PR #13847, opened 2024-08-01
lxd: Ensure file is created when receiving raw image

MusicDin:
In conversion API the file was not created on the server when receiving a raw image on directory storage pool.

In addition, there are two minor fixes:
- Remove unnecessary usage of VolumeCopy
- Set volume size directly on volume's config (instead of on args.Config which points to the same map).

### DIFF
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -2434,10 +2434,6 @@ func (b *lxdBackend) CreateInstanceFromConversion(inst instance.Instance, conn i
 		return err
 	}
 
-	// Override args.Name and args.Config to ensure volume is created based on instance.
-	args.Config = vol.Config()
-	args.Name = inst.Name()
-
 	// Get instance's root disk device from local devices. Do not use expanded devices, as we want
 	// to determine whether the root disk volume size was explicitly set by the client.
 	canResizeRootDiskSize := true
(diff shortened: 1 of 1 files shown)


lxd/storage/backend_lxd.go
```diff
@@ -1870,7 +1870,7 @@ func (b *lxdBackend) recvVolumeFiller(conn io.ReadWriteCloser, contentType drive
 			}
 		} else {
 			// Receive block volume.
-			to, err := os.OpenFile(rootBlockPath, os.O_WRONLY|os.O_TRUNC, 0)
+			to, err := os.OpenFile(rootBlockPath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0)
```
tomponline, line 1873: how come this worked before without this in <= 6.1?

Also if we are creating the file now, do we need to ensure the file is created with the same permissions as a normal empty block volume file for dir pools?

tomponline (approved):

lxd/storage/backend_lxd.go
MusicDin: > how come this worked before without this in <= 6.1?

Because in case of `migration` we first create the volume and then receive the image into it, not using the filler as in `conversion` case.

In case of `dir` storage pool the file is created (if missing) only once the filler functions has been executed.

https://github.com/canonical/lxd/blob/2a2bead82fefcc02039a92f2a437b3db3d80670c/lxd/storage/drivers/driver_dir_volumes.go#L65-L86

File is created within `ensureVolumeBlockFile` function:
https://github.com/canonical/lxd/blob/2a2bead82fefcc02039a92f2a437b3db3d80670c/lxd/storage/drivers/utils.go#L400-L405

MusicDin: > Also if we are creating the file now, do we need to ensure the file is created with the same permissions as a normal empty block volume file for dir pools?

Agree. I think 0600 should be sufficient.

tomponline: Is it the same as what is used for new volume files?

tomponline: > Because in case of migration we first create the volume and then receive the image into it, not using the filler as in conversion case.
Right makes sense thanks.

tomponline (approved):

PR #13848, opened 2024-08-01
lxd: Add tracker for qemu-img progress (from Incus)

MusicDin:
Track qemu-img convert progress.

Cherry-picked from https://github.com/lxc/incus/pull/1025 and https://github.com/lxc/incus/pull/1015


lxd/storage/backend_lxd.go
```diff
@@ -1912,7 +1912,7 @@ func (b *lxdBackend) recvBlockVol(toFile *os.File, volName string, conn io.ReadW
 
 	var wrapper *ioprogress.ProgressTracker
 	if args.TrackProgress {
-		wrapper = migration.ProgressTracker(op, "block_progress", volName)
+		wrapper = migration.ProgressTracker(op, "block_progress", "Transferring instance")
```
tomponline, line 1915: Are these ever used with non instance volumes?

tomponline (approved):

lxd/storage/backend_lxd.go
MusicDin: Currently, those are used only within conversion API. Therefore, they are used only for instance volume transfer

PR #13854, opened 2024-08-01
lxd/storage_pools: Don't crash on nil pool (from Incus)

kadinsayani:
cherry-picked from lxc/incus#359


tomponline:
@kadinsayani we may not need this one as LXD has a different authorizer structure. cc @markylaing

tomponline:
Added "(from Incus)" to PR description to help with attribution

PR #13859, opened 2024-08-02
shared/cmd: Allow a cmd asker to be created with a logger instance

gabrielmougard:
If a command asker is with a logger, the question and answer (even if it is invalid) are systematically logged to the logger destination. This will greatly help debugability in integration test suites like in MicroCloud which have a lot of interactions in the command line.

### DIFF
--- a/lxc/main.go
+++ b/lxc/main.go
@@ -89,7 +89,7 @@ For help with any of those, simply call them with --help.`))
 	app.CompletionOptions = cobra.CompletionOptions{DisableDefaultCmd: true}
 
 	// Global flags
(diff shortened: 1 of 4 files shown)


masnax (approved):
Sounds reasonable to me, though one nit I have is that `verbose` is kind of ambiguous for a feature like this which is intended for debugging. Not necessarily a blocker.

tomponline (changes requested):
As discussed the other day, lets change this to allow passing in a logger for verbose logging rather than outputting to stdout.

gabrielmougard:
@tomponline updated

tomponline:
@gabrielmougard linting issues need fixing

tomponline (approved):

PR #13864, opened 2024-08-03
lxd/firewall/drivers: Fix netprio error message

MggMuggins:
This was confusing when I ran into it; initially thought PyLXD had mixed up some json fields...


tomponline (approved):
Ta!

PR #13875, opened 2024-08-05
Improve VM shutdown

hamistao:
Includes cherry picks from https://github.com/lxc/incus/pull/362

### DIFF
--- a/lxd/instance_exec.go
+++ b/lxd/instance_exec.go
@@ -23,6 +23,7 @@ import (
 	"github.com/canonical/lxd/lxd/cluster"
 	"github.com/canonical/lxd/lxd/db/operationtype"
 	"github.com/canonical/lxd/lxd/instance"
+	"github.com/canonical/lxd/lxd/instance/drivers"
 	"github.com/canonical/lxd/lxd/instance/instancetype"
 	"github.com/canonical/lxd/lxd/operations"
 	"github.com/canonical/lxd/lxd/request"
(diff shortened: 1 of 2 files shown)


tomponline (commented):
Thanks for this!

Could we have a test added in lxd-ci VM tests that checks for the correct exit codes, I believe we have something in the main lxd test suite for this already for containers (if not we should add that too) and would be good to have something or VMs as well so we can catch regressions. Ta

tomponline:
@hamistao ready for review?

hamistao:
@tomponline This is still missing working tests for 129 exit code in containers, as there aren't any yet. I am still busy with the API metrics but I will return to this as soon as I finish the changes on the metrics PR.

lxd/instance_exec.go
```diff
@@ -289,7 +290,14 @@ func (s *execWs) Do(op *operations.Operation) error {
 			_ = pty.Close()
 		}
 
+		// Make VM disconnections (shutdown/reboot) match containers.
+		if cmdErr == drivers.ErrExecDisconnected {
+			cmdResult = 129
```
mihalicyn, line 295: exit code is `128+SIGHUP (=1)` = `129`. Looks good to me.

mihalicyn (approved):

tomponline:
@hamistao ready for a review, is there an accompanying test for this in lxd-ci?

hamistao:
@mihalicyn Thanks for the review, but my main question was regarding the failing tests. Although the VM case is currently working as expected, container disconnections result in a 137 exit code instead of 129 on my local tests. I wanted to confirm if this behavior can be considered a bug and if just catching this exit code and changing it to 129 would be an adequate solution or if there could be a deeper problem involved.

Also, the GH tests behave differentely from my local tests and return a 143 exit code, @simondeziel discovered that this exit code is also what we get when killing the agent process inside the instance to make it disconnect.

hamistao:
@tomponline This is ready for a review, the PR containing the tests was also opened [here](https://github.com/canonical/lxd-ci/pull/284). When both are merged I will also add tests for containers on the LXD repo.

tomponline (approved):

PR #13876, opened 2024-08-05
doc/tutorial: add a UI tutorial

ru-fu:
This commit also includes some small updates to the existing tutorial/UI instructions to make sure they are up-to-date.


doc/tutorial/ui.md
```diff
@@ -0,0 +1,422 @@
...
+```
+
+## Create and start instances
+
+Let's start by launching a few instances.
+With *instance*, we mean either a container or a virtual machine.
+See {ref}`containers-and-vms` for information about the difference between the two instance types.
+
+1. Select {guilabel}`Instances` from the navigation.
+1. Click {guilabel}`Create instance` to launch a container called `first` using the Ubuntu 24.04 LTS image:
```
mas-who, line 41: interesting the numbering here is all `1` but in the documentation preview it's correct. How does this work out of curiosity?

doc/tutorial/ui.md
```diff
@@ -0,0 +1,422 @@
...
+   See {ref}`instances-create-iso` for more information.
+   ```
+
+1. Click {guilabel}`Create and start` to launch the container.
+
+   ```{note}
+   Launching this container takes a few seconds, because the image must be downloaded and unpacked first.
+   ```
+
+1. Launch another container called `second`, using the same image.
```
mas-who, line 65: Is the reason for creating the `second` container to demonstrate the increase in speed since the image is now locally available? If so, I think creating the `third` container (even without starting) would demonstrate the same behaviour? Perhaps we can reduce the steps here by removing the creation of `second` container? This would of course require some adjustments for the rest of the document.

doc/tutorial/ui.md
```diff
@@ -0,0 +1,422 @@
...
+   ```
+
+  See {ref}`instances-console` for more information.
+- The {guilabel}`Logs` tab contains log files for inspection and download.
+  Running instances have only limited information.
+  More log files are added if an instance ends up in error state.
+
+  See {ref}`instances-troubleshoot` for more information.
+
+## Stop and delete instances
```
mas-who, line 170: related to the previous comment about potentially removing the `second` container creation. This section may need some adjustments

mas-who (commented):
@ru-fu Thanks for this write up, it is clear and looks great! Just a few minor comments for your consideration :)

doc/howto/access_ui.md
```diff
@@ -38,13 +40,17 @@ Complete the following steps to access the LXD web UI:
 1. Set up the certificates that are required for the UI client to authenticate with the LXD server by following the steps presented in the UI.
    These steps include creating a set of certificates, adding the private key to your browser, and adding the public key to the server's trust store.
 
+   If you already have a client certificate in your browser, you can choose to re-use the existing certificate and authorize it with the server.
```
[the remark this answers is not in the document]
ru-fu: I think you're right - it makes more sense to clearly present the two alternatives. Updated.

doc/tutorial/ui.md
ru-fu: It's a common feature in Markdown. The advantage is that if you use only `1.`, you won't mess anything up when you add another step somewhere later. :)

doc/tutorial/ui.md
ru-fu: Good catch - I removed the third one now. In the CLI tutorial, we create the third one by copying the second, but we don't have that in the UI.

tomponline (approved):

PR #13886, opened 2024-08-07
Effective project handling

### LINKED ISSUE #13863: Handling of project `features.*` in authorization
When any `features.*` are disabled on a project, entities of that type are created in the default project instead. 

Authorization checks are made with specific entity URLs. If the authorizer is not aware that the entity may not be in the project specified by the project query parameter in the URL, the authorization check may fail with a "not found" error.

I've investigated the work that is needed for this and outlined it below:
- [ ] In the embedded OpenFGA authorization driver, check for `request.CtxEffectiveProjectName` in the request context and handle appropriately.
- [ ] For `features.profiles`, the effective project is only set in the request context when listing profiles. If `features.profiles` is disabled for a project, getting/updating/deleting a single profile in that project works because the access check is using the request project. This will not work with the OpenFGA authorization driver, so we'll need to always set the effective project in the request context.
- [ ] For `features.networks` and `features.networks.zones` we have the same situation as with `features.profiles` above.
- [ ] For `features.images` we have a similar situation as with `features.profiles`, however we need to be careful to two key areas. Firstly, the process of getting the effective project is not standardised like it is for other entity types (with a function in the `lxd/project` package). Secondly, the database methods for getting/listing images are quite convoluted, and reperform this check on each call. I think we'll need to refactor a bit to simplify these methods by determining the effective project beforehand and passing it in.

I will treat each item here separately and create a PR for each.

markylaing:
Implements the following:
1. Always sets the effective project when querying entities that have associated `features.*` project config. This includes:
    i. Adding network, network zone, image, and profile specific access handlers.
    ii. For each entity type above, setting details in the request context to avoid repeated calculation (same pattern as for storage buckets and volumes).
2. Always uses the request project in calls to `(Authorizer).CheckPermission` and in calls to the `auth.PermissionChecker` returned by `(Authorizer).GetPermissionChecker`.
3. In the TLS driver, we remove effective project handling (we always expect calls to use the request project, this is what we check in their allowed project list).
4. In the OpenFGA driver, overwrite the request project with the effective project in calls to the embedded OpenFGA server. But do not "punch through" to the default project like with the TLS driver, as these permissions can be managed by an administrator.
5. Increased test coverage for project features with TLS authorization.
6. Adds tests for handling of project features with fine-grained authorization.

Closes #13863

### DIFF
--- a/doc/explanation/projects.md
+++ b/doc/explanation/projects.md
@@ -46,6 +46,15 @@ To edit them, you must remove all instances first.
 New features that are added in an upgrade are disabled for existing projects.
 ```
 
+```{important}
+In a multi-tenant environment, unless using {ref}`fine-grained-authorization`, all projects should have all features enabled.
+Otherwise, clients with {ref}`restricted-tls-certs` are able to create, edit, and delete resources in the default project. This might affect other tenants.
+
+For example, if project "foo" is created and `features.networks` is not set to true, then a restricted client certificate with access to "foo" can view, edit, and delete networks in the default project.
+
+Conversely, if a client's permissions are managed via {ref}`fine-grained-authorization`, resources may be inherited from the default project but access to those resources is not automatically granted.
+```
+
 (projects-confined)=
 ## Confined projects in a multi-user environment
 
--- a/lxd/images.go
+++ b/lxd/images.go
@@ -66,46 +66,147 @@ var imagesCmd = APIEndpoint{
 var imageCmd = APIEndpoint{
 	Path: "images/{fingerprint}",
 
-	Delete: APIEndpointAction{Handler: imageDelete, AccessHandler: allowPermission(entity.TypeImage, auth.EntitlementCanDelete, "fingerprint")},
+	Delete: APIEndpointAction{Handler: imageDelete, AccessHandler: imageAccessHandler(auth.EntitlementCanDelete)},
 	Get:    APIEndpointAction{Handler: imageGet, AllowUntrusted: true},
-	Patch:  APIEndpointAction{Handler: imagePatch, AccessHandler: allowPermission(entity.TypeImage, auth.EntitlementCanEdit, "fingerprint")},
-	Put:    APIEndpointAction{Handler: imagePut, AccessHandler: allowPermission(entity.TypeImage, auth.EntitlementCanEdit, "fingerprint")},
+	Patch:  APIEndpointAction{Handler: imagePatch, AccessHandler: imageAccessHandler(auth.EntitlementCanEdit)},
+	Put:    APIEndpointAction{Handler: imagePut, AccessHandler: imageAccessHandler(auth.EntitlementCanEdit)},
 }
 
 var imageExportCmd = APIEndpoint{
 	Path: "images/{fingerprint}/export",
 
 	Get:  APIEndpointAction{Handler: imageExport, AllowUntrusted: true},
-	Post: APIEndpointAction{Handler: imageExportPost, AccessHandler: allowPermission(entity.TypeImage, auth.EntitlementCanEdit, "fingerprint")},
+	Post: APIEndpointAction{Handler: imageExportPost, AccessHandler: imageAccessHandler(auth.EntitlementCanEdit)},
 }
 
 var imageSecretCmd = APIEndpoint{
 	Path: "images/{fingerprint}/secret",
 
-	Post: APIEndpointAction{Handler: imageSecret, AccessHandler: allowPermission(entity.TypeImage, auth.EntitlementCanEdit, "fingerprint")},
+	Post: APIEndpointAction{Handler: imageSecret, AccessHandler: imageAccessHandler(auth.EntitlementCanEdit)},
 }
 
 var imageRefreshCmd = APIEndpoint{
 	Path: "images/{fingerprint}/refresh",
 
-	Post: APIEndpointAction{Handler: imageRefresh, AccessHandler: allowPermission(entity.TypeImage, auth.EntitlementCanEdit, "fingerprint")},
+	Post: APIEndpointAction{Handler: imageRefresh, AccessHandler: imageAccessHandler(auth.EntitlementCanEdit)},
 }
 
 var imageAliasesCmd = APIEndpoint{
 	Path: "images/aliases",
 
-	Get:  APIEndpointAction{Handler: imageAliasesGet, AccessHandler: allowAuthenticated},
+	Get:  APIEndpointAction{Handler: imageAliasesGet, AccessHandler: allowProjectResourceList},
 	Post: APIEndpointAction{Handler: imageAliasesPost, AccessHandler: allowPermission(entity.TypeProject, auth.EntitlementCanCreateImageAliases)},
 }
 
 var imageAliasCmd = APIEndpoint{
 	Path: "images/aliases/{name:.*}",
 
-	Delete: APIEndpointAction{Handler: imageAliasDelete, AccessHandler: allowPermission(entity.TypeImageAlias, auth.EntitlementCanDelete, "name")},
+	Delete: APIEndpointAction{Handler: imageAliasDelete, AccessHandler: imageAliasAccessHandler(auth.EntitlementCanDelete)},
 	Get:    APIEndpointAction{Handler: imageAliasGet, AllowUntrusted: true},
-	Patch:  APIEndpointAction{Handler: imageAliasPatch, AccessHandler: allowPermission(entity.TypeImageAlias, auth.EntitlementCanEdit, "name")},
-	Post:   APIEndpointAction{Handler: imageAliasPost, AccessHandler: allowPermission(entity.TypeImageAlias, auth.EntitlementCanEdit, "name")},
-	Put:    APIEndpointAction{Handler: imageAliasPut, AccessHandler: allowPermission(entity.TypeImageAlias, auth.EntitlementCanEdit, "name")},
+	Patch:  APIEndpointAction{Handler: imageAliasPatch, AccessHandler: imageAliasAccessHandler(auth.EntitlementCanEdit)},
+	Post:   APIEndpointAction{Handler: imageAliasPost, AccessHandler: imageAliasAccessHandler(auth.EntitlementCanEdit)},
+	Put:    APIEndpointAction{Handler: imageAliasPut, AccessHandler: imageAliasAccessHandler(auth.EntitlementCanEdit)},
+}
+
+const ctxImageDetails request.CtxKey = "image-details"
+
+// imageDetails contains fields that are determined prior to the access check. This is set in the request context when
+// addImageDetailsToRequestContext is called.
+type imageDetails struct {
+	imageFingerprintPrefix string
+	imageID                int
+	image                  api.Image
+}
+
+// addImageDetailsToRequestContext sets request.CtxEffectiveProjectName (string) and ctxImageDetails (imageDetails)
+// in the request context.
+func addImageDetailsToRequestContext(s *state.State, r *http.Request) error {
+	imageFingerprintPrefix, err := url.PathUnescape(mux.Vars(r)["fingerprint"])
+	if err != nil {
+		return err
+	}
+
+	requestProjectName := request.ProjectParam(r)
+	effectiveProjectName := requestProjectName
+	var imageID int
+	var image *api.Image
+	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+		effectiveProjectName, err = projectutils.ImageProject(ctx, tx.Tx(), requestProjectName)
+		if err != nil {
+			return err
+		}
+
+		imageID, image, err = tx.GetImageByFingerprintPrefix(ctx, imageFingerprintPrefix, dbCluster.ImageFilter{Project: &requestProjectName})
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("Failed to check project %q image feature: %w", requestProjectName, err)
+	}
+
+	request.SetCtxValue(r, request.CtxEffectiveProjectName, effectiveProjectName)
+	request.SetCtxValue(r, ctxImageDetails, imageDetails{
+		imageFingerprintPrefix: imageFingerprintPrefix,
+		imageID:                imageID,
+		image:                  *image,
+	})
+
+	return nil
+}
+
+func imageAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *http.Request) response.Response {
+	return func(d *Daemon, r *http.Request) response.Response {
+		s := d.State()
+		err := addImageDetailsToRequestContext(s, r)
+		if err != nil {
+			return response.SmartError(err)
+		}
+
+		details, err := request.GetCtxValue[imageDetails](r.Context(), ctxImageDetails)
+		if err != nil {
+			return response.SmartError(err)
+		}
+
+		err = s.Authorizer.CheckPermission(r.Context(), entity.ImageURL(request.ProjectParam(r), details.image.Fingerprint), entitlement)
+		if err != nil {
+			return response.SmartError(err)
+		}
+
+		return response.EmptySyncResponse
+	}
+}
+
+func imageAliasAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *http.Request) response.Response {
+	return func(d *Daemon, r *http.Request) response.Response {
+		imageAliasName, err := url.PathUnescape(mux.Vars(r)["name"])
+		if err != nil {
+			return response.SmartError(err)
+		}
+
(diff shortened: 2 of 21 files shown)


markylaing:
I've created this issue to refactor the image database methods. For now there is an extra database call in image and image alias handlers to resolve the effective project and set it in the request context before the access check.

tomponline:
> I've created this issue to refactor the image database methods. For now there is an extra database call in image and image alias handlers to resolve the effective project and set it in the request context before the access check.

Missing link?

markylaing:
> > I've created this issue to refactor the image database methods. For now there is an extra database call in image and image alias handlers to resolve the effective project and set it in the request context before the access check.
> 
> Missing link?

Whoops, I've added it to the original comment.

test/suites/tls_restrictions.sh
```diff
@@ -96,43 +95,240 @@ test_tls_restrictions() {
...
-  test_image_fingerprint="$(lxc_remote image list localhost: --format csv --columns f --project blah)"
   lxc_remote image delete "localhost:${test_image_fingerprint}" --project blah
 
   # Ensure we can create and view resources that are not enabled for the project (e.g. their effective project is
   # the default project).
 
-  # Networks are disabled when projects are created.
+  ### IMAGES (initial value is true for new projects)
+
+  # Unset the images feature (the default is false).
```
simondeziel, line 108: Isn't the default value `true`? If so, the there are many such comment that would need fixing.

test/suites/tls_restrictions.sh
```diff
@@ -96,43 +95,240 @@ test_tls_restrictions() {
...
+  lxc_remote image import "${TEST_DIR}/${test_image_fingerprint}.tar.xz" localhost: --project blah
+
+  # Clean up
+  lxc_remote image delete "localhost:${test_image_fingerprint}" --project blah
+
+
+  ### NETWORKS (initial value is false in new projects).
+
+  # Create a network in the default project.
+  networkName="net$(tr -dc A-Za-z0-9 </dev/urandom | head -c 3)"
```
simondeziel, line 139: Same comment about `$$`.

test/suites/tls_restrictions.sh
```diff
@@ -96,43 +95,240 @@ test_tls_restrictions() {
...
+  ! lxc_remote network show localhost:blah-network --project default || false
+
+  # The restricted client can delete the network.
+  lxc_remote network delete localhost:blah-network --project blah
+
+
+  ### NETWORK ZONES (initial value is false in new projects).
+
+  # Create a network zone in the default project.
+  zoneName="zone$(tr -dc A-Za-z0-9 </dev/urandom | head -c 3)"
```
simondeziel, line 176: Same comment about `$$`.

test/suites/tls_restrictions.sh
```diff
@@ -96,43 +95,240 @@ test_tls_restrictions() {
...
   lxc_remote network zone delete localhost:blah-zone --project blah
 
+
+  ### PROFILES (initial value is true for new projects)
+
   # Unset the profiles feature (the default is false).
   lxc project unset blah features.profiles
+
+  # Create a profile in the default project.
+  profileName="prof$(tr -dc A-Za-z0-9 </dev/urandom | head -c 3)"
```
simondeziel, line 216: Same comment about `$$`.

test/suites/tls_restrictions.sh
```diff
@@ -96,43 +95,240 @@ test_tls_restrictions() {
...
   lxc_remote profile delete localhost:blah-profile --project blah
 
+
+  ### STORAGE VOLUMES (initial value is true for new projects)
+
   # Unset the storage volumes feature (the default is false).
   lxc project unset blah features.storage.volumes
+
+  # Create a storage volume in the default project.
+  volName="vol$(tr -dc A-Za-z0-9 </dev/urandom | head -c 3)"
```
simondeziel, line 256: Same comment about `$$`.

test/suites/tls_restrictions.sh
```diff
@@ -96,43 +95,240 @@ test_tls_restrictions() {
...
   lxc_remote storage volume delete "localhost:${pool_name}" blah-volume --project blah
 
+  ### STORAGE BUCKETS (initial value is true for new projects)
+  create_object_storage_pool s3
+
+  # Unset the storage buckets feature (the default is false).
+  lxc project unset blah features.storage.buckets
+
+  # Create a storage bucket in the default project.
+  bucketName="bucket$(tr -dc a-z0-9 </dev/urandom | head -c 3)"
```
simondeziel, line 296: Same comment about `$$`.

test/includes/storage.sh
```diff
@@ -130,3 +130,56 @@ umount_loops() {
...
+    echo "Storage pool pool ${poolName} already exists"
+    exit 1
+  fi
+
+  # Check cephobject.radosgw.endpoint is required for cephobject pools.
+  if [ "${lxd_backend}" = "ceph" ]; then
+    lxc storage create "${poolName}" cephobject cephobject.radosgw.endpoint="${LXD_CEPH_CEPHOBJECT_RADOSGW}"
+  else
+
+    # Create a loop device for dir pools as MinIO doesn't support running on tmpfs (which the test suite can do).
```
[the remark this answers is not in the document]
markylaing: I didn't actually investigate this as it was copied from the `storage_buckets` test. I'll add a comment for it :)

test/includes/storage.sh
```diff
@@ -130,3 +130,56 @@ umount_loops() {
...
+    if [ "${lxd_backend}" = "dir" ]; then
+      mkdir -p "${TEST_DIR}/s3/${poolName}"
+      configure_loop_device loop_file_1 loop_device_1
+      # shellcheck disable=SC2154
+      mkfs.ext4 "${loop_device_1}"
+      mount "${loop_device_1}" "${TEST_DIR}/s3/${poolName}"
+      losetup -d "${loop_device_1}"
```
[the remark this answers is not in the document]
markylaing: Yeah that is a bit strange. Again I copied this from the storage buckets test so I didn't think about it too much. This is the commit that introduced it: https://github.com/canonical/lxd/commit/6ef7de504cbc8657d7d7e80c21ebd3aaaba123b6

I have removed it and the tests are still passing. The device is still being detached via `delete_object_storage_pool` and `deconfigure_loop_device`.

test/suites/tls_restrictions.sh
markylaing: No, the default value is false for all project features. The initial value upon creating a project is true for images, storage volumes, storage buckets, and profiles, and false for networks and network zones.

simondeziel: At first I thought you were contradicting yourself but then I realized that LXD actually sets those config so unsetting brings a default that is different than the one you get on creation. Quite confusing.

```bash
$ lxc project create blah
Project blah created
$ lxc project show blah
name: blah
description: ""
config:
  features.images: "true"
  features.profiles: "true"
  features.storage.buckets: "true"
  features.storage.volumes: "true"
used_by:
- /1.0/profiles/default?project=blah
$ lxc project unset blah features.images
$ lxc project show blah
name: blah
description: ""
config:
  features.profiles: "true"
  features.storage.buckets: "true"
  features.storage.volumes: "true"
used_by:
- /1.0/profiles/default?project=blah
```

simondeziel (approved):

test/suites/tls_restrictions.sh
tomponline: These are so-called "create time defaults", similar to https://documentation.ubuntu.com/lxd/en/latest/reference/network_bridge/#network-bridge-network-conf:ipv6.nat

lxd/profiles.go
```diff
@@ -43,11 +43,70 @@ var profilesCmd = APIEndpoint{
...
+// profileDetails contains fields that are determined prior to the access check. This is set in the request context when
+// addProfileDetailsToRequestContext is called.
+type profileDetails struct {
+	profileName      string
+	effectiveProject api.Project
+}
+
+// addProfileDetailsToRequestContext sets request.CtxEffectiveProjectName (string) and ctxProfileDetails (profileDetails)
+// in the request context.
+func addProfileDetailsToRequestContext(d *Daemon, r *http.Request) error {
```
tomponline, line 65: Is there a reason you're requiring *Daemon here, rather than *state.State as an argument? It doesn't look like you're using d for any other reason that to access d.State(), which could be passed in from the caller, which better indicates the requirements for this function.

lxd/profiles.go
```diff
@@ -43,11 +43,70 @@ var profilesCmd = APIEndpoint{
...
+	s := d.State()
+	profileName, err := url.PathUnescape(mux.Vars(r)["name"])
+	if err != nil {
+		return err
+	}
+
+	requestProjectName := request.ProjectParam(r)
+	effectiveProject, err := project.ProfileProject(s.DB.Cluster, requestProjectName)
+	if err != nil && api.StatusErrorCheck(err, http.StatusNotFound) {
+		return api.StatusErrorf(http.StatusNotFound, http.StatusText(http.StatusNotFound))
```
tomponline, line 75: Why are you masking the error message from project.ProfileProject() with a generic "Not found" message rather than using SmartResponse and which would pass the "profile not found" type error to the caller?

tomponline (commented):
So in general 2 questions:

1. Can we avoid passing *Daemon to `addFooDetailsToRequestContext`?
2. Why are the not found errors from `project.FooRecord()` being masked, and if there is a reason, we should have a comment explaining it, as its non-obvious.

doc/explanation/projects.md
```diff
@@ -46,6 +46,15 @@ To edit them, you must remove all instances first.
 New features that are added in an upgrade are disabled for existing projects.
 ```
 
+```{important}
+In a multi-tenant environment, all projects should have all features enabled.
```
tomponline, line 50: Can we clarify this to say that its only recommend when using non-fine grained identities?

tomponline:
@markylaing please can you look at whether `func eventsSocket(s *state.State, r *http.Request, w http.ResponseWriter) error {` is doing the right thing (WRT to projects/all projects ) whilst you're working in this area? Thanks

lxd/profiles.go
tomponline: > The logic is starting to leak into lots of places around the code base though, I'm starting to think that `response.SmartError` should handle this instead. E.g. if `auth.IsDeniedError` we return a masked error message, otherwise we return the true error.

I think that makes sense.

> We're doing this masking in lots of places now to avoid resource discovery. If the project really doesn't exist, the user will get `Project not found`. If the project does exist but the user does not have access to it, they will get `Not found`.

I do remember discussing that. Although the way you've described it here makes me think, can an end user still figure out if a resource exists by comparing the error strings instead of the response codes?

doc/explanation/projects.md
markylaing: I think it's fine to disable features when using fine-grained identities. The administrator just needs to be aware that they will need to grant some level of permission to the group that allows them to interact with resources in the default project.

E.g.
```
lxc auth group create foo-group
lxc project create foo-project # Has networks disabled.
lxc auth group permission add foo-group project foo-project operator
lxc auth group permission add foo-group project default can_view_networks
```

`foo-group` can now create instances in `foo-project` using networks from the default project, but can't edit or delete the networks in the default project.

tomponline: exactly :) Which is why I don't think we should blanket say "In a multi-tenant environment, all projects should have all features enabled."

markylaing:
> @markylaing please can you look at whether `func eventsSocket(s *state.State, r *http.Request, w http.ResponseWriter) error {` is doing the right thing (WRT to projects/all projects ) whilst you're working in this area? Thanks

The events socket is doing the right thing WRT `project`/`all-projects`. If `all-projects` is passed, we get a permission checker with `can_view_events` on projects and use that for each event that is to be sent to the client. If a project name is given, we check that the identity has `can_view_events` on that project and deny the request if they don't, then only send events that are in that project.

This breaks down for effective projects of course. If an event occurs in the default project because the feature is not enabled in another project, we can't resolve the original request project because there may be more than one project with that feature disabled.

tomponline:
> This breaks down for effective projects of course. If an event occurs in the default project because the feature is not enabled in another project, we can't resolve the original request project because there may be more than one project with that feature disabled.

Mmm yeah that is what I was getting at. So this would mean that specifying a specific project which has certain features disabled would mean missing events for entities related to those features ? But it wouldn't mean seeing events that one shouldnt?

tomponline:
> > > This breaks down for effective projects of course. If an event occurs in the default project because the feature is not enabled in another project, we can't resolve the original request project because there may be more than one project with that feature disabled.
> > 
> > 
> > Mmm yeah that is what I was getting at. So this would mean that specifying a specific project which has certain features disabled would mean missing events for entities related to those features ? But it wouldn't mean seeing events that one shouldnt?
> 
> Yes exactly. E.g.
> 
> ```
> lxc project create foo-project # Networks and network zones disabled by default
> lxc monitor --project foo-project # Will not receive events for networks or network zones.
> ```
> 
> This would be true for both fine-grained or TLS users, though with fine-grained users you can grant `can_view_events` in the default project. Whereas TLS users would have to be granted full access to the default project to be able to see them.

Would you mind adding a comment to the code there explaining this for future readers.

I wonder if its worth opening an issue to think about whether we can/should address this?

lxd/profiles.go
markylaing: Fixed. I should have been using state, however the distinction between `*Deamon` and `*state.State` isn't clear to me in LXD. If the state is a snapshot of the daemon state at the time the request is received, I'd say this should be evaluated at the top-level mux handler passed into both the access handler and main handler.

lxd/profiles.go
markylaing: I've updated SmartError to mask not found and forbidden errors to use the default HTTP status text.

The end user can figure out if a resource exists by the error text in some cases I think. I haven't done a full audit of this. It wouldn't work for project names as we don't need to do any logic before the authorization check. The general case is as follows:
1. Caller makes request to view/edit/delete a specific entity.
2. The entity requires some pre-processing before we can perform the access check. This can be for the effective project, or it can be to expand an image or certificate fingerprint.
3. The error returned in the pre-processing step may indicate the resource does not exist with message like "<thing> not found.
4. If the resource is found but the caller does not have permission to view it, the authorizer returns "Not Found".

If the caller gets "Not Found" and not "<thing> not found", they can surmise that the entity does exist. In most cases they already need to be authenticated for this, but some endpoints (images in particular) allow untrusted requests.

I think updating SmartError will catch the majority of cases and I don't think it's a general concern as an attack vector, I'm just being thorough :)

doc/explanation/projects.md
markylaing: Ahhh I see what you mean now :) will update

markylaing: Done.

markylaing:
> Would you mind adding a comment to the code there explaining this for future readers.
> 
> I wonder if its worth opening an issue to think about whether we can/should address this?

I've added a comment in the event websocket handler and have also created this issue #13962.

lxd/response/smart.go
```diff
@@ -23,12 +25,25 @@ func SmartError(err error) Response {
 
 	statusCode, found := api.StatusErrorMatch(err)
 	if found {
+		// If the status code relates to an authentication error, return the generic HTTP status text.
+		// This is so that there is no difference in the error message when
+		// - a resource really doesn't exist (e.g. "Instance not found").
```
tomponline, line 30: Ah so this isn't going to work, because as far as I can tell, we are now always converting "instance not found" type errors into generic "not found" errors. Which undoes all the work we've done over the last few years to pass more intelligible errors back to the end user - i.e if you are doing something on an instance that depends on a profile and the profile doesn't exist you get a 404 "profile not found" rather than a 404 "not found" and you can't tell if its the instance or the profile that is not found.

This would be a retrograde step.

tomponline (changes requested):
I think we should decouple this PR from the 404 not found discussion so as not to prevent this from being merged. Can you rework this PR to not do any error masking and then we can revisit the discussion as a separate issue?

lxd/response/smart.go
markylaing: Sure thing. We only need to mask errors in the pre-processing step before the call to the authorizer. So I can revert back to the previous method.

markylaing: Done. Will re-request a review once the tests pass.

simondeziel (approved):

lxd/images.go
```diff
@@ -3001,7 +3081,13 @@ func imageGet(d *Daemon, r *http.Request) response.Response {
 	// Get the image. We need to do this before the permission check because the URL in the permission check will not
 	// work with partial fingerprints.
 	var info *api.Image
+	effectiveProjectName := projectName
```
tomponline, line 3084: `var effectiveProjectName string` would be clearer here no? Showing that its empty until its loaded from the DB below, rather than defaulting to the request project?

PR #13890, opened 2024-08-07
VM usb nic passthrough (from Incus)

MggMuggins:
What constitutes adequate testing for this set of changes? I don't have access (currently) to a USB/Ethernet adapter; while I'm sure it's _possible_ to simulate one I doubt that would be a straightforward/worthwhile process. Should I check if someone else on the team has appropriate hardware to run a quick check on this?

### DIFF
--- a/lxd/instance/instancetype/instance.go
+++ b/lxd/instance/instancetype/instance.go
@@ -1210,10 +1210,20 @@ func ConfigKeyChecker(key string, instanceType Type) (func(value string) error, 
 			return validate.IsAny, nil
 		}
 
+		// lxdmeta:generate(entities=network-physical; group=volatile; key=volatile.last_state.usb.bus)
+		//
+		// --
+		//  type: string
+		//  shortdesc: USB Bus Number
 		if strings.HasSuffix(key, ".bus") {
 			return validate.IsAny, nil
 		}
 
+		// lxdmeta:generate(entities=network-physical; group=volatile; key=volatile.last_state.usb.device)
+		//
+		// --
+		//  type: string
+		//  shortdesc: USB Device Number
 		if strings.HasSuffix(key, ".device") {
 			return validate.IsAny, nil
 		}
--- a/lxd/instance/instancetype/instance.go
+++ b/lxd/instance/instancetype/instance.go
@@ -1210,6 +1210,14 @@ func ConfigKeyChecker(key string, instanceType Type) (func(value string) error, 
 			return validate.IsAny, nil
 		}
 
+		if strings.HasSuffix(key, ".bus") {
+			return validate.IsAny, nil
+		}
+
+		if strings.HasSuffix(key, ".device") {
+			return validate.IsAny, nil
+		}
+
 		if strings.HasSuffix(key, ".uuid") {
 			return validate.IsAny, nil
 		}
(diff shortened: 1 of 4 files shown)


simondeziel:
> What constitutes adequate testing for this set of changes? I don't have access (currently) to a USB/Ethernet adapter;

When I plug my iphone to my laptop, an `AppleUSBEthernet` NIC appears, maybe you could use your phone for a local test? I could also do the test for you if you need.

> while I'm sure it's _possible_ to simulate one I doubt that would be a straightforward/worthwhile process.

Yeah, last I checked, simulating one was not easy which is why we still have no USB test part of `lxd-ci` :/

> Should I check if someone else on the team has appropriate hardware to run a quick check on this?

If the phone trick doesn't work, I also have cheap USB Wifi NICs laying around so let me know.

MggMuggins:
Had to enable tethering on my phone but that works on Android, I'll see what I can do. Thanks.

MggMuggins:
I sideloaded a `lxd.debug` into `latest/edge` and was able to pass my phone through to the VM. This is good to go!

lxd/instance/instancetype/instance.go
```diff
@@ -1210,6 +1210,14 @@ func ConfigKeyChecker(key string, instanceType Type) (func(value string) error,
 			return validate.IsAny, nil
 		}
 
+		if strings.HasSuffix(key, ".bus") {
```
tomponline, line 1213: Please can you add the lxdmeta:generate documentation for these 2 new keys. Thanks

tomponline (changes requested):
Looks good thanks.

Just needs the new volatile keys documenting.

MggMuggins:
Added the docstrings; it doesn't look like this combination of `entities=network-physical` and `group=volatile` is being used anywhere in the docs. The [volatile instance keys](https://documentation.ubuntu.com/lxd/en/latest/reference/instance_options/#volatile-internal-data) have a section dedicated, but the nic option docs are [nested pretty deeply](https://documentation.ubuntu.com/lxd/en/latest/reference/devices_nic/#nictype-physical), so I'm not sure if it makes sense to include the volatile keys. @ru-fu do you have memory/opinion on this?

tomponline (approved):

ru-fu:
> Added the docstrings; it doesn't look like this combination of `entities=network-physical` and `group=volatile` is being used anywhere in the docs. The [volatile instance keys](https://documentation.ubuntu.com/lxd/en/latest/reference/instance_options/#volatile-internal-data) have a section dedicated, but the nic option docs are [nested pretty deeply](https://documentation.ubuntu.com/lxd/en/latest/reference/devices_nic/#nictype-physical), so I'm not sure if it makes sense to include the volatile keys. @ru-fu do you have memory of previous discussion/an opinion on this?

No previous discussions about this.
My suggestion would be to just add the keys to the common `network-conf` group (and maybe add a sentence to the description that they are internal and shouldn't be changed).

In general, we break up the set of config options when there are a lot and "it makes sense", but there isn't a strict rule on when that needs to happen. And in this case, they will appear at the end of the list anyway, so I think it's totally fine to add them to the list [here](https://documentation.ubuntu.com/lxd/en/latest/reference/network_physical/#configuration-options).

PR #13897, opened 2024-08-08
lxd/instance: Allow importing snapshots with non-alphanumeric names

kadinsayani:
Bug fix for issue #13879: "lxc import" fails if a snapshot name contains non-alphanumeric characters.


tomponline:
@kadinsayani whilst https://github.com/canonical/lxd/pull/13899 fixes the headline issue, please could you retain this PR and remove the fix commits, but leave the other linter fixes you've picked up. Thanks

PR #13900, opened 2024-08-09
Network: Switch to long-lived OVN ports

tomponline:
- [x] Convert OVN logical switch ports and associated config to be created at NIC create time and deleted at NIC delete time (rather than start and stop time respectively).
- [x] Update tests to reflect this. See https://github.com/canonical/lxd-ci/pull/261
- [x] Add DB patch to clear out `volatile.*last_state.ip_addresses` instance config entries as no longer needed.
- [x] Explore if we can have a network level option to allow the OVN dynamic IPs to be refreshed on start and cleared on stop to allow the old dynamic behaviour as well, otherwise it won't be possible to create more instances than available IPs, even if not all of them are running at the same time.
- [x] Test live VM migration works - tested using both ceph and zfs migrations.

Confirmed that creating an OVN NIC when there is insufficient space causes an error:

```
Error: Failed instance creation: Failed creating instance record: Failed initialising instance: Failed to add device "eth0": Failed setting up OVN port: Insufficient dynamic addresses allocated
```

WRT to exploring if we can later add a network option to allow dynamic allocation at start time rather than create time, this should be possible to do by not setting the `dynamic` option when we create the port, but instead setting it at start time and removing it at stop time. **But will not add this functionality in this PR.**

E.g.

```
sudo ovn-nbctl find logical_switch_port
addresses           : ["00:16:3e:24:9c:4c"]
dhcpv4_options      : 592a6788-4c73-4773-a8fa-12808377eccf
dhcpv6_options      : 52436e24-f9d1-4436-9de4-96a264739aed
dynamic_addresses   : []
enabled             : []
external_ids        : {lxd_location=none, lxd_switch=lxd-net14-ls-int}
ha_chassis_group    : []
name                : lxd-net14-instance-3dd0d44c-f72f-4981-b04c-bdf6414e0bba-eth0
options             : {}
parent_name         : []
port_security       : []
tag                 : []
tag_request         : []
type                : ""
up                  : false
```

And then:

```
sudo ovn-nbctl lsp-set-addresses lxd-net14-instance-3dd0d44c-f72f-4981-b04c-bdf6414e0bba-eth0 00:16:3e:24:9c:4c dynamic
```

Will result in dynamic IPs being allocated:

```
addresses           : ["00:16:3e:24:9c:4c", dynamic]
dynamic_addresses   : "be:af:33:b8:e5:03 10.184.229.2 fd42:843c:cb92:3d28:bcaf:33ff:feb8:e503"
```

Then removing `dynamic`:

```
sudo ovn-nbctl lsp-set-addresses lxd-net14-instance-3dd0d44c-f72f-4981-b04c-bdf6414e0bba-eth0 00:16:3e:24:9c:4c
```

Causes the dynamic IPs to be removed too:

```
addresses           : ["00:16:3e:24:9c:4c"]
dynamic_addresses   : []
```

It would also need the DNS setup to be delayed until NIC start time and have it torn down at NIC stop time as otherwise we won't have IPs to associate to the DNS name.

### DIFF
--- a/lxd/device/nic_ovn.go
+++ b/lxd/device/nic_ovn.go
@@ -343,7 +343,41 @@ func (d *nicOVN) checkAddressConflict() error {
 
 // Add is run when a device is added to a non-snapshot instance whether or not the instance is running.
 func (d *nicOVN) Add() error {
-	return d.network.InstanceDevicePortAdd(d.inst.LocalConfig()["volatile.uuid"], d.name, d.config)
+	networkVethFillFromVolatile(d.config, d.volatileGet())
+
+	// Load uplink network config.
+	uplinkNetworkName := d.network.Config()["network"]
+
+	var err error
+	var uplink *api.Network
+
+	err = d.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+		_, uplink, _, err = tx.GetNetworkInAnyState(ctx, api.ProjectDefaultName, uplinkNetworkName)
+
+		return err
+	})
+	if err != nil {
+		return fmt.Errorf("Failed to load uplink network %q: %w", uplinkNetworkName, err)
+	}
+
+	err = d.network.InstanceDevicePortAdd(d.inst.LocalConfig()["volatile.uuid"], d.name, d.config)
+	if err != nil {
+		return err
+	}
+
+	// Add new OVN logical switch port for instance.
+	_, err = d.network.InstanceDevicePortStart(&network.OVNInstanceNICSetupOpts{
+		InstanceUUID: d.inst.LocalConfig()["volatile.uuid"],
+		DNSName:      d.inst.Name(),
+		DeviceName:   d.name,
+		DeviceConfig: d.config,
+		UplinkConfig: uplink.Config,
+	}, nil)
+	if err != nil {
+		return fmt.Errorf("Failed setting up OVN port: %w", err)
+	}
+
+	return nil
 }
 
 // PreStartCheck checks the managed parent network is available (if relevant).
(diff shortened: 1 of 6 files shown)


tomponline:
@MggMuggins

lxd/network/driver_ovn.go
```diff
@@ -4252,17 +4252,15 @@ func (n *ovn) InstanceDevicePortIPs(instanceUUID string, deviceName string) ([]n
 
 // InstanceDevicePortStop deletes an instance device port from the internal logical switch.
 func (n *ovn) InstanceDevicePortStop(ovsExternalOVNPort openvswitch.OVNSwitchPort, opts *OVNInstanceNICStopOpts) error {
```
roosterfish: Worth removing this func too if it just returns nil? Maybe it can be even removed from the `ovnNet` interface?

roosterfish: Nvm, it's removed in the next commit.
Is it worth not having the change to `return nil` in this commit?

lxd/device/nic_ovn.go
```diff
@@ -554,45 +554,18 @@ func (d *nicOVN) Start() (*deviceConfig.RunConfig, error) {
...
-	for _, ipStr := range shared.SplitNTrimSpace(v["last_state.ip_addresses"], ",", -1, true) {
-		lastStateIP := net.ParseIP(ipStr)
-		if lastStateIP != nil {
-			lastStateIPs = append(lastStateIPs, lastStateIP)
-		}
-	}
-
 	// Add new OVN logical switch port for instance.
-	logicalPortName, dnsIPs, err := d.network.InstanceDevicePortStart(&network.OVNInstanceNICSetupOpts{
+	logicalPortName, _, err := d.network.InstanceDevicePortStart(&network.OVNInstanceNICSetupOpts{
```
roosterfish: Worth getting rid of the returned `dnsIPs` argument from `InstanceDevicePortStart()`?

This now looks to be not used anymore.

roosterfish: Nvm, also done in one of the next commits.

roosterfish (approved):
Looks good! Only one comment.

lxd/network/driver_ovn.go
tomponline: ah just saw this now, yeah potentially, although it showed my logical steps towards the final result.

Issue #13887, opened 2024-08-07
Token can be created with name `--help`


MusicDin:
I've accidentally created a token with name `--help`, but I cannot remove it using `lxc` because actual help message will be shown.

```
$ lxc config trust add --name --help

$ lxc config trust list-tokens
+-----------+-------+------------+
| NAME      | TOKEN | EXIPRES AT |
+-----------+-------+------------+
| --help    |  ...  |            |
+-----------+-------+------------+

$ lxc config trust revoke-token '--help'
Description:
  Revoke certificate add token

Usage:
  lxc config trust revoke-token [<remote>:] <name> [flags]

Global Flags:
      --debug          Show all debug messages
      --force-local    Force using the local unix socket
  -h, --help           Print help
      --project        Override the source project
  -q, --quiet          Don't show progress information
      --sub-commands   Use with help or --help to view sub-commands
  -v, --verbose        Show all information messages
      --version        Print version number
```

MusicDin:
The token can be removed using `lxc config trust revoke-token -- --help`, but I still think we should prevent such names?

kadinsayani:
Should we just disallow names including ```--help```, or any name with the prefix ```--``` or ```-```?

kadinsayani:
Alternatively, we can disallow all names that are also command or global flags.

tomponline:
Option 2 at least please

kadinsayani:
The ```--help``` flag is defined as a ```PersistentFlag``` in ```lxc/main.go```, which correctly applies the help flag across all commands. However, the ```lxc``` cli uses ```Flags().StringVar()``` to accept input when the ```--name``` flag is passed, which allows any input to be accepted without validation.

I think we have two solutions here:
1. Add input validation to the ```lxc config trust add``` command; and/or
2. Add input validation on the LXD server side.

PR #13902, opened 2024-08-09
lxd/certificates: Add input validation for certificate names

kadinsayani:
Bug fix for issue #13887

### DIFF
--- a/lxd/certificates.go
+++ b/lxd/certificates.go
@@ -12,6 +12,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/gorilla/mux"
@@ -437,6 +438,11 @@ func certificatesPost(d *Daemon, r *http.Request) response.Response {
 
 	if req.Password != "" {
 		return response.NotImplemented(fmt.Errorf("Password authentication is no longer supported, please update your client"))
+	}
+
+	// Validate name.
+	if strings.HasPrefix(req.Name, "-") || strings.Contains(req.Name, "/") {
+		return response.BadRequest(fmt.Errorf("Invalid certificate name %q, name must not start with a hyphen or contain forward slashes", req.Name))
 	}
 
 	localHTTPSAddress := s.LocalConfig.HTTPSAddress()
--- a/test/suites/remote.sh
+++ b/test/suites/remote.sh
@@ -11,6 +11,10 @@ test_remote_url() {
     lxc_remote remote remove test
   done
 
+  # invalid certificate names returns an error
+  ! lxc_remote config trust add --name -foo || false
+  ! lxc_remote config trust add --name fo/o || false
+
   # shellcheck disable=2153
   urls="${LXD_DIR}/unix.socket unix:${LXD_DIR}/unix.socket unix://${LXD_DIR}/unix.socket"
 


lxd/certificates.go
```diff
@@ -439,6 +440,11 @@ func certificatesPost(d *Daemon, r *http.Request) response.Response {
 		return response.NotImplemented(fmt.Errorf("Password authentication is no longer supported, please update your client"))
 	}
 
+	// Validate name.
+	if strings.HasPrefix(req.Name, "--") || strings.HasPrefix(req.Name, "-") {
+		return response.BadRequest(fmt.Errorf("Invalid certificate name"))
+	}
```
markylaing, line 446: Do you think `strings.HasPrefix(req.Name, "--")` might be redundant?

Also has there been any discussion around any additional validation we might want to apply. Commonly for lxd entities we disallow forward slashes.

markylaing (changes requested):

markylaing (changes requested):
The first commit in this PR adds validation for certificate names and the second commit changes that validation. This makes sense in terms of tracking how the change was developed, but keeping track of multiple commits like this can become a headache when backporting fixes to older versions of LXD.

Where possible, try to keep each commit to one single logical change to the codebase. Could you please squash these commits together and update the comment to reflect the result (disallow forward slashes and hyphen prefixes)?

Thanks :)

lxd/certificates.go
```diff
@@ -439,6 +440,11 @@ func certificatesPost(d *Daemon, r *http.Request) response.Response {
 		return response.NotImplemented(fmt.Errorf("Password authentication is no longer supported, please update your client"))
 	}
 
+	// Validate name.
+	if strings.HasPrefix(req.Name, "-") || strings.Contains(req.Name, "/") {
+		return response.BadRequest(fmt.Errorf("Invalid certificate name"))
```
markylaing, line 445: Could you please add some more information to the error message so that the user will know what they need to change?

E.g. `fmt.Errorf("Invalid certificate name %q, name must not start with a hyphen or contain forward slashes")`

markylaing (changes requested):
One final nit and I think it's good to go :)

MusicDin (dismissed):
Thanks!

tomponline (commented):
What happens for users who've already created Identities with `-` or `/` in them?

kadinsayani:
> What happens for users who've already created Identities with `-` or `/` in them?

Since the input validation happens on the LXD server side, specifically in ```certificatesPost()```, users who have already created certificates with names with the prefix `-` or containing `/` shouldn't run into any issues. I only see one potentially problematic request.

https://github.com/canonical/lxd/blob/bb61ded59b315bdf70ffbac9bdfef8120023eeb2/lxd/cluster/connect.go#L139-L156

```SetupTrust()``` could cause the post request to throw the invalid name error, but I think this is unlikely since it uses the server's TLS certificate.

**Review by tomponline — CHANGES_REQUESTED** [review]
Thanks.

Please can we add some tests for this.

@markylaing maybe able to advise a good place in the test suite for this.

File: lxd/certificates.go [hunk 3]
```diff
@@ -439,6 +440,11 @@ func certificatesPost(d *Daemon, r *http.Request) response.Response {
 		return response.NotImplemented(fmt.Errorf("Password authentication is no longer supported, please update your client"))
 	}
 
+	// Validate name.
+	if strings.HasPrefix(req.Name, "-") || strings.Contains(req.Name, "/") {
+		return response.BadRequest(fmt.Errorf("Invalid certificate name %q, name must not start with a hyphen or contain forward slashes", req.Name))
```
tomponline, line 445: just "slash" rather than "slashes"

markylaing:
> Thanks.
> 
> Please can we add some tests for this.
> 
> @markylaing maybe able to advise a good place in the test suite for this.

I think the best place is probably in `test/suites/remote.sh` in `remote_url`. Alternatively, a good place would be in the `basic_usage` test (`test/suites/basic.sh`) before the call to `ensure_has_localhost_remote`.

markylaing (approved):
LGTM thanks!

tomponline (approved):

tomponline:
Thanks!

PR #13907, opened 2024-08-12
lxc: Add support for creating project from yaml (from Incus)

gabrielmougard:
Cherry-picked from https://github.com/lxc/incus/pull/765

### DIFF
--- a/lxc/project.go
+++ b/lxc/project.go
@@ -92,6 +92,11 @@ func (c *cmdProjectCreate) command() *cobra.Command {
 	cmd.Short = i18n.G("Create projects")
 	cmd.Long = cli.FormatSection(i18n.G("Description"), i18n.G(
 		`Create projects`))
+	cmd.Example = cli.FormatSection("", i18n.G(`lxc project create p1
+
+lxc project create p1 < config.yaml
+    Create a project with configuration from config.yaml`))
+
 	cmd.Flags().StringArrayVarP(&c.flagConfig, "config", "c", nil, i18n.G("Config key/value to apply to the new project")+"``")
 
 	cmd.RunE = c.run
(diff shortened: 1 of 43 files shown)


test/suites/projects.sh
```diff
@@ -1060,3 +1060,31 @@ test_projects_usage() {
   lxc image delete testimage --project test-usage
   lxc project delete test-usage
 }
+
+test_projects_yaml() {
+  lxc project create test-project-yaml < <(cat <<EOF
```
MusicDin, line 1065: Is the process substitution intentional here, or could we remove it to for readability? 
Something like that:
```sh
lxc project create test-project-yaml <<EOF
config:
  limits.cpu: 2
  limits.memory: 2048MiB
description: Test project using YAML
EOF
```

lxc/project.go
```diff
@@ -121,15 +141,18 @@ func (c *cmdProjectCreate) run(cmd *cobra.Command, args []string) error {
 	// Create the project
 	project := api.ProjectsPost{}
 	project.Name = resource.name
+	project.ProjectPut = stdinData
+
+	if project.Config == nil {
```
MusicDin, line 146: Is there any case where `project.Config` is not nil?

MusicDin (commented):
Thanks! Just a few questions, otherwise looks good to me.

lxc/project.go
gabrielmougard: Yes, when using YAML as an input for example (above, in the `if !termios.IsTerminal(getStdinFd())` block). This will populate `stdinData` which will be added to project.ProjectPut (therefore, indirectly to `project.Config`)... This is when you do `lxc project create --config k1=v1 --config k2=v2 ...` (so not using YAML) that we enter this `if project.Config == nil` block

MusicDin: I see, thanks

test/suites/projects.sh
gabrielmougard: fixed

MusicDin (dismissed):

tomponline (changes requested):
please rebase

tomponline (approved):

tomponline:
looks like it needs rebasing again

MusicDin (approved):

tomponline (approved):

PR #13924, opened 2024-08-12
Storage: Add support for storage buckets backup

boltmark:
This PR adds support for storage buckets backup.

Includes cherry picks from https://github.com/lxc/incus/pull/365.


boltmark:
@tomponline this is ready for review, though I'd appreciate some insight into how to address the CodeQL errors.

I see many CodeQL errors regarding user data in path expressions, but I'm not sure this is relevant here as we restrict resource names and are building path expressions in the same way we currently do in many other places, i.e. [here](https://github.com/canonical/lxd/blob/main/lxd/backup/backup_volume.go#L73). Should these errors be dismissed?

boltmark:
@tomponline here are examples of CodeQL errors even though validation occurs: 
https://github.com/canonical/lxd/pull/13924/files#diff-e63bb483ef763696d5679bfb31d544cdab0d26e6099e6152ede6570210493937

Validation of the variable it complains about happens here:
https://github.com/canonical/lxd/blob/c2a206b03c29312783a478d4354c4dfc9664421c/lxd/storage_buckets.go#L1221-L1224

boltmark:
@tomponline the CodeQL errors regarding user data in path expressions are now resolved. For storage bucket backups, we ensure that backup names do not contain `..`, `/`, or `\`. I've enforced this for other backups as well to ensure we are consistent with backup naming restrictions.

One CodeQL error remains, as incus sets `InsecureSkipVerify` to true when setting up the minio client. Does it make sense to use `InsecureSkipVerify` for the minio client or is this something we should look into?

tomponline:
> One CodeQL error remains, as incus sets `InsecureSkipVerify` to true when setting up the minio client. Does it make sense to use `InsecureSkipVerify` for the minio client or is this something we should look into?

To answer this I would need to understand more about what LXD is doing with the minio client; why does it need to talk to the minio server?

In general because this PR is introducing a new feature it would benefit from its description being filled in to explain what it does and how it works. Then the the reviewer can look at the code with some context as to how the author/importer believes it should work.

tomponline (changes requested):
Please expand the PR description to include info on what this PR does, how it works, new file formats generated etc. Thanks

tomponline:
please can this be rebased

boltmark:
> please can this be rebased

Done

tomponline:
please can you rebase and update i18n. ta

tomponline:
please can you rebase, it looks like another api extension beat this one to the post.

test/suites/storage_buckets.sh
```diff
@@ -186,3 +186,113 @@ EOF
...
 }
+
+test_storage_bucket_export() {
+  # shellcheck disable=2039,3043
+  local lxd_backend
+
+  lxd_backend=$(storage_backend "$LXD_DIR")
+
+  # Skip the test if we are not using a Ceph or Dir backend, because the test requires a storage pool
+  # larger than 1 GiB due to the Minio requirements: https://github.com/minio/minio/issues/6795
```
tomponline, line 197: why do we require 1GiB?

boltmark: Not sure why incus has that restriction, but tests pass for other backends so I've removed that check.

tomponline:
please can you rebase

lxd/instance_backup.go
tomponline, line 1: Would you mind landing this change and the associated cherry-pick for the function in a separate PR to isolate the new functionality only in this PR. Thanks

lxd/db/backups.go
```diff
@@ -541,3 +552,185 @@ func (c *ClusterTx) RenameVolumeBackup(ctx context.Context, oldName, newName str
...
+	}
+
+	return result, nil
+}
+
+// CreateStoragePoolBucketBackup creates a new storage bucket backup.
+func (c *ClusterTx) CreateStoragePoolBucketBackup(ctx context.Context, args StoragePoolBucketBackup) error {
+	_, err := c.getStoragePoolBucketBackupID(ctx, args.Name)
+	if err == nil {
+		return ErrAlreadyDefined
```
tomponline, line 584: we should use api.StatusErrorf() here like we do in `CreateStoragePoolBucketKey` then we wont need the new ErrAlreadyDefined var. 

```go
 api.StatusErrorf(http.StatusConflict, "A bucket key using that access key already exists on this server")
```

Then we can check it in the call site using:

```
api.StatusErrorCheck(err, http.StatusConflict)
```

lxd/storage/s3/transfer_manager.go
```diff
@@ -0,0 +1,196 @@
...
+		s3URL:     s3URL,
+		state:     state,
+		accessKey: accessKey,
+		secretKey: secretKey,
+	}
+}
+
+// DownloadAllFiles downloads all files from a bucket and writes them to a tar writer.
+func (t TransferManager) DownloadAllFiles(bucketName string, tarWriter *instancewriter.InstanceTarWriter) error {
+	logger.Debugf("Downloading all files from bucket %s", bucketName)
```
tomponline, line 43: we dont use `logger.*f` functions, but we can probably just remove these right?

tomponline:
please can you rebase

boltmark:
@tomponline I've isolated the storage bucket backups functionality as discussed, though I'm seeing the pipeline fail due to large lxc binary size:
`FAIL: lxc binary size exceeds 15MiB`

I'm not entirely sure what to do about this.

tomponline:
@simondeziel can point u in right direction to update this (if the growth is expected, but it sounds like you may have increasex the lxc command unexpectedly via import changes)

test/suites/storage_buckets.sh
```diff
@@ -186,3 +186,107 @@ EOF
 
   delete_object_storage_pool "${poolName}"
 }
+
+test_storage_bucket_export() {
```
simondeziel, line 190: At first glance, that's a rather extensive test script, thanks for that!!

lxd/db/backups.go
boltmark: Thanks for the suggestion -- updated

test/suites/storage_buckets.sh
```diff
@@ -186,3 +186,107 @@ EOF
...
+  # shellcheck disable=2039,3043
```
simondeziel, line 191: Disabling those shouldn't be needed AFAIK.

lxd/storage/s3/transfer_manager.go
boltmark: This cherry pick uses them quite a bit, should we remove all of them?

Just curious, I see `logger.*f` uses from `lxd/shared` all over currently in the repo -- what is different about this?

simondeziel:
> @tomponline I've isolated the storage bucket backups functionality as discussed, though I'm seeing the pipeline fail due to large lxc binary size: `FAIL: lxc binary size exceeds 15MiB`

`main` has `lxc` at `15717840` bytes ATM. You PR barely brings it above this with:

```
+ '[' 15746576 -gt 15728640 ']'
FAIL: lxc binary size exceeds 15MiB
```

15746576 (your PR) - 15717840 (main) = 28736 => ~28KiB.

> I'm not entirely sure what to do about this.

It seems like `lxc` has slowly been growing and your PR just happen to make it above the 15MiB limit. We can probably bump it to 16MiB for now and think about putting it to diet later on ;)

test/suites/storage_buckets.sh
boltmark: Removed

lxd/storage/s3/transfer_manager.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func getTransport() *http.Transport {
+	return &http.Transport{
+		MaxIdleConns:       10,
+		IdleConnTimeout:    30 * time.Second,
+		DisableCompression: true,
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
+			MinVersion:         tls.VersionTLS12,
```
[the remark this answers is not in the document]
boltmark: Updated to 1.3

client/lxd_storage_buckets.go
```diff
@@ -233,3 +241,151 @@ func (r *ProtocolLXD) DeleteStoragePoolBucketKey(poolName string, bucketName str
...
+
+	// Handle the data
+	body := response.Body
+	if req.ProgressHandler != nil {
+		body = &ioprogress.ProgressReader{
+			ReadCloser: response.Body,
+			Tracker: &ioprogress.ProgressTracker{
+				Length: response.ContentLength,
+				Handler: func(percent int64, speed int64) {
+					req.ProgressHandler(ioprogress.ProgressData{Text: fmt.Sprintf("%d%% (%s/s)", percent, units.GetByteSizeString(speed, 2))})
```
simondeziel, line 323:
```suggestion
					req.ProgressHandler(ioprogress.ProgressData{Text: strconv.FormatInt(percent, 10) + "% (" + units.GetByteSizeString(speed, 2) + "/s)"})
```

lxc/storage_bucket.go
```diff
@@ -1214,3 +1228,254 @@ func (c *cmdStorageBucketKeyShow) run(cmd *cobra.Command, args []string) error {
...
+		Quiet:  c.global.flagQuiet,
+	}
+
+	createArgs := lxd.StoragePoolBucketBackupArgs{
+		BackupFile: &ioprogress.ProgressReader{
+			ReadCloser: file,
+			Tracker: &ioprogress.ProgressTracker{
+				Length: fstat.Size(),
+				Handler: func(percent int64, speed int64) {
+					progress.UpdateProgress(ioprogress.ProgressData{Text: fmt.Sprintf("%d%% (%s/s)", percent, units.GetByteSizeString(speed, 2))})
```
simondeziel, line 1460:
```suggestion
					progress.UpdateProgress(ioprogress.ProgressData{Text: strconv.FormatInt(percent, 10) + "% (" + units.GetByteSizeString(speed, 2) + "/s)"})
```

test/main.sh
```diff
@@ -397,6 +397,7 @@ if [ "${1:-"all"}" != "cluster" ]; then
     run_test test_storage_driver_zfs "zfs storage driver"
     run_test test_storage_driver_pure "pure storage driver"
     run_test test_storage_buckets "storage buckets"
+    run_test test_storage_bucket_export "storage buckets export and import"
```
simondeziel, line 400: `storage_buckets` is part of the `LXD_REQUIRED_TESTS` for some reason @tomponline probably knows.
Maybe `storage_bucket_export` should be part of that list as well?

It shouldn't change much because the early returns in the `test_storage_bucket_export()` are for conditions that should all be met with GitHub CI but still, it might be safer to hard fail if the test is skipped.

test/suites/storage_buckets.sh
```diff
@@ -186,3 +186,106 @@ EOF
...
+      mkdir "${TEST_DIR}/${bucketPrefix}/s3"
+      # shellcheck disable=SC2034
+      lxc storage create s3 dir source="${TEST_DIR}/${bucketPrefix}/s3"
+      poolName="s3"
+    fi
+
+    buckets_addr="127.0.0.1:$(local_tcp_port)"
+    lxc config set core.storage_buckets_address "${buckets_addr}"
+    s3Endpoint="https://${buckets_addr}"
+  fi
```
simondeziel, line 235: This looks pretty similar to `create_object_storage_pool()`. Could it be reused/adapted or maybe a new helper function created?

test/suites/storage_buckets.sh
```diff
@@ -186,3 +186,106 @@ EOF
...
+    lxc storage delete "${poolName}"
+  fi
+
+  if [ "$lxd_backend" = "dir" ]; then
+    umount "${TEST_DIR}/${bucketPrefix}"
+    rmdir "${TEST_DIR}/${bucketPrefix}"
+
+    # shellcheck disable=SC2154
+    deconfigure_loop_device "${loop_file_1}" "${loop_device_1}"
+  fi
```
simondeziel, line 287: Looks very similar to `delete_object_storage_pool()`

lxd/storage_buckets_backup.go
```diff
@@ -0,0 +1,810 @@
...
+			if err != nil || count != 1 {
+				continue
+			}
+
+			if num >= m {
+				m = num + 1
+			}
+		}
+
+		req.Name = fmt.Sprintf("backup%d", m)
```
simondeziel, line 357:
```suggestion
		req.Name = fmt.Sprint("backup", m)
```

test/suites/storage_buckets.sh
boltmark: The difference I see is that the `storage_bucket_export` test script relies on vars which are set during the code block. Not sure how I could factor that part out. Any pointers?

test/suites/storage_buckets.sh
boltmark: Thanks! Using `delete_object_storage_pool()` now

test/suites/storage_buckets.sh
simondeziel: Are you referring to `s3Endpoint`? If so, a similar approach to what's done in `test_storage_buckets()` might do.

boltmark: Got it, updated as suggested -- thanks for the pointer! :)

lxc/storage_bucket.go
```diff
@@ -1214,3 +1229,254 @@ func (c *cmdStorageBucketKeyShow) run(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return err
+	}
+
+	defer func() { _ = target.Close() }()
+
+	// Prepare the download request
+	progress = cli.ProgressRenderer{
+		Format: i18n.G("Exporting backup of storage bucket %s"),
+		Quiet:  c.global.flagForceLocal,
```
simondeziel, line 1364: I'm curious as to why force local would mean quiet? Could it be a typo for `c.global.flagQuiet`?

test/suites/storage_buckets.sh
```diff
@@ -186,3 +186,78 @@ EOF
...
+  # Extract storage backup tarball.
+  mkdir "${LXD_DIR}/storage-bucket-export"
+  tar -xzf "${LXD_DIR}/testbucket.tar.gz" -C "${LXD_DIR}/storage-bucket-export"
+
+  # Check tarball content.
+  [ -f "${LXD_DIR}/storage-bucket-export/backup/index.yaml" ]
+  [ -f "${LXD_DIR}/storage-bucket-export/backup/bucket/${lxdTestFile}" ]
+  [ "$(cat "${LXD_DIR}/storage-bucket-export/backup/bucket/${lxdTestFile}")" = "hello world" ]
+
+  # Delete bucket and import exported bucket
```
simondeziel, line 243: Is the bucket size preserved through an export/import? If it is, it feels like something worth testing.

test/suites/storage_buckets.sh
```diff
@@ -186,3 +186,78 @@ EOF
...
+  lxc storage bucket delete "${poolName}" "${bucketPrefix}.foo"
+  lxc storage bucket import "${poolName}" "${LXD_DIR}/testbucket.tar.gz" "${bucketPrefix}.bar"
+  rm "${LXD_DIR}/testbucket.tar.gz"
+
+  # Test listing bucket files via S3.
```
simondeziel, line 248: Could you also compare the checksums (`s3cmdrun ... info` will give you the `md5sum`, IIRC) of the initial and imported buckets?

simondeziel (commented):
LGTM with only minor tweaks to the test script.

test/suites/storage_buckets.sh
```diff
@@ -186,3 +186,78 @@ EOF
...
+    s3Endpoint="${LXD_CEPH_CEPHOBJECT_RADOSGW}"
+  else
+    s3Endpoint="https://$(lxc config get core.storage_buckets_address)"
+  fi
+
+  # Create test bucket
+  initCreds=$(lxc storage bucket create "${poolName}" "${bucketPrefix}.foo" user.foo=comment)
+  initAccessKey=$(echo "${initCreds}" | awk '{ if ($2 == "access" && $3 == "key:") {print $4}}')
+  initSecretKey=$(echo "${initCreds}" | awk '{ if ($2 == "secret" && $3 == "key:") {print $4}}')
+  s3cmdrun "${lxd_backend}" "${initAccessKey}" "${initSecretKey}" ls | grep -F "${bucketPrefix}.foo"
```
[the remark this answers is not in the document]
boltmark: Done

lxc/storage_bucket.go
boltmark: I'd imagine it was supposed to be `c.global.flagQuiet` as you suggested. Fixed!

test/suites/storage_buckets.sh
boltmark: Done, let me know if it looks ok!

test/suites/storage_buckets.sh
boltmark: Done, let me know if it's what you're looking for

lxd/storage/s3/transfer_manager.go
```diff
@@ -0,0 +1,194 @@
...
+	return t.s3URL.Scheme == "https"
+}
+
+func getTransport() *http.Transport {
+	return &http.Transport{
+		MaxIdleConns:       10,
+		IdleConnTimeout:    30 * time.Second,
+		DisableCompression: true,
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
```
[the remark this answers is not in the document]
simondeziel: @boltmark Could you please make that `InsecureSkipVerify` configurable and defaulting to `false`? Thanks!

lxd/storage/s3/transfer_manager.go
tomponline: we're moving away from unstructured logging, but admittedly there are still places where they are used, but we should avoid adding more.

lxd/storage/backend_lxd.go
```diff
@@ -7897,3 +7923,194 @@ func (b *lxdBackend) getParentVolumeUUID(vol drivers.Volume, projectName string)
...
+	if err != nil {
+		return err
+	}
+
+	backupKey, err := b.getFirstReadStorageBucketPoolKey(bucket.ID)
+	if err != nil {
+		return err
+	}
+
+	bucketURL := b.GetBucketURL(bucket.Name)
```
tomponline, line 7958: Hrm, ive not reviewed this yet in detail, but @simondeziel comment about InsecureSkipVerify caught me eye, but not because of the security implications (clearly they are undesirable) but more that it indicates that we have the LXD process apparently (in the case of minio based buckets) looping back to itself (@boltmark this anti-pattern will be familiar to you from your project force deletion PR too) in order to download all the files in an s3 bucket.

Now, I'm guessing the reason its been done like this is in order to support Ceph RADOS buckets (https://documentation.ubuntu.com/lxd/en/latest/reference/storage_cephobject/) where the `b.GetBucketURL` will be an external service (not LXD).

However this then got me thinking, why haven't we come across this issue before, and the answer appears to be, we have, and it was solved by:

https://github.com/canonical/lxd/blob/b784423b1011d6a541fd7f7811735907dcc4d1a2/lxd/storage/drivers/driver_cephobject_buckets.go#L30-L75

and for minio backends (which uses http rather than https as LXD's minio is only bound to local-loopback):

https://github.com/canonical/lxd/blob/b784423b1011d6a541fd7f7811735907dcc4d1a2/lxd/storage/s3/miniod/miniod.go#L75-L85

So we need to check if using these existing client functions allow LXD to connect directly to the s3 service (rather than looping through LXD in the case of minio) and avoid the skipping of TLS checks when connecting to an external ceph rados gw service when TLS is enabled?

Also, suggest looking for follow up commits in Incus to see if this has already been addressed.

Also, if simpler, we could add to the storage drivers interface a `GetS3Client()` function which each storage driver could supply, to make the logic simpler in backendLXD.

PR #13926, opened 2024-08-13
Openfga dependency update

markylaing:
We don't need to care about the options that are passed into thedatabase methods. These are related to OpenFGA's concept of "consistency preference" (see https://openfga.dev/docs/interacting/consistency#when-to-use-higher-consistency) which is not applicable for our implementation.

Closes #13913


hamistao:
The changes look good to me, but if I understood [this](https://github.com/advisories/GHSA-3f6g-m4hr-59h8) correctly we should not merge this PR yet.

markylaing:
> The changes look good to me, but if I understood [this](https://github.com/advisories/GHSA-3f6g-m4hr-59h8) correctly we should not merge this PR yet.

Good catch! Looks like 1.5.9 is ok to use.

markylaing:
I've updated the PR to use 1.5.9

hamistao (approved):
Thanks for this!

PR #13928, opened 2024-08-13
VM: Improve QEMU passthrough device ID generation

### LINKED ISSUE #13597: Potential duplicate QEMU device ID when using long device names
# Required information

 * Distribution: Ubuntu
 * Distribution version: 22.04
 * snap: latest/stable
 * LXC version: 5.21.1 LTS
 * LXD version: 5.21.1 LTS

# Issue description

This is related to #13596.
If a QEMU device name is too long, QEMU will trim it down to 64 characters. For LXD, this means that two devices named "{long_name}1" and "{long_name}2"  may end up with the same QEMU device ID `dev-lxd_{trimmed_long_name}`, which will result in the VM failing to start.

# Steps to reproduce
 1. `LONG_DEVICE_NAME="device-with-very-long-name-for-long-qemu-property-handling-test"`
 2. `mkdir -p $HOME/device-test && mkdir -p $HOME/device-test && lxc init ubuntu:n vm --vm`
 4. `lxc config device add v2 "${LONG_DEVICE_NAME}1" pci address=01:00.0 # Requires having an available PCI device.`
 5. `lxc config device add v2 "${LONG_DEVICE_NAME}2" pci address=02:00.0`
 6. `lxc start vm`

hamistao:
Although this changes qemu device IDs for physical PCI and GPU under some specific circumstances, it is safe to merge since the device ID is only a reference inside QEMU and does not affect how the device is shown inside the instance (`lspci`  information is the same regardless of these changes). Applying this changes was tested and it does not affect previously created devices, they will may have a different device ID iin QEMU the next reboot.

Since `lxd-ci` does not include tests with these types of devices, no tests related to these changes will be added.

Fixes #13597

### DIFF
--- a/lxd/instance/drivers/driver_qemu_config_test.go
+++ b/lxd/instance/drivers/driver_qemu_config_test.go
@@ -936,7 +936,7 @@ func TestQemuConfigTemplates(t *testing.T) {
 				pciSlotName: "host-slot",
 			},
 			`# PCI card ("physical-pci-name" device)
-			[device "dev-lxd_physical-pci-name"]
+			[device "dev-lxd_physical--pci--name"]
 			driver = "vfio-pci"
 			bus = "qemu_pcie1"
 			addr = "00.0"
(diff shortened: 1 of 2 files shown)


hamistao:
@mionaalex This was my mistake, no changes to the docs were made here.

tomponline (approved):
Thanks!

PR #13930, opened 2024-08-14
lxd/storage: Mount pools before attempting to list volumes

### LINKED ISSUE #13860: Upgrading from 4.0 to 5.21 fails on patch `storage_prefix_bucket_names_with_project`
# Required information

 * Distribution: Ubuntu
 * Distribution version: 20.04
 * The output of "snap list --all lxd core20 core22 core24 snapd":

```
Name    Version         Rev    Tracking       Publisher   Notes
core20  20240416        2318   latest/stable  canonical✓  base
core22  20240408        1380   latest/stable  canonical✓  base
lxd     4.0.9-a29c6f1   24061  5.21/stable    canonical✓  disabled
lxd     5.21.2-34459c8  29568  5.21/stable    canonical✓  -
snapd   2.63            21759  latest/stable  canonical✓  snapd
```

 * The output of "lxc info" or if that fails:
   * Kernel version: `Linux rbasak-lxd 5.4.0-189-generic #209-Ubuntu SMP Fri Jun 7 14:05:13 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux`
   * LXC version: `5.21.2 LTS`
   * LXD version: `5.21.2 LTS`
   * Storage backend in use: dir

# Issue description

Following on from #13806 I tried updating the snap to `5.21/stable`. This caused lxc to fail entirely.

# Steps to reproduce

 1. Launch and log in to a Focal VM over ssh. I used: `uvt-kvm create --memory=1024 rbasak-lxd release=focal arch=amd64` with this image: `release=focal arch=amd64 label=release (20240710)`, then `uvt-kvm wait rbasak-lxd`, then `uvt-kvm ssh rbasak-lxd`.
 2. `sudo lxd init`, specifying all defaults except that I used the "dir" storage type.
 3. `lxc remote rm images`
 4. `lxc remote add images https://images.lxd.canonical.com --protocol=simplestreams`
 5. `lxc launch images:debian/sid/amd64 foo`
 6. `sudo snap refresh --channel=5.21/stable lxd`
 7. `lxc list`

Expected results: listing of the one container I created previously.

Actual results:

`Error: LXD unix socket not accessible: Get "http://unix.socket/1.0": EOF`

Logging out of the VM and back in again, and trying `lxc list` again, I get:

`Error: LXD unix socket "/var/snap/lxd/common/lxd/unix.socket" not accessible: Get "http://unix.socket/1.0": dial unix /var/snap/lxd/common/lxd/unix.socket: connect: connection refused`

Rebooting the VM doesn't help. After that, I get:

`Error: LXD unix socket not accessible: Get "http://unix.socket/1.0": EOF`

MggMuggins:
dir and btrfs pools use the generic_vfs, which assumes that the pool has been mounted when listing volumes.

Mount implicitly creates any missing dirs from [this list](https://github.com/canonical/lxd/blob/main/lxd/storage/drivers/volume.go#L80), which is what we ran into in #13860.

Tested locally on upgrade from `4.0/stable` -> `latest/edge` with my binary sideloaded; not sure if/where tests are appropriate.

Fixes #13860

LXD-1393

### DIFF
--- a/lxd/storage/backend_lxd_patches.go
+++ b/lxd/storage/backend_lxd_patches.go
@@ -14,6 +14,7 @@ import (
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/logger"
+	"github.com/canonical/lxd/shared/revert"
 )
 
 var lxdEarlyPatches = map[string]func(b *lxdBackend) error{
@@ -190,6 +191,26 @@ func patchBucketNames(b *lxdBackend) error {
 		return err
 	}
 
+	if len(buckets) == 0 {
+		return nil
+	}
+
+	reverter := revert.New()
+	defer reverter.Fail()
+
+	// ListVolumes assumes that the pool is mounted for dir & btrfs drivers
+	_, err = b.Mount()
+	if err != nil {
+		return err
+	}
+
+	reverter.Add(func() {
+		_, err := b.Unmount()
+		if err != nil {
+			logger.Error("Failed to unmount pool", logger.Ctx{"pool": b.name, "err": err})
+		}
+	})
+
 	// Get list of volumes.
 	volumes, err := b.driver.ListVolumes()
 	if err != nil {
@@ -217,5 +238,6 @@ func patchBucketNames(b *lxdBackend) error {
 		}
 	}
 
+	reverter.Success()
 	return nil
 }
--- a/lxd/storage/drivers/driver_zfs.go
+++ b/lxd/storage/drivers/driver_zfs.go
@@ -717,6 +717,21 @@ func (d *zfs) patchDropBlockVolumeFilesystemExtension() error {
 		poolName = d.name
 	}
 
+	reverter := revert.New()
+	defer reverter.Fail()
+
+	_, err := d.Mount()
+	if err != nil {
+		return fmt.Errorf("Failed to mount pool %s: %w", poolName, err)
+	}
+
+	reverter.Add(func() {
+		_, err := d.Unmount()
+		if err != nil {
+			logger.Error("Failed to unmount pool", logger.Ctx{"pool": poolName, "err": err})
+		}
+	})
+
 	out, err := shared.RunCommand("zfs", "list", "-H", "-r", "-o", "name", "-t", "volume", fmt.Sprintf("%s/images", poolName))
 	if err != nil {
 		return fmt.Errorf("Failed listing images: %w", err)
@@ -743,6 +758,7 @@ func (d *zfs) patchDropBlockVolumeFilesystemExtension() error {
 		}
 	}
 
+	reverter.Success()
 	return nil
 }
 


lxd/storage/backend_lxd_patches.go
```diff
@@ -190,6 +190,16 @@ func patchBucketNames(b *lxdBackend) error {
 		return err
 	}
 
+	if len(buckets) == 0 {
+		return nil
+	}
+
+	// ListVolumes assumes that the pool is mounted for dir & btrfs drivers
+	_, err = b.Mount()
```
MusicDin, line 198: Should we unmount the pool afterwards (to ensure clean exit in case of an error)?

Also, this will ensure the directories are created for dir and btrfs, but have you maybe tested with zfs and lvm? Just to ensure there is no side effects?

MusicDin:
Thanks, it is looking good, just few questions in the comment above.

lxd/storage/backend_lxd_patches.go
MggMuggins: :+1: for unmount.

MggMuggins:
I did a few more tests with the `if len(buckets) == 0` commented out so that the mount would run; lvm works fine and so does zfs.

I ran into this with zfs going from 4.0 -> 5.21 when totally stopping lxd and starting it again:
```
sudo snap stop --disable lxd
sudo snap refresh --channel=5.21/stable lxd
sudo snap start --enable lxd
```
```
2024-08-16T16:45:03-05:00 lxd.daemon[203084]: Error: Failed applying patch "storage_zfs_drop_block_volume_filesystem_extension": Failed applying patch to pool "default": Failed listing images: Failed to run: zfs list -H -r -o name -t volume default/images: exit status 1 (cannot open 'default/images': dataset does not exist)
```
Adding a mount/unmount pair there resolves the issue.

lxd/storage/backend_lxd_patches.go
```diff
@@ -190,9 +191,29 @@ func patchBucketNames(b *lxdBackend) error {
...
+		_, err := b.Unmount()
+		if err != nil {
+			logger.Error("Failed to unmount pool", logger.Ctx{"pool": b.name, "err": err})
+		}
+	})
+
 	// Get list of volumes.
 	volumes, err := b.driver.ListVolumes()
 	if err != nil {
+		reverter.Fail()
```
tomponline, line 216: Should use `defer reverter.Fail()` rather than calling `reverter.Fail` on each error scenario.

MggMuggins: Duh, ty.

tomponline:
@MggMuggins so in a strange twist of fate I was working on a DB patch for the long-lived OVN ports change and noticed that the state.ServerName value I needed wasn't populated when the patch was being run, this lead me to find that the patch system wasn't respecting the run stages argument and so I fixed it over in https://github.com/canonical/lxd/pull/13957

As I was reviewing your PR a thought crossed my mind that as `storage_prefix_bucket_names_with_project` patch was defined as supposed to be running in the `patchPostDaemonStorage` start up stage (which is after the storage pools are mounted) that perhaps the fix in https://github.com/canonical/lxd/pull/13957 that would cause patches to be run at the correct stage during LXD start up process, rather than running them all earlier might also fix this issue.

And indeed it does. If you setup from 4.0/stable with `lxd init --auto` and then refresh to `latest/stable` you get the issue, but if you refresh to `latest/edge` you do not. This is because the pools will be mounted, thus creating any missing dirs, before the patch is run.

So we don't need this PR I'm afraid.

PR #13931, opened 2024-08-15
lxd/networks: Don't fill network config when joining cluster (from Incus)

hamistao:
Based on https://github.com/lxc/incus/pull/351


tomponline:
Added "from Incus" to pr title

tomponline (approved):

PR #13933, opened 2024-08-15
lxd/firewall/drivers: Make sure to always use locking (from Incus)

hamistao:
Based on https://github.com/lxc/incus/pull/310.

This also fixes a bug where the Iptables rules were not being set correctly on xtables, it was included here because the linter was failing because of it.

### DIFF
--- a/lxd/firewall/drivers/drivers_xtables.go
+++ b/lxd/firewall/drivers/drivers_xtables.go
@@ -568,7 +568,7 @@ func (d Xtables) NetworkApplyACLRules(networkName string, rules []ACLRule) error
 
 		// Add rules to chain in table.
 		for _, iptRule := range iptRules {
-			_, err := shared.RunCommand(cmd, append([]string{"-t", "filter", "-A", chain}, iptRule...)...)
+			_, err := shared.RunCommand(cmd, append([]string{"-w", "-t", "filter", "-A", chain}, iptRule...)...)
 			if err != nil {
 				return fmt.Errorf("Failed adding rule to %q chain %q in table %q: %w", cmd, chain, "filter", err)
 			}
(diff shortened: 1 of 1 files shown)


hamistao:
@tomponline Maybe it makes sense to add unit tests for the firewall rules to help catching future bugs like the one that is fixed here. What do you think?

tomponline:
Added "from Incus" to pr title to aid with attribution

tomponline (approved):

tomponline:
> This also fixes a bug where the Iptables rules were not being set correctly on xtables, i

Please can you elaborate on this statement explaining what the issue was.

hamistao:
As you can see on [this line](https://github.com/canonical/lxd/pull/13933/commits/726f91053f0ed62223153354651e71ae092d9224#diff-00736e36bdbb013c7b8f026c259abfd8576e1d3d0174979929c1f93e63b9190bL1165), `generateFilterIptablesRules` was returning an empty slice due to a [recent lint fix commit](https://github.com/hamistao/lxd/commit/47e2ffe90a173b0ab6d113190ec26c8ae5676109#diff-00736e36bdbb013c7b8f026c259abfd8576e1d3d0174979929c1f93e63b9190bL1165), which resulted in the iptables rules generated on this function to never be set.

tomponline:
> As you can see on [this line](https://github.com/canonical/lxd/pull/13933/commits/726f91053f0ed62223153354651e71ae092d9224#diff-00736e36bdbb013c7b8f026c259abfd8576e1d3d0174979929c1f93e63b9190bL1165), `generateFilterIptablesRules` was returning an empty slice due to a [recent lint fix commit](https://github.com/hamistao/lxd/commit/47e2ffe90a173b0ab6d113190ec26c8ae5676109#diff-00736e36bdbb013c7b8f026c259abfd8576e1d3d0174979929c1f93e63b9190bL1165), which resulted in the iptables rules generated on this function to never be set.

Ah thats not good (cc @MggMuggins ) good spot thanks!

tomponline:
> @tomponline Maybe it makes sense to add unit tests for the firewall rules to help catching future bugs like the one that is fixed here. What do you think?

We have iptables checks in lxd-ci so perhaps they need extending to catch this area.

Issue #13542, opened 2024-06-04
Allow volumes to be moved from offline members regardless of the storage driver


roosterfish:
In https://github.com/canonical/lxd/pull/13517 the question was raised why the API runs a specific check only for the Ceph RBD driver but not for remote storage drivers in general (See https://github.com/canonical/lxd/pull/13517#discussion_r1620843911).

The same pattern can be found at another place https://github.com/canonical/lxd/blob/main/lxd/instance_post.go#L987. 

If this isn't specific for Ceph RBD make sure to check for the storage drivers `pool.Driver().Info().Remote` field and validate the changes with both the Ceph RBD and PowerFlex storage drivers.

Make sure to also extend the test suites if necessary.

PR #13934, opened 2024-08-15
Instance: For remote shared storage, make allow cluster self healing and migration from offline source member

heat1024:
Makes allow migration from offline source member and cluster self healing (by set cluster.healing_threshold > 0) if remote shared storage like `powerflex` , not only `ceph` .

For now, Remote shared storage is `ceph` and `powerflex` only.
- https://github.com/canonical/lxd/blob/main/doc/explanation/storage.md#data-storage-location

PR #12386, opened 2023-10-13
Support server side copy/move of custom storage volumes in clusters

monstermunchkin:
Fixes #11445

Issue #11445, opened 2023-03-06
Implement server side move for custom storage volumes


markylaing:
In the instances API it is possible to perform a server side move by sending `POST /1.0/instances/{instanceName}?target={clusterMember}`. Here, `clusterMember` is not the cluster member that contains the instance, but is the cluster member that the instance will be moved to.

See https://github.com/lxc/lxd/blob/e19b4dfd1f571d5466000b9c8601017dd155280a/lxc/move.go#L259 https://github.com/lxc/lxd/blob/e19b4dfd1f571d5466000b9c8601017dd155280a/lxd/instance_post.go#L107 https://github.com/lxc/lxd/blob/e19b4dfd1f571d5466000b9c8601017dd155280a/lxd/instance_post.go#L967

We should have the same behaviour for the storage volumes API, e.g. `POST /1.0/storage-pools/{pool}/volumes/{type}/{volume}?target={clusterMemberToMoveTo}`. This will allow the client to perform a move with one request (current behaviour is to copy the volume, then delete the original).

PR #13939, opened 2024-08-16
Storage: Allow rename of all remote storage vols if cluster member is offline

roosterfish:
This check was hardcoded to work only with the Ceph RBD storage driver as it was the only remote storage driver before we have added support for Dell PowerFlex.

Fixes https://github.com/canonical/lxd/issues/13542
The other part of the issue is already addressed in https://github.com/canonical/lxd/pull/13934.

### DIFF
--- a/lxd/storage_volumes.go
+++ b/lxd/storage_volumes.go
@@ -1485,7 +1485,7 @@ func storagePoolVolumePost(d *Daemon, r *http.Request) response.Response {
 				return resp
 			}
 
-			if details.pool.Driver().Info().Name == "ceph" {
+			if details.pool.Driver().Info().Remote {
 				var dbVolume *db.StorageVolume
 				var volumeNotFound bool
 				var targetIsSet bool


roosterfish:
Mh, I am not so sure about this block and fail to reproduce or reach this part of the code using the Ceph RBD driver.

There are various variations using `--target` and `--destination-target` but for example if you select an offline cluster member using `--target` you anyway get the error `Error: Missing event connection with target cluster member` as LXD cannot connect to it (buy using the `lxc` CLI).

This feature was added in https://github.com/canonical/lxd/pull/12386 based on issue https://github.com/canonical/lxd/issues/11445. But the latest `lxc` CLI is still using the create/delete approach when when moving the volume. So I assume in order to reproduce this you would need to use the API directly.

But detached from all of this, if you move a volume using the following command, shouldn't we just perform the rename instead of actually performing a copy? I guess by using the API additions from https://github.com/canonical/lxd/issues/11445 that would be the case as it just triggers a rename at the end.

lxd/storage_volumes.go
```diff
@@ -2691,7 +2691,7 @@ func addStoragePoolVolumeDetailsToRequestContext(s *state.State, r *http.Request
 	// Convert the volume type name to our internal integer representation.
 	volumeType, err := storagePools.VolumeTypeNameToDBType(volumeTypeName)
 	if err != nil {
-		return api.StatusErrorf(http.StatusBadRequest, err.Error())
+		return api.StatusErrorf(http.StatusBadRequest, "%s", err.Error())
```
tomponline, line 2694: Is this part actually needed? Without the 3rd argument  the whole string is used right?

roosterfish: It looks that starting from `golangci-lint` release `v1.60.1` (released last week) this now produces a `govet` error. See https://github.com/canonical/lxd/actions/runs/10416482319/job/28848800110.

It now also complains about cases where you do `fmt.Sprintf("static string")` which makes sense in my opinion as no format string is used here. See this microcluster PR as an example https://github.com/canonical/microcluster/pull/236.

tomponline: cool thanks!

tomponline:
@roosterfish please can we close this until you get time to work on it?

PR #13947, opened 2024-08-20
Backports (stable-5.21)

tomponline:


tomponline:
I will merge even with the static analysis checks failing as this will get fixed in subsequent backports from main, and shouldn't block the 5.21/edge builds. I need to get this merged in order to get the commit for powerflex to cherry-pick into 5.21/candidate.

MusicDin (approved):

PR #13953, opened 2024-08-20
Ubuntu Pro auto-attachment of guests

markylaing:
This PR implements the LXD side of things for auto-attachment of ubuntu guests:
- Adds a `lxd/ubuntupro` package containing a `Client` which contains an fsmonitor for LXD related configuration.
- Adds the `ubuntupro.Client` to Daemon and State.
- Adds a new config option `ubuntu_pro.guest_attach` with allowed values `on`, `off`, or `available`.
- Adds two `devlxd` endpoints: `GET /1.0/ubuntu-pro` and `POST /1.0/ubuntu-pro/token`. The first returns the value of `ubuntu_pro.guest_attach` (if enabled on the host, otherwise it returns "off"). The second executes a `pro` command on the host to request a guest token, which is returned to the guest.
- Adds the `devlxd` endpoints to the LXD agent.

To use the `fsmonitor` package for this, I needed to refactor it somewhat. This makes the fsmonitor package much more flexible, but we do need to watch for regressions in unix devices, as I'm not sure if our tests cover all the edge cases here.

See https://docs.google.com/document/d/1MJA1LcTcd1OGHweyiahxAJ-UjrjXKzGtslrNqHnc9uc/edit for more details.

Opening as draft while I conduct more testing.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -147,7 +147,7 @@ jobs:
 
           # bin/max (sizes are in MiB)
           SIZES="lxc 15
-                 lxd-agent 12"
+                 lxd-agent 13"
           MIB="$((1024 * 1024))"
 
           while read -r bin max; do
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2470,3 +2470,8 @@ Adds the following internal metrics:
 This introduces per-pool project disk limits, introducing a `limits.disk.pool.NAME`
 configuration option to the project limits. When `limits.disk.pool.POOLNAME: 0`
 for a project, the pool is excluded from `lxc storage list` in that project.
+
+## `ubuntu_pro_guest_attach`
+
+Adds a new {config:option}`instance-miscellaneous:ubuntu_pro.guest_attach` configuration option for instances.
+When set to `on`, if the host has guest attachment enabled, the guest can request a guest token for Ubuntu Pro via `devlxd`.
--- a/lxd/devlxd.go
+++ b/lxd/devlxd.go
@@ -293,6 +293,50 @@ func devlxdDevicesGetHandler(d *Daemon, c instance.Instance, w http.ResponseWrit
 	return response.DevLxdResponse(http.StatusOK, c.ExpandedDevices(), "json", c.Type() == instancetype.VM)
 }
 
+var devlxdUbuntuProGet = devLxdHandler{
+	path:        "/1.0/ubuntu-pro",
+	handlerFunc: devlxdUbuntuProGetHandler,
+}
+
+func devlxdUbuntuProGetHandler(d *Daemon, c instance.Instance, w http.ResponseWriter, r *http.Request) response.Response {
+	if shared.IsFalse(c.ExpandedConfig()["security.devlxd"]) {
+		return response.DevLxdErrorResponse(api.NewGenericStatusError(http.StatusForbidden), c.Type() == instancetype.VM)
+	}
+
+	if r.Method != http.MethodGet {
+		return response.DevLxdErrorResponse(api.NewGenericStatusError(http.StatusMethodNotAllowed), c.Type() == instancetype.VM)
+	}
+
+	settings := d.State().UbuntuPro.GuestAttachSettings(c.ExpandedConfig()["ubuntu_pro.guest_attach"])
+
+	// Otherwise, return the value from the instance configuration.
+	return response.DevLxdResponse(http.StatusOK, settings, "json", c.Type() == instancetype.VM)
+}
+
+var devlxdUbuntuProTokenPost = devLxdHandler{
+	path:        "/1.0/ubuntu-pro/token",
+	handlerFunc: devlxdUbuntuProTokenPostHandler,
+}
+
+func devlxdUbuntuProTokenPostHandler(d *Daemon, c instance.Instance, w http.ResponseWriter, r *http.Request) response.Response {
+	if shared.IsFalse(c.ExpandedConfig()["security.devlxd"]) {
+		return response.DevLxdErrorResponse(api.NewGenericStatusError(http.StatusForbidden), c.Type() == instancetype.VM)
+	}
+
+	if r.Method != http.MethodPost {
+		return response.DevLxdErrorResponse(api.NewGenericStatusError(http.StatusMethodNotAllowed), c.Type() == instancetype.VM)
+	}
+
+	// Return http.StatusForbidden if the host does not have guest attachment enabled.
+	tokenJSON, err := d.State().UbuntuPro.GetGuestToken(r.Context(), c.ExpandedConfig()["ubuntu_pro.guest_attach"])
+	if err != nil {
+		return response.DevLxdErrorResponse(fmt.Errorf("Failed to get an Ubuntu Pro guest token: %w", err), c.Type() == instancetype.VM)
+	}
+
+	// Pass it back to the guest.
+	return response.DevLxdResponse(http.StatusOK, tokenJSON, "json", c.Type() == instancetype.VM)
+}
+
 var handlers = []devLxdHandler{
 	{
 		path: "/",
@@ -307,6 +351,8 @@ var handlers = []devLxdHandler{
 	devlxdEventsGet,
 	devlxdImageExport,
 	devlxdDevicesGet,
+	devlxdUbuntuProGet,
+	devlxdUbuntuProTokenPost,
 }
 
 func hoistReq(f func(*Daemon, instance.Instance, http.ResponseWriter, *http.Request) response.Response, d *Daemon) func(http.ResponseWriter, *http.Request) {
--- a/lxd/instance/instancetype/instance.go
+++ b/lxd/instance/instancetype/instance.go
@@ -405,6 +405,20 @@ var InstanceConfigKeysAny = map[string]func(value string) error{
 		return err
 	},
 
+	// lxdmeta:generate(entities=instance; group=miscellaneous; key=ubuntu_pro.guest_attach)
+	// Indicate whether the guest should auto-attach Ubuntu Pro at start up.
+	// The allowed values are `off`, `on`, and `available`.
+	// If set to `off`, it will not be possible for the Ubuntu Pro client in the guest to obtain guest token via `devlxd`.
+	// If set to `available`, attachment via guest token is possible but will not be performed automatically by the Ubuntu Pro client in the guest at startup.
+	// If set to `on`, attachment will be performed automatically by the Ubuntu Pro client in the guest at startup.
+	// To allow guest attachment, the host must be an Ubuntu machine that is Pro attached, and guest attachment must be enabled via the Pro client.
+	// To do this, run `pro config set lxd_guest_attach=on`.
+	// ---
+	// type: string
+	// liveupdate: no
+	// shortdesc: Whether to auto-attach Ubuntu Pro.
+	"ubuntu_pro.guest_attach": validate.Optional(validate.IsOneOf("off", "on", "available")),
+
 	// Volatile keys.
 
 	// lxdmeta:generate(entities=instance; group=volatile; key=volatile.apply_template)
--- a/lxd/ubuntupro/client.go
+++ b/lxd/ubuntupro/client.go
@@ -1,0 +1,247 @@
+package ubuntupro
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/fs"
+	"net/http"
+	"os"
+	"path"
+
+	"github.com/canonical/lxd/lxd/fsmonitor"
+	"github.com/canonical/lxd/lxd/fsmonitor/drivers"
+	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/logger"
+)
+
+const (
+	// guestAttachSettingOff indicates that guest attachment is turned off.
+	// - When the host has this setting turned off, devlxd requests to `GET /1.0/ubuntu-pro` should return "off" and
+	//   `POST /1.0/ubuntu-pro/token` should return a 403 Forbidden (regardless of the guest setting).
+	// - When the guest has this setting turned off (`ubuntu_pro.guest_attach`), devlxd requests to `GET /1.0/ubuntu-pro`
+	//   should return "off" and `POST /1.0/ubuntu-pro/token` should return a 403 Forbidden (regardless of the host setting).
+	guestAttachSettingOff = "off"
+
+	// guestAttachSettingAvailable indicates that guest attachment is available.
+	// - When the host has this setting, devlxd requests to `GET /1.0/ubuntu-pro` should return the setting from the guest
+	//   (`ubuntu_pro.guest_attach) and `POST /1.0/ubuntu-pro/token` should retrieve a guest token via the Ubuntu Pro client.
+	// - When the guest has this setting, the pro client inside the guest will not try to retrieve a guest token at startup
+	//   (though attachment with a guest token can still be performed with `pro auto-attach`.
+	guestAttachSettingAvailable = "available"
+
+	// guestAttachSettingOn indicates that guest attachment is on.
+	// - When the host has this setting, devlxd requests to `GET /1.0/ubuntu-pro` should return the setting from the guest
+	//   (`ubuntu_pro.guest_attach) and `POST /1.0/ubuntu-pro/token` should retrieve a guest token via the Ubuntu Pro client.
+	// - When the guest has this setting, the pro client inside the guest will attempt to retrieve a guest token at startup.
+	guestAttachSettingOn = "on"
+)
+
+// isValid returns an error if the GuestAttachSetting is not one of the pre-defined values.
+func validateGuestAttachSetting(guestAttachSetting string) error {
+	if !shared.ValueInSlice(guestAttachSetting, []string{guestAttachSettingOff, guestAttachSettingAvailable, guestAttachSettingOn}) {
+		return fmt.Errorf("Invalid guest auto-attach setting %q", guestAttachSetting)
+	}
+
+	return nil
+}
+
+// ubuntuAdvantageDirectory is the base directory for Ubuntu Pro related configuration.
+const ubuntuAdvantageDirectory = "/var/lib/ubuntu-advantage"
+
+// Client is our wrapper for Ubuntu Pro configuration and the Ubuntu Pro CLI.
+type Client struct {
+	guestAttachSetting string
+	monitor            fsmonitor.FSMonitor
+	pro                pro
+}
+
+// pro is an internal interface that is used for mocking calls to the pro CLI.
+type pro interface {
+	getGuestToken(ctx context.Context) (*api.UbuntuProGuestTokenResponse, error)
+}
+
+// proCLI calls the actual Ubuntu Pro CLI to implement the interface.
+type proCLI struct{}
+
+// proAPIGetGuestTokenV1 represents the expected format of calls to `pro api u.pro.attach.guest.get_guest_token.v1`.
+// Not all fields are modelled as they are not required for guest attachment functionality.
+type proAPIGetGuestTokenV1 struct {
+	Result string `json:"result"`
+	Data   struct {
+		Attributes api.UbuntuProGuestTokenResponse `json:"attributes"`
+	} `json:"data"`
+	Errors []struct {
+		Title string
+	} `json:"errors"`
+}
+
+// getTokenJSON runs `pro api u.pro.attach.guest.get_guest_token.v1` and returns the result.
+func (proCLI) getGuestToken(ctx context.Context) (*api.UbuntuProGuestTokenResponse, error) {
+	// Run pro guest attach command.
+	response, err := shared.RunCommandContext(ctx, "pro", "api", "u.pro.attach.guest.get_guest_token.v1")
+	if err != nil {
+		return nil, api.StatusErrorf(http.StatusServiceUnavailable, "Ubuntu Pro client command unsuccessful: %w", err)
+	}
+
+	var getGuestTokenResponse proAPIGetGuestTokenV1
+	err = json.Unmarshal([]byte(response), &getGuestTokenResponse)
+	if err != nil {
+		return nil, api.StatusErrorf(http.StatusInternalServerError, "Received unexpected response from Ubuntu Pro contracts server: %w", err)
+	}
+
+	if getGuestTokenResponse.Result != "success" {
+		if len(getGuestTokenResponse.Errors) > 0 && getGuestTokenResponse.Errors[0].Title != "" {
+			return nil, api.StatusErrorf(http.StatusServiceUnavailable, "Ubuntu Pro contracts server returned %q when getting a guest token with error %q", getGuestTokenResponse.Result, getGuestTokenResponse.Errors[0].Title)
+		}
+
+		return nil, api.StatusErrorf(http.StatusServiceUnavailable, "Ubuntu Pro contracts server returned %q when getting a guest token", getGuestTokenResponse.Result)
+	}
+
+	return &getGuestTokenResponse.Data.Attributes, nil
+}
+
+// New returns a new Client that watches /var/lib/ubuntu-advantage for changes to LXD configuration and contains a shim
+// for the actual Ubuntu Pro CLI. If the host is not Ubuntu, it returns a static Client that always returns
+// guestAttachSettingOff.
+func New(osName string, ctx context.Context) *Client {
+	if osName != "Ubuntu" {
+		// If we're not on Ubuntu, return a static Client.
+		return &Client{guestAttachSetting: guestAttachSettingOff}
+	}
+
+	s := &Client{}
+	s.init(ctx, shared.HostPath(ubuntuAdvantageDirectory), proCLI{})
+	return s
+}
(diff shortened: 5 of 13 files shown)


markylaing:
@tomponline RE these lint failures: https://github.com/canonical/lxd/actions/runs/10487443271/job/29047794191#step:14:78

They seem to have appeared due to a new version of `golangci-lint`. It is to do with using non- string format values in a format string. I think `govet` has now learned to pick these up in `api.StatusErrorf`.

This means that we'll need to replace things like 
```go
api.StatusErrorf(http.StatusBadRequest, http.StatusText(http.StatusBadRequest)
```
with
```go
api.StatusErrorf(http.StatusBadRequest, "%s", http.StatusText(http.StatusBadRequest)
```

Which seems a bit redundant. I couldn't see anything in golangci-lint about this change. But this [CodeQL section](https://codeql.github.com/codeql-query-help/cpp/cpp-non-constant-format/#non-constant-format-string) makes sense of it a bit better. I'll make the change.

tomponline:
> Which seems a bit redundant. I couldn't see anything in golangci-lint about this change. But this [CodeQL section](https://codeql.github.com/codeql-query-help/cpp/cpp-non-constant-format/#non-constant-format-string) makes sense of it a bit better. I'll make the change.

Yeah, it makes sense to change these because, in theory, if there was a '%' character in the string variable passed in, it would be considered a placeholder rather than a literal and cause incorrect output

markylaing:
@tomponline I've marked the JIRA card as blocked as we're waiting for pro API and CLI changes. I've done all I can do on this for now. Feel free to review when you have a moment.

doc/api-extensions.md
```diff
@@ -2457,3 +2457,8 @@ This can be used by a virtual machine running LXD to access raw images from the
 
 Adds a new `virtio-blk` value for `io.bus` on `disk` devices which allows
 for the attached disk to be connected to the `virtio-blk` bus.
+
+## `ubuntu_pro_guest_attach`
+
+Adds a new {config:option}`instance-miscellaneous:ubuntu_pro.guest_attach` configuration option for instances.
+When set to `on`, if the host has guest attachment enabled, the guest can request a guest token for Ubuntu pro via `devlxd`.
```
tomponline, line 2464: Ubuntu Pro? rather than pro?

lxd/instance/instancetype/instance.go
```diff
@@ -372,6 +372,14 @@ var InstanceConfigKeysAny = map[string]func(value string) error{
 		return err
 	},
 
+	// lxdmeta:generate(entities=instance; group=miscellaneous; key=ubuntu_pro.guest_attach)
```
tomponline, line 375: can we include the allowed values in the description and their meanings?

lxd/devlxd.go
```diff
@@ -297,6 +298,64 @@ func devlxdDevicesGetHandler(d *Daemon, c instance.Instance, w http.ResponseWrit
...
+	handlerFunc: devlxdUbuntuProGetHandler,
+}
+
+func devlxdUbuntuProGetHandler(d *Daemon, c instance.Instance, w http.ResponseWriter, r *http.Request) response.Response {
+	if shared.IsFalse(c.ExpandedConfig()["security.devlxd"]) {
+		return response.DevLxdErrorResponse(api.StatusErrorf(http.StatusForbidden, "not authorized"), c.Type() == instancetype.VM)
+	}
+
+	if r.Method != http.MethodGet {
+		return response.DevLxdErrorResponse(api.StatusErrorf(http.StatusMethodNotAllowed, "%s", http.StatusText(http.StatusMethodNotAllowed)), c.Type() == instancetype.VM)
```
tomponline, line 312: You can use api.NewGenericStatusError here right?

doc/api-extensions.md
```diff
@@ -2457,3 +2457,8 @@ This can be used by a virtual machine running LXD to access raw images from the
 
 Adds a new `virtio-blk` value for `io.bus` on `disk` devices which allows
 for the attached disk to be connected to the `virtio-blk` bus.
+
+## `ubuntu_pro_guest_attach`
```
tomponline, line 2461: Please can the commit message start with `api: ` as per the documented convention:

https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#commit-structure

doc/api-extensions.md
```diff
@@ -2457,3 +2457,8 @@ This can be used by a virtual machine running LXD to access raw images from the
...
+
+Adds a new {config:option}`instance-miscellaneous:ubuntu_pro.guest_attach` configuration option for instances.
+When set to `on`, if the host has guest attachment enabled, the guest can request a guest token for Ubuntu Pro via `devlxd`.
```
tomponline, line 2464: > When set to on, *and* if the host has guest attachment enabled in the *Ubuntu Pro client*, the guest...

lxd/ubuntu-pro/watcher.go
```diff
@@ -0,0 +1,252 @@
...
+type GuestAttachSetting string
+
+const (
+	// GuestAttachSettingOff indicates that guest attachment is never available. The value of `GET /1.0/ubuntu-pro` over
+	// devlxd should always be "off".
+	GuestAttachSettingOff GuestAttachSetting = "off"
+
+	// GuestAttachSettingAvailable indicates that guest attachment is available and can be opted into via guest config.
+	// The value of `GET /1.0/ubuntu-pro` over devlxd should return the value of the `ubuntu_pro.guest_attach`
+	// configuration of the instance.
```
tomponline, line 26: We should mention here that its not only the instance configuration setting that influences the return value of `GET /1.0/ubuntu-pro` and that its also the status of the pro client on the LXD host too.

lxd/ubuntu-pro/watcher.go
```diff
@@ -0,0 +1,252 @@
...
+	GuestAttachSettingAvailable GuestAttachSetting = "available"
+
+	// GuestAttachSettingOn indicates that guest attachment is on and can be opted into via guest config. The value of
+	// `GET /1.0/ubuntu-pro` over devlxd should return the value of the `ubuntu_pro.guest_attach` configuration of the
+	// instance.
+	//
+	// NOTE: This setting currently behaves in the same way as GuestAttachSettingAvailable, but we will use this setting
```
tomponline, line 33: This isn't correct. The pro client inside the guest will trigger auto attach for `on` but not for `available` (that would require manual `pro auto-attach`).

tomponline: I'm not sure we should include the text about the future work as that is not 100% agreed yet.

lxd/ubuntu-pro/watcher.go
```diff
@@ -0,0 +1,252 @@
...
+	return filepath.Join(ubuntuAdvantageDir, "interfaces")
+}
+
+// lxdConfigFile returns the path to the ubuntu-pro LXD configuration file based on the given "interfaces" directory.
+func lxdConfigFile(interfacesDir string) string {
+	return filepath.Join(interfacesDir, "lxd-config.json")
+}
+
+// Settings contains methods for returning ubuntu-pro related settings.
+type Settings interface {
```
tomponline, line 62: Having an interface here feels a bit unnecessary to me.  There's only one implementation.
And in general interfaces are better defined where they are consumed rather than where the implementations are defined.

https://go.dev/wiki/CodeReviewComments

> Go interfaces generally belong in the package that uses values of the interface type, not the package that implements those values. The implementing package should return concrete (usually pointer or struct) types: that way, new methods can be added to implementations without requiring extensive refactoring.

lxd/ubuntu-pro/watcher.go
```diff
@@ -0,0 +1,252 @@
...
+	guestAttachSetting GuestAttachSetting
+}
+
+// GuestAttach returns the current GuestAttachSetting from the ubuntu-pro configuration.
+func (p *proWatcher) GuestAttach() GuestAttachSetting {
+	return p.guestAttachSetting
+}
+
+// SettingsFile contains ubuntu-pro settings relevant to LXD.
+type SettingsFile struct {
```
tomponline, line 86: `File` here seems unnecessary, this struct represents the logical pro settings, it shouldn't matter how it gets populated (file or otherwise).

lxd/ubuntu-pro/watcher.go
```diff
@@ -0,0 +1,252 @@
...
+	GuestAttach GuestAttachSetting `json:"guest_attach"`
+}
+
+// New returns the proWatcher implementation of Settings using a file watcher. When the given context is cancelled, all file
+// watchers are closed and GuestAttach will return GuestAttachSettingOff.
+func New(ctx context.Context) Settings {
```
tomponline, line 92: Suggest having this return a point to an `ubuntupro.State` struct that has `GuestAttachSetting()` function on it.

tomponline: or just a `pro` struct, which we can add to in the future, and have a function called `GuestAttachState()`.

lxd/ubuntu-pro/watcher.go
```diff
@@ -0,0 +1,252 @@
...
+	lxdConfigFile := lxdConfigFile(interfacesDir)
+
+	for {
+		select {
+		case <-ctx.Done():
+			// If the context is cancelled, we're no longer watching the file. Always return GuestAttachSettingOff.
+			p.guestAttachSetting = GuestAttachSettingOff
+			err := p.watcher.Close()
+			if err != nil {
+				logger.Error("Failed to close ubuntu-pro auto-attach configuration file watcher", logger.Ctx{"err": err})
```
tomponline, line 144: Thats a warning right?

lxd/ubuntu-pro/watcher.go
```diff
@@ -0,0 +1,252 @@
...
+			}
+
+			return
+		case event := <-p.watcher.Events:
+			switch {
+			case event.Name == interfacesDir && event.Has(fsnotify.Create):
+				// Interfaces directory created. Add a watcher to it.
+				err := p.watcher.Add(interfacesDir)
+				if err != nil {
+					logger.Error("Error occurred while adding file watcher for ubuntu-pro interface directory", logger.Ctx{"err": err})
```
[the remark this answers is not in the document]
tomponline: if the pro client isn't installed on the host and the interfacesDir doesn't exist, does this fail?

lxd/ubuntu-pro/watcher.go
```diff
@@ -0,0 +1,252 @@
...
+	err = w.Add(ubuntuAdvantageDir)
+	if err != nil {
+		return nil, nil, fmt.Errorf("Failed adding ubuntu-advantage directory to file watcher: %w", err)
+	}
+
+	// Check if the interfaces directory exists.
+	interfacesDir := interfacesDir(ubuntuAdvantageDir)
+	_, err = os.Stat(interfacesDir)
+	if err != nil && !os.IsNotExist(err) {
+		return nil, nil, fmt.Errorf("Failed to check if ubuntu-pro interface directory exists: %w", err)
```
tomponline, line 224: I suspect we'll need to keep checking if it exists periodically, as the package may be installed which creates the dir after LXD is started

lxd/ubuntu-pro/watcher.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	// Add the interfaces directory to the watch list.
+	err = w.Add(interfacesDir)
+	if err != nil {
+		return nil, nil, fmt.Errorf("Failed to add ubuntu-pro interface directory to file watcher: %w", err)
+	}
+
+	// Check if the LXD config file exists.
+	lxdConfigFile := lxdConfigFile(interfacesDir)
```
tomponline, line 236: Rather than checking here if it exists, can you call readLXDConfigFile and test for os.IsNotExist on the error returned?

lxd/instance/instancetype/instance.go
```diff
@@ -372,6 +372,19 @@ var InstanceConfigKeysAny = map[string]func(value string) error{
...
+	// Indicate whether the guest should auto-attach Ubuntu Pro at start up.
+	// The allowed values are `off`, `on`, and `available`.
+	// If set to `off`, it will not be possible for the Ubuntu Pro client to obtain guest token via devlxd.
+	// If set to `available`, attachment via guest token is possible but will not be performed automatically by the Ubuntu Pro client at startup.
+	// If set to `on`, attachment will be performed automatically by the Ubuntu Pro client at startup.
```
tomponline, line 380: Probably should add here a mention that the Ubuntu Pro client is required to be installed inside the guest.

lxd/devlxd.go
```diff
@@ -297,6 +298,64 @@ func devlxdDevicesGetHandler(d *Daemon, c instance.Instance, w http.ResponseWrit
...
+	if r.Method != http.MethodGet {
+		return response.DevLxdErrorResponse(api.NewGenericStatusError(http.StatusMethodNotAllowed), c.Type() == instancetype.VM)
+	}
+
+	s := d.State()
+	hostAttachSetting := s.UbuntuPro.GuestAttach()
+	guestAttachSetting := c.ExpandedConfig()["ubuntu_pro.guest_attach"]
+
+	// If either the host or the guest have the value set to "off" (or if the attachment is not configured on the instance), then return "off".
+	if hostAttachSetting == pro.GuestAttachSettingOff || guestAttachSetting == string(pro.GuestAttachSettingOff) || guestAttachSetting == "" {
```
tomponline, line 320: As you have access to d.Pro here, can you push this logic down into the ubuntupro package rather than having it in the API handler, then you can add unit tests for it.

lxd/devlxd.go
```diff
@@ -297,6 +298,64 @@ func devlxdDevicesGetHandler(d *Daemon, c instance.Instance, w http.ResponseWrit
...
+
+	// Return http.StatusForbidden if the host does not have guest attachment enabled.
+	s := d.State()
+	hostAttachSetting := s.UbuntuPro.GuestAttach()
+	if hostAttachSetting == pro.GuestAttachSettingOff {
+		return response.DevLxdErrorResponse(api.NewStatusError(http.StatusForbidden, "Guest attachment not allowed"), c.Type() == instancetype.VM)
+	}
+
+	// Run pro guest attach command.
+	stdout, err := shared.RunCommandContext(r.Context(), "pro", "api", "u.pro.attach.guest.get_guest_token.v1")
```
tomponline, line 350: Lets push this into the ubuntupro package please and add a function to s.UbuntuPro to get a token.

lxd/ubuntu-pro/watcher.go
markylaing: The `interfaces` directory is added to the watchlist when it is created. I've bailed out if `/var/lib/ubuntu-advantage` doesn't exist with the assumption being it is not an ubuntu host.

tomponline: I think we need to get clarity around when that dir is created incase its managed by the pro client tool. We may need a different indicator for Ubuntu guest.

markylaing: I'm beginning to think this file watcher approach is incorrect. The pro client has a `config` command (used for setting `lxd_guest_attach`) with a `show` subcommand:
```
root@v1:~# pro config show lxd_guest_attach
lxd_guest_attach off
```

How about the `devlxd` handler checks the instance's `ubuntu_pro.guest_attach` setting. If it is `available` or `on`, we call `pro config show lxd_guest_attach` and parse the output to get the current value. If it is `off` or there is an error we return `off`, otherwise we return the instance setting.

This leaves the pro client free to change it's internal representation of the configuration.

@orndorffgrant is the `pro config show` command considered stable for this purpose?

tomponline: Yes potentially that could be a better way instead. But I don't think we should be calling out to the pro client on every /dev/lxd request for guest attach status as it would be  away for the guests to DoS the host (execing external processes are relatively expensive), additionally if you are running lots of instances and you start/restart them they are all going to query the pro client around the same time.

So we would need some form caching in that scenario, so that we only get the current value at most once per X time period. And this could introduce unexpected behaviour when you enable it on the host and then immediately go into the guest to auto attach and it doesn't work straight away.

Hence the event driven approach was chosen using a config file.

markylaing: Makes sense, thanks.

markylaing:
This has now been tested manually against the Pro contracts staging environment with the test [Pro client PPA](https://launchpad.net/~orndorffgrant/+archive/ubuntu/pro-client-lxd-testing).

tomponline:
@markylaing please can you move the fsmonitor changes into a separate PR so we can review and merge first

markylaing:
The refactor of the fsmonitor package https://github.com/canonical/lxd/pull/14110 must be merged first.

lxd/ubuntupro/client.go
```diff
@@ -0,0 +1,261 @@
...
+		Attributes ProAPIGetGuestTokenV1DataAttributes `json:"attributes"`
+	} `json:"data"`
+	Errors []struct {
+		Title string
+	} `json:"errors"`
+}
+
+// ProAPIGetGuestTokenV1DataAttributes contains the expected fields of proAPIGetGuestTokenV1 that must be passed back to
+// the guest for pro attachment to succeed.
+type ProAPIGetGuestTokenV1DataAttributes struct {
```
tomponline, line 89: Having "Pro" here seems unnecessary given the whole package is "ubuntupro"?

tomponline: Also "Attributes" seems unnecessary, given that we know its a struct with fields (i.e attributes).

tomponline: Oh I see its because you've named it after the CLI command plus the sub-field you're extracting the data from.

lxd/ubuntupro/client.go
```diff
@@ -0,0 +1,261 @@
...
+
+	var getGuestTokenResponse proAPIGetGuestTokenV1
+	err = json.Unmarshal([]byte(response), &getGuestTokenResponse)
+	if err != nil {
+		return nil, api.StatusErrorf(http.StatusInternalServerError, "Received unexpected response from Ubuntu Pro contracts server: %w", err)
+	}
+
+	if getGuestTokenResponse.Result != "success" {
+		if len(getGuestTokenResponse.Errors) > 0 && getGuestTokenResponse.Errors[0].Title != "" {
+			return nil, api.StatusErrorf(http.StatusServiceUnavailable, "Server returned %q when getting a guest token with error %q", getGuestTokenResponse.Result, getGuestTokenResponse.Errors[0].Title)
```
tomponline, line 111: We should stick with saying either "Ubuntu Pro contracts server" like above or just "server" here, but not mix them.

lxd/ubuntupro/client.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+
+		return nil, api.StatusErrorf(http.StatusServiceUnavailable, "Server returned %q when getting a guest token", getGuestTokenResponse.Result)
```
tomponline, line 114: We should stick with saying either "Ubuntu Pro contracts server" like above or just "server" here, but not mix them.

lxd/ubuntupro/client.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		logger.Warn("Failed to configure Ubuntu configuration watcher", logger.Ctx{"err": err})
+	}
+}
+
+func (s *Client) watch(ctx context.Context, ubuntuAdvantageDir string) error {
+	// On first call, attempt to read the contents of the config file.
+	configFilePath := path.Join(ubuntuAdvantageDir, "interfaces", "lxd-config.json")
+	err := s.readConfigFile(configFilePath)
+	if err != nil && !os.IsNotExist(err) {
```
tomponline, line 190: https://pkg.go.dev/os#IsNotExist

> This function predates [errors.Is](https://pkg.go.dev/errors#Is). It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrNotExist).

lxd/ubuntupro/client.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		return fmt.Errorf("Failed to configure file monitor: %w", err)
+	}
+
+	s.monitor = monitor
+	return nil
+}
+
+// readConfigFile reads the Ubuntu Pro `lxd-config.json` file and sets appropriate values in the client.
+func (s *Client) readConfigFile(lxdConfigFile string) error {
```
tomponline, line 237: Can we call this `parseConfigFile` or `applyConfigFile` as "read" suggests a not mutating action, which this is not.

lxd/ubuntupro/client.go
```diff
@@ -0,0 +1,261 @@
...
+	s.guestAttachSetting = GuestAttachSettingOff
+
+	f, err := os.Open(lxdConfigFile)
+	if err != nil {
+		return fmt.Errorf("Failed to open Ubuntu Pro configuration file: %w", err)
+	}
+
+	defer f.Close()
+
+	settings := Settings{}
```
tomponline, line 248: just a code style nit, I tend to use `var settings Setting` in these sorts of situations where I'm creating a variable that is an empty vessel to be populated by `Decode`, and the `Settings{}` style when im initiating a variable with zero or more fields that I want set to specific values.

lxd/ubuntupro/client.go
```diff
@@ -0,0 +1,261 @@
...
+// isValid returns an error if the GuestAttachSetting is not one of the pre-defined values.
+func (g GuestAttachSetting) isValid() error {
+	if !shared.ValueInSlice(g, []GuestAttachSetting{GuestAttachSettingOff, GuestAttachSettingAvailable, GuestAttachSettingOn}) {
+		return fmt.Errorf("Invalid guest auto-attach setting %q", g)
+	}
+
+	return nil
+}
+
+// Settings contains Ubuntu Pro settings relevant to LXD.
```
tomponline, line 52: Because /dev/lxd is a public API (to the applications running inside the guest) we need to ensure that the API structs for it are defined in `shared/api` like `api.DevLXDGet`.

So I think this struct should be moved into `shared/api/devlxd.go`.

lxd/ubuntupro/client.go
tomponline, line 89: This needs to be in `shared/api/devlxd.go` too

tomponline: I would suggest we rename it to be shorter and not so tightly coupled to the pro client command, as the command may change in the future, but we wont want the API name to change.

test/godeps/lxd-agent.list
```diff
@@ -328,6 +330,7 @@ gopkg.in/mgo.v2/internal/json
 gopkg.in/tomb.v2
 gopkg.in/yaml.v2
 gopkg.in/yaml.v3
+k8s.io/utils/inotify
```
tomponline, line 333: this is incorrect, there is no reason for lxd-agent to need the inotify dependency, so it we should ensure that the ubuntupro package is split in such a way to achieve this.

We added this godeps check after the openfga functionality was added to catch just this sort of thing where unnecessary deps were being imported into lxd-agent rather than ensuring the separation of server and client imports.

test/godeps/lxd-agent.list
```diff
@@ -33,6 +33,7 @@ github.com/canonical/lxd/lxd/events
 github.com/canonical/lxd/lxd/firewall
 github.com/canonical/lxd/lxd/firewall/drivers
 github.com/canonical/lxd/lxd/fsmonitor
+github.com/canonical/lxd/lxd/fsmonitor/drivers
```
tomponline, line 36: i'd like to understand why lxd-agent already "needs" fsmonitor, and whether we can get rid of both of these dependencies. Does the lxd-agent actually use fsmonitor for anything?

.github/workflows/tests.yml
```diff
@@ -147,7 +147,7 @@ jobs:
 
           # bin/max (sizes are in MiB)
           SIZES="lxc 15
-                 lxd-agent 12"
+                 lxd-agent 13"
```
tomponline, line 150: this should be a warning sign we've accidentally imported something we shouldn't.

tomponline (changes requested):
This is looking good overall, thanks!

My two main requests are:

1. Move client facing API structs used in /dev/lxd responses into `shared/api/devlxd.go` and rename appropriately to align with existing conventions.
2. Review lxd-agent fsmonitor and inotify imports and double check whether lxd-agent actually needs them or whether they are being imported accidentally due to a package not being sufficiently separated into server and client roles.

tomponline:
> Have done an initial investigation. A huge chunk of imports including fsmonitor are coming from the import of the `operations` package, which in turn imports `state`. `state` brings in lots and lots of imports and it explodes from there.

I wonder if we can decouple either lxd-agent and operations or operations and state?

markylaing:
@tomponline this PR now doesn't add any dependencies to the LXD agent, but it is still too large: https://github.com/canonical/lxd/actions/runs/10945290233/job/30389176932?pr=13953#step:16:71

This is from a recent run and we can see there is very little headroom in the check: https://github.com/canonical/lxd/actions/runs/10940393014/job/30372708033#step:16:71

tomponline:
> @tomponline this PR now doesn't add any dependencies to the LXD agent, but it is still too large: https://github.com/canonical/lxd/actions/runs/10945290233/job/30389176932?pr=13953#step:16:71
> 
> This is from a recent run and we can see there is very little headroom in the check: https://github.com/canonical/lxd/actions/runs/10940393014/job/30372708033#step:16:71

Ok lets bump it a little then

tomponline (approved):

PR #13567, opened 2024-06-07
Server: Remove password authentication - From Incus

roosterfish:
This PR is mainly based on https://github.com/lxc/incus/pull/62 and effectively removes the `core.trust_password` config setting together with its internals from the API. The following API changes are covered by this PR:

* `POST /1.0/certificates`: Ignore the payloads `password` field as it is now marked as deprecated. Instead use the `trust_token` field.
* `PUT /1.0/cluster`: Ignore the payloads `cluster_password` field as it is now marked as deprecated. Instead use the `cluster_token` field.

For backwards compatibility the `lxd-migrate` and `lxc` clients still support using the password to be able to communicate with older LXD servers. The clients check for the presence of the `explicit_trust_token` extension in order to decide if the upstream server supports the token field.

The PR also contains an enhancement for the patch system to allow running patches before loading the cluster configuration for the first time. This ensures we can safely remove cluster configuration settings without getting any error in the daemons log.

PR #13957, opened 2024-08-21
Patches: Fix patch system so that it respects the patch stages

### LINKED ISSUE #13860: Upgrading from 4.0 to 5.21 fails on patch `storage_prefix_bucket_names_with_project`

tomponline:
It appears that ever since I introduced the concept of a [patch stage in LXD 3.23](https://github.com/canonical/lxd/pull/7012) the patch stage argument hasn't actually been taking effect.

This hasn't been noticed as each time we've introduced an earlier stage the existing patches were either compatible with the earlier stage or had already been run and so we weren't getting reports of issues.

However with the introduction of the [`patchPreLoadClusterConfig` stage in LXD 6.1](https://github.com/canonical/lxd/pull/13567) it has now become apparent because of a new patch I am working on to remove specific volatile instance config keys from instances on the local member, and I noticed that `state.ServerName` was empty even though it was being set after the `patchPreLoadClusterConfig` stage and my new patch was supposed to be running after that, but in fact, wasn't.

Fixes #13860

### DIFF
--- a/lxd/patches.go
+++ b/lxd/patches.go
@@ -1266,14 +1266,16 @@ func patchStorageRenameCustomISOBlockVolumesV2(name string, d *Daemon) error {
 			return fmt.Errorf("Failed loading pool %q: %w", poolName, err)
 		}
 
+		isRemotePool := p.Driver().Info().Remote
+
 		// Ensure the renaming is done only on the selected patch cluster member for remote storage pools.
-		if p.Driver().Info().Remote && !isSelectedPatchMember {
+		if isRemotePool && !isSelectedPatchMember {
 			continue
 		}
 
 		for _, vol := range volumes {
-			// In a non-clusted environment ServerName will be empty.
-			if s.ServerName != "" && vol.Location != s.ServerName {
+			// Skip volumes on local pools that are on other servers.
+			if !isRemotePool && s.ServerClustered && vol.Location != s.ServerName {
 				continue
 			}
 


roosterfish (dismissed):
Nice catch!

lxd/patches.go
```diff
@@ -941,116 +941,7 @@ func patchZfsSetContentTypeUserProperty(name string, d *Daemon) error {
...
-			err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
-				return tx.UpdateStoragePoolVolume(ctx, vol.Project, vol.Name, volType, pool, vol.Description, config)
-			})
-			if err != nil {
-				return fmt.Errorf("Failed updating volume %q in project %q on pool %q: %w", vol.Name, vol.Project, poolIDNameMap[pool], err)
-			}
-		}
-	}
-
+	// Superseded by patchStorageZfsUnsetInvalidBlockSettingsV2.
```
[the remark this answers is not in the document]
tomponline: It has been applied on systems already so good to keep a log that it existed.

lxd/patches.go
```diff
@@ -1266,14 +1266,16 @@ func patchStorageRenameCustomISOBlockVolumesV2(name string, d *Daemon) error {
...
-		if p.Driver().Info().Remote && !isSelectedPatchMember {
+		if isRemotePool && !isSelectedPatchMember {
 			continue
 		}
 
 		for _, vol := range volumes {
-			// In a non-clusted environment ServerName will be empty.
-			if s.ServerName != "" && vol.Location != s.ServerName {
+			// Skip volumes on local pools that are on other servers.
+			if !isRemotePool && s.ServerClustered && vol.Location != s.ServerName {
```
roosterfish, line 1278: I was now looking at this for a bit.

Having only `!isRemotePool && vol.Location != s.ServerName` would evaluate to `true` every time because if LXD is not clustered `s.ServerName` is empty? So therefore we need the extra check for `s.ServerClustered` (or `s.ServerName != ""` but the other is more elegant :) ).

Maybe for the future it would help to have a `isSelectedVolPatchMember(pool, vol) bool` that can be used for decision making when patching volumes?

roosterfish (approved):

lxd/patches.go
tomponline: > if LXD is not clustered `s.ServerName` is empty?

s.ServerName is "none" in those situations.

tomponline: > Maybe for the future it would help to have a `isSelectedVolPatchMember(pool, vol) bool` that can be used for decision making when patching volumes?

Yes standardising that logic into a helper function would be useful though.

roosterfish: Just thinking what will happen if you actually have a LXD cluster member called "none". I suspect this is allowed?

Maybe worth having another story that checks this too in the other places this pattern is used.

tomponline: I dont think you are allowed to do that, but if you can then it should be blocked.

roosterfish: Looks to be possible:

```
root@m1:~# lxc cluster ls
+------+-----------------------------+------------------+--------------+----------------+-------------+--------+-------------------+
| NAME |             URL             |      ROLES       | ARCHITECTURE | FAILURE DOMAIN | DESCRIPTION | STATE  |      MESSAGE      |
+------+-----------------------------+------------------+--------------+----------------+-------------+--------+-------------------+
| m1   | https://10.165.133.53:8443  | database-leader  | x86_64       | default        |             | ONLINE | Fully operational |
|      |                             | database         |              |                |             |        |                   |
+------+-----------------------------+------------------+--------------+----------------+-------------+--------+-------------------+
| none | https://10.165.133.128:8443 | database-standby | x86_64       | default        |             | ONLINE | Fully operational |
+------+-----------------------------+------------------+--------------+----------------+-------------+--------+-------------------+
```

I'll open up a PR to block it.

PR #13960, opened 2024-08-21
shared: Allow adding multiple names to certificate SAN

masnax:
Requested by the k8s team to use with their Canonical K8s which relies on microcluster. 

This will be used to supply both the `microcluster` internal cluster name as well as the system hostname to the certificate SAN, so either can be used to identify the certificate. 

`CommonName` is the same as the old `SubjectName` in that it is used both in the Subject Common Name field, as well as in the Subjcet Alternative Names field, if defined. If not defined, it will default to the hostname.

`SubjectAlternativeNames` are any alternative names that should be included in the SAN field in addition to `CommonName` and the local IPs.

### DIFF
--- a/shared/cert.go
+++ b/shared/cert.go
@@ -34,8 +34,11 @@ type CertOptions struct {
 	// AddHosts determines whether to populate the Subject Alternative Name DNS Names and IP Addresses fields.
 	AddHosts bool
 
-	// SubjectName will be used in place of the system hostname for the SAN DNS Name and Issuer Common Name.
-	SubjectName string
+	// CommonName will be used in place of the system hostname for the SAN DNS Name and Issuer Common Name.
+	CommonName string
+
+	// SubjectAlternativeNames contains other names to include in the SAN DNS name field in addition to CommonName.
+	SubjectAlternativeNames []string
 }
 
 // KeyPairAndCA returns a CertInfo object with a reference to the key pair and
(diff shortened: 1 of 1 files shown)


shared/cert.go
```diff
@@ -245,23 +248,28 @@ func TestingAltKeyPair() *CertInfo {
...
- * This will include the hostname and ip address.
- * If the `name` argument is non-empty,  it will be used in place of the system hostname.
- */
-func mynames(name string) ([]string, error) {
-	if name == "" {
+// generateSANNames creates a list of names for which the certificate will be valid.
+// - `commonName` will be the first entry if defined, otherwise the hostname will be used.
+// - `additionalNames` will be supplied next, if defined
+// - Finally, the localhost IPs will be added.
+func generateSANNames(commonName string, additionalNames []string) ([]string, error) {
```
tomponline, line 255: Can you make additionalNames variadic please?

masnax: Done. 

I thought about removing `commonName` as an argument and defaulting to the hostname if `additionalNames` is empty, but that doesn't actually reduce any complexity since we'd still have to append the hostname to the args passed into `generateSANNames` anyway.

roosterfish (approved):
LGTM!

tomponline (approved):

aliazlan-t:
@masnax hey, thanks for this PR. I see that you have added an option for ```SubjectAlternativeNames``` but how do I add a custom IP address or hostname that I want to be included in the server cert? From what I understand, this field is just added into the config but there is no code where we are adding values to it.

PR #13963, opened 2024-08-22
Cluster: Enforce member name cannot be `none`

roosterfish:
In https://github.com/canonical/lxd/pull/13957#discussion_r1726718542 we found that currently `none` is a valid LXD cluster member name.

This collides with the internally used concept which identifies if LXD is clustered or not by setting it's internal `ServerName` to `none` or it's actual cluster name.

The PR ensures that a LXD cluster member's name cannot be `none` as well as blocking the creation of join tokens with name `none`.


tomponline (approved):
Good spotting!

Issue #13553, opened 2024-06-05
Networks of type physical managed do not expose state


edlerd:
# Required information

 * Distribution: snap
 * Distribution version: 5.21

# Issue description

The `GET 1.0/networks/:name/state` endpoint returns 404 if the network is a managed physical network.

# Steps to reproduce

 1. Create a physical managed network `lxc network create foo --type physical parent=wlp0s20f3`
 2. Fetch the network state `lxc network info foo`
 3. `Error: Network interface "foo" not found` is returned because  `GET 1.0/networks/foo/state` returns 404. I would expect it to return something.

tomponline:
Whilst I agree this is a bug, I am not entirely sure what we should return here for this endpoint.

The problem is that for `physical` networks, the parent interface can disappear from the host if used by an `physical` NIC, and so most of the info will not be retrievable, or worse only retrievable when the instance isn't running (and the parent interface has been released back to the host).

But when the physical parent is being used as an OVN uplink it might be quite useful to see the counters of data flowing through it.

@edlerd would it be acceptable to sometimes return 404 on this endpoint if the parent interface is missing, either because its not been configured on the host or if its passed into a running instance?

PR #13965, opened 2024-08-22
lxd: Provide state for physical managed networks

boltmark:
This PR adds support for providing state for physical managed networks. When the parent of a physical managed network is available, we provide this parent state. When the parent is not available, we provide a response indicating that the state is `unavailable` and the type is `unknown`.

Closes https://github.com/canonical/lxd/issues/13553.


lxd/networks.go
```diff
@@ -1845,6 +1846,26 @@ func networkStateGet(d *Daemon, r *http.Request) response.Response {
...
 	}
 
+	// If network is of type physical managed, then return info on parent.
+	if n != nil && n.Type() == "physical" && n.IsManaged() {
+		config := n.Config()
+		state, err := resources.GetNetworkState(network.GetHostDevice(config["parent"], config["vlan"]))
+		if err != nil {
+			// If parent is not found, then return response indicating the parent is unavailable.
+			statusErr := api.StatusError{}
+			if errors.As(err, &statusErr) && statusErr.Status() == 404 {
```
tomponline, line 1856: Please use `api.StatusErrorCheck(err, http.StatusNotFound)` here rather than the manual errors.As call.

lxd/networks.go
```diff
@@ -1845,6 +1846,26 @@ func networkStateGet(d *Daemon, r *http.Request) response.Response {
...
+
+			return response.SmartError(err)
+		}
+
+		return response.SyncResponse(true, state)
+	}
+
 	var state *api.NetworkState
 	if n != nil {
 		state, err = n.State()
```
tomponline, line 1871: I've just seen this part of the code now (we didn't get this far down in our meeting) and I see now that the Network interface has a `State()` function that allows delegating the generation of the network state info to each Network driver.

So we shouldn't be adding network driver specific logic into this `networkStateGet` function but instead pushing down the driver specific implementation into the driver.

So we should be creating a State() function in https://github.com/canonical/lxd/blob/main/lxd/network/driver_physical.go to override the common function.

test/suites/network.sh
```diff
@@ -64,6 +64,15 @@ test_network() {
   lxc network create lxdt$$ ipv4.address=none ipv6.address=none
   lxc network delete lxdt$$
 
+  # Check that we can return state for physical managed networks
+  ip link add dummy0 type dummy
+  lxc network create lxdt$$ --type=physical parent=dummy0
+  lxc network info lxdt$$
```
tomponline, line 70: It would be good to test for Type and State in the positive scenario too

tomponline (changes requested):
Lets move the driver specific implementation into the network driver.

lxd/networks.go
boltmark: Fixed, thanks!

lxd/networks.go
boltmark: Moved the logic to the driver. Will remember to do so for similar tasks in the future. Thanks!

tomponline (approved):
Thanks!

Issue #11948, opened 2023-07-05
Return migration errors to client correctly


pcgeek86:
# Required information

 * Distribution: Ubuntu
 * Distribution version: 23.04 Lunar Lobster
 * The output of "lxc info" or if that fails:
   * Kernel version: `6.2.0-24-generic`
   * LXC version: 5.15
   * LXD version: 5.15
   * Storage backend in use: `zfs`

# Issue description

When I try to move an LXD virtual machine from one cluster member, to a different cluster member, I receive an error.

```
Error: Instance move to destination failed: Error transferring instance data: Failed migration on target: Failed getting migration target filesystem connection: websocket: bad handshake
```

```
+-----------------------------------------------------------------------------------------------------------------------+
|  +---------------------------+     +-------------------------+     +--------------------------+                       |
|  |lxd01                      |     |lxd03                    |     |lxd04                     |                       |
|  |LXD host                   |     |LXD host                 |     |LXD host                  |                       |
|  |Cluster member             |     |Cluster member           |     |Cluster member            |                       |
|  |                           |     |                         |     |                          |                       |
|  |                           |     |                         |     |                          |                       |
|  | +-----------------------+ |     |                         |     |                          |                       |
|  | |vm01                   | |     |                         |     |                          |                       |
|  | |Alpine Linux           | |     +-------------------------+     |                          |                       |
|  | |Default profile        | |                                     |                          |                       |
|  | +-----------------------+ |                                     |                          |                       |
|  |                   |       |                                     |                          |                       |
|  |                   |       |                                     |                          |                       |
|  |                   +---------------------------------------------|                          |                       |
|  |                           |  Move vm01 from lxd01 to lxd04      |                          |                       |
|  |                           |                                     |                          |                       |
|  |                           |  lxc move vm01 --target lxd04       |                          |                       |
|  |                           |                                     |                          |                       |
|  |                           |                                     |                          |                       |
|  +---------------------------+                                     +--------------------------+                       |
|                                                                                                                       |
|                                                                                                                       |
|                                                                                                                       |
|                                                                                                                       |
|LXD cluster (3 members)                                                                                                |
+-----------------------------------------------------------------------------------------------------------------------+
```

# Steps to reproduce

 1. Create a three-node LXD cluster
 2. Deploy a bunch of VMs across the cluster
 3. Try to move a VM between two different cluster members

# Information to attach

 - [ ] Any relevant kernel output (`dmesg`)
[...]

PR #13966, opened 2024-08-22
Migration: Forward migration errors (from Incus)

boltmark:
This PR adds functionality to report source errors during copy. This helps to solve the issue of the incorrect error being reported when a live migration is attempted while `migration.stateful` is missing or set to false.

(Also rectifies a few unchecked type assertions in `client/lxd_instances.go` since the linter complained)

Closes https://github.com/canonical/lxd/issues/11948.

### DIFF
--- a/client/lxd_instances.go
+++ b/client/lxd_instances.go
@@ -723,6 +723,9 @@ func (r *ProtocolLXD) tryCreateInstance(req api.InstancesPost, urls []string, op
 	operation := req.Source.Operation
 
 	// Forward targetOp to remote op
+	chConnect := make(chan error, 1)
+	chWait := make(chan error, 1)
+
 	go func() {
 		success := false
 		var errors []remoteOperationResult
@@ -762,13 +765,35 @@ func (r *ProtocolLXD) tryCreateInstance(req api.InstancesPost, urls []string, op
 			break
 		}
 
-		if !success {
-			rop.err = remoteOperationError("Failed instance creation", errors)
+		if success {
+			chConnect <- nil
+			close(chConnect)
+		} else {
+			chConnect <- remoteOperationError("Failed instance creation", errors)
+			close(chConnect)
+
 			if op != nil {
 				_ = op.Cancel()
 			}
 		}
+	}()
 
+	if op != nil {
+		go func() {
+			chWait <- op.Wait()
+			close(chWait)
+		}()
+	}
+
+	go func() {
+		var err error
+
+		select {
+		case err = <-chConnect:
+		case err = <-chWait:
+		}
+
+		rop.err = err
 		close(rop.chDone)
 	}()
 
--- a/client/lxd_instances.go
+++ b/client/lxd_instances.go
@@ -978,7 +978,12 @@ func (r *ProtocolLXD) CopyInstance(source InstanceServer, instance api.Instance,
 
 		targetSecrets := map[string]string{}
 		for k, v := range opAPI.Metadata {
-			targetSecrets[k] = v.(string)
+			vStr, ok := v.(string)
+			if !ok {
+				continue
+			}
+
+			targetSecrets[k] = vStr
 		}
 
 		// Prepare the source request
@@ -1006,7 +1011,12 @@ func (r *ProtocolLXD) CopyInstance(source InstanceServer, instance api.Instance,
 
 	sourceSecrets := map[string]string{}
 	for k, v := range opAPI.Metadata {
-		sourceSecrets[k] = v.(string)
+		vStr, ok := v.(string)
+		if !ok {
+			continue
+		}
+
+		sourceSecrets[k] = vStr
 	}
 
 	// Relay mode migration
@@ -1026,7 +1036,12 @@ func (r *ProtocolLXD) CopyInstance(source InstanceServer, instance api.Instance,
 		// Extract the websockets
 		targetSecrets := map[string]string{}
 		for k, v := range targetOpAPI.Metadata {
-			targetSecrets[k] = v.(string)
+			vStr, ok := v.(string)
+			if !ok {
+				continue
+			}
+
+			targetSecrets[k] = vStr
 		}
 
 		// Launch the relay
@@ -1268,9 +1283,16 @@ func (r *ProtocolLXD) ExecInstance(instanceName string, exec api.InstanceExecPos
 
 	value, ok := opAPI.Metadata["fds"]
 	if ok {
-		values := value.(map[string]any)
-		for k, v := range values {
-			fds[k] = v.(string)
+		values, ok := value.(map[string]any)
+		if ok {
+			for k, v := range values {
+				vStr, ok := v.(string)
+				if !ok {
+					continue
+				}
+
+				fds[k] = vStr
+			}
 		}
 	}
 
@@ -1285,7 +1307,12 @@ func (r *ProtocolLXD) ExecInstance(instanceName string, exec api.InstanceExecPos
 		outputs, ok := opAPI.Metadata["output"].(map[string]any)
 		if ok {
 			for k, v := range outputs {
-				outputFiles[k] = v.(string)
+				vStr, ok := v.(string)
+				if !ok {
+					continue
+				}
+
+				outputFiles[k] = vStr
 			}
 		}
 
@@ -2017,7 +2044,12 @@ func (r *ProtocolLXD) CopyInstanceSnapshot(source InstanceServer, instanceName s
 
 		targetSecrets := map[string]string{}
 		for k, v := range opAPI.Metadata {
-			targetSecrets[k] = v.(string)
+			vStr, ok := v.(string)
+			if !ok {
+				continue
+			}
+
+			targetSecrets[k] = vStr
 		}
 
 		// Prepare the source request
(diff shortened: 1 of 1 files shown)


lxd/instance/drivers/driver_qemu.go
```diff
@@ -6579,6 +6579,22 @@ func (d *qemu) MigrateSend(args instance.MigrateSendArgs) error {
 		return fmt.Errorf("Failed receiving migration offer response: %w", err)
 	}
 
+	// If we have unknown fields in the protobuf message, then we received
+	// unexpected data. This will only happen if we were sent a migration.MigrationControl
+	// instead of the expected migration.MigrationHeader. This implies that we have failed on
+	// the target before we were able to send the response header.
+	unknownFields := offerHeader.ProtoReflect().GetUnknown()
```
masnax, line 6586: Is this supposed to be `respHeader` instead of `offerHeader`? `offerHeader` is the type being sent, not the one being received for this block.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -6579,6 +6579,22 @@ func (d *qemu) MigrateSend(args instance.MigrateSendArgs) error {
...
+	if len(unknownFields) > 0 {
+		msg, err := shared.GetFirstUnknownField(unknownFields)
```
masnax, line 6588: Relying on the first field feels a bit arbitrary. I wonder if we can make the underlying `ControlReceive` a bit smarter to avoid this?

`ControlReceive` calls [`ProtoRecv`](https://github.com/canonical/lxd/blob/main/lxd/migration/wsproto.go#L14-L39) which unmarshals the data into the provided type (which it interprets as a `proto.Message`)

Could we split up `ProtoRecv` into two parts:
```go
// Reads from the conn and returns the received []byte.
func ProtoRead(ws *websocket.Conn) ([]byte, error)
```
and
```go
// Unmarshals a []byte into a proto.Message.
 func ProtoParse(buf []byte, msg proto.Message) error
```
Then I _think_ we could do something like this:
```go
func recv(m proto.Message) error {
    conn, err := c.conns[api.SecretNameControl].WebSocket(context.TODO())
    if err != nil {
        return fmt.Errorf("Control connection not initialized: %w", err)
    }

    // Get the underlying type of the proto.Message.
    t := reflect.TypeOf(m)
    if t.Kind() == reflect.Ptr {
        t = t.Elem()
    }

    // Read from the connection.
    buf, err := migration.ProtoRead(conn)
    if err != nil {
        return err
    }

    // If we are expecting a MigrationHeader, first check if we received a MigrationControl
    // which contains error messages indicating that we need to stop.
    if t == reflect.TypeOf(migration.MigrationHeader{}) {
        ctl := &migration.MigrationControl{}
        err := migration.ProtoParse(buf, ctl)
        if err != nil {
            return err
        }

        // Check if the fields are non-nil, indicating that we actually did receive a MigrationControl.
        // If Success if non-nil and false, then we received an error.
        if ctl.Success != nil && ctl.Message != nil && !ctl.GetSuccess() {
            return fmt.Errorf(ctl.GetMessage())
        }
    }

    // Properly parse into the correct type before returning.
    return migration.ProtoParse(buf, m)
}
```

masnax: In particular, I have no idea if this block is correct:
```go
 // Check if the fields are non-nil, indicating that we actually did receive a MigrationControl.
        // If Success if non-nil and false, then we received an error.
        if ctl.Success != nil && ctl.Message != nil && !ctl.GetSuccess() {
            return fmt.Errorf(ctl.GetMessage())
        }
```

masnax: Oh, you can actually simplify out the direct `reflect` package usage by just making a type assertion:
```go
_, ok := m.(*migration.MigrationHeader)
if ok {
    ctl := &migration.MigrationControl{}
    err := migration.ProtoParse(buf, ctl)
    if err != nil {
        return err
    }
 
    // Check if the fields are non-nil, indicating that we actually did receive a MigrationControl.
    // If Success if non-nil and false, then we received an error.
    if ctl.Success != nil && ctl.Message != nil && !ctl.GetSuccess() {
        return fmt.Errorf(ctl.GetMessage())
    }
}
```

lxd/instance/drivers/driver_qemu.go
boltmark: Oops yes, nice catch. Marking as resolved since it is no longer relevant for updated code :)

lxd/instance/drivers/driver_qemu.go
boltmark: I really appreciate this! I've pushed new changes to use a fix along these lines instead of relying on unknown fields.

lxd/migrate_instance.go
```diff
@@ -152,7 +152,10 @@ func (s *migrationSourceWs) Do(state *state.State, migrateOp *operations.Operati
 	})
 	if err != nil {
 		l.Error("Failed migration on source", logger.Ctx{"err": err})
-		return fmt.Errorf("Failed migration on source: %w", err)
+
+		errMsg := fmt.Errorf("Failed migration on source: %w", err)
+		s.sendControl(errMsg)
```
tomponline, line 157: @boltmark what was the reason we couldnt cherry-pick this one by itself?

tomponline: @boltmark in what scenarios does this help with compared with https://github.com/canonical/lxd/pull/13894/files ?

lxd/instance/drivers/driver_qemu.go
```diff
@@ -6509,6 +6504,11 @@ func (d *qemu) MigrateSend(args instance.MigrateSendArgs) error {
 		return err
 	}
 
+	// Check for stateful support.
+	if args.Live && shared.IsFalseOrEmpty(d.expandedConfig["migration.stateful"]) {
```
tomponline, line 6508: Please can you expand the commit message to explain what this fixes.

tomponline (commented):
Overall this looks encouraging @boltmark thanks!

lxd/migrate.go
```diff
@@ -72,7 +72,34 @@ func (c *migrationFields) recv(m proto.Message) error {
...
+		return err
+	}
+
+	// If we are expecting a MigrationHeader, first check if we received a MigrationControl
+	// which may contain a message indicating that we need to stop.
+	_, ok := m.(*migration.MigrationHeader)
+	if ok {
+		ctl := &migration.MigrationControl{}
+
+		// Enforce required fields present.
```
tomponline, line 86: Which part of the code does this? AllowPartialFalse? If so can you move the comment to next to the field?

lxd/migrate.go
```diff
@@ -72,7 +72,34 @@ func (c *migrationFields) recv(m proto.Message) error {
...
+			AllowPartial: false,
+		}
+
+		// Attempt to unmarshal into a MigrationControl.
+		err := opts.Unmarshal(buf, ctl)
+
+		// If we successfully parsed a MigrationControl and the Success field is false,
+		// then we received an error instead of a proper MigrationHeader and should fail.
+		if err == nil && migration.IsValidMigrationControlWithMessage(ctl) && !ctl.GetSuccess() {
+			return fmt.Errorf("%s", ctl.GetMessage())
```
tomponline, line 97: errors.New()

lxd/migrate_instance.go
boltmark: The reason we cannot use the cherry-pick by itself is because although it sends the error to the other side, this does not help if the other side does not read it. In the particular case for the `migration.stateful` check, the issue is twofold:
1. By not setting up the filesystem connection on the source before making this check, we cause the sink to fail before reading from the control connection. This means that even though we send over the error as per the cherry-pick, the sink never reads it in this case.
2. The first time the sink reads from the control connection, it does so expecting to receive a `MigrationHeader`. Therefore if the source fails before sending a header and instead sends an error, the sink will read the `MigrationControl` error message and (incorrectly) parse it into a `MigrationHeader`. The update to the `recv` function ensures we check for this case and return the error if one is found.

As for the particular scenarios this helps with compared to the other PR, this handles error checking for non-clustered servers, since in this case we cannot enforce that we always return the error for the source.

lxd/instance/drivers/driver_qemu.go
boltmark: Done!

lxd/migrate.go
boltmark: Yes, fixed!

lxd/migrate.go
boltmark: Thanks, fixed!

lxd/migrate_instance.go
boltmark: Ah, sorry, I may have misread that. Do you mean why are we cherry-picking the added `sendControl` for both the source and the sink `Do` functions as opposed to just the source as in this code block?

tomponline: Yes please, this fix is in the client similar to what we discussed originally. Ta

boltmark:
@tomponline I've updated this PR to cherry-pick the incus fix. Marking as ready.

tomponline (approved):

PR #13969, opened 2024-08-23
Shared: Add HMAC and cert utils

roosterfish:
This PR adds additional HMAC and cert utils to the `shared` package.

The HMAC implementation follows the practices from the [explicit trust establishment spec](https://discourse.ubuntu.com/t/explicit-trust-establishment-mechanism-for-microcloud/44261).
It's imported as a general utility in LXD to be used by MicroCloud during trust establishment.

In order to make the use of a KDF (in this case argon2) optional, there is a second factory function that allows using some extended functionality added by argon2.

To create a HMAC header, the caller runs:

```go
// Specific HMAC version to allow indicating protocol changes in the future.
const HMACMicroCloud10 trust.HMACVersion = "MicroCloud-1.0"

// Basic format
hmac := trust.NewHMAC(<key>, NewDefaultHMACConf("LXD1.0"))
// Create the "Authorization" header in the format `Authorization: <version> <HMAC>`
header, err := trust.HMACAuthorizationHeader(hmac, <message body to be sent>)

// Or custom format that extends HMAC with a KDF (in this case argon2).
hmac, err := trust.NewHMACArgon2(<password>, <salt or nil>, NewDefaultHMACConf("LXD1.0"))
// Create the "Authorization" header in the format `Authorization: <version> <salt>:<HMAC>`
header, err := trust.HMACAuthorizationHeader(hmac, <message body to be sent>)
```

On the server side the validation happens as follows:

```go
// Initialize the expected HMAC format
hmac, err := trust.NewHMACArgon2(<password>, <salt or nil>, NewDefaultHMACConf("LXD1.0"))

// Validate the HMAC based on the given *http.Request r
err = trust.HMACEqual(hmac, r)
```

### DIFF
--- a/shared/cert.go
+++ b/shared/cert.go
@@ -430,6 +430,16 @@ func GenerateMemCert(client bool, options CertOptions) ([]byte, []byte, error) {
 	return cert, key, nil
 }
 
+// ParseCert parse a X.509 certificate from the given byte slice and return its parsed content.
+func ParseCert(cert []byte) (*x509.Certificate, error) {
+	certBlock, _ := pem.Decode(cert)
+	if certBlock == nil {
+		return nil, fmt.Errorf("Invalid certificate file")
+	}
+
+	return x509.ParseCertificate(certBlock.Bytes)
+}
+
 // ReadCert reads a X.509 certificate from the filesystem, do PEM decoding and return its parsed content.
 func ReadCert(fpath string) (*x509.Certificate, error) {
 	cf, err := os.ReadFile(fpath)
@@ -437,12 +447,7 @@ func ReadCert(fpath string) (*x509.Certificate, error) {
 		return nil, err
 	}
 
-	certBlock, _ := pem.Decode(cf)
-	if certBlock == nil {
-		return nil, fmt.Errorf("Invalid certificate file")
-	}
-
-	return x509.ParseCertificate(certBlock.Bytes)
+	return ParseCert(cf)
 }
 
 // CertFingerprint returns the SHA256 fingerprint of a X.509 certificate.
--- a/shared/trust/hmac.go
+++ b/shared/trust/hmac.go
@@ -1,0 +1,202 @@
+package trust
+
+import (
+	"bytes"
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"hash"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/canonical/lxd/shared/api"
+)
+
+// HMACFormatter represents arbitrary formats to diplay and parse the actual HMAC.
+// For example implementations like argon2 extend the format with an additional salt.
+// Example using argon2: `Authorization: <version> <salt>:<HMAC>`.
+type HMACFormatter interface {
+	// The Write* methods allow the creation of an HMAC based on various inputs.
+	WriteBytes(b []byte) ([]byte, error)
+	WriteJSON(v any) ([]byte, error)
+	WriteRequest(r *http.Request) ([]byte, error)
+
+	// Version returns the current HMAC version set for the format.
+	Version() HMACVersion
+	// HTTPHeader expects the HMAC computed over the payload and returns the final Authorization header.
+	HTTPHeader(hmac []byte) string
+	// ParseHTTPHeader expects an Authorization header and returns the used version and HMAC.
+	ParseHTTPHeader(header string) (HMACVersion, []byte, error)
+	// Equal compares two HMACs and returns true in case of a match.
+	Equal(hmac1 []byte, hmac2 []byte) bool
+}
+
+// HMAC represents the the tooling for creating and validating HMACs.
+type HMAC struct {
+	conf HMACConf
+	key  []byte
+}
+
+// HMACVersion indicates the version used for the authorization header format.
+// This allows to define a format used by the header so that the scheme can be modified
+// in future implementations without breaking already existing versions.
+// An example version can be `LXD1.0` which indicates that this is version 1.0 of the
+// LXD HMAC authentication scheme.
+// The format used after the version is dependant on the actual implementation:
+// Example: `Authorization: <version> <format including the HMAC>`.
+type HMACVersion string
+
+// HMACConf represents the HMAC configuration.
+type HMACConf struct {
+	HashFunc func() hash.Hash
+	Version  HMACVersion
+}
+
+// NewDefaultHMACConf returns the default configuration for HMAC.
+func NewDefaultHMACConf(version HMACVersion) HMACConf {
+	return HMACConf{
+		HashFunc: sha256.New,
+		Version:  version,
+	}
+}
+
+// NewHMAC returns a new instance of HMAC.
+func NewHMAC(key []byte, conf HMACConf) HMACFormatter {
+	return &HMAC{
+		conf: conf,
+		key:  key,
+	}
+}
+
+func (h *HMAC) splitVersionFromHMAC(header string) (HMACVersion, string, error) {
+	authHeaderSplit := strings.Split(header, " ")
+	if len(authHeaderSplit) != 2 {
+		return "", "", errors.New("Version or HMAC is missing")
+	}
+
+	if authHeaderSplit[0] == "" {
+		return "", "", errors.New("Version cannot be empty")
+	}
+
+	if authHeaderSplit[1] == "" {
+		return "", "", errors.New("HMAC cannot be empty")
+	}
+
+	return HMACVersion(authHeaderSplit[0]), authHeaderSplit[1], nil
+}
+
+// HTTPHeader returns the actual HMAC together with the used version.
+func (h *HMAC) HTTPHeader(hmac []byte) string {
+	return fmt.Sprintf("%s %s", h.conf.Version, hex.EncodeToString(hmac))
+}
+
+// Version returns the used HMAC version.
+func (h *HMAC) Version() HMACVersion {
+	return h.conf.Version
+}
+
+// ParseHTTPHeader extracts the actual version and HMAC from the Authorization header.
+func (h *HMAC) ParseHTTPHeader(header string) (HMACVersion, []byte, error) {
+	version, hmacStr, err := h.splitVersionFromHMAC(header)
+	if err != nil {
+		return "", nil, err
+	}
+
+	hmac, err := hex.DecodeString(hmacStr)
+	if err != nil {
+		return "", nil, fmt.Errorf("Failed to decode the HMAC: %w", err)
+	}
+
+	return version, hmac, nil
+}
+
+// WriteBytes creates a new HMAC hash using the given bytes.
+func (h *HMAC) WriteBytes(b []byte) ([]byte, error) {
+	mac := hmac.New(h.conf.HashFunc, h.key)
+	_, err := mac.Write(b)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to create HMAC: %w", err)
+	}
+
+	return mac.Sum(nil), nil
+}
+
+// WriteJSON creates a new HMAC hash using the given struct.
+func (h *HMAC) WriteJSON(v any) ([]byte, error) {
+	payload, err := json.Marshal(v)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to marshal payload: %w", err)
+	}
+
+	return h.WriteBytes(payload)
+}
+
+// WriteRequest creates a new HMAC hash using the given request.
+// It will extract the requests body.
+func (h *HMAC) WriteRequest(r *http.Request) ([]byte, error) {
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to read request body: %w", err)
+	}
+
+	defer func() {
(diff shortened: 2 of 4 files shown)


roosterfish:
I am still not sure if we want to track the version within the format. Thinking about possible scenarios here.

shared/cert.go
```diff
@@ -430,19 +430,24 @@ func GenerateMemCert(client bool, options CertOptions) ([]byte, []byte, error) {
 	return cert, key, nil
 }
 
+// ParseCert parse a X.509 certificate from the given byte slice and return its parsed content.
+func ParseCert(cert []byte) (*x509.Certificate, error) {
```
tomponline, line 434: Would it be useful to adapt this to take an io.Reader instead so that a bytes.Buffer or actual file handle can be used? e.g. https://pkg.go.dev/bytes#example-Buffer-Reader

tomponline (changes requested):
Please could we have some tests that are usage examples for this package?

tomponline:
> This PR adds additional HMAC and cert utils to the `shared` package.
> 
> The HMAC implementation follows the practices from the [explicit trust establishment spec](https://discourse.ubuntu.com/t/explicit-trust-establishment-mechanism-for-microcloud/44261). It's imported as a general utility in LXD to be used by MicroCloud during trust establishment.
> 
> In order to make the use of a KDF (in this case argon2) optional, a format interface is added to allow plugging this functionality in on demand.
> 
> To create a HMAC header, the caller will run:
> 
> ```go
> // Specific HMAC version to allow indicating protocol changes in the future.
> const HMACMicroCloud10 trust.HMACVersion = "MicroCloud-1.0"
> 
> // Custom format that extends HMAC with a KDF (in this case argon2).
> argonFormat, err := trust.NewDefaultHMACFormatArgon()
> if err != nil {
>     return err
> }
> 
> hmac := trust.NewHMAC(trust.NewDefaultHMACConf(<passphrase>, HMACMicroCloud10)).WithFormat(argonFormat)
> 
> // Create the "Authorization" header in the format `Authorization: <version> <salt>:<HMAC>`
> header, err := hmac.AuthorizationHeader(<message body to be sent>)
> if err != nil {
>     return err
> }
> ```
> 
> On the server side the validation happens as follows:
> 
> ```go
> argonFormat, err := trust.NewDefaultHMACFormatArgon()
> if err != nil {
>     return response.InternalError(fmt.Errorf("Failed to setup argon: %w", err))
> }
> 
> requestHMAC := trust.NewHMAC(trust.NewDefaultHMACConf(<passphrase>, HMACMicroCloud10)).WithFormat(argonFormat)
> 
> // Validate the HMAC based on the given *http.Request r
> err = requestHMAC.ValidateAuthorizationHeader(r)
> if err != nil {
>     return response.SmartError(err)
> }
> ```

Is the expectation that this will be used in an http middleware? 

I'm a bit concerned how this is tightly coupled to the http package (although I do recall as discussing the balance of making it flexible and easy to test vs making it opinionated for the microcloud trust scenario).

Again, im wondering a bytes.Buffer would help to decouple this from the http package, and make setting/getting the Authorization header an excercise for the caller?

shared/cert.go
roosterfish: It looks that the upstream `encoding/pem` is a limitation here. Whilst it accepts a writer for encoding, decoding is only supported using a `[]byte` slice. See https://pkg.go.dev/encoding/pem#Decode. 

So even if we accept an `io.Reader` here, we would still need to read it's contents and create a slice before we can decode the contents.

shared/trust/hmac.go
```diff
@@ -0,0 +1,251 @@
...
+	"net/http"
+	"strings"
+
+	"golang.org/x/crypto/argon2"
+
+	"github.com/canonical/lxd/shared/api"
+)
+
+// HMACVersion indicates the version used for the authorization header format.
+type HMACVersion string
```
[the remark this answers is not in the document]
roosterfish: Fixed.

roosterfish:
> Please could we have some tests that are usage examples for this package?

Added. I have split the tests into HMAC creation and validation to demonstrate both usage scenarios.

shared/trust/hmac.go
```diff
@@ -0,0 +1,265 @@
...
+type HMAC struct {
+	conf   HMACConf
+	format HMACFormat
+}
+
+// HMACFormatDefault represents the default format.
+type HMACFormatDefault struct{}
+
+// HMACFormatArgon represents the argon2 format.
+type HMACFormatArgon struct {
```
tomponline, line 63: Shall we call this Argon2 rather than Argon to be clear what it implements?

shared/trust/hmac.go
```diff
@@ -0,0 +1,265 @@
...
+func NewHMAC(conf HMACConf) *HMAC {
+	return &HMAC{
+		conf:   conf,
+		format: &HMACFormatDefault{},
+	}
+}
+
+// WithFormat configures the HMAC instance to use the specified format.
+// The format can be used to extend the way of displaying and parsing the resulting header.
+func (h *HMAC) WithFormat(format HMACFormat) *HMAC {
```
tomponline, line 176: Can we just have different factory functions for the different formats, rather than having the ability to switch between them. I'm finding the method chaining functionality here a bit difficult to comprehend why it is needed?

shared/trust/hmac.go
```diff
@@ -0,0 +1,265 @@
...
+// HMACFormat represents arbitrary formats to diplay and parse the actual HMAC.
+// Implementations like argon extend the format with an additional salt.
+// The Header function accepts the used version and the HMAC computed over the payload.
+// It is expected to return the final Authorization header.
+// The ParseHeader function accepts a requests header and is expected to return
+// the used version and the HMAC.
+// The Passphrase function is invoked so that the format can learn about the secret
+// used for generating the HMAC and can apply modifications such as running it through a KDF.
+// Example using argon2: `Authorization: <version> <salt>:<HMAC>`.
+type HMACFormat interface {
```
tomponline, line 47: Lets call this HMACFormatter so it has a verb in it, which is ideally what interface names should have.

shared/trust/hmac.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+// HMAC represents the the tooling for creating and validating HMACs.
+type HMAC struct {
+	conf   HMACConf
+	format HMACFormat
+}
+
+// HMACFormatDefault represents the default format.
+type HMACFormatDefault struct{}
```
tomponline, line 60: Please can you separate the two different hmac format options and their functions into separate files, e.g. hmac_default.go and hmac_argon2.go for clarity? You can keep the common entities in hmac.go as they are now.

shared/trust/hmac.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+}
+
+// Passphrase accepts the outer HMAC passphrase and returns the same without modification.
+func (h *HMACFormatDefault) Passphrase(passphrase []byte) []byte {
+	return passphrase
+}
+
+// Header returns the actual HMAC alongside together with the used version.
+func (h *HMACFormatDefault) Header(version HMACVersion, hmac []byte) string {
```
tomponline, line 86: Maybe call this HTTPHeader as to indicate where it should be used?

shared/trust/hmac.go
```diff
@@ -0,0 +1,265 @@
...
+		return "", nil, fmt.Errorf("Failed to decode the HMAC: %w", err)
+	}
+
+	return HMACVersion(authHeaderSplit[0]), hmac, nil
+}
+
+// NewDefaultHMACFormatArgon returns the default format for argon2.
+// Recommended defaults according to https://www.rfc-editor.org/rfc/rfc9106#section-4-6.2.
+// We use the second recommended option to not require a system having 2 GiB of memory.
+func NewDefaultHMACFormatArgon() (*HMACFormatArgon, error) {
```
tomponline, line 108: Can this take a salt argument which can avoid generating a random one if provided?

shared/trust/hmac.go
roosterfish: Fixed.

shared/trust/hmac.go
roosterfish: Fixed. There is now a `NewHMAC` and `NewHMACArgon2` which both return a struct that implements the `HMACFormatter`.

shared/trust/hmac.go
roosterfish: Fixed.

shared/trust/hmac.go
roosterfish: Fixed.

shared/trust/hmac.go
```diff
@@ -0,0 +1,265 @@
...
+
+	h.Salt = salt
+
+	hmac, err := hex.DecodeString(authHeaderDetails[1])
+	if err != nil {
+		return "", nil, fmt.Errorf("Failed to decode the HMAC: %w", err)
+	}
+
+	return HMACVersion(authHeaderSplit[0]), hmac, nil
+}
```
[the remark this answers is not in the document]
roosterfish: Fixed.
Empty strings are now forbidden for the version, HMAC and salt (in case of argon2).
I have also added a few more test cases for these scenarios.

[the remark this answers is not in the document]
Fedqys: thanks !

shared/trust/hmac.go
```diff
@@ -0,0 +1,205 @@
...
+// HMACFormatter represents arbitrary formats to diplay and parse the actual HMAC.
+// For example implementations like argon2 extend the format with an additional salt.
+// Example using argon2: `Authorization: <version> <salt>:<HMAC>`.
+type HMACFormatter interface {
+	// The Write* methods allow the creation of an HMAC based on various inputs.
+	WriteBytes(b []byte) ([]byte, error)
+	WriteJSON(v any) ([]byte, error)
+	WriteRequest(r *http.Request) ([]byte, error)
+
+	splitVersionFromHMAC(header string) (HMACVersion, string, error)
```
tomponline, line 28: Does this need to be in the interface? Its not exported.

shared/trust/hmac.go
```diff
@@ -0,0 +1,205 @@
...
+
+	// Version returns the current HMAC version set for the format.
+	Version() HMACVersion
+	// HTTPHeader expects the HMAC computed over the payload and returns the final Authorization header.
+	HTTPHeader(hmac []byte) string
+	// ParseHeader expects an Authorization header and returns the used version and HMAC.
+	ParseHTTPHeader(header string) (HMACVersion, []byte, error)
```
tomponline, line 35: comment/name mismatch

shared/trust/hmac.go
```diff
@@ -0,0 +1,205 @@
...
+
+	if !hmac.Equal(hmacFromHeader, hmacFromBody) {
+		return api.StatusErrorf(http.StatusForbidden, "Invalid HMAC")
+	}
+
+	return nil
+}
+
+// HMACAuthorizationHeader returns the HMAC on v using the given formatter.
+func HMACAuthorizationHeader(hmac HMACFormatter, v any) (string, error) {
```
tomponline, line 198: the `hmac` argument covers the `hmac` package import - a small code smell.

shared/trust/hmac.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	if !hmac.Equal(hmacFromHeader, hmacFromBody) {
+		return api.StatusErrorf(http.StatusForbidden, "Invalid HMAC")
+	}
+
+	return nil
+}
+
+// HMACAuthorizationHeader returns the HMAC on v using the given formatter.
```
tomponline, line 197: Returns the HMAC as an Authorization header?

shared/trust/hmac.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	hmacVersion := hmac.Version()
+	if version != hmacVersion {
+		return api.StatusErrorf(http.StatusBadRequest, "Authorization header uses version %q but expected %q", version, hmacVersion)
+	}
+
+	hmacFromBody, err := hmac.WriteRequest(r)
+	if err != nil {
+		return api.StatusErrorf(http.StatusInternalServerError, "Failed to write request body: %w", err)
```
tomponline, line 187: This sounds a bit like its writing output in response to the request.

How about "Failed to calculate HMAC from request body"?

roosterfish: Fixed.

shared/trust/hmac.go
roosterfish: Fixed.

shared/trust/hmac.go
roosterfish: Thanks for the suggestion, I was thinking about an alternative here but couldn't find a good alternative.

shared/trust/hmac.go
roosterfish: Fixed. Also in the `HMACEqual` func.

shared/trust/hmac.go
roosterfish: Fixed.

shared/trust/hmac.go
roosterfish: No this doesn't need to be in the interface.
I put it there as I wanted to access it from all the implementations of the interface but realized it's enough to have it as part of the base struct as this one is already embedded in the argon struct.

Fixed and removed.

tomponline (approved):
Thanks LGTM

Issue #13464, opened 2024-05-09
`lxd recover` loses the expiration dates of snapshots.


simondeziel:
I just went through a successful `lxd recover` which reimport my container and its snapshots from the intact zpool. Snapshots are taken on a schedule:

```sh
# lxc config show -e ganymede | grep snapshot
  snapshots.expiry: 3d
  snapshots.schedule: '@daily, @startup'
```

However, after `lxd recover` brought those snapshots back, they lost their `expires at` field:

```
# lxc info ganymede | sed -n '/^Snapshots:$/,$ p'
Snapshots:
+---------+----------------------+----------------------+----------+
|  NAME   |       TAKEN AT       |      EXPIRES AT      | STATEFUL |
+---------+----------------------+----------------------+----------+
| snap222 | 2024/05/07 05:28 UTC |                      | NO       |
+---------+----------------------+----------------------+----------+
| snap223 | 2024/05/08 05:28 UTC |                      | NO       |
+---------+----------------------+----------------------+----------+
| snap224 | 2024/05/09 05:28 UTC |                      | NO       |
+---------+----------------------+----------------------+----------+
| snap225 | 2024/05/09 22:21 UTC | 2024/05/12 22:21 UTC | NO       |
+---------+----------------------+----------------------+----------+
```

In the above, `snap225` was taken after the `lxd recover`.

However, the instance's `backup.yaml` should have had this information which is where it learned about the `taken at` field. That said, it seems the recovery have overwritten the `backup.yaml` with the bogus values now:

```
# LD_LIBRARY_PATH=/snap/lxd/current/lib/:/snap/lxd/current/lib/x86_64-linux-gnu/ nsenter --mount=/run/snapd/ns/lxd.mnt sed -n '/^volume_snapshots:$/,$ p' /var/snap/lxd/common/lxd/storage-pools/default/containers/ganymede/backup.yaml
volume_snapshots:
- name: snap222
  description: ""
  content_type: filesystem
  created_at: 0001-01-01T00:00:00Z
  expires_at: 0001-01-01T00:00:00Z
  config:
    volatile.uuid: a951c197-c11f-4fcc-a76e-ec575b99e305
- name: snap223
  description: ""
  content_type: filesystem
  created_at: 0001-01-01T00:00:00Z
  expires_at: 0001-01-01T00:00:00Z
  config:
    volatile.uuid: 2b8f7c0c-92da-4170-af35-c5033ec6b89c
- name: snap224
  description: ""
[...]

PR #13972, opened 2024-08-23
lxd/recover: Add snapshot expiry when recovering instance snapshots

kadinsayani:
Fixes canonical/lxd#13464

### DIFF
--- a/lxd/api_internal_recover.go
+++ b/lxd/api_internal_recover.go
@@ -574,12 +574,23 @@ func internalRecoverImportInstanceSnapshot(s *state.State, pool storagePools.Poo
 		return nil, err
 	}
 
+	snapshotExpiry := snap.Config["snapshots.expiry"]
+	if snapshotExpiry != "" {
+		expiry, err := shared.GetExpiry(snap.CreatedAt, snapshotExpiry)
+		if err != nil {
+			return nil, err
+		}
+
+		snap.ExpiresAt = expiry
+	}
+
 	_, snapInstOp, cleanup, err := instance.CreateInternal(s, db.InstanceArgs{
 		Project:      projectName,
 		Architecture: arch,
 		BaseImage:    snap.Config["volatile.base_image"],
 		Config:       snap.Config,
 		CreationDate: snap.CreatedAt,
+		ExpiryDate:   snap.ExpiresAt,
 		Type:         instanceType,
 		Snapshot:     true,
 		Devices:      deviceConfig.NewDevices(snap.Devices),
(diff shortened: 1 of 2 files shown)


lxd/api_internal_recover.go
```diff
@@ -574,12 +574,23 @@ func internalRecoverImportInstanceSnapshot(s *state.State, pool storagePools.Poo
 		return nil, err
 	}
 
+	snapshotExpiry := snap.Config["snapshots.expiry"]
```
tomponline, line 577: Do we need this check?

simondeziel:
I'm not familiar enough with the code to review but I added a simple reproducer highlighting the bogus `backup.yaml` file LXD generates during snapshots: https://github.com/canonical/lxd/issues/13464

lxd/api_internal_recover.go
kadinsayani: `lxd restore` looks for the expiry date in the volume_snapshots section as shown here: https://github.com/canonical/lxd/issues/13464#issuecomment-2307782519. The check I've added imports the instance snapshot with the correct expiry date, but doesn't address the underlying issue of volume_snapshots having zeroed expiry dates in the `backup.yaml` and storage_volumes_snapshots table.  I'm going to look further into this and introduce a better fix.

tomponline: thanks.

>  but doesn't address the underlying issue of volume_snapshots having zeroed expiry dates in the backup.yaml and storage_volumes_snapshots table. I'm going to look further into this and introduce a better fix.

I'm not sure that is an issue TBH, because the associated storage volume for an instance snapshot is always deleted when the instance snapshot is removed.

tomponline (approved):

tomponline (changes requested):
Oh actually please can you update the recover tests to check for snapshot expiry restoration. Ta

test/suites/backup.sh
```diff
@@ -103,8 +103,10 @@ EOF
     echo "hello world" | lxc exec c1 --project test -- tee /mnt/test.txt
     [ "$(lxc exec c1 --project test -- cat /mnt/test.txt)" = "hello world" ]
     lxc stop -f c1
+    lxc config set c1 snapshots.expiry 1d
     lxc snapshot c1
     lxc info c1
+    snapshotExpiryDateBefore=$(lxc info c1 | grep "snap0" | awk -F '|' '{print $4' | xargs)
```
simondeziel, line 109: This has a bogus awk script but how about simply saving the whole `snap0` info for comparison?

```
snapshotExpiryDateBefore="$(lxc info c1 | grep -wF "snap0")"
```

kadinsayani: Thanks for pointing that out!

tomponline (approved):
thanks!

PR #13986, opened 2024-08-26
storage: Conserve LV state (From Incus)

MggMuggins:
Reactivates an already-active volume after its size has been increased. If a volume was active before being grown, it would be deactivated by `SetVolumeQuota`; this retains the state of the LV after a size adjustment.

Contains cherry-picks from https://github.com/lxc/incus/pull/1021 and https://github.com/lxc/incus/pull/1136

incus#1021 is only partially applied as its other commit depends on https://github.com/lxc/incus/pull/512; I've made a note of this in the tracking sheet.


tomponline (approved):
Thanks!

PR #13989, opened 2024-08-27
Add guest OS compatibility matrix

simondeziel:


doc/guest-os-compatibility.md
```diff
@@ -0,0 +1,64 @@
+(guest-os-compatibility)=
+# Guest OS compatibility
+
+Tested on LXD `5.21/stable`
+
+OS vendor | OS version         | OS support | LXD agent | VirtIO-SCSI | VirtIO-BLK | NVMe    | CSM (BIOS) | UEFI | Secure Boot
```
mionaalex, line 6: Can we also add a footnote on what the lxd-agent does (or link to a doc section that explains it, I couldn't find it)? This page is likely to be shared with users/customers who might not be familiar with LXD, so seeing a bunch of X's might make them feel like this is a problem.

mionaalex (approved):
This looks good to me, just one comment on the lxd-agent.

doc/guest-os-compatibility.md
simondeziel: I added a proper note as I couldn't find any existing doc for this mechanism.

Maybe I should move the `LXD agent` column to the right so that it's less prominent. WDYT?

tomponline (approved):

PR #13994, opened 2024-08-27
Project: Add `lxd/project/limits` and fix PowerFlex instance size limits

### LINKED ISSUE #12567: Update projects instance limit accounting
# Issue description

Restricted projects count its already occupied resources for each specified limit. Doing this for VMs also takes the `size.state` value into account. However starting with https://github.com/canonical/lxd/pull/12566 it depends on the storage drivers `DefaultVMBlockFilesystemSize` setting how large the actual default value is. 

Currently we cannot easily resolve the storage driver within the `lxd/project` package in order to pick the value from the drivers `Info{}` struct. The accounting process might also receive changes in the near future which could make this story obsolete.

In the meantime this is causing wrong values for the projects instance limits accounting in case the upcoming PowerFlex driver is used.

## Possible Solution

In case we still need to resolve the driver within `lxd/project`, one option would be to pass down the drivers `Info{}` as an interface, so that a package level interface in `lxd/project` can be used to implement a subset of the interface coming from the driver to prevent having an import loop:

```go
type Driver interface {
	// Internal.
	Info() Informer
```

MggMuggins:
Fixes #12567 

Moves some the contents of `lxd/project` into a new module `lxd/project/limits` to remove an import cycle `lxd/project -> lxd/storage/drivers -> lxd/project`.

`lxd/project` split fairly cleanly in two; I tried to leave as much project-related API in `lxd/project` as I could. Most of what is left there is the string processing code for prepending instance/volume names with the project name. This is used fairly heavily in `lxd/storage/drivers`.

`lxd/project/limits` now contains just the code for enforcing project limits; the change in usage reads better IMO:
```go
project.AllowInstanceCreation(s.GlobalConfig, tx, projectName, req)
// to 
limits.AllowInstanceCreation(s.GlobalConfig, tx, projectName, req)
```

This change allows us to import the storage drivers from `lxd/storage/drivers` in `lxd/project/limits` to get `DefaultVMBlockFilesystemSize`.

This introduces a new method to the internal driver interface (`defaultVMBlockFilesystemSize() string`) and exposes a function from `lxd/storage/drivers` to convert a driver name to the `DefaultVMBlockFilesystemSize`.


lxd/storage/backend_lxd.go
```diff
@@ -7565,7 +7566,7 @@ func (b *lxdBackend) CreateCustomVolumeFromISO(projectName string, volName strin
 	}
 
 	err := b.state.DB.Cluster.Transaction(b.state.ShutdownCtx, func(ctx context.Context, tx *db.ClusterTx) error {
-		return project.AllowVolumeCreation(b.state.GlobalConfig, tx, projectName, req)
+		return limits.AllowVolumeCreation(b.state.GlobalConfig, tx, projectName, req)
```
tomponline, line 7569: Can you move this package toe lxd/project/limits so we can see its still related to projects.

tomponline:
> I'm happy to put up another PR against `stable-5.21` with the "wrong way" to fix #12567 (simply checking against a literal `"powerflex"` in `lxd/project`) if that would be preferable as a backport.

I'll backport these changes as part of the normal process as they dont change the API or the DB.

MggMuggins:
I've updated the PR description, made the requested change, and reworked how the default size was being passed around.

lxd/project/limits/permissions.go
```diff
@@ -1233,10 +1241,11 @@ func fetchProject(globalConfig map[string]any, tx *db.ClusterTx, projectName str
 	}
 
 	info := &projectInfo{
-		Project:   *project,
-		Profiles:  profiles,
-		Instances: instances,
-		Volumes:   volumes,
+		Project:      *project,
+		Profiles:     profiles,
+		StoragePools: pools,
```
roosterfish, line 1246: Should this be limited to only pools which are actually used in the project? 
I guess `tx.GetStoragePools` returns all of them.
Not sure if there is a filter option like for the profiles.

lxd/storage/drivers/load_test.go
```diff
@@ -0,0 +1,18 @@
+package drivers
```
roosterfish, line 1: IIRC that was the reasoning back then for having the `DefaultVMBlockFilesystemSize` func on the `common` struct so that there is a default defined which can then be overwritten by the PowerFlex driver.

Is there a specific reason for getting rid of this pattern?

roosterfish (changes requested):

lxd/storage/drivers/load_test.go
MggMuggins: We can't load a driver from `lxd/project/limits` because we don't have easy access to a `state.State`. 

Initially I was going to grab an entire `Info{}` from the driver but that should really [require that the driver be loaded](https://github.com/canonical/lxd/blob/main/lxd/storage/drivers/driver_btrfs.go#L95). This was a compromise so that the default can be defined per-driver and also not require the drivers to be loaded.

The test is just designed to prevent this kind of behavior:
```go
func (d *btrfs) defaultVMBlockFilesystemSize() string {
	v := d.state.OS.whatever // state could be nil when used from DefaultVMBlockFilesystemSize
	// ...
}
```

lxd/project/limits/permissions.go
MggMuggins: I don't see a function under `lxd/db` that would do that. I took a look through the database schema and [the docs](https://documentation.ubuntu.com/lxd/en/latest/reference/projects/#project-features) and there isn't a clear/direct relationship between projects and pools; this would likely mean going through all the devices in the project to find the pools in use.

I can implement a filter if you'd like but I don't think the complexity is worth it.

roosterfish: My thinking was if such a `projectInfo` struct gets initialized it's probably used with the intention that everything in there is part of the project. Same as for profiles, volumes and instances.
But as storage pools are shared across projects that kind of breaks this logic.

Maybe you can do the DB lookup in `getInstanceLimits` directly?

lxd/storage/drivers/load_test.go
roosterfish: Ah sorry then what I mentioned was an initial idea we didn't end up implementing.

So you could have a `DefaultVMBlockFilesystemSize()` on the driver `common` struct which returns the default `100MiB` and the PowerFlex driver can overwrite it and return the `8GiB` instead. So same as you did in the PR but with exported funcs without adding it to the `Info{}` struct so it can be called directly.

IIRC the comment back then was that this would return driver specific configuration which is usually served via the `Info{}` struct.

roosterfish: So is the problematic you are seeing that if you run `driverFunc().Info()` parts of the struct might not be set (like the Btrfs version you have linked)?

roosterfish: And therefore you have added the unexported `defaultVMBlockFilesystemSize` so that you can access the value directly without going through the `Info{}` struct?

MggMuggins: Yes, that's precisely it. `Info{}` mixes stateless and stateful information so it's not easy to get the stateless stuff that way without hauling in a lot of context. What do you think about splitting `Info` into stateless (no `Load` required) and stateful (`Load` required) components rather than exposing a special function for each bit of stateless info? I'm not sure what to call those functions/structs off the top of my head (`Info`/`InfoLoaded`? :grimacing:).

lxd/project/limits/permissions.go
MggMuggins: I'll give that a shot; it also saves us from doing an extra query that we only need in a special case.

lxd/storage/drivers/load_test.go
tomponline: I'm happy with having this DefaultVMBlockFilesystemSize function that uses the unexported function on the driver. Perhaps we can change this later to a special "Stateless" info struct, but no need right now.

lxd/project/limits/permissions.go
tomponline: We dont want to be doing any extra DB queries in `getInstanceLimits` as its run per instance in the project.

I dont see an issue with storing a pool name -> driver name map on `projectInfo` even if it does have pools in it that are not used by the project (a comment could help here).

Then we can add a specialised function that retrieves just the needed info in a single query.

MggMuggins:
@roosterfish @tomponline This is ready to go again

lxd/project/limits/permissions.go
```diff
@@ -1348,12 +1359,18 @@ func getInstanceLimits(instance api.Instance, keys []string, skipUnset bool) (ma
...
 				if !ok {
-					// TODO: In case the VMs storage drivers config drive size isn't the default,
-					// the limits accounting will be incorrect.
-					// This applies for the PowerFlex storage driver whose config drive size
-					// is 8 GB as set in the DefaultVMPowerFlexBlockFilesystemSize variable.
-					// See https://github.com/canonical/lxd/issues/12567
-					sizeStateValue = deviceconfig.DefaultVMBlockFilesystemSize
+					driverName := ""
+					poolName := device["pool"]
+					for pool, driver := range storagePoolDrivers {
```
roosterfish, line 1364: Do you need a loop or can you just do this:

```golang
driverName, ok := storagePoolDrivers[poolName]
```

roosterfish (dismissed):
LGTM!

lxd/project/limits/permissions.go
tomponline: Yeh that would be nicer.

Also what happens if the poolName isn't matched to a driver @MggMuggins ?

MggMuggins: Oops. Adjusted this for better error messages if the root device is missing a `pool` attribute (sounds like this [shouldn't happen](https://github.com/canonical/lxd/blob/main/lxd/instance/instancetype/instance.go#L32)) or the pool doesn't have a driver.

roosterfish: @MggMuggins yeah I think the check doesn't hurt as the error message is more usable than `No driver found for pool ""` in case the pool attribute isn't set in the device map.

roosterfish (approved):
Looks good to me.

tomponline (approved):

tomponline:
Thanks @MggMuggins please can you rebase to fix the conflict?

MggMuggins:
@tomponline Done

tomponline (approved):

PR #13998, opened 2024-08-28
lxd/apparmor: allow signals receiving

mihalicyn:
This is a workaround for AppArmor 4.0+ and new
unconfined profile feature. It was found [1]
that the new unconfined profile mode is not fully
permissive. Some things are forbidden for no reason.

We need this change to prevent breakage of LXD
when we eventually enable a new unconfined mode for lxd-support plug in snapd.

This change is absolutely safe in general and there is no reason to put it under "if" condition. [2]

[1] https://bugs.launchpad.net/apparmor/+bug/2077413
[2] https://gitlab.com/apparmor/apparmor/-/merge_requests/1310


tomponline (approved):
Thanks!

PR #14003, opened 2024-08-28
Make lxd-agent install script compatible with SUSE

### LINKED ISSUE #13991: lxd-agent not compatible with SLES 12/15 due path differences
On SLES 12 (SP5), systemd is installed in `/usr/lib/systemd` instead of the more common `/lib/systemd`. The `install.sh` script and the `lxd-agent.service` unit would need to detect this.

On SLES 15 (SP6), systemd is also installed in `/usr/lib/systemd` but `udev` rules are also not in the usual location but in `/usr/lib/udev/rules.d`.

simondeziel:
Fixes #13991

### DIFF
--- a/lxd/instance/drivers/driver_qemu.go
+++ b/lxd/instance/drivers/driver_qemu.go
@@ -2866,17 +2866,44 @@ if [ ! -d "/run/systemd/system/" ]; then
     exit 1
 fi
 
+for path in "/lib/systemd" "/usr/lib/systemd"; do
+    [ -d "${path}/system" ] || continue
+    LIB_SYSTEMD="${path}"
+    break
+done
+
+if [ ! -d "${LIB_SYSTEMD:-}" ]; then
+    echo "Could not find path to systemd"
+    exit 1
+fi
+
+for path in "/lib/udev" "/usr/lib/udev"; do
+    [ -d "${path}/rules.d/" ] || continue
+    LIB_UDEV="${path}"
+    break
+done
+
+if [ ! -d "${LIB_UDEV:-}" ]; then
+    echo "Could not find path to udev"
+    exit 1
+fi
+
 # Cleanup former units.
-rm -f /lib/systemd/system/lxd-agent-9p.service \
-    /lib/systemd/system/lxd-agent-virtiofs.service \
+rm -f "${LIB_SYSTEMD}/system/lxd-agent-9p.service" \
+    "${LIB_SYSTEMD}/system/lxd-agent-virtiofs.service" \
     /etc/systemd/system/multi-user.target.wants/lxd-agent-9p.service \
     /etc/systemd/system/multi-user.target.wants/lxd-agent-virtiofs.service \
     /etc/systemd/system/multi-user.target.wants/lxd-agent.service
 
 # Install the units.
-cp udev/99-lxd-agent.rules /lib/udev/rules.d/
-cp systemd/lxd-agent.service /lib/systemd/system/
-cp systemd/lxd-agent-setup /lib/systemd/
+cp udev/99-lxd-agent.rules "${LIB_UDEV}/rules.d/"
+cp systemd/lxd-agent-setup "${LIB_SYSTEMD}/"
+if [ "/lib/systemd" = "${LIB_SYSTEMD}" ]; then
+  cp systemd/lxd-agent.service "${LIB_SYSTEMD}/system/"
+else
+  # Adapt paths for systemd's lib location
+  sed "/=\/lib\/systemd/ s|=/lib/systemd|=${LIB_SYSTEMD}|" systemd/lxd-agent.service > "${LIB_SYSTEMD}/system/lxd-agent.service"
+fi
 systemctl daemon-reload
 
 # SELinux handling.
--- a/lxd/instance/drivers/driver_qemu.go
+++ b/lxd/instance/drivers/driver_qemu.go
@@ -2860,7 +2860,8 @@ if [ ! -e "systemd" ] || [ ! -e "lxd-agent" ]; then
     exit 1
 fi
 
-if [ ! -e "/lib/systemd/system" ]; then
+# systemd systems always have /run/systemd/system/ created on boot.
+if [ ! -d "/run/systemd/system/" ]; then
     echo "This script only works on systemd systems"
     exit 1
 fi
--- a/doc/guest-os-compatibility.md
+++ b/doc/guest-os-compatibility.md
@@ -50,6 +50,7 @@ not supported  | ❌
 ### LXD agent
 
 The LXD agent provides the ability to execute commands inside of the virtual machine guest without relying on traditional access solution like secure shell (SSH) or Remote Desktop Protocol (RDP). This agent is only supported on Linux guests using `systemd`.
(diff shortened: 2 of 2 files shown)


simondeziel:
This was tested on Ubuntu 24.04, OpenSUSE 15.6 and Tumbleweed. Note that Tumbleweed has a compat symlink in place (`/lib/ -> /usr/lib/`) so didn't need the fix anyway.

To test this, I used `ubuntu-minimal-daily:24.04`, `images:opensuse/15.6` and `images:opensuse/tumbleweed` VMs where:

1. setup SSH
2. removed every traces of the LXD agent
3. rebooted
4. confirmed agent not working
5. mounted the `virtiofs` config share onto `/mnt`
6. copied the `install.sh` script on the VM
7. mount bind the `install.sh` script onto `/mnt/install.sh`
8. manually installed the agent
9. rebooted
10. confirmed the agent worked

I also tested this on a manually installed VM running SUSE 15 SP6 (ISO install).

lxd/instance/drivers/driver_qemu.go
```diff
@@ -2860,7 +2860,7 @@ if [ ! -e "systemd" ] || [ ! -e "lxd-agent" ]; then
     exit 1
 fi
 
-if [ ! -e "/lib/systemd/system" ]; then
```
markylaing, line 2863: Could you please add a comment to explain what's happening (as in the commit message) as this is not obvious.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -2865,17 +2865,35 @@ if [ ! -d "/run/systemd/system/" ]; then
...
+    exit 1
+fi
+
+LIB_UDEV="/lib/udev"
+[ -d "${LIB_UDEV}/rules.d/" ] || LIB_UDEV="/usr${LIB_UDEV}"
+if [ ! -d "${LIB_UDEV}" ]; then
+    echo "Could not find path to udev"
+    exit 1
+fi
+
```
markylaing, line 2881: I found this a little hard to read. It's more verbose but an alternative might be:
```bash
LIB_SYSTEMD=""
for path in "/lib/systemd/system/" "/usr/lib/systemd"; do
    if [ -d "${path}" ]; then
        LIB_SYSTEMD="${path}"
    fi
done

if [ -z "${LIB_SYSTEMD}" ]; then
    echo "Could not find path to systemd"
    exit 1
fi
```

With that said, you mention that Tumbleweed has symlinks to the expected location. Should we just symlink here as well?

simondeziel: The thing is I'm checking for the presence of a child dir before concluding the parent dir is the right one. For the systemd case, I check for `${...}/system`'s existence before retaining the `${...}` part. I had started with a `for` loop be felt it was too verbose. I'll take another shot.

Re Tumbleweed having a compat symlink for `/lib -> /usr/lib`, that's a decision for a distro to make as the effects are far reaching.

simondeziel: @markylaing this is what I can come up with:
```
for path in "/lib/systemd" "/usr/lib/systemd"; do
    [ -d "${path}/system" ] || continue
    LIB_SYSTEMD="${path}"
    break
done

if [ ! -d "${LIB_SYSTEMD:-}" ]; then
    echo "Could not find path to systemd"
    exit 1
fi
```
Tell me if you think that reads better.

tomponline: yep looks clearer to me too

lxd/instance/drivers/driver_qemu.go
```diff
@@ -2866,17 +2866,43 @@ if [ ! -d "/run/systemd/system/" ]; then
...
 # Install the units.
-cp udev/99-lxd-agent.rules /lib/udev/rules.d/
-cp systemd/lxd-agent.service /lib/systemd/system/
-cp systemd/lxd-agent-setup /lib/systemd/
+cp udev/99-lxd-agent.rules "${LIB_UDEV}/rules.d/"
+cp systemd/lxd-agent-setup "${LIB_SYSTEMD}/"
+if [ "/lib/systemd" = "${LIB_SYSTEM}" ]; then
+  cp systemd/lxd-agent.service "${LIB_SYSTEMD}/system/"
+else
+  sed "/=\/lib\/systemd/ s|=/lib/systemd|=${LIB_SYSTEMD}|" systemd/lxd-agent.service > "${LIB_SYSTEMD}/system/lxd-agent.service"
```
tomponline, line 2904: what is this part for?

lxd/instance/drivers/driver_qemu.go
```diff
@@ -2866,17 +2866,43 @@ if [ ! -d "/run/systemd/system/" ]; then
...
     /etc/systemd/system/multi-user.target.wants/lxd-agent-virtiofs.service \
     /etc/systemd/system/multi-user.target.wants/lxd-agent.service
 
 # Install the units.
-cp udev/99-lxd-agent.rules /lib/udev/rules.d/
-cp systemd/lxd-agent.service /lib/systemd/system/
-cp systemd/lxd-agent-setup /lib/systemd/
+cp udev/99-lxd-agent.rules "${LIB_UDEV}/rules.d/"
+cp systemd/lxd-agent-setup "${LIB_SYSTEMD}/"
+if [ "/lib/systemd" = "${LIB_SYSTEM}" ]; then
```
markylaing, line 2901: Is there a "D" missing here? e.g. `LIB_SYSTEMD`?

lxd/instance/drivers/driver_qemu.go
simondeziel: I put a comment that we need to adapt the path when systemd is installed in the unusual location.

lxd/instance/drivers/driver_qemu.go
simondeziel: Thanks!

simondeziel:
@tomponline the CI fails for what seems to be a regression with microceph as it's consistently failing in this unrelated spot: https://github.com/canonical/lxd/actions/runs/10685425570/job/29621723179#step:12:50510

lxd/instance/drivers/driver_qemu.go
tomponline: Thanks, what is it actually doing though?

tomponline (approved):

lxd/instance/drivers/driver_qemu.go
simondeziel: Replacing `=/lib/systemd` by `=${LIB_SYSTEMD}` only if both are different.

simondeziel: If `LIB_SYSTEMD` == `/lib/systemd`, `cp` is used. Otherwise, `sed` will take the `lxd-agent.service` file from the config and return a mangled version to stdout which is then written into the proper location (also based on `${LIB_SYSTEMD}`).

This way, `sed` is only used/needed if the unit needs mangling and there is no "in-place" rewrite as a `cp + sed -i` would have required.

tomponline: Thanks!

PR #14012, opened 2024-08-29
Makefile: pin gorilla/websocket version (stable-5.0)

simondeziel:
This avoids crashes in LP on riscv64


tomponline:
@simondeziel can we have this for main and stable-5.21 if not there already, thanks

tomponline (approved):

simondeziel:
`main` and `stable-5.21` have it already but I didn't cherry-pick due to the commit in question also pinning `openfga`.

PR #14021, opened 2024-08-30
lxd/instances: Don't bypass instance limit check (from Incus)

MggMuggins:
User reported in https://github.com/lxc/incus/issues/463 that project limits were not being enforced on the cluster member in direct communication with the client during instance creation.

~I also noticed that the limit checks were being skipped based on the value of the [`User-Agent` header](https://github.com/canonical/lxd/commit/d6871e49e4ae7c1ee173c49ca08e350e11f45b73), which is spoofable. This pushes the limits check back to the target cluster member.~

Contains cherry-picks from https://github.com/lxc/incus/pull/505

### DIFF
--- a/lxd/instances_post.go
+++ b/lxd/instances_post.go
@@ -1238,8 +1238,6 @@ func instancesPost(d *Daemon, r *http.Request) response.Response {
 			if err != nil {
 				return err
 			}
-
-			return nil
 		}
 
 		if !clusterNotification {


lxd/instances_post.go
```diff
@@ -1240,15 +1240,6 @@ func instancesPost(d *Daemon, r *http.Request) response.Response {
 			}
 		}
 
-		if !clusterNotification {
```
tomponline, line 1245: i dont think we want this commit, as we do not want to re-run the creation checks again during a redirection.

However please can you check whether `clusterNotification` is forgeable by a non-admin client?

MggMuggins: It's not forgeable; [client authentication will fail](https://github.com/canonical/lxd/blob/1e0b3535f2c0e0559205b019831bea7af1af21fb/lxd/daemon.go#L396) if the user passes `User-Agent: lxd-cluster-notifier`:
```
$ curl --cert client.crt --key client.key --insecure -H 'User-Agent: lxd-cluster-notifier' https://canaveral.cld.lan:8443/1.0
{"type":"error","status":"","status_code":0,"operation":"","error_code":403,"error":"Cluster notification isn't using trusted server certificate","metadata":null}
```
Dropping this commit.

tomponline (approved):

PR #14022, opened 2024-08-30
lxd/project: Allow managed volumes under "allow" (from Incus)

MggMuggins:
See [the docs](https://documentation.ubuntu.com/lxd/en/latest/reference/projects/#project-restricted:restricted.devices.disk) for disk device restrictions. This PR prevents checking the `restricted.devices.disk.paths` for managed devices when `restricted.devices.disk: allow`.

Contains cherry-picks from https://github.com/lxc/incus/pull/708


tomponline:
We should add a test or 2 for this

tomponline (approved):

PR #14024, opened 2024-08-30
Use `errors.New()` where appropriate

simondeziel:


simondeziel:
I had to back down on the `fmt.Println(...)` changes as those are not always 1:1 replacement for `fmt.Printf("%s\n", ...)`.

MusicDin (approved):

PR #14026, opened 2024-09-02
Tests: Check cluster member names cannot be 'none' (stable-5.21)

roosterfish:
As we have both the password and token support in `5.21`, we can completely backport https://github.com/canonical/lxd/commit/c49ef5f1c599c9c6883b20527387a7e8057b8155.

An additional commit ensures that also when joining cluster nodes using the legacy password (which is still in `5.21`) the name cannot be `none`.

This change works with the LXD snap from `5.21/edge`.


tomponline (approved):

tomponline:
Ta!

PR #14031, opened 2024-09-02
lxd/device/gpu_physical: forbid CDI GPU hotplugging

gabrielmougard:
CDI GPU are not hotpluggable because the configuration of a CDI GPU requires a LXC hook that is only run at instance start. A classic GPU device can be hotplugged.


tomponline (approved):

tomponline:
> A classic GPU device can be hotplugged.

Not strictly 100% true. As even classic GPUs have the same problem, in that if hotplugged, only the device itself and not the supporting libraries will be attached.

PR #14035, opened 2024-09-03
Add workflow to run e2e tests from lxd-ui

edlerd:
Added a step to the GitHub test workflow, executing the e2e test suite from lxd-ui with a lxd backend built from the current branch from a PR or main branch.

We might extend this on the stable-5.21 and stable-5.0 branches, as we have dedicated test suites in lxd-ui for those. For older versions, we should continue skipping the e2e tests. I'd keep this as a second step in a followup PR.

This was successful on a fork, i.e. [see this run](https://github.com/edlerd/lxd/actions/runs/10686642233/job/29622381613).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -474,6 +474,227 @@ jobs:
           name: ${{ runner.os }}
           path: bin/
 
+  ui:
+    name: UI e2e tests
+    runs-on: ubuntu-latest
+    if: "!startsWith(github.ref, 'refs/heads/stable-')"
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Performance tuning
+        run: |
+          set -eux
+          # optimize ext4 FSes for performance, not reliability
+          for fs in $(findmnt --noheading --type ext4 --list --uniq | awk '{print $1}'); do
+            # nombcache and data=writeback cannot be changed on remount
+            sudo mount -o remount,noatime,barrier=0,commit=6000 "${fs}" || true
+          done
+
+          # disable dpkg from calling sync()
+          echo "force-unsafe-io" | sudo tee /etc/dpkg/dpkg.cfg.d/force-unsafe-io
+
+      - name: Install Go (1.22)
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.22.x
+
+      - name: Install dependencies
+        run: |
+          set -eux
+          sudo add-apt-repository ppa:ubuntu-lxc/daily -y --no-update
+          sudo add-apt-repository ppa:dqlite/dev -y --no-update
+          sudo apt-get update
+
+          sudo systemctl mask lxc.service lxc-net.service
+
+          sudo apt-get install --no-install-recommends -y \
+            curl \
+            git \
+            libacl1-dev \
+            libcap-dev \
+            libdbus-1-dev \
+            libdqlite-dev \
+            liblxc-dev \
+            libseccomp-dev \
+            libselinux-dev \
+            libsqlite3-dev \
+            libtool \
+            libudev-dev \
+            make \
+            pkg-config\
+            acl \
+            attr \
+            bind9-dnsutils \
+            btrfs-progs \
+            busybox-static \
+            dnsmasq-base \
+            easy-rsa \
+            gettext \
+            jq \
+            lxc-utils \
+            lvm2 \
+            nftables \
+            quota \
+            rsync \
+            s3cmd \
+            socat \
+            sqlite3 \
+            squashfs-tools \
+            tar \
+            tcl \
+            thin-provisioning-tools \
+            uuid-runtime \
+            xfsprogs \
+            xz-utils \
+            zfsutils-linux
+
+          mkdir -p "$(go env GOPATH)/bin"
+          curl -sSfL https://dl.min.io/server/minio/release/linux-amd64/minio --output "$(go env GOPATH)/bin/minio"
+          chmod +x "$(go env GOPATH)/bin/minio"
+
+          # Also grab the latest minio client to maintain compatibility with the server.
+          curl -sSfL https://dl.min.io/client/mc/release/linux-amd64/mc --output "$(go env GOPATH)/bin/mc"
+          chmod +x "$(go env GOPATH)/bin/mc"
+
+      - name: Download go dependencies
+        run: |
+          set -eux
+          go mod download
+
+      - name: Run LXD build
+        run: |
+          set -eux
+          make lxd
+
+      - name: Install doc dependencies
+        run: |
+          set -eux
+          sudo apt-get install aspell aspell-en
+          sudo snap install mdl
+
+      - name: Build docs (for the objects.inv.txt file to be available for the e2e tests)
+        shell: 'script -q -e -c "export TERM=xterm-256color; bash {0}"'
+        run: |
+          set -eux
+          make doc
+
+      - name: Run LXD daemon
+        run: |
+          set -eux
+          echo "root:1000000:1000000000" | sudo tee /etc/subuid /etc/subgid
+          export CGO_CFLAGS="-I/home/runner/work/lxd/lxd-test/vendor/dqlite/include/"
+          export CGO_LDFLAGS="-L/home/runner/work/lxd/lxd-test/vendor/dqlite/.libs/"
+          export LD_LIBRARY_PATH="/home/runner/work/lxd/lxd-test/vendor/dqlite/.libs/"
+          export LXD_DOCUMENTATION="/home/runner/work/lxd/lxd/doc/_build/"
+          sudo rm -rf /var/lib/lxd
+          sudo -E PATH=${PATH} LD_LIBRARY_PATH=${LD_LIBRARY_PATH} $(go env GOPATH)/bin/lxd --group sudo &
+
+      - name: Setup LXD
+        shell: bash
+        run: |
+          set -eux
+          sudo LXD_DIR=/var/lib/lxd lxc storage create local-storage zfs
+          sudo LXD_DIR=/var/lib/lxd lxc profile device add default root disk path=/ pool=local-storage
+          sudo LXD_DIR=/var/lib/lxd lxc network create local-network
+          sudo LXD_DIR=/var/lib/lxd lxc profile device add default eth0 nic network=local-network
+          sudo LXD_DIR=/var/lib/lxd lxc config set core.https_address "[::]:8443"
+          sudo LXD_DIR=/var/lib/lxd lxc config set cluster.https_address "127.0.0.1"
+          sudo LXD_DIR=/var/lib/lxd lxc cluster enable local
+          sudo LXD_DIR=/var/lib/lxd lxc config set user.show_permissions=true
+
+      - name: Checkout LXD-UI
+        uses: actions/checkout@v4
+        with:
+            repository: 'canonical/lxd-ui'
+            ref: main
+            path: lxd-ui
+
+      - name: Install Dotrun
+        run: sudo pip3 install dotrun
+
+      - name: Restore cached keys
+        uses: actions/cache/restore@v3
+        with:
+          path: lxd-ui/keys
+          key: keys-folder
+
+      - name: Install LXD-UI dependencies
+        run: |
+          set -eux
+          sudo chmod 0777 ./lxd-ui
+          cd lxd-ui
+          dotrun install
+
+      - name: Run LXD-UI
+        env:
+          ENVIRONMENT: devel
+          PORT: 8407
+          LXD_UI_BACKEND_IP: 172.17.0.1
+        run: |
+          set -eux
+          cd lxd-ui
+          dotrun &
+          curl --head --fail --retry-delay 2 --retry 100 --retry-connrefused --insecure https://localhost:8407
+
+      - name: Set keys permissions
+        run: |
+          set -eux
+          sudo chmod -R 0666 lxd-ui/keys
+          sudo chmod 0777 lxd-ui/keys
+
+      - name: Save keys
+        uses: actions/cache/save@v3
+        with:
+          path: lxd-ui/keys
+          key: keys-folder
+
+      - name: Allow LXD-UI keys
+        shell: bash
+        run: |
+          set -eux
+          sudo LXD_DIR=/var/lib/lxd lxc config trust add lxd-ui/keys/lxd-ui.crt
+
+      - name: Create a custom image
+        shell: bash
+        run: |
+          set -eux
+          sudo LXD_DIR=/var/lib/lxd lxc launch ubuntu-minimal:22.04 my-instance
+          sudo LXD_DIR=/var/lib/lxd lxc publish my-instance --alias my-custom-image --force
+          sudo LXD_DIR=/var/lib/lxd lxc delete my-instance --force
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Install Playwright Browsers
+        run: |
+          set -eux
+          cd lxd-ui
+          npx playwright install --with-deps chromium
+
+      - name: Create OIDC users
+        shell: bash
+        run: |
+          set -eux
+          sudo LXD_DIR=/var/lib/lxd PATH=/home/runner/go/bin:$PATH ./lxd-ui/tests/scripts/create_oidc_identities
+
+      - name: Run Playwright tests
+        run: |
+          set -eux
+          cd lxd-ui
+          CI=true DISABLE_VM_TESTS=true npx playwright test --project chromium:lxd-latest-edge
+          #npx playwright test --project chromium:lxd-5.21-edge
+          #npx playwright test --project lxd-5.0-edge
+
+      - name: Upload lxd-ui test artifacts
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: lxd-ui-test-report
+          path: lxd-ui/blob-report
+
   documentation:
     name: Documentation
     runs-on: ubuntu-latest


tomponline (changes requested):
Hi @edlerd please can you work with @simondeziel to update this to use the pre-built LXD binaries we now have (that are compiled during the Code Tests phase) and to use the re-usable workflows for performance tuning and dependency installation.

Also i'd like to avoid building the docs twice, but they are built into the tarball i believe during the Code Tests phase so may be available already now.

.github/workflows/tests.yml
```diff
@@ -474,6 +474,227 @@ jobs:
...
+
+      - name: Run Playwright tests
+        run: |
+          set -eux
+          cd lxd-ui
+          CI=true DISABLE_VM_TESTS=true npx playwright test --project chromium:lxd-latest-edge
+          #npx playwright test --project chromium:lxd-5.21-edge
+          #npx playwright test --project lxd-5.0-edge
+
+      - name: Upload lxd-ui test artifacts
```
tomponline, line 691: lets set a short artifact lifetime on these please

edlerd: Good idea, set this to 14 days.

edlerd:
@simondeziel @tomponline Updated the workflow to consume the build artefacts from previous steps. Please give it another review.

edlerd:
Also using the docs artefact by depending and downloading on that build step.

.github/workflows/tests.yml
```diff
@@ -494,6 +494,240 @@ jobs:
...
           path: bin/
 
+  ui:
+    name: UI e2e tests
+    runs-on: ubuntu-latest
+    needs: [code-tests, documentation]
+    if: "!startsWith(github.ref, 'refs/heads/stable-')"
+    env:
+      LXD_DIR: "/var/lib/lxd"
+      LXD_OIDC_CLIENT_ID: "gxj297yfAjmklILK5WqPWDSbtVBAeSQm"
```
tomponline, line 504: Is this a secret key?

.github/workflows/tests.yml
```diff
@@ -494,6 +494,240 @@ jobs:
...
+      LXD_OIDC_ISSUER: "https://dev-xjrvvfikbsv4jxn7.us.auth0.com/"
+      LXD_OIDC_AUDIENCE: "https://dev-xjrvvfikbsv4jxn7.us.auth0.com/api/v2/"
+      LXD_OIDC_USER: "lxd-ui-e2e-tests@example.org"
+      LXD_OIDC_PASSWORD: "lxd-ui-e2e-password"
```
tomponline, line 508: This shouldnt be hardcoded into the test file, as it references an external service, and thus is a real password.

We should now change this password to something else and then securely transfer it to me so we can set it up as a GH secret.

.github/workflows/tests.yml
```diff
@@ -494,6 +494,240 @@ jobs:
...
+          # optimize ext4 FSes for performance, not reliability
+          for fs in $(findmnt --noheading --type ext4 --list --uniq | awk '{print $1}'); do
+            # nombcache and data=writeback cannot be changed on remount
+            sudo mount -o remount,noatime,barrier=0,commit=6000 "${fs}" || true
+          done
+
+          # disable dpkg from calling sync()
+          echo "force-unsafe-io" | sudo tee /etc/dpkg/dpkg.cfg.d/force-unsafe-io
+
+      - name: Install Go (1.22)
```
tomponline, line 526: please can you copy how we setup Go for the other steps, as we are now using the go mod version automatically to reduce maintenance.

.github/workflows/tests.yml
```diff
@@ -494,6 +494,240 @@ jobs:
...
+      LXD_OIDC_ISSUER: "https://dev-xjrvvfikbsv4jxn7.us.auth0.com/"
+      LXD_OIDC_AUDIENCE: "https://dev-xjrvvfikbsv4jxn7.us.auth0.com/api/v2/"
+      LXD_OIDC_USER: "lxd-ui-e2e-tests@example.org"
+      LXD_OIDC_PASSWORD: "lxd-ui-e2e-password"
+      LXD_OIDC_GROUPS_CLAIM: "lxd-idp-groups"
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Performance tuning
```
tomponline, line 514: can you use the workflows we have for this?

.github/workflows/tests.yml
edlerd: This is a dedicated free account, specially created for e2e tests. It is expected to be public. I don't see a reason for this to be kept secret.

Though once we start using the local OIDC service, we can remove this.

.github/workflows/tests.yml
```diff
@@ -494,6 +494,240 @@ jobs:
...
 
+  ui:
+    name: UI e2e tests
+    runs-on: ubuntu-latest
+    needs: [code-tests, documentation]
+    if: "!startsWith(github.ref, 'refs/heads/stable-')"
+    env:
+      LXD_DIR: "/var/lib/lxd"
+      LXD_OIDC_CLIENT_ID: "gxj297yfAjmklILK5WqPWDSbtVBAeSQm"
+      LXD_OIDC_ISSUER: "https://dev-xjrvvfikbsv4jxn7.us.auth0.com/"
```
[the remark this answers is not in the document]
edlerd: That is a good hint, I'll check how to adopt it.

.github/workflows/tests.yml
tomponline: Hrm, yeah, but we shouldnt be making the credentials public

.github/workflows/tests.yml
edlerd: Same as the [password over there](https://github.com/canonical/lxd/pull/14035#discussion_r1841906916)

.github/workflows/tests.yml
```diff
@@ -494,6 +494,240 @@ jobs:
           name: lxd-clients-${{ runner.os }}
           path: bin/
 
+  ui:
```
tomponline, line 497: please can you move this as part of the `system-tests:` workflow and make it a step after `name: "Run system tests (${{ matrix.suite }}, ${{ matrix.backend }})"` and before `name: Upload coverage data`

Then we can re-use the dependency install deps, and by ensuring LXD is run with the code coverage env var, we can collect coverage data from the UI tests too.

.github/workflows/tests.yml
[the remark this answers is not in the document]
edlerd: Using the mini OIDC implementation is not trivial. It is only implementing the device flow, which is incompatible with the web flow.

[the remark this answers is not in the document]
tomponline: OK so we can stick with the external service, but all identifiers should be GH secrets.

.github/workflows/tests.yml
```diff
@@ -494,6 +494,231 @@ jobs:
...
+      - name: Run LXD-UI
+        env:
+          ENVIRONMENT: devel
+          PORT: 8407
+          LXD_UI_BACKEND_IP: 172.17.0.1
+        run: |
+          set -eux
+          cd lxd-ui
+          dotrun &
+          curl --head --fail --retry-delay 2 --retry 100 --retry-connrefused --insecure https://localhost:8407
```
simondeziel, line 658:
```suggestion
          curl --head --fail --retry-delay 2 --retry 100 --retry-connrefused --insecure https://localhost:${{ env.PORT }}
```

.github/workflows/tests.yml
```diff
@@ -494,6 +494,231 @@ jobs:
...
+          set -eux
+          sudo chmod 0777 ./lxd-ui
+          cd lxd-ui
+          dotrun install
+
+      - name: Run LXD-UI
+        env:
+          ENVIRONMENT: devel
+          PORT: 8407
+          LXD_UI_BACKEND_IP: 172.17.0.1
```
simondeziel, line 653: Could that be one of the loopback IPs instead?

.github/workflows/tests.yml
```diff
@@ -494,6 +494,231 @@ jobs:
...
+        uses: actions/checkout@v4
+        with:
+          repository: 'canonical/lxd-ui'
+          ref: main
+          path: lxd-ui
+
+      - name: Install Dotrun
+        run: sudo pip3 install dotrun
+
+      - name: Restore cached keys
```
simondeziel, line 636: Could we have those generated on the fly please? You can potentially reuse `test/includes/certificates.sh` that has a handy function called `gen_cert_and_key`.

.github/workflows/tests.yml
```diff
@@ -494,6 +494,231 @@ jobs:
...
+      LXD_DIR: "/var/lib/lxd"
+      LXD_OIDC_CLIENT_ID: "gxj297yfAjmklILK5WqPWDSbtVBAeSQm"
+      LXD_OIDC_ISSUER: "https://dev-xjrvvfikbsv4jxn7.us.auth0.com/"
+      LXD_OIDC_AUDIENCE: "https://dev-xjrvvfikbsv4jxn7.us.auth0.com/api/v2/"
+      LXD_OIDC_USER: "lxd-ui-e2e-tests@example.org"
+      LXD_OIDC_PASSWORD: "lxd-ui-e2e-password"
+      LXD_OIDC_GROUPS_CLAIM: "lxd-idp-groups"
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
simondeziel, line 512: Please use SHA pinning for external actions.

.github/workflows/tests.yml
```diff
@@ -494,6 +494,231 @@ jobs:
...
+          set -eux
+          export PATH="/home/runner/go/bin:$PATH"
+          sudo -E LXD_DIR=/var/lib/lxd lxc config trust add lxd-ui/keys/lxd-ui.crt
+
+      - name: Create a custom image
+        shell: bash
+        run: |
+          set -eux
+          export PATH="/home/runner/go/bin:$PATH"
+          sudo -E LXD_DIR=/var/lib/lxd lxc launch ubuntu-minimal:22.04 my-instance
```
simondeziel, line 684: No need to launch, init will do. Also, we tend to use `minimal-daily` elsewhere:
```suggestion
          sudo -E LXD_DIR=/var/lib/lxd lxc init ubuntu-minimal-daily:22.04 my-instance
```

But if you don't actually need a full fledged instance, we have a simple busybox `testimage`. See `test/includes/setup.sh`.

.github/workflows/tests.yml
edlerd: Generation of the keys is taking ~1 minute and burning cpu cycles on each ci run seems easy to resolve, hence we introduced this cache a while ago in the lxd-ui pipeline. It is just a few kb and I think the benefits outweigh here. wdyt?

.github/workflows/tests.yml
edlerd: The ui is is running in a container, so we need to use this to reach out from the container that is running the ui. Loopback would be inside the container and not reach lxd which runs on the host.

.github/workflows/tests.yml
edlerd: This is a workaround to generate an image from an instance, before this was not possible in the UI. I think by now we can generate the image directly in our e2e test. So we might be able to get rid of this preparation step completely. I need to double-check if we can evolve our test, but I am optimistic it will work. Will keep this open for now.

.github/workflows/tests.yml
edlerd: Good idea.

.github/workflows/tests.yml
edlerd: This will slow down LXD test runtime, currently it runs in parallel to the system tests, which will not affect your total runtime. Because system tests take longer than the UI tests. I think a better option is to also record coverage in the UI step and download both coverage reports in the TICS step.

.github/workflows/tests.yml
simondeziel: Hmm, maybe that was due to RSA being slow as with our function, this should be even faster than fetching from the cache:

```
$ . test/includes/certificates.sh
$ export LXD_CONF=$(mktemp -d)
$ time gen_cert_and_key lxd-ui
-----

real	0m0.007s
user	0m0.003s
sys	0m0.004s
```

So hard to beat :)

And you get such cert/key:

```
$ openssl x509 -noout -text -in $LXD_CONF/lxd-ui.crt
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            0b:d3:1c:7f:64:54:df:b9:34:af:1f:cc:40:14:bd:8b:b2:92:03:0f
        Signature Algorithm: ecdsa-with-SHA384
        Issuer: CN = lxd-ui.local
        Validity
            Not Before: Nov 14 23:13:44 2024 GMT
            Not After : Nov 15 23:13:44 2024 GMT
        Subject: CN = lxd-ui.local
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (384 bit)
                pub:
                    04:bd:8c:4e:94:7d:bb:66:de:f7:52:a8:05:f8:19:
                    d4:32:0b:4c:8a:b2:60:42:77:27:33:7e:74:cb:af:
                    39:66:70:3c:d9:f6:79:26:c8:5c:43:cf:1e:b9:e5:
                    ee:02:42:55:ec:04:15:c8:4f:9a:93:ad:69:9c:87:
                    dd:95:c4:52:69:66:93:22:5a:e0:77:65:51:2e:fd:
                    76:21:28:87:23:82:8d:55:f6:8d:1e:2c:54:0e:1a:
                    98:1f:8f:4f:fd:44:98
                ASN1 OID: secp384r1
                NIST CURVE: P-384
        X509v3 extensions:
            X509v3 Subject Key Identifier: 
                C8:FC:A2:AC:F1:1C:85:0E:E8:7C:FD:A3:A3:7B:27:AE:D0:8C:54:BB
            X509v3 Authority Key Identifier: 
                C8:FC:A2:AC:F1:1C:85:0E:E8:7C:FD:A3:A3:7B:27:AE:D0:8C:54:BB
            X509v3 Basic Constraints: critical
                CA:TRUE
    Signature Algorithm: ecdsa-with-SHA384
    Signature Value:
        30:65:02:31:00:82:02:fc:69:6f:ec:88:f6:9f:53:94:b2:68:
        76:1c:b6:f4:33:7a:0f:62:49:df:7e:95:a7:37:b5:9a:df:78:
        08:a2:14:d4:57:4e:7e:ec:db:75:eb:7c:5d:c3:75:ac:cf:02:
        30:4f:b3:c5:22:63:a3:6b:c6:00:7c:98:80:cd:91:1e:62:11:
        5a:2c:e0:72:d0:6c:db:d7:d4:93:85:79:0c:8b:02:5f:0c:2b:
        bc:47:ab:06:b0:c3:d4:64:1d:ad:76:20:9f
```

edlerd: Oh, this is very nice. I will definitely use it instead, and we don't need the cache. Thanks for the hint :+1:

.github/workflows/tests.yml
edlerd: As per the [okta docs](https://www.oauth.com/oauth2-servers/client-registration/client-id-secret/), the client id is public.

.github/workflows/tests.yml
edlerd: removed this bit after adjusting the ui tests.

.github/workflows/tests.yml
edlerd: Created a new set of credentials and shared them to be setup as a repo secret. Once they are available, we can start using them here instead of the old ones.

tomponline: Thanks will set them up now.

As the creds will only be available after a PR is merged, please can you add a condition that this UI test only runs on push rather than pull requests. This way it wont cause our per-PR tests to take longer, but will still gate our edge builds.

.github/workflows/tests.yml
tomponline: You can still run the test itself concurrently, as long as it happens after the LXD binary is built with code coverage, and before the code coverage collection. This is what we do with the system tests.

.github/workflows/tests.yml
tomponline: Do we need to run the UI in a container?

.github/workflows/tests.yml
edlerd: Adjusted the flow to rely on secrets.

.github/workflows/tests.yml
edlerd: The container is the status quo of the dev environment and the CI for the LXD-UI. It is solving a couple of things, like running a web server for the UI, permission bootstrapping and using HAProxy to wire things together with the backend while not breaking CORS. Yes, the container is not strictly needed here. But we would create an exception for the LXD workflow to build and run the UI differently. I think it is simpler to keep dev environment, UI CI and LXD CI aligned. Especially for future maintenance, it seems beneficial to keep full control over the wiring and setup in the UI with the container.

.github/workflows/tests.yml
```diff
@@ -494,6 +494,211 @@ jobs:
...
+            xz-utils \
+            zfsutils-linux
+
+          mkdir -p "$(go env GOPATH)/bin"
+          curl -sSfL https://dl.min.io/server/minio/release/linux-amd64/minio --output "$(go env GOPATH)/bin/minio"
+          chmod +x "$(go env GOPATH)/bin/minio"
+
+          # Also grab the latest minio client to maintain compatibility with the server.
+          curl -sSfL https://dl.min.io/client/mc/release/linux-amd64/mc --output "$(go env GOPATH)/bin/mc"
+          chmod +x "$(go env GOPATH)/bin/mc"
```
simondeziel, line 578: This is now a dedicated action: `.github/actions/download-minio/action.yml`

.github/workflows/tests.yml
```diff
@@ -494,6 +494,211 @@ jobs:
...
+
+      - name: Tune disk performance
+        uses: ./.github/actions/tune-disk-performance
+
+      - name: Install Go
+        uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
+        with:
+          go-version-file: 'go.mod'
+
+      - name: Install dependencies
```
simondeziel, line 522: I think most if not all of it can be replaced by `.github/actions/install-lxd-builddeps`.

.github/workflows/tests.yml
```diff
@@ -494,6 +494,211 @@ jobs:
...
+      - name: Install Dotrun
+        run: sudo pip3 install dotrun
+
+      - name: Generate keys for the ui to use
+        run: |
+          . test/includes/certificates.sh
+          export LXD_CONF=lxd-ui/keys
+          mkdir -p $LXD_CONF
+          gen_cert_and_key lxd-ui
+          openssl dhparam -out lxd-ui/keys/dhparams.pem 2048
```
simondeziel, line 642: I'm curious to know why/where DH params are needed?

.github/workflows/tests.yml
```diff
@@ -494,6 +494,211 @@ jobs:
...
+          cat lxd-ui/keys/lxd-ui.key lxd-ui/keys/lxd-ui.crt > lxd-ui/keys/lxd-ui.pem
+          cp lxd-ui/keys/lxd-ui.key lxd-ui/keys/lxd-ui.crt.key
```
simondeziel, line 644: What's that `.crt.key` file for? I find the extension a bit misleading as it contains just the private key, isn't it?

.github/workflows/tests.yml
```diff
@@ -494,6 +494,211 @@ jobs:
...
+        run: sudo pip3 install dotrun
+
+      - name: Generate keys for the ui to use
+        run: |
+          . test/includes/certificates.sh
+          export LXD_CONF=lxd-ui/keys
+          mkdir -p $LXD_CONF
+          gen_cert_and_key lxd-ui
+          openssl dhparam -out lxd-ui/keys/dhparams.pem 2048
+          cat lxd-ui/keys/lxd-ui.key lxd-ui/keys/lxd-ui.crt > lxd-ui/keys/lxd-ui.pem
```
simondeziel, line 643: Is that combined `crt`+`key` `PEM` used by test UI test tooling?

.github/workflows/tests.yml
```diff
@@ -494,6 +494,211 @@ jobs:
...
+          set -eux
+          export PATH="/home/runner/go/bin:$PATH"
+          sudo -E LXD_DIR=/var/lib/lxd lxc storage create default zfs
+          sudo -E LXD_DIR=/var/lib/lxd lxc profile device add default root disk path=/ pool=default
+          sudo -E LXD_DIR=/var/lib/lxd lxc network create local-network
+          sudo -E LXD_DIR=/var/lib/lxd lxc profile device add default eth0 nic network=local-network
+          sudo -E LXD_DIR=/var/lib/lxd lxc config set core.https_address "[::]:8443"
+          sudo -E LXD_DIR=/var/lib/lxd lxc config set cluster.https_address "127.0.0.1"
+          sudo -E LXD_DIR=/var/lib/lxd lxc cluster enable local
+          sudo -E LXD_DIR=/var/lib/lxd lxc config set user.show_permissions=true
```
simondeziel, line 624: It's the first time I see this key used, is `user.show_permissions` some special key for the LXD-UI?

edlerd: Yes, we use this as a flag to show the permissions tab in the ui even when oidc is not configured.

.github/workflows/tests.yml
edlerd: The params file is used for the [HAProxy config](https://github.com/canonical/lxd-ui/blob/main/haproxy-dev.cfg#L3). See the [other comment on why we need HAProxy](https://github.com/canonical/lxd/pull/14035#discussion_r1846209294).

.github/workflows/tests.yml
edlerd: I think this is also needed for HAProxy, suspecting a convention for it to be named like that. I might be wrong, but don't have the time now to do an experiment of removing it, then rerunning the workflow and dev server and demo server and see if anything fails.

.github/workflows/tests.yml
edlerd: good idea.

.github/workflows/tests.yml
edlerd: good idea :)

.github/workflows/tests.yml
edlerd: Edit: Meant "show the permissions navigation entry", it is not a tab.

.github/workflows/tests.yml
edlerd: removed the explicit key generation from this workflow.

.github/workflows/tests.yml
```diff
@@ -486,6 +486,194 @@ jobs:
...
+
+      - name: Tune disk performance
+        uses: ./.github/actions/tune-disk-performance
+
+      - name: Install Go
+        uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
+        with:
+          go-version-file: 'go.mod'
+
+      - name: Install dependencies for running lxd
```
simondeziel, line 514: I've just reworked the actions so you can now replace that bit with:
```
      - name: Install runtime dependencies
        uses: ./.github/actions/install-lxd-runtimedeps
```

.github/workflows/tests.yml
edlerd: Adjusted the workflow, so the UI tests reuse the binaries and also upload coverage with a matching prefix. They now run in parallel to the system tests, and block the TICS run, which consumes the coverage from both.

image

tomponline: Excellent!

tomponline:
Please can this be rebased and squashed into fewer commits (perhaps only 1, depending on what you see as the logical changes in this PR), rather than each commit showing the evolution of the PR over the review. Thanks

edlerd:
> Please can this be rebased and squashed into fewer commits (perhaps only 1, depending on what you see as the logical changes in this PR), rather than each commit showing the evolution of the PR over the review. Thanks

Will do this after [testing this on a fork](https://github.com/edlerd/lxd/actions/runs/11957944302) shows a successful run.

.github/workflows/tests.yml
```diff
@@ -445,6 +445,159 @@ jobs:
...
+      - name: Run Playwright tests
+        run: |
+          set -eux
+          cd lxd-ui
+          sudo chown $USER -R /home/runner/.config
+          sudo chown $USER /var/lib/lxd/unix.socket
+          CI=true DISABLE_VM_TESTS=true LXD_DIR=/var/lib/lxd PATH=/home/runner/go/bin:$PATH npx playwright test --project chromium:lxd-latest-edge
+          #npx playwright test --project chromium:lxd-5.21-edge
+          #npx playwright test --project lxd-5.0-edge
+
```
tomponline, line 585: We should probably have a step here to cleanly stop LXD via `lxd shutdown` so we can be sure all coverage data is written before sending it up.

.github/workflows/tests.yml
edlerd: I am adding some old dependencies for dqlite back to encounter this issue:

image

simondeziel: That `libdqlite.so.0` should be coming from the system-test-deps artifact, no?

.github/workflows/tests.yml
edlerd: I added the `lxd shutdown`, but the coverage data for the ui is empty on [this run](https://github.com/edlerd/lxd/actions/runs/11958693042/job/33339373448) on my fork. I set up the secrets over there and the e2e test suite completed just fine. Any idea what could be missing is welcome.

.github/workflows/tests.yml
edlerd: All I know is with [this changeset](https://github.com/canonical/lxd/compare/58ba9f6f9d15574dc6d2f683a141711504b03e4d..5fdd983438347b63cb4077080ba4c64548a8e8a5) the UI e2e test suite completed. Before without those extra deps I get the error from the screenshot above.

.github/workflows/tests.yml
```diff
@@ -445,6 +445,168 @@ jobs:
...
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Tune disk performance
+        uses: ./.github/actions/tune-disk-performance
+
+      - name: Install Go
+        uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
+        with:
+          go-version-file: 'go.mod'
```
simondeziel, line 471: I don't think explicitly pulling go is needed here.

.github/workflows/tests.yml
```diff
@@ -445,6 +445,168 @@ jobs:
...
+    needs: [code-tests, documentation]
+    if: ${{ github.repository == 'canonical/lxd' && github.event_name == 'push' }}
+    env:
+      LXD_DIR: "/var/lib/lxd"
+      LXD_OIDC_CLIENT_ID: ${{ secrets.LXD_UI_OIDC_CLIENT_ID }}
+      LXD_OIDC_ISSUER: ${{ secrets.LXD_UI_OIDC_ISSUER }}
+      LXD_OIDC_AUDIENCE: ${{ secrets.LXD_UI_OIDC_AUDIENCE }}
+      LXD_OIDC_USER: ${{ secrets.LXD_UI_OIDC_TEST_USER }}
+      LXD_OIDC_PASSWORD: ${{ secrets.LXD_UI_OIDC_TEST_PASSWORD }}
+      LXD_OIDC_GROUPS_CLAIM: "lxd-idp-groups"
```
simondeziel, line 460: This is missing `LD_LIBRARY_PATH: "/home/runner/go/bin/dqlite/libs/"` which is probably why you have that error about missing `libdqlite.so.0`.

.github/workflows/tests.yml
```diff
@@ -445,6 +445,168 @@ jobs:
...
+          mkdir -p "${GOCOVERDIR}"
+        if: env.GOCOVERDIR != ''
+
+      - name: Run LXD daemon
+        run: |
+          set -eux
+          echo "root:1000000:1000000000" | sudo tee /etc/subuid /etc/subgid
+          export CGO_CFLAGS="-I/home/runner/work/lxd/lxd-test/vendor/dqlite/include/"
+          export CGO_LDFLAGS="-L/home/runner/work/lxd/lxd-test/vendor/dqlite/.libs/"
+          export LD_LIBRARY_PATH="/home/runner/work/lxd/lxd-test/vendor/dqlite/.libs/"
```
simondeziel, line 511: This path seems wrong as the system-test-deps are all unpacked into `~/go/bin`. The `LD_LIBRARY_PATH` suggested earlier should work.

.github/workflows/tests.yml
simondeziel: I think this is all tied to a wrong `LD_LIBRARY_PATH` and your change works by accident as it pulls in `libdqlite-dev` which happens to depend on `libdqlite0` that has a (random) copy of that `libdqlite.so.0`.

Fixing the `LD_LIBRARY_PATH` should make any modification of the actions unneeded. I realize there is a lot of back-n-forth and that must be frustrating for you so I'd like to propose doing a Meet tomorrow if you think that can help/speed things up.

edlerd: You are absolutely right, thx for the tip! I can see it worked with the fixed path in [this fork run](https://github.com/edlerd/lxd/actions/runs/11967855652/job/33366069682).

.github/workflows/tests.yml
edlerd: correct, removed it.

.github/workflows/tests.yml
edlerd: We don't need it here, this is the ui test runner in js, it doesn't need to talk to dqlite directly. Only talks to the ui, that is using the LXD api.

.github/workflows/tests.yml
edlerd: Yes, thanks for flagging this :)

.github/workflows/tests.yml
tomponline: I wonder if the GOCOVERDIR env var isn't being set in your fork's run because of this line

https://github.com/canonical/lxd/blob/5fdd983438347b63cb4077080ba4c64548a8e8a5/.github/workflows/tests.yml#L15

.github/workflows/tests.yml
tomponline: you need the dqlite libs and that env var to run lxd binary though

.github/workflows/tests.yml
edlerd: > I wonder if the GOCOVERDIR env var isn't being set in your fork's run because of this line

Yeah, I saw that and explicitly set `GOCOVERDIR` for the [ui step in recent versions](https://github.com/canonical/lxd/pull/14035/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR454) of the workflow.

.github/workflows/tests.yml
edlerd: Yes, that is in another previous step, which is launching the LXD api.

.github/workflows/tests.yml
tomponline: [I'd suggest ](https://github.com/canonical/lxd/blob/5fdd983438347b63cb4077080ba4c64548a8e8a5/.github/workflows/tests.yml#L122) then relies on that env var being set here:

https://github.com/canonical/lxd/blob/main/Makefile#L39

.github/workflows/tests.yml
```diff
@@ -445,6 +445,159 @@ jobs:
...
+
+      - name: Checkout LXD-UI
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          repository: 'canonical/lxd-ui'
+          ref: main
+          path: lxd-ui
+
+      - name: Install Dotrun
+        run: sudo pip3 install dotrun
```
tomponline, line 519: lets do this before starting any daemons too

.github/workflows/tests.yml
```diff
@@ -445,6 +445,159 @@ jobs:
...
+
+      - name: Install LXD-UI dependencies
```
tomponline, line 521: lets do this before starting any daemons

.github/workflows/tests.yml
```diff
@@ -445,6 +445,159 @@ jobs:
...
+      - name: Setup LXD
+        shell: bash
+        run: |
+          set -eux
+          export PATH="/home/runner/go/bin:$PATH"
+          sudo -E LXD_DIR=/var/lib/lxd lxc storage create default zfs
+          sudo -E LXD_DIR=/var/lib/lxd lxc profile device add default root disk path=/ pool=default
+          sudo -E LXD_DIR=/var/lib/lxd lxc network create local-network
+          sudo -E LXD_DIR=/var/lib/lxd lxc profile device add default eth0 nic network=local-network
+          sudo -E LXD_DIR=/var/lib/lxd lxc config set core.https_address "[::]:8443"
```
tomponline, line 548: should this be `::1`?

.github/workflows/tests.yml
```diff
@@ -445,6 +445,159 @@ jobs:
...
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Install Playwright Browsers
+        run: |
+          set -eux
+          cd lxd-ui
+          npx playwright install --with-deps chromium
```
tomponline, line 562: lets do this before starting any daemons

.github/workflows/tests.yml
edlerd: I think that doesn't work, because the UI is running in a container and can't access the host system loopback directly.

.github/workflows/tests.yml
edlerd: This [recent run](https://github.com/edlerd/lxd/actions/runs/12030477528) on my fork produced coverage data for the ui run. I am not sure what caused this before to fail though, but happy that it works now :)

edlerd:
I think all issues and concerns raised above have been addressed. Please give this another go @tomponline and @simondeziel

edlerd:
This [run on my fork](https://github.com/edlerd/lxd/actions/runs/12030477528) was passing and producing coverage data for the LXD-UI tests.

.github/workflows/tests.yml
```diff
@@ -445,6 +445,155 @@ jobs:
...
+          sudo -E LXD_DIR=/var/lib/lxd PATH=/home/runner/go/bin:$PATH ./lxd-ui/tests/scripts/setup_test
+
+      - name: Run Playwright tests
+        run: |
+          set -eux
+          cd lxd-ui
+          sudo chown $USER -R /home/runner/.config
+          sudo chown $USER /var/lib/lxd/unix.socket
+          CI=true DISABLE_VM_TESTS=true LXD_DIR=/var/lib/lxd PATH=/home/runner/go/bin:$PATH npx playwright test --project chromium:lxd-latest-edge
+          #npx playwright test --project chromium:lxd-5.21-edge
```
tomponline, line 573: please can you remove these comments

.github/workflows/tests.yml
```diff
@@ -445,6 +445,155 @@ jobs:
...
+          set -eux
+          sudo -E LXD_DIR=/var/lib/lxd PATH=/home/runner/go/bin:$PATH ./lxd-ui/tests/scripts/setup_test
+
+      - name: Run Playwright tests
+        run: |
+          set -eux
+          cd lxd-ui
+          sudo chown $USER -R /home/runner/.config
+          sudo chown $USER /var/lib/lxd/unix.socket
+          CI=true DISABLE_VM_TESTS=true LXD_DIR=/var/lib/lxd PATH=/home/runner/go/bin:$PATH npx playwright test --project chromium:lxd-latest-edge
```
tomponline, line 572: why dont you add `PATH=/home/runner/go/bin:$PATH` to the `env` section to save repeating it in a few places?

https://github.com/canonical/lxd/pull/14035/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR453

edlerd: I tried that, and I think the resolution of `$PATH` in the `env` section was not producing the same result as doing it in here.

.github/workflows/tests.yml
```diff
@@ -445,6 +445,155 @@ jobs:
...
+        shell: bash
+        run: |
+          set -eux
+          export PATH="/home/runner/go/bin:$PATH"
+          sudo -E LXD_DIR=/var/lib/lxd lxc storage create default zfs
+          sudo -E LXD_DIR=/var/lib/lxd lxc profile device add default root disk path=/ pool=default
+          sudo -E LXD_DIR=/var/lib/lxd lxc network create local-network
+          sudo -E LXD_DIR=/var/lib/lxd lxc profile device add default eth0 nic network=local-network
+          sudo -E LXD_DIR=/var/lib/lxd lxc config set core.https_address "[::]:8443"
+          sudo -E LXD_DIR=/var/lib/lxd lxc config set cluster.https_address "127.0.0.1"
```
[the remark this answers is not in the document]
edlerd: We have a series of tests around managing cluster groups, that only works if clustering is enabled.

.github/workflows/tests.yml
edlerd: removed them just now.

.github/workflows/tests.yml
```diff
@@ -445,6 +445,153 @@ jobs:
...
+          ref: main
+          path: lxd-ui
+
+      - name: Install Dotrun
+        run: sudo pip3 install dotrun
+
+      - name: Install LXD-UI dependencies
+        run: |
+          set -eux
+          sudo chmod 0777 ./lxd-ui
```
tomponline, line 479: out of interest how come this is needed?

edlerd: To be honest, this is pretty old code and I don't remember the exact details. I assume `dotrun` is creating files and folders under a different user, and we need to give it permission to do so.

.github/workflows/tests.yml
```diff
@@ -12,7 +12,7 @@ on:
 env:
   LXD_REQUIRED_TESTS: "storage_buckets"
   LXD_SKIP_TESTS: "clustering_upgrade clustering_upgrade_large"
-  GOCOVERDIR: ${{ ( github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' ) && '/home/runner/work/lxd/lxd/coverage' || '' }}
+  GOCOVERDIR: '/home/runner/work/lxd/lxd/coverage'
```
tomponline, line 15: we need this to stay conditional please as we dont need it on the PR runs.

tomponline: it activates the go coverage mode output

edlerd: Whoops, I forgot to revert this back to normal after testing. Did restore the original state just now.

tomponline:
@edlerd looks like we missed blocking the edge builds on the ui tests passing

https://github.com/canonical/lxd/pull/14035/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fL507

edlerd:
> > We might extend this on the stable-5.21 and stable-5.0 branches, as we have dedicated test suites in lxd-ui for those. For older versions, we should continue skipping the e2e tests. I'd keep this as a second step in a followup PR.
> 
> My understanding is that the LXD-UI from main should work on stable-5.21 and stable-5.0 branches, so can I just backport this test as part of our normal backport process?

Basically yes.

The problem is the test suite needs to be adjusted for older versions. You can backport it, but need to adjust the command `npx playwright test --project chromium:lxd-latest-edge` to `npx playwright test --project chromium:lxd-5.21-edge` or `npx playwright test --project chromium:lxd-5.0-edge` to run the right tests for the version.

tomponline:
> The problem is the test suite needs to be adjusted for older versions. You can backport it, but need to adjust the command `npx playwright test --project chromium:lxd-latest-edge` to `npx playwright test --project chromium:lxd-5.21-edge` or `npx playwright test --project chromium:lxd-5.0-edge` to run the right tests for the version.

Please can you update so this is automatic using logic like this:

https://github.com/canonical/lxd/blob/main/.github/workflows/tests.yml#L683-L687

edlerd:
> Please can you update so this is automatic using logic like this:
> 
> https://github.com/canonical/lxd/blob/main/.github/workflows/tests.yml#L683-L687

Addressing this and other comments from above in the follow up PR https://github.com/canonical/lxd/pull/14553

PR #14036, opened 2024-09-03
test/main: don't use any HTTP(S) proxy with localhost

simondeziel:


simondeziel:
@kadinsayani this change is untested but could you give it a try in your PS6 VM and report back please? It should hint `curl` and our `curl` wrapper to not use the proxy when trying to talk to a local LXD over HTTPS.

kadinsayani (commented):
I've tested on my prodstack vm and the previously failing tests are successfully passing. Thanks!

tomponline (approved):

PR #14039, opened 2024-09-04
Test `LXD_INSECURE_TLS`

simondeziel:
Make sure LXD accepts only TLS 1.3 in its default configuration. Also verify that `LXD_INSECURE_TLS=true` reduces the security just enough to provide TLS 1.2 for backward compatibility without allowing CBC nor SHA1.

The fine grain cipher checks will likely require changes over time as Go default cipher list evolves.


tomponline (approved):
Thanks!

PR #14040, opened 2024-09-04
Entity: Rename TypeNode to TypeClusterMember

hamistao:
Updates `TypeNode` to `TypeClusterMember` and its value from `node` to `cluster_member`.
This is done because the term node is inappropriate to be exposed to users, and it will be when the API metrics are introduced.
As of now this type is not exposed to users so an API extension is not needed.

### DIFF
--- a/lxd/cluster/heartbeat.go
+++ b/lxd/cluster/heartbeat.go
@@ -178,7 +178,7 @@ func (hbState *APIHeartbeat) Send(ctx context.Context, networkCert *shared.CertI
 			heartbeatData.Unlock()
 			logger.Debug("Successful heartbeat", logger.Ctx{"remote": address})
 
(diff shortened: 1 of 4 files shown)


lxd/cluster/heartbeat.go
```diff
@@ -557,7 +557,7 @@ func HeartbeatNode(taskCtx context.Context, address string, networkCert *shared.
 	defer func() { _ = response.Body.Close() }()
 
 	if response.StatusCode != http.StatusOK {
-		return fmt.Errorf("Heartbeat request failed with status: %w", api.StatusErrorf(response.StatusCode, response.Status))
+		return fmt.Errorf("Heartbeat request failed with status: %w", api.NewStatusError(response.StatusCode, response.Status))
```
markylaing, line 560: There's a typo in your commit message :)

tomponline (approved):

Issue #13862, opened 2024-08-02
Spelling error in instance snapshot deletion trigger name


markylaing:
See https://github.com/canonical/lxd/pull/13846#discussion_r1701870013_

We'll need a DB patch to fix this.

tomponline:
@markylaing can @kadinsayani help with this?

markylaing:
> @markylaing can @kadinsayani help with this?

Yes that would be great. For this, we need a patch see `lxd/patches.go` to be applied at stage `patchPreLoadClusterConfig` that deletes the `on_instance_snaphot_delete` trigger (note spelling error) if it exists. The [deletion trigger](https://github.com/canonical/lxd/blob/1e0b3535f2c0e0559205b019831bea7af1af21fb/lxd/db/cluster/entity_type_instance_snapshot.go#L43) for the `instance_snapshot` entity type needs to be updated to have the correct spelling.

PR #14043, opened 2024-09-04
Patches: Fix spelling error in instance snapshot deletion trigger name

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/13862

### DIFF
--- a/lxd/patches.go
+++ b/lxd/patches.go
@@ -92,6 +92,7 @@ var patches = []patch{
 	{name: "storage_move_custom_iso_block_volumes_v2", stage: patchPostDaemonStorage, run: patchStorageRenameCustomISOBlockVolumesV2},
 	{name: "storage_unset_invalid_block_settings_v2", stage: patchPostDaemonStorage, run: patchStorageUnsetInvalidBlockSettingsV2},
 	{name: "config_remove_core_trust_password", stage: patchPreLoadClusterConfig, run: patchRemoveCoreTrustPassword},
+	{name: "entity_type_instance_snapshot_on_delete_trigger_typo_fix", stage: patchPreLoadClusterConfig, run: patchEntityTypeInstanceSnapshotOnDeleteTriggerTypoFix},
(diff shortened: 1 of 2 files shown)


lxd/patches.go
```diff
@@ -1346,6 +1347,44 @@ func patchRemoveCoreTrustPassword(_ string, d *Daemon) error {
 	return nil
 }
 
+func patchEntityTypeInstanceSnapshotTypoFix(name string, d *Daemon) error {
+	for {
```
markylaing, line 1351: I've had a good think about the scenarios we need to account for here and I think we should actually run this patch on all cluster members.

Each member will only delete the trigger if it exists, and will have added the new trigger when loading the database, so I don't think we're in danger of missing anything. The worry would be that a member that has not yet been updated reloads the LXD daemon and reapplies the old trigger. However, we're less in danger of this becoming an issue if we apply the patch on all members because once they're all updated we can be certain that the old trigger is deleted. In contrast, if we only run this patch on the leader then we're in danger of another cluster member that has not yet been updated reloading the LXD daemon and re-adding the old trigger.

The upside is the patch will be much simpler. It should look like https://github.com/canonical/lxd/blob/f9b1a03ef3c5efdb2ec32c624e26f24eecaf296f/lxd/patches.go#L1334-L1347 but removing the old trigger.

Does this make sense? Could you sanity check this please @tomponline?

kadinsayani: That makes sense to me, I'll apply the changes after the go ahead from @tomponline

tomponline: Because triggers are setup on all members, the patch that affects triggers also has to be setup on all members right?

kadinsayani: Yes, exactly. Since triggers are associated with each database member, applying the patch to all members should ensure consistency.

tomponline:
@markylaing are you happy with this one?

markylaing (approved):
LGTM thanks :)

tomponline (approved):

PR #14044, opened 2024-09-04
Set OSD pool size when creating `ceph` and `cephfs` storage pools

### LINKED ISSUE #14006: Allow setting `ceph.osd.pool_size` on pool creation
At least with MicroCeph, OSD pools will default to a replication size of 3. That means if there are fewer than that many OSDs available for replication, LXD will fail to create the pool when running `lxc storage create pool ceph`. The command will block as the OSD pool will try to replicate unsuccessfully during initialization. 

In such cases (admittedly usually testing setups), it may be useful to specify `ceph.osd.pool_size` to apply as an initial argument when creating the pool. This may also be useful in the LXD test suite as well, where we need to set [`microceph.ceph config set global osd_pool_default_size 1`](https://github.com/canonical/lxd/blob/bd35d6d2cc3c550e2c1a4920f390437c398a910e/.github/workflows/tests.yml#L360) to change the default behaviour for all OSD pools in the test suite to disable replication.

Somewhat related to the MicroCloud OSD replication discussions going on, but MicroCloud would not leverage this capability in LXD because it also needs to be able to update the value when adding new cluster members, and it seems the ceph storage pool has [no update implementation either.](https://github.com/canonical/lxd/blob/bd35d6d2cc3c550e2c1a4920f390437c398a910e/lxd/storage/drivers/driver_ceph.go#L361)

masnax:
Closes #14006

Adds the keys `ceph.osd.pool_size` and `cephfs.osd_pool_size` (in keeping with the other naming schemes).

By default, if no value is supplied, the pool size will be pulled from the global default pool size. It can be set to any value larger than 0. 

On update, we will try to re-apply the OSD pool size value, if it has changed.

### DIFF
--- a/.github/actions/setup-microceph/action.yml
+++ b/.github/actions/setup-microceph/action.yml
@@ -53,8 +53,9 @@ runs:
 
           sudo snap install microceph --channel "${{ inputs.microceph-channel }}"
           sudo microceph cluster bootstrap
-          sudo microceph.ceph config set global osd_pool_default_size 1
+          sudo microceph.ceph config set global mon_allow_pool_size_one true
           sudo microceph.ceph config set global mon_allow_pool_delete true
+          sudo microceph.ceph config set global osd_pool_default_size 1
           sudo microceph.ceph config set global osd_memory_target 939524096
           sudo microceph.ceph osd crush rule rm replicated_rule
           sudo microceph.ceph osd crush rule create-replicated replicated default osd
--- a/.github/actions/setup-microceph/action.yml
+++ b/.github/actions/setup-microceph/action.yml
@@ -6,6 +6,10 @@ inputs:
     description: MicroCeph snap channel to install
     default: "latest/edge"
     type: string
+  osd-count:
+    description: Number of OSDs to add to MicroCeph
+    default: 1
+    type: integer
 
 runs:
   using: composite
@@ -51,6 +55,7 @@ runs:
           }
           trap cleanup ERR HUP INT TERM
 
+          ephemeral_disk="${{ steps.free_ephemeral_disk.outputs.ephemeral_disk }}"
           sudo snap install microceph --channel "${{ inputs.microceph-channel }}"
           sudo microceph cluster bootstrap
           sudo microceph.ceph config set global mon_allow_pool_size_one true
@@ -62,7 +67,29 @@ runs:
           for flag in nosnaptrim nobackfill norebalance norecover noscrub nodeep-scrub; do
               sudo microceph.ceph osd set $flag
           done
-          sudo microceph disk add --wipe "${{ steps.free_ephemeral_disk.outputs.ephemeral_disk }}"
+
+          # If there is more than one OSD, set up partitions.
+          if [ ${{ inputs.osd-count }} -gt 1 ]; then
+            sudo blkdiscard "${ephemeral_disk}" --force
+            sudo parted "${ephemeral_disk}" --script mklabel gpt
+
+            for i in $(seq 1 "${{ inputs.osd-count }}"); do
+              # Create equal sized partitions for each OSD.
+              min="$(( (${i}-1) *  100 / ${{ inputs.osd-count }} ))"
+              max="$(( ${i} * 100 / ${{ inputs.osd-count }} ))"
+              sudo parted "${ephemeral_disk}" --align optimal --script mkpart primary "${min}%" "${max}%"
+
+              # MicroCeph does not accept partitions directly.
+              # See: https://github.com/canonical/microceph/issues/251
+              disk="$(losetup -f)"
+              sudo losetup --direct-io=on "${disk}" "${ephemeral_disk}${i}"
+              sudo microceph disk add "${disk}"
+            done
+          else
+              sudo microceph disk add --wipe "${ephemeral_disk}"
+          fi
+
+
           sudo rm -rf /etc/ceph
           sudo ln -s /var/snap/microceph/current/conf/ /etc/ceph
           sudo microceph enable rgw
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2535,3 +2535,6 @@ Adds a new {config:option}`device-unix-hotplug-device-conf:ownership.inherit` co
 ## `unix_device_hotplug_subsystem_device_option`
 
 Adds a new {config:option}`device-unix-hotplug-device-conf:subsystem` configuration option for `unix-hotplug` devices. This adds support for detecting `unix-hotplug` devices by subsystem, and can be used in conjunction with {config:option}`device-unix-hotplug-device-conf:productid` and {config:option}`device-unix-hotplug-device-conf:vendorid`.
+
+## `storage_ceph_osd_pool_size`
+This introduces the configuration keys {config:option}`storage-ceph-pool-conf:ceph.osd.pool_size`, and {config:option}`storage-cephfs-pool-conf:cephfs.osd_pool_size` to be used when adding or updating a `ceph` or `cephfs` storage pool to instruct LXD to create set the replication size for the underlying OSD pools.
--- a/doc/metadata.txt
+++ b/doc/metadata.txt
@@ -4996,6 +4996,14 @@ Specify either a cron expression (`<minute> <hour> <dom> <month> <dow>`), a comm
 
 ```
 
+```{config:option} ceph.osd.pool_size storage-ceph-pool-conf
+:defaultdesc: "`3`"
+:shortdesc: "Number of RADOS object replicas. Set to 1 for no replication."
+:type: "string"
+This option specifies the name for the file metadata OSD pool that should be used when
+creating a file system automatically.
+```
+
 ```{config:option} ceph.rbd.clone_copy storage-ceph-pool-conf
 :defaultdesc: "`true`"
 :scope: "global"
@@ -5196,6 +5204,14 @@ creating a file system automatically.
 :type: "string"
 This option specifies the number of OSD pool placement groups (`pg_num`) to use
 when creating a missing OSD pool.
+```
+
+```{config:option} cephfs.osd_pool_size storage-cephfs-pool-conf
+:defaultdesc: "`3`"
+:shortdesc: "Number of RADOS object replicas. Set to 1 for no replication."
+:type: "string"
+This option specifies the number of OSD pool replicas to use
+when creating an OSD pool.
 ```
 
 ```{config:option} cephfs.path storage-cephfs-pool-conf
--- a/lxd/storage/drivers/driver_ceph.go
+++ b/lxd/storage/drivers/driver_ceph.go
@@ -406,6 +406,23 @@ func (d *ceph) Validate(config map[string]string) error {
 
 // Update applies any driver changes required from a configuration change.
 func (d *ceph) Update(changedConfig map[string]string) error {
+	newSize, changed := changedConfig["ceph.osd.pool_size"]
+	if changed {
+		_, err := shared.TryRunCommand("ceph",
+			"--name", "client."+d.config["ceph.user.name"],
+			"--cluster", d.config["ceph.cluster_name"],
+			"osd",
+			"pool",
+			"set",
+			d.config["ceph.osd.pool_name"],
+			"size",
+			newSize,
+			"--yes-i-really-mean-it")
+		if err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
--- a/lxd/storage/drivers/driver_ceph.go
+++ b/lxd/storage/drivers/driver_ceph.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os/exec"
+	"strconv"
 	"strings"
 
 	"github.com/canonical/lxd/lxd/migration"
@@ -111,6 +112,29 @@ func (d *ceph) FillConfig() error {
 
 	if d.config["ceph.osd.pg_num"] == "" {
 		d.config["ceph.osd.pg_num"] = "32"
+	}
+
+	if d.config["ceph.osd.pool_size"] == "" {
+		size, err := shared.TryRunCommand("ceph",
+			"--name", "client."+d.config["ceph.user.name"],
+			"--cluster", d.config["ceph.cluster_name"],
+			"config",
+			"get",
+			"mon",
+			"osd_pool_default_size",
+			"--format",
+			"json")
+		if err != nil {
+			return err
+		}
+
+		var sizeInt int
+		err = json.Unmarshal([]byte(size), &sizeInt)
+		if err != nil {
+			return err
+		}
+
+		d.config["ceph.osd.pool_size"] = strconv.Itoa(sizeInt)
 	}
 
 	return nil
--- a/lxd/storage/drivers/driver_ceph.go
+++ b/lxd/storage/drivers/driver_ceph.go
@@ -344,6 +344,14 @@ func (d *ceph) Validate(config map[string]string) error {
 		//  shortdesc: Number of placement groups for the OSD storage pool
 		//  scope: global
 		"ceph.osd.pg_num": validate.IsAny,
+		// lxdmeta:generate(entities=storage-ceph; group=pool-conf; key=ceph.osd.pool_size)
+		// This option specifies the name for the file metadata OSD pool that should be used when
+		// creating a file system automatically.
+		// ---
+		//  type: string
+		//  defaultdesc: `3`
+		//  shortdesc: Number of RADOS object replicas. Set to 1 for no replication.
+		"ceph.osd.pool_size": validate.Optional(validate.IsInRange(1, 255)),
 		// lxdmeta:generate(entities=storage-ceph; group=pool-conf; key=ceph.osd.pool_name)
 		//
 		// ---
--- a/lxd/storage/drivers/driver_ceph.go
+++ b/lxd/storage/drivers/driver_ceph.go
@@ -196,6 +196,20 @@ func (d *ceph) Create() error {
 
 		revert.Add(func() { _ = d.osdDeletePool() })
 
+		_, err = shared.TryRunCommand("ceph",
+			"--name", "client."+d.config["ceph.user.name"],
+			"--cluster", d.config["ceph.cluster_name"],
+			"osd",
+			"pool",
+			"set",
+			d.config["ceph.osd.pool_name"],
+			"size",
+			d.config["ceph.osd.pool_size"],
+			"--yes-i-really-mean-it")
+		if err != nil {
+			return err
+		}
+
 		// Initialize the pool. This is not necessary but allows the pool to be monitored.
 		_, err = shared.TryRunCommand("rbd",
 			"--id", d.config["ceph.user.name"],
--- a/lxd/storage/drivers/driver_ceph.go
+++ b/lxd/storage/drivers/driver_ceph.go
@@ -159,7 +159,7 @@ func (d *ceph) Create() error {
 	if !poolExists {
 		// Create new osd pool.
 		_, err := shared.TryRunCommand("ceph",
-			"--name", fmt.Sprintf("client.%s", d.config["ceph.user.name"]),
+			"--name", "client."+d.config["ceph.user.name"],
 			"--cluster", d.config["ceph.cluster_name"],
 			"osd",
 			"pool",
@@ -218,7 +218,7 @@ func (d *ceph) Create() error {
 
 		// Use existing OSD pool.
 		msg, err := shared.RunCommand("ceph",
-			"--name", fmt.Sprintf("client.%s", d.config["ceph.user.name"]),
+			"--name", "client."+d.config["ceph.user.name"],
 			"--cluster", d.config["ceph.cluster_name"],
 			"osd",
 			"pool",
@@ -398,7 +398,7 @@ func (d *ceph) GetResources() (*api.ResourcesStoragePool, error) {
 
 	err := shared.RunCommandWithFds(context.TODO(), nil, &stdout,
 		"ceph",
-		"--name", fmt.Sprintf("client.%s", d.config["ceph.user.name"]),
+		"--name", "client."+d.config["ceph.user.name"],
 		"--cluster", d.config["ceph.cluster_name"],
 		"df",
 		"-f", "json")
--- a/test/suites/storage_driver_ceph.sh
+++ b/test/suites/storage_driver_ceph.sh
@@ -129,7 +129,7 @@ test_storage_driver_ceph() {
     lxc profile device remove default root
     lxc storage delete "lxdtest-$(basename "${LXD_DIR}")-pool1"
     lxc storage delete "lxdtest-$(basename "${LXD_DIR}")-pool2"
-    ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool rm "lxdtest-$(basename "${LXD_DIR}")-existing-osd-pool" "lxdtest-$(basename "${LXD_DIR}")-existing-osd-pool"  --yes-i-really-really-mean-it
+    ! ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool ls | grep -q "lxdtest-$(basename "${LXD_DIR}")-existing-osd-pool" || false
 
 
     # Test that pre-existing OSD pools are not affected by the config option in LXD. Only the associated pool should be affected.
--- a/test/suites/storage_driver_ceph.sh
+++ b/test/suites/storage_driver_ceph.sh
@@ -130,6 +130,25 @@ test_storage_driver_ceph() {
     lxc storage delete "lxdtest-$(basename "${LXD_DIR}")-pool1"
     lxc storage delete "lxdtest-$(basename "${LXD_DIR}")-pool2"
     ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool rm "lxdtest-$(basename "${LXD_DIR}")-existing-osd-pool" "lxdtest-$(basename "${LXD_DIR}")-existing-osd-pool"  --yes-i-really-really-mean-it
+
+
+    # Test that pre-existing OSD pools are not affected by the config option in LXD. Only the associated pool should be affected.
+    # .mgr is auto-created by Ceph, so give it a value that is different from the size supplied to the `lxc storage create` commands.
+    ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool set ".mgr" size 3
+    pool1="lxdtest-$(basename "${LXD_DIR}")-pool1"
+    pool2="lxdtest-$(basename "${LXD_DIR}")-pool2"
+    lxc storage create "${pool1}" ceph volume.size=25MiB ceph.osd.pg_num=16 ceph.osd.pool_size=1
+    [[ "$(ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool get "${pool1}" size --format json | jq '.size')" = "1" ]]
+    [[ "$(ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool get ".mgr" size --format json | jq '.size')" = "3" ]]
+
+    lxc storage create "${pool2}" ceph volume.size=25MiB ceph.osd.pg_num=16 ceph.osd.pool_size=2
+    [[ "$(ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool get "${pool1}" size --format json | jq '.size')" = "1" ]]
+    [[ "$(ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool get "${pool2}" size --format json | jq '.size')" = "2" ]]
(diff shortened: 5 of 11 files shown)


masnax:
I can't seem to get the docs to pick up the new extensions, I just get
```
WARNING: Could not find target storage-ceph-pool-conf:ceph.osd.pool_size in api-extensions
```

masnax:
@simondeziel Got any recommendations for setting up more than 1 OSD for microceph? I'd like to add a test case for the new key but it's tied to the number of available disks supplied to MicroCeph, which is 1 in the github workflow tests.

masnax:
Looks like 1 thing I overlooked in my testing is that `size=1` is disabled by default unless the global config is changed.

So while the impetus for this PR was to avoid running `microceph.ceph config set global osd_pool_default_size 1` to set the default pool size, and adding these keys avoids that, a user will still need to run `ceph config set global mon_allow_pool_size_one true` to make the keys work if a user wants `size=1`

It's still added flexibility, but if you'd prefer to avoid managing these keys in LXD since it doesn't actually solve the underlying problem, I can close the PR @tomponline

simondeziel:
> @simondeziel Got any recommendations for setting up more than 1 OSD for microceph? I'd like to add a test case for the new key but it's tied to the number of available disks supplied to MicroCeph, which is 1 in the github workflow tests.

Could we maybe use the ephemeral disk, split it into 3 partitions and export each of those as loop dev backed by their respective disk part?

test/suites/storage_driver_ceph.sh
```diff
@@ -130,6 +130,22 @@ test_storage_driver_ceph() {
     lxc storage delete "lxdtest-$(basename "${LXD_DIR}")-pool1"
     lxc storage delete "lxdtest-$(basename "${LXD_DIR}")-pool2"
     ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool rm "lxdtest-$(basename "${LXD_DIR}")-existing-osd-pool" "lxdtest-$(basename "${LXD_DIR}")-existing-osd-pool"  --yes-i-really-really-mean-it
+
+
+    # Test that the pool size setting only applies to the created pool, and not any other pools.
+    pool1="$("lxdtest-$(basename "${LXD_DIR}")-pool1")"
+    pool2="$("lxdtest-$(basename "${LXD_DIR}")-pool2")"
+    lxc storage create "${pool1}" ceph volume.size=25MiB ceph.osd.pg_num=16 ceph.osd.pool_size=1
+    ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool get "${pool1}" size --format json | jq '.size' | grep -q 1
```
simondeziel, line 139: Using shell comparison should help with debugging as we will see what `.size` we got on the left hand side of the comparison:
```suggestion
    [[ "$(ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool get "${pool1}" size --format json | jq '.size')" = "1" ]]
```

masnax: Fixed

lxd/storage/drivers/driver_ceph.go
```diff
@@ -112,6 +114,29 @@ func (d *ceph) FillConfig() error {
 		d.config["ceph.osd.pg_num"] = "32"
 	}
 
+	if d.config["ceph.osd.pool_size"] == "" {
+		size, err := shared.TryRunCommand("ceph",
+			"--name", fmt.Sprintf("client.%s", d.config["ceph.user.name"]),
```
simondeziel, line 119: It's a tiny nit but simple string concat is more efficient and needs less keystrokes ;)
```suggestion
			"--name", "client." + d.config["ceph.user.name"],
```

masnax: Someone was paying attention at the performance sessions in Madrid :)

Fixed

doc/api-extensions.md
```diff
@@ -2517,6 +2517,10 @@ Adds support for using a bridge network with a specified VLAN ID as an OVN uplin
 Adds `logical_cpus` field to `GET /1.0/cluster/members/{name}/state` which
 contains the total available logical CPUs available when LXD started.
 
+<<<<<<< HEAD
```
simondeziel, line 2520: Merge/rebase leftover.

.github/actions/setup-microceph/action.yml
```diff
@@ -51,17 +51,33 @@ runs:
...
+          sudo losetup "${disk1}" "${ephemeral_disk}1"
+          disk2="$(losetup -f)"
+          sudo losetup "${disk2}" "${ephemeral_disk}2"
+          disk3="$(losetup -f)"
+          sudo losetup "${disk3}" "${ephemeral_disk}3"
+
           sudo snap install microceph --channel "${{ inputs.microceph-channel }}"
           sudo microceph cluster bootstrap
           sudo microceph.ceph config set global osd_pool_default_size 1
+          sudo microceph.ceph config set global mon_allow_pool_size_one true
```
simondeziel, line 70: How about having the 2 `mon_*` settings grouped before the OSD ones? I'm thinking this would avoid `mon` complaining about the OSD pool size being 1 for a very brief moment?

.github/actions/setup-microceph/action.yml
```diff
@@ -51,17 +51,33 @@ runs:
           }
           trap cleanup ERR HUP INT TERM
 
+          ephemeral_disk="${{ steps.free_ephemeral_disk.outputs.ephemeral_disk }}"
+          sudo parted "${ephemeral_disk}" --script mklabel gpt
+          sudo parted "${ephemeral_disk}" --script mkpart primary 0% 33%
+          sudo parted "${ephemeral_disk}" --script mkpart primary 33% 66%
+          sudo parted "${ephemeral_disk}" --script mkpart primary 66% 100%
```
simondeziel, line 58: I like it and didn't know it could take %.

.github/actions/setup-microceph/action.yml
```diff
@@ -51,17 +51,33 @@ runs:
...
+
+          disk1="$(losetup -f)"
+          sudo losetup "${disk1}" "${ephemeral_disk}1"
+          disk2="$(losetup -f)"
+          sudo losetup "${disk2}" "${ephemeral_disk}2"
+          disk3="$(losetup -f)"
+          sudo losetup "${disk3}" "${ephemeral_disk}3"
```
simondeziel, line 65: I take it that MicroCeph still cannot take partitions directly right? How about adding a link to https://github.com/canonical/microceph/issues/251 in a comment?

lxd/storage/drivers/driver_ceph.go
```diff
@@ -303,6 +342,13 @@ func (d *ceph) Validate(config map[string]string) error {
 		//  defaultdesc: `32`
 		//  shortdesc: Number of placement groups for the OSD storage pool
 		"ceph.osd.pg_num": validate.IsAny,
+		// lxdmeta:generate(entities=storage-ceph; group=pool-conf; key=ceph.osd.pool_size)
+		//
+		// ---
+		//  type: string
+		//  defaultdesc: `3`
+		//  shortdesc: Number of RADOS object replicas. Set to 1 for no replication.
+		"ceph.osd.pool_size": validate.Optional(validate.IsInRange(1, math.MaxInt64)),
```
simondeziel, line 351: While technically correct, maybe we could put an upper bound that's a little less... gigantic :) Hardcoding `255` feels future-proof enough I'd say.

simondeziel: I just noticed that `ceph.osd.pg_num` has no such validation.

doc/metadata.txt
```diff
@@ -5074,6 +5081,14 @@ This option specifies the number of OSD pool placement groups (`pg_num`) to use
 when creating a missing OSD pool.
 ```
 
+```{config:option} cephfs.osd_pool_size storage-cephfs-pool-conf
+:defaultdesc: "`3`"
+:shortdesc: "Number of RADOS object replicas. Set to 1 for no replication."
+:type: "string"
+This option specifies the number of OSD pool replicas to use
+when creating an OSD pool.
```
simondeziel, line 5089: Why have a long description for `cephfs` and not for `ceph`? Also, is this really something that needs to be done at creation time only?

simondeziel (commented):
LGTM in general with the caveat that the `setup-microceph` action is being used elsewhere and should probably not default to using 3 partitions. Could you maybe make that an action parameter?

masnax:
> LGTM in general with the caveat that the `setup-microceph` action is being used elsewhere and should probably not default to using 3 partitions. Could you maybe make that an action parameter?

Hmm, does it matter since we set the replication factor to 1 anyway?

simondeziel:
> > LGTM in general with the caveat that the `setup-microceph` action is being used elsewhere and should probably not default to using 3 partitions. Could you maybe make that an action parameter?
> 
> Hmm, does it matter since we set the replication factor to 1 anyway?

That cuts the available space in 3 (so ~25GiB instead of ~75GiB IIRC) which might be a bit tight for some `lxd-ci.git` tests. Also, it comes with the losetup overhead.

lxd/storage/drivers/driver_ceph.go
tomponline: Indeed!

doc/metadata.txt
masnax: Added it now.

lxd/storage/drivers/driver_ceph.go
masnax: Set it to 255 now. Interestingly Ceph totally lets you use pg_num=0 without erroring out. Not sure what it does though.

masnax:
This one's now rebased (due to API extensions) and ready for a review.

lxd/storage/drivers/driver_ceph.go
```diff
@@ -112,6 +113,29 @@ func (d *ceph) FillConfig() error {
 		d.config["ceph.osd.pg_num"] = "32"
 	}
 
+	if d.config["ceph.osd.pool_size"] == "" {
+		size, err := shared.TryRunCommand("ceph",
```
tomponline, line 117: Please can you help me understand the reason we are copying the osd_pool_default_size out of the ceph installation into the pool's config if its not specified. This seems to be setting a new precedent of reading a property of the environment and then persisting it into the pool's config.

tomponline:
> By default, if no value is supplied, the pool size will be pulled from the global default pool size.

Why dont we just leave the config setting unpopulated in that case?

lxd/storage/drivers/driver_ceph.go
masnax: So there is no difference between ceph using the default value or the LXD supplied value for pool size, so I thought it would be helpful to see the default value that the pool was created with, without having to check ceph directly. 

Otherwise the UX requires a lot of back-and-forth between ceph & LXD:
* Create ceph cluster & osds (using ceph)
* Create osd pool (using LXD)
* Check the osd pool size (using ceph)
* Update the osd pool size (using LXD) 

You're right that it's a bit inconsistent with other fields like `pg_num` though, so I can remove it.

tomponline: so it this setting just stored for posterity and not used again once the pool is created or the setting is updated?

masnax: If a user updates the LXD config on the storage pool manually, then the change will be propagated to ceph. LXD doesn't communicate with ceph periodically, so if the pool size changes through Ceph, LXD will not see the change. So this first value is effectively just a snapshot of the pool size from when the pool was created.

test/suites/storage_driver_ceph.sh
```diff
@@ -130,6 +130,24 @@ test_storage_driver_ceph() {
     lxc storage delete "lxdtest-$(basename "${LXD_DIR}")-pool1"
     lxc storage delete "lxdtest-$(basename "${LXD_DIR}")-pool2"
     ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool rm "lxdtest-$(basename "${LXD_DIR}")-existing-osd-pool" "lxdtest-$(basename "${LXD_DIR}")-existing-osd-pool"  --yes-i-really-really-mean-it
+
+
+    # Test that the pool size setting only applies to the created pool, and not any other pools.
+    microceph.ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool set ".mgr" size 3
```
tomponline, line 136: the tests shouldnt depend on microceph, can we do this in an agnostic way?

lxd/storage/drivers/driver_ceph.go
masnax: So shall I remove it since it doesn't match the behaviour of other keys?

.github/actions/setup-microceph/action.yml
```diff
@@ -51,17 +55,40 @@ runs:
...
+            for i in $(seq 1 "${{ inputs.osd-count }}"); do
+              # Create equal sized partitions for each OSD.
+              min="$(( (${i}-1) *  100 / ${{ inputs.osd-count }} ))"
+              max="$(( ${i} * 100 / ${{ inputs.osd-count }} ))"
+              sudo parted "${ephemeral_disk}" --script mkpart primary "${min}%" "${max}%"
+
+              # MicroCeph does not accept partitions directly.
+              # See: https://github.com/canonical/microceph/issues/251
+              disk="$(losetup -f)"
+              sudo losetup "${disk}" "${ephemeral_disk}${i}"
```
simondeziel, line 84: I'd probably throw in a `--direct-io=on` in there to avoid trashing the memory by essentially caching all the disk accesses.

.github/actions/setup-microceph/action.yml
```diff
@@ -51,17 +55,40 @@ runs:
...
+
+          # If there is more than one OSD, set up partitions.
+          if [ ${{ inputs.osd-count }} -gt 1 ]; then
+            sudo parted "${ephemeral_disk}" --script mklabel gpt
+
+            for i in $(seq 1 "${{ inputs.osd-count }}"); do
+              # Create equal sized partitions for each OSD.
+              min="$(( (${i}-1) *  100 / ${{ inputs.osd-count }} ))"
+              max="$(( ${i} * 100 / ${{ inputs.osd-count }} ))"
+              sudo parted "${ephemeral_disk}" --script mkpart primary "${min}%" "${max}%"
```
simondeziel, line 79: `man parted` doesn't specify what kind of alignment it does by default so to be sure we get the the optimal one, could you add `--align optimal` in there?

.github/actions/setup-microceph/action.yml
```diff
@@ -51,17 +55,40 @@ runs:
...
           sudo microceph.ceph osd crush rule rm replicated_rule
           sudo microceph.ceph osd crush rule create-replicated replicated default osd
           for flag in nosnaptrim nobackfill norebalance norecover noscrub nodeep-scrub; do
               sudo microceph.ceph osd set $flag
           done
-          sudo microceph disk add --wipe "${{ steps.free_ephemeral_disk.outputs.ephemeral_disk }}"
+
+          # If there is more than one OSD, set up partitions.
+          if [ ${{ inputs.osd-count }} -gt 1 ]; then
+            sudo parted "${ephemeral_disk}" --script mklabel gpt
```
simondeziel, line 73: How about adding a `blkdiscard ${ephemeral_disk}` before creating the GPT label?

.github/actions/setup-microceph/action.yml
```diff
@@ -51,17 +55,40 @@ runs:
...
+              # Create equal sized partitions for each OSD.
+              min="$(( (${i}-1) *  100 / ${{ inputs.osd-count }} ))"
+              max="$(( ${i} * 100 / ${{ inputs.osd-count }} ))"
+              sudo parted "${ephemeral_disk}" --script mkpart primary "${min}%" "${max}%"
+
+              # MicroCeph does not accept partitions directly.
+              # See: https://github.com/canonical/microceph/issues/251
+              disk="$(losetup -f)"
+              sudo losetup "${disk}" "${ephemeral_disk}${i}"
+              sudo microceph disk add --wipe "${disk}"
```
simondeziel, line 85: With the `blkdiscard` suggested earlier, we could potentially avoid this `--wipe`? I don't know how it's implemented under the hood. Maybe it's doing just that already? Or maybe it's doing something slow like `dd if=/dev/zero`?

Or maybe it's just always a bit slow to add a disk (3-4s) and not related to wiping at all.

simondeziel (commented):
Thanks for extending that action, it's a nice improvement! I've left a few comments but I'm already quite pleased by the fact that performance is ~identical to the single OSD case so good job!

.github/actions/setup-microceph/action.yml
masnax: It is indeed just a `dd if=/dev/zero` iirc.

.github/actions/setup-microceph/action.yml
```diff
@@ -51,17 +55,41 @@ runs:
...
           sudo microceph.ceph osd crush rule rm replicated_rule
           sudo microceph.ceph osd crush rule create-replicated replicated default osd
           for flag in nosnaptrim nobackfill norebalance norecover noscrub nodeep-scrub; do
               sudo microceph.ceph osd set $flag
           done
-          sudo microceph disk add --wipe "${{ steps.free_ephemeral_disk.outputs.ephemeral_disk }}"
+
+          # If there is more than one OSD, set up partitions.
+          if [ ${{ inputs.osd-count }} -gt 1 ]; then
+            sudo blkdiscard "${ephemeral_disk}"
```
simondeziel, line 73: It looks like it needs `--force` to ignore the existing `dos` partition label. Sorry for not checking this.

.github/actions/setup-microceph/action.yml
```diff
@@ -51,17 +55,41 @@ runs:
...
+              # Create equal sized partitions for each OSD.
+              min="$(( (${i}-1) *  100 / ${{ inputs.osd-count }} ))"
+              max="$(( ${i} * 100 / ${{ inputs.osd-count }} ))"
+              sudo parted "${ephemeral_disk}" --align optimal --script mkpart primary "${min}%" "${max}%"
+
+              # MicroCeph does not accept partitions directly.
+              # See: https://github.com/canonical/microceph/issues/251
+              disk="$(losetup -f)"
+              sudo losetup --direct-io=on "${disk}" "${ephemeral_disk}${i}"
+              sudo microceph disk add "${disk}"
```
simondeziel, line 86: Removing the `--wipe` didn't make any difference, the `disk add` still takes 3-4s. I'd still keep it like that unless you think otherwise :)

.github/actions/setup-microceph/action.yml
masnax: Fixed

.github/actions/setup-microceph/action.yml
masnax: I suppose it's fine to leave it. If MicroCeph stops working due to lack of OSDs we'll know pretty quick :)

simondeziel (dismissed):

lxd/storage/drivers/driver_ceph.go
tomponline: I think its ok to leave as-is, either way its potentially out-of-date with reality, so this way its just recording what has been used which can be useful.

lxd/storage/drivers/driver_ceph.go
```diff
@@ -195,6 +195,20 @@ func (d *ceph) Create() error {
 
 		revert.Add(func() { _ = d.osdDeletePool() })
 
+		_, err = shared.TryRunCommand("ceph",
```
tomponline, line 198: Now, is this a behavioral change, what was happening before to osd pools when we didn't explicitly set the osd pool size using the default size (if not specified)?

tomponline: can we specify the pool size as part of the creation command rather than use two commands btw?

lxd/storage/drivers/driver_ceph.go
```diff
@@ -405,6 +405,23 @@ func (d *ceph) Validate(config map[string]string) error {
 
 // Update applies any driver changes required from a configuration change.
 func (d *ceph) Update(changedConfig map[string]string) error {
+	newSize := changedConfig["ceph.osd.pool_size"]
+	if newSize != d.config["ceph.osd.pool_size"] && newSize != "" {
```
tomponline, line 417: Is this `newSize != d.config["ceph.osd.pool_size"]` redundant, if they were equal it wouldnt be in changedConfig right?

tomponline: Perhaps clearer to test for `newSize, changed := changedConfig["ceph.osd.pool_size"]` instead

.github/actions/setup-microceph/action.yml
```diff
@@ -54,6 +54,7 @@ runs:
           sudo snap install microceph --channel "${{ inputs.microceph-channel }}"
           sudo microceph cluster bootstrap
           sudo microceph.ceph config set global osd_pool_default_size 1
+          sudo microceph.ceph config set global mon_allow_pool_size_one true
```
tomponline, line 61: what does this do, that `global osd_pool_default_size 1` doesnt do btw?

tomponline: looks like the previous commit https://github.com/canonical/lxd/pull/14044/commits/6a376887db8e93006ac51a88f98fdc4ba8effcf7 needs to go as its been changed here.

.github/actions/setup-microceph/action.yml
```diff
@@ -51,18 +55,41 @@ runs:
...
           sudo microceph.ceph config set global osd_memory_target 939524096
           sudo microceph.ceph osd crush rule rm replicated_rule
           sudo microceph.ceph osd crush rule create-replicated replicated default osd
           for flag in nosnaptrim nobackfill norebalance norecover noscrub nodeep-scrub; do
               sudo microceph.ceph osd set $flag
           done
-          sudo microceph disk add --wipe "${{ steps.free_ephemeral_disk.outputs.ephemeral_disk }}"
+
+          # If there is more than one OSD, set up partitions.
+          if [ ${{ inputs.osd-count }} -gt 1 ]; then
```
tomponline, line 72: does this slow down our ceph tests further by much?

test/suites/storage_driver_ceph.sh
```diff
@@ -130,6 +130,24 @@ test_storage_driver_ceph() {
     lxc storage delete "lxdtest-$(basename "${LXD_DIR}")-pool1"
     lxc storage delete "lxdtest-$(basename "${LXD_DIR}")-pool2"
     ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool rm "lxdtest-$(basename "${LXD_DIR}")-existing-osd-pool" "lxdtest-$(basename "${LXD_DIR}")-existing-osd-pool"  --yes-i-really-really-mean-it
+
+
+    # Test that the pool size setting only applies to the created pool, and not any other pools.
+    ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool set ".mgr" size 3
```
tomponline, line 136: what does this do?

test/suites/storage_driver_ceph.sh
```diff
@@ -129,7 +129,7 @@ test_storage_driver_ceph() {
     lxc profile device remove default root
     lxc storage delete "lxdtest-$(basename "${LXD_DIR}")-pool1"
     lxc storage delete "lxdtest-$(basename "${LXD_DIR}")-pool2"
-    ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool rm "lxdtest-$(basename "${LXD_DIR}")-existing-osd-pool" "lxdtest-$(basename "${LXD_DIR}")-existing-osd-pool"  --yes-i-really-really-mean-it
+    ceph --cluster "${LXD_CEPH_CLUSTER}" osd pool rm "lxdtest-$(basename "${LXD_DIR}")-existing-osd-pool" "lxdtest-$(basename "${LXD_DIR}")-existing-osd-pool"  --yes-i-really-really-mean-it || true
```
tomponline, line 132: what do you mean by

"However, because it was the last line, the error was
not propagating."

If the pool is supposed to be removed by LXD, wouldn't it be better to change the check to confirm its really gone?

lxd/storage/drivers/driver_ceph.go
tomponline, line 1: please can you re-order the commits so that `lxd/storage/drivers: Remove unnecessary fmt.Sprintfs ` occurs first before the new functionality, and then ensure the new functionality commits avoid using unnecessary fmt.Sprintfs in the first place.

test/suites/storage_driver_ceph.sh
masnax: `.mgr` is a [reserved pool](https://docs.ceph.com/en/latest/mgr/administrator/#module-pool) created by ceph. Here we're testing that only the LXD-managed pool corresponding to the config is affected by the config, and not any other pool that may happen to exist.

lxd/storage/drivers/driver_ceph.go
masnax: Before, OSD pools were given the same pool size as the default OSD pool size. This is viewable by checking the OSD pool config after pool creation. So assuming a user did not explicitly set a value for the new config key, the `osd pool set` call should not result in any change to the osd pool.

Looking at the [ceph docs](https://teams.microsoft.com/l/meetup-join/19%3ameeting_M2YyMzE5OTMtY2UzNy00MjFjLWI2NTctZmMyM2U5NTM5NGQy%40thread.v2/0?context=%7b%22Tid%22%3a%22a9a572ee-9e57-46ea-b37b-b020adfd669a%22%2c%22Oid%22%3a%221170be4e-da29-46b4-9477-7628dc06a1d4%22%7d), it doesn't appear to me that there is a way to set many config options when creating a pool

.github/actions/setup-microceph/action.yml
masnax: `osd_pool_default_size` sets the default (unspecified) pool size to 1. `mon_allow_pool_size_one` allows you to explicitly set a pool's size to 1.

lxd/storage/drivers/driver_ceph.go
masnax: Before, OSD pools were given the same pool size as the default OSD pool size. This is viewable by checking the OSD pool config after pool creation. So assuming a user did not explicitly set a value for the new config key, the `osd pool set` call should not result in any change to the osd pool.

Looking at the [ceph docs](https://docs.ceph.com/en/reef/rados/operations/pools/#creating-a-pool), it doesn't appear to me that there is a way to set many config options when creating a pool

test/suites/storage_driver_ceph.sh
masnax: I'm not that familiar with shell quirks, but for some reason when this line was the last line of the function,  its exit code being non-zero was not causing the function to exit with a non-zero exit code (despite `set -e`). 

However taking the main branch (so without the changes here) and adding any new execution following this one causes the function to exit appropriately at this line, and not reach the new line.

I'll update the line to instead check that the pool no longer exists.

test/suites/storage_driver_ceph.sh
masnax: To give some more context, we're basically setting the size of a pre-existing pool that is not managed by LXD, before creating OSD pools with LXD that have explicitly defined sizes.

This is so we can verify only the newly created pools are affected by the config key.

lxd/storage/drivers/driver_ceph.go
masnax: Ah good point, thanks!

test/suites/storage_driver_ceph.sh
masnax: So... I tested this again, and despite the fact that I remember wasting a couple hours trying to figure out what was going on with this line, it seems the command no longer returns a non-zero exit code, even when it can't find an OSD pool to remove. 

Not sure if it got fixed with a ceph version update because it certainly was returning a non-zero exit code for a while...

lxd/storage/drivers/driver_ceph.go
masnax: Fixed.

.github/actions/setup-microceph/action.yml
masnax: Commit ordering is fixed too.

test/suites/storage_driver_ceph.sh
masnax: In any case, it's fixed now. I've replaced the removal attempt with a check that the osd pool was removed with the LXD storage pool.

test/suites/storage_driver_ceph.sh
tomponline: Great, thanks, please can you add that as a comment to the test.

lxd/storage/drivers/driver_ceph.go
tomponline: > Before, OSD pools were given the same pool size as the default OSD pool size. This is viewable by checking the OSD pool config after pool creation. So assuming a user did not explicitly set a value for the new config key, the `osd pool set` call should not result in any change to the osd pool.

So no behavioral change compared with inheriting the default before?

masnax: No change if the new config has not been explicitly included in the request.

test/suites/storage_driver_ceph.sh
masnax: Done.

PR #14049, opened 2024-09-05
Fix LXD agent to work on SELinux systems (from Incus)

simondeziel:
On most Red Hat systems, you need to use /var/run/XYZ as there is an equivalency rule in SELinux that points /run to it.

Attempting to directly put a rule on /run/XYZ fails with an error:

```
ValueError: File spec /run/lxd_agent/lxd-agent conflicts with equivalency rule '/run /var/run'; Try adding '/var/run/lxd_agent/lxd-agent' instead
```

With that fix in place, `lxd-agent` is now supported/recommended on distros using SELinux.


tomponline (approved):
@simondeziel is this from an incus cherry-pick?

tomponline:
> @simondeziel is this from an incus cherry-pick?

Ah yeah it is, ill update the PR description

PR #14051, opened 2024-09-06
CI Profiling precursor changes

tomponline:
This PR:

1. Moves the building of the LXD binaries for the system tests into the code test stage and uses GH artifacts system in order to avoid building the LXD binaries multiple times for each concurrent runner.
2. Adds functionality to build LXD binaries with code coverage support and uploads them to the GH artifact system for collection later (this will come in a future PR).
3. Various Makefile improvements and cleanups.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -283,7 +283,6 @@ jobs:
         run: |
           set -eux
           sudo add-apt-repository ppa:ubuntu-lxc/daily -y --no-update
-          sudo add-apt-repository ppa:dqlite/dev -y --no-update
           sudo apt-get update
 
           sudo systemctl mask lxc.service lxc-net.service
@@ -294,7 +293,6 @@ jobs:
             libacl1-dev \
             libcap-dev \
             libdbus-1-dev \
-            libdqlite-dev \
             liblxc-dev \
             libseccomp-dev \
             libselinux-dev \
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -192,7 +192,7 @@ jobs:
       - name: Unit tests (all)
         run: |
           set -eux
-          sudo --preserve-env=CGO_CFLAGS,CGO_LDFLAGS,CGO_LDFLAGS_ALLOW,LD_LIBRARY_PATH LD_LIBRARY_PATH=${LD_LIBRARY_PATH} env "PATH=${PATH}" go test -v ./...
+          sudo --preserve-env=CGO_CFLAGS,CGO_LDFLAGS,CGO_LDFLAGS_ALLOW,LD_LIBRARY_PATH LD_LIBRARY_PATH=${LD_LIBRARY_PATH} env "PATH=${PATH}" make check-unit
 
       - name: Upload system test dependencies
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -242,20 +289,10 @@ jobs:
...
-          go mod tidy -go="${GOMIN}"
-
-          DOC_GOMIN="$(sed -n 's/^LXD requires Go \([0-9.]\+\) .*/\1/p' doc/requirements.md)"
-          [ "${GOMIN}" = "${DOC_GOMIN}" ]
-
       - name: Install dependencies
         run: |
           set -eux
           sudo add-apt-repository ppa:ubuntu-lxc/daily -y --no-update
-          sudo add-apt-repository ppa:dqlite/dev -y --no-update
```
simondeziel, line 258: If we are not compiling LXD here, do we need the `ubuntu-lxc/daily` PPA and the `liblxc-dev` and other `-dev` packages?

.github/workflows/tests.yml
```diff
@@ -391,8 +421,15 @@ jobs:
           set -eux
           chmod +x ~
           echo "root:1000000:1000000000" | sudo tee /etc/subuid /etc/subgid
+          mkdir -p "${GOCOVERDIR}"
           cd test
-          sudo --preserve-env=PATH,GOPATH,GITHUB_ACTIONS,LXD_VERBOSE,LXD_BACKEND,LXD_CEPH_CLUSTER,LXD_CEPH_CEPHFS,LXD_CEPH_CEPHOBJECT_RADOSGW,LXD_OFFLINE,LXD_SKIP_TESTS,LXD_REQUIRED_TESTS, LXD_BACKEND=${{ matrix.backend }} ./main.sh ${{ matrix.suite }}
+          sudo --preserve-env=PATH,GOPATH,GOCOVERDIR,GITHUB_ACTIONS,LXD_VERBOSE,LXD_BACKEND,LXD_CEPH_CLUSTER,LXD_CEPH_CEPHFS,LXD_CEPH_CEPHOBJECT_RADOSGW,LXD_OFFLINE,LXD_SKIP_TESTS,LXD_REQUIRED_TESTS, LXD_BACKEND=${{ matrix.backend }} ./main.sh ${{ matrix.suite }}
```
simondeziel, line 426: In the `System` tests, do we still need Go tooling? Maybe to assist in collecting the coverage info?

simondeziel (approved):
LGTM thanks! There might be some things to further remove but those can wait till later.

One thing we lose with that is we are now only testing/building on Go 1.22. That said, maybe you wanted to have that done first and then we add another step to the `Code` job later on.

.github/workflows/tests.yml
tomponline: for now we do as LXD links against it dynamically

.github/workflows/tests.yml
tomponline: yeah we do for now at least because there are several go programs built during the tests to help run them

tomponline:
> One thing we lose with that is we are now only testing/building on Go 1.22. That said, maybe you wanted to have that done first and then we add another step to the `Code` job later on.

Note, thanks for highlighting.

PR #14052, opened 2024-09-06
api: Add `metrics_api_requests` extension.

markylaing:
NOTE: This change would normally be committed with an update to `doc/api-extensions.md` (582b2ccddeebbfa17df7becfd3979d4f5438a0a1) but this was missed. These commits need to be cherry-picked together.


tomponline (approved):
Good spot!

PR #14053, opened 2024-09-06
Fix: Also update HMAC key in case of argon2

roosterfish:
Due to the test cases having an expected salt configured for the argon2 tests, I didn't catch that when parsing an HTTP Authorization header, besides updating the structs salt we also have to rewrite the actual HMAC key as this one is tightly coupled to the salt.

If you initialize a new argon2 based HMAC using `hmac := trust.NewHMACArgon2()` you can provide a salt or leave it blank.
In a real world scenario when running the HMAC validation as part of some middelware, you cannot set a salt when initializing as this information gets populated later when you run `hmac.ParseHTTPHeader()`.

To fix this `hmac.ParseHTTPHeader()` now also rewrites the internal `key`.
The argon2 `IDKey` func is moved in its own unexported function so we can run it both when initializing the struct and when parsing an header.

To prevent further misses in this field, I have also removed the salt field from the validation test cases as it's not required at the end.


shared/trust/hmac_argon2.go
```diff
@@ -84,8 +91,12 @@ func (h *HMACArgon2) ParseHTTPHeader(header string) (HMACVersion, []byte, error)
 		return "", nil, fmt.Errorf("Failed to decode the argon2 salt: %w", err)
 	}
 
+	// Update the salt so we can properly render the header.
 	h.salt = salt
 
+	// Also update the overall HMAC key as it's based on the salt in case of argon2.
+	h.HMAC.key = idKey(h.password, salt)
```
tomponline, line 98: This is starting to look like "magic" again.

If I'm following this right a user can potentially call ` NewHMACArgon2` passing in a salt as an argument, and then call h.ParseHTTPHeader() (whose docblock only mentions "extract"...) which then modifies the salt I set as the caller to something else.

So the usability issue here is that one has to understand the relationship between NewHMACArgon2 and ParseHTTPHeader and know that `salt` argument is ignored when using `ParseHTTPHeader`.

Instead of this, why don't we have a `NewHMACArgon2FromHTTPHeader()` function that returns a formatter based on the provided header?

roosterfish: > a user can potentially call  NewHMACArgon2 passing in a salt as an argument, and then call h.ParseHTTPHeader() (whose docblock only mentions "extract"...) which then modifies the salt I set as the caller to something else.

That's right.

Thanks for the suggestion. I'll move it to a `NewHMACArgon2FromHTTPHeader` as this would also remove the "overwriting" of the salt in case you run `ParseHTTPHeader`.

I know we have discussed having a `NewFromHeader` and I have evaluated options that allow you to parse the right implementation (e.g. argon2) depending on the header but those implementations all didn't look straightforward as I tried to have a generic method that allows returning various formats.

However I haven't evaluated having a dedicated `NewArgon2FromHeader` which makes sense as it already implies the used format.

roosterfish: I have pushed the necessary changes and can now give you some reasoning why having a `NewHMACArgon2FromHTTPHeader` isn't straightforward.

Let's say you have a `NewHMACArgon2FromHTTPHeader(password string, header string)`. 
We can totally parse the version and salt from the header and populate the internal structs fields together with the password.

However as soon as you want to start validating if the header is what you expect, you have to compare it's version against something.
This something can be either a new instance (created using `NewHMACArgon2`) or a direct comparison against a variable of type `HMACVersion`.

Whilst this still seems feasible, the actual HMAC comparison now becomes interesting.
To not require parsing the header again `NewHMACArgon2FromHTTPHeader` can also return the actual HMAC found in the header.
Now you have to call any of the `Write*` funcs (ideally `WriteRequest`) to calculate the HMAC over the request's body.
Now you have both and can compare them.

But as we have both the default and argon2 formatters, this would mean lots of duplicated code for the verification as both need to call their own implementation of `NewHMAC...FromHTTPHeader`. 
This way I have settled with moving the idea of a `NewHMAC..FromHTTPHeader` into the already existing `ParseHTTPHeader`.
It now returns a completely new instance derived from the header using the password/key and version which was already configured for the parent instance.

There is an updated example in the PR's desc :)

roosterfish: The tl;dr is we don't anymore require to rewrite the salt and/or password of an already existing instance of HMAC.

tomponline (approved):

PR #14055, opened 2024-09-06
Completions: Fix completions for config keys

kadinsayani:
`cmpInstanceAllKeys()` was only grabbing config keys that apply to any instance, but we also want to grab/complete keys that are specific to containers or VM's.


tomponline (commented):
Has there been a later commit in incus we can cherry pick for this?

kadinsayani:
> Has there been a later commit in incus we can cherry pick for this?

From what I can tell, no. Incus introduced completions for config back in February and the corresponding code hasn't been updated since then.

tomponline (approved):
Thanks!

PR #14056, opened 2024-09-06
Request: Fix request not counted warning

hamistao:
This makes it so that we don't expect the metrics callback to be present on devlxd endpoints and neither on the `/` endpoint.

It does so by reworkin how we check for requests originating from `/dev/lxd`. Instead of  setting the request remote address, this sets a newly introduced authentication method on the request context for all devlxd requests and uses that to validate devlxd requests. So this makes it possible to avoid logging a warning for devlxd requests when they are not supposed to be tracked for the API metrics. Thanks for the suggestion @markylaing :D.

### DIFF
--- a/lxd/auth/types.go
+++ b/lxd/auth/types.go
@@ -26,6 +26,9 @@ const (
 	// Note: Regardless of whether `core.trust_ca_certificates` is enabled, we still check if the client certificate
 	// fingerprint is in the identity cache. If they are found, standard TLS restrictions will apply.
 	AuthenticationMethodPKI string = "pki"
+
+	// AuthenticationMethodDevLXD is the authentication method for interacting with the devlxd API.
+	AuthenticationMethodDevLXD = "devlxd"
 )
 
 // PermissionChecker is a type alias for a function that returns whether a user has required permissions on an object.
--- a/lxd/daemon.go
+++ b/lxd/daemon.go
@@ -387,11 +387,6 @@ func (d *Daemon) Authenticate(w http.ResponseWriter, r *http.Request) (trusted b
 		return true, "", auth.AuthenticationMethodUnix, nil, nil
 	}
 
-	// Devlxd unix socket credentials on main API.
-	if r.RemoteAddr == devlxdRemoteAddress {
-		return false, "", "", nil, fmt.Errorf("Main API query can't come from /dev/lxd socket")
-	}
-
 	// Cluster notification with wrong certificate.
 	if isClusterNotification(r) {
 		return false, "", "", nil, fmt.Errorf("Cluster notification isn't using trusted server certificate")
--- a/lxd/daemon.go
+++ b/lxd/daemon.go
@@ -53,6 +53,7 @@ import (
 	"github.com/canonical/lxd/lxd/lifecycle"
 	"github.com/canonical/lxd/lxd/loki"
 	"github.com/canonical/lxd/lxd/maas"
+	"github.com/canonical/lxd/lxd/metrics"
 	networkZone "github.com/canonical/lxd/lxd/network/zone"
 	"github.com/canonical/lxd/lxd/node"
 	"github.com/canonical/lxd/lxd/request"
@@ -630,7 +631,7 @@ func (d *Daemon) createCmd(restAPI *mux.Router, version string, c APIEndpoint) {
 		// Only endpoints from the main API (version 1.0) should be counted for the metrics.
 		// This prevents internal endpoints from being included as well.
 		if version == "1.0" {
-			request.CountStartedRequest(r)
+			metrics.TrackStartedRequest(r)
 		}
 
 		w.Header().Set("Content-Type", "application/json")
--- a/lxd/request/request.go
+++ b/lxd/request/request.go
@@ -4,12 +4,8 @@ import (
 	"context"
 	"net"
 	"net/http"
-	"strings"
-	"sync"
 
-	"github.com/canonical/lxd/lxd/metrics"
 	"github.com/canonical/lxd/shared/api"
-	"github.com/canonical/lxd/shared/logger"
 )
 
 // CreateRequestor extracts the lifecycle event requestor data from an http.Request context.
@@ -53,35 +49,4 @@ func CreateRequestor(r *http.Request) *api.EventLifecycleRequestor {
 // in the request context for later use.
 func SaveConnectionInContext(ctx context.Context, connection net.Conn) context.Context {
 	return context.WithValue(ctx, CtxConn, connection)
-}
-
-// CountStartedRequest tracks the request as started for the API metrics and
-// injects a callback function to track the request as completed.
-func CountStartedRequest(r *http.Request) {
-	requestURL := *r.URL
-
-	// Set the callback function to track the request as completed.
-	// Use sync.Once to ensure it can be called at most once.
-	var once sync.Once
-	callbackFunc := func(result metrics.RequestResult) {
-		once.Do(func() {
-			metrics.TrackCompletedRequest(requestURL, result)
-		})
-	}
-
-	SetCtxValue(r, MetricsCallbackFunc, callbackFunc)
-
-	metrics.TrackStartedRequest(requestURL)
-}
-
-// MetricsCallback retrieves a callback function from the request context and calls it.
-// The callback function is used to mark the request as completed for the API metrics.
-func MetricsCallback(request *http.Request, result metrics.RequestResult) {
-	callback, err := GetCtxValue[func(metrics.RequestResult)](request.Context(), MetricsCallbackFunc)
-	if err != nil && (strings.HasPrefix(request.URL.Path, "/1.0") || request.URL.Path == "/") {
-		// Log a warning if endpoint is part of the main API, and therefore should be counted fot the API metrics.
-		logger.Warn("Request will not be counted for the API metrics", logger.Ctx{"url": request.URL.Path, "method": request.Method, "remote": request.RemoteAddr})
-	} else if err == nil && callback != nil {
-		callback(result)
-	}
 }
--- a/test/suites/devlxd.sh
+++ b/test/suites/devlxd.sh
@@ -1,5 +1,9 @@
 test_devlxd() {
   ensure_import_testimage
+  fingerprint="$(lxc image info testimage | awk '/^Fingerprint:/ {print $2}')"
+
+  # Ensure testimage is not set as cached.
+  lxd sql global "UPDATE images SET cached=0 WHERE fingerprint=\"${fingerprint}\""
 
   (
     cd devlxd-client || return
@@ -13,6 +17,16 @@ test_devlxd() {
   lxc config unset devlxd security.devlxd
   lxc exec devlxd -- test -S /dev/lxd/sock
   lxc file push --mode 0755 "devlxd-client/devlxd-client" devlxd/bin/
+
+  # Try to get a host's private image from devlxd.
+  [ "$(lxc exec devlxd -- devlxd-client image-export "${fingerprint}")" = "not authorized" ]
+  lxc config set devlxd security.devlxd.images true
+  # Trying to get a private image should return a not found error so that the client can't infer the existence
+  # of an image with the provided fingerprint.
+  [ "$(lxc exec devlxd -- devlxd-client image-export "${fingerprint}" | jq -r '.error')" = "not found" ]
+  lxd sql global "UPDATE images SET cached=1 WHERE fingerprint=\"${fingerprint}\""
+  # No output means the export succeeded.
+  [ -z "$(lxc exec devlxd -- devlxd-client image-export "${fingerprint}")" ]
 
   lxc config set devlxd user.foo bar
   [ "$(lxc exec devlxd -- devlxd-client user.foo)" = "bar" ]
--- a/lxd/api.go
+++ b/lxd/api.go
@@ -205,7 +205,7 @@ func restServer(d *Daemon) *http.Server {
 	}
 
 	mux.NotFoundHandler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		lxdRequest.CountStartedRequest(r)
+		metrics.TrackStartedRequest(r)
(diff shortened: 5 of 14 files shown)


hamistao:
@tomponline This fixes the incorrect logging on devlxd endpoints.

@markylaing suggested checking the `RemoteAddr` field as being `@devlxd` to see if the request is coming from /dev/lxd. Unfortunately I wasn't able to make this work as the `RemoteAddr` value followed the pattern of `vm(104676174):2672481575`. I chose to check the `Host` field instead, if you agree on this approach this should be ready to merge.

lxd/request/request.go
```diff
@@ -78,7 +78,8 @@ func CountStartedRequest(r *http.Request) {
 // The callback function is used to mark the request as completed for the API metrics.
 func MetricsCallback(request *http.Request, result metrics.RequestResult) {
 	callback, err := GetCtxValue[func(metrics.RequestResult)](request.Context(), MetricsCallbackFunc)
-	if err != nil && (strings.HasPrefix(request.URL.Path, "/1.0") || request.URL.Path == "/") {
+	// If host is a custom socket the request is coming from the /dev/lxd socket.
+	if err != nil && strings.HasPrefix(request.URL.Path, "/1.0") && request.Host != "custom.socket" {
```
simondeziel, line 82: AFAIK, this `custom.socket` host can be basically anything:

```
$ lxc exec c1 -- curl -s --unix-socket /dev/lxd/sock http://custom.socket/1.0
{"state":"Started","api_version":"1.0","instance_type":"container","location":"sdeziel-lemur"}
$ lxc exec c1 -- curl -s --unix-socket /dev/lxd/sock http://foo.bar/1.0
{"state":"Started","api_version":"1.0","instance_type":"container","location":"sdeziel-lemur"}
```

I think @markylaing suggested checking for `@devlxd` somewhere in the request struct(?).

hamistao: Yes, but it unfortunately did not work :(

The `RemoteAddr` value followed the pattern of `vm(104676174):2672481575` instead of being `@devlxd`.

I was afraid checking for `custom.socket` could be too broad, but I think the alternative would be using a regex to check if `RemoteAddr` follows the expected pattern fot vm remotes. Or remove this log message entirely.

simondeziel: `vm(104676174):2672481575` looks like a `vsock` address so maybe that is a good way to distinguish them for the VM case? The `@devlxd` request address might be just for containers?

tomponline: @hamistao do you know why this function is being called at all for devlxd?

lxd/request/request.go
```diff
@@ -78,7 +78,8 @@ func CountStartedRequest(r *http.Request) {
...
 		// Log a warning if endpoint is part of the main API, and therefore should be counted fot the API metrics.
 		logger.Warn("Request will not be counted for the API metrics", logger.Ctx{"url": request.URL.Path, "method": request.Method, "remote": request.RemoteAddr})
```
simondeziel, line 84: This feels like debug logging to me. Maybe it's not even needed as you suggested @hamistao.
That said, it seems to have brought the question of why `/dev/lxd` requests ended up in that function.

lxd/request/request.go
hamistao: Because it uses some of the same `Response` types as the regular API. See [this](https://github.com/hamistao/lxd/blob/improve_reboot_experience/lxd/response/response.go#L94) line.

lxd/request/request.go
hamistao: I think a warning maybe makes more sense since a sys admin may want to know that some requests are not being counted as expected as this should help understand inconsistencies in the metrics values. That said I am fine with removing it if we are unable to properly check for devlxd requests.

lxd/request/request.go
hamistao: > The @devlxd request address might be just for containers

If this is the case, should we worry the existing checks for `@devlxd` may be flawed?

> vm(104676174):2672481575 looks like a vsock address so maybe that is a good way to distinguish them for the VM case

Cool, that works for me as well :D

lxd/request/request.go
tomponline: The warning is incorrect in that it is logging for a request that it shouldnt be tracking.
So we need to fix the logic that causes this to be logged on every request, but we can leave the general idea of logging a warning if it occurs for another unexpected reason.

lxd/request/request.go
tomponline: It seems to be like in a single API handler for devlxd we override the `RemoteAddr` with `devlxdRemoteAddress` and for all the others we do not do that.

Then there are checks for `RemoteAddr` being `devlxdRemoteAddress` to prevent access to things or to classify the request and thus this isn't working correctly for all devlxd endpoints.

I also don't really like us overwriting the `RemoteAddr` with `devlxdRemoteAddress` because we lose the original remote identifier.

Instead I think instead of this we should remove `devlxdRemoteAddress` and instead have the devlxd API server set something into the request context to indicate its a devlxd request.

We already have `request.CtxProtocol` which is used to indicate the authentication method, perhaps this would be an appropriate place to have "devlxd" as another authentication method. 

Then we can update the places that check for `devlxdRemoteAddress`  to check for `request.CtxProtocol` for as "devlxd".

What do you think @markylaing ?

hamistao: @tomponline
If @markylaing agrees, I can implement your suggestion as a part of PR since it will help with the problem at hand as well.

markylaing: > Instead I think instead of this we should remove devlxdRemoteAddress and instead have the devlxd API server set something into the request context to indicate its a devlxd request.

Yep this sounds better than setting the `RemoteAddr`. @hamistao you can add the new authentication method under `lxd/auth/types.go`.

markylaing: That being said. If we don't want these to count towards metrics at all... would it be better to just delete the metrics callback in the devlxd request hoist function?

hamistao: > you can add the new authentication method under lxd/auth/types.go

So this authentication method would be just checking the request context for whether the request is a devlxd request?

> Or better: don't set it when setting up the devlxd handlers.

It is not being set up, this log message is triggered by there being no callback function in the request context.

markylaing: > It is not being set up, this log message is triggered by there being no callback function in the request context.

Ah that makes sense, sorry my misunderstanding.

> So this authentication method would be just checking the request context for whether the request is a devlxd request?

Sorry this is overloading the term "method". The authentication methods are string constants: https://github.com/canonical/lxd/blob/dbefa5e712d5a026893957e79b277485b770eb20/lxd/auth/types.go#L11-L29 

more are defined in the `api` package as well: https://github.com/canonical/lxd/blob/dbefa5e712d5a026893957e79b277485b770eb20/shared/api/auth.go#L3-L9

but we'll keep the `devlxd` authentication method under `lxd/auth` because it's internal to LXD.

Currently, setting `r.RemoteAddr` is only being performed in the devlxd image export handler to distinguish between devlxd and main API queries in the main image export handler. We should:
1. Remove `devlxdRemoteAddress`.
2. Remove setting `r.RemoteAddr` in the devlxd image export handler.
3. Add a new authentication method constant for devlxd in `lxd/auth/types.go` with key `request.CtxProtocol`.
4. Set the authentication method in the request context in devlxd's `hoistReq` function.
5. Where we previously checked `r.RemoteAddr`, now call `(lxd/auth).GetAuthenticationMethodFromCtx` to get the authentication method.
6. Skip logging request not counted based on the value returned by `(lxd/auth).GetAuthenticationMethodFromCtx`.

hamistao:
@tomponline @markylaing This should be ready for a review.

tomponline (dismissed):
Thanks!

Do we have tests that check there haven't been any access regressions, i.e /dev/lxd endpoint can make requests to the main api?

shared/api/auth.go
```diff
@@ -6,6 +6,9 @@ const (
 
 	// AuthenticationMethodOIDC is a token based authentication method.
 	AuthenticationMethodOIDC = "oidc"
+
+	// AuthenticationMethodDevLxd is the authentication method for interacting with the devlxd API..
+	AuthenticationMethodDevLxd = "devlxd"
```
markylaing, line 11: This is an internal authentication method, so it should be in `lxd/auth/types.go`

markylaing (dismissed):

lxd/daemon.go
```diff
@@ -388,7 +388,8 @@ func (d *Daemon) Authenticate(w http.ResponseWriter, r *http.Request) (trusted b
 	}
 
 	// Devlxd unix socket credentials on main API.
-	if r.RemoteAddr == devlxdRemoteAddress {
+	authMethod, _ := auth.GetAuthenticationMethodFromCtx(r.Context())
+	if authMethod == api.AuthenticationMethodDevLxd {
```
markylaing, line 392: I wonder if it might be worth adding something like:
```golang
// lxd/auth/util.go
func IsAuthenticationMethod(ctx context.Context, method string) (bool, error) {
    actualMethod, err := GetAuthenticationMethodFromCtx(ctx)
    if err != nil {
        return false, err
    }

    return method == actualMethod
}
```

lxd/request/request.go
```diff
@@ -78,7 +78,10 @@ func CountStartedRequest(r *http.Request) {
 // The callback function is used to mark the request as completed for the API metrics.
 func MetricsCallback(request *http.Request, result metrics.RequestResult) {
 	callback, err := GetCtxValue[func(metrics.RequestResult)](request.Context(), MetricsCallbackFunc)
-	if err != nil && (strings.HasPrefix(request.URL.Path, "/1.0") || request.URL.Path == "/") {
+
+	// Verify the auth method in the request context to determine if the request comes from the /dev/lxd socket.
+	authMethod, _ := GetCtxValue[string](request.Context(), CtxProtocol)
+	if err != nil && strings.HasPrefix(request.URL.Path, "/1.0") && authMethod != api.AuthenticationMethodDevLxd {
```
markylaing, line 84: Once the authentication method is moved to `lxd/auth/types.go` the import cycle will be unavoidable here. I suggest moving the callback to `lxd/metrics.go`

lxd/daemon.go
hamistao: I was thinking the same, but based on the current usage I don't think this does much in making code more readable nor shorter.

hamistao:
> Do we have tests that check there haven't been any access regressions, i.e /dev/lxd endpoint can make requests to the main api?

@tomponline I looked for it and I think we don't, if needed I can add them on LXD CI on `devlxd-container`.

tomponline:
> > Do we have tests that check there haven't been any access regressions, i.e /dev/lxd endpoint can make requests to the main api?
> 
> @tomponline I looked for it and I think we don't, if needed I can add them on LXD CI on `devlxd-container`.

Can we add them in the main test suite for containers so they run on PRs too.

hamistao:
@tomponline Sorry but I may have misinterpreted the tests you wanted here, the server that listens over /dev/lxd do not even handle the main API endpoints so I just get a 404 when trying `lxc exec c -- curl -s --unix-socket /dev/lxd/sock http://custom.socket/1.0/metrics` for example. If that is what you intended and want these tests included please let me know, if not please elaborate so I can try to better understand.

tomponline:
> @tomponline Sorry but I may have misinterpreted the tests you wanted here, the server that listens over /dev/lxd do not even handle the main API endpoints so I just get a 404 when trying `lxc exec c -- curl -s --unix-socket /dev/lxd/sock http://custom.socket/1.0/metrics` for example. If that is what you intended and want these tests included please let me know, if not please elaborate so I can try to better understand.

It was really asking if we have a test for this check being effective:

https://github.com/canonical/lxd/pull/14056/commits/9b6d51901353e88d5927fb43f38e42b15a3945e9#diff-f3756e85146e37cc40847e0c6b78fe5fdc2a4775a9a492ae08f05860f01c97d1R4119

hamistao:
@tomponline Oh I see, so we already have them on [`tests/devlxd-container`](https://github.com/canonical/lxd-ci/blob/main/tests/devlxd-container#L107-L119) on lxd-ci. I am guessing they were not added in the main test suite because it uses a nested LXD so maybe it is slower that your average test. If think they are needed here, we can move/copy all the tests in that file to the main test suite.

tomponline:
@hamistao how is this coming along, ready for review?

hamistao:
@tomponline @markylaing This is ready for another review.
Should be a quick review inspite of the number of commits, since they are all small and/or trivial to understand.

tomponline:
tests failing

hamistao:
@tomponline @markylaing Sorry for the delay, the tests are now fixed and passing. Feel free to review this when possible. 
 As mentioned before, should be a quick review inspite of the number of commits, since they are all small and/or trivial to understand. Thanks!

lxd/auth/types.go
```diff
@@ -26,6 +26,9 @@ const (
 	// Note: Regardless of whether `core.trust_ca_certificates` is enabled, we still check if the client certificate
 	// fingerprint is in the identity cache. If they are found, standard TLS restrictions will apply.
 	AuthenticationMethodPKI string = "pki"
+
+	// AuthenticationMethodDevLxd is the authentication method for interacting with the devlxd API..
+	AuthenticationMethodDevLxd = "devlxd"
```
tomponline, line 31: Lxd should be LXD please

test/suites/devlxd.sh
```diff
@@ -1,4 +1,8 @@
 test_devlxd() {
+  if [ "$(lxc image info testimage | awk '/^Cached:/ {print $2}')" = "yes" ]; then
+    echo "==> Deleting cached testimage to ensure testimage will not be accessible through devlxd"
+    lxc image delete testimage
+  fi
   ensure_import_testimage
```
tomponline, line 6: wont this just reimport it here again?

tomponline: if there is a cached image available could we test that /dev/lxd can access it before we remove it?

hamistao: The cached image is one that lingers from a previous test.
The one imported by `ensure_import_testimage` is not cached.

hamistao:
@tomponline changes made and tests passing finally!

tomponline:
Thanks!

PR #14057, opened 2024-09-06
doc/authentication: clarify what `{client|server}.ca` are

simondeziel:


doc/authentication.md
```diff
@@ -115,7 +115,9 @@ The steps for enabling PKI mode differ slightly depending on whether you use an
 
 `````{tabs}
 ````{group-tab} Only PKI
-If you use a PKI system, both the server certificates and the client certificates are issued by a secondary CA.
+If you use a PKI system, both the server certificates and the client certificates are issued by intermediate CA(s). The `client.ca` file contains the certificate used by the client to verify the trust of the server certificate it receives when making a connection to a remote. The `server.ca` file contains the certificate used by the server to verify the trust of the client certificates associated with an incoming connection.
+
+Both files contain trust anchors used to evaluate if the received leaf certificate from the other end of the connection is to be trusted or not. If the leaf certificate's chain of trust leads to one of the trusted anchors it will be trusted (unless revoked).
```
markylaing, line 120: I think some general advice we had from Ruth was to have a sentence per line. So I would suggest (with some additional edits):
```
If you use a PKI system, both the server and client certificates are issued by intermediate CA(s).
The `client.ca` file contains the certificate used by the client to verify the server certificate it receives when making a connection to a remote. 
The `server.ca` file contains the certificate used by the server to verify the client certificates associated with an incoming connection.

Both files contain trust anchors used to evaluate if the received leaf certificate from the other end of the connection is to be trusted or not.
If the leaf certificate's chain of trust leads to one of the trusted anchors it will be trusted (unless revoked).
```

simondeziel: Thanks @markylaing and @ru-fu that makes a lot of sense when we think of minimizing the `diff` context. Fixed

markylaing (approved):
LGTM thanks :)

PR #14059, opened 2024-09-06
Auth: Remove certificate entity type

### LINKED ISSUE #13372: Potentially redundant entity types for permissions
# Required information

- Distribution: 5.21 LTS
- Distribution version: snap

# Issue description

Currently in lxd the `identity` and `certificate` entity type from the authorisation model may relate to the same entities. An `identity` entity refers to oidc and tls entities and a `certificate` entity refers to tls entities only. Therefore, certificate entities is a subset of identity entities. Permissions can be created for both entity types currently, but ideally we should consider combining both entity types in the authorisation model to minimise the possibility of conflicting permissions.

markylaing:
Removes the "certificate" entity type from the authorization model in favour of using "identity". This is because identities are a superset of certificates, and certificates are contained within the identities table.

Closes #13372


tomponline (changes requested):
Please can we add a test for " lxd/auth: Update the TLS driver to handle identity URLs. "

lxd/patches.go
```diff
@@ -1420,4 +1421,34 @@ func patchInstanceRemoveVolatileLastStateIPAddresses(_ string, d *Daemon) error
...
+		// Even though these reference the same row in the identities table, the permissions are different because the
+		// entity type codes are different. Updating the code of the certificate permission would then cause a uniqueness
+		// constraint violation.
+		_, err := tx.Tx().Exec(`UPDATE OR REPLACE auth_groups_permissions SET entity_type = 24 WHERE entity_type = 4`)
+		if err != nil {
+			return fmt.Errorf("Failed replacing certificate entities with identity entities in permissions table: %w", err)
+		}
+
+		// Not using "OR REPLACE" here because warnings do not have a uniqueness constraint applied to the entity_type_code column.
+		_, err = tx.Tx().Exec(`UPDATE warnings SET entity_type_code = 24 WHERE entity_type_code = 4`)
```
hamistao, line 1440: So this could leave two identical warnings?

lxd/auth/drivers/tls.go
```diff
@@ -161,6 +169,13 @@ func (t *tls) GetPermissionChecker(ctx context.Context, entitlement auth.Entitle
 			return false
 		}
 
+		// Allow identities (certificates if using this authorizer) to view, edit, delete themselves.
+		// Allow identities (certificates if using this authorizer) to view themselves.
```
hamistao, line 173: I am a bit confused by these two similar comments. In my understanding just the second line is needed here since the permissions for edition and deletion are explained below, what do you think?

markylaing: Ah whoops I didn't mean to leave the first comment in there!

markylaing:
> Please can we add a test for " lxd/auth: Update the TLS driver to handle identity URLs. "

If it's ok with you I'll include #14061 in this PR and add tests for that

tomponline:
> > Please can we add a test for " lxd/auth: Update the TLS driver to handle identity URLs. "
> 
> If it's ok with you I'll include #14061 in this PR and add tests for that

Ah sounds good!

lxd/patches.go
markylaing: I'm not aware of any warnings being introduced for the `identity` entity type. So any existing warnings for certificates will be moved there.

hamistao: Oh cool then, thanks for the info

markylaing:
Closing. We have decided to keep the certificate entity type for existing certificates.

PR #14060, opened 2024-09-06
Github: Align codesql and commits workflow with test workflows for concurrency and and "on" rules

tomponline:


simondeziel (approved):
Thanks!

tomponline (commented):
Seems to have upset the push commits check https://github.com/canonical/lxd/actions/runs/10749297801

.github/workflows/commits.yml
```diff
@@ -1,6 +1,14 @@
 name: Commits
 on:
-  - pull_request
+  push:
```
tomponline, line 3: Ah it should not run on push at all ill correct

PR #14061, opened 2024-09-06
Auth: Defensive authorization in TLS driver for non project specific entities

markylaing:
While working on #14059 I noticed that the error message that a restricted TLS client receives when attempting to view a group is
`User does not have permission for project ""`. This was because the entity.TypeAuthGroup was not being handled appropriately in the driver. If somehow a project query parameter was set on a group URL in an auth check, it would be possible to expose groups to restricted users.

Opening as draft as #14059 must be merged first.


tomponline:
What happened here?

markylaing:
> What happened here?

I'll add this to #14059 and add tests

Issue #14001, opened 2024-08-28
Ability to launch armhf VMs on arm64 hosts


panlinux:
# Required information

 * Distribution: Ubuntu
 * Distribution version: 24.04.1
 * The output of "snap list --all lxd core20 core22 core24 snapd":
```
Name    Version      Rev    Tracking       Publisher   Notes
core20  20240227     2267   latest/stable  canonical✓  base,disabled
core20  20240416     2321   latest/stable  canonical✓  base
core22  20240731     1568   latest/stable  canonical✓  base,disabled
core22  20240809     1590   latest/stable  canonical✓  base
lxd     6.1-4cd98ce  29733  latest/stable  canonical✓  disabled
lxd     6.1-efad198  29946  latest/stable  canonical✓  -
snapd   2.62         21467  latest/stable  canonical✓  snapd,disabled
snapd   2.63         21761  latest/stable  canonical✓  snapd
```
 * The output of "lxc info" or if that fails: not relevant, since this is a feature request I believe, but let me know if you still want it. It's long.

# Issue description
I would like to be able to launch armhf VMs on my arm64 host. I can launch armhf containers on that host already, but sometimes I need a VM.

Currently I get an error:
```
andreas@pi4:~$ lxc launch ubuntu-daily:jammy/armhf j-armhf-vm --vm
Creating j-armhf-vm
Starting j-armhf-vm
Error: Architecture isn't supported for virtual machines
Try `lxc info --show-log local:j-armhf-vm` for more info

andreas@pi4:~$ lxc list j-armhf-vm
+------------+---------+------+------+-----------------+-----------+
|    NAME    |  STATE  | IPV4 | IPV6 |      TYPE       | SNAPSHOTS |
+------------+---------+------+------+-----------------+-----------+
| j-armhf-vm | STOPPED |      |      | VIRTUAL-MACHINE | 0         |
+------------+---------+------+------+-----------------+-----------+
```
A container is launched just fine:
```
andreas@pi4:~$ lxc launch ubuntu-daily:jammy/armhf j-armhf-container 
Creating j-armhf-container
Starting j-armhf-container
andreas@pi4:~$ lxc list j-armhf-container
+-------------------+---------+------+------+-----------+-----------+
|       NAME        |  STATE  | IPV4 | IPV6 |   TYPE    | SNAPSHOTS |
+-------------------+---------+------+------+-----------+-----------+
| j-armhf-container | RUNNING |      |      | CONTAINER | 0         |
[...]

PR #14077, opened 2024-09-09
lxd/instance: Add support for armhf vm's on arm64 hosts

kadinsayani:
This improvement will resolve https://github.com/canonical/lxd/issues/14001 with the inclusion of an armhf 32-bit lxd-agent binary in the lxd snap.

### DIFF
--- a/lxd/instance/drivers/driver_qemu.go
+++ b/lxd/instance/drivers/driver_qemu.go
@@ -2007,7 +2007,7 @@ func (d *qemu) qemuArchConfig(arch int) (path string, bus string, err error) {
 		}
 
 		return path, "pcie", nil
-	} else if arch == osarch.ARCH_64BIT_ARMV8_LITTLE_ENDIAN {
+	} else if arch == osarch.ARCH_32BIT_ARMV7_LITTLE_ENDIAN || arch == osarch.ARCH_32BIT_ARMV8_LITTLE_ENDIAN || arch == osarch.ARCH_64BIT_ARMV8_LITTLE_ENDIAN {
 		path, err := exec.LookPath(basePath + "qemu-system-aarch64")
 		if err != nil {
 			return "", "", err
--- a/lxd/instance/drivers/driver_qemu_templates.go
+++ b/lxd/instance/drivers/driver_qemu_templates.go
@@ -74,7 +74,7 @@ func qemuMachineType(architecture int) string {
 	switch architecture {
 	case osarch.ARCH_64BIT_INTEL_X86:
 		machineType = "q35"
-	case osarch.ARCH_64BIT_ARMV8_LITTLE_ENDIAN:
+	case osarch.ARCH_32BIT_ARMV7_LITTLE_ENDIAN, osarch.ARCH_32BIT_ARMV8_LITTLE_ENDIAN, osarch.ARCH_64BIT_ARMV8_LITTLE_ENDIAN:
 		machineType = "virt"
 	case osarch.ARCH_64BIT_POWERPC_LITTLE_ENDIAN:
 		machineType = "pseries"


simondeziel (dismissed):
LGTM but should we rework those `else if` to use `or` comparison and avoid dup'ing the `if's` body? The aim would be to reduce the code duplication.

lxd/instance/drivers/driver_qemu_templates.go
```diff
@@ -74,6 +74,10 @@ func qemuMachineType(architecture int) string {
 	switch architecture {
 	case osarch.ARCH_64BIT_INTEL_X86:
 		machineType = "q35"
+	case osarch.ARCH_32BIT_ARMV7_LITTLE_ENDIAN:
```
simondeziel, line 77:
```suggestion
	case osarch.ARCH_32BIT_ARMV7_LITTLE_ENDIAN, osarch.ARCH_32BIT_ARMV8_LITTLE_ENDIAN, osarch.ARCH_64BIT_ARMV8_LITTLE_ENDIAN:
```

lxd/instance/drivers/driver_qemu.go
```diff
@@ -2007,7 +2007,7 @@ func (d *qemu) qemuArchConfig(arch int) (path string, bus string, err error) {
 		}
 
 		return path, "pcie", nil
-	} else if arch == osarch.ARCH_64BIT_ARMV8_LITTLE_ENDIAN {
+	} else if arch == osarch.ARCH_32BIT_ARMV7_LITTLE_ENDIAN || arch == osarch.ARCH_32BIT_ARMV8_LITTLE_ENDIAN || arch == osarch.ARCH_64BIT_ARMV8_LITTLE_ENDIAN {
```
simondeziel, line 2010: Actually, could that be a case for `slices.Contains()`?

lxd/instance/drivers/driver_qemu_templates.go
simondeziel: As always, my questions/suggestions on anything related with Go are to be taken with a large grain of salt. @tomponline is really the authoritative voice here. I'm still just a Go noob ;)

lxd/instance/drivers/driver_qemu.go
kadinsayani: This could be a case for `slices.Contains()`. Personally, I prefer using logical OR for readability with a smaller number of options, but we could create a slice of supported ARM architectures and check using `slices.Contains()`.

It would look something like:

```Go
supportedArchs := []int{
    osarch.ARCH_32BIT_ARMV7_LITTLE_ENDIAN,
    osarch.ARCH_32BIT_ARMV8_LITTLE_ENDIAN,
    osarch.ARCH_64BIT_ARMV8_LITTLE_ENDIAN,
}
...
else if slices.Contains(supportedArchs, arch) {
...
}
```

The `supportedArchs` slice could be located in `driver_qemu.go` or `architectures.go` (with a "getter" function).

simondeziel: Elsewhere in the code I've seen the slice inlined as in:
```
else if slices.Contains([]int{osarch.ARCH_32BIT_ARMV7_LITTLE_ENDIAN, osarch.ARCH_32BIT_ARMV8_LITTLE_ENDIAN, osarch.ARCH_64BIT_ARMV8_LITTLE_ENDIAN}, arch) {
```

Really a question of style... inspired by my little Python background where `if arch in ("osarch.ARCH_32BIT_ARMV7_LITTLE_ENDIAN", ... ""):` would have worked ;)

kadinsayani: Thanks for sharing that one-liner! I'm happy with any of the above approaches. I'll wait for @tomponline to comment.

tomponline:
@kadinsayani if this is ready for review please can you click the "Ready for review" button, ta

lxd/instance/drivers/driver_qemu.go
tomponline: I'm fine with it as it is.

But if we were to use that logic in multiple places then I would say we should have a function that answers the specific question/classifier, e.g. `architectureSupportsUEFI` rather than returning a slice.

tomponline (approved):
Has this been tested working @kadinsayani ?

kadinsayani:
> Has this been tested working @kadinsayani ?

Launching `armhf` vm's on a 64-bit arm rpi host, yes ✅

Running `lxc exec c1 -- sh` with a snap that includes an armhf 32-bit `lxd-agent` binary, not yet ❌

I'll provide an update on this tomorrow 😄

tomponline:
> Running `lxc exec c1 -- sh` with a snap that includes an armhf 32-bit `lxd-agent` binary, not yet ❌

Yeah doing multi arch agent is a different feature.

So this pr ready to review?

simondeziel (approved):

tomponline (approved):

PR #14081, opened 2024-09-11
github: Only build coverage data on workflow_dispatch

tomponline:


simondeziel (dismissed):

markylaing (approved):
LGTM

So we'll only run coverage tests manually? I'm not sure how long it takes to run the tiobe stuff but it would be cool if we had a metric for whether each PR improves or degrades our score :)

tomponline:
> So we'll only run coverage tests manually? I'm not sure how long it takes to run the tiobe stuff but it would be cool if we had a metric for whether each PR improves or degrades our score :)

For now going to run it manually to get it working, then on a daily basis automatically.

But not sure about running it on every PR as it seems to take a long time.

markylaing:
> For now going to run it manually to get it working, then on a daily basis automatically.
> 
> But not sure about running it on every PR as it seems to take a long time.

Sounds good. Yeah probably unnecessary anyway as we're getting pretty good at adding test coverage for new PRs :)

PR #14083, opened 2024-09-11
Backup before cluster reconfigure

MggMuggins:
This follows a similar feature in microcluster. While any failure of the recovery process may be recoverable, expecting a user to surgically determine the exact step where the failure occurred, manually correct their `database/`, and create a recovery tarball is entirely unreasonable.

This PR creates a database backup `/var/snap/lxd/common/lxd/db_backup.TIMESTAMP.tar.gz` before any sensitive recovery operations so that a failed recovery doesn't cause data loss.

This also prevents `lxd cluster edit` from attempting to reconfigure the database more than once, as a failure during the recovery should allow the user to troubleshoot and retry from the backup.

LXD-1468

### DIFF
--- a/lxd/cluster/recover.go
+++ b/lxd/cluster/recover.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


lxd/cluster/recover.go
```diff
@@ -538,3 +539,87 @@ func unpackTarball(tarballPath string, destRoot string) error {
...
+	}
+
+	gzWriter := gzip.NewWriter(tarball)
+	tarWriter := tar.NewWriter(gzWriter)
+
+	filesys := os.DirFS(rootDir)
+
+	err = fs.WalkDir(filesys, walkDir, func(filepath string, stat fs.DirEntry, err error) error {
+		if err != nil {
+			logger.Warn("Failed to read file while creating tarball; skipping", logger.Ctx{"file": filepath, "err": err})
```
tomponline, line 566: why skipping here?

lxd/cluster/recover.go
```diff
@@ -538,3 +539,87 @@ func unpackTarball(tarballPath string, destRoot string) error {
...
+		}
+
+		info, err := stat.Info()
+		if err != nil {
+			return err
+		}
+
+		header, err := tar.FileInfoHeader(info, filepath)
+		if err != nil {
+			return fmt.Errorf("create tar header for %q: %w", filepath, err)
```
tomponline, line 581: uc first on errors please

lxd/cluster/recover.go
```diff
@@ -538,3 +539,87 @@ func unpackTarball(tarballPath string, destRoot string) error {
 
 	return nil
 }
+
+// createTarball creates tarball at tarballPath, rooted at rootDir and including
+// all files in walkDir except those paths found in excludeFiles.
+// walkDir and excludeFiles elements are relative to rootDir.
+func createTarball(tarballPath string, rootDir string, walkDir string, excludeFiles []string) error {
```
tomponline, line 553: is this a copy + edit of an existing function?

lxd/cluster/recover.go
```diff
@@ -538,3 +539,87 @@ func unpackTarball(tarballPath string, destRoot string) error {
...
+	tarball, err := os.Create(tarballPath)
```
tomponline, line 554: we should add a revert in case the create files part way through to clean up this file

lxd/cluster/recover.go
```diff
@@ -507,6 +507,31 @@ func unpackTarball(tarballPath string, destRoot string) error {
...
+	walkDir := path.Base(databaseDir)
+
+	logger.Info("Creating database backup", logger.Ctx{"archive": tarballPath})
+
+	// Don't include the recovery tarball in a backup tarball
+	excludeFiles := []string{path.Join(walkDir, RecoveryTarballName)}
+
+	err := createTarball(tarballPath, varDir, walkDir, excludeFiles)
+	if err != nil {
+		_ = os.Remove(tarballPath)
```
tomponline, line 543: this should be the responsibility of createTarball otherwise every caller has to remember to do it.

lxd/cluster/recover.go
```diff
@@ -507,6 +507,31 @@ func unpackTarball(tarballPath string, destRoot string) error {
...
+
+	// tar interprets `:` as a remote drive; ISO8601 allows a 'basic format'
+	// with the colons omitted (as opposed to time.RFC3339)
+	// https://en.wikipedia.org/wiki/ISO_8601
+	backupFileName := fmt.Sprintf("db_backup.%s.tar.gz", time.Now().Format("2006-01-02T150405Z0700"))
+	tarballPath := filepath.Join(varDir, backupFileName)
+
+	walkDir := path.Base(databaseDir)
+
+	logger.Info("Creating database backup", logger.Ctx{"archive": tarballPath})
```
tomponline, line 536: Can we just use "path" here to show where its going to be created. You could change the message to "Creating database backup archive" if the format is important to clarify.

lxd/cluster/recover.go
```diff
@@ -82,6 +82,11 @@ func localRaftNode(database *db.Node) (*db.RaftNode, error) {
 // member in the cluster. Use `Reconfigure` if more members should remain in
 // the raft configuration.
 func Recover(database *db.Node) error {
+	_, err := createDatabaseBackup(database.Dir())
+	if err != nil {
+		return fmt.Errorf("Create backup: %w", err)
```
tomponline, line 87: Please can we use "Failed creating backup: %w" style errors.

lxd/cluster/recover.go
MggMuggins: [Yes](https://github.com/canonical/microcluster/blob/v3/internal/recover/recover.go#L541)... I can drop this in `shared/` and import it over there if you'd prefer.

tomponline: no need to put it in shared, but if you can avoid duplicating the function and instead expand the scope of the existing function that would be great.

tomponline: oh wait, its from microcluster :)

Ah ok then its fine as is :)

lxd/cluster/recover.go
MggMuggins: I think this must have been left over from my debugging; removed it.

MggMuggins:
RTD build failed due to a [discourse.ubuntu.com]() outage; unsure how to rerun.

tomponline (approved):
Ta!

PR #14088, opened 2024-09-12
lxc/config: create `oidctokens` dir with 0750

simondeziel:
The actual tokens are already safe with 0600 but let's also avoid disclosing which tokens we have by making the directory a bit more private. Same as the `servercerts` directory.


markylaing (approved):
Makes sense, thanks!

tomponline (approved):

PR #14091, opened 2024-09-12
Remove type bound public access

markylaing:
Removes the [type-bound public access](https://openfga.dev/docs/modeling/public-access) for server and storage pool and replace with an authentication check. This is to simplify the model and the underlying OpenFGADatastore implementation.

This does not change any users existing permissions, as it is not possible to create a permission with `can_view` on `server` or `storage_pool`.

Additionally, this PR improves the TLS authorization driver (see commit message) and adds tests.

### DIFF
--- a/lxd/auth/drivers/openfga.go
+++ b/lxd/auth/drivers/openfga.go
@@ -129,6 +129,16 @@ func (e *embeddedOpenFGA) load(ctx context.Context, identityCache *identity.Cach
 // check, but will not automatically allow "punching through" to the effective (default) project. An administrator can
 // allow specific permissions against those entities.
 func (e *embeddedOpenFGA) CheckPermission(ctx context.Context, entityURL *api.URL, entitlement auth.Entitlement) error {
+	entityType, projectName, location, pathArguments, err := entity.ParseURL(entityURL.URL)
+	if err != nil {
+		return fmt.Errorf("Failed to parse entity URL: %w", err)
+	}
+
+	err = auth.ValidateEntitlement(entityType, entitlement)
+	if err != nil {
+		return fmt.Errorf("Cannot check permissions for entity type %q and entitlement %q: %w", entityType, entitlement, err)
+	}
+
 	logCtx := logger.Ctx{"entity_url": entityURL.String(), "entitlement": entitlement}
 	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
@@ -182,12 +192,6 @@ func (e *embeddedOpenFGA) CheckPermission(ctx context.Context, entityURL *api.UR
 				groups = append(groups, lxdGroup)
 			}
 		}
-	}
-
-	// Deconstruct the given URL.
-	entityType, projectName, location, pathArguments, err := entity.ParseURL(entityURL.URL)
-	if err != nil {
-		return fmt.Errorf("Authorization driver failed to parse entity URL %q: %w", entityURL.String(), err)
 	}
 
 	// The project in the given URL may be for a project that does not have a feature enabled, in this case the auth check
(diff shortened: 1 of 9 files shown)


lxd/auth/drivers/tls.go
```diff
@@ -161,7 +140,51 @@ func (t *tls) GetPermissionChecker(ctx context.Context, entitlement auth.Entitle
...
+		if !projectSpecific {
+			return t.allowProjectUnspecificEntityType(entitlement, entityType, id, project, pathArguments)
+		}
+
 		// Otherwise, check if the project is in the list of allowed projects for the entity.
 		return shared.ValueInSlice(project, id.Projects)
 	}, nil
 }
+
+func (t *tls) allowProjectUnspecificEntityType(entitlement auth.Entitlement, entityType entity.Type, id *identity.CacheEntry, projectName string, pathArguments []string) bool {
```
tomponline, line 163: please can we have a docblock comment on this one?

lxd/auth/drivers/tls.go
```diff
@@ -154,10 +154,7 @@ func (t *tls) allowProjectUnspecificEntityType(entitlement auth.Entitlement, ent
 	switch entityType {
 	case entity.TypeServer:
 		// Restricted TLS certificates have the following entitlements on server.
-		return shared.ValueInSlice(entitlement, []auth.Entitlement{auth.EntitlementCanView, auth.EntitlementCanViewResources, auth.EntitlementCanViewMetrics})
-	case entity.TypeStoragePool:
```
tomponline: Should this return an error then?

> We should never expect the authorizer to be called with this entitlement on these entities.

lxd/auth/drivers/openfga.go
```diff
@@ -129,6 +129,16 @@ func (e *embeddedOpenFGA) load(ctx context.Context, identityCache *identity.Cach
 // check, but will not automatically allow "punching through" to the effective (default) project. An administrator can
 // allow specific permissions against those entities.
 func (e *embeddedOpenFGA) CheckPermission(ctx context.Context, entityURL *api.URL, entitlement auth.Entitlement) error {
+	entityType, projectName, location, pathArguments, err := entity.ParseURL(entityURL.URL)
```
tomponline, line 132: good idea

lxd/auth/drivers/openfga.go
```diff
@@ -256,11 +256,14 @@ func (e *embeddedOpenFGA) CheckPermission(ctx context.Context, entityURL *api.UR
 
 	// If not allowed, decide if the user can view the resource.
```
tomponline, line 257: Please can you tweak this comment as it reads a bit odd now:

> // If not allowed, decide if the user can view the resource.

This statement seems contradictory.

How about:

> // If not allowed, then decide if the user can view the resource in order to alter the HTTP response code.

lxd/auth/drivers/openfga.go
```diff
@@ -256,11 +256,14 @@ func (e *embeddedOpenFGA) CheckPermission(ctx context.Context, entityURL *api.UR
...
 	if !resp.GetAllowed() {
+		err := auth.ValidateEntitlement(entityType, auth.EntitlementCanView)
```
tomponline, line 259: Should we be checking for a specific type/class of error here? What if ValidateEntitlement later returns an error which doesn't mean "no error, but user cant view", but means "something went wrong internally"?

tomponline (approved):

lxd/auth/drivers/tls.go
markylaing: We could but the entitlements are now being validated when CheckPermission or GetPermissionChecker is called. So an error will have been returned at that point.

PR #14078, opened 2024-09-09
Per-pool project limits (from Incus)

MggMuggins:
Implements `limits.disk.pool.POOLNAME` config key on projects. When set, the key prevents a project using more than the configured amount of space within that pool.

When `limits.disk.pool.POOLNAME: 0`, the pool is excluded from `lxc storage list` in that project.

Contains cherry-picks from https://github.com/lxc/incus/pull/1098

PR #14095, opened 2024-09-13
Project: Don't fail creation on missing pools (from Incus)

roosterfish:
From https://github.com/lxc/incus/pull/1109.

After https://github.com/canonical/lxd/pull/14078 got merged the following error started to appear on MicroCloud when forming clusters: `
Error: System "micro02" failed to join the cluster: Failed to update cluster status of services: Failed to join "LXD" cluster: Failed to configure cluster: Failed to initialize member: Failed to initialize storage pools and networks: Failed to update local member project "default": Failed loading storage pool names: Storage pool(s) not found`

The DB`s `GetStoragePoolNames()` returns 404 in case of no storage pools. This has to be checked explicitly otherwise the error is returned back to the caller.

### DIFF
--- a/lxd/api_project.go
+++ b/lxd/api_project.go
@@ -1407,7 +1407,7 @@ func projectValidateConfig(s *state.State, config map[string]string) error {
 
 		return nil
 	})
-	if err != nil {
+	if err != nil && !api.StatusErrorCheck(err, http.StatusNotFound) {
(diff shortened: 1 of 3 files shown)


tomponline:
Thanks! good catch.

tomponline:
@roosterfish is it viable to add a test for this scenario?

tomponline (approved):
nice one

PR #14110, opened 2024-09-16
Refactor `fsmonitor` package in preparation for Ubuntu Pro guest attachment

markylaing:
The `fsmonitor` package is designed for watching `/dev` for available devices that can be passed into guest machines. The monitor is not configurable and only one can be instantiated.

This PR updates `fsmonitor` so that it can be re-used in multiple scenarios, including:
- Specifying which events to listen for.
- Removing the restriction that the given path must be a mount point (this is not required for inotify, but is kept for fanotify).
- Allowing more than one fsmonitor to be instantiated.

Some other improvements include:
- Using the `fsmonitor.Event` type to denote the event type outside of the package (previously this was `string`). 
- Improved logging for unrecognised events, and a potential fix for an unrecognised event related to `inotify.OnDeleteSelf`.

This is a prerequisite of #13953


lxd/device/device_utils_unix_events.go
```diff
@@ -140,18 +140,7 @@ func unixRunHandlers(state *state.State, event *UnixEvent) {
 }
 
 // unixNewEvent returns a newly created Unix device event struct.
-// If an empty action is supplied then the action of the event is derived from whether the path
-// exists (add) or not (removed). This allows the peculiarities of the inotify API to be somewhat
-// masked by the consuming event handler functions.
-func unixNewEvent(action string, path string) UnixEvent {
```
tomponline, line 146: did you test this with inotify and not just fanotify?

tomponline (approved):
Thanks!

lxd/device/device_utils_unix_events.go
markylaing: That's a good point. I'm not sure which driver is used in the test, but I'm guessing it's fanotify. I'll see if I can find a way to force one or the other and ensure we test both.

tomponline: try using an older ubuntu release, maybe ubuntu:focal

PR #14111, opened 2024-09-16
fsmonitor: Test unix devices with both drivers

markylaing:
The `fsmonitor` package has two drivers, but only fanotify is tested for unix devices (see https://github.com/canonical/lxd/pull/14110#discussion_r1761232451). This is the default driver but it may not be available in all kernels/architectures.

This PR adds an environment variable to force a specific driver to be used. The unix char and block device tests are then run with both drivers.

The `inotify` package was encountering some unknown events. As it turned out, mapping these events to `fsmonitor` events broke the test suite, so I've added exceptions for them to reduce log verbosity.


tomponline (approved):
Nice, thanks!

simondeziel (approved):

Issue #14098, opened 2024-09-13
Shell completion improvement suggestions


simondeziel:
`lxc config get <tab>` should suggest:
* local instances
* name of remotes minus builtin and simplestreams ones

`lxc config get core.<tab>` (`.` is the distinguishing factor) should suggest:
* `local:` server's options starting with `core.`

`lxc config get foo: core.<tab>` (`.` is the distinguishing factor) should suggest:
* `foo:` server's options starting with `core.`

`lxc config get foo:<tab>` (`:` is the distinguishing factor) should suggest:
* instances on `foo:`

`lxc config get foo:bar <tab>` should suggest:
* instance config options for `foo:bar`

tomponline:
@simondeziel thanks!

What about also:

```
lxc config get foo:core.<tab>
```

Remote server's options starting with core.

simondeziel:
> What about also:

Good point, added to the issue description! Thanks

kadinsayani:
Currently, `lxc config get <instance>` will return all config options (even those which aren't set). 

These are the two cases in question:

> `lxc config get core.<tab>` (`.` is the distinguishing factor) should suggest:
> 
> * `local:` server's options starting with `core.`
> 
> `lxc config get foo:core.<tab>` (`.` is the distinguishing factor) should suggest:
> 
> * `foo:` server's options starting with `core.`

With the current behaviour, regardless of the remote, all server options would be suggested. Do we want to change this behaviour to only suggest options which are set? Personally, I find it convenient to use commands like `lxc config get <some_config_option>` to see if something is already set, or set by default. Happy to take suggestions on this :)

kadinsayani:
Also, `completion.go` is currently undocumented. Would you like me to use this time to also add go doc comments?

simondeziel:
> Currently, `lxc config get <instance>` will return all config options (even those which aren't set).

That's not what I see with `latest/edge`:
```
$ snap list lxd
Name  Version      Rev    Tracking     Publisher   Notes
lxd   git-44a52ea  30256  latest/edge  canonical✓  -
```

That said, the previous completion script was completing **server config options only** (not instance) when doing `lxc config get <tab>`.

> With the current behaviour, regardless of the remote, all server options would be suggested. Do we want to change this behaviour to only suggest options which are set? Personally, I find it convenient to use commands like `lxc config get <some_config_option>` to see if something is already set, or set by default. Happy to take suggestions on this :)

[...]

PR #14115, opened 2024-09-16
lxc: Add completions for server keys and console

kadinsayani:
Resolves https://github.com/canonical/lxd/issues/14098.

All the improvements made to the `lxc config get/set` completions have been tested locally. I made a minor change that deviates from the requirements outlined in https://github.com/canonical/lxd/issues/14098. Instead of typing `lxc config get foo:core.<tab>` to retrieve server configuration options, typing `lxc config get foo: core.<tab>` (notice the [SPACE]) will now return server configuration options. This change was implemented to differentiate from the conventional usage of `remote:instance` across the CLI. However, please let me know if you'd prefer server configuration completions without the [SPACE] 😄.

### DIFF
--- a/client/interfaces.go
+++ b/client/interfaces.go
@@ -73,6 +73,7 @@ type InstanceServer interface {
 	ImageServer
 
 	// Server functions
+	GetMetadataConfiguration() (metadataConfiguration *api.MetadataConfiguration, err error)
 	GetMetrics() (metrics string, err error)
 	GetServer() (server *api.Server, ETag string, err error)
 	GetServerResources() (resources *api.Resources, err error)
--- a/client/lxd_server.go
+++ b/client/lxd_server.go
@@ -186,3 +186,21 @@ func (r *ProtocolLXD) GetMetrics() (string, error) {
 
 	return string(content), nil
 }
+
+// GetMetadataConfiguration returns metadata configuration for a server.
+func (r *ProtocolLXD) GetMetadataConfiguration() (*api.MetadataConfiguration, error) {
+	// Check that the server supports it.
+	err := r.CheckExtension("metadata_configuration")
+	if err != nil {
+		return nil, err
+	}
+
+	metadataConfiguration := api.MetadataConfiguration{}
+
+	_, err = r.queryStruct(http.MethodGet, api.NewURL().Path("metadata", "configuration").String(), nil, "", &metadataConfiguration)
+	if err != nil {
+		return nil, err
+	}
+
+	return &metadataConfiguration, err
+}
--- a/lxc/completion.go
+++ b/lxc/completion.go
@@ -211,6 +211,18 @@ func (g *cmdGlobal) cmpImages(toComplete string) ([]string, cobra.ShellCompDirec
 }
 
 func (g *cmdGlobal) cmpInstanceAllKeys(instanceName string) ([]string, cobra.ShellCompDirective) {
+	var keys []string
+	cmpDirectives := cobra.ShellCompDirectiveNoFileComp
+
+	_, instanceNameOnly, _ := strings.Cut(instanceName, ":")
+	if instanceNameOnly == "" {
+		serverKeys, directives := g.cmpServerAllKeys(instanceName)
+		keys = append(keys, serverKeys...)
+		cmpDirectives = directives
+
+		return keys, cmpDirectives
+	}
+
 	resources, err := g.ParseServers(instanceName)
 	if err != nil || len(resources) == 0 {
 		return nil, cobra.ShellCompDirectiveError
@@ -219,13 +231,12 @@ func (g *cmdGlobal) cmpInstanceAllKeys(instanceName string) ([]string, cobra.She
 	resource := resources[0]
 	client := resource.server
 
-	instanceNameOnly, _, err := client.GetInstance(instanceName)
+	instance, _, err := client.GetInstance(instanceNameOnly)
 	if err != nil {
 		return nil, cobra.ShellCompDirectiveError
 	}
 
-	var keys []string
-	instanceType := instanceNameOnly.Type
+	instanceType := instance.Type
 
 	if instanceType == "container" {
 		for k := range instancetype.InstanceConfigKeysContainer {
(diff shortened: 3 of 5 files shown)


simondeziel:
> I made a minor change that deviates from the requirements outlined in #14098. Instead of typing `lxc config get foo:core.<tab>` to retrieve server configuration options, typing `lxc config get foo: core.<tab>` (notice the [SPACE]) will now return server configuration options. This change was implemented to differentiate from the conventional usage of `remote:instance` across the CLI. However, please let me know if you'd prefer server configuration completions without the [SPACE] 😄.

That was the right thing to do as I got the example in the bug wrong, thanks for catching/fixing this!

lxc/completion.go
```diff
@@ -211,6 +212,21 @@ func (g *cmdGlobal) cmpImages(toComplete string) ([]string, cobra.ShellCompDirec
...
+		instanceName = strings.Split(instanceName, ":")[1]
+
+		if instanceName == "" {
+			serverKeys, directives := g.cmpServerAllKeys()
+			keys = append(keys, serverKeys...)
+			cmpDirectives = directives
+
+			return keys, cmpDirectives
+		}
+	}
```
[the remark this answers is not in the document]
kadinsayani: Thanks for the suggestion! I can see the potential for some optimization here, but we only want to modify `instanceName` if we're checking for server config keys. `instanceName` is later passed into `ParseServers()` and we want to make sure we're grabbing the right keys for an instance belonging to a remote. There may be an edge case where two distinct remotes have an instance with the same name, with one instance being a VM and one being a container (different options for config keys). 

We could use `strings.Cut()` but store the result in a variable named `remoteOnly`, or leave the code as is.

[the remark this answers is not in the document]
kadinsayani: I'm okay with either approach 😄

simondeziel (approved):

lxc/completion.go
```diff
@@ -211,6 +212,21 @@ func (g *cmdGlobal) cmpImages(toComplete string) ([]string, cobra.ShellCompDirec
 }
 
 func (g *cmdGlobal) cmpInstanceAllKeys(instanceName string) ([]string, cobra.ShellCompDirective) {
+	var keys []string
+	cmpDirectives := cobra.ShellCompDirectiveNoFileComp
+
+	if strings.Contains(instanceName, ":") {
+		instanceName = strings.Split(instanceName, ":")[1]
```
markylaing, line 219: You can likely use `strings.Cut` here to get the instance name and check if the colon is present in one step.

lxc/completion.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/canonical/lxd/lxd/instance/instancetype"
+	"github.com/canonical/lxd/lxd/node"
```
markylaing, line 11: You won't be able to import this package into the client unfortunately. That's why the build is failing.

kadinsayani:
I’ve updated the `cmpServerAllKeys()` function to use the metadata API by calling `GetMetadataConfiguration()`. I believe it would be more efficient to create a `MetadataConfiguration` struct in the `/shared/api` directory. This way, we can avoid parsing the response in the `completions.go` file in polynomial time. Instead, we can unmarshal the necessary information to a nested struct. I'm not sure if it would improve performance, but it would definitely be more elegant. Thoughts?

Also, @simondeziel, you were right all along about `strings.cut()` 😄.

client/lxd_server.go
```diff
@@ -186,3 +187,50 @@ func (r *ProtocolLXD) GetMetrics() (string, error) {
 
 	return string(content), nil
 }
+
+// GetMetadataConfiguration returns metadata configuration for a server.
+func (r *ProtocolLXD) GetMetadataConfiguration() (*map[string]any, error) {
+    // Check that the server supports it.
+	err := r.CheckExtension("metadata_configuration")
```
simondeziel, line 194: Tabs and spaces are mixed here.

client/interfaces.go
```diff
@@ -73,6 +73,7 @@ type InstanceServer interface {
 	ImageServer
 
 	// Server functions
+    GetMetadataConfiguration() (config *map[string]any, err error)
```
simondeziel, line 76: Tabs should be used instead of spaces. Thx

tomponline: Or more precisely: go fmt should be used for all code when saving in your ide

client/lxd_server.go
```diff
@@ -186,3 +187,50 @@ func (r *ProtocolLXD) GetMetrics() (string, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Unmarshal content.
+	config := make(map[string]any)
+	err = json.Unmarshal(content, &config)
+	if err != nil {
+		return nil, err
+	}
```
markylaing, line 233: I think you can use `r.queryStruct` here:
```go
config := make(map[string]any)
_, err = r.queryStruct(http.MethodGet, api.NewURL("metadata", "configuration").String(), nil, "", &config)
if err != nil {
    return err
}

return config, nil
```

Also note that this function will extract only the metadata from the request body. Your method here is returning the full API response as a `*map[string]any` which we don't need. The `queryStruct` function will return an error if the server returns an error.

Lastly, there's no need to return a `*map[string]any` as a map is already a pointer, can you please just return `map[string]any`, thanks.
```

**Review by markylaing — CHANGES_REQUESTED** [review]
I think it's fine to create a type in the `shared/api` package for the configuration metadata. It is exposed over the API and would break the UI if changed, so we are beholden to it.

**Comment by kadinsayani** [comment]
@markylaing I rebased my local branch on your PR branch https://github.com/canonical/lxd/pull/14132.

**Comment by tomponline** [comment]
@kadinsayani should we also import https://github.com/lxc/incus/pull/1231 ?

**Comment by kadinsayani** [comment]
> @markylaing I rebased my local branch on your PR branch #14132. Is this okay, or should I wait for your PR to be merged and then rebase on main?

I reverted this to avoid conflicts.

**Comment by tomponline** [comment]
tests failing

**Comment by kadinsayani** [comment]
> tests failing

Awaiting merge of https://github.com/canonical/lxd/pull/14132 to `latest/edge`.

**Comment by kadinsayani** [comment]
Rebased and good to go 👍

**Review by tomponline — APPROVED** [review]

**Review by simondeziel — DISMISSED** [review]

File: lxc/completion.go [hunk 7]
```diff
@@ -241,6 +252,38 @@ func (g *cmdGlobal) cmpInstanceAllKeys(instanceName string) ([]string, cobra.She
...
+		return nil, cobra.ShellCompDirectiveError
+	}
+
+	for _, field := range server {
+		for _, keyMap := range field.Keys {
+			for key := range keyMap {
+				keys = append(keys, key)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
kadinsayani: We can preallocate the slice:

```Go
keyCount := 0
	for _, field := range server {
		keyCount += len(field.Keys)
	}
	keys = make([]string, 0, keyCount)

	for _, field := range server {
		for _, keyMap := range field.Keys {
			for key := range keyMap {
				keys = append(keys, key)
			}
		}
	}
```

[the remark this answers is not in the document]
simondeziel: Thanks, that seems to be a pattern used by Go libs, like in https://cs.opensource.google/go/go/+/refs/tags/go1.23.1:src/strings/strings.go;l=243-250

lxc/completion.go
```diff
@@ -241,6 +252,45 @@ func (g *cmdGlobal) cmpInstanceAllKeys(instanceName string) ([]string, cobra.She
...
+	if !ok {
+		return nil, cobra.ShellCompDirectiveError
+	}
+
+	keyCount := 0
+	for _, field := range server {
+		keyCount += len(field.Keys)
+	}
+
+	keys = make([]string, 0, keyCount)
```
simondeziel, line 284:
```suggestion
	keys := make([]string, 0, keyCount)
```

Maybe and remove the earlier declaration?

kadinsayani: I also prefer short variable declarations in functions, thanks for pointing this out!

simondeziel (approved):

tomponline (approved):

Issue #13468, opened 2024-05-10
lxd init produces config which fails validation


mmokrejs:
# Required information

 * Distribution: Gentoo
 * Distribution version: current
 * The output of "snap list --all lxd core20 core22 core24 snapd": NA
 * The output of "lxc info" or if that fails:
```
# lxc info
config: {}
api_extensions:
- storage_zfs_remove_snapshots
- container_host_shutdown_timeout
- container_stop_priority
- container_syscall_filtering
- auth_pki
- container_last_used_at
- etag
- patch
- usb_devices
- https_allowed_credentials
- image_compression_algorithm
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
- unix_device_rename
- storage_lvm_use_thinpool
- storage_rsync_bwlimit
- network_vxlan_interface
- storage_btrfs_mount_options
- entity_description
- image_force_refresh
- storage_lvm_lv_resizing
- id_map_base
- file_symlinks
- container_push_target
- network_vlan_physical
- storage_images_delete
- container_edit_metadata
- container_snapshot_stateful_migration
- storage_driver_ceph
- storage_ceph_user_name
- resource_limits
- storage_volatile_initial_source
- storage_ceph_force_osd_reuse
- storage_block_filesystem_btrfs
- resources
- kernel_limits
- storage_api_volume_rename
- network_sriov
- console
- restrict_devlxd
- migration_pre_copy
- infiniband
- maas_network
- devlxd_events
- proxy
- network_dhcp_gateway
- file_get_symlink
- network_leases
- unix_device_hotplug
- storage_api_local_volume_handling
- operation_description
- clustering
- event_lifecycle
- storage_api_remote_volume_handling
- nvidia_runtime
- container_mount_propagation
- container_backup
- devlxd_images
- container_local_cross_pool_handling
- proxy_unix
- proxy_udp
- clustering_join
- proxy_tcp_udp_multi_port_handling
- network_state
[...]

PR #14119, opened 2024-09-17
lxd/init: Differentiate managed and unmanaged network

MusicDin:
When using `lxd init` to configure default profile with an existing network, differentiate between managed and unmanaged networks by first trying to retrieve an existing network (and checking if it is managed).

Unamanaged network:
```sh
...
Would you like to configure LXD to use an existing bridge or host interface? (yes/no) [default=no]: yes
Name of the existing bridge or host interface: tmpbr0
...
Would you like a YAML "lxd init" preseed to be printed? (yes/no) [default=no]: y
config: {}
networks: []
storage_pools: []
storage_volumes: []
profiles:
- config: {}
  description: ""
  devices:
    eth0:
      name: eth0
      nictype: bridged # <--
      parent: tmpbr0   # <--
      type: nic
  name: default
projects: []
cluster: null
```

Managed network:
```sh
...
Would you like to configure LXD to use an existing bridge or host interface? (yes/no) [default=no]: yes
Name of the existing bridge or host interface: lxdbr0
...
Would you like a YAML "lxd init" preseed to be printed? (yes/no) [default=no]: y
config: {}
networks: []
storage_pools: []
storage_volumes: []
profiles:
- config: {}
  description: ""
  devices:
    eth0:
      name: eth0
      network: lxdbr0 # <--
      type: nic
  name: default
projects: []
cluster: null
```

Closes https://github.com/canonical/lxd/issues/13468

### DIFF
--- a/lxd/main_init_interactive.go
+++ b/lxd/main_init_interactive.go
@@ -349,27 +349,51 @@ func (c *cmdInit) askNetworking(config *api.InitPreseed, d lxd.InstanceServer) e
 		}
 
 		if useExistingInterface {
+			networks, err := d.GetNetworks()
+			if err != nil {
+				return err
+			}
+
 			for {
 				interfaceName, err := c.global.asker.AskString("Name of the existing bridge or host interface: ", "", nil)
 				if err != nil {
 					return err
 				}
 
-				if !shared.PathExists(fmt.Sprintf("/sys/class/net/%s", interfaceName)) {
+				// Try to find the existing network.
+				var network *api.Network
+				for _, n := range networks {
+					if n.Name == interfaceName {
+						network = &n
+						break
+					}
+				}
+
+				if network == nil {
 					fmt.Println("The requested interface doesn't exist. Please choose another one.")
 					continue
 				}
 
-				// Add to the default profile
-				config.Node.Profiles[0].Devices["eth0"] = map[string]string{
-					"type":    "nic",
-					"nictype": "macvlan",
-					"name":    "eth0",
-					"parent":  interfaceName,
-				}
+				// Add to the default profile.
+				if network.Managed {
+					// Add managed network.
+					config.Node.Profiles[0].Devices["eth0"] = map[string]string{
+						"name":    "eth0",
+						"type":    "nic",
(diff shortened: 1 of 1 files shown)


lxd/main_init_interactive.go
```diff
@@ -360,16 +360,27 @@ func (c *cmdInit) askNetworking(config *api.InitPreseed, d lxd.InstanceServer) e
...
 
-				// Add to the default profile
-				config.Node.Profiles[0].Devices["eth0"] = map[string]string{
-					"type":    "nic",
-					"nictype": "macvlan",
-					"name":    "eth0",
-					"parent":  interfaceName,
-				}
+				// Add to the default profile.
+				network, _, err := d.GetNetwork(interfaceName)
```
tomponline, line 364: we should probably differentiate a not found error from other errors here and only continue on a not found error.

We shouldnt mask other errors and treat them as not found.

lxd/main_init_interactive.go
```diff
@@ -360,16 +360,27 @@ func (c *cmdInit) askNetworking(config *api.InitPreseed, d lxd.InstanceServer) e
...
+					config.Node.Profiles[0].Devices["eth0"] = map[string]string{
+						"name":    "eth0",
+						"type":    "nic",
+						"nictype": "macvlan",
+						"parent":  interfaceName,
+					}
 
-				if shared.PathExists(fmt.Sprintf("/sys/class/net/%s/bridge", interfaceName)) {
-					config.Node.Profiles[0].Devices["eth0"]["nictype"] = "bridged"
+					if shared.PathExists(fmt.Sprintf("/sys/class/net/%s/bridge", interfaceName)) {
```
tomponline, line 381: @MusicDin could we rework this section so it just pulls a list of networks from LXD client and then looks at their network type to decide what to do rather than probing /sys/class here (its what LXD is doing anyway when pulling a list of networks).

MusicDin: Yes. Thanks for suggestion

tomponline (approved):
Cool thanks

Issue #12130, opened 2023-08-09
lxd-migrate should not just tell the user that instance name is incorrect and quit


m0nkey-br4in:
lxd-migrate version 5.15

Name of new instance is the second question it asks you in the process of migration, then it makes you go through the motions of configuring everything, then finally you press `1` to begin the transfer process and get this:
`"Error: Invalid instance name: Name can only contain alphanumeric and hyphen characters"`
And then it exits.

I had some choice words to say about the designer of this UI interaction. Since I've cooled down a bit, all I can say is that the program shouldn't behave like this. It should tell you about incorrect name right after you enter it, and it shouldn't just quit, it should let you change it.

tomponline:
Yeah I think we should certainly make improvements to the user experience in lxd-migrate, some ideas come to mind:

-     Better client side validation.
-     Command flags to allow it to run in non-interactive mode.
-     Allow talking to local LXD server over unix socket.

PR #14122, opened 2024-09-18
lxd-migrate: Prevent invalid instance names

MusicDin:
Prevent `lxd-migrate` from progressing if instance name is not valid.

Noninteractive mode:
```sh
$ sudo ./lxd-migrate --name "test test"
Error: Invalid instance name "test test": Name can only contain alphanumeric and hyphen characters

$ sudo ./lxd-migrate --name "test"
The local LXD server is the target [default=yes]: ...
```

Interactive mode:
```sh
$ sudo ./lxd-migrate
The local LXD server is the target [default=yes]: 
Would you like to create a container (1) or virtual-machine (2)?: 1
Name of the new instance: test test
Invalid instance name "test test": Name can only contain alphanumeric and hyphen characters
Name of the new instance: test
Please provide the path to a root filesystem: ...
```

Closes https://github.com/canonical/lxd/issues/12130

### DIFF
--- a/lxd-migrate/main_migrate.go
+++ b/lxd-migrate/main_migrate.go
@@ -20,6 +20,7 @@ import (
 	"gopkg.in/yaml.v2"
 
 	"github.com/canonical/lxd/client"
+	"github.com/canonical/lxd/lxd/instance/instancetype"
 	"github.com/canonical/lxd/lxd/storage/block"
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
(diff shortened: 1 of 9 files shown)


MusicDin:
Hmm, can't include `lxd/instance` in `lxd-migrate` because it has reference to `lxd/db` package. Should we move the validator into shared?

tomponline:
> Hmm, can't include `lxd/instance` in `lxd-migrate` because it has reference to `lxd/db` package. Should we move the validator into shared?

which function do you need?

MusicDin:
`instance.ValidName()`

I've just extracted the part that checks non-snapshot instance name and added it into the `lxd-migrate/utils`.

tomponline:
We could move to instancetype package

tomponline (approved):

PR #14124, opened 2024-09-18
WIP: Refactor operations to not import state.

markylaing:
Refactors the `operations` package so that:
- `OperationCreate` accepts a `context.Context` (expected to be the daemon shutdown context), the operation location, an events server, and a `*operations.Opts`.
- `operations.Options` and `operations.ClusterOptions` can be used to specify optional operation create arguments.
- `state.State` is no longer imported by the operations package. Instead, a transaction function is required for LXD daemon operations. This can be used by the operation to add/delete itself. The transaction function has a hook with `*sql.Tx` rather than `db.ClusterTx` so that we don't import any database files.


markylaing:
@simondeziel @tomponline considering this PR has greatly reduced the number of imports for the lxd-agent, I'm baffled that the lxd-agent size has actually gone **up** (https://github.com/canonical/lxd/actions/runs/10921682830/job/30314339170?pr=14124#step:16:73)

any ideas?

tomponline:
> @simondeziel @tomponline considering this PR has greatly reduced the number of imports for the lxd-agent, I'm baffled that the lxd-agent size has actually gone **up** (https://github.com/canonical/lxd/actions/runs/10921682830/job/30314339170?pr=14124#step:16:73)
> 
> any ideas?

do you observe the size change when compiling locally?

tomponline:
btw im loving your work on removing deps from lxd-agent

simondeziel:
> @simondeziel @tomponline considering this PR has greatly reduced the number of imports for the lxd-agent, I'm baffled that the lxd-agent size has actually gone **up** (https://github.com/canonical/lxd/actions/runs/10921682830/job/30314339170?pr=14124#step:16:73)

Looking at the last run in `main`, it was very close to hitting the max limit already, 6088 bytes headroom:

```
+ '[' 12576824 -gt 12582912 ']'
OK: lxd-agent is between 11 and 12MiB
+ '[' 12576824 -lt 11534336 ']'
+ echo 'OK: lxd-agent is between 11 and 12MiB'
```

In your PR it grew by189760 bytes:

```
+ '[' 12766584 -gt 12582912 ']'
FAIL: lxd-agent binary size exceeds 12MiB
```

I cannot explain that increase as in both cases the `go.mod` and the Go version are identical. It's weird and I'd like to find the explanation but I don't think that should stop you from doing that refactor :)

test/godeps/lxd-agent.list
```diff
@@ -13,7 +13,6 @@ github.com/canonical/lxd/lxd/metrics
 github.com/canonical/lxd/lxd/operations
 github.com/canonical/lxd/lxd/request
 github.com/canonical/lxd/lxd/response
-github.com/canonical/lxd/lxd/state
```
markylaing, line 49: So when using the build tags to list dependencies, this refactor only removes a single package! I guess that explains why the binary size didn't get smaller - though it doesn't explain why it got larger.

markylaing:
Closing as this was not required and did not reduce the size of the LXD agent binary.

PR #14127, opened 2024-09-18
Use agent build tags when performing dependency check

markylaing:
Uses the correct build tags when checking the dependency list for the LXD agent.


test/lint/godeps.sh
```diff
@@ -9,8 +9,15 @@ rc=0
...
   echo "==> Checking for imports/deps that have been added to ${pkg}..."
+
+  OUT=""
   DEP_FILE="test/godeps/$(echo "${pkg}" | sed 's/\//-/g').list"
-  OUT="$(go list -f '{{ join .Deps "\n" }}' ./${pkg} | grep -F . | sort -u | diff --new-file -u "${DEP_FILE}" - || true)"
+  if [ "${pkg}" = "lxd-agent" ]; then
+    OUT="$(go list -tags agent,netgo -f '{{ join .Deps "\n" }}' ./${pkg} | grep -F . | sort -u | diff --new-file -u "${DEP_FILE}" - || true)"
+  else
+    OUT="$(go list -f '{{ join .Deps "\n" }}' ./${pkg} | grep -F . | sort -u | diff --new-file -u "${DEP_FILE}" - || true)"
+  fi
```
simondeziel, line 19:
```suggestion
  TAGS=""
  [ "${pkg}" = "lxd-agent" ] && TAGS="-tags agent,netgo"
  OUT="$(go list ${TAGS:+"${TAGS}"} -f '{{ join .Deps "\n" }}' ./${pkg} | grep -F . | sort -u | diff --new-file -u "${DEP_FILE}" - || true)"
```
It's not exactly pretty but avoid dup'ing the command pipeline

markylaing: This didn't quite work initially, but works with `${TAGS:+${TAGS}}`, e.g. removing the quotes.

simondeziel: Right, the quotes cause the `-tags` and `agent,netgo` to be seen as a single argument with a space in the middle, exactly what I wanted to prevent. This was done trying to shut up `shellcheck` :/

simondeziel (approved):
Nice job!

PR #14130, opened 2024-09-18
Storage: Return `growFileSystem` errors

MggMuggins:
`msg` only contains stdout from the child process, so this wasn't passing back any information about what actually failed.

Still working on root cause for these failures: https://github.com/canonical/lxd/actions/runs/10919085792/attempts/1?pr=14122

### DIFF
--- a/lxd/storage/drivers/utils.go
+++ b/lxd/storage/drivers/utils.go
@@ -537,21 +537,20 @@ func growFileSystem(fsType string, devPath string, vol Volume) error {
 	}
 
 	return vol.MountTask(func(mountPath string, op *operations.Operation) error {
-		var msg string
 		var err error
 		switch fsType {
 		case "ext4":
-			msg, err = shared.TryRunCommand("resize2fs", devPath)
+			_, err = shared.TryRunCommand("resize2fs", devPath)
 		case "xfs":
-			msg, err = shared.TryRunCommand("xfs_growfs", mountPath)
+			_, err = shared.TryRunCommand("xfs_growfs", mountPath)
 		case "btrfs":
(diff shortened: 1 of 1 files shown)


lxd/storage/drivers/utils.go
```diff
@@ -537,21 +537,21 @@ func growFileSystem(fsType string, devPath string, vol Volume) error {
...
 		case "btrfs":
-			msg, err = shared.TryRunCommand("btrfs", "filesystem", "resize", "max", mountPath)
+			_, err = shared.TryRunCommand("btrfs", "filesystem", "resize", "max", mountPath)
 		default:
 			return fmt.Errorf("Unrecognised filesystem type %q", fsType)
 		}
 
 		if err != nil {
-			return fmt.Errorf("Could not grow underlying %q filesystem for %q: %s", fsType, devPath, msg)
+			logger.Error("Failed to grow filesystem", logger.Ctx{"filesystem": fsType, "device": devPath, "err": err})
```
tomponline, line 553: can we lose the error logging here as we are returning it as an error and it isnt contextual to the volume or instance.

lxd/storage/drivers/utils.go
```diff
@@ -537,21 +537,21 @@ func growFileSystem(fsType string, devPath string, vol Volume) error {
...
+			return fmt.Errorf("Could not grow underlying %q filesystem for %q: %s", fsType, devPath, err)
```
tomponline, line 554: Please use %w rather than %s when wrapping errors so that the original error type (if any) is accessible by the caller.

tomponline (approved):

PR #14132, opened 2024-09-19
Add API types for configuration metadata

markylaing:
This is so that we can manage configuration metadata with API extensions as we do with any other API response over `/1.0`

* Adds an API extension for adding entity metadata to `GET /1.0/configuration/metadata`.
* Adds types under `shared/api` for configuration metadata.
* Updates entity metadata such that each entity maps to a JSON object, with `requires_project` and `entitlements` fields.
* Enforces that generated metadata conforms to the new API type.
* Updates the swagger doc for `GET /1.0/configuration/metadata` to include the newly defined type.

### DIFF
--- a/lxd/lxd-metadata/lxd_metadata.go
+++ b/lxd/lxd-metadata/lxd_metadata.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/api"
 
 	"gopkg.in/yaml.v2"
 )
@@ -462,7 +463,7 @@ func writeDocFile(inputJSONPath, outputTxtPath string) error {
 		}
 	}
 
-	entities := make(map[string][]map[string]string)
+	entities := make(map[string]api.MetadataConfigurationEntity)
 	err = json.Unmarshal(jsonDoc.Entities, &entities)
 	if err != nil {
 		return err
@@ -476,11 +477,11 @@ func writeDocFile(inputJSONPath, outputTxtPath string) error {
 	sort.Strings(sortedEntityNames)
 
 	for _, entityName := range sortedEntityNames {
-		entitlements := entities[entityName]
+		entity := entities[entityName]
 		buffer.WriteString(fmt.Sprintf("<!-- entity group %s start -->\n", entityName))
-		for _, entitlement := range entitlements {
-			buffer.WriteString(fmt.Sprintf("`%s`\n", entitlement["name"]))
-			buffer.WriteString(fmt.Sprintf(": %s\n\n", entitlement["description"]))
+		for _, entitlement := range entity.Entitlements {
+			buffer.WriteString(fmt.Sprintf("`%s`\n", entitlement.Name))
+			buffer.WriteString(fmt.Sprintf(": %s\n\n", entitlement.Description))
 		}
 
 		buffer.WriteString("\n")
--- a/lxd/lxd-metadata/lxd_metadata.go
+++ b/lxd/lxd-metadata/lxd_metadata.go
@@ -13,7 +13,6 @@ import (
 	"path/filepath"
 	"regexp"
 	"sort"
-	"strconv"
 	"strings"
 	"time"
 
@@ -41,37 +40,6 @@ type IterableAny interface {
 type doc struct {
 	Configs  map[string]map[string]map[string][]any `json:"configs"`
 	Entities json.RawMessage                        `json:"entities"`
-}
-
-// detectType detects the type of a string and returns the corresponding value.
-func detectType(s string) any {
-	i, err := strconv.Atoi(s)
-	if err == nil {
-		return i
-	}
-
-	b, err := strconv.ParseBool(s)
-	if err == nil {
-		return b
-	}
-
-	f, err := strconv.ParseFloat(s, 64)
-	if err == nil {
-		return f
-	}
-
-	t, err := time.Parse(time.RFC3339, s)
-	if err == nil {
-		return t
-	}
-
-	// special characters handling
-	if s == "-" {
-		return ""
-	}
-
-	// If all conversions fail, it's a string
-	return s
 }
 
 // sortConfigKeys alphabetically sorts the entries by key (config option key) within each config group in an entity.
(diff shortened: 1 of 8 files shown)


kadinsayani (approved):
LGTM - thanks!

lxd/lxd-metadata/lxd_metadata.go
```diff
@@ -285,7 +253,7 @@ func parse(path string, outputJSONPath string, excludedPaths []string, substitut
 						continue
 					}
 
-					configKeyEntry[metadataMap["key"]].(map[string]any)[dataKVMatch[1]] = detectType(dataKVMatch[2])
+					configKeyEntry[metadataMap["key"]].(map[string]any)[dataKVMatch[1]] = dataKVMatch[2]
```
tomponline, line 257: I'm not sure about this change.

Where is the value in `dataKVMatch[2]` coming from?

As for how these are classified, whilst I agree that all config values are strings, my understanding of the metadata definition here is in order to indicate in the docs what sort of string values are allowed for a particular config key.

lxd/lxd-metadata/lxd_metadata.go
```diff
@@ -309,6 +309,13 @@ func parse(path string, outputJSONPath string, excludedPaths []string, substitut
 		return nil, fmt.Errorf("Error while marshaling project documentation: %v", err)
 	}
 
+	// Validate that what we've generated is valid against our API definition.
+	var metadataConfiguration api.MetadataConfiguration
```
tomponline, line 313: nice touch

tomponline (commented):
Apart from the data type change at the start, this seems all good.

lxd/lxd-metadata/lxd_metadata.go
markylaing: It looks like `dataKVMatch` is coming from this regular expression `(?m)([\S]+):[\s]+([\S \"\']+)` on the key value section of the `lxdmeta:generate` section of the doc block e.g.:
```
type: string
defaultdesc: `auto`
liveupdate: no
shortdesc: What to do when evacuating the instance
```
In this example, the first `dataKVMatch` outputted from `lxdDocDataRegex.FindAllStringSubmatch(data, -1)` would have
```
dataKVMatch[1] -> "type"
dataKVMatch[2] -> "string"
```

My feeling is that lxd-generate shouldn't be too clever about figuring out which config keys are which type. We should just include that in the description of the key. There is only one config key for which this appears to work.

tomponline: >  There is only one config key for which this appears to work.
which one?

markylaing: The commit directly after this change shows the result of `make update-metadata`. Only one field is changed: https://github.com/canonical/lxd/pull/14132/commits/2614bb5bba2ea25288f0f2b391f70c6eab7843e3#diff-d3459341f992192c3b18b6130aa7e165d305183794a84f219c7ab7512b1b31bcL4316

It is changing the value of `defaultdesc` from a boolean `true` to a string `"true"`.

tomponline:
needs a rebase.

@edlerd are you happy with this change btw?

markylaing:
> needs a rebase.

Done :+1:

edlerd:
> @edlerd are you happy with this change btw?

Happy is a big word for a change breaking our integration ;-) It looks all right to me.

One concern is the `project_specific` field being boolean. It seems not 100% correct, as some entities can be project specific or not, depending on how the project is set up. So IMHO for networks there should be a third option `depends` and the `project_specific` field be modelled as non-binary.

markylaing:
> > @edlerd are you happy with this change btw?
> 
> Happy is a big word for a change breaking our integration ;-) It looks all right to me.
> 
> One concern is the `project_specific` field being boolean. It seems not 100% correct, as some entities can be project specific or not, depending on how the project is set up. So IMHO for networks there should be a third option `depends` and the `project_specific` field be modelled as non-binary.

I'm a bit confused about this, does this relate to the `features.*` config on a project?

edlerd:
> I'm a bit confused about this, does this relate to the `features.*` config on a project?

Yes, exactly. You can share a network between all projects, so it is not really project specific. Or you can enable the `features.networks` option on a project, and then the networks created in that project are project specific.

markylaing:
> > I'm a bit confused about this, does this relate to the `features.*` config on a project?
> 
> Yes, exactly. You can share a network between all projects, so it is not really project specific. Or you can enable the `features.networks` option on a project, and then the networks created in that project are project specific.

I guess this is a difference in phrasing. When a network is "shared" it is just a network in the default project. It's still project specific because it is in the default project.

Let's say you create project "foo" with `features.networks=false` and `lxdbr0` is available. `lxdbr0` is still in the default project, so while `lxc network show lxdbr0 --project foo` would work, `lxc auth group permission add <group> network lxdbr0 can_view project=foo` would not work because the network is not truly in the `foo` project.

markylaing:
> > > I'm a bit confused about this, does this relate to the `features.*` config on a project?
> > 
> > 
> > Yes, exactly. You can share a network between all projects, so it is not really project specific. Or you can enable the `features.networks` option on a project, and then the networks created in that project are project specific.
> 
> I guess this is a difference in phrasing. When a network is "shared" it is just a network in the default project. It's still project specific because it is in the default project.
> 
> Let's say you create project "foo" with `features.networks=false` and `lxdbr0` is available. `lxdbr0` is still in the default project, so while `lxc network show lxdbr0 --project foo` would work, `lxc auth group permission add <group> network lxdbr0 can_view project=foo` would not work because the network is not truly in the `foo` project.

Project features have been a real headache for fine-grained authorization for this reason. I've added a warning to the documentation on this: https://documentation.ubuntu.com/lxd/en/latest/explanation/projects/#isolation-of-projects.

tomponline:
> Happy is a big word for a change breaking our integration ;-) It looks all right to me.

Yes apologies for this, but good that it was caught before it was released so its not a breaking change for external consumers.

tomponline:
> You can share a network between all projects, so it is not really project specific

Networks are project specific, but ones in the default project can be viewed/accessed in non-default projects based on the project's features.

tomponline (approved):

PR #14135, opened 2024-09-19
lxd/image: Revert "Acquire image lock for uploaded images"

MggMuggins:
This reverts commit 8a59f7fd1a21f078d6f47cb7b7bfbfd35bd5bb2a.

Only going to revert this one to make sure that the locking is the cause of the test failures.


tomponline (approved):

tomponline:
not needed as the issue is still present.

PR #14136, opened 2024-09-19
lxc: Fix bulk unfreeze (from Incus)

kadinsayani:
Cherry-picked from https://github.com/lxc/incus/pull/387. Tested locally ✅.

### DIFF
--- a/lxd/instances_put.go
+++ b/lxd/instances_put.go
@@ -143,7 +143,7 @@ func instancesPut(d *Daemon, r *http.Request) response.Response {
 			}
 
 		case instancetype.Unfreeze:
-			if inst.IsRunning() {
+			if !inst.IsFrozen() {
 				continue
 			}
 		}


MggMuggins:
@kadinsayani Some additional context in https://github.com/canonical/lxd/pull/13452, not sure if we want to introduce a `lxc resume` since I think `lxc start` is overloaded to do the same.

kadinsayani:
> @kadinsayani Some additional context in #13452, not sure if we want to introduce a `lxc resume` since I think `lxc start` is overloaded to do the same.

Thanks for providing additional context on this one 😄. I'll wait for Tom's input on this and close the PR if we decide against adding `lxc resume`.

tomponline:
> Thanks for providing additional context on this one 😄. I'll wait for Tom's input on this and close the PR if we decide against adding `lxc resume`.

OK lets not introduce a new command as @MggMuggins says, but is there a bug to be fixed here anyway?

kadinsayani:
> OK lets not introduce a new command as @MggMuggins says, but is there a bug to be fixed here anyway?

Sounds good. Yes, I've rebased to only include the bug fix.

tomponline (approved):

PR #14137, opened 2024-09-19
lxc: Disallow launching aliased containers of instance type VM

kadinsayani:
An aliased vm image should not launch with `lxc launch <alias_name>` (unless the `--vm` flag is passed in).


test/suites/remote.sh
```diff
@@ -30,6 +30,11 @@ test_remote_url() {
   if [ -z "${LXD_OFFLINE:-}" ]; then
     lxc_remote remote add ubuntu1 https://cloud-images.ubuntu.com/releases --protocol=simplestreams
     lxc_remote remote add ubuntu2 https://cloud-images.ubuntu.com:443/releases --protocol=simplestreams
+
+    # Ensure that we can't launch an aliased VM image as a container
+    lxc_remote image copy ubuntu:noble local: --copy-aliases --vm
```
tomponline, line 35: we cant use external images in the main LXD test suite.

So this check should be moved to the https://github.com/canonical/lxd-ci/blob/main/tests/vm test which runs daily using edge snap

MusicDin: Just to add on this - In the LXD CI test suite, the remote `images:` is disabled in:
https://github.com/canonical/lxd-ci/blob/78076446370f6244f03ac0e12add906f537a0c9c/.github/workflows/tests.yml#L259-L261

If access to those images, you can re-enable the access within the test like this:
```sh
# Unblock images remote.
sed -i '/images\.lxd\.canonical\.com/d' /etc/hosts
```

tomponline: thanks Din

kadinsayani: I've moved the test to `lxc-ci` and opened https://github.com/canonical/lxd-ci/pull/294.

tomponline (approved):

PR #14145, opened 2024-09-23
Trivy vulnerability scanning

hamistao:
This adds two jobs to our workflows. One uses Trivy to scan LXD's dependencies and the other scans the snap.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -26,13 +26,23 @@ jobs:
         with:
           ref: main
 
+      - name: Create Trivy cache directory
+        run: mkdir -p /home/runner/vuln-cache
+
       - name: Run Trivy vulnerability scanner in repo mode
         uses: aquasecurity/trivy-action@master
         with:
           scan-type: "fs"
           format: "sarif"
           output: "trivy-lxd-repo-scan-results.sarif"
+          cache-dir: "/home/runner/vuln-cache"
           severity: "LOW,MEDIUM,HIGH,CRITICAL"
+
+      - name: Cache trivy and vulnerability database
+        uses: actions/cache/save@v4
+        with:
+          path: /home/runner/vuln-cache
+          key: trivy-cache
 
       - name: Upload Trivy scan results to GitHub Security tab
         uses: github/codeql-action/upload-sarif@v3
@@ -40,3 +50,55 @@ jobs:
           sarif_file: "trivy-lxd-repo-scan-results.sarif"
           sha: ${{ github.sha }}
           ref: refs/heads/main
+
+  trivy-snap:
+    name: Trivy vulnerability scanner - Snap
+    runs-on: ubuntu-22.04
+    needs: trivy-repo
+    strategy:
+      matrix:
+        version:
+          - "latest"
+          - "5.21"
+          - "5.0"
+          - "4.0"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          # Snap scanner results are placed on the the branch related to the scanned channel.
+          ref: ${{ (matrix.version == 'latest' && 'main') || format('stable-{0}', matrix.version) }}
+
+      - name: Restore cached Trivy and vulnerability database
+        uses: actions/cache/restore@v4
+        with:
+          path: /home/runner/vuln-cache
+          key: trivy-cache
+
+      - name: Download snap for scan
+        run: |
+          ls /home/runner/vuln-cache
+          snap download lxd --channel=${{ matrix.version }}/stable
+          unsquashfs ./lxd*.snap
+
+      - name: Run Trivy vulnerability scanner on the snap
+        uses: aquasecurity/trivy-action@master
+        with:
+          scan-type: "rootfs"
+          format: "sarif"
+          scan-ref: squashfs-root
+          output: "${{ matrix.version }}-stable.sarif"
+          cache-dir: "/home/runner/vuln-cache"
+          severity: "LOW,MEDIUM,HIGH,CRITICAL"
+
+      - name: Prepend channel to alert name
+        run: |
+          jq '.runs[].tool.driver.rules[] |= (.shortDescription.text |= "Snap scan - " + .)' ${{ matrix.version }}-stable.sarif > tmp.json
+          mv tmp.json ${{ matrix.version }}-stable.sarif
+
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: "${{ matrix.version }}-stable.sarif"
+          sha: ${{ github.sha }}
+          ref: refs/heads/${{ (matrix.version == 'latest' && 'main') || format('stable-{0}', matrix.version) }}
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -1,0 +1,42 @@
+name: Vulnerability Scanning with Trivy
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *'  # Test Trivy daily at midnight
+
+permissions:
+  contents: read
+  security-events: write # for uploading SARIF results to the security tab
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  trivy-repo:
+    name: Trivy vulnerability scanner - Repository
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: main
+
+      - name: Run Trivy vulnerability scanner in repo mode
+        uses: aquasecurity/trivy-action@master
+        with:
+          scan-type: "fs"
+          format: "sarif"
+          output: "trivy-lxd-repo-scan-results.sarif"
+          severity: "LOW,MEDIUM,HIGH,CRITICAL"
+
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: "trivy-lxd-repo-scan-results.sarif"
+          sha: ${{ github.sha }}
+          ref: refs/heads/main


hamistao:
image
This shows how the alerts are shown in the security page (those alerts shown are false positives and can be closed as such after this is merged)

.github/workflows/tests.yml
```diff
@@ -477,6 +477,38 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: main
+
+      - name: Run Trivy vulnerability scanner in repo mode
+        uses: aquasecurity/trivy-action@master
+        with:
+          scan-type: "fs"
+          ignore-unfixed: true
```
tomponline, line 496: what does this do?

.github/workflows/tests.yml
```diff
@@ -477,6 +477,38 @@ jobs:
...
+      - name: Get HEAD sha
+        run: |
+          SHA="$(git rev-parse HEAD)"
+          echo "head_sha=$SHA" >> "$GITHUB_ENV"
+
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: "trivy-lxd-repo-scan-results.sarif"
+          sha: ${{ env.head_sha }}
```
tomponline, line 509: can we use GITHUB_SHA here?

.github/workflows/tests.yml
hamistao: Whoops, this should be removed. It ignores alerts that aren't fixed in newer versions of the vulnerable dependency/lib, e.g. a vulnerability that can't be fixed by updating our dependencies/libs.

.github/workflows/tests.yml
hamistao: I wasn't aware of this variable, let me try it on my fork

.github/workflows/tests.yml
```diff
@@ -477,6 +477,38 @@ jobs:
           installTics: true
           calc: ALL
 
+  trivy-repo:
```
tomponline, line 480: Lets move this into a separate workflow called security.yml, as from what i can see here its not tied to a specific LXD build and is just parsing the source code (go mod) right?

hamistao: Yes, you are correct. Will do!

.github/workflows/tests.yml
```diff
@@ -509,6 +509,49 @@ jobs:
...
+          mkdir -p sarifs
+          VER=$(curl --silent -qI https://github.com/aquasecurity/trivy/releases/latest | awk -F '/' '/^location/ {print  substr($NF, 1, length($NF)-1)}');
+          wget https://github.com/aquasecurity/trivy/releases/download/${VER}/trivy_${VER#v}_Linux-64bit.tar.gz
+          tar -zxvf ./trivy_${VER#v}_Linux-64bit.tar.gz
+
+      - name: Run Trivy vulnerability scanner on the snap
+        run: |
+          snap download lxd --channel=${{ matrix.version }}/${{ matrix.risk }}
+          unsquashfs ./lxd*.snap
+          ./trivy rootfs ./squashfs-root/ --format sarif > ${{ matrix.version }}-${{ matrix.risk }}.sarif
```
tomponline, line 541: is trivy recognising the packages we have inside the snap, the trivy page mentions generation of an SBOM, can you try this too to check its producing a list of our snap deps and versions, otherwise it wont necessarily be able to find my vulnerabilities.

hamistao:
image
Another image showing some scans from the snap scanner (these alerts are fixed on latest channels)

tomponline:
> image Another image showing some alerts from the snap scanner (these alerts are fixed on `latest` channels)

why dont we set it up in each one of our source branches on this repo, but targeting `<series>/stable` as its the stable snaps we are interested most in.

hamistao:
> why dont we set it up in each one of our source branches on this repo, but targeting <series>/stable as its the stable snaps we are interested most in.

Works for me! Another option would also be appending the channel in the beginning of the alert name. For example: `5.0/stable - PyYAML: Incomplete fix for CVE-2020-1747`. That said, searching for branch should be simpler so I would still prefer the option of using the branches.

tomponline:
> > why dont we set it up in each one of our source branches on this repo, but targeting /stable as its the stable snaps we are interested most in.
> 
> Works for me! Another option would also be appending the channel in the beginning of the alert name. For example: `5.0/stable - PyYAML: Incomplete fix for CVE-2020-1747`. That said, searching for branch should be simpler so I would still prefer the option of using the branches.

can you do that too so its clear its coming from a snap alert

tomponline:
Ready for review?

hamistao:
@tomponline Now it is!

Example of alerts for the snap scanner, the repo scanner alerts are the same.

image

.github/workflows/tests.yml
hamistao: this unfortunately did not work :/

hamistao:
image

One more image to ilustrate one particular behavior, if more than one version has the same vulnerability, the most recent is shown in the alert title, I believe this is because the alert ID is the same so the title should stay constant across branches. They are still filtered by branch and have the indicator on the side so in my view this doesn't hurt the UX.

.github/workflows/tests.yml
tomponline: please advise what you use and why it didn't work?

.github/workflows/security.yml
```diff
@@ -17,6 +17,25 @@ defaults:
     shell: bash
 
 jobs:
+  set-up-trivy:
+    name: Set up Trivy
+    runs-on: ubuntu-22.04
+    steps:
+    - name: Setup Trivy vulnerability scanner
+      run: |
+        mkdir -p /home/runner/trivy
```
tomponline, line 26: this install process seems quite complicated, is there a reason we didnt use the deb?

https://aquasecurity.github.io/trivy/v0.55/getting-started/installation/

```
sudo apt-get install wget apt-transport-https gnupg
wget -qO - https://aquasecurity.github.io/trivy-repo/deb/public.key | gpg --dearmor | sudo tee /usr/share/keyrings/trivy.gpg > /dev/null
echo "deb [signed-by=/usr/share/keyrings/trivy.gpg] https://aquasecurity.github.io/trivy-repo/deb generic main" | sudo tee -a /etc/apt/sources.list.d/trivy.list
sudo apt-get update
sudo apt-get install trivy
```

tomponline:
Does the repo scan generate the vuln cache dir? If so can we get it to update the cache when it finishes and then have the snap task depend on it?

hamistao:
> Does the repo scan generate the vuln cache dir? If so can we get it to update the cache when it finishes and then have the snap task depend on it?

Yes, we can!

.github/workflows/security.yml
hamistao: Not much, I just figured that since we are caching the vuln DB, why not caching trivy itself as well. Also it should be little faster, maybe this is why the k8s team used it like this.

Since I will be caching the vuln DB in the repo scan, and that uses the GitHub action so it doesn't download Trivy, I plan on using the deb in each `test-snap` run.

tomponline: I would suggest we use the deb for both runs for consistency, we are not speed sensitive on this run.

hamistao:
@tomponline @simondeziel Requested changes made

image
Updated alerts above

.github/workflows/security.yml
```diff
@@ -26,17 +26,79 @@ jobs:
...
+      - name: Download snap for scan
+        run: |
+          ls /home/runner/vuln-cache
+          snap download lxd --channel=${{ matrix.version }}/stable
+          unsquashfs ./lxd*.snap
+
+      - name: Run Trivy vulnerability scanner on the snap
+        uses: aquasecurity/trivy-action@master
+        with:
+          scan-type: "rootfs"
```
tomponline, line 87: where do you instruct trivy to scan the uncompressed snap dir?

.github/workflows/security.yml
```diff
@@ -26,17 +26,79 @@ jobs:
...
+    strategy:
+      matrix:
+        version:
+          - "latest"
+          - "5.21"
+          - "5.0"
+          - "4.0"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
tomponline, line 67: why do we need to checkout the source code of the repository if we are scanning the snap?

.github/workflows/security.yml
hamistao: A few lines below, in the scan-ref field. `squashfs-root` is the name of the uncompressed snap dir.

```suggestion
      - name: Run Trivy vulnerability scanner on the snap
        uses: aquasecurity/trivy-action@master
        with:
          scan-type: "rootfs"
          format: "sarif"
          scan-ref: squashfs-root
          output: "${{ matrix.version }}-stable.sarif"
```

.github/workflows/security.yml
hamistao: To get the correct `github.sha` when uploading the alerts to the security tab and upload the results to the correct branch. See [this](https://github.com/canonical/lxd/pull/14145/commits/5f9f0cf88ac7e96ffe02ff08a291e0d707fbdbb0#diff-6102d4f2cddb56c446459bde9bf8e8044b87ce3f98a24c2bba99debfd62461dcR103).

.github/workflows/security.yml
tomponline: so it does, thanks for highlighting!

.github/workflows/security.yml
tomponline: I see thanks!

tomponline: Do we need to submit a sha, as its not really accurate at all to take the head of the branch and relate that to a scan of the stable snap, they could be months different.

.github/workflows/security.yml
```diff
@@ -0,0 +1,104 @@
...
+      - name: Prepend channel to alert name
+        run: |
+          jq '.runs[].tool.driver.rules[] |= (.shortDescription.text |= "Snap scan - " + .)' ${{ matrix.version }}-stable.sarif > tmp.json
+          mv tmp.json ${{ matrix.version }}-stable.sarif
+
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: "${{ matrix.version }}-stable.sarif"
+          sha: ${{ github.sha }}
```
tomponline, line 103: do we need to provide this line at all?

hamistao: I also find it confusing but I get an error if I don't provide both `sha` and `ref`

hamistao:
The PRs for microcloud and microcluster will be ready soon.

.github/workflows/security.yml
tomponline: can parse the version from the snap somehow, and use the ref of the associated tag perhaps?

PR #14146, opened 2024-09-23
Disable tests that are failing because of `resize2fs` kernel bug

markylaing:
[This kernel bug](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2081231) is causing CI failures for Ceph and LVM. This PR disables those tests *for now*. We need to re-enable these tests when the kernel bug is fixed.


markylaing:
Marked as draft to see the result of the workflow first, if many other tests need to be disabled, we'll turn those off for ceph and LVM only.

simondeziel:
> Marked as draft to see the result of the workflow first, if many other tests need to be disabled, we'll turn those off for ceph and LVM only.

I **think** that anything that involves a block device and ext4 resizing risks triggering the issue. This would mean that creating any instance on top of LVM and Ceph is at risk, unless we can avoid the resizing part but that seems to big for a quick workaround.

.github/workflows/tests.yml
```diff
@@ -230,13 +230,28 @@ jobs:
...
+          - event_type: schedule
+            backend: lvm
+          - event_type: pull_request
+            backend: ceph
+          - event_type: schedule
+            backend: ceph
+          - event_type: pull_request
+            backend: random
+          - event_type: schedule
+            backend: random
```
markylaing, line 254: @tomponline @simondeziel I can't figure out why these exclude rules aren't working. (Also, I'm aware we only want to disable ceph and lvm on `push` events, I just wanted to test it works first in a PR)

tomponline: Try adding something like `if: github.event_name == 'pull_request'` on the workflow step itself

tomponline: you should be able to expand the logic to consult the matrix

markylaing: Ok thanks :)

markylaing: Turns out I was using `github.event.name` and not `github.event_name` :facepalm:

markylaing:
@tomponline @simondeziel this works as is so I can change `pull_request` to `push` OR I can go back to excluding the event type in the matrix, which will stop all of the additional steps (e.g. `remove_docker`) from running. Let me know what you'd prefer.

markylaing:
@tomponline @simondeziel I've opted with the matrix exclusions as I think it's easier to read and also means we're not using unnecessary compute.

tomponline:
> @tomponline @simondeziel I've opted with the matrix exclusions as I think it's easier to read and also means we're not using unnecessary compute.

Ok but we need to exclude "schedule" not just "push" for the tiobe daily checks

tomponline (approved):
Merging this for now

markylaing:
> > @tomponline @simondeziel I've opted with the matrix exclusions as I think it's easier to read and also means we're not using unnecessary compute.
> 
> Ok but we need to exclude "schedule" not just "push" for the tiobe daily checks

Yep this excludes schedule as well 👍

PR #14148, opened 2024-09-23
Add support for binfmt misc namespace 

kadinsayani:
Cherry-picked from https://github.com/lxc/incus/pull/474.


lxd/instance/drivers/driver_lxc.go
```diff
@@ -800,6 +799,11 @@ func (d *lxc) initLXC(config bool) (*liblxc.Container, error) {
 		"/sys/kernel/tracing",
 	}
 
+	// Handle unprivileged binfmt_misc.
+	if d.IsPrivileged() || !d.state.OS.UnprivBinfmt {
```
kadinsayani, line 803: At this point, I'm only able to get `binfmt` support working after changing this check to `if !d.IsPrivileged || !d.state.OS.UnprivBinFmt`. Based on my understanding, that should be the correct check - @mihalicyn can you please advise?

kadinsayani:
> Is it possible/viable to add a test for this?

Certainly, should be a pretty straight forward test.

lxd/instance/drivers/driver_lxc.go
mihalicyn: `d.IsPrivileged() || !d.state.OS.UnprivBinfmt` is a correct condition. See, first of all, we remove `"/proc/sys/fs/binfmt_misc"` from the `bindMounts` array. Then we put it back to the same array but only if either container is privileged (in this case binfmt_misc can not be containerized) OR if kernel does not support binfmt_misc namespacing. No change is needed there.

mihalicyn: Now let's try to figure out what precisely does not work for you when you test this PR in its current state?

mihalicyn:
This looks good and correct to me.

mihalicyn (approved):

lxd/instance/drivers/driver_lxc.go
kadinsayani: Here's what I was originally testing:

```
kadinsayani@devbox:~/canonical/lxd$ uname -a
Linux devbox 6.8.0-45-generic #45-Ubuntu SMP PREEMPT_DYNAMIC Fri Aug 30 12:02:04 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

kadinsayani@devbox:~/canonical/lxd$ lxc launch ubuntu:24.04 c1
Launching c1

kadinsayani@devbox:~/canonical/lxd$ ls -l /proc/sys/fs/binfmt_misc
total 0
-rw-r--r-- 1 root root 0 Sep 25 21:15 python3.12
--w------- 1 root root 0 Sep 25 21:15 register
-rw-r--r-- 1 root root 0 Sep 25 21:15 status
kadinsayani@devbox:~/canonical/lxd$ lxc exec c1 -- ls -l /proc/sys/fs/binfmt_misc
total 0

kadinsayani@devbox:~/canonical/lxd$ lxc exec c1 -- mount | grep binfmt
```

Thanks for clarifying that we shouldn't be mounting `binfmt_misc` from the host if the guest OS supports it; that makes sense to me.

I've tested enabling `binfmt_misc` from within a container and it works as expected:

```
ubuntu@c1:~$ sudo mount binfmt_misc -t binfmt_misc /proc/sys/fs/binfmt_misc
ubuntu@c1:~$ cat /proc/sys/fs/binfmt_misc/status
enabled
ubuntu@c1:~$ uname -a
Linux c1 6.8.0-45-generic #45-Ubuntu SMP PREEMPT_DYNAMIC Fri Aug 30 12:02:04 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

I think we’re all set for this! Just wanted to double-check that we’ve covered everything. 😄

kadinsayani: For reference, there are details in the docs on auto mounting on boot: https://docs.kernel.org/admin-guide/binfmt-misc.html.

tomponline:
@kadinsayani can we add a test for this (to the test suite in this repo)?

kadinsayani:
> @kadinsayani can we add a test for this (to the test suite in this repo)?

Sure thing!

doc/container-environment.md
```diff
@@ -62,7 +62,7 @@ The following mounts are set up by default:
 
 If they are present on the host, the following paths will also automatically be mounted:
 
-- `/proc/sys/fs/binfmt_misc`
+- `/proc/sys/fs/binfmt_misc` (only on kernels that lack binfmt_misc namespace support)
```
simondeziel, line 65: If that `binfmt_misc` bit inside the parenthesis is what prompted you to add it to the custom word list, I think it'd be better to just surround it with backticks which is enough to get the spellchecker to stop looking for typos.

test/suites/basic.sh
```diff
@@ -374,6 +373,12 @@ test_basic_usage() {
   lxc list | grep foo | grep RUNNING
   lxc stop foo --force
 
+  # Test binfmt_misc support
+  lxc start foo
+  lxc exec foo -- mount -t binfmt_misc none /proc/sys/fs/binfmt_misc
+  [ "$(lxc exec foo -- cat /proc/sys/fs/binfmt_misc/status)" = "enabled" ]
+  lxc stop foo
```
simondeziel, line 380: `lxc stop -f foo` would be OK here as we don't care about an orderly shutdown.

lxd/apparmor/instance_lxc.go
```diff
@@ -28,7 +28,9 @@ profile "{{ .name }}" flags=(attach_disconnected,mediate_deleted) {
 
   # Handle binfmt
   mount fstype=binfmt_misc -> /proc/sys/fs/binfmt_misc/,
-  deny /proc/sys/fs/binfmt_misc/{,**} rwklx,
+  {{- if not .kernel_binfmt }}
+    deny /proc/sys/fs/binfmt_misc/{,**} rwklx,
```
simondeziel, line 32: Since it's akin to a preprocessor `if`, I wouldn't indent the body as it would stick out... cosmetic only OFC.

tomponline:
@kadinsayani please ensure you add "(from Incus)" to PR title in the future. Ta

tomponline (approved):

PR #14149, opened 2024-09-23
lxc: Prevent accept-certificate flag when using trust token

### LINKED ISSUE #13891: Improve remote token handling
When trying to add a remote using a token, I got this unexpected prompt for fingerprint verification:

```bash
$ lxc remote add joliet 10.245.130.24 --token eyJjbGllbnRfbmFtZSI6InNkZXppZWwiLCJmaW5nZXJwcmludCI6IjQ2MzVlZjZkN2I0OTFlNjUxNTRiYjljYzA3MGMxZTJhMGFkYzM4NmViNGE3NGU1MWMxNzk2YmRkNDgzYmRmYWQiLCJhZGRyZXNzZXMiOlsiMTAuMS4xMC4xMDU6ODQ0MyIsIjEwLjI0NS4xMzAuMjQ6ODQ0MyIsIjEwLjY5LjIzLjE6ODQ0MyIsIltmZDQyOmE3Yzk6N2NkZjo4NGQ6OjFdOjg0NDMiXSwic2VjcmV0IjoiMGM3OWFkZTE5YjVkYTQ5YTNhMTAwNWZjMmJmOTk3Njc4MjFjZTg3MWI4NTgwOWE0ZGU4NjYxZGZjMDgwMjg4ZiIsImV4cGlyZXNfYXQiOiIwMDAxLTAxLTAxVDAwOjAwOjAwWiJ9
Certificate fingerprint: 4635ef6d7b491e65154bb9cc070c1e2a0adc386eb4a74e51c1796bdd483bdfad
ok (y/n/[fingerprint])? n
Error: Server certificate NACKed by user
```

The token embeds the fingerprint so why ask for verification?

```bash
$ base64 -d <<< 'eyJjbGllbnRfbmFtZSI6InNkZXppZWwiLCJmaW5nZXJwcmludCI6IjQ2MzVlZjZkN2I0OTFlNjUxNTRiYjljYzA3MGMxZTJhMGFkYzM4NmViNGE3NGU1MWMxNzk2YmRkNDgzYmRmYWQiLCJhZGRyZXNzZXMiOlsiMTAuMS4xMC4xMDU6ODQ0MyIsIjEwLjI0NS4xMzAuMjQ6ODQ0MyIsIjEwLjY5LjIzLjE6ODQ0MyIsIltmZDQyOmE3Yzk6N2NkZjo4NGQ6OjFdOjg0NDMiXSwic2VjcmV0IjoiMGM3OWFkZTE5YjVkYTQ5YTNhMTAwNWZjMmJmOTk3Njc4MjFjZTg3MWI4NTgwOWE0ZGU4NjYxZGZjMDgwMjg4ZiIsImV4cGlyZXNfYXQiOiIwMDAxLTAxLTAxVDAwOjAwOjAwWiJ9' | jq -r
{
  "client_name": "sdeziel",
  "fingerprint": "4635ef6d7b491e65154bb9cc070c1e2a0adc386eb4a74e51c1796bdd483bdfad",
  "addresses": [
    "10.1.10.105:8443",
    "10.245.130.24:8443",
    "10.69.23.1:8443",
    "[fd42:a7c9:7cdf:84d::1]:8443"
  ],
  "secret": "0c79ade19b5da49a3a1005fc2bf99767821ce871b85809a4de8661dfc080288f",
  "expires_at": "0001-01-01T00:00:00Z"
}
```

Also, the IP address needs to be provided despite being one of the `addresses`. Here's what happens if the IP isn't provided when adding the remote:

```
$ lxc remote add joliet --token eyJjbGllbnRfbmFtZSI6InNkZXppZWwiLCJmaW5nZXJwcmludCI6IjQ2MzVlZjZkN2I0OTFlNjUxNTRiYjljYzA3MGMxZTJhMGFkYzM4NmViNGE3NGU1MWMxNzk2YmRkNDgzYmRmYWQiLCJhZGRyZXNzZXMiOlsiMTAuMS4xMC4xMDU6ODQ0MyIsIjEwLjI0NS4xMzAuMjQ6ODQ0MyIsIjEwLjY5LjIzLjE6ODQ0MyIsIltmZDQyOmE3Yzk6N2NkZjo4NGQ6OjFdOjg0NDMiXSwic2VjcmV0IjoiMGM3OWFkZTE5YjVkYTQ5YTNhMTAwNWZjMmJmOTk3Njc4MjFjZTg3MWI4NTgwOWE0ZGU4NjYxZGZjMDgwMjg4ZiIsImV4cGlyZXNfYXQiOiIwMDAxLTAxLTAxVDAwOjAwOjAwWiJ9
Error: Get "https://joliet:8443": lookup joliet on 127.0.0.53:53: server misbehaving
```

IMHO, all addresses should be tried in parallel or in quick succession.

### LINKED ISSUE #13973: `lxc remote add` should refuse `--accept-certificate` when provided with a token
Remote add tokens were designed to allow securely onboarding clients while avoiding the risk of MITM. Since the token consumed by the client embeds the fingerprint of the expected server certificate, there is really no reason to use `--accept-certificate`.

Accepting both a token and `--accept-certificate` is confusing and might imply (I didn't check the code) that LXD doesn't do the fingerprint verification. LXD should error out to clearly communicate that this combination is insecure and unneeded.

MusicDin:
This PR prevents `--accept-certificate` to be used with a trust token. Whenever the token is provided, the fingerprint from the token is compared to the remote certificate fingerprint.

In addition, validation for mutually exclusive flags is added, more specifically trust token cannot be used with public remotes or when auth-type is OIDC.

Finally, two minor fixes; do not stop interactive mode on wrong input, and do not fail if remote is already trusting us. 

Closes #13973
Closes #13891

### DIFF
--- a/lxc/remote.go
+++ b/lxc/remote.go
@@ -294,6 +294,16 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
 		return errors.New(i18n.G("Remote address must not be empty"))
 	}
 
+	// Trust token cannot be used when auth type is set to OIDC.
+	if c.flagToken != "" && c.flagAuthType == "oidc" {
+		return errors.New(i18n.G("Trust token cannot be used with OIDC authentication"))
+	}
+
+	// Trust token cannot be used for public remotes.
+	if c.flagToken != "" && c.flagPublic {
+		return errors.New(i18n.G("Trust token cannot be used for public remotes"))
+	}
+
 	// Validate the server name.
 	if strings.Contains(server, ":") {
 		return errors.New(i18n.G("Remote names may not contain colons"))
--- a/lxc/remote.go
+++ b/lxc/remote.go
@@ -153,8 +153,13 @@ func (c *cmdRemoteAdd) findProject(d lxd.InstanceServer, project string) (string
 	return project, nil
 }
 
-func (c *cmdRemoteAdd) runToken(server string, token string, rawToken *api.CertificateAddToken) error {
+func (c *cmdRemoteAdd) runToken(addr string, server string, token string, rawToken *api.CertificateAddToken) error {
 	conf := c.global.conf
+
+	// Certificate cannot be blindly accepted when using a trust token.
+	if c.flagAcceptCert {
+		return errors.New(i18n.G("The --accept-certificate flag is not supported when adding a remote using a trust token"))
+	}
 
 	if !conf.HasClientCertificate() {
 		fmt.Fprint(os.Stderr, i18n.G("Generating a client certificate. This may take a minute...")+"\n")
@@ -164,6 +169,12 @@ func (c *cmdRemoteAdd) runToken(server string, token string, rawToken *api.Certi
 		}
 	}
 
+	// If address is provided, use token on that specific address.
+	if addr != "" {
+		return c.addRemoteFromToken(addr, server, token, rawToken.Fingerprint)
+	}
+
+	// Otherwise, iterate over all addresses within the token.
 	for _, addr := range rawToken.Addresses {
 		addr = fmt.Sprintf("https://%s", addr)
 
@@ -179,6 +190,7 @@ func (c *cmdRemoteAdd) runToken(server string, token string, rawToken *api.Certi
 		return nil
 	}
 
+	// Finally, fallback to manual input.
 	fmt.Println(i18n.G("All server addresses are unavailable"))
 	fmt.Print(i18n.G("Please provide an alternate server address (empty to abort):") + " ")
 
@@ -304,6 +316,11 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
 		return errors.New(i18n.G("Trust token cannot be used for public remotes"))
 	}
 
+	// Certificate cannot be blindly accepted when using a trust token.
+	if c.flagToken != "" && c.flagAcceptCert {
+		return errors.New(i18n.G("The --accept-certificate flag is not supported when adding a remote using a trust token"))
+	}
+
 	// Validate the server name.
 	if strings.Contains(server, ":") {
 		return errors.New(i18n.G("Remote names may not contain colons"))
@@ -329,9 +346,11 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
 		conf.Remotes = map[string]config.Remote{}
 	}
 
+	// Check if the first argument is a trust token. In such case, we need to
+	// decode it and use it to connect to the remote.
 	rawToken, err := shared.CertificateTokenDecode(addr)
 	if err == nil {
-		return c.runToken(server, addr, rawToken)
+		return c.runToken("", server, addr, rawToken)
 	}
 
 	// Complex remote URL parsing
@@ -445,6 +464,16 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
 
 		conf.Remotes[server] = remote
 		return conf.SaveConfig(c.global.confPath)
+	}
+
+	// Handle adding a remote with trust token.
+	if c.flagToken != "" {
+		rawToken, err := shared.CertificateTokenDecode(c.flagToken)
+		if err != nil {
+			return fmt.Errorf(i18n.G("Failed to decode trust token: %w"), err)
+		}
+
+		return c.runToken(addr, server, c.flagToken, rawToken)
 	}
 
 	// Check if the system CA worked for the TLS connection
@@ -594,11 +623,9 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
 			// use the token instead and prompt for it if not present.
 			if d.(lxd.InstanceServer).HasExtension("explicit_trust_token") && c.flagPassword == "" {
 				// Prompt for trust token.
-				if c.flagToken == "" {
-					c.flagToken, err = c.global.asker.AskString(fmt.Sprintf(i18n.G("Trust token for %s: "), server), "", nil)
-					if err != nil {
-						return err
-					}
+				c.flagToken, err = c.global.asker.AskString(fmt.Sprintf(i18n.G("Trust token for %s: "), server), "", nil)
+				if err != nil {
+					return err
 				}
 
 				req.TrustToken = c.flagToken
--- a/lxc/remote.go
+++ b/lxc/remote.go
@@ -449,22 +449,39 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
 
 	// Handle certificate prompt
 	if certificate != nil {
+		// Prompt for certificate acceptance if user did not allow us to blindly
+		// accept the remote certificate.
 		if !c.flagAcceptCert {
 			digest := shared.CertFingerprint(certificate)
 
 			fmt.Printf("%s: %s\n", i18n.G("Certificate fingerprint"), digest)
 			fmt.Print(i18n.G("ok (y/n/[fingerprint])?") + " ")
-			line, err := shared.ReadStdin()
-			if err != nil {
-				return err
-			}
+			for {
+				line, err := shared.ReadStdin()
+				if err != nil {
+					return err
+				}
 
-			if string(line) != digest {
+				// Continue with adding the remote if digest matches, or the user
+				// confirmed a fingerprint.
+				if string(line) == digest || strings.ToLower(string(line[0])) == i18n.G("y") {
+					break
+				}
+
+				// If the input length matches the certificate fingerprint length
+				// but the fingerprints do not match, return an error. This ensures
+				// the scripts do not hang if incorrect fingerprint is provided.
+				if len(line) == len(digest) {
+					return errors.New(i18n.G("The provided fingerprint does not match the server certificate fingerprint"))
+				}
+
+				// Error out if the user didn't confirm the fingerprint.
 				if len(line) < 1 || strings.ToLower(string(line[0])) == i18n.G("n") {
 					return errors.New(i18n.G("Server certificate NACKed by user"))
-				} else if strings.ToLower(string(line[0])) != i18n.G("y") {
-					return errors.New(i18n.G("Please type 'y', 'n' or the fingerprint:"))
 				}
+
+				// Ask again for any other invalid input.
+				fmt.Print(i18n.G("Please type 'y', 'n' or the fingerprint: "))
 			}
 		}
 
--- a/lxc/remote.go
+++ b/lxc/remote.go
@@ -267,9 +267,22 @@ func (c *cmdRemoteAdd) addRemoteFromToken(addr string, server string, token stri
 		req.Password = token
 	}
 
+	// Add client certificate to trust store. Even if we are already trusted (src.Auth == "trusted"),
+	// we want to send the token to invalidate it. Therefore, we can ignore the conflict error, which
+	// is thrown if we are trying to add a client cert that is already trusted by LXD remote.
 	err = d.CreateCertificate(req)
+	if err != nil && !api.StatusErrorCheck(err, http.StatusConflict) {
+		return err
+	}
+
+	// And check if trusted now.
+	srv, _, err := d.GetServer()
 	if err != nil {
-		return fmt.Errorf(i18n.G("Failed to create certificate: %w"), err)
+		return err
+	}
+
+	if srv.Auth != "trusted" {
+		return errors.New(i18n.G("Server doesn't trust us after authentication"))
 	}
 
 	// Handle project.
--- a/lxd/certificates.go
+++ b/lxd/certificates.go
@@ -478,6 +478,28 @@ func certificatesPost(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(fmt.Errorf("Failed to get authentication status: %w", err))
 	}
 
+	// If caller is already trusted and the trust token is provided, we validate the token and cancel
+	// the corresponding token operation.
+	if trusted && req.TrustToken != "" {
+		// Decode the trust token.
+		joinToken, err := shared.CertificateTokenDecode(req.TrustToken)
+		if err != nil {
+			return response.Forbidden(nil)
+		}
+
+		// Validate the token and cancel the corresponding operation.
+		tokenReq, err := certificateTokenValid(s, r, joinToken)
+		if err != nil {
+			return response.InternalError(fmt.Errorf("Failed during search for certificate add token operation: %w", err))
+		}
+
+		if tokenReq == nil {
+			return response.Forbidden(fmt.Errorf("No matching certificate add operation found"))
+		}
+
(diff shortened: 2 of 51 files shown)


test/suites/pki.sh
```diff
@@ -189,9 +189,8 @@ test_pki() {
     # Remove remote
     lxc_remote remote remove pki-lxd
 
-    # Add the remote again using an incorrect token.
-    # This should succeed as is the same as the test above but with an incorrect token rather than no token.
-    lxc_remote remote add pki-lxd "${LXD5_ADDR}" --accept-certificate --token=bar
+    # Ensure remote can be added without the token, as we are already trusted by the remote.
+    echo "y" | lxc_remote remote add pki-lxd "${LXD5_ADDR}"
```
simondeziel, line 193: The remote still trusts the client but the client deleted the `remote.crt` so it has to prompt for `y/fingerprint` before proceeding.

Manually `echo`'ing `y` feel like emulating `--accept-certificate`.

simondeziel (dismissed):

test/suites/pki.sh
MusicDin: Correct, should use `--accept-certificate` instead.

simondeziel (approved):

lxc/remote.go
tomponline, line 1: Please can you add more detail to the commit message explaining what you fixed.

tomponline, line 1: > If a remote is added, and then removed from the local config (lxc remote delete)
and the certificate remains trusted on the remote server, when remote is added
again it will fail because it is already trusted.

Is this the current state now and what you've fixed, or what you are changing it to be?

I struggled to understand the change here.

tomponline: Because:

>  when remote is added again it will fail because it is already trusted.

Sounds like the wrong behaviour to me, so wanted to understand if this is new or what is being fixed?

lxc/remote.go
```diff
@@ -322,6 +329,11 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
 		return errors.New(i18n.G("Trust token cannot be used for public remotes"))
 	}
 
+	// Certificate cannot be blindly accepted when using a trust token.
```
tomponline, line 332: Please can you update the comment to explain the *why* of this statement.

lxc/remote.go
```diff
@@ -347,9 +359,16 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
 		conf.Remotes = map[string]config.Remote{}
 	}
 
+	// Check if the first argument is a trust token. In such case, we need to
+	// decode it and use it to connect to the remote.
 	rawToken, err := shared.CertificateTokenDecode(addr)
 	if err == nil {
-		return c.runToken(server, addr, rawToken)
+		// Certificate cannot be blindly accepted when using a trust token.
+		if c.flagAcceptCert {
```
tomponline, line 367: can we re-structure this commit to avoid needing to check the same logic twice and duplicating the error message?

test/includes/setup.sh
tomponline, line 1: is accept-token still tested anywhere?

test/suites/clustering.sh
```diff
@@ -2763,7 +2763,7 @@ test_clustering_image_refresh() {
   LXD_DIR="${LXD_REMOTE_DIR}" lxc config set core.https_address "10.1.1.104:8443"
 
   # Add remotes
-  lxc remote add public "https://10.1.1.104:8443" --accept-certificate --token foo --public
+  lxc remote add public "https://10.1.1.104:8443" --accept-certificate --public
```
tomponline, line 2766: Can we add tests for not being able to use --accept-certificate with tokens please?

lxc/remote.go
tomponline: Does `lxc remote add test` work after doing `lxc remote rm test`?

MusicDin: Yes, if target URL is specified:
```
lxc remote add test 10.70.164.239
```

tomponline: Got you. Will the new token still get expired once its used (even though its not really used)?

MusicDin: No, the token is not consumed.

tomponline: Should it not be? From the user's PoV we've used it.

MusicDin: Yes, I think that makes sense. Will fix that.

MusicDin:
Moving to draft until I address all of the above comments.

lxc/remote.go
MusicDin: @markylaing @tomponline Regarding our conversation, here is the short explanation of the issue with tokens.

Previously, before preventing `--token` and `--accept-certificate` to be used together, when token was used along `--accept-certificate`, the token was not used to check the remote certificate and not invalidated if remote already trusted us, so this error never occurred.

When we try to add remote using a trust token and the remote already trusts us, the returned error is "Client is already trusted". 

However, if the user also has permissions to create new certificates the remote will assume we are trying to add new certificate or request a new token. In this case the resulting error is:
```
$ lxc remote add test <url> --token <token>
Error: Failed to create certificate: Invalid certificate type
```

As it can be seen in this test run: https://github.com/canonical/lxd/actions/runs/11217715833/job/31180544404#step:12:17878

---

The above issues are addressed solely in commit 0998460713e3369477c836348237bd24aaae87df, which as per discussion may open more issues, such as: What if the client has new client certificate? But wouldn't in such case the client be already considered untrusted (which means the above scenarios would not be the issue in the first place)?

Alternatively, I can drop that commit and fix the tests where `--token` is used when adding a remote that already trust us.

MusicDin:
@markylaing Please see the comment https://github.com/canonical/lxd/pull/14149#discussion_r1791689403

lxc/remote.go
markylaing: I'm not clear why the token should be invalidated in this case.

I think that - rather than checking if the caller has permission to create certificates, we should only check if they are authenticated. If they are not trusted, a token must be provided. Otherwise a token cannot be provided.

tomponline: please can we decouple the change wrt to using tokens on with an existing trusted client from the rest of the changes in this PR so we can get the other changes merged. Ta

markylaing: So to clarify I think:
```go
if auth.IsTrusted(r.Context()) && req.Token != "" {
    return response.BadRequest(errors.New("Client is already trusted"))
}
```

I'd blanket deny all trusted requests that use a token, because that's not what a token is supposed to be used for. Since the client is already trusted they do not need to provide a token. And `lxc config trust add <name> <address>` will work. If the user does not know the address (e.g. it's in the token), the `remote add` command could ignore the error and continue to add the remote.

MusicDin: > I think that - rather than checking if the caller has permission to create certificates, we should only check if they are authenticated. If they are not trusted, a token must be provided. Otherwise a token cannot be provided.

We wanted to allow adding a remote using valid token, which previously allowed passing any token - even invalid (if token was passed as positional argument).

To me your suggestions seems a great solution, as we provide the user with a meaningful error if he passes the token, while avoiding potential issues when invalidating tokens.

tomponline: > add

Isn't that back to square one, with leaving the token unconsumed on the server?

test/suites/pki.sh
```diff
@@ -200,9 +200,8 @@ test_pki() {
     # Remove remote
     lxc_remote remote remove pki-lxd
 
-    # Add the remote again using an incorrect token.
-    # This should succeed as is the same as the test above but with an incorrect token rather than no token.
-    lxc_remote remote add pki-lxd "${LXD5_ADDR}" --token=bar
+    # Ensure remote can be added without the token, as we are already trusted by the remote.
+    lxc_remote remote add pki-lxd "${LXD5_ADDR}" --accept-certificate
```
markylaing, line 204: This comment is a little confusing. "trusted by the remote" implies that we're in the truststore of the remote, but we aren't. We're trusted by virtue of having our certificate signed by the CA and `core.trust_ca_certificates` being enabled.

Instead of re-testing this, I think a better test would be to ensure `--accept-certificate` is not needed if the CA certificate is in the client's config directory.

lxc/remote.go
MusicDin: Yes, but @markylaing has good point that the address can be extracted from the token if the user does not specify it.

I think we have two options:
- Just check if the server already trusts us, and if token is used print "token not consumed"
- Or invalidate the token (current implementation). But if there are concerns with this approach we can still go with the first one.

IMO, any of the above options is better then error `Error: Failed to create certificate: Invalid certificate type` when adding a remote server.

markylaing: > > add
> 
> Isn't that back to square one, with leaving the token unconsumed on the server?

Ok middle ground. If a token is provided, always check and invalidate it. If the caller is already trusted, then it has been "misused" but the caller may not know that. So check for the "Client is already trusted" error in `remote add` and continue adding the remote.

test/suites/pki.sh
MusicDin: Since the same test is a few lines above I've just removed it.

The error when adding a remote, it fails because the certificate is signed by an unknown authority. Therefore, `lxc` falls back to asking whether the certificate should be accepted. Flag `--accept-certificate` papers over the issue by automatically accepting it. (`DEBUG` part is printed when CA authentication fails)
```
++ timeout --foreground 120 /home/runner/go/bin/lxc remote add pki-lxd 127.0.0.1:47377 --verbose
time="2024-10-11T11:43:10Z" level=info msg="http: TLS handshake error from 127.0.0.1:43634: EOF"

(DEBUG) Error: Get "https://127.0.0.1:47377/1.0": tls: failed to verify certificate: x509: certificate signed by unknown authority

Certificate fingerprint: 4cb5d7cebdc68022bcc864878d895637d9c1d8c55a3e1a69cc828754229b50b3
ok (y/n/[fingerprint])? 
```

I've tested this before the changes in this PR and the result is exactly the same indicating this is not a regression. 

If you agree, can this be addressed in a separate PR?

markylaing: Did you add the CA certificate to the clients configuration directory as `client.ca`? If so and it still prompts to accept the certificate then that's a bug. Happy for it to be addressed separately.

I vaguely remember that this is difficult to test - something to do with hostnames not matching perhaps.... In any case, we can address in a separate issue and if we can't test it we should add a comment to explain why for next time.

lxc/remote.go
```diff
@@ -449,22 +449,32 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
...
+			for {
+				line, err := shared.ReadStdin()
+				if err != nil {
+					return err
+				}
 
-			if string(line) != digest {
+				// Continue with adding the remote if digest matches, or the user
+				// confirmed a fingerprint.
+				if string(line) == digest || strings.ToLower(string(line[0])) == i18n.G("y") {
```
markylaing, line 518: I wonder if we should still error if the input looks like a fingerprint but it doesn't match so that it doesn't block forever in a scripted scenario. E.g:
```
echo "${expectedFingerprint}" | lxc remote add <token>
```

MusicDin: Thanks for suggestion. This is fixed now - if provided fingerprint length matches the remote certificate fingerprint length, but they are not equal, an error is immediately returned. 
```
$ lxc remote add 127.0.0.1
Certificate fingerprint: 99303272e5e88ed5ddc69d54b12761ab45b8afcece5e1262bed46e7c3e145203
ok (y/n/[fingerprint])? test
Please type 'y', 'n' or the fingerprint:test
Please type 'y', 'n' or the fingerprint:99303272e5e88ed5ddc69d54b12761ab45b8afcece5e1262bed46e7c3e14520    # 1 character shorter
Please type 'y', 'n' or the fingerprint:99303272e5e88ed5ddc69d54b12761ab45b8afcece5e1262bed46e7c3e145203x  # 1 character longer
Please type 'y', 'n' or the fingerprint:99303272e5e88ed5ddc69d54b12761ab45b8afcece5e1262bed46e7c3e14520x   # Lengths are equal, but the token is mismatched.
Error: Provided fingerprint does not match the certificate fingerprint
```

simondeziel: In the above, when re-asking for the correct input, could you please add a space after the `:` in the prompt? Like that:
```
Please type 'y', 'n' or the fingerprint: test
```

MusicDin:
@markylaing  Forgot to ping for another review :)

lxc/remote.go
```diff
@@ -267,9 +267,21 @@ func (c *cmdRemoteAdd) addRemoteFromToken(addr string, server string, token stri
 		req.Password = token
 	}
 
+	// Add client certificate to trust store. Even if we are already trusted (src.Auth == "trusted"),
+	// we want to send the token to invalidate it. Therefore, we can ignore the client already trusted error.
 	err = d.CreateCertificate(req)
+	if err != nil && err.Error() != "Client is already trusted" {
```
markylaing, line 273: I don't think we should match on the error string but I'm not sure if there is a better way...

MusicDin: Yes, I was looking at the status code, but it is generic 400 BadRequest. Not sure if I can wrap the error over the connection - should try that

lxd/certificates.go
```diff
@@ -477,7 +477,28 @@ func certificatesPost(d *Daemon, r *http.Request) response.Response {
 	}
 
 	// If caller is already trusted and does not have permission to create certificates, they cannot create more certificates.
-	if trusted && !userCanCreateCertificates && req.Certificate == "" && !req.Token {
+	// If trust token is provided, also return an error because trust token can only be used to add a remote.
+	if trusted && (req.Certificate == "" && !req.Token && !userCanCreateCertificates || req.TrustToken != "") {
```
markylaing, line 481: I'm uncertain which edge cases the changes made here are catching.

Can we just have `if trusted and req.TrustToken != ""`?

Nobody, regardless of their permissions, should be sending a trust token if they are already trusted right?

MusicDin: It is `if trusted and req.TrustToken != ""` - I just left the previous condition `trusted && (req.Certificate == "" && !req.Token && !userCanCreateCertificates` there to not mess with anything else (and not duplicate the error).

So it can be read as if `trusted && req.TrustToken != ""` or `trusted && req.Certificate == "" && !req.Token && !userCanCreateCertificates`.

Should I split it?

lxc/remote.go
markylaing: No it won't be possible to do that. The client recreates an `api.StatusError` based on the status code of the request and the error message. So all the wrapping is gone.

We could potentially add the error message to the `api` package but I'm not a fan of that either tbh... Lets leave for now.

MusicDin: I can maybe put the error into shared and use from both `lxc` and `lxd` packages?

MusicDin: Maybe I can return a 409 conflict? Since conflict is returned if user fails to add a new certificate, the same could apply when using trust token, as it is only used to add client's certificate?

markylaing (approved):
LGTM thanks!

tomponline (approved):

tomponline:
Thanks!

PR #14153, opened 2024-09-24
lxd/storage: Check for direct I/O when converting image format

MusicDin:
Check for direct I/O when converting image format using `qemu-img` to bypass OS cache if possible.


MusicDin:
3 tests are failing due to `resize2fs` issue. Since conversion is tested only in LXD CI, no test should fail in this repository.

tomponline (approved):
ta

PR #14156, opened 2024-09-25
Remove fanotify warning on ubuntu pro fsmonitor initialisation

markylaing:
This change allows the caller to specify a driver when loading an fsmonitor, or they can specify "any" to try to load fanotify but fallback to inotify if it fails.

### DIFF
--- a/lxd/fsmonitor/fsmonitor_interface.go
+++ b/lxd/fsmonitor/fsmonitor_interface.go
@@ -1,5 +1,17 @@
 package fsmonitor
 
+const (
+	// DriverNameFANotify is the name of the FANotify driver.
+	//
+	// FANotify should be preferred over INotify because it is more performant and does not need to recursively watch
+	// subdirectories. However, it is not possible to use fanotify if the specified path is not a mountpoint because we
+	// need to use the unix.FAN_MARK_FILESYSTEM flag for this functionality.
+	DriverNameFANotify = "fanotify"
+
+	// DriverNameINotify is the name of the inotify driver.
+	DriverNameINotify = "inotify"
+)
+
 // FSMonitor represents a filesystem monitor.
 type FSMonitor interface {
 	DriverName() string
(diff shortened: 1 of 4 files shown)


tomponline (commented):
Is there a reason we cant handle this logic inside fsmonitor itself by detecting if the path being monitored is a mount path (using `filesystem.IsMountPoint`) and only trying inotify if its not a mount point, rather than expanding the API and putting the burden on the caller to work it out?

markylaing:
> Is there a reason we cant handle this logic inside fsmonitor itself by detecting if the path being monitored is a mount path (using `filesystem.IsMountPoint`) and only trying inotify if its not a mount point, rather than expanding the API and putting the burden on the caller to work it out?

Nope. I just didn't think of that... Will change.

simondeziel (approved):

lxd/fsmonitor/fsmonitor_interface.go
```diff
@@ -1,5 +1,17 @@
 package fsmonitor
 
+const (
+	// DriverNameFANotify is the name of the FANotify driver.
+	//
+	// FANotify should be preferred over INotify because it is more performant and does not need to recursively watch
+	// subdirectories. However, it is not possible to use fanotify if the specified path is not a mountpoint because we
+	// need to use the unix.FAN_MARK_FILESYSTEM flag for this functionality.
+	DriverNameFANotify = "fanotify"
```
tomponline, line 9: Do these constants need exporting?

tomponline (approved):

lxd/fsmonitor/fsmonitor_interface.go
markylaing: I was also thinking this, but I figured that they're constants and exporting the names gives the caller a small amount of information about the underlying implementation. I realise the point of an interface is so that the caller doesn't have to care, but it's nice documentation wise to get a little overview at the top level.

PR #14157, opened 2024-09-25
lxd/storage: Fix check for direct I/O write support

simondeziel:


lxd/storage/backend_lxd.go
```diff
@@ -1858,7 +1858,7 @@ func (b *lxdBackend) imageConversionFiller(imgPath string, imgFormat string, op
 			_ = from.Close()
 		}
 
-		to, err := os.OpenFile(diskPath, unix.O_DIRECT|unix.O_RDONLY, 0)
+		to, err := os.OpenFile(diskPath, unix.O_DIRECT|unix.O_WRONLY, 0)
```
tomponline, line 1861: do we need this, we use O_RDONLY in ImageUnpack.

The aim is just to figure out if the underlying storage supports direct I/O, not to actually use the file handle.

simondeziel: That's a good question I don't know the answer for. Can direct I/O support be different for writing and reading?

If not, then why check both `from` and `to`?

simondeziel: NVM re `from` and `to`, they can be on different FS :)

simondeziel: @mihalicyn do you know if `O_DIRECT` support can vary between read and write accesses? I'm well outside of my depth here but after reading the commit description of https://github.com/openzfs/zfs/pull/10018/commits, I believe it would be prudent to test for both type of operations.

mihalicyn (approved):

lxd/storage/backend_lxd.go
tomponline: OK thanks

MusicDin: Should this be similarly adjusted on `ImageUnpack`?

tomponline: yeah if its appropriate

Issue #14047, opened 2024-09-05
Aliases from `lxc image copy --copy-aliases` are exclusive to one image type


DanielDewberry:
# Required information

Distribution: `Kubuntu`
Distribution version: `24.04.1`

The output of `snap list --all lxd core20 core22 core24 snapd`:

```
Name    Version         Rev    Tracking       Publisher    Notes
core20  20240416        2318   latest/stable  canonical**  base
core22  20240731        1564   latest/stable  canonical**  base,disabled
core22  20240809        1586   latest/stable  canonical**  base
core24  20240528        423    latest/stable  canonical**  base,disabled
core24  20240710        490    latest/stable  canonical**  base
lxd     5.21.2-22f93f4  29948  5.21/stable    canonical**  disabled,held
lxd     5.21.2-2f4ba6b  30131  5.21/stable    canonical**  held
snapd   2.62            21465  latest/stable  canonical**  snapd,disabled
snapd   2.63            21759  latest/stable  canonical**  snapd
```

The output of `lxc info`:
 
```
config:
  core.https_address: :8443
api_extensions:
- storage_zfs_remove_snapshots
- container_host_shutdown_timeout
- container_stop_priority
- container_syscall_filtering
- auth_pki
- container_last_used_at
- etag
- patch
- usb_devices
- https_allowed_credentials
- image_compression_algorithm
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
- unix_device_rename
- storage_lvm_use_thinpool
- storage_rsync_bwlimit
- network_vxlan_interface
- storage_btrfs_mount_options
- entity_description
- image_force_refresh
- storage_lvm_lv_resizing
- id_map_base
- file_symlinks
- container_push_target
- network_vlan_physical
- storage_images_delete
- container_edit_metadata
- container_snapshot_stateful_migration
- storage_driver_ceph
- storage_ceph_user_name
- resource_limits
- storage_volatile_initial_source
[...]

PR #14162, opened 2024-09-25
Add support for image aliases of the same name for containers and virtual machines

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/14047.

This PR includes several changes to image alias functionality, including:

- `images_aliases` schema update which adds an image type column;
- improvements to the `/images/aliases` API including added support for image types;
- instance type checking for improved error handling; and
- client support for aliases with image types.

### DIFF
--- a/client/interfaces.go
+++ b/client/interfaces.go
@@ -61,7 +61,7 @@ type ImageServer interface {
 	GetImageAliases() (aliases []api.ImageAliasesEntry, err error)
 	GetImageAliasNames() (names []string, err error)
 
-	GetImageAlias(name string) (alias *api.ImageAliasesEntry, ETag string, err error)
+	GetImageAlias(name string, instanceType api.InstanceType) (alias *api.ImageAliasesEntry, ETag string, err error)
 	GetImageAliasType(imageType string, name string) (alias *api.ImageAliasesEntry, ETag string, err error)
 	GetImageAliasArchitectures(imageType string, name string) (entries map[string]*api.ImageAliasesEntry, err error)
 
--- a/client/simplestreams_images.go
+++ b/client/simplestreams_images.go
@@ -253,7 +253,7 @@ func (r *ProtocolSimpleStreams) GetImageAliasNames() ([]string, error) {
 }
 
 // GetImageAlias returns an existing alias as an ImageAliasesEntry struct.
-func (r *ProtocolSimpleStreams) GetImageAlias(name string) (*api.ImageAliasesEntry, string, error) {
+func (r *ProtocolSimpleStreams) GetImageAlias(name string, _ api.InstanceType) (*api.ImageAliasesEntry, string, error) {
 	alias, err := r.ssClient.GetAlias("container", name)
 	if err != nil {
 		alias, err = r.ssClient.GetAlias("virtual-machine", name)
@@ -268,7 +268,7 @@ func (r *ProtocolSimpleStreams) GetImageAlias(name string) (*api.ImageAliasesEnt
 // GetImageAliasType returns an existing alias as an ImageAliasesEntry struct.
 func (r *ProtocolSimpleStreams) GetImageAliasType(imageType string, name string) (*api.ImageAliasesEntry, string, error) {
 	if imageType == "" {
-		return r.GetImageAlias(name)
+		return r.GetImageAlias(name, "")
 	}
 
 	alias, err := r.ssClient.GetAlias(imageType, name)
--- a/doc/rest-api.yaml
+++ b/doc/rest-api.yaml
@@ -973,6 +973,11 @@ definitions:
                 example: ubuntu-24.04
                 type: string
                 x-go-name: Name
+            type:
+                description: Alias type (container or virtual-machine)
+                example: container
+                type: string
+                x-go-name: Type
         type: object
         x-go-package: github.com/canonical/lxd/shared/api
     ImageAliasesEntry:
@@ -1008,6 +1013,11 @@ definitions:
                 example: ubuntu-24.04
                 type: string
                 x-go-name: Name
+            type:
+                description: Alias type (container or virtual-machine)
+                example: container
+                type: string
+                x-go-name: Type
         type: object
         x-go-package: github.com/canonical/lxd/shared/api
     ImageAliasesEntryPut:
@@ -1023,6 +1033,11 @@ definitions:
                 example: 06b86454720d36b20f94e31c6812e05ec51c1b568cf3a8abd273769d213394bb
                 type: string
                 x-go-name: Target
+            type:
+                description: Alias type (container or virtual-machine)
+                example: container
+                type: string
+                x-go-name: Type
         type: object
         x-go-package: github.com/canonical/lxd/shared/api
     ImageAliasesPost:
--- a/lxc/utils.go
+++ b/lxc/utils.go
@@ -141,7 +141,7 @@ func parseDeviceOverrides(deviceOverrideArgs []string) (map[string]map[string]st
 
 // IsAliasesSubset returns true if the first array is completely contained in the second array.
 func IsAliasesSubset(a1 []api.ImageAlias, a2 []api.ImageAlias) bool {
-	set := make(map[string]interface{})
+	set := make(map[string]any)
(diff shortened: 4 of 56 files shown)


client/interfaces.go
```diff
@@ -61,7 +61,7 @@ type ImageServer interface {
 	GetImageAliases() (aliases []api.ImageAliasesEntry, err error)
 	GetImageAliasNames() (names []string, err error)
 
-	GetImageAlias(name string) (alias *api.ImageAliasesEntry, ETag string, err error)
```
hamistao, line 64: The commit message only mentions `lxc` but this commit changes many modules. What do you think of breaking this commit in parts for `client`, `lxc` and `lxd-benchmark`?

doc/rest-api.yaml
```diff
@@ -1008,6 +1013,11 @@ definitions:
                 example: ubuntu-24.04
                 type: string
                 x-go-name: Name
+            type:
+                description: Alias type (container or virtual-machine)
+                example: container
+                type: string
+                x-go-name: Type
```
hamistao, line 1020: Same comment regarding using `ImageType` instead of `Type`

client/simplestreams_images.go
```diff
@@ -268,7 +268,7 @@ func (r *ProtocolSimpleStreams) GetImageAlias(name string) (*api.ImageAliasesEnt
 // GetImageAliasType returns an existing alias as an ImageAliasesEntry struct.
 func (r *ProtocolSimpleStreams) GetImageAliasType(imageType string, name string) (*api.ImageAliasesEntry, string, error) {
 	if imageType == "" {
-		return r.GetImageAlias(name)
+		return r.GetImageAlias(name, "")
```
hamistao, line 271: Would it make sense to default to `container` here as well?

lxc/utils.go
```diff
@@ -141,7 +141,7 @@ func parseDeviceOverrides(deviceOverrideArgs []string) (map[string]map[string]st
 
 // IsAliasesSubset returns true if the first array is completely contained in the second array.
 func IsAliasesSubset(a1 []api.ImageAlias, a2 []api.ImageAlias) bool {
-	set := make(map[string]interface{})
+	set := make(map[string]any)
```
hamistao, line 144: Nice to keep our code up to date with the latest practices :D

client/interfaces.go
hamistao: I had some problems doing these rebases before. So if you do this and don't know how, I suggest doing an interactive rebase, set the previous commit as edit, cherry pick this commit with -`-no-commit` and commit the changes separately. Just say the word if you want a hand!

lxd/images.go
```diff
@@ -3559,20 +3559,34 @@ func imageAliasesGet(d *Daemon, r *http.Request) response.Response {
...
-				responseStr = append(responseStr, api.NewURL().Path(version.APIVersion, "images", "aliases", name).String())
-			} else {
-				_, alias, err := tx.GetImageAlias(ctx, projectName, name, "", true)
-				if err != nil {
+			for _, imageType := range imageTypes {
+				if visited[name][imageType] {
+					continue
+				}
+
+				if !userHasPermission(entity.ImageAliasURL(projectName, name)) {
```
hamistao, line 3574: @markylaing Now we can have two aliases with the same name (one for a container image and one for a VM image), but both would have the same URL and this I understand that the same permissions would be granted for both.
I think maybe this is best kept as is and we just grant the same permissions for aliases with the same name, but I would like to double check and ask how do you think we should handle this.

lxd/db/db_internal_test.go
```diff
@@ -242,7 +242,7 @@ func (s *dbTestSuite) Test_ImageExists_false() {
 
 func (s *dbTestSuite) Test_GetImageAlias_alias_exists() {
 	_ = s.db.Transaction(context.Background(), func(ctx context.Context, tx *ClusterTx) error {
-		_, alias, err := tx.GetImageAlias(ctx, "default", "somealias", true)
+		_, alias, err := tx.GetImageAlias(ctx, "default", "somealias", "container", true)
```
hamistao, line 245: To ensure we are getting both right, could you please introduce a similar test with a VM image alias?

lxd/images.go
```diff
@@ -3548,7 +3548,7 @@ func imageAliasesGet(d *Daemon, r *http.Request) response.Response {
 	var responseStr []string
 	var responseMap []api.ImageAliasesEntry
 	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
-		names, err := tx.GetImageAliases(ctx, projectName)
```
hamistao, line 3549: After this, I believe `GetImageAliases` is left unused, so you can just update `GetImageAliases` to also return the aliases' image types instead of creating a new function `GetImageAliasesTypes`

lxd/db/images.go
```diff
@@ -625,6 +625,47 @@ SELECT images_aliases.name
 	return names, nil
 }
 
+// GetImageAliasesTypes returns the names and image types of the aliases of all images.
```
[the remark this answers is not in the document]
hamistao: This can be disregarded if we are going to change `GetImageAliases` instead of introducing this function.

hamistao (changes requested):
Also, since this PR introduces an update to the API via the addition of the `Type` field, an API extension should also be added here. We use API extensions to keep track of API updates individually. [Here](https://github.com/canonical/lxd/commit/5ecafb1b81602cfecf09cef0fe4c860a759f2ee8) is an example commit so make this concept clearer.

lxd/images.go
```diff
@@ -3396,14 +3396,16 @@ func imageAliasesPost(d *Daemon, r *http.Request) response.Response {
...
+		// This is just to see if the alias already exists.
+		_, entry, err := tx.GetImageAlias(ctx, projectName, req.Name, true)
 		if !response.IsNotFoundError(err) {
 			if err != nil {
 				return err
 			}
 
-			return api.StatusErrorf(http.StatusConflict, "Alias %q already exists", req.Name)
+			if req.Type == entry.Type {
+				return api.StatusErrorf(http.StatusConflict, "Alias %q already exists for %q", req.Name, req.Type)
```
tomponline: Suggest

> Alias %q already exists for image type %q

lxd/images.go
tomponline: we cant have 2 different resources with the same URL - this might block the change indeed

tomponline: so whilst i am not too concerned about the authz part of this, we need to check if the image alias itself is addressable over the API, and if we have to introduce a breaking API change for this then we might have to can the idea.

client/interfaces.go
tomponline: yes agree, this should be several commits, at least 1 per package to aid with backporting.

hamistao:
@kadinsayani Sorry, one more coment I just thought of. If the tests you already did on the other PRs don't include tests for having one alias of each type, it could be a good addition here.

tomponline:
I suspect we will need to remove the old API endpoint, add new API endpoints to include a /type/ segment and then update the LXD Go SDK to detect (by way of API extension) whether or not the server supports the old way or the new way.

This does then raise the spectre of needing to change the Go SDK API too, but this remains to be seen if we can do it tastefully.

Closing for now.

kadinsayani:
Thanks for all your comments and feedback @hamistao and @tomponline. Even though we're closing this PR for now, I've noted all your suggestions 😄.

shared/api/image.go
```diff
@@ -239,6 +239,10 @@ type ImageAlias struct {
 	// Description of the alias
 	// Example: Our preferred Ubuntu image
 	Description string `json:"description" yaml:"description"`
+
+	// Alias type (container or virtual-machine)
+	// Example: container
+	Type string `json:"type" yaml:"type"`
```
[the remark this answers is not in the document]
kadinsayani: I chose to use `Type` to stay consistent with the already implemented `ImageAliasesEntry`. I think `ImageType` is more descriptive - we can probably refactor `ImageAliasesEntry` when we update the other alias types 😄.

PR #14170, opened 2024-09-26
Use `strings.Cut()` where appropriate.

simondeziel:
`strings.Cut()` is a wrapper around `stringslite.Cut()`: https://cs.opensource.google/go/go/+/refs/tags/go1.23.1:src/internal/stringslite/strings.go;l=108
`strings.SplitN()` is a wrapper around `strings.genSplit()`: https://cs.opensource.google/go/go/+/refs/tags/go1.23.1:src/strings/strings.go;l=236

The former (`strings.Cut()`) seems to be slightly more efficient as it doesn't do any slice allocation.


kadinsayani:
Nice!

tomponline (approved):

Issue #14133, opened 2024-09-19
VM CPU auto pinning causes slowdowns and stealtime


pkramme:
# Required information

 * Distribution: Ubuntu
 * Distribution version: 20.04.5
 * The output of "snap list --all lxd core20 core22 core24 snapd":
   ```
   Name    Version      Rev    Tracking       Publisher   Notes
   core22  20240809     1586   latest/stable  canonical✓  base,disabled
   core22  20240823     1612   latest/stable  canonical✓  base
   lxd     6.1-efad198  29943  latest/stable  canonical✓  disabled
   lxd     6.1-78a3d8f  30130  latest/stable  canonical✓  -
   snapd   2.62         21465  latest/stable  canonical✓  snapd,disabled
   snapd   2.63         21759  latest/stable  canonical✓  snapd
   ```
 * The output of "lxc info" or if that fails:
   ```
    config:
    api_extensions:
    - storage_zfs_remove_snapshots
    - container_host_shutdown_timeout
    - container_stop_priority
    - container_syscall_filtering
    - auth_pki
    - container_last_used_at
    - etag
    - patch
    - usb_devices
    - https_allowed_credentials
    - image_compression_algorithm
    - directory_manipulation
    - container_cpu_time
    - storage_zfs_use_refquota
    - storage_lvm_mount_options
    - network
    - profile_usedby
    - container_push
    - container_exec_recording
    - certificate_update
    - container_exec_signal_handling
    - gpu_devices
    - container_image_properties
    - migration_progress
    - id_map
    - network_firewall_filtering
    - network_routes
    - storage
    - file_delete
    - file_append
    - network_dhcp_expiry
    - storage_lvm_vg_rename
    - storage_lvm_thinpool_rename
    - network_vlan
    - image_create_aliases
    - container_stateless_copy
    - container_only_migration
    - storage_zfs_clone_copy
    - unix_device_rename
    - storage_lvm_use_thinpool
    - storage_rsync_bwlimit
    - network_vxlan_interface
    - storage_btrfs_mount_options
    - entity_description
    - image_force_refresh
    - storage_lvm_lv_resizing
    - id_map_base
    - file_symlinks
    - container_push_target
    - network_vlan_physical
    - storage_images_delete
    - container_edit_metadata
    - container_snapshot_stateful_migration
    - storage_driver_ceph
    - storage_ceph_user_name
    - resource_limits
    - storage_volatile_initial_source
    - storage_ceph_force_osd_reuse
[...]

PR #13670, opened 2024-06-26
Disable VM CPU auto pinning (stable-5.21)

tomponline:
Restores the previously reverted `SetAffinity` implementation for qemu driver so we can easily cherry-pick fixes from main in the future for this.

And instead adds logic to prevent auto pinning for VMs in the `deviceEventListener` which also stops the logging of the `level=debug msg="Scheduler: virtual-machine v1 changed: re-balancing"` message.

We can take this out later once needed.

PR #14171, opened 2024-09-26
Add `limits.cpu.pin_strategy` setting to disable VM CPU auto pinning by default

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/14133.

This pull request introduces a configuration option for virtual machines, `limits.cpu.pin_strategy`. This option allows users to specify the strategy for CPU auto pinning. The possible values are `none` (disables CPU auto pinning - also the new default) and `auto` (enables CPU auto pinning).

For reference when back-porting: https://github.com/canonical/lxd/pull/13670/commits/56bd496649a8c044e41272acae5ef693d3b48712

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2516,3 +2516,7 @@ Adds support for using a bridge network with a specified VLAN ID as an OVN uplin
 
 Adds `logical_cpus` field to `GET /1.0/cluster/members/{name}/state` which
 contains the total available logical CPUs available when LXD started.
+
+## `vm_limits_cpu_pin_strategy`
+
+Adds a new {config:option}`instance-resource-limits:limits.cpu.pin_strategy` configuration option for virtual machines. This option controls the CPU pinning strategy. When set to `none`, CPU auto pinning is disabled. When set to `auto`, CPU auto pinning is enabled.
--- a/doc/reference/instance_options.md
+++ b/doc/reference/instance_options.md
@@ -98,6 +98,9 @@ You have different options to limit CPU usage:
 - Set {config:option}`instance-resource-limits:limits.cpu.allowance` to restrict the load an instance can put on the available CPUs.
   This option is available only for containers.
   See {ref}`instance-options-limits-cpu-container` for how to set this option.
+- Set {config:option}`instance-resource-limits:limits.cpu.pin_strategy` to specify the strategy for virtual-machine CPU auto pinning.
+  This option is available only for virtual machines.
+  See {ref}`instance-options-limits-cpu-vm` for how to set this option.
 
 It is possible to set both options at the same time to restrict both which CPUs are visible to the instance and the allowed usage of those instances.
 However, if you use {config:option}`instance-resource-limits:limits.cpu.allowance` with a time limit, you should avoid using {config:option}`instance-resource-limits:limits.cpu` in addition, because that puts a lot of constraints on the scheduler and might lead to less efficient allocations.
@@ -116,6 +119,7 @@ You can specify either which CPUs or how many CPUs are visible and available to 
 - If you specify a number (for example, `4`) of CPUs, LXD will do dynamic load-balancing of all instances that aren't pinned to specific CPUs, trying to spread the load on the machine.
   Instances are re-balanced every time an instance starts or stops, as well as whenever a CPU is added to the system.
 
+(instance-options-limits-cpu-vm)=
 ##### CPU limits for virtual machines
 
 ```{note}
@@ -127,10 +131,10 @@ Depending on the guest operating system, you might need to either restart the in
 LXD virtual machines default to having just one vCPU allocated, which shows up as matching the host CPU vendor and type, but has a single core and no threads.
 
 When {config:option}`instance-resource-limits:limits.cpu` is set to a single integer, LXD allocates multiple vCPUs and exposes them to the guest as full cores.
-Those vCPUs are not pinned to specific physical cores on the host.
+Unless {config:option}`instance-resource-limits:limits.cpu.pin_strategy` is set to `auto`, those vCPUs are not pinned to specific cores on the host.
 The number of vCPUs can be updated while the VM is running.
 
-When {config:option}`instance-resource-limits:limits.cpu` is set to a range or comma-separated list of CPU IDs (as provided by [`lxc info --resources`](lxc_info.md)), the vCPUs are pinned to those physical cores.
+When {config:option}`instance-resource-limits:limits.cpu` is set to a range or comma-separated list of CPU IDs (as provided by [`lxc info --resources`](lxc_info.md)), the vCPUs are pinned to those cores.
 In this scenario, LXD checks whether the CPU configuration lines up with a realistic hardware topology and if it does, it replicates that topology in the guest.
 When doing CPU pinning, it is not possible to change the configuration while the VM is running.
 
--- a/lxd/cgroup/cgroup_cpu.go
+++ b/lxd/cgroup/cgroup_cpu.go
@@ -20,9 +20,7 @@ func TaskSchedulerTrigger(srcType string, srcName string, srcStatus string) {
 }
 
 // ParseCPU parses CPU allowances.
-func ParseCPU(cpuAllowance string, cpuPriority string) (int64, int64, int64, error) {
-	var err error
-
+func ParseCPU(cpuAllowance string, cpuPriority string) (cpuShares int64, cpuCfsQuota int64, cpuCfsPeriod int64, err error) {
 	// Max shares depending on backend.
 	maxShares := int64(1024)
 	if cgControllers["cpu"] == V2 {
@@ -30,7 +28,7 @@ func ParseCPU(cpuAllowance string, cpuPriority string) (int64, int64, int64, err
 	}
 
 	// Parse priority
-	cpuShares := int64(0)
+	cpuShares = 0
 	cpuPriorityInt := 10
 	if cpuPriority != "" {
 		cpuPriorityInt, err = strconv.Atoi(cpuPriority)
@@ -41,8 +39,8 @@ func ParseCPU(cpuAllowance string, cpuPriority string) (int64, int64, int64, err
 	cpuShares -= int64(10 - cpuPriorityInt)
 
 	// Parse allowance
-	cpuCfsQuota := int64(-1)
-	cpuCfsPeriod := int64(100000)
+	cpuCfsQuota = -1
+	cpuCfsPeriod = 100000
 	if cgControllers["cpu"] == V2 {
 		cpuCfsPeriod = -1
 	}
--- a/lxd/cgroup/cgroup_cpu.go
+++ b/lxd/cgroup/cgroup_cpu.go
@@ -4,16 +4,18 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+
+	"github.com/canonical/lxd/lxd/instance/instancetype"
 )
 
 // DeviceSchedRebalance channel for scheduling a CPU rebalance.
 var DeviceSchedRebalance = make(chan []string, 2)
 
 // TaskSchedulerTrigger triggers a CPU rebalance.
-func TaskSchedulerTrigger(srcType string, srcName string, srcStatus string) {
+func TaskSchedulerTrigger(srcType instancetype.Type, srcName string, srcStatus string) {
 	// Spawn a go routine which then triggers the scheduler
 	select {
-	case DeviceSchedRebalance <- []string{srcType, srcName, srcStatus}:
+	case DeviceSchedRebalance <- []string{srcType.String(), srcName, srcStatus}:
 	default:
 		// Channel is full, drop the event
 	}
--- a/lxd/devices.go
+++ b/lxd/devices.go
@@ -21,6 +21,7 @@ import (
 	"github.com/canonical/lxd/lxd/state"
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/logger"
+	"github.com/canonical/lxd/shared/validate"
 )
 
 type deviceTaskCPU struct {
@@ -482,6 +483,14 @@ func deviceTaskBalance(s *state.State) {
 			} else {
 				cpulimit = effectiveCpus
 			}
+		}
+
+		// Determine CPU pinning strategy and static pinning settings.
+		// When pinning strategy does not equal auto (none or empty), don't auto pin CPUs.
+		cpuPinStrategy := conf["limits.cpu.pin_strategy"]
+		err = validate.IsStaticCPUPinning(cpulimit)
+		if err != nil && c.Type() == instancetype.VM && cpuPinStrategy != "auto" {
+			continue
 		}
 
 		// Check that the instance is running.
--- a/lxd/instance/instance_utils.go
+++ b/lxd/instance/instance_utils.go
@@ -36,6 +36,7 @@ import (
 	"github.com/canonical/lxd/shared/logger"
 	"github.com/canonical/lxd/shared/osarch"
 	"github.com/canonical/lxd/shared/revert"
+	"github.com/canonical/lxd/shared/validate"
 	"github.com/canonical/lxd/shared/version"
 )
 
@@ -109,6 +110,14 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
 
 	if shared.IsTrue(config["security.privileged"]) && shared.IsTrue(config["nvidia.runtime"]) {
 		return fmt.Errorf("nvidia.runtime is incompatible with privileged containers")
+	}
+
+	// Validate pinning strategy when limits.cpu specifies static pinning.
+	cpuPinStrategy := config["limits.cpu.pin_strategy"]
+	cpuLimit := config["limits.cpu"]
+	err = validate.IsStaticCPUPinning(cpuLimit)
+	if err == nil && !expanded && cpuPinStrategy == "auto" {
+		return fmt.Errorf(`CPU pinning specified, but pinning strategy is set to "auto"`)
 	}
 
 	return nil
--- a/shared/validate/validate.go
+++ b/shared/validate/validate.go
@@ -884,3 +884,12 @@ func IsMultipleOfUnit(unit string) func(value string) error {
 		return nil
 	}
 }
+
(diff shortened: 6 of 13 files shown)


doc/reference/instance_options.md
```diff
@@ -98,6 +98,9 @@ You have different options to limit CPU usage:
 - Set {config:option}`instance-resource-limits:limits.cpu.allowance` to restrict the load an instance can put on the available CPUs.
   This option is available only for containers.
   See {ref}`instance-options-limits-cpu-container` for how to set this option.
+- Set {config:option}`instance-resource-limits:limits.cpu.pin_strategy` to specify the strategy for virtual-machine CPU auto pinning.
+  This option is available only for containers.
```
simondeziel, line 102: This `containers` bit seems out of place.

doc/reference/instance_options.md
```diff
@@ -127,10 +131,10 @@ Depending on the guest operating system, you might need to either restart the in
 LXD virtual machines default to having just one vCPU allocated, which shows up as matching the host CPU vendor and type, but has a single core and no threads.
 
 When {config:option}`instance-resource-limits:limits.cpu` is set to a single integer, LXD allocates multiple vCPUs and exposes them to the guest as full cores.
-Those vCPUs are not pinned to specific physical cores on the host.
+Unless {config:option}`instance-resource-limits:limits.cpu.pin_strategy` is set to `auto`, those vCPUs are not pinned to specific physical cores on the host.
```
simondeziel, line 134: I wouldn't drop the `physical` mention as there are also virtual cores (Hyperthread).

doc/reference/instance_options.md
```diff
@@ -127,10 +131,10 @@ Depending on the guest operating system, you might need to either restart the in
...
 The number of vCPUs can be updated while the VM is running.
 
-When {config:option}`instance-resource-limits:limits.cpu` is set to a range or comma-separated list of CPU IDs (as provided by [`lxc info --resources`](lxc_info.md)), the vCPUs are pinned to those physical cores.
+When {config:option}`instance-resource-limits:limits.cpu` is set to a range or comma-separated list of CPU IDs (as provided by [`lxc info --resources`](lxc_info.md)) and {config:option}`instance-resource-limits:limits.cpu.pin_strategy` is set to `auto`, the vCPUs are pinned to those physical cores.
```
simondeziel, line 137: Same note about `physical`.

doc/reference/instance_options.md
kadinsayani: Indeed, it should be virtual machines.

doc/reference/instance_options.md
kadinsayani: `those vCPUs are not pinned to specific cores on the host`?

kadinsayani:
If we modify the current tests in `lxd-ci` to launch a VM with the `limits.cpu.pin_strategy` set, there will be a regression since earlier LXD versions will return `Error: Failed instance creation: Failed creating instance record: Unknown configuration key: limits.cpu.pin_strategy`, causing the `cpu-vm` tests to fail. 

Currently, the checks in `lxd-ci` for cpu pinning only run if the feature is present:

```
if journalctl --quiet --no-hostname --no-pager --boot=0 --unit=snap.lxd.daemon.service | grep "Scheduler: virtual-machine"; then
...
```

However, due to my implementation of checking the config setting in `deviceTaskBalance`, the log message `Scheduler: virtual-machine` is printed regardless of whether auto cpu pinning is enabled or not. Hence, I have added a new log message which we can use to check if the feature is present.

```
if c.Type() == instancetype.VM {
	if conf["limits.cpu.pin_strategy"] == "none" {
		continue
	} else if conf["limits.cpu.pin_strategy"] == "auto" {
		logger.Debugf("CPU auto pinning enabled for %q", c.Name())
	}
}
```
 
Therefore, I think the only change required (for now) to ensure we don't introduce a regression to `lxd-ci` is to change the check to look for the new log message.

Please let me know if you have any suggestions 😄

tomponline:
> If we modify the current tests in `lxd-ci` to launch a VM with the `limits.cpu.pin_strategy` set, there will be a regression since earlier LXD versions will return `Error: Failed instance creation: Failed creating instance record: Unknown configuration key: limits.cpu.pin_strategy`, causing the `cpu-vm` tests to fail.

You will need to introduce a new API extension for the new config key and can test for support of that when testing to enable the functionality.

simondeziel:
@kadinsayani I couldn't find any reference to back that but I think it's best for the commit adding the API extension to be the first in the PR. This way, when @tomponline does the backporting (if any, dunno for this one) he can quickly identify that a collection of commits need to be considered at once for the backport.

kadinsayani:
> @kadinsayani I couldn't find any reference to back that but I think it's best for the commit adding the API extension to be the first in the PR. This way, when @tomponline does the backporting (if any, dunno for this one) he can quickly identify that a collection of commits need to be considered at once for the backport.

Sounds good, thanks for letting me know!

kadinsayani:
@tomponline will we be back porting the commit containing the new API extension to 6.1? Currently, the tests in `lxd-ci` are relying on LXD daemon logs for checking if the feature is present.

https://github.com/mihalicyn/lxd-ci/blob/972ef98d660c3c2efba3d8aa7408126a1b555ff9/tests/cpu-vm#L12

tomponline:
> @tomponline will we be back porting the commit containing the new API extension to 6.1? Currently, the tests in `lxd-ci` are relying on LXD daemon logs for checking if the feature is present:

yes we can because it defaults to disabled

simondeziel (dismissed):

tomponline:
@kadinsayani please can you ensure that the change to disable auto pin for containers is separate so i can choose to backport it or not.

doc/api-extensions.md
```diff
@@ -2480,3 +2480,7 @@ When set to `on`, if the host has guest attachment enabled, the guest can reques
 
 This adds entity type metadata to `GET /1.0/metadata/configuration`.
 The entity type metadata is a JSON object under the `entities` key.
+
+## `limits_cpu_pin_strategy`
```
tomponline, line 2484: Lets call this `vm_limits_cpu_ping_strategy`

lxd/instance/instance_utils.go
```diff
@@ -143,6 +143,10 @@ func validConfigKey(os *sys.OS, key string, value string, instanceType instancet
 		return fmt.Errorf("%s isn't supported on this architecture", key)
 	}
 
+	if key == "limits.cpu.pin_strategy" && value != "" && !shared.ValueInSlice(value, []string{"none", "auto"}) {
```
tomponline, line 146: We should add this to `InstanceConfigKeysVM` as its VM specific.

lxd/instance/instance_utils.go
```diff
@@ -693,6 +697,23 @@ func CreateInternal(s *state.State, args db.InstanceArgs, clearLogDir bool) (Ins
 
 	args.Config["volatile.uuid.generation"] = args.Config["volatile.uuid"]
 
+	cpuPinningSpecified := strings.ContainsAny(args.Config["limits.cpu"], ",-")
+
+	if args.Type == instancetype.VM {
```
tomponline, line 702: We dont want to populate the defaults into the instance DB record, this isn't how LXD's config defaults work.

Instead you should consult the instance's expanded config where you would use the `limits.cpu` value and consider then whether to use it or not based on the value of `limits.cpu.pin_strategy`.

lxd/devices.go
```diff
@@ -644,6 +645,11 @@ func deviceEventListener(stateFunc func() *state.State) {
 				continue
 			}
 
+			// Only auto pin VM CPUs when pin strategy is set to "auto"
```
tomponline, line 648: Why didnt you add this logic into ` TaskSchedulerTrigger` to avoid having to put another value into the event?

lxd/cgroup/cgroup_cpu.go
```diff
@@ -10,10 +10,10 @@ import (
 var DeviceSchedRebalance = make(chan []string, 2)
 
 // TaskSchedulerTrigger triggers a CPU rebalance.
-func TaskSchedulerTrigger(srcType string, srcName string, srcStatus string) {
+func TaskSchedulerTrigger(srcType string, srcName string, srcStatus string, srcPinStrategy string) {
```
tomponline, line 13: For clarity can we update the srcType field to be of type instancetype.Type, e.g.

```
func TaskSchedulerTrigger(srcType instancetype.Type, srcName string, srcStatus string) {
```

Then use `srcType.String()` as the value passed to the event channel.

tomponline: Then use `d.dbType` in the call sites.

lxd/devices.go
tomponline: Oh this logic is wrong anyway, where ever it is :)

It needs to be in deviceTaskBalance (which already has access to the instance's expanded config so it doesn't need to be passed via the event), because if `limits.cpu` is set to a static pinning then we should apply that irrespective of the value of the pin strategy setting.

tomponline: We'll also want some validation when the settings are applied directly to the instance's local config to ensure that the pin strategy isnt set to `auto` when using a non-integer `limits.cpu` setting.

tomponline (changes requested):
I'm afraid the current implementation isnt quite right.

doc/api-extensions.md
simondeziel: Just making sure a copy-n-pasto doesn't happen: `s/ping/pin/` :)

lxd/instance/instance_utils.go
```diff
@@ -693,6 +693,12 @@ func CreateInternal(s *state.State, args db.InstanceArgs, clearLogDir bool) (Ins
 
 	args.Config["volatile.uuid.generation"] = args.Config["volatile.uuid"]
 
+	// Validate pinning strategy when limits.cpu specifies static pinning
+	cpuPinningSpecified := strings.ContainsAny(args.Config["limits.cpu"], ",-")
```
tomponline, line 697: This validation shouldnt be here

lxd/devices.go
```diff
@@ -484,6 +484,12 @@ func deviceTaskBalance(s *state.State) {
 			}
 		}
 
+		cpuPinStrategy := conf["limits.cpu.pin_strategy"]
```
tomponline, line 487: this validation shouldnt be here, it should be done where the rest of instance validation is checked so that invalid configs cant be stored.

tomponline: although is this needed because of the fact we do allow this key to be set in the profile?

lxd/instance/instance_utils.go
kadinsayani: I've moved this input validation to `ValidConfig()`.

kadinsayani: Let me know if a validation function in `InstanceConfigKeysVM` would be preferred for this.

lxd/instance/instance_utils.go
```diff
@@ -111,6 +111,12 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
 		return fmt.Errorf("nvidia.runtime is incompatible with privileged containers")
 	}
 
+	// Validate pinning strategy when limits.cpu specifies static pinning
```
tomponline, line 114: Thanks, this looks like the correct place to do composite config key validation.

However this seems to allow me to set the `limits.cpu` to include a comma and set `limits.cpu.pin_strategy` to `auto` in a profile (i.e when `instanceType == instancetype.Any`).

Is this correct?

So the scenarios we want to test are:

1. `limits.cpu.pin_strategy=auto` and `limits.cpu` set to static pinning in profile: Not allowed.
1. `limits.cpu.pin_strategy=auto` in profile and `limits.cpu` set to static pinning in instance config: Allowed (with `limits.cpu` taking precedence).
1. `limits.cpu` set to static pinning in profile and `limits.cpu.pin_strategy=auto` in instance config: Allowed (with the auto pin strategy taking precedence).
1. `limits.cpu.pin_strategy=auto` and `limits.cpu` set to static pinning in instance: Not allowed.
1. `limits.cpu.pin_strategy=auto` set in container config: Not allowed

Any others ive missed or got wrong in your view?

lxd/devices.go
```diff
@@ -484,6 +484,12 @@ func deviceTaskBalance(s *state.State) {
...
+		cpuPinningSpecified := strings.ContainsAny(cpulimit, ",-")
+		if c.Type() == instancetype.VM && (cpuPinStrategy == "none" || cpuPinStrategy == "") && !cpuPinningSpecified {
```
tomponline, line 489: Would this be more concise?

```go
if c.Type() == instancetype.VM && cpuPinStrategy != "auto" && !cpuPinningSpecified {
```

tomponline (changes requested):
I'm still not sure the validation is quite thorough enough yet.

lxd/devices.go
kadinsayani: Sure thing!

lxd/instance/instance_utils.go
kadinsayani: > However this seems to allow me to set the `limits.cpu` to include a comma and set `limits.cpu.pin_strategy` to `auto` in a profile (i.e when `instanceType == instancetype.Any`).
> 
> Is this correct?

When `limits.cpu.pin_strategy` is set to `auto` in a profile, `limits.cpu` should take only take precedence in instance config - so no, this is not correct.

> 1. `limits.cpu.pin_strategy=auto` and `limits.cpu` set to static pinning in profile: Not allowed.
> 2. `limits.cpu.pin_strategy=auto` in profile and `limits.cpu` set to static pinning in instance config: Allowed (with `limits.cpu` taking precedence).
> 3. `limits.cpu` set to static pinning in profile and `limits.cpu.pin_strategy=auto` in instance config: Allowed (with the auto pin strategy taking precedence).
> 4. `limits.cpu.pin_strategy=auto` and `limits.cpu` set to static pinning in instance: Not allowed.
> 5. `limits.cpu.pin_strategy=auto` set in container config: Not allowed
> 
> Any others ive missed or got wrong in your view?

I think you've captured all cases above 👍.

lxd/devices.go
kadinsayani: I like your suggestion for conciseness, but it's less descriptive. I'll add a comment for clarity as well - best of both approaches!

lxd/instance/instance_utils.go
kadinsayani: I've covered all of the cases except #3. If `limits.cpu` is set to static pinning, we actually don't want to skip over the segment that handles fixed CPU pinning/load balancing. There is already logic to prevent rebalancing for static pinning in `fillFixedInstances`. Therefore, I don't think `limits.cpu.pin_strategy=auto` at the local level should take precedence over `limits.cpu` at the profile level. Auto pinning without a cpu limit provides 1 vCPU by default, so if `limits.cpu.pin_strategy` is set to `auto` at the instance level, how would we handle allocating a fixed range of CPUs as per the profile config?

lxd/instance/instance_utils.go
```diff
@@ -111,6 +112,14 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
 		return fmt.Errorf("nvidia.runtime is incompatible with privileged containers")
 	}
 
+	// Validate pinning strategy when limits.cpu specifies static pinning.
+	cpuPinStrategy := config["limits.cpu.pin_strategy"]
+	cpuLimit := config["limits.cpu"]
+	staticPinning, _ := validate.IsStaticCPUPinning(cpuLimit)
+	if !expanded && staticPinning && cpuPinStrategy == "auto" {
+		return fmt.Errorf("CPU pinning specified, but pinning strategy is set to 'auto'")
```
tomponline, line 120: Please can you use backticks for the fmt.Errorf here and then use double quotes to wrap "auto".

shared/validate/validate.go
```diff
@@ -884,3 +884,12 @@ func IsMultipleOfUnit(unit string) func(value string) error {
 		return nil
 	}
 }
+
+// IsStaticCPUPinning validates a static CPU pinning strategy.
+func IsStaticCPUPinning(value string) (static bool, err error) {
```
tomponline, line 893: No need to return bool and error, just error is in keeping with the format of the other functions, and a nil error can mean "is static CPU pinning"

lxd/instance/instance_utils.go
tomponline: > Therefore, I don't think `limits.cpu.pin_strategy=auto` at the local level should take precedence over `limits.cpu`

Yes that makes sense indeed, if there is a static pinning setting in the profile then that should take precedence over the local `limits.cpu.pin_strategy=auto`.

tomponline:
needs a rebase

lxd/instance/instance_utils.go
kadinsayani: Sure thing!

lxd/cgroup/cgroup_cpu.go
```diff
@@ -4,33 +4,33 @@ import (
...
+func ParseCPU(cpuAllowance string, cpuPriority string) (cpuShares int64, cpuCfsQuota int64, cpuCfsPeriod int64, err error) {
 	// Max shares depending on backend.
 	maxShares := int64(1024)
 	if cgControllers["cpu"] == V2 {
 		maxShares = 100
 	}
 
 	// Parse priority
-	cpuShares := int64(0)
+	cpuShares = int64(0)
```
simondeziel, line 33: Is the type casting `int64()` still needed?

lxd/cgroup/cgroup_cpu.go
```diff
@@ -41,8 +41,8 @@ func ParseCPU(cpuAllowance string, cpuPriority string) (int64, int64, int64, err
 	cpuShares -= int64(10 - cpuPriorityInt)
 
 	// Parse allowance
-	cpuCfsQuota := int64(-1)
-	cpuCfsPeriod := int64(100000)
+	cpuCfsQuota = int64(-1)
+	cpuCfsPeriod = int64(100000)
```
simondeziel, line 45: Are the type casting `int64()` still needed?

kadinsayani: Also not needed here! Thanks!

tomponline (approved):
thanks!

PR #14173, opened 2024-09-27
Auth: Redefine identity certificate entity types to prevent overlap

### LINKED ISSUE #13372: Potentially redundant entity types for permissions

markylaing:
This PR redefines the `identity` and `certificate` entity types such that a `certificate` is considered to be any existing certificate type (e.g. client, metrics, server) and an `identity` is any identity whose permissions are managed via group membership.

Closes #13372 

Opening as draft as I would like to perform more testing on the patch before merging.

### DIFF
--- a/lxd/db/cluster/auth_groups.go
+++ b/lxd/db/cluster/auth_groups.go
@@ -287,7 +287,7 @@ func SetAuthGroupPermissions(ctx context.Context, tx *sql.Tx, groupID int, authG
 	}
 
 	for _, permission := range authGroupPermissions {
-		_, err := tx.ExecContext(ctx, `INSERT INTO auth_groups_permissions (auth_group_id, entity_type, entity_id, entitlement) VALUES (?, ?, ?, ?);`, permission.GroupID, permission.EntityType, permission.EntityID, permission.Entitlement)
+		_, err := tx.ExecContext(ctx, `INSERT INTO auth_groups_permissions (auth_group_id, entity_type, entity_id, entitlement) VALUES (?, ?, ?, ?);`, groupID, permission.EntityType, permission.EntityID, permission.Entitlement)
 		if err != nil {
 			return fmt.Errorf("Failed to write group permissions: %w", err)
 		}
--- a/lxd/identities.go
+++ b/lxd/identities.go
@@ -99,9 +99,16 @@ func identityAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *http
 			return response.SmartError(err)
 		}
 
-		err = s.Authorizer.CheckPermission(r.Context(), entity.IdentityURL(authenticationMethod, id.Identifier), entitlement)
-		if err != nil {
-			return response.SmartError(err)
+		if identity.IsFineGrainedIdentityType(string(id.Type)) {
+			err = s.Authorizer.CheckPermission(r.Context(), entity.IdentityURL(authenticationMethod, id.Identifier), entitlement)
+			if err != nil {
+				return response.SmartError(err)
+			}
+		} else {
+			err = s.Authorizer.CheckPermission(r.Context(), entity.CertificateURL(id.Identifier), entitlement)
+			if err != nil {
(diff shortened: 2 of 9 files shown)


lxd/patches.go
```diff
@@ -1436,4 +1437,41 @@ func patchInstanceRemoveVolatileLastStateIPAddresses(_ string, d *Daemon) error
...
+			return err
+		}
+
+		args := make([]any, 0, len(incorrectIdentityPermissionIDs))
+		for _, id := range incorrectIdentityPermissionIDs {
+			args = append(args, id)
+		}
+
+		// Change these permissions to apply to certificates instead. Use "UPDATE OR REPLACE" in case of UNIQUE constraint violation.
+		stmt := fmt.Sprintf("UPDATE OR REPLACE auth_groups_permissions SET entity_type = 4 WHERE id IN %s", query.Params(len(args)))
```
tomponline, line 1466: Can you use a sub-query or join here rather than doing 2 queries?

tomponline, line 1466: Please can you define the magic numbers 24, 1, and 4 as local vars/constants in this function for future readability?

lxd/identities.go
```diff
@@ -692,11 +710,6 @@ func patchIdentity(d *Daemon, r *http.Request) response.Response {
 		return response.BadRequest(fmt.Errorf("Failed to unmarshal request body: %w", err))
 	}
 
-	authenticationMethod := mux.Vars(r)["authenticationMethod"]
```
tomponline, line 695: what happens in this case now?

lxd/db/cluster/auth_groups.go
```diff
@@ -287,7 +287,7 @@ func SetAuthGroupPermissions(ctx context.Context, tx *sql.Tx, groupID int, authG
 	}
 
 	for _, permission := range authGroupPermissions {
```
tomponline, line 289: As well as this change, should we check the Permissions against the groupID and return an error before trying to execute the query?

tomponline: Ah I see its removed in next commit.

lxd/patches.go
markylaing: Yeah I'll give it a go. Now that I have the unit test I can check it more easily :)

markylaing: Yep. Those numbers are the entity type and authentication method constants that are intentionally not exported from the `lxd/db/cluster` package because we should be using constants from `shared/api` or `shared/entity` everywhere else.

lxd/identities.go
```diff
@@ -599,10 +621,6 @@ func updateIdentity(d *Daemon, r *http.Request) response.Response {
 		return response.BadRequest(fmt.Errorf("Failed to unmarshal request body: %w", err))
 	}
 
-	if id.AuthMethod == api.AuthenticationMethodTLS && len(identityPut.Groups) > 0 {
-		return response.NotImplemented(fmt.Errorf("Adding TLS identities to groups is currently not supported"))
```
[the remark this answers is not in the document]
markylaing: This is now handled in the access handler. Now that I think about it though, this is conflating the usage of the access handler. I'll change this.

lxd/patches.go
markylaing: Rather than define the magic numbers here, I've used the `Value` method on the database type definitions for these. E.g.
```
certificateEntityTypeCode, _ := dbCluster.EntityType(entity.TypeCertificate).Value()
```

tomponline (approved):

PR #14176, opened 2024-09-27
github: tweak Trivy notifications to have the channel in the description

simondeziel:


tomponline (approved):

simondeziel:
> @simondeziel Did you try to run this and check the security tab on your fork?

I have not, actions are disabled on my fork. If tweaking the description doesn't work maybe we should just not do it as the `Snap scan - ` suffix isn't all that useful. We can already filter by tools and if the snap channel isn't usable, we can always rely on the associated branch information.

hamistao:
The idea there was to distinguish the snap alerts from the repo alerts. That said, it isn't really useful now since we do not currently have any open repo alerts.

simondeziel:
> The idea there was to distinguish the snap alerts from the repo alerts. That said, it isn't really useful now since we do not currently have any open repo alerts.

My bad, I missed the fact there was "fs" scanning done on the repo as well.

PR #14183, opened 2024-10-01
lxd: Minor fixes

tomponline:


lxd/patches.go
```diff
@@ -493,10 +493,10 @@ func patchVMRenameUUIDKey(name string, d *Daemon) error {
 					newUUIDKey: uuid,
 				}
 
-				logger.Debugf("Renaming config key %q to %q for VM %q (Project %q)", oldUUIDKey, newUUIDKey, inst.Name, inst.Project)
+				logger.Debugf("Renaming config key %q to %q for VM %q (project %q)", oldUUIDKey, newUUIDKey, inst.Name, inst.Project)
```
markylaing, line 496: Should we use `logger.Ctx` here instead?

tomponline: yep, but outside of the scope of this PR, happy to see a separate PR for it.

markylaing (approved):

PR #14186, opened 2024-10-01
godep: Update dependencies and bump min Go version to 1.22.7

tomponline:
Also add some extra protections to TICS and Trivy workflows.


MusicDin (approved):
Thanks

PR #14188, opened 2024-10-01
Add `lxc file create` subcommand (from Incus)

kadinsayani:
Cherry-picked from https://github.com/lxc/incus/pull/408. I've also added instance shell completions for `lxc file` subcommands.

Signed-off-by: Kadin Sayani <kadin.sayani@canonical.com>

### DIFF
--- a/lxc/file.go
+++ b/lxc/file.go
@@ -30,8 +30,12 @@ import (
 	"github.com/canonical/lxd/shared/units"
 )
 
-// DirMode represents the file mode for creating dirs on `lxc file pull/push`.
-const DirMode = 0755
+const (
+	// DirMode represents the file mode for creating dirs on `lxc file pull/push`.
+	DirMode = 0755
+	// FileMode represents the file mode for creating files on `lxc file create`.
+	FileMode = 0644
+)
 
 type cmdFile struct {
 	global *cmdGlobal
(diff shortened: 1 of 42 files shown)


lxc/file.go
```diff
@@ -110,6 +118,187 @@ func (c *cmdFile) command() *cobra.Command {
...
+	} else if c.flagType == "file" {
+		mode = os.FileMode(FileMode)
+	}
+
+	if c.file.flagMode != "" {
+		if len(c.file.flagMode) == 3 {
+			c.file.flagMode = "0" + c.file.flagMode
+		}
+
+		m, err := strconv.ParseInt(c.file.flagMode, 0, 32)
```
simondeziel, line 237: `os.FileMode` being a `uint32` (https://pkg.go.dev/os#FileMode => https://pkg.go.dev/io/fs#FileMode). Would it make sense to use `strconv.ParseUint` instead?

simondeziel (approved):
LGTM with a small question

lxc/file.go
kadinsayani: Good catch, I've updated both references in `file.go` and also set a base and upper bound.

simondeziel (dismissed):

tomponline:
tests are unhappy

kadinsayani:
> tests are unhappy

Should be good to go now 👍

kadinsayani:
> tests are unhappy

Should be good to go now 👍

lxc/file.go
```diff
@@ -532,7 +721,7 @@ func (c *cmdFilePush) run(cmd *cobra.Command, args []string) error {
 			c.file.flagMode = "0" + c.file.flagMode
 		}
 
-		m, err := strconv.ParseInt(c.file.flagMode, 0, 0)
+		m, err := strconv.ParseUint(c.file.flagMode, 8, 32)
```
tomponline, line 756: This should be a separate commit as not related to new command

PR #14192, opened 2024-10-02
Improve Trivy scanning

hamistao:
This renames a few steps names and, most importantly, fixes the vulnerability database caching.
The database caching is updated every run and is used for all snap scanning jobs in that run.
We had to drop the [`trivy-action`](https://github.com/aquasecurity/trivy-action) for the use of `cache-dir` to be effective and for us to be able to point to the right directory when downloading and using Trivy's vulnerability database.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -26,23 +26,30 @@ jobs:
         with:
           ref: main
 
-      - name: Create Trivy cache directory
-        run: mkdir -p /home/runner/vuln-cache
+      # We are using the deb package instead of the trivy action because we need to point to the cache directory
+      # with cache-dir, and there is currently a bug in that functionality on the Trivy GitHub action.
+      # See https://github.com/aquasecurity/trivy-action/issues/12.
+      - name: Install Trivy
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          mkdir -p /home/runner/vuln-cache
+          sudo apt-get install wget apt-transport-https gnupg
+          wget -qO - https://aquasecurity.github.io/trivy-repo/deb/public.key | gpg --dearmor | sudo tee /usr/share/keyrings/trivy.gpg > /dev/null
+          echo "deb [signed-by=/usr/share/keyrings/trivy.gpg] https://aquasecurity.github.io/trivy-repo/deb generic main" | sudo tee -a /etc/apt/sources.list.d/trivy.list
+          sudo apt-get update
+          sudo apt-get install trivy
 
-      - name: Run Trivy vulnerability scanner in repo mode
-        uses: aquasecurity/trivy-action@master
-        with:
-          scan-type: "fs"
-          format: "sarif"
-          output: "trivy-lxd-repo-scan-results.sarif"
-          cache-dir: "/home/runner/vuln-cache"
-          severity: "LOW,MEDIUM,HIGH,CRITICAL"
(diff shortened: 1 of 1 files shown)


hamistao:
@tomponline @simondeziel  This was also intended to show the commit used in each snap channel on snap scanning alerts on the section shown on the bottom of the following image
image

But specifying the right SHA on the upload step did not change this section on the UI (field highlighted below).
```
- name: Upload Trivy scan results to GitHub Security tab
  uses: github/codeql-action/upload-sarif@v3
  with:
    sarif_file: "${{ matrix.version }}-stable.sarif"
    **sha: ${{ github.sha }}**
    ref: refs/heads/${{ (matrix.version == 'latest' && 'main') || format('stable-{0}', matrix.version) }}
```

So I turned to the [GitHub REST API](https://docs.github.com/en/rest/code-scanning/code-scanning?apiVersion=2022-11-28#upload-an-analysis-as-sarif-data), as security alerts include a `commit_sha` field accessible through the API and that field was unchanged after the described approach above.
Unfortunately, POSTing alerts using the right `commit_sha`, although changing the `commit_sha` field on the alert when getting it from the API, also did not impact the view on the UI in any way, and this is why this improvement is not included in this PR.

As an additional note, PATCHing or PUTing an alert to change the value of `commit_sha` is not possible.

All that said, this is ready for a review :D

simondeziel (commented):
As you mentioned the action didn't work with `--cache-dir`, I went on locking in their bugs and found https://github.com/aquasecurity/trivy-action/issues/12 to which you commented (good).

It would be nice to have a comment in the workflow pointing to that bug so that we have a better change of dropping that workaround once the bug is fixed.

Thanks.

hamistao:
> It would be nice to have a comment in the workflow pointing to that bug so that we have a better change of dropping that workaround once the bug is fixed.

Very good point, done!

simondeziel (dismissed):

.github/workflows/security.yml
```diff
@@ -26,23 +26,28 @@ jobs:
...
-        run: mkdir -p /home/runner/vuln-cache
+      # We are using the deb package instead of the trivy action because we need to point to the cache directory
+      # with cache-dir, and there is currently a bug in that functionality on the Trivy GitHub action.
+      # See https://github.com/aquasecurity/trivy-action/issues/12.
+      - name: Install Trivy
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          mkdir -p /home/runner/vuln-cache
+          latest=$(gh release view --repo aquasecurity/trivy --json tagName -q '.tagName' | head -n 1)
```
tomponline, line 37: Lets use the official install steps for this (I think I mentioned this previously):

```
sudo apt-get install wget apt-transport-https gnupg
wget -qO - https://aquasecurity.github.io/trivy-repo/deb/public.key | gpg --dearmor | sudo tee /usr/share/keyrings/trivy.gpg > /dev/null
echo "deb [signed-by=/usr/share/keyrings/trivy.gpg] https://aquasecurity.github.io/trivy-repo/deb generic main" | sudo tee -a /etc/apt/sources.list.d/trivy.list
sudo apt-get update
sudo apt-get install trivy
```

https://aquasecurity.github.io/trivy/v0.56/getting-started/installation/

tomponline: speed isnt really an issue here, talking seconds, better in my view to use official install steps.

tomponline (approved):

Issue #14138, opened 2024-09-19
`lxc config device` shell completion improvements


kadinsayani:
`lxc config device add <instance_name>` should suggest:

- instance devices

`lxc config device add <instance_name> <device>` should suggest:

- device options

kadinsayani:
@masnax let me know if I'm missing anything here 😄

PR #14194, opened 2024-10-02
Add shell completions for `lxc config device`

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/14138

Adds shell completions for `lxc config device add`:

- `lxc config device add [TAB]` shows all possible devices; and
- `lxc config device add <device> [TAB]` shows all possible device options.

Shell completions for `lxc config device get/set/unset` are unchanged, as they currently return set config options from either the config or profile (whichever exists), and I think it makes sense to keep it this way. Please let me know if you have any suggestions 😄

### DIFF
--- a/lxc/completion.go
+++ b/lxc/completion.go
@@ -366,6 +366,37 @@ func (g *cmdGlobal) cmpInstanceAllDevices(instanceName string) ([]string, cobra.
 	return devices, cobra.ShellCompDirectiveNoFileComp
 }
 
+func (g *cmdGlobal) cmpInstanceAllDeviceOptions(instanceName string, deviceName string) ([]string, cobra.ShellCompDirective) {
+	resources, err := g.ParseServers(instanceName)
+	if err != nil || len(resources) == 0 {
+		return nil, cobra.ShellCompDirectiveError
+	}
+
+	resource := resources[0]
+	client := resource.server
+
+	metadataConfiguration, err := client.GetMetadataConfiguration()
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveError
+	}
+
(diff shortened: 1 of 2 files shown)


simondeziel:
> Please let me know if you have any suggestions 😄

Have you considered `lxc config device override`? Those should suggest only devices that are coming from the profile(s) so maybe that's out of scope... but then I'd retitle the PR to: Add shell completions for `lxc config device add` ;)

simondeziel:
> > Please let me know if you have any suggestions 😄
> 
> Have you considered `lxc config device override`? Those should suggest only devices that are coming from the profile(s) so maybe that's out of scope... but then I'd retitle the PR to: Add shell completions for `lxc config device add` ;)

Maybe that's asking too much because even commands like `lxc config device get v1 root path` don't track down devices coming from profiles:

```
$ lxc config device get v1 root path
Error: Device from profile(s) cannot be retrieved for individual instance
$ lxc profile device get default root path
/
```

lxc/completion.go
```diff
@@ -339,6 +339,64 @@ func (g *cmdGlobal) cmpInstanceDeviceNames(instanceName string) ([]string, cobra
...
+	metadataConfiguration, err := client.GetMetadataConfiguration()
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveError
+	}
+
+	devices := make([]string, 0, len(metadataConfiguration.Configs))
+
+	for key := range metadataConfiguration.Configs {
+		if strings.HasPrefix(key, "device-") {
+			parts := strings.Split(key, "-")
```
simondeziel, line 360: A case for `strings.SplitN(key, "-", 3)`?

simondeziel (approved):

PR #14195, opened 2024-10-03
fix:fix slice init length

cuishuang:
The intention here should be to initialize a slice with a capacity of len(vols) rather than initializing the length of this slice.

The online demo:  https://go.dev/play/p/q1BcVCmvidW


tomponline:
Thanks for this, we can cherry-pick from https://github.com/lxc/incus/pull/1285

Issue #12234, opened 2023-09-11
Support for VLAN tag on OVN uplink network


fnordahl:
When attaching an OVN network with an external network there is currently no way to specify a VLAN tag to use.

On systems where you have no extra NIC for external networking it is useful to be able to use the same NIC for multiple purposes. A way of doing this is for example to put the main interface into an OVS bridge without any tag configuration. An example of this could be done is laid out in #12233.

When creating the `localnet` port in OVN it is possible to specify which VLAN tag it should use, when doing it this way the traffic will be appropriately tagged even if the interface in the associated bridge is not tagged.

For reference, OpenStack Neutron makes use of the `tag` column when creating the `localnet` Logical_Switch_Port: https://github.com/openstack/neutron/blob/8cba9a2ee86cb3b65645674ef315c14cfb261143/neutron/plugins/ml2/drivers/ovn/mech_driver/ovsdb/ovn_client.py#L1872

tomponline:
Does using the `vlan` option on the `physical` uplink network work?

https://documentation.ubuntu.com/lxd/en/latest/reference/network_physical/#configuration-options

tomponline:
@boltmark see comment for how we may be able to approach this.

tomponline:
I have got it working as follows:

1. On the localnet OVN logical_switch_port that connects to the uplink provide bridge, we need to set vlan_request to the specified VLAN ID. E.g. ovn-nbctl set logical_switch_port lxd-net3-ls-ext-lsp-provider tag_request=101. This results in egress packets from the OVN router being tagged as they go onto the provide OVS bridge. This change is cluster wide - just OVN northbound DB.

2. Configure the OVS port that represents the connection to the uplink network parent interface to trunk the specified VLAN. E.g. ovs-vsctl set port lxdovn1b vlan_mode=trunk trunk=101 where lxdovn1b is the end of the veth pair that connects to the OVS provider bridge from lxdbr0 (acting as the uplink) to the . This results in packets from the OVN router to traverse the OVS provider switch to the uplink parent interface still being tagged. This change is per-cluster member as the OVS bridge is local to each cluster member. 

[...]

PR #14196, opened 2024-10-03
Network: Add support for OVN uplink networks attached to a VLAN

boltmark:
This PR adds support for OVN uplink networks attached to a VLAN. The implementation details closely track https://github.com/canonical/lxd/issues/12234#issuecomment-2379005076.

Testing here: https://github.com/canonical/lxd-ci/pull/302.

Closes https://github.com/canonical/lxd/issues/12234.

### DIFF
--- a/lxd/network/driver_ovn.go
+++ b/lxd/network/driver_ovn.go
@@ -1531,6 +1531,14 @@ func (n *ovn) startUplinkPortBridgeNative(uplinkNet Network, bridgeDevice string
 		return fmt.Errorf("Failed to bring up uplink veth interface %q: %w", vars.uplinkEnd, err)
 	}
 
+	// Add VLAN filter entry to the uplink end of the veth interface.
+	if uplinkNetConfig["vlan"] != "" {
+		err = link.BridgeVLANAdd(uplinkNetConfig["vlan"], true, true, false)
+		if err != nil {
+			return fmt.Errorf("Failed to configure VLAN for uplink veth interface %q: %w", vars.uplinkEnd, err)
+		}
+	}
+
 	// Ensure uplink OVS end veth interface is up.
 	link = &ip.Link{Name: vars.ovsEnd}
 	err = link.SetUp()
@@ -1644,6 +1652,11 @@ func (n *ovn) startUplinkPortPhysical(uplinkNet Network) error {
 	// Detect if uplink interface is a native bridge.
 	if IsNativeBridge(uplinkHostName) {
 		return n.startUplinkPortBridgeNative(uplinkNet, uplinkHostName)
+	}
+
+	// Handle case where uplink interface is bridge and VLAN is specified.
+	if IsNativeBridge(uplinkConfig["parent"]) && uplinkConfig["vlan"] != "" {
+		return n.startUplinkPortBridgeNative(uplinkNet, uplinkConfig["parent"])
 	}
 
 	// Detect if uplink interface is a OVS bridge.
@@ -1853,7 +1866,7 @@ func (n *ovn) deleteUplinkPortPhysical(uplinkNet Network) error {
 	uplinkHostName := GetHostDevice(uplinkConfig["parent"], uplinkConfig["vlan"])
 
 	// Detect if uplink interface is a native bridge.
-	if IsNativeBridge(uplinkHostName) {
+	if IsNativeBridge(uplinkHostName) || (IsNativeBridge(uplinkConfig["parent"]) && uplinkConfig["vlan"] != "") {
 		return n.deleteUplinkPortBridgeNative(uplinkNet)
 	}
 


lxd/network/openvswitch/ovn.go
```diff
@@ -119,8 +119,9 @@ type OVNSwitchPortOpts struct {
 	DHCPv4OptsID OVNDHCPOptionsUUID // Optional, if empty, no DHCPv4 enabled on port.
 	DHCPv6OptsID OVNDHCPOptionsUUID // Optional, if empty, no DHCPv6 enabled on port.
 	Parent       OVNSwitchPort      // Optional, if set a nested port is created.
-	VLAN         uint16             // Optional, use with Parent to request a specific VLAN for nested port.
+	NestedVLAN   uint16             // Optional, use with Parent to request a specific VLAN for nested port.
```
tomponline, line 122: Shall we call this ParentVLAN to show it is related to Parent field?

lxd/network/openvswitch/ovn.go
```diff
@@ -1269,6 +1270,10 @@ func (o *OVN) LogicalSwitchPortAdd(switchName OVNSwitch, portName OVNSwitchPort,
 		if opts.Location != "" {
 			args = append(args, "--", "set", "logical_switch_port", string(portName), fmt.Sprintf("external_ids:%s=%s", ovnExtIDLXDLocation, opts.Location))
 		}
+
+		if opts.VLAN != 0 {
```
tomponline, line 1274: If opts.VLAN == 0 we should unset the tag_request field (and confirm that doing that clears the tag field set by OVN) so that one can disable the VLAN setting later.

lxd/network/driver_ovn.go
```diff
@@ -3104,6 +3161,13 @@ func (n *ovn) Start() error {
 		return err
 	}
 
+	// If VLAN config is provided to the OVN network, configure the necessary ports
+	// to handle traffic on the specified VLAN.
+	err = n.configurePortsForVLAN(n.config["vlan"], false)
```
tomponline, line 3166: Can we make this called, or be part of, startUplinkPort() as that can then avoid loading the uplink network from the DB again, and also it keeps the uplink port configuration together, rather than splitting it out.

lxd/network/driver_ovn.go
```diff
@@ -3218,6 +3282,12 @@ func (n *ovn) Update(newNetwork api.NetworkPut, targetNode string, clientType re
 		delete(newNetwork.Config, ovnVolatileUplinkIPv6)
 	}
 
+	// Update port configurations for VLAN if necessary.
+	err = n.configurePortsForVLAN(newNetwork.Config["vlan"], true)
```
tomponline, line 3286: where do changes to uplink ports get applied during updates currently? Can we move this call there?

lxd/network/driver_ovn.go
```diff
@@ -2302,8 +2348,19 @@ func (n *ovn) setup(update bool) error {
 			return fmt.Errorf("Failed linking external router port to external switch port: %w", err)
 		}
 
+		// If VLAN is provided, ensure localnet port is properly tagged.
+		localnetOpts := &openvswitch.OVNSwitchPortOpts{}
+		if n.config["vlan"] != "" {
```
tomponline, line 2353: what if vlan is unset, how is the vlan setting removed?

lxd/network/driver_ovn.go
```diff
@@ -1430,6 +1430,52 @@ func (n *ovn) startUplinkPort() error {
...
+	}
+
+	uplinkNet, err := LoadByName(n.state, api.ProjectDefaultName, n.config["network"])
+	if err != nil {
+		return err
+	}
+
+	ovs := openvswitch.NewOVS()
+	vars := n.uplinkPortBridgeVars(uplinkNet)
+	uplinkHostName := GetHostDevice(uplinkNet.Config()["parent"], uplinkNet.Config()["vlan"])
```
tomponline, line 1447: Hrm, so referring to my question https://github.com/canonical/lxd/issues/12234#issuecomment-2379610382 that we didn't discuss yet:

> Should we add a vlan setting to the uplink network or each OVN network?

You've gone with adding the `vlan` setting to each OVN network in this PR.

Because each uplink network can be used by multiple OVN networks, this means there is the possibility of using multiple VLANs on a single uplink (which in principle would work as we are trunking VLANs through the uplink OVS bridge into the uplink interface).

But I don't think using the uplink's vlan setting in the call to ` GetHostDevice` makes sense in that case.

I also see some other places in the existing code where we've called `uplinkHostName := GetHostDevice` and used the uplink's VLAN setting which needs to be reviewed as part of this PR.

Right now, say I had an eth0.101 interface (a physical port eth0, with an untagged VLAN 101 interface layered ontop) which I want to use for my uplink, you could create a physical network that has `parent=eth0` and `vlan=101` and when passing directly to an instance, LXD would check if the eth0.101 interface existed, and if not create it, before passing it to the instance.

For OVN uplink networks, I can't see that LXD would create the eth0.101 interface if it doesn't exist, but if it does exist we would then connect eth0.101 to the intermediate uplink OVS bridge like any other port.

But in this PR you are introducing per-OVN network uplink VLAN settings, so this VLAN setting on the uplink network conceptually conflicts with what we are trying to do, as you are trying to allow multiple OVN networks to trunk multiple VLANs through a single uplink network, whilst at the same time allowing the uplink port to be configured with a single untagged VLAN (which wouldn't work with tagged frames).

But stepping back a bit further, lets forget the uplink's existing `vlan` setting for a moment.

Say we go ahead with your proposal here of adding `vlan` to the OVN network itself, then if I have two OVN networks both connected to the same uplink but using two different uplink VLAN IDs, how do we allocate an IP for the OVN network's external (volatile) IP on the uplink network, as each OVN network is effectively connected to different uplink networks (because each VLAN on the uplink network is likely going to have different IP ranges, or at least conceptually separate IP ranges).

So going back to my question above:

> Should we add a vlan setting to the uplink network or each OVN network?

I actually think we should keep the `vlan` setting on the uplink network. 
Don't introduce a `vlan` setting on the OVN network.

This would then require that the administrator define separate uplink networks in LXD for each VLAN on the parent interface, along with the separate IP range settings (which is what we want).

So the question then becomes: How should we use the `vlan` setting on the uplink network?

Also it seems that whilst the OVN network doesn't create the vlan interface (e.g. eth0.101), if the uplink is a physical network it does seem to create it:

https://github.com/canonical/lxd/blob/main/lxd/network/driver_physical.go#L318

So, because we already have `uplinkHostName := GetHostDevice(uplinkConfig["parent"], uplinkConfig["vlan"])` in `startUplinkPortPhysical` could it be that if you use an OVN network with a physical network as an uplink using a dedicated interface (not a bridge) that VLAN support already works?

tomponline: >  could it be that if you use an OVN network with a physical network as an uplink using a dedicated interface (not a bridge) that VLAN support already works?

I've just tested this and it works already, but only for physical uplinks, not bridged uplinks.

We should probably just add the last step from https://github.com/canonical/lxd/issues/12234#issuecomment-2379604497 `startUplinkPortBridgeNative` and the equivalent to `startUplinkPortBridgeOVS` to this PR instead:

> bridge vlan add dev <interface from the uplink OVS bridge to uplink> vid 101

boltmark: Thanks @tomponline. I've pushed changes which better handle the case where the uplink parent is a bridge and the `vlan` config is set -- we now avoid using the physical interface lxd creates on top of the bridge. This is WIP as I have a couple questions:

> We should probably just add the last step from https://github.com/canonical/lxd/issues/12234#issuecomment-2379604497 startUplinkPortBridgeNative and the equivalent to startUplinkPortBridgeOVS to this PR instead:

Can you help me to understand what we would need to add to `startUplinkPortBridgeOVS`?

Also, since the `vlan` config is set on the uplink network, I'm not sure what the best approach is to ensure ports are reconfigured when the uplink network `vlan` config is changed. Would we need to add a check in the network update handler which notifies cluster members to update their ports if (1) the updated network is used as an uplink for an OVN network and (2) the `vlan` config is changed? This seems a bit indirect so I'm wondering if there is something better we can do.

lxd/network/driver_ovn.go
```diff
@@ -1853,7 +1866,7 @@ func (n *ovn) deleteUplinkPortPhysical(uplinkNet Network) error {
 	uplinkHostName := GetHostDevice(uplinkConfig["parent"], uplinkConfig["vlan"])
 
 	// Detect if uplink interface is a native bridge.
-	if IsNativeBridge(uplinkHostName) {
+	if IsNativeBridge(uplinkHostName) || IsNativeBridge(uplinkConfig["parent"]) && uplinkConfig["vlan"] != "" {
```
tomponline, line 1869: do we need brackets for the logic here?

boltmark: Sure, added parens to improve readability

tomponline (approved):

PR #14099, opened 2024-09-13
Auth: Fine-grained access control for TLS clients

markylaing:
This PR adds:
- The ability to create TLS identities via `lxc auth identity create [<remote>:]tls/<name> [<path_to_cert_file>]`. If a file is provided, the certificate is sent as a base64 encoded, PEM encoded certificate in the request body. If a file is not provided, an identity of type `Client certificate (pending)` is created, whose identifier is a UUID. The pending client certificate identity works like a certificate add operation - it's metadata contains a secret and an expiry time.
- The ability for untrusted clients to update their pending client certificate to a `Client certificate`, with a valid token.
- Updates to `lxd remote add` to handle these cases.
- The ability to delete TLS and OIDC identities.
- Restrictions to `/1.0/certificates` so that the API does not show the new TLS identity type.
- The ability to add these new TLS identities to groups.

Here are some examples:
```
$ lxc auth group create my-group
Group my-group created
$ lxc auth group permission add my-group project default operator
$ lxc auth identity create tls/jane-doe
eyJjbGllbnRfbmFtZSI6ImphbmUtZG9lIiwiZmluZ2VycHJpbnQiOiI1MGYxZWUwYTUxYjRmNzQ0MDcyODAwNTJiMGQ5YzM3MzFkNGIzNjE2ZDE0ZWYwMjMzOTNhZjAzMGRmYzEyMjdhIiwiYWRkcmVzc2VzIjpbIjEyNy4wLjAuMTo4NDQzIl0sInNlY3JldCI6ImMxNjgzYzljYmNlOThhMmZjOTk3MmQwYWNhZTI1MDg5NTU1OWM5YjVkMmZhZDc5MDU5OWY2NTQ3OGYwMTRiMTkiLCJleHBpcmVzX2F0IjoiMDAwMS0wMS0wMVQwMDowMDowMFoiLCJpZGVudGl0eV90eXBlIjoiQ2xpZW50IGNlcnRpZmljYXRlIn0=
[...]

Issue #12815, opened 2024-02-04
Improve lxd-generate pluralisation


markylaing:
The plural form of `identity` is `identities`, but lxd-generate is outputting `identitys`.

See https://github.com/canonical/lxd/pull/12807#discussion_r1474731848_

markylaing:
I'm going to close this as it is so far down on the list of priorities I don't think we'll ever get to it. A misspelled variable that is generated does not impact functionality.

PR #14199, opened 2024-10-03
Certificate db follow up

markylaing:
In the specification for fine-grained authorization for TLS clients, the `/1.0/certificates` API does not return identities of the new type (because they can't be managed via this API).

In the PoC PR (#14099) this is done in a bit of a hacky way. They are essentially filtered out because they don't have a matching certificate type.

This PR does this in a more correct way, by ensuring that certificate database queries only return identities of the correct type. I've done some additional clean up as well.


lxd/db/cluster/certificates.go
```diff
@@ -115,6 +115,22 @@ func (cert Certificate) ToIdentity() (*Identity, error) {
...
+	AND type in (%d, %d, %d, %d, %d)
+`, authMethodTLS,
+	identityTypeCertificateServer,
+	identityTypeCertificateClientRestricted,
+	identityTypeCertificateClientUnrestricted,
+	identityTypeCertificateMetricsRestricted,
+	identityTypeCertificateMetricsUnrestricted,
+)
+
+var getCertificateIdentityByFingerprintStmt = fmt.Sprintf(`%s AND identifier = ?`, getCertificateIdentitiesStmt)
```
tomponline, line 130: Rather than store the query 3 times in memory, can you generate the query inside the relevant function using just getCertificateIdentitiesStmt as the basis?

Also, there's no need to use `fmt.Sprintf` as that uses reflection to detect the input type, instead you can just use `+` to concatenate the strings.

lxd/cluster/membership.go
```diff
@@ -1153,7 +1154,7 @@ func Purge(c *db.Cluster, name string) error {
 			return fmt.Errorf("Failed to remove member %q: %w", name, err)
 		}
 
-		err = cluster.DeleteCertificates(context.Background(), tx.Tx(), name, certificate.TypeServer)
+		err = cluster.DeleteIdentitys(ctx, tx.Tx(), name, api.IdentityTypeCertificateServer)
```
tomponline, line 1158: is this a DB generator bug? The incorrect pluralisation of  Identity?

tomponline (commented):
Do we also need to delete `type CertificateFilter`?

lxd/db/cluster/certificates.go
markylaing: Yep will do.

lxd/cluster/membership.go
markylaing: Yep, there's an issue for it here: https://github.com/canonical/lxd/issues/12815

There's also a JIRA ticket for it in my backlog.

markylaing:
> Do we also need to delete `type CertificateFilter`?

Probably yes. I'll double check.

lxd/db/cluster/certificates.go
markylaing: Done.

tomponline (approved):

tomponline:
Thanks

PR #14200, opened 2024-10-03
lxd/devices: use builtin `min()`

simondeziel:
`min()` is builtin since Go 1.21, https://tip.golang.org/ref/spec#Min_and_max


tomponline (approved):
Good catch!

PR #14201, opened 2024-10-03
doc/installing: update link to how to manage the LXD snap

simondeziel:


tomponline (approved):
Good catch!

simondeziel:
> Good catch!

Good thing you copied the post and updated the linuxcontainers.org one with a link to the Ubuntu one so thank **you**.

PR #14205, opened 2024-10-04
Makefile: Make lxd-user and lxd-benchmark static binaries

tomponline:


simondeziel (approved):

simondeziel:
Since the snap recipe doesn't use those make targets, https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft.yaml#L1403 will need updating too.

PR #14191, opened 2024-10-02
Auth: Implement identity deletion

markylaing:
- Separates identity endpoints by authentication method (instead of having the authentication method as a path argument) in preparation for fine-grained TLS identites. 
- Implements deletion of TLS and OIDC identities (for all current TLS identities, this will be rejected because they are not fine-grained).
- Adds a `access_management_tls` API extension which will apply to all upcoming changes for TLS fine-grained auth.

Note: The licence check was failing in my local static analysis, not sure what's happening there.
Note 2: (I just needed to update it locally :+1:).

Closes #13149

PR #14207, opened 2024-10-04
Auth: Creation of fine-grained TLS identites

markylaing:
Implements creation and subsequent authentication/authorization of fine-grained TLS identities.

Depends on https://github.com/canonical/lxd/pull/14191

Still to do:
- [x] Documentation

### DIFF
--- a/lxc/auth.go
+++ b/lxc/auth.go
@@ -1,13 +1,17 @@
 package main
 
 import (
+	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
+	"net/url"
 	"os"
 	"sort"
 	"strings"
 
+	"github.com/google/uuid"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v2"
 
@@ -18,6 +22,7 @@ import (
 	"github.com/canonical/lxd/shared/entity"
 	"github.com/canonical/lxd/shared/i18n"
 	"github.com/canonical/lxd/shared/termios"
+	"github.com/canonical/lxd/shared/version"
 )
 
 type cmdAuth struct {
@@ -732,6 +737,9 @@ func (c *cmdIdentity) command() *cobra.Command {
 	cmd.Long = cli.FormatSection(i18n.G("Description"), i18n.G(
 		`Manage identities`))
 
+	identityCreateCmd := cmdIdentityCreate{global: c.global}
+	cmd.AddCommand(identityCreateCmd.command())
+
 	identityListCmd := cmdIdentityList{global: c.global}
 	cmd.AddCommand(identityListCmd.command())
 
@@ -754,6 +762,147 @@ func (c *cmdIdentity) command() *cobra.Command {
 	cmd.Args = cobra.NoArgs
 	cmd.Run = func(cmd *cobra.Command, args []string) { _ = cmd.Usage() }
 	return cmd
+}
+
+type cmdIdentityCreate struct {
+	global     *cmdGlobal
+	flagGroups []string
+}
+
+func (c *cmdIdentityCreate) command() *cobra.Command {
+	cmd := &cobra.Command{}
+	cmd.Use = usage("create", i18n.G("[<remote>:]<authentication_method>/<name> [<path to PEM encoded certificate>] [[--group <group_name>]]"))
+	cmd.Short = i18n.G("Create an identity")
+	cmd.Long = cli.FormatSection(i18n.G("Description"), i18n.G(
+		`Create a TLS identity`))
+
+	cmd.RunE = c.run
+	cmd.Flags().StringSliceVarP(&c.flagGroups, "group", "g", []string{}, "Groups to add to the identity")
+
+	return cmd
+}
+
+func (c *cmdIdentityCreate) run(cmd *cobra.Command, args []string) error {
+	var stdinData api.IdentitiesTLSPost
+
+	// Quick checks.
+	exit, err := c.global.CheckArgs(cmd, args, 1, 2)
+	if exit {
+		return err
+	}
+
+	// If stdin isn't a terminal, read text from it
+	if !termios.IsTerminal(getStdinFd()) {
+		contents, err := io.ReadAll(os.Stdin)
+		if err != nil {
+			return err
+		}
+
+		err = yaml.Unmarshal(contents, &stdinData)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Parse remote
+	remoteName, resourceName, err := c.global.conf.ParseRemote(args[0])
+	if err != nil {
+		return err
+	}
+
+	transporter, wrapper := newLocationHeaderTransportWrapper()
+	client, err := c.global.conf.GetInstanceServerWithTransportWrapper(remoteName, wrapper)
+	if err != nil {
+		return err
+	}
+
+	if resourceName == "" {
+		return errors.New(i18n.G("Missing identity argument"))
+	}
+
+	authMethod, name, ok := strings.Cut(resourceName, "/")
+	if !ok {
+		return errors.New(i18n.G("Malformed argument, expected `[<remote>:]<authentication_method>/<name>`, got ") + args[0])
+	}
+
+	if authMethod != api.AuthenticationMethodTLS {
+		return errors.New(i18n.G("Identity creation only supported for TLS identities"))
+	}
+
+	// Add name and groups to any stdin data
+	stdinData.Name = name
+	for _, group := range c.flagGroups {
+		if !shared.ValueInSlice(group, stdinData.Groups) {
+			stdinData.Groups = append(stdinData.Groups, group)
+		}
+	}
+
+	// If the certificate argument is provided, read it and add it to the stdin data.
+	if len(args) == 2 {
+		pemEncodedX509Cert, err := os.ReadFile(args[1])
+		if err != nil {
+			return err
+		}
+
+		stdinData.Certificate = string(pemEncodedX509Cert)
+	}
+
+	// Expect that if the caller did not provide a certificate then they want to get a token.
+	if stdinData.Certificate == "" {
+		stdinData.Token = true
+		token, err := client.CreateIdentityTLSToken(stdinData)
+		if err != nil {
+			return err
+		}
+
+		if !c.global.flagQuiet {
+			pendingIdentityURL, err := url.Parse(transporter.location)
+			if err != nil {
+				return fmt.Errorf("Received invalid location header %q: %w", transporter.location, err)
+			}
+
+			var pendingIdentityUUIDStr string
+			identityURLPrefix := api.NewURL().Path(version.APIVersion, "auth", "identities", authMethod).String()
+			_, err = fmt.Sscanf(pendingIdentityURL.Path, identityURLPrefix+"/%s", &pendingIdentityUUIDStr)
+			if err != nil {
+				return fmt.Errorf("Received unexpected location header %q: %w", transporter.location, err)
+			}
+
+			pendingIdentityUUID, err := uuid.Parse(pendingIdentityUUIDStr)
+			if err != nil {
+				return fmt.Errorf("Received invalid pending identity UUID %q: %w", pendingIdentityUUIDStr, err)
+			}
+
+			fmt.Printf(i18n.G("TLS identity %q (%s) pending identity token:")+"\n", resourceName, pendingIdentityUUID.String())
+		}
+
+		// Encode certificate add token to JSON.
+		tokenJSON, err := json.Marshal(token)
+		if err != nil {
+			return fmt.Errorf("Failed to encode identity token: %w", err)
+		}
+
+		// Print the base64 encoded token.
+		fmt.Println(base64.StdEncoding.EncodeToString(tokenJSON))
+		return nil
+	}
+
+	fingerprint, err := shared.CertFingerprintStr(stdinData.Certificate)
+	if err != nil {
+		return err
+	}
+
+	// Otherwise create the identity directly.
+	err = client.CreateIdentityTLS(stdinData)
+	if err != nil {
+		return err
+	}
+
+	if !c.global.flagQuiet {
+		fmt.Printf(i18n.G("TLS identity %q created with fingerprint %q")+"\n", resourceName, fingerprint)
+	}
+
+	return nil
 }
 
 type cmdIdentityList struct {
--- a/lxc/remote.go
+++ b/lxc/remote.go
@@ -171,14 +171,14 @@ func (c *cmdRemoteAdd) runToken(addr string, server string, token string, rawTok
 
 	// If address is provided, use token on that specific address.
 	if addr != "" {
-		return c.addRemoteFromToken(addr, server, token, rawToken.Fingerprint)
+		return c.addRemoteFromToken(addr, server, token, *rawToken)
 	}
 
 	// Otherwise, iterate over all addresses within the token.
 	for _, addr := range rawToken.Addresses {
 		addr = fmt.Sprintf("https://%s", addr)
 
-		err := c.addRemoteFromToken(addr, server, token, rawToken.Fingerprint)
+		err := c.addRemoteFromToken(addr, server, token, *rawToken)
 		if err != nil {
 			if api.StatusErrorCheck(err, http.StatusServiceUnavailable) {
 				continue
@@ -203,7 +203,7 @@ func (c *cmdRemoteAdd) runToken(addr string, server string, token string, rawTok
 		return errors.New(i18n.G("Failed to add remote"))
 	}
 
-	err = c.addRemoteFromToken(string(line), server, token, rawToken.Fingerprint)
+	err = c.addRemoteFromToken(string(line), server, token, *rawToken)
 	if err != nil {
 		return err
 	}
@@ -211,7 +211,7 @@ func (c *cmdRemoteAdd) runToken(addr string, server string, token string, rawTok
 	return nil
 }
 
-func (c *cmdRemoteAdd) addRemoteFromToken(addr string, server string, token string, fingerprint string) error {
+func (c *cmdRemoteAdd) addRemoteFromToken(addr string, server string, token string, rawToken api.CertificateAddToken) error {
 	conf := c.global.conf
 
 	var certificate *x509.Certificate
@@ -227,7 +227,7 @@ func (c *cmdRemoteAdd) addRemoteFromToken(addr string, server string, token stri
 		}
 
 		certDigest := shared.CertFingerprint(certificate)
-		if fingerprint != certDigest {
+		if rawToken.Fingerprint != certDigest {
 			return fmt.Errorf(i18n.G("Certificate fingerprint mismatch between certificate token and server %q"), addr)
 		}
 
@@ -260,19 +260,33 @@ func (c *cmdRemoteAdd) addRemoteFromToken(addr string, server string, token stri
 		return api.StatusErrorf(http.StatusServiceUnavailable, "%s: %w", i18n.G("Unavailable remote server"), err)
 	}
 
-	req := api.CertificatesPost{}
-	if d.HasExtension("explicit_trust_token") {
-		req.TrustToken = token
-	} else {
-		req.Password = token
-	}
-
 	// Add client certificate to trust store. Even if we are already trusted (src.Auth == "trusted"),
 	// we want to send the token to invalidate it. Therefore, we can ignore the conflict error, which
 	// is thrown if we are trying to add a client cert that is already trusted by LXD remote.
-	err = d.CreateCertificate(req)
-	if err != nil && !api.StatusErrorCheck(err, http.StatusConflict) {
-		return err
+	//
+	// If "type" is not set on the token, the token was issued by the certificates API and CreateCertificate should be
+	// called. If "type" is set, the token was issued by the auth API and CreateIdentityTLS should be called.
+	if rawToken.Type == "" {
+		req := api.CertificatesPost{}
+		if d.HasExtension("explicit_trust_token") {
+			req.TrustToken = token
+		} else {
+			req.Password = token
+		}
+
+		err = d.CreateCertificate(req)
+		if err != nil && !api.StatusErrorCheck(err, http.StatusConflict) {
+			return err
+		}
+	} else {
+		req := api.IdentitiesTLSPost{
+			TrustToken: token,
+		}
+
+		err = d.CreateIdentityTLS(req)
+		if err != nil && !api.StatusErrorCheck(err, http.StatusConflict) {
+			return err
+		}
 	}
 
 	// And check if trusted now.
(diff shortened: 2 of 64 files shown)


markylaing:
@mionaalex I've requested your review here for the documentation changes. When you have a minute, could you please take a look at https://canonical-ubuntu-documentation-library--14207.com.readthedocs.build/lxd/en/14207/

The relevant pages are:
- https://canonical-ubuntu-documentation-library--14207.com.readthedocs.build/lxd/en/14207/authentication/
- https://canonical-ubuntu-documentation-library--14207.com.readthedocs.build/lxd/en/14207/explanation/authorization/
- https://canonical-ubuntu-documentation-library--14207.com.readthedocs.build/lxd/en/14207/howto/server_expose/
- https://canonical-ubuntu-documentation-library--14207.com.readthedocs.build/lxd/en/14207/howto/projects_confine/

Thank you :)

test/suites/auth.sh
```diff
@@ -79,11 +79,33 @@ test_authorization() {
...
+  ! LXD_CONF="${LXD_CONF3}" lxc remote add tls "${tls_identity_token2}" || false
+
+  # Check users have been added to the group.
+  tls_identity_fingerprint="$(cert_fingerprint "${LXD_CONF2}/client.crt")"
   lxc auth identity list --format csv | grep -Fq 'oidc,OIDC client," ",test-user@example.com,test-group'
+  lxc auth identity list --format csv | grep -Fq "tls,Client certificate,test-user,${tls_identity_fingerprint},test-group"
 
   # Test `lxc auth identity info`
-  expected=$(cat << EOF
+  expectedOIDCInfo=$(cat << EOF
```
simondeziel, line 108: I think we could do without the subshell and the `cat` by just using regular double quotes that spans multiple lines.

test/suites/auth.sh
```diff
@@ -95,7 +117,21 @@ effective_groups:
 effective_permissions: []
 EOF
 )
-  lxc auth identity info oidc: | grep -Fz "${expected}"
+  lxc auth identity info oidc: | grep -Fz "${expectedOIDCInfo}"
+
+  expectedTLSInfo=$(cat << EOF
```
simondeziel, line 122: Same here re subshell/`cat`.

test/suites/auth.sh
```diff
@@ -95,7 +117,21 @@ effective_groups:
 effective_permissions: []
 EOF
 )
-  lxc auth identity info oidc: | grep -Fz "${expected}"
+  lxc auth identity info oidc: | grep -Fz "${expectedOIDCInfo}"
```
simondeziel, line 120: I went to the man page for `grep -z` ;)
Still not sure why a `null` char is wanted here.

test/suites/auth.sh
```diff
@@ -79,11 +79,33 @@ test_authorization() {
...
   lxc auth identity group add oidc/test-user@example.com test-group # Valid
 
-  # Check user has been added to the group.
+  # Test fine-grained TLS identity creation
+  tls_identity_token="$(lxc auth identity create tls/test-user --quiet --group test-group)"
+  LXD_CONF2=$(mktemp -d -p "${TEST_DIR}" XXX)
+  LXD_CONF="${LXD_CONF2}" gen_cert_and_key "client"
+
+  # Cannot use the token with the certificates API and the correct error is returned.
+  [ "$(curl --insecure --cert "${LXD_CONF2}/client.crt" --key "${LXD_CONF2}/client.key" -X POST "https://${LXD_ADDR}/1.0/certificates" --data "{\"trust_token\": \"${tls_identity_token}\"}" | jq -er '.error')" = "Failed during search for certificate add token operation: TLS Identity token detected (you must update your client)" ]
```
simondeziel, line 88: Could we use our `my_curl` wrapper like this? Despite using `--silent`, the error still shows up.

```suggestion
  [ "$(LXD_CONF="${LXD_CONF2}" my_curl -X POST "https://${LXD_ADDR}/1.0/certificates" --data "{\"trust_token\": \"${tls_identity_token}\"}" | jq -er '.error')" = "Failed during search for certificate add token operation: TLS Identity token detected (you must update your client)" ]
```

test/suites/auth.sh
```diff
@@ -79,11 +79,33 @@ test_authorization() {
...
+
+  # Can use the token with remote add command.
+  LXD_CONF="${LXD_CONF2}" lxc remote add tls "${tls_identity_token}"
+  [ "$(LXD_CONF="${LXD_CONF2}" lxc_remote query tls:/1.0 | jq -r '.auth')" = 'trusted' ]
+
+  # Check a token cannot be used when expired
+  lxc config set core.remote_token_expiry=1S
```
simondeziel, line 95: Just a nit but the time unit capitalization sometimes matters like `M` means months and `m` minutes. That said, `S` is probably not ambiguous but it still caught my attention ;)

test/suites/auth.sh
simondeziel: **If** the full command output is supposed to match the `${expectedOIDCInfo}` content, this could be done in a simple comparison `[ "$(lxc auth identity info oidc:)" = "${expectedOIDCInfo}" ]` as that works on multi-line variables too.

test/suites/auth.sh
markylaing: Lower-case "s" isn't allowed as a value unfortunately.

test/suites/auth.sh
markylaing: Done

test/suites/auth.sh
markylaing: grep -z was used so that it would match across multiple lines. I've changed to use the plain `=` test.

test/suites/auth.sh
markylaing: I didn't know about `my_curl`. Sorted.

test/suites/auth.sh
simondeziel: I didn't know that `grep -z` trick, thanks!

shared/api/auth.go
```diff
@@ -105,6 +105,41 @@ type IdentityPut struct {
 	Groups []string `json:"groups" yaml:"groups"`
 }
 
+// IdentitiesPostTLS contains required information for the creation of a TLS identity.
+//
+// swagger:model
+//
+// API extension: access_management_tls.
+type IdentitiesPostTLS struct {
```
tomponline, line 113: I believe the verb goes at the end, so this should be IdentitiesTLSPost right?

shared/api/auth.go
```diff
@@ -105,6 +105,41 @@ type IdentityPut struct {
...
+	// Example: ["foo", "bar"]
+	Groups []string `json:"groups" yaml:"groups"`
+}
+
+// IdentityTokenTLS contains a token that can be used by an untrusted client to gain trust with a LXD system.
+//
+// swagger:model
+//
+// API extension: access_management_tls.
+type IdentityTokenTLS struct {
```
tomponline, line 139: whats this for?

markylaing: This is what is returned by `POST /1.0/auth/identities/tls` when `{"token": true}` is received by a sufficiently privileged identity.

Now that I think about it, we can just return the `CertificateAddToken` as `POST /1.0/certificates` does. The client can base64 encoded the token before printing it.

lxd/db/cluster/identities.go
```diff
@@ -360,6 +363,41 @@ func (i *Identity) ToAPI(ctx context.Context, tx *sql.Tx, canViewGroup auth.Perm
 	}, nil
 }
 
+// UnpendTLSIdentity updates a TLS identity to make it valid by adding the fingerprint, PEM encoded certificate, and setting
+// the type to api.IdentityTypeCertificateClient.
+func UnpendTLSIdentity(ctx context.Context, tx *sql.Tx, identifier uuid.UUID, cert *x509.Certificate) error {
```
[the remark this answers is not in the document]
markylaing: Shall I rename the database method too?

shared/api/auth.go
tomponline: yeah thats better because we leave the token presentation (base64 encoding it) to the CLI

lxd/db/cluster/identities.go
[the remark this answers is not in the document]
tomponline: yes please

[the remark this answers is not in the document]
markylaing: > Shall I rename the database method too?

Sorry got confused here and thought this was the client method. I've renamed the DB method. There's still a client method called `CreateIdentityTLSPending`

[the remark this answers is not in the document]
markylaing: I've changed the client method to:
```
CreateIdentityTLSToken(identitiesTLSPost api.IdentitiesTLSPost) (*api.CertificateAddToken, error)
```

shared/api/auth.go
markylaing: Done.

shared/api/auth.go
markylaing: Done.

lxd/db/cluster/identities.go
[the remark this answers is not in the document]
markylaing: Done.

shared/api/auth.go
```diff
@@ -105,6 +105,32 @@ type IdentityPut struct {
...
+	// Whether to create a certificate add token
+	// Example: true
+	Token bool `json:"token" yaml:"token"`
+
+	// The PEM encoded and base64 encoded public certificate of the identity
+	Certificate string `json:"certificate" yaml:"certificate"`
+
+	// Groups is the list of groups for which the identity is a member.
+	// Example: ["foo", "bar"]
+	Groups []string `json:"groups" yaml:"groups"`
```
tomponline, line 131: I noticed today that the API type for groups is called AuthGroups, should we call this field AuthGroups for consistency too?

lxd/certificates.go
```diff
@@ -614,20 +614,12 @@ func certificatesPost(d *Daemon, r *http.Request) response.Response {
...
-				return response.Forbidden(fmt.Errorf("The certificate is not trusted by the CA or has been revoked"))
-			}
-		}
 	} else {
 		return response.BadRequest(fmt.Errorf("Can't use TLS data on non-TLS link"))
 	}
 
 	// Check validity.
-	err = certificateValidate(cert)
+	err = certificateValidate(d.endpoints.NetworkCert(), cert)
```
tomponline, line 631: can we avoid calling d.endpoints.NetworkCert multiple times in same function (it takes a lock you see).

lxd/db/cluster/identities.go
```diff
@@ -360,6 +363,41 @@ func (i *Identity) ToAPI(ctx context.Context, tx *sql.Tx, canViewGroup auth.Perm
...
+	metadata := CertificateMetadata{Certificate: string(pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw}))}
+	b, err := json.Marshal(metadata)
+	if err != nil {
+		return fmt.Errorf("Failed to encode certificate metadata: %w", err)
+	}
+
+	stmt := `UPDATE identities SET type = ?, identifier = ?, metadata = ? WHERE identifier = ? AND auth_method = ?`
+	res, err := tx.ExecContext(ctx, stmt, identityTypeCertificateClient, fingerprint, string(b), identifier.String(), authMethodTLS)
+	if err != nil {
+		return fmt.Errorf("Failed to unpend TLS identity: %w", err)
```
tomponline, line 384: "unpend" to "activate"  or "update"

lxd/db/cluster/identities.go
```diff
@@ -360,6 +363,41 @@ func (i *Identity) ToAPI(ctx context.Context, tx *sql.Tx, canViewGroup auth.Perm
...
+	}
+
+	n, err := res.RowsAffected()
+	if err != nil {
+		return fmt.Errorf("Failed to check affected rows after unpending a TLS identity: %w", err)
```
tomponline, line 389: Lets have a more user friendly error here, something like "Failed to activate identity, no records update"

lxd/db/cluster/identities.go
```diff
@@ -360,6 +363,41 @@ func (i *Identity) ToAPI(ctx context.Context, tx *sql.Tx, canViewGroup auth.Perm
...
+	}
+
+	if n == 0 {
+		return api.StatusErrorf(http.StatusNotFound, "No pending TLS identity found with identifier %q", identifier)
+	} else if n > 1 {
+		return fmt.Errorf("Unknown error occurred when unpending a TLS identity: %w", err)
```
tomponline, line 395: please no "unpending" :)

lxd/db/cluster/identities.go
```diff
@@ -398,6 +398,30 @@ func ActivateTLSIdentity(ctx context.Context, tx *sql.Tx, identifier uuid.UUID,
...
+func GetPendingTLSIdentityByTokenSecret(ctx context.Context, tx *sql.Tx, secret string) (*Identity, error) {
+	identities, err := getIdentitysRaw(ctx, tx, getPendingTLSIdentityByTokenSecretStmt, secret)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(identities) == 0 {
+		return nil, api.NewStatusError(http.StatusNotFound, "No pending identities found with given secret")
+	} else if len(identities) > 1 {
+		return nil, errors.New("Multiple pending identities found with given secret")
```
tomponline, line 419: Do we check for conflicts when issuing tokens to prevent this scenario occurring?

lxd/identities.go
```diff
@@ -166,6 +174,358 @@ func identityAccessHandler(authenticationMethod string, entitlement auth.Entitle
...
+
+	localHTTPSAddress := s.LocalConfig.HTTPSAddress()
+
+	trusted, err := request.GetCtxValue[bool](r.Context(), request.CtxTrusted)
+	if err != nil {
+		return response.SmartError(fmt.Errorf("Failed to get authentication status: %w", err))
+	}
+
+	var cert *x509.Certificate
+	if trusted {
```
tomponline, line 259: In the interests of starting to reduce our cyclomatic complexity:

https://ticsdocumentation.tiobe.com/latest/docs/metrics/index.html

I think it would be worth separating this logic into a separate function called something like `createIdentityTLSValidate()`?

lxd/identities.go
```diff
@@ -166,6 +174,358 @@ func identityAccessHandler(authenticationMethod string, entitlement auth.Entitle
...
+//	    oneOf:
+//	      - $ref: "#/responses/CertificateAddToken"
+//	      - $ref: "#/responses/EmptySyncResponse"
+//	  "400":
+//	    $ref: "#/responses/BadRequest"
+//	  "403":
+//	    $ref: "#/responses/Forbidden"
+//	  "500":
+//	    $ref: "#/responses/InternalServerError"
+func createTLSIdentity(d *Daemon, r *http.Request) response.Response {
```
tomponline, line 241: Should this be `createIdentityTLS` to align with the API struct IdentityTLS?

lxd/identities.go
```diff
@@ -166,6 +174,358 @@ func identityAccessHandler(authenticationMethod string, entitlement auth.Entitle
...
+	}
+
+	// Validate certificate.
+	if cert != nil {
+		err = certificateValidate(s.Endpoints.NetworkCert(), cert)
+		if err != nil {
+			return response.BadRequest(err)
+		}
+	}
+
```
tomponline, line 323: The validate function would take everything up to this point.

lxd/identities.go
```diff
@@ -166,6 +174,358 @@ func identityAccessHandler(authenticationMethod string, entitlement auth.Entitle
...
+		lc, err := notify(lifecycle.IdentityCreated, identifier.String(), false)
+		if err != nil {
+			return response.SmartError(err)
+		}
+
+		return response.SyncResponseLocation(true, token, lc.Source)
+	}
+
+	if !trusted {
```
tomponline, line 433: this section could be a separate function dealing with token handling

lxd/identities.go
```diff
@@ -166,6 +174,358 @@ func identityAccessHandler(authenticationMethod string, entitlement auth.Entitle
...
+
+		lc, err := notify(lifecycle.IdentityUpdated, identifier.String(), true)
+		if err != nil {
+			return response.SmartError(err)
+		}
+
+		return response.SyncResponseLocation(true, nil, lc.Source)
+	}
+
+	// Calculate the fingerprint.
```
tomponline, line 461: This section could be a separate function dealing with creating identity directly

lxd/identities.go
```diff
@@ -166,6 +174,358 @@ func identityAccessHandler(authenticationMethod string, entitlement auth.Entitle
...
+
+	lc, err := notify(lifecycle.IdentityCreated, fingerprint, true)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	return response.SyncResponseLocation(true, nil, lc.Source)
+}
+
+func tlsIdentityTokenValid(ctx context.Context, s *state.State, token api.CertificateAddToken) (uuid.UUID, error) {
```
tomponline, line 501: could this end "Validate" so it aligns consistent with `certificateValidate` above?

lxd/identities.go
```diff
@@ -1518,6 +1518,10 @@ func updateIdentityCache(d *Daemon) {
 		}
 
 		if cacheEntry.AuthenticationMethod == api.AuthenticationMethodTLS {
+			if cacheEntry.IdentityType == api.IdentityTypeCertificateClientPending {
```
tomponline, line 1521: To avoid mistakes in the future, can we change this to be an explicit allow list of cacheable identities?

lxc/remote.go
```diff
@@ -191,15 +191,15 @@ func (c *cmdRemoteAdd) runToken(server string, token string, rawToken *api.Certi
...
+	err = c.addRemoteFromToken(string(line), server, token, rawToken)
 	if err != nil {
 		return err
 	}
 
 	return nil
 }
 
-func (c *cmdRemoteAdd) addRemoteFromToken(addr string, server string, token string, fingerprint string) error {
+func (c *cmdRemoteAdd) addRemoteFromToken(addr string, server string, token string, rawToken *api.CertificateAddToken) error {
```
tomponline, line 202: Can we avoid passing the pointer in here? We dont modify the rawToken right?

shared/api/auth.go
markylaing: IIRC we discussed this when fine-grained auth was initially implemented and we decided to use `AuthGroup` wherever it is ambiguous and `Group` otherwise. I'm not against changing it though if you prefer.

lxd/db/cluster/identities.go
markylaing: Ah sorry I changed the function name and missed this

lxc/auth.go
```diff
@@ -756,6 +764,120 @@ func (c *cmdIdentity) command() *cobra.Command {
...
 
+type cmdIdentityCreate struct {
+	global     *cmdGlobal
+	flagGroups []string
+}
+
+func (c *cmdIdentityCreate) command() *cobra.Command {
+	cmd := &cobra.Command{}
+	cmd.Use = usage("create", i18n.G("[<remote>:]<authentication_method>/<name> [<path to PEM encoded certificate>] [[--group <group_name>]]"))
+	cmd.Short = i18n.G("Create a TLS identity")
```
tomponline, line 775: This command is specifically for creating a TLS identity and yet the usage format above seems to indicate one can provide different authentication methods?

lxd/db/cluster/identities.go
markylaing: I'm using `shared.RandomCryptoString` to generate the secret as in `POST /1.0/certificates` which returns 32 random bytes. It's unlikely that there will be a collision, but we can check and regenerate?

lxd/identities.go
markylaing: Yep that's fine, the handler is getting a bit long.

tomponline (changes requested):
Looking good, mostly stylistic suggestions to reduce cyclomatic complexity for your reviewer :)

Also @mionaalex and @minaelee is it worth booking some time with @markylaing to go through the doc changes and check they make sense structurally from an end user perspective?

lxd/db/cluster/identities.go
tomponline: Probably fine for now then, and you've protected it anyway

mionaalex (commented):
@markylaing I went through the doc changes and it seems fine to me. As far as I can tell you made sure to cover the various pages and use cases concerning TLS. I suspect we might find some other references to fix later on, but for now this looks comprehensive enough.

lxc/remote.go
markylaing: Yep I'll need to fix this up after the recent changes to `--accept-certificate` usage as well.

lxc/auth.go
markylaing: Ah I had in my mind to not rule out creation of other identity types. I'll just change to `Create an identity`.

I'll look into the stdin changes that Gabriel has worked on and do something similar.

markylaing: I've implemented creating an identity from stdin. As in the cherry-picks from incus, stdin is modified with command line args. It is assumed that if no certificate is provided (either via stdin or as an argument) that the caller wants to get a token. In that case the token field is set to true.

lxd/identities.go
```diff
@@ -1061,26 +1040,43 @@ func deleteIdentity(d *Daemon, r *http.Request) response.Response {
...
+	return response.EmptySyncResponse
+}
 
-	// Send a lifecycle event for the identity update.
-	lc := lifecycle.IdentityDeleted.Event(string(id.AuthMethod), id.Identifier, request.CreateRequestor(r), nil)
-	s.Events.SendLifecycle(api.ProjectDefaultName, lc)
+// newIdentityNotificationFunc returns a function that creates and sends a lifecycle event for the identity.
+// If updateCache is true, the local identity cache is updated and a notification is sent to other members to do the same.
+func newIdentityNotificationFunc(s *state.State, r *http.Request, networkCert *shared.CertInfo, serverCert *shared.CertInfo) identityNotificationFunc {
+	return func(action lifecycle.IdentityAction, authenticationMethod string, identifier string, updateCache bool) (*api.EventLifecycle, error) {
```
tomponline, line 1417: when wouldnt we update the cache, why is it optional?

markylaing: When a pending identity is created we emit a lifecycle event because it's something that is accessible and manageable via the API, but we don't need to update the identity cache.

tomponline: understood thanks

lxc/remote.go
```diff
@@ -260,19 +260,33 @@ func (c *cmdRemoteAdd) addRemoteFromToken(addr string, server string, token stri
...
+		} else {
+			req.Password = token
+		}
+
+		err = d.CreateCertificate(req)
+		if err != nil && !api.StatusErrorCheck(err, http.StatusConflict) {
+			return err
+		}
+	} else {
+		req := api.IdentitiesTLSPost{
```
tomponline, line 282: should be checking an API extension here right?

tomponline (approved):
Couple of nits, but LGTM!

lxc/remote.go
tomponline: Yeah and its just consistent with how we check support for api calls

Issue #13311, opened 2024-04-11
Support growing the VM disk size without needing to reboot the VM


tomponline:
This has been requested as something users would expect.

tomponline:
https://github.com/lxc/incus/pull/1158

kadinsayani:
I think we'll also need a few commits from https://github.com/lxc/incus/pull/1166.

tomponline:
Good spot

tomponline:
Can you review that PR in full and see which commits are suitable (even if not directly related to resizing) ta.

kadinsayani:
> Can you review that PR in full and see which commits are suitable (even if not directly related to resizing) ta.

Sure thing!

kadinsayani:
@tomponline I've included the commits relevant to resizing VM disks in https://github.com/canonical/lxd/pull/14211. The other two commits in https://github.com/lxc/incus/pull/1166 relate to hashing block node names if they exceed the maximum length. Currently in LXD, AFAIK, we don't perform any hashing for block node names, so commits https://github.com/lxc/incus/pull/1166/commits/371745bb5831dd000d49ad5633c0b6be4eeb86a7 and https://github.com/lxc/incus/pull/1166/commits/d08911f6f54342cf092865463f05ddb2d41514e6 from Incus may be useful for us.

PR #14211, opened 2024-10-04
Storage: Add support for online disk growing of `zfs` and `lvm` block volumes (from Incus)

kadinsayani:
This PR adds support for resizing (growing) VM disks without rebooting, when using ZFS or LVM storage backends.

Resolves https://github.com/canonical/lxd/issues/13311.

### DIFF
--- a/lxd/api_internal.go
+++ b/lxd/api_internal.go
@@ -52,6 +52,7 @@ var apiInternal = []APIEndpoint{
 	internalContainerOnStartCmd,
 	internalContainerOnStopCmd,
 	internalContainerOnStopNSCmd,
+	internalVirtualMachineOnResizeCmd,
 	internalGarbageCollectorCmd,
 	internalImageOptimizeCmd,
 	internalImageRefreshCmd,
@@ -92,6 +93,12 @@ var internalContainerOnStopCmd = APIEndpoint{
 	Path: "containers/{instanceRef}/onstop",
 
 	Get: APIEndpointAction{Handler: internalContainerOnStop, AccessHandler: allowPermission(entity.TypeServer, auth.EntitlementCanEdit)},
+}
+
+var internalVirtualMachineOnResizeCmd = APIEndpoint{
+	Path: "virtual-machines/{instanceRef}/onresize",
+
+	Get: APIEndpointAction{Handler: internalVirtualMachineOnResize, AccessHandler: allowPermission(entity.TypeServer, auth.EntitlementCanEdit)},
 }
 
 var internalSQLCmd = APIEndpoint{
@@ -388,6 +395,56 @@ func internalContainerOnStop(d *Daemon, r *http.Request) response.Response {
 	if err != nil {
 		logger.Error("The stop hook failed", logger.Ctx{"instance": inst.Name(), "err": err})
 		return response.SmartError(err)
+	}
+
+	return response.EmptySyncResponse
+}
+
+func internalVirtualMachineOnResize(d *Daemon, r *http.Request) response.Response {
+	s := d.State()
+
+	// Get the instance ID.
+	instanceID, err := strconv.Atoi(mux.Vars(r)["instanceRef"])
+	if err != nil {
+		return response.BadRequest(err)
+	}
+
+	// Get the devices list.
+	devices := request.QueryParam(r, "devices")
+	if devices == "" {
+		return response.BadRequest(fmt.Errorf("Resize hook requires a list of devices"))
+	}
+
+	// Load by ID.
+	inst, err := instance.LoadByID(s, instanceID)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	// Update the local instance.
+	for _, dev := range strings.Split(devices, ",") {
+		fields := strings.SplitN(dev, ":", 2)
+		if len(fields) != 2 {
+			return response.BadRequest(fmt.Errorf("Invalid device/size tuple: %s", dev))
+		}
+
+		size, err := strconv.ParseInt(fields[1], 10, 64)
+		if err != nil {
+			return response.BadRequest(err)
+		}
+
+		runConf := deviceConfig.RunConfig{}
+		runConf.Mounts = []deviceConfig.MountEntryItem{
+			{
+				DevName: fields[0],
+				Size:    size,
+			},
+		}
+
+		err = inst.DeviceEventHandler(&runConf)
+		if err != nil {
+			return response.InternalError(err)
+		}
 	}
 
 	return response.EmptySyncResponse
--- a/lxd/storage/drivers/driver_lvm_utils.go
+++ b/lxd/storage/drivers/driver_lvm_utils.go
@@ -590,27 +590,27 @@ func (d *lvm) copyThinpoolVolume(vol, srcVol Volume, srcSnapshots []string, refr
 	}
 
 	if volExists {
-		if refresh {
-			newVolDevPath := d.lvmDevPath(d.config["lvm.vg_name"], vol.volType, vol.contentType, vol.name)
-			tmpVolName := vol.name + tmpVolSuffix
-			tmpVolDevPath := d.lvmDevPath(d.config["lvm.vg_name"], vol.volType, vol.contentType, tmpVolName)
-
-			// Rename existing volume to temporary new name so we can revert if needed.
-			err := d.renameLogicalVolume(newVolDevPath, tmpVolDevPath)
-			if err != nil {
-				return fmt.Errorf("Error temporarily renaming original LVM logical volume: %w", err)
-			}
-
-			// Record this volume to be removed at the very end.
-			removeVols = append(removeVols, tmpVolName)
-
-			revert.Add(func() {
-				// Rename the original volume back to the original name.
-				_ = d.renameLogicalVolume(tmpVolDevPath, newVolDevPath)
-			})
-		} else {
+		if !refresh {
 			return fmt.Errorf("LVM volume already exists %q", vol.name)
 		}
+
+		newVolDevPath := d.lvmDevPath(d.config["lvm.vg_name"], vol.volType, vol.contentType, vol.name)
+		tmpVolName := vol.name + tmpVolSuffix
+		tmpVolDevPath := d.lvmDevPath(d.config["lvm.vg_name"], vol.volType, vol.contentType, tmpVolName)
+
+		// Rename existing volume to temporary new name so we can revert if needed.
+		err := d.renameLogicalVolume(newVolDevPath, tmpVolDevPath)
+		if err != nil {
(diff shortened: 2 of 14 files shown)


simondeziel:
What would prevent growing live the `.raw` file backing a QEMU on another storage driver? Or maybe that was left for another day/PR?

tomponline:
> What would prevent growing live the `.raw` file backing a QEMU on another storage driver? Or maybe that was left for another day/PR?

I'd like it if we could explore adding suppport for that, we support growing the raw disk file offline, so not sure if there is a reason we cant do it online?

tomponline:
Needs a rebase too please

kadinsayani:
@simondeziel @tomponline re: online disk resize

I don't see an issue with adding online disk resizing for ceph. RBD has an [exclusive lock feature](https://docs.ceph.com/en/reef/rbd/rbd-exclusive-locks/) and supports online resizing with RBD client kernel > 3.10.

kadinsayani:
@tomponline rebased and good to go. Do we want to include support for live resizing ceph disks with this PR or open up a separate issue and save it for later?

tomponline:
> @tomponline rebased and good to go. Do we want to include support for live resizing ceph disks with this PR or open up a separate issue and save it for later?

Lets try and do it as part of this PR. And then we can add a single API extension.

kadinsayani:
Although it's possible to online resize ceph rbd volumes, I haven't been able to get this working in LXD. I believe this is the reason why:

https://github.com/canonical/lxd/blob/main/lxd/storage/drivers/driver_ceph_volumes.go#L1332-L1334

kadinsayani:
I've tested live resizing a Ceph RBD filesystem disk and it works as expected - it's just resizing Ceph RBD block volumes that doesn't work, which explains why I haven't been able to resize a Ceph backed rootfs.

kadinsayani:
It doesn't look like we'll be able to add support for online growing of Ceph RBD root disks. Ceph backed VM's have a read only snapshot which can't be updated when the root disk size is updated (see below). The snapshot is used for instance creation.

https://github.com/canonical/lxd/blob/9ac2433510c825833f260621b00fa0c19e6a6ff8/lxd/storage/drivers/driver_ceph_volumes.go#L1332-L1337

kadinsayani:
Rebased and good to go. 

In summary, we're adding support for online resizing (growing) any zfs or lvm disks. Online resizing Ceph RBD filesystems was possible before the changes in this PR, but we've confirmed that online resizing of Ceph RBD block volumes is not possible due to the read only snapshot used during instance creation.

lxd/api_internal.go
```diff
@@ -393,6 +400,56 @@ func internalContainerOnStop(d *Daemon, r *http.Request) response.Response {
...
+	}
+
+	// Update the local instance.
+	for _, dev := range strings.Split(devices, ",") {
+		fields := strings.SplitN(dev, ":", 2)
+		if len(fields) != 2 {
+			return response.BadRequest(fmt.Errorf("Invalid device/size tuple: %s", dev))
+		}
+
+		size, err := strconv.ParseInt(fields[1], 16, 64)
```
simondeziel, line 431: Is the `size` really expressed in base 16?

lxd/storage/drivers/driver_lvm_utils.go
```diff
@@ -572,27 +590,27 @@ func (d *lvm) copyThinpoolVolume(vol, srcVol Volume, srcSnapshots []string, refr
...
-				return fmt.Errorf("Error temporarily renaming original LVM logical volume: %w", err)
-			}
+		if !refresh {
+			return fmt.Errorf("LVM volume already exists %q", vol.name)
+		}
 
-			// Record this volume to be removed at the very end.
-			removeVols = append(removeVols, tmpVolName)
+		newVolDevPath := d.lvmDevPath(d.config["lvm.vg_name"], vol.volType, vol.contentType, vol.name)
+		tmpVolName := fmt.Sprintf("%s%s", vol.name, tmpVolSuffix)
```
simondeziel, line 598:
```suggestion
		tmpVolName := vol.name + tmpVolSuffix
```

simondeziel:
> In summary, we're adding support for online resizing (growing) of any zfs or lvm disks. Online resizing Ceph RBD filesystems was possible before the changes in this PR, but we've confirmed that online resizing of Ceph RBD block volumes is not possible due to the read only snapshot used during instance creation.

zvols have a similar read-only snapshot as their origin, I guess it's an inherent limitation of how CoW is implemented in Ceph. Thanks for digging into it.

I'm now wondering what's up with `dir` and `.raw` files though. Them being `raw` files they don't have any CoW going on so I'd expect a live grow would work for them too.

simondeziel:
> > In summary, we're adding support for online resizing (growing) of any zfs or lvm disks. Online resizing Ceph RBD filesystems was possible before the changes in this PR, but we've confirmed that online resizing of Ceph RBD block volumes is not possible due to the read only snapshot used during instance creation.
> 
> zvols have a similar read-only snapshot as their origin, I guess it's an inherent limitation of how CoW is implemented in Ceph. Thanks for digging into it.

https://docs.ceph.com/en/reef/rbd/rbd-snapshot/#layering seems to suggest it should just work:
> A copy-on-write clone of a snapshot behaves exactly like any other Ceph block device image. You can read to, write from, clone, and resize cloned images. **There are no special restrictions with cloned images.**

But since you ran into issues, maybe we need to flatten those cloned images before growing them? https://docs.ceph.com/en/reef/rbd/rbd-snapshot/#flattening-a-cloned-image

MusicDin:
Should we add a row for live VM disk resize in the storage driver features table? See: https://documentation.ubuntu.com/lxd/en/latest/reference/storage_drivers/#feature-comparison

simondeziel:
> Should we add a row for live VM disk resize in the storage driver features table? See: https://documentation.ubuntu.com/lxd/en/latest/reference/storage_drivers/#feature-comparison

+1

lxd/api_internal.go
kadinsayani: Nice catch!

kadinsayani:
> > zvols have a similar read-only snapshot as their origin, I guess it's an inherent limitation of how CoW is implemented in Ceph. Thanks for digging into it.
> 
> https://docs.ceph.com/en/reef/rbd/rbd-snapshot/#layering seems to suggest it should just work:
> 
> > A copy-on-write clone of a snapshot behaves exactly like any other Ceph block device image. You can read to, write from, clone, and resize cloned images. **There are no special restrictions with cloned images.**
> 
> But since you ran into issues, maybe we need to flatten those cloned images before growing them? https://docs.ceph.com/en/reef/rbd/rbd-snapshot/#flattening-a-cloned-image

Thanks for digging into this further :)

Given my initial research, your new findings, and what I've seen in the LXD codebase, I believe it is theoretically possible to online resize (grow) Ceph RBD block volumes, dir and .raw files. 

I think I have some more work to do for this PR.

kadinsayani:
> But since you ran into issues, maybe we need to flatten those cloned images before growing them? https://docs.ceph.com/en/reef/rbd/rbd-snapshot/#flattening-a-cloned-image

I don't think flattening the cloned image is a safe approach. From the docs:

> Since a flattened image contains all the data stored in the snapshot, a flattened image takes up more storage space than a layered clone does.

kadinsayani:
So although it is possible to online grow a Ceph RBD backed root disk, I found another problem:

When we create a Ceph RBD volume, a read only snapshot is created. This read only snapshot is used as the clone source for future non-image volumes. The read only or protected property of the snapshot is a precondition for creating RBD clones.

kadinsayani:
For reference, here is the error I'm getting after modifying the behaviour to allow for online growing the root disk, and adding a file system resize:

```bash
root@testbox:~# lxc config device set v1 root size=11GiB
Error: Failed to update device "root": Could not grow underlying "ext4" filesystem for "/dev/rbd0": Failed to run: resize2fs /dev/rbd0: exit status 1 (resize2fs 1.47.0 (5-Feb-2023)
resize2fs: Bad magic number in super-block while trying to open /dev/rbd0)
```

simondeziel:
> For reference, here is the error I'm getting after modifying the behaviour to allow for online growing the root disk, and adding a file system resize:
> 
> ```shell
> root@testbox:~# lxc config device set v1 root size=11GiB
> Error: Failed to update device "root": Could not grow underlying "ext4" filesystem for "/dev/rbd0": Failed to run: resize2fs /dev/rbd0: exit status 1 (resize2fs 1.47.0 (5-Feb-2023)
> resize2fs: Bad magic number in super-block while trying to open /dev/rbd0)

`underlying "ext4"` seems misleading as it seems to be code running in the host itself as operating on `/dev/rbd0`.  Also, why would it do that? I'd expect only the VM's `/dev/sda` to be bigger, no partition touched, no FS resized.

Same for `/dev/rbd0`, shouldn't it just be bigger?

kadinsayani:
I've updated the PR and the [tests](https://github.com/canonical/lxd-ci/pull/310) are good to go. I've opened a [new issue](https://github.com/canonical/lxd/issues/14462) to track adding support for Ceph RBD volumes.

kadinsayani:
> > I've updated the PR and the [tests](https://github.com/canonical/lxd-ci/pull/310) are good to go. I've opened a [new issue](https://github.com/canonical/lxd/issues/14462) to track adding support for Ceph RBD volumes.
> 
> I don't mind (too much) having this feature land in a per-driver fashion. However, I suspect/hope that Ceph is the special case here and all our other drivers would support live growing. I didn't hear back from you regarding the easy to test `dir` backend? Next, we'll need to consider Powerflex and the other driver that's still baking.

`dir` is not supported with the changes in this PR thus far. I'm working on adding support for it :)

@tomponline mentioned that Powerflex is out of scope for this PR.

hamistao:
@kadinsayani From what I can see this may also help with container live resizing (for both growing and shrinking) on block based drivers (i.e. lvm, ceph and zfs with `volumes.zfs.block_mode` enabled), as it currently is also not possible. I am also assuming this would not apply to ceph for the same reason we apparentely can't resize VMs on it. To what extent are these assumptions correct?

kadinsayani:
> @kadinsayani From what I can see this may also help with container live resizing (for both growing and shrinking) on block based drivers (i.e. lvm, ceph and zfs with `volumes.zfs.block_mode` enabled), as it currently is also not possible. I am also assuming this would not apply to ceph for the same reason we apparentely can't resize VMs on it. To what extent are these assumptions correct?

Online shrinking is only possible for `filesystem` volumes. Online growing of block based drivers (`zfs` and `lvm`) will be possible for containers once this PR is merged. Online growing of Ceph RBD block volumes is still under investigation, see https://github.com/canonical/lxd/issues/14462.

tomponline:
@kadinsayani  can we close this for now until you get chance to look at this again?

PR #14212, opened 2024-10-04
Replace `go/ast` deprecated types

masnax:
Uses `packages.Load` and walks through the parsed files' syntax trees instead of relying on the deprecated `ast.Package` and `ast.Object` types. 

Populating type information directly into the `packages.Package` type appears to be very slow, increasing the full code generation block to about 2 minutes, so instead all of the type information loading is disabled.

See: https://pkg.go.dev/golang.org/x/tools/go/packages#Load and https://pkg.go.dev/golang.org/x/tools/go/packages#LoadMode for more information on the general pattern for defining the set of information loaded into `packages.Package`, and links to some associated issues.

The workaround I've went with is to introduce the `GetVars` function, which parses the syntax trees per file included in `packages.Package` to find all variable declarations, since we can't populate `packages.Scope` to perform `packages.Scope.Lookup` without adding significant delays. If we need them, we can add `GetTypes` and `GetFuncs` in the future as well.

With this approach, generating the full set of db files takes about 20-25 seconds, which is close to the old implementation.


masnax:
@tomponline Lint errors are fixed, though it looks like there's a dqlite dependency issue unrelated to this PR.

tomponline:
> @tomponline Lint errors are fixed, though it looks like there's a dqlite dependency issue unrelated to this PR.

needs a rebase with the fix please

tomponline:
@masnax did you test this change with `make update-schema` btw to check for any changed output?

masnax:
> @masnax did you test this change with `make update-schema` btw to check for any changed output?

There were no changes to the generated code, but it seems there was a pre-existing issue with the code generation where the `schema.go` generated files weren't working right because the path we read from `runtime.Caller` was assumed to be relative to the current directory, but it's relative to `GOPATH/src` instead.

Fixed by appending `$GOPATH` and `src` in to the returned filepath.

tomponline (approved):
thanks!

PR #14213, opened 2024-10-07
PKI test improvements

markylaing:
There was a test failure when I last pushed #14207 (https://github.com/canonical/lxd/actions/runs/11183415209/job/31092412149#step:12:64275) that led to me noticing a few things we could do better in this test:
1. When asserting that a truststore entry is added, we should grep for exactly what we expect (e.g. more than just the common name).
2. When asserting that a truststore entry has not been added, we shouldn't grep for anything (nothing to grep for, also what if a truststore entry is added with a name we don't expect). We should check that the number of entries is what is expected.
3. When in the subshell, we always should be talking to the LXD running in `LXD5_DIR`. The setting of `LXD_DIR="${LXD5_DIR}"` preceding a command was missed in some cases, meaning that we were checking the output of the wrong LXD. To fix these, `LXD5_DIR` is exported as `LXD_DIR` in the subshell.

### DIFF
--- a/test/suites/pki.sh
+++ b/test/suites/pki.sh
@@ -63,6 +63,7 @@ test_pki() {
     set -e
     # shellcheck disable=2030
     export LXD_CONF="${LXC5_DIR}"
+    export LXD_DIR="${LXD5_DIR}"
 
     ### Unrestricted CA signed client certificate with `core.trust_ca_certificates` disabled.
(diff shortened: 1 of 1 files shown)


tomponline (approved):

simondeziel:
Thank you!

PR #14228, opened 2024-10-07
Define endpoint entity type explicitly

hamistao:
This proposes a change to the endpoint classification for the API metrics to make it more mantainable.

The proposal cosists in explicitly defining each endpoint's entity type through a `MetricsType` field in the `APIEndpoint` struct instead of deriving from the URL prefix.


hamistao:
@markylaing @tomponline I think this conveys more clearly the poposal of classifying endpoints for the metrics explicitly, feel free to take a look whenever you can.

markylaing:
I think this can work well, but I would propose that the LXD daemon should fail to start up if a metrics type isn't defined for an endpoint. This will make it very obvious that it's required and means we won't accidentally fall back to `TypeServer` when we don't want to.

markylaing (dismissed):
LGTM. Much simpler and now we always know the correct entity type is set!

lxd/api.go
```diff
@@ -183,6 +184,11 @@ func restServer(d *Daemon) *http.Server {
 	}
 
 	for _, c := range api10 {
+		// Every 1.0 endpoint should have a type for the API metrics.
+		if !shared.ValueInSlice(c.MetricsType, entity.APIMetricsEntityTypes()) {
+			panic(fmt.Sprintf("Endpoint \"/1.0/%s\" has invalid MetricsType: %s", c.Path, c.MetricsType))
```
tomponline, line 189: Please can you use backticks rather than escaping the double quotes, e.g.

`panic(fmt.Sprintf(`Endpoint "/1.0/%s" has invalid MetricsType: %s`, c.Path, c.MetricsType))`

Its easier to read :)

lxd/api.go
```diff
@@ -184,6 +184,11 @@ func restServer(d *Daemon) *http.Server {
 	}
 
 	for _, c := range api10 {
+		// Every 1.0 endpoint should have a type for the API metrics.
+		if !shared.ValueInSlice(c.MetricsType, entity.APIMetricsEntityTypes()) {
```
tomponline, line 188: why call entity.APIMetricsEntityTypes() for every API endpoint in the loop? Better to store in temp var right?

tomponline (approved):

PR #14247, opened 2024-10-08
Per cluster member resource reservations

MggMuggins:
As described in [LX089](https://docs.google.com/document/d/1yRE_sKGIynydluV0UHfTqWLzP3lkTG_H5prC_wzoEfY/edit).

The general approach (as with the other limits validation) is to query the current state of the cluster from the database, apply the delta (create an instance, update the cluster config or update a cluster member config), and then validate the resulting configuration.

This is performed in at most only two round-trips to the DB:
- `GetMemberConfigWithGlobalDefault` gives a column for global and member-specific config for each cluster member
- `GetClusterMemberInstanceConfig` gives the value of each limit for each instance, including null values when the config key is not set

I don't love the specialized types for each query, but without them the results are not super manageable. I also feel like the permissions checking code could be cleaner.

I'd like to hear thoughts on organization:
- These limits aren't enforced at the project level the same way others are, but I've included them in the `lxd/project/limits` package because other code there is similar.
- The db query functions are relevant to cluster members but could fit in any of a few files

## TODO:
- [x] Refactor `GetClusterInstanceConfig` (this was the bug I noted in our standup on Tuesday Oct 8)
- [x] Only request `GET /cluster/members/{name}/state` when cluster update includes reservation keys
- [x] Reservations are not yet taken into account during instance placement
- [x] More tests
- [ ] Pass `ClusterMemberState` to scriptlet so that the scriptlet doesn't need to re-request from the rest of the cluster
- [ ] Docs? Not sure if we should call this feature out in a place other than the config keys themselves.


lxd/db/node.go
```diff
@@ -1111,6 +1111,85 @@ func whereClause(conditions []inCondition, args *[]any) string {
 	return b.String()
 }
 
+// OverridableConfig is a config key that could be set on either a specific
+// cluster member or for the whole cluster.
+type OverridableConfig struct {
```
MggMuggins, line 1116: I don't love this name; `ClusterConfigValue`? `MemberConfigValue`?

MggMuggins:
@tomponline This should be ready for a first round. I'm not clear on the cause of the test failure but initially it looks unrelated to my changes... I'll investigate further when I'm back to work on Wednesday.

lxd/api_cluster.go
```diff
@@ -1829,6 +1829,28 @@ func clusterValidateConfig(config map[string]string) error {
...
+		// lxdmeta:generate(entities=cluster; group=cluster; key=limits.reserve.cpu)
+		// Number of CPUs to reserve for the LXD server. This setting only limits
+		// the sum of instance `limits.cpu` that can be located on a cluster member.
+		//
+		// When this key is set, all instances on that member must have
+		// `limits.cpu` set.
+		// ---
+		//  type: integer
+		//  shortdesc: Number of CPUs to reserve for the LXD server
+		"limits.reserve.cpu": validate.IsAny,
```
tomponline, line 1858: shouldnt this be an optional integer?

lxd/project/limits/permissions.go
```diff
@@ -356,7 +356,7 @@ func GetImageSpaceBudget(globalConfig *clusterConfig.Config, tx *db.ClusterTx, p
 
 // Check that we would not violate the project limits or restrictions if we
 // were to commit the given instances and profiles.
-func checkRestrictionsAndAggregateLimits(globalConfig *clusterConfig.Config, tx *db.ClusterTx, info *projectInfo) error {
+func checkRestrictionsAndAggregateLimits(globalConfig *clusterConfig.Config, info *projectInfo) error {
```
tomponline, line 371: does this function or any of the functions it calls use the DB? If so, I'd prefer to keep the tx arg and plumb it down so that it uses a single tx.

lxd/cluster/member_state.go
```diff
@@ -73,6 +74,10 @@ func LocalSysInfo() (*api.ClusterMemberSysInfo, error) {
 		return nil, fmt.Errorf("Failed getting load averages: %w", err)
 	}
 
+	// NumCPU gives the number of threads available to the LXD server at startup,
+	// not the currently available number of threads.
+	sysInfo.CPUThreads = uint64(runtime.NumCPU())
```
tomponline, line 84: Any particular reason for using "Threads" here? Was it to align with a similar field elsewhere?

The NumCPU function says "NumCPU returns the number of logical CPUs usable by the current process.", so I wondered if "LogicalCPUs" would be clearer?

lxd/db/node.go
```diff
@@ -1047,6 +1047,70 @@ func (c *ClusterTx) GetNodeOfflineThreshold(ctx context.Context) (time.Duration,
 	return threshold, nil
 }
 
+func paramList(count int) string {
```
tomponline, line 1050: Can we use `lxd/db/query.Params()` here?

Also if these are generic query helper functions I think it would be useful to move them into `lxd/db/query/expr.go`.

lxd/instances_post.go
```diff
@@ -720,6 +720,11 @@ func createFromBackup(s *state.State, r *http.Request, projectName string, data
 		return response.BadRequest(err)
 	}
 
+	sysinfo, err := cluster.LocalSysInfo()
+	if err != nil {
+		return response.InternalError(err)
```
tomponline, line 725: I suggest using response.SmartError() here when your not generating the error yourself, as it could return different types of error now or in the future.

lxd/db/node.go
```diff
@@ -1111,6 +1111,85 @@ func whereClause(conditions []inCondition, args *[]any) string {
...
+// cluster member or for the whole cluster.
+type OverridableConfig struct {
+	ClusterValue sql.NullString
+	MemberValue  sql.NullString
+}
+
+// GetMemberConfigWithGlobalDefault returns a mapping of
+// [clusterMemberName][key]cfg, with filters clusterMemberNames and keys. If
+// either slice is nil/empty, all clusterMembers/keys are returned.
+func (c *ClusterTx) GetMemberConfigWithGlobalDefault(ctx context.Context, clusterMemberNames []string, keys []string) (map[string]map[string]OverridableConfig, error) {
```
tomponline, line 1124: Is this function just used for access the effective config, or are you using it to display the config to the end user?

If the former, have you considered returning a map of effective config values for each cluster member, rather than requiring the caller to consult the ClusterValue and the MemberValue.

Also we dont use sqli.NullString for config values, unset keys are just not present in the map.

lxd/db/node.go
```diff
@@ -1190,6 +1190,99 @@ func (c *ClusterTx) GetMemberConfigWithGlobalDefault(ctx context.Context, cluste
...
+type InstanceKey struct {
+	InstanceName string
+	Key          string
+	Value        sql.NullString
+}
+
+// GetClusterMemberInstanceConfig returns a map from clusterMemberName -> InstanceKey
+// with optional filters clusterMemberNames and keys. If either slice is nil/empty,
+// the results will include all cluster members/config keys.
+func (c *ClusterTx) GetClusterMemberInstanceConfig(ctx context.Context, clusterMemberNames []string, keys []string) (map[string][]InstanceKey, error) {
```
tomponline, line 1203: This doesnt look workable to me because your not expanding the effective instance config from profile(s).

I dont think you can do this in a single query, you might need to use `InstanceList()` instead with a custom handler func.

MggMuggins: Focusing on this as it's the largest functional deficiency; the current commits are a bit of a mess but my feeling is that  `InstanceList` won't quite work here. In order to add an instance (with expanded config) we need a list of all profiles that will be applied to that instance, which we can't get from `InstanceList`. Still thinking on this...

tomponline: You can use InstanceList for this, we do it commonly. It doesn't expand it automatically but does provide profile info so you can expand if needed. See existing examples

tomponline: e.g.

```
err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
		return tx.InstanceList(ctx, func(dbInst db.InstanceArgs, p api.Project) error {
                    dbInst.Config = instancetype.ExpandInstanceConfig(globalConfigDump, dbInst.Config, dbInst.Profiles)
			return nil
		}, filter)
	})
	if err != nil {
		return nil, err
	}

```

lxd/cluster/member_state.go
```diff
@@ -81,6 +86,66 @@ func LocalSysInfo() (*api.ClusterMemberSysInfo, error) {
 	return sysInfo, nil
 }
 
+// ClusterSysInfo returns a map from clusterMemberName -> sysinfo for every member
+// of the cluster. This requires an HTTP API call to the rest of the cluster.
+// Fails with ClusterUnavailableError if any requested member is offline.
+func ClusterSysInfo(s *state.State, members []db.NodeInfo) (map[string]api.ClusterMemberSysInfo, error) {
```
tomponline, line 92: Rather than duplicate the NewNotifer logic, the changes you've needed here sound quite useful to add to NewNotifier and the returned Notifier function definition anyway.

So I'm thinking we should update NewNotifier to:

```go
type Notifier func(member db.NodeInfo, hook func(lxd.InstanceServer) error) error

func NewNotifier(state *state.State, networkCert *shared.CertInfo, serverCert *shared.CertInfo, policy NotifierPolicy, members ...[]db.NodeInfo) (Notifier, error) {
```

This would make `members` optional, and if not supplied then all members would be loaded as they are today. 

Also there looks to be some optimisations we can make in NewNotifier that you've used here:

1. Avoid the call to tx.GetNodeOfflineThreshold(ctx) and s.GlobalConfig.OfflineThreshold() instead.
2. Change `if localClusterAddress == "" {` to `if s.ServerClustered {`.

That way we get to use the more resilient down detection in NewNotifier here too.

tomponline: We probably still need the ClusterSysInfo function for convenience, but it can use the updated NewNotifier under the hood.

lxd/project/limits/permissions.go
```diff
@@ -1677,24 +1677,24 @@ func AllowClusterUpdate(ctx context.Context, tx *db.ClusterTx, sysinfo map[strin
 		return nil
 	}
 
-	aggregates, err := getClusterMemberAggregateLimits(ctx, tx, newConfig, requiredLimits)
+	clusterMemberNames := []string{}
```
tomponline, line 1680: When you're appending to a slice there's no need to initialise it first, so can just do `var clusterMemberNames []string`

tomponline (changes requested):
I like that you've identified and made some cleanups that arent directly related to this feature.

But I think it would be clearer to include them in a separate earlier PR so we can lay the ground work for the feature change itself separately.

So lets include in a separate PR:

- Any cleanup not related to the feature itself.
- The NewNotifier changes.
- The SysInfo change to add logical CPUs.
- The change the cluster member config update to redirect to the member in question.

lxd/db/node.go
tomponline: im confused, can you book a quick meeting to discuss as i dont see what the issue is.

lxd/cluster/member_state.go
MggMuggins: I had considered it but decided against it while drafting this since updating all the usages was going to take a while. Agree though, I'll put together a PR.

MggMuggins:
Going to close this as the code that isn't in #14292 will get largely rewritten

PR #14249, opened 2024-10-09
github: pin `actions/*` using SHA commit IDs

simondeziel:


hamistao (approved):
Thanks for this \ (•◡•) /

PR #14252, opened 2024-10-10
lxc exec: Fix exit codes for signaled processes 

hamistao:
With this, processes spawned by `lxc exec` stopped due to being signaled now return the proper exit code to LXD.

From the [docs](https://pkg.go.dev/os#ProcessState.Sys):

> Sys returns system-dependent exit information about the process. Convert it to the appropriate underlying type, such as [syscall.WaitStatus](https://pkg.go.dev/syscall#WaitStatus) on Unix, to access its contents.

### DIFF
--- a/shared/util_linux.go
+++ b/shared/util_linux.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"strings"
+	"syscall"
 	"time"
 	"unsafe"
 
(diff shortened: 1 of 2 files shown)


shared/util_linux.go
```diff
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"strings"
+	"syscall"
```
tomponline, line 16: > NOTE: Most of the functions, types, and constants defined in this package are also available in the [golang.org/x/sys](https://pkg.go.dev/golang.org/x/sys) package. That package has more system call support than this one, and most new code should prefer that package where possible. See https://golang.org/s/go1.4-syscall for more information.

Should we use [golang.org/x/sys](https://pkg.go.dev/golang.org/x/sys) instead?

hamistao: Casting will only work if using `syscall.WaitStatus`

hamistao:
@tomponline Tests are updated as discussed, no more clean stop

tomponline:
> @tomponline Tests are updated as discussed, no more clean stop

Looks like ur still importing "syscall# through

shared/util_linux.go
tomponline: Oh i missed this comment, ok ta

tomponline (approved):

Issue #14126, opened 2024-09-18
Add all projects support to fetch images


edlerd:
# Required information

 * Distribution: snap
 * Distribution version: 6.1

# Issue description

Add a way to fetch images from all projects. Suggesting a new param: `GET 1.0/images?all-projects=true`. A similar parameter exists on the instance endpoint.

Motivation: We need all images for the permission selection in LXD-UI. We present the list of images to create a permission, and the auth API only gives us fingerprints for the images. So we'd like to fetch all images to show the nice names. This currently only works for images in the default project, others stay with the fingerprint, as we can't load the images for all projects.

PR #14260, opened 2024-10-10
API: Add support for fetching images across all projects (from Incus)

kadinsayani:
Resolves https://github.com/canonical/lxd/issues/14126.

This PR adds support for fetching images across all projects by adding a query parameter `all-projects`.

### DIFF
--- a/client/interfaces.go
+++ b/client/interfaces.go
@@ -243,6 +243,7 @@ type InstanceServer interface {
 	UpdateImageAlias(name string, alias api.ImageAliasesEntryPut, ETag string) (err error)
 	RenameImageAlias(name string, alias api.ImageAliasesEntryPost) (err error)
 	DeleteImageAlias(name string) (err error)
+	GetImagesAllProjects() (images []api.Image, err error)
 
 	// Network functions ("network" API extension)
 	GetNetworkNames() (names []string, err error)
--- a/client/interfaces.go
+++ b/client/interfaces.go
@@ -244,6 +244,7 @@ type InstanceServer interface {
 	RenameImageAlias(name string, alias api.ImageAliasesEntryPost) (err error)
 	DeleteImageAlias(name string) (err error)
 	GetImagesAllProjects() (images []api.Image, err error)
+	GetImagesAllProjectsWithFilter(filters []string) (images []api.Image, err error)
 
 	// Network functions ("network" API extension)
 	GetNetworkNames() (names []string, err error)
--- a/client/lxd_images.go
+++ b/client/lxd_images.go
@@ -51,6 +51,29 @@ func (r *ProtocolLXD) GetImagesAllProjects() ([]api.Image, error) {
 	return images, nil
 }
 
+// GetImagesAllProjectsWithFilter returns a filtered list of images across all projects as Image structs.
+func (r *ProtocolLXD) GetImagesAllProjectsWithFilter(filters []string) ([]api.Image, error) {
+	err := r.CheckExtension("api_filtering")
+	if err != nil {
+		return nil, err
+	}
+
+	images := []api.Image{}
+
+	err = r.CheckExtension("images_all_projects")
+	if err != nil {
+		return nil, err
+	}
+
+	u := api.NewURL().Path("images").WithQuery("recursion", "1").WithQuery("all-projects", "true").WithQuery("filter", parseFilters(filters))
+	_, err = r.queryStruct("GET", u.String(), nil, "", &images)
+	if err != nil {
+		return nil, err
+	}
+
+	return images, nil
+}
+
 // GetImagesWithFilter returns a filtered list of available images as Image structs.
 func (r *ProtocolLXD) GetImagesWithFilter(filters []string) ([]api.Image, error) {
 	err := r.CheckExtension("api_filtering")
--- a/client/lxd_images.go
+++ b/client/lxd_images.go
@@ -33,6 +33,24 @@ func (r *ProtocolLXD) GetImages() ([]api.Image, error) {
 	return images, nil
 }
 
+// GetImagesAllProjects returns a list of images across all projects as Image structs.
+func (r *ProtocolLXD) GetImagesAllProjects() ([]api.Image, error) {
+	images := []api.Image{}
+
+	err := r.CheckExtension("images_all_projects")
+	if err != nil {
+		return nil, err
+	}
+
+	u := api.NewURL().Path("images").WithQuery("recursion", "1").WithQuery("all-projects", "true")
+	_, err = r.queryStruct("GET", u.String(), nil, "", &images)
+	if err != nil {
+		return nil, err
+	}
+
+	return images, nil
+}
+
 // GetImagesWithFilter returns a filtered list of available images as Image structs.
 func (r *ProtocolLXD) GetImagesWithFilter(filters []string) ([]api.Image, error) {
 	err := r.CheckExtension("api_filtering")
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2519,3 +2519,7 @@ Adds a new {config:option}`instance-resource-limits:limits.cpu.pin_strategy` con
 ## `gpu_cdi`
 
 Adds support for using the Container Device Interface (CDI) specification to configure GPU passthrough in LXD containers. The `id` field of GPU devices now accepts CDI identifiers (for example, `{VENDOR_DOMAIN_NAME}/gpu=gpu{INDEX}`) for containers, in addition to DRM card IDs. This enables GPU passthrough for devices that don't use PCI addressing (like NVIDIA Tegra iGPUs) and provides a more flexible way to identify and configure GPU devices.
+
+## `images_all_projects`
+
+This adds support for listing images across all projects using the `all-projects` parameter in `GET /1.0/images` requests.
--- a/lxc/image.go
+++ b/lxc/image.go
@@ -1078,8 +1078,9 @@ type cmdImageList struct {
 	global *cmdGlobal
 	image  *cmdImage
 
-	flagFormat  string
-	flagColumns string
+	flagFormat      string
+	flagColumns     string
+	flagAllProjects bool
 }
 
 func (c *cmdImageList) command() *cobra.Command {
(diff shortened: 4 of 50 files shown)


lxd/images.go
```diff
@@ -1501,6 +1501,26 @@ func doImagesGet(ctx context.Context, tx *db.ClusterTx, recursion bool, projectN
...
+func doImagesGetAllProjects(ctx context.Context, tx *db.ClusterTx, public bool, clauses *filter.ClauseSet, hasPermission auth.PermissionChecker) ([]*api.Image, error) {
+	var result []*api.Image
+
+	projects, err := dbCluster.GetProjectNames(ctx, tx.Tx())
+	if err != nil {
+		return nil, err
+	}
+
+	for _, projectName := range projects {
+		images, err := doImagesGet(ctx, tx, true, projectName, public, clauses, hasPermission)
```
tomponline, line 1513: Why is the `recursion` parameter hardcoded to true? And yet in `imagesGet` it is passed to doImagesGet via the ` util.IsRecursionRequest(r)` call?

kadinsayani: For the images endpoint there is an option to fetch image fingerprints or image structs (based on the recursion parameter). For this new endpoint, I've hardcoded recursion to true to match the spec of fetching all image structs across all projects. I'm assuming image structs are required since the GH issue describes a requirement for needing `image.Name`. We don't want to change the behaviour of `doImagesGet()` with this change, so when the `all-projects` parameter is set to `false`, `util.IsRecursionRequest(r)` is passed into `doImagesGet()`.

tomponline: It is expected that this API endpoint behaves in the same way as other endpoints in LXD, namely that `all-projects` influences which resources are returned, and `recursion` controls whether or not they are returned as a URL list or as a struct. The `all-projects` parameter should not change the format that the resources are returned in, so hardcoding `true` here seems wrong to me.

Also ive assigned you some Incus imports related to all-projects to consider whilst you're working on this.

Finally we should ask @markylaing to advise on what authz checks we should be making to ensure we dont list access to the wrong images.

tomponline (changes requested):
The all-projects parameter should only change which resources are returned, it should not change the output format.

lxd/images.go
kadinsayani: Sounds good 👍 I’ll take a look at those Incus imports tomorrow - thanks for pointing them out!

kadinsayani: I reviewed the Incus imports. Their implementation also includes support for a `--all-projects` flag in the CLI. It appears that they’ve extended this feature to other client interactions, such as networks and profiles. I think it's a good idea to adjust this PR to incorporate Incus cherry-picks relating to fetching images across all projects, along with the appropriate LXD authentication checks, to guarantee trusted client access. After that, I'll take a look at the other PR's which expand the `--all-projects` feature to other endpoints.

tomponline: Sounds good thanks!

kadinsayani:
This PR should be good to go. It adds support for the `--all-projects` flag when running `lxc image list`. I'm planning to open separate PR's for `--all-projects` support for network zones, profiles, networks, network-acls, and storage buckets (Incus cherry picks).

lxd/images.go
```diff
@@ -1452,30 +1452,46 @@ func getImageMetadata(fname string) (*api.ImageMetadata, string, error) {
...
+	for fingerprint, projects := range imagesProjectsMap {
+		curProjectName := projects[0]
+
+		image, err := doImageGet(ctx, tx, curProjectName, fingerprint, public)
 		if err != nil {
 			continue
 		}
 
-		if !image.Public && !hasPermission(entity.ImageURL(projectName, fingerprint)) {
+		if !image.Public && !hasPermission(entity.ImageURL(curProjectName, fingerprint)) {
```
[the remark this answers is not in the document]
kadinsayani: Also, `GetImages()` only returns the image fingerprints. When we iterate over the fingerprints and projects, `doImageGet()` is called with the `public` argument passed in. Maybe I'm missing something here?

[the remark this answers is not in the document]
kadinsayani: I think I'm following what you mean now. Maybe we can rework `GetImages()` to include a `public` parameter, similar to `GetImagesFingerprints()`, like so:

https://github.com/canonical/lxd/blob/9ac2433510c825833f260621b00fa0c19e6a6ff8/lxd/db/images.go#L183-L192

client/interfaces.go
```diff
@@ -48,6 +48,7 @@ type ImageServer interface {
 
 	// Image handling functions
 	GetImages() (images []api.Image, err error)
+	GetImagesAllProjects() (images []api.Image, err error)
```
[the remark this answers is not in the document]
kadinsayani: Agreed, thanks for pointing this out :)

lxc/image.go
```diff
@@ -1355,15 +1371,22 @@ func (c *cmdImageList) run(cmd *cobra.Command, args []string) error {
 
 	serverFilters, clientFilters := getServerSupportedFilters(filters, api.Image{})
 
-	var images []api.Image
-	allImages, err := remoteServer.GetImagesWithFilter(serverFilters)
-	if err != nil {
-		allImages, err = remoteServer.GetImages()
+	var allImages, images []api.Image
+	if c.flagAllProjects {
+		allImages, err = remoteServer.GetImagesAllProjects()
```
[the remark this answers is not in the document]
kadinsayani: I think that's a great idea!

lxd/images.go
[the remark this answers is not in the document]
markylaing: I think we should remove `GetImagesFingerprints` entirely. We're using it to get a list of fingerprints and then re-querying the same table to get values from the remaining columns. Again though I think this can come in a later refactor. For now I'd be happy if getting images from all projects follows the same pattern and only returns public images when the caller is not authenticated.

lxc/image.go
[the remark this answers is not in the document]
kadinsayani: Filtering is applied in `doImagesGet()`.

lxd/images.go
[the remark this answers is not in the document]
markylaing: Just had a thought. Now that we're disallowing `all-projects` requests for untrusted users we don't need to do any of the above refactoring here, since we'll always be authenticated when getting images across all projects :)

client/lxd_images.go
```diff
@@ -33,6 +33,27 @@ func (r *ProtocolLXD) GetImages() ([]api.Image, error) {
...
+	v := url.Values{}
+	v.Set("recursion", "1")
+	v.Set("all-projects", "true")
+
+	err := r.CheckExtension("images_all_projects")
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = r.queryStruct("GET", fmt.Sprintf("/images?%s", v.Encode()), nil, "", &images)
```
markylaing, line 49: Can we use `api.NewURL` here please. It has utils for setting query parameters.

client/lxd_images.go
```diff
@@ -54,6 +54,33 @@ func (r *ProtocolLXD) GetImagesAllProjects() ([]api.Image, error) {
...
+	v.Set("recursion", "1")
+	v.Set("all-projects", "true")
+	v.Set("filter", parseFilters(filters))
+
+	err = r.CheckExtension("images_all_projects")
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = r.queryStruct("GET", fmt.Sprintf("/images?%s", v.Encode()), nil, "", &images)
```
markylaing, line 76: Same here please

markylaing (changes requested):
Just a few nits that should be easy to fix :)

client/lxd_images.go
kadinsayani: I don't see `api.NewURL` being used in the client. Do you mean `http.NewRequest`?

kadinsayani:
> Just a few nits that should be easy to fix :)

Thanks for the review @markylaing :) I think I've addressed everything now.

lxd/images.go
```diff
@@ -1729,8 +1775,14 @@ func imagesGet(d *Daemon, r *http.Request) response.Response {
...
 
-	// If the caller is not trusted, we only want to list public images.
-	publicOnly := !auth.IsTrusted(r.Context())
+	// If the caller is not trusted, we only want to list public images in the default project.
+	trusted := auth.IsTrusted(r.Context())
+	publicOnly := !trusted
+
+	// Untrusted callers can't request images from all projects or projects other than default.
+	if !trusted && (allProjects || projectName != api.ProjectDefaultName) {
+		return response.Forbidden(errors.New("Untrusted requests can only access public images in the default project"))
```
markylaing, line 1784: I'm still not sure on this error message (sorry I know I suggested it!). Suggest changing
- `can` to `may`
- `requests` to `callers` (or `users` perhaps... I'm not sure).

markylaing (changes requested):
Just a couple of last bits. Thanks :)

lxd/images.go
kadinsayani: I think `callers` is more consistent with the language used in comments throughout `lxd/images.go`.

kadinsayani:
> Just a couple of last bits. Thanks :)

Should be good to go, thanks!

kadinsayani:
Rebased and good to go 👍

markylaing (commented):
Please double check the API extension commit and cherry-pick if it was added to Incus (splitting the doc change if necessary). Other than that, LGTM. cc @tomponline.

doc/api-extensions.md
```diff
@@ -2519,3 +2519,7 @@ Adds a new {config:option}`instance-resource-limits:limits.cpu.pin_strategy` con
 ## `gpu_cdi`
 
 Adds support for using the Container Device Interface (CDI) specification to configure GPU passthrough in LXD containers. The `id` field of GPU devices now accepts CDI identifiers (for example, `{VENDOR_DOMAIN_NAME}/gpu=gpu{INDEX}`) for containers, in addition to DRM card IDs. This enables GPU passthrough for devices that don't use PCI addressing (like NVIDIA Tegra iGPUs) and provides a more flexible way to identify and configure GPU devices.
+
+## `images_all_projects`
+
+This extension adds support for listing images across all projects.
```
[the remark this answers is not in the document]
tomponline: this needs to be a cherry-pick with correct attribution

tomponline (changes requested):
There are no tests, we especially need tests to check access control of images.

kadinsayani:
> There are no tests, we especially need tests to check access control of images.

Tests added! 😄

tomponline (approved):
thanks!

PR #14261, opened 2024-10-11
Auth: Prune pending TLS identities

markylaing:
As part of the [TLS fine-grained authorization specification](https://discourse.ubuntu.com/t/tls-fine-grained-authorization/48497), pending TLS identities must be pruned when their associated token expires. This pull request adds logic to the `autoRemoveExpiredTokens` task such that the leader will retrieve all the pending TLS identities and delete those that have expired. Non-leaders will continue to cancel "certificate add" and "cluster join" operations in-memory in each member.

There was a lot of duplication of logic to identify the database leader. This PR adds a method to `Daemon` and `State` to determine if the member is a leader and refactors existing code to use that method. If the Daemon is standalone it is considered to be the leader. It is the responsibility of the caller to check `(*State).ServerClustered` before calling `(*State).LeaderInfo` when necessary.

Marked as draft until #14207 is merged.

### DIFF
--- a/lxd/api_cluster.go
+++ b/lxd/api_cluster.go
@@ -1226,9 +1226,13 @@ func clusterNodesGet(d *Daemon, r *http.Request) response.Response {
 	recursion := util.IsRecursionRequest(r)
 	s := d.State()
 
-	leaderAddress, err := d.gateway.LeaderAddress()
+	leaderInfo, err := s.LeaderInfo()
 	if err != nil {
 		return response.InternalError(err)
+	}
+
+	if !leaderInfo.Clustered {
+		return response.InternalError(cluster.ErrNodeIsNotClustered)
 	}
 
 	var raftNodes []db.RaftNode
@@ -1268,7 +1272,7 @@ func clusterNodesGet(d *Daemon, r *http.Request) response.Response {
 		}
 
 		args := db.NodeInfoArgs{
-			LeaderAddress:        leaderAddress,
+			LeaderAddress:        leaderInfo.Address,
 			FailureDomains:       failureDomains,
 			MemberFailureDomains: memberFailureDomains,
 			OfflineThreshold:     s.GlobalConfig.OfflineThreshold(),
@@ -1503,9 +1507,13 @@ func clusterNodeGet(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
-	leaderAddress, err := d.gateway.LeaderAddress()
+	leaderInfo, err := s.LeaderInfo()
 	if err != nil {
 		return response.InternalError(err)
+	}
+
+	if !leaderInfo.Clustered {
+		return response.InternalError(cluster.ErrNodeIsNotClustered)
 	}
 
 	var raftNodes []db.RaftNode
@@ -1544,7 +1552,7 @@ func clusterNodeGet(d *Daemon, r *http.Request) response.Response {
 		}
 
 		args := db.NodeInfoArgs{
-			LeaderAddress:        leaderAddress,
+			LeaderAddress:        leaderInfo.Address,
 			FailureDomains:       failureDomains,
 			MemberFailureDomains: memberFailureDomains,
 			OfflineThreshold:     s.GlobalConfig.OfflineThreshold(),
@@ -1981,24 +1989,27 @@ func clusterNodeDelete(d *Daemon, r *http.Request) response.Response {
 	}
 
 	// Redirect all requests to the leader, which is the one with
-	// knowing what nodes are part of the raft cluster.
+	// knowledge of which nodes are part of the raft cluster.
 	localClusterAddress := s.LocalConfig.ClusterAddress()
-
-	leader, err := d.gateway.LeaderAddress()
+	leaderInfo, err := s.LeaderInfo()
 	if err != nil {
-		return response.InternalError(err)
+		return response.SmartError(err)
 	}
 
-	var localInfo, leaderInfo db.NodeInfo
+	if !leaderInfo.Clustered {
+		return response.InternalError(cluster.ErrNodeIsNotClustered)
+	}
+
+	var localInfo, leaderNodeInfo db.NodeInfo
 	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
 		localInfo, err = tx.GetNodeByAddress(ctx, localClusterAddress)
 		if err != nil {
 			return fmt.Errorf("Failed loading local member info %q: %w", localClusterAddress, err)
 		}
 
-		leaderInfo, err = tx.GetNodeByAddress(ctx, leader)
+		leaderNodeInfo, err = tx.GetNodeByAddress(ctx, leaderInfo.Address)
 		if err != nil {
-			return fmt.Errorf("Failed loading leader member info %q: %w", leader, err)
+			return fmt.Errorf("Failed loading leader member info %q: %w", leaderInfo.Address, err)
 		}
 
 		return nil
@@ -2018,7 +2029,7 @@ func clusterNodeDelete(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(fmt.Errorf("Unable to get raft nodes: %w", err))
 	}
 
-	if localClusterAddress != leader {
+	if !leaderInfo.Leader {
 		if localInfo.Name == name {
 			// If the member being removed is ourselves and we are not the leader, then lock the
 			// clusterPutDisableMu before we forward the request to the leader, so that when the leader
@@ -2035,8 +2046,8 @@ func clusterNodeDelete(d *Daemon, r *http.Request) response.Response {
 			}()
 		}
 
-		logger.Debugf("Redirect member delete request to %s", leader)
-		client, err := cluster.Connect(leader, s.Endpoints.NetworkCert(), s.ServerCert(), r, false)
+		logger.Debugf("Redirect member delete request to %s", leaderInfo.Address)
+		client, err := cluster.Connect(leaderInfo.Address, s.Endpoints.NetworkCert(), s.ServerCert(), r, false)
 		if err != nil {
 			return response.SmartError(err)
 		}
@@ -2048,7 +2059,7 @@ func clusterNodeDelete(d *Daemon, r *http.Request) response.Response {
 
 		// If we are the only remaining node, wait until promotion to leader,
 		// then update cluster certs.
-		if name == leaderInfo.Name && len(nodes) == 2 {
+		if name == leaderNodeInfo.Name && len(nodes) == 2 {
 			err = d.gateway.WaitLeadership()
 			if err != nil {
 				return response.SmartError(err)
@@ -2080,9 +2091,9 @@ func clusterNodeDelete(d *Daemon, r *http.Request) response.Response {
 	defer d.clusterMembershipMutex.Unlock()
 
 	// If we are removing the leader of a 2 node cluster, ensure the other node can be a leader.
-	if name == leaderInfo.Name && len(nodes) == 2 {
+	if name == leaderNodeInfo.Name && len(nodes) == 2 {
 		for i := range nodes {
-			if nodes[i].Address != leader && nodes[i].Role != db.RaftVoter {
+			if nodes[i].Address != leaderInfo.Address && nodes[i].Role != db.RaftVoter {
 				// Promote the remaining node.
 				nodes[i].Role = db.RaftVoter
 				err := changeMemberRole(s, r, nodes[i].Address, nodes)
@@ -2426,25 +2437,23 @@ func internalClusterPostAccept(d *Daemon, r *http.Request) response.Response {
 	}
 
 	// Redirect all requests to the leader, which is the one with
-	// knowning what nodes are part of the raft cluster.
-	localClusterAddress := s.LocalConfig.ClusterAddress()
-
-	leader, err := d.gateway.LeaderAddress()
+	// knowledge of which nodes are part of the raft cluster.
+	leaderInfo, err := s.LeaderInfo()
 	if err != nil {
 		return response.InternalError(err)
 	}
 
-	if localClusterAddress != leader {
-		logger.Debugf("Redirect member accept request to %s", leader)
+	if !leaderInfo.Clustered {
+		return response.InternalError(cluster.ErrNodeIsNotClustered)
+	}
 
-		if leader == "" {
-			return response.SmartError(fmt.Errorf("Unable to find leader address"))
-		}
+	if !leaderInfo.Leader {
+		logger.Debugf("Redirect member accept request to %s", leaderInfo.Address)
 
 		url := &url.URL{
 			Scheme: "https",
 			Path:   "/internal/cluster/accept",
-			Host:   leader,
+			Host:   leaderInfo.Address,
 		}
 
 		return response.SyncResponseRedirect(url.String())
@@ -2517,19 +2526,21 @@ func internalClusterPostRebalance(d *Daemon, r *http.Request) response.Response 
 
 	// Redirect all requests to the leader, which is the one with with
 	// up-to-date knowledge of what nodes are part of the raft cluster.
-	localClusterAddress := s.LocalConfig.ClusterAddress()
-
-	leader, err := d.gateway.LeaderAddress()
+	leaderInfo, err := s.LeaderInfo()
 	if err != nil {
 		return response.InternalError(err)
 	}
 
-	if localClusterAddress != leader {
-		logger.Debugf("Redirect cluster rebalance request to %s", leader)
+	if !leaderInfo.Clustered {
+		return response.InternalError(cluster.ErrNodeIsNotClustered)
+	}
+
+	if !leaderInfo.Leader {
+		logger.Debugf("Redirect cluster rebalance request to %s", leaderInfo.Address)
 		url := &url.URL{
 			Scheme: "https",
 			Path:   "/internal/cluster/rebalance",
-			Host:   leader,
+			Host:   leaderInfo.Address,
 		}
 
 		return response.SyncResponseRedirect(url.String())
@@ -2662,16 +2673,12 @@ func handoverMemberRole(s *state.State, gateway *cluster.Gateway) error {
 
 	// Find the cluster leader.
 findLeader:
-	leader, err := gateway.LeaderAddress()
+	leaderInfo, err := s.LeaderInfo()
 	if err != nil {
 		return err
 	}
 
-	if leader == "" {
-		return fmt.Errorf("No leader address found")
-	}
-
-	if leader == localClusterAddress {
+	if leaderInfo.Leader {
 		logger.Info("Transferring leadership", logCtx)
 		err := gateway.TransferLeadership()
 		if err != nil {
@@ -2682,7 +2689,7 @@ findLeader:
 	}
 
 	logger.Info("Handing over cluster member role", logCtx)
-	client, err := cluster.Connect(leader, s.Endpoints.NetworkCert(), s.ServerCert(), nil, true)
+	client, err := cluster.Connect(leaderInfo.Address, s.Endpoints.NetworkCert(), s.ServerCert(), nil, true)
 	if err != nil {
 		return fmt.Errorf("Failed handing over cluster member role: %w", err)
 	}
@@ -2752,21 +2759,17 @@ func internalClusterPostHandover(d *Daemon, r *http.Request) response.Response {
 	// authoritative knowledge of the current raft configuration.
 	localClusterAddress := s.LocalConfig.ClusterAddress()
 
-	leader, err := d.gateway.LeaderAddress()
+	leaderInfo, err := s.LeaderInfo()
 	if err != nil {
 		return response.InternalError(err)
 	}
 
-	if leader == "" {
-		return response.SmartError(fmt.Errorf("No leader address found"))
-	}
-
-	if localClusterAddress != leader {
-		logger.Debugf("Redirect handover request to %s", leader)
+	if !leaderInfo.Leader {
+		logger.Debugf("Redirect handover request to %s", leaderInfo.Address)
 		url := &url.URL{
 			Scheme: "https",
 			Path:   "/internal/cluster/handover",
-			Host:   leader,
+			Host:   leaderInfo.Address,
 		}
 
 		return response.SyncResponseRedirect(url.String())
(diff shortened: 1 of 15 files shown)


lxd/daemon.go
```diff
@@ -604,6 +604,7 @@ func (d *Daemon) State() *state.State {
 		LocalConfig:         localConfig,
 		ServerName:          d.serverName,
 		ServerClustered:     d.serverClustered,
+		LeaderInfo:          func() (bool, string, error) { return d.leaderInfo() },
```
tomponline, line 607: Cant we just use `LeaderInfo: d.leaderInfo` why another wrapping function?

lxd/instance_post.go
```diff
@@ -232,7 +232,7 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
 		}
 
 		if targetMemberInfo == nil && s.GlobalConfig.InstancesPlacementScriptlet() != "" {
-			leaderAddress, err := d.gateway.LeaderAddress()
```
tomponline, line 235: interesting, could this previously have returned an error if not clustered?

lxd/tokens.go
```diff
@@ -28,7 +31,44 @@ func autoRemoveExpiredTokens(ctx context.Context, s *state.State) {
...
-	if len(expiredTokenOps) == 0 {
+	isLeader, _, err := s.LeaderInfo()
+	if err != nil {
+		logger.Error("Failed to get database leader details", logger.Ctx{"err": err})
+	}
+
+	var expiredPendingTLSIdentities []cluster.Identity
+	if isLeader {
+		var pendingTLSIdentities []cluster.Identity
+		err := s.DB.Cluster.Transaction(ctx, func(ctx context.Context, tx *db.ClusterTx) error {
```
tomponline, line 49: Please can we move this part into a separate function called `getExpiredPendingIdentities` or similar?

test/suites/auth.sh
```diff
@@ -99,6 +99,11 @@ test_authorization() {
   LXD_CONF="${LXD_CONF3}" gen_cert_and_key "client"
   ! LXD_CONF="${LXD_CONF3}" lxc remote add tls "${tls_identity_token2}" || false
 
+  # Check token prune task works
+  [ "$(lxc auth identity list --format csv | grep -cF 'pending')" = 1 ]
```
tomponline, line 103: Please can you also add a test for checking the classic tokens are getting removed too now we have the /internal endpoint to trigger it?

lxd/daemon.go
```diff
@@ -2416,3 +2416,23 @@ func (d *Daemon) nodeRefreshTask(heartbeatData *cluster.APIHeartbeat, isLeader b
 
 	wg.Wait()
 }
+
+// leaderInfo returns whether this Daemon is the leader and the leader address, or an error if the leader cannot be found.
+// If the Daemon is not clustered then it is considered the leader and no address is returned.
+func (d *Daemon) leaderInfo() (bool, string, error) {
```
tomponline, line 2423: In this PR I've noticed that by masking the error from d.gateway.LeaderAddress() when not clustered has required adding several explicit checks for s.ServerClustered in the code base.

I'm wondering if instead this function should return a new struct type called, say, `LeaderInfo` that contains the fields:

- Clustered bool
- Leader bool (set to true if not clustered, or clustered and is leader)
- LeaderAddress (set to empty if not clustered, or leader address if clustered)

Then we can keep all the leader/clustered related logic together.

Furthermore, when this function is stored in State.LeaderInfo, its still going to access the config from d.localConfig, so I'm wondering if we should avoid having this function on Daemon at all, and instead, store a non-exported reference to d.gateway.LeaderAddress inside State, and then have State.LeaderInfo a first-class function that accesses the existing ServerClustered, and LocalConfig.

lxd/instance_post.go
hamistao: Apparentely yes: https://github.com/hamistao/lxd/blob/b98fa327315345e7927e65035e2cd8c7e63a8b5d/lxd/cluster/gateway.go#L618

hamistao: Shouldn't it stay that way though? If I understand this correctly, it seems pointless to excecute this bit in a standalone environment.

lxd/patches.go
```diff
@@ -413,27 +411,13 @@ func patchDBNodesAutoInc(name string, d *Daemon) error {
...
-			if errors.Is(err, cluster.ErrNodeIsNotClustered) {
-				break // Apply change on standalone node.
-			}
-
 			return err
 		}
 
-		if localConfig.ClusterAddress() == leaderAddress {
-			break // Apply change on leader node.
+		if !isLeader {
```
hamistao, line 419:
```suggestion
		if isLeader {
```

Could this perhaps be the correct form?

lxd/tokens.go
```diff
@@ -40,6 +80,20 @@ func autoRemoveExpiredTokens(ctx context.Context, s *state.State) {
...
+				if err != nil {
+					logger.Debug("Failed removing pending TLS identity", logger.Ctx{"err": err, "id": op.ID()})
+				}
+			}
+
+			return nil
+		})
+		if err != nil {
+			logger.Debug("Failed removing pending TLS identities", logger.Ctx{"err": err, "id": op.ID()})
+		}
```
hamistao, line 102: I am not too familiar with out logging standards, but I am wondering if this should be a `logger.Error`, to match the ones above on this same file. What do you think?

lxd/instance_post.go
hamistao: Oh sorry, this entire section is gated with a `if target != "" && s.ServerClustered {`, so my comments here can be disregarded.

lxd/daemon.go
```diff
@@ -2416,3 +2416,23 @@ func (d *Daemon) nodeRefreshTask(heartbeatData *cluster.APIHeartbeat, isLeader b
...
+
+	localClusterAddress := d.localConfig.ClusterAddress()
+	leaderAddress, err := d.gateway.LeaderAddress()
+	if err != nil {
+		return false, "", fmt.Errorf("Failed to get the address of the cluster leader: %w", err)
+	}
+
+	if leaderAddress == "" {
+		return false, "", fmt.Errorf("No leader address found")
+	}
```
hamistao, line 2436: Apparentely `d.gateway.LeaderAddress` already returns an error when the address is empty, so maybe this bit is redundant.

lxd/instance_post.go
hamistao: Oh sorry, this entire section is gated with a if target != "" && s.ServerClustered {, so my comments here can be disregarded.

lxd/daemon.go
markylaing: This all makes sense except for storing a non-exported reference to `d.gateway.LeaderAddress` inside `State`. We need to export it in order to be able to set it in the `(*Daemon).State` function.

That means we'll have both `State.LeaderAddress` and `State.LeaderInfo`, which is confusing to the caller. I'll play around with it to see if I can find something that works well.

tomponline: Why? `(*Daemon).State()` returns state.State, that doesn't have to have an exported `LeaderAddress` field, it can store that value as part of State.LeaderInfo no?

markylaing: Perhaps I'm misunderstanding what you're asking for. If I add a first class function to State to get the leader info as follows:

```go
type LeaderInfo struct {
    Clustered bool
    Leader bool
    Address string
}

func (s *State) LeaderInfo() (*LeaderInfo, error) {
	if !s.ServerClustered {
		return &state.LeaderInfo{
			Clustered: false,
			Leader:    true,
			Address:   "",
		}, nil
	}

	localClusterAddress := s.LocalConfig.ClusterAddress()
	leaderAddress, err := leaderAddressFunc()
	if err != nil {
		return nil, fmt.Errorf("Failed to get the address of the cluster leader: %w", err)
	}

	if leaderAddress == "" {
		return nil, fmt.Errorf("No leader address found")
	}

	return &state.LeaderInfo{
		Clustered: true,
		Leader:    localClusterAddress == leaderAddress,
		Address:   leaderAddress,
	}, nil
}
```

The question is where does the reference to `leaderAddressFunc` come from? I need to store it in state somewhere, but I can't set it in state via `(*Daemon).State` unless the field is exported.

I've made the changes that you've asked for except that `LeaderInfo` is not a first class function. It's being set in `(*Daemon).State`.

markylaing: Oh my bad. I did completely forget what a first class function was apparently. I've implemented it as you've asked :laughing:

lxd/patches.go
markylaing: Yep that one was wrong! Thanks :)

lxd/tokens.go
markylaing: It looks like the log messages that relate to collating the expired token operations are logged at error level and the log messages that related to deleting the expired token operations are logged at debug level. So this is consistent with what is already in the task. 

I'm happy to change it though :)

lxd/daemon.go
markylaing: Done

lxd/tokens.go
markylaing: Done

markylaing:
@tomponline @hamistao third time lucky, it has passed! Unfortunately this is probably bad news as it means we have a racy test that fails intermittently in CI. I'll keep looking but I can't see how my PR is affecting this.

markylaing:
> @tomponline @hamistao third time lucky, it has passed! Unfortunately this is probably bad news as it means we have a racy test that fails intermittently in CI. I'll keep looking but I can't see how my PR is affecting this.

I just realised that I posted this comment on the wrong PR. It's meant for #14324

lxd/api_cluster.go
```diff
@@ -1226,7 +1226,11 @@ func clusterNodesGet(d *Daemon, r *http.Request) response.Response {
 	recursion := util.IsRecursionRequest(r)
 	s := d.State()
 
-	leaderAddress, err := d.gateway.LeaderAddress()
+	if !s.ServerClustered {
```
tomponline, line 1229: For consistency shouldn't we be checking leaderInfo.Clustered here?

lxd/api_cluster.go
```diff
@@ -1498,12 +1502,16 @@ func clusterNodesPost(d *Daemon, r *http.Request) response.Response {
 func clusterNodeGet(d *Daemon, r *http.Request) response.Response {
 	s := d.State()
 
+	if !s.ServerClustered {
```
tomponline, line 1505: shouldnt we be checking leaderInfo.Clustered here?

lxd/api_cluster.go
```diff
@@ -1970,6 +1982,10 @@ func clusterNodePost(d *Daemon, r *http.Request) response.Response {
 func clusterNodeDelete(d *Daemon, r *http.Request) response.Response {
 	s := d.State()
 
+	if !s.ServerClustered {
```
tomponline, line 1985: shouldnt we be checking leaderInfo.Clustered here?

lxd/api_cluster.go
```diff
@@ -2412,6 +2427,10 @@ func updateClusterCertificate(ctx context.Context, s *state.State, gateway *clus
 func internalClusterPostAccept(d *Daemon, r *http.Request) response.Response {
 	s := d.State()
 
+	if !s.ServerClustered {
```
tomponline, line 2430: shouldnt we be using leaderInfo.Clustered here?

lxd/api_cluster.go
```diff
@@ -2515,21 +2528,23 @@ type internalRaftNode struct {
 func internalClusterPostRebalance(d *Daemon, r *http.Request) response.Response {
 	s := d.State()
 
+	if !s.ServerClustered {
```
tomponline, line 2531: leaderInfo.Clustered?

lxd/api_cluster.go
```diff
@@ -4462,17 +4469,17 @@ func autoHealClusterTask(d *Daemon) (task.Func, task.Schedule) {
 			return // Skip healing if it's disabled.
 		}
 
-		leader, err := d.gateway.LeaderAddress()
-		if err != nil {
-			if errors.Is(err, cluster.ErrNodeIsNotClustered) {
-				return // Skip healing if not clustered.
-			}
+		if !s.ServerClustered {
```
tomponline, line 4472: leaderInfo.Clustered

lxd/tokens.go
```diff
@@ -28,7 +38,20 @@ func autoRemoveExpiredTokens(ctx context.Context, s *state.State) {
 		}
 	}
 
-	if len(expiredTokenOps) == 0 {
+	leaderInfo, err := s.LeaderInfo()
+	if err != nil {
+		logger.Warn("Failed to get database leader details", logger.Ctx{"err": err})
```
tomponline, line 43: why do we continue here?

lxd/tokens.go
```diff
@@ -28,7 +38,20 @@ func autoRemoveExpiredTokens(ctx context.Context, s *state.State) {
...
+	}
+
+	var expiredPendingTLSIdentities []cluster.Identity
+	if leaderInfo != nil && leaderInfo.Leader {
+		expiredPendingTLSIdentities, err = getExpiredPendingIdentities(ctx, s)
+		if err != nil {
+			logger.Warn("Failed to retrieve expired pending TLS identities during removal of expired tokens task", logger.Ctx{"err": err})
```
tomponline, line 50: why do we continue here?

lxd/tokens.go
```diff
@@ -40,6 +63,20 @@ func autoRemoveExpiredTokens(ctx context.Context, s *state.State) {
...
+				err := cluster.DeleteIdentity(ctx, tx.Tx(), api.AuthenticationMethodTLS, expiredPendingTLSIdentity.Identifier)
+				if err != nil {
+					logger.Warn("Failed removing pending TLS identity", logger.Ctx{"err": err, "id": op.ID()})
+				}
+			}
+
+			return nil
+		})
+		if err != nil {
+			logger.Warn("Failed removing pending TLS identities", logger.Ctx{"err": err, "id": op.ID()})
```
tomponline, line 77: can we change "id" in this line and the one above (and any others you come across) to "operation", to match that of `ImageDownload()` as I feel `id` is a bit vague here.

lxd/tokens.go
```diff
@@ -66,6 +103,40 @@ func autoRemoveExpiredTokens(ctx context.Context, s *state.State) {
...
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	expiredPendingTLSIdentities := make([]cluster.Identity, 0, len(pendingTLSIdentities))
+	for _, pendingTLSIdentity := range pendingTLSIdentities {
+		metadata, err := pendingTLSIdentity.PendingTLSMetadata()
+		if err != nil {
```
tomponline, line 125: Can this be written without the duplicated append doing something like this (i.e in this case the happy path is an expired token)?

```go
metadata, err := pendingTLSIdentity.PendingTLSMetadata()

if metadata != nil &&  (metadata.Expiry.IsZero() || metadata.Expiry.After(time.Now())) {
    continue // Token has not expired.
} 

if err != {
    // If we fail to unmarshal it, it's not valid and therefore should be removed.
    logger.Warn("Failed to unmarshal pending TLS identity metadata", logger.Ctx{"err": err})
}

    expiredPendingTLSIdentities = append(expiredPendingTLSIdentities, pendingTLSIdentity)
}
```

Also slightly concerned about using `err != nil` as a reason to remove it, as looking at `PendingTLSMetadata` it can return an error for one other reason than invalid JSON.

lxd/identities.go
```diff
@@ -516,6 +517,18 @@ func tlsIdentityTokenValidate(ctx context.Context, s *state.State, token api.Cer
...
 
+	reverter := revert.New()
+	defer reverter.Fail()
+
+	reverter.Add(func() {
+		err := s.DB.Cluster.Transaction(s.ShutdownCtx, func(ctx context.Context, tx *db.ClusterTx) error {
+			return dbCluster.DeleteIdentity(ctx, tx.Tx(), id.AuthMethod, id.Identifier)
+		})
+		if err != nil {
+			logger.Warn("Failed to delete invalid or expired pending TLS identity", logger.Ctx{"err": err, "identity_id": id.Identifier})
```
tomponline, line 528: can we just use `"identity":` in the log context?

Have you logged an identifier ID previously, what context label did you use?

test/suites/auth.sh
```diff
@@ -102,6 +102,13 @@ test_authorization() {
   # The token was used, so the pending identity should be deleted.
   [ "$(lxc auth identity list --format csv | grep -cF 'pending')" = 0 ]
 
+  # Check token prune task works
+  lxc auth identity create tls/test-user2 --quiet
+  sleep 2
```
tomponline, line 107: why sleep 2?

markylaing:
> Please can you review the use of s.ServerClustered and in functions where there is also a call to s.LeaderInfo check leaderInfo.Clustered instead, that way we are keeping checks for cluster state to the leaderInfo struct. Perhaps ultimately we may unexport s.Clustered from State....

I have kept these checks because checking `s.Clustered` is quick, whereas `s.LeaderInfo` may call `d.gateway.LeaderAddress` which may be slow. I think it's fine to keep `s.Clustered` because sometimes we just need to know if we're clustered, and not necessary get the leader information.

tomponline:
> > Please can you review the use of s.ServerClustered and in functions where there is also a call to s.LeaderInfo check leaderInfo.Clustered instead, that way we are keeping checks for cluster state to the leaderInfo struct. Perhaps ultimately we may unexport s.Clustered from State....
> 
> I have kept these checks because checking `s.Clustered` is quick, whereas `s.LeaderInfo` may call `d.gateway.LeaderAddress` which may be slow. I think it's fine to keep `s.Clustered` because sometimes we just need to know if we're clustered, and not necessary get the leader information.

Yeah but wont we do an s.LeaderInfo call if we are clustered in those situations, and s.LeaderInfo is cheap because it tests s.ServerClustered internally first?

test/suites/auth.sh
markylaing: To wait for the token to expire. The expiry is still set to 1 second. I'll add a comment.

lxd/identities.go
hamistao: I did not find any other occurences of logging an identifier ID, so just using `identity` should be fine.

lxd/tokens.go
hamistao:
```
// PendingTLSMetadata returns the pending TLS identity metadata.
func (i Identity) PendingTLSMetadata() (*PendingTLSMetadata, error) {
	if i.Type != api.IdentityTypeCertificateClientPending {
		return nil, fmt.Errorf("Cannot extract pending TLS identity secret: Identity is not pending")
	}

	var metadata PendingTLSMetadata
	err := json.Unmarshal([]byte(i.Metadata), &metadata)
	if err != nil {
		return nil, fmt.Errorf("Failed to unmarshal pending TLS identity metadata: %w", err)
	}

	return &metadata, nil
}
```

The one other reason is if identity is not pending, but since we are only querying for pending identities it should not raise any immediate problems. That said, I think it would be nice to check for a specific error here as to not rely so much on another function's specific implementation.

lxd/tokens.go
markylaing: So that any local certificate add/cluster join token operations are still pruned. I'll add a comment.

lxd/tokens.go
markylaing: I'm confused about this. The only identities that are returned by the query are pending TLS identities. If we can't get their metadata then, regardless of the reason, they are invalid since they cannot be used e.g. when an untrusted used tries to activate it with a token. We should never enter this block but the warning is their in case we do.

I can add a new error type e.g. `ErrInvalidIdentityType` or similar to check if the error was returned by the other stanza but in this case we'd still want to delete it

tomponline: OK as long as an error from this function always means delete the identity token.

But in general we should avoid assuming that all errors from a function mean a single thing, as that can change over time and introduce logic bugs.

It would generally be preferable to be able to identify the specific error class that should lead to the deletion, and anything unexpected is then logged.

lxd/tokens.go
markylaing: So that any local certificate add/cluster join token operations are still pruned. I've added a comment.

markylaing:
I think I've addressed everything. Let me know if you have any more thoughts on https://github.com/canonical/lxd/pull/14261#discussion_r1814452259 but otherwise ready for another round.

lxd/tokens.go
markylaing: It always means delete the token as long as we're sure that the identity is a pending identity. The function itself is defined on the general identity... I'm beginning to understand the problem a bit better now.

I think what we can do in this case if have an error type defined in the DB:
```go
type InvalidIdentityMetadataError struct {
    err error
}
```
which is only returned if the JSON unmarshalling fails, but not if e.g. `CertificateMetadata()` is called on an OIDC identity. We should return this error when getting metadata for any identity type. Getting this error will always mean the identity is invalid and must be deleted.

tomponline: I suppose another way of looking at this is that `PendingTLSMetadata()` could just not return an error ever, and if it returns nil metadata (because its the incorrect identity type or invalid metadata) then we can then treat it the same way.

markylaing: Yeah I'm happy to go with that, though I'd like to do the same for all identity types. Can I update this in a separate PR?

markylaing: Though I suppose this might mask the underlying error and make it harder to reason about why the unmarshalling failed. We can add a warning log for it though

tomponline: hrm in that case dont warn in the function but return an error then.

But im confused, because we werent logging the error anyway, just skipping over it.

markylaing: I have updated `PendingTLSMetadata` to return a 400 Bad Request error if it is called against an identity that is not a pending TLS identity, and a 500 Internal Server Error if JSON unmarshalling fails. I've added a comment to the token prune task to explain why we want to remove the pending identity in either case.

lxd/tokens.go
```diff
@@ -55,7 +62,7 @@ func autoRemoveExpiredTokens(ctx context.Context, s *state.State) {
 		for _, op := range expiredTokenOps {
 			_, err := op.Cancel()
 			if err != nil {
-				logger.Warn("Failed removing expired token", logger.Ctx{"err": err, "operation": op.ID()})
+				logger.Debug("Failed removing expired token", logger.Ctx{"err": err, "operation": op.ID()})
```
tomponline, line 65: why not Warn?

tomponline (approved):
LGTM, just that Debug log line query?

lxd/api_cluster.go
```diff
@@ -1981,24 +1989,27 @@ func clusterNodeDelete(d *Daemon, r *http.Request) response.Response {
...
-
-	leader, err := d.gateway.LeaderAddress()
+	leaderInfo, err := s.LeaderInfo()
 	if err != nil {
-		return response.InternalError(err)
+		return response.SmartError(err)
+	}
+
+	if !leaderInfo.Clustered {
+		return response.InternalError(cluster.ErrNodeIsNotClustered)
```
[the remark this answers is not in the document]
markylaing: I thought that but I didn't want to change the existing HTTP code as that's technically an API change

lxd/tokens.go
markylaing: Hmm this is just wrong, it shouldn't have been changed in this commit. Will fix.

markylaing: Fixed.

hamistao (approved):
LGTM!

PR #14263, opened 2024-10-11
test/container_devices_nic_bridged_filtering: unload br_netfilter once done

simondeziel:
Also avoid forking `grep` to read read from `/proc`.


tomponline (approved):
Btw what's the difference between modprobe -r and rmmod?

simondeziel:
> Btw what's the difference between modprobe -r and rmmod?

None that I'm aware. I just liked the symmetry between the modprobe for loading and unloading ;)

Issue #14264, opened 2024-10-11
`lxc storage volume copy` completion issue


simondeziel:
Here's the "completion flow":

```
# good
$ lxc storage volume copy 
default/     juju-btrfs/  juju-zfs/    

# extraneous ` ` (space)
$ lxc storage volume copy default/ 
default     juju-btrfs  juju-zfs    

# good
$ lxc storage volume copy default/
default/container/ansible    default/custom/ci-sdb    default/container/ansible01

# extraneous `custom/`
$ lxc storage volume copy default/custom/ci-sdb 
default     juju-btrfs  juju-zfs    

# the 2nd storage pool should be tab complete-able but is not
$ lxc storage volume copy default/custom/ci-sdb juju-zfs 
Error: Storage pool volume not found

# removing the bogus `custom/` bit works
# but the 2nd storage pool should have `/` at the end because a volume name needs to be provided
$ lxc storage volume copy default/ci-sdb juju-zfs

$ lxc storage volume copy default/ci-sdb juju-zfs/ci-sdb
Storage volume copied successfully!
```

@kadinsayani, there is a lot to unpack in the above so please let me if anything is unclear!

simondeziel:
@MusicDin I hope that covers your bogus about the bogus `custom/` bit showing up when tab completing volumes.

MusicDin:
Yes, thanks

simondeziel:
I believe code introduced some different oddities:

```
$ lxc storage volume copy default/<tab>`
default/24.04-desktop                                                     default/b0ded57c28fa07670f2318f5e63ca333413a022065f06829c277742cd6216371  default/noble-builder
default/5a63bc87974e61c631567c0d171fefffb33ebb7525b8295672ef0c5bf2cbd898
```

In the above, it's listing image volumes (fingerprint) along with instances and image aliases.

```
$ lxc storage volume list default
+-----------------+------------------------------------------------------------------+-------------+--------------+---------+
|      TYPE       |                               NAME                               | DESCRIPTION | CONTENT-TYPE | USED BY |
+-----------------+------------------------------------------------------------------+-------------+--------------+---------+------+
| container       | noble-builder                                                    |             | filesystem   | 1       |
+-----------------+------------------------------------------------------------------+-------------+--------------+---------+
| custom          | 24.04-desktop                                                    |             | iso          | 0       |
+-----------------+------------------------------------------------------------------+-------------+--------------+---------+
[...]

PR #14265, opened 2024-10-11
lxc/completion: `lxc storage volume copy` shell completion fixes and improvements to `completions.go`

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/14264.

This PR also replaces the use of `Sprintf()` with string concatenation where possible, and adds go doc comments to all completion functions.

### DIFF
--- a/lxc/completion.go
+++ b/lxc/completion.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"fmt"
 	"regexp"
 	"strings"
 
@@ -65,7 +64,7 @@ func (g *cmdGlobal) cmpClusterGroups(toComplete string) ([]string, cobra.ShellCo
 		if resource.remote == g.conf.DefaultRemote && !strings.Contains(toComplete, g.conf.DefaultRemote) {
 			name = group
 		} else {
-			name = fmt.Sprintf("%s:%s", resource.remote, group)
+			name = resource.remote + ":" + group
 		}
 
 		results = append(results, name)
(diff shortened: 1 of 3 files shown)


lxc/completion.go
```diff
@@ -1174,13 +1174,14 @@ func (g *cmdGlobal) cmpStoragePoolWithVolume(toComplete string) ([]string, cobra
 
 	var results []string
 	for _, volume := range volumes {
-		results = append(results, fmt.Sprintf("%s/%s", pool, volume))
+		volName, _ := parseVolume("volume", volume)
+		results = append(results, fmt.Sprintf("%s/%s", pool, volName))
```
simondeziel, line 1178: I know the code base is filled with this `fmt.Sprintf("%s/%s")` or other simple concatenation of strings but those are less efficient than a simple `+` operation when fancy conversion/formatting isn't required.

simondeziel (commented):
I've yet to test it locally but it LGTM. I'll to get to it Tuesday.

lxc/completion.go
kadinsayani: You're right! I'll update that :)

lxc/completion.go
```diff
@@ -1174,13 +1174,14 @@ func (g *cmdGlobal) cmpStoragePoolWithVolume(toComplete string) ([]string, cobra
...
 	}
 
 	return results, cobra.ShellCompDirectiveNoFileComp
 }
 
-func (g *cmdGlobal) cmpStoragePools(toComplete string) ([]string, cobra.ShellCompDirective) {
+func (g *cmdGlobal) cmpStoragePools(toComplete string, noSpace bool) ([]string, cobra.ShellCompDirective) {
```
tomponline, line 1184: Docblock missing explaining function and args

kadinsayani: Thanks for pointing this out - I've added doc comments to all the other completion functions as well (in a separate commit).

tomponline (approved):
Thanks!

PR #14275, opened 2024-10-14
Improve API metrics logging

hamistao:
Recently we are getting unwanted API metrics logging on the LXD agent, MicroCloud and MicroCluster.

As a fix, this PR removes the logging logic, as it is not needed since https://github.com/canonical/lxd/pull/14228. We are now typing the endpoints manually and checking if all the appropriate endpoints are typed.

This also fixes handling communication errors for the metrics, to not let those errors affect the request result.

### DIFF
--- a/lxd/metrics/api_rates.go
+++ b/lxd/metrics/api_rates.go
@@ -2,14 +2,11 @@ package metrics
 
 import (
 	"net/http"
-	"strings"
 	"sync"
 	"sync/atomic"
 
-	"github.com/canonical/lxd/lxd/auth"
 	"github.com/canonical/lxd/lxd/request"
 	"github.com/canonical/lxd/shared/entity"
-	"github.com/canonical/lxd/shared/logger"
 )
 
 // RequestResult represents a completed request status category.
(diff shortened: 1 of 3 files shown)


lxd/metrics/api_rates.go
```diff
@@ -79,6 +81,11 @@ func GetCompletedRequests(entityType entity.Type, result RequestResult) int64 {
 func TrackStartedRequest(r *http.Request, endpointType entity.Type) {
 	// Set the callback function to track the request as completed.
 	// Use sync.Once to ensure it can be called at most once.
+	if !shared.ValueInSlice(endpointType, entity.APIMetricsEntityTypes()) {
```
markylaing, line 84: Is it possible to enter this block? I thought that all metrics entity types were checked at start up when the server is initialised?

lxd/daemon.go
```diff
@@ -826,6 +826,12 @@ func (d *Daemon) createCmd(restAPI *mux.Router, version string, c APIEndpoint) {
 			resp = response.NotFound(fmt.Errorf("Method %q not found", r.Method))
 		}
 
+		// Log a warning if endpoint is part of the main API, and therefore should be counted fot the API metrics.
```
tomponline, line 829: typo "fot"

tomponline: do we even need this logging anymore now we force the metrics type on the API handler?

lxd/metrics/api_rates.go
hamistao: It is possible if you use `TrackStartedRequest` on an endpoint that is not supposed to be counted. Or at least that was the reasoning behind this, if you think it is not needed it is fine for me :)

lxd/daemon.go
hamistao: I was asking myself the same thing. It would be nice if we could check the callback function was called somehow, but I don't see how we would do this.

tomponline: OK think we should just remove this part for now.

lxd/metrics/api_rates.go
tomponline: Yeh lets remove this for now.

hamistao: Done!

lxd/daemon.go
hamistao: Done!

tomponline (approved):

PR #14287, opened 2024-10-16
Makefile: opportunistically run yamllint against GH workflows

simondeziel:
GHA runners comes with `yamllint` installed by default (https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md)

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -55,11 +55,11 @@ jobs:
 
     steps:
     - name: Checkout repository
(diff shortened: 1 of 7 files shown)


simondeziel:
@tomponline by fixing the `uses: actions/...@SHA1  # vX.Y` syntax to not have 2 spaces, I hope to it will make dependabot consider those pinned GHA as valid actions needing to be kept up to date.

tomponline (approved):

PR #14290, opened 2024-10-17
github: use shorter job names for Trivy scanning

simondeziel:
The GitHub UI trims long names like this:

> Trivy vulnerability scanner - Rep...
> Trivy vulnerability scanner - Sna...
> Trivy vulnerability scanner - Sna...
> Trivy vulnerability scanner - Sna...
> Trivy vulnerability scanner - Sna...

Since the workflow title already makes it clear that Trivy is a vulnerability scanner, there is no need to repeat it.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -18,7 +18,7 @@ defaults:
 
 jobs:
   trivy-repo:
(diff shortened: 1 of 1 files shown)


tomponline (approved):

tomponline:
much better

PR #14292, opened 2024-10-17
lxd/rsync: check if the source provided to LocalCopy() is an existing dir

simondeziel:


tomponline:
@simondeziel if you're working on what i think you're working on, then I'm thinking we should target the fix in either `doAPI10UpdateTriggers` or `daemonStorageMove` by checking the volume exists in the DB. Rather than at the lower level that may prevent using this function more generally.

simondeziel:
> @simondeziel if you're working on what i think you're working on, then I'm thinking we should target the fix in either `doAPI10UpdateTriggers` or `daemonStorageMove` by checking the volume exists in the DB.

Good idea, thanks

> Rather than at the lower level that may prevent using this function more generally.

The function does `shared.AddSlash(source)` later on so I don't think that would prevent more general use.

lxd/rsync/rsync.go
```diff
@@ -68,6 +68,10 @@ func LocalCopy(source string, dest string, bwlimit string, xattrs bool, rsyncArg
 		return "", err
 	}
 
+	if !shared.IsDir(source) {
```
tomponline, line 71: What does this check help with?

lxd/daemon_storage.go
```diff
@@ -133,7 +133,7 @@ func daemonStorageMount(s *state.State) error {
 }
 
 func daemonStorageSplitVolume(volume string) (poolName string, volumeName string, err error) {
-	fields := strings.Split(volume, "/")
+	fields := strings.SplitN(volume, "/", 3)
```
tomponline, line 136: strings.Cut?

lxd/daemon_storage.go
```diff
@@ -316,6 +316,30 @@ func daemonStorageMove(s *state.State, storageType string, target string) error
 		return err
 	}
 
+	// Validate pool and volume exist.
+	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
```
tomponline, line 320: This will need moving higher up due to similar issues with the section starting on line 274 using unvalidated input.

Suggest doing this lookup earlier and then replacing all uses of sourcePool and sourceVolume with the vars loaded from the DB for pool and volume respectively.

tomponline: Can you also update `daemonStorageValidate` to use a similar approach, i.e only use the user input args to load the DB records, and all internal use for pool and volume name use the vars from the DB.

lxd/rsync/rsync.go
tomponline: This check, assuming its needed, would be better done before we make the target dir above.

lxd/rsync/rsync.go
```diff
@@ -68,6 +68,10 @@ func LocalCopy(source string, dest string, bwlimit string, xattrs bool, rsyncArg
...
+		return "", fmt.Errorf("rsync source (%s) is not a directory", source)
```
tomponline, line 72: We tend to use %q for quoting parameters in errors rather than "(%s)".

Also we tend to start errors with upper case characters, im not sure "Rsync" works here, perhaps just "Source %q is not a directory" (the rsync part could be added by the caller, i.e `"Failed to rsync: %w", err`, with err being `Source %q is not a directory`.

lxd/daemon_storage.go
tomponline: Actually, suggest we split daemonStorageValidate into multiple functions, with one just dealing with validating and returning the DB info and then the same method can be used both during setting the values as well as moving them.

lxd/daemon_storage.go
simondeziel: I stuck with `strings.Split` because this allows catching bogus param like `<pool>/<volume>/<junk>`. However as you suggested doing more validation by splitting `daemonStorageValidate` into multiple functions, maybe that's OK to use `strings.Cut` and let the DB lookup deal with the bogus volume name `<volume>/<junk>`?

lxd/rsync/rsync.go
simondeziel: My initial intention was to address the CodeQL alert but that `source` isn't the problem in the end.

Considering that `rsync` tries to `chdir` when the source has a trailing `/`, it might not be needed at all as you suggested.
```
# rsync -a /etc/hostname/ /tmp/foo/
rsync: [sender] change_dir "/etc/hostname" failed: Not a directory (20)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1338) [sender=3.2.7]
```
Happy to drop it if you think the `Source %q is not a directory` error isn't worth it.

tomponline: Yeah let's drop it, it's just an extra syscall not needed

lxd/daemon_storage.go
tomponline: Yeah let the db lookup handle it

tomponline:
@simondeziel did you need any assistance with this?

simondeziel:
I surely will, but @markylaing has already proposed peer programming with me so I might bring this up with him :)
Thanks Tom!

tomponline:
Closing for now to reduce PR list.

PR #14294, opened 2024-10-17
Notifier update; MemberState update

MggMuggins:
This includes preparatory changes as described in #14247:

- Notifier improvements
  - Use `state.ServerClustered`
  - Don't query for `OfflineThreshold` unless needed (see commit msg)
  - Allow passing `db.NodeInfo` to `NewNotifier`
  - Pass `db.NodeInfo` to the notifier hook to reduce the need for extra http calls for basic info (like the target server's name)
- Member state updates
  - Add `CPUThreads` to `ClusterMemberSysInfo` (name tbd)
  - Refactor `MemberState` to expose local `ClusterMemberSysinfo`
  - Implement helper to collect member state from all cluster members, using the notifier improvements above
- Other odds and ends

Performance improvements for project limits will be in a separate PR.

### DIFF
--- a/lxd/cluster/member_state.go
+++ b/lxd/cluster/member_state.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"runtime"
 	"strconv"
 	"strings"
 
@@ -72,6 +73,10 @@ func LocalSysInfo() (*api.ClusterMemberSysInfo, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Failed getting load averages: %w", err)
 	}
+
+	// NumCPU gives the number of threads available to the LXD server at startup,
+	// not the currently available number of threads.
+	sysInfo.LogicalCPUs = uint64(runtime.NumCPU())
 
 	return sysInfo, nil
 }
--- a/lxd/cluster/member_state.go
+++ b/lxd/cluster/member_state.go
@@ -77,7 +77,7 @@ func LocalSysInfo() (*api.ClusterMemberSysInfo, error) {
 }
 
 // MemberState retrieves state information about the cluster member.
-func MemberState(ctx context.Context, s *state.State, memberName string) (*api.ClusterMemberState, error) {
+func MemberState(ctx context.Context, s *state.State) (*api.ClusterMemberState, error) {
 	var memberState api.ClusterMemberState
 
 	sysInfo, err := LocalSysInfo()
--- a/lxd/cluster/member_state.go
+++ b/lxd/cluster/member_state.go
@@ -45,34 +45,47 @@ func getLoadAvgs() ([]float64, error) {
 	return loadAvgs, nil
 }
 
-// MemberState retrieves state information about the cluster member.
-func MemberState(ctx context.Context, s *state.State, memberName string) (*api.ClusterMemberState, error) {
-	var err error
-	var memberState api.ClusterMemberState
-
+// LocalSysInfo retrieves system information about a cluster member.
+func LocalSysInfo() (*api.ClusterMemberSysInfo, error) {
 	// Get system info.
 	info := unix.Sysinfo_t{}
-	err = unix.Sysinfo(&info)
+	err := unix.Sysinfo(&info)
 	if err != nil {
 		logger.Warn("Failed getting sysinfo", logger.Ctx{"err": err})
 
 		return nil, err
 	}
 
-	// Account for different representations of Sysinfo_t on different architectures.
-	memberState.SysInfo.Uptime = int64(info.Uptime)
-	memberState.SysInfo.TotalRAM = uint64(info.Totalram)
-	memberState.SysInfo.SharedRAM = uint64(info.Sharedram)
-	memberState.SysInfo.BufferRAM = uint64(info.Bufferram)
-	memberState.SysInfo.FreeRAM = uint64(info.Freeram)
-	memberState.SysInfo.TotalSwap = uint64(info.Totalswap)
-	memberState.SysInfo.FreeSwap = uint64(info.Freeswap)
+	sysInfo := &api.ClusterMemberSysInfo{}
 
-	memberState.SysInfo.Processes = info.Procs
-	memberState.SysInfo.LoadAverages, err = getLoadAvgs()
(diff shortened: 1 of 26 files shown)


lxd/cluster/notify.go
```diff
@@ -43,11 +48,12 @@ func NewNotifier(state *state.State, networkCert *shared.CertInfo, serverCert *s
 
 	var err error
 	var members []db.NodeInfo
-	var offlineThreshold time.Duration
 	err = state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
-		offlineThreshold, err = tx.GetNodeOfflineThreshold(ctx)
-		if err != nil {
-			return err
+		if state.GlobalConfig == nil {
+			offlineThreshold, err = tx.GetNodeOfflineThreshold(ctx)
```
tomponline: Could you add a comment to the effect of the commit message here explaining why we need it.

tomponline: which invocation of NewNotifier is called when there's no global DB btw?

lxd/cluster/notify.go
```diff
@@ -46,25 +46,28 @@ func NewNotifier(state *state.State, networkCert *shared.CertInfo, serverCert *s
...
-			offlineThreshold, err = tx.GetNodeOfflineThreshold(ctx)
-			if err != nil {
-				return fmt.Errorf("Failed getting node offline threshold: %w", err)
+	if len(members) == 0 || state.GlobalConfig == nil {
+		err := state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+			var err error
+			if state.GlobalConfig == nil {
+				offlineThreshold, err = tx.GetNodeOfflineThreshold(ctx)
+				if err != nil {
+					return fmt.Errorf("Failed getting node offline threshold: %w", err)
```
tomponline, line 56: Failed getting cluster member offline threshold

lxd/cluster/notify.go
```diff
@@ -84,36 +85,36 @@ func NewNotifier(state *state.State, networkCert *shared.CertInfo, serverCert *s
 			if !HasConnectivity(networkCert, serverCert, member.Address) {
 				switch policy {
 				case NotifyAll:
-					return nil, fmt.Errorf("peer node %s is down", member.Address)
+					return nil, fmt.Errorf("Peer node %s is down", member.Address)
```
tomponline, line 88: Peer member...

lxd/cluster/notify.go
```diff
@@ -84,36 +85,36 @@ func NewNotifier(state *state.State, networkCert *shared.CertInfo, serverCert *s
...
-	notifier := func(hook func(lxd.InstanceServer) error) error {
+	notifier := func(hook func(db.NodeInfo, lxd.InstanceServer) error) error {
 		errs := make([]error, len(peers))
 		wg := sync.WaitGroup{}
 		wg.Add(len(peers))
-		for i, address := range peers {
-			logger.Debugf("Notify node %s of state changes", address)
-			go func(i int, address string) {
+		for i, member := range peers {
+			logger.Debugf("Notify node %s at %s of state changes", member.Name, member.Address)
```
tomponline, line 104: Notify member...

tomponline: Also this is a good opportunity to move away from `logger.*f()` and use contextual logging

lxd/cluster/notify.go
```diff
@@ -123,7 +124,7 @@ func NewNotifier(state *state.State, networkCert *shared.CertInfo, serverCert *s
 				isDown := shared.IsConnectionError(err) || api.StatusErrorCheck(err, http.StatusServiceUnavailable)
 
 				if isDown && policy == NotifyAlive {
-					logger.Warnf("Could not notify node %s", peers[i])
+					logger.Warn("Could not notify node", logger.Ctx{"name": peers[i].Name, "address": peers[i].Address})
```
tomponline, line 127: Could not notify member

lxd/networks.go
```diff
@@ -710,7 +705,7 @@ func networksPostCluster(s *state.State, projectName string, netInfo *api.Networ
 		}
 
 		// Merge node specific config items into global config.
-		for key, value := range nodeConfigs[server.Environment.ServerName] {
+		for key, value := range nodeConfigs[member.Name] {
```
tomponline, line 708: Nice!

lxd/cluster/member_state.go
```diff
@@ -81,6 +82,50 @@ func LocalSysInfo() (*api.ClusterMemberSysInfo, error) {
 	return sysInfo, nil
 }
 
+// ClusterState returns a map from clusterMemberName -> state for every member
+// of the cluster. This requires an HTTP call to the rest of the cluster.
+func ClusterState(s *state.State, members []db.NodeInfo) (map[string]api.ClusterMemberState, error) {
```
tomponline, line 87: members could be a variadic here too?

tomponline:
cluster tests are sad

lxd/cluster/notify.go
MggMuggins: [`NotifyUpgradeCompleted`](https://github.com/canonical/lxd/blob/main/lxd/cluster/upgrade.go#L24), called [in the daemon](https://github.com/canonical/lxd/blob/main/lxd/daemon.go#L1390) before `patchesApply(d, patchPreLoadClusterConfig`. I didn't want to side-quest too far into why `NotifyUpgradeCompleted` is happening before we apply patches. Too late I guess:

I think the cluster notification has to happen before patches are applied so that we don't end up deadlocking without quorum:
- Old server starts
- New server starts, fails to open DB and waits for old server to send upgrade notification
- Old server is upgraded, restarts, starts global db and sends upgrade notification
- New server receives upgrade notification and retries opening the DB
- DB has reached quorum and both servers can run patches
If the old server tries to run patches or otherwise access the DB before sending a notification, the DB will only be started on that member and will not have quorum

MggMuggins: I also considered using the default offline threshold instead of going back to the DB, but I think this is probably more correct.

tomponline (approved):
LGTM

Please can you add an API extension for the new cluster member state LogicalCPUs field as a follow up PR. Thanks

PR #14295, opened 2024-10-17
doc: embed discourse post title as link text

minaelee:
This PR explicitly specifies the link text for Discourse posts (from the posts' titles) instead of extracting it at build time. Recent Discourse outages have caused build failures when the link text couldn't be fetched dynamically. Hard-setting the link text should prevent this.

**Note:** Docs on specifying link text [here](https://github.com/canonical/canonical-sphinx-extensions/?tab=readme-ov-file#specify-links-for-a-page). The docs also describes a backup link text option, but I did not use it due to buggy behavior. _My expectation:_ The backup link text will be used if the link text cannot be extracted dynamically, and the build will not fail. _Reality:_ The backup link text is recognized, yet the build still fails.

### DIFF
--- a/doc/explanation/authorization.md
+++ b/doc/explanation/authorization.md
@@ -1,5 +1,5 @@
 ---
-discourse: ubuntu:41516
+discourse: ubuntu:[Identity&#32;and&#32;Access&#32;Management&#32;for&#32;LXD](41516)
 ---
 
 (authorization)=
--- a/doc/howto/snap.md
+++ b/doc/howto/snap.md
@@ -1,5 +1,5 @@
 ---
-discourse: ubuntu:37214
+discourse: ubuntu:[Managing&#32;the&#32;LXD&#32;snap&#32;package](37214)
 ---
 
 (howto-snap)=
--- a/doc/installing.md
+++ b/doc/installing.md
@@ -1,5 +1,5 @@
 ---
-discourse: ubuntu:37327
+discourse: ubuntu:[Building&#32;custom&#32;LXD&#32;binaries&#32;for&#32;side&#32;loading&#32;into&#32;an&#32;existing&#32;snap&#32;installation](37327)
 ---
 
 (installing)=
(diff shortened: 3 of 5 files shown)


tomponline (changes requested):
LXD maintains several LTS series of releases and we backport selected changes from the main branch into those LTS branches.

We dont backport everthing, as anything that causes an API break or a DB schema change cannot be backported.

This means that the docs between the main and LTS branches do diverge over time.

By making this change as a single commit it means that when I come to back port this change (which I will want to do) the chances of it applying cleanly to the LTS series is reduced (because not all of the associated pages/sections maybe in the LTS branch).

I would suggest this change be broken down into 1 commit per file so that I can selectively backport the change and skip over any ones that are not relevant.

Thanks!

simondeziel (commented):
To get those backup link texts, did you use a tool? If so, it'd be nice to have it integrated in a `Makefile` target (or similar) to update them all in one go from time to time.

tomponline (approved):

minaelee:
> To get those backup link texts, did you use a tool? If so, it'd be nice to have it integrated in a `Makefile` target (or similar) to update them all in one go from time to time.

I added these link texts manually since there were so few Discourse links. I agree, with the move to embedded link titles in general, an update tool would be handy to integrate in the future.

Not all links in the documentation require this update. Most are already embedded via Markdown, in the `[Link text](https://link)` syntax. From what I can tell, aside from the now-fixed Discourse links, there are ~20 only other links that currently fetch their link text dynamically by querying the link URL, using the `canonical-sphinx-extensions` package. These are the ones set as values for the `relatedlinks:` key in the frontmatter of docs .md files. I can do a quick manual update of these for now, then look into integrating an updater.

simondeziel:
> > To get those backup link texts, did you use a tool? If so, it'd be nice to have it integrated in a `Makefile` target (or similar) to update them all in one go from time to time.
> 
> @simondeziel I added these link texts manually since there were so few Discourse links.

Good as those handful of links are enough to bring the discourse down to its knees when we push builds to `5.21/latest`, `5.21/candidate` and `latest/edge` like we did today. I spent most of the day scheduling builds one by one to avoid the thundering herd :/

> I agree, with the move to embedded link titles in general, an update tool would be handy to integrate in the future.

Excellent, thanks!

Issue #13412, opened 2024-05-01
OVN virtual router external addresses should show in `lxc network list-allocations` on the uplink network


tomponline:
Currently when listing IP allocations, one can see the external addresses allocated by network forwards and network load balancers from OVN networks on the uplink network, e.g.

```
lxc network list-allocations 
+------------------------------------------+---------------------------+-----------------+-------+------------------+
|                 USED BY                  |          ADDRESS          |      TYPE       |  NAT  | HARDWARE ADDRESS |
+------------------------------------------+---------------------------+-----------------+-------+------------------+
| /1.0/networks/lxdbr0                     | 10.199.41.1/24            | network         | true  |                  |
+------------------------------------------+---------------------------+-----------------+-------+------------------+
| /1.0/networks/lxdbr0                     | fd42:9542:ff80:fdb::1/64  | network         | true  |                  |
+------------------------------------------+---------------------------+-----------------+-------+------------------+
| /1.0/networks/ovn1                       | 10.179.69.1/24            | network         | true  |                  |
+------------------------------------------+---------------------------+-----------------+-------+------------------+
| /1.0/networks/ovn1                       | fd42:ca1d:2f42:bf38::1/64 | network         | true  |                  |
+------------------------------------------+---------------------------+-----------------+-------+------------------+
| /1.0/networks/ovn1/forwards/192.168.10.1 | 192.168.10.1/32           | network-forward | false |                  |
+------------------------------------------+---------------------------+-----------------+-------+------------------+
```

But the OVN networks also allocate external IPs on the uplink network for the external port of the virtual router:

```
lxc network show ovn1
name: ovn1
description: ""
type: ovn
managed: true
status: Created
config:
  bridge.mtu: "1500"
  ipv4.address: 10.179.69.1/24
  ipv4.nat: "true"
  ipv6.address: fd42:ca1d:2f42:bf38::1/64
  ipv6.nat: "true"
  network: lxdbr0
  volatile.network.ipv4.address: 10.199.41.21
  volatile.network.ipv6.address: fd42:9542:ff80:fdb:216:3eff:fe51:3bd1
used_by: []
locations:
- none
```

We should display the `volatile.network.ipv4.address` and `volatile.network.ipv6.address` addresses in `lxc network list-allocations` as an allocation owned by the ovn network, i.e in this case `/1.0/networks/ovn1`.

The `type` displayed for these entries is debatable, but suggest we differentiate it from the `network` addresses by using something like `uplink`

[...]

Issue #14291, opened 2024-10-17
Snapshots missing in `used_by` for custom volumes and storage pools on `latest/edge` LXD build


mas-who:
# Required information

 * Distribution: snap
 * Distribution version: 2.63
 * The output of "snap list --all lxd core20 core22 core24 snapd":
 ```
Name    Version      Rev    Tracking       Publisher   Notes
core20  20240416     2318   latest/stable  canonical✓  base,disabled
core20  20240705     2379   latest/stable  canonical✓  base
core22  20240823     1612   latest/stable  canonical✓  base,disabled
core22  20240904     1621   latest/stable  canonical✓  base
core24  20240528     423    latest/stable  canonical✓  base,disabled
core24  20240710     490    latest/stable  canonical✓  base
lxd     git-dcd70b1  30709  latest/edge    canonical✓  disabled
lxd     git-35332a1  30717  latest/edge    canonical✓  -
snapd   2.65.3       22991  latest/stable  canonical✓  snapd,disabled
snapd   2.63         21759  latest/stable  canonical✓  snapd
```
 * The output of "lxc info" or if that fails:
```
config:
  acme.agree_tos: "true"
  core.https_address: '[::]:8443'
  oidc.audience: https://dev-xjrvvfikbsv4jxn7.us.auth0.com/api/v2/
  oidc.client.id: gxj297yfAjmklILK5WqPWDSbtVBAeSQm
  oidc.groups.claim: lxd-idp-groups
  oidc.issuer: https://dev-xjrvvfikbsv4jxn7.us.auth0.com/
  user.show_permissions: "true"
api_extensions:
- storage_zfs_remove_snapshots
- container_host_shutdown_timeout
- container_stop_priority
- container_syscall_filtering
- auth_pki
- container_last_used_at
- etag
- patch
- usb_devices
- https_allowed_credentials
- image_compression_algorithm
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
[...]

PR #14299, opened 2024-10-18
Networt: Fix incorrect handling of instances in UsedByInstanceDevices

tomponline:
During cherry-pick from Incus of refactor of this function in e5f5b887ae11d50f9b6bb99df2a008d2edd6e4e9 the project filtering logic was incorrectly returning rather than continuing to next instance.

This meant some instances were being excluded incorrectly.

Also fixes a potential issue with concurrent modification of the leases slice in `bridge.Leases()` when collecting leases from each cluster member concurrently.

Fixes https://github.com/canonical/lxd/issues/13412

### DIFF
--- a/lxd/network/driver_bridge.go
+++ b/lxd/network/driver_bridge.go
@@ -3632,6 +3632,18 @@ func (n *bridge) Leases(projectName string, clientType request.ClientType) ([]ap
 			return nil, err
 		}
 
+		leasesCh := make(chan api.NetworkLease)
+
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go func() {
+			for lease := range leasesCh {
+				leases = append(leases, lease)
+			}
+
+			wg.Done()
+		}()
+
 		err = notifier(func(client lxd.InstanceServer) error {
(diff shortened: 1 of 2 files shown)


hamistao (approved):

tomponline:
This may also help with https://github.com/canonical/lxd/issues/14291

tomponline:
@hamistao working on adding a fix for concurrent collection of leases for the bridge in clusters too

tomponline:
FYI @masnax there was a bad cherry-pick from Incus.

hamistao (approved):
Thanks for this!

lxd/network/driver_bridge.go
```diff
@@ -3632,6 +3632,18 @@ func (n *bridge) Leases(projectName string, clientType request.ClientType) ([]ap
 			return nil, err
 		}
 
+		leasesCh := make(chan api.NetworkLease)
+
+		var wg sync.WaitGroup
+		wg.Add(1)
```
hamistao, line 3638: Sorry, just noticed this now. Since we are dealing with only one goroutine, would using a channel instead of a waitGroup be more appropriate?

PR #14300, opened 2024-10-18
Backports (stable-5.0)

MusicDin:
Based on https://github.com/canonical/lxd/pull/14283

The only change is that if the trust token is provided with `--password` flag, it is used to validate the remote certificate instead of blindly accepting it.

However, if `--accept-certificate` flag is used, this validation is skipped.


tomponline (commented):
How come there are i18n updates?

MusicDin:
There is nothing new, just the lines are shifted for the existing entries due to code additions. I think the CI then complains that it is not updated.

tomponline (approved):

PR #14301, opened 2024-10-18
lxd: Use explicit cluster addr when core addr is wildcard

masnax:
This was noticed in MicroCloud. When removing a node from the LXD cluster, and then adding it back to the cluster, the old `core.https_address` and `cluster.https_address` are preserved.

When handling the request, LXD will attempt to use the `core.https_address` value and apply it to `cluster.https_address`, but if this value is a wildcard, we should continue with the explicit address defined under `ServerAddress` in the request body.

### DIFF
--- a/lxd/api_cluster.go
+++ b/lxd/api_cluster.go
@@ -536,6 +536,10 @@ func clusterPutJoin(d *Daemon, r *http.Request, req api.ClusterPut) response.Res
 			}
 
 			localHTTPSAddress = req.ServerAddress
+		} else if util.IsWildCardAddress(localHTTPSAddress) {
+			// Clustering requires an explicit address,
+			// so if core.https_address is a wildcard, we should still use the explicitly defined address.
+			localHTTPSAddress = req.ServerAddress
 		}
 
 		// Update the cluster.https_address config key.


masnax:
Ideally this should get backported to `5.21` as well, to work with the MicroCloud LTS. Seems we don't do individual cherry picks, since the checks on #14302 failed.

tomponline:
> When removing a node from the LXD cluster, and then adding it back to the cluster, the old `core.https_address` and `cluster.https_address` are preserved.

One thing to point out here is that "removing a node from the LXD cluster, and then adding it back to the cluster" is not officially supported. Your mileage may vary, and if this is something that MicroCloud is expecting then I would recommend getting a work item scheduled to investigate scenarios where this won't work as expected. cc @mseralessandri

lxd/api_cluster.go
```diff
@@ -535,6 +535,10 @@ func clusterPutJoin(d *Daemon, r *http.Request, req api.ClusterPut) response.Res
 				return response.SmartError(err)
 			}
 
+			localHTTPSAddress = req.ServerAddress
+		} else if util.IsWildCardAddress(localHTTPSAddress) {
```
tomponline, line 539: should this be an `else if` or just a separate if/

masnax: It's slightly more efficient this way.

The block above calls [IsAddressCovered](https://github.com/canonical/lxd/blob/main/lxd/util/net.go#L164) which returns true in cases where either the `core` and `cluster` address are the same, or `core` is a wildcard. 

If the additional condition was a separate `if` block, then we would always run through the wildcard logic twice, whereas with an `else if` we only check the address again if the first condition fails.

masnax:
> One thing to point out here is that "removing a node from the LXD cluster, and then adding it back to the cluster" is not officially supported. Your mileage may vary, and if this is something that MicroCloud is expecting then I would recommend getting a work item scheduled to investigate scenarios where this won't work as expected. cc @mseralessandri


Good to know. From my understanding, re-adding a system isn't something that we explicitly advertise, and most of the other snaps' cleanups are just best-effort as well.

lxd/api_cluster.go
tomponline: Thanks, we will backport this to 5.21

tomponline (approved):

PR #14305, opened 2024-10-20
Show gateway leases when `features.networks` is disabled (for default project networks)

hamistao:
If project `foo` has `features.networks` disabled on a project, it does not have networks of its own and the networks from the default project are visible from that project. However, the gateway leases for networks on the default project don't show on `lxc network list-leases lxdbr0 --project foo`.
This PR proposes changing this behavior and showing those leases, being more consistent with how we handle `features.networks`.

### DIFF
--- a/lxd/network/driver_ovn.go
+++ b/lxd/network/driver_ovn.go
@@ -5442,8 +5442,21 @@ func (n *ovn) Leases(projectName string, clientType request.ClientType) ([]api.N
 	var err error
 	leases := []api.NetworkLease{}
 
-	// If requested project matches network's project then include gateway IPs.
-	if projectName == n.project || projectName == "" {
+	// Include gateway leases if network is visible from requested project or requesting all projects.
+	// Avoid querying project if we don't need to check features.networks.
+	includeGatewayLeases := projectName == "" || n.project == projectName
+	if !includeGatewayLeases {
+		var effectiveProject string
+		effectiveProject, _, err = project.NetworkProject(n.state.DB.Cluster, projectName)
+		if err != nil {
+			return nil, err
+		}
+
+		includeGatewayLeases = n.project == effectiveProject
+	}
+
+	// Include gateway IPs if applicable.
+	if includeGatewayLeases {
 		// Add our own gateway IPs.
 		for _, addr := range []string{n.config["ipv4.address"], n.config["ipv6.address"]} {
 			ip, _, _ := net.ParseCIDR(addr)
--- a/lxd/network/driver_bridge.go
+++ b/lxd/network/driver_bridge.go
@@ -3447,10 +3447,23 @@ func (n *bridge) Leases(projectName string, clientType request.ClientType) ([]ap
 	instanceProjects := make(map[string]string)
 	leases := []api.NetworkLease{}
 
+	// Include gateway leases if network is visible from requested project or requesting all projects.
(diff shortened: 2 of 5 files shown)


lxd/network/driver_ovn.go
```diff
@@ -5442,8 +5442,30 @@ func (n *ovn) Leases(projectName string, clientType request.ClientType) ([]api.N
 	var err error
 	leases := []api.NetworkLease{}
 
-	// If requested project matches network's project then include gateway IPs.
-	if projectName == n.project || projectName == "" {
+	var project *api.Project
+
+	// Only query for project when we need to check for "features.networks". That is, when the network project is
+	// the default project and projectName refers to any one other project.
+	if projectName != api.ProjectDefaultName && projectName != "" && n.project == api.ProjectDefaultName {
```
tomponline, line 5449: Can we move this logic into `NetworkProject` to avoid the duplication? Would that be correct?

lxd/network/driver_ovn.go
```diff
@@ -5442,8 +5442,30 @@ func (n *ovn) Leases(projectName string, clientType request.ClientType) ([]api.N
...
+			return err
+		})
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Include gateway IPs when the network belongs to requested project, getting leases for all projects or
+	// requested project doesn't have features.networks and network is in default project.
+	if projectName == n.project || projectName == "" ||
```
tomponline, line 5467: please dont put if statements on multiple lines.
If its too long please break it up into multiple variables to make the if statement clearer.

lxd/network/driver_ovn.go
hamistao: I don't think this would be correct since we have to consider the networks from project `default` and the instances from the requested project (e.g. `projectName`)

tomponline: Lets move this to a function in network.Common

hamistao:
@tomponline This is ready for a review

tomponline:
@hamistao please can you demonstrate the scenario you're fixing here?

It sounds like you're trying to include the networks gateway IPs when doing `lxc network list-leases lxdbr0 --project=foo` where project `foo` has `features.networks: false`, is that right?

If so, I'm not sure we want the gateway IPs to show in that list as they do not belong to project `foo`.

hamistao:
@tomponline you are spot on! I figured adding this made sense since it feels more consistent with how we handle `features.networks`. For example, the networks from project `default` show on `lxc network ls --project foo`, so why wouldn't it's leases show on `lxc network list-leases`.
All that said, I am just explaining my reasoning so I would be fine with closing this regardless. The majority of time spent on this was really understanding and learning about network zones on LXD, so I got something from working on this anyway :))

tomponline:
OK I see.

I think we should leave it as it is, so that, yes, the network itself is visible to the non-default project, but the same way we don't show instances from the default project when listing leases for the non-default project, the gateway IPs belong to the default project.

However when doing `all-projects` (if supported) then yes I do think the gateways should show, so perhaps these is still a change to be made here?

hamistao:
@tomponline Using `all-projects` with `lxc network list-leases` is not currently supported. It could be a nice addition to implement here.

hamistao:
Adding `all-projects` to `/1.0/network/{network}/leases` will be done in the future if needed, as per our current plans, so this can be closed.

Issue #13507, opened 2024-05-27
Automatically downloaded images prevent project removal with "Only empty projects can be removed"


lathiat:
Currently you cannot remove an otherwise empty project, because it contains automatically downloaded images from launching containers. This is surprising and confusing to the average user, and it can be difficult to figure out what is preventing the removal.

I have run into this problem "manually" many times, but it also confuses https://github.com/terraform-lxd/terraform-provider-lxd/issues because when you try to destroy a project you have created it fails for the same reason.

I don't think it makes sense to force the user to cleanup these automatically downloaded and "cached" images manually, we should just automatically remove them. I don't see a safety benefit to making this manual for such image types, presuming we can always be sure about the safety of removing them.

This also, I think, lends to people doing workarounds like this script which probably will accidentally resulting removing the wrong kinds of images from the wrong project at some point:
`lxc image ls --project testproj1 -f csv -c f | xargs -n1 lxc image rm --project testproj1`

Previously also noted in https://github.com/canonical/lxd/issues/5634

Tested on 5.20 and 5.21/stable 5.21.1-d46c406 (28460) today.

# Reproducer
```
root@jammy:~# lxd init --auto

root@jammy:~# lxc project create testproj1
Project testproj1 created

root@jammy:~# lxc project switch testproj1

root@jammy:~# lxc profile device add default root disk pool=default path=/
Device root added to default

root@jammy:~# lxc profile device add default eth0 nic network=lxdbr0

root@jammy:~# lxc launch ubuntu:22.04 test1

root@jammy:~# lxc rm test1

root@jammy:~# lxc project switch default

root@jammy:~# lxc project rm testproj1
Error: Only empty projects can be removed

root@jammy:~# lxc image ls --project testproj1
+-------+--------------+--------+---------------------------------------------+--------------+-----------+-----------+------------------------------+
| ALIAS | FINGERPRINT  | PUBLIC |                 DESCRIPTION                 | ARCHITECTURE |   TYPE    |   SIZE    |         UPLOAD DATE          |
+-------+--------------+--------+---------------------------------------------+--------------+-----------+-----------+------------------------------+
|       | a99d6cd59df2 | no     | ubuntu 22.04 LTS amd64 (release) (20240514) | x86_64       | CONTAINER | 411.86MiB | May 27, 2024 at 5:46am (UTC) |
+-------+--------------+--------+---------------------------------------------+--------------+-----------+-----------+------------------------------+

[...]

PR #14316, opened 2024-10-21
lxd: Disregard auto-cached images when checking if project is empty

boltmark:
This PR loosens the conditions in which we consider a project empty. Now, we consider a project empty even if it contains automatically cached images. This is done to allow project deletion in the case where an otherwise empty project contains automatically cached images.

Closes https://github.com/canonical/lxd/issues/13507.

### DIFF
--- a/lxd/api_project.go
+++ b/lxd/api_project.go
@@ -796,7 +796,7 @@ func projectPost(d *Daemon, r *http.Request) response.Response {
 				return fmt.Errorf("Failed loading project %q: %w", name, err)
 			}
 
-			empty, err := projectIsEmpty(ctx, project, tx)
+			empty, err := projectIsEmpty(ctx, project, tx, true)
 			if err != nil {
 				return err
 			}
@@ -867,13 +867,27 @@ func projectDelete(d *Daemon, r *http.Request) response.Response {
 			return fmt.Errorf("Fetch project %q: %w", name, err)
 		}
 
-		empty, err := projectIsEmpty(ctx, project, tx)
+		empty, err := projectIsEmpty(ctx, project, tx, true)
 		if err != nil {
 			return err
 		}
 
 		if !empty {
 			return fmt.Errorf("Only empty projects can be removed")
+		}
+
+		// Remove any automatically cached images.
+		filterForCached := true
+		cachedImages, err := cluster.GetImages(ctx, tx.Tx(), cluster.ImageFilter{Project: &project.Name, Cached: &filterForCached})
+		if err != nil {
+			return err
+		}
+
+		for _, image := range cachedImages {
+			err = tx.DeleteImage(ctx, image.ID)
+			if err != nil {
+				return err
+			}
 		}
 
 		return cluster.DeleteProject(ctx, tx.Tx(), name)
@@ -953,7 +967,7 @@ func projectStateGet(d *Daemon, r *http.Request) response.Response {
 }
 
 // Check if a project is empty.
-func projectIsEmpty(ctx context.Context, project *cluster.Project, tx *db.ClusterTx) (bool, error) {
+func projectIsEmpty(ctx context.Context, project *cluster.Project, tx *db.ClusterTx, ignoreCachedImages bool) (bool, error) {
 	instances, err := cluster.GetInstances(ctx, tx.Tx(), cluster.InstanceFilter{Project: &project.Name})
 	if err != nil {
 		return false, err
@@ -969,7 +983,15 @@ func projectIsEmpty(ctx context.Context, project *cluster.Project, tx *db.Cluste
 	}
 
 	if len(images) > 0 {
-		return false, nil
+		if !ignoreCachedImages {
+			return false, nil
+		}
+
+		for _, image := range images {
+			if !image.Cached {
+				return false, nil
+			}
+		}
 	}
 
 	profiles, err := cluster.GetProfiles(ctx, tx.Tx(), cluster.ProfileFilter{Project: &project.Name})


lxd/api_project.go
```diff
@@ -968,8 +968,11 @@ func projectIsEmpty(ctx context.Context, project *cluster.Project, tx *db.Cluste
 		return false, err
 	}
 
-	if len(images) > 0 {
-		return false, nil
+	// Check for images that were not automatically cached.
```
tomponline, line 971: Did you consult where `projectIsEmpty` is used (other than when deleting a project)?

As a cursory glance shows me its also used in projectPost just before allowing a project rename?

Does this change potentially allow images to become orphaned if a project is renamed?

Also for project deletion, does this change require any additional changes to ensure that cached images that previously needed to be manually removed before project deletion are now deleted automatically when the project is deleted?

tomponline (changes requested):
Please can you reassure me that the impact of this small change hasn't introduced additional bugs.

lxd/api_project.go
boltmark: Hi @tomponline, thanks for the comment.

> Did you consult where projectIsEmpty is used (other than when deleting a project)?
> As a cursory glance shows me its also used in projectPost just before allowing a project rename?
> Does this change potentially allow images to become orphaned if a project is renamed?

Yes, currently just used when renaming and deleting. Tested renaming while cached images are present, and the images are not orphaned after rename. To avoid potential issues in the future though, I've added a boolean parameter `ignoreCachedImages` to `projectIsEmpty` so that the callee can make the decision.
```
$ lxc project switch foo
$ lxc image list
+-------+--------------+--------+---------------------------------------------+--------------+-----------+-----------+------------------------------+
| ALIAS | FINGERPRINT  | PUBLIC |                 DESCRIPTION                 | ARCHITECTURE |   TYPE    |   SIZE    |         UPLOAD DATE          |
+-------+--------------+--------+---------------------------------------------+--------------+-----------+-----------+------------------------------+
|       | 74957a558028 | no     | ubuntu 24.04 LTS amd64 (release) (20241004) | x86_64       | CONTAINER | 241.51MiB | Oct 22, 2024 at 7:31pm (UTC) |
+-------+--------------+--------+---------------------------------------------+--------------+-----------+-----------+------------------------------+
$ lxc image info 74957a558028
Fingerprint: 74957a5580288913be8a8727d121f16616805e3183629133029ca907f210f541
Size: 241.51MiB
Architecture: x86_64
Type: container
Public: no
Timestamps:
    Created: 2024/10/04 00:00 UTC
    Uploaded: 2024/10/22 19:31 UTC
    Expires: 2029/05/31 00:00 UTC
    Last used: 2024/10/22 19:31 UTC
Properties:
    release: noble
    serial: 20241004
    type: squashfs
    version: 24.04
    architecture: amd64
    description: ubuntu 24.04 LTS amd64 (release) (20241004)
    label: release
    os: ubuntu
Aliases:
Cached: yes
Auto update: enabled
Source:
    Server: https://cloud-images.ubuntu.com/releases
    Protocol: simplestreams
    Alias: 24.04
Profiles:
    - default
$ lxc project switch default
$ lxc project rename foo bar
Project foo renamed to bar
$ lxc project switch bar
$ lxc image ls
+-------+--------------+--------+---------------------------------------------+--------------+-----------+-----------+------------------------------+
| ALIAS | FINGERPRINT  | PUBLIC |                 DESCRIPTION                 | ARCHITECTURE |   TYPE    |   SIZE    |         UPLOAD DATE          |
+-------+--------------+--------+---------------------------------------------+--------------+-----------+-----------+------------------------------+
|       | 74957a558028 | no     | ubuntu 24.04 LTS amd64 (release) (20241004) | x86_64       | CONTAINER | 241.51MiB | Oct 22, 2024 at 7:31pm (UTC) |
+-------+--------------+--------+---------------------------------------------+--------------+-----------+-----------+------------------------------+
```

> Also for project deletion, does this change require any additional changes to ensure that cached images that previously needed to be manually removed before project deletion are now deleted automatically when the project is deleted?

Thanks for catching this. I've updated the PR to include code which deletes the cached images before deleting the project, if they exist.

lxd/api_project.go
```diff
@@ -867,7 +867,7 @@ func projectDelete(d *Daemon, r *http.Request) response.Response {
 			return fmt.Errorf("Fetch project %q: %w", name, err)
 		}
 
-		empty, err := projectIsEmpty(ctx, project, tx)
+		empty, err := projectIsEmpty(ctx, project, tx, true)
```
tomponline, line 870: if there isnt anything on disk that contains the project name for cached images then i think we can allow rename when there are cached images.

lxd/api_project.go
```diff
@@ -876,6 +876,20 @@ func projectDelete(d *Daemon, r *http.Request) response.Response {
...
 
+		// Remove any automatically cached images.
+		filterForCached := true
+		cachedImages, err := cluster.GetImages(ctx, tx.Tx(), cluster.ImageFilter{Project: &project.Name, Cached: &filterForCached})
+		if err != nil {
+			return err
+		}
+
+		for _, image := range cachedImages {
+			err = tx.DeleteImage(ctx, image.ID)
```
tomponline, line 887: hrm, i still think this is going to end up leaving image files potentially orphaned on the disk.

If we take `imageDelete()` as an example of what has to be called today in order to clear out the cached images, we can see that it checks if the image is referenced by other projects, and if not, it then deletes the image file on each cluster member using `imageDeleteFromDisk`.

So if we aren't doing this when deleting the project then we will leaving image files on the disk without a reference in the DB.

Separately I've spotted an opportunity to remove some duplication by calling `imageDeleteFromDisk` from 

https://github.com/canonical/lxd/blob/33eeb706884cd868e84f6f4a29d1657b57a32902/lxd/images.go#L2707-L2719

boltmark:
@tomponline I see that we have an incus import candidate which adds forced project deletion:
https://docs.google.com/spreadsheets/d/1Q28pj5zVGDjWIXedCB9LEEy4jVhj8PxcB0XB6ff75os/edit?gid=0#gid=0&range=179:179

This solves the more general issue of project deletion being impeded by resources the user doesn't care about, as is the case in https://github.com/canonical/lxd/issues/13507. What do you think about closing this PR this and instead importing the forced project deletion feature from incus?

tomponline:
> @tomponline I see that we have an incus import candidate which adds forced project deletion: https://docs.google.com/spreadsheets/d/1Q28pj5zVGDjWIXedCB9LEEy4jVhj8PxcB0XB6ff75os/edit?gid=0#gid=0&range=179:179
> 
> This solves the more general issue of project deletion being impeded by resources the user doesn't care about, as is the case in #13507. What do you think about closing this PR this and instead importing the forced project deletion feature from incus?

mm, yeah thats a good idea. Although scanning through that Incus PR quickly it seems like its using the LXD client to perform loopback requests to its own API. When evaluating that PR please can you consider whether that is the best way or whether we can directly call internal functions and avoid making an HTTP request back to LXD.

tomponline:
will close this for now then

PR #14315, opened 2024-10-21
Profiles & instance listing efficency improvements (from Incus)

hamistao:
Improves listing profiles/instances by caching some data on the entities and making DB queries way more efficient.

Contains cherry picks from https://github.com/lxc/incus/pull/1140 and https://github.com/lxc/incus/pull/1314.

@MggMuggins

PR #14318, opened 2024-10-21
Project limits: Use `InstanceList`

MggMuggins:
Follow-up to #14315

This tidies up the queries used for project limits checking; it likely results in improved performance for clusters with many entities.

### DIFF
--- a/lxd/db/instances.go
+++ b/lxd/db/instances.go
@@ -17,6 +17,7 @@ import (
 	"github.com/canonical/lxd/lxd/instance/instancetype"
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/osarch"
 )
 
 // InstanceArgs is a value object holding all db-related details about an instance.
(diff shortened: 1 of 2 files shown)


MggMuggins:
Had a weird error on the CI, unlikely to be related but I'll at least document it here:
```
=== RUN   TestEndpoints_NetworkSocketBasedActivation
    endpoints_test.go:54: 
        	Error Trace:	/home/runner/work/lxd/lxd/lxd/endpoints/endpoints_test.go:54
        	            				/home/runner/work/lxd/lxd/lxd/endpoints/network_test.go:66
        	Error:      	Received unexpected error:
        	            	readdirnames /tmp/lxd-endpoints-test-1872275927: readdirent lxd-endpoints-test-1872275927: bad file descriptor
        	Test:       	TestEndpoints_NetworkSocketBasedActivation
--- FAIL: TestEndpoints_NetworkSocketBasedActivation (0.00s)
```

tomponline:
> Error: Failed checking if instance creation allowed: Failed getting usage of project entities: Instance "c1" in project "test-usage" has no "limits.memory" config, either directly or via a profile

MggMuggins:
`InstanceList` won't grab profiles that are not in use by any instances. We need all the profiles in the project so that new instances can expand their config/devices.

I've put up a revision here that at least gets us from `O(n)` queries to `O(1)` (where n is instances); it does a lot of unnecessary faffing around with profiles that ends up getting thrown out (looks like about 4 queries whose results end up being dropped).

What do you think of something like this?
```go
// lxd/db/projects.go
type ProjectArgs struct {
	Project   api.Project
	Profiles  []api.Profile
	Instances []api.Instance
}

func (c *ClusterTx) ProjectsList(ctx context.Context, filters ...cluster.ProjectFilter) ([]ProjectArgs, error)
```

I've done a little fiddling here and have about half an implementation if you'd like me to pursue that further. Otherwise I'll get back to bugs.

MggMuggins:
The changes I've made here go from `O(n)` queries to `O(1)` queries for `fetchProject`; I definitely think this is still worthwhile.

It's inefficient because `InstanceList` is still doing a couple of queries to get profiles that `fetchProject` must repeat in order to get all the Project's profiles. In general `InstanceList` is fine; it's just not quite the right tool for `fetchProject`.

PR #14319, opened 2024-10-21
Remove old retry logic (from Incus)

hamistao:
Just a cherry pick for https://github.com/lxc/incus/pull/1140/commits/7c60b4a520c7554bb4a8103783419ed144e511e1

### DIFF
--- a/lxd/instances_get.go
+++ b/lxd/instances_get.go
@@ -17,17 +17,14 @@ import (
 	"github.com/canonical/lxd/lxd/cluster"
 	"github.com/canonical/lxd/lxd/db"
 	dbCluster "github.com/canonical/lxd/lxd/db/cluster"
-	"github.com/canonical/lxd/lxd/db/query"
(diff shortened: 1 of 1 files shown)


lxd/instances_get.go
```diff
@@ -304,12 +280,12 @@ func doInstancesGet(s *state.State, r *http.Request) (any, error) {
...
 	})
 	if err != nil {
-		return nil, err
+		return response.SmartError(err)
 	}
 
 	userHasPermission, err := s.Authorizer.GetPermissionChecker(r.Context(), auth.EntitlementCanView, entity.TypeInstance)
 	if err != nil {
-		return nil, err
+		return response.InternalError(err)
```
tomponline, line 288: Lets not wrap this in response.InternalError() and instead use response.SmartError() as we dont always know that GetPermissionChecker want us to return an internal error.

hamistao: Done!

tomponline (approved):

Issue #14286, opened 2024-10-16
Create cluster group with more than one member fails


edlerd:
# Required information

 * Distribution: snap 
 * Distribution version: git-6c31513 / revision 30699 / from latest/edge

# Issue description

- Creating a cluster group with `POST /1.0/cluster/groups` returns 409 conflict in case more than one cluster member is present in the payload.
- Creating a cluster group with zero or one cluster member works fine.
- Editing the group after creation using the `PUT` endpoint with the same payload as above, to add multiple members, works fine.

# Steps to reproduce

 1. ` POST 1.0/cluster/groups` with payload `{"name":"ddff","description":"","members":["pet-shrew","integral-python"]}`.
 2. Get error 409 with the error message `This \"nodes_clusters_groups\" entry already exists`

Expected is for the post request to succeed, so we can create a cluster group with multiple cluster members.

PR #14321, opened 2024-10-22
lxd: Fix cluster group create when >1 node is defined in the request

boltmark:
This PR fixes cluster group creation when multiple nodes are present in the request. This is currently failing due to an incorrect check for existing mappings in the `nodes_cluster_groups` table. This is due to the `NodeClusterGroups` struct not properly matching the backing table. To fix this, we ensure the `NodeClusterGroups` uses the `primary` directive where necessary.

Fixes https://github.com/canonical/lxd/issues/14286.

### DIFF
--- a/lxd/db/cluster/nodes_cluster_groups.mapper.go
+++ b/lxd/db/cluster/nodes_cluster_groups.mapper.go
@@ -7,7 +7,6 @@ package cluster
 import (
 	"context"
 	"database/sql"
-	"errors"
 	"fmt"
 	"net/http"
 	"strings"
@@ -31,16 +30,6 @@ SELECT nodes_cluster_groups.group_id, nodes.name AS node
   JOIN nodes ON nodes_cluster_groups.node_id = nodes.id
   WHERE ( nodes_cluster_groups.group_id = ? )
   ORDER BY nodes_cluster_groups.group_id
-`)
-
-var nodeClusterGroupID = RegisterStmt(`
-SELECT nodes_cluster_groups.id FROM nodes_cluster_groups
-  WHERE nodes_cluster_groups.group_id = ?
-`)
-
-var nodeClusterGroupCreate = RegisterStmt(`
-INSERT INTO nodes_cluster_groups (group_id, node_id)
-  VALUES (?, (SELECT nodes.id FROM nodes WHERE nodes.name = ?))
 `)
 
 var nodeClusterGroupDeleteByGroupID = RegisterStmt(`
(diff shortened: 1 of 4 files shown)


lxd/db/cluster/nodes_cluster_groups.mapper.go
```diff
@@ -222,26 +224,25 @@ func NodeClusterGroupExists(ctx context.Context, tx *sql.Tx, groupID int) (bool,
 	return true, nil
 }
 
-// GetNodeClusterGroupID return the ID of the node_cluster_group with the given key.
-// generator: node_cluster_group ID
-func GetNodeClusterGroupID(ctx context.Context, tx *sql.Tx, groupID int) (int64, error) {
-	stmt, err := Stmt(tx, nodeClusterGroupID)
+// GetNodeClusterGroup returns the node_cluster_group entry with the given group ID and node name.
+// generator: node_cluster_group GetOne
+func GetNodeClusterGroup(ctx context.Context, tx *sql.Tx, groupID int, nodeName string) (*NodeClusterGroup, error) {
```
tomponline, line 229: you cannot modify any .mapper.go files as they are generated by the DB generator.

tomponline: instead it sounds like the definition of the table isnt correct in the metadata, here: 

https://github.com/canonical/lxd/blob/main/lxd/db/cluster/cluster_groups.go#L7-L25

Or we've hit a limitation with the DB generator.

Cc @masnax

boltmark: Thanks @masnax! I've made the suggested update

tomponline: Looking into this more, it seems in other places in the code base we use `tx.AddNodeToClusterGroup`, so can we instead of updating the generator rules use this function in `clusterGroupsPost` and `clusterGroupPatch`.

Potentially then we can remove some of the generated DB functions for this table entirely.
And then we a consistent way to add cluster members to a cluster group.

Thanks!

boltmark: @tomponline thanks, I've updated `clusterGroupsPost` and `clusterGroupPatch` to use `tx.AddNodeToClusterGroup` instead of the generated `CreateNodeClusterGroup`. I've also removed the now-unneeded generated functions for the `nodes_cluster_groups` table.

tomponline (approved):
Ta!

PR #14322, opened 2024-10-22
Github: Use go mod version of go

tomponline:
This simplifies the go version setting to automatically match that of the go mod version.

The removal of the go version matrix is ineffectual as it was already effectively removed when we switched to the 2-stage build + system test pattern, where LXD is only built once.

The snap tests do use the later versions of Go that LXD snap is built with. And the system tests in this repo check our support of the go mod minimum Go version.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -364,7 +364,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32  # v5.0.2
         with:
-          go-version: 1.22.x
+          go-version-file: 'go.mod'
 
       - name: Download coverage data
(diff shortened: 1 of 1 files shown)


MusicDin (approved):

.github/workflows/tests.yml
```diff
@@ -68,10 +68,10 @@ jobs:
           path: ${{ steps.ShellCheck.outputs.sarif }}
         if: github.event_name == 'pull_request'
 
-      - name: Install Go (1.22)
+      - name: Install Go
         uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32  # v5.0.2
         with:
-          go-version: 1.22.x
+          go-version-file: 'go.mod'
```
simondeziel, line 74: Wow, nice feature :)

simondeziel (commented):
Thanks!

PR #14323, opened 2024-10-22
API: Make Warning & Operation metrics member specific

hamistao:
`lxd_operations_total` and `lxd_warnings_total` are currently being taken clusterwide, instead of returning just the entities related to the Node responding the metrics request. This goes against the overall design of the metrics, that are supposed to be per node and queried on each node on a cluster.
To fix this, this PR filters the queries for those entities based on the Node.

### DIFF
--- a/test/main.sh
+++ b/test/main.sh
@@ -265,6 +265,7 @@ if [ "${1:-"all"}" != "standalone" ]; then
     run_test test_clustering_join_api "clustering join api"
     run_test test_clustering_shutdown_nodes "clustering shutdown"
     run_test test_clustering_projects "clustering projects"
+    run_test test_clustering_metrics "clustering metrics"
     run_test test_clustering_update_cert "clustering update cert"
     run_test test_clustering_update_cert_reversion "clustering update cert reversion"
     run_test test_clustering_update_cert_token "clustering update cert token"


hamistao:
> Do warnings always have a node?

They do not, this is marked as Draft while I/we decide on how to handle nodeless warnings, I am thinking just include them in all node's metrics.

tomponline:
> > Do warnings always have a node?
> 
> They do not, this is marked as Draft while I/we decide on how to handle nodeless warnings, I am thinking just include them in all node's metrics.

Yeah, or just on the leader?

hamistao:
> Yeah, or just on the leader?

I think this makes even more sense. Since Prometheus is supposed to scrape them on every node, counting them for each node can have the effect of readundantly counting these warnings many times when aggregating the measurements.

lxd/api_metrics.go
```diff
@@ -378,18 +378,21 @@ func getFilteredMetrics(s *state.State, r *http.Request, compress bool, metricSe
 	return response.SyncResponsePlain(true, compress, metricSet.String())
 }
 
-func internalMetrics(ctx context.Context, daemonStartTime time.Time, tx *db.ClusterTx) *metrics.MetricSet {
+func internalMetrics(ctx context.Context, nodeName string, daemonStartTime time.Time, tx *db.ClusterTx) *metrics.MetricSet {
```
tomponline, line 381: please use serverName or memberName, not "node", ta

hamistao:
@tomponline @markylaing I am thinking on waiting for https://github.com/canonical/lxd/pull/14261 before proceeding here since I intend to use the `LeaderInfo` function defined there. I have cherry picked some commits to test and it all works fine.

tomponline:
@hamistao please can we get this one fixed ready for LXD 6.2 release

lxd/api_metrics.go
```diff
@@ -378,10 +379,40 @@ func getFilteredMetrics(s *state.State, r *http.Request, compress bool, metricSe
...
+func clusterMemberWarnings(ctx context.Context, s *state.State, tx *db.ClusterTx) ([]dbCluster.Warning, error) {
+	var filters []dbCluster.WarningFilter
+
+	leaderInfo, err := s.LeaderInfo()
+	if err != nil {
+		return nil, err
+	}
+
+	// Use local variables to get pointers.
+	memberName := s.ServerName
```
tomponline, line 394: why not use `&s.ServerName`?

lxd/api_metrics.go
```diff
@@ -378,10 +379,40 @@ func getFilteredMetrics(s *state.State, r *http.Request, compress bool, metricSe
 	return response.SyncResponsePlain(true, compress, metricSet.String())
 }
 
-func internalMetrics(ctx context.Context, daemonStartTime time.Time, tx *db.ClusterTx) *metrics.MetricSet {
```
tomponline, line 381: the change to this function's signature should be a separate commit as not related to the commit message

tomponline: Personally I would have plumbed in the replacement of daemonStartTime argument with *state.State as single commit first, including passing s.StartTime instead of daemonStartTime to `out.AddSamples()`.

Then as another commit I would have introcuced the `clusterMemberWarnings` function and replaced the call to ` dbCluster.GetWarnings` in `internalMetrics`.

As it is now the changes are munged together and harder to review.

tomponline (changes requested):
Is it possible to add some tests for this behaviour by faking some warnings and checking they only appear in the leader endpoint?

lxd/api_metrics.go
hamistao: Noted! I find it useful to understand how you think about this so I can try to make your life easier :)

I will be pushing these changes along with the tests

tomponline:
@hamistao is this ready for review yet, its still in draft?

hamistao:
@tomponline This is still a draft because I broke the tests while trying to make use of the `/internal/testing/warnings` endpoint. This should be an hour (give or take) away from being ready.

hamistao:
@tomponline Please could you look first into this smallish PR that would help with the tests implemented here: https://github.com/canonical/lxd/pull/14496.

Then after that one is merged this one will be ready for review as well.

I tested it all together and it all works, but it would probably be more nitty to keep these in separate PRs.

tomponline:
ready for rebase

hamistao:
@tomponline This is now also ready for a review since the aforementioned PR got merged.
Mind that, on top of filtering the warnings by node, this also filters out resolved and acknowledged warnings when counting the warnings for the metrics. The resolved ones are pruned so definitely they should be filtered out, the acknowledged ones I am not so sure. Please let me know what you think.

tomponline (approved):
thanks! Yes i think excluding resolved and acknowledge warnings make sense.

Issue #12836, opened 2024-02-07
ACL logs with MicroOVN do not work because of hard coded file path


escabo:
# Required information

 * Found with a MicroCloud setup using LXD 5.20

# Issue description

 * The `lxc network acl show-log bla` command always produces this error: `Error: Only OVN log entries may be retrieved at this time`

Seems like this line https://github.com/canonical/lxd/blob/main/lxd/network/acl/driver_common.go#L754 should be snap aware as if I manually create a link on the servers of the cluster to the snap location of the file (when using MicroOVN), the error disappears.

tomponline:
@escabo the `HostPath()` call in line https://github.com/canonical/lxd/blob/main/lxd/network/acl/driver_common.go#L734 is snap aware. 

Was this tested using a microovn installation or a normal Ubuntu based OVN setup (such as the one described here https://documentation.ubuntu.com/lxd/en/latest/howto/network_ovn_setup/#set-up-a-standalone-ovn-network)?

If it works with the standard OVN setup and not with the microovn setup then its likely an issue with the snap packaging in LXD not setting up the correct symlinks internally for the microovn log directory.

escabo:
This was with MicroOVN (in a MicroCloud setup) and I reported it because when I added symlinks, the error message disappeared.

tomponline:
@masnax want to take a look at this one given you've done the work around microovn content interfaces previously?

simondeziel:
I think LXD and MicroOVN snaps should have a plug/interface to allow for easy retrieval of OVN logs by LXD. On the MicroOVN side, all I could find is for [local log files](https://canonical-microovn.readthedocs-hosted.com/en/latest/how-to/logs/).

We currently have a couple of interfaces/connections between the 2:
```
# snap connections lxd | grep ovn
content[ovn-certificates]  lxd:ovn-certificates  microovn:ovn-certificates  -
content[ovn-chassis]       lxd:ovn-chassis       microovn:ovn-chassis       -

# snap connections microovn | grep lxd
content[ovn-certificates]  lxd:ovn-certificates          microovn:ovn-certificates  -
content[ovn-chassis]       lxd:ovn-chassis               microovn:ovn-chassis       -
```

I think an eventual `ovn-logs` interface should only be manually connected as we would not always want to receive OVN logs. On the LXD side, this should also be conditional to having `core.syslog_socket=true`.

This would be a joint effort with the MicroOVN team.

tomponline:
[...]

PR #14327, opened 2024-10-23
Network: Return ACL logs from syslogs when the OVN controller is deployed in MicroOVN

gabrielmougard:
closes https://github.com/canonical/lxd/issues/12836

We need to remember to enable the log tests for microovn on https://github.com/canonical/lxd-ci/pull/368 when this is merged

### DIFF
--- a/doc/howto/network_ovn_setup.md
+++ b/doc/howto/network_ovn_setup.md
@@ -162,6 +162,10 @@ See the linked YouTube video for the complete tutorial using four machines.
 
        lxc config set network.ovn.northbound_connection <ovn-northd-nb-db>
 
+   ```{note}
+   If you are using a MicroOVN deployment, pass the value of the MicroOVN node IP address you want to target. Prefix the IP address with `ssl:`, and suffix it with the `:6641` port number that corresponds to the OVN central service within MicroOVN.
+   ```
+
 1. Finally, create the actual OVN network (on the first machine):
 
        lxc network create my-ovn --type=ovn
--- a/lxd/network/acl/acl_ovn.go
+++ b/lxd/network/acl/acl_ovn.go
@@ -1,9 +1,11 @@
 package acl
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net"
 	"strconv"
 	"strings"
@@ -1154,4 +1156,58 @@ func ovnParseLogEntry(logline string, syslogTimestamp string, prefix string) str
 	}
 
 	return string(out)
+}
+
+// ovnParseLogEntriesFromJournald reads the OVN log entries from the systemd journal and returns them as a list of string entries.
+// Also, we chose to output the last 1000 entries to avoid overloading the system with too many log entries.
+func ovnParseLogEntriesFromJournald(ctx context.Context, systemdUnitName string, filter string) ([]string, error) {
+	var logEntries []string
+	cmd := []string{
+		"/usr/bin/journalctl",
+		"--unit", systemdUnitName,
+		"--directory", shared.HostPath("/var/log/journal"),
+		"--no-pager",
+		"--boot", "0",
+		"--case-sensitive",
+		"--grep", filter,
+		"--output-fields", "MESSAGE",
+		"-n", "1000",
+		"-o", "json",
+	}
+
+	stdout := bytes.Buffer{}
+	err := shared.RunCommandWithFds(ctx, nil, &stdout, cmd[0], cmd[1:]...)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to run journalctl to fetch OVN ACL logs: %w", err)
+	}
+
+	decoder := json.NewDecoder(&stdout)
+	for {
+		var sdLogEntry map[string]any
+		err = decoder.Decode(&sdLogEntry)
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			return nil, fmt.Errorf("Failed to parse log entry: %w", err)
+		}
+
+		message, ok := sdLogEntry["MESSAGE"].(string)
+		if !ok {
+			continue
+		}
+
+		timestamp, ok := sdLogEntry["__REALTIME_TIMESTAMP"].(string)
+		if !ok {
+			continue
+		}
+
+		logEntry := ovnParseLogEntry(message, timestamp, filter)
+		if logEntry == "" {
+			continue
+		}
+
+		logEntries = append(logEntries, logEntry)
+	}
+
+	return logEntries, nil
 }
--- a/lxd/network/acl/acl_ovn.go
+++ b/lxd/network/acl/acl_ovn.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"strconv"
 	"strings"
 	"time"
 
@@ -1041,9 +1042,26 @@ type ovnLogEntry struct {
 	Action   string `json:"action"`
 }
 
-// ovnParseLogEntry takes a log line and expected ACL prefix and returns a re-formated log entry if matching.
-func ovnParseLogEntry(input string, prefix string) string {
-	fields := strings.Split(input, "|")
+// ovnParseLogEntry takes a log line (that comes from either an ovn controller log file or from the syslogs)
+// and expected ACL prefix and returns a re-formated log entry if matching.
+// The 'timestamp' string is in microseconds format. If empty, the timestamp is extracted from the log entry.
+func ovnParseLogEntry(logline string, syslogTimestamp string, prefix string) string {
+	parseLogTimeFromFields := func(fields []string) (time.Time, error) {
+		return time.Parse(time.RFC3339, fields[0])
+	}
+
+	parseLogTimeFromTimestamp := func(syslogTimestamp string) (time.Time, error) {
+		tsInt, err := strconv.ParseInt(syslogTimestamp, 10, 64)
+		if err != nil {
+			return time.Time{}, fmt.Errorf("Failed to parse timestamp: %w", err)
+		}
+
+		// The provided timestamp is in microseconds and need to be converted to nanoseconds.
+		tsNs := tsInt * 1000
+		return time.Unix(0, tsNs).UTC(), nil
+	}
+
+	fields := strings.Split(logline, "|")
 
 	// Skip unknown formatting.
 	if len(fields) != 5 {
@@ -1071,8 +1089,14 @@ func ovnParseLogEntry(input string, prefix string) string {
 		return ""
 	}
 
-	// Parse the timestamp.
-	logTime, err := time.Parse(time.RFC3339, fields[0])
+	var logTime time.Time
+	var err error
+	if syslogTimestamp == "" {
+		logTime, err = parseLogTimeFromFields(fields)
+	} else {
+		logTime, err = parseLogTimeFromTimestamp(syslogTimestamp)
+	}
+
 	if err != nil {
 		return ""
 	}
--- a/lxd/network/acl/driver_common.go
+++ b/lxd/network/acl/driver_common.go
@@ -749,14 +749,14 @@ func (d *common) Delete() error {
 }
 
 // GetLog gets the ACL log.
-func (d *common) GetLog(clientType request.ClientType) (string, error) {
+func (d *common) GetLog(ctx context.Context, clientType request.ClientType) (string, error) {
 	// ACLs aren't specific to a particular network type but the log only works with OVN.
 	var logEntries []string
 	var err error
 
 	if shared.IsMicroOVNUsed() {
 		prefix := fmt.Sprintf("lxd_acl%d-", d.id)
-		logEntries, err = ovnParseLogEntriesFromJournald(context.TODO(), "snap.microovn.chassis.service", prefix)
+		logEntries, err = ovnParseLogEntriesFromJournald(ctx, "snap.microovn.chassis.service", prefix)
 		if err != nil {
 			return "", fmt.Errorf("Failed to get OVN log entries from syslog: %w", err)
 		}
--- a/lxd/network/acl/driver_common.go
+++ b/lxd/network/acl/driver_common.go
@@ -767,7 +767,7 @@ func (d *common) GetLog(clientType request.ClientType) (string, error) {
 	logEntries := []string{}
 	scanner := bufio.NewScanner(logFile)
 	for scanner.Scan() {
-		logEntry := ovnParseLogEntry(scanner.Text(), fmt.Sprintf("lxd_acl%d-", d.id))
+		logEntry := ovnParseLogEntry(scanner.Text(), "", fmt.Sprintf("lxd_acl%d-", d.id))
 		if logEntry == "" {
 			continue
 		}
--- a/lxd/network/acl/driver_common.go
+++ b/lxd/network/acl/driver_common.go
@@ -751,33 +751,47 @@ func (d *common) Delete() error {
 // GetLog gets the ACL log.
 func (d *common) GetLog(clientType request.ClientType) (string, error) {
 	// ACLs aren't specific to a particular network type but the log only works with OVN.
-	logPath := shared.HostPath("/var/log/ovn/ovn-controller.log")
-	if !shared.PathExists(logPath) {
-		return "", fmt.Errorf("Only OVN log entries may be retrieved at this time")
-	}
+	var logEntries []string
+	var err error
 
-	// Open the log file.
-	logFile, err := os.Open(logPath)
-	if err != nil {
-		return "", fmt.Errorf("Couldn't open OVN log file: %w", err)
-	}
-
-	defer func() { _ = logFile.Close() }()
-
-	logEntries := []string{}
-	scanner := bufio.NewScanner(logFile)
-	for scanner.Scan() {
-		logEntry := ovnParseLogEntry(scanner.Text(), "", fmt.Sprintf("lxd_acl%d-", d.id))
-		if logEntry == "" {
-			continue
+	if shared.IsMicroOVNUsed() {
+		prefix := fmt.Sprintf("lxd_acl%d-", d.id)
+		logEntries, err = ovnParseLogEntriesFromJournald(context.TODO(), "snap.microovn.chassis.service", prefix)
+		if err != nil {
+			return "", fmt.Errorf("Failed to get OVN log entries from syslog: %w", err)
+		}
+	} else {
+		// Else, if the current LXD deployment does not use MicroOVN,
+		// then try to read the OVN controller log file directly (a standalone OVN controller might be built-in with LXD).
+		logEntries = []string{}
+		prefix := fmt.Sprintf("lxd_acl%d-", d.id)
+		logPath := shared.HostPath("/var/log/ovn/ovn-controller.log")
+		if !shared.PathExists(logPath) {
+			return "", fmt.Errorf("Only OVN log entries may be retrieved at this time")
 		}
 
-		logEntries = append(logEntries, logEntry)
-	}
+		// Open the log file.
+		logFile, err := os.Open(logPath)
+		if err != nil {
+			return "", fmt.Errorf("Failed to open OVN log file: %w", err)
+		}
 
-	err = scanner.Err()
-	if err != nil {
-		return "", fmt.Errorf("Failed to read OVN log file: %w", err)
+		defer func() { _ = logFile.Close() }()
+
+		scanner := bufio.NewScanner(logFile)
+		for scanner.Scan() {
+			logEntry := ovnParseLogEntry(scanner.Text(), "", prefix)
+			if logEntry == "" {
+				continue
+			}
+
+			logEntries = append(logEntries, logEntry)
+		}
+
+		err = scanner.Err()
+		if err != nil {
+			return "", fmt.Errorf("Failed to read OVN log file: %w", err)
+		}
 	}
 
 	// Aggregates the entries from the rest of the cluster.
--- a/doc/.custom_wordlist.txt
+++ b/doc/.custom_wordlist.txt
@@ -143,6 +143,7 @@ MiB
 Mibit
 MicroCeph
 MicroCloud
+MicroOVN
 MII
 MinIO
 MITM
--- a/lxd/network/acl/acl_interface.go
+++ b/lxd/network/acl/acl_interface.go
@@ -1,6 +1,8 @@
 package acl
 
 import (
+	"context"
+
 	"github.com/canonical/lxd/lxd/cluster/request"
 	"github.com/canonical/lxd/lxd/state"
 	"github.com/canonical/lxd/shared/api"
@@ -19,7 +21,7 @@ type NetworkACL interface {
 	UsedBy() ([]string, error)
 
 	// GetLog.
-	GetLog(clientType request.ClientType) (string, error)
+	GetLog(ctx context.Context, clientType request.ClientType) (string, error)
 
 	// Internal validation.
 	validateName(name string) error
--- a/lxd/network_acls.go
+++ b/lxd/network_acls.go
@@ -637,7 +637,7 @@ func networkACLLogGet(d *Daemon, r *http.Request) response.Response {
 	}
 
 	clientType := clusterRequest.UserAgentClientType(r.Header.Get("User-Agent"))
-	log, err := netACL.GetLog(clientType)
+	log, err := netACL.GetLog(r.Context(), clientType)
 	if err != nil {
 		return response.SmartError(err)
 	}
--- a/shared/util.go
+++ b/shared/util.go
@@ -1514,3 +1514,15 @@ func ApplyDeviceOverrides(localDevices map[string]map[string]string, profileDevi
 
 	return localDevices, nil
 }
+
+// IsMicroOVNUsed returns whether the current LXD deployment is using a built-in openvswitch
+// or is connected to MicroOVN, which in this case, would make `/run/openvswitch` a symlink to
+// `/var/snap/lxd/common/microovn/chassis/switch`.
+func IsMicroOVNUsed() bool {
+	targetPath, err := os.Readlink("/run/openvswitch")
+	if err == nil && strings.HasSuffix(targetPath, "/microovn/chassis/switch") {
+		return true
+	}
+
+	return false
+}


gabrielmougard:
@tomponline this should be working. Regarding the documentation, as we discussed earlier, we probably need to add a note in the doc section `Set up a LXD cluster on OVN` to tell that a northbound_db_connection can also be added as a unix socket pointing toward an existing MicroOVN installation (`unix:/var/snap/microovn/common/run/ovn/ovnnd_db.sock`). However has discovered here https://chat.canonical.com/canonical/pl/7fdhjesei3n9tcz1tkwbastozr, this seems to work only for a single node MicroOVN.. Do we really want to document that if it is not working for a multi node MicroOVN deployment?

gabrielmougard:
Here is the lxd-ci PR: https://github.com/canonical/lxd-ci/pull/330

doc/howto/network_ovn_setup.md
```diff
@@ -162,6 +162,10 @@ See the linked YouTube video for the complete tutorial using four machines.
 
        lxc config set network.ovn.northbound_connection <ovn-northd-nb-db>
 
+   ```{note}
+   If you are using a MicroOVN deployment, you can pass the value of the MicroOVN node IP address you want target prefixed with `ssl:` and suffixed with the `:6641` port number that should target the OVN central service within MicroOVN.
```
minaelee, line 166:
```suggestion
   If you are using a MicroOVN deployment, pass the value of the MicroOVN node IP address you want to target. Prefix the IP address with `ssl:`, and suffix it with the `:6641` port number that corresponds to the OVN central service within MicroOVN.
```

gabrielmougard: fixed

lxd/network/acl/driver_common.go
```diff
@@ -752,32 +752,45 @@ func (d *common) Delete() error {
...
-		logEntries = append(logEntries, logEntry)
-	}
+		defer func() { _ = logFile.Close() }()
 
-	err = scanner.Err()
-	if err != nil {
-		return "", fmt.Errorf("Failed to read OVN log file: %w", err)
+		scanner := bufio.NewScanner(logFile)
+		for scanner.Scan() {
+			logEntry := ovnParseLogEntry(scanner.Text(), "", fmt.Sprintf("lxd_acl%d-", d.id))
```
simondeziel, line 782: I don't know how smart is the compiler but it might be better to move that `fmt.Sprintf()` call outside of the loop.

lxd/network/acl/driver_common.go
```diff
@@ -752,32 +752,45 @@ func (d *common) Delete() error {
...
-		return "", fmt.Errorf("Couldn't open OVN log file: %w", err)
-	}
+		// Check if this is OVN builtin of if OVN is provided by a MicroOVN deployment.
+		targetPath, err := os.Readlink("/run/openvswitch")
+		if err != nil {
+			return "", fmt.Errorf("Failed to read symlink while checking for OVN logs: %v\n", err)
+		}
 
-	defer func() { _ = logFile.Close() }()
+		if !openvswitch.MicroovnTargetChassisPath.MatchString(targetPath) {
```
simondeziel, line 763: Could we instead use `strings.HasSuffix(targetPath, "/microovn/chassis/switch")` instead of a `regexp` match?

LXD's snap has this in `daemon.start`:
```
elif [ -d "${SNAP_DATA}/microovn/chassis/switch" ]; then
    ln -s "${SNAP_DATA}/microovn/chassis/switch" /run/openvswitch
```

lxd/network/acl/acl_ovn.go
```diff
@@ -1071,10 +1072,26 @@ func ovnParseLogEntry(input string, prefix string) string {
...
+	if timestamp == "" {
+		// Parse the timestamp.
+		logTime, err = time.Parse(time.RFC3339, fields[0])
+		if err != nil {
+			return ""
+		}
+	} else {
+		// Else, if from syslog, the timestamp is provided as a separate field
+		// which is passed separately.
+		tsInt, err := strconv.ParseInt(timestamp, 10, 64)
```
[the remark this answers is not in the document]
gabrielmougard: here, `time.Unix(0, tsNs)` requires an `int64` and `strconv.Atoi` will give an `int`.. If I use `strconv.Atoi` and manually cast to an `int64` would there be a risk of overflowing that would corrupt the timestamp?

lxd/network/acl/driver_common.go
gabrielmougard: fixed

lxd/network/acl/driver_common.go
gabrielmougard: yes correct! fixed.

lxd/network/acl/acl_ovn.go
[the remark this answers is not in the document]
tomponline: yes on 32bit armhf systems, we should stick with strconv.ParseInt in my view and be explicit about using `int64` over the variable width `int`. Correctness first. Especially given this isn't in the hot path for core functionality.

minaelee (dismissed):

lxd/network/acl/acl_ovn.go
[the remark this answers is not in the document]
gabrielmougard: fixed

lxd/network/acl/acl_ovn.go
```diff
@@ -1131,3 +1148,53 @@ func ovnParseLogEntry(input string, prefix string) string {
 
 	return string(out)
 }
+
+// ovnParseLogEntriesFromSyslog reads the OVN log entries from the systemd journal and returns them as a list of string entries.
+func ovnParseLogEntriesFromSyslog(l logger.Logger, systemdUnitName string, prefix string) ([]string, error) {
```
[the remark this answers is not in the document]
gabrielmougard: fixed

lxd/network/acl/acl_ovn.go
```diff
@@ -1131,3 +1148,53 @@ func ovnParseLogEntry(input string, prefix string) string {
...
+	var logEntries []string
+	res, err := shared.RunCommand(
+		"/usr/bin/journalctl",
+		"--unit", systemdUnitName,
+		"--directory", "/var/lib/snapd/hostfs/var/log/journal",
+		"--no-pager",
```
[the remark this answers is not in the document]
gabrielmougard: I agree. fixed.

lxd/network/acl/acl_ovn.go
```diff
@@ -1131,3 +1148,53 @@ func ovnParseLogEntry(input string, prefix string) string {
...
+		"--grep", "acl_log",
```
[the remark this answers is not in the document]
gabrielmougard: sure. fixed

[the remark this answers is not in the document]
gabrielmougard: The idea was to remain consistent with the logic inside the `ovnParseLogEntry` that first filter the bulk of the entries with `acl_log` and then with `prefix` (so that we're really sure we're getting acl logs and not a random log that might contain `prefix` in its content... I don't know, maybe this case never happens..)

lxd/network/acl/acl_ovn.go
```diff
@@ -1131,3 +1148,53 @@ func ovnParseLogEntry(input string, prefix string) string {
...
+		"-n", "5000",
+		"-o", "json",
```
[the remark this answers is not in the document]
gabrielmougard: Thanks for this tip! fixed.

lxd/network/acl/acl_ovn.go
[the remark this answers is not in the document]
tomponline: if we can pre-filter it earlier it would make sense to imho

lxd/network/acl/acl_ovn.go
```diff
@@ -1131,3 +1148,56 @@ func ovnParseLogEntry(input string, prefix string) string {
...
+		"--grep", "acl_log",
+		"--output-fields", "MESSAGE",
+		"-n", "5000",
+		"-o", "json",
+	)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to read journal entries: %v", err)
+	}
+
+	lines := strings.Split(res, "\n")
```
tomponline, line 1171: why are we reading up to 5000 lines of data into a variable res and then duplicating that data into a slice (up to 10k entries now), rather than consuming the line by line output of the command in a stream?

I think you should be able to use shared.RunCommandWithFds combined with either a bytes.Buffer and json.NewDecoder(&buf).Decode(), and/or bufio.NewScanner

simondeziel: I don't know if that could work but `journalctl` as a `tail` mode where it will keep feeding new data as it shows up. This can also work with `-n X` if some backlog is also needed.

tomponline: we dont want a blocking tail as that isnt how the command works today, but we also dont really want to read the entire output into memory (multiple times).

lxd/network/acl/acl_ovn.go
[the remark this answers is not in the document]
gabrielmougard: ok fixed

lxd/network/acl/acl_ovn.go
gabrielmougard: this should be fixed

lxd/network/acl/acl_ovn.go
```diff
@@ -1131,3 +1149,64 @@ func ovnParseLogEntry(input string, prefix string) string {
...
+	cmd := []string{
+		"/usr/bin/journalctl",
+		"--unit", systemdUnitName,
+		"--directory", "/var/lib/snapd/hostfs/var/log/journal",
+		"--no-pager",
+		"--boot", "0",
+		"--case-sensitive",
+		"--grep", prefix,
+		"--output-fields", "MESSAGE",
+		"-n", "5000",
```
simondeziel, line 1165: Are we still OK with pulling up to 5k log lines?

lxd/network/acl/driver_common.go
```diff
@@ -10,7 +10,7 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/canonical/lxd/client"
+	lxd "github.com/canonical/lxd/client"
```
simondeziel, line 13: This I think is unwanted change that VSCode automatically does IIRC.

gabrielmougard: ah yes, I forgot this one....

gabrielmougard: fixed

simondeziel (dismissed):

lxd/network/acl/driver_common.go
```diff
@@ -751,33 +751,46 @@ func (d *common) Delete() error {
...
+	targetPath, err := os.Readlink("/run/openvswitch")
 	if err != nil {
-		return "", fmt.Errorf("Couldn't open OVN log file: %w", err)
+		return "", fmt.Errorf("Failed to read symlink while checking for OVN logs: %v\n", err)
 	}
 
-	defer func() { _ = logFile.Close() }()
+	var logEntries []string
+	if strings.HasSuffix(targetPath, "/microovn/chassis/switch") {
+		logEntries, err = ovnParseLogEntriesFromJournald("snap.microovn.chassis.service", fmt.Sprintf("lxd_acl%d-", d.id))
```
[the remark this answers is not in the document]
tomponline: agreed

lxd/network/acl/acl_ovn.go
```diff
@@ -1067,14 +1069,30 @@ func ovnParseLogEntry(input string, prefix string) string {
 	}
 
 	// Filter for our ACL.
-	if !strings.HasPrefix(aclEntry["name"], prefix) {
+	if prefix != "" && !strings.HasPrefix(aclEntry["name"], prefix) {
```
tomponline, line 1072: should this change be in a separate commit? Its not clear what its fixing?

lxd/network/acl/acl_ovn.go
```diff
@@ -1067,14 +1069,30 @@ func ovnParseLogEntry(input string, prefix string) string {
...
 	}
 
-	// Parse the timestamp.
-	logTime, err := time.Parse(time.RFC3339, fields[0])
-	if err != nil {
-		return ""
+	var logTime time.Time
+	var err error
+	// When reading from a log file, the timestamp is provided as a message parameter.
+	if timestamp == "" {
```
tomponline, line 1079: Lets break the time parsing aspect of this into a separate function, and then have the timestamp always provided to this function as an time.Time  argument (reduce cyclomatic complexity) .

lxd/network/acl/acl_ovn.go
```diff
@@ -1131,3 +1149,64 @@ func ovnParseLogEntry(input string, prefix string) string {
...
 	return string(out)
 }
+
+// ovnParseLogEntriesFromJournald reads the OVN log entries from the systemd journal and returns them as a list of string entries.
+func ovnParseLogEntriesFromJournald(systemdUnitName string, prefix string) ([]string, error) {
+	var logEntries []string
+	cmd := []string{
+		"/usr/bin/journalctl",
+		"--unit", systemdUnitName,
+		"--directory", "/var/lib/snapd/hostfs/var/log/journal",
```
tomponline, line 1159: Lets use HostPath here please

lxd/network/acl/acl_ovn.go
```diff
@@ -1131,3 +1149,64 @@ func ovnParseLogEntry(input string, prefix string) string {
...
+	}
+
+	stdout := strings.Builder{}
+	err := shared.RunCommandWithFds(context.TODO(), nil, &stdout, cmd[0], cmd[1:]...)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to run journalctl to fetch OVN ACL logs: %w", err)
+	}
+
+	scanner := bufio.NewScanner(strings.NewReader(stdout.String()))
+	for scanner.Scan() {
```
tomponline, line 1176: was there a reason you went for processing each line here rather than using `json.NewDecoder(&scanner).Decode()`?

lxd/network/acl/driver_common.go
tomponline, line 1: commit message needs updating to remove the reference to 5000 lines as not in this commit

lxd/network/acl/driver_common.go
```diff
@@ -751,33 +751,46 @@ func (d *common) Delete() error {
...
 func (d *common) GetLog(clientType request.ClientType) (string, error) {
 	// ACLs aren't specific to a particular network type but the log only works with OVN.
-	logPath := shared.HostPath("/var/log/ovn/ovn-controller.log")
-	if !shared.PathExists(logPath) {
-		return "", fmt.Errorf("Only OVN log entries may be retrieved at this time")
-	}
-
-	// Open the log file.
-	logFile, err := os.Open(logPath)
+	targetPath, err := os.Readlink("/run/openvswitch")
```
tomponline, line 754: what if this path doesn't exist, this would break existing installations that do have `/var/log/ovn/ovn-controller.log"` right?

gabrielmougard: hmmm true. I initially thought that this symlink would be created even with OVN built in, but this isn't the case.. l'll fix that.

lxd/network/acl/acl_ovn.go
gabrielmougard: No special reasons.

tomponline: ack. please can we use the NewDecoder approach if it works, as its simpler (stream directly into the json decode, which handles separating by lines for you).

lxd/network/acl/driver_common.go
tomponline: we do have to consider non-snap installs too

lxd/network/acl/acl_ovn.go
```diff
@@ -1042,7 +1044,7 @@ type ovnLogEntry struct {
 }
 
 // ovnParseLogEntry takes a log line and expected ACL prefix and returns a re-formated log entry if matching.
-func ovnParseLogEntry(input string, prefix string) string {
+func ovnParseLogEntry(input string, timestamp string, prefix string) string {
```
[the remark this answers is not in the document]
gabrielmougard: fixed

lxd/network/acl/acl_ovn.go
gabrielmougard: fixed

lxd/network/acl/acl_ovn.go
gabrielmougard: fixed

lxd/network/acl/driver_common.go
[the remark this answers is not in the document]
gabrielmougard: fixed

lxd/network/acl/driver_common.go
gabrielmougard: fixed

lxd/network/acl/acl_ovn.go
gabrielmougard: Not needed anymore. Fixed

lxd/network/acl/driver_common.go
```diff
@@ -751,33 +751,48 @@ func (d *common) Delete() error {
...
 
-	// Open the log file.
-	logFile, err := os.Open(logPath)
-	if err != nil {
-		return "", fmt.Errorf("Couldn't open OVN log file: %w", err)
-	}
+		// Open the log file.
+		logFile, err := os.Open(logPath)
+		if err != nil {
+			return "", fmt.Errorf("Couldn't open OVN log file: %w", err)
```
tomponline, line 777: Lets not use contractions in error messages. "Failed to..." will be fine.

lxd/network/acl/acl_ovn.go
```diff
@@ -1093,8 +1096,7 @@ func ovnParseLogEntry(input string, prefix string) string {
 		return ""
 	}
 
-	// Parse the timestamp.
-	logTime, err := time.Parse(time.RFC3339, fields[0])
+	logTime, err := parseLogTime(timestamp, fields)
```
tomponline, line 1099: what does `fields[0]` when the ovn log comes from `ovnParseLogEntriesFromJournald`?

lxd/network/acl/acl_ovn.go
```diff
@@ -1153,3 +1155,56 @@ func ovnParseLogEntry(input string, prefix string) string {
 
 	return string(out)
 }
+
+// ovnParseLogEntriesFromJournald reads the OVN log entries from the systemd journal and returns them as a list of string entries.
+func ovnParseLogEntriesFromJournald(systemdUnitName string, prefix string) ([]string, error) {
```
tomponline, line 1160: this function isnt used in the changes of this commit, so feels like this should be separated into 2 commits, or have this function be added as part of a later commit.

This commit should just be concerned with changing the function signature of `ovnParseLogEntry`.

lxd/network/acl/driver_common.go
```diff
@@ -751,33 +751,48 @@ func (d *common) Delete() error {
 // GetLog gets the ACL log.
 func (d *common) GetLog(clientType request.ClientType) (string, error) {
 	// ACLs aren't specific to a particular network type but the log only works with OVN.
-	logPath := shared.HostPath("/var/log/ovn/ovn-controller.log")
-	if !shared.PathExists(logPath) {
-		return "", fmt.Errorf("Only OVN log entries may be retrieved at this time")
-	}
+	var logEntries []string
+	var err error
+	prefix := fmt.Sprintf("lxd_acl%d-", d.id)
```
tomponline, line 756: can the scope of the prefix var be reduced by moving it into one of the if statements?

lxd/network/acl/driver_common.go
gabrielmougard: fixed

lxd/network/acl/acl_ovn.go
```diff
@@ -1041,6 +1041,28 @@ type ovnLogEntry struct {
 	Action   string `json:"action"`
 }
 
+// parseLogTime parses a timestamp that is a string format and returns a time.Time object.
+// If 'timestamp' is empty (which is the case if the timestamp is contained in the 'fields' slice when we're reading an OVN log file for example),
+// we try to extract the timestamp from the 'fields'.
+func parseLogTime(timestamp string, fields []string) (time.Time, error) {
```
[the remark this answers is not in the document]
tomponline: im not sure if local functions increase cyclomatic complexity actually.

lxd/network/acl/acl_ovn.go
tomponline, line 1: This looks better thanks.

However as `ovnParseLogEntry` signature has changed, i would have expected to see either in this commit or in the next commit an update to where `ovnParseLogEntry` is called, before introducing the change for microovn.

lxd/network/acl/acl_ovn.go
```diff
@@ -1156,3 +1157,56 @@ func ovnParseLogEntry(input string, timestamp string, prefix string) string {
...
+		"--boot", "0",
+		"--case-sensitive",
+		"--grep", prefix,
+		"--output-fields", "MESSAGE",
+		"-n", "1000",
+		"-o", "json",
+	}
+
+	stdout := strings.Builder{}
+	err := shared.RunCommandWithFds(context.TODO(), nil, &stdout, cmd[0], cmd[1:]...)
```
tomponline, line 1178: Can this function accept a context so we can pass it here, this way if the client's request context gets cancelled we can skip running this program or cancel it  midway through.

gabrielmougard: fixed

lxd/network/acl/acl_ovn.go
gabrielmougard: fixed

tomponline:
doc tests failing

simondeziel:
> doc tests failing

The fix is a rebase away AFAICT.

lxd/network/acl/driver_common.go
```diff
@@ -751,33 +751,50 @@ func (d *common) Delete() error {
...
+	var err error
+
+	// First, check if we can resolve the /run/openvswitch symlink to determine if OVN is in use.
+	// This is used in case of a MicroOVN deployment is interfaced with LXD.
+	targetPath, err := os.Readlink("/run/openvswitch")
+	if err == nil && strings.HasSuffix(targetPath, "/microovn/chassis/switch") {
+		prefix := fmt.Sprintf("lxd_acl%d-", d.id)
+		logEntries, err = ovnParseLogEntriesFromJournald(context.TODO(), "snap.microovn.chassis.service", prefix)
+		if err != nil {
+			return "", fmt.Errorf("Failed to get OVN log entries from syslog: %v\n", err)
```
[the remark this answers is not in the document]
tomponline: We should use `%w` when wrapping errors

lxd/network/acl/driver_common.go
```diff
@@ -751,33 +751,50 @@ func (d *common) Delete() error {
...
-	}
+	var logEntries []string
+	var err error
+
+	// First, check if we can resolve the /run/openvswitch symlink to determine if OVN is in use.
+	// This is used in case of a MicroOVN deployment is interfaced with LXD.
+	targetPath, err := os.Readlink("/run/openvswitch")
+	if err == nil && strings.HasSuffix(targetPath, "/microovn/chassis/switch") {
+		prefix := fmt.Sprintf("lxd_acl%d-", d.id)
+		logEntries, err = ovnParseLogEntriesFromJournald(context.TODO(), "snap.microovn.chassis.service", prefix)
```
tomponline: Please pass in the d.state.ShutdownCtx here

lxd/network/acl/driver_common.go
```diff
@@ -751,33 +751,50 @@ func (d *common) Delete() error {
...
-	logPath := shared.HostPath("/var/log/ovn/ovn-controller.log")
-	if !shared.PathExists(logPath) {
-		return "", fmt.Errorf("Only OVN log entries may be retrieved at this time")
-	}
+	var logEntries []string
+	var err error
+
+	// First, check if we can resolve the /run/openvswitch symlink to determine if OVN is in use.
+	// This is used in case of a MicroOVN deployment is interfaced with LXD.
+	targetPath, err := os.Readlink("/run/openvswitch")
```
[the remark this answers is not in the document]
gabrielmougard: For now, there seems to be a circular dependency issue if the function is added in `lxd/network/network_utils.go` while being imported in the `lxd/network/acl` package. Could we introduce a new `lxd/network/acl/network_acl_utils.go` file and define the function there or should we do some refactoring of the `lxd/network` package ?

lxd/network/acl/driver_common.go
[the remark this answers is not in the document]
gabrielmougard: fixed

gabrielmougard:
@tomponline updated (with one question ;) ) ... and happy new year !

lxd/network/acl/acl_utils.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"os"
+	"strings"
+)
+
+// IsMicroovnUsed returns whether the current LXD deployment is using a built-in OVN switch
+// or is connected to MicroOVN, which in this case, would make `/run/openvswitch` a symlink to
+// `/var/snap/lxd/common/microovn/chassis/switch`.
+func IsMicroovnUsed() bool {
```
tomponline, line 11: Should be called `IsMicroOVNUsed()`

tomponline: Lets move this to shared/util.go alongside `InSnap()` as its detecting the use of the microovn snap.

lxd/network/acl/acl_utils.go
```diff
@@ -0,0 +1,18 @@
+package acl
+
+import (
+	"os"
+	"strings"
+)
+
+// IsMicroovnUsed returns whether the current LXD deployment is using a built-in OVN switch
```
tomponline, line 8: Lets update this comment.

lxd/network/acl/acl_ovn.go
```diff
@@ -1071,10 +1088,18 @@ func ovnParseLogEntry(input string, prefix string) string {
...
 
-	// Parse the timestamp.
-	logTime, err := time.Parse(time.RFC3339, fields[0])
-	if err != nil {
-		return ""
+	var logTime time.Time
+	var err error
+	if timestamp == "" {
+		logTime, err = parseLogTimeFromFields(fields)
+		if err != nil {
```
tomponline, line 1096: Suggest moving this if statement outside of the `if timestamp ==` statement to avoid the repetition in the else part.

lxd/network/acl/acl_ovn.go
```diff
@@ -1042,7 +1043,23 @@ type ovnLogEntry struct {
 }
 
 // ovnParseLogEntry takes a log line and expected ACL prefix and returns a re-formated log entry if matching.
-func ovnParseLogEntry(input string, prefix string) string {
+// The 'timestamp' string is in microseconds format. If empty, the timestamp is extracted from the log entry.
+func ovnParseLogEntry(input string, timestamp string, prefix string) string {
```
tomponline, line 1048: please rename `timestamp` argument to `syslogTimestamp` so its clear that this should only be populated when the logLine is from syslog.

lxd/network/acl/acl_ovn.go
```diff
@@ -1156,3 +1157,56 @@ func ovnParseLogEntry(input string, timestamp string, prefix string) string {
...
+	cmd := []string{
+		"/usr/bin/journalctl",
+		"--unit", systemdUnitName,
+		"--directory", shared.HostPath("/var/log/journal"),
+		"--no-pager",
+		"--boot", "0",
+		"--case-sensitive",
+		"--grep", prefix,
+		"--output-fields", "MESSAGE",
+		"-n", "1000",
```
tomponline, line 1173: this should be reflected in the function's description text too.

lxd/network/acl/acl_ovn.go
```diff
@@ -1156,3 +1157,56 @@ func ovnParseLogEntry(input string, timestamp string, prefix string) string {
...
+		"-o", "json",
+	}
+
+	stdout := strings.Builder{}
```
tomponline, line 1177: can we use a byte.Buffer here so we can stream the output through the json.NewDecoder below rather than reading the entire contents of the log into memory, and then converting it to a single large string on line 1183. As this does not achieve the streaming we were aiming for.

lxd/network/acl/driver_common.go
```diff
@@ -751,33 +751,47 @@ func (d *common) Delete() error {
...
-	// Open the log file.
-	logFile, err := os.Open(logPath)
-	if err != nil {
-		return "", fmt.Errorf("Couldn't open OVN log file: %w", err)
-	}
-
-	defer func() { _ = logFile.Close() }()
+	if IsMicroovnUsed() {
+		prefix := fmt.Sprintf("lxd_acl%d-", d.id)
+		logEntries, err = ovnParseLogEntriesFromJournald(context.TODO(), "snap.microovn.chassis.service", prefix)
```
tomponline: we should use d.state.ShutdownContext here.

tomponline (changes requested):
Some minor changes to make and then we are good to go, thanks!

lxd/network/acl/acl_utils.go
gabrielmougard: fixed

lxd/network/acl/acl_utils.go
gabrielmougard: fixed

lxd/network/acl/acl_ovn.go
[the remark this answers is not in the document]
gabrielmougard: fixed

lxd/network/acl/acl_ovn.go
gabrielmougard: fixed

lxd/network/acl/acl_ovn.go
```diff
@@ -1156,3 +1157,56 @@ func ovnParseLogEntry(input string, timestamp string, prefix string) string {
 
 	return string(out)
 }
+
+// ovnParseLogEntriesFromJournald reads the OVN log entries from the systemd journal and returns them as a list of string entries.
+func ovnParseLogEntriesFromJournald(ctx context.Context, systemdUnitName string, prefix string) ([]string, error) {
```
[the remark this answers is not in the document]
gabrielmougard: filter is good. Fixed.

lxd/network/acl/acl_ovn.go
gabrielmougard: fixed

lxd/network/acl/acl_ovn.go
gabrielmougard: Sure! fixed

gabrielmougard:
> LGTM thanks!
> 
> Just need to put up a PR with the snap ovn test now, based on the changes in [canonical/lxd-ci#368](https://github.com/canonical/lxd-ci/pull/368) from @simondeziel

@tomponline actually it's unclear. By the 'snap ovn test', do you mean connecting an OVN controller inside the OVN chassis snap (https://snapcraft.io/ovn-chassis) to the LXD snap and run the lxd-ci 'ovn' tests ?

tomponline:
> > LGTM thanks!
> > Just need to put up a PR with the snap ovn test now, based on the changes in [canonical/lxd-ci#368](https://github.com/canonical/lxd-ci/pull/368) from @simondeziel
> 
> @tomponline actually it's unclear. By the 'snap ovn test', do you mean connecting an OVN controller inside the OVN chassis snap (https://snapcraft.io/ovn-chassis) to the LXD snap and run the lxd-ci 'ovn' tests ?

yes, testing that this PR is working with microovn

gabrielmougard:
But this is already done in Simon's PR no ?

tomponline:
> > > LGTM thanks!
> > > Just need to put up a PR with the snap ovn test now, based on the changes in [canonical/lxd-ci#368](https://github.com/canonical/lxd-ci/pull/368) from @simondeziel
> > 
> > 
> > @tomponline actually it's unclear. By the 'snap ovn test', do you mean connecting an OVN controller inside the OVN chassis snap (https://snapcraft.io/ovn-chassis) to the LXD snap and run the lxd-ci 'ovn' tests ?
> 
> yes, testing that this PR is working with microovn

How is it passing if this isnt merged yet?

simondeziel:
Re the lxd-ci bit, it's only going to look at the ACL logs when OVN is a deb:

```
    if [ "${OVN_SOURCE:-latest/edge}" = "deb" ]; then
```

PR #14331, opened 2024-10-23
Auth: Allow updating TLS identity certificates

markylaing:
Implements the final part of the TLS fine-grained auth specification. This allows any user to update their own certificate, as well as allowing an administrator with `can_edit` on a given identity to update the certificate for that identity.

### DIFF
--- a/lxd/certificates.go
+++ b/lxd/certificates.go
@@ -1180,11 +1180,11 @@ func certificateDelete(d *Daemon, r *http.Request) response.Response {
 
 func certificateValidate(networkCert *shared.CertInfo, cert *x509.Certificate) error {
 	if time.Now().Before(cert.NotBefore) {
-		return fmt.Errorf("The provided certificate isn't valid yet")
+		return api.NewStatusError(http.StatusBadRequest, "The provided certificate isn't valid yet")
 	}
 
 	if time.Now().After(cert.NotAfter) {
-		return fmt.Errorf("The provided certificate is expired")
+		return api.NewStatusError(http.StatusBadRequest, "The provided certificate is expired")
 	}
 
 	if networkCert != nil && networkCert.CA() != nil {
@@ -1203,7 +1203,7 @@ func certificateValidate(networkCert *shared.CertInfo, cert *x509.Certificate) e
 
 		// Check that we're dealing with at least 2048bit (Size returns a value in bytes).
 		if pubKey.Size()*8 < 2048 {
-			return fmt.Errorf("RSA key is too weak (minimum of 2048bit)")
+			return api.NewStatusError(http.StatusBadRequest, "RSA key is too weak (minimum of 2048bit)")
 		}
 	}
 
--- a/lxd/identities.go
+++ b/lxd/identities.go
@@ -885,12 +885,23 @@ func getIdentities(authenticationMethod string) func(d *Daemon, r *http.Request)
 
 			apiIdentities := make([]api.Identity, 0, len(identities))
 			for _, id := range identities {
+				var certificate string
+				if id.AuthMethod == api.AuthenticationMethodTLS && id.Type != api.IdentityTypeCertificateClientPending {
+					metadata, err := id.CertificateMetadata()
+					if err != nil {
+						return response.SmartError(err)
+					}
+
+					certificate = metadata.Certificate
+				}
+
 				apiIdentities = append(apiIdentities, api.Identity{
 					AuthenticationMethod: string(id.AuthMethod),
 					Type:                 string(id.Type),
 					Identifier:           id.Identifier,
 					Name:                 id.Name,
 					Groups:               groupNamesByIdentityID[id.ID],
+					TLSCertificate:       certificate,
 				})
 			}
 
--- a/lxd/identities.go
+++ b/lxd/identities.go
@@ -137,27 +137,46 @@ const (
 	ctxClusterDBIdentity request.CtxKey = "cluster-db-identity"
 )
 
+// addIdentityDetailsToContext queries the database for the identity with the given authentication method and the
+// `nameOrIdentifier` path argument. This expands the `nameOrIdentifier` so that we can get the fully qualified URL
+// of the identity matching what is expected by the authorizer. It returns the Identity for convenience, and also adds
+// it to the request context with the ctxClusterDBIdentity context key for later use.
+func addIdentityDetailsToContext(s *state.State, r *http.Request, authenticationMethod string) (*dbCluster.Identity, error) {
+	muxVars := mux.Vars(r)
+	nameOrID, err := url.PathUnescape(muxVars["nameOrIdentifier"])
+	if err != nil {
+		return nil, fmt.Errorf("Failed to unescape path argument: %w", err)
+	}
+
+	var id *dbCluster.Identity
+	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+		id, err = dbCluster.GetIdentityByNameOrIdentifier(ctx, tx.Tx(), authenticationMethod, nameOrID)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
+	if err != nil {
+		if api.StatusErrorCheck(err, http.StatusNotFound) {
+			// Mask not found error to prevent discovery
+			return nil, api.NewGenericStatusError(http.StatusNotFound)
+		}
+
+		return nil, err
+	}
+
+	request.SetCtxValue(r, ctxClusterDBIdentity, id)
+	return id, nil
+}
+
 // identityAccessHandler performs some initial validation of the request and gets the identity by its name or
 // identifier. If one is found, the identifier is used in the URL that is passed to (auth.Authorizer).CheckPermission.
 // The cluster.Identity is set in the request context.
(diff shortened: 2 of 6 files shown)


lxd/identities.go
```diff
@@ -1201,46 +1201,154 @@ func updateIdentity(authenticationMethod string) func(d *Daemon, r *http.Request
...
+// updateIdentityUnprivileged is only invoked when an identity of type api.IdentityTypeClientCertificate updates their
+// own identity and does not have permission to change their own groups.
+func updateIdentityUnprivileged(s *state.State, r *http.Request, id dbCluster.Identity, identityPut api.IdentityPut) response.Response {
+	// Validate the given certificate
+	fingerprint, metadata, err := validateIdentityCert(s.Endpoints.NetworkCert(), identityPut.TLSCertificate)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	canViewGroup, err := s.Authorizer.GetPermissionChecker(r.Context(), auth.EntitlementCanView, entity.TypeAuthGroup)
```
tomponline, line 1243: whats this for?

lxd/identities.go
```diff
@@ -1201,46 +1201,154 @@ func updateIdentity(authenticationMethod string) func(d *Daemon, r *http.Request
...
+		if err != nil {
+			return err
+		}
+
+		// Return an error if the caller tries to update their own groups.
+		if !slices.Equal(identityPut.Groups, apiIdentity.Groups) {
+			return api.NewStatusError(http.StatusForbidden, "Only the certificate may be changed")
+		}
+
+		// If the fingerprint is identical there is nothing to do (this is checked here so that we still return an error
```
tomponline, line 1264: this comment didnt make sense to me.

lxd/identities.go
```diff
@@ -1201,46 +1201,154 @@ func updateIdentity(authenticationMethod string) func(d *Daemon, r *http.Request
...
+	}
+}
 
-			err = dbCluster.SetIdentityAuthGroups(ctx, tx.Tx(), id.ID, identityPut.Groups)
-			if err != nil {
-				return err
-			}
+// updateIdentityUnprivileged is only invoked when an identity of type api.IdentityTypeClientCertificate updates their
+// own identity and does not have permission to change their own groups.
+func updateIdentityUnprivileged(s *state.State, r *http.Request, id dbCluster.Identity, identityPut api.IdentityPut) response.Response {
```
tomponline, line 1236: shall we call this updateSelfIdentityUnprivileged?

lxd/identities.go
markylaing: The caller may not be able to view all of the groups that an identity is a member of. We need this to perform the ETag check in PUT/PATCH handlers.

lxd/identities.go
markylaing: We need to start a transaction to convert the `cluster.Identity` into an `api.Identity` and get it's list of groups. When a user without `can_edit` on the identity updates their own certificate, they cannot change the group list. We check that the groups are the same and if they are we update the certificate, but if the certificate is the same as the existing certificate there is no reason to perform the update.

lxd/identities.go
markylaing: Done

tomponline:
@markylaing is this ready for review?

markylaing:
> @markylaing is this ready for review?

Yes, apologies I thought I had clicked the "re-request review" button

lxd/identities.go
tomponline: please can you add a comment to that effect here?

lxd/identities.go
tomponline: please could you update the comment with some of this context, ta

lxd/identities.go
markylaing: Done

test/suites/auth.sh
```diff
@@ -199,6 +202,42 @@ effective_permissions: []"
...
+  [ "$(LXD_CONF="${LXD_CONF5}" lxc_remote query tls:/1.0 | jq -r '.auth')" = "trusted" ]
+
+  # Do the same tests with patch. test-user4 cannot change their group membership
+  [ "$(LXD_CONF="${LXD_CONF5}" my_curl "https://${LXD_ADDR}/1.0/auth/identities/tls/test-user4" -X PATCH --data "{\"tls_certificate\":\"$(awk '{printf "%s\\n", $0}' "${LXD_CONF4}/client.crt")\", \"groups\":[\"new-group\"]}" | jq -r '.error_code')" -eq 403 ]
+
+  # Change the certificate back to the original, using patch. Here no groups are in the request, only the certificate.
+  [ "$(LXD_CONF="${LXD_CONF5}" my_curl "https://${LXD_ADDR}/1.0/auth/identities/tls/test-user4" -X PATCH --data "{\"tls_certificate\":\"$(awk '{printf "%s\\n", $0}' "${LXD_CONF4}/client.crt")\"}" | jq -r '.status_code')" -eq 200 ]
+  [ "$(LXD_CONF="${LXD_CONF4}" lxc_remote query tls:/1.0 | jq -r '.auth')" = "trusted" ]
+  [ "$(LXD_CONF="${LXD_CONF5}" lxc_remote query tls:/1.0 | jq -r '.auth')" = "untrusted" ]
+
```
hamistao, line 240: Just a compliment: thanks for these beautiful tests and comments :D

hamistao (commented):
Looks good overall!
Most of the comments are more aimed to raise some discussion.

lxd/identities.go
```diff
@@ -136,27 +136,47 @@ const (
...
+		return nil
+	})
+	if err != nil {
+		if api.StatusErrorCheck(err, http.StatusNotFound) {
+			// Mask not found error to prevent discovery
+			return api.NewGenericStatusError(http.StatusNotFound)
+		}
+
+		return err
+	}
```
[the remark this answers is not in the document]
markylaing: This is a general pattern that we have adopted when performing database queries prior to any access control decisions (this function is called before we call the authorizer to get the fully qualified URL of the entity - e.g. expanding fingerprints/names).

The reason is that the authorizer will return a generic 404 if the caller does not have permission to view the resource. With the idea being that if they cannot view it, it shouldn't "exist" from an API perspective. If we don't mask these errors before the authorizer is called then there is a difference in the error message of the API response between an entity really not existing ("No identity found with name or identifier {x}") and an entity existing but not being allowed to view it ("Not found").

Let's say an attacker gets a list of email addresses from some server and they want to find out which of those users are using the LXD server. They could call `GET /1.0/auth/identities/oidc/{email_address}` with each email and then figure out who is using the server by the contents of the error message.

So all API responses should be masked up until an access control decision is made.

lxd/identities.go
```diff
@@ -136,27 +136,47 @@ const (
...
+
+	request.SetCtxValue(r, ctxClusterDBIdentity, id)
```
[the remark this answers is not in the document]
markylaing: Ah good spot, thanks!

lxd/certificates.go
```diff
@@ -1180,11 +1180,11 @@ func certificateDelete(d *Daemon, r *http.Request) response.Response {
 
 func certificateValidate(networkCert *shared.CertInfo, cert *x509.Certificate) error {
 	if time.Now().Before(cert.NotBefore) {
-		return fmt.Errorf("The provided certificate isn't valid yet")
+		return api.NewStatusError(http.StatusBadRequest, "The provided certificate isn't valid yet")
 	}
 
 	if time.Now().After(cert.NotAfter) {
-		return fmt.Errorf("The provided certificate is expired")
+		return api.NewStatusError(http.StatusBadRequest, "The provided certificate is expired")
```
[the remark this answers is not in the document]
markylaing: Hmm yes to the second one (RSA key too weak) but the second one is a type assertion that we don't expect to fail since we've already checked the public key algorithm of the cert, so I'll leave that as a plain error that will become a 500.

lxd/identities.go
```diff
@@ -365,29 +365,9 @@ func createIdentityTLSTrusted(ctx context.Context, s *state.State, networkCert *
...
-	// If a token is not requested, the caller must provide a certificate.
-	if req.Certificate == "" {
-		return response.BadRequest(fmt.Errorf("Must provide a certificate"))
-	}
-
-	// Parse the certificate.
-	cert, err := shared.ParseCert([]byte(req.Certificate))
+	fingerprint, metadata, err := validateIdentityCert(networkCert, req.Certificate)
 	if err != nil {
-		return response.BadRequest(fmt.Errorf("Invalid certificate material: %w", err))
```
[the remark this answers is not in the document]
markylaing: I don't think so. The CA revocation check always returned a 403, it's just that now we're surfacing it through `certificateValidate`.

test/suites/auth.sh
markylaing: Thanks :smile:

markylaing:
Thanks for the review @hamistao :) I think I've addressed everything if you want to take another look

lxd/identities.go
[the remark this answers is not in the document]
hamistao: Thanks for the elaborate and clear explanation!

hamistao (approved):
> Thanks for the review @hamistao :) I think I've addressed everything if you want to take another look

My pleasure (ﾉ◕ヮ◕)ﾉ*:･ﾟ✧

tomponline (approved):

tomponline:
thanks!

PR #14333, opened 2024-10-24
lxd: Switch to use go-dqlite v2 and dqlite lts-1.17.x (stable-5.21)

tomponline:
Also use same go version as specified in go mod file.


tomponline:
Will need to merge this to get a test run working as the go linter has picked up on unrelated code style issues.

Makefile
```diff
@@ -89,8 +90,10 @@ endif
 deps:
 	# dqlite (+raft)
 	@if [ ! -e "$(DQLITE_PATH)" ]; then \
-		git clone --depth=1 "https://github.com/canonical/dqlite" "$(DQLITE_PATH)"; \
+		echo "Retrieving dqlite from ${DQLITE_BRANCH} branch"; \
+		git clone --depth=1 --branch "${DQLITE_BRANCH}" "https://github.com/canonical/dqlite" "$(DQLITE_PATH)"; \
 	elif [ -e "$(DQLITE_PATH)/.git" ]; then \
+		echo "Updating existing dqlite branch"; \
 		cd "$(DQLITE_PATH)"; git pull; \
```
hamistao, line 97: is there a possibility that the existing dqlite clone will not be checked out to the LTS?

tomponline: yes, thats fine, i dont want to get bogged down in larger changes, and if you're working locally its nice to give devs a choice which dqlite they are using.

hamistao: indeed, good point. Thanks for clarifying :)

hamistao (approved):
Thanks for this :D

PR #14343, opened 2024-10-25
lxd: Add support for forced deletion of projects

boltmark:
Currently in draft state as this PR relies on deleting resources via requests to the server's own API. This is not ideal, and I am investigating if it is possible to avoid this. Right now it looks like this is the simplest option, but we could likely add implementation for each resource to perform the deletes without invoking the LXD API. Perhaps we could factor out some of the logic in the delete handlers so that we can reference it locally. Still looking into this -- WIP.

---

This PR adds support for forced deletion of projects. If the `--force` flag is provided when attempting to delete a project, and if the user confirms that forced deletion is what they want, then all resources associated with the project will be deleted and the project will be removed.

This PR also simplifies the `projectIsEmpty` function to utilize `projectUsedBy` instead of checking each resource manually.

Includes cherry-picks from https://github.com/lxc/incus/pull/900.

Closes: https://github.com/canonical/lxd/pull/14316.

### DIFF
--- a/client/interfaces.go
+++ b/client/interfaces.go
@@ -337,6 +337,7 @@ type InstanceServer interface {
 	UpdateProject(name string, project api.ProjectPut, ETag string) (err error)
 	RenameProject(name string, project api.ProjectPost) (op Operation, err error)
 	DeleteProject(name string) (err error)
+	DeleteProjectForce(name string) (err error)
 
 	// Storage pool functions ("storage" API extension)
 	GetStoragePoolNames() (names []string, err error)
--- a/client/lxd_projects.go
+++ b/client/lxd_projects.go
@@ -145,3 +145,18 @@ func (r *ProtocolLXD) DeleteProject(name string) error {
 
 	return nil
 }
+
+// DeleteProjectForce deletes a project and everything inside of it.
+func (r *ProtocolLXD) DeleteProjectForce(name string) error {
+	if !r.HasExtension("projects_force_delete") {
+		return fmt.Errorf("The server is missing the required \"projects_force_delete\" API extension")
+	}
+
+	// Send the request.
+	_, _, err := r.query("DELETE", fmt.Sprintf("/projects/%s?force=1", url.PathEscape(name)), nil, "")
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2535,3 +2535,8 @@ Adds a new {config:option}`device-unix-hotplug-device-conf:ownership.inherit` co
 ## `unix_device_hotplug_subsystem_device_option`
 
 Adds a new {config:option}`device-unix-hotplug-device-conf:subsystem` configuration option for `unix-hotplug` devices. This adds support for detecting `unix-hotplug` devices by subsystem, and can be used in conjunction with {config:option}`device-unix-hotplug-device-conf:productid` and {config:option}`device-unix-hotplug-device-conf:vendorid`.
+
+## `projects_force_delete`
+
+This extends `DELETE /1.0/projects` to allow `?force=true` which will
+delete everything inside of the project along with the project itself.
--- a/doc/rest-api.yaml
+++ b/doc/rest-api.yaml
@@ -14567,6 +14567,11 @@ paths:
         delete:
             description: Removes the project.
             operationId: project_delete
+            parameters:
+                - description: Delete project and related artifacts
+                  in: query
+                  name: force
+                  type: boolean
             produces:
                 - application/json
             responses:
--- a/lxc/project.go
+++ b/lxc/project.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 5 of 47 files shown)


tomponline:
@markylaing this is the PR I mentioned earlier wrt to doing loopback client requests, not @kadinsayani PR in the end. Apologies for confusion.

markylaing:
@boltmark also could you please add some tests for this. Thanks :)

boltmark:
Hi @tomponline, looping you into this question I had -- 

I'm facing a bit of an issue invoking deletion endpoints locally with dummy requests. Mainly, we currently seem to rely on everything going through the HTTP request routing/handling since we need mux to inject variables into the request context. For example, we expect to be able to access these mux vars here: https://github.com/canonical/lxd/blob/main/lxd/network_acls.go#L311.

I'm able to inject mux vars into the dummy requests via [mux.SetURLVars](https://pkg.go.dev/github.com/gorilla/mux#SetURLVars), but this function is pretty explicit about only being there for testing purposes. Am I missing something here? If not and this is in fact an issue, is some sort of refactoring of endpoints worth it for this?

CC @markylaing

markylaing:
> Hi @tomponline, looping you into this question I had --
> 
> I'm facing a bit of an issue invoking deletion endpoints locally with dummy requests. Mainly, we currently seem to rely on everything going through the HTTP request routing/handling since we need mux to inject variables into the request context. For example, we expect to be able to access these mux vars here: https://github.com/canonical/lxd/blob/main/lxd/network_acls.go#L311.
> 
> I'm able to inject mux vars into the dummy requests via [mux.SetURLVars](https://pkg.go.dev/github.com/gorilla/mux#SetURLVars), but this function is pretty explicit about only being there for testing purposes. Am I missing something here? If not and this is in fact an issue, is some sort of refactoring of endpoints worth it for this?
> 
> CC @markylaing

Hi @boltmark, sorry it's taken a while to get back to you on this one.

I agree that we shouldn't be using `mux.SetURLVars`.

I think there are two options:
1. We have a pattern of setting request details in the request context to avoid parsing data or calling the database more than once. For example, when deleting a certificate, we need to expand the certificate fingerprint before we perform the access check. The access handler sets the database entry for the certificate in the request context so that it is available in the main handler without re-querying the DB. We could adopt this pattern in all DELETE handlers such that all required information is in the request context. The project force delete code path could then set this information before calling the handlers. 
2. Refactor all required `DELETE` handlers such that they get the required mux vars and query parameters (such as `target`) and pass them in to another function that actually performs the deletion and returns an `api.StatusError`. Then in the force delete code path, call those functions rather than the handlers directly.

I think it's worth getting @tomponline thoughts on this before doing any refactoring work though.

tomponline:
> > Hi @tomponline, looping you into this question I had --
> > I'm facing a bit of an issue invoking deletion endpoints locally with dummy requests. Mainly, we currently seem to rely on everything going through the HTTP request routing/handling since we need mux to inject variables into the request context. For example, we expect to be able to access these mux vars here: https://github.com/canonical/lxd/blob/main/lxd/network_acls.go#L311.
> > I'm able to inject mux vars into the dummy requests via [mux.SetURLVars](https://pkg.go.dev/github.com/gorilla/mux#SetURLVars), but this function is pretty explicit about only being there for testing purposes. Am I missing something here? If not and this is in fact an issue, is some sort of refactoring of endpoints worth it for this?
> > CC @markylaing
> 
> Hi @boltmark, sorry it's taken a while to get back to you on this one.
> 
> I agree that we shouldn't be using `mux.SetURLVars`.
> 
> I think there are two options:
> 
>     1. We have a pattern of setting request details in the request context to avoid parsing data or calling the database more than once. For example, when deleting a certificate, we need to expand the certificate fingerprint before we perform the access check. The access handler sets the database entry for the certificate in the request context so that it is available in the main handler without re-querying the DB. We could adopt this pattern in all DELETE handlers such that all required information is in the request context. The project force delete code path could then set this information before calling the handlers.
> 
>     2. Refactor all required `DELETE` handlers such that they get the required mux vars and query parameters (such as `target`) and pass them in to another function that actually performs the deletion and returns an `api.StatusError`. Then in the force delete code path, call those functions rather than the handlers directly.
> 
> 
> I think it's worth getting @tomponline thoughts on this before doing any refactoring work though.

I dont think we should be bundling everything into the request context, especially stuff thats not altered from the request itself. Is that what you were meaning for 1?

It sounds like a combination of 1 (to reduce DB queries if needed) and 2 is the way to go.
But we should check whether we should be spending more time on this now or whether to park this or land as-is.

@escabo what would you like us to do with this one?

markylaing:
> I dont think we should be bundling everything into the request context, especially stuff thats not altered from the request itself. Is that what you were meaning for 1?

Not exactly, I was thinking to only add required information per-handler call. So rather than setting in the context of the incoming request, passing `request.WithContext(<deletion args>)` into the call to the appropriate handler. 

As you say though, a combination of 1 and 2 is likely needed.

> But we should check whether we should be spending more time on this now or whether to park this or land as-is.

I don't think we should land as is. We should at least iterate over the project used-by list and check that the caller has permission to delete the resource before calling back out via unix socket (to ourselves) to perform the deletion. Also, we should have a discussion as to whether we should ignore the `security.protection.delete` configuration when force deleting. Alternatively, we can add a new entitlement for this: `can_force_delete` on `project`.

tomponline:
@markylaing thanks.

@escabo this cherry-pick PR has escalated, is it OK to continue or shall we park for now?

escabo:
Sorry I missed the first ping, l have added this topic to next week's work
session.

PR #14347, opened 2024-10-25
Support NVIDIA CDI in Ubuntu Core environment

mihalicyn:
Tested using testflinger with `core24-latest` image deployed.

### DIFF
--- a/lxd/device/cdi/spec.go
+++ b/lxd/device/cdi/spec.go
@@ -13,6 +13,7 @@ import (
 	"tags.cncf.io/container-device-interface/specs-go"
 
 	"github.com/canonical/lxd/lxd/instance"
+	"github.com/canonical/lxd/lxd/state"
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/logger"
 )
@@ -39,7 +40,7 @@ func defaultNvidiaTegraCSVFiles(rootPath string) []string {
 }
 
 // generateNvidiaSpec generates a CDI spec for an Nvidia vendor.
-func generateNvidiaSpec(cdiID ID, inst instance.Instance) (*specs.Spec, error) {
+func generateNvidiaSpec(s *state.State, cdiID ID, inst instance.Instance) (*specs.Spec, error) {
 	l := logger.AddContext(logger.Ctx{"instanceName": inst.Name(), "projectName": inst.Project().Name, "cdiID": cdiID.String()})
 	mode := nvcdi.ModeAuto
 	if cdiID.Class == IGPU {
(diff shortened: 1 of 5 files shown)


lxd/device/cdi/spec.go
```diff
@@ -62,15 +63,26 @@ func generateNvidiaSpec(cdiID ID, inst instance.Instance) (*specs.Spec, error) {
...
 
 	rootPath := ""
-	if shared.InSnap() {
+	devRootPath := ""
+	if s.OS.InUbuntuCore() {
+		//
+		// This magic "gpu-2404-2" name comes from:
+		// https://github.com/canonical/mesa-2404/blob/0e48b4d1b8e5cb4d3098d64417025824decd9846/scripts/bin/gpu-2404-provider-wrapper.in#L5
+		//
+		// Also, we can't use `/snap/lxd/current/gpu-2404-2` as a rootPath because of a bug in nvcdi package.
```
tomponline, line 72: is there a link to the bug in question you could add here?

mihalicyn: > is there a link to the bug in question you could add here?

there is no buglink. That's what I found during my investigation of nvcdi internals.

tomponline (approved):

lxd/device/cdi/spec.go
tomponline: OK thanks, would it be worth opening an issue on their Github page?

lxd/instance/instance_utils.go
```diff
@@ -112,6 +112,10 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
 		return fmt.Errorf("nvidia.runtime is incompatible with privileged containers")
 	}
 
+	if sysOS.InUbuntuCore() && shared.IsTrue(config["nvidia.runtime"]) {
+		return fmt.Errorf("nvidia.runtime is incompatible with Ubuntu Core")
```
tomponline, line 116: @mihalicyn is nvidia.runtime compatible with CDI mode in general? I'm wondering if we need this check specifically (although happy to keep it if CDI does work with nvidia.runtime).

lxd/device/cdi/spec.go
mihalicyn: > OK thanks, would it be worth opening an issue on their Github page?

yeah, definitely, but I decided to go with a quick and simple workaround from LXD side at first.

I'll fill an issue or even put up a PR for nvidia thing a bit later.

lxd/instance/instance_utils.go
mihalicyn: As far as I understand `nvidia.runtime=true` thing and CDI thing are absolute independent from each other. `nvidia.runtime=true` uses https://github.com/NVIDIA/libnvidia-container and https://github.com/lxc/lxc/blob/main/hooks/nvidia hook but I wasn't able to make it work (and stopped trying as we decided to go with a CDI thing, i.e. https://github.com/NVIDIA/nvidia-container-toolkit).

I guess this check is really useful as I'm pretty sure that Ubuntu Core users will try to set `nvidia.runtime=true` and it will silently fail and all NVIDIA tooling won't be available inside the container. IMHO, it's better to explicitly say that it does not supported at all.

tomponline: i agree, but should we also refuse to support it for CDI mode too?

tomponline: @gabrielmougard should nvidia.runtime be blocked from being set when using CDI mode?

gabrielmougard: > @gabrielmougard should nvidia.runtime be blocked from being set when using CDI mode?

Yes I think so. As @mihalicyn said, the CDI approach and the LXC hook called if `nvidia.runtime=true` are independent. Plus, having both option set together could potentially introduce linking issues: both approaches update the container `ldcache`, so if libnvidia (used by the legacy NVIDIA hook for LXC) hasn't the same library lookup mechanisn than nvctk (CDI), it could override the ldcache config and break things inside the container that will only be seen at runtime (when executing a CUDA workload for example (something like `libnvml not found` for example))

gabrielmougard: It's safer to restrict the user from setting CDI mode and nvidia.runtime=true ... It'll avoid these nasty runtime bugs.

mihalicyn: Thanks!

Let me add some sanity logic around that then.

tomponline: Thanks @gabrielmougard 
And im happy to keep the existing check for ubuntu core and nvidia runtime too @mihalicyn

tomponline (approved):
I'll merge this now to aid with testing next week.

But if you could follow up with another PR wrt to the validation discussion that would be great. Ta

PR #14371, opened 2024-10-30
lxd: Deduplicate logic for deletion of main image file

boltmark:
This small PR removes the duplicated logic in the `imageDeleteFromDisk` function.

### DIFF
--- a/lxd/images.go
+++ b/lxd/images.go
@@ -2705,17 +2705,9 @@ func pruneExpiredImages(ctx context.Context, s *state.State, op *operations.Oper
 		}
 
 		// Remove main image file.
-		fname := filepath.Join(s.OS.VarDir, "images", fingerprint)
-		err = os.Remove(fname)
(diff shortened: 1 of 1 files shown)


lxd/images.go
```diff
@@ -2886,7 +2878,10 @@ func imageDelete(d *Daemon, r *http.Request) response.Response {
 		}
 
 		// Remove main image file from disk.
-		imageDeleteFromDisk(details.image.Fingerprint)
+		err = imageDeleteFromDisk(details.image.Fingerprint)
```
tomponline, line 2881: We should probably remove the DB record above last otherwise if the image delete fails the image record has now been removed so we can retry it.

boltmark: Thanks, I've moved the call to `imageDeleteFromDisk` before the DB record removal

tomponline (approved):

Issue #14266, opened 2024-10-13
Inconsistency in unix-hotplug when multiple devices share product/vendor IDs


jonathan-conder:
# Required information

 * Distribution: Ubuntu
 * Distribution version: 24.04.1
 * The output of "snap list --all lxd core20 core22 core24 snapd":
```
Name    Version         Rev    Tracking       Publisher   Notes
core20  20240705        2379   latest/stable  canonical✓  base
core22  20240904        1621   latest/stable  canonical✓  base
lxd     5.21.2-2f4ba6b  30131  5.21/stable    canonical✓  -
```
 * The output of "lxc info" or if that fails:
```
config: {}
api_extensions:
- storage_zfs_remove_snapshots
- container_host_shutdown_timeout
- container_stop_priority
- container_syscall_filtering
- auth_pki
- container_last_used_at
- etag
- patch
- usb_devices
- https_allowed_credentials
- image_compression_algorithm
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
- unix_device_rename
- storage_lvm_use_thinpool
- storage_rsync_bwlimit
- network_vxlan_interface
- storage_btrfs_mount_options
- entity_description
- image_force_refresh
- storage_lvm_lv_resizing
- id_map_base
- file_symlinks
- container_push_target
- network_vlan_physical
- storage_images_delete
- container_edit_metadata
- container_snapshot_stateful_migration
- storage_driver_ceph
- storage_ceph_user_name
- resource_limits
- storage_volatile_initial_source
- storage_ceph_force_osd_reuse
- storage_block_filesystem_btrfs
- resources
- kernel_limits
- storage_api_volume_rename
- network_sriov
- console
- restrict_devlxd
- migration_pre_copy
- infiniband
- maas_network
- devlxd_events
- proxy
- network_dhcp_gateway
- file_get_symlink
- network_leases
- unix_device_hotplug
- storage_api_local_volume_handling
- operation_description
[...]

Issue #14373, opened 2024-10-30
Feature request: filter by subsystem for unix-hotplug devices


jonathan-conder:
Currently `unix-hotplug` devices can be detected based on `productid` and `vendorid`. It would be great if this could be extended to include `subsystem` as well.

For example, it would be really convenient to allow containers to access all camera devices with a single command:
```console
lxc config device add u1 cameras unix-hotplug subsystem=video4linux
```
Or all cameras with a specific `vendorid`, and so on:
```console
lxc config device add u1 logitechcameras unix-hotplug subsystem=video4linux vendorid=****
```

Some devices don't currently have product and vendor IDs, so this change would enable those as well. For example fake `v4l2loopback` devices are useful for testing. My camera also exposes `/dev/media0`, which belongs to the `media` subsystem but lacks a product/vendor ID.

PR #6530, opened 2019-11-29
Uevent-based matching for `unix-char` and `unix-block`

Lilyjjo:
For https://github.com/lxc/lxd/issues/1750

PR #14375, opened 2024-10-30
Device: Add support for discovering multiple `unix-hotplug` devices

kadinsayani:
Fixes 1/2 related issues raised in https://github.com/canonical/lxd/issues/14266. This fix has been tested locally with a USB hard disk and by @jonathan-conder with USB camera devices.

Summary of changes:
- Adds support for discovering multiple `unix-hotplug` devices which share the same `productid` and/or `vendorid`;
- Improves `productid` and `vendorid` matching logic;
- Resolves inconsistent device matching bug.

### DIFF
--- a/lxd/device/unix_hotplug.go
+++ b/lxd/device/unix_hotplug.go
@@ -89,8 +89,7 @@ func (d *unixHotplug) validateConfig(instConf instance.ConfigReader) error {
 
 // Register is run after the device is started or when LXD starts.
 func (d *unixHotplug) Register() error {
-	// Extract variables needed to run the event hook so that the reference to this device
-	// struct is not needed to be kept in memory.
+	// Extract variables needed to run the event hook so that the reference to this device struct is not required to be stored in memory.
 	devicesPath := d.inst.DevicesPath()
 	devConfig := d.config
 	deviceName := d.name
@@ -212,8 +211,7 @@ func (d *unixHotplug) postStop() error {
 	return nil
 }
 
-// loadUnixDevices scans the host machine for unix devices with matching product/vendor ids
-// and returns the matching devices with subsystem types of char or block.
+// loadUnixDevices scans the host machine for unix devices with matching product/vendor ids and returns the matching devices with subsystem types of char or block.
 func (d *unixHotplug) loadUnixDevices() []udev.Device {
 	// Find device if exists
 	u := udev.Udev{}
--- a/lxd/device/unix_hotplug.go
+++ b/lxd/device/unix_hotplug.go
@@ -149,29 +149,38 @@ func (d *unixHotplug) Start() (*deviceConfig.RunConfig, error) {
 	runConf := deviceConfig.RunConfig{}
 	runConf.PostHooks = []func() error{d.Register}
 
-	device := d.loadUnixDevice()
-	if d.isRequired() && device == nil {
-		return nil, fmt.Errorf("Required Unix Hotplug device not found")
+	devices := d.loadUnixDevices()
+	if d.isRequired() && len(devices) <= 0 {
+		return nil, fmt.Errorf("Required unix hotplug device not found")
 	}
 
-	if device == nil {
-		return &runConf, nil
-	}
+	for _, device := range devices {
+		devnum := device.Devnum()
+		major := uint32(devnum.Major())
+		minor := uint32(devnum.Minor())
 
-	devnum := device.Devnum()
-	major := uint32(devnum.Major())
-	minor := uint32(devnum.Minor())
+		// Setup device.
+		var err error
+		if device.Subsystem() == "block" {
+			err = unixDeviceSetupBlockNum(d.state, d.inst.DevicesPath(), "unix", d.name, d.config, major, minor, device.Devnode(), false, &runConf)
 
-	// setup device
-	var err error
-	if device.Subsystem() == "block" {
-		err = unixDeviceSetupBlockNum(d.state, d.inst.DevicesPath(), "unix", d.name, d.config, major, minor, device.Devnode(), false, &runConf)
-	} else {
-		err = unixDeviceSetupCharNum(d.state, d.inst.DevicesPath(), "unix", d.name, d.config, major, minor, device.Devnode(), false, &runConf)
-	}
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			err = unixDeviceSetupCharNum(d.state, d.inst.DevicesPath(), "unix", d.name, d.config, major, minor, device.Devnode(), false, &runConf)
 
-	if err != nil {
-		return nil, err
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		// Remove unix device on failure to setup device.
+		runConf.Revert = func() { _ = unixDeviceRemove(d.inst.DevicesPath(), "unix", d.name, "", &runConf) }
+
+		if err != nil {
+			return nil, fmt.Errorf("Unable to setup unix hotplug device: %w", err)
+		}
 	}
 
 	return &runConf, nil
@@ -203,9 +212,9 @@ func (d *unixHotplug) postStop() error {
 	return nil
 }
 
-// loadUnixDevice scans the host machine for unix devices with matching product/vendor ids
-// and returns the first matching device with the subsystem type char or block.
-func (d *unixHotplug) loadUnixDevice() *udev.Device {
+// loadUnixDevices scans the host machine for unix devices with matching product/vendor ids
+// and returns the matching devices with subsystem types of char or block.
+func (d *unixHotplug) loadUnixDevices() []udev.Device {
 	// Find device if exists
 	u := udev.Udev{}
 	e := u.NewEnumerate()
@@ -230,27 +239,35 @@ func (d *unixHotplug) loadUnixDevice() *udev.Device {
 	}
 
 	devices, _ := e.Devices()
-	var device *udev.Device
+	var matchingDevices []udev.Device
 	for i := range devices {
-		device = devices[i]
+		device := devices[i]
 
-		if device == nil {
+		// Ignore USB bus devices (handled by `usb` device type) since we don't want `unix-hotplug` and `usb` devices conflicting. We want to add all device nodes besides those with a `usb` subsystem.
+		// We ignore devices with a major number of 0, since this indicates they are unnamed devices (e.g. non-device mounts).
+		// We ignore devices without an associated device node file name, as this indicates they are not accessible via the standard interface in /dev/.
+		if device == nil || device.Devnum().Major() == 0 || device.Devnode() == "" || strings.HasPrefix(device.Subsystem(), "usb") {
 			continue
 		}
 
-		devnum := device.Devnum()
-		if devnum.Major() == 0 || devnum.Minor() == 0 {
+		match := false
+		vendorIDMatch := device.PropertyValue("ID_VENDOR_ID") == d.config["vendorid"]
+		productIDMatch := device.PropertyValue("ID_MODEL_ID") == d.config["productid"]
+
+		if d.config["vendorid"] != "" && d.config["productid"] != "" {
+			match = vendorIDMatch && productIDMatch
+		} else if d.config["vendorid"] != "" {
+			match = vendorIDMatch
+		} else if d.config["productid"] != "" {
+			match = productIDMatch
+		}
+
+		if !match {
 			continue
 		}
 
-		if device.Devnode() == "" {
-			continue
-		}
-
-		if !strings.HasPrefix(device.Subsystem(), "usb") {
-			return device
-		}
+		matchingDevices = append(matchingDevices, *device)
 	}
 
-	return nil
+	return matchingDevices
 }
--- a/lxd/device/unix_hotplug.go
+++ b/lxd/device/unix_hotplug.go
@@ -16,16 +16,27 @@ import (
 	"github.com/canonical/lxd/shared/validate"
 )
 
-// unixHotplugIsOurDevice indicates whether the unixHotplug device event qualifies as part of our device.
-// This function is not defined against the unixHotplug struct type so that it can be used in event
-// callbacks without needing to keep a reference to the unixHotplug device struct.
-func unixHotplugIsOurDevice(config deviceConfig.Device, unixHotplug *UnixHotplugEvent) bool {
-	// Check if event matches criteria for this device, if not return.
-	if (config["vendorid"] != "" && config["vendorid"] != unixHotplug.Vendor) || (config["productid"] != "" && config["productid"] != unixHotplug.Product) {
+// unixHotplugDeviceMatch matches a unix-hotplug device based on vendorid and productid. USB bus and devices with a major number of 0 are ignored. This function is used to indicate whether a unix hotplug event qualifies as part of our registered devices, and to load matching devices.
+func unixHotplugDeviceMatch(config deviceConfig.Device, vendorid string, productid string, subsystem string, major uint32) bool {
+	match := false
+	vendorIDMatch := vendorid == config["vendorid"]
+	productIDMatch := productid == config["productid"]
+
+	// Ignore USB bus devices (handled by `usb` device type) since we don't want `unix-hotplug` and `usb` devices conflicting. We want to add all device nodes besides those with a `usb` subsystem.
+	// We ignore devices with a major number of 0, since this indicates they are unnamed devices (e.g. non-device mounts).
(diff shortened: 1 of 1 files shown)


lxd/device/unix_hotplug.go
```diff
@@ -149,29 +149,31 @@ func (d *unixHotplug) Start() (*deviceConfig.RunConfig, error) {
...
+		if device.Subsystem() == "block" {
+			err = unixDeviceSetupBlockNum(d.state, d.inst.DevicesPath(), "unix", d.name, d.config, major, minor, device.Devnode(), false, &runConf)
+		} else {
+			err = unixDeviceSetupCharNum(d.state, d.inst.DevicesPath(), "unix", d.name, d.config, major, minor, device.Devnode(), false, &runConf)
+		}
 
-	if err != nil {
-		return nil, err
+		if err != nil {
+			return nil, err
```
tomponline, line 175: Do we need to consider undoing any previously setup devices if one of the subsequent device setups fails here?

tomponline: I.e. doing the equivalent of when the instance is stopped.

lxd/device/unix_hotplug.go
```diff
@@ -230,6 +232,7 @@ func (d *unixHotplug) loadUnixDevice() *udev.Device {
 	}
 
 	devices, _ := e.Devices()
+	var matchingDevices []*udev.Device
 	var device *udev.Device
```
tomponline, line 236: I wonder why device var here is outside of the for loop, it should be just:

```go
device := devices[i]
```

kadinsayani: Agreed

lxd/device/unix_hotplug.go
tomponline: But ur doing this in a for loop which means we need to undo files created in earlier iterations if we encounter an error and return

kadinsayani: Gotcha, makes sense!

kadinsayani:
I was trying to reason about why the initial implementation of `loadUnixDevice()` only matched the first device with a matching product and vendor id since this logic is explicitly described in the go doc comment. I found the original PR and this comment: https://github.com/canonical/lxd/pull/6530#issuecomment-561348982. 

I don't think the approach in this PR is quite right. Furthermore, wrt https://github.com/canonical/lxd/issues/14373, I found this comment which provides reason to not add support for specifying subsystem types: https://github.com/canonical/lxd/pull/6530#issuecomment-561466424.

We'll have to rethink our solution here.

tomponline:
> I was trying to reason about why the initial implementation of `loadUnixDevice()` only matched the first device with a matching product and vendor id since this logic is explicitly described in the go doc comment. I found the original PR and this comment: [#6530 (comment)](https://github.com/canonical/lxd/pull/6530#issuecomment-561348982).
> 
> I don't think the approach in this PR is quite right. Furthermore, wrt #14373, I found this comment which provides reason to not add support for specifying subsystem types: [#6530 (comment)](https://github.com/canonical/lxd/pull/6530#issuecomment-561466424).
> 
> We'll have to rethink our solution here.

Which part of the comment makes you think multiple devices shouldn't be allowed?

kadinsayani:
> Which part of the comment makes you think multiple devices shouldn't be allowed?

> If we don't specify a subsystem we end up enumerating multiple devices in loadUnixDevice(). For example, if we plug in a usb with a FS on it, we get a device for the usb itself and then a device for the sda device node entry as well since they share the same product/vendor IDs

kadinsayani:
Looks like this was an issue raised in the original PR but was never actually addressed properly:

> The partitions (sda1, sda2 and sda3) only showed up during USB hotplug, on container start or when adding the device, only sda shows up

Also, this [commit](https://github.com/canonical/lxd/commit/a10be101fa4d9a0aedc9a24116d16df32c6f4457) added logic to not add a device with a minor number of 0, but for most usb devices, the first sub device has a minor number of 0.

kadinsayani:
Thanks to @jonathan-conder for testing this locally with his camera devices. The camera devices showed up as expected, but a mouse with a matching vendor id also appeared. This is because when `udev_enumerate_add_match_property` is called more than once, each matching result is ORed - only at least one property needs to match.

It's now clear why this [commit](https://github.com/canonical/lxd/commit/a10be101fa4d9a0aedc9a24116d16df32c6f4457) added logic to not add a device with a major or minor number of 0. However, this approach doesn't allow all devices with a matching product or vendor id to be captured. For example, if we ignore devices with a minor number of 0, we may see cases like this:

```
$ ls -l /dev/video*
crw-rw----+ 1 root video 81, 0 Oct 19 05:29 /dev/video0
crw-rw----+ 1 root video 81, 1 Oct 19 05:29 /dev/video1
crw-rw----+ 1 root video 81, 2 Oct 19 05:41 /dev/video2
crw-rw----+ 1 root video 81, 3 Oct 19 05:41 /dev/video3
$ lxc exec u1 -- sh -c 'ls -l /dev/video*'
crw-rw---- 1 root root  81, 1 Oct 18 16:39 /dev/video1
crw-rw---- 1 root video 81, 2 Oct 18 16:41 /dev/video2
crw-rw---- 1 root video 81, 3 Oct 18 16:41 /dev/video3
```

We should see `/dev/video[0-3]`.

I think the best approach is to require specifying subsystem type along with either vendor or product id. Requiring subsystem type to be specified will lead to less "edge cases". @tomponline, what do you think?

kadinsayani:
Or, we just need a way to filter on both vendor id and product id. This would also be helpful for https://github.com/canonical/lxd/issues/14373 so we could filter on subsystem && vendorid. I'll look into this further.

kadinsayani:
I believe I've got a fix working. Just waiting on a test with camera devices and then this PR will be good to go.

kadinsayani:
The fix is now confirmed to be working. We've noticed another inconsistency where `/dev/bus/usb/...` devices are excluded by what's added by `unix-hotplug`, however, the device(s) appear after replugging the device. [This comment](https://github.com/canonical/lxd/pull/6530#issuecomment-561466424) provides some insight into the original decision for excluding `/dev/bus/usb/...` devices.

jonathan-conder:
As Kadin says, I'm happy with this version. For our use case it doesn't matter whether the USB device is added or not, but based on that comment I agree it makes sense to avoid adding the USB device when replugging

kadinsayani:
I've added a commit which resolves the inconsistent device matching bug. `/dev/bus/usb/...` will be ignored when `unix-hotplug` devices are registered and started (replugged or after container restart). 

@tomponline this PR now fully resolves the issue.

lxd/device/unix_hotplug.go
```diff
@@ -149,29 +148,36 @@ func (d *unixHotplug) Start() (*deviceConfig.RunConfig, error) {
...
 
-	device := d.loadUnixDevice()
-	if d.isRequired() && device == nil {
+	devices := d.loadUnixDevices()
+	if d.isRequired() && devices == nil {
 		return nil, fmt.Errorf("Required Unix Hotplug device not found")
 	}
 
-	if device == nil {
+	if devices == nil {
```
tomponline, line 165: do we need this anymore? If the devices var is nil or zero length then the for loop wont iterate it and the return statement at the bottom will run as usual.

lxd/device/unix_hotplug.go
```diff
@@ -149,29 +148,36 @@ func (d *unixHotplug) Start() (*deviceConfig.RunConfig, error) {
 	runConf := deviceConfig.RunConfig{}
 	runConf.PostHooks = []func() error{d.Register}
 
-	device := d.loadUnixDevice()
-	if d.isRequired() && device == nil {
+	devices := d.loadUnixDevices()
+	if d.isRequired() && devices == nil {
```
tomponline, line 161: Can we do `if d.isRequired() && len(devices) <= 0 {` to catch the scenario where devices is non-nil but zero length?

lxd/device/unix_hotplug.go
```diff
@@ -149,29 +148,36 @@ func (d *unixHotplug) Start() (*deviceConfig.RunConfig, error) {
...
 
-	// setup device
-	var err error
-	if device.Subsystem() == "block" {
-		err = unixDeviceSetupBlockNum(d.state, d.inst.DevicesPath(), "unix", d.name, d.config, major, minor, device.Devnode(), false, &runConf)
-	} else {
-		err = unixDeviceSetupCharNum(d.state, d.inst.DevicesPath(), "unix", d.name, d.config, major, minor, device.Devnode(), false, &runConf)
-	}
+		if err != nil {
+			err := unixDeviceRemove(d.inst.DevicesPath(), "unix", d.name, "", &runConf)
```
tomponline, line 183: Please can you use the `revert` package here.

This makes it easier to maintain in the future because we wont have to worry about reverting every step manually on any new error cases.

lxd/device/unix_hotplug.go
```diff
@@ -203,9 +209,9 @@ func (d *unixHotplug) postStop() error {
 	return nil
 }
 
-// loadUnixDevice scans the host machine for unix devices with matching product/vendor ids
-// and returns the first matching device with the subsystem type char or block.
-func (d *unixHotplug) loadUnixDevice() *udev.Device {
+// loadUnixDevices scans the host machine for unix devices with matching product/vendor ids
+// and returns the matching devices with subsystem types of char or block.
+func (d *unixHotplug) loadUnixDevices() []*udev.Device {
```
tomponline, line 223: can we return just `[]udev.Device` ?

lxd/device/unix_hotplug.go
```diff
@@ -230,27 +236,32 @@ func (d *unixHotplug) loadUnixDevice() *udev.Device {
...
 
 	devices, _ := e.Devices()
-	var device *udev.Device
+	var matchingDevices []*udev.Device
 	for i := range devices {
-		device = devices[i]
+		device := devices[i]
 
-		if device == nil {
+		if device == nil || device.Devnum().Major() == 0 || device.Devnode() == "" || strings.HasPrefix(device.Subsystem(), "usb") {
```
tomponline: Please can you add a comment here explaining the *why* for skipping in these cases.

lxd/device/unix_hotplug.go
```diff
@@ -230,27 +236,32 @@ func (d *unixHotplug) loadUnixDevice() *udev.Device {
...
 			continue
 		}
 
-		devnum := device.Devnum()
-		if devnum.Major() == 0 || devnum.Minor() == 0 {
-			continue
+		match := false
+		vendorIDMatch := device.PropertyValue("ID_VENDOR_ID") == d.config["vendorid"]
```
tomponline: why is this needed when there is apparently filtering being setup via `e.AddMatchProperty` above?

tomponline (changes requested):
A few minor tweaks and a question about filtering. Thanks

lxd/device/unix_hotplug.go
kadinsayani: `e.AddMatchProperty` calls `udev_enumerate_add_match_property` under the hood. Each matching result is ORed - only at least one property needs to match.

lxd/device/unix_hotplug.go
kadinsayani: You're right, we don't need this check anymore. Good catch.

lxd/device/unix_hotplug.go
kadinsayani: Sure thing 👍

lxd/device/unix_hotplug.go
kadinsayani: Sounds good

lxd/device/unix_hotplug.go
kadinsayani: Yes, we can just just dereference `device` when we append it to `matchingDevices`.

lxd/device/unix_hotplug.go
kadinsayani: Sure thing!

lxd/device/unix_hotplug.go
kadinsayani: From the udevadm man page:

```
-p, --property-match=PROPERTY=VALUE
           Trigger events for devices with a matching property value. This option supports shell style pattern matching. When this option is specified more than once, then each matching result is
           ORed, that is, devices which have one of the specified properties are triggered.
```

I've removed the redundant `e.AddMatchProperty` calls since we have the new consolidated matching logic now.

tomponline: Ok i see. The AddMatchProperty calls should remain though to provide early push down filtering.

kadinsayani: Sounds good - I've added them back.

lxd/device/unix_hotplug.go
```diff
@@ -149,31 +149,42 @@ func (d *unixHotplug) Start() (*deviceConfig.RunConfig, error) {
...
-	if device.Subsystem() == "block" {
-		err = unixDeviceSetupBlockNum(d.state, d.inst.DevicesPath(), "unix", d.name, d.config, major, minor, device.Devnode(), false, &runConf)
-	} else {
-		err = unixDeviceSetupCharNum(d.state, d.inst.DevicesPath(), "unix", d.name, d.config, major, minor, device.Devnode(), false, &runConf)
-	}
+		// Remove unix device on failure to setup device.
+		revert.Add(func() {
+			err = unixDeviceRemove(d.inst.DevicesPath(), "unix", d.name, "", &runConf)
+			if err != nil {
+				d.logger.Error("Failed to remove unix device", logger.Ctx{"device": d.name, "err": err})
```
tomponline, line 188: Perhaps "Failed to revert..." or "Failed to clean up..." would be clearer here, as otherwise users may think the LXD instance device itself was being removed even though it technically never got started. Also "removed" indicates the device was trying to be "added", but that is not whats happening from a user perspective, the device is actually being "started".

lxd/device/unix_hotplug.go
```diff
@@ -149,31 +149,42 @@ func (d *unixHotplug) Start() (*deviceConfig.RunConfig, error) {
 	runConf := deviceConfig.RunConfig{}
 	runConf.PostHooks = []func() error{d.Register}
 
-	device := d.loadUnixDevice()
-	if d.isRequired() && device == nil {
-		return nil, fmt.Errorf("Required Unix Hotplug device not found")
+	devices := d.loadUnixDevices()
+	if d.isRequired() && len(devices) <= 0 {
+		return nil, fmt.Errorf("Required unix hotplug device %q not found", d.name)
```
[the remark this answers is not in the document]
tomponline: Ah wait, GH was munging the two changes together, this is in Start().

Still, I don't think we should be including the device name in the error.

Because we can see from the call site, returns like this:

```
return nil, fmt.Errorf("Failed to start device %q: %w", dev.Name(), err)
```

lxd/device/unix_hotplug.go
```diff
@@ -149,31 +149,42 @@ func (d *unixHotplug) Start() (*deviceConfig.RunConfig, error) {
...
+			err = unixDeviceRemove(d.inst.DevicesPath(), "unix", d.name, "", &runConf)
+			if err != nil {
+				d.logger.Error("Failed to remove unix device", logger.Ctx{"device": d.name, "err": err})
+			}
+		})
 
-	if err != nil {
-		return nil, err
+		if err != nil {
+			return nil, fmt.Errorf("Unable to setup unix hotplug device %q: %w", d.name, err)
```
tomponline, line 193: No need for including device name in error, see above.

lxd/device/unix_hotplug.go
```diff
@@ -230,27 +241,33 @@ func (d *unixHotplug) loadUnixDevice() *udev.Device {
...
 	devices, _ := e.Devices()
-	var device *udev.Device
+	var matchingDevices []udev.Device
 	for i := range devices {
-		device = devices[i]
+		device := devices[i]
 
-		if device == nil {
+		// Ignore USB bus devices (handled by `usb` device type) and devices without a valid devpath or major number.
+		if device == nil || device.Devnum().Major() == 0 || device.Devnode() == "" || strings.HasPrefix(device.Subsystem(), "usb") {
```
tomponline: youve added a comment for the *what* but not yet the *why* of the skipping logic here.

tomponline: the aim here is to document the git archeology you've spent a lot of time on to save future readers the work.

lxd/device/unix_hotplug.go
kadinsayani: Agreed.

tomponline (approved):

PR #14402, opened 2024-11-05
Fix/ss creation date

MusicDin:
Parse creation date in 24-hour format instead of 12 hour format where am/pm is not differentiated. Also replace usage of `ValueInSlice` with `slices.Contains`.

Based on comment: https://github.com/canonical/lxd/commit/4b26fc21f402b20286a31c89e145092d59f8de01#commitcomment-148595914


tomponline (approved):

tomponline:
Ta!

PR #14403, opened 2024-11-05
Fix handling of long running connections (from Incus)

markylaing:
When connecting to LXD via a load-balancer, long-polling LXD often fails because the connection is timing out. This PR mitigates the issue by adding a keep-alive timeout on the LXD server.

### DIFF
--- a/lxd/api.go
+++ b/lxd/api.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/gorilla/mux"
 
@@ -223,6 +224,7 @@ func restServer(d *Daemon) *http.Server {
 	return &http.Server{
 		Handler:     &lxdHTTPServer{r: mux, d: d},
 		ConnContext: lxdRequest.SaveConnectionInContext,
+		IdleTimeout: 30 * time.Second,
 	}
 }
 


markylaing:
The CLA checker failed a few times because I used the British spelling `Licence` instead of `License` :facepalm:

tomponline:
@markylaing would you mind adding some additional info to the PR description to describe what "fix" means in this context? Thanks

markylaing:
> @markylaing would you mind adding some additional info to the PR description to describe what "fix" means in this context? Thanks

Yep of course, apologies

markylaing:
@tomponline I've updated the description. I've also removed the second commit related to OIDC. I'm not aware of this causing any issues and it's an easy fix if it is ever reported.

tomponline:
> When connecting to LXD via a load-balancer, long-polling LXD often fails because the connection is timing out. This PR mitigates the issue by adding a keep-alive timeout on the LXD server.

Do you know what the default IdleTimeout value was?

markylaing:
> > When connecting to LXD via a load-balancer, long-polling LXD often fails because the connection is timing out. This PR mitigates the issue by adding a keep-alive timeout on the LXD server.
> 
> Do you know what the default IdleTimeout value was?

The go doc has:
```
// IdleTimeout is the maximum amount of time to wait for the
// next request when keep-alives are enabled. If IdleTimeout
// is zero, the value of ReadTimeout is used. If both are
// zero, there is no timeout.
```

The ReadTimeout is not set so it's currently defaulting to zero.

tomponline:
> The ReadTimeout is not set so it's currently defaulting to zero.

So if not timeout now, then how come "long-polling LXD often fails because the connection is timing out. "?

markylaing:
I'm going to close this because I can't see any issues relating to what it purports to fix.

PR #14404, opened 2024-11-05
lxd: Expose `IsFineGrained` to `api.IdentityInfo`

gabrielmougard:
JIRA ticket: https://warthogs.atlassian.net/browse/LXD-2047

This is a first PR to expose if the currently used identity is fine-grained.


markylaing:
I can't remember, did we agree that this will require a specification? We'll need to figure out how to efficiently return all this data

tomponline:
@gabrielmougard @markylaing lets get a spec sorted out for the entitlements list (formerly "dry run") feature which may need to include this change, but we can add them together rather than piece meal before we understand the full solution.

tomponline:
Closing this for now.

PR #14408, opened 2024-11-06
Remove non- stdlib errors from `lxd/response`

markylaing:
The package is being consumed by microcluster and microcloud. To avoid import conflicts with other versions of DQLite, we need to remove it's import.


tomponline (approved):
Thanks.

@roosterfish please let me know if this is OK to backport to stable-5.21?

lxd/db/errors.go
```diff
@@ -2,9 +2,19 @@ package db
...
+	"github.com/mattn/go-sqlite3"
 )
 
 var (
 	// ErrNoClusterMember is used to indicate no cluster member has been found for a resource.
 	ErrNoClusterMember = fmt.Errorf("No cluster member found")
 )
+
+// SmartErrors are used to return more appropriate errors to the caller.
+var SmartErrors = map[int][]error{
```
roosterfish, line 17: So in case of `microcluster/v2` we create our own `SmartErrors` using `go-dqlite/v2`?

roosterfish: And the call `Init()` using our own to prevent importing the `v3`.

roosterfish (approved):

lxd/db/errors.go
tomponline: yeah thats right

PR #14413, opened 2024-11-06
Allow vm `security.shared`

MggMuggins:
This is a preparatory change for attaching `virtual-machine` ~~and `container`~~ type volumes to instances.

`security.shared` will be required for in order to attach `virtual-machine` volumes to other VMs.

Also a few bits I noticed in the docs.

### DIFF
--- a/doc/explanation/storage.md
+++ b/doc/explanation/storage.md
@@ -122,7 +122,7 @@ Storage volumes can be of the following types:
 : LXD automatically creates one of these storage volumes when you launch an instance.
   It is used as the root disk for the instance, and it is destroyed when the instance is deleted.
 
-  This storage volume is created in the storage pool that is specified in the profile used when launching the instance (or the default profile, if no profile is specified).
+  This storage volume is created in the storage pool that is specified when launching the instance (or in the default profile, if no pool or profile is specified).
   The storage pool can be explicitly specified by providing the `--storage` flag to the launch command.
 
 `image`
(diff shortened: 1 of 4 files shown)


tomponline:
> `security.shared` will be required for in order to attach `virtual-machine` volumes to other instances.

Hrm, im not sure why that is needed, because that `security.shared` setting is designed to allow concurrently mounting custom block volumes on multiple running instances, but for the feature you're working on we will require that the instance's whose root volume you're attaching to a different instance wont be running.

tomponline:
> This is a preparatory change for attaching `virtual-machine` and `container` type volumes to instances.

Also interesting that you mention container volumes here, im not opposed to this at all, but this might introduce additional complexities wrt to idmapping, so might be a "phase 2" feature, as the primary driver for the feature you're working on is to add support for attaching VM root volumes to another VM for recovery/export/import.

MggMuggins:
> attaching VM root volumes to another VM for recovery/export/import

I totally `s/VM/instance/` when I read the roadmap item; my bad. This will be reflected in the spec.

`security.shared` only guards creation of multiple disk devices referring to the same volume (it's only checked [during config validation](https://github.com/canonical/lxd/blob/main/lxd/device/disk.go#L151) or when [updating storage volumes](https://github.com/canonical/lxd/blob/main/lxd/storage/backend_lxd.go#L6019)), basically `--yes-i-really-mean-it` for creating the devices.

As we found while I was working on cluster member limits, making decisions based on the running state of instances is prone to races.

We should require `security.shared: true` to attach a VM's root volume to another instance, since we have no way of guarding against corruption by checking the running state.

We can put this on hold until I've put together a spec.

tomponline:
Lets close this for now.

PR #14415, opened 2024-11-07
github: Use cache when Trivy DB download fails

hamistao:
This updates the Trivy workflow to cache the vulnerability database when the download is successful and, when it is not, use the latest cached database instead. Motivated by some failures seen on some recent runs: https://github.com/canonical/lxd/actions/runs/11603622604/job/32310941454


tomponline (approved):

tomponline:
ta!

PR #14417, opened 2024-11-07
`unix-hotplug` ownership fix

kadinsayani:
Fixes the second bug raised in https://github.com/canonical/lxd/issues/14266.


simondeziel (dismissed):

tomponline:
@kadinsayani please can you update the PR description to explain the change of behaviour in this PR and what, if any, will change for existing users?

kadinsayani:
I noticed some wording inconsistencies in the documentation for `unix-hotplug` and the other unix device types. I've opened https://github.com/canonical/lxd/issues/14451 to track the issue and engage @minaelee for feedback. I'll open up a separate PR to update the docs and resolve the issue.

lxd/device/device_utils_unix.go
```diff
@@ -176,13 +186,19 @@ func UnixDeviceCreate(s *state.State, idmapSet *idmap.IdmapSet, devicesPath stri
 		if err != nil {
 			return nil, fmt.Errorf("Invalid uid %s in device %s", m["uid"], srcPath)
 		}
+	} else {
+		_, uid, _ := unixDeviceOwnership(srcPath)
+		d.UID = int(uid)
```
simondeziel, line 191: Here we are coercing a `uint32` into a `int`, right?

kadinsayani: Yes 👍

simondeziel: I know we do this elsewhere but it doesn't feel like the right thing to do :/

kadinsayani: I don't think there is any potential for overflow here, but we could consider modifying the structure:

```Go
// UnixDevice contains information about a created UNIX device.
type UnixDevice struct {
	HostPath     string      // Absolute path to the device on the host.
	RelativePath string      // Relative path where the device will be mounted inside instance.
	Type         string      // Type of device; c (for char) or b for (block).
	Major        uint32      // Major number.
	Minor        uint32      // Minor number.
	Mode         os.FileMode // File mode.
	UID          int         // Owner UID.
	GID          int         // Owner GID.
}
```

I don't see any reason why the `UID` and `GID` fields can't be of type `uint32`. What do you think?

kadinsayani: Or, we can add some overflow validation to `unixDeviceOwnership` and return the gid and uid with type `int`.

kadinsayani: But since `unixDeviceOwnership` returns the gid and uid with type `uint32`, we aren't at risk for any overflow by using the `int()` cast function.

simondeziel: As pointed out by Tom on another PR (https://github.com/canonical/lxd/pull/14327#discussion_r1829047630), `int()` on 32-bit platforms is only 32 bit long. This means it can overflow there, isn't it?

kadinsayani: In this case, it can't overflow because the gid and uid returned by `unixDeviceOwnership` is guaranteed to be of type `uint32`. I'd prefer either keeping the `int()` cast as is or adding an `int()` cast to the `unixDeviceOwnership` return values rather than modifying the `UnixDevice` structure.

simondeziel: So we discussed over IM and got to the conclusion that despite `uint32()` risking to overflow once cast to `int()`, it was harmless as `int()` always return `-1` when wrapping around and @kadinsayani found it:

```
// Chown changes the numeric uid and gid of the named file.
// If the file is a symbolic link, it changes the uid and gid of the link's target.
// A uid or gid of -1 means to not change that value.
// If there is an error, it will be of type [*PathError].
//
// On Windows or Plan 9, Chown always returns the [syscall.EWINDOWS] or
// EPLAN9 error, wrapped in *PathError.
func Chown(name string, uid, gid int) error {
    e := ignoringEINTR(func() error {
        return syscall.Chown(name, uid, gid)
    })
    if e != nil {
        return &PathError{Op: "chown", Path: name, Err: e}
    }
    return nil
}
```

Meaning that in case of overflow, the UID/GID will remain unchanged anyway.

simondeziel (dismissed):

lxd/device/device_utils_unix.go
kadinsayani: Thanks for summarizing our findings @simondeziel :)

kadinsayani:
@tomponline, do we need an API extension for the new setting even if we're not adding tests?

tomponline:
> @tomponline, do we need an API extension for the new setting even if we're not adding tests?

yes new settings require API extensions in order for clients to be able to ascertain whether the server supports the setting.

kadinsayani:
> > @tomponline, do we need an API extension for the new setting even if we're not adding tests?
> 
> yes new settings require API extensions in order for clients to be able to ascertain whether the server supports the setting.

Gotcha, makes sense.

kadinsayani:
I've added the API extension :)

doc/api-extensions.md
```diff
@@ -2520,3 +2520,7 @@ contains the total available logical CPUs available when LXD started.
 ## `vm_limits_cpu_pin_strategy`
 
 Adds a new {config:option}`instance-resource-limits:limits.cpu.pin_strategy` configuration option for virtual machines. This option controls the CPU pinning strategy. When set to `none`, CPU auto pinning is disabled. When set to `auto`, CPU auto pinning is enabled.
+
+## `unix_hotplug_ownership_inherit`
```
tomponline, line 2524: should this be `unix_hotplug_ownership_inherit` to be consistent with existing `unix_device_hotplug` api extension?

lxd/device/device_utils_unix.go
```diff
@@ -170,18 +184,28 @@ func UnixDeviceCreate(s *state.State, idmapSet *idmap.IdmapSet, devicesPath stri
 		d.Type = "c"
 	}
 
-	// Get the device owner.
-	if m["uid"] != "" {
-		d.UID, err = strconv.Atoi(m["uid"])
+	if shared.IsTrue(m["ownership.inherit"]) {
```
tomponline, line 187: Shall we tweak this so that ` unixDeviceOwnership` is run when either m["uid"] or m["gid"] are unset, and then only set the key(s) that aren't explicitly overridden?

lxd/device/device_utils_unix.go
```diff
@@ -170,18 +184,28 @@ func UnixDeviceCreate(s *state.State, idmapSet *idmap.IdmapSet, devicesPath stri
...
-		if err != nil {
-			return nil, fmt.Errorf("Invalid gid %s in device %s", m["gid"], srcPath)
+		d.GID = int(gid)
+		d.UID = int(uid)
+	} else {
+		// Get the device owner.
+		if m["uid"] != "" {
+			d.UID, err = strconv.Atoi(m["uid"])
+			if err != nil {
+				return nil, fmt.Errorf("Invalid uid %s in device %s", m["uid"], srcPath)
```
tomponline, line 200: whilst we changing this, please can you use %q for quoting values in errors rather than %q

doc/api-extensions.md
kadinsayani: If you mean `unix_device_hotplug_ownership_inherit`, yes I think that's more consistent 👍

lxd/device/unix_hotplug.go
```diff
@@ -91,6 +99,10 @@ func (d *unixHotplug) validateConfig(instConf instance.ConfigReader) error {
 		return fmt.Errorf("Unix hotplug devices require a vendorid or a productid")
 	}
 
+	if d.config["gid"] != "" && d.config["uid"] != "" && shared.IsTrue(d.config["ownership.inherit"]) {
+		return fmt.Errorf("Unix hotplug device ownership cannot be inherited from host while GID and UID are set")
+	}
```
simondeziel, line 104: Shouldn't this error be return whenever one of `gid` or `uid` is set while `ownership.inherit == true`?

kadinsayani: We're allowing one of `gid` or `uid` to be set with `ownership.inherit` set to `true.

kadinsayani: See https://github.com/canonical/lxd/pull/14417#discussion_r1846459783.

kadinsayani:
@tomponline rebased 👍

tomponline:
rebase please

kadinsayani:
> rebase please

Rebased.

tomponline (approved):
Thanks!

PR #14419, opened 2024-11-08
Network: add target parameter for GET 1.0/networks and network_get_ta…

### LINKED ISSUE #14406: Add locations to network GET
# Required information

 * Distribution: snap
 * Distribution version: git-9bbe4d7             30973  latest/edge

# Issue description

When fetching networks, we need the information on which cluster member they exist on. The current `GET 1.0/networks?project=:project&recursion=1` response includes a `locations` field. The field is correctly populated for managed networks.

However, for non-managed networks, the `locations` field is always null. This seems wrong, because a physical network can exist only on some members, as per the reproducer below.

# Steps to reproduce

 1. Initialize a 3 node microcluster to get a clustered LXD 
 2. On one of the cluster members, create a physical network. In a shell, run: `sudo ip link add eth10 type dummy`
 3. Fetch LXD network list via API `GET 1.0/networks?project=:project&recursion=1`.
 4. The newly created `eth10` appears in the list with `locations: null`. 

Expected output would be a `locations: [ "member-name" ]` value for the `eth10` network. Because it only exists on exactly one cluster member and not on the others.

edlerd:
# Done

* add target parameter for GET 1.0/networks 
* add network_get_target api extension

fixes #14406 

The UI can fetch local networks for each member, and we don't need to populate the `locations` field for non managed networks in the response of `GET 1.0/netwroks`. Populating that field is expensive, because LXD would have to fan out the request to all members to collect the network interface lists for all of them.


doc/api-extensions.md
```diff
@@ -2520,3 +2520,7 @@ contains the total available logical CPUs available when LXD started.
 ## `vm_limits_cpu_pin_strategy`
 
 Adds a new {config:option}`instance-resource-limits:limits.cpu.pin_strategy` configuration option for virtual machines. This option controls the CPU pinning strategy. When set to `none`, CPU auto pinning is disabled. When set to `auto`, CPU auto pinning is enabled.
+
+## `network_get_target`
+
+Adds optional `target` parameter to `GET /1.0/network`. When target is set, forward the request to the specified cluster member and return the non managed interfaces from that member.
```
minaelee, line 2526:
```suggestion
Adds optional `target` parameter to `GET /1.0/network`. When target is set, forward the request to the specified cluster member and return the non-managed interfaces from that member.
```
```suggestion
Adds optional `target` parameter to `GET /1.0/network`. When target is set, forward the request to the specified cluster member and return the non-managed interfaces from that member.
```

tomponline (changes requested):
@edlerd @mas-who

Some reservations I have with this approach:

1. It requires the UI to make an API request to its LXD server (which is in turn forwarded to the right cluster member) for each cluster member it knows about. This isn't particularly efficient.
2. For each cluster member that handles the forward API request, they will return all the managed networks (which in turn requires loading those same records from the DB repeatedly from the leader). This would be OK if you were trying to access per-cluster member config (like `parent` setting when using `recursion=1`)

But IIRC the reason for this was so the UI can generate a list of candidate parent network interfaces to be used as the parent for a managed network.

The other option is to change the output of `/1.0/network` to list all physical interfaces across the cluster, and add a `Location` field to the struct. We would also need to reflect this in the `lxc network list` output to show a `Location` col otherwise we'd get repeated entries and it would be confusing as to why. This could be considered a CLI experience break too.

So that seems less desirable.

But on the flip side, I am worried that adding a `target` param support to `list` endpoint is an inconsistent API feature. I can't think of an API we have at the moment that supports targeting at list level (if you know of one please remind me).

I thought of this because I was thinking we would need to add some tests to https://github.com/canonical/lxd/blob/main/test/suites/clustering.sh#L1041C1-L1041C24 (by creating dummy interfaces of different names inside each cluster member's network namespace and then testing the targeting works as appropriate with `lxc network list --target=<member>`), and it occurred to me that that invocation looks a bit unusual.

Remind me, did we discuss/discount using the `/1.0/resources` API for this?
Perhaps we need to add targeting and filtering by network interface there instead.

edlerd:
@tomponline 

> [...] I am worried that adding a `target` param support to `list` endpoint is an inconsistent API feature. I can't think of an API we have at the moment that supports targeting at list level (if you know of one please remind me).

There are a few list endpoints, that allow sending a target parameter:
* list network forwards 
* list network load balancers
* list network peers
* list storage volumes
* list resources
* metrics get

In my opinion, this is a catching up of the network list to also support the same to be consistent with the other get endpoints that can return cluster member specific information.

> I thought of this because I was thinking we would need to add some tests to https://github.com/canonical/lxd/blob/main/test/suites/clustering.sh#L1041C1-L1041C24 (by creating dummy interfaces of different names inside each cluster member's network namespace and then testing the targeting works as appropriate with `lxc network list --target=<member>`), and it occurred to me that that invocation looks a bit unusual.

Sure, the changes to the CLI can be discussed and changed accordingly when we get to it. This PR is only about adding the target filter/forwarding to the network list API endpoint.

> Remind me, did we discuss/discount using the `/1.0/resources` API for this? Perhaps we need to add targeting and filtering by network interface there instead.

The resources API already has the target parameter and provides us with the network names under the `network.cards.[:id].ports.[:index].id`. Though this is not including any non-managed bridges. I think we support those as uplink/parent as well, and we'd be missing those in our selectors. So this might not be sufficient for our use case. Though, I am not 100% sure if we need the bridges in the selector or not.:

image

tomponline:
> There are a few list endpoints, that allow sending a target parameter:

Oh well thats good.

Interestingly, I cant see why `list network peers` would need a target option, as there's no peering that is member specific.

tomponline:
> Sure, the changes to the CLI can be discussed and changed accordingly when we get to it. This PR is only about adding the target filter/forwarding to the network list API endpoint.

We'd want CLI support as part of this PR especially as adding `--target` is not going to break the output. That way we can include tests for both in this PR.

tomponline:
> Though this is not including any non-managed bridges

`/1.0/resources` seems to be only including physical interfaces, not bridges (managed or otherwise).

tomponline (changes requested):
OK thanks @edlerd 

Lets go forward with this with the understanding that the way the UI plans to use it will mean that managed networks within the default project (the UI must always sent `project=default` when getting uplink parent candidates) will be retrieved multiple times once for each cluster member which is somewhat inefficient.

Please can you add the CLI and test changes as discussed.

Thanks

edlerd:
- Added target parameter for `lxc network list --target=:member`.
- Added test for the cli, unfortunately we don't have dedicated hosts in the cluster test, so falling back to creating dummy interfaces and checking them per member instead.

test/suites/clustering.sh
```diff
@@ -1217,6 +1217,14 @@ test_clustering_network() {
   LXD_DIR="${LXD_ONE_DIR}" lxc config device remove c3 eth0
   LXD_DIR="${LXD_ONE_DIR}" lxc config device add c3 eth0 nic hwaddr="${c1MAC}" nictype=bridged parent="${net}"
 
+  # Check networks local to a cluster member show up when targeting that cluster member
```
tomponline, line 1220: the per-member dummy interfaces wont be being created inside the network namespace for each cluster member.

tomponline: See https://github.com/canonical/lxd/blob/a82bf163bf28f8c28e2e7e95e6201d0ce9bca0dd/test/includes/clustering.sh#L26

tomponline: Here's an example of running network setup inside a network namespace

https://github.com/canonical/lxd/blob/a82bf163bf28f8c28e2e7e95e6201d0ce9bca0dd/test/includes/clustering.sh#L77-L85

tomponline: and the pid file for the ns seems to be accessible here

https://github.com/canonical/lxd/blob/a82bf163bf28f8c28e2e7e95e6201d0ce9bca0dd/test/includes/clustering.sh#L70C9-L70C42

edlerd: Thanks for the pointers, with your help, I think the tests are fine now :)

test/suites/clustering.sh
```diff
@@ -1217,6 +1217,13 @@
   LXD_DIR="${LXD_ONE_DIR}" lxc config device remove c3 eth0
   LXD_DIR="${LXD_ONE_DIR}" lxc config device add c3 eth0 nic hwaddr="${c1MAC}" nictype=bridged parent="${net}"
 
+  # Check networks local to a cluster member show up when targeting that member
+  # and hidden when targeting other cluster members. Setup is in includes/clustering.sh
+  LXD_DIR="${LXD_ONE_DIR}" lxc network list --target=node1 | grep localBridge1
+  ! LXD_DIR="${LXD_ONE_DIR}" lxc network list --target=node1 | grep localBridge2
```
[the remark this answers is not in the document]
edlerd: this was resolved

test/suites/clustering.sh
```diff
@@ -1217,6 +1217,13 @@
...
+  ! LXD_DIR="${LXD_ONE_DIR}" lxc network list --target=node2 | grep localBridge1
```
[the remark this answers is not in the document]
edlerd: this was resolved

test/includes/clustering.sh
```diff
@@ -82,6 +82,7 @@ ip link set dev "${veth2}" name eth0
 ip link set eth0 up
 ip addr add "10.1.1.10${id}/16" dev eth0
 ip route add default via 10.1.1.1
+ip link add localBridge${id} type bridge
```
tomponline, line 85: can we access the net ns via id to only add the per-member dummy nics in the test_clustering_network

Also this is removing the interface not needed because its destroyed when the netns is destroyed?

edlerd: Yes, destroying should not be needed as netns gets destroyed after the test run.

I tried to move the dummy bridge setup into the test code, but it turned out quite verbose. Because we need to talk to nsenter, and getting the right path to read the `nspid` is not straight forward:
```
id="${1}"
shift
prefix="lxd$$"
ns="${prefix}${id}"
```
Especially the 2nd and 3rd line are a bit of a mystery to me. Duplicating those lines in the test seemed like a bad idea, as it clashes with the single source of truth. That's why I resorted to adding a local bridge on each node with the id of that node in the setup code directly, where the ns pid is already resolved and in use.

tomponline (approved):

PR #14420, opened 2024-11-08
Set Target after network/storage lookup (from Incus)

markylaing:
In LXD init, if the `--storage` or `--network` flags are provided the CLI will check the storage pool or network exists before setting up a device on the instance config. When targeting a cluster group this would lead to calls like `GET /1.0/storage-pools/foo?target=@bar` which doesn't make sense. 

This commit (cherry-picked from Incus) moves setting the cluster group target to after those lookups.

I've added tests have checked that they fail before this change and pass afterwards.


tomponline (approved):
Ta

PR #14422, opened 2024-11-08
Network: Properly handle OVN uplink with disabled DHCP/DHCPv6 (from Incus)

kadinsayani:
(cherry picked from https://github.com/lxc/incus/pull/343)

### DIFF
--- a/lxd/network/driver_ovn.go
+++ b/lxd/network/driver_ovn.go
@@ -1245,7 +1245,16 @@ func (n *ovn) allocateUplinkPortIPs(uplinkNet Network, routerMAC net.HardwareAdd
 					return fmt.Errorf(`Missing required "ipv4.ovn.ranges" config key on uplink network`)
 				}
 
-				ipRanges, err := shared.ParseIPRanges(uplinkNetConf["ipv4.ovn.ranges"], uplinkNet.DHCPv4Subnet())
+				dhcpSubnet := uplinkNet.DHCPv4Subnet()
+				allowedNets := []*net.IPNet{}
+
+				if dhcpSubnet != nil {
+					allowedNets = append(allowedNets, dhcpSubnet)
+				} else {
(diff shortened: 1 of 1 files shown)


lxd/network/driver_ovn.go
```diff
@@ -1245,7 +1245,16 @@ func (n *ovn) allocateUplinkPortIPs(uplinkNet Network, routerMAC net.HardwareAdd
 					return fmt.Errorf(`Missing required "ipv4.ovn.ranges" config key on uplink network`)
 				}
 
-				ipRanges, err := shared.ParseIPRanges(uplinkNetConf["ipv4.ovn.ranges"], uplinkNet.DHCPv4Subnet())
+				dhcpSubnet := uplinkNet.DHCPv4Subnet()
+				allowedNets := []*net.IPNet{}
```
MggMuggins, line 1249: Optional: Would it be possible to dispense with the array and just use a *net.IPNet instead? Could default to uplinkIPv4Net and change the pointer if dhcpSubnet isn't nil.

MggMuggins (commented):
Just one nit.

I assume this makes it possible to use a managed bridge with DHCP disabled as an UPLINK, are there any other use-cases this enables?

kadinsayani:
> 
> 
> 
> 
> Just one nit.
> 
> 
> 
> I assume this makes it possible to use a managed bridge with DHCP disabled as an UPLINK, are there any other use-cases this enables?

You got it, we want to properly handle 'lxc network create <name> ipv4.dhcp=false ipv6.dhcp=false'. There aren't any other use cases that I'm aware of.

lxd/network/driver_ovn.go
kadinsayani: I think we should keep it as a slice in case IPv6 and IPv4 OVN ranges are specified.

tomponline (approved):

PR #13878, opened 2024-08-05
Image export over devlxd for virtual machines

markylaing:
Adds an endpoint to the LXD agent `/dev/lxd` socket listener that proxies requests to `/1.0/images/{fingerprint}/export` on the host LXD devlxd listener.

Additionally cleans up `lxd-agent/devlxd.go` and moves the `security.devlxd.images` config key from the container specific config key map into the shared container/VM config key map and regenerates the metadata.

I just worked on this to speed up my local test runs :)

Closes #12589

PR #14445, opened 2024-11-11
Make `security.devlxd.images` live-updatable for VMs

### LINKED ISSUE #14206: `security.devlxd.images` is not live-updatable for VMs
`security.devlxd.images` is live-updatable for containers but not for VMs:

```
$ lxc launch ubuntu-minimal-daily:24.04 c1
$ lxc config set c1 security.devlxd.images true

$ lxc launch ubuntu-minimal-daily:24.04 v1 --vm
$ lxc config set v1 security.devlxd.images true
Error: Key "security.devlxd.images" cannot be updated when VM is running
```

I haven't look at the implementation but I suspect this config key doesn't change anything on the guest side as the permission to access the needed API endpoint is evaluated at on the LXD host side.

markylaing:
#13878 Added `security.devlxd.images` for virtual machines but this key was not live-updatable. There was no particular reason for this. The endpoint is always available in the LXD agent and access is restricted by the host daemon.

This PR allows the config key to be set while the instance is running.

Closes #14206

### DIFF
--- a/doc/metadata.txt
+++ b/doc/metadata.txt
@@ -2144,7 +2144,7 @@ See {ref}`dev-lxd` for more information.
 
 ```{config:option} security.devlxd.images instance-security
 :defaultdesc: "`false`"
-:liveupdate: "no"
+:liveupdate: "yes"
 :shortdesc: "Controls the availability of the `/1.0/images` API over `devlxd`"
 :type: "bool"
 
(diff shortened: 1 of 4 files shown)


simondeziel:
> There was particular reason for this. The endpoint is always available in the LXD agent and access is restricted by the host daemon.

"There was particular" or "There was no particular" ?
Looking at the diff, it'd think it's the later and you simply omitted the `no` but wanted to be 100% sure :)

markylaing:
> > There was particular reason for this. The endpoint is always available in the LXD agent and access is restricted by the host daemon.
> 
> "There was particular" or "There was no particular" ? Looking at the diff, it'd think it's the later and you simply omitted the `no` but wanted to be 100% sure :)

:laughing: yes I meant that there was *no* particular reason for this.

tomponline (approved):

Issue #14085, opened 2024-09-12
Unable to see non-LXD managed networks for OIDC identity with server admin entitlement


mas-who:
# Issue description

When making an API request to `GET /1.0/networks` as a TLS authenticated identity, all networks are reported by LXD in the response as shown in the example below:
```
{
  "type": "sync",
  "status": "Success",
  "status_code": 200,
  "operation": "",
  "error_code": 0,
  "error": "",
  "metadata": [
    {
      "name": "lxdbr0",
      "description": "",
      "type": "bridge",
      "managed": true,
      "status": "Created",
      "config": {
        "ipv4.address": "10.173.68.1/24",
        "ipv4.nat": "true",
        "ipv6.address": "fd42:fd46:adbb:ef2f::1/64",
        "ipv6.nat": "true"
      },
      "used_by": [
        "/1.0/profiles/default",
        "/1.0/instances/node-1",
        "/1.0/instances/node-2",
        "/1.0/instances/c1",
        "/1.0/instances/c2",
        "/1.0/instances/c3"
      ],
      "locations": [
        "none"
      ]
    },
    {
      "name": "lo",
      "description": "",
      "type": "loopback",
      "managed": false,
      "status": "",
      "config": {},
      "used_by": [],
      "locations": null
    },
    {
      "name": "eno1",
      "description": "",
      "type": "physical",
      "managed": false,
      "status": "",
      "config": {},
      "used_by": [],
      "locations": null
    },
    {
      "name": "wlo1",
      "description": "",
      "type": "physical",
      "managed": false,
      "status": "",
      "config": {},
      "used_by": [],
      "locations": null
    },
    {
      "name": "ovs-system",
      "description": "",
      "type": "unknown",
      "managed": false,
      "status": "",
      "config": {},
      "used_by": [],
      "locations": null
    },
    {
      "name": "genev_sys_6081",
      "description": "",
      "type": "unknown",
      "managed": false,
      "status": "",
      "config": {},
      "used_by": [],
      "locations": null
    },
    {
      "name": "br-int",
      "description": "",
      "type": "bridge",
      "managed": false,
      "status": "",
      "config": {},
      "used_by": [],
      "locations": null
    },
    {
      "name": "docker0",
      "description": "",
      "type": "bridge",
      "managed": false,
      "status": "",
      "config": {},
      "used_by": [],
      "locations": null
    },
    {
      "name": "tapb04dfee5",
      "description": "",
      "type": "unknown",
      "managed": false,
[...]

PR #14446, opened 2024-11-11
Auth: Mask errors returned by OpenFGADatastore

markylaing:
While working on https://github.com/canonical/lxd/issues/14085 I set up a new fine-grained TLS identity and issued the following commands as that identity, without any permissions yet (I forgot I'd changed my default remote):
```
$ lxc auth group create tmp
Error: Forbidden
$ lxc auth group permission add tmp server admin
Error: Failed to check OpenFGA relation: No such entity "/1.0/auth/groups/tmp"
```
Creating the group failed, this is correct behaviour. 

When attempting to add a permission to the non-existent group, the request failed (correct) but the OpenFGA Authorization driver returned the above error. This is incorrect.

This PR checks if the error returned by a `Check` request on the embedded OpenFGA server is a `Not Found` error and returns a generic not found error. This makes errors returned by the authorizer consistent. We are masking all not found errors returned before access control decisions are made to prevent discovery. After this change, the same command returns:
```
$ lxc auth group permission add tmp server admin
Error: Not Found
```

### DIFF
--- a/lxd/auth/drivers/openfga.go
+++ b/lxd/auth/drivers/openfga.go
@@ -250,6 +250,12 @@ func (e *embeddedOpenFGA) CheckPermission(ctx context.Context, entityURL *api.UR
 	l.Debug("Checking OpenFGA relation")
 	resp, err := e.server.Check(ctx, req)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


hamistao (approved):
LGTM!

PR #14447, opened 2024-11-11
Auth: Allow listing unmanaged networks with fine-grained auth

markylaing:
Fixes issue where fine-grained authorization was filtering out unmanaged networks (because they aren't in the database). To do this I added a new entitlement at the server level: `can_view_unmanaged_networks`. Restricted TLS clients are granted this by default to not break existing behaviour.

Opening this now but will mark as draft until I add some tests.

Closes https://github.com/canonical/lxd/issues/14085


lxd/networks.go
```diff
@@ -241,20 +241,32 @@ func networksGet(d *Daemon, r *http.Request) response.Response {
...
 
-	var networkNames []string
+	// networks holds the network names of the managed and unmanaged networks. They are in two different slices so that
+	// we can perform access control checks differently.
+	var networks [2][]string
 
 	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
 		// Get list of managed networks (that may or may not have network interfaces on the host).
-		networkNames, err = tx.GetNetworks(ctx, effectiveProjectName)
+		networks[0], err = tx.GetNetworks(ctx, effectiveProjectName)
```
tomponline, line 250: please can we have local constants for the magic numbers 0 and 1 so we dont end up with hard to read logic like this:

```go
		if !shared.ValueInSlice(iface.Name, networks[0]) {
				networks[1] = append(networks[1], iface.Name)
			}
```

markylaing: Yep sure. I wasn't sure about this either and contemplated using a map but it seemed overkill for something so simple. I'll add some constants.

tomponline (approved):

tomponline:
Thanks!

PR #14449, opened 2024-11-12
Tests: Remove `gen_cert` util

markylaing:
The `gen_cert` util had a weird side-effect if I was concurrently running a LXD snap with OIDC configured. (I realise that there can be some test failures when doing this related to networks etc. but I do it regularly when running a single suite as I'm often using the snapped LXD for something else). I've put more details in the commit messages.

Additionally simplifies the `user_is_server_admin` function in the fine-grained auth tests.

### DIFF
--- a/test/includes/lxc.sh
+++ b/test/includes/lxc.sh
@@ -36,19 +36,3 @@ lxc_remote() {
         eval "timeout --foreground 120 ${cmd}"
     fi
 }
-
-gen_cert() {
-    # Temporarily move the existing cert to trick LXC into generating a
-    # second cert.  LXC will only generate a cert when adding a remote
-    # server with a HTTPS scheme.  The remote server URL just needs to
-    # be syntactically correct to get past initial checks; in fact, we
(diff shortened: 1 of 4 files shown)


tomponline:
> The `gen_cert` util had a weird side-effect 

Please don't keep us in suspense, what was it? :)

tomponline:
> I've put more details in the commit messages.

ah ok will read those :)

tomponline (dismissed):

markylaing:
Hmmm test failures. Weird because the auth tests passed locally. Will look into it.

markylaing:
> Hmmm test failures. Weird because the auth tests passed locally. Will look into it.

Should be sorted now. The value of `LXD_CONF2` was set in another function and that's were the `tls` remote was defined. When I ran the test locally I had named it `TMP_LXD_CONF` and I didn't think the rename would matter. Sorry for the wasted CI time :cry:

tomponline (approved):

PR #14453, opened 2024-11-12
Improve USB ID type handling

simondeziel:


shared/usbid/parse.go
```diff
@@ -242,7 +242,7 @@ parseLines:
 		case "":
 			err = parseVendor(level, id, name)
 		case "C":
-			err = parseClass(level, id, name)
+			err = parseClass(level, uint8(id), name)
```
kadinsayani, line 245: I'm not sure about casting from `string` to `uint8` without a bound check.

simondeziel: I think `id` comes from `split()` that has the 3rd return arg as `uint16` which is then coerced into `uint8`. This is then used to lookup in a struct with a fallback to `strconv.Itoa()`.

I must admit to not fully understand that code :/

simondeziel:
This is trying to address https://github.com/canonical/lxd/security/code-scanning?query=is%3Aopen+branch%3Amain+rule%3Ago%2Fincorrect-integer-conversion+path%3Ashared%2Fusbid

tomponline (approved):

PR #14458, opened 2024-11-13
lxc: Use TrustToken field if supported by the server (stable-5.21)

roosterfish:
When adding a remote by putting the token in the positional argument (not `--token`) also check if the server supports the `explicit_trust_token` extension and send the password using the `TrustToken` field instead.

This seems to be fixed already in [`main`](https://github.com/canonical/lxd/blob/main/lxc/remote.go#L271).

### DIFF
--- a/lxc/remote.go
+++ b/lxc/remote.go
@@ -255,13 +255,17 @@ func (c *cmdRemoteAdd) addRemoteFromToken(addr string, server string, token stri
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


roosterfish:
Tested it on a `5.21/edge` against `latest/edge` LXD (as remote).

tomponline (approved):
Thanks!

PR #14459, opened 2024-11-13
github: Use squid/candidate for microceph (stable-5.21)

tomponline:
To align with expectations of microcloud 2.x LTS to be used with microceph squid.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -313,6 +313,8 @@ jobs:
       - name: Setup MicroCeph
         if: ${{ matrix.backend == 'ceph' }}
         uses: ./.github/actions/setup-microceph
+        with:
+          microceph-channel: "squid/candidate"
 
       - name: Make GOCOVERDIR
         run: |


.github/workflows/tests.yml
```diff
@@ -313,6 +313,8 @@ jobs:
       - name: Setup MicroCeph
         if: ${{ matrix.backend == 'ceph' }}
         uses: ./.github/actions/setup-microceph
+        with:
+          microceph-channel: "squid/candidate"
```
roosterfish, line 317:
```suggestion
          microceph-channel: "squid/stable"
```

simondeziel (approved):

.github/workflows/tests.yml
roosterfish: Why not using stable?

roosterfish (changes requested):

.github/workflows/tests.yml
roosterfish: I see, thx.

roosterfish (approved):

PR #14461, opened 2024-11-13
VM: Add missing `disk` block node name overflow handling

kadinsayani:
Two commits were identified as potential imports in https://github.com/canonical/lxd/issues/13311:
The first commit adds a function `blockNodeName` which hashes block node names if they exceed 31 characters. The second commit applies the node name overflow function when adding a QEMU config to a drive.

@simondeziel pointed out that we already have a function `qemuDeviceNameOrID` so this PR just adds missing node name overflow handling.

### DIFF
--- a/lxd/instance/drivers/driver_qemu.go
+++ b/lxd/instance/drivers/driver_qemu.go
@@ -4113,7 +4113,7 @@ func (d *qemu) addDriveConfig(qemuDev map[string]string, bootIndexes map[string]
 		revert := revert.New()
 		defer revert.Fail()
 
-		nodeName := fmt.Sprintf("%s%s", qemuDeviceNamePrefix, escapedDeviceName)
+		nodeName := qemuDeviceNameOrID(qemuDeviceNamePrefix, escapedDeviceName, "", qemuDeviceNameMaxLength)
 
 		if isRBDImage {
 			secretID := fmt.Sprintf("pool_%s_%s", blockDev["pool"], blockDev["user"])


lxd/instance/drivers/driver_qemu.go
```diff
@@ -8962,6 +8964,23 @@ func (d *qemu) deviceDetachUSB(usbDev deviceConfig.USBDeviceItem) error {
 	return nil
 }
 
+// Block node names may only be up to 31 characters long, so use a hash if longer.
+func (d *qemu) blockNodeName(name string) string {
```
simondeziel, line 8968: https://github.com/canonical/lxd/blob/main/lxd/instance/drivers/driver_qemu_templates.go#L16-L35 seems to be doing something similar, would it be possible to reuse it?

kadinsayani: Thanks for pointing this out! We certainly can!

tomponline:
Yeah do we need this @hamistao ?

lxd/instance/drivers/driver_qemu.go
```diff
@@ -4113,7 +4113,7 @@ func (d *qemu) addDriveConfig(qemuDev map[string]string, bootIndexes map[string]
 		revert := revert.New()
 		defer revert.Fail()
 
-		nodeName := fmt.Sprintf("%s%s", qemuDeviceNamePrefix, escapedDeviceName)
+		nodeName := qemuDeviceNameOrID(qemuDeviceNamePrefix, escapedDeviceName, "", qemuDeviceNameMaxLength)
```
hamistao, line 4116: Do you think we could just use the value from `blockDev["node-name"]`, defined [here](https://github.com/canonical/lxd/pull/14461/commits/668d666c5777aae921912dd5dc88bee6895f30e9#diff-87040b4a9b898c058e0f889f869c7a0448135f0a7f02db851824f48290c3a2f5R3954), instead of creating a new variable?

tomponline: Are they conceptually the same, because in the disk removal you linked above its called blockDevName rather than nodeName, making me think that the node name and the FD set name can be different conceptually.

But apart from that I can see the benefit of referencing the node name directly as you suggest.

hamistao: Indeed they are conceptually different, the FD name only has to be unique per device and predictable so we can derive it on device removal. And using the node name accomplishes this, as this PR already does but we are computing it again effectively.

kadinsayani: In this case, I think it makes sense to add `nodeName` for conceptual clarity and readability.

tomponline (approved):

Issue #13453, opened 2024-05-06
lxc start fails despite stopped state


holmanb:
# Required information

 * Distribution: Ubuntu
 * Distribution version: all
 * The output of "snap list --all lxd core20 core22 core24 snapd":
```
Name    Version         Rev    Tracking       Publisher   Notes
core22  20240111        1122   latest/stable  canonical✓  base,disabled
core22  20240408        1380   latest/stable  canonical✓  base
lxd     5.21.1-d46c406  28460  5.21/stable    canonical✓  -
snapd   2.61.2          21184  latest/stable  canonical✓  snapd,disabled
snapd   2.62            21465  latest/stable  canonical✓  snapd
```
 * The output of "lxc info" or if that fails:
```
config: {}
api_extensions:
- storage_zfs_remove_snapshots
- container_host_shutdown_timeout
- container_stop_priority
- container_syscall_filtering
- auth_pki
- container_last_used_at
- etag
- patch
- usb_devices
- https_allowed_credentials
- image_compression_algorithm
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
- unix_device_rename
- storage_lvm_use_thinpool
- storage_rsync_bwlimit
- network_vxlan_interface
- storage_btrfs_mount_options
- entity_description
- image_force_refresh
- storage_lvm_lv_resizing
- id_map_base
- file_symlinks
- container_push_target
- network_vlan_physical
- storage_images_delete
- container_edit_metadata
- container_snapshot_stateful_migration
- storage_driver_ceph
- storage_ceph_user_name
- resource_limits
- storage_volatile_initial_source
- storage_ceph_force_osd_reuse
- storage_block_filesystem_btrfs
- resources
- kernel_limits
- storage_api_volume_rename
- network_sriov
- console
- restrict_devlxd
- migration_pre_copy
- infiniband
- maas_network
- devlxd_events
- proxy
- network_dhcp_gateway
- file_get_symlink
[...]

PR #14463, opened 2024-11-14
Workaround container self-stop race

MggMuggins:
If an instance self-stops while `statusCode()` is waiting for `getLxcState()` to finish, `statusCode()` may return a stale instance state.

This PR is a workaround for the use-case in #13453 and significantly reduces the likelihood that `statusCode` returns a stale status.

In an ideal world, LXD would maintain a canonical cluster-wide view of instance state. This would allow making race-free decisions based on whether an instance is running or not. For example:
- Project CPU/RAM limits could be enforced at instance start instead of at instance creation
- Volumes with content-type block could be attached to more than one instance without `security.shared`; instance start could fail if another instance with any shared block volumes is already running.

### DIFF
--- a/lxd/instance/drivers/driver_lxc.go
+++ b/lxd/instance/drivers/driver_lxc.go
@@ -8142,19 +8142,9 @@ func (d *lxc) NextIdmap() (*idmap.IdmapSet, error) {
 // statusCode returns instance status code.
 func (d *lxc) statusCode() api.StatusCode {
 	// Shortcut to avoid spamming liblxc during ongoing operations.
-	op := operationlock.Get(d.Project().Name, d.Name())
-	if op != nil {
-		if op.Action() == operationlock.ActionStart {
-			return api.Stopped
-		}
-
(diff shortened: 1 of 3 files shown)


lxd/instance/drivers/driver_lxc.go
```diff
@@ -8142,26 +8142,25 @@ func (d *lxc) NextIdmap() (*idmap.IdmapSet, error) {
...
 	state, err := d.getLxcState()
 	if err != nil {
 		return api.Error
 	}
 
+	// The state that we get from LXC could be stale; if the container is self-stopping,
+	// the on-stop hook handler may be called while we are waiting for getLxcState. If
+	// that happens, we might return `STOPPED` even though a Stop operation is still
+	// running.
+	operationStatus = d.operationStatusCode()
```
tomponline, line 8159: shall we only call this if the getLxcState returns STOPPED? So that we dont incur the extra overhead for running containers?

tomponline (approved):

PR #14467, opened 2024-11-15
tests: print Apport crash reports in case of failure

mihalicyn:
trying to investigate https://github.com/canonical/lxd/actions/runs/11852057329/job/33029976380?pr=14324#step:12:44044

### DIFF
--- a/test/main.sh
+++ b/test/main.sh
@@ -110,6 +110,13 @@ cleanup() {
     if [ "${expandDmesg}" = "no" ]; then
       echo "::endgroup::"
     fi
+
+    if [ "$(ls -A /var/crash)" ]; then
+      echo "::group::crashes"
+      ls -la /var/crash
+      find /var/crash -type f -exec cat {} +
+      echo "::endgroup::"
+    fi
   fi
 
   if [ -n "${GITHUB_ACTIONS:-}" ]; then


test/main.sh
```diff
@@ -110,6 +110,11 @@ cleanup() {
     if [ "${expandDmesg}" = "no" ]; then
       echo "::endgroup::"
     fi
+
+    echo "::group::crashes"
```
tomponline, line 114: should we make this group conditional based on whether there is anything in /var/crash/?

test/main.sh
```diff
@@ -110,6 +110,11 @@ cleanup() {
...
+    ls -la /var/crash
+    cat /var/crash/*
```
simondeziel, line 116: TIL: I thought that apport crash reports were binary files :)

Looking at https://askubuntu.com/questions/1210651/how-to-read-or-open-crash-file-from-var-crash, I see that you can get away with just looking at the textual representation or use a tool to unpack them. If we need more than `cat`, I think we should consider attaching them as CI artifacts.

simondeziel (approved):

test/main.sh
mihalicyn: Have done!

simondeziel (approved):

tomponline (approved):

PR #14470, opened 2024-11-15
Rework how binaries are stripped during CI

simondeziel:


.github/workflows/tests.yml
```diff
@@ -132,6 +132,11 @@ jobs:
           cd /home/runner/work/lxd/lxd-test
           make
 
+      - name: Strip binaries/libs
```
tomponline: why is this change needed?

.github/workflows/tests.yml
```diff
@@ -138,7 +138,8 @@ jobs:
           set -eux
 
           strip --strip-all -v /home/runner/go/bin/dqlite/libs/*.so*
-          strip --strip-all /home/runner/go/bin/{lxc*,lxd*} -v
+          strip --strip-all -v /home/runner/go/bin/lxc*
```
tomponline: why is this change needed?

.github/workflows/tests.yml
```diff
@@ -205,6 +196,15 @@ jobs:
           path: ${{env.GOCOVERDIR}}
         if: env.GOCOVERDIR != ''
 
+      - name: Strip binaries/libs
```
tomponline, line 199: why is this ordering change needed?

.github/workflows/tests.yml
simondeziel: This was me trying to figure out why only some of the binaries were stripped. If you look at https://github.com/canonical/lxd/actions/runs/11862081364/artifacts/2194695458, you'll see:

```
$ file lx* dqlite/libs/libdqlite.so* | grep -vF ', stripped'
lxc-to-lxd:                     ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=6cb0742305291d444e8d1ca580089e25dca77160, for GNU/Linux 3.2.0, with debug_info, not stripped
lxd:                            ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=d75ec2a67943c7dc655388a947ea942cbb0da4c4, for GNU/Linux 3.2.0, with debug_info, not stripped
lxd-benchmark:                  ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=Z0dY2xyk7xY1-vsbMMoV/-HYIQVlqoBAXQwkJjDdV/zIubv6YhiEkB-NZjcv-f/3wjhnzMmqNn5aX6xH-Ax, with debug_info, not stripped
lxd-metadata:                   ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=iSHeRoxOxHPFSKZZxeCD/YP00P_D7Ju4yTubKWWSP/UyXW_mUklLl_qvVvV5mu/GuBCDdilMNUdWbPv1WyS, with debug_info, not stripped
lxd-user:                       ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=W8j-wX--DoFLg5U9xyqv/vL74aihhgDOlNLC7oSje/LpVQlOSkesbA70M2CX1s/QHcbYYpQSVYElN7eiPEN, with debug_info, not stripped
dqlite/libs/libdqlite.so:       ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=f2e9e054edfe6ea3417ca3c3d097c16a4045f49f, with debug_info, not stripped
dqlite/libs/libdqlite.so.0:     ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=f2e9e054edfe6ea3417ca3c3d097c16a4045f49f, with debug_info, not stripped
dqlite/libs/libdqlite.so.0.0.1: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=f2e9e054edfe6ea3417ca3c3d097c16a4045f49f, with debug_info, not stripped
```

.github/workflows/tests.yml
simondeziel: Seems like doing it this late is what made them all be stripped in the uploaded `.zip` that goes from ~145 MiB to 107MiB.

tomponline: weird, are they being built twice or something?

simondeziel: That's what I suspected but couldn't track down.

tomponline: Do we need to not strip them for the tics run?

simondeziel:
@tomponline this now does unconditional stripping of everything. Once merged, I'll check on the next day if TICS is still happy or not.

tomponline (approved):

PR #14476, opened 2024-11-16
auth: Add entitlements to LXD resources if the current identity is fine-grained

gabrielmougard:
JIRA ticket: https://warthogs.atlassian.net/browse/LXD-2208
Specification link: https://docs.google.com/document/d/1GxWV5J57MLrjGEY5RDG7eS86J99A8RMmjKTpn0mEhgY/edit?tab=t.0

## How to test

```bash
# Add TLS certificate to LXD (this can be done through LXD UI. Download the .pfx file)
# Create an auth group
lxc auth group create test-group

# Add an new identity to this group ("lxd-ui" is my TLS certificate identifier created by LXD UI)
lxc auth identity group add tls/lxd-ui test-group-1

# Create resources and add permissions to them
lxc auth group permission add test-group-1 instance c1 can_view project=default

# Create a .pfx certificate (with password "1234") and associate it to an auth group and give it some permissions. You can call the API like the following to perform fine-grained calls:
curl --insecure --cert-type P12 --cert lxd-ui-fine-grained-tls.pfx:1234 "https://127.0.0.1:8443/1.0/instances/c1?recursion=1&with-entitlements=true" | jq '.'
```

The CURL output should give you something like:

```json
{
  "type": "sync",
  "status": "Success",
  "status_code": 200,
  "operation": "",
  "error_code": 0,
  "error": "",
  "metadata": {
    "entitlements": [
      "can_view"
    ],
    "name": "c1",
    "description": "this is a test ",
    "status": "Running",
    "status_code": 103,
    "created_at": "2024-11-12T09:58:54.960586253Z",
    "last_used_at": "2024-11-16T12:15:13.572661467Z",
    "location": "none",
    "type": "container",
    ...
  }
}
```


lxd/daemon.go
```diff
@@ -359,6 +361,220 @@ func allowProjectResourceList(d *Daemon, r *http.Request) response.Response {
...
+
+			err = processEntity(objSlice[1], tuple.Key.Relation)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
```
markylaing, line 576: I'm very confused by this. There are a few things to address here.
1. We can't bypass calls to OpenFGA by accessing the cache directly. We don't know that the cache contains all the necessary information. We will need to call the OpenFGA server itself to check the entitlements on a per-entity basis. The cache should be read from/written to only from in OpenFGA datastore implementation itself.
2. This is leaking a lot of OpenFGA specific logic from the `lxd/db/openfga` and `lxd/auth/drivers` packages. We should try our utmost to keep that logic where it is.
3. Using reflection should be a last resort IMO. If we absolutely require it then we need to have an awful lot of comments to explain why and what it's doing.

For 1, we still need to call the OpenFGA server. I suggest calling `CheckPermission` for each entitlement in the `auth.EntitlementsByEntityType` map value for the entity type. This will be inefficient when listing entitlements for a list of resources (e.g. `GET /1.0/instances?recursion=1&with-entitlements=true`) so I suggest in this calling `auth.GetPermissionChecker` with each entitlement in the `auth.EntitlementsByEntityType` map.

Alternatively, I suggest adding two methods to the `auth.Authorizer` interface to perform the enrichment. One for a single entity which will call `Check` on the OpenFGA server, and one for a list of entities that will call `ListObjects`. The implementation for the TLS authorizer can return a 501 not implemented. The authorizer itself can return a bad request error if the caller is not fine-grained.

For 3, I suggest creating an interface with method `AddEntitlements([]auth.Entitlement)` and implement it for the API types in question (all API types represented by the OpenFGA model).

markylaing (changes requested):

lxd/daemon.go
gabrielmougard: Thanks. I agree with you on the three points. I'll come up with a new approach.

minaelee (changes requested):

doc/rest-api.yaml
```diff
@@ -1,11 +1,34 @@
 definitions:
+    AttachedEntitlements:
+        properties:
+            entitlements:
+                description: Entitlements is a list of entitlements that may be attached to whatever LXD entity.
```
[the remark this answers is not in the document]
edlerd: Shouldn't this say "entitlements that are granted to the requesting user"?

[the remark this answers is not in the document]
gabrielmougard: @minaelee this is the `list comprising only the entitlements that are currently attached to a LXD entity`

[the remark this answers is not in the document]
markylaing: :+1: to @edlerd suggestion. The list should represent the entitlements that the requesting user has on the specific resource. 

So while `can_edit` might be a valid entitlement for the entity type. The point is to attach only the entitlements that the requestor actually has on the resource.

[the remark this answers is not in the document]
gabrielmougard: Thanks. This should be fixed.

doc/api-extensions.md
```diff
@@ -2520,3 +2520,10 @@ contains the total available logical CPUs available when LXD started.
 ## `vm_limits_cpu_pin_strategy`
 
 Adds a new {config:option}`instance-resource-limits:limits.cpu.pin_strategy` configuration option for virtual machines. This option controls the CPU pinning strategy. When set to `none`, CPU auto pinning is disabled. When set to `auto`, CPU auto pinning is enabled.
+
+## `resources_with_entitlements`
+
+Adds `is_fine_grained` field to `GET /1.0/auth/identities/current` to indicate if the current identity
+interacting with the LXD API is fine-grained (i.e, associated permissions are managed via group membership) and
```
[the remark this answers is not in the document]
gabrielmougard: fixed

lxd/daemon.go
gabrielmougard: @markylaing I updated the approach. What do you think now?

lxd/db/openfga/openfga.go
```diff
@@ -484,12 +616,15 @@ WHERE auth_groups_permissions.entitlement = ? AND auth_groups_permissions.entity
...
+	cache.WriteReadStartingFromUserCache(cacheKey, tuples)
+
 	return storage.NewStaticTupleIterator(tuples), nil
 }
 
 // ReadPage is not implemented. It is not required for the functionality we need.
-func (*openfgaStore) ReadPage(ctx context.Context, store string, tk *openfgav1.TupleKey, opts storage.ReadPageOptions) ([]*openfgav1.Tuple, []byte, error) {
-	return nil, nil, api.StatusErrorf(http.StatusNotImplemented, "not implemented")
+func (*openfgaStore) ReadPage(ctx context.Context, store string, tk *openfgav1.TupleKey, opts storage.ReadPageOptions) ([]*openfgav1.Tuple, string, error) {
+	return nil, "", api.StatusErrorf(http.StatusNotImplemented, "not implemented")
```
markylaing, line 627: These changes are related to the update to OpenFGA 1.8.0. Could you include the commit(s) from  https://github.com/canonical/lxd/pull/14440 instead?

lxd/db/openfga/openfga.go
```diff
@@ -28,6 +31,102 @@ func NewOpenFGAStore(clusterDB *db.Cluster) storage.OpenFGADatastore {
...
+	if err != nil {
+		return false, "", nil, err
+	}
+
+	tuples, ok := c.ReadUsersetTuplesCache[string(b)]
+	if ok {
+		return true, string(b), tuples, nil
+	}
+
+	return false, string(b), nil, nil
```
markylaing, line 122: I'm not sure the JSON marshalling/unmarshalling is necessary for the map keys as I think some fields are ignored by the Datastore. For example, wildcard relations were removed from our model because they introduced more complexity than they were worth for our use-case, so we don't need to handle them.

I would check which fields from the input filters to each function are actually used. Then either concatenate them to get a key, or define a struct containing those fields and use that.

If it turns out that marshalling is absolutely necessary, I *think* we can use `google.golang.org/protobuf/encoding/protojson` to marshal/unmarshal directly, without the intermediate struct.

lxd/db/openfga/openfga.go
```diff
@@ -28,6 +31,102 @@ func NewOpenFGAStore(clusterDB *db.Cluster) storage.OpenFGADatastore {
...
+}
+
+// WriteReadUsersetTuplesCache writes tuples to the cache.
+func (c *OpenFGACache) WriteReadUsersetTuplesCache(cacheKey string, tuples []*openfgav1.Tuple) {
+	c.ReadUsersetTuplesCache[cacheKey] = tuples
```
markylaing, line 127: Currently we allow only one call to the datastore at a time (see OpenFGA server options in `lxd/auth/drivers/openfga.go`). Given that we now have caching at the datastore level and that some calls to the datastore don't require database calls at all, I can see us removing that restriction to improve performance. Could you please add a `sync.RWMutex` to these caches so that they are thread safe? Thanks.

lxd/db/openfga/openfga.go
```diff
@@ -231,6 +330,23 @@ func (o *openfgaStore) ReadUsersetTuples(ctx context.Context, store string, filt
...
+		return nil, fmt.Errorf("ReadUsersetTuples: Failed to get OpenFGA request ID from context: %w", err)
+	}
+
+	ok, cacheKey, tuples, err := cache.ReadReadUsersetTuplesCache(filter)
+	if err != nil {
+		return nil, fmt.Errorf("ReadUsersetTuples: Failed to read cache: %w", err)
+	}
+
+	if ok {
+		logger.Error("ReadUsersetTuples: Cache hit", logger.Ctx{"cacheKey": cacheKey, "tuples": tuples})
```
markylaing, line 345: Why the error log here?

lxd/db/openfga/openfga.go
```diff
@@ -231,6 +330,23 @@ func (o *openfgaStore) ReadUsersetTuples(ctx context.Context, store string, filt
 		return nil, fmt.Errorf("ReadUsersetTuples: Failed to parse entity URL %q: %w", entityURL, err)
 	}
 
+	// Check the request cache.
+	cache, err := request.GetCtxValue[*OpenFGACache](ctx, request.CtxOpenFGARequestCache)
+	if err != nil {
+		return nil, fmt.Errorf("ReadUsersetTuples: Failed to get OpenFGA request ID from context: %w", err)
```
markylaing, line 336: I don't think we should error out if the cache is not present. We may not always want to use the cache. For example on long-running connections like `/1.0/events`. Instead I think we should just call the database as normal.

lxd/daemon.go
```diff
@@ -764,6 +764,12 @@ func (d *Daemon) createCmd(restAPI *mux.Router, version string, c APIEndpoint) {
 			return
 		}
 
+		// Generate an OpenFGA request ID
+		_, err = request.GetCtxValue[*openfga.OpenFGACache](r.Context(), request.CtxOpenFGARequestCache)
+		if err != nil {
+			request.SetCtxValue(r, request.CtxOpenFGARequestCache, openfga.NewOpenFGACache())
+		}
+
```
markylaing, line 772: Why get it first and then set it if not present? At this point we can be sure that it has not been set. 

Also could you please move this up to where the username and protocol are being set and use the same pattern. Thanks.

lxd/identities.go
```diff
@@ -1126,6 +1131,7 @@ func getCurrentIdentityInfo(d *Daemon, r *http.Request) response.Response {
 		Identity:             *apiIdentity,
 		EffectiveGroups:      effectiveGroups,
 		EffectivePermissions: effectivePermissions,
+		IsFineGrained:        isFineGrained,
```
markylaing, line 1134: Can we just set `identity.IsFineGrainedIdentityType(string(id.Type))` as the field value? Why does it need to be evaluated in the transaction?

shared/api/auth.go
```diff
@@ -84,6 +84,11 @@ func (i *Identity) SetWritable(put IdentityPut) {
 	i.TLSCertificate = put.TLSCertificate
 }
 
+// AddEntitlement adds an entitlement to the identity.
+func (i *Identity) AddEntitlement(entitlement string) {
+	i.Entitlements = append(i.Entitlements, entitlement)
+}
+
```
markylaing, line 91: I'm not sure but can this function just be defined on the `AddedEntitlements` struct? It's embedded, so the parent struct will then implement the method automatically and you won't need to implement it separately everywhere.

lxd/auth/types.go
```diff
@@ -52,6 +52,12 @@ type Authorizer interface {
 	// the entity. The effective project for the entity must be set in the request context as request.CtxEffectiveProjectName
 	// *before* the call to GetPermissionChecker.
 	GetPermissionChecker(ctx context.Context, entitlement Entitlement, entityType entity.Type) (PermissionChecker, error)
+
+	// AddEntitlements attaches entitlements to an entity.
+	AddEntitlements(ctx context.Context, entityType entity.Type, entityURL *api.URL, entity entity.EntityWithEntitlements) error
+
+	// AddEntitlementsToEntities attaches entitlements to multiple entities.
+	AddEntitlementsToEntities(ctx context.Context, entityType entity.Type, entityURLs []*api.URL, entities []entity.EntityWithEntitlements) error
```
markylaing, line 60: I'm not sure on these function signatures. Especially in the second case where the caller must prepare a list of URLs to correspond by index to the list of entities. 

It seems odd to me that we're passing in e.g. an `api.Instance`, along with it's URL and it's type, both of which can be derived from the fact that it is an `api.Instance`.

Do you think it's worth adding two more methods to `EntityWithEntitlements` e.g.:
```
type EntityWithEntitlements interface {
    AddEntitlement(auth.Entitlement)
    URL() string
    EntityType() entity.Type
}
```

Then these function signatures can become:
```
AddEntityEntitlements(ctx context.Context, entity entity.EntityWithEntitlements)
AddEntitiesEntitlements(ctx context.Context, entities entity.EntityWithEntitlements)
```

Let me know what you think.

lxd/instance_get.go
```diff
@@ -124,6 +126,9 @@ func instanceGet(d *Daemon, r *http.Request) response.Response {
 		recursion = 0
 	}
 
+	// Detect if we want to also returns entitlements for each instance.
+	withEntitlements := request.QueryParam(r, "with-entitlements") == "true"
```
markylaing, line 130: We might want to use `shared.IsTrue` here to also capture `yes`, and `1`, or anything capitalised.

lxd/instance_get.go
```diff
@@ -152,5 +157,19 @@ func instanceGet(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	if withEntitlements {
+		if recursion == 0 {
```
markylaing, line 161: Isn't recursion `0` just a list of URLs?

doc/rest-api.yaml
```diff
@@ -1,11 +1,34 @@
...
+            entitlements:
+                description: Entitlements that are granted to the requesting user on the attached entity.
+                example:
+                    - can_view
+                    - can_edit
+                items:
+                    type: string
+                type: array
+                x-go-name: Entitlements
+        title: AttachedEntitlements is part of any LXD entity that may have entitlements attached to it.
```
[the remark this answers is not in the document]
gabrielmougard: `AttachedEntitlements` is always part of any LXD entity yes (by default, this is always empty if we don't specify the `with-entitlements=true` in the URL query param)

lxd/api_project.go
```diff
@@ -200,6 +201,23 @@ func projectsGet(d *Daemon, r *http.Request) response.Response {
...
+		entitiesWithEntitlements := make([]entity.EntityWithEntitlements, len(apiProjects))
+		for i, p := range apiProjects {
+			entitiesWithEntitlements[i] = p
+		}
+
+		err = d.authorizer.AddEntitlementsToEntities(r.Context(), entity.TypeProject, openFGAurls, entitiesWithEntitlements)
+		if err != nil {
+			return response.SmartError(err)
+		}
+	}
```
markylaing, line 219: I think expanding the `EntityWithEntitlements` interface would be useful in simplifying the logic at the call site as we wouldn't have to build the list of URLs.

markylaing (changes requested):
Looking good overall. Just a few tweaks and we'll be nearly there I think.

lxd/db/openfga/openfga.go
gabrielmougard: Ok, fixed

lxd/db/openfga/openfga.go
gabrielmougard: yes ok. fixed

lxd/db/openfga/openfga.go
gabrielmougard: No real reason. Removing this line. Fixed

lxd/daemon.go
gabrielmougard: But what happens if this is an API call from an other cluster member that already created this cache in the request context? This check would avoid re-creating the cache right?

lxd/identities.go
gabrielmougard: no reason :) fixed

lxd/auth/types.go
gabrielmougard: I like the the idea but I'd prefer to call the `URL() string` method `EntityURL() string` instead to be more explicit. For example, the URL scheme used for an LXD instance can be defined like:
`url := api.NewURL().Path(version.APIVersion, instancePath, instance.Name).Project(instance.Project)`
and if `instance.Project` is `default` then no query param will be added whereas an OpenFGA entity URL will systematically have the `project` query param set even if its `default`.

gabrielmougard: @markylaing Actually, I realize this will be a bit more complicated than that:

There is an existing logic for URL generation in the `shared/entity` package, but I won't be able to use that from the `shared/api` package (circular dependency issues) where my implementations of `EntityWithEntitlements` are defined...

gabrielmougard: @markylaing can't we just move the `url.go` file from `shared/entity` into the `shared/api` package? We'd keep the `type.go` in `shared/entity`. This way we can reuse the existing `<Entity>URL` functions for the API LXD entity that would implement  the `EntityWithEntitlements` interface. I think this would make the circular dep issues go away..

gabrielmougard: This seems difficult. Instead could we do something like:

```go
type EntityWithEntitlements interface {
	AddEntitlement(entitlement string)
	EntityType() Type
}

type EntityWithEntitlementsAndURL struct {
	Entity EntityWithEntitlements
	EntityURL    *api.URL
}

...

// In the interface
AddEntityEntitlements(ctx context.Context, entity entity.EntityWithEntitlementsAndURL)
AddEntitiesEntitlements(ctx context.Context, entities []*entity.EntityWithEntitlementsAndURL)
```

This way, we can set the EntityURL in the `lxd` package when calling the `AddEntityEntitlements` / `AddEntitiesEntitlements` function without having any issues. What do you think?

markylaing: Can we just define the interface somewhere else? In fact, it is only used internally, so perhaps it is better placed in `lxd/auth`? Would that help with the circular dependencies?

lxd/daemon.go
markylaing: The approach looks good :) It might be possible to make the new Authorizer functions more efficient by calling `ListObjects` and `Check` on the OpenFGA server directly. We can revisit that at a later date though, for now re-using `CheckPermission` and `GetPermissionChecker` is reasonable.

lxd/daemon.go
markylaing: If the request is forwarded by the other member then the cache was in-memory in the other member and is no longer available in the request context.

lxd/instance_get.go
gabrielmougard: in this endpoint (`GET /1.0/instances/{name}`), recusion=0 return an `*api.Instance` (with recursion=1, `*api.InstanceFull` is returned)

lxd/db/openfga/openfga.go
gabrielmougard: fixed

lxd/instance_get.go
gabrielmougard: fixed

lxd/instance_get.go
gabrielmougard: fixed

tomponline:
@gabrielmougard closing for now until you can work on this in the new year.

PR #14479, opened 2024-11-18
VM: Handle not found file descriptor

hamistao:
This simply adds a check for whether a file descriptor with the provided name exists and return an error if it doesn't.


lxd/instance/drivers/qmp/commands.go
```diff
@@ -249,6 +253,11 @@ func (m *Monitor) RemoveFDFromFDSet(name string) error {
 		}
 	}
 
+	// Return an error if no fd matched the provided name.
+	if !found {
+		return fmt.Errorf("No fd with name %q", name)
```
tomponline, line 258: This should use `api.StatusErrorf(http.StatusNotFound, "No fd with name %q", name)` so that the caller can detect this error using `api.StatusErrorCheck(err, http.StatusNotFound) ` and ignore it if needed.

hamistao: Done!

PR #14490, opened 2024-11-18
Safer path operations

simondeziel:
Tentatively addresses https://github.com/canonical/lxd/security/code-scanning?query=path%3Alxd%2Finstance_logs.go+is%3Aopen+branch%3Amain+

### DIFF
--- a/lxd/project/project.go
+++ b/lxd/project/project.go
@@ -18,7 +18,7 @@ const separator = "_"
 // Instance adds the "<project>_" prefix to instance name when the given project name is not "default".
 func Instance(projectName string, instanceName string) string {
 	if projectName != api.ProjectDefaultName {
-		return fmt.Sprintf("%s%s%s", projectName, separator, instanceName)
+		return projectName + separator + instanceName
 	}
 
 	return instanceName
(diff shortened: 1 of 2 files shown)


kadinsayani (approved):

tomponline (changes requested):
To avoid the duplication, and to help with unit tests etc, can this logic be moved into a function like `FilePathSafeJoin(prefix string, items ...string)` or similar which can then be called from each function instead of `filepath.Join`?

tomponline (changes requested):
Also this change only protects against a particular class of errors, but really we should *also* be checking the input is valid higher up the call chain where we know the context of the input (e.g. its an instance name).

So for instance log file delete for example, there's

```go
err = os.Remove(shared.LogPath(project.Instance(projectName, name), file))
```

But `name` should be replaced with `inst.Name()`, because we are already loading the instance from the DB using the `name` input, so that is itself de-coupling the raw input from the validated input.

Going further, there is also `inst.LogPath()` which does this for you. :)

We could also update `validLogFileName` to return the validated file name, in order to de-couple the raw input `file` from the file name passed to `os.Remove`.

tomponline:
@simondeziel can we close this for now until you get chance to look at this again?

PR #14493, opened 2024-11-19
github: Move chmod of go.mod, go.sum before go mod download

tomponline:
Sometimes it needs to modify them.

Related to fixing tests on https://github.com/canonical/lxd/pull/14388

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -79,13 +79,13 @@ jobs:
       - name: Download go dependencies
         run: |
           set -eux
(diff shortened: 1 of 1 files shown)


simondeziel (approved):
Would it make sense to do add a `git diff -- go.mod` once we are done with all the `go mod` invocations?

tomponline:
> Would it make sense to do add a `git diff -- go.mod` once we are done with all the `go mod` invocations?

Yeah that would be useful potentially.

PR #14498, opened 2024-11-20
github: stop stripping libs and binaries

simondeziel:
There is no real speed benefit to stripping and no known downside to keeping debug symbols. The only minor change is going to be bigger cached artifacts.

On the plus side though, having debug symbols will make crash investigations much easier.


markylaing (approved):
LGTM Thanks

tomponline (approved):

PR #14499, opened 2024-11-20
TEST: Force a DQLite crash and get a core dump

markylaing:


mihalicyn:
Please, can you also revert this https://github.com/canonical/lxd/commit/a80b942d4c9e1579bd26efacef98283148105df8 as a part of this PR. As we don't need this anymore and it's not valid (as after our `core_pattern` change our coredumps are not text files anymore).

test/main.sh
```diff
@@ -113,13 +113,6 @@ cleanup() {
     if [ "${expandDmesg}" = "no" ]; then
       echo "::endgroup::"
     fi
-
-    if [ "$(ls -A /var/crash)" ]; then
```
mihalicyn, line 117: That's what I was about to propose ;-)

mihalicyn (approved):

tomponline:
Ready for rebase

markylaing:
I've rebased, split the commits a bit better, and actually reverted a80b942d4c9e1579bd26efacef98283148105df8 rather than just deleting the block. Should be ready to go :)

tomponline (approved):

PR #14502, opened 2024-11-21
doc: fix defaultdesc format for instance configuration related to boot

edlerd:
# Done
- fix defaultdesc format for instance configuration related to boot for the documentation and metadata api


tomponline (approved):
Ta

PR #14504, opened 2024-11-21
github: Set `GOTRACEBACK=crash` on system test runs.

markylaing:
Sets `GOTRACEBACK=crash` as a follow up to #14499 because in @mihalicyn's words

> yeah, as withoutGOTRACEBACK="crash" we won't get a coredump at all, as golang runtime handles segfaults gracefully and Linux kernel won't realize that coredump collecting is required.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -209,6 +209,7 @@ jobs:
       LXD_VERBOSE: "1"
       LXD_OFFLINE: "1"
       LXD_TMPFS: "1"
+      GOTRACEBACK: "crash"
     name: System
     runs-on: ubuntu-22.04
     needs: code-tests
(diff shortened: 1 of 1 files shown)


markylaing:
Note that I did have this set when looking for coredumps in previous runs. Including when we set `ulimit` before running tests. We missed this as an oversight in #14499

.github/workflows/tests.yml
```diff
@@ -209,6 +209,7 @@ jobs:
       LXD_VERBOSE: "1"
       LXD_OFFLINE: "1"
       LXD_TMPFS: "1"
+      GOTRACEBACK: "crash"
```
tomponline, line 212: So this will prevent Go based panics from displaying a stack trace in the log and we will need to the manually inspect the core dump files? Or will it do both?

mihalicyn: On my local environment I can still see everything in logs and in addition to that I'm getting a coredump.
Without this parameter I coredumps are not generated. `go version go1.23.3 linux/amd64`

tomponline (approved):

.github/workflows/tests.yml
markylaing: I found a description of the different settings here: https://dave.cheney.net/tag/gotraceback

With `GOTRACEBACK=crash` we should get everything outputted and additionally a core dump

PR #14506, opened 2024-11-21
Snapcraft: Improved LXD Snap Completions

kadinsayani:
Fixes the extraneous space issue raised in https://github.com/canonical/lxd/issues/14264.

This PR adds `snap_completer.sh`, which provides improved bash completions for the LXD snap. Rather than generating the completion script during the LXD Snap `override-build` step and stream editing the file to work with `snapd` completions, including a modified script is more reliable. Moving forward, completions for the LXC CLI can be added/changed directly in `lxc/completion.go` with a guarantee that Cobra completion directives (such as `ShellCompDirectiveNoSpace` and `ShellCompDirectiveNoFileComp`) will get along with `snapd`. 

For reference, here is the [associated PR](https://github.com/canonical/lxd-pkg-snap/pull/629) for `lxd-pkg-snap`.

### DIFF
--- a/scripts/bash/snap_completer.sh
+++ b/scripts/bash/snap_completer.sh
@@ -1,0 +1,333 @@
+# shellcheck shell=bash
+
+# Derived from: https://github.com/spf13/cobra/blob/main/bash_completionsV2.go
+# Modifications made by Canonical Ltd, 2024
+# Original work is licensed under the Apache License, Version 2.0
+
+# bash completion V2 for lxc
+# This script provides bash completions for the LXD Snap. Snap completions are achieved by a marshalling script in snapd. Completion requests are intercepted and scanned before serialization by complete.sh, sent to etelpmoc.sh for de-serialization and processing, and finally, results are serialized. See https://github.com/canonical/snapd/blob/master/data/completion/bash/etelpmoc.sh and https://github.com/canonical/snapd/blob/master/data/completion/bash/complete.sh for detailed descriptions on the snapd bash completion process.
+
+__lxc_debug()
+{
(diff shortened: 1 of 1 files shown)


simondeziel (commented):
As annoying/agonizing as it is, I think we could still get away with keeping the `snapcraft` build time inline patching if we add the following (albeit ugly) bit:

```
lxc completion bash | sed ... -e 's|$(type -t compopt)|"builtin"|' ... 
```

I've put up https://github.com/canonical/lxd-pkg-snap/pull/630 to that effect, still untested though.

kadinsayani:
> As annoying/agonizing as it is, I think we could still get away with keeping the `snapcraft` build time inline patching if we add the following (albeit ugly) bit:
> 
> ```
> lxc completion bash | sed ... -e 's|$(type -t compopt)|"builtin"|' ... 
> ```
> 
> I've put up [canonical/lxd-pkg-snap#630](https://github.com/canonical/lxd-pkg-snap/pull/630) to that effect, still untested though.

I’ll give this a try tomorrow. If both approaches work, it’s just a matter of preference. I like using the modified script in this PR because it has comments about my findings throughout the debugging process with the hope that future issues will be easier to resolve.

kadinsayani:
> I've put up [canonical/lxd-pkg-snap#630](https://github.com/canonical/lxd-pkg-snap/pull/630) to that effect, still untested though.

Thanks for putting this up :)

kadinsayani:
Closing in favour of https://github.com/canonical/lxd-pkg-snap/pull/630. Thanks @simondeziel 😄

PR #14513, opened 2024-11-24
Auth: Fine-grained permission request cache

markylaing:
I was playing with some ideas to get some quick wins for performance in the fine-grained authorization driver.

The OpenFGADatastore receives very specific queries such as:
- Which groups are related to `instance:/1.0/instances/c1?project=default` as `can_view`?
- Which objects of type `instance` are members of group `foo` related to as `can_view`?

Each of these queries are executed case-by-case. This always hits the database, even if there are no permissions at all for entities of type `instance`.

This PR implements pre-loading of the `auth_groups_permissions` table to reduce the total number of database calls. Qualitatively, I've seen this reduce the total number of database calls when listing projects with recursion from 85 to 10.

@gabrielmougard it might be worth comparing this with your work in #14476. We can test performance and try to make further optimisations so that loading the additional entitlements is low-cost.


gabrielmougard:
@markylaing thanks! Taking a look

markylaing:
Closing in favour of #14557

PR #14514, opened 2024-11-25
lxd: Log error on `resultErrListAppend`

hamistao:
I noticed this `err` parameter was unused. So I had this idea of, instead of removing it, logging these erros, as the errors used with that function are currently not being disclosed to the user in any way. This way it is clear why one was not able to get info on that particular instance when looking in the logs. If this logging is not desirable, I will update the PR to just remove the unused parameter.

### DIFF
--- a/lxd/instances_get.go
+++ b/lxd/instances_get.go
@@ -25,6 +25,7 @@ import (
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/entity"
 	"github.com/canonical/lxd/shared/filter"
+	"github.com/canonical/lxd/shared/logger"
 	"github.com/canonical/lxd/shared/version"
 )
 
@@ -304,6 +305,8 @@ func instancesGet(d *Daemon, r *http.Request) response.Response {
 	}
 
 	resultErrListAppend := func(inst db.Instance, err error) {
+		logger.Error("Failed getting instance info", logger.Ctx{"err": err, "project": inst.Project, "instance": inst.Name})
+
 		instFull := &api.InstanceFull{
 			Instance: api.Instance{
 				Name:       inst.Name,


lxd/instances_get.go
```diff
@@ -304,6 +305,8 @@ func instancesGet(d *Daemon, r *http.Request) response.Response {
 	}
 
 	resultErrListAppend := func(inst db.Instance, err error) {
+		logger.Errorf("Failed getting instance %s info: %v", inst.Name, err)
```
MggMuggins, line 308: Can you use `logger.Error` here with a `logger.Ctx`? I think the eventual goal is to remove all uses of `logger.*f`.

MggMuggins (changes requested):
Good idea. (For future, not now) I wonder if this could be refactored to use the cluster notifier instead.

lxd/instances_get.go
hamistao: I was wondering if that was the case, thanks!

MggMuggins (dismissed):
LGTM!

tomponline:
> Good idea. (For future, not now) I wonder if this could be refactored to use the cluster notifier instead.

yeah probably.

lxd/instances_get.go
```diff
@@ -304,6 +305,8 @@ func instancesGet(d *Daemon, r *http.Request) response.Response {
 	}
 
 	resultErrListAppend := func(inst db.Instance, err error) {
+		logger.Error("Failed getting instance info", logger.Ctx{"err": err, "instance": inst.Name})
```
tomponline, line 308: we should log `project` in the error context too otherwise its not clear which instance we are logging about.

`project:` comes first before `instance:`.

hamistao: Done!

Issue #14125, opened 2024-09-18
Metadata API to indicate cluster member specificity


edlerd:
# Required information

 * Distribution: snap
 * Distribution version: 6.1

# Issue description

The `GET 1.0/metadata/configuration` response should include an indicator if a field is global or cluster member specific, when LXD is clustered. This information should also be added to the documentation.

For instance, the key `config.network-physical.network-conf.keys[18].parent` has no notion of being specific per cluster member. When LXD is clustered, the key cannot be set without targetting a specific cluster member. See screenshot below for its current value.

![image](https://github.com/user-attachments/assets/809e7a25-dd93-4975-a19b-709d46533dfd)

PR #14530, opened 2024-11-25
API: Add `scope` to metadata configuration API

kadinsayani:
Resolves https://github.com/canonical/lxd/issues/14125.

This PR adds `scope` to the metadata configuration API endpoint, `/1.0/metadata/configuration`.

Summary of changes:
- Adds api extension, `metadata_configuration_scope`;
- Adds `Scope` field to `MetadataConfigurationConfigKey` struct;
- Adds `scope` to metadata for member specific config keys in `lxd/network/driver_{bridge/macvlan/physical/sriov}.go`;
- Updates API docs and metadata.

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2523,3 +2523,7 @@ Adds support for using the Container Device Interface (CDI) specification to con
 ## `images_all_projects`
 
 This adds support for listing images across all projects using the `all-projects` parameter in `GET /1.0/images` requests.
+
+## `metadata_configuration_scope`
+
+This adds scope metadata to `GET /1.0/metadata/configuration`. Options marked with a `global` scope are applied to all cluster members. Options with a `local` scope must be set on a per-member basis.
--- a/lxd/network/driver_bridge.go
+++ b/lxd/network/driver_bridge.go
@@ -203,6 +203,7 @@ func (n *bridge) Validate(config map[string]string) error {
 		//  type: string
 		//  condition: BGP server
 		//  shortdesc: Peer address (IPv4 or IPv6)
+		//  scope: global
 
 		// lxdmeta:generate(entities=network-bridge; group=network-conf; key=bgp.peers.NAME.asn)
 		//
@@ -210,6 +211,7 @@ func (n *bridge) Validate(config map[string]string) error {
 		//  type: integer
 		//  condition: BGP server
 		//  shortdesc: Peer AS number
+		//  scope: global
 
 		// lxdmeta:generate(entities=network-bridge; group=network-conf; key=bgp.peers.NAME.password)
 		//
@@ -219,6 +221,7 @@ func (n *bridge) Validate(config map[string]string) error {
 		//  defaultdesc: (no password)
 		//  required: no
 		//  shortdesc: Peer session password
+		//  scope: global
 
 		// lxdmeta:generate(entities=network-bridge; group=network-conf; key=bgp.peers.NAME.holdtime)
 		// Specify the hold time in seconds.
@@ -228,6 +231,7 @@ func (n *bridge) Validate(config map[string]string) error {
 		//  defaultdesc: `180`
 		//  required: no
 		//  shortdesc: Peer session hold time
+		//  scope: global
 
 		// lxdmeta:generate(entities=network-bridge; group=network-conf; key=bgp.ipv4.nexthop)
 		//
@@ -236,6 +240,7 @@ func (n *bridge) Validate(config map[string]string) error {
 		//  condition: BGP server
 		//  defaultdesc: local address
 		//  shortdesc: Override the IPv4 next-hop for advertised prefixes
+		//  scope: local
 		"bgp.ipv4.nexthop": validate.Optional(validate.IsNetworkAddressV4),
 		// lxdmeta:generate(entities=network-bridge; group=network-conf; key=bgp.ipv6.nexthop)
 		//
@@ -244,6 +249,7 @@ func (n *bridge) Validate(config map[string]string) error {
 		//  condition: BGP server
 		//  defaultdesc: local address
 		//  shortdesc: Override the IPv6 next-hop for advertised prefixes
+		//  scope: local
 		"bgp.ipv6.nexthop": validate.Optional(validate.IsNetworkAddressV6),
 		// lxdmeta:generate(entities=network-bridge; group=network-conf; key=bridge.driver)
 		// Possible values are `native` and `openvswitch`.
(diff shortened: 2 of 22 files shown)


kadinsayani:
cc. @edlerd

doc/api-extensions.md
```diff
@@ -2519,3 +2519,7 @@ Adds a new {config:option}`instance-resource-limits:limits.cpu.pin_strategy` con
 ## `gpu_cdi`
 
 Adds support for using the Container Device Interface (CDI) specification to configure GPU passthrough in LXD containers. The `id` field of GPU devices now accepts CDI identifiers (for example, `{VENDOR_DOMAIN_NAME}/gpu=gpu{INDEX}`) for containers, in addition to DRM card IDs. This enables GPU passthrough for devices that don't use PCI addressing (like NVIDIA Tegra iGPUs) and provides a more flexible way to identify and configure GPU devices.
+
+## metadata_configuration_scope
+
+This adds scope metadata to `GET /1.0/metadata/configuration`. Options marked with a `global` scope are immediately applied to all cluster members. Options with a `local` scope must be set on a per-member basis.
```
markylaing, line 2525: I think we can drop "immediately" from this sentence.

shared/api/metadata_configuration.go
```diff
@@ -64,6 +64,12 @@ type MetadataConfigurationConfigKey struct {
 	//
 	// Example: yes.
 	Managed string `json:"managed"`
+
+	// Scope describes the cluster member specificity of a configuration key. Options marked with a `global` scope are immediately applied to all cluster members. Options marked with a `local` scope are set on a per-member basis.
```
markylaing, line 68: Again here.

lxd/network/driver_bridge.go
```diff
@@ -273,20 +281,23 @@ func (n *bridge) Validate(config map[string]string) error {
 		// ---
 		//  type: string
 		//  shortdesc: MAC address for the bridge
+		//  scope: global
```
markylaing, line 284: I don't think MAC addresses can be global.

markylaing (changes requested):
We'll need @tomponline or @simondeziel to review the local/global-ness of the config keys but other than that I just have a couple of small nits :)

lxd/network/driver_bridge.go
kadinsayani: Here's my thinking: since all cluster members need to have an identical defined network, the bridge MAC address shouldn't differ on a per-member basis.

simondeziel: The bridge MAC address outta be unique per member for proper function. Otherwise you get conflicting MAC addresses in your L2 network.

tomponline: If it cant be set per-member today then its global currently.

Whether that is appropriate or not is the matter of a separate GH issue :)

kadinsayani: A quick test: 

```bash
❯ lxc network create foo --type=bridge --target=devbox bridge.hwaddr=00:00:00:00:00:00
Error: Config key "bridge.hwaddr" may not be used as member-specific key
```

kadinsayani: I've opened https://github.com/canonical/lxd/issues/14548 to track.

lxd/network/driver_macvlan.go
```diff
@@ -29,46 +29,53 @@ func (n *macvlan) Validate(config map[string]string) error {
...
 		//  type: string
 		//  shortdesc: Parent interface to create `macvlan` NICs on
+		//  scope: local
 		"parent": validate.Required(validate.IsNotEmpty, validate.IsInterfaceName),
 		// lxdmeta:generate(entities=network-macvlan; group=network-conf; key=mtu)
 		//
 		// ---
 		//  type: integer
 		//  shortdesc: MTU of the new interface
+		//  scope: global
```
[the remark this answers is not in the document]
tomponline: There maybe some confusion around the scope (pun intended) of this PR.

It is to update the metdata API with the current scope of each setting.

The definition for "local" scope in the context of the docs and metadata API is whether the config key in question can be defined on a per-cluster member basis.

As instance devices are always tied to an instance, the value of their settings applies irrespective to which cluster member they are running on, and thus instance and device settings are inherently global.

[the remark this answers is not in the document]
simondeziel: I'm probably confused but I understood this `mtu` config to be for a `macvlan` bridge, not an instance's NIC device.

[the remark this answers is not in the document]
tomponline: Ah yeah true, i somehow saw this as for the macvlan device.

Indeed `parent` should likely be local indeed, as it is the network indirection concept that allows an instance NIC device referencing a network to take on per-cluster member settings.

So for this PR specifically we are really only interested on what is currently implemented as local/global, rather than potential future changes. 

@kadinsayani there are 2 variables you'll want to look at for this:

```go
// NodeSpecificNetworkConfig lists all network config keys which are node-specific.
var NodeSpecificNetworkConfig = []string{
	"bgp.ipv4.nexthop",
	"bgp.ipv6.nexthop",
	"bridge.external_interfaces",
	"parent",
}
```

and 

```go
// NodeSpecificStorageConfig lists all storage pool config keys which are node-specific.
var NodeSpecificStorageConfig = []string{
	"size",
	"source",
	"source.wipe",
	"volatile.initial_source",
	"zfs.pool_name",
	"lvm.thinpool_name",
	"lvm.vg_name",
}
```

lxd/network/driver_macvlan.go
```diff
@@ -29,46 +29,53 @@ func (n *macvlan) Validate(config map[string]string) error {
...
 		"mtu": validate.Optional(validate.IsNetworkMTU),
 		// lxdmeta:generate(entities=network-macvlan; group=network-conf; key=vlan)
 		//
 		// ---
 		//  type: integer
 		//  shortdesc: VLAN ID to attach to
+		//  scope: global
```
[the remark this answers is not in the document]
tomponline: right nows its not

lxd/network/driver_macvlan.go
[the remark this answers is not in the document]
simondeziel: > So for this PR specifically we are really only interested on what is currently implemented as local/global, rather than potential future changes.

Yes indeed. But it's nice that Mark caught a discrepancy which will be followed through :)

lxd/storage/drivers/driver_powerflex.go
```diff
@@ -261,6 +270,7 @@ func (d *powerflex) Validate(config map[string]string) error {
 		//  type: string
 		//  defaultdesc: `8GiB`
 		//  shortdesc: Size/quota of the storage volume
+		//  scope: local
```
tomponline, line 273: this doesnt look right

lxd/storage/drivers/driver_powerflex_volumes.go
```diff
@@ -454,6 +457,7 @@ func (d *powerflex) commonVolumeRules() map[string]func(value string) error {
 		//  type: string
 		//  defaultdesc: same as `volume.size`
 		//  shortdesc: Size/quota of the storage volume
+		//  scope: local
```
tomponline, line 460: check this please

tomponline (approved):
Thanks!

tomponline:
@edlerd this will be in latest/edge soon.

PR #14491, opened 2024-11-18
Storage: Specify storage volume type in disk device `source`

MggMuggins:
This change lays the groundwork for attaching VM root disk devices to instances by making two changes:

### Refactor
Renames `(Un)MountCustomVolume` -> `(Un)MountVolume` to allow for generic activation of `virtual-machine` and `container` volumes.

### Specify volume type in disk device
Allows the following syntax for specifying `custom` storage volumes in disk devices:
```
devices:
  cosmic-dust-analyzer-data:
	pool: default
	source: custom/cosmic-dust-analyzer-data
	type: disk
```

(I was incorrect in my assessment that we already support this syntax; the code was there but wasn't being used consistently)

PR #13821, opened 2024-07-25
Instance: Fix deadlock during failed snapshot creation

MggMuggins:
Fixes #13466

We need a non-locking delete on an `Instance`; otherwise a failure while mounting the instance/updating the backup file will deadlock.

I don't love the type assertion here. However, since `CreateInternal` returns `Instance`, we don't have a lot of options beyond enriching that interface or providing an additional one.

In order to avoid using type assertions we'd need to define an internal interface `instance` for `delete` and possibly a few other functions in the drivers/ package and implement it for `lxc`/`qemu`. Alternatively, moving the `Interface` interface into the same package as the drivers would allow us to define some private methods on `Interface` instead.

(Updating the backup file on the full volume was very likely the failure in #13466; this is after applying the fix):
```
$ lxc snapshot lucky-bulldog
Error: Failed to create file "/var/lib/lxd/containers/microcloud-test_lucky-bulldog/backup.yaml": open /var/lib/lxd/containers/microcloud-test_lucky-bulldog/backup.yaml: disk quota exceeded
```

I will throw a snapshot failure test on here this afternoon.

LXD-1117

PR #14532, opened 2024-11-26
Attach VM root volumes as disk devices

MggMuggins:
Includes commits from #14491 

This PR enables attaching a virtual machine's root storage volume to another virtual machine via a disk device:
```
# vm2 yaml
...
devices:
  v1-root:
    type: disk
    pool: default
    source: virtual-machine/vm1
```

This has some constraints because simultaneous access to storage volumes with content-type `block` is unsafe:
- `vm1`'s root volume can be attached to exactly one other instance if `vm1` has `security.protection.start: true`
- `vm1`'s root volume can be attached to any number of other instances if the storage volume `virtual-machine/vm1` has `security.shared: true`

`security.protection.start` is recommended for interactive use; e.g. a user temporarily needs to access a bricked machine's root volume to fix it or recover data. `security.shared` can be used if more than one running instance must have access to the block volume.

## TODO
- [x] Docs
- [x] [lxd-ci](https://github.com/canonical/lxd-ci/) tests to validate attachments to running VMs. [Done](https://github.com/canonical/lxd-ci/pull/366)
- [x] Bootorder bug (see comment) - current workaround is to hotplug disk devices to avoid UUID/LABEL collisions at boot time.

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2546,3 +2546,9 @@ Adds optional `target` parameter to `GET /1.0/network`. When target is set, forw
 ## `network_zones_all_projects`
 
 This adds support for listing network zones across all projects using the `all-projects` parameter in `GET /1.0/network-zones` requests.
+
+## `instance_root_volume_attachment`
+
+Adds support for instance root volumes to be attached to other instances as disk
+devices. Introduces the `<type>/<volume>` syntax for the `source` property of
+disk devices.
--- a/doc/explanation/storage.md
+++ b/doc/explanation/storage.md
@@ -120,10 +120,10 @@ Storage volumes can be of the following types:
 
 `container`/`virtual-machine`
 : LXD automatically creates one of these storage volumes when you launch an instance.
-  It is used as the root disk for the instance, and it is destroyed when the instance is deleted.
+  It is used as the root disk for the instance and is destroyed when the instance is deleted.
 
-  This storage volume is created in the storage pool that is specified in the profile used when launching the instance (or the default profile, if no profile is specified).
-  The storage pool can be explicitly specified by providing the `--storage` flag to the launch command.
+  The storage pool can be explicitly specified by providing the `--storage` flag to the {ref}`launch command <lxc_launch.md>`.
+  If no pool or profile is specified, LXD uses the storage pool of the default profile's root disk device.
 
 `image`
 : LXD automatically creates one of these storage volumes when it unpacks an image to launch one or more instances from it.
@@ -157,7 +157,7 @@ Each storage volume uses one of the following content types:
 
   Custom storage volumes of content type `block` can only be attached to virtual machines.
   By default, they can only be attached to one instance at a time, because simultaneous access can lead to data corruption.
-  Sharing a custom storage volumes of content type `block` is made possible through the usage of the `security.shared` configuration key.
+  Sharing custom storage volumes of content type `block` is made possible through the usage of the `security.shared` configuration key.
 
 `iso`
 : This content type is used for custom ISO volumes.
--- a/doc/howto/storage_volumes.md
+++ b/doc/howto/storage_volumes.md
@@ -179,6 +179,34 @@ For example, to set the default volume size for `my-pool`, use the following com
 
     lxc storage set my-pool volume.size=15GiB
 
+## Attach instance root volumes to other instances
+Virtual-machine root volumes can be attached as disk devices to other virtual machines.
+In order to prevent concurrent access, `security.protection.start` must be set on
+an instance before its root volume can be attached to another virtual-machine.
+
+```{caution}
+Because instances created from the same image share the same partition and file system
+UUIDs and labels, booting an instance with two root file systems mounted may result
+in the wrong root file system being used. This may result in unexpected behavior
+or data loss. **It is strongly recommended to only attach virtual-machine root
+volumes to other virtual machines when the target virtual-machine is running.**
+```
+
+Assuming `vm1` is stopped and `vm2` is running, attach the `virtual-machine/vm1` storage
+volume to `vm2`:
+
+    lxc config set vm1 security.protection.start=true
+    lxc storage volume attach my-pool virtual-machine/vm1 vm2
+
+`virtual-machine/vm1` must be detached from `vm2` before `security.protection.start`
+can be unset from `vm1`:
+
+    lxc storage volume detach my-pool virtual-machine/vm1 vm2
+    lxc config unset vm1 security.protection.start
+
+`security.shared` can also be used on `virtual-machine` volumes to enable concurrent
+access. Note that concurrent access to block volumes may result in data loss.
+
 ## Resize a storage volume
 
 If you need more storage in a volume, you can increase the size of your storage volume.
--- a/lxc/storage_volume.go
+++ b/lxc/storage_volume.go
@@ -164,10 +164,12 @@ type cmdStorageVolumeAttach struct {
 
 func (c *cmdStorageVolumeAttach) command() *cobra.Command {
 	cmd := &cobra.Command{}
-	cmd.Use = usage("attach", i18n.G("[<remote>:]<pool> <volume> <instance> [<device name>] [<path>]"))
+	cmd.Use = usage("attach", i18n.G("[<remote>:]<pool> [<type>/]<volume> <instance> [<device name>] [<path>]"))
 	cmd.Short = i18n.G("Attach new storage volumes to instances")
 	cmd.Long = cli.FormatSection(i18n.G("Description"), i18n.G(
-		`Attach new storage volumes to instances`))
+		`Attach new storage volumes to instances
+
+<type> must be one of "custom" or "virtual-machine"`))
 
 	cmd.RunE = c.run
 
@@ -210,8 +212,8 @@ func (c *cmdStorageVolumeAttach) run(cmd *cobra.Command, args []string) error {
 	}
 
 	volName, volType := parseVolume("custom", args[1])
-	if volType != "custom" {
-		return errors.New(i18n.G("Only \"custom\" volumes can be attached to instances"))
+	if volType != "custom" && volType != "virtual-machine" {
+		return errors.New(i18n.G(`Only "custom" and "virtual-machine" volumes can be attached to instances`))
 	}
 
 	// Attach the volume
@@ -257,7 +259,7 @@ func (c *cmdStorageVolumeAttach) run(cmd *cobra.Command, args []string) error {
 	device := map[string]string{
 		"type":   "disk",
 		"pool":   resource.name,
-		"source": volName,
+		"source": args[1],
 		"path":   devPath,
 	}
 
@@ -279,10 +281,12 @@ type cmdStorageVolumeAttachProfile struct {
 
 func (c *cmdStorageVolumeAttachProfile) command() *cobra.Command {
 	cmd := &cobra.Command{}
-	cmd.Use = usage("attach-profile", i18n.G("[<remote:>]<pool> <volume> <profile> [<device name>] [<path>]"))
+	cmd.Use = usage("attach-profile", i18n.G("[<remote:>]<pool> [<type>/]<volume> <profile> [<device name>] [<path>]"))
 	cmd.Short = i18n.G("Attach new storage volumes to profiles")
 	cmd.Long = cli.FormatSection(i18n.G("Description"), i18n.G(
-		`Attach new storage volumes to profiles`))
+		`Attach new storage volumes to profiles
+
+<type> must be one of "custom" or "virtual-machine"`))
 
 	cmd.RunE = c.run
 
@@ -340,8 +344,8 @@ func (c *cmdStorageVolumeAttachProfile) run(cmd *cobra.Command, args []string) e
 	}
 
 	volName, volType := parseVolume("custom", args[1])
-	if volType != "custom" {
-		return errors.New(i18n.G("Only \"custom\" volumes can be attached to instances"))
+	if volType != "custom" && volType != "virtual-machine" {
+		return errors.New(i18n.G(`Only "custom" and "virtual-machine" volumes can be attached to profiles`))
 	}
 
 	// Check if the requested storage volume actually exists
@@ -354,7 +358,7 @@ func (c *cmdStorageVolumeAttachProfile) run(cmd *cobra.Command, args []string) e
 	device := map[string]string{
 		"type":   "disk",
 		"pool":   resource.name,
-		"source": vol.Name,
+		"source": args[1],
 	}
 
 	// Ignore path for block volumes
--- a/lxc/storage_volume.go
+++ b/lxc/storage_volume.go
@@ -807,7 +807,7 @@ type cmdStorageVolumeDetach struct {
 
 func (c *cmdStorageVolumeDetach) command() *cobra.Command {
 	cmd := &cobra.Command{}
-	cmd.Use = usage("detach", i18n.G("[<remote>:]<pool> <volume> <instance> [<device name>]"))
+	cmd.Use = usage("detach", i18n.G("[<remote>:]<pool> [<type>/]<volume> <instance> [<device name>]"))
 	cmd.Short = i18n.G("Detach storage volumes from instances")
 	cmd.Long = cli.FormatSection(i18n.G("Description"), i18n.G(
 		`Detach storage volumes from instances`))
@@ -865,14 +865,13 @@ func (c *cmdStorageVolumeDetach) run(cmd *cobra.Command, args []string) error {
 	}
 
 	volName, volType := parseVolume("custom", args[1])
-	if volType != "custom" {
-		return errors.New(i18n.G(`Only "custom" volumes can be attached to instances`))
-	}
 
 	// Find the device
 	if devName == "" {
 		for n, d := range inst.Devices {
-			if d["type"] == "disk" && d["pool"] == resource.name && d["source"] == volName {
+			sourceName, sourceType := parseVolume("custom", d["source"])
+
+			if d["type"] == "disk" && d["pool"] == resource.name && volType == sourceType && volName == sourceName {
 				if devName != "" {
 					return errors.New(i18n.G("More than one device matches, specify the device name"))
 				}
@@ -910,7 +909,7 @@ type cmdStorageVolumeDetachProfile struct {
 
 func (c *cmdStorageVolumeDetachProfile) command() *cobra.Command {
 	cmd := &cobra.Command{}
-	cmd.Use = usage("detach-profile", i18n.G("[<remote:>]<pool> <volume> <profile> [<device name>]"))
+	cmd.Use = usage("detach-profile", i18n.G("[<remote:>]<pool> [<type>/]<volume> <profile> [<device name>]"))
 	cmd.Short = i18n.G("Detach storage volumes from profiles")
 	cmd.Long = cli.FormatSection(i18n.G("Description"), i18n.G(
 		`Detach storage volumes from profiles`))
@@ -967,14 +966,13 @@ func (c *cmdStorageVolumeDetachProfile) run(cmd *cobra.Command, args []string) e
 	}
 
 	volName, volType := parseVolume("custom", args[1])
-	if volType != "custom" {
-		return errors.New(i18n.G(`Only "custom" volumes can be attached to instances`))
-	}
 
 	// Find the device
 	if devName == "" {
 		for n, d := range profile.Devices {
-			if d["type"] == "disk" && d["pool"] == resource.name && d["source"] == volName {
+			sourceName, sourceType := parseVolume("custom", d["source"])
+
+			if d["type"] == "disk" && d["pool"] == resource.name && volType == sourceType && volName == sourceName {
 				if devName != "" {
 					return errors.New(i18n.G("More than one device matches, specify the device name"))
 				}
--- a/lxd/device/disk.go
+++ b/lxd/device/disk.go
@@ -480,6 +480,17 @@ func (d *disk) validateConfig(instConf instance.ConfigReader) error {
 					return err
 				}
 
+				if d.inst != nil {
+					instVolType, err := storagePools.InstanceTypeToVolumeType(d.inst.Type())
+					if err != nil {
+						return err
+					}
+
+					if instVolType == volumeType && d.inst.Name() == volumeName {
+						return errors.New("Instance root device cannot be attached to itself")
+					}
+				}
+
 				// Derive the effective storage project name from the instance config's project.
 				storageProjectName, err = project.StorageVolumeProject(d.state.DB.Cluster, instConf.Project().Name, dbVolumeType)
 				if err != nil {
--- a/lxd/device/disk.go
+++ b/lxd/device/disk.go
@@ -147,7 +147,8 @@ func (d *disk) sourceIsLocalPath(source string) bool {
 	return true
 }
 
-// Check that unshared custom storage block volumes are not added to profiles or multiple instances.
+// Check that unshared custom storage block volumes are not added to profiles or
+// multiple instances unless they will not be accessed concurrently.
 func (d *disk) checkBlockVolSharing(instanceType instancetype.Type, projectName string, volume *api.StorageVolume) error {
 	// Skip the checks if the volume is set to be shared or is not a block volume.
 	if volume.ContentType != cluster.StoragePoolVolumeContentTypeNameBlock || shared.IsTrue(volume.Config["security.shared"]) {
@@ -155,7 +156,7 @@ func (d *disk) checkBlockVolSharing(instanceType instancetype.Type, projectName 
 	}
 
 	if instanceType == instancetype.Any {
-		return fmt.Errorf("Cannot add custom storage block volume to profiles if security.shared is false or unset")
+		return fmt.Errorf("Cannot add block volume to profiles if security.shared is false or unset")
 	}
 
 	err := storagePools.VolumeUsedByInstanceDevices(d.state, d.pool.Name(), projectName, volume, true, func(inst db.InstanceArgs, project api.Project, usedByDevices []string) error {
@@ -164,13 +165,24 @@ func (d *disk) checkBlockVolSharing(instanceType instancetype.Type, projectName 
 			return nil
 		}
 
-		return db.ErrListStop
-	})
-	if err != nil {
-		if err == db.ErrListStop {
-			return fmt.Errorf("Cannot add custom storage block volume to more than one instance if security.shared is false or unset")
+		// Don't count a VM volume's instance if security.protection.start is preventing that instance from starting.
+		// It's safe to share block volumes with an instance that cannot start.
+		if volume.Type == cluster.StoragePoolVolumeTypeNameVM && volume.Project == inst.Project && volume.Name == inst.Name {
+			apiInst, err := inst.ToAPI()
+			if err != nil {
+				return err
+			}
+
+			apiInst.ExpandedConfig = instancetype.ExpandInstanceConfig(d.state.GlobalConfig.Dump(), apiInst.Config, inst.Profiles)
+
+			if shared.IsTrue(apiInst.ExpandedConfig["security.protection.start"]) {
+				return nil
+			}
 		}
 
+		return fmt.Errorf("Cannot add block volume to more than one instance if security.shared is false or unset")
+	})
+	if err != nil {
 		return err
 	}
 
--- a/lxd/device/disk.go
+++ b/lxd/device/disk.go
@@ -1579,9 +1579,6 @@ func (d *disk) mountPoolVolume() (func(), string, *storagePools.MountInfo, error
 		return nil, "", nil, err
 	}
 
-	volStorageName := project.StorageVolume(storageProjectName, volumeName)
-	srcPath := storageDrivers.GetVolumeMountPath(d.config["pool"], volumeType, volStorageName)
-
 	mountInfo, err = d.pool.MountVolume(storageProjectName, volumeName, volumeType, nil)
 	if err != nil {
 		return nil, "", nil, fmt.Errorf(`Failed mounting storage volume "%s/%s" from storage pool %q: %w`, volumeTypeName, volumeName, d.pool.Name(), err)
@@ -1600,6 +1597,15 @@ func (d *disk) mountPoolVolume() (func(), string, *storagePools.MountInfo, error
 		return nil, "", nil, fmt.Errorf("Failed to fetch local storage volume record: %w", err)
 	}
 
+	var volStorageName string
+	if dbVolume.Type == cluster.StoragePoolVolumeTypeNameCustom {
+		volStorageName = project.StorageVolume(storageProjectName, volumeName)
+	} else {
+		volStorageName = project.Instance(storageProjectName, volumeName)
+	}
+
+	srcPath := storageDrivers.GetVolumeMountPath(d.config["pool"], volumeType, volStorageName)
+
 	if d.inst.Type() == instancetype.Container {
 		if dbVolume.ContentType != cluster.StoragePoolVolumeContentTypeNameFS {
 			return nil, "", nil, fmt.Errorf("Only filesystem volumes are supported for containers")
@@ -1612,8 +1618,6 @@ func (d *disk) mountPoolVolume() (func(), string, *storagePools.MountInfo, error
 	}
 
 	if dbVolume.ContentType == cluster.StoragePoolVolumeContentTypeNameBlock || dbVolume.ContentType == cluster.StoragePoolVolumeContentTypeNameISO {
-		volStorageName := project.StorageVolume(storageProjectName, volumeName)
-
 		volume := d.pool.GetVolume(volumeType, storageDrivers.ContentType(dbVolume.ContentType), volStorageName, dbVolume.Config)
 
 		srcPath, err = d.pool.Driver().GetVolumeDiskPath(volume)
--- a/lxd/instance/drivers/driver_common.go
+++ b/lxd/instance/drivers/driver_common.go
@@ -1697,6 +1697,55 @@ func (d *common) deleteSnapshots(deleteFunc func(snapInst instance.Instance) err
 	return nil
 }
 
+// checkRootVolumeNotInUse fails if the instance's root volume is in use on
+// another instance.
+func (d *common) checkRootVolumeNotInUse() error {
+	// Make sure that the instance's root volume is not attached to another instance
+	storagePool, err := d.getStoragePool()
+	if err != nil {
+		return err
+	}
+
+	rootVolumeType, err := storagePools.InstanceTypeToVolumeType(d.Type())
+	if err != nil {
+		return err
+	}
+
+	rootVolumeDBType, err := storagePools.VolumeTypeToDBType(rootVolumeType)
+	if err != nil {
+		return err
+	}
+
+	var rootVolume *db.StorageVolume
+	err = d.state.DB.Cluster.Transaction(d.state.ShutdownCtx, func(ctx context.Context, tx *db.ClusterTx) error {
+		rootVolume, err = tx.GetStoragePoolVolume(ctx, storagePool.ID(), d.Project().Name, rootVolumeDBType, d.Name(), true)
+		return err
+	})
+	if err != nil {
+		return err
+	}
+
+	err = storagePools.VolumeUsedByProfileDevices(d.state, storagePool.Name(), d.Project().Name, &rootVolume.StorageVolume, func(profileID int64, profile api.Profile, p api.Project, usedByDevices []string) error {
+		return fmt.Errorf(`"%s/%s" is attached to a profile`, rootVolume.Type, rootVolume.Name)
+	})
+	if err != nil {
+		return err
+	}
+
+	err = storagePools.VolumeUsedByInstanceDevices(d.state, storagePool.Name(), d.Project().Name, &rootVolume.StorageVolume, false, func(inst db.InstanceArgs, p api.Project, usedByDevices []string) error {
+		if inst.Name == d.Name() && inst.Project == d.Project().Name {
+			return nil
+		}
+
+		return fmt.Errorf(`"%s/%s" is attached to another instance`, rootVolume.Type, rootVolume.Name)
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // removeUnixDevices reads the devices path and remove all unix devices.
 func (d *common) removeUnixDevices() error {
 	// Check that we indeed have devices to remove
--- a/lxd/instance/drivers/driver_qemu.go
+++ b/lxd/instance/drivers/driver_qemu.go
@@ -5190,6 +5190,11 @@ func (d *qemu) Rename(newName string, applyTemplateTrigger bool) error {
 		return err
 	}
 
+	err = d.checkRootVolumeNotInUse()
+	if err != nil {
+		return err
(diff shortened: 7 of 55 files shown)


MggMuggins:
Currently when booting a VM with two root disk devices attached, the kernel chooses which partitions to mount at `/boot/efi`, `/boot` and `/` seemingly at random. I've seen it pick the correct disk for all three partitions, the incorrect disk for all three partitions, and any combination of the three:
```
# lsblk
NAME    MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
sda       8:0    0   10G  0 disk
├─sda1    8:1    0    9G  0 part /
├─sda14   8:14   0    4M  0 part
├─sda15   8:15   0  106M  0 part
└─sda16 259:0    0  913M  0 part
sdb       8:16   0   10G  0 disk
├─sdb1    8:17   0    9G  0 part
├─sdb14   8:30   0    4M  0 part
├─sdb15   8:31   0  106M  0 part /boot/efi
└─sdb16 259:1    0  913M  0 part /boot

# lsblk
NAME    MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
sda       8:0    0   10G  0 disk
├─sda1    8:1    0    9G  0 part /
├─sda14   8:14   0    4M  0 part
├─sda15   8:15   0  106M  0 part
└─sda16 259:0    0  913M  0 part /boot
sdb       8:16   0   10G  0 disk
├─sdb1    8:17   0  2.1G  0 part
├─sdb14   8:30   0    4M  0 part
└─sdb15   8:31   0  106M  0 part /boot/efi

# lsblk
NAME    MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
sda       8:0    0   10G  0 disk
├─sda1    8:1    0    9G  0 part
├─sda14   8:14   0    4M  0 part
├─sda15   8:15   0  106M  0 part
└─sda16 259:1    0  913M  0 part
sdb       8:16   0   10G  0 disk
├─sdb1    8:17   0    9G  0 part /
├─sdb14   8:30   0    4M  0 part
├─sdb15   8:31   0  106M  0 part /boot/efi
└─sdb16 259:0    0  913M  0 part /boot

# lsblk
NAME    MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
sda       8:0    0   10G  0 disk
├─sda1    8:1    0    9G  0 part
├─sda14   8:14   0    4M  0 part
├─sda15   8:15   0  106M  0 part /boot/efi
└─sda16 259:1    0  913M  0 part
sdb       8:16   0   10G  0 disk
├─sdb1    8:17   0    9G  0 part /
├─sdb14   8:30   0    4M  0 part
├─sdb15   8:31   0  106M  0 part
└─sdb16 259:0    0  913M  0 part /boot
```
Still investigating the cause.

MggMuggins:
A little more info on the bootorder bug; this is worse than I thought. When two VMs are launched from the same cloud image, the UUIDs for the filesystems and partitions are all the same:

```
root@vm1:~# efibootmgr -v
BootCurrent: 0001
Timeout: 0 seconds
BootOrder: 0007,0001,0003,0004,0005,0006,0000,0002
Boot0000* UiApp	FvVol(7cb8bdc9-f8eb-4f34-aaea-3ee4af6516a1)/FvFile(462caa21-7614-4503-836e-8ab6f4662331)
Boot0001* UEFI QEMU QEMU HARDDISK 	PciRoot(0x0)/Pci(0x1,0x1)/Pci(0x0,0x0)/SCSI(0,1)N.....YM....R,Y.
Boot0002* EFI Internal Shell	FvVol(7cb8bdc9-f8eb-4f34-aaea-3ee4af6516a1)/FvFile(7c04a583-9e3e-4f1c-ad65-e05268d0b4d1)
Boot0003* UEFI PXEv4 (MAC:00163EF03AE4)	PciRoot(0x0)/Pci(0x1,0x4)/Pci(0x0,0x0)/MAC(00163ef03ae4,1)/IPv4(0.0.0.00.0.0.0,0,0)N.....YM....R,Y.
Boot0004* UEFI PXEv6 (MAC:00163EF03AE4)	PciRoot(0x0)/Pci(0x1,0x4)/Pci(0x0,0x0)/MAC(00163ef03ae4,1)/IPv6([::]:<->[::]:,0,0)N.....YM....R,Y.
Boot0005* UEFI HTTPv4 (MAC:00163EF03AE4)	PciRoot(0x0)/Pci(0x1,0x4)/Pci(0x0,0x0)/MAC(00163ef03ae4,1)/IPv4(0.0.0.00.0.0.0,0,0)/Uri()N.....YM....R,Y.
Boot0006* UEFI HTTPv6 (MAC:00163EF03AE4)	PciRoot(0x0)/Pci(0x1,0x4)/Pci(0x0,0x0)/MAC(00163ef03ae4,1)/IPv6([::]:<->[::]:,0,0)/Uri()N.....YM....R,Y.
Boot0007* ubuntu	HD(15,GPT,6714cd0e-2211-4b5f-8daa-341fcbae2865,0x2800,0x35000)/File(\EFI\ubuntu\shimx64.efi)
root@vm1:~# blkid
/dev/sda15: LABEL_FATBOOT="UEFI" LABEL="UEFI" UUID="F1D8-37B4" BLOCK_SIZE="512" TYPE="vfat" PARTUUID="6714cd0e-2211-4b5f-8daa-341fcbae2865"
/dev/sda1: LABEL="cloudimg-rootfs" UUID="fec1c9ae-0df3-419c-80dd-f3035049b845" BLOCK_SIZE="4096" TYPE="ext4" PARTUUID="40c8f90e-558a-4475-b818-ec6e9e5d02fb"
/dev/loop1: TYPE="squashfs"
/dev/loop2: TYPE="squashfs"
/dev/loop0: TYPE="squashfs"
/dev/sda14: PARTUUID="69646288-a3aa-469f-96fc-0782de904d84"
```
```
root@vm2:~# efibootmgr -v
BootCurrent: 0001
Timeout: 0 seconds
BootOrder: 0007,0001,0003,0004,0005,0006,0000,0002
Boot0000* UiApp	FvVol(7cb8bdc9-f8eb-4f34-aaea-3ee4af6516a1)/FvFile(462caa21-7614-4503-836e-8ab6f4662331)
Boot0001* UEFI QEMU QEMU HARDDISK 	PciRoot(0x0)/Pci(0x1,0x1)/Pci(0x0,0x0)/SCSI(0,1)N.....YM....R,Y.
Boot0002* EFI Internal Shell	FvVol(7cb8bdc9-f8eb-4f34-aaea-3ee4af6516a1)/FvFile(7c04a583-9e3e-4f1c-ad65-e05268d0b4d1)
Boot0003* UEFI PXEv4 (MAC:00163EA0318E)	PciRoot(0x0)/Pci(0x1,0x4)/Pci(0x0,0x0)/MAC(00163ea0318e,1)/IPv4(0.0.0.00.0.0.0,0,0)N.....YM....R,Y.
Boot0004* UEFI PXEv6 (MAC:00163EA0318E)	PciRoot(0x0)/Pci(0x1,0x4)/Pci(0x0,0x0)/MAC(00163ea0318e,1)/IPv6([::]:<->[::]:,0,0)N.....YM....R,Y.
Boot0005* UEFI HTTPv4 (MAC:00163EA0318E)	PciRoot(0x0)/Pci(0x1,0x4)/Pci(0x0,0x0)/MAC(00163ea0318e,1)/IPv4(0.0.0.00.0.0.0,0,0)/Uri()N.....YM....R,Y.
Boot0006* UEFI HTTPv6 (MAC:00163EA0318E)	PciRoot(0x0)/Pci(0x1,0x4)/Pci(0x0,0x0)/MAC(00163ea0318e,1)/IPv6([::]:<->[::]:,0,0)/Uri()N.....YM....R,Y.
Boot0007* ubuntu	HD(15,GPT,6714cd0e-2211-4b5f-8daa-341fcbae2865,0x2800,0x35000)/File(\EFI\ubuntu\shimx64.efi)
root@vm2:~# blkid
/dev/sda15: LABEL_FATBOOT="UEFI" LABEL="UEFI" UUID="F1D8-37B4" BLOCK_SIZE="512" TYPE="vfat" PARTUUID="6714cd0e-2211-4b5f-8daa-341fcbae2865"
/dev/sda1: LABEL="cloudimg-rootfs" UUID="fec1c9ae-0df3-419c-80dd-f3035049b845" BLOCK_SIZE="4096" TYPE="ext4" PARTUUID="40c8f90e-558a-4475-b818-ec6e9e5d02fb"
/dev/loop1: TYPE="squashfs"
/dev/loop2: TYPE="squashfs"
/dev/loop0: TYPE="squashfs"
/dev/sda14: PARTUUID="69646288-a3aa-469f-96fc-0782de904d84"
```

The UUID for the root filesystem is passed via kernel cmdline in Jammy:
```
# cat /proc/cmdline
BOOT_IMAGE=/boot/vmlinuz-5.15.0-1067-kvm root=PARTUUID=40c8f90e-558a-4475-b818-ec6e9e5d02fb ro console=tty1 console=ttyS0
```

But `ubuntu:noble` images use the partition label instead of the UUID:
```
# cat /proc/cmdline
BOOT_IMAGE=/vmlinuz-6.8.0-49-generic root=LABEL=cloudimg-rootfs ro console=tty1 console=ttyS0
```

TL;DR, VM root volume attachment is only safe between machines deployed using different cloud images, using Jammy or older as the recovery machine. Working on figuring out how(/if) the public clouds get around this.

MggMuggins:
Done some more reading:
- Openstack VMs behave exactly the same as LXD; two VMs created from the same image have the same partition/fs UUIDs. OpenStack doesn't support removing boot disks from instances at all; there is a [spec](https://blueprints.launchpad.net/nova/+spec/detach-boot-volume) and part of an implementation that was abandoned in 2018
- I haven't looked super hard at GCP but the docs are pretty thin on attaching boot disks to other instances; [this](https://cloud.google.com/compute/docs/disks/detach-reattach-boot-disk#attach_disk) implies that an instance can only have one boot disk at a time. Unclear if "boot disk" is an immutable feature of disk devices or something else; I have little/no experience with GCP.
- The workaround documented in the [AWS docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-booting-from-wrong-volume.html) is likely to work for the root device. It looks to me like the EFI boot entries in the instance's NVRAM use the UUID of `/boot/efi`, so changing labels will only force the kernel to select the correct root partition. I'm not sure that this is super viable, since one of the primary use cases for this feature is "the kernel is broken and the system won't boot" and the kernel executable is stored in `/boot/efi`.

I need to do some more reading on how we can control the EFI boot entries in LXD. More Friday.

MggMuggins:
In OpenStack, the procedure for getting access to another VM's root volume is to create an image from the VM, a volume from the image, and then mount the volume on another VM:
```
os server image create --name vm1_snap0 vm1
os volume create --image vm1_snap0 vm1_snap0_image --size 20GiB
os server add volume vm2 vm1_snap0_image --device /dev/vdb
```
This does suffer from the same duplicate UUID problem as described above. A VM can be created based on the modified volume.

tomponline:
please can you rebase

lxc/storage_volume.go
```diff
@@ -963,7 +966,10 @@ func (c *cmdStorageVolumeDetachProfile) run(cmd *cobra.Command, args []string) e
 	// Find the device
 	if devName == "" {
 		for n, d := range profile.Devices {
-			if d["type"] == "disk" && d["pool"] == resource.name && d["source"] == args[1] {
+			volType, volName := parseVolume("custom", args[1])
+			sourceType, sourceName := parseVolume("custom", d["source"])
```
hamistao, line 970: While I see the purpose of parsing the source to be safe, I don't think the source name is ever supposed to be the full volume name.
And if it somehow show as type/name it would be nice for it to be flagged right away so we can fix it, so in my view we can drop this second `parseVolume`.

hamistao: Oh I see, sow the type will be included because of `virtual-machine` volumes. You can disregard this.

doc/howto/storage_volumes.md
```diff
@@ -151,6 +151,34 @@ For example, to set a default volume size for a storage pool, use the following
 
     lxc storage set [<remote>:]<pool_name> volume.size <value>
 
+## Attach instance root volumes to other instances
+Virtual-machine root volumes can be attached as disk devices to other virtual machines.
+In order to prevent concurrent access, `security.protection.start` must be set on
+an instance before it's root volume can be attached to another virtual-machine.
```
minaelee, line 157:
```suggestion
an instance before its root volume can be attached to another virtual-machine.
```

MggMuggins:
Rebase is complete @tomponline

doc/explanation/storage.md
```diff
@@ -122,7 +122,7 @@ Storage volumes can be of the following types:
 : LXD automatically creates one of these storage volumes when you launch an instance.
   It is used as the root disk for the instance, and it is destroyed when the instance is deleted.
 
-  This storage volume is created in the storage pool that is specified in the profile used when launching the instance (or the default profile, if no profile is specified).
+  This storage volume is created in the storage pool that is specified when launching the instance (or in the default profile, if no pool or profile is specified).
```
minaelee, line 125:
```suggestion
  This storage volume is created in the storage pool that is specified when launching the instance. This pool can be specified explicitly or by specifying a profile. If no pool or profile is specified, LXD uses the storage pool of the default profile.
```

minaelee: I find this suggested version to be easier to understand because it explains that the storage pool can be specified explicitly or via the profile. The current version only implies (in "if no pool or profile is specified") that a profile can be used to specify the pool.

Without this edit, it's also difficult to parse what clause the parenthesized content modifies--that is, what does the part following "**or**" replace? I know after reading it a couple times that it's meant to replace "when launching the instance". However, at first glance, it's natural to parse it as "This storage volume is created in the default profile, if no pool or profile is specified" due to the symmetry of "**in the storage pool**" and "**in the default profile**", and that doesn't make sense.

MggMuggins: I totally agree RE clarity. Having looked at this again today I'm not sure that this is the right spot in the docs to nail down the semantics of which pool is selected for a root disk device at all; the [section above](https://documentation.ubuntu.com/lxd/en/latest/explanation/storage/#default-storage-pool) is probably more appropriate.

I'm wondering if the relationship between the root disk device and the `container`/`virtual-machine` volume type is a bit of a docs gap; figuring out the relationship between those two took me some time. Probably out of scope for this PR.

lxd/storage_volumes_utils.go
```diff
@@ -86,16 +87,6 @@ func storagePoolVolumeUpdateUsers(s *state.State, projectName string, oldPoolNam
...
-	// Handle instance volumes.
-	if vol.Type == cluster.StoragePoolVolumeTypeNameContainer || vol.Type == cluster.StoragePoolVolumeTypeNameVM {
-		volName, snapName, isSnap := api.GetParentAndSnapshotName(vol.Name)
-		if isSnap {
-			return []string{api.NewURL().Path(version.APIVersion, "instances", volName, "snapshots", snapName).Project(vol.Project).String()}, nil
-		}
-
-		return []string{api.NewURL().Path(version.APIVersion, "instances", volName).Project(vol.Project).String()}, nil
-	}
-
```
hamistao, line 98: For containers, doesn't it make sense to return early?

You are claiming it doesn't affect performance, but how can that be if we are running `VolumeUsedByInstanceDevices` and `VolumeUsedByProfileDevices` unnecessarily for containers?

I may have missed something here, if so sorry about this.

lxd/storage_volumes_utils.go
```diff
@@ -132,6 +123,21 @@ func storagePoolVolumeUsedByGet(s *state.State, requestProjectName string, vol *
...
+		if isSnap {
+			return []string{api.NewURL().Path(version.APIVersion, "instances", volName, "snapshots", snapName).Project(vol.Project).String()}, nil
+		}
+
+		// If vol is the instance's root volume and it is defined in a profile,
+		// it won't be added to the list by VolumeUsedByInstanceDevices.
+		instancePath := api.NewURL().Path(version.APIVersion, "instances", volName).Project(vol.Project).String()
+		if !slices.Contains(volumeUsedBy, instancePath) {
+			volumeUsedBy = append(volumeUsedBy, instancePath)
+		}
```
hamistao, line 138: If the volume is being used in a profile device as an additional drive, instances with this device won't be added to `volumeUsedBy`, right?

This is probably expected anyway, I am asking more for my own understanding.

lxd/storage/backend_lxd.go
```diff
@@ -2920,6 +2920,13 @@ func (b *lxdBackend) UpdateInstance(inst instance.Instance, newDesc string, newC
...
 
+		if shared.IsFalseOrEmpty(changedConfig["security.shared"]) && volDBType == cluster.StoragePoolVolumeTypeVM {
+			err = allowRemoveSecurityShared(b.state, inst.Project().Name, &curVol.StorageVolume)
+			if err != nil {
+				return err
+			}
+		}
+
 		// Load storage volume from database.
 		dbVol, err := VolumeDBGet(b, inst.Project().Name, inst.Name(), volType)
```
hamistao, line 2931: Not related to your changes, but are we getting the same volume from the database again?

On line 2895 we have `curVol, err := VolumeDBGet(b, inst.Project().Name, inst.Name(), volType)`

hamistao (commented):
I really like your idea of using `security.protection.start`, good one! 

Sorry for so many comments, hope they aren't too much of a pain to address.

https://github.com/canonical/lxd/pull/14593 has been merged so you can rebase.

Lastly, I was wondering if we should now check for volume type [here](https://github.com/hamistao/lxd/blob/cephfs_tests/lxd/storage/utils.go#L1149) and [here](https://github.com/hamistao/lxd/blob/cephfs_tests/lxd/storage/utils.go#L1075)

lxd/instance_delete.go
```diff
@@ -84,6 +88,45 @@ func instanceDelete(d *Daemon, r *http.Request) response.Response {
 		return response.BadRequest(fmt.Errorf("Instance is running"))
 	}
 
+	// Make sure that the instance's root volume is not attached to another instance
```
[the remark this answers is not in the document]
tomponline: > Would it be an option to delete the instance but keeping the volume?
No this should not be supported.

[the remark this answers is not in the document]
tomponline: > Plus, all these checks only need to be performed for VMs if I understand correctly, so doing them for container deletion is a waste.

True, for now at least, but making it a common function would still make sense to me.

lxd/device/disk.go
```diff
@@ -164,13 +164,23 @@ func (d *disk) checkBlockVolSharing(instanceType instancetype.Type, projectName
...
+			apiInst, err := inst.ToAPI()
+			if err != nil {
+				return err
+			}
+
+			apiInst.ExpandedConfig = instancetype.ExpandInstanceConfig(d.state.GlobalConfig.Dump(), apiInst.Config, inst.Profiles)
+
+			if shared.IsTrue(apiInst.ExpandedConfig["security.protection.start"]) {
+				return nil
+			}
```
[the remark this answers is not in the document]
tomponline: > I am wondering now if we should also allow custom block volumes to be attached to more than one instance if .`security.protection.start` is set for all but one of them. If so, it would be nice to keep those consistent I think.

This doesn't really make any sense to me as it has no use though.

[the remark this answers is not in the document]
tomponline: > If `security.protection.start` is set on the instance using the volume as an additional drive instead of the one using it as root, it should also be okay to attach the volume to it. Am I correct?

I dont follow this scenario, can you expand your idea?

lxc/storage_volume.go
```diff
@@ -257,7 +259,7 @@ func (c *cmdStorageVolumeAttach) run(cmd *cobra.Command, args []string) error {
 	device := map[string]string{
 		"type":   "disk",
 		"pool":   resource.name,
-		"source": volName,
+		"source": args[1],
```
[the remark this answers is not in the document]
MggMuggins: The server will still interpret `c1` and `custom/c1` as the same volume name.

lxd/storage/utils.go
```diff
@@ -971,6 +971,40 @@ func InstanceContentType(inst instance.Instance) drivers.ContentType {
 	return contentType
 }
 
+// volumeIsUsedByDevice
```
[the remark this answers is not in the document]
MggMuggins: No :sweat_smile:; I'll add a bit

lxd/storage/utils.go
```diff
@@ -971,6 +971,40 @@ func InstanceContentType(inst instance.Instance) drivers.ContentType {
...
+	}
+
+	if inst != nil && instancetype.IsRootDiskDevice(dev) {
+		rootVolumeType := cluster.StoragePoolVolumeTypeNameContainer
+		if inst.Type == instancetype.VM {
+			rootVolumeType = cluster.StoragePoolVolumeTypeNameVM
+		}
+
+		if inst.Name == vol.Name && rootVolumeType == vol.Type {
+			return true, nil
```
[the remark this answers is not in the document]
MggMuggins: Having 3+ different sets of constants floating around for the same idea is super cursed :sob:
I think I agree that this is more readable; more future-proof to use the functions too.

lxd/device/disk.go
```diff
@@ -164,13 +164,23 @@ func (d *disk) checkBlockVolSharing(instanceType instancetype.Type, projectName
...
-			return fmt.Errorf("Cannot add custom storage block volume to more than one instance if security.shared is false or unset")
+		// Don't count a VM volume's instance if security.protection.start is preventing that instance from starting
+		if volume.Type == cluster.StoragePoolVolumeTypeNameVM && volume.Project == inst.Project && volume.Name == inst.Name {
+			apiInst, err := inst.ToAPI()
+			if err != nil {
+				return err
+			}
+
+			apiInst.ExpandedConfig = instancetype.ExpandInstanceConfig(d.state.GlobalConfig.Dump(), apiInst.Config, inst.Profiles)
+
```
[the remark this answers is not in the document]
MggMuggins: What is missing from lines 150 and 167? I'll update the function's comment (150).

lxd/device/disk.go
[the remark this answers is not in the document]
MggMuggins: @tomponline gets all the credit for this one :wink: 

Making this consistent for all block volumes has some appeal to me; however, it would also mean that an instance can't unset `security.protection.start` until it's checked all of the instances in its project for every volume it currently has attached. [`VolumeUsedByInstanceDevices`](https://github.com/canonical/lxd/blob/main/lxd/storage/utils.go#L1096) would have to be called in a loop and it does a bunch of queries.

lxd/storage_volumes_utils.go
MggMuggins: Maybe I wasn't super clear. You're right, the way I've done this definitely affects performance for containers unnecessarily.

I'm still hoping to get this feature implemented for containers too, so this shouldn't be unnecessary for long.

lxd/storage_volumes_utils.go
MggMuggins: Yes, that's correct; we don't expand devices from profiles here. We want to report everywhere that references exist, not necessarily where the resources are actually attached.

This spot is a special case for root disk devices are defined on a profile. I'll expand the comment.

lxd/storage/utils.go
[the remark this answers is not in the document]
hamistao: > more future-proof to use the functions too.

Yeah this is precisely my reasoning, I agree that yours is perhaps more readable but future-proofing probably comes first.

lxc/storage_volume.go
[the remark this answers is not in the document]
hamistao: oh cool then, closing this

lxd/device/disk.go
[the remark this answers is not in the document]
hamistao: Maybe just something like `// We can share with instances that cannot be started without running the risk of data corruption` is enough, to clarify what `security.protection.start` does and why it is being checked for here.

lxd/storage_volumes_utils.go
hamistao: Cool, makes sense to me.
Also good idea on expanding the comment.

lxd/storage_volumes_utils.go
hamistao: Hmm I see. Probably best to keep this efficient so we don't run the risk of making a release/round of backports that only include this commit and not the future implementation for containers. What do you think?

lxd/instance_delete.go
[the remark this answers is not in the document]
hamistao: > True, for now at least, but making it a common function would still make sense to me.

Indeed, this reminds me of our discussion earlier on uplink IP quota validation for different network drivers.
And now I know that we intend to support attaching container volumes as well in the near future, so this makes even more sense.

lxd/device/disk.go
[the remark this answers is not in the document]
hamistao: > This doesn't really make any sense to me as it has no use though.

> it would also mean that an instance can't unset security.protection.start until it's checked all of the instances in its project for every volume it currently has attached.

Yeah this isn't my sharpest thought, sorry just throwing ideas around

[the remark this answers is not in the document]
hamistao: > I dont follow this scenario, can you expand your idea?

Say we have an instance `vm2` using `vm1` volume as a source for an additional(not-root) disk drive and `vm1` doesn't have `security.shared`. Both `vm1` and `vm2` have `security.protection.start` enabled. Should we allow attaching `vm1` to another instance?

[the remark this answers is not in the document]
hamistao: > [VolumeUsedByInstanceDevices](https://github.com/canonical/lxd/blob/main/lxd/storage/utils.go#L1096) would have to be called in a loop and it does a bunch of queries.

Though this would also require doing this heavy work when unsetting `security.protection.start`

tomponline:
@hamistao @MggMuggins please can you go through and resolve any chat threads that you deem resolved so i can get an idea of whether this is ready for a final review?

Also needs a rebase please

hamistao:
I will defer closing the threads to @MggMuggins to do when he implements the changes discussed in them.
The ones that don't require any change have been closed.

lxd/instance_delete.go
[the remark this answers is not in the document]
MggMuggins: I totally agree; this should be in a common function.

See https://github.com/canonical/lxd/pull/13821 for a little bit of context on why a common delete function is less trivial than I'd like. We have the following structure:
- `package lxd/instance`
  - `Instance interface`
  - `package drivers`
    - `common struct`
    - `lxc struct` (implements `Instance`)
    - `qemu struct` (implements `Instance`)

It's fairly common (this case included) to want to use methods of the `Instance` interface from the drivers (e.g. `LoadByInstance` or `Delete`). However, `common` does not implement `Instance`, so methods receiving a `common` aren't able to use the interface's methods.

This leads to a _ton_ of code duplication in the instance drivers (see [lxc.Delete](https://github.com/canonical/lxd/blob/d537da48f329eaefa3c77c9e1479d893fbe12a3d/lxd/instance/drivers/driver_lxc.go#L3711) and [qemu.Delete](https://github.com/canonical/lxd/blob/main/lxd/instance/drivers/driver_qemu.go#L6011); I suspect that a substantial chunk of both `lxc` and `qemu`'s methods could be refactored if `common` implemented `Instance`. Alternatively, the package could be refactored into something more like `lxd/storage`, where `lxd/storage.Pool` and `lxd/storage/drivers.Driver` are distinct interfaces that provide a lot of similar capabilities but aren't interchangeable.

Either way, solving this problem the "right" way is out of scope for this PR. A few options for what I should do here:
- [ ] Leave this in the `DELETE` handler for now and open a follow-up PR to move it to a common delete function, likely with some more widespread refactoring
- [ ] Push this into the qemu driver and make container attachment harder/increase code duplication later
- [x] Create a stopgap common function and call it only from the qemu driver (for now)

lxd/storage/utils.go
```diff
@@ -971,6 +971,40 @@ func InstanceContentType(inst instance.Instance) drivers.ContentType {
...
+		}
+	}
+
+	_, _, _, volumeName, err := DiskVolumeSourceParse(dev["source"])
```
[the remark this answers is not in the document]
MggMuggins: Yes; I described the change in behavior rather than the refactoring

lxd/storage_volumes_utils.go
MggMuggins: I re-introduced the check for containers, I'll remove it when I implement container root attachments.

doc/api-extensions.md
```diff
@@ -2535,3 +2535,9 @@ Adds a new {config:option}`device-unix-hotplug-device-conf:ownership.inherit` co
 ## `unix_device_hotplug_subsystem_device_option`
 
 Adds a new {config:option}`device-unix-hotplug-device-conf:subsystem` configuration option for `unix-hotplug` devices. This adds support for detecting `unix-hotplug` devices by subsystem, and can be used in conjunction with {config:option}`device-unix-hotplug-device-conf:productid` and {config:option}`device-unix-hotplug-device-conf:vendorid`.
+
+## `instance_root_volume_attachment`
+
+Adds support for instance root volumes to be attached to other instances as disk
```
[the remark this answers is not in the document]
MggMuggins: I was hoping to avoid needing two extensions to refer to the same change; I'll defer to @tomponline here

MggMuggins:
@tomponline I've addressed all of @hamistao's feedback except the bits I'd like you to weigh in on.

doc/api-extensions.md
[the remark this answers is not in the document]
hamistao: Fine by me then

hamistao:
@MggMuggins Hey! Just some quick questions that are bugging me:
 - Should [this](https://github.com/hamistao/lxd/blob/main/lxd%2Fstorage%2Futils.go#L1075) and [this](https://github.com/hamistao/lxd/blob/main/lxd%2Fstorage%2Futils.go#L1149) also check for volume type now?
 - Should [this](https://github.com/hamistao/lxd/blob/main/lxd%2Fstorage_volumes_utils.go#L33) be updated to keep the "source" format consistent?

doc/api-extensions.md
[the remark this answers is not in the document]
tomponline: it depends if the container root disk attachment lands before the next release after this Pr is merged.

MggMuggins:
Thanks for the check-in:
- The volume type is [now checked](https://github.com/canonical/lxd/pull/14532/commits/c2a9fd6e8d7f1e47a48b3ca802fafbb4b9a71c0a#diff-33511e3a6e5abae8b54480aa23c933de0ec50584842b582286029955063c389dR995)
- Likely yes; I'll take a look this afternoon

MggMuggins:
I've put up #14681 with a fix for your second bullet, as it's not strictly related to this PR. But it brings up a good point: LXD should update any uses of an instance's root volume when the instance is renamed. Still working on the fix.

tomponline:
> But it brings up a good point: LXD should update any uses of an instance's root volume when the instance is renamed. Still working on the fix.

From what I've seen in LXD, it prefers blocking an action that would cause an entity to reference a non-existent entity, rather than updating the reference.


```
lxc network rename lxdbr0 foo
Error: Network is currently in use
```

So for now I think we should block the rename of an instance if its in use by another instance.

And later think about a more general approach for renaming and updating references (as an option).

tomponline:
@MggMuggins please can you rebase this ready for our 1:1 on Friday so we can go through together and hopefully merge. Ta

lxd/storage/utils.go
```diff
@@ -970,6 +970,45 @@ func InstanceContentType(inst instance.Instance) drivers.ContentType {
 	return contentType
 }
 
+// volumeIsUsedByDevice; true when vol is referred to by dev
+// inst is the instance dev belongs to, or nil if dev is part of a profile.
+func volumeIsUsedByDevice(vol *api.StorageVolume, inst *db.InstanceArgs, dev map[string]string) (bool, error) {
```
tomponline, line 975: Shall we avoid passing vol as a pointer to avoid the implication/possibility for this function being able to modify it?

lxd/storage/utils.go
```diff
@@ -1107,9 +1138,6 @@ func VolumeUsedByInstanceDevices(s *state.State, poolName string, projectName st
 			}
 
 			instStorageProject := project.StorageVolumeProjectFromRecord(&p, volumeType)
-			if err != nil {
```
tomponline, line 1112: if this was a redundant error check can you remove it as a separate commit so we can backport this change independent of the feature?

lxd/instance/drivers/driver_qemu.go
```diff
@@ -5598,6 +5598,47 @@ func (d *qemu) Update(args db.InstanceArgs, userRequested bool) error {
 		if newErr != nil {
 			return fmt.Errorf("Invalid root disk device: %w", newErr)
 		}
+
+		// If security.protection.start is being removed, we need to make sure that
+		// our root disk device is not attached to another instance.
+		if shared.IsTrue(oldExpandedConfig["security.protection.start"]) && shared.IsFalseOrEmpty(d.expandedConfig["security.protection.start"]) {
```
tomponline, line 5609: Lets move this into a separate function to avoid increasing the Update functions' cyclomatic complexity any further.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -5598,6 +5598,47 @@ func (d *qemu) Update(args db.InstanceArgs, userRequested bool) error {
...
+			if err != nil {
+				return err
+			}
+
+			volumeType := dbCluster.StoragePoolVolumeTypeVM
+			volumeName := d.common.name
+			volumeProject := project.StorageVolumeProjectFromRecord(&d.common.project, volumeType)
+
+			var dbVolume *db.StorageVolume
+			err = d.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
```
tomponline, line 5622: where you dont have a request context, you can use d.state.ShutdownCtx

lxd/storage/backend_lxd.go
```diff
@@ -5951,6 +5951,32 @@ func (b *lxdBackend) detectChangedConfig(curConfig, newConfig map[string]string)
...
+		return fmt.Errorf("Cannot disable security.shared on custom storage block volume as it is attached to profile(s)")
+	})
+	if err != nil {
+		return err
+	}
+
+	var usedByInstanceDevices []string
+
+	err = VolumeUsedByInstanceDevices(s, volume.Pool, projectName, volume, true, func(inst db.InstanceArgs, project api.Project, usedByDevices []string) error {
+		usedByInstanceDevices = append(usedByInstanceDevices, inst.Name)
```
tomponline, line 5980: this caught my eye. Why do we need this usedByInstanceDevices slice, when the very act of this function being called means it is in use and thus we can just return an error?

lxd/storage/backend_lxd.go
```diff
@@ -2920,6 +2920,13 @@ func (b *lxdBackend) UpdateInstance(inst instance.Instance, newDesc string, newC
 			return fmt.Errorf(`Instance volume "volatile.uuid" property cannot be changed`)
 		}
 
+		if shared.IsFalseOrEmpty(changedConfig["security.shared"]) && volDBType == cluster.StoragePoolVolumeTypeVM {
```
tomponline, line 2923: what do you think about adding a function that indicates if the volDBType is an instance volume (container or VM), and then preemptively use that function in places like this to apply to both VM and containers, even though we dont yet allow adding container's root disk to other containers?

lxd/storage/backend_lxd.go
tomponline, line 1: nice catch

lxd/instance/drivers/driver_common.go
```diff
@@ -1697,6 +1697,55 @@ func (d *common) deleteSnapshots(deleteFunc func(snapInst instance.Instance) err
...
+		return err
+	}
+
+	rootVolumeDBType, err := storagePools.VolumeTypeToDBType(rootVolumeType)
+	if err != nil {
+		return err
+	}
+
+	var rootVolume *db.StorageVolume
+	err = d.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
```
tomponline, line 1720: you can use ShutdownCtx here i think

lxd/storage/utils.go
MggMuggins: :+1: #14764

lxd/instance/drivers/driver_qemu.go
MggMuggins: `changedConfig` would also pick up a `lxc config set v1 security.protection.start=true`; we only care about when the option is being removed/set `false`. This way we can trim a few queries off of setting the value to `true`.

lxd/storage/backend_lxd.go
MggMuggins: Agreed, the slice is completely unnecessary. But, removing `security.shared` is perfectly valid if there are 0 or 1 instances using the volume.

lxd/storage/backend_lxd.go
MggMuggins: There's only one other spot it's used this way, but it's a reasonable idea. I went ahead and added it.

lxd/storage_volumes_utils.go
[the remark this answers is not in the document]
MggMuggins: Sorry, I forgot to update the commit message. There's certainly an argument that this isn't the clearest code...

lxd/storage/backend_lxd.go
MggMuggins: I'll tuck that away in a branch and polish it off after this is merged.

lxd/instance/drivers/driver_qemu.go
tomponline: understood thanks

lxd/storage/utils.go
```diff
@@ -1054,20 +1093,12 @@ func VolumeUsedByProfileDevices(s *state.State, poolName string, projectName str
...
-			if dev["type"] != cluster.TypeDisk.String() {
-				continue
-			}
-
-			if dev["pool"] != poolName {
-				continue
-			}
-
-			_, _, _, volumeName, err := DiskVolumeSourceParse(dev["source"])
+			usesVol, err := volumeIsUsedByDevice(*vol, nil, dev)
```
MggMuggins, line 1096: This probably should correctly pick up `virtual-machine`/`container` volumes when attached as the root device of a profile.

tomponline:
Please can you rebase again

MggMuggins:
Done.

lxd/storage/utils.go
MggMuggins: @tomponline I will open a new PR for this as it will require another DB query... we need a list of the instances using each profile. I do think that from a correctness standpoint `VolumeUsedByProfileDevices` should call its callback when the volume is associated with a Profile's root disk device.

tomponline (approved):

lxd/storage/utils.go
tomponline: > I will open a new PR for this as it will require another DB query... we need a list of the instances using each profile

Why do you think this is needed? Isn' the job of `VolumeUsedByProfileDevices` to match disk devices in profiles to an instance device, not other instances? That is what `VolumeUsedByInstanceDevices` is for right?

> I do think that from a correctness standpoint VolumeUsedByProfileDevices should call its callback when the volume is associated with a Profile's root disk device.

Yeah that makes sense, as the profile is using that disk.

PR #14533, opened 2024-11-26
lxd: Simplify `allowAuthenticated`

hamistao:


markylaing (approved):
The consequence of this change is that if `allowAuthenticated` is called before `request.CtxTrusted` is set the API response will be `401 Forbidden` rather than a `500 Internal Server Error` with message `Failed to get expected value "trusted" from context`.

However, `request.CtxTrusted` is always set in `(*Daemon).createCmd`, and `auth.IsTrusted` is already used in many situations without checking if the value is actually set, so I think this is fine.

Issue #14534, opened 2024-11-26
Check whether `security.shared` makes sense for `cephfs` and if not remove


tomponline:
The security.shared volume setting is only relevant for custom block volumes (you can see there is a "condition: custom block volume " section in the drop down for that setting.

https://documentation.ubuntu.com/lxd/en/latest/reference/storage_cephfs/#storage-cephfs-volume-conf:security.shared

However as cephfs storage driver doesn't support block volumes this doesn't make sense to me.

https://documentation.ubuntu.com/lxd/en/latest/reference/storage_cephfs/#storage-cephfs-volume-conf:security.shared

Discussion: https://chat.canonical.com/canonical/pl/6wry93xjcfn78g8tbitstp4fmw

tomponline:
Looks like it was this commit

https://github.com/canonical/lxd/pull/13183/commits/362734e5ae886b43310b9a42fa01cec4345cbe73

@hamistao was this an oversight or added for cephfs particularly?

Thanks

hamistao:
This was one of my very first contributions to LXD, it was definitely an oversight. Will fix in a minute.

PR #14536, opened 2024-11-26
lxd/storage: Remove `security.shared` from cephfs keys

hamistao:
cephfs does not support block volumes, thus `security.shared` does not apply to it.

Closes https://github.com/canonical/lxd/issues/14534

### DIFF
--- a/doc/metadata.txt
+++ b/doc/metadata.txt
@@ -5090,15 +5090,6 @@ when creating a missing OSD pool.
 
 <!-- config group storage-cephfs-pool-conf end -->
 <!-- config group storage-cephfs-volume-conf start -->
-```{config:option} security.shared storage-cephfs-volume-conf
-:condition: "custom block volume"
-:defaultdesc: "same as `volume.security.shared` or `false`"
-:shortdesc: "Enable volume sharing"
-:type: "bool"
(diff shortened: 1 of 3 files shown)


hamistao:
> Please could you add a check for this key not being allowed on cephfs volumes in test_storage_driver_cephfs around the # Test invalid key combinations for auto-creation of cephfs entities. section?

Isn't this better framed as a test for setting `security.shared` for non-block volumes in general, instead of something specific for `cephfs`?

tomponline:
> Isn't this better framed as a test for setting `security.shared` for non-block volumes in general, instead of something specific for `cephfs`? Or maybe both?

I'm assuming we already have this (but good idea to check), but was more thinking about whether that setting can be set on cephfs pools or not via `volume.security.shared` setting (you shouldn't be able to set this).

tomponline (approved):
merging now as 6.2 is imminent, but tests to follow :)

hamistao:
> I'm assuming we already have this (but good idea to check), but was more thinking about whether that setting can be set on cephfs pools or not via volume.security.shared setting (you shouldn't be able to set this).

Got it, I will probably also have to add some validation for `volume.security.shared`.

Issue #13698, opened 2024-07-03
`lxc config show <instance>` for an instance on an offline cluster member fails


tomponline:
`lxc config show <instance>` for an instance on an offline cluster member fails.

Can this be made to work? As the data for the response should be purely coming from the database.

PR #14539, opened 2024-11-26
lxd: Get config for instance on offline cluster member

hamistao:
If not using recursion on `GET /1.0/instances/:instanceName`, almost all the instance information used in the response comes from the database. So when forwarding the request is not possible, fallback to reading the instance info from the node being queried. The only change is the response is the instance state, that in this case would show as Error since we can't determine the state of a remote instance.
 
 closes https://github.com/canonical/lxd/issues/13698

### DIFF
--- a/lxd/instance/drivers/driver_lxc.go
+++ b/lxd/instance/drivers/driver_lxc.go
@@ -8141,6 +8141,11 @@ func (d *lxc) NextIdmap() (*idmap.IdmapSet, error) {
 
 // statusCode returns instance status code.
 func (d *lxc) statusCode() api.StatusCode {
+	// If instance is running on a remote cluster member, we cannot determine instance state.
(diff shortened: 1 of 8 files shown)


tomponline:
typo in pr title

lxd/instance/drivers/driver_lxc.go
```diff
@@ -8141,6 +8141,11 @@ func (d *lxc) NextIdmap() (*idmap.IdmapSet, error) {
 
 // statusCode returns instance status code.
 func (d *lxc) statusCode() api.StatusCode {
+	// If instance is running on a remote cluster member, we cannot determine instance state.
```
tomponline, line 8144: in what situation are we calling statusCode on non-local containers?

tomponline (approved):
Thanks

Issue #14537, opened 2024-11-26
Missing instance config suggestions from tab completion


simondeziel:
```
$ lxc init --empty c1
Creating c1
$ lxc config set c1 linux<tab>   # completes to linux.kernel_modules
$ lxc config set c1 linux.kernel_modules
```

But it skips the `linux.sysctl.` key prefix. Same for `user.` and `environment.`. Please note that for those key prefix, it's only possible to complete up to the trailing `.` as afterward it's all unknown/user provided.

kadinsayani:
Thanks for putting this issue up! Would you prefer if `linux.sysctl.*` was completed, or `linux.sysctl.`?

simondeziel:
> Would you prefer if `linux.sysctl.*` was completed, or `linux.sysctl.`?

Completing up to the 2nd `.` makes sense to me.

Also while playing around it, I noticed that `lxc config unset <inst> <tab>` could be made smarter in proposing to complete only the config keys that are already `set` and thus `unset`'able. Happy to fill this one as a separated issue if you want.

simondeziel:
> Should we also throw in a no space directive for `lxc config get/set/unset`?

Definitely as those don't accept dir/file arguments anyway

PR #14542, opened 2024-11-26
CLI: Improvements to `lxc config` and `lxc profile` shell completions

kadinsayani:
This PR includes updates to `lxc/completion.go`, improving the shell completion functionality for instance configuration keys. 

Summary of changes:
- Refactorings that improve performance and correctness;
- Updates `cmpInstanceKeys` and `cmpInstanceAllKeys` functions to use the metadata API (resolves https://github.com/canonical/lxd/issues/14537);
- Adds `cmpInstanceSetKeys` function for improved `lxc config unset` completions.

### DIFF
--- a/lxc/completion.go
+++ b/lxc/completion.go
@@ -512,9 +512,8 @@ func (g *cmdGlobal) cmpInstances(toComplete string) ([]string, cobra.ShellCompDi
 	}
 
 	if !strings.Contains(toComplete, ":") {
-		remotes, directives := g.cmpRemotes(false)
+		remotes, _ := g.cmpRemotes(false)
 		results = append(results, remotes...)
-		cmpDirectives |= directives
 	}
 
 	return results, cmpDirectives
--- a/lxc/completion.go
+++ b/lxc/completion.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/canonical/lxd/lxd/instance/instancetype"
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
 )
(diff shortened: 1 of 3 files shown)


lxc/completion.go
```diff
@@ -250,44 +248,144 @@ func (g *cmdGlobal) cmpInstanceKeys(instanceName string) ([]string, cobra.ShellC
...
+		keyCount += len(field.Keys)
+	}
+
+	configKeys := make([]string, 0, keyCount)
+
+	for _, field := range instanceConfig {
+		for _, key := range field.Keys {
+			for configKey, configKeyField := range key {
+				if strings.HasSuffix(configKey, "*") {
+					configKey = strings.Trim(configKey, "*")
```
simondeziel, line 273: `strings.TrimSuffix()` is a faster implementation.

lxc/completion.go
```diff
@@ -250,44 +245,144 @@ func (g *cmdGlobal) cmpInstanceKeys(instanceName string) ([]string, cobra.ShellC
...
+	client := resource.server
+
+	instance, _, err := client.GetInstance(instanceName)
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveError
+	}
+
+	// Pre-allocate configKeys slice capacity.
+	keyCount := 0
+	keyCount += len(instance.Config)
```
simondeziel, line 366:
```suggestion
	keyCount := len(instance.Config)
```

lxc/completion.go
```diff
@@ -250,44 +245,144 @@ func (g *cmdGlobal) cmpInstanceKeys(instanceName string) ([]string, cobra.ShellC
...
+		keyCount += len(field.Keys)
+	}
+
+	configKeys := make([]string, 0, keyCount)
+
+	for _, field := range instanceConfig {
+		for _, key := range field.Keys {
+			for configKey := range key {
+				if strings.HasSuffix(configKey, "*") {
+					configKey = strings.Trim(configKey, "*")
```
simondeziel, line 329: `strings.TrimSuffix()` is a faster implementation than .Trim()`.

`configKey = configKey[:len(configKey)-1]` would also work but might be less readable?

lxc/completion.go
```diff
@@ -250,44 +245,144 @@ func (g *cmdGlobal) cmpInstanceKeys(instanceName string) ([]string, cobra.ShellC
...
+				if instanceType == string(api.InstanceTypeContainer) && configKeyField.Condition == "container" {
+					configKeys = append(configKeys, configKey)
+				} else if instanceType == string(api.InstanceTypeVM) && configKeyField.Condition == "virtual machine" {
+					configKeys = append(configKeys, configKey)
+				}
+
+				// InstanceTypeAny config keys.
+				if configKeyField.Condition == "" {
+					configKeys = append(configKeys, configKey)
+				}
```
simondeziel, line 285: Doing this `if` check before the instance type would be a tad faster.

lxc/completion.go
kadinsayani: I say let's go with the slice operation for performance :)

kadinsayani: Since it's just `len(configKey)-1`, I think readability is intact with the slicing syntax.

kadinsayani: Hmm my language server is suggesting `strings.TrimSuffix()` over the conditional in conjunction with slicing syntax. Since `strings.TrimSuffix()` returns the string unchanged and has early exits, the overhead may be negligible if we drop the conditional.

lxc/completion.go
kadinsayani: I missed this 😆

lxc/completion.go
kadinsayani: Great suggestion, thanks!

lxc/completion.go
kadinsayani: This came up while copying code across completion functions, which got me thinking that there's a lot of room for refactoring and de-duplicating logic in `lxc/completions.go`. As `lxc/completions.go` has grown, so has duplication. I don't see any immediate benefit to making such improvements, but I'll open an issue and self-assign.

tomponline:
@kadinsayani @simondeziel has any consideration been thought about wrt to caching the metdata api response for a particular remote so it doesnt have to keep re-requesting it (perhaps tied to the version number returned by the remote's /1.0 api)?

kadinsayani:
> @kadinsayani @simondeziel has any consideration been thought about wrt to caching the metdata api response for a particular remote so it doesnt have to keep re-requesting it (perhaps tied to the version number returned by the remote's /1.0 api)? 

I'll include this with https://github.com/canonical/lxd/issues/14543 :)

PR #14551, opened 2024-11-28
doc: reword explanation articles' titles

minaelee:
- Remove repetitive "About" word for each Explanation article title in side navigation & Explanation index page
- Rephrase titles for increased specificity as well as to avoid duplicating titles from how-to/reference articles
- Update links to a renamed file and add redirect

### DIFF
--- a/doc/custom_conf.py
+++ b/doc/custom_conf.py
@@ -145,7 +145,8 @@ slug = "lxd"
 redirects = {
     'howto/instances_snapshots/index': '../instances_backup/',
     'reference/network_external/index': '../networks/',
-    'explanation/containers_and_vms/index': '../instances/'
+    'explanation/containers_and_vms/index': '../instances/',
+    'explanation/clustering/index': '../clusters/'
 }
 
 ############################################################
--- a/doc/database.md
+++ b/doc/database.md
@@ -3,7 +3,7 @@ relatedlinks: "[Canonical&#32;Dqlite](https://dqlite.io/), https://github.com/ca
 ---
 
 (database)=
-# About the LXD database
+# The LXD Dqlite database
 
 LXD uses a distributed database to store the server configuration and state, which allows for quicker queries than if the configuration was stored inside each instance's directory (as it is done by LXC, for example).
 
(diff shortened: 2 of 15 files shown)


tomponline (changes requested):
Thanks @minaelee please can you break these changes down into a few separate commits so that they are more likely to be backportable into our LTS series without your intervention.

I would suggest a commit for each section, including any related redirection changes.

doc/authentication.md
```diff
@@ -4,7 +4,7 @@ relatedlinks: https://www.youtube.com/watch?v=6O0q3rSWr8A
 ---
 
 (authentication)=
-# Remote API authentication
+# API authentication
```
tomponline, line 7: The "Remote" part here was relevant because it only applies when accessing the API remotely (i.e over the network), as opposed to accessing the API over the local unix socket. What was the reason for removing it?

tomponline:
@minaelee  can we close this for now until you get chance to look at this again?

doc/authentication.md
minaelee: Currently the side navigation lists "About authentication" and "About authorization" without any mention of either "Remote" or "API", overriding default use of the page title for the navlink texts. This PR deletes those overrides for these pages and defaults to the page titles; the intent in updating the page titles was to strike a middle ground with a text that's more descriptive than "About authentication" but more succinct than "Remote API authentication". 

Thanks for your explanation about needing to leave "Remote" in the page titles. I will revert those docs. Do you think it's as important to use "Remote" in the sidebar nav, or is "API authentication" sufficient there?

tomponline: > The "Remote" part here was relevant because it only applies when accessing the API remotely (i.e over the network), as opposed to accessing the API over the local unix socket. What was the reason for removing it?

I've not got strong opinions about that. If you would prefer for them to be consistent then it makes sense for remote to be in the sidebar nav too, but then it might not be succinct enough. This maybe why the compromise was made originally.

Maybe we could go for "Remote authentication" instead?

minaelee: Let's go ahead and make it consistent. I don't think it will be too long, and I think it would be helpful for users to see the word API there.

PR #14557, opened 2024-11-29
OpenFGA: Add request cache to the OpenFGA datastore

gabrielmougard:
This PR decrease the latency of API calls that are going through the OpenFGA fine-grained authorizer. Thanks to a per-request cache mechanism, we avoid calling the database when the cache key is present.

This work has been benchmarked and has been used in the following PR: https://github.com/canonical/lxd/pull/14476

[openfga_benchmark.pdf](https://github.com/user-attachments/files/17960500/openfga_benchmark.pdf)

The benchmarking script can be found at: https://paste.ubuntu.com/p/WCwsk6gSSK/

### DIFF
--- a/lxd/auth/drivers/openfga.go
+++ b/lxd/auth/drivers/openfga.go
@@ -84,10 +84,6 @@ func (e *embeddedOpenFGA) load(ctx context.Context, identityCache *identity.Cach
 		server.WithDatastore(opts.openfgaDatastore),
 		// Use our logger.
 		server.WithLogger(openfgaLogger{l: e.logger}),
-		// Set the max concurrency to 1 for both read and check requests.
-		// Our driver cannot perform concurrent reads.
-		server.WithMaxConcurrentReadsForListObjects(1),
-		server.WithMaxConcurrentReadsForCheck(1),
 		// Enable context propagation to the datastore so that the OpenFGA cache
 		// created for each request can be accessed from our OpenFGA datastore implementation.
 		server.WithContextPropagationToDatastore(true),
--- a/lxd/auth/drivers/openfga.go
+++ b/lxd/auth/drivers/openfga.go
@@ -88,6 +88,9 @@ func (e *embeddedOpenFGA) load(ctx context.Context, identityCache *identity.Cach
 		// Our driver cannot perform concurrent reads.
 		server.WithMaxConcurrentReadsForListObjects(1),
 		server.WithMaxConcurrentReadsForCheck(1),
+		// Enable context propagation to the datastore so that the OpenFGA cache
+		// created for each request can be accessed from our OpenFGA datastore implementation.
+		server.WithContextPropagationToDatastore(true),
 	}
 
 	e.server, err = server.NewServerWithOpts(openfgaServerOptions...)
--- a/lxd/daemon.go
+++ b/lxd/daemon.go
@@ -754,6 +754,9 @@ func (d *Daemon) createCmd(restAPI *mux.Router, version string, c APIEndpoint) {
 			return
 		}
 
+		// Set OpenFGA cache in request context.
+		request.SetCtxValue(r, request.CtxOpenFGARequestCache, &openfga.RequestCache{})
+
 		// Dump full request JSON when in debug mode
 		if daemon.Debug && r.Method != "GET" && util.IsJSONRequest(r) {
 			newBody := &bytes.Buffer{}
--- a/lxd/db/openfga/openfga.go
+++ b/lxd/db/openfga/openfga.go
@@ -8,13 +8,17 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
+	"sync/atomic"
 
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
 	"github.com/openfga/openfga/pkg/storage"
 
+	"github.com/canonical/lxd/lxd/auth"
 	"github.com/canonical/lxd/lxd/db"
 	"github.com/canonical/lxd/lxd/db/cluster"
 	"github.com/canonical/lxd/lxd/db/query"
+	"github.com/canonical/lxd/lxd/request"
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/entity"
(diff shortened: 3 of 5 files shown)


lxd/daemon.go
```diff
@@ -754,6 +754,9 @@ func (d *Daemon) createCmd(restAPI *mux.Router, version string, c APIEndpoint) {
 			return
 		}
 
+		// Generate an OpenFGA request ID
```
markylaing, line 757: Can you update the comment here please?
Suggestion `Set OpenFGA cache in request context to improve driver performance`

lxd/daemon.go
```diff
@@ -754,6 +754,9 @@ func (d *Daemon) createCmd(restAPI *mux.Router, version string, c APIEndpoint) {
 			return
```
markylaing, line 754: Suggested change for commit message: `lxd: Set OpenFGA cache in request context for all requests`.

lxd/db/openfga/openfga.go
```diff
@@ -205,6 +219,79 @@ func (o *openfgaStore) ReadUserTuple(ctx context.Context, store string, tk *open
...
+	if err != nil {
+		return err
+	}
+
+	// Rearrange our map to optimise for ReadUsersetTuples and ReadStartingWithUser.
+	// - ReadUsersetTuples asks the question: Which groups have this permission?
+	//   So it essentially maps permissions to a list of groups. This is the permissionCacheByEntityType.
+	// - ReadStartingFromUser asks the question: Which entities does this group have this entitlement on?
+	//   So it essentially maps groups to permissions. This is the permisionCacheByGroup.
+	permissionCacheByEntityType := make(map[entity.Type]map[auth.Entitlement]map[int][]string)
```
markylaing, line 252: Could you please move these descriptions to their definition in the OpenFGA cache struct?

lxd/db/openfga/openfga.go
```diff
@@ -245,13 +332,82 @@ func (o *openfgaStore) ReadUsersetTuples(ctx context.Context, store string, filt
...
+		if err != nil {
+			return nil, err
+		}
+
+		return storage.NewStaticTupleIterator(usersetTuples(filter.Object, filter.Relation, groups)), nil
+	}
+
+	err = o.ensureCacheLoaded(ctx, cache)
+	if err != nil {
+		return nil, err
```
markylaing, line 350: Could you please add some error wrapping?

gabrielmougard:
Thanks! Fixing this right away

lxd/db/openfga/openfga.go
gabrielmougard: fixed

lxd/db/openfga/openfga.go
gabrielmougard: fixed

lxd/daemon.go
gabrielmougard: fixed

lxd/daemon.go
gabrielmougard: fixed

gabrielmougard:
@tomponline @markylaing updated (thanks Mark for helping documenting this)

markylaing (approved):
LGTM, thanks!

lxd/db/openfga/openfga.go
```diff
@@ -248,10 +345,93 @@ func (o *openfgaStore) ReadUsersetTuples(ctx context.Context, store string, filt
 		return nil, fmt.Errorf("ReadUsersetTuples: Failed to parse entity URL %q: %w", entityURL, err)
 	}
 
-	var groupNames []string
+	// Get cache from context. If it is not present, we'll fall back to calling the database directly.
+	cache, err := request.GetCtxValue[*RequestCache](ctx, request.CtxOpenFGARequestCache)
```
tomponline, line 349: why do we need this section, if the cache is not in the context we have a bug right? Why would we mask over that here?

lxd/auth/drivers/openfga.go
```diff
@@ -84,10 +84,6 @@ func (e *embeddedOpenFGA) load(ctx context.Context, identityCache *identity.Cach
 		server.WithDatastore(opts.openfgaDatastore),
 		// Use our logger.
 		server.WithLogger(openfgaLogger{l: e.logger}),
-		// Set the max concurrency to 1 for both read and check requests.
```
tomponline, line 87: why did we need this before? 

What changed?

The commit message would be improved with a *why* rather than just a *what*.

tomponline (commented):
I have some questions about the concurrency aspect.

- Why did we restrict openfga concurrency before and why dont we need to now?
- For the per-request cache in the context, in what scenarios will openfga concurrently read or write to the cache for a single request?

lxd/db/openfga/openfga.go
markylaing: I asked @gabrielmougard to do this in the original PR; which performed caching slightly differently, but the reasoning is the same.

The reason is that for long running requests we may not want to use the cache. For example, currently have an open issue to improve authorization for the events API. I anticipate that this will check `can_view` on a per-resource basis. If we use this cache for this kind of request, we'll have a snapshot of permissions at the time of the original request, but this may change while the web socket is still connected.

So this logic is in anticipation of setting the value of the request cache to nil in the request context, in the edge cases where we don't want to use a cache.

lxd/auth/drivers/openfga.go
tomponline: > With caching there is now a significant benefit to concurrency. Since pre-fetching the cache on the first encountered call means that subsequent calls with the same cache may not need to hit the DB.

So let me check I understand this.

There is a single long-lived openfga embedded instance, so enabling concurrency allows for that openfga server to check access for multiple API requests concurrently right?

But within a single request, the cache exists only for the duration of the request, but what is concurrently happening in the openga server then for a single request? Are you saying that we are pushing down the locking from openga server-level to request-level inside the request cache itself?

tomponline: BTW the reason I'm asking is im trying to understand whether we actually need the complexity of 2 rw locks + an atomic variable or whether we can use a single mutex and check for whether the variables themselves are initialized.

tomponline: @markylaing explained in a call that the openfga server will traverse the module and make concurrent calls to the DB driver (and thus the cache) even for a single API request.

PR #14562, opened 2024-11-29
doc/support: detail the 3 levels of LTS support

simondeziel:


simondeziel:
Assuming we are OK with pointing to Ubuntu Pro for extended OS vendor support, I'd need @minaelee's help with the footnote :) ATM it is broken as we can see the `[1]` reference on https://canonical-ubuntu-documentation-library--14562.com.readthedocs.build/lxd/en/14562/support/

doc/support.md
```diff
@@ -8,13 +8,22 @@ LXD maintains different release branches in parallel.
 Long term support (LTS) releases
 : The current LTS releases are LXD 5.21.x (snap channel `5.21/stable` - this is the default channel), LXD 5.0.x (snap channel `5.0/stable`) and LXD 4.0.x (snap channel `4.0/stable`).
 
+  An LTS release starts its life with full support, then moves to maintenance support and finally to extended support. OS vendors may provide additional support[1].
```
minaelee, line 11:
```suggestion
  An LTS release starts its life with full support, then moves to maintenance support, and finally to extended support. OS vendors might provide additional support[^1].
```

doc/support.md
```diff
@@ -32,3 +41,5 @@ Feature releases
     :start-after: <!-- Include start support -->
     :end-before: <!-- Include end support -->
 ```
+
+[1]: Canonical provides additional support through the [Ubuntu Pro](https://ubuntu.com/pro) offering.
```
minaelee, line 45:
```suggestion
[^1]: Canonical provides additional support through the [Ubuntu Pro](https://ubuntu.com/pro) offering.
```

minaelee (dismissed):

doc/support.md
simondeziel: Looks like that didn't work :/

minaelee: The footnote syntax is correct and works in local build (as well as in a couple other pages that are live), but I wonder if possibly the include directive that's between the footnote reference and the corresponding definition is confusing the RTD build.

You might try moving the footnote definition above the include directive, or I can try spinning up a test RTD site in my morning and testing it there, and I can let you know if it makes a difference.

tomponline:
@simondeziel please can you remove references to `latest/*` and `6.1/stable` in the docs?

https://documentation.ubuntu.com/lxd/en/latest/installing/ in Feature releases

tomponline:
I updated https://discourse.ubuntu.com/t/managing-the-lxd-snap-package/37214 too

tomponline:
@simondeziel @minaelee it would be useful to mention the `--revision` snap flag for cluster users to avoid this situation 

https://github.com/canonical/lxd/issues/14575#issuecomment-2514540731

doc/support.md
minaelee: I tried it on a test RTD site, and as I suspected, the issue did have to do with the RTD build regarding footnotes and directives.

It's not the include directive (to the README) that's the issue though, the issue seems to be that the comments on lines 4 and 39 (`<!-- Include start release -->` and `<!-- Include end release -->`) delineate a section of Markdown to be inserted via an include directive into another page, `installing.md` (lines 18-21 there). The fix should hopefully be to move the footnote reference so that it's before the `<!-- Include end release -->` comment, which should add it to the included section and keep RTD from complaining that there's a footnote reference without a definition.

I will add it as a suggested change and we can see if that does the trick.

simondeziel:
> @simondeziel @minaelee it would be useful to mention the `--revision` snap flag for cluster users to avoid this situation

I add a note about the use of cohorts when installing (we only ever mentioned them for refreshes) as I feel that's going to avoid most of those problems. I added the `--revision` bit as well because it's possible that with the existing cluster being on `hold` or just not yet refreshed, new cluster members would get a different rev.

tomponline:
> I added the `--revision` bit as well because it's possible that with the existing cluster being on `hold` or just not yet refreshed, new cluster members would get a different rev.

Thanks, yeah this is what happened to the user I linked to.

doc/support.md
```diff
@@ -8,23 +8,33 @@ LXD maintains different release branches in parallel.
 Long term support (LTS) releases
 : The current LTS releases are LXD 5.21.x (snap channel `5.21/stable` - this is the default channel), LXD 5.0.x (snap channel `5.0/stable`) and LXD 4.0.x (snap channel `4.0/stable`).
 
+  An LTS release starts its life with full support, then moves to maintenance support, and finally to extended support. OS vendors might provide additional support[^1].
+
+  - Full support: frequent bugfixes and security updates are provided.
```
tomponline, line 13: we could also mention that it gets some new features too.

doc/support.md
```diff
@@ -8,23 +8,33 @@ LXD maintains different release branches in parallel.
...
     It gets frequent bugfix and security updates, but does not receive any feature additions.
     Updates to this release happen approximately every six months, but this schedule should be seen as a rough estimation that can change based on priorities and discovered bugs.
+    Currently in full support phase.
   - LXD 5.0 is supported until June 2027.
+    Currently in maintenance support phase.
   - LXD 4.0 is supported until June 2025.
+    Currently in extended support phase.
 
 Feature releases
 : After LXD 5.21 is released, the next feature release will be LXD 6.x (starting with 6.1).
```
tomponline, line 29: can we say here "the next feature release series will be LXD 6.x (starting with 6.1), available from the `6/stable` channel"?

tomponline (approved):

tomponline:
looks like doc tests still failing from my earlier PR merge.

doc/installing.md
```diff
@@ -64,6 +64,20 @@ Complete the following steps to install the snap:
 
         sudo snap install lxd --channel=5.0/stable
 
+```{note}
+LXD cluster members all need to use the exact same version of LXD and this requires special consideration due to how snaps are deployed as explained in {ref}`Keep cluster members in sync<howto-snap-cohort>`. To ensure all the cluster members are offered the same snap revision, it is possible to specify a cohort parameter which will bypass the progressive deployment:
```
tomponline, line 68: @minaelee @simondeziel I wonder whether this should be in https://documentation.ubuntu.com/lxd/en/latest/howto/cluster_manage/#upgrade-cluster-members and linked to from here or vice versa. There's significant overlap in these pages.

doc/support.md
simondeziel: Yes, that did work, thanks!

doc/installing.md
simondeziel: I do agree that our installation/upgrade/snap doc is all over the place and especially messy when we consider clusters needing cohorts often wanting holds.

That doc reorg feels like a bigger task, a needed one that's for sure. Do you have any objections to having it split off from that PR though?

simondeziel:
@tomponline tests have now passed and I believe I've addressed @minaelee's feedback.

tomponline (approved):

Issue #14561, opened 2024-11-29
[regression]: `lxc image list <OFFICIAL REMOTE>:` broken on `latest/edge`


simondeziel:
Listing public images doesn't work with LXD from `latest/edge`:

```
$ lxc image list ubuntu-minimal-daily:
Error: The remote isn't a private LXD server
```
But that remote config seems to be in order:

```
$ lxc remote list | sed -n '2,3 p; /ubuntu-minimal-daily/ p'
|                NAME                 |                        URL                        |   PROTOCOL    |  AUTH TYPE  | PUBLIC | STATIC | GLOBAL |
+-------------------------------------+---------------------------------------------------+---------------+-------------+--------+--------+--------+
| ubuntu-minimal-daily                | https://cloud-images.ubuntu.com/minimal/daily/    | simplestreams | none        | YES    | YES    | NO     |
```

Other official remotes are also affected but they all work with `5.21/stable` and `latest/candidate`.

Additional information:

```
$ lxc version
Client version: 6.2
Server version: 6.2
$ snap list lxd
Name  Version      Rev    Tracking     Publisher   Notes
lxd   git-dc7b9b1  31502  latest/edge  canonical✓  -
```

kadinsayani:
This regression was probably introduced by https://github.com/canonical/lxd/pull/14260. I'll have a PR up with a fix pretty quickly, I think I've tracked down the problem. More to follow.

PR #14563, opened 2024-11-29
LXC: Fix `lxc image list` regression

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/14561 introduced with https://github.com/canonical/lxd/pull/14260.

This PR fixes a regression introduced with https://github.com/canonical/lxd/pull/14260 by adding `GetImagesAllProjects` and
`GetImagesAllProjectsWithFilter` functions to `client/simplestreams_images.go`, and moving their method signatures to
the `ImageServer` interface. While adding support for fetching images across all projects, we changed the logic in the `lxc image list` command to call different functions based on whether the `--all-projects` flag is passed in or not:

https://github.com/canonical/lxd/pull/14260/commits/1fae6e6d99d68538ae29f01937d4dfc8338bf923

This is fine, but we needed to add `simplestreams` protocol support as well to facilitate listing public images from remotes such as `images` and `ubuntu-minimal-daily`.

The actual bug comes from the following LOC: 

https://github.com/canonical/lxd/blob/4a3c4a343e4a9490105bd3da893f9892b86e01d8/lxc/image.go#L1356-L1359

Fetching from an instance server isn't applicable for `simplestreams` servers, so `GetImagesAllProjects` is never reached due to the error raised by `c.global.conf.GetInstanceServer`. Instead, we should fetch an image server before fetching images, hence why moving the new client functions method signatures to the `ImageServer` interface fixes the regression.

### DIFF
--- a/client/interfaces.go
+++ b/client/interfaces.go
@@ -50,6 +50,8 @@ type ImageServer interface {
 	GetImages() (images []api.Image, err error)
 	GetImageFingerprints() (fingerprints []string, err error)
 	GetImagesWithFilter(filters []string) (images []api.Image, err error)
+	GetImagesAllProjects() (images []api.Image, err error)
+	GetImagesAllProjectsWithFilter(filters []string) (images []api.Image, err error)
 
 	GetImage(fingerprint string) (image *api.Image, ETag string, err error)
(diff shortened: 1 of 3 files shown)


tomponline (approved):
Thanks for quick fix

PR #14564, opened 2024-11-30
Test image list of some builtin remotes

simondeziel:


test/suites/image.sh
```diff
@@ -83,11 +83,24 @@ test_image_list_all_aliases() {
...
+
+    lxc remote list -f csv | while IFS=, read -r name url _; do
+        if [ "${name}" != "images" ] && [ "${name}" != "ubuntu-minimal" ]; then
+            continue
+        fi
+
+        # Check if there is connectivity
+        curl --head --silent "${url}" > /dev/null || continue
 
+        lxc image list "${name}:" > /dev/null
```
kadinsayani, line 102: Can we redirect to `stderr` if `lxc image list` throws an error?

tomponline (approved):

PR #14576, opened 2024-12-03
README: Update references to latest documentation to stable-5.21 (stable-5.21)

tomponline:
We should link to the relevant LTS series docs in the LTS code branches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,11 +14,11 @@ You should consider using LXD if you want to containerize different environments
 
 ## Get started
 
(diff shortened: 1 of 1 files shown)


simondeziel (approved):
Should we have a lint check comparing the git branch with the versioned URL? Similar to the Go version requirement we check in GH CI?

tomponline:
> Should we have a lint check comparing the git branch with the versioned URL? Similar to the Go version requirement we check in GH CI?

Maybe, although its really only the readme file that has these absolute links so i dont think it will come up often.

minaelee (approved):

PR #14577, opened 2024-12-03
Make `shared.IsSnapshot()` a bit more strict

simondeziel:


lxd/instance/instance_utils.go
```diff
@@ -302,10 +302,8 @@ func LoadInstanceDatabaseObject(ctx context.Context, tx *db.ClusterTx, project,
 	var container *cluster.Instance
 	var err error
 
-	if strings.Contains(name, shared.SnapshotDelimiter) {
-		parts := strings.SplitN(name, shared.SnapshotDelimiter, 2)
-		instanceName := parts[0]
-		snapshotName := parts[1]
+	if shared.IsSnapshot(name) {
```
tomponline, line 305: Maybe we should just use `api.GetParentAndSnapshotName` everywhere (or at least in more places) and update it to use strings.Cut. This way we don't have to repeat the check for `/` in the string.

shared/util.go
```diff
@@ -651,9 +651,10 @@ func (r BytesReadCloser) Close() error {
 	return nil
 }
 
-// IsSnapshot returns true if a given name contains the snapshot delimiter.
+// IsSnapshot returns true if a given name contains a snapshot delimiter not at the end.
 func IsSnapshot(name string) bool {
```
tomponline, line 655: Probably we should centralise the "is snapshot/extract snapshot parts" into api.GetParentAndSnapshotName instead.

tomponline (commented):
I've not gone through this in detail yet, but one comment is that we've consciously avoided using `IsSnapshot` or `api.GetParentAndSnapshotName` in the past in places where the variable being checked isn't representing a possible instance or volume fully qualified snapshot name.

I.e instance backups follows the same style but they represent different things and could evolve separately, so maybe we also need a separate function for other conceptual fully qualified entities.

tomponline:
@simondeziel can we close this for now until you get chance to look at this again?

PR #14582, opened 2024-12-04
Storage: Per project image and backup storage options

boltmark:
This PR adds support for per-project image and backup storage options. We maintain the option to select storage options at the server level, and add a more fine-grained option to select this at the project level.

---

Overview: We currently support configuration of storage.backups_volume and storage.images_volume on the server level. We would like to make this option more granular by introducing the ability to configure it on a per-project basis. We will continue to support the server-level configuration, but project-level configuration will take priority.

If these options are configured on a given project, then all backup/image tarballs created for resources within said project will be stored in the selected volume. If not, then we will default to the config set at the server level (if any).

Implementation:
We will maintain support for `storage.[images,backups]_volume` on the server level, and add this config option at the project level. Users may specify `storage.images_volume` and `storage.backups_volume` at the project level, which will ensure images or backups for the given project are stored on the specified storage volume. These config keys can be changed, and data will move accordingly.

Images and backups will continue to be stored in the images and backups directory. However, we need some way to be able to store images and backups for a certain projects on a specified storage volume. We will follow the current pattern of using a symlink to the mountpoint of the specified volume. Special directories `backups/daemon` and `images/daemon` will be created, and will function exactly as the images and backups directories function now -- to store images and backups that are not project specific. If `storage.[images,backups]_volume` is set at the server level, then these will be symlinked to the selected volume's mountpoint. 

For each project that is created, directories `images/project_X` and `backups/project_X` will also be created, which will store images and backups for the given project. If `storage.[images,backups]_volume` is set at the project level, then the directory will be symlinked to the mountpoint of the selected storage volume. If not set on the project level, then the directory will be symlinked to `[images,backups]/daemon/project_X`. This follows the current implementation at the server level. When project-level `storage.[images,backups]_volume` is changed or unset, data is moved accordingly.

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2535,3 +2535,7 @@ Adds a new {config:option}`device-unix-hotplug-device-conf:ownership.inherit` co
 ## `unix_device_hotplug_subsystem_device_option`
 
 Adds a new {config:option}`device-unix-hotplug-device-conf:subsystem` configuration option for `unix-hotplug` devices. This adds support for detecting `unix-hotplug` devices by subsystem, and can be used in conjunction with {config:option}`device-unix-hotplug-device-conf:productid` and {config:option}`device-unix-hotplug-device-conf:vendorid`.
+
+## `project_daemon_storage`
+This introduces two new configuration keys `storage.images_volume` and `storage.backups_volume` to the project level to allow for a storage volume
+on an existing pool be used for storing the project-wide images and backups artifacts.
--- a/lxd/backup.go
+++ b/lxd/backup.go
@@ -101,7 +101,7 @@ func backupCreate(s *state.State, args db.InstanceBackup, sourceInst instance.In
 	}
 
 	// Create the target path if needed.
-	backupsPath := shared.VarPath("backups", "instances", project.Instance(sourceInst.Project().Name, sourceInst.Name()))
+	backupsPath := shared.VarPath("backups", fmt.Sprintf("project_%s", op.Project()), "instances", project.Instance(sourceInst.Project().Name, sourceInst.Name()))
 	if !shared.PathExists(backupsPath) {
 		err := os.MkdirAll(backupsPath, 0700)
 		if err != nil {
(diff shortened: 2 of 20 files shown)


doc/api-extensions.md
```diff
@@ -2535,3 +2535,7 @@ Adds a new {config:option}`device-unix-hotplug-device-conf:ownership.inherit` co
 ## `unix_device_hotplug_subsystem_device_option`
 
 Adds a new {config:option}`device-unix-hotplug-device-conf:subsystem` configuration option for `unix-hotplug` devices. This adds support for detecting `unix-hotplug` devices by subsystem, and can be used in conjunction with {config:option}`device-unix-hotplug-device-conf:productid` and {config:option}`device-unix-hotplug-device-conf:vendorid`.
+
+## `project_daemon_storage`
+This introduces two new configuration keys `storage.images_volume` and `storage.backups_volume` to the project level to allow for a storage volume
```
minaelee, line 2540:
```suggestion
This introduces two new configuration keys, `storage.images_volume` and `storage.backups_volume`, to the project level. These keys allow for a storage volume
```

doc/api-extensions.md
```diff
@@ -2535,3 +2535,7 @@ Adds a new {config:option}`device-unix-hotplug-device-conf:ownership.inherit` co
...
+on an existing pool be used for storing the project-wide images and backups artifacts.
```
minaelee, line 2541:
```suggestion
on an existing pool to be used for storing the project-wide images and backups artifacts.
```

lxd/backup.go
```diff
@@ -101,7 +101,7 @@ func backupCreate(s *state.State, args db.InstanceBackup, sourceInst instance.In
 	}
 
 	// Create the target path if needed.
-	backupsPath := shared.VarPath("backups", "instances", project.Instance(sourceInst.Project().Name, sourceInst.Name()))
+	backupsPath := shared.VarPath("backups", fmt.Sprintf("project_%s", op.Project()), "instances", project.Instance(sourceInst.Project().Name, sourceInst.Name()))
```
tomponline, line 104: we should move this logic into a function in the instance interface I think to avoid the duplication of logic, something like `sourceInstance.BackupsPath()`.

boltmark: How would we handle cases where a reference to the instance is not available in the function/call stack? For example https://github.com/canonical/lxd/blob/main/lxd/images.go#L830.

Maybe I'm missing something here

tomponline: no you're not, we wouldnt be able to use an instance level helper function for images as they arent associated to instances.

skozina:
We should also ensure that the storage volume used for image/backup store is always available. That means, adjust daemon startup so that such volumes would be mounted right in `storageStartup` (rather than `daemonStorageMount` as we do today), and similarly ensure that the volumes are mounted any time the respective project config options are set.

tomponline:
@skozina shall we close this for now?

skozina:
Yes, let me close this. There are couple of comments I'd like to keep in mind, but the PR should be closed. Thanks!

Issue #14570, opened 2024-12-02
Auto-completion for `unset` ignores `user.*` configuration keys


MusicDin:
The auto-completion for `unset` works impressively well for known keys.
```sh
$ lxc config set c1 limits.cpu=4
$ lxc config unset c1 [TAB]
limits.cpu
...
```

However, `user.` keys are ignored:
```sh
$ lxc config set c1 limits.cpu=4
$ lxc config set c1 user.test=123
$ lxc config unset c1 [TAB]
limits.cpu 
... # No option for user.test
```

@kadinsayani Is there a reason `user.*` keys are ignored? If so, feel free to close the issue.

PR #13051, opened 2024-03-05
Reject limits kernel vm config

MggMuggins:
Fixes https://github.com/canonical/lxd/issues/12874

Should I write a test?

PR #14584, opened 2024-12-04
LXC: Improve configuration key validation and add missing completions for `lxc config unset`

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/14570.

This PR reworks configuration key validation in `validConfigKey`, in `lxd/instance/instance_utils.go`. The intent of the first two commits is to:
1. Validate keys that begin with a prefix, such as `user.*` and `linux.sysctl.*`; and
2. Validate keys based on their instance type.

For example, when attempting to set `linux.sysctl.*` for a VM instance, a meaningful error message should be shown indicating the key is not valid for VMs, rather than the generic `Unknown configuration key: <key>`.

This PR also expands the use of the new slices `instancetype.ConfigKeyPrefixesAny` and `instancetype.ConfigKeyPrefixesContainer` to contextually provide completions for prefixed config keys when running`lxc config unset`, based on instance type.

### DIFF
--- a/lxc/completion.go
+++ b/lxc/completion.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/canonical/lxd/lxd/instance/instancetype"
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
 )
@@ -357,8 +358,10 @@ func (g *cmdGlobal) cmpInstanceSetKeys(instanceName string) ([]string, cobra.She
 		return nil, cobra.ShellCompDirectiveError
 	}
 
-	// Fetch all config keys that can be set by a user.
-	allInstanceConfigKeys, _ := g.cmpInstanceAllKeys(instanceName)
+	instanceType := instance.Type
+
+	// Fetch all config keys that can be set by a user based on instance type.
+	allInstanceConfigKeys, _ := g.cmpInstanceKeys(instanceName)
 
 	// Convert slice to map[string]struct{} for O(1) lookups.
 	keySet := make(map[string]struct{}, len(allInstanceConfigKeys))
@@ -374,7 +377,11 @@ func (g *cmdGlobal) cmpInstanceSetKeys(instanceName string) ([]string, cobra.She
 		// We only want to return the intersection between allInstanceConfigKeys and configKeys to avoid returning the full instance config.
 		_, exists := keySet[configKey]
 		if exists {
-			configKeys = append(configKeys, configKey)
+			if shared.StringHasPrefix(configKey, instancetype.ConfigKeyPrefixesAny...) {
+				configKeys = append(configKeys, configKey)
+			} else if instanceType == string(api.InstanceTypeContainer) && shared.StringHasPrefix(configKey, instancetype.ConfigKeyPrefixesContainer...) {
+				configKeys = append(configKeys, configKey)
+			}
 		}
 	}
 
--- a/lxd/instance/instance_utils.go
+++ b/lxd/instance/instance_utils.go
@@ -128,6 +128,22 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
 }
 
 func validConfigKey(os *sys.OS, key string, value string, instanceType instancetype.Type) error {
+	// Check if the key is a valid prefix and whether or not it requires a subkey.
+	knownPrefixes := append(instancetype.ConfigKeyPrefixesAny, instancetype.ConfigKeyPrefixesContainer...)
+	if strings.HasSuffix(key, ".") {
+		// Disallow keys with container-specific prefixes such as "linux.sysctl." and "limits.kernel." for VMs.
+		if instanceType == instancetype.VM && shared.StringHasPrefix(key, instancetype.ConfigKeyPrefixesContainer...) {
+			return fmt.Errorf("%q isn't supported for %q", key, instanceType)
+		}
+
+		if !(key == instancetype.ConfigVolatilePrefix || shared.ValueInSlice(key, knownPrefixes)) {
+			// Not a known prefix.
+			return fmt.Errorf("Unknown configuration key: %q", key)
+		}
+
+		return fmt.Errorf("%q requires a subkey", key)
+	}
+
 	f, err := instancetype.ConfigKeyChecker(key, instanceType)
 	if err != nil {
 		return err
--- a/lxd/instance/instance_utils.go
+++ b/lxd/instance/instance_utils.go
@@ -144,6 +144,24 @@ func validConfigKey(os *sys.OS, key string, value string, instanceType instancet
 		return fmt.Errorf("%q requires a subkey", key)
 	}
 
+	// Validate the configuration key against instance type for containers and VMs.
+	// Ignore configuration keys with known prefixes since usage has already been validated, and ConfigKeyChecker validates keys syntactically.
+	if instanceType != instancetype.Any && !shared.StringHasPrefix(key, knownPrefixes...) && !strings.HasPrefix(key, instancetype.ConfigVolatilePrefix) {
+		// Ensure key is present in instance config key map based on type.
+		exists := false
+		switch instanceType {
+		case instancetype.VM:
+			_, exists = instancetype.InstanceConfigKeysVM[key]
+		case instancetype.Container:
+			_, exists = instancetype.InstanceConfigKeysContainer[key]
+		}
+
+		_, existsAny := instancetype.InstanceConfigKeysAny[key]
+		if !(exists || existsAny) {
+			return fmt.Errorf("%q isn't supported for %q", key, instanceType)
+		}
+	}
+
 	f, err := instancetype.ConfigKeyChecker(key, instanceType)
 	if err != nil {
 		return err
@@ -151,10 +169,6 @@ func validConfigKey(os *sys.OS, key string, value string, instanceType instancet
 
 	if err = f(value); err != nil {
 		return err
-	}
-
-	if strings.HasPrefix(key, "limits.kernel.") && instanceType == instancetype.VM {
-		return fmt.Errorf("%s isn't supported for VMs", key)
 	}
 
 	if key == "raw.lxc" {
--- a/lxd/instance/instancetype/instance.go
+++ b/lxd/instance/instancetype/instance.go
@@ -28,6 +28,12 @@ const (
 // ConfigVolatilePrefix indicates the prefix used for volatile config keys.
 const ConfigVolatilePrefix = "volatile."
 
+// ConfigKeyPrefixesAny indicates valid prefixes for configuration options.
+var ConfigKeyPrefixesAny = []string{"environment.", "user.", "image."}
+
+// ConfigKeyPrefixesContainer indicates valid prefixes for container configuration options.
+var ConfigKeyPrefixesContainer = []string{"linux.sysctl.", "limits.kernel."}
+
 // ValidName validates an instance name. There are different validation rules for instance snapshot names
 // so it takes an argument indicating whether the name is to be used for a snapshot or not.
 func ValidName(instanceName string, isSnapshot bool) error {
--- a/lxd/instance/instancetype/instance.go
+++ b/lxd/instance/instancetype/instance.go
@@ -1280,25 +1280,12 @@ func ConfigKeyChecker(key string, instanceType Type) (func(value string) error, 
 		}
 	}
 
-	if strings.HasPrefix(key, "environment.") {
+	if (instanceType == Any || instanceType == Container) && strings.HasPrefix(key, "linux.sysctl.") {
 		return validate.IsAny, nil
 	}
 
-	if strings.HasPrefix(key, "user.") {
-		return validate.IsAny, nil
-	}
-
-	if strings.HasPrefix(key, "image.") {
-		return validate.IsAny, nil
-	}
-
-	if strings.HasPrefix(key, "limits.kernel.") &&
-		(len(key) > len("limits.kernel.")) {
-		return validate.IsAny, nil
-	}
-
-	if (instanceType == Any || instanceType == Container) &&
-		strings.HasPrefix(key, "linux.sysctl.") {
+	knownPrefixes := append(ConfigKeyPrefixesAny, ConfigKeyPrefixesContainer...)
+	if shared.StringHasPrefix(key, knownPrefixes...) {
 		return validate.IsAny, nil
 	}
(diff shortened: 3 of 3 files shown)


lxc/completion.go
```diff
@@ -372,8 +372,10 @@ func (g *cmdGlobal) cmpInstanceSetKeys(instanceName string) ([]string, cobra.She
 
 	for configKey := range instance.Config {
 		// We only want to return the intersection between allInstanceConfigKeys and configKeys to avoid returning the full instance config.
+		// We also need to account for user defined keys such as user.* and linux.sysctl.*.
 		_, exists := keySet[configKey]
-		if exists {
+		if exists || strings.HasPrefix(configKey, "user.") ||
+			strings.HasPrefix(configKey, "linux.sysctl.") {
```
simondeziel, line 378: Searching for `.*` on https://documentation.ubuntu.com/lxd/en/latest/reference/instance_options/ I think only the `environment.` prefix is missing from your extended list.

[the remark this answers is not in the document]
kadinsayani: Thanks for catching that!

lxc/completion.go
```diff
@@ -372,8 +372,11 @@ func (g *cmdGlobal) cmpInstanceSetKeys(instanceName string) ([]string, cobra.She
 
 	for configKey := range instance.Config {
 		// We only want to return the intersection between allInstanceConfigKeys and configKeys to avoid returning the full instance config.
+		// We also need to account for user defined keys such as user.* and linux.sysctl.*.
 		_, exists := keySet[configKey]
-		if exists {
+		if exists || strings.HasPrefix(configKey, "user.") ||
+			strings.HasPrefix(configKey, "linux.sysctl.") ||
+			strings.HasPrefix(configKey, "environment.") {
```
simondeziel, line 379: We have `shared.StringPrefixInSlice()` that could be used here.

simondeziel (approved):

lxc/completion.go
kadinsayani: I think `shared.StringPrefixInSlice()` would introduce extra complexity since we're already looping through `instance.Config`. Nice to know we have this function though!

simondeziel, line 379: I just realized I pointed you to the wrong `shared.` util. Here's what I meant:
```suggestion
		if exists || shared.StringHasPrefix(configKey, "user.", "linux.sysctl.", "environment.") {
```

kadinsayani: Gotcha - this one will work well :)

simondeziel (dismissed):

lxc/completion.go
```diff
@@ -372,8 +372,9 @@ func (g *cmdGlobal) cmpInstanceSetKeys(instanceName string) ([]string, cobra.She
 
 	for configKey := range instance.Config {
 		// We only want to return the intersection between allInstanceConfigKeys and configKeys to avoid returning the full instance config.
+		// We also need to account for user defined keys such as user.* and linux.sysctl.*.
 		_, exists := keySet[configKey]
-		if exists {
+		if exists || shared.StringHasPrefix(configKey, "user.", "linux.sysctl.", "environment.") {
```
tomponline, line 377: Shouldnt we make this a slice of prefixes somewhere so that both the lxc client and the lxd server can use it when validating keys so we can see the references ? Otherwise how will we remember to update this list when a new prefix is added?


For example, it looks like there's some other prefix keys not accounted for

limits.kernel.
image.
environment.

See `ConfigKeyChecker()`

kadinsayani: Good call, I'll add the slices to `lxd/instance/instancetype`.

lxd/instance/instancetype/instance.go
```diff
@@ -1274,15 +1282,8 @@ func ConfigKeyChecker(key string, instanceType Type) (func(value string) error,
...
-		return validate.IsAny, nil
-	}
-
-	if strings.HasPrefix(key, "user.") {
-		return validate.IsAny, nil
-	}
-
-	if strings.HasPrefix(key, "image.") {
+	if (instanceType == Any || instanceType == Container) &&
+		strings.HasPrefix(key, "linux.sysctl.") {
```
simondeziel, line 1286: I think `limits.kernel.` should be treated the same, see https://github.com/canonical/lxd/pull/13051/commits/f0d0128ed636065080607170969244033b2d4616 which I think might have been better done here?

I'm mentioning it here because you introduced `ConfigKeyPrefixesContainer` in the PR.

kadinsayani: Yeah, `limits.kernel.` is a container only config, good catch.

lxd/instance/instance_utils.go
```diff
@@ -128,6 +128,42 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
...
+
+	if shared.StringHasPrefix(key, instancetype.ConfigKeyPrefixesAny...) {
+		for _, configKey := range instancetype.ConfigKeyPrefixesAny {
+			if len(key) <= len(configKey) {
+				return fmt.Errorf("%s requires a subkey", key)
+			}
+		}
+	}
+
+	if shared.StringHasPrefix(key, instancetype.ConfigKeyPrefixesContainer...) {
```
simondeziel, line 159:
```suggestion
	} else if shared.StringHasPrefix(key, instancetype.ConfigKeyPrefixesContainer...) {
```

simondeziel: Also, could those 2 `shared.StringHasPrefix()` checks be moved in the top `if`?

lxd/instance/instance_utils.go
```diff
@@ -128,6 +128,42 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
 }
 
 func validConfigKey(os *sys.OS, key string, value string, instanceType instancetype.Type) error {
+	if !strings.HasPrefix(key, instancetype.ConfigVolatilePrefix) {
```
simondeziel, line 131: I'm curious why `volatile.*` are special and not say `image.*`?

kadinsayani: For `volatile.*` keys, we check suffixes. For example:

https://github.com/canonical/lxd/blob/ac7aa741dc57cd7eb0a07dd7b5a66b9658a6e760/lxd/instance/instancetype/instance.go#L1139-L1159

kadinsayani: So we don't want to rule them as invalid, but rather pass them to `ConfigKeyChecker`.

lxd/instance/instancetype/instance.go
```diff
@@ -28,6 +28,12 @@ const (
 // ConfigVolatilePrefix indicates the prefix used for volatile config keys.
 const ConfigVolatilePrefix = "volatile."
 
+// ConfigKeyPrefixesAny indicates valid prefixes for configuration options.
+var ConfigKeyPrefixesAny = []string{"environment.", "user.", "image."}
```
simondeziel, line 32: Should this slice also include `ConfigVolatilePrefix`?

kadinsayani: Sure!

lxd/instance/instance_utils.go
```diff
@@ -128,6 +128,45 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
...
+	for _, prefix := range append(instancetype.ConfigKeyPrefixesAny, instancetype.ConfigKeyPrefixesContainer...) {
+		if strings.HasPrefix(key, prefix) {
+			matchedPrefix = prefix
+			break
+		}
+	}
+
+	if matchedPrefix != "" && len(key) <= len(matchedPrefix) {
+		return fmt.Errorf("%s requires a subkey", key)
+	}
```
simondeziel, line 168: I think this simplification would work
```suggestion
	for _, prefix := range append(instancetype.ConfigKeyPrefixesAny, instancetype.ConfigKeyPrefixesContainer...) {
                // Reject the key if it matches a known prefix as it indicates the subkey was omitted.
		if key == prefix {
			return fmt.Errorf("%s requires a subkey", key)
		}
	}
```

lxd/instance/instancetype/instance.go
kadinsayani: I don't think so - we want to leave handling of `volatile` keys to `ConfigKeyChecker`.

simondeziel (dismissed):

simondeziel (dismissed):

lxd/instance/instance_utils.go
```diff
@@ -128,6 +128,46 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
 }
 
 func validConfigKey(os *sys.OS, key string, value string, instanceType instancetype.Type) error {
+	if instanceType != instancetype.Any &&
+		!shared.StringHasPrefix(key, instancetype.ConfigKeyPrefixesAny...) &&
```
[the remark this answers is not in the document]
tomponline: although it doesnt appear to be a problem in this case to put it all on one line

lxd/instance/instance_utils.go
```diff
@@ -128,6 +128,46 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
...
+		!strings.HasPrefix(key, instancetype.ConfigVolatilePrefix) {
+		// Determine valid keys for instance type.
+		var validKeys map[string]func(string) error
+		containerPrefix := false
+
+		switch instanceType {
+		case instancetype.VM:
+			if shared.StringHasPrefix(key, instancetype.ConfigKeyPrefixesContainer...) {
+				return fmt.Errorf("%s isn't supported for VMs", key)
```
tomponline, line 141: Lets quote the problem key here using %q

lxd/instance/instance_utils.go
```diff
@@ -128,6 +128,46 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
 }
 
 func validConfigKey(os *sys.OS, key string, value string, instanceType instancetype.Type) error {
+	if instanceType != instancetype.Any &&
```
tomponline, line 131: we could do with a comment here explaining why instancetype.Any is ignored.

lxd/instance/instance_utils.go
```diff
@@ -128,6 +128,46 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
...
+
+			validKeys = instancetype.InstanceConfigKeysContainer
+		}
+
+		// Check if the key is valid.
+		if !containerPrefix && validKeys != nil {
+			_, exists := validKeys[key]
+			_, existsAny := instancetype.InstanceConfigKeysAny[key]
+			if !(exists || existsAny) {
+				return fmt.Errorf("%s isn't supported for %ss", key, instanceType)
```
tomponline, line 158: lets not use contractions in error messages
lets use %q for the the key

lxd/instance/instance_utils.go
tomponline, line 1: Ive had a hard time reviewing this commit as the commit message just says "improve error messages..." but doesnt explain what changes you've made intentionally, and the actual implementation contains a fair bit of logic change. So its hard to review against what you intended to do as you've not explained it.

lxd/instance/instancetype/instance.go
```diff
@@ -1280,25 +1280,11 @@ func ConfigKeyChecker(key string, instanceType Type) (func(value string) error,
 		}
 	}
 
-	if strings.HasPrefix(key, "environment.") {
+	if shared.StringHasPrefix(key, ConfigKeyPrefixesContainer...) {
```
tomponline, line 1283: we seem to have lost a check for instanceType from line 1300 here.

tomponline (changes requested):
I'm afraid i didn't get on well with reviewing this PR.

I had a hard time following the changes. Please could you expand the comments, commit descriptions and potentially refactor the code so its clearer to follow.

Thanks

lxd/instance/instance_utils.go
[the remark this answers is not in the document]
kadinsayani: Sure, sounds good. I was trying to follow the 80/120 guideline a bit too precisely at the cost of readability here. I'll make note of this moving forward.

kadinsayani:
> I'm afraid i didn't get on well with reviewing this PR.
> 
> I had a hard time following the changes. Please could you expand the comments, commit descriptions and potentially refactor the code so its clearer to follow.
> 
> Thanks

I've refactored the code, expanded the use of comments, and reworded the commit messages and descriptions.

lxd/instance/instancetype/instance.go
kadinsayani: I'll need to rework how `instancetype.Any` keys are validated so we don't lose this check.

lxd/instance/instance_utils.go
```diff
@@ -128,6 +128,22 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
 }
 
 func validConfigKey(os *sys.OS, key string, value string, instanceType instancetype.Type) error {
+	// Check if the key is a valid prefix and whether or not it requires a subkey.
+	knownPrefixes := append(instancetype.ConfigKeyPrefixesAny, instancetype.ConfigKeyPrefixesContainer...)
+	if strings.HasSuffix(key, ".") {
```
tomponline, line 133: in what scenario would this occur?

kadinsayani: If a key such as `linux.sysctl.` is passed in, it is invalid since there is no subkey provided. 

For this check, we check if a known key is passed in, if so, we inform the user that the key requires a subkey. Otherwise, the key is unknown.

lxd/instance/instance_utils.go
```diff
@@ -128,6 +128,22 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
 }
 
 func validConfigKey(os *sys.OS, key string, value string, instanceType instancetype.Type) error {
+	// Check if the key is a valid prefix and whether or not it requires a subkey.
+	knownPrefixes := append(instancetype.ConfigKeyPrefixesAny, instancetype.ConfigKeyPrefixesContainer...)
```
tomponline, line 132: can this knownPrefixes definition be reduced in scope by moving it inside the if below?

tomponline: oh i see you use it in next commit

tomponline (approved):

simondeziel:
@kadinsayani I think this PR caused a regression: https://github.com/canonical/lxd-ci/actions/runs/12367790805/job/34516721509#step:11:93

PR #14585, opened 2024-12-04
API: Add all projects support for network zones (from Incus)

kadinsayani:
This PR adds support for fetching network zones across all projects.

### DIFF
--- a/lxd/network_zones.go
+++ b/lxd/network_zones.go
@@ -283,7 +283,7 @@ func networkZonesGet(d *Daemon, r *http.Request) response.Response {
 			}
 
 			if err != nil {
-				continue
+				return response.SmartError(err)
 			}
 
 			netzoneInfo := netzone.Info()
--- a/lxd/network_zones.go
+++ b/lxd/network_zones.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -19,6 +20,7 @@ import (
 	"github.com/canonical/lxd/lxd/response"
 	"github.com/canonical/lxd/lxd/state"
 	"github.com/canonical/lxd/lxd/util"
+	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/entity"
 	"github.com/canonical/lxd/shared/version"
@@ -116,6 +118,11 @@ func networkZoneAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *h
 //      description: Project name
 //      type: string
 //      example: default
+//    - in: query
+//      name: all-projects
+//      description: Retrieve network zones from all projects
+//      type: boolean
+//      example: true
 //  responses:
 //    "200":
 //      description: API endpoints
@@ -165,6 +172,11 @@ func networkZoneAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *h
 //	    description: Project name
 //	    type: string
 //	    example: default
+//	  - in: query
+//	    name: all-projects
+//	    description: Retrieve network zones from all projects
+//	    type: boolean
+//	    example: true
 //	responses:
 //	  "200":
 //	    description: API endpoints
@@ -196,19 +208,50 @@ func networkZoneAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *h
 func networkZonesGet(d *Daemon, r *http.Request) response.Response {
 	s := d.State()
 
-	requestProjectName := request.ProjectParam(r)
-	effectiveProjectName, _, err := project.NetworkZoneProject(s.DB.Cluster, requestProjectName)
-	if err != nil {
-		return response.SmartError(err)
+	allProjects := shared.IsTrue(request.QueryParam(r, "all-projects"))
+	requestProjectName := request.QueryParam(r, "project")
+
+	// requestProjectName is only valid for project specific requests.
+	if allProjects && requestProjectName != "" {
+		return response.BadRequest(errors.New("Cannot specify a project when requesting all projects"))
+	}
+
+	var effectiveProjectName string
+	var err error
+	if !allProjects {
+		if requestProjectName == "" {
+			requestProjectName = api.ProjectDefaultName
+		}
+
+		// Project specific requests require an effective project, when "features.networks.zones" is enabled this is the requested project, otherwise it is the default project.
+		effectiveProjectName, _, err = project.NetworkZoneProject(s.DB.Cluster, requestProjectName)
+		if err != nil {
+			return response.SmartError(err)
+		}
+
+		// If the request is project specific, then set effective project name in the request context so that the authorizer can generate the correct URL.
+		request.SetCtxValue(r, request.CtxEffectiveProjectName, effectiveProjectName)
 	}
 
 	recursion := util.IsRecursionRequest(r)
 
-	var zoneNames []string
-
+	var zoneNamesMap map[string]string
 	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
-		// Get list of Network zones.
-		zoneNames, err = tx.GetNetworkZonesByProject(ctx, effectiveProjectName)
+		if allProjects {
+			zoneNamesMap, err = tx.GetNetworkZones(ctx)
+		} else {
+			// Get list of Network zones.
+			zoneNames, err := tx.GetNetworkZonesByProject(ctx, effectiveProjectName)
+			if err != nil {
+				return err
+			}
+
+			// Network zones should be mapped to the requested project for project specific requests.
+			zoneNamesMap = make(map[string]string, len(zoneNames))
+			for _, zoneName := range zoneNames {
+				zoneNamesMap[zoneName] = requestProjectName
+			}
+		}
 
 		return err
 	})
@@ -216,7 +259,6 @@ func networkZonesGet(d *Daemon, r *http.Request) response.Response {
 		return response.InternalError(err)
 	}
 
-	request.SetCtxValue(r, request.CtxEffectiveProjectName, effectiveProjectName)
 	userHasPermission, err := s.Authorizer.GetPermissionChecker(r.Context(), auth.EntitlementCanView, entity.TypeNetworkZone)
 	if err != nil {
 		return response.InternalError(err)
@@ -224,15 +266,22 @@ func networkZonesGet(d *Daemon, r *http.Request) response.Response {
 
 	resultString := []string{}
 	resultMap := []api.NetworkZone{}
-	for _, zoneName := range zoneNames {
-		if !userHasPermission(entity.NetworkZoneURL(requestProjectName, zoneName)) {
+	for zoneName, projectName := range zoneNamesMap {
+		// Check permission for each network zone against the requested project.
+		if !userHasPermission(entity.NetworkZoneURL(projectName, zoneName)) {
 			continue
 		}
 
 		if !recursion {
 			resultString = append(resultString, api.NewURL().Path(version.APIVersion, "network-zones", zoneName).String())
 		} else {
-			netzone, err := zone.LoadByNameAndProject(s, effectiveProjectName, zoneName)
+			var netzone zone.NetworkZone
+			if !allProjects {
+				netzone, err = zone.LoadByNameAndProject(s, effectiveProjectName, zoneName)
+			} else {
+				netzone, err = zone.LoadByNameAndProject(s, projectName, zoneName)
+			}
+
 			if err != nil {
 				continue
 			}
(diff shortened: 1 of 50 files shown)


markylaing:
@kadinsayani can you rebase please? Also do you still want me to review?

kadinsayani:
> @kadinsayani can you rebase please? Also do you still want me to review?

Sure thing. And yes, please - I'm currently having trouble getting the TLS authorization tests to pass.

kadinsayani:
I'm going to mark this as draft until I have the tests passing, so we can hold off on reviews for now. I've just found this which we may want to include here as well: https://github.com/lxc/incus/pull/1117/commits/9bdeb59150c8b07f479f138860bfd6d8b93b7e67.

cc. @markylaing @tomponline

kadinsayani:
Ready for review @markylaing, thanks!

markylaing:
> Ready for review @markylaing, thanks!

What was the issue in the end?

lxd/network_zones.go
```diff
@@ -196,50 +208,69 @@ func networkZoneAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *h
...
 		return response.SmartError(err)
 	}
 
-	recursion := util.IsRecursionRequest(r)
+	if allProjects && request.QueryParam(r, "project") != "" {
+		return response.BadRequest(errors.New("Cannot specify a project when requesting all projects"))
+	}
 
-	var zoneNames []string
+	recursion := util.IsRecursionRequest(r)
```
markylaing, line 222: This code section is:
1. Performing a database call before performing a validation that can be performed prior to the call.
2. Parsing the request query parameter for "project" twice.
3. Getting the effective project (likely via a superfluous DB call) even when the request is not project specific.

Here's a suggested change which should do the same thing:
```golang
allProjects := shared.IsTrue(request.QueryParam(r, "all-projects"))
requestProjectName := request.QueryParam(r, "project")

if allProjects && requestProjectName != "" {
    return response.BadRequest(errors.New("Cannot specify a project when requesting all projects"))
}

if !allProjects {
    if requestProjectName == "" {
        requestProjectName = api.ProjectDefaultName
    }

    effectiveProjectName, _, err := project.NetworkZoneProject(s.DB.Cluster, requestProjectName)
    if err != nil {
    }

    // If the request is project specific, then set effective project name in the request context so that the authorizer can generate the correct URL.
    request.SetCtxValue(r, request.CtxEffectiveProjectName, effectiveProjectName)
}

```

With this change, you can remove the setting of the effective project name in the request context inside the transaction (which also sticks out as a little odd).

test/suites/network_zone.sh
```diff
@@ -40,6 +40,10 @@ test_network_zone() {
   # Create zone in project.
   lxc network zone create lxdfoo.example.net --project foo
 
+  # Check listing zones from all projects.
+  lxc network zone list --all-projects -fcsv | grep -q default,lxd.example.net || false
+  lxc network zone list --all-projects -fcsv | grep -q foo,lxdfoo.example.net || false
```
markylaing, line 45: Why the `|| false`? This is making it a little confusing to read for me as if the pattern is not found, grep will exit 1. If it is found then the "or" will return anyway. In either case, we don't get to the `false`.

Additionally, can we use `-F` for a fixed string here, and have a space between the format flag and `csv`? Thanks.

markylaing: Also can we remove the `-q` from the grep as I think it's generally preferred that we can see it in test output. Thanks.

markylaing (changes requested):
Just a couple of small nits :)

kadinsayani:
> > Ready for review @markylaing, thanks!
> 
> What was the issue in the end?

We need to check permission on `requestProjectName` for both individual and all project requests, and call the `LoadByNameAndProject` function on `effectiveProjectName`. Also, we were populating the map with `requestProjectName`, when it should have been `effectiveProjectName`.

lxd/network_zones.go
```diff
@@ -196,50 +208,78 @@ func networkZoneAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *h
...
 	s := d.State()
 
-	requestProjectName := request.ProjectParam(r)
-	effectiveProjectName, _, err := project.NetworkZoneProject(s.DB.Cluster, requestProjectName)
-	if err != nil {
-		return response.SmartError(err)
+	allProjects := shared.IsTrue(request.QueryParam(r, "all-projects"))
+	requestProjectName := request.QueryParam(r, "project")
+
+	if allProjects && request.QueryParam(r, "project") != "" {
```
markylaing, line 214: This is already computed two lines above. Can you replace with `requestProjectName` please?

lxd/network_zones.go
```diff
@@ -196,50 +208,78 @@ func networkZoneAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *h
...
 	userHasPermission, err := s.Authorizer.GetPermissionChecker(r.Context(), auth.EntitlementCanView, entity.TypeNetworkZone)
 	if err != nil {
 		return response.InternalError(err)
 	}
 
 	resultString := []string{}
 	resultMap := []api.NetworkZone{}
-	for _, zoneName := range zoneNames {
+	for zoneName, projectName := range zoneNamesMap {
 		if !userHasPermission(entity.NetworkZoneURL(requestProjectName, zoneName)) {
```
markylaing, line 267: This is still not making sense. The `requestProjectName` is only valid for project specific requests.

markylaing: I think this needs to be `projectName`. If the tests fail with this change, let's look in more detail at what the tests are doing and make sure they are correct.

kadinsayani:
> > > > Ready for review @markylaing, thanks!
> > > 
> > > 
> > > What was the issue in the end?
> > 
> > 
> > We need to check permission on `requestProjectName` for both individual and all project requests, and call the `LoadByNameAndProject` function with `effectiveProjectName` passed in. Also, we were populating the map with `requestProjectName`, when it should have been `effectiveProjectName`.
> 
> I'm confused, it doesn't make sense to check permission on `requestProjectName` for an `all-projects` request since the request is not project specific. Auth checks should use the project that the network zone is defined in for all-projects.

You're right, we should check permission for `effectiveProjectName`. I believe the issue we were running into during our meeting was `requestProjectName` being empty for default project requests when we switched from `request.ProjectParam(...)` to `request.QueryParam(...)`, which we have resolved by adding a check and setting it.

lxd/network_zones.go
kadinsayani: It appears that the TLS restrictions test suite fails with this change - I’ll look into this further.

kadinsayani: Still failing, here's where the failure occurs:

https://github.com/canonical/lxd/blob/da43979dab584872e6df2362834deb7fbd4187c3/test/suites/tls_restrictions.sh#L185-L187

I don't think it makes sense for the network zone to be visible in the project `blah` if it was created in the `default` project in this case since we are testing access for a restricted user.

markylaing: It does make sense in this case. The project `blah` does not have `features.networks.zones` enabled, so it uses the network zones from the default project.

markylaing: It does make sense in this case. The project `blah` does not have `features.networks.zones` enabled, so it uses the network zones from the default project.

markylaing: It does make sense in this case. The project `blah` does not have `features.networks.zones` enabled, so it uses the network zones from the default project.

kadinsayani: It looks like the initial value for `features.networks.zones` is set to `false`.

kadinsayani: Looks like we commented at the same time :)

markylaing:
@kadinsayani The issue in this case is that we are setting the `requestProjectName` in the map of network zone to project name when getting network zones for a single project, this is required for the authorization check, but causes `zone.LoadByNameAndProject` to fail silently later (because it just has a `continue` on error).

Please can you update the calls to `zone.LoadByNameAndProject` to use the effective project name (only when allProjects is false). This should fix it.

Also, can you please investigate why there is a `continue` if a network zone fails to load? This seems very unusual, so if there is a good reason for it we should have a comment explaining it.

kadinsayani:
> @kadinsayani The issue in this case is that we are setting the `requestProjectName` in the map of network zone to project name when getting network zones for a single project, this is required for the authorization check, but causes `zone.LoadByNameAndProject` to fail silently later (because it just has a `continue` on error).
> 
> Please can you update the calls to `zone.LoadByNameAndProject` to use the effective project name (only when allProjects is false). This should fix it.
> 
> Also, can you please investigate why there is a `continue` if a network zone fails to load? This seems very unusual, so if there is a good reason for it we should have a comment explaining it.

Thanks - I've updated the calls to `zone.LoadByNameAndProject` and added a comment for the `continue` when a network zone fails to load. We don't want to return an error when a network zone fails to load because a restricted client may be able to use specific network zones specified by `restricted.networks.zones`. Does that sound accurate?

kadinsayani:
> @kadinsayani The issue in this case is that we are setting the `requestProjectName` in the map of network zone to project name when getting network zones for a single project, this is required for the authorization check, but causes `zone.LoadByNameAndProject` to fail silently later (because it just has a `continue` on error).
> 
> Please can you update the calls to `zone.LoadByNameAndProject` to use the effective project name (only when allProjects is false). This should fix it.
> 
> Also, can you please investigate why there is a `continue` if a network zone fails to load? This seems very unusual, so if there is a good reason for it we should have a comment explaining it.

Thanks - I've updated the calls to `zone.LoadByNameAndProject`. After further investigation, since we have already checked permission for a client, I don't think we should `continue` when a network zone fails to load, but rather return an error. This will allow us to remain consistent with database transaction failures at other endpoints which also return errors.

lxd/network_zones.go
```diff
@@ -196,50 +208,84 @@ func networkZoneAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *h
...
+			var netzone zone.NetworkZone
+			if !allProjects {
+				netzone, err = zone.LoadByNameAndProject(s, effectiveProjectName, zoneName)
+			} else {
+				netzone, err = zone.LoadByNameAndProject(s, projectName, zoneName)
+			}
+
 			if err != nil {
-				continue
+				return response.SmartError(err)
```
markylaing, line 282: I think it might be worth adding this (change from `continue` to returning an error) as a separate commit. This is a change in behaviour for the endpoint when getting network zones within a project as well, and not necessarily related to adding `all-projects` functionality.

test/suites/network_zone.sh
```diff
@@ -40,6 +40,10 @@ test_network_zone() {
   # Create zone in project.
   lxc network zone create lxdfoo.example.net --project foo
 
+  # Check listing zones from all projects.
+  lxc network zone list --all-projects -f csv | grep -F default,lxd.example.net
+  lxc network zone list --all-projects -f csv | grep -F foo,lxdfoo.example.net
```
markylaing, line 45: Could you please wrap the grep string in single quotes? Thanks.

test/suites/auth.sh
```diff
@@ -628,6 +628,7 @@ user_is_not_project_operator() {
   # Should not be able to see or create network zones.
   lxc_remote network zone create zone1
   [ "$(lxc_remote network zone list "${remote}:" -f csv | wc -l)" = 0 ]
+  [ "$(lxc_remote network zone list "${remote}:" -f csv --all-projects | wc -l)" = 0 ]
```
markylaing, line 631: This one I think can remove the `wc -l` and equate to an empty string, as below. (The one above can be changed also, perhaps you can come back to that in a separate PR?)

markylaing (commented):
Just a couple of small nits and then we're good to go :)

kadinsayani:
Should be good to go, thanks :)

@markylaing @tomponline

markylaing (dismissed):
LGTM, thanks!

tomponline:
Please can you rebase

kadinsayani:
> Please can you rebase

Rebased ✅

minaelee (dismissed):

lxd/network_zones.go
tomponline, line 1: Please can you add some comments around the non-obvious project logic being used in this commit.
Specifically the differences between requested project, effective project and the zone's project when using the all projects mode, it wasn't clear why each was being used.

tomponline (changes requested):
Do the tests we have here cover the effective project scenarios that were making the permission checks for complicated?

lxd/network_zones.go
kadinsayani: Sure thing!

kadinsayani:
> Do the tests we have here cover the effective project scenarios that were making the permission checks for complicated?

Yes, the tests for fetching network zones across all projects in the auth test suite cover those cases.

minaelee (approved):

tomponline:
Please rebase to include @minaelee doc fix ta

kadinsayani:
> Please rebase to include @minaelee doc fix ta

Rebased ✅

tomponline (approved):

PR #14586, opened 2024-12-05
Fix detection of suitable architecture for conversion when LXD is clustered

MusicDin:
Thanks to @escabo for reporting the issue where conversion fails if LXD is clustered and supports API extension `instance_import_conversion`.

Not sure how this slipped through for such a long time. Seems I have messed something up when juggling with commit order on initial implementation, as I remember configuring this.

Anyway, the issue is detection of suitable architecture, where source type was not matched with any case resulting in an error:
```
Error: Unknown instance source type "conversion"
```

### DIFF
--- a/lxd/instance/instance_utils.go
+++ b/lxd/instance/instance_utils.go
@@ -547,7 +547,7 @@ func ResolveImage(ctx context.Context, tx *db.ClusterTx, projectName string, sou
 // A nil list indicates that we can't tell at this stage, typically for private images.
 func SuitableArchitectures(ctx context.Context, s *state.State, tx *db.ClusterTx, projectName string, sourceInst *cluster.Instance, sourceImageRef string, req api.InstancesPost) ([]int, error) {
 	// Handle cases where the architecture is already provided.
-	if shared.ValueInSlice(req.Source.Type, []string{"migration", "none"}) && req.Architecture != "" {
+	if shared.ValueInSlice(req.Source.Type, []string{"conversion", "migration", "none"}) && req.Architecture != "" {
 		id, err := osarch.ArchitectureId(req.Architecture)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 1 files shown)


lxd/instance/instance_utils.go
```diff
@@ -547,7 +547,7 @@ func ResolveImage(ctx context.Context, tx *db.ClusterTx, projectName string, sou
 // A nil list indicates that we can't tell at this stage, typically for private images.
 func SuitableArchitectures(ctx context.Context, s *state.State, tx *db.ClusterTx, projectName string, sourceInst *cluster.Instance, sourceImageRef string, req api.InstancesPost) ([]int, error) {
 	// Handle cases where the architecture is already provided.
-	if shared.ValueInSlice(req.Source.Type, []string{"migration", "none"}) && req.Architecture != "" {
+	if shared.ValueInSlice(req.Source.Type, []string{"conversion", "migration", "none"}) && req.Architecture != "" {
```
tomponline, line 550: As a separate PR please can we have constants created for these values so we can see where they are referenced in the code.

tomponline (approved):

tomponline:
@MusicDin @escabo why was architecture not populated btw?

tomponline:
@MusicDin once merged please can you prepare a cherry-pick against lxd-pkg-snap latest-candidate branch to include this commit in 6.2

MusicDin:
Architecture was populated, but `source.Type=conversion` did not match any case in the function, reaching the end, where an error `Error: Unknown instance source type "conversion"` is returned.

tomponline:
> Architecture was populated, but `source.Type=conversion` did not match any case in the function, reaching the end, where an error `Error: Unknown instance source type "conversion"` is returned.

why was clustering relevant?

MusicDin:
> why was clustering relevant?

When target server is not provided the architecture is used to decide upon suitable cluster member to run the new instance.

https://github.com/canonical/lxd/blob/6d29cd671b5653441272aeff11c4f4eb18d1887d/lxd/instances_post.go#L1227-L1259

PR #14587, opened 2024-12-05
Introduce source type constants

MusicDin:
Replace hardcoded strings for container/instance source type.

Also fixed some issue reported by static analysis.

### DIFF
--- a/client/lxd.go
+++ b/client/lxd.go
@@ -221,7 +221,7 @@ func lxdParseResponse(resp *http.Response) (*api.Response, string, error) {
 
 	// Handle errors
 	if response.Type == api.ErrorResponse {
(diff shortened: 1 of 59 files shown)


client/lxd.go
```diff
@@ -221,7 +221,7 @@ func lxdParseResponse(resp *http.Response) (*api.Response, string, error) {
 
 	// Handle errors
 	if response.Type == api.ErrorResponse {
-		return nil, "", api.StatusErrorf(resp.StatusCode, response.Error)
+		return nil, "", api.StatusErrorf(resp.StatusCode, "%s", response.Error)
```
tomponline, line 224: @markylaing added api.NewStatusError for this scenario to avoid needing the parsing cost of fmt.Errorf

tomponline (approved):
Nice!

PR #14593, opened 2024-12-05
lxc: Accept volume full name on `detach`

hamistao:
The completions suggest the volume name but the client doesn't parse it correctly and this has been annoying me, so I took some of my time off to fix it and also improved the completions for detach/attach commands. Along with other small improvements.

### DIFF
--- a/lxc/completion.go
+++ b/lxc/completion.go
@@ -1549,7 +1549,8 @@ func (g *cmdGlobal) cmpStoragePoolVolumeSnapshots(poolName string, volumeName st
 
 // cmpStoragePoolVolumes provides shell completion for storage pool volumes.
 // It takes a storage pool name and returns a list of storage pool volumes along with a shell completion directive.
-func (g *cmdGlobal) cmpStoragePoolVolumes(poolName string) ([]string, cobra.ShellCompDirective) {
+// Parameter volumeTypes determines which types of volumes are valid as completion options, none being provided means all types are valid.
+func (g *cmdGlobal) cmpStoragePoolVolumes(poolName string, volumeTypes ...string) ([]string, cobra.ShellCompDirective) {
 	// Parse remote
 	resources, err := g.ParseServers(poolName)
 	if err != nil || len(resources) == 0 {
@@ -1569,5 +1570,29 @@ func (g *cmdGlobal) cmpStoragePoolVolumes(poolName string) ([]string, cobra.Shel
 		return nil, cobra.ShellCompDirectiveError
 	}
 
-	return volumes, cobra.ShellCompDirectiveNoFileComp
+	// If no volume type is provided, don't filter on type.
+	if len(volumeTypes) == 0 {
+		return volumes, cobra.ShellCompDirectiveNoFileComp
+	}
+
+	// Pre-allocate slice capacity.
+	customKeyCount := 0
+	for _, volume := range volumes {
+		_, volType := parseVolume("custom", volume)
+		if shared.ValueInSlice(volType, volumeTypes) {
+			customKeyCount++
+		}
+	}
+
+	customVolumeNames := make([]string, 0, customKeyCount)
+
+	// Only include custom volumes
+	for _, volume := range volumes {
+		_, volType := parseVolume("custom", volume)
+		if shared.ValueInSlice(volType, volumeTypes) {
+			customVolumeNames = append(customVolumeNames, volume)
+		}
+	}
+
+	return customVolumeNames, cobra.ShellCompDirectiveNoFileComp
 }
--- a/lxc/completion.go
+++ b/lxc/completion.go
@@ -1368,10 +1368,8 @@ func (g *cmdGlobal) cmpStoragePoolWithVolume(toComplete string) ([]string, cobra
 
 	var results []string
 	for _, volume := range volumes {
-		volName, volType := parseVolume("custom", volume)
-		if volType == "custom" {
-			results = append(results, pool+"/"+volName)
-		}
+		volName, _ := parseVolume("custom", volume)
+		results = append(results, pool+"/"+volName)
 	}
 
 	return results, cobra.ShellCompDirectiveNoFileComp
--- a/lxc/storage_volume.go
+++ b/lxc/storage_volume.go
@@ -236,6 +236,7 @@ func (c *cmdStorageVolumeAttach) run(cmd *cobra.Command, args []string) error {
 		case "block", "iso":
 			devName = args[3]
 		case "filesystem":
+			// If using a filesystem volume, the path must also be provided as the fourth argument.
 			if !strings.HasPrefix(args[3], "/") {
 				devPath = path.Join("/", args[3])
 			} else {
--- a/lxc/storage_volume.go
+++ b/lxc/storage_volume.go
@@ -177,7 +177,7 @@ func (c *cmdStorageVolumeAttach) command() *cobra.Command {
 		}
 
 		if len(args) == 1 {
-			return c.global.cmpStoragePoolVolumes(args[0])
+			return c.global.cmpStoragePoolVolumes(args[0], "custom")
 		}
 
 		if len(args) == 2 {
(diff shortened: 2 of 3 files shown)


lxc/storage_volume.go
```diff
@@ -323,26 +324,26 @@ func (c *cmdStorageVolumeAttachProfile) run(cmd *cobra.Command, args []string) e
 		return errors.New(i18n.G("Missing pool name"))
 	}
 
+	volName, volType := parseVolume("custom", args[1])
+	if volType != "custom" {
+		return errors.New(i18n.G("Only \"custom\" volumes can be attached to instances"))
```
tomponline, line 329: lets use backticks here to save needing the escaping

lxc/storage_volume.go
```diff
@@ -859,10 +860,15 @@ func (c *cmdStorageVolumeDetach) run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	volName, volType := parseVolume("custom", args[1])
+	if volType != "custom" {
+		return errors.New(i18n.G("Only \"custom\" volumes can be attached to instances"))
```
tomponline, line 865: backticks for the outer string please

hamistao: Done!

tomponline:
tests failing, converted to draft

kadinsayani:
Can we please update `cmpStoragePoolWithVolume` and references to it as well?

lxc/storage_volume.go
```diff
@@ -236,13 +236,14 @@ func (c *cmdStorageVolumeAttach) run(cmd *cobra.Command, args []string) error {
 		case "block", "iso":
 			devName = args[3]
 		case "filesystem":
+			// If using a filesystem volume, the path must also be provided as the forth argument.
```
kadinsayani, line 239:
```suggestion
			// If using a filesystem volume, the path must also be provided as the fourth argument.
```

kadinsayani (dismissed):
Just a few small nits - thanks :)

lxc/storage_volume.go
hamistao: Thanks for catching this :)

lxc/completion.go
```diff
@@ -1569,5 +1567,19 @@ func (g *cmdGlobal) cmpStoragePoolVolumes(poolName string) ([]string, cobra.Shel
 		return nil, cobra.ShellCompDirectiveError
 	}
 
-	return volumes, cobra.ShellCompDirectiveNoFileComp
+	if !customOnly {
+		return volumes, cobra.ShellCompDirectiveNoFileComp
+	}
+
+	var results []string
```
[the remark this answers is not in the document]
hamistao: We can indeed, but don't you think this seems like over-complicating the algorithm for little benefit?

hamistao:
> Can we please update cmpStoragePoolWithVolume and references to it as well?

@kadinsayani Sorry I am confused by this, why would it need ot be updated on this PR?

lxc/completion.go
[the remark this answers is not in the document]
kadinsayani: Probably, but I've been pre-allocating slices elsewhere in `lxc/completion.go` so thought I would suggest for consistency :)

kadinsayani:
> > Can we please update cmpStoragePoolWithVolume and references to it as well?
> 
> @kadinsayani Sorry I am confused by this, why would it need ot be updated on this PR?

Sorry, I was confused by the function naming - we don't need to update `cmpStoragePoolWithVolume`.

lxc/completion.go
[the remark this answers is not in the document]
hamistao: sure then, all fine for me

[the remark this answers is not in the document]
kadinsayani: Thanks!

hamistao:
@MggMuggins I happened to parse the volume name on `detach` similarly to what you have done on https://github.com/canonical/lxd/pull/14532. Since this PR is pretty simple and straightforward overall, I suggest merging this as is and you can soon rebase your PR afterwards, what do you think?

lxc/storage_volume.go
```diff
@@ -218,7 +218,7 @@ func (c *cmdStorageVolumeAttach) run(cmd *cobra.Command, args []string) error {
 	devPath := ""
 	devName := ""
 	if len(args) == 3 {
-		devName = args[1]
+		devName = volName
```
MggMuggins, line 221: I think I disagree with this commit/change; as the `note` in the docs [here](https://documentation.ubuntu.com/lxd/en/latest/howto/storage_volumes/#view-storage-volumes) points out, custom storage volumes and instances can have the same name. After #14532 is merged, `lxc storage volume attach/detach` would generate the same device name for `virtual-machine/vm1` and `custom/vm1`. I don't like introducing that possibility if we can avoid it.

lxc/completion.go
```diff
@@ -1549,7 +1549,7 @@ func (g *cmdGlobal) cmpStoragePoolVolumeSnapshots(poolName string, volumeName st
 
 // cmpStoragePoolVolumes provides shell completion for storage pool volumes.
 // It takes a storage pool name and returns a list of storage pool volumes along with a shell completion directive.
-func (g *cmdGlobal) cmpStoragePoolVolumes(poolName string) ([]string, cobra.ShellCompDirective) {
+func (g *cmdGlobal) cmpStoragePoolVolumes(poolName string, customOnly bool) ([]string, cobra.ShellCompDirective) {
```
MggMuggins, line 1550: A boolean param that's used as a constant everywhere the function is called is a code smell; can you refactor this into `cmpStoragePoolVolumes` and `cmpStoragePoolCustomVolumes` (or similar)?

E.g. when I run into this in the wild, I have zero context for what `true` means here. There's no connection at the call site between the parameter's value and its meaning.
```go
volumes, compdir := g.cmpStoragePoolVolumes(pool, true)
```

MggMuggins (changes requested):
With the exception of 8128f85, I am happy to rebase over this.

lxc/completion.go
hamistao: I get your point, but this would result in a lot of code duplication, which is something we are always trying to avoid. I can extend the function comment and then it would be clear what 'true' means just by looking at the comment.

lxc/storage_volume.go
hamistao: You make a fine argument, dropping this commit

hamistao:
All comments were addressed @kadinsayani @MggMuggins @tomponline

lxc/completion.go
tomponline, line 1: Please can you expand the commit message to explain what this new flag does and why its needed.
Hard to review otherwise as im not sure of your intentions vs the actual code.

lxc/completion.go
tomponline: can we change this to a variadic argument that accepts one or more volume types to include?

test/suites/storage.sh
tomponline, line 1: Please can we move away from "improve xyx" type commit messages, as the commit message should explain what is specifically changing. Thanks

lxc/storage_volume.go
```diff
@@ -859,10 +859,15 @@ func (c *cmdStorageVolumeDetach) run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	volName, volType := parseVolume("custom", args[1])
+	if volType != "custom" {
+		return errors.New(i18n.G(`Only "custom" volumes can be attached to instances`))
```
tomponline, line 865: @MggMuggins is this going to need to change in your PR that adds support for attaching root disks to instances?

hamistao: Yup, he has taken a look at these changes. I believe these checks will also consider `virtual-machine` volumes.

MggMuggins: Yes, this should conflict with my PR; I'm happy to rebase.

test/suites/storage.sh
hamistao: Since this commit has a few very small changes I instead extended the commit message getting a bit more into detail of what it introduces.

lxc/completion.go
hamistao: Included some comments elaborating on the new parameter

tomponline (approved):
Thanks!

tomponline:
@MggMuggins please feel free to make changes to this in your later PR that feel appropriate when being able to attached root disks too. Ta

PR #14594, opened 2024-12-06
Storage: Define per-pool default block size

hamistao:
This includes some improvements that were originally a part of the fix implemented in https://github.com/canonical/lxd/pull/14511, but ended up not being needed for that fix. They are still nice to have and thus were included here.

### DIFF
--- a/lxd/scriptlet/instance_placement.go
+++ b/lxd/scriptlet/instance_placement.go
@@ -15,7 +15,7 @@ import (
 	"github.com/canonical/lxd/lxd/resources"
 	scriptletLoad "github.com/canonical/lxd/lxd/scriptlet/load"
 	"github.com/canonical/lxd/lxd/state"
-	storageDrivers "github.com/canonical/lxd/lxd/storage/drivers"
+	"github.com/canonical/lxd/lxd/storage"
 	"github.com/canonical/lxd/shared/api"
 	apiScriptlet "github.com/canonical/lxd/shared/api/scriptlet"
 	"github.com/canonical/lxd/shared/logger"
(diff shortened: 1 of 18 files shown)


lxd/scriptlet/instance_placement.go
```diff
@@ -221,7 +221,12 @@ func InstancePlacementRun(ctx context.Context, l logger.Logger, s *state.State,
 
 			// Apply VM root disk size defaults if not specified.
 			if req.Type == api.InstanceTypeVM && rootDiskSizeStr == "" {
-				rootDiskSizeStr = storageDrivers.DefaultBlockSize
+				driver, err := storageDrivers.Load(s, rootDiskConfig["pool"], "", nil, nil, nil, nil)
```
tomponline, line 224: please can you add a function like `storageDrivers.DefaultVMBlockFilesystemSize` for this.

lxd/storage/drivers/driver_powerflex_volumes.go
```diff
@@ -532,7 +532,7 @@ func (d *powerflex) GetVolumeUsage(vol Volume) (int64, error) {
 
 	// Getting the usage of an unmounted volume is not supported.
 	// PowerFlex reports the usage on pool level only.
-	return 0, ErrNotSupported
+	return -1, ErrNotSupported
```
tomponline, line 535: @hamistao is this the current (i.e whats in main now) standard?

lxd/storage/utils.go
```diff
@@ -1377,3 +1377,13 @@ func ValidVolumeName(volumeName string) error {
 
 	return nil
 }
+
+// GetPoolDefaultBlockSize returns the default block size for the specified storage pool according to its driver.
+func GetPoolDefaultBlockSize(s *state.State, poolName string) (string, error) {
```
tomponline, line 1382: This isn't quite the same as `DefaultVMBlockFilesystemSize` as it has a state argument.
Why is that?
Can we not use the same pattern as `DefaultVMBlockFilesystemSize`?

lxd/storage/drivers/driver_powerflex_volumes.go
hamistao: Yes, it is!

lxd/storage/utils.go
hamistao: We can use the same pattern as `DefaultVMBlockFilesystemSize`, but `GetPoolDefaultBlockSize` aims to get the default block size using the pool name, not the driver name. So we need to get the pool data from the database to know what its driver is. And this is like this because this is what we need for `instance_placement`.

Alternatively, I can create a function like `DefaultVMBlockFilesystemSize` and get the pool info before calling it. Hope this was clear.

tomponline: understood thanks, this is fine

tomponline (approved):

Issue #14349, opened 2024-10-26
doc: Missing documentation for quota management


rptb1:
[Reference / Storage Drivers / ZFS / Quotas](https://documentation.ubuntu.com/lxd/en/latest/reference/storage_zfs/#quotas) says:

> By default, LXD uses the quota property when you set up a quota for your storage volume.

But [How to manage storage volumes](https://documentation.ubuntu.com/lxd/en/latest/howto/storage_volumes/#how-to-manage-storage-volumes) makes no mention of quotas.  (It would be nice if the phrase "set up a quota" and all similar phrases were linked, btw.)

A [search of the documentation for quota](https://documentation.ubuntu.com/lxd/en/latest/search/?q=quota&check_keywords=yes&area=default) does not reveal anything.

I'm guessing there are secret configuration keywords on containers and storage volumes that controls quotas.  We have [zfs.use_refquota](https://documentation.ubuntu.com/lxd/en/latest/reference/storage_zfs/#storage-zfs-volume-conf:zfs.use_refquota) that says:

> Use refquota instead of quota for space

but again this does not link to how to actually use set or manage a "quota for space".

The only clue I've found is [in an *example*](https://documentation.ubuntu.com/lxd/en/latest/reference/manpages/lxc/storage/volume/unset/#examples):

>   lxc storage volume unset default data size
>       Remotes the size/quota of a custom volume "data" in pool "default".

All of the above applies for instances and profiles as well as "storage volumes".  For example [How to manage instances](https://documentation.ubuntu.com/lxd/en/latest/howto/instances_manage/) does not mention quotas.

Please could we have some documentation for size quotas.

Thanks.

---
Document: reference/storage_zfs.md

minaelee:
https://github.com/canonical/lxd/pull/14604 addresses the issue of not finding mentions of quotas in the documentation. Wherever `size` is mentioned in regard to storage, the word `quota` is mentioned alongside it at least once.

PR #14595, opened 2024-12-06
Storage: Show correct disk size

hamistao:
This is a follow up PR for https://github.com/canonical/lxd/pull/14511 and this aims to address the problems outlined in https://github.com/canonical/lxd/pull/14511#issuecomment-2502356227.

For this, we using the `volatile.rootfs.size` config key, previously only available for images, for instance volumes. This config key is used to represent the current total size of an instance volume at any point in time. This way we can determine whether a volume had its quota set based on `volume.size` or not, and also determine if has been resized after changing the disk's `size` property, as some instances are not live-resizable.

This key was chosen instead of the `size` key because the intended behavior of this config key fits better the pattern we see with `volatile.*` keys, that can be changed by LXD itself, without the user specifying them.

### DIFF
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -250,10 +250,14 @@ func (b *lxdBackend) Create(clientType request.ClientType, op *operations.Operat
 }
 
 // GetNewVolume returns a drivers.Volume that doesn't yet exist in the database.
-// It contains copies of the supplied volume config, including a new UUID, and the pools config.
+// It contains copies of the supplied volume config, including a new UUID and
+// default configuration for the poo driver, as well as the pool's config.
 // Use the returned drivers.Volume as the base for actions performed on the new volume.
 func (b *lxdBackend) GetNewVolume(volType drivers.VolumeType, contentType drivers.ContentType, volName string, volConfig map[string]string) drivers.Volume {
 	newVol := b.GetVolume(volType, contentType, volName, volConfig)
+
+	// Fill default config.
+	b.Driver().FillVolumeConfig(newVol)
 
 	// Set a new UUID.
 	newVol.Config()["volatile.uuid"] = uuid.New().String()
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -915,9 +915,12 @@ func (b *lxdBackend) CreateInstanceFromBackup(srcBackup backup.Info, srcData io.
 
 			newSnapshotName := drivers.GetSnapshotVolumeName(inst.Name(), backupFileSnap)
 
+			// Ensure driver specific config is filled in new volume.
+			snapVol := b.GetNewVolume(volType, contentType, newSnapshotName, volumeSnapConfig)
+
 			// Validate config and create database entry for new storage volume.
 			// Strip unsupported config keys (in case the export was made from a different type of storage pool).
-			err = VolumeDBCreate(b, inst.Project().Name, newSnapshotName, volumeSnapDescription, volType, true, volumeSnapConfig, volumeSnapCreationDate, volumeSnapExpiryDate, contentType, true, true)
+			err = VolumeDBCreate(b, inst.Project().Name, newSnapshotName, volumeSnapDescription, volType, true, snapVol.Config(), volumeSnapCreationDate, volumeSnapExpiryDate, contentType, true, true)
 			if err != nil {
 				return err
 			}
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -874,6 +874,14 @@ func (b *lxdBackend) CreateInstanceFromBackup(srcBackup backup.Info, srcData io.
 			}
 		}
 
+		// Generate the effective root device volume for instance.
+		volStorageName := project.Instance(inst.Project().Name, inst.Name())
+		vol := b.GetVolume(volType, contentType, volStorageName, volumeConfig)
+		err = b.applyInstanceRootDiskOverrides(inst, &vol)
+		if err != nil {
+			return err
+		}
+
 		// Validate config and create database entry for new storage volume.
 		// Strip unsupported config keys (in case the export was made from a different type of storage pool).
 		err = VolumeDBCreate(b, inst.Project().Name, inst.Name(), volumeDescription, volType, false, volumeConfig, volumeCreationDate, time.Time{}, contentType, true, true)
@@ -926,14 +934,6 @@ func (b *lxdBackend) CreateInstanceFromBackup(srcBackup backup.Info, srcData io.
 			}
 
 			postHookRevert.Add(func() { _ = VolumeDBDelete(b, inst.Project().Name, newSnapshotName, volType) })
-		}
-
-		// Generate the effective root device volume for instance.
-		volStorageName := project.Instance(inst.Project().Name, inst.Name())
-		vol := b.GetVolume(volType, contentType, volStorageName, volumeConfig)
-		err = b.applyInstanceRootDiskOverrides(inst, &vol)
-		if err != nil {
-			return err
 		}
 
 		// Save any changes that have occurred to the instance's config to the on-disk backup.yaml file.
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -2241,6 +2241,12 @@ func (b *lxdBackend) CreateInstanceFromMigration(inst instance.Instance, conn io
 	revert := revert.New()
 	defer revert.Fail()
 
+	// Generate the effective root device volume for instance.
+	err = b.applyInstanceRootDiskOverrides(inst, &vol)
+	if err != nil {
+		return err
+	}
+
 	if !args.Refresh {
 		if volExists {
 			if !isRemoteClusterMove {
@@ -2301,12 +2307,6 @@ func (b *lxdBackend) CreateInstanceFromMigration(inst instance.Instance, conn io
 
 			revert.Add(func() { _ = VolumeDBDelete(b, inst.Project().Name, newSnapshotName, volType) })
 		}
-	}
-
-	// Generate the effective root device volume for instance.
-	err = b.applyInstanceRootDiskOverrides(inst, &vol)
-	if err != nil {
-		return err
 	}
 
 	// Override args.Name and args.Config to ensure volume is created based on instance.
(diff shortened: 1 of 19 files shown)


tomponline:
@minaelee @MggMuggins shall we try and address the concerns in https://github.com/canonical/lxd/issues/14349 and mention how size is equivalent to quota?

doc/api-extensions.md
```diff
@@ -2535,3 +2535,8 @@ Adds a new {config:option}`device-unix-hotplug-device-conf:ownership.inherit` co
 ## `unix_device_hotplug_subsystem_device_option`
 
 Adds a new {config:option}`device-unix-hotplug-device-conf:subsystem` configuration option for `unix-hotplug` devices. This adds support for detecting `unix-hotplug` devices by subsystem, and can be used in conjunction with {config:option}`device-unix-hotplug-device-conf:productid` and {config:option}`device-unix-hotplug-device-conf:vendorid`.
+
+## `storage_volume_rootfs_size`
+
+Introduces a `volatile.rootfs.size` key, previously only supported for images, for container and virtual machine volumes that is used to store that volume's current size at any point.
```
minaelee, line 2541: Should this be showing up in one of the reference documents that are 
```suggestion
Introduces a `volatile.rootfs.size` key, previously only supported for images, for container and virtual machine volumes. This key is used to store a volume's current size at any point.
```

minaelee: Clarifying that the clause following "virtual machine volumes that ..." is meant to refer to `key` and not `volumes`. 

Question: Should the `volatile.rootfs.size` key show up in a reference page somewhere? I could not find it in the PR's RTD build. For example, should it show up [here](https://canonical-ubuntu-documentation-library--14595.com.readthedocs.build/lxd/en/14595/config-options/)? Or in the storage volume configuration options like [here](https://canonical-ubuntu-documentation-library--14595.com.readthedocs.build/lxd/en/14595/reference/storage_zfs/)?

doc/rest-api.yaml
```diff
@@ -2516,13 +2516,13 @@ definitions:
...
             total:
-                description: Total size in bytes
+                description: Total size in bytes. We use -1 to convey that the instance has access to the entire pool's storage.
                 example: 502239232
                 format: int64
                 type: integer
                 x-go-name: Total
             usage:
-                description: Disk usage in bytes
+                description: Disk usage in bytes. We use -1 to indicate the storage driver for the instance's pool does not support getting the disk usage.
```
minaelee, line 2525:
```suggestion
                description: Disk usage in bytes. We use -1 to indicate that the storage driver for the instance's pool does not support retrieving the disk usage.
```

lxd/storage/backend_lxd.go
```diff
@@ -3287,12 +3302,20 @@ func (b *lxdBackend) GetInstanceUsage(inst instance.Instance) (*VolumeUsage, err
 		}
 	}
 
+	// If the instance volume is not block based/typed and neither bound by the device's size config key,
```
minaelee, line 3305:
```suggestion
	// If the instance volume is neither block based/typed nor bound by the device's size config key,
```

doc/api-extensions.md
hamistao: > Clarifying that the clause following "virtual machine volumes that ..." is meant to refer to key and not volumes.

Yes, it is. Now I see how that is confusing,

> Question: Should the volatile.rootfs.size key show up in a reference page somewhere?

I always assumed it was a pattern to not include `volatile.*` keys in the docs as they are not intended (I think) to be set by users. Although for some reason it is apparentely allowed.
There seems to be a lack of pattern, the only `volatile.*` keys that are included are defined [here](https://github.com/hamistao/lxd/blob/disk_size_fixes/lxd/storage/utils.go#L495) and here(https://github.com/hamistao/lxd/blob/disk_size_fixes/lxd/instance/instancetype/instance.go#L111). Everywhere else they are do not follow a metadata doc comment and I do not see a pattern for why some of them are included and some aren't.

tomponline:
> For this, we are introducing the `volatile.rootfs.size` config key for instance volumes

Is this only for container filesystem based volumes or for VM block volumes too?

Is this key used for VM image volumes too?

hamistao:
> Is this only for container filesystem based volumes or for VM block volumes too?

This is for all instance types, since all can be restrained by setting the device's `size` key. 

> Is this key used for VM image volumes too?

Yes, it is currently only used for image volumes.

tomponline:
> Yes, it is currently only used for image volumes.

thats not what i asked, is it currently populated for VM image volumes as well as container image volumes?

hamistao:
> thats not what i asked, is it currently populated for VM image volumes as well as container image volumes?

Oh sorry, I misread your question. Yes, it is used for both container and VM images.

tomponline:
@hamistao as discussed we will revert/continue to return `0` for volume size for both instance state and volume state when a volume is unbounded, rather than `-1`.

doc/api-extensions.md
tomponline: @hamistao please can you log a separate GH issue to review this question and allow us to decide whether to document volatile keys for all entity types. Thanks

tomponline:
@hamistao there's a comment in `validateVolumeCommonRules` that says:

> // volatile.rootfs.size is only used for image volumes.

Does this need to be updated/considered/changed?

tomponline:
@hamistao please can you explore whether we can prevent volatile.rootfs.size being editable for volumes that have it set, as a separate PR.

hamistao:
We have reached a crossroads when it comes to what value to use to represent the size of unbounded (sizeless) volumes.

Previously, we got size 0 in every scenario except when a user explicitely set the volume size, either through `lxc config device set c1 root size=10GiB`, for instance volumes, or `lxc storage volume set pool custom_vol size 10GiB`. Besides disregarding the size that block-based volumes need to have, this also doesn't consider the fact that block-based instances are not live-resizable, and such getting a non-zero could also be innacurate. Also, when getting usage was not supported, the API used to show an empty state struct for the instance disk, so no value was designed for usage nor total.

Mind that these 0's were never exposed though CLI command outputs, so `lxc storage volume info pool volume`, `lxc info instance` and `lxc config show instance` would simply abstain from mentioning the disk/volume total size.

Since https://github.com/canonical/lxd/pull/14511, we now show -1 for the total size of sizeless instances (not custom volumes yet) and -1 for usage when it is not supported. We also now show the default block size for block-based instances (10GiB). Custom volumes still behave the same as they did previously.

So now we have two possible ways to think of the size of sizeless volumes.
A) The 0 size for sizeless volumes we got previously was not a bug, but only the innacurate reporting of the size. So we should change it back to 0.
B) We never should show 0 as a size for a volume since the correct way to represent no size is by using -1. So showing 0 was never correct. So we should fix custom volumes accordingly.

One one hand, option A implies we should be "changing less" and also takes into account the possibility of some users relying on this 0 to check wether the instance has a root disk size set or not. It is also more flexible in case we want to make it a uunsigned integet in the future.
On the other, B implies we should be "changing more" and using -1. This is more consisten with our API. As examples, we have cpu usage and project limits defaulting to -1 on the API level. Also, although changing to -1 can be seen as an API break, the disk size reported by the API was extremely unreliable, so if one was almost always getting 0 size, it is hard to say that this is meaningful/useful information to begin with. And this is why I am more inclined to go with B and keep it as -1.

I hope this was clear, if any info is missing we can discuss it in our 1-1 later, cc @tomponline

tomponline:
> We have reached a crossroads when it comes to what value to use to represent the size of unbounded (sizeless) volumes.
> 
> Previously, we got size 0 in every scenario except when a user explicitely set the volume size, either through `lxc config device set c1 root size=10GiB`, for instance volumes, or `lxc storage volume set pool custom_vol size 10GiB`. Besides disregarding the size that block-based volumes need to have, this also doesn't consider the fact that block-based instances are not live-resizable, and such getting a non-zero could also be innacurate. Also, when getting usage was not supported, the API used to show an empty state struct for the instance disk, so no value was designed for usage nor total.
> 
> Mind that these 0's were never exposed though CLI command outputs, so `lxc storage volume info pool volume`, `lxc info instance` and `lxc config show instance` would simply abstain from mentioning the disk/volume total size.
> 
> Since #14511, we now show -1 for the total size of sizeless instances (not custom volumes yet) and -1 for usage when it is not supported. We also now show the default block size for block-based instances (10GiB). Custom volumes still behave the same as they did previously.
> 
> So now we have two possible ways to think of the size of sizeless volumes. A) The 0 size for sizeless volumes we got previously was not a bug, but only the innacurate reporting of the size. So we should change it back to 0. B) We never should show 0 as a size for a volume since the correct way to represent no size is by using -1. So showing 0 was never correct. So we should fix custom volumes accordingly.
> 
> One one hand, option A implies we should be "changing less" and also takes into account the possibility of some users relying on this 0 to check wether the instance has a root disk size set or not. It is also more flexible in case we want to make it a uunsigned integet in the future. On the other, B implies we should be "changing more" and using -1. This is more consisten with our API. As examples, we have cpu usage and project limits defaulting to -1 on the API level. Also, although changing to -1 can be seen as an API break, the disk size reported by the API was extremely unreliable, so if one was almost always getting 0 size, it is hard to say that this is meaningful/useful information to begin with. And this is why I am more inclined to go with B and keep it as -1.
> 
> I hope this was clear, if any info is missing we can discuss it in our 1-1 later, cc @tomponline

@hamistao as discussed in our 1:1 today, we tested the current state APIs for disk usage for instances and custom volumes as follows:

**Instance state:**

`lxc query /1.0/instances/c1/state`

```
dir: "disk": {},
zfs:	"disk": {
    	"root": {
        	"total": 0,
        	"usage": 7117312
    	}
	},
btrfs: 	"disk": {},
lvm: 	"disk": {
    	"root": {
        	"total": 0,
        	"usage": 1182437376
    	}
	},
```

**Custom volume state:**

`lxc query /1.0/storage-pools/default/volumes/custom/vol1/state`

```
dir: Error: Not supported
zfs: {
	"usage": {
    	"total": 0,
    	"used": 24576
	}
}
btrfs: Error: Not supported
lvm: Error: Not supported
```

So it looks like we do return `total: 0` for ZFS when the volume is unbounded.

So we agreed to maintain the existing premise that unbounded volumes should return a size of 0.

We can use -1 for total size in an error scenario if needed. We see this pattern already for CPU usage and processes fields.

tomponline:
@hamistao shall we close this for now to declutter the PR list?

PR #14598, opened 2024-12-06
Draft: shared/api: clusterNodesPost change metadata serverName to server_name

smcavoy:
This should result in the metadata being returned to be base64 encode and used as a cluster add token


simondeziel:
@smcavoy if you are indeed the original author of the code, could you please add your `Signed-off-by`? To do that: `git commit --amend -s` then force push.

Thanks!

tomponline (commented):
This is an API breaking change, so I dont think we can accept this without understanding more about why the current field name is an issue.

PR #14599, opened 2024-12-06
Storage: Add Pure Storage storage driver

MusicDin:
This PR adds `pure` storage driver for Pure Storage remote storage.
Driver supports communication over either iSCSI or NVMe/TCP.

- [x] For iSCSI, this PR is required to ensure LXD redirects `iscsiadm` to the LXD host: https://github.com/canonical/lxd-pkg-snap/pull/646 
- [x] Rebase after https://github.com/canonical/lxd/pull/14710
- [x] Rebase after https://github.com/canonical/lxd/pull/14865 (for WaitDiskSize func)

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2587,3 +2587,15 @@ The following new pool level configuration keys have been added:
 The following configuration keys have been added for volumes backed by PowerFlex:
 
 1. {config:option}`storage-powerflex-volume-conf:block.type`
+
+## `storage_driver_pure`
+
+Adds a new `pure` storage driver which allows the consumption of storage volumes from a Pure Storage storage array using either iSCSI or NVMe/TCP.
+
+The following pool level configuration keys have been added:
+
+1. {config:option}`storage-pure-pool-conf:pure.gateway`
+1. {config:option}`storage-pure-pool-conf:pure.gateway.verify`
+1. {config:option}`storage-pure-pool-conf:pure.api.token`
+1. {config:option}`storage-pure-pool-conf:pure.mode`
+1. {config:option}`storage-pure-pool-conf:pure.target`
--- a/doc/howto/storage_pools.md
+++ b/doc/howto/storage_pools.md
@@ -173,6 +173,24 @@ Create a storage pool named `pool5` that explicitly uses the PowerFlex SDC:
 
     lxc storage create pool5 powerflex powerflex.mode=sdc powerflex.pool=<id of sp1> powerflex.gateway=https://powerflex powerflex.user.name=lxd powerflex.user.password=foo
 
+#### Create a Pure Storage pool
+
+Create a storage pool named `pool1` that uses NVMe/TCP by default:
+
+    lxc storage create pool1 pure pure.gateway=https://<pure-storage-address> pure.api.token=<pure-storage-api-token>
+
+Create a storage pool named `pool2` that uses a Pure Storage gateway with a certificate that is not trusted:
+
+    lxc storage create pool2 pure pure.gateway=https://<pure-storage-address> pure.gateway.verify=false pure.api.token=<pure-storage-api-token>
+
+Create a storage pool named `pool3` that uses iSCSI to connect to Pure Storage array:
+
+    lxc storage create pool3 pure pure.gateway=https://<pure-storage-address> pure.api.token=<pure-storage-api-token> pure.mode=iscsi
+
+Create a storage pool named `pool4` that uses NVMe/TCP to connect to Pure Storage array via specific target addresses:
+
+    lxc storage create pool4 pure pure.gateway=https://<pure-storage-address> pure.api.token=<pure-storage-api-token> pure.mode=iscsi pure.target=<target_address_1>,<target_address_2>
+
 (storage-pools-cluster)=
 ## Create a storage pool in a cluster
 
@@ -238,6 +256,19 @@ Storage pool my-remote-pool2 pending on member vm02
 Storage pool my-remote-pool2 pending on member vm03
 :input: lxc storage create my-remote-pool2 powerflex powerflex.mode=sdc powerflex.pool=sp1 powerflex.domain=pd1 powerflex.gateway=https://powerflex powerflex.user.name=lxd powerflex.user.password=foo
 Storage pool my-remote-pool2 created
+```
+
+Create a third storage pool named `my-remote-pool3` using the Pure Storage driver:
+
+```{terminal}
+:input: lxc storage create my-remote-pool3 pure --target=vm01
+Storage pool my-remote-pool3 pending on member vm01
+:input: lxc storage create my-remote-pool3 pure --target=vm02
+Storage pool my-remote-pool3 pending on member vm02
+:input: lxc storage create my-remote-pool3 pure --target=vm03
+Storage pool my-remote-pool3 pending on member vm03
+:input: lxc storage create my-remote-pool3 pure pure.gateway=https://<pure-storage-address> pure.api.token=<pure-storage-api-token>
+Storage pool my-remote-pool3 created
 ```
 
 ## Configure storage pool settings
--- a/doc/reference/storage_drivers.md
+++ b/doc/reference/storage_drivers.md
@@ -15,6 +15,7 @@ storage_cephfs
 storage_cephobject
 storage_ceph
 storage_powerflex
+storage_pure
 storage_dir
 storage_lvm
 storage_zfs
@@ -27,23 +28,23 @@ See the corresponding pages for driver-specific information and configuration op
 
 Where possible, LXD uses the advanced features of each storage system to optimize operations.
 
-Feature                                     | Directory | Btrfs | LVM   | ZFS    | Ceph RBD | CephFS | Ceph Object | Dell PowerFlex
-:---                                        | :---      | :---  | :---  | :---   | :---     | :---   | :---        | :---
-{ref}`storage-optimized-image-storage`      | ❌        | ✅   | ✅     | ✅     | ✅       | ➖     | ➖          | ❌
-Optimized instance creation                 | ❌        | ✅   | ✅     | ✅     | ✅       | ➖     | ➖          | ❌
-Optimized snapshot creation                 | ❌        | ✅   | ✅     | ✅     | ✅       | ✅     | ➖          | ✅
-Optimized image transfer                    | ❌        | ✅   | ❌     | ✅     | ✅       | ➖     | ➖          | ❌
-Optimized backup (import/export)            | ❌        | ✅   | ❌     | ✅     | ❌       | ➖     | ➖          | ❌
-{ref}`storage-optimized-volume-transfer`    | ❌        | ✅   | ❌     | ✅     | ✅[^1]   | ➖     | ➖          | ❌
-{ref}`storage-optimized-volume-refresh`     | ❌        | ✅   | ✅[^2] | ✅     | ✅[^3]   | ➖     | ➖          | ❌
-Copy on write                               | ❌        | ✅   | ✅     | ✅     | ✅       | ✅     | ➖          | ✅
-Block based                                 | ❌        | ❌   | ✅     | ❌      | ✅      | ❌     | ➖          | ✅
-Instant cloning                             | ❌        | ✅   | ✅     | ✅     | ✅       | ✅     | ➖          | ❌
-Storage driver usable inside a container    | ✅        | ✅   | ❌     | ✅[^4] | ❌       | ➖     | ➖          | ❌
-Restore from older snapshots (not latest)   | ✅        | ✅   | ✅     | ❌      | ✅      | ✅     | ➖          | ✅
-Storage quotas                              | ✅[^5]    | ✅   | ✅     | ✅     | ✅       | ✅     | ✅          | ✅
-Available on `lxd init`                     | ✅        | ✅   | ✅     | ✅     | ✅       | ❌     | ❌          | ❌
-Object storage                              | ✅        | ✅   | ✅     | ✅     | ❌       | ❌     | ✅          | ❌
+Feature                                     | Directory | Btrfs | LVM   | ZFS    | Ceph RBD | CephFS | Ceph Object | Dell PowerFlex | Pure Storage
+:---                                        | :---      | :---  | :---  | :---   | :---     | :---   | :---        | :---           | :---
+{ref}`storage-optimized-image-storage`      | ❌        | ✅   | ✅     | ✅     | ✅       | ➖     | ➖          | ❌              | ✅
+Optimized instance creation                 | ❌        | ✅   | ✅     | ✅     | ✅       | ➖     | ➖          | ❌              | ✅
+Optimized snapshot creation                 | ❌        | ✅   | ✅     | ✅     | ✅       | ✅     | ➖          | ✅              | ✅
+Optimized image transfer                    | ❌        | ✅   | ❌     | ✅     | ✅       | ➖     | ➖          | ❌              | ✅
+Optimized backup (import/export)            | ❌        | ✅   | ❌     | ✅     | ❌       | ➖     | ➖          | ❌              | ❌
+{ref}`storage-optimized-volume-transfer`    | ❌        | ✅   | ❌     | ✅     | ✅[^1]   | ➖     | ➖          | ❌              | ❌
+{ref}`storage-optimized-volume-refresh`     | ❌        | ✅   | ✅[^2] | ✅     | ✅[^3]   | ➖     | ➖          | ❌              | ❌
+Copy on write                               | ❌        | ✅   | ✅     | ✅     | ✅       | ✅     | ➖          | ✅              | ✅
+Block based                                 | ❌        | ❌   | ✅     | ❌      | ✅      | ❌     | ➖          | ✅              | ✅
+Instant cloning                             | ❌        | ✅   | ✅     | ✅     | ✅       | ✅     | ➖          | ❌              | ✅
+Storage driver usable inside a container    | ✅        | ✅   | ❌     | ✅[^4] | ❌       | ➖     | ➖          | ❌              | ❌
+Restore from older snapshots (not latest)   | ✅        | ✅   | ✅     | ❌      | ✅      | ✅     | ➖          | ✅              | ✅
+Storage quotas                              | ✅[^5]    | ✅   | ✅     | ✅     | ✅       | ✅     | ✅          | ✅              | ✅
+Available on `lxd init`                     | ✅        | ✅   | ✅     | ✅     | ✅       | ❌     | ❌          | ❌              | ❌
+Object storage                              | ✅        | ✅   | ✅     | ✅     | ❌       | ❌     | ✅          | ❌              | ❌
 
 [^1]: Volumes of type `block` will fall back to non-optimized transfer when migrating to an older LXD server that doesn't yet support the `RBD_AND_RSYNC` migration type.
 [^2]: Requires {config:option}`storage-lvm-pool-conf:lvm.use_thinpool` to be enabled. Only when refreshing local volumes.
--- a/doc/reference/storage_pure.md
+++ b/doc/reference/storage_pure.md
@@ -1,0 +1,106 @@
+(storage-pure)=
+# Pure Storage - `pure`
+
+[Pure Storage](https://www.purestorage.com/) is a software-defined storage solution. It offers the consumption of redundant block storage across the network.
+
+LXD supports connecting to Pure Storage storage clusters through two protocols: either {abbr}`iSCSI (Internet Small Computer Systems Interface)` or {abbr}`NVMe/TCP (Non-Volatile Memory Express over Transmission Control Protocol)`.
+In addition, Pure Storage offers copy-on-write snapshots, thin provisioning, and other features.
+
+To use Pure Storage with LXD requires a Pure Storage API version of at least `2.21`, corresponding to a minimum Purity//FA version of `6.4.2`.
+
+Additionally, ensure that the required kernel modules for the selected protocol are installed on your host system.
+For iSCSI, the iSCSI CLI named `iscsiadm` needs to be installed in addition to the required kernel modules.
+
+## Terminology
+
+Each storage pool created in LXD using a Pure Storage driver represents a Pure Storage *pod*, which is an abstraction that groups multiple volumes under a specific name.
+One benefit of using Pure Storage pods is that they can be linked with multiple Pure Storage arrays to provide additional redundancy.
+
+LXD creates volumes within a pod that is identified by the storage pool name.
+When the first volume needs to be mapped to a specific LXD host, a corresponding Pure Storage host is created with the name of the LXD host and a suffix of the used protocol.
+For example, if the LXD host is `host01` and the mode is `nvme`, the resulting Pure Storage host would be `host01-nvme`.
+
+The Pure Storage host is then connected with the required volumes, to allow attaching and accessing volumes from the LXD host.
+The created Pure Storage host is automatically removed once there are no volumes connected to it.
+
+## The `pure` driver in LXD
+
+The `pure` driver in LXD uses Pure Storage volumes for custom storage volumes, instances, and snapshots.
+All created volumes are thin-provisioned block volumes. If required (for example, for containers and custom file system volumes), LXD formats the volume with a desired file system.
+
+LXD expects Pure Storage to be pre-configured with a specific service (e.g. iSCSI) on network interfaces whose address is provided during storage pool configuration.
+Furthermore, LXD assumes that it has full control over the Pure Storage pods it manages.
+Therefore, you should never maintain any volumes in Pure Storage pods that are not owned by LXD because LXD might disconnect or even delete them.
+
+This driver behaves differently than some of the other drivers in that it provides remote storage.
+As a result, and depending on the internal network, storage access might be a bit slower compared to local storage.
+On the other hand, using remote storage has significant advantages in a cluster setup: all cluster members have access to the same storage pools with the exact same contents, without the need to synchronize them.
+
+When creating a new storage pool using the `pure` driver in either `iscsi` or `nvme` mode, LXD automatically discovers the array's qualified name and target address (portal).
+Upon successful discovery, LXD attaches all volumes that are connected to the Pure Storage host that is associated with a specific LXD server.
+Pure Storage hosts and volume connections are fully managed by LXD.
+
+Volume snapshots are also supported by Pure Storage. However, each snapshot is associated with a parent volume and cannot be directly attached to the host.
+Therefore, when a snapshot is being exported, LXD creates a temporary volume behind the scenes. This volume is attached to the LXD host and removed once the operation is completed.
+Similarly, when a volume with at least one snapshot is being copied, LXD sequentially copies snapshots into destination volume, from which a new snapshot is created.
+Finally, once all snapshots are copied, the source volume is copied into the destination volume.
+
+(storage-pure-volume-names)=
+### Volume names
+
+Due to a [limitation](storage-pure-limitations) in Pure Storage, volume names cannot exceed 63 characters.
+Therefore, the driver uses the volume's {config:option}`storage-pure-volume-conf:volatile.uuid` to generate a shorter volume name.
+
+For example, a UUID `5a2504b0-6a6c-4849-8ee7-ddb0b674fd14` is first trimmed of any hyphens (`-`), resulting in the string `5a2504b06a6c48498ee7ddb0b674fd14`.
+To distinguish volume types and snapshots, special identifiers are prepended and appended to the volume names, as depicted in the table below:
+
+Type            | Identifier   | Example
+:--             | :---         | :----------
+Container       | `c-`         | `c-5a2504b06a6c48498ee7ddb0b674fd14`
+Virtual machine | `v-`         | `v-5a2504b06a6c48498ee7ddb0b674fd14-b` (block volume) and `v-5a2504b06a6c48498ee7ddb0b674fd14` (file system volume)
+Image (ISO)     | `i-`         | `i-5a2504b06a6c48498ee7ddb0b674fd14-i`
+Custom volume   | `u-`         | `u-5a2504b06a6c48498ee7ddb0b674fd14`
+Snapshot        | `s`          | `sc-5a2504b06a6c48498ee7ddb0b674fd14` (container snapshot)
+
+(storage-pure-limitations)=
+### Limitations
+
+The `pure` driver has the following limitations:
+
+Volume size constraints
+: Minimum volume size (quota) is `1MiB` and must be a multiple of `512B`.
+
+Snapshots cannot be mounted
+: Snapshots cannot be mounted directly to the host. Instead, a temporary volume must be created to access the snapshot's contents.
+  For internal operations, such as copying instances or exporting snapshots, LXD handles this automatically.
+
+Sharing the Pure Storage storage pool between installations
+: Sharing the same Pure Storage storage pool between multiple LXD installations is not supported.
+  If a different LXD installation tries to create a storage pool with a name that already exists, an error is returned.
+
+Recovering Pure Storage storage pools
+: Recovery of Pure Storage storage pools using `lxd recover` is currently not supported.
+
+## Configuration options
+
+The following configuration options are available for storage pools that use the `pure` driver, as well as storage volumes in these pools.
+
+(storage-pure-pool-config)=
+### Storage pool configuration
+
+% Include content from [../metadata.txt](../metadata.txt)
+```{include} ../metadata.txt
+    :start-after: <!-- config group storage-pure-pool-conf start -->
+    :end-before: <!-- config group storage-pure-pool-conf end -->
+```
+
+{{volume_configuration}}
+
+(storage-pure-vol-config)=
+### Storage volume configuration
+
+% Include content from [../metadata.txt](../metadata.txt)
+```{include} ../metadata.txt
+    :start-after: <!-- config group storage-pure-volume-conf start -->
+    :end-before: <!-- config group storage-pure-volume-conf end -->
+```
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -4175,8 +4175,10 @@ func (b *lxdBackend) EnsureImage(fingerprint string, op *operations.Operation) e
 				}
 
 				// Reset img volume variables as we just deleted the old one.
+				// Since the old volume has been removed, ensure the new volume
+				// is instantiated with its own UUID.
 				imgDBVol = nil
-				imgVol = b.GetVolume(drivers.VolumeTypeImage, contentType, image.Fingerprint, nil)
+				imgVol = b.GetNewVolume(drivers.VolumeTypeImage, contentType, image.Fingerprint, nil)
 			} else if err != nil {
 				return err
 			} else {
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -1155,6 +1155,16 @@ func (b *lxdBackend) CreateInstanceFromCopy(inst instance.Instance, src instance
 		srcVolStorageName := project.Instance(src.Project().Name, src.Name())
 		srcVol := b.GetVolume(volType, contentType, srcVolStorageName, srcConfig.Volume.Config)
 
+		// Set the parent volume's UUID.
+		if b.driver.Info().PopulateParentVolumeUUID {
+			parentUUID, err := b.getParentVolumeUUID(srcVol, src.Project().Name)
+			if err != nil {
+				return err
+			}
+
+			srcVol.SetParentUUID(parentUUID)
+		}
+
 		volCopy := drivers.NewVolumeCopy(vol, targetSnapshots...)
 		srcVolCopy := drivers.NewVolumeCopy(srcVol, sourceSnapshots...)
 
@@ -2016,6 +2026,16 @@ func (b *lxdBackend) CreateInstanceFromImage(inst instance.Instance, fingerprint
 	// will cause a non matching configuration which will always fall back to non optimized storage.
 	vol := b.GetNewVolume(volType, contentType, volStorageName, volumeConfig)
 
+	// Set the parent volume UUID.
+	if b.driver.Info().PopulateParentVolumeUUID {
+		parentUUID, err := b.getParentVolumeUUID(vol, inst.Project().Name)
+		if err != nil {
+			return err
+		}
+
+		vol.SetParentUUID(parentUUID)
+	}
+
 	// Validate config and create database entry for new storage volume.
 	err = VolumeDBCreate(b, inst.Project().Name, inst.Name(), "", volType, false, vol.Config(), inst.CreationDate(), time.Time{}, contentType, true, false)
 	if err != nil {
@@ -2364,6 +2384,16 @@ func (b *lxdBackend) CreateInstanceFromMigration(inst instance.Instance, conn io
 
 		snapshotStorageName := project.Instance(inst.Project().Name, instSnapshot.Name())
 		targetSnapshots = append(targetSnapshots, b.GetVolume(volType, contentType, snapshotStorageName, snap.Config))
+	}
+
+	// Set the parent volume's UUID.
+	if b.driver.Info().PopulateParentVolumeUUID {
+		parentUUID, err := b.getParentVolumeUUID(vol, projectName)
+		if err != nil {
+			return err
+		}
+
+		vol.SetParentUUID(parentUUID)
 	}
 
 	volCopy := drivers.NewVolumeCopy(vol, targetSnapshots...)
@@ -3078,6 +3108,16 @@ func (b *lxdBackend) MigrateInstance(inst instance.Instance, conn io.ReadWriteCl
 		_ = filesystem.SyncFS(inst.RootfsPath())
 	}
 
+	// Set the parent volume UUID.
+	if b.driver.Info().PopulateParentVolumeUUID {
+		parentUUID, err := b.getParentVolumeUUID(vol, inst.Project().Name)
+		if err != nil {
+			return err
+		}
+
+		vol.SetParentUUID(parentUUID)
+	}
+
 	volCopy := drivers.NewVolumeCopy(vol, sourceSnapshots...)
 
 	err = b.driver.MigrateVolume(volCopy, conn, args, op)
@@ -3710,6 +3750,16 @@ func (b *lxdBackend) DeleteInstanceSnapshot(inst instance.Instance, op *operatio
 
 	vol := b.GetVolume(volType, contentType, snapVolName, dbVol.Config)
 
+	// Set the parent volume UUID.
+	if b.driver.Info().PopulateParentVolumeUUID {
+		parentUUID, err := b.getParentVolumeUUID(vol, inst.Project().Name)
+		if err != nil {
+			return err
+		}
+
+		vol.SetParentUUID(parentUUID)
+	}
+
 	volExists, err := b.driver.HasVolume(vol)
 	if err != nil {
 		return err
@@ -3911,6 +3961,17 @@ func (b *lxdBackend) MountInstanceSnapshot(inst instance.Instance, op *operation
 		return nil, err
 	}
 
+	// Set the parent volume UUID.
+	if b.driver.Info().PopulateParentVolumeUUID {
+		parentUUID, err := b.getParentVolumeUUID(vol, inst.Project().Name)
+		if err != nil {
+			return nil, err
+		}
+
+		vol.SetParentUUID(parentUUID)
+	}
+
+	// Mount the snapshot.
 	err = b.driver.MountVolumeSnapshot(vol, op)
 	if err != nil {
 		return nil, err
@@ -3960,6 +4021,17 @@ func (b *lxdBackend) UnmountInstanceSnapshot(inst instance.Instance, op *operati
 		return err
 	}
 
+	// Set the parent volume UUID.
+	if b.driver.Info().PopulateParentVolumeUUID {
+		parentUUID, err := b.getParentVolumeUUID(vol, inst.Project().Name)
+		if err != nil {
+			return err
+		}
+
+		vol.SetParentUUID(parentUUID)
+	}
+
+	// Unmount volume.
 	_, err = b.driver.UnmountVolumeSnapshot(vol, op)
 
 	return err
@@ -5726,6 +5798,15 @@ func (b *lxdBackend) CreateCustomVolumeFromMigration(projectName string, conn io
 		return err
 	}
 
+	if b.driver.Info().PopulateParentVolumeUUID {
+		parentUUID, err := b.getParentVolumeUUID(vol, projectName)
+		if err != nil {
+			return err
+		}
+
+		vol.SetParentUUID(parentUUID)
+	}
+
 	revert := revert.New()
 	defer revert.Fail()
 
@@ -6615,6 +6696,16 @@ func (b *lxdBackend) DeleteCustomVolumeSnapshot(projectName, volName string, op 
 	volStorageName := project.StorageVolume(projectName, volName)
 
 	vol := b.GetVolume(drivers.VolumeTypeCustom, contentType, volStorageName, volume.Config)
+
+	// Set the parent volume's UUID.
+	if b.driver.Info().PopulateParentVolumeUUID {
+		parentUUID, err := b.getParentVolumeUUID(vol, projectName)
+		if err != nil {
+			return err
+		}
+
+		vol.SetParentUUID(parentUUID)
+	}
 
 	// Delete the snapshot from the storage device.
 	// Must come before DB VolumeDBDelete so that the volume ID is still available.
@@ -7797,4 +7888,31 @@ func (b *lxdBackend) CreateCustomVolumeFromBackup(srcBackup backup.Info, srcData
 
 	revert.Success()
 	return nil
+}
+
+// getParentVolumeUUID returns the UUID of the parent's volume.
+// If the volume has no parent, an empty string is returned.
+func (b *lxdBackend) getParentVolumeUUID(vol drivers.Volume, projectName string) (string, error) {
+	parentName, _, isSnapshot := api.GetParentAndSnapshotName(vol.Name())
+	if !isSnapshot {
+		// Volume has no parent.
+		return "", nil
+	}
+
+	// Ensure the parent name does not contain a project prefix.
+	_, parentName = project.StorageVolumeParts(parentName)
+
+	// Load storage volume from the database.
+	parentDBVol, err := VolumeDBGet(b, projectName, parentName, vol.Type())
+	if err != nil {
+		return "", fmt.Errorf("Failed to extract parent UUID from snapshot %q in project %q: %w", vol.Name(), projectName, err)
+	}
+
+	// Extract parent volume UUID.
+	parentUUID := parentDBVol.Config["volatile.uuid"]
+	if parentUUID == "" {
+		return "", fmt.Errorf("Parent volume %q of snapshot %q in project %q does not have UUID set)", parentName, projectName, vol.Name())
+	}
+
+	return parentUUID, nil
 }
--- a/lxd/storage/drivers/driver_pure.go
+++ b/lxd/storage/drivers/driver_pure.go
@@ -19,6 +19,9 @@ type pure struct {
 	// Holds the low level HTTP client for the Pure Storage API.
 	// Use pure.client() to retrieve the client struct.
 	httpClient *pureClient
+
+	// apiVersion indicates the Pure Storage API version.
+	apiVersion string
 }
 
 // load is used initialize the driver. It should be used only once.
--- a/lxd/storage/drivers/driver_pure.go
+++ b/lxd/storage/drivers/driver_pure.go
@@ -61,6 +61,7 @@ func (d *pure) Info() Info {
 		DirectIO:                     true,
 		IOUring:                      true,
 		MountedRoot:                  false,
+		PopulateParentVolumeUUID:     true,
 	}
 }
 
--- a/lxd/storage/drivers/driver_pure.go
+++ b/lxd/storage/drivers/driver_pure.go
@@ -1,0 +1,119 @@
+package drivers
+
+import (
+	"github.com/canonical/lxd/lxd/migration"
+	"github.com/canonical/lxd/lxd/operations"
+	"github.com/canonical/lxd/shared/api"
+)
+
+// pureLoaded indicates whether load() function was already called for the Pure Storage driver.
+var pureLoaded = false
+
+// pureVersion indicates Pure Storage version.
+var pureVersion = ""
+
+type pure struct {
+	common
+
+	// Holds the low level HTTP client for the Pure Storage API.
+	// Use pure.client() to retrieve the client struct.
+	httpClient *pureClient
+}
+
+// load is used initialize the driver. It should be used only once.
+func (d *pure) load() error {
+	// Done if previously loaded.
+	if pureLoaded {
+		return nil
+	}
+
+	pureLoaded = true
+	return nil
+}
+
+// client returns the drivers Pure Storage client. A new client is created only if it does not already exist.
+func (d *pure) client() *pureClient {
+	if d.httpClient == nil {
+		d.httpClient = newPureClient(d)
+	}
+
+	return d.httpClient
+}
+
+// isRemote returns true indicating this driver uses remote storage.
+func (d *pure) isRemote() bool {
+	return true
+}
+
+// Info returns info about the driver and its environment.
+func (d *pure) Info() Info {
+	return Info{
+		Name:                         "pure",
+		Version:                      pureVersion,
+		DefaultBlockSize:             d.defaultBlockVolumeSize(),
+		DefaultVMBlockFilesystemSize: d.defaultVMBlockFilesystemSize(),
+		OptimizedImages:              false,
+		PreservesInodes:              false,
+		Remote:                       d.isRemote(),
+		VolumeTypes:                  []VolumeType{VolumeTypeCustom, VolumeTypeVM, VolumeTypeContainer, VolumeTypeImage},
+		BlockBacking:                 true,
+		RunningCopyFreeze:            true,
+		DirectIO:                     true,
+		IOUring:                      true,
+		MountedRoot:                  false,
+	}
+}
+
+// FillConfig populates the storage pool's configuration file with the default values.
+func (d *pure) FillConfig() error {
+	return nil
+}
+
+// Validate checks that all provided keys are supported and there is no conflicting or missing configuration.
+func (d *pure) Validate(config map[string]string) error {
+	return nil
+}
+
+// Create is called during pool creation and is effectively using an empty driver struct.
+// WARNING: The Create() function cannot rely on any of the struct attributes being set.
+func (d *pure) Create() error {
+	err := d.FillConfig()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Update applies any driver changes required from a configuration change.
+func (d *pure) Update(changedConfig map[string]string) error {
+	return nil
+}
+
+// Delete removes the storage pool (Pure Storage pod).
+func (d *pure) Delete(op *operations.Operation) error {
+	return nil
+}
+
+// Mount mounts the storage pool.
+func (d *pure) Mount() (bool, error) {
+	// Nothing to do here.
+	return true, nil
+}
+
+// Unmount unmounts the storage pool.
+func (d *pure) Unmount() (bool, error) {
+	// Nothing to do here.
+	return true, nil
+}
+
+// GetResources returns the pool resource usage information.
+func (d *pure) GetResources() (*api.ResourcesStoragePool, error) {
+	res := &api.ResourcesStoragePool{}
+	return res, nil
+}
+
+// MigrationTypes returns the type of transfer methods to be used when doing migrations between pools in preference order.
+func (d *pure) MigrationTypes(contentType ContentType, refresh bool, copySnapshots bool) []migration.Type {
+	return []migration.Type{}
+}
--- a/lxd/storage/drivers/driver_pure.go
+++ b/lxd/storage/drivers/driver_pure.go
@@ -4,6 +4,7 @@ import (
 	"github.com/canonical/lxd/lxd/migration"
 	"github.com/canonical/lxd/lxd/operations"
 	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/validate"
 )
 
 // pureLoaded indicates whether load() function was already called for the Pure Storage driver.
@@ -72,6 +73,41 @@ func (d *pure) FillConfig() error {
 
 // Validate checks that all provided keys are supported and there is no conflicting or missing configuration.
 func (d *pure) Validate(config map[string]string) error {
+	rules := map[string]func(value string) error{
+		"size": validate.Optional(validate.IsSize),
+		// lxdmeta:generate(entities=storage-pure; group=pool-conf; key=pure.api.token)
+		//
+		// ---
+		//  type: string
+		//  shortdesc: API token for Pure Storage gateway authentication
+		"pure.api.token": validate.Optional(),
+		// lxdmeta:generate(entities=storage-pure; group=pool-conf; key=pure.gateway)
+		//
+		// ---
+		//  type: string
+		//  shortdesc: Address of the Pure Storage gateway
+		"pure.gateway": validate.Optional(validate.IsRequestURL),
+		// lxdmeta:generate(entities=storage-pure; group=pool-conf; key=pure.gateway.verify)
+		//
+		// ---
+		//  type: bool
+		//  defaultdesc: `true`
+		//  shortdesc: Whether to verify the Pure Storage gateway's certificate
+		"pure.gateway.verify": validate.Optional(validate.IsBool),
+		// lxdmeta:generate(entities=storage-pure; group=pool-conf; key=volume.size)
+		// Default Pure Storage volume size rounded to 512B. The minimum size is 1MiB.
+		// ---
+		//  type: string
+		//  defaultdesc: `10GiB`
+		//  shortdesc: Size/quota of the storage volume
+		"volume.size": validate.Optional(validate.IsMultipleOfUnit("512B")),
+	}
+
+	err := d.validatePool(config, rules, d.commonVolumeRules())
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/lxd/storage/drivers/driver_pure.go
+++ b/lxd/storage/drivers/driver_pure.go
@@ -1,9 +1,15 @@
 package drivers
 
 import (
+	"fmt"
+	"net/http"
+
 	"github.com/canonical/lxd/lxd/migration"
 	"github.com/canonical/lxd/lxd/operations"
+	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/revert"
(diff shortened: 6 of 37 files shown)


doc/howto/storage_pools.md
```diff
@@ -173,6 +173,20 @@ Create a storage pool named `pool5` that explicitly uses the PowerFlex SDC:
 
     lxc storage create pool5 powerflex powerflex.mode=sdc powerflex.pool=<id of sp1> powerflex.gateway=https://powerflex powerflex.user.name=lxd powerflex.user.password=foo
 
+#### Create a PureStorage pool
+
+Create a storage pool named `pool1` that uses NVMe/TCP by default:
+
+    lxc storage create pool1 pure pure.gateway=https://<purestorage-address> pure.api.token=<purestorage-api-token> pure.array.address=<nvme-ip>
```
simondeziel, line 180: That gateway HTTPS verification presumably relies on the OS provided trust store so it would work if it's an "official TLS" cert like one signed by Let's Encrypt.

I must admit not knowing how to provide a different trusted CA if one doesn't have an official cert.

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
...
+Custom volume   | `u-`         | `u-5a2504b06a6c48498ee7ddb0b674fd14`
+Snapshot        | `s`          | `sc-5a2504b06a6c48498ee7ddb0b674fd14` (container snapshot)
+
+(storage-pure-limitations)=
+### Limitations
+
+The `pure` driver has the following limitations:
+
+Volume size constraints
+: Minimum volume size is `1MiB` and has to be a multiply of `512B`.
```
minaelee, line 65:
```suggestion
: Minimum volume size is `1MiB` and has to be a multiple of `512B`.
```

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
...
+
+Snapshots cannot be mounted
+: Snapshots cannot be mounted directly to the host, which requires creating a temporary volume when accessing snapshot's contents.
```
minaelee, line 68:
```suggestion
: Snapshots cannot be mounted directly to the host. Instead, a temporary volume must be created to access the snapshot's contents.
```

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
...
+  For internal operations, such as copying instances or exporting snapshots, LXD handles this automatically.
+
+Sharing the PureStorage storage pool between installations
+: Sharing the same PureStorage storage pool between multiple LXD installations is not supported.
+  If different LXD installation tries to create a storage pool with the name that already exists, an error is be returned.
```
minaelee, line 73:
```suggestion
  If a different LXD installation tries to create a storage pool with a name that already exists, an error is returned.
```

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
...
+
+Recovering PureStorage storage pools
+: Recovery of PureStorage storage pools using `lxd recover` is currently not supported.
+
+## Configuration options
+
+The following configuration options are available for storage pools that use the `pure` driver and for storage volumes in these pools.
```
minaelee, line 80:
```suggestion
The following configuration options are available for storage pools that use the `pure` driver, as well as storage volumes in these pools.
```

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
...
+Another benefit of using PureStorage pods, is that they can be linked with multiple PureStorage arrays to provide additional redundancy.
+LXD creates volumes within a pod that is identified by the storage pool name.
+When the first volume needs to be mapped to a specific LXD host, a corresponding PureStorage host is created with the name of the LXD host and a suffix of used protocol.
+The PureStorage host is then connected with the required volumes, to allow attaching and accessing volumes from the LXD host.
+Created PureStorage hosts are automatically removed once there is no volumes connected to it.
+
+## `pure` driver in LXD
+
+The `pure` driver in LXD uses PureStorage volumes for custom storage volumes, instances, and snapshots.
+All created volumes are thin-provisioned block volumes, that LXD formats with a desired file system if required (for containers and custom file system volumes).
```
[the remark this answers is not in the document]
MusicDin: Thanks for pointing this out. I've just provided some examples. LXD formats the block volume with a file system for custom file system volumes and containers. However, one file system volume is also created for each VM.

IMO, this option nicely describes the situation:
```
If required (for example, for containers and custom file system volumes), LXD formats the volume with a desired file system.
```

MusicDin:
@minaelee Thanks for the comments. I've addressed all of them, but will ask you for another review just before we merge the PR, as there may still be some changes to the docs. Thanks again :)

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -0,0 +1,1411 @@
...
+
+	mountPath := snapVol.MountPath()
+
+	if snapVol.contentType == ContentTypeFS && shared.PathExists(mountPath) {
+		err = wipeDirectory(mountPath)
+		if err != nil {
+			return err
+		}
+
+		err = os.Remove(mountPath)
```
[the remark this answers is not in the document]
tomponline: I believe @simondeziel is planning to address this issue more generally higher up the call stack, so wont block this PR.

lxd/storage/drivers/driver_pure.go
```diff
@@ -0,0 +1,118 @@
...
+func (d *pure) isRemote() bool {
+	return true
+}
+
+// Info returns info about the driver and its environment.
+func (d *pure) Info() Info {
+	return Info{
+		Name:                         "pure",
+		Version:                      pureVersion,
+		DefaultVMBlockFilesystemSize: d.defaultVMBlockFilesystemSize(),
```
tomponline, line 107: needs `DefaultBlockSize:             d.defaultBlockVolumeSize()`, also now

lxd/storage/backend_lxd.go
```diff
@@ -7778,3 +7869,26 @@ func (b *lxdBackend) CreateCustomVolumeFromBackup(srcBackup backup.Info, srcData
 	revert.Success()
 	return nil
 }
+
+// getSnapshotParentUUID returns the UUID of the snapshot's parent volume.
+func (b *lxdBackend) getSnapshotParentUUID(vol drivers.Volume, projectName string) (string, error) {
+	parentName, _, isSnapshot := api.GetParentAndSnapshotName(vol.Name())
+	if !isSnapshot {
+		// Volume has no parent.
+		return "", nil
```
tomponline, line 7878: should this be an error as it means the vol passed in is not a snapshot and the function is being used incorrectly?

lxd/storage/backend_lxd.go
```diff
@@ -7778,3 +7869,26 @@ func (b *lxdBackend) CreateCustomVolumeFromBackup(srcBackup backup.Info, srcData
...
+	}
+
+	// Load storage volume from database.
+	parentDBVol, err := VolumeDBGet(b, projectName, parentName, vol.Type())
+	if err != nil {
+		return "", fmt.Errorf("Failed to extract parent UUID from snapshot %q: %w", vol.Name(), err)
```
tomponline, line 7884: lets include project in the error too

lxd/storage/backend_lxd.go
```diff
@@ -7778,3 +7869,26 @@ func (b *lxdBackend) CreateCustomVolumeFromBackup(srcBackup backup.Info, srcData
...
+	}
+
+	// Extract UUID.
+	parentUUID := parentDBVol.Config["volatile.uuid"]
+	if parentUUID == "" {
+		return "", fmt.Errorf("Parent volume %q of snapshot %q does not have UUID set", parentName, vol.Name())
```
tomponline, line 7890: include project please

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -12,3 +49,186 @@ func newPureClient(driver *pure) *pureClient {
...
+
+	return body, nil
+}
+
+// request issues a HTTP request against the PureStorage gateway.
+func (p *pureClient) request(method string, path string, reqBody io.Reader, reqHeaders map[string]string, respBody any, respHeaders map[string]string) error {
+	var url string
+
+	// Construct the request URL.
+	if strings.HasPrefix(path, "/api") {
```
tomponline, line 205: why are we using this function in such a way that the function needs to translate the path into what we actually want? This could do with a comment explaining i think.

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -12,3 +49,186 @@ func newPureClient(driver *pure) *pureClient {
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("Failed to send request: %w", err)
+	}
+
+	defer resp.Body.Close()
+
+	// Wrap unauthorized requests into an API status error.
+	if resp.StatusCode == http.StatusUnauthorized {
+		return api.StatusErrorf(http.StatusUnauthorized, "Unauthorized request")
```
tomponline, line 256: Are you masking the response body on purpose (it might contain useful info right)?

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -12,3 +49,186 @@ func newPureClient(driver *pure) *pureClient {
...
+		return pureErr
+	}
+
+	return nil
+}
+
+// requestAuthenticated issues an authenticated HTTP request against the PureStorage gateway. In case
+// the access token is expired, the function will try to obtain a new one.
+func (p *pureClient) requestAuthenticated(method string, path string, reqBody io.Reader, respBody any) error {
+	retries := 1
```
tomponline, line 292: what are the retries for?

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -12,3 +49,186 @@ func newPureClient(driver *pure) *pureClient {
...
+			// Either the error is not of type unauthorized or the maximum number of
+			// retries has been exceeded.
+			return err
+		}
+
+		return nil
+	}
+}
+
+// getAPIVersion returns the list of API version that are supported by the PureStorage.
```
tomponline, line 325: API versions

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -12,3 +49,186 @@ func newPureClient(driver *pure) *pureClient {
...
+		APIVersions []string `json:"version"`
+	}
+
+	err := p.request(http.MethodGet, "/api/api_version", nil, nil, &resp, nil)
+	if err != nil {
+		return nil, fmt.Errorf("Failed retrieve available API versions from PureStorage: %w", err)
+	}
+
+	if len(resp.APIVersions) == 0 {
+		return nil, fmt.Errorf("PureStorage does not support any API version")
```
tomponline, line 337: versions

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -12,3 +49,186 @@ func newPureClient(driver *pure) *pureClient {
...
+// an access token is retrieved and stored within a client. The access token is then used for further
+// authentication.
+func (p *pureClient) login() error {
+	if p.accessToken != "" {
+		// Token has been already obtained.
+		return nil
+	}
+
+	reqHeaders := map[string]string{
+		"Api-Token": p.driver.config["pure.api.token"],
```
tomponline, line 353: does this need to have this casing of Api?

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -12,3 +49,186 @@ func newPureClient(driver *pure) *pureClient {
...
+	respHeaders := make(map[string]string)
+
+	err := p.request(http.MethodPost, "/login", nil, reqHeaders, nil, respHeaders)
+	if err != nil {
+		return fmt.Errorf("Failed to login: %w", err)
+	}
+
+	p.accessToken = respHeaders["X-Auth-Token"]
+	if p.accessToken == "" {
+		return errors.New("Failed to obtain access token")
```
tomponline, line 365: do we get any other sort of error message from this post?

lxd/storage/drivers/driver_pure.go
```diff
@@ -121,6 +127,24 @@ func (d *pure) Create() error {
 		return err
 	}
 
+	revert := revert.New()
+	defer revert.Fail()
+
+	poolSizeBytes, err := units.ParseByteSizeString(d.config["size"])
+	if err != nil {
+		return fmt.Errorf("Failed to parse storage size (quota limit): %w", err)
```
tomponline, line 234: im interested to know why you mentioned "quota" in this message, is this a pure specific concept?

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -37,6 +38,51 @@ func (p *pureError) Error() string {
 	return strings.TrimSuffix(p.Errors[0].Message, ".")
 }
 
+// Matches returns true if the error status code is equal to the provided status code and the error message
+// contains the provided substring.
+func (p *pureError) Matches(statusCode int, substring string) bool {
```
tomponline, line 77: not enjoying that we need this hacky approach, but ill reserve judgement until i see where its needed.

lxd/storage/drivers/driver_pure_util.go
MusicDin: Yep, will include the comment. 

The `/api` paths do not include API version in front of them. This way, if `/api/...` is provided, we literally use that path, otherwise we prepend the required API version. 

Also to point out, currently the latest API version is always used (list of supported versions is retrieved from PureStorage), but could be something worth limiting or exposing through the config in the future.

lxd/storage/backend_lxd.go
MusicDin: I was thinking about this one, and decided to just return an empty string to avoid checking for snapshot on every place where the function is called. However, in such case the function name "GetParentVolumeUUID" would probably be more appropirate, as non-snapshot volumes have no parent.

On the other side, returning an error as you have suggested prevents unnecessary or incorrect usage of the function.

lxd/storage/drivers/driver_pure_util.go
tomponline: If you made substring a variadic arg you could use this function in `IsNotFoundError`

lxd/storage/drivers/driver_pure_util.go
tomponline, line 1: Lets use api.NewURL for URL building

tomponline: Because as is often the way when building urls manually you've not handled URL encoding.

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -232,3 +278,120 @@ func (p *pureClient) login() error {
...
+		if err != nil {
+			return err
+		}
+
+		err = p.requestAuthenticated(http.MethodPatch, fmt.Sprintf("/pods?names=%s", poolName), req, nil)
+		if err != nil {
+			return fmt.Errorf("Failed to restore storage pool %q: %w", poolName, err)
+		}
+
+		logger.Warn("Storage pool has been restored", logger.Ctx{"pool": poolName})
```
tomponline, line 425: Info level?

tomponline: we can return nil here and avoid the else

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -232,3 +278,120 @@ func (p *pureClient) login() error {
...
+		req, err := p.createBodyReader(map[string]any{
+			"destroyed": true,
+		})
+		if err != nil {
+			return err
+		}
+
+		err = p.requestAuthenticated(http.MethodPatch, fmt.Sprintf("/pods?names=%s&destroy_contents=true", poolName), req, nil)
+		if err != nil {
+			perr, ok := err.(*pureError)
```
tomponline, line 487: Dont you have a helper function for not found errors?

lxd/storage/drivers/driver_pure.go
MusicDin: This makes no sense since the value is parsed from config. Will fix it.

lxd/storage/drivers/driver_pure_util.go
tomponline, line 487: feels like we could do with a helper function for this pattern of casting the err to pureError and then checking if its not found, i can see it repeated twice in this function alone.

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -232,3 +278,120 @@ func (p *pureClient) login() error {
...
+	// Eradicate the storage pool by permanently deleting it along all of its contents.
+	err = p.requestAuthenticated(http.MethodDelete, fmt.Sprintf("/pods?names=%s&eradicate_contents=true", poolName), nil, nil)
+	if err != nil {
+		perr, ok := err.(*pureError)
+		if ok {
+			if perr.IsNotFoundError() {
+				return nil
+			}
+
+			if perr.Matches(http.StatusBadRequest, "Cannot eradicate pod") {
```
tomponline, line 505: does pure storage have the ability to return error strings in other languages?

Or are these documented "constants"?

lxd/storage/drivers/driver_pure_util.go
MusicDin: Yes the casting can be moved into the helper function to make it cleaner. Will fix, thanks.

lxd/storage/drivers/driver_pure_util.go
tomponline: please add a comment, as it wasnt clear if this was because of some flaky endpoint

lxd/storage/drivers/driver_pure_util.go
tomponline: we should be targeting the API version we support right? Not using the latest as otherwise something might change and the pure appliance would expect us to use that newer version not the older one. This should be documented too.

lxd/storage/drivers/driver_pure_util.go
MusicDin: Completely agree and am option to suggestions, but the main issue is that most errors (with just few rare exceptions) return status code `400`. Therefore, the substring matching is the important part of the function.

lxd/storage/drivers/driver_pure_util.go
tomponline: > The `/api` paths do not include API version in front of them

What are the "/api" paths? Given that in the else statement you then add /api/version to the path provided, it seems that infact some `/api` paths do have a version?

lxd/storage/drivers/driver_pure_util.go
MusicDin: > If you made substring a variadic arg you could use this function in IsNotFoundError

Yes indeed. Thanks.

lxd/storage/drivers/driver_pure.go
```diff
@@ -76,6 +102,11 @@ func (d *pure) Info() Info {
 
 // FillConfig populates the storage pool's configuration file with the default values.
 func (d *pure) FillConfig() error {
+	// Use iSCSI by default.
+	if d.config["pure.mode"] == "" {
+		d.config["pure.mode"] = pureModeISCSI
```
tomponline: should we prefer nvme by default like powerflex?

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -507,3 +531,33 @@ func (d *pure) serverName() (string, error) {
 
 	return hostname, nil
 }
+
+// getVolumeName returns the fully qualified name derived from the volume's UUID.
+func (d *pure) getVolumeName(vol Volume) (string, error) {
```
tomponline, line 1493: can we add tests for this sort of well-isolated function? Would help get our coverage rates up, especially for code we cannot easily test in the CI system tests

lxd/storage/drivers/driver_pure.go
```diff
@@ -19,6 +20,11 @@ var pureLoaded = false
 // pureVersion indicates PureStorage version.
 var pureVersion = ""
 
+// PureStorage modes.
+const (
+	pureModeISCSI = "iscsi"
```
tomponline, line 25: where's nvme?

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -439,9 +463,33 @@ func (p *pureClient) getHosts() ([]pureHost, error) {
...
+}
+
+// createHost creates a new host with provided initiator qualified names that can be associated
+// with specific volumes.
+func (p *pureClient) createHost(hostName string, qns []string) error {
+	body := make(map[string]any, 1)
+
+	switch p.driver.config["pure.mode"] {
+	case pureModeISCSI:
+		body["iqns"] = qns
```
tomponline, line 780: should we have a default case here that returns an error if the mode is unhandled?

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -460,8 +508,15 @@ func (p *pureClient) createHost(hostName string) error {
 }
 
 // updateHost updates an existing host.
-func (p *pureClient) updateHost(hostName string) error {
-	req, err := p.createBodyReader(map[string]any{})
+func (p *pureClient) updateHost(hostName string, qns []string) error {
+	body := make(map[string]any, 1)
+
+	switch p.driver.config["pure.mode"] {
```
tomponline, line 807: should we have a default case here that returns an error if the mode is unhandled?

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -517,6 +572,475 @@ func (p *pureClient) disconnectHostFromVolume(poolName string, volName string, h
...
+// getTarget retrieves the target that can be connected to using the initiator's qualified name.
+func (p *pureClient) getTarget() (*pureTarget, error) {
+	var resp pureResponse[pureTarget]
+
+	targetAddr := p.driver.config["pure.array.address"]
+
+	// Append service port to the address if missing.
+	// This way we can use exact matching in the filter to find the target.
+	if !strings.Contains(targetAddr, ":") {
+		// Default NVMe port is 4420.
```
tomponline, line 879: you've used switch cases elsewhere for this sort of logic, suggest doing the same here, with a default case that returns an error if mode is unhandled.

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -517,6 +572,475 @@ func (p *pureClient) disconnectHostFromVolume(poolName string, volName string, h
...
+
+		// Fetch the host by IQN.
+		host, err := d.client().getHostByIQN(iqn)
+		if err != nil {
+			if !api.StatusErrorCheck(err, http.StatusNotFound) {
+				return "", nil, err
+			}
+
+			// The host with a given IQN does not exist, therefore, create a new one.
+			hostname, err = d.serverName()
```
tomponline, line 927: is the assumption that if the LXD host becomes clustered after the pool is setup the then IQN mapping will break?

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -517,6 +572,475 @@ func (p *pureClient) disconnectHostFromVolume(poolName string, volName string, h
...
+
+			cleanup := revert.Clone().Fail
+			revert.Success()
+			return cleanup, nil
+		}
+
+		// Attempt to login into discovered iSCSI targets.
+		_, stderr, err := shared.RunCommandSplit(d.state.ShutdownCtx, nil, nil, "iscsiadm", "--mode", "node", "--targetname", targetIQN, "--portal", targetAddr, "--login")
+		if err != nil {
+			return nil, fmt.Errorf("Failed to connect to PureStorage array %q via iSCSI: %s\n%v", targetAddr, stderr, err)
```
tomponline, line 1032: we shouldn't be putting newlines in error messages.

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -517,6 +572,475 @@ func (p *pureClient) disconnectHostFromVolume(poolName string, volName string, h
...
+func (d *pure) mapVolume(vol Volume) error {
+	revert := revert.New()
+	defer revert.Fail()
+
+	volName, err := d.getVolumeName(vol)
+	if err != nil {
+		return err
+	}
+
+	unlock, err := locking.Lock(d.state.ShutdownCtx, d.config["pure.mode"])
```
tomponline, line 1115: As this appears to be copying following the powerflex driver's lead of serialising access to the storage mapping subsystem I think we need to move this out to a separate set of functions that can be called from both drivers to clearly show that this lock is spanning multiple drivers.

Also, and this is my fault as I clearly missed this on the PF review, but we should add a prefix to the lock name, something like "storage_".

As each implementation is different after the lock, suggest something like we've done before in storage with `MountTask()` taking a function as an argument to avoid forgetting to unlock.

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -517,6 +572,475 @@ func (p *pureClient) disconnectHostFromVolume(poolName string, volName string, h
...
+	}
+
+	// Connect to the array using iscsi. Connection can be established only
+	// when at least one volume is connected to the host.
+	cleanup, err = d.connect()
+	if err != nil {
+		return err
+	}
+
+	revert.Add(cleanup)
```
tomponline, line 1147: this can never be called, as you mark the reverter as succeeded on next line?

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -517,6 +572,475 @@ func (p *pureClient) disconnectHostFromVolume(poolName string, volName string, h
...
+	// Disconnect the volume from the host and ignore error if connection does not exist.
+	err = d.client().disconnectHostFromVolume(vol.pool, volName, host.Name)
+	if err != nil && !api.StatusErrorCheck(err, http.StatusNotFound) {
+		return err
+	}
+
+	// Wait until the volume has disappeared.
+	volumePath, _, _ := d.getMappedDevPath(vol, false)
+	if volumePath != "" {
+		if d.config["pure.mode"] == pureModeISCSI {
```
tomponline, line 1199: Is there some utility functions we can create/reuse from powerflex for the volume mapping/unmapping stuff?

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -517,6 +572,475 @@ func (p *pureClient) disconnectHostFromVolume(poolName string, volName string, h
...
+		}
+	}
+
+	// Disconnect the volume from the host and ignore error if connection does not exist.
+	err = d.client().disconnectHostFromVolume(vol.pool, volName, host.Name)
+	if err != nil && !api.StatusErrorCheck(err, http.StatusNotFound) {
+		return err
+	}
+
+	// Wait until the volume has disappeared.
```
tomponline, line 1196: this comment seems to be in the wrong place, as it should be down by waitGone, as we've not yet request it be disconnected by this point, so its not going to have disappeared yet right?

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -517,6 +572,475 @@ func (p *pureClient) disconnectHostFromVolume(poolName string, volName string, h
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// getMappedDevPath returns the local device path for the given volume.
+// Indicate with mapVolume if the volume should get mapped to the system if it isn't present.
+func (d *pure) getMappedDevPath(vol Volume, mapVolume bool) (string, revert.Hook, error) {
```
tomponline, line 1251: can this be shared with powerflex?

lxd/storage/drivers/driver_pure_util.go
MusicDin: Huh, a very good point. But it should not break. The host's name will be different and it will result in each LXD cluster member having its own host with different IQN (that is read from the `/etc/iscsi/initiatorname.iscsi`). Individual host will then be mapped with the volumes that are mounted on that particular host.

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -517,6 +572,475 @@ func (p *pureClient) disconnectHostFromVolume(poolName string, volName string, h
...
+		if err != nil {
+			return "", nil, err
+		}
+
+		revert.Add(func() { _ = d.unmapVolume(vol) })
+	}
+
+	// findDevPathFunc has to be called in a loop with a set timeout to ensure
+	// all the necessary directories and devices can be discovered.
+	findDevPathFunc := func(diskPrefix string, diskSuffix string) (string, error) {
```
tomponline, line 1266: or can at least parts of it be shared with powerflex?

lxd/storage/drivers/driver_pure_util.go
MusicDin: Yes, I wanted to call it anyway to ensure we do not miss the reverter entry if we ever extend the function. Can remove if desired.

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -517,6 +572,475 @@ func (p *pureClient) disconnectHostFromVolume(poolName string, volName string, h
...
+	case pureModeISCSI:
+		diskPrefix = "scsi-"
+		diskSuffix = fmt.Sprintf("%s::%s", vol.pool, volName)
+	}
+
+	timeout := time.Now().Add(15 * time.Second)
+
+	// It might take a while to create the local disk.
+	// Retry until it can be found.
+	for {
```
tomponline, line 1335: Please can we use a pattern like `tryExists` or `waitGone` that accepts a context with a WithTimeout, that way we can include the state's shutdown context too and immediately return on next iteration rather than keeping going until the timeout hits.

lxd/storage/drivers/driver_pure_util.go
MusicDin: Yes this part can be shared with PowerFlex, however, we use different prefix/suffix to locate a volume.

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -517,6 +572,475 @@ func (p *pureClient) disconnectHostFromVolume(poolName string, volName string, h
...
+			break
+		}
+
+		// Exit if the volume wasn't explicitly mapped.
+		// Doing a retry would run into the timeout when the device isn't mapped.
+		if !mapVolume {
+			break
+		}
+
+		time.Sleep(10 * time.Millisecond)
```
tomponline, line 1361: this seems too short, and will likely cause this to be repeated many times in a tight loop causing CPU spikes.

Other sleeps in the storage package tend to me in the hundreds of milliseconds, for example `waitGone` uses `time.Sleep(500 * time.Millisecond)`

tomponline: powerflex likely needs to be updated here too.

lxd/storage/drivers/driver_pure_util.go
MusicDin: Would it be acceptable if I address shared functionality with PowerFlex in a separate smaller PR afterwards? My point being is that messing with PowerFlex without being able to test seems a bit risky, but also do not want to put burden on Julian to keep testing PowerFlex after each change to this big PR.

lxd/storage/drivers/driver_pure_util.go
MusicDin: Should not matter. I think it can be all lower case if preferred?

lxd/storage/drivers/driver_pure.go
```diff
@@ -102,9 +120,9 @@ func (d *pure) Info() Info {
 
 // FillConfig populates the storage pool's configuration file with the default values.
 func (d *pure) FillConfig() error {
-	// Use iSCSI by default.
+	// Use NVMe by default.
```
tomponline, line 123: ah ok so we dont want to change history inside a single PR, so we need to rebase this, otherwise the reviewer has to keep track of how the commits change each file over time. Better to present the final version inside the PR.

lxd/storage/drivers/driver_pure.go
```diff
@@ -61,6 +62,23 @@ func (d *pure) load() error {
...
+		// Detect and record the version of the NVMe CLI.
+		// The NVMe CLI is shipped with the snap.
+		out, err := shared.RunCommand("nvme", "version")
+		if err != nil {
+			return fmt.Errorf("Failed to get nvme-cli version: %w", err)
+		}
+
+		fields := strings.Split(strings.TrimSpace(out), " ")
+		if strings.HasPrefix(out, "nvme version ") && len(fields) > 2 {
+			powerFlexVersion = fmt.Sprintf("%s (nvme-cli)", fields[2])
```
tomponline, line 75: powerflexVersion?

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -1001,6 +1103,22 @@ func (d *pure) getMappedDevPath(vol Volume, mapVolume bool) (string, revert.Hook
...
 		diskSuffix = fmt.Sprintf("%s::%s", vol.pool, volName)
+	case pureModeNVMe:
+		diskPrefix = "nvme-eui."
+
+		pureVol, err := d.client().getVolume(vol.pool, volName)
+		if err != nil {
+			return "", nil, err
+		}
+
+		if len(pureVol.Serial) <= 10 {
```
tomponline, line 1321: what is the magic number 10 here?

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -186,14 +189,99 @@ func (d *pure) RefreshVolume(vol VolumeCopy, srcVol VolumeCopy, refreshSnapshots
...
+	}
+
+	volName, err := d.getVolumeName(vol)
+	if err != nil {
+		return err
+	}
+
+	var host *pureHost
+
+	switch d.config["pure.mode"] {
```
tomponline, line 585: can we abstract this whole section for disconnecting a volume into a separate function that deals with the different modes?

lxd/storage/drivers/driver_pure_util.go
MusicDin: Yes, comment is missing.

For iscsi, when volumes are mounted, they will have entire volume name in the device suffix "<pod>::<volume_name>".

However, for nvme/tcp this is not the case. Instead, device name contains last 10 characters from the volume serial.

lxd/storage/drivers/driver_pure.go
```diff
@@ -283,7 +283,34 @@ func (d *pure) Unmount() (bool, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
 	res := &api.ResourcesStoragePool{}
+
+	res.Space.Total = uint64(pool.Quota)
+	res.Space.Used = uint64(pool.Space.UsedBytes)
+
+	if pool.Quota == 0 {
```
tomponline, line 313: @hamistao is this aligned with how we are going to report storage pool space for unbounded pools (like dir)? 

Is it just unbounded volumes which we will report as 0 size?

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -786,7 +948,8 @@ func (d *pure) CreateVolumeSnapshot(snapVol Volume, op *operations.Operation) er
 	revert.Add(func() { _ = d.DeleteVolumeSnapshot(snapVol, op) })
 
 	// For VMs, create a snapshot of the filesystem volume too.
-	if snapVol.IsVMBlock() {
+	// Skip if snapshotVMfilesystem is false to prevent overwriting separately copied volumes.
+	if snapVol.IsVMBlock() && snapshotVMfilesystem {
```
tomponline, line 1170: is snapshotVMfilesystem a pure specific thing? Why is it needed, seems unusual?

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -342,8 +342,208 @@ func (d *pure) CreateVolumeFromMigration(vol VolumeCopy, conn io.ReadWriteCloser
...
+		return err
+	}
+
+	revert.Add(cleanup)
+
+	revert.Success()
+	return nil
+}
+
+// refreshVolume updates an existing volume to match the state of another. For VM volumes and snaphots,
```
tomponline, line 401: >  For VM volumes and snaphots,

This only makes sense to me for VM volumes, not snapshots in general, what does it mean?

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -1219,14 +1219,107 @@ func (d *pure) DeleteVolumeSnapshot(snapVol Volume, op *operations.Operation) er
...
+
+	// Get the snapshot volume name.
+	snapVolName, err := d.getVolumeName(snapVol)
+	if err != nil {
+		return err
+	}
+
+	// A PureStorage snapshot cannot be mounted. To mount a snapshot, a new volume
+	// has to be created from the snapshot.
+	err = d.client().copyVolumeSnapshot(snapVol.pool, parentVolName, snapVolName, snapVol.pool, snapVolName)
```
tomponline, line 1261: can this temporary name conflict with other volume names?

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -342,8 +342,208 @@ func (d *pure) CreateVolumeFromMigration(vol VolumeCopy, conn io.ReadWriteCloser
...
+	volName, err := d.getVolumeName(vol.Volume)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create new reverter snapshot, which is used to revert the original volume in case of
+	// an error. Snapshots are also required to be first copied into destination volume,
+	// from which a new snapshot is created to effectively copy a snapshot. If any error
+	// occurs, the destination volume has been already modified and needs reverting.
+	reverterSnapshotName := "lxd-reverter-snapshot"
```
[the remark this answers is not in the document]
MusicDin: Is there any other storage driver that sequentially copies snapshots through the "main volume"? The idea behind `lxd-reverter-snapshot` is just to prevent breaking the target/destination instance - leaving it in the state of unknown snapshot.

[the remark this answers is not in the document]
tomponline: can you add that to a list of things to discuss in our next call please

lxd/storage/drivers/driver_pure_volumes.go
MusicDin: No. Snapshots are prefixed with `s` which is only the case for snapshots. Similarly, it *should* not conflict with other snapshots as they have a unique volatile UUID.

The only potential case where it could cause the problem: if the same volume with snapshots is being refreshed simultaneously to two different locations, but AFAIK the locks should prevent such case?

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
...
+
+LXD expects PureStorage to be pre-configured with a specific service (e.g. iSCSI) on network interfaces whose address is provided during storage pool configuration.
+Furthermore, LXD assumes that it has full control over the PureStorage pods it manages.
+Therefore, you should never maintain any volumes in PureStorage pods that are not owned by LXD because LXD might disconnect or even delete them.
+
+This driver behaves differently than some of the other drivers in that it provides remote storage.
+As a result, and depending on the internal network, storage access might be a bit slower compared to local storage.
+On the other hand, using remote storage has significant advantages in a cluster setup: all cluster members have access to the same storage pools with the exact same contents, without the need to synchronize them.
+
+When creating a new storage pool using the `pure` driver in either `iscsi` or `nvme-tcp` mode, LXD tries to discover one of the array's qualified names on the address `pure.array.address`.
```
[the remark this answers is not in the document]
MusicDin: To avoid confusion if we go for `nvme-fc` in the future. But I agree, having it consistent with PowerFlex would make more sense. Will update everywhere.

tomponline (commented):
I made it through! :)

Thanks @MusicDin this is looking great.

My main areas of concern:

1. URL construction, lets use the URL builder package rather than raw string manipulation - this has burned us in the past, lets not repeat the mistake.
2. Code sharing/de-duplication with powerflex driver. I accept your point of de-risking impacting powerflex driver by duplication, but I am concerned that we are also going to end up diverging unnecessarily and having some shared functions that deal with mapping/unmapping nvme and iscsi volumes is going to be helpful going forward.
3. On the point of nvme/iscsi mode, it'd be nice to try and abstract that stuff out into some smaller functions and then have the driver call those functions for the specific operation we want to do. I saw quite a few iqn resolution function calls being duplicated.
4. Unit tests. If there are some areas of the driver that can be unit tested for correct logic (like volume name generation) then we should as these external appliance drivers are more difficult to test so would be good to get some code coverage in these areas.

doc/reference/storage_pure.md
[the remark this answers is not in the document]
tomponline: would there also be potentially an iscsi-fc?

lxd/storage/drivers/driver_pure.go
MusicDin: Currently, if the storage pool is unbound, the entire storage space is reported. Let me know what is desired outcome in such case. Probably just `0`?

doc/reference/storage_pure.md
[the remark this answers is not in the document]
MusicDin: There is `scsi-fc` (note missing `i` :))

lxd/storage/drivers/driver_pure.go
MusicDin: Oops..

doc/reference/storage_pure.md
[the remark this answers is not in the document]
tomponline: So does pure call nvme over tcp "nvme-tcp"?

lxd/storage/drivers/driver_pure_util.go
MusicDin: If we get any error from this post, it will be returned as "Failed to login: <err>".
For example: `Error: Failed to create storage pool "pure": Failed to login: Unable to list user for API token` where `Unable to list user for API token` is PureStorage error.

The check `p.accessToken == ""` is just to ensure that we also successfully parsed the the access token from the response. This code is only reached if Post request was successful.

doc/reference/storage_pure.md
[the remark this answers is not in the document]
MusicDin: Yes.

Those are values in PureStorage:
- `nvme-tcp`
- `nvme-fc`
- `scsi-fc`
- `iscsi`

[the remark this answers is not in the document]
tomponline: @mionaalex shall we go for consistency with powerflex and have "nvme" mode to mean "nvme over tcp" for both powerflex and pure, or use pure specific terminology of "nvme-tcp" and be different from powerflex driver?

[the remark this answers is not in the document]
mionaalex: @tomponline I'd go for consistency, so either we change Pure, or we change PowerFlex as well to be `nvme-tcp` (don't know if that's feasible at this point). If we go for the former, we can still use `nvme-fc` for fiber channel to distinguish between the two.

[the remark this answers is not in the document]
tomponline: we shouldnt change powerflex as its published and would break deployments, so "nvme" for pure it is.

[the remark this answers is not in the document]
MusicDin: Perfect, thank you both. Will update accordingly.

lxd/storage/drivers/driver_pure.go
MusicDin: Fixed - it was set to NVMe by default, but in the later commit.

lxd/storage/drivers/driver_pure_volumes.go
MusicDin: Should be `For VMs ...`

lxd/storage/drivers/driver_pure_volumes.go
MusicDin: This has to do sequential snapshot copying through the main volume, where in such case we do not want to create filesystem volume. Let's talk this one through on the meeting as it will be easier for me to explain.

lxd/storage/drivers/driver_pure.go
hamistao: We had decided to use `-1` in such cases. The PR setting this standard wasn't merged yet so we have time to change if needed, but personally I prefer `-1`.

lxd/storage/drivers/driver_pure_util.go
MusicDin: Just `/api/api_versions`

tomponline: So should we do an exact match then?

lxd/storage/drivers/driver_pure.go
tomponline: -1 for unbounded volume size is ok, but i dont think -1 for unbounded pools is a good idea, i thought we had agreed to present the entire capacity of the underlying storage device so that it was possible to ascertain space for a particular unbounded volume by considering the pool capacity?

lxd/storage/drivers/driver_pure_util.go
MusicDin: That will work as well. Will fix.

doc/reference/storage_drivers.md
```diff
@@ -27,29 +28,30 @@ See the corresponding pages for driver-specific information and configuration op
...
+Object storage                              | ✅        | ✅   | ✅     | ✅     | ❌       | ❌     | ✅          | ❌              | ❌
 
 [^1]: Volumes of type `block` will fall back to non-optimized transfer when migrating to an older LXD server that doesn't yet support the `RBD_AND_RSYNC` migration type.
 [^2]: Requires {config:option}`storage-lvm-pool-conf:lvm.use_thinpool` to be enabled. Only when refreshing local volumes.
 [^3]: Only for volumes of type `block`.
 [^4]: Requires {config:option}`storage-zfs-volume-conf:zfs.delegate` to be enabled.
 [^5]: % Include content from [storage_dir.md](storage_dir.md)
+[^6]: The optimized refresh works only on the single PureStorage array that is used by single LXD host or cluster.
 
       ```{include} storage_dir.md
```
minaelee, line 56: This include directive needs to go below the footnote definition for `[^5]:` on line 53, before the footnote definition for `[^6]`. It looks like this currently in the [preview build](https://canonical-ubuntu-documentation-library--14599.com.readthedocs.build/lxd/en/14599/reference/storage_drivers/#id13):
image

minaelee (changes requested):

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
+(storage-pure)=
+# PureStorage - `pure`
+
+[PureStorage](https://www.purestorage.com/) is a software-defined storage solution. It offers the consumption of redundant block storage across the network.
+
+LXD offers access to PureStorage storage clusters using either iSCSI or NVMe/TCP.
```
minaelee, line 6:
```suggestion
LXD supports connecting to PureStorage storage clusters through two protocols: either {abbr}`iSCSI (Internet Small Computer Systems Interface)` or {abbr}`NVMe/TCP (Non-Volatile Memory Express over Transmission Control Protocol)`.
```

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
...
+In addition, PureStorage offers copy-on-write snapshots, thin provisioning, and other features.
+
+To use PureStorage, ensure the required kernel modules for the selected protocol are installed on your host system.
```
minaelee, line 9:
```suggestion
To use PureStorage, ensure that the required kernel modules for the selected protocol are installed on your host system.
```

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
...
+For iSCSI, the iSCSI CLI named `iscsiadm` needs to be installed in addition to the required kernel modules.
+
+## Terminology
+
+Each storage pool created in LXD using a PureStorage driver represents a PureStorage *pod*, which is an abstraction that groups multiple volumes under a specific name.
+Another benefit of using PureStorage pods is that they can be linked with multiple PureStorage arrays to provide additional redundancy.
+LXD creates volumes within a pod that is identified by the storage pool name.
```
minaelee, line 16:
```suggestion

LXD creates volumes within a pod that is identified by the storage pool name.
```

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
...
+LXD offers access to PureStorage storage clusters using either iSCSI or NVMe/TCP.
+In addition, PureStorage offers copy-on-write snapshots, thin provisioning, and other features.
+
+To use PureStorage, ensure the required kernel modules for the selected protocol are installed on your host system.
+For iSCSI, the iSCSI CLI named `iscsiadm` needs to be installed in addition to the required kernel modules.
+
+## Terminology
+
+Each storage pool created in LXD using a PureStorage driver represents a PureStorage *pod*, which is an abstraction that groups multiple volumes under a specific name.
+Another benefit of using PureStorage pods is that they can be linked with multiple PureStorage arrays to provide additional redundancy.
```
minaelee, line 15:
```suggestion
One benefit of using PureStorage pods is that they can be linked with multiple PureStorage arrays to provide additional redundancy.
```

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
...
+LXD creates volumes within a pod that is identified by the storage pool name.
+When the first volume needs to be mapped to a specific LXD host, a corresponding PureStorage host is created with the name of the LXD host and a suffix of used protocol.
```
minaelee, line 17:
```suggestion
When the first volume needs to be mapped to a specific LXD host, a corresponding PureStorage host is created with the name of the LXD host and a suffix of the used protocol.
```

minaelee: Can we add an example here of what that PureStorage host name might look like? Like, "For example, a PureStorage host name might be ...".

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
...
+The PureStorage host is then connected with the required volumes, to allow attaching and accessing volumes from the LXD host.
```
minaelee, line 18:
```suggestion

The PureStorage host is then connected with the required volumes, to allow attaching and accessing volumes from the LXD host.
```

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
...
+Created PureStorage hosts are automatically removed once there is no volumes connected to it.
```
minaelee, line 19:
```suggestion
The created PureStorage host is automatically removed once there are no volumes connected to it.
```

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
...
+
+## `pure` driver in LXD
```
minaelee, line 21:
```suggestion
## The `pure` driver in LXD
```

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
...
+Similarly, when a volume with at least one snapshot is being copied, LXD sequentially copies snapshots into destination volume, from which a new snapshot is created.
+Finally, once all snapshots are copied, the source volume is copied into the destination volume.
+
+(storage-pure-volume-names)=
+### Volume names
+
+Due to a [limitation](storage-pure-limitations) in PureStorage, volume names cannot exceed 63 characters.
+Therefore, the driver uses the volume's {config:option}`storage-pure-volume-conf:volatile.uuid` to generate a shorter volume name.
+
+For example, a UUID `5a2504b0-6a6c-4849-8ee7-ddb0b674fd14` is first trimmed of any hyphens (`-`) resulting in the string `5a2504b06a6c48498ee7ddb0b674fd14`. To distinguish volume types and snapshots, special identifiers are prepended and appended to the volume names, as depicted in the table below:
```
minaelee, line 49:
```suggestion
For example, a UUID `5a2504b0-6a6c-4849-8ee7-ddb0b674fd14` is first trimmed of any hyphens (`-`), resulting in the string `5a2504b06a6c48498ee7ddb0b674fd14`. To distinguish volume types and snapshots, special identifiers are prepended and appended to the volume names, as depicted in the table below:
```

doc/reference/storage_pure.md
```diff
@@ -0,0 +1,100 @@
...
+Custom volume   | `u-`         | `u-5a2504b06a6c48498ee7ddb0b674fd14`
+Snapshot        | `s`          | `sc-5a2504b06a6c48498ee7ddb0b674fd14` (container snapshot)
+
+(storage-pure-limitations)=
+### Limitations
+
+The `pure` driver has the following limitations:
+
+Volume size constraints
+: Minimum volume size is `1MiB` and has to be a multiple of `512B`.
```
minaelee, line 65:
```suggestion
: Minimum volume size (quota) is `1MiB` and must be a multiple of `512B`.
```

doc/reference/storage_drivers.md
MusicDin: Thanks. I've removed this footnote as optimized refresh will be addressed in a follow-up PR and I will ensure it is correctly placed then.

doc/reference/storage_drivers.md
```diff
@@ -27,29 +28,30 @@ See the corresponding pages for driver-specific information and configuration op
...
+Storage quotas                              | ✅[^5]    | ✅   | ✅     | ✅     | ✅       | ✅     | ✅          | ✅              | ✅
+Available on `lxd init`                     | ✅        | ✅   | ✅     | ✅     | ✅       | ❌     | ❌          | ❌              | ❌
+Object storage                              | ✅        | ✅   | ✅     | ✅     | ❌       | ❌     | ✅          | ❌              | ❌
 
 [^1]: Volumes of type `block` will fall back to non-optimized transfer when migrating to an older LXD server that doesn't yet support the `RBD_AND_RSYNC` migration type.
 [^2]: Requires {config:option}`storage-lvm-pool-conf:lvm.use_thinpool` to be enabled. Only when refreshing local volumes.
 [^3]: Only for volumes of type `block`.
 [^4]: Requires {config:option}`storage-zfs-volume-conf:zfs.delegate` to be enabled.
 [^5]: % Include content from [storage_dir.md](storage_dir.md)
+[^6]: The optimized refresh works only on the single PureStorage array that is used by single LXD host or cluster.
```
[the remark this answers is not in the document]
MusicDin: I've removed this footnote as and marked optimized refresh as a unsupported feature, as this will be addressed in a follow-up PR, where I will ensure to explain it properly. Maybe this limitation will not even apply.

lxd/storage/backend_lxd.go
MusicDin: > Rename to GetParentVolumeUUID

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -12,3 +51,200 @@ func newPureClient(driver *pure) *pureClient {
...
+		return nil, fmt.Errorf("Failed to write request body: %w", err)
+	}
+
+	return body, nil
+}
+
+// request issues a HTTP request against the Pure Storage gateway.
+func (p *pureClient) request(method string, url url.URL, reqBody io.Reader, reqHeaders map[string]string, respBody any, respHeaders map[string]string) error {
+	// Extract scheme and host from the gateway URL.
+	urlParts := strings.Split(p.driver.config["pure.gateway"], "://")
```
tomponline, line 214: strings.Cut?

tomponline: url.Scheme, url.Host, valid := strings.Cut()

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -478,6 +478,78 @@ func (p *pureClient) deleteStoragePool(poolName string) error {
 	return nil
 }
 
+// getVolume returns the volume behind volumeID.
+func (p *pureClient) getVolume(poolName string, volName string) (*pureVolume, error) {
+	var resp pureResponse[pureVolume]
+
+	url := api.NewURL().Path("volumes").WithQuery("names", fmt.Sprintf("%s::%s", poolName, volName))
```
tomponline, line 554: lets try and avoid fmt.Sprintf and use concat instead to avoid reflection.

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -12,3 +51,200 @@ func newPureClient(driver *pure) *pureClient {
...
+// request issues a HTTP request against the Pure Storage gateway.
+func (p *pureClient) request(method string, url url.URL, reqBody io.Reader, reqHeaders map[string]string, respBody any, respHeaders map[string]string) error {
+	// Extract scheme and host from the gateway URL.
+	urlParts := strings.Split(p.driver.config["pure.gateway"], "://")
+	if len(urlParts) != 2 {
+		return fmt.Errorf("Invalid Pure Storage gateway URL: %q", p.driver.config["pure.gateway"])
+	}
+
+	// Set request URL schema and host.
+	url.Scheme = urlParts[0]
```
simondeziel, line 220: Should we ensure that `url.Scheme` is one of `http` or `https`?

lxd/storage/drivers/driver_pure.go
hamistao: Sorry, I missed that notification. As I stated, the a `dir` pool does not have any persistent size on its configuration, but when querying for its `lxc info` it gets reported as the entire capacity of the underlying storage device, just as @tomponline said. The implementation for this can be seen [here](https://github.com/hamistao/lxd/blob/364ad5e4c611cda853da581ba323298fe9815b2d/lxd/storage/drivers/generic_vfs.go#L38).

tomponline:
@MusicDin ready for rebase

tomponline:
@MusicDin just a heads up, but please see https://github.com/canonical/lxd/pull/14900#discussion_r1937491546 for a tip on parsing the target addresses.

doc/api-extensions.md
```diff
@@ -2568,3 +2568,7 @@ GET request to fetch the LXD entities has the `with-access-entitlements=<comma_s
 ## `profiles_all_projects`
 
 This adds support for listing profiles across all projects using the `all-projects` parameter in `GET /1.0/profiles` requests.
+
+## `storage_driver_pure`
+
+Adds support for Pure Storage storage driver.
```
tomponline, line 2574: We should mention the new config keys added as well as the storage driver name here

tomponline: @roosterfish please can you put up a PR to add an API extension for powerflex, it seems this was missed when you originally added it. Thanks

roosterfish: https://github.com/canonical/lxd/pull/14914

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -927,9 +927,29 @@ func (d *pure) unmapVolume(vol Volume) error {
 			}
 
 			devName := filepath.Base(volumePath)
-			err := removeDevice(devName)
-			if err != nil {
-				return fmt.Errorf("Failed to unmap volume %q: Failed to remove device %q: %w", vol.name, devName, err)
+			if strings.HasPrefix(devName, "dm-") {
+				// Multipath device (/dev/dm-*) itself is not removavle.
```
tomponline, line 1382: typo removavle

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -186,14 +189,83 @@ func (d *pure) RefreshVolume(vol VolumeCopy, srcVol VolumeCopy, refreshSnapshots
...
+		return err
+	}
+
+	host, err := d.client().getCurrentHost()
+	if err != nil {
+		if !api.StatusErrorCheck(err, http.StatusNotFound) {
+			return err
+		}
+	} else {
+		// Dicsconnect the volume from the host.
```
tomponline, line 590: typo

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -186,14 +189,83 @@ func (d *pure) RefreshVolume(vol VolumeCopy, srcVol VolumeCopy, refreshSnapshots
...
+	volName, err := d.getVolumeName(vol)
+	if err != nil {
+		return err
+	}
+
+	host, err := d.client().getCurrentHost()
+	if err != nil {
+		if !api.StatusErrorCheck(err, http.StatusNotFound) {
+			return err
+		}
```
tomponline, line 588: should there be a `return err` here?

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -566,3 +697,13 @@ func (d *pure) RenameVolumeSnapshot(snapVol Volume, newSnapshotName string, op *
 	// Renaming a volume snapshot won't change an actual name of the Pure Storage volume snapshot.
 	return nil
 }
+
+// getSnapshotParentVolume returns a parent volume that has volatile.uuid set to the snapshot's parent UUID.
+func getSnapshotParentVolume(snapVol Volume) Volume {
```
tomponline, line 1424: lets move this to generic utils rather than in the pool driver file, as although only the pure driver uses this, it is not pure specific logic.

tomponline: or snapVol.GetParentVolume()?

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -375,7 +375,17 @@ func (d *pure) UpdateVolume(vol Volume, changedConfig map[string]string) error {
 
 // GetVolumeUsage returns the disk space used by the volume.
 func (d *pure) GetVolumeUsage(vol Volume) (int64, error) {
-	return 0, ErrNotSupported
+	volName, err := d.getVolumeName(vol)
+	if err != nil {
+		return -1, err
```
tomponline, line 763: @hamistao is this correct to return -1 for volume usage error?

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -389,8 +397,137 @@ func (d *pure) GetVolumeUsage(vol Volume) (int64, error) {
 }
 
 // SetVolumeQuota applies a size limit on volume.
-// Does nothing if supplied with an empty/zero size.
+// Does nothing if supplied with an non-positive size.
```
tomponline, line 775: @MusicDin could other storage drivers do with this clarification too?

lxd/storage/drivers/driver_pure_volumes.go
tomponline: @MusicDin this still looks potentially problematic/different from other storage drivers, i think we need to evaluate why this is "special" and document better whats going on here.

lxd/storage/drivers/driver_pure_volumes.go
MusicDin: Interesting the spellcheck did not catch it

roosterfish:
> We should consider whether we'd be better off moving the connection/disconnection of targets to pool mount/unmount time, and have an ongoing go routine that keeps retrying connections the entire time and maintaining connections even if they drop.

I was about to comment that for PowerFlex we have to be careful as you can only connect after the first volume got mapped to the host. However this isn't true anymore when using `nvme connect` instead of `connect-all`.
However the host already has to be created before running `nvme connect` otherwise it will return `Failed to write to /dev/nvme-fabrics: Input/output error`.

On a side note I have tried supplying a list of addresses to the `nvme connect` command which is supported hoping that it will return after the first successful connection. But instead it errors out if it cannot connect to all of the given addresses.

lxd/storage/drivers/driver_pure_volumes.go
tomponline: sounds like a good place for a comment :)

tomponline:
@MusicDin please can you break the linter commits (and any other commits not related to pure) out to a different PR.

lxd/storage/drivers/driver_pure_volumes.go
tomponline: @MusicDin is there a reason we dont do the filesystem and block volume snapshot copy together, rather than doing all the filesystem snapshots first and then doing all of the block volume snapshots next?

MusicDin:
Sure: https://github.com/canonical/lxd/pull/14915

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -1224,14 +1224,107 @@ func (d *pure) DeleteVolumeSnapshot(snapVol Volume, op *operations.Operation) er
...
+	}
+
+	// A Pure Storage snapshot cannot be mounted. To mount a snapshot, a new volume
+	// has to be created from the snapshot.
+	err = d.client().copyVolumeSnapshot(snapVol.pool, parentVolName, snapVolName, snapVol.pool, snapVolName)
+	if err != nil {
+		return err
+	}
+
+	// Ensure temporary snapshot volume is remooved in case of an error.
```
tomponline, line 1289: not a blocker to merge, but typo here

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -1224,14 +1224,107 @@ func (d *pure) DeleteVolumeSnapshot(snapVol Volume, op *operations.Operation) er
...
+
+	// Get the snapshot volume name.
+	snapVolName, err := d.getVolumeName(snapVol)
+	if err != nil {
+		return err
+	}
+
+	// A Pure Storage snapshot cannot be mounted. To mount a snapshot, a new volume
+	// has to be created from the snapshot.
+	err = d.client().copyVolumeSnapshot(snapVol.pool, parentVolName, snapVolName, snapVol.pool, snapVolName)
```
tomponline, line 1284: @MusicDin question for you:

What happens if I've got a snapshot mounted (so a temporary volume has been created), and then lxd gets stopped and the host restarted uncleanly, and i go to mount the snapshot again, will this call fail because the temporary volume already exists?

doc/howto/storage_pools.md
```diff
@@ -173,6 +173,24 @@ Create a storage pool named `pool5` that explicitly uses the PowerFlex SDC:
...
+
+    lxc storage create pool2 pure pure.gateway=https://<pure-storage-address> pure.gateway.verify=false pure.api.token=<pure-storage-api-token>
+
+Create a storage pool named `pool3` that uses iSCSI to connect to Pure Storage array:
+
+    lxc storage create pool3 pure pure.gateway=https://<pure-storage-address> pure.api.token=<pure-storage-api-token> pure.mode=iscsi
+
+Create a storage pool named `pool4` that uses NVMe/TCP to connect to Pure Storage array via specific target addresses:
+
+    lxc storage create pool4 pure pure.gateway=https://<pure-storage-address> pure.api.token=<pure-storage-api-token> pure.mode=iscsi pure.target=<target_address_1>,<target_address_2>
```
tomponline, line 192: > uses NVMe/TCP to connect

> pure.mode=iscsi

copy pasto? Please can you follow up in separate PR these small fixes.

tomponline (approved):
LGTM thanks! There's a couple of minor things to follow up with, but ill merge this as-is.

lxd/storage/drivers/driver_pure_volumes.go
MusicDin: ^ Until the pool itself is removed, then everything is wiped.

tomponline: > or overwrites an existing one if it already exists.

OK cool,that was the bit I was mainly interested in.

> For example, if the instance is then removed, this temporary snapshot could remain in Pure Storage.

Can we scan volumes on instance (or snapshot) delete to find any temporary volumes related and delete them too?

MusicDin: > Can we scan volumes on instance (or snapshot) delete to find any temporary volumes related and delete them too?

I cannot foreseen anything that would block us from doing that. 
What is your suggestion, at which phase should this happen? Should we run the cleanup on LXD boot / when storage is loaded?

tomponline: when deleting the instance snapshot, and when deleting the instance (which by extension is deleting the snapshot)

MusicDin: Ahh yeah of course. So in practice this would mean when we are deleting a specific snapshot, we check if a volume exists with such uuid.

For example:
1. Delete snapshot v1/snap0
3. Check if there is an existing volume with `snap0-uuid` (obviously correctly resolving volume prefixes and suffixes in the process)
4. Delete such volume if exists
5. Delete an actual snapshot `snap0`

PR #14601, opened 2024-12-06
Simplification of `actions/lp-snap-build` to be reusable

simondeziel:
I had not properly looked at all the possible variation of the `Trigger Launchpad snap build` step that all our repos and branches have. This step being repo/branch specific, it's hard to push it to a generic action.

As such, go back a little and keep the actual build trigger and git interaction parts out of the action.


tomponline (commented):
Does passing the secret to the action cause it to be logged somewhere as the input to the actions?

simondeziel:
> Does passing the secret to the action cause it to be logged somewhere as the input to the actions?

Since this is the official recommended way, I hope not. From what I've read, we should see this obfuscated with `***`.

tomponline:
> Since this is the official recommended way, I hope not. From what I've read, we should see this obfuscated with `***`.

Can you check on your own fork with a test secret in a workflow and double check, as dont want that ssh key published.

simondeziel:
Proper obfuscation happens (https://github.com/simondeziel/lxd/actions/runs/12243790782/job/34154120501) so the GH doc is accurate.

.github/workflows/tests.yml
```diff
@@ -665,49 +665,28 @@ jobs:
...
-          git config --global user.email "lxd@lists.canonical.com"
-          git config --global commit.gpgsign true
-          git config --global gpg.format "ssh"
-          git config --global user.signingkey ~/.ssh/id_ed25519.pub
           localRev="$(git rev-parse HEAD)"
-          go install github.com/canonical/lxd-ci/lxd-snapcraft@latest
-          git clone -b "${TARGET}" git+ssh://lxdbot@git.launchpad.net/~canonical-lxd/lxd ~/lxd-pkg-snap-lp
-          cd ~/lxd-pkg-snap-lp
-          lxd-snapcraft -package lxd -set-version "git-${localRev:0:7}" -set-source-commit "${localRev}"
+          git clone --depth 1 -b "${BRANCH}" "${REPO}" ~/"${PACKAGE}-pkg-snap-lp"
```
tomponline, line 686: can we move these steps into the workflow too?

simondeziel:
@tomponline PTAL

tomponline (commented):
What was the reason we cant move the call to lxd-snapscraft and the commit, push etc into the action as well?

tomponline (approved):

PR #14602, opened 2024-12-06
test: add markdown table with the duration of each test

simondeziel:


simondeziel:
@tomponline please have a look at this: https://github.com/canonical/lxd/actions/runs/12238740428?pr=14602#summary-34138270509

ATM, I'm just echo'ing to the designated file as we don't have any other use for step summaries but that will be easy to change later on if needed.

ATM, the tests are listed in execution order but I was wondering if we should not sort them by duration instead? Any opinion?

tomponline (approved):

PR #14604, opened 2024-12-07
doc: clarify that storage size is equal to quota

minaelee:
To address #14349 where users seek information about disk quotas, these commits clarify that when storage `size` is discussed in LXD docs, it means the same as `quota`. This already exists in some of the docs, but sparsely and inconsistently.

### DIFF
--- a/doc/cloud-init.md
+++ b/doc/cloud-init.md
@@ -19,7 +19,7 @@ Possible actions include, for example:
 * Adding users
 * Enabling services
 * Running commands or scripts
-* Automatically growing the file system of a VM to the size of the disk
+* Automatically growing the file system of a VM to the size (quota) of the disk
 
 See the {ref}`cloud-init:index` for detailed information.
 
--- a/doc/howto/storage_pools.md
+++ b/doc/howto/storage_pools.md
@@ -17,7 +17,7 @@ To create a storage pool, use the following command:
 
     lxc storage create <pool_name> <driver> [configuration_options...]
 
-Unless specified otherwise, LXD sets up loop-based storage with a sensible default size (20% of the free disk space, but at least 5 GiB and at most 30 GiB).
+Unless specified otherwise, LXD sets up loop-based storage with a sensible default size/quota (20% of the free disk space, but at least 5 GiB and at most 30 GiB).
 
 See the {ref}`storage-drivers` documentation for a list of available configuration options for each driver.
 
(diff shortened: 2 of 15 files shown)


doc/cloud-init.md
```diff
@@ -19,7 +19,7 @@ Possible actions include, for example:
 * Adding users
 * Enabling services
 * Running commands or scripts
-* Automatically growing the file system of a VM to the size of the disk
+* Automatically growing the file system of a VM to the size (quota) of the disk
```
MggMuggins, line 22: Is there any reason to prefer `size (quota)` over `size/quota`? Aesthetically I think I prefer the latter as parenthesizing it makes "quota" seem to be the less apt descriptor. My feeling is that we really do mean "quota" when we say "size".

minaelee: I went with that because all the configuration options, CLI commands, etc, use `size`, so it seems to deserve a heavier weight. Otherwise, I do understand what you mean about the word quota being a more accurate description.

tomponline: quota is a less apt description in some senses.

LXD models the concept of "volumes" that have a size. 

Depending on the storage driver the volume's size is enforced using the storage driver's quota functionality.

But LXD itself doesn't have storage quota support as a primary concept.

Instead the aim here is to allow users who are looking for "quota" to discover the "size" concept.

doc/howto/storage_pools.md
```diff
@@ -17,7 +17,7 @@ To create a storage pool, use the following command:
 
     lxc storage create <pool_name> <driver> [configuration_options...]
 
-Unless specified otherwise, LXD sets up loop-based storage with a sensible default size (20% of the free disk space, but at least 5 GiB and at most 30 GiB).
+Unless specified otherwise, LXD sets up loop-based storage with a sensible default size/quota (20% of the free disk space, but at least 5 GiB and at most 30 GiB).
```
tomponline, line 20: why did we use size/quota here instead of brackets?

doc/reference/storage_zfs.md
```diff
@@ -86,7 +86,7 @@ ZFS provides two different quota properties: `quota` and `refquota`.
 `quota` restricts the total size of a {spellexception}`dataset`, including its snapshots and clones.
 `refquota` restricts only the size of the data in the {spellexception}`dataset`, not its snapshots and clones.
 
-By default, LXD uses the `quota` property when you set up a quota for your storage volume.
+By default, LXD uses the `quota` property when you set up a size/quota for your storage volume.
```
tomponline, line 89: why did we use size/quota here rather than brackets?

tomponline (approved):

PR #14631, opened 2024-12-10
Per-project uplink IP quotas

hamistao:
Introduces per-network project uplink IP limits, adding a `limits.networks.uplink_ips.NETWORK_NAME` configuration key to projects.
This config key defines the maximum value of IPs made available on a network named NETWORK_NAME to be assigned as uplink IPs for entities inside a cetain project.

### DIFF
--- a/lxd/api_project.go
+++ b/lxd/api_project.go
@@ -294,7 +294,7 @@ func projectsPost(d *Daemon, r *http.Request) response.Response {
 	}
 
 	// Validate the configuration.
-	err = projectValidateConfig(s, project.Config)
+	err = projectValidateConfig(s, project.Config, project.Name)
 	if err != nil {
 		return response.BadRequest(err)
 	}
@@ -682,7 +682,7 @@ func projectChange(s *state.State, project *api.Project, req api.ProjectPut) res
 	}
 
 	// Validate the configuration.
-	err := projectValidateConfig(s, req.Config)
+	err := projectValidateConfig(s, req.Config, project.Name)
 	if err != nil {
 		return response.BadRequest(err)
 	}
@@ -1024,7 +1024,7 @@ func isEitherAllowOrBlockOrManaged(value string) error {
 	return validate.Optional(validate.IsOneOf("block", "allow", "managed"))(value)
 }
 
-func projectValidateConfig(s *state.State, config map[string]string) error {
+func projectValidateConfig(s *state.State, config map[string]string, projectName string) error {
 	// Validate the project configuration.
 	projectConfigKeys := map[string]func(value string) error{
 		// lxdmeta:generate(entities=project; group=specific; key=backups.compression_algorithm)
--- a/lxd/api_project.go
+++ b/lxd/api_project.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 
 	"github.com/gorilla/mux"
@@ -950,6 +951,54 @@ func projectStateGet(d *Daemon, r *http.Request) response.Response {
 	}
 
 	return response.SyncResponse(true, &state)
+}
+
+// uplinkIPLimitValidator returns a validator function for uplink IP limits.
+// The protocol argument specifies whether we should validate ipv4 or ipv6.
+func uplinkIPLimitValidator(s *state.State, projectName string, networkName string, protocol string) func(string) error {
+	return func(value string) error {
+		// Perform cheaper checks on the value first.
+		providedIPQuota, err := strconv.Atoi(value)
+		if err != nil {
+			return err
+		}
+
+		if providedIPQuota < 0 {
+			return fmt.Errorf("Value must be non-negative")
+		}
+
+		// The results for the quota we are not interested in will be ignored in the end, so
+		// here -1 is used to prevent the quota that is not relevant from stopping UplinkAddressQuotasExceeded
+		// from returning early.
+		IPV4AddressQuota := -1
+		IPV6AddressQuota := -1
+
+		if protocol == "ipv6" {
+			IPV6AddressQuota = providedIPQuota
+		}
+
+		if protocol == "ipv4" {
+			IPV4AddressQuota = providedIPQuota
+		}
+
+		// Check if the provided value is equal or lower to the number of uplink addresses currently in use
+		// on the provided project and in the specified network.
+		// We are only interested on the result for the desired protocol, the other will always come out as true.
+		err = s.DB.Cluster.Transaction(s.ShutdownCtx, func(ctx context.Context, tx *db.ClusterTx) error {
+			invalidIPV4Quota, invalidIPV6Quota, err := limits.UplinkAddressQuotasExceeded(ctx, tx, projectName, networkName, IPV4AddressQuota, IPV6AddressQuota)
+			if err != nil {
+				return err
+			}
+
+			if protocol == "ipv4" && invalidIPV4Quota || protocol == "ipv6" && invalidIPV6Quota {
+				return fmt.Errorf("Uplink %s limit %q is below current number of used uplink addresses", protocol, value)
+			}
+
+			return nil
+		})
+
+		return err
+	}
 }
 
 // Check if a project is empty.
--- a/lxd/api_project.go
+++ b/lxd/api_project.go
@@ -1463,6 +1463,36 @@ func projectValidateConfig(s *state.State, config map[string]string, projectName
 		return fmt.Errorf("Failed loading storage pool names: %w", err)
 	}
 
+	// Per-network project limits for uplink IPs only make sense for projects with their own networks.
+	if shared.IsTrue(config["features.networks"]) {
+		// Get networks that are allowed to be used as uplinks by this project.
+		allowedUplinkNetworks, err := network.AllowedUplinkNetworks(s, config)
+		if err != nil {
+			return err
+		}
+
+		// Add network-specific config keys.
+		for _, networkName := range allowedUplinkNetworks {
+			// lxdmeta:generate(entities=project; group=limits; key=limits.networks.uplink_ips.ipv4.NETWORK_NAME)
+			// Maximum number of IPv4 addresses that this project can consume from the specified uplink network.
+			// This number of IPs can be consumed by networks, forwards and load balancers in this project.
+			//
+			// ---
+			//  type: string
+			//  shortdesc: Quota of IPv4 addresses from a specified uplink network that can be used by entities in this project
+			projectConfigKeys["limits.networks.uplink_ips.ipv4."+networkName] = validate.Optional(uplinkIPLimitValidator(s, projectName, networkName, "ipv4"))
+
+			// lxdmeta:generate(entities=project; group=limits; key=limits.networks.uplink_ips.ipv6.NETWORK_NAME)
+			// Maximum number of IPv6 addresses that this project can consume from the specified uplink network.
+			// This number of IPs can be consumed by networks, forwards and load balancers in this project.
+			//
+			// ---
+			//  type: string
+			//  shortdesc: Quota of IPv4 addresses from a specified uplink network that can be used by entities in this project
+			projectConfigKeys["limits.networks.uplink_ips.ipv6."+networkName] = validate.Optional(uplinkIPLimitValidator(s, projectName, networkName, "ipv6"))
+		}
+	}
+
 	for k, v := range config {
 		key := k
 
--- a/lxd/network/driver_ovn.go
+++ b/lxd/network/driver_ovn.go
@@ -1232,15 +1232,47 @@ func (n *ovn) allocateUplinkPortIPs(uplinkNet Network, routerMAC net.HardwareAdd
 	routerExtPortIPv4 := net.ParseIP(n.config[ovnVolatileUplinkIPv4])
 	routerExtPortIPv6 := net.ParseIP(n.config[ovnVolatileUplinkIPv6])
 
+	// Get project's config.
+	var p *api.Project
+	var ipv4QuotaAvailable bool
+	var ipv6QuotaAvailable bool
+
+	err = n.state.DB.Cluster.Transaction(n.state.ShutdownCtx, func(ctx context.Context, tx *db.ClusterTx) error {
+		dbProject, err := dbCluster.GetProject(ctx, tx.Tx(), n.project)
+		if err != nil {
+			return err
+		}
+
+		p, err = dbProject.ToAPI(ctx, tx.Tx())
+		if err != nil {
+			return err
+		}
+
+		// Check if we have quota available for the addresses we want to allocate for the new network.
+		ipv4QuotaAvailable, ipv6QuotaAvailable, err = n.projectUplinkIPQuotaAvailable(ctx, tx, p, uplinkNet.Name())
+		return err
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	allocatingIPV4 := uplinkIPv4Net != nil && routerExtPortIPv4 == nil
+	allocatingIPV6 := uplinkIPv6Net != nil && routerExtPortIPv6 == nil
+
+	// A quota check result is only relevant if we intent to allocate an IP fot that quota's protocol.
+	if allocatingIPV4 && !ipv4QuotaAvailable || allocatingIPV6 && !ipv6QuotaAvailable {
+		return nil, fmt.Errorf("Project quota for uplink IPs on network %q is exhausted", uplinkNet.Name())
+	}
+
 	// Decide whether we need to allocate new IP(s) and go to the expense of retrieving all allocated IPs.
-	if (uplinkIPv4Net != nil && routerExtPortIPv4 == nil) || (uplinkIPv6Net != nil && routerExtPortIPv6 == nil) {
+	if (allocatingIPV4) || (allocatingIPV6) {
 		err := n.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
 			allAllocatedIPv4, allAllocatedIPv6, err := n.uplinkAllAllocatedIPs(ctx, tx, uplinkNet.Name())
 			if err != nil {
 				return fmt.Errorf("Failed to get all allocated IPs for uplink: %w", err)
 			}
 
-			if uplinkIPv4Net != nil && routerExtPortIPv4 == nil {
+			if allocatingIPV4 {
 				if uplinkNetConf["ipv4.ovn.ranges"] == "" {
 					return fmt.Errorf(`Missing required "ipv4.ovn.ranges" config key on uplink network`)
 				}
@@ -1267,7 +1299,7 @@ func (n *ovn) allocateUplinkPortIPs(uplinkNet Network, routerMAC net.HardwareAdd
 				n.config[ovnVolatileUplinkIPv4] = routerExtPortIPv4.String()
 			}
 
-			if uplinkIPv6Net != nil && routerExtPortIPv6 == nil {
+			if allocatingIPV6 {
 				// If IPv6 OVN ranges are specified by the uplink, allocate from them.
 				if uplinkNetConf["ipv6.ovn.ranges"] != "" {
 					dhcpSubnet := uplinkNet.DHCPv6Subnet()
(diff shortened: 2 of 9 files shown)


lxd/api_project.go
```diff
@@ -1414,6 +1415,74 @@ func projectValidateConfig(s *state.State, config map[string]string) error {
...
+			if err != nil {
+				return err
+			}
+
+			if providedLimit < 0 {
+				return fmt.Errorf("Value must be non-negative")
+			}
+
+			// Check if the provided value is equal or lower to the number of uplink addresses currently in use
+			// on the providided project and in the specified network.
```
minaelee, line 1466:
```suggestion
			// on the provided project and in the specified network.
```

lxd/network/driver_ovn.go
```diff
@@ -4759,90 +4773,138 @@ func (n *ovn) forwardFlattenVIPs(listenAddress net.IP, defaultTargetAddress net.
 	return vips
 }
 
-// ForwardCreate creates a network forward.
-func (n *ovn) ForwardCreate(forward api.NetworkForwardsPost, clientType request.ClientType) (net.IP, error) {
-	revert := revert.New()
-	defer revert.Fail()
+// projectUplinkIPQuotaAvailable checks if a project has quota available to assingn new uplink IPs in a certain network.
```
minaelee, line 4776:
```suggestion
// projectUplinkIPQuotaAvailable checks if a project has quota available to assign new uplink IPs in a certain network.
```

lxd/network/driver_ovn.go
```diff
@@ -4759,90 +4773,138 @@ func (n *ovn) forwardFlattenVIPs(listenAddress net.IP, defaultTargetAddress net.
...
+func (n *ovn) projectUplinkIPQuotaAvailable(p *api.Project, uplinkName string) (bool, error) {
+	uplinkIPLimitKey := fmt.Sprintf("limits.networks.uplink_ips.%s", uplinkName)
+	uplinkLimitValue, ok := p.Config[uplinkIPLimitKey]
 
-	// Convert listen address to subnet so we can check its valid and can be used.
-	listenAddressNet, err := ParseIPToNet(forward.ListenAddress)
-	if err != nil {
-		return nil, fmt.Errorf("Failed parsing %q: %w", forward.ListenAddress, err)
+	// If no quota is defined for that network return right away.
```
minaelee, line 4781:
```suggestion
	// If no quota is defined for that network, return right away.
```

lxd/network/driver_ovn.go
```diff
@@ -4759,90 +4773,138 @@ func (n *ovn) forwardFlattenVIPs(listenAddress net.IP, defaultTargetAddress net.
...
+	uplinkIPLimit, err := strconv.Atoi(uplinkLimitValue)
+	if err != nil {
+		return false, err
+	}
 
-		// Load the project to get uplink network restrictions.
-		var p *api.Project
-		var uplink *api.Network
+	// If the current value of used uplink IPs surpasses the limit-1, that means there is not room for one more IP.
+	// Use network caching to avoid quetying for the same networks twice.
```
minaelee, line 4793:
```suggestion
	// Use network caching to avoid querying for the same networks twice.
```

lxd/network/driver_ovn.go
```diff
@@ -4759,90 +4773,138 @@ func (n *ovn) forwardFlattenVIPs(listenAddress net.IP, defaultTargetAddress net.
...
+	var uplinkIPLimit int
+	uplinkIPLimit, err := strconv.Atoi(uplinkLimitValue)
+	if err != nil {
+		return false, err
+	}
 
-		// Load the project to get uplink network restrictions.
-		var p *api.Project
-		var uplink *api.Network
+	// If the current value of used uplink IPs surpasses the limit-1, that means there is not room for one more IP.
```
minaelee, line 4792:
```suggestion
	// If the current value of used uplink IPs surpasses the limit-1, that means there is no room for any more IPs.
```

lxd/network/driver_ovn.go
```diff
@@ -4759,90 +4773,138 @@ func (n *ovn) forwardFlattenVIPs(listenAddress net.IP, defaultTargetAddress net.
...
-			ipVersion := 4
-			if forward.ListenAddress == net.IPv6zero.String() {
-				ipVersion = 6
-			}
+		return listenAddressNet.IP.String(), nil
+	}
 
-			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-			defer cancel()
+	// Check the listen address subnet is allowed within both the uplink's external routes and any
```
minaelee, line 4872:
```suggestion
	// Check if the listen address subnet is allowed within both the uplink's external routes and any
```

hamistao:
@minaelee Thanks for the early review, every comment was addressed as you suggested.

minaelee:
> @minaelee Thanks for the early review, every comment was addressed as you suggested.

Thanks! Sorry, I missed that it was marked as Draft.  Looks great!

hamistao:
@tomponline @markylaing Some observations on this:
- I did not consider bridge network forwards as something that is taking away IPs from another network, as it would be more tricky to check if they are taking IPs from other managed networks and they are exclusive to the default project. If needed we can add this check in the future.
- Project uplink IP limits are exclusive to projects with `features.networks`
- I am not currently distinguishing between ipv6 and ipv4. Maybe I am not 100% clear on the details of the use case of this feature, as that could help decide how to approach the different protocols.
- Tests are in the works and should be added to lxd CI shortly.

Manual tests are looking fine so far so I am opening this for review, feel free to look whenever you are able.

tomponline:
> I am not currently distinguishing between ipv6 and ipv4. Maybe I am not 100% clear on the details of the use case of this feature, as that could help decide how to approach the different protocols.

I think we should have separate ipv4 and ipv6 quotas as the routes on the uplinks are defined per protocol.

tomponline:
> * I did not consider bridge network forwards as something that is taking away IPs from another network, as it would be more tricky to check if they are taking IPs from other managed networks and they are exclusive to the default project. If needed we can add this check in the future.

I'm not quite following here. We aren't checking if IPs are being taken away from other networks, but rather whether the quota for the project that the IP usage has exceeded the quota.

For any managed networks in the default project they should still be limited by the quota set on the default project (which is likely to be nothing).

> Project uplink IP limits are exclusive to projects with features.networks

I think that probably makes sense. The default project has features.networks so that should be fine.

hamistao:
> I'm not quite following here. We aren't checking if IPs are being taken away from other networks, but rather whether the quota for the project that the IP usage has exceeded the quota.

Let's make this a topic for our one to one tomorrow so I can better explain.

hamistao:
Going back to driver to implement suggested changes.

lxd/network/network_utils.go
```diff
@@ -1463,6 +1463,11 @@ func ProxyParseAddr(data string) (*deviceConfig.ProxyAddress, error) {
 func AllowedUplinkNetworks(s *state.State, projectConfig map[string]string) ([]string, error) {
 	var uplinkNetworkNames []string
 
+	// There are no allowed networks if project is restricted and restricted.networks.uplinks is not set.
```
tomponline, line 1466: ideally this would have been in a separate PR so we can merge and test it earlier.

lxd/api_project.go
```diff
@@ -952,6 +953,48 @@ func projectStateGet(d *Daemon, r *http.Request) response.Response {
...
+		if err != nil {
+			return err
+		}
+
+		if providedIPQuota < 0 {
+			return fmt.Errorf("Value must be non-negative")
+		}
+
+		// Populate quota variables accordingly.
+		IPV4AddressQuota := -1
```
tomponline, line 971: I assume -1 is unlimited, but would be good to clarify that via a comment/constant.

lxd/api_project.go
```diff
@@ -952,6 +953,48 @@ func projectStateGet(d *Daemon, r *http.Request) response.Response {
...
+		// Check if the provided value is equal or lower to the number of uplink addresses currently in use
+		// on the provided project and in the specified network.
+		// We are only interested on the result for the desired protocol, the other will always come out as true.
+		invalidIPV4Quota, invalidIPV6Quota, err := limits.UplinkAddressQuotasExceeded(s, projectName, networkName, IPV4AddressQuota, IPV6AddressQuota)
+		if err != nil {
+			return err
+		}
+
+		if protocol == "ipv4" && invalidIPV4Quota || protocol == "ipv6" && invalidIPV6Quota {
+			return fmt.Errorf("Value %s is below current number of used uplink addresses", value)
```
tomponline, line 991: We use %q for user inputted strings

tomponline: Also "Value" here is a bit vague, can we say "Uplink IPv4 limit %q is below current number of used uplink address"

lxd/api_project.go
```diff
@@ -1457,6 +1457,38 @@ func projectValidateConfig(s *state.State, config map[string]string, projectName
 		return fmt.Errorf("Failed loading storage pool names: %w", err)
 	}
 
+	// Per-network project limits for uplink IPs only make sense for projects with their own networks.
+	if shared.IsTrue(config["features.networks"]) {
+		// Get networks that are allowed to be used as uplinks by this project.
+		allowedUplinkNetworks, err := network.AllowedUplinkNetworks(s, config)
+		if err != nil {
+			return nil
```
tomponline, line 1465: return nil seems wrong here?

lxd/api_project.go
```diff
@@ -1457,6 +1457,38 @@ func projectValidateConfig(s *state.State, config map[string]string, projectName
...
+		}
+
+		// Add network-specific config keys.
+		for _, networkName := range allowedUplinkNetworks {
+			// lxdmeta:generate(entities=project; group=limits; key=limits.networks.uplink_ips.ipv4.NETWORK_NAME)
+			// This represents the maximum value for IPV4 addresses available on a network
```
tomponline, line 1471: I think this can be clarified a bit, how about

> Maximum number of IPv4 addresses that this project can consume from the specified uplink network. These IPs can be consumed by networks, forwards and load-balancers.

as "maximum value" is a bit vague IMHO.

lxd/api_project.go
```diff
@@ -1457,6 +1457,38 @@ func projectValidateConfig(s *state.State, config map[string]string, projectName
...
+			// named NETWORK_NAME to be assigned as uplink addresses for entities inside
+			// a specific project.
+			//
+			// ---
+			//  type: string
+			//  shortdesc: Quota of IPs on a certain network used by entities on this project
```
tomponline, line 1477: I think this can be clarified a bit, how about:

> Quota of IPv4 addresses from a specified uplink network that can be used by entities in this project

The key bits I changed were changing "on" to "from" to avoid confusion around where the quota was counting resources from.

Also changed "on this project" to "in this project", as resources go "into" a project not "onto" them.

lxd/api_project.go
```diff
@@ -1489,6 +1489,26 @@ func projectValidateConfig(s *state.State, config map[string]string, projectName
 		}
 	}
 
+	// If project does not have access to a network and have limits defined for it, unset those limits
+	// before validation.
+	// These changes should be persisted on the database.
+	if shared.IsTrue(config["restricted"]) {
```
tomponline, line 1495: I would prefer to see this as an earlier validation step that prevents setting `limits.networks.uplink_ips.*.UPLINK` where the uplink isn't accessible in the project, rather than silently deleting the user specified key.

lxd/network/driver_ovn.go
```diff
@@ -4725,6 +4725,84 @@ func (n *ovn) forwardFlattenVIPs(listenAddress net.IP, defaultTargetAddress net.
 	return vips
 }
 
+// triageExternalAddressUsage performs common steps between creating network forwards and load balancers.
+// This includes validating the provided listen address or auto-allocating one if needed.
+func (n *ovn) triageExternalAddressUsage(listenAddressNet *net.IPNet, proposedAddress string) (*net.IPNet, string, error) {
```
tomponline: can we call this `alocateUplinkAddress`?

Also please can you define in the doc block what the input arguments are, as its not obvious to me. Thanks

tomponline: from what I can see in the usage in the next commit, the `listenAddresNet` and `proposedAddress` arguments are the same value in different formats (the requested listen address or empty/undefined for auto allocation) and the return values are the same value in different formats.

This seems quite confusing to me.

Can we accept just a `listenAddresNet  *net.IPNet`  argument here and return `(net.IP, error)`, and then we can update `forward.ListenAddress` as needed by converting the returned `ip.Net` to a string.

Otherwise we are blurring the lines between allocation & validation logic and formatting logic.

lxd/network/driver_ovn.go
```diff
@@ -4766,6 +4767,19 @@ func (n *ovn) triageExternalAddressUsage(listenAddressNet *net.IPNet, proposedAd
...
+	ipv4QuotaAvailable, ipv6QuotaAvailable, err := n.projectUplinkIPQuotaAvailable(p, uplink.Name)
+	if err != nil {
+		return nil, "", err
+	}
+
+	usingIPV6 := validate.IsNetworkAddressV6(proposedAddress) == nil
+
+	// If there is no quota available for the required protocol, return an error.
+	if usingIPV6 && !ipv6QuotaAvailable || !usingIPV6 && !ipv4QuotaAvailable {
+		return nil, "", fmt.Errorf("Required project %s's quota for uplink IPs on network %s is exhausted", p.Name, uplink.Name)
```
tomponline: I dont think we need to include the project name in this error.

lxd/network/driver_ovn.go
```diff
@@ -1232,15 +1232,44 @@ func (n *ovn) allocateUplinkPortIPs(uplinkNet Network, routerMAC net.HardwareAdd
...
+	ipv4QuotaAvailable, ipv6QuotaAvailable, err := n.projectUplinkIPQuotaAvailable(p, uplinkNet.Name())
+	if err != nil {
+		return nil, err
+	}
+
+	allocatingIPV4 := uplinkIPv4Net != nil && routerExtPortIPv4 == nil
+	allocatingIPV6 := uplinkIPv6Net != nil && routerExtPortIPv6 == nil
+
+	if allocatingIPV4 && !ipv4QuotaAvailable || allocatingIPV6 && !ipv6QuotaAvailable {
+		return nil, fmt.Errorf("Required project %s's quota for uplink IPs on network %s is exhausted", p.Name, uplinkNet.Name())
```
tomponline, line 1261: I dont think we need to include the project name in this error.

tomponline: network name should be quoted using %q

lxd/network/driver_ovn.go
tomponline, line 1: this commit looks a bit confused, there's a lot of changes in it which i think could do with being broken up into a logical order.

lxd/project/limits/permissions.go
tomponline, line 1: lets put this commit, plus the 2 earlier unrelated changes (the fail fast check commit, and then de-dupe logic on allocating forward IP commit) into a separate PR to simplify this PR.

lxd/project/limits/permissions.go
```diff
@@ -471,21 +471,22 @@ func checkAggregateLimits(info *projectInfo, aggregateKeys []string) error {
 
 // parseHostIDMapRange parse the supplied list of host ID map ranges into a idmap.IdmapEntry slice.
 func parseHostIDMapRange(isUID bool, isGID bool, listValue string) ([]idmap.IdmapEntry, error) {
-	var idmaps []idmap.IdmapEntry
+	mapRanges := shared.SplitNTrimSpace(listValue, ",", -1, true)
+	idmaps := make([]idmap.IdmapEntry, len(mapRanges))
```
tomponline, line 475: We tend to use the `make([]idmap.IdmapEntry, 0, len(mapRanges))` initialization style so that we create the underlying slice with sufficient capacity to avoid extra allocations, but dont initialise the actual slice elements, and instead use `append`. 

That way if we ever alter the logic to skip certain elements, we dont end up introducing a bug where certain pre-initialised elements are left present with their zero value.

tomponline (commented):
Please ensure you're using `ShutdownCtx` in your transactions rather than `context.TODO()` when retrieving data from the DB.

tomponline (commented):
Is there a possibility of passing a transaction argument to UplinkAddressQuotasExceeded so that we can validate the uplink quotas in a single transaction? To avoid concurrent requests potentially allowing the quota to be exceeded?

tomponline:
@hamistao as this change is all about validation and logic rather than actually testing end-to-end OVN functionality, I think the tests for this would be well suited to be in the main lxd repo, rather than in lxd-ci https://github.com/canonical/lxd-ci/pull/390 as @markylaing has now added support for OVN tests in the main lxd repo.

lxd/api_project.go
hamistao: The problem of having this as a validation step is when a user has uplink IP limits set and then wants to restrict a project. Having to unset your limits before setting restricted seems like bad UX to me.

hamistao: But my approach has a problem of silently denying to set limits on a restricted project instead of showing an erro. I will try and come up with something better that checks both boxes.

lxd/project/limits/permissions.go
hamistao: Makes perfect sense, thanks for the insight!

lxd/api_project.go
tomponline: > Having to unset your limits before setting restricted seems like bad UX to me.

This seems fine to me as we are requiring the project's config is consistent.

tomponline: But we probably also need a check to ensure that if not existing already, that ensures that any existing resources are only using allowed uplinks in restricted mode.

hamistao: Sorry, could you perhaps rephrase this last message to help me understand?

tomponline: You said:

> The problem of having this as a validation step is when a user has uplink IP limits set and then wants to restrict a project. Having to unset your limits before setting restricted seems like bad UX to me.

My assumption there was that you meant limits on uplinks that weren't yet allowed in the project would be prevented from being set when restricted=true.

But that is OK, thats proper validation, if you can to set restricted=true then you should also set the restricted uplinks to the ones that have uplink limits on too.

But equally we should ensure that even if there are no uplink quota limits set, that if restricted=true that the project doesn't have any resources in it that are using uplinks not specified in the allowed uplinks list.

hamistao: Cool, got it! Thanks for clarifying

lxd/project/limits/permissions.go
hamistao: Would you be ok with just dropping the fail fast check commit?
This commit is here to avoid linter errors and the de-dupe logic on allocating forward IPs, although not necessary, is useful so I can just add the quota check logic in one place for both forwards and load balancers.

lxd/api_project.go
hamistao: > But equally we should ensure that even if there are no uplink quota limits set, that if restricted=true that the project doesn't have any resources in it that are using uplinks not specified in the allowed uplinks list.

@tomponline As a follow up to this PR, I am getting another PR ready to include this check that is indeed not currently performed. In the meantime, this is rebased and ready for review.

tomponline (approved):
Thanks!

lxd/api_project.go
```diff
@@ -1463,6 +1463,36 @@ func projectValidateConfig(s *state.State, config map[string]string, projectName
...
+			projectConfigKeys["limits.networks.uplink_ips.ipv4."+networkName] = validate.Optional(uplinkIPLimitValidator(s, projectName, networkName, "ipv4"))
+
+			// lxdmeta:generate(entities=project; group=limits; key=limits.networks.uplink_ips.ipv6.NETWORK_NAME)
+			// Maximum number of IPv6 addresses that this project can consume from the specified uplink network.
+			// This number of IPs can be consumed by networks, forwards and load balancers in this project.
+			//
+			// ---
+			//  type: string
+			//  shortdesc: Quota of IPv4 addresses from a specified uplink network that can be used by entities in this project
+			projectConfigKeys["limits.networks.uplink_ips.ipv6."+networkName] = validate.Optional(uplinkIPLimitValidator(s, projectName, networkName, "ipv6"))
```
MusicDin, line 1492: Caught while writing weekly news, there is a typo in short description: `IPv4` -> `IPv6`

hamistao: Thanks for catching this, will fix in a minute

PR #14633, opened 2024-12-10
storage: Disallow `volume.security.shared` on cephfs 

hamistao:
This adds a check to deny enabling `security.shared` on a cephfs pool. Also includes tests for this check and for setting `security.shared` in many volume types.

### DIFF
--- a/test/suites/storage_driver_cephfs.sh
+++ b/test/suites/storage_driver_cephfs.sh
@@ -11,10 +11,10 @@ test_storage_driver_cephfs() {
   lxc storage delete cephfs
 
   # Test invalid key combinations for auto-creation of cephfs entities.
-  ! lxc storage create cephfs cephfs source="${LXD_CEPH_CEPHFS}/$(basename "${LXD_DIR}")" cephfs.osd_pg_num=32 || true
-  ! lxc storage create cephfs cephfs source="${LXD_CEPH_CEPHFS}/$(basename "${LXD_DIR}")" cephfs.meta_pool=xyz || true
-  ! lxc storage create cephfs cephfs source="${LXD_CEPH_CEPHFS}/$(basename "${LXD_DIR}")" cephfs.data_pool=xyz || true
-  ! lxc storage create cephfs cephfs source="${LXD_CEPH_CEPHFS}/$(basename "${LXD_DIR}")" cephfs.create_missing=true cephfs.data_pool=xyz_data cephfs.meta_pool=xyz_meta || true
+  ! lxc storage create cephfs cephfs source="${LXD_CEPH_CEPHFS}/$(basename "${LXD_DIR}")" cephfs.osd_pg_num=32 || false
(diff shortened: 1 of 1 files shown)


hamistao:
@masnax Could you please just look at the last commit and confirm if my changes there make sense?

lxd/storage/drivers/driver_cephfs.go
```diff
@@ -437,6 +437,8 @@ func (d *cephfs) Validate(config map[string]string) error {
 		//  shortdesc: Whether the CephFS file system was empty on creation time
 		//  scope: global
 		"volatile.pool.pristine": validate.IsAny,
+		// Override general 'security.shared' rule since only cephfs does not support block volumes.
+		"volume.security.shared": func(value string) error { return fmt.Errorf("") },
```
tomponline, line 441: empty error message looks odd?

test/suites/storage_local_volume_handling.sh
```diff
@@ -66,6 +66,18 @@ test_storage_local_volume_handling() {
     lxc storage volume set "${pool}" vol1 user.foo=snap0
     lxc storage volume set "${pool}" vol1 snapshots.expiry=1H
 
+    lxc storage volume create "${pool}" blockVol --type=block
+    truncate -s 25MiB foo.iso
+    lxc storage volume import "${pool}" ./foo.iso isoVol
+
+    # security.shared is only allowed for block volumes
+    ! lxc storage volume set "${pool}" vol1 security.shared true || false
+    ! lxc storage volume set "${pool}" isoVol security.shared true || false
```
tomponline, line 75: So does `security.shared` not work for iso volumes because they are allowed to be attached concurrently to multiple instances (as this should be allowed because they are readonly)?

test/suites/storage_driver_cephfs.sh
tomponline, line 1: Please can you expand the commit message as to why these were ineffective?

test/suites/storage_local_volume_handling.sh
hamistao: Yes, they are sharable by default

tomponline:
@hamistao ceph tests failing

```
+ lxc storage create cephfs cephfs source=cephfs/Q5A
++ timeout --foreground 120 /home/runner/go/bin/lxc storage create cephfs cephfs source=cephfs/Q5A --verbose
Error: Invalid value for option "volume.security.shared": 
```

lxd/storage/drivers/driver_cephfs.go
hamistao: Whoops, that was my mistake, fixed

test/suites/storage_driver_cephfs.sh
hamistao: Done!

hamistao:
@tomponline Tests fixed and comments addressed

PR #14642, opened 2024-12-11
Check default osd pool size before setting on create.

masnax:
Fixes `lxd-ci` pipeline errors from `mon_allow_pool_size_one` not being set


tomponline (approved):
Great thanks!

PR #14643, opened 2024-12-12
doc: fix lxc network load-balancer backend add command syntax

minaelee:
This PR removes an unnecessary <listen_ports> option in the syntax for `lxc network load-balancer backend add`. This is correct in [the man page](https://documentation.ubuntu.com/lxd/en/latest/reference/manpages/lxc/network/load-balancer/backend/add/) but incorrect in [the how-to guide this PR fixes](https://documentation.ubuntu.com/lxd/en/latest/howto/network_load_balancers/#configure-backends). There is no `listen_ports` backend property. `listen_ports` is a port property that is used in `lxc network load-balancer port add`.

### DIFF
--- a/doc/howto/network_load_balancers.md
+++ b/doc/howto/network_load_balancers.md
@@ -59,7 +59,7 @@ The backend target address must be within the same subnet as the network that th
 Use the following command to add a backend specification:
 
 ```bash
-lxc network load-balancer backend add <network_name> <listen_address> <backend_name> <listen_ports> <target_address> [<target_ports>]
+lxc network load-balancer backend add <network_name> <listen_address> <backend_name> <target_address> [<target_ports>]
 ```
 
 The target ports are optional.


minaelee:
@tomponline: I'm also concerned about [this line](https://github.com/canonical/lxd/blob/bf56b737514d63d28a68afac0a532dfa24bef7e3/doc/howto/network_load_balancers.md?plain=1#L66) that appears in the docs, right under the syntax for `lxc network load-balancer backend add`:

>The target ports are optional. **If not specified, the load balancer will use the listen ports for the backend for the backend target ports.**

Since we've established that there are no listen ports defined on the backend configuration, the phrase **listen ports for the backend** doesn't seem applicable. Should this be **the listen ports defined in the load balancer port properties** instead? 

[This line in the source](https://github.com/canonical/lxd/blob/main/shared/api/network_load_balancer.go#L41) seems to indicate that is the case. That the target ports, if not specified, fall back to the listen ports on the `NetworkLoadBalancerPort`.

If that's true, would I be correct to assume that whenever a backend is created, if no port options (such as `listen_port`) are configured explicitly, there is a default set of port options configured on the load balancer that includes a `listen_port`?

tomponline (approved):

tomponline:
> @tomponline: I'm also concerned about [this line](https://github.com/canonical/lxd/blob/bf56b737514d63d28a68afac0a532dfa24bef7e3/doc/howto/network_load_balancers.md?plain=1#L66) that appears in the docs, right under the syntax for `lxc network load-balancer backend add`:
> 
> > The target ports are optional. **If not specified, the load balancer will use the listen ports for the backend for the backend target ports.**
> 
> Since we've established that there are is no listen_port property of the load balancer backend, only of the load balancer port, the phrase **listen ports for the backend** doesn't seem applicable. Should this read **...the listen ports defined in the load balancer port properties...** instead?
> 
> [This line in the source](https://github.com/canonical/lxd/blob/main/shared/api/network_load_balancer.go#L41) seems to indicate that is the case. That the target ports, if not specified, fall back to the listen ports on the `NetworkLoadBalancerPort`.
> 
> If that's true, would I be correct to assume that whenever a backend is created, if no port options (such as `listen_port`) are configured explicitly, there is a default set of port options configured on the load balancer that includes a `listen_port`?

Yeah I agree.

Suggest something like

> The target ports are optional. If not specified, the load balancer listener ports will be used for the backend target ports.


> If that's true, would I be correct to assume that whenever a backend is created, if no port options (such as listen_port) are configured explicitly, there is a default set of port options configured on the load balancer that includes a listen_port?

No thats not correct, creating a load balancer simply reserves a listen IP (and no ports) then listen ports are added using:

```
lxc network load-balancer port add
```

minaelee:
> > If that's true, would I be correct to assume that whenever a backend is created, if no port options (such as listen_port) are configured explicitly, there is a default set of port options configured on the load balancer that includes a listen_port?
> 
> No thats not correct, creating a load balancer simply reserves a listen IP (and no ports) then listen ports are added using:
> 
> ```
> lxc network load-balancer port add
> ```

What is the behavior if the `lxc network load-balancer backend add` command is run without specifying target ports _and_ no listen ports have yet been added via the `...port add` command? Is there another fallback in that case, or will it simply not specify any target ports?

PR #14644, opened 2024-12-12
lxd: Deny creation of empty ISO volume

hamistao:
I noticed `lxc storage volume create "${pool}" isoVol --type=iso` worked inconsistently across drivers. Which led me to see that, according to the docs: `A custom storage volume of type `iso` can only be created by importing an ISO file using`, which makes sense because an empty ISO volume is useless AFAIK. So this implements this prohibition of creating empty ISO volumes.


tomponline (approved):

tomponline:
Nice catch

PR #14650, opened 2024-12-13
Fix typo: Reusuable -> Reusable

simondeziel:
I added some "quality of life" improvement to our `make static-analysis` target after looking at the `revive` error flagged during CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -241,13 +241,13 @@ po/%.mo: po/%.po
 	msgfmt --statistics -o $@ $<
 
 po/%.po: po/$(DOMAIN).pot
-	msgmerge -U po/$*.po po/$(DOMAIN).pot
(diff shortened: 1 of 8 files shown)


tomponline (approved):

tomponline:
> is it OK to do such typo fix for an exported variable or could that affect downstream users?

should be fine as not part of our shared or api packages.

tomponline:
> regarding the CodeQL complaint (https://github.com/canonical/lxd/pull/14650/checks?check_run_id=34387757044) about missing config for Python, is there some cleanup you can do with your higher privs than mine? Or should we simply re-introduce CodeQL scanning for Python even if we have ~0.

I cant see anything in admin section that allows specifying languages to scan for, it just points to the yaml file in our repo for config.

PR #14654, opened 2024-12-14
Doc: Improve network load-balancer instructions and add examples

minaelee:
The existing doc says that if target ports are not provided, "the listen ports for the backend" will be used. This is incorrect (or at least, incomplete) as the backend specification does not define listen ports; the ports specification associated with that backend does. This PR clarifies and expands on how this works, including what happens if target ports are not provided _and_ no ports specification associated with that backend exists.

Also:
* Adds command syntax examples
* Removes redundancy in explaining how allocate flag is used
* Minor grammar/formatting improvements
* Shows example YAML configuration file

### DIFF
--- a/doc/howto/network_load_balancers.md
+++ b/doc/howto/network_load_balancers.md
@@ -9,9 +9,9 @@ discourse: lxc:14317
 Network load balancers are currently available for the {ref}`network-ovn`.
 ```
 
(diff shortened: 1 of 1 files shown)


tomponline (approved):

tomponline:
doc tests are failing

minaelee:
> doc tests are failing

Fixed, thanks!

tomponline (approved):

PR #14651, opened 2024-12-13
OVN tests for main test suite

markylaing:
This PR adds OVN networking tests to the main suite. This is primarily to give faster feedback when driver changes are made. This initial PR just validates that the configuration is applied correctly to the ovn northbound database. The `LXD_OVN_NB_CONNECTION` environment variable must be set to run the tests. The tests will fail if there is any existing data in the northbound database. The test expects that `ovn-nbctl` is available on the PATH.

PR #14670, opened 2024-12-16
Add `--format` flag to `lxd sql`

markylaing:
This allows specifying the output of `lxd sql` as `sql`, `csv`, `table`, `compact`, `json`, or `yaml`. The new `sql` format is added so that the column names are not auto-formatted (upper cased, replacing underscores with spaces).

This is added to ease scripting in tests and is required for #14651.

### DIFF
--- a/lxd/main_sql.go
+++ b/lxd/main_sql.go
@@ -161,7 +161,7 @@ func sqlPrintSelectResult(format string, result internalSQLResult) error {
 	for _, row := range result.Rows {
 		r := make([]string, 0, len(row))
 		for _, col := range row {
-			r = append(r, fmt.Sprintf("%v", col))
+			r = append(r, fmt.Sprint(col))
 		}
 
 		data = append(data, r)
--- a/lxd/main_sql.go
+++ b/lxd/main_sql.go
@@ -6,15 +6,16 @@ import (
 	"io"
 	"os"
 
-	"github.com/olekukonko/tablewriter"
 	"github.com/spf13/cobra"
 
 	"github.com/canonical/lxd/client"
 	"github.com/canonical/lxd/shared"
+	cli "github.com/canonical/lxd/shared/cmd"
 )
 
 type cmdSql struct {
-	global *cmdGlobal
+	global     *cmdGlobal
+	flagFormat string
 }
 
 func (c *cmdSql) Command() *cobra.Command {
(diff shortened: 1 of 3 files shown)


lxd/main_sql.go
```diff
@@ -161,7 +161,7 @@ func sqlPrintSelectResult(format string, result internalSQLResult) error {
 	for _, row := range result.Rows {
 		r := make([]string, 0, len(row))
 		for _, col := range row {
-			r = append(r, fmt.Sprintf("%v", col))
+			r = append(r, fmt.Sprint(col))
```
tomponline, line 164: i got hit by this bug today i think, so good spot.

shared/cmd/table.go
```diff
@@ -35,6 +36,10 @@ func RenderTable(format string, header []string, data [][]string, raw any) error
 		table.SetHeaderLine(false)
 		table.SetBorder(false)
 		table.Render()
+	case TableFormatSQLResult:
+		table := getBaseTable(header, data)
+		table.SetAutoFormatHeaders(false)
```
tomponline, line 47: why dont we want headers?

lxd/main_sql.go
markylaing: This is a bug? Also credit to @simondeziel for spotting this one!

shared/cmd/table.go
markylaing: This preserves the output format of `lxd sql` as it is. For example:
```
# Existing lxd sql format
$ lxd sql global 'select * from networks'
+----+------------+--------+-------------+-------+------+
| id | project_id |  name  | description | state | type |
+----+------------+--------+-------------+-------+------+
| 6  | 1          | lxdbr0 |             | 1     | 0    |
+----+------------+--------+-------------+-------+------+

# With auto formatted headers
$ lxd sql global 'select * from networks'
+----+------------+--------+-------------+-------+------+
| ID | PROJECT ID |  NAME  | DESCRIPTION | STATE | TYPE |
+----+------------+--------+-------------+-------+------+
| 6  | 1          | lxdbr0 |             | 1     | 0    |
+----+------------+--------+-------------+-------+------+
```

lxd/main_sql.go
tomponline: yeah i think so, or something similar as was  using `%` wildcards in my query and then i was seeing the unmatched parameter message.

shared/cmd/table.go
tomponline: cheers

tomponline (approved):

PR #14674, opened 2024-12-17
test: Improve usage of `lxd sql` in tests

markylaing:
Support for the `--format` flag was added in #14670. This PR uses the format flag where possible in our test suite to avoid e.g. grepping for table formatted output.

In some cases I've changed queries to use a `SELECT count(*) FROM ...` instead of a `SELECT * FROM ...` where the output was piped into a word count or grepped with the -c flag.


simondeziel (approved):
Lovely, thanks!

tomponline (approved):

Issue #14635, opened 2024-12-10
`lxc remote switch` should suggest completion of any locally configured remote


simondeziel:
It seems that only builtin remote are proposed during tab-completion:
```
$ lxc remote switch <tab>
images                local                 ubuntu                ubuntu-daily          ubuntu-minimal        ubuntu-minimal-daily 
```
But I have many more remotes. Icing on the cake would be to remove the `current` one from the list of suggestion.

Additional information
```
$ snap list lxd
Name  Version      Rev    Tracking     Publisher   Notes
lxd   git-de3f147  31657  latest/edge  canonical✓  -
```

markylaing:
TIL that `set-default` is an alias for `switch` and not the other way around :laughing:

kadinsayani:
I'm seeing all of my remotes (including non-default remotes) appear when tab completing `lxc remote switch` on my end. For this completion, we load remotes from `g.conf.Remotes`, which should include all remotes. Do you have a reproducer I can try to continue investigating further?

simondeziel:
@kadinsayani if you've built snaps before, you should have `craft-com.ubuntu.cloud-buildd` and `craft-com.ubuntu.cloud-buildd-daily` remotes. Can you tab complete those when on `latest/edge` cause I cannot.

kadinsayani:
It looks like this issue is only prevalent with the LXD snap - upon regenerating the completion script and sourcing it locally, all remotes appear as expected.

A fresh install of the LXD snap yields the following results:

```bash
root@c1:~# lxd.lxc __complete remote switch cd
ubuntu
ubuntu-daily
ubuntu-minimal
ubuntu-minimal-daily
images
cdimage
local
:4
Completion ended with directive: ShellCompDirectiveNoFileComp
root@c1:~# lxd.lxc remote switch [TAB]
images                ubuntu                ubuntu-minimal
local                 ubuntu-daily          ubuntu-minimal-daily
```

The completion directive returns all of the available remotes, but it appears that `snap` is intercepting the response and excluding non-default remotes (`cdimage` in this case).

PR #14677, opened 2024-12-17
CLI: Exclude default remote from `lxc remote switch|remove` shell completions

kadinsayani:
Resolves 1/2 issues raised in https://github.com/canonical/lxd/issues/14635.

This PR adds logic to exclude the default (current) remote from shell completions when running `lxc remote switch|remove`. As described in [this comment](https://github.com/canonical/lxd/issues/14635#issuecomment-2547022037), it appears that `snapd` is intercepting the response and excluding non-default remotes (`cdimage` in this case). I'll see if I can land a fix for the second issue in https://github.com/canonical/lxd-pkg-snap.

### DIFF
--- a/lxc/completion.go
+++ b/lxc/completion.go
@@ -1347,10 +1347,13 @@ func (g *cmdGlobal) cmpRemotes(includeAll bool) ([]string, cobra.ShellCompDirect
 
 // cmpRemoteNames provides shell completion for remote names.
 // It returns a list of remote names provided by `g.conf.Remotes` along with a shell completion directive.
-func (g *cmdGlobal) cmpRemoteNames() ([]string, cobra.ShellCompDirective) {
-	var results []string
-
+func (g *cmdGlobal) cmpRemoteNames(includeDefaultRemote bool) ([]string, cobra.ShellCompDirective) {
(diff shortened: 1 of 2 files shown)


lxc/completion.go
```diff
@@ -1347,10 +1347,13 @@ func (g *cmdGlobal) cmpRemotes(includeAll bool) ([]string, cobra.ShellCompDirect
 
 // cmpRemoteNames provides shell completion for remote names.
 // It returns a list of remote names provided by `g.conf.Remotes` along with a shell completion directive.
-func (g *cmdGlobal) cmpRemoteNames() ([]string, cobra.ShellCompDirective) {
-	var results []string
-
+func (g *cmdGlobal) cmpRemoteNames(includeDefaultRemote bool) ([]string, cobra.ShellCompDirective) {
+	results := make([]string, len(g.conf.Remotes))
```
simondeziel, line 1351:
```suggestion
	results := make([]string, 0, len(g.conf.Remotes))
```

kadinsayani: Thanks, nice catch.

simondeziel (approved):

tomponline (approved):

PR #14680, opened 2024-12-17
VM: Move check for container-specific prefixed keys applied to VMs up

kadinsayani:
This PR fixes the following regression:
https://github.com/canonical/lxd-ci/actions/runs/12367790805/job/34516721509#step:11:93.

We always need to check config keys applied to VM's to ensure container-specific prefixed keys are disallowed.

### DIFF
--- a/lxd/instance/instance_utils.go
+++ b/lxd/instance/instance_utils.go
@@ -128,14 +128,14 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
 }
 
 func validConfigKey(os *sys.OS, key string, value string, instanceType instancetype.Type) error {
+	// Disallow keys with container-specific prefixes such as "linux.sysctl." and "limits.kernel." for VMs.
+	if instanceType == instancetype.VM && shared.StringHasPrefix(key, instancetype.ConfigKeyPrefixesContainer...) {
+		return fmt.Errorf("%q isn't supported for %q", key, instanceType)
+	}
+
 	// Check if the key is a valid prefix and whether or not it requires a subkey.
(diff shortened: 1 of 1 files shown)


lxd/instance/instance_utils.go
```diff
@@ -128,14 +128,14 @@ func ValidConfig(sysOS *sys.OS, config map[string]string, expanded bool, instanc
...
 
 func validConfigKey(os *sys.OS, key string, value string, instanceType instancetype.Type) error {
+	// Disallow keys with container-specific prefixes such as "linux.sysctl." and "limits.kernel." for VMs.
+	if instanceType == instancetype.VM && shared.StringHasPrefix(key, instancetype.ConfigKeyPrefixesContainer...) {
+		return fmt.Errorf("%q isn't supported for %q", key, instanceType)
+	}
+
 	// Check if the key is a valid prefix and whether or not it requires a subkey.
 	knownPrefixes := append(instancetype.ConfigKeyPrefixesAny, instancetype.ConfigKeyPrefixesContainer...)
 	if strings.HasSuffix(key, ".") {
```
simondeziel, line 138: Shouldn't we also reject anything that ends with a trailing `.` as that's never going to be a valid config key anyway so why bother doing more checks?

simondeziel (approved):

lxd/instance/instance_utils.go
kadinsayani: We can do that, however, the additional check allows us to throw more specific error messages:
`"Unknown configuration key: %q"` if the key is not known and `"%q requires a subkey"` if the key is known but missing a subkey. Also, the check is pretty inexpensive computationally since the `knownPrefixes` slice is only 5 elements.

tomponline (approved):

PR #14681, opened 2024-12-17
Preserve `custom/c1` syntax during storage volume rename

MggMuggins:
Follow-up to #14491; if a volume is specified as `source: custom/c1`, then a rename of `c1`->`c2` should leave the device with `source: custom/c2`.

### DIFF
--- a/lxd/storage_volumes_utils.go
+++ b/lxd/storage_volumes_utils.go
@@ -30,7 +30,12 @@ func storagePoolVolumeUpdateUsers(s *state.State, projectName string, oldPoolNam
 			_, exists := localDevices[devName]
 			if exists {
 				localDevices[devName]["pool"] = newPoolName
-				localDevices[devName]["source"] = newVol.Name
+
+				if strings.Contains(localDevices[devName]["source"], "/") {
+					localDevices[devName]["source"] = newVol.Type + "/" + newVol.Name
+				} else {
+					localDevices[devName]["source"] = newVol.Name
+				}
 			}
 		}
 
@@ -62,7 +67,12 @@ func storagePoolVolumeUpdateUsers(s *state.State, projectName string, oldPoolNam
 		for name, dev := range profile.Devices {
 			if shared.ValueInSlice(name, usedByDevices) {
 				dev["pool"] = newPoolName
-				dev["source"] = newVol.Name
+
+				if strings.Contains(dev["source"], "/") {
+					dev["source"] = newVol.Type + "/" + newVol.Name
+				} else {
+					dev["source"] = newVol.Name
+				}
 			}
 		}
 
--- a/test/suites/storage_volume_attach.sh
+++ b/test/suites/storage_volume_attach.sh
@@ -116,8 +116,15 @@ EOF
 
   lxc storage volume show "${pool}" testvolume | grep -q '/1.0/instances/c1'
 
+  # rename should preserve source=custom/testvolume syntax
+  lxc stop c1
+
+  lxc storage volume rename "${pool}" testvolume testvolume2
+
+  [ "$(lxc config device get c1 testvolume source)" = "custom/testvolume2" ]
+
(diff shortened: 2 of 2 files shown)


tomponline (approved):

MggMuggins:
You an me both :laughing:

MggMuggins:
@tomponline There's no bug when used with instance devices and renaming devices attached to profiles fails, but with an unhelpful error message.

## Instances
Assume the following configuration in a lxd cluster with cluster members `c1`, `c2`, and `c3`:
```
$ lxc ls
+------+---------+------+------+-----------------+-----------+----------+
| NAME |  STATE  | IPV4 | IPV6 |      TYPE       | SNAPSHOTS | LOCATION |
+------+---------+------+------+-----------------+-----------+----------+
| v1   | STOPPED |      |      | VIRTUAL-MACHINE | 0         | c1       |
+------+---------+------+------+-----------------+-----------+----------+
| v2   | STOPPED |      |      | VIRTUAL-MACHINE | 0         | c2       |
+------+---------+------+------+-----------------+-----------+----------+
$ lxc storage volume ls
+-------+-----------------+------------------------------------------------------------------+-------------+--------------+---------+----------+
| POOL  |      TYPE       |                               NAME                               | DESCRIPTION | CONTENT-TYPE | USED BY | LOCATION |
+-------+-----------------+------------------------------------------------------------------+-------------+--------------+---------+----------+
| local | custom          | blk1                                                             |             | block        | 1       | c1       |
+-------+-----------------+------------------------------------------------------------------+-------------+--------------+---------+----------+
| local | custom          | blk1                                                             |             | block        | 1       | c2       |
+-------+-----------------+------------------------------------------------------------------+-------------+--------------+---------+----------+
...
+-------+-----------------+------------------------------------------------------------------+-------------+--------------+---------+----------+
| local | virtual-machine | v1                                                               |             | block        | 1       | c1       |
+-------+-----------------+------------------------------------------------------------------+-------------+--------------+---------+----------+
| local | virtual-machine | v2                                                               |             | block        | 1       | c2       |
+-------+-----------------+------------------------------------------------------------------+-------------+--------------+---------+----------+
$ lxc config show v1
...
devices:
  blk1:
    pool: local
    source: blk1
    type: disk
ephemeral: false
profiles:
- default
- ...
$ lxc config show v2
...
devices:
  blk1:
    pool: local
    source: blk1
    type: disk
ephemeral: false
profiles:
- default
...
```
Renaming `blk1` on `c1` only modifies the disk device on `v1`:
```
$ lxc storage volume rename local blk1 blk2 --target c1
Renamed storage volume from "blk1" to "blk2"
$ lxc config show v1
...
devices:
  blk1:
    pool: local
    source: blk2
    type: disk
ephemeral: false
profiles:
- default
...
$ lxc config show v2
...
devices:
  blk1:
    pool: local
    source: blk1
    type: disk
ephemeral: false
profiles:
- default
...
```

## Profiles
The block volume sharing rules aren't strictly correct; adding the same device to the default profile hits this:
```
$ lxc profile device add default blk1 disk pool=local source=blk1
Config parsing error: Device validation failed for "blk1": Cannot add custom storage block volume to profiles if security.shared is false or unset
```

Setting that aside for a moment, the rename fails here:
```
$ lxc storage volume set local blk1 security.shared=true --target c1
$ lxc storage volume set local blk1 security.shared=true --target c2
$ lxc storage volume rename local blk1 blk2 --target c1
Error: Device validation failed for "blk1": Failed loading custom volume: Storage volume not found
```

I tried a similar setup with filesystem volumes in profiles and get the same thing. I think the current implementation is sound, at least insofar as it [validates](https://github.com/canonical/lxd/blob/main/lxd/storage_volumes_utils.go#L83) changes after making them.

Looking at the code I think there's a corner case involving how revert happens (or not) when the rename fails like this, but I need a little more time to work it out properly.

PR #14684, opened 2024-12-18
CLI: Rework `cmpInstanceKeys` for contextual completions based on instance type

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/14689.

Summary of changes:
- Correctly provides instance config key completions based on instance type.

### DIFF
--- a/lxc/completion.go
+++ b/lxc/completion.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/canonical/lxd/lxd/instance/instancetype"
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
 )
@@ -284,6 +285,21 @@ func (g *cmdGlobal) cmpInstanceKeys(instanceName string) ([]string, cobra.ShellC
 		return nil, cobra.ShellCompDirectiveError
 	}
 
+	instanceTypeAnyKeys := make([]string, 0, len(instancetype.InstanceConfigKeysAny))
+	for key := range instancetype.InstanceConfigKeysAny {
+		instanceTypeAnyKeys = append(instanceTypeAnyKeys, key)
+	}
+
+	instanceTypeContainerKeys := make([]string, 0, len(instancetype.InstanceConfigKeysContainer))
+	for key := range instancetype.InstanceConfigKeysContainer {
+		instanceTypeContainerKeys = append(instanceTypeContainerKeys, key)
+	}
+
+	instanceTypeVMKeys := make([]string, 0, len(instancetype.InstanceConfigKeysVM))
+	for key := range instancetype.InstanceConfigKeysVM {
+		instanceTypeVMKeys = append(instanceTypeVMKeys, key)
+	}
+
 	// Pre-allocate configKeys slice capacity.
 	keyCount := 0
 	for _, field := range instanceConfig {
@@ -294,18 +310,16 @@ func (g *cmdGlobal) cmpInstanceKeys(instanceName string) ([]string, cobra.ShellC
 
 	for _, field := range instanceConfig {
 		for _, key := range field.Keys {
-			for configKey, configKeyField := range key {
+			for configKey := range key {
 				configKey = strings.TrimSuffix(configKey, "*")
 
-				// InstanceTypeAny config keys.
-				if configKeyField.Condition == "" {
+				if shared.ValueInSlice(configKey, instanceTypeAnyKeys) || shared.StringHasPrefix(configKey, instancetype.ConfigKeyPrefixesAny...) {
 					configKeys = append(configKeys, configKey)
-					continue
 				}
 
-				if instanceType == string(api.InstanceTypeContainer) && configKeyField.Condition == "container" {
+				if instanceType == string(api.InstanceTypeContainer) && (shared.ValueInSlice(configKey, instanceTypeContainerKeys) || shared.StringHasPrefix(configKey, instancetype.ConfigKeyPrefixesContainer...)) {
 					configKeys = append(configKeys, configKey)
-				} else if instanceType == string(api.InstanceTypeVM) && configKeyField.Condition == "virtual machine" {
+				} else if instanceType == string(api.InstanceTypeVM) && shared.ValueInSlice(configKey, instanceTypeVMKeys) {
 					configKeys = append(configKeys, configKey)
 				}
 			}
--- a/lxc/completion.go
+++ b/lxc/completion.go
@@ -1550,9 +1550,9 @@ func (g *cmdGlobal) cmpStoragePoolVolumeConfigs(poolName string, volumeName stri
 	resource := resources[0]
 	client := resource.server
 
-	var pool = poolName
-	if strings.Contains(poolName, ":") {
-		pool = strings.Split(poolName, ":")[1]
+	_, pool, found := strings.Cut(poolName, ":")
+	if !found {
+		pool = poolName
 	}
 
 	volName, volType := parseVolume("custom", volumeName)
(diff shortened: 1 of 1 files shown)


lxc/completion.go
```diff
@@ -1634,9 +1634,17 @@ func (g *cmdGlobal) cmpStoragePoolVolumes(poolName string, volumeTypes ...string
...
+	// Only complete volumes specified by volumeTypes.
 	for _, volume := range volumes {
 		_, volType := parseVolume("custom", volume)
+
+		// Parse snapshots returned by GetStoragePoolVolumeNames.
+		if strings.Contains(volume, "/snapshots/") {
+			fields := strings.SplitN(volume, "/snapshots/", 2)
+			customVolumeNames = append(customVolumeNames, fields[0]+"/"+fields[1])
+			continue
+		}
```
simondeziel, line 1647: Feels like a good candidate for `before, after, found := strings.Cut` as it combines the `strings.Contains` and `strings.SplitN` into one operation.

lxc/completion.go
```diff
@@ -274,7 +274,7 @@ func (g *cmdGlobal) cmpInstanceKeys(instanceName string) ([]string, cobra.ShellC
 				configKey = strings.TrimSuffix(configKey, "*")
 
 				// InstanceTypeAny config keys.
-				if configKeyField.Condition == "" {
+				if configKeyField.Condition == "" || configKeyField.Condition == "If supported by image" {
```
tomponline, line 277: this looks fishy, whats it comparing to?

kadinsayani: The condition field for a `cloud-init.*` instance config key:

```
			"cloud-init": {
				"keys": [
					{
						"cloud-init.network-config": {
							"condition": "If supported by image",
							"defaultdesc": "`DHCP on eth0`",
							"liveupdate": "no",
							"longdesc": "The content is used as seed value for `cloud-init`.",
							"shortdesc": "Network configuration for `cloud-init`",
							"type": "string"
						}
					},

```

Alternatively, we could just add all keys under the `cloud-init` map.

tomponline: why are we needing to consider the Condition field?

kadinsayani: We check the condition field to determine if a key is valid for instance type any, container, or virtual-machine.

tomponline: ok, so we really dont want to be comparing to arbitrary free-text values.

This seem super hacky to me.

tomponline: > Alternatively, we could add all keys under the cloud-init map.
What do you mean please?

kadinsayani: Under metadata, we have configs->instance->config key group->keys. Currently, we add all keys from every config key group as long as they match the condition of instance type. The `cloud-init.*` keys are skipped because their condition is "If supported by image" and not "container" or "virtual machine".

If we want to move away from comparing to strings found in metadata, we can grab the keys from `instancetype.InstanceConfigKeysAny/VM/Container`.

kadinsayani: That would mean reverting https://github.com/canonical/lxd/pull/14542/commits/4ae5a7b2ded2680337d3a94b5276f088bf0e044c.

kadinsayani: Or, we can check if a key is found in `instancetype.InstanceConfigKeysAny/VM/Container` to deduce the type, but this adds complexity.

tomponline: > Currently, we add all keys from every config key group as long as they match the condition of instance type.

Do we make clear the link anywhere between the value of `condition` and how it affects auto complete? 
It feels like this link has been only recently established in these changes and could easily be broken due to not making clear the link, and hard to detect (due to there being no tests for auto complete).

I wonder if lxd-metadata should be updated to test for valid "conditions" values from a slice it has so we can only use accepted ones.

Then we could expose those as constants and have them used as checks for the auto complete.

> If we want to move away from comparing to strings found in metadata, we can grab the keys from instancetype.InstanceConfigKeysAny/VM/Container.

But that would then be restricted to keys known about by the client, and not by the server leading to situations where the client may suggest something the server doesnt support or miss out on suggesting something it does support.


Also, by checking for `condition` as it stands today, you're missing out on other instance fields like `raw.idmap` and `security.idmap.*` which have a condition of `unprivileged container`.

One acceptable workaround would be to have a list of prefixes that you can accept from the metadata API irrespective of condition.

kadinsayani: I'm in agreement with moving away from checking "condition" values.

> But that would then be restricted to keys known about by the client, and not by the server leading to situations where the client may suggest something the server doesnt support or miss out on suggesting something it does support.

We can also consider checking the keys from the metadata API against `instancetype.InstanceConfigKeysAny/VM/Container` to deduce type. Aren't the keys generated by the metadata API sourced from `lxd/instance/instancetype/instance.go`?

tomponline: Yes, but remember lxc would be built against a list at a point in time, but the server maybe newer and have additional keys only in metadata api

kadinsayani: What about adding a "supported instance type" field to the metadata? I'm trying to avoid hard coding a list of accepted keys based on instance type in the client, since these may not line up with the server. Also, there are cases where a prefix contains keys supported by containers and VMs. For example, the `migration` prefix contains `migration.incremental.memory` which is only supported by containers, and `migration.stateful` which is only supported by VMs.

tomponline:
Is it possible to split this pr into two, so we can merge the non contentious parts?

lxc/completion.go
kadinsayani: Another consideration for adding an instance type field to config keys in metadata is the potential usability improvements that could be achieved in `lxd-ui`. I don't use it much so I'm not sure if completions are served when setting profile or instance config keys, but having an instance type field in the metadata could facilitate smart completions for config keys in the ui.

kadinsayani: Another option could be including the `condition` metadata string with the completion results in brackets. For example:

```
$ lxc config set c1 [TAB]
limits.cpu.pin_strategy (virtual machine)
cloud-init.network-config (If supported by image)
...
```

With this option, we effectively mirror the documentation and inform users of all available configuration keys and their corresponding condition (virtual machine only, unprivileged container, if supported by image, storage volumes managed by LXD, etc.)

kadinsayani: Resolved by using a best-effort approach comparing config keys from the metadata API to keys known at compile time of the `instancetype` module.

lxc/completion.go
```diff
@@ -1536,7 +1550,7 @@ func (g *cmdGlobal) cmpStoragePoolVolumeConfigs(poolName string, volumeName stri
 	resource := resources[0]
 	client := resource.server
 
-	var pool = poolName
+	pool := poolName
 	if strings.Contains(poolName, ":") {
 		pool = strings.Split(poolName, ":")[1]
```
simondeziel, line 1555: Could we use `strings.Cut` here instead of `Contains` + `Split`, please?

lxc/completion.go
```diff
@@ -1605,7 +1619,7 @@ func (g *cmdGlobal) cmpStoragePoolVolumeProfiles(poolName string, volumeName str
...
```
simondeziel, line 1624: Same here, `Cut`, please.

lxc/completion.go
```diff
@@ -1568,7 +1582,7 @@ func (g *cmdGlobal) cmpStoragePoolVolumeInstances(poolName string, volumeName st
...
```
simondeziel, line 1587: Same here, `Cut`, please.

lxc/completion.go
```diff
@@ -1642,7 +1656,7 @@ func (g *cmdGlobal) cmpStoragePoolVolumeSnapshots(poolName string, volumeName st
...
```
simondeziel, line 1661: Same here, `Cut`, please.

lxc/completion.go
kadinsayani: Sure thing!

simondeziel (approved):

tomponline (approved):

PR #14693, opened 2024-12-18
doc/cloud-init: add a note that recent `cloud-init` on older Ubuntu still doesn't support modern keys

simondeziel:


simondeziel:
@holmanb I couldn't find where I read it before but I think that `cloud-init` backports to old releases (like `20.04`) intentionally disable support for the LXD datasource and causes LXD guests to use the `nocloud` one. Is that right?

```
$ lxc init ubuntu-minimal-daily:20.04 f1

$ lxc config set f1 cloud-init.user-data - << EOF
#cloud-config
runcmd:
  - touch /root/cloud-init.user-data
EOF

$ lxc config set f1 user.user-data - << EOF
#cloud-config
runcmd:
  - touch /root/user.user-data
EOF

$ lxc start f1 && lxc exec f1 -- cloud-init status --wait
$ lxc exec f1 -- cloud-init status --long | grep detail
detail: DataSourceNoCloud [seed=/var/lib/cloud/seed/nocloud-net]
$ lxc exec f1 -- cloud-init query userdata
#cloud-config
runcmd:
  - touch /root/user.user-data

$ lxc exec f1 -- ls -l /root/
total 1
-rw-r--r-- 1 root root 0 Dec 18 23:27 user.user-data
```

In the above, we see that only the legacy key `user.user-data` was taken into account.

simondeziel:
For comparison, here is with a more modern Ubuntu version (`22.04`):

```
$ lxc init ubuntu-minimal-daily:22.04 j1

$ lxc config set j1 cloud-init.user-data - << EOF
#cloud-config
runcmd:
  - touch /root/cloud-init.user-data
EOF

$ lxc config set j1 user.user-data - << EOF
#cloud-config
runcmd:
  - touch /root/user.user-data
EOF

$ lxc start j1 && lxc exec j1 -- cloud-init status --wait
$ lxc exec j1 -- cloud-init status --long | grep detail
detail: DataSourceLXD


$ lxc exec j1 -- cloud-init query userdata
#cloud-config
runcmd:
  - touch /root/cloud-init.user-data


$ lxc exec j1 -- ls -l /root/
total 1
-rw-r--r-- 1 root root 0 Dec 18 23:31 cloud-init.user-data
```

tomponline (approved):

holmanb:
> @holmanb I couldn't find where I read it before but I think that `cloud-init` backports to old releases (like `20.04`) intentionally disable support for the LXD datasource and causes LXD guests to use the `nocloud` one. Is that right?

Yes. Thanks for this @simondeziel!

Issue #14159, opened 2024-09-25
GetResources is returning unexpected data when it is executed on VMs with Passthrough CPU


r00ta:
# Issue description

In MAAS we are using some utilities from the LXD library to extract the hardware info of a machine (see [here](https://github.com/canonical/maas/blob/4b019c3dcae62decbb60cffff8573d1ccdbed49c/src/host-info/pkg/info/info.go#L180)). If this is run on a VM with Passthrough CPU it returning that the machine has N sockets where N is the number of CPU threads. 
The problem is that the socket name changes from time to time (causing MAAS not to display the cpu model name). See [this](https://pastebin.ubuntu.com/p/8ST8skMDcP/) for the full output and [this](https://pastebin.ubuntu.com/p/PgHFgGVRTX/) for the output of /proc/cpuinfo. In this specific case it is flipping from Intel(R) Xeon(R) CPU Max 9462 to pc-q35-6.2 (with vendor QEMU). 

Is this a known limitation or can this be considered as a bug?

tomponline:
Associated MM thread https://chat.canonical.com/canonical/pl/wzmwbz7ixbbk9rh4847wni7kua

mihalicyn:
Hi @r00ta,

please can you give me a bit of context in addition to what Simon asked.

Are you using CPU hotplug for that VM? Is this thing is stable-reproducible or not?

mihalicyn:
`GetResources` generate it's output based on `/proc/cpuinfo` and `/sys/devices/system/cpu`.

>See [this](https://pastebin.ubuntu.com/p/8ST8skMDcP/) for the full output and [this](https://pastebin.ubuntu.com/p/PgHFgGVRTX/) for the output of /proc/cpuinfo

These two outputs doesn't appear to be collected from the same machine. Even CPUs are different:
https://www.intel.com/content/www/us/en/products/sku/232597/intel-xeon-cpu-max-9462-processor-75m-cache-2-70-ghz/specifications.html
https://www.intel.com/content/www/us/en/products/sku/192478/intel-xeon-platinum-8280-processor-38-5m-cache-2-70-ghz/specifications.html

I can imagine inconsistency like mixture of `pc-q35-6.2` cpu with `Intel(R) Xeon(R) CPU Max 9462` but having absolutely another one `Intel(R) Xeon(R) Platinum 8280` on the same machine looks suspicious.

[...]

PR #14683, opened 2024-12-17
Minor tweaks

simondeziel:
Address a few small things noticed while doing code reviews.

PR #14695, opened 2024-12-19
lxd/resources/cpu: fix GetCPU() to handle multisockets systems

mihalicyn:
When VM has more than 9 vCPUs and they are exposed as sitting on a different "physical" sockets we don't handle it properly [1] as we implicitly assume that entries in /sys/devices/system/cpu/ are sorted the same way as in /proc/cpuinfo. But it's not correct assumption.

To reproduce that, the following QEMU command line parameter can be used: -smp 12,sockets=12,cores=1,threads=1

Affected systems have more than one line in the following command output: cat /sys/devices/system/cpu/cpu*/topology/physical_package_id | uniq and more than 9 CPUs.

[1] https://github.com/canonical/lxd/issues/14159#issuecomment-2551838382
Fixes: #14159
Signed-off-by: Alexander Mikhalitsyn <aleksandr.mikhalitsyn@canonical.com>

### DIFF
--- a/lxd/resources/cpu.go
+++ b/lxd/resources/cpu.go
@@ -174,7 +174,7 @@ func getCPUCache(path string) ([]api.ResourcesCPUCache, error) {
 	return caches, nil
 }
 
-func getCPUdmi() (string, string, error) {
+func getCPUdmi() (vendor string, model string, err error) {
 	// Open the system DMI tables.
 	stream, _, err := smbios.Stream()
 	if err != nil {
--- a/lxd/resources/cpu.go
+++ b/lxd/resources/cpu.go
@@ -206,6 +206,11 @@ func getCPUdmi() (string, string, error) {
 	return "", "", fmt.Errorf("No DMI table found")
 }
 
+type cpuInfo struct {
+	Name   string
+	Vendor string
+}
+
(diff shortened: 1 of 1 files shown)


lxd/resources/cpu.go
```diff
@@ -227,7 +232,68 @@ func GetCPU() (*api.ResourcesCPU, error) {
...
+
+	// CPU information
+	for cpuInfoScanner.Scan() {
+		line := strings.TrimSpace(cpuInfoScanner.Text())
+		if !strings.HasPrefix(line, "processor") {
+			return nil, fmt.Errorf("Failed to parse /proc/cpuinfo: Unexpected line %q", line)
+		}
+
+		// Extract cpu index
+		fields := strings.SplitN(line, ":", 2)
```
tomponline, line 246: strings.Cut would likely be more appropriate here as an FYI only

tomponline (approved):
Thanks!

lxd/resources/cpu.go
simondeziel: https://github.com/canonical/lxd/pull/14683/commits/de6416c7f889b735592a0d47b5b7655979a0b757 is my attempt at using `strings.Cut`.

Issue #14679, opened 2024-12-17
`tests/vm-migration` regression with `latest/edge` snap


simondeziel:
The last step in [`tests/vm-migration`](https://github.com/canonical/lxd-ci/blob/main/tests/vm-migration#L135) it to do a live `lxc move` of `v1` from `member1` to `member2`. However, that doesn't work anymore in CI with `latest/edge`. This is reproducible locally and when doing this, I found that `member2` should have had a receiving QEMU process but it had none:

```
root@member2:~# ps faux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           2  0.0  0.0      0     0 ?        S    17:59   0:00 [kthreadd]
...
root        2275  0.0  0.0      0     0 ?        S    18:08   0:00  \_ [psimon]
root           1  0.7  0.6  22304 13424 ?        Ss   17:59   0:05 /sbin/init
root         127  0.0  0.5  34356 10504 ?        S<s  17:59   0:00 /usr/lib/systemd/systemd-journald
root         190  0.0  0.3  25436  7020 ?        Ss   17:59   0:00 /usr/lib/systemd/systemd-udevd
root         258  0.0  0.8 1238664 17268 ?       Ssl  17:59   0:00 /run/lxd_agent/lxd-agent
root        2265  0.0  0.1   6428  3840 pts/0    Ss   18:08   0:00  \_ su -l
root        2278  0.0  0.2   5144  4224 pts/0    S    18:08   0:00      \_ -bash
root        2312  0.0  0.2   8280  4096 pts/0    R+   18:11   0:00          \_ ps faux
systemd+     333  0.0  0.6  21584 12928 ?        Ss   17:59   0:00 /usr/lib/systemd/systemd-resolved
systemd+     338  0.0  0.3  91020  7808 ?        Ssl  17:59   0:00 /usr/lib/systemd/systemd-timesyncd
systemd+     553  0.0  0.4  18988  9472 ?        Ss   17:59   0:00 /usr/lib/systemd/systemd-networkd
message+     639  0.0  0.2   8276  4480 ?        Ss   17:59   0:00 @dbus-daemon --system --address=systemd: --nofork --nopidfile --systemd-activation --syslog-only
root         663  0.0  0.4  17800  8576 ?        Ss   17:59   0:00 /usr/lib/systemd/systemd-logind
[...]

PR #14696, opened 2024-12-19
VM: Fix live migration

tomponline:
PR https://github.com/canonical/lxd/pull/14539 broke live migration in commit https://github.com/canonical/lxd/pull/14539/commits/d734721b3d3e2000dd71bfc1c2cb7c92ff8c298d

This is because when live migrating a VM, the VM's `Start()` function is called on the target side (so that live migration can occur) before the Location of the instance is updated in the DB. This in turn calls `statusCode` during the start process.

The change to the `statusCode` function meant that this returned an error status because the instance was in theory non-local, and that prevented the qemu process from being started.

This PR reverts the `statusCode` error change and moves that logic into the Render function.

Fixes https://github.com/canonical/lxd/issues/14679


hamistao (approved):
Thanks for this and apologies for introducing this regression. TIL more about live migration on LXD.

Issue #14682, opened 2024-12-17
`lxc storage volume <action>` completion issue with snapshots


simondeziel:
Steps to reproduce:

```
$ lxc storage volume create default foo
$ lxc storage volume snapshot default foo
$ lxc storage volume show default custom/foo<tab>
custom/foo                  custom/foo/snapshots/snap0

$ lxc storage volume show default custom/foo/snapshots/snap0 
Error: Invalid snapshot name

$ lxc storage volume show default custom/foo/snap0 
name: snap0
description: ""
content_type: filesystem
created_at: 2024-12-17T22:30:17.958279135Z
expires_at: 0001-01-01T00:00:00Z
config:
  volatile.uuid: 6ea5154a-6c91-4776-ad52-dcc84c4e9211
```

`tl;dr`: there is a bogus `snapshots/` part being added during completion.

PR #14702, opened 2024-12-19
CLI: `lxc storage volume` completion fixes and command description update

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/14682.

Summary of changes:

- Fixes shell completions for `lxc storage volume show`. `GetStoragePoolVolumeNames` returns the full volume name including "/snapshots/", which is incorrect in the context of the CLI. To address this issue, parsing has been added to serve valid completions for snapshot volumes.
- Updates the command description for `lxc storage volume snapshot`.
- Fixes `cmpInstanceSetKeys` logic to provide completions for any instance keys which are currently set, not just prefixed keys.


simondeziel (dismissed):

lxc/completion.go
```diff
@@ -377,12 +375,9 @@ func (g *cmdGlobal) cmpInstanceSetKeys(instanceName string) ([]string, cobra.She
 	for configKey := range instance.Config {
 		// We only want to return the intersection between allInstanceConfigKeys and configKeys to avoid returning the full instance config.
 		_, exists := keySet[configKey]
-		if exists {
-			if shared.StringHasPrefix(configKey, instancetype.ConfigKeyPrefixesAny...) {
-				configKeys = append(configKeys, configKey)
-			} else if instanceType == string(api.InstanceTypeContainer) && shared.StringHasPrefix(configKey, instancetype.ConfigKeyPrefixesContainer...) {
-				configKeys = append(configKeys, configKey)
-			}
+		knownPrefixes := append(instancetype.ConfigKeyPrefixesAny, instancetype.ConfigKeyPrefixesContainer...)
```
tomponline, line 378: why are we building this slice for every iteration of the loop? It seems static?

tomponline, line 378: Separately, so following on from yesterday's comment, it seems like you're already depending on build-time config keys from `instancetype` package, which could differ from the keys supported by the target server (as lxc client version may differ from server version). Here it looks to me like this could end up with a situation where valid keys supported by the server are not suggested by the client, or vice versa.


However, if you've already gone this route here, then why did you go the route of using the metadata `Condition` field in the other PR? It seems to me we should pick one approach for gathering supported keys, either a build-time list (and accept the drawbacks that it may mismatch what is supported) or use the metadata API.

Mixing the approach seems confusing to me.

lxc/completion.go
```diff
@@ -377,12 +375,9 @@ func (g *cmdGlobal) cmpInstanceSetKeys(instanceName string) ([]string, cobra.She
...
+		if exists || strings.HasPrefix(configKey, instancetype.ConfigVolatilePrefix) || shared.StringHasPrefix(configKey, knownPrefixes...) {
```
tomponline, line 379: Also, why are we including the volatile key prefix, should we be encouraging users to change these keys at all?

kadinsayani:
> I'm still concerned with the approaches we're using to gather keys.
> 
> 
> 
> We need to have a single source of truth for config keys and not mix `instancetype` and metadata API sources.

With the addition of improved configuration key validation with https://github.com/canonical/lxd/pull/14584, we actually don't need any validation in `cmpInstanceSetKeys`, since we can assume any set key is valid for an instance type. I'll update this PR to only filter out `volatile.*` keys. Should we also filter out `image.*` keys?

lxc/completion.go
kadinsayani: I don't think `volatile.*` keys should be completed.

kadinsayani:
> I'm still concerned with the approaches we're using to gather keys.
> 
> We need to have a single source of truth for config keys and not mix `instancetype` and metadata API sources.

I've removed the `instancetype` package import. I think we should stick to the metadata API to ensure compatibility.

tomponline (approved):
Thanks!

PR #14703, opened 2024-12-20
Storage: Add reverter to disk device update & test

MggMuggins:
This makes `storagePoolVolumeUpdateUsers` safe to call without using a reverter.

This also adds a few tests around renaming attached local custom storage volumes in a clustered context. As we've discussed previously (#14681), updating disk devices in profiles isn't necessarily sound. These tests give me a little more confidence that updates behave sanely for local storage volumes in a cluster.

We should consider eliminating this feature altogether and simply making it a hard error to rename a custom storage volume while it is referred to by any disk device.

### DIFF
--- a/lxd/storage_volumes.go
+++ b/lxd/storage_volumes.go
@@ -1878,10 +1878,12 @@ func storagePoolVolumeTypePostRename(s *state.State, r *http.Request, poolName s
 	defer revert.Fail()
 
 	// Update devices using the volume in instances and profiles.
-	err = storagePoolVolumeUpdateUsers(s, projectName, pool.Name(), vol, pool.Name(), &newVol)
+	cleanup, err := storagePoolVolumeUpdateUsers(s, projectName, pool.Name(), vol, pool.Name(), &newVol)
 	if err != nil {
 		return response.SmartError(err)
 	}
+
+	revert.Add(cleanup)
 
 	// Use an empty operation for this sync response to pass the requestor
 	op := &operations.Operation{}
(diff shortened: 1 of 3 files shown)


lxd/storage_volumes_utils.go
```diff
@@ -58,6 +63,17 @@ func storagePoolVolumeUpdateUsers(s *state.State, projectName string, oldPoolNam
 			return err
 		}
 
+		revert.Add(func() {
+			err := inst.Update(dbInst, false)
+			if err != nil {
+				logger.Error("Failed to revert instance update", logger.Ctx{
```
tomponline, line 69: nit: we tend to keep these on one line please

lxd/storage_volumes.go
```diff
@@ -1924,10 +1924,6 @@ func storagePoolVolumeTypePostMove(s *state.State, r *http.Request, poolName str
 			return err
 		}
 
-		revert.Add(func() {
```
tomponline, line 1927: the downside with this approach is that now if any subsequent function fails, like `newPool.CreateCustomVolumeFromCopy` or `pool.DeleteCustomVolume` the updated volume users wont get reverted.

Suggest instead that you keep your updated implementation, but update `storagePoolVolumeUpdateUsers` to return a `revert.Hook` (like `clusterInitMember`).

That way the contract can be that `storagePoolVolumeUpdateUsers` will deal with its own reverts if it returns a non-nil error, but if it returns a nil error then it will also return a revert.Hook that can be added to the outer reverter to be undone if a subsequent step fails.

tomponline: and `storagePoolVolumeTypePostRename` can then benefit from that approach to and use an outer reverter too.

tomponline (approved):

PR #14704, opened 2024-12-20
DB: Fix update-schema to support lxd in non-GOPATH locations

tomponline:


lxd/db/schema/update.go
```diff
@@ -40,9 +39,10 @@ func DotGo(updates map[int]Update, name string) error {
 	// Passing 1 to runtime.Caller identifies our caller.
 	_, filename, _, _ := runtime.Caller(1)
 
-	// runtime.Caller returns the path after "${GOPATH}/src" when used with `go generate`.
+	// runtime.Caller returns source file path starting from github.com.
+	// Translate into relative path to source file.
 	if strings.HasPrefix(filename, "github.com") {
-		filename = filepath.Join(os.Getenv("GOPATH"), "src", filename)
+		filename = strings.TrimPrefix(filename, "github.com/canonical/lxd/lxd/db/")
```
markylaing, line 45: Honestly I'm finding it strange that this is required at all.

As far as I can tell, schema generation is only used by LXD and not any consumers or `lxd-generate` (microcloud, microceph, etc.), they only use it for query generation.

`lxd-generate` is directly importing 
```
	"github.com/canonical/lxd/lxd/db/cluster"
	"github.com/canonical/lxd/lxd/db/node"
```
to call `cluster.SchemaDotGo` and `node.SchemaDotGo` for generating schemas.

I think it'd be easier to reason about if we take this out of `lxd-generate` and have a small go program call these functions and input the exact filepath we want.

tomponline: @markylaing @masnax tells me that its used by micro* too.

tomponline: I'm happy to have it reworked, but we need it to be working now (its broken right now), and I dont have time to focus on a large rework right now im afraid.

tomponline: @markylaing for context it was broken in this commit https://github.com/canonical/lxd/pull/14212/commits/63b017e84dab2082af8a84b81238b50a82e4e0a3

tomponline: > I think it'd be easier to reason about if we take this out of `lxd-generate` and have a small go program call these functions and input the exact filepath we want.

I dont follow you. It's lxd-generate's job to update these files?

tomponline: We could increase the specificity of `if strings.HasPrefix(filename, "github.com") {` to check for `github.com/canonical/lxd/lxd/db/` if you think that would help with microcloud usage?

markylaing: > @masnax tells me that its used by micro* too.

`micro*` may be using query generation but I don't see how it's possible that they're using `lxd-generate` schema since it doesn't take any arguments, it only generates the LXD schema.

I've just seen the thread for this (I didn't realise it was blocking a PR) so will approve for now and we can discuss more in the new year :)

markylaing (approved):

lxd/db/schema/update.go
tomponline: > `micro*` may be using query generation but I don't see how it's possible that they're using `lxd-generate` schema since it doesn't take any arguments, it only generates the LXD schema.

You may well be right, all I got from @masnax so far that gave me that impression was:

> One small issue might be with goimports which does not properly handle non-v1/v0 go module packages. But that wouldn't affect LXD, only Micro*, which also use lxd-generate

PR #14705, opened 2024-12-20
Remove `runtime.Caller` from `schema.DotGo`

markylaing:
This is an alternative to #14704 

I didn't understand why `runtime.Caller` was required. It was because of where the `SchemaDotGo` functions were defined in the `node` and `cluster` packages.

We can instead just split the `go:generate` directives into their respective packages so that invocation there will cause the `schema.go` files to be created in the current directory.

As an aside, since `lxd-generate` is used by `micro*` packages, I'd also be in favour of removing the `schema` subcommand from it, as it only supports generating the LXD node or cluster schemas.

### DIFF
--- a/lxd/db/cluster/update.go
+++ b/lxd/db/cluster/update.go
@@ -1,5 +1,7 @@
 package cluster
 
+//go:generate lxd-generate db schema cluster
+
 import (
 	"context"
 	"database/sql"
--- a/lxd/db/cluster/update.go
+++ b/lxd/db/cluster/update.go
@@ -30,7 +30,7 @@ func FreshSchema() string {
 // SchemaDotGo refreshes the schema.go file in this package, using the updates
 // defined here.
 func SchemaDotGo() error {
-	return schema.DotGo(updates, "schema")
+	return schema.DotGo(updates, "cluster", "schema.go")
 }
 
 // SchemaVersion is the current version of the cluster database schema.
--- a/lxd/db/generate/db.go
+++ b/lxd/db/generate/db.go
@@ -36,7 +36,11 @@ func newDbSchema() *cobra.Command {
 		Use:   "schema",
 		Short: "Generate database schema by applying updates.",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return db.UpdateSchema()
+			if len(args) < 1 {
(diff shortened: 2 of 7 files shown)


tomponline (approved):
Ah yeah this is nicer. Thanks.

I don't know the history of lxd-generate so dont know why it wasnt done like this originally.

Does this break micro* usage of it though?

tomponline:
> I'd also be in favour of removing the `schema` subcommand from it, as it only supports generating the LXD node or cluster schemas

Removing the schema sub-command from what? micro* or from lxd-generate?

If its removed how will the schema get generated in lxd?

tomponline (approved):

markylaing:
> I don't know the history of lxd-generate so dont know why it wasnt done like this originally.

Who could say!

> Does this break micro* usage of it though?

`lxd-generate db schema` only does one thing, it generates the LXD cluster and node schemas. So if `micro*` are using it I would be quite surprised. 

The `schema.DotGo` function may be used elsewhere but I couldn't find any evidence of this. I don't think any microcluster Apps have a reason for generating a fresh schema because the design is predicated on applying schema updates at startup to the core tables.

> Removing the schema sub-command from what? micro* or from lxd-generate?

From `lxd-generate`.

> If its removed how will the schema get generated in lxd?

You could have a tiny `freshschema` package in `lxd/db/{cluster,node}` that has a main function that calls `(lxd/db/{cluster,node}).SchemaDotGo` and change the generation directive to `//go:generate go run ./freshschema/main.go`.

tomponline:
> You could have a tiny `freshschema` package in `lxd/db/{cluster,node}` that has a main function that calls `(lxd/db/{cluster,node}).SchemaDotGo` and change the generation directive to `//go:generate go run ./freshschema/main.go`

I see.

Could you have a single freshchema program in say, `lxd/db` and call it via `go run ../freshschema/main.go` and have it take an argument for the schema to generate?

markylaing:
> Could you have a single freshchema program in say, `lxd/db` and call it via `go run ../freshschema/main.go` and have it take an argument for the schema to generate?

Yeah I don't see why not :)

tomponline:
> Yeah I don't see why not :)

Cool, go for it, thanks

PR #14700, opened 2024-12-19
Storage: Deduplicate logic between Pure Storage and PowerFlex storage drivers

MusicDin:
This PR introduces storage connectors. The aim of the connector is to handle connections with the target storage. Currently, connectors support NVMe, iSCSI, and SDC. 

NVMe and iSCSI were tested on Pure Storage. Powerflex and SDC have yet to be tested and improved.

---

- [ ] Needs rebase once https://github.com/canonical/lxd/pull/14599 is merged
- [ ] Test PowerFlex

PR #14708, opened 2024-12-20
Storage: Fix PowerFlex NVMe subsystem ENOENT

roosterfish:
In case no NVMe subsystem is connected, it appears that the `/sys/devices/virtual/nvme-subsystem` might not exist.

This can be due to a newer kernel version (saw this happen on `5.15.0-130-generic`).
The fix is to check for ENOENT and only try to parse the subsystems in case the directory exists.

Looks this is already fixed in https://github.com/canonical/lxd/pull/14700/files#diff-715b401f96a8a50ff84a548172cb47b0c2ded1c49ce881ee56a2377319927a97R69.

### DIFF
--- a/lxd/storage/drivers/driver_powerflex_utils.go
+++ b/lxd/storage/drivers/driver_powerflex_utils.go
@@ -1173,55 +1173,56 @@ func (d *powerflex) unmapVolume(vol Volume) error {
 func (d *powerflex) connectNVMeSubsys() (revert.Hook, error) {
 	basePath := "/sys/devices/virtual/nvme-subsystem"
 
-	// Retrieve list of existing NVMe subsystems on this host.
-	directories, err := os.ReadDir(basePath)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


MusicDin (commented):
Thanks, I should adjust that for Pure Storage as well.

There is one linter issue for pre-allocation of `errorStrings` slice. In addition, I just left one comment about reverting connection (although it has no impact).

tomponline:
@roosterfish @MusicDin can we separate the pure driver from the changed in https://github.com/canonical/lxd/pull/14700 so we can include this in LXD (including the fix from this PR) in LXD 5.21.3 before pure lands?

MusicDin:
I've cherry-picked storage connector and PowerFlex changes from #14700 in https://github.com/canonical/lxd/pull/14710

roosterfish:
Can be closed in favor of https://github.com/canonical/lxd/pull/14710.

PR #13550, opened 2024-06-05
Storage: Prevent race when unmapping PowerFlex volume

roosterfish:
https://github.com/canonical/lxd/pull/13073 was meant to resolve an error that happens during the resizing of a volume. As the error wasn't anymore present I thought this has fixed it by unmapping the volume prior to the resize.

But this breaks the `lxc import ...` command as the `driver.SetVolumeQuota()` function isn't expected to unmap/map the volume which might cause the path of the device to be different after the map. This can lead to an error during the unpack operations.

Instead it looks that the actual error comes from a race if LXD doesn't wait for the volume to disappear from the system. A subsequent map operation might pick the path from the old volume that is still to be removed from the system.
For Ceph RBD this is not an issue as we call `rbd unmap` directly on the host and the command presumably blocks until the actual device is removed from the system.

Furthermore during the execution of `lxd-ci/tests/storage-vm` I found that PowerFlex backed volume refreshes (`lxc cp abc remote:def --refresh` and `lxc storage volume cp pool/abc remote:pool/def --refresh` to a remote LXD currently do not work as the volumes receive a new `volatile.uuid` due to the call to `GetNewVolume()` when doing a refresh. The conditions are changed for both custom volume and instance refreshes.
This didn't surface until now as the tests (https://github.com/canonical/lxd-ci/pull/101) were landed after we have merged PowerFlex.

Additionally the `nvme` lock got moved to ensure that multiple LXD storage pools backed by the same PowerFlex storage cluster will not remove the NVMe/TCP connection from each other.

PR #14710, opened 2024-12-20
Storage: Add NVMe and SDC storage connectors

MusicDin:
This PR contains cherry-picks from #14700 to include only connectors and PowerFlex changes (without Pure Storage bits)

The aim of the connector is to handle connections with the target storage. Currently, connectors support NVMe, and SDC.

Once PowerFlex will use `Connect`/`Diconnect`, the following connector functions can be removed:
- `ConnectAll`
- `DisconnectAll`
- `SessionID`

Closes https://github.com/canonical/lxd/pull/14708

### DIFF
--- a/lxd/storage/connectors/connector.go
+++ b/lxd/storage/connectors/connector.go
@@ -13,6 +13,9 @@ const (
 
 	// TypeNVME represents an NVMe/TCP storage connector.
 	TypeNVME string = "nvme"
+
+	// TypeSDC represents Dell SDC storage connector.
+	TypeSDC string = "sdc"
 )
 
 // session represents a connector session that is established with a target.
@@ -53,6 +56,11 @@ func NewConnector(connectorType string, serverUUID string) (Connector, error) {
 	switch connectorType {
 	case TypeNVME:
 		return &connectorNVMe{
+			common: common,
+		}, nil
+
+	case TypeSDC:
+		return &connectorSDC{
 			common: common,
 		}, nil
 
--- a/lxd/storage/connectors/connector.go
+++ b/lxd/storage/connectors/connector.go
@@ -70,3 +70,26 @@ func NewConnector(connectorType string, serverUUID string) (Connector, error) {
 		return nil, fmt.Errorf("Unknown storage connector type %q", connectorType)
 	}
 }
+
+// GetSupportedVersions returns the versions for the given connector types
+// ignoring those that produce an error when version is being retrieved
+// (e.g. due to a missing required tools).
+func GetSupportedVersions(connectorTypes []string) []string {
+	versions := make([]string, 0, len(connectorTypes))
+
+	// Iterate over the provided connector types, and extract
+	// their versions.
+	for _, connectorType := range connectorTypes {
+		connector, err := NewConnector(connectorType, "")
+		if err != nil {
+			continue
+		}
+
+		version, _ := connector.Version()
+		if version != "" {
+			versions = append(versions, version)
+		}
+	}
+
+	return versions
+}
--- a/lxd/storage/connectors/connector.go
+++ b/lxd/storage/connectors/connector.go
@@ -1,0 +1,34 @@
+package connectors
+
+import (
+	"context"
+
+	"github.com/canonical/lxd/shared/revert"
+)
+
+// session represents a connector session that is established with a target.
+type session struct {
+	// id is a unique identifier of the session.
+	id string
+
+	// targetQN is the qualified name of the target.
+	targetQN string
+
+	// addresses is a list of active addresses associated with the session.
+	addresses []string
+}
+
+// Connector represents a storage connector that handles connections through
+// appropriate storage subsystem.
+type Connector interface {
+	Type() string
+	Version() (string, error)
+	QualifiedName() (string, error)
+	LoadModules() error
+	Connect(ctx context.Context, targetQN string, targetAddrs ...string) (revert.Hook, error)
+	ConnectAll(ctx context.Context, targetAddr string) error
+	Disconnect(targetQN string) error
+	DisconnectAll() error
+	SessionID(targetQN string) (string, error)
+	findSession(targetQN string) (*session, error)
+}
--- a/lxd/storage/connectors/connector.go
+++ b/lxd/storage/connectors/connector.go
@@ -2,8 +2,17 @@ package connectors
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/canonical/lxd/shared/revert"
+)
+
+const (
+	// TypeUnknown represents an unknown storage connector.
+	TypeUnknown string = "unknown"
+
+	// TypeNVME represents an NVMe/TCP storage connector.
+	TypeNVME string = "nvme"
 )
 
 // session represents a connector session that is established with a target.
@@ -31,4 +40,25 @@ type Connector interface {
 	DisconnectAll() error
 	SessionID(targetQN string) (string, error)
 	findSession(targetQN string) (*session, error)
+}
+
+// NewConnector instantiates a new connector of the given type.
+// The caller needs to ensure connector type is validated before calling this
+// function, as common (empty) connector is returned for unknown type.
+func NewConnector(connectorType string, serverUUID string) (Connector, error) {
+	common := common{
+		serverUUID: serverUUID,
+	}
+
+	switch connectorType {
+	case TypeNVME:
+		return &connectorNVMe{
+			common: common,
+		}, nil
+
+	default:
+		// Return common connector if the type is unknown. This removes
+		// the need to check for nil or handle the error in the caller.
+		return nil, fmt.Errorf("Unknown storage connector type %q", connectorType)
+	}
 }
--- a/lxd/storage/connectors/connector_nvme.go
+++ b/lxd/storage/connectors/connector_nvme.go
@@ -1,0 +1,258 @@
+package connectors
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"github.com/canonical/lxd/lxd/util"
+	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/revert"
+)
+
+var _ Connector = &connectorNVMe{}
+
+type connectorNVMe struct {
+	common
+}
+
+// Type returns the type of the connector.
+func (c *connectorNVMe) Type() string {
+	return TypeNVME
+}
+
+// Version returns the version of the NVMe CLI.
+func (c *connectorNVMe) Version() (string, error) {
+	// Detect and record the version of the NVMe CLI.
+	out, err := shared.RunCommand("nvme", "version")
+	if err != nil {
+		return "", fmt.Errorf("Failed to get nvme-cli version: %w", err)
+	}
+
+	fields := strings.Split(strings.TrimSpace(out), " ")
+	if strings.HasPrefix(out, "nvme version ") && len(fields) > 2 {
+		return fmt.Sprintf("%s (nvme-cli)", fields[2]), nil
+	}
+
+	return "", fmt.Errorf("Failed to get nvme-cli version: Unexpected output %q", out)
+}
+
+// LoadModules loads the NVMe/TCP kernel modules.
+// Returns true if the modules can be loaded.
+func (c *connectorNVMe) LoadModules() error {
+	err := util.LoadModule("nvme_fabrics")
+	if err != nil {
+		return err
+	}
+
+	return util.LoadModule("nvme_tcp")
+}
+
+// QualifiedName returns a custom NQN generated from the server UUID.
+// Getting the NQN from /etc/nvme/hostnqn would require the nvme-cli
+// package to be installed on the host.
+func (c *connectorNVMe) QualifiedName() (string, error) {
+	return fmt.Sprintf("nqn.2014-08.org.nvmexpress:uuid:%s", c.serverUUID), nil
+}
+
+// Connect establishes a connection with the target on the given address.
+func (c *connectorNVMe) Connect(ctx context.Context, targetQN string, targetAddresses ...string) (revert.Hook, error) {
+	// Connects to the provided target address, if the connection is not yet established.
+	connectFunc := func(ctx context.Context, session *session, targetAddr string) error {
+		if session != nil && slices.Contains(session.addresses, targetAddr) {
+			// Already connected.
+			return nil
+		}
+
+		hostNQN, err := c.QualifiedName()
+		if err != nil {
+			return err
+		}
+
+		_, err = shared.RunCommandContext(ctx, "nvme", "connect", "--transport", "tcp", "--traddr", targetAddr, "--nqn", targetQN, "--hostnqn", hostNQN, "--hostid", c.serverUUID)
+		if err != nil {
+			return fmt.Errorf("Failed to connect to target %q on %q via NVMe: %w", targetQN, targetAddr, err)
+		}
+
+		return nil
+	}
+
+	return connect(ctx, c, targetQN, targetAddresses, connectFunc)
+}
+
+// ConnectAll establishes a connection with all targets available on the given address.
+func (c *connectorNVMe) ConnectAll(ctx context.Context, targetAddr string) error {
+	hostNQN, err := c.QualifiedName()
+	if err != nil {
+		return err
+	}
+
+	_, err = shared.RunCommandContext(ctx, "nvme", "connect-all", "--transport", "tcp", "--traddr", targetAddr, "--hostnqn", hostNQN, "--hostid", c.serverUUID)
+	if err != nil {
+		return fmt.Errorf("Failed to connect to any target on %q via NVMe: %w", targetAddr, err)
+	}
+
+	return nil
+}
+
+// Disconnect terminates a connection with the target.
+func (c *connectorNVMe) Disconnect(targetQN string) error {
+	// Find an existing NVMe session.
+	session, err := c.findSession(targetQN)
+	if err != nil {
+		return err
+	}
+
+	// Disconnect from the NVMe target if there is an existing session.
+	if session != nil {
+		// Do not restrict the context as the operation is relatively short
+		// and most importantly we do not want to "partially" disconnect from
+		// the target, potentially leaving some unclosed sessions.
+		_, err := shared.RunCommandContext(context.Background(), "nvme", "disconnect", "--nqn", targetQN)
+		if err != nil {
+			return fmt.Errorf("Failed disconnecting from NVMe target %q: %w", targetQN, err)
+		}
+	}
+
+	return nil
+}
+
+// DisconnectAll terminates all connections with all targets.
+func (c *connectorNVMe) DisconnectAll() error {
+	_, err := shared.RunCommand("nvme", "disconnect-all")
+	if err != nil {
+		return fmt.Errorf("Failed disconnecting from NVMe targets: %w", err)
+	}
+
+	return nil
+}
+
+// SessionID returns the identifier of a session that matches the targetQN.
+// If no session is found, an empty string is returned.
+func (c *connectorNVMe) SessionID(targetQN string) (string, error) {
+	session, err := c.findSession(targetQN)
+	if err != nil || session == nil {
+		return "", err
+	}
+
+	return session.id, nil
+}
+
+// findSession returns an active NVMe subsystem (referred to as session for
+// consistency across connectors) that matches the given targetQN.
+// If the session is not found, nil is returned.
+//
+// This function handles the distinction between an "inactive" session (with no
+// active controllers/connections) and a completely "non-existent" session. While
+// checking "/sys/class/nvme" for active controllers is sufficient to identify if
+// the session is currently in use, it does not account for cases where a session
+// exists but is temporarily inactive (e.g., due to network issues). Removing
+// such a session during this state would prevent it from automatically
+// recovering once the connection is restored.
+//
+// To ensure we detect "existing" sessions, we first check for the session's
+// presence in "/sys/class/nvme-subsystem", which tracks all associated NVMe
+// subsystems regardless of their current connection state. If such session is
+// found the function determines addresses of the active connections by checking
+// "/sys/class/nvme", and returns a non-nil result (except if an error occurs).
+func (c *connectorNVMe) findSession(targetQN string) (*session, error) {
+	// Base path for NVMe sessions/subsystems.
+	subsysBasePath := "/sys/class/nvme-subsystem"
+
+	// Retrieve the list of existing NVMe subsystems on this host.
+	subsystems, err := os.ReadDir(subsysBasePath)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			// If NVMe subsystems directory does not exist,
+			// there is no sessions.
+			return nil, nil
+		}
+
+		return nil, fmt.Errorf("Failed getting a list of existing NVMe subsystems: %w", err)
+	}
+
+	sessionID := ""
+	for _, subsys := range subsystems {
+		// Get the target NQN.
+		nqnBytes, err := os.ReadFile(filepath.Join(subsysBasePath, subsys.Name(), "subsysnqn"))
+		if err != nil {
+			return nil, fmt.Errorf("Failed getting the target NQN for subystem %q: %w", subsys.Name(), err)
+		}
+
+		// Compare using contains, as targetQN may not be the entire NQN.
+		// For example, PowerFlex targetQN is a substring of the full NQN.
+		if strings.Contains(string(nqnBytes), targetQN) {
+			// Found matching session.
+			sessionID = strings.TrimPrefix(subsys.Name(), "nvme-subsys")
+			break
+		}
+	}
+
+	if sessionID == "" {
+		// No matching session found.
+		return nil, nil
+	}
+
+	session := &session{
+		id:       sessionID,
+		targetQN: targetQN,
+	}
+
+	basePath := "/sys/class/nvme"
+
+	// Retrieve the list of currently active (operational) NVMe controllers.
+	controllers, err := os.ReadDir(basePath)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			// No active connections for any session.
+			return session, nil
+		}
+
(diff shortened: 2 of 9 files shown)


MusicDin:
As previously discussed with @roosterfish, I have also replaced nvme "disconnect-all" and "connect-all" by connecting to just a specific target. This prevents disconnecting all nvme volumes, which would affect other storage drivers as well. However, needs to be checked that NQN is correctly passed.

lxd/storage/connectors/connector_nvme.go
```diff
@@ -0,0 +1,171 @@
...
+
+	if stderr != "" {
+		return fmt.Errorf("Failed to connect to target %q on %q via NVMe: %s", targetQN, targetAddr, stderr)
+	}
+
+	return nil
+}
+
+// ConnectAll establishes a connection with all targets available on the given address.
+func (c *connectorNVMe) ConnectAll(ctx context.Context, targetAddr string) error {
```
roosterfish, line 126: I am about to suggest that we can drop the `ConnectAll` and `DisconnectAll` methods and replace the ones in PowerFlex with the respective `Connect` and `Disconnect`. 
In case of PowerFlex the SDT (NVMe targets) can be queried easily so that shouldn't be too much of an issue.

But I think this can land right after this PR so we don't have to mix up things.

roosterfish: Maybe `ConnectAll` can be kept for convenience, but really PowerFlex shouldn't use `DisconnectAll` anymore.

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -931,19 +914,26 @@ func (d *powerflex) mapVolume(vol Volume) (revert.Hook, error) {
...
-		reverter.Add(cleanup)
+	domain, err := d.client().getProtectionDomain(pool.ProtectionDomainID)
+	if err != nil {
+		return nil, err
+	}
+
+	targetQN := domain.SystemID
+	targetAddr := d.config["powerflex.sdt"]
+
+	err = d.connector().Connect(d.state.ShutdownCtx, targetAddr, targetQN)
```
roosterfish, line 930: `ConnectAll`?

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -931,19 +914,26 @@ func (d *powerflex) mapVolume(vol Volume) (revert.Hook, error) {
...
+	if err != nil {
+		return nil, err
 	}
 
+	reverter.Add(func() { _ = d.connector().Disconnect(targetQN) })
```
roosterfish, line 935: `DisconnectAll`?

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -810,13 +784,17 @@ func (d *powerflex) getVolumeType(vol Volume) powerFlexVolumeType {
 // createNVMeHost creates this NVMe host in PowerFlex.
 func (d *powerflex) createNVMeHost() (string, revert.Hook, error) {
 	var hostID string
-	nqn := d.getHostNQN()
+
+	targetNQN, err := d.connector().QualifiedName()
```
roosterfish, line 788: `hostNQN` as target would be PowerFlex?

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -847,8 +825,13 @@ func (d *powerflex) createNVMeHost() (string, revert.Hook, error) {
 // deleteNVMeHost deletes this NVMe host in PowerFlex.
 func (d *powerflex) deleteNVMeHost() error {
 	client := d.client()
-	nqn := d.getHostNQN()
-	host, err := client.getNVMeHostByNQN(nqn)
+
+	targetNQN, err := d.connector().QualifiedName()
```
roosterfish, line 829: `hostNQN`? Same as above.

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -1114,28 +1041,35 @@ func (d *powerflex) unmapVolume(vol Volume) error {
...
+			if err != nil {
+				return err
+			}
+
+			targetQN := domain.SystemID
+
 			// Disconnect from the NVMe subsystem.
 			// Do this first before removing the host from PowerFlex.
-			err := d.disconnectNVMeSubsys()
+			err = d.connector().Disconnect(targetQN)
```
roosterfish, line 1072: `DisconnectAll`?

roosterfish (changes requested):
Thanks for this, looking great already. 

I'll grab your branch and make some tests on the PowerFlex cluster for validation. I think the only change that has to be performed upfront is using the actual (Connect|Disconnect)All funcs to resemble the original behavior.
When we have merged the PR I can take care of switching to the Connect/Disconnect instead.

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -885,7 +885,7 @@ func (d *powerflex) mapVolume(vol Volume) (revert.Hook, error) {
 
 	switch d.config["powerflex.mode"] {
 	case powerFlexModeNVMe:
-		unlock, err := locking.Lock(d.state.ShutdownCtx, "nvme")
+		unlock, err := locking.Lock(d.state.ShutdownCtx, "storage_powerflex_nvme")
```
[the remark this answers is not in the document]
tomponline: we've got `imageOperationLock(fingerprint string)` which we could have something similar, e.g. `storageConnectLock(connectorName string, driverName string)`

lxd/storage/connectors/connector_nvme.go
MusicDin: Sure, connect-all should not interfere. Will fix

lxd/storage/drivers/driver_powerflex_utils.go
MusicDin: I would avoid DisconnectAll, as previously discussed, this would disconnect all NVMe volumes from the LXD host, which is not desired.

lxd/storage/drivers/driver_powerflex_utils.go
MusicDin: Should be hostNQN. This is mistake

lxd/storage/drivers/driver_powerflex_utils.go
MusicDin: I would avoid DisconnectAll, as previously discussed, this would disconnect all NVMe volumes from the LXD host, which is not desired.

MusicDin:
@roosterfish Thanks for the comments. I've addressed all of them, and reverted fixes for `ConnectAll -> Connect`, and `DisconnectAll -> Disconnect` to retain the previous implementation.

roosterfish:
> I've addressed all of them, and reverted fixes

Thanks, I'll run it through some tests on the PowerFlex env and report back here.

lxd/storage/connectors/utils.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	for _, diskPath := range diskPaths {
+		// Skip the disk if it is only a partition of the actual volume.
+		if strings.Contains(diskPath, "-part") {
+			continue
+		}
+
+		// Skip volumes that do not contain the desired name suffix.
+		if !strings.HasSuffix(diskPath, diskNameSuffix) {
```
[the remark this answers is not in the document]
MusicDin: Good point, accepting an arbitrary function also seems versatile solution for any potential upcoming storage drivers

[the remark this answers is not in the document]
MusicDin: I've added the `devicePathFilter` which is a function that accepts device path, and returns bool indicating whether the path matches the required criteria.

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -946,19 +913,19 @@ func (d *powerflex) mapVolume(vol Volume) (revert.Hook, error) {
...
-			return nil, err
-		}
+	targetAddr := d.config["powerflex.sdt"]
 
-		reverter.Add(cleanup)
+	// Connect to the storage subsystem.
+	// In case of NVMe/TCP, we have to connect after the first mapping was established,
+	// as PowerFlex does not offer any discovery log entries until a volume gets mapped
+	// to the host.
+	err = d.connector().ConnectAll(d.state.ShutdownCtx, targetAddr)
```
roosterfish, line 922: Ok, I made progress in the test suite running those changes on PowerFlex. 
Whilst doing so I found a bug as the environment is quite slow :) But wanted to check if it's related to the changes in this PR, but it's something different.

Anyway the `lxd-ci/tests/storage-vm` is now running until vol resizing which fails. I tried to remember what the error was about as it seems familiar, and if I am not mistaken it has something to do with this line here.

Prior to this change the `connectNVMeSubsys()` func was returning a cleanup hook which only contained the disconnect instructions in case the call to `connectNVMeSubsys` actually caused a connection (as it is idempotent and the connection might already be established).
If we are already connected to the subsystem, the returned cleanup hook didn't contain the disconnect statement. Now it does in any case.
IIRC that was one of the problems I found and fixed in https://github.com/canonical/lxd/pull/13550.
And I assume this is now causing the resizing to fail as LXD rips out the connection whilst trying to resize the vol.
It calls the `SetVolumeQuota` func which tries to `mapVolume`. Before this PR the returned cleanup hook didn't contain the disconnect but now it does. So if I get the timing right the resize succeeds, but then the subsystem gets disconnected but LXD tries to bring up the VM again which fails. 

If you like let's have a short meeting to discuss this. Hope I was able to explain it well enough.

roosterfish: Maybe that is also something to check for the pure driver? Simply start a VM, resize its root vol and the do a `lxc restart <vm>`. This should cause the error if pure is affected too.

lxd/storage/connectors/connector.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	switch connectorType {
+	case TypeNVME:
+		return &connectorNVMe{
+			common: common,
+		}
+
+	case TypeSDC:
+		return &connectorNVMe{
```
roosterfish, line 47: Should be `connectorSDC` :)

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -946,17 +915,36 @@ func (d *powerflex) mapVolume(vol Volume) (revert.Hook, error) {
 		reverter.Add(func() { _ = client.deleteHostVolumeMapping(hostID, volumeID) })
```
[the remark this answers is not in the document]
roosterfish: Here is the summary of changes:

```go
    // Add this new outerRevert which handles reverts we can safely pass to the caller.
	outerRevert := revert.New()
	defer outerRevert.Fail()

	if !mapped {
		err = client.createHostVolumeMapping(hostID, volumeID)
		if err != nil {
			return nil, err
		}

        // Run umapVolume instead as it contains the logic required to know if the caller can still delete the NVMe host and close the NVMe/TCP connection. Furthermore it aquires the lock to safely perform those actions.
		outerRevert.Add(func() { _ = d.unmapVolume(vol) })
	}

    ...

    // Get the cleanup from the outerRevert
    cleanup := outerRevert.Clone().Fail

    // Set success for the inner revert as deletion of the host is something we don't want the caller to run as it might already affect other volumes as we have given up the lock between returning the cleanup and actually calling it.
	reverter.Success()
	outerRevert.Success()
	return cleanup, nil
```

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -946,17 +915,36 @@ func (d *powerflex) mapVolume(vol Volume) (revert.Hook, error) {
...
+	// Connect to the storage subsystem.
+	// In case of NVMe/TCP, we have to connect after the first mapping was established,
+	// as PowerFlex does not offer any discovery log entries until a volume gets mapped
+	// to the host.
+	err = d.connector().ConnectAll(d.state.ShutdownCtx, targetAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	if sessionID == "" {
```
roosterfish, line 945: As we add `unmapVolume` above in the reverter we don't need to add the disconnect to the reverter as this is handled within `unmapVolume` after properly acquiring the lock.

roosterfish (changes requested):
Thanks for the short call, here a summary of what we have discussed:

roosterfish (approved):
Thanks, just passed lxd-ci using nvme mode. LGTM!

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -946,22 +915,27 @@ func (d *powerflex) mapVolume(vol Volume) (revert.Hook, error) {
...
 	}
 
-	cleanup := reverter.Clone().Fail
+	// Reverting mapping or connection outside mapVolume function
+	// could conflict with other ongoing operations as lock will
+	// already be released. Therefore, use unmapVolume instead
+	// because it ensures the lock is acquired and accounts for
+	// an existing session before unmapping a volume.
+	outerReverter := revert.New()
+	outerReverter.Add(func() { _ = d.unmapVolume(vol) })
```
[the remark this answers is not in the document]
tomponline: My understanding from @MusicDin is that `unapVolume` already handles that logic?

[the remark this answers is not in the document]
roosterfish: I have just discussed this with @MusicDin. 
Within `unmapVolume` you have already lost the information whether or not you or somebody else has mapped the volume. 
All you can see is if the volume is is mapped to the host or not. 
Whereas in `mapVolume` you still have the knowledge if you actually performed the mapping because you obtain the lock and then check if the volume is currently mapped to the host.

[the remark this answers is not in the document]
MusicDin: In some cases, such as online resizing of the volume, we attempt to map the volume even if it is already mapped.

If an error occurs in the above case, assuming `mapVolume` always returns `unmapVolume` in the reverter, it would cause the host to be disassociated from the volume (on remote storage). As a result the volume would disappear from the host even if the connection with the remote storage is still established. (disassociating the volume can be though of as removing permission from the client to access the volume, not breaking the connection)

However, `unmapVolume` will only prevents disconnection if there are still volumes associated with the host (after the desired one is disassociated).

With that being said, @roosterfish is correct, and we should only ever revert the mapping if the volume was just mapped.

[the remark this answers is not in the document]
tomponline: @MusicDin @roosterfish so whats the outcome?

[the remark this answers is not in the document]
tomponline: and in other cases we return an empty reverter that does nothing right?

roosterfish:
NVMe Test passed with latest changes!

roosterfish (approved):

lxd/storage/connectors/utils.go
```diff
@@ -0,0 +1,128 @@
...
+			return "", err
+		}
+
+		// If the device is found, return the device path.
+		if diskPath != "" {
+			break
+		}
+
+		// Check if context is cancelled.
+		if ctx.Err() != nil {
```
tomponline, line 65: small nit that we are calling ctx.Err() twice unnecessarily when it is non-nil.

lxd/storage/connectors/utils.go
```diff
@@ -126,3 +130,120 @@ func WaitDiskDeviceGone(ctx context.Context, diskPath string) bool {
...
+		cancel()
+
+		// Wait until all connection attempts have finished.
+		wg.Wait()
+
+		// If no active session was found, ensure all established
+		// connections are closed. Otherwise, keep open connections
+		// intact for other volumes using the session.
+		if session == nil {
+			revert.Add(func() { _ = c.Disconnect(targetQN) })
```
tomponline, line 225: @MusicDin this looks wrong, you're adding stuff to the reverter that only is added after the revert.Fail function is run.

lxd/storage/connectors/utils.go
```diff
@@ -126,3 +130,120 @@ func WaitDiskDeviceGone(ctx context.Context, diskPath string) bool {
...
+		}
+	})
+
+	// Wait until at least one successful connections is established, or
+	// exit if all connections fail.
+	doneConns := 0
+	for {
+		success := <-doneChan
```
tomponline, line 233: When the go routine that calls `wg.Wait()` ends, it closes `doneChan` which will cause all reads from that channel to return `false` (the default bool value), so I think we can just `range` over this channel, and return the first true value or wait until the for range loop ends (due to the channel being closed), and then there's no need for the `doneConns` logic.

tomponline: So like this (also suggest rename to resChan for clarity, as `doneChan` usually indicates its closed when done only):

```go
for success := range resChan {
     if success {
        return nil // Successfully connected to one endpoint
     }
}

return nil, fmt.Errorf("Failed to connect to any address on target %q", targetQN)
```

lxd/storage/connectors/connector.go
```diff
@@ -37,3 +40,24 @@ type Connector interface {
...
+		serverUUID: serverUUID,
+	}
+
+	switch connectorType {
+	case TypeNVME:
+		return &connectorNVMe{
+			common: common,
+		}
+
+	default:
```
tomponline, line 66: why wouldn't we expect the caller to do this, or for this function to return an error if an invalid connectorType string is provided? Seems like its going to go wrong sooner or later anyway, and sooner is preferable right?

lxd/storage/connectors/connector_nvme.go
```diff
@@ -0,0 +1,261 @@
...
+		if err != nil {
+			return err
+		}
+
+		_, stderr, err := shared.RunCommandSplit(ctx, nil, nil, "nvme", "connect", "--transport", "tcp", "--traddr", targetAddr, "--nqn", targetQN, "--hostnqn", hostNQN, "--hostid", c.serverUUID)
+		if err != nil {
+			return fmt.Errorf("Failed to connect to target %q on %q via NVMe: %w", targetQN, targetAddr, err)
+		}
+
+		if stderr != "" {
```
tomponline, line 82: could do with a comment here explaining in what scenarios we'd expect to get a zero (success) exit code with non-empty stderr?

lxd/storage/connectors/connector_nvme.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, stderr, err := shared.RunCommandSplit(ctx, nil, nil, "nvme", "connect-all", "--transport", "tcp", "--traddr", targetAddr, "--hostnqn", hostNQN, "--hostid", c.serverUUID)
+	if err != nil {
+		return fmt.Errorf("Failed to connect to any target on %q via NVMe: %w", targetAddr, err)
+	}
+
+	if stderr != "" {
```
tomponline, line 104: could do with a comment here explaining in what scenarios we'd expect to get a zero (success) exit code with non-empty stderr?

lxd/storage/connectors/utils.go
```diff
@@ -126,3 +130,120 @@ func WaitDiskDeviceGone(ctx context.Context, diskPath string) bool {
...
+	//
+	// Context cancellation is not deferred here to ensure connection attempts
+	// continue even if the function exits before all attempts are completed.
+	timeoutCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+
+	var wg sync.WaitGroup
+	doneChan := make(chan bool, len(targetAddrs))
+
+	// Connect to all target addresses.
+	for _, addr := range targetAddrs {
```
tomponline, line 188: I think I'd be tempted to move this into the same go routine as the cleanup routine below so that its keeps these parts locked together and more readable, e.g.

```go
go func() {
	// Connect to all target addresses.
	for _, addr := range targetAddrs {
          go func(addr string) {
           ....
          })(addr)
    }

        wg.Wait()
		close(doneChan)
		cancel()
		unlock()
}()
```

lxd/storage/connectors/utils.go
tomponline: Further more if you made `resChan` type `error` you could do this, and return the last error to the caller.

```go
var err error
for err = range resChan {
     if err == nil {
        return nil // Successfully connected to one endpoint
     }
}

return fmt.Errorf("Failed to connect to any address on target %q: %w", targetQN, err)
```

lxd/storage/drivers/utils.go
```diff
@@ -911,3 +912,15 @@ func ResolveServerName(serverName string) (string, error) {
...
+// remoteVolumeMapLock acquires a lock used when mapping or unmapping remote
+// storage volumes. This lock prevents conflicts between operations trying to
+// associate or disassociate volumes with the LXD host. If the lock is
+// successfully acquired, unlock function is returned.
+func remoteVolumeMapLock(connectorName string, driverName string) (locking.UnlockFunc, error) {
+	l := logger.AddContext(logger.Ctx{"connector": connectorName, "driver": driverName})
+	l.Debug("Acquiring lock for storage connector")
+	defer l.Debug("Lock acquired for storage connector")
+
+	return locking.Lock(context.TODO(), fmt.Sprintf("StorageConnector_%s_%s", connectorName, driverName))
```
tomponline, line 908: Can the logging and lock name be updated to reflect the new function name please.

lxd/storage/drivers/driver_powerflex_utils.go
tomponline: would be good to have constants for these so we can see where they are being referenced?

tomponline (commented):
Nearly there, few small changes to make. Ta

lxd/storage/connectors/utils.go
MusicDin: Oops

lxd/storage/connectors/connector_nvme.go
MusicDin: To be sincere, I am not sure if there is any such case. Since I never ran into any issues with that, I assumed that it will either be skipped immediately or show some other underlying error.

But now that you have pointed it out, I think this can make more harm than good in case some warnings are printed to stderr and the command exits successfully. This would result in count as the *connection failed* although if exit status is 0, I quite sure the connection succeeded.

@roosterfish Do you have any specific use case for that, otherwise I'll happily remove this.

lxd/storage/drivers/driver_powerflex_utils.go
MusicDin: This commit can be dropped. It is a leftover, as this is fixed in one of the following commits using `remoteVolumeMapLock`

lxd/storage/connectors/connector_nvme.go
roosterfish: I don't remember why this was useful. But I chose `RunCommandSplit` for a reason to explicitly get `stderr`. 
Maybe it was for testing purposes so I would say let's get rid of it until we find a use case we can then document.

lxd/storage/connectors/connector.go
MusicDin: The main idea behind that was to avoid returning error from `driver.Connector()`. The mode is always checked before initializing the connector, and to avoid checking that error is not nil (or ignoring it) across utils function, I have instead returned the common connector in such case. The common connector errors out for almost all connector functions anyway and returns type "unknown".

In short to be able to do:
```go
if d.connector().Type() == connectors.NVME {
   ...
}

or 

switch d.connector().Type() {
 ...
}
```

Instead of:
```go
con, err := d.connector().Type()
if err != nil {
  return err
}

if con.Type() == connectors.NVME {
   ...
}

or 

con, _ := d.connector().Type()
switch con.Type() {
 ...
}
```

If the error is not ignored, we would always need to either  "nil" error

It allows then to do:
```
err := NewConnector(conntectors.TypeNVME).LoadModules()
```

So if it is acceptable, we can return error from `Connector()` and still ignore it in `driver.connector()` wrapper?

tomponline: My view is that `NewConnector` should return nil with an error if an invalid type is provided, like the driver loaders do. https://github.com/canonical/lxd/blob/main/lxd/instance/drivers/load.go#L65

lxd/storage/connectors/utils.go
```diff
@@ -127,3 +131,113 @@ func WaitDiskDeviceGone(ctx context.Context, diskPath string) bool {
...
+		unlock()
+	}()
+
+	// Revert successful connections in case of an unexpected error.
+	revert.Add(func() {
+		// Cancel the context to immediately stop all connection attempts.
+		cancel()
+
+		// Wait until all connection attempts have finished.
+		wg.Wait()
```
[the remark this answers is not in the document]
MusicDin: Hmm, this one is tricky. It *could* happen. But how long should we keep it?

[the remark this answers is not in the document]
MusicDin: Having a counter would be a viable solution. But when do we release the initial lock/reduce the counter

lxd/storage/connectors/utils.go
```diff
@@ -0,0 +1,243 @@
...
+
+		// If no active session was found, ensure all established
+		// connections are closed. Otherwise, keep open connections
+		// intact for other volumes using the session.
+		if session == nil {
+			_ = c.Disconnect(context.Background(), targetQN)
```
roosterfish, line 226: Instead should this be the `ctx` passed inside `connect()`?

lxd/storage/connectors/utils.go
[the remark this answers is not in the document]
tomponline: @MusicDin @roosterfish and I discussed it and we'll merge this as-is, and then @roosterfish is going to see if he can address this with his other up coming PowerFlex fixes.

That way we can unblock the Pure driver.

lxd/storage/connectors/utils.go
MusicDin: I don't think so because the caller may use context with shorter timeout for connect, so we would never run disconnect if the timeout is reached.

tomponline: no, i dont think so because it'll be guaranteed cancelled as its cancelled above on line 217.

MusicDin: I would keep this question open until we figure out the previous issue, which may solve this one as well - we may end up not returning the reverter at all (or at least not disconnecting in the reverter)

roosterfish (approved):
Ran another bunch of tests on NVMe/SDC with VMs and containers. All good and cleaned up afterwards. LGTM!

tomponline (approved):
LGTM thanks!

PR #14714, opened 2024-12-20
Misc fixes

simondeziel:


lxd/db/instances.go
```diff
@@ -559,7 +556,7 @@ func (c *ClusterTx) instanceProfilesFillWithProfiles(ctx context.Context, snapsh
 
 		first = false
 
-		q.WriteString(fmt.Sprintf("%d", instanceID))
+		q.WriteString(fmt.Sprint(instanceID))
```
tomponline, line 559: strconv.Itoa maybe better here?

doc/instance-exec.md
```diff
@@ -87,7 +87,7 @@ Non-interactive mode
 
       lxc query --request DELETE /1.0/instances/<instance_name>/logs/exec-output/<record-output-file>
 
-  See [`GET /1.0/instances/{name}/logs/exec-output`](swagger:/instances/instance_exec-outputs_get), [`GET /1.0/instances/{name}/logs/exec-output/{filename}`](swagger:/instances/instance_exec-output_get), and [`DELETE /1.0/instances/{name}/logs/exec-output/{filename}`](swagger:/instances/instance_exec-output_delete) for more information.
+  See [`GET /1.0/instances/{name}/logs/exec-output`](swagger:/instances/instance_exec-output_get), [`GET /1.0/instances/{name}/logs/exec-output/{filename}`](swagger:/instances/instance_exec-output_get), and [`DELETE /1.0/instances/{name}/logs/exec-output/{filename}`](swagger:/instances/instance_exec-output_delete) for more information.
```
minaelee, line 90: Changing `instance_exec-outputs_get` to `instance_exec-output_get` would result in a broken link, since the `operationId` is still `instance_exec-outputs_get` in `rest-api.yaml` - see [line 10910 in that file](https://github.com/canonical/lxd/blob/393216c4d66c0a65969c65fa8b0bd372b985b56f/doc/rest-api.yaml#L10910).

I tried to confirm this by testing in the RTD preview. The link was broken. However, I realized after further testing that ALL links written using `swagger:` are non-functioning in the RTD preview (either an RTD issue or an issue with the Swagger plugin), so that doesn't help to prove that this PR breaks the link. 

Instead, you can see that if you go to the API from the RTD preview via the sidebar nav instead of from the `swagger:` link and look for the specific endpoint, it's found at: https://canonical-ubuntu-documentation-library--14714.com.readthedocs.build/lxd/en/14714/api/#/instances/instance_exec-outputs_get <-- see that it says **`outputs_get`**

Thus, you will need to update the operationId in rest-api.yaml!

lxd/db/instances.go
```diff
@@ -106,16 +106,15 @@ func (c *ClusterTx) GetNodeAddressOfInstance(ctx context.Context, project string
 		args = append(args, instType)
 	}
 
-	if strings.Contains(name, shared.SnapshotDelimiter) {
-		parts := strings.SplitN(name, shared.SnapshotDelimiter, 2)
-
+	instanceName, snapshotName, _ := strings.Cut(name, shared.SnapshotDelimiter)
```
tomponline, line 110: This has subtly changed behaviour?

Previously passing `foo/` as `name` would have caused the first if statement part to be run, using "" as the snapshot name.  Which I think would have then returned a not found error as no snapshot named "" exists.

But now using `foo/` will return an empty snapshotName, and the new logic will then use the else block with the instance name, which would find the "foo" instance potentially.

Instead, how about we use the third return value from strings.Cut as the condition for the if statement?

lxc/completion.go
```diff
@@ -188,9 +188,8 @@ func (g *cmdGlobal) cmpImages(toComplete string) ([]string, cobra.ShellCompDirec
 	var remote string
```
tomponline, line 188: no need for this line now.

lxd/instances_post.go
```diff
@@ -924,7 +924,7 @@ func setupInstanceArgs(s *state.State, instType instancetype.Type, projectName s
 				break
 			}
 
-			rootDevName = fmt.Sprintf("root%d", i)
+			rootDevName = fmt.Sprint("root", i)
```
tomponline, line 927: maybe "root" + strconv.ItoA(i)`?

tomponline (changes requested):
Looks good, just some minor changes.

tomponline (approved):

PR #14744, opened 2025-01-06
VM: Avoid `Invalid compat argument` from `virtiofsd`

hamistao:
When using a `=` in a device path, as in `lxc storage volume attach default vol vm /mnt/path=`, we get a stale operation for that VM, the disk device set up freezes midway and further operations on that VM become impossible and always halts. Upon looking the `virtiofsd` logs we can see something like: `error: Invalid compat argument '-o source=/var/snap/lxd/common/lxd/devices/vm/disk.vol.path='`. As a fix, this PR escapes '=' from the source path for VMs.

### DIFF
--- a/lxd/device/disk.go
+++ b/lxd/device/disk.go
@@ -1637,12 +1637,9 @@ func (d *disk) mountPoolVolume() (func(), string, *storagePools.MountInfo, error
 // createDevice creates a disk device mount on host.
 // The srcPath argument is the source of the disk device on the host.
 // Returns the created device path, and whether the path is a file or not.
-func (d *disk) createDevice(srcPath string) (func(), string, bool, error) {
+func (d *disk) createDevice(srcPath string, devPath string) (func(), bool, error) {
 	revert := revert.New()
 	defer revert.Fail()
-
-	// Paths.
-	devPath := d.getDevicePath(d.name, d.config)
 
 	isReadOnly := shared.IsTrue(d.config["readonly"])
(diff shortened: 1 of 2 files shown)


lxd/device/disk.go
```diff
@@ -1637,13 +1637,10 @@ func (d *disk) mountPoolVolume() (func(), string, *storagePools.MountInfo, error
 // createDevice creates a disk device mount on host.
 // The srcPath argument is the source of the disk device on the host.
 // Returns the created device path, and whether the path is a file or not.
-func (d *disk) createDevice(srcPath string) (func(), string, bool, error) {
+func (d *disk) createDevice(srcPath string, devPath string) (func(), bool, error) {
```
tomponline, line 1645: as an aside, the createDevice function returns a revert hook `func()` but to make this clearer please can you change the type of the return value to `revert.Hook`.

tomponline (changes requested):
@hamistao thanks for this

As discussed lets explore probing the virtiofsd version at LXD start, and storing its location, as well as an indicator if it is the newer unbundled version or the older bundled version. Then we can switch to using the non-deprecated flags that dont exhibit this issue.

We have a struct called `type OS struct` that is used to store stuff like this, e.g. `AppArmorFeatures`, so we could probe and store it there.

tomponline:
@hamistao shall we close this for now?

tomponline:
@hamistao please can you create a GH issue to track the problem you found.

tomponline:
> so is it necessary

Yes please so its tracked, ta

PR #14745, opened 2025-01-06
auth: Add entitlements to LXD entities (part 1: introduce `IsFineGrained` field)

gabrielmougard:
JIRA ticket: https://warthogs.atlassian.net/browse/LXD-2208
Specification link: https://docs.google.com/document/d/1GxWV5J57MLrjGEY5RDG7eS86J99A8RMmjKTpn0mEhgY/edit?tab=t.0

____

This is the first part of a group of three stacked PRs.

* Second part: https://github.com/canonical/lxd/pull/14746
* Third part: https://github.com/canonical/lxd/pull/14748

____

This introduces a new `FineGrained` boolean field in `IdentityInfo` (returned from `GET /1.0/auth/identities/current`) to let a user know if he is currently using a fine grained authentication method. This PR also introduce the `entities_with_entitlements` extension. 

____

## How to query the API using fine grained auth (this is just a reminder for me)

```bash
# Add TLS certificate to LXD (this can be done through LXD UI. Download the .pfx file)
# Create an auth group
lxc auth group create test-group

# Add an new identity to this group ("lxd-ui" is my TLS certificate identifier created by LXD UI)
lxc auth identity group add tls/lxd-ui test-group-1

# Create resources and add permissions to them
lxc auth group permission add test-group-1 instance c1 can_view project=default

# Create a .pfx certificate (with password "1234") and associate it to an auth group and give it some permissions. You can call the API like the following to perform fine-grained calls:
curl --insecure --cert-type P12 --cert lxd-ui-fine-grained-tls.pfx:1234 "https://127.0.0.1:8443/1.0/instances/c1?recursion=1&with-entitlements=can_view" | jq '.'
```

The CURL output should give you something like:

```json
{
  "type": "sync",
  "status": "Success",
  "status_code": 200,
  "operation": "",
  "error_code": 0,
  "error": "",
  "metadata": {
    "entitlements": [
      "can_view"
    ],
    "name": "c1",
    "description": "this is a test ",
    "status": "Running",
    "status_code": 103,
    "created_at": "2024-11-12T09:58:54.960586253Z",
    "last_used_at": "2024-11-16T12:15:13.572661467Z",
    "location": "none",
    "type": "container",
    ...
  }
}
```

PR #14748, opened 2025-01-06
auth: Add entitlements to LXD entities (part 2: Enrich LXD resources with entitlements)

gabrielmougard:
This is the second part of a group of three stacked PRs.

* First part: https://github.com/canonical/lxd/pull/14745

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2557,3 +2557,10 @@ disk devices.
 
 Introduces per-project uplink IP limits for each available uplink network, adding `limits.networks.uplink_ips.ipv4.NETWORK_NAME` and `limits.networks.uplink_ips.ipv6.NETWORK_NAME` configuration keys for projects with `features.networks` enabled.
 These keys define the maximum value of IPs made available on a network named NETWORK_NAME to be assigned as uplink IPs for entities inside a certain project. These entities can be other networks, network forwards or load balancers.
+
+## `entities_with_entitlements`
+
+Adds `fine_grained` field to `GET /1.0/auth/identities/current` to indicate if the current identity
+interacting with the LXD API is fine-grained (that is, associated permissions are managed via group membership).
+Allows LXD entities to be returned with an `access_entitlements` field if the current identity is fine-grained and the
+GET request to fetch the LXD entities has the `with-access-entitlements=<comma_separated_list_of_candidate_entitlements>` query parameter.
--- a/lxd/api_project.go
+++ b/lxd/api_project.go
@@ -144,6 +144,10 @@ func projectsGet(d *Daemon, r *http.Request) response.Response {
 	s := d.State()
 
 	recursion := util.IsRecursionRequest(r)
+	withEntitlements, err := extractEntitlementsFromQuery(r, entity.TypeProject, true)
+	if err != nil {
+		return response.SmartError(err)
+	}
 
 	userHasPermission, err := s.Authorizer.GetPermissionChecker(r.Context(), auth.EntitlementCanView, entity.TypeProject)
 	if err != nil {
@@ -199,6 +203,19 @@ func projectsGet(d *Daemon, r *http.Request) response.Response {
 
 	for _, apiProject := range apiProjects {
 		apiProject.UsedBy = projecthelpers.FilterUsedBy(s.Authorizer, r, apiProject.UsedBy)
+	}
+
+	if len(withEntitlements) > 0 {
+		urlToProject := make(map[*api.URL]auth.EntitlementReporter, len(apiProjects))
+		for _, p := range apiProjects {
+			u := entity.ProjectURL(p.Name)
+			urlToProject[u] = p
+		}
+
+		err = reportEntitlements(r.Context(), s.Authorizer, s.IdentityCache, entity.TypeProject, withEntitlements, urlToProject)
+		if err != nil {
+			return response.SmartError(err)
+		}
 	}
 
 	return response.SyncResponse(true, apiProjects)
@@ -397,6 +414,11 @@ func projectGet(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	withEntitlements, err := extractEntitlementsFromQuery(r, entity.TypeProject, false)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
 	// Get the database entry
 	var project *api.Project
 	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
@@ -415,6 +437,13 @@ func projectGet(d *Daemon, r *http.Request) response.Response {
 	})
 	if err != nil {
 		return response.SmartError(err)
+	}
+
+	if len(withEntitlements) > 0 {
+		err = reportEntitlements(r.Context(), s.Authorizer, s.IdentityCache, entity.TypeProject, withEntitlements, map[*api.URL]auth.EntitlementReporter{entity.ProjectURL(name): project})
+		if err != nil {
+			return response.SmartError(err)
+		}
 	}
 
 	etag := []any{
--- a/lxd/auth/types.go
+++ b/lxd/auth/types.go
@@ -35,6 +35,13 @@ const (
 // It is returned by Authorizer.GetPermissionChecker.
 type PermissionChecker func(entityURL *api.URL) bool
 
+// EntitlementReporter is an interface for adding entitlements to an entity.
+type EntitlementReporter interface {
+	// ReportEntitlements adds entitlements to the entity.
+	// Note: this needs to be a list of string because the implementations of this method will be for the API types.
+	ReportEntitlements([]string)
+}
+
 // Authorizer is the primary external API for this package.
 type Authorizer interface {
 	// Driver returns the driver name.
--- a/lxd/auth_groups.go
+++ b/lxd/auth_groups.go
@@ -163,7 +163,7 @@ func validateGroupName(name string) error {
 //	  "500":
 //	    $ref: "#/responses/InternalServerError"
 func getAuthGroups(d *Daemon, r *http.Request) response.Response {
-	recursion := request.QueryParam(r, "recursion")
+	recursion := util.IsRecursionRequest(r)
 	s := d.State()
 
 	canViewGroup, err := s.Authorizer.GetPermissionChecker(r.Context(), auth.EntitlementCanView, entity.TypeAuthGroup)
@@ -179,6 +179,11 @@ func getAuthGroups(d *Daemon, r *http.Request) response.Response {
 	canViewIDPGroup, err := s.Authorizer.GetPermissionChecker(r.Context(), auth.EntitlementCanView, entity.TypeIdentityProviderGroup)
 	if err != nil {
 		return response.SmartError(fmt.Errorf("Failed to get a permission checker: %w", err))
+	}
+
+	withEntitlements, err := extractEntitlementsFromQuery(r, entity.TypeAuthGroup, true)
+	if err != nil {
+		return response.SmartError(err)
 	}
 
 	var groups []dbCluster.AuthGroup
@@ -203,7 +208,7 @@ func getAuthGroups(d *Daemon, r *http.Request) response.Response {
 			return nil
 		}
 
-		if recursion == "1" {
+		if recursion {
 			// If recursing, we need all identities for all groups, all IDP groups for all groups,
 			// all permissions for all groups, and finally the URLs that those permissions apply to.
 			groupsIdentities, err = dbCluster.GetAllIdentitiesByAuthGroupIDs(ctx, tx.Tx())
@@ -234,13 +239,14 @@ func getAuthGroups(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
-	if recursion == "1" {
+	if recursion {
 		authGroupPermissionsByGroupID := make(map[int][]dbCluster.Permission, len(groups))
 		for _, permission := range authGroupPermissions {
 			authGroupPermissionsByGroupID[permission.GroupID] = append(authGroupPermissionsByGroupID[permission.GroupID], permission)
 		}
 
 		apiGroups := make([]api.AuthGroup, 0, len(groups))
+		urlToGroup := make(map[*api.URL]auth.EntitlementReporter, len(groups))
 		for _, group := range groups {
 			var apiPermissions []api.Permission
 
@@ -272,13 +278,23 @@ func getAuthGroups(d *Daemon, r *http.Request) response.Response {
 				}
 			}
 
-			apiGroups = append(apiGroups, api.AuthGroup{
+			group := api.AuthGroup{
 				Name:                   group.Name,
 				Description:            group.Description,
 				Permissions:            apiPermissions,
 				Identities:             apiIdentities,
 				IdentityProviderGroups: idpGroups,
-			})
+			}
+
+			apiGroups = append(apiGroups, group)
+			urlToGroup[entity.AuthGroupURL(group.Name)] = &group
+		}
+
+		if len(withEntitlements) > 0 {
+			err = reportEntitlements(r.Context(), s.Authorizer, s.IdentityCache, entity.TypeAuthGroup, withEntitlements, urlToGroup)
+			if err != nil {
+				return response.SmartError(err)
+			}
 		}
 
 		return response.SyncResponse(true, apiGroups)
@@ -406,6 +422,11 @@ func getAuthGroup(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	withEntitlements, err := extractEntitlementsFromQuery(r, entity.TypeAuthGroup, false)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
 	ctx, cancel := context.WithTimeout(r.Context(), 10*time.Second)
 	defer cancel()
 
@@ -436,6 +457,13 @@ func getAuthGroup(d *Daemon, r *http.Request) response.Response {
 	})
 	if err != nil {
 		return response.SmartError(err)
+	}
+
+	if len(withEntitlements) > 0 {
+		err = reportEntitlements(r.Context(), s.Authorizer, s.IdentityCache, entity.TypeAuthGroup, withEntitlements, map[*api.URL]auth.EntitlementReporter{entity.AuthGroupURL(groupName): apiGroup})
+		if err != nil {
+			return response.SmartError(err)
+		}
 	}
 
 	return response.SyncResponseETag(true, *apiGroup, *apiGroup)
--- a/lxd/certificates.go
+++ b/lxd/certificates.go
@@ -143,9 +143,15 @@ func certificatesGet(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	withEntitlements, err := extractEntitlementsFromQuery(r, entity.TypeCertificate, true)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
 	if recursion {
 		var certResponses []api.Certificate
 		var baseCerts []dbCluster.Certificate
+		urlToCertificate := make(map[*api.URL]auth.EntitlementReporter)
 		var err error
 		err = d.State().DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
 			baseCerts, err = dbCluster.GetCertificates(ctx, tx.Tx())
@@ -165,12 +171,20 @@ func certificatesGet(d *Daemon, r *http.Request) response.Response {
 				}
 
 				certResponses = append(certResponses, *apiCert)
+				urlToCertificate[entity.CertificateURL(apiCert.Fingerprint)] = apiCert
 			}
 
 			return nil
 		})
 		if err != nil {
 			return response.SmartError(err)
+		}
+
+		if len(withEntitlements) > 0 {
+			err = reportEntitlements(r.Context(), s.Authorizer, s.IdentityCache, entity.TypeCertificate, withEntitlements, urlToCertificate)
+			if err != nil {
+				return response.SmartError(err)
+			}
 		}
 
 		return response.SyncResponse(true, certResponses)
@@ -768,6 +782,11 @@ func certificateGet(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	withEntitlements, err := extractEntitlementsFromQuery(r, entity.TypeCertificate, false)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
 	var cert *api.Certificate
 	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
 		dbCertInfo, err := dbCluster.GetCertificateByFingerprintPrefix(ctx, tx.Tx(), fingerprint)
@@ -785,6 +804,13 @@ func certificateGet(d *Daemon, r *http.Request) response.Response {
 	err = s.Authorizer.CheckPermission(r.Context(), entity.CertificateURL(cert.Fingerprint), auth.EntitlementCanView)
 	if err != nil {
 		return response.SmartError(err)
+	}
+
+	if len(withEntitlements) > 0 {
+		err = reportEntitlements(r.Context(), s.Authorizer, s.IdentityCache, entity.TypeCertificate, withEntitlements, map[*api.URL]auth.EntitlementReporter{entity.CertificateURL(cert.Fingerprint): cert})
+		if err != nil {
+			return response.SmartError(err)
+		}
 	}
 
 	return response.SyncResponseETag(true, cert, cert)
--- a/lxd/daemon.go
+++ b/lxd/daemon.go
@@ -416,6 +416,33 @@ func reportEntitlements(ctx context.Context, authorizer auth.Authorizer, identit
 	return nil
 }
 
+// extractEntitlementsFromQuery extracts the entitlements from the query string of the request.
+func extractEntitlementsFromQuery(r *http.Request, entityType entity.Type, allowRecursion bool) ([]auth.Entitlement, error) {
+	rawEntitlements := request.QueryParam(r, "with-access-entitlements")
+	if rawEntitlements == "" {
+		return nil, nil
+	}
+
+	allowedEntitlements := auth.EntityTypeToEntitlements[entityType]
+	entitlements := strings.Split(rawEntitlements, ",")
+	validEntitlements := make([]auth.Entitlement, 0, len(entitlements))
+	for _, e := range entitlements {
+		if !shared.ValueInSlice(auth.Entitlement(e), allowedEntitlements) {
+			return nil, api.StatusErrorf(http.StatusBadRequest, "Requested entitlement %q is not valid for entity type %q", e, entityType)
+		}
+
+		validEntitlements = append(validEntitlements, auth.Entitlement(e))
+	}
+
+	// Entitlements can only be requested when recursion is enabled for a request returning multiple entities (this function call uses `allowRecursion=true`).
+	// If the request is meant to return a single entity, the entitlements can be requested regardless of the recursion setting (in this case, the function is called with `allowRecursion=false`).
+	if len(validEntitlements) > 0 && (!util.IsRecursionRequest(r) && allowRecursion) {
+		return nil, fmt.Errorf("Entitlements can only be requested when recursion is enabled")
+	}
+
+	return validEntitlements, nil
+}
+
 // Authenticate validates an incoming http Request
 // It will check over what protocol it came, what type of request it is and
 // will validate the TLS certificate or OIDC token.
--- a/lxd/daemon.go
+++ b/lxd/daemon.go
@@ -350,6 +350,72 @@ func allowProjectResourceList(d *Daemon, r *http.Request) response.Response {
 	return response.EmptySyncResponse
 }
 
+// reportEntitlements takes a map of entity URLs to EntitlementReporters (in practice, API types that implement the ReportEntitlements method), and
+// reports the entitlements that the caller has on each entity URL to the corresponding EntitlementReporter.
+func reportEntitlements(ctx context.Context, authorizer auth.Authorizer, identityCache *identity.Cache, entityType entity.Type, requestedEntitlements []auth.Entitlement, entityURLToEntitlementReporter map[*api.URL]auth.EntitlementReporter) error {
+	// Nothing to do
+	if len(entityURLToEntitlementReporter) == 0 {
+		return nil
+	}
+
+	id, err := auth.GetIdentityFromCtx(ctx, identityCache)
+	if err != nil {
+		return fmt.Errorf("Failed to get caller identity: %w", err)
+	}
+
+	if !identity.IsFineGrainedIdentityType(id.IdentityType) {
+		return fmt.Errorf("Not fine grained")
+	}
+
+	// In the case where we have only one entity URL, we'll use the authorizer's CheckPermission method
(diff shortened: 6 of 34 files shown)


lxd/instance_get.go
```diff
@@ -120,6 +121,12 @@ func instanceGet(d *Daemon, r *http.Request) response.Response {
 	// Parse the recursion field
 	recursive := util.IsRecursionRequest(r)
 
+	// Detect if we want to also returns entitlements for each instance.
```
minaelee, line 124:
```suggestion
	// Detect if we want to also return entitlements for each instance.
```

lxd/auth/types.go
```diff
@@ -52,6 +52,12 @@ type Authorizer interface {
 	// the entity. The effective project for the entity must be set in the request context as request.CtxEffectiveProjectName
 	// *before* the call to GetPermissionChecker.
 	GetPermissionChecker(ctx context.Context, entitlement Entitlement, entityType entity.Type) (PermissionChecker, error)
+
+	// GetEntitlements gets entitlements for an entity from a list of entitlement candidates.
+	GetEntitlements(ctx context.Context, entityType entity.Type, entityURL *api.URL, queriedEntitlementCandidates []Entitlement) ([]string, error)
+
+	// GetEntitlementsToEntities returns the entitlements for multiple entity URLs and types given a list of entitlement candidates.
```
minaelee, line 59:
```suggestion
	// GetEntitlementsToEntities returns the entitlements for multiple entity URLs and types, given a list of entitlement candidates.
```

lxd/instances_get.go
```diff
@@ -247,6 +247,12 @@ func instancesGet(d *Daemon, r *http.Request) response.Response {
 	projectName := request.QueryParam(r, "project")
 	allProjects := shared.IsTrue(r.FormValue("all-projects"))
 
+	// Detect if we want to also returns entitlements for each instance.
```
minaelee, line 250:
```suggestion
	// Detect if we want to also return entitlements for each instance.
```

gabrielmougard:
@minaelee thanks for the feedbacks. Fixing that right now

lxd/auth/types.go
gabrielmougard: fixed

lxd/instances_get.go
gabrielmougard: fixed

lxd/instance_get.go
gabrielmougard: fixed

shared/api/image.go
```diff
@@ -307,6 +309,8 @@ type ImageAliasesEntryPut struct {
 //
 // swagger:model
 type ImageAliasesEntry struct {
+	AccessInfo `yaml:",inline"`
+
```
edlerd, line 313: Not sure if this is needed. I think this is the reason the entitlements shows up also on the ImageAliasesPost entity in the `rest-api.yaml`, which seems a bit odd. Maybe not an issue and can stay as it is.

test/suites/auth.sh
edlerd, line 1: Shall we add a test case like the one below?

A restricted user has some entitlements on a project. Request the project list with that user and with a list of entitlements. Then assert the api response contains the correct list of entitlements for that project. Maybe request some entitlements the user has on the project, and some that the user has not to make it cover all the possible cases.

If we build this test case reusable, we can repeat it for all the different entity types, but not sure if that level of coverage is needed.

edlerd (commented):
Approach and format seems all right from my point of view. Left some comments below.

shared/api/image.go
tomponline: ah yeah thats a problem, we only want them in the GET response type

gabrielmougard: fixed

lxd/storage_volumes.go
```diff
@@ -623,6 +623,12 @@ func storagePoolVolumesGet(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	// Detect if we want to also returns entitlements for each volume.
```
minaelee, line 626:
```suggestion
	// Detect if we want to also return entitlements for each volume.
```

lxd/storage_volumes.go
```diff
@@ -2016,6 +2035,12 @@ func storagePoolVolumeGet(d *Daemon, r *http.Request) response.Response {
...
```
minaelee, line 2038:
```suggestion
	// Detect if we want to also return entitlements for each volume.
```

shared/api/auth.go
```diff
@@ -153,6 +155,8 @@ type IdentitiesTLSPost struct {
 //
 // API extension: access_management.
 type AuthGroup struct {
+	AccessInfo `yaml:",inline"`
```
tomponline, line 158: @edlerd @gabrielmougard @markylaing thinking about this further, shall we not inline this embedded struct, so that it becomes `foo.AccessInfo.Entitlements` that way we can avoid preventing adding an `Entitlements` field to any entity struct later, and also avoids any conflicts in the future should we want to add additional fields to AccessInfo.

What do you think?

lxd/daemon.go
```diff
@@ -350,6 +350,31 @@ func allowProjectResourceList(d *Daemon, r *http.Request) response.Response {
...
 
+// extractEntitlementsFromQuery extracts the entitlements from the query string of the request.
+func extractEntitlementsFromQuery(r *http.Request, entityType entity.Type) ([]auth.Entitlement, error) {
+	rawEntitlements := request.QueryParam(r, "with-entitlements")
+	if rawEntitlements == "" {
+		return nil, nil
+	}
+
+	allowedEntitlements := auth.EntityTypeToEntitlements[entityType]
+	if rawEntitlements == "all" {
```
tomponline, line 361: @gabrielmougard im not sure we should keep this, as we dont want to encourage "lazy" clients to just use this option all the time as it would have (potentially significant) performance impacts, which is why we chose not to do this by default originally.

lxd/daemon.go
```diff
@@ -350,6 +350,31 @@ func allowProjectResourceList(d *Daemon, r *http.Request) response.Response {
...
+		return allowedEntitlements, nil
+	}
+
+	entitlements := strings.Split(rawEntitlements, ",")
+	validEntitlements := make([]auth.Entitlement, len(entitlements))
```
tomponline, line 366: @gabrielmougard in general we prefer the `make([]auth.Entitlement, 0, len(entitlements))` form of this initialization, so that rather than creating a slice with empty indexes, we create an empty slice with an underlying pre-allocated capacity.

Then use `append` in the loop below rather than referencing by index.

This way we are not pre-initialising zeroed entites in the slice and then immediately replacing them in the loop.

Additionally, if this loop ever gets modified to include some skipping logic, there is no risk of ending up with empty indexes.

lxd/daemon.go
```diff
@@ -350,6 +350,31 @@ func allowProjectResourceList(d *Daemon, r *http.Request) response.Response {
...
+	for i, e := range entitlements {
+		if !shared.ValueInSlice(auth.Entitlement(e), allowedEntitlements) {
+			return nil, fmt.Errorf("entitlement %q part of the URL query is not compatible with the entity type %q", e, entityType)
```
tomponline, line 369: errors starting with capital letter please

lxd/auth/drivers/openfga.go
```diff
@@ -476,6 +476,77 @@ func (e *embeddedOpenFGA) GetPermissionChecker(ctx context.Context, entitlement
 	}, nil
 }
 
+// GetEntitlements gets entitlements to an entity using the embedded OpenFGA server.
+func (e *embeddedOpenFGA) GetEntitlements(ctx context.Context, entityType entity.Type, entityURL *api.URL, queriedEntitlementCandidates []auth.Entitlement) ([]string, error) {
```
tomponline, line 480: Please can you improve this doc block to cover what `queriedEntitlementCandidates` does (and can it be made a little less verbose too?), what happens if its empty/nil etc?

Also can this function return a `[]Entitlement` rather than `[]string`?

lxd/instances_get.go
```diff
@@ -506,15 +512,50 @@ func instancesGet(d *Daemon, r *http.Request) response.Response {
...
+	var entityURLs []*api.URL
+	var entitlementsMap map[string][]string
+
+	if len(withEntitlements) > 0 {
+		resultList = make([]*api.Instance, 0, len(resultFullList))
 		for i := range resultFullList {
 			resultList = append(resultList, &resultFullList[i].Instance)
 		}
 
+		entityURLs = make([]*api.URL, len(resultList))
```
tomponline, line 525: use capacity and append here

shared/api/auth.go
edlerd: I am fine with both options, inline or not-inline. `foo.AccessInfo.Entitlements` is trading the benefit of making it easier to add more access related info later on for the downside of a bit more complexity in the structure. Another benefit I can think of is that it becomes easier to understand. But the `entities_with_entitlements` value of api extension should probably change to indicate the presence of access info instead of entitlements. Also the `with-entitlements` should probably be changed to `with-access-info` or alike.

lxd/instance_get.go
```diff
@@ -120,6 +121,12 @@ func instanceGet(d *Daemon, r *http.Request) response.Response {
 	// Parse the recursion field
 	recursive := util.IsRecursionRequest(r)
 
+	// Detect if we want to also return entitlements for each instance.
+	withEntitlements, err := extractEntitlementsFromQuery(r, entity.TypeInstance)
```
tomponline, line 125: should we check if recursion param >0 and if not then return bad request error if with-entitlements is provided?

shared/api/auth.go
tomponline: Or we just add an `AccessEntitlements` field to each struct directly, rather than embedding it. By prefixing it with "Access" we clarify what it is used for. And then the API ext can be `with-access-entitlements`.

gabrielmougard: fixed

lxd/daemon.go
gabrielmougard: fixed

lxd/daemon.go
gabrielmougard: fixed

lxd/daemon.go
gabrielmougard: fixed

lxd/auth/drivers/openfga.go
gabrielmougard: fixed

lxd/auth/drivers/openfga.go
```diff
@@ -476,6 +476,77 @@ func (e *embeddedOpenFGA) GetPermissionChecker(ctx context.Context, entitlement
...
+			return nil, fmt.Errorf("Failed to add entitlements %q for entity URL %q: %w", entitlement, entityURL.String(), err)
+		}
+
+		entitlements = append(entitlements, string(entitlement))
+	}
+
+	return entitlements, nil
+}
+
+// GetEntitlementsToEntities adds entitlements to multiple entities using the embedded OpenFGA server.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

lxd/instance_get.go
gabrielmougard: yes. Fixed.

lxd/instances_get.go
gabrielmougard: fixed

lxd/storage_volumes.go
gabrielmougard: fixed

lxd/storage_volumes.go
gabrielmougard: fixed

test/suites/auth.sh
```diff
@@ -289,6 +288,34 @@ fine_grained: true"
...
   lxc config trust remove "${lxdconf6_fingerprint_short}"
 
+  # Create a new test project, add some entitlements on it and check that these are reflected in the 'access_entitlements' field returned from the API.
+  lxc project create test-project
+  lxc auth group permission add test-group project test-project can_view
+  lxc auth group permission add test-group project test-project can_edit
+  lxc auth group permission add test-group project test-project can_delete
+
+  # Check the created project entitlements given a list of candidate entitlements (some are wrong: `can_create_instances` and `can_create_networks`. These should not be returned).
+  [ "$(lxc_remote query "oidc:/1.0/projects/test-project?recursion=1&with-access-entitlements=can_view,can_edit,can_delete,can_create_instances,can_create_networks" | jq -c '.access_entitlements')" = '["can_view","can_edit","can_delete"]' ]
```
gabrielmougard, line 298: @markylaing The CI complains with this error:

```
lxc query 'oidc:/1.0/projects/test-project?recursion=1&with-access-entitlements=can_view,can_edit,can_delete,can_create_instances,can_create_networks' --verbose
time="2025-01-21T17:32:52Z" level=info msg="Access denied" driver=embedded-openfga entitlement=can_view entity_url=/1.0/projects/test-project http_code=404 protocol=oidc username=test-user@example.com
Error: Not Found
```

I don't quite understand why this is happening (in my local setup with keycloak, I manage to make that call though). Do you have an idea ?

markylaing: These tests need to come before this block: https://github.com/canonical/lxd/pull/14748/commits/3c7327b1fe154d34c7d53b076cda0deb4d59ff9b#diff-b27cd748067850b04a61be123a228ffca482cc8874deabf0df620a4b775ce00bR223-R232

The OIDC user has deleted themselves. They are authenticated but no longer a member of `test-group` and therefore do not have the permissions you are granting.

gabrielmougard: Ah of course... thanks a lot!

markylaing: Alternatively you can just add them to the group again before running these tests:
`lxc auth identity group add oidc/test-user@example.com test-group`

gabrielmougard: fixed

test/suites/auth.sh
gabrielmougard: fixed

gabrielmougard:
@markylaing @edlerd Can you review the tests please? They should be passing now

lxd/storage_pools.go
```diff
@@ -685,6 +718,20 @@ func storagePoolGet(d *Daemon, r *http.Request) response.Response {
...
+	if len(withEntitlements) > 0 {
+		entitlements, err := d.authorizer.GetEntitlements(r.Context(), entity.TypeStoragePool, entity.StoragePoolURL(poolName), withEntitlements)
+		if err != nil {
+			return response.SmartError(err)
+		}
+
+		accessEntitlements := make([]string, 0, len(entitlements))
+		for _, entitlement := range entitlements {
+			accessEntitlements = append(accessEntitlements, string(entitlement))
+		}
```
markylaing, line 730: Since these entitlements are used only for an API response, it's kind of annoying that they are returned as a `[]auth.Entitlement`. That being said, I still think it's the correct type for the interface.

Where did we get to with having each API type implement an interface such as:

```
// EntitlementAdder should be defined in the `lxd/auth` package and not in `shared/api`.
type EntitlementAdder interface {
    URL() *api.URL
    AddEntitlements([]string)
}
```

This could be an argument to the new methods on the Authorizer interface, so we don't need to add so much logic to the handlers.

test/suites/auth.sh
```diff
@@ -289,6 +288,37 @@ fine_grained: true"
...
+
+  # Storage volume
+  # Storage volume entitlements test
+  pool_name="$(lxc storage list -f csv | cut -d, -f1)"
+  lxc storage volume create "${pool_name}" test-volume
+  lxc auth group permission add test-group storage_volume test-volume can_view project=default pool="${pool_name}" type=custom
+  lxc auth group permission add test-group storage_volume test-volume can_edit project=default pool="${pool_name}" type=custom
+  lxc auth group permission add test-group storage_volume test-volume can_delete project=default pool="${pool_name}" type=custom
+  [ "$(lxc_remote query "oidc:/1.0/storage-pools/${pool_name}/volumes/custom/test-volume?project=default&recursion=1&with-access-entitlements=can_view,can_edit,can_delete,can_manage_backups,can_manage_snapshots" | jq -c '.access_entitlements')" = '["can_view","can_edit","can_delete"]' ]
+  lxc storage volume delete "${pool_name}" test-volume
```
markylaing, line 320: These look good. My only suggestion would be to use `sort` and `@csv` in `jq`. We aren't guaranteeing the order of the output and I think `@csv` is easier to perform an exact match with. @simondeziel might have another idea though :)
```
[ "$(lxc_remote query "oidc:/1.0/storage-pools/${pool_name}/volumes/custom/test-volume?project=default&recursion=1&with-access-entitlements=can_view,can_edit,can_delete,can_manage_backups,can_manage_snapshots" | jq -c '.access_entitlements | sort | @csv')" = '"can_delete","can_edit","can_view"' ]
```

markylaing (changes requested):
Tests look good :) Other than that there are a few issues and I'd like to revisit the idea to have the API types implement an interface so that we can just add the entitlements directly, rather than adding so much logic to the handlers.

Also, is this PR supposed to go over all entity types and add entitlements to them? If so I think we're missing a few. There is no issue having follow up PRs for those though.

doc/api-extensions.md
```diff
@@ -2552,3 +2552,10 @@ This adds support for listing network zones across all projects using the `all-p
 Adds support for instance root volumes to be attached to other instances as disk
 devices. Introduces the `<type>/<volume>` syntax for the `source` property of
 disk devices.
+
+## `entities_with_entitlements`
+
+Adds `is_fine_grained` field to `GET /1.0/auth/identities/current` to indicate if the current identity
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/api-extensions.md
```diff
@@ -2552,3 +2552,10 @@ This adds support for listing network zones across all projects using the `all-p
...
+interacting with the LXD API is fine-grained (that is, associated permissions are managed via group membership).
+Allows LXD entities to be returned with an `entitlements` field if the current identity is fine-grained and the
+GET request to fetch the LXD entities has the `with-entitlements=<comma_separated_list_of_candidate_entitlements>` query parameter.
```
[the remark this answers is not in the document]
gabrielmougard: fixed

doc/api-extensions.md
```diff
@@ -2552,3 +2552,10 @@ This adds support for listing network zones across all projects using the `all-p
 Adds support for instance root volumes to be attached to other instances as disk
 devices. Introduces the `<type>/<volume>` syntax for the `source` property of
 disk devices.
+
+## `entities_with_entitlements`
+
+Adds `is_fine_grained` field to `GET /1.0/auth/identities/current` to indicate if the current identity
+interacting with the LXD API is fine-grained (that is, associated permissions are managed via group membership).
+Allows LXD entities to be returned with an `entitlements` field if the current identity is fine-grained and the
```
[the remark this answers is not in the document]
gabrielmougard: fixed

lxd/auth/drivers/openfga.go
```diff
@@ -476,6 +476,83 @@ func (e *embeddedOpenFGA) GetPermissionChecker(ctx context.Context, entitlement
...
+
+	if !identity.IsFineGrainedIdentityType(id.IdentityType) {
+		return e.tlsAuthorizer.GetEntitlements(ctx, entityType, entityURL, entitlementCandidates)
+	}
+
+	entitlements := make([]auth.Entitlement, 0)
+	for _, entitlement := range entitlementCandidates {
+		err = e.CheckPermission(ctx, entityURL, entitlement)
+		if err != nil {
+			if err == api.NewGenericStatusError(http.StatusNotFound) || err == api.NewGenericStatusError(http.StatusForbidden) {
```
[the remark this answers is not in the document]
gabrielmougard: fixed

lxd/auth/drivers/openfga.go
```diff
@@ -476,6 +476,83 @@ func (e *embeddedOpenFGA) GetPermissionChecker(ctx context.Context, entitlement
...
+	}
+
+	return entitlements, nil
+}
+
+// GetEntitlementsToEntities gets entitlements for multiple entities using the embedded OpenFGA server.
+// Again, `entitlementCandidates` is a list of entitlements that the caller is interested in checking.
+// The returned map will contain the entity URL as the key and a list of entitlements as the value.
+// The list of entitlements will be a subset of the input entitlement candidates list.
+func (e *embeddedOpenFGA) GetEntitlementsToEntities(ctx context.Context, entityType entity.Type, entityURLs []*api.URL, entitlementCandidates []auth.Entitlement) (map[string][]auth.Entitlement, error) {
```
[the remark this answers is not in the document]
gabrielmougard: fixed

lxd/api_project.go
```diff
@@ -143,6 +143,10 @@ func projectsGet(d *Daemon, r *http.Request) response.Response {
 	s := d.State()
 
 	recursion := util.IsRecursionRequest(r)
+	withEntitlements, err := extractEntitlementsFromQuery(r, entity.TypeProject)
+	if err != nil {
+		return response.SmartError(err)
+	}
```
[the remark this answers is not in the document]
gabrielmougard: fixed

lxd/storage_pools.go
gabrielmougard: @markylaing That is fine to implement `AddEntitlements` for the various api entities but the `URL()` method takes arguments and they are not the same from one type to an other... Could we instead have the method `URL(parts ...string) *api.URL` and adapt the existing api type `URL` method to match that signature ?

lxd/instances_get.go
```diff
@@ -510,6 +520,32 @@ func instancesGet(d *Daemon, r *http.Request) response.Response {
...
+	if len(withEntitlements) > 0 {
+		instances := make([]*api.Instance, 0, len(resultFullList))
+		for _, inst := range resultFullList {
+			instances = append(instances, &inst.Instance)
+		}
+
+		entityURLs := make([]*api.URL, 0, len(instances))
+		for _, inst := range instances {
+			entityURLs = append(entityURLs, entity.InstanceURL(inst.Project, inst.Name))
+		}
```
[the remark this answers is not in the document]
gabrielmougard: fixed

lxd/storage_pools.go
gabrielmougard: fixed

lxd/instances_get.go
```diff
@@ -510,6 +520,32 @@ func instancesGet(d *Daemon, r *http.Request) response.Response {
 		return response.SyncResponse(true, resultList)
 	}
 
+	if len(withEntitlements) > 0 {
+		instances := make([]*api.Instance, 0, len(resultFullList))
+		for _, inst := range resultFullList {
+			instances = append(instances, &inst.Instance)
+		}
```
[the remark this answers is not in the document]
gabrielmougard: fixed

gabrielmougard:
@markylaing updated following the approach you suggested this morning. I also added the enrichment for all the other entities.

lxd/auth_groups.go
```diff
@@ -181,6 +181,15 @@ func getAuthGroups(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(fmt.Errorf("Failed to get a permission checker: %w", err))
 	}
 
+	withEntitlements, err := extractEntitlementsFromQuery(r, entity.TypeAuthGroup)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	if len(withEntitlements) > 0 && recursion != "1" {
```
edlerd, line 189:
```suggestion
	if len(withEntitlements) > 0 && !recursion {
```
Should be more future-proof this way in case higher recursion levels are introduced. wdyt?

lxd/identities.go
```diff
@@ -797,6 +797,15 @@ func getIdentities(authenticationMethod string) func(d *Daemon, r *http.Request)
 			return response.SmartError(err)
 		}
 
+		withEntitlements, err := extractEntitlementsFromQuery(r, entity.TypeIdentity)
+		if err != nil {
+			return response.SmartError(err)
+		}
+
+		if len(withEntitlements) > 0 && recursion != "1" {
```
edlerd, line 805: Similar to above, there are a couple of more occurrences like this below.

edlerd (approved):
New tests are great, thanks for adding. I skimmed through the code without detailed reading, just one non-blocking observation below. LGTM :+1:

minaelee (dismissed):

tomponline (changes requested):
@gabrielmougard please add the `// API extension: ` comment line to the new `AccessEntitlments` fields like we've done for earlier API extensions that add new fields.

tomponline (commented):
Can the `ReportEntitlements` function and the `AccessEntitlements` field not be added to each APIs struct by way of an embedded type?

lxd/identities.go
gabrielmougard: fixed

lxd/auth_groups.go
gabrielmougard: fixed

edlerd (approved):
Thanks for adjusting, LGTM

gabrielmougard:
> Can the `ReportEntitlements` function and the `AccessEntitlements` field not be added to each APIs struct by way of an embedded type?

@tomponline should be updated now

lxd/instance_get.go
```diff
@@ -120,6 +122,16 @@ func instanceGet(d *Daemon, r *http.Request) response.Response {
...
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	if len(withEntitlements) > 0 && !recursive {
```
tomponline, line 131: Can we add this check inside `extractEntitlementsFromQuery` to avoid duplication?

lxd/instances_get.go
```diff
@@ -250,6 +250,16 @@ func instancesGet(d *Daemon, r *http.Request) response.Response {
 	projectName := request.QueryParam(r, "project")
 	allProjects := shared.IsTrue(r.FormValue("all-projects"))
 
+	// Detect if we want to also return entitlements for each instance.
+	withEntitlements, err := extractEntitlementsFromQuery(r, entity.TypeInstance)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	if len(withEntitlements) > 0 && !mustLoadObjects {
```
tomponline, line 259: why are we not checking `recursion` here? As `mustLoadObjects` is true even when `recursion` is `0` but when filtering clauses are requested. And we dont need to do entitlement enrichment when `recursion` is false. Also if we can do this then we can use the same logic we are moving into `extractEntitlementsFromQuery` for the other endpoints.

lxd/api_project.go
```diff
@@ -417,13 +443,19 @@ func projectGet(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	project.UsedBy = projecthelpers.FilterUsedBy(s.Authorizer, r, project.UsedBy)
```
tomponline, line 446: wasnt expecting to see this line being moved, can it be undone please

lxd/storage_buckets.go
```diff
@@ -239,7 +248,7 @@ func storagePoolBucketsGet(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
-	var filteredDBBuckets []*db.StorageBucket
+	filteredDBBuckets := make([]*db.StorageBucket, 0)
```
tomponline, line 251: separate commit please

lxd/storage_buckets.go
```diff
@@ -358,6 +385,17 @@ func storagePoolBucketGet(d *Daemon, r *http.Request) response.Response {
...
 
+	if util.IsRecursionRequest(r) {
+		if len(withEntitlements) > 0 {
+			err = reportEntitlements(r.Context(), s.Authorizer, s.IdentityCache, entity.TypeStorageBucket, withEntitlements, map[*api.URL]auth.EntitlementReporter{entity.StorageBucketURL(effectiveProjectName, bucket.Location, details.pool.Name(), bucket.Name): &bucket.StorageBucket})
+			if err != nil {
+				return response.SmartError(err)
+			}
+		}
+
+		return response.SyncResponse(true, &bucket.StorageBucket)
```
tomponline, line 396: this doesnt look like it should be there as we have a (correct) SyncReponseETag call below, and here we have no ETag generated which would be a regression.

lxd/certificates.go
```diff
@@ -143,9 +143,19 @@ func certificatesGet(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	withEntitlements, err := extractEntitlementsFromQuery(r, entity.TypeCertificate)
```
tomponline, line 146: suggest we add an `allowRecursion bool` argument to this function to allow us to remove this repeated stanza:

```go

	if len(withEntitlements) > 0 && !recursion {
		return response.BadRequest(fmt.Errorf("Entitlements can only be requested when recursion is enabled"))
	}
```

lxd/images.go
```diff
@@ -1818,6 +1828,24 @@ func imagesGet(d *Daemon, r *http.Request) response.Response {
...
 	}
 
+	if len(withEntitlements) > 0 {
+		// 'result' must be a slice of Image pointers.
+		images, ok := result.([]*api.Image)
+		if !ok {
+			return response.InternalError(fmt.Errorf("Images response is not a slice of Image pointer structs"))
+		}
+
+		urlToImage := make(map[*api.URL]auth.EntitlementReporter)
```
tomponline, line 1838: I believe we can pre-allocate the map by doing `make(map[*api.URL]auth.EntitlementReporter, len(images))` we should do this here and in the other API endpoints.

lxd/profiles.go
```diff
@@ -212,6 +212,14 @@ func profilesGet(d *Daemon, r *http.Request) response.Response {
 	}
 
 	recursion := util.IsRecursionRequest(r)
+	withEntitlements, err := extractEntitlementsFromQuery(r, entity.TypeProfile)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	if len(withEntitlements) > 0 && !recursion {
+		return response.BadRequest(errors.New("Cannot specify a project when requesting all projects"))
```
tomponline, line 221: good job we are moving this into `extractEntitlementsFromQuery` as this error is wrong.

test/suites/auth.sh
tomponline: @gabrielmougard to address please

gabrielmougard: fixed

lxd/instance_get.go
gabrielmougard: fixed

lxd/api_project.go
gabrielmougard: fixed

lxd/certificates.go
gabrielmougard: fixed

lxd/images.go
```diff
@@ -1818,6 +1828,24 @@ func imagesGet(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	if len(withEntitlements) > 0 {
+		// 'result' must be a slice of Image pointers.
```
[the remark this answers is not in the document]
gabrielmougard: Actually no because I need the pointer's `Project` and `Fingerprint` fields to build the URL map down below. I'll add a comment to explain that.

lxd/images.go
gabrielmougard: fixed

lxd/profiles.go
gabrielmougard: fixed

lxd/instances_get.go
gabrielmougard: fixed

lxd/storage_buckets.go
gabrielmougard: fixed

tomponline (approved):
LGTM thanks!

gabrielmougard:
@tomponline auth test passed. Let's wait for the rest.

tomponline (approved):

tomponline (approved):

PR #14749, opened 2025-01-06
CLI: Add `lxc file` completions and fix `lxc profile copy` completions (from Incus)

kadinsayani:
This PR improves completions for `lxc file` and fixes an issue with `lxc profile copy` completions.

Summary of changes:
- Improves remote and instance completions by adding a comparison of possible completions to the completion string.
- Adds `cmpFiles` for dynamic completions for `lxc file`.
- Fixes an issue with `lxc profile copy` completions by making use of the `cmpProfiles` function.


lxc/file.go
```diff
@@ -824,7 +846,7 @@ func (c *cmdFilePush) run(cmd *cobra.Command, args []string) error {
 	defer reverter.Fail()
 
 	// Make sure all of the files are accessible by us before trying to push any of them
-	var files []*os.File
+	files := make([]*os.File, len(sourcefilenames))
```
simondeziel, line 849:
```suggestion
	files := make([]*os.File, 0, len(sourcefilenames))
```
Since we are appending to it, better have the initial length set to 0, I think.

lxc/image.go
```diff
@@ -1394,6 +1394,7 @@ func (c *cmdImageList) run(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	images := make([]api.Image, len(allImages))
```
simondeziel, line 1397:
```suggestion
	images := make([]api.Image, 0, len(allImages))
```

Same rational re length.

lxc/completion.go
```diff
@@ -1635,3 +1650,89 @@ func (g *cmdGlobal) cmpStoragePoolVolumes(poolName string, volumeTypes ...string
...
+	sep := string(filepath.Separator)
+	dir, prefix := filepath.Split(toComplete)
+	switch prefix {
+	case ".":
+		files = append(files, dir+"."+sep)
+		fallthrough
+	case "..":
+		files = append(files, dir+".."+sep)
+		directives |= cobra.ShellCompDirectiveNoSpace
+	}
```
simondeziel, line 1701: I'm puzzled by the check for `.` first then falling through. This seems backward to me but I don't fully understand the intent it seems. Also, why is `cobra.ShellCompDirectiveNoSpace` only wanted when the `..` prefix is found?

A few comments might help me understand the code :)

kadinsayani: The intent is to ensure both "./" and "../" are included if the user types "." as the prefix. The `fallthrough` ensures both cases are handled and preventing a space when a user types `../` allows them to continue typing a sub directory.

I've added some comments as well :)

kadinsayani: The same effect could be achieved by including a check for `.` and `..` in separate `if` clauses:

```Go
if prefix == "." {
    files = append(files, dir+"."+sep)
}

if prefix == ".." {
    files = append(files, dir+".."+sep)
    directives |= cobra.ShellCompDirectiveNoSpace
}
```

simondeziel: If the user types `.`, wouldn't it `append` twice to the `files` due to the `fallthrough`?

Is this snippet an accurate translation of your previous explanation?

```
if (prefix == "." || prefix == "..") {
  	files = append(files, dir+prefix+sep)
  	directives |= cobra.ShellCompDirectiveNoSpace
}
```

simondeziel: Your `if`-based alternative doesn't seem equivalent to the switch case with fallthrough but maybe I'm not just fully back from vacation ;)

kadinsayani: If the user types `.`,  it would append `dir/./` and `dir/../`. You're right, my original `if` based alternative is missing the cascading logic achieved by the`fallthrough` - a more accurate alternative could be:

```Go
if prefix == "." {
    // Add "./"
    files = append(files, dir+"."+sep)

    // Also add "../" as a fallthrough effect
    files = append(files, dir+".."+sep)
    directives |= cobra.ShellCompDirectiveNoSpace
} else if prefix == ".." {
    // Add "../" only
    files = append(files, dir+".."+sep)
    directives |= cobra.ShellCompDirectiveNoSpace
}
```

lxc/completion.go
```diff
@@ -1635,3 +1650,99 @@ func (g *cmdGlobal) cmpStoragePoolVolumes(poolName string, volumeTypes ...string
...
+func (g *cmdGlobal) cmpFiles(toComplete string, includeLocalFiles bool) ([]string, cobra.ShellCompDirective) {
+	// Get instance completions first.
+	instances, directives := g.cmpInstances(toComplete)
+	// Append "/" to instances to indicate directory-like behavior.
+	for i := range instances {
+		if strings.HasSuffix(instances[i], ":") {
+			continue
+		}
+
+		instances[i] += "/"
```
MusicDin, line 1676: Should we use filepath.Separator for consistency?

lxc/completion.go
```diff
@@ -1635,3 +1650,99 @@ func (g *cmdGlobal) cmpStoragePoolVolumes(poolName string, volumeTypes ...string
...
+	switch prefix {
+	case ".":
+		// Add "./" for the current directory and fall through to handle "..".
+		files = append(files, dir+"."+sep)
+		fallthrough
+	case "..":
+		// Add "../" for the parent directory and ensure no space is added after "..".
+		// We want to handle cases where users continue typing (e.g. "../subdir").
+		files = append(files, dir+".."+sep)
+		directives |= cobra.ShellCompDirectiveNoSpace
```
MusicDin, line 1707: `directives |= cobra.ShellCompDirectiveNoSpace`

This was already set before the switch statement.

lxc/completion.go
```diff
@@ -1635,3 +1650,99 @@ func (g *cmdGlobal) cmpStoragePoolVolumes(poolName string, volumeTypes ...string
...
+		}
+
+		base := filepath.Base(path)
+		if strings.HasPrefix(base, prefix) {
+			// Match files and directories starting with the given prefix.
+			file := dir + base
+			switch {
+			case d.IsDir():
+				// Add "/" for directories and set no space directive.
+				directives |= cobra.ShellCompDirectiveNoSpace
```
MusicDin, line 1727: Again, shouldn't `directives |= cobra.ShellCompDirectiveNoSpace` be already set at this point?

MusicDin (commented):
Just some minor comments/questions.

lxc/completion.go
tomponline: this depends on whether its for the local machine (that may be Windows) or remote which will be Linux

MusicDin: Just to clarify, this separator is just for LXD to indicate file on the instance?

kadinsayani: Yes, the separator is used to indicate files and directory-like behavior on the instance.

lxc/completion.go
kadinsayani: Yes, I've cleaned up the setting of shell completion directives in `cmpFiles`.

simondeziel (approved):

PR #14760, opened 2025-01-08
devlxd: Unmask Render errors

hamistao:
This implements a better handling of errors during Render on devlxd handlers, following the same pattern seen on https://github.com/hamistao/lxd/blob/unmask_write_errors/lxd/daemon.go#L846
When rendering a response (send it to the client) fails, instead of masking the error we try to send a response using `DevLxdErrorResponse` to the client communicating the error, since this determines the response type used for communicating erros on the devlxd API. If that also fails, an error is logged on the server.

### DIFF
--- a/lxd/api.go
+++ b/lxd/api.go
@@ -243,7 +243,15 @@ func hoistReqVM(f func(*Daemon, instance.Instance, http.ResponseWriter, *http.Re
 		}
 
 		resp := f(d, inst, w, r)
-		_ = resp.Render(w, r)
+		if resp != nil {
+			err = resp.Render(w, r)
+			if err != nil {
+				writeErr := response.DevLxdErrorResponse(err, true).Render(w, r)
+				if writeErr != nil {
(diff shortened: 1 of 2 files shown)


lxd/api.go
```diff
@@ -243,7 +243,15 @@ func hoistReqVM(f func(*Daemon, instance.Instance, http.ResponseWriter, *http.Re
 		}
 
 		resp := f(d, inst, w, r)
-		_ = resp.Render(w, r)
+		if resp != nil {
```
tomponline, line 246: is resp sometimes nil?

tomponline:
> This implements a better handling of errors during Render on devlxd handlers, following the same pattern seen on https://github.com/hamistao/lxd/blob/unmask_write_errors/lxd/daemon.go#L846

Please can you expand the PR description to explain how the API behaviour has changed, as it seems like error response formats have changed?

lxd/api.go
hamistao: I don't think they should ever be nil, just thought it was more coherent to not assume anything about the handler's behavior

tomponline (approved):

PR #14761, opened 2025-01-08
Storage: Add block reset logic

tomponline:


lxd/storage/drivers/driver_lvm_utils.go
```diff
@@ -572,27 +610,27 @@ func (d *lvm) copyThinpoolVolume(vol, srcVol Volume, srcSnapshots []string, refr
...
-				return fmt.Errorf("Error temporarily renaming original LVM logical volume: %w", err)
-			}
+		if !refresh {
+			return fmt.Errorf("LVM volume already exists %q", vol.name)
+		}
 
-			// Record this volume to be removed at the very end.
-			removeVols = append(removeVols, tmpVolName)
+		newVolDevPath := d.lvmDevPath(d.config["lvm.vg_name"], vol.volType, vol.contentType, vol.name)
+		tmpVolName := fmt.Sprintf("%s%s", vol.name, tmpVolSuffix)
```
simondeziel, line 618:
```suggestion
		tmpVolName := vol.name + tmpVolSuffix
```

lxd/storage/block/discard.go
```diff
@@ -0,0 +1,263 @@
...
+
+	if found != 3 {
+		return fmt.Errorf("Some of our initial markers weren't written properly")
+	}
+
+	// Start clearing the block.
+	_ = fd.Close()
+
+	// Attempt a secure discard run.
+	_, err = shared.RunCommand("blkdiscard", "--force", "--offset", fmt.Sprintf("%d", blockOffset), "--secure", blockPath)
```
simondeziel, line 159:
```suggestion
	_, err = shared.RunCommand("blkdiscard", "--force", "--offset", strconv.Itoa(blockOffset), "--secure", blockPath)
```

lxd/storage/block/discard.go
```diff
@@ -0,0 +1,263 @@
...
+			// All markers are gone, secure discard succeeded.
+			return nil
+		}
+
+		// Some markers were found, go to the next clearing option.
+		_ = fd.Close()
+	}
+
+	// Attempt a regular discard run.
+	_, err = shared.RunCommand("blkdiscard", "--force", "--offset", fmt.Sprintf("%d", blockOffset), blockPath)
```
simondeziel, line 184: `strconv.Itoa()` here as well.

lxd/storage/block/discard.go
```diff
@@ -0,0 +1,263 @@
...
+			// All markers are gone, regular discard succeeded.
+			return nil
+		}
+
+		// Some markers were found, go to the next clearing option.
+		_ = fd.Close()
+	}
+
+	// Attempt device zero-ing.
+	_, err = shared.RunCommand("blkdiscard", "--force", "--offset", fmt.Sprintf("%d", blockOffset), "--zeroout", blockPath)
```
simondeziel, line 209: Same here.

lxd/storage/block/discard.go
tomponline: strconv.FormatInt(blockOffset, 10)

lxd/storage/block/discard.go
tomponline: strconv.FormatInt(blockOffset, 10)

lxd/storage/block/discard.go
tomponline: strconv.FormatInt(blockOffset, 10)

lxd/storage/block/discard.go
tomponline: yep, just that one takes an int64

lxd/storage/drivers/driver_lvm_utils.go
```diff
@@ -501,11 +508,42 @@ func (d *lvm) lvmDevPath(vgName string, volType VolumeType, contentType ContentT
...
+	_, err = shared.TryRunCommand("lvresize", args...)
 	if err != nil {
 		return err
 	}
 
+	// On thick pools, discard the blocks in the additional space when the volume is grown.
+	if needsClearing {
+		if oldSizeBytes < sizeBytes {
+			// Discard the new blocks.
+			err := block.ClearBlock(lvPath, sizeBytes)
```
simondeziel, line 540: At that point, the LV has already been grown so shouldn't we call `ClearBlock` with `oldSizeBytes` so that it zero's from `oldSizeBytes` up to `sizeBytes`?

lxd/storage/block/discard.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+
+		if found == 0 {
+			// All markers are gone, secure discard succeeded.
+			return nil
+		}
+
+		// Some markers were found, go to the next clearing option.
+		_ = fd.Close()
+	}
```
simondeziel, line 182: That boilerplate code around `checkMarkers` is repeated 3 times. Any way to avoid the repetition?

lxd/storage/block/discard.go
```diff
@@ -0,0 +1,264 @@
...
+			return nil
+		}
+
+		return err
+	}
+
+	defer fd.Close()
+
+	// Get the size of the file/block.
+	size, err := fd.Seek(0, io.SeekEnd)
```
simondeziel, line 35: Should it error out if the retrieved `size` is identical to the provided `blockOffset`?

lxd/storage/drivers/driver_lvm_utils.go
tomponline: yes i think you could be right here, will take a look thanks

tomponline: thanks, ive reworked the thick volume grown code as it had additional problems.

It didn't work basically, because the lvm volume wasn't activated when ClearBlock was called and the ClearBlock function silently ignored that. I strongly like that kind of "magic" behaviour, so instead of having ClearBlock ignore missing files, its now up to the caller to decide if that is a valid error to ignore or not.

tomponline: Have now confirmed that creating a block volume at 10MiB, filling it with "A"s, counting them and then growing the volume, checking that it contained the same amount of As as before and no other data.

lxd/storage/block/discard.go
tomponline: i tried a couple of ways of removing this repetition but it didnt result in much de-dupe in the end, and in one case it make the code less efficient by requiring another file handle. So I'm OK to leave as is.

lxd/storage/block/discard.go
tomponline: im not sure. Will the function run OK, if so i dont see an issue with this? Could you expand on your concern here? Thanks

simondeziel: If both sizes are identical, it means no clearing will actually happen so the caller likely fed the wrong offset. Same as what I flagged with the fat LVM case where the LV was already resized when block clearing was requested for the new size (instead of the old size).

lxd/storage/drivers/driver_lvm_volumes.go
```diff
@@ -548,10 +564,17 @@ func (d *lvm) SetVolumeQuota(vol Volume, size string, allowUnsafeResize bool, op
 			return err
 		}
 
-		// Move the VM GPT alt header to end of disk if needed (not needed in unsafe resize mode as it is
-		// expected the caller will do all necessary post resize actions themselves).
-		if vol.IsVMBlock() && !allowUnsafeResize {
-			// Activate the volume for resizing.
+		// The new blocks in a grown volume will need clearing if using a thick pool.
+		needsClearing := !d.usesThinpool() && (oldSizeBytes < sizeBytes)
```
simondeziel, line 568: So this deals with zeroing the added bytes to a fat LV being grown. What about the shrinking case? I don't know if LVM thin does discard the freed block when shrinking but it's something we should be doing to avoid leaving customer data on disk.

tomponline: This issue is about not exposing old data to newly created volumes, rather than securely deleting data off of the disk when the volume is deleted/shrunk, so I think this doesn't need to be addressed in this PR.

lxd/storage/block/discard.go
tomponline: right i see, ill update to error on that scenario thanks

tomponline: have updated

simondeziel (approved):

PR #14765, opened 2025-01-09
doc: update ref to cloud-init docs

minaelee:
Recent update in cloud-init docs https://github.com/canonical/cloud-init/pull/5909 included changing an anchor #vendordata to #vendor-data in one of their pages, which caused `WARNING: undefined label: 'cloud-init:vendordata' [ref.ref]` error when building our docs. This PR simply updates that reference to `cloud-init:vendor-data`.

### DIFF
--- a/doc/cloud-init.md
+++ b/doc/cloud-init.md
@@ -47,7 +47,7 @@ However, there might be exceptions to that rule.
 
 The following configuration options are supported:
 
-* `cloud-init.vendor-data` or `user.vendor-data` (see {ref}`cloud-init:vendordata`)
+* `cloud-init.vendor-data` or `user.vendor-data` (see {ref}`cloud-init:vendor-data`)
 * `cloud-init.user-data` or `user.user-data` (see {ref}`cloud-init:user_data_formats`)
 * `cloud-init.network-config` or `user.network-config` (see {ref}`cloud-init:network_config`)
 


MusicDin (approved):
Thanks 🚀

tomponline (approved):
Thanks!

BTW how do we know that the ref to `cloud-init` references an external documentation set?

minaelee:
> Thanks!
> 
> BTW how do we know that the ref to `cloud-init` references an external documentation set?

Anytime you see the pattern ``{ref}`foo:bar` ``, that tells you the reference is to an external RTD docs set via the [Intersphinx extension](https://docs.readthedocs.io/en/stable/guides/intersphinx.html). In this case, the reference was to `cloud-init:vendor-data`.

Then you can find the external docs set URL in `docs/custom_conf.py` [here](https://github.com/canonical/lxd/blob/c0d7e8f22bbcf5972749c6b578ff8877aff9f707/doc/custom_conf.py#L283). In this case, that looks like: `'cloud-init': ('https://cloudinit.readthedocs.io/en/latest/', None),`

The LXD documentation doesn't use this all that much, but MicroCloud's depends on it heavily for linking to the docs for LXD, Ceph, MicroCeph, MicroOVN.

tomponline:
excellent thanks!

Issue #14706, opened 2024-12-20
LXD 5.21.2 crashed 


mastier:
# Required information

 * Distribution: Ubuntu
 * Distribution version: 22.04
 * The output of "snap list --all lxd core20 core22 core24 snaps":
```
$ snap list --all lxd core20 core22 core24 snaps
Name    Version         Rev    Tracking       Publisher   Notes
core20  20240911        2434   latest/stable  canonical✓  base
core22  20241119        1722   latest/stable  canonical✓  base
core24  20240920        609    latest/stable  canonical✓  base
lxd     5.0.4-497fe1e   31333  5.21/stable    canonical✓  disabled
lxd     5.21.2-084c8c8  31214  5.21/stable    canonical✓  -
```
 * The output of "lxc info" or if that fails:
   * Kernel version: 6.8.0-49-generic
   * LXC version: 5.21.2 LTS
   * LXD version: 5.21.2 LTS
   * Storage backend in use: dir

# Issue description

I was deploying Openstack cloud with Juju. Some containers on one of the nodes failed to get the IP address (4 of them).
After restarting containers the could get the address normally.
Around the time of the deployment I saw that LXD crashed in the logs. So that might be related
```
Dec 19 17:54:12 controller-2 lxd.daemon[17490]: => First LXD execution on this system
Dec 19 17:54:12 controller-2 lxd.daemon[17490]: => LXD is ready
Dec 19 17:56:33 controller-2 lxd.daemon[17656]: fatal error: concurrent map read and map write
Dec 19 17:56:33 controller-2 lxd.daemon[17656]: goroutine 2928 [running]:
Dec 19 17:56:33 controller-2 lxd.daemon[17656]: main.hoistReq.func1({0x265fe88, 0xc000279340}, 0xc00082f900)
Dec 19 17:56:33 controller-2 lxd.daemon[17656]:         github.com/canonical/lxd/lxd/devlxd.go:277 +0xf0
Dec 19 17:56:33 controller-2 lxd.daemon[17656]: net/http.HandlerFunc.ServeHTTP(0xc00082f540?, {0x265fe88?, 0xc000279340?}, 0x538389?)
Dec 19 17:56:33 controller-2 lxd.daemon[17656]:         net/http/server.go:2220 +0x29
[...]

PR #14771, opened 2025-01-13
Devlxd: Fix crash due to concurrent pidMapper map access

tomponline:
Fixes https://github.com/canonical/lxd/issues/14706


MusicDin (approved):
Thanks 🚀

PR #14792, opened 2025-01-13
lxd-agent: If the target path of a mount is relative, make it absolute.

### LINKED ISSUE #13406: Hot plugging filesystem volume into VM with relative `path` causes it to be mounted under `/run/lxd_agent/`
When hot-plugging a filesystem volume into a VM the behaviour of where that volume is mounted inside the guest when using a relative `path` setting differs from the behaviour of hot-plugging a filesystem volume into a container.

```
lxc storage volume create default foo --type=filesystem
lxc launch ubuntu:24.04 c1
lxc launch ubuntu:24.04 v1 --vm

# Hot-plugging it into a container with path=mnt causes it to be mounted at /mnt as expected.
lxc config device add c1 d0 disk source=foo pool=default path=mnt
lxc exec c1 -- mount | grep mnt
/dev/mapper/vgubuntu-root on /mnt type ext4 (rw,relatime,errors=remount-ro)

# Hot-plugging it into a VM with path=mnt causes it to be mounted at /run/lxd_agent/mnt (this is wrong, it should be /mnt).
lxc config device add v1 d0 disk source=foo pool=default path=mnt
lxc exec v1 -- mount | grep mnt
lxd_d0 on /run/lxd_agent/mnt type virtiofs (rw,relatime)
```

Note: This does not occur when the VM instance has the disk added before it is started, it only occurs when hot-plugging the disk.

We should ensure the agent uses the same behaviour during hot-plugging as it does at boot time.

gabrielmougard:
closes #13406

If the path is not absolute, the mount will be created at `/run/lxd_agent/<path>`, since the mount command executed below originates from the `lxd-agent` binary that is in the `/run/lxd_agent` directory. This is not ideal and not consistent with the way mounts are handled with containers. We then make the path absolute.

### DIFF
--- a/lxd-agent/events.go
+++ b/lxd-agent/events.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


tomponline (approved):

tomponline:
ta

PR #14794, opened 2025-01-14
Storage: Revert "lxd/storage/drivers/generic/vfs: Truncate/Discard ahead of sparse write"

tomponline:
This reverts commit ab910ac2cfbe2c91d6b5bcfe303dfe350b79ebe9 from https://github.com/canonical/lxd/pull/14761.

It is causing failures with our LVM CI tests due to the additional time it takes to do a slow `blkdiscard --zeroout` run when resetting the volume after an initial snapshot is transferred.

As the other commit in this PR deals with resetting the newly created volume, no old data from previously deleted volumes will be present, and this commit is just designed to ensure that "holes" of zeroes in the incoming data stream don't corrupt the volume when using the sparse writer.

As we've not landed the sparse writer yet, all of the bytes from the incoming data stream are written to the volume and thus there is no issue.

So we can hold off from landing this commit until we land the sparse writer.

### DIFF
--- a/lxd/storage/drivers/generic_vfs.go
+++ b/lxd/storage/drivers/generic_vfs.go
@@ -2,10 +2,8 @@ package drivers
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
-	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


hamistao (approved):
Thanks for the clear commit message :))
I was able to understand a process outside my comfort zone in a single read.

PR #14797, opened 2025-01-14
lxd-agent/metrics: ignore fuse.lxcfs FSes

simondeziel:


simondeziel:
This was brought to our attention by the observability team:
> Hey LXD team, we're seeing Prometheus metrics in node exporter about used_disk_space, with fstype=fuse.lxcfs, with NaN as a value; is that something we can safely ignore?

tomponline (approved):

PR #14798, opened 2025-01-14
doc: Add file command as requirement for lxd-migrate

MusicDin:
Add missing command `file` requirement for `lxd-migrate` in the docs.


tomponline (approved):
Thanks!

tomponline (approved):

Issue #14155, opened 2024-09-25
Copy custom storage volume fails if target cluster member and target project are set at the same time


mas-who:
# Issue description

When trying to copy a custom storage volume using the CLI, an error message "Project and target project are the same" is returned when the target cluster member and project are both different to those of the source volume.  See below an example for the lxc command that would cause the error.
```
lxc storage volume copy local/test-local-volume local/test-local-volume-duplicate --target node-1 --destination-target node-2 --target-project test-project
```
When calling the API endpoint `POST /1.0/storage-pools/{poolName}/volumes/{type}` directly with `target` and `project` query params both set, this results in the same error response with a 400 status code.

Below is the output of `snap list`
```
Name                       Version                 Rev    Tracking         Publisher         Notes
bare                       1.0                     5      latest/stable    canonical✓        base
code                       38c31bc7                169    latest/stable    vscode✓           classic
core                       16-2.61.4-20240607      17200  latest/stable    canonical✓        core
core18                     20240612                2829   latest/stable    canonical✓        base
core20                     20240705                2379   latest/stable    canonical✓        base
core22                     20240904                1621   latest/stable    canonical✓        base
core24                     20240710                490    latest/stable    canonical✓        base
firefox                    130.0.1-1               4955   latest/stable/…  mozilla✓          -
gnome-3-34-1804            0+git.3556cb3           93     latest/stable    canonical✓        -
gnome-3-38-2004            0+git.efb213a           143    latest/stable/…  canonical✓        -
gnome-42-2204              0+git.510a601           176    latest/stable/…  canonical✓        -
gtk-common-themes          0.1-81-g442e511         1535   latest/stable/…  canonical✓        -
konf                       0+git.129ff3c           50     latest/stable    canonicalwebteam  -
lxd                        git-aa06597             30396  latest/edge      canonical✓        -
microceph                  0+git.7b5672b           707    quincy/stable    canonical✓        held
microcloud                 1.1-04a1c49             734    latest/stable    canonical✓        -
microovn                   22.03.3+snap0e23a0e4f5  395    22.03/stable     canonical✓        -
parca-agent                0.31.1                  2474   latest/stable    parca-team✓       classic
rockcraft                  1.5.3                   2218   latest/stable    canonical✓        classic
slack                      4.39.95                 158    latest/stable    slack✓            -
snap-store                 41.3-72-g80e7130        1216   latest/stable/…  canonical✓        -
[...]

PR #14804, opened 2025-01-15
Storage: Fix copying storage volumes between cluster nodes when target and project parameters are set

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/14155.

Summary of changes:
- Includes the request project name in the `clusterCopyCustomVolumeInternal` function when building the `storageVolumePost` request. This is required to ensure the request project name is not always default when copying storage volumes with a target project set.
- Fixes storage volume migration error handling by checking if effective project and target project are the same only when the target project is not `default`.
- Adds a test for copying storage volumes across cluster members when both a target and target project are set.


simondeziel (approved):

tomponline:
Thanks!

PR #14806, opened 2025-01-16
Avoid `fmt.Sprintf` where possible as it is slow

simondeziel:
Also:

* adds a missing `url.PathEscape()` to `MoveStoragePoolVolume()`
* don't ask `mkfs.ext4` to skip `discard`


tomponline (approved):
Ta!

PR #14807, opened 2025-01-16
Storage: Align Ceph block device name encoding

MggMuggins:
Different calls of `DiskGetRBDFormat` and `DiskParseRBDFormat` are currently using both senses of "volume":
- A Ceph block device (RBD Image)
- A LXD storage volume

Because a disk device can specify an RBD image with source `ceph:<pool_name>/<volume_name>"`, these functions should use the first meaning: the name passed should be the name of a RADOS block device (RBD image), not a LXD storage volume.

These functions should (in future) be replaced with a more comprehensive abstraction for what `MountInfo.DevPath` is currently doing. This serialization doesn't seem to be written anywhere, it's just deserialized when it gets to the instance driver.

This fixes the failures from https://github.com/canonical/lxd-ci/pull/366; the QEMU driver was inferring a virtual-machine `VolumeType` only when `TargetPath == "/"`

### DIFF
--- a/lxd/instance/drivers/driver_qemu.go
+++ b/lxd/instance/drivers/driver_qemu.go
@@ -3727,7 +3727,9 @@ func (d *qemu) addRootDriveConfig(qemuDev map[string]string, mountInfo *storageP
 				clusterName = storageDrivers.CephDefaultUser
 			}
 
-			driveConf.DevPath = device.DiskGetRBDFormat(clusterName, userName, config["ceph.osd.pool_name"], vol.Name())
+			rbdImageName := storageDrivers.CephGetRBDImageName(vol, "", false)
+
+			driveConf.DevPath = device.DiskGetRBDFormat(clusterName, userName, config["ceph.osd.pool_name"], rbdImageName)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


lxd/instance/drivers/driver_qemu.go
```diff
@@ -3727,7 +3727,11 @@ func (d *qemu) addRootDriveConfig(qemuDev map[string]string, mountInfo *storageP
 				clusterName = storageDrivers.CephDefaultUser
 			}
 
-			driveConf.DevPath = device.DiskGetRBDFormat(clusterName, userName, config["ceph.osd.pool_name"], vol.Name())
+			// Driver and pool name arguments can be ignored as CephGetRBDImageName doesn't need them
+			vol := storageDrivers.NewVolume(nil, "", storageDrivers.VolumeTypeVM, storageDrivers.ContentTypeBlock, vol.Name(), nil, nil)
```
tomponline, line 3731: why are we creating a new locally scope `vol` here rather than using the one above from line 3715?

tomponline (commented):
I agree wrt to the future plans to change how mount info from a disk device is passed to an instance driver. Its gotten rather complicated and obtuse over the years with per-device custom encoding options. 

I'd prefer to see us return different concrete struct types, via interface smuggling using `any`, and then use type assertion in the instance drive to detect the type of struct returned and therefore know which type of disk it is and have access to the custom fields required for that disk type without needing to encode/decode.

When support for using qemu's ceph support was added, this blurred the lines between responsibility of each subsystem.

Before that it was the `disk` device's responsibility to activate and mount the volume and then pass that mount path/block path to the instance driver.  

But after that it became the instance driver's job to know all of the details required to activate the ceph volume.

lxd/instance/drivers/driver_qemu.go
tomponline: I'm thinking that it would be clearer to update `CephGetRBDImageName` to accept specific arguments that it needs rather than a `Volume` type, as it seems several places are having to craft specific `Volume` vars to make it work.

MggMuggins: I missed the `vol` on line 3715, thanks.

`CephGetRBDImageName` uses `vol.ConfigBlockFilesystem()`, so we need the volume's config as well. I've updated all usages so that the config is passed as well, rather than using `NewVolume`.

tomponline (approved):

tomponline (approved):

Issue #14709, opened 2024-12-20
VM: can't set explicit pinning on isolated CPUs


mihalicyn:
This was originally reported on [Discourse](https://discourse.ubuntu.com/t/after-this-years-update-host-isolated-cpu-cores-are-no-longer-being-used-on-lxd-vm-that-isolated-cores-have-been-pinned-to/47814
) and I was able to reproduce it too.

Commit which introduces the issue https://github.com/canonical/lxd/pull/13257/commits/801e0310d5f17dc00784e7c3d06d9bc89e131f02

PR #14817, opened 2025-01-20
Devices: CPU pinning fix (allow explicit pinning on isolated CPUs for VM instances)

mihalicyn:
Fixes: https://github.com/canonical/lxd/issues/14709

### DIFF
--- a/lxd/devices.go
+++ b/lxd/devices.go
@@ -431,6 +431,22 @@ func getNumaNodeToCPUMap() (numaNodeToCPU map[int64][]int64, err error) {
 	return numaNodeToCPU, err
 }
 
+func getNumaCPUs(numaNodeToCPU map[int64][]int64, cpuNodes string) ([]int64, error) {
+	var numaCpus []int64
+	if cpuNodes != "" {
+		numaNodeSet, err := resources.ParseNumaNodeSet(cpuNodes)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, numaNode := range numaNodeSet {
+			numaCpus = append(numaCpus, numaNodeToCPU[numaNode]...)
+		}
+	}
+
+	return numaCpus, nil
+}
+
 // deviceTaskBalance is used to balance the CPU load across instances running on a host.
 // It first checks if CGroup support is available and returns if it isn't.
 // It then retrieves the effective CPU list (the CPUs that are guaranteed to be online) and isolates any isolated CPUs.
@@ -479,17 +495,10 @@ func deviceTaskBalance(s *state.State) {
 	for _, c := range instances {
 		conf := c.ExpandedConfig()
 		cpuNodes := conf["limits.cpu.nodes"]
-		var numaCpus []int64
-		if cpuNodes != "" {
-			numaNodeSet, err := resources.ParseNumaNodeSet(cpuNodes)
-			if err != nil {
-				logger.Error("Error parsing numa node set", logger.Ctx{"numaNodes": cpuNodes, "err": err})
-				return
-			}
-
-			for _, numaNode := range numaNodeSet {
-				numaCpus = append(numaCpus, numaNodeToCPU[numaNode]...)
-			}
+		numaCpus, err := getNumaCPUs(numaNodeToCPU, cpuNodes)
+		if err != nil {
+			logger.Error("Error parsing numa node set", logger.Ctx{"numaNodes": cpuNodes, "err": err})
+			return
 		}
 
 		cpulimit, ok := conf["limits.cpu"]
--- a/lxd/devices.go
+++ b/lxd/devices.go
@@ -25,9 +25,10 @@ import (
 )
 
 type deviceTaskCPU struct {
-	id    int64
-	strID string
-	count *int
+	id       int64
+	strID    string
+	count    *int
+	isolated bool
 }
 
 type deviceTaskCPUs []deviceTaskCPU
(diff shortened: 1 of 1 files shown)


gabrielmougard (dismissed):
One comment but else very good !

lxd/devices.go
```diff
@@ -431,6 +445,32 @@ func getNumaNodeToCPUMap() (numaNodeToCPU map[int64][]int64, err error) {
 	return numaNodeToCPU, err
 }
 
+func makeCPUUsageMap(cpus []int64, isolCpus []int64) map[int64]deviceTaskCPU {
```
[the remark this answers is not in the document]
mihalicyn: Ah, thanks! Let me fix that :)

lxd/devices.go
```diff
@@ -366,53 +366,28 @@ func fillFixedInstances(fixedInstances map[int64][]instance.Instance, inst insta
...
-// Finally, the pinning map is used to set the new CPU pinning for each instance, updating it to the new balanced state.
-//
-// Overall, this function ensures that the CPU resources of the host are utilized effectively amongst all the instances running on it.
-func deviceTaskBalance(s *state.State) {
-	// Don't bother running when CGroup support isn't there
-	if !s.OS.CGInfo.Supports(cgroup.CPUSet, nil) {
-		return
-	}
-
+func getCPULists() (effectiveCpus string, cpus []int64, err error) {
```
tomponline: please can we have a comment on this function explaining what the return values are and their format, primarily the effectiveCpus string.

Also should this be called `effectiveCPUs`?

mihalicyn: Thanks! Fixed!

lxd/devices.go
```diff
@@ -410,6 +410,27 @@ func getCPULists() (effectiveCpus string, cpus []int64, err error) {
 	return effectiveCpus, cpus, nil
 }
 
+func getNumaNodeToCPUMap() (numaNodeToCPU map[int64][]int64, err error) {
```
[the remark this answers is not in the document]
mihalicyn: Have done! Thanks!

lxd/devices.go
```diff
@@ -549,65 +595,56 @@ func deviceTaskBalance(s *state.State) {
 				logger.Warn("The pinned CPUs override the NUMA configuration CPUs", logger.Ctx{"pinnedCPUs": instanceCpus, "numaCPUs": numaCpus})
 			}
 
-			fillFixedInstances(fixedInstances, c, cpus, instanceCpus, len(instanceCpus), false)
+			usableCpus := cpus
+			//
+			// For VM instances, historically, we allow explicit pinning to a CPUs
```
tomponline, line 600: what happens if you try to explicitly pin a container's cpu to one in the isolcpus list, does the container refuse to start?

lxd/devices.go
```diff
@@ -366,50 +380,55 @@ func fillFixedInstances(fixedInstances map[int64][]instance.Instance, inst insta
 	}
 }
 
-func getCPULists() (effectiveCpus string, cpus []int64, err error) {
+// getCPULists returns lists of CPUs:
+// effectiveCPUs: list of CPUs which are available for instances to be automatically pinned on
```
tomponline, line 384: what format does this string come back in?
Is there a need to keep this returned string given that `cpus` return value is the same data just as an int slice?

lxd/devices.go
mihalicyn: Nope, historically, it will just silently ignore such pinning and produce a warning in the logs `Internal error: instance using unavailable cpu`. I'm not changing behavior for containers here for compatibility reasons.

lxd/devices.go
mihalicyn: > what format does this string come back in?

It's just a comma-separated list of CPUs like `3,4,5`.

> Is there a need to keep this returned string given that `cpus` return value is the same data just as an int slice?

we use this CPU list formatted as a string one time in the code:
```
		cpulimit, ok := conf["limits.cpu"]
		if !ok || cpulimit == "" {
			// For VMs empty limits.cpu means 1,
			// but for containers it means "unlimited"
			if c.Type() == instancetype.VM {
				cpulimit = "1"
			} else {
				cpulimit = effectiveCPUs // <<<<< HERE
			}
		}
```

but I guess we can get rid of that case with something like this:
```
			} else {
				cpulimit = strconv.FormatInt(len(cpus), 10),
			}
```
this means that instead of explicitly specifying all available CPUs as a comma-separated list, we just set a number of CPUs and then instance will be automatically balanced to use the same list of CPUs. It's a bit less explicit but seems logically equivalent.

and then we can get rid of this extra return value. What do you think?

lxd/devices.go
tomponline: OK thanks. Can we update the logged message to use contextual logging (i.e not `logger.Errorf`) and include the instance name, project name and, perhaps CPU ID?

lxd/devices.go
mihalicyn: >can we use something like

yes, we can. But I went with even simpler way `cpulimit = strconv.Itoa(len(cpus))` instead of `cpulimit = effectiveCPUs`. Which is logically equivalent.

> What im looking to achieve here is separate the logic of CPU selection from the formatting of said CPU selection into the instance config.

yeah, I like this idea!

I have removed this redunant `effectiveCPUs` variable.

lxd/devices.go
mihalicyn: I wasn't fully right when I said that this incorrect container pinning produces `Internal error: instance using unavailable cpu`. Actually, this error is actually internal and it's not easy to get it. But what I did is added a new:
```
				logger.Warn("Instance using unavailable cpu", logger.Ctx{"project": inst.Project().Name, "instance": inst.Name(), "cpu": nr})
```
in the `fillFixedInstances` function.

lxd/devices.go
```diff
@@ -431,6 +431,22 @@ func getNumaNodeToCPUMap() (numaNodeToCPU map[int64][]int64, err error) {
 	return numaNodeToCPU, err
 }
 
+func getNumaCPUs(numaNodeToCPU map[int64][]int64, cpuNodes string) ([]int64, error) {
```
tomponline, line 480: missing doc block on this function.

As it would be useful to the caller and reviewer to understand what is in the returned values and what format the cpuNodes string is in.

lxd/devices.go
tomponline: thanks!

lxd/devices.go
```diff
@@ -549,65 +595,56 @@ func deviceTaskBalance(s *state.State) {
...
-		usage[id] = cpu
-	}
-
-	for cpu, ctns := range fixedInstances {
+	// Handle instances with explicit CPU pinnings
+	for cpu, instances := range fixedInstances {
 		c, ok := usage[cpu]
 		if !ok {
-			logger.Error("Internal error: instance using unavailable cpu")
+			logger.Error("Internal error: instance using unavailable cpu", logger.Ctx{"cpu": cpu})
```
tomponline, line 620: do we still need this error now we have the other error? As it doesnt have the context of the instance, it'd be difficult to use to identity the problem.

tomponline (approved):

lxd/devices.go
mihalicyn: yes, as this error would indicate a bug in our code rather than misconfiguration.

PR #14830, opened 2025-01-21
Storage: Resize instance on `size` unset

hamistao:
Before this change, unsetting a size config on a disk, as in
`lxc config device unset isnt root size`, doesn't resize the volume and only unsets the config key, masking the instance's real size. This now tries to resize the volume to the default size for that pool when unsetting the size config.

### DIFF
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -3323,8 +3323,15 @@ func (b *lxdBackend) SetInstanceQuota(inst instance.Instance, size string, vmSta
 		return err
 	}
 
-	// Apply the main volume quota.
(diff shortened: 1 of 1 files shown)


lxd/storage/backend_lxd.go
```diff
@@ -3323,8 +3323,15 @@ func (b *lxdBackend) SetInstanceQuota(inst instance.Instance, size string, vmSta
 		return err
 	}
 
-	// Apply the main volume quota.
+	// If no size is provided, block volumes should get resized to the pool's default size for block volumes.
```
tomponline, line 3326: @hamistao im not sure we want this behavior, lets discuss before going forward with this.

Issue #14141, opened 2024-09-20
OIDC - oidc.groups.claim is added as additional scope which breaks Microsoft Entra integration


mrbranan:
I just worked on integrating LXD with Microsoft Entra ID. When you create an app registration in Entra for LXD, you can add a group claim to your token configuration. That claim is called "groups". If you then set oidc.groups.claim = "groups", when you attempted to log in, LXD adds "groups" as an additional requested scope. This caused Entra to return an error because that scope does not exist. 

Claims and scopes are two different things, and you shouldn't assume that there will be a scope with the same name as your groups claim. I would suggest that you create a separate configuration parameter for users to specify any additional scopes that might be needed (maybe "oidc.additional_scopes"?)

As a work around, I had to choose a scope that Entra would consider valid, and then create a Custom claims provider to inject a claim with that same name into the token.

ech0-de:
This issue is also relevant when trying to use GitLab as an identity provider. GitLab shares the groups a user is a member of as `groups` and `groups_direct` claims, which are included in the `openid` scope. However, requesting `groups` or `groups_direct` as a scope, will cause the requests to be rejected due to invalid scopes.

However, I also ran into another problem with GitLab due to the `offline_access` scope, which would neither be solved by having an `oidc.additional_scopes` setting, nor by introducing an `oidc.groups.scope` setting.

What do you think about instead introducing an `oidc.scopes` setting instead, which overwrites the current behavior of adding the groups claim as an additional scope as well as the default scopes that are requested. This change would be backword-compatible by reverting back to the current behavior if not defined, and enables lxd to work with more peculiar identity providers, such as GitLab.

I created a [pull request](https://github.com/canonical/lxd/pull/14832) to add this setting, as a starting point for discussion :)

markylaing:
@ech0-de @mrbranan With the addition of the `oidc.scopes` configuration key in #14935 you can now remove `offline_access` and the `oidc.groups.claim` config key is no longer used as a scope. The feature is only available in `latest/edge` at the moment.

Apologies for the turn around time on this. If you could confirm it is now working for you that would be great.

ech0-de:
> Apologies for the turn around time on this. If you could confirm it is now working for you that would be great.

Sorry for the delayed reply. I just tested the changes and it works great for my use case, thanks for working on this!

PR #14832, opened 2025-01-21
Auth: Make OIDC scopes configurable

ech0-de:
This PR adds a new server configuration key `oidc.scopes`. When set, the value of this config key is used to determine which scopes to request during OIDC flows.

The current behavior is hardcoded to request the `openid offline_access profile email`. The `offline_access` scope is marked as optional in the OpenID Connect specification and is not supported by all auth providers. For example, [GitLab as an OpenID Connect identity provider](https://docs.gitlab.com/ee/integration/openid_connect_provider.html) rejects requests with the `offline_access` scope, even though it supports refresh tokens. The proposed `oidc.scopes` option could be used in this case, to only request `openid profile email` scopes, so that LXD can be used with GitLab as an identity provider.

Additionally, #14141 shows how the `oidc.groups.claim` configuration option is added as an additional scope which breaks various identity providers. This pull request also provides a backwards-compatible fix for this issue. As long as `oidc.scopes` is not configured, the current behavior is unchanged and the default scopes are used with the value of `oidc.groups.claim` as an additional scope. When `oidc.scopes` is set, it overwrites all scopes, so that the scopes can be configured to work with existing identity providers.

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2552,3 +2552,8 @@ This adds support for listing network zones across all projects using the `all-p
 Adds support for instance root volumes to be attached to other instances as disk
 devices. Introduces the `<type>/<volume>` syntax for the `source` property of
 disk devices.
+
+## `oidc_scopes`
+
+This optional API extension enables setting an {config:option}`server-oidc:oidc.scopes` configuration key, which can be used to overwrite the default scopes that are requested during the OIDC flow.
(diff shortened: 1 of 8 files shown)


tomponline:
Thanks for this! This has prompted additional discussion and will be handled by this PR https://github.com/canonical/lxd/pull/14888

PR #14837, opened 2025-01-21
Storage: Use zero for unbound volumes' total size on the API

hamistao:
As per discussed on https://github.com/canonical/lxd/pull/14595#issuecomment-2589868308, this implements the standard of showing the size of unbound volumes as 0.

To keep the API consistent, also updates getting custom volume state so that total is shown even when usage is not supported and shows 0 to represent unsupporte usage of both custom and instance volumes.

### DIFF
--- a/doc/rest-api.yaml
+++ b/doc/rest-api.yaml
@@ -2516,13 +2516,13 @@ definitions:
     InstanceStateDisk:
         properties:
             total:
-                description: Total size in bytes. Uses -1 to convey that the instance has access to the entire pool's storage.
+                description: Total size in bytes. Uses 0 to convey that the instance has access to the entire pool's storage.
                 example: 502239232
                 format: int64
                 type: integer
                 x-go-name: Total
             usage:
-                description: Disk usage in bytes. Uses -1 to indicate that the storage driver for the instance's pool does not support retrieving the disk usage.
+                description: Disk usage in bytes. Uses 0 to indicate that the storage driver for the pool does not support retrieving disk usage.
                 example: 502239232
                 format: int64
                 type: integer
@@ -6703,13 +6703,13 @@ definitions:
         description: StorageVolumeStateUsage represents the disk usage of a volume
         properties:
             total:
-                description: Storage volume size in bytes
+                description: Storage volume size in bytes. Uses 0 to convey that the volume has access to the entire pool's storage.
                 example: 5189222192
                 format: int64
                 type: integer
                 x-go-name: Total
             used:
-                description: Used space in bytes
+                description: Used space in bytes. Uses 0 to indicate that the storage driver for the pool does not support retrieving volume usage.
                 example: 1693552640
                 format: uint64
                 type: integer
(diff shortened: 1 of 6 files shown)


hamistao:
@tomponline In addition, we did not reach a conclusion regarding the value for usage in our last 1-1. But throughout our discussions we stablished that:
  1- It would be nice for the value we use for Usage when not supported to match the value for an unset total.
  2- Total should be 0.
  3- -1 is a value typically used to communicate an error.

For these reasons, volume/disk usage is being set here to be reported as 0 if not supported and -1 if an error ocurred while trying to retrieve its value.

tomponline:
> For these reasons, volume/disk usage is being set here to be reported as 0 if not supported and -1 if an error ocurred while trying to retrieve its value.

Doea this change any existing usage values compared to lxd 6.2?

tomponline:
> 1- It would be nice for the value we use for Usage when not supported to match the value for an unset total. 2- Total should be 0. 3- -1 is a value typically used to communicate an error.
> 
> For these reasons, volume/disk usage is being set here to be reported as 0 if not supported and -1 if an error ocurred while trying to retrieve its value.

What situations are there when usage is not suppprted/unavailable. I suspect we may want to consider unavailable/unsupported as an error (-1) to allow differentiation from zero usage (a possibility as a genuine value).

tomponline:
@hamistao does this PR effectively restore the behaviour that we had in LXD 6.1?

tomponline:
Thanks. Did you check each storage driver's implementation for total size 0 if unbounded?

tomponline:
@hamistao i think we'll land this one for LXD 6.3 to partially restore the previous behaviour.

hamistao:
I agree.

Thinking practically, I suppose our users will hardly be bothered by this reverting.
 This above is conjecture, but only the value for unsupported usage and the instance state's total are affected, and the latter is so broken and useless that I doubt anyone is currently relying on it.

tomponline:
> I agree.
> 
> Thinking practically, although this is just conjecture, I suppose our users will hardly be bothered by this reverting. I say this because only the value for unsupported usage and the instance state's total are affected, and the latter is so broken and useless that I doubt anyone is currently relying on it.

Yes it didnt make it into any LTS which is good.

tomponline:
@hamistao so is this good to merge?

hamistao:
@tomponline I am concerned about this failing test, please give me some time to check on it

hamistao:
@tomponline Sorry, I should have pinged you after I fixed the broken tests. This is ready.

tomponline (approved):
Thanks!

tomponline:
@edlerd is this OK for LXD UI's consumption ?

edlerd:
> @edlerd is this OK for LXD UI's consumption ?

Yes, that seems all right.

tomponline:
@MusicDin @roosterfish are all volumes in Powerflex and Pure bounded in size? Please can you double check they use the convention specified in this PR? Thanks

PR #14840, opened 2025-01-22
lxd: Factor out delete endpoint logic

boltmark:
The logic for the delete endpoints for images, instances, networks, network zones, network ACLs, profiles, and storage pool volumes has been factored out such that it can be invoked without going through the lxd client API.

### DIFF
--- a/lxd/networks.go
+++ b/lxd/networks.go
@@ -1023,6 +1023,66 @@ func doNetworkGet(s *state.State, r *http.Request, allNodes bool, requestProject
 	return apiNet, nil
 }
 
+// handleNetworkDelete provides the logic for deleting a network.
+func handleNetworkDelete(s *state.State, n network.Network, effectiveProjectName string, networkName string, details networkDetails, clientType clusterRequest.ClientType, clusterNotification bool, requestor *api.EventLifecycleRequestor) error {
+	// Check if project allows access to network.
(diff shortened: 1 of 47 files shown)


tomponline (changes requested):
Thanks for this @boltmark 

Please can you break it out into a logical commit per API endpoint being re-worked to aid review.

Thanks

boltmark:
> Thanks for this @boltmark
> 
> Please can you break it out into a logical commit per API endpoint being re-worked to aid review.
> 
> Thanks

Good point, should have done this in the first place. Done now!

lxd/networks.go
```diff
@@ -1023,6 +1023,66 @@ func doNetworkGet(s *state.State, r *http.Request, allNodes bool, requestProject
 	return apiNet, nil
 }
 
+// handleNetworkDelete provides the logic for deleting a network.
+func handleNetworkDelete(s *state.State, n network.Network, effectiveProjectName string, networkName string, details networkDetails, clientType clusterRequest.ClientType, clusterNotification bool, requestor *api.EventLifecycleRequestor) error {
```
tomponline, line 1027: you have access to the network name via n.Name() so why do you need the `networkName` arg too?

tomponline:
@boltmark @escabo as discussed, will close this for now until we've investigated Temporal more.

PR #14841, opened 2025-01-22
Network: Begin transition to libovsdb

markylaing:
This begins a series of cherry-picks and amendments to transition LXD to use libovsdb, rather than OVN CLI tooling.


tomponline:
@markylaing im going to wait until after LXD 6.3 to merge this btw.

markylaing:
> @markylaing im going to wait until after LXD 6.3 to merge this btw.

Yep no problem. There are some issues with it in any case.

markylaing:
Tests passing now (both in the main test suite and the `network-ovn` LXD CI tests).

A few things for discussion:
- The code scanning errors above. I'm not sure if they're a false positive since the OVN CA certificate it used it present.
- When adding the `network.ovs.connection` config key, the "scope" of the key in the Incus commit was set to "global", but it was added to the local config for the node. I changed the scope to "local" with the thinking that OVS is a per-cluster member config. We could change it to "global" and move to `lxd/cluster/config` instead, and require that the path be the same on all members.
- Also when adding the `network.ovs.connection` I cherry-picked [this commit](https://github.com/lxc/incus/pull/1159/commits/ab44781257c29ab35c51a629f113f8e4ac0df66f) (see [here](https://github.com/canonical/lxd/pull/14841/commits/555322d9c4a1cf64611b3d0b023e730bb998bae1)) to update the call sites. However, I'm not sure I should have because `VSwitch` is not yet defined on our state (it's not long-lived yet) so I had to edit all the calls to pass in the connection string instead. I'm thinking to remove this cherry-pick and implement myself, then cherry-pick the commit when we actually move to long-lived connections.

PR #14842, opened 2025-01-23
Device: Remove existing GPU CDI device files before new device files are added

gabrielmougard:
If a host machine is not shut down properly [ i.e. it loses power ], the instances having CDI GPU attached to them won't start again after the host is started, even if a start command is issued manually.

LXD returns as exception:

`Failed to start device "nvidia-gpu": Failed to create device "/var/snap/lxd/common/lxd/devices/oel-ogrp623/cdi.unix.nvidia--gpu.dev-nvidia0" for "/dev/nvidia0": file exists`

To solve that, we must remove any remaining device files before adding new CDI device files in the instance GPU device directory. These old files are still present if the host crash because the GPU device stop hook is not called.

Fixes https://github.com/canonical/lxd/issues/14843

### DIFF
--- a/lxd/device/gpu_physical.go
+++ b/lxd/device/gpu_physical.go
@@ -3,6 +3,7 @@ package device
 import (
 	"encoding/json"
 	"fmt"
+	"io/fs"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -157,6 +158,27 @@ func (d *gpuPhysical) startCDIDevices(configDevices cdi.ConfigDevices, runConf *
 		minor, err := strconv.ParseUint(conf["minor"], 10, 32)
 		if err != nil {
 			return fmt.Errorf("Failed to parse minor number %q when starting CDI device: %w", conf["minor"], err)
+		}
+
+		// Check if there are any remaining CDI devices in the instance devices directory.
+		// If there are, we need to remove them. These can be present in the case where the device stop hook was not called
+		// (e.g. due to an abrupt host shutdown).
+		err = filepath.WalkDir(d.inst.DevicesPath(), func(path string, e fs.DirEntry, _ error) error {
+			if e.IsDir() {
+				return nil
+			}
+
+			if strings.HasPrefix(e.Name(), cdi.CDIUnixPrefix+"."+d.name) {
+				err := os.Remove(path)
+				if err != nil {
+					return err
+				}
+			}
+
+			return nil
+		})
+		if err != nil {
+			return err
 		}
 
 		// Here putting a `cdi.CDIUnixPrefix` prefix with 'd.name' as a device name will create an directory entry like:


gabrielmougard:
@tomponline I'm wondering how we can test that in lxd-ci though. We would need to start a VM, do a physical GPU passthrough to the VM first, create a container inside the VM and add this GPU as a CDI GPU, force stop the VM, restart it, and try to start the container. Can we do such a thing in lxd-ci?

lxd/device/gpu_physical.go
```diff
@@ -159,6 +160,27 @@ func (d *gpuPhysical) startCDIDevices(configDevices cdi.ConfigDevices, runConf *
 			return fmt.Errorf("Failed to parse minor number %q when starting CDI device: %w", conf["minor"], err)
 		}
 
+		// Check if there are any remaining CDI devices in the instance devices directory.
+		// If there are, we need to remove them. These can be present in the case where the device stop hook was not called
+		// (e.g. due to an abrupt host shutdown).
+		err = filepath.Walk(d.inst.DevicesPath(), func(path string, info fs.FileInfo, err error) error {
```
tomponline, line 166: As per https://pkg.go.dev/path/filepath#Walk

> Walk is less efficient than [WalkDir](https://pkg.go.dev/path/filepath#WalkDir), introduced in Go 1.16, which avoids calling os.Lstat on every visited file or directory.

So as we aren't using `info` argument here, it'd make sense to use `WalkDir` instead to save the extra syscalls.

tomponline: its still using Walk and WalkDir so not fixed

gabrielmougard: ah sorry. Now it is updated

tomponline:
@gabrielmougard 

Error: lxd/device/gpu_physical.go:166:81: SA4009: argument err is overwritten before first use (staticcheck)
		err = filepath.WalkDir(d.inst.DevicesPath(), func(path string, e fs.DirEntry, err error) error {
		                                                                              ^
Error: lxd/device/gpu_physical.go:172:5: SA4009(related information): assignment to err (staticcheck)
				err = os.Remove(path)

gabrielmougard:
@tomponline also, I opened this draft PR on lxd-ci on the side: https://github.com/canonical/lxd-ci/pull/393

lxd/device/gpu_physical.go
```diff
@@ -159,6 +160,27 @@ func (d *gpuPhysical) startCDIDevices(configDevices cdi.ConfigDevices, runConf *
 			return fmt.Errorf("Failed to parse minor number %q when starting CDI device: %w", conf["minor"], err)
 		}
 
+		// Check if there are any remaining CDI devices in the instance devices directory.
+		// If there are, we need to remove them. These can be present in the case where the device stop hook was not called
+		// (e.g. due to an abrupt host shutdown).
+		err = filepath.WalkDir(d.inst.DevicesPath(), func(path string, e fs.DirEntry, err error) error {
+			if e.IsDir() {
+				return nil
```
tomponline, line 168: why wouldnt we deleted dirs?

tomponline: shouldnt we check err here too ?

tomponline: > The err argument reports an error related to path, signaling that [WalkDir](https://pkg.go.dev/io/fs#WalkDir) will not walk into that directory

https://pkg.go.dev/path/filepath#WalkFunc

So perhaps we should ignore it and just try to delete what we can. In which case we should replace `err error` with `_ error` to show we're not using it.

lxd/device/gpu_physical.go
```diff
@@ -159,6 +160,27 @@ func (d *gpuPhysical) startCDIDevices(configDevices cdi.ConfigDevices, runConf *
...
+			}
+
+			if strings.HasPrefix(e.Name(), fmt.Sprintf("%s.%s", cdi.CDIUnixPrefix, d.name)) {
```
tomponline, line 171: lets not use `fmt.Sprintf()` for concatenating strings as it uses reflection which is more expensive than just doing:

```go
if strings.HasPrefix(e.Name(), cdi.CDIUnixPrefix+"."+d.name,) {
```

lxd/device/gpu_physical.go
gabrielmougard: fixed

lxd/device/gpu_physical.go
gabrielmougard: fixed

tomponline (approved):

PR #14844, opened 2025-01-23
Add missing `swtpm` command for container_devices_tpm test

simondeziel:
ATM, GHA runners don't have the needed kconfig:

```
$ grep TPM_PROXY /boot/config-*
/boot/config-6.8.0-1020-azure:# CONFIG_TCG_VTPM_PROXY is not set
```
but those running the test suite locally should have no problem when using a non-Azure kernel (like `linux-image-virtual`).

### DIFF
--- a/.github/actions/install-lxd-runtimedeps/action.yml
+++ b/.github/actions/install-lxd-runtimedeps/action.yml
@@ -37,6 +37,7 @@ runs:
           socat \
           sqlite3 \
           squashfs-tools \
+          swtpm \
           tar \
           tcl \
           thin-provisioning-tools \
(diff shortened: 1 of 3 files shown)


tomponline (approved):

tomponline:
Was the test previously being skipped?

simondeziel:
> Was the test previously being skipped?

Yep

simondeziel:
@tomponline GHA runners are a missing the require kernel module but if they ever gain it, the test would start running. I've added an early skip if the kmod is not available.

This tested fine locally though.

tomponline (approved):

PR #14845, opened 2025-01-23
Godoc deprecation notice

simondeziel:
Deprecation notices need to be on their own paragraph so all our `// Deprecated: ` markers were missing a `//` before them.

https://go.dev/blog/godoc:
> To signal that an identifier should not be used, **add a paragraph to its doc comment** that begins with “Deprecated:” followed by some information about the deprecation.

This PR fixes most of them except for `RunCommand` as that's used in so many places.

### DIFF
--- a/client/lxd_containers.go
+++ b/client/lxd_containers.go
@@ -21,10 +21,12 @@ import (
 
 // Container handling functions
 //
+//
 // Deprecated: Those functions are deprecated and won't be updated anymore.
 // Please use the equivalent Instance function instead.
 
 // GetContainerNames returns a list of container names.
+//
 // Deprecated: Use GetInstanceNames instead.
 func (r *ProtocolLXD) GetContainerNames() ([]string, error) {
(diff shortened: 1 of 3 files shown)


simondeziel:
The depration of `HasExtension` and `btrfsIsSubVolume` should be looked at more closely. The former is still being used by the function that's meant to be it's replacement.

The `btrfsIsSubVolume` still has one caller but apparently that's not tripping the staticcheck linter, maybe because it's not exported?

tomponline (approved):
Ta

simondeziel:
> The deprecation of `HasExtension` and `btrfsIsSubVolume` should be looked at more closely. The former is still being used by the function that's meant to be it's replacement.

@roosterfish if you have time to comment on the above, I'd appreciate it!
 
> The `btrfsIsSubVolume` still has one caller but apparently that's not tripping the `staticcheck` linter, maybe because it's not exported?

Same here, if anyone has guidance on what to do, I'm all ears.

PR #14851, opened 2025-01-24
client: Fall back to OIDC `verification_uri`

markylaing:
The client was attempting to open the browser at `verification_uri_complete` which is marked as optional and might not be populated: https://www.rfc-editor.org/rfc/rfc8628#section-3.2

This commit falls back to the `verification_uri` if the former is not present.


tomponline:
Added "OIDC" to PR description for ease of search-ability later.

tomponline (approved):

PR #14865, opened 2025-01-27
Storage: PowerFlex fixes and improvements

roosterfish:
This has to land after https://github.com/canonical/lxd/pull/14710 got merged as it was tested together. 

### Summary of changes

1. Don't map the volume before actually performing the resize in `SetVolumeQuota`. As we can get the current volume's size directly from PowerFlex it does not need to be mapped beforehand. This also allows reducing errors in `dmesg` caused by the resize in case the vol is already mapped to the system.
2. Prevent volume creation in case the PowerFlex pool isn't using zero-padding. This is to prevent leaking data from a previous volume in case the storage pool on PowerFlex wasn't setup appropriately.
3. Lock the PowerFlex client's API token to not cause concurrent logins and to save time in case of parallel access.
4. Always copy the PowerFlex client's request body for retries. Before, if the to be retried request contained a request body, it was already read in the first request which leads any follow up request to not anymore have a request body.
5. Small improvement to use `NewStatusError`
6. Always allow the filler to resize as PowerFlex doesn't use optimized image storage. This caused issues when trying to launch images from an image that were bigger than the volume size. There is now a test for this in `lxd-ci/tests/storage-vm`.
7. Move common disk discovery functions into the `block` package.

### DIFF
--- a/lxd/storage/connectors/utils.go
+++ b/lxd/storage/connectors/utils.go
@@ -2,145 +2,14 @@ package connectors
 
 import (
 	"context"
-	"errors"
 	"fmt"
-	"path/filepath"
-	"strings"
 	"sync"
 	"time"
 
-	"golang.org/x/sys/unix"
-
 	"github.com/canonical/lxd/lxd/locking"
-	"github.com/canonical/lxd/lxd/resources"
-	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/logger"
 	"github.com/canonical/lxd/shared/revert"
 )
-
-// devicePathFilterFunc is a function that accepts device path and returns true
-// if the path matches the required criteria.
-type devicePathFilterFunc func(devPath string) bool
-
-// GetDiskDevicePath checks whether the disk device with a given prefix and suffix
-// exists in /dev/disk/by-id directory. A device path is returned if the device is
-// found, otherwise an error is returned.
-func GetDiskDevicePath(diskNamePrefix string, diskPathFilter devicePathFilterFunc) (string, error) {
-	devPath, err := findDiskDevicePath(diskNamePrefix, diskPathFilter)
-	if err != nil {
-		return "", err
-	}
-
-	if devPath == "" {
-		return "", fmt.Errorf("Device not found")
-	}
-
-	return devPath, nil
-}
-
-// WaitDiskDevicePath waits for the disk device to appear in /dev/disk/by-id.
-// It periodically checks for the device to appear and returns the device path
-// once it is found. If the device does not appear within the timeout, an error
-// is returned.
-func WaitDiskDevicePath(ctx context.Context, diskNamePrefix string, diskPathFilter devicePathFilterFunc) (string, error) {
-	var err error
-	var diskPath string
-
-	_, ok := ctx.Deadline()
-	if !ok {
-		// Set a default timeout of 30 seconds for the context
-		// if no deadline is already configured.
-		var cancel context.CancelFunc
-		ctx, cancel = context.WithTimeout(ctx, 30*time.Second)
-		defer cancel()
-	}
-
-	for {
-		// Check if the device is already present.
-		diskPath, err = findDiskDevicePath(diskNamePrefix, diskPathFilter)
-		if err != nil && !errors.Is(err, unix.ENOENT) {
-			return "", err
-		}
-
-		// If the device is found, return the device path.
-		if diskPath != "" {
-			break
-		}
-
-		// Check if context is cancelled.
-		err := ctx.Err()
-		if err != nil {
-			return "", err
-		}
-
-		time.Sleep(500 * time.Millisecond)
-	}
-
-	return diskPath, nil
-}
-
-// findDiskDevivePath iterates over device names in /dev/disk/by-id directory and
-// returns the path to the disk device that matches the given prefix and suffix.
-// Disk partitions are skipped, and an error is returned if the device is not found.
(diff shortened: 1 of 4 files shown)


roosterfish:
Linter errors should disappear with https://github.com/canonical/lxd/pull/14710.

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -250,17 +252,28 @@ func (p *powerFlexClient) request(method string, path string, body io.Reader, re
 func (p *powerFlexClient) requestAuthenticated(method string, path string, body io.Reader, response any) error {
 	retries := 0
 	for {
-		err := p.login()
+		// Returns a copy of the client's token.
+		// This allows issuing multiple requests in parallel.
+		// Otherwise the lock on the token can only be released after
+		// the actual request has been finished as the token might be expired.
+		token, err := p.login()
```
MusicDin, line 259: Should this be wrapped with the lock as well (similar to how it is locked when reseting the token)?

Maybe having p.getTokenLock and p.setTokenLock could ensure safe access?

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -274,30 +287,36 @@ func (p *powerFlexClient) requestAuthenticated(method string, path string, body
...
+		err = p.request(http.MethodPost, "/rest/auth/login", "", body, &actualResponse)
+		if err != nil {
+			return "", fmt.Errorf("Failed to login: %w", err)
+		}
 
-	err = p.request(http.MethodPost, "/rest/auth/login", body, &actualResponse)
-	if err != nil {
-		return fmt.Errorf("Failed to login: %w", err)
+		// Save the new token for later use.
+		p.token = actualResponse.AccessToken
```
MusicDin, line 323: Similar to above comment, should we acquire token lock here?

lxd/storage/connectors/utils.go
```diff
@@ -132,6 +133,29 @@ func WaitDiskDeviceGone(ctx context.Context, diskPath string) bool {
 	}
 }
 
+// WaitDiskDeviceResize waits until the disk device reflects the new size.
+func WaitDiskDeviceResize(ctx context.Context, diskPath string, newSizeBytes int64) error {
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
MusicDin, line 138: Check if deadline is already configured

roosterfish: I have switched it to what you have added in https://github.com/canonical/lxd/pull/14887

lxd/storage/drivers/driver_powerflex_utils.go
roosterfish: The lock is obtained within `login()` right at the beginning. 
I just haven't put it outside of the func to not leave it up to the caller to obtain the lock.

lxd/storage/drivers/driver_powerflex_utils.go
roosterfish: The lock is obtained at the beginning of the func.

lxd/storage/drivers/driver_powerflex_utils.go
MusicDin: Ahh thanks for clarifying, I've missed that one

roosterfish:
Running the test suite now for `nvme` and `sdc` mode, will mark ready for review on success.

lxd/storage/connectors/utils.go
```diff
@@ -132,6 +133,35 @@ func WaitDiskDeviceGone(ctx context.Context, diskPath string) bool {
 	}
 }
 
+// WaitDiskDeviceResize waits until the disk device reflects the new size.
+func WaitDiskDeviceResize(ctx context.Context, diskPath string, newSizeBytes int64) error {
```
MusicDin, line 137: Just to confirm this goes into connectors package and not into `block` package?
Noticed that when cherry-picking this commit into Pure PR, where I used `block.WaitDiskDeviceResize`.

roosterfish: As it follows the same approach as `WaitDiskDeviceGone` I put it there. Do you think it's better placed in `block`?

tomponline: Block makes more sense to me. Might be useful for non-nvme/iscsi situations (ceph)?

roosterfish: Good call with Ceph, also created https://github.com/canonical/lxd/issues/14897 so we can fix this for Ceph too.
Then I'll move it over there.

MusicDin: My opinion was since `WaitDiskDevice` relies on the `block.DiskSizeBytes(diskPath)` and is not used within the connectors package itself, it reduces one dependency between those two subpackages. Could be as well used for non-nvme/iscsi situations, but as Julian pointed out it also nicely fits with `WaitDiskDeviceGone`, `WaitDiskDevicePath`.. within connectors package :)

roosterfish: Fixed.

tomponline: Those other wait ones make sense to move to `block` too.

roosterfish: Moved.

roosterfish:
Tests for both NVMe and SDC passed successfully.

MusicDin (approved):

tomponline (approved):
LGTM thanks!

PR #14867, opened 2025-01-27
Projects: Deny restricting a project that is using a forbidden uplink a project that is using a forbidden uplink

hamistao:
As discussed in https://github.com/canonical/lxd/pull/14631#discussion_r1926820927.

On the "restricted" key metadata, it is stated that "To temporarily remove the restrictions, you can disable this option instead of clearing the related keys", which makes me think the behavior of only enforcing project restrictions from the point restricted is enabled is intended and users may be relying on it.

Mind that many project restrictions can still be broken if they were already broken at the time the project was set as "restricted".

Changes to LXD CI tests will probably be needed in order to reflect these changes.


test/suites/clustering_move.sh
```diff
@@ -170,7 +170,7 @@ EOF
   # - c2 is deployed on node2
   # - c3 is deployed on node3
   # - default project restricted to cluster groups foobar1,foobar2
-  LXD_DIR="${LXD_ONE_DIR}" lxc project set default restricted=true
+  LXD_DIR="${LXD_ONE_DIR}" lxc project set default restricted=true restricted.networks.uplinks=${bridge}
```
tomponline, line 173: @hamistao what is this change of behaviour, is this something you wanted to flag up/are concerned by?

hamistao: This line was restricting a project that was already using an uplink, and that is not allowed anymore, as of this PR. So to prevent the test from failing, I am adding the uplink that is already in use to the uplink allowlist.

tomponline: @hamistao so just so im clear, previously, before this PR, if I had `restricted=true` but `restricted.networks.uplinks` unset (or empty) could I use any uplink or no uplinks?

hamistao: no uplinks, as opposed to most restricted keys, it is an allowList as opposed to a denyList (not sure why).

Here is the `restricted.networks.uplinks` metadata description: `// Specify a comma-delimited list of network names that can be used as uplink for networks in this project.`

tomponline: Not really, works the same as `restricted.cluster.groups` and `restricted.networks.zones`, but it looks like `restricted.networks.access` is somewhat different indeed, in that by default it allows all. This likely came from how disruptive it would be to start limiting this on all existing restricted projects.

tomponline (approved):
Thanks!

test/suites/clustering_move.sh
hamistao: Oh indeed, also `restricted.devices.disk.paths` and `restricted.idmap.uid` follow this pattern. Allow/Block keys usually default to block and I may have switched these concepts in my memory.

Issue #14835, opened 2025-01-21
Config update for loki.api.url is blocking for a long time


edlerd:
# Required information

 * Distribution: snap
 * Distribution version: lxd     git-4010d6d  32032  latest/edge

# Issue description

Updates to the `loki.api.url` config are eventually blocking for up to 10 minutes. I think this should be changed to more reasonable timeouts, likely less than a minute, maybe 10 seconds. This happens when the URL is invalid, and I change it to something valid. I suspect the [loki stop](https://github.com/canonical/lxd/blob/main/lxd/loki/loki.go#L223) function might be the problem.

In some cases I had no messages appear in loki, but after a `sudo systemctl reload snap.lxd.daemon` it started working.

# Steps to reproduce

 1. lxc config set loki.api.url=http://127.0.0.1:12345
 2. lxc config set loki.api.url=http://127.0.0.1:1234

The 2nd call takes up to 10 minutes. If the cli is cancelled, a `lxc config get loki.api.url` shows the new config as applied.

tomponline:
Which of the URLs in your example is an invalid one?

edlerd:
> Which of the URLs in your example is an invalid one?

Both are invalid. But only on the 2nd call is blocking for a long time. Hence, my suspicion of the Loki stop and close to be the cause for it.

kadinsayani:
The issue is in the `sendBatch` function, specifically the number of retries and the retry interval. The config key is set even when the Loki address is invalid, so the maximum number of retries (30) at the retry interval of 10 seconds must complete before the `Stop` function runs. We should validate the Loki address (perhaps with a HEAD request) before setting up the client and attempting to send batches, and possibly consider reducing the number of retries.

edlerd:
Loki could become unresponsive or change ip or go down after it was correctly configured. Can the the stop call force quit pending send batch calls after a grace period of like 10 seconds or so?

kadinsayani:
Yes, the `Stop` function could force quit the pending `sendBatch` calls, but I think `Stop` was designed to wait for the calls (including retries) to complete so logging data loss doesn't occur, hence why we have the `wg.Wait()` call.

I think we should validate Loki is reachable on the address and port passed in to `lxc config set loki.api.url`.

edlerd:
> I think we should validate that Loki is reachable on the address and port passed into `lxc config set loki.api.url`.

I agree, that sounds like a good improvement.

[...]

PR #14868, opened 2025-01-27
CLI: Improve config key update for `loki.api.url`

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/14835.

cc. @edlerd

### DIFF
--- a/lxd/loki/loki.go
+++ b/lxd/loki/loki.go
@@ -55,7 +55,6 @@ type entry struct {
 type Client struct {
 	cfg     config
 	client  *http.Client
-	ctx     context.Context
 	quit    chan struct{}
 	once    sync.Once
 	entries chan entry
@@ -80,7 +79,6 @@ func NewClient(ctx context.Context, u *url.URL, username string, password string
 			url:       u,
 		},
 		client:  &http.Client{},
-		ctx:     ctx,
 		entries: make(chan entry),
 		quit:    make(chan struct{}),
 	}
@@ -117,16 +115,11 @@ func (c *Client) run() {
 	maxWaitCheck := time.NewTicker(maxWaitCheckFrequency)
 
 	defer func() {
-		// Send all pending batches
-		c.sendBatch(batch)
 		c.wg.Done()
 	}()
 
 	for {
 		select {
-		case <-c.ctx.Done():
-			return
-
 		case <-c.quit:
 			return
 
@@ -168,9 +161,11 @@ func (c *Client) sendBatch(batch *batch) {
 	var status int
 
 	for i := 0; i < 30; i++ {
-		// Try to send the message.
-		status, err = c.send(c.ctx, buf)
-		if err == nil {
+		ctx, cancel := context.WithTimeout(context.Background(), c.cfg.timeout)
+		status, err = c.send(ctx, buf)
+		cancel()
+
+		if err != nil {
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


lxd/loki/loki.go
tomponline, line 1: This approach looks like it isnt cancelling the go routine that is trying to send the data, but instead just returning from the Stop call after a certain time. Leaking a go routine or more in the process.

Instead we should plumb a context all the way through and cancel that when calling stop.

kadinsayani: Sounds good, I've added a cancellation context.

edlerd (approved):
LGTM, thanks for the quick fix on this one!

lxd/loki/loki.go
```diff
@@ -220,9 +265,21 @@ func (c *Client) send(ctx context.Context, buf []byte) (int, error) {
...
+	go func() {
+		c.wg.Wait()
+		close(done)
+	}()
+
+	select {
+	case <-done:
+		return nil
+	case <-time.After(c.cfg.timeout):
+		return fmt.Errorf("shutdown timed out")
```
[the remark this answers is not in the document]
kadinsayani: You're right, I see that now. With the addition of the `checkLoki()` function in this PR's second commit, it is no longer possible to setup a new Loki client with `loki.api.url` set to an invalid address. In the case when `loki.api.url` is changed from one valid address to another, we should wait for the wait group to finish so all logs are still sent to the old address, and in this case it won't block for an extended period of time since the address is valid and we won't attempt the maximum number of retries under normal conditions.

So I think we can drop the first commit and effectively close the issue, unless we see any other reason to refactor the `Stop()` function following discussion in our 1:1.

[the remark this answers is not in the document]
edlerd: > changed from one valid address to another, we should wait for the wait group to finish so all logs are still sent to the old address, and in this case it won't block for an extended period of time since the address is valid and we won't attempt the maximum number of retries under normal conditions.

I would doubt the assumption that loki is responsive, because it was reachable at the time of configuration in LXD.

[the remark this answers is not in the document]
edlerd: Services can become unreachable at random times in distributed systems.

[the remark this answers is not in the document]
tomponline: My view is that Stop() should immediately stop the routine sending messages to loki.

lxd/loki/loki.go
```diff
@@ -56,14 +56,16 @@ type Client struct {
...
 	quit    chan struct{}
 	once    sync.Once
 	entries chan entry
 	wg      sync.WaitGroup
 }
 
 // NewClient returns a Client.
-func NewClient(ctx context.Context, u *url.URL, username string, password string, caCert string, instance string, location string, logLevel string, labels []string, types []string) *Client {
+func NewClient(ctx context.Context, u *url.URL, username string, password string, caCert string, instance string, location string, logLevel string, labels []string, types []string) (*Client, error) {
+	ctx, cancel := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
tomponline: If this NewClient function is trying to connect to loki (i,e via the `checkLoki` function) then the `ctx` passed here should be passed down to `checkLoki` to allow the caller of `NewClient` to cancel and/or define a custom timeout for the check attempt.

One can create a chained context with a default timeout if necessary.

But the ctx to NewClient shouldn't be stored, as we already have the `Stop` function which is used for cancellation of an ongoing client.

We should only have method to stop/cancel an ongoing client.

This similarly came up with a discussion @MusicDin, @roosterfish and I had this morning.

lxd/loki/loki.go
```diff
@@ -99,12 +97,12 @@ func NewClient(ctx context.Context, u *url.URL, username string, password string
 	}
 
 	client.wg.Add(1)
-	go client.run()
+	go client.run(ctx)
```
tomponline, line 105: this still feels like a long-lived context to me, as NewClient will return and the context will live on inside the client.

kadinsayani: Instead, can we use the `Stop` function to close the `quit` channel and then create a short-lived context by applying a timeout to a supplied context before passing it into `sendBatch`?

tomponline: yeah sounds reasonable. Why is the a deferred sendBatch in run?

kadinsayani: The deferred `sendBatch` sends any last pending batches and decrements the wait group once the routine is cancelled. Since we want `Stop` to immediately stop the routine and sending messages we don't need the deferred `sendBatch`.

tomponline (approved):

tomponline:
@kadinsayani has this been tested with a loki server btw?

kadinsayani:
> @kadinsayani has this been tested with a loki server btw?

Yes 👍

PR #14869, opened 2025-01-27
Properly mark functions/fields as deprecated

simondeziel:
Also, remove deprecation markers where not applicable (still used internally).

### DIFF
--- a/lxd/firewall/drivers/drivers_xtables.go
+++ b/lxd/firewall/drivers/drivers_xtables.go
@@ -160,19 +160,19 @@ func (d Xtables) ebtablesInUse() bool {
 
 // networkIPTablesComment returns the iptables comment that is added to each network related rule.
 func (d Xtables) networkIPTablesComment(networkName string) string {
-	return fmt.Sprintf("LXD network %s", networkName)
+	return "LXD network " + networkName
 }
 
 // networkForwardIPTablesComment returns the iptables comment that is added to each network forward related rule.
 func (d Xtables) networkForwardIPTablesComment(networkName string) string {
-	return fmt.Sprintf("LXD network-forward %s", networkName)
+	return "LXD network-forward " + networkName
 }
 
 // networkSetupNICFilteringChain creates the NIC filtering chain if it doesn't exist, and adds the jump rules to
 // the INPUT and FORWARD filter chains. Must be called after networkSetupForwardingPolicy so that the rules are
 // prepended before the default fowarding policy rules.
 func (d Xtables) networkSetupNICFilteringChain(networkName string, ipVersion uint) error {
-	chain := fmt.Sprintf("%s_%s", iptablesChainNICFilterPrefix, networkName)
+	chain := iptablesChainNICFilterPrefix + "_" + networkName
 
 	// Create the NIC filter chain if it doesn't exist.
 	exists, _, err := d.iptablesChainExists(ipVersion, "filter", chain)
@@ -203,7 +203,7 @@ func (d Xtables) networkSetupNICFilteringChain(networkName string, ipVersion uin
 
 // networkSetupACLFilteringChains creates any missing ACL chains and adds jump rules.
 func (d Xtables) networkSetupACLFilteringChains(networkName string) error {
-	chain := fmt.Sprintf("%s_%s", iptablesChainACLFilterPrefix, networkName)
+	chain := iptablesChainACLFilterPrefix + "_" + networkName
 
 	for _, ipVersion := range []uint{4, 6} {
 		// Create the ACL filter chain if it doesn't exist.
@@ -528,7 +528,7 @@ func (d Xtables) NetworkSetup(networkName string, ipv4Address net.IP, ipv6Addres
 
 // NetworkApplyACLRules applies ACL rules to the existing firewall chains.
 func (d Xtables) NetworkApplyACLRules(networkName string, rules []ACLRule) error {
-	chain := fmt.Sprintf("%s_%s", iptablesChainACLFilterPrefix, networkName)
+	chain := iptablesChainACLFilterPrefix + "_" + networkName
 
 	// Parse rules for both IP families before applying either family of rules.
 	iptCmdRules := make(map[string][][]string)
@@ -684,9 +684,9 @@ func (d Xtables) aclRuleCriteriaToArgs(networkName string, ipVersion uint, rule 
 			args = append(args, "-m", extName)
 
 			if rule.ICMPCode == "" {
-				args = append(args, fmt.Sprintf("--%s-type", protoName), rule.ICMPType)
+				args = append(args, "--"+protoName+"-type", rule.ICMPType)
 			} else {
-				args = append(args, fmt.Sprintf("--%s-type", protoName), fmt.Sprintf("%s/%s", rule.ICMPType, rule.ICMPCode))
+				args = append(args, "--"+protoName+"-type", rule.ICMPType+"/"+rule.ICMPCode)
 			}
 		}
 	}
@@ -706,7 +706,7 @@ func (d Xtables) aclRuleCriteriaToArgs(networkName string, ipVersion uint, rule 
 
 		if rule.LogName != "" {
 			// Add a trailing space to prefix for readability in logs.
-			logArgs = append(logArgs, "--log-prefix", fmt.Sprintf("%s ", rule.LogName))
+			logArgs = append(logArgs, "--log-prefix", rule.LogName+" ")
 		}
 	}
 
@@ -742,7 +742,7 @@ func (d Xtables) aclRuleSubjectToACLMatch(direction string, ipVersion uint, subj
 	}
 
 	if len(fieldParts) > 0 {
-		return []string{fmt.Sprintf("--%s", direction), strings.Join(fieldParts, ",")}, nil
+		return []string{"--" + direction, strings.Join(fieldParts, ",")}, nil
 	}
 
 	return nil, nil // No subjects suitable for ipVersion.
@@ -756,13 +756,13 @@ func (d Xtables) aclRulePortToACLMatch(direction string, portCriteria ...string)
 	for _, portCriterion := range portCriteria {
 		criterionParts := strings.SplitN(portCriterion, "-", 2)
 		if len(criterionParts) > 1 {
-			fieldParts = append(fieldParts, fmt.Sprintf("%s:%s", criterionParts[0], criterionParts[1]))
+			fieldParts = append(fieldParts, criterionParts[0]+":"+criterionParts[1])
 		} else {
 			fieldParts = append(fieldParts, criterionParts[0])
 		}
 	}
 
-	return []string{"-m", "multiport", fmt.Sprintf("--%s", direction), strings.Join(fieldParts, ",")}
+	return []string{"-m", "multiport", "--" + direction, strings.Join(fieldParts, ",")}
 }
 
 // NetworkClear removes network rules from filter, mangle and nat tables.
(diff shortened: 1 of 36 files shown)


client/lxd_server.go
```diff
@@ -55,8 +55,6 @@ func (r *ProtocolLXD) UpdateServer(server api.ServerPut, ETag string) error {
 }
 
 // HasExtension returns true if the server supports a given API extension.
-//
-// Deprecated: Use CheckExtension instead.
```
tomponline, line 59: Its still deprecated, and places where its used need to be updated.

shared/util.go
```diff
@@ -1076,7 +1076,6 @@ func RunCommandContext(ctx context.Context, name string, arg ...string) (string,
 
 // RunCommand runs a command with optional arguments and returns stdout. If the command fails to
 // start or returns a non-zero exit code then an error is returned containing the output of stderr.
-// Deprecated: Use RunCommandContext.
```
tomponline, line 1079: Its still deprecated and places where its used should be updated.

lxd/storage/drivers/utils.go
```diff
@@ -739,7 +739,7 @@ func BTRFSSubVolumesGet(path string) ([]string, error) {
 
 // btrfsIsSubvolume checks if a given path is a subvolume.
 //
-// Deprecated: Use IsSubvolume from the Btrfs driver instead.
+// Deprecated: Use btrfs.isSubvolume().
 func btrfsIsSubVolume(subvolPath string) bool {
```
tomponline, line 743: can we remove this and update where its used to use btrfs.isSubVolume? Are the implementations different?

lxd/storage/drivers/utils.go
```diff
@@ -765,7 +765,9 @@ func BTRFSSubVolumeIsRo(path string) bool {
 	return strings.HasPrefix(string(output), "ro=true")
 }
 
-// BTRFSSubVolumeMakeRo makes a subvolume read only. Deprecated use btrfs.setSubvolumeReadonlyProperty().
+// BTRFSSubVolumeMakeRo makes a subvolume read only.
+//
+// Deprecated: Use btrfs.setSubvolumeReadonlyProperty().
```
tomponline, line 770: Can we remove this function and update the callers to use btrfs.setSubvolumeReadonlyProperty?

lxd/storage/drivers/utils.go
```diff
@@ -773,7 +773,9 @@ func BTRFSSubVolumeMakeRo(path string) error {
 	return err
 }
 
-// BTRFSSubVolumeMakeRw makes a sub volume read/write. Deprecated use btrfs.setSubvolumeReadonlyProperty().
+// BTRFSSubVolumeMakeRw makes a sub volume read/write.
+//
+// Deprecated: Use btrfs.setSubvolumeReadonlyProperty().
```
tomponline, line 778: Can we remove this function and update the callers to use btrfs.setSubvolumeReadonlyProperty

lxd/storage/storage.go
```diff
@@ -51,7 +51,8 @@ func InstanceImportingFilePath(instanceType instancetype.Type, poolName, project
 
 // GetStoragePoolMountPoint returns the mountpoint of the given pool.
 // {LXD_DIR}/storage-pools/<pool>
-// Deprecated, use GetPoolMountPath in storage/drivers package.
+//
+// Deprecated: Use GetPoolMountPath in storage/drivers package.
```
tomponline, line 55: Can we remove this function and update the callers to use GetPoolMountPath?

client/lxd_server.go
simondeziel: I see a need for both `HasExtension` and `CheckExtension` as the former just tells you if you have it or not and the later gives you either an error or nil.

All the places I've seen `HasExtension` being used was because a nicer/more in-context error message needed to be conveyed to the user.

shared/util.go
simondeziel: I think that's a lot of places were we'd just add a placeholder context, no?:
```
$ git grep -F 'shared.RunCommand(' | wc -l
254
```

tomponline: OK, but the function should still be marked as deprecated as we should avoid using it and guide people to using the other function that supports cancellation via context.

client/lxd_server.go
tomponline: I'd suggest we leave it deprecated to encourage a gradual switch over to the preferred CheckExtension.

lxd/storage/storage.go
simondeziel: There are no callers of `GetStoragePoolMountPoint()` but since it's exported we don't get notified about an unused function.

I'm not sure what to do in such cases. Is that OK to simply drop it since it's under `lxd/storage` which is not something other software should depend on?

tomponline: If its not used in microcluster or microcloud we can drop it as its not part of our public api

shared/util.go
simondeziel: Alright, I've properly marked it as deprecated which forced me to annotate a bunch of call sites with `// nolint: staticcheck` as I'm not sure how to build a proper context.

Also, is that cancellable context something that would allow me to `Ctrl-C` an operation I feel is too long? Like an image refresh for example?

lxd/storage/drivers/utils.go
simondeziel: I tried something, please have a look as I'm unsure that's the right way to go.

shared/util.go
tomponline: > Alright, I've properly marked it as deprecated which forced me to annotate a bunch of call sites with `// nolint: staticcheck` as I'm not sure how to build a proper context.
> 
> Also, is that cancellable context something that would allow me to `Ctrl-C` an operation I feel is too long? Like an image refresh for example?

You can pass context.TODO() and switch to non deprecated function if youve had to update old call sites anyway.

lxd/storage/drivers/utils.go
```diff
@@ -727,7 +727,8 @@ func BTRFSSubVolumesGet(path string) ([]string, error) {
 		}
 
 		// Check if a btrfs subvolume.
-		if btrfsIsSubVolume(fpath) {
+		d := &btrfs{}
```
tomponline, line 732: The wider issue here is that we have storage driver specific logic outside of the storage driver itself.

In this case the `BTRFSSubVolumesGet` is being called by `lxc` instance driver, we can see the issue in places like this:

https://github.com/canonical/lxd/blob/main/lxd/instance/drivers/driver_lxc.go#L1883-L1896

```go
	// Revert the currently applied on-disk idmap.
	if diskIdmap != nil {
		if storageType == "zfs" {
			err = diskIdmap.UnshiftRootfs(d.RootfsPath(), storageDrivers.ShiftZFSSkipper)
		} else if storageType == "btrfs" {
			err = storageDrivers.UnshiftBtrfsRootfs(d.RootfsPath(), diskIdmap)
		} else {
			err = diskIdmap.UnshiftRootfs(d.RootfsPath(), nil)
		}

		if err != nil {
			return idmap.IdmapStorageNone, nil, err
		}
	}
```

That function already has access to the pool driver via `pool`, so this should be pushed down into the storage driver itself, so we end up with something like this:

```
pool.UnshiftFs(d)
```

And then we can call the driver specific directly from the driver.

We can see 3 places in the lxc driver where the `d.getStorageType()` function value is used to get the pool driver of the instance's root disk and then perform storage driver specific logic based on that related to FS shifting.

Interestingly though (and counter to my suggestion above of pushing this down into the storage driver), I wonder if this logic actually *should be pool driver specific at all* and whether, like in the `qemu` driver's `addDriveConfig` where there is a filesystem specific logic for direct I/O that isn't related to the storage pool driver itself, it wouldn't be better to detect the container's root fs filesystem type.

The reason I ask is because it might be that we want to do filesystem specific shifting logic even when using a `dir` pool backed by a btrfs filesystem (for instance).

That being said, looking at the `shiftBtrfsRootfs` function, the only BTRFS specific thing about it is making the volume writable and then read only again, which is specific to the BTRFS storage driver logic of making snapshots read only.

So at the moment I'm thinking this would be best placed to push it down into the storage driver itself.

tomponline (changes requested):
All looks good, except please replace the ` // nolint: staticcheck` on `RunCommand` with the `RunCommandContext` call and use `context.TODO()` as the argument. This way we won't silence other potential linting issues on that line and will move away from the deprecated function. Ta

Don't worry about the storage driver push down piece I mentioned, it was really just an explanation of the wider issue I see there. But your approach to it works for now.

Thanks

simondeziel:
> All looks good, except please replace the ` // nolint: staticcheck` on `RunCommand` with the `RunCommandContext` call and use `context.TODO()` as the argument. This way we won't silence other potential linting issues on that line and will move away from the deprecated function. Ta

I've done it and replaced a few more `RunCommand` while in there. We go from 254 in `main` down to 183 with this PR, still a long way to go :/

> Don't worry about the storage driver push down piece I mentioned, it was really just an explanation of the wider issue I see there. But your approach to it works for now.

Ack, thanks :)

simondeziel:
@tomponline ready for another review when you have time. Thanks

shared/api/cluster.go
```diff
@@ -82,12 +82,6 @@ type ClusterPut struct {
 	// API extension: clustering_join
 	ServerAddress string `json:"server_address" yaml:"server_address"`
 
-	// The trust password of the cluster you're trying to join (deprecated, use cluster_token)
-	// Example: blah
-	//
-	// API extension: clustering_join
-	ClusterPassword string `json:"cluster_password" yaml:"cluster_password"` // Deprecated, use ClusterToken.
```
tomponline, line 89: @simondeziel so you didnt find any justifiable use of this field from an external client perspective, all former uses were intra-cluster only?

lxd/firewall/drivers/drivers_xtables.go
```diff
@@ -1332,7 +1332,7 @@ func (d Xtables) iptablesClear(ipVersion uint, comments []string, fromTables ...
 				fields[0] = "-D"
 
 				args = append(baseArgs, fields...)
-				_, err = shared.TryRunCommand("sh", "-c", cmd+" "+strings.Join(args, " "))
+				_, err := shared.TryRunCommand(cmd, args...)
```
tomponline, line 1335: This use of "sh" has been around since the very earliest iptables management commit in LXD 3.0.x

https://github.com/canonical/lxd/commit/5165b7f2dc2c2a214ce8e3c16c691dba198634e8#diff-fbd1acc2af1668d9e7a950499d2609e52057361c523285f7b868625661b2692dR71

What im concerned about is why is it there, seeing that all the other invocations even in the original commit didn't use the shell wrapper (suggesting it was done on purpose).

I too would like to get rid of this "sh" invocation, but  im concerned we may be introducing a regression with this change.

Possibly the output of the lines we are parsing can contain comments perhaps?

shared/api/cluster.go
simondeziel: > @simondeziel so you didnt find any justifiable use of this field from an external client perspective, all former uses were intra-cluster only?

In our 1:1, it seems you did the search for that field in `stable-5.21` which is why you had many hits. In `main`, that's the only place that appeared.

lxd/firewall/drivers/drivers_xtables.go
simondeziel: Noted, I dropped that commit hoping to get that PR merged before 6.3 is cut. I'll come again at it after doing some testing.

tomponline (approved):
LGTM

lxd/firewall/drivers/drivers_xtables.go
simondeziel: I just tested it and indeed, the `sh -c` wrapper is needed as removing it gives me this:

```
root@v1:~# lxc network set lxdbr0 security.acls=
Error: Failed clearing firewall: Failed to run: iptables -w -t filter -D INPUT -i lxdbr0 -p icmp -m icmp --icmp-type 12 -m comment --comment "generated for LXD network lxdbr0" -j ACCEPT: exit status 2 (iptables v1.8.10 (legacy): wait seconds not numeric
Try `iptables -h' or 'iptables --help' for more information.)
root@v1:~# iptables -w -t filter -D INPUT -i lxdbr0 -p icmp -m icmp --icmp-type 12 -m comment --comment "generated for LXD network lxdbr0" -j ACCEPT
root@v1:~# 
```

Considering that copy-n-pasting the command worked, I'm assuming the `--comment "generated for LXD network lxdbr0"` needs the `sh -c` wrapper to avoid being split up into multiple args like `--comment`, `"generated`, `for`, `LXD`, ...

PR #14871, opened 2025-01-28
OIDC: Use proxy enabled HTTP client for relying party remote key set

markylaing:
The remote key set will automatically fetch new keys when the current ones expire. If LXD is behind a proxy, this will fail unless the proxy enabled HTTP client is used.


tomponline:
Error: lxd/auth/oidc/oidc.go:512:151: redefines-builtin-id: redefinition of the built-in function delete (revive)
func (*Verifier) setCookies(w http.ResponseWriter, secureCookie *securecookie.SecureCookie, sessionID uuid.UUID, idToken string, refreshToken string, delete bool) error {

tomponline (approved):

PR #14873, opened 2025-01-28
Project: Avoid using empty context

hamistao:
Pre-requisite for https://github.com/canonical/lxd/pull/14867


tomponline (approved):
The PR description didn't explain why passing contexts is a prerequisite for https://github.com/canonical/lxd/pull/14867 so i don't understand why you needed to do this now.

However in general passing contexts to database querying functions is a good idea to allow prompt cancellation, so this LGTM thanks.

hamistao:
> The PR description didn't explain why passing contexts is a prerequisite for https://github.com/canonical/lxd/pull/14867 so i don't understand why you needed to do this now.

Just so that the newly introduced database queries can also use the correct context.

PR #14880, opened 2025-01-29
lxd: Entitlement enrichment for remaining API entities

gabrielmougard:
This should fix the issues discussed here: https://chat.canonical.com/canonical/pl/c3bjkis1g3nmbytao6q5gcaxuc

This also add entitlement enrichment for API entities that weren't added as part as https://github.com/canonical/lxd/pull/14748

### DIFF
--- a/doc/rest-api.yaml
+++ b/doc/rest-api.yaml
@@ -1106,6 +1106,15 @@ definitions:
     ImageAliasesEntry:
         description: ImageAliasesEntry represents a LXD image alias
         properties:
+            access_entitlements:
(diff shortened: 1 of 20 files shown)


tomponline:
@gabrielmougard as discussed please can you check whether we need to additional slices that convert the pointers back to concrete elements?

gabrielmougard:
@tomponline Now we should have support for all the entities supported by the OpenFGA permission model. I also ensured that we systematically use slice of pointers in the API handlers during the enrichment.

tomponline (approved):

tomponline:
@gabrielmougard just to confirm, did you test each endpoint is working correctly with-entitlements now?

PR #14882, opened 2025-01-29
Projects: Peform uplink quota validation in a single transaction

hamistao:
Moving the checks for uplink quotas to `AllowProjectUpdate` enables us to:
1- Be more aligned with our code standards;
2- Perform the check and update the quota values in a single transaction;
3- Be more efficient if setting many uplink quotas at once;


tomponline (approved):
Thanks!

PR #14886, opened 2025-01-30
Volume disk device sourcetype

MggMuggins:
I'm torn between the `volumeType/volumeName` and `sourcetype` syntaxes for specifying other types of storage volumes in disk devices. A few factors here, some I knew when I wrote my spec, but a few I didn't:
- By using `volumeType/volumeName`, we introduced novel syntax that the server doesn't use anywhere else, and doesn't fit nicely with the principle of "one value per map key".
- This syntax is already overloaded in the client to mean both `volumeType/volumeName` and `poolName/volumeName` (see [`lxc storage volume copy`](https://canonical-microcloud.readthedocs-hosted.com/en/latest/lxd/reference/manpages/lxc/storage/volume/copy/#lxc-storage-volume-copy-md))
- The `storage.backups_volume` and `storage.image_volume` [server configuration keys](https://canonical-microcloud.readthedocs-hosted.com/en/latest/lxd/server/#miscellaneous-options) also use `poolName/volumeName`

The current syntax for specifying volumes as disk devices is:
```
devices:
  cosmic-dust-analyzer-data:
	pool: default
	source: custom/cosmic-dust-analyzer-data
	type: disk
```

This PR uses this style instead:
```
devices:
  cosmic-dust-analyzer-data:
	pool: default
	source: cosmic-dust-analyzer-data
        sourcetype: custom
	type: disk
```

`sourcetype` is `custom` by default so it can be omitted for custom volumes.


tomponline:
> sourcetype: custom

I think `source-type` would be more readable here.

Would having a source-type of "instance" require a lot of extra DB lookups in order to figure out the instance type btw?

MggMuggins:
I went with `sourcetype` to be consistent with [`nictype`](https://canonical-microcloud.readthedocs-hosted.com/en/latest/lxd/reference/devices_nic/#nictype-vs-network) but I totally agree; will change.

Yes, If the volume type is not derivable from the disk device we'd need to look it up to know which `Mount{CustomVolume,CustomVolumeSnapshot,Instance,InstanceSnapshot}` function to call. Each of those functions also load the volume from the DB, so you'd end up doing basically the same query twice :face_with_diagonal_mouth:

tomponline:
> I went with `sourcetype` to be consistent with [`nictype`](https://canonical-microcloud.readthedocs-hosted.com/en/latest/lxd/reference/devices_nic/#nictype-vs-network) but I totally agree; will change.
> 
> Yes, If the volume type is not derivable from the disk device we'd need to look it up to know which `Mount{CustomVolume,CustomVolumeSnapshot,Instance,InstanceSnapshot}` function to call. Each of those functions also load the volume from the DB, so you'd end up doing basically the same query twice 🫤

Looks like we are doing this already anyway?

https://github.com/canonical/lxd/pull/14886/files#diff-30511c7ce3bcabd6a4ddc8624541cc2f5240a13b5e3b19aac8c787fdbd0e4720R1138

MggMuggins:
Looks like a second query isn't done for 50% of the times we parse that field. [1](https://github.com/canonical/lxd/pull/14886/files#diff-30511c7ce3bcabd6a4ddc8624541cc2f5240a13b5e3b19aac8c787fdbd0e4720R753) [2](https://github.com/canonical/lxd/pull/14886/files#diff-30511c7ce3bcabd6a4ddc8624541cc2f5240a13b5e3b19aac8c787fdbd0e4720L1615) [3](https://github.com/canonical/lxd/pull/14886/files#diff-30511c7ce3bcabd6a4ddc8624541cc2f5240a13b5e3b19aac8c787fdbd0e4720R2132)

I think we've discussed this before, but I don't love `instance` from a consistency point of view (`lxc storage volume ls <pool>` `TYPE` column). I'm willing to implement it that way but I think it muddies the waters between the `source-type` field and the `volumeType/volumeName` syntax that's still being used in `lxc`.

tomponline:
> Looks like a second query isn't done for 50% of the times we parse that field. [1](https://github.com/canonical/lxd/pull/14886/files#diff-30511c7ce3bcabd6a4ddc8624541cc2f5240a13b5e3b19aac8c787fdbd0e4720R753) [2](https://github.com/canonical/lxd/pull/14886/files#diff-30511c7ce3bcabd6a4ddc8624541cc2f5240a13b5e3b19aac8c787fdbd0e4720L1615) [3](https://github.com/canonical/lxd/pull/14886/files#diff-30511c7ce3bcabd6a4ddc8624541cc2f5240a13b5e3b19aac8c787fdbd0e4720R2132)
> 
> I think we've discussed this before, but I don't love `instance` from a consistency point of view (`lxc storage volume ls <pool>` `TYPE` column). I'm willing to implement it that way but I think it muddies the waters between the `source-type` field and the `volumeType/volumeName` syntax that's still being used in `lxc`.
> 
> I was able to cut out an extra query for each usage by using the already-loaded project.

OK makes sense, lets use the existing "container" and "virtual-machine" prefixes then. Ta

lxd/device/disk.go
```diff
@@ -74,7 +74,7 @@ type diskSourceNotFoundError struct {
 }
 
 func (e diskSourceNotFoundError) Error() string {
-	return fmt.Sprint(e.msg, ": ", e.err)
+	return fmt.Sprintf("%s: %v", e.msg, e.err)
```
tomponline, line 77: This was changed just 5 days ago by @simondeziel in https://github.com/canonical/lxd/commit/fcf892a6d8801a53a972d0a241433cf68be1b313 to avoid using sprintf that uses reflection.

lxd/device/disk.go
```diff
@@ -1215,7 +1249,7 @@ func (d *disk) startVM() (*deviceConfig.RunConfig, error) {
 				// virtfs-proxy-helper 9p share. The 9p share will only be used as a fallback.
 				err = func() error {
 					sockPath, pidPath := d.vmVirtiofsdPaths()
-					logPath := filepath.Join(d.inst.LogPath(), "disk."+filesystem.PathNameEncode(d.name)+".log")
+					logPath := filepath.Join(d.inst.LogPath(), fmt.Sprintf("disk.%s.log", filesystem.PathNameEncode(d.name)))
```
tomponline, line 1244: @MggMuggins why switching back to Sprintf when @simondeziel changed this recently, and also in an unrelated commit?

tomponline (changes requested):
I think a rebase is required as it looks like you've inadvertently changed a bunch of unrelated sprintf removals back.

tomponline (approved):
LGTM!

PR #14888, opened 2025-01-30
OIDC: Add additional scopes

### LINKED ISSUE #14141: OIDC - oidc.groups.claim is added as additional scope which breaks Microsoft Entra integration

markylaing:
Adds an `oidc.additional_scopes` configuration key which defaults to `profile,offline_access`.
Removes the `oidc.groups.claim` value from requested scopes.

Things I'm not sure on:
- Using a comma to delimit the scope values. Identity providers appear to be quite freeform. It might be better to accept this as a space delimited as in #14832 (as the author originally suggested) to avoid issues with splitting on the comma.
- Backwards and forwards compatibility. New clients talking to old servers will not request the configured scopes. Old clients talking to new servers will continue to request the groups claim, but will continue to request the `profile` and `offline_access` scopes. 

Requires #14871 
Closes #14141

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2599,3 +2599,9 @@ The following pool level configuration keys have been added:
 1. {config:option}`storage-pure-pool-conf:pure.api.token`
 1. {config:option}`storage-pure-pool-conf:pure.mode`
 1. {config:option}`storage-pure-pool-conf:pure.target`
+
+## `oidc_additional_scopes`
+
+This API extension enables setting an {config:option}`server-oidc:oidc.additional_scopes` configuration key.
+This configuration option can be used to request additional scopes that might be required for retrieving {ref}`identity provider groups <identity-provider-groups>` from the identity provider.
+Additionally, the optional scopes `profile` and `offline_access` can be unset via this setting.
--- a/lxd/auth/oidc/oidc.go
+++ b/lxd/auth/oidc/oidc.go
@@ -51,7 +51,7 @@ type Verifier struct {
 	clientID       string
 	issuer         string
 	audience       string
-	groupsClaim    string
+	opts           Opts
 	clusterCert    func() *shared.CertInfo
 	httpClientFunc func() (*http.Client, error)
 
@@ -273,26 +273,26 @@ func (o *Verifier) getEmailFromClaims(claims map[string]any) (string, error) {
 // getGroupsFromClaims attempts to get the configured groups claim from the token claims and warns if it is not present
 // or is not a valid type. The custom claims are an unmarshalled JSON object.
 func (o *Verifier) getGroupsFromClaims(customClaims map[string]any) []string {
-	if o.groupsClaim == "" {
+	if o.opts.GroupsClaim == "" {
 		return nil
 	}
 
-	groupsClaimAny, ok := customClaims[o.groupsClaim]
+	groupsClaimAny, ok := customClaims[o.opts.GroupsClaim]
 	if !ok {
-		logger.Warn("OIDC groups custom claim not found", logger.Ctx{"claim_name": o.groupsClaim})
+		logger.Warn("OIDC groups custom claim not found", logger.Ctx{"claim_name": o.opts.GroupsClaim})
 		return nil
 	}
 
 	groupsArr, ok := groupsClaimAny.([]any)
 	if !ok {
-		logger.Warn("Unexpected type for OIDC groups custom claim", logger.Ctx{"claim_name": o.groupsClaim, "claim_value": groupsClaimAny})
+		logger.Warn("Unexpected type for OIDC groups custom claim", logger.Ctx{"claim_name": o.opts.GroupsClaim, "claim_value": groupsClaimAny})
 	}
 
 	groups := make([]string, 0, len(groupsArr))
 	for _, groupNameAny := range groupsArr {
 		groupName, ok := groupNameAny.(string)
 		if !ok {
-			logger.Warn("Unexpected type for OIDC groups custom claim", logger.Ctx{"claim_name": o.groupsClaim, "claim_value": groupsClaimAny})
+			logger.Warn("Unexpected type for OIDC groups custom claim", logger.Ctx{"claim_name": o.opts.GroupsClaim, "claim_value": groupsClaimAny})
 		}
 
 		groups = append(groups, groupName)
@@ -359,7 +359,7 @@ func (o *Verifier) WriteHeaders(w http.ResponseWriter) error {
 	w.Header().Set("X-LXD-OIDC-issuer", o.issuer)
 	w.Header().Set("X-LXD-OIDC-clientid", o.clientID)
 	w.Header().Set("X-LXD-OIDC-audience", o.audience)
-	w.Header().Set("X-LXD-OIDC-groups-claim", o.groupsClaim)
+	w.Header().Set("X-LXD-OIDC-groups-claim", o.opts.GroupsClaim)
 
 	return nil
 }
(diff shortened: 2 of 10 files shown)


tomponline:
Can be rebased now to remove linter commit

doc/api-extensions.md
```diff
@@ -2553,6 +2553,7 @@ Adds support for instance root volumes to be attached to other instances as disk
 devices. Introduces the `<type>/<volume>` syntax for the `source` property of
 disk devices.
 
+<<<<<<< HEAD
```
minaelee, line 2556:
```suggestion
```

doc/api-extensions.md
```diff
@@ -2564,3 +2565,9 @@ Adds `fine_grained` field to `GET /1.0/auth/identities/current` to indicate if t
 interacting with the LXD API is fine-grained (that is, associated permissions are managed via group membership).
 Allows LXD entities to be returned with an `access_entitlements` field if the current identity is fine-grained and the
 GET request to fetch the LXD entities has the `with-access-entitlements=<comma_separated_list_of_candidate_entitlements>` query parameter.
+
+## `oidc_additional_scopes`
+
+This API extension enables setting an {config:option}`server-oidc:oidc.additional_scopes` configuration key.
+This configuration option can be used to request additional scopes that might be required for retrieving {ref}`identity-provider-groups` from the identity provider.
```
minaelee, line 2572:
```suggestion
This configuration option can be used to request additional scopes that might be required for retrieving {ref}`identity provider groups <identity-provider-groups>` from the identity provider.
```

minaelee: otherwise the sentence reads like this:
>This configuration option can be used to request additional scopes that might be required for retrieving [Use groups defined by the identity provider](https://canonical-ubuntu-documentation-library--14888.com.readthedocs.build/lxd/en/14888/explanation/authorization/#identity-provider-groups) from the identity provider.

minaelee (changes requested):

lxd/auth/oidc/oidc.go
```diff
@@ -153,6 +153,13 @@ func (o *Verifier) authenticateAccessToken(ctx context.Context, accessToken stri
 		return nil, AuthError{Err: fmt.Errorf("Provided OIDC token doesn't allow the configured audience")}
 	}
 
+	// If we already have all the required information, return now.
+	res, err := o.getResultFromClaims(claims, claims.Claims)
+	if err == nil {
+		return res, nil
+	}
+
```
BarcoMasile, line 161: in case this `err != nil` the error get swallowed. It could be useful to log it (even at debug)

doc/api-extensions.md
markylaing: Ah this is a rebase artefact. Thanks for catching.

lxd/auth/oidc/oidc.go
```diff
@@ -335,6 +336,17 @@ func (o *Verifier) WriteHeaders(w http.ResponseWriter) error {
 	w.Header().Set("X-LXD-OIDC-issuer", o.issuer)
 	w.Header().Set("X-LXD-OIDC-clientid", o.clientID)
 	w.Header().Set("X-LXD-OIDC-audience", o.audience)
+
+	if len(o.opts.AdditionalScopes) > 0 {
```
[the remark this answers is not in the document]
markylaing: `o.opts` is not nullable. `o.opts.AdditionalScopes` may be nil but there is no need to nil check because the type is `[]string` and `len` of a nil slice is zero.

lxd/cluster/config/config.go
```diff
@@ -224,8 +224,8 @@ func (c *Config) RemoteTokenExpiry() string {
 }
 
 // OIDCServer returns all the OpenID Connect settings needed to connect to a server.
-func (c *Config) OIDCServer() (issuer string, clientID string, audience string, groupsClaim string) {
-	return c.m.GetString("oidc.issuer"), c.m.GetString("oidc.client.id"), c.m.GetString("oidc.audience"), c.m.GetString("oidc.groups.claim")
+func (c *Config) OIDCServer() (issuer string, clientID string, audience string, groupsClaim string, additionalScopes []string) {
+	return c.m.GetString("oidc.issuer"), c.m.GetString("oidc.client.id"), c.m.GetString("oidc.audience"), c.m.GetString("oidc.groups.claim"), strings.Split(c.m.GetString("oidc.additional_scopes"), ",")
```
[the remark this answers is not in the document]
markylaing: Yeah I will do this, thanks.

lxd/auth/oidc/oidc.go
markylaing: I've dropped this commit as it is not necessary for the purpose of this PR.

markylaing:
@tomponline this is ready for review

lxd/cluster/config/config.go
```diff
@@ -680,15 +681,29 @@ var ConfigSchema = config.Schema{
...
+	// The value of the claim is expected to be a JSON string array.
 	// ---
 	//  type: string
 	//  scope: global
 	//  shortdesc: A claim used for mapping identity provider groups to LXD groups.
 	"oidc.groups.claim": {},
+
+	// lxdmeta:generate(entities=server; group=oidc; key=oidc.additional_scopes)
+	// Specify additional scopes to be requested when performing OIDC flows.
+	// Additional scopes might be required if managing access control in the Identity Provider (see {ref}`identity-provider-groups`).
```
minaelee, line 695:
```suggestion
	// Additional scopes might be required if managing access control in the Identity Provider (see: {ref}`identity-provider-groups`).
```

lxd/cluster/config/config.go
```diff
@@ -680,15 +681,29 @@ var ConfigSchema = config.Schema{
...
+	// The default value for this configuration option contains two scopes which are optional, but will degrade LXD OIDC functionality if removed.
+	// These are:
+	//   - `profile`: This is requested by LXD to present display names for the identification of end users. If it is removed, only email addresses will be displayed.
```
minaelee, line 698:
```suggestion
	//   - `profile`: This is requested by LXD to present display names for the identification of end users. If it is removed, only email addresses are displayed.
```

markylaing:
Discussed with Tom that this needs extra work. If both `profile` and `offline_access` scopes are removed, LXD will continue to use them as the default values.

tomponline:

markylaing:
> @markylaing shall we close this for now?

I don't think so. Making this configurable is useful for @ech0-de (for Gitlab OIDC) and as far as I know the only contentious thing is how to unset both `profile` and `offline_access` scopes.

How about we introduce two keys?
- `oidc.scopes.additional`: No default. Can be used to set additional scopes if required for retrieving group info, or if the OIDC provider requires any additional scopes (I've seen Entra ID use a `user.read` scope for example). Validation logic will not allow setting an additional scope that is present in `oidc.scopes.omit`.
- `oidc.scopes.omit`: No default. Can be used to omit the `profile` and/or `offline_access` scopes. Validation logic will prevent setting `openid`, `email`, or any scopes present in `oidc.scopes.additional`.

That said, I guess another point of contention I have with our OIDC implementation is the use of headers to send e.g. the issuer URL to the LXD client. I think these values should be returned via `/1.0` and this PR adds yet another header.

markylaing:
> > @markylaing shall we close this for now?
> 
> I don't think so. Making this configurable is useful for @ech0-de (for Gitlab OIDC) and as far as I know the only contentious thing is how to unset both `profile` and `offline_access` scopes.
> 
> How about we introduce two keys?
> 
> * `oidc.scopes.additional`: No default. Can be used to set additional scopes if required for retrieving group info, or if the OIDC provider requires any additional scopes (I've seen Entra ID use a `user.read` scope for example). Validation logic will not allow setting an additional scope that is present in `oidc.scopes.omit`.
> * `oidc.scopes.omit`: No default. Can be used to omit the `profile` and/or `offline_access` scopes. Validation logic will prevent setting `openid`, `email`, or any scopes present in `oidc.scopes.additional`.
> 
> That said, I guess another point of contention I have with our OIDC implementation is the use of headers to send e.g. the issuer URL to the LXD client. I think these values should be returned via `/1.0` and this PR adds yet another header.

I don't like this idea either. We can just have `oidc.scopes` and validation that requires `openid` and `email` be included.

markylaing:
Closing in favour of #14935.

PR #14889, opened 2025-01-30
OIDC: Add Entra ID support

markylaing:
[Entra ID](https://www.microsoft.com/en-gb/security/business/identity-access/microsoft-entra-id) is a cloud IAM offering from Microsoft. It is based on Azure Active Directory and as such, has a few idiosyncrasies that are not part of the OIDC standard.

When adding a LXD server as a remote in the LXD CLI, we use the [Device Authorization Grant](https://oauth.net/2/device-flow/) flow. We then send the access token retrieved by that flow to LXD as an `Authorization: Bearer <token>` header. LXD verifies the token and then calls the `/userinfo` endpoint to retrieve the user's name and email address to associate them with the IdP subject.

However, LXD can't verify the access token for 2 reasons:
1. The access token issuer `https://sts.windows.net/<tenant_id>` is different from the configured issuer `https://login.microsoftonline.com/<tenant_id>/v2.0` from which OIDC discovery was initially performed.
2. The access token contains a `nonce` key in the header which needs to be hashed and recombined into the token before validation (see https://medium.com/@codeatom/manual-validate-azure-ad-graph-api-access-token-35f1d451b210 and https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/609#issuecomment-405683736).

The reason for this is that the access token is not meant for our API. It is intended for use with the Microsoft Graph API only. 

It is possible to force Entra ID to return an access token containing the correct issuer and no `nonce` value in the header by adding an "API" to the application configuration in Entra, and sending it's associated scope when performing the device flow (see https://learn.microsoft.com/en-us/answers/questions/1669290/oauth-jwt-issuer-(sts-windows-net)-is-different-fr). However, the returned token is then no longer valid for use with the microsoft graph API which is where the `/userinfo` endpoint is located. Additionally, if access tokens are configured in Entra ID to contain any optional claims such as `email`, then the token will revert back to the `sts.windows.net` issuer.

To summarise, we can't verify access tokens that are required for retrieving compulsory user information. This leaves our options pretty limited:
1. Don't verify access tokens locally. Instead, just call `/userinfo` with the provided token and if it's valid, we'll get the required user information.
2. Luckily, Entra ID returns an ID token as part of the Device Flow which contains an email address (if configured) has the correct issuer, and can be verified. We can send this from the client instead in certain scenarios.

I discounted option 1 because this means an outbound request to Entra ID *for every API call to LXD*. Additionally, the `sub` claim returned by the `/userinfo` call *is different to the one in the access token*. In fact, both the ID token and the `/userinfo` endpoint contain/return the same subject, but the access token as a different one. This is absolutely not part of the OIDC spec.

So finally we land on Option 2. This pull request sets the ID token as the bearer token in the following scenario:
1. The ID Token is not empty, and
2. The Access token contains a `nonce` value in the header, and
3. The `iss` claim in the access token and ID token are different, and
4. The `sub` claim in the access token and ID token are different, and
5. The ID token contains an `email` claim.

On the server side, we now initially expect the bearer token to be an ID token, falling back to an access token. This is because ID token verification is more strict than access token verification (it checks more fields, e.g. it checks that the audience contains the client ID).

To configure Entra ID authentication in LXD. These are the steps:
1. *EntraID*: Create an application:
2. *EntraID*: {application} > Authentication > + Add a platform > Mobile and desktop applications > Custom redirect URLs > Enter the following values and save:
    - `https://{LXD host}/oidc/callback`
    - `https://{LXD host}/oidc/login`
    - `https://{LXD host}/oidc/logout`
3. *EntraID*: {application} > API permissions > + Add a permission > Microsoft Graph > Delegated permissions > OpenId permissions > Check `email`, `offline_access`, `openid`, `profile` and save.
4. *EntraID*: {application} > Token configuration > + Add optional claim > ID > Check `email` and save.
5. *EntraID*: {application} > Authentication > Advanced settings > Set "Allow public client flows" to "Yes" and save.
6. *LXD*: `lxc config set oidc.client.id = {Application (client) ID found in Entra ID application overview}`.
7. *LXD*: `lxc config set oidc.audience = {Application (client) ID found in Entra ID application overview}`.
8. *LXD*: `lxc config set oidc.issuer = https://login.microsoftonline.com/{Directory (tenant) ID found in Entra ID application overview}/v2.0`.


markylaing:
@tomponline this PR didn't really depend on the other ones so I've removed those commits. This PR can be reviewed first. There's a lint fix commit I'll just have to remove from the other two when this is merged.

markylaing:
Should be fixed now. I was missing [this commit](https://github.com/canonical/lxd/pull/14888/commits/aebf6b2b3cef5e2ced8804774d24f7555c2f84c3)

tomponline:
Can be rebased now to remove linter commit

nsklikas (commented):
Tried to write down some thoughts about this, interested to see what other people think.

Even though EntraID is not OIDC complaint, unfortunately `access_tokens` can be anything the Provider wants.  
To add to that, the [JWT access token rfc](https://datatracker.ietf.org/doc/html/rfc9068#section-2.2) states:
> sub
    REQUIRED - as defined in [Section 4.1.2](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.2) of [[RFC7519](https://datatracker.ietf.org/doc/html/rfc7519)]. In cases of access tokens obtained through grants where a resource owner is involved, such as the authorization code grant, the value of "sub" SHOULD correspond to the subject identifier of the resource owner. In cases of access tokens obtained through grants where no resource owner is involved, such as the client credentials grant, the value of "sub" SHOULD correspond to an identifier the authorization server uses to indicate the client application. See [Section 5](https://datatracker.ietf.org/doc/html/rfc9068#SecurityConsiderations) for more details on this scenario. Also, see [Section 6](https://datatracker.ietf.org/doc/html/rfc9068#PrivacyConsiderations) for a discussion about how different choices in assigning "sub" values can impact privacy.

Meaning that the `sub` does not have to match the one in the idtoken or userinfo.

[Access tokens](https://datatracker.ietf.org/doc/html/rfc6749#section-1.4) are:
>    credentials used to access protected resources.  An
   access token is a string representing an authorization issued to the
   client.  The string is usually opaque to the client.  Tokens
   represent specific scopes and durations of access, granted by the
   resource owner, and enforced by the resource server and authorization
   server.

[ID tokens](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) are:
> security tokens that contain Claims about the Authentication of an End-User by an Authorization Server when using a Client, and potentially other requested Claims

After reading the above descriptions, IMO it makes sense to use id tokens to keep track of user sessions (just like you do with the browser cookies) and it makes sense to use access tokens to authenticate external entities (CLI) accessing the resource's (LXD) API.

Unfortunately, I agree that it is probably unwanted to make a call to EntraID on every API call (+ if the connection is broken or if EntraID is down, the LXD API will be inaccessible from the CLI).

One problem with the current approach is that AFAIK you will not be able to get the user's groups from the ID token, as that requires calling the graph API using the access token (https://learn.microsoft.com/en-us/graph/api/user-list-memberof). Also the ID token does not contain the requested scopes, which is something that you may want to validate (although not a requirement, especially for Entra).

Another thing to consider is that other public clouds, may not properly implement the protocol either (tbh I'm not sure if anyone else is as bad as microsoft) OR they may also require extra calls to fetch the groups. To adjust to such idiosyncrasies I think that introducing a `provider` configuration variable is needed, this way you can have custom logic for different providers.

~IF you don't care for supporting EntraID groups, then I think that using access tokens issued by https://sts.windows.net/... is fine, as they can be configured to have the email and the correct audience.~
(EDIT: I now realize that you are using userinfo to fetch the user's attributes, which makes sense, so using this approach won't work).

### Microsoft specific solution
I think that my preferred solution for this would be to concat the access token + id token and send them in the header. If the user is in the cache, then you only validate the id token, else you use the access token to fetch the user's groups. The access token hash could be stored in the cache alongside the userinfo to allow you to validate it. This will allow you to validate access tokens whose signature is invalid, while not calling the `userinfo` endpoint on every request.

### Alternatively
IMO, if you wish to support public clouds + fetching group information + not making calls to external APIs on every API request, you would need to mint your own tokens (like you do for the UI). The CLI would be exchanging the access token from the external OIDC Provider, with an LXD signed token and then it would be using that to interact with the lxd API.

If you wish to support public clouds + fetching group information, but you are fine with making a call to the external API. Then using the access token as you do now, is fine. A new lxd config variable could be introduced, specifying whether the access token should be validated using it's signature or an external API (this logic could be included in the `provider` config I mentioned above).

If you only care about using public clouds to authenticate users and you don't care about fetching the user's groups, then the current approach is fine BUT I would suggest to continue using access_tokens and to expect that if the user uses Entra to properly configure the application (with a custom scope and to include `email` and `aud` claims)

BarcoMasile (changes requested):
Requesting changes for the comment about the failing ID token check that leads to an extra check on the Access Token.

About the approach in general: I agree with everything Nikos said, although I would personally prefer instead of a provider configuration variable, a different implementation based on the supported providers (the strategy pattern I mentioned in our call from this week).

That said, I understand that MS Entra ID has a very weird behavior, but probably the non-matching issuer thing could be bypassed by configuring the library (some library allow to disable the `iss` check for example) and adding a "manual" check on the expected value, plus applying the nonce spec by MS Entra to correctly verify the access token. This way LXD could stay in what is considered the best approach for validating access (as Nikos beautifully explained in his comment).

wood-push-melon:
Just a NON-BLOCKING comment:

I lack much context and don't know much about lxd implementation. I can understand the pitfalls of MS Entra ID and understand the initiative of the PR (although the workaround solution diverges from the intended use of ID tokens).

From the pure perspective of code implementation, I strongly agree with @BarcoMasile's recommendation to use a certain pattern for such an MS-specific provider solution. This follows the open-closed principle and is better for maintainable development and testing.

PR #14890, opened 2025-01-30
doc(instance) fix cluster.evacuate doc block to avoid newlines in the generated metadata api response

edlerd:
# Done
fix cluster.evacuate doc block to avoid newlines in the generated metadata api response


tomponline (approved):
Ta

PR #14892, opened 2025-01-30
doc: increase linkcheck max rate limit timeout and retries

minaelee:
Increasing the max rate limit timeout should help with linkcheck issues due to rate limits. 
Increasing the number of retries might also help with this & other linkcheck issues.


tomponline:
@minaelee lets try this, but I suspect this may exacerbate the situation by retrying a system that is already rejecting the doc builder's requests.

Instead I'd like to see us A) try less concurrency if possible B) move away from requesting external links in the doc build process entirely and instead embed the page titles directly.

tomponline (approved):

Issue #14560, opened 2024-11-29
Improvements to shell completions for `lxc network|storage|config get|set|unset`


kadinsayani:
- `lxc network|storage set|get` should show configurable options;
- `lxc network|storage unset` should show options that can be unset;
- `lxc config set|get` should show configurable options based on `scope`;
- `lxc config unset` should show options that can be unset based on `scope`.

Completion functions should use the [new `scope` field in the metadata configuration API](https://github.com/canonical/lxd/pull/14530) to contextually source applicable options.

kadinsayani:
From the LXD docs:

> When you configure an option with global scope on any cluster member, the changes are propagated to the other cluster members through the distributed database. The locally scoped options are set only on the cluster member where you configure them, unless you use the --target flag to specify a different cluster member.

Therefore, we don't want to contextually provide completions for server configuration options based on cluster member roles.

Furthermore:

> In addition to the server configuration, there are [cluster member configuration options](https://documentation.ubuntu.com/lxd/en/latest/reference/cluster_member_config/#cluster-member-config) that are specific to each cluster member.

We already have completions for getting and unsetting cluster member configuration options by way of the [`cmpClusterMemberConfigs`](https://github.com/canonical/lxd/blob/lxd-remotes/lxc/completion.go#L92-L92) function, so we just need to add completions for setting cluster member configuration options.

PR #14894, opened 2025-01-30
CLI: Cluster member config key completions

kadinsayani:
Closes https://github.com/canonical/lxd/issues/14560.

### DIFF
--- a/lxc/completion.go
+++ b/lxc/completion.go
@@ -87,6 +87,54 @@ func (g *cmdGlobal) cmpClusterGroups(toComplete string) ([]string, cobra.ShellCo
 	return results, cmpDirectives
 }
 
+// cmpClusterMemberAllConfigKeys provides shell completion for all cluster member configuration keys.
+// It takes a partial input string and returns a list of all cluster member configuration keys along with a shell completion directive.
+func (g *cmdGlobal) cmpClusterMemberAllConfigKeys(memberName string) ([]string, cobra.ShellCompDirective) {
+	cmpDirectives := cobra.ShellCompDirectiveNoFileComp | cobra.ShellCompDirectiveNoSpace
+
+	// Parse remote
+	resources, err := g.ParseServers(memberName)
+	if err != nil || len(resources) == 0 {
+		return nil, cobra.ShellCompDirectiveError
+	}
+
+	resource := resources[0]
+	client := resource.server
+
+	cluster, _, err := client.GetCluster()
+	if err != nil || !cluster.Enabled {
+		return nil, cobra.ShellCompDirectiveError
+	}
+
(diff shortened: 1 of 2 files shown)


lxc/completion.go
```diff
@@ -87,6 +87,54 @@ func (g *cmdGlobal) cmpClusterGroups(toComplete string) ([]string, cobra.ShellCo
...
+
+	for _, field := range clusterConfig {
+		for _, key := range field.Keys {
+			for configKey := range key {
+				configKeys = append(configKeys, configKey)
+			}
+		}
+	}
+
+	return configKeys, cmpDirectives | cobra.ShellCompDirectiveNoSpace
```
simondeziel, line 135: Could that `cobra.ShellCompDirectiveNoSpace` be done at the beginning of the function?

lxc/completion.go
```diff
@@ -87,6 +87,54 @@ func (g *cmdGlobal) cmpClusterGroups(toComplete string) ([]string, cobra.ShellCo
 	return results, cmpDirectives
 }
 
+// cmpClusterMemberAllConfigKeys provides shell completion for all cluster member configuration keys.
+// It takes a partial input string and returns a list of all cluster member configuration keys along with a shell completion directive.
+func (g *cmdGlobal) cmpClusterMemberAllConfigKeys(memberName string) ([]string, cobra.ShellCompDirective) {
+	cmpDirectives := cobra.ShellCompDirectiveNoSpace
```
simondeziel, line 93: We lost the no file comp directive when you moved the no space one here. Not sure if that was intentional or not.

kadinsayani: That was unintentional, thanks for pointing it out!

lxc/cluster_role.go
```diff
@@ -58,6 +58,10 @@ func (c *cmdClusterRoleAdd) command() *cobra.Command {
 			return c.global.cmpClusterMembers(toComplete)
 		}
 
+		if len(args) == 1 {
+			return []string{"event-hub", "ovn-chassis"}, cobra.ShellCompDirectiveNoSpace
```
tomponline, line 62: @kadinsayani lets drop this as we dont want to hard code possible cluster member roles into the CLI - how will we remember to update these later when we add new ones (similar point as I've raised before).

Either we get the info from the metadata or we use a compile time shared list of valid cluster member roles that are referenced both in lxc and on the server.

Please don't add these types of hard coded list in the CLI itself.

Thanks

kadinsayani: I've dropped the commit adding cluster member role completions.

simondeziel (approved):

tomponline (approved):

PR #14895, opened 2025-01-30
Storage: Disk device doc updates

MggMuggins:
Documentation and comment changes that I made while working on container root attachment. Since that item is on hold I thought I'd at least get these improvements put up.

Hopefully this improves clarity around IDmaps by a small measure.


tomponline (approved):
Thanks!

Issue #13677, opened 2024-06-28
Mount flags for VM directory share disks are not passed to lxd-agent when hotplugged


tomponline:
The resulting mount flags  (such as `ro`) used for directory pass-through disks are different depending on whether the disk is hotplugged or available at start time.

```
lxc init ubuntu-minimal:24.04 v1 -d root,size=5GiB --vm

# Add disk device before VM is started.
lxc config device add v1 foo disk source=/foo path=/mnt readonly=true
lxc start v1
lxc exec v1 -- mount | grep foo
lxd_foo on /mnt type virtiofs (ro,relatime) # `ro` flag is added due to `readonly=true` in device config.

# Add disk whilst VM is running (hotplug).
lxc config device remove v1 foo
lxc exec v1 -- umount /mnt
lxc config device add v1 foo disk source=/foo path=/mnt readonly=true
lxc exec v1 -- mount | grep foo
lxd_foo on /mnt type virtiofs (rw,relatime) # `ro` flag is *not* added even though `readonly=true` in device config.
```

**This is not a security issue because LXD still enforces the read-only nature of the disk on the host side:**

```
lxc exec v1 -- touch /mnt/foo
touch: cannot touch '/mnt/foo': Read-only file system
```

But it would be nice to be consistent, and in the future we may want to support other mount flags that will need to be passed both at start time and hotplug time.

PR #14898, opened 2025-01-31
VM: Apply mount options to hotplugged disks

hamistao:
Closes https://github.com/canonical/lxd/issues/13677

Related CI tests in: https://github.com/canonical/lxd-ci/pull/403

### DIFF
--- a/lxd-agent/events.go
+++ b/lxd-agent/events.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -161,8 +162,16 @@ func eventsProcess(event api.Event) {
 
 	_ = os.MkdirAll(e.Config["path"], 0755)
 
+	// Parse mount options, if provided.
+	var args []string
+	if len(e.Mount.Options) > 0 {
+		args = append(args, "-o", strings.Join(e.Mount.Options, ","))
+	}
+
+	args = append(args, "-t", "virtiofs", mntSource, e.Config["path"])
+
 	for i := 0; i < 5; i++ {
-		_, err = shared.RunCommand("mount", "-t", "virtiofs", mntSource, e.Config["path"])
+		_, err = shared.RunCommandContext(context.Background(), "mount", args...)
 		if err == nil {
 			l.Info("Mounted hotplug")
 			return
--- a/lxd/instance/drivers/driver_common.go
+++ b/lxd/instance/drivers/driver_common.go
@@ -1575,12 +1575,25 @@ func (d *common) devicesUpdate(inst instance.Instance, removeDevices deviceConfi
 
 			if runConf != nil && len(runConf.Mounts) > 0 {
 				for _, opt := range runConf.Mounts[0].Opts {
-					if strings.HasPrefix(opt, "mountTag=") {
-						parts := strings.SplitN(opt, "=", 2)
-						event["mount"] = instancetype.VMAgentMount{
-							Source: parts[1],
-						}
+					key, value, _ := strings.Cut(opt, "=")
+					if key != "mountTag" {
+						continue
 					}
+
+					if value == "" {
+						return nil, errors.New(`Empty "mountTag" on device's mount options`)
+					}
+
+					agentMount := instancetype.VMAgentMount{
+						Source: value,
+					}
+
+					if shared.IsTrue(dev.Config()["readonly"]) {
+						// Tell the agent to mount with "ro" option for consistency.
+						agentMount.Options = []string{"ro"}
+					}
+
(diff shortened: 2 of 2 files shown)


hamistao:
Two possible improvements for the future:
- Also supporting passing mount options to the lxd-agent through `raw.mount.options`, which should be easily doable but would require some discussion first, so I did not want to rush it.
- Removing old mount entries when removing devices, when removing/updating a disk device, old mount entries might linger. In the following example, I created a dir share device called foo with `readonly` enabled, then disabled readonly and enabled it once again, resulting in:

```
$ mount | grep foo
lxd_foo on /mnt type virtiofs (ro,relatime)
lxd_foo on /mnt type virtiofs (rw,relatime)
lxd_foo on /mnt type virtiofs (ro,relatime)
```

lxd/instance/drivers/driver_common.go
```diff
@@ -1574,12 +1574,30 @@ func (d *common) devicesUpdate(inst instance.Instance, removeDevices deviceConfi
 			}
 
 			if runConf != nil && len(runConf.Mounts) > 0 {
+				var source string
+
 				for _, opt := range runConf.Mounts[0].Opts {
 					if strings.HasPrefix(opt, "mountTag=") {
 						parts := strings.SplitN(opt, "=", 2)
```
tomponline, line 1581: we can use strings.Cut here instead

lxd-agent/events.go
```diff
@@ -151,6 +151,13 @@ func eventsProcess(event api.Event) {
 		mntSource = e.Mount.Source
 	}
 
+	// Parse mount options, if provided.
+	var args []string
+	if len(e.Mount.Options) > 0 {
+		args = append(args, "-o")
```
tomponline, line 157: this can be simplified to:

```go
args = append(args, "-o", strings.Join(e.Mount.Options, ","))
```

hamistao:
@tomponline this should be ready to merge

lxd/instance/drivers/driver_common.go
```diff
@@ -1201,7 +1201,10 @@ func (d *common) setCoreSched(pids []int) error {
 		args = append(args, strconv.Itoa(pid))
 	}
 
-	_, err := shared.RunCommand(d.state.OS.ExecPath, args...)
+	ctx, cancel := context.WithTimeout(d.state.ShutdownCtx, 10*time.Second)
```
tomponline, line 1204: lets use just `d.state.ShutdownCtx` here, rather than arbitrary timeout.

lxd-agent/events.go
```diff
@@ -151,6 +152,12 @@ func eventsProcess(event api.Event) {
 		mntSource = e.Mount.Source
 	}
 
+	// Parse mount options, if provided.
+	var args []string
```
tomponline, line 156: @hamistao should we move all the `args` related stuff down to line 171 to keep it all together, or is there a reason you split it up?

lxd/instance/drivers/driver_common.go
```diff
@@ -1574,12 +1577,29 @@ func (d *common) devicesUpdate(inst instance.Instance, removeDevices deviceConfi
...
+						if value == "" {
+							return nil, errors.New("Empty mountTag on device mount options")
 						}
+
+						source = value
+					}
+				}
+
+				if source != "" {
+					var mountOptions []string
```
tomponline, line 1594: Lets rewrite this block more concisely to:

```go
event["mount"] = instancetype.VMAgentMount{
						Source:  source,					
}

if shared.IsTrue(dev.Config()["readonly"]) {
    event["mount"] = append(event["mount"], "ro")
}
```

hamistao: I can't append with events["mount"], as events is a `map[string]any`, but I did the closest thing to your suggestion.

lxd-agent/events.go
hamistao: Good idea! Moved.

lxd/instance/drivers/driver_common.go
```diff
@@ -1201,7 +1201,7 @@ func (d *common) setCoreSched(pids []int) error {
 		args = append(args, strconv.Itoa(pid))
 	}
 
-	_, err := shared.RunCommand(d.state.OS.ExecPath, args...)
+	_, err := shared.RunCommandContext(d.state.ShutdownCtx, d.state.OS.ExecPath, args...)
```
tomponline, line 1204: we dont want to prematurely exit this call, lets use context.Background().

hamistao: Fixed!

tomponline (approved):
Thanks!

PR #14899, opened 2025-01-31
Device: Allow containers with CDI GPU to be restarted even after abrupt host shutdown

gabrielmougard:
The initial error was that the WalkDir callback (see `startCDIDevices` method) meant to cleanup the files was wrong.. The `HasPrefix` conditions were never met. Now they should be correct. This will remove:

- Any unix-char mount prefixed with `cdi.unix`
- Any disk mount prefixed with `cdi.disk`
- Any CDI metadata files (the cdi_config_devices.json file and the cdi_hooks.json file)

This has been tested locally on an dGPU NVIDIA-enabled laptop. Here is the tested approach:

```
lxc init ubuntu:22.04 c1
lxc config device add c1 gpu0 gpu gputype=physical id=nvidia.com/gpu=0
lxc start c1
lxc exec c1 -- nvidia-smi # Wokrking
sudo shutdown -i
# After powering up back the laptop, `c1` was still in a running state and calling nvidia-smi from inside it worked as well.
lxc exec c1 -- nvidia-smi
```

### DIFF
--- a/lxd/device/gpu_physical.go
+++ b/lxd/device/gpu_physical.go
@@ -141,6 +141,31 @@ func (d *gpuPhysical) startCDIDevices(configDevices cdi.ConfigDevices, runConf *
 		}
 	}()
 
+	hooksFilePath := d.generateCDIHooksFilePath()
+	deviceConfigFilePath := d.generateCDIConfigDevicesFilePath()
+
+	// Check if there are any remaining CDI devices in the instance devices directory.
+	// If there are, we need to remove them. These can be present in the case where the device stop hook was not called
+	// (e.g. due to an abrupt host shutdown).
+	err := filepath.WalkDir(d.inst.DevicesPath(), func(path string, e fs.DirEntry, _ error) error {
+		if e.IsDir() {
+			return nil
+		}
+
+		// Remove the CDI device files (both unix-char and disk devices as long as JSON CDI metadata files).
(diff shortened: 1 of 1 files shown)


lxd/device/gpu_physical.go
```diff
@@ -141,6 +141,28 @@ func (d *gpuPhysical) startCDIDevices(configDevices cdi.ConfigDevices, runConf *
...
+	// Check if there are any remaining CDI devices in the instance devices directory.
+	// If there are, we need to remove them. These can be present in the case where the device stop hook was not called
+	// (e.g. due to an abrupt host shutdown).
+	err := filepath.WalkDir(d.inst.DevicesPath(), func(path string, e fs.DirEntry, _ error) error {
+		if e.IsDir() {
+			return nil
+		}
+
+		// Remove the CDI device files (both unix-char and disk devices as long as JSON CDI metadata files).
+		if strings.HasPrefix(e.Name(), cdi.CDIUnixPrefix) || strings.HasPrefix(e.Name(), cdi.CDIDiskPrefix) || strings.HasSuffix(e.Name(), cdi.CDIConfigDevicesFileSuffix) || strings.HasSuffix(e.Name(), cdi.CDIHooksFileSuffix) {
```
tomponline, line 153: The two suffix checks look a bit dangerous to me as we could theoretically have other non-GPU CDI devices witht he same suffice.

Wouldn't it be better to test the file name against generateCDIHooksFilePath() and generateCDIConfigDevicesFilePath()?

But glad to see its been tested this time, thanks.

gabrielmougard: yes! I can change that

tomponline (approved):
ta

Issue #14814, opened 2025-01-17
Let PowerFlex disconnect from specific targest instead of all


roosterfish:
Right now the PowerFlex driver is issuing `nvme disconnect-all` to disconnect from the NVMe/TCP subsystems. However this might produce a conflict if LXD isn't the only user of NVMe/TCP on this host or another storage driver like Pure is used in parallel to PowerFlex.

We can adapt the same techniques as they are performed in the Pure driver to connect to each individual subsystem's address.
Those addresses can be queried from PowerFlex by listing the SDT's availble for the used storage pool.

PR #14900, opened 2025-01-31
Storage: Powerflex use `Connect` and `Disconnect` for NVMe/TCP

roosterfish:
Closes https://github.com/canonical/lxd/issues/14814

This aligns the PowerFlex driver with the implementation of Pure in https://github.com/canonical/lxd/pull/14599 to use the NVMe/TCP `Connect` and `Disconnect` functions instead of their `(Connect|Disconnect)All` counterparts.

Furthermore a patch is introduced to remove the `powerflex.sdt` config key from already existing storage pools. 
Currently only a single address is stored in the key and using `nvme connect-all` all of the other targets (SDTs) are discovered anyway.
As the driver is now querying for all the SDTs in case `powerflex.sdt` config key is not set, the behavior is the exact same.
In addition `powerflex.sdt` now accepts multiple addresses separated by comma which allows overriding the addresses discovered by asking the PowerFlex REST API.

Some additional minor tweaks have been put into place to make more consistent use of the connector's `LoadModules` in case of SDC and to clean up some error messages.

### DIFF
--- a/lxd/storage/connectors/connector.go
+++ b/lxd/storage/connectors/connector.go
@@ -41,9 +41,7 @@ type Connector interface {
 	QualifiedName() (string, error)
 	LoadModules() error
 	Connect(ctx context.Context, targetQN string, targetAddrs ...string) (revert.Hook, error)
-	ConnectAll(ctx context.Context, targetAddr string) error
 	Disconnect(targetQN string) error
-	DisconnectAll() error
 	SessionID(targetQN string) (string, error)
 	findSession(targetQN string) (*session, error)
 }
--- a/lxd/storage/connectors/connector.go
+++ b/lxd/storage/connectors/connector.go
@@ -42,7 +42,6 @@ type Connector interface {
 	LoadModules() error
 	Connect(ctx context.Context, targetQN string, targetAddrs ...string) (revert.Hook, error)
 	Disconnect(targetQN string) error
-	SessionID(targetQN string) (string, error)
 	findSession(targetQN string) (*session, error)
 }
 
--- a/lxd/storage/drivers/driver_powerflex_utils.go
+++ b/lxd/storage/drivers/driver_powerflex_utils.go
@@ -1241,12 +1241,12 @@ func (d *powerflex) resolvePool() (*powerFlexStoragePool, error) {
 func (d *powerflex) getVolumeName(vol Volume) (string, error) {
 	volUUID, err := uuid.Parse(vol.config["volatile.uuid"])
 	if err != nil {
-		return "", fmt.Errorf(`Failed parsing "volatile.uuid" from volume %q: %w`, vol.name, err)
+		return "", fmt.Errorf("Failed parsing %q from volume %q: %w", "volatile.uuid", vol.name, err)
 	}
 
 	binUUID, err := volUUID.MarshalBinary()
 	if err != nil {
-		return "", fmt.Errorf(`Failed marshalling the "volatile.uuid" of volume %q to binary format: %w`, vol.name, err)
+		return "", fmt.Errorf("Failed marshalling the %q of volume %q to binary format: %w", "volatile.uuid", vol.name, err)
 	}
 
 	// The volume's name in base64 encoded format.
--- a/lxd/storage/drivers/driver_powerflex_utils.go
+++ b/lxd/storage/drivers/driver_powerflex_utils.go
@@ -996,15 +996,28 @@ func (d *powerflex) mapVolume(vol Volume) (revert.Hook, error) {
 		reverter.Add(func() { _ = client.deleteHostVolumeMapping(hostID, volumeID) })
 	}
 
-	targetAddr := d.config["powerflex.sdt"]
+	var cleanup revert.Hook
+	if d.config["powerflex.mode"] == connectors.TypeNVME {
+		// Discover all SDTs from PowerFlex for the respective storage pool.
+		targetAddresses, err := d.getNVMeTargetAddresses()
+		if err != nil {
+			return nil, err
+		}
 
-	// Connect to the storage subsystem.
-	// In case of NVMe/TCP, we have to connect after the first mapping was established,
-	// as PowerFlex does not offer any discovery log entries until a volume gets mapped
-	// to the host.
-	err = connector.ConnectAll(d.state.ShutdownCtx, targetAddr)
-	if err != nil {
-		return nil, err
+		// Discover the SDT's targetQN from any of the addresses.
+		targetQN, err := d.getNVMeTargetQN(targetAddresses...)
+		if err != nil {
+			return nil, err
+		}
+
+		// Connect to the storage subsystem.
+		// In case of NVMe/TCP, we have to connect after the first mapping was established,
+		// as PowerFlex does not offer any discovery log entries until a volume gets mapped
+		// to the host.
+		cleanup, err = connector.Connect(d.state.ShutdownCtx, targetQN, targetAddresses...)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	// Reverting mapping or connection outside mapVolume function
@@ -1013,8 +1026,17 @@ func (d *powerflex) mapVolume(vol Volume) (revert.Hook, error) {
 	// because it ensures the lock is acquired and accounts for
 	// an existing session before unmapping a volume.
 	outerReverter := revert.New()
+
 	if !mapped {
 		outerReverter.Add(func() { _ = d.unmapVolume(vol) })
+	}
+
+	// Add the cleanup hooks of the connection attempt to the outer reverter.
+	// This ensures that ongoing connection attempts that haven't yet finished are cancelled
+	// before potentially running unmap volume.
+	// As the revert hooks are called in reverse order add the connection cleanup after unmap.
+	if d.config["powerflex.mode"] == connectors.TypeNVME {
+		outerReverter.Add(cleanup)
 	}
 
 	reverter.Success()
@@ -1149,9 +1171,19 @@ func (d *powerflex) unmapVolume(vol Volume) error {
 		}
 
 		if len(mappings) == 0 {
+			targetAddresses, err := d.getNVMeTargetAddresses()
+			if err != nil {
+				return err
+			}
+
+			targetQN, err := d.getNVMeTargetQN(targetAddresses...)
+			if err != nil {
+				return err
+			}
+
 			// Disconnect from the NVMe subsystem.
 			// Do this first before removing the host from PowerFlex.
-			err = connector.DisconnectAll()
+			err = connector.Disconnect(targetQN)
 			if err != nil {
 				return err
 			}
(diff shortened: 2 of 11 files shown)


lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -933,17 +938,51 @@ func (d *powerflex) mapVolume(vol Volume) (revert.Hook, error) {
...
+			for _, ip := range relation.IPList {
+				targetAddresses = append(targetAddresses, ip.IP)
+			}
+		}
+
+		if len(targetAddresses) == 0 {
+			return nil, fmt.Errorf("Failed to retrieve at least one SDT for the given storage pool: %q", pool.ID)
+		}
+	} else {
+		targetAddresses = strings.Split(d.config["powerflex.sdt"], ",")
```
tomponline, line 967: Suggest using 

```go

targetAddresses : =shared.SplitNTrimSpace(d.config["powerflex.sdt"], ",", -1, true)

if len(targetAddresses) < 1 {
    relations, err := client.getProtectionDomainSDTRelations(pool.ProtectionDomainID)
    ...
}
```

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -858,6 +858,11 @@ func (d *powerflex) deleteNVMeHost() error {
 	return client.deleteHost(host.ID)
 }
 
+// targetQN returns the unique NQN of the target SDTs using PowerFlex's system ID.
+func (d *powerflex) targetQN(systemID string) string {
+	return fmt.Sprintf("nqn.1988-11.com.dell:powerflex:00:%s", systemID)
```
MusicDin, line 863: I guess the nqn prefix is the same for all PF system, just wanted to point out if we are sure this is true? Because if the prefix does not match, the connection will fail.

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -933,17 +938,51 @@ func (d *powerflex) mapVolume(vol Volume) (revert.Hook, error) {
...
+	cleanup, err := connector.Connect(d.state.ShutdownCtx, d.targetQN(domain.SystemID), targetAddresses...)
 	if err != nil {
 		return nil, err
 	}
 
+	reverter.Add(cleanup)
+
 	// Reverting mapping or connection outside mapVolume function
 	// could conflict with other ongoing operations as lock will
 	// already be released. Therefore, use unmapVolume instead
```
MusicDin, line 988: Could you just correct this comment? It is mistake from my PR - the cleanup from `connector.Connect()` function no longer reverts the connections.

lxd/storage/connectors/connector.go
```diff
@@ -38,9 +38,7 @@ type Connector interface {
 	QualifiedName() (string, error)
 	LoadModules() error
 	Connect(ctx context.Context, targetQN string, targetAddrs ...string) (revert.Hook, error)
-	ConnectAll(ctx context.Context, targetAddr string) error
 	Disconnect(targetQN string) error
-	DisconnectAll() error
 	SessionID(targetQN string) (string, error)
```
MusicDin, line 42: Is `SessionID` used anywhere? Otherwise, this can go out as well

lxd/storage/drivers/driver_powerflex_utils.go
roosterfish: Had the same thoughts. Unfortunately this string doesn't seem to be returned by any PowerFlex API. I tried querying the relevant ones with no luck.

Also all the online examples always use `nqn.1988-11.com.dell:powerflex:00:` so I suspect this to be the same every time.

roosterfish: IIRC in case of pure you get it straight from the API right?

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -933,17 +938,49 @@ func (d *powerflex) mapVolume(vol Volume) (revert.Hook, error) {
...
 	// In case of NVMe/TCP, we have to connect after the first mapping was established,
 	// as PowerFlex does not offer any discovery log entries until a volume gets mapped
 	// to the host.
-	err = connector.ConnectAll(d.state.ShutdownCtx, targetAddr)
+	cleanup, err := connector.Connect(d.state.ShutdownCtx, d.targetQN(domain.SystemID), targetAddresses...)
 	if err != nil {
 		return nil, err
 	}
 
+	reverter.Add(cleanup)
```
MusicDin, line 982: I think the cleanup from connect should also be added to the outer reverter, to ensure connections attempts are canceled before the unmapVolume is called (for example, if some connection is failing in the background but has not yet timed out)

lxd/storage/drivers/driver_powerflex_utils.go
MusicDin: > IIRC in case of pure you get it straight from the API right?

Yes.

> Unfortunately this string doesn't seem to be returned by any PowerFlex API. I tried querying the relevant ones with no luck.

I think the prefix should be the same on all PF arrays, as the prefix is constructed as:
`<protocol>.<year_of_company_creation?>.<domain_in_reverse>:<system_id>`

Just not sure if this is really a standard, or if it can vary for unknown reasons

lxd/storage/drivers/driver_powerflex_utils.go
roosterfish: Fixed.

lxd/storage/drivers/driver_powerflex_utils.go
roosterfish: You are right, was missing the fact that we have introduced the `outerReverter` for this exact reason. Fixed.

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -933,23 +938,56 @@ func (d *powerflex) mapVolume(vol Volume) (revert.Hook, error) {
...
-	// could conflict with other ongoing operations as lock will
-	// already be released. Therefore, use unmapVolume instead
-	// because it ensures the lock is acquired and accounts for
-	// an existing session before unmapping a volume.
 	outerReverter := revert.New()
+
+	// Add the cleanup hooks of the connection attempt to the outer reverter.
+	// This ensures that ongoing connection attempts that haven't yet finished are cancelled
+	// before potentially running unmap volume.
+	outerReverter.Add(cleanup)
```
MusicDin, line 987: I did the same mistake in Pure, but the cleanup from `connect` should be added after `unmapVolume` (or before it if in the same function) as the reverter calls functions in the reverse order.

lxd/storage/drivers/driver_powerflex_utils.go
roosterfish: Thanks, fixed. In the end I was using `if targetAddresses == nil` as nil is returned by `SplitNTrimSpace` in case the config key is empty.

lxd/storage/drivers/driver_powerflex_utils.go
roosterfish: An alternative would be to keep `ConnectAll` to make use of the discovery. This way we don't need to craft the target's nqn as we get it from NVMe discovery.
But then `ConnectAll` should support multiple addresses, the same as `Connect`.

And `Disconnect` already supports just providing a string that is contained in any of the current session so in case of PowerFlex we can just pass the system id.

lxd/storage/drivers/driver_powerflex_utils.go
roosterfish: Thanks, fixed.

lxd/storage/drivers/driver_powerflex_utils.go
roosterfish: To be on the safe side we now either fetch the list of SDTs (targets) from PowerFlex or use the addresses from `powerflex.sdt` config key to run the NVMe/TCP discovery in order to lookup the used `targetQN`.
It is cached in the drivers struct to speed up subsequent lookups.

roosterfish:
Ready for review now. In parallel I am running the `nvme`/`sdc` mode tests.

lxd/storage/drivers/driver_powerflex_utils.go
tomponline: As discussed in our 1:1, lets do this concurrently to avoid delay in initial connection when some of the targets are unreachable.

roosterfish:
I have moved the ideas around managing many concurrent connections in case of timeouts to another PR https://github.com/canonical/lxd/pull/14955.

roosterfish:
Given the recent findings in https://github.com/canonical/lxd/pull/14957#discussion_r1951287352 regarding the blocking `nvme connect` commands, I wonder if we should rather use `nvme connect-all` overall.

tomponline:
@roosterfish is this blocked atm wrt to the nvme concurrent connect issue?

roosterfish:
> @roosterfish is this blocked atm wrt to the nvme concurrent connect issue?

No this can be merged/is ready for review.

As it's now using the common connector it is affected by the recent findings around the sequential execution of `nvme connect` but this should not be a blocker for this PR.

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -996,15 +996,28 @@ func (d *powerflex) mapVolume(vol Volume) (revert.Hook, error) {
...
 
-	// Connect to the storage subsystem.
-	// In case of NVMe/TCP, we have to connect after the first mapping was established,
-	// as PowerFlex does not offer any discovery log entries until a volume gets mapped
-	// to the host.
-	err = connector.ConnectAll(d.state.ShutdownCtx, targetAddr)
-	if err != nil {
-		return nil, err
+		// Discover the SDT's targetQN from any of the addresses.
+		targetQN, err := d.getNVMeTargetQN(targetAddresses...)
```
tomponline, line 1008: @roosterfish so the QN discovery is done on every map volume request?

tomponline (approved):
Thanks!

lxd/storage/drivers/driver_powerflex_utils.go
roosterfish: If it's not already stored in the drivers struct field then yes. 

Another option would have been storing the QN on package level in another variable and protect it with a lock but I found this approach to be more in line with the rest of the client calls that try to fetch information.

tomponline: ah ok, i got the impression you were going to cache the result, but not really as its only stored per-operation effectively.

I suppose we still have to get the SDT addresses via the powerflex API too, because otherwise we'd have no initial addresses to use for discovery.

roosterfish: Yes, the SDT addresses are fetched in any case (to cope with changes if an admin decides to move some SDT's to other addresses).

There are some scenarios where the current cache on struct level is helpful, e.g. when you launch a VM and it tries to mount both block vol and FS vol.
Or when you perform storage operations with snapshots and so on as long as the same instance of the struct is used.

PR #14901, opened 2025-01-31
Follow-up on `fmt.Sprintf()` replacement

simondeziel:
This fixes an unintended change introduced by https://github.com/canonical/lxd/commit/963f555865fbf8cdc4855fe458430f48bfd32aad

### DIFF
--- a/client/lxd_images.go
+++ b/client/lxd_images.go
@@ -1052,7 +1052,7 @@ func (r *ProtocolLXD) ExportImage(fingerprint string, image api.ImageExportPost)
 	}
 
 	// Send the request
-	op, _, err := r.queryOperation("POST", "/images/"+url.PathEscape(fingerprint)+"/export", image, "", true)
+	op, _, err := r.queryOperation("POST", "/images/"+url.PathEscape(fingerprint)+"/export", &image, "", true)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


tomponline:
@simondeziel please can you check if any of these made it into stable-5.21 branch, ta

client/lxd_images.go
```diff
@@ -1052,7 +1052,7 @@ func (r *ProtocolLXD) ExportImage(fingerprint string, image api.ImageExportPost)
 	}
 
 	// Send the request
-	op, _, err := r.queryOperation("POST", "/images/"+url.PathEscape(fingerprint)+"/export", image, "", true)
+	op, _, err := r.queryOperation("POST", "/images/"+url.PathEscape(fingerprint)+"/export", &image, "", true)
```
tomponline, line 1055: @simondeziel please can you add a test for this as looks like we have a gap that allowed this regression through, ta.

simondeziel:
> @simondeziel please can you check if any of these made it into stable-5.21 branch, ta

Good news, none of those `Sprintf` commits have been backported to `stable-5.21`.

client/lxd_images.go
simondeziel: @tomponline could you help me with how to test this out? I can hit the `GET` handler with `lxc image export <fingerprint>` but I still haven't found a way to get to the `POST` handler :/

tomponline: Which function is the post handler?

tomponline (approved):

client/lxd_images.go
simondeziel: `ExportImage()` from `client/lxd_images.go`

PR #14904, opened 2025-01-31
Network: Fix OVN ChassisGroupChassisDelete to convert chassis_name

### LINKED ISSUE #14902: OVN chassis stick around when evacuating a cluster member
The `ValueInSlice` test here: https://github.com/canonical/lxd/blob/main/lxd/network/openvswitch/ovn.go#L1628 never works as we are comparing the chassis' name with uuids.

A quick test forcing the code into executing `ha-chassis-group-remove-chassis` evacuates and restores the chassis correctly.

escabo:
Now gets the list of ha_chassis first to convert the received chassis_name into a _uuid for later comparison.
Fixes #14902.

### DIFF
--- a/lxd/network/openvswitch/ovn.go
+++ b/lxd/network/openvswitch/ovn.go
@@ -575,7 +575,7 @@ func (o *OVN) LogicalRouterPortSetIPv6Advertisements(portName OVNRouterPort, opt
 		fmt.Sprintf("ipv6_ra_configs:send_periodic=%t", opts.SendPeriodic),
 	}
 
-	var removeRAConfigKeys []string
+	removeRAConfigKeys := make([]string, 0) //nolint:prealloc
 
 	if opts.AddressMode != "" {
 		args = append(args, fmt.Sprintf("ipv6_ra_configs:address_mode=%s", string(opts.AddressMode)))
@@ -766,7 +766,7 @@ func (o *OVN) logicalSwitchParseExcludeIPs(ips []shared.IPRange) ([]string, erro
 
 // LogicalSwitchSetIPAllocation sets the IP allocation config on the logical switch.
 func (o *OVN) LogicalSwitchSetIPAllocation(switchName OVNSwitch, opts *OVNIPAllocationOpts) error {
-	var removeOtherConfigKeys []string
+	removeOtherConfigKeys := make([]string, 0) //nolint:prealloc
 	args := []string{"set", "logical_switch", string(switchName)}
 
 	if opts.PrefixIPv4 != nil {
@@ -814,7 +814,7 @@ func (o *OVN) LogicalSwitchSetIPAllocation(switchName OVNSwitch, opts *OVNIPAllo
 
 // LogicalSwitchDHCPv4RevervationsSet sets the DHCPv4 IP reservations.
 func (o *OVN) LogicalSwitchDHCPv4RevervationsSet(switchName OVNSwitch, reservedIPs []shared.IPRange) error {
-	var removeOtherConfigKeys []string
+	removeOtherConfigKeys := make([]string, 0) //nolint:prealloc
 	args := []string{"set", "logical_switch", string(switchName)}
 
 	if len(reservedIPs) > 0 {
@@ -1187,7 +1187,7 @@ func (o *OVN) LogicalSwitchIPs(switchName OVNSwitch) (map[OVNSwitchPort][]net.IP
 	for _, line := range lines {
 		fields := shared.SplitNTrimSpace(line, ",", -1, true)
 		portName := OVNSwitchPort(fields[0])
-		var ips []net.IP
+		ips := make([]net.IP, 0, len(fields))
 
 		// Parse all IPs mentioned in addresses and dynamic_addresses fields.
 		for i := 1; i < len(fields); i++ {
@@ -1290,7 +1290,7 @@ func (o *OVN) LogicalSwitchPortIPs(portName OVNSwitchPort) ([]net.IP, error) {
 	}
 
 	addresses := strings.Split(strings.Replace(strings.TrimSpace(addressesRaw), ",", " ", 1), " ")
-	ips := make([]net.IP, 0)
+	ips := make([]net.IP, 0, len(addresses))
 
 	for _, address := range addresses {
 		ip := net.ParseIP(address)
@@ -1449,7 +1449,7 @@ func (o *OVN) LogicalSwitchPortGetDNS(portName OVNSwitchPort) (OVNDNSUUID, strin
 	dnsUUID := strings.TrimSpace(parts[0])
 
 	var dnsName string
-	var ips []net.IP
+	ips := make([]net.IP, 0) //nolint:prealloc
 
 	// Try and parse the DNS name and IPs.
 	if len(parts) > 1 {
@@ -1730,7 +1730,7 @@ func (o *OVN) PortGroupAdd(projectID int64, portGroupName OVNPortGroup, associat
 
 // PortGroupDelete deletes port groups along with their ACL rules.
 func (o *OVN) PortGroupDelete(portGroupNames ...OVNPortGroup) error {
-	args := make([]string, 0)
+	args := make([]string, 0, 5*len(portGroupNames))
 
 	for _, portGroupName := range portGroupNames {
 		if len(args) > 0 {
@@ -1925,7 +1925,7 @@ func (o *OVN) loadBalancerUUIDs(loadBalancerName OVNLoadBalancer) ([]string, err
 	lbTCPName := fmt.Sprintf("%s-tcp", loadBalancerName)
 	lbUDPName := fmt.Sprintf("%s-udp", loadBalancerName)
 
-	var lbUUIDs []string
+	lbUUIDs := make([]string, 0) //nolint:prealloc
 
 	// Use find command in order to workaround OVN bug where duplicate records of same name can exist.
 	for _, lbName := range []string{lbTCPName, lbUDPName} {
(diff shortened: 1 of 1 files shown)


lxd/network/openvswitch/ovn.go
```diff
@@ -1613,19 +1613,48 @@ func (o *OVN) ChassisGroupChassisAdd(haChassisGroupName OVNChassisGroup, chassis
...
+			break
+		}
+	}
+
+	if chassisUUID == "" {
+		return fmt.Errorf("Could not find chassis with name %q", chassisID)
+	}
+
+	// Check if chassis group exists. ovn-nbctl doesn't provide an "--if-exists" option for this.
+	output, err = o.nbctl("--no-headings", "--data=bare", "--colum=name,ha_chassis", "find", "ha_chassis_group", fmt.Sprintf("name=%s", string(haChassisGroupName)))
```
simondeziel, line 1646:
```suggestion
	output, err = o.nbctl("--no-headings", "--data=bare", "--colum=name,ha_chassis", "find", "ha_chassis_group", "name="+string(haChassisGroupName))
```

`fmt.Sprintf()` is slow so we recently started replacing it with simpler/faster string concatenation.

tomponline (changes requested):
Ultimately this issue was introduced by myself in https://github.com/canonical/lxd/commit/1d98019ecb8fb28669f1eb5d3442da42c7b7fb96 trying to workaround the issue that the `ovn-nbctl ha-chassis-group-remove-chassis` command doesn't provide an `--if-exists` option where if the chassis member doesn't exist in the group, then no error will be raised.

> Previously if the chassis was a member of multiple chassis groups (OVN networks) then if one of the OVN networks couldn't
be started (and thus its chassis wasn't added to that network's chassis group) then attempting to remove unavailable OVN
network would fail because the previous implementation thought the chassis was in the chassis group, when infact it was
in another chassis group.

However this PR as it stands introduces another issue because it does not consider that the same chassis can have multiple `ha_chassis` entries (one per ovn network). This is done to achieve the per-ovn network chassis failover priority, in order to distribute uplink traffic across the cluster.

Taking the example of a microcloud with 2 OVN networks:

```
ovn-nbctl find ha_chassis
_uuid               : f139a82d-7d32-4799-80ce-6feecbb4b3df
chassis_name        : micro01
external_ids        : {}
priority            : 26056

_uuid               : 6fe7b2ed-d98b-4e64-9585-0eef369ba082
chassis_name        : micro01
external_ids        : {}
priority            : 1138

_uuid               : a1a8d8c2-2cc6-4806-8bd8-5a4ddc485890
chassis_name        : micro03
external_ids        : {}
priority            : 15471

_uuid               : a07811af-327d-4467-ae7e-3be7afc98098
chassis_name        : micro02
external_ids        : {}
priority            : 10207

_uuid               : a7efafc7-9f65-4383-aa9d-ab8cad5443ae
chassis_name        : micro03
external_ids        : {}
priority            : 29049

_uuid               : 701780b3-8930-4365-abe7-85ab0865356e
chassis_name        : micro02
external_ids        : {}
priority            : 22819
```

```
ovn-nbctl find ha_chassis_group
_uuid               : 2df9866e-03a8-4bb9-9447-5f928014f338
external_ids        : {}
ha_chassis          : [a07811af-327d-4467-ae7e-3be7afc98098, a1a8d8c2-2cc6-4806-8bd8-5a4ddc485890, f139a82d-7d32-4799-80ce-6feecbb4b3df]
name                : lxd-net3

_uuid               : 9cc51612-237f-43dd-bb32-aebf191bf3c6
external_ids        : {}
ha_chassis          : [6fe7b2ed-d98b-4e64-9585-0eef369ba082, 701780b3-8930-4365-abe7-85ab0865356e, a7efafc7-9f65-4383-aa9d-ab8cad5443ae]
name                : lxd-net2
```

We see that `micro01` has 2 `ha_chassis` entries:

```
_uuid               : f139a82d-7d32-4799-80ce-6feecbb4b3df
chassis_name        : micro01
external_ids        : {}
priority            : 26056
```

and

```
_uuid               : 6fe7b2ed-d98b-4e64-9585-0eef369ba082
chassis_name        : micro01
external_ids        : {}
priority            : 1138
```

The `f139a82d-7d32-4799-80ce-6feecbb4b3df` is used in the `ha_chassis_group` for ` lxd-net3` network and `6fe7b2ed-d98b-4e64-9585-0eef369ba082` is used in the `has_chassis_group` for `lxd-net2` network respectively.

So the problem with the PR is that it takes the entries in the `ha_chassis` table and matches the first entry that matches the cluster member's chassis ID.

Then it will use that UUID to match against the `ha_chassis_group` members entries for the `ha_chassis_group` entry for the associated network.

This is likely to not match correctly, at least some of the time.

The naive approach we could use is to simply lookup each chassis member using the UUID and convert it to a chassis ID/name via:

```
ovn-nbctl get ha_chassis <group member UUID> chassis_name
```

But because each `ovn-nbctl` invocation is expensive, because it connects to the ovn cluster and loads part of the database, I think we should load all ha_chassis members and iterate through them.

Once we switch to libovsdb, we can switch to just doing a chassis name lookup for each chassis member like Incus do here:

https://github.com/lxc/incus/blob/main/internal/server/network/ovn/ovn_nb_actions.go#L2401-L2410

So for now  I think we should do this:

1. First, get chassis members from the `ha_chassis_group` table, using the existing command we are using.
2. Next, get list of `ha_chassis` entries, and see if there is an entry that matches the cluster member's chassis ID *and* where its UUID is inside the chassis members list for this network's chassis group.


We can also add a test to https://github.com/canonical/lxd/blob/main/test/suites/network_ovn.sh such that when `lxd shutdown` is called, the chassis group member entry should be removed.

tomponline:
@escabo id like to get this into 5.21/candidate as a cherry-pick before we push 5.21.3 to 5.21/stable if timing works out.

simondeziel:
@escabo I think you need to rebase (`git rebase -i main`) to avoid pushing back old commits that are already in `main`.

simondeziel (approved):

tomponline (changes requested):
If you have time @escabo please can you separate the chassis change and the linter fixes into separate commits?

lxd/network/openvswitch/ovn.go
```diff
@@ -1613,22 +1613,48 @@ func (o *OVN) ChassisGroupChassisAdd(haChassisGroupName OVNChassisGroup, chassis
 
 // ChassisGroupChassisDelete deletes a chassis ID from an HA chassis group.
 func (o *OVN) ChassisGroupChassisDelete(haChassisGroupName OVNChassisGroup, chassisID string) error {
-	// Check if chassis group exists. ovn-nbctl doesn't provide an "--if-exists" option for this.
-	output, err := o.nbctl("--no-headings", "--data=bare", "--colum=name,ha_chassis", "find", "ha_chassis_group", fmt.Sprintf("name=%s", string(haChassisGroupName)))
+	// Map UUIDs with chassis_names.
+	output, err := o.nbctl("--format=csv", "--no-headings", "--data=bare", "--column=_uuid,chassis_name", "list", "ha_chassis")
```
tomponline, line 1617: We've tended to use `--data=bare` when we we know that the first col wont contain the space char that we can split the columns by (so in this case uuid wont contain space), as this avoids having to undo the quoting that csv output may add.

We also tend to use the `find` command rather than `list`, but i would need to check why we've used that more commonly.

lxd/network/openvswitch/ovn.go
```diff
@@ -1613,22 +1613,48 @@ func (o *OVN) ChassisGroupChassisAdd(haChassisGroupName OVNChassisGroup, chassis
...
 		return err
 	}
 
 	lines := shared.SplitNTrimSpace(output, "\n", -1, true)
+
+	uuidToChassis := make(map[string]string, len(lines))
+
+	for _, line := range lines {
+		// a74125a8-b580-4763-b389-11ce2c8c5509,node2
+		fields := strings.Split(line, ",")
```
tomponline, line 1628: Irrespective of whether you're splitting by comma (for csv) or space (for bare) we should use either `strings.Cut` or `strings.SplitN(line, ",",2)` to make explicit we expect only 2 fields.

`strings.Cut` here would also avoid the need for the `len` check below.

tomponline (commented):
Why did you pick 10 for pre allocation in some cases?

escabo:
> Why did you pick 10 for pre allocation in some cases?

for occurrences where there wasn't a source to get the size from and we needed a handful of elements, I thought this was a nice round number :-)

lxd/network/openvswitch/ovn.go
escabo: like this?
```
	for _, line := range lines {
		// a74125a8-b580-4763-b389-11ce2c8c5509,node2
		key, value, match := strings.Cut(line, ",")
		if match {
			uuidToChassis[key] = value
		}
	}
```

simondeziel: Exactly.

tomponline:
Please can you rebase as ive merged the ovn ptr pr now and there are some conflicts, then this should be ready to merge and ill follow up with some tesrs

escabo:
> Please can you rebase as ive merged the ovn ptr pr now and there are some conflicts, then this should be ready to merge and ill follow up with some tests. Thanks

done, thank you!

tomponline:
Got a couple of whitespace linter errors there

escabo:
> Got a couple of whitespace linter errors there

Yep, fixed.

tomponline (approved):
Thanks!

PR #14917, opened 2025-02-03
Network: Update physical NIC device options for added clarity

kadinsayani:
This PR updates the long description for `agent.nic_config` and adds a container condition to container-specific physical NIC device options.

### DIFF
--- a/lxd/device/nic.go
+++ b/lxd/device/nic.go
@@ -97,6 +97,7 @@ func nicValidationRules(requiredFields []string, optionalFields []string, instCo
 		//  defaultdesc: parent MTU
 		//  managed: no
 		//  shortdesc: MTU of the new interface
+		//  condition: container
 
 		// lxdmeta:generate(entities=device-nic-{ipvlan+routed}; group=device-conf; key=mtu)
 		//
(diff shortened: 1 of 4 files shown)


simondeziel:
Looks like your first commit was overridden by the `make update-api`. Was the edit made to a generated file?

kadinsayani:
> Looks like your first commit was overridden by the `make update-api`. Was the edit made to a generated file?

Not sure what happened there, thanks for catching.

lxd/device/nic.go
```diff
@@ -126,13 +127,21 @@ func nicValidationRules(requiredFields []string, optionalFields []string, instCo
...
 		//  type: integer
 		//  managed: no
 		//  shortdesc: VLAN ID to attach to
 
+		// lxdmeta:generate(entities=device-nic-physical; group=device-conf; key=vlan)
+		//
+		// ---
+		//  type: integer
+		//  managed: no
+		//  shortdesc: VLAN ID to attach to
```
tomponline, line 142: @kadinsayani why is this container specific?

lxd/device/nic.go
```diff
@@ -176,6 +185,7 @@ func nicValidationRules(requiredFields []string, optionalFields []string, instCo
 		//  defaultdesc: parent MAC address
 		//  managed: no
 		//  shortdesc: MAC address of the new interface
+		//  condition: container
```
tomponline, line 188: @kadinsayani does this not work for VMs?

lxd/device/nic.go
kadinsayani: For VM PCI passthrough these settings are dependent on the host device so we only allow them to be configured for containers. Here's where we set the required and optional fields:

https://github.com/canonical/lxd/blob/update-agent-nic-desc/lxd/device/nic_physical.go#L34-L51

tomponline: dead link

tomponline: but understood, as its being detached from the host, makes sense

tomponline (approved):

lxd/device/nic.go
kadinsayani: It isn't directly configurable for VMs - it can only set by `lxd-agent` when `agent.nic_config` is set. @simondeziel and I discovered this when test running https://github.com/canonical/lxd-ci/commit/447d40974feae8de3e023d977894d1e13f2256d2 on a testflinger node. We also discovered that `tests/network` does not run successfully on the node due to the VFs not having assigned IPs on the testflinger node (similar to the issue we faced when configuring a DPU).

tomponline: OK, I have an SR-IOV card locally and the tests were working fine for LXD 5.21.3 and LXD 6.2, so perhaps its a recent break in the test itself?

tomponline: In my lab I have the SR-IOV VFs linked to a network that has dhcp.

kadinsayani: I don't think it's a recent break in the tests but rather an issue with the testflinger node network.

PR #14920, opened 2025-02-04
Storage: Delete potentially left snapshot temporary volume

MusicDin:
Pure Storage does not allow mounting snapshots directly. Instead, we create a temporary volume with the snapshot's volume name. If the LXD was abruptly closed when temporary volume is being used, it may remain in the Pure Storage until the pool is removed.

To prevent such situation, when snapshot is removed also remove the potentially existing temporary snapshot volume.

Addresses comment: https://github.com/canonical/lxd/pull/14599#discussion_r1940802947


tomponline (approved):

tomponline (approved):
Ta

PR #14921, opened 2025-02-04
Storage: Restore specialized mount functions 

MggMuggins:
This is largely a revert of #14491. The functionality provided by the reverted commits is implemented in the last commit here.

Rather than using a common `MountVolume` which changes its behavior based on volume type, the disk device driver should choose the function to call based on the volume type provided by the user in the `source-type` field.


tomponline (approved):
Thanks!

PR #14923, opened 2025-02-05
Storage: Avoid filling volume config on `VolumeDBCreate`

hamistao:
This is a pre-requisite for https://github.com/canonical/lxd/pull/14595.

The whole idea here is that we want to adapt the provided volume config on `VolumeDBCreate` to populate `volatile.rootfs.size` without changing the original config passed in, as this would be a very easy and consistent way to keep track of volume sizes.

But we are currently relying on `VolumeDBCreate` to populate some fields within the volume for future use, which is giving many unrelated purposes to this function. So this instead proposed filling this config on `GetNewVolume` instead, as it makes sense for a new volume to have default conifg for the driver filled in. That, along with a few more tweaks, allows us to use `VolumeDBCreate` exclusively to create the database entry for the volume, adapting the volume config as needed without compromising the original config passed in as an argument.

### DIFF
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -250,10 +250,14 @@ func (b *lxdBackend) Create(clientType request.ClientType, op *operations.Operat
 }
 
 // GetNewVolume returns a drivers.Volume that doesn't yet exist in the database.
-// It contains copies of the supplied volume config, including a new UUID, and the pools config.
+// It contains copies of the supplied volume config, including a new UUID and
+// default configuration for the poo driver, as well as the pool's config.
 // Use the returned drivers.Volume as the base for actions performed on the new volume.
 func (b *lxdBackend) GetNewVolume(volType drivers.VolumeType, contentType drivers.ContentType, volName string, volConfig map[string]string) drivers.Volume {
 	newVol := b.GetVolume(volType, contentType, volName, volConfig)
+
+	// Fill default config.
+	b.Driver().FillVolumeConfig(newVol)
 
 	// Set a new UUID.
 	newVol.Config()["volatile.uuid"] = uuid.New().String()
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -874,6 +874,9 @@ func (b *lxdBackend) CreateInstanceFromBackup(srcBackup backup.Info, srcData io.
 			}
 		}
 
+		// Fill volume config with driver defaults before writing to the database.
+		b.driver.FillVolumeConfig(vol)
+
 		// Validate config and create database entry for new storage volume.
 		// Strip unsupported config keys (in case the export was made from a different type of storage pool).
 		err = VolumeDBCreate(b, inst.Project().Name, inst.Name(), vol, volumeDescription, false, volumeCreationDate, time.Time{}, true, true)
@@ -2244,6 +2247,9 @@ func (b *lxdBackend) CreateInstanceFromMigration(inst instance.Instance, conn io
 				return fmt.Errorf("Cannot create volume, already exists on migration target storage")
 			}
 		} else {
+			// Ensure driver specific default config is filled in new volume.
+			b.driver.FillVolumeConfig(vol)
+
 			// Validate config and create database entry for new storage volume if not refreshing.
 			// Strip unsupported config keys (in case the export was made from a different type of storage pool).
 			err = VolumeDBCreate(b, inst.Project().Name, inst.Name(), vol, volumeDescription, false, inst.CreationDate(), time.Time{}, true, true)
@@ -5800,6 +5806,9 @@ func (b *lxdBackend) CreateCustomVolumeFromMigration(projectName string, conn io
 	defer revert.Fail()
 
 	if !args.Refresh {
+		// Ensure driver specific default config is filled in new volume.
+		b.driver.FillVolumeConfig(vol)
+
 		// Validate config and create database entry for new storage volume.
 		// Strip unsupported config keys (in case the export was made from a different type of storage pool).
 		err = VolumeDBCreate(b, projectName, args.Name, vol, args.Description, false, time.Now().UTC(), time.Time{}, true, true)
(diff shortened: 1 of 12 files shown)


lxd/storage/utils.go
tomponline, line 1: I would suggest the avoidance of the term "weird" in the commit message, as its likely that there has been a change in situation where previously it made sense to do this here, but perhaps now it doesnt.

tomponline: It was actually added here as a safety precaution to avoid forgetting to call it. Taking it out needs careful consideration.

lxd/storage/backend_lxd.go
tomponline, line 1: Indeed `GetNewVolume` did not exist when `VolumeDBCreate` was made to fill volume config.

It made sense at the time that if you're creating a new DB volume you are likely making a new volume.

But this subtlety has changed over the years.

lxd/storage/backend_lxd.go
```diff
@@ -250,11 +250,15 @@ func (b *lxdBackend) Create(clientType request.ClientType, op *operations.Operat
 }
 
 // GetNewVolume returns a drivers.Volume that doesn't yet exist in the database.
-// It contains copies of the supplied volume config, including a new UUID, and the pools config.
+// It contains copies of the supplied volume config, including a new UUID and
+// default configuration for the poo driver, as well as the pool's config.
```
tomponline, line 254: typo here

tomponline: Lets rephase to:

> // It contains copies of the supplied volume config and populates a new UUID and any pool level defaults.

lxd/storage/backend_lxd.go
```diff
@@ -7481,8 +7478,7 @@ func (b *lxdBackend) ImportInstance(inst instance.Instance, poolVol *backupConfi
 			for _, poolVolSnap := range poolVol.VolumeSnapshots {
 				fullSnapName := drivers.GetSnapshotVolumeName(inst.Name(), poolVolSnap.Name)
 
-				// Copy volume config from backup file if present,
-				// so VolumeDBCreate can safely modify the copy if needed.
+				// Create new volume object to get a proper configuration for the driver in use.
```
tomponline, line 7481: > // Ensure driver specific default config is filled in new volume.

lxd/storage/backend_lxd.go
```diff
@@ -7503,8 +7499,7 @@ func (b *lxdBackend) ImportInstance(inst instance.Instance, poolVol *backupConfi
 			for _, i := range snapshots {
 				fullSnapName := i // Local var for revert.
 
-				// Copy the parent volume's config,
-				// so VolumeDBCreate can safely modify the copy if needed.
+				// Create new volume object to get a proper configuration for the driver in use.
```
tomponline, line 7502: > // Ensure driver specific default config is filled in new volume.

lxd/storage/drivers/driver_ceph_volumes.go
tomponline, line 1: Hrm I'd be hesitant to start forcefully killing ongoing storage commands that are mutating the local or remote storage layers during a clean shutdown. This feels like its asking for trouble to me.

Was there a reason you didn't go with `context.Background()` here to ensure the commands safely completed?

lxd/storage/drivers/driver_common.go
```diff
@@ -289,7 +289,7 @@ func (d *common) moveGPTAltHeader(devPath string) error {
 		return nil
 	}
 
-	_, err = shared.RunCommand(path, "--move-second-header", devPath)
+	_, err = shared.RunCommandContext(d.state.ShutdownCtx, path, "--move-second-header", devPath)
```
tomponline, line 292: i dont think we should use d.state.ShutdownCtx for anything mutating the disk, as Go will then forcefully kill the command, leaving it in an inconsistent state.

We've been bitten before by this in the past where we killed an ongoing ceph map command, which then left unrecoverable structures in the kernel and the system had to be rebooted.

tomponline (changes requested):
Lets avoid introducing potential new issues by forcefully canceling commands that are mutating on disk structures during a clean shutdown. Use `context.Background()` for these until such time as you're prepared to test each one for its resulting behaviour.

lxd/storage/backend_lxd.go
hamistao: This is aligned to my research as well, I am glad you agree with the proposal :D

lxd/storage/utils.go
hamistao: Indeed, I am taking some time to perform as many tests as I can to make sure that any new volume will always have their default config filled in by testing all execution branches. But I believe this will clear the path for a simple and efficient solution for reporting accurate volume sizes for all types of volumes.

lxd/storage/drivers/driver_common.go
hamistao: > We've been bitten before by this in the past where we killed an ongoing ceph map command, which then left unrecoverable structures in the kernel and the system had to be rebooted.

Always useful to know some history, thanks for the insight!
I will make the necessary updates.

lxd/storage/drivers/driver_ceph_volumes.go
hamistao: I just failed to consider the risks of stopping these commands midway. Considering that doing so would simply leave the storage layers unharmed is far too optimistic.

lxd/storage/utils.go
tomponline: > Such as not enabling to perform operations that require having a filled in volume config before create the DB volume unless one adds a redundant call to `FillVolumeConfig`

Yes this is why `GetNewVolume` came along.

lxd/storage/drivers/driver_common.go
tomponline: In general if a command isnt mutating something we can use d.state.shutdownCtx to exit more quickly, but if we're mutating something it'd be better to let it finish in my view (ofc if thats going to take a long time and some cleanup is going to happen on error, such as writing a large file out and then removing it on error in a defer then that would be different).

tomponline:
@hamistao shall we close this for now to declutter the PR list?

PR #14926, opened 2025-02-05
doc: update guide on moving instances and live migration

minaelee:
- Separated move and copy into own sections & moved text out of unnecessary annotation boxes
- Improved section on live-migrating VMs, including info about setting `migration.stateful` only while VM is stopped, and a syntax example
- Added brief section about cluster evacuate/restore for moving all instances temporarily

### DIFF
--- a/doc/howto/move_instances.md
+++ b/doc/howto/move_instances.md
@@ -100,3 +100,9 @@ With this configuration, LXD instructs CRIU to perform a series of memory dumps 
 After each dump, LXD sends the memory dump to the specified remote.
 In an ideal scenario, each memory dump will decrease the delta to the previous memory dump, thereby increasing the percentage of memory that is already synced.
 When the percentage of synced memory is equal to or greater than the threshold specified via {config:option}`instance-migration:migration.incremental.memory.goal`, or the maximum number of allowed iterations specified via {config:option}`instance-migration:migration.incremental.memory.iterations` is reached, LXD instructs CRIU to perform a final memory dump and transfers it.
+
+## Temporarily move all instances from a cluster member
+
+For LXD servers that are members of a cluster, you can use the evacuate and restore operations to temporarily move all instances from one cluster member to another. These operations can live-migrate eligible instances.
+
+For more information, see: {ref}`cluster-evacuate-restore`.
--- a/doc/howto/move_instances.md
+++ b/doc/howto/move_instances.md
@@ -14,27 +14,30 @@ Therefore, there is no direct equivalent for moving instances in the API or the 
 However, you can {ref}`export an instance <instances-backup-export-instance>` from one server and {ref}`import it <instances-backup-import-instance>` to another server.
 ```
 
+## Move instances
+
 To move an instance from one LXD server to another, use the [`lxc move`](lxc_move.md) command:
 
     lxc move [<source_remote>:]<source_instance_name> <target_remote>:[<target_instance_name>]
 
-```{note}
 When moving a container, you must stop it first.
 See {ref}`live-migration-containers` for more information.
 
 When moving a virtual machine, you must either enable {ref}`live-migration-vms` or stop it first.
-```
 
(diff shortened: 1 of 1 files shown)


doc/howto/move_instances.md
```diff
@@ -53,28 +56,32 @@ If you need to adapt the configuration for the instance to run on the target ser
...
-```
+  ```{note}
+  When {config:option}`instance-migration:migration.stateful` is enabled in LXD, virtiofs shares are disabled, and files are only shared via the 9P protocol. Consequently, guest OSes lacking 9P support, such as CentOS 8, cannot share files with the host unless stateful migration is disabled. Additionally, the `lxd-agent` will not function for these guests under these conditions.
+  ```
+
+- The {config:option}`device-disk-device-conf:size.state` of the virtual machine's root disk device must be set to at least the size of the virtual machine's {config:option}`instance-resource-limits:limits.memory` setting.
+
+  ```{note}
+  If you are using a shared storage pool like Ceph RBD to back your instance, you don't need to set {config:option}`device-disk-device-conf:size.state` to perform live migration.
```
roosterfish, line 81: Should this rather say "remote storage pool" instead of "shared storage pool"?

roosterfish (dismissed):

doc/howto/move_instances.md
```diff
@@ -53,28 +53,32 @@ If you need to adapt the configuration for the instance to run on the target ser
...
+  ```
 
-```{note}
-When {config:option}`instance-migration:migration.stateful` is enabled in LXD, virtiofs shares are disabled, and files are only shared via the 9P protocol. Consequently, guest OSes lacking 9P support, such as CentOS 8, cannot share files with the host unless stateful migration is disabled. Additionally, the `lxd-agent` will not function for these guests under these conditions.
-```
+  ```{note}
+  When {config:option}`instance-migration:migration.stateful` is enabled in LXD, virtiofs shares are disabled, and files are only shared via the 9P protocol. Consequently, guest OSes lacking 9P support, such as CentOS 8, cannot share files with the host unless stateful migration is disabled. Additionally, the `lxd-agent` will not function for these guests under these conditions.
+  ```
+
+- The {config:option}`device-disk-device-conf:size.state` of the virtual machine's root disk device must be set to at least the size of the virtual machine's {config:option}`instance-resource-limits:limits.memory` setting.
```
tomponline, line 78: This sentence is undermined somewhat by the note immediately following it.

This, in my view, introduces the possibility of confusion where a reader may just read the statement and not the note and get confused.

Can we qualify this statement by prefixing "When using a local pool, "

> When using a local pool, the {config:option}`device-disk-device-conf:size.state` of the virtual machine's root disk..."

Then the note below becomes an explanation of the statement rather than a contradiction to it.

doc/howto/move_instances.md
```diff
@@ -100,3 +100,9 @@ With this configuration, LXD instructs CRIU to perform a series of memory dumps
 After each dump, LXD sends the memory dump to the specified remote.
 In an ideal scenario, each memory dump will decrease the delta to the previous memory dump, thereby increasing the percentage of memory that is already synced.
 When the percentage of synced memory is equal to or greater than the threshold specified via {config:option}`instance-migration:migration.incremental.memory.goal`, or the maximum number of allowed iterations specified via {config:option}`instance-migration:migration.incremental.memory.iterations` is reached, LXD instructs CRIU to perform a final memory dump and transfers it.
+
+## Temporarily move all instances in a cluster member
```
tomponline, line 107: Can we say "from" rather than "in" here?

doc/howto/move_instances.md
minaelee: Done

doc/howto/move_instances.md
minaelee: Makes sense, done!

tomponline (approved):
Thanks!

PR #14927, opened 2025-02-06
Network: Improve ovn validation efficiency and logic

tomponline:
- Reduces the number of transactions used.
- Fixes validation of ovn uplink volatile IP to ensure its within ranges allowed from uplink.

### DIFF
--- a/lxd/network/driver_ovn.go
+++ b/lxd/network/driver_ovn.go
@@ -659,12 +659,6 @@ func (n *ovn) Validate(config map[string]string) error {
 		return err
 	}
 
-	// Get uplink routes.
-	uplinkRoutes, err := n.uplinkRoutes(uplink)
-	if err != nil {
-		return err
-	}
-
 	// Get project restricted routes.
 	projectRestrictedSubnets, err := n.projectRestrictedSubnets(p, uplink.Name)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


lxd/network/driver_ovn.go
```diff
@@ -676,18 +680,12 @@ func (n *ovn) Validate(config map[string]string) error {
...
-	var externalSubnets []*net.IPNet // Subnets to check for conflicts with other networks/NICs.
-	for _, keyPrefix := range []string{"ipv4", "ipv6"} {
-		addressKey := fmt.Sprintf("%s.address", keyPrefix)
-		netSubnet := netSubnets[addressKey]
-
-		if shared.IsFalseOrEmpty(config[fmt.Sprintf("%s.nat", keyPrefix)]) && netSubnet != nil {
-			// Add to list to check for conflicts.
-			externalSubnets = append(externalSubnets, netSubnet)
+			// If NAT disabled, record the external subnets that are being requested.
+			if shared.IsFalseOrEmpty(config[fmt.Sprintf("%s.nat", keyPrefix)]) {
```
hamistao: This is not a blocker, but should we use `keyPrefix+ ".nat"` instead of `fmt.Sprintf("%s.nat", keyPrefix)` to be compliant with our new standard of avoiding `fmt.Sprintf` when possible?

lxd/network/driver_ovn.go
```diff
@@ -673,19 +673,21 @@ func (n *ovn) Validate(config map[string]string) error {
...
-			netSubnets[addressKey] = ipNet
+		netSubnets[addressKey] = ipNet
 
-			// If NAT disabled, record the external subnets that are being requested.
-			if shared.IsFalseOrEmpty(config[fmt.Sprintf("%s.nat", keyPrefix)]) {
-				// Add to list to check for conflicts.
-				externalSubnets = append(externalSubnets, ipNet)
-			}
+		// If NAT disabled, record the external subnets that are being requested.
+		if shared.IsFalseOrEmpty(config[fmt.Sprintf("%s.nat", keyPrefix)]) {
```
hamistao: Same comment here regarding the usage of `fmt.Sprintf`

hamistao (approved):
Looks good to me, thanks!

PR #14929, opened 2025-02-06
lxd/device: Unmount instance disk devices

MggMuggins:
This is a follow-up from #14921; I missed using the specialized `Unmount*`. Working on tests in lxd-ci


tomponline (approved):
Ta

PR #14930, opened 2025-02-06
Storage: Attach VM snapshots as disk devices

MggMuggins:
Attach a snapshot of `v1` to another VM as a disk device using this device config:
```
devices:
  v1-snap0:
    pool: default
    source: v1
    source.type: virtual-machine
    source.snapshot: snap0
    type: disk
```

Attempting to attach custom volume snapshots using this API will fail. I am planning to complete that functionality next week.

~~Tests in the works for lxd-ci~~https://github.com/canonical/lxd-ci/pull/413

### DIFF
--- a/lxc/storage_volume.go
+++ b/lxc/storage_volume.go
@@ -163,6 +163,68 @@ func (c *cmdStorageVolume) parseVolumeWithPool(name string) (volumeName string, 
 	return fields[1], fields[0]
 }
 
+func cmdAttachArgsAsDevice(client lxd.InstanceServer, poolName string, flagTarget string, args []string) (devName string, device map[string]string, err error) {
+	volName, volType, snapshot := parseVolumeSnapshot("custom", args[1])
+	if volType != "custom" && volType != "virtual-machine" {
+		return "", nil, errors.New(i18n.G(`Only "custom" and "virtual-machine" volumes can be attached to instances`))
+	}
+
+	// Attach the volume
+	devPath := ""
+	if len(args) == 3 {
+		devName = args[1]
+	} else if len(args) == 4 {
+		// Use the provided target.
+		if flagTarget != "" && client.IsClustered() {
+			client = client.UseTarget(flagTarget)
+		}
+
+		vol, _, err := client.GetStoragePoolVolume(poolName, volType, volName)
+		if err != nil {
+			return "", nil, err
+		}
+
+		switch vol.ContentType {
+		case "block", "iso":
+			devName = args[3]
+		case "filesystem":
+			// If using a filesystem volume, the path must also be provided as the fourth argument.
+			if !strings.HasPrefix(args[3], "/") {
+				devPath = path.Join("/", args[3])
+			} else {
+				devPath = args[3]
+			}
+
+			devName = args[1]
+		default:
+			return "", nil, errors.New(i18n.G("Unsupported content type for attaching to instances"))
+		}
+	} else if len(args) == 5 {
+		// Path and device name have been given to us.
+		devName = args[3]
+		devPath = args[4]
+	}
+
+	// Prepare the instance's device entry
+	device = map[string]string{
+		"type":   "disk",
+		"pool":   poolName,
+		"source": volName,
+		"path":   devPath,
+	}
+
+	// Only specify sourcetype when not the default
+	if volType != "custom" {
+		device["source.type"] = volType
+	}
+
+	if snapshot != "" {
+		device["source.snapshot"] = snapshot
+	}
+
+	return devName, device, nil
+}
+
 // Attach.
 type cmdStorageVolumeAttach struct {
 	global        *cmdGlobal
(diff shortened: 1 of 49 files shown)


tomponline:
@MggMuggins when we added `source-type` field it looks like we didnt update the `instance_root_volume_attachment` API extension, as it still says "Introduces the `<type>/<volume>` syntax for the `source` property of
disk devices."

Also can you update that API extension to include the new `snapshot` field too? Then that single API extension can cover both new features.

MggMuggins:
Good catch; should be good to go.

minaelee (dismissed):

MggMuggins:
Still chasing down a bug, this is **not** GTM.

tomponline:
@MggMuggins please can you rebase

MggMuggins:
Spun my wheels a bit today on checking that deleting a snapshot isn't possible if it's attached to an instance. Added a test to the lxd-ci PR to check that; this is good for review.

Still working on the LVM refcount bug. When the LVM driver [deactivates a volume snapshot](https://github.com/canonical/lxd/blob/main/lxd/storage/drivers/driver_lvm_utils.go#L817) it checks to see if the parent volume is still in use. We need the inverse; the driver also needs to make sure when deactivating the parent that there aren't any snapshots in use. It looks like deactivating the parent also deactivates the snapshot (hence the error below), but I doubt the opposite is true. More on this tomorrow.

Putting this here for future travelers:
```
$ lxc storage volume detach vmpool292382 virtual-machine/v2 v1
Error: Failed to stop device "v2-root": Failed to deactivate LVM logical volume "/dev/vmpool292382/virtual-machines_test_v2": Failed to run: lvchange --activate n --ignoreactivationskip /dev/vmpool292382/virtual-machines_test_v2: exit status 5 (Logical volume vmpool292382/virtual-machines_test_v2-snap0 contains a filesystem in use.
```

tomponline (commented):
@MggMuggins did you consider supporting `source: vol/snap` rather than `source-snapshot`? As we tend to use vol/snap format most other places?

lxc/storage_volume.go
tomponline, line 1: nice catch!

lxd/storage/utils.go
tomponline, line 1: >  Checks for the removal of security.shared and security.protection.start will no longer identify a snapshot as its parent volume

Doesnt seem quite relevant to this commit?

lxd/storage/utils.go
```diff
@@ -969,12 +969,25 @@ func volumeIsUsedByDevice(vol api.StorageVolume, inst *db.InstanceArgs, dev map[
...
 
+	var volName string
+	var snapName string
+
+	if shared.IsSnapshot(vol.Name) {
+		parts := strings.SplitN(vol.Name, shared.SnapshotDelimiter, 2)
+		volName, snapName = parts[0], parts[1]
+	} else if dev["source.snapshot"] != "" {
+		// vol is not a snapshot but dev refers to one
+		return false, nil
```
tomponline, line 980: shouldnt this be an error?

MggMuggins: `volumeIsUsedByDevice` is just determining if `dev` describes `vol`; if `vol` isn't a snapshot but `dev` describes one, then it was correct use of the function, just the two don't match. For example, the function should return false if it gets a storage volume `virtual-machine/v1` but this disk device:
```
  v1-snap0:
    type: disk
    pool: default
    source: v1
    source.type: virtual-machine
    source.snapshot: snap0
```

lxd/storage/utils.go
MggMuggins: I'll rewrite the commit message

tomponline:
> @MggMuggins did you consider supporting `source: vol/snap` rather than `source-snapshot`? As we tend to use vol/snap format most other places?

Just want to chexk u didnt miss this question?

MggMuggins:
I did not :wink: 

[Here](https://github.com/canonical/lxd/compare/main...MggMuggins:snapshot-disk-devices-separator?expand=1) is this set of changes with that syntax instead (tests failing atm, def broken). From an implementation perspective it makes complete sense (much smaller delta with the existing disk device driver and lxc, consistent with `StorageVolume.Name`, etc); and there are a few places where `<instance>/<snap>` synxtax is exposed in the API.

I don't love it for the same reasons that I switched to `source.type` though: multiple discrete pieces of information in the same field is very opaque when reading the code. It also requires parsing the field before using it (even if, as is the case here, my changes aren't doing the parsing). Using `source.snapshot` is more of an ugly band-aid than a real fix though, so I'm willing to go with the delimited syntax if that's what you'd prefer. I will get that other branch in order tomorrow.

Fixed 2 bugs (ceph `ro` block vols and docs) in the process :slightly_smiling_face:

tomponline:
> Using `source.snapshot` is more of an ugly band-aid than a real fix though, so I'm willing to go with the delimited syntax if that's what you'd prefer. I will get that other branch in order tomorrow.


How about we stick with `source.snapshot` because it makes the config and code itself clearer, and we can then still use vol/snapshot in the CLI if needed (like volume attach)?

MggMuggins:
I'm happy with that; the `[<type>/]<vol>[/<snapshot>]` syntax will stick around regardless.

I'll get this ready to go ASAP.

MggMuggins:
@tomponline This is ready to go.

Added tests for readonly ceph block devices to https://github.com/canonical/lxd-ci/pull/413

MggMuggins:
Nevermind, the readonly fix regressed something else and I don't have time today to dig into it. I'll pull those commits out so that this PR is unblocked and keep working on it on Monday.

lxc/storage_volume.go
```diff
@@ -51,6 +51,17 @@ func parseVolume(defaultType string, name string) (volName string, volType strin
 	return volName, volType
 }
 
+func parseVolumeSnapshot(defaultType string, name string) (volName string, volType string, snapshot string) {
+	volName, volType = parseVolume(defaultType, name)
+
+	parts := strings.SplitN(volName, "/", 2)
```
tomponline, line 57: strings.Cut would also be an option here

tomponline:
@MggMuggins once you've made the API extension change, please can you also rebase and then this should be good to merge and unblock LXD 6.3, thanks!

MggMuggins:
@tomponline This is all set; I just ran a quick check of lxd-ci tests as well.

tomponline (approved):
Thanks!

PR #14935, opened 2025-02-07
OIDC: Make scopes configurable

### LINKED ISSUE #14141: OIDC - oidc.groups.claim is added as additional scope which breaks Microsoft Entra integration

markylaing:
In the end, just having `oidc.scopes` was the way to go (see https://github.com/canonical/lxd/pull/14888#issuecomment-2642783581). I've added tests for the validation of the config key and also tested it locally with Keycloak by disabling the `profile` claim.

Closes #14141.

### DIFF
--- a/client/lxd_oidc.go
+++ b/client/lxd_oidc.go
@@ -3,6 +3,7 @@ package lxd
 import (
 	"context"
 	"crypto/rand"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -65,7 +66,6 @@ func (o *oidcTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 }
 
 var errRefreshAccessToken = fmt.Errorf("Failed refreshing access token")
-var oidcScopes = []string{oidc.ScopeOpenID, oidc.ScopeOfflineAccess, oidc.ScopeEmail, oidc.ScopeProfile}
 
 type oidcClient struct {
 	httpClient    *http.Client
@@ -101,7 +101,8 @@ func (o *oidcClient) getAccessToken() string {
 
 // do function executes an HTTP request using the oidcClient's http client, and manages authorization by refreshing or authenticating as needed.
 // If the request fails with an HTTP Unauthorized status, it attempts to refresh the access token, or perform an OIDC authentication if refresh fails.
-func (o *oidcClient) do(req *http.Request) (*http.Response, error) {
+// The oidcScopesExtensionPresent argument changes the behaviour of this function based on the presence of an API extension.
+func (o *oidcClient) do(req *http.Request, oidcScopesExtensionPresent bool) (*http.Response, error) {
 	resp, err := o.httpClient.Do(req)
 	if err != nil {
 		return nil, err
@@ -115,11 +116,32 @@ func (o *oidcClient) do(req *http.Request) (*http.Response, error) {
 	issuer := resp.Header.Get("X-LXD-OIDC-issuer")
 	clientID := resp.Header.Get("X-LXD-OIDC-clientid")
 	audience := resp.Header.Get("X-LXD-OIDC-audience")
-	groupsClaim := resp.Header.Get("X-LXD-OIDC-groups-claim")
 
-	err = o.refresh(issuer, clientID, groupsClaim)
+	var scopes []string
+	if oidcScopesExtensionPresent {
+		// If we have the `oidc_scopes` extension, get the scopes from the header and ignore the groups claim header.
+		scopesJSON := resp.Header.Get("X-LXD-OIDC-scopes")
+		if scopesJSON == "" {
+			return nil, fmt.Errorf("LXD server did not return OIDC scopes")
+		}
+
+		err = json.Unmarshal([]byte(scopesJSON), &scopes)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to parse OIDC scopes: %w", err)
+		}
+	} else {
+		// Otherwise, use the default scopes from before the API extension was added, and append the groups claim header
+		// if set.
+		scopes = []string{oidc.ScopeOpenID, oidc.ScopeEmail, oidc.ScopeOfflineAccess, oidc.ScopeProfile}
+		groupsClaim := resp.Header.Get("X-LXD-OIDC-groups-claim")
+		if groupsClaim != "" {
+			scopes = append(scopes, groupsClaim)
+		}
+	}
+
+	err = o.refresh(issuer, clientID, scopes)
 	if err != nil {
-		err = o.authenticate(issuer, clientID, audience, groupsClaim)
+		err = o.authenticate(issuer, clientID, audience, scopes)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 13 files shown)


client/lxd_oidc.go
```diff
@@ -115,11 +115,22 @@ func (o *oidcClient) do(req *http.Request) (*http.Response, error) {
 	issuer := resp.Header.Get("X-LXD-OIDC-issuer")
 	clientID := resp.Header.Get("X-LXD-OIDC-clientid")
 	audience := resp.Header.Get("X-LXD-OIDC-audience")
```
markylaing, line 117: @tomponline I've never liked that we use headers for this. The API documentation for the [`/1.0?public` endpoint](https://documentation.ubuntu.com/lxd/en/latest/api/#/server/server_get_untrusted) says:

> Shows a small subset of the server environment and configuration which is required by untrusted clients to reach a server.

IMO we should have a small subset of config available over /1.0 for untrusted clients, including these values.

doc/api-extensions.md
```diff
@@ -2505,6 +2505,7 @@ And lastly, adds a `project` field on leases, leases can be retrieved via `/1.0/
 
 ## `network_ovn_uplink_vlan`
 
+<<<<<<< HEAD
```
simondeziel, line 2508: Conflict resolution leftover ;)

lxd/auth/oidc/oidc.go
```diff
@@ -453,12 +463,7 @@ func (o *Verifier) setRelyingParty(ctx context.Context, host string) error {
 		rp.WithHTTPClient(httpClient),
 	}
 
-	oidcScopes := []string{oidc.ScopeOpenID, oidc.ScopeOfflineAccess, oidc.ScopeEmail, oidc.ScopeProfile}
-	if o.groupsClaim != "" {
-		oidcScopes = append(oidcScopes, o.groupsClaim)
-	}
-
-	relyingParty, err := rp.NewRelyingPartyOIDC(ctx, o.issuer, o.clientID, "", fmt.Sprintf("https://%s/oidc/callback", host), oidcScopes, options...)
+	relyingParty, err := rp.NewRelyingPartyOIDC(ctx, o.issuer, o.clientID, "", fmt.Sprintf("https://%s/oidc/callback", host), o.scopes, options...)
```
simondeziel, line 466:
```suggestion
	relyingParty, err := rp.NewRelyingPartyOIDC(ctx, o.issuer, o.clientID, "", "https://"+host+"/oidc/callback", o.scopes, options...)
```

test/suites/oidc.sh
```diff
@@ -4,6 +4,14 @@ test_oidc() {
   # shellcheck disable=2153
   ensure_has_localhost_remote "${LXD_ADDR}"
 
+  # Check OIDC scopes validation
+  ! lxc config set oidc.scopes "my-scope" || false # Doesn't contain "email" or "openid"
+  ! lxc config set oidc.scopes "my-scope email" || false # Doesn't contain "openid"
+  ! lxc config set oidc.scopes "my-scope openid" || false # Doesn't contain "openid"
```
simondeziel, line 10:
```suggestion
  ! lxc config set oidc.scopes "my-scope openid" || false # Doesn't contain "email"
```

doc/api-extensions.md
```diff
@@ -2599,3 +2600,10 @@ The following pool level configuration keys have been added:
 1. {config:option}`storage-pure-pool-conf:pure.api.token`
 1. {config:option}`storage-pure-pool-conf:pure.mode`
 1. {config:option}`storage-pure-pool-conf:pure.target`
+
+## `oidc_scopes`
+
+This API extension enables setting an {config:option}`server-oidc:oidc.scopes` configuration key which accepts a space separated list of OIDC scopes to request from the identity provider.
```
minaelee, line 2606:
```suggestion
This API extension enables setting an {config:option}`server-oidc:oidc.scopes` configuration key, which accepts a space-separated list of OIDC scopes to request from the identity provider.
```

markylaing: Fixed, thanks :)

test/suites/oidc.sh
markylaing: Fixed.

lxd/auth/oidc/oidc.go
markylaing: Fixed.

doc/api-extensions.md
markylaing: Whoopsie!

lxd/cluster/config/config.go
```diff
@@ -679,11 +680,33 @@ var ConfigSchema = config.Schema{
...
+	// A list of OpenID Connect scopes to request from the identity provider.
+	// This must include the `openid` and `email` scopes.
+	// The remaining optional scopes are `offline_access` and `profile`.
+	// If you remove the `offline_access` scope, users might be required to log in more frequently.
+	// If you remove the `profile` scope, user information may not be displayed in LXD UI (or in `lxc auth identity` commands).
+	// You may add additional scopes if this is required by your identity provider, or if necessary for configuration of {ref}`identity provider groups <identity-provider-groups>`.
+	// ---
+	//  type: space-delimited string
+	//  scope: global
+	//  shortdesc: Space separated list of OpenID Connect scopes
```
minaelee, line 693:
```suggestion
	//  shortdesc: Space-separated list of OpenID Connect scopes
```

simondeziel (dismissed):

lxd/cluster/config/config.go
markylaing: Fixed.

markylaing:
@tomponline this is ready when you have time. Thanks.

client/lxd_oidc.go
tomponline, line 1: @markylaing as discussed in order to maintain backwards and forwards compat between client and server, lets:

1. Have the server continue to send the `X-LXD-OIDC-groups-claim` header - so that old clients continue to request this as a scope.
2. Have the new client detect that the server supports the new API extension and disregard the `X-LXD-OIDC-groups-claim` header for scopes and instead use `X-LXD-OIDC-scopes`.

Update commit to remove references to `X-LXD-OIDC-additional-scopes` too please.

markylaing: I've implemented this now.

markylaing:
@tomponline ready for another round when you're ready. I realised that I forgot to cherry-pick the patch from #14888 so I've added it now and also added a unit test for that.

lxd/patches.go
```diff
@@ -1445,4 +1447,40 @@ DELETE FROM storage_pools_config WHERE key = "powerflex.sdt"
 	return err
 }
 
+// patchOIDCGroupsClaimScope adds the contents of oidc.groups.claim to the new configuration for oidc.scopes if present.
+// The oidc.groups.claim value was initially added to scopes but shouldn't have been. This patch will allow users with
+// working identity provider group mappings to continue using them by continuing to request the claim as an additional
+// scope.
+func patchOIDCGroupsClaimScope(_ string, d *Daemon) error {
```
tomponline, line 1454: @markylaing this function will be run on each cluster member, at different times.

I notice that it is loading the cluster config from the DB inside the function, and then modifying the global config.

But is it safe when potentially this function is run again on a different cluster member after the global config has been modified by this function at an earlier stage?

tomponline (commented):
one comment

lxd/patches.go
markylaing: Ahh of course :facepalm: will fix

markylaing: Fixed. Thanks for catching :)

tomponline (approved):

PR #14937, opened 2025-02-08
doc: fix looping local sphinx-autobuild

minaelee:
This overrides the behavior of `make run` to `--ignore` the manpages. It was infinitely looping after a single change was made due to the change triggering a build, which would auto-generate manpages, which would trigger another build, which would auto-generate manpages, and so on. Only affects local builds for testing (`make run` in doc folder builds, serves, and watches for changes locally).


tomponline (approved):
Ta

PR #14958, opened 2025-02-11
lxd/instance/drivers/qemu: Fix double-escaped device name

MggMuggins:
Volumes with a `-` in their name would fail to detach as the device name used at device removal was different than the one used at creation.

This only manifested in the LVM driver, although I didn't check zfs.

Fixes the failures from https://github.com/canonical/lxd-ci/pull/409

### DIFF
--- a/lxd/instance/drivers/driver_qemu.go
+++ b/lxd/instance/drivers/driver_qemu.go
@@ -4136,7 +4136,7 @@ func (d *qemu) addDriveConfig(qemuDev map[string]string, bootIndexes map[string]
 		revert := revert.New()
 		defer revert.Fail()
 
-		nodeName := qemuDeviceNameOrID(qemuDeviceNamePrefix, escapedDeviceName, "", qemuDeviceNameMaxLength)
+		nodeName := qemuDeviceNameOrID(qemuDeviceNamePrefix, driveConf.DevName, "", qemuDeviceNameMaxLength)
 
 		if isRBDImage {
 			secretID := fmt.Sprintf("pool_%s_%s", blockDev["pool"], blockDev["user"])


hamistao (approved):
Thanks for this!
Not related to this PR and this would have to be confirmed, but some of the previous uses of `escapedDeviceName` could potentially be improved with `qemuDeviceNameOrID` as well.

tomponline:
Thanks @MggMuggins 

But im confused why devname here is already escaped, as that is coming from the devices config. So at first glance it should need escaping here too.

@hamistao i thought you had gone through the device naming escaping a while back?

Cc @kadinsayani

tomponline:
@MggMuggins also im not seeing this in main branch when adding/removing a block disk device for a custom volume;

```
lxc config device add v1 foo-bah disk source=foo pool=default
lxc config device remove v1 foo-bah
```


```
Adding logging of args to `monitor.AddBlockDevice`:
ERROR  [2025-02-11T08:20:21Z] tomp1                                         blockDev="map[aio:io_uring cache:map[direct:true no-flush:false] discard:unmap driver:file filename:/dev/fdset/1 locking:off node-name:lxd_foo--bah read-only:false]" device="map[bus:qemu_scsi.0 channel:0 drive:lxd_foo--bah driver:scsi-hd id:dev-lxd_foo--bah lun:1 serial:lxd_foo--bah]"

Adding logging of arg to `RemoveBlockDevice`:
ERROR  [2025-02-11T08:20:45Z] tomp2                                         blockDevName=lxd_foo--bah
```

tomponline:
Oh I see now, its not the block device itself, but the name of the filehandles via `SendFileWithFDSet` and `RemoveFDFromFDSet` that is at fault. Got you.

tomponline:
@MggMuggins please can you backport this to stable-5.21 branch

tomponline:
> Thanks for this! Not related to this PR and this would have to be confirmed, but some of the previous uses of `escapedDeviceName` could potentially be improved with `qemuDeviceNameOrID` as well.

@hamistao please can you go through and see if there are other issues related to this if you get chance between your ssh roadmap item, thanks.

tomponline:
Also @MggMuggins i noticed in my debug messages above that `RemoveBlockDevice` seems to be called twice on removal of a disk device, thats probably worth investigating too.

hamistao:
@tomponline I have a task for this purpose on my backlog, but perhaps I can arrange some time this week

tomponline:
> @tomponline I have a task for this purpose on my backlog, but perhaps I can arrange some time this week

Only if u get chance though, ta

PR #14959, opened 2025-02-11
cloud-init: Support injection of additional SSH keys on instances

hamistao:
This adds supports for addition of SSH keys on instances in such a way that neither `cloud-init.vendor-data` and `cloud-init.user-data` are not overwritten.

This is done through the addition of the config key `cloud-init.ssh-keys.KEYNAME`, where KEYNAME can be any arbitrary name for the added key. The content of this key is merged into any pre-existing `cloud-config` defined in either `cloud-init.vendor-data` or `cloud-init.user-data`.

LXD CI tests in https://github.com/canonical/lxd-ci/pull/412

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2599,3 +2599,11 @@ The following pool level configuration keys have been added:
 1. {config:option}`storage-pure-pool-conf:pure.api.token`
 1. {config:option}`storage-pure-pool-conf:pure.mode`
 1. {config:option}`storage-pure-pool-conf:pure.target`
+
+## `cloud_init_ssh_keys`
+
+Adds support for injecting additional SSH public keys into instances through {ref}`cloud-init <instance-options-cloud-init>` without conflicting with any configuration present on {config:option}`instance-cloud-init:cloud-init.vendor-data` or {config:option}`instance-cloud-init:cloud-init.user-data`.
+
+To achieve this, the `cloud-init.ssh-keys.KEYNAME` configuration key is added for both instances and profiles. This key is used to define a public key to be injected. `KEYNAME` can be any arbitrary name for the injected key.
+
+The value for `cloud-init.ssh-keys.KEYNAME` should be `<user>:<key>`, where `<user>` is the name of the user for whom to inject the key. For `<key>`, provide either the public key or a `cloud-init` import ID for a key hosted elsewhere. Example valid values for `cloud-init.ssh-keys.KEYNAME` are `root:gh:githubUser` or `myUser:ssh-keyAlg base64PublicKey`.
--- a/doc/cloud-init.md
+++ b/doc/cloud-init.md
@@ -296,3 +296,33 @@ config:
       - type: nameserver
         address: 10.10.10.254
 ```
+
+## How to inject SSH keys into instances
+
+To inject SSH keys into LXD instances for an arbitrary user, use the configuration key `cloud-init.ssh-keys.<keyName>`.
+
+Use the format `<user>:<key>` for its value, where `<user>` is a Linux username and `<key>` can be either a pure SSH public key or an import ID for a key hosted elsewhere. For example, `root:gh:githubUser` and `myUser:ssh-keyAlg publicKeyHash` are valid values.
+
+`cloud-init.ssh-keys.<keyName>` cannot be applied if LXD is unable to parse both the existing `cloud-config.vendor-data` and `cloud-config.user-data` for that instance. This could happen, for example, if those keys contain badly formatted YAML.
+
+You can define SSH keys via `cloud-init.vendor-data` or `cloud-init.user-data` directly. Keys defined using `cloud-init.ssh-keys.<keyName>` do not conflict with those defined with either `cloud-init.vendor-data` or `cloud-init.user-data` in any way. For more information on how to use `cloud-config` to define SSH keys, see [the cloud-init docs for SSH configuration](https://cloudinit.readthedocs.io/en/latest/reference/yaml_examples/ssh.html).
+
+Since `cloud-init` only runs on instance start, updates to `cloud-init.*` keys on a running instance only take effect after restart.
+
+### Examples
+
+The following command injects `someuser`'s key from Launchpad into the newly created `container`:
+
+```bash
+lxc launch ubuntu:24.04 container -c cloud-init.ssh-keys.mykey root:lp:someuser
+```
+
+The example profile configuration below defines a key to be injected on an instance. The injected key enables the owner of the private key to SSH into the instance as a user named `user`:
+
+```yaml
+config:
+  cloud-init.vendor-data: |
+    users:
+      - name: user
+        ssh_authorized_keys: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIJFDWcYmMrCZdk9JI29bAiHKD90oEUr8tqK5VvoO8Vcj
+```
--- a/doc/metadata.txt
+++ b/doc/metadata.txt
@@ -1716,6 +1716,17 @@ The instance with the highest value is shut down first.
 The content is used as seed value for `cloud-init`.
 ```
 
+```{config:option} cloud-init.ssh-keys.KEYNAME instance-cloud-init
+:condition: "If supported by image"
+:liveupdate: "no"
+:shortdesc: "Additional SSH key to be injected on the instance by `cloud-init`"
+:type: "string"
+Represents an additional SSH public key to be merged into existing `cloud-init` seed data
+and injected into an instance. Has the format `{user}:{key}`, where {user} is a Linux username and
+{key} can be either a pure SSH public key or an import ID for a key hosted elsewhere.
+// For example: `root:gh:githubUser`, `myUser:ssh-keyAlg publicKeyHash`
+```
+
 ```{config:option} cloud-init.user-data instance-cloud-init
 :condition: "If supported by image"
 :defaultdesc: "`#cloud-config`"
--- a/lxd/devlxd.go
+++ b/lxd/devlxd.go
@@ -25,6 +25,7 @@ import (
 	"github.com/canonical/lxd/lxd/response"
 	"github.com/canonical/lxd/lxd/state"
 	"github.com/canonical/lxd/lxd/ucred"
+	"github.com/canonical/lxd/lxd/util"
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/logger"
@@ -97,8 +98,18 @@ func devlxdConfigKeyGetHandler(d *Daemon, c instance.Instance, w http.ResponseWr
 		return response.DevLxdErrorResponse(api.StatusErrorf(http.StatusForbidden, "not authorized"), c.Type() == instancetype.VM)
 	}
 
-	value, ok := c.ExpandedConfig()[key]
-	if !ok {
+	value := c.ExpandedConfig()[key]
+
+	// If parsing the config is not possible, abstain from merging the additional keys into the config.
+	if strings.HasSuffix(key, ".vendor-data") || strings.HasSuffix(key, ".user-data") {
+		value, err = util.MergeSSHKeyCloudConfig(c.ExpandedConfig(), value)
+		if err != nil {
+			logger.Warn("Failed merging SSH keys into cloud-init seed data, abstain from injecting additional keys", logger.Ctx{"err": err, "project": c.Project(), "instance": c.Name(), "requestedKey": key})
+		}
+	}
+
+	// If the requested key is not defined and there isn't any addition SSH keys for this instance, return a 'not found' response.
+	if value == "" {
 		return response.DevLxdErrorResponse(api.StatusErrorf(http.StatusNotFound, "not found"), c.Type() == instancetype.VM)
 	}
 
--- a/lxd/devlxd.go
+++ b/lxd/devlxd.go
@@ -70,10 +70,28 @@ func devlxdConfigGetHandler(d *Daemon, c instance.Instance, w http.ResponseWrite
 	}
 
 	filtered := []string{}
+	hasSSHKeys := false
+	hasCustomConfig := false
 	for k := range c.ExpandedConfig() {
-		if strings.HasPrefix(k, "user.") || strings.HasPrefix(k, "cloud-init.") {
+		// cloud-init.ssh-keys keys should not be retireved by cloud-init directly but instead should be merged into
+		// cloud-init.vendor-data and/or cloud-init.user-data.
+		if strings.HasPrefix(k, "cloud-init.ssh-keys.") {
+			hasSSHKeys = true
+		} else if strings.HasPrefix(k, "user.") || strings.HasPrefix(k, "cloud-init.") {
+			if strings.HasSuffix(k, ".vendor-data") || strings.HasSuffix(k, ".user-data") {
+				hasCustomConfig = true
+			}
+
 			filtered = append(filtered, "/1.0/config/"+k)
 		}
+	}
+
+	// If the instance has SSH keys to be injected into it and do not have any custom seed data for cloud-init defined,
+	// include cloud-init.vendor-data in the response, since the SSH keys will be included in it.
+	// We are including vendor-data because it makes more sense conceptually, as the vendor (LXD) is generating the
+	// configuration for `cloud-init` to fetch.
+	if hasSSHKeys && !hasCustomConfig {
+		filtered = append(filtered, "/1.0/config/cloud-init.vendor-data")
 	}
 
 	return response.DevLxdResponse(http.StatusOK, filtered, "json", c.Type() == instancetype.VM)
--- a/lxd/instance/instancetype/instance.go
+++ b/lxd/instance/instancetype/instance.go
@@ -29,7 +29,7 @@ const (
 const ConfigVolatilePrefix = "volatile."
 
 // ConfigKeyPrefixesAny indicates valid prefixes for configuration options.
-var ConfigKeyPrefixesAny = []string{"environment.", "user.", "image."}
+var ConfigKeyPrefixesAny = []string{"environment.", "user.", "image.", "cloud-init.ssh-keys."}
 
 // ConfigKeyPrefixesContainer indicates valid prefixes for container configuration options.
 var ConfigKeyPrefixesContainer = []string{"linux.sysctl.", "limits.kernel."}
--- a/lxd/instance/instancetype/instance.go
+++ b/lxd/instance/instancetype/instance.go
@@ -1137,6 +1137,21 @@ func ConfigKeyChecker(key string, instanceType Type) (func(value string) error, 
 		}
 	}
 
+	// lxdmeta:generate(entities=instance; group=cloud-init; key=cloud-init.ssh-keys.KEYNAME)
+	// Represents an additional SSH public key to be merged into existing `cloud-init` seed data
+	// and injected into an instance. Has the format `{user}:{key}`, where {user} is a Linux username and
+	// {key} can be either a pure SSH public key or an import ID for a key hosted elsewhere.
+	// // For example: `root:gh:githubUser`, `myUser:ssh-keyAlg publicKeyHash`
+	// ---
+	//  type: string
+	//  liveupdate: no
+	//  condition: If supported by image
+	//  shortdesc: Additional SSH key to be injected on the instance by `cloud-init`
+	sshKeyName := strings.TrimPrefix(key, "cloud-init.ssh-keys.")
+	if sshKeyName != key && sshKeyName != "" {
+		return validate.Optional(validate.IsUserSSHKey), nil
+	}
+
 	if strings.HasPrefix(key, ConfigVolatilePrefix) {
 		// lxdmeta:generate(entities=instance; group=volatile; key=volatile.<name>.last_state.hwaddr)
 		// The original MAC that was used when moving a physical device into an instance.
--- a/lxd/device/disk.go
+++ b/lxd/device/disk.go
@@ -27,6 +27,7 @@ import (
 	storagePools "github.com/canonical/lxd/lxd/storage"
 	storageDrivers "github.com/canonical/lxd/lxd/storage/drivers"
 	"github.com/canonical/lxd/lxd/storage/filesystem"
+	"github.com/canonical/lxd/lxd/util"
 	"github.com/canonical/lxd/lxd/warnings"
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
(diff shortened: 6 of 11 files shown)


hamistao:
This feature can and will be improved as the current implementation does not parse `cloud-config` in JSON format, and this can be an issue if both `cloud-init.vendor-data` and `cloud-init.user-data` are defined in this format. The YAML formatting is far more common than JSON, so this is why YAML is included here.

Also, `vendor-data` and `user-data` can take the form of commands to be run on the instance. But this also would only be a problem if both `cloud-init.vendor-data` and `cloud-init.user-data` are commands and this can also be worked around by replacing the command for a `cloud-config` and inserting the command as a `rumcmd` field on the resulting YAML, keeping the intended behavior.

doc/api-extensions.md
```diff
@@ -2599,3 +2599,9 @@ The following pool level configuration keys have been added:
 1. {config:option}`storage-pure-pool-conf:pure.api.token`
 1. {config:option}`storage-pure-pool-conf:pure.mode`
 1. {config:option}`storage-pure-pool-conf:pure.target`
+
+## `cloud_init_ssh_keys`
+
+Adds support for injecting additional SSH public keys on instances through `cloud-init` without conflicting with any configuration present on `cloud-init.vendor-data` or `cloud-init.user-data`.
+
+To achieve this, the `cloud-init.ssh-keys.KEYNAME` configuration key is added for both instances and profiles, that can be used for specifying one desired additional public key to be injected. Where `KEYNAME` can be any arbitrary name for the injected key.
```
tomponline, line 2607: Please can you document here what the value of this key should be too (as well as in the docs).

lxd/instance/instancetype/instance.go
```diff
@@ -1137,6 +1137,20 @@ func ConfigKeyChecker(key string, instanceType Type) (func(value string) error,
 		}
 	}
 
+	// lxdmeta:generate(entities=instance; group=cloud-init; key=cloud-init.ssh-keys.KEYNAME)
+	// Represent an additional SSH public key to be merged into existent `cloud-init` seed data
```
tomponline, line 1141: existing not existent

lxd/instance/instancetype/instance.go
```diff
@@ -1137,6 +1137,20 @@ func ConfigKeyChecker(key string, instanceType Type) (func(value string) error,
...
+	// and injected on an instance. Has the format `{user}:{key}`, where user is a Linux {user} and
```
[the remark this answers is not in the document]
tomponline: > where user is a Linux {user} 

Can we say "username" and lose the brackets.

lxd/instance/instancetype/instance.go
```diff
@@ -1137,6 +1137,20 @@ func ConfigKeyChecker(key string, instanceType Type) (func(value string) error,
...
+	// {key} can be either a pure SSH public keys or import IDs for keys hosted elsewhere.
+	// // For example: `root:gh:githubUser` or `myUser:ssh-keyAlg publicKeyHash`
+	// ---
+	//  type: string
+	//  liveupdate: no
+	//  condition: If supported by image
+	//  shortdesc: Additional SSH key to be injected on the instance by `cloud-init`
+	if strings.HasPrefix(key, "cloud-init.ssh-keys.") && strings.TrimPrefix(key, "cloud-init.ssh-keys.") != "" {
```
tomponline, line 1150: Would strings.Cut be more appropriate here?


```go
prefix, suffix, found := strings.Cut(key, "cloud-init.ssh-keys.")
if found && prefix == "" && suffix != "" {
}
```

go.mod
```diff
@@ -65,6 +65,7 @@ require (
 	google.golang.org/protobuf v1.36.4
 	gopkg.in/tomb.v2 v2.0.0-20161208151619-d5d1b5820637
 	gopkg.in/yaml.v2 v2.4.0
```
tomponline, line 67: @hamistao we should aim to only use a single version of the yaml parser as a direct depedency

lxd/devlxd.go
```diff
@@ -560,3 +561,161 @@ func findContainerForPid(pid int32, s *state.State) (instance.Container, error)
 
 	return nil, errPIDNotInContainer
 }
+
+// userKeys is a struct that keeps the keys to be injected to allow SSHing as a certain user.
```
tomponline, line 573: no mention of cloud init here, i think it should have something, maybe `cloudInitUserSSHKeys`?

lxd/devlxd.go
```diff
@@ -560,3 +561,161 @@ func findContainerForPid(pid int32, s *state.State) (instance.Container, error)
...
+				continue
+			}
+
+			// Create an empty userKeys if the user is not configured.
+			_, ok := users[user]
+			if !ok {
+				users[user] = &userKeys{}
+			}
+
+			// Check if ssh key is an import ID with with a format "keyServer:UserName".
```
tomponline, line 602: @hamistao based on our chat, can you expand the comment here explaining the logic of the validation, ta

lxd/devlxd.go
```diff
@@ -560,3 +561,161 @@ func findContainerForPid(pid int32, s *state.State) (instance.Container, error)
...
+			if strings.Contains(sshKey, ":") && !strings.Contains(sshKey, " ") {
+				users[user].importIDs = append(users[user].importIDs, sshKey)
+				continue
+			}
+
+			users[user].publicKeys = append(users[user].publicKeys, sshKey)
+		}
+	}
+
+	// If no additional keys are defined, return the original config passed in.
```
tomponline, line 612: "additional" seems misleading here, additional to what? I would suggest we just say "no keys".

lxd/devlxd.go
```diff
@@ -560,3 +561,161 @@ func findContainerForPid(pid int32, s *state.State) (instance.Container, error)
...
+	}
+
+	// Use map[string]any to allow us to parse any YAML.
+	cloudConfigMap := make(map[string]any)
+	err := yaml.Unmarshal([]byte(cloudConfig), cloudConfigMap)
+	if err != nil {
+		return cloudConfig, fmt.Errorf("Could not unmarshall cloud-config: %w", err)
+	}
+
+	// Get previously defined user sequence in provided config, if present.
```
tomponline, line 624: lets say "users list"

lxd/devlxd.go
```diff
@@ -560,3 +561,161 @@ func findContainerForPid(pid int32, s *state.State) (instance.Container, error)
...
+
+	// Preserve the previous non-list users field as an item on the new users list so it is still applied if applicable.
+	// For example, the users field can be a scalar containing the name of a user that has to be created, so this behavior has
+	// to be preserved.
+	if hasUsers && !isSlice {
+		userSequence = append(userSequence, userField)
+	}
+
+	// Define comment to be added on added lines.
+	sshKeyExtendedConfigTag := "#cloud-init.ssh-keys"
```
tomponline, line 642: Can we include "LXD" in this comment to show it was added by LXD?

lxd/devlxd.go
```diff
@@ -103,6 +103,14 @@ func devlxdConfigKeyGetHandler(d *Daemon, c instance.Instance, w http.ResponseWr
 		return response.DevLxdErrorResponse(api.StatusErrorf(http.StatusNotFound, "not found"), c.Type() == instancetype.VM)
 	}
 
+	// If parsing the config is not possible, abstain from merge the additional keys into the config.
+	if strings.HasSuffix(key, ".vendor-data") || strings.HasSuffix(key, ".user-data") {
+		value, err = mergeSSHKeyCloudConfig(c.ExpandedConfig(), value)
+		if err != nil {
+			logger.Warn("Failed merging SSH keys into cloud config, abstain from injecting additional keys")
```
tomponline, line 110: Lets log the request key here too please

tomponline: we also should include `instance` and `project` in the logging context as we have the instance available.

doc/cloud-init.md
```diff
@@ -296,3 +296,25 @@ config:
...
+
+```yaml
+config:
+  cloud-init.vendor-data: |
+    users:
+      - name: user
+        ssh_authorized_keys: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIJFDWcYmMrCZdk9JI29bAiHKD90oEUr8tqK5VvoO8Vcj
+```
+
+It is also possible to inject an additional key for the user `user` on an instance that uses this profile in such a way that the key defined on the profile is still applied. This can be done using the configuration key `cloud-init.ssh-keys.<keyName>`.
```
tomponline, line 314: into an instance

tomponline: lets re-order this to promote the new simpler way of injecting ssh keys rather than having the older manual way first.

doc/cloud-init.md
```diff
@@ -296,3 +296,25 @@ config:
       - type: nameserver
         address: 10.10.10.254
 ```
+
+## How to inject SSH keys on instances
+
+It is possible to inject SSH keys into LXD instances using `cloud-init.vendor-data` or `cloud-init.user-data`. For more information on how to do this using `cloud-config`, see [the cloud-init docs for SSH configuration](https://cloudinit.readthedocs.io/en/latest/reference/yaml_examples/ssh.html) for more information.
```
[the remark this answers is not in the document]
tomponline: lets re-phrase this section to say something like:

- Its possible to inject ssh keys into LXD instances using cloud-init.
- This can be done using LXD's ``cloud-init.ssh-keys.` configuration options 
- Or via manually constructed  `cloud-init.vendor-data` or `cloud-init.user-data` (see cloud-init docs)

doc/metadata.txt
```diff
@@ -1716,6 +1716,17 @@ The instance with the highest value is shut down first.
...
 
+```{config:option} cloud-init.ssh-keys.KEYNAME instance-cloud-init
+:condition: "If supported by image"
+:liveupdate: "no"
+:shortdesc: "Additional SSH key to be injected on the instance by `cloud-init`"
+:type: "string"
+Represent an additional SSH public key to be merged into existent `cloud-init` seed data
+and injected on an instance. Has the format `{user}:{key}`, where user is a Linux {user} and
+{key} can be either a pure SSH public keys or import IDs for keys hosted elsewhere.
+// For example: `root:gh:githubUser` or `myUser:ssh-keyAlg publicKeyHash`
```
tomponline, line 1727: this is what would be useful in the api definition doc too

doc/api-extensions.md
minaelee, line 2607:
```suggestion
To achieve this, the `cloud-init.ssh-keys.KEYNAME` configuration key is added for both instances and profiles, used to define a public key to be injected. `KEYNAME` can be any arbitrary name for the injected key.
```

doc/cloud-init.md
```diff
@@ -296,3 +296,25 @@ config:
...
+
+Here is an example of a profile configuration that defines a key to be injected on the instance. This enables the owner of the private key to SSH into the instance as a user named `user`:
```
minaelee, line 304:
```suggestion
The example profile configuration below defines a key to be injected on an instance. The injected key enables the owner of the private key to SSH into the instance as a user named `user`:
```

doc/cloud-init.md
```diff
@@ -296,3 +296,25 @@ config:
...
+        ssh_authorized_keys: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIJFDWcYmMrCZdk9JI29bAiHKD90oEUr8tqK5VvoO8Vcj
+```
+
+It is also possible to inject an additional key for the user `user` on an instance that uses this profile in such a way that the key defined on the profile is still applied. This can be done using the configuration key `cloud-init.ssh-keys.<keyName>`.
+
+```bash
+lxc launch <image> <instance-name> -c cloud-init.ssh-keys.<keyName> <public-key>
+```
+
+Additional keys defined via `cloud-init.ssh-keys.<keyName>` may not be applied if LXD is not able to parse neither the existing `cloud-config.vendor-data` and `cloud-config.user-data` for that instance. This could happen, for example, if the content of those keys is a badly formatted YAML.
```
minaelee, line 320:
```suggestion
Additional keys defined via `cloud-init.ssh-keys.<keyName>` cannot be applied if LXD is unable to parse either the existing `cloud-config.vendor-data` or `cloud-config.user-data` for that instance. This could happen, for example, if one of those keys contains badly formatted YAML.
```

lxd/devlxd.go
```diff
@@ -560,3 +569,161 @@ func findContainerForPid(pid int32, s *state.State) (instance.Container, error)
...
+				continue
+			}
+
+			// Create an empty userKeys if the user is not configured.
+			_, ok := users[user]
+			if !ok {
+				users[user] = &userKeys{}
+			}
+
+			// Check if ssh key is an import ID with with a format "keyServer:UserName".
```
minaelee, line 602:
```suggestion
			// Check if ssh key is an import ID with the format "keyServer:UserName".
```

lxd/instance/instancetype/instance.go
```diff
@@ -1137,6 +1137,20 @@ func ConfigKeyChecker(key string, instanceType Type) (func(value string) error,
 		}
 	}
 
+	// lxdmeta:generate(entities=instance; group=cloud-init; key=cloud-init.ssh-keys.KEYNAME)
+	// Represent an additional SSH public key to be merged into existent `cloud-init` seed data
+	// and injected on an instance. Has the format `{user}:{key}`, where user is a Linux {user} and
+	// {key} can be either a pure SSH public keys or import IDs for keys hosted elsewhere.
```
minaelee, line 1143:
```suggestion
	// {key} can be either a pure SSH public key or an import ID for a key hosted elsewhere.
```

minaelee (changes requested):

go.mod
hamistao: After doing some research, I did not see any good reason to use `v3` instead of `v2`, as the new features are not useful anywhere we do yaml parsing. Perhaps in the future we can consider bumping it but for now I suggest we keep using `v2`.

v3 release post: https://ubuntu.com/blog/api-v3-of-the-yaml-package-for-go-is-available

lxd/devlxd.go
hamistao: Makes sense since the separation between import IDs and public keys is strictly a `cloud-init` concept. Done!

lxd/devlxd.go
```diff
@@ -560,3 +561,161 @@ func findContainerForPid(pid int32, s *state.State) (instance.Container, error)
...
+	err := yaml.Unmarshal([]byte(cloudConfig), cloudConfigMap)
+	if err != nil {
+		return cloudConfig, fmt.Errorf("Could not unmarshall cloud-config: %w", err)
+	}
+
+	// Get previously defined user sequence in provided config, if present.
+	userField, hasUsers := cloudConfigMap["users"]
+	userSequence, isSlice := userField.([]any)
+
+	// If the users field is empty or set to something other than a list, which would not be recognized by
```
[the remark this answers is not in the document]
hamistao: In fact I was simply mistaken here as it is valid to have a `users` field that is not a list, for example:
```
users: testUser
```
Has the effect of creating a user named testUser, and this behavior can be preserverd by appending `testUser` to the resulting `users` list.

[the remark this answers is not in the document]
hamistao: The same is not valid for a map field though, so the following configuration would be innefective and we can handle it as a parsing error:
```
users:
  name: testUser
```

go.mod
tomponline: Sounds good thanks

doc/cloud-init.md
```diff
@@ -296,3 +296,33 @@ config:
       - type: nameserver
         address: 10.10.10.254
 ```
+
+## How to inject SSH keys into instances
+
+It is possible to inject SSH keys into LXD instances for an arbitrary user. This can be done using the configuration key `cloud-init.ssh-keys.<keyName>`.
+
+This configuration should respect the format `<user>:<key>`, where `<user` is a Linux username and `<key>` can be either a pure SSH public key or an import ID for a key hosted elsewhere. For example, `root:gh:githubUser` and `myUser:ssh-keyAlg publicKeyHash` are valid settings.
```
minaelee, line 304:
```suggestion
Use the format `<user>:<key>` for its value, where `<user>` is a Linux username and `<key>` can be either a pure SSH public key or an import ID for a key hosted elsewhere. For example, `root:gh:githubUser` and `myUser:ssh-keyAlg publicKeyHash` are valid values.
```

doc/cloud-init.md
```diff
@@ -296,3 +296,33 @@ config:
...
+
+`cloud-init.ssh-keys.<keyName>` cannot be applied if LXD is unable to parse both the existing `cloud-config.vendor-data` and `cloud-config.user-data` for that instance. This could happen, for example, if those keys contain badly formatted YAML.
+
+It is also possible to define SSH keys via `cloud-init.vendor-data` or `cloud-init.user-data` directly. Defining keys using `cloud-init.ssh-keys.<keyName>` does not conflict with neither `cloud-init.vendor-data` or `cloud-init.user-data` in any way. For more information on how to do this using `cloud-config`, see [the cloud-init docs for SSH configuration](https://cloudinit.readthedocs.io/en/latest/reference/yaml_examples/ssh.html).
```
minaelee, line 308:
```suggestion
You can define SSH keys via `cloud-init.vendor-data` or `cloud-init.user-data` directly. Keys defined using `cloud-init.ssh-keys.<keyName>` do not conflict with those defined with either `cloud-init.vendor-data` or `cloud-init.user-data` in any way. For more information on how to use `cloud-config` to define SSH keys, see [the cloud-init docs for SSH configuration](https://cloudinit.readthedocs.io/en/latest/reference/yaml_examples/ssh.html).
```

doc/cloud-init.md
```diff
@@ -296,3 +296,33 @@ config:
...
+
+Since `cloud-init` only runs on instance start, updates to `cloud-init.*` keys on a running instance will only take effect after restart.
```
minaelee, line 310:
```suggestion
Since `cloud-init` only runs on instance start, updates to `cloud-init.*` keys on a running instance only take effect after restart.
```

minaelee (changes requested):

doc/api-extensions.md
```diff
@@ -2599,3 +2599,11 @@ The following pool level configuration keys have been added:
 1. {config:option}`storage-pure-pool-conf:pure.api.token`
 1. {config:option}`storage-pure-pool-conf:pure.mode`
 1. {config:option}`storage-pure-pool-conf:pure.target`
+
+## `cloud_init_ssh_keys`
+
+Adds support for injecting additional SSH public keys into instances through `cloud-init` without conflicting with any configuration present on `cloud-init.vendor-data` or `cloud-init.user-data`.
```
minaelee, line 2605:
```suggestion
Adds support for injecting additional SSH public keys into instances through {ref}`cloud-init <instance-options-cloud-init>` without conflicting with any configuration present on {config:option}`instance-cloud-init:cloud-init.vendor-data` or {config:option}`instance-cloud-init:cloud-init.user-data`.
```

doc/cloud-init.md
```diff
@@ -296,3 +296,33 @@ config:
...
         address: 10.10.10.254
 ```
+
+## How to inject SSH keys into instances
+
+It is possible to inject SSH keys into LXD instances for an arbitrary user. This can be done using the configuration key `cloud-init.ssh-keys.<keyName>`.
+
+This configuration should respect the format `<user>:<key>`, where `<user` is a Linux username and `<key>` can be either a pure SSH public key or an import ID for a key hosted elsewhere. For example, `root:gh:githubUser` and `myUser:ssh-keyAlg publicKeyHash` are valid settings.
+
+`cloud-init.ssh-keys.<keyName>` cannot be applied if LXD is unable to parse both the existing `cloud-config.vendor-data` and `cloud-config.user-data` for that instance. This could happen, for example, if those keys contain badly formatted YAML.
```
minaelee, line 306:
```suggestion
`cloud-init.ssh-keys.<keyName>` cannot be applied if LXD is unable to parse both the existing {config:option}`instance-cloud-init:cloud-init.vendor-data` and {config:option}`instance-cloud-init:cloud-init.user-data` keys for that instance. This could happen, for example, if those keys contain badly formatted YAML.
```

lxd/instance/instancetype/instance.go
```diff
@@ -1137,6 +1137,21 @@ func ConfigKeyChecker(key string, instanceType Type) (func(value string) error,
 		}
 	}
 
+	// lxdmeta:generate(entities=instance; group=cloud-init; key=cloud-init.ssh-keys.KEYNAME)
+	// Represents an additional SSH public key to be merged into existing `cloud-init` seed data
+	// and injected into an instance. Has the format `{user}:{key}`, where {user} is a Linux username and
+	// {key} can be either a pure SSH public keys or import IDs for keys hosted elsewhere.
```
minaelee, line 1143:
```suggestion
	// {key} can be either pure SSH public keys or import IDs for keys hosted elsewhere.
```

lxd/instance/instancetype/instance.go
minaelee: Need to either get rid of "a" (before "pure SSH public keys") or make "key" singular throughout. If you don't want to make "key" singular throughout, then delete "a".

minaelee (changes requested):

doc/api-extensions.md
hamistao: tbh I wasn't aware of this syntax used in `{config:option}`instance-cloud-init:cloud-init.user-data` until now. Thanks for catching this and I will pay attention to it from now on.

lxd/instance/instancetype/instance.go
hamistao: Since the {key} part only represents one key, perhaps it would be better `// {key} can be either a pure SSH public key or an import ID for a key hosted elsewhere.`

hamistao:
@minaelee Thanks for the quick review, all of your suggestions were addressed (with one exception but for a good reason)

@tomponline more news! I got a response from `cloud-init` and confirmed the approach of using the same key for public keys and import IDs is reliable and we should not run into problems there.

Regarding the unit tests, both versions of the `yaml` library do not have a predictable order for map fields, which results in unpredictable ordering of users in the users list. Since the configuration is correct regardless of the order of the fields, I updated the tests to not fail based on the order or the fields.

As a useful info, I did some performance tests and found that parsing the colossal Juju generated `cloud-config`, injecting some keys in it and marshalling it back takes roughly 950µs in my machine, so peformance should not be an issue (and it works as expected!).

And I believe this is the end of every thread related to this. This should be ready to merge.

lxd/devlxd.go
```diff
@@ -70,12 +70,30 @@ func devlxdConfigGetHandler(d *Daemon, c instance.Instance, w http.ResponseWrite
 	}
 
 	filtered := []string{}
+	hasSSHKeys := false
+	hasCustomConfig := false
 	for k := range c.ExpandedConfig() {
-		if strings.HasPrefix(k, "user.") || strings.HasPrefix(k, "cloud-init.") {
+		// cloud-init.ssh-keys keys should not be retireved by cloud-init directly but instead should be merged into
```
tomponline, line 76: retireved

tomponline: I think a comment here explaining why we are merging the SSH keys into both vendor and user data would be useful in the future, as it seems counter intuitive.

doc/cloud-init.md
```diff
@@ -296,3 +296,33 @@ config:
...
+
+## How to inject SSH keys into instances
+
+To inject SSH keys into LXD instances for an arbitrary user, use the configuration key `cloud-init.ssh-keys.<keyName>`.
+
+Use the format `<user>:<key>` for its value, where `<user>` is a Linux username and `<key>` can be either a pure SSH public key or an import ID for a key hosted elsewhere. For example, `root:gh:githubUser` and `myUser:ssh-keyAlg publicKeyHash` are valid values.
+
+`cloud-init.ssh-keys.<keyName>` cannot be applied if LXD is unable to parse both the existing `cloud-config.vendor-data` and `cloud-config.user-data` for that instance. This could happen, for example, if those keys contain badly formatted YAML.
+
+You can define SSH keys via `cloud-init.vendor-data` or `cloud-init.user-data` directly. Keys defined using `cloud-init.ssh-keys.<keyName>` do not conflict with those defined with either `cloud-init.vendor-data` or `cloud-init.user-data` in any way. For more information on how to use `cloud-config` to define SSH keys, see [the cloud-init docs for SSH configuration](https://cloudinit.readthedocs.io/en/latest/reference/yaml_examples/ssh.html).
```
tomponline, line 308: This section would benefit from being expanded to explain that the SSH keys defined are merged into the config provided here and presented to the guest.

tomponline (approved):
I'm going to merge this now, but please follow up with the comments I mentioned, ta.

Issue #14965, opened 2025-02-12
Copy request body for Pure client retries


roosterfish:
Recently in the PowerFlex driver we found an issue when performing retries of a request in case the token expired.
In case the request that was retried had a non nil body, the bodies contents were already read when performing the request for the first time so any attempt to retry the request used an empty request body.

The fix was to always pass the actual map into the request function and create the bodies reader on demand for every retry. See https://github.com/canonical/lxd/pull/14865/commits/238651cc34ca2816e801b2bbc7af8a47e67a2d81.

Just saw this bug also went into Pure https://github.com/canonical/lxd/blob/main/lxd/storage/drivers/driver_pure_util.go#L363 and requires a fix.

PR #14968, opened 2025-02-12
Storage: Always create new request body reader 

MusicDin:
Always create new request body reader and ensure safe access to `accessToken` in Pure Storage driver.

Fixes https://github.com/canonical/lxd/issues/14965

### DIFF
--- a/lxd/storage/drivers/driver_pure_util.go
+++ b/lxd/storage/drivers/driver_pure_util.go
@@ -237,7 +237,7 @@ func (p *pureClient) createBodyReader(contents map[string]any) (io.Reader, error
 }
 
 // request issues a HTTP request against the Pure Storage gateway.
-func (p *pureClient) request(method string, url url.URL, reqBody io.Reader, reqHeaders map[string]string, respBody any, respHeaders map[string]string) error {
+func (p *pureClient) request(method string, url url.URL, reqBody map[string]any, reqHeaders map[string]string, respBody any, respHeaders map[string]string) error {
 	// Extract scheme and host from the gateway URL.
 	scheme, host, found := strings.Cut(p.driver.config["pure.gateway"], "://")
 	if !found {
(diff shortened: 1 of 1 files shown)


lxd/storage/drivers/driver_pure_util.go
```diff
@@ -355,17 +357,22 @@ func (p *pureClient) requestAuthenticated(method string, url url.URL, reqBody io
 		}
 
 		// Set access token as request header.
+		p.accessTokenLock.RLock()
```
roosterfish, line 370: Mhhh I am just questioning the locking at all around the token.

I know I have added it to the PowerFlex driver but are there ever two independent callers using the same `*pureclient` struct? I don't think so which means you don't have to add the locking here, neither do we need in PowerFlex.

Or maybe I am missing something now.

MusicDin: One client is per one storage pool. So doing multiple operations on the same pool can call request/login being called multiple times. I liked your approach primarily as it ensures we do not hit any panic when starting many instances, as in such case the multiple operations can call request() as well as login().

roosterfish: Let's use the callchain of the drivers `CreateVolume()` as an example:

```go
instancesPost() // API handler
createFromImage()
instanceCreateFromImage() // -> creates a new instance of the driver's struct
CreateInstanceFromImage()
EnsureImage()
CreateVolume()
```

An individual driver's struct is never shared across multiple callers. And as the client is stored inside the driver's struct there also cannot be two two callers working on the same struct which would require the lock.

roosterfish: Only in case you for example create a VM and mount both the root vol and FS vol the same struct is used because it's happening in "the same call".

MusicDin: > An individual driver's struct is never shared across multiple callers. And as the client is stored inside the driver's struct there also cannot be two two callers working on the same struct which would require the lock.

Thanks, fixed

roosterfish: I'll also add a small PR to revert this in PowerFlex. Thanks for raising it unknowingly :)

roosterfish: https://github.com/canonical/lxd/pull/14969

PR #14977, opened 2025-02-13
github: split documentation job into build and test

mas-who:
## Done
- Separated a portion of the documentation job from the test workflow into its own job. This prevents the UI E2E tests from depending on unrelated parts of the documentation job.

NOTE: On my fork [run](https://github.com/mas-who/lxd/actions/runs/13309237496/job/37167572965), the documentation test job was executed successfully, but it seems like the doc link checker is failing at the same spot


tomponline (approved):
Thanks!

PR #14979, opened 2025-02-13
Auth: Return missing group names on error

markylaing:
The error message displayed when adding an identity to groups is currently:
```
fmt.Errorf("Failed to write expected number of rows to identity auth group association table (expected %d, got %d)", len(groupNames), rowsAffected)
```
Which is a bit gross. It also isn't returning an appropriate HTTP error code.

I've added some error handling to check all the inputted groups and return a nice error containing all the groups that are missing.

### DIFF
--- a/lxd/db/cluster/identities.go
+++ b/lxd/db/cluster/identities.go
@@ -12,6 +12,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


tomponline (approved):
Cheers

lxd/db/cluster/identities.go
```diff
@@ -574,6 +574,26 @@ WHERE auth_groups.name IN %s
...
+			}
+
+			if !exists {
+				missingGroups = append(missingGroups, groupName)
+				msg += "%q"
+			}
+		}
+
+		if len(missingGroups) > 0 {
+			return api.StatusErrorf(http.StatusNotFound, msg, missingGroups...)
```
simondeziel, line 594: If multiple groups are missing, would they have a separator between them?

lxd/db/cluster/identities.go
```diff
@@ -574,6 +574,26 @@ WHERE auth_groups.name IN %s
 	}
 
 	if int(rowsAffected) != len(groupNames) {
+		// No longer on happy "fast" path.
+		// Check each group name exists to return a nice error
+		missingGroups := make([]any, 0, len(groupNames))
```
tomponline, line 579: Why slice of any?

test/suites/auth.sh
```diff
@@ -100,6 +100,8 @@ test_authorization() {
   BROWSER=curl lxc remote add --accept-certificate oidc "${LXD_ADDR}" --auth-type oidc
 
   ! lxc auth identity group add oidc/test-user@example.com not-found || false # Group not found
+  result="$(! lxc auth identity group add oidc/test-user@example.com not-found 2>&1)"
```
tomponline, line 103: Can we try a multi group error to check separator

lxd/db/cluster/identities.go
markylaing: Nope :facepalm:. Fixed.

lxd/db/cluster/identities.go
markylaing: This was to pass into the `NewStatusError` function as it requires a slice of any. I've changed it now to use strings.Join and string concatenation rather than a format string.

test/suites/auth.sh
markylaing: Done. I didn't because the command doesn't let you add more than one. Now I'm using curl to check it.

simondeziel (approved):

tomponline (approved):

PR #14982, opened 2025-02-13
Replace `source-type` with `source.type`

MggMuggins:
This is the kind of thing that makes people wonder "what the heck were they thinking" (see [docs](https://documentation.ubuntu.com/lxd/en/latest/reference/devices_disk/#device-disk-device-conf:size)):
Screenshot from 2025-02-13 16-30-00

I couldn't unsee it. Namespacing these keys under `source` just seems like a really good idea. I will use `source.snapshot` for snapshots.

There shouldn't be any changes needed for the lxd-ci tests; I ran them on this locally and they pass.

### DIFF
--- a/doc/metadata.txt
+++ b/doc/metadata.txt
@@ -186,7 +186,7 @@ See {ref}`devices-disk-types` for details.
 
 ```
 
-```{config:option} source-type device-disk-device-conf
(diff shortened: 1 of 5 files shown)


tomponline (approved):
Good spot

Issue #14803, opened 2025-01-15
user.* keys do not work with the cloud-init nocloud datasource provided by LXD


stew3254:
When using jinja templates for cloud-init, it's very convenient to pass custom values in by `user.*` and then access them by `ds.config.user_*`. Here is a LXD profile demonstrating an example.

```yaml
name: demo
description: Jinja cloud-init demo
config:
  cloud-init.user-data: |
    ## template: jinja
    #cloud-config
    users:
      - name: "{{ ds.config.user_username }}"
        sudo: ALL=(ALL) NOPASSWD:ALL
        shell: /bin/bash
        plain_text_passwd: "{{ ds.config.user_password }}"
        lock_passwd: false
  user.password: test
  user.username: test
```

If you were to launch this profile on a jammy or later vm, or any container, it should create a user named `test` with the password `test` instead of the `ubuntu` user.

Now, add a config drive to the profile under the devices section:
```yaml
devices:
  config:
    source: cloud-init:config
    type: disk
```

Now, the `user.username` and `user.password` variables don't get passed in via the NoCloud datasource. Therefore, the default `ubuntu` user is created instead. This occurs because the variables were not actually passed into the `meta-data` file in the root of `/dev/sr0` which the drive added. Therefore, no custom variables can be passed to instances launched using this method.

Currently, this breaks tooling using Focal vms as a base image, such as those trying to reproduce Ubuntu FIPS issues because Focal does not properly implement the standard LXD datasource.

tomponline:
@holmanb what do you think about this? Something that cloud-init can support?

stew3254:
@tomponline Sorry for the late reply. Cloud-init already supports using the `meta-data` file for the NoCloud provider. You can try for yourself by creating a LXD volume that is a vfat partiton with the label CIDATA. Create the following files:

`user-data`
```jinja
## template: jinja
#cloud-config
users:
  - name: "{{ ds.meta_data.user }}"
    sudo: ALL=(ALL) NOPASSWD:ALL
    shell: /bin/bash
    plain_text_passwd: "{{ ds.meta_data.password }}"
    lock_passwd: false
```

`meta-data`
```yaml
user: test
password: mypassword
```

Then, init a new LXD vm  without specifying anything to do with cloud-init in a LXD profile / instance config. Attach this volume, start it, and watch it initialize the `test` user instead of `ubuntu`.

```bash
# Create the machine and volume to set up for cloud-init
lxc init ubuntu:n --vm setup
[...]

Issue #13853, opened 2024-08-01
User-defined meta-data can create malformed YAML


holmanb:
# Required information

 * Distribution: Ubuntu Noble
 * The output of "snap list --all lxd core20 core22 core24 snapd":
```
Name    Version      Rev    Tracking       Publisher   Notes
core20  20240227     2264   latest/stable  canonical✓  base,disabled
core20  20240416     2318   latest/stable  canonical✓  base
core22  20240111     1122   latest/stable  canonical✓  base,disabled
core22  20240408     1380   latest/stable  canonical✓  base
lxd     6.1-0d4d89b  29469  latest/stable  canonical✓  disabled
lxd     6.1-c14927a  29551  latest/stable  canonical✓  -
snapd   2.62         21465  latest/stable  canonical✓  snapd,disabled
snapd   2.63         21759  latest/stable  canonical✓  snapd
```
 * The output of "lxc info" or if that fails:
```
config: {}
api_extensions:
- storage_zfs_remove_snapshots
- container_host_shutdown_timeout
- container_stop_priority
- container_syscall_filtering
- auth_pki
- container_last_used_at
- etag
- patch
- usb_devices
- https_allowed_credentials
- image_compression_algorithm
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
- unix_device_rename
- storage_lvm_use_thinpool
- storage_rsync_bwlimit
- network_vxlan_interface
- storage_btrfs_mount_options
- entity_description
- image_force_refresh
- storage_lvm_lv_resizing
- id_map_base
- file_symlinks
- container_push_target
- network_vlan_physical
- storage_images_delete
- container_edit_metadata
- container_snapshot_stateful_migration
- storage_driver_ceph
- storage_ceph_user_name
- resource_limits
- storage_volatile_initial_source
- storage_ceph_force_osd_reuse
- storage_block_filesystem_btrfs
- resources
- kernel_limits
- storage_api_volume_rename
[...]

PR #14983, opened 2025-02-14
cloud-init: Include `user.*` keys in `meta-data` file

hamistao:
We can't add make these values available through `ds.config` as we do when using the LXD datastore, so this is the next best thing.

This does raise the risk of adding duplicate keys since `user.meta-data` could include the same key as some `user.*`. Similarly to what is reported in https://github.com/canonical/lxd/issues/13853. But this won't be a concern after `user.meta-data` is removed.

Closes https://github.com/canonical/lxd/issues/14803

### DIFF
--- a/lxd/device/disk.go
+++ b/lxd/device/disk.go
@@ -2597,13 +2597,29 @@ func (d *disk) generateVMConfigDrive() (string, error) {
 		}
 	}
 
-	// Append any custom meta-data to our predefined meta-data config.
-	metaData := fmt.Sprintf(`instance-id: %s
-local-hostname: %s
-%s
-`, d.inst.Name(), d.inst.Name(), instanceConfig["user.meta-data"])
+	var metaDataBuilder strings.Builder
 
-	err = os.WriteFile(filepath.Join(scratchDir, "meta-data"), []byte(metaData), 0400)
+	// Append strings to the builder
+	metaDataBuilder.WriteString("instance-id: " + d.inst.Name() + "\n")
+	metaDataBuilder.WriteString("local-hostname: " + d.inst.Name() + "\n")
+
+	// These keys shouldn't be appended to the meta as it would be redundant as their values are already available
+	// for cloud-init. Only the content of `user.meta-data` is appended to meta_data.
+	excludedKeys := []string{"user.meta-data", "user.user-data", "user.vendor-data", "user.network-config"}
+
+	// Add keys that are exposed to cloud-init to meta-data so one can use in jinja templates.
+	// The added values are single quoted to prevent rendering `meta-data` unparseable.
+	// Single quotes included in the value itself are escaped by being replaced with `''`.
+	for key, value := range instanceConfig {
+		if strings.HasPrefix(key, "user.") && !shared.ValueInSlice(key, excludedKeys) {
+			metaDataBuilder.WriteString(key + ": '" + strings.Replace(value, "'", "''", -1) + "'\n")
+		}
+	}
+
+	// Append any custom meta-data.
+	metaDataBuilder.WriteString(instanceConfig["user.meta-data"])
+
+	err = os.WriteFile(filepath.Join(scratchDir, "meta-data"), []byte(metaDataBuilder.String()), 0400)
 	if err != nil {
 		return "", err
 	}
--- a/lxd/device/disk.go
+++ b/lxd/device/disk.go
@@ -2600,7 +2600,7 @@ func (d *disk) generateVMConfigDrive() (string, error) {
 	var metaDataBuilder strings.Builder
 
 	// Append strings to the builder
-	metaDataBuilder.WriteString("instance-id: " + d.inst.Name() + "\n")
+	metaDataBuilder.WriteString("instance-id: " + d.inst.CloudInitID() + "\n")
 	metaDataBuilder.WriteString("local-hostname: " + d.inst.Name() + "\n")
 
 	// These keys shouldn't be appended to the meta as it would be redundant as their values are already available


stew3254:
> > We can't add make these values available through `ds.config` as we do when using the LXD datastore, so this is the next best thing.
> 
> Why is that?

Cloud-init limitations. The NoCloud DS doesn't support this so it would require changing upstream to my understanding. LXD is the only cloud I'm aware of not following the convention of `ds.meta_data`.

tomponline:
> > > We can't add make these values available through `ds.config` as we do when using the LXD datastore, so this is the next best thing.
> > 
> > 
> > Why is that?
> 
> Cloud-init limitations. The NoCloud DS doesn't support this so it would require changing upstream to my understanding. LXD is the only cloud I'm aware of not following the convention of `ds.meta_data`.

I wonder why cloud-init chose to expose the user vars from /dev/lxd under `ds.config` rather than `ds.meta_data`?

Any ideas @holmanb ?

hamistao:
> Its rather unfortunate that we cant offer a consistent experience here.

Indeed :cry: 

> I wonder why cloud-init chose to expose the user vars from /dev/lxd under ds.config rather than ds.meta_data?

Just as additional info, and I am not entirely sure of this, but apparentely that accessing values through `ds.config` is only possible for the LXD datasource and no other.

tomponline:
OK I'm going to hold off on this until we've had confirmation from @holmanb that there's no way of doing this consistently. Thanks

lxd/device/disk.go
```diff
@@ -2597,11 +2597,24 @@ func (d *disk) generateVMConfigDrive() (string, error) {
...
-local-hostname: %s
-%s
-`, d.inst.Name(), d.inst.Name(), instanceConfig["user.meta-data"])
+	local-hostname: %s
+`, d.inst.Name(), d.inst.Name())
+
+	// Add other user.* keys to meta-data so that cloud-init has access to this information.
+	for key, value := range instanceConfig {
+		if strings.HasPrefix(key, "user.") && !shared.ValueInSlice(key, reservedUserKeys) {
+			fmt.Println(key)
```
tomponline, line 2609: is this old debug ?

lxd/device/disk.go
```diff
@@ -2597,11 +2597,24 @@ func (d *disk) generateVMConfigDrive() (string, error) {
...
+			metaData += key + ": " + value + "\n"
+		}
+	}
+
+	// Append any custom meta-data.
+	metaData += instanceConfig["user.meta-data"]
+
+	fmt.Println(metaData)
```
tomponline, line 2617: debug?

lxd/device/disk.go
tomponline: Rather than using fmt.Sprintf in this function, how about using strings.Builder to build up the contents of the file?

lxd/device/disk.go
hamistao: Whoops, fixed :sweat_smile:

hamistao:
> OK I'm going to hold off on this until we've had confirmation from @holmanb that there's no way of doing this consistently. Thanks

@tomponline  We had confirmation from another member of the cloud-init team as mentioned in the issue. You can check the cloud-init channel on Mattermost if you'd like :wink:

tomponline:
> > OK I'm going to hold off on this until we've had confirmation from @holmanb that there's no way of doing this consistently. Thanks
> 
> @tomponline We had confirmation from another member of the cloud-init team as mentioned in the issue (I am afraid I don't know his GH handle). You can check the cloud-init channel on Mattermost if you'd like 😉

Will you provide a link?

hamistao:
> Will you provide a link?

Here it is, I was on my phone and preferred to come to the computer to get the link: https://chat.canonical.com/canonical/pl/ieahsfitgjrd9p6okssof9r4yw

lxd/device/disk.go
```diff
@@ -2597,11 +2597,21 @@ func (d *disk) generateVMConfigDrive() (string, error) {
 		}
 	}
 
-	// Append any custom meta-data to our predefined meta-data config.
+	reservedUserKeys := []string{"user.network-config", "user.user-data", "user.vendor-data", "user.meta-data"}
```
tomponline, line 2600: @hamistao @holmanb btw, does cloud-init itself have similar logic when generating the `ds.config` var when using the lxd datasource, so as not to request these keys from /dev/lxd?

tomponline (changes requested):
please can you use strings.Builder to build up the file data rather than fmt.Sprintf, it should work out clearer than having placeholders.

lxd/device/disk.go
hamistao: It does not, every `user.*` key is added to `ds.config`. I have adjusted the implementation to match it.

Your other suggestion regarding the use of strings.Builder was also addressed.

lxd/device/disk.go
```diff
@@ -2597,13 +2597,27 @@ func (d *disk) generateVMConfigDrive() (string, error) {
...
+	metaDataBuilder.WriteString(d.inst.Name())
+
+	// Add other user.* keys to meta-data so that cloud-init has access to this information.
+	for key, value := range instanceConfig {
+		if strings.HasPrefix(key, "user.") {
+			metaDataBuilder.WriteString("\n" + key + ": " + value)
+		}
+	}
+
+	metaDataBuilder.WriteString("\n")
```
tomponline, line 2615: then we can lose this line

lxd/device/disk.go
```diff
@@ -2597,13 +2597,27 @@ func (d *disk) generateVMConfigDrive() (string, error) {
...
-	metaData := fmt.Sprintf(`instance-id: %s
-local-hostname: %s
-%s
-`, d.inst.Name(), d.inst.Name(), instanceConfig["user.meta-data"])
+	var metaDataBuilder strings.Builder
 
-	err = os.WriteFile(filepath.Join(scratchDir, "meta-data"), []byte(metaData), 0400)
+	// Append strings to the builder
+	metaDataBuilder.WriteString("instance-id: ")
+	metaDataBuilder.WriteString(d.inst.Name())
```
tomponline, line 2604: Lets review whether the use of d.inst.Name() is still valid here or whether it should be d.inst.CloudInitID() instead.

Please also check that both /dev/lxd in containers and VMs when using the LXD DS are using d.inst.CloudInitID().

lxd/device/disk.go
```diff
@@ -2597,13 +2597,27 @@ func (d *disk) generateVMConfigDrive() (string, error) {
...
-	// Append any custom meta-data to our predefined meta-data config.
-	metaData := fmt.Sprintf(`instance-id: %s
-local-hostname: %s
-%s
-`, d.inst.Name(), d.inst.Name(), instanceConfig["user.meta-data"])
+	var metaDataBuilder strings.Builder
 
-	err = os.WriteFile(filepath.Join(scratchDir, "meta-data"), []byte(metaData), 0400)
+	// Append strings to the builder
+	metaDataBuilder.WriteString("instance-id: ")
```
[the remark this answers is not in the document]
hamistao: The string build refacotring was done.

lxd/device/disk.go
hamistao: Indeed using `d.inst.CloudInitID()` is more appropriate here, to quote my commit message:

```
Using `d.inst.CloudInitID()` fits the standard that we use for the LXD datasource, and there is no reason to use a different format in NoCloud. This also hepls to make changes to cloud-init configuration apply on every boot by updating the `instance-id` and allows a user to set a custom `instance-id` through `volatile.cloud-init.instance-id.`
```

lxd/device/disk.go
```diff
@@ -2597,13 +2597,25 @@ func (d *disk) generateVMConfigDrive() (string, error) {
...
-	metaData := fmt.Sprintf(`instance-id: %s
-local-hostname: %s
-%s
-`, d.inst.Name(), d.inst.Name(), instanceConfig["user.meta-data"])
+	var metaDataBuilder strings.Builder
 
-	err = os.WriteFile(filepath.Join(scratchDir, "meta-data"), []byte(metaData), 0400)
+	// Append strings to the builder
+	metaDataBuilder.WriteString("instance-id: " + d.inst.CloudInitID() + "\n")
+	metaDataBuilder.WriteString("\nlocal-hostname: " + d.inst.Name() + "\n")
```
tomponline, line 2604: we dont need the leading `\n` on this line right?

lxd/device/disk.go
```diff
@@ -2597,13 +2597,25 @@ func (d *disk) generateVMConfigDrive() (string, error) {
...
+
+	// Add other user.* keys to meta-data so that cloud-init has access to this information.
+	for key, value := range instanceConfig {
+		if strings.HasPrefix(key, "user.") {
```
tomponline, line 2608: do we need to skip the legacy cloud-init keys here?

hamistao: No, we do not since the LXD datasource doesn't do this when including `user.*` keys on `ds.config`.

tomponline: Hrm that sounds like a bug though. It shouldnt serving up the legacy cloud-init config keys right?

hamistao: I am wondering if we should exclude `user.meta-data` though since its content is already included on `ds.meta_data`. But it could be used to distinguish the source of the meta-data and we are planning to remove it so this shouldn't be a big problem.

tomponline: I dont think we should include user.user-data, user.vendor-data or user.meta-data as these are already included and are not custom user vars for accessing in the template.

tomponline: > cloud-init just includes every defined config key that it has access to into ds.config. So this includes all user.* and cloud-init.*

Yeah, this sounds like a bug, why would I want to reference `user.user-data` or `cloud-init.user-data` as a custom user template var?

hamistao: We cannot do anything about it on the LXD DataSource, as we just show the defined keys and the logic for creating the datasource structure, such as `ds.config`, is included in `cloud-init`. But we can limit the keys we are including on `meta_data` on `NoCloud` datasource disk device.

tomponline: @hamistao indeed.

Please can you bring this up with @holmanb to see if this needs also to be a bug report in cloud-init?

lxd/device/disk.go
```diff
@@ -2597,13 +2597,25 @@ func (d *disk) generateVMConfigDrive() (string, error) {
...
-`, d.inst.Name(), d.inst.Name(), instanceConfig["user.meta-data"])
+	var metaDataBuilder strings.Builder
 
-	err = os.WriteFile(filepath.Join(scratchDir, "meta-data"), []byte(metaData), 0400)
+	// Append strings to the builder
+	metaDataBuilder.WriteString("instance-id: " + d.inst.Name() + "\n")
+	metaDataBuilder.WriteString("local-hostname: " + d.inst.Name() + "\n")
+
+	// Add keys that are exposed to cloud-init to meta-data so one can use in jinja templates.
+	for key, value := range instanceConfig {
```
tomponline, line 2607: @hamistao we still need to exclude the `user.meta-data`, `user.user-data`, `user.vendor-data`, `cloud-init.user-data` and `cloud-init.vendor-data` keys here please.

If we are going to need this logic in multiple places suggest having a constant slice of excluded keys.

lxd/device/disk.go
hamistao: @holmanb Since we already called your attention, we are working on making `user.*` keys availlable for use in jinja templates when using the NoCloud datasource on LXD and we are discussing the fact that keys like `cloud-init.user-data` and `user.network-config` are included in `ds.config`, but this seems redundant since these keys are already used as a source of other data available to cloud-init, such as `user-data` or `network-config`. So the same information is included in more than one place. Do you think this could use some improvement?

tomponline (commented):
One thought I had with this now.

Is it possible that by introducing this change we break an existing working instance that uses the NoCloud data store drive by including `user.*` keys that contain characters that render the meta-data unparseable by cloud-init?

hamistao:
> Is it possible that by introducing this change we break an existing working instance that uses the NoCloud data store drive by including user.* keys that contain characters that render the meta-data unparseable by cloud-init?

Good point. In the current implementation it is possible indeed, but single quoting added values prevents this. I will be pushing a new version.

tomponline (approved):
Thanks!

PR #14984, opened 2025-02-14
Daemon: Update storage.backups_volume and storage.images_volume validation

tomponline:
Should close https://github.com/canonical/lxd/security/code-scanning/1 (although not real security issue because codeql is not correctly detecting that the existing validation prevents using unvalidated raw input anyway).

### DIFF
--- a/lxd/api_1.0.go
+++ b/lxd/api_1.0.go
@@ -648,14 +648,26 @@ func doAPI10Update(d *Daemon, r *http.Request, req api.ServerPut, patch bool) re
 
 		// Validate the storage volumes
 		if nodeValues["storage.backups_volume"] != nil && nodeValues["storage.backups_volume"] != newNodeConfig.StorageBackupsVolume() {
-			err := daemonStorageValidate(s, nodeValues["storage.backups_volume"].(string))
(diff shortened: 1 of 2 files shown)


lxd/api_1.0.go
```diff
@@ -648,14 +648,26 @@ func doAPI10Update(d *Daemon, r *http.Request, req api.ServerPut, patch bool) re
 
 		// Validate the storage volumes
 		if nodeValues["storage.backups_volume"] != nil && nodeValues["storage.backups_volume"] != newNodeConfig.StorageBackupsVolume() {
-			err := daemonStorageValidate(s, nodeValues["storage.backups_volume"].(string))
+			backupsPoolVolume, ok := nodeValues["storage.backups_volume"].(string)
```
MusicDin, line 651: Just for my understanding? Is this check only meant to prevent security warning, or could the type actually be non-string?

MusicDin (approved):

roosterfish (approved):

lxd/api_1.0.go
tomponline: this is just a safety check because the var type is any. The code scanners pick this up a potential crash point.

PR #14986, opened 2025-02-14
github: Rework documentation tests

tomponline:
- Move inclusive name checker into main documentation build and test workflow.
 - Separate out the link checker workflow from rest of documentation workflow as unreliable - this avoids unnecessary checkout and download of documentation build just to run inclusive name check on normal PRs.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -465,7 +465,7 @@ jobs:
   ui-e2e-tests:
     name: UI e2e tests
     runs-on: ubuntu-latest
-    needs: [code-tests, documentation-build]
+    needs: [code-tests, documentation]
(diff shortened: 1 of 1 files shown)


mas-who (approved):
LGTM, thanks!

MusicDin (approved):
Thanks! Should we do something similar for lxd-imagebuilder? Although, currently there is no scheduled tests, so that could be a reason to introduce them?

tomponline:
> Thanks! Should we do something similar for lxd-imagebuilder? Although, currently there is no scheduled tests, that could be a reason to introduce them?

You can use this technique for problem URLs btw https://github.com/canonical/lxd/pull/14980

MusicDin:
> > Thanks! Should we do something similar for lxd-imagebuilder? Although, currently there is no scheduled tests, that could be a reason to introduce them?
> 
> You can use this technique for problem URLs btw #14980

Even better, thanks!

PR #14989, opened 2025-02-14
lxd: Extra validation of daemon storage options pool and volume names

simondeziel:
Should hopefully please CodeQL for https://github.com/canonical/lxd/security/code-scanning/1031 and possibly https://github.com/canonical/lxd/security/code-scanning/1

### DIFF
--- a/lxd/daemon_storage.go
+++ b/lxd/daemon_storage.go
@@ -133,13 +133,21 @@ func daemonStorageMount(s *state.State) error {
 }
 
 func daemonStorageSplitVolume(volume string) (poolName string, volumeName string, err error) {
-	fields := strings.Split(volume, "/")
-	if len(fields) != 2 {
+	if strings.Count(volume, "/") != 1 {
 		return "", "", fmt.Errorf("Invalid syntax for volume, must be <pool>/<volume>")
 	}
 
-	poolName = fields[0]
-	volumeName = fields[1]
+	poolName, volumeName, _ = strings.Cut(volume, "/")
+
+	// Validate pool name.
+	if strings.Contains(poolName, "\\") || strings.Contains(poolName, "..") {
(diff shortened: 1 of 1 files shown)


tomponline:
The other day in https://github.com/canonical/lxd/pull/14967 I initially followed the 2nd suggestion 

```
func handler(w http.ResponseWriter, r *http.Request) {
	path := r.URL.Query()["path"][0]

	// GOOD: ensure that the filename has no path separators or parent directory references
	// (Note that this is only suitable if `path` is expected to have a single component!)
	if strings.Contains(path, "/") || strings.Contains(path, "\\") || strings.Contains(path, "..") {
		http.Error(w, "Invalid file name", http.StatusBadRequest)
		return
	}
	data, _ := ioutil.ReadFile(filepath.Join("/home/user/", path))
	w.Write(data)
}
```

But it still didn't consider the matter resolved even though I followed it's own suggestion. :(

tomponline (approved):

tomponline:
I actually dont like this, because now we're putting in use-case specific logic for what is/is not a valid name, rather than just doing a DB lookup.

simondeziel:
The DB validation happens in `daemonStorageValidate()` which actually calls the `daemonStorageSplitVolume()` one. I've checked all other callers of `daemonStorageSplitVolume()` and they all validate the pool/volume exist in the DB.

tomponline:
> The DB validation happens in `daemonStorageValidate()` which actually calls the `daemonStorageSplitVolume()` one. I've checked all other callers of `daemonStorageSplitVolume()` and they all validate the pool/volume exist in the DB.

yep i did that same trace this morning :)

My point being that codeql seems to inconsistently require additional string validation in some places, but not in others.

tomponline (approved):

PR #14990, opened 2025-02-14
Storage: Prevent using remote storage pools for `(images|backups)_volume`

roosterfish:
This prevents using FS vols from unsupported remote storage drivers for the `storage.(images|backups)_volume` config keys.
It prevents concurrent reads/writes from multiple hosts when used inside a cluster.

Just checking for remote drivers is not sufficient as CephFS volumes can be used.

### DIFF
--- a/lxd/daemon_storage.go
+++ b/lxd/daemon_storage.go
@@ -167,6 +167,10 @@ func daemonStorageValidate(s *state.State, target string) (validatedTarget strin
 		return "", fmt.Errorf("Storage pool %q cannot be used when in %q status", poolName, poolState)
 	}
 
+	if pool.Driver().Info().Remote {
+		return "", fmt.Errorf("Remote storage pool %q cannot be used", pool)
+	}
+
 	var snapshots []db.StorageVolumeArgs
 	var dbVol *db.StorageVolume
 
--- a/test/suites/serverconfig.sh
+++ b/test/suites/serverconfig.sh
@@ -25,6 +25,10 @@ _server_config_storage() {
 
   lxd_backend=$(storage_backend "$LXD_DIR")
   if [ "$lxd_backend" = "ceph" ]; then
+    # The volume doesn't have to be present as the check errors after testing for the remote storage pool.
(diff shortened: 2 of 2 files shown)


roosterfish:
@tomponline it was indeed allowed.

tomponline (approved):

tomponline:
ta

tomponline:
> Just checking for remote drivers is not sufficient as CephFS volumes can be used.

Wait, does this block using cephfs volumes now too? We should allow that.

lxd/daemon_storage.go
```diff
@@ -167,6 +167,10 @@ func daemonStorageValidate(s *state.State, target string) (validatedTarget strin
 		return "", fmt.Errorf("Storage pool %q cannot be used when in %q status", poolName, poolState)
 	}
 
+	if pool.Driver().Info().Remote {
```
tomponline, line 170: @roosterfish doesnt this block using cephfs volumes for daemon storage? This should be allowed.

tomponline: We only want to prevent using volumes on block backed remote storage drivers.

roosterfish: I have actually reverted from this after checking the code for `storage.images_volume` [here](https://github.com/canonical/lxd/blob/main/lxd/images.go#L2202), see https://github.com/canonical/lxd/compare/bf104110330885582ae04cd3989fd341f938d094..69d482cf36355b0dd8cb75f90bd0e1a94705d60d.

If the `distributeImage` func is executed on all cluster nodes don't they all try to create the image (if it does not exist)? I know CephFS will handle the concurrent read/write but isn't that still not ideal to run those operations concurrently?

tomponline: Iirc theres code to skip distribution for remote pools

tomponline: We will need to revert this pr

roosterfish: On it.

roosterfish: Fixed in https://github.com/canonical/lxd/pull/14993

roosterfish: > Iirc theres code to skip distribution for remote pools

Yes there is https://github.com/canonical/lxd/blob/main/lxd/images.go#L2155. Was looking too faster over the code

PR #14991, opened 2025-02-14
Remove a few more `fmt.Sprintf()` and do some preallocations

simondeziel:
Also try to address some issues found by CodeQL.


lxd-agent/events.go
```diff
@@ -163,8 +163,10 @@ func eventsProcess(event api.Event) {
 	// If the path is not absolute, the mount will be created at `/run/lxd_agent/<path>`
 	// (since the mount command executed below originates from the `lxd-agent` binary that is in the `/run/lxd_agent` directory).
 	// This is not ideal and not consistent with the way mounts are handled with containers. We then make the path absolute.
-	if !filepath.IsAbs(e.Config["path"]) {
-		e.Config["path"] = filepath.Join("/", e.Config["path"])
+	e.Config["path"], err = filepath.Abs(e.Config["path"])
+	if err != nil {
+		l.Error("Failed to make path absolute")
```
tomponline, line 168: Would be good to log path here

tomponline: Ah yeah i couldnt see that in review pane ta

tomponline (approved):

PR #15015, opened 2025-02-18
cloud-init: SSH key injection improvements

hamistao:
user-data's fields overwrite vendor-data's fields, so merging SSH keys can result in adding a "users" fields that did not exist before, having the side effect of overwriting vendor-data's "users" field.

In order to fix we had to extend the logic for merging SSH keys into cloud-config, and that prompted us to refactor the merging logic, moving it to a newly created `cloudinit` package.


lxd/instance/drivers/driver_common.go
```diff
@@ -1250,7 +1250,7 @@ func (d *common) needsNewInstanceID(changedConfig []string, oldExpandedDevices d
 		"user.user-data",
 		"user.network-config",
 	} {
-		if shared.ValueInSlice(key, changedConfig) {
+		if shared.ValueInSlice(key, changedConfig) || strings.HasPrefix(key, "cloud-init.ssh-keys.") {
```
tomponline, line 1253: this new logic is unreachable.

tomponline (changes requested):
As discussed in our 1:1 lets move this into a `cloudinit` package and out of `util`.

tomponline:
@hamistao please ping me when ready for another review

hamistao:
@tomponline This is now ready

lxd/cloudinit/config.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// MergeSSHKeyCloudConfig merges any existing SSH keys defined in an instance config into a provided
+// cloud-config YAML string.
+// In the case where we were not able to parse the cloud config, return the original, unchanged config and
+// the error.
+func MergeSSHKeyCloudConfig(instanceConfig map[string]string, cloudConfig string) (string, error) {
+	// Check if cloudConfig is in a supported format.
+	// A YAML cloud config without #cloud-config is invalid.
+	// The "#cloud-config" tag can be either on the first or second lines.
```
tomponline, line 108: why not the third line? This seems an odd statement.

lxd/devlxd.go
```diff
@@ -118,17 +118,19 @@ func devlxdConfigKeyGetHandler(d *Daemon, c instance.Instance, w http.ResponseWr
...
 
-	// If parsing the config is not possible, abstain from merging the additional keys into the config.
-	if strings.HasSuffix(key, ".vendor-data") || strings.HasSuffix(key, ".user-data") {
-		value, err = util.MergeSSHKeyCloudConfig(c.ExpandedConfig(), value)
-		if err != nil {
-			logger.Warn("Failed merging SSH keys into cloud-init seed data, abstain from injecting additional keys", logger.Ctx{"err": err, "project": c.Project().Name, "instance": c.Name(), "requestedKey": key})
-		}
+	// For values containing cloud-init seed data, try to merge into them additional SSH keys present on the instance config.
+	// If parsing the config is not possible, abstain from merging the additional keys.
+	if strings.HasSuffix(key, ".user-data") {
```
tomponline, line 125: Please can we do exact matching on the relevant keys rather than suffix matching otherwise customer keys like user.foo.user-data would be in scope incorrectly.

lxd/device/disk.go
```diff
@@ -2540,55 +2540,32 @@ func (d *disk) generateVMConfigDrive() (string, error) {
...
-	if !ok {
-		vendorDataKey = "user.vendor-data"
-		vendorData = instanceConfig[vendorDataKey]
-		if vendorData == "" {
-			vendorData = "#cloud-config\n{}"
-		}
+	// Get raw data from instance config.
+	vendorDataKey := cloudinit.GetEffectiveConfigKey(instanceConfig, "vendor-data")
+	userDataKey := cloudinit.GetEffectiveConfigKey(instanceConfig, "user-data")
+	vendorData, userData := cloudinit.GetResultingCloudConfig(instanceConfig, vendorDataKey, userDataKey, d.inst.Name(), d.inst.Project().Name)
```
tomponline, line 2546: In my view it would be clearer to return a `cloudinit.Config` (or similarly named) struct that contained fields for userdata, vendordata etc rather than multiple return values as it allows you to pass around the overall cloud config.

tomponline, line 2546: Could we have the logic for `GetEffectiveConfigKey` moved inside `GetResultingCloudConfig` and then avoid the caller needing to figure this stuff out by just having a `requestedKey` argument that the the `cloudinit` package figures out what needs to be returned?

tomponline: Rather than "Resulting" lets use "Effective" here as that aligns more with other concepts in LXD (effective project for example).

lxd/cloudinit/config.go
tomponline: this already seems to be in an earlier commit, so looks like a bad rebase somewhere

tomponline:
I'd left some comments about trying to reduce the size of the API to keep as much of the cloud-init key-specific logic away from the caller's concern and inside the package itself.

Also lets try and avoid using prefix matching as that is prone to unexpected failure.

lxd/cloudinit/config.go
hamistao: > why not the third, fourth or n lines? This seems an odd statement.

The YAML is supposed to have `#cloud-config` on the first line. It can be the second line if using a template like jinja, in which case the first lines would be `## template: jinja` followed by `#cloud-config`.

hamistao:
@tomponline This is now ready

lxd/cloudinit/config.go
```diff
@@ -73,14 +73,14 @@ func GetEffectiveConfig(instanceConfig map[string]string, requestedKey string, i
 	var vendorErr error
 	var userErr error
 
-	// Defer logging a warning for each desired output in case of a parsing error.
+	// Defer logging a warning for each desired output in case of failing to merge SSH keys into cloud-init data due to a parsing error.
 	// An output is considered desired if it is the expected output for requestedKey's effective value or no requestedKey was provided.
 	defer func() {
-		if (requestedKey == "" || vendorKeyProvided) && vendorErr != nil {
+		if len(userKeys) == 0 && (requestedKey == "" || vendorKeyProvided) && vendorErr != nil {
```
tomponline, line 79: Should this be >0

hamistao: Whoops, fixed :sweat_smile:

tomponline (approved):

PR #15022, opened 2025-02-18
Improved validation of project and instance names

simondeziel:


simondeziel:
@tomponline you were right, CodeQL was apparently losing track of the `newName` being validated in the `Rename()` function. Repeating the check in `RenameInstance()` and `RenameInstanceSnapshot()` seem to have appeased it, frustrating!

simondeziel:
@tomponline what are your thoughts on restricting volume and project names to just `[a-zA-Z0-9-]+` + `.` (but not 2 consecutive dots anywhere)?

Allowing other weird chars like `*` and such make me worry with our use of `rsync` and `tar` to just name those.

tomponline:
> @tomponline what are your thoughts on restricting volume and project names to just `[a-zA-Z0-9-]+` + `.` (but not 2 consecutive dots anywhere)?
> 
> Allowing other weird chars like `*` and such make me worry with our use of `rsync` and `tar` to just name those.

Anything that works today and wouldn't work after the restriction has the potential to break existing users.
So we'd need to run this past @escabo and have a plan for how to support any affected users in recovering their LXD operation.

simondeziel:
> > @tomponline what are your thoughts on restricting volume and project names to just `[a-zA-Z0-9-]+` + `.` (but not 2 consecutive dots anywhere)?
> > Allowing other weird chars like `*` and such make me worry with our use of `rsync` and `tar` to just name those.
> 
> Anything that works today and wouldn't work after the restriction has the potential to break existing users. So we'd need to run this past @escabo and have a plan for how to support any affected users in recovering their LXD operation.

Yeah, I've thought about it more and that idea is really not viable. It's guarantied to break stuff and would be hard and risky to migrate stuff around.

This was a bad idea and I even forgot about non-ASCII chars :/

escabo:
Can we identify non-compliant names during the upgrade and either stop or automatically rename (é->e) and report changes? I don't know that a lot of users would have used characters outside of that new proposed range.

lxd/util/http.go
```diff
@@ -225,7 +226,7 @@ func CheckMutualTLS(cert x509.Certificate, trustedCerts map[string]x509.Certific
 
 	// Check whether client certificate is in the map of trusted certs.
 	for fingerprint, v := range trustedCerts {
-		if bytes.Equal(cert.Raw, v.Raw) {
+		if subtle.ConstantTimeCompare(cert.Raw, v.Raw) == 1 {
```
kadinsayani, line 229: Nice to see this change!

kadinsayani (dismissed):
LGTM!

lxd/util/http.go
kadinsayani: I did a quick grep and I don't see any concerning use of `bytes.Equal` elsewhere in the codebase.

kadinsayani: Perhaps this one: https://github.com/canonical/lxd/blob/lxd-remotes/lxd/ip/neigh.go#L74-L79.

simondeziel: I found a few, hopefully all addressed by https://github.com/canonical/lxd/pull/15036

kadinsayani: Thanks!

lxd/util/http.go
```diff
@@ -267,7 +267,7 @@ func ListenAddresses(configListenAddress string) ([]string, error) {
 	listenIP := net.ParseIP(unwrappedConfigListenAddress)
 	if listenIP != nil || !strings.Contains(unwrappedConfigListenAddress, ":") {
 		// Use net.JoinHostPort so that IPv6 addresses are correctly wrapped ready for parsing below.
-		configListenAddress = net.JoinHostPort(unwrappedConfigListenAddress, fmt.Sprintf("%d", shared.HTTPSDefaultPort))
+		configListenAddress = net.JoinHostPort(unwrappedConfigListenAddress, fmt.Sprint(shared.HTTPSDefaultPort))
```
tomponline, line 271: note for the future, lets prefer using strconv.Itoa or Format rather than fmt.Sprint in the future to avoid any unexpected formatting bugs.

shared/cert.go
```diff
@@ -490,7 +490,7 @@ func GetRemoteCertificate(address string, useragent string) (*x509.Certificate,
 	}
 
 	// Connect
-	req, err := http.NewRequest("GET", address, nil)
+	req, err := http.NewRequest("HEAD", address, nil)
```
tomponline, line 493: @simondeziel I'd prefer functional changes like this aren't included in PR that it isnt related to ("Improved validation of project and instance names").

If the remote doesnt support HEAD requests then this could be a regression.

Some things that should be considered before we land this change (if not already done):

- How far back in terms of LXD server versions did you test this?
- Is this function only ever called for LXD remotes or can it be used for anything else?
- Is it used in microcluster/micro* and do their endpoints also support HEAD?

lxd/storage/backend_lxd.go
```diff
@@ -1439,6 +1439,14 @@ func (b *lxdBackend) RefreshCustomVolume(projectName string, srcProjectName stri
 
 		targetSnapshots := make([]drivers.Volume, 0, len(targetSnaps))
 		for _, targetSnap := range targetSnaps {
+			// Quick check.
+			err := instancetype.ValidName(targetSnap.Name, true)
+			if err != nil {
+				// This should not happen as the target snapshots list is coming from the database.
```
tomponline, line 1445: indeed, why is it needed?

I dont like the fact that it doesnt error here but instead logs only a debug line and then silently stops considering the snapshot.

We should be validating input before we do anything, not midway through the operation.

lxd/api_project.go
```diff
@@ -307,7 +307,7 @@ func projectsPost(d *Daemon, r *http.Request) response.Response {
 	}
 
 	// Quick checks.
-	err = projectValidateName(project.Name)
+	err = projecthelpers.ValidName(project.Name)
```
tomponline, line 310: Although not a change you've introduced here, a note for future change is to stop using the `projecthelpers` import alias and just use `project` like everywhere else in the code base, and then change any shadowing variables in this file to use `p` rather than `project`. 

This is for consistency and helps with code searches.

tomponline (changes requested):
23b974305da4836881d8f197f8a5ae3953a42025 needs reworking too please

lxd/api_project.go
simondeziel: Noted for the next time. Thanks

lxd/storage/backend_lxd.go
simondeziel: I've changed it to error out now.

The sole reason for this commit is to please CodeQL that apparently lost track that the snapshot list came from the DB.

tomponline: hrm, perhaps we should mark that alert as invalid then rather than add low level internal validation for something that has already been validated.

shared/cert.go
simondeziel: I've dropped it, not related indeed.

Those are going point regarding compatibility and I admit making the assumption that HEAD would be permitted/support anywhere GET is. Probably not a given.

Since we are only interested at the TLS layer, I concluded the upper HTTP protocol didn't matter much. That was wrong, as in case of error, we stop before peaking at the TLS info we obtained.

simondeziel: > @simondeziel I'd prefer functional changes like this aren't included in PR that it isnt related to ("Improved validation of project and instance names").

I note you specifically said "functional changes" which I think is very reasonable to keep separated and I'll try to improve that in the future, thanks for pointing that out!

I am often tempted to include minor changes (fmt.Sprintf replacement) like I did in this PR too. Those unfortunately sometimes cause issues as the team has noticed and cleaned up after myself :/

Should I also keep those "minor/drive-by" fixes for separated PRs?  Would that make it easier to review/catch mistakes?

lxd/api_project.go
simondeziel: I'm happy to bundle that `unshadowing` change so that we can avoid the `projecthelpers`.

tomponline: cool thanks

lxd/util/http.go
simondeziel: Fixed for the few occurrences in that PR.

lxd/api_project.go
simondeziel: Hmm, can I backpedal on that one? This being in `lxd/api_project.go`, there are more than 200 occurrences of "project".

shared/cert.go
tomponline: yes probably, it also makes it easier to get the other PRs merged to keep moving things forward.

lxd/project/project.go
```diff
@@ -21,6 +21,10 @@ func ValidName(name string) error {
 		return fmt.Errorf("No name provided")
 	}
 
+	if strings.Contains(name, "\\") {
+		return fmt.Errorf("Project names may not contain backs lashes")
```
tomponline, line 33: "back lashes"

tomponline (approved):

Issue #14175, opened 2024-09-27
LXD OIDC against ADFS should support different URL for issuer and acces_token_issuer 


qch2012:
# Required information

 * Distribution: Ubuntu
 * Distribution version: 20.04, 22.04
 * The output of "snap list --all lxd core20 core22 core24 snapd":
 ```
snap list --all lxd core20 core22 core24 snapd
Name    Version         Rev    Tracking       Publisher   Notes
core20  20240416        2318   latest/stable  canonical✓  base,disabled
core20  20240705        2379   latest/stable  canonical✓  base
core22  20240823        1612   latest/stable  canonical✓  base,disabled
core22  20240904        1621   latest/stable  canonical✓  base
core24  20240528        423    latest/stable  canonical✓  base,disabled
core24  20240710        490    latest/stable  canonical✓  base
lxd     5.21.2-22f93f4  29948  5.21/stable    canonical✓  disabled
lxd     5.21.2-2f4ba6b  30131  5.21/stable    canonical✓  -
```
 * The output of "lxc info" or if that fails:
   * Kernel version: 5.4.0-195-generic
   * LXC version: 5.21.2 LTS
   * LXD version: 5.21.2 LTS
   * Storage backend in use: ZFS

# Issue description

When trying OIDC authentication against ADFS, the LXD server may return issuer does not match error 
```
Error: Failed OIDC Authentication: Failed to authenticate: Failed to verify access token: issuer does not match: Expected: https://ADFS.lab.ob32.com/adfs, got: http://ADFS.lab.ob32.com/adfs/services/trust
```

This is because ADFS by default use different URLs for access_token_issuer and issuer, as seen in the adfs/.well-known/openid-configuration endpoint
```
{
  "issuer": "https://ADFS.lab.ob32.com/adfs",
  ...
  "access_token_issuer": "https://ADFS.lab.ob32.com/adfs/service/trust",
}
```
The access_token_issuer value seems to be taken from the federation service identifier

![image](https://github.com/user-attachments/assets/4989c5e9-c256-442b-81c9-6398ae401864)

So one workaround is to change the federation service id URL to match the issuers URL, but this approach is far from ideal as it is a global setting change that would affect all users 

It seems access_token_issuer is ADFS specific that do not conform to [OIDC spec](https://openid.net/specs/openid-connect-discovery-1_0.html#IssuerDiscovery)
>.  
[...]

Issue #14174, opened 2024-09-27
OIDC via ADFS failed with error "Failed to encrypt refresh token: securecookie: the value is too long"


qch2012:
# Required information

 * Distribution: Ubuntu
 * Distribution version: 20.04
 * The output of "snap list --all lxd core20 core22 core24 snapd":
```
snap list --all lxd core20 core22 core24 snapd
Name    Version         Rev    Tracking       Publisher   Notes
core20  20240416        2318   latest/stable  canonical✓  base,disabled
core20  20240705        2379   latest/stable  canonical✓  base
core22  20240823        1612   latest/stable  canonical✓  base,disabled
core22  20240904        1621   latest/stable  canonical✓  base
core24  20240528        423    latest/stable  canonical✓  base,disabled
core24  20240710        490    latest/stable  canonical✓  base
lxd     5.21.2-22f93f4  29948  5.21/stable    canonical✓  disabled
lxd     5.21.2-2f4ba6b  30131  5.21/stable    canonical✓  -
```
 * The output of "lxc info" or if that fails:
   * Kernel version: 5.4.0-195-generic
   * LXC version:  5.21.2 LTS
   * LXD version: 5.21.2 LTS
   * Storage backend in use: zfs 

# Issue description

ODIC authentication against ADFS failed to return refresh token due to returned token size exceed the cookie size limit

Following errors are seen in browser after login to ADFS 
```
{"type":"error","status":"","status_code":0,"operation":"","error_code":500,"error":"Failed to set login information: Failed to encrypt refresh token: securecookie: the value is too long: 5888","metadata":null}
```

# Steps to reproduce

 1. Step 1 ,  configure application groups in ADFS (in windows 2022 server) 
 
![image](https://github.com/user-attachments/assets/9f428aa0-91c5-429f-a370-6ce372d7211f)

 2. Step 2 ,  configure LXD oidc.client.id  and oidc.issuer
 
![image](https://github.com/user-attachments/assets/9c622109-653d-40d0-ba51-d080b8965029)

 3. Step 3,   open lxd gui https://<lxd fqdn>:8443,   click "login with SSO", the login page should be redirected to ADFS 
 
 4. Step 4,   Login with the AD account in ADFS page,  the browswer returns token exceeding cookie size error 

![image](https://github.com/user-attachments/assets/03243fe8-2b10-4342-b39a-e040741baaa1)

markylaing:
Initial thoughts on this:
[...]

PR #15030, opened 2025-02-19
OIDC: Sessions

markylaing:
This is for discussion and would add direct support for Entra ID and ADFS (#14174 #14175). It:
* Adds logic to set a JWT as a cookie after successfully logging in (both via CLI and UI).
* The JWT is signed using a key derived from the cluster-wide key (added in #15011) using the session ID as a salt.
* The session ID is saved both in the token itself ("sub" claim) and also in the metadata for the identity.
* The identity metadata for OIDC identities now contains fields for the refresh token, for identity provider groups, and for a boolean to indicate if the session has been revoked. These are updated whenever a session times out.
* Sessions can be revoked by session the `session_terminated` field in the identity metadata to true. I haven't implemented an API endpoint for this but it should be easy to do.
* Adds a new config key `oidc.session.lifetime`, which determines the expiry of the JWTs issued by LXD.
* Updates the OIDC Verifier to return an `*api.IdentityInfo` or error. This is set in the request context so that authorization can account for effective groups (LXD and IdP groups). The OIDC Verifier performs DB queries only when the user has authenticated.
* Removes reliance on the identity cache for authorization. It is now used only for mTLS authentication.
* Updates the Authenticate function in `daemon.go` to add an `*api.IdentityInfo` and `*api.Certificate` to the request context when appropriate. (Again, DB calls are made only when the caller has authenticated).
* Updates the `client` and `lxc/config` to handle session cookies (and saves cookies on disk on exit).


tomponline:
@markylaing shall we close this for now to declutter the PR list?

PR #15031, opened 2025-02-20
Network: Fix DHCPv4 after int formatting regression in `LogicalSwitchDHCPv4OptionsSet`

tomponline:
Fixes regression from https://github.com/canonical/lxd/commit/6b1d1200a69befffe107b32124ac0833e0faa306

This caused DHCP options for lease_time to be stored in OVN as `lease_time="3.6µs"` rather than `lease_time="3600"`.

OVN then logged:

 ovn-controller[1939]: ovs|00067|lflow|WARN|error parsing actions "reg0[3] = put_dhcp_opts(offerip = 10.36.76.2, dns_server = {10.123.123.1}, domain_name = "lxd", lease_time = 3.6µs, mtu = 1442, netmask = 255.255.255.0, router = 10.36.76.1, server_id = 10.36.76.1); next;": Invalid numeric constant.

And DHCPv4 broke.


MusicDin (approved):

simondeziel:
> Fixes regression from [6b1d120](https://github.com/canonical/lxd/commit/6b1d1200a69befffe107b32124ac0833e0faa306)

Oops! I remember the test suite catching a very similar issue.

> This caused DHCP options for lease_time to be stored in OVN as `lease_time="3.6µs"` rather than `lease_time="3600"`.

I've put a `XXX:` note in https://github.com/canonical/lxd/blob/main/lxd/network/openvswitch/ovn.go#L586-L598 and even made a typo as it should have read:
> // XXX: Sprint(%d) includes the unit (ns).

Thanks for fixing and finding the right explanation! I'll fix those 2 in a similar way as you did.

PR #15034, opened 2025-02-20
Network: Don't ask OVN to allocate dynamic addresses when DHCP is disabled

tomponline:
This PR changes the way LXD configures OVN's logical switch ports so that it doesn't request `dynamic` IPs for the switch port when `ipv4.dhcp` **and** `ipv6.dhcp` are disabled on the network.

Because OVN doesn't currently allow us to request protocol specific dynamic allocation, LXD has a workaround where if `ipv4.dhcp` is disabled, but `ipv6.dhcp` is enabled, LXD will generate an EUI64 address from the MAC address and set that as a static address in OVN. This avoids OVN needing to also allocate an IPv4 address. It also means that no DNS A records will be created for that port.

However in the case where `ipv4.dhcp` is enabled and `ipv6.dhcp` is false, we need to ask OVN to assign `dynamic` addresses, and so the logical switch port will get both an IPv4 and IPv6 address (and DNS entries for both by extension).

Also fixes a regression from https://github.com/canonical/lxd/pull/14991

### DIFF
--- a/lxd/network/openvswitch/ovn.go
+++ b/lxd/network/openvswitch/ovn.go
@@ -1997,7 +1997,7 @@ func (o *OVN) LoadBalancerApply(loadBalancerName OVNLoadBalancer, routers []OVNR
 			}
 
 			if r.ListenPort > 0 {
-				targetArgs = append(targetArgs, ipToString(target.Address)+":"+fmt.Sprint(target.Port))
+				targetArgs = append(targetArgs, ipToString(target.Address)+":"+strconv.FormatUint(target.Port, 10))
 			} else {
 				targetArgs = append(targetArgs, ipToString(target.Address))
 			}
--- a/lxd/network/openvswitch/ovn.go
+++ b/lxd/network/openvswitch/ovn.go
@@ -1239,21 +1239,26 @@ func (o *OVN) LogicalSwitchPortAdd(switchName OVNSwitch, portName OVNSwitchPort,
 			args = append(args, string(opts.Parent), fmt.Sprint(opts.VLAN))
 		}
 
-		ipStr := make([]string, 0, len(opts.IPs))
-		for _, ip := range opts.IPs {
-			ipStr = append(ipStr, ip.String())
+		var addresses []string
+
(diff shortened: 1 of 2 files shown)


tomponline:
cc @fnordahl

MggMuggins (dismissed):
These changes are consistent with my understanding of the problem. Improved readability as well IMO.

lxd/network/openvswitch/ovn.go
```diff
@@ -1241,21 +1241,26 @@ func (o *OVN) LogicalSwitchPortAdd(switchName OVNSwitch, portName OVNSwitchPort,
...
+			}
+		} else if opts.DHCPv4OptsID != "" || opts.DHCPv6OptsID != "" {
+			// Or if either DHCPv4 or DHCPv6 is enabled then request dynamic IPs for both protocols.
+			// OVN currently doesn't allow us to request protocol specific dynamic IPs.
+			addresses = append(addresses, "dynamic")
 		}
 
-		args = append(args, "--", "lsp-set-addresses", string(portName), addresses)
+		if len(addresses) > 0 {
+			args = append(args, "--", "lsp-set-addresses", string(portName), strings.Join(addresses, " "))
```
simondeziel, line 1262: Is it intentional to join the addresses into an single long space-separated strings rather than append the slice?

simondeziel (dismissed):

lxd/network/openvswitch/ovn.go
tomponline: yep its intended

tomponline: ill double check to be sure, but that is what it was previously anyway and is still working as far as i can tell.

tomponline: although its not clear from the man page, it does seemingly make a difference as dhpv4 didnt work when separate args.

markylaing (approved):
LGTM and well explained, thanks :)

MusicDin (approved):

PR #15036, opened 2025-02-20
Constant time comparisons

simondeziel:


kadinsayani (approved):

kadinsayani (approved):

tomponline (approved):

tomponline:
@simondeziel have these changes been tested?

simondeziel:
> @simondeziel have these changes been tested?

They've been tested by the test suite which immediately turned all red with my first few attempts. While this shows I'm still newb, it's a nice confirmation that what I've been altering was well tested. It also gives me some confidence now that it's all green.

I believe the only part that was not tested was fetching images requiring a secret. Maybe there would be a way to expand our tests for it?

`tl;dr`: It is functional (now at least), but I've not verified if the constant time comparisons are delivering on their promises of not leaking information through timing analysis.

markylaing:
LGTM :rocket: 

However, there is one more place where secrets are being compared. Though in this case it is happening as part of a database query, so I'm not sure it would have the same issue since the time taken will be obfuscated by the time taken to perform the query:
- https://github.com/canonical/lxd/blob/0b56114e4f6fc2e2ebe54f99f942719b92c6e32f/lxd/identities.go#L512
- https://github.com/canonical/lxd/blob/0b56114e4f6fc2e2ebe54f99f942719b92c6e32f/lxd/db/cluster/identities.go#L431-L453

tomponline (approved):

simondeziel:
> Though in this case it is happening as part of a database query, so I'm not sure it would have the same issue since the time taken will be obfuscated by the time taken to perform the query:

The theory is that with enough attempts, the duration of an operation shouldn't matter. You will still be able to extract the information from the noise, even if the DB query time fluctuates itself.

As such, to protect that bit, the secret piece really needs to be compared in constant time. Let's look at it in our 1:1 ;)

simondeziel:
> > Though in this case it is happening as part of a database query, so I'm not sure it would have the same issue since the time taken will be obfuscated by the time taken to perform the query:
> 
> The theory is that with enough attempts, the duration of an operation shouldn't matter. You will still be able to extract the information from the noise, even if the DB query time fluctuates itself.
> 
> As such, to protect that bit, the secret piece really needs to be compared in constant time. Let's look at it in our 1:1 ;)

The DB is doing the comparison so it's probably not worth the effort to mitigate a theoretical problem.

PR #15040, opened 2025-02-21
Network: Fix another regression from moving away from fmt.Sprintf

tomponline:
Fixes missing colon from https://github.com/canonical/lxd/pull/14991


tomponline:
Actually going to just include this fix in https://github.com/canonical/lxd/pull/15034

PR #15041, opened 2025-02-21
Network: Remove unnecessary inclusion of OVN NIC static IPs in logical switch `exclude_ips` setting 

tomponline:
Since https://github.com/canonical/lxd/pull/13900 LXD has used persistent OVN logical switch ports that are created at NIC device add time. The IPv4 address allocated to these ports (whether static or dynamic) act as DHCPv4 reservations (in that OVN's DHCP server won't allocate the same IP to another port), so we no longer need to maintain the exclude_ips list such that it contains all the static IPs from the NICs connected to the network. This was only needed when the logical switch port was removed when the instance device was stopped.


MusicDin (approved):
Nice

simondeziel (approved):
Awesome!

PR #15043, opened 2025-02-21
doc: Improvements to  `cloud-init` docs

hamistao:
Some improvements to the `cloud-init` docs.


tomponline (approved):

doc/cloud-init.md
```diff
@@ -75,11 +75,7 @@ See {ref}`cloud-init:merging_user_data` for instructions.
...
-That means any changes to `cloud-init` configuration will only take effect after the next instance start.
-Some configuration options only take effect on the first boot of an instance. If `cloud-init` notices that an
-instance's `instance-id` has changed, it behaves as if this was the instance's first boot. For more information, see
-the [cloud-init docs](https://docs.cloud-init.io/en/latest/explanation/first_boot.html#first-boot-determination)
-If you are using the CLI client, create the instance with [`lxc init`](lxc_init.md) instead of [`lxc launch`](lxc_launch.md), and then start it after completing the configuration.
+That means any changes to `cloud-init` configuration will only take effect after the next instance start. We make sure `cloud-init` configurations are applied on any boot by changing the instance ID if any configuration keys relevant to `cloud-init` change, as this prompts `cloud-init` to fetch and apply the relevant data from LXD as if it was the instance's first boot. For more information, see the [cloud-init docs](https://docs.cloud-init.io/en/latest/explanation/first_boot.html#first-boot-determination).
```
minaelee, line 78:
```suggestion
This means any changes to `cloud-init` configuration only take effect after the next instance start. To ensure `cloud-init` configurations are applied on every boot, LXD changes the instance ID whenever relevant `cloud-init` configuration keys are modified. This triggers `cloud-init` to fetch and apply the updated data from LXD as if it were the instance's first boot. For more information, see the `cloud-init` docs regarding {ref}`cloud-init:first_boot_determination`.
```

doc/cloud-init.md
```diff
@@ -303,9 +299,9 @@ To inject SSH keys into LXD instances for an arbitrary user, use the configurati
...
+The contents of `cloud-init.ssh-keys.<keyName>` keys are merged into the contents of both `cloud-config.vendor-data` and `cloud-config.user-data` before presenting their content to the guest. This is done according to the [`cloud-config` specification](https://cloudinit.readthedocs.io/en/latest/explanation/about-cloud-config.html). Therefore, keys defined via `cloud-init.ssh-keys.<keyName>` cannot be applied if LXD is unable to parse both the existing `cloud-config.vendor-data` and `cloud-config.user-data` for that instance. This could happen, if those keys are in a format other than YAML or contain badly formatted YAML. Other configuration formats are not yet supported.
```
minaelee, line 302:
```suggestion
The contents of the `cloud-init.ssh-keys.<keyName>` keys are merged into both {config:option}`instance-cloud-init:cloud-init.vendor-data` and {config:option}`instance-cloud-init:cloud-init.user-data` before being passed to the guest, following the {ref}`cloud-config specification <cloud-init:about-cloud-config>`. Therefore, keys defined via `cloud-init.ssh-keys.<keyName>` cannot be applied if LXD cannot parse the existing `cloud-init.vendor-data` and `cloud-init.user-data` for that instance. This might occur if those keys are not in YAML format or contain invalid YAML. Other configuration formats are not yet supported.
```

minaelee (changes requested):

doc/cloud-init.md
hamistao: Both this and the previous comment were addressed

doc/cloud-init.md
```diff
@@ -303,9 +299,9 @@ To inject SSH keys into LXD instances for an arbitrary user, use the configurati
...
 
-You can define SSH keys via `cloud-init.vendor-data` or `cloud-init.user-data` directly. Keys defined using `cloud-init.ssh-keys.<keyName>` do not conflict with those defined with either `cloud-init.vendor-data` or `cloud-init.user-data` in any way. For more information on how to use `cloud-config` to define SSH keys, see [the cloud-init docs for SSH configuration](https://cloudinit.readthedocs.io/en/latest/reference/yaml_examples/ssh.html).
+You can define SSH keys via `cloud-init.vendor-data` or `cloud-init.user-data` directly. Keys defined using `cloud-init.ssh-keys.<keyName>` do not conflict with those defined with either `cloud-init.vendor-data` or `cloud-init.user-data` in any way. For more information on how to use `cloud-config` to define SSH keys, see [the cloud-init docs for SSH configuration](https://cloudinit.readthedocs.io/en/latest/reference/yaml_examples/ssh.html). Previously applied keys will not be removed upon changing a `cloud-init.*` key.
```
minaelee, line 304:
```suggestion
You can define SSH keys via `cloud-init.vendor-data` or `cloud-init.user-data` directly. Keys defined using `cloud-init.ssh-keys.<keyName>` do not conflict with those defined in either of those settings. For details on defining SSH keys with `cloud-config`, see {ref}`the cloud-init documentation for SSH configuration <cloud-init:cce-ssh>`. Changing a `cloud-init.*` key does not remove previously applied keys.
```

hamistao: Done!

minaelee (approved):
Lgtm!

tomponline:
Doc tests failing

hamistao:
Indeed, I assumed it was the same link checks that were failing earlier.

@minaelee the `config` in `{ref}\`cloud-config specification <cloud-init:about-cloud-config>\`` is what is causing this, not sure what to do about it tbh. Any ideas?

minaelee:
> Indeed, I assumed it was the same link checks that were failing earlier.
> 
> @minaelee the second `config` in `{ref}cloud-config specification <cloud-init:about-cloud-config>` is what is causing this, not sure what to do about it tbh. Any ideas?

Can you add `cloud-config` to the list of allowed words in .sphinx/wordlist.txt?

hamistao:
Sorry, I should have mentioned in my previous post.It seems  `cloud-config` is interpreted as two words by the spellchecker, so we would have to add `config` to the wordlist. But that would not be right since  I think `config` is not a worrd we want to allow outside of `cloud-config`.

minaelee:
> Sorry, I should have mentioned in my previous post.It seems `cloud-config` is interpreted as two words by the spellchecker, so we would have to add `config` to the wordlist. But that would not be right since I think `config` is not a word we want to allow outside of `cloud-config`.

I agree; we don't want the spellcheck to ignore `config` in general. There's a `{spellexception}` role that would normally work for one-off inline exceptions from the spellcheck, but it won't work inside a `{ref}` unfortunately. I tested a workaround that passes spellcheck for me locally; I'll suggest it as a change.

doc/cloud-init.md
```diff
@@ -303,9 +299,9 @@ To inject SSH keys into LXD instances for an arbitrary user, use the configurati
...
+The contents of the `cloud-init.ssh-keys.<keyName>` keys are merged into both {config:option}`instance-cloud-init:cloud-init.vendor-data` and {config:option}`instance-cloud-init:cloud-init.user-data` before being passed to the guest, following the {ref}`cloud-config specification <cloud-init:about-cloud-config>`. Therefore, keys defined via `cloud-init.ssh-keys.<keyName>` cannot be applied if LXD cannot parse the existing `cloud-init.vendor-data` and `cloud-init.user-data` for that instance. This might occur if those keys are not in YAML format or contain invalid YAML. Other configuration formats are not yet supported.
```
minaelee, line 302:
```suggestion
The contents of the `cloud-init.ssh-keys.<keyName>` keys are merged into both {config:option}`instance-cloud-init:cloud-init.vendor-data` and {config:option}`instance-cloud-init:cloud-init.user-data` before being passed to the guest, following the `cloud-config` specification. (See the {ref}`cloud-init documentation <cloud-init:about-cloud-config>` for details.) Therefore, keys defined via `cloud-init.ssh-keys.<keyName>` cannot be applied if LXD cannot parse the existing `cloud-init.vendor-data` and `cloud-init.user-data` for that instance. This might occur if those keys are not in YAML format or contain invalid YAML. Other configuration formats are not yet supported.
```

tomponline (approved):

PR #15044, opened 2025-02-21
Network: Add OVS bridge information 

### LINKED ISSUE #14606: GetNetworkState can't detect bridge interfaces configured with OVS
# Required information

We are NOT using LXD but rather LXD as a library, see https://github.com/canonical/maas/blob/7265eeecb055e573d8c40e63b3964c46e657bae6/src/host-info/pkg/info/info.go#L192

# Issue description

[GetNetworkState](https://github.com/canonical/lxd/blob/d08925bb554979182c8ba480b298f71d06174448/lxd/resources/network.go#L477) can't detect a bridge interface configured with OVS. 

It returns 
```
        "br-provider": {
            "addresses": [
                {
                    "family": "inet",
                    "address": "REDACTED",
                    "netmask": "20",
                    "scope": "global"
                },
                {
                    "family": "inet6",
                    "address": "REDACTED",
                    "netmask": "64",
                    "scope": "link"
                }
            ],
            "counters": {
                "bytes_received": 26676298785368,
                "bytes_sent": 3704328664385,
                "packets_received": 3148152714,
                "packets_sent": 111961554
            },
            "hwaddr": "REDACTED",
            "mtu": 1500,
            "state": "up",
            "type": "broadcast",
            "bond": null,
            "bridge": null,                               ----------------------------> This should not be `null`
            "vlan": null,
            "ovn": null
        }
 ``` 

# Steps to reproduce

 1. Step one: Configure a machine with a bridge interface with OVS
 2. Step two: call `GetNetworkState` like in https://github.com/canonical/lxd/blob/main/lxd/resources/network.go#L603-L650

### PR DESCRIPTION
**Pull request by ghadi-rahme**
Adds extra functions to ovs.go to allow for the needed OVS bridge information to be read. 
These are needed for `GetNetworkState()` to be able to retrieve OVS bridge information.  

Fixes #14606

### DIFF
--- a/lxd/network/openvswitch/ovs.go
+++ b/lxd/network/openvswitch/ovs.go
@@ -1,6 +1,7 @@
 package openvswitch
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"os/exec"
@@ -43,7 +44,7 @@ func (o *OVS) Installed() bool {
 
 // BridgeExists returns true if OVS bridge exists.
 func (o *OVS) BridgeExists(bridgeName string) (bool, error) {
-	_, err := shared.RunCommand("ovs-vsctl", "br-exists", bridgeName)
+	_, err := shared.RunCommandContext(context.TODO(), "ovs-vsctl", "br-exists", bridgeName)
 	if err != nil {
 		runErr, ok := err.(shared.RunError)
 		if ok {
@@ -79,7 +80,7 @@ func (o *OVS) BridgeAdd(bridgeName string, mayExist bool, hwaddr net.HardwareAdd
 		args = append(args, "--", "set", "int", bridgeName, fmt.Sprintf(`mtu_request=%d`, mtu))
 	}
 
-	_, err := shared.RunCommand("ovs-vsctl", args...)
+	_, err := shared.RunCommandContext(context.TODO(), "ovs-vsctl", args...)
 	if err != nil {
 		return err
 	}
@@ -89,7 +90,7 @@ func (o *OVS) BridgeAdd(bridgeName string, mayExist bool, hwaddr net.HardwareAdd
 
 // BridgeDelete deletes an OVS bridge.
 func (o *OVS) BridgeDelete(bridgeName string) error {
-	_, err := shared.RunCommand("ovs-vsctl", "del-br", bridgeName)
+	_, err := shared.RunCommandContext(context.TODO(), "ovs-vsctl", "del-br", bridgeName)
 	if err != nil {
 		return err
 	}
@@ -106,8 +107,7 @@ func (o *OVS) BridgePortAdd(bridgeName string, portName string, mayExist bool) e
 	}
 
 	args = append(args, "add-port", bridgeName, portName)
-
-	_, err := shared.RunCommand("ovs-vsctl", args...)
+	_, err := shared.RunCommandContext(context.TODO(), "ovs-vsctl", args...)
 	if err != nil {
 		return err
 	}
@@ -117,7 +117,7 @@ func (o *OVS) BridgePortAdd(bridgeName string, portName string, mayExist bool) e
 
 // BridgePortDelete deletes a port from the bridge (if already detached does nothing).
 func (o *OVS) BridgePortDelete(bridgeName string, portName string) error {
-	_, err := shared.RunCommand("ovs-vsctl", "--if-exists", "del-port", bridgeName, portName)
+	_, err := shared.RunCommandContext(context.TODO(), "ovs-vsctl", "--if-exists", "del-port", bridgeName, portName)
 	if err != nil {
 		return err
 	}
@@ -127,7 +127,7 @@ func (o *OVS) BridgePortDelete(bridgeName string, portName string) error {
 
 // BridgePortSet sets port options.
 func (o *OVS) BridgePortSet(portName string, options ...string) error {
-	_, err := shared.RunCommand("ovs-vsctl", append([]string{"set", "port", portName}, options...)...)
+	_, err := shared.RunCommandContext(context.TODO(), "ovs-vsctl", append([]string{"set", "port", portName}, options...)...)
 	if err != nil {
 		return err
 	}
@@ -139,7 +139,7 @@ func (o *OVS) BridgePortSet(portName string, options ...string) error {
 // and then associates the specified interfaceName to the OVN switch port.
 func (o *OVS) InterfaceAssociateOVNSwitchPort(interfaceName string, ovnSwitchPortName OVNSwitchPort) error {
 	// Clear existing ports that were formerly associated to ovnSwitchPortName.
-	existingPorts, err := shared.RunCommand("ovs-vsctl", "--format=csv", "--no-headings", "--data=bare", "--colum=name", "find", "interface", fmt.Sprintf("external-ids:iface-id=%s", string(ovnSwitchPortName)))
+	existingPorts, err := shared.RunCommandContext(context.TODO(), "ovs-vsctl", "--format=csv", "--no-headings", "--data=bare", "--colum=name", "find", "interface", fmt.Sprintf("external-ids:iface-id=%s", string(ovnSwitchPortName)))
 	if err != nil {
 		return err
 	}
@@ -147,7 +147,7 @@ func (o *OVS) InterfaceAssociateOVNSwitchPort(interfaceName string, ovnSwitchPor
 	existingPorts = strings.TrimSpace(existingPorts)
 	if existingPorts != "" {
 		for _, port := range strings.Split(existingPorts, "\n") {
-			_, err = shared.RunCommand("ovs-vsctl", "del-port", port)
+			_, err = shared.RunCommandContext(context.TODO(), "ovs-vsctl", "del-port", port)
 			if err != nil {
 				return err
 			}
@@ -160,7 +160,7 @@ func (o *OVS) InterfaceAssociateOVNSwitchPort(interfaceName string, ovnSwitchPor
 		}
 	}
 
-	_, err = shared.RunCommand("ovs-vsctl", "set", "interface", interfaceName, fmt.Sprintf("external_ids:iface-id=%s", string(ovnSwitchPortName)))
+	_, err = shared.RunCommandContext(context.TODO(), "ovs-vsctl", "set", "interface", interfaceName, fmt.Sprintf("external_ids:iface-id=%s", string(ovnSwitchPortName)))
 	if err != nil {
 		return err
 	}
@@ -170,7 +170,7 @@ func (o *OVS) InterfaceAssociateOVNSwitchPort(interfaceName string, ovnSwitchPor
 
 // InterfaceAssociatedOVNSwitchPort returns the OVN switch port associated to the OVS interface.
 func (o *OVS) InterfaceAssociatedOVNSwitchPort(interfaceName string) (OVNSwitchPort, error) {
-	ovnSwitchPort, err := shared.RunCommand("ovs-vsctl", "get", "interface", interfaceName, "external_ids:iface-id")
+	ovnSwitchPort, err := shared.RunCommandContext(context.TODO(), "ovs-vsctl", "get", "interface", interfaceName, "external_ids:iface-id")
 	if err != nil {
 		return "", err
 	}
@@ -184,7 +184,7 @@ func (o *OVS) ChassisID() (string, error) {
 	// However ovs-vsctl's find and list commands don't support retrieving a single column's map field.
 	// And ovs-vsctl's JSON output is unfriendly towards statically typed languages as it mixes data types
 	// in a slice. So stick with "get" command and use Go's strconv.Unquote to return the actual values.
-	chassisID, err := shared.RunCommand("ovs-vsctl", "get", "open_vswitch", ".", "external_ids:system-id")
+	chassisID, err := shared.RunCommandContext(context.TODO(), "ovs-vsctl", "get", "open_vswitch", ".", "external_ids:system-id")
 	if err != nil {
 		return "", err
 	}
@@ -204,7 +204,7 @@ func (o *OVS) OVNEncapIP() (net.IP, error) {
 	// However ovs-vsctl's find and list commands don't support retrieving a single column's map field.
 	// And ovs-vsctl's JSON output is unfriendly towards statically typed languages as it mixes data types
 	// in a slice. So stick with "get" command and use Go's strconv.Unquote to return the actual values.
-	encapIPStr, err := shared.RunCommand("ovs-vsctl", "get", "open_vswitch", ".", "external_ids:ovn-encap-ip")
+	encapIPStr, err := shared.RunCommandContext(context.TODO(), "ovs-vsctl", "get", "open_vswitch", ".", "external_ids:ovn-encap-ip")
 	if err != nil {
 		return nil, err
 	}
@@ -229,7 +229,7 @@ func (o *OVS) OVNBridgeMappings(bridgeName string) ([]string, error) {
 	// However ovs-vsctl's find and list commands don't support retrieving a single column's map field.
 	// And ovs-vsctl's JSON output is unfriendly towards statically typed languages as it mixes data types
 	// in a slice. So stick with "get" command and use Go's strconv.Unquote to return the actual values.
-	mappings, err := shared.RunCommand("ovs-vsctl", "--if-exists", "get", "open_vswitch", ".", "external-ids:ovn-bridge-mappings")
+	mappings, err := shared.RunCommandContext(context.TODO(), "ovs-vsctl", "--if-exists", "get", "open_vswitch", ".", "external-ids:ovn-bridge-mappings")
 	if err != nil {
 		return nil, err
 	}
@@ -267,7 +267,7 @@ func (o *OVS) OVNBridgeMappingAdd(bridgeName string, providerName string) error 
 	mappings = append(mappings, newMapping)
 
 	// Set new mapping string back into OVS database.
-	_, err = shared.RunCommand("ovs-vsctl", "set", "open_vswitch", ".", fmt.Sprintf("external-ids:ovn-bridge-mappings=%s", strings.Join(mappings, ",")))
+	_, err = shared.RunCommandContext(context.TODO(), "ovs-vsctl", "set", "open_vswitch", ".", fmt.Sprintf("external-ids:ovn-bridge-mappings=%s", strings.Join(mappings, ",")))
 	if err != nil {
 		return err
 	}
@@ -299,13 +299,13 @@ func (o *OVS) OVNBridgeMappingDelete(bridgeName string, providerName string) err
 	if changed {
 		if len(newMappings) < 1 {
 			// Remove mapping key in OVS database.
-			_, err = shared.RunCommand("ovs-vsctl", "remove", "open_vswitch", ".", "external-ids", "ovn-bridge-mappings")
+			_, err = shared.RunCommandContext(context.TODO(), "ovs-vsctl", "remove", "open_vswitch", ".", "external-ids", "ovn-bridge-mappings")
 			if err != nil {
 				return err
 			}
 		} else {
 			// Set updated mapping string back into OVS database.
-			_, err = shared.RunCommand("ovs-vsctl", "set", "open_vswitch", ".", fmt.Sprintf("external-ids:ovn-bridge-mappings=%s", strings.Join(newMappings, ",")))
+			_, err = shared.RunCommandContext(context.TODO(), "ovs-vsctl", "set", "open_vswitch", ".", fmt.Sprintf("external-ids:ovn-bridge-mappings=%s", strings.Join(newMappings, ",")))
 			if err != nil {
 				return err
 			}
@@ -318,7 +318,7 @@ func (o *OVS) OVNBridgeMappingDelete(bridgeName string, providerName string) err
 // BridgePortList returns a list of ports that are connected to the bridge.
 func (o *OVS) BridgePortList(bridgeName string) ([]string, error) {
 	// Clear existing ports that were formerly associated to ovnSwitchPortName.
-	portString, err := shared.RunCommand("ovs-vsctl", "list-ports", bridgeName)
+	portString, err := shared.RunCommandContext(context.TODO(), "ovs-vsctl", "list-ports", bridgeName)
 	if err != nil {
 		return nil, err
 	}
@@ -341,7 +341,7 @@ func (o *OVS) HardwareOffloadingEnabled() bool {
 	// However ovs-vsctl's find and list commands don't support retrieving a single column's map field.
 	// And ovs-vsctl's JSON output is unfriendly towards statically typed languages as it mixes data types
 	// in a slice. So stick with "get" command and use Go's strconv.Unquote to return the actual values.
-	offload, err := shared.RunCommand("ovs-vsctl", "--if-exists", "get", "open_vswitch", ".", "other_config:hw-offload")
+	offload, err := shared.RunCommandContext(context.TODO(), "ovs-vsctl", "--if-exists", "get", "open_vswitch", ".", "other_config:hw-offload")
 	if err != nil {
 		return false
 	}
@@ -361,7 +361,7 @@ func (o *OVS) HardwareOffloadingEnabled() bool {
 
 // OVNSouthboundDBRemoteAddress gets the address of the southbound ovn database.
 func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
-	result, err := shared.RunCommand("ovs-vsctl", "get", "open_vswitch", ".", "external_ids:ovn-remote")
+	result, err := shared.RunCommandContext(context.TODO(), "ovs-vsctl", "get", "open_vswitch", ".", "external_ids:ovn-remote")
 	if err != nil {
 		return "", err
 	}
--- a/lxd/network/openvswitch/ovs.go
+++ b/lxd/network/openvswitch/ovs.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"os/exec"
+	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ghadi-rahme** [comment]
### Size increase justification
Functions were added to the `openvswitch` package, which itself depends on other packages. The `openvswitch` package was then imported into the `network.go` file in the `resources` package. This leads to an increase in the packages used by lxd-agent causing the size of the binary to increase.

File: lxd/network/openvswitch/ovs.go [hunk 1]
```diff
@@ -373,3 +374,151 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
 
 	return addr, nil
 }
+
+// Default STP priority is 32768 (0x8000). This value will be used if the
+// priority cannot be retrieved or if STP is disabled
+func GetStpPriority(bridgeName string) (string, error) {
+	output, err := shared.RunCommand("ovs-appctl", "stp/show", bridgeName)
```
markylaing, line 381: I think `RunCommand` is deprecated. Instead you can use `RunCommandContext` and pass a `context.Context` into this method.

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +374,151 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+	if err != nil {
```
markylaing, line 382: Please can you use `errors.Is` and/or `errors.As` for these kind of checks.

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +374,151 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+		runErr, ok := err.(shared.RunError)
+		if ok {
+			exitError, ok := runErr.Unwrap().(*exec.ExitError)
+
+			// STP is disabled for that bridge
+			if ok && exitError.ExitCode() == 2 {
+				return "8000", nil
+			}
+		}
+		return "8000", err
```
markylaing, line 392: You will always need a newline after a closing brace and before any subsequent statements.

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +374,151 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+		}
+	}
+
+	// convert the priority to hexadecimal format
+	decimalStp, err := strconv.Atoi(stpPriority)
+	if err != nil {
+		return "8000", err
+	}
+	hexStp := fmt.Sprintf("%4X", decimalStp)
+	return hexStp, nil
```
markylaing, line 421: Why do we return the STP as a hex string?

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +374,151 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+		}
+		if bridgeSection {
+			if strings.HasPrefix(line, "stp-priority") {
+				sections := strings.Fields(line)
+				if len(sections) > 1 {
+					stpPriority = sections[1]
+				}
+			}
+		}
+	}
```
markylaing, line 413: Most `ovs-*` commands have options for formatting. Are there any options you can use here to simplify the output format? Or perhaps rather than calling `ovs-appctl stp/show` there are `ovs-vsctl` commands you can use to get the value directly from the database?

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +374,151 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+		return "8000", err
+	}
+	hexStp := fmt.Sprintf("%4X", decimalStp)
+	return hexStp, nil
+
+}
+
+// The bridge IDs follow the following format <STP priority>.<MAC Address>
+// The STP priority is in hexadecimal format
+func (o *OVS) GenerateOVSBridgeID(bridgeName string) (string, error) {
```
markylaing, line 427: Why? Is there a link to some documentation we could link to here?

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +374,151 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+	// Get the STP priority
+	stpPriority, err := GetStpPriority(bridgeName)
+	if err != nil {
+		return "None", err
+	}
+
+	return stpPriority + "." + bridgeHwID, nil
+}
+
+func (o *OVS) StpEnabled(bridgeName string) (bool, error) {
```
markylaing, line 444: These exported functions will all need to be commented in godoc style

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +374,151 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+			}
+		}
+	}
+
+	// remove the "s" at the end
+	stpOvsDelay = stpOvsDelay[:len(stpOvsDelay)-1]
+
+	stpFwDelay, err := strconv.ParseUint(stpOvsDelay, 10, 64)
+	return stpFwDelay * 1000, err
+
```
markylaing, line 483: Unnecessary newline

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +374,151 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+				}
+			}
+		}
+	}
+
+	// remove the "s" at the end
+	stpOvsDelay = stpOvsDelay[:len(stpOvsDelay)-1]
+
+	stpFwDelay, err := strconv.ParseUint(stpOvsDelay, 10, 64)
+	return stpFwDelay * 1000, err
```
markylaing, line 482: Why the multiplier? I'm guessing that the delay is reported by OVS in seconds and you would like it in milliseconds?

lxd/resources/network.go
```diff
@@ -605,54 +606,106 @@ func GetNetworkState(name string) (*api.NetworkState, error) {
...
+	if ovs.Installed() {
+		exists, err = ovs.BridgeExists(name)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	bridge := api.NetworkStateBridge{}
+
+	if !exists {
```
markylaing, line 621: `exists` is a confusing variable name. Can you please call it something like `isOVSBridge`?

lxd/resources/network.go
```diff
@@ -605,54 +606,106 @@ func GetNetworkState(name string) (*api.NetworkState, error) {
...
-				}
-			}
+		//Upper devices
+		entries, err := ovs.BridgePortList(name)
+		if err == nil {
+			bridge.UpperDevices = append(bridge.UpperDevices, entries...)
 		}
-
 		network.Bridge = &bridge
+
```
markylaing, line 708: Rather than this large if/else block, can you please refactor to two functions: `getOVSBridgeState` and `getNativeBridgeState`?

lxd/state/notlinux.go
```diff
@@ -5,12 +5,14 @@ package state
 import (
 	"context"
 
+	clusterConfig "github.com/canonical/lxd/lxd/cluster/config"
```
markylaing, line 8: We don't wan't to add anything to this file that isn't absolutely necessary for the LXD agent to function. I suggest if `state.State` is causing issues, refactor the `resources` package to not require it (e.g. only pass in the fields of `state.State` that are actually used).

markylaing (changes requested):
Overall I'm happy to add these functions to the `openvswitch` package but the parsing of `ovs-appctl stp/show` output could use some work. Ideally a format can be specified or `ovs-vsctl` can be used instead. Additionally, I think the best long-term solution for having `lxd/resources` as a dependency will be to refactor the package so that it does not import `state.State`.

lxd/network/openvswitch/ovs.go
ghadi-rahme: The reason is because the IEEE802.1D standard represents the STP priority in binary terms. Linux (and many switches) already uses the hexadecimal format to represent them but OVS defaults to showing the value in base 10. Also to represent a bridge ID which is a combination of the STP priority and the MAC address [1], the STP will need to be in hex format just like the MAC address is. In fact IEEE802.1D is the same standard that defines the MAC address as also being represented in binary format (hexadecimal is used on MAC addresses to simplify the representation, the same is done for STP priority).

[1] https://www.kernel.org/doc/Documentation/networking/bridge.rst
```
Bridge ID: The Bridge ID is composed of two components: Bridge Priority
     and the MAC address of the bridge. It uniquely identifies each bridge
     in the network. The Bridge ID is used to compare the priorities of
     different bridges.
```

lxd/network/openvswitch/ovs.go
ghadi-rahme: I have rewrote this function to make use of `ovs-vsctl`. I had avoided this approach before  because the error codes returned when querying the DB for config variables do not differentiate between a bad variable name (not used)  and an undefined variable.

lxd/network/openvswitch/ovs.go
ghadi-rahme: Yes I will add a reference in the source

lxd/network/openvswitch/ovs.go
markylaing: Thanks. Ultimately all of these functions will be migrated to use `libovsdb`. When that happens, MAAS will need to instantiate and use the `openvswitch` package differently. Even the import path will change.

lxd/state/notlinux.go
ghadi-rahme: I am not sure how feasible this is without splitting the` openvswitch` package in two packages. 
The `state.State` dependency comes from importing `openvswitch` into the `network.go` file and openvswitch makes use of `state.State`. 
The only other solution I could think off would be to move the functions in the `ovs.go`  into their own separate package and import that into `network.go`, which I was already told not to do.
Let me know if you are aware of any other workarounds or how you would like me to proceed.

ghadi-rahme:
> > ### Size increase justification
> > Functions were added to the `openvswitch` package, which itself depends on other packages. The `openvswitch` package was then imported into the `network.go` file in the `resources` package. This leads to an increase in the packages used by lxd-agent causing the size of the binary to increase.
> 
> IIRC we dont want lxd-agent needing to depend on ovs package

`resources` needs to access these function in order to retrieve the data from ovs. One solution I though off was to split the `openvswitch` package in two, but I was told not to do it (and I agree it does make things messy).

lxd/state/notlinux.go
ghadi-rahme: Also one extra note on the scenario of splitting the `openvswitch` package in two. If `ovs.go` were to be in a separate package from `ovn.go`, it would still need to import the `openvswitch` package and result in `state.State` still being used because functions such as `OVNBridgeMappingAdd` and `OVNBridgeMappings` make calls to functions present in `ovn.go` .
This means that these functions will need to be moved back to the `openvswitch` package resulting in ovs related tasks being split across two packages.... which is messy.

markylaing: The only usage of `state.State` is where it is passed into `openvswitch.NewOVN`, where it access `(*State).GlobalConfig`. I have a PR to remove it and once merged you should be able to use the OVS utils with no issues.

markylaing:
Once #15084 is merged and this is rebased I'll take another look.

ghadi-rahme:
I applied the requested changes. I also made the context a timeout context with a 5s duration for each exec function. 
Let me know if the current context is good or if I should switch to `shared.RunCommandLocal` maybe.
Hopefully I didn't miss anything.

markylaing:
> I applied the requested changes. I also made the context a timeout context with a 5s duration for each exec function. Let me know if the current context is good or if I should switch to `shared.RunCommandLocal` maybe. Hopefully I didn't miss anything.

Thanks for making the changes. As for the context, can you just make this an argument to the OVS functions? Thanks.

ghadi-rahme:
I squashed the commits and made the ctx an argument, however the tests failing seem to be a result of some sort of network error unrelated to the changes. Would it be possible to restart the test?

markylaing (dismissed):

lxd/network/openvswitch/ovs.go
```diff
@@ -42,8 +44,8 @@ func (o *OVS) Installed() bool {
 }
 
 // BridgeExists returns true if OVS bridge exists.
-func (o *OVS) BridgeExists(bridgeName string) (bool, error) {
-	_, err := shared.RunCommand("ovs-vsctl", "br-exists", bridgeName)
+func (o *OVS) BridgeExists(ctx context.Context, bridgeName string) (bool, error) {
+	_, err := shared.RunCommandContext(ctx, "ovs-vsctl", "br-exists", bridgeName)
```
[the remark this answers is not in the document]
markylaing: I've double checked and yes, changes to `lxd/network/openvswitch/ovs.go` causes static analysis to fail because `shared.RunCommand` is deprecated. There is also no clear context to pass in from the callers (OVN network driver, OVN network instance device, OVN ACLs). 

The simplest solution will be to replace `RunCommand` with `RunCommandContext` in `ovs.go` and pass in a `context.TODO()`. Thanks

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +373,137 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
 
 	return addr, nil
 }
+
+// GetStpPriority returns the STP priority of the OVS bridge.
+// Default STP priority is 32768 (0x8000). This value will be used if the
+// priority cannot be retrieved or if STP is disabled.
+func GetStpPriority(ctx context.Context, bridgeName string) (string, error) {
```
tomponline, line 380: Should be GetSTPPriority as acronyms should be uppercased.

https://go.dev/talks/2014/names.slide#5
Initialisms[¶](https://go.dev/wiki/CodeReviewComments#initialisms)

lxd/network/openvswitch/ovs.go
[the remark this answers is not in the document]
ghadi-rahme: That is correct the analysis does fail with a deprecation error. I had originally only modified the extra functions I added, but after running the static analysis checks I got the deprecation errors and decided to change the rest of the functions as well.

tomponline (changes requested):
Please can you avoid creating unnecessary ctxN vars and instead you can shadow the existing ctx var if you are creating a new variable within a limited scope.

lxd/networks.go
```diff
@@ -14,7 +14,7 @@ import (
 
 	"github.com/gorilla/mux"
 
-	"github.com/canonical/lxd/client"
+	lxd "github.com/canonical/lxd/client"
```
tomponline, line 17: please remove this change

lxd/device/nic_bridged.go
tomponline, line 1: The commit message doesnt align with the actual change, "Add the following OVS features:..." isnt present in the actual code change.

tomponline: Please can you split these context changes in to 1 commit per file or at most 1 commit per package. Thanks

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +373,137 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+	if err != nil {
+		return 15 * 1000, err
+	}
+
+	return stpFwdDelay * 1000, nil
+}
+
+// VlanFilteringEnabled checks if a vlans are enabled on the OVS bridge.
+// In OVS, Vlan filtering is enabled when Vlan related settings are configured.
+func (o *OVS) VlanFilteringEnabled(ctx context.Context, bridgeName string) (bool, error) {
```
tomponline, line 472: VLANFilteringEnabled

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +373,137 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+// GetVlanPvid returbs the PVID of the ovs bridge.
+// In OVS a PVID of 0 means that the port is not associated with any VLAN.
+func (o *OVS) GetVlanPvid(ctx context.Context, bridgeName string) (uint64, error) {
```
tomponline, line 492: GetVLANPVID

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +373,137 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+
+	output = strings.TrimSpace(output)
+	return strconv.ParseBool(output)
+}
+
+// GetStpForwardDelay returns the STP forward delay in ms. OVS returns the value in seconds, so it needs to be
+// converted to ms to satisfy the api.NetworkStateBridge struct which expects the value in ms.
+// Default value is 15s.
+// Check the "other_config : stp-forward-delay:" section on http://www.openvswitch.org/support/dist-docs/ovs-vswitchd.conf.db.5.txt
+func (o *OVS) GetStpForwardDelay(ctx context.Context, bridgeName string) (uint64, error) {
```
tomponline, line 444: GetSTPForwardDelay

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +373,137 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+	if err != nil {
+		return "None", err
+	}
+
+	return stpPriority + "." + bridgeHwID, nil
+}
+
+// StpEnabled checks if STP is enabled by looking up the "stp_enable" boolean config variable.
+// Returns the value stored in "stp_enable", or false if it is undefined.
+func (o *OVS) StpEnabled(ctx context.Context, bridgeName string) (bool, error) {
```
tomponline, line 430: STPEnabled

lxd/device/nic_bridged.go
ghadi-rahme: I will be undoing all the context calls for the functions that were already present in `ovs.go` as per @markylaing requested. In the end there will only be two files modified: `ovs.go` and `lxd/resources/networks.go` by this PR

tomponline: sounds good to me

tomponline:


Looks like you have not signed your commits.

ghadi-rahme:
Let me know how everything looks, and than you for the reviews!

ghadi-rahme:
Quick side note. When running `make static-analysis` I keep running into an error asking me to verify imports. To get around it I have been removing these imports to get the analysis to go through locally then undoing the changes to them before pushing the commits. In the github runners it seems to pass fine so not sure what I need to do to make it ignore these imports on my end. Below is the output I get:
```
==> Checking for imports/deps that have been added to client...
ERROR: you added a new dependency to client; please make sure this is what you want
--- test/godeps/client.list     2025-03-04 13:59:10.592828763 +0000
+++ -   2025-03-04 14:17:07.704694926 +0000
@@ -75,10 +75,8 @@
 vendor/golang.org/x/crypto/chacha20poly1305
 vendor/golang.org/x/crypto/cryptobyte
 vendor/golang.org/x/crypto/cryptobyte/asn1
-vendor/golang.org/x/crypto/hkdf
 vendor/golang.org/x/crypto/internal/alias
 vendor/golang.org/x/crypto/internal/poly1305
-vendor/golang.org/x/crypto/sha3
 vendor/golang.org/x/net/dns/dnsmessage
 vendor/golang.org/x/net/http/httpguts
 vendor/golang.org/x/net/http/httpproxy

==> Checking for imports/deps that have been added to lxc/config...
ERROR: you added a new dependency to lxc/config; please make sure this is what you want
--- test/godeps/lxc-config.list 2025-03-04 13:59:10.592828763 +0000
+++ -   2025-03-04 14:17:07.812829823 +0000
@@ -77,10 +77,8 @@
 vendor/golang.org/x/crypto/chacha20poly1305
 vendor/golang.org/x/crypto/cryptobyte
 vendor/golang.org/x/crypto/cryptobyte/asn1
-vendor/golang.org/x/crypto/hkdf
 vendor/golang.org/x/crypto/internal/alias
 vendor/golang.org/x/crypto/internal/poly1305
-vendor/golang.org/x/crypto/sha3
 vendor/golang.org/x/net/dns/dnsmessage
 vendor/golang.org/x/net/http/httpguts
 vendor/golang.org/x/net/http/httpproxy

==> Checking for imports/deps that have been added to lxd-agent...
ERROR: you added a new dependency to lxd-agent; please make sure this is what you want
--- test/godeps/lxd-agent.list  2025-03-04 13:59:10.592828763 +0000
+++ -   2025-03-04 14:17:07.929844444 +0000
@@ -116,10 +116,8 @@
 vendor/golang.org/x/crypto/chacha20poly1305
 vendor/golang.org/x/crypto/cryptobyte
 vendor/golang.org/x/crypto/cryptobyte/asn1
-vendor/golang.org/x/crypto/hkdf
 vendor/golang.org/x/crypto/internal/alias
 vendor/golang.org/x/crypto/internal/poly1305
-vendor/golang.org/x/crypto/sha3
 vendor/golang.org/x/net/dns/dnsmessage
 vendor/golang.org/x/net/http/httpguts
 vendor/golang.org/x/net/http/httpproxy

==> Checking for imports/deps that have been added to shared/api...
ERROR: you added a new dependency to shared/api; please make sure this is what you want
--- test/godeps/shared-api.list 2025-03-04 13:59:10.592828763 +0000
+++ -   2025-03-04 14:17:08.072605510 +0000
@@ -2,10 +2,8 @@
 vendor/golang.org/x/crypto/chacha20poly1305
 vendor/golang.org/x/crypto/cryptobyte
 vendor/golang.org/x/crypto/cryptobyte/asn1
-vendor/golang.org/x/crypto/hkdf
 vendor/golang.org/x/crypto/internal/alias
 vendor/golang.org/x/crypto/internal/poly1305
-vendor/golang.org/x/crypto/sha3
 vendor/golang.org/x/net/dns/dnsmessage
 vendor/golang.org/x/net/http/httpguts
 vendor/golang.org/x/net/http/httpproxy
run-parts: test/lint/godeps.sh exited with return code 1
make: *** [Makefile:299: static-analysis] Error 1
```
These were there since before my changes as well.

lxd/network/openvswitch/ovs.go
tomponline, line 1: Please split this commit into 2, one for switching to RunCommandContext in existing functions in this file and the 2nd to add the new functions.

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +373,137 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
 
 	return addr, nil
 }
+
+// GetSTPPriority returns the STP priority of the OVS bridge.
+// Default STP priority is 32768 (0x8000). This value will be used if the
+// priority cannot be retrieved or if STP is disabled.
```
tomponline, line 379: no need to wrap this line

lxd/network/openvswitch/ovs.go
ghadi-rahme: Apologies, I though this was only to be done to the functions called outside the openvswitch package before I changed them to use `context.TODO()`. I will do that change now

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +373,137 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+	// get bridge does not differentiate in it's error codes between variable undefined and other errors, such as bridge not found
+	// the error output will need to be checked to make sure if stp-priority is not defined (defaults to 0x8000) or if a breaking error happened
+	if err != nil {
+		// default stp-priority is used if stp-priority is not defined
+		if strings.Contains(err.Error(), fmt.Sprintf("no key \"stp-priority\" in Bridge record \"%s\" column other_config", bridgeName)) {
+			return "8000", nil
+		}
+
+		// All other errors are considered errors
+		return "8000", err
```
tomponline, line 391: why are you returning the hex value in as a string rather than the decimal value as an int?

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +373,137 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+	output = strings.TrimSpace(output)
+	// remove surrounding quotes
+	output = output[1 : len(output)-1]
+	stpPriority, err := strconv.Atoi(output)
+	if err != nil {
+		return "8000", err
+	}
+
+	hexStp := fmt.Sprintf("%4X", stpPriority)
+	return hexStp, nil
```
tomponline, line 404: why return the hex value as a string rather than as an base 10 int?

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +373,137 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+// Default value is 15s.
+// Check the "other_config : stp-forward-delay:" section on http://www.openvswitch.org/support/dist-docs/ovs-vswitchd.conf.db.5.txt
+func (o *OVS) GetSTPForwardDelay(ctx context.Context, bridgeName string) (uint64, error) {
+	output, err := shared.RunCommandContext(ctx, "ovs-vsctl", "get", "bridge", bridgeName, "other_config:stp-forward-delay")
+	// get bridge does not differentiate in it's error codes between variables undefined and other errors, such as bridge not found
+	// the error output will need to be checked to make sure if stp-forward-delay is not defined (defaults to 15000) or if a breaking error happened
+	if err != nil {
+		// default stp-priority is used if stp-priority is not defined
+		if strings.Contains(err.Error(), fmt.Sprintf("no key \"stp-forward-delay\" in Bridge record \"%s\" column other_config", bridgeName)) {
+			return 15 * 1000, nil
```
tomponline, line 451: Please can you define a constant for this once rather than defining it multiple times.

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +373,137 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+		return "8000", err
+	}
+
+	// Convert to hex format
+	output = strings.TrimSpace(output)
+	// remove surrounding quotes
+	output = output[1 : len(output)-1]
+	stpPriority, err := strconv.Atoi(output)
+	if err != nil {
+		return "8000", err
```
tomponline, line 400: Please can you define a constant for this once rather than defining it multiple times.

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +373,137 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+// converted to ms to satisfy the api.NetworkStateBridge struct which expects the value in ms.
+// Default value is 15s.
+// Check the "other_config : stp-forward-delay:" section on http://www.openvswitch.org/support/dist-docs/ovs-vswitchd.conf.db.5.txt
+func (o *OVS) GetSTPForwardDelay(ctx context.Context, bridgeName string) (uint64, error) {
+	output, err := shared.RunCommandContext(ctx, "ovs-vsctl", "get", "bridge", bridgeName, "other_config:stp-forward-delay")
+	// get bridge does not differentiate in it's error codes between variables undefined and other errors, such as bridge not found
+	// the error output will need to be checked to make sure if stp-forward-delay is not defined (defaults to 15000) or if a breaking error happened
+	if err != nil {
+		// default stp-priority is used if stp-priority is not defined
+		if strings.Contains(err.Error(), fmt.Sprintf("no key \"stp-forward-delay\" in Bridge record \"%s\" column other_config", bridgeName)) {
```
tomponline, line 450: errors should start with upper case letter please

ghadi-rahme: The message `no key...` is not a returned error by this code but it is comparing the output returned by OVS if the variable is not defined in the DB, I cannot change it to be `No key...` or else the comparison will fail.

tomponline: right got you, thanks

lxd/network/openvswitch/ovs.go
tomponline: In general please try to separate logical changes into separate commits.

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +373,137 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+// Check the Bridge ID section on https://www.kernel.org/doc/Documentation/networking/bridge.rst.
+// The STP priority is in hexadecimal format just like the MAC address.
+func (o *OVS) GenerateOVSBridgeID(ctx context.Context, bridgeName string) (string, error) {
+	// get the MAC address
+	netIf, err := net.InterfaceByName(bridgeName)
+	if err != nil {
+		return "None", err
+	}
+
+	bridgeHwID := strings.ReplaceAll(strings.ToLower(netIf.HardwareAddr.String()), ":", "")
```
tomponline, line 418: @ghadi-rahme does OVS allow accessing the bridge ID rather than us generating it?

lxd/network/openvswitch/ovs.go
ghadi-rahme: The reason is because this function is used to compute the Bridge ID, which expects the STP to be in hex form. Although I have this function exported, I should have left it local since it is only used by `GenerateOVSBridgeID`.
I could have it return a base 10 and reconvert the value in `GenerateOVSBridgeID` if that is preferred. What doe you think? Also feedback on if I should keep this global or not is appreciated.

ghadi-rahme: Another reason to keep it in hex form is that STP is defined in binary form just like MAC addresses, https://github.com/canonical/lxd/pull/15044#discussion_r1971690548
But again I don't mind going with either solution.

tomponline: My preference is that it shouldnt be a string at least.


Probably this function should also not be exported if its just used for an internal sub-step.

tomponline: My preference would be to use base 10 int (or smaller), so 32768, and then convert to hex where needed.

lxd/network/openvswitch/ovs.go
ghadi-rahme: OVS does allow returning the UUID of a bridge, but it has a completely different form of what a linux bridge ID is or what a bridge ID in general should look like. Generally when talking about bridge ID's it is of the form stp-prio.MAC-add. Which to my understanding is an IEEE 802.1D standard and used by multiple vendors:
https://docs.oracle.com/cd/E39109_01/html/E24664/z400276b1406190.html

lxd/network/openvswitch/ovs.go
ghadi-rahme: sounds good! I think I had it exported originally while running tests on it and forgot to set it back to local

lxd/network/openvswitch/ovs.go
tomponline: What purpose does returning a non-OVS bridge ID serve in this case (for LXD's `NetworkState` struct) - are you just trying to keep to the same format as the native bridge uses?

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +373,137 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+// GetSTPPriority returns the STP priority of the OVS bridge.
+// Default STP priority is 32768 (0x8000). This value will be used if the
+// priority cannot be retrieved or if STP is disabled.
+func GetSTPPriority(ctx context.Context, bridgeName string) (string, error) {
+	output, err := shared.RunCommandContext(ctx, "ovs-vsctl", "get", "bridge", bridgeName, "other_config:stp-priority")
+	// get bridge does not differentiate in it's error codes between variable undefined and other errors, such as bridge not found
+	// the error output will need to be checked to make sure if stp-priority is not defined (defaults to 0x8000) or if a breaking error happened
+	if err != nil {
+		// default stp-priority is used if stp-priority is not defined
+		if strings.Contains(err.Error(), fmt.Sprintf("no key \"stp-priority\" in Bridge record \"%s\" column other_config", bridgeName)) {
```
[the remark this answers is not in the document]
ghadi-rahme: I will have to use the `\"stp-priority\"` syntax instead of the backticks to be able to match the output of OVS.
%q is fine.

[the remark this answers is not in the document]
tomponline: does it output the error with backslashes then?

lxd/network/openvswitch/ovs.go
ghadi-rahme: Also although a bridge ID is also used for identifying a bridge, it also serves as a purpose of electing the root bridge, and it does so by looking at the priority and then the MAC address if the priorities are equal between the different bridges. This is why the bridge ID is formed as `Priority.MAC` because it turns this theoretical 2 stage comparison into only comparing the bridge ID (The lowest Bridge ID will be the root bridge). 
While a UUID can only be used for identification. 
This quirk and extra usage of a bridge ID is what made me generate it for OVS bridges.

lxd/network/openvswitch/ovs.go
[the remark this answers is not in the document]
ghadi-rahme: Double quotes are used by the ovs output as seen above, this means that if I were to use the backticks, the comparison would fail.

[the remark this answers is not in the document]
tomponline:
```go
if strings.Contains(err.Error(), fmt.Sprintf(`no key "stp-priority" in Bridge record %q column other_config`, bridgeName)) {
```

lxd/network/openvswitch/ovs.go
tomponline: > This is why the bridge ID is formed as `Priority.MAC` because it turns this theoretical 2 stage comparison into only comparing the bridge ID (The lowest Bridge ID will be the root bridge).
> While a UUID can only be used for identification.
> This quirk and extra usage of a bridge ID is what made me generate it for OVS bridges.

I didnt quite follow what you meant here, please can you expand?

lxd/network/openvswitch/ovs.go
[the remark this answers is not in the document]
ghadi-rahme: Ahh right... been doing a lot of C, forgot go accepts backticks instead of double quotes for strings

lxd/network/openvswitch/ovs.go
ghadi-rahme: When selecting the root bridge, the STP algorithm first checks who has the lowest priority and elects the one with the lowest one. If they happen to have the same priority, the comparison will be done on the MAC address instead and the one with the lowest MAC address will be chosen as the root bridge.
To avoid the scenario where a second comparison will need to be made if the priorities are the same, the bridge ID is compared since it contains both the priority and the MAC address, meaning that only one comparison is needed. As an example

Scenario 1:
br0 has priority 0x8000 and mac FFFFFFFFFFF1. The ID is 8000FFFFFFFFFFF1
br1 has priority 0x8000 and mac FFFFFFFFFFF2. The ID is 8000FFFFFFFFFFF2

The ID of br0 < br1 so br0 is the root bridge. Only the full bridge ID value can be compared instead of comparing the priority, noticing that they are the same and moving on to comparing the MAC address.

Scenario 2:
br0 has priority 0x8100 and mac FFFFFFFFFFF1. The ID is 8100FFFFFFFFFFF1
br1 has priority 0x8000 and mac FFFFFFFFFFF2. The ID is 8000FFFFFFFFFFF2

The ID of br1 < br0 so br1 is the root bridge. In this case comparing the priorities would have directly given us br1 as the root bridge, but we can have the same answer by comparing the bridge IDs and being sure that we will not need to compare the MAC addresses individually.

This is why bridge IDs are used for root bridge election and why I made sure to generate one for the OVS bridge. Although I am not sure for what purpose the bridge ID is used for in LXD.
A UUID cannot be used for such a purpose and since the struct specified a bridge ID explicitly, it made me uncomfortable just using the UUID as is.

tomponline: > Although I am not sure for what purpose the bridge ID is used for in LXD.

Yes this is the root of my question really. I wanted to check what assumptions, if any, you had made as to why use the STP style ID rather than the UUID one.

Was it just to align with what we already have for native Linux bridges?

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +374,141 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
...
+
+// GenerateOVSBridgeID returns the bridge ID of the OVS bridge.
+// The bridge IDs follow the following format <STP priority>.<MAC Address>.
+// Check the Bridge ID section on https://www.kernel.org/doc/Documentation/networking/bridge.rst.
+// The STP priority is in hexadecimal format just like the MAC address.
+func (o *OVS) GenerateOVSBridgeID(ctx context.Context, bridgeName string) (string, error) {
+	// get the MAC address
+	netIf, err := net.InterfaceByName(bridgeName)
+	if err != nil {
+		return "None", err
```
tomponline, line 416: Can we return empty string here rather than special value "None", empty means none right?

lxd/network/openvswitch/ovs.go
```diff
@@ -373,3 +374,141 @@ func (o *OVS) OVNSouthboundDBRemoteAddress() (string, error) {
 
 	return addr, nil
 }
+
+// GetSTPPriority returns the STP priority of the OVS bridge.
+// Default STP priority is 32768 (0x8000). This value will be used if the priority cannot be retrieved or if STP is disabled.
+func GetSTPPriority(ctx context.Context, bridgeName string) (int, error) {
```
tomponline, line 380: should this return a uint with a particular size? What is the max priority value supported?

tomponline (commented):
Couple of small changes, then good to go.

lxd/network/openvswitch/ovs.go
ghadi-rahme: yeas you are correct I should turn this into an unsigned. The range is 0 to 61440 generally, but I will check if this is the case for OVS as well

tomponline: so uint16 should suffice and be arch independent.

ghadi-rahme: yup it should suffice, I'll just need to confirm is OVS is not doing odd things with the priority

ghadi-rahme: slight correction, in OVS the range is a bit different, it's from 0 to 65,535. Good news is that it fits exactly in the unsigned 16 bits range.

```
 other_config : stp-priority
                                     optional  string,  containing an integer,
                                     in range 0 to 65,535
```
http://www.openvswitch.org/support/dist-docs/ovs-vswitchd.conf.db.5.txt
I will add this link and range value as a comment above the function for clarity.
Also forgot to make the function local in the last push.

ghadi-rahme:
Let me know how it looks!

PR #15067, opened 2025-02-24
Assorted small fixes and improvements

simondeziel:
Addressing CodeQL warnings:

* https://github.com/canonical/lxd/security/code-scanning/148
* https://github.com/canonical/lxd/security/code-scanning/122

### DIFF
--- a/lxd/cgroup/abstraction.go
+++ b/lxd/cgroup/abstraction.go
@@ -832,31 +832,34 @@ func (cg *CGroup) GetMemoryStats() (map[string]uint64, error) {
 	}
 
 	for _, stat := range strings.Split(stats, "\n") {
-		field := strings.Split(stat, " ")
+		key, value, found := strings.Cut(stat, " ")
+		if !found {
+			continue
+		}
 
-		switch field[0] {
+		switch key {
 		case "total_active_anon", "active_anon":
-			out["active_anon"], _ = strconv.ParseUint(field[1], 10, 64)
+			out["active_anon"], _ = strconv.ParseUint(value, 10, 64)
 		case "total_active_file", "active_file":
-			out["active_file"], _ = strconv.ParseUint(field[1], 10, 64)
+			out["active_file"], _ = strconv.ParseUint(value, 10, 64)
 		case "total_inactive_anon", "inactive_anon":
-			out["inactive_anon"], _ = strconv.ParseUint(field[1], 10, 64)
+			out["inactive_anon"], _ = strconv.ParseUint(value, 10, 64)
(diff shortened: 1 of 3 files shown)


simondeziel:
@tomponline I've again bundled not immediately related changes in this PR but that's because I find it hard to split them for such "drive-by" PRs. I've kept this to a minimum but let me know if that's still too much.

lxd/cgroup/abstraction.go
```diff
@@ -242,13 +242,13 @@ func (cg *CGroup) SetMemorySwapLimit(limit int64) error {
 			return fmt.Errorf("Failed parsing %q: %w", val, err)
 		}
 
-		return cg.rw.Set(version, "memory", "memory.memsw.limit_in_bytes", fmt.Sprintf("%d", limit+valInt))
+		return cg.rw.Set(version, "memory", "memory.memsw.limit_in_bytes", fmt.Sprint(limit+valInt))
```
tomponline, line 241: Can we use strconv here?

lxd/cgroup/abstraction.go
```diff
@@ -1046,6 +1046,11 @@ func (cg *CGroup) GetIOStats() (map[string]*IOStats, error) {
 					continue
 				}
 
+				// Skip irrelevant stats related to direct IO (dbytes and dios).
```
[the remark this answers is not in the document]
simondeziel: I added another comment outside of the loop showing what the line we are splitting looks like. Let me know if that's still not clear enough.

lxd/cgroup/abstraction.go
```diff
@@ -1046,6 +1047,11 @@ func (cg *CGroup) GetIOStats() (map[string]*IOStats, error) {
 					continue
 				}
 
+				// Skip irrelevant stats related to direct IO (dbytes= and dios=).
```
tomponline, line 1045: this wasnt skipped before right? So the will the metrics values change as a result of this, and are you saying they were wrong before?

simondeziel: It's always been skipped but just a bit further down in the function when comparing the key names:

```
                switch statName {
                case "rbytes":
                    ioStats.ReadBytes = statValue
                case "wbytes":
                    ioStats.WrittenBytes = statValue
                case "rios":
                    ioStats.ReadsCompleted = statValue
                case "wios":
                    ioStats.WritesCompleted = statValue
                }
```

tomponline: ok thanks, this wasnt clear at all.

tomponline (approved):

PR #15072, opened 2025-02-25
Tweaks for backup tests

simondeziel:
This mostly aims at being faster by using smaller instances/volumes.

It also renames the `backup_different_instance_uuid` test to `backup_instance_uuid` to avoid any confusion. As explained in cb5bb787dbf, the UUIDs should be preserved through `export`/`import`.


simondeziel:
@tomponline the standalone+ceph issue seems unrelated and is being investigated by Mark.

simondeziel:
@tomponline all green now.

tomponline (approved):

PR #14973, opened 2025-02-13
auth: Add more tests for entity enrichment with entitlements

gabrielmougard:
This adds tests for the enrichment of the remaining LXD entities that are eligible. This also remove support for entitlement enrichment for instance<snapshot|backup> and volume<snapshot|backup>. These are managed by the parent instance/volume `can_mange_<snapshots|backups>`

PR #15077, opened 2025-02-26
Auth: Permission deletion trigger bug

markylaing:
Permissions for auth groups and certificates were not being deleted properly by the triggers. This caused dangling permissions to be logged when new tests were added by @gabrielmougard in #14973. More details on the trigger changes can be found in the commit messages.

I've also added a regex to `deps/panic-checker` to check for this log line and fail any future tests if it is found again. I've run the tests with this addition to the test suite before and after adding the fix to ensure it works and catches the log message.

Note that these dangling permissions are *not* a security issue because they reference entities by their ID. If they are dangling, it means they are no longer referencing anything and therefore not granting access. They also can never reference another new entity, since all of our database IDs are set to auto-increment.

### DIFF
--- a/lxd/db/cluster/entity_type_identity.go
+++ b/lxd/db/cluster/entity_type_identity.go
@@ -64,17 +64,18 @@ WHERE '' = ?
 }
 
 func (e entityTypeIdentity) onDeleteTriggerSQL() (name string, sql string) {
+	typeCertificate := entityTypeCertificate{}
 	name = "on_identity_delete"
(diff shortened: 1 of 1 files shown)


tomponline (approved):
Thanks!

Do triggers get reapplied on startup?

simondeziel (approved):
Thanks!

markylaing:
> Do triggers get reapplied on startup?

Yep. They are deleted (if existing) and re-applied on every restart.

PR #15084, opened 2025-02-28
OVN: Remove `state.State` from `openvswitch`

markylaing:
State was imported only to get some config keys for setting up the OVN command connection arguments. We should try to not import state where possible, and only import the fields we need at the lowest level. This helps not only with reducing the size of imports, but also for use of packages externally, and also for unit testing.

This is required for using the OVS package externally. See #15044 and #14606. cc @ghadi-rahme


tomponline (approved):

tomponline:
ta

PR #15085, opened 2025-02-28
Network: Fix OVN NIC dynamic IPs not being allocated when DHCP is disabled but NIC device has routes

tomponline:
It is a tested behaviour that `ipv{n}.routes` and `ipv{n}.routes.external` can be set on an OVN NIC when its network has `ipv{n}.dhcp=false`, so we need to ensure OVN is still allocating dynamic IPs for the port in this case.

Also refactors `InstanceDevicePortStart` to reduce scope of variables and, hopefully, improve clarity of IP assignment logic.

### DIFF
--- a/lxd/network/openvswitch/ovn.go
+++ b/lxd/network/openvswitch/ovn.go
@@ -116,7 +116,7 @@ type OVNDHCPv6Opts struct {
 // OVNSwitchPortOpts options that can be applied to a swich port.
 type OVNSwitchPortOpts struct {
 	MAC          net.HardwareAddr   // Optional, if nil will be set to dynamic.
-	IPs          []net.IP           // Optional, if empty IPs will be set to dynamic.
+	IPs          []net.IP           // Optional, if non-nil but empty, then dynamic IPs are requested.
 	DHCPv4OptsID OVNDHCPOptionsUUID // Optional, if empty, no DHCPv4 enabled on port.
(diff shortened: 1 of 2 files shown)


lxd/network/openvswitch/ovn.go
```diff
@@ -116,7 +116,7 @@ type OVNDHCPv6Opts struct {
 // OVNSwitchPortOpts options that can be applied to a swich port.
 type OVNSwitchPortOpts struct {
 	MAC          net.HardwareAddr   // Optional, if nil will be set to dynamic.
-	IPs          []net.IP           // Optional, if empty IPs will be set to dynamic.
```
markylaing, line 119: I wonder if it would be clearer to just have a boolean for dynamic IPs and then error if `(o.DynamicIPs && len(o.IPs) > 0) || (!o.DynamicIPs && len(o.IPs) == 0)`

markylaing: Hmm looking further in the PR this might actually complicate logic elsewhere :thinking:

markylaing (approved):

lxd/network/openvswitch/ovn.go
tomponline: yeah maybe, ill merge this to see if the lxd-ci tests are fixed, and revisit. thanks

PR #15086, opened 2025-02-28
Init: Allow PowerFlex pool creation

roosterfish:
I just found that when running `lxd init` we have to ask for the mandatory PowerFlex config keys. 

Furthermore the creation of new pools on PowerFlex should not be asked as PowerFlex always expects that the storage pool is already set up on PowerFlex by an administrator.

### DIFF
--- a/lxd/main_init_interactive.go
+++ b/lxd/main_init_interactive.go
@@ -726,9 +726,14 @@ func (c *cmdInit) askStoragePool(config *api.InitPreseed, d lxd.InstanceServer, 
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


tomponline (commented):
It feels like this should be broken up and pushed down into each storage driver in order for them to acquire the necessary info interactively that each needs.

Otherwise, as we've found, its easy to forget this section, and its getting pretty unwieldy.

roosterfish:
> It feels like this should be broken up and pushed down into each storage driver 

Right, looks we also require this for Pure? I'll move it out.

roosterfish:
Added a note to the planning sheet.

tomponline:
> Added a note to the planning sheet.

Please can you log this as a GH bug too, thanks

PR #15089, opened 2025-02-28
Client: Improve image error

roosterfish:
When running `lxc (init|launch)` a dedicated remote can be provided to set the instance's image:

`lxc launch r1:abcd r2:efgh`

However if the image does not exist, the CLI returns only `Error: not found` which is not quite clear.


roosterfish:
@tomponline to follow up on our discussion from earlier.
If you don't specify a remote for the image, it picks `local` by default.

tomponline (approved):

tomponline:
@roosterfish needs a rebase please

tomponline (approved):

PR #15109, opened 2025-03-03
AppArmor-related cherry-picks due to core20 update (stable-4.0)

mihalicyn:
- core20 is updated
- core20 update contains a fixed AppArmor user space utilities version which contains a fix for an old bug https://bugs.launchpad.net/apparmor/+bug/1597017
- this makes AppArmor to be more strict than before
- and... we have a problem with devpts mount after.

### DIFF
--- a/lxd/apparmor/instance_lxc.go
+++ b/lxd/apparmor/instance_lxc.go
@@ -480,8 +480,13 @@ profile "{{ .name }}" flags=(attach_disconnected,mediate_deleted) {
   ### Configuration: unprivileged containers
   pivot_root,
 
+  # We need to allow all these filesystems because they were allowed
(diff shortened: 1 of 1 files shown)


tomponline (approved):
thanks!

tomponline:
@mihalicyn will this cause any issues on systems with LXD 4.0 and a non-updated core20 snap (i.e is the old apparmor parser compatible with these rules)?

mihalicyn:
> @mihalicyn will this cause any issues on systems with LXD 4.0 and a non-updated core20 snap (i.e is the old apparmor parser compatible with these rules)?

No, as these rule syntax is old. Everything will be fine for any core20 versions.

PR #15111, opened 2025-03-03
lxd/instance/drivers/lxc: tiny improvements

simondeziel:
The handling of `NVIDIA_REQUIRE_CUDA` and `NVIDIA_REQUIRE_DRIVER` env variables seem to have been backward causing empty variables to be exported when the respective configuration keys were empty and not defined when actually populated.

### DIFF
--- a/lxd/instance/drivers/driver_lxc.go
+++ b/lxd/instance/drivers/driver_lxc.go
@@ -1071,14 +1071,12 @@ func (d *lxc) initLXC(config bool) (*liblxc.Container, error) {
 		nvidiaDriver := d.expandedConfig["nvidia.driver.capabilities"]
 		if nvidiaDriver == "" {
 			err = lxcSetConfigItem(cc, "lxc.environment", "NVIDIA_DRIVER_CAPABILITIES=compute,utility")
-			if err != nil {
-				return nil, err
-			}
 		} else {
 			err = lxcSetConfigItem(cc, "lxc.environment", "NVIDIA_DRIVER_CAPABILITIES="+nvidiaDriver)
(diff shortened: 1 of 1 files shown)


simondeziel:
The rework of `NVIDIA_REQUIRE_CUDA` and `NVIDIA_REQUIRE_DRIVER` was not tested which is why I'm going to ask Gabriel for his input :)

simondeziel:
The "backward logic" was introduced in 2325ba266da4ffa95084f4e38d1765047ce9b58c which is from 2018 so I'm questioning if it's not my reading of the code that's backward ;)

@mihalicyn @gabrielmougard as SMEs for NVIDIA toolkit integration, please take a look.

tomponline (approved):

tomponline:
> The "backward logic" was introduced in [2325ba2](https://github.com/canonical/lxd/commit/2325ba266da4ffa95084f4e38d1765047ce9b58c) which is from 2018 so I'm questioning if it's not my reading of the code that's backward ;)
> 
> @mihalicyn @gabrielmougard as SMEs for NVIDIA toolkit integration, please take a look.

Your fix looks good as per the docs:

> nvidia.require.cuda
> The specified version expression is used to set libnvidia-container NVIDIA_REQUIRE_CUDA.

simondeziel:
> @simondeziel SMEs?

Subject Matter Experts

tomponline:
> I'm wondering why it would have gone unnoticed for so long. Also, I would have assumed this code path was exercised by Gabriel and Aleks not too long ago.

It looks fairly niche as the default is to allow all.

PR #15112, opened 2025-03-03
lxd/device/disk: Consider `readonly` for other volume types

MggMuggins:
Prior to this change, the following disk types could have `readonly: true` in their device config and still be mounted rw in VMs:
- root disks (not likely to work very well but the disk device config semantics should be consistent)
- cloud-init config drive (unknown to me if a CD image would be writable)
- `source: ceph:...` disks
- Delegated ceph block volumes (all ceph block volumes are delegated)

I alluded in #14930 to a regression introduced by this change; that was a mistake on my part. This is good to go, tests at https://github.com/canonical/lxd-ci/pull/422.


tomponline (approved):
Ta!

PR #15114, opened 2025-03-05
lxd: expose CPU flags for a LXD node and allow stateful export/import of VM between similar x86 machines 

gabrielmougard:
JIRA ticket: https://warthogs.atlassian.net/browse/LXD-2264

This work expose a LXD node CPU flags through the resources endpoint. It also compute a x86 CPU flag baseline across the cluster member so that it is possible to migrate a stateful VM from an x86 system on a target x86 system that might share common CPU flags.


tomponline:
@gabrielmougard @mihalicyn shall we close this for now?

PR #15119, opened 2025-03-06
Prevent cluster groups from being deleted when referenced by a projects' configuration

### LINKED ISSUE #15118: project config 'restricted.cluster.groups' not checked when deleting a cluster group
**problem**

spotted this while trying to restore a cluster member.
if you could make the error bit more verbose it would help. logs point into a more clear direction on the culprit

```
ubuntu@hpc-02:~$ lxc cluster restore hpc-01
Are you sure you want to restore cluster member "hpc-01"? (yes/no) [default=no]: yes
Error: Migration API failure: Project isn't allowed to use this cluster member: "hpc-01" 
```

ubuntu-desktop-remote is a stopped instance, on project `stable` which has a `restricted.cluster.groups` that does not exist anymore (it was deleted some days back)

```
ubuntu@hpc-02:~$ lxc monitor --pretty  | grep hpc-01
...
time="2025-03-06T08:56:23+01:00" level=debug msg="Handling API request" fingerprint=9198a46812cfa3434c18bf56f7e519bc083db03008a1527c5f9c003198e0ed73 ip="REDACTED.24.228:44572" method=POST protocol=cluster url="/1.0/instances/ubuntu-desktop-remote?project=stable&target=hpc-01"
time="2025-03-06T08:56:23+01:00" level=debug msg="Database error" err="Project isn't allowed to use this cluster member: \"hpc-01\""
time="2025-03-06T08:56:23+01:00" level=info msg="ID: 6b6f9060-306d-44a5-98ee-9c3003279461, Class: task, Description: Restoring cluster member" CreatedAt="2025-03-06 08:56:23.037108935 +0100 CET" Err="Migration API failure: Project isn't allowed to use this cluster member: \"hpc-01\"" Location=hpc-01 MayCancel=false Metadata="map[evacuation_progress:Migrating \"ubuntu-desktop-remote\" in project \"stable\" from \"hpc-02\"]" Resources="map[]" Status=Failure StatusCode=Failure UpdatedAt="2025-03-06 08:56:23.371603162 +0100 CET"


ubuntu@hpc-01:~$ lxc project show stable
name: stable
description: use-cases for validation purpose
config:
...
  restricted.cluster.groups: hardened-nodes
...
used_by:
...

ubuntu@hpc-01:~$ lxc cluster group list
+---------+-----------------------+---------+
|  NAME   |      DESCRIPTION      | MEMBERS |
+---------+-----------------------+---------+
| default | Default cluster group | 5       |
+---------+-----------------------+---------+
```

**howto reproduce**

- create a cluster group with a node
- create a project and assign group as `restricted.cluster.groups`
 - delete group and now project is corrupted, since it's not validating that is not in use

more specific reproduction of our use case ,though i did not test it fully:
- create a cluster group with a node
- create a project and assign group as `restricted.cluster.groups`
- create an instance in that project.
- stop it (did not check if problem happens also with running instances, or even without an instance)
- evacuate node
- delete group and now project is corrupted and will probably get the first error 
- try restore node

markylaing:
Adds an API extension `clustering_groups_used_by` and adds a `UsedBy` field to `api.ClusterGroup`. On GET requests, the `UsedBy` URLs are projects that contain the cluster group in `restricted.cluster.groups`. DELETE requests will fail if the cluster group is used by a project.

Closes #15118

### DIFF
--- a/lxd/api_cluster.go
+++ b/lxd/api_cluster.go
@@ -16,6 +16,7 @@ import (
 	"sync"
 	"time"
 
+	dqlite "github.com/canonical/go-dqlite/v3/client"
 	"github.com/gorilla/mux"
 
 	"github.com/canonical/lxd/client"
@@ -774,11 +775,15 @@ func clusterPutJoin(d *Daemon, r *http.Request, req api.ClusterPut) response.Res
 		s.UpdateIdentityCache()
 
 		// Update local setup and possibly join the raft dqlite cluster.
-		nodes := make([]db.RaftNode, len(info.RaftNodes))
-		for i, node := range info.RaftNodes {
-			nodes[i].ID = node.ID
-			nodes[i].Address = node.Address
-			nodes[i].Role = db.RaftRole(node.Role)
+		nodes := make([]db.RaftNode, 0, len(info.RaftNodes))
+		for _, node := range info.RaftNodes {
+			nodes = append(nodes, db.RaftNode{
+				NodeInfo: dqlite.NodeInfo{
+					ID:      node.ID,
+					Address: node.Address,
+					Role:    db.RaftRole(node.Role),
+				},
+			})
 		}
 
 		err = cluster.Join(s, d.gateway, networkCert, serverCert, req.ServerName, nodes)
@@ -1378,6 +1383,7 @@ func clusterNodesPost(d *Daemon, r *http.Request) response.Response {
 		}
 
 		// Filter to online members.
+		onlineNodeAddresses = make([]any, 0, len(members))
 		for _, member := range members {
 			if member.State == db.ClusterMemberStateEvacuated || member.IsOffline(s.GlobalConfig.OfflineThreshold()) {
 				continue
@@ -2481,14 +2487,16 @@ func internalClusterPostAccept(d *Daemon, r *http.Request) response.Response {
 	}
 
 	accepted := internalClusterPostAcceptResponse{
-		RaftNodes:  make([]internalRaftNode, len(nodes)),
+		RaftNodes:  make([]internalRaftNode, 0, len(nodes)),
 		PrivateKey: s.Endpoints.NetworkPrivateKey(),
 	}
 
-	for i, node := range nodes {
-		accepted.RaftNodes[i].ID = node.ID
-		accepted.RaftNodes[i].Address = node.Address
-		accepted.RaftNodes[i].Role = int(node.Role)
+	for _, node := range nodes {
+		accepted.RaftNodes = append(accepted.RaftNodes, internalRaftNode{
+			ID:      node.ID,
+			Address: node.Address,
+			Role:    int(node.Role),
+		})
 	}
 
 	return response.SyncResponse(true, accepted)
@@ -2718,12 +2726,16 @@ func internalClusterPostAssign(d *Daemon, r *http.Request) response.Response {
 		return response.BadRequest(fmt.Errorf("No raft members provided"))
 	}
 
-	nodes := make([]db.RaftNode, len(req.RaftNodes))
-	for i, node := range req.RaftNodes {
-		nodes[i].ID = node.ID
-		nodes[i].Address = node.Address
-		nodes[i].Role = db.RaftRole(node.Role)
-		nodes[i].Name = node.Name
+	nodes := make([]db.RaftNode, 0, len(req.RaftNodes))
+	for _, node := range req.RaftNodes {
+		nodes = append(nodes, db.RaftNode{
+			NodeInfo: dqlite.NodeInfo{
+				ID:      node.ID,
+				Address: node.Address,
+				Role:    db.RaftRole(node.Role),
+			},
+			Name: node.Name,
+		})
 	}
 
 	err = cluster.Assign(s, d.gateway, nodes)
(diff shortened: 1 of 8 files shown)


lxd/db/cluster/cluster_groups.go
```diff
@@ -51,3 +57,43 @@ func (c *ClusterGroup) ToAPI() (*api.ClusterGroup, error) {
...
 }
+
+// GetClusterGroupUsedBy collates references to the cluster group with the given name.
+// This currently only returns the URLs of projects whose `restricted.cluster.groups` configuration
+// contains the cluster group.
+func GetClusterGroupUsedBy(ctx context.Context, tx *sql.Tx, groupName string) ([]string, error) {
+	q := `
+SELECT projects.name, projects_config.value FROM projects 
+JOIN projects_config ON projects.id = projects_config.project_id 
+WHERE projects_config.key = 'restricted.cluster.groups' AND projects_config.value LIKE '%` + groupName + `%'`
```
tomponline, line 74: This is a SQL injection security risk.

tomponline: Also the use of LIKE here seems potentially incorrect as it might match partial group names.

Can we use JSON collation and direct matching with query parameters instead?

tomponline (changes requested):
Looks good, except the SQL injection flaw ;)

markylaing:
> Looks good, except the SQL injection flaw ;)

Whoopsie :laughing: Thank you CodeQL :pray: Not sure how I convinced myself that this was ok...

lxd/db/cluster/cluster_groups.go
markylaing: Unfortunately not, because the config key is a comma separated list.

Rather than doing anything clever, I'm now just querying for all projects that have the `restricted.cluster.groups` configuration key set. Then parsing the output to see if it contains the group.

tomponline: makes sense because they may be spaces in the comma delimited list too

lxd/db/cluster/cluster_groups.go
```diff
@@ -51,3 +57,39 @@ func (c *ClusterGroup) ToAPI() (*api.ClusterGroup, error) {
...
+	err = query.SelectObjects(ctx, stmt, func(scan func(dest ...any) error) error {
+		var projectName string
+		var configValue string
+		err := scan(&projectName, &configValue)
+		if err != nil {
+			return err
+		}
+
+		if shared.ValueInSlice(groupName, shared.SplitNTrimSpace(configValue, ",", -1, false)) {
+			projectURLs = append(projectURLs, entity.ProjectURL(projectName).String())
```
tomponline, line 91: I'm interested why you convert the api.URL to a string here rather than return a slice of api.URLs themselves, I feel that would be clearer as to the contents of the slice rather than a string?

lxd/db/cluster/cluster_groups.go
```diff
@@ -51,3 +57,39 @@ func (c *ClusterGroup) ToAPI() (*api.ClusterGroup, error) {
...
+// GetClusterGroupUsedBy collates references to the cluster group with the given name.
+// This currently only returns the URLs of projects whose `restricted.cluster.groups` configuration
+// contains the cluster group.
+func GetClusterGroupUsedBy(ctx context.Context, tx *sql.Tx, groupName string) ([]string, error) {
+	q := `
+SELECT projects.name, projects_config.value FROM projects 
+JOIN projects_config ON projects.id = projects_config.project_id 
+WHERE projects_config.key = 'restricted.cluster.groups'`
+
+	stmt, err := tx.Prepare(q)
```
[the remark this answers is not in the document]
tomponline: SelectObjects is very similar to query.Scan, we should probably get rid of SelectObjects

lxd/api_cluster.go
```diff
@@ -3817,26 +3817,24 @@ func clusterGroupsGet(d *Daemon, r *http.Request) response.Response {
 				return err
 			}
 
-			for i := range clusterGroups {
-				nodeClusterGroups, err := dbCluster.GetNodeClusterGroups(ctx, tx.Tx(), dbCluster.NodeClusterGroupFilter{GroupID: &clusterGroups[i].ID})
+			apiClusterGroups := make([]*api.ClusterGroup, len(clusterGroups))
```
tomponline, line 3820: please can you use the form `make([]*api.ClusterGroup, 0, len(clusterGroups))` to avoids accidents in the future.

lxd/db/cluster/cluster_groups.go
markylaing: I was just following the convention of other `UsedBy` methods but happy to change.

lxd/db/cluster/cluster_groups.go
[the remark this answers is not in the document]
markylaing: Ahh I knew there was a method that didn't require preparation of the statement somewhere!

lxd/api_cluster.go
markylaing: Will fix in separate commits

markylaing:
> Static analysis failure

Fixed.

tomponline (approved):
Thanks!

PR #14870, opened 2025-01-28
doc: add live-migration info, command syntax, misc improvements

minaelee:
- Add missing command syntax code blocks and examples
- Expand on cluster evacuation section, including adding subsections and additional information about live migrations
- Note that migration.stateful must be set on instances before they can be live-migrated, and that this can only be set while the instance is stopped
- Misc grammar / readability improvements

PR #15120, opened 2025-03-06
Backport docs (stable-5.21)

roosterfish:
This backports https://github.com/canonical/lxd/pull/14870 into `stable-5.21` to get the doc reference labels `cluster-restore` and `cluster-evacuate-restore` into the stable branch to unblock the MicroCloud [`v2-edge` backports](https://github.com/canonical/microcloud/pull/681) which reference those labels using the docs from `stable-5.21`.


tomponline (approved):
@roosterfish thanks!

Was there a particular reason these commits need to be done now rather than as part of the normal backport process?

PR #15122, opened 2025-03-06
Client cert presence check in GET /1.0

edlerd:
# Done

* show if client cert was present for an unauthenticated request to GET /1.0
* this is for the UI to use, as it can't know if a cert is present otherwise
* following a suggesting from @markylaing

### DIFF
--- a/shared/api/server.go
+++ b/shared/api/server.go
@@ -196,6 +196,13 @@ type ServerUntrusted struct {
 	//
 	// API extension: oidc
 	AuthMethods []string `json:"auth_methods" yaml:"auth_methods"`
+
+	// Whether the requester sent a client certificate with the request
+	// Read only: true
+	// Example: false
+	//
+	// API extension: client_cert_presence
+	ClientCertificate bool `json:"client_certificate" yaml:"client_certificate"`
 }
 
 // Server represents a LXD server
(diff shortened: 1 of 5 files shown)


shared/api/server.go
```diff
@@ -196,6 +196,11 @@ type ServerUntrusted struct {
 	//
 	// API extension: oidc
 	AuthMethods []string `json:"auth_methods" yaml:"auth_methods"`
+
+	// Whether the client sent a client certificate with the request
+	// Read only: true
+	// Example: false
+	ClientCertificate bool `json:"client_certificate" yaml:"client_certificate"`
```
edlerd, line 203: Decided against adding an API extension as this endpoint is needed to fetch the API extensions, so we can just check if the field is present in the same response and don't need the new extension IMHO.

tomponline: I think we will need one in order for the go client to decide if the field value can be used.

edlerd: I suspect as the go client has information about the presence of a client certificate. The client wouldn't need this flag. The LXD-UI needs it, because the browser is not exposing the presence of the cert to the JavaScript layer.

tomponline: my thinking here was that if you have the latest version of the shared/api package and you parse a server response from /1.0 into that struct, you cannot know if the ClientCertificate field is False whether thats because its untrue or whether its because the server doesnt support that extension. You're correct that the client can probably figure out if it sent a client cert or not so wouldnt need to send this field, but there's also a documentation aspect that normally we document against the field definition itself which api extension introduced it, which is helpful with testing too (to detect whether it should work or not).

edlerd: I see, will add the extension to this PR then.

tomponline (changes requested):
Please split the api extension into its own commit

edlerd:
> Please split the api extension into its own commit

Split the commits accordingly.

tomponline (approved):

tomponline:
Ta!

PR #15123, opened 2025-03-06
lxd/ubuntupro: Reduce log verbosity when Pro config directory does not exist

markylaing:


lxd/ubuntupro/client.go
```diff
@@ -162,8 +162,20 @@ func (s *Client) init(ctx context.Context, ubuntuProDir string, proShim pro) {
...
+		} else {
+			logger.Error("Ubuntu Pro guest attachment disabled - failed to check existence of Ubuntu Pro configuration directory", logger.Ctx{"err": err})
+		}
+
+		return
+	}
+
 	// Set up a watcher on the ubuntu pro directory.
-	err := s.watch(ctx, ubuntuProDir)
+	err = s.watch(ctx, ubuntuProDir)
```
tomponline, line 178: Can you do the errors.Is check here instead of doing the additional stat call?

tomponline (approved):

lxd/ubuntupro/client.go
tomponline: Lets leave as it is then.

PR #15124, opened 2025-03-07
Fix Snapshot Root attachment for RBD Volumes

MggMuggins:
`DiskGetRBDFormat` and `DiskParseRBDFormat` have been the gifts that keep on giving... bugs! Instead of trying to hack yet another field into these functions, I went ahead and refactored `MountEntryItem`'s `DevPath` field to use `any` instead of `string`.

Beyond being more comprehensible and less error-prone, it also makes adding fields to delegated RBD disk devices pretty trivial.

The second piece of this PR makes attaching RBD volume snapshots work (@tomponline was correct, this never worked and somehow slipped through my local tests). LXD uses Ceph in two ways:
- The `rbd` cli utility (`man 8 rbd`)
- Via Qemu QMP ([docs](https://www.qemu.org/docs/master/interop/qemu-qmp-ref.html))

The rbd utility requires the snapshot's name to be appended to the volume name with an `@`.

QMP expects a snapshot name to be passed as a separate parameter (see [docs](https://www.qemu.org/docs/master/interop/qemu-storage-daemon-qmp-ref.html#qapidoc-708)).

LXD has been conflating these things; `CephGetRBDImageName` was trying to provide image names for both `rbd` (used in the Ceph driver) and QMP for delegated block volumes (including a confusing reference to [these docs](https://docs.ceph.com/en/squid/rbd/qemu-rbd/#usage)). `CephGetRBDImageName` now separates the snapshot component in its return value so that it can be used correctly in both contexts.

### DIFF
--- a/lxd/device/disk.go
+++ b/lxd/device/disk.go
@@ -1207,7 +1207,7 @@ func (d *disk) startVM() (*deviceConfig.RunConfig, error) {
 					projectStorageVolumeName := project.StorageVolume(d.inst.Project().Name, volumeName)
 
 					vol := d.pool.GetVolume(volumeType, contentType, projectStorageVolumeName, dbVolume.Config)
-					rbdImageName := storageDrivers.CephGetRBDImageName(vol, "", false)
+					rbdImageName, snapName := storageDrivers.CephGetRBDImageName(vol, false)
 
 					mount := deviceConfig.MountEntryItem{
 						DevSource: DevSourceRBD{
@@ -1215,6 +1215,7 @@ func (d *disk) startVM() (*deviceConfig.RunConfig, error) {
 							UserName:    userName,
 							PoolName:    poolName,
 							ImageName:   rbdImageName,
+							Snapshot:    snapName,
 						},
 						DevName: d.name,
 						Opts:    opts,
--- a/lxd/device/disk.go
+++ b/lxd/device/disk.go
@@ -979,7 +979,7 @@ func (d *disk) startContainer() (*deviceConfig.RunConfig, error) {
 		// Instruct LXD to perform the mount.
 		runConf.Mounts = append(runConf.Mounts, deviceConfig.MountEntryItem{
 			DevName:    d.name,
-			DevPath:    sourceDevPath,
+			DevSource:  DevSourcePath{Path: sourceDevPath},
 			TargetPath: relativeDestPath,
 			FSType:     "none",
 			Opts:       options,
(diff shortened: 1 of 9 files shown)


lxd/device/disk.go
```diff
@@ -1235,7 +1245,8 @@ func (d *disk) startVM() (*deviceConfig.RunConfig, error) {
 			// If the source being added is a directory or cephfs share, then we will use the lxd-agent
 			// directory sharing feature to mount the directory inside the VM, and as such we need to
 			// indicate to the VM the target path to mount to.
-			if shared.IsDir(mount.DevPath) || d.sourceIsCephFs() {
```
tomponline, line 1238: @MggMuggins not a blocker for this PR, but please can you check that the removal of the IsDir check here isn't changing the behaviour in the scenario where a non-dir path is used as disk source for VMs?

tomponline: hrm on 2nd thoughts this does look to be a problem, as i cant see where the differentiation between a dir and non-dir is taking place anymore.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -3829,12 +3834,12 @@ func (d *qemu) addDriveDirConfig(cfg *[]cfgSection, bus *qemuBus, fdFiles *[]*os
...
-	fd, err := strconv.Atoi(driveConf.DevPath)
-	if err != nil {
-		return fmt.Errorf("Invalid file descriptor %q for drive %q: %w", driveConf.DevPath, driveConf.DevName, err)
+	fdSource, ok := driveConf.DevSource.(device.DevSourceFD)
+	if !ok {
+		return fmt.Errorf("Drive config for %q was not a file descriptor", driveConf.DevName)
 	}
 
-	proxyFD := d.addFileDescriptor(fdFiles, os.NewFile(uintptr(fd), driveConf.DevName))
+	proxyFD := d.addFileDescriptor(fdFiles, os.NewFile(uintptr(fdSource.FD), driveConf.DevName))
```
tomponline, line 3843: do you need to cast this to uintptr now that fdSource.FD is a uintptr?

tomponline (changes requested):
Looking good, im  really pleased we can move away from the multitude of encoding formats for device source, and instead use type assertion as we discussed a while back, nice!

Just one main concern around directory detection.

lxd/device/disk.go
MggMuggins: Agreed; I've reintroduced the check

lxd/device/disk.go
```diff
@@ -1348,7 +1362,7 @@ func (d *disk) startVM() (*deviceConfig.RunConfig, error) {
 					}
 				}
 			} else {
```
tomponline, line 1364: should we have a check for isPath here, and then an else that returns an error about unexpected type if no matching type at that stage?

tomponline (approved):
Thanks!

MggMuggins:
Initial test failure was an [unrelated clustering issue](https://github.com/canonical/lxd/actions/runs/13723225630/job/38384035259#step:14:4045)

PR #15129, opened 2025-03-07
Storage: Improve volume metadata

roosterfish:
Spec here https://docs.google.com/document/d/1AKm3Jgt08ovNShrrP27P7AIjbdtTZjc7RZ4fMpCiHK8/edit?tab=t.0

### Overview

This PR improves the storage volume metadata to be able to account for multiple volumes.
This is especially useful in case of instances to allow recovery of attached custom storage volumes.
It adds a new format identified by a new version field which can be set by the most recent CLI when exporting backups using `--export-version=(1|2)`.

Changing the format requires modifications for backup/restore and migration as both make use of the same data structures.
Importing backups from newer LXDs on older LXDs and being able to migrate from newer to older should still work seamlessly.

In addition a new API extension `backup_metadata_version` is added which acts as the gatekeeper for being able to set the version of the backup's format.

### Backport info

Most of the PR can be backported to 5.21, except the last commits indicating that this is only intended for 6.x going forward.
As the PR changes lots of lines, this should make backporting of other features much easier as the internal changes are getting backported whilst the artifacts that LXD writes to file still remain to use the old format (1).

### PR dependencies

PRs to fix the panic in case of invalid backup file format. This panic would currently happen when importing newer backups on older versions of LXD or when modifying the backup artifact:
* https://github.com/canonical/lxd/pull/15165 + https://github.com/canonical/lxd/pull/15168
* https://github.com/canonical/lxd/pull/15163
* https://github.com/canonical/lxd/pull/15164
* https://github.com/canonical/lxd/pull/15166

Connected PR in lxd-ci for tests: https://github.com/canonical/lxd-ci/pull/443

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2642,3 +2642,17 @@ This adds a request option to set snapshot's target profile on instance copy to 
 ## `resources_device_fs_uuid`
 
 Adds the field `device_fs_uuid` including the respective UUID to each disk and partition indicating whether or not a filesystem is located on the device.
+
+## `backup_metadata_version`
+
+Adds the field `version` when exporting instances and custom storage volumes to define the backup file format.
+In case the field is omitted, the server's default version is used.
+This maintains backwards compatibility with older clients.
+
+When exporting an instance, the specific version can be provided using the `--export-version` flag:
+
+`lxc export v1 --export-version 2`
+
+The same applies when exporting a custom storage volume:
+
+`lxc storage volume export pool1 vol1 --export-version 2`
--- a/doc/howto/storage_backup_volume.md
+++ b/doc/howto/storage_backup_volume.md
@@ -139,6 +139,11 @@ You can add any of the following flags to the command:
 
   Exporting a volume in optimized mode is usually quicker than exporting the individual files.
   Snapshots are exported as differences from the main volume, which decreases their size (quota) and makes them easily accessible.
+
+`--export-version`
+: If you intend to import the backup to an older version of LXD, set the version to `1` which will use the original (old) backup metadata format.
+Backups using the old format can always be imported on newer versions of LXD.
+If the flag is not specified and the server has support for the `backup_metadata_version` API extension, version `2` is used by default.
 <!-- Include end export info -->
 
 `--volume-only`
--- a/lxc/export.go
+++ b/lxc/export.go
@@ -1,11 +1,13 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"net/url"
 	"os"
 	"path"
+	"strconv"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -23,6 +25,7 @@ type cmdExport struct {
 	flagInstanceOnly         bool
 	flagOptimizedStorage     bool
 	flagCompressionAlgorithm string
+	flagExportVersion        string
 }
 
 func (c *cmdExport) command() *cobra.Command {
@@ -41,6 +44,8 @@ func (c *cmdExport) command() *cobra.Command {
 	cmd.Flags().BoolVar(&c.flagOptimizedStorage, "optimized-storage", false,
 		i18n.G("Use storage driver optimized format (can only be restored on a similar pool)"))
 	cmd.Flags().StringVar(&c.flagCompressionAlgorithm, "compression", "", i18n.G("Compression algorithm to use (none for uncompressed)")+"``")
+	cmd.Flags().StringVar(&c.flagExportVersion, "export-version", "",
+		i18n.G("Use a different metadata format version than the latest one supported by the server")+"``")
 
 	return cmd
 }
@@ -74,6 +79,37 @@ func (c *cmdExport) run(cmd *cobra.Command, args []string) error {
 		InstanceOnly:         instanceOnly,
 		OptimizedStorage:     c.flagOptimizedStorage,
 		CompressionAlgorithm: c.flagCompressionAlgorithm,
+	}
+
+	backupVersionSupported := d.HasExtension("backup_metadata_version")
+
+	// Don't allow explicitly setting 0 as it will implicitly create a backup using version 1.
+	if c.flagExportVersion == "0" {
+		return fmt.Errorf(i18n.G("Invalid export version %q"), "0")
+	}
+
+	// In case no version is set, default to 0 so we can convert it to an uint32.
+	if c.flagExportVersion == "" {
+		c.flagExportVersion = "0"
+	}
+
+	versionUint, err := strconv.ParseUint(c.flagExportVersion, 10, 32)
+	if err != nil {
+		return fmt.Errorf(i18n.G("Invalid export version %q: %w"), c.flagExportVersion, err)
+	}
+
+	versionUint32 := uint32(versionUint)
+
+	// If the server supports setting the backup version, set the selected version.
+	// If supported but the version is not set, the server picks its default version.
+	// If unsupported but the version is set to 1, the field isn't set so its up to the server to pick the old version.
+	if backupVersionSupported {
+		if versionUint32 != 0 {
+			req.Version = versionUint32
+		}
+	} else if !backupVersionSupported && versionUint32 > api.BackupMetadataVersion1 {
+		// Any version beyond 1 isn't supported by an older server without the backup_metadata_version extension.
+		return errors.New(i18n.G("The server doesn't support setting the metadata format version"))
 	}
 
 	op, err := d.CreateInstanceBackup(name, req)
--- a/lxc/export.go
+++ b/lxc/export.go
@@ -45,7 +45,7 @@ func (c *cmdExport) command() *cobra.Command {
 		i18n.G("Use storage driver optimized format (can only be restored on a similar pool)"))
 	cmd.Flags().StringVar(&c.flagCompressionAlgorithm, "compression", "", i18n.G("Compression algorithm to use (none for uncompressed)")+"``")
 	cmd.Flags().StringVar(&c.flagExportVersion, "export-version", "",
-		i18n.G("Use a different metadata format version than the latest one supported by the server")+"``")
+		i18n.G("Use a different metadata format version than the latest one supported by the server (to support imports on older LXD versions)")+"``")
 
 	return cmd
 }
--- a/lxc/storage_volume.go
+++ b/lxc/storage_volume.go
@@ -2639,6 +2639,7 @@ type cmdStorageVolumeExport struct {
 	flagVolumeOnly           bool
 	flagOptimizedStorage     bool
 	flagCompressionAlgorithm string
+	flagExportVersion        string
 }
 
 func (c *cmdStorageVolumeExport) command() *cobra.Command {
@@ -2652,6 +2653,8 @@ func (c *cmdStorageVolumeExport) command() *cobra.Command {
 	cmd.Flags().BoolVar(&c.flagOptimizedStorage, "optimized-storage", false,
 		i18n.G("Use storage driver optimized format (can only be restored on a similar pool)"))
 	cmd.Flags().StringVar(&c.flagCompressionAlgorithm, "compression", "", i18n.G("Define a compression algorithm: for backup or none")+"``")
+	cmd.Flags().StringVar(&c.flagExportVersion, "export-version", "",
+		i18n.G("Use a different metadata format version than the latest one supported by the server")+"``")
 	cmd.Flags().StringVar(&c.storage.flagTarget, "target", "", i18n.G("Cluster member name")+"``")
 	cmd.RunE = c.run
 
@@ -2708,6 +2711,37 @@ func (c *cmdStorageVolumeExport) run(cmd *cobra.Command, args []string) error {
 		VolumeOnly:           volumeOnly,
 		OptimizedStorage:     c.flagOptimizedStorage,
 		CompressionAlgorithm: c.flagCompressionAlgorithm,
+	}
+
+	backupVersionSupported := d.HasExtension("backup_metadata_version")
+
+	// Don't allow explicitly setting 0 as it will implicitly create a backup using version 1.
+	if c.flagExportVersion == "0" {
+		return fmt.Errorf(i18n.G("Invalid export version %q"), "0")
+	}
+
+	// In case no version is set, default to 0 so we can convert it to an uint32.
+	if c.flagExportVersion == "" {
+		c.flagExportVersion = "0"
+	}
+
+	versionUint, err := strconv.ParseUint(c.flagExportVersion, 10, 32)
+	if err != nil {
+		return fmt.Errorf(i18n.G("Invalid export version %q: %w"), c.flagExportVersion, err)
+	}
+
+	versionUint32 := uint32(versionUint)
+
+	// If the server supports setting the backup version, set the selected version.
+	// If supported but the version is not set, the server picks its default version.
+	// If unsupported but the version is set to 1, the field isn't set so its up to the server to pick the old version.
+	if backupVersionSupported {
+		if versionUint32 != 0 {
+			req.Version = versionUint32
+		}
+	} else if !backupVersionSupported && versionUint32 > api.BackupMetadataVersion1 {
+		// Any version beyond 1 isn't supported by an older server without the backup_metadata_version extension.
+		return errors.New(i18n.G("The server doesn't support setting the metadata format version"))
 	}
 
 	op, err := d.CreateStoragePoolVolumeBackup(name, volName, req)
--- a/lxc/storage_volume.go
+++ b/lxc/storage_volume.go
@@ -2654,7 +2654,7 @@ func (c *cmdStorageVolumeExport) command() *cobra.Command {
 		i18n.G("Use storage driver optimized format (can only be restored on a similar pool)"))
 	cmd.Flags().StringVar(&c.flagCompressionAlgorithm, "compression", "", i18n.G("Define a compression algorithm: for backup or none")+"``")
 	cmd.Flags().StringVar(&c.flagExportVersion, "export-version", "",
-		i18n.G("Use a different metadata format version than the latest one supported by the server")+"``")
+		i18n.G("Use a different metadata format version than the latest one supported by the server (to support imports on older LXD versions)")+"``")
 	cmd.Flags().StringVar(&c.storage.flagTarget, "target", "", i18n.G("Cluster member name")+"``")
 	cmd.RunE = c.run
 
--- a/lxd/api_internal.go
+++ b/lxd/api_internal.go
@@ -713,28 +713,33 @@ func internalImportFromBackup(s *state.State, projectName string, instName strin
 		return err
 	}
 
-	if backupConf.Container == nil {
+	if backupConf.Instance == nil {
 		return fmt.Errorf("Instance definition in backup config is missing")
 	}
 
 	if allowNameOverride && instName != "" {
-		backupConf.Container.Name = instName
+		backupConf.Instance.Name = instName
 	}
 
-	if instName != backupConf.Container.Name {
-		return fmt.Errorf("Instance name requested %q doesn't match instance name in backup config %q", instName, backupConf.Container.Name)
+	if instName != backupConf.Instance.Name {
+		return fmt.Errorf("Instance name requested %q doesn't match instance name in backup config %q", instName, backupConf.Instance.Name)
 	}
 
-	if backupConf.Pool == nil {
+	if len(backupConf.Pools) == 0 {
 		// We don't know what kind of storage type the pool is.
 		return fmt.Errorf("No storage pool struct in the backup file found. The storage pool needs to be recovered manually")
+	}
+
+	rootVolPool, err := backupConf.RootVolumePool()
+	if err != nil {
+		return fmt.Errorf("Failed getting the root volume's pool: %w", err)
 	}
 
 	// Try to retrieve the storage pool the instance supposedly lives on.
 	pool, err := storagePools.LoadByName(s, instancePoolName)
 	if response.IsNotFoundError(err) {
 		// Create the storage pool db entry if it doesn't exist.
-		_, err = storagePoolDBCreate(s, instancePoolName, "", backupConf.Pool.Driver, backupConf.Pool.Config)
+		_, err = storagePoolDBCreate(s, instancePoolName, "", rootVolPool.Driver, rootVolPool.Config)
 		if err != nil {
 			return fmt.Errorf("Create storage pool database entry: %w", err)
 		}
@@ -747,12 +752,12 @@ func internalImportFromBackup(s *state.State, projectName string, instName strin
 		return fmt.Errorf("Find storage pool database entry: %w", err)
 	}
 
-	if backupConf.Pool.Name != instancePoolName {
-		return fmt.Errorf(`The storage pool %q the instance was detected on does not match the storage pool %q specified in the backup file`, instancePoolName, backupConf.Pool.Name)
+	if rootVolPool.Name != instancePoolName {
+		return fmt.Errorf(`The storage pool %q the instance was detected on does not match the storage pool %q specified in the backup file`, instancePoolName, rootVolPool.Name)
 	}
 
-	if backupConf.Pool.Driver != pool.Driver().Info().Name {
-		return fmt.Errorf(`The storage pool's %q driver %q conflicts with the driver %q recorded in the instance's backup file`, instancePoolName, pool.Driver().Info().Name, backupConf.Pool.Driver)
+	if rootVolPool.Driver != pool.Driver().Info().Name {
+		return fmt.Errorf(`The storage pool's %q driver %q conflicts with the driver %q recorded in the instance's backup file`, instancePoolName, pool.Driver().Info().Name, rootVolPool.Driver)
 	}
 
 	// Check snapshots are consistent.
@@ -764,7 +769,7 @@ func internalImportFromBackup(s *state.State, projectName string, instName strin
 	// Check if a storage volume entry for the instance already exists.
 	var dbVolume *db.StorageVolume
 	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
-		dbVolume, err = tx.GetStoragePoolVolume(ctx, pool.ID(), projectName, instanceDBVolType, backupConf.Container.Name, true)
+		dbVolume, err = tx.GetStoragePoolVolume(ctx, pool.ID(), projectName, instanceDBVolType, backupConf.Instance.Name, true)
 		if err != nil && !response.IsNotFoundError(err) {
 			return err
 		}
@@ -776,12 +781,12 @@ func internalImportFromBackup(s *state.State, projectName string, instName strin
 	}
 
 	if dbVolume != nil {
-		return fmt.Errorf(`Storage volume for instance %q already exists in the database`, backupConf.Container.Name)
+		return fmt.Errorf(`Storage volume for instance %q already exists in the database`, backupConf.Instance.Name)
 	}
 
 	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
 		// Check if an entry for the instance already exists in the db.
-		_, err := tx.GetInstanceID(ctx, projectName, backupConf.Container.Name)
+		_, err := tx.GetInstanceID(ctx, projectName, backupConf.Instance.Name)
 
 		return err
 	})
@@ -790,25 +795,30 @@ func internalImportFromBackup(s *state.State, projectName string, instName strin
 	}
 
 	if err == nil {
-		return fmt.Errorf(`Entry for instance %q already exists in the database`, backupConf.Container.Name)
+		return fmt.Errorf(`Entry for instance %q already exists in the database`, backupConf.Instance.Name)
 	}
 
-	if backupConf.Volume == nil {
+	if len(backupConf.Volumes) == 0 {
 		return fmt.Errorf(`No storage volume struct in the backup file found. The storage volume needs to be recovered manually`)
 	}
 
 	if dbVolume != nil {
-		if dbVolume.Name != backupConf.Volume.Name {
-			return fmt.Errorf(`The name %q of the storage volume is not identical to the instance's name "%s"`, dbVolume.Name, backupConf.Container.Name)
+		rootVol, err := backupConf.RootVolume()
+		if err != nil {
+			return fmt.Errorf("Failed getting the root volume: %w", err)
 		}
 
-		if dbVolume.Type != backupConf.Volume.Type {
-			return fmt.Errorf(`The type %q of the storage volume is not identical to the instance's type %q`, dbVolume.Type, backupConf.Volume.Type)
+		if dbVolume.Name != rootVol.Name {
+			return fmt.Errorf(`The name %q of the storage volume is not identical to the instance's name "%s"`, dbVolume.Name, backupConf.Instance.Name)
 		}
 
-		err := s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+		if dbVolume.Type != rootVol.Type {
+			return fmt.Errorf(`The type %q of the storage volume is not identical to the instance's type %q`, dbVolume.Type, rootVol.Type)
+		}
+
+		err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
 			// Remove the storage volume db entry for the instance since force was specified.
-			return tx.RemoveStoragePoolVolume(ctx, projectName, backupConf.Container.Name, instanceDBVolType, pool.ID())
+			return tx.RemoveStoragePoolVolume(ctx, projectName, backupConf.Instance.Name, instanceDBVolType, pool.ID())
 		})
 		if err != nil {
 			return err
@@ -818,7 +828,7 @@ func internalImportFromBackup(s *state.State, projectName string, instName strin
 	var profiles []api.Profile
 
 	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
-		profiles, err = tx.GetProfiles(ctx, projectName, backupConf.Container.Profiles)
+		profiles, err = tx.GetProfiles(ctx, projectName, backupConf.Instance.Profiles)
 
 		return err
 	})
@@ -827,31 +837,31 @@ func internalImportFromBackup(s *state.State, projectName string, instName strin
 	}
 
 	// Initialise the devices maps.
-	if backupConf.Container.Devices == nil {
-		backupConf.Container.Devices = make(map[string]map[string]string, 0)
+	if backupConf.Instance.Devices == nil {
+		backupConf.Instance.Devices = make(map[string]map[string]string, 0)
 	}
 
-	if backupConf.Container.ExpandedDevices == nil {
-		backupConf.Container.ExpandedDevices = make(map[string]map[string]string, 0)
+	if backupConf.Instance.ExpandedDevices == nil {
+		backupConf.Instance.ExpandedDevices = make(map[string]map[string]string, 0)
 	}
 
 	// Apply device overrides.
 	// Do this before calling internalImportRootDevicePopulate so that device overrides are taken into account.
-	resultingDevices, err := shared.ApplyDeviceOverrides(backupConf.Container.Devices, backupConf.Container.ExpandedDevices, deviceOverrides)
+	resultingDevices, err := shared.ApplyDeviceOverrides(backupConf.Instance.Devices, backupConf.Instance.ExpandedDevices, deviceOverrides)
 	if err != nil {
 		return err
 	}
 
-	backupConf.Container.Devices = resultingDevices
+	backupConf.Instance.Devices = resultingDevices
 
 	// Add root device if needed.
 	// And ensure root device is associated with same pool as instance has been imported to.
-	internalImportRootDevicePopulate(instancePoolName, backupConf.Container.Devices, backupConf.Container.ExpandedDevices, profiles)
+	internalImportRootDevicePopulate(instancePoolName, backupConf.Instance.Devices, backupConf.Instance.ExpandedDevices, profiles)
 
 	revert := revert.New()
 	defer revert.Fail()
 
-	if backupConf.Container == nil {
+	if backupConf.Instance == nil {
 		return fmt.Errorf("No instance config in backup config")
 	}
 
@@ -868,8 +878,8 @@ func internalImportFromBackup(s *state.State, projectName string, instName strin
 	revert.Add(cleanup)
 	defer instOp.Done(err)
 
-	instancePath := storagePools.InstancePath(instanceType, projectName, backupConf.Container.Name, false)
-	isPrivileged := backupConf.Container.Config["security.privileged"] == ""
+	instancePath := storagePools.InstancePath(instanceType, projectName, backupConf.Instance.Name, false)
+	isPrivileged := backupConf.Instance.Config["security.privileged"] == ""
 
 	err = storagePools.CreateContainerMountpoint(instanceMountPoint, instancePath, isPrivileged)
 	if err != nil {
@@ -877,11 +887,11 @@ func internalImportFromBackup(s *state.State, projectName string, instName strin
 	}
 
 	for _, snap := range existingSnapshots {
-		snapInstName := backupConf.Container.Name + shared.SnapshotDelimiter + snap.Name
+		snapInstName := backupConf.Instance.Name + shared.SnapshotDelimiter + snap.Name
 
 		snapErr := s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
 			// Check if an entry for the snapshot already exists in the db.
-			_, err := tx.GetInstanceSnapshotID(ctx, projectName, backupConf.Container.Name, snap.Name)
+			_, err := tx.GetInstanceSnapshotID(ctx, projectName, backupConf.Instance.Name, snap.Name)
 
 			return err
 		})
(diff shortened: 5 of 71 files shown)


tomponline:
Can we get b3a889348666546f13be1763af27eb6edef7b1db as a separate PR please

roosterfish:
Ready now for a first review.

doc/api-extensions.md
```diff
@@ -2630,3 +2630,8 @@ Adds the field `client_certificate` to `GET /1.0` to indicate if the current req
 
 This API extension adds a `used_by` field to the API response for a {ref}`cluster group <cluster-groups>`.
 Deletion of a cluster group is disallowed if the cluster group is referenced by project configuration (see {config:option}`project-restricted:restricted.cluster.groups`).
+
+## `backup_version`
+
+Adds the field `version` when exporting instances and custom storage volumes to define the backup file format.
+In case the field is omitted the original (old) version is used to allow staying compatible with old clients.
```
minaelee, line 2637:
```suggestion
In case the field is omitted, the original (old) version is used to maintain compatibility with older clients.
```

doc/howto/storage_backup_volume.md
```diff
@@ -139,6 +139,11 @@ You can add any of the following flags to the command:
 
   Exporting a volume in optimized mode is usually quicker than exporting the individual files.
   Snapshots are exported as differences from the main volume, which decreases their size (quota) and makes them easily accessible.
+
+`--version`
+: If you intend to import the backup on an older version of LXD, you can set the version to `1.0` which will use the original (old)
```
minaelee, line 144:
```suggestion
: If you intend to import the backup to an older version of LXD, set the version to `1.0` which will use the original (old)
```

doc/howto/storage_backup_volume.md
```diff
@@ -139,6 +139,11 @@ You can add any of the following flags to the command:
...
+backup format. Backups using the old format can always be imported on newer versions of LXD.
+By default version `2.0` is used if the flag is not specified and the server has support for the `backup_version` API extension.
```
minaelee, line 146:
```suggestion
If the flag is not specified and the server has support for the `backup_version` API extension, version `2.0` is used by default.
```

lxd/api_internal.go
```diff
@@ -663,28 +663,33 @@ func internalImportFromBackup(s *state.State, projectName string, instName strin
...
 
 	if allowNameOverride && instName != "" {
-		backupConf.Container.Name = instName
+		backupConf.Instance.Name = instName
 	}
 
-	if instName != backupConf.Container.Name {
-		return fmt.Errorf("Instance name requested %q doesn't match instance name in backup config %q", instName, backupConf.Container.Name)
+	if instName != backupConf.Instance.Name {
+		return fmt.Errorf("Instance name requested %q doesn't match instance name in backup config %q", instName, backupConf.Instance.Name)
```
minaelee, line 675:
```suggestion
		return fmt.Errorf("Requested instance name %q doesn't match instance name %q in backup config ", instName, backupConf.Instance.Name)
```

lxd/api_internal.go
```diff
@@ -740,25 +745,30 @@ func internalImportFromBackup(s *state.State, projectName string, instName strin
...
-			return fmt.Errorf(`The name %q of the storage volume is not identical to the instance's name "%s"`, dbVolume.Name, backupConf.Container.Name)
+		rootVol, err := backupConf.RootVolume()
+		if err != nil {
+			return fmt.Errorf("Failed getting the root volume: %w", err)
 		}
 
-		if dbVolume.Type != backupConf.Volume.Type {
-			return fmt.Errorf(`The type %q of the storage volume is not identical to the instance's type %q`, dbVolume.Type, backupConf.Volume.Type)
+		if dbVolume.Name != rootVol.Name {
+			return fmt.Errorf(`The name %q of the storage volume is not identical to the instance's name "%s"`, dbVolume.Name, backupConf.Instance.Name)
```
minaelee, line 762:
```suggestion
			return fmt.Errorf(`The storage volume name %q does not match the instance's name "%s"`, dbVolume.Name, backupConf.Instance.Name)
```

lxd/api_internal.go
```diff
@@ -740,25 +745,30 @@ func internalImportFromBackup(s *state.State, projectName string, instName strin
...
 		}
 
-		err := s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+		if dbVolume.Type != rootVol.Type {
+			return fmt.Errorf(`The type %q of the storage volume is not identical to the instance's type %q`, dbVolume.Type, rootVol.Type)
```
minaelee, line 766:
```suggestion
			return fmt.Errorf(`The storage volume type %q does not match the instance's type %q`, dbVolume.Type, rootVol.Type)
```

lxd/backup/backup_config_utils.go
```diff
@@ -81,6 +81,65 @@ func ConfigToInstanceDBArgs(state *state.State, c *config.Config, projectName st
...
+	// Unset the deprecated keys.
+	backupConf.Container = nil
+	backupConf.Pool = nil
+	backupConf.Volume = nil
+	backupConf.VolumeSnapshots = nil
+}
+
+// DowngradeConfigFile changes from the new to the old metadata file format.
+// It's a noop in case the config is already using the old format.
+// Downgrading looses the information about any additional custom storage volumes
```
minaelee, line 120:
```suggestion
// Downgrading loses the information about any additional custom storage volumes
```

lxc/storage_volume.go
```diff
@@ -2710,6 +2713,21 @@ func (c *cmdStorageVolumeExport) run(cmd *cobra.Command, args []string) error {
 		CompressionAlgorithm: c.flagCompressionAlgorithm,
 	}
 
+	backupVersionSupported := d.HasExtension("backup_version")
+	backupVersion := api.BackupVersion(c.flagVersion)
+
+	// If the server supports setting the backup version set the selection version
```
minaelee, line 2719:
```suggestion
	// If the server supports setting the backup version, set the selection version
```

doc/api-extensions.md
roosterfish: Thanks, fixed.

doc/howto/storage_backup_volume.md
roosterfish: Fixed.

doc/howto/storage_backup_volume.md
roosterfish: Fixed.

lxd/api_internal.go
roosterfish: Fixed.

lxd/api_internal.go
roosterfish: Fixed.

lxd/api_internal.go
roosterfish: Fixed.

lxd/backup/backup_config_utils.go
roosterfish: Fixed.

lxc/storage_volume.go
roosterfish: Fixed.

roosterfish:
The new linter errors should disappear after https://github.com/canonical/lxd/pull/15258 gets merged.

shared/version/api.go
```diff
@@ -443,6 +443,7 @@ var APIExtensions = []string{
 	"client_cert_presence",
 	"clustering_groups_used_by",
 	"container_bpf_delegation",
+	"backup_version",
```
simondeziel, line 446: IIRC, such commit is best kept as the first in a series to make Tom's backport effort easier.

roosterfish: Thanks for the hint, moved to the head now.

roosterfish: Actually I moved it back now right before the changes to the public API (structs). Because all the commits before that should be allowed to be backported whilst the rest shouldn't (starting from the API extension commits).
So it's as you suggested but separating the "to-be-backported" commits from the ones starting with the API extension.

simondeziel: Sounds good, thanks!

roosterfish: Changed it again after realizing they are best suited at the beginning of the PR as the commits before wouldn't make sense as standalone as I have planned :)

lxd/backup/config/backup_config.go
```diff
@@ -4,13 +4,32 @@ import (
 	"github.com/canonical/lxd/shared/api"
 )
 
+// VolumeConfig represents the config of a volume including its snapshots.
+type VolumeConfig struct {
```
tomponline, line 14: "Config" here stands out compared to the other struct type names in the `Config` struct below.

As its all "config", the repeat of "Config" here seems unnecessary and inconsistent, what do you think?

lxd/backup/backup_config_utils.go
```diff
@@ -81,6 +81,65 @@ func ConfigToInstanceDBArgs(state *state.State, c *config.Config, projectName st
...
+			{
+				StorageVolume: *backupConf.Volume,
+				Snapshots:     backupConf.VolumeSnapshots,
+			},
+		}
+	}
+
+	// Set the corresponding backup format version if not set.
+	if backupConf.Version == 0 {
+		backupConf.Version = api.BackupMetadataVersion2
```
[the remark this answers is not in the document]
tomponline: version 2? so we have no version 1, only 0 and 2?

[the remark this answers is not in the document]
tomponline: or is version 0 equivalent to version 1 from a UX perspective?

lxd/instance_backup.go
```diff
@@ -328,6 +329,15 @@ func instanceBackupsPost(d *Daemon, r *http.Request) response.Response {
 		req.Name = fmt.Sprintf("backup%d", backupNo)
 	}
 
+	// In case no version was selected for the backup format use the globally set format by default.
+	// This allows staying backwards compatible with older CLIs which don't yet support
+	// sending this field.
+	if req.Version == 0 {
+		req.Version = config.FileMetadataVersion
+	} else if !shared.ValueInSlice(req.Version, []api.BackupMetadataVersion{api.BackupMetadataVersion1, api.BackupMetadataVersion2}) {
```
tomponline, line 337: Can we use `req.Version >  config.FileMetadataVersion` here?

lxd/backup/config/backup_config.go
```diff
@@ -6,6 +6,9 @@ import (
 	"github.com/canonical/lxd/shared/api"
 )
 
+// FileMetadataVersion represents the current version of the format used when writing a backup's metadata to file.
+const FileMetadataVersion = api.BackupMetadataVersion1
```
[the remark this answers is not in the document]
tomponline: Is it used for migration as well, if so we should drop the `File` prefix.

lxd/storage/backend_lxd.go
```diff
@@ -7068,6 +7063,17 @@ func (b *lxdBackend) UpdateInstanceBackupFile(inst instance.Instance, snapshots
 			return err
 		}
 
+		// Downgrade the config in case the old backup format was requested.
+		// Do this as one of the last steps to allow working on the latest format beforehand.
+		if version == api.BackupMetadataVersion1 {
+			backup.DowngradeConfigFile(config)
```
tomponline, line 7069: I wonder if we should have a `backup.ConvertFormat(version uint)` function instead of upgrade/downgrade (also still not enjoying the `File` suffix here as we dont know its from a file).

Then the `ConvertFormat` can contain the logic to switch between the desired versions, rather than needing if statements before each call.

shared/api/storage_pool_volume_backup.go
```diff
@@ -56,6 +56,12 @@ type StoragePoolVolumeBackupsPost struct {
 	// What compression algorithm to use
 	// Example: gzip
 	CompressionAlgorithm string `json:"compression_algorithm" yaml:"compression_algorithm"`
+
+	// What backup format version to use
+	// Example: 1
+	//
+	// API extension: backup_metadata_version
+	Version BackupMetadataVersion
```
tomponline, line 64: I'm thinking we should stick with uint as data type rather than its own named type, as its pretty clear what a version is here.

doc/howto/storage_backup_volume.md
```diff
@@ -139,6 +139,11 @@ You can add any of the following flags to the command:
 
   Exporting a volume in optimized mode is usually quicker than exporting the individual files.
   Snapshots are exported as differences from the main volume, which decreases their size (quota) and makes them easily accessible.
+
+`--export-version`
+: If you intend to import the backup to an older version of LXD, set the version to `1` which will use the original (old)
+backup metadata format. Backups using the old format can always be imported on newer versions of LXD.
```
tomponline, line 145: do we need a newline here?

tomponline: I think we should split lines on  "Backup using..."

doc/howto/storage_backup_volume.md
```diff
@@ -139,6 +139,11 @@ You can add any of the following flags to the command:
...
+If the flag is not specified and the server has support for the `backup_metadata_version` API extension, version `2` is used by default.
```
tomponline, line 146: do you think we should add the latest supported backup version to the output of `/1.0`?

tomponline: or will any future changes be via API extensions, like `backup_metadata_version3`?

tomponline (changes requested):
Thanks @roosterfish this is looking great.

There's a few ordering issues and a few minor questions/suggestions.

Thanks

tomponline:
@roosterfish does this PR deliver the roadmap item for improvements to `lxd recover` for restoring custom volumes?

tomponline (changes requested):
Please can you include some tests for the old and new formats too. Ta

roosterfish:
> @roosterfish does this PR deliver the roadmap item for improvements to `lxd recover` for restoring custom volumes?

Not yet. After this one we have to add custom storage vols to the now available list and then we can enhance `lxd recover`.

> Please can you include some tests for the old and new formats too. Ta

This is in the lxd-ci repo, see https://github.com/canonical/lxd-ci/pull/443

lxd/storage/backend_lxd.go
```diff
@@ -774,30 +774,42 @@ func (b *lxdBackend) CreateInstanceFromBackup(srcBackup backup.Info, srcData io.
...
+	// Check if the backup config contains a root volume and populate it for later use.
+	var rootVol *backupConfig.VolumeConfig
+	if srcBackup.Config != nil {
+		rootVol, err = srcBackup.Config.RootVolume()
+		if err != nil {
+			return nil, nil, fmt.Errorf("Failed getting the root volume: %w", err)
+		}
+	}
+
+	if rootVol != nil {
```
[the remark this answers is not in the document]
roosterfish: Tbh I just tried to keep the old logic but have made some tests now but in all cases (also using --empty) there is a root vol. 
So the only case where this might not be present is if somebody exports an instance, modifies the contents and then imports it. 
But in this case we can error out as the format seems corrupted.

lxd/backup/backup_config_utils.go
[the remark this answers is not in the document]
roosterfish: > or is version 0 equivalent to version 1 from a UX perspective?

Yep.

0 is just the default for uint which means if its not set then it must be the old format. The `.Version` field in the backup file won't ever be set to 1 as version 1 doesn't yet populate this field.

lxc/export.go
```diff
@@ -41,6 +43,8 @@ func (c *cmdExport) command() *cobra.Command {
 	cmd.Flags().BoolVar(&c.flagOptimizedStorage, "optimized-storage", false,
 		i18n.G("Use storage driver optimized format (can only be restored on a similar pool)"))
 	cmd.Flags().StringVar(&c.flagCompressionAlgorithm, "compression", "", i18n.G("Compression algorithm to use (none for uncompressed)")+"``")
+	cmd.Flags().UintVar(&c.flagExportVersion, "export-version", 0,
+		i18n.G("Use a different metadata format version to be able to import the backup on older LXD versions")+"``")
```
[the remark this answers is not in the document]
tomponline: You could make the flag a string type, defaulting to empty string, and then convert to a number internally before submitting to the API, with 0 being use the latest server version. This would avoid indicating `0` is a valid version on the CLI.

lxd/backup/backup_info.go
```diff
@@ -130,5 +130,8 @@ func GetInfo(r io.ReadSeeker, sysOS *sys.OS, outputPath string) (*Info, error) {
 		return nil, fmt.Errorf("Backup is missing at %q", backupIndexPath)
 	}
 
+	// Upgrade the config file in any case to the new format.
+	UpgradeConfigFile(result.Config)
```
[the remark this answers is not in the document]
roosterfish: I have squashed it into another (earlier) commit which introduced the upgrade/downgrade and already made use of the funcs.

doc/howto/storage_backup_volume.md
roosterfish: Good catch, thanks!

doc/howto/storage_backup_volume.md
roosterfish: I like the idea of adding a list of supported versions to the `/1.0`. 
This way a user doesn't need to search for API extensions but can rather consult this list if `backup_metadata_version` extension is present.

lxd/backup/config/backup_config.go
```diff
@@ -33,3 +35,40 @@ type Config struct {
 	VolumeSnapshots []*api.StorageVolumeSnapshot `yaml:"volume_snapshots,omitempty"`
 	Bucket          *api.StorageBucket           `yaml:"bucket,omitempty"`
 }
+
+// RootVolumePool returns the pool of the root volume.
+// This is the first pool in the list of pools.
```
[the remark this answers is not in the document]
roosterfish: When upgrading from the version 1 format to version 2 the `Pool` field gets moved into `Pools` so it is the first element from the list.

lxd/backup/config/backup_config.go
```diff
@@ -33,3 +35,40 @@ type Config struct {
...
+func (c *Config) UpdateRootVolumePool(pool *api.StoragePool) {
+	if c.Pools == nil {
+		c.Pools = make([]*api.StoragePool, 0, 1)
+	}
+
+	c.Pools[0] = pool
+}
+
+// RootVolume returns the root volume.
+// This is the first volume in the list of volumes.
```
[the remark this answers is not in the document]
roosterfish: When upgrading from the version 1 format to version 2 the Volume field gets moved into Volumes so it is the first element from the list.
Can you think of a case where this might be different?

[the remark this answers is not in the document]
tomponline: I'm not too keen on using positional elements to indicate special meaning, especially so when we have a pre-existing convention that is used through the rest of LXD.

lxd/backup/config/backup_config.go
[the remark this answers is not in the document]
tomponline: It would be more proper to resolve the pool volume from the embedded profile and instance config in my view.

Is there a scenario where you see this wouldnt work?

lxd/storage/backend_lxd.go
tomponline: yes lets stick to uint32.

lxd/backup/config/backup_config.go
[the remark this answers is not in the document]
roosterfish: Makes sense. I have changed it accordingly. 

This also surfaced something else that was unintentionally working before. The same backup file format is used for both instances and custom storage vols. For the latter there always is only a single volume but having the function to return this vol for custom storage volumes be named `RootVolume()` feels odd.

Instead There is now a `RootVolume()` used when retrieving the actual root vol for instances and a `CustomVolume()` which returns the first and only volume in the list when retrieving the custom storage volume.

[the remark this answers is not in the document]
tomponline: You could call it `PrimaryVolume()` if that helps for both?

lxd/backup/config/backup_config.go
roosterfish: Agreed and fixed.

lxd/backup/config/backup_config.go
```diff
@@ -4,13 +4,32 @@ import (
...
-	Profiles        []*api.Profile               `yaml:"profiles,omitempty"`
-	Volume          *api.StorageVolume           `yaml:"volume,omitempty"`
+	Version api.BackupMetadataVersion `yaml:"version,omitempty"`
+	// Deprecated.
+	// Use Instance instead.
+	Container *api.Instance           `yaml:"container,omitempty"`
+	Instance  *api.Instance           `yaml:"instance,omitempty"`
+	Snapshots []*api.InstanceSnapshot `yaml:"snapshots,omitempty"`
+	// Deprecated.
+	// Use Pools instead.
```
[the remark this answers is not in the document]
roosterfish: Fixed.

lxd/backup/config/backup_config.go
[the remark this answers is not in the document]
roosterfish: Fixed as discussed yesterday. Also the `CustomVolume()` func now checks if the file contains any traces that makes it look like its coming from an instance and errors out in this case.

lxd/storage/backend_lxd.go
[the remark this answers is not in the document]

lxd/storage/backend_lxd.go
```diff
@@ -6832,11 +6891,14 @@ func (b *lxdBackend) GenerateCustomVolumeBackupConfig(projectName string, volNam
...
 			}
 
-			config.VolumeSnapshots = append(config.VolumeSnapshots, &snapshot)
+			volConfig.Snapshots = append(volConfig.Snapshots, &snapshot)
 		}
 	}
 
-	return config, nil
+	return &backupConfig.Config{
+		Version: api.BackupMetadataVersion2,
```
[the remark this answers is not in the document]
roosterfish: Fixed. All constants commits are now moved to the beginning for clarity.

lxc/export.go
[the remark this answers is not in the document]
roosterfish: Fixed as discussed.

lxc/export.go
```diff
@@ -76,6 +80,19 @@ func (c *cmdExport) run(cmd *cobra.Command, args []string) error {
...
+	backupVersionSupported := d.HasExtension("backup_metadata_version")
+	exportVersion := api.BackupMetadataVersion(c.flagExportVersion)
+
+	// If the server supports setting the backup version set the selected version.
+	// Not setting a version will let the API fallback to version 1.
+	if backupVersionSupported {
+		if exportVersion != 0 {
+			req.Version = exportVersion
+		}
+	} else if !backupVersionSupported && !shared.ValueInSlice(exportVersion, []api.BackupMetadataVersion{api.BackupMetadataVersion1, 0}) {
```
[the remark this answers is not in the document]
roosterfish: Fixed.

lxd/backup.go
```diff
@@ -255,6 +255,11 @@ func backupWriteIndex(sourceInst instance.Instance, pool storagePools.Pool, opti
 		return fmt.Errorf("Failed generating instance backup config: %w", err)
 	}
 
+	// Downgrade the config in case the old backup format was requested.
+	if version == api.BackupMetadataVersion1 {
```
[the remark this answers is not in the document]
roosterfish: Fixed.

lxd/backup/backup_info.go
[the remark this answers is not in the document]
roosterfish: Fixed.

lxd/instance_backup.go
roosterfish: Fixed.

lxd/backup/config/backup_config.go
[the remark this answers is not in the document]
roosterfish: Fixed.

lxd/storage/backend_lxd.go
roosterfish: Fixed, now using `uint32` everywhere.

shared/api/storage_pool_volume_backup.go
roosterfish: Fixed.

lxc/storage_volume.go
```diff
@@ -2652,6 +2653,8 @@ func (c *cmdStorageVolumeExport) command() *cobra.Command {
 	cmd.Flags().BoolVar(&c.flagOptimizedStorage, "optimized-storage", false,
 		i18n.G("Use storage driver optimized format (can only be restored on a similar pool)"))
 	cmd.Flags().StringVar(&c.flagCompressionAlgorithm, "compression", "", i18n.G("Define a compression algorithm: for backup or none")+"``")
+	cmd.Flags().UintVar(&c.flagExportVersion, "export-version", 0,
+		i18n.G("Use a different metadata format version to be able to import the backup on older LXD versions")+"``")
```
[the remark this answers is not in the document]
roosterfish: Fixed.
Also the last commit for 6.x adds a note that this flag can be used to produce backups which can be imported on older LXDs.

lxc/storage_volume.go
```diff
@@ -2710,6 +2713,19 @@ func (c *cmdStorageVolumeExport) run(cmd *cobra.Command, args []string) error {
...
+	backupVersionSupported := d.HasExtension("backup_metadata_version")
+	exportVersion := api.BackupMetadataVersion(c.flagExportVersion)
+
+	// If the server supports setting the backup version, set the selected version.
+	// Not setting a version will let the API fallback to version 1.
+	if backupVersionSupported {
+		if exportVersion != 0 {
+			req.Version = exportVersion
+		}
+	} else if !backupVersionSupported && !shared.ValueInSlice(exportVersion, []api.BackupMetadataVersion{api.BackupMetadataVersion1, 0}) {
```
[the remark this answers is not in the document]
roosterfish: Fixed.

doc/howto/storage_backup_volume.md
roosterfish: Fixed.

doc/howto/storage_backup_volume.md
roosterfish: Fixed, as discussed its now represented by a range which is also tested in the added test suite.

roosterfish:
@tomponline I have addressed all of the comments. You will find all of the requested changes in the first push (haven't rebased `main` in the middle to make reviewing easier).

There are a few more pushes after that for smaller tweaks in CI and two other ones when it comes to lookup of the instances root vol from the list of vols as it required special handling for instance snapshots. But I have added comments to the respective sections of the code.

doc/api-extensions.md
```diff
@@ -2642,3 +2642,8 @@ This adds a request option to set snapshot's target profile on instance copy to
 ## `resources_device_fs_uuid`
 
 Adds the field `device_fs_uuid` including the respective UUID to each disk and partition indicating whether or not a filesystem is located on the device.
+
+## `backup_metadata_version`
+
+Adds the field `version` when exporting instances and custom storage volumes to define the backup file format.
+In case the field is omitted, the original (old) version is used to maintain compatibility with older clients.
```
tomponline, line 2649: > In case the field is omitted, the original (old) version is used to maintain compatibility with older clients.

Is this still true, aren't we defaulting to the default preferred version that the server chooses if not specified?

Also, can you include some CLI usage examples here as this is useful for the docs. Ta

lxd/backup/config/backup_config.go
```diff
@@ -4,6 +4,10 @@ import (
 	"github.com/canonical/lxd/shared/api"
 )
 
+// MetadataVersion represents the current version of the format used when writing a backup's metadata.
+// The metadata is used both for exporting backups and for migration.
+const MetadataVersion = api.BackupMetadataVersion1
```
tomponline: Shall we call this DefaultMetadataVersion, and document it as the version used when the client doesnt specify a preferred version?

tomponline: Presumably we will also need a MaxMetadataVersion constant too?

lxd/backup/config/backup_config.go
```diff
@@ -8,13 +8,28 @@ import (
 // The metadata is used both for exporting backups and for migration.
 const MetadataVersion = api.BackupMetadataVersion1
 
+// Volume represents the config of a volume including its snapshots.
+type Volume struct {
+	// Make sure to have the embedded structs fields inline to avoid nesting.
+	api.StorageVolume `yaml:",inline"`
+	Snapshots         []*api.StorageVolumeSnapshot `yaml:"snapshots,omitempty"`
```
tomponline, line 18: Usually we have a new line between embedded types and struct fields IIRC?

lxd/backup/config/backup_config.go
```diff
@@ -33,3 +35,116 @@ type Config struct {
...
 	VolumeSnapshots []*api.StorageVolumeSnapshot `yaml:"volume_snapshots,omitempty"`
 }
+
+func (c *Config) rootVolPoolName() (string, error) {
+	if c.Instance == nil {
+		return "", fmt.Errorf("Instance config is missing")
+	}
+
+	for deviceName, deviceConfig := range c.Instance.ExpandedDevices {
+		if deviceName == "root" {
```
tomponline, line 46: The root device doesnt have to be called "root".

Can we use `instancetype.GetRootDiskDevice(entry.Devices)` here or in `RootVolumePool()` below instead of this function?

lxd/backup/config/backup_config.go
```diff
@@ -33,3 +35,116 @@ type Config struct {
...
+		if volume.Name == c.Instance.Name {
+			return volume, nil
+		}
+	}
+
+	// Second try fetching the single volume for snapshot instances.
+	// Snapshot instances don't have the Instance field populated.
+	// A snapshot is always represented by a single volume.
+	// Therefore reuse the same tooling as when retrieving a custom volume.
+	volume, _ := c.CustomVolume()
```
tomponline, line 123: hrm i dont like this because the error returned here might mention "custom volumes"

doc/api-extensions.md
roosterfish: True, I missed to update this sentence. Will also add some examples.

lxd/backup/config/backup_config.go
roosterfish: I have found both in the code but I think it makes sense for readability.

lxd/backup/backup_config_utils.go
```diff
@@ -81,6 +81,67 @@ func ConfigToInstanceDBArgs(state *state.State, c *config.Config, projectName st
 	return inst, nil
 }
 
+// ConvertFormat converts a backup config's metadata file format between versions.
+// In case the requested format is already present it's a noop.
+func ConvertFormat(backupConf *config.Config, version uint32) {
```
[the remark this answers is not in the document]
roosterfish: Properly copying the `*config.Config` is tricky as its full of pointers. Instead why not returning a cleanup hook which can be used by the caller if wanted?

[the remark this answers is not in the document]
tomponline: the data has already gone by then though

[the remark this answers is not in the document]
tomponline: we do have shared.DeepCopy btw.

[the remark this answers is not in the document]
roosterfish: I knew there was something regarding deep copy but I didn't remember anymore. Great, I'll put it in then.

lxd/backup/config/backup_config.go
roosterfish: The `MaxMetadataVersion` made me thinking. Remember this comment https://github.com/canonical/lxd/pull/15129#discussion_r2034931624. 
By using `req.Version >  config.FileMetadataVersion` instead, the API now rejects any version higher than the default which is not right for LXD 5.21 as it will support version 2 so we should have this max constant or the check that was in there previously.

But as the check is basically duplicated for instances and custom storage vols, I'll go with the const. Thanks.
I'll also split the test case so that there is something to be "backportable" to 5.21 too that would catch this error.

roosterfish: I think we can add the "min" equivalent in the future when starting to deprecate version 1.

roosterfish: The test as it is right now is actually fine and can be backported. Nvm.

lxd/backup/config/backup_config.go
roosterfish: Fixed. I now went with a `primaryVolume` which is generic enough to be called by `CustomVolume` and `RootVolume` depending on the use case.

lxd/backup/config/backup_config.go
roosterfish: Fixed.

lxd/backup/config/backup_config.go
roosterfish: Fixed.

lxd/backup/config/backup_config.go
roosterfish: Fixed.

doc/api-extensions.md
roosterfish: Fixed.

lxd/backup/config/backup_config.go
```diff
@@ -11,13 +11,29 @@ const DefaultMetadataVersion = api.BackupMetadataVersion1
...
 type Config struct {
-	Container       *api.Instance                `yaml:"container,omitempty"` // Used by VM backups too.
-	Snapshots       []*api.InstanceSnapshot      `yaml:"snapshots,omitempty"`
-	Pool            *api.StoragePool             `yaml:"pool,omitempty"`
-	Profiles        []*api.Profile               `yaml:"profiles,omitempty"`
-	Volume          *api.StorageVolume           `yaml:"volume,omitempty"`
+	Version   uint32                  `yaml:"version,omitempty"`
+	Instance  *api.Instance           `yaml:"instance,omitempty"`
+	Snapshots []*api.InstanceSnapshot `yaml:"snapshots,omitempty"`
+	Pools     []*api.StoragePool      `yaml:"pools,omitempty"`
```
tomponline, line 31: As I was looking through the code it occurred to me that whilst `Snapshots` makes sense as a slice as order is important, the `Pools` and `Volumes` fields do not have order. Would it make sense to make these maps, keyed on their respective pool or volume name? This would make loading a specific pool/volume for read/update easier.

lxd/backup/config/backup_config.go
```diff
@@ -11,13 +11,29 @@ const DefaultMetadataVersion = api.BackupMetadataVersion1
...
+	Profiles  []*api.Profile          `yaml:"profiles,omitempty"`
+	Volumes   []*Volume               `yaml:"volumes,omitempty"`
+	Bucket    *api.StorageBucket      `yaml:"bucket,omitempty"`
```
tomponline, line 34: I wonder if we should also give the same treatment to `Bucket` for consistency as it stands out a bit now.

tomponline: hrm although we do not attach buckets to instances, so probably not infact.

Still we should at least consider https://github.com/canonical/lxd/pull/13924 and ensure this new format is compatible with that.

lxd/backup/config/backup_config.go
```diff
@@ -37,3 +40,130 @@ type Config struct {
...
+			continue
+		}
+
+		if volume.Name == c.Instance.Name {
+			return volume, nil
+		}
+	}
+
+	// Second try fetching the single volume for snapshot instances.
+	// Snapshot instances don't have the Instance field populated.
```
tomponline, line 146: > 	// Snapshot instances don't have the Instance field populated.

Why is this?

lxd/backup/config/backup_config.go
tomponline: oh now, ofc, we cant do that for volumes as they may have the same name in different pools. Ignore me.

lxd/backup/config/backup_config.go
```diff
@@ -37,3 +40,130 @@ type Config struct {
...
+		}
+	}
+
+	// There already exists a root volume pool and it's name doesn't match the given pool.
+	return fmt.Errorf("Cannot apply invalid root volume pool")
+}
+
+// RootVolume returns an instance's root volume from the list of volumes.
+// The volume's name matches the one of the instance.
+func (c *Config) RootVolume() (*Volume, error) {
```
tomponline, line 132: I'm a bit confused why this function doesn't use the same technique as `rootVolPoolName` does to figure out the instance's root volume? There probably needs to be some explanation about this.

lxc/export.go
```diff
@@ -76,6 +81,37 @@ func (c *cmdExport) run(cmd *cobra.Command, args []string) error {
...
+	if c.flagExportVersion == "" {
+		c.flagExportVersion = "0"
+	}
+
+	versionUint, err := strconv.ParseUint(c.flagExportVersion, 10, 32)
+	if err != nil {
+		return fmt.Errorf(i18n.G("Invalid export version %q: %w"), c.flagExportVersion, err)
+	}
+
+	versionUint32 := uint32(versionUint)
```
tomponline, line 101: this intermediate var is probably unnecessary, as you can perform the logic comparisons on versiobnUint first and then just do:

```go
req.Version = uint32(versionUint)
```

lxc/export.go
```diff
@@ -76,6 +81,37 @@ func (c *cmdExport) run(cmd *cobra.Command, args []string) error {
...
+
+	// If the server supports setting the backup version, set the selected version.
+	// If supported but the version is not set, the server picks its default version.
+	// If unsupported but the version is set to 1, the field isn't set so its up to the server to pick the old version.
+	if backupVersionSupported {
+		if versionUint32 != 0 {
+			req.Version = versionUint32
+		}
+	} else if !backupVersionSupported && versionUint32 > api.BackupMetadataVersion1 {
```
tomponline, line 110: although does it prevent comparison with the constant?

lxd/storage/backend_lxd.go
```diff
@@ -7074,6 +7069,18 @@ func (b *lxdBackend) UpdateInstanceBackupFile(inst instance.Instance, snapshots
 			return err
 		}
 
+		// Downgrade the config in case the old backup format was requested.
```
tomponline, line 7072: can we do this just before vol.MountTask so we dont go through the overhead of mounting the volume only to fail on yaml conversion?

tomponline:
LGTM, there's some minor questions/nits, the most interesting one for me was the statements around .Instance not being populated for instance snapshots.

tomponline (approved):

lxd/backup/config/backup_config.go
roosterfish: Ultimately this was added as part of [this commit](https://github.com/canonical/lxd/commit/1fedc735327358ba07767a58373852cd492931f3#diff-029fdd0e1afdacb0e66630c0d7e813772122db9a7d036060902838db315d9c4dR4553).
I suspect the reason is to keep the amount of data stored in the backup file for snapshots low as the same data is already written in the backup file of the instance itself.

This surfaced when running the instance migration tests when trying to migrate (copy) an instance snapshot to another new instance.
In this case getting the root volume is easy as the snapshot only has a single volume which is the root volume.

tomponline: ah so this is only within the context of migration though right?

tomponline: you cant export just a snapshot?

lxd/backup/config/backup_config.go
roosterfish: Regardless of the name of the device, the volume under `volumes:` (v2) and `volume:` (v1) in the backup file is named after the name of the instance. Therefore the lookup has to be less involved.

lxd/backup/config/backup_config.go
roosterfish: Yes, currently snapshots of both instances and custom storage vols cannot be exported separately.

lxd/backup/config/backup_config.go
roosterfish: Makes sense to me. I guess what we should have is, like for volumes, a dedicated struct which embeds the bucket api type as well as snapshots.
But just don't use a list as for volumes as there really can only ever be a single bucket.
The list of volumes only makes sense as we use the same structure for both instances (with multiple vols) and custom storage vols (with a single vol).

So I would propose this:

```go
type Bucket struct {
    // Make sure to have the embedded structs fields inline to avoid nesting.
    *api.StorageBucket   `yaml:",inline"`

    Snapshots []*api.StorageBackupSnapshot `yaml:"snapshots,omitempty"` // If this is the name of the field after landing support for bucket snapshots
}

type Config struct {
    Volumes []*Volume // support instances and custom storage vols with snapshots
    Bucket *Bucket    // support a single bucket with snapshots
}

```

tomponline: yeah i think this makes sense as it allows for back up of bucket snapshots with differing config to its underlying volume config.

lxc/export.go
roosterfish: Yeah I have added the variable so that I can both use it for assignment to `req.Version` and in this check.

roosterfish: But I'll move the entire block into its own func as suggested in the other comment so we have it at a single place only.

lxc/storage_volume.go
```diff
@@ -2710,6 +2713,37 @@ func (c *cmdStorageVolumeExport) run(cmd *cobra.Command, args []string) error {
 		CompressionAlgorithm: c.flagCompressionAlgorithm,
 	}
 
+	backupVersionSupported := d.HasExtension("backup_metadata_version")
```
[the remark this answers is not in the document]
roosterfish: Fixed in https://github.com/canonical/lxd/pull/15430.

lxd/storage/backend_lxd.go
roosterfish: Fixed in https://github.com/canonical/lxd/pull/15430.

roosterfish: Thanks for the suggestion, I'll try to keep this in mind for the future when doing refactoring as it's clearly an improvement.

lxc/export.go
roosterfish: Fixed in https://github.com/canonical/lxd/pull/15430.

lxc/export.go
roosterfish: Fixed in https://github.com/canonical/lxd/pull/15430.

lxd/backup/config/backup_config.go
roosterfish: Fixed in https://github.com/canonical/lxd/pull/15430.

PR #15130, opened 2025-03-07
Resources: Prioritize native bridges over ovs ones

ghadi-rahme:
In cases when the ovs daemon is not active, retrieving native bridge information would fail. Check if the bridge is native first before doing any OVS checks in order to avoid this scenario

### DIFF
--- a/lxd/resources/network.go
+++ b/lxd/resources/network.go
@@ -477,49 +477,46 @@ func GetNetwork() (*api.ResourcesNetwork, error) {
 }
 
 // Fetch native linux bridge information.
-func getNativeBridgeState(name string) *api.NetworkStateBridge {
+func getNativeBridgeState(bridgePath string, name string) *api.NetworkStateBridge {
 	bridge := api.NetworkStateBridge{}
-	bridgePath := fmt.Sprintf("/sys/class/net/%s/bridge", name)
-	if sysfsExists(bridgePath) {
-		// Bridge ID.
-		strValue, err := os.ReadFile(filepath.Join(bridgePath, "bridge_id"))
-		if err == nil {
-			bridge.ID = strings.TrimSpace(string(strValue))
-		}
+	// Bridge ID.
+	strValue, err := os.ReadFile(filepath.Join(bridgePath, "bridge_id"))
+	if err == nil {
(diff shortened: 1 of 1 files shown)


lxd/resources/network.go
```diff
@@ -528,43 +525,53 @@ func getNativeBridgeState(name string) *api.NetworkStateBridge {
...
-	if err == nil {
-		bridge.UpperDevices = append(bridge.UpperDevices, entries...)
+		// Upper devices
+		entries, err := ovs.BridgePortList(name)
+		if err == nil {
+			bridge.UpperDevices = append(bridge.UpperDevices, entries...)
+		}
 	}
 
 	return &bridge
```
tomponline, line 577: should probably return nil if we've not been able to establish if this was a bridge in the first place, rather than asserting that this is an empty bridge incorrectly.

lxd/resources/network.go
```diff
@@ -528,43 +525,53 @@ func getNativeBridgeState(name string) *api.NetworkStateBridge {
 }
 
 // Fetch OVS bridge information.
-func getOVSBridgeState(ctx context.Context, ovs *openvswitch.OVS, name string) *api.NetworkStateBridge {
+func getOVSBridgeState(name string) *api.NetworkStateBridge {
+	ovs := openvswitch.NewOVS()
+	isOVSBridge := false
+	if ovs.Installed() {
+		isOVSBridge, _ = ovs.BridgeExists(name)
```
tomponline, line 532: seeing as we're ignored the error here, do we even need to call ovs.Installed()?

tomponline (approved):

lxd/resources/network.go
ghadi-rahme: The only reason I went with this approach was to avoid having to go through each command and letting them fail one by one.

lxd/resources/network.go
ghadi-rahme: I have that change ready if you would like me to push it?

lxd/resources/network.go
tomponline: ah yeah because Installed only looks for the file being existing. OK lets leave this as is

PR #15131, opened 2025-03-07
doc: hard-code discourse and related links titles

minaelee:
Explicitly set all titles for links set using the [`relatedlinks` Sphinx extension](https://github.com/canonical/canonical-sphinx-extensions?tab=readme-ov-file#related-links) (both Discourse links and general related links) rather than fetching at build, so that external sites' outages do not affect LXD builds.


tomponline (approved):
Thanks!

Issue #13135, opened 2024-03-13
Instance-types refresh should be optimized


simondeziel:
ATM, when LXD refreshes the instance-types definition, it makes a first HTTPS request to download `.yaml` which references the instance types of the 3 big cloud providers:

* `aws.yaml`
* `azure.yaml`
* `gcp.yaml`

Ideally, all those would be combined in a `all.yaml` and that file would be directly fetched by (up to date) LXD. For older LXD we'll need to keep the old way of fetching all 4 files due to how the YAML parsing is done.

With the above implemented, new LXDs would make a single HTTPS request.

PR #15132, opened 2025-03-08
lxd/instance_instance_types: Update `instanceRefreshTypes` function to use `all.yaml` instance type file

kadinsayani:
Closes https://github.com/canonical/lxd/issues/13135.

This pull request simplifies the process of updating instance types in the `instance_instance_types.go` file by consolidating the parsing of instance type sources into a single step.

### DIFF
--- a/lxd/instance_instance_types.go
+++ b/lxd/instance_instance_types.go
@@ -155,28 +155,12 @@ func instanceRefreshTypes(ctx context.Context, s *state.State) error {
 		_ = instanceLoadCache()
 	}
 
-	// Get the list of instance type sources
-	sources := map[string]string{}
-	err := downloadParse(".yaml", &sources)
+	// Parse the "all.yaml" file and update the global map
+	err := downloadParse("all.yaml", &instanceTypes)
 	if err != nil {
+		logger.Warnf("Failed to update instance types: %v", err)
 		return err
 	}
-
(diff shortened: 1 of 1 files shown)


simondeziel (approved):
LGTM, thanks!

FYI, I've updated the referenced issue regarding the need to preserve the old 4 `yaml` files to remain compatible with unpatched/old LXD.

Fixing modern ones is already a great win so thanks!

lxd/instance_instance_types.go
```diff
@@ -155,29 +155,13 @@ func instanceRefreshTypes(ctx context.Context, s *state.State) error {
 		_ = instanceLoadCache()
 	}
 
-	// Get the list of instance type sources
-	sources := map[string]string{}
-	err := downloadParse(".yaml", &sources)
+	// Parse the "all.yaml" file and update the global map
+	err := downloadParse("all.yaml", &instanceTypes)
 	if err != nil {
+		logger.Warnf("Failed to update instance types: %v", err)
```
tomponline, line 161: Ideally this would have been converted to contextual logging as is our convention.

tomponline (approved):

lxd/instance_instance_types.go
kadinsayani: https://github.com/canonical/lxd/pull/15153

PR #15152, opened 2025-03-10
Backports (stable-5.21)

kadinsayani:


tomponline (approved):
Thanks! The revert and then restore messed up git cherry i think.

PR #15153, opened 2025-03-10
lxd/instance_instance_types: Convert logging to contextual logging in `instanceRefreshTypes` function

kadinsayani:


lxd/instance_instance_types.go
```diff
@@ -158,14 +158,14 @@ func instanceRefreshTypes(ctx context.Context, s *state.State) error {
 	// Parse the "all.yaml" file and update the global map
 	err := downloadParse("all.yaml", &instanceTypes)
 	if err != nil {
-		logger.Warnf("Failed to update instance types: %v", err)
+		logger.Error("Failed updating instance types", logger.Ctx{"err": err})
```
tomponline, line 161: logger.Warn will be fine i think, no need for error level.

kadinsayani: Shall we change the following to `logger.Warn` as well: https://github.com/canonical/lxd/blob/main/lxd/instance_instance_types.go#L87 and https://github.com/canonical/lxd/blob/main/lxd/instance_instance_types.go#L93.

tomponline: Yeah makes sense ta

tomponline (approved):

Issue #15078, opened 2025-02-27
Cannot copy an instance that has snapshots across projects


dmitry-lyfar:
I suspect the issue is that the snapshot was created when the instance had a profile associated. The `--no-profiles` option is ignored. Reproducer:

```
lxc launch ubuntu:22.04 u1
lxc profile create test
lxc profile add u1 test
lxc snapshot u1 snap
lxc project create test-project
lxc copy u1 u1 --no-profiles -s default  --project default --target-project test-project

Error: Failed creating instance snapshot record "u1/snap": Requested profile "test" doesn't exist
```

6.2-bde4d03

skozina:
Looks like the issue is in lxc, where we first copy the instance into the new instance, and only then we overwrite the profiles with the `--no-profiles` option:
https://github.com/canonical/lxd/blob/e8de07109fc3e1f14c2cda8119e57a8f337c6955/lxc/copy.go#L284
The same applies to copies of snapshots.

tomponline:
@skozina 

The no-profiles flag causes an empty profiles slice to be specified here on the local copy of the source instance's config:

https://github.com/canonical/lxd/blob/e8de07109fc3e1f14c2cda8119e57a8f337c6955/lxc/copy.go#L298-L299


This config is then used in the actual server-side copy request here:

https://github.com/canonical/lxd/blob/e8de07109fc3e1f14c2cda8119e57a8f337c6955/lxc/copy.go#L372

What I suspect is happening is that on the server-side, the empty profiles slice is not being used to also wipe out the snapshot's config.

skozina:
Yes. On the server side, when we copy the snapshots of the source instance, we assign each snapshot the profiles of the source snapshot rather than the target instance. Here:
https://github.com/canonical/lxd/blob/d2a981a22ae4de8db7918e6a6be3ea9a78a56045/lxd/instance.go#L391-L404

As the fix we could just assign each snapshot the target instance profiles:
--- a/lxd/instance.go
+++ b/lxd/instance.go
@@ -397,7 +397,7 @@ func instanceCreateAsCopy(s *state.State, opts instanceCreateAsCopyOpts, op *ope
                                Description:  srcSnap.Description(),
                                Ephemeral:    srcSnap.IsEphemeral(),
                                Name:         newSnapName,
-                               Profiles:     srcSnap.Profiles(),
+                               Profiles:     opts.targetInstance.Profiles,
                                Project:      opts.targetInstance.Project,
[...]

PR #15155, opened 2025-03-10
lxd: Fix target profile on snapshot copy

skozina:
Fixes https://github.com/canonical/lxd/issues/15078

### DIFF
--- a/lxc/copy.go
+++ b/lxc/copy.go
@@ -229,9 +229,10 @@ func (c *cmdCopy) copyInstance(conf *config.Config, sourceResource string, destR
 		}
 
 		// Allow overriding the ephemeral status
-		if ephemeral == 1 {
+		switch ephemeral {
+		case 1:
 			entry.Ephemeral = true
-		} else if ephemeral == 0 {
+		case 0:
 			entry.Ephemeral = false
 		}
 
@@ -299,6 +300,15 @@ func (c *cmdCopy) copyInstance(conf *config.Config, sourceResource string, destR
 			entry.Profiles = []string{}
 		}
 
+		// Traditionally, if instance with snapshots is copied across projects,
+		// the snapshots keep their own profiles.
+		// This doesn't work if the profiles don't exist in target instance.
+		// If --no-profiles is applied, instruct the server to apply the profiles
+		// of the source instance, ie. none.
+		if c.flagNoProfiles {
+			args.OverrideSnapshotProfiles = true
+		}
+
 		// Check to see if any of the devices overrides are for devices that are not yet defined in the
 		// local devices and thus are expected to be coming from profiles.
 		needProfileExpansion := false
@@ -333,9 +343,10 @@ func (c *cmdCopy) copyInstance(conf *config.Config, sourceResource string, destR
 		}
 
 		// Allow overriding the ephemeral status
-		if ephemeral == 1 {
+		switch ephemeral {
+		case 1:
 			entry.Ephemeral = true
-		} else if ephemeral == 0 {
+		case 0:
 			entry.Ephemeral = false
 		}
 
--- a/lxd/instance.go
+++ b/lxd/instance.go
@@ -239,12 +239,13 @@ func instanceRebuildFromEmpty(inst instance.Instance, op *operations.Operation) 
 
 // instanceCreateAsCopyOpts options for copying an instance.
 type instanceCreateAsCopyOpts struct {
-	sourceInstance       instance.Instance // Source instance.
-	targetInstance       db.InstanceArgs   // Configuration for new instance.
-	instanceOnly         bool              // Only copy the instance and not it's snapshots.
-	refresh              bool              // Refresh an existing target instance.
-	applyTemplateTrigger bool              // Apply deferred TemplateTriggerCopy.
-	allowInconsistent    bool              // Ignore some copy errors
+	sourceInstance           instance.Instance // Source instance.
+	targetInstance           db.InstanceArgs   // Configuration for new instance.
+	instanceOnly             bool              // Only copy the instance and not it's snapshots.
+	refresh                  bool              // Refresh an existing target instance.
+	applyTemplateTrigger     bool              // Apply deferred TemplateTriggerCopy.
+	allowInconsistent        bool              // Ignore some copy errors
+	overrideSnapshotProfiles bool              // Copy the target instance profiles to the instance snapshots
 }
 
 // instanceCreateAsCopy create a new instance by copying from an existing instance.
@@ -401,6 +402,11 @@ func instanceCreateAsCopy(s *state.State, opts instanceCreateAsCopyOpts, op *ope
 				Project:      opts.targetInstance.Project,
 				ExpiryDate:   srcSnap.ExpiryDate(),
 				CreationDate: srcSnap.CreationDate(),
+			}
+
+			// Fix target profiles
+			if opts.overrideSnapshotProfiles {
+				snapInstArgs.Profiles = opts.targetInstance.Profiles
 			}
 
 			// Create the snapshots.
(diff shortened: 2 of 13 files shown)


lxd/instance.go
```diff
@@ -397,7 +397,7 @@ func instanceCreateAsCopy(s *state.State, opts instanceCreateAsCopyOpts, op *ope
 				Description:  srcSnap.Description(),
 				Ephemeral:    srcSnap.IsEphemeral(),
 				Name:         newSnapName,
-				Profiles:     srcSnap.Profiles(),
+				Profiles:     opts.targetInstance.Profiles,
```
tomponline, line 400: Thanks @skozina this indeed looks like the culprit.

I'm slightly concerned that this is going to change the default behaviour where previously the historic profiles were retained within each snapshot, but now we are going to update each copied snapshot to use the same profiles as request by the client (or default if not specified at all as per line 1158 in instances_post.go).

This is a classic example of existing historical behaviour not being well defined, and so the "correct thing" to do is not always clear. Some users may be expecting the existing behaviour, as we aim to not alter snapshots where possible.

The question here I suppose can be more concisely stated as:

> Do the `--profile` and `--no-profiles`  flags refer to just the main instance or the main instance and its snapshots?

I've run out of time now as have a meeting, but we can discuss more later on.

skozina: Thanks for quick review!
Yes, I'm not sure what's the best way to proceed. I suppose the best "fix" would be to only fix the behavior if some explicit flags are specified by the user. Ie. apply to target instance profiles only if the `--no-profiles` was specified. I'm not sure if we can detect such case as the arguments are not passed to lxd, and I'm not sure if the "empty profiles list" is a definitive indication if the `--no-profiles` was specified?
Lets talk later. Thanks!

tomponline: A similar issue exists for snapshot config:

https://github.com/canonical/lxd/pull/15155/files#diff-272a0e97fc7cdf4bb7a22457a8bf507889f75f3b6df9689074820b8c1168e4feL393
https://github.com/canonical/lxd/pull/15155/files#diff-272a0e97fc7cdf4bb7a22457a8bf507889f75f3b6df9689074820b8c1168e4feL396

`api.InstanceSource` (which is used in `api.InstancesPost` as the `Source` field when copying/migrating instances) has the `InstanceOnly` field indicating to copy only the main volume and not snapshots from the source.

I wonder if we could add a field to `api.InstanceSource` with an API extension to alter how copies/migrations treat snapshot config (profiles, local config and local devices).

tomponline: Got another meeting to attend now, but this is where im at currently...

tomponline:
converted to draft until we figure out approach.

shared/api/instance.go
```diff
@@ -72,6 +72,12 @@ type InstancesPost struct {
 	//
 	// API extension: instance_create_start
 	Start bool `json:"start" yaml:"start"`
+
+	// Whether the instances's snapshot should receive target instances profile on copy
+	// Example: true
+	//
+	// API extension: instance_copy_snapshot_target_profile
+	CopySnapshotTargetProfile bool `json:"copy_snapshot_target_profile" yaml:"copy_snapshot_target_profile"`
```
tomponline, line 80: I wonder if this would be better suited to being added to the `Source` fields (i.e `InstanceSource` struct, as that already has `InstanceOnly` as an indicator as to not copy snapshots at all).

Maybe we could call this field `OverrideSnapshotProfiles`, would this sound clearer?

tomponline:
@skozina please can you sign your commits:

https://docs.github.com/en/authentication/managing-c

shared/api/instance.go
skozina: Thanks for the tip! Yes, InstanceSource seems like a better place (I know you suggested that originally, sorry for missing that). And OverrideSnapshotProfiles is a better name. Applied both. Thanks!

tomponline (changes requested):
Thanks! We should add some tests for this scenario to avoid regressions in the future (as well as help to contribute toward our code coverage score in TICS).

In `test_container_move()` in `test/suites/container_move.sh` there is already a similar test doing:

https://github.com/canonical/lxd/blob/main/test/suites/container_move.sh#L30-L36

If you add a snapshot to that instance before it is moved to the different project with `--no-profiles` specified then that should test your change. It should fail without your PR and work correctly with your PR by ensuring the snapshot also has no profiles in its config.

skozina:
> Thanks! We should add some tests for this scenario to avoid regressions in the future (as well as help to contribute toward our code coverage score in TICS).
> 
> In `test_container_move()` in `test/suites/container_move.sh` there is already a similar test doing:
> 
> https://github.com/canonical/lxd/blob/main/test/suites/container_move.sh#L30-L36
> 
> If you add a snapshot to that instance before it is moved to the different project with `--no-profiles` specified then that should test your change. It should fail without your PR and work correctly with your PR by ensuring the snapshot also has no profiles in its config.

Thanks for the tips! This is a bit more complex.
In `container_move` test, we also need to add secondary profile, so I've done that.
However, `container_move` is only testing the `lxc move` command, which does server-side migration in this case rather than full instance copy. Turns out the server side migration is broken in the same manner as copy, as it runs `instanceCreateAsCopy()` under the hood as well. So I'm fixing the server-side `lxc move` too.

test/suites/snapshots.sh
```diff
@@ -63,6 +63,17 @@
   # Delete the snapshot
   lxc delete foo/tester_yaml
 
+  # Check if the instance with snapshots and non-default profile can be copied to a new project
+  lxc profile create source-profile
+  lxc profile add foo source-profile
+  lxc project create test-project
+  lxc copy foo foo --no-profiles -s ${pool} --project default --target-project test-project
```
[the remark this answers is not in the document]
skozina: This is now fixed.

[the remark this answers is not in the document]
skozina: This is fixed now.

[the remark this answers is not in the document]
skozina: I should also highlight that changes in `lxd/instance_post.go` are not just refactoring. Since I'm passing the request structure to `instancePostMigration()` (and similarly the `struct cmdMove` into `moveInstance()` in `lxc/move.go`), the functions changes its content. Eg. here:
https://github.com/canonical/lxd/pull/15155/files#diff-319655ca6de81bfabaaf9438c986b2597e46884fba01f9c8b5c233fcc81737cbR463

skozina:
> please can you rebase as we recently merged another api extension, thanks

Rebased, and fixed some additional lint errors which now occured.

tomponline:
Looks like there are stilk conflicts with the api extensions

skozina:
Rebased (again)

lxc/copy.go
```diff
@@ -229,9 +229,10 @@ func (c *cmdCopy) copyInstance(conf *config.Config, sourceResource string, destR
 		}
 
 		// Allow overriding the ephemeral status
-		if ephemeral == 1 {
+		switch ephemeral {
```
tomponline, line 232: for future reference unrelated linter fixes should be their own commit, as sometimes we would want to backport these but not the feature change itself. But in this case its fine as ill backport this.

lxc/copy.go
```diff
@@ -299,6 +300,15 @@ func (c *cmdCopy) copyInstance(conf *config.Config, sourceResource string, destR
 			entry.Profiles = []string{}
 		}
 
+		// Traditionally, if instance with snapshots is copied across projects,
+		// the snapshots keep their own profiles.
+		// This doesn't work if the profiles don't exist in target instance.
+		// If --no-profiles is applied, instruct the server to apply the profiles
+		// of the source instance, ie. none.
+		if c.flagNoProfiles {
```
tomponline, line 308: Should we also be setting this if `--profiles` is provided too?

tomponline (approved):
LGTM thanks!

lxc/copy.go
skozina: Thanks, will do

Issue #10302, opened 2022-04-21
Copying snapshot of running instance fails on non-thin LVM pools


tomponline:
```
lxc storage create lvm lvm lvm.use_thinpool=false
lxc init images:ubuntu/focal v1 --vm -s lvm
lxc snapshot v1
lxc copy v1/snap0 v2 # OK
lxc delete v2
lxc start v1
lxc copy v1/snap0 v2 # Fails with Error: Create instance from copy: Failed to deactivate LVM logical volume "/dev/lvm/virtual-machines_v1.block": Failed to run: lvchange --activate n --ignoreactivationskip /dev/lvm/virtual-machines_v1.block: Logical volume lvm/virtual-machines_v1.block in use.
```

tomponline:
This issue is due to the way that with non-thin LVM pools, activating a snapshot volume will also activate the origin volume (and vice versa).

It is also to do with the limitation the storage subsystem currently has around expressing the concept of a 2-stage "mount" (i.e that some volumes need both activating and mounting, and thus when stopping the operation we may only need to undo one of those actions depending on what was needed to "mount").

I think only affects the VM instance volumes, because only they have this form of mount duality.

PR #15162, opened 2025-03-12
Storage: Reference counts for LVM LV activation

MggMuggins:
When a thick LVM volume is activated, its snapshots are activated, and vice versa.

When I introduced snapshot attachment for VM volumes (https://github.com/canonical/lxd/pull/14930), it became possible for a snapshot to be mounted independently of its parent volume. The [previous logic](https://github.com/canonical/lxd/commit/44d4b432fb032fc57084afe4cad08d71979c7ae1) (see https://github.com/canonical/lxd/issues/10302) allowed for snapshots to be mounted while their parent volumes were mounted, but not vice versa.

In order to correctly identify when an LV can be safely deactivated, this introduces a reference count for LVs to prevent deactivation of an LV which was activated by another instance/operation.

### DIFF
--- a/lxd/storage/drivers/driver_lvm_volumes.go
+++ b/lxd/storage/drivers/driver_lvm_volumes.go
@@ -804,54 +804,44 @@ func (d *lvm) UnmountVolume(vol Volume, keepBlockDev bool, op *operations.Operat
 
 	refCount := vol.MountRefCountDecrement()
 
+	// For VMs, unmount the filesystem volume.
+	if vol.IsVMBlock() {
+		fsVol := vol.NewVMBlockFilesystemVolume()
+		ourUnmount, err = d.UnmountVolume(fsVol, false, op)
+		if err != nil {
+			return false, err
+		}
+	}
+
+	if refCount > 0 {
+		d.logger.Debug("Skipping unmount as in use", logger.Ctx{"volName": vol.name, "refCount": refCount})
+		return false, ErrInUse
+	}
+
 	// Check if already mounted.
 	if vol.contentType == ContentTypeFS && filesystem.IsMountPoint(mountPath) {
-		if refCount > 0 {
-			d.logger.Debug("Skipping unmount as in use", logger.Ctx{"volName": vol.name, "refCount": refCount})
-			return false, ErrInUse
-		}
-
 		err = TryUnmount(mountPath, 0)
 		if err != nil {
 			return false, fmt.Errorf("Failed to unmount LVM logical volume: %w", err)
 		}
 
-		d.logger.Debug("Unmounted logical volume", logger.Ctx{"volName": vol.name, "path": mountPath, "keepBlockDev": keepBlockDev})
-
-		// We only deactivate filesystem volumes if an unmount was needed to better align with our
-		// unmount return value indicator.
-		if !keepBlockDev {
-			_, err = d.deactivateVolume(vol)
-			if err != nil {
-				return false, err
-			}
-		}
-
 		ourUnmount = true
 	} else if vol.contentType == ContentTypeBlock {
-		// For VMs, unmount the filesystem volume.
-		if vol.IsVMBlock() {
-			fsVol := vol.NewVMBlockFilesystemVolume()
-			ourUnmount, err = d.UnmountVolume(fsVol, false, op)
-			if err != nil {
-				return false, err
-			}
-		}
-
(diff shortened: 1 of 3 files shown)


lxd/storage/drivers/driver_lvm_volumes.go
```diff
@@ -803,55 +803,44 @@ func (d *lvm) UnmountVolume(vol Volume, keepBlockDev bool, op *operations.Operat
 	mountPath := vol.MountPath()
 
 	refCount := vol.MountRefCountDecrement()
+	if refCount > 0 {
```
tomponline, line 806: I am concerned that by moving this check here we are no longer calling d.UnmountVolume for the vol.NewVMBlockFilesystemVolume() volume first, which means we've not given this outer UnmountVolume request to also decrement the associated config filesystem volume for VM.

I think we need to ensure both VM volumes have their ref counts decremented before checking the ref count is >0 for either of them.

tomponline: Otherwise, say both volumes are activated/mounted and the ref counter sits at 2 for both:

The first outer UnmountVolume() is called, and the VM block volume's ref counter is decremented to 1, but the VM filesystem volume's ref counter still sits at 2.

Then we check if the VM block volume's ref counter is >0, which it is, so we return an ErrInUse error.
This leaves the VM block's ref counter at 1 and the VM filesystem's ref counter at 2.

Later the other user of the volume finishes and calls UnmountVolume() again.
This time the VM block volume's ref counter is again decremented, now sits at 0.
Its no longer in use, so we call the VM config filesystem volume's UnmountVolume(), but this decrements the ref counter from 2 to 1, and is then considered to be in use and the unmount attempt fails with ErrInUse.

MggMuggins: Good catch; I shuffled this around a bit so both refcounts between the VM FS vol and the VM block vol should stay the same.

MggMuggins: In order to catch this bug in the tests, it would require two VMs, both running, to have the same `virtual-machine` volume attached. The tests I have only ever start [one VM](https://github.com/canonical/lxd-ci/blob/main/tests/storage-volumes-vm#L271). Adding a commit that will test this, but it comes at the cost of another VM boot.

MggMuggins: :+1:

lxd/storage/drivers/driver_lvm_volumes.go
```diff
@@ -804,54 +805,59 @@ func (d *lvm) UnmountVolume(vol Volume, keepBlockDev bool, op *operations.Operat
...
-		if refCount > 0 {
-			d.logger.Debug("Skipping unmount as in use", logger.Ctx{"volName": vol.name, "refCount": refCount})
-			return false, ErrInUse
+	// For VMs, unmount the filesystem volume.
+	if vol.IsVMBlock() {
+		fsVol := vol.NewVMBlockFilesystemVolume()
+		ourUnmount, err = d.UnmountVolume(fsVol, false, op)
+
+		// If the VMBlockFilesystem volume is still in use, we use the refCount
+		// of the block volume instead.
```
MggMuggins, line 814: Sorry to push back on this, but I'm still not clear on why this will break if the FS volume is mounted on its own. If some future code calls `MountVolume` on just the FS volume, then we expect that code to also call `UnmountVolume`, so it should keep the refCount accurate.

I guess the real question for me here is: why should trying to unmount the block volume return ErrInUse if the FS volume is still mounted elsewhere? This code (to my knowledge) just considers them as separate volumes for the purpose of refCounting.

I'm just not sure I understand the situation that breaks this :slightly_frowning_face:

tomponline: No worries, quite alright.

My worry here is that we are ignoring an InUse error from the underlying VM config filesystem volume and masking it to the caller of `UnmountVolume` of the VM block volume itself, which previously we did not.

Now, because at the current time our expectation is that the ref counters for the VM block volume and the VM config filesystem volume are in lock-step, this shouldnt cause a problem in practice as on line 820 we are then detecting the refCount > 0 of the VM block volume and returning `ErrInUse` anyway.

So I think the comment we discussed adding on line 813 should be more a statement of that assumption and why its ok to ignore the ErrInUse error there.

tomponline (approved):

PR #15168, opened 2025-03-12
API: Don't panic if backup config is missing (stable-4.0)

roosterfish:
Follow up on https://github.com/canonical/lxd/pull/15165.

The field is still accessed down the line (the code is structured differently in 4.0 therefore I have missed this bit).
The previous PR only had half of the fixes.


roosterfish:
Still the import of a backup from LXD 6.x to 4.0 is tricky due to unrecognized config settings but this prevents the daemon from running into a panic.

tomponline (approved):

PR #15170, opened 2025-03-12
tests/suites/network.sh: Add test for fetching bridge info

ghadi-rahme:
This tests makes sure that we are able to retrieve OVS bridge information as well as native linux bridge information after disabling the openvswitch service using "$ lxc network info <bridge>".


tomponline (approved):
Thanks!

PR #14885, opened 2025-01-29
Storage: Newtype for Database Volume/Content types

MggMuggins:
This:
- Communicates intent via a variable's type, not just its name
- Prevents accidental assignment of e.g. a volume type where a content type was expected
- Enforces the type safety at the database boundary; any type IDs coming out of the database layer can be assumed to be correct
- Provides convenient conversions using `Name()`

PR #15175, opened 2025-03-13
Rename `StoragePoolVolume{Content}Type` `Name` methods to `String`

MggMuggins:
As requested in the review of https://github.com/canonical/lxd/pull/14885

Yields a few readability improvements in the disk device driver as well; kudos on the suggestion.

### DIFF
--- a/lxd/db/cluster/storage_volume_type.go
+++ b/lxd/db/cluster/storage_volume_type.go
@@ -57,12 +57,12 @@ func StoragePoolVolumeTypeFromName(volTypeName string) (StoragePoolVolumeType, e
 	return StoragePoolVolumeTypeCustom, errors.New("Invalid storage volume type")
 }
 
-// Name gives the name of a StoragePoolVolumeType.
+// String gives the name of a StoragePoolVolumeType.
 //
 // # Safety
 // This function assumes that `t` is one of the StoragePoolVolumeType enums
 // defined above.
-func (t StoragePoolVolumeType) Name() string {
+func (t StoragePoolVolumeType) String() string {
 	switch t {
 	case StoragePoolVolumeTypeContainer:
 		return StoragePoolVolumeTypeNameContainer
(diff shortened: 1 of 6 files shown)


markylaing:
Hi @MggMuggins @tomponline, I'm just reading through these PRs and I'm wondering if you would instead consider the pattern used for authentication methods and identity types:
```
type AuthMethod string

const (
	authMethodTLS  int64 = 1
	authMethodOIDC int64 = 2
)

// Scan implements sql.Scanner for AuthMethod. This converts the integer value back into the correct API constant or
// returns an error.
func (a *AuthMethod) Scan(value any) error {
	if value == nil {
		return fmt.Errorf("Authentication method cannot be null")
	}

	intValue, err := driver.Int32.ConvertValue(value)
	if err != nil {
		return fmt.Errorf("Invalid authentication method type: %w", err)
	}

	authMethodInt, ok := intValue.(int64)
	if !ok {
		return fmt.Errorf("Authentication method should be an integer, got `%v` (%T)", intValue, intValue)
	}

	switch authMethodInt {
	case authMethodTLS:
		*a = api.AuthenticationMethodTLS
	case authMethodOIDC:
		*a = api.AuthenticationMethodOIDC
	default:
		return fmt.Errorf("Unknown authentication method `%d`", authMethodInt)
	}

	return nil
}

// Value implements driver.Valuer for AuthMethod. This converts the API constant into an integer or throws an error.
func (a AuthMethod) Value() (driver.Value, error) {
	switch a {
	case api.AuthenticationMethodTLS:
		return authMethodTLS, nil
	case api.AuthenticationMethodOIDC:
		return authMethodOIDC, nil
	}

	return nil, fmt.Errorf("Invalid authentication method %q", a)
}
```

This is useful because the API type can be directly casted to this type, reducing the need for conversion functions. Validation is then performed as part of reading from or writing to the database. Making it impossible to write invalid data when this type is used.

tomponline:
Good point @markylaing

Issue #15024, opened 2025-02-18
Proxy device early disconnect


jonathan-conder:
I noticed an inconsistency between proxy devices and ordinary sockets.

Normal behaviour:
```console
$ systemd-socket-activate --accept --inetd --listen 127.0.0.1:1111 -- cat 2>/dev/null &
[1] 48529
$ echo foo | nc -N localhost 1111
foo
$ kill %systemd-socket-activate
```

Proxy behaviour:
```console
$ lxc launch ubuntu:24.04 u1
Launching u1
$ lxc config device add u1 d1 proxy listen=tcp:127.0.0.1:1111 connect=tcp:127.0.0.1:1111
Device d1 added to u1
$ lxc exec u1 -- systemd-socket-activate --accept --inetd --listen 127.0.0.1:1111 -- cat &>/dev/null &
[1] 49806
$ echo foo | nc -N localhost 1111
$ kill %lxc
```

I would expect to see `foo` printed here too. As far as I can tell `nc -N` will call [`shutdown(2)`](https://man.archlinux.org/man/shutdown.2) when it's done writing to the socket. But it's not done reading yet. I'm guessing the proxy device closes both sides of the connection when that happens.

Just to check it's not `nc -N` being racy, I tried replacing `cat` above with `bash -c 'sleep 1; cat; sleep 1; echo bar'`, which prints `foo bar` on the host but nothing via the proxy.

jonathan-conder:
Another one is `seq 1 1000000 | nc -N localhost 1111`. This does echo through the proxy device (if not, just increase the number), but often cuts off early.

kadinsayani:
I've been investigating this issue, but I'm starting to question whether it's an LXD issue or simply a quirk of how `nc` behaves when `stdin` is closed. From my testing, `cat | nc -N localhost 1111` and using `socat` both work as expected with the LXD proxy device.

`nc -N` explicitly shuts down the network socket after EOF on input.

jonathan-conder:
Here's a go program that exhibits the same behaviour, in case that helps:
```go
package main

import (
	"io"
	"net"
	"os"
	"slices"
)

func serve(listener net.Listener) error {
	conn, err := listener.Accept()
	if err != nil {
		return err
	}
	defer conn.Close()

	data, err := io.ReadAll(conn)
	if err != nil {
		return err
	}
	if _, err := os.Stdout.Write(data); err != nil {
		return err
	}

	if _, err = conn.Write(data); err != nil {
		return err
[...]

PR #15178, opened 2025-03-14
forkproxy: proxy device early disconnect fix

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/15024.

### DIFF
--- a/lxd/main_forkproxy.go
+++ b/lxd/main_forkproxy.go
@@ -835,14 +835,14 @@ func genericRelay(dst net.Conn, src net.Conn) {
 	}
 
 	select {
-	case errSnd := <-chSend:
-		if daemon.Debug && errSnd != nil {
-			fmt.Printf("Warning: Error while sending data: %v\n", errSnd)
+	case errSend := <-chSend:
+		if daemon.Debug && errSend != nil {
+			fmt.Printf("Warning: Error while sending data: %v\n", errSend)
 		}
 
-	case errRcv := <-chRecv:
-		if daemon.Debug && errRcv != nil {
-			fmt.Printf("Warning: Error while reading data: %v\n", errRcv)
+	case errRecv := <-chRecv:
+		if daemon.Debug && errRecv != nil {
+			fmt.Printf("Warning: Error while reading data: %v\n", errRecv)
 		}
 	}
 
--- a/lxd/main_forkproxy.go
+++ b/lxd/main_forkproxy.go
@@ -819,41 +819,53 @@ func proxyCopy(dst net.Conn, src net.Conn) error {
 }
 
 func genericRelay(dst net.Conn, src net.Conn) {
+	// relayer is a helper function that handles one-way data copying and signals completion through the provided channel
 	relayer := func(src net.Conn, dst net.Conn, ch chan error) {
-		ch <- proxyCopy(src, dst)
+		err := proxyCopy(src, dst)
+		// Support half-closed TCP connections after copying completes. This allows data to still be received.
+		tcpConn, ok := src.(*net.TCPConn)
+		if ok {
+			_ = tcpConn.CloseWrite()
+		}
+
+		// Signal completion
+		ch <- err
 		close(ch)
 	}
 
 	chSend := make(chan error)
 	chRecv := make(chan error)
 
+	// Start copying in both directions
 	go relayer(src, dst, chRecv)
 
 	_, isUDP := dst.(*net.UDPConn)
 	if !isUDP {
-		go relayer(dst, src, chSend)
+		go relayer(dst, src, chSend) // Bidirectional copying is only supported for non-UDP connections
 	}
 
-	select {
-	case errSend := <-chSend:
-		if daemon.Debug && errSend != nil {
-			fmt.Printf("Warning: Error while sending data: %v\n", errSend)
-		}
+	// Wait for both copy operations to complete
+	for chSend != nil || chRecv != nil {
+		select {
+		case errSend := <-chSend:
+			if daemon.Debug && errSend != nil {
+				fmt.Printf("Warning: Error while sending data: %v\n", errSend)
+			}
 
-	case errRecv := <-chRecv:
-		if daemon.Debug && errRecv != nil {
-			fmt.Printf("Warning: Error while reading data: %v\n", errRecv)
+			chSend = nil // Sending complete
+
+		case errRecv := <-chRecv:
+			if daemon.Debug && errRecv != nil {
+				fmt.Printf("Warning: Error while reading data: %v\n", errRecv)
+			}
+
+			chRecv = nil // Receiving complete
 		}
 	}
 
+	// Fully close both connections once both directions are complete
 	_ = src.Close()
 	_ = dst.Close()
-
-	// Empty the channels
-	if !isUDP {
-		<-chSend
-	}
-	<-chRecv
 }
 
 func unixRelayer(src *net.UnixConn, dst *net.UnixConn, ch chan error) {
--- a/test/suites/container_devices_proxy.sh
+++ b/test/suites/container_devices_proxy.sh
@@ -117,22 +117,36 @@ container_devices_proxy_tcp() {
     false
   fi
 
-  # Initial test
+  # Initial test: Setting up multiple TCP port proxies
   lxc config device remove proxyTester proxyDev
   HOST_TCP_PORT2=$(local_tcp_port)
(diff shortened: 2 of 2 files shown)


simondeziel (approved):

nmezhenskyi (approved):

tomponline:
I'm going to wait until the tests pass before reviewing :)

```
Status: Started
Warning: Failed to connect to target: dial tcp 127.0.0.1:4321: connect: connection refused
Warning: Failed to prepare new listener instance: dial tcp 127.0.0.1:4321: connect: connection refused
Proxy device did not properly send data from host to container
```

kadinsayani:
> I'm going to wait until the tests pass before reviewing :)
> 
> ```
> Status: Started
> Warning: Failed to connect to target: dial tcp 127.0.0.1:4321: connect: connection refused
> Warning: Failed to prepare new listener instance: dial tcp 127.0.0.1:4321: connect: connection refused
> Proxy device did not properly send data from host to container
> ```

Tests passing ✅

lxd/main_forkproxy.go
```diff
@@ -827,25 +827,46 @@ func genericRelay(dst net.Conn, src net.Conn) {
...
 
-	case errRcv := <-chRecv:
-		if daemon.Debug && errRcv != nil {
-			fmt.Printf("Warning: Error while reading data: %v\n", errRcv)
+	var errSend, errRecv error
+
+	// Wait for send direction
+	if !isUDP {
+		select {
+		case errSend = <-chSend:
```
tomponline, line 846: @kadinsayani maybe im reading this wrong, but previously it looks like it waited for an error on either send or receive channels and then proceeded to close both sockets, but now it appears to wait for an error on send or for 1s to elapse, and then does the same on the receive channel or for 1s to elapse.

So what happens if the send/receive data isnt transmitted within 1s?

kadinsayani: That's correct. If the send/receive data isn't transmitted within 1s, the connection will close. I see now that this is not the behavior we want. 

I've been trying to "fix" the issue but I'm now questioning if this is an LXD issue or an issue with how `nc` behaves when `stdin` is closed, since `cat | nc -N localhost 1111` works as expected with a LXD proxy device.

kadinsayani: What we really need is a state machine with 4 possible states:

Both sides open
Send closed, read open
Send open, read closed
Both sides closed

When send is closed but read is open we should call `CloseWrite` for TCP connections, and only close both sides after the read side is also closed.

test/suites/container_devices_proxy.sh
```diff
@@ -120,11 +120,14 @@ container_devices_proxy_tcp() {
   # Initial test
   lxc config device remove proxyTester proxyDev
   HOST_TCP_PORT2=$(local_tcp_port)
+  HOST_TCP_PORT3=$(local_tcp_port)
   nsenter -n -U -t "$(lxc query /1.0/containers/proxyTester/state | jq .pid)" -- socat tcp-listen:4321 exec:/bin/cat &
   NSENTER_PID=$!
   nsenter -n -U -t "$(lxc query /1.0/containers/proxyTester/state | jq .pid)" -- socat tcp-listen:4322 exec:/bin/cat &
   NSENTER_PID1=$!
-  lxc config device add proxyTester proxyDev proxy "listen=tcp:127.0.0.1:$HOST_TCP_PORT,$HOST_TCP_PORT2" connect=tcp:127.0.0.1:4321-4322 bind=host
+  nsenter -n -U -t "$(lxc query /1.0/containers/proxyTester/state | jq .pid)" -- socat tcp-listen:4323 exec:/bin/cat &
```
simondeziel, line 128: While in there, could you please cache that PID (`lxc query /1.0/containers/proxyTester/state | jq .pid`) to avoid looking it up every time?

kadinsayani: Sure!

simondeziel: Also, if https://github.com/canonical/lxd/pull/15268 goes through, you'll want to update the `tcp-listen` to be `tcp4-listen` for forward compat with 24.04 runners.

kadinsayani: Thanks for the heads up! I've a added a commit updating all references of `socat tcp-listen` to `socat tcp4-listen`.

simondeziel: Ack, this might cause some minor rebase pain but I'm sure you've seen much worse ;)

simondeziel (approved):

lxd/main_forkproxy.go
tomponline, line 1: please can you update the commit message to expand on what has changed. Thanks

lxd/main_forkproxy.go
```diff
@@ -827,25 +827,46 @@ func genericRelay(dst net.Conn, src net.Conn) {
...
-		if daemon.Debug && errSnd != nil {
-			fmt.Printf("Warning: Error while sending data: %v\n", errSnd)
-		}
+	sendClosed := false
+	recvClosed := false
 
-	case errRcv := <-chRecv:
-		if daemon.Debug && errRcv != nil {
-			fmt.Printf("Warning: Error while reading data: %v\n", errRcv)
+	for !sendClosed || !recvClosed {
```
tomponline, line 841: A comment or two explaining whats going on here and why would be useful in my view.
From reading the code, my interpretation is that you're waiting for both sender and receiver go processes to finish before moving onto fully closing both streams, but as soon as one side finishes you're doing a CloseWrite on that side.

Would it work to move the CloseWrite logic into the `relayer()` function to reduce duplication, e.g.:

```go
	relayer := func(src net.Conn, dst net.Conn, ch chan error) {
		err :=  proxyCopy(src, dst)

			tcpConn, ok := src.(*net.TCPConn)
			if ok {
				_ = tcpConn.CloseWrite()
			}

        ch <- err
		close(ch)
	}

```

lxd/main_forkproxy.go
```diff
@@ -827,25 +827,46 @@ func genericRelay(dst net.Conn, src net.Conn) {
...
+		select {
+		case errSnd := <-chSend:
```
tomponline: I'm concerned you're introducing a busy wait loop due to the way that chSend and chRecv are closed by `relayer` once they've sent their result. Because a closed channel always returns its default value, so the select statement will always spin.

Instead you should be setting the chSend or chRecv to nil inside the select statement once it has been read from, as reading from a nil channel always blocks.

And once you've got that, you can use `for chSend != nil || chRecv != nil {` and lose the `*Done` vars.

test/suites/container_devices_proxy.sh
```diff
@@ -120,11 +120,15 @@ container_devices_proxy_tcp() {
   # Initial test
   lxc config device remove proxyTester proxyDev
   HOST_TCP_PORT2=$(local_tcp_port)
-  nsenter -n -U -t "$(lxc query /1.0/containers/proxyTester/state | jq .pid)" -- socat tcp4-listen:4321 exec:/bin/cat &
+  HOST_TCP_PORT3=$(local_tcp_port)
```
tomponline, line 123: please can you add comments to this function explaining what it is testing as its non-obvious.

tomponline (changes requested):
Looks like some changes needed to the for loop.

test/suites/container_devices_proxy.sh
simondeziel: I'd go even further and suggest adding a bunch of `echo` instead of comments, these make following along in debug logs much easier and also make it easy to `grep` for a given string ;)

lxd/main_forkproxy.go
kadinsayani: Done.

test/suites/container_devices_proxy.sh
kadinsayani: I've added comments and `echo` calls.

lxd/main_forkproxy.go
kadinsayani: I've added comments and moved the `CloseWrite` call into the `relayer` function.

lxd/main_forkproxy.go
kadinsayani: Thanks for pointing that out. I've updated the loop to follow your suggested approach.

kadinsayani:
> Looks like some changes needed to the for loop.

Thanks for your review. I've addressed all comments.

tomponline (approved):
Lgtm thanks!

PR #15194, opened 2025-03-17
gomod: Update deps

tomponline:
Without using the toolchain directive.


tomponline:
@simondeziel i tried also doing `go get toolchain@1.23.3` to match GOMIN and it worked fine, except because it was the same as the module's go version it didnt add a `toolchain` line. So is equivalent to `go get toolchain@none`.

So that dependabot is now trying to add a toolchain that is higher than we declare is a bug in dependabot IMHO.

MusicDin (approved):

simondeziel:
> So that dependabot is now trying to add a toolchain that is higher than we declare is a bug in dependabot IMHO.

I think that's what the toolchain directive is all about. Is your main grip with the toolchain directive associated to the churn in `go.mod` or are you more concerned by bugs going unnnoticed or introduced by using a toolchain (which is by definition more recent than the GOMIN)?

tomponline:
> > So that dependabot is now trying to add a toolchain that is higher than we declare is a bug in dependabot IMHO.
> 
> I think that's what the toolchain directive is all about. Is your main grip with the toolchain directive associated to the churn in `go.mod` or are you more concerned by bugs going unnnoticed or introduced by using a toolchain (which is by definition more recent than the GOMIN)?

My point is that our toolchain only needs to be set at GOMIN, and we dont need dependabot bumping it unnecessarily.

tomponline:
There's no reason to be bumping it and having 2 different versions to consider. More unwanted complexity.

tomponline:
@simondeziel if need be we can disable dependabot until its fixed

PR #15197, opened 2025-03-17
Require GHA pinning

simondeziel:
This PR is supposed to fail until I bring in the commit to pin the `actions/setup-node@v4`.

**Update:** this failed as needed until the last commit  was added.


tomponline (approved):
Ta

Issue #15179, opened 2025-03-15
LXD storage leaves leftover datasets when it is not supposed to.


dmitry-lyfar:
```
$ lxc project create p
Project p created
$ lxc project switch p
$ lxc storage create s zfs
Storage pool s created
$ lxc storage volume create s v
Storage volume v created
$ lxc project delete p
Project p deleted
$ lxc storage volume list s
+------+------+-------------+--------------+---------+
| TYPE | NAME | DESCRIPTION | CONTENT-TYPE | USED BY |
+------+------+-------------+--------------+---------+
lxc storage volume delete s
$ lxc storage delete s
Error: ZFS pool has leftover datasets: custom/p_v
```

So, the volume is not in the table anymore, but the storage cannot be deleted.

An attempt to recreate the project and volume leads to:

```
Error: Failed to run: zfs create -o mountpoint=legacy -o canmount=noauto s/custom/p_v: exit status 1 (cannot create 's/custom/p_v': dataset already exists)
```

6.2-bde4d03

tomponline:
@nmezhenskyi would you mind taking a look at this one? Looks like the "in use" checks arent working fully.

tomponline:
Regressed in https://github.com/canonical/lxd/commit/6096458d3a75623ea4d2e604a6fb0f5f4ba10424

PR #15198, opened 2025-03-17
lxd: Prevent deleting non-empty project

nmezhenskyi:
Fixes https://github.com/canonical/lxd/issues/15179.  

Fixed early return in `projectIsEmpty()` by simplifying if statement to check associated profiles.

### DIFF
--- a/lxd/api_project.go
+++ b/lxd/api_project.go
@@ -1041,12 +1041,8 @@ func projectIsEmpty(ctx context.Context, project *cluster.Project, tx *db.Cluste
 		return false, err
 	}
 
-	if len(profiles) > 0 {
-		// Consider the project empty if it is only used by the default profile.
-		if len(profiles) == 1 && profiles[0].Name == "default" {
-			return true, nil
-		}
-
+	// Consider the project empty if it is only used by the default profile.
+	if len(profiles) > 1 || (len(profiles) == 1 && profiles[0].Name != "default") {
 		return false, nil
 	}
 
--- a/test/suites/clustering.sh
+++ b/test/suites/clustering.sh
@@ -661,7 +661,7 @@ test_clustering_storage() {
     LXD_DIR="${LXD_ONE_DIR}" lxc storage volume copy pool1/vol1 pool1/vol1 --target=node1 --destination-target=node2 --refresh
     LXD_DIR="${LXD_ONE_DIR}" lxc project create foo
     LXD_DIR="${LXD_ONE_DIR}" lxc storage volume copy pool1/vol1 pool1/vol1 --target=node1 --destination-target=node2 --target-project foo
-    LXD_DIR="${LXD_ONE_DIR}" lxc project delete foo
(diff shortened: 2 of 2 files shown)


lxd/api_project.go
```diff
@@ -1041,12 +1041,8 @@ func projectIsEmpty(ctx context.Context, project *cluster.Project, tx *db.Cluste
...
 	}
 
-	if len(profiles) > 0 {
-		// Consider the project empty if it is only used by the default profile.
-		if len(profiles) == 1 && profiles[0].Name == "default" {
-			return true, nil
-		}
-
+	// Consider the project empty if it is only used by the default profile.
+	if len(profiles) > 0 && !(len(profiles) == 1 && profiles[0].Name == "default") {
```
kadinsayani, line 1045:
```suggestion
	if len(profiles) > 1 || (len(profiles) == 1 && profiles[0].Name != "default") {
```

kadinsayani (changes requested):

lxd/api_project.go
kadinsayani: Just a small nit for readability :). LGTM otherwise.

kadinsayani (approved):
Thanks!

nmezhenskyi:
Hm, it seems that `test/suites/clustering` is failing [here](https://github.com/canonical/lxd/blob/1a14cfe1c432fd5e8ab88f053c36c706aafc30d4/test/suites/clustering.sh#L664) during the System (cluster, dir) test.  

@tomponline maybe I should change the test to delete copied volumes before deleting the `foo` project? It shouldn't be possible to delete a non-empty project, right?

tomponline:
@nmezhenskyi what was the commit that introduce the issue btw?

tomponline:
> Hm, it seems that `test/suites/clustering` is failing [here](https://github.com/canonical/lxd/blob/1a14cfe1c432fd5e8ab88f053c36c706aafc30d4/test/suites/clustering.sh#L664) during the System (cluster, dir) test.
> 
> @tomponline maybe I should change the test to delete copied volumes before deleting the `foo` project? It shouldn't be possible to delete a non-empty project, right?

Yes please as the test was added by @kadinsayani and incorrectly expects the project to be deletable even with custom volumes in it.

test/suites/clustering.sh
```diff
@@ -661,7 +661,6 @@ test_clustering_storage() {
     LXD_DIR="${LXD_ONE_DIR}" lxc storage volume copy pool1/vol1 pool1/vol1 --target=node1 --destination-target=node2 --refresh
     LXD_DIR="${LXD_ONE_DIR}" lxc project create foo
     LXD_DIR="${LXD_ONE_DIR}" lxc storage volume copy pool1/vol1 pool1/vol1 --target=node1 --destination-target=node2 --target-project foo
-    LXD_DIR="${LXD_ONE_DIR}" lxc project delete foo
```
tomponline, line 664: can we keep this here but change it to a negative to check that the empty check is working for custom volumes please

tomponline (approved):
Thanks!

Issue #15202, opened 2025-03-18
LXD fails to start when Loki is unavailable


vosdev:
I just updated to 6.3 and I noticed that LXD wouldn't start right away.

It took 10 restarts before it would finally come and stay up.

journalctl shows 10x the exact same log entries as you can see in the paste below.

Loki runs inside an LXC guest on my lxd host, set up using https://github.com/canonical/lxd-ui/blob/ac979083e43f44b8bdaa4ba67cdcee250ccb9a3f/public/assets/scripts/setup-grafana.sh

After the 10th attempt it suddenly did work.

It also fails to advertise vsock address to an instance agent.

```
Mar 18 12:35:44 dragonstone lxd.daemon[3973790]: - cpuview_daemon
Mar 18 12:35:44 dragonstone lxd.daemon[3973790]: - loadavg_daemon
Mar 18 12:35:44 dragonstone lxd.daemon[3973790]: - pidfds
Mar 18 12:35:45 dragonstone lxd.daemon[3973653]: => Starting LXD
Mar 18 12:35:45 dragonstone lxd.daemon[3973844]: time="2025-03-18T12:35:45Z" level=warning msg=" - Couldn't find the CGroup network priority controller, per-instance network priority will be ignored. Please use per-device limits.priority instead"
Mar 18 12:35:47 dragonstone lxd.daemon[3973844]: time="2025-03-18T12:35:47Z" level=error msg="Failed to start the daemon" err="Loki is not ready, server returned HTTP status 503 Service Unavailable (503): Pattern Ingester not ready: waiting for 15s after being ready"
Mar 18 12:35:47 dragonstone lxd.daemon[3973844]: time="2025-03-18T12:35:47Z" level=warning msg="Failed to advertise vsock address to instance agent" err="Failed sending VM sock address to lxd-agent: Failed to fetch https://custom.socket/1.0: 500 Internal Server Error" instance=HomeassistantOS instanceType=virtual-machine project=default
Mar 18 12:35:47 dragonstone lxd.daemon[3973844]: time="2025-03-18T12:35:47Z" level=warning msg="Dqlite last entry" index=0 term=0
Mar 18 12:35:47 dragonstone lxd.daemon[3973844]: Error: Loki is not ready, server returned HTTP status 503 Service Unavailable (503): Pattern Ingester not ready: waiting for 15s after being ready
Mar 18 12:35:47 dragonstone lxd.daemon[3973653]: Killed
[...]

PR #15203, opened 2025-03-18
lxd/daemon: Log warning when Loki client is unreachable rather than return `err`

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/15202 introduced by https://github.com/canonical/lxd/pull/14868/commits/cd8966843783eee339329f13e06f68cd2cfbb902. If a Loki client is not reachable, we don't want LXD to be blocked from starting.


lxd/loki/loki.go
```diff
@@ -160,34 +162,46 @@ func (c *Client) run() {
...
+				scanner := bufio.NewScanner(io.LimitReader(resp.Body, maxErrMsgLen))
+				line := ""
 
-		return fmt.Errorf("Loki is not ready, server returned HTTP status %s (%d): %s", resp.Status, resp.StatusCode, line)
-	}
+				if scanner.Scan() {
+					line = scanner.Text()
+				}
+
+				logger.Warnf("Loki is not ready, server returned HTTP status %s (%d): %s", resp.Status, resp.StatusCode, line)
```
tomponline, line 198: Lets switch to contextual logging whilst we r here

lxd/loki/loki.go
```diff
@@ -160,34 +162,46 @@ func (c *Client) run() {
...
-	}
+func (c *Client) checkLoki(ctx context.Context) {
+	retryFrequency := time.NewTicker(10 * time.Second)
+	defer retryFrequency.Stop()
 
-	req = req.WithContext(ctx)
-	req.Header.Set("Content-Type", contentType)
+	for {
+		select {
+		case <-c.quit:
```
tomponline, line 171: We should check ctx.Done() here too right?

kadinsayani: Yes, thanks for pointing that out!

lxd/loki/loki.go
kadinsayani: Done ✅

kadinsayani:
Since we don't push any errors up the call stack and only return a client from `NewClient`, we no longer have error messages printed to standard output when setting the `loki.api.url` config key, however, I've added more robust logging and a retry mechanism to `checkLoki`.

tomponline:
How will incorrect urks be rejected when being set?

kadinsayani:
> How will incorrect urls be rejected when being set? Or are we not doimh that currently?

With the current approach in this PR incorrect urls will not be rejected, but there will be an error message in the logs. Currently, `NewClient` returns an error which prints to standard output.

tomponline:
Where does the `lxc config set` value get checked and rejected now?

kadinsayani:
> Where does the `lxc config set` value get checked and rejected now?

It currently doesn't with the changes in this PR. I'm going to work that in.

PR #15209, opened 2025-03-19
docs: add scripted grafana setup WD-20339

edlerd:
# Done
- added section to the Grafana setup to describe setup with script on a one node setup.


doc/howto/grafana.md
```diff
@@ -82,3 +82,25 @@ To set the `loki.instance` configuration key, run the following command:
...
+
+## Scripted setup and LXD-UI integration
+
+Alternatively use scripted setup of the Grafana dashboard. This only supports a single node LXD installation.
+
+1. Launch a new instance on your LXD server:
+
+       lxc launch ubuntu:20.04 grafana --project default
+
+1. Download and run the script to set up Grafana on the instance:
```
minaelee, line 94:
```suggestion
1. Run the following commands to download and execute the script to set up Grafana on the `grafana` instance:
```

doc/howto/grafana.md
```diff
@@ -82,3 +82,25 @@ To set the `loki.instance` configuration key, run the following command:
...
+
+       curl -s https://raw.githubusercontent.com/canonical/lxd-ui/refs/heads/main/public/assets/scripts/setup-grafana.sh -o /tmp/setup-grafana.sh
+       chmod +x /tmp/setup-grafana.sh
+       /tmp/setup-grafana.sh grafana default
+
+1. After the script finishes, sign in to Grafana with the default credentials `admin`/`admin` and change the password. Then import the LXD dashboard as described in step 3. from the guide above.
+
+The script installs Grafana, Prometheus, and Loki on a LXD instance. It also configures LXD to send metrics to Prometheus and logs to Loki. Additionally, it configures LXD-UI to be aware of the Grafana dashboard. This enables the UI to render a deep link {guilabel}`Metrics` to the Grafana dashboard from instance detail pages:
```
minaelee, line 102: In testing, all steps were successful _but_ the `Metrics` tab did not show up in my UI at the end:
image

minaelee (changes requested):

doc/howto/grafana.md
edlerd: Are you running the latest version 6.3? Other versions don't have the feature yet.

minaelee: Gotcha! Switched to 6.3 and the Metrics tab shows up now.

edlerd: Do you think we should mention this, or is it usually implicit through the docs version selector? As you look at the docs always for a specific lxd version.

edlerd: > Will the feature be backported to stable-5.2?

That is currently not planned. We might, but not right now.

tomponline: Please make sure the commit message flags this so i dont backport it

edlerd: - Added a note that LXD 6.3 minimum is needed to the docs.
- Added it to the commit message.

edlerd:
@minaelee thank you for the review. Please have another pass at this one :)

doc/howto/grafana.md
```diff
@@ -82,3 +82,27 @@ To set the `loki.instance` configuration key, run the following command:
...
+
+       curl -s https://raw.githubusercontent.com/canonical/lxd-ui/refs/heads/main/public/assets/scripts/setup-grafana.sh -o /tmp/setup-grafana.sh
+       chmod +x /tmp/setup-grafana.sh
+       /tmp/setup-grafana.sh grafana default
+
+1. After the script finishes, sign in to Grafana with the default credentials `admin`/`admin` and change the password.
+
+1. Import the LXD dashboard as described in step 3 of the manual steps in the preceding section.
+
+The script installs Grafana, Prometheus, and Loki on a LXD instance. It also configures LXD to send metrics to Prometheus and logs to Loki. Additionally, it configures LXD-UI to be aware of the Grafana dashboard. This enables the UI to render a deep link {guilabel}`Metrics` to the Grafana dashboard from instance detail pages (available since LXD 6.3):
```
minaelee, line 104:
```suggestion
The script installs Grafana, Prometheus, and Loki on a LXD instance. It also configures LXD to send metrics to Prometheus and logs to Loki. Additionally, it configures the LXD UI to be aware of the Grafana dashboard. This enables the UI to render a deep link {guilabel}`Metrics` to the Grafana dashboard from instance details pages (available since LXD 6.3):
```

doc/howto/grafana.md
```diff
@@ -82,3 +82,27 @@ To set the `loki.instance` configuration key, run the following command:
...
+
+![Metrics link in the LXD-UI instance detail page](/images/grafana_lxd_ui_metrics_integration.png)
```
minaelee, line 106:
```suggestion
![Metrics link in the LXD UI instance detail page](/images/grafana_lxd_ui_metrics_integration.png)
```

minaelee (changes requested):
Couple very small nits then LGTM :) Thank you!
Also, have the PNGs been optimized?

edlerd:
> Couple very small nits then LGTM :) Thank you! Also, have the PNGs been optimized?

Thanks for the review, adjusted the wording and optimized images with `optipng -o7`. Do you use another tool?

Please have another pass at this @minaelee

minaelee:
> Thanks for the review, adjusted the wording and optimized images with `optipng -o7`. Do you use another tool?

I use the same :) Thanks for the updates!

minaelee (dismissed):
LGTM, thank you!

doc/howto/grafana.md
```diff
@@ -82,3 +82,27 @@ To set the `loki.instance` configuration key, run the following command:
...
 `lxc config get loki.instance`
 ```
+
+## Scripted setup and LXD UI integration
+
+As an alternative to the manual steps above, we provide a script to set up the Grafana dashboard. This only supports a single-node LXD installation.
+
+1. Launch a new instance on your LXD server:
+
+       lxc launch ubuntu:20.04 grafana --project default
```
tomponline, line 92: 20.04 is almost eol, is this required?

doc/howto/grafana.md
```diff
@@ -82,3 +82,27 @@ To set the `loki.instance` configuration key, run the following command:
...
+
+1. Run the following commands to download and execute the script to set up Grafana on the `grafana` instance:
+
+       curl -s https://raw.githubusercontent.com/canonical/lxd-ui/refs/heads/main/public/assets/scripts/setup-grafana.sh -o /tmp/setup-grafana.sh
```
tomponline, line 96: Can we move this into the lxd repo so we arent askimg users to execute a script outside of the main repo.

tomponline:
@simondeziel would you mind giving the install script a once over for with your bash-foo please?

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,129 @@
...
+CONTAINER_UPLINK_IP=$(echo "$CONTAINER_IP" | cut -d "." -f1,2,3)".1"
+echo "Found container IP as '$CONTAINER_IP' and uplink as '$CONTAINER_UPLINK_IP'"
+
+set -e
+set -x
+
+# upload server.crt to container
+lxc info | sed -n "/BEGIN CERTIFICATE/,/END CERTIFICATE/p" | sed 's/^[ \t]*//;s/[ \t]*$//' > /tmp/server.crt
+lxc file push /tmp/server.crt "$INSTANCE"/root/server.crt --project="$PROJECT"
+rm /tmp/server.crt
```
simondeziel, line 29:
```suggestion
lxc info | sed -n "/BEGIN CERTIFICATE/,/END CERTIFICATE/ s/^\s*// p" | lxc file push - "$INSTANCE"/root/server.crt --project="$PROJECT"
```

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,129 @@
...
+apt-get update
+apt-get install -y apt-transport-https software-properties-common wget prometheus
+mkdir -p /etc/apt/keyrings/
+wget -q -O - https://apt.grafana.com/gpg.key | gpg --dearmor | tee /etc/apt/keyrings/grafana.gpg > /dev/null
+echo "deb [signed-by=/etc/apt/keyrings/grafana.gpg] https://apt.grafana.com stable main" | tee -a /etc/apt/sources.list.d/grafana.list
+apt-get update
+apt-get install -y grafana=11.4.0 loki=3.3.2 promtail=3.3.2
+systemctl daemon-reload
+systemctl start grafana-server
+systemctl enable grafana-server.service
```
simondeziel, line 45:
```suggestion
systemctl enable --now grafana-server.service
```

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,129 @@
...
+sed -ie '44d' /etc/loki/config.yml # fix the loki configuration, see https://github.com/grafana/loki/issues/15039
```
simondeziel, line 46: Snipping a line using a hardcoded number is too brittle. Is the goal to remove the `enabled: true` line under the `metric_aggregation` part?
https://github.com/grafana/loki/blob/v3.3.0/cmd/loki/loki-local-config.yaml#L41

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,129 @@
...
+systemctl start loki
+systemctl enable loki
+systemctl start promtail
+systemctl enable promtail
```
simondeziel, line 50:
```suggestion
systemctl enable --now loki
systemctl enable --now promtail
```

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,129 @@
...
+
+# generate ssl key for grafana to serve via https
+openssl req -x509 -newkey rsa:4096 -keyout /etc/grafana/grafana.key -out /etc/grafana/grafana.crt -days 365 -nodes -subj "/CN=metrics.local"
+chown grafana:grafana /etc/grafana/grafana.crt
+chown grafana:grafana /etc/grafana/grafana.key
+chmod 400 /etc/grafana/grafana.key /etc/grafana/grafana.crt
+sed -i "s#;protocol = http#protocol = https#" /etc/grafana/grafana.ini
```
simondeziel, line 57: It's nice to use HTTPS for Grafana's dashboard but I feel it might be a bit out of scope.
Anyway, I'd go with an ECDSA key (like the metrics one) and keep the `.key` as `root:grafana` with `0640`.

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,129 @@
...
+
+if [ "$#" -ne 2 ]; then
+    echo "Usage: $0 <instance> <project>"
+    echo "Error: Both 'instance' and 'project' arguments are required."
+    exit 1
+fi
+
+INSTANCE=$1
+PROJECT=$2
+
```
simondeziel, line 11: This will come handy later on.
```suggestion
SERVER_NAME="$(lxc info | awk '{if ($1 == "server_name:") print $2}')"
```

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,129 @@
...
+    scheme: 'https'
+    static_configs:
+      - targets: ['$CONTAINER_UPLINK_IP:8443']
+    tls_config:
+      ca_file: '/etc/prometheus/tls/server.crt'
+      cert_file: '/etc/prometheus/tls/metrics.crt'
+      key_file: '/etc/prometheus/tls/metrics.key'
+      # XXX: server_name is required if the target name
+      #      is not covered by the certificate (not in the SAN list)
+      server_name: '$HOSTNAME'
```
simondeziel, line 101:
```suggestion
      server_name: '$SERVER_NAME'
```

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,129 @@
...
+EOT
+systemctl daemon-reload
+systemctl start prometheus
+systemctl enable prometheus.service
```
simondeziel, line 105:
```suggestion
systemctl enable --now prometheus.service
```

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,129 @@
...
+EOF
+
+# download metrics.crt from container and add to host lxd trust store
+lxc file pull "$INSTANCE"/etc/prometheus/tls/metrics.crt --project="$PROJECT" /tmp/metrics.crt
+lxc config trust add /tmp/metrics.crt --type=metrics
+rm -rf /tmp/metrics.crt
+
+# configure host lxd for loki and grafana
+lxc config set user.grafana_base_url=https://"$CONTAINER_IP":3000/d/bGY-LSB7k/lxd?orgId=1
+lxc config set loki.api.url=http://"$CONTAINER_IP":3100 loki.instance=lxd &
```
simondeziel, line 115: Not sure why this config set is ran in the background? Is that to workaround the issue with Loki being unavailable making it slow?

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,129 @@
...
+
+# restart container
+lxc exec "$INSTANCE" --project="$PROJECT" reboot
```
simondeziel, line 118:
```suggestion
lxc restart "$INSTANCE" --project="$PROJECT"
```

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,129 @@
...
+EOT
+systemctl restart grafana-server
+
+# generate certs for prometheus
+openssl req -x509 -newkey ec -pkeyopt ec_paramgen_curve:secp384r1 -sha384 -keyout metrics.key -nodes -out metrics.crt -days 3650 -subj "/CN=metrics.local"
+mkdir /etc/prometheus/tls
+mv metrics.* /etc/prometheus/tls/
+mv server.crt /etc/prometheus/tls/
+chown -R prometheus:root /etc/prometheus/tls
+chmod 400 /etc/prometheus/tls/metrics.key /etc/prometheus/tls/metrics.crt /etc/prometheus/tls/server.crt
```
simondeziel, line 79: The only file that needs special handling is `metrics.key` which needs to be `root:prometheus` with `0640` perms. The rest can be `0644` perms.

scripts/setup-grafana.sh
edlerd: Yes, exactly, not to block the script execution in case loki is down.

scripts/setup-grafana.sh
edlerd: Yes, exactly, removing that line to work around the referenced bug in the github issue.

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,129 @@
...
+set -x
+set -e
+# install grafana and prometheus
+apt-get update
+apt-get install -y apt-transport-https software-properties-common wget prometheus
+mkdir -p /etc/apt/keyrings/
+wget -q -O - https://apt.grafana.com/gpg.key | gpg --dearmor | tee /etc/apt/keyrings/grafana.gpg > /dev/null
+echo "deb [signed-by=/etc/apt/keyrings/grafana.gpg] https://apt.grafana.com stable main" | tee -a /etc/apt/sources.list.d/grafana.list
+apt-get update
+apt-get install -y grafana=11.4.0 loki=3.3.2 promtail=3.3.2
```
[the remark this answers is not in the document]
edlerd: When weakening this, we run into a problem of having to ensure the installation still works the same and the script succeeds.

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,123 @@
...
+apt-get update
+apt-get install -y software-properties-common wget prometheus
+mkdir -p /etc/apt/keyrings/
+wget -q -O - https://apt.grafana.com/gpg.key | gpg --dearmor | tee /etc/apt/keyrings/grafana.gpg > /dev/null
+echo "deb [signed-by=/etc/apt/keyrings/grafana.gpg] https://apt.grafana.com stable main" | tee -a /etc/apt/sources.list.d/grafana.list
+apt-get update
+apt-get install -y grafana=11.4.0 loki=3.3.2 promtail=3.3.2
+systemctl daemon-reload
+systemctl enable --now grafana-server.service
+sed -ie '44d' /etc/loki/config.yml # fix the loki configuration, see https://github.com/grafana/loki/issues/15039
```
simondeziel, line 43:
```suggestion
sed -ie '/metric_aggregation:$/,/enabled:/ s/\(^\s*enabled:\s*\)true$/\1false/' /etc/loki/config.yml # fix the loki configuration, see https://github.com/grafana/loki/issues/15039
```

scripts/setup-grafana.sh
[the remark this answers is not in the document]
edlerd: updated the versions to latest.

scripts/setup-grafana.sh
edlerd: we don't need this anymore, because in loki 3.4.2 this was fixed :)

edlerd: removed the line.

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,122 @@
...
+systemctl enable --now prometheus.service
+EOF
+
+# download metrics.crt from container and add to host lxd trust store
+lxc file pull "$INSTANCE"/etc/prometheus/tls/metrics.crt --project="$PROJECT" /tmp/metrics.crt
+lxc config trust add /tmp/metrics.crt --type=metrics
+rm -rf /tmp/metrics.crt
+
+# configure host lxd for loki and grafana
+lxc config set user.grafana_base_url=https://"$CONTAINER_IP":3000/d/bGY-LSB7k/lxd?orgId=1
```
simondeziel, line 107: Is that HTTPS URL directly exposed to the user via the LXD UI? If so the user needs to accept the self-signed cert warning, right?

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,122 @@
...
+    url: http://$CONTAINER_IP:3100
+EOT
+systemctl restart grafana-server
+
+# generate certs for prometheus
+openssl req -x509 -newkey ec -pkeyopt ec_paramgen_curve:secp384r1 -sha384 -keyout metrics.key -nodes -out metrics.crt -days 3650 -subj "/CN=metrics.local"
+mkdir /etc/prometheus/tls
+mv metrics.* /etc/prometheus/tls/
+mv server.crt /etc/prometheus/tls/
+chown -R prometheus:root /etc/prometheus/tls
```
simondeziel, line 72:
```suggestion
chown root:prometheus /etc/prometheus/tls/metrics.key
```
This way, the prometheus daemon can read the private key to use it but cannot change it.

scripts/setup-grafana.sh
edlerd: yes, that is right.

scripts/setup-grafana.sh
edlerd: I think this didn't work for me earlier, I wouldn't get metrics with this change.

edlerd: Yes, this breaks the setup.

simondeziel: Hmm, could you share the output of `namei -l /etc/prometheus/tls/metrics.key`, please?

It should look like this:
```
# namei -l /etc/prometheus/tls/metrics.key
f: /etc/prometheus/tls/metrics.key
drwxr-xr-x root root       /
drwxr-xr-x root root       etc
drwxr-x--- root prometheus prometheus
drwxr-x--- root prometheus tls
-rw-r----- root prometheus metrics.key
```

edlerd: What is the problem with owning the whole directory and contents to the prometheus user and root group? Why change it in the first place to root user and prometheus group? The former works well already :)

simondeziel: Indeed, it's not about functional or not but more a question of least privileges.

edlerd: I think the existing `prometheus:root` is giving lower permissions than the suggested `root:prometheus`.

simondeziel: With your method, the prometheus daemon can alter any file under `/etc/prometheus/tls` which IMHO is undesirable. It only needs read access to all files which is why I suggested tweaking the permissions/ownership.

I just ran the script with my suggested changes and it works. Here's the resulting `namei` output:

```
root@grafana:~# namei -l /etc/prometheus/tls/metrics.key 
f: /etc/prometheus/tls/metrics.key
drwxr-xr-x root root       /
drwxr-xr-x root root       etc
drwxr-xr-x root root       prometheus
drwxr-xr-x root root       tls
-rw-r----- root prometheus metrics.key
```

The bit that was broken is when the `server.crt` file was pushed into the container. I'll propose a change to address this.

scripts/setup-grafana.sh
```diff
@@ -0,0 +1,121 @@
...
+
+CONTAINER_IP="$(lxc list "${INSTANCE}" --project "${PROJECT}" -c 4 -f csv | awk '{print $1}')"
+CONTAINER_UPLINK_IP=$(echo "$CONTAINER_IP" | cut -d "." -f1,2,3)".1"
+echo "Found container IP as '$CONTAINER_IP' and uplink as '$CONTAINER_UPLINK_IP'"
+
+set -e
+set -x
+
+# upload server.crt to container
+lxc info | sed -n "/BEGIN CERTIFICATE/,/END CERTIFICATE/ s/^\s*// p" | lxc file push - "$INSTANCE"/root/server.crt --project="$PROJECT"
```
simondeziel, line 27:
```suggestion
lxc info | sed -n "/BEGIN CERTIFICATE/,/END CERTIFICATE/ s/^\s*// p" | lxc file push --uid 0 --gid 0 --mode 0644 - "$INSTANCE"/root/server.crt --project="$PROJECT"
```

simondeziel, line 27:
```suggestion
lxc info | sed -n "/BEGIN CERTIFICATE/,/END CERTIFICATE/ s/^\s*// p" | lxc file push --uid 0 --gid 0 --mode 0644 - "$INSTANCE"/root/server.crt --project="$PROJECT"
```

simondeziel (changes requested):

edlerd:
Thanks for the review @simondeziel I can confirm with recent suggestions all works fine and the script become much simpler :)

Please have another pass at the script.

simondeziel (approved):
LGTM, thanks!

PR #15211, opened 2025-03-19
lxd/storage/drivers/pure: Unmap volume when connection is freshly created

### LINKED ISSUE #15210: PureStorage volume not removed if mount fails
When creating a PureStorage volume, if the mount fails due to a timeout while waiting for the device to appear on the LXD host, the volume remains on the PureStorage array.

The reverter handles volume deletion. However, if the volume is still connected (as is the case when the mount fails), the deletion fails because the volume is still connected (associated with a PureStorage host).

The reverter must ensure that the volume is disconnected before attempting deletion.

Reported-by: @escabo 

---

To reproduce the issue, I have set the timeout to wait for the disk device to appear to `0`. This results in the volume being consistently left on the PureStorage array.

MusicDin:
Fixes left-over volumes on the PureStorage array when volume creation fails.

The `unmapVolume` must be added to `outerReverter` when we are the first to associate (connect) the volume with the PureStorage host. The comment above is correct, just the condition was wrong.

Fixes #15210

### DIFF
--- a/lxd/storage/drivers/driver_pure_util.go
+++ b/lxd/storage/drivers/driver_pure_util.go
@@ -1270,7 +1270,7 @@ func (d *pure) mapVolume(vol Volume) (cleanup revert.Hook, err error) {
 	// returned (outer) reverter. Unmap ensures the target is disconnected only when
 	// no other device is using it.
 	outerReverter := revert.New()
-	if !connCreated {
+	if connCreated {
 		outerReverter.Add(func() { _ = d.unmapVolume(vol) })
 	}
 


MusicDin:
Tests are still running, will mark it as ready once all tests pass.

roosterfish:
Checked the corresponding section in PowerFlex, looks to be not affected.

MusicDin:
Thanks for confirmation. I've checked it as well and only PureStorage is affected.

tomponline (approved):
thanks!

Issue #15208, opened 2025-03-19
Ceph RBD driver doesn't apply updates of `pg_num` setting


roosterfish:
When creating a new storage pool using the Ceph RBD storage driver, you can set the number of placement groups (pg) by specifying the `ceph.osd.pg_num` setting.

LXD allows the modification of this setting using `lxc storage set <pool> ceph.osd.pg_num 64` but doesn't apply the change on the Ceph backend and only changes the config setting in the database.

Currently it seems that only the `ceph.osd.pool_size` setting gets updated on the backend when modified, see https://github.com/canonical/lxd/blob/main/lxd/storage/drivers/driver_ceph.go#L412.

As part of this it might be worth checking the other config keys too to allow modification after pool creation.

PR #15214, opened 2025-03-19
Storage: Fix persisting Ceph RBD config

roosterfish:
This PR ensures when updating `ceph.osd.pg_num` the change is also applied on the backend.

Furthermore it blocks `ceph.cluster_name` and `ceph.osd.pool_name` from being updated after pool creation as it will break the pool.

Fixes https://github.com/canonical/lxd/issues/15208


tomponline (approved):
ta

PR #15217, opened 2025-03-19
Do not use Go toolchain directive

simondeziel:
We don't want to use any [Go toolchain](https://go.dev/doc/toolchain) as it introduces another variable (an uncontrolled one at that) in the testing matrix.


tomponline (approved):
Thanks!

PR #15219, opened 2025-03-20
LVM: Fix ref count inconsistency

hamistao:
Fixes the problem caught in https://github.com/canonical/lxd-ci/actions/runs/13923914369/job/38963535130.

When deleting a snapshot for a running instance, its deactivation ref count was decremented to 0, which prompted LXD to thing deativating the volume group was safe, when in fact the VM root volume that is part of the same volume groups as the snapshot was still in use as the VM was running.

To fix this, we first use `UnmountVolumeSnapshot` instead of `UnmountVolume` when deleting snapshots and rework `UnmountVolumeSnapshot` to decrement the ref count only if the snapshot was just unmounted.

### DIFF
--- a/lxd/storage/drivers/driver_lvm_volumes.go
+++ b/lxd/storage/drivers/driver_lvm_volumes.go
@@ -1007,7 +1007,7 @@ func (d *lvm) DeleteVolumeSnapshot(snapVol Volume, op *operations.Operation) err
 	}
 
 	if lvExists {
-		_, err = d.UnmountVolume(snapVol, false, op)
+		_, err = d.UnmountVolumeSnapshot(snapVol, op)
 		if err != nil {
 			return fmt.Errorf("Error unmounting LVM logical volume: %w", err)
 		}
--- a/lxd/storage/drivers/driver_lvm_volumes.go
+++ b/lxd/storage/drivers/driver_lvm_volumes.go
@@ -791,7 +791,7 @@ func (d *lvm) MountVolume(vol Volume, op *operations.Operation) error {
 }
 
 // UnmountVolume unmounts volume if mounted and not in use. Returns true if this unmounted the volume.
-// keepBlockDev indicates if backing block device should be not be deactivated when volume is unmounted.
+// keepBlockDev indicates if backing block device should not be deactivated when volume is unmounted.
 func (d *lvm) UnmountVolume(vol Volume, keepBlockDev bool, op *operations.Operation) (bool, error) {
 	unlock, err := vol.MountLock()
 	if err != nil {
--- a/lxd/storage/drivers/driver_lvm_volumes.go
+++ b/lxd/storage/drivers/driver_lvm_volumes.go
@@ -834,6 +834,8 @@ func (d *lvm) UnmountVolume(vol Volume, keepBlockDev bool, op *operations.Operat
 			return false, fmt.Errorf("Failed to unmount LVM logical volume: %w", err)
 		}
 
+		d.logger.Debug("Unmounted logical volume", logger.Ctx{"volName": vol.name, "path": mountPath, "keepBlockDev": keepBlockDev})
+
 		ourUnmount = true
 	} else if vol.contentType == ContentTypeBlock {
 		volDevPath := d.lvmDevPath(d.config["lvm.vg_name"], vol.volType, vol.contentType, vol.name)
@@ -847,10 +849,6 @@ func (d *lvm) UnmountVolume(vol Volume, keepBlockDev bool, op *operations.Operat
 		if err != nil {
 			return false, err
 		}
-
-		d.logger.Debug("Unmounted logical volume", logger.Ctx{"volName": vol.name, "path": mountPath, "keepBlockDev": keepBlockDev})
-
-		ourUnmount = true
 	} else {
 		// Since activation of the LV on mount is unconditional, the activation
 		// refcount needs to be updated regardless of whether we actually
--- a/lxd/storage/drivers/driver_lvm_volumes.go
+++ b/lxd/storage/drivers/driver_lvm_volumes.go
@@ -1169,13 +1169,22 @@ func (d *lvm) UnmountVolumeSnapshot(snapVol Volume, op *operations.Operation) (b
 
 	refCount := snapVol.MountRefCountDecrement()
 
+	// For VMs, unmount the filesystem volume.
+	if snapVol.IsVMBlock() {
+		fsVol := snapVol.NewVMBlockFilesystemVolume()
+		ourUnmount, err = d.UnmountVolumeSnapshot(fsVol, op)
+		if err != nil {
+			return false, err
+		}
+	}
+
+	if refCount > 0 {
+		d.logger.Debug("Skipping unmount as in use", logger.Ctx{"volName": snapVol.name, "refCount": refCount})
+		return false, ErrInUse
+	}
+
 	// Check if already mounted.
 	if snapVol.contentType == ContentTypeFS && filesystem.IsMountPoint(mountPath) {
-		if refCount > 0 {
-			d.logger.Debug("Skipping unmount as in use", logger.Ctx{"volName": snapVol.name, "refCount": refCount})
-			return false, ErrInUse
-		}
-
 		err = TryUnmount(mountPath, 0)
 		if err != nil {
 			return false, fmt.Errorf("Failed to unmount LVM snapshot volume: %w", err)
@@ -1198,37 +1207,23 @@ func (d *lvm) UnmountVolumeSnapshot(snapVol Volume, op *operations.Operation) (b
 			}
 		}
 
-		// We only deactivate filesystem volumes if an unmount was needed to better align with our
-		// unmount return value indicator.
-		_, err = d.deactivateVolume(snapVol)
-		if err != nil {
-			return false, err
-		}
-
 		ourUnmount = true
 	} else if snapVol.contentType == ContentTypeBlock {
-		// For VMs, unmount the filesystem volume.
-		if snapVol.IsVMBlock() {
-			fsVol := snapVol.NewVMBlockFilesystemVolume()
-			ourUnmount, err = d.UnmountVolumeSnapshot(fsVol, op)
-			if err != nil {
-				return false, err
-			}
-		}
-
 		volDevPath := d.lvmDevPath(d.config["lvm.vg_name"], snapVol.volType, snapVol.contentType, snapVol.name)
 		if shared.PathExists(volDevPath) {
 			if refCount > 0 {
 				d.logger.Debug("Skipping unmount as in use", logger.Ctx{"volName": snapVol.name, "refCount": refCount})
 				return false, ErrInUse
 			}
+		}
+	}
 
-			_, err = d.deactivateVolume(snapVol)
-			if err != nil {
-				return false, err
-			}
-
-			ourUnmount = true
+	// We only deactivate filesystem volumes if an unmount was needed to better align with our
+	// unmount return value indicator.
+	if ourUnmount {
+		_, err = d.deactivateVolume(snapVol)
+		if err != nil {
+			return false, err
 		}
 	}
 


lxd/storage/drivers/driver_lvm_volumes.go
```diff
@@ -1198,37 +1207,23 @@ func (d *lvm) UnmountVolumeSnapshot(snapVol Volume, op *operations.Operation) (b
...
-				return false, err
-			}
-
-			ourUnmount = true
+	// We only deactivate filesystem volumes if an unmount was needed to better align with our
+	// unmount return value indicator.
+	if ourUnmount {
+		_, err = d.deactivateVolume(snapVol)
+		if err != nil {
+			return false, err
```
hamistao, line 1226: @MggMuggins I am still unsure about this part, since the `UnmountVolume` decreases the ref count even if the volume was not unmounted. But since this is a snapshot, decreasing it here would trigger the problem we are trying to fix so I believe this to be correct here. What do you think?

lxd/storage/drivers/driver_lvm_volumes.go
```diff
@@ -1198,37 +1207,23 @@ func (d *lvm) UnmountVolumeSnapshot(snapVol Volume, op *operations.Operation) (b
...
 			}
+		}
+	}
 
-			_, err = d.deactivateVolume(snapVol)
-			if err != nil {
-				return false, err
-			}
-
-			ourUnmount = true
```
MggMuggins, line 1233: Don't we still need to deactivate the volume if `volDevPath` exists and the refcount was 0?

lxd/storage/drivers/driver_lvm_volumes.go
MggMuggins: Every time `MountVolume` is called, there is a corresponding `UnmountVolume`. Since `MountVolume` _always_ activates the volume, we _always_ need to decrement the refCount when `UnmountVolume` is called.

I would think that the same set of semantics applies to `MountVolumeSnapshot` and `UnmountVolumeSnapshot`. I'm guessing you'll need to add an `activationRefCountIncrement` to `MountVolumeSnapshot` as it doesn't activate FS volumes when they are already mounted. Then you just need to make sure that `activationRefCountDecrement` is called every time `UnmountVolumeSnapshot` is called, and the refcounts should accurately reflect how many different things are using that "activation group" (vol and its snaps).

Does that make any sense?

MggMuggins (changes requested):

lxd/storage/drivers/driver_lvm_volumes.go
```diff
@@ -834,6 +834,8 @@ func (d *lvm) UnmountVolume(vol Volume, keepBlockDev bool, op *operations.Operat
 			return false, fmt.Errorf("Failed to unmount LVM logical volume: %w", err)
 		}
 
+		d.logger.Debug("Unmounted logical volume", logger.Ctx{"volName": vol.name, "path": mountPath, "keepBlockDev": keepBlockDev})
```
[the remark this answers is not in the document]
hamistao: In this case, I don't think logging a message for the Unmount start and finish makes much sense. Plus this log message was referrering to the the linux "Unmount" so moving it to just after the actual operation is more intuitive.

lxd/storage/drivers/driver_lvm_volumes.go
```diff
@@ -1007,7 +1007,7 @@ func (d *lvm) DeleteVolumeSnapshot(snapVol Volume, op *operations.Operation) err
 	}
 
 	if lvExists {
-		_, err = d.UnmountVolume(snapVol, false, op)
+		_, err = d.UnmountVolumeSnapshot(snapVol, op)
```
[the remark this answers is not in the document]
hamistao: Still, I think it is worth it to factor out the common logic into a separate function.

lxd/storage/drivers/driver_lvm_volumes.go
hamistao: I think I follow for the most part. You say that `MountVolume always activates the volume`, but can a volume (and its snapshots) be activated again if it is already activated?

If we want to deactivate the volume when it and its snapshots are all unmounted, why not simply increment/decrement when the volumes are actually mounted/unmounted?

This way the refcount value would always reflect the current number of mounted volumes in that group of volumes and we can safely deactivate when that number reaches 0. This seems simpler to understand and mantain while serving the same purpose, but perhaps I am missing something.

lxd/storage/drivers/driver_lvm_volumes.go
hamistao: Indeed, thanks for catching!

lxd/storage/drivers/driver_lvm_volumes.go
MggMuggins: > why not simply increment/decrement when the volumes are actually mounted/unmounted
Because it's unknown how many disk devices are using a volume, so we don't know when its safe to deactivate it. It's trivial to query the state of the volume, but that doesn't tell us how many users it has.

`activateVolume` succeeding is only a guarantee that the volume is activated and available (it can be called multiple times for the same volume if already activated). There should be a 1:1 correspondence between calls to `activateVolume` and `deactivateVolume`.

When we check if an FS volume is already mounted, it implies a use of `activateVolume`, and a call to `deactivateVolume` when the storage volume is unmounted:
https://github.com/canonical/lxd/blob/edb3c75644f486bf2000767193cf7e4cad4c9c8d/lxd/storage/drivers/driver_lvm_volumes.go#L1172-L1175

My thought is that the if statement on 1175 just needs an `else` that calls `incrementActivationRefCount`.

hamistao: Got it, thanks! I will add the increment as you suggested and I may add some tests to check we are deactivating the volume appropriately.

hamistao:
Back to draft to address comments from @MggMuggins and add tests to check if we are deactivating logical volumes when appropriate.

lxd/storage/drivers/driver_lvm_volumes.go
MggMuggins: Tests would be really awesome, especially in the container suite

hamistao: Sorry, one more question, I understand that for every call of `UnmountVolume[Snapshot]` we must decrement the ref count, so why add an increment call to `UnmountVolumeSnapshot`? Perhaps on the imprementation you marked somehow we decremented twice, could you confirm this applies to `unmountCommon`, the new function introduced here?

MggMuggins: Crud, I'm sorry. I meant to refer to these lines in the comment above:
https://github.com/canonical/lxd/blob/edb3c75644f486bf2000767193cf7e4cad4c9c8d/lxd/storage/drivers/driver_lvm_volumes.go#L1061-L1064

No need to increment the refcount in Unmount, my bad.

hamistao: Oh cool, on the refactoring this has been dealt with already, thanks!

tomponline:
@hamistao tests failing

hamistao:
@tomponline Tests fixed

lxd/storage/drivers/driver_lvm_volumes.go
```diff
@@ -1079,9 +1080,12 @@ func (d *lvm) DeleteVolumeSnapshot(snapVol Volume, op *operations.Operation) err
...
 	if lvExists {
-		_, err = d.UnmountVolumeSnapshot(snapVol, op)
-		if err != nil {
-			return fmt.Errorf("Error unmounting LVM logical volume: %w", err)
+		// Only call UnmountVolumeSnapshot if mounted to avoid breaking deactivaion ref counts.
+		if snapVol.contentType == ContentTypeFS && filesystem.IsMountPoint(snapVol.MountPath()) {
+			_, err = d.UnmountVolumeSnapshot(snapVol, op)
+			if err != nil {
+				return fmt.Errorf("Error unmounting LVM logical volume: %w", err)
+			}
```
MggMuggins, line 1088: Under what conditions can `DeleteVolumeSnapshot` be called and still have the logical volume mounted? I sorta assume that `DeleteVolume{Snapshot}` should fail if there are still nonzero refcounts around (indicating that the volume is still being used by something).

lxd/storage/drivers/driver_lvm_volumes.go
```diff
@@ -1048,193 +1122,13 @@ func (d *lvm) DeleteVolumeSnapshot(snapVol Volume, op *operations.Operation) err
...
-			})
-
-			// We are going to mount the temporary volume instead.
-			mountVol = tmpVol
-		}
-
-		volDevPath := d.lvmDevPath(d.config["lvm.vg_name"], mountVol.volType, mountVol.contentType, mountVol.name)
-
-		// Activate volume if needed.
-		_, err = d.activateVolume(mountVol)
```
MggMuggins, line 1106: I may be very wrong here, but I'm not sure I understand why this is safe to remove; have you tested this code with an lvm-thin pool? For thick pools, I would expect not to need another snapshot activation (as its a snapshot of the same parent), but I'm not sure that this will work in a thin pool.

hamistao: Yes I tested it with all pools. Pehaps there is some confusion here, this commit does not remove this `activateVolume` call, but makes sure that `MountVolume` always calls it by moving it outside an `if` block.

hamistao: Matching the previous behavior of `MountVolume`.

lxd/storage/drivers/driver_lvm_volumes.go
hamistao: To be honest I didn't look super deep into this, but `DeleteVolume{Snapshot}` are used in revert calls and I think in some ocasions the volumes deleted for a revert operation can be mounted. Probably an edge case but enough to make me not feel confident just failing here, I can look deeper into this if we feel it is needed.

lxd/storage/drivers/driver_lvm_volumes.go
MggMuggins: Sure, that much is correct AFAICT. My concern is that the `regenerateFSUUID` stuff changes the underlying volume that's getting mounted (`tmpVol`); does that volume need to be activated as well?

hamistao: Oh I see. I managed to reproduce this by attaching a custom volume snapshot with a xfs filesystem and indeed this raises an error. Nice spotting!

hamistao: Converting to draft until this is addressed

hamistao: Fixed

MggMuggins:
Thanks for the fix :slightly_smiling_face: 

Do you mean to replace [this](https://github.com/hamistao/lxd/blob/604f533509f3649e8f053065332a8634b8b6ab7f/lxd/storage/drivers/driver_lvm_volumes.go#L880-L888)?

Does it even need to be deferred? Is an unconditional call to `deactivateVolume` at the end of the function appropriate? I don't remember the purpose of the `keepBlockDev` parameter off the top of my head, but I would think that an unconditional `deactivateVolume` would violate it. Is there any functional reason for the change?

hamistao:
@MggMuggins No there isn't any functional reason for the change. Indeed it would beak `keepBlockDev=true`, so let's keep this as is.

lxd/storage/drivers/driver_lvm_volumes.go
```diff
@@ -218,7 +218,8 @@ func (d *lvm) DeleteVolume(vol Volume, op *operations.Operation) error {
 	}
 
 	if lvExists {
-		if vol.contentType == ContentTypeFS {
+		// Only call UnmountVolume if mounted to avoid breaking deactivaion ref counts.
+		if vol.contentType == ContentTypeFS && filesystem.IsMountPoint(vol.MountPath()) {
```
tomponline, line 222: Good catches.

This is interesting, as whenever I see a change like this I ask myself "what do the other storage drivers do"?
As these situations can be an opportunity to fix a "theme" of bugs, rather than just in one storage driver.

From a quick scan of the other storage driver's implementations of `DeleteVolume` it looks like only `lvm` and `ceph` have this problem.

The other similar storage drivers, zfs, powerflex, pure etc all seem to just deactivate the volume and delete it and don't trigger an unmount of the filesystem. There appears to be an expectation/convention that when calling `.DeleteVolume()` that volumes are unused by this point (not attached to anything, nor mounted).

So now I'm thinking, why only these 2 storage drivers do this?

Perhaps this has been a done as historical this would have been shortcut do deactivating/unmap the volumes as well (which both lvm and ceph rbd need).

I'm happy to leave this as you propose, but I do think we should also look at the `ceph` driver to check if it has the same issue.

tomponline (approved):
LGTM thanks! Nice sleuthing :)

lxd/storage/drivers/driver_lvm_volumes.go
hamistao: Thanks for the insight, will look into it :D

PR #15220, opened 2025-03-20
lxd/network/common: Handle missing BGP peer (from Incus)

nmezhenskyi:
Includes cherry-pick from https://github.com/lxc/incus/pull/1818.


tomponline (approved):
Ta!

PR #15154, opened 2025-03-10
VM: Fix QMP arguments typing to support QEMU 9.2 (from Incus)

nmezhenskyi:
Cherry-picks from https://github.com/lxc/incus/pull/1531, https://github.com/lxc/incus/pull/1601, and https://github.com/lxc/incus/pull/1196.

Fixes https://github.com/canonical/lxd/issues/14916.

PR #15225, opened 2025-03-21
VM: Use full device name in qemu's `device_id`

hamistao:
In QEMU 9.1.1, using the device serial as a default for its ID fails if the serial is longer than 20 characters (https://github.com/lxc/incus/issues/1536). So a cherry pick that we performed from Incus made it so we explicitly provide the `device_id` (https://github.com/canonical/lxd/pull/15154/commits/e0f99467da0ae6431c716d1ce6a66d41d34d104b). This changed the format the `device_id` takes and broke our tests (https://github.com/canonical/lxd-ci/actions/runs/13923914369/job/38963532720).

This proposes an improvement to the format of `device_id`. I concluded from manually testing that we don't have a lenght limit for `device_id`, we can just use the full `lxd_<escaped_lxd_device_name>`. lxc-ci tests are adapted in https://github.com/canonical/lxd-ci/pull/434.

### DIFF
--- a/lxd/instance/drivers/driver_qemu.go
+++ b/lxd/instance/drivers/driver_qemu.go
@@ -4074,6 +4074,13 @@ func (d *qemu) addDriveConfig(qemuDev map[string]any, bootIndexes map[string]int
 
 	qemuDeviceSerial := qemuDeviceNamePrefix + escapedDeviceName
 	qemuDev["serial"] = qemuDeviceSerial
+
+	// Trim the serial down to 36 characters if longer than that, since this is the max size of a serial in QEMU.
+	// Do not hash as to not break older guests that were relying on QEMU to reduce the size of the device serial.
+	if len(qemuDeviceSerial) > 36 {
+		qemuDev["serial"] = qemuDeviceSerial[:36]
+	}
+
 	if bus == "virtio-scsi" {
 		qemuDev["device_id"] = qemuDeviceSerial
 		qemuDev["channel"] = 0
--- a/lxd/instance/drivers/driver_qemu.go
+++ b/lxd/instance/drivers/driver_qemu.go
@@ -4071,10 +4071,11 @@ func (d *qemu) addDriveConfig(qemuDev map[string]any, bootIndexes map[string]int
 	}
 
 	qemuDev["drive"] = qemuDevDrive
-	qemuDev["serial"] = qemuDeviceNamePrefix + escapedDeviceName
 
+	qemuDeviceSerial := qemuDeviceNamePrefix + escapedDeviceName
+	qemuDev["serial"] = qemuDeviceSerial
 	if bus == "virtio-scsi" {
-		qemuDev["device_id"] = qemuDeviceNameOrID(qemuDeviceNamePrefix, driveConf.DevName, "", qemuDeviceNameMaxLength)
+		qemuDev["device_id"] = qemuDeviceSerial
 		qemuDev["channel"] = 0
 		qemuDev["lun"] = 1
 		qemuDev["bus"] = "qemu_scsi.0"


tomponline:
> lxc-ci tests are adapted in [canonical/lxd-ci#434](https://github.com/canonical/lxd-ci/pull/434).

@hamistao @skatsaounis is this not an "API break" for the guest OS where the disk device path can change between LXD versions, this might break custom volume mount rules inside the guest.

We've discussed in the past that we need to be careful about changing stuff like this.

tomponline:
> The VM keeps its `/dev/disk/by-id` path until the next restart, in which it uses the new format, the device works fine both before and after restart.

Does the /dev/disk/by-id path change after restart? If so then this is a problem.

tomponline:
> > The VM keeps its `/dev/disk/by-id` path until the next restart, in which it uses the new format, the device works fine both before and after restart.
> 
> Does the /dev/disk/by-id path change after restart? If so then this is a problem.

Is the only change to support longer device names? Are shorter ones unchanged?

hamistao:
> Does the /dev/disk/by-id path change after restart? If so then this is a problem.

@tomponline They do for long names. They had already changed after 6.3 with https://github.com/canonical/lxd/commit/e0f99467da0ae6431c716d1ce6a66d41d34d104b, so this changes it to something else instead.

If we want to address the limitation on QEMU 9.1.1 AND keep the existing behavior the same, we can explicitly set the device's `device_id` to a trimmed down version of the full `lxd_<escaped_device_name>`, let me know if that would be preferred.

> Is the only change to support longer device names? Are shorter ones unchanged?

Yes, shorter ones are unchanged. This is just to make long names not trimmed down when used as a `device_id`.

hamistao:
>Does the /dev/disk/by-id path change after restart? If so then this is a problem.

@tomponline We can also consider a method to prevent the `device_id` to change after the first start of an instance, but seems too complex of a solution here.

tomponline:
> > Does the /dev/disk/by-id path change after restart? If so then this is a problem.
> 
> @tomponline We can also consider a method to prevent the `device_id` to change after the first start of an instance, but seems too complex of a solution here.

Lets discuss further as i do not understand why the change has been introduced in the first place.

We need to consider any potential change to paths exposed inside the guest as a breaking change, this is why we have those tests afterall. So when the tests fail we shouldnt just update the tests.

tomponline:
> They had already changed after 6.3 with [e0f9946](https://github.com/canonical/lxd/commit/e0f99467da0ae6431c716d1ce6a66d41d34d104b), so this changes it to something else instead.

Yes that change needs reverting.

tomponline:
> Yes, shorter ones are unchanged. This is just to make long names not trimmed down when used as a `device_id`

And has any consideration been made for users/admins inside guest VMs that were relying on the previous form of path?

hamistao:
> Lets discuss further as i do not understand why the change has been introduced in the first place.

@tomponline It is all in the issue I linked to: https://github.com/lxc/incus/issues/1536

> And has any consideration been made for users/admins inside guest VMs that were relying on the previous form of path?

I was never completely set on any approach but the changes are very simple so I decided to open this PR to at least raise some discussion. And the first change to the path was approved both on Incus and here so I figured this was a good approach to start. 

Taking all into account, I suppose it makes sense to simply revert the changes for now and including this and the tests PR in a future release where we bump our QEMU up, this way we can save the change to the path for a release that will probably also require other changes on the QEMU driver, instead of chaging its behavior bit by bit. And at the same time improve the path generation. What do you think?

tomponline:
@hamistao so as discussed in our meeting earlier lets:

1. Check that the unbounded device_id is supported by QEMU 9.2
2. If so, then decide whether we will land the unbounded support in LXD 6.x (it would be nice to because it supports longer device IDs in a clearer way), but my view is we shouldn't land this in LXD 5.21.x as its a potentially breaking changed to the /dev/by-id paths in guests. 

cc @skozina

hamistao:
@tomponline The commit that introduced the new behavior of failing when a serial longer than 20 characters when being used as a default for `device_id` also makes it so qemu fails when the serial is longer than 36 characters, even if a `device_id` is provided. 

A bit of reasoning and history is contained on the commit message: https://gitlab.com/qemu-project/qemu/-/commit/75997e182b695f2e3f0a2d649734952af5caf3ee.

So I updated this PR to also truncate the `serial` instead of relying on qemu to do so, avoiding the `The serial number can't be longer than 36 characters` error on newer qemu.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -37,7 +37,7 @@ import (
 	"google.golang.org/protobuf/proto"
 	"gopkg.in/yaml.v2"
 
-	"github.com/canonical/lxd/client"
+	lxd "github.com/canonical/lxd/client"
```
tomponline, line 40: Oops

tomponline:
> Looking at this changed file also shows that no limit is enforced on `device_id`s provided by the user. I made some experiments to confirm this

Thanks great investigation!

lxd/instance/drivers/driver_qemu.go
hamistao: Hmm that worries me, my vscode added that since this package is already referenced by that name throughout this file. I added another commit proposing calling this package `lxdClient` and preventing others from making the same mistake.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -37,7 +37,7 @@ import (
 	"google.golang.org/protobuf/proto"
 	"gopkg.in/yaml.v2"
 
-	"github.com/canonical/lxd/client"
+	lxdClient "github.com/canonical/lxd/client"
```
tomponline, line 40: lets stick with what we use everywhere else in the code base of just `"github.com/canonical/lxd/client"` and then update your editor to stop auto changing it.

We can discuss whether we want to use a different import across the entire code base separately, but is outside of scope of this PR. Ta

hamistao: OK, done!

tomponline (approved):
Thanks!

Issue #12667, opened 2023-12-12
Investigate LXD shutdown timeout/behavior


simondeziel:
Someone complained that LXD was taking too long to shutdown while rebooting their machine (https://chat.canonical.com/canonical/pl/e8w39m7axi8z9xu5khnjhj7p6h). There are at least 2 timeouts at play here, one is from LXD itself `5m` (https://documentation.ubuntu.com/lxd/en/latest/server/#server-core:core.shutdown_timeout) and the other is from the systemd unit of the snap `TimeoutStopSec=600` (https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft.yaml#L83).

While users would like shutdown to be quick, LXD needs to ensure instances are provided with enough time to do an orderly shutdown otherwise they risk loosing data. On top of that, LXD has to pace how many instances are signaled for parallel shutdown to avoid a deluge of IOs associated with too many instances shutting down at the same time. Having a static limit off either 5m or 10m is probably not ideal if there are many instances or if some take a lot of time to properly shutdown. Windows VMs applying updates can easily take upward 10 minutes to shutoff for example.

On the systemd side, there is a way to postpone the timeout:

> If a service of Type=notify sends "EXTEND_TIMEOUT_USEC=...", this may cause the stop time to be extended beyond
> TimeoutStopSec=. The first receipt of this message must occur before TimeoutStopSec= is exceeded, and once the stop time
> has extended beyond TimeoutStopSec=, the service manager will allow the service to continue to stop, provided the service
> repeats "EXTEND_TIMEOUT_USEC=..."  within the interval specified, or terminates itself (see sd_notify(3)).

In the LXD case, it would require to change the `daemon` unit from `Type=simple` to `Type=notify` which is possible for snaps to do (https://forum.snapcraft.io/t/services-and-daemons/12601).

MggMuggins:
Marginally related behavior:

My cluster decided to snap refresh while I had a spice console open (`lxc console --type=vga machine0` to the remote that was refreshing); looks like that blocked LXD from shutting down for a while (causing some of my other API requests to fail). I assume this would also be true for terminal console sessions and lxd shell. LXD did eventually close the websocket from the server side and restart (`core.shutdown_timeout`).

In the moment it was a little unclear to me as an operator what was going on (even with my LXD experience). @tomponline mentioned the possibility of requesting that snapd delay a refresh until later if the user has a long-running interactive request open.

PR #15247, opened 2025-03-24
shutdown: Extend the systemd LXD daemon stop timeout internally 

gabrielmougard:
closes https://github.com/canonical/lxd/issues/12667

If LXD is still shutting down but we detect that we're coming 'close' to the `stop-timeout` value defined in the lxd snap (original stop-timeout minus a padding duration, in our case we chose `20s`), we extend this duration by a small duration (`1min`)


gabrielmougard:
tests WIP

lxd/resources/system.go
```diff
@@ -426,3 +432,57 @@ func systemGetMotherboard() (*api.ResourcesSystemMotherboard, error) {
...
+	}
+
+	file, err := os.Open(serviceFile)
+	if err != nil {
+		return 0, fmt.Errorf("Failed to open %s: %w", serviceFile, err)
+	}
+
+	defer file.Close()
+	scanner := bufio.NewScanner(file)
+	timeoutRegex := regexp.MustCompile(`^\s*TimeoutStopSec\s*=\s*(\d+)?`)
```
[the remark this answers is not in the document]
simondeziel: `regexp.MustCompile` should be avoided as it would cause a `panic` if the regex compilation fails due to OOM for example.

lxd/resources/system.go
```diff
@@ -426,3 +432,57 @@ func systemGetMotherboard() (*api.ResourcesSystemMotherboard, error) {
...
+	if err != nil {
+		return fmt.Errorf("Failed to extend systemd timeout: %w", err)
+	}
+
+	return nil
+}
+
+// GetSystemdStopTimeout returns the systemd stop timeout
+func GetSystemdStopTimeout() (time.Duration, error) {
+	serviceFile := "/var/lib/snapd/hostfs/etc/systemd/system/snap.lxd.daemon.service"
```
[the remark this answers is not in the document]
gabrielmougard: If we want to use `systemctl` we need to add a `run-host` wrapper on `/run/bin/systemctl` in the lxd snap otherwise we wont be allowed to use it. Do we want to do that?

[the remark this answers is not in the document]
simondeziel: I don't like the added complexity but it would be safer IMHO. People can provide overrides to units so it's possible that some are already tweaking that `TimeoutStopSec=` directive to better suite their environment.

simondeziel (commented):
By switching the LXD daemon to using `Type=nofity`, I think we also need to send the `READY=1` notification once initialization is completed.

lxd/resources/system.go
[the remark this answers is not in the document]
gabrielmougard: fixed

lxd/resources/system.go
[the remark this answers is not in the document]
gabrielmougard: fixed

gabrielmougard:
local testing works!

I created an instance c1 and just shell into it. I then sent the kill SIGPWR sig to the lxd process and after the snap stop-timeout - 20secs (my padding duration), can see the `LXD snap daemon stop timeout extended` message with a new timeout being `11min` . After closing the shell session, the ticking goroutine closes normally.

gabrielmougard:
@tomponline CI is green.

lxd/resources/system.go
```diff
@@ -426,3 +430,99 @@ func systemGetMotherboard() (*api.ResourcesSystemMotherboard, error) {
...
+
+		total += time.Duration(num * float64(multiplier))
+	}
+
+	return total, nil
+}
+
+// GetSystemdStopTimeout returns the systemd stop timeout for the LXD daemon.
+func GetSystemdStopTimeout(ctx context.Context) (time.Duration, error) {
+	output, err := shared.RunCommandContext(ctx, "systemctl", "show", "snap.lxd.daemon", "--property=TimeoutStopUSec")
```
simondeziel, line 516:
```suggestion
	output, err := shared.RunCommandContext(ctx, "systemctl", "show", "snap.lxd.daemon.service", "--value", "--property=TimeoutStopUSec")
```

This way, you won't get the unwanted `TimeoutStopUSec=` prefix to trim.

lxd/resources/system.go
```diff
@@ -426,3 +430,99 @@ func systemGetMotherboard() (*api.ResourcesSystemMotherboard, error) {
...
+}
+
+// parseSystemdDuration parses a systemd duration string and returns the equivalent time.Duration.
+func parseSystemdDuration(value string) (time.Duration, error) {
+	if value == "" {
+		return 0, fmt.Errorf("Empty duration string")
+	}
+
+	// Validate the string doesn't have invalid number formats like "1.2.3".
+	invalidNumPattern := regexp.MustCompile(`\d+\.\d+\.\d+`)
```
simondeziel, line 456: `s/MustCompile/Compile/` as the former risks failing if out of memory and thus cause a (unwanted) panic.

lxd/resources/system.go
```diff
@@ -426,3 +430,99 @@ func systemGetMotherboard() (*api.ResourcesSystemMotherboard, error) {
...
 }
+
+// ExtendSystemdTimeout extends the systemd stop timeout by the specified duration.
+func ExtendSystemdTimeout(ctx context.Context, duration time.Duration) error {
+	if os.Getenv("NOTIFY_SOCKET") == "" {
+		return fmt.Errorf("NOTIFY_SOCKET not set while trying to extend systemd stop timeout for LXD daemon")
+	}
+
+	message := fmt.Sprintf("EXTEND_TIMEOUT_USEC=%d", int64(duration/time.Microsecond))
+	_, err := shared.RunCommandContext(ctx, "systemd-notify", message, "READY=1")
```
simondeziel, line 441: The `READY=1` bit is AFAIK, something to send only during startup/initialization, to signal readiness. Is it actually needed when postponing the stop timeout?

lxd/resources/system_test.go
```diff
@@ -0,0 +1,75 @@
+package resources
+
+import (
+	"testing"
+	"time"
+)
+
+func TestParseSystemdDuration(t *testing.T) {
```
simondeziel, line 8: That's a very good use of unit tests :)

gabrielmougard: Thanks ;)

lxd/resources/system.go
gabrielmougard: Thanks. Fixing this.

lxd/resources/system.go
gabrielmougard: Actually, this is not needed in our case. Removing it. Thanks.

gabrielmougard: fixed

lxd/resources/system.go
gabrielmougard: fixed

lxd/resources/system.go
gabrielmougard: fixed

lxd/resources/system.go
```diff
@@ -426,3 +430,98 @@ func systemGetMotherboard() (*api.ResourcesSystemMotherboard, error) {
...
+		"h":   time.Hour,
+		"d":   24 * time.Hour,
+		"w":   7 * 24 * time.Hour,
+		"m":   30 * 24 * time.Hour,
+		"y":   365 * 24 * time.Hour,
+	}
+
+	// Regular expression to match number+unit combinations.
+	// This matches decimal numbers followed by alphabetic units.
+	re := regexp.MustCompile(`(\d+\.?\d*)([a-zA-Z]+)`)
```
simondeziel, line 486: `s/MustCompile/Compile/`

simondeziel (commented):
I didn't find any place where the `READY=1` notification was sent to indicate that the daemon is ready.

simondeziel (changes requested):
As mentioned before, `Type=notify` requires the daemon to send a `READY=1` notification during startup. I couldn't find this anywhere.

lxd/resources/system.go
```diff
@@ -426,3 +430,102 @@ func systemGetMotherboard() (*api.ResourcesSystemMotherboard, error) {
 
 	return &motherboard, nil
 }
+
+// ExtendSystemdTimeout extends the systemd stop timeout by the specified duration.
```
simondeziel, line 434: https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html#EXTEND_TIMEOUT_USEC=%E2%80%A6 indicate that we can use it for more than the stop timeout.

lxd/daemon.go
```diff
@@ -2044,6 +2045,47 @@ func (d *Daemon) Stop(ctx context.Context, sig os.Signal) error {
 	// Cancelling the context will make everyone aware that we're shutting down.
 	d.shutdownCancel()
 
+	// If we're in a snap and we asked for a graceful shutdown,
+	// we start a ticker gorountine that in case of a very long shutdown process
+	// can augment the snapd timeout of the LXD daemon.
```
simondeziel, line 2050: "can extend the systemd stop timeout of the LXD daemon"?

lxd/daemon.go
```diff
@@ -2044,6 +2045,47 @@ func (d *Daemon) Stop(ctx context.Context, sig os.Signal) error {
...
+		stopTimeout, err := resources.GetSystemdStopTimeout(ctx)
+		if err != nil {
+			logger.Error("Failed to get systemd stop timeout", logger.Ctx{"err": err})
+		}
+
+		paddingDuration := 20 * time.Second
+		augmentedDuration := time.Minute
+
+		// If the shutdown process is still not finished after `stopTimeout - paddingDuration`,
+		// we will augment the timeout by `augmentedDuration`. This is to avoid the snapd
```
simondeziel, line 2063: `s/snapd/systemd/`

lxd/daemon.go
```diff
@@ -2044,6 +2045,47 @@ func (d *Daemon) Stop(ctx context.Context, sig os.Signal) error {
...
+				case <-ticker.C:
+					if time.Since(start) > stopTimeout-paddingDuration {
+						logger.Warn("Shutdown process is taking a long time, extending LXD snap daemon stop timeout")
+						err = resources.ExtendSystemdTimeout(ctx, augmentedDuration)
+						if err != nil {
+							logger.Error("Failed to extend systemd stop timeout", logger.Ctx{"err": err})
+						}
+
+						stopTimeout += augmentedDuration
+						logger.Info("LXD snap daemon stop timeout extended", logger.Ctx{"newStopTimeout": stopTimeout})
```
simondeziel, line 2082: The logs are not consistently referring to the "systemd stop timeout".

lxd/resources/system.go
simondeziel: Is the `READY=1` message really required here to bump the stop timeout?

simondeziel (changes requested):
I'm still quite confused at how `READY=1` is handled. This, AFAIK, requires a dedicated message after the daemon is done initializing and this message should be a `READY=1`.

Or maybe I don't really understand `Type=notify`?

gabrielmougard:
> I'm still quite confused at how READY=1 is handled. This, AFAIK, requires a dedicated message after the daemon is done initializing and this message should be a READY=1.

I don't think I understand what you mean here. I thought that passing `READY=1` in the `ExtendSystemdTimeout` was enough. Kind of like how this is done in the `Run` function of the lxd-agent:
```
...
        // Done with early setup, tell systemd to continue boot.
	// Allows a service that needs a file that's generated by the agent to be able to declare After=lxd-agent
	// and know the file will have been created by the time the service is started.
	if os.Getenv("NOTIFY_SOCKET") != "" {
		_, err := shared.RunCommandContext(context.TODO(), "systemd-notify", "READY=1")
		if err != nil {
			cancelStatusNotifier() // Ensure STOPPED status is written to QEMU status ringbuffer.
			cancelFunc()

			return fmt.Errorf("Failed to notify systemd of readiness: %w", err)
		}
	}
```

lxd/daemon.go
gabrielmougard: fixed

lxd/daemon.go
```diff
@@ -1012,11 +1013,21 @@ func (d *Daemon) Init() error {
 	// cleanup any state we produced so far. Errors happening here will be
 	// ignored.
 	if err != nil {
+		_, err2 := shared.RunCommandContext(d.shutdownCtx, "systemd-notify", "ERRNO=125") // 125=ECANCELED code
```
simondeziel, line 1016: I did some more digging and it seems that systemd has generic exit codes too: https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#Process%20Exit%20Codes

I propose simply exiting with `1` to signal a generic failure.

simondeziel (commented):
LGTM with the tiny nit about the failure code.

lxd/resources/system.go
gabrielmougard: fixed

simondeziel (commented):
LGTM conceptually, thanks! I'll let others do a proper review of the Go implementation.

Issue #14839, opened 2025-01-22
Failed parsing stream while listing or starting instances when using `http_proxy` or middle boxes


klanso:
I have searched for this issue but none seems to match. When I try to list or start new instance I get this error:

Error: Failed parsing stream: Get "https://cloud-images.ubuntu.com/releases/streams/v1/index.json": read tcp localip:51604->185.125.190.37:443: read: connection reset by peer

Tried to connect with NC to the url and it works.

 from lxc info:
 os_name: Ubuntu
  os_version: "22.04"
  project: default
  server: lxd
  server_clustered: false
  server_event_mode: full-mesh
  server_name: host1-lo
  server_pid: 1425
  server_version: "6.2"
  server_lts: false
 driver: lxc | qemu
  driver_version: 6.0.2 | 8.2.1
  instance_types:
  - container
  - virtual-machine
  firewall: nftables
  kernel: Linux
  kernel_architecture: x86_64

Thanks for any help.

tomponline:
Is this reproducible or temporary network issue?

malabay:
I have the same error,see attachment

![Image](https://github.com/user-attachments/assets/a1c09b41-9a3b-4cea-b213-ce783d890f76)
if  install  stable version (4 or 5) then Ok, no error

tomponline:
what about 5.21/stable LTS?

malabay:
![Image](https://github.com/user-attachments/assets/0cd4a459-cc38-4437-85ef-05a661a9bf51)

tpsailer:
6.2 also has this issue

tomponline:
Can you do "curl https://cloud-images.ubuntu.com/releases/streams/v1/index.json"

p-ouellette:
I'm getting the same error message with the 5.0/edge snap, but 5.0/candidate works fine:
```
> snap refresh lxd --channel=5.0/candidate
> lxc image info ubuntu:24.04
<ok>
> snap refresh lxd --channel=5.0/edge
> lxc image info ubuntu:24.04
Error: Failed fetching fingerprint "24.04": Failed getting image: Failed parsing stream: Get "https://cloud-images.ubuntu.com/releases/streams/v1/index.json": read tcp [...]: read: connection reset by peer
```
Let me know if there's anything I can do to help debug this.

tomponline:
are you using an HTTP proxy?

p-ouellette:
[...]

PR #15250, opened 2025-03-24
client/connection: use LegacyCurvesOnly for simplestreams connections

simondeziel:
The modern post-quantum curve that has been enabled by default with [Go 1.23+](https://go.dev/doc/go1.23) causes the `ClientHello` message to be much bigger than before due to MLKEM/Kyber needing big `key_share`. This size inflation sometimes causes the `ClientHello` message to be split over multiple (usually 2) TCP packets.

This message splitting can confuse broken middleboxes and lead them to abort the connection (`connection reset by peer`). This problem is well explained [here](https://tldr.fail/).

Those middleboxes are more likely to be intercepting connections to images servers (`images:`, `ubuntu:`, etc) than `lxc` to `lxd` communication. As such, do not use modern post-quantum curves when dealing with `simplestreams` remote.

Addresses #14839.

### DIFF
--- a/client/connection.go
+++ b/client/connection.go
@@ -267,7 +267,10 @@ func ConnectSimpleStreams(url string, args *ConnectionArgs) (ImageServer, error)
 	}
 
 	// Setup the HTTP client
-	httpClient, err := tlsHTTPClient(args.HTTPClient, args.TLSClientCert, args.TLSClientKey, args.TLSCA, args.TLSServerCert, args.InsecureSkipVerify, args.Proxy, args.TransportWrapper)
+	// Do not include modern post-quantum curves in the ClientHello to avoid
+	// compatibility issues (connection resets) with broken middleboxes that
+	// can't handle large ClientHello messages split over multiple TCP packets.
+	httpClient, err := tlsHTTPClient(args.HTTPClient, args.TLSClientCert, args.TLSClientKey, args.TLSCA, args.TLSServerCert, args.InsecureSkipVerify, true, args.Proxy, args.TransportWrapper)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


tomponline (approved):
Thanks!

client/connection.go
```diff
@@ -259,6 +262,11 @@ func ConnectSimpleStreams(url string, args *ConnectionArgs) (ImageServer, error)
 		args = &ConnectionArgs{}
 	}
 
+	// Do not include modern post-quantum curves in the ClientHello to avoid
+	// compatibility issues (connection resets) with broken middleboxes that
+	// can't handle large ClientHello messages split over multiple TCP packets.
+	args.LegacyCurvesOnly =	true
```
tomponline, line 268: Shall we move this to be setting to true in the call site of `ConnectSimpleStreams` rather than overriding something could have been potentially set in `args`?

simondeziel: There are 3 call sites which would cause some comment duplication.

I was debating if this flag should be part of the `args` struct to begin with. If an environment requires this quirk, it needs it all the time so there should be no need to vary it per call site.

My initial version simply passed the right value to the `tlsHTTPClient()` instead. Maybe that would be best?

tomponline: > My initial version simply passed the right value to the tlsHTTPClient() instead. Maybe that would be best?

Sounds good

simondeziel:
PQC is also coming to OpenSSL's next LTS release (3.5): https://github.com/openssl/openssl/releases/tag/openssl-3.5.0-beta1:

* The default TLS supported groups list has been changed to include and
prefer hybrid PQC KEM groups. Some practically unused groups were removed
from the default list.

* The default TLS keyshares have been changed to offer X25519MLKEM768 and
and X25519.

This will align OpenSSL's default TLS keyshares with Go's.

tomponline (approved):

tomponline:
Ta

Issue #15196, opened 2025-03-17
Custom ISO import fails when submitted with flag iso


edlerd:
# Version
LXD 5.21 / snap

# Description
First creating a test file:

`truncate -s 25MiB foo.iso`

Using a storage pool called `remote` with the ceph driver on a cluster with three nodes:

`lxc storage volume  import remote /root/foo.iso foo` succeeds and creates a custom ISO.
 `lxc storage volume  import remote /root/foo.iso foo --type=iso` fails with the below error:

`Error: Failed creating custom volume from ISO: Failed creating volume: write /dev/rbd0: no space left on device`

When running both commands against another storage pool with a local driver, such as directory, both succeed.

Related is the [import of ISO test](https://github.com/canonical/lxd/blob/1a14cfe1c432fd5e8ab88f053c36c706aafc30d4/test/suites/storage_local_volume_handling.sh#L75-L77). The test is doing nearly the right thing, just omitting the `--type=iso` flag.

Some logs from my daemon when issuing the failing command:

```
Mar 17 12:26:01 micro1 systemd[1]: Started snap.lxd.lxc-a82d02f5-28f1-4eea-821c-74914b938db4.scope.
Mar 17 12:26:01 micro1 kernel: libceph: mon2 (1)10.131.25.30:6789 session established
Mar 17 12:26:01 micro1 kernel: libceph: client1534634 fsid 31c511a3-0539-4bcb-829f-b9821f4ed62c
Mar 17 12:26:01 micro1 kernel: rbd: rbd0: capacity 0 features 0x3d
Mar 17 12:26:01 micro1 systemd-udevd[8571]: rbd0: Process '/usr/bin/unshare -m /usr/bin/snap auto-import --mount=/dev/rbd0' failed with exit code 1.
Mar 17 12:26:01 micro1 microceph.daemon[575]: time="2025-03-17T12:26:01Z" level=debug msg="start: monitors unchanged, sleeping: [micro1 micro2 micro3]"
Mar 17 12:26:02 micro1 systemd[1]: snap.lxd.lxc-a82d02f5-28f1-4eea-821c-74914b938db4.scope: Deactivated successfully.
```

Expected behavior is that, also with the `--type=iso` flag, the import succeeds.

tomponline:
@edlerd is this specific to the ceph storage driver or does it occur for local storage drivers too?

edlerd:
[...]

PR #15251, opened 2025-03-24
Storage: Fix import of iso images smaller than 512B on ceph

skozina:
Ceph storage has a limitation that if the size of the image is less than 512B, the image is not writable.
Fixes: https://github.com/canonical/lxd/issues/15196

### DIFF
--- a/lxd/storage/drivers/driver_ceph_utils.go
+++ b/lxd/storage/drivers/driver_ceph_utils.go
@@ -101,6 +101,15 @@ func (d *ceph) osdDeletePool() error {
 	return nil
 }
 
+func (d *ceph) roundUpTo512(a int64) int64 {
+	result := (a / 512) * 512
+	if a%512 != 0 {
+		result += 512
+	}
+
+	return result
+}
+
 // rbdCreateVolume creates an RBD storage volume.
 // Note that the default set of features is intentionally limited
 // by passing --image-feature explicitly. This is done to ensure that
@@ -130,6 +139,9 @@ func (d *ceph) rbdCreateVolume(vol Volume, size string) error {
 	if d.config["ceph.osd.data_pool_name"] != "" {
 		cmd = append(cmd, "--data-pool", d.config["ceph.osd.data_pool_name"])
 	}
+
+	// Ceph allows writing only to images of size in multiples of 512B
+	sizeBytes = d.roundUpTo512(sizeBytes)
 
 	cmd = append(cmd,
 		"--size", fmt.Sprintf("%dB", sizeBytes),
--- a/test/suites/storage_volume_import.sh
+++ b/test/suites/storage_volume_import.sh
@@ -1,6 +1,9 @@
 test_storage_volume_import() {
-  truncate -s 25MiB foo.iso
-  truncate -s 25MiB foo.img
+  # Some storage types, such as lvm and ceph, support only 512-aligned image sizes
+  # Test with weird sizes of images
+  dd if=/dev/urandom of=foo.iso bs=359 count=1
+  dd if=/dev/urandom of=foo.img bs=1M count=1
+  echo -n "a" >> foo.img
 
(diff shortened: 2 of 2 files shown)


tomponline:
@skozina please can you rebase to get the fixes for the linter. Ta

lxd/storage/drivers/driver_ceph_utils.go
```diff
@@ -131,6 +131,10 @@ func (d *ceph) rbdCreateVolume(vol Volume, size string) error {
 		cmd = append(cmd, "--data-pool", d.config["ceph.osd.data_pool_name"])
 	}
 
+	// Ceph seems to have a limitation that if an image is smaller than 512B, it is not writable.
+	// Make sure we create an image of a size at least 512B.
+	sizeBytes = max(sizeBytes, 512)
```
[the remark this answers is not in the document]
skozina: Looks like you're right. I just tried to import an image of size 513B, and that fails as well.
I tried even larger files (20M), which work fine. Once I append a single letter by the end, it fails.
I'll update the PR to round up to 512, similarly like we do for LVM volumes. Thank you!

tomponline:
@skozina thanks! Can we update the tests in this repo to check for this to avoid regressions in the future?


Perhaps rather than using truncate here https://github.com/canonical/lxd/blob/578c23cc99b6dda6d4d679594556c98dae5f20c0/test/suites/storage_volume_import.sh#L2-L3 we use contents from /dev/urandom for a specific size (suggest < 512 bytes).

skozina:
> @skozina thanks! Can we update the tests in this repo to check for this to avoid regressions in the future?

Sure! Added. Thanks for the review!

test/suites/storage_volume_import.sh
```diff
@@ -1,6 +1,8 @@
 test_storage_volume_import() {
-  truncate -s 25MiB foo.iso
-  truncate -s 25MiB foo.img
+  # Some storage types, such as lvm and ceph, support only 512-aligned image sizes
+  # Test with weird sizes of images
+  dd if=/dev/urandom of=foo.iso bs=1 count=359
+  dd if=/dev/urandom of=foo.img bs=1024 count=1024 && echo -n "a" >> foo.img
```
simondeziel, line 5: Tiny nit to avoid multiple small IOs:

```suggestion
  dd if=/dev/urandom of=foo.iso bs=359 count=1
  dd if=/dev/urandom of=foo.img bs=1M count=1
  echo -n "a" >> foo.img
```

test/suites/storage_volume_import.sh
```diff
@@ -1,6 +1,9 @@
 test_storage_volume_import() {
-  truncate -s 25MiB foo.iso
-  truncate -s 25MiB foo.img
+  # Some storage types, such as lvm and ceph, support only 512-aligned image sizes
+  # Test with weird sizes of images
+  dd if=/dev/urandom of=foo.iso bs=359 count=1
+  dd if=/dev/urandom of=foo.img bs=1M count=1
+  echo -n "a" >> foo.img
```
tomponline, line 6: do we need this line anymore?

simondeziel: I'd say yes as it's the one pushing it just over the limit. 1M is a nice round number but adding one byte forces the rounding up to 1M + 512B

tomponline: ah ok i see

tomponline:
@skozina seeing as that ceph test failure is the volume export one weve been chatting about, do u fancy taking a stab at it as part of this pr too?

tomponline:
@skozina im afraid the code tests are now failing because of the upgraded linter:

https://github.com/canonical/lxd/actions/runs/14064452040/job/39577612484?pr=15251

skozina:
Rebased on top of latest `main` which has the linter fixes applied.

PR #15270, opened 2025-03-28
doc: Fix swagger definition for `POST /1.0/auth/identities/tls`

markylaing:
The swagger comment was incorrectly referencing the struct type name.


tomponline (approved):
ta

PR #15274, opened 2025-03-29
doc: improve network ACLs docs

minaelee:
Improvements to network ACLs docs, including:
- Document the ability to provide a YAML file to create an ACL
- Clarify that the only arguments accepted for CLI-based ACL creation (aside from a YAML file) are the ACL name and the custom user keys
- Add missing sections for listing ACLs, showing an ACL, renaming an ACL, deleting an ACL
- Add API instruction counterparts to all CLI instructions
- Add links to relevant parts of API reference
- Add command syntax examples
- Improve descriptions of ACL and ACL rule properties
- Detailed information about how to edit the ACL via the API, where and how to use PATCH and POST
- Clarify how logging works
- General formatting improvements and copyedits

### DIFF
--- a/doc/howto/network_acls.md
+++ b/doc/howto/network_acls.md
@@ -836,7 +836,7 @@ lxc config device set ubuntu-container ovn-nic security.acls="web-traffic,intern
 
 #### View the existing NIC configuration
 
-To update the configuration for an instance's device using the API, you must include all the required keys for the device—even if you're only changing one key. The required keys are the `type` of `nic` and the `network` name. Include these along with the key to update, and any other existing keys (unless you want to remove them). Omitted keys are reset to default values.
+To update the configuration for an instance's device using the API, you must include all the required keys for the device—even if you're only changing one key. For an instance NIC device, the required keys are the `type` of `nic` and the `network` name. Include these along with the key to update, and any other existing keys (unless you want to remove them). Omitted keys are reset to default values.
 
 To view the existing instance NIC configuration, query the following endpoint:
 
@@ -919,23 +919,138 @@ To view additional properties of the `security.acls` lists, refer to the configu
 (network-acls-defaults)=
 ## Configure default actions
 
-When one or more ACLs are applied to a NIC (either explicitly or implicitly through a network), a default reject rule is added to the NIC.
+When one or more ACLs are applied to a NIC—either directly or through its network—a default reject rule is added to the NIC.
 This rule rejects all traffic that doesn't match any of the rules in the applied ACLs.
 
-You can change this behavior with the network and NIC level `security.acls.default.ingress.action` and `security.acls.default.egress.action` settings.
-The NIC level settings override the network level settings.
+You can change this behavior with the network- and NIC-level `security.acls.default.ingress.action` and `security.acls.default.egress.action` settings. The NIC-level settings override the network-level settings.
 
-For example, to set the default action for inbound traffic to `allow` for all instances connected to a network, use the following command:
+`````{tabs}
+````{group-tab} CLI
 
-```bash
-lxc network set <network_name> security.acls.default.ingress.action=allow
-```
+### Configure a default action for a network
 
-To configure the same default action for an instance NIC, use the following command:
+To set the default action for a network's egress or ingress traffic, run:
 
 ```bash
-lxc config device set <instance_name> <device_name> security.acls.default.ingress.action=allow
+lxc network set <network_name> security.acls.default.<egress|ingress>.action=<allow|reject|drop>
 ```
+
+#### Example
+
+To set the default action for inbound traffic to `allow` for all instances on the `ovntest` network, run:
+
+```bash
+lxc network set ovntest security.acls.default.ingress.action=allow
+```
+
+### Configure a default action for an instance NIC device
+
+To set the default action for an instance NIC's egress or ingress traffic, run:
+
+```bash
+lxc config device set <instance_name> <NIC_name> security.acls.default.<egress|ingress>.action=<allow|reject|drop>
+```
+
+#### Example
+
+To set the default action for inbound traffic to `allow` for the `ubuntu-container` instance's `ovn-nic` NIC device, run:
+
+```bash
+lxc config device set ubuntu-container ovn-nic security.acls.default.ingress.action=allow
+```
+
+````
+% End of group-tab CLI
+
+````{group-tab} API
+
+### Configure a default action for a network
+
+To set the default action for a network's egress or ingress traffic, query the following endpoint:
+
+```bash
+lxc query --request PATCH /1.0/networks/{networkName} --data '{
+  "config": {
+    "security.acls.default.egress.action": "<allow|reject|drop>",
+    "security.acls.default.ingress.action": "<allow|reject|drop>",
+  }
+}'
+```
+
+See [the API reference](swagger:/networks/network_patch) for more information.
+
+#### Example
+
+Set the `ovntest` network's default egress action to `allow`:
+
+```bash
+lxc query --request PATCH /1.0/networks/ovntest --data '{
+  "config": {
+    "security.acls.default.egress.action": "allow"
+  }
+}'
+```
+
+### Configure a default action for an instance NIC device
+
+#### View the existing NIC configuration
+
+To update the configuration for an instance's device using the API, you must include all the required keys for the device—even if you're only changing one key. For an instance NIC device, the required keys are the `type` of `nic` and the `network` name. Include these along with the key to update, and any other existing keys (unless you want to remove them). Omitted keys are reset to default values.
+
+To view the existing instance NIC configuration, query the following endpoint:
+
+```bash
+lxc query /1.0/instances/{instanceName} | jq '.devices["<NIC name>"]'
+```
+
+See [the API reference](swagger:/networks/network_get) for more information.
+
+##### Example
+
+```bash
+lxc query /1.0/instances/ubuntu-container | jq '.devices["ovn-nic"]'
+```
+
+#### Configure the default action
+
+To set the default action for an instance NIC's traffic, query the following endpoint:
+
+```bash
+lxc query --request PATCH /1.0/instances/{instanceName} --data '{
+  "devices": {
+    "<NIC name>": {
+      "network": <network_name>,
+      "type": "nic",
+      "security.acls.default.<egress|ingress>.action": "<allow|reject|drop>"
+      <other optional keys>
+    }
+  }
+}'
+```
+
+See [the API reference](swagger:/instances/instance_patch) for more information.
+
+#### Example
+
+This request sets the default action for inbound traffic to `allow` for the `ovn-nic` device of the `ubuntu-container` instance:
+
+```bash
+lxc query --request PATCH /1.0/instances/ubuntu-container --data '{
+  "devices": {
+    "ovn-nic": {
+      "network": "ovntest",
+      "type": "nic",
+      "security.acls.default.ingress.action": "allow"
+    }
+  }
+}'
+```
+
+````
+% End of group-tab API
+
+`````
+% End of tab
 
 (network-acls-bridge-limitations)=
 ## Bridge limitations
--- a/doc/howto/network_acls.md
+++ b/doc/howto/network_acls.md
@@ -740,6 +740,36 @@ Note that the request _inserted_ the new `user.limit` key without affecting the 
 
 `````
 
+(network-acls-delete)=
+## Delete an ACL
+
+You can only delete an ACL that is not assigned to a NIC or network. See the {ref}`network-acls-assign` section for more information on ACL assignment.
+
+`````{tabs}
+````{group-tab} CLI
+
+To delete an ACL, run:
+
+```bash
+lxc network acl delete <ACL_name>
+```
+
+````
+% End of group-tab CLI
+
+````{group-tab} API
+
+```bash
+lxc query --request DELETE /1.0/network-acls/{name}
+```
+
+See [the API reference](swagger:/network-acls/network_acl_delete) for more information.
+
+````
+% End of group-tab API
+
+`````
+
 (network-acls-assign)=
 ## Assign an ACL
 
--- a/doc/howto/network_acls.md
+++ b/doc/howto/network_acls.md
@@ -107,7 +107,7 @@ Network ACL names must meet the following requirements:
 To create an ACL, run:
 
 ```bash
-lxc network acl create <ACL_name> [user.KEY=value ...] 
+lxc network acl create <ACL_name> [user.KEY=value ...]
 ```
 
 - You must provide an ACL name that meets the requirements described above.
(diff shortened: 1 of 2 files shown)


doc/howto/network_acls.md
```diff
@@ -92,22 +92,162 @@ lxc query --request GET /1.0/network-acls/my-acl
...
+First, create a file named `config.yaml` with the following content:
+
+```yaml
+description: Allow web traffic from internal network
+config:
+  user.owner: devops
+egress:
+  - action: allow
+    description: Allow DNS to Google
+    destination: 8.8.8.8/32,8.8.4.4/32
```
tomponline, line 155: we tend to use the documentation IP ranges in examples.

https://en.wikipedia.org/wiki/Reserved_IP_addresses

doc/howto/network_acls.md
```diff
@@ -621,21 +621,273 @@ Note that the request _inserted_ the new `user.limit` key without affecting the
...
+lxc query --request PATCH /1.0/networks/{networkName} --data '{
+  "config": {
+    "security.acls": "<ACL_name>[,<ACL_name>,...]"
+  }
+}'
+```
+
+#### Examples
+
+Set the `br01` network's `security.acls` to contain only the `web-traffic` ACL:
```
tomponline, line 854: "br" prefix by convention means unmanaged bridges (which cant have ACLs assigned), so we should avoid implicitly indicating that here. Instead suggest using `lxdbr0` here.

doc/howto/network_acls.md
```diff
@@ -891,24 +891,133 @@ To view additional properties of the `security.acls` lists, refer to the configu
 (network-acls-defaults)=
 ## Configure default actions
 
-When one or more ACLs are applied to a NIC (either explicitly or implicitly through a network), a default reject rule is added to the NIC.
+When one or more ACLs are applied to a NIC—either directly or through its network—a default reject rule is added to the NIC.
```
tomponline: shouldnt there be spaces between the "-" here?

doc/howto/network_acls.md
```diff
@@ -891,24 +891,133 @@ To view additional properties of the `security.acls` lists, refer to the configu
...
 ```
 
-To configure the same default action for an instance NIC, use the following command:
+### Configure a default action for an instance NIC device
+
+To set the default action for an instance NIC's egress or ingress traffic, run:
 
 ```bash
-lxc config device set <instance_name> <device_name> security.acls.default.ingress.action=allow
+lxc config device set <instance_name> <NIC_name> security.acls.default.<egress|ingress>.action=<allow|reject|drop>
```
tomponline, line 1020: Would be worth going through the rest of the docs for examples where we've used `device_name` in `lxc [profile] config device [add|set|unset|remove]` commands in the NIC context and updating it to this form for consistency.

doc/howto/network_acls.md
```diff
@@ -891,24 +891,133 @@ To view additional properties of the `security.acls` lists, refer to the configu
...
 ```
 
+#### Example
+
+To set the default action for inbound traffic to `allow` for all instances on the `ovntest` network, run:
+
+```bash
+lxc config device set ovntest security.acls.default.ingress.action=allow
```
tomponline, line 1028: this is missing either an instance name or a device name argument - its an invalid command.

doc/howto/network_acls.md
minaelee: These are em-dashes which render longer than hyphens and [do not typically have spaces around them except in newspapers](https://prowritingaid.com/grammar/1000066/Should-I-use-spaces-around-an-em-dash). Rendered view:
image

doc/howto/network_acls.md
minaelee: Changed all to `ovn1`, thanks!

doc/howto/network_acls.md
minaelee: Fixed, thanks!

doc/howto/network_acls.md
minaelee: Agreed - added as docs backlog item (https://warthogs.atlassian.net/browse/LXD-2854)

doc/howto/network_acls.md
minaelee: Removed the egress section as it didn't make sense in the context of the entire ACL description (allow web traffic from internal network) anyway.

minaelee:
Marking as draft because I realized I need to add a section for removing ACLs (not just ACL rules).

edlerd:
I just stumbled over [ACL assignment](https://documentation.ubuntu.com/lxd/en/latest/howto/network_acls/#assign-an-acl) missing some limitations:

When assigning ACLs to an instance nic, this only works if the network on the instance is of type `OVN`. For any other nictype, this is not allowed. I think the assignment section should mention that limitation of network type. See also the [available fields per nictype](https://documentation.ubuntu.com/lxd/en/latest/reference/devices_nic/#device-nic-ovn-device-conf:security.acls).

This limitation does not apply, when assigning an ACL directly to a network.

minaelee:
> I just stumbled over [ACL assignment](https://documentation.ubuntu.com/lxd/en/latest/howto/network_acls/#assign-an-acl) missing some limitations:
> 
> When assigning ACLs to an instance nic, this only works if the network on the instance is of type `OVN`. For any other nictype, this is not allowed. I think the assignment section should mention that limitation of network type. See also the [available fields per nictype](https://documentation.ubuntu.com/lxd/en/latest/reference/devices_nic/#device-nic-ovn-device-conf:security.acls).
> 
> This limitation does not apply, when assigning an ACL directly to a network.

The [top of the Assign an ACL section in the update](https://canonical-ubuntu-documentation-library--15274.com.readthedocs.build/lxd/en/15274/howto/network_acls/#assign-an-acl) mentions this:

image

There's a note at the very top of the document as well:

image

Unless I misunderstand, and you mean something different?

doc/howto/network_acls.md
```diff
@@ -20,6 +20,76 @@ When assigned to a network, the ACL applies to all NICs connected to the network
...
+
+## Show an ACL
+
+`````{tabs}
+````{group-tab} CLI
+
+To show details about a specific ACL, run:
+
+```
+lxc network acl show <acl_name>
```
tomponline, line 64: @minaelee there's some inconsistency with the placeholder formatting on this commit and in the next commit for creating ACLs, here you use `acl_name` and on the next commit you use `ACL_name`.

I'm fine with either, but we should use consistent placeholder names to avoid any implication that case matters.

doc/howto/network_acls.md
```diff
@@ -92,22 +92,155 @@ lxc query --request GET /1.0/network-acls/my-acl
...
+To create an ACL, run:
+
+```bash
+lxc network acl create <ACL_name> [user.KEY=value ...] 
+```
+
+- You must provide an ACL name that meets the requirements described above.
+- You can optionally provide one or more custom `user` keys to store metadata or other information.
+
+ACLs have no rules upon creation via command line, so as a next step, {ref}`add rules <network-acls-rules>` to the ACL. You can also {ref}`edit the ACL configuration <network-acls-edit>`, or {ref}`assign the ACL to a network or NIC <network-acls-assign>`.
```
tomponline, line 116: You could mention here that an ACL with no rules defaults to blocking all traffic.

doc/howto/network_acls.md
```diff
@@ -92,22 +92,155 @@ lxc query --request GET /1.0/network-acls/my-acl
...
+
+Another way to create ACLs from the command line is to provide a YAML configuration file:
+
+```bash
+lxc network acl create <ACL_NAME> < <filename.yaml>
+```
+
+This file must contain at least the ACL `name`, and it can include any other {ref}`network-acls-properties`, including the `egress` and `ingress` properties for defining {ref}`ACL rules <network-acls-rules>`. See the final example in the set below.
```
tomponline, line 124: > This file must contain at least the ACL `name`

This isnt true, the `name` would be ignored as its taken from the CLI argument.

doc/howto/network_acls.md
```diff
@@ -92,22 +92,155 @@ lxc query --request GET /1.0/network-acls/my-acl
...
+
+### Examples
+
+Create an ACL with the name `acl1`:
+
+```bash
+lxc network acl create acl1
```
tomponline, line 131: in the earlier example for showing an ACL you use `my-acl` but here you're using `acl1`, it'd be nice to use consistent example names.

doc/howto/network_acls.md
```diff
@@ -92,22 +92,155 @@ lxc query --request GET /1.0/network-acls/my-acl
...
+```
+
+Create an ACL with the name `acl2` and a custom user key of `environment`:
 
 ```bash
-lxc network acl create <ACL_name> [configuration_options...]
+lxc network acl create acl2 user.environment=dev
```
tomponline, line 137: as the example here isn't related to the example above, we could use the same ACL name `my-acl` again here.

doc/howto/network_acls.md
```diff
@@ -92,22 +92,155 @@ lxc query --request GET /1.0/network-acls/my-acl
...
 ```
 
-This command creates an ACL without rules.
-As a next step, {ref}`add rules <network-acls-rules>` to the ACL.
+Create an ACL with the name `acl3`, and custom user keys of `environment` and `level`:
 
-Valid network ACL names must adhere to the following rules:
+```bash
+lxc network acl create acl3 user.environment=dev user.level=5
```
tomponline, line 143: Same here, we can use `my-acl` as its a different example.

doc/howto/network_acls.md
```diff
@@ -92,22 +92,155 @@ lxc query --request GET /1.0/network-acls/my-acl
...
+Create an ACL with the name `acl2` and a custom user key of `environment`:
 
 ```bash
-lxc network acl create <ACL_name> [configuration_options...]
+lxc network acl create acl2 user.environment=dev
 ```
 
-This command creates an ACL without rules.
-As a next step, {ref}`add rules <network-acls-rules>` to the ACL.
+Create an ACL with the name `acl3`, and custom user keys of `environment` and `level`:
```
tomponline, line 140: I feel like having multiple examples of custom user keys is perhaps labouring the point. 
Having one example showing custom user keys is probably sufficient, especially given that the example user keys being used here start to give the impression that these keys mean something to LXD.

tomponline: The ACL and the each rule inside the ACL support a description property, so also showing an example of custom user keys for environment and level seem redundant or imply LXD understands those in my view.

doc/howto/network_acls.md
```diff
@@ -359,14 +361,245 @@ lxc network acl show-log <ACL_name>
...
+
+Requirements:
+
+- You can only rename an ACL that is not currently assigned to a NIC or network. See the {ref}`network-acls-assign` section for more information.
+- The new name must meet the naming requirements listed in {ref}`network-acls-create`.
+
+To rename an ACL, query the following endpoint:
+
+```bash
+lxc query --request POST /1.0/network-acls/{name} --data '{
```
tomponline, line 557: We should use consistent placeholder names, so rather than `name` here, suggest `ACL_NAME` or `acl_name`, which ever we also use in the CLI examples.

doc/howto/network_acls.md
```diff
@@ -359,14 +361,245 @@ lxc network acl show-log <ACL_name>
...
+
+```{warning}
+Be careful! Any properties you omit from this request (aside from the ACL `name`) will be reset to defaults.
+```
+
+The `PUT` method used in this request performs a full replacement of the ACL's properties. With the exception of the `name` property, all other properties are overwritten by the data you provide. Any omitted properties are reset to default values. To preserve any specific properties when making an update, first {ref}`retrieve the current ACL properties <network-acls-show>`, then copy the values you want to keep into your request body. See [the API reference](swagger:/network-acls/network_acl_put) for more information.
+
+If you only need to add or update a `config.user.*` key, see: {ref}`network-acls-edit-api-config`.
+
+#### Example
```
tomponline, line 598: @minaelee can we avoid the need to repeat basic REST API fundamentals as an example for every API entity, and instead have an "API Basics" tutorial that we can link to from the warning?

Its starting to feel like these examples are being unnecessarily bloated and will be very repetitive across every API entity type we have. Can we cover the fundamental concept of "PUT replaces the entity's config, so you need to GET first and merge if you want to update" in a single place?

doc/howto/network_acls.md
```diff
@@ -359,14 +361,245 @@ lxc network acl show-log <ACL_name>
...
+Consider an ACL named `test` with the following properties (shown in JSON):
+
+```json
+{
+  "name": "test",
+  "description": "My test ACL",
+  "config": {
+    "user.type": "test"
+  },
+  "description": ""
```
tomponline, line 708: there seems to be a rogue description line here

doc/howto/network_acls.md
```diff
@@ -348,16 +348,51 @@ Otherwise, the ACL cannot be applied to it.
...
+lxc query --request GET /1.0/network-acls/my-logged-acl/log
+```
+
+````
+% End of API group-tab
+`````
+% End of tabs
+
+```{note}
+LXD does not validate whether the specified ACL contains any rules with a `state` of `logged`. As a result, if your attempt to view logs returns no data, it could due to one of the following:
```
tomponline, line 494: we should probably explain why this is here, its because the ACL may have formerly had rules with log action in them and currently do not, but the old logs can still be view.

doc/howto/network_acls.md
tomponline: Ah you covered this here https://github.com/canonical/lxd/pull/15274/commits/04ade3398bd8407350cf4890649d532d8da77a15

doc/howto/network_acls.md
```diff
@@ -639,21 +639,283 @@ Note that the request _inserted_ the new `user.limit` key without affecting the
...
+```
+
+````
+% End of group-tab CLI
+
+````{group-tab} API
+
+#### View the existing NIC configuration
+
+To update the configuration for an instance's device using the API, you must include all the required keys for the device—even if you're only changing one key. The required keys are the `type` of `nic` and the `network` name. Include these along with the key to update, and any other existing keys (unless you want to remove them). Omitted keys are reset to default values.
```
tomponline: This is starting to feel repetitive again. Especially as here we are crossing the boundary into the Instance Device realm. Can we document which config key needs updating for instance devices and then link to an instance device howto for updating devices which would have this warning?

tomponline: We dont want to start documenting instance device required keys outside of the instance or instance device pages, otherwise we are repeating ourselves.

doc/howto/network_acls.md
```diff
@@ -836,7 +836,7 @@ lxc config device set ubuntu-container ovn-nic security.acls="web-traffic,intern
 
 #### View the existing NIC configuration
 
-To update the configuration for an instance's device using the API, you must include all the required keys for the device—even if you're only changing one key. The required keys are the `type` of `nic` and the `network` name. Include these along with the key to update, and any other existing keys (unless you want to remove them). Omitted keys are reset to default values.
+To update the configuration for an instance's device using the API, you must include all the required keys for the device—even if you're only changing one key. For an instance NIC device, the required keys are the `type` of `nic` and the `network` name. Include these along with the key to update, and any other existing keys (unless you want to remove them). Omitted keys are reset to default values.
```
tomponline, line 963: Is this section modified twice within the PR? We should avoid this an ensure each commit in the PR is the final form for review.

doc/howto/network_acls.md
tomponline: There seems to be various conventions on this one. Personally I found this quite jarring to read, I thought it was a formatting mistake to start with (missed spaces).

tomponline: My other concern here is we'll end up with 2 different standards across the docs based on who authored each document. If this is the new required standard then it should be applied across the docs consistently, but if its being left down to personal preference of each author I would prefer us to stick with what is the more common convention across the docs as it is now.

doc/howto/network_acls.md
```diff
@@ -919,24 +919,139 @@ To view additional properties of the `security.acls` lists, refer to the configu
...
+```bash
+lxc config device set <instance_name> <NIC_name> security.acls.default.<egress|ingress>.action=<allow|reject|drop>
+```
+
+#### Example
+
+To set the default action for inbound traffic to `allow` for the `ubuntu-container` instance's `ovn-nic` NIC device, run:
+
+```bash
+lxc config device set ubuntu-container ovn-nic security.acls.default.ingress.action=allow
```
tomponline, line 1082: The `ubuntu-container` instance name here got me thinking.

I don't think we should be using the instance type in the instance name for examples, unless we want to draw attention to something that is instance type specific.

Otherwise this could be misconstrued as being something specific to containers, and some users are only interested in VMs and are actively avoiding containers. 

Instead I think we should be coming up with a standard set of instance names we use for examples.
In the past I've seen us use things like "c1" and "v1", which suffer, albeit less obviously, from the same problem of embedding instance type in the name.

doc/howto/network_acls.md
```diff
@@ -27,7 +27,7 @@ See {ref}`network-acls-groups` for more information.
 
 To list all ACLs, run:
 
-```
+```bash
```
tomponline, line 30: When we talk about each logical change being a commit, these changes would be a good example of a separate commit. Something like "Updated code blocks in ACLs section to use bash highlighting" along with an explanation as to why.

Then the actual content changes can be a separate commit explaining what has been changed.

Otherwise we are ending up with large commits that change a variety of different things in the docs.

tomponline: I would also encourage you to split up formatting changes from content changes, the former needing less focused review than the latter. If the two are mixed in together it is harder to know which bits to focus on.

doc/howto/network_acls.md
```diff
@@ -739,16 +840,16 @@ lxc network set <network_name> security.acls="<ACL_name>[,<ACL_NAME>,...]"
 
 #### Examples
 
-Set the `br01` network's `security.acls` to contain only the `web-traffic` ACL:
+Set the `ovn1` network's `security.acls` to contain only the `web-traffic` ACL:
```
tomponline, line 873: as this section is network type agnostic (the header says "bridge or OVN network"),  how about we just use the network name "my-network" here?

tomponline (changes requested):
Thanks @minaelee I was able to get time to do a thorough review on this today.

Some areas of improvement:

1. Probably the most concerning one for me is the pattern I'm now starting to see where the examples in the how-to are covering our API fundamentals (GET/PUT/POST/PATCH behaviour) over and over again for each entity type. This is understandable because you've been focusing on improving the docs in specific areas. But if we continue this approach to its logical conclusion I am worried we're going to end up with a lot of repetition, not only on the topic of REST API fundamentals, but also cross-entity type updates (instance devices for example). I think we consider whether, as part of focusing on a particular of the docs, that if a a generic or cross-entity type theme emerges then we should also expand the scope to add these generic sections and then link to them from the focus area docs. What do you think?
2. There's some minor inconsistency in CLI and API placeholder naming (`acl_name` vs `ACL_NAME` vs `name` for example) we should settle on a form and use it throughout.
3. Entity naming in the examples. This is a historical problem, but as you're re-working areas of the docs now it would seem to be a good time to start addressing it. I think that where an instance type or network type is not relevant for the example we should avoid embedding that type into the entity name (e.g. `ubuntu-container` instance name or `ovn1` network name), but rather pick friendly generic names such as "my-instance" or "my-network" to allow easy identification of which part of the command is user specified, and to avoid any incorrect implication that the examples are restricted to a specific type of instance/network.

edlerd:
When commenting initially, I was only referring to the current live docs, not on the PR contents.

> There's a note at the very top of the document:

I missed that, didn't read the whole doc at the time, just the examples.

I think the current PR is clarifying on my initial point. Because the limitation is mentioned right next to the examples. Thanks for that :)

edlerd:
Another note: Based on a recent request, we are adding network ACLs to the UI shortly. This is the relevant PR: https://github.com/canonical/lxd-ui/pull/1171

doc/howto/network_acls.md
minaelee: I've standardized (with `ACL_name`).

doc/howto/network_acls.md
minaelee: Right! Fixed.

doc/howto/network_acls.md
minaelee: > I feel like having multiple examples of custom user keys is perhaps labouring the point. Having one example showing custom user keys is probably sufficient, especially given that the example user keys being used here start to give the impression that these keys mean something to LXD.

I wanted to show how you _would_ add multiple custom user keys during creation, as it's possible but the syntax is not intuitive. If you think it's not important and places too much emphasis on custom user keys, I see that point and can remove it.

minaelee: > The ACL and the each rule inside the ACL support a description property, so also showing an example of custom user keys for environment and level seem redundant or imply LXD understands those in my view.

What would you suggest as an example of a custom user key instead?

doc/howto/network_acls.md
```diff
@@ -92,22 +92,155 @@ lxc query --request GET /1.0/network-acls/my-acl
...
+  "config": {
+    "user.<custom key name>": "<custom key value>"
+  },
+  "description": "<description of the ACL>",
+  "egress": [<egress rule>, <another egress rule...>],
+  "ingress": [<ingress rule>, <another ingress rule...>]
+}'
+```
+
+- You must provide an ACL `name` that meets the requirements described above.
```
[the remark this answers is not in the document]
minaelee: Done

doc/howto/network_acls.md
minaelee: Agreed, changed to `my-instance`

doc/howto/network_acls.md
minaelee: Added an explanation

minaelee:
> Thanks @minaelee I was able to get time to do a thorough review on this today.
> 
> Some areas of improvement:
> 
> 1. Probably the most concerning one for me is the pattern I'm now starting to see where the examples in the how-to are covering our API fundamentals (GET/PUT/POST/PATCH behaviour) over and over again for each entity type. This is understandable because you've been focusing on improving the docs in specific areas. But if we continue this approach to its logical conclusion I am worried we're going to end up with a lot of repetition, not only on the topic of REST API fundamentals, but also cross-entity type updates (instance devices for example). I think we consider whether, as part of focusing on a particular of the docs, that if a a generic or cross-entity type theme emerges then we should also expand the scope to add these generic sections and then link to them from the focus area docs. What do you think?

Agreed, we can extract these out. 

> 2. There's some minor inconsistency in CLI and API placeholder naming (`acl_name` vs `ACL_NAME` vs `name` for example) we should settle on a form and use it throughout.

ACL-name makes sense to me since you can't use underscores in ACL names.

> 3. Entity naming in the examples. This is a historical problem, but as you're re-working areas of the docs now it would seem to be a good time to start addressing it. I think that where an instance type or network type is not relevant for the example we should avoid embedding that type into the entity name (e.g. `ubuntu-container` instance name or `ovn1` network name), but rather pick friendly generic names such as "my-instance" or "my-network" to allow easy identification of which part of the command is user specified, and to avoid any incorrect implication that the examples are restricted to a specific type of instance/network.

Agreed, updating to my-instance and my-network.

> 4. Commit structure. You're already breaking commits down by general section, which is much appreciated. But for these larger reworks some of the commits were pretty large. I found that where formatting changes were mixed with content changes it made carefully reviewing the content changes more difficult. It would be great if document wide mechanical formatting changes that dont change the actual content could be separated into different commits.
> 
> Thanks!

I'll aim to keep the document formatting scoped to the specific section's commit rather than document-wide.

tomponline:
> ACL-name makes sense to me since you can't use underscores in ACL names.
> 
> >

The name requirements for an entity are not relevant here as these are just placeholders, they are not to be run.

In the examples that show actual runnable commands, we can use "my-<entity>", e.g "my-acl" for example.

doc/howto/network_acls.md
tomponline: > I wanted to show how you _would_ add multiple custom user keys during creation, as it's possible but the syntax is not intuitive. If you think it's not important and places too much emphasis on custom user keys, I see that point and can remove it.

mmm yes I see, I suppose again this is applicable to many LXD entity types, so perhaps we have have a shared section on specifying config at create time, and include there a section on what custom user keys can be useful for.

My perspective is that they are quite a niche feature that I'd prefer to not have them enlarging the examples for every entity type.

Maybe we can add a one liner that says something long the lines of "This entity allows specifying config at create time, including [custom user fields](link to section explaining custom user fieds)"

minaelee:
Re-submitted with more modular commits and other issues from PR review addressed in https://github.com/canonical/lxd/pull/15362

PR #15277, opened 2025-03-31
Configurable instance placement

markylaing:
Adds the ability to define placement rules on profiles and instances. Placement rules can be added with:
```bash
lxc {config,profile} placement add <rule-name> <rule-kind> <entity-type> <property>=<<value>,...> [--priority=<priority>]
```
For example:
```bash
lxc profile placement add my-rule affinity cluster_group name=gpu
```
Creates a required affinity rule for the cluster group "gpu". All instances with this profile will be placed on cluster members that are part of the "gpu" cluster group.

This PR shows the basic principle of the implementation but there are many more things to do:
- [ ] Rule validation in relation to project configuration (e.g. `restricted.cluster.groups`).
- [ ] Disallow rule creation if server is not clustered.
- [ ] Lots more testing.
- [ ] Concurrent scheduling.
- [ ] Further generalisation of rule validation via entity type metadata.
- [ ] Expose more entity type metadata to client, to allow auto-completion and client side validation.
- [ ] Documentation.
- [ ] Rule evaluation of existing instances on rule change.

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2638,3 +2638,9 @@ Adds new {config:option}`instance-security:security.delegate_bpf`.* group of opt
 ## `override_snapshot_profiles_on_copy`
 
 This adds a request option to set snapshot's target profile on instance copy to be inherited from target instance.
+
+## `instance_placement_rules`
+
+This API extension adds a new `placement_rules` field to instances and profiles.
+Placement rules are used to configure where an instance will be created in the cluster.
+Multiple placement rules can be used to define complex scheduling scenarios.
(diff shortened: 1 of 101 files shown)


tomponline:
@markylaing when you get a chance, please can you open a separate PR with the non-placement related commits, to reduce the size of this one, ta.

markylaing:
> @markylaing when you get a chance, please can you open a separate PR with the non-placement related commits, to reduce the size of this one, ta, or anything that is generic that can be added before the main PR.

Yep, a lot of the unrelated ones are linter errors flagged by file changes in this PR, but there's no reason these can't be separated.

markylaing:
Ah annoyingly I was using an old golangci-lint version locally and there are even more lint errors to fix :cry:

markylaing:
Closing in favour of #15408

PR #15291, opened 2025-03-31
github: Bump max lxd-agent size to 14MiB to accomodate dep updates

tomponline:


MusicDin:
Could it be that binary size checks are reversed?
`-gt` -> `-lt "${max_mib}"`
`-lt` -> `-gt "${min_mib}"`

```
            if [ "${cur}" -gt "${max_mib}" ]; then
              echo "FAIL: ${bin} binary size exceeds ${max}MiB"
              exit 1
            fi
            
            if [ "${cur}" -lt "${min_mib}" ]; then
              echo "Congratulations: ${bin} binary size reduced below ${min}MiB"
              echo "It is now time to edit the workflow job to use smaller min/max sizes for ${bin}"
              exit 1
            fi
```

.github/workflows/tests.yml
```diff
@@ -135,7 +135,7 @@ jobs:
 
           # bin/max (sizes are in MiB)
           SIZES="lxc 16
-                 lxd-agent 13"
+                 lxd-agent 12"
```
MusicDin, line 138: The current size seems to be `13354520` which is ~`12.7MiB`, why was this change needed?

tomponline: im not sure whats going on here https://github.com/canonical/lxd/actions/runs/14169474053/job/39689877082?pr=15282

simondeziel: Maybe 1.23.7 produces slightly smaller binaries? Or maybe one of the dependency shrunk a bit in the one brought up to date?

tomponline: removed it lets see what happens

tomponline:
looking happier now.

MusicDin (approved):

PR #15292, opened 2025-03-31
Pre- instance-placement fixes

markylaing:
Work on #15277 has thrown a lot of linter errors. I think this is due to updates to `golangci-lint` as a lot of these files have already had lint errors fixed, but now the linter seems to be catching a lot more cases (especially unused parameters).

There were also a couple of other things I fixed/removed along the way. Adding them here to keep the size of the other PR down and focused.

### DIFF
--- a/lxd/db/cluster/update.go
+++ b/lxd/db/cluster/update.go
@@ -150,7 +150,7 @@ ALTER TABLE identities ADD COLUMN updated_date DATETIME NOT NULL DEFAULT "0001-0
 	return nil
 }
 
-func updateFromV70(ctx context.Context, tx *sql.Tx) error {
+func updateFromV70(_ context.Context, tx *sql.Tx) error {
 	_, err := tx.Exec(`
 CREATE TABLE auth_groups (
     id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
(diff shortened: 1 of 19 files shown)


lxd/db/cluster/update_test.go
```diff
@@ -33,9 +34,12 @@ func TestUpdateFromV0(t *testing.T) {
 	require.Error(t, err)
 
 	// Unique constraint on address
-	stmt = "INSERT INTO nodes VALUES (3, 'bar', 'gasp', '1.2.3.4:666', 9, 11), ?, 0)"
+	stmt = "INSERT INTO nodes VALUES (3, 'bar', 'gasp', '1.2.3.4:666', 9, 11, ?, 0)"
```
simondeziel, line 37: Nice catch!

simondeziel (approved):
Thanks! I'm looking forward to the day where we can enable golangci-lint on all the repo, not just the changed files ;)

lxd/db/cluster/update.go
```diff
@@ -4709,7 +4709,7 @@ CREATE TABLE storage_volumes_config (
 	return err
 }
 
-func updateFromV0(ctx context.Context, tx *sql.Tx) error {
+func updateFromV0(_ context.Context, tx *sql.Tx) error {
```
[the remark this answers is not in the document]
markylaing: Unfortunately this was a manual search and replace, since some functions do use the argument and some don't :cry:

lxd/db/cluster/update_test.go
markylaing: I have my moments :laughing:

tomponline (approved):
Ta!

PR #15293, opened 2025-03-31
lxd/storage/drivers/zfs: suggest `zfs.external` if in snap and ZFS tools are missing

simondeziel:
ATM, if no `zpool` of the right version can be found in the LXD snap, this error is displayed:

```
root@v1:~# lxc storage create zfs zfs
Error: Required tool 'zpool' is missing
```

There is also a more verbose error logged by the snap'ed daemon but it requires the user to look at `journalctl -u snap.lxd.daemon`.

Make the workaround more easily discoverable by tweaking the error message.

### DIFF
--- a/lxd/storage/drivers/driver_zfs.go
+++ b/lxd/storage/drivers/driver_zfs.go
@@ -68,7 +68,7 @@ func (d *zfs) load() error {
 	for _, tool := range []string{"zpool", "zfs"} {
 		_, err := exec.LookPath(tool)
 		if err != nil {
-			return fmt.Errorf("Required tool '%s' is missing", tool)
+			return fmt.Errorf("Required tool %q is missing", tool)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


lxd/storage/drivers/driver_zfs.go
```diff
@@ -68,6 +68,10 @@ func (d *zfs) load() error {
 	for _, tool := range []string{"zpool", "zfs"} {
 		_, err := exec.LookPath(tool)
 		if err != nil {
+			if shared.InSnap() {
+				return fmt.Errorf("Required tool '%s' is missing, consider installing ZFS tools in the host and use 'snap set lxd zfs.external=true'", tool)
```
tomponline, line 72: Please use %q to double quote placeholder strings.

Also can we is IsSnap() to alter message to say something about the host kernel version not matching the bundle zfs tooling?

simondeziel: Let me know what you think of the reworded error. Thanks

tomponline (approved):
Great stuff thanks!

PR #15301, opened 2025-04-01
lxd/device/cdi: ubuntu core: add mesa-2404 contents to config search …

mihalicyn:
…paths

Some files, particularly 10_nvidia_wayland.json are now distributed not as a part of pc-kernel nvidia component, but as a part of mesa-2404 snap. We should add it to a configuration files search locations list too.


tomponline (approved):
Ta

Issue #15221, opened 2025-03-20
feat(validation): network collision when setting ipv4.routes


wideawakening:
**context**
when in need to configure ovn network forwarders, you set `ipv4.routes` CIDR
this range can be in conflict with the `ipv4.ovn.ranges` FIRST-LAST format configuration done during the init process of the cloud. 
the range-ips on both groups should be exclusive and avoid having the same IP being used for different purposes causing a network conflict.

this came from a wrong design, and since we used the auto-asignation of ips, we spotted 
**feature**
if we could get some error when trying to set this up, like 'these IPs are already in use/range for OVN routes (ipv4.ovn.ranges)” would be of help.

for https://canonical-microcloud.readthedocs-hosted.com/en/latest/lxd/reference/network_physical/#network-physical-network-conf:ipv4.routes

**reproducer**
 setup

**side-quest**
check what other network validations we can do to avoid collision when configuring network ranges/routes/gateways either on the network/instance/...  level.

PR #14344, opened 2024-10-25
Network: Consider volatile IP in ovnNetworkExternalSubnets

boltmark:
This PR updates `ovn.ovnNetworkExternalSubnets` to consider the volatile IP of each OVN network using a given uplink. This is added to prevent the use of one OVN network's volatile IP on another OVN network, namely in the cases of creating network forwards and IP passthrough.

PR #15305, opened 2025-04-01
Network: Improve network validation

nmezhenskyi:
Fixes https://github.com/canonical/lxd/issues/15221.

**Changes:**
- Validate that listen addresses of network forwards and load balancers do not overlap with OVN ranges. However, allow forwards and load balancers on OVN networks using `volatile.network.ipv4.address` and `volatile.network.ipv6.address` as listen address.
- Validate that `ipv4.routes` and `ipv6.routes` on physical and bridge networks are compatible with existing forwards and load balancers.
- Prevent using IPs from OVN range for NIC passthrough to an instance using `ipv4.routes.external` and `ipv6.routes.external`.
- Add tests to check that forwards and load balancers cannot be created with a listen address that overlaps with OVN ranges.
- Add tests to check that `ipv4.routes` and `ipv6.routes` cannot be removed for existing OVN forwards and load balancers.
- Add tests to check that forwards and load balancers can be set up on `volatile.network.ipv4.address` and `volatile.network.ipv6.address` of the associated OVN network.
- Add tests to check that instance NIC passthrough does not allow using IPs from OVN ranges.

### DIFF
--- a/lxd/device/nic_ovn.go
+++ b/lxd/device/nic_ovn.go
@@ -278,6 +278,95 @@ func (d *nicOVN) validateConfig(instConf instance.ConfigReader) error {
 		}
 	}
 
+	// Validate that NIC passthrough does not allow IPs from OVN range.
+	err = d.validateExternalRoutes()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// validateExternalRoutes checks that external routes do not allow to passthrough IPs from OVN range.
+func (d *nicOVN) validateExternalRoutes() error {
+	var (
+		routesListIPv4 []*net.IPNet
+		routesListIPv6 []*net.IPNet
+
+		ovnRangesListIPv4 []*shared.IPRange
+		ovnRangesListIPv6 []*shared.IPRange
+
+		uplink *api.Network
+
+		err error
+	)
+
+	if d.network.Config() == nil {
+		return fmt.Errorf("Network config is missing for NIC device %q", d.name)
+	}
+
+	uplinkName := d.network.Config()["network"]
+	if uplinkName == "" {
+		return fmt.Errorf(`OVN network %q is missing "network" config option`, d.network.Name())
+	}
+
+	// Get uplink network config.
+	err = d.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+		// Uplink has to be in the "default" project.
+		_, uplink, _, err = tx.GetNetworkInAnyState(ctx, api.ProjectDefaultName, uplinkName)
+
+		return err
+	})
+	if err != nil {
+		return fmt.Errorf("Failed to get config for network %q: %w", uplinkName, err)
+	}
+
+	if d.config["ipv4.routes.external"] != "" {
+		routesListIPv4, err = shared.ParseNetworks(d.config["ipv4.routes.external"])
+		if err != nil {
+			return fmt.Errorf("Failed parsing ipv4.routes: %w", err)
+		}
+
+		if uplink.Config["ipv4.ovn.ranges"] != "" {
+			ovnRangesListIPv4, err = shared.ParseIPRanges(uplink.Config["ipv4.ovn.ranges"])
+			if err != nil {
+				return fmt.Errorf("Failed parsing ipv4.ovn.ranges: %w", err)
+			}
+		}
+	}
+
+	if d.config["ipv6.routes.external"] != "" {
+		routesListIPv6, err = shared.ParseNetworks(d.config["ipv6.routes.external"])
+		if err != nil {
+			return fmt.Errorf("Failed parsing ipv6.routes: %w", err)
+		}
+
+		if uplink.Config["ipv6.ovn.ranges"] != "" {
+			ovnRangesListIPv6, err = shared.ParseIPRanges(uplink.Config["ipv6.ovn.ranges"])
+			if err != nil {
+				return fmt.Errorf("Failed parsing ipv6.ovn.ranges: %w", err)
+			}
+		}
+	}
+
+	// Validate "ipv4.routes.external".
+	for _, routes := range routesListIPv4 {
+		for _, ovnRange := range ovnRangesListIPv4 {
+			if ovnRange.ContainsIP(routes.IP.To16()) {
+				return fmt.Errorf(`Route %q overlaps with "ipv4.ovn.ranges" (%q)`, routes, ovnRange)
+			}
+		}
+	}
+
+	// Validate "ipv6.routes.external".
+	for _, routes := range routesListIPv6 {
+		for _, ovnRange := range ovnRangesListIPv6 {
+			if ovnRange.ContainsIP(routes.IP.To16()) {
+				return fmt.Errorf(`Route %q overlaps with "ipv6.ovn.ranges" (%q)`, routes, ovnRange)
+			}
+		}
+	}
+
 	return nil
 }
 
--- a/lxd/device/nic_ovn.go
+++ b/lxd/device/nic_ovn.go
@@ -536,7 +536,8 @@ func (d *nicOVN) Start() (*deviceConfig.RunConfig, error) {
 	if d.config["nested"] != "" {
 		delete(saveData, "host_name") // Nested NICs don't have a host side interface.
 	} else {
-		if d.config["acceleration"] == "sriov" {
+		switch d.config["acceleration"] {
+		case "sriov":
 			ovs := openvswitch.NewOVS()
 			if !ovs.HardwareOffloadingEnabled() {
 				return nil, fmt.Errorf("SR-IOV acceleration requires hardware offloading be enabled in OVS")
@@ -583,7 +584,7 @@ func (d *nicOVN) Start() (*deviceConfig.RunConfig, error) {
 
 			integrationBridgeNICName = vfRepresentor
 			peerName = vfDev
-		} else if d.config["acceleration"] == "vdpa" {
+		case "vdpa":
 			ovs := openvswitch.NewOVS()
 			if !ovs.HardwareOffloadingEnabled() {
 				return nil, fmt.Errorf("SR-IOV acceleration requires hardware offloading be enabled in OVS")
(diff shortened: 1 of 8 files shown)


nmezhenskyi:
Static analysis was giving me errors like this one:
```
Error: lxd/network/driver_bridge.go:1017:11: QF1008: could remove embedded field "common" from selector (staticcheck)
	return n.common.delete()
```

So I'm removing unnecessary `.common` selectors.

test/suites/network_ovn.sh
```diff
@@ -80,6 +80,25 @@ test_network_ovn() {
       ipv4.address=10.24.140.1/24 ipv4.nat=true \
       ipv6.address=fd42:bd85:5f89:5293::1/64 ipv6.nat=true
 
+  # Create a dummy uplink physical network to test network collision validation.
```
simondeziel, line 83: Could you turn all those comments into `echo`'s with the same content?

This will look out of place at first, but I believe we should be `echo`'ing a lot more in tests to aid debugging/following along. So we need to start somewhere :)

nmezhenskyi: Sure, will do! :slightly_smiling_face:

simondeziel (approved):
Thanks, LGTM

shared/network_ip.go
```diff
@@ -147,10 +147,31 @@ func (r *IPRange) Overlaps(otherRange *IPRange) bool {
...
 func (r *IPRange) String() string {
 	if r.End == nil {
 		return r.Start.String()
 	}
 
 	return fmt.Sprintf("%v-%v", r.Start, r.End)
 }
+
+// ParseNetworks parses a comma separated list of IP networks in CIDR notation.
+func ParseNetworks(netList string) ([]*net.IPNet, error) {
```
tomponline, line 164: Similar to `ParseIPRanges` above, I think we need to trim the space around each network, as the validator uses:

```go
validate.Optional(validate.IsListOf(validate.IsNetworkV6))
```

Which allows for example `10.0.0.0/24,      192.168.0.1/32     , 127.0.0.0/8`

simondeziel: Such parser functions make good candidate for quick unit tests. Not asking for this specifically but would be nice to have if you can spare the time ;)

simondeziel (dismissed):
LGTM, thanks for the unit and system tests!

tomponline (changes requested):
@nmezhenskyi looking at this PR and the reported issue more deeply I think the issue's suggestion of:

> if we could get some error when trying to set this up, like 'these IPs are already in use/range for OVN routes (ipv4.ovn.ranges)” would be of help.

Is not quite right, whilst at a high level does make sense.

If you have a single subnet, say a /24 that you want to use for your OVN network(s) on your uplink network, then you should be able to use a configuration like this:

```
ipv4.gateway: 10.0.0.1/24
ipv4.ovn.ranges: 10.0.0.2-10.0.0.20
ipv4.routes: 10.0.0.0/24
```

This says "10.0.0.1 and 10.0.0.2-10.0.0.20 are reserved for OVN virtual routers, but the rest of 10.0.0.0/24 is available for use with NIC ipv4.routes.external, load balancers and network forwarders".

Further more, it is acceptable that one can setup a load balancer or network forward (but not a NIC ipv4.routes.external passthrough) using the same `volatile.network.ipv4.address` as the associated OVN network (assuming its not been taken by any other load balancer or network forward).

So I think we need to reconsider that the validation should actually occur when we are being asked to create a specific listener (network forward or load balancer) or auto allocating an address.

nmezhenskyi:
@tomponline This makes sense to me. I am working on changing the validation approach.  

While I was investigating this issue further, I found that it is possible to unset `ipv4.routes` on the uplink network even if the OVN network already has some network forwards. How should we go about that? Should we prevent removing `ipv4.routes` on the uplink if the dependent OVN network has forwards, or should we just issue a warning message? As another option, we could delete the affected network forwards while removing `ipv4.routes` on the uplink.

tomponline:
> Should we prevent removing `ipv4.routes` on the uplink if the dependent OVN network has forwards

Yes i think we should. Thanks for spotting.

lxd/network/driver_common.go
```diff
@@ -866,6 +866,59 @@ func (n *common) forwardValidate(listenAddress net.IP, forward api.NetworkForwar
...
+	ovnRangesKey := "ipv4.ovn.ranges"
+	if !listenIsIP4 {
+		ovnRangesKey = "ipv6.ovn.ranges"
+	}
+
+	var ovnRanges []*shared.IPRange
+	var err error
+
+	// Validate listen address for collision with OVN ranges for a "bridge" type network.
+	if n.netType == "bridge" && n.config[ovnRangesKey] != "" {
```
nmezhenskyi, line 878: I'm wondering about this part. Should I keep this logic in `driver_common.go` or should I split these cases into `driver_bridge.go` and `driver_ovn.go`? On the one hand, current approach avoids some code duplication, but on the other, it creates additional coupling in the `driver_common`.

tomponline: i think we can restructure to have a per-network driver `forwardValidate` function that then calls into the common.forwardValidate function as well as doing any per-driver validation logic.

nmezhenskyi:
@tomponline I was looking into this part:

> Further more, it is acceptable that one can setup a load balancer or network forward (but not a NIC ipv4.routes.external passthrough) using the same volatile.network.ipv4.address as the associated OVN network (assuming its not been taken by any other load balancer or network forward).  

I think the existing validation for NIC `ipv4.routes.external` and `ipv6.routes.external` already prevents using an OVN volatile network address. There was a PR for it https://github.com/canonical/lxd/pull/14344. I tried manually testing a NIC passthrough using volatile.network.ipv4.address and it didn't let me.

nmezhenskyi:
UPD:

> However, NIC passthrough currently allows using other IPs from the OVN range.

Fixed it, that's no longer the case.

Also, this PR breaks a small section of `network-ovn` test in `lxd-ci` because of new validation logic that prevents removal of ipv4/ipv6 routes on the uplink if there are exisiting forwards. I'll create a PR to fix the test there.

simondeziel (approved):
Tests are looking good, thanks again!

tomponline:
please rebase

nmezhenskyi:
Rebased.  

System (cluster, btrfs) test failed because it couldn't install dependencies for some reason, I'll rerun it.

lxd/network/driver_ovn.go
```diff
@@ -5887,3 +5907,76 @@ func (n *ovn) checkAddressNotInUse(netip *net.IPNet) (bool, error) {
...
+func (n *ovn) checkAddressNotInOVNRange(addr net.IP) error {
+	if addr == nil {
+		return fmt.Errorf("Invalid listen address")
+	}
+
+	if n.config["network"] == "" {
+		return fmt.Errorf(`OVN network %q is missing "network" config option`, n.name)
+	}
+
+	addrIsIP4 := addr.To4() != nil
```
tomponline, line 5925: just a nit but i would have moved that down to where its tested on line 5930 to keep it together.

lxd/network/driver_ovn.go
```diff
@@ -5887,3 +5907,76 @@ func (n *ovn) checkAddressNotInUse(netip *net.IPNet) (bool, error) {
...
+
+		if addr.Equal(volatileAddr) {
+			return nil
+		}
+	}
+
+	var uplink *api.Network
+	var err error
+
+	uplinkName := n.config["network"]
```
tomponline, line 5948: Again, just a nit, but i would suggest defining this var above the check for `if n.config["network"] == "" {` and then changing to `if uplinkName == "" {` so you're testing the same var you use further down to avoid the chance of the two diverging later on.

lxd/network/driver_common.go
```diff
@@ -209,6 +209,113 @@ func (n *common) validateZoneNames(config map[string]string) error {
...
+			return fmt.Errorf("Failed to get listen addresses of OVN load balancers: %w", err)
+		}
+
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	checkListenAddressesOVN := func(project, network string, ips []string) error {
```
tomponline, line 257: we prefer not using compact argument type style, so instead

```go
func(project string, network string, ips []string) error
```

lxd/network/driver_common.go
```diff
@@ -209,6 +209,113 @@ func (n *common) validateZoneNames(config map[string]string) error {
...
+			}
+		}
+
+		return nil
+	}
+
+	// Check that IP routes are provided for already existing OVN forwards.
+	for project, networks := range forwards {
+		for network, ips := range networks {
+			if n.name == network && n.project == project {
```
tomponline, line 295: as this check would be safe to do for load balancers as well, it seems like we could combine these two sections for forwards and loadBalancers into a single look, e.g.

```go
for _, routes := range []{forwards, loadBalancers} {
    for project, networks := range routes {
        ....
    }
}
```

```

File: lxd/network/driver_physical.go [hunk 8]
```diff
@@ -219,6 +219,12 @@ func (n *physical) Validate(config map[string]string) error {
 		return err
 	}
 
+	// Validate IP routes.
```
tomponline, line 222: The commit message:

> Check that ipv4.routes and ipv6.routes contain the routes for existing OVN network forwards and load balancers.

Would have made a better comment here than "Validate IP routes" which is clear from the function name ;)

lxd/device/nic_ovn.go
```diff
@@ -278,6 +278,91 @@ func (d *nicOVN) validateConfig(instConf instance.ConfigReader) error {
...
+
+		ovnRangesListIPv4 []*shared.IPRange
+		ovnRangesListIPv6 []*shared.IPRange
+
+		uplink *api.Network
+
+		err error
+	)
+
+	uplinkName := d.network.Config()["network"]
```
tomponline, line 304: the TICS checker might complain that a panic is possible here if `Config()` returns nil, best to check.

tomponline (changes requested):
Excellent work thank you.

A few very small nits and suggestions.

test/suites/network_ovn.sh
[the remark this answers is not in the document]
nmezhenskyi: No, I forgot to add them here, will do. I think we have some in the `lxd-ci`.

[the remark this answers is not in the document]
tomponline: Also the tests we add to the main LXD test suite will count towards our TICS coverage report whereas right now the ones in lxd-ci do not.

nmezhenskyi:
Thanks for the review! I've made the changes and added tests for `ipvN.routes.external`. Should be ready now.

lxd/device/nic_ovn.go
tomponline: @nmezhenskyi fyi this wasnt addressed

wideawakening:
thanks for those extra checks/validations! 🧡

PR #15306, opened 2025-04-02
DON'T MERGE: disable all tests but backup_volume_export

skozina:
This is to debug the flaky cephfs backup_volume_export test.
We disable all other tests, and instead repeat the core of backup_volume_export test on cephfs many times to see if it consistently fails.


tomponline:
@skozina can we close this now?

skozina:
> @skozina can we close this now?

I thought I'd re-use this to test the sync() call from LXD directly. But that won't be that easy, so yes, let me close this now. Thanks!

PR #15308, opened 2025-04-02
test/suites/backup: On cephfs add sync calls after `file push`

skozina:
Fixes the flaky test of volume export on cephfs.

Tested with custom PR on github which isolates the core of the backup_volume_export test and runs it in a loop:
https://github.com/canonical/lxd/pull/15306

Without the sync call, the test suite failed after 17 runs of the inner loop. With the sync, it run for 100 runs without failure.

### DIFF
--- a/test/suites/backup.sh
+++ b/test/suites/backup.sh
@@ -736,6 +736,7 @@ _backup_volume_export_with_project() {
 
   # Create file on the custom volume.
   echo foo | lxc file push - c1/mnt/test
+  LXC_LOCAL='' lxc_remote exec c1 -- sync /mnt/test
 
   # Snapshot the custom volume.
   lxc storage volume set "${custom_vol_pool}" testvol user.foo=test-snap0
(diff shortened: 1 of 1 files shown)


tomponline (commented):
I suggest we just always add the sync call irrespective of the storage driver.
This way we can ensure we have flushed the filesystem cache before taking the snapshot. We do this already for VM tests in lxd-ci.

skozina:
I've removed the storage driver conditional.
Also I am fixing the test failure -- here we inject `--force-local` in the `lxc exec c1 -- sync` call, which causes LXD *not* to use the provided configuration via LXD_DIR, which makes LXD *not* to use the currently selected project:
https://github.com/canonical/lxd/blob/main/test/includes/lxc.sh
The fix is to use `lxc_remote` instead.

tomponline:
Thanks! Another option would have been to add an explicit `--project` flag to the exec calls I imagine.

skozina:
> Thanks! Another option would have been to add an explicit `--project` flag to the exec calls I imagine.

Yes, that should work too. I thought it would just be harder to understand why there's an explicit project setting on this specific line.

PR #15310, opened 2025-04-02
Resources: Add device's filesystem UUID if present

roosterfish:
This allows MicroCloud checking (https://github.com/canonical/microcloud/pull/724) whether or not a device is already used after fetching the list of disks from `/1.0/resources`. 

The PR adds another field `device_fs_uuid` to both disks and partitions returned from the resources endpoint.
A new API extension `resources_device_fs_uuid` is added to indicate this change.

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2638,3 +2638,7 @@ Adds new {config:option}`instance-security:security.delegate_bpf`.* group of opt
 ## `override_snapshot_profiles_on_copy`
 
 This adds a request option to set snapshot's target profile on instance copy to be inherited from target instance.
+
+## `resources_device_fs_uuid`
+
+Adds the field `device_fs_uuid` including the respective UUID to each disk and partition indicating whether or not a filesystem is located on the device.
--- a/lxd/resources/storage.go
+++ b/lxd/resources/storage.go
@@ -5,7 +5,6 @@ import (
 	"bytes"
 	"fmt"
 	"os"
-	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -13,7 +12,6 @@ import (
 
 	"golang.org/x/sys/unix"
 
-	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
 )
 
@@ -438,24 +436,4 @@ func GetStorage() (*api.ResourcesStorage, error) {
 	}
 
 	return &storage, nil
-}
-
-// GetDisksByID returns all disks whose ID contains the filter prefix.
-func GetDisksByID(filterPrefix string) ([]string, error) {
-	disks, err := os.ReadDir(devDiskByID)
-	if err != nil {
-		return nil, fmt.Errorf("Failed getting disks by ID: %w", err)
-	}
-
-	var filteredDisks []string
-	for _, disk := range disks {
-		// Skip the disk if it does not have the prefix.
-		if !shared.StringHasPrefix(disk.Name(), filterPrefix) {
-			continue
-		}
-
-		filteredDisks = append(filteredDisks, path.Join(devDiskByID, disk.Name()))
-	}
-
-	return filteredDisks, nil
 }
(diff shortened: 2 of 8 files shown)


shared/util.go
```diff
@@ -819,6 +819,16 @@ func StringMapHasStringKey(m map[string]string, keys ...string) bool {
 	return false
 }
 
+// BlockFsUUID returns the UUID of a filesystem from the block device.
+func BlockFsUUID(pathName string) (string, error) {
```
tomponline, line 823: We should really only be moving things *out* of the bloated catch-call `shared` package and not into it.

As `lxd/resources` is already AGPL 3.0, there should be no problem with it consuming other packages that are outside of `shared`.

So I suggest moving this function into `lxd/storage/filesystem` as that seems more appropriate.

tomponline: It could also be in `lxd/storage/block` potentially.

And the "Fs" should be "FS" I think as its an acronym.

doc/api-extensions.md
```diff
@@ -2638,3 +2638,7 @@ Adds new {config:option}`instance-security:security.delegate_bpf`.* group of opt
 ## `override_snapshot_profiles_on_copy`
 
 This adds a request option to set snapshot's target profile on instance copy to be inherited from target instance.
+
+## `resources_device_fs_uuid`
```
tomponline, line 2642: this should be part of the previous commit as per https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#commit-structure

shared/api/resource.go
```diff
@@ -679,6 +679,12 @@ type ResourcesStorageDisk struct {
 	//
 	// API extension: resources_disk_address
 	USBAddress string `json:"usb_address,omitempty" yaml:"usb_address,omitempty"`
+
+	// UUID of the filesystem on the device
+	// Example: 9313518c-0e13-4067-9746-5c1703830b78
+	//
+	// API extension: resources_device_fs_uuid
+	DeviceFSUUID string `json:"device_fs_uuid" yaml:"device_fs_uuid"`
```
tomponline, line 687: Presumably the population of this field is mutually exclusive to the Partitions field right?

doc/api-extensions.md
roosterfish: Fixed.

lxd/resources/storage.go
tomponline, line 1: nice

tomponline, line 1: nice

lxd/resources/storage.go
```diff
@@ -420,6 +427,12 @@ func GetStorage() (*api.ResourcesStorage, error) {
 				}
 			}
 
+			// Pull device filesystem UUID information.
+			disk.DeviceFSUUID, err = block.DiskFSUUID(filepath.Join("/dev", entryName))
+			if err != nil {
+				return nil, err
```
tomponline, line 433: do we want this to be an error scenario?

lxd/resources/storage.go
```diff
@@ -357,6 +358,12 @@ func GetStorage() (*api.ResourcesStorage, error) {
 
 				partition.Size = partitionSize * 512
 
+				// Pull device filesystem UUID information.
+				partition.DeviceFSUUID, err = block.DiskFSUUID(filepath.Join("/dev", subEntryName))
+				if err != nil {
+					return nil, err
```
tomponline, line 364: do we want this to be an error scenario?

lxd/resources/storage.go
roosterfish: Same as in https://github.com/canonical/lxd/pull/15310/files#r2030934069.

tomponline (approved):

tomponline (changes requested):
Error: lxd/storage/block/utils.go:18:7: exported: exported const DevDiskByID should have comment or be unexported (revive)
const DevDiskByID = "/dev/disk/by-id"

roosterfish:
@tomponline lint errors fixed now, missed a few due to the renaming.

tomponline (approved):

roosterfish:
@tomponline can I add a PR to backport this to 5.21?

tomponline:
> @tomponline can I add a PR to backport this to 5.21?

Be my guest :)

Issue #15298, opened 2025-03-31
feat: improve CLI completion for `lxc auth` subcommands


simondeziel:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

Some of the `lxc auth` subcommands are missing CLI completion

### Describe the solution you'd like

It'd be nice if `lxc auth group delete` would autocomplete based on existing groups:

```
root@micro1:~# lxc auth group list
+-------------+-------------+
|    NAME     | DESCRIPTION |
+-------------+-------------+
| e2e-testing |             |
+-------------+-------------+
root@micro1:~# lxc auth group delete <tab>
```

Similarly, `lxc auth group permission add` should also propose completing based on existing groups.

Also, `lxc auth identity group add` should propose completing existing entities and groups:

```
root@micro1:~# lxc auth identity list
+-----------------------+--------------------+-------------+------------------------------------------------------------------+-------------+
| AUTHENTICATION METHOD |        TYPE        |    NAME     |                            IDENTIFIER                            |   GROUPS    |
+-----------------------+--------------------+-------------+------------------------------------------------------------------+-------------+
| tls                   | Client certificate | e2e-testing | 6c9e84dc87c5e7b6f90d43e7e5055b2121a0e28549c8a84936192e8455e27108 | e2e-testing |
+-----------------------+--------------------+-------------+------------------------------------------------------------------+-------------+
| tls                   | Server certificate | micro1      | 9528207a660b89fa4b82e1b20d415f68019c92e513d6e3a747b707b21c37ead0 |             |
+-----------------------+--------------------+-------------+------------------------------------------------------------------+-------------+
| tls                   | Server certificate | micro2      | a1d5ef70354a730d4fff99fe9260ebb091e810cf58e3fcecf78ed2eeb941c33b |             |
+-----------------------+--------------------+-------------+------------------------------------------------------------------+-------------+
| tls                   | Server certificate | micro3      | 415c0ff0bb007bc4b5ddd249a956b3fa342283aeccdac2b25cf19284caab74be |             |
+-----------------------+--------------------+-------------+------------------------------------------------------------------+-------------+
root@micro1:~# lxc auth group list
+-------------+-------------+
|    NAME     | DESCRIPTION |
+-------------+-------------+
| e2e-testing |             |
+-------------+-------------+

# this is what I would have liked to tab complete to:
[...]

PR #15311, opened 2025-04-02
Completions refactor

markylaing:
While working on #15298 I noticed a great deal of code duplication in `lxc/completions.go`. This PR attempts to generalise things a bit by using common methods for resources that are "top level". By "top level" resources, I mean resources that can be listed without any path parameters. These are all resources that can be referenced with `<remote>:<name>`.

To do this I'm using an entity type concept again, I haven't used the `shared/entity` package for this because the ultimate goal will be to remove the `shared/entity` package and rely only on the metadata API.

For reviewers: Does anyone know how we can test code completions programatically?


kadinsayani:
> For reviewers: Does anyone know how we can test code completions programatically?

@simondeziel and I chatted about this a few months ago. We can call completion functions directly using `__complete` and assert the resulting completions and completion directives. For example:

```
❯ lxc __complete stop v
v1
v2
:4
Completion ended with directive: ShellCompDirectiveNoFileComp
```

Where `v1` and `v2` are the expected resulting completions.

kadinsayani (dismissed):
Thanks!

markylaing:
> > For reviewers: Does anyone know how we can test code completions programatically?
> 
> @simondeziel and I chatted about this a few months ago. We can call completion functions directly using `__complete` and assert the resulting completions and completion directives. For example:
> 
> ```
> ❯ lxc __complete stop v
> v1
> v2
> :4
> Completion ended with directive: ShellCompDirectiveNoFileComp
> ```
> 
> Where `v1` and `v2` are the expected resulting completions.

Cool, in that case I'd like to add some tests so that we can verify that the behaviour is identical before and after.

markylaing:
Marking as draft while I work on tests. Initial tests are showing some unexpected results (mainly, completions returning all resources and ignoring the beginning of the completion text) so there will be more refactoring commits.

tomponline:
@markylaing is this something that will likely be mergeable soon, or if its likely to be an ongoing background project can we close the PR?

markylaing:
> @markylaing is this something that will likely be mergeable soon, or if its likely to be an ongoing background project can we close the PR?

I've just go to a reasonable cut off point with it so I'm going to be pushing the changes soon :+1: 

Also, RE:

> Initial tests are showing some unexpected results (mainly, completions returning all resources and ignoring the beginning of the completion text) so there will be more refactoring commits.

It is ok for completions to return all results because these are filtered out by the bash part of the completion. However, this can make testing quite laborious so wherever possible I'm prefiltering in go

markylaing:
This PR now adds some completions testing but there is a lot still to do. Here is a list of all commands yet to be tested:

- [ ] network      Manage and attach instances to networks
  - [ ] acl               Manage network ACLs
    - [ ] create       Create new network ACLs
    - [ ] delete       Delete network ACLs
    - [ ] edit         Edit network ACL configurations as YAML
    - [ ] get          Get values for network ACL configuration keys
    - [ ] list         List available network ACLS
    - [ ] rename       Rename network ACLs
    - [ ] rule         Manage network ACL rules
      - [ ] add          Add rules to an ACL
      - [ ] remove       Remove rules from an ACL
    - [ ] set          Set network ACL configuration keys
    - [ ] show         Show network ACL configurations
    - [ ] show-log     Show network ACL log
    - [ ] unset        Unset network ACL configuration keys
  - [ ] attach            Attach network interfaces to instances
  - [ ] attach-profile    Attach network interfaces to profiles
  - [ ] create            Create new networks
  - [ ] delete            Delete networks
  - [ ] detach            Detach network interfaces from instances
  - [ ] detach-profile    Detach network interfaces from profiles
  - [ ] edit              Edit network configurations as YAML
  - [ ] forward           Manage network forwards
    - [ ] create       Create new network forwards
    - [ ] delete       Delete network forwards
    - [ ] edit         Edit network forward configurations as YAML
    - [ ] get          Get values for network forward configuration keys
    - [ ] list         List available network forwards
    - [ ] port         Manage network forward ports
      - [ ] add          Add ports to a forward
      - [ ] remove       Remove ports from a forward
    - [ ] set          Set network forward keys
    - [ ] show         Show network forward configurations
    - [ ] unset        Unset network forward configuration keys
  - [ ] get               Get values for network configuration keys
  - [ ] info              Get runtime information on networks
  - [ ] list              List available networks
  - [ ] list-allocations  List network allocations in use
  - [ ] list-leases       List DHCP leases
  - [ ] load-balancer     Manage network load balancers
    - [ ] backend      Manage network load balancer backends
      - [ ] add          Add backends to a load balancer
      - [ ] remove       Remove backends from a load balancer
    - [ ] create       Create new network load balancers
    - [ ] delete       Delete network load balancers
    - [ ] edit         Edit network load balancer configurations as YAML
    - [ ] get          Get values for network load balancer configuration keys
    - [ ] list         List available network load balancers
    - [ ] port         Manage network load balancer ports
      - [ ] add          Add ports to a load balancer
      - [ ] remove       Remove ports from a load balancer
    - [ ] set          Set network load balancer keys
    - [ ] show         Show network load balancer configurations
    - [ ] unset        Unset network load balancer configuration keys
  - [ ] peer              Manage network peerings
    - [ ] create       Create new network peering
    - [ ] delete       Delete network peerings
    - [ ] edit         Edit network peer configurations as YAML
    - [ ] get          Get values for network peer configuration keys
    - [ ] list         List available network peers
    - [ ] set          Set network peer keys
    - [ ] show         Show network peer configurations
    - [ ] unset        Unset network peer configuration keys
  - [ ] rename            Rename networks
  - [ ] set               Set network configuration keys
  - [ ] show              Show network configurations
  - [ ] unset             Unset network configuration keys
  - [ ] zone              Manage network zones
    - [ ] create       Create new network zones
    - [ ] delete       Delete network zones
    - [ ] edit         Edit network zone configurations as YAML
    - [ ] get          Get values for network zone configuration keys
    - [ ] list         List available network zoneS
    - [ ] record       Manage network zone records
      - [ ] create       Create new network zone record
      - [ ] delete       Delete network zone record
      - [ ] edit         Edit network zone record configurations as YAML
      - [ ] entry        Manage network zone record entries
      - [ ] get          Get values for network zone record configuration keys
      - [ ] list         List available network zone records
      - [ ] set          Set network zone record configuration keys
      - [ ] show         Show network zone record configuration
      - [ ] unset        Unset network zone record configuration keys
    - [ ] set          Set network zone configuration keys
    - [ ] show         Show network zone configurations
    - [ ] unset        Unset network zone configuration keys
- [ ] operation    List, show and delete background operations
  - [ ] delete       Delete a background operation (will attempt to cancel)
  - [ ] list         List background operations
  - [ ] show         Show details on a background operation
- [ ] profile      Manage profiles
  - [ ] add          Add profiles to instances
  - [ ] assign       Assign sets of profiles to instances
  - [ ] copy         Copy profiles
  - [ ] create       Create profiles
  - [ ] delete       Delete profiles
  - [ ] device       Manage devices
    - [ ] add          Add instance devices
    - [ ] list         List instance devices
    - [ ] set          Set device configuration keys
  - [ ] edit         Edit profile configurations as YAML
  - [ ] get          Get values for profile configuration keys
  - [ ] list         List profiles
  - [ ] remove       Remove profiles from instances
  - [ ] rename       Rename profiles
  - [ ] set          Set profile configuration keys
  - [ ] show         Show profile configurations
  - [ ] unset        Unset profile configuration keys
- [ ] project      Manage projects
  - [ ] create       Create projects
  - [ ] delete       Delete projects
  - [ ] edit         Edit project configurations as YAML
  - [ ] get          Get values for project configuration keys
  - [ ] info         Get a summary of resource allocations
  - [ ] list         List projects
  - [ ] rename       Rename projects
  - [ ] set          Set project configuration keys
  - [ ] show         Show project options
  - [ ] switch       Switch the current project
  - [ ] unset        Unset project configuration keys
- [ ] query        Send a raw query to LXD
- [ ] storage      Manage storage pools and volumes
  - [ ] bucket       Manage storage buckets
    - [ ] create       Create new custom storage buckets
    - [ ] delete       Delete storage buckets
    - [ ] edit         Edit storage bucket configurations as YAML
    - [ ] get          Get values for storage bucket configuration keys
    - [ ] key          Manage storage bucket keys
      - [ ] create       Create key for a storage bucket
      - [ ] delete       Delete key from a storage bucket
      - [ ] edit         Edit storage bucket key as YAML
      - [ ] list         List storage bucket keys
      - [ ] show         Show storage bucket key configurations
    - [ ] list         List storage buckets
    - [ ] set          Set storage bucket configuration keys
    - [ ] show         Show storage bucket configurations
    - [ ] unset        Unset storage bucket configuration keys
  - [ ] create       Create storage pools
  - [ ] delete       Delete storage pools
  - [ ] edit         Edit storage pool configurations as YAML
  - [ ] get          Get values for storage pool configuration keys
  - [ ] info         Show useful information about storage pools
  - [ ] list         List available storage pools
  - [ ] set          Set storage pool configuration keys
  - [ ] show         Show storage pool configurations and resources
  - [ ] unset        Unset storage pool configuration keys
  - [ ] volume       Manage storage volumes
    - [ ] attach          Attach new storage volumes to instances
    - [ ] attach-profile  Attach new storage volumes to profiles
    - [ ] copy            Copy storage volumes
    - [ ] create          Create new custom storage volumes
    - [ ] delete          Delete storage volumes
    - [ ] detach          Detach storage volumes from instances
    - [ ] detach-profile  Detach storage volumes from profiles
    - [ ] edit            Edit storage volume configurations as YAML
    - [ ] export          Export custom storage volume
    - [ ] get             Get values for storage volume configuration keys
    - [ ] import          Import custom storage volumes
    - [ ] info            Show storage volume state information
    - [ ] list            List storage volumes
    - [ ] move            Move storage volumes between pools
    - [ ] rename          Rename storage volumes and storage volume snapshots
    - [ ] restore         Restore storage volume snapshots
    - [ ] set             Set storage volume configuration keys
    - [ ] show            Show storage volume configurations
    - [ ] snapshot        Snapshot storage volumes
    - [ ] unset           Unset storage volume configuration keys
- [ ] version      Show local and remote versions
- [ ] warning      Manage warnings
  - [ ] acknowledge  Acknowledge warning
  - [ ] delete       Delete warning
  - [ ] list         List warnings
  - [ ] show         Show warning
- [ ] cluster      Manage cluster members
  - [ ] add                 Request a join token for adding a cluster member
  - [ ] edit                Edit cluster member configurations as YAML
  - [ ] enable              Enable clustering on a single non-clustered LXD server
  - [ ] evacuate            Evacuate cluster member
  - [ ] get                 Get values for cluster member configuration keys
  - [ ] group               Manage cluster groups
    - [ ] add          Add member to group
    - [ ] assign       Assign sets of groups to cluster members
    - [ ] create       Create a cluster group
    - [ ] delete       Delete a cluster group
    - [ ] edit         Edit a cluster group
    - [ ] list         List all the cluster groups
    - [ ] remove       Remove member from group
    - [ ] rename       Rename a cluster group
    - [ ] show         Show cluster group configurations
  - [ ] info                Show useful information about a cluster member
  - [ ] list                List all the cluster members
  - [ ] list-tokens         List all active cluster member join tokens
  - [ ] remove              Remove a member from the cluster
  - [ ] rename              Rename a cluster member
  - [ ] restore             Restore cluster member
  - [ ] revoke-token        Revoke cluster member join token
  - [ ] role                Manage cluster roles
    - [ ] add          Add roles to a cluster member
    - [ ] remove       Remove roles from a cluster member
  - [ ] set                 Set a cluster members configuration keys
  - [ ] show                Show details of a cluster member
  - [ ] unset               Unset a cluster members configuration keys
  - [ ] update-certificate  Update cluster certificate

tomponline:
> This means the output of `lxc __complete` that we use for testing can be different from the actual output. So where I've refactored, I've done the filtering in Go, without relying on the subsequent bash script, so that we can be certain it is working as expected.

Thanks, this sounds preferable anyway, to do as much in Go as possible and as little in bash.

markylaing:
Tests should pass now :crossed_fingers:

As discussed @tomponline I've also added completions for `gputype=` and narrowed the completions based on the subtype (as for nictype)

test/suites/completions.sh
```diff
@@ -0,0 +1,330 @@
...
+    [ "$(complete export c)" = 'c1,c2' ]
+
+    (
+      import_wd=$(mktemp -d -p "${TEST_DIR}" XXX)
+      # shellcheck disable=SC2164
+      # (tests are run with set -e)
+      cd "${import_wd}"
+      mkdir foo
+      touch bar.txt fizz.tar buzz.tar.gz bazz.tar.xz
+      [ "$(complete import '')" = 'buzz.tar.gz,foo/,localhost:' ] # Correct file extensions, directories, and non-default remotes
```
simondeziel, line 87: It seems that `tar.xz` are also importable:
```
$ file c1.tar.xz 
c1.tar.xz: XZ compressed data, checksum CRC64
$ lxc import c1.tar.xz; echo $?
0
$ lxc info c1
Name: c1
Status: STOPPED
Type: container
Architecture: x86_64
Created: 2025/04/10 13:57 EDT
```

test/suites/completions.sh
```diff
@@ -0,0 +1,330 @@
...
+      [ "$(complete import localhost: '')" = 'buzz.tar.gz,foo/' ]
+      rm -rf "${import_wd}"
+    )
+
+    [ "$(complete info '')" = 'c1,c2,localhost:' ] # Info should show only instance server remotes and not the default
+    [ "$(complete info l)" = 'localhost:' ]
+    [ "$(complete info c)" = 'c1,c2' ]
+    testimage_fingerprint="$(lxc query /1.0/images/aliases/testimage | jq -r .target)"
+    testimage_fingerprint_prefix="$(echo "${testimage_fingerprint}" | cut -c1-12)"
+    [ "$(complete init '')" = "${testimage_fingerprint_prefix},images:,localhost:,testimage,ubuntu-daily:,ubuntu-minimal-daily:,ubuntu-minimal:,ubuntu:" ] # Init should show images from the default remote, and all image remotes
```
simondeziel, line 97: ATM, `lxc init` and `lxc launch` do not propose completing with fingerprints of local images. I believe the current behaviour is better as fingerprints are not so human friendly.

Having the alias (`testimage`) is good though but that's already the case now.

test/suites/completions.sh
```diff
@@ -0,0 +1,330 @@
...
+    [ "$(complete move l)" = 'localhost:' ]
+    [ "$(complete move c)" = 'c1,c2' ]
+    [ "$(complete pause '')" = 'c2,localhost:' ] # Only c2 running
+    [ "$(complete publish '')" = 'c1,c2,localhost:' ]
+    [ "$(complete publish l)" = 'localhost:' ]
+    [ "$(complete publish c)" = 'c1,c2' ]
+    [ "$(complete publish c1/s)" = 'c1/snap0' ]
+    [ "$(complete publish c1/)" = 'c1/snap0' ]
+    [ "$(complete rebuild '')" = "${testimage_fingerprint_prefix},images:,localhost:,testimage,ubuntu-daily:,ubuntu-minimal-daily:,ubuntu-minimal:,ubuntu:" ]
+    [ "$(complete rebuild u)" = "ubuntu-daily:,ubuntu-minimal-daily:,ubuntu-minimal:,ubuntu:" ]
```
simondeziel, line 113: Can we actually rebuild a given image fingerprint?

Simplestreams remote should also be filtered out IMHO:

```
$ lxc rebuild ubuntu:noble --empty
Error: The remote isn't a private LXD server
```

test/suites/completions.sh
```diff
@@ -0,0 +1,330 @@
...
+    [ "$(complete config device add c1 devname infiniband '')" = 'hwaddr=,mtu=,name=,nictype=,parent=' ]
+    [ "$(complete config device add c1 devname pci '')" = 'address=' ]
+    [ "$(complete config device add c1 devname proxy '')" = 'bind=,connect=,gid=,listen=,mode=,nat=,proxy_protocol=,security.,uid=' ]
+    [ "$(complete config device add c1 devname tpm '')" = 'path=,pathrm=' ]
+    [ "$(complete config device add c1 devname unix-block '')" = 'gid=,major=,minor=,mode=,path=,required=,source=,uid=' ]
+    [ "$(complete config device add c1 devname unix-char '')" = 'gid=,major=,minor=,mode=,path=,required=,source=,uid=' ]
+    [ "$(complete config device add c1 devname unix-hotplug '')" = 'gid=,mode=,ownership.,productid=,required=,subsystem=,uid=,vendorid=' ]
+    [ "$(complete config device add c1 devname usb '')" = 'busnum=,devnum=,gid=,mode=,productid=,required=,serial=,uid=,vendorid=' ]
+    [ "$(complete config device add c1 devname nic '')" = 'network=,nictype=' ]
+    [ "$(complete config device add c1 devname nic network=)" = "$(lxc query /1.0/networks?recursion=1 | jq -r '["network="+.[].name] | sort | @csv | sub("\"";"";"g")')" ]
```
simondeziel, line 216: Color me impressed ;)

simondeziel (commented):
Wow, such a nice improvement and massive testing, I love it!

I think you know already but there seems to be some pollution from an earlier test causing the `lxc config unset` test to [fail](https://github.com/canonical/lxd/actions/runs/14385268668/job/40339586265#step:14:8689):

```
 ++ /home/runner/go/bin/lxc __complete config unset ''
++ sort
++ grep -vP '^:\d+'
++ tr '\n' ,
++ sed 's/.$//'
+ '[' c1,c2,core.https_address,core.storage_buckets_address,localhost:,user.foo = c1,c2,core.https_address,localhost:,user.foo ']'
```

I think it's due to `test_tls_restrictions` in `test/suites/tls_restrictions.sh` calling `create_object_storage_pool s3`.

simondeziel:
> I think it's due to `test_tls_restrictions` in `test/suites/tls_restrictions.sh` calling `create_object_storage_pool s3`.

Better to instead fix the corresponding `destroy` function, https://github.com/canonical/lxd/pull/15365

test/suites/completions.sh
markylaing: Ok in that case I'll revert to only using aliases :)

markylaing:
> > I think it's due to `test_tls_restrictions` in `test/suites/tls_restrictions.sh` calling `create_object_storage_pool s3`.
> 
> Better to instead fix the corresponding `delete` function, #15365

Ahh thanks!

test/suites/completions.sh
```diff
@@ -0,0 +1,330 @@
...
+    # TODO: This can probably be performed more concisely with awk or some other tool.
+    "${cmd}" __complete "${@}" 2>/dev/null | sort | grep -vP '^:\d+' | tr '\n' ',' | sed 's/.$//'
+}
+
+completion_directive() {
+  # Can't use lxc function as it adds other arguments (like the --verbose flag).
+  cmd=$(unset -f lxc; command -v lxc)
+
+  # Run the completion and grep for the directive.
+  "${cmd}" __complete "${@}" 2>/dev/null | grep -P '^:\d+'
```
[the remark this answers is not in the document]
markylaing: There is only one `:[0-9]`, however it is always the last thing to be outputted, so perhaps just `tail -n 1` would be better

test/suites/completions.sh
```diff
@@ -0,0 +1,330 @@
...
+    [ "$(complete remote switch u)" = 'ubuntu,ubuntu-daily,ubuntu-minimal,ubuntu-minimal-daily' ]
+    [ "$(complete image list '')" = 'images:,localhost:,ubuntu-daily:,ubuntu-minimal-daily:,ubuntu-minimal:,ubuntu:' ] # Image list should show all remotes except the default
+    [ "$(complete image list i)" = 'images:' ]
+    [ "$(complete list '')" = 'localhost:' ] # Instance list should show only instance server remotes and not the default
+    [ "$(complete list l)" = 'localhost:' ]
+    lxc remote switch localhost
+    [ "$(complete remote remove '')" = 'images' ] # The old default is "local", which can't be removed, so is omitted
+    [ "$(complete remote remove u)" = '' ]
+    [ "$(complete remote remove i)" = 'images' ]
+    [ "$(complete remote rename '')" = 'images,localhost' ] # Only images and localhost remotes may be renamed
```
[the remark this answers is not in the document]
markylaing: I think we need to mark them all as static and prevent deletion and rename of all of them

test/suites/completions.sh
```diff
@@ -0,0 +1,330 @@
...
+    [ "$(complete remote rename i)" = 'images' ]
+    [ "$(complete remote set-url '')" = 'images,localhost' ] # Only URLs of images and localhost remotes may be changed
+    [ "$(complete remote set-url l)" = 'localhost' ]
+    [ "$(complete remote switch '')" = 'images,local,ubuntu,ubuntu-daily,ubuntu-minimal,ubuntu-minimal-daily' ] # Can switch to any remote
+    [ "$(complete remote switch u)" = 'ubuntu,ubuntu-daily,ubuntu-minimal,ubuntu-minimal-daily' ]
```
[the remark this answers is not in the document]
markylaing: I thought this as well, but then I tried:
```
lxc remote switch images
lxc init alpine/edge local:c1
```
and it worked fine.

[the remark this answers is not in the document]
markylaing: We can continue to allow it I guess but not suggest it. I'll filter the public remotes from the completions :+1:

test/suites/completions.sh
markylaing: Do you think it's worth keeping the fingerprints for e.g. `image edit` completions? Or should we only ever offer completions for aliases? cc @tomponline

test/suites/completions.sh
```diff
@@ -0,0 +1,330 @@
+complete() {
+    # Can't use lxc function as it adds other arguments (like the --verbose flag).
+    cmd=$(unset -f lxc; command -v lxc)
+
+    # Run the completion and sort the output alphabetically, ignore stderr and the ":\d+" directive,
+    # Then convert to a comma-separated list, removing the last comma.
+    # TODO: This can probably be performed more concisely with awk or some other tool.
+    "${cmd}" __complete "${@}" 2>/dev/null | sort | grep -vP '^:\d+' | tr '\n' ',' | sed 's/.$//'
```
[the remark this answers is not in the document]
markylaing: I don't think it's necessarily worth the effort, I'm guessing bash is also formatting the output into columns and so forth as well.

test/suites/completions.sh
simondeziel: oh, my bad, I didn't understand how rebuild worked. You indeed need to provide a remote for the source of the image to use :/

test/suites/completions.sh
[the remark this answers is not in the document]
simondeziel: Ack, sounds good to keep that bit into `bash`. Thanks

test/suites/completions.sh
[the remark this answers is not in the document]
simondeziel: `tail -n1` is even better then.

test/suites/completions.sh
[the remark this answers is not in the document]
simondeziel: Indeed, it makes sense to continue allowing it. Filtering it will avoid possible confusion so I like your idea.

test/suites/completions.sh
markylaing: I have updated the `cmpImages` function to only return image fingerprints when:
- there are no aliases for the image
- the image is not cached
- the remote is private

I think this is best of both, since users will never see fingerprint completions for public remotes (where aliases should be expected), nor will they see fingerprint completions when the image has a more friendly name elsewhere (e.g. they won't see the fingerprint of cached images from `images:`), but they will see their own images if they haven't created an alias for them yet.

simondeziel: Sounds like a good heuristic to me, thanks!

test/suites/completions.sh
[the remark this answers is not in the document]
markylaing: I've used `tail -n 1` for this :)

test/suites/completions.sh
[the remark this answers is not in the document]
markylaing: TIL about `paste`, thanks! I'm also using `head -n -1` to omit the `:[0-9]` directive now.

test/suites/completions.sh
markylaing: *takes a bow* :laughing:

test/suites/completions.sh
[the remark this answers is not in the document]
simondeziel: I like the `tail`/`head` symmetry, thx.

lxc/config/remote.go
```diff
@@ -52,20 +51,95 @@ func (c *Config) ParseRemote(raw string) (remoteName string, resourceName string
...
 
+	args := defaultArgs
+	if inArgs != nil {
+		args = mergeConnectionArgs(*defaultArgs, inArgs)
+	}
+
 	return c.connectRemote(*remote, args)
 }
 
+func mergeConnectionArgs(defaultArgs lxd.ConnectionArgs, customisableArgs *lxd.ConnectionArgs) *lxd.ConnectionArgs {
```
tomponline, line 79: why is customisableArgs a pointer?

tomponline: also we dont use compacted argument types.

lxc/completion.go
tomponline, line 1: RE

>    ii. Use metadata API and config key "condition" to determine
       key applicability, rather than importing `lxd/instancetype`.

This was iterated on a few times with @kadinsayani , and I believe the `condition` field in the metdata API is not suitable for this use case because not all keys are suitably marked and the field is not well enough defined.

So I think in this case, although not desirable, we do need to build the CLI against the lxd/instancetype keys.

lxc/config.go
```diff
@@ -86,6 +86,43 @@ func (c *cmdConfig) command() *cobra.Command {
 	return cmd
 }
 
+func (c *cmdConfig) configSetValidArgsFunc(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
tomponline, line 89: please add a doc block to this function explaining what it does.

lxc/config_device.go
```diff
@@ -65,6 +65,99 @@ func (c *cmdConfigDevice) command() *cobra.Command {
 	return cmd
 }
 
+func (c *cmdConfigDevice) deviceAddValidArgsFunc(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
tomponline, line 68: please add a doc block

tomponline (changes requested):
see comments thanks

lxc/config/remote.go
markylaing: compacted argument types?

lxc/completion.go
markylaing: In that case can we define the field more concretely? Or if not shall we move `lxd/instancetype` under `shared/instancetype` since ultimately this is the API surface of instance configuration

lxc/config/remote.go
tomponline:
```go
defaultArgs lxd.ConnectionArgs, customisableArgs *lxd.ConnectionArgs
```

rather than

```go
defaultArgs lxd.ConnectionArgs  *lxd.ConnectionArgs, customisableArgs *lxd.ConnectionArgs
```

and anyway in this case we cant use them because we should use:

```go
defaultArgs lxd.ConnectionArgs  *lxd.ConnectionArgs, customisableArgs lxd.ConnectionArgs
```

tomponline: @markylaing is there a more standard name for that?

lxc/completion.go
tomponline: > In that case can we define the field more concretely? 

Perhaps but we have other higher priorities right now.

>  Or if not shall we move lxd/instancetype under shared/instancetype since ultimately this is the API surface of instance configuration

No, because it changes per LXD server version.

tomponline: @markylaing this is ground hog day conversation I had with @kadinsayani 

https://github.com/canonical/lxd/pull/14684#discussion_r1892453313

tomponline:
I'd like @kadinsayani review of this before merging please.

lxc/config/remote.go
markylaing: I'm sorry I'm still confused :confused: 

The first option you've mention is what is in the PR. The the second and third options don't look valid to me?

lxc/completion.go
markylaing: Righto. The difference now is that we do have tests. I'll revert to `shared/instancetype` for now, but would like to take a little time on this early next cycle.

lxc/config/remote.go
tomponline: updated it to make sense, i was rushing as need to work on other higher prio items

tomponline: its quite an odd function as its merging 2 structs,but one is pointer, suggesting its going to be modified, but it isn't, and the one that might be modified (the defaults) isn't a pointer, suggesting it won't be modified, but it is, internally anyway, and then it returns a modified copy of default args.

Quite for a code reviewer.

lxc/config/remote.go
```diff
@@ -52,20 +51,95 @@ func (c *Config) ParseRemote(raw string) (remoteName string, resourceName string
 
 // GetInstanceServer returns a lxd.InstanceServer for the remote with the given name.
 func (c *Config) GetInstanceServer(name string) (lxd.InstanceServer, error) {
+	return c.GetInstanceServerWithAdditionalConnectionArgs(name, nil)
+}
+
+// GetInstanceServerWithAdditionalConnectionArgs returns a lxd.InstanceServer for the remote with the given name. Any
+// populated fields of the given connection arguments override the default connection arguments for the remote.
+func (c *Config) GetInstanceServerWithAdditionalConnectionArgs(name string, inArgs *lxd.ConnectionArgs) (lxd.InstanceServer, error) {
```
tomponline, line 59: This function name seems excessively long, almost Java-esque.

How about just `GetInstanceServerWithConnectionArgs`, the "additional" is clear because the caller is passing them in.

lxc/config/remote.go
markylaing: Right I understand. Will fix in follow up after I've addressed https://github.com/canonical/lxd/pull/15311#discussion_r2044166402

markylaing:
@tomponline I've reverted to using `lxd/instancetype` for the config key completions and have created a ticket in my backlog to come back and address the other comments

lxc/completion.go
```diff
@@ -10,81 +11,204 @@ import (
...
+	return func(option string) {
+			trimmed, found := strings.CutPrefix(option, toComplete)
+			if !found {
+				return // Does not have prefix
+			}
+
+			var key string
+			remaining, _, ok := strings.Cut(trimmed, ".")
+			if ok {
+				// The option contains a `.<something>` after the completion, append only up to the '.'
```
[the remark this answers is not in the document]
kadinsayani: Currently, we complete `limits.cpu` as follows:

```bash
❯ lxc config set c1 limits.cpu
limits.cpu            limits.cpu.allowance  limits.cpu.nodes      limits.cpu.priority
```

lxc/completion.go
```diff
@@ -10,81 +11,204 @@ import (
...
-	if err != nil {
-		return nil, cobra.ShellCompDirectiveError
+	directive := cobra.ShellCompDirectiveNoFileComp
+	if !strings.Contains(toComplete, ":") {
+		filters := instanceServerRemoteCompletionFilters(*g.conf)
+		if shared.ValueInSlice(entityType, []string{"image", "image_alias"}) {
+			filters = imageServerRemoteCompletionFilters(*g.conf)
+		}
+
+		remotes, directives := g.cmpRemotes(toComplete, ":", true, filters...)
```
kadinsayani, line 137:
```suggestion
		remotes, directives := g.cmpRemotes(toComplete, ":", false, filters...)
```

I don't think we should include a `cobra.ShellCompDirectiveNoSpace` directive for every top level resource. For example, when completing `lxc config set c[TAB]` where `c1` is a running instance, `lxc config set c1 ` should be completed rather than `lxc config set c1`. This allows us to continue tab comleting possible instance keys without needing to enter a `[SPACE]`.

lxc/completion.go
```diff
@@ -10,81 +11,204 @@ import (
...
+
+			var key string
+			remaining, _, ok := strings.Cut(trimmed, ".")
+			if ok {
+				// The option contains a `.<something>` after the completion, append only up to the '.'
+				key = toComplete + remaining + "."
+			} else {
+				// The option does not contain any more `.<something>'s after the completion, append the suffix.
+				key = option + suffix
+			}
```
kadinsayani, line 203:
```suggestion
			var key string
			before, after, ok := strings.Cut(trimmed, ".")
			if ok {
				// The option contains a `.<something>` after the completion, append only up to the '.'
				key = toComplete + remaining + "."
			} else if after == "*" {
			        // The option ends with "*", don't append suffix.
			        key = option
			} else {
				// The option does not contain any more `.<something>'s after the completion, append the suffix.
				key = option + suffix
			}
```

We shouldn't append the suffix for keys such as `linux.sysctl.`, or else we end up completing `linux.sysctl.=`.

lxc/completion.go
```diff
@@ -10,81 +11,204 @@ import (
...
+}
 
-	cluster, _, err := resource.server.GetCluster()
-	if err != nil || !cluster.Enabled {
-		return nil, cobra.ShellCompDirectiveError
-	}
+// topLevelInstanceServerResourceNameFuncs is a map of functions that can return LXD API resource names without any arguments.
+// This is used when returning completions for arguments like `<remote>:<name>` where the remote is an instance server.
+var topLevelInstanceServerResourceNameFuncs = map[string]func(server lxd.InstanceServer) ([]string, error){
+	"certificate": func(server lxd.InstanceServer) ([]string, error) {
```
kadinsayani, line 45: Can we represent top level resources as constants? That way we can use the constants rather than strings at all call sites.

kadinsayani (changes requested):
Looks really good overall. Thanks for your work on this.

Just a few nits :)

lxc/completion.go
markylaing: Done.

lxc/completion.go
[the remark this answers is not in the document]
markylaing: Yes this can be a future improvement but showing config option prefixes is good enough for now.

lxc/completion.go
markylaing: Fixed.

lxc/completion.go
markylaing: Discussed this yesterday. I'm happy to either a) just use the entities from the `shared/entity` package, or b) add some new consts to `lxc`.

The reason I didn't do this initially is that the plan will ultimately be to remove `shared/entity` entirely and rely on the metadata API

markylaing: For either case, I'm happy to address in a follow up.

simondeziel (approved):

lxc/completion.go
```diff
@@ -612,10 +609,26 @@ func (g *cmdGlobal) cmpInstanceAllDevices(instanceName string) ([]string, cobra.
...
+		if !strings.HasPrefix(key, "device-") {
+			continue
+		}
+
+		parts := strings.Split(key, "-")
+		deviceType := parts[1]
+
+		// "unix" is not a device, get the next part.
+		if deviceType == "unix" && len(parts) > 2 {
+			// The metadata API has "unix-usb", but the device type is just "usb"
```
tomponline, line 621: is this a metadata API error, should we fix that rather than paper over it here?

lxc/config_device.go
tomponline: not done yet

tomponline (commented):
@markylaing Im finding this PR quite tricky to grok given the size and amount of changes occurring.

Is it possible to break this PR up to focus first on the de-duplication changes, and then on the fixes/additional completions?

kadinsayani (approved):

kadinsayani:
Happy to provide another review when this PR is split up. 

The changes we discussed yesterday LGTM :)

markylaing:
I'll break this up and submit separate PRs.

Issue #15313, opened 2025-04-02
downgrading a cluster within an LTS channel fails


tjwsl:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

24.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
root@case-repro-1:~# snap list --all lxd core20 core22 core24 snapd
Name    Version         Rev    Tracking       Publisher   Notes
core22  20250210        1802   latest/stable  canonical✓  base
core24  20241217        739    latest/stable  canonical✓  base
lxd     5.21.2-084c8c8  31214  5.21/stable    canonical✓  disabled,in-cohort,held
lxd     5.21.3-c5ae129  33110  5.21/stable    canonical✓  in-cohort,held
snapd   2.67.1          23771  latest/stable  canonical✓  snapd
```

### Output of "lxc info" or system info if it fails

```shell
config:
  cluster.https_address: 10.1.2.41:8443
  core.https_address: 10.1.2.41:8443
api_extensions:
- storage_zfs_remove_snapshots
- container_host_shutdown_timeout
- container_stop_priority
- container_syscall_filtering
- auth_pki
- container_last_used_at
- etag
- patch
- usb_devices
- https_allowed_credentials
- image_compression_algorithm
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
- unix_device_rename
- storage_lvm_use_thinpool
- storage_rsync_bwlimit
- network_vxlan_interface
- storage_btrfs_mount_options
- entity_description
- image_force_refresh
- storage_lvm_lv_resizing
- id_map_base
- file_symlinks
- container_push_target
- network_vlan_physical
- storage_images_delete
- container_edit_metadata
- container_snapshot_stateful_migration
- storage_driver_ceph
- storage_ceph_user_name
- resource_limits
- storage_volatile_initial_source
[...]

PR #15314, opened 2025-04-03
Cluster: Support downgrades within point releases of an LXD LTS series

tomponline:
Fixes https://github.com/canonical/lxd/issues/15313

When a cluster member is downgraded between point releases and has had newer API extensions previously, the API extension counter was not being updated to reflect the new reduced API extension count due to the scenario of a member being behind the others not completing the DB transaction and so not recording the update of the current member's version info in the global database.

Subsequently even downgrading all members within a cluster was not sufficient to get an operational cluster again.

This PR removes this restriction by having LXD wait until all cluster member versions are aligned, be they older or newer.

It also moves away from using several different sentinel errors to use the convention we already have of using `api.StatusError` with an error class of `StatusPreconditionFailed` for when the upgrade requirements havent been met.

Finally this PR also re-introduces `lxc cluster list` member status info during a version change.

Members now report either "LXD version is older than other members" or "LXD version is newer than other members" with a "blocked" state.


tomponline:
This PR still needs work as we need to ensure schema version isnt downgraded.

lxd/daemon.go
```diff
@@ -1478,7 +1478,7 @@ func (d *Daemon) init() error {
 		if err == nil {
 			logger.Info("Initialized global database")
 			break
-		} else if errors.Is(err, db.ErrSomeNodesAreBehind) {
+		} else if api.StatusErrorCheck(err, http.StatusUpgradeRequired) {
```
MggMuggins, line 1481: To clarify; we still fail on `http.StatusPreconditionFailed` (another cluster member's schema/api version was greater). So we should still expect the daemon to refuse to start until the entire cluster has been downgraded? Wouldn't that prevent the other cluster members from updating their `nodes` entries due to dqlite loss of quorum?

With members c1, c2 and c3, downgrade c1. An admin will need to power ahead with the downgrade of c2 with c1 refusing to start. Once c1 and c2 have been started, they won't be able to establish quorum (since `err` here will still be `http.StatusPreconditionFailed`), so there's no way for the third cluster member to update its schema entry.

MggMuggins (changes requested):
One clarification; likely that I'm incorrect there but want to be sure.

One more thing (shouldn't necessarily block this): I find the use of the http error codes here pretty opaque. I'm wondering if it wouldn't make more sense to define an enum for the return value of [`CompareVersions`](https://github.com/canonical/lxd/blob/main/lxd/util/version.go#L18), something like `SchemaEqual`, `SchemaLess`, and `SchemaGreater`; those would likely make at least as much sense as the http error codes, and could replace the `if`/`switch` statements over the integer return values too.

lxd/daemon.go
tomponline: Not quite.

Today LXD still restarts when it is behind other cluster members, it just does so in a way that means the query it ran against the database to update its `nodes` entry with its updated version isn't persisted. This means the cluster never converges.

However the database itself is up and running and quorum has been reached, otherwise we wouldn't be able to run any queries against it.

markylaing (approved):
Looks good to me!

We have tests for `clustering_upgrade` (though last I recall these were being skipped due to flakiness. Do you think it would be worth adding a `clustering_downgrade` test?

tomponline:
> Looks good to me!
> 
> We have tests for `clustering_upgrade` (though last I recall these were being skipped due to flakiness. Do you think it would be worth adding a `clustering_downgrade` test?

I'm actually working on getting them running again now :)

tomponline:
> There's more work we can do here in the future with regard to `lxc cluster list` reporting that some members are in a pending state (which it currently doesn't do even for normal upgrades), but it outside of the scope of this issue.

I'm actually working on improving this too.

lxd/db/cluster/open.go
```diff
@@ -289,18 +278,16 @@ func checkClusterIsUpgradable(ctx context.Context, tx *sql.Tx, target [2]int) er
...
 			// and wait for other nodes to be upgraded and
 			// restarted.
-			return errSomeNodesAreBehind
+			return api.StatusErrorf(http.StatusPreconditionFailed, "A cluster member's version (%v) is behind this cluster member's version (%v), please ensure versions match", version, target)
 		case 2:
 			// Another node has a version greater than ours
-			// and presumeably is waiting for other nodes
+			// and presumably is waiting for other nodes
 			// to upgrade. Let's error out and shutdown
 			// since we need a greater version.
```
MggMuggins, line 286: I don't think this comment is accurate anymore, since LXD now waits for an upgrade notification in this case as well.

MggMuggins (approved):
This looks great! Thanks for the tests. One nit but no blockers.

lxd/db/cluster/open.go
tomponline: Ah yeah ill do a follow up pr

tomponline: https://github.com/canonical/lxd/pull/15322

PR #15316, opened 2025-04-03
github: enable code coverage for client tests

simondeziel:


.github/workflows/tests.yml
```diff
@@ -416,60 +416,78 @@ jobs:
...
 
+      - name: Make GOCOVERDIR
+        run: mkdir -p "${GOCOVERDIR}"
+        if: env.GOCOVERDIR != ''
+
       - name: Build static lxc (x86_64)
         env:
           CGO_ENABLED: 0
           GOARCH: amd64
+          EXTRAS: ${{ ( env.GOCOVERDIR != '' ) && '-cover' || '-ldflags "-s -w"' }}
```
tomponline: is this making ldflags optional?

simondeziel: It's either coverage build with full debug symbols OR stipped binaries.

So yes.

simondeziel: It is on purpose as I think coverage info may need debug symbols? If you know for sure they don't I'm happy to tweak that.

tomponline (approved):

PR #15317, opened 2025-04-04
Refactor devlxd API

MusicDin:
Refactors the devlxd by introducing the API endpoints and register functions (similar to LXD REST API).

Additionally, it adds devlxd client that can be used in other tools for interacting with the devlxd endpoints from virtual machines and containers. This PR uses it test devlxd binary.

### DIFF
--- a/client/devlxd.go
+++ b/client/devlxd.go
@@ -31,6 +31,8 @@ type ProtocolDevLXD struct {
 	httpBaseURL   url.URL
 	httpUnixPath  string
 	httpUserAgent string
+
+	eventListenerManager *eventListenerManager
 }
 
 // GetConnectionInfo returns the basic connection information used to interact with the server.
--- a/client/devlxd.go
+++ b/client/devlxd.go
@@ -62,6 +62,14 @@ func (r *ProtocolDevLXD) Disconnect() {
 	r.ctxConnectedCancel()
 }
 
+// RawQuery allows directly querying the devLXD.
+//
+// This should only be used by internal LXD tools.
+func (r *ProtocolDevLXD) RawQuery(method string, path string, data any, ETag string) (*api.DevLXDResponse, string, error) {
+	url := r.httpBaseURL.String() + path
+	return r.rawQuery(method, url, data, ETag)
+}
+
 // rawQuery is a method that sends HTTP request to the devLXD with the provided
 // method, URL, data, and ETag. It processes the request based on the data's
 // type and handles the HTTP response, returning parsed results or an error
--- a/client/devlxd.go
+++ b/client/devlxd.go
@@ -1,0 +1,187 @@
+package lxd
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+
+	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/logger"
+	"github.com/canonical/lxd/shared/version"
+)
+
+// ProtocolDevLXD represents a devLXD API server.
+type ProtocolDevLXD struct {
+	ctx context.Context
+
+	// Context related to the current connection.
+	ctxConnected       context.Context
+	ctxConnectedCancel context.CancelFunc
+
+	// HTTP client information.
+	http          *http.Client
+	httpBaseURL   url.URL
+	httpUnixPath  string
+	httpUserAgent string
+}
+
+// GetConnectionInfo returns the basic connection information used to interact with the server.
+func (r *ProtocolDevLXD) GetConnectionInfo() (*ConnectionInfo, error) {
+	return &ConnectionInfo{
+		Protocol:   "devlxd",
+		URL:        r.httpBaseURL.String(),
+		SocketPath: r.httpUnixPath,
+	}, nil
+}
+
+// GetHTTPClient returns the http client used for the connection. This can be used to set custom http options.
+func (r *ProtocolDevLXD) GetHTTPClient() (*http.Client, error) {
+	if r.http == nil {
+		return nil, fmt.Errorf("HTTP client isn't set, bad connection")
+	}
+
+	return r.http, nil
+}
+
+// DoHTTP performs a Request.
+func (r *ProtocolDevLXD) DoHTTP(req *http.Request) (resp *http.Response, err error) {
+	// Set the user agent
+	if r.httpUserAgent != "" {
+		req.Header.Set("User-Agent", r.httpUserAgent)
+	}
+
+	return r.http.Do(req)
+}
+
+// Disconnect is a no-op for devLXD.
+func (r *ProtocolDevLXD) Disconnect() {
+	r.ctxConnectedCancel()
+}
+
+// rawQuery is a method that sends HTTP request to the devLXD with the provided
+// method, URL, data, and ETag. It processes the request based on the data's
+// type and handles the HTTP response, returning parsed results or an error
+// if it occurs.
+func (r *ProtocolDevLXD) rawQuery(method string, url string, data any, ETag string) (devLXDResp *api.DevLXDResponse, etag string, err error) {
+	var req *http.Request
+
+	// Log the request
+	logger.Debug("Sending request to devLXD", logger.Ctx{
+		"method": method,
+		"url":    url,
+		"etag":   ETag,
+	})
+
+	// Get a new HTTP request setup
+	if data != nil {
+		// Encode the provided data
+		buf := bytes.Buffer{}
+		err := json.NewEncoder(&buf).Encode(data)
+		if err != nil {
+			return nil, "", err
+		}
+
+		// Some data to be sent along with the request
+		// Use a reader since the request body needs to be seekable
+		req, err = http.NewRequestWithContext(r.ctx, method, url, bytes.NewReader(buf.Bytes()))
+		if err != nil {
+			return nil, "", err
+		}
+
+		// Set the encoding accordingly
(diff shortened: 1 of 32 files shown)


shared/api/devlxd.go
```diff
@@ -1,5 +1,20 @@
...
+	"encoding/json"
+)
+
+// DevLXDResponse represents the response from the devLXD API.
+type DevLXDResponse struct {
+	Content    []byte `json:"content" yaml:"content"`
+	StatusCode int    `json:"status_code" yaml:"status_code"`
+}
+
+// ContentAsStruct parses the Response metadata into a provided struct.
```
MusicDin, line 13: Fix comment -> Unmarshals the response content.
Do the same on api.Response - MetadataAsStruct

client/devlxd_events.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/api"
+)
+
+// DevLXDEventListener is a wrapper around the EventListener struct
+// that is used to interact with the devLXD event stream.
+type DevLXDEventListener struct {
+	EventListener
+	r *ProtocolDevLXD
+}
```
MusicDin, line 18: I wanted to reuse the EventListener here, but it accepts the ProtocolLXD, so I wrapped it DevLXDEventListener and overrid the Disconnect function, which is the only function that uses it.

I don't like this, because ProtocolLXD can be used later used in other functions, which would panic if the ProtocolLXD is not set in EventListener.

We could pass the disconnect function when creating event listener, or  introduce some kind of "EventServer" - either as an embedded struct that contains event listeners, or an interface that allows adding, modifying, and retrieving event listeners.

Any thoughts?

client/devlxd_images.go
```diff
@@ -0,0 +1,106 @@
...
+	"mime/multipart"
+	"net/http"
+	"strings"
+
+	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/version"
+)
+
+// ExportImage exports the image that is cached in host's LXD to the guest LXD.
+func (r *ProtocolDevLXD) ExportImage(fingerprint string, receiveFileFunc func(filename string, content io.Reader) error) error {
```
MusicDin, line 17: I was not entirely sure how to handle image export. The unified image formats are transferred as octet-stream, which means the function could accept the writer and copy the content to it. But for split images, they are transferred as multipart http, so the writer would not be sufficient.

I went with a `receiveFileFunc` which is called for each file (if the func is provided). Caller can then decide what to do with it, but maybe there is a cleaner approach that I am missing.

MusicDin: Comment: Check `lxdImageDownload`

client/devlxd_events.go
MusicDin: Introduce EventListenerManager to get rid of parent reference in EventListener

client/devlxd_images.go
MusicDin: The `lxdImageDownload` handles both octet-stream and multipart http. The client now simply calls the function and returns the same as `(ProtocolLXD).GetImageFile`.

lxd/api.go
tomponline, line 1: Would a `devlxd` package be useful to move these utilities into, and avoid the repetition of "DevLXD" as a prefix?

lxd/api_devlxd.go
tomponline, line 1: For `lxd/devlxd: Remove http.Writer from handler func` what does the `/1.0/events` handler in the main LXD API do here? Use ManualResponse too?

tomponline, line 1: `lxd/devlxd: Pass request instance using request context` commit seems to have quite a few formatting changes unrelated to commit message, making it harder to review.

lxd/devlxd.go
```diff
@@ -458,7 +458,17 @@ func registerDevLXDEndpoint(d *Daemon, apiRouter *mux.Router, apiVersion string,
 		// Set devLXD auth method to identify this request as coming from the /dev/lxd socket.
 		request.SetCtxValue(r, request.CtxProtocol, auth.AuthenticationMethodDevLXD)
 
-		handleRequest := func(action devLXDAPIEndpointAction) response.Response {
+		handleRequest := func(action devLXDAPIEndpointAction) (resp response.Response) {
+			// Handle panic in the handler.
+			defer func() {
```
tomponline, line 458: @MusicDin do the main API handlers have this too?

shared/api/devlxd.go
```diff
@@ -1,5 +1,20 @@
 package api
 
+import (
+	"encoding/json"
+)
+
+// DevLXDResponse represents the response from the devLXD API.
+type DevLXDResponse struct {
```
tomponline, line 8: This will need an API extension then please

client/devlxd_state.go
```diff
@@ -0,0 +1,29 @@
...
+	_, err := r.queryStruct(http.MethodGet, "", nil, "", &info)
+	if err != nil {
+		return nil, err
+	}
+
+	return &info, nil
+}
+
+// PatchState patches the guest state.
+func (r *ProtocolDevLXD) PatchState(req api.DevLXDPut) error {
```
tomponline, line 22: hrm, the function names in the other client functions dont mention HTTP verbs, but rather use `Update` verbs?
This seems a little inconsistent.

tomponline: infact the client package today doesnt mention "patch" at all.

tomponline: Also we accept an argument for DevLXDPut but the function name is Patch... super confusing.

client/devlxd.go
```diff
@@ -163,6 +167,73 @@ func (r *ProtocolDevLXD) queryStruct(method string, urlPath string, data any, ET
 	return etag, nil
 }
 
+// RawWebsocket allows directly connection to LXD API websockets
```
tomponline, line 172: "allows directly"

client/devlxd_images.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"fmt"
+
+	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/version"
+)
+
+// ExportImage exports the image with a given fingerprint from the host's LXD.
+// Note that only cached and public images can be exported.
+func (r *ProtocolDevLXD) ExportImage(fingerprint string, req ImageFileRequest) (*ImageFileResponse, error) {
```
tomponline, line 12: is ExportImage the right name here? What about `GetImageFile` which seems to align more with the existing client package in terms of function signature.

lxd/devlxd.go
MusicDin: Yes they do. The only benefit I see using this is that the error is nicer and in `devLXD` format.

shared/api/devlxd.go
MusicDin: It's existing structure, the responses are the same as they were previously

lxd/devlxd.go
tomponline: i couldnt see any other calls to `recover()`?

lxd/api_devlxd.go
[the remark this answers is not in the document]
MusicDin: Mechanical, it just abstracts the handlers into endpoint structs.

MusicDin: It uses a custom response, that handles event retrieval in Render function, similar to manual response.

test/suites/devlxd.sh
```diff
@@ -48,33 +48,41 @@ test_devlxd() {
 
   (
     cat << EOF
-metadata:
```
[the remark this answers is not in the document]
tomponline: makes sense ta

lxd/devlxd.go
tomponline: >  Yes it does, so this is not necessary, just seemed suitable.

So a panic in a devlxd handler wouldnt crash the daemon, but this change captures and presents the error more clearly?

lxd/api.go
MusicDin: I think it would, but since we are expecting `Daemon` struct to be passed around, it cannot be accessed from the sub-package. The API handlers (all of them are prefixed with `devLXD`) expect Daemon as input. Personally, I prefer consistency with LXD API, but we may be able to pass just the state around.

lxd/api_devlxd.go
MusicDin: Fixed.

client/devlxd_state.go
MusicDin: Fixed.

lxd/api.go
MusicDin: Still, IMO would be nice to move everything into `devlxd` package, not just the utils.

lxd/devlxd.go
MusicDin: Yes.

Here is an example on a artificial endpoint `/panic` that just panics:

```go
devLXDAPIEndpoint{
	{
		Path: "/panic",
		Get: devLXDAPIEndpointAction{
			Handler: func(d *Daemon, r *http.Request) response.Response {
				panic("This is a test panic")
			},
		},
	},
}
```

With recovery:
```
$ lxc exec c1 -- curl -i --unix-socket /dev/lxd/sock devlxd/1.0/panic
HTTP/1.1 500 Internal Server Error
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Date: Thu, 17 Apr 2025 08:28:08 GMT
Content-Length: 21

This is a test panic
```

Without recovery:
```
$ lxc exec c1 -- curl -i --unix-socket /dev/lxd/sock devlxd/1.0/panic
curl: (52) Empty reply from server
```

PR #15319, opened 2025-04-04
doc: hard code link titles (stable-5.0)

minaelee:
This PR hard-codes all titles for `relatedlinks:` and `discourse:` links that exist in `stable-5.0` docs. This is intended to prevent failed live link title fetching at build time (typically due to external site downtime) from blocking builds.


tomponline (approved):
Cheers!

PR #15324, opened 2025-04-07
Ubuntu Pro: Re-initialise watcher on guest request

markylaing:
The next release of the Ubuntu Pro client will use `/var/lib/ubuntu-pro` for new configuration. This is where the `interfaces/lxd-config.json` file will be placed, which LXD will watch for host configuration updated. Unfortunately, the new configuration directory isn't created until `pro config set lxd_guest_attach <value>` is actually called. Since LXD creates a file watcher on startup, it attempts to watch `/var/lib/ubuntu-pro` and this fails (doesn't exist yet). The only way to get guest attachment working is to restart the LXD daemon.

This PR works around this issue by allowing the guest to re-trigger a file watcher in limited scenarios. The tl;dr is as follows:
1. If the host is not ubuntu, then client is static and no file watcher will ever be triggered.
2. If the host is ubuntu, but no watcher is configured, and the watcher context has not been cancelled, when the guest calls `GET /1.0/ubuntu-pro` or `POST /1.0/ubuntu-pro/token` LXD will attempt to watch `/var/lib/ubuntu-pro` again to get the host config. If that fails, a cooldown of 5 minutes will prevent a guest from requesting either the setting or a token again.

### DIFF
--- a/lxd/devlxd.go
+++ b/lxd/devlxd.go
@@ -351,7 +351,10 @@ func devlxdUbuntuProGetHandler(d *Daemon, c instance.Instance, w http.ResponseWr
 		return response.DevLxdErrorResponse(api.NewGenericStatusError(http.StatusMethodNotAllowed), c.Type() == instancetype.VM)
 	}
 
-	settings := d.State().UbuntuPro.GuestAttachSettings(c.ExpandedConfig()["ubuntu_pro.guest_attach"])
+	settings, err := d.State().UbuntuPro.GuestAttachSettings(c.ExpandedConfig()["ubuntu_pro.guest_attach"])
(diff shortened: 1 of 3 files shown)


tomponline:
@markylaing im going to hold back on this as potentially doesn't solve the problem at hand,e.g.

```
lxc launch ubuntu:25.04 c1 # Will not get correct setting as pro client not upgraded yet
# upgrade pro client
pro config set lxd_guest_attach=on
lxc launch ubuntu:25.04 c2 # Will not get correct setting as rate limited
```

But happy to be told im wrong here :)

tomponline:
@markylaing this can be closed right?

PR #15346, opened 2025-04-07
Block: Fix `DiskFSUUID` for pristine disks

roosterfish:
Whilst backporting https://github.com/canonical/lxd/pull/15310 and testing with an isolated 5.21 LXD with completely clean storage disks I found the blkid command returns exit 2 in case the disk doesn't contain any traces of device identifiers or device contents.

The [man page](https://www.man7.org/linux/man-pages/man8/blkid.8.html) says:

> If the specified token was not found, or no (specified) devices
       could be identified, or it is impossible to gather any information
       about the device identifiers or device content an exit status of 2
       is returned.

### DIFF
--- a/lxd/storage/block/utils.go
+++ b/lxd/storage/block/utils.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


roosterfish:
Interestingly running `blkid` on pristine partitions without FS doesn't return exit code 2. This only applies for the actual disks.

tomponline (approved):

PR #15349, opened 2025-04-08
storage: Support multi volume snapshot and restore

hamistao:
This adds support for snapshotting an instance and its attached volumes while mantaining crash consistency across all relevant volumes.

This follows Phase 1 of the to be published specification: Sync instance and attached volume snapshots. 

Including shared volumes on a multi volume snapshot or restore is not allowed.

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2672,3 +2672,15 @@ This adds support for listing networks across all projects using the `all-projec
 ## `clustering_restore_skip_mode`
 
 Adds a `skip` mode to the restore request. This mode restores a cluster member's status to `ONLINE` without restarting any of its stopped local instances or migrating back instances that were evacuated to other cluster members.
+
+## `instance_snapshots_multi_volume`
+
+This API extension enables creating a snapshot and restoring an instance along with its attached volumes, while guaranteeing crash consistency across volumes. To support this, this extension adds a `disks` field to both `POST /1.0/instances/{name}/snapshots` and
+`PUT /1.0/instances/{name}` with the following possible values:
+
+1. `root`: Represents a snapshot of only the instance root disk. This same behavior is kept if `disks` is unset.
+1. `volumes`: Represents a multi-volume snapshot of the instance root volume and all non-shared attached volumes. Fails if any attached volumes are being shared.
+
+Also adds a `exclude_disks` field to both endpoints to allow a user to specify disk names whose volumes should not be included as part of the multi-volume snapshot.
+
+To track which volumes' snapshots are created together, this extension introduces a new volatile configuration key, {config:option}`instance-volatile:volatile.attached_volumes`, in the configuration of supported storage drivers for instance snapshots. This key contains a JSON-serialized map of attached volume UUIDs to the UUIDs of their corresponding snapshots.
--- a/lxc/snapshot.go
+++ b/lxc/snapshot.go
@@ -20,9 +20,11 @@ import (
 type cmdSnapshot struct {
 	global *cmdGlobal
 
-	flagStateful bool
-	flagNoExpiry bool
-	flagReuse    bool
+	flagStateful     bool
+	flagNoExpiry     bool
+	flagReuse        bool
+	flagDisks        string
+	flagExcludeDisks string
 }
 
 func (c *cmdSnapshot) command() *cobra.Command {
@@ -44,6 +46,8 @@ running state, including process memory state, TCP connections, ...`))
 	cmd.Flags().BoolVar(&c.flagStateful, "stateful", false, i18n.G("Whether or not to snapshot the instance's running state"))
 	cmd.Flags().BoolVar(&c.flagNoExpiry, "no-expiry", false, i18n.G("Ignore any configured auto-expiry for the instance"))
 	cmd.Flags().BoolVar(&c.flagReuse, "reuse", false, i18n.G("If the snapshot name already exists, delete and create a new one"))
+	cmd.Flags().StringVar(&c.flagDisks, "disks", "", i18n.G(`Which disks should be included in the snapshot; can be "root" or "volumes"`))
+	cmd.Flags().StringVar(&c.flagExcludeDisks, "exclude-disks", "", i18n.G(`Comma-separated list of names for which disks should be ignored when performing a multi-volume snapshot`))
 	cmd.ValidArgsFunction = func(cmd *cobra.Command, args []string, toComplete string) ([]cobra.Completion, cobra.ShellCompDirective) {
 		if len(args) > 0 {
 			return nil, cobra.ShellCompDirectiveNoFileComp
@@ -121,8 +125,10 @@ func (c *cmdSnapshot) run(cmd *cobra.Command, args []string) error {
 	}
 
 	req := api.InstanceSnapshotsPost{
-		Name:     snapname,
-		Stateful: c.flagStateful,
+		Name:         snapname,
+		Stateful:     c.flagStateful,
+		DisksMode:    c.flagDisks,
+		ExcludeDisks: shared.SplitNTrimSpace(c.flagExcludeDisks, ",", -1, true),
 	}
 
 	if !stdinData.ExpiresAt.IsZero() {
--- a/lxd/instance/drivers/driver_common.go
+++ b/lxd/instance/drivers/driver_common.go
@@ -324,6 +324,58 @@ func (d *common) Snapshots() ([]instance.Instance, error) {
 	return snapshots, nil
 }
 
+// The volumeKey type is used to identify a volume on a map for getSharedVolumes to efficiently check if
+// each volume on a list is being shared.
+// If the volume under the key is a snapshot, volumeName should be its parent's name so getSharedVolumes
+// can check if the parent is being shared.
+type volumeKey struct {
+	pool       string
+	volumeName string
+}
+
+// getSharedVolumes filters shared volumes from a map of volumes, each volume can be only attached to inst.
+// Useful when checking for many volumes at once, since using VolumeUsedByInstanceDevices would result in a lot of redundancy.
+func (d *common) getSharedVolumes(ctx context.Context, tx *db.ClusterTx, original map[volumeKey]*api.StorageVolume, inst instance.Instance) (shared []*api.StorageVolume, err error) {
+	// Return early if an empty map was provided.
+	if len(original) == 0 {
+		return shared, nil
+	}
+
+	instanceProject := inst.Project()
+	instanceName, _, _ := api.GetParentAndSnapshotName(inst.Name()) // Use the parent name to check for sharing if inst is a snapshot.
+	effectiveProject := project.StorageVolumeProjectFromRecord(&instanceProject, dbCluster.StoragePoolVolumeTypeCustom)
+
+	err = tx.InstanceList(ctx, func(listedInstance db.InstanceArgs, p api.Project) error {
+		// Ignore the provided instance.
+		if instanceProject.Name == listedInstance.Project && instanceName == listedInstance.Name {
+			return nil
+		}
+
+		// Ignore instances with a different effective project for volumes.
+		if effectiveProject != project.StorageVolumeProjectFromRecord(&p, dbCluster.StoragePoolVolumeTypeCustom) {
+			return nil
+		}
+
+		expandedDevices := instancetype.ExpandInstanceDevices(listedInstance.Devices, listedInstance.Profiles)
+
+		// Iterate through each of the instance's devices, looking for disks sourced by volumes attached to inst.
+		for _, dev := range expandedDevices {
+			volume, usesVol := original[volumeKey{dev["pool"], dev["source"]}]
+
+			if !usesVol {
+				continue
+			}
+
+			// Delete the shared volume from the map and move on.
+			shared = append(shared, volume)
+		}
+
+		return nil
+	})
+
+	return shared, err
+}
+
 // VolatileSet sets one or more volatile config keys.
 func (d *common) VolatileSet(changes map[string]string) error {
 	// Quick check.
--- a/lxd/instance/drivers/driver_common.go
+++ b/lxd/instance/drivers/driver_common.go
@@ -538,16 +538,22 @@ func (d *common) getAttachedVolumeSnapshots(sourceSnapshot instance.Instance, di
 // restoreCommon contains the common logic of restoring different instance types.
 // This includes loading the storage pool, deriving which volumes should be restored, stopping the instance
 // and updating instance config.
-func (d *common) restoreCommon(inst instance.Instance, source instance.Instance) (storagePools.Pool, bool, *operationlock.InstanceOperation, error) {
+func (d *common) restoreCommon(inst instance.Instance, source instance.Instance, disks deviceConfig.Devices) ([]*api.StorageVolume, storagePools.Pool, bool, *operationlock.InstanceOperation, error) {
 	// Load the storage driver.
 	pool, err := storagePools.LoadByInstance(d.state, inst)
 	if err != nil {
-		return nil, false, nil, err
+		return nil, nil, false, nil, err
+	}
+
+	// If performing a multi volume restore, get the snapshots used to restore attached volumes.
+	targetSnapshots, err := d.getAttachedVolumeSnapshots(source, disks)
+	if err != nil {
+		return nil, nil, false, nil, fmt.Errorf("Failed to restore snapshot rootfs: %w", err)
 	}
 
 	op, err := operationlock.Create(d.Project().Name, d.Name(), operationlock.ActionRestore, false, false)
 	if err != nil {
-		return nil, false, nil, fmt.Errorf("Failed to create instance restore operation: %w", err)
+		return nil, nil, false, nil, fmt.Errorf("Failed to create instance restore operation: %w", err)
 	}
 
 	// Stop the instance.
@@ -571,7 +577,7 @@ func (d *common) restoreCommon(inst instance.Instance, source instance.Instance)
 			err := inst.Update(args, false)
 			if err != nil {
 				op.Done(err)
-				return nil, false, nil, err
+				return nil, nil, false, nil, err
 			}
 
 			// On function return, set the flag back on.
@@ -585,15 +591,18 @@ func (d *common) restoreCommon(inst instance.Instance, source instance.Instance)
 		err := inst.Stop(false)
 		if err != nil {
 			op.Done(err)
-			return nil, false, nil, err
+			return nil, nil, false, nil, err
 		}
 
 		// Refresh the operation as that one is now complete.
 		op, err = operationlock.Create(d.Project().Name, d.Name(), operationlock.ActionRestore, false, false)
 		if err != nil {
-			return nil, false, nil, fmt.Errorf("Failed to create instance restore operation: %w", err)
+			return nil, nil, false, nil, fmt.Errorf("Failed to create instance restore operation: %w", err)
 		}
 	}
+
+	// "volatile.attached_volumes" should not be included in the instance config.
+	delete(source.LocalConfig(), "volatile.attached_volumes")
 
 	// Restore the configuration.
 	args := db.InstanceArgs{
@@ -613,10 +622,10 @@ func (d *common) restoreCommon(inst instance.Instance, source instance.Instance)
 	err = inst.Update(args, false)
 	if err != nil {
 		op.Done(err)
-		return nil, false, nil, err
+		return nil, nil, false, nil, err
 	}
 
-	return pool, wasRunning, op, nil
+	return targetSnapshots, pool, wasRunning, op, nil
 }
 
 // VolatileSet sets one or more volatile config keys.
--- a/lxd/instance/drivers/driver_common.go
+++ b/lxd/instance/drivers/driver_common.go
@@ -535,6 +535,90 @@ func (d *common) getAttachedVolumeSnapshots(sourceSnapshot instance.Instance, di
 	return slices.Collect[*api.StorageVolume](maps.Values(targetSnapshots)), nil
 }
 
+// restoreCommon contains the common logic of restoring different instance types.
+// This includes loading the storage pool, deriving which volumes should be restored, stopping the instance
+// and updating instance config.
+func (d *common) restoreCommon(inst instance.Instance, source instance.Instance) (storagePools.Pool, bool, *operationlock.InstanceOperation, error) {
+	// Load the storage driver.
+	pool, err := storagePools.LoadByInstance(d.state, inst)
+	if err != nil {
+		return nil, false, nil, err
+	}
+
+	op, err := operationlock.Create(d.Project().Name, d.Name(), operationlock.ActionRestore, false, false)
+	if err != nil {
+		return nil, false, nil, fmt.Errorf("Failed to create instance restore operation: %w", err)
+	}
+
+	// Stop the instance.
+	wasRunning := inst.IsRunning()
+	if wasRunning {
+		ephemeral := d.IsEphemeral()
+		if ephemeral {
+			// Unset ephemeral flag.
+			args := db.InstanceArgs{
+				Architecture: d.Architecture(),
+				Config:       d.LocalConfig(),
+				Description:  d.Description(),
+				Devices:      d.LocalDevices(),
+				Ephemeral:    false,
+				Profiles:     d.Profiles(),
+				Project:      d.Project().Name,
+				Type:         d.Type(),
+				Snapshot:     d.IsSnapshot(),
+			}
+
+			err := inst.Update(args, false)
+			if err != nil {
+				op.Done(err)
+				return nil, false, nil, err
+			}
+
+			// On function return, set the flag back on.
+			defer func() {
+				args.Ephemeral = ephemeral
+				_ = inst.Update(args, false)
+			}()
+		}
+
+		// This will unmount the instance storage.
+		err := inst.Stop(false)
+		if err != nil {
+			op.Done(err)
+			return nil, false, nil, err
+		}
+
+		// Refresh the operation as that one is now complete.
+		op, err = operationlock.Create(d.Project().Name, d.Name(), operationlock.ActionRestore, false, false)
+		if err != nil {
+			return nil, false, nil, fmt.Errorf("Failed to create instance restore operation: %w", err)
+		}
+	}
+
+	// Restore the configuration.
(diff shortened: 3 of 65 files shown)


test/suites/snapshots.sh
```diff
@@ -449,3 +449,97 @@ test_snap_fail() {
...
+
+  lxc exec c1 -- rm snap1
+  lxc exec c1 -- rm /mnt/non-shared/snap1
+  lxc exec c1 -- rm /mnt/shared/snap1
+
+  lxc storage volume delete default non-shared/snap0 # The snapshot taken with c1-snap1 was "non-shared"'s first snapshot.
+
+  # Restoring all does not work anymore since we lost one snapshot, restore all available ones instead.
+  ! lxc restore c1 c1/c1-snap1 --volumes=all || false
+  lxc start c1 # A failed restore leaves the instance as stopped.
```
hamistao, line 536: I noticed this behavior while doing the tests. While I can see the reasoning behind it, perhaps it would be an improvement to try to start the instance back if it was running before, but not fail if starting is not possible.

lxd/storage/backend_lxd.go
```diff
@@ -3749,6 +3783,21 @@ func (b *lxdBackend) CreateInstanceSnapshot(inst instance.Instance, src instance
...
+	// Snapshot attached volumes.
+	// TODO: Improve this by implementing concurrency.
+	for _, volume := range targetVolumes {
+		snapshotName, err := VolumeDetermineNextSnapshotName(b.state, volume.Pool, volume.Name, volume.Config)
+		if err != nil {
+			return err
+		}
+
+		// Add a description to easily identify attached volume snapshots.
+		description := "Attached volume snapshot for " + inst.Type().String() + "/" + inst.Name() + " on project " + inst.Project().Name
```
hamistao, line 3845: This expands on what is in the spec as it was an idea I had while working on the implementation. I find this to be a nice replacement for the proposed attached volume snapshot naming, as it heps the user identify which snapshots were taken together easily. Let me know if this is not desirable and I can remove it.

doc/api-extensions.md
```diff
@@ -2642,3 +2642,21 @@ This adds a request option to set snapshot's target profile on instance copy to
...
+
+1. `root`: Represents a restore of only the instance root volume. This same behavior is kept if `volumes` is unset.
+1. `all`: Represents a multi volume restore of the instance root volume and all attached volumes. Fails if one of the attached volumes snapshots is not available.
+1. `available`: Represents a multi volume restore of the instance root volume and all available attached volumes. Ignore attached volumes that are not available to restore.
+
+To keep track of which volumes are taken together, a new volatile configuration key {config:option}`volume-conf:volatile.attached_volumes` is introduced for instance snapshots. This key contains a comma separated list of UUIDs for snapshots of attached volumes that were taken together with the instance snapshot.
+
+Lastly, a new configuration key for instances `snapshots.schedule.volumes` is introduced to open the possibility of performing a multi volume snapshot on a schedule. This key defines the behavior of scheduled snapshots and follows the same possible values (`root`, `all` and `exclusive`) and describe the same respective behaviours as `volumes` in `POST /1.0/instances/{name}/snapshots`.
```
hamistao, line 2662: @minaelee `snapshots.schedule.volumes` is missing `{config:option}` because it would break the tests as `snapshots.schedule.volumes` will be added in a follow up PR. So I am planning to add `{config:option}` then if that is ok.

hamistao: Also, I had to use `storage-btrfs-volume-conf` in this file because `storage-btrfs` is the first mentioned entity, but it gives the wrong impression that this config key is exclusive to `btrfs`, which it isn't. Would you have an idea to improve this?

hamistao:
@tomponline I am hitting this confusing lint error, and since I am running late I will take a look at this tomorrow. Please feel free to review regardless of this problem, tests are looking fine locally.

tomponline:
@hamistao please can we start by moving all linter fixes and any simplifications or non-functional refactors into a separate PR first. Thanks

hamistao:
Although marked as ready, this should be rebased when https://github.com/canonical/lxd/pull/15352is mergef as this includes commits from there.

shared/api/instance_snapshot.go
```diff
@@ -18,6 +18,15 @@ type InstanceSnapshotsPost struct {
 	// Example: false
 	Stateful bool `json:"stateful" yaml:"stateful"`
 
+	// Which volumes should be snapshotted.
+	// Can be "root" or empty, for just the instance root volume,
+	// "exclusive" for all non-shared attached volumes and
+	// "all" for all attached volumes, fails if has a shared attached volume.
```
minaelee, line 24:
```suggestion
	// "all" (for all attached volumes; fails if has a shared attached volume).
```

minaelee (changes requested):

doc/api-extensions.md
minaelee: I added a suggestion to improve it by linking the Btrfs-specific config key as an example only :)

hamistao: Works for me, thanks!

hamistao:
Converting to draft to fix conflicts with main

lxd/instance/drivers/driver_common.go
```diff
@@ -545,6 +545,84 @@ func (d *common) getAttachedVolumeSnapshots(sourceSnapshot instance.Instance, di
...
+// This includes loading the storage pool, deriving which volumes should be restored, stopping the instance
+// and updating instance config.
+func (d *common) restoreCommon(inst instance.Instance, source instance.Instance, volumeDisks deviceConfig.Devices) (storagePools.Pool, bool, *operationlock.InstanceOperation, error) {
+	// Load the storage driver.
+	pool, err := storagePools.LoadByInstance(d.state, inst)
+	if err != nil {
+		return nil, false, nil, err
+	}
+
+	op, err := operationlock.Create(d.Project().Name, d.Name(), operationlock.ActionRestore, false, false)
```
hamistao, line 564: @tomponline  Probably you have no memory of this, but as reviewer of this PR and the one that included the change I am about to link to, would you be able to clarify on why [this](https://github.com/canonical/lxd/blob/26356ac250618774fc49d929ded6238d4c0119b2/lxd/instance/drivers/driver_qemu.go#L5117) was being performed here?

From what I can see, both operations are exactly the same and will finish effectively at the same time, so why is this refreshing needed?

tomponline: Please do check for the validity of this, but: I think its because `err := d.Stop(false)` will cancel any ongoing operations for an instance.

hamistao: In fact, after looking into `Stop`, I believe it [inherits the ongoing operation being used for restore](https://github.com/hamistao/lxd/blob/49834eeed91b3a1e6b1330926fe5aca059bd3243/lxd/instance/drivers/driver_lxc.go#L2672) and the [`onStop` hook finishes it](https://github.com/hamistao/lxd/blob/49834eeed91b3a1e6b1330926fe5aca059bd3243/lxd/instance/drivers/driver_lxc.go#L2816).

hamistao:
@tomponline This is now ready for a full review

tomponline:
Please rebase

doc/api-extensions.md
```diff
@@ -2668,3 +2668,17 @@ This adds support for listing network ACLs across all projects using the `all-pr
...
+1. `volumes`: Represents a multi-volume snapshot of the instance root volume and all non-shared attached volumes. Fails if any attached volumes are being shared.
+
+Also adds a `exclude_disks` field to both endpoints to allow a user to specify disk names whose volumes should not be included as part of the multi-volume snapshot.
+
+To track which volumes' snapshots are created together, this extension introduces a new volatile configuration key, {config:option}`instance-volatile:volatile.attached_volumes`, in the configuration of supported storage drivers for instance snapshots. This key contains a JSON-serialized map of attached volume UUIDs to the UUIDs of their corresponding snapshots.
+
+Lastly, a couple new configuration keys for instances to enable scheduling multi-volume snapshots. The `snapshots.schedule.disks` key defines the behavior of scheduled snapshots and follows the same possible values (`root` and `volumes`). It describes the same respective behaviors as `disks` in `POST /1.0/instances/{name}/snapshots`. The `snapshots.schedule.exclude-disks` key allows the user to explicitly exclude a disk sourced from a volume from scheduled multi-volume snapshots.
```
tomponline, line 2684: How about `snapshots.schedule.disks.exclude`?

lxd/instance/instance_utils.go
tomponline, line 1: can this commit be separated from this PR?

Also the commit message would benefit from a *why* statement, seeing as we are skipping validation.

shared/api/instance_snapshot.go
```diff
@@ -18,6 +18,20 @@ type InstanceSnapshotsPost struct {
...
 
+	// Which disks should be included in the multi-volume snapshot.
+	// Can be "root" or empty (for just the instance root volume), and
+	// "volumes" (for all disks sourced by volumes; fails an attached volume is being shared).
+	// Example: "volumes"
+	//
+	// API extension: instance_snapshot_multi_volume
+	Disks string `json:"disks,omitempty" yaml:"disks,omitempty"`
+
+	// Disks sourced by volumes that should not included in a multi-volume snapshot.
```
tomponline, line 29: "should not be..."

shared/api/instance_snapshot.go
```diff
@@ -18,6 +18,20 @@ type InstanceSnapshotsPost struct {
 	// Example: false
 	Stateful bool `json:"stateful" yaml:"stateful"`
 
+	// Which disks should be included in the multi-volume snapshot.
+	// Can be "root" or empty (for just the instance root volume), and
+	// "volumes" (for all disks sourced by volumes; fails an attached volume is being shared).
```
tomponline, line 23: "fails if..."

lxd/instance/instance_interface.go
```diff
@@ -43,6 +43,16 @@ const (
 	ConsoleTypeVGA     = "vga"
 )
 
+// Disks indicates which disks should be part of an instance operation.
+type Disks int8
```
tomponline, line 47: Is this new type really needed, and if so, could we use string type and constants as the input is already in a string.

lxd/instance/instance_utils.go
```diff
@@ -1223,3 +1223,15 @@ func SnapshotProtobufToInstanceArgs(s *state.State, inst Instance, snap *migrati
 
 	return &args, nil
 }
+
+// DisksFromString converts a string into a Disks object.
+func DisksFromString(volumes string) (Disks, error) {
```
tomponline, line 1228: `volumes` arg name here seems misleading.

Can we call this function `ValidateDisks()`?

Also, now I'm looking at the API extension, I'm wondering if it wouldnt be clearer to call the new API field `DisksMode` and then this function could be called `ValidateDisksMode()` which would be clearer.

It also avoid having a field called `Disks` and a field called `ExcludeDisks` which at first glance seems counter intuitive.

lxd/instance_put.go
```diff
@@ -188,11 +198,21 @@ func instancePut(d *Daemon, r *http.Request) response.Response {
...
 		opType = operationtype.InstanceUpdate
 	} else {
+		disks, err := instance.DisksFromString(configRaw.Disks)
+		if err != nil {
+			return response.BadRequest(err)
+		}
+
+		// "excluded_disks" do not apply on single volume snapshots.
+		if disks == instance.DisksRoot && len(configRaw.ExcludeDisks) > 0 {
+			return response.BadRequest(errors.New(`Field "excluded_disks" was provided on a single volume snapshot`))
```
tomponline, line 208: "root volume only snapshots" would be clearer in my view as it aligns with the default "root" Disks Mode

lxd/instance_put.go
```diff
@@ -229,7 +249,7 @@ func instanceSnapRestore(s *state.State, projectName string, name string, snap s
 	if err != nil {
 		switch {
 		case response.IsNotFoundError(err):
-			return fmt.Errorf("Snapshot %s does not exist", snap)
+			return api.NewStatusError(http.StatusBadRequest, "Snapshot "+snap+" does not exist")
```
tomponline, line 252: this should be a separate commit, ideally in a different PR

lxd/instance_snapshot.go
```diff
@@ -351,9 +352,26 @@ func instanceSnapshotsPost(d *Daemon, r *http.Request) response.Response {
...
 	}
 
+	disks, err := instance.DisksFromString(req.Disks)
+	if err != nil {
+		return response.BadRequest(err)
+	}
+
+	// "exclude_disks" only make sense for multi-volume snapshots.
+	if len(req.ExcludeDisks) > 0 && disks == instance.DisksRoot {
+		return response.BadRequest(errors.New(`Field "exclude_disks" was provided in a single-volume snapshot`))
```
tomponline, line 362: "root volume only snapshots"

lxd/instance_snapshot.go
```diff
@@ -351,9 +352,26 @@ func instanceSnapshotsPost(d *Daemon, r *http.Request) response.Response {
 		}
 	}
 
+	disks, err := instance.DisksFromString(req.Disks)
```
tomponline, line 355: Looking at this pattern now, would it be more useful and avoid duplication to have a function like this:

```go
extraDisks, err := instance.ParseSnapshotDisks(req) // returns []string of extra disks to snapshot and any errors (like checking for `len(req.ExcludeDisks) > 0 && disks == instance.DisksRoot `).
```

Then this can be passed to `inst.Snapshot()` as a list of disks to snapshot along with the root volume, rather than passing the mode and excluded disks?

shared/util.go
```diff
@@ -731,6 +731,16 @@ func ValueInSlice[T comparable](key T, list []T) bool {
 	return false
 }
 
+// MapToSlice organizes the values in a map into a slice.
+func MapToSlice[K comparable, V any](originalMap map[K]V) []V {
```
tomponline, line 735: It would be useful to see this used in the next commit otherwise its difficult to see when/why this is needed.

Also, can we avoid needing this function by using the new `maps` and `slices` packages?

lxd/instance/drivers/driver_common.go
```diff
@@ -324,6 +324,61 @@ func (d *common) Snapshots() ([]instance.Instance, error) {
...
+
+			_, usesVol := original[volumeKey{dev["pool"], dev["source"]}]
+
+			if !usesVol {
+				continue
+			}
+
+			// Delete the shared volume from the map and move on.
+			shared = append(shared, original[key])
+			delete(original, key)
```
tomponline, line 374: why does this modify the `original` map, seems like this function is doing too much?

lxd/instance/drivers/driver_common.go
```diff
@@ -324,6 +324,61 @@ func (d *common) Snapshots() ([]instance.Instance, error) {
...
+			return nil
+		}
+
+		expandedDevices := instancetype.ExpandInstanceDevices(listedInstance.Devices, listedInstance.Profiles)
+
+		// Iterate through each of the instance's devices, looking for disks sourced by volumes attached to inst.
+		for _, dev := range expandedDevices {
+			key := volumeKey{dev["pool"], dev["source"]}
+
+			_, usesVol := original[volumeKey{dev["pool"], dev["source"]}]
```
tomponline, line 366: Can we use `volumeIsUsedByDevice` here so at least we can find this function later when looking for functions to do with volume usage and ensure it uses a common definition of "using".

lxd/instance/drivers/driver_common.go
```diff
@@ -379,6 +379,61 @@ func (d *common) getSharedVolumes(ctx context.Context, tx *db.ClusterTx, origina
 	return shared, err
 }
 
+// attachedCustomVolumes returns a map of custom volumes attached to the provided instance.
+func (d *common) attachedCustomVolumes(ctx context.Context, tx *db.ClusterTx, inst instance.Instance, devices deviceConfig.Devices) (map[volumeKey]*api.StorageVolume, error) {
+	instanceProject := inst.Project()
+
+	// Get the instance's effective project for volumes.
+	// Create a variables so we can point to them.
```
tomponline, line 388: a variable

lxd/instance/drivers/driver_common.go
```diff
@@ -379,6 +379,61 @@ func (d *common) getSharedVolumes(ctx context.Context, tx *db.ClusterTx, origina
...
+				continue
+			}
+
+			volumeType, err := dbCluster.StoragePoolVolumeTypeFromName(dbVol.Type)
+			if err != nil {
+				return nil, err
+			}
+
+			// Skip if volume if not from the correct project
+			if effectiveProject != project.StorageVolumeProjectFromRecord(&instanceProject, volumeType) {
```
tomponline, line 418: didnt you already filter by project on line 392?

tomponline (commented):
@hamistao I wanted to check something, originally we were going to have a mode for selecting only exclusively attached disks right? But now this is replaced by a disk exclusion mechanism. Good. But I wanted to check do we silently skip over shared volumes now or do we/should we fail to snapshot and then require shared ones to be skipped?

lxd/instance/drivers/driver_common.go
```diff
@@ -801,10 +802,62 @@ func (d *common) runHooks(hooks []func() error) error {
...
+	snapshotConfig := inst.LocalConfig()
+	var volumeMap map[volumeKey]*api.StorageVolume
+
+	// If any disks were provided, include attached volumes.
+	if len(disks) > 0 {
+		// Get a map of attached volumes and check for shared ones.
+		// if any disks were provided, fail if a volume used as source is shared.
+		var shared []*api.StorageVolume
+		err := d.state.DB.Cluster.Transaction(d.state.ShutdownCtx, func(ctx context.Context, tx *db.ClusterTx) error {
+			volumeMap, err = d.attachedCustomVolumes(ctx, tx, inst, disks)
```
tomponline, line 1021: would combining these functions and returning a slice of shared volumes and a slice of volumes allow us to reduce the number of queries/loops done?

lxd/storage/backend_lxd.go
```diff
@@ -6582,7 +6582,7 @@ func (b *lxdBackend) ImportCustomVolume(projectName string, poolVol *backupConfi
 }
 
 // CreateCustomVolumeSnapshot creates a snapshot of a custom volume.
-func (b *lxdBackend) CreateCustomVolumeSnapshot(projectName, volName string, newSnapshotName string, newDescription string, newExpiryDate time.Time, op *operations.Operation) error {
+func (b *lxdBackend) CreateCustomVolumeSnapshot(projectName, volName string, newSnapshotName string, newDescription string, newExpiryDate time.Time, newSnapshotUUID string, op *operations.Operation) error {
```
tomponline, line 6669: How about we keep the volume UUID generation the responsibility of the storage subsystem and update this function signature to return the UUID used for the snapshot?

lxd/storage/backend_lxd.go
```diff
@@ -3683,6 +3699,43 @@ func (b *lxdBackend) CreateInstanceSnapshot(inst instance.Instance, src instance
...
+		l.Debug("Creating attached volume snapshot", logger.Ctx{"volumeProject": volume.Project, "volumePool": volume.Pool, "volumeName": volume.Name})
+
+		// Use shutdown context here as we don't have access to the request context.
+		snapshotName, err := VolumeDetermineNextSnapshotName(b.state.ShutdownCtx, b.state, volume.Pool, volume.Name, volume.Config)
+		if err != nil {
+			return err
+		}
+
+		// Add a description to easily identify attached volume snapshots.
+		description := "Attached volume snapshot related to " + inst.Type().String() + "/" + inst.Name() + " on project " + inst.Project().Name
```
tomponline, line 3719: Maybe "Created by snapshot of ..."?

lxd/storage/backend_lxd.go
```diff
@@ -3683,6 +3699,43 @@ func (b *lxdBackend) CreateInstanceSnapshot(inst instance.Instance, src instance
...
+			volumePool, err = LoadByName(b.state, volume.Pool)
+			if err != nil {
+				return err
+			}
+
+			poolMap[volume.Pool] = volumePool
+		}
+
+		// Attached volume snapshots inherit expiration time from instance snapshot.
+		err = volumePool.CreateCustomVolumeSnapshot(volume.Project, volume.Name, snapshotName, description, inst.ExpiryDate(), attachedVolumeSnapshotUUIDs[volume.Config["volatile.uuid"]], op)
```
tomponline, line 3733: we should add a delete of the snapshots to the reverter so they are removed if anything goes wrong

lxc/snapshot.go
```diff
@@ -44,6 +46,8 @@ running state, including process memory state, TCP connections, ...`))
 	cmd.Flags().BoolVar(&c.flagStateful, "stateful", false, i18n.G("Whether or not to snapshot the instance's running state"))
 	cmd.Flags().BoolVar(&c.flagNoExpiry, "no-expiry", false, i18n.G("Ignore any configured auto-expiry for the instance"))
 	cmd.Flags().BoolVar(&c.flagReuse, "reuse", false, i18n.G("If the snapshot name already exists, delete and create a new one"))
+	cmd.Flags().StringVar(&c.flagDisks, "disks", "", i18n.G(`Which disks should be included in the snapshot; can be "root" or "volumes"`))
+	cmd.Flags().StringVar(&c.flagExcludeDisks, "exclude-disks", "", i18n.G(`Which volume-sourced disks should be ignored when performing a multi-volume snapshot; contains a comma-separated list of device names`))
```
tomponline, line 50: I'd exclude `volume-sourced` part from this as its covered by `--disks=volumes`.

lxd/instance/instance_utils.go
hamistao: Yes, it can. The why here is so we can add `volatile.attached_volumes` to the snapshot config without breaking while not allowing it on the instance config. I can add that to the commit message.

lxd/instance/instance_utils.go
```diff
@@ -1235,3 +1235,33 @@ func DisksFromString(volumes string) (Disks, error) {
...
+// according to the provided "volumeDisks" argument.
+// The "excludedDevices" field indicates the names of devices that should be excluded from the return value.
+func VolumeDisksFromInstance(inst Instance, volumeDisks Disks, excludedDisks []string) (deviceConfig.Devices, error) {
+	// If we only care about the root disk, return an empty map.
+	if volumeDisks == DisksRoot {
+		return nil, nil
+	}
+
+	volumeDevices := inst.ExpandedDevices().Clone()
+	// Remove devices that are disks sourced by volumes
```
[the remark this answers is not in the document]
hamistao: yes, there is

lxd/instance_put.go
hamistao: I will include this one in a separate PR along with the other ones that can be taken apart.

hamistao:
> I wanted to check something, originally we were going to have a mode for selecting only exclusively attached disks right? But now this is replaced by a disk exclusion mechanism. Good. But I wanted to check do we silently skip over shared volumes now or do we/should we fail to snapshot and then require shared ones to be skipped?

@tompoline We now fail to snapshot and then require shared volumes to be skipped. I now included the disk names for the shared volumes on the error message to enhance the user experience and to match restore.

lxc/snapshot.go
hamistao: makes sense to me

shared/util.go
hamistao: Oh, we can use `slices.Collect(maps.Values(myMap))`

lxd/instance/instance_utils.go
hamistao: makes sense to me, but the same reasoning would apply for the names of the fields themselves: `disks` and `exclude_disks` (and the CLI flags). Perhaps we could then use `disks_mode` or something like `include_disks`?

shared/util.go
tomponline: bingo!

lxd/instance/instance_interface.go
hamistao: Not really, we can just use strings if you prefer.

doc/api-extensions.md
hamistao: I had some different ideas of how to implement scheduling here, I am thinking of adding them to the spec for future reference and leave this part out of this PR.

lxd/instance/drivers/driver_common.go
hamistao: Not really, for queries we would still have to query for the project custom volumes once and query for project instances to check to check for volume sharing. And for loops, we would stiil need to find the attached volumes and then iterate through instances to check for sharing, so it would still be a `O(projectCustomVolumes+projectInstances)`.

lxd/instance_snapshot.go
hamistao: I just think we could return a `Devices` (map of Device which is itself a map[string]string) instead of a []string, since converting it to []string would require an extra step and we would need to get the full Devices object anyway.

lxd/storage/backend_lxd.go
hamistao: I only did not go for this right away because that would require us to either:
1- Add one more query to update the instance snapshot config after it was created;
2- Create the DB record after creating the volumes, which I assume is better but breaks the logic of "doing the easy thing first"

lxd/instance_snapshot.go
hamistao: Also using the disk mode and excluded disks as arguments allows us to use it for both snapshot and restore.

lxd/storage/backend_lxd.go
hamistao: I think "Created by snapshot of ..." hints at the instance itself, but perhaps including the snapshot name would be nice, what do you think of `Created alongside {full snapshot name}`?

hamistao:
@tomponline Please comment on the last few unresolved conversations when you have the time. Also, There are only two commits that I can separate from this PR, is it still worth it to do it?

hamistao:
@tomponline SInce time is short today, I am thinking of leaving this as is. I believe it is looking good enough so it shouldn't give you much trouble getting it merged.

kadinsayani:
I'm going to close this and open several smaller pull requests that cherry-pick commits from this one and address the outstanding comments. I'll also include documentation and add tests to `lxd-ci`.

tomponline:
> I'm going to close this and open several smaller pull requests that cherry-pick commits from this one and address the outstanding comments. I'll also include documentation and add tests to `lxd-ci`.

Good idea!

PR #15350, opened 2025-04-08
Device: Fix disk attachment of ceph instance volumes in default project

tomponline:
The startVM function was always using the project.StorageVolume() function to derive the underlying ceph rbd image name. This is incorrect when the volume being attached is from an instance in the default project because the `default_` prefix is only added for custom volumes, not instance volumes.

Additionally this function was also using the instance or volume's project name directly rather than using the existing `storageProjectName := project.StorageVolumeProjectFromRecord()` variable. This is a problem because for custom volumes it wasn't taking into account the value of `features.storage.volumes` set on the project (indicating the effective project for a custom volume).

So instead I switched to using the same invocation pattern we already have from disk.mountPoolVolume() function.

Also starts the process of pushing the generation of DevSource type values down into the storage subsystem.

### DIFF
--- a/lxd/storage/pool_interface.go
+++ b/lxd/storage/pool_interface.go
@@ -8,6 +8,7 @@ import (
 	"github.com/canonical/lxd/lxd/backup"
 	backupConfig "github.com/canonical/lxd/lxd/backup/config"
 	"github.com/canonical/lxd/lxd/cluster/request"
+	deviceConfig "github.com/canonical/lxd/lxd/device/config"
 	"github.com/canonical/lxd/lxd/instance"
 	"github.com/canonical/lxd/lxd/instancewriter"
 	"github.com/canonical/lxd/lxd/migration"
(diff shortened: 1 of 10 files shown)


tomponline:
> @tomponline It seems I was outsped in fixing this 😅
> 
> Do you still need me to look into any CI test failures today?

No its fine, i know what the problem is (mentioned in MM), but not yet got a handle on the fix. I noticed this additional bug along the way.

tomponline:
> > No its fine, i know what the problem is (mentioned in MM), but not yet got a handle on the fix. I noticed this additional bug along the way.
> 
> Cool, if you don't have the time today to get the fix ready just say the word as today I have more time to focus on this.

@escabo would prefer you to focus on roadmap items at this time please.

lxd/storage/pool_interface.go
```diff
@@ -26,7 +27,7 @@ type VolumeUsage struct {
 
 // MountInfo represents info about the result of a mount operation.
 type MountInfo struct {
-	DiskPath  string                               // The location of the block disk (if supported).
+	DevSource deviceConfig.DevSource               // The location of the block disk (if supported).
```
hamistao, line 30: If we are always expecting this to be a `DevSourcePath`, why not have this be a `DevSourcePath` instead of `DevSource`?

tomponline: we arent in the long run, the idea is to push down the ceph rbd stuff too.

hamistao (approved):
LGTM!

PR #15352, opened 2025-04-08
storage: Pre-multi volume snapshots

hamistao:
Includes some simplifications and other changes that will be useful in the multi volume snapshot implementation, see https://github.com/canonical/lxd/pull/15349.


lxd/storage/utils.go
```diff
@@ -1357,3 +1361,106 @@ func GetPoolDefaultBlockSize(s *state.State, poolName string) (string, error) {
 
 	return pool.Driver().Info().DefaultBlockSize, nil
 }
+
+// VolumeDetermineNextSnapshotName determines a name for next snapshot of a volume
+// following the volume's snapshots.pattern or the provided default pattern.
+func VolumeDetermineNextSnapshotName(s *state.State, pool string, volumeName string, volumeConfig map[string]string) (string, error) {
```
tomponline, line 1369: can this function take a context argument we can pass to it using r.Context()?

hamistao: Yes that would be possible. I assume your idea is to use the request context instead of `ShutdownCtx`.

tomponline: yeah thats right where possible

lxd/storage/utils.go
```diff
@@ -1364,3 +1368,106 @@ func GetPoolDefaultBlockSize(s *state.State, poolName string) (string, error) {
...
+	})
+	if err != nil {
+		return "", err
+	}
+
+	for _, pool := range pools {
+		var poolID int64
+
+		err = s.DB.Cluster.Transaction(ctx, func(ctx context.Context, tx *db.ClusterTx) error {
+			poolID, err = tx.GetStoragePoolID(ctx, pool)
```
tomponline, line 1432: can we merge this work into the transaction above to avoid starting another transaction?

lxd/device/disk.go
```diff
@@ -205,11 +205,8 @@ func (d *disk) checkBlockVolSharing(instanceType instancetype.Type, projectName
 
 		return fmt.Errorf("Cannot add block volume to more than one instance if security.shared is false or unset")
 	})
-	if err != nil {
-		return err
-	}
 
-	return nil
+	return err
```
tomponline, line 209: simpler still would be `return storagePools.VolumeUsedByInstanceDevices(...`

lxd/storage/backend_lxd.go
```diff
@@ -3621,7 +3621,10 @@ func (b *lxdBackend) CreateInstanceSnapshot(inst instance.Instance, src instance
 		defer func() { _ = src.Unfreeze() }()
 
 		// Attempt to sync the filesystem.
-		_ = filesystem.SyncFS(src.RootfsPath())
+		err = filesystem.SyncFS(src.RootfsPath())
+		if err != nil {
+			l.Warn("Failed to flush writes to instance volume", logger.Ctx{"err": err})
```
tomponline, line 3626: Ah good, you did sync the comment here, thanks (see my earlier comment).

However there are more things to consider along this theme.

tomponline (changes requested):
Lets drop the syncfs changes and consider them in a separate PR. Ta

lxd/storage/backend_lxd.go
```diff
@@ -6537,6 +6537,15 @@ func (b *lxdBackend) CreateCustomVolumeSnapshot(projectName, volName string, new
 		return fmt.Errorf("Volume of content type %q does not support snapshots", contentType)
 	}
 
+	// If the volume is a filesystem, attempt to sync the filesystem before taking the snapshot.
+	if parentVol.ContentType == cluster.StoragePoolVolumeContentTypeNameFS {
```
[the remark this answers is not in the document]
hamistao: 3 is definetely a mistake here and will be fixed (not in this PR of course). As for the rest, I will elaborate more on them on the upcoming PR.

lxd/storage/utils.go
hamistao: Yeah, we can!

hamistao:
@tomponline All done over here!

tomponline (approved):

tomponline:
ta

PR #15355, opened 2025-04-08
Device: Include src path in MountInfo returned from mountPoolVolume() in `disk` device

tomponline:
Part of the journey to push DeviceSource concept down into the storage subsystem.

Follows https://github.com/canonical/lxd/pull/15350


hamistao (approved):
LGTM

PR #15357, opened 2025-04-09
Ubuntu Pro: Revert "lxd/ubuntu-pro: Change Ubuntu Pro directory."

markylaing:
As discussed internally, the Ubuntu Pro client will place `interfaces/lxd-config.json` under `/var/lib/ubuntu-advantage`. This is because the `/var/lib/ubuntu-advantage` directory is already present on the Ubuntu system, whereas `/var/lib/ubuntu-pro` was only created when `pro config set lxd_guest_attach <value>` was called.

### DIFF
--- a/lxd/ubuntupro/client.go
+++ b/lxd/ubuntupro/client.go
@@ -48,8 +48,8 @@ func validateGuestAttachSetting(guestAttachSetting string) error {
 	return nil
 }
 
-// ubuntuProDirectory is the base directory for Ubuntu Pro related configuration.
-const ubuntuProDirectory = "/var/lib/ubuntu-pro"
+// ubuntuAdvantageDirectory is the base directory for Ubuntu Pro related configuration.
+const ubuntuAdvantageDirectory = "/var/lib/ubuntu-advantage"
 
 // Client is our wrapper for Ubuntu Pro configuration and the Ubuntu Pro CLI.
 type Client struct {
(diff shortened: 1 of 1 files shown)


tomponline (commented):
Once the failed tests are passing this is good.

tomponline:
@markylaing can we add some tests for this functionality by simulating the files that the pro client should create (i realise the full pro support cant be tested, but just partial /dev/lxd responses?)

markylaing:
Ah of course I forgot to sign it off :facepalm:

markylaing:
git commit -s is natural to me now

> @markylaing can we add some tests for this functionality by simulating the files that the pro client should create (i realise the full pro support cant be tested, but just partial /dev/lxd responses?)

There are a bunch of unit tests in https://github.com/canonical/lxd/blob/main/lxd/ubuntupro/client_test.go. The client uses a shim for the `pro` CLI to cover as many cases as possible. Should I add an environment variable to set the path during testing perhaps?

tomponline:
> git commit -s is natural to me now
> 
> > @markylaing can we add some tests for this functionality by simulating the files that the pro client should create (i realise the full pro support cant be tested, but just partial /dev/lxd responses?)
> 
> There are a bunch of unit tests in https://github.com/canonical/lxd/blob/main/lxd/ubuntupro/client_test.go. The client uses a shim for the `pro` CLI to cover as many cases as possible. Should I add an environment variable to set the path during testing perhaps?

Thanks, seems fine as it is.

tomponline (approved):
ta

PR #15359, opened 2025-04-09
storage: Instance and custom volume snapshot consistency

hamistao:
This implements `SyncFS` when snapshotting custom volumes, similarly to what we already do with instance volumes.

This also includes a warning it fit fails for either instance or custom volumes. In a different manner as `fsfreeze`, we do not expect `sync` to fail even if the filesystem is busy.

I was at odds on whether I should gate SyncFS on `RunningCopyFreeze` as even after some digging on the PRs that populated this field on each of the drivers, it was unclear on what specifically determined the value of  `RunningCopyFreeze` for each driver.

After some more research on each driver's inner workings, I got to the conclusion that `RunningCopyFreeze=false` indeed means the driver syncs the volume on snapshot (as snapshots generally follow a similar mechanism to the one used for copy).

As reference, I found useful sources for [btrfs](https://btrfs.readthedocs.io/en/latest/btrfs-subvolume.html#performance) and [cephfs](https://docs.ceph.com/en/quincy/dev/cephfs-snapshots/), both of which have `RunningCopyFreeze=false`. Unfortunately I found nothing for `zfs`, that also falls within this same category

### DIFF
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -6542,6 +6542,18 @@ func (b *lxdBackend) CreateCustomVolumeSnapshot(projectName, volName string, new
 		return fmt.Errorf("Volume of content type %q does not support snapshots", contentType)
 	}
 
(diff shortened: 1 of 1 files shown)


tomponline (approved):
Thanks

PR #15360, opened 2025-04-09
Refactor cluster client functions

kadinsayani:
Updates client functions in `lxd_cluster.go` to construct URLs using `api.NewURL()` and replaces HTTP method literals with constants from `http` package.

### DIFF
--- a/client/lxd_cluster.go
+++ b/client/lxd_cluster.go
@@ -1,6 +1,8 @@
 package lxd
 
 import (
+	"net/http"
+
 	"github.com/canonical/lxd/shared/api"
 )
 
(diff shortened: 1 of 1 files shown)


client/lxd_cluster.go
```diff
@@ -79,14 +83,14 @@ func (r *ProtocolLXD) GetClusterMemberNames() ([]string, error) {
...
-	_, err = r.queryStruct("GET", baseURL, nil, "", &urls)
+	u := api.NewURL().Path("cluster", "members")
+	_, err = r.queryStruct(http.MethodGet, u.String(), nil, "", &urls)
 	if err != nil {
 		return nil, err
 	}
 
 	// Parse it.
-	return urlsToResourceNames(baseURL, urls...)
+	return urlsToResourceNames("/cluster/members", urls...)
```
simondeziel, line 93: Would that also work?
```suggestion
	return urlsToResourceNames(u.String(), urls...)
```

simondeziel (approved):
LGTM with a tiny question.

client/lxd_cluster.go
kadinsayani: Indeed!

simondeziel (approved):
LGTM, thanks

tomponline (approved):

PR #15362, opened 2025-04-10
doc: improve network ACLs docs

minaelee:
Re-submit of https://github.com/canonical/lxd/pull/15274

Improvements to network ACLs docs, including:
- Document the ability to provide a YAML file to create an ACL (previously only in man page)
- Clarify that the only arguments accepted for CLI-based ACL creation (aside from a YAML file) are the ACL name and the custom user keys
- Add missing sections for listing, showing, renaming, and deleting ACLs
- Add API instruction counterparts to all CLI instructions
- Add links to relevant parts of API reference
- Add command syntax examples
- Improve descriptions of ACL and ACL rule properties
- Detailed information about how to edit the ACL via the API, where and how to use PATCH and POST
- Clarify how logging works
- General formatting improvements and copyedits
- Also updated the PUT vs PATCH section in rest-api.md & added a section about required device keys in instance-configuration.md

### DIFF
--- a/doc/howto/instances_configure.md
+++ b/doc/howto/instances_configure.md
@@ -180,10 +180,14 @@ See [`lxc config device --help`](lxc_config_device.md) for a full list of availa
 ````
 
 ````{group-tab} API
-To add and configure an instance device for your instance, use the same mechanism of patching the instance configuration.
+To add or configure an instance device for your instance, use the same mechanism of patching the instance configuration.
 The device configuration is located under the `devices` field of the configuration.
 
-Specify the instance name, a device name, the device type and maybe device options (depending on the {ref}`device type <devices>`):
+```{caution}
+Patching a device's configuration unsets any omitted options for that device, along with the instance's `description` property. See {ref}`instances-configure-devices-api-patch-effects` for details.
+```
+
+Specify the instance name, a device name, and any {ref}`instances-configure-devices-api-required` (depending on the {ref}`device type <devices>`):
 
     lxc query --request PATCH /1.0/instances/<instance_name> --data '{
       "devices": {
@@ -208,6 +212,69 @@ For example, to add the storage at `/share/c1` on the host system to your instan
     }'
 
 See [`PATCH /1.0/instances/{name}`](swagger:/instances/instance_patch) for more information.
+
+(instances-configure-devices-api-required)=
+### Required device options
+
+When using a PATCH request to update an instance's `devices` property, you must include any required options for each device in the request body. The device's `type` option is always required. To find any other required keys for a specific device type, view the {ref}`devices` reference guides. For example, for an OVN NIC device, the {config:option}`device-nic-ovn-device-conf:network` key is required.
+
+(instances-configure-devices-api-patch-effects)=
+### Effects of patching device options
+
+For any device in your PATCH request, the request acts similar to a conventional PUT: it replaces all options for that device. This means that if you omit a non-required option, it is unset. Thus, include not only the options you want to add or update in your patch, but also any other existing options whose values you want to keep.
+
+This behavior only affects the specific device or devices that you are patching; if there are other devices, you don't need to include them. It also does not affect any other instance properties, with one exception: if the instance includes a `description` property, that property must be passed along with `devices`; otherwise, it is unset.
+
+For example, consider an instance that contains this `devices` property:
+
+```bash
+"devices": {
+  "my-bridge-nic": {
+    "name": "my-bridge-nic-name",
+    "network": "my-bridge-network",
+    "type": "nic"
+  },
+  "my-ovn-nic": {
+    "name": "my-ovn-nic-name",
+    "network": "my-ovn-network",
+    "type": "nic"
+  }
+}
+```
+
+Let's say the following PATCH request is sent for this instance:
+
+```bash
+lxc query --request PATCH /1.0/instances/my-instance --data '{
+  "devices": {
+    "my-bridge-nic": {
+      "type": "nic",
+      "network": "test-bridge",
+      "ipv4.address": "192.0.2.10"
+    }
+  }
+}'
+```
+
+This PATCH request updates only the `my-bridge-nic` device, without affecting the `my-ovn-nic` device. The device options defined in the request body replace the existing options. After the request, this is the `devices` property's configuration:
+
+```bash
+"devices": {
+  "my-bridge-nic": {
+    "network": "my-bridge-network",
+    "type": "nic",
+    "ipv4.address": "192.0.2.10"
+  },
+  "my-ovn-nic": {
+    "name": "my-ovn-nic-name",
+    "network": "my-ovn-network",
+    "type": "nic"
+  }
+}
+```
+
+Notice that in the updated `my-bridge-nic` device, the `name` option is unset and no longer appears, due to not being sent in the PATCH request.
+
 ````
 
 ````{group-tab} UI
--- a/doc/howto/network_acls.md
+++ b/doc/howto/network_acls.md
@@ -354,21 +354,18 @@ There is no specific endpoint for removing a rule. Instead, you must {ref}`edit 
 
 You cannot edit a rule directly. Instead, you must {ref}`edit the full ACL <network-acls-edit>`, which contains the `egress` and `ingress` lists.
 
-### Rule ordering and priorities
+### Rule ordering and application of actions
 
-Rules are provided as lists.
-However, the order of the rules in the list is not important and does not affect
-filtering.
+ACL rules are defined as lists, but their order within the list does not affect how they are applied.
 
-LXD automatically orders the rules based on the `action` property as follows:
+LXD automatically prioritizes rules based on the action property, in the following order:
 
 - `drop`
 - `reject`
 - `allow`
-- Automatic default action for any unmatched traffic (defaults to `reject`, see {ref}`network-acls-defaults`).
+- The default action for unmatched traffic (defaults to `reject`, see {ref}`network-acls-defaults`)
 
-This means that when you apply multiple ACLs to a NIC, there is no need to specify a combined rule ordering.
-If one of the rules in the ACLs matches, the action for that rule is taken and no other rules are considered.
+When you assign multiple ACLs to a NIC, you do not need to coordinate rule order across them. As soon as a rule matches, its action is applied and no further rules are evaluated.
 
 ### Rule properties
 
--- a/doc/howto/network_acls.md
+++ b/doc/howto/network_acls.md
@@ -385,9 +385,14 @@ ACL rules have the following properties:
 This feature is supported only for the {ref}`OVN NIC type <nic-ovn>` and the {ref}`network-ovn`.
 ```
 
-The `source` field (for ingress rules) and the `destination` field (for egress rules) support using selectors instead of CIDR or IP ranges.
+In ACL rules, the `source` and `destination` properties support using selectors instead of CIDR blocks or IP ranges. You can only use selectors in the `source` of `ingress` rules, and in the `destination` of `egress` rules.
 
-With this method, you can use ACL groups or network selectors to define rules for groups of instances without needing to maintain IP lists or create additional subnets.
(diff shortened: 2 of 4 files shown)


doc/howto/instances_configure.md
```diff
@@ -208,6 +208,14 @@ For example, to add the storage at `/share/c1` on the host system to your instan
     }'
 
 See [`PATCH /1.0/instances/{name}`](swagger:/instances/instance_patch) for more information.
+
+(instances-configure-devices-api-required)=
+### Required keys
```
tomponline, line 213: I think we should use the word "Properties" here rather than "Keys" so we align with this part of the docs:

https://documentation.ubuntu.com/lxd/en/latest/reference/instance_properties/

That way we dont confuse Properties with config keys (which are within a specific property).

doc/howto/network_acls.md
```diff
@@ -535,6 +535,15 @@ For instance NICs, use the following command:
 lxc config device set <instance_name> <device_name> security.acls="<ACL_name>"
 ```
 
+(network-acls-assign-additional)=
+### Additional properties
```
tomponline, line 871: Similar to my earlier comment, "Properties" refers to the fields on an API struct (like Name, Description etc) and not config options specified in maps. So here I think "Additional options" is more consistent.

tomponline (changes requested):
This looks great, thanks @minaelee 
It was a lot easier to go through with smaller commits.

There is only one thing I think we need to address on this PR, and that is to disambiguate the difference between Properties and Options.


For example if we take this API request:

```
lxc query --request PUT /1.0/network-acls/{ACL-name} --data '{
  "config": {
    "user.<custom key name>": "<custom key value>"
  },
  "description": "<description of the ACL>",
  "egress": [<egress rule>, <another egress rule...>,...],
  "ingress": [<ingress rule>, <another ingress rule...>,...]
}'
```

Here the `user.*` keys are *options* because they are stored inside the `config` *property*. Similarly `description`, `egress` and `ingress` are properties.

So when using a PATCH request only *options* can be merged, e.g. 

```
lxc query --request PATCH /1.0/network-acls/{ACL-name} --data '{
  "config": {
    "user.foo": "bar"
  },
}'
```

Here the `user.foo` config option will be merged into the existing config options.
However the properties will be reset to default, which means `description`, `ingress` and `egress` properties will be wiped.

So we should probably draw attention to this in the examples to avoid surprised users who might think only the `config` options would be updated.

doc/howto/instances_configure.md
minaelee: I'm afraid I don't follow. This section is intended to discuss required keys when updating device options, such as the required `network` option key for OVN NIC devices:
image

My understanding is that the top-level keys, such as `devices`, are what we're calling _properties_, but that the key-value pairs within a property that's a JSON object type, such as `devices` within `network`, are what we're calling _keys_ or _option keys_, such as in the screenshot above.

This section refers to needing to send option keys that are required when sending an API request to patch any key. For example, a GET request to  /1.0/instances/ubuntu-container on my local machine shows the `devices` property like this:


If I try to PATCH any option key within `devices`, I get failures unless I include both the `type` and `network` option keys:
image

I don't follow why this section should refer to required properties instead?

doc/howto/network_acls.md
minaelee: Ty, fixed

doc/howto/instances_configure.md
tomponline: I see what you mean. Although what I was getting at was that the PATCH examples on this page will result in the Instance Properties being reset too.

So I had thought this section was a warning that although the examples don't show it, you do need to include the Properties in all PATCH requests.

tomponline: @minaelee how would you like to proceed here?

minaelee: > I see what you mean. Although what I was getting at was that the PATCH examples on this page will result in the Instance Properties being reset too.
> 
> So I had thought this section was a warning that although the examples don't show it, you do need to include the Properties in all PATCH requests.

The PATCH examples for updating an instance device do not reset the instance properties, unless I misunderstand what you mean there.

For example, here is my `ubuntu-container` instance configuration:

```
minae@iris:~$ lxc query --request GET /1.0/instances/ubuntu-container
{
	"architecture": "x86_64",
	"config": {
		"image.architecture": "amd64",
		"image.description": "ubuntu 24.04 LTS amd64 (release) (20250228)",
		"image.label": "release",
		"image.os": "ubuntu",
           ... (I cut the rest of this part out just to make this example shorter! it's a lot of lines) ...
	},
	"created_at": "2025-03-04T00:54:32.583471218Z",
	"description": "",
	"devices": {
		"bridge-nic": {
			"host_name": "test-host",
			"network": "test-bridge",
			"type": "nic"
		},
		"ovn-nic": {
			"network": "ovntest",
			"security.acls.default.ingress.action": "reject",
			"type": "nic"
		}
	},
	"ephemeral": false,
	"expanded_config": {
           ... (I cut this part out just to make this example shorter! it's a lot of lines) ...
	},
	"expanded_devices": {
		"bridge-nic": {
			"network": "test-bridge",
			"type": "nic"
		},
		"eth0": {
			"name": "eth0",
			"network": "lxdbr0",
			"type": "nic"
		},
		"ovn-nic": {
			"network": "ovntest",
			"security.acls.default.ingress.action": "reject",
			"type": "nic"
		},
		"root": {
			"path": "/",
			"pool": "default",
			"type": "disk"
		}
	},
	"last_used_at": "2025-04-08T19:16:42.694881104Z",
	"location": "none",
	"name": "ubuntu-container",
	"profiles": [
		"default"
	],
	"project": "default",
	"stateful": false,
	"status": "Running",
	"status_code": 103,
	"type": "container"
}
```

The instance properties in this case are `architecture`, `config`, `created_at`, `description`, `devices`, and every other top-level key. 

If I send a PATCH request to update the content of `devices` (or any other property), the behavior that I see is what would conventionally be expected from PATCH:  the _other_ properties are left untouched.

However, for the `devices` property, the behavior is more tricky. It behaves like a PUT, but only for devices that are included in the PATCH request. It leaves other devices alone, but any included device is replaced instead of updated. This is the behavior I was attempting to address in the section I added to `instances_configure.md`. 

Example: 

In the configuration for `ubuntu_container` shown above, this is the `devices` property:

```
	"devices": {
		"bridge-nic": {
			"host_name": "test-host",
			"network": "test-bridge",
			"type": "nic"
		},
		"ovn-nic": {
			"network": "ovntest",
			"security.acls.default.ingress.action": "reject",
			"type": "nic"
		}
	},
```

Note that there are two devices. I sent a PATCH request to update the configuration for the `bridge-nic` device _only_, without sending any other instance properties, _and_ without sending the `ovn-nic` device configuration either. Also note that I included a new option for the `bridge-nic` device called `name`, and I deliberately did _not_ send the `host_name` option which is in the existing configuration, to show what happens. I had to send the `network` and `type` option keys because otherwise the PATCH would fail; required options must be sent with every PATCH to a device.

```
minae@iris:~$ lxc query --request PATCH /1.0/instances/ubuntu-container --data '{ 
  "devices": {
    "bridge-nic": {
      "type": "nic",
      "network": "test-bridge",
      "name": "test-name"
    }
  }
}'
```

Then I checked the configuration again. Note that the other instance properties, which were not sent in the request body, remained unchanged, which matches the conventional behavior for PATCH:

```
minae@iris:~$ lxc query --request GET /1.0/instances/ubuntu-container
{
	"architecture": "x86_64",
	"config": {
		"image.architecture": "amd64",
		"image.description": "ubuntu 24.04 LTS amd64 (release) (20250228)",
		"image.label": "release",
		"image.os": "ubuntu",
                ... (I cut the rest of this part out just to make this example shorter! it's a lot of lines) ...
	},
	"created_at": "2025-03-04T00:54:32.583471218Z",
	"description": "",
	"devices": {
		"bridge-nic": {
			"name": "test-name",
			"network": "test-bridge",
			"type": "nic"
		},
		"ovn-nic": {
			"network": "ovntest",
			"security.acls.default.ingress.action": "reject",
			"type": "nic"
		}
	},
	"ephemeral": false,
	"expanded_config": {
           ... (I cut this part out just to make this example shorter! it's a lot of lines) ...
	},
	"expanded_devices": {
		"bridge-nic": {
			"name": "test-name",
			"network": "test-bridge",
			"type": "nic"
		},
		"eth0": {
			"name": "eth0",
			"network": "lxdbr0",
			"type": "nic"
		},
		"ovn-nic": {
			"network": "ovntest",
			"security.acls.default.ingress.action": "reject",
			"type": "nic"
		},
		"root": {
			"path": "/",
			"pool": "default",
			"type": "disk"
		}
	},
	"last_used_at": "2025-04-08T19:16:42.694881104Z",
	"location": "none",
	"name": "ubuntu-container",
	"profiles": [
		"default"
	],
	"project": "default",
	"stateful": false,
	"status": "Running",
	"status_code": 103,
	"type": "container"
}
```

Here's a closer look at the `devices` property:
```
	"devices": {
		"bridge-nic": {
			"name": "test-name",
			"network": "test-bridge",
			"type": "nic"
		},
		"ovn-nic": {
			"network": "ovntest",
			"security.acls.default.ingress.action": "reject",
			"type": "nic"
		}
	},
```

Compare it to the one before I sent the previous PATCH request:
```
	"devices": {
		"bridge-nic": {
			"host_name": "test-host",
			"network": "test-bridge",
			"type": "nic"
		},
		"ovn-nic": {
			"network": "ovntest",
			"security.acls.default.ingress.action": "reject",
			"type": "nic"
		}
	},
```
You can see like the instance properties, the `ovn-nic` device, which was not included in the request body, was not changed/unset. The `bridge-nic` device WAS changed however, in the way that a conventional PUT request would change it: a replacement of that entire device's configuration. The result is that "name" was added, but "host_name" was dropped, since it was not included in the request.

This is the behavior that the section I added to the API describes. I've edited it to include a similar example as the above, and I hope that's clearer. I also changed the word "key" to "option" as that's more consistent. Though "option key" is also used in the existing page, just "option" by itself occurs more frequently. If this is still not making sense, perhaps we can schedule some time early next week to discuss it further.

minaelee:
> This looks great, thanks @minaelee It was a lot easier to go through with smaller commits.
> 
> There is only one thing I think we need to address on this PR, and that is to disambiguate the difference between Properties and Options.
> 
> For example if we take this API request:
> 
> ```
> lxc query --request PUT /1.0/network-acls/{ACL-name} --data '{
>   "config": {
>     "user.<custom key name>": "<custom key value>"
>   },
>   "description": "<description of the ACL>",
>   "egress": [<egress rule>, <another egress rule...>,...],
>   "ingress": [<ingress rule>, <another ingress rule...>,...]
> }'
> ```
> 
> Here the `user.*` keys are _options_ because they are stored inside the `config` _property_. Similarly `description`, `egress` and `ingress` are properties.
> 
> So when using a PATCH request only _options_ can be merged, e.g.
> 
> ```
> lxc query --request PATCH /1.0/network-acls/{ACL-name} --data '{
>   "config": {
>     "user.foo": "bar"
>   },
> }'
> ```
> 
> Here the `user.foo` config option will be merged into the existing config options. However the properties will be reset to default, which means `description`, `ingress` and `egress` properties will be wiped.
> 
> So we should probably draw attention to this in the examples to avoid surprised users who might think only the `config` options would be updated.

I believe the above has all been addressed, with cautions [here](https://github.com/canonical/lxd/pull/15362/commits/0cfb75fda6d9694797d73f3d857848b218a0448e) and [here](https://github.com/canonical/lxd/pull/15362/commits/82d123d05da405445b2a815699f4801977d16bf0), and examples [here](https://github.com/canonical/lxd/pull/15362/commits/4e9a5341f953ee96db307ae2bc4d434d98b141b0) and [here](https://github.com/canonical/lxd/pull/15362/commits/620069ed5288b0e0413cadeaecdbe4a2051ef4c2). Hopefully. :)

tomponline (approved):
LGTM thanks!

PR #15364, opened 2025-04-10
Storage: Fix missing mount reference counting when attaching ceph VM root volume to another VM

tomponline:
This fixes an issue where detaching a running VM's (v1) root volume (using ceph) from another running VM (v2) (only allowed when `security.shared=true`) was incorrectly triggering an unmount of the attached VM's root volume from the running VM (v1) due to the mount reference counters not being correctly maintained when attaching/detaching v1's root volume from v2.

Reproducer:

```
lxc launch ubuntu:24.04 v1 --vm -s ceph
lxc launch ubuntu:24.04 v2 --vm -s dir

lxc storage volume set ceph virtual-machine/v1 security.shared=true
lxc config device add v2 v1-root disk source=v1 pool=ceph source.type=virtual-machine

# This would trigger an incorrect unmount of running v1's root volumes and fail due to them being in use.
lxc config device remove v2 v1-root 
```

### DIFF
--- a/lxd/device/disk.go
+++ b/lxd/device/disk.go
@@ -1175,6 +1175,16 @@ func (d *disk) startVM() (*deviceConfig.RunConfig, error) {
 					mount.FSType = "iso9660"
 				}
 
+				revertFunc, mountedPath, _, err := d.mountPoolVolume()
+				if err != nil {
+					return nil, diskSourceNotFoundError{msg: "Failed mounting volume", err: err}
+				}
+
+				mount.DevSource = deviceConfig.DevSourcePath{Path: mountedPath}
+				mount.Opts = append(mount.Opts, d.detectVMPoolMountOpts()...)
+
+				revert.Add(revertFunc)
+
 				// If the pool is ceph backed and a block device, don't mount it, instead pass config to QEMU instance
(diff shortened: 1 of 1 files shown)


tomponline:
@MggMuggins this seems to fix the immediate issue causing:

```
+ lxc storage volume detach vmpool10341 virtual-machine/v2 v1
Error: Failed to stop device "v2-root": Failed to run: rbd --id admin --cluster ceph --pool vmpool10341 unmap virtual-machine_test_v2: exit status 16 (rbd: sysfs write failed
rbd: unmap failed: (16) Device or resource busy)
```

https://github.com/canonical/lxd-ci/actions/runs/14374040526/job/40302435408#step:12:1287

And this aligns with what we are doing for VM root volumes when starting the actual VM.

It is, though, somewhat redundant because we are mapping a krbd block device on the host and then never using it (this too aligns with what we do when starting a VM on ceph normally, so at least its consistent behaviour).

By doing this the mount reference counters are maintained correctly as the storage subsystem gets visibility into what is going on.

There is some question around how this will work with our plans to use Exclusive RBD lock mode for VM volumes, but this will need to be resolved more generally, not just for VM root volume attachment.

PR #15369, opened 2025-04-11
Add all projects support for network ACLs (from Incus)

kadinsayani:
Cherry-picked from https://github.com/lxc/incus/pull/838. Tests have also been included.

### DIFF
--- a/lxd/db/network_acls.go
+++ b/lxd/db/network_acls.go
@@ -42,6 +42,39 @@ func (c *ClusterTx) GetNetworkACLs(ctx context.Context, project string) ([]strin
 	return aclNames, nil
 }
 
+// GetNetworkACLsAllProjects returns the names of existing Network ACLs.
+func (c *ClusterTx) GetNetworkACLsAllProjects(ctx context.Context) (map[string][]string, error) {
+	q := `SELECT projects.name, networks_acls.name FROM networks_acls
+		JOIN projects ON projects.id=networks_acls.project_id
+		ORDER BY networks_acls.id
+	`
+
+	aclNames := map[string][]string{}
+	err := query.Scan(ctx, c.tx, q, func(scan func(dest ...any) error) error {
+		var projectName string
+		var aclName string
+
+		err := scan(&projectName, &aclName)
+		if err != nil {
+			return err
+		}
+
+		_, ok := aclNames[projectName]
+		if !ok {
+			aclNames[projectName] = []string{}
+		}
+
+		aclNames[projectName] = append(aclNames[projectName], aclName)
+
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return aclNames, nil
+}
+
 // GetNetworkACLIDsByNames returns a map of names to IDs of existing Network ACLs.
 func (c *ClusterTx) GetNetworkACLIDsByNames(ctx context.Context, project string) (map[string]int64, error) {
 	q := `SELECT id, name FROM networks_acls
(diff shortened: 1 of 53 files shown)


test/suites/network_acl.sh
```diff
@@ -7,16 +7,23 @@ test_network_acl() {
...
+  lxc network acl ls --project testproj3 | grep testacl2
   lxc network acl delete testacl
   lxc network acl delete testacl --project testproj
+  lxc network acl delete testacl2 --project testproj3
   ! lxc network acl ls | grep testacl || false
+  ! lxc network acl ls | grep testacl2 || false
   ! lxc network acl ls --project testproj | grep testacl || false
+  ! lxc network acl ls --project testproj3 | grep testacl2 || false
   lxc project delete testproj
+  lxc project delete testproj3
```
markylaing, line 26: I think for the system tests we're converging on a pattern of:
- When matching positively (checking something is there), use direct equality if possible (`[ "$(lxc ..)" = "<value>" ]`). If matching a substring then use `grep -F` where possible.
- When matching negatively, don't use grep as this can lead to erroneous tests where we're grepping for something that was never there (e.g. due to a typo), and instead use a line count or direct equality

markylaing (changes requested):
Looks good overall, just one nit about tests :)

kadinsayani:
> Looks good overall, just one nit about tests :)

Thanks for your review :)

I've updated the network acl tests to follow your suggested pattern.

tomponline:
please rebase

kadinsayani:
> As with #15392 can you please add a couple of tests to the `tls_restrictions` suite to ensure that a restricted TLS client can still list network ACLs in a project they are restricted to where that project has `features.networks=false`, and that the reported project for those networks is the project they are restricted to (e.g. not `default`).

Thanks for your review. I've added TLS restrictions tests for network ACLs.

lxd/db/network_acls.go
```diff
@@ -42,6 +42,39 @@ func (c *ClusterTx) GetNetworkACLs(ctx context.Context, project string) ([]strin
...
 
+// GetNetworkACLsAllProjects returns the names of existing Network ACLs.
+func (c *ClusterTx) GetNetworkACLsAllProjects(ctx context.Context) (map[string][]string, error) {
+	q := `SELECT projects.name, networks_acls.name FROM networks_acls
+		JOIN projects ON projects.id=networks_acls.project_id
+		ORDER BY networks_acls.id
+	`
+
+	aclNames := map[string][]string{}
+	err := query.Scan(ctx, c.tx, q, func(scan func(dest ...any) error) error {
```
tomponline, line 53: Good to see query.Scan being used here.

tomponline (approved):
Thanks!

lxd/network_acls.go
```diff
@@ -184,25 +228,34 @@ func networkACLsGet(d *Daemon, r *http.Request) response.Response {
...
 			}
 
-			netACLInfo := netACL.Info()
-			netACLInfo.UsedBy, _ = netACL.UsedBy() // Ignore errors in UsedBy, will return nil.
-			netACLInfo.UsedBy = project.FilterUsedBy(s.Authorizer, r, netACLInfo.UsedBy)
-
-			resultMap = append(resultMap, netACLInfo)
-			urlToNetworkACL[entity.NetworkACLURL(requestProjectName, aclName)] = netACLInfo
+			if !recursion {
+				resultString = append(resultString, fmt.Sprintf("/%s/network-acls/%s", version.APIVersion, aclName))
```
tomponline, line 238: just a nit/suggestion for a follow up PR, but here the URL could be generated using the URL builder to ensure aclName is path escaped.

PR #15370, opened 2025-04-11
Refactor network allocation client methods (from Incus)

kadinsayani:
This refactors network allocation client methods: extracts `GetNetworkAllocationsAllProjects` method from
`GetNetworkAllocations` and updates call site to use new function when `c.flagAllProjects` is `true` (includes cherry-picks from https://github.com/lxc/incus/pull/838).

### DIFF
--- a/po/ar.po
+++ b/po/ar.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: lxd\n"
 "Report-Msgid-Bugs-To: lxd@lists.canonical.com\n"
-"POT-Creation-Date: 2025-03-03 09:58-0600\n"
+"POT-Creation-Date: 2025-04-17 15:05-0600\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Automatically generated\n"
 "Language-Team: none\n"
@@ -60,7 +60,7 @@ msgid ""
 "###   size: \"61203283968\""
 msgstr ""
 
-#: lxc/config.go:1241
+#: lxc/config.go:1242
 msgid ""
 "### This is a YAML representation of the UEFI variables configuration.\n"
 "### Any line starting with a '# will be ignored.\n"
@@ -189,7 +189,7 @@ msgid ""
 "### Note that the name is shown but cannot be modified"
 msgstr ""
 
-#: lxc/image.go:419
+#: lxc/image.go:420
 msgid ""
 "### This is a YAML representation of the image properties.\n"
 "### Any line starting with a '# will be ignored.\n"
@@ -417,7 +417,7 @@ msgstr ""
 msgid "%d (id: %d, online: %v, NUMA node: %v)"
 msgstr ""
 
-#: lxc/image.go:1180
+#: lxc/image.go:1181
 #, c-format
 msgid "%s (%d more)"
 msgstr ""
@@ -427,12 +427,12 @@ msgstr ""
 msgid "%s (%s) (%d available)"
 msgstr ""
 
-#: lxc/file.go:1175
+#: lxc/file.go:1179
 #, c-format
 msgid "%s is not a directory"
 msgstr ""
 
-#: lxc/file.go:1065
+#: lxc/file.go:1069
 #, c-format
 msgid "'%s' isn't a supported file type"
 msgstr ""
@@ -472,7 +472,7 @@ msgstr ""
 msgid "--empty cannot be combined with an image name"
 msgstr ""
 
-#: lxc/config.go:497 lxc/config.go:817
+#: lxc/config.go:497 lxc/config.go:818
 msgid "--expanded cannot be used with a server"
 msgstr ""
 
@@ -492,7 +492,7 @@ msgstr ""
 msgid "--refresh can only be used with instances"
 msgstr ""
 
-#: lxc/config.go:170 lxc/config.go:447 lxc/config.go:636 lxc/config.go:843
+#: lxc/config.go:170 lxc/config.go:447 lxc/config.go:637 lxc/config.go:844
 #: lxc/info.go:461
 msgid "--target cannot be used with instances"
 msgstr ""
@@ -521,11 +521,11 @@ msgstr ""
 msgid "<remote> <new-name>"
 msgstr ""
 
-#: lxc/file.go:699
+#: lxc/file.go:701
 msgid "<source path>... [<remote>:]<instance>/<path>"
 msgstr ""
 
-#: lxc/image.go:695
+#: lxc/image.go:696
 msgid ""
 "<tarball>|<directory>|<URL> [<rootfs tarball>] [<remote>:] [key=value...]"
 msgstr ""
@@ -538,15 +538,15 @@ msgstr ""
 msgid "ADDRESS"
 msgstr ""
 
-#: lxc/alias.go:139 lxc/image.go:1143 lxc/image_alias.go:267
+#: lxc/alias.go:139 lxc/image.go:1144 lxc/image_alias.go:267
 msgid "ALIAS"
 msgstr ""
 
-#: lxc/image.go:1144
+#: lxc/image.go:1145
 msgid "ALIASES"
 msgstr ""
 
-#: lxc/cluster.go:195 lxc/image.go:1138 lxc/list.go:562
+#: lxc/cluster.go:195 lxc/image.go:1139 lxc/list.go:562
 msgid "ARCHITECTURE"
 msgstr ""
 
@@ -737,7 +737,7 @@ msgstr ""
 msgid "Aliases already exists: %s"
 msgstr ""
 
-#: lxc/image.go:1045
+#: lxc/image.go:1046
 msgid "Aliases:"
 msgstr ""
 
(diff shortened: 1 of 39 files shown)


markylaing (changes requested):
I think if we're going for `api.NewURL` across the board in client functions it's worth change all (internal) methods to use an `api.URL`. Otherwise, every time we call `r.setQueryAttributes` we're parsing the URL again to add query parameters, then calling `String` on it again.

Also small nit but can we have smaller commit messages with a description please? (Subject to others opinion on this, it's just a preference really)

client/lxd_networks.go
```diff
@@ -16,14 +15,14 @@ func (r *ProtocolLXD) GetNetworkNames() ([]string, error) {
...
-	_, err = r.queryStruct("GET", baseURL, nil, "", &urls)
+	u := api.NewURL().Path("networks")
+	_, err = r.queryStruct(http.MethodGet, u.String(), nil, "", &urls)
 	if err != nil {
 		return nil, err
 	}
 
 	// Parse it.
-	return urlsToResourceNames(baseURL, urls...)
+	return urlsToResourceNames(u.String(), urls...)
```
simondeziel, line 25: Same note about caching `u.String()`.

client/lxd_network_acls.go
```diff
@@ -73,13 +74,14 @@ func (r *ProtocolLXD) GetNetworkACLLogfile(name string) (io.ReadCloser, error) {
 	}
 
 	// Prepare the HTTP request
-	url := r.httpBaseURL.String() + "/1.0/network-acls/" + url.PathEscape(name) + "/log"
+	u := api.NewURL().Path("network-acls", name, "log")
+	url := r.httpBaseURL.String() + "/1.0" + u.String()
```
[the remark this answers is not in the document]
kadinsayani: Here we don't use our internal `query*` methods but rather `http.NewRequest`, so we need to generate `url` with  `https://` and `/1.0` before making the call to `NewRequest`. Normally, generation of the `url` is done by `query`.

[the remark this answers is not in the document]
kadinsayani: We use `http.NewRequest` because we return a reader for the ACL log file, with the caller taking responsibility for closing the reader.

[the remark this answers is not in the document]
kadinsayani: > `api.NewURL().Path()` can accept multiple path components so would it make sense to have the `/1.0` bit added when building `u`?

Yes, thanks for the suggestion :)

kadinsayani:
> I think if we're going for `api.NewURL` across the board in client functions it's worth change all (internal) methods to use an `api.URL`. Otherwise, every time we call `r.setQueryAttributes` we're parsing the URL again to add query parameters, then calling `String` on it again.
> 
> Also small nit but can we have smaller commit messages with a description please? (Subject to others opinion on this, it's just a preference really)

For now, I've incorporated @simondeziel's suggestions to reduce the `String()` calls, that way we can merge this PR which also includes cherry-picks for `GetNetworkAllocationsAllProjects`. I'm going to open a follow up PR which includes broader client refactorings including your suggested approach of passing around `*api.URL`, and I'll also update all client methods accordingly (replacing HTTP method string literals with `http` constants as well). 

I've also gone ahead and used smaller commit messages with description. I like this approach better as well, thanks for the suggestion :)

client/lxd_network_acls.go
```diff
@@ -73,13 +75,13 @@ func (r *ProtocolLXD) GetNetworkACLLogfile(name string) (io.ReadCloser, error) {
 	}
 
 	// Prepare the HTTP request
-	url := r.httpBaseURL.String() + "/1.0/network-acls/" + url.PathEscape(name) + "/log"
+	url := api.NewURL().Path(r.httpBaseURL.String(), "1.0", "network-acls", name, "log").String()
```
simondeziel, line 78: Is `r.httpBaseURL.String()` actually a Path? Since it comes before `/1.0`, I'd expect it to be a scheme+hostname (like `https://example.com`), no?

simondeziel (commented):
LGTM with a tiny question.

client/lxd_network_acls.go
kadinsayani: Yes, you're right. That should not be included in the `Path()` call since an extra `/` would be included.

simondeziel: Ah OK so `r.httpBaseURL.String()` contained the scheme+hostname+rootpath (`https://example.com/`). Cool!

simondeziel (dismissed):

simondeziel (approved):

client/interfaces.go
```diff
@@ -293,7 +293,8 @@ type InstanceServer interface {
 	DeleteNetworkACL(name string) (err error)
 
 	// Network allocations functions ("network_allocations" API extension)
-	GetNetworkAllocations(allProjects bool) (allocations []api.NetworkAllocations, err error)
+	GetNetworkAllocations() (allocations []api.NetworkAllocations, err error)
```
tomponline, line 296: This is a Go SDK API break @kadinsayani and not acceptable I'm afraid.

tomponline (changes requested):
The signature change of `GetNetworkAllocations(allProjects bool)` to `GetNetworkAllocations()` and `GetNetworkAllocationsAllProjects()` is a LXD Go SDK API break.

Any user of `GetNetworkAllocations(allProjects bool)` would have their code break if I were to merge this PR.

We should always be aware of API breaks in the various forms and work hard to avoid them, or at the very least flag them up in the PR description so they can be discussed and evaluated rather than proposing them without consideration.

kadinsayani:
> I think if we're going for `api.NewURL` across the board in client functions it's worth change all (internal) methods to use an `api.URL`. Otherwise, every time we call `r.setQueryAttributes` we're parsing the URL again to add query parameters, then calling `String` on it again.

I've opened https://github.com/canonical/lxd/pull/15433 to replace all http string literals with `http` package constants. I've also started refactoring internal methods to use `api.URL` rather than raw `string` URLs. This is a fairly large refactoring with many implications across the client library, including breaking changes to interfaces such as:

https://github.com/canonical/lxd/blob/53b22c049db603c08d0ff935f63f20cff316bf66/client/interfaces.go#L466-L469

We likely won't observe significant performance improvements by reducing our `String()` calls across the client library when using the LXD snap. Additionally, our typical call chain looks like: `SomeClientFunction` → `queryStruct` → `query` → `rawQuery` → `http.NewRequestWithContext`, where the final method requires a string URL anyway. This raises the question of whether it's worth always using `api.URL` internally, given that we'll still end up calling `String()` at the end of the chain. An alternative might be to standardize on always using `string` URLs and introduce a helper method for safe URL construction and escaping.

Happy to go over what I’ve done so far during our 1:1 and chat about the best way forward :)

tomponline:
> An alternative might be to standardize on always using `string` URLs and introduce a helper method for safe URL construction and escaping.

api.URL and its functions are those helper functions :)

tomponline:
> including breaking changes to interfaces such as

Breaking changes to the public API SDK interface are not acceptable for this change, so will need to accept/return strings for URLs at that boundary point.

tomponline:
> > including breaking changes to interfaces such as
> 
> Breaking changes to the public API SDK interface are not acceptable for this change, so will need to accept/return strings for URLs at that boundary point.

But you can add unexported wrapper functions that accept non-string URLs if that is helpful

kadinsayani:
> api.URL and its functions are those helper functions :)

True, though we still need to call `String()` on an `api.URL` before calling `http.NewRequestWithContext`. We wouldn't have this problem with `string` literals and a helper function for cases such as `fmt.Sprintf("/%s/networks/%s/forwards/%s", version.APIVersion, url.PathEscape(n.Name()), url.PathEscape(listenAddress))`. We could have a helper function that wraps `fmt.Sprintf` and path escapes all parts.

> But you can add unexported wrapper functions that accept non-string URLs if that is helpful, as if its truly internal, then the like of rawQuery() would be fine.

I also tried this approach:

```Go
func (r *ProtocolLXD) buildRequestURL(apiURL *api.URL, params ...string) string {
	// Add project/target
	r.setQueryAttributes(apiURL)

	// Generate the URL
	return r.httpBaseURL.String() + "/1.0" + apiURL.String()
}
```

It works fine, but it only reduces 1 `String()` call in the stack for `SomeClientFunction`. So the benefit might be pretty minimal overall.

tomponline:
> > api.URL and its functions are those helper functions :)
> 
> True, though we still need to call `String()` on an `api.URL` before calling `http.NewRequestWithContext`. We wouldn't have this problem with `string` literals and a helper function for cases such as `fmt.Sprintf("/%s/networks/%s/forwards/%s", version.APIVersion, url.PathEscape(n.Name()), url.PathEscape(listenAddress))`. We could have a helper function that wraps `fmt.Sprintf` and path escapes all parts.

I've not got the full context clearly. What is the problem with callinf u.String()?

tomponline:
> fmt.Sprintf("/%s/networks/%s/forwards/%s", version.APIVersion, url.PathEscape(n.Name()), url.PathEscape(listenAddress))

Building URLs like this is exactly what we're trying to avoid.

PR #15392, opened 2025-04-14
Add all projects support for networks (from Incus) 

kadinsayani:
Cherry-picked from https://github.com/lxc/incus/pull/838. Tests have also been included.

### DIFF
--- a/lxd/db/networks.go
+++ b/lxd/db/networks.go
@@ -480,6 +480,36 @@ func (c *ClusterTx) GetNetworks(ctx context.Context, project string) ([]string, 
 	return c.networks(ctx, project, "")
 }
 
+// GetNetworksAllProjects returns the names of all networks across all projects.
+func (c *ClusterTx) GetNetworksAllProjects(ctx context.Context) (map[string][]string, error) {
+	q := "SELECT projects.name, networks.name FROM networks JOIN projects ON networks.project_id=projects.id"
+
+	networkNames := map[string][]string{}
+	err := query.Scan(ctx, c.tx, q, func(scan func(dest ...any) error) error {
+		var projectName string
+		var networkName string
+
+		err := scan(&projectName, &networkName)
+		if err != nil {
+			return err
+		}
+
+		_, ok := networkNames[projectName]
+		if !ok {
+			networkNames[projectName] = []string{}
+		}
+
+		networkNames[projectName] = append(networkNames[projectName], networkName)
+
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return networkNames, nil
+}
+
 // Get all networks matching the given WHERE filter (if given).
 func (c *ClusterTx) networks(ctx context.Context, project string, where string, args ...any) ([]string, error) {
 	q := "SELECT name FROM networks WHERE project_id = (SELECT id FROM projects WHERE name = ?)"
--- a/lxd/networks.go
+++ b/lxd/networks.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
@@ -151,6 +152,11 @@ func networkAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *http.
 //      description: Project name
 //      type: string
 //      example: default
+//    - in: query
+//      name: all-projects
+//      description: Retrieve networks from all projects
+//      type: boolean
+//      example: true
 //  responses:
 //    "200":
 //      description: API endpoints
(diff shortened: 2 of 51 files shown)


kadinsayani:
Moved back to draft as tests are failing.

tomponline:
please rebase

tomponline:
@markylaing please can you approve. Ta

lxd/networks.go
```diff
@@ -258,20 +281,36 @@ func networksGet(d *Daemon, r *http.Request) response.Response {
...
+				return err
+			}
+		} else {
+			// Get list of managed networks (that may or may not have network interfaces on the host).
+			networkNames, err := tx.GetNetworks(ctx, effectiveProjectName)
+			if err != nil {
+				return err
+			}
+
+			networks[managed][effectiveProjectName] = networkNames
```
markylaing, line 307: Slight concern here using the effective project name in this map. I think it needs to be the requested project name because later when ranging over this map, we're using the effective project name in the call to `userHasPermission`. Auth checks should always use the requested project name (the fine-grained authorizer will add replace the requested project with the effective project, but the TLS authorizer uses the requested project directly to account for project features).

Can we add a test to verify whether or not this is working as expected and then make changes here as required? The test should be in the `tls_restrictions` suite and should:
1. Create a project `foo` with `features.networks=false`
2. Create a restricted TLS client with access only to project `foo`. 
3. List networks in project `foo` as the TLS client. They should see networks in the default project, and the project reported on those networks should state `foo` and not `default`.

markylaing (changes requested):
Just the one concern about using the effective project in the authorization check :)

lxd/networks.go
kadinsayani: Yes, the map should store `requestProjectName`, thanks for pointing that out :)

Maybe I'm overlooking something, but I think for the test case you've described the project reported would be `default` since we disallow `requestProjectName` in all projects requests, and we include `effectiveProjectName` in the response, which is the `default` project when `features.networks=false`.

markylaing: The networks in project `foo` with `features.networks=false` should still have `foo` as the project, even if they are in the default project. 

These wouldn't be displayed via `lxc network list --project foo` because the project column will not be displayed, but the API response should have `foo` in the project field.

So the `requestProjectName` should be in the API response, not the `effectiveProjectName`.

kadinsayani: Got it, thanks for clarifying!

markylaing (approved):
LGTM thanks!

lxd/db/networks.go
```diff
@@ -480,6 +480,42 @@ func (c *ClusterTx) GetNetworks(ctx context.Context, project string) ([]string,
...
 
+// GetNetworksAllProjects returns the names of all networks across all projects.
+func (c *ClusterTx) GetNetworksAllProjects(ctx context.Context) (map[string][]string, error) {
+	q := "SELECT projects.name, networks.name FROM networks JOIN projects ON networks.project_id=projects.id"
+
+	var projectName string
+	var networkName string
+	outfmt := []any{projectName, networkName}
+
+	result, err := queryScan(ctx, c, q, nil, outfmt)
```
tomponline, line 491: @kadinsayani please can you use query.Scan instead, we are phasing out queryScan

lxd/networks.go
```diff
@@ -313,23 +352,25 @@ func networksGet(d *Daemon, r *http.Request) response.Response {
...
+			for _, networkName := range networkNames {
+				// Filter out managed networks that the caller doesn't have permission to view.
+				if kind == managed && !userHasPermission(entity.NetworkURL(projectName, networkName)) {
 					continue
 				}
 
-				resultMap = append(resultMap, &net)
-				urlToNetwork[entity.NetworkURL(requestProjectName, networkName)] = &net
+				if !recursion {
+					resultString = append(resultString, "/"+version.APIVersion+"/networks/"+networkName)
```
tomponline, line 364: here's a good place to use URL builder so that networkName is path escaped automatically

tomponline, line 364: here's a good place to use URL builder so that networkName is path escaped automatically

tomponline (changes requested):
We're phasing out the `queryScan` function as it requires the somewhat awkward creation of `outfmt := []any{name}` slice rather than just being able to scan directly into a struct or variable.

lxd/networks.go
kadinsayani: Thanks, fixed.

lxd/db/networks.go
kadinsayani: Sure thing, fixed.

kadinsayani:
> @kadinsayani if you have time & interest, then a separate PR to switch over from queryScan to query.Scan and remove queryScan altogether would be great.

https://github.com/canonical/lxd/pull/15444 - I've kept it as draft for now while tests are running.

tomponline (approved):
Thanks!

PR #15396, opened 2025-04-15
Auth: Filter cluster group used by URLs

markylaing:
Follow up to https://github.com/canonical/lxd/pull/15119

Cluster group used-by URLs we're not being filtered by what the caller is able to view. This allowed restricted users to see the URLs of projects that they do not have access to.

### DIFF
--- a/lxd/api_cluster.go
+++ b/lxd/api_cluster.go
@@ -36,6 +36,7 @@ import (
 	"github.com/canonical/lxd/lxd/lifecycle"
 	"github.com/canonical/lxd/lxd/node"
 	"github.com/canonical/lxd/lxd/operations"
+	"github.com/canonical/lxd/lxd/project"
 	"github.com/canonical/lxd/lxd/project/limits"
 	"github.com/canonical/lxd/lxd/request"
 	"github.com/canonical/lxd/lxd/response"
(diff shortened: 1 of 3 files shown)


lxd/api_project.go
```diff
@@ -1244,7 +1244,25 @@ func projectValidateConfig(s *state.State, config map[string]string, defaultNetw
...
 		//  type: string
 		//  shortdesc: Cluster groups that can be targeted
-		"restricted.cluster.groups": validate.Optional(validate.IsListOf(validate.IsAny)),
+		"restricted.cluster.groups": func(value string) error {
+			if value == "" {
+				return nil
+			}
+
+			groupNames := shared.SplitNTrimSpace(value, ",", -1, true)
+			return s.DB.Cluster.Transaction(s.ShutdownCtx, func(ctx context.Context, tx *db.ClusterTx) error {
```
tomponline, line 1253: Rather than a new ClusterGroupsExist function, can we use `GetClusterGroups` and then test for which item(s) weren't found and either return the first one not to be found or all of the not found ones? I feel this would be more useful to the end user and avoid the need for a new query function.

lxd/api_cluster.go
```diff
@@ -36,6 +36,7 @@ import (
 	"github.com/canonical/lxd/lxd/lifecycle"
 	"github.com/canonical/lxd/lxd/node"
 	"github.com/canonical/lxd/lxd/operations"
+	projecthelpers "github.com/canonical/lxd/lxd/project"
```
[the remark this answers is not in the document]
markylaing: For some reason I thought it was preferred to consistently use that alias. I'll remove it.

lxd/api_project.go
markylaing: Done

tomponline (approved):
Cheers

PR #15398, opened 2025-04-15
Instance: Run CPU scheduler after DB changes made (from Incus)

tomponline:
This isn't a cherry-pick, but rather inspiration from a change in Incus.
At this point LXD's code base has diverged somewhat from Incus and the exact same bug was not present, however I have taken the point and run with it by making the scheduler re-run occur after 2 other places that might impact the scheduler.

Inspired by https://github.com/lxc/incus/pull/1938

### DIFF
--- a/lxd/instance/drivers/driver_lxc.go
+++ b/lxd/instance/drivers/driver_lxc.go
@@ -2534,14 +2534,14 @@ func (d *lxc) onStart(_ map[string]string) error {
 		return err
 	}
 
-	// Trigger a rebalance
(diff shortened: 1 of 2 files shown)


tomponline:
> I'll let the analysis to Aleks but the changes themselves LGTM.
> 
> As for the attribution, kernel.org seems to use an `Inspired-by: <link to patch>` header, see [here](https://cdn.kernel.org/pub/linux/kernel/v5.x/ChangeLog-5.4.281).

Updated.

mihalicyn (approved):
LGTM

PR #15401, opened 2025-04-15
doc: use external file to store redirects

minaelee:
Currently there are ~40+ files in the `doc/` directory that should be inside one of the category directories (`doc/howto`, etc). When those files are moved, their URLs will change, so we will need to redirect them to preserve links from external sources.

Redirects are handled by Sphinx using a `redirects` dict currently defined in `custom_conf.py`. Adding ~40+ new redirects to this dict would clutter `custom_conf.py`. This PR prepares for the move by extracting that dict to a new `redirects.py`, which is then imported by `custom_conf.py`.

### DIFF
--- a/doc/custom_conf.py
+++ b/doc/custom_conf.py
@@ -5,6 +5,7 @@ import subprocess
 import yaml
 from git import Repo
 import filecmp
+from redirects import redirects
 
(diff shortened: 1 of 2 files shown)


tomponline (approved):
Good idea, much easier to find! Cheers

PR #15403, opened 2025-04-16
Fix: Return cluster members in non-recursive mode

roosterfish:
This fixes a regression introduced in https://github.com/canonical/lxd/pull/15314/commits/cc14d2b45785a58b74315ef581ccd6753d5ed563.

When querying `GET /1.0/cluster/members` without `recursive=1` a simple list of cluster members should be returned.
This was causing errors in MicroCloud as it polls LXD until the API returns the first cluster member after enabling clustering. As an example see https://github.com/canonical/microcloud/actions/runs/14488020785/job/40637889229?pr=754#step:3:4054


tomponline (approved):
Great catch thanks!

PR #15406, opened 2025-04-16
Doc: Ignore link checking of https://www.schlachter.tech/solutions/pongo2-template-engine/

tomponline:
Seems to be denied when running from Github (403 response).


simondeziel (approved):
Thanks, it's been bugging me too :)

PR #15408, opened 2025-04-16
Placement Rulesets

markylaing:
This PR takes a different approach to instance placement than in #15277. Rather than adding placement rules directly to instances and profiles, rules are added to a new top-level API type called a `PlacementRuleset`. 

Placement rulesets are referenced by name and unique within their project. They are referenced by instance or profile configuration via the `placement.ruleset` config key.

Still to do:
- [ ] Apply rulesets on cluster member evacuation
- [ ] Test ruleset permissions and relationship with cluster groups
- [ ] Test warnings are created when a ruleset is updated and running instances are not conformant with new configuration.
- [ ] Test a warning is created when an instance is updated with a ruleset configuration and is not conformant.
- [ ] Documentation


client/util.go
```diff
@@ -252,6 +253,32 @@ func urlsToResourceNames(matchPathPrefix string, urls ...string) ([]string, erro
...
+// the final element of each given URL.
+func urlsToResourceNamesAllProjects(matchPathPrefix string, urls ...string) (map[string][]string, error) {
+	resourceNames := make(map[string][]string)
+	for _, urlRaw := range urls {
+		u, err := url.Parse(urlRaw)
+		if err != nil {
+			return nil, fmt.Errorf("Failed parsing URL %q: %w", urlRaw, err)
+		}
+
+		_, after, found := strings.Cut(u.Path, matchPathPrefix+"/")
```
simondeziel, line 266: I'd suggest using `strings.CutPrefix`

lxd/placement/filtering.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+
+	isSelfMatch := func(selector cluster.Selector) bool {
+		if entity.Type(selector.EntityType) != entity.TypeInstance {
+			return false
+		}
+
+		for _, m := range selector.Matchers {
+			_, key, ok := strings.Cut(m.Property, "config.")
```
simondeziel, line 48: `strings.CutPrefix`?

simondeziel (commented):
2 small drive-by comments :)

client/util.go
markylaing: Ah copy/pasto

lxd/placement/filtering.go
markylaing: copy/pasto again from a time before I knew that CutPrefix existed ;)

tomponline:

PR #15413, opened 2025-04-17
lxd/instance_instance_types: Update instanceRefreshTypes function to use all.yaml instance type file (stable-5.0)

kadinsayani:
Includes https://github.com/canonical/lxd/pull/15132.

### DIFF
--- a/lxd/instance_instance_types.go
+++ b/lxd/instance_instance_types.go
@@ -155,9 +155,8 @@ func instanceRefreshTypes(ctx context.Context, s *state.State) error {
 		_ = instanceLoadCache()
 	}
 
-	// Get the list of instance type sources
-	sources := map[string]string{}
-	err := downloadParse(".yaml", &sources)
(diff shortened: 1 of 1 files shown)


simondeziel (commented):
@kadinsayani LGTM but I'm curious if you had the chance to test it locally because we don't have any test exercising this part of the code due to the unwanted network dependency.

Removing `/var/snap/lxd/common/lxd/cache/instance_types.yaml` and seeing it coming back after restarting LXD should be a good starters.

kadinsayani:
> @kadinsayani LGTM but I'm curious if you had the chance to test it locally because we don't have any test exercising this part of the code due to the unwanted network dependency.
> 
> Removing `/var/snap/lxd/common/lxd/cache/instance_types.yaml` and seeing it coming back after restarting LXD should be a good starters.

Yes, all good locally ✅

simondeziel (approved):

tomponline (approved):

PR #15423, opened 2025-04-19
doc/idmap: Clarify subuid/subgid configuration (from Incus)

minaelee:
Cherry-picked from https://github.com/lxc/incus/pull/1353
Adapted during cherry-pick to use LXD terminology (say LXD where it says Incus)
Confirmed that the change from `lxd/incus` to `root` for user name occurred in [this 2017 commit](https://github.com/canonical/lxd/commit/1f6218101f16c1a1d5521dc5bfd97d532626d03b)

### DIFF
--- a/doc/userns-idmap.md
+++ b/doc/userns-idmap.md
@@ -25,7 +25,7 @@ kernels but will refuse to start containers.
 ## Allowed ranges
 
 On most hosts, LXD will check `/etc/subuid` and `/etc/subgid` for
-allocations for the `lxd` user and on first start, set the default
+allocations for the `root` user and on first start, set the default
 profile to use the first 65536 UIDs and GIDs from that range.
 
 If the range is shorter than 65536 (which includes no range at all),
@@ -90,3 +90,7 @@ host, and the third entry is the range inside the container. These ranges must
 be the same size.
 
 This property requires a container reboot to take effect.
+
+Remember that you may need to add an entry for the `root` user into
+`/etc/subid` and/or `/etc/subgid` so the container is allowed to make
+use of it.


tomponline:
@minaelee i found this https://ubuntu.com/blog/custom-user-mappings-in-lxd-containers

> The maps for “lxd” and “root” should always be kept in sync. LXD itself is restricted by the “root” allocation. The “lxd” entry is used to track what needs to be removed if LXD is uninstalled.

mihalicyn:
> Also I think (but need to check) that with the LXD snap it doesnt respect the /etc/sub{uid|gid} files anyway, so we should probably state that if true.

yeah, LXD when installed from official snap doesn't use `newuidmap`/`newgidmap` tools. So we don't care about `/etc/sub{u,g}id` files in this case.

Typical log messages (from `less /var/snap/lxd/common/lxd/logs/microk8s-unpriv/lxc.log`):
```
lxc microk8s-unpriv 20250326070830.659 INFO     idmap_utils - ../src/src/lxc/idmap_utils.c:lxc_map_ids:165 - newuidmap binary is missing
lxc microk8s-unpriv 20250326070830.659 INFO     idmap_utils - ../src/src/lxc/idmap_utils.c:lxc_map_ids:171 - newgidmap binary is missing
lxc microk8s-unpriv 20250326070830.659 DEBUG    idmap_utils - ../src/src/lxc/idmap_utils.c:lxc_map_ids:186 - No newuidmap and newgidmap binary found. Trying to write directly with euid 0
lxc microk8s-unpriv 20250326070
```

mihalicyn (approved):

tomponline:
@minaelee would you mind following up with a PR that explains this section is not relevant for snap users?

minaelee:
> @minaelee would you mind following up with a PR that explains this section is not relevant for snap users?

Addressed in https://github.com/canonical/lxd/pull/15547

PR #15433, opened 2025-04-22
Replace http string literals with `http` package constants

kadinsayani:
Replaces http method string literals with `http` package constants across entire codebase.

### DIFF
--- a/lxd/main_shutdown.go
+++ b/lxd/main_shutdown.go
@@ -66,7 +66,7 @@ func (c *cmdShutdown) Run(cmd *cobra.Command, args []string) error {
 		httpTransport := httpClient.Transport.(*http.Transport)
 		httpTransport.ResponseHeaderTimeout = 3600 * time.Second
 
-		_, _, err = d.RawQuery("PUT", fmt.Sprintf("/internal/shutdown?%s", v.Encode()), nil, "")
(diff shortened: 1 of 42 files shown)


simondeziel (commented):
The changes look good but have turned up some more issues with linters :)

kadinsayani:
> The changes look good but have turned up some more issues with linters :)

I've added fixes for the lint errors :)

lxd/main_shutdown.go
```diff
@@ -63,7 +63,12 @@ func (c *cmdShutdown) Run(cmd *cobra.Command, args []string) error {
 		}
 
 		// Request shutdown, this shouldn't return until daemon has stopped so use a large request timeout.
-		httpTransport := httpClient.Transport.(*http.Transport)
+		httpTransport, ok := httpClient.Transport.(*http.Transport)
+		if !ok {
+			chResult <- fmt.Errorf("httpClient.Transport is not *http.Transport")
```
tomponline, line 70: for future reference using %T here of the httpClient.Transport value would be useful to see what it is.

tomponline (approved):

tomponline:
ta

lxd/main_shutdown.go
kadinsayani: Thanks, I will keep that in mind :)

PR #15435, opened 2025-04-23
devlxd: Enable access to GET devlxd/1.0 (regression)

MusicDin:
Fixes the regression introduced in https://github.com/canonical/lxd/pull/15317 - the devLXD endpoint `GET /1.0` should be allowed despite `security.devlxd` not being enabled.

### DIFF
--- a/lxd/devlxd.go
+++ b/lxd/devlxd.go
@@ -95,7 +95,7 @@ var devLXD10Endpoint = devLXDAPIEndpoint{
 }
 
 func devLXDAPIGetHandler(d *Daemon, r *http.Request) response.Response {
-	inst, err := getInstanceFromContextAndCheckSecurityFlags(r.Context(), devLXDSecurityKey)
+	inst, err := getInstanceFromContextAndCheckSecurityFlags(r.Context())
 	if err != nil {
 		return response.DevLXDErrorResponse(err, inst != nil && inst.Type() == instancetype.VM)
 	}
--- a/lxd/response/response.go
+++ b/lxd/response/response.go
@@ -45,7 +45,7 @@ type Response interface {
 	String() string
 }
 
-// Devlxd response.
+// devLXDResponse represents a devLXD API response.
 type devLXDResponse struct {
 	content     any
 	code        int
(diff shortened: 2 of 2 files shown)


tomponline (commented):
Can we have a test for this?

MusicDin:
The test is actually in `lxd-ci` (and it failed) as this only applies to VMs when enabling `security.devlxd`. 

What happened here:
- The devlxd is enabled by default.
- The devlxd is disabled by setting `security.devlxd=false`.
- The devlxd is re-enabled by setting `security.devlxd=true` (or unsetting `security.devlxd`). When devlxd is enabled, the lxd-agent establishes connection with LXD. After successful connection, the `/1.0` is queried (`GetServer()`) to retrieve server information (devlxd in this context). This failed, because to update the `security.devlxd` first the connection needs to be established from the agent, which queries `/1.0` that is not enabled yet.

tomponline:
> The test is actually in `lxd-ci` (and it failed) as this only applies to VMs when enabling `security.devlxd`.

ah ok if its VM only then we can only do it in lxd-ci right now

tomponline (approved):

lxd/devlxd.go
```diff
@@ -95,7 +95,7 @@ var devLXD10Endpoint = devLXDAPIEndpoint{
 }
 
 func devLXDAPIGetHandler(d *Daemon, r *http.Request) response.Response {
-	inst, err := getInstanceFromContextAndCheckSecurityFlags(r.Context(), devLXDSecurityKey)
+	inst, err := getInstanceFromContextAndCheckSecurityFlags(r.Context())
```
tomponline, line 98: is this handler only used for the lxd-agent vsock API, or is it also used for the container devlxd API too?

MusicDin: > is this handler only used for the lxd-agent vsock API, or is it also used for the container devlxd API too?

It is used for both containers and VMs (vsock).

> I'm wondering whether we can test this policy in the main test suite using the go based devlxd client you added for testing.

Hmm, maybe there would be a way, but when `security.devlxd` is disabled, this API is not accessible.

tomponline: isnt that worth testing?

PR #15436, opened 2025-04-23
Clustering shutdown improvements

tomponline:


lxd/main_daemon.go
```diff
@@ -91,9 +86,23 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
...
+	err := d.Init()
+	if err != nil {
+		logger.Error("Failed starting daemon", logger.Ctx{"err": err})
+
+		// Only trigger d.Stop() and return start up error if manual shutdown hasn't been requested.
+		// This avoids calling d.Stop() twice and ensures start up errors don't trigger systemd restart.
+		if d.shutdownCtx.Err() == nil {
+			// If an error occurred while starting up, try to cleanup any setup done so far.
+			// Return the error from the failed d.Init() call so the original start up error can be
+			// translated into an exist status and considered by systemd (which may restart).
```
markylaing, line 101: `exist -> exit`

tomponline: ta fixed

markylaing (commented):
Just the typo and a few things to double check my understanding:
1. The `startStopLock` is locked on `Init` and also on `Stop`, is my understanding correct that `Init` will return when the context is cancelled, releasing the lock for use in `Stop`?
2. Is it possible that any networks or storage have been initialised if the database has not yet been initialised? If so, could `(*Daemon).Init` potentially return a reverter so that the changes can be reverted if the shutdown context is cancelled?

markylaing (approved):
LGTM :+1:

PR #15440, opened 2025-04-24
lxd: Improve clustering start up checks

tomponline:
In some cases the `cluster.https_address` can become unset in the local DB and this can cause LXD to try and start up
in standalone mode and enter a start up loop that cannot be easily stopped.

Adds sense check to detect when this situation occurs.

### DIFF
--- a/lxd/daemon.go
+++ b/lxd/daemon.go
@@ -1362,6 +1362,20 @@ func (d *Daemon) init() error {
 		}
 	}
 
+	// Load local config (must come after processing incoming recovery tarball as it can update local config).
+	logger.Info("Loading daemon configuration")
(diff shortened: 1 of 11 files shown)


lxd/daemon.go
```diff
@@ -1362,6 +1362,20 @@ func (d *Daemon) init() error {
 		}
 	}
 
+	// Load local config (must come after processing incoming recovery tarball as it can update local config).
+	logger.Info("Loading daemon configuration")
+	err = d.db.Node.Transaction(context.TODO(), func(ctx context.Context, tx *db.NodeTx) error {
```
MusicDin, line 1367: Since it is read operation, can we use at shutdown context?

MusicDin (approved):

PR #15443, opened 2025-04-24
Random assorted tweaks

simondeziel:
Those tweaks were accumulated while manually reviewing the commits moving away from `fmt.Sprintf()` and `fmt.Sprint()`.

### DIFF
--- a/lxd-agent/state.go
+++ b/lxd-agent/state.go
@@ -200,7 +200,7 @@ func networkState() map[string]api.InstanceStateNetwork {
 			}
 
 			scope := "global"
-			if strings.HasPrefix(address, "127") {
+			if strings.HasPrefix(address, "127.") {
 				scope = "local"
 			}
 
--- a/lxd-agent/state.go
+++ b/lxd-agent/state.go
@@ -189,36 +189,36 @@ func networkState() map[string]api.InstanceStateNetwork {
 		addrs, _ := iface.Addrs()
 
 		for _, addr := range addrs {
-			addressFields := strings.SplitN(addr.String(), "/", 2)
-			if len(addressFields) != 2 {
+			address, netmask, found := strings.Cut(addr.String(), "/")
+			if !found {
 				continue
 			}
 
 			networkAddress := api.InstanceStateNetworkAddress{
-				Address: addressFields[0],
-				Netmask: addressFields[1],
+				Address: address,
+				Netmask: netmask,
 			}
 
 			scope := "global"
-			if strings.HasPrefix(addressFields[0], "127") {
+			if strings.HasPrefix(address, "127") {
 				scope = "local"
 			}
 
-			if addressFields[0] == "::1" {
+			if address == "::1" {
 				scope = "local"
 			}
 
-			if strings.HasPrefix(addressFields[0], "169.254") {
+			if strings.HasPrefix(address, "169.254") {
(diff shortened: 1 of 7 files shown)


lxd-agent/state.go
```diff
@@ -189,36 +189,30 @@ func networkState() map[string]api.InstanceStateNetwork {
...
 			networkAddress := api.InstanceStateNetworkAddress{
-				Address: addressFields[0],
-				Netmask: addressFields[1],
+				Address: address,
+				Netmask: netmask,
 			}
 
 			scope := "global"
-			if strings.HasPrefix(addressFields[0], "127") {
+			if strings.HasPrefix(address, "127.") {
```
kadinsayani, line 203:
```suggestion
			if strings.HasPrefix(address, "127.") || address == "::1" {
```

lxd-agent/state.go
```diff
@@ -189,36 +189,30 @@ func networkState() map[string]api.InstanceStateNetwork {
...
 				scope = "local"
-			}
-
-			if addressFields[0] == "::1" {
+			} else if address == "::1" {
 				scope = "local"
-			}
-
-			if strings.HasPrefix(addressFields[0], "169.254") {
+			} else if strings.HasPrefix(address, "169.254") {
```
kadinsayani, line 207:
```suggestion
			} else if strings.HasPrefix(address, "169.254") || strings.HasPrefix(address, "fe80:" {
```

lxd-agent/state.go
```diff
@@ -189,36 +189,30 @@ func networkState() map[string]api.InstanceStateNetwork {
...
 				scope = "link"
 			}
 
 			networkAddress.Scope = scope
 
-			if strings.Contains(addressFields[0], ":") {
+			if strings.Contains(address, ":") {
 				networkAddress.Family = "inet6"
 			} else {
 				networkAddress.Family = "inet"
```
kadinsayani, line 218:
```suggestion
networkAddress.Family = "inet"
if strings.Contains(address, ":") {
	networkAddress.Family = "inet6"
} else {
```

lxd/storage/drivers/driver_zfs_volumes.go
```diff
@@ -376,7 +376,7 @@ func (d *zfs) CreateVolumeFromBackup(vol VolumeCopy, srcBackup backup.Info, srcD
 	unpackVolume := func(v Volume, r io.ReadSeeker, unpacker []string, srcFile string, target string) error {
 		d.Logger().Debug("Unpacking optimized volume", logger.Ctx{"source": srcFile, "target": target})
 
-		targetPath := shared.VarPath("") + "/storage-pools/" + target
+		targetPath := shared.VarPath("/storage-pools/", target)
```
kadinsayani, line 379: Do we need to include the `/` before and after `storage-pools`?

https://github.com/canonical/lxd/blob/main/shared/util.go#L227-L238

lxd-agent/state.go
simondeziel: I took a slightly different approach to avoid having `networkAddress.Family` init'ed with `""` then be replaced by either `inet` or `inet6`. I did the same with `networkAddress.Scope`.

kadinsayani (approved):
LGTM!

tomponline (approved):

Issue #14791, opened 2025-01-13
Cluster members missing logs directory for VMs that are Running / qemu.console missing


smcavoy-workwork:
# Required information

 * Distribution: Ubuntu
 * Distribution version: 22.04
 * The output of "snap list --all lxd core20 core22 core24 snapd":
 ```
Name    Version      Rev    Tracking       Publisher   Notes
core22  20241119     1722   latest/stable  canonical✓  base
lxd     6.2-bde4d03  31820  6/stable       canonical✓  -
snapd   2.66.1       23258  latest/stable  canonical✓  snapd,disabled
snapd   2.67         23545  latest/stable  canonical✓  snapd
```
 * The output of "lxc info" or if that fails:
 ```
smcavoy@server7:~$ lxc info
config:
  cluster.https_address: 11.12.12.103:8443
  core.https_address: 11.12.12.103:8443
  storage.images_volume: default/images
api_extensions:
- storage_zfs_remove_snapshots
- container_host_shutdown_timeout
- container_stop_priority
- container_syscall_filtering
- auth_pki
- container_last_used_at
- etag
- patch
- usb_devices
- https_allowed_credentials
- image_compression_algorithm
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
- unix_device_rename
- storage_lvm_use_thinpool
- storage_rsync_bwlimit
- network_vxlan_interface
- storage_btrfs_mount_options
- entity_description
- image_force_refresh
- storage_lvm_lv_resizing
- id_map_base
- file_symlinks
- container_push_target
- network_vlan_physical
- storage_images_delete
- container_edit_metadata
- container_snapshot_stateful_migration
- storage_driver_ceph
- storage_ceph_user_name
- resource_limits
- storage_volatile_initial_source
- storage_ceph_force_osd_reuse
- storage_block_filesystem_btrfs
- resources
- kernel_limits
- storage_api_volume_rename
- network_sriov
- console
[...]

PR #15445, opened 2025-04-24
VM: Use fresh state in `expireLogsTask`

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/14791.

This pull request refactors the `expireLogsTask` functions to use the `Daemon` struct directly instead of the `state.State` object to avoid using a long lived state.

### DIFF
--- a/lxd/logging.go
+++ b/lxd/logging.go
@@ -19,7 +19,7 @@ import (
 
 // This task function expires logs when executed. It's started by the Daemon
 // and will run once every 24h.
-func expireLogsTask(state *state.State) (task.Func, task.Schedule) {
+func expireLogsTask(d *Daemon) (task.Func, task.Schedule) {
 	f := func(ctx context.Context) {
 		opRun := func(op *operations.Operation) error {
 			return expireLogs(ctx, state)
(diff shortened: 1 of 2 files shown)


simondeziel (commented):
The code LGTM and the rational seems sensible but I honestly am not qualified to properly evaluate this PR :/

lxd/logging.go
```diff
@@ -19,8 +19,10 @@ import (
 
 // This task function expires logs when executed. It's started by the Daemon
 // and will run once every 24h.
-func expireLogsTask(state *state.State) (task.Func, task.Schedule) {
+func expireLogsTask(d *Daemon) (task.Func, task.Schedule) {
```
tomponline, line 22: Another potentially nice approach here to avoid needing to lass daemon as an arg here is to pass d.State (the fu ction itselfl as an argument.

kadinsayani:
> Ta, please can you check other tasks for the same issue.

Np! I did take a look and the other tasks follow the same pattern as this PR 👍

lxd/logging.go
kadinsayani: That's a great suggestion! I see that pattern in other areas of the code but the "*Task" functions all accept `*Daemon`, hence why I went with that approach.

kadinsayani: Shall I make it so all "*Task" functions take accept a `stateFunc func() *state.State`?

tomponline: Yes please

kadinsayani: https://github.com/canonical/lxd/pull/15446

PR #15450, opened 2025-04-25
doc: Move contributing out of "Getting Started"

MggMuggins:
Contributing guidelines should be a little easier to find in the docs.

Also, I noticed that the diff of `CONTRIBUTING.md` and `doc/contributing.md` is quite substantial; should these documents be maintained as one document instead?

### DIFF
--- a/doc/getting_started.md
+++ b/doc/getting_started.md
@@ -31,14 +31,13 @@ How to enable access to the UI and the documentation:
 ```
 
 ```{only} diataxis
-How to get support and contribute:
+How to get support:
 ```
 
 ```{filtered-toctree}
 :maxdepth: 1
 
 :diataxis:Get support </support>
-:diataxis:Contribute to LXD </contributing>
 ```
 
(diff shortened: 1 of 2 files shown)


minaelee:
> Contributing guidelines should be a little easier to find in the docs.
> 
> Also, I noticed that the diff of `CONTRIBUTING.md` and `doc/contributing.md` is quite substantial; should these documents be maintained as one document instead?

The way it's set up is that most of CONTRIBUTING.md is pulled into doc/contributing.md with an `include` directive.

doc/getting_started.md
```diff
@@ -38,7 +38,6 @@ How to get support and contribute:
 :maxdepth: 1
 
 :diataxis:Get support </support>
-:diataxis:Contribute to LXD </contributing>
 ```
```
minaelee, line 41: Please also remove "and contribute" from line 34

minaelee (changes requested):

PR #15455, opened 2025-04-28
lxc/config: Allow overriding connection arguments

markylaing:
This is part 1 or breaking down #15311 into smaller chunks


tomponline (approved):
Thanks. I like these smaller PRs.

Ive got a feeling this would be useful for the `lxd waitready` command too.

PR #15458, opened 2025-04-28
lxc/completions: Generalise completions for top level API resources

markylaing:
This is part 3 of splitting up #15311 and was the main reason to start the refactoring due to the repetition of existing functions, and since #15298 needs completions for resource names without also completing names of remotes.

Depends on #15455


markylaing:
#15455 is now merged so I am unmarking this as draft and pinging for review @tomponline @kadinsayani @simondeziel

tomponline (approved):

PR #15467, opened 2025-04-28
lxc/completion: Use filters when completing remotes

markylaing:
We currently have a boolean to indicate if `cmpRemotes` should return all remotes, or just those that are not image servers (minus the default). This isn't specific enough when completing for commands such as `lxc remote set-url`, which cannot be performed on static remotes.

This is part 4 of splitting up #15311 and depends on #15458 (marking as draft until that is merged).


markylaing:
@tomponline this is ready for review now that #15458 is merged. Thanks :)

tomponline (approved):

PR #15468, opened 2025-04-28
lxd: No need to call /1.0 repeatedly in waitready

tomponline:
We are calling `/internal/ready` in a loop already, so no need to call `/1.0` in a loop too.


markylaing (approved):
LGTM :rocket:

PR #15470, opened 2025-04-28
lxc: Config option completion

markylaing:
Updates completions for config options to suggest only `core.` rather than including all sub-options. This reduces the number of completions shown to the user for readability, while also allowing discoverability.

The driver behind this change was the testing added in #15311 (which will be re-added in a later PR) from which I found that `lxc config set <TAB>` did not show any server completions unless I specified the remote (e.g. `local:`). Similarly, I noticed lots of incorrect config keys being suggested when running `lxc profile device add ...` which could only be fixed by parsing the subtype of the device from the metadata API.

This is part 5 of splitting up #15311 and depends on #15467

### DIFF
--- a/lxc/completion.go
+++ b/lxc/completion.go
@@ -180,6 +180,43 @@ func (g *cmdGlobal) cmpTopLevelResourceInRemote(remote string, entityType string
 	return completionsFor(names, "", toComplete), cobra.ShellCompDirectiveNoFileComp
 }
 
+// configOptionAppender returns two functions, the first appends config options to a slice, the second returns the current
+// slice value. Config options are trimmed to the last '.' to improve discoverability. If there is no trailing '.', the
+// given suffix will be appended. This allows lists like: ["gid=", "id=", "mdev=", "mig."] when setting configuration.
+func configOptionAppender(toComplete string, suffix string, size int) (func(option string), func() []string) {
+	var out []string
+	if size > 0 {
+		out = make([]string, 0, size)
+	}
+
+	return func(option string) {
+			trimmed, found := strings.CutPrefix(option, toComplete)
+			if !found {
+				return // Does not have prefix
+			}
+
+			var key string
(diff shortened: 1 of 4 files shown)


kadinsayani (dismissed):

lxc/completion.go
```diff
@@ -384,138 +384,88 @@ func (g *cmdGlobal) cmpImages(toComplete string, instanceServerOnly bool) ([]str
...
-	if err != nil || len(resources) == 0 {
-		return nil, cobra.ShellCompDirectiveError
-	}
-
-	resource := resources[0]
-	client := resource.server
-
-	instance, _, err := client.GetInstance(instanceName)
+// getInstanceType gets the instance type by the "lightest" means, by not requiring the server to recurse.
+func (g *cmdGlobal) getInstanceType(remote string, instanceName string) api.InstanceType {
```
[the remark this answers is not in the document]
markylaing: PS the reason for trying to make things faster here is to make completions as snappy as possible.

lxc/completion.go
```diff
@@ -612,10 +609,26 @@ func (g *cmdGlobal) cmpInstanceAllDevices(instanceName string) ([]string, cobra.
...
+		}
+
+		parts := strings.Split(key, "-")
+		deviceType := parts[1]
+
+		// "unix" is not a device, get the next part.
+		if deviceType == "unix" && len(parts) > 2 {
+			// The metadata API has "unix-usb", but the device type is just "usb"
+			if parts[2] == "usb" {
+				deviceType = "usb"
```
markylaing, line 623: Also a comment from #15311 that this should be fixed in the metadata API. I'm happy to address this in a follow up as I'm not sure how it will affect the documentation.

tomponline: yes please - this feels like a kludge that im not particular happy about.

tomponline (approved):
@markylaing just to confirm, the completion logic doesn't use the "condition" field of the metadata API right?

PR #15473, opened 2025-04-28
lxc: Refactor snapshot completions

markylaing:
This refactors snapshot completions to use the new established patterns and top-level resource comparisons introduced in #15458. Completions for snapshots are factored out into another method which accepts a remote name, an instance name, and a partial snapshot name.


kadinsayani (approved):

tomponline (approved):
Thanks!

PR #15476, opened 2025-04-28
lxc: Refactor file completion

markylaing:
While adding tests to #15311 I found that we were missing completions for many top-level commands. `lxc import` and `lxc export` were two of these. I figured we can have a quick win by refactoring the existing `cmpFiles` function, which wasn't just comparing files, but also comparing instances. This PR makes the function usage clearer and adds the missing completions.

Additionally, I've removed a completion function that was defined twice for `lxc file delete` and also prevented file completions from happening by default when no other completion function has been set.

### DIFF
--- a/lxc/completion.go
+++ b/lxc/completion.go
@@ -1556,6 +1556,69 @@ func isSymlinkToDir(path string, d fs.DirEntry) bool {
 	return true
 }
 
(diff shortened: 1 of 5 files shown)


kadinsayani (approved):

tomponline (approved):
Lgtm!

PR #15485, opened 2025-04-29
lxd: Make d.gateway.Cluster available even if OpenCluster returns with PreconditionFailed

tomponline:
So its available for use with gateway heartbeats.

Also several other LXD start up and shutdown improvements.


mihalicyn (approved):
LGTM!

Issue #14538, opened 2024-11-26
Empty expires_at on snapshots, tiny API inconsistency


edlerd:
# Required information

 * Distribution: Snap
 * Distribution version: 6.1

# Issue description

A tiny inconsistency with responses on the snapshot API exists.

* Instance snapshots without an expiration date:
`GET 1.0/instances/:name?project=default&recursion=2` returns `expires_at: "0001-01-01T00:00:00Z".`
* Volume snapshots without an expiration date:
`GET 1.0/storage-pools/:pool/volumes/custom/:name/snapshots?project=default&recursion=2` returns `expires_at: null`.

I think both instance and volume snapshots APIs should return the same format for an empty expiration date.

roosterfish:
As `0001-01-01T00:00:00Z` seems to be used consistently everywhere else I have decided to get rid of the `null` output and instead return `0001-01-01T00:00:00Z` for custom storage volume snapshots.

PR #15486, opened 2025-04-29
Storage: Fix snapshot expiry

roosterfish:
Whilst working on adding custom storage volumes to the metadata, I found that the wrong timezone is used when computing the `expires_at` property for both instances and custom storage volumes.

The volume's `created_at` property is created using UTC based time but the expiry isn't.

Also fixes https://github.com/canonical/lxd/issues/14538

### DIFF
--- a/test/suites/snapshots.sh
+++ b/test/suites/snapshots.sh
@@ -381,6 +381,9 @@ test_snap_expiry() {
   lxc config show c1/snap0 | grep -q 'expires_at: 0001-01-01T00:00:00Z'
   [ "$(lxc config get --property c1/snap0 expires_at)" = "0001-01-01 00:00:00 +0000 UTC" ]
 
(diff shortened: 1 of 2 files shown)


tomponline:
@roosterfish whilst ur looking at this, is it worth looking at https://github.com/canonical/lxd/issues/14538 too?

roosterfish:
I was setting `snapshots.expiry=1m` with the intention it sets the expiry to one minute. Later I learned that it's 1 month instead but still the wrong hours catched my eye.

tomponline (dismissed):
Good catch!

roosterfish:
> @roosterfish whilst ur looking at this, is it worth looking at https://github.com/canonical/lxd/issues/14538 too?

Yep, will add a new commit and test shortly. Seems to be only a condition that requires removal.

tomponline (approved):
Great thanks!

PR #15487, opened 2025-04-29
Storage: Skip GPT alt‐header move on non-GPT disks (handle `sgdisk` exits 2 & 3)

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/15411.

When moving the GPT alternative header, `sgdisk` will exit with code 2 if no valid GPT header is found (read error) or code 3 if it detects an MBR disk without `-g` (write skipped). Neither condition represents a failure in our context, there is simply no GPT backup header to move. Treat both codes as non-fatal so raw/MBR images continue without error, while still propagating any other `sgdisk` failures.


tomponline (approved):
Thanks!

Issue #14810, opened 2025-01-17
Unsafe network parent in use check


edlerd:
# Required information

 * Distribution: snap
 * Distribution version: lxd     git-4010d6d  32032  latest/edge

# Issue description

On cluster networking: I can sometimes use the same physical interface from a cluster member for two physical managed networks. In other times I get an error: `Parent interface in use by another network`. I think the error should always be displayed.

Suspecting the [parent in use](https://github.com/canonical/lxd/blob/4010d6d732a1d4243aca0a7e626f5731c58c9c08/lxd/network/driver_physical.go#L225-L260) function to be insufficient in the cluster case. It should likely be loading the network configuration for each cluster member and ensure no parent interface collisions per member.

# Steps to reproduce

 1. LXD Cluster of 3 members: m1, m2 and m3. m1 and m2 have two interfaces i1 and i2, m3 has one interface (m3 having only one interface might not be relevant)
 2. Create a physical network n1 using interface i1 on all members
 3. Create another physical network n2 using interface i2 on member m1, and interface i1 on m2 and m3. Issue this command against m1.
 4. No error is displayed, even though we are already using i1 on n1 for the members m2 and m3.

edlerd:
An open, related question is: Is there a situation, where the same physical non-managed interface is used as a parent for more than one physical managed network? Or should the interfaces be used for a maximum of one time?

Motivation for the question is that in the LXD-UI. We could disable any already used interface in the parent selector if an interface may be used only once as a parent.

tomponline:
> Is there a situation, where the same physical non-managed interface is used as a parent for more than one physical managed network? Or should the interfaces be used for a maximum of one time?

I think:

> the interfaces be used for a maximum of one time

PR #15489, opened 2025-04-29
Network: Improve checking parent interface in use for physical networks

nmezhenskyi:
Fixes https://github.com/canonical/lxd/issues/14810.

### DIFF
--- a/lxd/db/networks.go
+++ b/lxd/db/networks.go
@@ -41,6 +41,59 @@ func (c *ClusterTx) GetNetworksLocalConfig(ctx context.Context) (map[string]map[
 	return networks, nil
 }
 
+// GetNetworksNodeParent returns a map associating each node ID in a cluster to networks and their
+// node-specific parent value. If network has no parent, it is omitted.
+func (c *ClusterTx) GetNetworksNodeParent(ctx context.Context) (map[int64]map[string]string, error) {
+	query := `
+   SELECT networks_config.node_id, networks.name, networks_config.value
+   FROM networks_config
+   JOIN networks ON networks.id=networks_config.network_id
+   WHERE (
+      networks_config.key="parent" AND
+      networks_config.node_id IS NOT NULL
+   )`
+
+	rows, err := c.tx.QueryContext(ctx, query)
+	if err != nil {
+		return nil, err
+	}
+
+	defer func() { _ = rows.Close() }()
+
+	nodesNetworksParent := make(map[int64]map[string]string)
+
+	for rows.Next() {
+		var (
+			nodeID      int64
+			networkName string
+			value       string
+		)
+
+		err = rows.Scan(&nodeID, &networkName, &value)
+		if err != nil {
+			return nil, err
+		}
+
+		if nodeID == 0 || networkName == "" || value == "" {
+			continue
+		}
+
+		_, nodeInMap := nodesNetworksParent[nodeID]
+		if !nodeInMap {
+			nodesNetworksParent[nodeID] = map[string]string{}
+		}
+
+		nodesNetworksParent[nodeID][networkName] = value
+	}
+
+	err = rows.Err()
+	if err != nil {
+		return nil, err
+	}
+
+	return nodesNetworksParent, nil
+}
+
 // GetNonPendingNetworkIDs returns a map associating each network name to its ID.
 //
 // Pending networks are skipped.
--- a/lxd/network/driver_physical.go
+++ b/lxd/network/driver_physical.go
@@ -231,7 +231,8 @@ func (n *physical) Validate(config map[string]string) error {
 }
 
 // checkParentUse checks if parent is already in use by another network or instance device.
-func (n *physical) checkParentUse(ourConfig map[string]string) (bool, error) {
+// Returns an error if parent is already in use or the check has failed.
+func (n *physical) checkParentUse(ourConfig map[string]string) error {
 	var err error
 	var projectNetworks map[string]map[int64]api.Network // All managed networks across all projects.
 	var nodesNetworksParent map[int64]map[string]string  // Node IDs mapped to networks and their node-specific parent.
@@ -255,7 +256,7 @@ func (n *physical) checkParentUse(ourConfig map[string]string) (bool, error) {
 		return nil
 	})
 	if err != nil {
-		return false, err
+		return err
 	}
 
 	for projectName, networks := range projectNetworks {
@@ -273,7 +274,7 @@ func (n *physical) checkParentUse(ourConfig map[string]string) (bool, error) {
 				// If either network doesn't specify a vlan, or both specify same vlan,
 				// then we can't use this parent.
 				if (network.Config["vlan"] == "" || ourConfig["vlan"] == "") || network.Config["vlan"] == ourConfig["vlan"] {
-					return true, nil
+					return fmt.Errorf("Parent interface %q in use by another network", ourConfig["parent"])
 				}
 			}
 		}
@@ -292,14 +293,14 @@ func (n *physical) checkParentUse(ourConfig map[string]string) (bool, error) {
 		for _, parent := range networksParent {
 			_, alreadyInUse := parentsInUse[parent]
 			if alreadyInUse {
-				return true, nil
+				return fmt.Errorf("Parent interface %q in use by another network on cluster member %q", parent, n.nodes[nodeID].Name)
 			}
 
 			parentsInUse[parent] = struct{}{}
 		}
 	}
 
-	return false, nil
+	return nil
 }
 
 // Create checks whether the referenced parent interface is used by other networks or instance devices, as we
@@ -309,13 +310,9 @@ func (n *physical) Create(clientType request.ClientType) error {
 
 	// We only need to check in the database once, not on every clustered node.
 	if clientType == request.ClientTypeNormal {
-		inUse, err := n.checkParentUse(n.config)
+		err := n.checkParentUse(n.config)
 		if err != nil {
 			return err
-		}
-
-		if inUse {
-			return fmt.Errorf("Parent interface %q in use by another network", n.config["parent"])
 		}
 	}
 
@@ -500,13 +497,9 @@ func (n *physical) Update(newNetwork api.NetworkPut, targetNode string, clientTy
 				return errors.New("Cannot update network parent interface when in use")
 			}
 
-			inUse, err := n.checkParentUse(newNetwork.Config)
+			err = n.checkParentUse(newNetwork.Config)
 			if err != nil {
 				return err
-			}
-
-			if inUse {
-				return fmt.Errorf("Parent interface %q in use by another network", newNetwork.Config["parent"])
 			}
 		}
 	}
--- a/lxd/network/driver_physical.go
+++ b/lxd/network/driver_physical.go
@@ -232,16 +232,30 @@ func (n *physical) Validate(config map[string]string) error {
 
 // checkParentUse checks if parent is already in use by another network or instance device.
 func (n *physical) checkParentUse(ourConfig map[string]string) (bool, error) {
-	// Get all managed networks across all projects.
 	var err error
-	var projectNetworks map[string]map[int64]api.Network
+	var projectNetworks map[string]map[int64]api.Network // All managed networks across all projects.
+	var nodesNetworksParent map[int64]map[string]string  // Node IDs mapped to networks and their node-specific parent.
 
 	err = n.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+		// Get all managed networks across all projects.
+		// If in clustered mode, returned network configs are for the current node.
 		projectNetworks, err = tx.GetCreatedNetworks(ctx)
-		return err
+		if err != nil {
+			return fmt.Errorf("Failed to load all networks: %w", err)
+		}
+
+		// Get parent interfaces for all networks on all nodes.
+		if n.state.ServerClustered && len(n.nodes) > 1 {
+			nodesNetworksParent, err = tx.GetNetworksNodeParent(ctx)
+			if err != nil {
+				return fmt.Errorf("Failed to load node-specific configs: %w", err)
+			}
+		}
+
+		return nil
 	})
 	if err != nil {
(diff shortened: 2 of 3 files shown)


lxd/network/driver_physical.go
```diff
@@ -258,13 +259,51 @@ func (n *physical) checkParentUse(ourConfig map[string]string) (bool, error) {
...
 		}
 	}
 
-	return false, nil
+	// Check that parent interfaces on other nodes are not already in use.
+	if n.state.ServerClustered && len(n.nodes) > 1 {
+		// Node names mapped to networks.
+		var nodesNetworksParent = make(map[string]map[string]string, len(n.nodes)-1)
+
+		err = n.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
```
tomponline, line 273: In dqlite all transactions and queries are sent to the leader node (even SELECTs) so we need to avoid starting unnecessary transactions and running many smaller queries, because the RTT is incurred for each.

Also by getting all of the data we need in a single transaction we can ensure we get a consistent point in time snapshot of the data, rather than potentially it being modified between transactions.

So here, I think we need to change this a bit:

1. Perform all queries in the earlier transaction on line 239.
2. Modify GetNetworksNodeParent to accept a variadic argument of Network IDs so that it can retrieve the per-node parents for multiple networks in a single query (and alter the return map to reflect that).
3. `n.state.DB.Cluster.GetNodeID()` doesn't need to be called on every iteration of the for loop, you can store it in a var outside of the for loop.

lxd/network/driver_physical.go
```diff
@@ -258,13 +259,51 @@ func (n *physical) checkParentUse(ourConfig map[string]string) (bool, error) {
 				// If either network doesn't specify a vlan, or both specify same vlan,
 				// then we can't use this parent.
 				if (network.Config["vlan"] == "" || ourConfig["vlan"] == "") || network.Config["vlan"] == ourConfig["vlan"] {
-					return true, nil
+					return fmt.Errorf("Parent interface %q in use by another network", n.config["parent"])
```
tomponline, line 262: Can this change be moved into a separate commit please?

tomponline (changes requested):
LGTM, just some improvements needed in the querying approach.

Also, we have `test_clustering_network` test function. I wonder if it is feasible to extend this to also test clustering physical network creation by using several `dummy` interfaces as the parent?

lxd/network/driver_physical.go
```diff
@@ -258,13 +287,29 @@ func (n *physical) checkParentUse(ourConfig map[string]string) (bool, error) {
...
+
+	// Check that parent interfaces on other nodes are not already in use.
+	if n.state.ServerClustered {
+		for nodeID, networksParent := range nodesNetworksParent {
+			parentsInUse := make(map[string]struct{})
+
+			for _, parent := range networksParent {
+				_, alreadyInUse := parentsInUse[parent]
+				if alreadyInUse {
+					return fmt.Errorf("Parent interface %q in use by another network on node %q", parent, n.nodes[nodeID].Name)
```
nmezhenskyi, line 304: This should be safe because `n.nodes` has type `map[int64]db.NetworkNode`.

nmezhenskyi:
I've made the changes to the querying approach and added a test for clustering physical network creation to the `test_clustering_network`.

lxd/db/networks.go
```diff
@@ -41,6 +41,38 @@ func (c *ClusterTx) GetNetworksLocalConfig(ctx context.Context) (map[string]map[
...
+				return nil, err
+			}
+
+			if config["parent"] != "" {
+				networks[name] = config["parent"]
+			}
+		}
+
+		nodesNetworksParent[nodeID] = networks
+	}
```
nmezhenskyi, line 71: I just realized that I could also change this loop to a single SQL query like: 
`WHERE ... networks_config.node_id IN (<node id list>) ...`.

lxd/db/networks.go
```diff
@@ -41,6 +42,74 @@ func (c *ClusterTx) GetNetworksLocalConfig(ctx context.Context) (map[string]map[
 	return networks, nil
 }
 
+// GetNetworksNodeParent returns a map associating each given node ID to networks and their
+// node-specific parent value. If network has no parent, it is omitted.
+func (c *ClusterTx) GetNetworksNodeParent(ctx context.Context, nodeID int64, nodeIDs ...int64) (map[int64]map[string]string, error) {
```
tomponline, line 47:
```go
nodeID int64, nodeIDs ...int64
```

is super confusing argument naming.

And the function description doesn't help much to illuminate as to why there are 2 different arguments, and then they get appended into the same list anyway.

If this is used to exclude our current node, then i think that would be better done outside of this function, as there maybe cases where we can use this function where we dont want to exclude ourselves.

lxd/db/networks.go
```diff
@@ -41,6 +42,74 @@ func (c *ClusterTx) GetNetworksLocalConfig(ctx context.Context) (map[string]map[
...
+	}
+
+	query := `
+   SELECT networks_config.node_id, networks.name, networks_config.value
+   FROM networks_config
+   JOIN networks ON networks.id=networks_config.network_id
+   WHERE (
+      networks_config.key="parent" AND
+      networks_config.node_id IN (` + nodeIDsList.String() +
+		"))\nORDER BY networks_config.node_id ASC"
```
tomponline, line 70: can we avoid the \n and just put it on a separate line like you did with the other lines?

lxd/db/networks.go
```diff
@@ -41,6 +42,74 @@ func (c *ClusterTx) GetNetworksLocalConfig(ctx context.Context) (map[string]map[
...
+		nodeIDsList.WriteString(strconv.FormatInt(id, 10))
+	}
+
+	query := `
+   SELECT networks_config.node_id, networks.name, networks_config.value
+   FROM networks_config
+   JOIN networks ON networks.id=networks_config.network_id
+   WHERE (
+      networks_config.key="parent" AND
+      networks_config.node_id IN (` + nodeIDsList.String() +
```
tomponline, line 69: whilst building the query like this is fine in this case (because the input can only be integers), I'd like to see us use query params and `query.Params()` function as a habit, unless theres very specific reasons not to (like in `InstanceList()`).

lxd/db/networks.go
```diff
@@ -41,6 +42,74 @@ func (c *ClusterTx) GetNetworksLocalConfig(ctx context.Context) (map[string]map[
...
+		return nil, err
+	}
+
+	defer func() { _ = rows.Close() }()
+
+	nodesNetworksParent := make(map[int64]map[string]string, len(nodeIDs))
+
+	for rows.Next() {
+		var (
+			nodeID      *int64
```
tomponline, line 83: Should we use `sql.NullInt64` here instead of the pointer, to clarify why there is that difference (because its representing the schema presumably). 

https://pkg.go.dev/database/sql#NullInt64

tomponline: or you could bake the filter on line 93 into the query itself to avoid having to deal with it in Go potentially?

lxd/network/driver_physical.go
```diff
@@ -232,13 +232,41 @@ func (n *physical) Validate(config map[string]string) error {
...
+		// If in cluster mode, returned network configs are for the current node.
 		projectNetworks, err = tx.GetCreatedNetworks(ctx)
-		return err
+		if err != nil {
+			return err
+		}
+
+		// Get parent interfaces for all networks on other nodes.
+		if n.state.ServerClustered && len(otherNodeIDs) > 0 {
+			nodesNetworksParent, err = tx.GetNetworksNodeParent(ctx, otherNodeIDs[0], otherNodeIDs[1:]...)
```
tomponline, line 264: does this include our own network at this stage?

lxd/db/networks.go
nmezhenskyi: The idea here is to require the caller to provide at least 1 node ID to fetch the information for, and optionally allow providing a list of additional IDs. At first, I tried using a single argument `nodeIDs ...int64` to avoid the confusion of having two separate arguments, but that required an additional runtime check to ensure that the caller passed at least 1 node ID. Then, the question is also how to handle the situation where the caller does not pass any IDs: to return an error or to panic since that is unintended use of function.   

I thought it would be better to create a separate argument for the required ID because the compiler would do the check and force the caller to provide an ID. I do agree that the naming is confusing though. Maybe we could call the arguments something like `firstNodeID int64, restNodeIDs ...int64` or `requiredNodeID int64, additionalNodeIDs ...int64`.  

What do you think? Or should we use a single argument to pass IDs? @tomponline

tomponline: why do we need to filter by specific node IDs, can we just have a function that returns for all of them in the cluster?

lxd/db/networks.go
nmezhenskyi: Yes, I used the pointer here to account for NULL value in the database. I'll try using SQL filter instead.

lxd/db/networks.go
nmezhenskyi: Right, we can. I guess my instinct was to filter by specific node IDs to skip fetching networks for the current node, since we don't need to check it a second time, but maybe it's better to filter it outside of the query.

tomponline: yes lets do that, keep it simple/boring

lxd/network/driver_physical.go
nmezhenskyi: Yes, that includes the current network as well.

lxd/db/networks.go
nmezhenskyi: Sounds good, I'll change it.

tomponline:
@nmezhenskyi hows this coming along? Do you think we can get in for LXD 6.4?

nmezhenskyi:
@tomponline Yes, I'm finishing it up, should be ready soon.

lxd/network/driver_physical.go
```diff
@@ -265,6 +279,28 @@ func (n *physical) checkParentUse(ourConfig map[string]string) (bool, error) {
...
+
+			parentsInUse := make(map[string]struct{})
+
+			for _, parent := range networksParent {
+				_, alreadyInUse := parentsInUse[parent]
+				if alreadyInUse {
+					return true, nil
+				}
+
+				parentsInUse[parent] = struct{}{}
```
tomponline, line 300: should there not be a node dimension to this variable, otherwise what happens if 2 different nodes use the same parent name for a single network, thats valid, but wont that trigger an in use error here?

lxd/network/driver_physical.go
```diff
@@ -265,6 +279,28 @@ func (n *physical) checkParentUse(ourConfig map[string]string) (bool, error) {
 		}
 	}
 
+	// Check that parent interfaces on other nodes are not already in use.
+	if n.state.ServerClustered && len(n.nodes) > 1 {
```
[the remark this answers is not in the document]
nmezhenskyi: True, I'll remove it.

lxd/network/driver_physical.go
nmezhenskyi: `parentsInUse` variable is isolated to a single node. The variable is recreated on each iteration for each consecutive node. So it only tracks parents for one node at a time and prevents reuse of the interface if it's already been used on that node. It is possible to have 2 different nodes use the same parent name for a single network.

nmezhenskyi: > what happens if 2 different nodes use the same parent name for a single network, thats valid, but wont that trigger an in use error here?

If I understood this part correctly, I added a test for this behavior:

```sh
  echo "Create a physical network net1."
  LXD_DIR="${LXD_ONE_DIR}" lxc network create "${net1}" --type=physical parent=i1 --target=node1
  LXD_DIR="${LXD_ONE_DIR}" lxc network create "${net1}" --type=physical parent=i1 --target=node2
  LXD_DIR="${LXD_ONE_DIR}" lxc network create "${net1}" --type=physical
  LXD_DIR="${LXD_ONE_DIR}" lxc network show "${net1}" | grep status: | grep -q Created
```

This test uses the same interface name `i1` as a parent on two different nodes to create a physical network.

tomponline: Excellent thanks

lxd/network/driver_physical.go
```diff
@@ -292,14 +293,14 @@ func (n *physical) checkParentUse(ourConfig map[string]string) (bool, error) {
 		for _, parent := range networksParent {
 			_, alreadyInUse := parentsInUse[parent]
 			if alreadyInUse {
-				return true, nil
+				return fmt.Errorf("Parent interface %q in use by another network on node %q", parent, n.nodes[nodeID].Name)
```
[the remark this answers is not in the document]
nmezhenskyi: Got it, I'll change it

test/suites/clustering.sh
```diff
@@ -1267,6 +1267,52 @@ test_clustering_network() {
...
+  LXD_DIR="${LXD_ONE_DIR}" lxc network create "${net2}" --type=physical parent=i2 --target=node2
+  ! LXD_DIR="${LXD_ONE_DIR}" lxc network create "${net2}" --type=physical || false
+  LXD_DIR="${LXD_ONE_DIR}" lxc network show "${net2}" | grep status: | grep -q Errored
+  LXD_DIR="${LXD_ONE_DIR}" lxc network delete "${net2}"
+
+  echo "Check that parent interface i1 on node2 cannot be used for another physical network net2."
+  LXD_DIR="${LXD_ONE_DIR}" lxc network create "${net2}" --type=physical parent=i2 --target=node1
+  LXD_DIR="${LXD_ONE_DIR}" lxc network create "${net2}" --type=physical parent=i1 --target=node2
+  ! LXD_DIR="${LXD_ONE_DIR}" lxc network create "${net2}" --type=physical || false
+  LXD_DIR="${LXD_ONE_DIR}" lxc network show "${net2}" | grep status: | grep -q Errored
```
tomponline, line 1297: @simondeziel will likely have suggestions around this line.

I'll merge for now, but will likely need a follow up with better grep matching here.

tomponline (approved):

tomponline:
Thanks!

test/suites/clustering.sh
simondeziel: https://github.com/canonical/lxd/pull/15637

PR #15495, opened 2025-04-30
lxd: Rework waitready to use contexts and server-side timeout support

tomponline:
- Changes `/internal/wait` to accept a `timeout` parameter rather than a `wait` parameter.
- Updates `lxd waitready` to pass its `--timeout` flag value to `/interna/wait` as `timeout` parameter.
- Updates `lxd waitready` to use contexts and to detect `http.StatusServiceUnavailable` response from server after timeout is reached in order to differentiate between an unreachable server and a not-ready server.
- Simplifies logging logic.

### DIFF
--- a/lxd/main_waitready.go
+++ b/lxd/main_waitready.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


lxd/main_waitready.go
```diff
@@ -38,67 +40,64 @@ func (c *cmdWaitready) Command() *cobra.Command {
...
-	}()
 
-	if c.flagTimeout > 0 {
-		select {
-		case <-finger:
-		case <-time.After(time.Second * time.Duration(c.flagTimeout)):
-			return fmt.Errorf("LXD still not running after %ds timeout (%v)", c.flagTimeout, errLast)
+		log(i, "Checking if LXD daemon is ready (attempt %d)", i)
+
+		u := api.NewURL().Path("internal", "ready").WithQuery("timeout", strconv.Itoa(c.flagTimeout))
```
simondeziel, line 89: For consistency, should `flagTimeout` be treated (and changed to) `uint64`?
```suggestion
		u := api.NewURL().Path("internal", "ready").WithQuery("timeout", strconv.FormatUint(c.flagTimeout, 10))
```

simondeziel (approved):
LGTM with a tiny nit.

lxd/main_waitready.go
tomponline: Yep will change

tomponline: https://github.com/canonical/lxd/pull/15501

PR #15496, opened 2025-04-30
Storage: Improve error message wording in `CreateInstanceFromConversion`

kadinsayani:
`...lower then source disk...`->`...less than source disk...`


tomponline (approved):
Nice

Issue #15417, opened 2025-04-17
Replace `golang.org/x/crypto/hkdf` by `crypto/hkdf` from Go 1.24


simondeziel:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

`lxd/auth/oidc/oidc.go` uses the HKDF package from `golang.org/x/crypto/hkdf` but with Go 1.24, this made it into the stdlib with very similar functions prototypes. See https://pkg.go.dev/crypto/hkdf@go1.24.2

### Describe the solution you'd like

`crypto/hkdf` to replace `golang.org/x/crypto/hkdf`

simondeziel:
> We can do this one now

Unfortunately not at as I had to backpedal on the switch to Go 1.24 due to the many `golangci-lint` errors it turns up.

PR #15497, opened 2025-04-30
Switch to Go 1.24 and update Go modules

simondeziel:
Note to the reviewers, the commits with their message ending with `(perfsprint)` were all automatically generated by committing changes made with the `perfsprint` tool invoked like this:

```
perfsprint -fix -fiximports -bool-format=false -errorf=true -hex-format=false -int-conversion=false -integer-format=false --sprintf1=false -strconcat=false -string-format=false ./...
```

Fixes https://github.com/canonical/lxd/security/dependabot/19
Unblocks https://github.com/canonical/lxd/issues/15417


lxd/db/query/marshal.go
tomponline, line 1: @markylaing is this expected to be unused?

shared/api/url.go
```diff
@@ -45,7 +45,7 @@ func (u *URL) Path(pathParts ...string) *URL {
 	}
 
 	u.URL.Path = path.String()
```
tomponline, line 47: @simondeziel weird that this line didnt trigger the linter either, as it presumably should also be updated.

tomponline: ah its because of the u.Path() function.

lxd/db/migration.go
```diff
@@ -271,9 +281,14 @@ func importNodeAssociation(entity string, columns []string, row []any, tx *sql.T
 	stmt := fmt.Sprintf(
 		"INSERT INTO %ss_nodes(%s_id, node_id) VALUES(?, 1)", entity, entity)
 	var id int64
+	var ok bool
 	for i, column := range columns {
 		if column == "id" {
-			id = row[i].(int64)
+			id, ok = row[i].(int64)
+			if !ok {
+				return fmt.Errorf("Failed to convert %s ID to int64", entity)
```
tomponline, line 289: as entity is a string %q would be nicer to allow seeing where the string value starts and the error message stops

.golangci.yaml
```diff
@@ -86,6 +86,7 @@ linters:
       - third_party$
       - builtin$
       - examples$
+      - lxd/idmap/idmapset_linux\.go$
```
tomponline, line 89: this is fine if its short term, but i worry we will forget to look at it again.

simondeziel: I've been avoid touching this file as it's full of linter problems. I intend to eventually get to it but I'll need to carve out some time for it.

PR #15500, opened 2025-04-30
github: Add KEV matching to security alerts

hamistao:
This downloads the KEV json list in `trivy-repo` and caches it to use in `trivy-snap`. Then in both jobs uses it to check which alerts IDs are included in it and if so, tag the alert with `(KEV)`

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -59,6 +59,20 @@ jobs:
           path: /home/runner/vuln-cache
           key: trivy-cache-${{ github.run_id }}
 
+      - name: Tag KEV alerts
+        run: |
+            cd /home/runner
+            curl -s --compressed -o kev.json https://www.cisa.gov/sites/default/files/feeds/known_exploited_vulnerabilities.json
+            kev_ids="$(jq -r '.vulnerabilities[].cveID' kev.json)"
+            jq --arg ids "$kev_ids" '($ids | split("\n")) as $id_list | .runs[].tool.driver.rules[] |= (
+              if (.id as $id | $id_list | index($id)) then
+                .shortDescription.text |= . + " (KEV)"
+              else
+                .
+              end
+            )' "$GITHUB_WORKSPACE/trivy-lxd-repo-scan-results.sarif" > trivy-modified.sarif
+            mv trivy-modified.sarif "$GITHUB_WORKSPACE/trivy-lxd-repo-scan-results.sarif"
+
       - name: Upload Trivy scan results to GitHub Security tab
         uses: github/codeql-action/upload-sarif@28deaeda66b76a05916b6923827895f2b14ab387 # v3.28.16
         with:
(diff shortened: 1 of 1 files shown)


hamistao:
I can't show a screenshot of how the alerts look because aparentely none fit the KEV matching (at least on the main branch), but I added a dummy alert manually to test and it worked fine.

.github/workflows/security.yml
```diff
@@ -110,8 +133,26 @@ jobs:
       - name: Flag snap scanning alerts
         run: |
           cd /home/runner
-          jq '.runs[].tool.driver.rules[] |= (.shortDescription.text |= "Snap scan - " + .)' ${{ matrix.version }}-stable.sarif > tmp.json
-          mv tmp.json ${{ matrix.version }}-stable.sarif
+          cat <<< $(jq '.runs[].tool.driver.rules[] |= (.shortDescription.text |= "Snap scan - " + .)' ${{ matrix.version }}-stable.sarif) > ${{ matrix.version }}-stable.sarif
```
tomponline, line 136: @kadinsayani please can you add a comment as to what this is doing, its not that clear.

.github/workflows/security.yml
```diff
@@ -59,6 +59,29 @@ jobs:
           path: /home/runner/vuln-cache
           key: trivy-cache-${{ github.run_id }}
 
+      - name: Get KEV list
+        run: |
+          cd /home/runner
+          curl -s -o kev.json https://www.cisa.gov/sites/default/files/feeds/known_exploited_vulnerabilities.json
```
simondeziel, line 65:
```suggestion
          curl -s --compressed -o kev.json https://www.cisa.gov/sites/default/files/feeds/known_exploited_vulnerabilities.json
```

Downloads 408K instead of ~3M.

.github/workflows/security.yml
```diff
@@ -59,6 +59,29 @@ jobs:
...
+
+      - name: Cache KEV list
```
simondeziel, line 67: That file is so cheap to download I don't think we even need to concern ourselves with caching it.

.github/workflows/security.yml
kadinsayani: Nice!

.github/workflows/security.yml
```diff
@@ -59,6 +59,19 @@ jobs:
           path: /home/runner/vuln-cache
           key: trivy-cache-${{ github.run_id }}
 
+      - name: Tag KEV alerts
```
simondeziel, line 62: I didn't quite find where this step was altering the `.sarif` file that is fed back into GH.

.github/workflows/security.yml
```diff
@@ -110,8 +123,21 @@ jobs:
       - name: Flag snap scanning alerts
         run: |
           cd /home/runner
-          jq '.runs[].tool.driver.rules[] |= (.shortDescription.text |= "Snap scan - " + .)' ${{ matrix.version }}-stable.sarif > tmp.json
-          mv tmp.json ${{ matrix.version }}-stable.sarif
+          # Modify the SARIF file in-place by prefixing each rule's shortDescription.text with "Snap scan - "
+          cat <<< $(jq '.runs[].tool.driver.rules[] |= (.shortDescription.text |= "Snap scan - " + .)' ${{ matrix.version }}-stable.sarif) > ${{ matrix.version }}-stable.sarif
+
+      - name: Tag KEV alerts
```
simondeziel, line 129: I didn't quite find where this step was altering the `.sarif` file that is fed back into GH.

.github/workflows/security.yml
```diff
@@ -110,8 +123,21 @@ jobs:
       - name: Flag snap scanning alerts
         run: |
           cd /home/runner
-          jq '.runs[].tool.driver.rules[] |= (.shortDescription.text |= "Snap scan - " + .)' ${{ matrix.version }}-stable.sarif > tmp.json
-          mv tmp.json ${{ matrix.version }}-stable.sarif
+          # Modify the SARIF file in-place by prefixing each rule's shortDescription.text with "Snap scan - "
+          cat <<< $(jq '.runs[].tool.driver.rules[] |= (.shortDescription.text |= "Snap scan - " + .)' ${{ matrix.version }}-stable.sarif) > ${{ matrix.version }}-stable.sarif
```
simondeziel, line 127: This pattern of reading and writing to the same file sounds risky to me. What it the file is too big to fit into memory?

Also, since we are editing the `.shortDescription.text` already, could we also add the ` (KEV)` suffix while at it? If that's too complicated/unreadable with a single `jq` pass, we can probably use pipes and do it in 2 steps. This should also ensure the `KEV` bits end up in the uploaded `.sarif`.

simondeziel (changes requested):

.github/workflows/security.yml
kadinsayani: It was missing! I've added it now.

.github/workflows/security.yml
kadinsayani: Agreed, we don't need to modify the file in place. I think combining the steps is a good idea - I've made it so :)

kadinsayani:
Test run on my fork: https://github.com/kadinsayani/lxd/actions/runs/14788067825/job/41520025008.

tomponline (approved):

tomponline:
Thanks!

PR #15501, opened 2025-05-01
lxd: Use uint64 for waitready timeout flag for consistency with server side

tomponline:
https://github.com/canonical/lxd/pull/15495#pullrequestreview-2807703090


markylaing (approved):
LGTM

PR #15502, opened 2025-05-01
Replace `fmt.Errorf()`  by `errors.new()`

simondeziel:
Note to the reviewers, the commits with their message ending with `(perfsprint)` were all automatically generated by committing changes made with the `perfsprint` tool invoked like this:

```
perfsprint -fix -fiximports -bool-format=false -errorf=true -hex-format=false -int-conversion=false -integer-format=false --sprintf1=false -strconcat=false -string-format=false ./...
```

### DIFF
--- a/lxd/storage/drivers/driver_cephfs.go
+++ b/lxd/storage/drivers/driver_cephfs.go
@@ -130,11 +130,11 @@ func (d *cephfs) Create() error {
 
 	// Config validation.
 	if d.config["source"] == "" {
-		return fmt.Errorf("Missing required source name/path")
+		return errors.New("Missing required source name/path")
 	}
 
 	if d.config["cephfs.path"] != "" && d.config["cephfs.path"] != d.config["source"] {
-		return fmt.Errorf("cephfs.path must match the source")
+		return errors.New("cephfs.path must match the source")
 	}
 
 	d.config["cephfs.path"] = d.config["source"]
@@ -317,7 +317,7 @@ func (d *cephfs) Create() error {
 	// Check that the existing path is empty.
 	ok, _ := shared.PathIsEmpty(filepath.Join(mountPoint, fsPath))
 	if !ok {
-		return fmt.Errorf("Only empty CephFS paths can be used as a LXD storage pool")
+		return errors.New("Only empty CephFS paths can be used as a LXD storage pool")
 	}
 
 	revert.Success()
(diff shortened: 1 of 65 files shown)


simondeziel:
All green minus the CodeQL (duplicated/existing/refreshed alerts).

tomponline:
Please rebase

lxd/db/cluster/devices.mapper.go
```diff
@@ -118,7 +119,7 @@ func GetDevices(ctx context.Context, tx *sql.Tx, parent string, filters ...Devic
 		}
 
 		if len(entries) == 0 {
-			return nil, fmt.Errorf("Cannot filter on empty DeviceFilter")
+			return nil, errors.New("Cannot filter on empty DeviceFilter")
```
markylaing, line 122: These files are generated by `make update-schema` - It might be worth updating https://github.com/canonical/lxd/blob/300a2f7af72ff99d0735fa1417dccf7317162d4c/lxd/db/generate/db/method.go#L317 to use `errors.New` and adding `errors` to the list of imports here https://github.com/canonical/lxd/blob/300a2f7af72ff99d0735fa1417dccf7317162d4c/lxd/db/generate/db/constants.go#L6

Then running `make update-schema`, and then running the `perfsprint` tool. Otherwise these will revert when schema generation is run again.

markylaing (changes requested):

lxd/db/cluster/devices.mapper.go
tomponline: good spot!

markylaing: I've just realised that the update I've commented on is a different error to the one I found in `lxd-generate` so presumably there are a few of these to fix first.

markylaing: I only found a few occurrences so just fixed it as I went. Here's the PR https://github.com/canonical/lxd/pull/15507

tomponline:
can be rebased

lxd/db/cluster/devices.mapper.go
simondeziel: Ah thanks for catching and fixing it properly! I'll try and be more careful with future mechanical changes :)

simondeziel:
@tomponline rebase done on top of Mark's PR for the `mapper.go` generated files. That reduced the number of changes a bit.

tomponline (approved):
Ta!

PR #15503, opened 2025-05-01
lxd/db/migration: use `%q` specifier to quote `entity`

simondeziel:


test/suites/snapshots.sh
```diff
@@ -391,8 +391,8 @@ test_snap_expiry() {
   created_at="$(lxc config get c1/snap1 --property created_at | awk -F. '{print $1}')"
   expires_at="$(lxc config get c1/snap1 --property expires_at | awk -F. '{print $1}')"
 
-  # Check if the expires_at propery is exactly 1d ahead.
-  [ "$(date -d "${created_at} today + 1days")" = "$(date -d "${expires_at}")" ]
+  # Check if the expires_at propery is 1d (or 1d+1s) ahead.
```
tomponline, line 394: @simondeziel @roosterfish can we fix this in the LXD code itself? The issue appears to be that time.Now.UTC() is called twice at different points, once early on to derive the expiry, and later in CreateInternal to get the created at timestamp.

Looking at this now:

The `instance.CreateInternal(d.state, args, true)` call is where the creation time is being defaulted to UTC now if not supplied in args:

https://github.com/canonical/lxd/blob/ee31a202f24663904bfd0d2b3c90100cca13ba54/lxd/instance/instance_utils.go#L796-L798

If we change the instance.Snapshot() `expiry` arg to be a *time.Time pointer instead, then we could pass `req.ExpiresAt` directly into `instance.Snapshot()` and then move this logic:

https://github.com/canonical/lxd/blob/main/lxd/instance_snapshot.go#L344-L352

All the way down into `(d *common) snapshotCommon()`

https://github.com/canonical/lxd/blob/main/lxd/instance/drivers/driver_common.go#L694-L711

At which point we can use the same `time.Now().UTC()` value to set the `args.CreationDate` as well as be used to derive the `args.ExpiryDate` value.

This way we would remove the possibility of them ever being slightly offset.

simondeziel: > This way we would remove the possibility of them ever being slightly offset.

Indeed and this would avoid the need for stripping the microseconds:

```
  # Get snapshot created_at and expires_at properties without microseconds.
  created_at="$(lxc config get c1/snap1 --property created_at | awk -F. '{print $1}')"
  expires_at="$(lxc config get c1/snap1 --property expires_at | awk -F. '{print $1}')"
```

simondeziel: @roosterfish I think I'd feel more comfortable leave the refactor Tom suggested to you. If you don't have time it's OK and I'll put it in my queue.

tomponline (approved):

test/suites/snapshots.sh
roosterfish: Fixed in https://github.com/canonical/lxd/pull/15528. Also found another bug whilst fixing it.

simondeziel: Many thanks @roosterfish!

PR #15507, opened 2025-05-02
Stop generating `fmt.Errorf` calls without format directives in `lxd-generate`

markylaing:
Discussed here https://github.com/canonical/lxd/pull/15502#discussion_r2071295337 and quickly put up a PR as I only found a few other cases. Verified that there are no more generated occurrences of this using `grep -r 'fmt\.Errorf[^%]*)' lxd/db/cluster | grep -c 'mapper.go'`.


tomponline (approved):
cheers

PR #15508, opened 2025-05-02
lxd/response/swagger: annotate individual strucs for nolint:unused

simondeziel:
The TICS scanner doesn't seem to recognise the file level `//nolint:unused` directive and report each structs as being unused so try to workaround that limitation.

Note that those structs are actually needed and they end up in the generate REST API doc.


markylaing (approved):
LGTM

Let's hope TICS accepts it!

tomponline (approved):

simondeziel:
It didn't work so I'll try Mark's suggestion instead.

Issue #15300, opened 2025-04-01
Instance network device rename fails, validation error


edlerd:
### Distribution

ubuntu

### Distribution version

24.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
lxd                                   6.3-f22a395                     33085  6/stable         canonical**     -
```

### Issue description

When changing the name for an instance NIC device, a validation error occurs. Expected is that the operation succeeds.

I.e. changing from this:
```
devices:
  eth-1:
    network: my-bridge
    type: nic
```
to this:
```
devices:
  eth-2:
    network: my-bridge
    type: nic
```
causes an error:
```
Failed add validation for device "eth-2": Instance DNS name "precise-ray" conflict between "eth-2" and "eth-1" because both are connected to same network
```
When removing the device, then saving, then adding back with new name, the procedure succeeds. Also, when doing this via a profile device, it succeeds. Only for instances, the validation seems to be broken.

### Steps to reproduce

1. Create an instance
2. Assign a network device to the instance
3. Rename the network device name via the `PUT /1.0/instances/:name` endpoint or via `lxc config edit :name`
4. See error, validation fails, expected is this action to succeed

edlerd:
@Kxiru found this issue in https://github.com/canonical/lxd-ui/pull/1151

edlerd:
There is another problem with the validation:
1. An instance with the name "foo" exists in two projects: "project-1" and "project-2". Instance foo in project-1 is connected to a network.
2. Now editing instance "foo" in project-2, adding a new nic device with the same network.
3. We get a validation error, that the DNS name is already taken. Even though, the instance DNS names should be considered project specific.

tomponline:
@edlerd are you in agreement that this isnt a bug?

edlerd:
> [@edlerd](https://github.com/edlerd) are you in agreement that this isnt a bug?

I think the original issue of renaming the network device failing as an open bug. It reproduces on edge, 5.21 and 6.3.

tomponline:
I'm not entirely following you.

> Starting the instance fails with the dns error as in the issue description above. I think this should be prevented by validation or the dns should be made project aware.

It is prevented with validation - hence the failure of the instance to start - its start time validation.

If you could show full reproducer steps I think it would help demonstrate the issue.

edlerd:
## Reproducer for instance network device rename

```
lxc launch ubuntu:20.04 u1
lxc config edit u1
```
add a network device with an existing network:
```
[...]

PR #15510, opened 2025-05-02
lxd/device: Skip address conflict validation for instance NIC rename

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/15300.

### DIFF
--- a/lxd/device/nic_bridged.go
+++ b/lxd/device/nic_bridged.go
@@ -402,6 +402,12 @@ func (d *nicBridged) checkAddressConflict() error {
 		// that has DNS enabled and is connected to the same untagged VLAN.
 		if d.network != nil && d.network.Config()["dns.mode"] != "none" && nicCheckDNSNameConflict(d.inst.Name(), inst.Name) {
 			if sameLogicalInstance {
+				// Skip NICs that are being renamed.
+				_, nicInPendingExpandedDevices := d.inst.ExpandedDevices()[nicName]
+				if !nicInPendingExpandedDevices {
+					return nil
+				}
+
 				return api.StatusErrorf(http.StatusConflict, "Instance DNS name %q conflict between %q and %q because both are connected to same network", strings.ToLower(inst.Name), d.name, nicName)
 			}
 
--- a/test/suites/container_devices_nic_bridged.sh
+++ b/test/suites/container_devices_nic_bridged.sh
@@ -36,6 +36,21 @@ test_container_devices_nic_bridged() {
   # Record how many nics we started with.
   startNicCount=$(find /sys/class/net | wc -l)
 
+  # Test device renaming without "eth0" device applied by profile.
+  lxc profile copy default "${ctName}"
+  lxc profile device remove "${ctName}" eth0
+  lxc init --empty "${ctName}" -d "${SMALL_ROOT_DISK}" -p "${ctName}"
+  lxc config device add "${ctName}" eth0 nic network="${brName}" name=eth0
+  [ "$(lxc config device get "${ctName}" eth0 name)" = "eth0" ]
+  [ "$(lxc config device get "${ctName}" eth0 network)" = "${brName}" ]
+  lxc config show -e "${ctName}" | sed '/^devices:/,/type:/ s/eth0/eth1/' | lxc config edit "${ctName}" # Rename added "eth0" to "eth1"
+  ! lxc config device get "${ctName}" eth0 name || false
+  [ "$(lxc config device get "${ctName}" eth1 name)" = "eth1" ]
+  [ "$(lxc config device get "${ctName}" eth1 network)" = "${brName}" ]
+  # Cleanup for remaining tests.
+  lxc delete "${ctName}" -f
+  lxc profile delete "${ctName}"
+
   # Test pre-launch profile config is applied at launch
   lxc profile copy default "${ctName}"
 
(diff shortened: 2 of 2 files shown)


lxd/device/nic_bridged.go
```diff
@@ -390,6 +390,13 @@ func (d *nicBridged) checkAddressConflict() error {
 			return nil
 		}
 
+		// Skip our own NIC if it is in the pending expanded config.
+		// This allows renaming a NIC without triggering a duplicate device error.
+		_, nicInPendingExpandedConfig := d.inst.ExpandedConfig()[nicName]
```
simondeziel, line 395: Would it be worth it to move the `d.inst.ExpandedConfig` call inside the `if sameLogicalInstance`?

simondeziel (dismissed):

simondeziel (commented):
@kadinsayani could you turn [Tom's reproducer](https://github.com/canonical/lxd/issues/15300#issuecomment-2838819725) into a quick test?

With `sed` address selection, you could limit the editing scope to just the NIC device, like this:

```
lxc config show -e v3 | sed '/^devices:/,/type:/ s/eth0/eth1/' | lxc config edit v3
```

tomponline:
Please can we have some tests for this?

test/suites/container_devices_nic_bridged.sh
```diff
@@ -36,6 +36,17 @@ test_container_devices_nic_bridged() {
   # Record how many nics we started with.
   startNicCount=$(find /sys/class/net | wc -l)
 
+  # Test device renaming without "eth0" device applied by profile.
+  lxc profile copy default "${ctName}"
+  lxc profile device remove "${ctName}" eth0
+  lxc init testimage "${ctName}" -p "${ctName}"
```
simondeziel, line 42: If the instance can be offline it'd be best to go with:

```suggestion
  lxc init --empty "${ctName}" -d "${SMALL_ROOT_DISK}" -p "${ctName}"
```

but if it needs to be running:

```suggestion
  lxc launch testimage "${ctName}" -d "${SMALL_ROOT_DISK}" -p "${ctName}"
```

FYI, I intend to do a mass change to add those `-d "${SMALL_ROOT_DISK}"` bits but I have yet to get around to doing it :/

test/suites/container_devices_nic_bridged.sh
```diff
@@ -36,6 +36,17 @@ test_container_devices_nic_bridged() {
...
+  lxc config device add "${ctName}" eth0 nic network=${brName}
```
simondeziel, line 43:
```suggestion
  lxc config device add "${ctName}" eth0 nic network="${brName}"
  [ "$(lxc config device get "${ctName}" eth0 name)" = "eth0" ]
  [ "$(lxc config device get "${ctName}" eth0 network)" = "${brName}" ]
```

test/suites/container_devices_nic_bridged.sh
```diff
@@ -36,6 +36,17 @@ test_container_devices_nic_bridged() {
...
+  lxc config show -e "${ctName}" | sed '/^devices:/,/type:/ s/eth0/eth1/' | lxc config edit "${ctName}" # Rename added "eth0" to "eth1"
+  lxc config show -e "${ctName}" | grep -q "eth1"
```
simondeziel, line 45: Since `lxc config show -e` include the volatile keys that contain the NIC name it would be safer to directly get the device key you expect to be there (or absent):

```suggestion
  ! lxc config device get "${ctName}" eth0 name || false
  [ "$(lxc config device get "${ctName}" eth1 name)" = "eth1" ]
  [ "$(lxc config device get "${ctName}" eth1 network)" = "${brName}" ]
```

test/suites/container_devices_nic_bridged.sh
```diff
@@ -155,6 +166,9 @@ test_container_devices_nic_bridged() {
     hwaddr="${ctMAC}" \
     mtu=1401
 
+  # Test hot plugging a container nic with a different name.
+  ! lxc config device add "${ctName}" eth1 nic nictype=bridged name=eth1 parent=${brName} || false
```
simondeziel, line 170:
```suggestion
  ! lxc config device add "${ctName}" eth1 nic nictype=bridged name=eth1 parent="${brName}" || false
```

simondeziel (commented):
Shaping good!

test/suites/container_devices_nic_bridged.sh
kadinsayani: > FYI, I intend to do a mass change to add those `-d "${SMALL_ROOT_DISK}"` bits but I have yet to get around to doing it :/

Do you think a `gopatch` would help here, or are there enough exceptions where `SMALL_ROOT_DISK` isn’t applicable that we’d need to review each `lxc init/launch` manually?

kadinsayani: For now I've added `-d "${SMALL_ROOT_DISK}"` where applicable in the container device NIC bridge tests.

simondeziel (approved):

test/suites/container_devices_nic_bridged.sh
simondeziel: I've ran into regressions when applying it massively without much consideration. That's why I put the PR back on the back burner :/

Thanks for taking care of that script!

tomponline (approved):
LGTM thanks!

PR #15523, opened 2025-05-05
Storage: Add all custom storage vols to instance's backup config

roosterfish:
Preceded by:
* https://github.com/canonical/lxd/pull/15775
* https://github.com/canonical/lxd/pull/15776
* https://github.com/canonical/lxd/pull/15777
* https://github.com/canonical/lxd/pull/15859

### Overview

This PR uses the new backup config metadata format introduced in https://github.com/canonical/lxd/pull/15129 to track all of an instance's custom storage volumes.
In addition all of the storage volume's storage pools are now listed too.

Whenever a custom volume (or its snapshots) gets updated, LXD has to check whether or not this custom volume is attached to any instance. If it's attached, with the changes proposed in this PR, LXD rewrites the backup config file to propagate the latest state of the volume.

### Detail

In order to pick up an instance's custom volume changes, the outcome of such a change has to be persisted in the parent instance's backup config.
LXD currently only has support for `GenerateCustomVolumeBackupConfig` which produces a backup config file (using the standard format) to be included in custom volume exports.

This PR grows a new `UpdateCustomVolumeBackupFile` (same as `UpdateInstanceBackupFile`) which allows updates of a custom volume to be persisted to file. With the only difference that those updates are written to all the instances using the custom volume.

As multiple instances can use the same volume, looking up all the necessary information to rebuild an instance's backup config might be resource intensive. Therefore the instance grows another `GenerateInstanceCustomVolumeBackupConfig` function which allows generating the backup config for the instance's custom volumes by leveraging caching mechanisms. Also an already prepared backup config can now be passed down to `GenerateInstanceBackupConfig` and `UpdateInstanceBackupFile` if its already crafted from the cache.

Having the `GenerateInstanceCustomVolumeBackupConfig` exported on the `Pool` interface allows arbitrary callers to either generate the custom volume's config to be passed when creating an instance's backup config or simply provide nothing. This is helpful in the case of exporting instances as the received backup doesn't contain any information about custom volumes attached to the respective instance.

Fixes https://github.com/canonical/lxd/issues/11712

### DIFF
--- a/lxd/backup.go
+++ b/lxd/backup.go
@@ -255,7 +255,8 @@ func backupWriteIndex(sourceInst instance.Instance, pool storagePools.Pool, opti
 		return os.ErrNotExist
 	}
 
-	config, err := pool.GenerateInstanceBackupConfig(sourceInst, snapshots, nil)
+	// Do not include any custom storage volumes (and their pools) in the backup config.
+	config, err := pool.GenerateInstanceBackupConfig(sourceInst, snapshots, nil, nil)
 	if err != nil {
 		return fmt.Errorf("Failed generating instance backup config: %w", err)
 	}
--- a/lxd/instance/drivers/driver_lxc.go
+++ b/lxd/instance/drivers/driver_lxc.go
@@ -8283,8 +8283,13 @@ func (d *lxc) UpdateBackupFile() error {
 		return err
 	}
 
+	volBackupConf, err := pool.GenerateInstanceCustomVolumeBackupConfig(d, nil, true, nil)
+	if err != nil {
+		return fmt.Errorf("Failed generating instance custom volume config: %w", err)
+	}
+
 	// Use the global metadata version.
-	return pool.UpdateInstanceBackupFile(d, true, nil, config.DefaultMetadataVersion, nil)
+	return pool.UpdateInstanceBackupFile(d, true, volBackupConf, config.DefaultMetadataVersion, nil)
 }
 
 // Info returns "lxc" and the currently loaded version of LXC.
--- a/lxd/instance/drivers/driver_lxc.go
+++ b/lxd/instance/drivers/driver_lxc.go
@@ -5324,7 +5324,7 @@ func (d *lxc) MigrateSend(args instance.MigrateSendArgs) (err error) {
 		}
 	}
 
-	srcConfig, err := pool.GenerateInstanceBackupConfig(d, args.Snapshots, d.op)
+	srcConfig, err := pool.GenerateInstanceBackupConfig(d, args.Snapshots, nil, d.op)
 	if err != nil {
 		return fmt.Errorf("Failed generating instance migration config: %w", err)
 	}
@@ -8284,7 +8284,7 @@ func (d *lxc) UpdateBackupFile() error {
 	}
 
 	// Use the global metadata version.
-	return pool.UpdateInstanceBackupFile(d, true, config.DefaultMetadataVersion, nil)
+	return pool.UpdateInstanceBackupFile(d, true, nil, config.DefaultMetadataVersion, nil)
 }
 
 // Info returns "lxc" and the currently loaded version of LXC.
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -6956,9 +6956,8 @@ func (b *lxdBackend) DeleteCustomVolumeSnapshot(projectName, volName string, op 
 	l.Debug("DeleteCustomVolumeSnapshot started")
 	defer l.Debug("DeleteCustomVolumeSnapshot finished")
 
-	isSnap := shared.IsSnapshot(volName)
-
-	if !isSnap {
+	parentVolName, _, isSnapshot := api.GetParentAndSnapshotName(volName)
+	if !isSnapshot {
 		return errors.New("Volume name must be a snapshot")
 	}
 
@@ -6991,6 +6990,28 @@ func (b *lxdBackend) DeleteCustomVolumeSnapshot(projectName, volName string, op 
 		vol.SetParentUUID(parentUUID)
 	}
 
+	// Get the parent volume.
+	parentVol, err := VolumeDBGet(b, projectName, parentVolName, drivers.VolumeTypeCustom)
+	if err != nil {
+		return err
+	}
+
+	var instances []instance.Instance
+
+	// Fetch all instances which are currently using the custom volume in one of their devices.
+	err = VolumeUsedByInstanceDevices(b.state, b.name, projectName, &parentVol.StorageVolume, true, func(dbInst db.InstanceArgs, project api.Project, _ []string) error {
+		inst, err := instance.Load(b.state, dbInst, project)
+		if err != nil {
+			return err
+		}
+
+		instances = append(instances, inst)
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
 	// Delete the snapshot from the storage device.
 	// Must come before DB VolumeDBDelete so that the volume ID is still available.
 	volExists, err := b.driver.HasVolume(vol)
@@ -7007,6 +7028,12 @@ func (b *lxdBackend) DeleteCustomVolumeSnapshot(projectName, volName string, op 
 
 	// Remove the snapshot volume record from the database.
 	err = VolumeDBDelete(b, projectName, volName, vol.Type())
+	if err != nil {
+		return err
+	}
+
+	// Update the backup config file of the corresponding instances.
+	err = b.UpdateCustomVolumeBackupFiles(projectName, parentVolName, true, instances, op)
 	if err != nil {
 		return err
 	}
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -940,9 +940,14 @@ func (b *lxdBackend) CreateInstanceFromBackup(srcBackup backup.Info, srcData io.
 			return err
 		}
 
+		volBackupConf, err := b.GenerateInstanceCustomVolumeBackupConfig(inst, nil, true, op)
+		if err != nil {
+			return fmt.Errorf("Failed generating instance custom volume config: %w", err)
+		}
+
 		// Save any changes that have occurred to the instance's config to the on-disk backup.yaml file.
 		// Use the global metadata version.
-		err = b.UpdateInstanceBackupFile(inst, false, backupConfig.DefaultMetadataVersion, op)
+		err = b.UpdateInstanceBackupFile(inst, false, volBackupConf, backupConfig.DefaultMetadataVersion, op)
 		if err != nil {
 			return fmt.Errorf("Failed updating backup file: %w", err)
 		}
@@ -1059,8 +1064,13 @@ func (b *lxdBackend) CreateInstanceFromCopy(inst instance.Instance, src instance
 		return errors.New("Source pool is not a lxdBackend")
 	}
 
+	volSrcConfig, err := srcPool.GenerateInstanceCustomVolumeBackupConfig(src, nil, true, op)
+	if err != nil {
+		return fmt.Errorf("Failed generating instance custom volume copy config: %w", err)
+	}
+
 	// Check source volume exists, and get its config including all of the snapshots.
-	srcConfig, err := srcPool.GenerateInstanceBackupConfig(src, true, op)
+	srcConfig, err := srcPool.GenerateInstanceBackupConfig(src, true, volSrcConfig, op)
 	if err != nil {
 		return fmt.Errorf("Failed generating instance copy config: %w", err)
 	}
@@ -1617,8 +1627,13 @@ func (b *lxdBackend) RefreshInstance(inst instance.Instance, src instance.Instan
 		return errors.New("Source pool is not a lxdBackend")
 	}
 
+	volSrcConfig, err := srcPool.GenerateInstanceCustomVolumeBackupConfig(src, nil, true, op)
+	if err != nil {
+		return fmt.Errorf("Failed generating instance custom volume refresh config: %w", err)
+	}
+
 	// Check source volume exists, and get its config including all of the snapshots.
-	srcConfig, err := srcPool.GenerateInstanceBackupConfig(src, true, op)
+	srcConfig, err := srcPool.GenerateInstanceBackupConfig(src, true, volSrcConfig, op)
 	if err != nil {
 		return fmt.Errorf("Failed generating instance refresh config: %w", err)
 	}
@@ -3302,9 +3317,14 @@ func (b *lxdBackend) BackupInstance(inst instance.Instance, tarWriter *instancew
 		return err
 	}
 
+	volBackupConf, err := b.GenerateInstanceCustomVolumeBackupConfig(inst, nil, snapshots, op)
+	if err != nil {
+		return fmt.Errorf("Failed generating instance custom volume config: %w", err)
+	}
+
 	// Ensure the backup file reflects current config.
 	// Use the version requested by the caller to write the correct backup file format.
-	err = b.UpdateInstanceBackupFile(inst, snapshots, version, op)
+	err = b.UpdateInstanceBackupFile(inst, snapshots, volBackupConf, version, op)
 	if err != nil {
 		return err
 	}
@@ -3786,9 +3806,14 @@ func (b *lxdBackend) RenameInstanceSnapshot(inst instance.Instance, newName stri
 		})
 	})
 
+	volBackupConf, err := b.GenerateInstanceCustomVolumeBackupConfig(inst, nil, true, op)
+	if err != nil {
+		return fmt.Errorf("Failed generating instance custom volume config: %w", err)
+	}
+
 	// Ensure the backup file reflects current config.
 	// Use the global metadata version.
-	err = b.UpdateInstanceBackupFile(inst, true, backupConfig.DefaultMetadataVersion, op)
+	err = b.UpdateInstanceBackupFile(inst, true, volBackupConf, backupConfig.DefaultMetadataVersion, op)
 	if err != nil {
 		return err
 	}
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -7019,6 +7019,7 @@ func (b *lxdBackend) GenerateCustomVolumeBackupConfig(projectName string, volNam
 	return &backupConfig.Config{
 		Version: api.BackupMetadataVersion2,
 		Volumes: []*backupConfig.Volume{volConfig},
+		Pools:   []*api.StoragePool{&b.db},
 	}, nil
 }
 
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -6247,6 +6247,27 @@ func (b *lxdBackend) UpdateCustomVolume(projectName string, volName string, newD
 		return nil
 	}
 
+	var instances []instance.Instance
+	err = VolumeUsedByInstanceDevices(b.state, b.name, projectName, &curVol.StorageVolume, true, func(dbInst db.InstanceArgs, project api.Project, _ []string) error {
+		inst, err := instance.Load(b.state, dbInst, project)
+		if err != nil {
+			return err
+		}
+
+		if changedConfig["security.shifted"] != "" {
+			// Confirm that no running instances are using it when changing shifted state.
+			if inst.IsRunning() && changedConfig["security.shifted"] != "" {
+				return errors.New("Cannot modify shifting with running instances using the volume")
+			}
+		}
+
+		instances = append(instances, inst)
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
 	if len(changedConfig) != 0 {
 		// Forbid changing the config for ISO custom volumes as they are read-only.
 		if contentType == drivers.ContentTypeISO {
@@ -6261,26 +6282,6 @@ func (b *lxdBackend) UpdateCustomVolume(projectName string, volName string, newD
 		// Check that the volume's volatile.uuid property isn't being changed.
 		if changedConfig["volatile.uuid"] != "" {
 			return errors.New(`Custom volume "volatile.uuid" property cannot be changed`)
-		}
-
-		// Check for config changing that is not allowed when running instances are using it.
-		if changedConfig["security.shifted"] != "" {
-			err = VolumeUsedByInstanceDevices(b.state, b.name, projectName, &curVol.StorageVolume, true, func(dbInst db.InstanceArgs, project api.Project, _ []string) error {
-				inst, err := instance.Load(b.state, dbInst, project)
-				if err != nil {
-					return err
-				}
-
-				// Confirm that no running instances are using it when changing shifted state.
-				if inst.IsRunning() && changedConfig["security.shifted"] != "" {
-					return errors.New("Cannot modify shifting with running instances using the volume")
-				}
-
-				return nil
-			})
-			if err != nil {
-				return err
-			}
 		}
 
 		sharedVolume, ok := changedConfig["security.shared"]
@@ -6309,6 +6310,13 @@ func (b *lxdBackend) UpdateCustomVolume(projectName string, volName string, newD
 	revert := revert.New()
 	defer revert.Fail()
 
+	// Add the backup file revert before changing the DB record.
+	// This ensures the updated DB entry is reverted before trying to reset the file.
+	revert.Add(func() {
+		// Reset the instance backup file if the custom volume update didn't succeed.
+		_ = b.UpdateCustomVolumeBackupFiles(projectName, volName, true, instances, op)
+	})
+
 	// Update the database.
 	err = b.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
 		return tx.UpdateStoragePoolVolume(ctx, projectName, volName, cluster.StoragePoolVolumeTypeCustom, b.ID(), newDesc, newConfig)
@@ -6323,6 +6331,12 @@ func (b *lxdBackend) UpdateCustomVolume(projectName string, volName string, newD
 			return tx.UpdateStoragePoolVolume(ctx, projectName, volName, cluster.StoragePoolVolumeTypeCustom, b.ID(), curVol.Description, curVol.Config)
 		})
 	})
+
+	// Update the instance's backup files.
+	err = b.UpdateCustomVolumeBackupFiles(projectName, volName, true, instances, op)
+	if err != nil {
+		return err
+	}
 
 	b.state.Events.SendLifecycle(projectName, lifecycle.StorageVolumeUpdated.Event(newVol, string(newVol.Type()), projectName, op, nil))
 
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -6351,7 +6351,8 @@ func (b *lxdBackend) UpdateCustomVolumeSnapshot(projectName string, volName stri
 	l.Debug("UpdateCustomVolumeSnapshot started")
 	defer l.Debug("UpdateCustomVolumeSnapshot finished")
 
-	if !shared.IsSnapshot(volName) {
+	parentName, _, isSnap := api.GetParentAndSnapshotName(volName)
+	if !isSnap {
 		return errors.New("Volume must be a snapshot")
 	}
 
@@ -6384,8 +6385,36 @@ func (b *lxdBackend) UpdateCustomVolumeSnapshot(projectName string, volName stri
 		return nil
 	}
 
+	// Fetch parent vol.
+	parentVol, err := VolumeDBGet(b, projectName, parentName, drivers.VolumeTypeCustom)
+	if err != nil {
+		return err
+	}
+
+	var instances []instance.Instance
+
+	// Fetch all instances which are currently using the custom volume in one of their devices.
+	err = VolumeUsedByInstanceDevices(b.state, b.name, projectName, &parentVol.StorageVolume, true, func(dbInst db.InstanceArgs, project api.Project, _ []string) error {
+		inst, err := instance.Load(b.state, dbInst, project)
+		if err != nil {
+			return err
+		}
+
+		instances = append(instances, inst)
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
 	revert := revert.New()
 	defer revert.Fail()
+
+	// Add the instance's backup file revert before adding the DB record reverter.
+	// This ensures the renamed DB entry is reverted before trying to reset the file.
+	revert.Add(func() {
+		_ = b.UpdateCustomVolumeBackupFiles(projectName, parentName, true, instances, op)
+	})
 
 	// Update the database. Use current config.
 	err = b.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
@@ -6400,6 +6429,12 @@ func (b *lxdBackend) UpdateCustomVolumeSnapshot(projectName string, volName stri
 			return tx.UpdateStorageVolumeSnapshot(ctx, projectName, volName, cluster.StoragePoolVolumeTypeCustom, b.ID(), curVol.Description, curVol.Config, curExpiryDate)
 		})
 	})
+
+	// Update the instance's backup files.
+	err = b.UpdateCustomVolumeBackupFiles(projectName, parentName, true, instances, op)
+	if err != nil {
+		return err
+	}
 
 	vol := b.GetVolume(drivers.VolumeTypeCustom, drivers.ContentType(curVol.ContentType), curVol.Name, curVol.Config)
 	b.state.Events.SendLifecycle(projectName, lifecycle.StorageVolumeSnapshotUpdated.Event(vol, string(vol.Type()), projectName, op, nil))
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -6764,6 +6764,28 @@ func (b *lxdBackend) CreateCustomVolumeSnapshot(projectName, volName string, new
 		newExpiryDate = &expiry
 	}
 
+	var instances []instance.Instance
+
+	// Fetch all instances which are currently using the custom volume in one of their devices.
+	err = VolumeUsedByInstanceDevices(b.state, b.name, projectName, &parentVol.StorageVolume, true, func(dbInst db.InstanceArgs, project api.Project, _ []string) error {
+		inst, err := instance.Load(b.state, dbInst, project)
+		if err != nil {
+			return err
+		}
+
+		instances = append(instances, inst)
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	// Add the instance's backup file revert before adding the DB record reverter.
+	// This ensures the added DB entry is reverted before trying to reset the file.
+	revert.Add(func() {
+		_ = b.UpdateCustomVolumeBackupFiles(projectName, volName, true, instances, op)
+	})
+
 	// Validate config and create database entry for new storage volume.
 	// Copy volume config from parent.
 	err = VolumeDBCreate(b, projectName, fullSnapshotName, description, drivers.VolumeTypeCustom, true, vol.Config(), snapshotCreationDate, *newExpiryDate, drivers.ContentType(parentVol.ContentType), false, true)
@@ -6780,6 +6802,12 @@ func (b *lxdBackend) CreateCustomVolumeSnapshot(projectName, volName string, new
 	}
 
 	revert.Add(func() { _ = b.driver.DeleteVolumeSnapshot(vol, op) })
+
+	// Update the backup config file of the corresponding instances.
+	err = b.UpdateCustomVolumeBackupFiles(projectName, volName, true, instances, op)
+	if err != nil {
+		return err
+	}
 
 	b.state.Events.SendLifecycle(projectName, lifecycle.StorageVolumeSnapshotCreated.Event(vol, string(vol.Type()), projectName, op, logger.Ctx{"type": vol.Type()}))
 
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -35,6 +35,7 @@ import (
 	"github.com/canonical/lxd/lxd/db"
 	"github.com/canonical/lxd/lxd/db/cluster"
 	"github.com/canonical/lxd/lxd/device/config"
+	"github.com/canonical/lxd/lxd/device/filters"
 	"github.com/canonical/lxd/lxd/instance"
 	"github.com/canonical/lxd/lxd/instance/instancetype"
 	"github.com/canonical/lxd/lxd/instancewriter"
@@ -8129,4 +8130,77 @@ func (b *lxdBackend) getParentVolumeUUID(vol drivers.Volume, projectName string)
 	}
 
 	return parentUUID, nil
+}
+
+// GenerateInstanceCustomVolumeBackupConfig returns the backup config entry for this instance's custom volumes.
+// To not require looking up all the entities from the DB for every instance, the custom volume's backup config
+// can be created independently from the actual instance's main backup config by specifying an optional cache.
+// The caller can decide to use this cache across multiple instances.
+// That is helpful in situations where a custom volume gets updated which causes the backup config files of all
+// instances using this volume to be updated.
+func (b *lxdBackend) GenerateInstanceCustomVolumeBackupConfig(inst instance.Instance, cache *backupConfigCache, snapshots bool, op *operations.Operation) (*backupConfig.Config, error) {
+	// Setup a cache if not provided.
+	// This will allow caching pool and volume backup configs for the given instance.
+	if cache == nil {
+		cache = newBackupConfigCache(b)
+	}
+
+	// Get the right project name for the disk device.
+	instanceProject := inst.Project()
+	projectName := project.StorageVolumeProjectFromRecord(&instanceProject, cluster.StoragePoolVolumeTypeCustom)
+
+	var instanceBackupConf = &backupConfig.Config{}
+
+	// Skip non-disk devices, host filesystem shares without a pool and the instance's root disk itself.
+	for _, device := range inst.ExpandedDevices().Filter(filters.IsCustomVolumeDisk) {
+		vol, err := cache.getVolume(projectName, device["pool"], device["source"], snapshots, op)
+		if err != nil {
+			// When restoring an instance from snapshot, some of the custom vols which were attached
+			// whilst taking the snapshot might not exist anymore.
+			// To not cause failures during restore, skip those volumes because we cannot include them in the instance's backup config.
+			// These volumes are still listed in the list of instance devices and will cause an error when
+			// trying to start the instance.
+			// The instance cannot be started in any case until the non-existing disk device gets removed.
+			if api.StatusErrorCheck(err, http.StatusNotFound) {
+				logger.Warn("Missing custom volume to include in instance's backup config", logger.Ctx{"project": projectName, "pool": device["pool"], "volume": device["source"], "instance": inst.Name()})
+
+				continue
+			}
+
+			return nil, err
+		}
+
+		volPool, err := cache.getPool(device["pool"])
+		if err != nil {
+			return nil, err
+		}
+
+		volFound := false
+		for _, existingVol := range instanceBackupConf.Volumes {
+			// A volume can be considered identical if both it's name and pool is matching.
+			if existingVol.Name == vol.Name && existingVol.Pool == volPool.Name() {
+				volFound = true
+			}
+		}
+
+		// Append vol to the backup config if it doesn't yet exist.
+		if !volFound {
+			instanceBackupConf.Volumes = append(instanceBackupConf.Volumes, vol)
+		}
+
+		poolFound := false
+		for _, existingPool := range instanceBackupConf.Pools {
+			if existingPool.Name == volPool.Name() {
+				poolFound = true
+			}
+		}
+
+		// Append vol pool to the backup config if it doesn't yet exist.
+		if !poolFound {
+			apiPool := volPool.ToAPI()
+			instanceBackupConf.Pools = append(instanceBackupConf.Pools, &apiPool)
+		}
+	}
+
+	return instanceBackupConf, nil
 }
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -6830,6 +6830,7 @@ func (b *lxdBackend) RenameCustomVolumeSnapshot(projectName, volName string, new
 		return errors.New("Invalid new snapshot name")
 	}
 
+	// Fetch the snapshot vol.
 	volume, err := VolumeDBGet(b, projectName, volName, drivers.VolumeTypeCustom)
 	if err != nil {
 		return err
@@ -6839,6 +6840,28 @@ func (b *lxdBackend) RenameCustomVolumeSnapshot(projectName, volName string, new
 	volStorageName := project.StorageVolume(projectName, volName)
 
 	vol := b.GetVolume(drivers.VolumeTypeCustom, drivers.ContentType(volume.ContentType), volStorageName, volume.Config)
+
+	// Fetch the parent vol.
+	parentVol, err := VolumeDBGet(b, projectName, parentName, drivers.VolumeTypeCustom)
+	if err != nil {
+		return err
+	}
+
+	var instances []instance.Instance
+
+	// Fetch all instances which are currently using the custom volume in one of their devices.
+	err = VolumeUsedByInstanceDevices(b.state, b.name, projectName, &parentVol.StorageVolume, true, func(dbInst db.InstanceArgs, project api.Project, _ []string) error {
+		inst, err := instance.Load(b.state, dbInst, project)
+		if err != nil {
+			return err
+		}
+
+		instances = append(instances, inst)
+		return nil
+	})
+	if err != nil {
+		return err
+	}
 
 	revert := revert.New()
 	defer revert.Fail()
@@ -6861,6 +6884,12 @@ func (b *lxdBackend) RenameCustomVolumeSnapshot(projectName, volName string, new
 		_ = b.driver.RenameVolumeSnapshot(newVol, oldSnapshotName, op)
 	})
 
+	// Add the instance's backup file revert before adding the DB record reverter.
+	// This ensures the renamed DB entry is reverted before trying to reset the file.
+	revert.Add(func() {
+		_ = b.UpdateCustomVolumeBackupFiles(projectName, parentName, true, instances, op)
+	})
+
 	err = b.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
 		return tx.RenameStoragePoolVolume(ctx, projectName, volName, newVolName, cluster.StoragePoolVolumeTypeCustom, b.ID())
 	})
@@ -6873,6 +6902,12 @@ func (b *lxdBackend) RenameCustomVolumeSnapshot(projectName, volName string, new
 			return tx.RenameStoragePoolVolume(ctx, projectName, newVolName, volName, cluster.StoragePoolVolumeTypeCustom, b.ID())
 		})
 	})
+
+	// Update the backup config file of the corresponding instances.
+	err = b.UpdateCustomVolumeBackupFiles(projectName, parentName, true, instances, op)
+	if err != nil {
+		return err
+	}
 
 	b.state.Events.SendLifecycle(projectName, lifecycle.StorageVolumeSnapshotRenamed.Event(vol, string(vol.Type()), projectName, op, logger.Ctx{"old_name": oldSnapshotName}))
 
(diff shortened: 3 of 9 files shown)


tomponline:
> Furthermore a general bug is fixed in LXD which caused an instance's backup file not to be updated when updating the storage pool of its root volume. This action is now performed for both an instance's root vol as well as all of its attached custom vols.
> 
> Fixes #11712

Please can we separate that into another pr ?

lxd/backup/config/backup_config.go
```diff
@@ -165,12 +165,12 @@ func (c *Config) RootVolume() (*Volume, error) {
 // Unlike RootVolume, CustomVolume always returns the first and only volume in the list.
 func (c *Config) CustomVolume() (*Volume, error) {
 	if c.Instance != nil {
-		return nil, errors.New("Instance config cannot be set for custom volumes")
+		return nil, errors.New("Invalid custom volume config")
```
tomponline: The old error message seemed clearer and more specific as to what the issue was, the new one seems vaguer?

lxd/storage/backend_lxd.go
```diff
@@ -7090,9 +7091,32 @@ func (b *lxdBackend) GenerateInstanceBackupConfig(inst instance.Instance, snapsh
...
 			}
 		}
+
+		for _, device := range inst.ExpandedDevices() {
+			if device["type"] != "disk" || instancetype.IsRootDiskDevice(device) {
+				continue
+			}
+
+			// Custom storage volume might be in a different pool, so load the pool.
+			pool, err := LoadByName(b.state, device["pool"])
```
tomponline: we should be more efficient here and avoid loading the same pool over and over again if we have multiple disks from the same pool.

tomponline: also if the volume is from the same pool as the current backend context then we dont need to load it at all.

lxd/storage/backend_lxd.go
```diff
@@ -7090,9 +7091,32 @@ func (b *lxdBackend) GenerateInstanceBackupConfig(inst instance.Instance, snapsh
...
+			if err != nil {
+				return nil, err
+			}
+
+			volConfig, err := pool.GenerateCustomVolumeBackupConfig(inst.Project().Name, device["source"], true, op)
+			if err != nil {
+				return nil, fmt.Errorf("Failed generating volume backup config: %w", err)
```
tomponline: adding context for the device, pool and volume name would be useful here.

lxd/backup/config/backup_config.go
```diff
@@ -159,7 +159,10 @@ func (c *Config) RootVolume() (*Volume, error) {
 			continue
 		}
 
-		if volume.Name == c.Instance.Name {
+		// An instance's root volume uses the same name as its parent instance.
```
tomponline: Does the logic below for getting an instance snapshot's primary volume hold true once custom volumes are included in the config?

lxd/storage/backend_lxd.go
```diff
@@ -7093,36 +7093,78 @@ func (b *lxdBackend) GenerateInstanceBackupConfig(inst instance.Instance, snapsh
...
+				// Skip non-disk devices, directory shares without a pool and the instance's root disk itself.
+				if device["type"] != "disk" || device["pool"] == "" || instancetype.IsRootDiskDevice(device) {
+					continue
+				}
 
-			volConfig, err := pool.GenerateCustomVolumeBackupConfig(inst.Project().Name, device["source"], true, op)
-			if err != nil {
-				return nil, fmt.Errorf("Failed generating volume backup config: %w", err)
+				// Custom volume might be in a different pool, so load the pool.
+				pool, err := LoadByName(b.state, device["pool"])
```
tomponline: same comment as earlier about not loading the same pool multiple times from db

lxd/storage/backend_lxd.go
```diff
@@ -7090,9 +7091,32 @@ func (b *lxdBackend) GenerateInstanceBackupConfig(inst instance.Instance, snapsh
 				})
 			}
 		}
+
+		for _, device := range inst.ExpandedDevices() {
+			if device["type"] != "disk" || instancetype.IsRootDiskDevice(device) {
```
tomponline: later you have similar logic but it also contains a pool name check:

```go
device["pool"] == ""
```

shouldnt this also have this check?

lxd/storage/backend_lxd.go
```diff
@@ -7093,36 +7093,78 @@ func (b *lxdBackend) GenerateInstanceBackupConfig(inst instance.Instance, snapsh
...
+					// When restoring an instance from snapshot, some of the custom vols which were attached
+					// whilst taking the snapshot might not exist anymore.
+					// To not cause failures during restore, skip those volumes.
+					// These volumes are still listed in the list of instance devices and will cause an error when
+					// trying to start the instance.
+					if api.StatusErrorCheck(err, http.StatusNotFound) {
+						continue
+					}
+
+					return nil, fmt.Errorf("Failed generating volume backup config: %w", err)
```
tomponline: adding context around which device, pool and volume would be useful here

lxd/backup.go
tomponline, line 1: `lxd: Update invocations` isnt going to be particularly useful when backporting.

lxd/storage/backend_lxd.go
```diff
@@ -7093,36 +7093,78 @@ func (b *lxdBackend) GenerateInstanceBackupConfig(inst instance.Instance, snapsh
...
-			}
+		var customVols []*backupConfig.Volume
+		var customVolPools []*api.StoragePool
 
-			// Custom storage volume might be in a different pool, so load the pool.
-			pool, err := LoadByName(b.state, device["pool"])
-			if err != nil {
-				return nil, err
-			}
+		if volBackupConf != nil {
```
tomponline, line 7326: This part triggered me later on when i saw you are passing an empty (but non-nil) list when generating an export index file.

My feeling on this now is that this function `GenerateInstanceBackupConfig` is trying to do too much and instead we should be breaking out the generation of custom volume config selection into a separate function, so that we can just this function `GenerateInstanceBackupConfig` take the proposed `volBackupConf` (but called `customVolConf` or similar to indicate its just for custom volumes) and then have this function append them to its root volume config.

Then we can have the call sites decide whether or not to include custom volume config, and it should then also be possible to be more efficient about loading that data too.

What do you think?

lxd/backup.go
```diff
@@ -252,7 +252,13 @@ func backupWriteIndex(sourceInst instance.Instance, pool storagePools.Pool, opti
 		return os.ErrNotExist
 	}
 
-	config, err := pool.GenerateInstanceBackupConfig(sourceInst, snapshots, nil)
+	// Provide an empty list (non-nil) to not include any custom storage volumes in the backup config.
+	volConfig := &backupConfig.Config{
```
tomponline: This feels like a kludge and makes me think that GenerateInstanceBackupConfig is being too clever, and rather we should have an external function generate custom volume config and have the caller decide to call it and pass the config to `GenerateInstanceBackupConfig`.

lxd/storage/backend_lxd.go
```diff
@@ -6952,6 +6952,58 @@ func (b *lxdBackend) createStorageStructure(path string) error {
...
+	l.Debug("UpdateCustomVolumeBackupFile started")
+	defer l.Debug("UpdateCustomVolumeBackupFile finished")
+
+	vol, err := VolumeDBGet(b, projectName, volName, drivers.VolumeTypeCustom)
+	if err != nil {
+		return err
+	}
+
+	// Fetch all instances which are currently using the custom volume in one of their devices if not provided already.
+	if instances == nil {
```
tomponline: this pattern of "if nil then lookup internally" feels a bit confusing to the caller as it differentiates between nil and empty slice - can we instead break this out into its own function that can then be used to provide a list of instances to this function?

lxd/storage/backend_lxd.go
```diff
@@ -6952,6 +6952,58 @@ func (b *lxdBackend) createStorageStructure(path string) error {
...
+	// Update the backup config file of all instances.
+	// Try updating all of the instance's backup files using a best effort strategy.
+	// In case a custom volume is used by many instances, it might happen that an instance gets deleted
+	// whilst this function tries to propagate an update of a custom volume.
+	// A lock isn't acquired in this case for all the instances as this might cause too much interruption
+	// as every update of an instance's backup file requires mounting the corresponding volume.
+	for _, inst := range instances {
+		// Don't invoke the the update of the instance's backup file directly on *lxdBackend.
+		// The instance's root vol might be located on a different storage pool which gets loaded when calling the instance's UpdateBackupFile.
+		err = inst.UpdateBackupFile(volConfig)
```
tomponline: this looks wrong to me, you're getting the single update volume's config and passing it to `inst.UpdateBackupFile` which in turn will pass it to `pool.UpdateInstanceBackupFile` which in turn will pass it to `GenerateInstanceBackupConfig` which means it wont lookup the config for any other attached volumes on the instance (https://github.com/canonical/lxd/pull/15523/files#diff-029fdd0e1afdacb0e66630c0d7e813772122db9a7d036060902838db315d9c4dR7228) and will remove them from the backup.yaml file.

lxd/storage/backend_lxd.go
tomponline, line 1: can some of this commit be moved into its own PR - it seems partially unrelated to the main change?

tomponline, line 1: can the reverter usage be separated into a PR that is separate to this one?

tomponline (changes requested):
A few comments:

1. Are there any changes that can be moved out into a separate PR that are not directly related to this PR (to reduce the size and keep the focus of this PR).
2. The pattern of "if you pass nil we will do a lookup internally" is not one I've found to be that helpful, and I would rather see the separation of logic into a separate function that can be used to populate the argument to main function rather than embedding it inside. This helps also to reduce the cyclomatic complexity in TICs.
3. Do you have any concerns around the DB query amplification, especially wrt to repeat queries for loading pools and volume configs? I am wondering if some sort of object cacher could be used for the duration of an operation to provide `get pool` and `get volume` functions which then only load from the DB if they've not already done so?

lxd/backup/config/backup_config.go
roosterfish: I have reverted this exact message, not sure anymore why I started modifying it but I did some experiments with vol lookups so probably that is a leftover from it.

lxd/storage/backend_lxd.go
roosterfish: Good point, added caching now.

lxd/backup/config/backup_config.go
roosterfish: Once the `Config` struct is used to represent an instance snapshot, its `.Instance` field [does not get populated](https://github.com/canonical/lxd/blob/main/lxd/storage/backend_lxd.go#L7009-L7010). Also the [pools](https://github.com/canonical/lxd/blob/main/lxd/storage/backend_lxd.go#L6997-L7000) and additional custom vols (added in this PR) aren't added to the struct. Only [the root vol](https://github.com/canonical/lxd/blob/main/lxd/storage/backend_lxd.go#L7070).

This allows the `RootVolume()` func falling back to using the `primaryVolume()` func as there only is a single vol in case of instance snapshots.
The only time the `Config` struct is used to represent instance snapshots is during migration, e.g. when running `lxc cp instance/snap0 remote:instance2`. And in such a case it is desirable to not have the custom vols (and their pools) as part of the `Config`.

tomponline: So an instance snapshot's backup.yaml wont contain custom volumes?

lxd/storage/backend_lxd.go
roosterfish: Duplicate of https://github.com/canonical/lxd/pull/15523#discussion_r2123765178, added caching.

lxd/storage/backend_lxd.go
roosterfish: I wonder how you where able to see this outdated version. Last push to this PR was a week ago before you did the review.

But in any case, this line is only used once in the PR and does use the `device["pool"] == ""` check :)

lxd/backup/config/backup_config.go
roosterfish: > So an instance snapshot's backup.yaml wont contain custom volumes?

Yes that is right.

lxd/storage/backend_lxd.go
roosterfish: Ah you are right, all the other vols except the updated custom vol and root vol are getting removed from the file. Thanks!
I'll also check if this could be covered by a test case too because clearly this went unnoticed.

roosterfish: I think this will fix itself after separating a few things out of `GenerateInstanceBackupConfig` and leave the config generation up to the caller.

lxd/storage/backend_lxd.go
roosterfish: I like the idea and will tackle it as part of https://github.com/canonical/lxd/pull/15523#discussion_r2123888586. Also we can then get rid of the "non-nil" empty list passing to exclude pools/vols for backups.

lxd/backup/config/backup_config.go
tomponline: how does that work? If there is a backup.yaml file on disk in the main instance volume (containing custom vol info) and I snapshot it, wont that file get included in the instance snapshot's backup.yaml file?

lxd/storage/backend_lxd.go
tomponline: possibly you have a commit that has the old code included in it accidentally and then it gets modified later?

lxd/backup/config/backup_config.go
roosterfish: We have to be careful with the use case. One use case is storing the instances (and instance snapshots) backup config to file. In this case it is exactly as you are describing it.

But the backup config is also used for migration of instances (and snapshots):
* (1) Copy instance i1 to i2
* (2) Copy instance snapshot i1/snap0 to i2

And in case (2) the corresponding `Config` struct (which is used during migration) doesn't contain the `.Instance` field and neither does it contain the additional custom vols (+ their pools) that might be attached to this instance.

lxd/storage/backend_lxd.go
roosterfish: Mh, the comment has the GitHub `Outdated` label on it. I found this usually to be present if a comment is made on an old version of the code? I'll check the commits though.

roosterfish:
Once https://github.com/canonical/lxd/pull/15777 is merged I'll rebase this one and then it should be ready for another pass.

lxd/backup.go
roosterfish: Fixed.

lxd/storage/backend_lxd.go
roosterfish: Fixed, updated the Pr's desc to give some more reasoning on what is going on.

lxd/backup.go
roosterfish: Removed this pattern.

lxd/storage/backend_lxd.go
roosterfish: Removed this pattern.

lxd/storage/backend_lxd.go

lxd/storage/backend_lxd.go
roosterfish: Moved, see linked Pr's.

roosterfish: Moved, see linked Pr's.

roosterfish:
Implemented the suggested changes, ready for another review. The PR's description now also contains a summary of the changes including a more in depth explanation of the changes.

lxd/storage/backend_lxd.go
```diff
@@ -7127,9 +7147,20 @@ func (b *lxdBackend) UpdateInstanceBackupFile(inst instance.Instance, snapshots
 		return nil
 	}
 
-	config, err := b.GenerateInstanceBackupConfig(inst, snapshots, op)
+	// When updating the instance backup file and no extra custom volume config is provided,
+	// generate the corresponding config for the custom volumes.
+	if volBackupConf == nil {
```
tomponline: Can we move this part out to the caller of this function too, rather than having the special "if volume config not supplied then load from else where" type logic? Its a bit "magical" to me (in that supplying a nil volume config results in potentially volume config still being added to the resulting backup file).

lxd/storage/utils.go
```diff
@@ -1460,3 +1461,67 @@ func VolumeDetermineNextSnapshotName(ctx context.Context, s *state.State, pool s
 
 	return pattern, nil
 }
+
+// backupConfigCache is used to cache pools and volumes during backup config creation.
```
tomponline: Can we avoid growing the generic utils.go file for this, it feels like it could be well place to go into the backup/config package? Seeing as we're importing it here now anwyay. What do you think?

tomponline: If not can we add it to a new `lxd/storage/backup` package?

lxd/storage/utils.go
```diff
@@ -1460,3 +1461,67 @@ func VolumeDetermineNextSnapshotName(ctx context.Context, s *state.State, pool s
...
+type backupConfigCache struct {
+	pools   map[string]Pool
+	volumes map[string]map[string]*backupConfig.Volume
+	state   *state.State
```
tomponline: It looks like this struct will need initiating, so to help with the usage of this I suggest adding a `New...` function that can accept the state as an argument, and initialise the maps.  Alternatively, if this struct is going to be long lived, then you could also have state passed as an argument to each function to avoid needing this var.

lxd/storage/utils.go
```diff
@@ -1460,3 +1461,67 @@ func VolumeDetermineNextSnapshotName(ctx context.Context, s *state.State, pool s
...
+			return nil, fmt.Errorf("Failed generating backup config for volume %q in project %q: %w", volName, projectName, err)
+		}
+
+		vol, err := volConfig.CustomVolume()
+		if err != nil {
+			return nil, fmt.Errorf("Failed getting the custom volume: %w", err)
+		}
+
+		// Cache the volume.
+		b.volumes[projectName][volName] = vol
```
tomponline: As volume names are unique per pool and project, I would have expected to see pool name be a dimension in these maps.

lxd/storage/backend_lxd.go
```diff
@@ -8131,3 +8131,14 @@ func (b *lxdBackend) getParentVolumeUUID(vol drivers.Volume, projectName string)
 
 	return parentUUID, nil
 }
+
+func (b *lxdBackend) newBackupConfigCache(state *state.State) *backupConfigCache {
```
tomponline: Ah I see, you've initialised the cache here, perhaps having a `New...` function alongside where the cache struct is defined would be clearer, this `New...` function could still take arguments to allow pre-loading the cache as you have done here though.

lxd/storage/backend_lxd.go
```diff
@@ -8132,6 +8132,64 @@ func (b *lxdBackend) getParentVolumeUUID(vol drivers.Volume, projectName string)
 	return parentUUID, nil
 }
 
+// GenerateInstanceCustomVolumeBackupConfig returns the backup config entry for this instance's custom volumes.
+func (b *lxdBackend) GenerateInstanceCustomVolumeBackupConfig(inst instance.Instance, cache *backupConfigCache, snapshots bool, op *operations.Operation) (*backupConfig.Config, error) {
+	// Get the right project name for the device.
```
tomponline, line 8256: device?

tomponline: "disk device" would be clearer IMHO

lxd/storage/backend_lxd.go
```diff
@@ -8132,6 +8132,64 @@ func (b *lxdBackend) getParentVolumeUUID(vol drivers.Volume, projectName string)
...
+	instanceProject := inst.Project()
+	projectName := project.StorageVolumeProjectFromRecord(&instanceProject, cluster.StoragePoolVolumeTypeCustom)
+
+	// Setup a cache if not provided.
```
tomponline, line 8345: Just a nit, but putting this section earlier would help keep the var definitions and comments above close to the device for loop below - helps with context.

lxd/storage/backend_lxd.go
```diff
@@ -8132,6 +8132,64 @@ func (b *lxdBackend) getParentVolumeUUID(vol drivers.Volume, projectName string)
...
+	// This will allow caching pool and volume backup configs for the given instance.
+	if cache == nil {
+		cache = b.newBackupConfigCache(b.state)
+	}
+
+	var instanceBackupConf = &backupConfig.Config{}
+
+	for _, device := range inst.ExpandedDevices() {
+		// Skip non-disk devices, directory shares without a pool and the instance's root disk itself.
+		if device["type"] != "disk" || device["pool"] == "" || instancetype.IsRootDiskDevice(device) {
```
tomponline: I've seen this sort of filter repeated a fair bit in various PRs, I'm wondering if we would all benefit from adding some common pre-defined filters to `lxd/devices/config/devices.Devices`, so we can do something like this:

```go
for _, device := range inst.ExpandedDevices().Disks(){
```
We could have `CustomVolumeDisks()` too that applies the filter you have here.

lxd/storage/backend_lxd.go
```diff
@@ -8132,6 +8132,64 @@ func (b *lxdBackend) getParentVolumeUUID(vol drivers.Volume, projectName string)
...
+			continue
+		}
+
+		vol, err := cache.getVolume(projectName, device["pool"], device["source"], snapshots, op)
+		if err != nil {
+			// When restoring an instance snapshot, the old instance's list of attached custom volumes
```
tomponline: I'd like to understand this comment a bit more (thanks for adding it btw, its of great help).

If I'm reading correctly, this part is need when restoring an instance snapshot that has in its snapshotted configuration disk devices that attach custom volumes that have since been deleted.

In this scenario LXD presumable allows the snapshot restore, but prevents starting of the instance until the issue is corrected?


But the problem you're trying to solve here is that the on-disk backup.yaml is being regenerated on snapshot restore too?

lxd/storage/backend_lxd.go
```diff
@@ -8132,6 +8132,64 @@ func (b *lxdBackend) getParentVolumeUUID(vol drivers.Volume, projectName string)
...
+		}
+
+		instanceBackupConf.Volumes = append(instanceBackupConf.Volumes, vol)
+
+		volPool, err := cache.getPool(device["pool"])
+		if err != nil {
+			return nil, err
+		}
+
+		poolFound := false
```
tomponline, line 8394: I wonder if this section would benefit from being moved to a function called something like `backupConfig.Config.AddPool()` which only adds it if not already in the list? Would this pattern help to de-dupe similar logic elsewhere?

lxd/instance/drivers/driver_lxc.go
```diff
@@ -5350,7 +5350,12 @@ func (d *lxc) MigrateSend(args instance.MigrateSendArgs) error {
 		}
 	}
 
-	srcConfig, err := pool.GenerateInstanceBackupConfig(d, args.Snapshots, d.op)
+	volSrcConfig, err := pool.GenerateInstanceCustomVolumeBackupConfig(d, nil, args.Snapshots, d.op)
```
tomponline: Why do we need to send custom volume config during migrations (at this time, as we do not support migrating volumes along with instances right now)?

tomponline: I'm surprised to see this here, because although the backup.yaml file needs to be materialised on the receiving side, it should be done on the receiving side by taking the instance config and resolving it custom volume and its config on the receiving side.

As the resulting custom volume config might be totally different than on the sending side as the volumes will likely be different volumes with different config.

So I wanted to check we are not blindly accepting custom volume config and pool volume config from the migration sender and writing it into the backup.yaml on the receiver, as that would not be correct.

lxd/storage/pool_interface.go
```diff
@@ -126,6 +126,7 @@ type Pool interface {
 	UnmountCustomVolume(projectName string, volName string, op *operations.Operation) (bool, error)
 	ImportCustomVolume(projectName string, poolVol *backupConfig.Config, op *operations.Operation) (revert.Hook, error)
 	RefreshCustomVolume(projectName string, srcProjectName string, volName, desc string, config map[string]string, srcPoolName, srcVolName string, snapshots bool, op *operations.Operation) error
+	UpdateCustomVolumeBackupFile(projectName string, volName string, snapshots bool, instances []instance.Instance, op *operations.Operation) error
```
tomponline: As this is primarily operating on Instance Backup Files, lets put the Instances front and center (especially as custom volumes dont have backup files).

Suggest a name like `UpdateInstancesBackupFileUsingCustomVolume` - i know its a bit of a mouthful, but seems clearer than the proposed - as it also indicates that the full instance backup file will be re-written, but only for instances using custom volume.

lxd/storage/backend_lxd.go
tomponline: In the earlier commit I suggest renaming this function to `UpdateInstancesBackupFileUsingCustomVolume` for clarity of purpose, but the optional argument of `instances` here makes this problematic as the function is doing 2 different things, only one of which is "UpdateInstancesBackupFileUsingCustomVolume".

Instead we could remove this special logic and instead always require a slice of instances, and move the "using" logic out of this function.

What do you think?

lxd/storage/backend_lxd.go
```diff
@@ -6981,6 +6981,59 @@ func (b *lxdBackend) createStorageStructure(path string) error {
...
+	// A lock isn't acquired in this case for all the instances as this might cause too much interruption
+	// as every update of an instance's backup file requires mounting the corresponding volume.
+	for _, inst := range instances {
+		instanceBackupConf, err := b.GenerateInstanceCustomVolumeBackupConfig(inst, backupVolConfCache, snapshots, op)
+		if err != nil {
+			return err
+		}
+
+		// Don't invoke the the update of the instance's backup file directly on *lxdBackend.
+		// The instance's root vol might be located on a different storage pool which gets loaded when calling the instance's UpdateBackupFile.
```
tomponline: Good note.

lxd/storage/backend_lxd.go
tomponline, line 1: Looks like this needs rebasing now as its still in this PR.

tomponline (changes requested):
Can the "Update instance backup file if custom volume..." commits at the end of the PR be split off from this one?

lxd/storage/backend_lxd.go
roosterfish: What do you think about having a `inst.ExpandedDevices().FilterDisks(someFilterFunc, anotherFilterFunc)`? And the filters accept a `Device` (or `map[string]string`) as arg.

To prevent import loops we cannot really import any useful utils from `devices/config` like the `instancetype.IsRootDiskDevice` I have used here.

roosterfish: I'll send a proposal in another PR which modifies the already existing uses cases.

tomponline: > What do you think about having a `inst.ExpandedDevices().FilterDisks(someFilterFunc, anotherFilterFunc)`? And the filters accept a `Device` (or `map[string]string`) as arg.
> 
> To prevent import loops we cannot really import any useful utils from `devices/config` like the `instancetype.IsRootDiskDevice` I have used here.

It depends if the filters are predefined functions, as if you have to define them in every call site it doesnt give us much benefit compared to filtering them in the loop as we do today.

If the functions are predefined, then maybe we can support chaining instead, e.g.

```
instance.ExpandedDevices().Disks().DiskPool("foo")
```

roosterfish: Added a proposal here using pre-defined filter funcs https://github.com/canonical/lxd/pull/15859.

lxd/backup/config/backup_config.go
roosterfish: Fixed.

lxd/storage/backend_lxd.go

lxd/storage/backend_lxd.go

lxd/storage/backend_lxd.go

lxd/backup.go
roosterfish: Fixed.

lxd/storage/backend_lxd.go
roosterfish: I have continued with what you also suggested on another place, moving the logic out to the caller. You now have to pass instances and I have kept the function name as is using `UpdateCustomVolumeBackupFile`.

The benefit of using this name over the `..Using..` is that it follows the exact same approach as for instances where we have `GenerateInstanceBackupConfig` and `UpdateInstanceBackupFile`.

The caller doesn't need in depth knowledge of how the custom volume's metadata is persisted (inside the instance's backup file which use the custom vol) but can use the same approach as for instances which is:

* One func to generate the config (`GenerateInstanceBackupConfig` or `GenerateCustomVolumeBackupConfig`)
* One func to persist the config (`UpdateInstanceBackupFile` or `UpdateCustomVolumeBackupFile`)

lxd/storage/backend_lxd.go

lxd/storage/utils.go
roosterfish: I tried moving it around but left it at its current place to prevent import loops. But I have moved it into a dedicated file to separate it from the generic utils.

Ultimately it should only be used by the backend funcs and also doesn't need to be exported (which would be required if its placed in another package).

lxd/storage/utils.go
roosterfish: Fixed, see other comment.

lxd/storage/utils.go
roosterfish: Good catch! I have added a comment to explain the dimensions of the cache.

lxd/storage/backend_lxd.go
roosterfish: Fixed. Moved the `new...` func away from the backend. But I have kept it unexported as its only used in this package.

lxd/storage/backend_lxd.go

lxd/storage/backend_lxd.go
roosterfish: Fixed.

lxd/storage/backend_lxd.go
roosterfish: Fixed using the new filter mechanism.

lxd/storage/backend_lxd.go
roosterfish: Yes, this is exactly what is happening. As I didn't account for this at the beginning, I have added another test case that explicitly tests that the additions from this PR won't break LXD's existing behavior which allows the restore but prevents the start (due to not anymore existing custom vols).

See the [test case](https://github.com/canonical/lxd/pull/15523/commits/1b761bb25a6489f219d13a8c26fa02b9ba26a966).

lxd/storage/backend_lxd.go
roosterfish: The pattern is only used here. The initial pool is just added to the struct when initializing the backup config.
I think in this case we don't need to put the logic into another func.

lxd/storage/pool_interface.go
roosterfish: Fixed by https://github.com/canonical/lxd/pull/15523#discussion_r2123849311.

lxd/instance/drivers/driver_lxc.go
roosterfish: > Why do we need to send custom volume config during migrations

We don't. When moving the logic to create the backup config for custom vols from inside `GenerateInstanceBackupConfig` to the caller I falsely added the call to `GenerateInstanceCustomVolumeBackupConfig` here even though it's not required and we can just pass `nil`.

As far as I can see the instance backup config is used solely on the sender and not passed to the receiving side.

After looking more carefully how the struct is passed during migration I think it was "easy" to use the backup config struct back then to pass around some of the information even though a struct with less information could have been used too.

It's fixed by removing the lookups of the custom vol backup config.

lxd/storage/backend_lxd.go
roosterfish: I think this can be closed now. Not sure what it refers to.

roosterfish:
> Can the "Update instance backup file if custom volume..." commits at the end of the PR be split off from this one?

Those are the actual commits that trigger the update of the instance's backup config in case any of their custom vols gets updated.

roosterfish:
@tomponline ready for another pass.

lxd/storage/backend_lxd.go
```diff
@@ -7111,13 +7111,33 @@ func (b *lxdBackend) GenerateInstanceBackupConfig(inst instance.Instance, snapsh
 				})
 			}
 		}
+
+		if volBackupConf != nil {
+			// Append all custom volumes to the list containing the instance's root volume.
+			config.Volumes = append(config.Volumes, volBackupConf.Volumes...)
```
tomponline, line 7330: @roosterfish if the same volume is attached twice - i believe this can be done - would we expect this to be de-duped here or by the caller of this function. I ask because we are de-duping the pools in this function below.

lxd/storage/cache.go
```diff
@@ -0,0 +1,95 @@
...
+	backupConfig "github.com/canonical/lxd/lxd/backup/config"
+	"github.com/canonical/lxd/lxd/operations"
+	"github.com/canonical/lxd/lxd/state"
+)
+
+// backupConfigCache is used to cache pools and volumes during backup config creation.
+type backupConfigCache struct {
+	pools map[string]Pool
+	// The volume cache is using the project as its first dimension.
+	// By default all projects use features.storage.volumes=true which uses the volumes from the default project.
```
tomponline: this statement is wrong?

lxd/storage/cache.go
```diff
@@ -0,0 +1,95 @@
...
+	// In this case the top level dimension only ever has one entry (default project) which causes the cache to stay small.
+	// Having the pool as the top level dimension would cause a default project entry under every pool from which volumes are consumed.
+	volumes map[string]map[string]map[string]*backupConfig.Volume
```
tomponline, line 26: Would be worth explaining what the dimensions are, in order, of this map.

lxd/storage/cache.go
```diff
@@ -0,0 +1,95 @@
...
+
+		// Cache the pool.
+		b.pools[name] = pool
+	}
+
+	return b.pools[name], nil
+}
+
+// getVolume returns the volume's backup config either by loading it from the DB or from the cache (preferred).
+func (b *backupConfigCache) getVolume(projectName string, poolName string, volName string, snapshots bool, op *operations.Operation) (*backupConfig.Volume, error) {
```
tomponline, line 64: Can you update the docblock to explain what `snapshots` arg does, or rename it to better explain what it is for?

lxd/storage/cache.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		volConfig, err := pool.GenerateCustomVolumeBackupConfig(projectName, volName, snapshots, op)
+		if err != nil {
+			// When restoring an instance from snapshot, some of the custom vols which were attached
+			// whilst taking the snapshot might not exist anymore.
+			// To not cause failures during restore, skip those volumes.
+			// These volumes are still listed in the list of instance devices and will cause an error when
+			// trying to start the instance.
+			return nil, fmt.Errorf("Failed generating backup config for volume %q in project %q: %w", volName, projectName, err)
```
tomponline, line 86: You dont appear to be skipping the volume here, but still returning an error?

lxd/storage/cache.go
tomponline: Maybe `includeSnapshots`?

lxd/storage/cache.go
tomponline: I'm not entirely following the comment, but I wondered:

1. Should we be tying this condition somehow to the `snapshots` arg?
2. Should we be updating `GenerateCustomVolumeBackupConfig` to return a specific api.StatusError we can detect so we dont accidentally mask other errors from it?

lxd/storage/backend_lxd.go
```diff
@@ -8120,3 +8121,57 @@ func (b *lxdBackend) getParentVolumeUUID(vol drivers.Volume, projectName string)
 
 	return parentUUID, nil
 }
+
+// GenerateInstanceCustomVolumeBackupConfig returns the backup config entry for this instance's custom volumes.
+func (b *lxdBackend) GenerateInstanceCustomVolumeBackupConfig(inst instance.Instance, cache *backupConfigCache, snapshots bool, op *operations.Operation) (*backupConfig.Config, error) {
```
tomponline, line 8344: @roosterfish can we include part of your excellent commit message in the docblock here explaining the purpose of the `cache` argument? This will make the reasoning more accessible to future readers.

lxd/storage/backend_lxd.go
tomponline: Thanks!

lxd/storage/backend_lxd.go
```diff
@@ -8120,3 +8121,57 @@ func (b *lxdBackend) getParentVolumeUUID(vol drivers.Volume, projectName string)
...
+			// cannot include it in the instance's backup config.
+			// The instance cannot be started in any case until the non-existing disk device gets removed.
+			if api.StatusErrorCheck(err, http.StatusNotFound) {
+				continue
+			}
+
+			return nil, err
+		}
+
+		instanceBackupConf.Volumes = append(instanceBackupConf.Volumes, vol)
```
tomponline: Do we need to de-dupe if the same volume is attached multiple times via 2 different disk devices?

lxd/storage/cache.go
roosterfish: I see, `features.storage.volumes=true` is right but the text and the ordering of the dimensions isn't. Thanks!

I just wondered probably in a prod setup someone has a big pool and x projects which means having the pool as the first dimension should be the way to go.
Not sure what was my thinking when implementing this. Probably because I already started with the wrong assumption on the intention of `features.storage.volumes=true` and then decided for the opposite in all cases.

lxd/storage/backend_lxd.go
```diff
@@ -6997,6 +6997,45 @@ func (b *lxdBackend) createStorageStructure(path string) error {
 	return nil
 }
 
+// UpdateCustomVolumeBackupFile writes the custom volume's config to the backup.yaml file of the corresponding instances.
+func (b *lxdBackend) UpdateCustomVolumeBackupFile(projectName string, volName string, snapshots bool, instances []instance.Instance, op *operations.Operation) error {
```
tomponline: Would you mind adding an "s" to the end of this function name, as it still irks me that we are implying there is a custom volume backup file, when this is actually a multiplier operation were we are updating potentially many files.

lxd/storage/backend_lxd.go
```diff
@@ -6997,6 +6997,45 @@ func (b *lxdBackend) createStorageStructure(path string) error {
...
+		}
+
+		pool, err := backupVolConfCache.getPool(poolName)
+		if err != nil {
+			return err
+		}
+
+		err = pool.UpdateInstanceBackupFile(inst, snapshots, instanceVolBackupConf, backupConfig.DefaultMetadataVersion, op)
+		if err != nil {
+			logger.Error("Failed updating backup file", logger.Ctx{"instance": inst.Name(), "err": err})
```
tomponline: why do we continue here, but not on the earlier errors? I feel this would benefit from a comment explaining the logic.

tomponline: message is missing `project` too.

lxd/storage/backend_lxd.go
roosterfish: Mh good point, wasn't aware that this is allowed.

For containers no problem, but for VMs it actually provides an error so looks like a bug `Error: Failed to start device "foo": Failed to add the virtiofs device: PCI: slot 0 function 0 not available for vhost-user-fs-pci, in use by vhost-user-fs-pci,id=dev-lxd_foo2-virtio-fs
`.

This test was using FS vols.

roosterfish: No issues with block devs though.

lxd/storage/backend_lxd.go
```diff
@@ -6300,9 +6280,37 @@ func (b *lxdBackend) UpdateCustomVolume(projectName string, volName string, newD
 		delete(newConfig, "volatile.idmap.next")
 	}
 
+	var instances []instance.Instance
```
tomponline: Can we move this new section up above the call to `b.UpdateVolume()` so we've done all the validation checks first before doing on-disk volume mutation at the driver level?

lxd/storage/backend_lxd.go
```diff
@@ -6320,6 +6328,12 @@ func (b *lxdBackend) UpdateCustomVolume(projectName string, volName string, newD
 		})
 	}
 
+	// Update the instance's backup files.
```
tomponline, line 6335: You even used "s" in the comment here.

Thats a +1 for `UpdateCustomVolumeBackupFiles` right? ;)

tomponline (changes requested):
I think we are nearly there with this PR, thanks!

A few questions, comment suggestions mainly.

lxd/storage/cache.go
roosterfish: Fixed. Added a JSON like representation to better understand the structure.

lxd/storage/cache.go
roosterfish: I specifically picked `snapshot` as this is the same as in the various backend functions when it comes to including snaps.
But I have added another doc line to explain the behavior.

lxd/storage/cache.go
roosterfish: I think the confusion came from still having the doc string here. The `VolumeDBGet` func already returns a 404 which is already checked here https://github.com/canonical/lxd/pull/15523/commits/1b5bac66cd3d1529935b8c69ae610de90db5247b#diff-029fdd0e1afdacb0e66630c0d7e813772122db9a7d036060902838db315d9c4dR8149-R8159. Also there is almost the same comment so I merged the best of both together at this place.

I have removed the comment from `getVolume()` as it's unrelated here but the caller can decide how to move on if a 404 is encountered.

lxd/storage/backend_lxd.go
roosterfish: Fixed.

lxd/storage/backend_lxd.go
roosterfish: There are essentially two places where having this de-dupe logic makes sense. 
One is in `GenerateInstanceBackupConfig` and the other is in `GenerateInstanceCustomVolumeBackupConfig`. 
I have decided for the latter as in there we should apply the logic that the same volume can be attached multiple times to the same instance but is only included once in the final backup config.
When running `GenerateInstanceBackupConfig` and passing the custom volume's backup config you shouldn't anymore care about filtering out duplicate vols. 

Fixed by adding a check for `GenerateInstanceCustomVolumeBackupConfig`.

lxd/storage/backend_lxd.go
roosterfish: Sure, think this strikes a good balance between keeping the func names consistent but actually representing what is happening. Fixed.

lxd/storage/backend_lxd.go
roosterfish: I have moved down the doc block right above this function to make it clear it's related to this function call.

lxd/storage/backend_lxd.go
roosterfish: Good point, fixed.

lxd/storage/backend_lxd.go
roosterfish: Not against, makes sense to me :) Fixed, see comment above.

lxd/storage/backend_lxd.go
tomponline: this still needs de-duping right?

roosterfish:
Added one more commit to check explicitly the behavior of not having duplicate vols in the backup config in case the same vol is attached multiple times using different devices to the same instance.

lxd/storage/cache.go
```diff
@@ -0,0 +1,99 @@
...
+	_, ok = b.volumes[poolName][projectName][volName]
+	if !ok {
+		pool, err := b.getPool(poolName)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to retrieve pool of volume %q in pool %q: %w", volName, poolName, err)
+		}
+
+		volConfig, err := pool.GenerateCustomVolumeBackupConfig(projectName, volName, snapshots, op)
+		if err != nil {
+			return nil, fmt.Errorf("Failed generating backup config for volume %q in project %q: %w", volName, projectName, err)
```
tomponline, line 86: worth including pool name in the message? Just a nit.

lxd/storage/backend_lxd.go
roosterfish: The input `volBackupConf` is already de-duped. Left a comment here https://github.com/canonical/lxd/pull/15523#discussion_r2182356918 to explain it and just added another test case for it.

lxd/storage/backend_lxd.go
```diff
@@ -8123,3 +8124,74 @@ func (b *lxdBackend) getParentVolumeUUID(vol drivers.Volume, projectName string)
...
+	for _, device := range inst.ExpandedDevices().Filter(filters.IsCustomVolumeDisk) {
+		vol, err := cache.getVolume(projectName, device["pool"], device["source"], snapshots, op)
+		if err != nil {
+			// When restoring an instance from snapshot, some of the custom vols which were attached
+			// whilst taking the snapshot might not exist anymore.
+			// To not cause failures during restore, skip those volumes because we cannot include them in the instance's backup config.
+			// These volumes are still listed in the list of instance devices and will cause an error when
+			// trying to start the instance.
+			// The instance cannot be started in any case until the non-existing disk device gets removed.
+			if api.StatusErrorCheck(err, http.StatusNotFound) {
```
tomponline, line 8367: Maybe worth logging a warning here? Just a nit.

lxd/storage/cache.go
roosterfish: If you don't mind I'll update all invocations of `GenerateCustomVolumeBackupConfig` in a follow up PR instead to add more context to the error message.

tomponline: go for it!

lxd/storage/backend_lxd.go
tomponline: got you, thanks.

lxd/storage/backend_lxd.go
tomponline: got you. thanks

lxd/storage/backend_lxd.go
roosterfish: Fixed, added:

`logger.Warn("Missing custom volume to include in instance's backup config", logger.Ctx{"project": projectName, "pool": device["pool"], "volume": device["source"], "instance": inst.Name()})`

lxd/storage/backend_lxd.go
```diff
@@ -6997,6 +6997,45 @@ func (b *lxdBackend) createStorageStructure(path string) error {
 	return nil
 }
 
+// UpdateCustomVolumeBackupFiles writes the custom volume's config to the backup.yaml file of the corresponding instances.
+func (b *lxdBackend) UpdateCustomVolumeBackupFiles(projectName string, volName string, snapshots bool, instances []instance.Instance, op *operations.Operation) error {
+	l := b.logger.AddContext(logger.Ctx{"project": projectName, "volume": volName, "snapshots": snapshots})
+	l.Debug("UpdateCustomVolumeBackupFile started")
```
tomponline: UpdateCustomVolumeBackupFiles

lxd/storage/backend_lxd.go
```diff
@@ -6997,6 +6997,45 @@ func (b *lxdBackend) createStorageStructure(path string) error {
...
+	defer l.Debug("UpdateCustomVolumeBackupFile finished")
```
tomponline: UpdateCustomVolumeBackupFiles

lxd/storage/backend_lxd.go
```diff
@@ -6997,6 +6997,45 @@ func (b *lxdBackend) createStorageStructure(path string) error {
...
+		}
+
+		// Try updating all of the instance's backup files using a best effort strategy.
+		// In case a custom volume is used by many instances, it might happen that an instance gets deleted
+		// whilst this function tries to propagate an update of a custom volume.
+		// A lock isn't acquired in this case for all the instances as this might cause too much interruption
+		// as every update of an instance's backup file requires mounting the corresponding volume.
+		err = pool.UpdateInstanceBackupFile(inst, snapshots, instanceVolBackupConf, backupConfig.DefaultMetadataVersion, op)
+		if err != nil {
+			logger.Error("Failed updating backup file", logger.Ctx{"instance": inst.Name(), "project": inst.Project().Name, "err": err})
```
tomponline: we put "project" first by convention (in hierarchy order), just a nit.

lxd/storage/backend_lxd.go
roosterfish: Reminder to my future self to not forget running STRG+F because renaming the doc string isn't enough :) Ty!

lxd/storage/backend_lxd.go

lxd/storage/backend_lxd.go

lxd/storage/backend_lxd.go

lxd/storage/backend_lxd.go
```diff
@@ -6241,6 +6241,28 @@ func (b *lxdBackend) UpdateCustomVolume(projectName string, volName string, newD
 
 	// Apply config changes if there are any.
 	changedConfig, userOnly := b.detectChangedConfig(curVol.Config, newConfig)
+
+	var instances []instance.Instance
+	err = VolumeUsedByInstanceDevices(b.state, b.name, projectName, &curVol.StorageVolume, true, func(dbInst db.InstanceArgs, project api.Project, _ []string) error {
```
tomponline, line 6251: should we only do this work if:

```go
if len(changedConfig) != 0 || newDesc != curVol.Description {
``

? 

Like the DB record update for the custom volume at the bottom of this function, and the `len(changedConfig) != 0 ` check for the actual on-disk volume update.

File: lxd/storage/backend_lxd.go [hunk 42]
```diff
@@ -6855,6 +6856,34 @@ func (b *lxdBackend) RenameCustomVolumeSnapshot(projectName, volName string, new
 		_ = b.driver.RenameVolumeSnapshot(newVol, oldSnapshotName, op)
 	})
 
+	// Fetch the parent vol.
```
tomponline, line 6879: In UpdateCustomVolume and CreateCustomVolumeSnapshot the list of instances using the volume was done before the driver level on-disk work was done using the principle of getting the data needed first before doing any modifications.

But here the on-disk driver level rename has been done first before the instance list has been retrieved.

Was this intentional?

lxd/storage/backend_lxd.go
```diff
@@ -7005,6 +7004,34 @@ func (b *lxdBackend) DeleteCustomVolumeSnapshot(projectName, volName string, op
 		return err
 	}
 
+	var instances []instance.Instance
```
tomponline: can the list of instances using volume be retrieved before the on-disk volume snapshot is removed?

lxd/storage/backend_lxd.go
roosterfish: Hm, I stumbled across this too when moving the block earlier.
I initially had it inside the `if len(changedConfig) != 0 || newDesc != curVol.Description {` but moved it out as I had in mind with this endpoint you can both update config keys and the desc.
But I didn't check close enough to realize the `|| newDes != currentDesc`. 

To prevent any of this to happen in the future, should we rather return the `nil` right away when `len(changedConfig) == 0 && newDesc == curVol.Description`? Because then it's a noop and we just skip the remainder of the func.

lxd/storage/backend_lxd.go
roosterfish: The reason it's so early in `UpdateCustomVolume` is because I wanted to prevent running the `VolumeUsedByInstanceDevices` twice as it's already used in there for the config key check which happens prior to the driver/DB calls.

In `CreateCustomVolumeSnapshot` we require the list of instances to add the reverter which has to be added before doing the first DB modification (`VolumeDBCreate` and it's own reverter) to be able to restore the instance's backup file in reverse order after the `VolumeDBCreate` got reverted. `UpdateCustomVolumeBackupFiles` is accessing the DB and then expects the already reset state.

Here in `RenameCustomVolumeSnapshot` the first DB modification is happening later so therefore the list of instances can also be retrieved at a later point in time. That should be the same on the other modified funcs too.

lxd/storage/backend_lxd.go
roosterfish: We can do this yes. I think I see where you are heading. Let's try to fetch as much as possible upfront before we start talking with the (maybe) external storage. Or just prevent any revert operation as best as we can so that we only ever start touching the actual storage if everything on LXD end is ready. 

I'll go through the backend funcs and move them all up.

lxd/storage/backend_lxd.go
tomponline: I think we should try and retrieve as much as possible from DB before doing on-disk mods so that if there is anything wrong with DB loading there's nothing to revert on disk.

lxd/storage/backend_lxd.go
tomponline: yes that probably makes sense indeed.

roosterfish: Added another small PR to have it outside of this one https://github.com/canonical/lxd/pull/15927.

lxd/storage/backend_lxd.go
roosterfish: Fixed. Moved the ones that came after after the call to the storage driver.

lxd/storage/backend_lxd.go
roosterfish: Fixed, see https://github.com/canonical/lxd/pull/15523#discussion_r2182681003.

tomponline (approved):
LGTM thanks!

PR #15526, opened 2025-05-06
Storage: Fix nested transaction in `storagePoolVolumeUpdateUsers`

roosterfish:
In the transaction only collect the necessary information but don't spend time persisting the data. This can be done outside of the transaction for performance reasons and also allows fetching further information down the line.

This fixes the following warning message in the log when running `lxc storage volume rename`: 

`level=warning msg="Transaction timed out. Retrying once" err="Failed to begin transaction: context deadline exceeded" member=1`

### DIFF
--- a/lxd/storage_volumes_utils.go
+++ b/lxd/storage_volumes_utils.go
@@ -8,6 +8,7 @@ import (
 	"github.com/canonical/lxd/lxd/backup"
 	"github.com/canonical/lxd/lxd/db"
 	"github.com/canonical/lxd/lxd/db/cluster"
+	"github.com/canonical/lxd/lxd/device/config"
 	"github.com/canonical/lxd/lxd/instance"
 	"github.com/canonical/lxd/lxd/state"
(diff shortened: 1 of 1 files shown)


tomponline:
@roosterfish did the rename action failure (if so can we add a test for it), or do we have a test, and this was just a warning message that was the problem?

roosterfish:
> @roosterfish did the rename action failure (if so can we add a test for it), or do we have a test, and this was just a warning message that was the problem?

Sorry, accidentally edited your comment. 

The operation itself doesn't fail, it just takes a few seconds until it returns and logs the warning. With this fix it returns immediately.

roosterfish:
The `storage.sh` test does check the rename of vols attached to an instance, see https://github.com/canonical/lxd/blob/main/test/suites/storage.sh#L366. 

It only causes issues if the vol is attached to an instance.

tomponline:
> It only causes issues if the vol is attached to an instance.

should we add a test to catch regressions of this logic?

tomponline (approved):

PR #15527, opened 2025-05-06
Storage: Update instance backup file when root vol pool gets updated

roosterfish:
Fixes a general bug in LXD which caused an instance's backup file not to be updated when updating the storage pool of its root volume.

This was discovered whilst working on https://github.com/canonical/lxd/pull/15523.


tomponline (approved):
Thanks!

PR #15528, opened 2025-05-06
Storage: Fix snapshot expiry date test regression

roosterfish:
Applies the changes suggested in https://github.com/canonical/lxd/pull/15503#discussion_r2070417058 for both instances and custom vols. 

In addition the test suite is modified to not anymore strip off the microsecond portion of the timestamps.

Furthermore a bug is fixed which always set the zero timestamp for an instance's snapshot vol expiry. Before only the actual instance snapshot DB entry contained the expiry date. A small test for this is added too.

### DIFF
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -6582,7 +6582,7 @@ func (b *lxdBackend) ImportCustomVolume(projectName string, poolVol *backupConfi
 }
 
 // CreateCustomVolumeSnapshot creates a snapshot of a custom volume.
-func (b *lxdBackend) CreateCustomVolumeSnapshot(projectName, volName string, newSnapshotName string, newDescription string, newExpiryDate time.Time, op *operations.Operation) error {
+func (b *lxdBackend) CreateCustomVolumeSnapshot(projectName, volName string, newSnapshotName string, newDescription string, newExpiryDate *time.Time, op *operations.Operation) error {
 	l := b.logger.AddContext(logger.Ctx{"project": projectName, "volName": volName, "newSnapshotName": newSnapshotName, "newDescription": newDescription, "newExpiryDate": newExpiryDate})
 	l.Debug("CreateCustomVolumeSnapshot started")
 	defer l.Debug("CreateCustomVolumeSnapshot finished")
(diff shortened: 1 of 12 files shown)


lxd/instance/instance_utils.go
```diff
@@ -794,10 +794,6 @@ func CreateInternal(s *state.State, args db.InstanceArgs, clearLogDir bool) (Ins
 		checkedProfiles[profile.Name] = true
 	}
 
-	if args.CreationDate.IsZero() {
```
tomponline, line 797: Id like to keep this as a failsafe please

lxd/storage/backend_lxd.go
```diff
@@ -3649,7 +3649,7 @@ func (b *lxdBackend) CreateInstanceSnapshot(inst instance.Instance, src instance
 	defer unlock()
 
 	// Validate config and create database entry for new storage volume.
-	err = VolumeDBCreate(b, inst.Project().Name, inst.Name(), srcDBVol.Description, volType, true, vol.Config(), inst.CreationDate(), time.Time{}, contentType, false, true)
+	err = VolumeDBCreate(b, inst.Project().Name, inst.Name(), srcDBVol.Description, volType, true, vol.Config(), inst.CreationDate(), inst.ExpiryDate(), contentType, false, true)
```
tomponline, line 3652: i think this was on purpose, as the instance snapshot expires, which then also removes the instance volume snapshot.

roosterfish: I found it odd that `lxc config show c1/snap2` shows the expiry whilst `lxc storage volume show <pool> container/c1/snap2` shows the zero time even though both the snapshot and its vol are tightly coupled.
 
But do you think it would hurt putting the actual expiry?

tomponline: depends on how you think about it I guess, the instance snapshot expires taking with it the volume snapshot linked to it. The value in the volume doesn't govern when the snapshot expires, you couldn't edit it for example, and we'd have to make sure if the instance snapshot expiry was edited this was also updated to reflect in the volume snapshot. Additionally we'd need to make sure that the expiry in the volume snapshot isnt taken as a trigger to delete the volume without the instance snapshot being deleted.

tomponline: this feels like a separate GH issue independent of the main issue trying to be fixed in this PR though

roosterfish: I realized there already is a test that checks explicitly for the zero timestamp on the snapshot's vol because it now failed on this. 

Thanks for your thoughts, that all sounds reasonable and makes sense, I have reverted those bits.

tomponline (approved):
Thanks!

PR #15529, opened 2025-05-07
lxd: Separate API handlers from business logic

MusicDin:
This PR aims to separate the business logic from the handlers. 

The main requirement for this is to stop passing an actual request around, and pass request context instead. 

The second change addresses the types returned from the BL functions. Instead of returning the responses, we want to get concrete types and a potential error. This also affected forwarding response, which is forwards the request when the response is rendered. Now, we return a `ForwardRequiredError` instead, which is handled by `response.SmartError`.

Only a subset of handlers are addressed, and the remaining will be done once we agree upon this approach.


markylaing:
I've had a quick look over this out of curiosity and I have a few initial thoughts:
1. Since this is such a large scale refactor. I think it's worth standardising how requests are parsed. I have a few ideas for this.
2. I think it's worth formalising what a request context is and what it should contain. Currently we have a lot of `request.CtxKey` definitions, where the true nature of the request typically needs to be computed from a combination of them (e.g. `CtxUsername` vs. `CtxForwardedUsername` have a very important distinction when it comes to auth). This could be something like:
```golang
type RequestContext interface {
    context.Context
    ID() string // ID of request
    IsAuthenticated() bool
    AuthenticationMethod() string // Always returning the authentication method of the original requestor, even if forwarded
    Identifier() string // Always returning the identifier of the original requestor, even if forwarded
    IsForwarded() bool // Indicate that the request has been forwarded.
    ForwardedFrom() string
    /* etc. */
}
```
Then, when a request arrives, an implementation of this interface is instantiated and passed around as a `context.Context`. Callers can then check if it's an internal call via type assertion (e.g. `requestContext, ok := ctx.(RequestContext)`, `if !ok { /* internal */`).
4. It would be nice to have an interface for the business logic that looks similar to the client interface. This would separate handling of e.g. `GetInstances` and `GetInstancesFull`, meaning that implementations would be lots of smaller functions, rather than one giant handler that catches lots of cases.

MusicDin:
Thanks for taking a look at this 🚀 

We previously briefly spoken about that and I completely agree. 
Although, I would personally avoid passing `RequestContext` as `context.Context` when this is not entirely necessary. This way we avoid the need for type assertion, and we can pass `nil` to emphasize that there is no request context (and make the linter happy at the same time). 

Also we avoid confusion where functions accept multiple contexts `someFunc(ctx context.Context, reqContext context.Context, ...)`.

And this still leaves us the room to pass the request context as `context.Context` if really necessary (e.g. in helper function). In such case, the receiver can still use the type assertion as per your suggestion instead of detecting the request context by checking if a specific context key is set.

> 4. It would be nice to have an interface for the business logic that looks similar to the client interface. This would separate handling of e.g. GetInstances and GetInstancesFull, meaning that implementations would be lots of smaller functions, rather than one giant handler that catches lots of cases.

Could you elaborate on this one a bit? I'm a bit confused because those smaller functions generally accept different input arguments.

tomponline:
I'm not entirely convinced on the need for an interface (as opposed to a concrete struct), is there going to be multiple implementations? If not then a concrete struct should be sufficient to hold the functions/variables, rather than needing to mask it behind an interface. See https://medium.com/@rahulgorai/why-interfaces-can-be-expensive-in-go-and-when-you-should-still-use-them-a1510c162851

markylaing:
> Could you elaborate on this one a bit? I'm a bit confused because those smaller functions generally accept different input arguments, and we would introduce an interface that just maps our current implementation?

Currently we have handlers that accept various parameters and return different things based on path arguments and query parameters.

My meaning here is the business logic should never return e.g. `[]any` to represent either `[]api.Instance` or `[]api.InstanceFull` based an a recursion parameter. Instead, the business logic should have something like:
```go
GetInstanceNames(ctx context.Context, projectName string) ([]string, error)
GetInstances(ctx context.Context, projectName string) ([]api.Instance, error)
GetInstancesFull(ctx context.Context, projectName string) ([]api.InstanceFull, error)
GetInstanceNamesAllProjects(ctx context.Context) ([]string, error)
GetInstancesAllProjects(ctx context.Context) ([]api.Instance, error)
GetInstanceFullAllProjects(ctx context.Context) ([]api.InstanceFull, error)
```

Depending on the initial request parsing, the handler will call out to one of these functions. These functions would be almost identical to those defined in `client/interfaces.go` but with have a context parameter.

> I'm not entirely convinced on the need for an interface (as opposed to a concrete struct), is there going to be multiple implementations? If not then a concrete struct should be sufficient to hold the functions/variables, rather than needing to mask it behind an interface. See https://medium.com/@rahulgorai/why-interfaces-can-be-expensive-in-go-and-when-you-should-still-use-them-a1510c162851

I agree, there's no reason for the interface. Instead we could just have a struct for this that is set in the request context. If this struct is not present then it is an internal request. This would just bundle the various existing context keys together to make them easier to use.

MusicDin:
@markylaing @tomponline Removed handler refactoring logic and just left the bits that pass around the request context instead of the http.Request. The request info is coming in a followup PR.

lxd/devlxd.go
```diff
@@ -453,6 +453,9 @@ func registerDevLXDEndpoint(d *Daemon, apiRouter *mux.Router, apiVersion string,
 		// Set devLXD auth method to identify this request as coming from the /dev/lxd socket.
 		request.SetCtxValue(r, request.CtxProtocol, auth.AuthenticationMethodDevLXD)
 
+		// Set request source address value to the request context.
+		request.SetCtxValue(r, request.CtxRequestSourceAddress, r.RemoteAddr)
```
tomponline: what is this value for devlxd clients?

lxd/cluster/connect.go
```diff
@@ -24,10 +24,13 @@ import (
...
 // with the correct parameters for node-to-node communication.
 //
+// If a context is passed, the proxy is configured as well. The context must be a request
+// context, as it is used to extract the various information from the request.
+//
 // If 'notify' switch is true, then the user agent will be set to the special
 // to the UserAgentNotifier value, which can be used in some cases to distinguish
 // between a regular client request and an internal cluster request.
-func Connect(address string, networkCert *shared.CertInfo, serverCert *shared.CertInfo, r *http.Request, notify bool) (lxd.InstanceServer, error) {
+func Connect(reqContext context.Context, address string, networkCert *shared.CertInfo, serverCert *shared.CertInfo, notify bool) (lxd.InstanceServer, error) {
```
tomponline: for future reference please introduce function signature changes as a separate commit first, followed by usage of it in the next commit, otherwise the reviewing has to search through reams of usage changes to find the actual function change.

lxd/cluster/connect.go
```diff
@@ -24,10 +24,13 @@ import (
 // Connect is a convenience around lxd.ConnectLXD that configures the client
 // with the correct parameters for node-to-node communication.
 //
+// If a context is passed, the proxy is configured as well. The context must be a request
```
tomponline: this doesnt read how i think you intended it, because you have to pass in a context.

How about:

> If a request context is passed (as defined by request.IsRequestContext) then the proxy is configured as well.

lxd/cluster/connect.go
```diff
@@ -50,23 +53,22 @@ func Connect(address string, networkCert *shared.CertInfo, serverCert *shared.Ce
 		args.UserAgent = clusterRequest.UserAgentNotifier
 	}
 
-	if r != nil {
+	if request.IsRequestContext(reqContext) {
 		proxy := func(req *http.Request) (*url.URL, error) {
```
tomponline, line 58: BTW @MusicDin @markylaing I find this entire proxy logic inside `cluster.Connect()` a bit suspect.

Why would a proxy need to be used for intra-cluster communication? And is this same proxy logic used for dqlite setup?

tomponline: I think this `proxy` logic is also used as a way to modify the request to send the request username and other auth details across to the cluster member. Not really to do with proxying.

lxd/cluster/connect.go
tomponline: Based on https://github.com/canonical/lxd/pull/15529/commits/a6893228e17e6c383a947a616e0d9e7057f271df#r2123359562

How about:

> If a request context is passed (as defined by request.IsRequestContext) then the identity info from the context is sent in the request, as well as respective proxy environment variables.

lxd/request/request.go
```diff
@@ -3,14 +3,12 @@ package request
...
-	"net/http"
 
 	"github.com/canonical/lxd/shared/api"
 )
 
-// CreateRequestor extracts the lifecycle event requestor data from an http.Request context.
-func CreateRequestor(r *http.Request) *api.EventLifecycleRequestor {
-	ctx := r.Context()
+// CreateRequestor extracts the lifecycle event requestor data from the request context.
+func CreateRequestor(ctx context.Context) *api.EventLifecycleRequestor {
```
tomponline, line 11: like the earlier commit, i had to scroll a long way to find this, please separate into an earlier commit.

lxd/instance_post.go
```diff
@@ -211,7 +211,7 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
 
 			var targetGroupName string
 
-			targetMemberInfo, targetGroupName, err = limits.CheckTarget(ctx, s.Authorizer, r, tx, targetProject, target, allMembers)
+			targetMemberInfo, targetGroupName, err = limits.CheckTarget(ctx, r.Context(), s.Authorizer, tx, targetProject, target, allMembers)
```
tomponline: this looks suspect, passing the same context in twice as 2 separate arguments?

lxd/project/limits/permissions.go
```diff
@@ -1585,10 +1585,10 @@ func projectHasRestriction(project *api.Project, restrictionKey string, blockVal
 }
 
 // CheckClusterTargetRestriction check if user is allowed to use cluster member targeting.
-func CheckClusterTargetRestriction(authorizer auth.Authorizer, r *http.Request, project *api.Project, targetFlag string) error {
+func CheckClusterTargetRestriction(reqContext context.Context, authorizer auth.Authorizer, project *api.Project, targetFlag string) error {
```
tomponline: should be separate commit

lxd/project/limits/permissions.go
```diff
@@ -1713,11 +1713,11 @@ func CheckTargetGroup(ctx context.Context, tx *db.ClusterTx, p *api.Project, gro
 // If target is a cluster member and is found in allMembers it returns the resolved node information object.
 // If target is a cluster group it returns the cluster group name.
 // In case of error, neither node information nor cluster group name gets returned.
-func CheckTarget(ctx context.Context, authorizer auth.Authorizer, r *http.Request, tx *db.ClusterTx, p *api.Project, target string, allMembers []db.NodeInfo) (*db.NodeInfo, string, error) {
+func CheckTarget(ctx context.Context, reqContext context.Context, authorizer auth.Authorizer, tx *db.ClusterTx, p *api.Project, target string, allMembers []db.NodeInfo) (*db.NodeInfo, string, error) {
```
tomponline: separate commit please

tomponline: please can you explain the need for 2 context arguments, this looks pretty suspect to me.

lxd/cluster/connect.go
```diff
@@ -92,14 +92,14 @@ func Connect(reqContext context.Context, address string, networkCert *shared.Cer
 // ConnectIfInstanceIsRemote figures out the address of the cluster member which is running the instance with the
 // given name in the specified project. If it's not the local member will connect to it and return the connected
 // client (configured with the specified project), otherwise it will just return nil.
-func ConnectIfInstanceIsRemote(s *state.State, projectName string, instName string, r *http.Request, instanceType instancetype.Type) (lxd.InstanceServer, error) {
+func ConnectIfInstanceIsRemote(reqContext context.Context, s *state.State, projectName string, instName string, instanceType instancetype.Type) (lxd.InstanceServer, error) {
```
tomponline: separate commit please

lxd/api_cluster.go
```diff
@@ -2580,7 +2580,7 @@ func internalClusterPostRebalance(d *Daemon, r *http.Request) response.Response
 
 // Check if there's a dqlite node whose role should be changed, and post a
 // change role request if so.
-func rebalanceMemberRoles(s *state.State, gateway *cluster.Gateway, r *http.Request, unavailableMembers []string) error {
+func rebalanceMemberRoles(reqContext context.Context, s *state.State, gateway *cluster.Gateway, unavailableMembers []string) error {
```
tomponline: separate commit please

lxd/daemon.go
```diff
@@ -2544,7 +2544,7 @@ func (d *Daemon) nodeRefreshTask(heartbeatData *cluster.APIHeartbeat, isLeader b
 		if isDegraded || onlineVoters < maxVoters || onlineStandbys < maxStandBy {
 			d.clusterMembershipMutex.Lock()
 			logger.Debug("Rebalancing member roles in heartbeat", logger.Ctx{"local": localClusterAddress})
-			err := rebalanceMemberRoles(d.State(), d.gateway, nil, unavailableMembers)
+			err := rebalanceMemberRoles(context.TODO(), d.State(), d.gateway, unavailableMembers)
```
tomponline: why TODO() rather than Background() like the others?

lxd/api_cluster.go
```diff
@@ -2651,7 +2651,7 @@ func upgradeNodesWithoutRaftRole(s *state.State, gateway *cluster.Gateway) error
 
 // Post a change role request to the member with the given address. The nodes
 // slice contains details about all members, including the one being changed.
-func changeMemberRole(s *state.State, r *http.Request, address string, nodes []db.RaftNode) error {
+func changeMemberRole(reqContext context.Context, s *state.State, address string, nodes []db.RaftNode) error {
```
tomponline: separate commit please

lxd/daemon_images.go
```diff
@@ -60,7 +60,7 @@ func imageOperationLock(fingerprint string) (locking.UnlockFunc, error) {
 }
 
 // ImageDownload resolves the image fingerprint and if not in the database, downloads it.
-func ImageDownload(r *http.Request, s *state.State, op *operations.Operation, args *ImageDownloadArgs) (*api.Image, error) {
+func ImageDownload(reqContext context.Context, s *state.State, op *operations.Operation, args *ImageDownloadArgs) (*api.Image, error) {
```
tomponline: separate commit please

lxd/instances_post.go
```diff
@@ -44,7 +44,7 @@ import (
 	"github.com/canonical/lxd/shared/version"
 )
 
-func ensureDownloadedImageFitWithinBudget(s *state.State, r *http.Request, op *operations.Operation, p api.Project, imgAlias string, source api.InstanceSource, imgType string) (*api.Image, error) {
+func ensureDownloadedImageFitWithinBudget(reqContext context.Context, s *state.State, op *operations.Operation, p api.Project, imgAlias string, source api.InstanceSource, imgType string) (*api.Image, error) {
```
tomponline: can this just be ctx?

lxd/instances_post.go
```diff
@@ -83,7 +83,7 @@ func ensureDownloadedImageFitWithinBudget(s *state.State, r *http.Request, op *o
 	return imgDownloaded, nil
 }
 
-func createFromImage(s *state.State, r *http.Request, p api.Project, profiles []api.Profile, img *api.Image, imgAlias string, req *api.InstancesPost) response.Response {
+func createFromImage(reqContext context.Context, s *state.State, p api.Project, profiles []api.Profile, img *api.Image, imgAlias string, req *api.InstancesPost) response.Response {
```
tomponline: can this just be ctx?

tomponline (changes requested):
Hi @MusicDin I didnt get all the way through this, but in general we need to:

1. Split the function signature changes into their own commits before the updated usage commit(s).
2. Revisit and justify why some functions need to accept 2 context args, which in at least some cases, are the same context. This looks problematic.
3. Why do some functions accept a ctx arg and others accept a reqContext arg, can we use consistent naming and just use `ctx` for all?

lxd/cluster/connect.go
markylaing: Yeah it's not really a proxy, it just sets details about the original request to be forwarded to the other cluster members. We could rename to something more suitable

MusicDin:
> Split the function signature changes into their own commits before the updated usage commit(s).

👍 

> Revisit and justify why some functions need to accept 2 context args, which in at least some cases, are the same context. This looks problematic.

This is because previously we were passing the context and request. Since we are interested into request's context only for the context values, the context is passed in case it contains the timeouts. I will revisit and check whether we can use just request's context.

> Why do some functions accept a ctx arg and others accept a reqContext arg, can we use consistent naming and just use ctx for all?

Sure, we can. I just wanted to emphasize that we are expecting request context which has the desired context values set (username, protocol, ...).

lxd/cluster/connect.go
MusicDin: This is cleaner. Thanks!

lxd/instance_post.go
MusicDin: Fixed

lxd/project/limits/permissions.go
MusicDin: There is no need. Fixed.

lxd/instances_post.go
MusicDin: Yes. Fixed

lxd/daemon.go
MusicDin: Fixed

lxd/devlxd.go
MusicDin: This request source is used to reliably detect whether it is a request context (even if value is empty, it will be present in each request context). I've removed it, as there is currently no need for that on the devLXD part, and this in particular will be replaced by checking if the request context contains `RequestInfo`.

tomponline:
@MusicDin please rebase

Issue #15480, opened 2025-04-29
Empty snapshot list on clustered directory pool


edlerd:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

ubuntu

### Distribution version

5.21

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
root@micro1:~# snap list --all lxd core20 core22 core24 snapd
Name    Version         Rev    Tracking       Publisher   Notes
core20  20241206        2496   latest/stable  canonical✓  base,disabled
core20  20250213        2501   latest/stable  canonical✓  base
core22  20250210        1802   latest/stable  canonical✓  base,disabled
core22  20250315        1908   latest/stable  canonical✓  base
core24  20241217        739    latest/stable  canonical✓  base,disabled
core24  20250318        888    latest/stable  canonical✓  base
lxd     5.21.3-c5ae129  33110  5.21/edge      canonical✓  disabled,in-cohort
lxd     git-54945ae     33432  5.21/edge      canonical✓  in-cohort
snapd   2.67.1          23771  latest/stable  canonical✓  snapd,disabled
snapd   2.68.4          24505  latest/stable  canonical✓  snapd
```

### Output of "lxc info" or system info if it fails

```shell
-
```

### Issue description

Given a directory based pool on a cluster:
```
root@micro1:~# lxc storage show dirydir
name: dirydir
description: ""
driver: dir
status: Created
config: {}
used_by:
- /1.0/storage-pools/dirydir/volumes/custom/asd-copy-cli?target=micro1
- /1.0/storage-pools/dirydir/volumes/custom/asd-copy-cli?target=micro2
- /1.0/storage-pools/dirydir/volumes/custom/asd/snapshots/snap0?target=micro2
- /1.0/storage-pools/dirydir/volumes/custom/asd/snapshots/snap1?target=micro2
- /1.0/storage-pools/dirydir/volumes/custom/asd/snapshots/snap2?target=micro2
- /1.0/storage-pools/dirydir/volumes/custom/asd/snapshots/snap3?target=micro2
- /1.0/storage-pools/dirydir/volumes/custom/asd/snapshots/snap4?target=micro2
- /1.0/storage-pools/dirydir/volumes/custom/asd/snapshots/snap5?target=micro2
- /1.0/storage-pools/dirydir/volumes/custom/asd/snapshots/snap6?target=micro2
[...]

PR #15536, opened 2025-05-07
Cluster: Add cluster target response forwarding to storage volume snapshot `GET` handler

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/15480.


tomponline (approved):
Thanks!

Issue #14256, opened 2024-10-10
vm import fails with no message if the size of the vm is bigger


mindtheegab:
# Required information

 * Distribution: ubuntu 
 * Distribution version: 24.04.1
 * The output of "snap list --all lxd core20 core22 core24 snapd":
```
snap list --all lxd core20 core22 core24 snapd
Name    Version         Rev    Tracking       Publisher   Notes
core22  20240904        1621   latest/stable  canonical✓  base
lxd     5.21.2-2f4ba6b  30131  5.21/stable    canonical✓  -
snapd   2.65.3          22991  latest/stable  canonical✓  snapd
```

```
lxc info
config:
  core.https_address: :8443
api_extensions:
- storage_zfs_remove_snapshots
- container_host_shutdown_timeout
- container_stop_priority
- container_syscall_filtering
- auth_pki
- container_last_used_at
- etag
- patch
- usb_devices
- https_allowed_credentials
- image_compression_algorithm
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
- unix_device_rename
- storage_lvm_use_thinpool
- storage_rsync_bwlimit
- network_vxlan_interface
- storage_btrfs_mount_options
- entity_description
- image_force_refresh
- storage_lvm_lv_resizing
- id_map_base
- file_symlinks
- container_push_target
- network_vlan_physical
- storage_images_delete
- container_edit_metadata
- container_snapshot_stateful_migration
- storage_driver_ceph
- storage_ceph_user_name
- resource_limits
- storage_volatile_initial_source
- storage_ceph_force_osd_reuse
- storage_block_filesystem_btrfs
- resources
- kernel_limits
- storage_api_volume_rename
- network_sriov
- console
- restrict_devlxd
- migration_pre_copy
- infiniband
- maas_network
- devlxd_events
- proxy
- network_dhcp_gateway
- file_get_symlink
- network_leases
- unix_device_hotplug
[...]

PR #15537, opened 2025-05-07
Improve migration API error handling (from Incus)

kadinsayani:
Closes https://github.com/canonical/lxd/issues/14256.

This improves error handling in `CreateInstanceFromMigration` by validating instance root disk size against the migration source volume size.

Cherry-picked from https://github.com/lxc/incus/pull/1971 and https://github.com/lxc/incus/pull/2041.

### DIFF
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -2234,6 +2234,26 @@ func (b *lxdBackend) CreateInstanceFromMigration(inst instance.Instance, conn io
 		return err
 	}
 
+	// Now that we got the source details, validate against the instance limits.
+	_, rootDiskConf, err := instancetype.GetRootDiskDevice(inst.ExpandedDevices().CloneNative())
+	if err != nil {
+		return err
+	}
+
+	if rootDiskConf["size"] != "" {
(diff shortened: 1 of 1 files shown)


simondeziel (dismissed):

MusicDin (commented):
What happens if the source block disk size is larger than volume size, but the user does not explicitly configure the root disk size? 

In such case the volume should be resized to the source disk size (source disk size hint is passed via a header). Will the root disk size be unset in such case - effectively ignoring the check?

The volume size adjustment can be found here:
https://github.com/canonical/lxd/blob/a24a27d7878f4e35febafa4898af2ab9d984f524/lxd/storage/backend_lxd.go#L2412-L2420

kadinsayani:
> What happens if the source block disk size is larger than volume size, but the user does not explicitly configure the root disk size?
> 
> In such case the volume should be resized to the source disk size (source disk size hint is passed via a header). Will the root disk size be unset in such case - effectively ignoring the check?

Yes, the check will be ignored and the volume will be resized to the source disk size.

MusicDin:
> Yes, the check will be ignored and the volume will be resized to the source disk size.

Cool stuff. Thanks! 🚀

tomponline (approved):
Ta

PR #15541, opened 2025-05-07
Build `liblxc` as part of `make deps`

simondeziel:


simondeziel:
> Does https://discourse.ubuntu.com/t/building-custom-lxd-binaries-for-side-loading-into-an-existing-snap-installation/37327 need updating too?

Yes, but it seems that page hasn't been updated in a long time as it still mentioned non-unified `dqlite` builds (where `raft` was a standalone lib).

Should I fill a Jira card for someone (Minea? me?) to update?

PR #15542, opened 2025-05-08
Storage: Fix CephFS mount timeout

roosterfish:
When LXD starts up, it tries to mount all storage pools.
In case of CephFS, if the storage cluster is not reachable, each mount operation using `TryMount` will timeout after 60s ([default](https://docs.ceph.com/en/latest/man/8/mount.ceph/#basic)). As LXD retries 20 times (over 10s) this will cause LXD to wait for 10*60=600s (10mins) until it reaches ready state.

The PR sets the `mount_timeout` for CephFS vols to `1s` which seems to be the lowest possible. Setting it to `0` has the same effect as not setting it.

### DIFF
--- a/lxd/storage/drivers/driver_cephfs.go
+++ b/lxd/storage/drivers/driver_cephfs.go
@@ -301,7 +301,7 @@ func (d *cephfs) Create() error {
 
 	// Mount the pool.
 	srcPath := strings.Join(monAddresses, ",") + ":/"
-	err = TryMount(srcPath, mountPoint, "ceph", 0, fmt.Sprintf("name=%v,secret=%v,mds_namespace=%v", d.config["cephfs.user.name"], userSecret, fsName))
+	err = TryMount(srcPath, mountPoint, "ceph", 0, d.getMountOptions(d.config["cephfs.user.name"], userSecret, fsName))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


tomponline:
Perhaps TryMount should take a context.Context and during the loop we exit if the context is cancelled, allowing the caller to specify a timeout.

tomponline:
I'm happy for us to lower the cephfs level timeout to something lower than 60s, but worried about going as low as 1s.

roosterfish:
> Perhaps TryMount should take a context.Context and during the loop we exit if the context is cancelled, allowing the caller to specify a timeout.

I have tried this now, but even when providing a context, the underlying `unix.Mount` from Go doesn't accept a context so in case of CephFS which hangs for a few minutes by default, the context isn't considered as the first mount operation will run way longer than the context's timeout.
I'll check some more options here.

roosterfish:
@tomponline I have now added an "optimistic" context to `TryMount` to allow the caller having more flexibility. 
However CephFS is still an exception but I tried to make it clear in the comments/commit messages.

Let me know what you think about this approach but I found it to be the most balanced to unblock LXD's starting procedure in case the CephFS pool cannot be started.

tomponline:
Please rebase

roosterfish:
Rebased.

Just tested this again with a fresh MicroCloud and two storage pools `remote` and `remote-fs` (CephFS). 
In case Ceph is offline when the LXD daemons starts up and tries to start the storage pools, it now exits accordingly after the 10s context deadline for CephFS and logs this:

`2025-06-06T08:24:47Z lxd.daemon[12873]: time="2025-06-06T08:24:47Z" level=error msg="Failed mounting storage pool" err="Failed to mount \"10.87.144.48:6789:/\" on \"/var/snap/lxd/common/lxd/storage-pools/remote-fs\" using \"ceph\": context deadline exceeded" pool=remote-fs`

Afterwards the LXD daemon becomes ready and will retry again after 60s.

PR #15543, opened 2025-05-08
Daemon: Extend `waitready` with networks and storage pools

roosterfish:
The new MicroCloud e2e test suite surfaced a situation where after rebooting a cluster member, MicroCeph/MicroOVN aren't yet fully up and running which causes LXD not being able to start the networks and storage pools in the first shot. 

Whilst LXD will retry in the background, the `lxc cluster restore` command should not be called until the networks and storage pools are ready.

Therefore two new flags `--network` and `--storage` are added to the `lxd waitready` command to allow a client to wait for all resources to be ready.

### DIFF
--- a/lxd/api_internal.go
+++ b/lxd/api_internal.go
@@ -241,7 +241,12 @@ func internalWaitReady(d *Daemon, r *http.Request) response.Response {
 	}
 
 	timeoutSeconds := request.QueryParam(r, "timeout")
+	waitNetwork := shared.IsTrue(request.QueryParam(r, "network"))
+	waitStorage := shared.IsTrue(request.QueryParam(r, "storage"))
+
 	notReadyErr := errors.New("LXD daemon not ready yet")
+	networkNotReadyErr := errors.New("Networks not ready yet")
+	storageNotReadyErr := errors.New("Storage pools not ready yet")
 
 	if timeoutSeconds != "" {
 		timeoutSecondsUint, err := strconv.ParseUint(timeoutSeconds, 10, 32)
@@ -257,6 +262,24 @@ func internalWaitReady(d *Daemon, r *http.Request) response.Response {
 			defer cancel()
 		}
 
+		// If network is true, then block until all networks are marked as ready.
+		if waitNetwork {
+			select {
+			case <-d.waitNetworkReady.Done(): // Block until all networks are ready.
+			case <-ctx.Done(): // Don't leave this go routine around if client disconnects or timeout reached.
+				return response.Unavailable(networkNotReadyErr)
+			}
+		}
+
+		// If storage is true, then block until all storage pools are marked as ready.
+		if waitStorage {
+			select {
+			case <-d.waitStorageReady.Done(): // Block until all storage pools are ready.
+			case <-ctx.Done(): // Don't leave this go routine around if client disconnects or timeout reached.
+				return response.Unavailable(storageNotReadyErr)
+			}
+		}
+
 		select {
 		case <-d.waitReady.Done(): // Block until LXD is ready and then return EmptySyncResponse.
 		case <-ctx.Done(): // Don't leave this go routine around if client disconnects or timeout reached.
@@ -264,6 +287,10 @@ func internalWaitReady(d *Daemon, r *http.Request) response.Response {
 		}
 	} else if d.waitReady.Err() == nil {
 		return response.Unavailable(notReadyErr)
+	} else if waitNetwork && d.waitNetworkReady.Err() == nil {
+		return response.Unavailable(networkNotReadyErr)
+	} else if waitStorage && d.waitStorageReady.Err() == nil {
+		return response.Unavailable(storageNotReadyErr)
 	}
 
 	return response.EmptySyncResponse // LXD is ready.
--- a/lxd/state/state.go
+++ b/lxd/state/state.go
@@ -24,6 +24,7 @@ import (
 	"github.com/canonical/lxd/lxd/sys"
 	"github.com/canonical/lxd/lxd/ubuntupro"
 	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/cancel"
 )
 
 // State is a gateway to the two main stateful components of LXD, the database
@@ -102,6 +103,12 @@ type State struct {
 
 	// Ubuntu pro settings.
 	UbuntuPro *ubuntupro.Client
+
+	// NetworkReady can be used to track whether all networks are successfully started.
+	NetworkReady cancel.Canceller
+
+	// StorageReady can be used to track whether all storage pools are successfully started.
+	StorageReady cancel.Canceller
 }
 
 // LeaderInfo represents information regarding cluster member leadership.
--- a/test/suites/waitready.sh
+++ b/test/suites/waitready.sh
@@ -1,0 +1,82 @@
+test_waitready() {
+  # Create storage pool.
+  local lxd_backend storage_pool
+  lxd_backend=$(storage_backend "${LXD_DIR}")
+  storage_pool="lxdtest-$(basename "${LXD_DIR}")-pool"
+  br_name="lxdt$$"
+
+  lxc storage create "${storage_pool}" "${lxd_backend}"
+  lxc network create "${br_name}"
+
+  echo "==> Corrupt the network by setting an invalid external interface"
+  ip link add foo type bridge
+  ip addr add dev foo 10.1.123.10/24
+  # Inject the config setting manually to prevent validation by LXD.
+  network_id="$(lxd sql global "select id from networks where name='${br_name}'" -f csv)"
+  lxd sql global "INSERT INTO networks_config (network_id, node_id, key, value) VALUES (${network_id}, null, 'bridge.external_interfaces', 'foo')"
+
+  # Stop the daemon.
+  shutdown_lxd "${LXD_DIR}"
+
+  echo "==> Corrupt the storage pool directory to cause errors when LXD starts trying to start the pool"
+  # Perform this after stopping the daemon to make sure all mounts of the storage pool directory are given up.
+  rm -rf "${LXD_DIR}/storage-pools/${storage_pool}"
+  touch "${LXD_DIR}/storage-pools/${storage_pool}"
+
+  # Start the daemon.
+  respawn_lxd "${LXD_DIR}" true
+
+  # Wait for storage and network.
+  # When requesting both --network and --storage the request will already timeout on --network.
+  echo "==> LXD started but fails to start all networks and storage pools"
+  [ "$(lxd waitready --network --storage --timeout 1 2>&1)" = "Error: Networks not ready yet after 1s timeout" ]
+
+  # Not setting a timeout should have the same effect and return instantly.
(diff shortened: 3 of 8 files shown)


lxd/state/state.go
```diff
@@ -70,6 +70,14 @@ type State struct {
 	// The cache is also refreshed on dqlite heartbeat to synchronise with other members.
 	UpdateIdentityCache func()
 
+	// UpdateNetworkReady can be used to indicate all networks were successfully started.
+	// It cancels the daemon's underlying canceller to unblock all waits on networks.
+	UpdateNetworkReady func()
```
tomponline, line 75: Can we pas the whole cancel.Canceller into state for these so that other areas of the code base can see if the networks and storage pools are ready too?

lxd/api_internal.go
```diff
@@ -257,6 +259,24 @@ func internalWaitReady(d *Daemon, r *http.Request) response.Response {
 			defer cancel()
 		}
 
+		// If network is 1 then block until all networks are marked as ready.
+		if waitNetwork == "1" {
```
tomponline, line 263: can we use `shared.IsTrue()` here?

roosterfish: Thanks for the hint, I was looking for something similar but haven't found something. Fixed.

lxd/state/state.go
roosterfish: Initially my thought was to not expose too much. Earlier I used a channel (as suggested yesterday) and I have protected it's `close()` with a `sync.OnceFunc` and this was what was originally behind those funcs in the state.

But now since its using the `cancel.Canceller` I think we can expose everything for the reasons you mentioned.

Fixed :)

lxd/state/state.go
```diff
@@ -102,6 +103,12 @@ type State struct {
 
 	// Ubuntu pro settings.
 	UbuntuPro *ubuntupro.Client
+
+	// NetworkStatus can be used to track whether all networks are successfully started.
+	NetworkStatus cancel.Canceller
```
tomponline, line 108: NetworkReady?

tomponline (changes requested):
Please can we have some tests to check for this working?

You could, for each network and pool, start LXD and create the entity, shutdown LXD, then break it somehow (create a conflicting interface, or remove a directory or something) and then start LXD up and use lxd waitready with an appropriate flag and timeout and check it doesn't return. Then resolve the issue and issue waitready again to check its resolved?

lxd/api_internal.go
```diff
@@ -241,6 +241,8 @@ func internalWaitReady(d *Daemon, r *http.Request) response.Response {
 	}
 
 	timeoutSeconds := request.QueryParam(r, "timeout")
+	waitNetwork := request.QueryParam(r, "network")
+	waitStorage := request.QueryParam(r, "storage")
```
simondeziel, line 245: We're only acting on those network/storage params if the timeout param was provided. Is that intentional? What it I'd like to wait indefinitely for everything to be ready?

simondeziel, line 245: We're only acting on those network/storage params if the timeout param was provided. Is that intentional? What it I'd like to wait indefinitely for everything to be ready?

roosterfish: There are three cases:

1. `timeout=0`: No deadline gets set which means you can wait for all daemon, storage and network indefinitely
2. `timeout=x`: The deadline gets set to x and you can wait for all daemon, storage and network for x seconds
3. omit timeout: Returns 503 immediately in case the daemon is not yet ready, otherwise 200. 

So case 1) is what you are looking for but I just realized we have to include storage and network also in case 3).

lxd/state/state.go
roosterfish: I was looking for something suitable, but yeah "Ready" sounds good too. And its using the same scheme as the daemon's internal `wait(Network|Storage)Ready`. Will fix.

lxd/api_internal.go
tomponline: case 1 is already handled, but 3 is missing as you say

lxd/state/state.go
roosterfish: Fixed.

lxd/api_internal.go
roosterfish: Fixed, also added a check in the new test suite.

test/suites/waitready.sh
```diff
@@ -0,0 +1,73 @@
...
+  # Replace the directory with a simple file which should cause an error.
+  rm -r "${LXD_DIR}/storage-pools/${storage_pool}"
+  touch "${LXD_DIR}/storage-pools/${storage_pool}"
+
+  # Restart the daemon.
+  shutdown_lxd "${LXD_DIR}"
+  respawn_lxd "${LXD_DIR}" true
+
+  # Wait 10s for storage.
+  # This should not return as LXD cannot start the storage due to error.
```
tomponline, line 21: LXD storage cannot start (LXD did start)

test/suites/waitready.sh
```diff
@@ -0,0 +1,73 @@
...
+  ! lxd waitready --storage --timeout 10 | grep -q "Error: Storage pools not ready yet after 10s timeout"
```
tomponline, line 22: I believe we are doing exact match on errors these days rather than grep.

roosterfish:
Mhh, if you run the test as standalone you don't get this error: `rm: cannot remove '/home/runner/work/lxd/lxd/test/tmp.YSe/rIW/storage-pools/lxdtest-rIW-pool': Device or resource busy`. I suppose the suite should spawn another LXD dedicated for the purpose of this test? 

I never looked closely enough into the test suite setup but I would guess the same underlying test LXD daemon which is getting spawned is used across all test suites which is why this error surfaces?

test/suites/waitready.sh
roosterfish: Rephrased.

test/suites/waitready.sh
roosterfish: You mean using `[ "$(... 2>&1)" = "Error: ..." ]`?

test/suites/waitready.sh
```diff
@@ -0,0 +1,83 @@
...
+  # Corrupt the network by setting an invalid MTU.
+  lxd sql global 'INSERT INTO networks_config (network_id, node_id, key, value) VALUES (1, null, "bridge.mtu", "42")'
+
+  # Restart the daemon.
+  shutdown_lxd "${LXD_DIR}"
+  respawn_lxd "${LXD_DIR}" true
+
+  # Wait 10s for network.
+  # LXD started but fails to start all networks.
+  ! lxd waitready --network --timeout 10 | grep -q "Error: Networks not ready yet after 10s timeout"
```
simondeziel, line 58: Could we get away with shorter timeouts when failures are expected? I'm asking because the average runtime of our test functions is 13s (but it's heavily skewed by `shutdown` taking 5.5 minutes :/).

test/suites/waitready.sh
```diff
@@ -0,0 +1,83 @@
...
+  # Replace the directory with a simple file which should cause an error.
+  rm -r "${LXD_DIR}/storage-pools/${storage_pool}"
+  touch "${LXD_DIR}/storage-pools/${storage_pool}"
+
+  # Restart the daemon.
+  shutdown_lxd "${LXD_DIR}"
+  respawn_lxd "${LXD_DIR}" true
+
+  # Wait 10s for storage.
+  # LXD started but fails to start all storage pools.
```
simondeziel, line 28: Could you turn some of those comments into `echo`? This makes debugging so much easier ;)

test/suites/waitready.sh
simondeziel: To add to that, by using `[ "$(lxd waitready ...)" = "The error we expect" ]` we will see what is compared even if the comparison doesn't match (thanks to `set -e`). This is handy when error messages change.

test/suites/waitready.sh
```diff
@@ -0,0 +1,83 @@
...
+  # Create storage pool.
+  local lxd_backend storage_pool
+  lxd_backend=$(storage_backend "$LXD_DIR")
+  storage_pool="lxdtest-$(basename "${LXD_DIR}")-pool"
+  br_name="lxdt$$"
+
+  lxc storage create "${storage_pool}" "${lxd_backend}"
+  lxc network create "${br_name}"
+
+  # Corrupt the storage pool directory to cause errors when LXD starts trying to start the storage pools.
```
simondeziel, line 18: What do you think of testing the scenario where both the storage and the network are corrupted but you ask to wait for just one of them. Would it make sense to start from this very broken state and slowly move to a fully functional LXD?

simondeziel (commented):
Very nice and I especially appreciate seeing system tests for it!

test/suites/waitready.sh
simondeziel: Good spot on the `stderr` redirection missing :)

test/suites/waitready.sh
roosterfish: I really much like this idea. Less restarts and you can actually see that the errors resolve itself. I have changed the test suite accordingly.
However now we need to have longer `timeout` settings as we have to wait for LXD to retry the network/storage pool start (which happens every 60s). But now the wait happens for a good reason!

test/suites/waitready.sh
roosterfish: Fixed.

test/suites/waitready.sh
roosterfish: Fixed.

test/suites/waitready.sh
simondeziel: Correctness > Performance so I'm happy with the balance you struck ;)

test/suites/waitready.sh
```diff
@@ -0,0 +1,71 @@
...
+  # Wait for storage and network.
+  # When requesting both --network and --storage the request will already timeout on --network.
+  echo "==> LXD started but fails to start all networks and storage pools"
+  [ "$(lxd waitready --network --storage --timeout 1 2>&1)" = "Error: Networks not ready yet after 1s timeout" ]
+
+  # Not setting a timeout should have the same effect and return instantly.
+  [ "$(DEBUG="" lxc query "/internal/ready?network=1&storage=1" 2>&1)" = "Error: Networks not ready yet" ]
+
+  # The standard waitready without extra flags should still return with success.
+  lxd waitready
```
simondeziel, line 39: Is it worth checking that `lxc network list` has failed networks and `lxc storage list` has missing pools?

roosterfish: Took me a bit to identify something which consistently breaks all pool drivers. But it should be in now including the check for `lxc (network|storage) ls`. I'll ping you again when the pipeline is finally green.

roosterfish: Ready now.

roosterfish:
@tomponline would be great if we can backport this to 5.21 too, as it is required at least in 5.21/edge so we can make use of it in the MicroCloud test pipeline.

tomponline:
> @tomponline would be great if we can backport this to 5.21 too, as it is required at least in 5.21/edge so we can make use of it in the MicroCloud test pipeline.

it will be

tomponline (approved):
LGTM

PR #15545, opened 2025-05-08
Remove `lxc-to-lxd`

simondeziel:


simondeziel:
One of the benefit of dropping the `lxc-to-lxd` binary is that our `system-test-deps` artifacts are now smaller (156MiB vs 170MiB).

tomponline (approved):
Thanks

minaelee:
Cheers, thanks Simon! As a nice side effect, removing the "Migrate from lxd" (`migrate-from-lxc.md`) document from the docs makes it much easier to rename the "Move instances" doc to "Migrate instances".

Note: I was scratching my head trying to figure out why there was no `migration.md` document accessible from the docs site (the one edited by this PR, not deleted). Turns out that `custom_conf.py` is set up to redirect `migration.md` to the page about using the (hopefully soon-to-be-renamed) lxc-migration tool. Not sure why the `migration.md` page exists at all when it's always redirected; will look into it and probably delete the page & redirect the redirect to "Migrate instances".

PR #15547, opened 2025-05-09
doc: add info about snaps and user idmaps

minaelee:
This PR notes that the Allowed ranges section is not needed for the LXD snap, due to the daemon running as root in the snap environment and disregarding the `/etc/sub{u,g}id` files.


tomponline (approved):
Thanks!

PR #15548, opened 2025-05-09
Database: Adds image.ToAPI() and uses it

tomponline:
This is a first step to rework `distributeImage()` and its callers by implementing and using `ToAPI()` for `db/cluster.Image` type.

This will allow the `GetImageByFingerprintPrefix` and `GetImageFromAnyProject` amongst other functions to be later changed to return a `cluster.Image` type rather than `api.Image` type, to avoid duplicate lookups because different functions use different types in the callers of `distributeImage`.

This continues the pattern of adding `ToAPI()` to our DB types.


simondeziel (approved):
LGTM for what it's worth.

PR #15552, opened 2025-05-10
Cluster: Add post-migration instance device cleanup

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/15511.

### DIFF
--- a/lxd/device/nic_bridged.go
+++ b/lxd/device/nic_bridged.go
@@ -876,6 +876,20 @@ func (d *nicBridged) postStop() error {
 	return nil
 }
 
+// PostMigrateSend is run after an instance is migrated to another cluster member.
+func (d *nicBridged) PostMigrateSend() error {
+	// Populate device config with volatile fields (hwaddr) if needed.
+	networkVethFillFromVolatile(d.config, d.volatileGet())
+
+	// Remove device (removing dnsmasq lease and config). This is required to reset leases post-migration.
+	err := d.Remove()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // Remove is run when the device is removed from the instance or the instance is deleted.
 func (d *nicBridged) Remove() error {
 	// Handle the case where validation fails but the device still must be removed.
--- a/lxd/instance/drivers/driver_common.go
+++ b/lxd/instance/drivers/driver_common.go
@@ -510,6 +510,24 @@ func (d *common) deviceVolatileSetFunc(devName string) func(save map[string]stri
 	}
 }
 
+// postMigrateSendCommon handles the common part of instance post-migration send.
+func (d *common) postMigrateSendCommon(inst instance.Instance) error {
+	// Perform post-migration device cleanup.
+	for devName, devConfig := range d.ExpandedDevices() {
+		dev, err := d.deviceLoad(inst, devName, devConfig)
+		if err != nil {
+			return err
+		}
+
+		err = dev.PostMigrateSend()
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 // expandConfig applies the config of each profile in order, followed by the local config.
 func (d *common) expandConfig() error {
 	var globalConfigDump map[string]any
--- a/lxd/instance/drivers/driver_lxc.go
+++ b/lxd/instance/drivers/driver_lxc.go
@@ -6382,6 +6382,11 @@ func (d *lxc) MigrateReceive(args instance.MigrateReceiveArgs) error {
 	}
 }
 
+// PostMigrateSend performs any required cleanup steps after an instance has been migrated to another member.
+func (d *lxc) PostMigrateSend() error {
+	return d.postMigrateSendCommon(d)
+}
+
 // ConversionReceive establishes the filesystem connection, transfers the filesystem / block volume,
 // and creates an instance from it.
 func (d *lxc) ConversionReceive(args instance.ConversionReceiveArgs) error {
--- a/lxd/instance/drivers/driver_qemu.go
+++ b/lxd/instance/drivers/driver_qemu.go
@@ -7568,6 +7568,11 @@ func (d *qemu) MigrateReceive(args instance.MigrateReceiveArgs) error {
 	}
 }
 
+// PostMigrateSend performs any required cleanup steps after an instance has been migrated to another member.
+func (d *qemu) PostMigrateSend() error {
+	return d.postMigrateSendCommon(d)
+}
+
 // ConversionReceive establishes the filesystem connection, transfers the filesystem / block volume,
 // and creates an instance from it.
 func (d *qemu) ConversionReceive(args instance.ConversionReceiveArgs) error {
--- a/lxd/instance_post.go
+++ b/lxd/instance_post.go
@@ -651,7 +651,8 @@ func instancePostMigration(s *state.State, inst instance.Instance, req api.Insta
 	return nil
 }
 
-// Move a non-ceph instance to another cluster node. Source and target members must be online.
+// Migrate an instance to another cluster node (supports both local and remote storage).
+// Source and target members must be online.
 func instancePostClusteringMigrate(ctx context.Context, s *state.State, srcPool storagePools.Pool, srcInst instance.Instance, newInstName string, srcMember db.NodeInfo, newMember db.NodeInfo, stateful bool, allowInconsistent bool) (func(op *operations.Operation) error, error) {
 	srcMemberOffline := srcMember.IsOffline(s.GlobalConfig.OfflineThreshold())
 
--- a/lxd/migrate_instance.go
+++ b/lxd/migrate_instance.go
@@ -156,6 +156,12 @@ func (s *migrationSourceWs) Do(state *state.State, migrateOp *operations.Operati
 		return fmt.Errorf("Failed migration on source: %w", err)
 	}
 
+	err = s.instance.PostMigrateSend()
+	if err != nil {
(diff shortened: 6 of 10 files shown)


simondeziel (commented):
Looking at the fixed issue, it seems like it should be relatively easy to add some system tests. Maybe there is already a test where a container is migrated between cluster members? If so, checking for the relevant files to be cleaned up/update would be nice.

kadinsayani:
> Looking at the fixed issue, it seems like it should be relatively easy to add some system tests. Maybe there is already a test where a container is migrated between cluster members? If so, checking for the relevant files to be cleaned up/update would be nice.

I'll add some tests :)

test/suites/clustering.sh
```diff
@@ -2514,6 +2514,15 @@ test_clustering_fan() {
   LXD_DIR="${LXD_ONE_DIR}" lxc exec c1 -- ping -c2 -W5 "${IP_C2}"
   LXD_DIR="${LXD_ONE_DIR}" lxc exec c2 -- ping -c2 -W5 "${IP_C1}"
 
+  echo "Check that the DHCP leases are cleaned up post-migration"
+  lxc stop c1
```
simondeziel, line 2518:
```suggestion
  lxc stop -f c1
```

test/suites/clustering.sh
```diff
@@ -2514,6 +2514,15 @@ test_clustering_fan() {
...
+  LXD_DIR="${LXD_ONE_DIR}" lxc move c1 --target node2
+  lxc start c1
+  if grep " c1" "${LXD_ONE_DIR}/networks/${fanbridge}/dnsmasq.leases" ; then
```
simondeziel, line 2521:
```suggestion
  if grep -qF " c1 " "${LXD_ONE_DIR}/networks/${fanbridge}/dnsmasq.leases" ; then
```

This slight tweak is to avoid any risk of matching with any IPv6 hex (granted FAN doesn't support v6) or a MAC address.

test/suites/clustering.sh
```diff
@@ -2514,6 +2514,15 @@ test_clustering_fan() {
   LXD_DIR="${LXD_ONE_DIR}" lxc exec c1 -- ping -c2 -W5 "${IP_C2}"
   LXD_DIR="${LXD_ONE_DIR}" lxc exec c2 -- ping -c2 -W5 "${IP_C1}"
 
+  echo "Check that the DHCP leases are cleaned up post-migration"
```
simondeziel, line 2517: Could you copy the exact same `grep` (without the surrounding `if`) here to ascertain that the pattern we look for is there when it should be and goes away when it should.

simondeziel (commented):
Thanks for the test, I just have tiny nitpicks.

simondeziel (dismissed):
Awesome, thanks!

tomponline (changes requested):
Whilst we're looking at this the comment:

> // Move a non-ceph instance to another cluster node. Source and target members must be online.

Is not correct, is it has explicit logic for remote storage pools.

Instead this should say something like:

> // Move an instance to another cluster node (supports both local and remote storage). Source and target members must be online.

lxd/instance_post.go
```diff
@@ -873,6 +874,30 @@ func instancePostClusteringMigrate(s *state.State, r *http.Request, srcPool stor
 			if err != nil {
 				return fmt.Errorf("Failed deleting instance on source member: %w", err)
 			}
+
+			// Perform post-migration device cleanup.
```
tomponline, line 878: This seems like a bit of a kludge as its calling the devices subsystem from outside of the instance driver subsystem (unlike the rest of the calls AFAIK).

I can see why you've done it like this because the rest of the "cleanup" done for the source instance is done outside of the instance subsystem too, like the calls to `srcPool.DeleteInstance()` above.

But I'm concerned that we've crossed a line where it doesn't make sense to continue to complicate this function, and instead perhaps we would be better off adding a `PostMigrateSend()` to the instance interface, as we already have a `MigrateSend()` function called from `migrationSourceWs.Do()`.

And then move the stuff from line 817 to `if !stateful && srcInstRunning {` (not inclusive)  into the instance subsystem.

Then we would rename the device's `PostMigrate` to `PostMigrateSend` to mirror the instance's one and to show this should only be called on the sender.

This way we would nicely segregate the source instance post migration send logic into the instance subsystem.

kadinsayani:
I think it makes sense to land https://github.com/canonical/lxd/pull/15595 first, since that PR significantly reworks server-side migration logic. I'll keep this PR as draft for now.

kadinsayani:
> I think it makes sense to land #15595 first, since that PR significantly reworks server-side migration logic. I'll keep this PR as draft for now.

On second thought, it’s fine if we decide to merge this PR first. After all, #15595 is significantly larger and requires more review.

lxd/instance_post.go
kadinsayani: I’ve implemented a `PostMigrateSend()` function for the new cleanup steps. However, I haven’t moved the cleanup steps from `instancePostClusteringMigrate` because that function is meant to be called by the target/receiver.

lxd/device/nic_bridged.go
```diff
@@ -876,6 +876,20 @@ func (d *nicBridged) postStop() error {
 	return nil
 }
 
+// PostMigrateSend is run after an instance is migrated to another cluster member.
+func (d *nicBridged) PostMigrateSend() error {
+	// Populate device config with volatile fields (hwaddr) if needed.
+	networkVethFillFromVolatile(d.config, d.volatileGet())
+
+	// Remove device (removing dnsmaq lease and config). This is required to reset leases post-migration.
```
simondeziel: `s/dnsmaq/dnsmasq/`

lxd/instance/drivers/driver_lxc.go
```diff
@@ -6382,6 +6382,16 @@ func (d *lxc) MigrateReceive(args instance.MigrateReceiveArgs) error {
...
 }
 
+// PostMigrateSend performs any required cleanup steps after an instance has been migrated to another member.
+func (d *lxc) PostMigrateSend() error {
+	err := d.postMigrateSendCommon(d)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
simondeziel: Maybe it's a question of style/convention but I find this to read better.
```suggestion
	return d.postMigrateSendCommon(d)
```

lxd/instance/drivers/driver_qemu.go
```diff
@@ -7568,6 +7568,16 @@ func (d *qemu) MigrateReceive(args instance.MigrateReceiveArgs) error {
 	}
 }
 
+// PostMigrateSend performs any required cleanup steps after an instance has been migrated to another member.
+func (d *qemu) PostMigrateSend() error {
+	err := d.postMigrateSendCommon(d)
```
simondeziel: Same here.

simondeziel (approved):

nmezhenskyi (approved):
LGTM!

lxd/instance/drivers/driver_common.go
```diff
@@ -510,6 +510,32 @@ func (d *common) deviceVolatileSetFunc(devName string) func(save map[string]stri
...
 
+// postMigrateSendCommon handles the common part of instance post-migration send.
+func (d *common) postMigrateSendCommon(inst instance.Instance) error {
+	// Perform post-migration device cleanup.
+	devices := d.ExpandedDevices()
+	localConfig := d.LocalConfig()
+
+	for devName, devConfig := range devices {
+		// Restore hardware address for NIC devices from volatile config.
+		if devConfig["type"] == "nic" && devConfig["hwaddr"] == "" && localConfig["volatile."+devName+".hwaddr"] != "" {
```
tomponline: This isn't needed right? Because this is what `networkVethFillFromVolatile` does in the earlier commit?

I'd prefer to avoid putting device specific logic into device agnostic areas of the code.

lxd/instance/drivers/driver_common.go
```diff
@@ -510,6 +510,32 @@ func (d *common) deviceVolatileSetFunc(devName string) func(save map[string]stri
 	}
 }
 
+// postMigrateSendCommon handles the common part of instance post-migration send.
+func (d *common) postMigrateSendCommon(inst instance.Instance) error {
+	// Perform post-migration device cleanup.
+	devices := d.ExpandedDevices()
+	localConfig := d.LocalConfig()
+
+	for devName, devConfig := range devices {
```
tomponline: Can just do `devName, devConfig := range d.ExpandedDevices()`

lxd/migrate_instance.go
```diff
@@ -156,6 +156,12 @@ func (s *migrationSourceWs) Do(state *state.State, migrateOp *operations.Operati
 		return fmt.Errorf("Failed migration on source: %w", err)
 	}
 
+	err = s.instance.PostMigrateSend()
+	if err != nil {
+		l.Error("Failed post-migration cleanup on source", logger.Ctx{"err": err})
```
tomponline: "cleanup" isn't appropriate here, as we dont know that what each instance/device will do is "cleanup".

tomponline (changes requested):
A few nits, but overall this looks nice, thanks.

lxd/instance/drivers/driver_common.go
kadinsayani: Correct, thanks for pointing that out.

kadinsayani:
> A few nits, but overall this looks nice, thanks.

Thanks for your review, I've addressed your comments :)

tomponline (approved):
thanks!

PR #15555, opened 2025-05-12
lxd/db/cluster/images: Only populate update image source if found

tomponline:
Otherwise leave UpdateSource as nil.

Fixes regression from https://github.com/canonical/lxd/pull/15548


kadinsayani (approved):

markylaing (approved):
LGTM

PR #15579, opened 2025-05-14
github: only check lxc/lxd-agent bin sizes of not built for coverage

simondeziel:
Binaries built with coverage information are inherently bigger than
regular/production builds.


tomponline (approved):
very good thanks!

PR #15580, opened 2025-05-14
staticcheck.conf: add dedicated config to ignore ST1005

simondeziel:
Some linters like VSCode's Go extension and TIOBE scanner do not consult the `.golangci.yaml` so reporting many problems with error strings not starting with a capital letter.


tomponline (approved):
Great!

PR #15581, opened 2025-05-14
Avoid integer overflows

simondeziel:


shared/util_linux.go
```diff
@@ -61,6 +62,15 @@ func GetPathMode(path string) (os.FileMode, error) {
...
 	var dimensions [4]uint16
+
+	if width > math.MaxUint16 || height > math.MaxUint16 {
+		return fmt.Errorf("Width or height too large: %d %d", width, height)
+	}
+
+	if width < 0 || height < 0 {
+		return fmt.Errorf("Width and height must not be negative: %d %d", width, height)
+	}
+
```
markylaing, line 73: I was thinking we could refactor this function to accept arguments of the correct type but this has a lot of impacts elsewhere. For example, I expected the file descriptor to be a `uint` but it actually needs to be an `int` because a file descriptor of `-1` is special cased to be the control websocket... So I think we'd best play it safe and check for overflows exactly as you have done :laughing:

markylaing (approved):
LGTM

PR #15582, opened 2025-05-14
Makefile: allow easy sideloading of lxc/lxd onto the snap

simondeziel:


tomponline:
As discussed, lets not include snap specific items in the LXD makefile.

Issue #15554, opened 2025-05-11
Server Certificate Mechanism


irtaza9:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

I am facing an issue in `lxd`, I successfully configured the lxd cluster and then I checked `lxc config trust ls` which give me the rows of all my members in the cluster then I removed them using `lxc config trust remove <fingerprint>` after that my cluster is not able to communicate. Now I want to add the server certificate for my servers so that I can re-configure the tls authentication.

I am using self-signed certificates and my new certificates are in `/var/snap/lxd/common/lxd` 

### Describe the solution you'd like

You guys can better comment on the solution.

### Additional context

communication over forum:

https://discourse.ubuntu.com/t/cluster-configurations-issues-lxd-cluster-crashes-when-powered-on-after-force-removing-offline-node/60738/4

https://discourse.ubuntu.com/t/lxd-cluster-invalid-client-certificate/38393

tomponline:
This scenario is now prevented via https://github.com/canonical/lxd/pull/15583

PR #15583, opened 2025-05-14
Prevent server certificate deletion

markylaing:
In #15554 the server certificates were removed, this broke the cluster. This PR does not add a mechanism for re-adding server certificates as the author suggests, instead it prevents deletion of server certificates via `DELETE /1.0/certificates`. There is no reason that a server certificate should be deleted. Updating server certificates is still allowed as this could be used to rotate expiring certificates, but no process is documented for this yet.


tomponline (approved):
Ta!

PR #15584, opened 2025-05-14
Replace `ValueInSlice` with `slices.Contains`

kadinsayani:
Here are the two `gopatch` patches I used for this:

```
# Replace shared.ValueInSlice with slices.Contains
@@
var x expression
var y expression
@@
-shared.ValueInSlice(x,y)
+slices.Contains(y,x)
```

```
# Replace ValueInSlice with slices.Contains
@@
var x expression
var y expression
@@
-ValueInSlice(x,y)
+slices.Contains(y,x)
```

### DIFF
--- a/client/connection.go
+++ b/client/connection.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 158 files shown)


simondeziel (approved):
Thanks for doing this, LGTM.

While reviewing it I noticed that some bits could be improved but only in a follow-up PR.

kadinsayani:
> Thanks for doing this, LGTM.
> 
> While reviewing it I noticed that some bits could be improved but only in a follow-up PR.

Thanks for your review! I've opened up a follow-up PR: https://github.com/canonical/lxd/pull/15589.

PR #15585, opened 2025-05-15
util/http: Improve etag error message to guide towards a solution

edlerd:
# Done
- Improve etag error message to guide towards a solution
- resolving https://github.com/canonical/lxd-ui/issues/1228

### DIFF
--- a/lxd/util/http.go
+++ b/lxd/util/http.go
@@ -93,7 +93,7 @@ func EtagCheck(r *http.Request, data any) error {
 	}
 
 	if hash != match {
-		return api.StatusErrorf(http.StatusPreconditionFailed, "ETag doesn't match: %s vs %s", hash, match)
+		return api.StatusErrorf(http.StatusPreconditionFailed, "ETag doesn't match: %s vs %s. The configuration has been modified since this change began. Please retrieve the updated configuration before proceeding.", hash, match)
 	}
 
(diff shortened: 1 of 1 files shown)


simondeziel (dismissed):
Rebasing would fix the broken tests as it would get changes from https://github.com/canonical/lxd/pull/15592.

However, re-running CI just for this error string improvement feels wasteful.

lxd/util/http.go
```diff
@@ -93,7 +93,7 @@ func EtagCheck(r *http.Request, data any) error {
 	}
 
 	if hash != match {
-		return api.StatusErrorf(http.StatusPreconditionFailed, "ETag doesn't match: %s vs %s", hash, match)
+		return api.StatusErrorf(http.StatusPreconditionFailed, "ETag doesn't match: %s vs %s. The configuration has been modified since this change began. Please reload the configuration before proceeding.", hash, match)
```
tomponline, line 96: How about "Please retrieve the updated configuration before proceed" as im worried "reload" here might suggest to the users they need to reload LXD.

edlerd: Yes, good idea as the "reload" was bit UI focused and can be misunderstood on CLI. Adopting to the suggestion.

tomponline (approved):
Ta!

PR #15587, opened 2025-05-15
Update context (from Incus)

kadinsayani:
Cherry-picked from https://github.com/lxc/incus/pull/567.


simondeziel (dismissed):
Not being very familiar with the concept of contexts, I won't review the code itself. The cherry-picked commits all have proper attribution and license tag so I'll ACK that part.

tomponline:
@kadinsayani ready for rebase

lxd/api_cluster.go
```diff
@@ -59,7 +59,7 @@ import (
 )
 
 type evacuateStopFunc func(inst instance.Instance) error
-type evacuateMigrateFunc func(s *state.State, r *http.Request, inst instance.Instance, targetMemberInfo *db.NodeInfo, live bool, startInstance bool, metadata map[string]any, op *operations.Operation) error
+type evacuateMigrateFunc func(ctx context.Context, s *state.State, r *http.Request, inst instance.Instance, targetMemberInfo *db.NodeInfo, live bool, startInstance bool, metadata map[string]any, op *operations.Operation) error
```
MusicDin: I quickly checked few occurrences where this func type is used, and I think we could get rid of `r *http.Request` from input params - but needs to be double checked.

MusicDin (commented):
Already looking good. I think only the change in `LocateImage` is causing context deadline issues

lxd/api_cluster.go
kadinsayani: Thanks for spotting this, I've removed the unused `*http.Request` parameter.

tomponline:
@kadinsayani please update the import spreadsheet. Ta

PR #15593, opened 2025-05-18
tests: Fix `command not found` error

kadinsayani:
This was missed with https://github.com/canonical/lxd/pull/15534/commits/f31af2ba509efe9a1bb8c4a55d76d06f845c4bea.

Excerpt from today's tests:

```
++ query cluster:/1.0/instances/c1
suites/clustering_move.sh: line 65: query: command not found
```

### DIFF
--- a/test/suites/clustering_move.sh
+++ b/test/suites/clustering_move.sh
@@ -62,12 +62,12 @@ test_clustering_move() {
   lxc info cluster:c1 | grep -q "Location: node1"
 
   # c1 can be moved within the same cluster group if it has multiple members
-  current_location="$(query cluster:/1.0/instances/c1 | jq -r '.location')"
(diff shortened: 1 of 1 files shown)


simondeziel (approved):
Thanks

tomponline (approved):
Nice catch!

simondeziel:
> Nice! Thanks and sorry for the search and replace errors here. I'm wondering how the tests ever passed!

They passed because we don't run `bash` with `set -o pipefail` so errors in pipelines are silently ignored. I'm working on fixing that shortcoming in #15594.

kadinsayani:
> Nice! Thanks and sorry for the search and replace errors here. I'm wondering how the tests ever passed!

No worries it happens! I just happened to notice this while running the `clustering_move` tests locally :)

Thanks @simondeziel for following up with #15594.

PR #15594, opened 2025-05-18
test/main: catch errors in pipelines

simondeziel:
https://github.com/canonical/lxd/commit/f31af2ba509efe9a1bb8c4a55d76d06f845c4bea

Reported-by: Kadin Sayani <kadin.sayani@canonical.com>


simondeziel:
This PR is expected to fail during cluster system tests until #15593 is merged to address the underlying issue.

kadinsayani (dismissed):
LGTM thanks!

tomponline:
Ready for rebase

simondeziel:
> Ready for rebase

`pipefail` turned up many more errors due to `something | grep -q pattern` aborting the pipeline early when the first `pattern` match hits. This happens for tools (the `something`) that don't gracefully handle their `stdout` being closed abruptly. This is explained here: https://www.pixelbeat.org/programming/sigpipe_handling.html

I've used that opportunity to improve many of the `grep` patterns to me more precise but it is still too early to know if it will be a viable thing to use `pipefail` as we have many occurrences of `grep -q` :/

simondeziel:
@kadinsayani and @tomponline after the many commits to handle `grep -q` halting the pipeline early on match, you should find commits addressing different problems if you check the changes starting from those touching `test/includes/lxd`.

Hopefully that will make the review a tad less tedious.

test/suites/clustering.sh
```diff
@@ -10,7 +10,7 @@ test_clustering_enable() {
     # shellcheck disable=SC2034,SC2030
     LXD_DIR=${LXD_INIT_DIR}
 
-    lxc config show | grep "core.https_address" | grep -qE "127.0.0.1:[0-9]{4,5}$"
+    lxc config show | grep -xE "^\s+core\.https_address: 127\.0\.0\.1:[0-9]{4,5}"
```
tomponline, line 13: "simplify" :)

tomponline (approved):
Thanks

What is wrong with `grep -q` and do we have a way to prevent future use of `grep -q` and `grep -v`?

tomponline:
@simondeziel the following commits do not backport cleanly, please can you follow up with a separate PR targeting stable-5.21 branch?

```
04176417583dcefec7845112a91a3b86099697bb test/includes/setup: don't abort pipelines early with `grep -q`
c24a9f9632d627f883dd54c503701269121cd8f6 test/suites/clustering: don't abort pipelines early with `grep -q`
30d1af3856d1de45c2c57e3025778c8c9032641d test/suites/projects: don't abort pipelines early with `grep -q`
b47298642fa1261f0612210e80d83fa696c381d6 test/suites/storage_snapshots: don't abort pipelines early with `grep -q`
c7bbb839dae51c7ddd98be1228897f67b1f75d19 test/suites/remote: use safer `grep` patterns
e85565437fa82279c8826c1f07a299bc6b8ebed1 test/suites/remote: ignore failure due to bad token when inspecting expected error message
a2648c749a41576aa64d1b566e06effd45e95830 test/suites/remote: `echo` what's going on in remote_admin
cdb2bc8f67cbcdb31643a4359520d01102b5663b test/suites/tls_restrictions: don't abort on expected failure from `curl`
```

tomponline:
@simondeziel actually this was causing test failures unless 100% backported (:() so ive backed it out, please can you backport separately, ta.

test/suites/clustering.sh
tomponline: How come?

simondeziel: I stuck with `lxc config show` as this is the only place where it's output is verified:

```
$ git grep 'lxc config show | '
test/suites/clustering.sh:    lxc config show | grep -xE "^\s+core\.https_address: 127\.0\.0\.1:[0-9]{4,5}"
```

It is also the only place where the test suite call `lxc config show` to get the global server config and not an instance's config. So it came down to a question of code coverage.

simondeziel:
> What is wrong with `grep -q` 

It's only a problem when used in pipelines as it causes `grep` to exit on first match (same as with `-m1`) which results in the command feeding the pipeline to be signaled that it's `stdout` was closed. Not all applications handle this well meaning they return an error that becomes problematic with `set -eo pipefail`.

> and do we have a way to prevent future use of `grep -q`

I'll add a check to flag it's use in pipelines, good idea.

> and `grep -v`?

This one requires to be a bit more creative when flagging its use as it's sometimes OK. I have a regex that should be a good first approximation ;)

Thanks for those good suggestions!

test/suites/clustering.sh
tomponline: makes sense thanks

PR #15621, opened 2025-05-19
Allow using the same volume for storage.images_volume and storage.backups_volume

skozina:
This depends on https://github.com/canonical/lxd/pull/15482 .

This also contains the move of images and backups to their own sub-directories if storage.{backups,images}_volume is set.

### DIFF
--- a/lxd/daemon_storage.go
+++ b/lxd/daemon_storage.go
@@ -271,6 +271,7 @@ func daemonStorageMove(s *state.State, storageType string, oldConfig string, new
 
 		sourceVolume = project.StorageVolume(api.ProjectDefaultName, sourceVolume)
 		sourcePath = storageDrivers.GetVolumeMountPath(sourcePool, storageDrivers.VolumeTypeCustom, sourceVolume)
+		sourcePath = filepath.Join(sourcePath, storageType)
 	}
 
 	moveContent := func(source string, target string) error {
@@ -280,20 +281,7 @@ func daemonStorageMove(s *state.State, storageType string, oldConfig string, new
 			return err
 		}
 
-		// Remove the source content.
-		entries, err := os.ReadDir(source)
-		if err != nil {
-			return err
-		}
-
-		for _, entry := range entries {
-			err := os.RemoveAll(filepath.Join(source, entry.Name()))
-			if err != nil {
-				return err
-			}
-		}
-
-		return nil
+		return os.RemoveAll(source)
 	}
 
 	// Deal with unsetting.
@@ -355,6 +343,12 @@ func daemonStorageMove(s *state.State, storageType string, oldConfig string, new
 		destPath = s.BackupsStoragePath()
 	}
 
+	// Ensure the destination directory structure exists within the target volume.
+	err = os.MkdirAll(destPath, 0700)
+	if err != nil {
+		return fmt.Errorf("Failed to create directory %q: %w", destPath, err)
+	}
+
 	err = os.Chmod(destPath, 0700)
 	if err != nil {
 		return fmt.Errorf("Failed to set permissions on %q: %w", destPath, err)
@@ -392,12 +386,6 @@ func daemonStorageMove(s *state.State, storageType string, oldConfig string, new
 	err = moveContent(sourcePath, destPath)
 	if err != nil {
 		return fmt.Errorf("Failed to move data over to directory %q: %w", destPath, err)
-	}
-
-	// Remove the old data.
-	err = os.RemoveAll(sourcePath)
-	if err != nil {
-		return fmt.Errorf("Failed to cleanup old directory %q: %w", sourcePath, err)
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


lxd/daemon_storage.go
```diff
@@ -233,16 +234,18 @@ func daemonStorageValidate(s *state.State, target string) (validatedTarget strin
 		return "", fmt.Errorf("Failed to list %q: %w", mountpoint, err)
 	}
 
+	allowedEntries := []string{
+		"lost+found", // Clean ext4 volumes.
+		".zfs",       // Systems with snapdir=visible
```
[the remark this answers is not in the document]
skozina: Not mine, that was already in the code, I just refactored it a bit;-)

tomponline:
@skozina ready for rebase

skozina:
> @skozina ready for rebase

Thanks! Rebased.

lxd/patches.go
```diff
@@ -1518,4 +1519,53 @@ func patchRemoveBackupsImagesSymlinks(_ string, d *Daemon) error {
 	return nil
 }
 
+func moveStorage(destPath string) error {
```
tomponline: can we define this as a function inside `patchMoveBackupsImagesStorage` so its scope is limited to that path?

lxd/daemon_storage.go
```diff
@@ -252,6 +255,19 @@ func daemonStorageValidate(s *state.State, target string) (validatedTarget strin
 	return pool.Name() + "/" + dbVol.Name, nil
 }
 
+func numberOfVolumeReferences(s *state.State, storageType string, volume string) int {
```
tomponline: Can this be called `hasVolumeReferences` and return a `bool`, then you can return true on the first reference and dont need to check for subsequent references, as the only usage of this function is if it returns  0.

lxd/daemon_storage.go
```diff
@@ -308,11 +324,13 @@ func daemonStorageMove(s *state.State, storageType string, oldConfig string, new
 			return err
 		}
 
-		// Unmount old volume.
-		projectName, sourceVolumeName := project.StorageVolumeParts(sourceVolume)
-		_, err = pool.UnmountCustomVolume(projectName, sourceVolumeName, nil)
-		if err != nil {
-			return fmt.Errorf(`Failed to umount storage volume "%s/%s": %w`, sourcePool, sourceVolumeName, err)
+		// Unmount old volume if noone else is using it.
+		if numberOfVolumeReferences(s, storageType, oldConfig) == 0 {
```
tomponline: Actually looking at the use of `numberOfVolumeReferences` it seems to be partially re-implementing the volume mount reference tracking that LXD already does in the Storage subsystem.

I wonder if it would be simpler to consider each setting `StorageBackupsVolume` and `StorageImagesVolume` as their own isolated lifecycle, and for each at start up, edit etc call `pool.{Mount|Unmount}CustomVolume()` then if you're using the same volume for both LXD will call `pool.{Mount|Unmount}CustomVolume()` 2 times and the internal reference counting will ensure that the volume is only actually mounted and unmounted when needed.

tomponline: Also the existing `daemonStorageMove` function looks like its conflating a few different concerns.

It seems to be handling:

1. Activating the new volume (if needed).
2. Moving the content of the old volume to the new volume.
3. Deactivating the old volume (if needed).

I wonder if this can be refactored to be a bit clearer such that the activation/deactivation of the volumes is handled separately to moving the content between them.

lxd/patches.go
skozina: Sure, moved.

lxd/daemon_storage.go
skozina: Per below, lets just drop this function entirely and use built-in reference counting in the storage drivers.

lxd/daemon_storage.go
skozina: As agreed, I've removed the `numberOfVolumeReferences()` function and instead we allow `ErrInUse` to be returned from the `UnmountCustomVolume()` call.

The `daemonStorageMove()` function is complex, but I'm not sure it could be simplified much. The pseudo-code basically is:
```
if movingToExternalStorage() {
    mount(target_storage)
}
moveContent()
if movingFromExternalStorage() {
    unmount(source_storage)
}
```
We can isolate the mount/unmount calls to separate function, but that would just make `moveContent()` a separate function outside of the `daemonStorageMove()`. I'm not sure if that's much better. Honestly I'd just leave it as it is. WDYT?

tomponline: OK lets continue as-is for now and see how we get on.

tomponline (approved):
Thanks!

Once this lands in latest/edge, perhaps @mseralessandri and/or @escabo might like to take it for a test drive and see if there are any issues with it (I'm thinking especially around upgrading from an existing installation of 6.4 or less to latest/edge?

PR #15627, opened 2025-05-20
Revert "Storage: Update instance backup file when root vol pool gets updated"

roosterfish:
This reverts https://github.com/canonical/lxd/pull/15527.

Instead of updating each instance's backup config file when updating the pool, we should rather update the file for individual instances when starting them to not require mounting each instance's vol when updating the pool.


tomponline (approved):
ta

PR #15628, opened 2025-05-20
Use `net.JoinHostPort()` instead of `fmt.Sprintf("%s:%d", host, port)`

simondeziel:
This is in preparation of Go 1.25 new vet warning: https://tip.golang.org/doc/next#vet


tomponline (approved):
nice catch!

PR #15632, opened 2025-05-20
Improve CI setup for `liblxc` support

simondeziel:


.github/workflows/tests.yml
```diff
@@ -400,6 +400,16 @@ jobs:
...
 
+      - name: Prepare for running LXD daemon
+        run: |
+          set -eux
+
+          # Update env variables for liblxc
+          LIBLXC_ARCH_LIBS="$(readlink -e /home/runner/go/bin/liblxc/libs/*-linux-gnu)"
+          echo "CGO_LDFLAGS=${CGO_LDFLAGS} -L${LIBLXC_ARCH_LIBS}"       >> "${GITHUB_ENV}"
+          echo "LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${LIBLXC_ARCH_LIBS}" >> "${GITHUB_ENV}"
+          echo "LD_RUN_PATH=${LD_RUN_PATH}:${LIBLXC_ARCH_LIBS}"         >> "${GITHUB_ENV}"
```
tomponline, line 411: why do we need this here and on the earlier commit?

tomponline (approved):

.github/workflows/tests.yml
simondeziel: The whole block is needed because `liblxc` is saved to an arch-dependent path (`x86_64-linux-gnu/`) so it cannot be hardcoded the same as the `dqlite` one.

Now if you asked specifically about `LD_RUN_PATH`, I must admit I'm not so sure it's needed. Could be a leftover from a previous attempt at fixing some other problem. I'll check with Aleks.

PR #13458, opened 2024-05-08
Seccomp: Tighten container capability checks a bit to align with kernel behaviour

mihalicyn:
Let's tighten capability checks in mknod interception code to align this with what we have in the kernel. So, if mknod interception is enabled then only user with CAP_MKNOD in the container's initial user namespace can use it.

PR #15634, opened 2025-05-21
lxd/seccomp: allow unprivileged whiteout creation

mihalicyn:
Mimic the kernel behavior [1] and allow unprivileged whiteout creation in all cases, included nested user namespace.

It is a bugfix for 67b16ca026d ("lxd/seccomp/seccomp: tighten capability checks a bit") from PR #13458. We should consider backports to stable releases.

[1] https://github.com/torvalds/linux/commit/a3c751a50fe6bbe50eb7622a14b18b361804ee0c

Locally tested.

### DIFF
--- a/lxd/seccomp/seccomp.go
+++ b/lxd/seccomp/seccomp.go
@@ -64,6 +64,11 @@ static int seccomp_notify_get_sizes(struct seccomp_notif_sizes *sizes)
 	return 0;
 }
 
+static bool is_whiteout(dev_t dev, mode_t mode)
+{
+	return ((mode & S_IFMT) == S_IFCHR) && (dev == makedev(0, 0));
+}
+
 static int device_allowed(dev_t dev, mode_t mode)
 {
 	switch (mode & S_IFMT) {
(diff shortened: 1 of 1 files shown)


tomponline (approved):
Thanks!

What is a "whiteout"?

tomponline:
Thanks!

Can we add a test for this in the lxd test suite?

mihalicyn:
> Can we add a test for this in the lxd test suite?

I'll add a testcase to lxd-ci.

tomponline:
> > Can we add a test for this in the lxd test suite?
> 
> I'll add a testcase to lxd-ci.

Can we add it in the test suite in this repo so it counts towards code coverage? Or does it need a specific guest OS?

lxd/seccomp/seccomp.go
```diff
@@ -1359,7 +1364,7 @@ func (s *Server) doDeviceSyscall(c Instance, args *MknodArgs, siov *Iovec) int {
 		return int(-C.EPERM)
 	}
 
-	if !hasCapability {
+	if !bool(C.is_whiteout(args.cDev, args.cMode)) && !hasCapability {
```
simondeziel, line 1367: How about swapping the 2 conditions? `hasCapability` being a `bool` should be cheaper to check so a good candidate.

simondeziel (approved):
Looking forward to tests for this. AFAIK this will have to be in `lxd-ci` due to needing something more fancy than our usual `busybox` instance we using LXD test suite.

tomponline:
> Looking forward to tests for this. AFAIK this will have to be in `lxd-ci` due to needing something more fancy than our usual `busybox` instance we using LXD test suite.

@mihalicyn and I discussed adding them to the main test suite by way of a helper Go binary that makes the syscalls we want to test.

mihalicyn:
In case if you want to have it merged as a part of release prep, please do. I can submit test as a separate PR as it will take time.

PR #15641, opened 2025-05-22
Build tweaks

simondeziel:


Makefile
```diff
@@ -296,10 +296,18 @@ dist: doc
 	(cd $(TMP)/lxd-$(VERSION)/vendor/liblxc ; git show-ref HEAD | cut -d' ' -f1 > .gitref)
 
 	# Copy doc output
-	cp -r doc/_build $(TMP)/lxd-$(VERSION)/doc/html/
+	cp -r --preserve=mode doc/_build $(TMP)/lxd-$(VERSION)/doc/html/
 
 	# Assemble tarball
-	tar --exclude-vcs -C $(TMP) -zcf $(ARCHIVE).gz lxd-$(VERSION)/
+	$(eval SOURCE_EPOCH := $(shell TZ=UTC0 git log -1 --format=tformat:%cd --date=format:%Y-%m-%dT%H:%M:%SZ $(HASH)))
```
tomponline, line 302: This is very complicated and not clear what these lines are doing and why (beyond the stated goal in the commit message). Please can you add some comments to thd file explaining what we are doing here and why for future readers. Thanks

test/deps/import-busybox
```diff
@@ -316,6 +316,8 @@ user.foo: {{ config_get("user.foo", "_unset_") }}
 if __name__ == "__main__":
     if "LXD_DIR" in os.environ:
         lxd_socket = os.path.join(os.environ["LXD_DIR"], "unix.socket")
+    elif os.path.exists("/var/snap/lxd/common/lxd/unix.socket"):
```
tomponline, line 319: Whats ur thinking with this one, for a future plan regarding testing?

.github/workflows/tests.yml
```diff
@@ -672,6 +672,20 @@ jobs:
...
           sudo -E ./lxd-ui/tests/scripts/setup_test
 
+      - name: Test basic LXD functionality
+        shell: bash
+        run: |
+          set -eux
+          export PATH="/home/runner/go/bin:$PATH"
+
+          # launch a test instance
+          sudo -E ./test/deps/import-busybox --alias testimage
```
tomponline, line 695: Ah i see what the busybox change was for now.

test/deps/import-busybox
simondeziel: Mostly usability for devs. This mimics `pyLXD` behaviour of trying to guess where the socket is.

.github/workflows/tests.yml
simondeziel: Not exactly as this snippet has a `LXD_DIR` already exported and doesn't use the snap anyway. However it's when I did those steps locally that I figured I could improve the ease of use of this script.

Makefile
simondeziel: I meant to include a link to https://www.gnu.org/software/tar/manual/html_node/Reproducibility.html but this was lost during a rebase it seems. I added it back along with more explanation as that is indeed rather cryptic.

tomponline (approved):

PR #15645, opened 2025-05-23
Test: Add timeout to `ceph status` command

roosterfish:
In case you have leftover connection details present on your system, the 'ceph status' command will hang for a few minutes until it returns which causes unnecessary wait times when facing errors during test execution.


tomponline (approved):
Yeah i hit this the other day too, good thinking

PR #15646, opened 2025-05-23
Storage: Fix volume rename DB record update order

roosterfish:
In case a volume is attached to a profile (which is attached to instances), performing a rename operation of the volume was causing an error because it's new name was written to the DB only after trying to update the volume users (devices). 
As the latter operation is also performing a DB lookup, this was causing the following error in the test suite:

`Error: Device validation failed for "webbaz": Failed loading "custom/webbaz2" from project "default": Storage volume not found`

The PR changes the order in which DB entries gets updated and updates the test suite to remove the check for the false positive.


tomponline (approved):
Nice catch thanks

PR #15647, opened 2025-05-23
API: Add missing 'version' struct tag

roosterfish:
Not having the tag shouldn't be a big deal for those two fields as their default is 'version' anyway. But to stay consistent will all the other fields also add the 'version' struct tag for instance and custom vol backups.

The problematic piece is that the metadata used `Version` and not `version`.

Thanks @edlerd for raising this in https://github.com/canonical/lxd-ui/pull/1270#discussion_r2104038375.

### DIFF
--- a/shared/api/instance_backup.go
+++ b/shared/api/instance_backup.go
@@ -42,7 +42,7 @@ type InstanceBackupsPost struct {
 	// Example: 1
 	//
 	// API extension: backup_metadata_version
-	Version uint32
+	Version uint32 `json:"version" yaml:"version"`
 }
 
 // InstanceBackup represents a LXD instance backup.
--- a/doc/rest-api.yaml
+++ b/doc/rest-api.yaml
@@ -1859,11 +1859,6 @@ definitions:
         x-go-package: github.com/canonical/lxd/shared/api
     InstanceBackupsPost:
         properties:
-            Version:
-                description: What backup format version to use
-                example: 1
-                format: uint32
-                type: integer
             compression_algorithm:
                 description: What compression algorithm to use
                 example: gzip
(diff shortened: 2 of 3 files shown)


tomponline (approved):

tomponline:
That was close, if we had cut lxd 6.4 it would have had to stay like that. Thanks @edlerd

shared/api/instance_backup.go
```diff
@@ -42,7 +42,7 @@ type InstanceBackupsPost struct {
 	// Example: 1
 	//
 	// API extension: backup_metadata_version
-	Version uint32
+	Version uint32 `json:"version" yaml:"version"`
```
simondeziel, line 45: I feel that's an easy annotation to forget. Could we maybe rely on linters to catch those? https://github.com/go-simpler/musttag might fit the bill and is readily available through `golangci-lint`.

simondeziel: @roosterfish I think it would have caught it:

```
# add musttag to .golangci-lint.yml
# a19e2a15f8b6c949d496a22d5e9dc70bbf317ccd is the prior to changes missing the tags
$ golangci-lint run --timeout 5m --new --new-from-rev a19e2a15f8b6c949d496a22d5e9dc70bbf317ccd --whole-files --max-same-issues=30
...
                                                ^
lxd/storage/s3/policy.go:68:36: the given struct should be annotated with the `json` tag (musttag)
        err := json.Unmarshal(jsonPolicy, &policy)
                                          ^
lxd/storage/s3/policy.go:82:48: the given struct should be annotated with the `json` tag (musttag)
                err = json.Unmarshal([]byte(jsonRolePolicy), &rolePolicy)
                                                             ^
lxd/storage_volumes_backup.go:352:29: the given struct should be annotated with the `json` tag (musttag)
        err = json.Unmarshal(body, &req)
                                   ^
20 issues:
* musttag: 20
```

And this bit leads to the struct with `Version uint32` missing the tag.
```
lxd/storage_volumes_backup.go:352:29: the given struct should be annotated with the `json` tag (musttag)
        err = json.Unmarshal(body, &req)
```

roosterfish: Does it test both `json` and `yaml` tags? Overall great idea, would you mind adding a PR for it?

Issue #15648, opened 2025-05-23
`musttag` linter should be abled in `golangci-lint`


simondeziel:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

The `musttag` linter could have caught a missing tag on a struct newly introduced. See https://github.com/canonical/lxd/pull/15647

### Describe the solution you'd like

`musttag` should be enabled and all the code (not just new) should be audited and fixed.

### Additional context

Enabling the [`tagalign`](https://golangci-lint.run/usage/linters/#tagalign) linter at the same time might be useful as it improves readability.

simondeziel:
> ### Additional context
> 
> Enabling the [`tagalign`](https://golangci-lint.run/usage/linters/#tagalign) linter at the same time might be useful as it improves readability.

The only remaining bit was for `tagalign` which was NACK'ed for now: #15769
As such, closing.

PR #15649, opened 2025-05-23
Add `musttag` linter now to catch newly missed tags

simondeziel:
This is the first step to addressing https://github.com/canonical/lxd/issues/15648. It however does not deal with fixing existing violation as our `test/lint/golangci.sh` script only looks at new code.

Also update `golangci-lint` itself as it ships an updated `musttag` version.


tomponline (approved):
Cool!

PR #15652, opened 2025-05-23
Makefile: use `git rev-parse HEAD` to get current commit hash

simondeziel:
Kadin found out that PR tests would be run with a `lxc` binary built from
sources that didn't match those associated with the PR. He then tracked the
issue down to a faulty commit where the archive produced by `make dist` would
use the sources from the main branch.

The problem is that `git show-ref --hash HEAD` gives the commit hash of the
origin/main's HEAD, not the one corresponding to the current branch.

Switch to using `git rev-parse HEAD` that is specifically made to get the hash
of the current commit.

Also, rename the variable to `COMMIT_HASH` to avoid shadowing the exising
`HASH` one.

Fixes https://github.com/canonical/lxd/commit/113279fb493cc638e37906c5b1afc81e0c8e888a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -280,9 +280,9 @@ dist: doc
 
 	# Create build dir
 	$(eval TMP := $(shell mktemp -d))
-	$(eval HASH := $(shell git show-ref --hash HEAD))
-	git archive --prefix=lxd-$(VERSION)/ $(HASH) | tar -x -C $(TMP)
-	echo $(HASH) > $(TMP)/lxd-$(VERSION)/.gitref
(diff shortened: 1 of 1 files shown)


kadinsayani (approved):
LGTM! Thanks!

kadinsayani (approved):

tomponline (approved):
Nice catch @kadinsayani !

simondeziel:
> Nice catch @kadinsayani !

Indeed, he did all the work between noticing, finding and figuring out a fix ;)

FYI, it also outlined that the `.gitref` we used to save for information purposes was wrong.

Issue #14768, opened 2025-01-11
LXD snap hold on to old swapfile


simondeziel:
I noticed that freeing the GHA ephemeral disk sometimes failed due to `Device or resource busy`. After some investigation, it seems to be due to LXD holding the block device active despite having `swapoff` the swapfile and `umount` the underlying FS.

Here's the simplest reproducer I could come up with:

Prepare the VM with an extra disk to hold onto the swapfile:

```
$ lxc init ubuntu-minimal-daily:24.04 v1 --vm
$ lxc storage volume create default v1-sdb --type block size=1GiB
$ lxc storage volume attach default v1-sdb v1
$ lxc start v1
```

Setup swapfile on extra disk:

```
$ lxc shell v1
root@v1:~# lsblk
NAME    MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
sda       8:0    0   40G  0 disk 
├─sda1    8:1    0   39G  0 part /
├─sda14   8:14   0    4M  0 part 
├─sda15   8:15   0  106M  0 part /boot/efi
└─sda16 259:0    0  913M  0 part /boot
sdb       8:16   0    1G  0 disk 
root@v1:~# mkfs.ext4 /dev/sdb
mke2fs 1.47.0 (5-Feb-2023)During CI debugging, I
Discarding device blocks: done                            
Creating filesystem with 262144 4k blocks and 65536 inodes
Filesystem UUID: a0d0f62b-d0f6-4b43-90a5-6a2cc07ff8ed
Superblock backups stored on blocks: 
	32768, 98304, 163840, 229376

Allocating group tables: done                            
Writing inode tables: done                            
Creating journal (8192 blocks): done
Writing superblocks and filesystem accounting information: done

root@v1:~# mount /dev/sdb /mnt/
root@v1:~# fallocate -l 512M /mnt/swapfile
root@v1:~# chmod 0000 /mnt/swapfile 
root@v1:~# mkswap /mnt/swapfile 
Setting up swapspace version 1, size = 512 MiB (536866816 bytes)
no label, UUID=0b1d0374-a33e-43dc-b1b1-132b49ff37ee
root@v1:~# swapon /mnt/swapfile 
```

Trying to discard the extra disk but failing due to LXD holding on to it:

```
root@v1:~# snap install lxd --channel latest/edge
2025-01-11T19:56:23Z INFO Waiting for automatic snapd restart...
lxd (edge) git-0d2dc0a from Canonical✓ installed
[...]

PR #15655, opened 2025-05-24
Daily caching of external assets

simondeziel:
Also tweak the cluster tests to **not** reclaim disk space as they don't use much so it's best to skip this operation that is sometimes very slow (~2 minutes). This required working around https://github.com/canonical/lxd/issues/14768


tomponline (approved):
Thanks!

PR #15657, opened 2025-05-26
API: Add missing struct tags (musttag)

roosterfish:
Whilst working on https://github.com/canonical/lxd/pull/15523 the linter started to complain about more structs not having proper tags defined (due to https://github.com/canonical/lxd/pull/15649).

The PR adds the missing tags for the structs used inside the storage backend package.

### DIFF
--- a/shared/api/instance.go
+++ b/shared/api/instance.go
@@ -227,7 +227,7 @@ type InstanceRebuildPost struct {
 //
 // API extension: instances.
 type Instance struct {
-	WithEntitlements `yaml:",inline"`
+	WithEntitlements `yaml:",inline"` //nolint:musttag
 
 	// Instance name
 	// Example: foo
--- a/shared/api/profile.go
+++ b/shared/api/profile.go
@@ -41,7 +41,7 @@ type ProfilePut struct {
 //
 // swagger:model
 type Profile struct {
-	WithEntitlements `yaml:",inline"`
+	WithEntitlements `yaml:",inline"` //nolint:musttag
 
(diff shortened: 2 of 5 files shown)


roosterfish:
> Would it be clearer to explicity tag the fields that are exported even in there capitalised format to make it clear they are part of api rather than nolint, or is this a linter error?

You mean something like this instead of `//nolint:musttag`?

```go
type Instance struct {
	WithEntitlements `json:"WithEntitlements" yaml:",inline"`
```

Problem is as soon as `WithEntitlements` gets a JSON tag, the marshalled JSON isn't anymore nested/inline:

```json
{
	"WithEntitlements": {
		"access_entitlements": [
			"can_view"
		]
	},

```

tomponline:
> > Would it be clearer to explicity tag the fields that are exported even in there capitalised format to make it clear they are part of api rather than nolint, or is this a linter error?
> 
> You mean something like this instead of `//nolint:musttag`?
> 
> ```go
> type Instance struct {
> 	WithEntitlements `json:"WithEntitlements" yaml:",inline"`
> ```
> 
> Problem is as soon as `WithEntitlements` gets a JSON tag, the marshalled JSON isn't anymore nested/inline:
> 
> ```json
> {
> 	"WithEntitlements": {
> 		"access_entitlements": [
> 			"can_view"
> 		]
> 	},
> ```
> 
> And there is no `inline` tag for JSON.

Ah ok got you. Makes sense how it is.

simondeziel (approved):

PR #15595, opened 2025-05-18
Rework server-side migration (from Incus)

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/15525.

This is a significant rework of server-side migration cherry-picked from https://github.com/lxc/incus/pull/727, improving instance migration logic. 

Key changes include refactoring the instance migration logic and adding support for changing one or more of name, project, location and storage pool.

`instancePost` now handles all request validation, and `migrateInstance` acts as an entry point that calls the following functions (in order):

`migrateFromOfflineNode`: Manages the specific case of migrating an instance from a server that is offline but uses shared storage.
`prepareMigration`: Prepares the target instance's configuration by applying overrides for config, devices, profiles, and storage pools.
`migrateHandleRename`: Handles the renaming of an instance during a migration. Note that simple rename requests (rename only, no pool, project, or location changes) are handled by `instancePost`.
`migrateLocal`: Handles local migrations, where the instance moves between projects or storage pools on the same server.
`migrateRemote`: Handles remote migrations, where the instance is moved to a different cluster member.

`migrateFromOfflineNode` simply updates the database records, imports the instance into the storage, and performs any renames. The rest of the functions sequentially update the instance's properties, first preparing the target configuration, then handling any renames, followed by local moves (project or storage pool changes), and finally executing a remote migration to a new cluster member if required. Therefore, clearing certain parts of the request throughout the steps allows for indication of "what is left to do".

This also includes a bug-fix cherry-pick from https://github.com/lxc/incus/pull/1019 (https://github.com/lxc/incus/pull/1019/commits/c84cd644a7ea9e9e17aff9577a884634ec4ab310) that correctly sets the target project in `migrateLocal`.

PR #15663, opened 2025-05-26
Defend against path traversal attacks

kadinsayani:


simondeziel (commented):
Instead of `IsSafePath()`, I'm wondering if we would be better off with a `IsChildDir()` or something similar. Let's have a chat in our 1:1.

lxd/storage/backend_lxd.go
```diff
@@ -518,8 +518,27 @@ func (b *lxdBackend) ensureInstanceSymlink(instanceType instancetype.Type, proje
 		return errors.New("Instance must not be snapshot")
 	}
 
+	// Defend against path traversal attacks.
+	if !shared.IsFileName(instanceName) {
```
simondeziel, line 522: Could we use `instancetype.ValidName(instanceName, false)` instead?

lxd/storage/backend_lxd.go
```diff
@@ -518,8 +518,27 @@ func (b *lxdBackend) ensureInstanceSymlink(instanceType instancetype.Type, proje
...
+		return fmt.Errorf("Invalid instance name %q", instanceName)
+	}
+
+	if !shared.IsFileName(projectName) {
```
simondeziel, line 526: Could we use ` projecthelpers.ValidName(projectName)` instead?

lxd/storage/backend_lxd.go
```diff
@@ -518,8 +518,27 @@ func (b *lxdBackend) ensureInstanceSymlink(instanceType instancetype.Type, proje
...
 
+	// Defend against path traversal attacks.
+	baseDir := shared.VarPath()
+	if !shared.IsPathWithinBaseDir(baseDir, symlinkPath) {
+		return fmt.Errorf("Invalid symlink path %q", symlinkPath)
+	}
+
+	if !shared.IsPathWithinBaseDir(baseDir, mountPath) {
+		return fmt.Errorf("Invalid mount path %q", mountPath)
+	}
```
simondeziel, line 540: I'm wondering if those would not be redundant once the instance and project names were validated?

lxd/storage/drivers/driver_cephfs_volumes.go
```diff
@@ -379,6 +379,11 @@ func (d *cephfs) UnmountVolume(vol Volume, keepBlockDev bool, op *operations.Ope
 
 // RenameVolume renames the volume and all related filesystem entries.
 func (d *cephfs) RenameVolume(vol Volume, newVolName string, op *operations.Operation) error {
+	// Defend against path traversal attacks.
+	if !shared.IsFileName(newVolName) {
```
simondeziel, line 383: There is a `storagePools.ValidVolumeName()` utility function but it does not reject `..` so maybe check both?

lxd/storage/backend_lxd.go
tomponline: > CodeQL doesn't like this either. 

If its an instance name we should be using `instancetype.ValidName(instanceName, false)`, what is the reason codeql thinks thats an issue?

tomponline: If we take calls to `ensureInstanceSymlink`, which specific routes to this function are being identified as issues?

Is it `CreateInstanceFromBackup` and `RenameInstance` only?

kadinsayani: The two problematic paths are:
`instances_post.go: createFromBackup()`
`instance_post.go: io.ReadAll(r.Body) -> Rename() -> RenameInstance()`

kadinsayani: I'll try using `instancetype.ValidName(instanceName, false)` higher up in the call stack.

tomponline: Right, thanks, so yeah we should be validating the input as early as possible.

lxd/backup/backup_instance.go
```diff
@@ -78,6 +79,16 @@ func (b *InstanceBackup) Rename(newName string) error {
 		}
 	}
 
+	// Defend against path traversal attacks.
```
tomponline: rather than checking the `name` is safe by checking the resulting paths, is the reason you've not validated the `newName` itself and then used the sanitised version of that to build the paths?

lxd/storage/drivers/driver_cephfs_volumes.go
tomponline, line 1: Instead of validating this at a per-driver level we should rather ensure rather only sanitised `newVolName` makes it into the function call.

simondeziel:
It seems like doing the verification earlier gave a massive payoff, 42 issues fixed: https://github.com/canonical/lxd/security/code-scanning?query=pr%3A15663+tool%3ACodeQL+is%3Aclosed

lxd/storage/backend_lxd.go
simondeziel: > `instancetype.ValidName(instanceName, false)` isn't sufficient for defending against path traversal. There aren't checks for path separators

https://github.com/canonical/lxd/blob/main/shared/validate/validate.go#L735 only allows for the harmless set.

Could it be that CodeQL isn't smart enough to understand the regex does the job? If so, adding a check before the regex that specifically refuses `/` and `\` might do.

> or `..`.

DNS hostnames should not have any `.` in them as that's the label separator. Could we clue CodeQL in with a simple `strings.ContainsAny(name, './\')` before the more thorough `regex` check?

kadinsayani: That could be the case. 

I've stepped through the flagged paths and we are validating `newName` or it's ancestral instance name in most cases, but CodeQL isn't "counting" this as path traversal defense.

I'll give that a try, thanks :)

kadinsayani:
Most of the alerts flagged by #15595 appear to be false positives, as we already call `instancetype.ValidName()` higher up in the call stack. I tried storing the sanitized instance name in a separate variable, but that didn’t satisfy CodeQL. I've also tried modifying `instancetype.ValidName()` to return a sanitized result and call `shared.IsFileName()`, but this didn't work either.

I’ll go ahead and close this for now. I’ll ensure proper validation is included in #15595. Since the alerts seem to be triggered by the size of the diff (this feature seems to be problematic for others too: https://github.com/actions/toolkit/issues/457), I think we’re okay to ignore them.

simondeziel:
@kadinsayani I'm surprised to see this PR closed. Did it not "address" 42 CodeQL alerts [as reported](https://github.com/canonical/lxd/security/code-scanning?query=pr%3A15663+tool%3ACodeQL+is%3Aclosed) ?

kadinsayani:
> @kadinsayani I'm surprised to see this PR closed. Did it not "address" 42 CodeQL alerts [as reported](https://github.com/canonical/lxd/security/code-scanning?query=pr%3A15663+tool%3ACodeQL+is%3Aclosed) ?

It did, but we were "double-validating", and in some cases, "triple-validating". The name already gets validated higher up in the call stack by `instancetype.ValidName()`, CodeQL just isn't picking this up.

Most of the 42 closed were coming from adding `IsPathWithinSafeDir()`, but as @tomponline mentioned, we shouldn't need this if the new names are properly validated.

kadinsayani:
For example, this check is redundant: https://github.com/canonical/lxd/commit/bf6536ba15fd6781122c1431891749dce1b4ac62#diff-2ab8ba8696630342b067053a70d715c9641767a5ad7d85c5f526e1fe3e098bb9.

The call stack is: `RenameInstance()->RenameVolume()->genericVFSRenameVolume()`. The new volume name is [derived from the new instance name](https://github.com/canonical/lxd/blob/cf419f5e963f91e1ce83b3fff066eb980f646e8c/lxd/storage/backend_lxd.go#L2821), which is already validated:

https://github.com/canonical/lxd/blob/cf419f5e963f91e1ce83b3fff066eb980f646e8c/lxd/storage/backend_lxd.go#L2741-L2741

simondeziel:
> > @kadinsayani I'm surprised to see this PR closed. Did it not "address" 42 CodeQL alerts [as reported](https://github.com/canonical/lxd/security/code-scanning?query=pr%3A15663+tool%3ACodeQL+is%3Aclosed) ?
> 
> It did, but we were "double-validating", and in some cases, "triple-validating". The name already gets validated higher up in the call stack by `instancetype.ValidName()`, CodeQL just isn't picking this up.

I see. It seems that TiCS is also not tracking things properly which means we'll have to live with a worst score, oh well, you tried so thank you for that!

> Most of the 42 closed were coming from adding `IsPathWithinSafeDir()`, but as @tomponline mentioned, we shouldn't need this if the new names are properly validated.

Hmm, I would have thought such helper to be handy elsewhere like before doing some path existence checks or `rm -rf`. Anyway, if need be we can always resurrect it.

PR #15665, opened 2025-05-26
Update install docs to split between build and runtime/test depencies

simondeziel:


doc/installing.md
```diff
@@ -260,14 +255,22 @@ To run the test suite, you'll also need:
 
 ```bash
 sudo apt install \
```
tomponline, line 259: Previously this section was intended just to contain packages needed to run the test suite (as opposed to those needed to build & run LXD), but now the packages needed to run LXD have been moved to the "run the test suite" section.

So whilst I agree its good to separate build & run packages, I think we need to make clearer which packages are needed to run LXD, and which additional ones are needed to run the test suite.

simondeziel: Sounds good. I'll make that 3 separated sections.

tomponline (approved):
ta

PR #15666, opened 2025-05-27
idmap: Add missing struct tags (musttag)

roosterfish:
Required as part of https://github.com/canonical/lxd/pull/15523

The lowercase representation can be used as it won't break existing `volatile.idmap.*` config keys. When doing `json.Unmarshal()`, the correct config key is pulled in any case:

* If the JSON key is in uppercase, it will get unmarshalled accordingly even if the lowercase tag is set
* If the JSON key is in lowercase, it matches the struct tag and will work as expected
* If the JSON key is in lowercase, but LXD expects the uppercase (no tag set, old version), it still unmarshalls accordingly

A patch is not required as all of the containers `volatile.idmap.*` config keys are getting updated progressively when LXD rewrites the keys.

### DIFF
--- a/lxd/idmap/structs.go
+++ b/lxd/idmap/structs.go
@@ -2,14 +2,14 @@ package idmap
 
 // IdmapEntry is a single idmap entry (line).
 type IdmapEntry struct {
-	Isuid    bool
-	Isgid    bool
-	Hostid   int64 // id as seen on the host - i.e. 100000
-	Nsid     int64 // id as seen in the ns - i.e. 0
-	Maprange int64
+	Isuid    bool  `json:"Isuid"`
+	Isgid    bool  `json:"Isgid"`
+	Hostid   int64 `json:"Hostid"` // id as seen on the host - i.e. 100000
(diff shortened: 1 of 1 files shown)


tomponline:
> Required as part of #15523
> 
> The lowercase representation can be used as it won't break existing `volatile.idmap.*` config keys. When doing `json.Unmarshal()`, the correct config key is pulled in any case:
> 
> * If the JSON key is in uppercase, it will get unmarshalled accordingly even if the lowercase tag is set
> * If the JSON key is in lowercase, it matches the struct tag and will work as expected
> 
> A patch is not required as all of the containers `volatile.idmap.*` config keys are getting updated progressively when LXD rewrites the keys.

Something you haven't considered is that this will break downgrades in an LTS series if the volatile keys have been updated to user lowercase and then switch back to a version that expects upper case.

I think we should keep the tags as upper case values.

roosterfish:
It also matches what we did for the migration info struct in https://github.com/canonical/lxd/pull/15657.

tomponline (approved):
Ta

PR #15669, opened 2025-05-27
test/main: detect cloud-init presence and wait for it

simondeziel:


tomponline (commented):
when would cloud-init be relevant for the lxd test suite (that uses busy box)?

PR #15670, opened 2025-05-27
API: Revert linter fixes and use legacy tags instead

roosterfish:
In the PR https://github.com/canonical/lxd/pull/15657 we were trying to satisfy the new musttag linter rules. However this breaks the migration with older versions of LXD.

To stay on the safe side, explicitly define the JSON tags to use the uppercase representation.


tomponline (commented):
Were we able to edit backups via the yaml editor? What im asking is whether the yaml tags are an API break too?

tomponline (approved):

PR #15673, opened 2025-05-28
Linter: Fix remaining `musttag` issues across the entire codebase

roosterfish:
Running `golangci-lint run --enable-only musttag` doesn't report any outstanding structs after those commits.


roosterfish:
@tomponline I think this is fine to be landed after the 6.4 release.

tomponline:
@roosterfish it will have to be now anyway as i just merged the release commit.

PR #15679, opened 2025-05-28
github: provide pkg-config path to `lxc.pc` for TiCS run

simondeziel:


tomponline:
Is this needed?

simondeziel:
> Is this needed?
I'm trying to address this from TiCS folks:
lxc pc-tics

tomponline:
Btw i use LD_RUN_PATH when i build lxd so that the linker knows where to locate the libs i want lxd to use without any pkg-config or runtime env vars. Ive found this useful when using sudo to launch lxd too.

PR #15682, opened 2025-05-28
Network: Show info for macvlan networks

nmezhenskyi:
Fixes https://github.com/canonical/lxd/issues/15479.

This is to prevent `lxc network info <net>` failing when used on a macvlan network.

Example output:
```
root@vm1:~/lxd# lxc network info net1
Name: net1
MAC address: ab:cd:ef:12:34:56
MTU: 1500
State: up
Type: broadcast

Network usage:
  Bytes received: 0B
  Bytes sent: 0B
  Packets received: 0
  Packets sent: 0
```

### DIFF
--- a/lxd/network/driver_macvlan.go
+++ b/lxd/network/driver_macvlan.go
@@ -2,9 +2,13 @@ package network
 
 import (
 	"fmt"
+	"math"
+	"net/http"
+	"strconv"
 
 	"github.com/canonical/lxd/lxd/cluster/request"
 	"github.com/canonical/lxd/lxd/db"
+	"github.com/canonical/lxd/lxd/resources"
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/logger"
 	"github.com/canonical/lxd/shared/revert"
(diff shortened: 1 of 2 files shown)


lxd/network/driver_macvlan.go
```diff
@@ -21,6 +24,44 @@ func (n *macvlan) DBType() db.NetworkType {
...
+
+		// In all other cases, return the original error.
+		return nil, err
+	}
+
+	var mtu int
+
+	configMtu := n.config["mtu"]
+	if configMtu != "" {
+		mtu, err = strconv.Atoi(configMtu)
```
simondeziel, line 47:
```suggestion
		mtu, err := strconv.ParseUint(d.config["mtu"], 10, 32)
```

To get a `uint` and for consistency with other places where we handle the MTU.

simondeziel (commented):
LGTM, just a tiny comment.

lxd/network/driver_macvlan.go
```diff
@@ -21,6 +25,53 @@ func (n *macvlan) DBType() db.NetworkType {
...
+		} else if uintMTU > 0 {
+			mtu = int(uintMTU)
+		} else {
+			mtu = 1500
+		}
+	} else {
+		mtu = parentState.Mtu
+	}
+
+	return &api.NetworkState{
```
tomponline, line 65: Is there a reason you didnt use the state returned by `resources.GetNetworkState` directly, and just modify the MTU field if the network config overrode the MTU from the parent?

Did you explicitly want to exclude the Counters and Addresses?

test/suites/container_devices_nic_macvlan.sh
```diff
@@ -104,8 +104,13 @@ test_container_devices_nic_macvlan() {
 
   echo "==> Test using macvlan network."
 
-  echo "==> Create macvlan network and add NIC device using that network."
+  echo "==> Create macvlan network."
   lxc network create "${ctName}net" --type=macvlan parent="${ctName}"
+
+  echo "==> Check that lxc network info succeeds for macvlan network."
+  lxc network info "${ctName}net"
```
tomponline, line 111: Can we also test for the MTU override logic?

lxd/network/driver_macvlan.go
nmezhenskyi: > Did you explicitly want to exclude the Counters and Addresses?

Yes, I thought these might be confusing for the user since they're related to the parent interface. As a user, I'd probably expect the Addresses to show all associated container/vm NIC IPs. And I saw that we also explicitly exclude the Counters for `ovn` networks.  

What are your thoughts on this?

test/suites/container_devices_nic_macvlan.sh
nmezhenskyi: Yes, will do

lxd/network/driver_macvlan.go
tomponline: > And I saw that we also explicitly exclude the Counters for ovn networks.

That is because OVN networks dont have a parent as such.

> I thought these might be confusing for the user since they're related to the parent interface. 

Yes I suppose so, the addresses and counters of the macvlan parent, whilst might be partially relevant/associated to the NICs hanging off of it are not exclusively related to the macvlan NICs, so I think your decision makes sense.


Thanks!

PR #15686, opened 2025-05-29
Apply `modernize` changes

simondeziel:
Those changes were done mechanically using [`modernize`](https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize):

```
$ go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix -test ./...

# Apply changes related to `slices`
$ for f in $(git ls-files -m); do p="$(echo "${f}" | sed 's/\.go$//')"; git diff "${f}" | grep -wF slices && git commit -spm "${p}: use \`slices\` operations where applicable (modernize)" "${f}"; done

...

# Apply changes related to `maps`
 for f in $(git ls-files -m); do p="$(echo "${f}" | sed 's/\.go$//')"; git diff "${f}" | grep -wF maps && git commit -spm "${p}: use \`maps\` operations where applicable (modernize)" "${f}"; done
```


kadinsayani (commented):
Nice one! Can we run `modernize` and flag suggestions in our CI?

kadinsayani (approved):
Thanks!

tomponline (commented):
Any API breaks in the shared/* packages?

simondeziel:
> > Any API breaks in the shared/* packages?
> 
> There is a function name change in `shared/osarach`: [f4b648a](https://github.com/canonical/lxd/commit/f4b648aae91af75da0914476e2aa09453a6a1041). It looks like all references have been updated though.
> 
> Also, `Get()` is no longer exported by `shared/proxy.go`: [db88729](https://github.com/canonical/lxd/commit/db88729e1ba841ed6863b087b4c491e5ad999943).

Should I revert those and silence the linter(s) to preserve compatibility with some downstream modules?

tomponline:
> > > Any API breaks in the shared/* packages?
> > 
> > 
> > There is a function name change in `shared/osarach`: [f4b648a](https://github.com/canonical/lxd/commit/f4b648aae91af75da0914476e2aa09453a6a1041). It looks like all references have been updated though.
> > Also, `Get()` is no longer exported by `shared/proxy.go`: [db88729](https://github.com/canonical/lxd/commit/db88729e1ba841ed6863b087b4c491e5ad999943).
> 
> Should I revert those and silence the linter(s) to preserve compatibility with some downstream modules?

Yes for now. We shouldnt sneak api breaks into a mechanical code improvement pr, but rather make it an explicit decision.

Anything else in here that could potentially affect rest clients or consumers of our sdk?

simondeziel:
> > > > Any API breaks in the shared/* packages?
> > > 
> > > 
> > > There is a function name change in `shared/osarach`: [f4b648a](https://github.com/canonical/lxd/commit/f4b648aae91af75da0914476e2aa09453a6a1041). It looks like all references have been updated though.
> > > Also, `Get()` is no longer exported by `shared/proxy.go`: [db88729](https://github.com/canonical/lxd/commit/db88729e1ba841ed6863b087b4c491e5ad999943).
> > 
> > 
> > Should I revert those and silence the linter(s) to preserve compatibility with some downstream modules?
> 
> Yes for now. We shouldnt sneak api breaks into a mechanical code improvement pr, but rather make it an explicit decision.

Ack, I didn't try to sneak those in I was just trying to tame the linters without thinking about API breaks I must admit.

> Anything else in here that could potentially affect rest clients or consumers of our sdk?

Does that mean I need to pay attention to changes under `client/` (none in this PR) and `shared/` (a few some of which I'll revert) ?

tomponline:
> > > > > Any API breaks in the shared/* packages?
> > > > 
> > > > 
> > > > There is a function name change in `shared/osarach`: [f4b648a](https://github.com/canonical/lxd/commit/f4b648aae91af75da0914476e2aa09453a6a1041). It looks like all references have been updated though.
> > > > Also, `Get()` is no longer exported by `shared/proxy.go`: [db88729](https://github.com/canonical/lxd/commit/db88729e1ba841ed6863b087b4c491e5ad999943).
> > > 
> > > 
> > > Should I revert those and silence the linter(s) to preserve compatibility with some downstream modules?
> > 
> > 
> > Yes for now. We shouldnt sneak api breaks into a mechanical code improvement pr, but rather make it an explicit decision.
> 
> Ack, I didn't try to sneak those in I was just trying to tame the linters without thinking about API breaks I must admit.
> 
> > Anything else in here that could potentially affect rest clients or consumers of our sdk?
> 
> Does that mean I need to pay attention to changes under `client/` (none in this PR) and `shared/` (a few some of which I'll revert) ?

Ah yeah "sneak" was a loaded term sorry, "bundle" wouldve been better. My meaning was we should make changes like that intentionally and explicitly.

Yes we should be mindful of entities in shared and client.

simondeziel:
@tomponline I've undone the unintended API breakage in `shared/osarch/architectures.go`.

I reviewed the output of `git diff main.. shared/` and don't think there are any API breaking changes. Of course please take a second look.

PR #15694, opened 2025-05-31
test/mini-oidc/storage: use stdlib `log/slog`

Juneezee:
The `log/slog` package is available in the Go standard library since Go 1.21.

Reference: https://go.dev/doc/go1.21#slog


tomponline (approved):
Thanks!

PR #15706, opened 2025-06-02
OIDC: Use client secret for authorization code flow

nmezhenskyi:
This addition allows using client secret for OIDC authorization code flow. For this to work, the client type has to be set to ["confidential"](https://tools.ietf.org/html/rfc6749#section-2.1) in the IdP.  

### Changes:
- Adds `oidc.client.secret` configuration key for LXD to use for OpenID Connect configuration.
- If set, `oidc.client.secret` will be used for OIDC authorization code flow.

### Notes:
- If IdP is configured to use the "confidential" client type, CLI clients of LXD will no longer be able to authenticate using OIDC. This can be mitigated in the future by implementing the support for two separate OIDC clients in LXD.

### DIFF
--- a/doc/authentication.md
+++ b/doc/authentication.md
@@ -136,6 +136,11 @@ The LXD client then retrieves and stores an access token, which it provides to L
 The identity provider might also provide a refresh token.
 In this case, the LXD client uses this refresh token to attempt to retrieve another access token when the current access token has expired.
 
+```{warning}
+Only set `oidc.client.secret` if required by the Identity Provider. Once set, this key allows the LXD UI client to authenticate.
+However, the secret is not shared with other LXD clients (such as the LXD CLI).
+```
+
 When an OIDC client initially authenticates with LXD, it does not have access to the majority of the LXD API.
 OIDC clients must be granted access by an administrator, see {ref}`fine-grained-authorization`.
 
--- a/doc/server.md
+++ b/doc/server.md
@@ -54,6 +54,11 @@ The following server options configure external user authentication through {ref
     :end-before: <!-- config group server-oidc end -->
 ```
 
+```{important}
+Setting `oidc.client.secret` may prevent CLI clients from authenticating depending on the Identity Provider policies.
+Set this key only if required by the Identity Provider.
+```
+
 (server-options-cluster)=
 ## Cluster configuration
 


doc/metadata.txt
```diff
@@ -4939,6 +4939,13 @@ This value is required by some providers.
 
 ```
 
+```{config:option} oidc.client.secret server-oidc
+:scope: "global"
+:shortdesc: "OpenID Connect client secret"
+:type: "string"
+
```
simondeziel, line 4952: We are not consistently doing it but I think this empty line could be dropped.

nmezhenskyi: Hm, I think I just ran `make update-metadata` to generate this file after adding a new key to `lxd/cluster/config/config.go`:

```go
// lxdmeta:generate(entities=server; group=oidc; key=oidc.client.secret)
//
// ---
//  type: string
//  scope: global
//  shortdesc: OpenID Connect client secret
"oidc.client.secret": {},
```

I could do this instead:

```go
// lxdmeta:generate(entities=server; group=oidc; key=oidc.client.secret)
//
// ---
//  type: string
//  scope: global
//  shortdesc: OpenID Connect client secret
//  longdesc: OpenID Connect client secret
"oidc.client.secret": {},
```

It will produce this in `doc/metadata.txt` (instead of the empty line, it'll have the description):

```
```{config:option} oidc.client.secret server-oidc
:scope: "global"
:shortdesc: "OpenID Connect client secret"
:type: "string"
OpenID Connect client secret
```

Should I do this instead? I'm not sure how to skip the line completely.

simondeziel: Ah sorry, what you did is likely fine, sorry for the noise.

nmezhenskyi:
Hey @markylaing ,

Yes, I tested this PR with Keycloak locally and it is working for UI, but there is a major limitation for CLI.  

In order to enable Keycloak to generate and use client secret, the OIDC client type has to be set to "confidential" (set Client authentication to "On" in Keycloak client settings). Once that is done and `oidc.client.secret` is set on the LXD server, the authorization flow works correctly for the UI. However, the CLI client won't be able to authenticate anymore because Keycloak would still require the client secret, even though the CLI uses the Device Authorization flow and the `Device Authorization Grant` is enabled in Keycloak settings. I couldn't find a way to override this behavior in Keycloak.  

Keycloak seems to only allow Device authorization if the OIDC client type is "public" and, therefore, has no client secret. Do you know if it is possible to override this?  

I want to demo this for everyone on Thursday and describe the limitation.  

Regarding the e2e UI tests, I haven't looked into it yet but will do this.

markylaing:
> the CLI client won't be able to authenticate anymore because Keycloak would still require the client secret

I had a feeling this might happen :cry:

The [specification](https://datatracker.ietf.org/doc/html/rfc8628) for the device authorization grant makes no mention of a client secret, but it seems different IdPs each have their own interpretation of the specification (we've been bitten by this before). 

I think a more typical set up, and probably one that should have been considered when OIDC was originally added to LXD, is to have two separate clients defined in the IdP. These would be e.g. "LXD UI" with a client secret configured, and "LXD Device" with the device flow configured. Then when performing the code flow, LXD would use the client id of "LXD UI", but it would send the client ID of "LXD Device" to clients attempting to authenticate with OIDC. The reasoning here is that LXD itself is one client, but the CLI or any other application is another third party "client" of the IdP.

Right now I'm thinking we can go ahead with the PR in it's current form, with some documentation to explain the limitations (this can be part of the description of the config key which will make it's way into the docs via `make update-metadata`). Then we can consider how to tackle this problem later (we can create an issue for it).

markylaing:
> I think a more typical set up, and probably one that should have been considered when OIDC was originally added to LXD, is to have two separate clients defined in the IdP. These would be e.g. "LXD UI" with a client secret configured, and "LXD Device" with the device flow configured. Then when performing the code flow, LXD would use the client id of "LXD UI", but it would send the client ID of "LXD Device" to clients attempting to authenticate with OIDC. The reasoning here is that LXD itself is one client, but the CLI or any other application is another third party "client" of the IdP.

PS this is the approach anbox are taking - one client for the CLI and another for the UI.

nmezhenskyi:
> I think a more typical set up, and probably one that should have been considered when OIDC was originally added to LXD, is to have two separate clients defined in the IdP. These would be e.g. "LXD UI" with a client secret configured, and "LXD Device" with the device flow configured. Then when performing the code flow, LXD would use the client id of "LXD UI", but it would send the client ID of "LXD Device" to clients attempting to authenticate with OIDC. The reasoning here is that LXD itself is one client, but the CLI or any other application is another third party "client" of the IdP.

Yes, this makes sense. I thought about this approach as well during testing.

> Right now I'm thinking we can go ahead with the PR in it's current form, with some documentation to explain the limitations (this can be part of the description of the config key which will make it's way into the docs via make update-metadata). Then we can consider how to tackle this problem later (we can create an issue for it).

Sounds good. I'll make a demo to gather additional feedback from everyone and mention that we can consider implementing the approach with two separate clients in the future.

markylaing:
> Sounds good. I'll make a demo to gather additional feedback from everyone and mention that we can consider implementing the approach with two separate clients in the future.

Great stuff, thank you

nmezhenskyi:
@markylaing @tomponline Should I update LXD documentation regarding the use of client secrets as part of this PR or create a separate PR for it?

tomponline:
> @markylaing @tomponline Should I update LXD documentation regarding the use of client secrets as part of this PR or create a separate PR for it?

Yes please.we need an api extension too please

doc/api-extensions.md
```diff
@@ -2676,3 +2676,7 @@ Adds a `skip` mode to the restore request. This mode restores a cluster member's
 ## `disk_io_threads_virtiofsd`
 
 Adds the {config:option}`device-disk-device-conf:io.threads` option on `disk` devices which is used to control the `virtiofsd` thread pool size when sharing file systems into VMs. This can help improve I/O performance.
+
+## `oidc_client_secret`
+
+This adds support for the new configuration key {config:option}`server-oidc:oidc.client.secret`. This client secret can be used for OpenID Connect authorization code flow.
```
```suggestion
This adds support for the configuration key {config:option}`server-oidc:oidc.client.secret`. This client secret can be used for the OpenID Connect authorization code flow.
```

doc/authentication.md
```diff
@@ -136,6 +136,11 @@ The LXD client then retrieves and stores an access token, which it provides to L
 The identity provider might also provide a refresh token.
 In this case, the LXD client uses this refresh token to attempt to retrieve another access token when the current access token has expired.
 
+```{warning}
+The `oidc.client.secret` key should only be set if required by the Identity Provider. Setting this key will prevent CLI clients from authenticating.
```
minaelee:
```suggestion
Only set  `oidc.client.secret` if required by the Identity Provider. Once set, this key allows only the LXD UI client to authenticate, blocking CLI clients.
```

doc/authentication.md
```diff
@@ -136,6 +136,11 @@ The LXD client then retrieves and stores an access token, which it provides to L
...
+Only LXD UI clients will be able to authenticate if this key is set.
```
minaelee:
```suggestion
```

doc/server.md
```diff
@@ -54,6 +54,11 @@ The following server options configure external user authentication through {ref
     :end-before: <!-- config group server-oidc end -->
 ```
 
+```{important}
+Setting `oidc.client.secret` key will prevent CLI clients from authenticating, only LXD UI clients will be able to authenticate.
```
minaelee:
```suggestion
Setting `oidc.client.secret` prevents CLI clients from authenticating; only the LXD UI client will be able to authenticate.
```

doc/server.md
```diff
@@ -54,6 +54,11 @@ The following server options configure external user authentication through {ref
...
+We recommended setting this key only if required by the Identity Provider.
```
minaelee:
```suggestion
Set this key only if required by the Identity Provider.
```

minaelee (changes requested):

doc/api-extensions.md
markylaing: I think it's worth stating that the client secret will be used for the OIDC flow in the UI only. Perhaps:
```
If set, the client secret will be use for the OpenID Connect authorization code flow, which is used by LXD UI.
```

shared/version/api.go
```diff
@@ -451,6 +451,7 @@ var APIExtensions = []string{
 	"networks_all_projects",
 	"clustering_restore_skip_mode",
 	"disk_io_threads_virtiofsd",
+	"oidc_client_secret",
```
simondeziel, line 454: The commit adding this should be the first of the series to serve as a hint to Tom when going through backports for stable branches.

doc/api-extensions.md
```diff
@@ -2676,3 +2676,8 @@ Adds a `skip` mode to the restore request. This mode restores a cluster member's
 ## `disk_io_threads_virtiofsd`
 
 Adds the {config:option}`device-disk-device-conf:io.threads` option on `disk` devices which is used to control the `virtiofsd` thread pool size when sharing file systems into VMs. This can help improve I/O performance.
+
+## `oidc_client_secret`
+
+This adds support for the configuration key {config:option}`server-oidc:oidc.client.secret`. This client secret can be used for the OpenID Connect authorization code flow.
+Once set, this key allows only the LXD UI client to authenticate, blocking CLI clients.
```
nmezhenskyi: @markylaing @minaelee What do you think about this? I've added the line 2683 to state that the client secret will be used for the OIDC flow in the UI only.

shared/version/api.go
nmezhenskyi: Makes sense, done

doc/api-extensions.md
nmezhenskyi: I've added:
```
Once set, this key allows only the LXD UI client to authenticate, blocking CLI clients.
```

doc/api-extensions.md
markylaing: It's very tricky to word this correctly. Setting this value doesn't necessarily mean that authentication with the CLI won't work. That depends on the IdP. 

My recommendation would be:
```
This adds support for the {config:option}`server-oidc:oidc.client.secret` configuration key.
If set, the LXD server will use this value in the OpenID Connect (OIDC) authorization code flow, which is used by LXD UI.
This configuration value is not shared with clients of LXD (e.g. the LXD CLI).
```

minaelee:
```
This configuration value is not shared with other LXD clients (such as the LXD CLI).
```

doc/authentication.md
```diff
@@ -136,6 +136,10 @@ The LXD client then retrieves and stores an access token, which it provides to L
 The identity provider might also provide a refresh token.
 In this case, the LXD client uses this refresh token to attempt to retrieve another access token when the current access token has expired.
 
+```{warning}
+Only set `oidc.client.secret` if required by the Identity Provider. Once set, this key allows only the LXD UI client to authenticate, blocking CLI clients.
```
minaelee: Will need to also update the second line here, if it doesn't actually block.

doc/server.md
```diff
@@ -54,6 +54,11 @@ The following server options configure external user authentication through {ref
     :end-before: <!-- config group server-oidc end -->
 ```
 
+```{important}
+Setting `oidc.client.secret` prevents CLI clients from authenticating; only the LXD UI client will be able to authenticate.
```
minaelee: Again, will need to update this per Mark's comment that it doesn't necessarily prevent other clients from authenticating.

doc/api-extensions.md
nmezhenskyi: Makes sense, thank you both!

doc/authentication.md
nmezhenskyi: Updated to:
```
Only set `oidc.client.secret` if required by the Identity Provider. Once set, this key allows the LXD UI client to authenticate.
However, the secret is not shared with other LXD clients (such as the LXD CLI).
```

doc/server.md
nmezhenskyi: Updated to:
```
Setting `oidc.client.secret` may prevent CLI clients from authenticating depending on the Identity Provider policies.
Set this key only if required by the Identity Provider.
```

minaelee (approved):
LGTM docs-wise, thanks!

markylaing (approved):
LGTM thanks :rocket:

tomponline (approved):
Thanks @nmezhenskyi 

Are there any tests we can add for this? Perhaps to mini-oidc?

nmezhenskyi:
@tomponline 

> Are there any tests we can add for this? Perhaps to mini-oidc?

`mini-oidc` only tests the Device Authorization code flow and the client secret is currently not being used there.

We can add a UI e2e test for this. I've reached out to @edlerd regarding this.

To do that, we need to:
1. Update the workflows and repository secrets to also provide the client secret [here](https://github.com/canonical/lxd/blob/main/.github/workflows/tests.yml#L616-L629) and [here](https://github.com/canonical/lxd-ui/blob/main/.github/workflows/pr.yaml#L82-L98).
2. Use the new environment variable with the client secret in the [LXD UI setup test script](https://github.com/canonical/lxd-ui/blob/main/tests/scripts/setup_test).

tomponline:
> @tomponline
> 
> > Are there any tests we can add for this? Perhaps to mini-oidc?
> 
> `mini-oidc` only tests the Device Authorization code flow and the client secret is currently not being used there.
> 
> We can add a UI e2e test for this. I've reached out to @edlerd regarding this.
> 
> To do that, we need to:
> 
> 1. Update the workflows and repository secrets to also provide the client secret [here](https://github.com/canonical/lxd/blob/main/.github/workflows/tests.yml#L616-L629) and [here](https://github.com/canonical/lxd-ui/blob/main/.github/workflows/pr.yaml#L82-L98).
> 2. Use the new environment variable with the client secret in the [LXD UI setup test script](https://github.com/canonical/lxd-ui/blob/main/tests/scripts/setup_test).

Cool stuff, thanks, let me know when you need changes to the repo secrets.

PR #15707, opened 2025-06-02
lxd/cloudinit: Fix parsing of invalid cloud-init yaml

skozina:
This is a fix for the recent `lxd-ci` failures (or rather lxd crashes): https://github.com/canonical/lxd-ci/pull/488 .

Fixes regression in https://github.com/canonical/lxd/pull/15689

The problem is that this is not a valid yaml file: https://github.com/canonical/lxd-ci/blob/c0ba34b31fe3412208ae3f5eefb620fe866f20cf/tests/cloud-init#L66-L69
(or at least `yaml.Unmarshall()` call returns an error here: https://github.com/canonical/lxd/blob/825c4eb370a7dab14ed5fbab7980f64cb6cf4631/lxd/cloudinit/config.go#L150)

Previously lxd didn't crash, despite having `userCloudConfig == nil` here: https://github.com/canonical/lxd/blob/825c4eb370a7dab14ed5fbab7980f64cb6cf4631/lxd/cloudinit/config.go#L92
because golang allows calling methods on type `map[any]any` even if the map has value of `nil`, and because `yaml.Marshal(nil)` simply returns string `null` here: https://github.com/canonical/lxd/blob/825c4eb370a7dab14ed5fbab7980f64cb6cf4631/lxd/cloudinit/config.go#L209

The important part of this PR is the error checking on L92-99. The change of pointer type on methods is just a nit (which would lead to slightly nicer backtrace in case of crash).

### DIFF
--- a/lxd/cloudinit/config.go
+++ b/lxd/cloudinit/config.go
@@ -94,7 +94,7 @@ func GetEffectiveConfig(instanceConfig map[string]string, requestedKey string, i
 	// user-data's fields overwrite vendor-data's fields, so merging SSH keys can result in adding a "users" field
 	// that did not exist before, having the side effect of overwriting vendor-data's "users" field.
 	// So only merge into "user-data" when safe to do.
-	canMergeUserData := userCloudConfig.hasUsers() || !vendorCloudConfig.hasUsers()
+	canMergeUserData := (userErr == nil && userCloudConfig.hasUsers()) || (vendorErr == nil && !vendorCloudConfig.hasUsers())
 
 	// Merge additional SSH keys into parsed config.
 	// If merging is not possible return the raw value for the target key.
@@ -205,7 +205,7 @@ type cloudConfig struct {
 }
 
 // string marshals a cloud-config map into a YAML string.
-func (config cloudConfig) string() (string, error) {
+func (config *cloudConfig) string() (string, error) {
 	resultingConfigBytes, err := yaml.Marshal(config.keys)
 	if err != nil {
 		return "", err
@@ -218,7 +218,7 @@ func (config cloudConfig) string() (string, error) {
 }
 
 // string marshals a cloud-config map into a YAML string.
-func (config cloudConfig) hasUsers() bool {
+func (config *cloudConfig) hasUsers() bool {
 	value, ok := config.keys["users"]
 	return ok && value != ""
 }
@@ -226,7 +226,7 @@ func (config cloudConfig) hasUsers() bool {
 // mergeSSHKeyCloudConfig merges keys present in a map of userSSHKeys into a CloudConfig.
 // The provided map can be obtained by extracting user keys from an instance config with extractAdditionalSSHKeys.
 // This also returns the resulting YAML string after the merging is done.
-func (config cloudConfig) mergeSSHKeyCloudConfig(userKeys map[string]*userSSHKeys) (string, error) {
+func (config *cloudConfig) mergeSSHKeyCloudConfig(userKeys map[string]*userSSHKeys) (string, error) {
 	// If no keys are defined, return the original config passed in.
 	if len(userKeys) == 0 {
 		return config.string()


skozina:
@simondeziel while this PR would stop LXD from crashing (and would restore previous behavior), it is likely not sufficient, as we should understand why the yaml file in lxd-ci is not valid.
`yaml.Unmarshal()` on this yaml:
https://github.com/canonical/lxd-ci/blob/c0ba34b31fe3412208ae3f5eefb620fe866f20cf/tests/cloud-init#L66-L69
fails with:
`yaml: invalid map key: map[interface {}]interface {}{"ds.meta_data.user_user":interface {}(nil)}`

Any idea what "ds.meta_data.user_user" here means?

simondeziel:
> @simondeziel while this PR would stop LXD from crashing (and would restore previous behavior), it is likely not sufficient, as we should understand why the yaml file in lxd-ci is not valid. `yaml.Unmarshal()` on this yaml: https://github.com/canonical/lxd-ci/blob/c0ba34b31fe3412208ae3f5eefb620fe866f20cf/tests/cloud-init#L66-L69 fails with: `yaml: invalid map key: map[interface {}]interface {}{"ds.meta_data.user_user":interface {}(nil)}`

Feeding the jinja string to `yamllint` gives me this:

```
$ yamllint - << EOF
## template: jinja
#cloud-config
users:
  - name: {{ ds.meta_data.user_user }}
EOF
stdin
  2:2       warning  missing starting space in comment  (comments)
  4:13      error    too many spaces inside braces  (braces)
  4:36      error    too many spaces inside braces  (braces)
```

Removing the spaces inside the `{{}}` seem to appease it:

```
$ yamllint - << EOF
## template: jinja
#cloud-config
users:
  - name: {{ds.meta_data.user_user}}
EOF
stdin
  2:2       warning  missing starting space in comment  (comments)
```

I'm wondering if the `cloud-init.user-data` string should be treated as YAML before the jinja stuff has been processed?

> Any idea what "ds.meta_data.user_user" here means?

AFAIK, for `cloud-init`, DS stands for Data Source.

skozina:
> I'm wondering if the cloud-init.user-data string should be treated as YAML before the jinja stuff has been processed?

Yes, I am thinking we should disallow provisioning of SSH keys if Jinja template style of cloud-init config is provided. Just because currently we embed SSH keys in yaml, but Jinja templates are not yaml.

skozina:
Alternatively we'd need to get Jinja template processing in LXD, so that we could embed SSH keys the right way.

tomponline:
> > I'm wondering if the cloud-init.user-data string should be treated as YAML before the jinja stuff has been processed?
> 
> Yes, I am thinking we should disallow provisioning of SSH keys if Jinja template style of cloud-init config is provided. Just because currently we embed SSH keys in yaml, but Jinja templates are not yaml.

Yeah i agree with this. I was concerned originally when Pedro added this feature that we may face difficulty with the various formats cloud-init supports and we planned to only inject the ssh keys if the provided config was compatible with it being injected.

simondeziel:
> > I'm wondering if the cloud-init.user-data string should be treated as YAML before the jinja stuff has been processed?
> 
> Yes, I am thinking we should disallow provisioning of SSH keys if Jinja template style of cloud-init config is provided. Just because currently we embed SSH keys in yaml, but Jinja templates are not yaml.

Sounds fair to not touch the `cloud-init` user-data unless it's the simple/known variant starting with `#cloud-config`.

skozina:
@simondeziel I spoke with @tomponline , we agreed that we should try to parse whatever cloud-init config is provided, and if we manage to parse it as valid yaml, try to embed the SSH keys, if parsing or embedding fails, then return the original config file unmodified.
I've reviewed the code, I think it works that way, so I've updated the fix for rather minimal change so it doesn't crash any more, but keeps the previous behavior.
I'll also look at lxd-ci and see if we could introduce few more tests for cloud-config in there. Thanks!

tomponline:
@nmezhenskyi this might fix the local issue you're seeing with LXD panicking on:

```
--- FAIL: TestGetEffectiveConfig (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
    panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x6df23a]

goroutine 20 [running]:
testing.tRunner.func1.2({0x780e00, 0xaa3e10})
    /snap/go/10888/src/testing/testing.go:1734 +0x21c
testing.tRunner.func1()
    /snap/go/10888/src/testing/testing.go:1737 +0x35e
panic({0x780e00?, 0xaa3e10?})
    /snap/go/10888/src/runtime/panic.go:792 +0x132
github.com/canonical/lxd/lxd/cloudinit.GetEffectiveConfig(0xc000074de8, {0x0, 0x0}, {0x7dfa20, 0x8}, {0x7df4ce, 0x7})
    /home/nikita-mezhenskyi/dev/lxd/lxd/cloudinit/config.go:97 +0x79a
github.com/canonical/lxd/lxd/cloudinit.TestGetEffectiveConfig(0xc000102e00)
    /home/nikita-mezhenskyi/dev/lxd/lxd/cloudinit/config_test.go:16 +0x15c
testing.tRunner(0xc000102e00, 0x7fd880)
    /snap/go/10888/src/testing/testing.go:1792 +0xf4
created by testing.(*T).Run in goroutine 1
    /snap/go/10888/src/testing/testing.go:1851 +0x413
FAIL    github.com/canonical/lxd/lxd/cloudinit    0.007s
signal: segmentation fault (core dumped)
FAIL    github.com/canonical/lxd/lxd/cluster    0.174s
```

tomponline:
Interesting @skozina that `TestGetEffectiveConfig` fails for me too but is apparently not being run by the test suite.

tomponline:
@simondeziel i see that `make check-unit` is failing on main, but apparently the github workflows are not checking them?

```
make check-unit
=== RUN   TestGetEffectiveConfig
--- FAIL: TestGetEffectiveConfig (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x6df23a]

goroutine 7 [running]:
testing.tRunner.func1.2({0x780e00, 0xaa3e10})
	/snap/go/10888/src/testing/testing.go:1734 +0x21c
testing.tRunner.func1()
	/snap/go/10888/src/testing/testing.go:1737 +0x35e
panic({0x780e00?, 0xaa3e10?})
	/snap/go/10888/src/runtime/panic.go:792 +0x132
github.com/canonical/lxd/lxd/cloudinit.GetEffectiveConfig(0xc000071de8, {0x0, 0x0}, {0x7dfa20, 0x8}, {0x7df4ce, 0x7})
	/home/user/lxd/lxd/cloudinit/config.go:97 +0x79a
github.com/canonical/lxd/lxd/cloudinit.TestGetEffectiveConfig(0xc0000f4540)
	/home/user/lxd/lxd/cloudinit/config_test.go:16 +0x15c
testing.tRunner(0xc0000f4540, 0x7fd880)
	/snap/go/10888/src/testing/testing.go:1792 +0xf4
created by testing.(*T).Run in goroutine 1
	/snap/go/10888/src/testing/testing.go:1851 +0x413
FAIL	github.com/canonical/lxd/lxd/cloudinit	0.011s
FAIL
make: *** [Makefile:282: check-unit] Error 1

```

tomponline:
@simondeziel ah it looks like the unit tests are being skipped!

https://github.com/canonical/lxd/actions/runs/15419036694/job/43388882855?pr=15707

PR #15708, opened 2025-06-03
lxc: Prevent panic when overwriting the progress renderer

roosterfish:
The progress renderer's UpdateOp gets added as a handler to the operation. This allows the operation to update the progress on its own pace.

However when overwriting the initial progress renderer and assigning a new instance of it, this also replaces the underlying mutex which can cause a race if the operation is in the process of calling the orignial progress renderer's UpdateOp, obtains a lock of the mutex but the mutex is in the process of getting replaced because a new instance of the progress renderer is assigned.

This causes the following error https://github.com/canonical/lxd-ci/actions/runs/15387812020/job/43290215796?pr=485

In order to protect the initial mutex, assign the progress renderer for the actual export to a new variable.

### Reproducer

```go
func main() {
	progress := cmd.ProgressRenderer{
		Format: "Running operation: %s",
		Quiet:  false,
	}

	go func() {
		// Try to update the progress renderer in the background.
		// Under the hood the UpdateOp() calls Update() which locks/unlocks the progress renderer's mutex.
		for {
			progress.UpdateOp(api.Operation{
				Metadata: map[string]any{
					"test_progress": time.Now().String(),
				},
			})
		}
	}()

        progress.Done("")

	// Cause the next progress.UpdateOp to block after locking the progress renderer's mutex without unlocking:
	/*
		func (p *ProgressRenderer) Update(status string) {
			...
			// Acquire rendering lock
			p.lock.Lock()
			defer p.lock.Unlock()

			// Check if we're already done
			if p.done {
				if GlobalInt > 0 {
					<-GlobalCh
				}

				return
			}
			...
		}
	*/
	cmd.GlobalInt = 1
	time.Sleep(1 * time.Second)

	// Set a new progress renderer by overwriting the initial one.
	// This also replaces the progress renderer's mutex.
	progress = cmd.ProgressRenderer{
		Format: "Running new operation: %s",
		Quiet:  false,
	}

	// Release the blocking UpdateOp (Update) operation from the original progress renderer.
	// It will try to unlock the previously obtained lock which fails as the lock from the new progress renderer is already unlocked.
	close(cmd.GlobalCh)
	time.Sleep(1 * time.Second)
}
```

```
fatal error: sync: unlock of unlocked mutex

goroutine 7 [running]:
internal/sync.fatal({0x83f84c?, 0x4b?})
        /snap/go/10888/src/runtime/panic.go:1068 +0x18
internal/sync.(*Mutex).unlockSlow(0xc0000ec8ac, 0xffffffff)
        /snap/go/10888/src/internal/sync/mutex.go:204 +0x35
internal/sync.(*Mutex).Unlock(...)
        /snap/go/10888/src/internal/sync/mutex.go:198
sync.(*Mutex).Unlock(...)
        /snap/go/10888/src/sync/mutex.go:65
github.com/canonical/lxd/shared/cmd.(*ProgressRenderer).Update(0xc0000ec870, {0xc00046ca40, 0x37})
        /home/julian/dev/lxd/shared/cmd/progress.go:148 +0x38a
github.com/canonical/lxd/shared/cmd.(*ProgressRenderer).UpdateOp(0xc0000ec870, {{0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0, 0x0}, ...})
        /home/julian/dev/lxd/shared/cmd/progress.go:198 +0xf9
main.main.func1()
        /home/julian/dev/test/main.go:20 +0x70
created by main.main in goroutine 1
        /home/julian/dev/test/main.go:16 +0x8f

goroutine 1 [sleep]:
time.Sleep(0x3b9aca00)
        /snap/go/10888/src/runtime/time.go:338 +0x165
main.main()
        /home/julian/dev/test/main.go:60 +0x109
exit status 2
```

### DIFF
--- a/lxc/export.go
+++ b/lxc/export.go
@@ -163,22 +163,23 @@ func (c *cmdExport) run(cmd *cobra.Command, args []string) error {
 		}
 	}()
 
-	// Prepare the download request
-	progress = cli.ProgressRenderer{
+	// Prepare the download request.
+	// Assign the renderer to a new variable to not interfer with the old one.
+	exportProgress := cli.ProgressRenderer{
 		Format: i18n.G("Exporting the backup: %s"),
 		Quiet:  c.global.flagQuiet,
 	}
 
 	backupFileRequest := lxd.BackupFileRequest{
 		BackupFile:      io.WriteSeeker(target),
-		ProgressHandler: progress.UpdateProgress,
+		ProgressHandler: exportProgress.UpdateProgress,
 	}
 
 	// Export tarball
 	_, err = d.GetInstanceBackupFile(name, backupName, &backupFileRequest)
 	if err != nil {
 		_ = os.Remove(targetName)
-		progress.Done("")
+		exportProgress.Done("")
 		return fmt.Errorf("Fetch instance backup file: %w", err)
 	}
 
@@ -205,6 +206,6 @@ func (c *cmdExport) run(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("Failed to close export file: %w", err)
 	}
 
-	progress.Done(i18n.G("Backup exported successfully!"))
+	exportProgress.Done(i18n.G("Backup exported successfully!"))
 	return nil
 }
--- a/lxc/storage_volume.go
+++ b/lxc/storage_volume.go
@@ -2800,26 +2800,27 @@ func (c *cmdStorageVolumeExport) run(cmd *cobra.Command, args []string) error {
 
 	defer func() { _ = target.Close() }()
 
-	// Prepare the download request
-	progress = cli.ProgressRenderer{
+	// Prepare the download request.
+	// Assign the renderer to a new variable to not interfer with the old one.
+	exportProgress := cli.ProgressRenderer{
 		Format: i18n.G("Exporting the backup: %s"),
 		Quiet:  c.global.flagQuiet,
 	}
 
 	backupFileRequest := lxd.BackupFileRequest{
 		BackupFile:      io.WriteSeeker(target),
-		ProgressHandler: progress.UpdateProgress,
+		ProgressHandler: exportProgress.UpdateProgress,
 	}
 
 	// Export tarball
 	_, err = d.GetStoragePoolVolumeBackupFile(name, volName, backupName, &backupFileRequest)
 	if err != nil {
 		_ = os.Remove(targetName)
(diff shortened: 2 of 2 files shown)


roosterfish:
It looks the export of instances and custom volumes are the only two places where we reassign the `progress` variable a second time.

tomponline (commented):
Does this mean the old one is still running/updating when we move onto starting a new progress tracker?

roosterfish:
> Does this mean the old one is still running/updating when we move onto starting a new progress tracker?

The old `UpdateOp` handler is still added to the operation. In any case if the operation runs the handler, it checks if `done` and simply returns.
However if the mutex gets replaced (by assigning a new progress renderer) [whilst the old mutex is being locked](https://github.com/canonical/lxd/blob/main/shared/cmd/progress.go#L91-L97), then this causes the panic.

tomponline:
> > Does this mean the old one is still running/updating when we move onto starting a new progress tracker?
> 
> The old `UpdateOp` handler is still added to the operation. In any case if the operation runs the handler, it checks if `done` and simply returns. However if the mutex gets replaced (by assigning a new progress renderer) [whilst the old mutex is being locked](https://github.com/canonical/lxd/blob/main/shared/cmd/progress.go#L91-L97), then this causes the panic.

Why did the operation progress tracker need to be updated part way through the running operation?

Is this an indicator of a broader problem, or is it a common a pattern we should support safely.

I'm slightly concerned that the fix in this PR is more of a workaround and the same issue could occur/reoccur elsewhere easily.

roosterfish:
> I'm slightly concerned that the fix in this PR is more of a workaround and the same issue could occur/reoccur elsewhere easily.

Absolutely, this could happen again if the faulty pattern is reused in the future as the PR only ensures the original progress renderer's state isn't modified and instead another renderer is used for subsequent tasks.

> Is this an indicator of a broader problem, or is it a common a pattern we should support safely.

I think so yes. The operation allows adding handlers, but there is no option to remove a handler. And in this case we have two progress renderers which each add their `UpdateOp` as a handler to the operation. 
Usually you can freely call `UpdateOp` on a progress renderer which is already marked as "done" but this old renderer still tries to obtain its internal lock to check whether or not it is done. 
And this check is causing the panic as the lock might be replaced whilst checking.

tomponline:
I see so you're using a separate progress reporter for each API call.

PR #15718, opened 2025-06-04
test: Check that network info shows correct MTU for macvlan network

nmezhenskyi:
Follow-up to https://github.com/canonical/lxd/pull/15682.  

Adds tests to check that `lxc network info` shows correct MTU values for a macvlan network.

### DIFF
--- a/test/suites/container_devices_nic_macvlan.sh
+++ b/test/suites/container_devices_nic_macvlan.sh
@@ -110,6 +110,43 @@ test_container_devices_nic_macvlan() {
   echo "==> Check that lxc network info succeeds for macvlan network."
   lxc network info "${ctName}net"
 
+  echo "==> Check that macvlan network info shows parent's MTU by default."
+  parentMTU=$(</sys/class/net/"${ctName}"/mtu)
+  if ! lxc network info "${ctName}net" | grep -xF "MTU: ${parentMTU}" ; then
+    echo "default mtu not inherited from parent"
+    false
+  fi
+
+  echo "==> Set a valid MTU config value for macvlan network."
+  customMTU=1492
+  lxc network set "${ctName}net" mtu="${customMTU}"
+
+  echo "==> Check that network info shows correct MTU value for macvlan network."
+  if ! lxc network info "${ctName}net" | grep -xF "MTU: ${customMTU}" ; then
+    echo "config mtu ignored"
+    false
+  fi
+
+  echo "==> Check that setting MTU config value out of range (1280-16384) is not allowed."
+  ! lxc network set "${ctName}net" mtu=0 || false
+  ! lxc network set "${ctName}net" mtu=1000 || false
+  ! lxc network set "${ctName}net" mtu=1279 || false
+  ! lxc network set "${ctName}net" mtu=16385 || false
+  ! lxc network set "${ctName}net" mtu=50000 || false
+
+  echo "==> Check that setting MTU config value to the boundaries of range (1280-16384) is allowed."
+  lxc network set "${ctName}net" mtu=1280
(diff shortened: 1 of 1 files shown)


test/suites/container_devices_nic_macvlan.sh
```diff
@@ -110,6 +110,37 @@ test_container_devices_nic_macvlan() {
   echo "==> Check that lxc network info succeeds for macvlan network."
   lxc network info "${ctName}net"
 
+  echo "==> Check that macvlan network info shows parent's MTU by default."
+  parentMTU=$(</sys/class/net/"${ctName}"/mtu)
```
simondeziel, line 114: Saving a `fork` by not calling `cat`, I like ;)

test/suites/container_devices_nic_macvlan.sh
```diff
@@ -110,6 +110,37 @@ test_container_devices_nic_macvlan() {
...
+  if ! lxc network info "${ctName}net" | grep -F "${parentMTU}" ; then
```
simondeziel: Since the `ctName` includes `$$` for the current PID, there is a small risk of matching in the `used_by` section. Using `grep -xF "MTU: ${parentMTU}"` here would avoid this (albeit) tiny risk.

test/suites/container_devices_nic_macvlan.sh
```diff
@@ -110,6 +110,37 @@ test_container_devices_nic_macvlan() {
...
+
+  echo "==> Check that network info shows correct MTU value for macvlan network."
+  if ! lxc network info "${ctName}net" | grep -F "${customMTU}" ; then
+    echo "config mtu ignored"
+    false
+  fi
+
+  echo "==> Check that setting MTU config value out of range (1280-16384) is not allowed."
+  ! lxc network set "${ctName}net" mtu=0 || false
+  ! lxc network set "${ctName}net" mtu=1000 || false
```
simondeziel, line 132: Since there is a known good range, testing at boundaries may be useful (`1279`, `1280`, `16384` and `16385`)

test/suites/container_devices_nic_macvlan.sh
```diff
@@ -110,6 +110,37 @@ test_container_devices_nic_macvlan() {
...
+  customMTU=1492
+  lxc network set "${ctName}net" mtu="${customMTU}"
+
+  echo "==> Check that network info shows correct MTU value for macvlan network."
+  if ! lxc network info "${ctName}net" | grep -F "${customMTU}" ; then
+    echo "config mtu ignored"
+    false
+  fi
+
+  echo "==> Check that setting MTU config value out of range (1280-16384) is not allowed."
```
simondeziel, line 130: The 1280 is likely for IPv6 but I'm curious where that upper bound comes from? Is that a macvlan thing?

nmezhenskyi: Hm, I believe some vendors use this value as jumbo frame size.

https://github.com/canonical/lxd/blob/b3c7efadc73cbbbb08482f6eb72ac36ccc6862bd/shared/validate/validate.go#L460C1-L475C2

I could find IBM mentioning this in their [docs](https://www.ibm.com/docs/en/aix/7.2.0?topic=tuning-tcp-streaming-workload):

> For example, use the ifconfig at0 mtu 16384 command to set the ATM MTU size to 16 KB. This causes TCP to compute a smaller MSS value. With a 16 KB MTU size, TCP can send four packets for a 64 KB window size.

simondeziel: Thanks for researching this!

tomponline (approved):
LGTM!

PR #15719, opened 2025-06-04
Misc tweaks

simondeziel:


lxd/storage/utils.go
```diff
@@ -972,7 +972,7 @@ func volumeIsUsedByDevice(vol api.StorageVolume, instanceType instancetype.Type,
 // VolumeUsedByProfileDevices finds profiles using a volume and passes them to profileFunc for evaluation.
 // The profileFunc is provided with a profile config, project config and a list of device names that are using
 // the volume.
-func VolumeUsedByProfileDevices(s *state.State, poolName string, projectName string, vol *api.StorageVolume, profileFunc func(profileID int64, profile api.Profile, project api.Project, usedByDevices []string) error) error {
+func VolumeUsedByProfileDevices(s *state.State, poolName string, projectName string, vol *api.StorageVolume, profileFunc func(profile api.Profile, project api.Project, usedByDevices []string) error) error {
```
simondeziel: I think this `usedByDevices` variable is shadowed by https://github.com/canonical/lxd/pull/15719/files#diff-33511e3a6e5abae8b54480aa23c933de0ec50584842b582286029955063c389dL1051

simondeziel: @kadinsayani helped me understand that there is no shadowing going on, I got the scoping wrong ;)

lxd/instance_post.go
```diff
@@ -896,7 +896,7 @@ func instancePostClusteringMigrate(s *state.State, r *http.Request, srcPool stor
 
 // instancePostClusteringMigrateWithRemoteStorage handles moving a remote shared storage instance from a source member that is offline.
 // This function must be run on the target cluster member to move the instance to.
-func instancePostClusteringMigrateWithRemoteStorage(s *state.State, r *http.Request, srcPool storagePools.Pool, srcInst instance.Instance, newInstName string, newMember db.NodeInfo, stateful bool) (func(op *operations.Operation) error, error) {
+func instancePostClusteringMigrateWithRemoteStorage(s *state.State, srcPool storagePools.Pool, srcInst instance.Instance, newInstName string, newMember db.NodeInfo) (func(op *operations.Operation) error, error) {
```
kadinsayani, line 899: Can the commit message or description be expanded to indicate that the `stateful` parameter is also being dropped here?

lxd/profiles_utils.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/canonical/lxd/shared/api"
 )
 
-func doProfileUpdate(s *state.State, p api.Project, profileName string, id int64, profile *api.Profile, req api.ProfilePut) error {
+func doProfileUpdate(s *state.State, p api.Project, profileName string, profile *api.Profile, req api.ProfilePut) error {
```
kadinsayani: I think we can use `id` in the below database transactions:

https://github.com/canonical/lxd/blob/74f662df66b345a3d2026df2d8b087436c1c13e4/lxd/profiles_utils.go#L104C3-L117C4

kadinsayani: It might be worth checking if `id` is already available at the call sites. We should minimize calls to `cluster.GetProfileID()`.

simondeziel: https://github.com/canonical/lxd/blob/74f662df66b345a3d2026df2d8b087436c1c13e4/lxd/profiles_utils.go#L95 calls `cluster.UpdateProfile()` which immediately get the profile ID: https://github.com/canonical/lxd/blob/main/lxd/db/cluster/profiles.mapper.go#L479 ... only to get it again later. This being in a `.mapper.go` makes it a bit more annoying to fix.

I think I'll back down on those changes as this sounds like it require more work than I'm willing to do ATM.

simondeziel: @kadinsayani I think you've found a good thing to look at more closely but I'm just saying that I don't have the cycles to do that ATM ;)

lxd/instance_post.go
simondeziel: Thanks for catching this!

kadinsayani (approved):
Thanks!

PR #15720, opened 2025-06-04
github: add golangci-lint action

simondeziel:


test/lint/golangci.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 set -eu
 
+# Skip if already ran via golangci/golangci-lint-action
+if [ -n "${GITHUB_ACTIONS:-}" ]; then
+    echo "GitHub Action runner detected, skipping golangci-lint script (already done)"
+    exit 0
+fi
+
```
tomponline, line 9: If we arent using this on github any more we should remove the github references below.

simondeziel: I've put the PR back to draft as I don't like it to operate only on a patch between main's HEAD and the PR. I'll try to bring it to parity with what we have/had in the script where `whole-files` get scooped in as well.

The long term goal would be to just run it on everything but it's still a long road till we get there :/

simondeziel: I'm happy with what it looks like now. It runs:

```
/home/runner/golangci-lint-2.1.6-linux-amd64/golangci-lint run  --timeout=5m --new --new-from-rev=74f662df66b345a3d2026df2d8b087436c1c13e4 --whole-files
```

Where 74f662df66b345a3d2026df2d8b087436c1c13e4 is the current main's HEAD which is the exact same commit the old script used to test with as can be seen in https://github.com/canonical/lxd/actions/runs/15452542118/job/43497869117#step:22:41

.github/workflows/tests.yml
```diff
@@ -234,6 +234,13 @@ jobs:
             exit 1
           fi
 
+      # XXX: `make static-analysis` also run golangci-lint but this one provides
+      #      useful feedback in the PR through github-code-scanning bot
+      - name: Run golangci-lint
+        uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 # v8
+        with:
+          args: --timeout=5m --new --new-from-rev=${{ github.event.pull_request.base.sha }} --whole-files
```
tomponline, line 243: Does this work with "push" action for when its merged? 
I remember it took a long time to figure out something that worked for both.

tomponline:
Please rebase

.github/workflows/tests.yml
markylaing: The script uses `GITHUB_BEFORE` (`github.event.before`) if `GITHUB_BASE_REF` (`github.event.pull_request.base_sha`) is unset. I haven't done the archaeology to find out why.

simondeziel: Thanks for pointing that out, I've also cleaned that bit now.

I decided to align with how shellcheck differential runs were only done during PR tests.

tomponline (approved):
Thanks!

PR #15722, opened 2025-06-04
Run shellcheck only once

simondeziel:


tomponline (approved):

tomponline:
@simondeziel how are we going to track the lists of which shell scripts are tested remain in sync?

simondeziel:
> @simondeziel how are we going to track the lists of which shell scripts are tested remain in sync?

Is your concern if we add another directory with scripts (like `test/foo/`)? If so, that shouldn't be different than before as we'd need to remember to update the list wherever it would be (`Makefile` or `test/lint/shellcheck.sh`).

Also, the GH action one would flag it as we don't give it any list, it has the smart to figure what to look for in the diff so it would remind us eventually if we forget as it would fail CI while a local `make static-analysis` would be passing.

Or maybe I completely missed your point.

tomponline:
> Is your concern if we add another directory with scripts (like `test/foo/`)? If so, that shouldn't be different than before as we'd need to remember to update the list wherever it would be (`Makefile` or `test/lint/shellcheck.sh`).

My concern is having to update it in two places, so if the github one doesnt need a list thats good.

PR #15724, opened 2025-06-05
Add `lxd-benchmark` tests

simondeziel:


simondeziel:
@tomponline the new tests ran as can be seen here: https://github.com/canonical/lxd/actions/runs/15457406077?pr=15724#summary-43512346053

In terms of runtime cost:

* 45 seconds for `btrfs`
* ~1 minute for `dir`
* ~1m15s for `zfs`
* ~2 minutes for `lvm`
* ~8 minutes for `ceph`

That's a steep price to pay IMHO to get semi-decent comparison of storage backend perfs. I say semi-decent due to GHA performance over time being all over the place.

simondeziel:
> That's a steep price to pay IMHO to get semi-decent comparison of storage backend perfs. I say semi-decent due to GHA performance over time being all over the place.

Those 8 extra minutes are really not welcome, by me at least as I have a personal goal of halving the runtime of the `ceph` `standalone` tests.

Should we disable them except the `basic` variant when on `ceph`? I'm open to suggestions.

tomponline:
Hi @simondeziel what is this PR trying to achieve, is there any additional context?

Historically the lxd-benchmark tests ran daily rather on each PR?

Would this make more sense?

simondeziel:
> Hi @simondeziel what is this PR trying to achieve, is there any additional context?

The main goal was to get code coverage data from that bit of code. I was hoping for something that had a cheap runtime cost so didn't give much thought to a daily run.

> Historically the lxd-benchmark tests ran daily rather on each PR?
> 
> Would this make more sense?

Right. I think we should keep at least the `lxd_benchmark_basic` one in the regular/per-PR tuns just so that we get early feedback if this ever gets touched. This is rather quick, only ~21s with `ceph`.

As for the serial and parallel variant they don't give reliable numbers due to running in GHA runners but they give an overview of the relative cost of creating instances en masse for the different storage drivers. Since this provide some value, I introduced a `LXD_EXPENSIVE` toggle that will enable those scripts during scheduled or manually triggered runs.

Let me know what you think, please.

tomponline:
> Right. I think we should keep at least the `lxd_benchmark_basic` one in the regular/per-PR tuns just so that we get early feedback if this ever gets touched. This is rather quick, only ~21s with `ceph`.

Lets just do this for now and then revisit the benchmarking as a separate item later.

simondeziel:
> > Right. I think we should keep at least the `lxd_benchmark_basic` one in the regular/per-PR tuns just so that we get early feedback if this ever gets touched. This is rather quick, only ~21s with `ceph`.
> 
> Lets just do this for now and then revisit the benchmarking as a separate item later.

So I let go of the `LXD_EXPENSIVE` flag? and drop the serial and parallel tests?

tomponline:
> So I let go of the `LXD_EXPENSIVE` flag? and drop the serial and parallel tests?

Yeah, I see these as not belonging to the main LXD test suite.

simondeziel:
> Comments suppressed due to low confidence (2)
> 
> **test/suites/lxd-benchmark.sh:3**
> 
>     * [nitpick] Consider adding an assertion after init to verify that exactly 5 containers were created (for example by using `lxc list` and counting entries), ensuring the init behavior is validated.
> 
> 
> ```
>   lxd-benchmark init --count 5 testimage
> ```

That is actually a good idea. I've added some checks which leverage the filtering capabilities of `lxd list`.

> **test/main.sh:413**
> 
>     * Indentation of this `run_test` call should align with the surrounding entries (e.g., use the same number of spaces as the lines above and below).
> 
> 
> ```
>     run_test test_lxd_benchmark_basic "lxd-benchmark basic init/start/stop/delete"
> ```

I am not sure of what you mean here. The `run_test` part is well aligned and we don't care about aligning the other parts.

simondeziel:
The `standalone`+`zfs` issue is an unrelated regression.

tomponline (approved):

simondeziel:
This made the TiCS code coverage score of the `lxd-benchmark` go from 0 to 78.55%.

This moved the overall code coverage score from 72.60% to 72.82%... at least it registered :)

PR #15727, opened 2025-06-05
devLXD: Response format based on vsock usage 

MusicDin:
This PR changes the devLXD response functions in a way that it determines the response format based on the vsock usage. If devLXD is being interacted over vsock (from LXD agent), the responses are returned as `api.Response`. Otherwise, the response are formatted as `api.DevLXDResponse`. This information is stored in the request context, and is retrieved from it during the response rendering.

Additionally, the LXD agent now uses the devLXD client to interact with the devLXD. This means that the devLXD client is adjusted to expect `api.Response` when used from LXD agent.

### DIFF
--- a/lxd-agent/devlxd.go
+++ b/lxd-agent/devlxd.go
@@ -301,37 +301,19 @@ var devLXDUbuntuProEndpoint = devLXDAPIEndpoint{
 }
 
 func devLXDUbuntuProGetHandler(d *Daemon, r *http.Request) *devLXDResponse {
-	// Get a http.Client.
-	client, err := getClient(d.serverCID, int(d.serverPort), d.serverCertificate)
+	client, err := getDevLXDVsockClient(d)
 	if err != nil {
-		return smartResponse(fmt.Errorf("Failed connecting to LXD over vsock: %w", err))
+		return smartResponse(fmt.Errorf("Failed connecting to devLXD over vsock: %w", err))
 	}
 
-	// Remove the request URI, this cannot be set on requests.
-	r.RequestURI = ""
+	defer client.Disconnect()
 
-	// Set up the request URL with the correct host.
-	r.URL = &api.NewURL().Scheme("https").Host("custom.socket").Path(version.APIVersion, "ubuntu-pro").URL
-
-	// Proxy the request.
-	resp, err := client.Do(r)
-	if err != nil {
-		return errorResponse(http.StatusInternalServerError, err.Error())
-	}
-
-	var apiResponse api.Response
-	err = json.NewDecoder(resp.Body).Decode(&apiResponse)
+	settings, err := client.GetUbuntuPro()
 	if err != nil {
 		return smartResponse(err)
 	}
 
-	var settingsResponse api.UbuntuProSettings
-	err = json.Unmarshal(apiResponse.Metadata, &settingsResponse)
-	if err != nil {
-		return errorResponse(http.StatusInternalServerError, fmt.Sprintf("Invalid Ubuntu Token settings response received from host: %v", err))
-	}
-
-	return okResponse(settingsResponse, "json")
+	return okResponse(settings, "json")
 }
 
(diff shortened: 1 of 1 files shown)


lxd/response/response.go
```diff
@@ -15,6 +15,7 @@ import (
 
 	"github.com/canonical/lxd/client"
 	"github.com/canonical/lxd/lxd/metrics"
+	"github.com/canonical/lxd/lxd/request"
```
MusicDin, line 18: I would like to draw some attention here.
Since the information whether the vsock is being used is required during response rendering, the `response` package is now depending on `request` package (to reference a context key - `request.DevLXDOverVsock`).

If this is not desired, we can use *plain string* to reference the key. Or maybe there is even better approach that I am not seeing.

markylaing: I'm not certain about the side effects of this but one option would be to define the context key elsewhere. The context key should be of a particular type and not just a plain string (you'll get a linter complaint otherwise, I think this is required for key uniqueness or something)

tomponline:
Please rebase

markylaing:
Initial review - all looking very neat and tidy! Especially in the lxd-agent :rocket:

lxd/response/response.go
MusicDin: Cool, so if there are any undesired side effects then we should move the key in some package that can be referenced from `lxd` and `lxd/response` package. We can use `lxd/response` package itself as this would remove the need for a new dependency and the key is mainly used to determine the response format.

markylaing:
```go

	trnsprt, ok := r.http.Transport.(*http.Transport)
	if !ok {
		panic("error")
	}
	
	if trnsprt.TLSClientConfig != nil {
		// When communicating with devLXD over vsock, the response is
		// expected to be in devLXD format.
		resp, etag, err := lxdParseResponse(resp)
		if err != nil {
			return nil, "", err
		}

		// Wrap into devLXD response.
		return &api.DevLXDResponse{
			Content:    resp.Metadata,
			StatusCode: resp.StatusCode,
		}, etag, nil
	}
```

Suggested in 1:1 using the above instead of having an `isVSock` field :)

MusicDin:
Just for context, @markylaing's suggestion is a way to get rid of setting/hardcoding `isDevLXDOverVsock=true` when `ConnectDevLXDHTTP` is called.

client/connection.go
```diff
@@ -362,6 +362,52 @@ func ConnectDevLXDWithContext(ctx context.Context, socketPath string, args *Conn
 	}, nil
 }
 
+// ConnectDevLXDHTTP lets you connect to devLXD over a VM socket.
+func ConnectDevLXDHTTP(args *ConnectionArgs, client *http.Client) (DevLXDServer, error) {
```
tomponline: do we need this function - why wouldnt we provide a context?

MusicDin: For consistency, but it is not being used. Will remove

lxd/response/response.go
tomponline: This might be an issue for MicroCluster, but I think @roosterfish is in the process of untangling that stuff anyway.

client/connection.go
```diff
@@ -362,6 +362,46 @@ func ConnectDevLXDWithContext(ctx context.Context, socketPath string, args *Conn
 	}, nil
 }
 
+// ConnectDevLXDHTTPWithContext lets you connect to devLXD over a VM socket.
```
tomponline, line 365: VM socket? You mean Instance socket - is this for container or VMs?

tomponline (approved):
LGTM thanks!

client/connection.go
MusicDin: For containers we have just `ConnectDevLXD[WithContext]()`, which connects to the provided unix path (e.g. `/dev/lxd/sock`).

This is for VMs, and it was meant to be `... over a vsock`.

tomponline: Oh I see the important difference was "HTTP".

PR #15728, opened 2025-06-05
Add tests for cloud-init

skozina:


test/suites/cloud-init.sh
```diff
@@ -27,4 +27,79 @@ test_cloud_init() {
   [ -n "${ID6}" ] && [ "${ID6}" != "${ID5}" ]
 
   lxc delete -f c1 c2
+
+  ensure_import_testimage
```
simondeziel: That's already been done at the beginning of the function.

test/suites/cloud-init.sh
```diff
@@ -27,4 +27,79 @@ test_cloud_init() {
...
+  lxd sql global "UPDATE images SET cached=0 WHERE fingerprint=\"${fingerprint}\""
+
+  (
+    cd devlxd-client || return
+    # Use -buildvcs=false here to prevent git complaining about untrusted directory when tests are run as root.
+    CGO_ENABLED=0 go build -tags netgo -v -buildvcs=false ./...
+  )
+
+  lxc launch testimage devlxd
+  lxc file push --mode 0755 "devlxd-client/devlxd-client" devlxd/bin/
```
simondeziel: Please add `--quiet` here.

test/suites/cloud-init.sh
```diff
@@ -27,4 +27,79 @@ test_cloud_init() {
...
+
+  # With SSH keys configured, unknown cloud-init formats should be passed to the instance unmodified
+  lxc config set devlxd cloud-init.user-data="invalid-yaml"
+  lxc config set devlxd cloud-init.ssh-keys.mykey="root:gh:user1"
+  [ "$(lxc exec devlxd -- devlxd-client cloud-init user-data)" = "invalid-yaml" ]
+  lxc config unset devlxd cloud-init.ssh-keys.mykey
+  lxc config unset devlxd cloud-init.user-data
+
+  # Configured ssh-keys in user-data should not replace vendor-data with users provided
+  CLOUD_INIT="$(cat <<EOF
```
simondeziel: `cat` is unneeded here, we can have `""` spanning multiple lines.

test/suites/cloud-init.sh
```diff
@@ -27,4 +27,79 @@ test_cloud_init() {
...
+  [ "$(lxc exec devlxd -- devlxd-client cloud-init user-data)" = "not found" ]
+  lxc config unset devlxd cloud-init.ssh-keys.mykey
+  lxc config unset devlxd cloud-init.vendor-data
+
+  # Valid format should work with ssh keys embedded
+  CLOUD_INIT="$(cat << EOF
+## template: jinja
+#cloud-config
+runcmd:
+  - echo {{v1.local_hostname}} > /var/tmp/runcmd_output
```
simondeziel: Is `v1` OK here or does that need to match the actual instance name?

test/suites/cloud-init.sh
```diff
@@ -27,4 +27,79 @@ test_cloud_init() {
...
+  ssh-import-id:
+  - gh:user1 #lxd:cloud-init.ssh-keys
+  ssh_import_id:
+  - gh:user1 #lxd:cloud-init.ssh-keys
+EOF
+  )"
+  lxc exec devlxd -- devlxd-client cloud-init user-data
+  [ "$(lxc exec devlxd -- devlxd-client cloud-init user-data)" = "${EXPECTED_CLOUD_INIT}" ]
+
+  lxc stop devlxd
```
simondeziel: I'd skip the `lxc stop` here as it's quicker to `lxc delete -f` it... we tend to not be nice with test instances ;)

test/suites/cloud-init.sh
```diff
@@ -27,4 +27,79 @@ test_cloud_init() {
...
+  - echo {{merged_system_cfg._doc}} >> /var/tmp/runcmd_output
+EOF
+  )"
+  lxc config set devlxd cloud-init.user-data="${CLOUD_INIT}"
+  lxc config set devlxd cloud-init.ssh-keys.mykey="root:gh:user1"
+  EXPECTED_CLOUD_INIT="$(cat << EOF
+## template: jinja
+#cloud-config
+runcmd:
+- echo {{v1.local_hostname}} > /var/tmp/runcmd_output
```
simondeziel: Same deal with `v1`.

test/suites/cloud-init.sh
```diff
@@ -27,4 +27,68 @@ test_cloud_init() {
   [ -n "${ID6}" ] && [ "${ID6}" != "${ID5}" ]
 
   lxc delete -f c1 c2
+
+  fingerprint="$(lxc image info testimage | awk '/^Fingerprint:/ {print $2}')"
```
simondeziel:
```suggestion
  fingerprint="$(lxc image list -c F -f csv testimage)"
```

test/suites/cloud-init.sh
skozina: Gee, how could have I missed that. Thanks, removed.

test/suites/cloud-init.sh
skozina: Added, thanks.

test/suites/cloud-init.sh
```diff
@@ -27,4 +27,68 @@ test_cloud_init() {
...
+  (
+    cd devlxd-client || return
+    # Use -buildvcs=false here to prevent git complaining about untrusted directory when tests are run as root.
+    CGO_ENABLED=0 go build -tags netgo -v -buildvcs=false ./...
+  )
+
+  lxc launch testimage devlxd
+  lxc file push --quiet --mode 0755 "devlxd-client/devlxd-client" devlxd/bin/
+
+  # Unknown cloud-init formats should be passed to the instance unmodified
```
simondeziel: Please turn those comments into `echo`'s.

test/suites/cloud-init.sh
skozina: Right, removed cat from all the use. Thanks!

test/suites/cloud-init.sh
skozina: It's probably fine for this use, since we don't actually care if cloud-config will work. We only test that what we set is configured.
I've fixed this so the use case feels more real, thanks for spotting it!

test/suites/cloud-init.sh
skozina: Removed, thanks!

test/suites/cloud-init.sh
```diff
@@ -27,4 +27,68 @@ test_cloud_init() {
...
+  CLOUD_INIT="#cloud-config
+users:
+  - name: root
+    ssh-import-id: gh:user2"
+
+  lxc config set devlxd cloud-init.vendor-data="${CLOUD_INIT}"
+  lxc config set devlxd cloud-init.ssh-keys.mykey="root:gh:user1"
+  lxc exec devlxd -- devlxd-client cloud-init vendor-data | grep "gh:user1"
+  lxc exec devlxd -- devlxd-client cloud-init vendor-data | grep "gh:user2"
+  [ "$(lxc exec devlxd -- devlxd-client cloud-init user-data)" = "not found" ]
```
simondeziel, line 63: Should we expect a non-0 return from `devlxd-client` or is that client just very simplistic/only for testing purposes?

simondeziel (commented):
LGTM overall, sorry for the piece-meal reviews.

test/suites/cloud-init.sh
skozina: Hm... I actually don't know. If there's no data, devlxd-client will still succeed (as LXD will return `not found` string).
devlxd-client will return non-0 if there's an error return by the http library used for the request, but I'm not sure if that can realistically happen with a unix socket connection.
I think that checking only for the returned strings, rather than the return value, is sufficient.

test/suites/cloud-init.sh
skozina: Thanks, done.

test/suites/cloud-init.sh
skozina: Thanks, done.

test/suites/cloud-init.sh
simondeziel: Sounds good to me, thanks

test/suites/cloud-init.sh
```diff
@@ -27,4 +27,68 @@ test_cloud_init() {
...
+runcmd:
+- echo {{devlxd.local_hostname}} > /var/tmp/runcmd_output
+- echo {{merged_system_cfg._doc}} >> /var/tmp/runcmd_output
+users:
+- name: root
+  ssh-import-id:
+  - gh:user1 #lxd:cloud-init.ssh-keys
+  ssh_import_id:
+  - gh:user1 #lxd:cloud-init.ssh-keys"
+  lxc exec devlxd -- devlxd-client cloud-init user-data
```
simondeziel: This call is extraneous as the next one will show up in the `set -e` output as can be seen here: https://github.com/canonical/lxd/actions/runs/15471326177/job/43556869815?pr=15728#step:15:46979

test/suites/cloud-init.sh
```diff
@@ -27,4 +27,68 @@ test_cloud_init() {
...
+  echo "Check that configured ssh-keys in user-data do not replace vendor-data with users provided"
+  CLOUD_INIT="#cloud-config
+users:
+  - name: root
+    ssh-import-id: gh:user2"
+
+  lxc config set devlxd cloud-init.vendor-data="${CLOUD_INIT}"
+  lxc config set devlxd cloud-init.ssh-keys.mykey="root:gh:user1"
+  lxc exec devlxd -- devlxd-client cloud-init vendor-data | grep "gh:user1"
+  lxc exec devlxd -- devlxd-client cloud-init vendor-data | grep "gh:user2"
```
simondeziel:
```suggestion
  VENDOR_DATA="$(lxc exec devlxd -- devlxd-client cloud-init vendor-data)"
  grep -F "gh:user1" <<< "${VENDOR_DATA}"
  grep -F "gh:user2" <<< "${VENDOR_DATA}"
```

test/suites/cloud-init.sh
skozina: Thanks for spotting it, removed.

test/suites/cloud-init.sh
skozina: Good point, I've changed this as you suggest. Thanks!

test/suites/cloud-init.sh
```diff
@@ -27,4 +27,68 @@ test_cloud_init() {
   [ -n "${ID6}" ] && [ "${ID6}" != "${ID5}" ]
 
   lxc delete -f c1 c2
+
+  fingerprint="$(lxc image list -c F -f csv testimage)"
+
+  # Ensure testimage is not set as cached.
+  lxd sql global "UPDATE images SET cached=0 WHERE fingerprint=\"${fingerprint}\""
```
tomponline: @skozina what is this part for, could we have a comment explaining why?

tomponline (dismissed):
LGTM, but I'd like @MusicDin to give it a brief once over too.

client/devlxd_cloudinit.go
```diff
@@ -0,0 +1,35 @@
+package lxd
+
+import (
+	"net/http"
+)
+
+// GetCloudInitUserData returns cloud-init user data.
+func (r *ProtocolDevLXD) GetCloudInitUserData() (string, error) {
+	data, _, err := r.query(http.MethodGet, "/config/cloud-init.user-data", nil, "")
```
MusicDin: I like the util functions for cloud init.

Although, the calls could be simplified in each function:
```
return r.GetConfigByKey("cloud-init.user-data")

return r.GetConfigByKey("cloud-init.vendor-data")

return r.GetConfigByKey("cloud-init.network-config")
```

MusicDin:
May I ask who is the target audience of cloud-init util functions in devLXD client? 

IMO, if this is mainly meant to be used externally, I'm seeing benefit in providing shortcuts for parsing cloud init configuration. If it is only for internal usage, I would be tempted to avoid introducing new functions in the devLXD client and just use `client.GetConfigByKey("cloud-init.<key>")` in devlxd test binary.

tomponline:
> May I ask who is the target audience of cloud-init util functions in devLXD client?
> 
> IMO, if this is mainly meant to be used externally, I'm seeing benefit in providing shortcuts for parsing cloud init configuration. If it is only for internal usage, I would be tempted to avoid introducing new functions in the devLXD client and just use `client.GetConfigByKey("cloud-init.<key>")` in devlxd test binary.

ah yeah good call @MusicDin i concur.

test/suites/cloud-init.sh
skozina: Sorry, remnant of my testing where I took some code initially from test/suites/devlxd.sh, where it was used for testing of private images sharing via devlxd. Removed now. Thanks for catching this!

skozina:
> May I ask who is the target audience of cloud-init util functions in devLXD client?

I think the target audience is internal testing only. Our exported interface is the config setting on the host, what's visible in the guest (which is different from the host setting as the SSH keys can be embedded in that) is internal implementation to LXD and should not confuse people.

> IMO, if this is mainly meant to be used externally, I'm seeing benefit in providing shortcuts for parsing cloud init configuration. If it is only for internal usage, I would be tempted to avoid introducing new functions in the devLXD client and just use `client.GetConfigByKey("cloud-init.<key>")` in devlxd test binary.

Good point. I've removed all changes to the client library, and instead we're calling now the `client.GetConfigByKey()` functions directly from devlxd-client. Thanks for the tip!

tomponline (approved):
LGTM thanks!

test/devlxd-client/devlxd-client.go
```diff
@@ -112,6 +112,31 @@ func run(args []string) error {
...
+		var err error
+		switch args[2] {
+		case "user-data":
+			config, err = client.GetConfigByKey("cloud-init.user-data")
+		case "vendor-data":
+			config, err = client.GetConfigByKey("cloud-init.vendor-data")
+		case "network-config":
+			config, err = client.GetConfigByKey("cloud-init.network-config")
+		default:
+			return fmt.Errorf("Usage: %s cloud-init <user-data|vendor-data|network-config>", args[0])
```
simondeziel, line 131: It's a tiny thing but the exact error is returned if the number of args != 3.

simondeziel (approved):

MusicDin (approved):
Thanks! :rocket:

test/devlxd-client/devlxd-client.go
skozina: Ah, I didn't even think about it as an error! Just as a handy message. Hopefully it should be pretty clear if we're passing wrong number of arguments, or an invalid argument. Or do you think I should change one of the messages to be different?

simondeziel: The messages are perfectly fine IMHO. It's such a tiny "code dup" that TiCS probably won't penalize us.

skozina: Ah, TiCS, I need to keep that in mind. Thanks for the reminder!

PR #15729, opened 2025-06-05
Makefile: Pin dqlite to last known good version

tomponline:


Makefile
```diff
@@ -104,7 +104,8 @@ dqlite:
 	# dqlite (+raft)
 	@if [ ! -e "$(DQLITE_PATH)" ]; then \
 		echo "Retrieving dqlite from ${DQLITE_BRANCH} branch"; \
-		git clone --depth=1 --branch "${DQLITE_BRANCH}" "https://github.com/canonical/dqlite" "$(DQLITE_PATH)"; \
+		git clone --branch "${DQLITE_BRANCH}" "https://github.com/canonical/dqlite" "$(DQLITE_PATH)"; \
+		git -C "$(DQLITE_PATH)" checkout 8fcffefa8df34628459545786602bbe7c5d839e0 ; \
```
roosterfish, line 108: Worth adding a comment what this is all about? Most probably it will be reverted anyway in a few days but just in case.

tomponline: id like to just get it merged ASAP - we will revert shortly hopefully.

nmezhenskyi (dismissed):
Thanks!

Makefile
simondeziel: FYI, it seems that `--revision=$HASH --depth=1` would still allow to do both shallow while still have a specific commit checked out.

Requires `git` that is newer than 2.49 so may work on GHA runner but not for folks with 22.04 nor 24.04... so not a good suggestion ;)

simondeziel (dismissed):

tomponline:
@simondeziel this isnt working - any ideas why dqlite still appears to being built from main?

roosterfish:
I remember the "Install Go" step/action also has a cache. And because we store the dqlite bits inside the `go/` directory it might be cached in there.

.github/workflows/tests.yml
```diff
@@ -154,6 +154,7 @@ jobs:
           # Build from unpacked dist tarball.
           cd /home/runner/work/lxd/lxd-test
           make deps
+          git -C "/home/runner/work/lxd/lxd-test/vendor/dqlite" log
```
simondeziel: The `make deps` found a `DQLITE_PATH` but it was not a git repo so it kept going without running your `git checkout` of the specific commit.

simondeziel: The cache is working against us here: https://github.com/canonical/lxd/actions/runs/15469677712/job/43550527186?pr=15729#step:16:25

simondeziel: I blew the cache and re-triggered a build.

simondeziel: Found the problem, I think tentative fix in #15731

tomponline:
@simondeziel lets see if this works now.

Makefile
```diff
@@ -104,7 +104,8 @@ dqlite:
...
 	elif [ -e "$(DQLITE_PATH)/.git" ]; then \
 		echo "Updating existing dqlite branch"; \
 		git -C "$(DQLITE_PATH)" pull; \
```
simondeziel, line 111: For completeness, we might want to stub that one out.

simondeziel (approved):
LGTM to unblock CI ASAP.

Makefile
tomponline: stub out?

tomponline:
@simondeziel ready for approval as cluster tests are passing

tomponline:
@simondeziel thanks you approved already :)

PR #15731, opened 2025-06-05
Makefile: pin dqlite to last known good commit in `make dist`

simondeziel:


simondeziel:
https://github.com/canonical/lxd/actions/runs/15470029028/job/43551692499?pr=15731#step:14:469 shows that it worked.

simondeziel:
#15729 has a more complete fix, let's carry on with it.

PR #15734, opened 2025-06-05
Make it easy to build and test LXD

simondeziel:
Here what the workflow look like:

```
$ lxc launch ubuntu-minimal-daily:24.04 v1 --vm -p lxd-test
$ lxc shell v1
root@v1:~# cd lxd && make deps && make && make test-shell
LXD-TEST root@v1:~/lxd/test$ 
```

### DIFF
--- a/doc/contributing.md
+++ b/doc/contributing.md
@@ -50,7 +50,7 @@ git push -u myfork <name_of_your_new_branch>
 ### Important notes for new LXD contributors
 
 - Persistent data is stored in the `LXD_DIR` directory, which is created by running `lxd init`.
-   - By default, `LXD_DIR` is located at `/var/lib/lxd` (for non-snap installations) or `/var/snap/lxd/common/lxd` (f or snap users).
+   - By default, `LXD_DIR` is located at `/var/lib/lxd` (for non-snap installations) or `/var/snap/lxd/common/lxd` (for snap users).
    - To prevent version conflicts, consider setting a separate `LXD_DIR` for your development fork.
 - Binaries compiled from your source are placed in `$(go env GOPATH)/bin` by default.
    - When testing, explicitly invoke these binaries instead of the global `lxd` you might have installed.


test/main.sh
tomponline, line 1: > if that dir is backed by `virtiofs`, the tests fail.

How so?

test/main.sh
```diff
@@ -176,7 +176,7 @@ TEST_DIR="$(mktemp -d -t lxd-test.tmp.XXXX)"
 chmod +x "${TEST_DIR}"
 
 # Verify the dir chain is accessible for other users
-INACCESSIBLE_DIRS="$(namei -m "${TEST_DIR}" | awk '/^ d/ {print $1}' | grep -v 'x$' || true)"
+INACCESSIBLE_DIRS="$(namei -m "${TEST_DIR}" | awk '/^ d/ {print $1}' | grep -v '[xt]$' || true)"
```
tomponline, line 183: What is this fixing? Can we add a comment explaining the cryptic command please?

test/main.sh
```diff
@@ -273,13 +273,22 @@ if [ -n "${GITHUB_ACTIONS:-}" ]; then
...
 # allow for running a specific set of tests
-if [ "$#" -gt 0 ] && [ "$1" != "all" ] && [ "$1" != "cluster" ] && [ "$1" != "standalone" ]; then
+if [ "$#" -gt 0 ] && [ "$1" != "all" ] && [ "$1" != "cluster" ] && [ "$1" != "standalone" ] && [ "$1" != "tshell" ]; then
   run_test "test_${1}"
   # shellcheck disable=SC2034
   TEST_RESULT=success
   exit
 fi
 
+if [ "${1:-"all"}" = "tshell" ]; then
```
tomponline: What is this and why do we need it? Commit message is lacking explanation.

Makefile
```diff
@@ -175,6 +175,11 @@ deps: dqlite liblxc
 	@echo "# Please set the following in your environment (possibly ~/.bashrc)"
 	@make -s env
 
+.PHONY: tshell
```
tomponline: What is a tshell?

doc/lxd-test.yaml
```diff
@@ -0,0 +1,102 @@
+name: lxd-test
+description: "Profile to build and test LXD in ephemeral VM"
+config:
+  cloud-init.user-data: |-
+    #cloud-config
+    apt:
+      # Speed things up by not pulling from backports/security and avoid restricted/multiverse pockets.
+      # In general, backported packages or those from restricted/multiverse shouldn't be relied on because
+      # they don't come with the same level of support as those from main for example.
+      # The security repo doesn't make much sense when pulling from a Canonical maintained archive mirror.
```
tomponline: Why?

doc/lxd-test.yaml
```diff
@@ -0,0 +1,102 @@
...
+    - shellcheck
+    - socat
+    - sqlite3
+    - swtpm
+    - yq
+    snap:
+      commands:
+      - snap install go --classic
+  limits.cpu: "4"
+  limits.memory: 16GiB
```
tomponline, line 91: This seems excessive and many users will have to reduce this (myself included!)

CONTRIBUTING.md
```diff
@@ -132,6 +132,51 @@ If you add or update configuration options, regenerate and commit the documentat
...
+sed "s|@@PATH_TO_LXD_GIT@@|${GIT_ROOT}|" "${GIT_ROOT}/doc/lxd-test.yaml" | lxc profile create lxd-test
+```
+
+This profile can then be used to launch an Ubuntu Noble VM and start using it:
+
+```sh
+lxc launch ubuntu-minimal-daily:24.04 v1 --vm -p lxd-test
+sleep 30
+# this may take a while as many packages need to be installed
+lxc exec v1 -- cloud-init status --wait --long
```
tomponline, line 164: Would it be more transparent for the cloudinit config to write a script which sets up the dep packages rather than installing them directly, then the user can be instructed to run this script and they can see how things are progressing?

test/main.sh
simondeziel: Yeah, I need to fill a bug for Aleks to have a proper look. He already expressed interest in fixing this behavior which IMHO will probably also help other use cases.

That said, having `tmp.XXX` dirs littering our local git clone still feels wrong to me.

doc/lxd-test.yaml
simondeziel: This is to align with CPU/memory specs of GH action runners.

Would you like me to update the `sed` command doing the mangling to also tweak those 2 limits based on user provided variables?

CONTRIBUTING.md
simondeziel: I'd rather not as that adds an extra step to be typed by the user. This whole PR is aiming for the easiest/quickest way to get a throw away env that is 100% clean. I use that multiple times a day so every bit count IMHO.

doc/lxd-test.yaml
kadinsayani: Can we also make the root disk size configurable? 12GiB also seems excessive since the only mount is the lxd directory.

```bash
❯ du -h --max-depth=0 ./lxd
870M    ./lxd
```

CONTRIBUTING.md
```diff
@@ -132,6 +132,51 @@ If you add or update configuration options, regenerate and commit the documentat
 
     make update-metadata
 
+#### Development environment setup
+
+Several pieces of software are needed in order to build and test LXD. Here is an easy way to create a virtual-machine to use as development environment. LXD itself is needed to power that virtual-machine so install it first: {ref}`installing`.
```
kadinsayani:
```suggestion
Several pieces of software are needed in order to build and test LXD. Here is an easy way to create a virtual-machine to use as a development environment. LXD itself is needed to power that virtual-machine so install it first: {ref}`installing`.
```

CONTRIBUTING.md
```diff
@@ -132,6 +132,51 @@ If you add or update configuration options, regenerate and commit the documentat
...
+
+Once LXD is installed and {ref}`initialized <initialize>`, a special profile (`lxd-test`) needs to be loaded:
+
+```sh
```
kadinsayani: This `sh` block is empty.

CONTRIBUTING.md
```diff
@@ -132,6 +132,51 @@ If you add or update configuration options, regenerate and commit the documentat
...
+```
+
+The profile requires includes a `lxd-git` device that will share LXD's git repository with the virtual-machine. Since this path is specific to your environment you need to adjust it when loading the profile:
```
kadinsayani: Perhaps we can link to https://documentation.ubuntu.com/lxd/latest/reference/devices_disk/#devices-disk here.

CONTRIBUTING.md
```diff
@@ -132,6 +132,51 @@ If you add or update configuration options, regenerate and commit the documentat
...
+```sh
+# this needs to be run from inside the git repostory
+GIT_ROOT="$(git rev-parse --show-toplevel)"
+# delete any old copy of the profile
+lxc profile list | grep -qwF lxd-test && lxc profile delete lxd-test
+# create the profile based on the provided template
+sed "s|@@PATH_TO_LXD_GIT@@|${GIT_ROOT}|" "${GIT_ROOT}/doc/lxd-test.yaml" | lxc profile create lxd-test
+```
+
+This profile can then be used to launch an Ubuntu Noble VM and start using it:
```
kadinsayani, line 158:
```suggestion
This profile can then be used to launch an Ubuntu VM and start using it:
```

Will this also work with older releases?

CONTRIBUTING.md
```diff
@@ -132,6 +132,51 @@ If you add or update configuration options, regenerate and commit the documentat
...
+# go into the git repo
+root@v1:~# cd lxd
+
+# build deps and LXD binaries
+root@v1:~/lxd# make deps && make
+
+# get an interactive test shell session with all the needed environment variables to use and test LXD
+root@v1:~/lxd# make tshell
+```
+
```
kadinsayani, line 182: While we are here, would it be helpful to add a link to https://documentation.ubuntu.com/lxd/latest/debugging/ somewhere in this file?

kadinsayani:
What happens to the shell running lxd after the test exits (by ^C)? Will the normal cleanup steps happen?

doc/lxd-test.yaml
tomponline: We dont need to align to gh runners, this large seems like it would be a barrier for people copy pasting to get up n running.

CONTRIBUTING.md
tomponline: Ack

doc/lxd-test.yaml
simondeziel: There is a lot to unpack here. Let's start by pointing out that there is a precedent in `lxd-ci.git`: https://github.com/canonical/lxd-ci?tab=readme-ov-file#lxd-ci-profile  (not that it's a good justification ;) )

So packages from `backports` or `multiverse` are not supported by Canonical, only `main` and `restricted` are (ignoring ESM/Pro). `restricted` is supported but is for proprietary software. In both cases, we don't need any of that as we only depend on bits from `main` and `universe` which are the 2 "pockets" not disabled.

As for why it's safe to `security` repo? Let's first start by why is `security.ubuntu.com` part of every default install. That is specifically to avoid a malicious apt mirror (country specific or other, like say `http://mirrors.dc.clear.net.ar/ubuntu/` to pick on the first one from list here https://launchpad.net/ubuntu/+archivemirrors). If that malicious mirror stops syncing with Ubuntu primary archives with the intention of serving old/stale/known vulnerable package, the apt client would still get the patched version from `security.ubuntu.com` that comes from Canonical directly.

Now in our case, `ubuntu:` images all pull packages from `archive.ubuntu.com` which is not a random mirror but the primary source of update, on par with `security.ubuntu.com` in terms of trust.

So all that to say that it's OK to "cut those corners" as we only save time, not reliability not security.

doc/lxd-test.yaml
simondeziel: @kadinsayani while the git repo itself is smallish it shouldn't even matter as it comes from your host directly. The real use of the 12GiB is for tests that will consume space in the guest VM.

I have no problem making it tunable but I never felt the need here, probably because I use ZFS which doesn't really mind about this.

CONTRIBUTING.md
simondeziel: It has to be Noble. If you do go with 22.04 you miss the `yq` .deb package so you need to pull the snap.

If you go with a newer release, you are in uncharted territories. Should work but might run into some packages needing to be have their name updated.

CONTRIBUTING.md
simondeziel: Ah thanks, fixed.

doc/lxd-test.yaml
simondeziel: I settled with having another code block where the CPU, memory and disk size are tweaked for a more modest machine.
I figured nobody likes `sed` ;)

test/main.sh
simondeziel: This is mostly a guard rail in case the needed `sudo chmod +x ~` was not run.

test/main.sh
simondeziel: It gives you an interactive "test shell session". I augmented the commit messages and added comments in the script and `Makefile`.

simondeziel:
> What happens to the shell running lxd after the test exits (by ^C)? Will the normal cleanup steps happen?

Good catch, I now added a reentry protection by deactivating all traps when `cleanup()` kicks in. Out of curiosity, did you take the changes for a test drive and ran into this? Or you just have super review powers?

kadinsayani:
> Good catch, I now added a reentry protection by deactivating all traps when `cleanup()` kicks in. Out of curiosity, did you take the changes for a test drive and ran into this? Or you just have super review powers?

'Twas a lucky catch 😉

test/main.sh
tomponline: can it have a comment explaining that?

test/main.sh
tomponline: can we call it `make testshell` to be clearer?

doc/lxd-test.yaml
tomponline: Thanks for the explanation, can we extend slightly the line "The security repo doesn't make much sense when pulling from a Canonical maintained archive mirror." to explain why it doesnt make sense as this was the bit that confused me. Thanks

tomponline:
@simondeziel please can you rebase too

doc/lxd-test.yaml
```diff
@@ -0,0 +1,102 @@
...
+    # Faster dpkg installs
+    - content: "force-unsafe-io\n"
+      path: /etc/dpkg/dpkg.cfg
+      append: true
+    # wait for cloud-init to be done
+    - content: "#!/bin/sh\nexec cloud-init status --wait --long\n"
+      path: /usr/local/bin/ci-wait
+      permissions: "0755"
+
+    runcmd:
```
skozina, line 29: Where did the `chmod -x /usr/sbin/lxc /usr/sbin/lxd` go? It is in your (excellent!) lxd-ci profile in lxd-ci.

doc/lxd-test.yaml
```diff
@@ -0,0 +1,102 @@
...
+    - systemctl stop unattended-upgrades.service
+    - apt-get autopurge -y cron lxd-installer needrestart networkd-dispatcher unattended-upgrades
+    - cd /etc/systemd/system/timers.target.wants/ && systemctl disable --now *.timer
+    # Prepare LXD build
+    - printf '[safe]\n    directory = *\n' >> /root/.gitconfig
+    - chmod o+x /root
+
+    package_update: true
+    package_upgrade: true
+    packages:
```
skozina, line 44: Can we also remove `uidmap` package as part of the setup? That was sometimes causing the uid remap errors.

doc/lxd-test.yaml
```diff
@@ -0,0 +1,102 @@
...
+    - busybox-static
+    - curl
+    - dnsmasq-base
+    - iptables
+    - jq
+    - shellcheck
+    - socat
+    - sqlite3
+    - swtpm
+    - yq
```
skozina, line 86: Bold ask -- can we include `vim` in the list, even if it's not needed?

skozina (approved):
Few nits, I really appreciate we could build a standard test profile;-) I was using the one from lxd-ci since I found it... Thanks!

doc/lxd-test.yaml
kadinsayani: +1 for this, can we also set `environment.EDITOR` to `vim`?

kadinsayani: Or we can just inherit `EDITOR` from the host.

simondeziel: Both good suggestions and you've hit a sweet spot as I've been annoyed myself due to the lack of `vim` and the annoying presence of `nano` ;)

Will do.

test/main.sh
simondeziel: I'll test with `LXD_TMPFS=1` as that might sidestep the `virtiofs` issue.

test/main.sh
simondeziel: Naming is hard. It started as `tash` for `test + bash`, then felt that `tshell` was more descriptive while still being short but I guess we go all in with explicit and use `test-shell`. I added the `-` as our other targets separate words with it already.

CONTRIBUTING.md
```diff
@@ -132,6 +132,56 @@ If you add or update configuration options, regenerate and commit the documentat
...
+# create the profile based on the provided template
+sed "s|@@PATH_TO_LXD_GIT@@|${GIT_ROOT}|" "${GIT_ROOT}/doc/lxd-test.yaml" | lxc profile create lxd-test
+```
+
+The `lxd-test` profile assigns CPU and memory limits similar to those available in free GitHub Action runners. Those can be adapted to the specifications of a more modest physical machine:
+
+```sh
+lxc profile set lxd-test limits.cpu=2
+lxc profile set lxd-test limits.memory=4GiB
+lxc profile set lxd-test root size=8GiB
```
kadinsayani:
```suggestion
lxc profile device set lxd-test root size=8GiB
```

test/main.sh
tomponline: `make test-shell` sounds good.

`make tash` lol

test/main.sh
tomponline: good idea, we should do that for sure as it speeds things up a lot

CONTRIBUTING.md
```diff
@@ -132,6 +132,56 @@ If you add or update configuration options, regenerate and commit the documentat
...
+sleep 30
+# this may take a while as many packages need to be installed
+lxc exec v1 -- cloud-init status --wait --long
+```
+
+Then it is possible to build all the dependencies, LXD binaries and even run tests either automatically or manually:
+
+```sh
+# enter the VM
+$ lxc exec v1 -- bash
```
kadinsayani:
```suggestion
lxc exec v1 -- bash
```

simondeziel: I usually am with you in not including `$ ` as that makes copy pasting bloody annoying but I wanted to include the PS1 in those. Now looking back, I agree with you. Will drop them.

doc/lxd-test.yaml
simondeziel: This package is not in `ubuntu-minimal-daily:24.04` and I just checked that nothing installed by the `lxd-test` profile bring it along with it.

doc/lxd-test.yaml
simondeziel: So our minimal images don't even have `nano` so simply bringing `vim` along will do the right thing. No need to fiddle with the `EDITOR` env as `vim` seems to be used by default as the only one present.

I test with a `git commit --amend` and got `vim`.

test/main.sh
simondeziel: `LXD_TMPFS=1` avoids the `virtiofs` issue. However, even if `LXD_TMPFS=1` is desirable, I don't feel like `make test-shell` should set this environment variable, it has to be a conscious decision IMHO.

WDYT?

tomponline: Id default it on if it works as its a controlled environment a preferable to be enabled.

simondeziel: We enable it in GH CI too so should we default to having this flag enabled? I still believe `test-shell` shouldn't add much magic on its own.

tomponline: Can you add it to bash profile in cloud-init?

simondeziel: I can and will do that but I'm still wondering why not change the default at this point ;)

tomponline: im not sure

simondeziel: Ah shoot, I just realized that using `environment.LXD_TMPFS` would break when using `lxc shell` as that starts with clean env :/

simondeziel: NVM, will use the same trick as in `lxd-ci`'s profile:

```
    runcmd:
    - echo "LXD_TMPFS=1" >> /etc/environment
```

simondeziel:
@tomponline I think it's ready for a (final?) round of review. I've addressed all the feedback point and am waiting on this PR to be merged to report the `virtiofs` issue as this will simplify the reproducer.

CONTRIBUTING.md
```diff
@@ -132,6 +132,56 @@ If you add or update configuration options, regenerate and commit the documentat
...
+lxc exec v1 -- bash
+
+# go into the git repo
+cd lxd
+
+# build deps and LXD binaries
+make deps && make
+
+# get an interactive test shell session with all the needed environment variables to use and test LXD
+make test-shell
```
tomponline, line 180: it'd be good to give an example of what you can do once you're in the shell in my view.

tomponline (approved):
LGTM thanks!

CONTRIBUTING.md
simondeziel: https://github.com/canonical/lxd/pull/15764

test/main.sh
simondeziel: > > if that dir is backed by `virtiofs`, the tests fail.
> 
> How so?

I finally got round to reporting it with a simple reproducer: https://github.com/canonical/lxd/issues/15803

PR #15736, opened 2025-06-05
lxd/images: accumulate node IDs while building the node list

simondeziel:


kadinsayani (approved):

tomponline:
Personally I wouldn't add a change like this because whilst it saves a loop, the number of iterations are small (nodes are usually less than 50) and this makes refactoring code more complicated, whereas how it is now the scope of the integer nodes representation is limited to where its needed, and the primary representation (the slice of structs) is what is carried throughout the function.

simondeziel:
> Personally I wouldn't add a change like this because whilst it saves a loop, the number of iterations are small (nodes are usually less than 50) and this makes refactoring code more complicated, whereas how it is now the scope of the integer nodes representation is limited to where its needed, and the primary representation (the slice of structs) is what is carried throughout the function.

I totally agree and now want to revert it :/

PR #15726, opened 2025-06-05
cluster: Fix panic during exclusive DB mode

tomponline:
Replaces `EnterExclusive` and `ExitExclusive` with `RunExclusive` which takes a function as an argument to run inside the exclusive mode. The `RunExclusive` function takes care of acquiring and releasing the lock, which simplifies what the caller has to care about.

Fixes panic caused by error during cluster exclusive DB mode:

```
fatal error: sync: Unlock of unlocked RWMutex

goroutine 453 gp=0xc0004e7500 m=3 mp=0xc000089008 [running]:
runtime.fatal({0x26eb90a, 0x20})
	runtime/panic.go:1123 +0x5c fp=0xc00143f1d8 sp=0xc00143f1a8 pc=0x45a09c
sync.fatal({0x26eb90a?, 0x2b8ccb0?})
	runtime/panic.go:1038 +0x18 fp=0xc00143f1f8 sp=0xc00143f1d8 pc=0x492df8
sync.(*RWMutex).Unlock(0xc000d6d710)
	sync/rwmutex.go:212 +0x45 fp=0xc00143f230 sp=0xc00143f1f8 pc=0x4a71e5
github.com/canonical/lxd/lxd/db.(*Cluster).ExitExclusive.deferwrap1()
	github.com/canonical/lxd/lxd/db/db.go:362 +0x25 fp=0xc00143f248 sp=0xc00143f230 pc=0xe2fe65
github.com/canonical/lxd/lxd/db.(*Cluster).ExitExclusive(0xc000d6d700, {0x2b8ccb0, 0x44f5560}, 0x2834470)
	github.com/canonical/lxd/lxd/db/db.go:363 +0xc5 fp=0xc00143f2b8 sp=0xc00143f248 pc=0xe2fde5
github.com/canonical/lxd/lxd/cluster.Join.func3()
	github.com/canonical/lxd/lxd/cluster/membership.go:408 +0x38 fp=0xc00143f318 sp=0xc00143f2b8 pc=0x14b6418
github.com/canonical/lxd/shared/revert.(*Reverter).Fail(0xc00143f5c0)
```

PR #15743, opened 2025-06-07
lxd/db: Rework RunExclusive to use context with timeout internally

tomponline:
This is to avoid a potential race where the go routine takes the mutex and tries to send the acquired message to the outer function before it is waiting on the channel and then incorrectly detects this as the outer function having returned, which then results in the outer function waiting until the 20s timeout and returning an error.

Fixes issue introduced by https://github.com/canonical/lxd/pull/15726

### DIFF
--- a/lxd/db/db.go
+++ b/lxd/db/db.go
@@ -349,20 +349,24 @@ func (c *Cluster) RunExclusive(f func(t Transactor) error) error {
 		c.mu.Unlock()
 	}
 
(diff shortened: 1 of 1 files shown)


simondeziel:
You probably know but just in case, the `main` branch also ran into this during unit tests: https://github.com/canonical/lxd/actions/runs/15513043189/job/43676314476#step:24:738

PR #15745, opened 2025-06-09
lxc/remote/add: Fixed runtime error (index out of range [0] with length 0)

TanishkBansode:
When run,` lxc remote add <name> <url>`
It outputs this and waits for input

```
Certificate fingerprint: <fingerprint>
ok (y/n/[fingerprint])?
```
but when press just enter without inputting any character it return error, 

```
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
main.(*cmdRemoteAdd).run(0xc00016f2d0, 0xc00023da10?, {0xc00040e620, 0x2, 0x0?})
        github.com/canonical/lxd/lxc/remote.go:534 +0x31b0
github.com/spf13/cobra.(*Command).execute(0xc0003f8608, {0xc00040e5c0, 0x2, 0x2})
        github.com/spf13/cobra@v1.8.1/command.go:985 +0xaaa
github.com/spf13/cobra.(*Command).ExecuteC(0xc000300f08)
        github.com/spf13/cobra@v1.8.1/command.go:1117 +0x3ff
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra@v1.8.1/command.go:1041
main.main()
        github.com/canonical/lxd/lxc/main.go:294 +0x1a73
```
So i fixed it, to check if the length is greater than 0, and then read the line.


tomponline:
Hi @TanishkBansode please can you sign (as well as sign off) your commits?

lxc/remote.go
```diff
@@ -529,23 +529,24 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
...
+						return errors.New(i18n.G("The provided fingerprint does not match the server certificate fingerprint"))
+					}
 
-				// Error out if the user didn't confirm the fingerprint.
-				if len(line) < 1 || strings.ToLower(string(line[0])) == i18n.G("n") {
-					return errors.New(i18n.G("Server certificate NACKed by user"))
+					// Error out if the user didn't confirm the fingerprint.
+					if len(line) < 1 || strings.ToLower(string(line[0])) == i18n.G("n") {
+						return errors.New(i18n.G("Server certificate NACKed by user"))
+					}
```
MusicDin: The check `len(line) < 1` is no longer needed here

MusicDin: Also to prevent changing the behavior of the current CLI, we should error out if `len(line) == 0`

MusicDin: IMO we should also error out if no input was provided (`len(line) == 0`)

lxc/remote.go
```diff
@@ -530,22 +530,25 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
 					return err
 				}
 
-				// Continue with adding the remote if digest matches, or the user
-				// confirmed a fingerprint.
-				if string(line) == digest || strings.ToLower(string(line[0])) == i18n.G("y") {
-					break
-				}
+				// Check length before accessing line [0]
```
kadinsayani: This comment tells readers what the check is doing, but it doesn't address _why_ this check is required. 

```suggestion
				// Check length before accessing line to prevent runtime panic.
```

lxc/remote.go
```diff
@@ -530,22 +530,25 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
...
+				if len(line) > 0 {
+					// Continue with adding the remote if digest matches, or the user
+					// confirmed a fingerprint.
+					if string(line) == digest || strings.ToLower(string(line[0])) == i18n.G("y") {
+						break
+					}
```
kadinsayani:
```suggestion
					// Continue with adding the remote if digest matches, or the user
					// confirmed a fingerprint.
					if string(line) == digest || (len(line) > 0 && strings.ToLower(string(line[0])) == i18n.G("y")) {
						break
					}
```

lxc/remote.go
```diff
@@ -530,22 +530,25 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
...
+						return errors.New(i18n.G("The provided fingerprint does not match the server certificate fingerprint"))
+					}
 
-				// Error out if the user didn't confirm the fingerprint.
-				if len(line) < 1 || strings.ToLower(string(line[0])) == i18n.G("n") {
-					return errors.New(i18n.G("Server certificate NACKed by user"))
+					// Error out if the user didn't confirm the fingerprint.
+					if len(line) < 1 || strings.ToLower(string(line[0])) == i18n.G("n") {
+						return errors.New(i18n.G("Server certificate NACKed by user"))
+					}
```
kadinsayani:
```suggestion
					// Error out if the user didn't confirm the fingerprint.
					if len(line) == 0 || strings.ToLower(string(line[0])) == i18n.G("n") {
						return errors.New(i18n.G("Server certificate NACKed by user"))
					}
```

kadinsayani (changes requested):
Thanks for fixing this bug! I just have a few suggestions to simplify the logic and ensure we error out when `len(line) == 0`.

kadinsayani:
@TanishkBansode instead of adding new commits for the suggestions, could you please rebase and amend the original commit? The fix should ideally be contained within a single commit.

TanishkBansode:
Hi ,
I've rebased it and turned the changes into a single commit.
If I’ve misunderstood anything or done something incorrectly, please let me know. Thanks!

lxc/remote.go
```diff
@@ -544,9 +545,10 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
 				}
 
 				// Error out if the user didn't confirm the fingerprint.
-				if len(line) < 1 || strings.ToLower(string(line[0])) == i18n.G("n") {
+				if len(line) == 0 || strings.ToLower(string(line[0])) == i18n.G("n") {
 					return errors.New(i18n.G("Server certificate NACKed by user"))
 				}
+
```
MusicDin: Please remove this extra empty line

MusicDin (dismissed):
Thank you 🚀 There's just one extra empty line that needs to be removed

lxc/remote.go
TanishkBansode: Done 👍

kadinsayani: Thanks, you may need to run `gofmt` on the file. It looks like there is still some extra spacing.

tomponline:
Thanks!

PR #15762, opened 2025-06-09
lxd: Replace request context keys with request info

MusicDin:
This PR introduces request info which is a structure set on a request context that wraps commonly used context values. 

Request info replaces majority of request context key constants, and removes the need to parse each request context value into a particular type (now only the `RequestInfo` needs to be parsed). This simplifies addition of new context keys, as they can be added directly to the request info.

Additionally, it enables reliable checks for request context (by checking for the presence of the `RequestInfo` instead of `CtxRequestSourceAddress`).


tomponline:
Please rebase

lxd/daemon.go
```diff
@@ -818,6 +818,13 @@ func (d *Daemon) createCmd(restAPI *mux.Router, version string, c APIEndpoint) {
 			return
 		}
 
+		// Fetch the existing context info (in case it was already set) or create a new one.
```
tomponline: Can this code (that is duplicated below) be added to a function in the `request` package called `SetupContext()` or similar?

lxd/request/context.go
```diff
@@ -34,6 +28,32 @@ func GetCtxValue[T any](ctx context.Context, key CtxKey) (T, error) {
...
+	if ctx == nil {
+		return false
+	}
+
+	_, ok := GetCtxInfo(ctx)
+	return ok
+}
+
+// GetCtxInfo gets the request information from the request context.
+func GetCtxInfo(reqContext context.Context) (*Info, bool) {
```
tomponline: It appears to be that in Go function names about contexts use the word "Context" rather than "Ctx", and its only the argument names that use `ctx`. So I think this should be `GetContextInfo`, `ContextInfo` or perhaps just `Info`.

lxd/request/context.go
```diff
@@ -34,6 +28,32 @@ func GetCtxValue[T any](ctx context.Context, key CtxKey) (T, error) {
...
+	info, ok := reqContext.Value(CtxRequestInfo).(*Info)
+	if !ok {
+		return &Info{}, false
+	}
+
+	return info, true
+}
+
+// SetCtxInfo sets the request information in the request context.
+func SetCtxInfo(r *http.Request, info *Info) {
```
tomponline: `SetContextInfo` like `IsRequestContext`

lxd/request/context.go
```diff
@@ -34,6 +28,32 @@ func GetCtxValue[T any](ctx context.Context, key CtxKey) (T, error) {
...
+
+	_, ok := GetCtxInfo(ctx)
+	return ok
+}
+
+// GetCtxInfo gets the request information from the request context.
+func GetCtxInfo(reqContext context.Context) (*Info, bool) {
+	info, ok := reqContext.Value(CtxRequestInfo).(*Info)
+	if !ok {
+		return &Info{}, false
```
tomponline: Suggest just return nil if not and then there's no need for the bool return param.

lxd/request/request.go
```diff
@@ -22,7 +24,7 @@ func CreateRequestor(ctx context.Context) *api.EventLifecycleRequestor {
 		requestor.Protocol = val
 	}
 
-	requestor.Address, _ = ctx.Value(CtxRequestSourceAddress).(string)
+	requestor.Address = info.SourceAddress
```
tomponline: This would then become:

```go
if info != nil {
    requestor.Address = info.SourceAddress
}
```

lxd/request/context.go
MusicDin: A small nit, but my initial idea was to always return a non-nil value without a boolean, as this allows direct value access when we are not interested in checking whether it is a request context:
```go
trusted := request.GetContextInfo(ctx).Trusted
```

In case we want to check whether it is a request context, we can use `request.IsRequestContext(ctx)`.

Returning nil adds some boiler plate when for example retrieving `EffectiveProjectName`:
```go
var effectiveProjectName string
info := request.GetContextInfo(ctx)
if info != nil {
  effectiveProjectName = info.EffectiveProjectName
}

// Instead we could use.
effectiveProjectName := request.GetContextInfo(ctx).EffectiveProjectName
```

lxd/request/context.go
MusicDin: Actually, the outlier here is `request.IsRequestContext`, as we already have `[Get/Set]CtxValue`. Should I also rename those for consistency? -> `IsRequestContext`, `SetContextInfo`, `SetContextValue`, `GetContextValue`, ...

tomponline: Yes please.

Its not really the outlier though, at least in stdlib, because there is:

- https://pkg.go.dev/net/http#NewRequestWithContext
- https://pkg.go.dev/net/http#Request.Context
- https://pkg.go.dev/golang.org/x/sync/errgroup#WithContext

etc

lxd/request/context.go
tomponline: Hrm, is the point of `GetContextInfo()` that you can modify it and it is then persisted into the context because it is a pointer?

If so, then returning a pointer to an empty struct that is not stored in the context seems like a potential source of weird bugs where the caller may get confused as to whether the pointer returned is actually from the context or not and modify it only to see that its not available later.

lxd/request/context.go
MusicDin: > Its not really the outlier though, at least in stdlib, because there is:

I meant in our `request` package.

Will change 👍

lxd/request/context.go
MusicDin: I cannot argue that. Will fix :)

lxd/cluster/connect.go
```diff
@@ -57,26 +57,14 @@ func Connect(ctx context.Context, address string, networkCert *shared.CertInfo,
...
-				req.Header.Add(request.HeaderForwardedUsername, val)
-			}
-
-			val, ok = ctx.Value(request.CtxProtocol).(string)
-			if ok {
-				req.Header.Add(request.HeaderForwardedProtocol, val)
-			}
-
 			req.Header.Add(request.HeaderForwardedAddress, reqInfo.SourceAddress)
+			req.Header.Add(request.HeaderForwardedUsername, reqInfo.Username)
```
tomponline, line 61: This is a subtle change in behaviour that would be useful to draw attention to in a separate commit rather tha n bundling in with other changes.

Previously the header was only added if the context contained the fields in questions, now we always add the headers just with empty string values.

Is this correct?

lxd/auth/drivers/tls_test.go
tomponline, line 1: Please try and avoid having commits that span multiple packages. Thanks

lxd/request/context.go
tomponline, line 1: Just a word of caution here, its alright at the moment, but something to be mindful of in the future is that `lxd/request` is used by lxd-agent, which means we need to be careful about the imports this package uses, right now its all stdlib so no problem, but we may be tempted to include fields that are non-stdlib types in the future which we will have to avoid.

MusicDin: Agree, this is why we avoided adding any fields that represent internal structures in request Info.

lxd/cluster/connect.go
MusicDin: Previously, the header was set only if the value was found in the context, which may result in the header not being set.

However, when we read the values from the header in the handler, the values will be set to an empty string if the header is not found.

https://github.com/canonical/lxd/blob/299dcbd5ba48df3450b83069de7dd3804eb94691/lxd/daemon.go#L832-L837

Not setting the headers would not change the functionality, but it would prevent sending empty headers.

In short, good spot, I'll fix this.

PR #15764, opened 2025-06-09
Improve `make test-shell` a bit

simondeziel:
Add examples of how to make use of the interactive `test-shell` feature.
Add necessary bits to have `lxc` command completions working.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -178,6 +178,19 @@ make deps && make
 
 # get an interactive test shell session with all the needed environment variables to use and test LXD
 make test-shell
+
+# run the `exec` and `query` tests
+./main.sh exec
+./main.sh query
+
+# or manually interact with LXD, for example:
+lxc launch ubuntu:24.04 u1
+lxc exec u1 -- hostname
+lxc delete --force u1
+
+# for a barebones test instance with just busybox (note: no IP automatically configured)
+./deps/import-busybox --alias testimage
+lxc launch testimage c1
 ```
 
 At this point you might want to learn more on {doc}`debugging`.
--- a/test/main.sh
+++ b/test/main.sh
@@ -289,7 +289,7 @@ fi
 if [ "${1:-"all"}" = "test-shell" ]; then
   # yellow
   export PS1="\[\033[0;33mLXD-TEST\033[0m ${PS1:-\u@\h:\w\$ }\]"
-  bash --norc
+  bash --norc || true
   # shellcheck disable=SC2034
   TEST_RESULT=success
   exit
--- a/test/main.sh
+++ b/test/main.sh
@@ -223,7 +223,7 @@ run_test() {
   TEST_CURRENT=${1}
   TEST_CURRENT_DESCRIPTION=${2:-${1#test_}}
   TEST_UNMET_REQUIREMENT=""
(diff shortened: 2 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -178,6 +178,15 @@ make deps && make
 
 # get an interactive test shell session with all the needed environment variables to use and test LXD
 make test-shell
+
+# run the `exec` and `query` tests
```
kadinsayani, line 182:
```suggestion
cd test/
# run the `exec` and `query` tests
```

CONTRIBUTING.md
```diff
@@ -178,6 +178,15 @@ make deps && make
...
+./main.sh exec
+./main.sh query
+
+# or manually interact with LXD
```
kadinsayani:
```suggestion
# or manually interact with LXD, for example:
```

CONTRIBUTING.md
simondeziel: `make test-shell` already drops you in that `test/` dir.

kadinsayani: Awesome!

kadinsayani (approved):

kadinsayani (dismissed):

test/main.sh
```diff
@@ -289,7 +289,7 @@ fi
 if [ "${1:-"all"}" = "test-shell" ]; then
   # yellow
   export PS1="\[\033[0;33mLXD-TEST\033[0m ${PS1:-\u@\h:\w\$ }\]"
-  bash --norc
+  bash --rcfile test-shell.bashrc || true
```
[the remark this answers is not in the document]
simondeziel: The error being silenced is not due to a potentially missing RC file but to avoid triggering the `cleanup` helper that is called due to `trap`'ing `EXIT` signals.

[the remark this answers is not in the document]
simondeziel: The `main.sh` script already assumed it was only run from the `test` directory. I've now changed it so that is `cd` into that dir if not done when called like `./test/main.sh exec`.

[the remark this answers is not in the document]
simondeziel: I've improved it a bit by having `main.sh` check early if the current dir is the expected one and if not change to `dirname $0`.

simondeziel:
@tomponline ready for a final review.

test/main.sh
```diff
@@ -289,7 +289,7 @@ fi
 if [ "${1:-"all"}" = "test-shell" ]; then
   # yellow
   export PS1="\[\033[0;33mLXD-TEST\033[0m ${PS1:-\u@\h:\w\$ }\]"
-  bash --norc
+  bash --norc || true
```
tomponline: for future reference I'd encourage you to add comments to the code explaining this rather (or at least only) the commit message, as for future readers it will not be clear why this is here. We should prioritise comments in code first before (but not excluding) detailed explanations in commit messages as they may be obscured over time by future changes.

tomponline (approved):
Thanks!

The docs will be super useful for folks using this new feature you've added.

test/main.sh
simondeziel: https://github.com/canonical/lxd/pull/15780

PR #15774, opened 2025-06-11
devlxd: Fix regression in LXD agent where string responses are escaped

MusicDin:
When devLXD is sending string content over vsock, the `(api.Response).Metadata` contains JSON encoded (escaped string) content. This affects LXD-agent endpoints `/1.0/meta-data` and `/1.0/config/<key>`, which also resulted in broken cloud-init LXD data source. The unix socket (container) responses are unaffected.

We handle two affected endpoints in devLXD client by unmarshaling the content into a string when response is received over vsock.

### DIFF
--- a/client/devlxd_metadata.go
+++ b/client/devlxd_metadata.go
@@ -1,6 +1,7 @@
 package lxd
 
 import (
+	"encoding/json"
 	"net/http"
 )
 
@@ -9,6 +10,18 @@ func (r *ProtocolDevLXD) GetMetadata() (metadata string, err error) {
 	resp, _, err := r.query(http.MethodGet, "/meta-data", nil, "")
 	if err != nil {
 		return "", err
+	}
+
+	if r.isDevLXDOverVsock {
+		var metadata string
+
+		// The returned string value is JSON encoded.
+		err = json.Unmarshal(resp.Content, &metadata)
+		if err != nil {
+			return "", err
+		}
+
+		return metadata, nil
 	}
 
 	return string(resp.Content), nil


client/devlxd_config.go
```diff
@@ -56,5 +57,16 @@ func (r *ProtocolDevLXD) GetConfigByKey(key string) (string, error) {
 		return "", err
 	}
 
+	if r.isDevLXDOverVsock {
+		// The returned string value is JSON encoded.
```
tomponline, line 63: So just to be 100% clear, before your recent changes, LXD was serving this value as a JSON encoded string right?

So this is just a regression in the devlxd-client/lxd-agent and not the LXD server?

MusicDin: Correct.

For example, this is how LXD-Agent handled the devLXD response for `/1.0/meta-data` before using devLXD client:
https://github.com/canonical/lxd/blob/299dcbd5ba48df3450b83069de7dd3804eb94691/lxd-agent/devlxd.go#L223-L233

Since resp.Metadata is of type `json.RawMessage` which is effectively `[]byte`, we could just do `string(resp.Metadata)` if it was not encoded.

tomponline:
@MusicDin please may you add some tests for this into lxd-ci?

tomponline (approved):
Thanks!

MusicDin:
Seems like SnapStore is having some issues, can't get tests to pass:
```
+ sudo snap install microovn --channel latest/edge
error: cannot perform the following tasks:
- Download snap "core24" (988) from channel "stable" (Get "https://canonical-bos01.cdn.snapcraftcontent.com/download-origin/canonical-lgw01/dwTAh7MZZ01zyriOZErqd1JynQLiOGvM_988.snap?interactive=1&token=1749654000_bf054585e8762891240b3c126e576c38cd08e1dc": read tcp 10.1.0.73:45012->91.189.91.99:443: read: connection reset by peer)
```

client/devlxd_config.go
```diff
@@ -56,5 +57,15 @@ func (r *ProtocolDevLXD) GetConfigByKey(key string) (string, error) {
 		return "", err
 	}
 
-	return string(resp.Content), nil
+	value := string(resp.Content)
```
simondeziel: This type conversion happens at run time so could it be done only `if !r.isDevLXDOverVsock`? With an optional early return, maybe?

client/devlxd_metadata.go
```diff
@@ -11,5 +12,15 @@ func (r *ProtocolDevLXD) GetMetadata() (metadata string, err error) {
 		return "", err
 	}
 
-	return string(resp.Content), nil
+	metadata = string(resp.Content)
```
simondeziel: Same thing about the run time conversion.

tomponline:
@MusicDin happy to merge this?

simondeziel:
I can report that fixed the `cloud-init` issue affecting VMs. Works flawlessly with:

```
$ snap list lxd
Name  Version      Rev    Tracking     Publisher   Notes
lxd   git-86df71d  34179  latest/edge  canonical✓  in-cohort,held
```

PR #15776, opened 2025-06-11
Backup: Rearrange consts to allow narrowing down volume lookup

roosterfish:
Small PR to move unrelated commits from https://github.com/canonical/lxd/pull/15523.

This moves some backup related constants from the main `backup` into the backup's dedicated `config` package which allows reusing them for a more narrow volume lookup (and to prevent an import cycle while doing so).

In addition when trying to retrieve the root vol from the list of vols in the backup config, the volume's type is now checked for being non-custom. 
This is in preparation for adding custom storage vols to the list of volumes in the backup config file.

### DIFF
--- a/lxd/backup.go
+++ b/lxd/backup.go
@@ -13,6 +13,7 @@ import (
 	"gopkg.in/yaml.v2"
 
 	"github.com/canonical/lxd/lxd/backup"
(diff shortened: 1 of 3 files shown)


tomponline (approved):
Ta

PR #15777, opened 2025-06-11
Storage: Add reverter pattern for custom volume backend actions

roosterfish:
Small PR to move unrelated bits from https://github.com/canonical/lxd/pull/15523.

Even though in some cases the "to be reverted" action is the last one in the row, having the reverter pattern in place allows adding additional instructions in the future without having to care about reverting existing actions in case of error.

This is starting to get useful with https://github.com/canonical/lxd/pull/15523 when reverting the instance's backup config file in case of errors.


tomponline (approved):
Thanks!

PR #15779, opened 2025-06-11
Makefile: Add `make update-fmt`

kadinsayani:
Add a `make update-fmt` target that runs `gofmt` on all files and update contribution guidelines.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -418,3 +418,7 @@ update-auth:
 			git commit -S -sm "lxd/auth: Update auth" -- ./lxd/auth/;\
 		fi;\
 	fi
+
+.PHONY: update-fmt
+update-fmt:
+	gofmt -w -s ./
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -99,6 +99,10 @@ endif
 
 	@echo "LXD-MIGRATE built successfully"
 
+.PHONY: fmt
+fmt:
+	gofmt -w -s ./
```
simondeziel: It is super fast:

```
$ time gofmt -w -s ./

real	0m0.290s
user	0m2.222s
sys	0m0.194s
```

but do we want to be altering the code in a non `update-*` make target?

kadinsayani: Most editors with the `gopls` LSP run `gofmt` or `goimports` on save anyways, I thought this would be helpful for new contributors (see https://github.com/canonical/lxd/pull/15745#pullrequestreview-2913210835).

In the future, we can also consider extending it to run `goimports` and `golangci-lint` on changed files, although we would need some fancy scripting to deal with `"github.com/canonical/lxd/client"` formatting to `lxd "github.com/canonical/lxd/client"`.

kadinsayani: We can add `make update-fmt` instead.

simondeziel: Yeah, I like that idea, being able to run `golangci-lint` in `--fix` mode would be SO nice.

tomponline (approved):

PR #15780, opened 2025-06-11
Install storage driver tools on-demand

simondeziel:
This allows speeding up both CI and local dev envs using `test-shell` by pulling just the needed bits, when they are needed.


simondeziel:
>     * This key is duplicated (also defined on line 17); remove the redundant entry to avoid confusion.
> 
> 
> ```
> APT::Get::Show-Versions "true";
> ```

This was actually a valid suggestion, thanks @Copilot.

doc/lxd-test.yaml
```diff
@@ -23,6 +23,10 @@ config:
     - content: "force-unsafe-io\n"
       path: /etc/dpkg/dpkg.cfg
       append: true
+    # Smaller initramfs (tailored to detected HW) and faster to build
```
tomponline: How does it know which modules are needed?

simondeziel: The `update-initramfs` tool has the smart to include the modules that are needed by the env it executes in.

However, you made me look again and the **24.04** minimal images are now booting without the usual `initrd` so this is dead code (I just dropped it). Thanks for catching this!

Server/physical machines running Noble (or VMs < 24.04) can still benefit from this trick as they keep using the traditional `initrd`. Here it is in action, with the various level of setting:

# Regular, big == boots everywhere
```
root@mars:~# time update-initramfs -uk all
update-initramfs: Generating /boot/initrd.img-6.8.0-60-generic
...
real	0m8.063s
user	0m2.678s
sys	0m3.009s
root@mars:~# ll -h /boot/initrd.img-6.8.0-60-generic 
-rw-r--r-- 1 root root 62M Jun 12 09:09 /boot/initrd.img-6.8.0-60-generic
```

# Tailored, smaller == boots only on ~identical machine
```
root@mars:~# echo MODULES=dep > /etc/initramfs-tools/conf.d/lxd-test.conf
root@mars:~# time update-initramfs -uk all
...
real	0m5.417s
user	0m1.467s
sys	0m2.668s
root@mars:~# ll -h /boot/initrd.img-6.8.0-60-generic 
-rw-r--r-- 1 root root 37M Jun 12 09:10 /boot/initrd.img-6.8.0-60-generic
```

tomponline: Thanks for looking into it, interesting stuff, TIL.

simondeziel: FYI, the speed difference between default and `MODULES=dep` was much bigger but since [23.10](https://discourse.ubuntu.com/t/reduce-initramfs-size-and-speed-up-the-generation-in-ubuntu-23-10/38972), they changed it so that modules are already `zstd` compressed and building the `initrd` is now essentially just concatenating them as-is into a `cpio` archive. There is no re-compression like there used to be.

tomponline (approved):
Thanks!

PR #15787, opened 2025-06-12
devlxd: Separate UbutuPro type used in LXD and in devLXD

MusicDin:
As part of devLXD refactoring, we want devLXD response types to be prefixed with `DevLXD`.
This change only rename the UbuntuPro response types, but the actual responses from devLXD are intact.

~Since UbuntuPro types (without DevLXD prefix) are not used outside devLXD, I would also prefer moving them into `lxd/ubuntupro` package - to be used only internally.~

Let me know if we want to proceed with the renaming for consistency.

### DIFF
--- a/client/devlxd_ubuntu_pro.go
+++ b/client/devlxd_ubuntu_pro.go
@@ -7,8 +7,8 @@ import (
 )
 
 // GetUbuntuPro retrieves the guest's Ubuntu Pro settings.
-func (r *ProtocolDevLXD) GetUbuntuPro() (*api.UbuntuProSettings, error) {
-	var info api.UbuntuProSettings
+func (r *ProtocolDevLXD) GetUbuntuPro() (*api.DevLXDUbuntuProSettings, error) {
+	var info api.DevLXDUbuntuProSettings
 
 	_, err := r.queryStruct(http.MethodGet, "/ubuntu-pro", nil, "", &info)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


MusicDin:
@tomponline We discussed that in our 1:1, but would like to confirm we still want to proceed with the renaming. Also, @markylaing what is your opinion on that?

markylaing:
Good morning. Just getting up to speed on my emails after being away.

I agree in principle with the change. 

Do you think we can simply rename `UbuntuProSettings` to `DevLXDUbuntuProSettings` instead? I realise that this is in `shared/api` but this should only ever have been used over devlxd. It wasn't part of the client until now, and since it's a fairly recent feature I doubt we'd be breaking anything.

MusicDin:
> Do you think we can simply rename UbuntuProSettings to DevLXDUbuntuProSettings instead? I realise that this is in shared/api but this should only ever have been used over devlxd. It wasn't part of the client until now, and since it's a fairly recent feature I doubt we'd be breaking anything.

While I agree, I would still prefer having UbuntuPro types directly in `lxd/ubuntupro` package (instead of in `shared/api` as it is in current PR) and separately only for DevLXD - I think this way it remains clean if we ever use that outside the scope of devlxd or add some new fields that are only relevant for LXD and not devLXD

tomponline:
> > Do you think we can simply rename UbuntuProSettings to DevLXDUbuntuProSettings instead? I realise that this is in shared/api but this should only ever have been used over devlxd. It wasn't part of the client until now, and since it's a fairly recent feature I doubt we'd be breaking anything.
> 
> While I agree, I would still prefer having UbuntuPro types directly in `lxd/ubuntupro` package (instead of in `shared/api` as it is in current PR) and separately only for DevLXD - I think this way it remains clean if we ever use that outside the scope of devlxd or add some new fields that are only relevant for LXD and not devLXD

We have other functions in lxd that return api types though, its ok (even encouraged) to use api types internally.

Issue #15742, opened 2025-06-06
Poor write performance of LXD storage volume with cephfs driver on VM


mastier:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

24.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
$ snap list --all lxd core20 core22 core24 snapd
Name    Version         Rev    Tracking       Publisher   Notes
core22  20250408        1963   latest/stable  canonical✓  base,disabled
core22  20250425        1981   latest/stable  canonical✓  base
core24  20250504        988    latest/stable  canonical✓  base
lxd     5.21.3-c5ae129  33110  5.21/stable    canonical✓  in-cohort
snapd   2.68.4          24505  latest/stable  canonical✓  snapd
```

### Output of "lxc info" or system info if it fails

```shell
$ sudo lxc info                                                                                      
config:                                           
  cluster.https_address: 10.46.111.4:8443                                                                              
  core.https_address: '[::]:8443'                          
  core.proxy_ignore_hosts: 10.46.16.251,10.46.16.252,10.46.16.253,127.0.0.1,::1,localhost                              
  network.ovn.northbound_connection: ssl:10.46.111.4:6641,ssl:10.46.111.5:6641,ssl:10.46.111.6:6641                    
api_extensions:                                                                                                        
- storage_zfs_remove_snapshots      
- container_host_shutdown_timeout                                                                                                                                                                                                             
- container_stop_priority                                                                                                                                                                                                                     
- container_syscall_filtering                                                                                                                                                                                                                 
- auth_pki                           
- container_last_used_at                                                                                               
- etag                               
- patch                                
- usb_devices                           
- https_allowed_credentials        
- image_compression_algorithm                                                                                                                                                                                                       [448/1338]
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
- unix_device_rename
[...]

Issue #15248, opened 2025-03-24
VM: Invalid compat argument from virtiofsd for paths containing `=`


hamistao:
When using a `=` in a device path for a VM directory passthrough device, as in `lxc storage volume attach default vol vm /mnt/path=`, we get a stale operation for that VM, the disk device set up freezes midway and further operations on that VM become impossible and always halts. Upon looking the virtiofsd logs we can see something like: error: Invalid compat argument '-o source=/var/snap/lxd/common/lxd/devices/vm/disk.vol.path='.

The PR https://github.com/canonical/lxd/pull/14744 is currently closed due to inactivity but contains a detailed approach to fixing this.

ljluestc:
```

package apparmor

import (
	"crypto/sha256"
	"fmt"
	"os"
	"path/filepath"
	"strings"

	"github.com/lxc/lxd/shared"
)

// VirtiofsdProfile generates an AppArmor profile for virtiofsd.
func VirtiofsdProfile(instanceName, sourcePath string, readOnly bool) (string, error) {
	// Sanitize source path and create a unique profile name.
	sourcePath = filepath.Clean(sourcePath)
	hash := sha256.Sum256([]byte(instanceName + sourcePath))
	profileName := fmt.Sprintf("lxd-virtiofsd-%s-%x", instanceName, hash[:8])

	// Define the AppArmor profile content.
	profileContent := fmt.Sprintf(`#include <tunables/global>

profile %s flags=(attach_disconnected,mediate_deleted) {
  #include <abstractions/base>

  # Allow virtiofsd binary execution
  /snap/lxd/*/bin/virtiofsd mr,

  # Allow access to the source directory
  %s/** %s,

  # Basic runtime requirements
  /proc/** r,
  /sys/** r,
  /dev/** r,
  /run/** rw,
  /tmp/** rw,

  # Logging and communication
  /var/log/** rw,
  /var/snap/lxd/common/lxd/logs/** rw,

  # Deny dangerous operations
  deny /etc/** w,
  deny /root/** w,
  deny /sbin/** x,
  deny /usr/sbin/** x,
}
`, profileName, sourcePath, getAccessMode(readOnly))

	// Write the profile to /etc/apparmor.d/
	profilePath := filepath.Join("/etc/apparmor.d", profileName)
	err := os.WriteFile(profilePath, []byte(profileContent), 0644)
	if err != nil {
		return "", fmt.Errorf("failed to write AppArmor profile %s: %w", profilePath, err)
[...]

PR #15792, opened 2025-06-13
Device: Add thread pool option for VM filesystem `disk` shares using `virtiofsd`

tomponline:
- Switches to modern flags provided by Rust implementation of `virtiofsd`, in the process fixes https://github.com/canonical/lxd/issues/15248
- Enables direct I/O passthrough from guest when using virtiofs (for correctness when an application specifies to open a file with direct I/O rather than silently converting that to a non-direct operation).
- Adds `io.threads` option for filesystem `disk` devices when being passed to the VM as `virtiofsd` based shares - this can improve performance. Fixes https://github.com/canonical/lxd/issues/15742
- Restricts use of this new `io.threads` option in restricted projects unless `restricted.virtual-machines.lowlevel` is set to `allow` to prevent restricted VMs from setting a high thread pool count and consuming extra host resources.
- Adds `disk_io_threads_virtiofsd` API extension.

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2672,3 +2672,7 @@ This adds support for listing networks across all projects using the `all-projec
 ## `clustering_restore_skip_mode`
 
 Adds a `skip` mode to the restore request. This mode restores a cluster member's status to `ONLINE` without restarting any of its stopped local instances or migrating back instances that were evacuated to other cluster members.
+
+## `disk_io_threads_virtiofsd`
+
+Adds the {config:option}`device-disk-device-conf:io.threads` option on `disk` devices which is used to control the `virtiofsd` thread pool size when sharing file systems into VMs. This can help improve I/O performance.
--- a/lxd/device/device_utils_disk.go
+++ b/lxd/device/device_utils_disk.go
@@ -450,8 +450,7 @@ func DiskVMVirtiofsdStart(kernelVersion version.DottedVersion, inst instance.Ins
 	args := []string{
 		"--fd=3",
 		"--shared-dir", sharePath,
-		// use -o flags for support in wider versions of virtiofsd.
-		"-o", "xattr",
+		"--xattr",
 	}
 
 	// Virtiofsd defaults to namespace sandbox mode which requires pidfd_open support.
--- a/lxd/device/device_utils_disk.go
+++ b/lxd/device/device_utils_disk.go
@@ -451,6 +451,7 @@ func DiskVMVirtiofsdStart(kernelVersion version.DottedVersion, inst instance.Ins
 		"--fd=3",
 		"--shared-dir", sharePath,
 		"--cache", "auto", // "never" and "metadata" modes do not allow execution at this time.
+		"--allow-direct-io",
 		"--xattr",
 	}
 
--- a/lxd/device/device_utils_disk.go
+++ b/lxd/device/device_utils_disk.go
@@ -449,9 +449,9 @@ func DiskVMVirtiofsdStart(kernelVersion version.DottedVersion, inst instance.Ins
 	// Start the virtiofsd process in non-daemon mode.
 	args := []string{
 		"--fd=3",
+		"--shared-dir", sharePath,
 		// use -o flags for support in wider versions of virtiofsd.
 		"-o", "xattr",
-		"-o", "source=" + sharePath,
 	}
 
 	// Virtiofsd defaults to namespace sandbox mode which requires pidfd_open support.
--- a/lxd/device/device_utils_disk.go
+++ b/lxd/device/device_utils_disk.go
@@ -450,6 +450,7 @@ func DiskVMVirtiofsdStart(kernelVersion version.DottedVersion, inst instance.Ins
 	args := []string{
 		"--fd=3",
 		"--shared-dir", sharePath,
+		"--cache", "auto", // "never" and "metadata" modes do not allow execution at this time.
 		"--xattr",
 	}
 
--- a/lxd/device/device_utils_disk.go
+++ b/lxd/device/device_utils_disk.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"slices"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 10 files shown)


lxd/device/device_utils_disk.go
```diff
@@ -449,9 +449,9 @@ func DiskVMVirtiofsdStart(kernelVersion version.DottedVersion, inst instance.Ins
 	// Start the virtiofsd process in non-daemon mode.
 	args := []string{
 		"--fd=3",
+		"--shared-dir", sharePath,
```
simondeziel, line 453: The (C?) version bundled in Jammy's QEMU doesn't mention this `--shared-dir` option in the man page: https://manpages.ubuntu.com/manpages/jammy/en/man1/virtiofsd.1.html The same page also says that in this old version they use a thread pool of 64 :)

This might cause problems to non-snap users.

lxd/device/device_utils_disk.go
```diff
@@ -450,8 +450,7 @@ func DiskVMVirtiofsdStart(kernelVersion version.DottedVersion, inst instance.Ins
 	args := []string{
 		"--fd=3",
 		"--shared-dir", sharePath,
-		// use -o flags for support in wider versions of virtiofsd.
-		"-o", "xattr",
```
simondeziel, line 453: Same concern with the old version bundled with Jammy's QEMU.

lxd/device/device_utils_disk.go
```diff
@@ -450,6 +450,7 @@ func DiskVMVirtiofsdStart(kernelVersion version.DottedVersion, inst instance.Ins
 	args := []string{
 		"--fd=3",
 		"--shared-dir", sharePath,
+		"--cache", "auto", // "never" and "metadata" modes do not allow execution at this time.
```
simondeziel, line 454: This one at least is supported by Jammy's QEMU version. They show it as having an `=` sign as in: `--cache=auto` which may or may not be relevant.

lxd/device/device_utils_disk.go
```diff
@@ -451,6 +451,7 @@ func DiskVMVirtiofsdStart(kernelVersion version.DottedVersion, inst instance.Ins
 		"--fd=3",
 		"--shared-dir", sharePath,
 		"--cache", "auto", // "never" and "metadata" modes do not allow execution at this time.
+		"--allow-direct-io",
```
simondeziel, line 455: Same issue with the old version bundled in Jammy's QEMU not advertising that flag in the man page.

lxd/device/device_utils_disk.go
```diff
@@ -452,6 +453,7 @@ func DiskVMVirtiofsdStart(kernelVersion version.DottedVersion, inst instance.Ins
 		"--shared-dir", sharePath,
 		"--cache", "auto", // "never" and "metadata" modes do not allow execution at this time.
 		"--allow-direct-io",
+		"--thread-pool-size", strconv.FormatUint(uint64(threadPoolSize), 10),
```
simondeziel, line 456: This one is supported by Jammy with the `=` in between, dunno if relevant as arg parsing is so unpredictable.

simondeziel (dismissed):
LGTM with the caveat that I think it breaks compat with the bundled version of virtiofsd in Jammy's QEMU.

tomponline:
> LGTM with the caveat that I think it breaks compat with the bundled version of virtiofsd in Jammy's QEMU.

Indeed, unfortunately only the rust version has the direct io option so think we will be forced to support only that. We wont backport anyhow to 5.0.

lxd/device/disk.go
```diff
@@ -404,6 +404,15 @@ func (d *disk) validateConfig(instConf instance.ConfigReader) error {
 		//  condition: virtual machine
 		//  shortdesc: Bus for the device
 		"io.bus": validate.Optional(validate.IsOneOf("nvme", "virtio-blk", "virtio-scsi")),
+		// lxdmeta:generate(entities=device-disk; group=device-conf; key=io.threads)
+		// This option controls the `virtiofsd` thread pool size. Only applies to virtiofs filesystem shares.
```
minaelee:
```suggestion
		// This option controls the `virtiofsd` thread pool size. Only applies to virtiofs file system shares.
```
```suggestion
		// This option controls the `virtiofsd` thread pool size. Only applies to virtiofs filesystem shares.
```
```suggestion
		// This option controls the `virtiofsd` thread pool size. Only applies to virtiofs filesystem shares.
```

lxd/device/disk.go
```diff
@@ -404,6 +404,15 @@ func (d *disk) validateConfig(instConf instance.ConfigReader) error {
...
+		// ---
+		//  type: integer
+		//  defaultdesc: `0`
+		//  required: no
+		//  condition: virtual machine
+		//  shortdesc: Thread pool for virtiofs filesystem shares
```
minaelee:
```suggestion
		//  shortdesc: Thread pool for virtiofs file system shares
```
```suggestion
		//  shortdesc: Thread pool for virtiofs file system shares
```
```suggestion
		//  shortdesc: Thread pool for virtiofs file system shares
```

doc/api-extensions.md
```diff
@@ -2672,3 +2672,7 @@ This adds support for listing networks across all projects using the `all-projec
 ## `clustering_restore_skip_mode`
 
 Adds a `skip` mode to the restore request. This mode restores a cluster member's status to `ONLINE` without restarting any of its stopped local instances or migrating back instances that were evacuated to other cluster members.
+
+## `disk_io_threads_virtiofsd`
+
+Adds the {config:option}`device-disk-device-conf:io.threads` option on `disk` devices which is used to specify the `virtiofsd` thread pool size when sharing filesystems into VMs.
```
```suggestion
Adds the {config:option}`device-disk-device-conf:io.threads` option on `disk` devices, which is used to specify the `virtiofsd` thread pool size when sharing file systems into VMs.
```
```suggestion
Adds the {config:option}`device-disk-device-conf:io.threads` option on `disk` devices which is used to specify the `virtiofsd` thread pool size when sharing filesystems into VMs.
```
```suggestion
Adds the {config:option}`device-disk-device-conf:io.threads` option on `disk` devices which is used to specify the `virtiofsd` thread pool size when sharing filesystems into VMs.
```

minaelee: Alternatively, add `filesystem` and `filesystems` to `doc/.custom_wordlist.txt` to pass the spellcheck test.

skozina (dismissed):
LGTM, thanks!

doc/api-extensions.md
tomponline: FWIW filesystem is already in the custom word list, but i went ahead and split them up.

simondeziel (approved):

Issue #15358, opened 2025-04-09
Failure in copying an instance with custom volume attached


mseralessandri:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

Ubuntu 24.04.2 LTS

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
LXD 6.3 
Name    Version      Rev    Tracking       Publisher   Notes
core22  20250210     1802   latest/stable  canonical✓  base
core24  20241217     739    latest/stable  canonical✓  base
lxd     git-482d520  33213  latest/edge    canonical✓  disabled
lxd     git-ce40f2f  33224  latest/edge    canonical✓  -
snapd   2.67.1       23771  latest/stable  canonical✓  snapd

LXD 5.21.3 
snap list --all lxd core20 core22 core24 snapd
Name    Version         Rev    Tracking       Publisher   Notes
core22  20250210        1802   latest/stable  canonical✓  base
lxd     5.21.3-c5ae129  33110  5.21/stable    canonical✓  -
snapd   2.67.1          23771  latest/stable  canonical✓  snapd
```

### Output of "lxc info" or system info if it fails

```shell
LXD 6.3
lxc info
config:
  core.https_address: '[::]:8443'
api_extensions:
- storage_zfs_remove_snapshots
- container_host_shutdown_timeout
- container_stop_priority
- container_syscall_filtering
- auth_pki
- container_last_used_at
- etag
- patch
- usb_devices
- https_allowed_credentials
- image_compression_algorithm
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
- unix_device_rename
- storage_lvm_use_thinpool
- storage_rsync_bwlimit
- network_vxlan_interface
- storage_btrfs_mount_options
- entity_description
- image_force_refresh
- storage_lvm_lv_resizing
- id_map_base
- file_symlinks
- container_push_target
[...]

PR #15798, opened 2025-06-13
lxd/instance/qemu: Fix shared storage volume setup during instance migration (from Incus)

nmezhenskyi:
Fixes https://github.com/canonical/lxd/issues/15358.

Includes cherry-pick from https://github.com/lxc/incus/pull/1669.   

Previously `MigrateReceive()` function would incorrectly try calling `CreateVolumeFromMigration()` for attached local volumes during instance migration and fail with "not supported" error. This change fixes the issue. Now, `CreateVolumeFromMigration()` is called only for shared storage volumes in order to notify the shared storage driver to setup the volume on the receiving cluster member.

### DIFF
--- a/lxd/instance/drivers/driver_qemu.go
+++ b/lxd/instance/drivers/driver_qemu.go
@@ -6927,8 +6927,9 @@ func (d *qemu) migrateSendLive(pool storagePools.Pool, clusterMoveSourceName str
 			diskPools[poolName] = diskPool
 		}
 
-		// Check that we're on shared storage.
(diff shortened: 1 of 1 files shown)


nmezhenskyi:
It seems that containers can be copied correctly even with custom volumes attached if the destination server already has volumes with the same name.

tomponline:
@nmezhenskyi hey, please can u book a meeting for us to work through the problem together as the scope of this bug fix remains unclear. Ta

tomponline (approved):
LGTM thanks!

PR #15799, opened 2025-06-13
test/suites/concurrent: ignore PIDs that vanished

simondeziel:


tomponline (approved):
Ta

simondeziel:
It seems there is a genuine issue as even with vanished PIDs now ignored, we still get failures: https://github.com/simondeziel/lxd/actions/runs/15669971643/job/44139636130#step:15:48490

I've only seen this affect `btrfs` so maybe it's because it fails one of the `lxc delete`?

PR #15800, opened 2025-06-13
Tiny speed and precision tweak to a few tests

simondeziel:
`lxc list -c n` is a fair bit cheaper than a regular `lxc list` which fetches much more than instance names. In a few places where only a subset of the information listing is needed, use the cheaper variant.


tomponline (approved):
Ta

PR #15818, opened 2025-06-16
Run CI on `tmpfs` mount

simondeziel:


tomponline (commented):
Interested to see speed improvement if any on this PR.

simondeziel:
> Interested to see speed improvement if any on this PR.

Locally it changed very little. I think the bulk of the saving will come from not reclaiming disk space (which is not directly related anyway).

PR #15820, opened 2025-06-17
Test tweaks

simondeziel:


test/main.sh
```diff
@@ -228,11 +228,11 @@ if [ -n "${INACCESSIBLE_DIRS:-}" ]; then
 fi
 
 if [ "${LXD_TMPFS:-0}" = "1" ]; then
-  mount -t tmpfs tmpfs "${TEST_DIR}" -o mode=0751 -o size=7G
+  mount -t tmpfs tmpfs "${TEST_DIR}" -o mode=0751,noatime -o size=7G
```
tomponline, line 234: I think we should stick use relatime as it gives similar performance benefits than noatime but still updates atime for written files.

tomponline (commented):
LGTM, apart from that noatime change, i'd prefer we use relatime unless there's a specific measurable reason to not use it.

simondeziel:
> LGTM, apart from that noatime change, i'd prefer we use relatime unless there's a specific measurable reason to not use it.

This only changes the "in-memory" time accounting as `relatime` only reaches the disk once per 24-hours which shouldn't be relevant for CI.

Since this has a non-measurable impact I'm perfectly fine to drop it but could it be in a follow up PR, to avoid the delay?

tomponline:
> > LGTM, apart from that noatime change, i'd prefer we use relatime unless there's a specific measurable reason to not use it.
> 
> This only changes the "in-memory" time accounting as `relatime` only reaches the disk once per 24-hours which shouldn't be relevant for CI.
> 
> Since this has a non-measurable impact I'm perfectly fine to drop it but could it be in a follow up PR, to avoid the delay?

Yes please

tomponline (approved):

PR #15825, opened 2025-06-17
[WIP] Add basic VM tests

simondeziel:
This works on 24.04 local `test-shell` machines but still needs QA on `22.04` ones.


test/suites/vm.sh
tomponline, line 1: > On 22.04, it just hangs:

Yes this is an annoying side effect of the way we start virtiofsd (and then it fails with invalid flags) but then we pass the FD to the created socket file to qemu anyway.

We should check the virtiofsd is available at the end of the unit socket file before passing it to qemu, but this might cause it to exit when we disconnect, in which case we should test that the virtiofsd process is still running after some point in time?

test/suites/vm.sh
```diff
@@ -16,4 +21,8 @@ test_vm_empty() {
   lxc launch --vm --empty --ephemeral v1 -c limits.memory=128MiB -d "${SMALL_ROOT_DISK}"
   lxc stop -f v1
   [ "$(lxc list -f csv -c n)" = "" ]
+
+  if grep -qxF 'VERSION_ID="22.04"' /etc/os-release; then
```
tomponline, line 30: any reason to not leave this set?

.github/workflows/tests.yml
```diff
@@ -774,7 +774,11 @@ jobs:
           sudo chown $USER /var/lib/lxd/unix.socket
           export PATH="/home/runner/go/bin:$PATH"
           export CI=true
-          export DISABLE_VM_TESTS=true
+
+          # On 22.04, VMs do not work due to `virtiofsd` incompat.
+          if grep -qxF 'VERSION_ID="22.04"' /etc/os-release; then
```
tomponline: do we know if this is going to work or is it likely to break in GH runners?

tomponline:
@simondeziel thanks for this, we can test quite a lot of VM functionality and logic without needing to actually boot the guest fully (like storage operations for example) :)

tomponline:
> also running into Apparmor issues on local VMs. Fortunately, the `vm_empty` test does work in CI despite running on 22.04 machines.

Lets hold off merging this until we can resolve these.

But if you want to break out some of your non-VM related commits from this PR then these are good to go in now.

test/suites/vm.sh
simondeziel: When looking at it after it hung, IIRC there was a `qemu-system` process marked as 'defunct'.

test/suites/vm.sh
simondeziel: It breaks any container creation afterward. Presumably due to not having `criu` working and thus `migration.stateful` not working for them?

tomponline: we dont run criu tests afaik

tomponline: what is the error?

.github/workflows/tests.yml
simondeziel: I don't understand your question. It does work in the sense the 22.04 runners are able to go through the `vm_empty` test.

test/suites/vm.sh
tomponline: Feel free to open a GH bug about this.

.github/workflows/tests.yml
tomponline: but this is for the lxd-ui tests, so do we know what their test suite does for VMs?

simondeziel:
> > also running into Apparmor issues on local VMs. Fortunately, the `vm_empty` test does work in CI despite running on 22.04 machines.
> 
> Lets hold off merging this until we can resolve these.
> 
> But if you want to break out some of your non-VM related commits from this PR then these are good to go in now.

Since the 22.04 runners are not hitting the Apparmor issues I wouldn't mind trying to get the local `test-shell` version working at least to a point where I can either fix the Apparmor profile generation or fill a bug with a reproducer.

.github/workflows/tests.yml
simondeziel: Oh sorry, I didn't read the context so I missed this was for the UI. I have not consulted with them and this has not been tested as this needs to be merged into `main` to have the needed bits for auth.

Should I drop that bit and ask them to test themselves?

tomponline: yeah lets do that before we break the daily tests

tomponline:
> > > also running into Apparmor issues on local VMs. Fortunately, the `vm_empty` test does work in CI despite running on 22.04 machines.
> > 
> > 
> > Lets hold off merging this until we can resolve these.
> > But if you want to break out some of your non-VM related commits from this PR then these are good to go in now.
> 
> Since the 22.04 runners are not hitting the Apparmor issues I wouldn't mind trying to get the local `test-shell` version working at least to a point where I can either fix the Apparmor profile generation or fill a bug with a reproducer.

Can you not do this on a PR, why does it need merging first?

simondeziel:
The

> > > > also running into Apparmor issues on local VMs. Fortunately, the `vm_empty` test does work in CI despite running on 22.04 machines.
> > > 
> > > 
> > > Lets hold off merging this until we can resolve these.
> > > But if you want to break out some of your non-VM related commits from this PR then these are good to go in now.
> > 
> > 
> > Since the 22.04 runners are not hitting the Apparmor issues I wouldn't mind trying to get the local `test-shell` version working at least to a point where I can either fix the Apparmor profile generation or fill a bug with a reproducer.
> 
> Can you not do this on a PR, why does it need merging first?

Since the Apparmor issue I run into only manifests for local VM created with `lxc launch ubuntu-minimal-daily:22.04 v2 --vm -p lxd-test`, I don't think we should hold off on testing VMs in CI.

In fact, I started abusing the `lxd-test` profile with 22.04 VMs just because we are stucked with this old version on GH. The `lxd-test` profile was never meant to be used on anything older than 24.04 :/

simondeziel:
Here is the Apparmor problem I run into with a local 22.04 machine:
```
root@v2:~/lxd/test# LXD_VM_TEST=1 LXD_VERBOSE=1 LXD_INSPECT=1 ./main.sh vm_empty
+ test_vm_empty
+ '[' 1 = 0 ']'
+ grep -qxF 'VERSION_ID="22.04"' /etc/os-release
+ echo 'Using migration.stateful to force 9p config drive thus avoiding the old/incompatible virtiofsd'
Using migration.stateful to force 9p config drive thus avoiding the old/incompatible virtiofsd
+ lxc profile set default migration.stateful=true
++ timeout --foreground 120 /root/go/bin/lxc profile set default migration.stateful=true --verbose
+ lxc init --vm --empty v1 -c limits.memory=128MiB -d root,size=32MiB
++ timeout --foreground 120 /root/go/bin/lxc init --vm --empty v1 -c limits.memory=128MiB -d root,size=32MiB --verbose
Creating v1
INFO   [2025-06-19T13:49:48Z] Creating instance                             ephemeral=false instance=v1 instanceType=virtual-machine project=default
INFO   [2025-06-19T13:49:48Z] Created instance                              ephemeral=false instance=v1 instanceType=virtual-machine project=default
+ lxc start v1
++ timeout --foreground 120 /root/go/bin/lxc start v1 --verbose
WARNING[2025-06-19T13:49:49Z] Unable to use virtio-fs for config drive, using 9p as a fallback  err="Stateful migration unsupported" instance=v1 instanceType=virtual-machine project=default
WARNING[2025-06-19T13:49:49Z] Using writeback cache I/O                     devPath=/tmp/lxd-test.tmp.csVG/1ga/storage-pools/lxdtest-1ga/virtual-machines/v1/root.img device=root fsType=tmpfs instance=v1 instanceType=virtual-machine project=default
Error: Failed setting up device via monitor: Failed opening file descriptor for disk device "root": open /tmp/lxd-test.tmp.csVG/1ga/storage-pools/lxdtest-1ga/virtual-machines/v1/root.img: invalid argument
Try `lxc info --show-log v1` for more info
+ cleanup
==> FAILED TEST: vm_empty (vm_empty)
==> Test result: failure
Dropping to a shell for inspection.
Once done, exit (Ctrl-D) to continue
LXD-TEST root@v2:~/lxd/test$ dmesg 
[ 3000.569868] audit: type=1400 audit(1750340989.101:323): apparmor="STATUS" operation="profile_load" profile="unconfined" name="lxd-v1_</tmp/lxd-test.tmp.csVG/1ga>" pid=76047 comm="apparmor_parser"
[ 3000.607670] audit: type=1400 audit(1750340989.137:324): apparmor="DENIED" operation="open" profile="lxd-v1_</tmp/lxd-test.tmp.csVG/1ga>" name="/root/lxd/test/" pid=76048 comm="qemu-system-x86" requested_mask="r" denied_mask="r" fsuid=0 ouid=1000
[ 3000.665248] IPv6: ADDRCONF(NETDEV_CHANGE): tapc758fe55: link becomes ready
```

test/suites/vm.sh
simondeziel: https://github.com/canonical/lxd/actions/runs/15759914517/job/44424421810?pr=15825#step:15:31127:

```
++ lxc config device get nt12307 eth0 name
+++ timeout --foreground 120 /home/runner/go/bin/lxc config device get nt12307 eth0 name --verbose
+ '[' eth0 = eth0 ']'
++ lxc config device get nt12307 eth0 network
+++ timeout --foreground 120 /home/runner/go/bin/lxc config device get nt12307 eth0 network --verbose
+ '[' lxdt12307 = lxdt12307 ']'
+ lxc config show -e nt12307
+ sed '/^devices:/,/type:/ s/eth0/eth1/'
+ lxc config edit nt12307
++ timeout --foreground 120 /home/runner/go/bin/lxc config show -e nt12307 --verbose
++ timeout --foreground 120 /home/runner/go/bin/lxc config edit nt12307 --verbose
Error: Invalid config: "migration.stateful" isn't supported for "container"
+ cleanup
```

But that's in `container_devices_nic_bridged`, way after the `vm_empty` test.

simondeziel: Ah, that's because we take the container's expanded config and alter it:

```
lxc config show -e "${ctName}" | sed '/^devices:/,/type:/ s/eth0/eth1/' | lxc config edit "${ctName}" # Rename added "eth0" to "eth1"
```

simondeziel:
> Here is the Apparmor problem I run into with a local 22.04 machine:

OK so after looking into this with @mihalicyn the conclusion is that Appamor is a red herring and the real problem is when using `tmpfs` with an old kernel (22.04's 5.15 aka GA) where direct-io is not supported.

tomponline:
> > Here is the Apparmor problem I run into with a local 22.04 machine:
> 
> OK so after looking into this with @mihalicyn the conclusion is that Appamor is a red herring and the real problem is when using `tmpfs` with an old kernel (22.04's 5.15 aka GA) where direct-io is not supported.

Well done chaps thanks

simondeziel:
Moving back to draft as I still need to deactivate some bits that are not compatible with 22.04's 5.15 kernel, sometimes when combined with the now default `LXD_TMPFS=1`.

tomponline (approved):
LGTM thanks!

tomponline:
@simondeziel the skips for virtiofs are only relevant for running inside a test shell VM right?

simondeziel:
> @simondeziel the skips for virtiofs are only relevant for running inside a test shell VM right?

Indeed, this is one of the many differences between "local runners" and GH runners.

tomponline:
@simondeziel id like us to use the Canonical runners eventually to get a more representative test env, especially for our LTS series. Thanks

simondeziel:
I do agree that would give us more flexibility and a environment that better represent what "out there".

PR #15827, opened 2025-06-18
WIP per-storage backups and images storage

skozina:


doc/metadata.txt
```diff
@@ -4458,6 +4458,20 @@ Possible values are `bzip2`, `gzip`, `lzma`, `xz`, or `none`.
 Specify the number of days after which the unused cached image expires.
 ```
 
+```{config:option} storage.backups_volume project-specific
+:defaultdesc: "`block`"
```
simondeziel: Isn't the default value an empty string?

test/suites/projects.sh
```diff
@@ -1172,3 +1173,119 @@ test_projects_before_init() {
 
   shutdown_lxd "${LXD_INIT_DIR}"
 }
+
+test_projects_images_volume() {
+  if [ "${LXD_BACKEND}" = "ceph" ]; then
+    # This won't work on ceph because it's not a multi-node storage
+    ! lxc project create foo -c storage.images_volume="${pool}/vol" | grep -F "Error: Invalid project configuration key \"storage.images_volume\" value: Remote storage pool \"${pool}\" cannot be used"
```
simondeziel: I'm not sure why your version worked. Isn't the error supposed to be on `stderr`?

```suggestion
    ERR_OUT="$(! lxc project create foo -c storage.images_volume="${pool}/vol" 2>&1)"
    echo "${ERR_OUT}" | grep -F "Error: Invalid project configuration key \"storage.images_volume\" value: Remote storage pool \"${pool}\" cannot be used"
```

test/suites/projects.sh
```diff
@@ -1172,3 +1173,119 @@ test_projects_before_init() {
...
+  ! lxc project create foo -c features.images=false -c storage.images_volume="${pool}/vol" || false
+  lxc project create foo -c storage.images_volume="${pool}/vol"
+
+  # It should be possible to change the setting on empty projects
+  lxc project unset foo storage.images_volume
+  lxc project set foo storage.images_volume="${pool}/vol"
+
+  # Import an image into the project and grab its fingerprint
+  deps/import-busybox --project foo --alias testimage
+  fingerprint="$(lxc image list --project foo -c f --format json | jq .[0].fingerprint | sed -e 's/"//g')"
```
simondeziel:
```suggestion
  fingerprint="$(lxc image list --project foo -f csv -c F testimage)"
```

test/suites/projects.sh
```diff
@@ -1172,3 +1173,119 @@ test_projects_before_init() {
...
+
+  # The image should not exist in the default storage, only in the project images volume
+  [ ! -e "${LXD_DIR}/images/${fingerprint}" ]
+  [ -z "$(ls -A "${LXD_DIR}/images")" ]
+  [ -f "${LXD_DIR}/storage-pools/${pool}/custom/default_vol/images/${fingerprint}" ]
+
+  # It should not be possible to change the setting on empty projects
+  ! lxc project unset foo storage.images_volume
```
simondeziel:
```suggestion
  ! lxc project unset foo storage.images_volume || false
```

test/suites/projects.sh
```diff
@@ -1172,3 +1173,119 @@ test_projects_before_init() {
...
+  # It should be possible to change the setting on empty projects
+  lxc project unset foo storage.backups_volume
+  lxc project set foo storage.backups_volume="${pool}/vol"
+
+  # It should not be possible to move daemon storage to the same volume
+  ! lxc config set storage.backups_volume="${pool}/vol" || false
+
+  # Import an image into the project and grab its fingerprint
+  deps/import-busybox --project foo --alias testimage
+  fingerprint="$(lxc image list --project foo -c f --format json | jq .[0].fingerprint | sed -e 's/"//g')"
```
simondeziel:
```suggestion
  fingerprint="$(lxc image list --project foo -f csv -c F testimage)"
```

simondeziel: This does not seem to be used in this function though.

test/suites/projects.sh
```diff
@@ -1172,3 +1173,119 @@ test_projects_before_init() {
...
+
+  # Ensure the backup is gone
+  [ -d "${LXD_DIR}/storage-pools/${pool}/custom/default_vol/backups/instances" ]
+  [ -z "$(ls -A "${LXD_DIR}/storage-pools/${pool}/custom/default_vol/backups/instances")" ]
+
+  # Clean up
+  lxc image delete --project foo testimage
+  lxc project delete foo
+  lxc storage volume delete "${pool}" vol
+}
```
simondeziel: Missing `\n` at EOF.

test/suites/remote.sh
```diff
@@ -228,45 +228,47 @@ test_remote_usage() {
...
   lxc_remote image alias create localhost:testimage "${sum}"
 
   # test remote publish
-  lxc_remote init testimage pub
-  lxc_remote publish pub lxd2: --alias bar --public a=b
+  lxc_remote init --quiet testimage pub
+  lxc_remote publish --quiet pub lxd2: --alias bar --public a=b
   lxc_remote image show lxd2:bar | grep -F "a: b"
   lxc_remote image show lxd2:bar | grep -xF "public: true"
+  fingerprint="$(lxc image list lxd2:bar -c f --format json | jq .[0].fingerprint | sed -e 's/"//g')"
```
simondeziel: In JSON output, `-c` doesn't do anything. I think we should reject it.

See the other comment for a simpler way to extract the fingerprint, unless the deletion could be done through the remote+name?

simondeziel (commented):
Looking quite good! Thanks

doc/metadata.txt
skozina: It is. Thanks! Removed.

test/suites/projects.sh
skozina: Heh;-) I think this worked because, if the `lxc project create` fails, the grep is not even executed.
I don't think we need to check for the specific error message, the important piece is just that the `lxc project create` is expected to fail. So I've dropped the grep completely.
Thanks for spotting this!

test/suites/projects.sh
skozina: Nice, thank you! Done.

test/suites/projects.sh
skozina: I've done this, though... Can you help me to understand the meaning here? The unary operator has higher precedence than the OR. If `lxc` fails, `false` is not even evaluated. If `lxc` succeeds, how is `$(! 0 || false)` better than simple `$(! 0)` ?

test/suites/projects.sh
skozina: Indeed. Removed, thanks!

test/suites/projects.sh
skozina: Thanks, added.

test/suites/remote.sh
skozina: Thanks, I've switched this to the csv format.
remote+name would not work. Here we `lxc init` from `lxd-public:` remote to `local:`:
https://github.com/canonical/lxd/blob/b102391a19820a33d035873528d5a8fa6e1c83cc/test/suites/remote.sh#L261
Because the image is brought from remote node through cluster synchronization, it has no alias on the `local:`. So we need to use fingerprint to delete it.

test/suites/projects.sh
```diff
@@ -1172,3 +1173,119 @@ test_projects_before_init() {
...
+
+  # Clean up
+  lxc project delete foo
+  lxc storage volume delete "${pool}" vol
+}
+
+test_projects_backups_volume() {
+  if [ "${LXD_BACKEND}" = "ceph" ]; then
+    # This won't work on ceph because it's not a multi-node storage
+    ! lxc project create foo -c storage.backups_volume="${pool}/vol"
```
simondeziel:
```suggestion
    ! lxc project create foo -c storage.backups_volume="${pool}/vol" || false
```

test/suites/projects.sh
```diff
@@ -1172,3 +1173,119 @@ test_projects_before_init() {
 
   shutdown_lxd "${LXD_INIT_DIR}"
 }
+
+test_projects_images_volume() {
+  if [ "${LXD_BACKEND}" = "ceph" ]; then
+    # This won't work on ceph because it's not a multi-node storage
+    ! lxc project create foo -c storage.images_volume="${pool}/vol"
```
simondeziel:
```suggestion
    ! lxc project create foo -c storage.images_volume="${pool}/vol" || false
```

test/suites/projects.sh
simondeziel: It is very confusing due to `bash` quirks!

When `set -e` is used, `!` creates a compound command which is special cased. `man 1 bash`:

```
 -e Exit immediately if a pipeline (which may consist of a single simple command), a list,
    or  a  compound  command (see  SHELL  GRAMMAR above), exits with a non-zero
    status.  The shell does not exit if the command that fails is part of the command list
    immediately following a while or until keyword, part of the test following  the  if  or elif
    reserved  words, part of any command executed in a && or || list except the command
    following the final && or ||, any command in a pipeline but the last, or if the command's
    return value is being inverted with !.  If a compound command other than a subshell
    returns a non-zero status because a command failed while -e was being ignored, the
    shell does not exit.  A trap on ERR, if set, is executed before the shell exits. This option
    applies to the shell environment and each subshell environment separately (see
    COMMAND EXECUTION ENVIRONMENT above), and may cause subshells to exit
    before executing all the commands in the subshell.
```

So when you expect a failure, `! command_should_fail` should work but if it unexpectedly succeeds, `set -e` won't stop the execution:

```
$ bash -ec '! false; echo "should run"'
should run
$ bash -ec '! true; echo "should not run"'
should not run
```

Now with the ` || false`:

```
$ bash -ec '! false || false; echo "should run"'
should run
$ bash -ec '! true || false; echo "will not run"'
$ echo $?
1
```

Now you may ask what about the `[ ! -e "should/not/exist" ]`? It's surprisingly different as the negation is inside the test `[]` and there, `set -e` will do the right thing so there is no need for ` || false`.
It would be different if the negation was outside like this: `! [ -e "should/not/exist" ]`. This would require a ` || false` too as we now have a compound command.

In the end, it would all be much clearer and safer to use the longer forms:

```
if command_that_is_expected_to_fail; then
    echo "ERROR: Command succeeded unexpectedly" >&2
    exit 1
fi

if [ -e "should/not/exist" ]; then
    echo "ERROR: File exists unexpectedly" >&2
    exit 1
fi
```

But it is longer :/

skozina: I see. Thank you, I wouldn't figure it out otherwise;-)
Now I'm wondering if we should add this reasoning to the `test/README.md` file? While I feel I should have known this, I'm pretty sure I'll forget (perhaps again;-/).

test/suites/projects.sh
skozina: Thank you! Fixed both, and I've also verified there are no more missing `|| false`s in these files.
That said, I've added a new commit, if you could review that too, that would be great;-) Thank you!

test/suites/projects.sh
skozina: Thanks, fixed.

test/suites/projects.sh
simondeziel: That's a good idea, I always need to research it myself ;)

simondeziel: Actually, since you've fixed a few other places, I will take your `README.md` suggestion and will also add a lint script to catch those. Thanks!

test/suites/remote.sh
```diff
@@ -381,15 +383,16 @@ test_remote_usage() {
...
   lxc_remote image delete lxd2:bar
-  lxc_remote init localhost:testimage lxd2:c1
+  lxc_remote init --quiet localhost:testimage lxd2:c1
   cached=$(lxc_remote image info "lxd2:${fingerprint}" | awk '/^Cached/ { print $2 }')
   # The `cached` field should be set to `yes` since the image was implicitly downloaded by the instance create operation
   [ "${cached}" = "yes" ]
   # There should be no alias for the image
-  ! lxc_remote image info "lxd2:${fingerprint}" | grep -F "Aliases:"
+  aliases=$(lxc_remote image ls "lxd2:${fingerprint}" -f csv -c l,F)
+  [ "${aliases}" = ",${fingerprint}" ]
```
simondeziel:
```suggestion
  [ "$(lxc_remote image ls "lxd2:${fingerprint}" -f csv -c l)" = "" ]
```

Nice catch, AFAIK, this test was broken as the `lxc image info` output always include `Aliases:` no matter if there is or isn't any aliases.

simondeziel (approved):

test/suites/remote.sh
skozina: Yes, exactly:-)

simondeziel: I ended up taking that one with me in https://github.com/canonical/lxd/pull/15891 as I added the check for missing ` || false` fallbacks and documenting the `! cmd_should_fail` quirk.

test/suites/projects.sh
simondeziel: Thanks for sparking that discussion, this lead to finding a real bug in our test suite that made a LXD bug invisible: https://github.com/canonical/lxd/issues/15894

skozina: Thank you for the proposed changes to the lint script and the test/README file, I think both will be super useful!

test/suites/remote.sh
skozina: Cool, thanks! I've dropped the last commit, which was only fixing the `suites/remote` and `suites/projects`, from this PR.

simondeziel (dismissed):

lxd/api_project.go
```diff
@@ -712,22 +712,33 @@ func projectChange(ctx context.Context, s *state.State, project *api.Project, re
...
 	// Record which features have been changed.
 	var featuresChanged []string
+	var storageConfig string
 	for _, configKeyChanged := range configChanged {
 		_, isFeature := cluster.ProjectFeatures[configKeyChanged]
 		if isFeature {
 			featuresChanged = append(featuresChanged, configKeyChanged)
 		}
+
+		fmt.Println("checking config ", configKeyChanged)
```
tomponline: left over debug logging?

lxd/api_project.go
```diff
@@ -712,22 +712,33 @@ func projectChange(ctx context.Context, s *state.State, project *api.Project, re
...
+			storageConfig = configKeyChanged
+		}
 	}
 
+	usedByLen := len(project.UsedBy)
+	projectInUse := usedByLen > 1 || (usedByLen == 1 && !strings.Contains(project.UsedBy[0], "/profiles/default"))
+
 	// Quick checks.
+	if projectInUse && storageConfig != "" {
+		return response.BadRequest(fmt.Errorf("Project config %q cannot be configured on non-empty projects", storageConfig))
```
tomponline: "configured" implies that it cannot be active on non-empty projects, how about "changed" instead?

lxd/api_project.go
```diff
@@ -1085,6 +1096,11 @@ func isEitherAllowOrBlockOrManaged(value string) error {
 	return validate.Optional(validate.IsOneOf("block", "allow", "managed"))(value)
 }
 
+func isValidStorage(s *state.State, storage string) (err error) {
```
tomponline: The "storage" here in the function name and argument of the same name seems rather vague, combined with no docblock on the function makes it quite hard to reason about what this function does without inspecting the call sites and its implementation.

Its a tricky one for naming though as "storage" is too vague, but its specific niche format of `<pool>/<volume>` so it cant even be called "storage volume".

What about `isValidProjectStorageVolume(s, daemonStorageVolume)` with a doc block that says

> Validates a daemon storage volume in the form "<pool_name>/<volume_name>"

?

lxd/daemon_storage.go
```diff
@@ -255,6 +255,17 @@ func daemonStorageValidate(s *state.State, target string) (validatedTarget strin
 	return pool.Name() + "/" + dbVol.Name, nil
 }
 
+func daemonStoragePath(config string, target string) string {
```
tomponline: Please can you add a docblock here explaining what config and target are?

lxd/db/cluster/projects.go
```diff
@@ -96,6 +96,27 @@ func (p *Project) ToAPI(ctx context.Context, tx *sql.Tx) (*api.Project, error) {
 	return apiProject, nil
 }
 
+// ProjectConfig is a helper to return value of an config option of a project.
```
tomponline: just a nit "a config option" not "an config option"

lxd/db/cluster/projects.go
```diff
@@ -96,6 +96,27 @@ func (p *Project) ToAPI(ctx context.Context, tx *sql.Tx) (*api.Project, error) {
...
+func ProjectConfig(ctx context.Context, tx *sql.Tx, project string, config string) (string, error) {
```
tomponline: When I see a very niche function like this I try and to consider if making it slightly more generic can allow us to use it in more places/or more scenarios in the future without incurring a lot of extra cost in terms of query time and allocations (especially as project config is often small).

In this case my immediate thought was: Can we have a function that returns a map[string]string of all the project's config, as it should be almost the same cost as getting just one config item, and would be useful in more scenarios.

I then went to look to see if we had any similar functions, and indeed we do, `GetProjectConfig`, although this takes a project ID, so would require an additional project name -> ID lookup, as in this case we probably don't have the ID to hand.

However as I was looking into this `GetProjectConfig` function, I noticed something sub-optimal, it uses `GetConfig()` which is one of those "generated" functions, and it seems to get all config for a specific entity type. So this is loading all project config across the entire cluster, and then only returning the rows that match the specified project ID in Go.

Then I looked at the callsites for `GetProjectConfig` and found 2 places:

- func ProjectHasImages(ctx context.Context, tx *sql.Tx, name string) (bool, error)
- func (p *Project) ToAPI(ctx context.Context, tx *sql.Tx) (*api.Project, error)

Both `ProjectHasImages` and `ToAPI` have the project name available. And interestingly `ProjectHasImages` is doing a project name -> project ID lookup internally.

So what I think we should do here is to lose this new function and re-work `GetProjectConfig` to accept a project name rather than ID. Then use the same approach you've used here, but instead retrieve all of a project's config into a `map[string]string` to return.

Then `ProjectHasImages` and `ToAPI` can be updated to use it and pass in the project name.
This will improve both of those functions; `ProjectHasImages` wont have to do the name -> ID lookup anymore as a separate query, and both functions will no longer load all config from all projects from the DB.

tomponline: I'm also quite concerned about `GetConfig` in general and I think we need to audit where its used to retrieve a single entity's config. 

In many places it seems to be used correctly to get the config for multiple entities of the same time to improve efficiency, however it does seem that the DB generator is using it to in its own generated functions to retrieve the config for a single entity and then filtering for that specific entity on the Go side.

For instance `GetInstanceConfig` is a prime example of a function we need to remove and replace with a manually written function that doesn't use `GetConfig`, as it looks like that is loading all instances config across the entire cluster to get a single instance's config!

Thankfully `GetInstanceConfig` isn't actually used.

But `GetProjectConfig`, `GetProfileConfig` and `GetInstanceSnapshotConfig` are for example and that needs to change.

cc @markylaing

lxd/db/cluster/projects.go
```diff
@@ -117,6 +117,42 @@ SELECT projects_config.value
 	return values[0], nil
 }
 
+// GetProjectsNameAndConfigValue is a helper to return names of all projects keyd by the config value.
```
tomponline: keyed instead of keyd

lxd/api_1.0.go
```diff
@@ -631,7 +632,32 @@ func doAPI10Update(d *Daemon, r *http.Request, req api.ServerPut, patch bool) re
 	var newNodeConfig *node.Config
 	oldNodeConfig := make(map[string]any)
 
-	err := s.DB.Node.Transaction(r.Context(), func(ctx context.Context, tx *db.NodeTx) error {
+	var projectsImagesStorage map[string]string
+	var projectsBackupsStorage map[string]string
+	var err error
+	if nodeValues["storage.backups_volume"] != nil {
+		err := s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
```
tomponline: can we avoid starting a transaction per query, and instead move these together into a single transaction?

lxd/daemon.go
```diff
@@ -729,12 +729,20 @@ func (d *Daemon) State() *state.State {
 		}, nil
 	}
 
-	s.ImagesStoragePath = func() string {
-		return daemonStoragePath(s.LocalConfig.StorageImagesVolume(), "images")
+	s.ImagesStoragePath = func(config string) string {
```
tomponline: If you can in the future please, separate the function signature change into an earlier separate commit before updating the call sites.

This helps with 2 things; firstly its easier to review because the function change comes first, rather than being presented with updated call site changes without seeing the actual function change yet, and having to scroll down to some mid-way point to try and find the change.

Secondly, it helps with backports, as the callsites may well be different between LTS and current feature release, so often the function signature change commit will backport OK, but the callsite one needs to be tweaked.

tomponline: can we call `config` the same term we use for `isValidStorage`, e.g. `daemonStorageVolume`

lxd/daemon_storage.go
tomponline: can we call `config` the same term we use for `isValidStorage`, e.g. `daemonStorageVolume`

lxd/images.go
```diff
@@ -1124,8 +1124,18 @@ func imagesPost(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	var projectImagesVolume string
+	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
```
tomponline: as no state modification has happened yet we can use r.Context() here safely rather than context.TODO()

tomponline: if you can separate the callsite updates to accommodate the new argument added to `s.ImagesStoragePath` and `s.BackupsStoragePath` functions from the changes that add the new project storage functionality that would help with reviews and backports.

lxd/daemon_storage.go
```diff
@@ -21,6 +21,27 @@ import (
 	"github.com/canonical/lxd/shared/api"
 )
 
+func unmountCustomVolume(s *state.State, source string) error {
```
tomponline: Can we call this `unmountDaemonStorageVolume()` and the `source` argument be renamed to `daemonStorageVolume` to clearly indicate this is specific to daemon storage volumes and not general custom volume unmounts.

lxd/daemon_storage.go
```diff
@@ -83,6 +83,27 @@ func daemonStorageVolumesUnmount(s *state.State, ctx context.Context) error {
 	return nil
 }
 
+func mountCustomVolume(s *state.State, source string) error {
```
tomponline: same feedback re func name and source arg as above.

lxd/api_project.go
```diff
@@ -319,6 +347,20 @@ func projectsPost(d *Daemon, r *http.Request) response.Response {
 		return response.BadRequest(err)
 	}
 
+	if project.Config["storage.images_volume"] != "" {
+		err = projectStorageSetup(s, project.Config["storage.images_volume"], "images")
+		if err != nil {
+			return response.BadRequest(err)
```
tomponline: please use response.SmartError here as otherwise this will always send a 400 HTTP response, even if the error indicates an internal server error or some other error.

lxd/api_project.go
```diff
@@ -319,6 +347,20 @@ func projectsPost(d *Daemon, r *http.Request) response.Response {
...
+		}
+	}
+
+	if project.Config["storage.backups_volume"] != "" {
+		err = projectStorageSetup(s, project.Config["storage.backups_volume"], "backups")
+		if err != nil {
+			return response.BadRequest(err)
```
tomponline: response.SmartError

lxd/api_project.go
```diff
@@ -952,13 +996,35 @@ func projectDelete(d *Daemon, r *http.Request) response.Response {
...
 	})
 
 	if err != nil {
 		return response.SmartError(err)
 	}
 
+	if imagesVolume != "" {
+		err := unmountCustomVolume(s, imagesVolume)
+		if err != nil {
+			return response.BadRequest(fmt.Errorf("Failed to mount images storage: %w", err))
```
tomponline: response.SmartError

lxd/api_project.go
```diff
@@ -952,13 +996,35 @@ func projectDelete(d *Daemon, r *http.Request) response.Response {
...
+		}
+	}
+
+	if backupsVolume != "" {
+		err := unmountCustomVolume(s, backupsVolume)
+		if err != nil {
+			return response.BadRequest(fmt.Errorf("Failed to mount backups storage: %w", err))
```
tomponline: response.SmartError

lxd/db/cluster/projects.go
```diff
@@ -117,6 +117,42 @@ SELECT projects_config.value
...
+`
+	rows, err := tx.QueryContext(ctx, stmt, config)
+	if err != nil {
+		return nil, err
+	}
+
+	defer func() { _ = rows.Close() }()
+
+	result := make(map[string]string)
+	for i := 0; rows.Next(); i++ {
```
tomponline: for rows.Next() should suffice here, as i isnt used

lxd/db/cluster/projects.go
```diff
@@ -117,6 +117,42 @@ SELECT projects_config.value
...
+		var name string
```
tomponline: can condense into `var name, configValue string`

lxd/api_1.0.go
```diff
@@ -663,28 +689,42 @@ func doAPI10Update(d *Daemon, r *http.Request, req api.ServerPut, patch bool) re
...
-			nodeValues["storage.backups_volume"], err = daemonStorageValidate(s, backupsPoolVolume)
+			nodeValues[config], err = daemonStorageValidate(s, backupsPoolVolume)
 			if err != nil {
-				return fmt.Errorf("Failed validation of %q: %w", "storage.backups_volume", err)
+				return fmt.Errorf("Failed validation of %q: %w", config, err)
+			}
+
+			for project, volume := range projectsBackupsStorage {
+				if volume == backupsPoolVolume {
+					return fmt.Errorf(`Failed validation of "%q": storage volume already configured as %q of project %q`, config, config, project)
```
tomponline: double quoting here, only %q is needed, no need to wrap in quotes

lxd/api_1.0.go
```diff
@@ -663,28 +689,42 @@ func doAPI10Update(d *Daemon, r *http.Request, req api.ServerPut, patch bool) re
...
-			nodeValues["storage.images_volume"], err = daemonStorageValidate(s, imagesPoolVolume)
+			nodeValues[config], err = daemonStorageValidate(s, imagesPoolVolume)
 			if err != nil {
-				return fmt.Errorf("Failed validation of %q: %w", "storage.images_volume", err)
+				return fmt.Errorf("Failed validation of %q: %w", config, err)
+			}
+
+			for project, volume := range projectsImagesStorage {
+				if volume == imagesPoolVolume {
+					return fmt.Errorf(`Failed validation of "%q": storage volume already configured as %q of project %q`, config, config, project)
```
tomponline: double quoting here, only %q is needed, no need to wrap in quotes

lxd/api_project.go
```diff
@@ -257,6 +258,33 @@ func projectUsedBy(ctx context.Context, tx *db.ClusterTx, project *cluster.Proje
 	return usedBy, nil
 }
 
+func projectStorageSetup(s *state.State, config string, storageType string) error {
```
tomponline: would benefit from a docblock explaining what it does.

lxd/daemon_storage.go
```diff
@@ -40,40 +61,19 @@ func daemonStorageVolumesUnmount(s *state.State, ctx context.Context) error {
 		return err
 	}
 
-	unmount := func(source string) error {
```
tomponline, line 43: Whilst looking at this change I noticed an optimisation/correctness change we can make in `daemonStorageVolumesUnmount`, instead of loading the node config again, can we use `s.LocalConfig`?

lxd/daemon_storage.go
```diff
@@ -101,36 +122,15 @@ func daemonStorageMount(s *state.State) error {
 		return err
 	}
 
-	mount := func(source string) error {
```
tomponline, line 104: Whilst looking at this change I noticed an optimisation/correctness change we can make in `daemonStorageMount`, instead of loading the node config again, can we use `s.LocalConfig`?

lxd/api_project.go
tomponline: Can this function return a revert.Hook which can then be used by the callers to undo this if the subsequent steps (like DB changes) fail?

lxd/api_project.go
```diff
@@ -319,6 +347,20 @@ func projectsPost(d *Daemon, r *http.Request) response.Response {
 		return response.BadRequest(err)
 	}
 
+	if project.Config["storage.images_volume"] != "" {
```
tomponline, line 380: I think it'd be good to start using the

```go
	revert := revert.New()
	defer revert.Fail()
```

pattern here so that we can undo on-disk changes if the DB steps fail later.

lxd/db/cluster/projects.go
```diff
@@ -96,8 +96,8 @@ func (p *Project) ToAPI(ctx context.Context, tx *sql.Tx) (*api.Project, error) {
 	return apiProject, nil
 }
 
-// ProjectConfig is a helper to return value of an config option of a project.
-func ProjectConfig(ctx context.Context, tx *sql.Tx, project string, config string) (string, error) {
+// GetProjectConfigValue is a helper to return value of an config option of a project.
+func GetProjectConfigValue(ctx context.Context, tx *sql.Tx, project string, config string) (string, error) {
```
tomponline: this function was added in an earlier commit and changed name here, so we should avoid doing this.

test/suites/remote.sh
tomponline, line 1: Can this commit be moved into a separate PR so we can keep the changes in this PR specifically related to project daemon storage?

tomponline (changes requested):
Hi @skozina, thanks for this, I spent quite a while going through this closely.

Overall its looking good. Although by the time I got to commit `lxd/daemon: Implement per-project images storage` I was starting to flag and the commit was too large with too many only tangentially related changes for me effectively reason about them or review it.

So I'd like it if you can break that commit up a bit into logical changes.

One thing I thought of whilst going through this PR that struck me that I don't recall us discussing before is how a single cluster-wide project setting like `storage.images_volume` is going to translate to the custom volumes on local storage pools (i.e those drivers which dont have `MultiVolume=true`) where the custom volume specified is specific to a cluster member.

This means when I do `lxc storage volume create foo bar --target=member1` the bar volume only exists on member1.

With the daemon level `storage.images_volume` this is stored in the per-cluster member local config so is nicely aligned with the per-cluster member concept.

However project config is cluster wide and so we cant specify a per-member per-project config value (would we even want to for this feature?).

So that begs the question, how do we deal with this? 
One approach perhaps is to require that a matching volume exists on every cluster member before the project setting can be specified. We would then also need to consider joining new cluster members, and how the init/preseed process would need to create those volumes.

What led me to this is what I couldn't see any evidence of cluster wide notification requests when setting up a project's daemon storage volume (mkdir dir, mounting, chmod etc), which would need to occur on each cluster member.

skozina:
Thank you for your review @tomponline !

> Overall its looking good. Although by the time I got to commit `lxd/daemon: Implement per-project images storage` I was starting to flag and the commit was too large with too many only tangentially related changes for me effectively reason about them or review it.
> 
> So I'd like it if you can break that commit up a bit into logical changes.

Sure, I'll look into breaking that commit, and the following `lxd/daemon: Implement per-project backups storage`, into smaller pieces. Note that without this full commit, the new feature will not work completely.

> One thing I thought of whilst going through this PR that struck me that I don't recall us discussing before is how a single cluster-wide project setting like `storage.images_volume` is going to translate to the custom volumes on local storage pools (i.e those drivers which dont have `MultiVolume=true`) where the custom volume specified is specific to a cluster member.
> 
> This means when I do `lxc storage volume create foo bar --target=member1` the bar volume only exists on member1.
> 
> With the daemon level `storage.images_volume` this is stored in the per-cluster member local config so is nicely aligned with the per-cluster member concept.

Thank you for bringing this up, I missed this aspect. I can see that for local storage pools (those without `MultiVolume=true`), the volumes are created only on that particular node. The daemon-level setting, when set on that particular node, then uses the volume on that particular node.
 
> However project config is cluster wide and so we cant specify a per-member per-project config value (would we even want to for this feature?).

It wouldn't feel intuitive to have to configure this setting for every single node. If we'd want to do that, I suppose we could introduce the node-level configs such as `storage.project_foo.images_volume`, which would nicely align all the options and code regarding the images storage to the same place.

> So that begs the question, how do we deal with this? One approach perhaps is to require that a matching volume exists on every cluster member before the project setting can be specified. We would then also need to consider joining new cluster members, and how the init/preseed process would need to create those volumes.

I think we can do that with the following requirements (which both seem to be met):
 - the newly joining cluster members share the storage pool configuration (so the only thing to do on `cluster add` would be to create the respective volume, if it's a local storage, and mount it)
 -  there are no images stored on the new cluster members (so there's no need to migrate images to the per-project storage on cluster join).

So to be more specific:
 - When project-level setting is set, the verification will ensure volume exists on all cluster members (if it's a local storage). If yes, it will be mounted on all cluster members.
 - When new nodes join the cluster, the joining mechanism will ensure the volume is created on the new members, and mount it.

I'll look at that. Thanks for the tip!

> What led me to this is what I couldn't see any evidence of cluster wide notification requests when setting up a project's daemon storage volume (mkdir dir, mounting, chmod etc), which would need to occur on each cluster member.

Yup.

lxd/api_project.go
skozina: Yup, thanks, I've removed this from the commit which introduced this (I only had further commit deleting the line, which is not ideal).

lxd/api_project.go
skozina: Sounds good, changed to "changed". TY!

lxd/api_project.go
skozina: Sounds good, done. I've also renamed the arguments in the calling functions from `storage` to `daemonStorageVolume`.

lxd/daemon_storage.go
skozina: Sure, added a docblock, and renamed `config` to `daemonStorageVolume`, and `target` to `storageType`.

lxd/db/cluster/projects.go
skozina: Thanks, fixed.

lxd/db/cluster/projects.go
skozina: Thanks, fixed.

lxd/api_1.0.go
skozina: Sure, done.

lxd/daemon.go
skozina: You mean the first commit, changing just the function signature, would basically break build (because the call sites are not yet updated), and only the second commit would make it work again, is that right?
Sure, I can do that next time. I am sorry I didn't do that first time right, I guess I'm still trying to make every commit to work well, forgetting about backports;-/

skozina: Sure, renamed.

lxd/images.go
skozina: I've changed the context here, thanks for catching that.

Wrt. the arguments isolation, you mean to first extend the function signature, and use default values at the call sites (which could be safely backported), and only in a separate commit actually start implementing the new per-project options with the new use of the arguments, is that it?

lxd/daemon_storage.go
skozina: Sure, renamed.

lxd/daemon_storage.go
skozina: Renamed.

lxd/api_project.go
skozina: Ah, makes sense. Thanks! Fixed.

lxd/api_project.go
skozina: Fixed.

lxd/api_project.go
skozina: Fixed.

lxd/api_project.go
skozina: Fixed.

lxd/db/cluster/projects.go
skozina: True, done.

lxd/db/cluster/projects.go
skozina: Right, done.

lxd/api_1.0.go
skozina: Right, fixed.

lxd/api_1.0.go
skozina: Fixed.

test/suites/remote.sh
skozina: Sure, opened #15952 and dropped from this PR.

lxd/daemon.go
tomponline: @skozina yes thats right. I've seen some projects that require every commit to be buildable and functional, but that isn't something I've ever seen be required on LXD.

Instead I've found it more useful from a review perspective to have a commit that adds/changes the function first, followed by a subsequent commit to update the existing usage.

lxd/images.go
tomponline: See https://github.com/canonical/lxd/pull/15827#discussion_r2189309538

In general this is less about backporting and more about easing reviews by highlighting the change of the function first before the updated call sites.

It can help with backports too if the call sites are different between the main and LTS branches and so require manual modification anyway.

tomponline:
> It wouldn't feel intuitive to have to configure this setting for every single node. If we'd want to do that, I suppose we could introduce the node-level configs such as `storage.project_foo.images_volume`, which would nicely align all the options and code regarding the images storage to the same place.

Yes this is an option, it would allow per-member project storage settings indeed and would align with how its configured today at the daemon level. 

As for the alternative, I agree its more intuitive once the cluster is configured:

> the newly joining cluster members share the storage pool configuration (so the only thing to do on cluster add would be to create the respective volume, if it's a local storage, and mount it)
> there are no images stored on the new cluster members (so there's no need to migrate images to the per-project storage on cluster join).

I suggest we check these statements to check for validity, as I believe a non-empty server can be joined to a cluster.
Though not sure how functional that is. However when I've worked on the join process in the past I remember that the joining machine sends its current network and storage configuration to the cluster for validation before proceeding with the join. This is something that @roosterfish is also exploring to understand/change its behaviour to allow for `lxd recover` and `lxd init` on a DR site when using storage appliance provided replication.

Specifically I am referring to this sequence here chaps:

https://github.com/canonical/lxd/blob/2d863c75f7898dd6eb173ae96472f18578b5b248/lxd/api_cluster.go#L648-L717

We should understand what happens today if projects ans volumes exist on the standalone server that conflict with the cluster settings that would be applied when joining, and perhaps need to extend the pre-join validation to reject it.

lxd/daemon_storage.go
skozina: Good catch! Thank you, added a new commit for this one.

tomponline:
> What led me to this is what I couldn't see any evidence of cluster wide notification requests when setting up a project's daemon storage volume (mkdir dir, mounting, chmod etc), which would need to occur on each cluster member.

@skozina the general pattern for implementing cluster-wide actions from an API request is to use:

```
cluster.NewNotifier(s, s.Endpoints.NetworkCert(), s.ServerCert(), cluster.NotifyAll)
```

Combined with API endpoint specific logic that detects the request is coming as part of a intra-cluster notification and only applies the member specific changes for the request.

This is detected using either `isClusterNotification` or the newer `UserAgentClientType` combined with the associated constants for client type.

The latter option might be useful as it also allows differentiating between a member joining the cluster and a post-join change.

lxd/api_project.go
skozina: Added a docblock to the function.

I've improved the logic of creation of the directories a bit. Instead of returning a `revert.Reverter` I've added the Reverter as a new parameter of the function, so the caller is responsible to allocate it and `Fail()` / `Success()` as needed.

lxd/daemon_storage.go
skozina: Thanks, added in a new commit.

lxd/api_project.go
skozina: Right, added handling a the `revert.Reverter` here, and adjusted `projectStorageSetup()` to add the necessary Hooks to it.

lxd/db/cluster/projects.go
skozina: Yup, thanks, updated the older commit to stick with the new name.

lxd/db/cluster/projects.go
skozina: Thanks for the idea! I've re-worked `GetProjectConfig()` so it's now hand written with SQL instance loading only this project config, and also to take project name as an argument rather than ID. I've updated `ProjectHasImages` and `ToAPI` and all the rest of this PR to use this function, and dropped the `ProjectConfig`, which was later renamed to `GetProjectConfigValue`, from this PR.

skozina: As for the `GetConfig` audit, I've opened a new issue for that one: #15972

lxd/db/cluster/projects.go
skozina: Actually, based on other review feedback, this function was dropped completely.

test/suites/projects.sh
```diff
@@ -1170,3 +1171,119 @@ test_projects_before_init() {
...
+  ! lxc project unset foo storage.images_volume || false
+
+  # It should be possible to share the volume among multiple projects
+  lxc project create foo2 -c storage.images_volume="${pool}/vol"
+  deps/import-busybox --project foo2 --alias testimage
+  lxc image delete --project foo2 testimage
+  lxc project delete foo2
+
+  # Import the image in the default project and storage, and ensure it's gone after removal
+  deps/import-busybox --alias testimage
```
simondeziel: `ensure_import_testimage` has some additional smart for caching and always operate on the current project. As such, it should be preferred when injecting the `testimage` in the current project.
```suggestion
  ensure_import_testimage
```

lxd/daemon.go
skozina: Thanks, makes sense.

test/suites/projects.sh
skozina: Thanks! I've replaced this with `ensure_import_testimage` on all the 4 places, and added project switching as necessary.

lxd/api_project.go
```diff
@@ -941,13 +1015,35 @@ func projectDelete(d *Daemon, r *http.Request) response.Response {
...
+		err := unmountDaemonStorageVolume(s, imagesVolume)
+		if err != nil {
+			return response.SmartError(fmt.Errorf("Failed to mount images storage: %w", err))
+		}
+	}
+
+	if backupsVolume != "" {
+		err := unmountDaemonStorageVolume(s, backupsVolume)
+		if err != nil {
+			return response.SmartError(fmt.Errorf("Failed to mount backups storage: %w", err))
```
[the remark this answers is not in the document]
skozina: Good point, done.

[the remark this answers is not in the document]
skozina: Dup with the above, done.

lxd/storage/utils.go
```diff
@@ -838,7 +838,7 @@ func ImageUnpack(s *state.State, imageFile string, vol drivers.Volume, destBlock
 		}
 	} else {
 		// Dealing with unified tarballs require an initial unpack to a temporary directory.
-		tempDir, err := os.MkdirTemp(s.ImagesStoragePath(), "lxd_image_unpack_")
+		tempDir, err := os.MkdirTemp(s.ImagesStoragePath(""), "lxd_image_unpack_")
```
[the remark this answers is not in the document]
skozina: Heh, there's no `ProjectSpecificImagesVolumePath()` function;-)

There are two places where the temporary image files are unpacked under the default daemon storage rather than the project specific image storage -- `driver_qemu.go:Export()` and this `ImageUnpack()` function, which is only called from `storage/backend_lxd.go:imageFiller()`. I've noticed that the qemu driver structure already contains pointer to the `api.Project` with all the project config. So we can easily adjust the driver_qemu.go to unpack temporary files on export to the project specific storage. Similarly `imageFiller()` already has parameter with the project-specific storage for temporary files, so we can just easily pass it to the `ImageUnpack()`.

So yes, lets put even the temporary files under the project specific storage. Good point Copilot!

skozina:
> We should understand what happens today if projects ans volumes exist on the standalone server that conflict with the cluster settings that would be applied when joining, and perhaps need to extend the pre-join validation to reject it.

I've played a bit with the current behavior.

If there's an existing project `foo` on a node joining the existing cluster, the project is removed.

If a cluster has a storage pool `foo`, and the node joining the cluster has a different storage pool `bar`, then the `bar` pool is removed on the node and a new pool `foo` is created. So there's a precedence on creating resources on the node joining the cluster.

If a cluster has a storage pool `foo`, and the node also has a storage pool `foo`, I can't get the node to join the cluster. It's failing on:
```
Error: Failed to join cluster: Failed to initialize member: Failed to initialize storage pools and networks: Failed to update storage pool "foo": Config key "size" is cluster member specific
```
This comes from https://github.com/canonical/lxd/blob/69aa6f7b1e1bc8c2f8f2944cba4d0bab6ae13979/lxd/storage_pools.go#L410

When a node leaves a cluster, the project and storage pools from the cluster are removed from the node.

So I think it's fair to assume that the storage pools are freshly created on the node joining the cluster, so for the purpose of this PR we can also create the per-project configure storage volumes.

skozina:
> Specifically I am referring to this sequence here chaps:
> 
> https://github.com/canonical/lxd/blob/2d863c75f7898dd6eb173ae96472f18578b5b248/lxd/api_cluster.go#L648-L717

This piece of code (PUT call on `/1.0/cluster` endpoint) doesn't seem to be running at all in my case on new node joining the cluster.

What I am seeing is that the node does a bunch of GET calls to get all of the projects, storage pools, network in the cluster. I suppose it then creates all the local resources as needed. Finally, it does a single POST call to `/internal/cluster/accept`, which is where the cluster leader verifies the new node configuration. Here: https://github.com/canonical/lxd/blob/2d863c75f7898dd6eb173ae96472f18578b5b248/lxd/api_cluster.go#L2494

skozina:
The last two commits are adding the notification calls to other cluster members to mount the configured project-level storage. This feels cumbersome, so I'm going to close this PR now and look at re-working this to be specific to each cluster member.

PR #15832, opened 2025-06-19
Auth: Cluster secret

markylaing:
Adds a slice of secrets to the database for encrypting OIDC credential cookies.

### DIFF
--- a/test/suites/serverconfig.sh
+++ b/test/suites/serverconfig.sh
@@ -5,8 +5,21 @@ test_server_config() {
 
   _server_config_access
   _server_config_storage
+  _server_config_auth_secret
 
   kill_lxd "${LXD_SERVERCONFIG_DIR}"
+}
+
+_server_config_auth_secret() {
+    # Validate core.auth_secret_expiry cannot be set to less than one day
+    lxc config set core.auth_secret_expiry="1d"
+    ! lxc config set core.auth_secret_expiry='23H 59M 59S' || false
+    lxc config set core.auth_secret_expiry='23H 59M 60S'
+    ! lxc config set core.auth_secret_expiry='1439M 59S' || false
+    lxc config set core.auth_secret_expiry='1439M 60S'
+    ! lxc config set core.auth_secret_expiry='86399S' || false
+    lxc config set core.auth_secret_expiry='86400S'
+    lxc config unset core.auth_secret_expiry
 }
 
 _server_config_access() {


doc/api-extensions.md
```diff
@@ -2676,3 +2676,7 @@ Adds a `skip` mode to the restore request. This mode restores a cluster member's
 ## `disk_io_threads_virtiofsd`
 
 Adds the {config:option}`device-disk-device-conf:io.threads` option on `disk` devices which is used to control the `virtiofsd` thread pool size when sharing file systems into VMs. This can help improve I/O performance.
+
+## `cluster_secret`
+
+Adds the {config:option}`server-core:core.secret_lifetime` option to configure how long a given secret should be used for.
```
```suggestion
Adds the {config:option}`server-core:core.secret_lifetime` option to configure how long a given secret is used.
```

lxd/cluster/config/config.go
```diff
@@ -501,6 +511,23 @@ var ConfigSchema = config.Schema{
 	//  shortdesc: Whether to automatically trust clients signed by the CA
 	"core.trust_ca_certificates": {Type: config.Bool, Default: "false"},
 
+	// lxdmeta:generate(entities=server; group=core; key=core.secret_lifetime)
+	// The cluster secret is used for various cryptographic purposes, such as cookie encryption.
+	// When a given secret is older than the configured lifetime, a new secret is generated.
+	// The secret should be rotated regularly.
```
minaelee: It's unclear to me what action is advised by "The secret should be rotated regularly." Does this mean that someone has to manually rotate something? Or that we recommend the user sets this config option so that the secret is automatically rotated regularly?

lxd/cluster/config/config.go
```diff
@@ -501,6 +511,23 @@ var ConfigSchema = config.Schema{
...
+	//
+	// This configuration option accepts multiple space separated values of the form `[0-9]+(S|M|H|d|w|m|y)`,
```
minaelee:
```suggestion
	// This configuration option accepts multiple space-separated values of the form `[0-9]+(S|M|H|d|w|m|y)`,
```

minaelee (changes requested):

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,84 @@
...
+
+// IsValid returns true if the first Secret in Secrets is younger than the given lifetime.
+func (s Secrets) IsValid(lifetime int64) bool {
+	return len(s) > 0 && s[0].IsValid(lifetime)
+}
+
+// Rotate prepends a new secret. Currently, the slice length is limited to 2.
+func (s *Secrets) Rotate() {
+	secretValue := make([]byte, 512)
+	_, _ = rand.Read(secretValue)
```
alanmcanonical: I saw the `DeviceNextInterfaceHWAddr` func in lxd/network/network_utils.go return and handle the error from crypto/rand, so would be better to handle err here too?

lxd/daemon.go
```diff
@@ -661,6 +665,57 @@ func (d *Daemon) handleOIDCAuthenticationResult(r *http.Request, result *oidc.Au
...
+
+	if d.internalSecrets.IsValid(lifetime) {
+		return d.internalSecrets, nil
+	}
+
+	err = d.db.Cluster.Transaction(ctx, func(ctx context.Context, tx *db.ClusterTx) error {
+		row := tx.Tx().QueryRowContext(ctx, `SELECT value FROM config WHERE key = 'volatile.secrets'`)
+		var secrets dbCluster.Secrets
+		err := row.Scan(&secrets)
+		if err != nil && !errors.Is(err, sql.ErrNoRows) {
```
alanmcanonical: What if err is sql.ErrNoRows?

lxd/cluster/config/config.go
markylaing: I've just removed the line. No action needs to be taken by the user.

lxd/db/cluster/secrets.go
markylaing: I ignored the error here per the golang docs https://pkg.go.dev/crypto/rand#Read

> Read fills b with cryptographically secure random bytes. It never returns an error, and always fills b entirely.

lxd/daemon.go
markylaing: This leaves `secrets` to be an empty slice. The next call to `IsValid` will return false, and then the secrets will be rotated. Rotating the secrets on an empty slice will return a new slice with one element, containing a newly minted secret.

minaelee (approved):
Thanks for the changes! One very tiny nit, otherwise LGTM.

alanmcanonical (approved):
Lgtm, great work.

nmezhenskyi (approved):
LGTM, thanks!

tomponline:
@markylaing is this updated with new DB schema or should it be in draft?

markylaing:
> @markylaing is this updated with new DB schema or should it be in draft?

It will be updated with the new DB schema in a few minutes :wink:

markylaing:
But yes I should have marked as draft while updating the spec, apologies

markylaing:
Marking as draft until #16120 is merged anyway

markylaing:
Will rebase to main when #16120 is merged and test locally before unmarking as draft and requesting review

markylaing:
@tomponline ready for review when you have time, thanks

doc/api-extensions.md
```diff
@@ -2690,3 +2690,7 @@ This adds PCI device hotplugging for VMs.
 
 ## `device_patch_removal`
 The `PATCH /1.0/instances/{name}` endpoint allows removing an instance device by setting its value to `null` in the devices map.
+
+## `auth_secrets`
```
tomponline, line 2694: The description makes it seem like a name of `auth_secrets_expiry` would be more relevant. 
However I think the description could do with being expanded to explain that this introduces a core auth secret and what its used for.

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,135 @@
...
+// authSecretLength defines the length of an AuthSecretValue.
+const authSecretLength = 64
+
+// AuthSecretValue is used to byte slice of length authSecretLength in a text column as a base64 encoded string.
+type AuthSecretValue []byte
+
+// newAuthSecretValue returns a AuthSecretValue, read from [rand.Reader].
+func newAuthSecretValue() AuthSecretValue {
+	secretValue := make([]byte, authSecretLength)
+	_, _ = rand.Read(secretValue)
```
tomponline, line 33: Does this need seeding or is that not required these days in Go?

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// ScanText implements [query.TextScanner] for AuthSecretValue to simplify the [sql.Scanner] implementation.
+func (s *AuthSecretValue) ScanText(str string) error {
+	out, err := base64.StdEncoding.DecodeString(str)
+	if err != nil {
+		return err
+	}
+
+	*s = out
```
tomponline: Is it normal for unvalidated values to be stored in the struct (and remain there) even if the ScanText function returns error? Seems like it could lead to use of the unvalidated value.

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,135 @@
...
+func (s *AuthSecretValue) Scan(value any) error {
+	return query.ScanValue(value, s, false)
+}
+
+// SecretType represents the "type" column in the secrets table.
+type SecretType string
+
+const (
+	// SecretTypeClusterAuth is the SecretType for global auth secrets.
+	SecretTypeClusterAuth SecretType = "cluster_auth"
```
tomponline: this secret is used for non-clusters too right, so would `CookieAuth` be more suitable?

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,135 @@
...
+func newAuthSecret() AuthSecret {
+	return AuthSecret{
+		Value:     newAuthSecretValue(),
+		CreatedAt: time.Now().UTC(),
+	}
+}
+
+// IsValid returns true if the time now is less than the time that the secret was created, plus the expiry.
+func (s AuthSecret) IsValid(expirySeconds int64) bool {
+	return s.CreatedAt.Add(time.Duration(expirySeconds)*time.Second).After(time.Now().UTC()) && s.Value.Validate() == nil
```
tomponline: As we already have `Validate` for the type above, `IsValid` seems to stand out as similar in name, but quite different in purpose, how about we call this `IsExpired()`?

lxd/daemon.go
```diff
@@ -669,6 +674,88 @@ func (d *Daemon) handleOIDCAuthenticationResult(r *http.Request, result *oidc.Au
...
+	// Check if the current in-memory secrets are valid.
+	if d.internalSecrets.IsValid(expiry) {
+		// If valid, return a copy.
+		return slices.Clone(d.internalSecrets), nil
+	}
+
+	// Otherwise, start a transaction.
+	err = d.db.Cluster.Transaction(ctx, func(ctx context.Context, tx *db.ClusterTx) error {
+		// Get the secrets from the database.
+		q := `SELECT value, created_at FROM secrets WHERE entity_type = ? AND entity_id = 0 AND type = ? ORDER BY created_at DESC`
```
tomponline: Lets not have queries in the daemon itself and instead push this down into a cluster db function.

tomponline: Reduces cyclomatic complexity for one thing.

lxd/daemon.go
```diff
@@ -669,6 +674,88 @@ func (d *Daemon) handleOIDCAuthenticationResult(r *http.Request, result *oidc.Au
...
+			// If valid, set internal secrets to the value defined in the database and exit transaction.
+			d.internalSecrets = secrets
+			return nil
+		}
+
+		// Rotate the secrets. If there were none, this will populate the array.
+		rotatedSecrets := secrets.Rotate()
+
+		// Add the new secret to the database.
+		res, err := tx.Tx().ExecContext(ctx, `INSERT INTO secrets (entity_type, entity_id, type, value, created_at) VALUES (?, 0, ?, ?, ?)`, dbCluster.EntityType(entity.TypeServer), dbCluster.SecretTypeClusterAuth, rotatedSecrets[0].Value, rotatedSecrets[0].CreatedAt)
```
tomponline: this should be a separate function in the cluster db package.

lxd/daemon.go
```diff
@@ -669,6 +674,88 @@ func (d *Daemon) handleOIDCAuthenticationResult(r *http.Request, result *oidc.Au
 	return nil
 }
 
+// getSecrets gets a copy of the current, cluster-wide secrets.
+func (d *Daemon) getSecrets(ctx context.Context) (dbCluster.AuthSecrets, error) {
```
tomponline, line 683: in general this function is too complex and should be broken up, starting with at least moving the DB queries to functions with the other DB query functions.

lxd/auth/oidc/oidc.go
```diff
@@ -543,49 +557,49 @@ func (o *Verifier) startSession(ctx context.Context, w http.ResponseWriter, idTo
 }
 
 // getCookies gets the sessionID, identity and refresh tokens from the request cookies and decrypts them.
-func (o *Verifier) getCookies(r *http.Request) (sessionIDPtr *uuid.UUID, idToken string, refreshToken string, err error) {
+func (o *Verifier) getCookies(r *http.Request) (idToken string, refreshToken string, startNewSession bool, err error) {
```
tomponline, line 562: Please can you update the doc block to explain when startNewSession is true?

lxd/db/cluster/secrets.go
markylaing: The `math/rand` package can be seeded because it uses a psuedorandom number generator thats defined in Go, but I don't think `crypto/rand` can be seeded because it calls OS APIs ([getrandom](https://manpages.debian.org/unstable/manpages-dev/getrandom.2.en.html), falling back to reading from `/dev/urandom`).

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// ScanInteger implements [query.IntegerScanner] for SecretType to simplify the [sql.Scanner] implementation.
+func (s *SecretType) ScanInteger(code int64) error {
+	switch code {
+	case secretTypeCodeClusterAuth:
+		*s = SecretTypeClusterAuth
+	}
+
+	return fmt.Errorf("Invalid secret type code %d", code)
```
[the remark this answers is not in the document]
markylaing: The point is that it's not possible to scan an invalid or empty value.

lxd/db/cluster/secrets.go
tomponline: ah yes ofcourse, thanks

lxd/metadata/configuration.json
```diff
@@ -5110,6 +5110,15 @@
 			},
 			"core": {
 				"keys": [
+					{
+						"core.auth_secret_expiry": {
+							"defaultdesc": "`1m`",
+							"longdesc": "The secret is used for various cryptographic purposes, such as cookie encryption.\nWhen a given secret is older than the configured expiry, a new secret is generated.\n\nThis configuration option accepts multiple space-separated values of the form `[0-9]+(S|M|H|d|w|m|y)`,\nwhere `S` is seconds, `M` is minutes, `H` is hours, `d` is days, `w` is weeks, `m` is months, and `y` is years.\nFor example, `1d 3H` is 1 day and 3 hours.\n\nThe default value is `1m` (1 month).",
+							"scope": "global",
+							"shortdesc": "How long to use a given cluster secret.",
+							"type": "bool"
```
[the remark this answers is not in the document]
markylaing: Fixed

doc/api-extensions.md
markylaing: Added more description

lxd/db/cluster/secrets.go
markylaing: Good catch, thanks :)

lxd/db/cluster/secrets.go
markylaing: I changed this to `core_auth` since we'll be using it for signing tokens in future too

lxd/db/cluster/secrets.go
markylaing: Changed to `IsExpired` and changed all `if IsValid() { ... }` to `if !IsExpired() { ... }`.

lxd/daemon.go
markylaing: Moved all the sql into `lxd/db/cluster/secrets.go`. My motivation behind keeping it all in one place was so that secret rotation can only happen in one place, but I appreciate it's harder to reason about.

lxd/daemon.go
markylaing: Moved to `lxd/db/cluster/secrets.go`

lxd/daemon.go
markylaing: Made this much simpler. The `(AuthSecrets).Rotate` function now accepts a transaction and will perform the rotation in-memory and in the database at once.

lxd/auth/oidc/oidc.go
markylaing: Updated doc blocks for `getCookies` and `secureCookieFromSession`

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,249 @@
...
+	"time"
+
+	"github.com/canonical/lxd/lxd/db/query"
+	"github.com/canonical/lxd/shared/entity"
+)
+
+// authSecretLength defines the length of an AuthSecretValue.
+const authSecretLength = 64
+
+// AuthSecretValue is used to byte slice of length authSecretLength in a text column as a base64 encoded string.
```
markylaing: typo here, will fix now

markylaing:
Remarking as draft as I've found an oddity regarding the expiry configuration

markylaing:
@tomponline ready for review, thanks :)

markylaing:
Not sure why the LVM and ZFS cluster tests failed but there's nothing in this PR that should affect the storage layer.

lxd/cluster/config/config.go
```diff
@@ -500,6 +505,34 @@ var ConfigSchema = config.Schema{
...
+	//
+	// The default value is `1m` (1 month).
+	// The minimum value is `1d` (1 day).
+	// ---
+	//  type: string
+	//  scope: global
+	//  defaultdesc: `1m`
+	//  shortdesc: How long to use a given cluster secret.
+	"core.auth_secret_expiry": {Default: "1m", Validator: func(s string) error {
+		exp, err := shared.GetExpiry(time.Unix(0, 0), s)
```
tomponline: Due to the issues around date differences you mentioned the other day, would it makes things easier to have `shared.GetExpiry` also return a time.Duration (which it is already calculating internally anyway) so that you can simply test the duration of the expiry?

lxd/cluster/config/config.go
```diff
@@ -500,6 +505,34 @@ var ConfigSchema = config.Schema{
...
+	// where `S` is seconds, `M` is minutes, `H` is hours, `d` is days, `w` is weeks, `m` is months, and `y` is years.
+	// For example, `1d 3H` is 1 day and 3 hours.
+	//
+	// The default value is `1m` (1 month).
+	// The minimum value is `1d` (1 day).
+	// ---
+	//  type: string
+	//  scope: global
+	//  defaultdesc: `1m`
+	//  shortdesc: How long to use a given cluster secret.
```
tomponline: All other `shortdec` dont end with dot.

lxd/db/cluster/schema.go
```diff
@@ -481,6 +481,17 @@ CREATE TABLE "projects_config" (
     FOREIGN KEY (project_id) REFERENCES "projects" (id) ON DELETE CASCADE,
     UNIQUE (project_id, key)
 );
+CREATE TABLE secrets (
+    id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+    entity_type INTEGER NOT NULL,
+    entity_id INTEGER NOT NULL,
+    type INTEGER NOT NULL,
+    value TEXT NOT NULL,
+    created_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP
```
tomponline: all other time fields in LXD use "_date" suffix on their columns, so this one stands out as using a new convention.

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,270 @@
...
+
+// newAuthSecretValue returns a AuthSecretValue, read from [rand.Reader].
+func newAuthSecretValue() AuthSecretValue {
+	secretValue := make([]byte, authSecretLength)
+	_, _ = rand.Read(secretValue)
+	return secretValue
+}
+
+// Validate checks that the AuthSecretValue has length authSecretLength.
+func (s AuthSecretValue) Validate() error {
```
tomponline, line 38: Can you avoid mixing function receiver types for the same type (Ive seen some linters pick up on this in the past) so this should be (s *AuthSecretValue)

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,270 @@
...
+	if len(s) != authSecretLength {
+		return fmt.Errorf("Secret must have length %d", authSecretLength)
+	}
+
+	return nil
+}
+
+// Value implements [driver.Valuer] for AuthSecretValue.
+func (s AuthSecretValue) Value() (driver.Value, error) {
```
tomponline, line 47: (s *AuthSecretValue)

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,270 @@
...
+	SecretTypeCoreAuth SecretType = "core_auth"
+)
+
+const (
+	// secretTypeCodeCoreAuth is the database code for SecretTypeCoreAuth.
+	secretTypeCodeCoreAuth int64 = 1
+)
+
+// Value implements [driver.Valuer] for SecretType.
+func (s SecretType) Value() (driver.Value, error) {
```
tomponline, line 96: (s *SecretType)

markylaing: These are for a good reason.

We want to be able to use a `SecretType` when writing to the database without it being a pointer, but when reading from the database we always need to pass a reference. This is the same as when implementing a JSON marshaler/unmarshaler.

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,270 @@
...
+	return nil
+}
+
+// Scan implements [sql.Scanner] for AuthSecretValue.
+func (s *AuthSecretValue) Scan(value any) error {
+	return query.ScanValue(value, s, false)
+}
+
+// SecretType represents the "type" column in the secrets table.
+type SecretType string
```
tomponline, line 83: @markylaing we need to decide whether these new types are going to internally represent their DB value (in this case an int) or whether they represent their human readable/API representation (in this case a string), as I want us to ensure we use a consistent approach cc @kadinsayani 

My view on this is that we should use the DB representation internally and provide `String()` function to convert to the human readable/API value, because this way it allows us to use more efficient integers internally for values that are stored in the DB as such.

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,270 @@
...
+	if err != nil {
+		return err
+	}
+
+	*s = value
+	return nil
+}
+
+// Scan implements [sql.Scanner] for AuthSecretValue.
+func (s *AuthSecretValue) Scan(value any) error {
```
tomponline, line 78: Should this type have a `String()` function to print its human readable form?

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,270 @@
...
+	if err != nil {
+		return err
+	}
+
+	expiresAt, err := shared.GetExpiry(s.CreatedAt, expiry)
+	if err != nil {
+		return fmt.Errorf("Failed to check auth secret expiry: %w", err)
+	}
+
+	if time.Now().UTC().After(expiresAt) {
```
tomponline, line 151: nice and simple like it.

lxd/cluster/config/config.go
markylaing: Perhaps the clearest would be:
```
now := time.Now()
exp, err := shared.GetExpiry(now, s)
if err != nil {
    return err
}

if exp.Sub(now) < 24*time.Hour {
    return api.NewStatusError("Auth secret expiry cannot be less than one day")
}

return nil
```

markylaing: Because returning a duration when passing a unix zero is precisely what was wierd. It's not correct if a month is passed in.

lxd/db/cluster/secrets.go
markylaing: Yep sounds good.

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,270 @@
...
+
+	// Add the new value to the database
+	err := createCoreAuthSecret(ctx, tx, newSecret)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to rotate secrets: %w", err)
+	}
+
+	// If the secrets were truncated, then there should be at most one old secret in the database. Delete it and assert
+	// that we deleted one row.
+	if truncated {
```
tomponline: Wouldn't it be simpler and safer (more correct) to take the secrets stored in `rotatedSecrets` and then delete any rows (of the relevant type) that *dont* match those values? Rather than doing another set of date logic in the DB. In my experience the less date comparison you can do the better, especially when relying on 2 different systems to come to the same result :)

tomponline: You could just delete all the relevant rows and insert the new keys - this is what the config setting does for most entity types.

lxd/db/cluster/secrets.go
markylaing: The idea behind using strings here is that it will be human readable everywhere (e.g. if you forget to call `String` or if you forget to write a format string with `%s`/`%q`) and you can never write an invalid value to the database.

markylaing: It also means you can just cast to string when converting to an API type, rather than calling another `ConvertDBTypeToAPIType` function, which will return an error that needs to be handled.

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,270 @@
...
+
+// deleteCoreAuthSecretsOlderThanTimestamp deletes core auth secrets older than the given timestamp. It returns the
+// number of affected rows.
+func deleteCoreAuthSecretsOlderThanTimestamp(ctx context.Context, tx *sql.Tx, timestamp time.Time) (int64, error) {
+	return deleteSecretsOlderThanTimestamp(ctx, tx, entity.TypeServer, 0, SecretTypeCoreAuth, timestamp)
+}
+
+// deleteSecretsOlderThanTimestamp deletes secrets with the given [entity.Type], entity ID, and SecretType that are
+// older than the given timestamp. It returns the number of affected rows.
+func deleteSecretsOlderThanTimestamp(ctx context.Context, tx *sql.Tx, entityType entity.Type, entityID int, secretType SecretType, timestamp time.Time) (int64, error) {
```
tomponline: I suggest replacing this with a function as follows:

```go
deleteSecrets(ctx context.Context, tx *sql.Tx, entityType entity.Type, entityID int, secretType SecretType, secret ...string)
```

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,270 @@
...
+	return fmt.Errorf("Invalid secret type code %d", code)
+}
+
+// Scan implements sql.Scanner for SecretType.
+func (s *SecretType) Scan(value any) error {
+	return query.ScanValue(value, s, false)
+}
+
+// AuthSecret contains an AuthSecretValue and a creation time.
+type AuthSecret struct {
```
tomponline, line 125: should this also contain an optionally populated DB ID for the secret?

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// GetCoreAuthSecrets returns the server core auth secrets.
+func GetCoreAuthSecrets(ctx context.Context, tx *sql.Tx) (AuthSecrets, error) {
+	return getAuthSecrets(ctx, tx, entity.TypeServer, 0, SecretTypeCoreAuth)
+}
+
+// getAuthSecrets returns a slice of AuthSecrets. It scans the "secrets" table for rows of the given [entity.Type], entity
+// ID, and SecretType. Since we are scanning for AuthSecrets, the "value" column must be an AuthSecretValue.
+func getAuthSecrets(ctx context.Context, tx *sql.Tx, entityType entity.Type, entityID int, secretType SecretType) (AuthSecrets, error) {
```
tomponline: This seems like it should be named `getSecrets()` as it doesn't filter just by a specific secret type but rather accepts an argument.

Also the return type of `AuthSecrets` seems too specific and wouldn't we first have a struct type that represents a row in the secrets DB table and then we can convert that to a more specific representation if needed?

I'm a bit concerned we're going a bit type heavy TBH.

lxd/db/cluster/secrets.go
```diff
@@ -0,0 +1,270 @@
...
+	"github.com/canonical/lxd/lxd/db/query"
+	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/entity"
+)
+
+// authSecretLength defines the length of an AuthSecretValue.
+const authSecretLength = 64
+
+// AuthSecretValue is used to store byte slice of length authSecretLength in a text column as a base64 encoded string.
+type AuthSecretValue []byte
```
tomponline, line 22: should this be type string so it aligns with the underlying DB representation?

lxd/daemon.go
```diff
@@ -669,6 +673,56 @@ func (d *Daemon) handleOIDCAuthenticationResult(r *http.Request, result *oidc.Au
...
+	expiry := d.globalConfig.AuthSecretExpiry()
+
+	// Check if the current in-memory secrets are valid.
+	err := d.internalSecrets.Validate(expiry)
+	if err == nil {
+		// If valid, return a copy.
+		return slices.Clone(d.internalSecrets), nil
+	}
+
+	// Otherwise, start a transaction.
```
tomponline, line 698: Might be worth adding a comment here explaining how you've approach the "update the secrets across the cluster" problem, because IIUC you're relying on every member having the same time (reasonable) and so they will all consider the secret expired at the same time and then start checking the database for a freshly generated secret or generating a new one inside a transaction.

lxd/cluster/config/config.go
tomponline: sounds good thanks. Suggest using `.UTC()` too.

lxd/db/cluster/secrets.go
markylaing: As discussed I will add an ID field to AuthSecret so that we can delete by ID :)

lxd/db/cluster/secrets.go
markylaing: Will remove this function and move the contents to GetAuthSecrets.

lxd/db/cluster/secrets.go
markylaing: As discussed, the idea behind these types is so that the returned value that it is the most convenient Go representation. In this case, []byte is passed into `hmac.New` to derive encryption keys, so we'll leave as is.

lxd/daemon.go
markylaing: Yep will do

markylaing:
@tomponline this is ready for another round 🤠

lxd/state/state.go
```diff
@@ -115,6 +116,9 @@ type State struct {
 
 	// StorageReady can be used to track whether all storage pools are successfully started.
 	StorageReady cancel.Canceller
+
+	// Secrets returns the current secrets.
+	Secrets func(ctx context.Context) (cluster.AuthSecrets, error)
```
tomponline, line 121: Not a blocker, but can you rename this to CoreAuthSecrets please?

lxd/daemon.go
```diff
@@ -669,6 +673,62 @@ func (d *Daemon) handleOIDCAuthenticationResult(r *http.Request, result *oidc.Au
...
 }
 
+// getSecrets gets a copy of the current, cluster-wide secrets. The approach can be summarized as follows:
+// 1. Check if the current in-memory value is valid. If valid, return a copy.
+// 2. Check if the current in-database value is valid. If valid, replace in-memory value and return a copy.
+// 3. Rotate the in-database value within the transaction, then replace in-memory value and return a copy.
+// Steps 2 and 3 must happen within the same transaction so that database locking enforces consistency across the
+// cluster. Everything is performed with a lock on the in-memory value. Note that this approach assumes that the UTC
+// time is synchronized across all cluster members, as this is used for validity checking.
+func (d *Daemon) getSecrets(ctx context.Context) (dbCluster.AuthSecrets, error) {
```
tomponline, line 683: Not a blocker, but can you rename this as `getCoreAuthSecrets` please

tomponline (approved):
LGTM thanks!

PR #15834, opened 2025-06-19
Makefile: Set GOTOOLCHAIN to match GOMIN version

kadinsayani:


simondeziel (commented):
What's the intention for this change?

We picked `local` to disable the toolchain mechanism because it introduced unwanted changes that are hard to predict.

kadinsayani:
Determined to be a local env issue.

PR #15836, opened 2025-06-20
Backports (stable-5.21)

tomponline:
Brings 5.21 up to LXD 6.4 level + cloud-init fixes, virtiofsd io threads + macvlan network state and various test, docs and modernize changes.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -664,6 +664,8 @@ jobs:
 
       - name: Install runtime dependencies
         uses: ./.github/actions/install-lxd-runtimedeps
+        with:
+          optional: true
 
       - name: Download system test dependencies
         uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
--- a/.github/actions/install-lxd-runtimedeps/action.yml
+++ b/.github/actions/install-lxd-runtimedeps/action.yml
@@ -1,14 +1,29 @@
 name: Install LXD runtime dependencies
 description: Installs LXD runtime dependencies
 
+inputs:
+  optional:
+    description: Whether to install optional dependencies
+    default: false
+    type: boolean
+
 runs:
   using: composite
   steps:
     - name: Installs LXD runtime dependencies
       shell: bash
+      env:
+        OPTIONAL: ${{ inputs.optional }}
       run: |
         set -eux
         sudo apt-get update
+
+        if [ "${OPTIONAL}" != "false" ]; then
+          echo "Installing optional runtime dependencies"
+          sudo apt-get install --no-install-recommends -y \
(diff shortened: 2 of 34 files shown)


.github/workflows/tests.yml
```diff
@@ -36,12 +47,12 @@ jobs:
       LD_RUN_PATH: "/home/runner/go/bin/dqlite/libs/"
       CGO_LDFLAGS_ALLOW: "(-Wl,-wrap,pthread_create)|(-Wl,-z,now)"
     name: Code
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
simondeziel: This probably slipped in by accident or is missing the following revert as in `main` we reverted back to 22.04 runner.

.github/workflows/tests.yml
```diff
@@ -249,13 +297,12 @@ jobs:
...
       LXD_OVN_NB_CLIENT_KEY_FILE: "/var/snap/microovn/common/data/pki/client-privkey.pem"
       LXD_OVN_NB_CA_CRT_FILE: "/var/snap/microovn/common/data/pki/cacert.pem"
-      LXD_CONCURRENT: "1"
       LXD_VERBOSE: "1"
       LXD_OFFLINE: "1"
       LXD_TMPFS: "1"
       GOTRACEBACK: "crash"
     name: System
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
simondeziel: Same deal here, after a brief period of time on 24.04, we reverted to 22.04.

.github/workflows/tests.yml
```diff
@@ -348,7 +399,7 @@ jobs:
 
   tics:
     name: Tiobe TICS
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
simondeziel, line 421: Same.

tomponline: @simondeziel thanks, i wasnt quite finished (still not actually) with this phase of the backports, but wanted to get an early indicating if the per-branch cache key was working.

simondeziel: FYI, I was wrong for the TiCS job, as we stuck to 24.04 runners as those don't need to run the actual tests, they just process coverage data collected by other jobs.

tomponline: Cool, well land that in main and itll get duly backported

simondeziel:
@tomponline it's now failing in `backup_rename`, probably due to missing https://github.com/canonical/lxd/commit/611b16b8f0eec35448b9a20dd0a313fa92632d95.

Once done, there will only be 20 test cases to go through so it's really nearing the end.

MusicDin (approved):

PR #14884, opened 2025-01-29
Cluster links API and CLI

kadinsayani:
Depends on https://github.com/canonical/lxd/pull/16165.

Introduces LXD cluster links, allowing for secure interaction between separate LXD clusters.

PR #15838, opened 2025-06-20
Auth: Extend TLS identity helper functions to support multiple identity types

kadinsayani:
This pull request extends TLS identity helper functions in preparation for #14884, which introduces cluster link identities, allowing for getting and activating pending TLS identities of different types. Error messages are also improved to return identity type. `createCertificateAddToken` is extracted from `createIdentityTLSPending` to allow for reuse.


tomponline:
@markylaing please can you take a look at this one?

lxd/db/cluster/identities.go
```diff
@@ -428,17 +428,33 @@ func ActivateTLSIdentity(ctx context.Context, tx *sql.Tx, identifier uuid.UUID,
...
+	typeStrings := make([]string, len(identityTypes))
+	for i, idType := range identityTypes {
+		typeStrings[i] = fmt.Sprintf("%d", idType)
+	}
+
+	// Construct statement with identity types.
+	stmt := fmt.Sprintf(getPendingTLSIdentityByTokenSecretStmt, strings.Join(typeStrings, ","))
+
+	// Get identities using secret.
+	identities, err := getIdentitysRaw(ctx, tx, stmt, secret)
```
markylaing: It took me a bit to figure out what's going on here. I'm not clear on why multiple identity types are needed in this function? The secret is unique, so the identity type is determined by the type of the identity that holds it. The only reason the existing query filters on identity type is to prevent unnecessary JSON parsing at the DQLite/SQLite layer.

I think it's clearer to update the existing statement to include the pending cluster link identity type. E.g.

```go
var getPendingTLSIdentityByTokenSecretStmt = fmt.Sprintf(`
SELECT identities.id, identities.auth_method, identities.type, identities.identifier, identities.name, identities.metadata
	FROM identities
	WHERE identities.type IN (%d,%d)
	AND json_extract(identities.metadata, '$.secret') = ?
`, identityTypeCertificateClientPending, identityTypeClusterLinkPending)
```
and leave the function as is.

lxd/identities.go
```diff
@@ -322,7 +322,7 @@ func createIdentityTLSUntrusted(ctx context.Context, s *state.State, peerCertifi
 
 	// Activate the pending identity with the certificate.
 	err = s.DB.Cluster.Transaction(ctx, func(ctx context.Context, tx *db.ClusterTx) error {
-		return dbCluster.ActivateTLSIdentity(ctx, tx.Tx(), identifier, cert)
+		return dbCluster.ActivateTLSIdentity(ctx, tx.Tx(), identifier, cert, dbCluster.IdentityType(api.IdentityTypeCertificateClient))
```
markylaing: Is the explicit cast required here or can it be coerced?

markylaing: I also wonder if we shouldn't just keep this within the function rather than passing the type that the identity will be changed to.

E.g. we currently only have `Pending TLS -> Certificate client`, and after adding cluster links we'll have `Pending cluster link -> cluster link` so maybe define a method like:
```go
func (i IdentityType) Activate() (IdentityType, error) {
    switch i {
    case api.IdentityTypeCertificateClientPending:
        return api.IdentityTypeCertificateClient, nil
    case api.IdentityTypeClusterLinkPending:
        return api.IdentityTypeClusterLink, nil
    default:
        return "", fmt.Errorf("Identities of type %s cannot be activated", i)
    }
}
```
and call that within this function.

lxd/db/cluster/identities.go
```diff
@@ -447,7 +448,7 @@ func GetPendingTLSIdentityByTokenSecret(ctx context.Context, tx *sql.Tx, secret
 	// Convert identity types to strings for the IN clause.
 	typeStrings := make([]string, len(identityTypes))
 	for i, idType := range identityTypes {
-		typeStrings[i] = fmt.Sprintf("%d", idType)
+		typeStrings[i] = strconv.FormatInt(idType, 10)
```
markylaing: This probably isn't needed any more (see earlier comment). In any case, generally when I'm adding new code I'll fix lint errors in the addition rather than adding code and fixing linter errors in a separate commit.

lxd/identities.go
```diff
@@ -408,35 +408,35 @@ func createIdentityTLSTrusted(ctx context.Context, s *state.State, networkCert *
 	return response.SyncResponseLocation(true, nil, lc.Source)
 }
 
-func createIdentityTLSPending(ctx context.Context, s *state.State, req api.IdentitiesTLSPost, notify identityNotificationFunc) response.Response {
+func createCertificateAddToken(s *state.State, clientName string, identityType string) (*api.CertificateAddToken, error) {
```
markylaing, line 411: Can this function also be used in `POST /1.0/certificates`?

lxd/identities.go
tomponline: i like this idea

lxd/db/cluster/identities.go
tomponline: agreed

lxd/db/cluster/identities.go
tomponline: > In any case, generally when I'm adding new code I'll fix lint errors in the addition rather than adding code and fixing linter errors in a separate commit.

agreed

lxd/db/cluster/identities.go
kadinsayani: I included a variadic param for identity type to make `GetPendingTLSIdentityByTokenSecret` more general for future extensibility. We can just update the query for now, I'll remove those changes from this PR and include the query update with #14884.

kadinsayani: Since we only use `getPendingTLSIdentityByTokenSecretStmt` in `GetPendingTLSIdentityByTokenSecret`, I'll move it into the function.

lxd/identities.go
kadinsayani: You're right, the explicit cast shouldn't be required since `driver.Value` is implemented for `IdentityType`. I like that idea as well!

lxd/identities.go
kadinsayani: Yes, nice catch!

lxd/db/cluster/identities.go
```diff
@@ -193,6 +193,16 @@ func (i IdentityType) Value() (driver.Value, error) {
 	return nil, fmt.Errorf("Invalid identity type %q", i)
 }
 
+// Activate returns the active version of the identity type.
+func (i IdentityType) Activate() (IdentityType, error) {
```
tomponline: ActiveType() rather than Activate() - as the latter makes it sound like its mutating the DB?

lxd/identities.go
```diff
@@ -445,15 +445,37 @@ func createIdentityTLSPending(ctx context.Context, s *state.State, req api.Ident
...
+	}
+
+	return &token, nil
+}
+
+func createIdentityTLSPending(ctx context.Context, s *state.State, req api.IdentitiesTLSPost, notify identityNotificationFunc) response.Response {
+	// Create CertificateAddToken token.
+	token, err := createCertificateAddToken(s, req.Name, api.IdentityTypeCertificateClient)
+	if err != nil {
+		return response.InternalError(fmt.Errorf("Failed to create certificate add token: %w", err))
```
tomponline: Need to use response.SmartError here otherwise you lose the benefit of returning StatusBadRequest here

https://github.com/canonical/lxd/pull/15838/commits/ef03498e5e9cf570712684f07c892646a9e966f5#diff-db86f7404a81beaeda08c11b8309cd12b260cdf3e6b5912797fdc200d42787b6R416

Because it will just get masked over as a 500 error rather than a 400 error.

tomponline: There's rarely a good reason to use something other than response.SmartError when wrapping an error variable.

lxd/certificates.go
tomponline, line 1: Nice de-dupe

lxd/identities.go
kadinsayani: ack, thanks

lxd/db/cluster/identities.go
kadinsayani: I like that name better too, thanks for the suggestion :)

kadinsayani:
Thanks for the reviews @markylaing and @tomponline!

lxd/db/cluster/identities.go
```diff
@@ -428,16 +428,16 @@ func ActivateTLSIdentity(ctx context.Context, tx *sql.Tx, identifier uuid.UUID,
...
-var getPendingTLSIdentityByTokenSecretStmt = fmt.Sprintf(`
+// GetPendingTLSIdentityByTokenSecret gets a single identity of type identityTypeCertificateClientPending with the given
+// secret in its metadata. If no pending identity is found, an api.StatusError is returned with http.StatusNotFound.
+func GetPendingTLSIdentityByTokenSecret(ctx context.Context, tx *sql.Tx, secret string) (*Identity, error) {
+	getPendingTLSIdentityByTokenSecretStmt := fmt.Sprintf(`
 SELECT identities.id, identities.auth_method, identities.type, identities.identifier, identities.name, identities.metadata
 	FROM identities
 	WHERE identities.type = %d
 	AND json_extract(identities.metadata, '$.secret') = ?
 `, identityTypeCertificateClientPending)
```
markylaing: I don't understand what we're gaining here. I guess the variable is moved to a lower scope, but the variable is also not exported. This also means that the `fmt.Sprintf` call is made every time the function is called, rather than once when the variable is instantiated at the package level (though I realise this is a micro-optimisation)

lxd/db/cluster/identities.go
```diff
@@ -404,8 +404,8 @@ func (i *Identity) ToAPI(ctx context.Context, tx *sql.Tx, canViewGroup auth.Perm
 }
 
 // ActivateTLSIdentity updates a TLS identity to make it valid by adding the fingerprint, PEM encoded certificate, and setting
-// the type to api.IdentityTypeCertificateClient.
-func ActivateTLSIdentity(ctx context.Context, tx *sql.Tx, identifier uuid.UUID, cert *x509.Certificate) error {
+// the type.
+func ActivateTLSIdentity(ctx context.Context, tx *sql.Tx, identifier uuid.UUID, cert *x509.Certificate, identityType IdentityType) error {
```
markylaing: The idea behind adding the `Activate` or `ActiveType` function here was so that the identity type doesn't need to be passed in.

Can we get the identity from the database by its identifier (not just it's ID) and then call `ActiveType` on the identity type that the returned identity gives us?

lxd/certificates.go
```diff
@@ -615,37 +615,19 @@ func certificatesPost(d *Daemon, r *http.Request) response.Response {
 			return response.BadRequest(fmt.Errorf("Invalid certificate material: %w", err))
 		}
 	} else if req.Token {
-		// Get all addresses the server is listening on. This is encoded in the certificate token,
-		// so that the client will not have to specify a server address. The client will iterate
-		// through all these addresses until it can connect to one of them.
-		addresses, err := util.ListenAddresses(localHTTPSAddress)
+		token, err := createCertificateAddToken(s, req.Name, api.IdentityTypeCertificateClient)
```
markylaing: I suspect here that we don't want to pass in this identity type, since we're not creating a fine-grained TLS identity via this API. 

I would pass an empty string instead, as this API never used that field.

markylaing (changes requested):
Just a couple more nits and then I think it's ready

lxd/db/cluster/identities.go
kadinsayani: We previously discussed this in our 1:1: https://github.com/canonical/lxd/pull/14884#discussion_r2017062518. I think this change makes more sense if we were going with a variadic parameter for identity types, which is what we initially discussed. Since we will just modify the query by adding an `IN` clause, I'll drop this change.

lxd/db/cluster/identities.go
kadinsayani: Gotcha, yes that works.

lxd/db/cluster/identities.go
markylaing: Great, thanks and sorry for the confusion

kadinsayani: No worries :)

tomponline (approved):
LGTM thanks!

tomponline:
@markylaing happy with this?

PR #15840, opened 2025-06-20
lxd/instances_post: Use correct project on cross-project copy (from Incus)

kadinsayani:
Cherry-picked from https://github.com/lxc/incus/pull/503.


tomponline (approved):
Thanks @kadinsayani 

Please can we have a test added for this.

Suggest in `clustering_move.sh`.

Thanks

PR #15828, opened 2025-06-18
doc: add sphinx-sitemap to generate sitemap for SEO

minaelee:
Adds `sphinx-sitemap` extension to generate `sitemap.xml` for SEO purposes. This is in response to a DOCPR-wide request that applies to all public-facing RTD documentation sets.

PR #15843, opened 2025-06-20
doc: rename variable used by sitemap per conflict

minaelee:
In custom_conf.py, this renames the version variable (added in https://github.com/canonical/lxd/pull/15828 while adding sitemap functionality) to rtd_version. This removes a conflict with a version variable that's used when creating Sphinx's inventory of files in objects.inv.

Also updates tests.yml to use a `fetch-depth=0` for documentation job as sphinx-sitemap otherwise produces 'git clone too shallow' errors.


tomponline (approved):
Ta

PR #15859, opened 2025-06-24
Devices: Add common filter capabilities

roosterfish:
Based on a comment in https://github.com/canonical/lxd/pull/15523#discussion_r2156631992 to streamline the device filtering.

The PR adds a central Filter function type which can be used to filter for arbitrary devices using pre-defined filter functions from a new `device/filter/` package.


lxd/device/filters/disk.go
```diff
@@ -0,0 +1,36 @@
+package filters
+
+// IsDisk evaluates whether or not the given device is of type disk.
+func IsDisk(device map[string]string) bool {
+	return device["type"] == "disk"
+}
+
+// IsRootDisk evaluates whether or not the given device is a root volume.
+func IsRootDisk(device map[string]string) bool {
+	return IsDisk(device) && HasRootPath(device) && !HasSource(device)
```
roosterfish: I have just copied the condition from the already existing code but I think a root disk also has the `pool` field set in all cases so we can check for it too.

lxd/device/filters/disk.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// IsCustomVolumeDisk evaluates whether or not the given device is a custom volume.
+func IsCustomVolumeDisk(device map[string]string) bool {
+	return IsDisk(device) && !HasRootPath(device) && HasPool(device)
```
roosterfish: I have just copied the condition from the already existing code but I think a custom volume disk also has the `source` field set in all cases so we can check for it too.

lxd/device/filters/disk.go
roosterfish: There is one exception, see https://github.com/canonical/lxd/pull/15859/commits/28751be995578a38b124f7ed5997a1c191ff0035.

tomponline: I like it when we write to our future selves :)

lxd/device/filters/disk.go
roosterfish: Fixed.

lxd/device/config/devices.go
```diff
@@ -242,3 +242,26 @@ func (list Devices) Reversed() DevicesSortable {
...
+// Combining filters using the OR operator can be done using the filters.Or function.
+func (list Devices) Filter(filters ...func(map[string]string) bool) Devices {
+	filteredDevices := Devices{}
+
+	for deviceName, device := range list {
+		allFiltersPassed := true
+
+		for _, filter := range filters {
+			if !filter(device) {
+				allFiltersPassed = false
```
tomponline, line 259: can we `break` here?

lxd/instance/drivers/driver_lxc.go
```diff
@@ -8536,11 +8533,7 @@ func (d *lxc) getFSStats() (*metrics.MetricSet, error) {
 	}
 
 	// Get disk devices
-	for _, dev := range d.expandedDevices {
-		if dev["type"] != "disk" || dev["path"] == "" {
-			continue
-		}
-
+	for _, dev := range d.expandedDevices.Filter(filters.IsDisk, filters.HasPath) {
```
tomponline: So this sort of construct isn't really delivering any value in my mind, because its just replicating the logic we already had, just in a more abstracted form.

What I was hoping for was more of a set of higher *definition of concepts* filters.

So rather than filtering for "device type disk with non-empty path", we could instead ask "why are we asking for devices with those properties - what does it represent to us?'

Using that mindset, and from just a cursory glance, this function is getting metrics for *filesystem disks* (and so it is using a non-empty `path` as a proxy for that meaning).

So how about `d.expandedDevices.Filter(filters.FilesystemDisk)`?

tomponline: Ultimately what I was hoping to achieve with this work is to avoid needing to duplicate the definition of each logical concept multiple times.

tomponline: So `IsRootDisk` is great. And general device type filters like `IsDisk` and `IsNIC` are useful in that they allow using the filtering method consistently, but be cautious of not re-inventing the logic we already have in multiple places.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -6905,16 +6902,8 @@ func (d *qemu) migrateSendLive(pool storagePools.Pool, clusterMoveSourceName str
 
 	// Notify the shared disks that they're going to be accessed from another system.
 	diskPools := make(map[string]storagePools.Pool, len(d.expandedDevices))
-	for _, dev := range d.expandedDevices.Sorted() {
-		if dev.Config["type"] != "disk" || dev.Config["path"] == "/" {
-			continue
-		}
-
+	for _, dev := range d.expandedDevices.Filter(filters.IsCustomVolumeDisk).Sorted() {
```
tomponline, line 6905: This is exactly the sort of conceptual filter I was looking for, thanks.

lxd/device/filters/disk.go
```diff
@@ -6,6 +6,10 @@ func IsDisk(device map[string]string) bool {
 }
 
 // IsRootDisk evaluates whether or not the given device is a root volume.
+// Root disk devices also need a non-empty "pool" property, but we can't check that here
+// because this function is used with clients talking to older servers where there was no
+// concept of a storage pool, and also it is used for migrating from old to new servers.
+// The validation of the non-empty "pool" property is done inside the disk device itself.
 func IsRootDisk(device map[string]string) bool {
 	return IsDisk(device) && HasRootPath(device) && !HasSource(device)
```
tomponline: We probably dont need HasRootPath and HasSource functions, they feel too specific without giving much meaning beyond the original hand-crafted filter.

tomponline (changes requested):
Overall I really like this, just a few comments around what conceptual level we should define the filters.

lxd/device/config/devices.go
roosterfish: thanks, fixed.

lxd/instance/drivers/driver_lxc.go
roosterfish: This is a good point. As you pointed out I just tried to model some of the conditions with "simple" filters. But after looking at those lines more carefully I found most of the times only the most relevant attributes where checked, e.g. does the device have a `path` so it can be mounted and analyzed whilst ignoring other required attributes like `source`, `pool` etc.

I have now put a few more high level filters in place which implicitly validate also the other attributes of a disk device but provide a human readable name so it makes it easy to filter for a specific set of disks using a single filter. 
Whilst doing so I was able to reuse the filtering logic too which now makes it feel quite solid. 

The simple `IsDisk` and `IsNIC` filters are kept as they are used various times and I think those make sense as they really filter disks out of the pool of available devices.

Overall there are many places in the code base which could now use this filtering pattern (also other devices than `type=disk`) so I would suggest doing it iteratively whenever touching/filtering devices. 
In this PR I tried to cover all places in the code base that try to identify root disks so that this is consistently using the new filters.

lxd/device/filters/disk.go
roosterfish: Fixed by deleting the simple filters.

lxd/device/filters/disk.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// IsCustomVolumeDiskFS evaluates whether or not the given device is a custom volume of type filesystem.
+func IsCustomVolumeDiskFS(device map[string]string) bool {
+	return IsDisk(device) && device["path"] != "/" && device["path"] != "" && device["source"] != "" && device["pool"] != ""
+}
+
+// IsCustomVolumeShare evaluates whether or not the given device is a directory share.
+func IsCustomVolumeShare(device map[string]string) bool {
+	return IsDisk(device) && device["path"] != "/" && device["source"] != "" && device["pool"] == ""
```
tomponline: Shouldnt this also check for a non empty path?

tomponline: "IsDirectoryShare" ? As this isnt always related to a custom volume, might be a host directory share.

lxd/instance/drivers/driver_lxc.go
```diff
@@ -8536,11 +8533,7 @@ func (d *lxc) getFSStats() (*metrics.MetricSet, error) {
 	}
 
 	// Get disk devices
-	for _, dev := range d.expandedDevices {
-		if dev["type"] != "disk" || dev["path"] == "" {
-			continue
-		}
-
+	for _, dev := range d.expandedDevices.Filter(filters.IsDiskMountable) {
```
tomponline: How about IsFilesystemDisk?

tomponline: Also this .IsDiskMountable seems to include vm root disks too which isnt right?

roosterfish: I like `IsFilesystemDisk` as it doesn't imply something gets mounted. Would you be fine if this filter also evaluates to `true` for VM root disks? 
Based on the device's attributes we cannot differentiate if it is a container or VM root vol as both just have `path=/`.

lxd/device/filters/disk.go
roosterfish: Added the check for non empty path and moved to `IsDirectoryShare`. 

I was just thinking about attaching vols but as you add shares through the `lxc config device` cmd there is no point in tracking that under the "custom volume" realm. Fixed.

lxd/device/filters/disk.go
```diff
@@ -0,0 +1,39 @@
...
+	return IsDisk(device) && device["path"] == "" && device["source"] != "" && device["pool"] != ""
+}
+
+// IsCustomVolumeFilesystemDisk evaluates whether or not the given device is a custom volume of type filesystem.
+func IsCustomVolumeFilesystemDisk(device map[string]string) bool {
+	return IsDisk(device) && device["path"] != "/" && device["path"] != "" && device["source"] != "" && device["pool"] != ""
+}
+
+// IsDirectoryShare evaluates whether or not the given device is a directory share.
+func IsDirectoryShare(device map[string]string) bool {
```
tomponline: This might not be a directory, for containers you can share single files through, so how about:

```
IsHostFilesystemShareDisk
```

lxd/device/filters/disk.go
```diff
@@ -0,0 +1,39 @@
...
+
+// IsDisk evaluates whether or not the given device is of type disk.
+func IsDisk(device map[string]string) bool {
+	return device["type"] == "disk"
+}
+
+// IsFilesystemDisk evaluates whether or not the given device contains a filesystem.
+// A filesystem disk device always holds a target path.
+// It returns true for custom volumes of type filesystem, root volumes and directory shares.
+func IsFilesystemDisk(device map[string]string) bool {
```
tomponline, line 13: The usage of this is to replace:

```go
if dev["type"] != "disk" || dev["path"] == "" {
     continue
}
```

VM root disks are an exception to this because they are not mountable, nor are they filesystem volumes, and yet they have `path=/` property in their associated disk device.

Its probably OK for now, but we'll see if we run into issues with this down the line as we switch over to it.

lxd/device/filters/disk.go
roosterfish: Missed this, thanks for the hint, fixed :)

lxd/device/filters/disk.go
roosterfish: I have added a comment to the `IsFilesystemDisk` filter to make this clear. Fixed.

tomponline (approved):
Great thanks, I like this pattern, makes the code more understandable.

PR #15862, opened 2025-06-24
github: the CLA check is now an Org-wide check

simondeziel:


tomponline:
@simondeziel we need to hold off on this until we're notified its ready.

> they are experimenting with org-wide cla check required rules

simondeziel:
> @simondeziel we need to hold off on this until we're notified its ready.
> 
> > they are experimenting with org-wide cla check required rules

I saw it ran successfully and concluded it was good already. Should I close for now?

tomponline:
> > @simondeziel we need to hold off on this until we're notified its ready.
> > > they are experimenting with org-wide cla check required rules
> 
> I saw it ran successfully and concluded it was good already. Should I close for now?

yes please

PR #15865, opened 2025-06-24
Bump Go min to 1.24.4 and update Go modules (stable-5.0)

simondeziel:


simondeziel:
Will come back at it later.

Issue #11112, opened 2022-11-12
Copy from volume snapshot to remote fails: "Error: not found"


bazaah:
# Required information

 * Distribution: `Arch Linux`
 * Distribution version: -
 * The output of "lxc info" or if that fails:

<details>
<summary>lxc info</summary>

```
config: <snip>
api_extensions:
- storage_zfs_remove_snapshots
- container_host_shutdown_timeout
- container_stop_priority
- container_syscall_filtering
- auth_pki
- container_last_used_at
- etag
- patch
- usb_devices
- https_allowed_credentials
- image_compression_algorithm
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
- unix_device_rename
- storage_lvm_use_thinpool
- storage_rsync_bwlimit
- network_vxlan_interface
- storage_btrfs_mount_options
- entity_description
- image_force_refresh
- storage_lvm_lv_resizing
- id_map_base
- file_symlinks
- container_push_target
- network_vlan_physical
- storage_images_delete
- container_edit_metadata
- container_snapshot_stateful_migration
- storage_driver_ceph
- storage_ceph_user_name
- resource_limits
- storage_volatile_initial_source
- storage_ceph_force_osd_reuse
- storage_block_filesystem_btrfs
- resources
- kernel_limits
- storage_api_volume_rename
- macaroon_authentication
- network_sriov
- console
- restrict_devlxd
- migration_pre_copy
- infiniband
- maas_network
- devlxd_events
- proxy
- network_dhcp_gateway
- file_get_symlink
- network_leases
- unix_device_hotplug
- storage_api_local_volume_handling
- operation_description
- clustering
- event_lifecycle
- storage_api_remote_volume_handling
- nvidia_runtime
- container_mount_propagation
- container_backup
- devlxd_images
- container_local_cross_pool_handling
- proxy_unix
- proxy_udp
- clustering_join
- proxy_tcp_udp_multi_port_handling
- network_state
[...]

PR #15868, opened 2025-06-24
lxd: Only run `PostMigrateSend` when migrating instances

kadinsayani:
Fixes a regression introduced by https://github.com/canonical/lxd/pull/15552. We only want to run `PostMigrateSend` on instance devices.

### DIFF
--- a/lxd/migrate_instance.go
+++ b/lxd/migrate_instance.go
@@ -156,10 +156,12 @@ func (s *migrationSourceWs) Do(state *state.State, migrateOp *operations.Operati
 		return fmt.Errorf("Failed migration on source: %w", err)
 	}
 
-	err = s.instance.PostMigrateSend()
-	if err != nil {
-		l.Error("Failed post-migration steps on source", logger.Ctx{"err": err})
-		return fmt.Errorf("Failed post-migration steps on source: %w", err)
+	if !shared.IsSnapshot(s.instance.Name()) {
+		err = s.instance.PostMigrateSend()
+		if err != nil {
+			l.Error("Failed post-migration steps on source", logger.Ctx{"err": err})
+			return fmt.Errorf("Failed post-migration steps on source: %w", err)
+		}
 	}
 
 	return nil


tomponline (approved):
Thanks!

tomponline:
@kadinsayani possible to have some tests in main test suite for this?

kadinsayani:
> @kadinsayani possible to have some tests in main test suite for this?

general vm tests for migration and post-migration steps, or testing a snapshot migration and making sure post-migration steps don't happen?

tomponline:
> > @kadinsayani possible to have some tests in main test suite for this?
> 
> general vm tests for the post-migration steps or testing a snapshot migration and making sure post-migration steps don't happen?

https://github.com/canonical/lxd/pull/15552 was not VM specific, so we should have a test in the main test suite that catches this regression for the future.

kadinsayani:
> > > @kadinsayani possible to have some tests in main test suite for this?
> > 
> > 
> > general vm tests for the post-migration steps or testing a snapshot migration and making sure post-migration steps don't happen?
> 
> #15552 was not VM specific, so we should have a test in the main test suite that catches this regression for the future.

https://github.com/canonical/lxd/pull/15870

tomponline:
> > > > @kadinsayani possible to have some tests in main test suite for this?
> > > 
> > > 
> > > general vm tests for the post-migration steps or testing a snapshot migration and making sure post-migration steps don't happen?
> > 
> > 
> > #15552 was not VM specific, so we should have a test in the main test suite that catches this regression for the future.
> 
> #15870

@kadinsayani I think we have misinterpreted what that test failure earlier was telling us, and I think the fix in https://github.com/canonical/lxd/pull/15868 needs to be reverted and a different fix applied.

Lets remind ourselves of the original problem description from the issue https://github.com/canonical/lxd/issues/15511:

> dnsmasq.lease not being deleted from the server when the container is moved in a cluster

You approached this by introducing a `PostMigrateSend` hook for devices, which is called each time an instance migration has finished.

But "migration" here means a specific thing, it means sending the instance over a byte stream to receiver (usually a different host - and not specifically a host in the cluster). 

It does *not* *only* mean "move an instance to a different cluster member".

And we see this in the test failure https://github.com/canonical/lxd-ci/actions/runs/15843329391/job/44659969441#step:12:901 as the command failing is a #11112 

> lxc copy v1/snap0 localhost:v2

So this means that your post migration send hook is now removing leases on all migrations, not just cluster member moves.

This is incorrect.

**We only want the `bridged` NIC lease to be cleared on the source member if the operation is a cluster member move.**

A cluster member move is defined by the `clusterMoveSourceName` argument passed to `newMigrationSource()`, and we can see that `newMigrationSource` is called in several places, and only in `instancePostClusteringMigrate()` is that argument passed as non-empty.
 
The question then becomes: Do we want the post migration hook to only fire for cluster member moves? In which case the hook should be renamed to `PostClusterMemberMoveSend()` or should we update the definition of the `PostMigrateSend()` to accept an argument that indicates if the migration send was part of a cluster member move (perhaps by passing in `clusterMoveSourceName`) and then update the bridged NIC hook implementation to only release the lease when it is a cluster member move.

Finally, the comment change in https://github.com/canonical/lxd/pull/15552/commits/29bc1433c6a187c3587499f4f0b46ca23be2e466 should be further tweaked to clarify that this function deals with moving an instance between cluster members (maybe even re-name the function too).

The original comment:

> // Move a non-ceph instance to another cluster node. Source and target members must be online.

was good, except for the mention of ceph and "node" rather than "member". The switch way from "move" to "migrate" was misleading it turns out.

PR #15869, opened 2025-06-24
test: Add test for container migration with attached local volume

nmezhenskyi:
Follow-up to https://github.com/canonical/lxd/pull/15798.

Adds a test for container migration (`lxc copy` and `lxc move`) with attached local volume.

### DIFF
--- a/test/suites/migration.sh
+++ b/test/suites/migration.sh
@@ -511,6 +511,42 @@ migration() {
   lxc_remote storage volume delete l2:"$remote_pool2" bar
   lxc_remote storage unset l1:"$remote_pool1" rsync.compression
 
+  echo "==> Test container migration with attached local volumes."
+
+  echo "==> Create a local storage with the same name."
+  lxc_remote storage create l1:dir dir
+  lxc_remote storage create l2:dir dir
+
+  echo "==> Create a volume to attach to container."
+  lxc_remote storage volume create l1:dir vol1 size=4MiB
+
+  echo "==> Create a container to test migration with attached local volume."
+  lxc_remote init --empty l1:c1
+  lxc_remote storage volume attach l1:dir vol1 c1 /files
+
+  echo "==> Check that copying a container with attached local volume fails, if the destination does not have a volume with the same name."
+  ! lxc_remote copy l1:c1 l2: || false
+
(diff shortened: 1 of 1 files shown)


test/suites/migration.sh
```diff
@@ -511,6 +511,24 @@ migration() {
   lxc_remote storage volume delete l2:"$remote_pool2" bar
   lxc_remote storage unset l1:"$remote_pool1" rsync.compression
 
+  echo "==> Test container migration with attached local volumes."
+
+  echo "==> Create a volume to attach to container."
+  lxc_remote storage volume create l1:"${remote_pool1}" vol1 size=4MiB
+
+  echo "==> Create a container to test migration with attached local volume."
+  lxc_remote init testimage l1:c1
```
simondeziel: Since the container is never started, could we get away with `--empty`?

test/suites/migration.sh
```diff
@@ -511,6 +511,24 @@ migration() {
...
+  lxc_remote storage volume attach l1:"${remote_pool1}" vol1 c1 /files
+
+  echo "==> Check that copying a container with attached local volume fails, if the destination does not have a volume with the same name."
+  ! lxc_remote copy l1:c1 l2: || false
+
+  echo "==> Copy the volume."
+  lxc_remote storage volume copy l1:"${remote_pool1}/vol1" l2:"${remote_pool2}/vol2"
+
+  echo "==> Check that copying a container with attached local volume succeeds, if the destination has a volume with the same name."
+  lxc_remote copy l1:c1 l2:
```
simondeziel, line 537: I think, the original and the copied containers and volumes need cleanup.

nmezhenskyi: Yes, sorry it's still WIP

tomponline (commented):
@nmezhenskyi do you think we could try adding VM tests for this (as we dont need to start the VMs), similar to https://github.com/canonical/lxd/blob/main/test/suites/vm.sh utilising `LXD_VM_TESTS` to skip in non-VM environments.

nmezhenskyi:
@tomponline 
> do you think we could try adding VM tests for this?

Hm, I think so, we should be able to, I'll get on it

PR #15870, opened 2025-06-24
tests: Test snapshot can be copied with volatile MAC conflict

kadinsayani:
Additional test for `PostMigrateSend` introduced with https://github.com/canonical/lxd/pull/15552.


test/suites/container_devices_nic_bridged.sh
```diff
@@ -822,11 +822,16 @@ test_container_devices_nic_bridged() {
...
+  lxc snapshot foo snap0
 
   # Test container start will fail due to volatile MAC conflict.
   [ "$(lxc config get foo volatile.eth0.hwaddr)" = "$(lxc config get foo2 volatile.eth0.hwaddr)" ]
   ! lxc start foo2 || false
-  lxc delete -f foo foo2
+  lxc delete -f foo2
+
+  # Test snapshot can be copied.
+  lxc copy foo/snap0 foo3
```
tomponline: This doesn't seem to be testing the regression in https://github.com/canonical/lxd-ci/actions/runs/15843329391/job/44659969441#step:12:901 as that was testing the migration subsystem by copying to a "localhost" remote.

Can you try reverting https://github.com/canonical/lxd/pull/15868 locally and then try this test change and check it fails.

kadinsayani: I've updated the test to copy to a `localhost` remote and can confirm the test fails after reverting the fix.

tomponline (approved):
Thanks!

PR #15873, opened 2025-06-24
Cluster: Rework post-migration steps to cleanup NIC bridged devices for cluster member moves only

kadinsayani:
Closes https://github.com/canonical/lxd/issues/15511.

### DIFF
--- a/lxd/instance/drivers/driver_common.go
+++ b/lxd/instance/drivers/driver_common.go
@@ -510,6 +510,26 @@ func (d *common) deviceVolatileSetFunc(devName string) func(save map[string]stri
 	}
 }
 
+// postMigrateSendCommon handles common instance post-migration steps.
+func (d *common) postMigrateSendCommon(inst instance.Instance, clusterMoveSourceName string) error {
+	// Perform post-migration device cleanup.
+	for devName, devConfig := range d.ExpandedDevices() {
+		dev, err := d.deviceLoad(inst, devName, devConfig)
+		if err != nil {
+			logger.Error("Failed to load device %q during post-migration steps on source: %v", logger.Ctx{"devName": devName, "err": err})
+		}
+
+		if dev != nil {
+			err = dev.PostMigrateSend(clusterMoveSourceName)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
 // expandConfig applies the config of each profile in order, followed by the local config.
 func (d *common) expandConfig() error {
 	var globalConfigDump map[string]any
--- a/lxd/instance/drivers/driver_common.go
+++ b/lxd/instance/drivers/driver_common.go
@@ -510,24 +510,6 @@ func (d *common) deviceVolatileSetFunc(devName string) func(save map[string]stri
 	}
 }
 
-// postMigrateSendCommon handles the common part of instance post-migration send.
-func (d *common) postMigrateSendCommon(inst instance.Instance) error {
-	// Perform post-migration device cleanup.
-	for devName, devConfig := range d.ExpandedDevices() {
-		dev, err := d.deviceLoad(inst, devName, devConfig)
-		if err != nil {
-			return err
-		}
-
-		err = dev.PostMigrateSend()
-		if err != nil {
-			return err
-		}
-	}
-
-	return nil
-}
-
 // expandConfig applies the config of each profile in order, followed by the local config.
 func (d *common) expandConfig() error {
 	var globalConfigDump map[string]any
--- a/lxd/instance/drivers/driver_lxc.go
+++ b/lxd/instance/drivers/driver_lxc.go
@@ -5273,7 +5273,7 @@ fi
 }
 
 // MigrateSend controls the sending side of a migration.
-func (d *lxc) MigrateSend(args instance.MigrateSendArgs) error {
+func (d *lxc) MigrateSend(args instance.MigrateSendArgs) (err error) {
 	d.logger.Info("Migration send starting")
 	defer d.logger.Info("Migration send stopped")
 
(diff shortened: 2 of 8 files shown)


lxd/instance/drivers/driver_lxc.go
```diff
@@ -5273,9 +5273,18 @@ fi
 }
 
 // MigrateSend controls the sending side of a migration.
-func (d *lxc) MigrateSend(args instance.MigrateSendArgs) error {
+func (d *lxc) MigrateSend(args instance.MigrateSendArgs) (err error) {
 	d.logger.Info("Migration send starting")
-	defer d.logger.Info("Migration send stopped")
+	defer func() {
+		d.logger.Info("Migration send stopped")
```
tomponline: hrm, shouldnt we only be calling d.postMigrateSendCommon if the migration send completed successfully, can this be called at the bottom of the function before we return err?

tomponline (commented):
Did you see this fix the test regression, did it not need overcoming the validation problem wrt to duplicate mac addresses when doing `deviceLoad()`?

kadinsayani:
> Did you see this fix the test regression, did it not need overcoming the validation problem wrt to duplicate mac addresses when doing `deviceLoad()`?

Yes, it fixes the regression. The `Failed remove validation for device` error is logged, not returned. Also, in `postMigrateSendCommon`, `PostMigrateSend` is called on `dev` even if `deviceLoad` returns an `err`.

lxd/instance/drivers/driver_lxc.go
kadinsayani: Yeah, we can call `postMigrateSendCommon` at the end. I was unsure about doing this in `driver_lxc.go` because we wait for goroutines to finish and return the first error (if any). I think I've worked around this by capturing the error returned by `postMigrateSendCommon` separately, and returning that error before the `err` from the goroutines.

lxd/instance/drivers/driver_common.go
```diff
@@ -510,6 +510,24 @@ func (d *common) deviceVolatileSetFunc(devName string) func(save map[string]stri
 	}
 }
 
+// postMigrateSendCommon handles common instance post-migration steps.
+func (d *common) postMigrateSendCommon(inst instance.Instance, clusterMoveSourceName string) error {
+	// Perform post-migration device cleanup.
+	for devName, devConfig := range d.ExpandedDevices() {
+		dev, err := d.deviceLoad(inst, devName, devConfig)
```
tomponline, line 517: @kadinsayani I'd like us to explore why nicBridged.checkAddressConflict() is failing during a cluster member move and see if we can avoid triggering this error message. 

There's a section in `checkAddressConflict`:

```go
		// Skip our own device. This avoids triggering duplicate device errors during
		// updates or when making temporary copies of our instance during migrations.
		sameLogicalInstance := instance.IsSameLogicalInstance(d.inst, &inst)
		if sameLogicalInstance && d.Name() == nicName {
			return nil
		}
```

Which I would have thought would have held true here.

lxd/instance/drivers/driver_lxc.go
```diff
@@ -5700,6 +5700,12 @@ func (d *lxc) MigrateSend(args instance.MigrateSendArgs) error {
 			}
 		}
 
+		postMigrateSendErr := d.postMigrateSendCommon(d, args.ClusterMoveSourceName)
```
tomponline: I think we should only run postMigrateSendCommin if the `err` from `g.Wait()` is non-nil indicating the migration send completed successfully.

lxd/instance/drivers/driver_common.go
kadinsayani: I don't see it failing for cluster member moves. That check holds true.

lxd/instance/drivers/driver_lxc.go
kadinsayani: Sounds good!

lxd/instance/drivers/driver_common.go
kadinsayani: I believe we saw that error for a snapshot copy where `volatile.hwaddr` was filled as a result of `PostMigrateSend` being called:

https://github.com/canonical/lxd/blob/e34037843083355703233af0da01dd9b04cb84ab/lxd/device/nic_bridged.go#L879-L891

Resulting failed check:

https://github.com/canonical/lxd/blob/e34037843083355703233af0da01dd9b04cb84ab/lxd/device/nic_bridged.go#L417-L425

tomponline: Ah ok thanks

lxd/instance/drivers/driver_common.go
```diff
@@ -510,6 +510,24 @@ func (d *common) deviceVolatileSetFunc(devName string) func(save map[string]stri
...
+		if err != nil {
+			logger.Error("Failed to load device %q during post-migration steps on source: %v", logger.Ctx{"devName": devName, "err": err})
+		}
+
+		err = dev.PostMigrateSend(clusterMoveSourceName)
```
tomponline: You need to check dev isnt nil

lxd/instance/drivers/driver_lxc.go
```diff
@@ -5700,6 +5700,14 @@ func (d *lxc) MigrateSend(args instance.MigrateSendArgs) error {
 			}
 		}
 
+		if err == nil {
+			postMigrateSendErr := d.postMigrateSendCommon(d, args.ClusterMoveSourceName)
+			if postMigrateSendErr != nil {
+				d.logger.Error("Post-migration steps failed on source", logger.Ctx{"postMigrateSendErr": postMigrateSendErr})
```
tomponline: Please use "err" label in the context here for consistency

lxd/instance/drivers/driver_lxc.go
```diff
@@ -5700,6 +5700,14 @@ func (d *lxc) MigrateSend(args instance.MigrateSendArgs) error {
 			}
 		}
 
+		if err == nil {
+			postMigrateSendErr := d.postMigrateSendCommon(d, args.ClusterMoveSourceName)
+			if postMigrateSendErr != nil {
+				d.logger.Error("Post-migration steps failed on source", logger.Ctx{"err": postMigrateSendErr})
+				return postMigrateSendErr
```
tomponline, line 5707: I'll merge this for now, but as the migration as fully succeeded at this point, I wonder if we should fail the operation here at all, or whether the logging alone will be sufficient (both both qemu and lxc drivers)?

kadinsayani: You raise a good point, I think logging is sufficient. 

https://github.com/canonical/lxd/pull/15886

PR #15879, opened 2025-06-25
Improve `snapshots.pattern` tests

simondeziel:


shared/util.go
```diff
@@ -1324,21 +1324,32 @@ func (r *ReadSeeker) Seek(offset int64, whence int) (int64, error) {
...
 // RenderTemplate renders a pongo2 template.
 func RenderTemplate(template string, ctx pongo2.Context) (string, error) {
+	// Create custom TemplateSet
+	set := pongo2.NewSet("restricted", pongo2.DefaultLoader)
+
+	// Ban tags that could be used to access the host's filesystem.
+	for _, tag := range []string{"extends", "import", "include", "ssi"} {
+		err := set.BanTag(tag)
+		if err != nil {
+			return "", fmt.Errorf("failed to ban tag %q: %w", tag, err)
```
tomponline: Errors should start with upper case letter please

shared/util.go
```diff
@@ -1324,21 +1324,32 @@ func (r *ReadSeeker) Seek(offset int64, whence int) (int64, error) {
...
 		return "", err
 	}
 
 	// Get rendered template
 	ret, err := tpl.Execute(ctx)
 	if err != nil {
-		return ret, err
-	}
+		// Looks like we're nesting templates so run pongo again
+		if strings.Contains(ret, "{{") || strings.Contains(ret, "{%") {
```
sudhackar, line 1348: why is this block moved to `err!=nil`? this will render `ret` as `""` and it would never be executed

simondeziel: Thanks, I'm reverting that bit in https://github.com/canonical/lxd/pull/15998 along with disabling template nesting.

PR #15880, opened 2025-06-26
API: Validate Content-Type if supplied, or if non-zero Content-Length supplied

tomponline:


lxd/daemon.go
```diff
@@ -929,6 +930,15 @@ func (d *Daemon) createCmd(restAPI *mux.Router, version string, c APIEndpoint) {
 				return response.NotImplemented(nil)
 			}
 
+			if len(action.ContentTypes) == 0 {
+				// Require application/json if not specified by handler.
+				action.ContentTypes = []string{"application/json"}
+			}
+
+			if r.Header.Get("Content-Lenth") != "" && !slices.Contains(action.ContentTypes, r.Header.Get("Content-Type")) {
```
markylaing: Typo in `Content-Length`, so I suspect this condition is not being hit.

tomponline: ugh, good spot, this is what comes of rushing stuff between meetings! Thanks

tomponline: I actually want to see a failure on this latest push as custom volume import of iso files should be tested and fail (as ive not put in allowed content types for that handler yet).

tomponline:
> LGTM but I think we need to update a handful of `curl` example we provide in the doc so they use `application/json` too.

Good point thanks.

lxd/daemon.go
```diff
@@ -929,6 +930,19 @@ func (d *Daemon) createCmd(restAPI *mux.Router, version string, c APIEndpoint) {
 				return response.NotImplemented(nil)
 			}
 
+			if len(action.ContentTypes) == 0 {
+				// Require application/json if not specified by handler.
+				action.ContentTypes = []string{"application/json"}
+			}
+
+			// Validate Content-Type if supplied, or if non-zero Content-Length supplied.
+			contentType, _, _ := strings.Cut(r.Header.Get("Content-Type"), ";") // Ignore multi-part boundary part.
```
simondeziel: Since it's [valid](https://www.rfc-editor.org/rfc/rfc7231#section-3.1.1.1) to have white spaces around `;`, maybe we should `TrimSpaces()`?

tomponline: fixed

tomponline:
> LGTM but I think we need to update a handful of `curl` example we provide in the doc so they use `application/json` too.

Done.

tomponline:
OK lets try this again @markylaing @simondeziel

test/suites/tls_restrictions.sh
```diff
@@ -423,15 +423,15 @@ test_certificate_edit() {
   # Trying to change other fields should fail as a non-admin.
   ! lxc_remote config trust show "${FINGERPRINT}" | sed -e "s/restricted: true/restricted: false/" | lxc_remote config trust edit localhost:"${FINGERPRINT}" || false
 
-  my_curl -X PATCH -d "{\"restricted\": false}" "https://${LXD_ADDR}/1.0/certificates/${FINGERPRINT}" | grep -F '"error_code":403'
+  [ "$(my_curl -X PATCH -H 'Content-Type: application/json' -d "{\"restricted\": false}" "https://${LXD_ADDR}/1.0/certificates/${FINGERPRINT}" | jq -r '.error_code')" -eq 403 ]
```
simondeziel, line 426: Good idea to avoid `grep` here.

Issue #15117, opened 2025-03-05
vm copies get duplicit IP addresses


skozina:
**Issue**
I've created a VM, and then made a copy of it. Both instances got the same IPv4 address.
**How to reproduce**
`$ lxc launch ubuntu:24.04 test-vm --vm`
... wait for test-vm to boot ...
`$ lxc copy test-vm test2-vm`
`$ lxc start test2-vm`
... wait for test2-vm to start ...

```
$ lxc list
+-----------+---------+------------------------+-------------------------------------------------+-----------------+-----------+
|   NAME    |  STATE  |          IPV4          |                      IPV6                       |      TYPE       | SNAPSHOTS |
+-----------+---------+------------------------+-------------------------------------------------+-----------------+-----------+
...
+-----------+---------+------------------------+-------------------------------------------------+-----------------+-----------+
| test2-vm  | RUNNING | 10.142.210.32 (enp5s0) | fd42:40e3:c742:ce37:216:3eff:fe89:2a20 (enp5s0) | VIRTUAL-MACHINE | 0         |
+-----------+---------+------------------------+-------------------------------------------------+-----------------+-----------+
| test-vm   | RUNNING | 10.142.210.32 (enp5s0) | fd42:40e3:c742:ce37:216:3eff:febe:9d17 (enp5s0) | VIRTUAL-MACHINE | 0         |
+-----------+---------+------------------------+-------------------------------------------------+-----------------+-----------+
```

Notice the duplicit IP 10.142.210.32
**Environment**
```
$ snap list lxd
Name  Version         Rev    Tracking     Publisher   Notes
lxd   5.21.3-75def3c  32455  5.21/stable  canonical✓  -
```
**Cause**
Both VMs are getting different uuids:
```
$ lxc config show test-vm | grep uuid
  volatile.uuid: c54d8dbe-eb11-45eb-85ee-0bfd95b3c461
  volatile.uuid.generation: c54d8dbe-eb11-45eb-85ee-0bfd95b3c461
stkozina@ersin-ubuntu:~$ lxc config show test2-vm | grep uuid
  volatile.uuid: 6475536f-2e60-43d1-90cf-7066300867d4
  volatile.uuid.generation: 6475536f-2e60-43d1-90cf-7066300867d4
```
However, dnsmasq on host is seeing identical client ids:
```
$ cat /var/snap/lxd/common/lxd/networks/lxdbr0/dnsmasq.leases
[...]

PR #15884, opened 2025-06-27
Fix duplicate IP assignment to VM copies

skozina:
Fixes https://github.com/canonical/lxd/issues/15117

### DIFF
--- a/lxd/network/driver_bridge.go
+++ b/lxd/network/driver_bridge.go
@@ -1044,6 +1044,276 @@ func (n *bridge) Start() error {
 	return nil
 }
 
+func (n *bridge) getDnsmasqArgs(bridge *ip.Bridge) ([]string, error) {
+	dnsmasqCmd := []string{"--keep-in-foreground", "--strict-order", "--bind-interfaces",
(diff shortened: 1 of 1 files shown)


lxd/network/driver_bridge.go
```diff
@@ -1044,6 +1044,276 @@ func (n *bridge) Start() error {
 	return nil
 }
 
+func getDnsmasqArgs(n *bridge, bridge *ip.Bridge) ([]string, error) {
```
tomponline: Can we move these new functions onto the `(n *bridge)` struct itself to keep them together in the bridge driver itself and save having `n` as an argument?

skozina: Sure, done.
I was thinking it would be nice to move these to the `dnsmasq` package. But then we'd need to create a new structure to pass the arguments through, as the functions need many details from the `bridge` struct and we can't get the `dnsmasq` package to import the `network` package;-/ So perhaps having these under the `bridge` struct is good for now.

tomponline (approved):
Thanks!

PR #15885, opened 2025-06-27
Storage: Reduce code duplication across remote storage drivers

mihalicyn:


lxd/storage/connectors/connector_nvme.go
```diff
@@ -220,3 +231,53 @@ func (c *connectorNVMe) findSession(targetQN string) (*session, error) {
 
 	return session, nil
 }
+
+// Discover returns the targets found on the first reachable targetAddr.
+func Discover(ctx context.Context, c Connector, serverUUID string, targetAddresses ...string) ([]nvmeDiscoveryLogRecord, error) {
+	if c.Type() != TypeNVME {
```
roosterfish, line 242: Why not adding the `Discover` func to the `Connector` interface directly alongside the `Connect` and `Disconnect`? This would save having to run this check. 

For all other connectors except NVMe the func can just return a not implemented error.

lxd/storage/connectors/connector_nvme.go
```diff
@@ -220,3 +231,53 @@ func (c *connectorNVMe) findSession(targetQN string) (*session, error) {
...
+			logger.Warn("Failed to find discovery log entries", logger.Ctx{"target_address": targetAddr, "err": err})
+			continue
+		}
+
+		// Try to unmarshal the returned log entries.
+		err = json.Unmarshal([]byte(stdout), &discoveryLog)
+		if err != nil {
+			// Don't just log this error.
+			// Something is clearly wrong with the returned output.
+			return nil, fmt.Errorf("Failed to unmarshal the returned discovery log entries from %q", targetAddr)
```
roosterfish: I see it's already missing in the original version in the PowerFlex driver, but should we also append the `err` at the end of the message?

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -753,8 +744,13 @@ func (d *powerflex) getHostGUID() (string, error) {
...
 func (d *powerflex) getNVMeTargetQN(targetAddresses ...string) (string, error) {
 	if d.nvmeTargetQN == "" {
+		connector, err := d.connector()
+		if err != nil {
+			return "", err
+		}
+
 		// The discovery log from the first reachable target address is returned.
-		discoveryLogRecords, err := d.discover(d.state.ShutdownCtx, targetAddresses...)
+		discoveryLogRecords, err := connectors.Discover(d.state.ShutdownCtx, connector, d.state.ServerUUID, targetAddresses...)
```
roosterfish: Having the `Discover` on the `Connector` interface would also save use from passing the UUID a second time as it's already passed when setting up the connector I believe.

lxd/storage/drivers/remote_storage_helpers.go
```diff
@@ -0,0 +1,159 @@
+package drivers
```
roosterfish: Should the file rather be called `remote_driver_helpers.go` to align with the interface name and the general idea of the `driver` package?

lxd/storage/drivers/remote_storage_helpers.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/canonical/lxd/shared/logger"
+	"github.com/canonical/lxd/shared/revert"
+	"golang.org/x/sys/unix"
+)
+
+// This file here is to consolidate all useful helpers to reuse across different remote storage
+// drivers which use NVMe/iSCSI/etc connectors like PowerFlex, Pure, HPE Alletra.
+
+// remoteDriver is the extended internal interface for use only with helpers below.
+type remoteDriver interface {
```
roosterfish: Why not just adding those funcs to the [`common`](https://github.com/canonical/lxd/blob/main/lxd/storage/drivers/driver_common.go#L28) driver implementation? It's already embedded by all the storage drivers.

Maybe I am missing something but I think the drivers can then still implement their own version of `unmapVolume` and `getMappedDevPath`.

roosterfish (changes requested):
Thanks for creating those generic funcs! 

I assume the HPE driver also needs to use discovery to retrieve the target NQN? Good to see that we don't use this pattern only for PowerFlex but also for HPE too in the future.

Left a few minor comments. I'll run some tests against the real PowerFlex cluster to validate the changes.

lxd/storage/connectors/connector_nvme.go
MusicDin: Very cool idea. And I assume the `Discover` will also be required for iSCSI in the future (for the same purpose).

lxd/storage/drivers/remote_storage_helpers.go
mihalicyn: My idea here was to explicitly separate all of these helpers as "for remote storages only" and keep all these in a single place.

So you think we should just add `unmapVolume` and `getMappedDevPath` to all storage drivers by default? Even not remote ones?

lxd/storage/drivers/remote_storage_helpers.go
mihalicyn: Will do, thanks!

lxd/storage/connectors/connector_nvme.go
mihalicyn: Sure! Will fix this. Thanks!

lxd/storage/drivers/remote_storage_helpers.go
roosterfish: Mhh I see. My thinking was to just have noop implementation on the `common` struct (returning not implemented like some of the others already do) and then it's up to the driver to decide implementing them.

roosterfish:
Update, just tested it on PowerFlex, still works and no regression when it comes to volume map/unmap as well as NVMe/TCP target discovery.

lxd/storage/drivers/utils_api.go
```diff
@@ -0,0 +1,20 @@
+package drivers
```
tomponline: No more "utils" files please.

https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

lxd/storage/drivers/remote_storage_helpers.go
mihalicyn: What I'm afraid here is that I'll add a lot of remote storage-only specific methods into `driver_common`. A benefit of having this in a separate structure is to make it easier to read the code and understand what has to be implemented for a new remote/local driver as these will live in a separate files, basically.

tomponline: I've not read through this entire PR yet, but taking a brief look my position on this would be:

1. For true "util" functions, that are divorced from any specific driver implementation and would be called from specific drivers themselves, purely to reduce code duplication, then it makes sense to me to put these into a separate package (or re-use an existing one, such as `lxd/storage/block`, `lxc/storage/filesystem` or `lxd/storage/connectors`. This means that they probably wont and shouldn't need to take a `remoteDriver` as an argument - but rather than take specific pieces of information that the remote driver can provide itself.
2. My (admittedly quick) observation here is that some of the functions at least are, what I would say, overstepping the role as a helper function and should be dialed back in their responsibility or moved into the `common` driver implementation. For example in `UnmountRemoteVolume` I don't see anything specific to "remote" pools in there, and that seems to be a good candidate for being added to the `common` driver. On the other hand `createBodyReader` seems like a good candidate to be moved into a "util" package (that isn't named "util" but named based on the subject of the collection of functions it containers).

I also see no need for a new interface, but I think this is just being done to avoid import loops, which is a sign that its likely in the wrong place too.

tomponline: So to summarise:

- Split up the remote specific code from the generic code
- Generic code that can be duplicated can go into the `common` driver function or if not wide enough in scope into a helper function into well-named sub-package - so that non-remote drivers can use it too.
- Remote specific code helpers should be factored out into a well-named package (or into an existing package).
- Avoid adding generic functionality into a remote specific helper function (like `UnmountRemoteVolume` is doing).

mihalicyn: >For example in UnmountRemoteVolume I don't see anything specific to "remote" pools in there, and that seems to be a good candidate for being added to the common driver.

What is specific is that these functions require a driver to implement `getMappedDevPath`/`mapVolume`/`unmapVolume` methods. Which makes absolutely no sense for a "common" driver. That's was the only reason why I decided to introduce a new interface, because it feels like a remote storage is a less abstract concept than a common driver with it's own specifics and I didn't want to bring that specifics into a common interface.

>I also see no need for a new interface, but I think this is just being done to avoid import loops, which is a sign that its likely in the wrong place too.

I hadn't seen any import loops as it's all anyways sits in the same package.

tomponline: Thanks, understood, OK so lets break down `MountRemoteVolume`:

The remote specific parts are:

1. getMappedDevPath - which gets calls to the appliance to map the volume to the host, and then block volume path on the host (I can't see any call to `mapVolume` in that function, presumably thats happening in the implementation of `getMappedDevPath`).

Now `UnmountRemoteVolume`:

1. getMappedDevPath - which it calls to get the mapped volume path (if exists).
2. unmapVolume - which it calls to deactive the volume on the host and remove the mapping on the appliance.

The rest of both of those functions is entirely generic.

Now I notice that in `MountRemoteVolume` the call to `getMappedDevPath` returns a `cleanup` function.
Am I right in thinking that this will cause the volume to be unmapped if needed?

Could we map the `getMappedDevPath` return a cleanup function even if `false` was passed to `mapVolume`, such that if its not mapped the returned function is a no-op, and if it is mapped the returned function would trigger an unmap?

Then we could split the majority of these functions into a function in the `common` driver, that accepts a `deactivate func() error` argument, and then each driver can pass it a driver-specific deactivate function, whilst allowing us to keep the generic stuff available for non-remote drivers too.

So something like this:

```go
func (d *common) mountVolume(devPath (vol Volume, activate bool) (string, revert.Hook, error), vol Volume, op *operations.Operation) error {
revert := revert.New()
defer revert.Fail()

volDevPath, deactivate, err := devPath(v, true)
if err != nil {
return err
}

revert.Add(deactivate)

switch vol.contentType {
	case ContentTypeFS:
		mountPath := vol.MountPath()
		if !filesystem.IsMountPoint(mountPath) {
			err = vol.EnsureMountPath()
			if err != nil {
				return err
			}

			fsType := vol.ConfigBlockFilesystem()

			if vol.mountFilesystemProbe {
				fsType, err = fsProbe(volDevPath)
				if err != nil {
					return fmt.Errorf("Failed probing filesystem: %w", err)
				}
			}

			mountFlags, mountOptions := filesystem.ResolveMountOptions(strings.Split(vol.ConfigBlockMountOptions(), ","))
			err = TryMount(context.TODO(), volDevPath, mountPath, fsType, mountFlags, mountOptions)
			if err != nil {
				return err
			}

			d.Logger().Debug("Mounted volume", logger.Ctx{"driver": d.Info().Name, "volName": vol.name, "dev": volDevPath, "path": mountPath, "options": mountOptions})
		}

	case ContentTypeBlock:
		// For VMs, mount the filesystem volume.
		if vol.IsVMBlock() {
			fsVol := vol.NewVMBlockFilesystemVolume()
			err := d.MountVolume(fsVol, op)
			if err != nil {
				return err
			}
		}
	}
}
```


Then for unmount:

```go
func (d *common) unmountVolume(devPath (vol Volume, activate bool) (string, revert.Hook, error), vol Volume, op *operations.Operation) (bool, error) {

    ourUnmount := false
	mountPath := vol.MountPath()
	refCount := vol.MountRefCountDecrement()

	// Attempt to unmount the volume.
	if vol.contentType == ContentTypeFS && filesystem.IsMountPoint(mountPath) {
		if refCount > 0 {
			d.Logger().Debug("Skipping unmount as in use", logger.Ctx{"volName": vol.name, "refCount": refCount})
			return false, ErrInUse
		}

		err := TryUnmount(mountPath, unix.MNT_DETACH)
		if err != nil {
			return false, err
		}

		d.Logger().Debug("Unmounted volume", logger.Ctx{"driver": d.Info().Name, "volName": vol.name, "path": mountPath, "keepBlockDev": keepBlockDev})

		// Attempt to unmap.
		if !keepBlockDev {
                _, deactivate, err := devPath(vol, false)
                if err != nil {
                     return false, err
                }

                err = deactivate(vol)
    			if err != nil {
    				return false, err
    			}
		}

		ourUnmount = true
	} else if vol.contentType == ContentTypeBlock {
		// For VMs, unmount the filesystem volume.
		if vol.IsVMBlock() {
			fsVol := vol.NewVMBlockFilesystemVolume()
			ourUnmount, err = d.UnmountVolume(fsVol, false, op)
			if err != nil {
				return false, err
			}
		}

		if !keepBlockDev {
			// Check if device is currently mapped (but don't map if not).
			devPath, _, _ := d.getMappedDevPath(vol, false)
			if devPath != "" && shared.PathExists(devPath) {
				if refCount > 0 {
					d.Logger().Debug("Skipping unmount as in use", logger.Ctx{"volName": vol.name, "refCount": refCount})
					return false, ErrInUse
				}

				// Attempt to unmap.
                _, deactivate, err := devPath(vol, false)
                if err != nil {
                     return false, err
                }

                err = deactivate(vol)
    			if err != nil {
    				return false, err
    			}

				ourUnmount = true
			}
		}
	}

return ourUnmount, nil
}

```

mihalicyn: Thanks!

roosterfish: > Could we map the getMappedDevPath return a cleanup function even if false was passed to mapVolume

Do I understand you right that when running `_, deactivate, err := devPath(vol, false)` the `deactivate` func should be a noop in case the vol is not yet mounted.
But running the same again when the vol is mounted causes `deactivate` to contain the necessary steps to unmap the vol? 

That should work as currently in PowerFlex/Pure none of the invocations of `getMappedDevPath` which pass `false` make use of the returned cleanup hook so we can start using it for this use case.

I like the idea as `devPath (vol Volume, activate bool) (string, revert.Hook, error)` is effectively the signature of the `getMappedDevPath` funcs so we can just pass those. And Pure will likely do the same.

tomponline: > Do I understand you right that when running `_, deactivate, err := devPath(vol, false)` the `deactivate` func should be a noop in case the vol is not yet mounted.
> But running the same again when the vol is mounted causes `deactivate` to contain the necessary steps to unmap the vol?

yes exactly

lxd/storage/drivers/utils_api.go
mihalicyn: ok, I've removed this part.

lxd/storage/drivers/remote_storage_helpers.go
mihalicyn: Dear friends,

thanks for all the input in there. I've completely reworked this as you suggested.

The only thing is that I kept the behavior of `getMappedDevPath(vol, false)` to return an error in case if volume is not mapped. Because, just from a function name caller expects to get a mapped device path and if device is not mapped then we should return an error, IMHO. Let me know if you disagree with this and I can fix this.

lxd/storage/drivers/driver_powerflex_utils.go
mihalicyn: Have done. Thanks!

lxd/storage/connectors/connector_nvme.go
mihalicyn: Have fixed, thanks!

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -1027,7 +1027,8 @@ func (d *powerflex) mapVolume(vol Volume) (revert.Hook, error) {
 
 // getMappedDevPath returns the local device path for the given volume.
 // Indicate with mapVolume if the volume should get mapped to the system if it isn't present.
-func (d *powerflex) getMappedDevPath(vol Volume, mapVolume bool) (string, revert.Hook, error) {
+func (d *powerflex) getMappedDevPath(vol Volume, mapVolume bool) (string, func() error, error) {
```
tomponline: can you use named return values in the function signature for clarity for the reader, e.g.

```go
getMappedDevPath(vol Volume, mapVolume bool) (devPath string, deactivate func() error,  err error) {
```

lxd/storage/drivers/generic_vfs.go
```diff
@@ -1194,3 +1195,144 @@ func genericVFSListVolumes(d Driver) ([]Volume, error) {
...
+	revert := revert.New()
+	defer revert.Fail()
+
+	// Activate volume if needed.
+	volDevPath, cleanup, err := getDevicePath(vol, true)
+	if err != nil {
+		return err
+	}
+
+	revert.Add(func() { _ = cleanup() })
```
tomponline: Can you log as a warning any error from cleanup please using d.logger.

lxd/storage/drivers/driver_powerflex_volumes.go
```diff
@@ -78,7 +75,7 @@ func (d *powerflex) CreateVolume(vol Volume, filler *VolumeFiller, op *operation
 			return err
 		}
 
-		revert.Add(cleanup)
+		revert.Add(func() { _ = cleanup() })
```
tomponline: Can you log as a warning any error from cleanup please using d.logger.

lxd/storage/drivers/driver_pure_volumes.go
tomponline, line 1: Can you log as a warning any error from `cleanup` please using d.logger.

tomponline: or if the error from `cleanup` is never used, it it better to have the cleanup function itself log the warning?

lxd/storage/drivers/generic_vfs.go
```diff
@@ -1194,3 +1195,144 @@ func genericVFSListVolumes(d Driver) ([]Volume, error) {
 
 	return vols, nil
 }
+
+// mountVolume mounts a volume and increments ref counter. Please call UnmountRemoteVolume() when done with the volume.
```
tomponline: comment is referencing non-existent `UnmountRemoteVolume`

lxd/storage/connectors/connector_nvme.go
tomponline, line 1: Please address the linter errors in this file

lxd/storage/drivers/generic_vfs.go
```diff
@@ -1194,3 +1195,144 @@ func genericVFSListVolumes(d Driver) ([]Volume, error) {
...
+			d.Logger().Debug("Skipping unmount as in use", logger.Ctx{"volName": vol.name, "refCount": refCount})
+			return false, ErrInUse
+		}
+
+		err := TryUnmount(mountPath, unix.MNT_DETACH)
+		if err != nil {
+			return false, err
+		}
+
+		d.Logger().Debug("Unmounted volume", logger.Ctx{"driver": d.Info().Name, "volName": vol.name, "path": mountPath, "keepBlockDev": keepBlockDev})
```
tomponline: d.Logger() includes pool name, so we probably dont need "driver" here as well.

lxd/storage/drivers/generic_vfs.go
```diff
@@ -1194,3 +1195,144 @@ func genericVFSListVolumes(d Driver) ([]Volume, error) {
...
+				return false, err
+			}
+		}
+
+		if !keepBlockDev {
+			// Check if device is currently mapped (but don't map if not).
+			devPath, _, _ := getDevicePath(vol, false)
+			if devPath != "" && shared.PathExists(devPath) {
+				if refCount > 0 {
+					d.Logger().Debug("Skipping unmount as in use", logger.Ctx{"volName": vol.name, "refCount": refCount})
```
tomponline: There's 2 calls to `d.Logger` in this function with similar context, so suggest creating a function specific logger context at the top of the function e.g.

```go
l := d.Logger.().AddContext(logger.Ctx{"volName": vol.name})
```

Then use it, e.g. `l.Debug("Skipping unmount as in use", logger.Ctx{"refCount": refCount})`

mihalicyn: Have done! Thanks!

lxd/storage/drivers/generic_vfs.go
mihalicyn: Fixed, thanks!

lxd/storage/connectors/connector_nvme.go
mihalicyn: Have done.

lxd/storage/drivers/generic_vfs.go
mihalicyn: Good catch. Fixed ;)

lxd/storage/drivers/driver_pure_volumes.go
mihalicyn: >or if the error from cleanup is never used, it it better to have the cleanup function itself log the warning?

it is used in mount/unmount helpers we introduce as a part of a logic, but I agree with your suggestion to put a warning in a deactivate function itself. Have done this. Thanks! ;)

lxd/storage/drivers/driver_powerflex_utils.go
mihalicyn: Fixed!

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -1075,9 +1074,20 @@ func (d *powerflex) getMappedDevPath(vol Volume, mapVolume bool) (string, revert
 		return "", nil, fmt.Errorf("Failed to locate device for volume %q: %w", vol.name, err)
 	}
 
-	cleanup := revert.Clone().Fail
+	// At this point, we know that device is mapped and device path was found.
+	// We want to return a deactivation function to a caller so it can
+	// unmap the volume if needed.
+	deactivate = func() error {
```
roosterfish: There is one problem with this.

`getMappedDevPath` is also used in cases where the volume is already mapped, e.g. `SetVolumeQuota`. In this case no matter if you call `getMappedDevPath` with `mapVolume=(true|false)`, it will return the path but without performing the mapping.

However the caller of `getMappedDevPath` does not know if he mapped the volume or if it was already mapped. He can just say using `mapVolume` if the volume should be mapped if it doesn't yet exist. But the returned `revert.Hook` is executed in all cases to ensure stuff is cleaned up if the caller performed the mapping.

Now if we always add `unmapVolume`, the caller will always unmap the volume even if he wasn't the one performing the mapping. See https://github.com/canonical/lxd/blob/main/lxd/storage/drivers/driver_powerflex_volumes.go#L604-L609.

roosterfish: I have just checked the Pure driver and it looks the returned `revert.Hook` is added to the caller's `revert` which IMHO will cause the volume not to be unmapped when running `SetVolumeQuota` and the volume wasn't mapped before. See https://github.com/canonical/lxd/blob/main/lxd/storage/drivers/driver_pure_volumes.go#L781.

roosterfish: Let's populate `deactivate` only if `mapVolume=true` and the volume didn't exist before. Maybe we should call `devicePath, err = block.GetDiskDevicePath(prefix, devicePathFilter)` right at the beginning to identify this?

mihalicyn: Yeah, you are absolutely right. Let me think a bit what we can do. I really liked Tom's idea to somehow hide map/unmap details and use these callback functions as a return values...

mihalicyn: >Let's populate deactivate only if mapVolume=true and the volume didn't exist before. Maybe we should call devicePath, err = block.GetDiskDevicePath(prefix, devicePathFilter) right at the beginning to identify this?

yes, this would fix the bug, but breaks Tom's idea. Because what we wanted is to have some way to unmap the volume even if caller is not the one who mapped it.

Ah! I know. We can introduce a parameter to the `deactivate` callback, something like `deactivate = func(unmapAnyways bool) error {` and if it is set to `true` then we do unmap even if we are not the one who mapped it, if it is `false` then we just keep the old semantics. WDYT?

tomponline: >  IMHO will cause the volume not to be unmapped when running SetVolumeQuota and the volume wasn't mapped before

@roosterfish isnt that what we want to happen though?

tomponline: @roosterfish can we update the pure driver to always call cleanup even on success?

roosterfish: > Because what we wanted is to have some way to unmap the volume even if caller is not the one who mapped it.

In which scenario do we want this to happen? It sounds like the type of issues we iterated over a few months ago when adding the `connector` package. 

> We can introduce a parameter to the deactivate callback

That might work yes. But you would still move the `block.GetDiskDevicePath` check right before the actual volume map operation to be able to figure out if `unmapAnyways=false` is allowed to trigger `unampVolume` (if the caller was mapping it)?

mihalicyn: > In which scenario do we want this to happen? It sounds like the type of issues we iterated over a few months ago when adding the `connector` package.

here https://github.com/canonical/lxd/pull/15885/files#diff-2ab8ba8696630342b067053a70d715c9641767a5ad7d85c5f526e1fe3e098bb9R1292

idea is that we want replace an explicit calls to `unmapVolume` with calls to the `deactivate` callback to make functions more generic.

mihalicyn: I've updated the PR. As we discussed I went the most conservative way and introduced a new callback function argument (unmapVolume). It allows us to keep the current logic, but generalize mount/unmount functions easily.

lxd/storage/drivers/generic_vfs.go
```diff
@@ -1194,3 +1196,138 @@ func genericVFSListVolumes(d Driver) ([]Volume, error) {
...
+		if vol.IsVMBlock() {
+			fsVol := vol.NewVMBlockFilesystemVolume()
+			err := d.MountVolume(fsVol, op)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	vol.MountRefCountIncrement() // From here on it is up to caller to call UnmountVolume() when done.
```
roosterfish: The comment requires an update to use lower case `unmountVolume()` after the code was moved from the driver level func to this helper.

lxd/storage/drivers/driver_powerflex_volumes.go
```diff
@@ -692,135 +689,13 @@ func (d *powerflex) defaultBlockVolumeSize() string {
...
-					return false, err
-				}
-
-				ourUnmount = true
-			}
-		}
-	}
-
-	return ourUnmount, nil
+	return unmountVolume(d, vol, keepBlockDev, d.getMappedDevPath, d.unmapVolume, op)
```
roosterfish, line 698: I really like how "wrapped" the operations now actually are.

roosterfish (changes requested):
Looks good to me, only one comment that requires an update.
Once I get my hands on another PowerFlex env which supports NVMe/TCP, I'll give it a test! The one in the Dell lab I am currently using as an alternative only has SDC.

lxd/storage/drivers/generic_vfs.go
mihalicyn: Good catch! Have fixed, thanks ;-)

roosterfish (dismissed):
Tests pending for PowerFlex due to env access, but I don't want to block this any longer. IMHO the latest changes don't impact the existing functionality in any way.

mihalicyn:
Force pushed once again and rebased on top of main without any changes in my commits.

mihalicyn:
Thanks!

roosterfish:
> Just tested again, looks good to me.

This also just passed the `lxd-ci/tests/storage-vm` using both PowerFlex `nvme` and `sdc` modes.

Issue #10660, opened 2022-07-12
Add required=false option for nic devices with an unavailable parent


m-ueberall:
# Required information

 * Distribution: Ubuntu
 * Distribution version: 20.04 ("Focal Fossa")
 * Kernel version: 5.15.x (x86_64, aarch64) 
 * LXC, LXD version: 5.3

# Issue description

If a container uses a network which is unavailable, it cannot be started (even if additional, available networks exist).
It would be helpful to be able to mark certain networks (network types?) – in my case: an OVN network – as `required: "false"` like with devices/disk paths so that it's not necessary to modify the container's configuration in order to bring it up after, e.g., rebooting the host machine. (Note that _running_ containers are "unaffected" if the network becomes unavailable.)

# Steps to reproduce

 1. Ensure that a certain network used by one or more containers is available
 2. Start all containers
 3. Bring down the network in question (while all containers from the previous step are still up)
 4. Stop and try to restart one of the above containers; the last step will now fail:
 `Error: Failed pre-start check for device "eth0": Network "ovn01" unavailable on this server`

SooryR:
If no one else has taken this issue, I would like to try to tackle it. It would help me considerably if you could show me where are the related files and any tips.

tomponline:
Thanks assigned to you.

tomponline:
If you look for the `PreStartCheck()` functions in the code base, you'll see some of the NICs have this bit:

```go
	// If managed network is not available, don't try and start instance.
	if d.network.LocalStatus() == api.StoragePoolStatusUnvailable {
		return api.StatusErrorf(http.StatusServiceUnavailable, "Network %q unavailable on this server", d.network.Name())
	}
```

You'll need to introduce a NIC device setting called `required` (similar to the `disk` one here https://linuxcontainers.org/lxd/docs/master/instances/#type-disk) that defaults to `true` if not specified. If set to `false` then you can skip the check in PreStartCheck() and then in `Start()` if the network isn't available just return `nil`.

tomponline:
@bakhtiar56 perhaps you would like to look at this one as it looks like @SooryR has not been available to do it.

bakhtiar56:
sure, I would love to tackle it.

jxie0615:
Hi! Is this issue still open to be worked on? My partner and I are students from UT Austin, and we were wondering if we could tackle this as a good first issue for our virtualization class?

tomponline:
[...]

PR #15887, opened 2025-06-27
Network: Add `required` configuration option for NIC devices

jade-abc:
Closes https://github.com/canonical/lxd/issues/10660.

implemented required flag across ovn, macvlan and bridged. Can change across all nictypes but others seemed less appropriate.

I have used the disk required implementation as a guide however there were more updates across files than originally expected. 

Tested on my system with ovn, macvlan and bridged network, required=true (default behaviour) works as required. 

Tested network being unavailable on startup, available on startup and becoming unavailable whilst container is running.

Unsure about how I'm handling the failure in nictypes.go and guidance would be appreciated if incorrect. 

Thanks for checking it out, looking forward to hearing any changes you may suggest. Let me know if I've done anything wrong!

### DIFF
--- a/lxd/device/nictype/nictype.go
+++ b/lxd/device/nictype/nictype.go
@@ -10,6 +10,7 @@ import (
 	deviceConfig "github.com/canonical/lxd/lxd/device/config"
 	"github.com/canonical/lxd/lxd/project"
 	"github.com/canonical/lxd/lxd/state"
+	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
 )
 
@@ -18,43 +19,46 @@ import (
 // resolved from the network's type. Otherwise the device's "nictype" property is returned (which may be empty if
 // used with non-NIC device configs).
 func NICType(s *state.State, deviceProjectName string, d deviceConfig.Device) (string, error) {
-	// NIC devices support resolving their "nictype" from their "network" property.
-	if d["type"] == "nic" {
-		if d["network"] != "" {
-			// Translate device's project name into a network project name.
-			networkProjectName, _, err := project.NetworkProject(s.DB.Cluster, deviceProjectName)
-			if err != nil {
-				return "", fmt.Errorf("Failed to translate device project %q into network project: %w", deviceProjectName, err)
+	required := shared.IsTrueOrEmpty(d["required"])
+	if required {
+		// NIC devices support resolving their "nictype" from their "network" property.
+		if d["type"] == "nic" {
+			if d["network"] != "" {
+				// Translate device's project name into a network project name.
+				networkProjectName, _, err := project.NetworkProject(s.DB.Cluster, deviceProjectName)
+				if err != nil {
+					return "", fmt.Errorf("Failed to translate device project %q into network project: %w", deviceProjectName, err)
+				}
+
+				var netInfo *api.Network
+
+				err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+					_, netInfo, _, err = tx.GetNetworkInAnyState(ctx, networkProjectName, d["network"])
+
+					return err
+				})
+				if err != nil {
+
+					return "", fmt.Errorf("Failed to load network %q for project %q: %w", d["network"], networkProjectName, err)
+				}
+				var nicType string
+				switch netInfo.Type {
+				case "bridge":
+					nicType = "bridged"
+				case "macvlan":
+					nicType = "macvlan"
+				case "sriov":
+					nicType = "sriov"
+				case "ovn":
+					nicType = "ovn"
+				case "physical":
+					nicType = "physical"
+				default:
+					return "", fmt.Errorf("Unrecognised NIC network type for network %q", d["network"])
+				}
+
+				return nicType, nil
 			}
-
-			var netInfo *api.Network
-
-			err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
-				_, netInfo, _, err = tx.GetNetworkInAnyState(ctx, networkProjectName, d["network"])
-
-				return err
-			})
-			if err != nil {
-				return "", fmt.Errorf("Failed to load network %q for project %q: %w", d["network"], networkProjectName, err)
-			}
-
-			var nicType string
-			switch netInfo.Type {
-			case "bridge":
-				nicType = "bridged"
-			case "macvlan":
-				nicType = "macvlan"
-			case "sriov":
-				nicType = "sriov"
-			case "ovn":
-				nicType = "ovn"
-			case "physical":
-				nicType = "physical"
-			default:
-				return "", fmt.Errorf("Unrecognised NIC network type for network %q", d["network"])
-			}
-
-			return nicType, nil
 		}
 	}
 


kadinsayani:
Hi @jade-abc 👋 

Thanks for your first contribution to LXD! Please take a look at https://documentation.ubuntu.com/lxd/latest/contributing/ and ensure the following: 

a) Commits are cryptographically signed, see https://docs.github.com/en/authentication/managing-commit-signature-verification.
b) Commit messages follow our style guide: https://documentation.ubuntu.com/lxd/latest/contributing/#commit-structure.
c) Formatting changes are included in the original commit with the change.

Furthermore, rather than adding additional commits for changes that depend on a previous commit, please modify the original commit and rebase. 

Thanks!

lxd/device/nic_ovn.go
```diff
@@ -112,6 +113,9 @@ func (d *nicOVN) validateConfig(instConf instance.ConfigReader) error {
 	// Lookup network settings and apply them to the device's config.
 	n, err := network.LoadByName(d.state, networkProjectName, d.config["network"])
 	if err != nil {
+		if d.config["required"] == "false" {
+			return nil
+		}
```
kadinsayani, line 118:
```suggestion
		}
		
```

lxd/device/nic_bridged.go
```diff
@@ -214,6 +215,9 @@ func (d *nicBridged) validateConfig(instConf instance.ConfigReader) error {
 		var err error
 		d.network, err = network.LoadByName(d.state, api.ProjectDefaultName, d.config["network"])
 		if err != nil {
+			if d.config["required"] == "false" {
+				return nil
+			}
```
kadinsayani:
```suggestion
			}
			
```

lxd/device/nic_bridged.go
```diff
@@ -486,6 +490,11 @@ func (d *nicBridged) PreStartCheck() error {
 		return nil
 	}
 
+	//skip status check and start container if not required
```
kadinsayani:
```suggestion
	// Skip status check and start container.
```

lxd/device/nic_bridged.go
```diff
@@ -501,6 +510,9 @@ func (d *nicBridged) Start() (*deviceConfig.RunConfig, error) {
 		return nil, err
 	}
 
+	if (d.network == nil || d.network.LocalStatus() == api.NetworkStatusUnavailable) && d.config["required"] == "false" {
+		return nil, nil
+	}
```
kadinsayani, line 508:
```suggestion
	}
	
```

lxd/device/nic_bridged.go
```diff
@@ -1796,6 +1808,9 @@ func (d *nicBridged) getHostMTU() (int, error) {
 
 // Register sets up anything needed on LXD startup.
 func (d *nicBridged) Register() error {
+	if d.network == nil && d.config["required"] == "false" {
+		return nil
+	}
```
kadinsayani:
```suggestion
	}
	
```

lxd/device/nic_macvlan.go
```diff
@@ -71,6 +72,9 @@ func (d *nicMACVLAN) validateConfig(instConf instance.ConfigReader) error {
 		var err error
 		d.network, err = network.LoadByName(d.state, api.ProjectDefaultName, d.config["network"])
 		if err != nil {
+			if d.config["required"] == "false" {
+				return nil
+			}
```
kadinsayani, line 77:
```suggestion
			}
			
```

lxd/device/nic_ovn.go
kadinsayani, line 118:
```suggestion
		}
		
```

lxd/device/nic_ovn.go
```diff
@@ -478,6 +485,11 @@ func (d *nicOVN) PreStartCheck() error {
 		return nil
 	}
 
+	//skip status check and start container if not requireu
```
kadinsayani:
```suggestion
	// Skip status check and start container.
```

lxd/device/nic_ovn.go
```diff
@@ -514,6 +526,9 @@ func (d *nicOVN) Start() (*deviceConfig.RunConfig, error) {
 	saveData := make(map[string]string)
 	saveData["host_name"] = d.config["host_name"]
 
+	if d.network == nil && d.config["required"] == "false" {
+		return nil, nil
+	}
```
kadinsayani, line 532:
```suggestion
	}
	
```

lxd/device/nictype/nictype.go
```diff
@@ -35,6 +35,9 @@ func NICType(s *state.State, deviceProjectName string, d deviceConfig.Device) (s
 				return err
 			})
 			if err != nil {
+				if d["required"] == "false" {
+					return "bridged", nil
+				}
```
kadinsayani, line 58:
```suggestion
				}
				
```

lxd/device/nic.go
```diff
@@ -59,6 +59,14 @@ func nicValidationRules(requiredFields []string, optionalFields []string, instCo
...
 		//  shortdesc: Name of the host device to join the instance to
 		"parent": validate.IsAny,
+		// lxdmeta:generate(entities=device-nic-{bridged+macvlan+ovn} group=device-conf; key=required)
+		// Container will start regardless of this networks availability
+		// ---
+		//  type: bool
+		//  defaultdesc: `true`
+		//  required: no
+		//  shortdesc: This network is required
+		"required": validate.Optional(validate.IsBool),
```
kadinsayani:
```suggestion
		// lxdmeta:generate(entities=device-nic-{bridged+macvlan+ovn} group=device-conf; key=required)
		// 
		// ---
		//  type: bool
		//  defaultdesc: `true`
		//  required: no
		//  shortdesc: Whether this interface is required to start the instance 
		"required": validate.Optional(validate.IsBool),
```

jade-abc:
> Hi @jade-abc 👋
> 
> Thanks for your first contribution to LXD! Please take a look at https://documentation.ubuntu.com/lxd/latest/contributing/ and ensure the following:
> 
> a) Commits are cryptographically signed, see https://docs.github.com/en/authentication/managing-commit-signature-verification. b) Commit messages follow our style guide: https://documentation.ubuntu.com/lxd/latest/contributing/#commit-structure. c) Formatting changes are included in the original commit with the change.
> 
> Furthermore, rather than adding additional commits for changes that depend on a previous commit, please modify the original commit and rebase.
> 
> Thanks!

Thanks for the help! Would you like me to rebase this commit? :)

jade-abc (commented):
Thank you for looking through that, I really appreciate it:)

go1.24.4.linux-amd64.tar.gz
kadinsayani, line 1: This file should not be included in any commit.

build.log
kadinsayani, line 1: This file should not be included in any commit.

kadinsayani (changes requested):
Once https://github.com/canonical/lxd/pull/15887#issuecomment-3014225970 has been addressed, I would expect to see something like this for the commits:

Commit 1: "lxd/device: Add `required` configuration option for NIC devices"
Commit 2: "lxd/nictype: Add check for `required` configuration option"

Thanks!

lxd/device/nic.go
```diff
@@ -59,6 +59,14 @@ func nicValidationRules(requiredFields []string, optionalFields []string, instCo
 		//  type: string
 		//  shortdesc: Name of the host device to join the instance to
 		"parent": validate.IsAny,
+		// lxdmeta:generate(entities=device-nic-{bridged+macvlan+ovn} group=device-conf; key=required)
```
kadinsayani:
```suggestion
		// lxdmeta:generate(entities=device-nic-{bridged+macvlan+ovn}; group=device-conf; key=required)
```

jade-abc: Thank you for the thorough code review, I have made the changes and recommitted with the requested commit structure and signed commits

kadinsayani: Can this change please be included in the first commit?

lxd/device/nictype/nictype.go
```diff
@@ -35,6 +35,10 @@ func NICType(s *state.State, deviceProjectName string, d deviceConfig.Device) (s
 				return err
 			})
 			if err != nil {
+				if d["required"] == "false" {
```
kadinsayani: Rather than checking `required` here, perhaps we can have something like this:

```Go
required := shared.IsTrueOrEmpty(d["required"])
// NIC devices support resolving their "nictype" from their "network" property.
if required && d["type"] == "nic" {
	if d["network"] != "" {
	...
	}
}

var nicType string
switch netInfo.Type {
case "bridge":
	nicType = "bridged"
...
```

We can follow a similar approach with the other checks.

kadinsayani (changes requested):

lxd/device/nic.go
```diff
@@ -59,6 +59,14 @@ func nicValidationRules(requiredFields []string, optionalFields []string, instCo
 		//  type: string
 		//  shortdesc: Name of the host device to join the instance to
 		"parent": validate.IsAny,
+		// lxdmeta:generate(entities=device-nic-{bridged+macvlan+ovn}; group=device-conf; key=required)
+		// 
+		// ---
+		//  type: bool
+		//  defaultdesc: `true`
+		//  required: no
+		//  shortdesc: Whether this network is required to start the instance
```
kadinsayani:
```suggestion
		//  shortdesc: Whether this interface is required to start the instance 
```

lxd/device/nictype/nictype.go
jade-abc: Had a go at this, slightly differs from what your implementations braces imply in that the netinfo object is created within the validation logic so the switch statement is nested one more level. Happy to amend again if necessary

lxd/device/nic.go
kadinsayani: I see you opted for `device` here. Can we please go with `interface` since that is the type of device we are referring to? Thanks.

jade-abc: Yes of course, genuinely not sure how I managed to mix those two up. Thanks for spotting

kadinsayani: Actually, let's align this with the existing `required` configuration option for `disk` devices. How about `Whether to fail if the parent network isn't available`?

jade-abc: changed that:)

jade-abc: > Actually, let's align this with the existing `required` configuration option for `disk` devices. How about `Whether to fail if the parent network isn't available`?

Sure, I'll do this now

kadinsayani: Awesome thanks!

lxd/device/nictype/nictype.go
kadinsayani: Cool, thanks. I had a closer look at these changes and I think we can avoid changing the behavior of `NICType()` if we take a slightly different approach to where we check the `required` configuration option. Please see my other comment: https://github.com/canonical/lxd/pull/15887#pullrequestreview-2972179718.

jade-abc:
> Thanks again for your work on this!
> 
> A few more suggestions:
> 
> 1. We can add a helper `isRequired` function, similar to:
> 
> https://github.com/canonical/lxd/blob/a6fd94390c044bcc3efd689ad231d9873c664733/lxd/device/disk.go#L124-L133
> 
> 2. I think we're making more changes than required here. Can you try following the suggested approach outlined in this comment: [#10660 (comment)](https://github.com/canonical/lxd/issues/10660#issuecomment-1209059522)?
> 
> Thanks!

Hi, I'll have to drop off for a few hours but happy to add that helper function. Makes things more clean. I'll take a look at the implementation and try to align it towards that comment. In the mean time, can you clarify where you see the unnecessary changes if at all possible?

lxd/device/nic.go
```diff
@@ -59,6 +59,14 @@ func nicValidationRules(requiredFields []string, optionalFields []string, instCo
...
 		//  shortdesc: Name of the host device to join the instance to
 		"parent": validate.IsAny,
+		// lxdmeta:generate(entities=device-nic-{bridged+macvlan+ovn}; group=device-conf; key=required)
+		//
+		// ---
+		//  type: bool
+		//  defaultdesc: `true`
+		//  required: no
+		//  shortdesc: Whether to fail if the parent network isn't available
+		"required": validate.Optional(validate.IsBool),
```
kadinsayani, line 69: Since we are updating the metadata, can you please run `make update-metadata` and respond `Y` when prompted to commit changes? Thanks!

kadinsayani:
Looking good and definitely getting closer here :)

Still missing:

> a) Commits are cryptographically signed, see https://docs.github.com/en/authentication/managing-commit-signature-verification.

jade-abc:
> Thanks again for your work on this!
> 
> A few more suggestions:
> 
> 1. We can add a helper `isRequired` function, similar to:
> 
> https://github.com/canonical/lxd/blob/a6fd94390c044bcc3efd689ad231d9873c664733/lxd/device/disk.go#L124-L133
> 
> 2. I think we're making more changes than required here. Can you please try following the suggested approach outlined in this comment: [#10660 (comment)](https://github.com/canonical/lxd/issues/10660#issuecomment-1209059522)?
> 
> Thanks!

Hello, quick question. Would you like an isRequired function to be added to the individual NIC devices that it touches, or would it be worth elevating to device_common? Might be a mistake as the disk requires an extra check of the optional flag but just thought I'd ask.

lxd/device/nic.go
jade-abc: Yep, done that!

kadinsayani:
> Hello, quick question. Would you like an isRequired function to be added to the individual NIC devices that it touches, or would it be worth elevating to device_common? Might be a mistake as the disk requires an extra check of the optional flag but just thought I'd ask.

We probably want to include the `required` option for all NIC types, so we should add the `isRequired` helper function to `lxd/device/nic.go`. Is there a reason why the other NIC types are excluded?

The check for the `optional` configuration option is not relevant here, only for `disk` devices.

lxd/device/nic.go
kadinsayani: I don't see the commit here, maybe you forgot to push?

jade-abc:
> > Hello, quick question. Would you like an isRequired function to be added to the individual NIC devices that it touches, or would it be worth elevating to device_common? Might be a mistake as the disk requires an extra check of the optional flag but just thought I'd ask.
> 
> We probably want to include the `required` option for all NIC types, so we should add the `isRequired` helper function to `lxd/device/nic.go`. Is there a reason why the other NIC types are excluded?
> 
> The check for the `optional` configuration option is not relevant here, only for `disk` devices.

I agree with adding it to nic.go if appropriate. The only reason sriov doesn't have a check is I was unsure what was in scope. Happy to add it to sriov and elevate isRequired to nic.go if thats the plan

kadinsayani:
> > > Hello, quick question. Would you like an isRequired function to be added to the individual NIC devices that it touches, or would it be worth elevating to device_common? Might be a mistake as the disk requires an extra check of the optional flag but just thought I'd ask.
> > 
> > 
> > We probably want to include the `required` option for all NIC types, so we should add the `isRequired` helper function to `lxd/device/nic.go`. Is there a reason why the other NIC types are excluded?
> > The check for the `optional` configuration option is not relevant here, only for `disk` devices.
> 
> I agree with adding it to nic.go if appropriate. The only reason sriov doesn't have a check is I was unsure what was in scope. Happy to add it to sriov and elevate isRequired to nic.go if thats the plan

@tomponline do we want to include the `required` configuration option for all NIC types?

lxd/device/nic_bridged.go
```diff
@@ -214,6 +215,10 @@ func (d *nicBridged) validateConfig(instConf instance.ConfigReader) error {
 		var err error
 		d.network, err = network.LoadByName(d.state, api.ProjectDefaultName, d.config["network"])
 		if err != nil {
+			if d.config["required"] == "false" {
```
tomponline: I believe the intention with this PR was to allow instance device to start if the parent network was not in the "up" state.

I still think it should fail to start if it references an unknown managed network - this is a configuration issue rather than an environmental issue.

lxd/device/nic_bridged.go
```diff
@@ -486,6 +491,11 @@ func (d *nicBridged) PreStartCheck() error {
 		return nil
 	}
 
+	// Skip status check and start container.
+	if shared.IsFalse(d.config["required"]) {
```
tomponline: Lets move this as part of the if statement on line 500 to keep the related concepts together, so something like this:

```go
// If managed network is not available, don't try and start instance.
if shared.IsTrueOrEmpty(d.config["required"]) && d.network.LocalStatus() == api.NetworkStatusUnavailable {
    return api.StatusErrorf(http.StatusServiceUnavailable, "Network %q unavailable on this server", d.network.Name())
}
```

jade-abc:
> > > > Hello, quick question. Would you like an isRequired function to be added to the individual NIC devices that it touches, or would it be worth elevating to device_common? Might be a mistake as the disk requires an extra check of the optional flag but just thought I'd ask.
> > > 
> > > 
> > > We probably want to include the `required` option for all NIC types, so we should add the `isRequired` helper function to `lxd/device/nic.go`. Is there a reason why the other NIC types are excluded?
> > > The check for the `optional` configuration option is not relevant here, only for `disk` devices.
> > 
> > 
> > I agree with adding it to nic.go if appropriate. The only reason sriov doesn't have a check is I was unsure what was in scope. Happy to add it to sriov and elevate isRequired to nic.go if thats the plan
> 
> @tomponline do we want to include the `required` configuration option for all NIC types?

Any updates on this? :)

lxd/device/nic_bridged.go
```diff
@@ -501,6 +511,10 @@ func (d *nicBridged) Start() (*deviceConfig.RunConfig, error) {
 		return nil, err
 	}
 
+	if (d.network == nil || d.network.LocalStatus() == api.NetworkStatusUnavailable) && d.config["required"] == "false" {
```
tomponline: I think this is a regression when using `parent=<unmanaged interface>` as `d.network` will always be nil then, and then the NIC will silently not start.

lxd/device/nic_bridged.go
```diff
@@ -1799,6 +1813,10 @@ func (d *nicBridged) getHostMTU() (int, error) {
 
 // Register sets up anything needed on LXD startup.
 func (d *nicBridged) Register() error {
+	if d.network == nil && d.config["required"] == "false" {
```
tomponline: i think we need to stop considering `d.network == nil` as an unavailable network, as it can be so for NICs connecting to unmanaged parent interfaces.

jade-abc:
> Looking good and definitely getting closer here :)
> 
> Still missing:
> 
> > a) Commits are cryptographically signed, see https://docs.github.com/en/authentication/managing-commit-signature-verification.

Apologies. Will sign next commits

jade-abc:
> > Looking good and definitely getting closer here :)
> > Still missing:
> > > a) Commits are cryptographically signed, see https://docs.github.com/en/authentication/managing-commit-signature-verification.
> 
> Apologies. Will sign next commits

As in, will sign these when I recommit again

lxd/device/nic.go
```diff
@@ -59,6 +59,14 @@ func nicValidationRules(requiredFields []string, optionalFields []string, instCo
 		//  type: string
 		//  shortdesc: Name of the host device to join the instance to
 		"parent": validate.IsAny,
+		// lxdmeta:generate(entities=device-nic-{bridged+macvlan+ovn}; group=device-conf; key=required)
+		//
+		// ---
+		//  type: bool
+		//  defaultdesc: `true`
+		//  required: no
+		//  shortdesc: Whether to fail if the parent network isn't available
```
tomponline, line 68: I think we need to clarify here, and in this overall PR whether we are considering only **managed** network status or whether we are also considering **unmanaged** parent interface status. As those use different statuses.

tomponline:
> > @tomponline do we want to include the `required` configuration option for all NIC types?

Only for those that make sense to.

lxd/device/nic_bridged.go
jade-abc: Thanks, removed this

lxd/device/nic.go
jade-abc: Any suggestions as to how I can clarify? :)

tomponline: @jade-abc lets first agree if you are intending to consider unavailable managed networks or whether its also unmanaged parent interfaces?

jade-abc: I presumed the use case was for managed devices and unmanaged parent interfaces, as if either become unavailable I doubt someone would want the instance to fail or unable to start if they've clarified the network is not required. Happy to reduce the scope to just managed interfaces if I'm wrong :)

tomponline: why dont we start with just managed networks and then we can add unmanaged parent interfaces in a follow up PR?

jade-abc: Good shout! I will update the PR promptly to reflect this. Thank you both for your patience in checking out my first contribution by the way!

tomponline: Suggest define this as

> Require associated network to be operational when starting the device

We cant use "parent" because a NIC that has `network=foo` doesnt have `parent` defined at all.

jade-abc:
> > > @tomponline do we want to include the `required` configuration option for all NIC types?
> 
> Only for those that make sense to.

my instincts say that it's probably inappropriate to add to nic_sriov.go but I could be misunderstanding something. I haven't added any code to it, but if it should be in scope I am happy to.

tomponline:
> > > > @tomponline do we want to include the `required` configuration option for all NIC types?
> > 
> > 
> > Only for those that make sense to.
> 
> my instincts say that it's probably inappropriate to add to nic_sriov.go but I could be misunderstanding something. I haven't added any code to it, but if it should be in scope I am happy to.

why do you think this?

jade-abc:
> > > > > @tomponline do we want to include the `required` configuration option for all NIC types?
> > > 
> > > 
> > > Only for those that make sense to.
> > 
> > 
> > my instincts say that it's probably inappropriate to add to nic_sriov.go but I could be misunderstanding something. I haven't added any code to it, but if it should be in scope I am happy to.
> 
> why do you think this?

I guess in the model of SR IOV I have if the underlying infrastructure was unavailable it would imply to me that the hardware was unavailable, which I just felt I didn't have enough knowledge to determine whether it's acceptable to degrade gracefully without clarification. 
I suppose there are other reasons that SR IOV may be unavailable outside of hardware availability and that if the user decides its not required who am I to tell them it's required? I also feel like SR IOV is the domain I understand the least here!

still feeling unsure, but erring on the side of adding it, though I would really love some clarification :)

tomponline:
> > > > > > @tomponline do we want to include the `required` configuration option for all NIC types?
> > > > 
> > > > 
> > > > Only for those that make sense to.
> > > 
> > > 
> > > my instincts say that it's probably inappropriate to add to nic_sriov.go but I could be misunderstanding something. I haven't added any code to it, but if it should be in scope I am happy to.
> > 
> > 
> > why do you think this?
> 
> I guess in the model of SR IOV I have if the underlying infrastructure was unavailable it would imply to me that the hardware was unavailable, which I just felt I didn't have enough knowledge to determine whether it's acceptable to degrade gracefully without clarification. I suppose there are other reasons that SR IOV may be unavailable outside of hardware availability and that if the user decides its not required who am I to tell them it's required? I also feel like SR IOV is the domain I understand the least here!
> 
> still feeling unsure, but erring on the side of adding it, though I would really love some clarification :)

I think it makes sense for SRIOV to be included in this new functionality, as hardware unavailability could affect other NIC types too.

jade-abc:
> > > > > > > @tomponline do we want to include the `required` configuration option for all NIC types?
> > > > > 
> > > > > 
> > > > > Only for those that make sense to.
> > > > 
> > > > 
> > > > my instincts say that it's probably inappropriate to add to nic_sriov.go but I could be misunderstanding something. I haven't added any code to it, but if it should be in scope I am happy to.
> > > 
> > > 
> > > why do you think this?
> > 
> > 
> > I guess in the model of SR IOV I have if the underlying infrastructure was unavailable it would imply to me that the hardware was unavailable, which I just felt I didn't have enough knowledge to determine whether it's acceptable to degrade gracefully without clarification. I suppose there are other reasons that SR IOV may be unavailable outside of hardware availability and that if the user decides its not required who am I to tell them it's required? I also feel like SR IOV is the domain I understand the least here!
> > still feeling unsure, but erring on the side of adding it, though I would really love some clarification :)
> 
> I think it makes sense for SRIOV to be included in this new functionality, as hardware unavailability could affect other NIC types too.

fab, thanks for that. I'll get on with that:)

jade-abc:
> > > > > > > > @tomponline do we want to include the `required` configuration option for all NIC types?
> > > > > > 
> > > > > > 
> > > > > > Only for those that make sense to.
> > > > > 
> > > > > 
> > > > > my instincts say that it's probably inappropriate to add to nic_sriov.go but I could be misunderstanding something. I haven't added any code to it, but if it should be in scope I am happy to.
> > > > 
> > > > 
> > > > why do you think this?
> > > 
> > > 
> > > I guess in the model of SR IOV I have if the underlying infrastructure was unavailable it would imply to me that the hardware was unavailable, which I just felt I didn't have enough knowledge to determine whether it's acceptable to degrade gracefully without clarification. I suppose there are other reasons that SR IOV may be unavailable outside of hardware availability and that if the user decides its not required who am I to tell them it's required? I also feel like SR IOV is the domain I understand the least here!
> > > still feeling unsure, but erring on the side of adding it, though I would really love some clarification :)
> > 
> > 
> > I think it makes sense for SRIOV to be included in this new functionality, as hardware unavailability could affect other NIC types too.
> 
> fab, thanks for that. I'll get on with that:)

Apologies for the lack of progress. This is now being actively worked on

tomponline:
I'll close this for now, but please do re-open if/when you get the chance to work on it again. Thanks!

Issue #15882, opened 2025-06-26
Launching a VM with `limits.memory` expressed in `%` should be rejected earlier


simondeziel:
### Issue description

It seems that VMs cannot have their memory based on a % of the host's value which is fine. However, it seems the validation only happens at start/launch time.

### Steps to reproduce

```
$ lxc launch --vm --empty v3 -c limits.memory=30% -d root,size=32MiB
Launching v3
Error: limits.memory invalid: Invalid value: 30%
$ lxc config get v3 limits.memory
30%
$ lxc delete v1
```

```
$ lxc init --empty --vm v1 -c limits.memory=30%
Creating v1
$ lxc config get v1 limits.memory 
30%
$ lxc delete v1
```

### Additional information

The doc on [`limits.memory`](https://documentation.ubuntu.com/lxd/latest/reference/instance_options/#instance-resource-limits:limits.memory) does not mention `%` as being valid but it works fine with containers.

Incus has a fix to allow `%` and also tweak the memory allocation while live: https://github.com/lxc/incus/pull/1270 & https://github.com/lxc/incus/pull/1287

PR #15888, opened 2025-06-27
VM: Fix `limits.memory` when using % of host memory (from Incus)

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/15882.

We should allow setting `limits.memory` to a percentage of host memory to match container `limits.memory` behavior.

### DIFF
--- a/lxd/instance/drivers/util.go
+++ b/lxd/instance/drivers/util.go
@@ -1,0 +1,32 @@
+package drivers
+
+import (
+	"strconv"
+	"strings"
+
+	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/units"
+)
+
+// parseMemoryStr parses a human-readable representation of a memory value.
+func parseMemoryStr(memory string) (valueInt int64, err error) {
+	if strings.HasSuffix(memory, "%") {
+		var percent, memoryTotal int64
+
+		percent, err = strconv.ParseInt(strings.TrimSuffix(memory, "%"), 10, 64)
+		if err != nil {
+			return 0, err
+		}
+
+		memoryTotal, err = shared.DeviceTotalMemory()
+		if err != nil {
+			return 0, err
+		}
+
+		valueInt = (memoryTotal / 100) * percent
+	} else {
+		valueInt, err = units.ParseByteSizeString(memory)
+	}
+
+	return valueInt, err
+}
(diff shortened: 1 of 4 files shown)


test/suites/vm.sh
```diff
@@ -24,9 +24,10 @@ test_vm_empty() {
   echo "Too small VMs"
   ! lxc launch --vm --empty v1 -c limits.memory=0 -d "${SMALL_ROOT_DISK}" || false
   ! lxc launch --vm --empty v1 -c limits.memory=0% -d "${SMALL_ROOT_DISK}" || false
-  # VMs don't support limits.memory in % but it's only detect at start time so needs cleanup
-  ! lxc launch --vm --empty v1 -c limits.memory=10% -d "${SMALL_ROOT_DISK}" || false
-  lxc delete v1
+
+  echo "Tiny VM with tiny memory"
+  lxc launch --vm --empty v1 -c limits.memory=1% -d "${SMALL_ROOT_DISK}"
+  lxc delete -f v1
```
simondeziel: Thanks for also adapting the tests. BTW, when I first ran into this, I had the working `lxc launch` on line 46 use a `%` limit.

The reason to tweak an existing one is because init/launch is relatively expensive so it's best to limit/combine them when possible.

lxd/instance/drivers/util.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"strconv"
+	"strings"
+
+	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/units"
+)
+
+// ParseMemoryStr parses a human-readable representation of a memory value.
+func ParseMemoryStr(memory string) (valueInt int64, err error) {
```
simondeziel: Does it need to be exported? It's only used in the drivers package.

test/suites/vm.sh
kadinsayani: I've updated the working `lxc launch` to use a `%` limit.

lxd/instance/drivers/util.go
kadinsayani: No need to export right now, thanks for pointing that out.

simondeziel (approved):
Thank you!

tomponline (approved):
Ta, do we have tests for container memory percent limits?

simondeziel:
> Ta, do we have tests for container memory percent limits?

https://github.com/canonical/lxd/pull/15890

PR #15889, opened 2025-06-28
VM: Informative error message for QEMU version

xypron:
When reading the QEMU version fails a numeric error code is the only clue to the reason why a failure occurred. It is preferable to provide the output to stderr.

In an instance that I experienced the new message was
```
Failed getting version during QEMU initialization:
/snap/lxd/x2/bin/qemu-system-riscv64:
error while loading shared libraries: libatomic.so.1:
cannot open shared object file: No such file or directory
```
instead of
```
Failed getting version during QEMU initialization: 127
```
With the new message it becomes clear why qemu-system-riscv64 cannot be executed.

### DIFF
--- a/lxd/instance/drivers/driver_qemu.go
+++ b/lxd/instance/drivers/driver_qemu.go
@@ -8716,16 +8716,16 @@ func (d *qemu) Info() instance.Info {
 		return data
 	}
 
-	out, err := exec.Command(qemuPath, "--version").Output()
+	stdout, stderr, err := shared.RunCommandSplit(context.TODO(), nil, nil, qemuPath, "--version")
 	if err != nil {
-		logger.Errorf("Failed getting version during QEMU initialization: %v", err)
+		logger.Errorf("Failed getting version during QEMU initialization: %v (%s)", err, stderr)
 		data.Error = errors.New("Failed getting QEMU version")
 		return data
 	}
 
-	qemuOutput := strings.Fields(string(out))
+	qemuOutput := strings.Fields(stdout)
 	if len(qemuOutput) >= 4 {
-		qemuVersion := strings.Fields(string(out))[3]
+		qemuVersion := qemuOutput[3]
 		data.Version = qemuVersion
 	} else {
 		data.Version = "unknown" // Not necessarily an error that should prevent us using driver.


lxd/instance/drivers/driver_qemu.go
```diff
@@ -8716,9 +8716,12 @@ func (d *qemu) Info() instance.Info {
 		return data
 	}
 
-	out, err := exec.Command(qemuPath, "--version").Output()
+	cmd := exec.Command(qemuPath, "--version")
+	cmd.Stderr = new(bytes.Buffer)
+	out, err := cmd.Output()
 	if err != nil {
-		logger.Errorf("Failed getting version during QEMU initialization: %v", err)
+		logger.Errorf("Failed getting version during QEMU initialization: %s",
```
simondeziel: We tend to use `%w` for errors.

xypron:
Am 29. Juni 2025 17:46:12 MESZ schrieb Simon Deziel ***@***.***>:
>@simondeziel commented on this pull request.
>
>
>
>>  	if err != nil {
>-		logger.Errorf("Failed getting version during QEMU initialization: %v", err)
>+		logger.Errorf("Failed getting version during QEMU initialization: %s",
>
>We tend to use `%w` for errors.
>

According to https://pkg.go.dev/fmt %w is not applicable to strings: "It is invalid to supply the %w verb with an operand that does not implement the error interface."

I wonder why in the current code %v was uses.

Do you want me to print both err and stderr?

Best regards

Heinrich

simondeziel:
Sorry, I missed that you were dealing with a regular string at that point. Please ignore me. Thanks

lxd/instance/drivers/driver_qemu.go
```diff
@@ -8716,9 +8716,12 @@ func (d *qemu) Info() instance.Info {
 		return data
 	}
 
-	out, err := exec.Command(qemuPath, "--version").Output()
+	cmd := exec.Command(qemuPath, "--version")
```
tomponline: Would you mind switching this to `shared.RunCommandSplit()` for consistency with how LXD runs other commands.

lxd/instance/drivers/driver_qemu.go
tomponline: I think we should still include `err` in the message too, so how about:

```go
logger.Errorf("Failed getting version during QEMU initialization: %v (%s)", err, stdErr)
```

tomponline (changes requested):
Thanks, just a couple of suggested tweaks.

lxd/instance/drivers/driver_qemu.go
xypron: Yes, shared.RunCommandSplit() seems to be what is needed. It is just unclear to me why context.TODO() is used.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -8716,16 +8716,16 @@ func (d *qemu) Info() instance.Info {
 		return data
 	}
 
-	out, err := exec.Command(qemuPath, "--version").Output()
+	stdout, stderr, err := shared.RunCommandSplit(context.TODO(), nil, nil, qemuPath, "--version")
 	if err != nil {
-		logger.Errorf("Failed getting version during QEMU initialization: %v", err)
+		logger.Errorf("Failed getting version during QEMU initialization: %s", stderr);
```
simondeziel: I think the issue is the trailing `;`.

lxd/instance/drivers/driver_qemu.go
tomponline: Its in place of having a cancellable command

lxd/instance/drivers/driver_qemu.go
```diff
@@ -8716,16 +8716,16 @@ func (d *qemu) Info() instance.Info {
 		return data
 	}
 
-	out, err := exec.Command(qemuPath, "--version").Output()
+	stdout, stderr, err := shared.RunCommandSplit(context.TODO(), nil, nil, qemuPath, "--version")
 	if err != nil {
-		logger.Errorf("Failed getting version during QEMU initialization: %v", err)
+		logger.Errorf("Failed getting version during QEMU initialization: %s", stderr)
```
tomponline:
```go
logger.Errorf("Failed getting version during QEMU initialization: %v (%s)", err, stderr)
```

tomponline (approved):
Thanks!

simondeziel (approved):

PR #15891, opened 2025-06-29
test/lint/test-tests: check for missing ` || false` fallbacks

simondeziel:


test/README.md
```diff
@@ -30,3 +30,101 @@ Name                           | Default                   | Description
...
+  lxc launch --vm --empty --ephemeral v1 -c limits.memory=128MiB -d "${SMALL_ROOT_DISK}"
+  lxc stop -f v1
+  [ "$(lxc list -f csv -c n)" = "" ]
+}
+```
+
+Good:
+
+```sh
+  echo "Tiny VMs"
```
skozina: Just a thought here... Should we even suggest some special characters in the "test progress" outputs, such as `echo "# Tiny VMs"`? :-)

skozina (approved):
This is awesome! Thank you for the extra work. Appreciate the clear notes in the test/README.md file;-)

test/README.md
simondeziel: Good point. In some places we use `echo "==> ..."` which helps readability.

test/lint/test-tests.sh
tomponline, line 1: Great thanks, this whole failure detection and various special cases is a very sorry situation in bash. 
Kind of unbelievable to me that a language can have this many rules and complexity for something so simple.

tomponline (approved):
ta

PR #15901, opened 2025-06-30
actions/disable-docker: Remove passt package

MusicDin:
Remove `passt` package. @simondeziel suggested adding this in docker action, instead of in conversion test only.

Closes: https://github.com/canonical/lxd-ci/pull/508


simondeziel (commented):
Sorry, this was a bad suggestion I made. This `passt` is not related to Docker in any way so it'd be confusing.

I think your original fix in `lxd-ci.git` itself made more sense.

PR #15906, opened 2025-07-01
test/suites: Use testimage for auth checks.

markylaing:
It looks like the use of the images: remote was originally added in b0a63b97c95dfdabca2c713786d32caee1c3fab1, but there is no reason to pull an image for these tests.

### DIFF
--- a/test/suites/auth.sh
+++ b/test/suites/auth.sh
@@ -1240,9 +1240,8 @@ entities_enrichment_with_entitlements() {
   lxc auth identity-provider-group delete test-idp-group3
 
   # Image
-  lxc init images:busybox/1.36.1 c1
-  lxc delete c1
-  imgFingerprint=$(lxc image list --format json | jq -r '.[] | select(.update_source.alias == "busybox/1.36.1") | .fingerprint')
+  ensure_import_testimage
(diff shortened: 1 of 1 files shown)


test/suites/auth.sh
```diff
@@ -1240,9 +1240,8 @@ entities_enrichment_with_entitlements() {
   lxc auth identity-provider-group delete test-idp-group3
 
   # Image
-  lxc init images:busybox/1.36.1 c1
-  lxc delete c1
-  imgFingerprint=$(lxc image list --format json | jq -r '.[] | select(.update_source.alias == "busybox/1.36.1") | .fingerprint')
+  ensure_import_testimage
+  imgFingerprint="$(lxc query /1.0/images/aliases/testimage | jq -r .target)"
```
simondeziel, line 1244:
```suggestion
  imgFingerprint="$(lxc image info testimage| awk '/^Fingerprint/ {print $2}')"
```
Looks like the `.target` was a typo/copy-n-pasto?

simondeziel: Oh, never mind, I didn't know what the return JSON was like.

PR #15907, opened 2025-07-01
Auth: Improvements to network allocation used by URL checks

### LINKED ISSUE #15855: Project specifier ignored in network list-allocations cli and API
### Please confirm

### Distribution

Ubuntu

### Distribution version

snap

```shell
Name    Version      Rev    Tracking       Publisher   Notes
core20  20250429     2582   latest/stable  canonical✓  base,disabled
core20  20250526     2599   latest/stable  canonical✓  base
core24  20250504     988    latest/stable  canonical✓  base,disabled
core24  20250526     1006   latest/stable  canonical✓  base
lxd     git-1d144a2  34173  latest/edge    canonical✓  disabled
lxd     git-a8d328a  34320  latest/edge    canonical✓  -
snapd   2.68.4       24505  latest/stable  canonical✓  snapd,disabled
snapd   2.68.5       24718  latest/stable  canonical✓  snapd
```

```shell
-
```

### Issue description

The project specifier on the cli for `lxc network list-allocations` is ignored. The api endpoint at `GET /1.0/network-allocations?project=foo` is ignoring the project parameter

### Steps to reproduce

1. Create a project `lxc project create foo --network lxdbr0 --storage default`
2. Create an instance: `lxc launch ubuntu:24.04 u1 --project foo`
3. Create another project: `lxc project create bar --network lxdbr0 --storage default`
4. Create an instance: `lxc launch ubuntu:24.04 u2 --project bar`
5. List network allocations in the first project: `lxc network list-allocations --project=foo`

Instances from both projects are shown, expected is only instances from the specified project `foo`:

```
+-------------------------------------------+--------------------------------------------+---------+----------+------+-------------------+
|                  USED BY                  |                  ADDRESS                   | NETWORK |   TYPE   | NAT  | HARDWARE ADDRESS  |
+-------------------------------------------+--------------------------------------------+---------+----------+------+-------------------+
| /1.0/networks/lxdbr0                      | 10.166.99.1/24                             | lxdbr0  | network  | true |                   |
+-------------------------------------------+--------------------------------------------+---------+----------+------+-------------------+
| /1.0/networks/lxdbr0                      | fd42:f8d8:8280:856a::1/64                  | lxdbr0  | network  | true |                   |
+-------------------------------------------+--------------------------------------------+---------+----------+------+-------------------+
| /1.0/instances/u1?project=foo             | fd42:f8d8:8280:856a:216:3eff:fee2:210c/128 | lxdbr0  | instance | true | 00:16:3e:e2:21:0c |
+-------------------------------------------+--------------------------------------------+---------+----------+------+-------------------+
| /1.0/instances/u2?project=bar             | fd42:f8d8:8280:856a:216:3eff:fef8:2925/128 | lxdbr0  | instance | true | 00:16:3e:f8:29:25 |
+-------------------------------------------+--------------------------------------------+---------+----------+------+-------------------+
| /1.0/instances/u2?project=bar             | 10.166.99.7/32                             | lxdbr0  | instance | true | 00:16:3e:f8:29:25 |
+-------------------------------------------+--------------------------------------------+---------+----------+------+-------------------+
| /1.0/instances/u1?project=foo             | 10.166.99.231/32                           | lxdbr0  | instance | true | 00:16:3e:e2:21:0c |
+-------------------------------------------+--------------------------------------------+---------+----------+------+-------------------+
```

### Information to attach

markylaing:
Checks the URLs returned by the network allocations endpoint. This required jumping through a few feature flag related hoops (see comments).

I've confirmed the tests are failing before this change and passing afterwards.

Closes #15855

### DIFF
--- a/lxd/network_allocations.go
+++ b/lxd/network_allocations.go
@@ -80,7 +80,7 @@ func networkAllocationsGet(d *Daemon, r *http.Request) response.Response {
 	s := d.State()
 
 	requestProjectName := request.ProjectParam(r)
-	effectiveProjectName, _, err := project.NetworkProject(d.State().DB.Cluster, requestProjectName)
+	effectiveProjectName, _, err := project.NetworkProject(s.DB.Cluster, requestProjectName)
 	if err != nil {
 		return response.SmartError(err)
 	}
@@ -91,7 +91,7 @@ func networkAllocationsGet(d *Daemon, r *http.Request) response.Response {
 	allProjects := shared.IsTrue(request.QueryParam(r, "all-projects"))
 
 	var projectNames []string
-	err = d.db.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
 		// Figure out the projects to retrieve.
 		if !allProjects {
 			projectNames = []string{effectiveProjectName}
@@ -156,7 +156,7 @@ func networkAllocationsGet(d *Daemon, r *http.Request) response.Response {
 
 		var networkNames []string
 
-		err := d.db.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+		err := s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
 			var err error
 
 			networkNames, err = tx.GetNetworks(ctx, projectName)
@@ -173,7 +173,7 @@ func networkAllocationsGet(d *Daemon, r *http.Request) response.Response {
 				continue
 			}
 
-			n, err := network.LoadByName(d.State(), projectName, networkName)
+			n, err := network.LoadByName(s, projectName, networkName)
 			if err != nil {
 				return response.SmartError(fmt.Errorf("Failed loading network %q in project %q: %w", networkName, projectName, err))
 			}
@@ -241,7 +241,7 @@ func networkAllocationsGet(d *Daemon, r *http.Request) response.Response {
 
 			var forwards map[int64]*api.NetworkForward
 
-			err = d.db.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+			err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
 				forwards, err = tx.GetNetworkForwards(ctx, n.ID(), false)
 
 				return err
@@ -271,7 +271,7 @@ func networkAllocationsGet(d *Daemon, r *http.Request) response.Response {
 
 			var loadBalancers map[int64]*api.NetworkLoadBalancer
 
-			err = d.db.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+			err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
 				loadBalancers, err = tx.GetNetworkLoadBalancers(ctx, n.ID(), false)
 
 				return err
--- a/lxd/network_allocations.go
+++ b/lxd/network_allocations.go
@@ -126,7 +126,21 @@ func networkAllocationsGet(d *Daemon, r *http.Request) response.Response {
 
 	result := make([]api.NetworkAllocations, 0)
 
-	userHasPermission, err := s.Authorizer.GetPermissionChecker(r.Context(), auth.EntitlementCanView, entity.TypeNetwork)
+	canViewNetwork, err := s.Authorizer.GetPermissionChecker(r.Context(), auth.EntitlementCanView, entity.TypeNetwork)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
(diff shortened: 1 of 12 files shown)


lxd/network_allocations.go
```diff
@@ -155,7 +193,7 @@ func networkAllocationsGet(d *Daemon, r *http.Request) response.Response {
 
 		// Get all the networks, their attached instances, their network forwards and their network load balancers.
 		for _, networkName := range networkNames {
-			if !userHasPermission(entity.NetworkURL(authCheckProjectName, networkName)) {
+			if !canViewNetwork(entity.NetworkURL(authCheckProjectName, networkName)) {
```
tomponline, line 172: this should be a separate commit right as is not related to the lease filtering fix?

tomponline: oh wait, you renamed the check for clarity, fine.

lxd/network_allocations.go
```diff
@@ -126,11 +126,49 @@ func networkAllocationsGet(d *Daemon, r *http.Request) response.Response {
...
 	if err != nil {
 		return response.SmartError(err)
 	}
 
+	canViewInstance, err := s.Authorizer.GetPermissionChecker(r.Context(), auth.EntitlementCanView, entity.TypeInstance)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	// If a project query parameter is passed, but the given project does not have `features.networks` enabled, then
```
tomponline: Thanks for the comment.

How would you feel about introducing a `s.Authorizer.GetPermissionCheckWithoutEffectiveProject()` or similar that we could use here for the `canViewInstance` checker (or add an argument to `GetPermissinChecker` to do the same)?

And do we need to use `canViewIgnoreEffectiveProject` for the `canViewNetwork` check? As wouldn't all networks be in the effective project if using `project=foo` where `foo` has `features.networks=false`?

markylaing: I'll play with it a bit more. I initially changed the authorizer to stop getting the effective project from the request context and instead expect an "overrideProject" argument, but changing the call sites became quite large.

For the second question: I wasn't 100% sure on this. The networks that are checked with this function are derived from the network lease here https://github.com/canonical/lxd/blob/ce8fa3309da16a2d86af9f0845200ee819459d5c/lxd/network_allocations.go#L200-L201 - where the lease type is "uplink". So I wasn't certain if these networks would be in the same project.

markylaing: > How would you feel about introducing a s.Authorizer.GetPermissionCheckWithoutEffectiveProject() or similar that we could use here for the canViewInstance checker (or add an argument to GetPermissinChecker to do the same)?

This seems to be the best worst option.

tomponline: > For the second question: I wasn't 100% sure on this. The networks that are checked with this function are derived from the network lease here

For the ovn and bridge network's `Leases()` function they both have a comment:

> // If requested project matches network's project then include gateway...

https://github.com/canonical/lxd/blob/main/lxd/network/driver_bridge.go#L3590-L3635

https://github.com/canonical/lxd/blob/main/lxd/network/driver_ovn.go#L5362-L5376

So I think as we're passing empty string to `Leases()` in this function we're OK.

This is most likely there to handle the `lxc network list-leases` use-case instead.

markylaing: Great thank :) Updated PR to add new methods to the auth interface that ignore the effective project and have used those methods only for the instance auth check

lxd/network_allocations.go
```diff
@@ -126,7 +126,12 @@ func networkAllocationsGet(d *Daemon, r *http.Request) response.Response {
 
 	result := make([]api.NetworkAllocations, 0)
 
-	userHasPermission, err := s.Authorizer.GetPermissionChecker(r.Context(), auth.EntitlementCanView, entity.TypeNetwork)
+	canViewNetwork, err := s.Authorizer.GetPermissionChecker(r.Context(), auth.EntitlementCanView, entity.TypeNetwork)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	canViewInstanceIgnoringEffectiveProject, err := s.Authorizer.GetPermissionCheckerWithoutEffectiveProject(r.Context(), auth.EntitlementCanView, entity.TypeInstance)
```
tomponline, line 143: It'd be good to have a comment here explaining why we are using this.

markylaing: Ah yeah I'll reword the comment from the commit message.

markylaing: Added a comment with examples.

markylaing:
@tomponline this is a pretty weird unit test failure: https://github.com/canonical/lxd/actions/runs/16020208379/job/45195111146?pr=15907#step:25:908

This is checking that when deleting an instance, values in `instance_config` are also deleted via cascade. The test sets up a test DQLite database. I'm wondering if there's been a regression. https://github.com/canonical/dqlite/pull/778 seems a likely candidate

tomponline:
> @tomponline this is a pretty weird unit test failure: https://github.com/canonical/lxd/actions/runs/16020208379/job/45195111146?pr=15907#step:25:908
> 
> This is checking that when deleting an instance, values in `instance_config` are also deleted via cascade. The test sets up a test DQLite database. I'm wondering if there's been a regression. [canonical/dqlite#778](https://github.com/canonical/dqlite/pull/778) seems a likely candidate

Yep thats the regression indeed.

PR #15913, opened 2025-07-01
doc: ignore anchor when checking maas link (stable-5.0)

minaelee:
Unresolved Sphinx linkcheck bug with anchors (as filed [here](https://github.com/sphinx-doc/sphinx/issues/12688) and [here](https://github.com/sphinx-doc/sphinx/issues/9255)) sometimes causes our link to maas.io to fail with:

```
broken [https://maas.io/docs/how-to-manage-machines#p-9078-use-lxd-vms](https://maas.io/docs/how-to-manage-machines#p-9078-use-lxd-vms) - Anchor 'p-9078-use-lxd-vms' not found
```

...even though the anchor exists and works when going directly to `https://maas.io/docs/how-to-manage-machines#p-9078-use-lxd-vms`, and linkcheck does not fail it consistently. This is an intermittent upstream bug; this PR works around it by adding the maas.io page to `linkcheck_anchors_ignore_for_url` in the Sphinx configuration file, which should cause the linkcheck to check that the page exists but ignore checking for the anchor.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -164,6 +164,11 @@ linkcheck_ignore = [
     'https://www.schlachter.tech/solutions/pongo2-template-engine/'
 ]
 
+# Ignore anchors for these URLs in linkcheck, but still check the URLs themselves
+linkcheck_anchors_ignore_for_url = [
+    'https://maas.io/docs/how-to-manage-machines',
+]
+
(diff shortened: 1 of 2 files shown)


tomponline (approved):
Thanks!

PR #15916, opened 2025-07-01
doc: remove unnecessary step to remove hold to refresh snap

minaelee:
The documentation on holding snap updates provided steps to `--unhold` the LXD snap before manually updating/refreshing, then reinstating the hold. This PR removes these steps as they are not necessary; you can manually update a snap while it is held.


tomponline (approved):
Cool

PR #15918, opened 2025-07-01
doc: add migration.stateful restriction for local volumes

minaelee:
This PR documents an additional restriction for attaching a custom storage volume: you cannot attach a storage volume from a local pool to an instance that has `migration.stateful` set to `true`. You must set it to `false` first, which results in the instance becoming ineligible for live migration.


tomponline (approved):

roosterfish:
Thanks! I have created https://github.com/canonical/microcloud/pull/846 to cross-link this also from the MicroCloud docs.

Issue #15038, opened 2025-02-20
`Unable to locate a VM UEFI firmware` at `lxd.activate` time


simondeziel:
On `amd64` machine with `latest/edge` and KVM supported and functional VMs, I get this warning:

```
# snap stop lxd; journalctl -f | grep -F 'lxd.activate[' & snap start lxd; lxc ls >/dev/null; kill %1
...
Feb 20 12:50:27 sdeziel-lemur lxd.activate[942949]: ==> Setting LXD user socket ownership
Feb 20 12:50:27 sdeziel-lemur lxd.activate[942949]: ==> Checking if LXD needs to be activated
Started.
Feb 20 12:50:27 sdeziel-lemur lxd.activate[943016]: time="2025-02-20T12:50:27-05:00" level=error msg="Unable to run feature checks during QEMU initialization: Unable to locate a VM UEFI firmware"
Feb 20 12:50:27 sdeziel-lemur lxd.activate[943016]: time="2025-02-20T12:50:27-05:00" level=warning msg="Instance type not operational" driver=qemu err="QEMU failed to run feature checks" type=virtual-machine
```

I wonder if it could be due to the snap only shipping a `debug` variant of the `CODE`:
```
$ find /snap/lxd/current/ -iname '*ovmf*'
/snap/lxd/current/share/qemu/OVMF_CODE.4MB.debug.fd
/snap/lxd/current/share/qemu/OVMF_CODE.4MB.fd
/snap/lxd/current/share/qemu/OVMF_VARS.4MB.fd
/snap/lxd/current/share/qemu/OVMF_VARS.4MB.ms.fd
```

And the code not looking for the `debug.fd` `CODE`: https://github.com/canonical/lxd/blob/main/lxd/instance/drivers/edk2/edk2.go#L182-L185

@tomponline on MM said:
> we just need to disable that during activation check reallly

mihalicyn:
I believe it's because here https://github.com/canonical/lxd-pkg-snap/blob/4d919878699c4ee5c3ff9ea73fe2e153dab15214/snapcraft/commands/daemon.activate#L112:
```
if ! "${LXD}" activateifneeded; then                      <<< HERE
    echo "====> Activation check failed, forcing activation"
    nsenter -t 1 -m systemctl start snap."${SNAP_INSTANCE_NAME}".daemon
fi
```

we call LXD binary without `LXD_QEMU_FW_PATH` environment variable set like we do in `daemon.start`.

[...]

PR #15919, opened 2025-07-02
Daemon: Don't request full `State` in `activateifneeded` command to avoid running instance driver feature checks

tomponline:
By avoiding calling `d.State()` this PR avoids running unnecessary instance driver feature checks during `activateifneeded` command, which in turn avoids doing unnecessary work and logging errors like:

```
Unable to locate a VM UEFI firmware
```

Fixes https://github.com/canonical/lxd/issues/15038

### DIFF
--- a/lxd/main_activateifneeded.go
+++ b/lxd/main_activateifneeded.go
@@ -74,16 +74,15 @@ func (c *cmdActivateifneeded) run(cmd *cobra.Command, args []string) error {
 	d.db.Node = db.DirectAccess(sqldb)
 
 	// Load the configured address from the database
-	var localConfig *node.Config
 	err = d.db.Node.Transaction(context.TODO(), func(ctx context.Context, tx *db.NodeTx) error {
-		localConfig, err = node.ConfigLoad(ctx, tx)
+		d.localConfig, err = node.ConfigLoad(ctx, tx)
 		return err
 	})
 	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


lxd/main_activateifneeded.go
```diff
@@ -116,36 +129,45 @@ func (c *cmdActivateifneeded) run(cmd *cobra.Command, args []string) error {
...
+	// core.https_address is unset (as there is an early check above) this is not inefficient because we will
+	// only get to this part for standalone servers.
+	instances, err := instance.LoadNodeAll(s, instancetype.Any)
 	if err != nil {
 		return err
 	}
 
 	for _, inst := range instances {
 		if instanceShouldAutoStart(inst) {
 			logger.Debugf("Daemon has auto-started instances, activating...")
```
simondeziel, line 150: It's too bad we don't have a linter flagging those `logger.XXXf()` calls lacking format specifier cause we should be using `logger.XXX()`.

simondeziel (approved):
Very nice!

lxd/main_activateifneeded.go
simondeziel: This seems easy enough to add a check so I'll do just that.

tomponline: Cool thanks

simondeziel: https://github.com/canonical/lxd/pull/15924

PR #15920, opened 2025-07-02
Add util to get both "project" and "all-projects" parameters

markylaing:
I noticed that multiple handlers were performing close to identical logic to get the project and all-projects parameters - returning an error if both are set.

This just factors it out into one place.

### DIFF
--- a/lxd/db/cluster/projects.go
+++ b/lxd/db/cluster/projects.go
@@ -236,3 +236,29 @@ func InitProjectWithoutImages(ctx context.Context, tx *sql.Tx, project string) e
 	_, err = tx.Exec(stmt, defaultProfileID)
 	return err
 }
+
+// GetAllProjectsConfig returns a map of project name to config map.
+func GetAllProjectsConfig(ctx context.Context, tx *sql.Tx) (map[string]map[string]string, error) {
+	projectConfigs := make(map[string]map[string]string)
+	err := query.Scan(ctx, tx, "SELECT projects.name, projects_config.key, projects_config.value FROM projects JOIN projects_config ON projects.id = projects_config.project_id", func(scan func(dest ...any) error) error {
+		var projectName, key, value string
+		err := scan(&projectName, &key, &value)
+		if err != nil {
+			return err
+		}
+
+		projectConfig, ok := projectConfigs[projectName]
+		if !ok {
+			projectConfigs[projectName] = map[string]string{key: value}
+			return nil
+		}
+
+		projectConfig[key] = value
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return projectConfigs, nil
+}
--- a/lxd/images.go
+++ b/lxd/images.go
@@ -1767,8 +1767,6 @@ func doImagesGet(ctx context.Context, tx *db.ClusterTx, recursion bool, projectN
 //	  "500":
 //	    $ref: "#/responses/InternalServerError"
 func imagesGet(d *Daemon, r *http.Request) response.Response {
-	projectName := request.ProjectParam(r)
-	allProjects := shared.IsTrue(r.FormValue("all-projects"))
 	filterStr := r.FormValue("filter")
 
 	recursion := util.IsRecursionRequest(r)
@@ -1777,24 +1775,22 @@ func imagesGet(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
-	// ProjectParam returns default if not set
-	if allProjects && projectName != api.ProjectDefaultName {
-		return response.BadRequest(errors.New("Cannot specify a project when requesting all projects"))
-	}
-
-	s := d.State()
-	var effectiveProjectName string
-	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
-		var err error
-		effectiveProjectName, err = projectutils.ImageProject(ctx, tx.Tx(), projectName)
-		return err
-	})
+	projectName, allProjects, err := request.ProjectParams(r)
 	if err != nil {
 		return response.SmartError(err)
 	}
 
-	reqInfo := request.SetupContextInfo(r)
-	reqInfo.EffectiveProjectName = effectiveProjectName
+	s := d.State()
+	if !allProjects {
+		reqInfo := request.SetupContextInfo(r)
+		err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+			reqInfo.EffectiveProjectName, err = projectutils.ImageProject(ctx, tx.Tx(), projectName)
+			return err
+		})
+		if err != nil {
+			return response.SmartError(err)
+		}
+	}
 
 	// If the caller is not trusted, we only want to list public images in the default project.
 	trusted := auth.IsTrusted(r.Context())
--- a/lxd/networks.go
+++ b/lxd/networks.go
@@ -321,7 +321,7 @@ func networksGet(d *Daemon, r *http.Request) response.Response {
 
 	// Get list of actual network interfaces on the host if the effective project is default and the caller has permission.
 	var getUnmanagedNetworks bool
-	if reqInfo.EffectiveProjectName == api.ProjectDefaultName {
+	if reqInfo.EffectiveProjectName == api.ProjectDefaultName || allProjects {
 		err := s.Authorizer.CheckPermission(r.Context(), entity.ServerURL(), auth.EntitlementCanViewUnmanagedNetworks)
 		if err == nil {
 			getUnmanagedNetworks = true
@@ -342,9 +342,14 @@ func networksGet(d *Daemon, r *http.Request) response.Response {
 				continue
 			}
 
+			unmanagedNetworkProject := requestProjectName
+			if allProjects {
+				unmanagedNetworkProject = api.ProjectDefaultName
+			}
+
 			// Append to the list of networks if a managed network of same name doesn't exist.
-			if !slices.Contains(networks[managed][requestProjectName], iface.Name) {
-				networks[unmanaged][requestProjectName] = append(networks[unmanaged][requestProjectName], iface.Name)
+			if !slices.Contains(networks[managed][unmanagedNetworkProject], iface.Name) {
+				networks[unmanaged][unmanagedNetworkProject] = append(networks[unmanaged][unmanagedNetworkProject], iface.Name)
 			}
 		}
 	}
--- a/lxd/networks.go
+++ b/lxd/networks.go
@@ -287,6 +287,7 @@ func networksGet(d *Daemon, r *http.Request) response.Response {
 		unmanaged
 	)
 
+	projectConfigs := make(map[string]map[string]string)
 	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
 		networks[managed] = map[string][]string{}
 		networks[unmanaged] = map[string][]string{}
@@ -294,6 +295,11 @@ func networksGet(d *Daemon, r *http.Request) response.Response {
 		if allProjects {
 			// Get list of managed networks from all projects.
 			networks[managed], err = tx.GetNetworksAllProjects(ctx)
+			if err != nil {
+				return err
+			}
+
+			projectConfigs, err = dbCluster.GetAllProjectsConfig(ctx, tx.Tx())
 			if err != nil {
 				return err
 			}
@@ -363,7 +369,14 @@ func networksGet(d *Daemon, r *http.Request) response.Response {
 				if !recursion {
 					resultString = append(resultString, api.NewURL().Path(version.APIVersion, "networks", networkName).String())
 				} else {
-					net, err := doNetworkGet(s, r, s.ServerClustered, projectName, reqProject.Config, networkName)
+					var projectConfig map[string]string
+					if allProjects {
+						projectConfig = projectConfigs[projectName]
+					} else {
+						projectConfig = reqProject.Config
+					}
+
+					net, err := doNetworkGet(s, r, s.ServerClustered, projectName, projectConfig, networkName)
 					if err != nil {
 						continue
 					}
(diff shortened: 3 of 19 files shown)


markylaing:
This refactor exploded slightly because of the `GET /1.0/networks` handler.

The handler always gets an effective project - regardless of whether the request is for all projects or not: https://github.com/canonical/lxd/blob/684b4b77314ed8d9310405ca4acc3ff1ba2cc0c5/lxd/networks.go#L269-L280

So for "all projects" requests, the effective project is set in the request context to be the default project. Additionally, the `reqProject` returned by `project.NetworkProject` is always the default project.

Later, we only get the unmanaged networks if the effective project is the default project: https://github.com/canonical/lxd/blob/684b4b77314ed8d9310405ca4acc3ff1ba2cc0c5/lxd/networks.go#L323-L331

This doesn't really make sense for all projects requests, but works because the effective project is set to default.

When iterating over networks to expand the response, the call to `doNetworkGet` accepts a `projectConfig` parameter which is expected to be the configuration of the requested project: https://github.com/canonical/lxd/blob/684b4b77314ed8d9310405ca4acc3ff1ba2cc0c5/lxd/networks.go#L372

For all projects requests, this is always the configuration of the default project. `doNetworkGet` uses this configuration to disallow viewership of the network based on `restricted.*` config keys. The default project is unlikely to have network restrictions set because it is the only project that can contain non-virtualised networks.

I've modified this to pass in the configuration of the project that the network is contained in. But for non-virtualised networks this is the default project anyway.

Lastly, the `doNetworkGet` function overwrites the given project with the effective project set in the request context: https://github.com/canonical/lxd/blob/684b4b77314ed8d9310405ca4acc3ff1ba2cc0c5/lxd/networks.go#L985-L989

This means that `all-projects` requests for fine-grained users would always return only networks in the default project - it would fail to find any other networks.


Since project restrictions apply to networks that are shared from the default project, there is no security concern here and the only bug is listing networks from all projects as a fine-grained identity. However, due to the convoluted API handler this feels like a happy accident. We need to be very careful about the interaction between project restrictions, the identity type of the caller, and all projects requests. To catch things of this nature in future I've reworked the `projects_restrictions` tests to run tests for unix, TLS (unrestricted), TLS (restricted), and TLS (fine-grained) users.

I'm also in the process of adding tests for listing networks across projects when some networks are actually defined in other projects (since we now have OVN). Because currently these tests only return networks in the default project.

lxd/images.go
```diff
@@ -1777,24 +1775,23 @@ func imagesGet(d *Daemon, r *http.Request) response.Response {
...
 		return response.SmartError(err)
 	}
 
-	reqInfo := request.SetupContextInfo(r)
-	reqInfo.EffectiveProjectName = effectiveProjectName
+	s := d.State()
+	if !allProjects {
+		reqInfo := request.SetupContextInfo(r)
+		err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+			var err error
```
tomponline: no need for this line

lxd/db/cluster/projects.go
```diff
@@ -236,3 +236,29 @@ func InitProjectWithoutImages(ctx context.Context, tx *sql.Tx, project string) e
 	_, err = tx.Exec(stmt, defaultProfileID)
 	return err
 }
+
+// GetAllProjectsConfig returns a map of project name to config map.
+func GetAllProjectsConfig(ctx context.Context, tx *sql.Tx) (map[string]map[string]string, error) {
```
tomponline, line 241: Can you use `dbCluster.GetConfig(ctx, tx.Tx(), "project")`

tomponline: oh that keys it by project ID though hrm

lxd/images.go
markylaing: Removed.

lxd/db/cluster/projects.go
markylaing: Yeah it's not ideal for this use case unfortunately.

markylaing:
Not sure what's going on with the tests here. It's passing locally. Will come back to it after lunch :sandwich:

markylaing:
@tomponline in the end the problem was that I had changed the default project in the `localhost` remote and didn't clean up the change :facepalm: 

Tests should pass now and I can get to looking at the public image endpoints.

test/suites/projects.sh
```diff
@@ -967,7 +1059,7 @@ test_projects_restrictions() {
   # It's not possible to use forbidden low-level options
   ! lxc profile set default "raw.idmap=both 0 0" || false
   ! lxc init testimage c1 -c "raw.idmap=both 0 0" || false
-  ! lxc init testimage c1 -c volatile.uuid="foo" || false
+  ! lxc init testimage c1  -c volatile.uuid="foo" || false
```
kadinsayani:
```suggestion
  ! lxc init testimage c1 -c volatile.uuid="foo" || false
```

test/suites/projects.sh
```diff
@@ -978,7 +1070,7 @@ test_projects_restrictions() {
 
   # It's not possible to change low-level options
   ! lxc config set c1 "raw.idmap=both 0 0" || false
-  ! lxc config set c1 volatile.uuid="foo" || false
+  ! lxc config set c1  volatile.uuid="foo" || false
```
kadinsayani:
```suggestion
  ! lxc config set c1 volatile.uuid="foo" || false
```

test/suites/auth.sh
```diff
@@ -1145,6 +1152,65 @@ auth_project_features() {
...
+  fi
+
+  lxc config set network.ovn.northbound_connection "${LXD_OVN_NB_CONNECTION}"
+
+  # If the connection uses SSL we have more required environment variables.
+  # Set the client cert, key, and CA cert.
+  if [[ "${LXD_OVN_NB_CONNECTION}" =~ ^ssl: ]]; then
+    lxc config set network.ovn.client_cert="$(< "${LXD_OVN_NB_CLIENT_CRT_FILE}")"
+    lxc config set network.ovn.client_key="$(< "${LXD_OVN_NB_CLIENT_KEY_FILE}")"
+    lxc config set network.ovn.ca_cert="$(< "${LXD_OVN_NB_CA_CRT_FILE}")"
```
simondeziel: I wonder if there would be a benefit in doing those config changes in one go.
```suggestion
    lxc config set network.ovn.client_cert="$(< "${LXD_OVN_NB_CLIENT_CRT_FILE}")" network.ovn.client_key="$(< "${LXD_OVN_NB_CLIENT_KEY_FILE}")" network.ovn.ca_cert="$(< "${LXD_OVN_NB_CA_CRT_FILE}")"
```
Would it avoid configuring OVN multiple times or something like that?

test/suites/auth.sh
```diff
@@ -1145,6 +1152,65 @@ auth_project_features() {
...
+
+  # Clean up
+  lxc network delete "${uplink_network}"
+  ip link delete dummy0
+  lxc project delete foo
+  lxc config unset network.ovn.northbound_connection
+  if [[ "${LXD_OVN_NB_CONNECTION}" =~ ^ssl: ]]; then
+    lxc config unset network.ovn.client_cert
+    lxc config unset network.ovn.client_key
+    lxc config unset network.ovn.ca_cert
```
simondeziel: Heh, apparently we cannot batch `lxc config unset`. To do those all in one go we need:

```suggestion
    lxc config set network.ovn.client_cert="" network.ovn.client_key="" network.ovn.ca_cert=""
```
I've filled: https://github.com/canonical/lxd/issues/15933

lxd/request/parse.go
```diff
@@ -4,10 +4,26 @@ import (
 	"net/http"
 	"net/url"
 
+	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/logger"
 )
 
+// ProjectParams returns the project name and the value of the all projects query parameter. It returns a 400 Bad
+// Request error if both parameters are specified.
```
kadinsayani: I recently discovered you can link to documentation by surrounding functions and types with square brackets. Just a nit :)

```suggestion
// ProjectParams returns the project name and the value of the all projects query parameter. 
// It returns a [http.StatusBadRequest] wrapped by [api.StatusErrorf] if both parameters are specified.
```

lxd/request/parse.go
```diff
@@ -4,10 +4,26 @@ import (
...
+func ProjectParams(r *http.Request) (string, bool, error) {
+	projectName := QueryParam(r, "project")
+	allProjects := shared.IsTrue(QueryParam(r, "all-projects"))
+
```
kadinsayani, line 17:
```suggestion

// A requested project name is only valid for project specific requests.
```

lxd/request/parse.go
markylaing: It should link without the square brackets I think. I'll update the comment.

test/suites/auth.sh
markylaing: Thanks :)

test/suites/projects.sh
markylaing: Hmm not sure what I did there :thinking:

test/suites/auth.sh
markylaing: Ah yeah good point. I was also thinking of extracting some OVN functions to a helper to clean it up as this is copy/paste from the `network_ovn` suite.

lxd/request/parse.go
markylaing: Nope you're right, it needs the square brackets https://pkg.go.dev/github.com/fluhus/godoc-tricks#hdr-Documentation_Links-Links

I got confused by the syntax highlighting in my editor.

markylaing: Created this issue: https://github.com/canonical/lxd/issues/15937

I've just added identifiers like `http.StatusBadRequest` to godoc comments in the past thinking they were linked automagically so it would be good to have a look for them and update to square brackets.

Will update this one in this PR of course.

test/suites/auth.sh
markylaing: I've gone ahead and added some OVN utils under `includes` :)

test/includes/ovn.sh
```diff
@@ -0,0 +1,57 @@
...
+  fi
+
+  # Already configured.
+  if [ "$(lxc config get network.ovn.northbound_connection)" != "" ]; then
+    return
+  fi
+
+  # Set base connection
+  lxc config set network.ovn.northbound_connection "${LXD_OVN_NB_CONNECTION}"
+
```
simondeziel: That can go as you've put it in the if and else branches.

test/includes/ovn.sh
```diff
@@ -0,0 +1,57 @@
...
+unset_ovn_configuration() {
+  if ! ovn_enabled; then
+    return
+  fi
+
+  # Unset the config keys.
+  lxc config unset network.ovn.northbound_connection
+  lxc config unset network.ovn.client_cert
+  lxc config unset network.ovn.client_key
+  lxc config unset network.ovn.ca_cert
```
simondeziel: Should we set all 4 keys to empty values to do it in one go?

test/suites/auth.sh
```diff
@@ -1145,6 +1152,52 @@ auth_project_features() {
   lxc project delete blah
 }
 
+auth_ovn() {
+  remote="${1}"
+
+  if ! ovn_enabled; then
+    echo "OVN not configured. Skipping OVN tests..."
```
simondeziel: We use the form `echo "==> SKIP: ..."` elsewhere.

test/suites/network_ovn.sh
```diff
@@ -1,6 +1,6 @@
 test_network_ovn() {
-  if [ -z "${LXD_OVN_NB_CONNECTION:-}" ]; then
-    echo "OVN northbound connection not set. Skipping OVN tests..."
+  if ! ovn_enabled; then
+    echo "OVN not configured. Skipping OVN tests..."
```
simondeziel: We use the form `echo "==> SKIP: ..."` elsewhere.

kadinsayani (approved):
LGTM, thanks!

test/includes/ovn.sh
markylaing: Fixed, thanks

test/includes/ovn.sh
markylaing: Fixed, thanks

test/suites/auth.sh
markylaing: Fixed, thanks

test/suites/network_ovn.sh
markylaing: Fixed, thanks

markylaing:
@tomponline this one is ready to go when you have time, thanks.

lxd/networks.go
```diff
@@ -258,26 +258,20 @@ func networksGet(d *Daemon, r *http.Request) response.Response {
...
 	if err != nil {
 		return response.SmartError(err)
 	}
 
 	reqInfo := request.SetupContextInfo(r)
-	reqInfo.EffectiveProjectName = effectiveProjectName
+	var reqProject *api.Project
+	if !allProjects {
+		// Project specific requests require an effective project, when "features.networks" is enabled this is the requested project, otherwise it is the default project.
+		reqInfo.EffectiveProjectName, reqProject, err = project.NetworkProject(s.DB.Cluster, requestProjectName)
```
tomponline, line 270: Out of interest, should this not be done in the access control layer, I thought this was normally where EffectiveProjectName was set?

markylaing: No, the effective project name must be set before the access control layer. This is so the authorizer knows to overwrite the project in the access check.

tomponline: No I meant in the access handler, such as `addNetworkZoneDetailsToRequestContext`?

tomponline: Is it because this is in a list endpoint that it cant add effective project in the access handler?

markylaing: Ah I see, sorry.

I think the list endpoints tend to use `allowProjectResourceList`, which does the initial logic disallowing restricted TLS clients from using the `all-projects` query parameter. Then there's nowhere else to put access logic except the main handler. This could be handled better for sure. I'll look into it.

tomponline: thanks!

PR #15921, opened 2025-07-02
Network: Fix state endpoint for OVN networks without port binding

roosterfish:
This seems to affect both 5.21 and edge.

When running `lxc network info <name>` on an OVN network which doesn't have a `volatile.network.ipv4.address` and therefore no port binding and active chassis, the API returns:

`Error: Failed to run: ovn-sbctl --timeout=10 --db ssl:10.87.144.48:6642 -c /proc/self/fd/3 -p /proc/self/fd/4 -C /proc/self/fd/5 get Chassis  hostname: exit status 1 (ovn-sbctl: no row "" in table Chassis)` 

as it tries to fetch the chassis with an `""` empty name. 

The PR ensures the chassis is only looked up in case the network actually has a port binding.
This probably anyway is an edge case I just ran across when initializing a MicroCloud and not providing any IP details for the uplink network.


roosterfish:
To not cause any API break I left the `OVN:       &api.NetworkStateOVN{Chassis: chassis},` which is returned by the state endpoint. In case of no chassis the string is empty but we could argue it anyway always returned an error so we can instead of returning an empty string also return `nil` which would also cause the `lxc` client to not display it:

```bash
root@micro01:~# lxc network info default
Name: default
MAC address: 
MTU: 1442
State: up
Type: broadcast

IP addresses:
  inet	10.128.172.1/24 (link)
  inet6	fd42:d1ff:deb9:6d17::1/64 (link)

Network usage:
  Bytes received: 0B
  Bytes sent: 0B
  Packets received: 0
  Packets sent: 0

OVN:
  Chassis: 
```

lxd/network/driver_ovn.go
```diff
@@ -146,9 +146,15 @@ func (n *ovn) State() (*api.NetworkState, error) {
...
 	}
 
-	chassis, err := client.GetLogicalRouterPortActiveChassisHostname(n.getRouterExtPortName())
-	if err != nil {
-		return nil, err
+	var chassis string
+
+	// The logical router's port active chassis is only available in case the network
+	// has an IP on the uplink network assigned.
+	if n.config["volatile.network.ipv4.address"] != "" {
```
tomponline: in that scenario can this occur? When the ovn network only uses IPv6?

lxd/network/openvswitch/ovn.go
```diff
@@ -2372,6 +2372,10 @@ func (o *OVN) GetLogicalRouterPortActiveChassisHostname(ovnRouterPort OVNRouterP
 		return "", err
 	}
 
+	if chassisID == "" {
+		return "", fmt.Errorf("No port binding for the router port %q", ovnRouterPort)
```
tomponline: Incus use a somewhat more concise & friendly error for the end user:

https://github.com/lxc/incus/blob/a082b02285925ee66d95bcc0fe2ba65718438990/internal/server/network/ovn/ovn_sb_actions.go#L28

lxd/network/driver_ovn.go
roosterfish: Thanks, we actually have to check if neither the IPv4 nor the IPv6 address is present.

It happens if the uplink doesn't have `ipv(4|6).gateway` set.

Another option would be to check if the UPLINK network has a `ipv(4|6).gateway` as this is the [root cause](https://github.com/canonical/lxd/blob/main/lxd/network/driver_ovn.go#L1295-L1304) for getting a `volatile.network.ipv(4|6).address` allocated on the uplink for the network.

roosterfish: Just found there are constants for `"volatile.network.ipv(4|6).address" so let me use those instead.

roosterfish: Fixed.

lxd/network/openvswitch/ovn.go
roosterfish: Good call, changed the message and added proper credit to the commit message.

PR #15927, opened 2025-07-03
Storage: Skip custom vol updates early

roosterfish:
This PR is based on a comment in https://github.com/canonical/lxd/pull/15523#discussion_r2182670606.

If neither a custom volume's config nor its description got modified the function can return early.
The exact same applies when trying to update custom vol snaps.


roosterfish:
Just found the same also applies for the `UpdateCustomVolumeSnapshot` func, I'll add one more commit in a sec.

tomponline (approved):

Issue #15894, opened 2025-06-30
The `fanotify` mechanism does not notice dynamic removal of underlying devices


simondeziel:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

24.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
n/a
```

### Output of "lxc info" or system info if it fails

```shell
n/a
```

### Issue description

When running the `container_devices_unix` test, a set of tests are ran 3 times with different notify drivers. When using the `fanotify` one, live removing the underlying devices in the host does not trigger cleanup in the container:

```
+ lxc config device set ct29140 test-dev-dynamic source=/tmp/lxd-test.tmp.zVO3/dev/testdev-ct29140/subdev
++ timeout --foreground 120 /root/go/bin/lxc config device set ct29140 test-dev-dynamic source=/tmp/lxd-test.tmp.zVO3/dev/testdev-ct29140/subdev --verbose
INFO   [2025-06-30T12:13:02Z] Unwatching path                               driver=fanotify path=/tmp/lxd-test.tmp.zVO3/dev/testdev-ct29140
INFO   [2025-06-30T12:13:02Z] Watching path                                 driver=fanotify path=/tmp/lxd-test.tmp.zVO3/dev/testdev-ct29140/subdev
+ lxc exec ct29140 -- mount
++ timeout --foreground 120 /root/go/bin/lxc exec ct29140 --verbose --force-local -- mount
+ grep -F /tmp/testdev
+ lxc exec ct29140 -- test -e /tmp/testdev
++ timeout --foreground 120 /root/go/bin/lxc exec ct29140 --verbose --force-local -- test -e /tmp/testdev
+ test -e /tmp/lxd-test.tmp.zVO3/2eB/devices/ct29140/unix.test--dev--dynamic.tmp-testdev
+ mkdir /tmp/lxd-test.tmp.zVO3/dev/testdev-ct29140
+ mknod /tmp/lxd-test.tmp.zVO3/dev/testdev-ct29140/subdev b 0 0
+ sleep 2
+ lxc exec ct29140 -- mount
++ timeout --foreground 120 /root/go/bin/lxc exec ct29140 --verbose --force-local -- mount
+ grep /tmp/testdev
tmpfs on /tmp/testdev type tmpfs (rw,relatime,size=7340032k,mode=751,inode64)
[...]

PR #15931, opened 2025-07-03
test/suites/container_devices_unix: do not completely skip tests with `fanotify`

simondeziel:
Only skip the part that is known broken ATM due to https://github.com/canonical/lxd/issues/15894


tomponline (approved):
Nice

PR #15936, opened 2025-07-04
daemon: Update Sec-Fetch-Site checks (stable-5.0)

tomponline:


lxd/daemon.go
```diff
@@ -654,7 +654,7 @@ func (d *Daemon) createCmd(restAPI *mux.Router, version string, c APIEndpoint) {
 
 			// Protect against CSRF when using LXD-UI with browser that supports Fetch metadata.
 			// Deny Sec-Fetch-Site when set to cross-origin or same-site.
```
roosterfish: Comment still says `cross-origin`.

tomponline: thanks fixed

tomponline: @roosterfish it needed a custom backport compared to 5.21

PR #15939, opened 2025-07-04
Storage: Add more context to errors

roosterfish:
This PR adds more context to errors returned from the backend's `GenerateCustomVolumeBackupConfig` func.
It's a follow up on https://github.com/canonical/lxd/pull/15523#discussion_r2182616424.


roosterfish:
I have explicitly moved all the errors to say "backup config" too to name it consistently.

lxd/migrate_storage_volumes.go
```diff
@@ -100,7 +100,7 @@ func (s *migrationSourceWs) DoStorage(state *state.State, projectName string, po
 
 	srcConfig, err := pool.GenerateCustomVolumeBackupConfig(projectName, volName, !s.volumeOnly, migrateOp)
 	if err != nil {
-		return fmt.Errorf("Failed generating volume migration config: %w", err)
+		return fmt.Errorf("Failed generating backup config of volume %q in pool %q and project %q: %w", volName, poolName, projectName, err)
```
tomponline: would you mind changing the word backup to migration here so that we can more context of where this specific error is occurring?

roosterfish: I was thinking a bit here and then picked the same message as there really is no "migration config" or "refresh config" (on one of the other errors) and to keep it consistent when we talk about the backup config.

But If you rather want to keep the context I'll modify it. Maybe it's more confusing to mention backup when you do refresh/migration. Mh.

tomponline: > Maybe it's more confusing to mention backup when you do refresh/migration.

Thats what i was thinking indeed.

roosterfish: Ok sure I'll push a change shortly.

roosterfish: Fixed.

PR #15940, opened 2025-07-04
Storage: Improve pool name and volume name validation

tomponline:


lxd/storage/drivers/utils.go
```diff
@@ -895,3 +896,45 @@ func remoteVolumeMapLock(connectorName string, driverName string) (locking.Unloc
...
+
+// ValidVolumeName validates a volume name.
+func ValidVolumeName(volumeName string) error {
+	if volumeName == "" {
+		return errors.New("Cannot be empty")
+	}
+
+	if strings.Contains(volumeName, "\\") {
+		return errors.New("Cannot contain backslashes")
+	}
```
simondeziel, line 939: Don't we need the same in `ValidPoolName()`?

simondeziel: Also for both, should we not block `..` like some open coded validation did before?

lxd/daemon_storage.go
```diff
@@ -146,14 +146,14 @@ func daemonStorageSplitVolume(volume string) (poolName string, volumeName string
...
-	// Validate pool name.
-	if strings.Contains(poolName, "\\") || strings.Contains(poolName, "..") {
-		return "", "", fmt.Errorf("Invalid pool name: %q", poolName)
+	err = storageDrivers.ValidPoolName(poolName)
+	if err != nil {
+		return "", "", fmt.Errorf("Invalid pool name %q: %w", poolName, err)
 	}
 
-	// Validate volume name.
-	if strings.Contains(volumeName, "\\") || strings.Contains(volumeName, "..") {
```
simondeziel, line 155: The `..` should still be banned, isn't it?

lxd/storage/drivers/utils.go
```diff
@@ -903,7 +903,11 @@ func ValidPoolName(poolName string) error {
 		return errors.New("Cannot be empty")
 	}
 
-	if strings.Contains(poolName, shared.SnapshotDelimiter) {
+	if strings.HasPrefix(poolName, "-") {
```
simondeziel, line 910: Should we do the same for `ValidVolumeName()`?

lxd/storage/drivers/utils.go
```diff
@@ -903,7 +903,11 @@ func ValidPoolName(poolName string) error {
...
+		return errors.New("Cannot start with a hyphen")
+	}
+
+	if strings.Contains(poolName, "/") {
```
simondeziel: Should we do the same for `ValidVolumeName()`?

test/suites/storage.sh
```diff
@@ -7,10 +7,17 @@ test_storage() {
...
-  # edit storage and pool description
   local storage_pool storage_volume
   storage_pool="lxdtest-$(basename "${LXD_DIR}")-pool"
   storage_volume="${storage_pool}-vol"
+
+  # Check for pool name validation
+  ! lxc query -X POST --wait /1.0/storage-pools --data '{\"driver\":\"'"${lxd_backend}"'\"}' || false
+  ! lxc query -X POST --wait /1.0/storage-pools --data '{\"name\":\"-'"${storage_pool}"'\", \"driver\":\"'"${lxd_backend}"'\"}' || false
+  ! lxc storage create "${storage_pool}/" "$lxd_backend" || false
+  ! lxc storage create "${storage_pool} " "$lxd_backend" || false
```
simondeziel: Nit pick: this new occurrence could use `{}`: `${lxd_backend}`
I know we are absolutely not consistent about this rule anywhere but since you've done it for the other variables :)

lxd/storage/drivers/utils.go
tomponline: > Don't we need the same in `ValidPoolName()`?

Thats outside of the scope of this change and may introduce further potential breakages.

lxd/daemon_storage.go
tomponline: The volume name policy should be applied consistently throughout he code base.
But only here was ".." blocked.
I'm not sure if blocking `/` is sufficient for `..` to not be a concern.

lxd/storage/drivers/utils.go
tomponline: Its outside of the scope of this particular PR as I'm aiming to fix the specific issue without introducing the risk of too many potential regressions.

lxd/storage/drivers/utils.go
tomponline: yes we do with `if strings.Contains(volumeName, shared.SnapshotDelimiter) {`

lxd/storage/drivers/utils.go
simondeziel: OK but in the long run blocking `-` would be a safe thing to do considering how easy to turn the name into a command argument.

lxd/storage/drivers/utils.go
tomponline: oh right i already fixed this once in an earlier commit and then git reintroduced it again later

test/suites/storage.sh
tomponline: I hate bash so much, but will fix.

lxd/daemon_storage.go
simondeziel: What about a pool named *just* `..`?

lxd/storage/drivers/utils.go
tomponline: It needs a full impact audit though - i wont include it in this PR, but if I get chance i'll evaluate it.

Whilst im fully onboard with the change, we do have to consider regressions with any policy change.

lxd/daemon_storage.go
tomponline: yeah ill add a check for that and volume name as should be very low impact.

simondeziel:
> **lxd/storage/drivers/utils.go:900**
> 
>     * Consider adding unit tests for `ValidPoolName` and `ValidVolumeName` to verify all edge cases (empty, whitespace, leading hyphens, "..", backslashes, etc.).

I do agree with CodeQL here but that's material for another PR probably.

lxd/storage/drivers/utils.go
simondeziel: `;` would also be another candidate to block. Maybe `{`, `}`, `|` and `&`...

It's a bit late to try to retrofit those restrictions :/

tomponline:
> > **lxd/storage/drivers/utils.go:900**
> > ```
> > * Consider adding unit tests for `ValidPoolName` and `ValidVolumeName` to verify all edge cases (empty, whitespace, leading hyphens, "..", backslashes, etc.).
> > ```
> 
> I do agree with CodeQL here but that's material for another PR probably.

yep wouldn't hurt, at least we've got integration tests that cover the pool name stuff.

lxd/storage/drivers/utils.go
tomponline: indeed - baby steps

tomponline: @simondeziel tested creating volumes starting with `-` on dir, zfs, ceph, lvm and btrfs, and all work fine because LXD prefixes the project name to the volume name.

PR #15942, opened 2025-07-04
golangci: Disable package name checks in var-naming revive checker (stable-5.21)

tomponline:
It was failing on the lxd/util package naming which we dont want to change in this series.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -31,12 +31,6 @@ linters:
         - name: unchecked-type-assertion
           arguments:
             - acceptIgnoredAssertionResult: true
-        # https://github.com/mgechev/revive/blob/2a1701aadbedfcc175cb92836a51407bec382652/RULES_DESCRIPTIONS.md#var-naming
-        - name: var-naming
-          arguments: # The arguments here are quite odd looking. See the rule description.
-            - []
-            - []
-            - - upperCaseConst: true
         # https://github.com/mgechev/revive/blob/2a1701aadbedfcc175cb92836a51407bec382652/RULES_DESCRIPTIONS.md#early-return
         - name: early-return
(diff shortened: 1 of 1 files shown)


kadinsayani (commented):
Removing the linter entirely for 5.21 is fine but I think we could also just add an exclusion for `lxd/util`.

tomponline:
> Removing the linter entirely for 5.21 is fine but I think we could also just add an exclusion for `lxd/util`.

how?

tomponline:
> Removing the linter entirely for 5.21 is fine but I think we could also just add an exclusion for `lxd/util`.

the revive docs show an option to disable package name checks, but this must not be the revive version included in golangci as that doesnt work.

tomponline:
> * exclude-rules

@kadinsayani @simondeziel copilot appears to be suggesting nonsense, because the link is for early-return detection :(

simondeziel (approved):

kadinsayani:
> > Removing the linter entirely for 5.21 is fine but I think we could also just add an exclusion for `lxd/util`.
> 
> how?

Something like this should work, similar to how we have an exclusion rule for `staticcheck ST1005`:

```yaml
settings:
...
exclusions:
...
    rules:
      - path: lxd/util\.go
        linters:
          - var-naming
```

tomponline:
> > > Removing the linter entirely for 5.21 is fine but I think we could also just add an exclusion for `lxd/util`.
> > 
> > 
> > how?
> 
> Something like this should work, similar to how we have an exclusion rule for `staticcheck ST1005`:
> 
> ```yaml
> settings:
> ...
> exclusions:
> ...
>     rules:
>       - path: lxd/util\.go
>         linters:
>           - var-naming
> ```

Would you mind trying that out on main, and we can leave this disabled on 5.21-stable as it doesnt make sense to be too strict on this series as we're only going to be including stuff backported from main anyway.

Issue #15733, opened 2025-06-05
Deleting an instance with limited disk causes each file to be touched twice


mariusklocke:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

22.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
Name    Version         Rev    Tracking        Publisher   Notes
core20  20250407        2571   latest/stable   canonical✓  base,disabled
core20  20250429        2582   latest/stable   canonical✓  base
core22  20250408        1963   latest/stable   canonical✓  base,disabled
core22  20250425        1981   latest/stable   canonical✓  base
lxd     5.21.3-c99e75b  32119  5.21/candidate  canonical✓  disabled,held
lxd     5.21.3-75def3c  32455  5.21/candidate  canonical✓  held
snapd   2.67.1          23771  latest/stable   canonical✓  snapd,disabled
snapd   2.68.4          24505  latest/stable   canonical✓  snapd
```

### Output of "lxc info" or system info if it fails

```shell
N/A
```

### Issue description

When deleting instances with limited disk size LXD first removes the quota and then deletes all files. Removing a quota means updating the extended file attributes of every file in the volume. Which means every file is updated before it is deleted. I tested this with a container with 10 mio files. Removing the quota took about 8 minutes and deleting all files another 8 minutes. Performance can improved significantly by changing the execution order of these two steps. If the files are already deleted, the quota removal should be done in a second.

It is rather a performance issue than a functional impairment. When using the `dir` storage driver, my workaround is simple: Just delete the container files before invoking `lxc delete`.

Just change the execution order in `DeleteVolume` function and halve the deletion time for instances with limited disks:
https://github.com/canonical/lxd/blob/9bc9730488387883fb5c9a0efc6d72ada24ffdf3/lxd/storage/drivers/driver_dir_volumes.go#L176

Please understand that I don't like to create a PR due to the hassle with the CLA on my last report.

### Steps to reproduce

1. Create a storage pool with `dir` driver (make sure the path is on XFS partition with quota support)
2. Create a container instance with a limited disk size
3. Generate 10 million blank files inside the container
4. Delete the container
5. Watch the output of `xfs_quota -x -c "report -i $mountpoint"` and `du --inodes $mountpoint`

### Information to attach

- [ ] Any relevant kernel output (`dmesg`)
- [ ] Instance log (`lxc info NAME --show-log`)
[...]

PR #15943, opened 2025-07-04
Storage: Remove quotas after removing volume from `dir` storage

nmezhenskyi:
Fixes https://github.com/canonical/lxd/issues/15733.  

This change improves performance when deleting a volume with a large amount of files stored in a `dir` storage pool on an XFS partition with project quotas enabled (particularly with the quota set for inodes usage).

The PR also adds a test for `dir` storage driver backed by XFS to ensure that LXD correctly configures XFS quotas.

## Benchmark results:
On my local set up there is about **34%** speed improvement on a sample size of 1,000,000 empty files stored in a container on an XFS partition of 10GiB with project quota `bhard=1G ihard=1001000`.

## How to benchmark:
1. Get a loop device that can be formatted (i.e., `/dev/sdb`).
1. Format the device with XFS: `sudo mkfs.xfs /dev/sdb`.
1. Create a mount point: `sudo mkdir -p <mount_dir>`.
1. Get UUID of the device: i.e., run `blkid`.
1. Add this line to the `/etc/fstab` file: `UUID="<device_uuid>"  <mount_dir>  xfs  defaults,prjquota  0  0`.
1. Mount the filesystem: `sudo mount -a`.
1. Create LXD storage pool using the `dir` driver: `lxc storage create xfs_pool dir source=<mount_dir>`.
1. Create a profile to use for test containers:
   ```
   lxc profile create xfs_profile
   lxc profile device add xfs_profile root disk pool=xfs_pool path=/ size=1GB
   ```
1. Launch a container:
   ```
   lxc launch ubuntu-minimal:24.04 test-container -p default -p xfs_profile
   ```
1. Set XFS project quota:
   ```
   container_path="${mount_dir}/containers/test-container"

   project_id=$(sudo lsattr -p "${container_path}" | awk '{print $1}')

   sudo xfs_quota -x -c "limit -p bhard=1G ihard=1001000 ${project_id}" "${LXD_DIR}/storage-pools/xfs_pool"
   ```
1. Generate 1,000,000 empty files:
   ```
   lxc exec test-container -- sh -c 'mkdir -p /root/manyfiles && cd /root/manyfiles && for i in $(seq 1 1000000); do truncate -s 0 "file$i"; done'
   ```
1. Stop the container: `lxc stop test-container`.

**Then,**  
1. Watch the output of: `watch 'xfs_quota -x -c "report -i <mount_dir>"'`.
1. Watch the output of: `watch 'du --inodes <mount_dir>'`.
1. Time the deletion of container: `time lxc delete test-container`.

### DIFF
--- a/test/suites/storage_driver_dir.sh
+++ b/test/suites/storage_driver_dir.sh
@@ -10,6 +10,7 @@ test_storage_driver_dir() {
   fi
 
   do_dir_on_empty_fs
+  do_dir_xfs_project_quotas
 }
 
 do_dir_on_empty_fs() {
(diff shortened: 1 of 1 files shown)


tomponline (commented):
@nmezhenskyi do we have any tests for xfs/ext4 project quotas btw?

nmezhenskyi:
@tomponline 

We have tests for project `limits.disk` and `limits.disk.pool` usage in `test/suites/projects.sh`, but I'm not sure if these tests are run for XFS/ext4 specifically.

nmezhenskyi:
@tomponline 

> do we have any tests for xfs/ext4 project quotas btw?

Sorry, I misunderstood the question. No, it seems we don't have any tests for XFS/ext4 prjquota.

tomponline:
> @tomponline
> 
> > do we have any tests for xfs/ext4 project quotas btw?
> 
> Sorry, I misunderstood the question. No, it seems we don't have any tests for XFS/ext4 prjquota.

Could we add some? I believe you can use a loop back file mounted with a filesystem and enable project quota at mount time for use as a dir pool source.

test/suites/storage_driver_dir.sh
```diff
@@ -42,3 +43,52 @@ do_dir_on_empty_fs() {
...
+  lxc profile device add xfs_profile root disk pool=xfs_pool path=/ size=1GB
+
+  echo "==> Launch a container using the dir storage pool backed by XFS."
+  lxc launch images:alpine/edge foo -p default -p xfs_profile
+
+  container_path="${mount_point}/containers/foo"
+  project_id=$(lsattr -p "${container_path}" | awk '{print $1}')
+
+  echo "==> Set XFS project quota for the container."
+  xfs_quota -x -c "limit -p bhard=850mb ihard=1000 ${project_id}" "${mount_point}/storage-pools/xfs_pool"
```
tomponline: This should be done using lxc to set disk size otherwise we arent testing lxd project quota functionality

tomponline: Can we check it got applied by lxd?

tomponline (changes requested):

test/suites/storage_driver_dir.sh
nmezhenskyi: Yes, makes sense

nmezhenskyi:
Interesting, in the `do_dir_ext4_project_quotas()` test:
 `mount -o loop "${tmp_file}" "${mount_point}"` always fails after calling   
 `tune2fs -O project -Q prjquota "${tmp_file}"`.
 
 The error is: `mount(2) system call failed: No such process.`   
 
 Without calling `tune2fs` the mount succeeds, but the project quotas are not enabled in that case.
 
 On my local setup the script is working correctly, so this is probably something specific to GitHub runners.
 
 @tomponline do you know what could be the issue?

nmezhenskyi:
The `do_dir_xfs_project_quotas()` test is passing fine.

nmezhenskyi:
@tomponline I removed the `ext4` test from this PR, so that we can merge it. Once Simon is back I'll ask him about it, maybe we can figure out together why an `ext4` volume with project quotas fails to mount inside a GitHub runner.

test/suites/storage_driver_dir.sh
```diff
@@ -42,3 +43,60 @@ do_dir_on_empty_fs() {
...
+
+  echo "==> Create LXD dir storage pool backed by XFS."
+  lxc storage create xfs_pool dir source="${mount_point}"
+
+  echo "==> Create a profile that uses xfs_pool as root disk. Limit the root disk to 850MiB."
+  lxc profile create xfs_profile
+  lxc profile device add xfs_profile root disk pool=xfs_pool path=/ size=850MiB
+
+  echo "==> Launch a container using the dir storage pool backed by XFS."
+  lxc launch images:alpine/edge foo -p default -p xfs_profile
```
tomponline, line 74: you probably didnt need the profile here, could just have used `--storage xfs_pool --device root,size=850MiB`

nmezhenskyi: True, it would've made the test a bit smaller. I initially thought it might be useful for testing multiple instances, but ended up using just a single instance.

PR #15946, opened 2025-07-04
HPE Alletra storage driver

mihalicyn:
Work in progress

### DIFF
--- a/lxd/storage/drivers/driver_hpe_alletra.go
+++ b/lxd/storage/drivers/driver_hpe_alletra.go
@@ -1,0 +1,286 @@
+package drivers
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/canonical/lxd/lxd/operations"
+	"github.com/canonical/lxd/lxd/storage/connectors"
+	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/logger"
+	"github.com/canonical/lxd/shared/validate"
+)
+
+// hpeAlletraLoaded indicates whether load() function was already called for this storage driver.
+var hpeAlletraLoaded = false
+
+// hpeAlletraVersion indicates storage driver version.
+var hpeAlletraVersion = ""
+
+// hpeAlletraSupportedConnectors represents a list of storage connectors that can be used.
+var hpeAlletraSupportedConnectors = []string{
+	connectors.TypeISCSI,
+	connectors.TypeNVME,
+}
+
+type hpeAlletra struct {
+	common
+
+	// Holds the low level connector (iSCSI, NVMEoTCP).
+	// Use .connector() method to retrieve the initialized connector.
+	storageConnector connectors.Connector
+
+	// Holds the targetQN of NVMe target.
+	nvmeTargetQN string
+
+	// Holds the low level HTTP client for the HPE Alletra Storage API.
+	// Use .client() method to retrieve the client struct.
+	httpClient *hpeAlletraClient
+}
+
+// load is used to run one-time action per-driver rather than per-pool.
+func (d *hpeAlletra) load() error {
+	// Done if previously loaded.
+	if hpeAlletraLoaded {
+		return nil
+	}
+
+	versions := connectors.GetSupportedVersions(hpeAlletraSupportedConnectors)
+	hpeAlletraVersion = strings.Join(versions, " / ")
+	hpeAlletraLoaded = true
+
+	// Load the kernel modules of the respective connector, ignoring those that cannot be loaded.
+	// Support for a specific connector is checked during pool creation. However, this
+	// ensures that the kernel modules are loaded, even if the host has been rebooted.
+	connector, err := d.connector()
+	if err == nil {
+		_ = connector.LoadModules()
+	}
+
+	return nil
+}
+
+// connector retrieves an initialized storage connector based on the configured
+// storage driver mode (iSCSI, NVMEoTCP). The connector is cached in the driver struct.
+func (d *hpeAlletra) connector() (connectors.Connector, error) {
+	if d.storageConnector == nil {
+		connector, err := connectors.NewConnector(d.config["hpe_alletra.mode"], d.state.ServerUUID)
+		if err != nil {
+			return nil, err
+		}
+
+		d.storageConnector = connector
+	}
+
+	return d.storageConnector, nil
+}
+
+// client returns the drivers HPE Alletra Storage client. A new client is created only if it does not already exist.
+func (d *hpeAlletra) client() *hpeAlletraClient {
+	if d.httpClient == nil {
+		d.httpClient = newHPEAlletraClient(d)
+	}
+
+	return d.httpClient
+}
+
+// isRemote returns true indicating this driver uses remote storage.
+func (d *hpeAlletra) isRemote() bool {
+	return true
+}
+
+// Info returns info about the driver and its environment.
+func (d *hpeAlletra) Info() Info {
+	return Info{
+		Name:                         "hpeAlletra",
+		Version:                      "1",
+		DefaultBlockSize:             d.defaultBlockVolumeSize(),
+		DefaultVMBlockFilesystemSize: d.defaultVMBlockFilesystemSize(),
+		OptimizedImages:              false,
+		PreservesInodes:              true,
+		Remote:                       d.isRemote(),
+		VolumeTypes:                  []VolumeType{VolumeTypeCustom, VolumeTypeImage, VolumeTypeContainer, VolumeTypeVM},
+		BlockBacking:                 true,
+		RunningCopyFreeze:            true,
+		DirectIO:                     true,
+		MountedRoot:                  false,
+		PopulateParentVolumeUUID:     true,
+	}
+}
+
+// FillConfig populates the driver's config with default values.
+func (d *hpeAlletra) FillConfig() error {
+	// Use NVMe by default.
+	if d.config["hpe_alletra.mode"] == "" {
+		d.config["hpe_alletra.mode"] = connectors.TypeNVME
+	}
+
+	return nil
+}
+
+// Validate checks that all provide keys are supported and that no conflicting or missing configuration is present.
+func (d *hpeAlletra) Validate(config map[string]string) error {
+	rules := map[string]func(value string) error{
+		// lxdmeta:generate(entities=storage-hpe-alletra; group=pool-conf; key=hpe_alletra.wsapi.url)
+		//
+		// ---
+		//  type: string
+		//  shortdesc: Address of the HPE Alletra Storage UI/WSAPI
+		"hpe_alletra.wsapi.url": validate.Optional(validate.IsRequestURL),
+		// lxdmeta:generate(entities=storage-hpe-alletra; group=pool-conf; key=hpe_alletra.wsapi.verify)
+		//
+		// ---
+		//  type: bool
+		//  defaultdesc: `true`
+		//  shortdesc: Whether to verify the HPE Alletra Storage UI/WSAPI certificate
+		"hpe_alletra.wsapi.verify": validate.Optional(validate.IsBool),
+		// lxdmeta:generate(entities=storage-hpe-alletra; group=pool-conf; key=hpe_alletra.wsapi.user.name)
+		//
+		// ---
+		//  type: string
+		//  shortdesc: HPE Alletra storage admin username
+		"hpe_alletra.wsapi.user.name": validate.IsAny,
+		// lxdmeta:generate(entities=storage-hpe-alletra; group=pool-conf; key=hpe_alletra.wsapi.user.password)
+		//
+		// ---
+		//  type: string
+		//  shortdesc: HPE Alletra storage admin password
+		"hpe_alletra.wsapi.user.password": validate.IsAny,
+		// lxdmeta:generate(entities=storage-hpe-alletra; group=pool-conf; key=hpe_alletra.wsapi.cpg)
+		//
+		// ---
+		//  type: string
+		//  shortdesc: HPE Alletra CPG name
+		"hpe_alletra.wsapi.cpg": validate.IsAny,
+		// lxdmeta:generate(entities=storage-hpe-alletra; group=pool-conf; key=hpe_alletra.target)
+		// A comma-separated list of target addresses. If empty, LXD discovers and connects to all available targets. Otherwise, it only connects to the specified addresses.
+		// ---
+		//  type: string
+		//  defaultdesc: the discovered mode
+		//  shortdesc: List of target addresses.
+		"hpe_alletra.target": validate.Optional(validate.IsListOf(validate.IsNetworkAddress)),
+		// lxdmeta:generate(entities=storage-hpe-alletra; group=pool-conf; key=hpe_alletra.mode)
+		// The mode to use to map Pure Storage volumes to the local server.
+		// Supported values are `iscsi` and `nvme`.
+		// ---
+		//  type: string
+		//  defaultdesc: the discovered mode
+		//  shortdesc: How volumes are mapped to the local server
+		"hpe_alletra.mode": validate.Optional(validate.IsOneOf(pureSupportedConnectors...)),
+		// lxdmeta:generate(entities=storage-hpe-alletra; group=pool-conf; key=volume.size)
+		// Default storage volume size rounded to 256MiB. The minimum size is 256MiB.
+		// ---
+		//  type: string
+		//  defaultdesc: `10GiB`
+		//  shortdesc: Size/quota of the storage volume
+		"volume.size": validate.Optional(validate.IsMultipleOfUnit("256MiB")),
+	}
+
+	err := d.validatePool(config, rules, d.commonVolumeRules())
+	if err != nil {
+		return err
+	}
+
+	newMode := config["hpe_alletra.mode"]
+	oldMode := d.config["hpe_alletra.mode"]
+
+	// Ensure hpe_alletra.mode cannot be changed to avoid leaving volume mappings
+	// and prevent disturbing running instances.
+	if oldMode != "" && oldMode != newMode {
+		return errors.New("HPE Alletra Storage mode cannot be changed")
+	}
+
+	// Check if the selected HPE Alletra Storage mode is supported on this node.
+	// Also when forming the storage pool on a LXD cluster, the mode
+	// that got discovered on the creating machine needs to be validated
+	// on the other cluster members too. This can be done here since Validate
+	// gets executed on every cluster member when receiving the cluster
+	// notification to finally create the pool.
+	if newMode != "" {
+		connector, err := connectors.NewConnector(newMode, "")
+		if err != nil {
+			return fmt.Errorf("HPE Alletra Storage mode %q is not supported: %w", newMode, err)
+		}
+
+		err = connector.LoadModules()
+		if err != nil {
+			return fmt.Errorf("HPE Alletra Storage mode %q is not supported due to missing kernel modules: %w", newMode, err)
+		}
+	}
+
+	return nil
+}
+
+// Create is called during pool creation and is effectively using an empty driver struct.
+// WARNING: The Create() function cannot rely on any of the struct attributes being set.
+func (d *hpeAlletra) Create() error {
+	err := d.FillConfig()
+	if err != nil {
+		return err
+	}
+
+	// Validate both pool and gateway here and return an error if they are not set.
+	// Since those aren't any cluster member specific keys the general validation
+	// rules allow empty strings in order to create the pending storage pools.
+	if d.config["hpe_alletra.wsapi.url"] == "" {
+		return errors.New("The hpe_alletra.wsapi.url cannot be empty")
+	}
+
+	if d.config["hpe_alletra.wsapi.user.name"] == "" {
+		return errors.New("The hpe_alletra.wsapi.user.name cannot be empty")
+	}
+
+	if d.config["hpe_alletra.wsapi.user.password"] == "" {
+		return errors.New("The hpe_alletra.wsapi.user.password cannot be empty")
+	}
+
+	// TODO: do we need to *actually* go to WSAPI and create something?
+	// Volumeset? Hostset? something else?
+	logger.Warn("Going to createVolumeSet")
+	err = d.client().createVolumeSet()
+	if err != nil {
+		logger.Warn("createVolumeSet failed", logger.Ctx{"err": err})
+		return err
+	}
+
+	return nil
+}
+
+// Delete removes a storage pool.
+func (d *hpeAlletra) Delete(op *operations.Operation) error {
+	// FIXME: recursive?
+	err := d.client().deleteVolumeSet()
+	if err != nil {
+		return err
+	}
+
+	// If the user completely destroyed it, call it done.
+	if !shared.PathExists(GetPoolMountPath(d.name)) {
+		return nil
+	}
+
+	// On delete, wipe everything in the directory.
+	return wipeDirectory(GetPoolMountPath(d.name))
+}
+
+// Update applies any driver changes required from a configuration change.
+func (d *hpeAlletra) Update(changedConfig map[string]string) error {
+	return nil
+}
+
+// Mount mounts the storage pool.
+func (d *hpeAlletra) Mount() (bool, error) {
+	return true, nil
+}
+
+// Unmount unmounts the storage pool.
+func (d *hpeAlletra) Unmount() (bool, error) {
+	return true, nil
+}
+
+// GetResources returns the pool resource usage information.
+func (d *hpeAlletra) GetResources() (*api.ResourcesStoragePool, error) {
+	return nil, nil
+}
--- a/lxd/storage/drivers/driver_hpe_alletra_volumes.go
+++ b/lxd/storage/drivers/driver_hpe_alletra_volumes.go
@@ -1,0 +1,977 @@
+package drivers
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"math"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/canonical/lxd/lxd/backup"
+	"github.com/canonical/lxd/lxd/instancewriter"
+	"github.com/canonical/lxd/lxd/migration"
+	"github.com/canonical/lxd/lxd/operations"
+	"github.com/canonical/lxd/lxd/storage/block"
+	"github.com/canonical/lxd/lxd/storage/connectors"
+	"github.com/canonical/lxd/lxd/storage/filesystem"
+	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/revert"
+	"github.com/canonical/lxd/shared/units"
+	"github.com/canonical/lxd/shared/validate"
+	"github.com/google/uuid"
+)
+
+// factorMiB divides a byte size value into Mebibytes.
+const factorMiB = 1024 * 1024
+
+// hpeAlletraVolTypePrefixes maps volume type to storage volume name prefix.
+// Use smallest possible prefixes since Pure Storage volume names are limited to 63 characters.
+var hpeAlletraVolTypePrefixes = map[VolumeType]string{
+	VolumeTypeContainer: "c",
+	VolumeTypeVM:        "v",
+	VolumeTypeImage:     "i",
+	VolumeTypeCustom:    "u",
+}
+
+// hpeAlletraContentTypeSuffixes maps volume's content type to storage volume name suffix.
+var hpeAlletraContentTypeSuffixes = map[ContentType]string{
+	// Suffix used for block content type volumes.
+	ContentTypeBlock: "b",
+
+	// Suffix used for ISO content type volumes.
+	ContentTypeISO: "i",
+}
+
+// hpeAlletraSnapshotPrefix is a prefix used for Pure Storage snapshots to avoid name conflicts
+// when creating temporary volume from the snapshot.
+var hpeAlletraSnapshotPrefix = "s"
+
+// defaultVMBlockFilesystemSize is the size of a VM root device block volume's associated filesystem volume.
+const hpeAlletraVMBlockFilesystemSize = "256MiB"
+
+// DefaultVMBlockFilesystemSize returns the size of a VM root device block volume's associated filesystem volume.
+func (d *hpeAlletra) defaultVMBlockFilesystemSize() string {
+	return hpeAlletraVMBlockFilesystemSize
+}
+
+// getVolumeName returns the fully qualified name derived from the volume's UUID.
+// TODO: copied as-is from pure. refactor?
+func (d *hpeAlletra) getVolumeName(vol Volume) (string, error) {
+	volUUID, err := uuid.Parse(vol.config["volatile.uuid"])
+	if err != nil {
+		return "", fmt.Errorf(`Failed parsing "volatile.uuid" from volume %q: %w`, vol.name, err)
+	}
+
+	// Remove hypens from the UUID to create a volume name.
+	volName := strings.ReplaceAll(volUUID.String(), "-", "")
+
+	// Search for the volume type prefix, and if found, prepend it to the volume name.
+	volumeTypePrefix, ok := hpeAlletraVolTypePrefixes[vol.volType]
+	if ok {
+		volName = volumeTypePrefix + "-" + volName
+	}
+
+	// Search for the content type suffix, and if found, append it to the volume name.
+	contentTypeSuffix, ok := hpeAlletraContentTypeSuffixes[vol.contentType]
+	if ok {
+		volName = volName + "-" + contentTypeSuffix
+	}
+
+	// If volume is snapshot, prepend snapshot prefix to its name.
+	if vol.IsSnapshot() {
+		volName = hpeAlletraSnapshotPrefix + volName
+	}
+
+	return volName, nil
+}
+
+// ensureHost returns a name of the host that is configured with a given IQN. If such host
+// does not exist, a new one is created, where host's name equals to the server name with a
+// mode included as a suffix because Pure Storage does not allow mixing IQNs, NQNs, and WWNs
+// on a single host.
+// TODO: copied as-is from pure. refactor?
+func (d *hpeAlletra) ensureHost() (hostName string, cleanup revert.Hook, err error) {
+	var hostname string
+
+	revert := revert.New()
+	defer revert.Fail()
+
+	connector, err := d.connector()
+	if err != nil {
+		return "", nil, err
+	}
+
+	// Get the qualified name of the host.
+	qn, err := connector.QualifiedName()
+	if err != nil {
+		return "", nil, err
+	}
+
+	// Fetch an existing Pure Storage host.
+	host, err := d.client().getCurrentHost()
+	if err != nil {
+		if !api.StatusErrorCheck(err, http.StatusNotFound) {
+			return "", nil, err
+		}
+
+		// The Pure Storage host with a qualified name of the current LXD host does not exist.
+		// Therefore, create a new one and name it after the server name.
+		serverName, err := ResolveServerName(d.state.ServerName)
+		if err != nil {
+			return "", nil, err
+		}
+
+		// Append the mode to the server name because Pure Storage does not allow mixing
+		// NQNs, IQNs, and WWNs for a single host.
+		hostname = serverName + "-" + connector.Type()
+
+		err = d.client().createHost(hostname, []string{qn})
+		if err != nil {
+			if !api.StatusErrorCheck(err, http.StatusConflict) {
+				return "", nil, err
+			}
+
+			// The host with the given name already exists, update it instead.
+			err = d.client().updateHost(hostname, []string{qn})
+			if err != nil {
+				return "", nil, err
+			}
+		} else {
+			revert.Add(func() { _ = d.client().deleteHost(hostname) })
+		}
+	} else {
+		// Hostname already exists with the given IQN.
+		hostname = host.Name
+	}
+
+	cleanup = revert.Clone().Fail
+	revert.Success()
+	return hostname, cleanup, nil
+}
+
+// mapVolume maps the given volume onto this host.
+// TODO: copied as-is (with driver rename in one place) from pure. refactor?
+func (d *hpeAlletra) mapVolume(vol Volume) (cleanup revert.Hook, err error) {
+	reverter := revert.New()
+	defer reverter.Fail()
+
+	connector, err := d.connector()
+	if err != nil {
+		return nil, err
+	}
+
+	volName, err := d.getVolumeName(vol)
+	if err != nil {
+		return nil, err
+	}
+
+	unlock, err := remoteVolumeMapLock(connector.Type(), d.Info().Name)
+	if err != nil {
+		return nil, err
+	}
+
+	defer unlock()
+
+	// Ensure the host exists and is configured with the correct QN.
+	hostname, cleanup, err := d.ensureHost()
+	if err != nil {
+		return nil, err
+	}
+
+	reverter.Add(cleanup)
+
+	// Ensure the volume is connected to the host.
+	connCreated, err := d.client().connectHostToVolume(vol.pool, volName, hostname)
+	if err != nil {
+		return nil, err
+	}
+
+	if connCreated {
+		reverter.Add(func() { _ = d.client().disconnectHostFromVolume(vol.pool, volName, hostname) })
+	}
+
+	// Find the array's qualified name for the configured mode.
+	targetQN, targetAddrs, err := d.client().getTarget()
+	if err != nil {
+		return nil, err
+	}
+
+	// Connect to the array.
+	connReverter, err := connector.Connect(d.state.ShutdownCtx, targetQN, targetAddrs...)
+	if err != nil {
+		return nil, err
+	}
+
+	reverter.Add(connReverter)
+
+	// If connect succeeded it means we have at least one established connection.
+	// However, it's reverter does not cleanup the establised connections or a newly
+	// created session. Therefore, if we created a mapping, add unmapVolume to the
+	// returned (outer) reverter. Unmap ensures the target is disconnected only when
+	// no other device is using it.
+	outerReverter := revert.New()
+	if connCreated {
+		outerReverter.Add(func() { _ = d.unmapVolume(vol) })
+	}
+
+	// Add connReverter to the outer reverter, as it will immediately stop
+	// any ongoing connection attempts. Note that it must be added after
+	// unmapVolume to ensure it is called first.
+	outerReverter.Add(connReverter)
+
+	reverter.Success()
+	return outerReverter.Fail, nil
+}
+
+// unmapVolume unmaps the given volume from this host.
+// TODO: copied as-is (with driver rename in one place) from pure. refactor?
+func (d *hpeAlletra) unmapVolume(vol Volume) error {
+	connector, err := d.connector()
+	if err != nil {
+		return err
+	}
+
+	volName, err := d.getVolumeName(vol)
+	if err != nil {
+		return err
+	}
+
+	unlock, err := remoteVolumeMapLock(connector.Type(), d.Info().Name)
+	if err != nil {
+		return err
+	}
+
+	defer unlock()
+
+	host, err := d.client().getCurrentHost()
+	if err != nil {
+		return err
+	}
+
+	// Disconnect the volume from the host and ignore error if connection does not exist.
+	err = d.client().disconnectHostFromVolume(vol.pool, volName, host.Name)
+	if err != nil && !api.StatusErrorCheck(err, http.StatusNotFound) {
+		return err
+	}
+
+	volumePath, _, _ := d.getMappedDevPath(vol, false)
+	if volumePath != "" {
+		// When iSCSI volume is disconnected from the host, the device will remain on the system.
+		//
+		// To remove the device, we need to either logout from the session or remove the
+		// device manually. Logging out of the session is not desired as it would disconnect
+		// from all connected volumes. Therefore, we need to manually remove the device.
+		if connector.Type() == connectors.TypeISCSI {
+			// removeDevice removes device from the system if the device is removable.
+			removeDevice := func(devName string) error {
+				path := "/sys/block/" + devName + "/device/delete"
+				if shared.PathExists(path) {
+					// Delete device.
+					err := os.WriteFile(path, []byte("1"), 0400)
+					if err != nil {
+						return err
+					}
+				}
+
+				return nil
+			}
+
+			devName := filepath.Base(volumePath)
+			if strings.HasPrefix(devName, "dm-") {
+				// Multipath device (/dev/dm-*) itself is not removable.
+				// Therefore, we remove its slaves instead.
+				slaves, err := filepath.Glob("/sys/block/" + devName + "/slaves/*")
+				if err != nil {
+					return fmt.Errorf("Failed to unmap volume %q: Failed to list slaves for device %q: %w", vol.name, devName, err)
+				}
+
+				// Remove slave devices.
+				for _, slave := range slaves {
+					slaveDevName := filepath.Base(slave)
+
+					err := removeDevice(slaveDevName)
+					if err != nil {
+						return fmt.Errorf("Failed to unmap volume %q: Failed to remove slave device %q: %w", vol.name, slaveDevName, err)
+					}
+				}
+			} else {
+				// For non-multipath device (/dev/sd*), remove the device itself.
+				err := removeDevice(devName)
+				if err != nil {
+					return fmt.Errorf("Failed to unmap volume %q: Failed to remove device %q: %w", vol.name, devName, err)
+				}
+			}
+		}
+
+		// Wait until the volume has disappeared.
+		ctx, cancel := context.WithTimeout(d.state.ShutdownCtx, 30*time.Second)
+		defer cancel()
+
+		if !block.WaitDiskDeviceGone(ctx, volumePath) {
+			return fmt.Errorf("Timeout exceeded waiting for Pure Storage volume %q to disappear on path %q", vol.name, volumePath)
+		}
+	}
+
+	// If this was the last volume being unmapped from this system, disconnect the active session
+	// and remove the host from HPE storage.
+
+	//FIXME: it looks like we have no option than iterating over VLUNs to find out if there is any other active sessions
+
+	return nil
+}
+
+// getMappedDevPath returns the local device path for the given volume.
+// Indicate with mapVolume if the volume should get mapped to the system if it isn't present.
+// TODO: copied as-is (except switch/case) from pure. refactor?
+func (d *hpeAlletra) getMappedDevPath(vol Volume, mapVolume bool) (string, revert.Hook, error) {
+	revert := revert.New()
+	defer revert.Fail()
+
+	connector, err := d.connector()
+	if err != nil {
+		return "", nil, err
+	}
+
+	if mapVolume {
+		cleanup, err := d.mapVolume(vol)
+		if err != nil {
+			return "", nil, err
+		}
+
+		revert.Add(cleanup)
+	}
+
+	volName, err := d.getVolumeName(vol)
+	if err != nil {
+		return "", nil, err
+	}
+
+	hpeVol, err := d.client().getVolume(vol.pool, volName)
+	if err != nil {
+		return "", nil, err
+	}
+
+	// Ensure the NGUID is exactly 32 characters long, as it uniquely
+	// identifies the device. This check should never succeed, but prevents
+	// out-of-bounds errors when slicing the string later.
+	if len(hpeVol.NGUID) != 32 { //
+		return "", nil, fmt.Errorf("Failed to locate device for volume %q: Unexpected length of NGUID %q (%d)", vol.name, hpeVol.NGUID, len(hpeVol.NGUID))
+	}
+
+	var diskPrefix string
+	var diskSuffix string
+
+	//TODO: only this part was changed. Too many copied code. We need to generalize this.
+	switch connector.Type() {
+	case connectors.TypeISCSI:
+		diskPrefix = "scsi-"
+		diskSuffix = strings.ToLower(hpeVol.NGUID)
+	case connectors.TypeNVME:
+		diskPrefix = "nvme-eui."
+		diskSuffix = strings.ToLower(hpeVol.NGUID)
+	default:
+		return "", nil, fmt.Errorf("Unsupported HPE Storage mode %q", connector.Type())
+	}
+
+	// Filters devices by matching the device path with the lowercase disk suffix.
+	// Pure Storage reports serial numbers in uppercase, so the suffix is converted
+	// to lowercase.
+	diskPathFilter := func(devPath string) bool {
+		return strings.HasSuffix(devPath, strings.ToLower(diskSuffix))
+	}
+
+	var devicePath string
+	if mapVolume {
+		// Wait until the disk device is mapped to the host.
+		devicePath, err = block.WaitDiskDevicePath(d.state.ShutdownCtx, diskPrefix, diskPathFilter)
+	} else {
+		// Expect device to be already mapped.
+		devicePath, err = block.GetDiskDevicePath(diskPrefix, diskPathFilter)
+	}
+
+	if err != nil {
+		return "", nil, fmt.Errorf("Failed to locate device for volume %q: %w", vol.name, err)
+	}
+
+	cleanup := revert.Clone().Fail
+	revert.Success()
+	return devicePath, cleanup, nil
+}
+
+// CreateVolume creates an empty volume and can optionally fill it by executing the supplied filler function.
+// TODO: copied as-is from pure. refactor?
+func (d *hpeAlletra) CreateVolume(vol Volume, filler *VolumeFiller, op *operations.Operation) error {
+	client := d.client()
+
+	revert := revert.New()
+	defer revert.Fail()
+
+	volName, err := d.getVolumeName(vol)
+	if err != nil {
+		return err
+	}
+
+	sizeBytes, err := units.ParseByteSizeString(vol.ConfigSize())
+	if err != nil {
+		return err
+	}
+
+	// Create the volume.
+	err = client.createVolume(vol.pool, volName, sizeBytes)
+	if err != nil {
+		return err
+	}
+
+	revert.Add(func() { _ = client.deleteVolume(vol.pool, volName) })
+
+	volumeFilesystem := vol.ConfigBlockFilesystem()
+	if vol.contentType == ContentTypeFS {
+		devPath, cleanup, err := d.getMappedDevPath(vol, true)
+		if err != nil {
+			return err
+		}
+
+		revert.Add(cleanup)
+
+		_, err = makeFSType(devPath, volumeFilesystem, nil)
+		if err != nil {
+			return err
+		}
+	}
+
+	// For VMs, also create the filesystem volume.
+	if vol.IsVMBlock() {
+		fsVol := vol.NewVMBlockFilesystemVolume()
+
+		err := d.CreateVolume(fsVol, nil, op)
+		if err != nil {
+			return err
+		}
+
+		revert.Add(func() { _ = d.DeleteVolume(fsVol, op) })
+	}
+
+	err = vol.MountTask(func(mountPath string, op *operations.Operation) error {
+		// Run the volume filler function if supplied.
+		if filler != nil && filler.Fill != nil {
+			var err error
+			var devPath string
+
+			if IsContentBlock(vol.contentType) {
+				// Get the device path.
+				devPath, err = d.GetVolumeDiskPath(vol)
+				if err != nil {
+					return err
+				}
+			}
+
+			allowUnsafeResize := false
+			if vol.volType == VolumeTypeImage {
+				// Allow filler to resize initial image volume as needed.
+				// Some storage drivers don't normally allow image volumes to be resized due to
+				// them having read-only snapshots that cannot be resized. However when creating
+				// the initial image volume and filling it before the snapshot is taken resizing
+				// can be allowed and is required in order to support unpacking images larger than
+				// the default volume size. The filler function is still expected to obey any
+				// volume size restrictions configured on the pool.
+				// Unsafe resize is also needed to disable filesystem resize safety checks.
+				// This is safe because if for some reason an error occurs the volume will be
+				// discarded rather than leaving a corrupt filesystem.
+				allowUnsafeResize = true
+			}
+
+			// Run the filler.
+			err = d.runFiller(vol, devPath, filler, allowUnsafeResize)
+			if err != nil {
+				return err
+			}
+
+			// Move the GPT alt header to end of disk if needed.
+			if vol.IsVMBlock() {
+				err = d.moveGPTAltHeader(devPath)
+				if err != nil {
+					return err
+				}
+			}
+		}
+
+		if vol.contentType == ContentTypeFS {
+			// Run EnsureMountPath again after mounting and filling to ensure the mount directory has
+			// the correct permissions set.
+			err = vol.EnsureMountPath()
+			if err != nil {
+				return err
+			}
+		}
+
+		return nil
+	}, op)
+	if err != nil {
+		return err
+	}
+
+	revert.Success()
+	return nil
+}
+
+// CreateVolumeFromBackup re-creates a volume from its exported state.
+func (d *hpeAlletra) CreateVolumeFromBackup(vol VolumeCopy, srcBackup backup.Info, srcData io.ReadSeeker, op *operations.Operation) (VolumePostHook, revert.Hook, error) {
+	return genericVFSBackupUnpack(d, d.state, vol, srcBackup.Snapshots, srcData, op)
+}
+
+// CreateVolumeFromCopy provides same-pool volume copying functionality.
+func (d *hpeAlletra) CreateVolumeFromCopy(vol VolumeCopy, srcVol VolumeCopy, allowInconsistent bool, op *operations.Operation) error {
+	return fmt.Errorf("CreateVolumeFromCopy: unsupported operation")
+}
+
+// CreateVolumeFromMigration creates a volume being sent via a migration.
+func (d *hpeAlletra) CreateVolumeFromMigration(vol VolumeCopy, conn io.ReadWriteCloser, volTargetArgs migration.VolumeTargetArgs, preFiller *VolumeFiller, op *operations.Operation) error {
+	return fmt.Errorf("CreateVolumeFromMigration: unsupported operation")
+}
+
+// RefreshVolume provides same-pool volume and specific snapshots syncing functionality.
+func (d *hpeAlletra) RefreshVolume(vol VolumeCopy, srcVol VolumeCopy, refreshSnapshots []string, allowInconsistent bool, op *operations.Operation) error {
+	return fmt.Errorf("RefreshVolume: unsupported operation")
+}
+
+// DeleteVolume deletes the volume and all associated snapshots.
+// TODO: copied as-is from pure. refactor?
+// Q: why in some drivers it removes only volume without snapshots?
+func (d *hpeAlletra) DeleteVolume(vol Volume, op *operations.Operation) error {
+	volExists, err := d.HasVolume(vol)
+	if err != nil {
+		return err
+	}
+
+	if !volExists {
+		return nil
+	}
+
+	volName, err := d.getVolumeName(vol)
+	if err != nil {
+		return err
+	}
+
+	host, err := d.client().getCurrentHost()
+	if err != nil {
+		// If the host doesn't exist, continue with the deletion of
+		// the volume and do not try to delete the volume mapping as
+		// it cannot exist.
+		if !api.StatusErrorCheck(err, http.StatusNotFound) {
+			return err
+		}
+	} else {
+		// Delete the volume mapping with the host.
+		err = d.client().disconnectHostFromVolume(vol.pool, volName, host.Name)
+		if err != nil && !api.StatusErrorCheck(err, http.StatusNotFound) {
+			return err
+		}
+	}
+
+	err = d.client().deleteVolume(vol.pool, volName)
+	if err != nil {
+		return err
+	}
+
+	// For VMs, also delete the filesystem volume.
+	if vol.IsVMBlock() {
+		fsVol := vol.NewVMBlockFilesystemVolume()
+
+		err := d.DeleteVolume(fsVol, op)
+		if err != nil {
+			return err
+		}
+	}
+
+	mountPath := vol.MountPath()
+
+	if vol.contentType == ContentTypeFS && shared.PathExists(mountPath) {
+		err := wipeDirectory(mountPath)
+		if err != nil {
+			return err
+		}
+
+		err = os.Remove(mountPath)
+		if err != nil && !os.IsNotExist(err) {
+			return fmt.Errorf("Failed to remove %q: %w", mountPath, err)
+		}
+	}
+
+	return nil
+}
+
+// HasVolume indicates whether a specific volume exists on the storage pool.
+// TODO: copied as-is from pure. refactor?
+func (d *hpeAlletra) HasVolume(vol Volume) (bool, error) {
+	volName, err := d.getVolumeName(vol)
+	if err != nil {
+		return false, err
+	}
+
+	// If volume represents a snapshot, also retrieve (encoded) volume name of the parent,
+	// and check if the snapshot exists.
+	if vol.IsSnapshot() {
+		parentVol := vol.GetParent()
+		parentVolName, err := d.getVolumeName(parentVol)
+		if err != nil {
+			return false, err
+		}
+
+		_, err = d.client().getVolumeSnapshot(vol.pool, parentVolName, volName)
+		if err != nil {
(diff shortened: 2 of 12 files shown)


lxd/storage/drivers/driver_hpe_alletra.go
```diff
@@ -0,0 +1,286 @@
...
+// hpeAlletraSupportedConnectors represents a list of storage connectors that can be used.
+var hpeAlletraSupportedConnectors = []string{
+	connectors.TypeISCSI,
+	connectors.TypeNVME,
+}
+
+type hpeAlletra struct {
+	common
+
+	// Holds the low level connector (iSCSI, NVMEoTCP).
```
roosterfish: `NVMe/TCP` is the term I have observed most of the times.

lxd/storage/drivers/driver_hpe_alletra.go
```diff
@@ -0,0 +1,286 @@
...
+	connector, err := d.connector()
+	if err == nil {
+		_ = connector.LoadModules()
+	}
+
+	return nil
+}
+
+// connector retrieves an initialized storage connector based on the configured
+// storage driver mode (iSCSI, NVMEoTCP). The connector is cached in the driver struct.
```
roosterfish: NVMe/TCP.

lxd/storage/drivers/driver_hpe_alletra.go
```diff
@@ -0,0 +1,286 @@
...
+// isRemote returns true indicating this driver uses remote storage.
+func (d *hpeAlletra) isRemote() bool {
+	return true
+}
+
+// Info returns info about the driver and its environment.
+func (d *hpeAlletra) Info() Info {
+	return Info{
+		Name:                         "hpeAlletra",
+		Version:                      "1",
```
roosterfish: Instead of 1 this should show the version of the tooling, see the PowerFlex and Pure drivers for reference:

```go
versions := connectors.GetSupportedVersions(powerflexSupportedConnectors)
powerFlexVersion = strings.Join(versions, " / ")
```

lxd/storage/drivers/driver_hpe_alletra.go
```diff
@@ -0,0 +1,286 @@
...
+
+// isRemote returns true indicating this driver uses remote storage.
+func (d *hpeAlletra) isRemote() bool {
+	return true
+}
+
+// Info returns info about the driver and its environment.
+func (d *hpeAlletra) Info() Info {
+	return Info{
+		Name:                         "hpeAlletra",
```
roosterfish: I wonder if we should just call it `alletra`. Also in case of PowerFlex we don't mention Dell in driver name for simplicity and just call it `powerflex`, but we do in the documentation so that all written text about the driver uses the correct vendor and product name together. 

IMHO it's unlikely we will ever face name collisions here as the vendors probably won't come up with storage arrays that share the same name with competitors.

lxd/storage/drivers/driver_hpe_alletra.go
```diff
@@ -0,0 +1,286 @@
...
+
+// Validate checks that all provide keys are supported and that no conflicting or missing configuration is present.
+func (d *hpeAlletra) Validate(config map[string]string) error {
+	rules := map[string]func(value string) error{
+		// lxdmeta:generate(entities=storage-hpe-alletra; group=pool-conf; key=hpe_alletra.wsapi.url)
+		//
+		// ---
+		//  type: string
+		//  shortdesc: Address of the HPE Alletra Storage UI/WSAPI
+		"hpe_alletra.wsapi.url": validate.Optional(validate.IsRequestURL),
```
roosterfish: That was the reason I wondered about the name. I think all of those config keys would be much cleaner if we just name the driver `alletra`.

lxd/storage/drivers/driver_hpe_alletra.go
roosterfish: See https://documentation.ubuntu.com/lxd/latest/reference/storage_drivers/

lxd/storage/drivers/driver_hpe_alletra.go
```diff
@@ -0,0 +1,286 @@
...
+		return err
+	}
+
+	newMode := config["hpe_alletra.mode"]
+	oldMode := d.config["hpe_alletra.mode"]
+
+	// Ensure hpe_alletra.mode cannot be changed to avoid leaving volume mappings
+	// and prevent disturbing running instances.
+	if oldMode != "" && oldMode != newMode {
+		return errors.New("HPE Alletra Storage mode cannot be changed")
```
roosterfish: Also when we settle on `alletra` for the name, think also all error messages can just refer to `Alletra`.

lxd/storage/drivers/driver_hpe_alletra.go
```diff
@@ -0,0 +1,286 @@
...
+
+	// Validate both pool and gateway here and return an error if they are not set.
+	// Since those aren't any cluster member specific keys the general validation
+	// rules allow empty strings in order to create the pending storage pools.
+	if d.config["hpe_alletra.wsapi.url"] == "" {
+		return errors.New("The hpe_alletra.wsapi.url cannot be empty")
+	}
+
+	if d.config["hpe_alletra.wsapi.user.name"] == "" {
+		return errors.New("The hpe_alletra.wsapi.user.name cannot be empty")
```
roosterfish: In case of PowerFlex I found that usually (after deployment and also in the examples) there always is an `admin` user so I set this one as a default. Is there something equivalent in Alletra too?

lxd/storage/drivers/driver_hpe_alletra.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	if d.config["hpe_alletra.wsapi.user.password"] == "" {
+		return errors.New("The hpe_alletra.wsapi.user.password cannot be empty")
+	}
+
+	// TODO: do we need to *actually* go to WSAPI and create something?
```
roosterfish: Probably you can do this on demand. But it would help to understand if the API is actually reachable and to validate if user/pass is correct.

roosterfish: Otherwise you might be able to create the pool, but only after creating the first vol/instance you know if the access is working as LXD reaches out to the array.

lxd/storage/drivers/driver_hpe_alletra_volumes.go
```diff
@@ -0,0 +1,977 @@
...
+	"github.com/canonical/lxd/shared/units"
+	"github.com/canonical/lxd/shared/validate"
+	"github.com/google/uuid"
+)
+
+// factorMiB divides a byte size value into Mebibytes.
+const factorMiB = 1024 * 1024
+
+// hpeAlletraVolTypePrefixes maps volume type to storage volume name prefix.
+// Use smallest possible prefixes since Pure Storage volume names are limited to 63 characters.
```
roosterfish: Alletra

lxd/storage/drivers/driver_hpe_alletra_volumes.go
```diff
@@ -0,0 +1,977 @@
...
+	}
+
+	return volName, nil
+}
+
+// ensureHost returns a name of the host that is configured with a given IQN. If such host
+// does not exist, a new one is created, where host's name equals to the server name with a
+// mode included as a suffix because Pure Storage does not allow mixing IQNs, NQNs, and WWNs
+// on a single host.
+// TODO: copied as-is from pure. refactor?
```
roosterfish: Should we start introducing common helpers for those if it's really a copy/paste? Maybe you could extend this new interface that you are adding in https://github.com/canonical/lxd/pull/15885.

roosterfish: I think that would be especially useful for the map/unmap operations as working out those took lots of time in the past and maybe all drivers can benefit from improvements fixes in the common helpers.

lxd/storage/drivers/driver_hpe_alletra_volumes.go
```diff
@@ -0,0 +1,977 @@
...
+		//  defaultdesc: `10GiB`
+		//  shortdesc: Size/quota of the storage volume
+		"volume.size": validate.Optional(validate.IsMultipleOfUnit("256MiB")),
+	}
+}
+
+// ValidateVolume validates the supplied volume config.
+// TODO: copied as-is from pure. refactor?
+func (d *hpeAlletra) ValidateVolume(vol Volume, removeUnknownKeys bool) error {
+	// When creating volumes from ISO images, round its size to the next multiple of 256MiB.
```
roosterfish: IIRC I have added this rounding initially to the PowerFlex driver for the multiples of 8GiB. Let's add a util function which accepts an initial value for the "multiples of" and the actual `vol.ConfigSize()` currently set.

roosterfish: That should then work for all PowerFlex, Pure and Alletra.

lxd/storage/drivers/driver_hpe_alletra_volumes.go
```diff
@@ -0,0 +1,977 @@
...
+}
+
+// RenameVolumeSnapshot renames a volume snapshot.
+func (d *hpeAlletra) RenameVolumeSnapshot(snapVol Volume, newSnapshotName string, op *operations.Operation) error {
+	// Renaming a volume snapshot won't change an actual name of the HPE Alletra volume snapshot.
+	return nil
+}
+
+// getNVMeTargetQN discovers the targetQN used for the given addresses.
+// The targetQN is unqiue per PowerFlex storage pool.
```
roosterfish: Alletra

lxd/storage/drivers/driver_hpe_alletra_volumes.go
```diff
@@ -0,0 +1,977 @@
...
+	if d.nvmeTargetQN == "" {
+		connector, err := d.connector()
+		if err != nil {
+			return "", err
+		}
+
+		// The discovery log from the first reachable target address is returned.
+		discoveryLogRecords, err := connectors.Discover(d.state.ShutdownCtx, connector, d.state.ServerUUID, targetAddresses...)
+		if err != nil {
+			return "", fmt.Errorf("Failed to discover SDT NQN: %w", err)
```
roosterfish: SDT is PowerFlex's name for NVMe target.

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
```diff
@@ -0,0 +1,1056 @@
...
+
+// hpeServiceNameMapping maps HPE Storage mode in LXD to the corresponding HPE Storage
+// service name.
+var hpeServiceNameMapping = map[string]string{
+	connectors.TypeISCSI: "iscsi",
+	connectors.TypeNVME:  "nvme-tcp",
+}
+
+type hpeVolume struct {
+	ID                    int         `json:"id"`
```
roosterfish: Do we require all fields from this struct? Maybe we can start only with the ones actually required?

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
```diff
@@ -0,0 +1,1056 @@
...
+		}
+	}
+
+	return false
+}
+
+// hpeIsNotFoundError returns true if the error is of type hpeError, its status code is 400 (bad request),
+// and the error message contains a substring indicating the resource was not found.
+func isHpeErrorNotFound(err error) bool {
+	return isHpeErrorOf(err, http.StatusNotFound, "Not found", "Does not exist", "No such volume or snapshot")
```
roosterfish: I wonder if this might hide important information down the line if we generalize all "not found"'s.
Also `isHpeErrorNotFound` seems to be used only in two places to check for 404 and then return a custom error message. Should we rather just relay the error from the API directly? Maybe that is more helpful for debugging because it's using the vendor terms instead of our own wrapped error.

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
```diff
@@ -0,0 +1,1056 @@
...
+}
+
+// request issues a HTTP request against HPE Storage WSAPI
+func (p *hpeAlletraClient) request(method string, url url.URL, reqBody map[string]any, reqHeaders map[string]string, respBody any, respHeaders map[string]string) error {
+	logger.Debug("HPE request()")
+	// Extract scheme and host from the gateway URL.
+	scheme, host, found := strings.Cut(p.driver.config["hpe_alletra.wsapi.url"], "://")
```
roosterfish: Looks this is coming from Pure driver. PowerFlex is doing `http.NewRequest` and simply passes the address of the array (including the scheme). 
@MusicDin was there a specific reason to split the address and perform those steps manually?

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
```diff
@@ -0,0 +1,1056 @@
...
+
+	// logger.Debugf("Request Body:\n%s", reqBody)
+
+	// reqBodyDisplay := make([]string, 0, len(reqBody))
+	// for k, v := range reqBody {
+	// 	reqBodyDisplay = append(reqBodyDisplay, fmt.Sprintf("%s=%v", k, v))
+	// }
+	// logger.Debugf("Request Body (KV):\n%s", strings.Join(reqBodyDisplay, "\n"))
+
+	reqBodyJSON, err := json.MarshalIndent(reqBody, "", "    ")
```
roosterfish: Do we have to indent the JSON? I assume it's for the debugging line below only.

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
```diff
@@ -0,0 +1,1056 @@
...
+		"Content-Security-Policy",
+		"Strict-Transport-Security",
+		"Cache-Control",
+		"Vary",
+		"X-Content-Type-Options",
+		"Content-Language",
+		"X-Xss-Protection",
+		"Pragma",
+		"Accept-Ranges",
+		"Last-Modified",
```
roosterfish: I wonder where this list is coming from? Do you think we should keep those?

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
```diff
@@ -0,0 +1,1056 @@
...
+	}
+	logger.Debugf("Response Headers (filtered):\n%s", parsedRespHeaders)
+
+	var prettyBody any
+	if json.Unmarshal(bodyBytes, &prettyBody) == nil {
+		b, _ := json.MarshalIndent(prettyBody, "", "  ")
+		lines := strings.Split(string(b), "\n")
+		logger.Debugf("Response Body size (JSON): %d", len(b))
+		var outputBuffer strings.Builder
+		if len(lines) > 100000 {
```
roosterfish: Can't we just print the indented response body when on debug log level? I don't understand this complexity with all the lines/loops etc.

Maybe if you come to this section of the code, can you check if we need all of this and clean it up?

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
```diff
@@ -0,0 +1,1056 @@
...
+		logger.Debugf("Response Body size (RAW): %d", len(bodyBytes))
+		logger.Debugf("Response Body (RAW):\n%s", string(bodyBytes))
+	}
+
+	// The unauthorized error is reported when an invalid (or expired) access token is provided.
+	// Wrap unauthorized requests into an API status error to allow easier checking for expired
+	// token in the requestAuthenticated function.
+	// if resp.StatusCode == http.StatusUnauthorized {
+	// 	return api.StatusErrorf(http.StatusUnauthorized, "xxx 1 Unauthorized request")
+	// } else if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusCreated {
```
roosterfish: This part is crucial when it comes to expired tokens. Usually the `request` func returns the 401 error so that the outer `requestAuthenticated` can then initiate a token refresh and will retry the request.

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
```diff
@@ -0,0 +1,1056 @@
...
+	url := api.NewURL().Path("api", "v1", "credentials")
+	respHeaders := make(map[string]string)
+
+	err := p.request(http.MethodPost, url.URL, body, nil, &respBody, respHeaders)
+	if err != nil {
+		return fmt.Errorf("HPE failed to send login request: %w", err)
+	}
+
+	if respBody.Key != "" {
+		p.sessionKey = respBody.Key
```
roosterfish: Do they call it session key in the API, or is it just a key/token?

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
```diff
@@ -0,0 +1,1056 @@
...
+	return nil
+}
+
+func (p *hpeAlletraClient) createVolume(poolName string, volName string, sizeBytes int64) error {
+	err := p._createVolume(poolName, volName, sizeBytes)
+	if err != nil {
+		return err
+	}
+
+	err = p.modifyVolumeSet(1, volName) // 1 = memAdd
```
roosterfish: This always implicitly adds the created vol to the volume set. Should we rather call the `modifyVolumeSet` on demand? This would also allow getting rid of the `_` prefixed `createVolume` which feels odd.

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
```diff
@@ -0,0 +1,1056 @@
...
+
+		logger.Debugf("HPE debug failed to delete hpeVolume %s from CPG %s", volName, poolName)
+		return fmt.Errorf("Failed to delete hpeVolume %q in CPG %q: %w", volName, poolName, err)
+	}
+
+	return nil
+}
+
+func (p *hpeAlletraClient) deleteVolume(poolName string, volName string) error {
+	err := p.modifyVolumeSet(2, volName) // 2 = memRemove
```
roosterfish: Same here regarding the implicit removal from the volume set.
Btw do we have to use the volume sets now or can we just create/delete vols without the vol sets?

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
```diff
@@ -0,0 +1,1056 @@
...
+
+	return true, nil
+}
+
+// getVLUN returns VLUN related data of given volumeName
+func (p *hpeAlletraClient) getVLUN(volumeName string) (*hpeVLUN, error) {
+	logger.Debugf("HPE getVLUN()")
+	var resp hpeRespMembers[hpeVLUN]
+
+	url := api.NewURL().Path("api", "v1", "vluns").WithQuery("query", "\"volumeName"+"=="+volumeName+"\"")
```
roosterfish: Should we have a common "formatter" func that does this conversion of the query param?

roosterfish (changes requested):
Thanks for your efforts on moving commonly used code into generic helpers. I have added some comments for a few more of those.

I know it's still WIP but just a few comments along the way :) It feels the API client requires some cleanups here and there once you get to it.

tomponline:
Please rebase

lxd/storage/drivers/driver_hpe_alletra.go
mihalicyn: Fixed, thanks!

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
mihalicyn: yeah, it is called a session key [1]

[1] https://support.hpe.com/hpesc/public/docDisplay?docId=a00113792en_us&page=v20296988.html

lxd/storage/drivers/driver_hpe_alletra.go
mihalicyn: Fixed, thanks!

lxd/storage/drivers/driver_hpe_alletra.go
mihalicyn: Yeah, sorry it was just a temporary todo. In a current state we create a volume set during storage pool creation explicitly.

lxd/storage/drivers/driver_hpe_alletra_volumes.go
mihalicyn: Yes, I think we can do this refactoring and generalize map/unmap logic the same we've done this for mount/unmount. But I would do it after merging this as are getting closer to a deadline.

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
mihalicyn: Yeah, it is only for debugging and I've removed this part entirely.

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
mihalicyn: It was only for debugging too. Just to filter some useless headers to have less noise in the logs.

Removed in a current version. Thanks!

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
mihalicyn: It was only for debugging too. Removed in a current version. Thanks!

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
mihalicyn: For sure, this is fixed too. Thanks, Julian!

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
mihalicyn: Please, can you elaborate on this one?

roosterfish: Mh not sure anymore what was my exact thinking here. 
Probably I was thinking about the points raised by the consultant that you need to be careful when formatting/encoding the `query` so that it's still a valid URL even with special characters in the volume's name. 

In case there are special characters in the vols name, does this break the URL? Or is this filtered out beforehand?

lxd/storage/drivers/driver_hpe_alletra.go
mihalicyn: Have done. Thanks!

mihalicyn:
> Please rebase
> Please fix linter errors.

Have done. Thanks!

lxd/metadata/configuration.json
```diff
@@ -5614,6 +5614,102 @@
...
+						}
+					},
+					{
+						"alletra.wsapi.verify": {
+							"defaultdesc": "`true`",
+							"longdesc": "",
+							"shortdesc": "Whether to verify the HPE Alletra Storage UI/WSAPI certificate",
+							"type": "bool"
+						}
+					},
```
MusicDin: Just wondering whether we could use the following names to make it consistent with Powerflex and Pure?

`alletra.wsapi.url` -> `alletra.gateway`
`alletra.wsapi.verify` -> `alletra.gateway.verify`
`alletra.wsapi.user` -> `alletra.user.name`
`alletra.wsapi.password` -> `alletra.user.password`

lxd/storage/drivers/driver_alletra.go
```diff
@@ -0,0 +1,279 @@
...
+	// Holds the low level connector (iSCSI, NVMEoTCP).
+	// Use .connector() method to retrieve the initialized connector.
+	storageConnector connectors.Connector
+
+	// Holds the targetQN of NVMe target.
+	nvmeTargetQN string
+
+	// Holds the low level HTTP client for the HPE Alletra Storage API.
+	// Use .client() method to retrieve the client struct.
+	httpClient *alletraClient
```
tomponline: This commit is too early in the PR as it is referencing a type, alletraClient, that has not been added yet.

lxd/storage/drivers/driver_alletra_wsapi1.go
tomponline, line 1: As discussed in our 1:1 lets create a `lxd/storage/drivers/clients` package and move the alletra client into that.

In the future we can move the pure and powerflex clients in there as well (cc @roosterfish @MusicDin ).

This will allow us to share code between clients easily, and allow isolation/clarification of which functions the storage driver itself should use (exported) and which ones are for internal usage (unexported).

lxd/storage/drivers/driver_alletra.go
```diff
@@ -0,0 +1,279 @@
...
+
+// Validate checks that all provide keys are supported and that no conflicting or missing configuration is present.
+func (d *alletra) Validate(config map[string]string) error {
+	rules := map[string]func(value string) error{
+		// lxdmeta:generate(entities=storage-alletra; group=pool-conf; key=alletra.wsapi.url)
+		//
+		// ---
+		//  type: string
+		//  shortdesc: Address of the HPE Alletra Storage UI/WSAPI
+		"alletra.wsapi.url": validate.Optional(validate.IsRequestURL),
```
tomponline: do we need the `wsapi` prefix on these settings?

lxd/storage/drivers/driver_alletra.go
```diff
@@ -0,0 +1,279 @@
...
+		//
+		// ---
+		//  type: string
+		//  shortdesc: HPE Alletra storage admin password
+		"alletra.wsapi.user.password": validate.IsAny,
+		// lxdmeta:generate(entities=storage-alletra; group=pool-conf; key=alletra.wsapi.cpg)
+		//
+		// ---
+		//  type: string
+		//  shortdesc: HPE Alletra CPG name
```
tomponline: Please can we expand the acronym CPG in the description.

lxd/storage/drivers/driver_alletra.go
```diff
@@ -0,0 +1,279 @@
...
+		return nil
+	}
+
+	// On delete, wipe everything in the directory.
+	return wipeDirectory(GetPoolMountPath(d.name))
+}
+
+// Update applies any driver changes required from a configuration change.
+func (d *alletra) Update(changedConfig map[string]string) error {
+	return nil
```
tomponline: should this return an ErrNotSupported error?

lxd/storage/drivers/driver_alletra.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+// Unmount unmounts the storage pool.
+func (d *alletra) Unmount() (bool, error) {
+	return true, nil
+}
+
+// GetResources returns the pool resource usage information.
+func (d *alletra) GetResources() (*api.ResourcesStoragePool, error) {
+	return nil, nil
```
tomponline: @mihalicyn this to be implemented in a future PR? Should probably return error ErrNotSupported for now.

lxd/storage/drivers/driver_alletra_volumes.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// GetVolumeUsage returns the disk space used by the volume.
+func (d *alletra) GetVolumeUsage(vol Volume) (int64, error) {
+	return 0, errors.New("GetVolumeUsage: unsupported operation")
+}
+
+// SetVolumeQuota applies a size limit on volume.
+func (d *alletra) SetVolumeQuota(vol Volume, size string, allowUnsafeResize bool, op *operations.Operation) error {
+	return errors.New("SetVolumeQuota: unsupported operation")
```
tomponline: You can return ErrNotSupported here

lxd/storage/drivers/driver_alletra_volumes.go
```diff
@@ -0,0 +1,174 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// GetVolumeUsage returns the disk space used by the volume.
+func (d *alletra) GetVolumeUsage(vol Volume) (int64, error) {
+	return 0, errors.New("GetVolumeUsage: unsupported operation")
```
tomponline: You can return ErrNotSupported here

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -11,3 +40,237 @@ func newAlletraClient(driver *alletra) *alletraClient {
...
+			return true
+		}
+	}
+
+	return false
+}
+
+// hpeIsNotFoundError returns true if the error is of type hpeError, its status code is 400 (bad request),
+// and the error message contains a substring indicating the resource was not found.
+func isHpeErrorNotFound(err error) bool {
```
tomponline: When we have the new `clients` package, this can be moved inside it, and then used to convert HPE specific not found errors to `api.StatusErrorf(http.StatusNotFound)` errors for use in the driver itself (and beyond allowing it to bubble up to influence the HTTP return code in the API if needed).

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -11,3 +40,237 @@ func newAlletraClient(driver *alletra) *alletraClient {
...
+				// Session key seems to be expired.
+				// Reset the token and try one more time.
+				p.sessionKey = ""
+				retries--
+				continue
+			}
+
+			hpeErr, assert := err.(*hpeError)
+			if assert {
+				logger.Debugf("HPE debug hpeErr.statusCode: %d", hpeErr.HTTPStatusCode)
```
tomponline: Lets remove these debug statements, or if they are needed, lets switch to a single contextual log line.

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -11,3 +40,237 @@ func newAlletraClient(driver *alletra) *alletraClient {
...
+
+	var respBody struct {
+		Key  string `json:"key"`
+		Desc string `json:"desc"`
+	}
+
+	body := map[string]any{
+		"user":        p.driver.config["alletra.wsapi.user.name"],
+		"password":    p.driver.config["alletra.wsapi.user.password"],
+		"sessionType": 1, // REGULAR
```
tomponline: As discussed in our 1:1 lets define constants for these in the new client package, rather than having magic numbers with comments.

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -274,3 +274,43 @@ func (p *alletraClient) login() error {
 	p.sessionKey = respBody.Key
 	return nil
 }
+
+func (p *alletraClient) createVolumeSet() error {
+	req := map[string]any{
+		"name":    p.driver.name,
+		"comment": "Created and managed by LXD",
```
tomponline: Just leaving a note here to ensure we consider and implement adding a `comment` for volumes and snapshots too - what will be in that comment still remains to be discussed.

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -274,3 +274,43 @@ func (p *alletraClient) login() error {
...
+	url := api.NewURL().Path("api", "v1", "volumesets", p.driver.name)
+	err := p.requestAuthenticated(http.MethodDelete, url.URL, nil, nil)
+	if err != nil {
+		return fmt.Errorf("Failed to delete volume set for a storage pool %q: %w", p.driver.name, err)
+	}
+
+	return nil
+}
+
+func (p *alletraClient) modifyVolumeSet(action int, volName string) error {
```
tomponline: Could you define a type for `action` and constants for the various actions supported. Then we can find the various values easily. Ta

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -314,3 +356,130 @@ func (p *alletraClient) modifyVolumeSet(action int, volName string) error {
...
+	url := api.NewURL().Path("api", "v1", "hosts")
+	err = p.requestAuthenticated(http.MethodPost, url.URL, req, nil)
+	if err != nil {
+		hpeErr, ok := err.(*hpeError)
+		if ok {
+			switch hpeErr.Code {
+			case 16: // EXISTENT_HOST
+				return nil
+			default:
+				return fmt.Errorf("HPE debug. hpeErr.Code: %d. hpeErr.Desc: %s", hpeErr.Code, hpeErr.Desc)
```
tomponline: This is an odd error, debug?

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -314,3 +356,130 @@ func (p *alletraClient) modifyVolumeSet(action int, volName string) error {
...
+		return fmt.Errorf("Unsupported HPE Alletra Storage mode %q", connector.Type())
+	}
+
+	url := api.NewURL().Path("api", "v1", "hosts")
+	err = p.requestAuthenticated(http.MethodPost, url.URL, req, nil)
+	if err != nil {
+		hpeErr, ok := err.(*hpeError)
+		if ok {
+			switch hpeErr.Code {
+			case 16: // EXISTENT_HOST
```
tomponline: constant needed

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -314,3 +356,130 @@ func (p *alletraClient) modifyVolumeSet(action int, volName string) error {
...
+
+// deleteHost deletes an existing host.
+func (p *alletraClient) deleteHost(hostName string) error {
+	url := api.NewURL().Path("api", "v1", "hosts", hostName)
+	err := p.requestAuthenticated(http.MethodDelete, url.URL, nil, nil)
+	if err != nil {
+		hpeErr, ok := err.(*hpeError)
+		if ok {
+			switch hpeErr.Code {
+			case 26: // EXPORTED_VLUN
```
tomponline: constant needed

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -483,3 +533,91 @@ func (p *alletraClient) deleteHost(hostName string) error {
 func (p *alletraClient) updateHost(hostName string, qns []string) error {
 	return fmt.Errorf("Failed to update host %q. Operation not supported.", hostName)
 }
+
+func (p *alletraClient) _createVolume(poolName string, volName string, sizeBytes int64) error {
+	req := map[string]any{
+		"name":    volName,
+		"cpg":     p.driver.config["alletra.wsapi.cpg"],
+		"sizeMiB": sizeBytes / 1024 / 1024,
+		"tpvv":    true, // thinly provisioned volume
```
tomponline: love this comment, as would have no idea what `tpvv` was :)

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -483,3 +533,91 @@ func (p *alletraClient) deleteHost(hostName string) error {
 func (p *alletraClient) updateHost(hostName string, qns []string) error {
 	return fmt.Errorf("Failed to update host %q. Operation not supported.", hostName)
 }
+
+func (p *alletraClient) _createVolume(poolName string, volName string, sizeBytes int64) error {
```
tomponline: As discussed, when this is moved into the `clients` package, we can then differentiate between external and internal functions, by having an exported `CreateVolume()`  function and then renaming this to `createVolume()` for internal use.

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -483,3 +533,91 @@ func (p *alletraClient) deleteHost(hostName string) error {
...
+// createVolume creates a new volume in the given storage pool. The volume is created with
+// supplied size in bytes. Upon successful creation, volume's ID is returned.
+func (p *alletraClient) createVolume(poolName string, volName string, sizeBytes int64) error {
+	err := p._createVolume(poolName, volName, sizeBytes)
+	if err != nil {
+		return err
+	}
+
+	// Add a newly created volume to a volume set
+	err = p.modifyVolumeSet(1, volName) // 1 = memAdd
```
tomponline: constant needed

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -483,3 +533,91 @@ func (p *alletraClient) deleteHost(hostName string) error {
...
+func (p *alletraClient) getVolume(poolName string, volName string) (*hpeVolume, error) {
+	var resp hpeVolume
+
+	url := api.NewURL().Path("api", "v1", "volumes", volName)
+	err := p.requestAuthenticated(http.MethodGet, url.URL, nil, &resp)
+	if err != nil {
+		hpeErr, ok := err.(*hpeError)
+		if ok {
+			switch hpeErr.Code {
+			case 23: // NON_EXISTENT_VOL
```
tomponline: constant needed

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -483,3 +533,91 @@ func (p *alletraClient) deleteHost(hostName string) error {
...
+
+func (p *alletraClient) _deleteVolume(poolName string, volName string) error {
+	url := api.NewURL().Path("api", "v1", "volumes", volName)
+
+	err := p.requestAuthenticated(http.MethodDelete, url.URL, nil, nil)
+	if err != nil {
+		hpeErr, ok := err.(*hpeError)
+		if ok {
+			switch hpeErr.Code {
+			case 23: // NON_EXISTENT_VOL
```
tomponline: constant needed

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -483,3 +533,91 @@ func (p *alletraClient) deleteHost(hostName string) error {
...
+
+	return &resp, nil
+}
+
+func (p *alletraClient) _deleteVolume(poolName string, volName string) error {
+	url := api.NewURL().Path("api", "v1", "volumes", volName)
+
+	err := p.requestAuthenticated(http.MethodDelete, url.URL, nil, nil)
+	if err != nil {
+		hpeErr, ok := err.(*hpeError)
```
tomponline: I'm seeing this pattern duplicated a fair bit, sounds like we could do with a helper function that detects specific errors, similar to `api.StatusErrorCheck`

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -483,3 +533,91 @@ func (p *alletraClient) deleteHost(hostName string) error {
...
+
+		return fmt.Errorf("Failed to delete hpeVolume %q in pool %q: %w", volName, poolName, err)
+	}
+
+	return nil
+}
+
+// deleteVolume deletes an exisiting volume in the given storage pool.
+func (p *alletraClient) deleteVolume(poolName string, volName string) error {
+	err := p.modifyVolumeSet(2, volName) // 2 = memRemove
```
tomponline: constant needed

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -621,3 +628,94 @@ func (p *alletraClient) deleteVolume(poolName string, volName string) error {
...
+	req["volumeName"] = volName
+	req["lun"] = 0
+	req["autoLun"] = true
+
+	err := p.requestAuthenticated(http.MethodPost, url.URL, req, nil)
+	if err != nil {
+		hpeErr, ok := err.(*hpeError)
+		if ok {
+			switch hpeErr.Code {
+			case 29: // Error: VVs cannot be exported to the same NVMe host twice.
```
tomponline: constant needed

tomponline: VVs?

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -621,3 +628,94 @@ func (p *alletraClient) deleteVolume(poolName string, volName string) error {
...
+
+// getVLUN returns VLUN related data of given volumeName.
+func (p *alletraClient) getVLUN(volumeName string) (*hpeVLUN, error) {
+	var resp hpeRespMembers[hpeVLUN]
+
+	url := api.NewURL().Path("api", "v1", "vluns").WithQuery("query", "\"volumeName"+"=="+volumeName+"\"")
+
+	err := p.requestAuthenticated(http.MethodGet, url.URL, nil, &resp)
+	if err != nil {
+		return nil, fmt.Errorf("Fail to get vLUN data for volume %s: %w", volumeName, err)
```
tomponline: %q rather than %s

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -621,3 +628,94 @@ func (p *alletraClient) deleteVolume(poolName string, volName string) error {
...
+
+// disconnectHostFromVolume deletes a connection between a host and volume.
+func (p *alletraClient) disconnectHostFromVolume(poolName string, volName string, hostName string) error {
+	vlun, errVLUN := p.getVLUN(volName)
+	if errVLUN != nil {
+		return fmt.Errorf("HPE Error %w", errVLUN)
+	}
+
+	if vlun == nil {
+		logger.Debugf("HPE debug no need to disconnect host %s from volume %s", hostName, volName)
```
tomponline: %q rather than %s

tomponline: although we probably need to remove the debug lines anyway

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -621,3 +628,94 @@ func (p *alletraClient) deleteVolume(poolName string, volName string) error {
...
+
+// getVLUNsForHost returns VLUNs list for a given hostName.
+func (p *alletraClient) getVLUNsForHost(hostName string) ([]hpeVLUN, error) {
+	var resp hpeRespMembers[hpeVLUN]
+
+	url := api.NewURL().Path("api", "v1", "vluns").WithQuery("query", "\"hostname"+"=="+hostName+"\"")
+
+	err := p.requestAuthenticated(http.MethodGet, url.URL, nil, &resp)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get vLUNs list for host %s: %w", hostName, err)
```
tomponline: %q instead of %s

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -629,6 +636,81 @@ func (p *alletraClient) deleteVolume(poolName string, volName string) error {
...
+		}
+
+		for _, member := range portData.Members {
+			if member.LinkState != 4 {
+				continue // skip down or unlinked ports
+			}
+
+			switch mode {
+			case connectors.TypeISCSI:
+				if member.Protocol != 2 {
```
tomponline: constants for these numbers please

lxd/storage/drivers/driver_alletra_volumes.go
```diff
@@ -346,6 +346,92 @@ func (d *alletra) unmapVolume(vol Volume) error {
...
+	if err != nil {
+		return "", nil, err
+	}
+
+	hpeVol, err := d.client().getVolume(vol.pool, volName)
+	if err != nil {
+		return "", nil, err
+	}
+
+	// Ensure the NGUID is exactly 32 characters long, as it uniquely
```
tomponline, line 403: Thank you for the comment!

lxd/storage/drivers/driver_alletra_volumes.go
```diff
@@ -432,6 +435,220 @@ func (d *alletra) GetVolumeDiskPath(vol Volume) (string, error) {
...
+}
+
+// CreateVolumeFromBackup re-creates a volume from its exported state.
+func (d *alletra) CreateVolumeFromBackup(vol VolumeCopy, srcBackup backup.Info, srcData io.ReadSeeker, op *operations.Operation) (VolumePostHook, revert.Hook, error) {
+	return genericVFSBackupUnpack(d, d.state, vol, srcBackup.Snapshots, srcData, op)
+}
+
+// CreateVolumeFromCopy provides same-pool volume copying functionality.
+func (d *alletra) CreateVolumeFromCopy(vol VolumeCopy, srcVol VolumeCopy, allowInconsistent bool, op *operations.Operation) error {
+	return errors.New("CreateVolumeFromCopy: unsupported operation")
```
tomponline, line 602: ErrNotSupported

lxd/storage/drivers/driver_alletra_volumes.go
```diff
@@ -432,6 +435,220 @@ func (d *alletra) GetVolumeDiskPath(vol Volume) (string, error) {
...
+}
+
+// CreateVolumeFromMigration creates a volume being sent via a migration.
+func (d *alletra) CreateVolumeFromMigration(vol VolumeCopy, conn io.ReadWriteCloser, volTargetArgs migration.VolumeTargetArgs, preFiller *VolumeFiller, op *operations.Operation) error {
+	return errors.New("CreateVolumeFromMigration: unsupported operation")
```
tomponline, line 607: ErrNotSupported

lxd/storage/drivers/driver_alletra_volumes.go
```diff
@@ -435,6 +435,17 @@ func (d *alletra) GetVolumeDiskPath(vol Volume) (string, error) {
 	return "", ErrNotSupported
 }
 
+// MountVolume mounts a volume and increments ref counter. Please call UnmountVolume() when done with the volume.
+func (d *alletra) MountVolume(vol Volume, op *operations.Operation) error {
+	return mountVolume(d, vol, d.getMappedDevPath, op)
```
tomponline, line 461: Yay, de-duplication :)

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -11,3 +41,281 @@ func newAlletraClient(driver *alletra) *alletraClient {
...
+			return api.StatusErrorf(http.StatusForbidden, "Unauthorized request")
+		}
+
+		return hpeErr
+	}
+
+	return nil
+}
+
+var sessionKeys = make(map[string]string)
```
tomponline: what clears old entries from this map?

also, related, as map memory allocations are never released AFAIK, we should seek to make each map entry as small as possible, so rather than `string` it could be `*string`.

mihalicyn: > what clears old entries from this map?

`invalidateSessionKey()` function removes the session key if it is not valid anymore.

lxd/storage/drivers/driver_hpe_alletra_wsapi1.go
mihalicyn: > In case there are special characters in the vols name, does this break the URL? Or is this filtered out beforehand?

yeah, volume names are generated the same way as in Pure. No special characters.

Also, `WithQuery()` includes necessary URL encoding.

roosterfish: Okay thanks.

lxd/storage/drivers/driver_alletra.go
tomponline: @mihalicyn outstanding question.

lxd/storage/drivers/driver_alletra.go
tomponline: @mihalicyn outstanding question.

lxd/storage/drivers/driver_alletra.go
tomponline: @mihalicyn outstanding question.

lxd/storage/drivers/driver_alletra.go
tomponline: @mihalicyn outstanding question.

lxd/storage/drivers/driver_alletra_volumes.go
tomponline: @mihalicyn outstanding question.

lxd/storage/drivers/driver_alletra_volumes.go
tomponline: @mihalicyn outstanding question.

lxd/storage/drivers/clients/alletra_wsapi1.go
```diff
@@ -1,10 +1,332 @@
...
+				// Session key seems to be expired.
+				// Reset the token and try one more time.
+				p.invalidateSessionKey()
+				retries--
+				continue
+			}
+
+			hpeErr, assert := err.(*hpeError)
+			if assert {
+				logger.Debugf("HPE debug hpeErr.statusCode: %d", hpeErr.HTTPStatusCode)
```
tomponline: Lets remove these debug statements please.

lxd/storage/drivers/clients/alletra_wsapi1.go
tomponline, line 1: Please add docblocks to exported functions and variables.

lxd/storage/drivers/clients/alletra_wsapi1.go
```diff
@@ -330,3 +330,43 @@ func (p *AlletraClient) login() error {
...
+	url := api.NewURL().Path("api", "v1", "volumesets", volumeSetName)
+	err := p.requestAuthenticated(http.MethodDelete, url.URL, nil, nil)
+	if err != nil {
+		return fmt.Errorf("Failed to delete volume set for a storage pool %q: %w", volumeSetName, err)
+	}
+
+	return nil
+}
+
+func (p *AlletraClient) modifyVolumeSet(volumeSetName string, action int, volName string) error {
```
tomponline, line 482: although not exported, i think it would be useful to include a comment here as to what this does and what the valid values of `action int` are. Ta

lxd/storage/drivers/clients/alletra_wsapi1.go
```diff
@@ -370,3 +415,113 @@ func (p *AlletraClient) modifyVolumeSet(volumeSetName string, action int, volNam
...
+	hosts, err := p.getHosts()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, host := range hosts {
+		if connectorType == connectors.TypeISCSI {
+			for _, iscsiPath := range host.ISCSIPaths {
+				if iscsiPath.Name == qn {
+					logger.Debugf("HPE iSCSI QN found: %s", qn)
```
tomponline: Lets remove these debug statements please

lxd/storage/drivers/clients/alletra_wsapi1.go
```diff
@@ -370,3 +415,113 @@ func (p *AlletraClient) modifyVolumeSet(volumeSetName string, action int, volNam
...
+					return &host, nil
+				}
+			}
+		}
+
+		if connectorType == connectors.TypeNVME {
+			for _, nvmePath := range host.NVMETCPPaths {
+				if nvmePath.NQN == qn {
+					logger.Debugf("HPE NVMe/TCP QN found: %s", qn)
```
tomponline: Remove debug please

lxd/storage/drivers/clients/alletra_wsapi1.go
```diff
@@ -370,3 +415,113 @@ func (p *AlletraClient) modifyVolumeSet(volumeSetName string, action int, volNam
...
+	url := api.NewURL().Path("api", "v1", "hosts")
+	err := p.requestAuthenticated(http.MethodPost, url.URL, req, nil)
+	if err != nil {
+		hpeErr, ok := err.(*hpeError)
+		if ok {
+			switch hpeErr.Code {
+			case apiErrorExistentHost:
+				return nil
+			default:
+				return fmt.Errorf("HPE debug. hpeErr.Code: %d. hpeErr.Desc: %s", hpeErr.Code, hpeErr.Desc)
```
tomponline: We dont want to be returning debug statements as errors. 
Please can you improve this error message so it makes sense to end users. Thanks

lxd/storage/drivers/clients/alletra_wsapi1.go
```diff
@@ -370,3 +415,113 @@ func (p *AlletraClient) modifyVolumeSet(volumeSetName string, action int, volNam
...
+// DeleteHost deletes an existing host.
+func (p *AlletraClient) DeleteHost(hostName string) error {
+	url := api.NewURL().Path("api", "v1", "hosts", hostName)
+	err := p.requestAuthenticated(http.MethodDelete, url.URL, nil, nil)
+	if err != nil {
+		hpeErr, ok := err.(*hpeError)
+		if ok {
+			switch hpeErr.Code {
+			case apiErrorExportedVLUN:
+				logger.Debugf("HPE debug. Host %s will not be deleted since it has exported volumes", hostName)
```
tomponline: Debug to remove.

lxd/storage/drivers/clients/alletra_wsapi1.go
```diff
@@ -370,3 +415,113 @@ func (p *AlletraClient) modifyVolumeSet(volumeSetName string, action int, volNam
...
+		return fmt.Errorf("Failed to delete host %q: %w", hostName, err)
+	}
+
+	return nil
+}
+
+// UpdateHost updates an existing host. This should be never called
+// and only needed to make code sharing with Pure easier.
+func (p *AlletraClient) UpdateHost(hostName string, qns []string) error {
+	return fmt.Errorf("Failed to update host %q. Operation not supported.", hostName)
```
tomponline: No trailing fullstop on errors please

lxd/storage/drivers/clients/alletra_wsapi1.go
```diff
@@ -525,3 +579,91 @@ func (p *AlletraClient) DeleteHost(hostName string) error {
...
+	var resp hpeVolume
+
+	url := api.NewURL().Path("api", "v1", "volumes", volName)
+	err := p.requestAuthenticated(http.MethodGet, url.URL, nil, &resp)
+	if err != nil {
+		hpeErr, ok := err.(*hpeError)
+		if ok {
+			switch hpeErr.Code {
+			case apiErrorNonExistentVol:
+				return nil, api.StatusErrorf(http.StatusNotFound, "Volume/snapshot %q not found", volName)
```
tomponline: What does "volume/snapshot" mean in this context, is it "volume or snapshot" or "snapshot of volume"?

tomponline (commented):
For the "HPE debug" (as opposed to the debug logging), it makes sense to keep this, but can we improve the message to something like:

> Unexpected HPE appliance API response: ...

Should we also include the URL we were sending too?

lxd/storage/drivers/clients/alletra_wsapi1.go
```diff
@@ -667,3 +675,94 @@ func (p *AlletraClient) DeleteVolume(poolName string, volName string) error {
...
+	req["lun"] = 0
+	req["autoLun"] = true
+
+	err := p.requestAuthenticated(http.MethodPost, url.URL, req, nil)
+	if err != nil {
+		hpeErr, ok := err.(*hpeError)
+		if ok {
+			switch hpeErr.Code {
+			case apiErrorVolumeIsAlreadyExported:
+				logger.Debugf("HPE using existing settings. Volume %s already attached to %s", volName, hostName)
```
tomponline: debug to remove

tomponline (changes requested):
We dont use `logger.Debugf` anymore, and in general if we can avoid adding unnecessary debug statements we should.

However if you feel that some debug logging statements are worth keeping in, then please update the client to accept a logger and then use that.

That way we can pass in a contextual logger with the pool name preconfigured, and then instead of using `logger.Debugf` you should use `l.Debug()` and use contextual logging.

Also avoid calling the l.Debug repeatedly for the same "event" and instead include multiple fields in the context.

lxd/storage/drivers/clients/alletra_wsapi1.go
```diff
@@ -667,3 +675,94 @@ func (p *AlletraClient) DeleteVolume(poolName string, volName string) error {
...
+	}
+
+	return nil
+}
+
+// GetVLUN returns VLUN related data of given volumeName.
+func (p *AlletraClient) GetVLUN(volumeName string) (*hpeVLUN, error) {
+	var resp hpeRespMembers[hpeVLUN]
+
+	url := api.NewURL().Path("api", "v1", "vluns").WithQuery("query", "\"volumeName"+"=="+volumeName+"\"")
```
tomponline: the query string here caught my eye, do we need to quote the `query` value?
If so can https://pkg.go.dev/strconv#Quote do the job?

lxd/storage/drivers/clients/alletra_wsapi1.go
```diff
@@ -676,6 +683,49 @@ func (p *AlletraClient) DeleteVolume(poolName string, volName string) error {
...
 }
 
+func (p *AlletraClient) GetTargetAddrs(connectorType string) (targetAddrs []string, err error) {
+	var portData hpeRespMembers[hpePort]
+
+	apiPorts := api.NewURL().Path("api", "v1", "ports")
+
+	err = p.requestAuthenticated(http.MethodGet, apiPorts.URL, nil, &portData)
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve port list: %w", err)
```
tomponline: Errors should start with capital letter

lxd/storage/drivers/clients/alletra_wsapi1.go
```diff
@@ -676,6 +683,49 @@ func (p *AlletraClient) DeleteVolume(poolName string, volName string) error {
...
+	}
+
+	for _, member := range portData.Members {
+		if member.LinkState != 4 {
+			continue // skip down or unlinked ports
+		}
+
+		switch connectorType {
+		case connectors.TypeISCSI:
+			if member.Protocol != 2 {
```
tomponline: Please can we use constants for these magic numbers (above and below).

lxd/storage/drivers/driver_alletra_volumes.go
```diff
@@ -437,6 +440,237 @@ func (d *alletra) GetVolumeDiskPath(vol Volume) (string, error) {
...
+		return err
+	}
+
+	// Create the volume.
+	err = client.CreateVolume(vol.pool, volName, d.config["alletra.wsapi.cpg"], sizeBytes)
+	if err != nil {
+		return err
+	}
+
+	revert.Add(func() { _ = client.DeleteVolume(vol.pool, volName) })
```
tomponline: @mihalicyn @simondeziel has asked that we start logging warnings/errors inside revert actions that fail (to improve our TICS score).

tomponline (changes requested):
Thanks @mihalicyn 

The new client is what I had in mind, great!

I've gone through a quick pass, but in general:

- logger.Debugf shouldn't appear anyway in the PR. Debug logging is acceptable if needed but we should accept a logger to the client so we can log with context of the pool, and the client itself should use contextual logging.
- Returning debug errors for unexpected API responses is fine, we just need to improve the error message itself.
- Errors start with capital letter and don't have trailing fullstop.
- Errors that occur inside a `defer` should log an error/warning (depending on the impact) so as to improve our TICS score.

mihalicyn:
> * logger.Debugf shouldn't appear anyway in the PR. Debug logging is acceptable if needed but we should accept a logger to the client so we can log with context of the pool, and the client itself should use contextual logging.

yeah, I think I should add a logger to the Constructor for a client

Thanks for review, Tom! I'll fix these and other issues as well.

lxd/storage/drivers/driver_hpe_alletra.go
mihalicyn: AFAIK, there is no default credentials for HPE Alletra.

lxd/metadata/configuration.json
mihalicyn: I'm happy to change this to anything we all agree on.

cc @roosterfish @tomponline

lxd/storage/drivers/driver_alletra_wsapi1.go
mihalicyn: Have done this. Thanks!

lxd/storage/drivers/driver_alletra_wsapi1.go
mihalicyn: Good, sure! We can overwrite contents in there any time, as we never rely/read this comment back from the API. Is it there "just in case" for a system administrator of a storage appliance to know that "this volumeset is there because of LXD". Nothing more.

lxd/storage/drivers/driver_alletra_wsapi1.go
mihalicyn: Yeah, I did it precisely like you've described.

lxd/storage/drivers/driver_alletra_wsapi1.go
tomponline: Sounds good, lets use the same comment as you've added for the volume set for now.

lxd/storage/drivers/driver_alletra_wsapi1.go
```diff
@@ -0,0 +1,803 @@
...
+				continue
+			}
+
+			hpeErr, assert := err.(*hpeError)
+			if assert {
+				logger.Debugf("HPE debug hpeErr.statusCode: %d", hpeErr.HTTPStatusCode)
+				logger.Debugf("HPE debug hpeErr.Code: %d", hpeErr.Code)
+				logger.Debugf("HPE debug hpeErr.Desc: %s", hpeErr.Desc)
+			} else {
+				logger.Debugf("HPE debug cannot assert hpeError: %s", hpeErr)
```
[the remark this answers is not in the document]
mihalicyn: Fixed.

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: Have fixed.

lxd/storage/drivers/driver_alletra_volumes.go
mihalicyn: Have fixed.

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: Have fixed.

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: We don't need to quote in there, but I've switched to raw string literal to make this a bit more readable.

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: Switched to a context logger.

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: Error message was improved.

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: Fixed. Thanks!

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: Switched to a context logger.

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: Have fixed.

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: Done.

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: Done.

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: Have done!

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: Switched to a context logger.

lxd/storage/drivers/driver_alletra_wsapi1.go
mihalicyn: Have done!

lxd/storage/drivers/driver_alletra_wsapi1.go
mihalicyn: Have done.

lxd/storage/drivers/driver_alletra_wsapi1.go
mihalicyn: Have done!

lxd/storage/drivers/driver_alletra_volumes.go
mihalicyn: Have done.

lxd/storage/drivers/driver_alletra_volumes.go
mihalicyn: Addressed.

lxd/storage/drivers/driver_alletra.go
mihalicyn: > this to be implemented in a future PR?

Nope, because there is no clean or good way to obtain such an information. There is a way to get some very-very low level details about disk sizes and CPGs, but nothing to get that statistics for a volume set.

This is not going to be implemented.

lxd/storage/drivers/driver_alletra.go
mihalicyn: Have fixed!

lxd/storage/drivers/driver_alletra.go
mihalicyn: Have fixed!

lxd/storage/drivers/driver_alletra.go
mihalicyn: Removed.

lxd/storage/drivers/driver_alletra.go
mihalicyn: Have fixed.

lxd/metadata/configuration.json
mihalicyn: Have reworked this part. Hope now it is better.

lxd/storage/drivers/driver_hpe_alletra.go
mihalicyn: Have fixed. Thanks!

lxd/storage/drivers/driver_hpe_alletra.go
mihalicyn: Addressed.

lxd/storage/drivers/driver_hpe_alletra_volumes.go
mihalicyn: Fixed ;)

lxd/storage/drivers/driver_hpe_alletra_volumes.go
mihalicyn: Fixed!

lxd/storage/drivers/driver_alletra_wsapi1.go
mihalicyn: Have fixed!

lxd/storage/drivers/driver_alletra_wsapi1.go
mihalicyn: Fixed, thanks!

lxd/storage/drivers/driver_alletra_wsapi1.go
mihalicyn: Fixed, thanks!

lxd/storage/drivers/driver_alletra_wsapi1.go
mihalicyn: Fixed, thanks!

lxd/storage/drivers/driver_alletra_wsapi1.go
mihalicyn: Fixed, thanks!

lxd/storage/drivers/driver_alletra_volumes.go
mihalicyn: Fixed, thanks!

lxd/storage/drivers/driver_alletra_volumes.go
mihalicyn: Fixed, thanks!

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: Fixed, thanks!

lxd/storage/drivers/clients/alletra_wsapi1.go
```diff
@@ -0,0 +1,828 @@
...
+	AdditionalStates      []string    `json:"additionalStates"`
+	TotalReservedMiB      int64       `json:"totalReservedMiB"`
+	TotalUsedMiB          int64       `json:"totalUsedMiB"`
+	SizeMiB               int64       `json:"sizeMiB"`
+	HostWriteMiB          int64       `json:"hostWriteMiB"`
+	WWN                   string      `json:"wwn"`
+	NGUID                 string      `json:"nguid"`
+	CreationTimeSec       int         `json:"creationTimeSec"`
+	CreationTime8601      string      `json:"creationTime8601"`
+	UsrSpcAllocWarningPct int         `json:"usrSpcAllocWarningPct"`
```
roosterfish, line 138: Do we need to carry those unused fields (e.g. `UsrSpcAllocWarningPct`) in the struct or should it be reduced down to only the used ones?

roosterfish (changes requested):

tomponline:
@mihalicyn I cant find the comment/response now, but I think we should discuss further the metrics reporting issue.

Did you say that pool size and usage reporting won't be possible with the approach taken?

lxd/storage/drivers/clients/alletra_wsapi1.go
tomponline: @mihalicyn will this be used in a later PR?

lxd/storage/drivers/driver_alletra_volumes.go
```diff
@@ -75,6 +85,286 @@ func (d *alletra) getVolumeName(vol Volume) (string, error) {
...
+	connCreated, err := d.client().ConnectHostToVolume(vol.pool, volName, hostname)
+	if err != nil {
+		return nil, err
+	}
+
+	if connCreated {
+		reverter.Add(func() {
+			err := d.client().DisconnectHostFromVolume(vol.pool, volName, hostname)
+			if err != nil {
+				d.logger.Error("DisconnectHostFromVolume API call failed on error path", logger.Ctx{"err": err, "volName": volName, "hostname": hostname})
```
tomponline, line 200: Suggest using `d.logger.Warn` here so as not to distract from the main error that is happening.

lxd/storage/drivers/clients/alletra_wsapi1.go
```diff
@@ -376,3 +421,112 @@ func (p *AlletraClient) modifyVolumeSet(volumeSetName string, action int, volNam
...
+// DeleteHost deletes an existing host.
+func (p *AlletraClient) DeleteHost(hostName string) error {
+	url := api.NewURL().Path("api", "v1", "hosts", hostName)
+	err := p.requestAuthenticated(http.MethodDelete, url.URL, nil, nil)
+	if err != nil {
+		hpeErr, ok := err.(*hpeError)
+		if ok {
+			switch hpeErr.Code {
+			case apiErrorExportedVLUN:
+				p.logger.Debug("Host won't be deleted since it has exported volumes", logger.Ctx{"hostName": hostName})
```
tomponline, line 589: shouldnt this be returning an error to the caller rather than silently succeeding but not deleting the host?

Or is this a case of the "last one out turns off the lights"?

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: Only few of them. I can remove some unused fields in the second PR.

lxd/storage/drivers/driver_alletra_volumes.go
mihalicyn: Will be covered in a 2nd PR. Thanks!

mihalicyn:
> I cant find the comment/response now, but I think we should discuss further the metrics reporting issue.

Yeah, I replied somewhere, but it is quite hard to find it :)
> 
> Did you say that pool size and usage reporting won't be possible with the approach taken?

It is not impossible, but it will be quite hacky. We can obtain full volume list in a volumeset and then iterate over them and sum up the values. I'm not sure if this makes any sense and has any value. I feel like not, but you know better.

There is a storage statistics API in Alletra, but it is more low-level it's like about physical drives (CPGs and LDs).
https://support.hpe.com/hpesc/public/docDisplay?docId=sd00003740en_us&page=GUID-E098B4BE-A8ED-4F56-B058-E3C94CF798CF.html

tomponline:
> > I cant find the comment/response now, but I think we should discuss further the metrics reporting issue.
> 
> Yeah, I replied somewhere, but it is quite hard to find it :)
> 
> > Did you say that pool size and usage reporting won't be possible with the approach taken?
> 
> It is not impossible, but it will be quite hacky. We can obtain full volume list in a volumeset and then iterate over them and sum up the values. I'm not sure if this makes any sense and has any value. I feel like not, but you know better.
> 
> There is a storage statistics API in Alletra, but it is more low-level it's like about physical drives (CPGs and LDs). https://support.hpe.com/hpesc/public/docDisplay?docId=sd00003740en_us&page=GUID-E098B4BE-A8ED-4F56-B058-E3C94CF798CF.html

@mihalicyn thanks, please would you add this as an agenda item to discuss briefly in the team meeting so @skozina and @escabo are aware. Ta

lxd/storage/drivers/driver_alletra_volumes.go
mihalicyn: Fixed, thanks!

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: Fixed, thanks!

lxd/storage/drivers/clients/alletra_wsapi1.go
mihalicyn: Fixed, thanks!

PR #15947, opened 2025-07-04
Assorted ZFS tweaks

simondeziel:
This includes some cherry-picks from Incus.


lxd/storage/drivers/driver_zfs_volumes.go
tomponline, line 1: @simondeziel should we include this yet, it would seem this would require supporting changes elsewhere in the code base?

lxd/storage/drivers/driver_zfs_volumes.go
```diff
@@ -1936,7 +1936,7 @@ func (d *zfs) getVolumeDiskPathFromDataset(dataset string) (string, error) {
 	zvolUdevLink := filepath.Join("/dev/zvol", dataset)
 
 	// Shortcut for udev.
-	if shared.PathExists(zvolUdevLink) {
+	if shared.PathExists(zvolUdevLink) && shared.IsBlockdevPath(zvolUdevLink) {
```
tomponline: @simondeziel what do you think about using an os.Stat call here for `zvolUdevLink` so that the symlink is followed, and then using that to check for `fs.ErrNotExist` error and then if not passing the return file info to `shared.IsBlockdev()` to avoid doing an os.Lstat and and os.Stat?

simondeziel: I added https://github.com/canonical/lxd/pull/15947/commits/a932a8e0659e3b2afbdee88627c2eca311ba228d to that effect. Let me know if I understood your proposed solution wrong.

lxd/storage/drivers/driver_zfs_volumes.go
simondeziel: This PR was meant to not accumulate too many changes while working on switching to 24.04 runners. Those changes should stand on their own.

tomponline: yeah agreed, so i dont think we should land this just now.

lxd/storage/drivers/driver_zfs_volumes.go
tomponline: LGTM thanks!

lxd/storage/drivers/driver_zfs_volumes.go
simondeziel: Hmm, looks like we are not on the same page. I'd like to see those changes go in without waiting for what I believe is not related to the exacerbated race problem we have with ZFS on 24.04.

The race still exist and none of those commits actually aims at fixing it.

tomponline: yep and thats fine except for this commit - as LXD does not support online resize of block volumes yet

simondeziel: Ah, I see, I was missing the context that your comment was about a specific commit. I dropped it now. Thanks.

PR #15956, opened 2025-07-07
lint: Fix `var-naming` lint error on `util` package

markylaing:
The config to exclude `var-naming` on `lxd/util` doesn't work because the linter is not called `var-naming`. Rather, var-naming is a rule defined on the revive linter.

This PR reverts that commit and then disables package name checks on the `var-naming` rule for `revive`.


.golangci.yaml
```diff
@@ -37,6 +37,7 @@ linters:
             - []
             - []
             - - upperCaseConst: true
+                skipPackageNameChecks: true
```
tomponline: @markylaing i found this doesnt work on the latest release of the revive linter

markylaing: great :facepalm: 

Maybe just a text exclusion will be most foolproof then https://golangci-lint.run/usage/false-positives/#exclude-issue-by-text

markylaing: Done.

tomponline (approved):

kadinsayani:
Thanks for fixing this @markylaing!

PR #15957, opened 2025-07-07
Initial refactor of image API handlers

### LINKED ISSUE #15075: Disallow public image creation in non-default projects
It is not possible for unauthenticated clients to retrieve public images from any project other than the default project. The same is also true for images made available over the devlxd API. 

For now we should disallow this behaviour to avoid confusion. We might later re-enable it.

markylaing:
Access handling in the image API handlers is complex because:
1. Public images in the default project can be viewed and exported by untrusted callers.
2. Private images in any project can be viewed and exported by callers that do not have permission (either untrusted or authenticated and no permissions) if they have a valid image secret.
3. Image aliases in the default project can be viewed if they reference a public image.
4. Images in the default project can be exported via DevLXD if they are public or cached.

This PR attempts to simplify the logic for how these access control decisions are made by:
1. Refactoring the image export handler to remove DevLXD related logic, and that directly to the DevLXD handler.
2. Adding an access handler to perform access control decisions for image GET and image export.
3. Secrets can now only be obtained for private images in the default project. No official client for LXD supports getting a private image from non-default projects.
4. Preventing public images from being created in non-default projects.

Additionally, all image API handlers get images by their fingerprint prefix (returning an error if more than one fingerprint matches the prefix). However, we generally expect clients to retrieve a full fingerprint via an alias and then use it. Clients should not be querying e.g. `GET /1.0/images/a`. This PR restricts the fingerprint prefix path argument to be at least 12 lowercase hexadecimal characters.

Closes #15075

### DIFF
--- a/lxc/utils.go
+++ b/lxc/utils.go
@@ -393,6 +393,15 @@ func getImgInfo(d lxd.InstanceServer, conf *config.Config, imgRemote string, ins
 		if err != nil {
 			return nil, nil, err
 		}
+
+		// If the image server is a private LXD server, conf.GetImageServer will set the project of the image server to
+		// the value of `--project`. This means for `lxc init A:img B:inst --project foo` the client will try to get
+		// "img" from project "foo" and not the default project. Since there is no way to specify the project on the image
+		// remote we reset it to the default project.
+		privateServer, ok := imgRemoteServer.(*lxd.ProtocolLXD)
+		if ok {
+			imgRemoteServer = privateServer.UseProject(api.ProjectDefaultName)
+		}
 	}
 
 	// Optimisation for simplestreams
--- a/lxd/db/cluster/images.mapper.go
+++ b/lxd/db/cluster/images.mapper.go
@@ -41,6 +41,14 @@ SELECT images.id, projects.name AS project, images.fingerprint, images.type, ima
   ORDER BY projects.id, images.fingerprint
 `)
 
+var imageObjectsByProjectAndFingerprint = RegisterStmt(`
+SELECT images.id, projects.name AS project, images.fingerprint, images.type, images.filename, images.size, images.public, images.architecture, images.creation_date, images.expiry_date, images.upload_date, images.cached, images.last_use_date, images.auto_update
+  FROM images
+  JOIN projects ON images.project_id = projects.id
+  WHERE ( project = ? AND images.fingerprint = ? )
+  ORDER BY projects.id, images.fingerprint
+`)
+
 var imageObjectsByProjectAndCached = RegisterStmt(`
 SELECT images.id, projects.name AS project, images.fingerprint, images.type, images.filename, images.size, images.public, images.architecture, images.creation_date, images.expiry_date, images.upload_date, images.cached, images.last_use_date, images.auto_update
   FROM images
@@ -168,6 +176,30 @@ func GetImages(ctx context.Context, tx *sql.Tx, filters ...ImageFilter) ([]Image
 			}
 
 			query, err := StmtString(imageObjectsByProjectAndPublic)
+			if err != nil {
+				return nil, fmt.Errorf("Failed to get \"imageObjects\" prepared statement: %w", err)
+			}
+
+			parts := strings.SplitN(query, "ORDER BY", 2)
+			if i == 0 {
+				copy(queryParts[:], parts)
+				continue
+			}
+
+			_, where, _ := strings.Cut(parts[0], "WHERE")
+			queryParts[0] += "OR" + where
+		} else if filter.Project != nil && filter.Fingerprint != nil && filter.ID == nil && filter.Public == nil && filter.Cached == nil && filter.AutoUpdate == nil {
+			args = append(args, []any{filter.Project, filter.Fingerprint}...)
+			if len(filters) == 1 {
+				sqlStmt, err = Stmt(tx, imageObjectsByProjectAndFingerprint)
+				if err != nil {
+					return nil, fmt.Errorf("Failed to get \"imageObjectsByProjectAndFingerprint\" prepared statement: %w", err)
+				}
+
+				break
+			}
+
+			query, err := StmtString(imageObjectsByProjectAndFingerprint)
 			if err != nil {
 				return nil, fmt.Errorf("Failed to get \"imageObjects\" prepared statement: %w", err)
 			}
--- a/lxd/db/images.go
+++ b/lxd/db/images.go
@@ -447,10 +447,10 @@ SELECT images_aliases.name
 }
 
 // GetImageAlias returns the alias with the given name in the given project.
-func (c *ClusterTx) GetImageAlias(ctx context.Context, projectName string, imageName string, isTrustedClient bool) (int, api.ImageAliasesEntry, error) {
+func (c *ClusterTx) GetImageAlias(ctx context.Context, projectName string, imageName string, isTrustedClient bool) (int, bool, api.ImageAliasesEntry, error) {
 	id := -1
 	entry := api.ImageAliasesEntry{}
-	q := `SELECT images_aliases.id, images.fingerprint, images.type, images_aliases.description
+	q := `SELECT images_aliases.id, images.public, images.fingerprint, images.type, images_aliases.description
 			 FROM images_aliases
 			 INNER JOIN images
 			 ON images_aliases.image_id=images.id
@@ -463,7 +463,7 @@ func (c *ClusterTx) GetImageAlias(ctx context.Context, projectName string, image
 
 	enabled, err := cluster.ProjectHasImages(ctx, c.tx, projectName)
 	if err != nil {
-		return -1, api.ImageAliasesEntry{}, fmt.Errorf("Check if project has images: %w", err)
+		return -1, false, api.ImageAliasesEntry{}, fmt.Errorf("Check if project has images: %w", err)
 	}
 
 	if !enabled {
@@ -472,16 +472,17 @@ func (c *ClusterTx) GetImageAlias(ctx context.Context, projectName string, image
 
 	var fingerprint, description string
 	var imageType int
+	var public bool
 
 	arg1 := []any{projectName, imageName}
-	arg2 := []any{&id, &fingerprint, &imageType, &description}
+	arg2 := []any{&id, &public, &fingerprint, &imageType, &description}
 	err = c.tx.QueryRowContext(ctx, q, arg1...).Scan(arg2...)
 	if err != nil {
 		if err == sql.ErrNoRows {
-			return -1, api.ImageAliasesEntry{}, api.StatusErrorf(http.StatusNotFound, "Image alias not found")
+			return -1, false, api.ImageAliasesEntry{}, api.StatusErrorf(http.StatusNotFound, "Image alias not found")
 		}
 
-		return 0, entry, err
+		return 0, false, entry, err
 	}
 
 	entry.Name = imageName
@@ -489,7 +490,7 @@ func (c *ClusterTx) GetImageAlias(ctx context.Context, projectName string, image
 	entry.Description = description
 	entry.Type = instancetype.Type(imageType).String()
 
-	return id, entry, nil
+	return id, public, entry, nil
 }
 
 // RenameImageAlias renames the alias with the given ID.
--- a/lxd/devlxd.go
+++ b/lxd/devlxd.go
@@ -16,6 +16,8 @@ import (
 
 	"github.com/canonical/lxd/lxd/auth"
 	"github.com/canonical/lxd/lxd/cloudinit"
+	"github.com/canonical/lxd/lxd/db"
+	"github.com/canonical/lxd/lxd/db/cluster"
 	"github.com/canonical/lxd/lxd/events"
 	"github.com/canonical/lxd/lxd/instance"
 	"github.com/canonical/lxd/lxd/lifecycle"
@@ -278,7 +280,39 @@ func devLXDImageExportHandler(d *Daemon, r *http.Request) response.Response {
 		return response.Forbidden(err)
 	}
 
-	return imageExport(d, r)
+	fingerprint, err := url.PathUnescape(mux.Vars(r)["fingerprint"])
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	projectName := request.ProjectParam(r)
+	if projectName != api.ProjectDefaultName {
+		return response.NotFound(nil)
+	}
+
+	s := d.State()
+
+	var imgInfo *api.Image
+	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+		dbImage, err := cluster.GetImage(ctx, tx.Tx(), api.ProjectDefaultName, fingerprint)
+		if err != nil {
+			return err
+		}
+
+		imgInfo, err = dbImage.ToAPI(ctx, tx.Tx(), api.ProjectDefaultName)
+		return err
+	})
+	if err != nil && api.StatusErrorCheck(err, http.StatusNotFound) {
+		return response.NotFound(nil)
+	} else if err != nil {
+		return response.SmartError(err)
+	}
+
+	if !imgInfo.Public && !imgInfo.Cached {
+		return response.NotFound(nil)
+	}
+
+	return imageExportFiles(r.Context(), s, imgInfo)
 }
 
 var devLXDMetadataEndpoint = devLXDAPIEndpoint{
--- a/lxd/images.go
+++ b/lxd/images.go
@@ -4207,16 +4207,12 @@ func imageExport(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
-	// Verify the auth method in the request context to determine if the request comes from the /dev/lxd socket.
-	authMethod, _ := auth.GetAuthenticationMethodFromCtx(r.Context())
-	isDevLXDQuery := authMethod == auth.AuthenticationMethodDevLXD
-
 	secret := r.FormValue("secret")
 	trusted := auth.IsTrusted(r.Context())
 
 	// Unauthenticated remote clients that do not provide a secret may only view public images.
 	// For devlxd, we allow querying for private images. We'll subsequently perform additional access checks.
-	publicOnly := !trusted && secret == "" && !isDevLXDQuery
+	publicOnly := !trusted && secret == ""
 
 	// Get the image. We need to do this before the permission check because the URL in the permission check will not
 	// work with partial fingerprints.
@@ -4260,20 +4256,6 @@ func imageExport(d *Daemon, r *http.Request) response.Response {
 		}
 	}
 
-	if isDevLXDQuery {
-		// A devlxd query must contain the full fingerprint of the image (no partials).
-		if fingerprint != imgInfo.Fingerprint {
-			return response.NotFound(nil)
-		}
-
-		// A devlxd query must be for a public or cached image.
-		if !imgInfo.Public && !imgInfo.Cached {
-			return response.NotFound(nil)
-		}
-
-		userCanViewImage = true
-	}
-
 	if !userCanViewImage {
 		if imgInfo.Public {
 			// If the image is public any client can view it.
@@ -4297,12 +4279,15 @@ func imageExport(d *Daemon, r *http.Request) response.Response {
 		return response.NotFound(nil)
 	}
 
+	return imageExportFiles(r.Context(), s, imgInfo)
+}
+
+func imageExportFiles(ctx context.Context, s *state.State, imgInfo *api.Image) response.Response {
 	var address string
-
-	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+	err := s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
 		// Check if the image is only available on another node.
+		var err error
 		address, err = tx.LocateImage(ctx, imgInfo.Fingerprint)
-
 		return err
 	})
 	if err != nil {
@@ -4311,7 +4296,7 @@ func imageExport(d *Daemon, r *http.Request) response.Response {
 
 	if address != "" {
 		// Forward the request to the other node
-		client, err := cluster.Connect(r.Context(), address, s.Endpoints.NetworkCert(), s.ServerCert(), false)
+		client, err := cluster.Connect(ctx, address, s.Endpoints.NetworkCert(), s.ServerCert(), false)
 		if err != nil {
 			return response.SmartError(err)
 		}
@@ -4354,8 +4339,8 @@ func imageExport(d *Daemon, r *http.Request) response.Response {
 		files[1].Path = rootfsPath
 		files[1].Filename = filename
 
-		requestor := request.CreateRequestor(r.Context())
-		s.Events.SendLifecycle(projectName, lifecycle.ImageRetrieved.Event(imgInfo.Fingerprint, projectName, requestor, nil))
+		requestor := request.CreateRequestor(ctx)
+		s.Events.SendLifecycle(imgInfo.Project, lifecycle.ImageRetrieved.Event(imgInfo.Fingerprint, imgInfo.Project, requestor, nil))
 
 		return response.FileResponse(files, nil)
 	}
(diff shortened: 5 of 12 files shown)


lxd/images.go
```diff
@@ -3110,7 +3110,7 @@ func imageValidSecret(s *state.State, r *http.Request, projectName string, finge
 			continue
 		}
 
-		if subtle.ConstantTimeCompare([]byte(opSecretStr), []byte(secret)) == 1 {
```
markylaing: @tomponline tests are failing because of this commit. 

It turns out that while image secrets are intended to be single use. They can be used more than once if the operation is not deleted before subsequent calls are made (this happens 5s after the operation is cancelled). 

I thought this was a bug, but it turns out we actually rely on this behaviour when creating an instance. Lets say a user has view permission on a private image `foo` on cluster `A`. If they want to create an instance using it on cluster `B` they'll run
```
lxc init A:foo B:bar
```

At which point, the client will recognise that `foo` is a private image and get an image secret for it. Then add it to the instance source in the body of `POST /1.0/instances` to cluster B. Cluster B gets information about the image with a call to `GET /1.0/images/{fingerprint}?secret={secret}` and then tried to download it with a call to `GET /1.0/images/{fingerprint}/export?secret={secret}` immediately afterwards. So image secrets are not intended for one time use, but two-time use :facepalm: 

I'm not sure what to do about this. For now I'll likely need to drop this commit modify my tests to wait for 5 seconds until the secret is no longer valid

markylaing:
Converting to draft. Have added more tests to cover behaviour of public images when the caller is trusted but has no permissions and it's a mess :facepalm:

markylaing:
@tomponline I think this is ready for review now. It's not perfect but further work can rely on the tests to make sure it's not broken.

Creating public images in non-default projects is now disallowed. There is no way to use them unless using some unsupported client. We can relax this at a later point if it becomes a priority.

Also, creating image secrets is disallowed unless for a private image in the default project. Again, there is officially supported way of using the secret since the `GetPrivateImage` method on the client only queries the default project.

lxd/images.go
tomponline: Right understood. But the secret is still unusable if the operation is cancelled right?

tomponline: @markylaing where did the idea come from that they were single use btw?

lxd/devlxd.go
```diff
@@ -278,7 +280,39 @@ func devLXDImageExportHandler(d *Daemon, r *http.Request) response.Response {
...
 	}
 
-	return imageExport(d, r)
+	fingerprint, err := url.PathUnescape(mux.Vars(r)["fingerprint"])
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	projectName := request.ProjectParam(r)
+	if projectName != api.ProjectDefaultName {
```
tomponline, line 289: Could do with a comment i think.

lxd/images.go
```diff
@@ -4297,12 +4279,15 @@ func imageExport(d *Daemon, r *http.Request) response.Response {
 		return response.NotFound(nil)
 	}
 
-	var address string
+	return imageExportFiles(r.Context(), s, imgInfo)
+}
 
-	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+func imageExportFiles(ctx context.Context, s *state.State, imgInfo *api.Image) response.Response {
```
tomponline, line 4431: Missing doc block

lxd/images.go
markylaing: > Right understood. But the secret is still unusable if the operation is cancelled right?

The secret is usable after the operation is cancelled, until the operation is deleted (always a 5s delay).

> @markylaing where did the idea come from that they were single use btw?

From a comment in the `imageValidSecret` function: https://github.com/canonical/lxd/blob/962d5542f62df82ea4c5e2db70c8404c8f606ace/lxd/images.go#L3022

tomponline: > The secret is usable after the operation is cancelled, until the operation is deleted (always a 5s delay).
> 
> >

OK so I think that needs to change.

tomponline: > From a comment in the `imageValidSecret` function:

Oh right I see. So it seems to be that we should make that secret valid for the lifetime of the operation and only cancel said operation when we're done with it. So its single operation secret.

Will this break API compat?

lxd/devlxd.go
tomponline: Also it looks a bit odd to create a var called `projectName` and then not use anywhere else, instead can you just do:

```go
if request.ProjectParam(r) != api.ProjectDefaultName {
```

lxd/devlxd.go
```diff
@@ -278,7 +280,39 @@ func devLXDImageExportHandler(d *Daemon, r *http.Request) response.Response {
...
+		imgInfo, err = dbImage.ToAPI(ctx, tx.Tx(), api.ProjectDefaultName)
+		return err
+	})
+	if err != nil && api.StatusErrorCheck(err, http.StatusNotFound) {
+		return response.NotFound(nil)
+	} else if err != nil {
+		return response.SmartError(err)
+	}
+
+	if !imgInfo.Public && !imgInfo.Cached {
```
tomponline, line 311: comment here too please explaining why

lxd/devlxd.go
```diff
@@ -278,7 +280,39 @@ func devLXDImageExportHandler(d *Daemon, r *http.Request) response.Response {
...
+		dbImage, err := cluster.GetImage(ctx, tx.Tx(), api.ProjectDefaultName, fingerprint)
+		if err != nil {
+			return err
+		}
+
+		imgInfo, err = dbImage.ToAPI(ctx, tx.Tx(), api.ProjectDefaultName)
+		return err
+	})
+	if err != nil && api.StatusErrorCheck(err, http.StatusNotFound) {
+		return response.NotFound(nil)
```
tomponline, line 306: comment here explaining whats going on as is not clear.

lxd/images.go
```diff
@@ -4297,12 +4279,15 @@ func imageExport(d *Daemon, r *http.Request) response.Response {
...
+	var address string
+	err := s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
 		// Check if the image is only available on another node.
+		var err error
```
tomponline, line 4435: no need for this

lxd/images.go
```diff
@@ -4297,12 +4279,15 @@ func imageExport(d *Daemon, r *http.Request) response.Response {
...
 	}
 
-	var address string
+	return imageExportFiles(r.Context(), s, imgInfo)
+}
 
-	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+func imageExportFiles(ctx context.Context, s *state.State, imgInfo *api.Image) response.Response {
+	var address string
+	err := s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
```
tomponline, line 4433: should this use ctx from function arg?

lxd/images.go
```diff
@@ -4354,8 +4339,8 @@ func imageExport(d *Daemon, r *http.Request) response.Response {
 		files[1].Path = rootfsPath
 		files[1].Filename = filename
 
-		requestor := request.CreateRequestor(r.Context())
-		s.Events.SendLifecycle(projectName, lifecycle.ImageRetrieved.Event(imgInfo.Fingerprint, projectName, requestor, nil))
+		requestor := request.CreateRequestor(ctx)
+		s.Events.SendLifecycle(imgInfo.Project, lifecycle.ImageRetrieved.Event(imgInfo.Fingerprint, imgInfo.Project, requestor, nil))
```
tomponline, line 4489: this used to use `projectName := request.ProjectParam(r)` but now is using `imgInfo.Project`, is there any difference wrt to effective image project?

lxd/images.go
```diff
@@ -4365,8 +4350,8 @@ func imageExport(d *Daemon, r *http.Request) response.Response {
 	files[0].Path = imagePath
 	files[0].Filename = filename
 
-	requestor := request.CreateRequestor(r.Context())
-	s.Events.SendLifecycle(projectName, lifecycle.ImageRetrieved.Event(imgInfo.Fingerprint, projectName, requestor, nil))
+	requestor := request.CreateRequestor(ctx)
+	s.Events.SendLifecycle(imgInfo.Project, lifecycle.ImageRetrieved.Event(imgInfo.Fingerprint, imgInfo.Project, requestor, nil))
```
tomponline, line 4500: same question as above

lxd/images.go
```diff
@@ -159,6 +160,7 @@ func addImageDetailsToRequestContext(s *state.State, r *http.Request) (*imageDet
 	}
 
 	details := imageDetails{
+		requestProjectName:     requestProjectName,
```
tomponline, line 185: Surely this is already in the context and isn't need to be added to every request details entity struct?

tomponline: this feels like a step too far to me in terms of bundling everything about a request into a entity specific struct, maybe im missing something, but it seems like an indirection too far (blurring the lines between retrieving an entity's info, validating access to it and then storing in the context for later use, and also duplicating request information).

lxd/images.go
```diff
@@ -244,6 +244,69 @@ func imageAliasAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *ht
...
+
+	// Some checks should be performed before hitting the database, because here the caller can be untrusted.
+	// In which case the caller may only query images in the default project.
+	trusted := auth.IsTrusted(r.Context())
+	projectName := request.ProjectParam(r)
+	secret := r.FormValue("secret")
+
+	// If not trusted and project is not default. Exporting or viewing images is only allowed with a secret.
+	if !trusted && projectName != api.ProjectDefaultName && secret == "" {
+		return response.NotFound(nil)
```
tomponline, line 357: do we want a comment here about why its generic?

lxd/images.go
tomponline, line 1: One question I have is: Why does `addImageDetailsToRequestContext` need to use `GetImageByFingerprintPrefix` rather than an exact match function?

lxd/images.go
```diff
@@ -73,7 +73,7 @@ var imageCmd = APIEndpoint{
 	MetricsType: entity.TypeImage,
 
 	Delete: APIEndpointAction{Handler: imageDelete, AccessHandler: imageAccessHandler(auth.EntitlementCanDelete)},
-	Get:    APIEndpointAction{Handler: imageGet, AllowUntrusted: true},
+	Get:    APIEndpointAction{Handler: imageGet, AccessHandler: imageGetOrExportAccessHandler, AllowUntrusted: true},
```
tomponline, line 77: do we still need `AllowUntrusted: true` here?

lxd/images.go
```diff
@@ -3201,9 +3201,7 @@ func imageValidSecret(s *state.State, r *http.Request, projectName string, finge
 //	    $ref: "#/responses/InternalServerError"
 func imageGet(d *Daemon, r *http.Request) response.Response {
 	s := d.State()
-
-	projectName := request.ProjectParam(r)
```
tomponline, line 3124: i think its fine to keep this and not include in `details`.

lxd/images.go
tomponline, line 1: Much cleaner, nice!

lxd/db/images.go
```diff
@@ -447,10 +447,10 @@ SELECT images_aliases.name
 }
 
 // GetImageAlias returns the alias with the given name in the given project.
-func (c *ClusterTx) GetImageAlias(ctx context.Context, projectName string, imageName string, isTrustedClient bool) (int, api.ImageAliasesEntry, error) {
+func (c *ClusterTx) GetImageAlias(ctx context.Context, projectName string, imageName string, isTrustedClient bool) (int, bool, api.ImageAliasesEntry, error) {
```
tomponline, line 450: doc block needs to be updated, and ideally switching to named return vars would make this function self-document, e.g. `(aliasID int, public bool, alias api.ImageAliasesEntry, err error)`

lxd/images.go
```diff
@@ -212,6 +213,98 @@ func imageAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *http.Re
...
+}
+
+// imageAliasGetAccessHandler sets imageAliasGetDetails in the request context, which indicates to the handler if it
+// should restrict the database query to return only aliases of public images. If the image is private, the handler must
+// still perform an access check.
+func imageAliasGetAccessHandler(d *Daemon, r *http.Request) response.Response {
+	requestProjectName := request.ProjectParam(r)
+	details := imageAliasGetDetails{}
+	defer func() {
+		request.SetContextValue(r, ctxImageAliasGetDetails, details)
```
tomponline, line 229: do we really want to set the details into the context if an error occurs in this function?

lxd/images.go
```diff
@@ -212,6 +213,98 @@ func imageAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *http.Re
...
+	}()
+
+	imageAliasName, err := url.PathUnescape(mux.Vars(r)["name"])
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	// If the caller is not trusted, only allow requests to the default project and restrict to aliases of public images.
+	trusted := auth.IsTrusted(r.Context())
+	if !trusted {
```
tomponline, line 239: > restrict to aliases of public images.

This part of the check doesnt appear to be happening here

lxd/images.go
```diff
@@ -212,6 +213,98 @@ func imageAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *http.Re
...
+	// regardless of the image target.
+	s := d.State()
+	if requestProjectName != api.ProjectDefaultName {
+		info := request.SetupContextInfo(r)
+		err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+			info.EffectiveProjectName, err = projectutils.ImageProject(ctx, tx.Tx(), requestProjectName)
+			return err
+		})
+		if err != nil && api.StatusErrorCheck(err, http.StatusNotFound) {
+			return response.NotFound(nil)
```
tomponline, line 257: need a comment here explaining why its like this

lxd/images.go
```diff
@@ -212,6 +213,98 @@ func imageAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *http.Re
 	}
 }
 
+const ctxImageAliasGetDetails request.CtxKey = "image-alias-details"
+
+type imageAliasGetDetails struct {
+	getPrivateAliases bool
```
tomponline, line 219: missing comment

lxd/images.go
```diff
@@ -212,6 +213,98 @@ func imageAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *http.Re
...
+		details.getPrivateAliases = true
+		return response.EmptySyncResponse
+	}
+
+	id, err := auth.GetIdentityFromCtx(r.Context(), s.IdentityCache)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	// Fine-grained identity types need to load the alias and perform an access check if it is private.
```
tomponline, line 287: It took a few reads of this figure out that you mean this will happen elsewhere and is not what the code is doing here. Think this could be clearer.

lxd/images.go
```diff
@@ -212,6 +213,98 @@ func imageAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *http.Re
...
+	if identity.IsFineGrainedIdentityType(id.IdentityType) {
+		details.getPrivateAliases = true
```
tomponline, line 289: I was thinking earlier that perhaps details.canViewPrivateImageAliases be clearer name, but now this variable is somewhat tricky to understand, as in this case its true, but could turn out later the requestor cannot see private image aliases right?

lxd/images.go
```diff
@@ -212,6 +213,98 @@ func imageAccessHandler(entitlement auth.Entitlement) func(d *Daemon, r *http.Re
...
+const ctxImageAliasGetDetails request.CtxKey = "image-alias-details"
+
+type imageAliasGetDetails struct {
+	getPrivateAliases bool
+}
+
+// imageAliasGetAccessHandler sets imageAliasGetDetails in the request context, which indicates to the handler if it
+// should restrict the database query to return only aliases of public images. If the image is private, the handler must
+// still perform an access check.
+func imageAliasGetAccessHandler(d *Daemon, r *http.Request) response.Response {
```
tomponline, line 225: I found this function a bit of a slog. I realise this image access control stuff is pretty convoluted, but I'm wondering if can be presented in an easier more concise way. 

The part I found most tricky to follow this function was the mixture of happy-path and sad-path style, e.g.

"if this then allow, if this then dont, else if this other thing do, but if this then dont, for everything else allow"

Would it be clearer if we do it in phases, i.e switch by broad client type, then for each do their specific exception checks (perhaps in smaller functions?)?

tomponline (commented):
I've not made it all the way through this yet so thought id give you what ive got so farv :)

Thanks!

lxd/images.go
markylaing: Yeah I think that will work better. We did this for some identity APIs and that was a lot cleaner.

lxd/patches.go
```diff
@@ -1568,4 +1569,33 @@ func patchMoveBackupsImagesStorage(name string, d *Daemon) error {
 	return nil
 }
 
+// patchPublicImagesInNonDefaultProjects sets all images that are not in the default project to private. This is because
+// images that are not in the default project can never be exported or viewed by untrusted callers.
+func patchPublicImagesInNonDefaultProjects(name string, d *Daemon) error {
+	err := d.State().DB.Cluster.Transaction(d.shutdownCtx, func(ctx context.Context, tx *db.ClusterTx) error {
+		q := `UPDATE images SET public = 0 WHERE project_id = (SELECT id FROM projects WHERE name = ?)`
```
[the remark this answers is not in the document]
markylaing: Wow 😯 thank you copilot!

lxc/utils.go
```diff
@@ -393,6 +393,15 @@ func getImgInfo(d lxd.InstanceServer, conf *config.Config, imgRemote string, ins
 		if err != nil {
 			return nil, nil, err
 		}
+
+		// If the image server is a private LXD server, conf.GetImageServer will set the project of the image server to
+		// the value of `--project`. This means for `lxc init A:img B:inst --project foo` the client will try to get
+		// "img" from project "foo" and not the default project. Since there is no way to specify the project on the image
+		// remote we reset it to the default project.
```
kadinsayani, line 400: Instead of resetting the project, can we disallow the `--project` flag for private image servers entirely by checking at the call site in `lxc/init.go`? That way we can return a useful error message.

lxd/images.go
```diff
@@ -113,28 +114,50 @@ var imageAliasCmd = APIEndpoint{
...
 }
 
+// validateImageFingerprintPrefix validates that the given string is at least 12 characters long contains only lowercase
+// hex characters.
+func validateImageFingerprintPrefix(prefix string) error {
+	if len(prefix) < 12 {
+		return api.NewStatusError(http.StatusBadRequest, "Image fingerprint prefix must contain 12 characters or more")
+	}
+
+	for _, b := range []byte(prefix) {
```
kadinsayani, line 130:
```suggestion
	for _, b := range prefix {
```

lxd/images.go
```diff
@@ -155,41 +178,131 @@ func addImageDetailsToRequestContext(s *state.State, r *http.Request) error {
...
+			return err
+		})
+		if err != nil && api.StatusErrorCheck(err, http.StatusNotFound) {
+			return response.NotFound(nil)
+		} else if err != nil {
+			return response.SmartError(err)
+		}
+
+		err := s.Authorizer.CheckPermission(r.Context(), entity.ImageAliasURL(requestProjectName, imageAliasName), auth.EntitlementCanView)
+		if err != nil && auth.IsDeniedError(err) {
```
kadinsayani, line 263: It would help to have a comment here explaining why the error is only returned when it is a `http.StatusNotFound` or `http.StatusForbidden` status error. It appears that any other error is allowed here since we don't have a check for when `err == nil` before setting `details.getPrivateAliases` to `true`.

test/suites/remote.sh
```diff
@@ -237,7 +237,7 @@ test_remote_usage() {
   lxc_remote image copy --quiet "localhost:${sum}" lxd2:
   lxc_remote image delete "lxd2:${sum}"
 
-  lxc_remote image copy --quiet "localhost:$(echo "${sum}" | cut -c 1-2)" lxd2:
+  lxc_remote image copy --quiet "localhost:$(echo "${sum}" | cut -c 1-12)" lxd2:
```
simondeziel, line 240: With `bash` we can directly access a substring:
```suggestion
  lxc_remote image copy --quiet "localhost:$(echo "${sum:0:12}" lxd2:
```

lxd/images.go
```diff
@@ -113,28 +114,50 @@ var imageAliasCmd = APIEndpoint{
...
+	Get:    APIEndpointAction{Handler: imageAliasGet, AccessHandler: imageAliasGetAccessHandler, AllowUntrusted: true},
 	Patch:  APIEndpointAction{Handler: imageAliasPatch, AccessHandler: imageAliasAccessHandler(auth.EntitlementCanEdit)},
 	Post:   APIEndpointAction{Handler: imageAliasPost, AccessHandler: imageAliasAccessHandler(auth.EntitlementCanEdit)},
 	Put:    APIEndpointAction{Handler: imageAliasPut, AccessHandler: imageAliasAccessHandler(auth.EntitlementCanEdit)},
 }
 
+// validateImageFingerprintPrefix validates that the given string is at least 12 characters long contains only lowercase
+// hex characters.
+func validateImageFingerprintPrefix(prefix string) error {
+	if len(prefix) < 12 {
```
simondeziel, line 126: We should also reject anything longer than 64.

lxd/db/cluster/images.mapper.go
```diff
@@ -41,6 +41,14 @@ SELECT images.id, projects.name AS project, images.fingerprint, images.type, ima
   ORDER BY projects.id, images.fingerprint
 `)
 
+var imageObjectsByProjectAndFingerprint = RegisterStmt(`
+SELECT images.id, projects.name AS project, images.fingerprint, images.type, images.filename, images.size, images.public, images.architecture, images.creation_date, images.expiry_date, images.upload_date, images.cached, images.last_use_date, images.auto_update
+  FROM images
+  JOIN projects ON images.project_id = projects.id
+  WHERE ( project = ? AND images.fingerprint = ? )
+  ORDER BY projects.id, images.fingerprint
```
simondeziel, line 49: Should we use a `LIMIT` of 1 to abort the table scan once we find one? Or maybe there is already a `UNIQUE` constraint on the `images.fingerprint` causing this implicitly. I don't know DB engines :/

lxd/devlxd.go
```diff
@@ -278,7 +280,39 @@ func devLXDImageExportHandler(d *Daemon, r *http.Request) response.Response {
...
+	} else if err != nil {
```
simondeziel, line 307: Both branches have `if err != nil` so would it be worth it to nest 2 ifs?

lxd/db/cluster/images.mapper.go
markylaing: IMHO unique constraints should enforce these kinds of things. Without the correct unique constraints, having a limit might be masking something going wrong internally. Unfortunately the images table has `UNIQUE (project_id, fingerprint)` - which I think is a bit weird given the behaviour of images in LXD. Not something that is likely to change any time soon though unfortunately :frowning_face:

lxd/images.go
markylaing: Fixed in https://github.com/canonical/lxd/pull/16002 :)

test/suites/remote.sh
markylaing: Fixed in https://github.com/canonical/lxd/pull/16002 :)

lxc/utils.go
markylaing: This is now addressed separately in https://github.com/canonical/lxd/pull/16005

lxd/images.go
markylaing: > Oh right I see. So it seems to be that we should make that secret valid for the lifetime of the operation and only cancel said operation when we're done with it. So its single operation secret.

The problem is the "when we're done with it" part. I think we have two options:
1. Make the secret valid for one `GET /1.0/images/{fingerprint}` call and one `GET /1.0/images/{fingerprint}/export` call. We would need to add more information to the operation metadata to do this, which would require that the request be forwarded to the machine containing the operation.
2. Make the secret valid for multiple `GET /1.0/images/{fingerprint}` calls but only one `GET /1.0/images/{fingerprint}/export` call. This way we can validate but not cancel the secret when getting the image info, and then validate and cancel on export.

I also think these secrets should have an expiry. They currently don't, and there is no task to clean them up, so they'll hang around forever.

> Will this break API compat?

If we make the token truly single use without any other changes then yes, this would stop private images from being copied between remotes (and also prevent instances from being created from private images).

tomponline: > I also think these secrets should have an expiry. They currently don't, and there is no task to clean them up, so they'll hang around forever.
> 
> >

They are part of the operation metadata so they should go once the operation completes.
We shouldnt be leaking operations though, so perhaps the more accurate issue here is operation lifetime?

markylaing: Yes, my meaning is that if the secret is never used, the operation is never deleted. I would expect an expiry like for CertificateAdd or ClusterJoin tokens.

PR #15975, opened 2025-07-07
Daemon: Only enforce content-type checks for browser clients

tomponline:
And change browser client detection to look for user-agents that start with `Mozilla`.


skozina (dismissed):

simondeziel (approved):
Thanks!

PR #15985, opened 2025-07-08
Use generic status text in LXD responses

markylaing:
When using e.g. `response.NotFound(nil)`, the error message is different from `response.SmartError(api.NewGenericStatusError(http.StatusNotFound))` (`not found` vs `Not Found`). 

This might seem inconsequential, but generic status errors are used to mask permission errors when view access is denied. If an access check requires some pre-processing (such as expanding an image fingerprint) then differences between these error messages can expose information.

This commit makes these `response.NotFound`, `response.Forbidden` etc. more consistent by using the `http.StatusText` for the error code rather than custom error messages.


lxd/response/response.go
```diff
@@ -290,75 +290,69 @@ type errorResponse struct {
 	msg  string // Message to return in the Error field of the response body.
 }
 
+// errorMessage returns the contents of the error if non-nil. Otherwise, the default [http.StatusText] is returned for
+// the code.
+func errorMessage(code int, err error) string {
```
MusicDin: Could we just do that when rendering the response? For example, if error message is not provided, to default to status code message?

markylaing: I'm open to that, though I'm not sure it will be any cleaner since we'll still need to check if the error is non-nil... I'll see if it works better :)

markylaing: Hmm no I don't like it as much. To make it concise you still end up with e.g.:
```go
// errorMessage returns the contents of the error if non-nil.
func errorMessage(err error) string {
	if err != nil {
		return err.Error()
	}

	return ""
}

// Unavailable return an unavailable response (503) with the given error.
func Unavailable(err error) Response {
	code := http.StatusServiceUnavailable
	return &errorResponse{code: code, msg: errorMessage(err)}
}

// Render renders a response that indicates an error on the request handling.
func (r *errorResponse) Render(w http.ResponseWriter, req *http.Request) error {
	if r.msg == "" {
		r.msg = http.StatusText(r.code)
	}

        /* etc. */
}
```

MusicDin: Could `errorResponse` store `err` instead of `msg`?

tomponline (commented):
Does this need to be added to our `api` package, thus expanding the size of the API that we provide for this package?

This seems like an internal concern.

markylaing:
> Does this need to be added to our `api` package, thus expanding the size of the API that we provide for this package?

I'm not sure what you mean. The response package imports a bunch of packages under `lxd/`, so I don't think it's suitable for moving under `shared/api` at this stage.

tomponline:
> > Does this need to be added to our `api` package, thus expanding the size of the API that we provide for this package?
> 
> I'm not sure what you mean. The response package imports a bunch of packages under `lxd/`, so I don't think it's suitable for moving under `shared/api` at this stage.

I mean i dont think it should be in the `api` package. That is a shared package, whereas this is a server-side thing.

markylaing:
> > > Does this need to be added to our `api` package, thus expanding the size of the API that we provide for this package?
> > 
> > 
> > I'm not sure what you mean. The response package imports a bunch of packages under `lxd/`, so I don't think it's suitable for moving under `shared/api` at this stage.
> 
> I mean i dont think it should be in the `api` package. That is a shared package, whereas this is a server-side thing.

I agree - This PR doesn't add anything to the `api` package.

lxd/response/response.go
markylaing: That works well actually. Will change.

tomponline:
> > > > Does this need to be added to our `api` package, thus expanding the size of the API that we provide for this package?
> > > 
> > > 
> > > I'm not sure what you mean. The response package imports a bunch of packages under `lxd/`, so I don't think it's suitable for moving under `shared/api` at this stage.
> > 
> > 
> > I mean i dont think it should be in the `api` package. That is a shared package, whereas this is a server-side thing.
> 
> I agree - This PR doesn't add anything to the `api` package.

Oh right i thought `NewGenericStatusError` was new for this PR. All good then

markylaing:
I've just updated `errorResponse` to take an `err error` field instead of a `msg string` field as per @MusicDin's suggestion.

MusicDin (approved):
Thanks! I really appreciate this change :)

PR #15988, opened 2025-07-09
lxd/db/cluster/projects: Optimize GetProjectConfig()

skozina:
@tomponline this is cherry-picked from #15827.
Now that the project-specific settings won't actually be project config options, this commit won't be needed, but could be useful on its own.

### DIFF
--- a/lxd/db/cluster/projects.go
+++ b/lxd/db/cluster/projects.go
@@ -26,7 +26,7 @@ import (
 //go:generate mapper stmt -e project update struct=Project
 //go:generate mapper stmt -e project delete-by-Name
 //
-//go:generate mapper method -i -e project GetMany references=Config
+//go:generate mapper method -i -e project GetMany
 //go:generate mapper method -i -e project GetOne struct=Project
 //go:generate mapper method -i -e project Exists struct=Project
 //go:generate mapper method -i -e project Create references=Config
@@ -88,7 +88,7 @@ func (p *Project) ToAPI(ctx context.Context, tx *sql.Tx) (*api.Project, error) {
 	}
 
 	var err error
-	apiProject.Config, err = GetProjectConfig(ctx, tx, p.ID)
+	apiProject.Config, err = GetProjectConfig(ctx, tx, p.Name)
 	if err != nil {
 		return nil, fmt.Errorf("Failed loading project config: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


lxd/db/cluster/projects.go
```diff
@@ -117,6 +117,41 @@ SELECT projects_config.value
 	return shared.IsTrue(values[0]), nil
 }
 
+// GetProjectConfig is a helper to return a config of a project.
+func GetProjectConfig(ctx context.Context, tx *sql.Tx, project string) (map[string]string, error) {
```
tomponline: just a nit, not blocker, but we tend to try to be more explicit about what `project` arg is here, so suggest `projectName`.

skozina: Sounds good.
I am actually wondering if it's better to have a project name as an argument (which will require the JOIN in this function, and will penalize ToAPI() calls), or if it's better to leave ID as an argument?
Still this commit would make sense to get rid of the GetConfig() call which is not efficient.

lxd/db/cluster/projects.go
```diff
@@ -117,6 +117,41 @@ SELECT projects_config.value
...
+	stmt := `
+	SELECT projects_config.key, projects_config.value
+	  FROM projects_config
+	  JOIN projects ON projects.id=projects_config.project_id
+	 WHERE projects.name=?
+	`
+	rows, err := tx.QueryContext(ctx, stmt, project)
```
tomponline: Just an FYI, we can we use `Scan()` 

https://github.com/canonical/lxd/blob/main/lxd/db/query/objects.go#L31-L33

as a small wrapper around `tx.QueryContext` to avoid having to remember to defer rows.Close().

lxd/db/cluster/projects.go
tomponline: This would mean `ProjectHasImages` still has to do a separate query to convert name to ID as it does today.

I suspect the join cost (on tables this size) is cheap compared to the overhead of doing an additional query from a non-leader node.

lxd/db/cluster/projects.go
skozina: Ah, sure, we can use it. Thanks, updated.

lxd/db/cluster/projects.go
skozina: Ack, so I've only updated the name of the argument to `projectName`.

tomponline (approved):
Thanks!

PR #15991, opened 2025-07-09
Events filtering

### LINKED ISSUE #13962: Effective project handling in the events API for fine grained users
Consider the following:
```
# Via unix or as administrator
$ lxc project create foo # `features.networks` & `features.networks.zones` disabled on creation be default.
$ lxc config trust add <cert> --restricted --projects foo

# Restricted cert
$ lxc monitor --all-projects --type lifecycle
$ lxc network create foo-network --project foo
```

The restricted certificate will not see the lifecycle event relating to the creation of their network, because the network was created in the default project.

We should discuss whether it is possible to address this. Some initial ideas would be:
1. Duplicate events in the default project and emit one for each project where that feature is disabled.
2. Add some "source project" property to events.
3. Don't handle it for existing TLS users, expect that an administrator will add `can_view_events` for the default project where a group is confined to a project with any features disabled.
4. Don't handle it for existing TLS users, add logic to include a source entity in the event, fine-grained users can view the event if they have `can_view` on the entity.

Note that 4. would be generally useful. We currently require that the `can_exec` entitlement on an instance be paired with `can_view_events` on the parent project because the client uses the events websocket instead of operations (to avoid creation of multiple connections). We've discussed how to fix this before in https://github.com/canonical/lxd/pull/12885, as using operations can be problematic (connections dropping when performing a long pool).

See https://github.com/canonical/lxd/issues/13886#issuecomment-2304208449_

markylaing:
This PR changes event filtering on `/1.0/events` such that:
- Callers with `can_view_privileged_events` see all events.
- Callers with `can_view_events` on a project (and restricted TLS clients with access to the project) can see all lifecycle events for that project, plus operations that they initiated.
- Remaining authenticated callers will only see lifecycle events and operations that they initiated.

What this means:
- Administrators will continue to see all events when running `lxc monitor`.
- The `can_view_events` entitlement grants less privilege, since it only concerns lifecycle events.
- The `can_view_events` entitlement is no longer required to start an exec or console session via the CLI.

Note that lifecycle events are still filtered on a per-project basis. While implementing this, I was able to filter lifecycle events at a fine-grained level by inspecting the event `source` property. However, the outputted lifecycle events contain a requestor field with information about other users in the system, so I think this should remain gated by the `can_view_events` entitlement.

Also note that the `filter` function that is now accepted by `(*events.Server).AddListener` **MUST NOT** make any calls to the `logger` package. Otherwise the event server goes into deadlock. I did investigate this but felt my solution with buffered channels was overkill. We can discuss in review.

This partially addresses #13962. However, I don't think it's ever going to be possible to account for effective projects in the events API, so I think we should say that this closes #13962 and another issue can be opened if this becomes a problem in the future.


doc/rest-api.yaml
```diff
@@ -747,6 +747,23 @@ definitions:
                 x-go-name: Type
         type: object
         x-go-package: github.com/canonical/lxd/shared/api
+    EventLifecycleRequestor:
+        description: 'API extension: event_lifecycle_requestor.'
```
markylaing: I've just noticed this description. We should probably have a separate type for this for operations, in case any information differs between operations and lifecycle events in the future.

nmezhenskyi: I think it could be useful in theory, but I'm not sure how likely it is for operations and lifecycle requestors to have different fields.

lxd/events/events.go
```diff
@@ -67,13 +65,13 @@ func (s *Server) SetLocalLocation(location string) {
 }
 
 // AddListener creates and returns a new event listener.
-func (s *Server) AddListener(projectName string, allProjects bool, projectPermissionFunc auth.PermissionChecker, connection EventListenerConnection, messageTypes []string, excludeSources []EventSource, recvFunc EventHandler, excludeLocations []string) (*Listener, error) {
+func (s *Server) AddListener(projectName string, allProjects bool, filter func(api.Event) bool, connection EventListenerConnection, messageTypes []string, excludeSources []EventSource, recvFunc EventHandler, excludeLocations []string) (*Listener, error) {
```
nmezhenskyi: Maybe it is worth mentioning in the `AddListener` function's doc comment that calls to the `logger` package must not be made inside of the `filter()` func?   

For example:  
```
// AddListener creates and returns a new event listener.
//
// NOTE: The filter function must not make any calls to the logger package. Otherwise the event server goes into deadlock.
```

markylaing: I've added a comment :)

doc/rest-api.yaml
markylaing: I've left this as is for now as it became a pain refactoring. I think it's fine because this will always represent the initial operation requestor, and any auditing data relevant here will also be relevant for lifecycle events.

markylaing:
@tomponline note that this PR also will remove OVN events from the event stream for non-admin users. I'm not sure of the consequences of that, but it seems to me these should always have been considered a privileged event (like logging events).

Marking as draft while this is discussed and while I figure out how to test this.

markylaing:
This is now blocked by https://github.com/canonical/lxd/issues/16072. I was testing that a fine-grained identity can view lifecycle events that they initiate, but there is no requestor field for instance-created events.

test/suites/auth.sh
```diff
@@ -324,6 +326,65 @@ fine_grained: true"
...
+  lxc auth group permission add test-group image "$(lxc query /1.0/images/aliases/testimage | jq -r '.target')" can_view project=default
+  lxc remote switch tls
+  lxc monitor --all-projects --format json > "${monfile}" &
+  monitor_pid=$!
+  sleep 0.1
+  lxc remote switch local
+
+  lxc init testimage tls:c1
+  kill -9 "${monitor_pid}" || true
+  jq -s -e 'any(.type == "lifecycle" and .metadata.action == "instance-created")' "${monfile}"
```
markylaing: This is the line that fails because of https://github.com/canonical/lxd/issues/16072

test/suites/auth.sh
```diff
@@ -324,6 +326,65 @@ fine_grained: true"
...
+
+  # Monitor as fine-grained identity that creates the instance with minimal permissions.
+  lxc auth group permission add test-group project default can_create_instances
+  lxc auth group permission add test-group image_alias testimage can_view project=default
+  lxc auth group permission add test-group image "$(lxc query /1.0/images/aliases/testimage | jq -r '.target')" can_view project=default
+  lxc remote switch tls
+  lxc monitor --all-projects --format json > "${monfile}" &
+  monitor_pid=$!
+  sleep 0.1
+  lxc remote switch local
```
markylaing, line 396: @simondeziel I found some things a bit weird when trying to get consistent results with `lxc monitor` in testing.

1. It doesn't seem to work at all when using the `tls` or `oidc` remotes that are set up in this suite.
2. You need to sleep a little, otherwise the `remote switch` command can actually run before the monitor starts.

I tried using `${_LXC}` directly, rather than the shell wrapper functions but nothing seemed to help. This was the only way I could get it to work. Any advice appreciated!

simondeziel: Could you get away without doing the remote switching? `lxc monitor tls: --all-projects --format json > "${monfile}" &` ?

markylaing: Nope, doesn't seem to work and I couldn't figure out why 😕

tomponline:
@markylaing @nmezhenskyi what is the current status of this as @edlerd is needing this for a UI issue.

cc @skozina

markylaing:
> @markylaing @nmezhenskyi what is the current status of this as @edlerd is needing this for a UI issue.
> 
> cc @skozina

Aside from requiring a rebase we need to:
- Create a new api type `OperationRequestor`, which will be the same as an event lifecycle requestor initially, but it feels odd to have a `Requestor` field in `api.Operation` that is an `api.EventLifecycleRequestor`. I don't know of any reason why these would deviate from one another, but to be thorough I think it should be a different type.
- Allow logging in the event filtering function. Just adding a comment to say that logging anything will cause the event server to deadlock seems risky and half baked.
- Finish off testing, probably with some events omitted out for now, because the requestor is missing on some events.

markylaing:
> > @markylaing @nmezhenskyi what is the current status of this as @edlerd is needing this for a UI issue.
> > cc @skozina
> 
> Aside from requiring a rebase we need to:
> 
> * Create a new api type `OperationRequestor`, which will be the same as an event lifecycle requestor initially, but it feels odd to have a `Requestor` field in `api.Operation` that is an `api.EventLifecycleRequestor`. I don't know of any reason why these would deviate from one another, but to be thorough I think it should be a different type.
> * Allow logging in the event filtering function. Just adding a comment to say that logging anything will cause the event server to deadlock seems risky and half baked.
> * Finish off testing, probably with some events omitted out for now, because the requestor is missing on some events.

I'll have a crack at it today since this has been in my backlog for a long while now

markylaing:
@tomponline this is ready for review now:

- [x] ~Create a new api type OperationRequestor, which will be the same as an event lifecycle requestor initially, but it feels odd to have a Requestor field in api.Operation that is an api.EventLifecycleRequestor. I don't know of any reason why these would deviate from one another, but to be thorough I think it should be a different type.~ To do this I've changed the `requestor` field of `operations.Operation` to be a `request.Requestor`, which can be converted to an `api.OperationRequestor` or `api.EventLifecycleRequestor` as necessary.
- [x] ~Allow logging in the event filtering function. Just adding a comment to say that logging anything will cause the event server to deadlock seems risky and half baked.~ To do this I've refactored the `shared/logger` package to allow instantiating a new logger (rather than using the global one) and then created a `logger` field on the event server itself, where that logger does not have a hook for logging events. This allows the event server to use it for logging anywhere without getting into deadlock.
- [ ] Finish off testing, probably with some events omitted out for now, because the requestor is missing on some events. All I've done here is change the existing test to create a profile instead of an instance. We're in no way checking every lifecycle event, so coverage is low. Happy to add more tests as a follow up. The test that is added demonstrates the principle.

tomponline:
> * To do this I've refactored the `shared/logger` package to allow instantiating a new logger (rather than using the global one) and then created a `logger` field on the event server itself, where that logger does not have a hook for logging events. This allows the event server to use it for logging anywhere without getting into deadlock.

Can this be a separate standalone PR please

tomponline:
> * To do this I've changed the `requestor` field of `operations.Operation` to be a `request.Requestor`, which can be converted to an `api.OperationRequestor` or `api.EventLifecycleRequestor` as necessary.


I've not followed what this part means, but im sure you'll explain it in the separate PR when it arrives. Thanks

tomponline:
Then we can have this PR just add the event filtering change.

Helps keep the size and scope of PR reduced. Thanks

markylaing:
Now depends on #16443 #16444 - will rebase when these are merged.

doc/rest-api.yaml
markylaing: I've added an `api.OperationRequestor` just in case. It was a little weird for `api.Operation` to contain an `api.EventLifecycleRequestor` anyway.

test/suites/auth.sh
markylaing: We're not going to audit every lifecycle event for this PR, so we'll no longer consider this a blocker. I've changed the test to use profile creation events instead.

markylaing:
@tomponline rebased and ready when you are, thanks.

test/suites/auth.sh
tomponline: What are we doing instead then?

markylaing: The test I've added checks that:
- An identity without `can_view_events` on `default` cannot see lifecycle events for that project when the event requestor is not them.
- An identity with `can_view_events` on `default` can see lifecycle events for that project when the event requestor is not them.
- An identity without `can_view_events` on `default` can see lifecycle events for that project if they are the event requestor.

I've utilised profile creation to do this, because `instance-created` events don't actually have a requestor (they should - I'll raise a bug).

markylaing: Oh I already did: https://github.com/canonical/lxd/issues/16072

tomponline: @markylaing @edlerd does https://github.com/canonical/lxd/issues/16072 need to be fixed to address the UI blocking issue?

tomponline: > The test I've added checks that:
> 
> * An identity without `can_view_events` on `default` cannot see lifecycle events for that project when the event requestor is not them.
> * An identity with `can_view_events` on `default` can see lifecycle events for that project when the event requestor is not them.
> * An identity without `can_view_events` on `default` can see lifecycle events for that project if they are the event requestor.
> 
> I've utilised profile creation to do this, because `instance-created` events don't actually have a requestor (they should - I'll raise a bug).

So more concisely, if you don't have `can_view_events` you can only view events where the requestor is yourself?

tomponline:
> * Callers with `can_view_privileged_events` see all events.

@markylaing did this change, could they not see all events before?

tomponline:
> * The `can_view_events` entitlement grants less privilege, since it only concerns lifecycle events.

What did it do before, please can you explain what has changed in more detail? Thanks

tomponline:
> so I think this should remain gated by the `can_view_events` entitlement.

Agree. Although there's an argument that it should be gated by `can_view_privileged_events` right?

tomponline:
> Also note that the `filter` function that is now accepted by `(*events.Server).AddListener` **MUST NOT** make any calls to the default logger via the `logger` package. Otherwise the event server goes into deadlock. ~I did investigate this but felt my solution with buffered channels was overkill. We can discuss in review.~ The filter function accepts a logger as an argument. This logger can be used without causing deadlock.

This is out of date now right @markylaing and should be removed/updated?

doc/api-extensions.md
tomponline, line 1: @markylaing im interested why this needs to be added to the public API in order to do internal event filtering?

Are you trying to also tackle and unstated additional aim?

markylaing: I added a `Requestor` field to `api.Operation` so that this could be read from event metadata. Otherwise there would be no way to perform the filtering on operation events that had been forwarded by other cluster members.

markylaing:
> > Also note that the `filter` function that is now accepted by `(*events.Server).AddListener` **MUST NOT** make any calls to the default logger via the `logger` package. Otherwise the event server goes into deadlock. ~I did investigate this but felt my solution with buffered channels was overkill. We can discuss in review.~ The filter function accepts a logger as an argument. This logger can be used without causing deadlock.
> 
> This is out of date now right @markylaing and should be removed/updated?

It's still absolutely a problem if e.g. `logger.Error(..)` is called within the filter. This will cause a deadlock. I added the logger as an argument to the filter so that it's clear that that is the logger that should be used

markylaing:
> > * Callers with `can_view_privileged_events` see all events.
> 
> @markylaing did this change, could they not see all events before?

This allowed them to see logging events, but they still needed `can_view_events` on a project in order to view events for that project. They likely will have had this entitlement anyway, a `server` `admin` has `can_view_privileged_events` and also has `can_view_events` on all projects.

markylaing:
> > * The `can_view_events` entitlement grants less privilege, since it only concerns lifecycle events.
> 
> What did it do before, please can you explain what has changed in more detail? Thanks

`can_view_events` now only allows viewing lifecycle events. Whereas previously it allowed the caller to view operation events as well.

test/suites/auth.sh
markylaing: Yes. This is updated now so that:
- If the event type is "operation" you can only view it if you can `can_view_operations` on the project OR you are the owner of the operation.
- If the event type is "lifecycle" you can only view it if you can `can_view_lifecycle_events` on the project OR you triggered the event.

markylaing:
Have confirmed that cluster upgrade works with the patch. An API extension is required for this patch, however I'm happy to land that separately with another API extension if required.

tomponline:
@markylaing ready for rebase

shared/api/operation.go
```diff
@@ -68,6 +68,28 @@ type Operation struct {
...
+	Requestor *OperationRequestor `json:"requestor,omitempty" yaml:"requestor,omitempty"`
+}
+
+// OperationRequestor represents the initial requestor of an operation
+//
+// API extension: operation_requestor.
+type OperationRequestor struct {
+	// Username is the username of the requestor. This is the identifier of the identity, or the username if using the unix socket.
+	// Example: jane.doe@example.com
+	Username string `yaml:"username" json:"username"`
```
tomponline, line 84: @markylaing is Username here following an existing convention, as I've seen elsewhere, such as @MusicDin PR that he refers to them as Identity IDs?

https://github.com/canonical/lxd/pull/16278/files#diff-7e939cda8409558eced96e827115c6e205e43c24e704c2549c12e806bcca4de3R178

Should we be using the same concept name universally?

lxd/patches.go
```diff
@@ -1788,4 +1790,25 @@ func patchRemoveInstancesPlacementScriptlet(name string, d *Daemon) error {
 	})
 }
 
+func patchEventEntitlementNames(name string, d *Daemon) error {
+	s := d.State()
+	return s.DB.Cluster.Transaction(d.shutdownCtx, func(ctx context.Context, tx *db.ClusterTx) error {
+		q := `UPDATE auth_groups_permissions SET entitlement = ? WHERE entitlement = ? AND entity_type = ?`
+
+		// Rename `can_view_events` on `project` to `can_view_lifecycle_events`.
```
tomponline: @markylaing what do you think about having `can_view_events` for project level lifecycle events too, so we don't have 2 types of "event" permissions? This way it aligns with having the same permission `can_view_operations` at server and project level.

Remind me again why you feel the need to separate `can_view_operations` from `can_view_events`, the PR description captured the *what* but not the *why*.

shared/api/operation.go
markylaing: It's the identifier for the identity. I'm happy to call it that, but here I went with consistency with the existing `api.EventLifecycleRequestor`.

markylaing: Also `IdentityIdentifier` is a little verbose for my liking.

lxd/patches.go
markylaing: I'm happy to keep `can_view_events` on `project` as is. The rename was to make it clear that the entitlement applies to lifecycle events only.

We already have `can_view_operations` at the project level. The reason for separating it at `server` level is so that we have some entitlement that defines access for operations that don't have a project (such as server background tasks).

shared/api/operation.go
tomponline: OK so @MusicDin should use same naming (username) in his pr to avoid confusion.

lxd/patches.go
tomponline: OK cool let's leave  can_view_events as is and introduce the operations perm at project level

tomponline:
@markylaing ready for rebase, please let me know when you want another review, ta

lxd/patches.go
markylaing: The operations perm already exists at project level. I think for now this PR can be simplified by:
1. Renaming `can_view_privileged_events` on `server` to `can_view_events`. This will grant access to `ovn` and `logging` events, plus lifecycle events that are not specific to a project.
2. Adding `can_view_operations` at server level. This will grant access to view server level operations, both via the operations API (in #16453) and via the events websocket.
3. Keeping `can_view_events` at project level. This will be only for project specific lifecycle events via the events websocket.
4. Keeping `can_view_operations` at project level. This will be for the operations API, and project specific operation events in the events websocket.

tomponline: Sounds good to me!

One question, does `can_view_privileged_events` server level see operations today? And after the proposed change will not right?

If so should we have a patch to assign `can_view_operations` to anyone that has `can_view_privileged_events` currently?

markylaing:
Ready for another round @tomponline thanks

lxd/patches.go
markylaing: I'll need to think on this a little as it a bit complicated. Server level operations and lifecycle events were never properly handled, because they were (almost) all emitted in the default project before #16464.

I'll get back to you

markylaing: I've looked into this and I think we shouldn't patch to automatically grant `can_view_operations` based on any current entitlements.

`tl;dr`: Viewership of server level events was granted to admins by way of computed relation to `can_view_events` in the default project. It's always been wrong and this PR fixes it.

`can_view_privileged_events` has only ever granted a group permission to view `logging` events. This PR will change this so that `can_view_privileged_events` is required to view `logging` events, `ovn` events, and any server level lifecycle event.

Currently, all access checks for events are project specific. If an operation or lifecycle event does not have a project, then no access check is performed. This means that a fine-grained identity calling `GET /1.0/events?all-projects=true` will receive all operation and lifecycle events where no project is set. 

Prior to #16464, all server level lifecycle events were emitted in the default project. So the fine-grained identity would require `can_view_events` in the default project. After #16464 this requirement was removed, because the event server only checks access if the event has a project.

tomponline:
@markylaing can you double check the spelling in the docs please

markylaing:
> @markylaing can you double check the spelling in the docs please

Hmmm weird that `bcache` is caught, I don't recall adding that.

markylaing:
> > @markylaing can you double check the spelling in the docs please
> 
> Hmmm weird that `bcache` is caught, I don't recall adding that.

Fixed - I borked a rebase

lxd/events.go
```diff
@@ -162,6 +180,58 @@ func eventsSocket(s *state.State, r *http.Request, w http.ResponseWriter) error
...
+		// Unmarshal the requestor from the event metadata.
+		var m requestorMetadata
+		err := json.Unmarshal(event.Metadata, &m)
+		if err != nil {
+			l.Error("Failed to unmarshal event metadata during requestor filtering")
+			return false
+		}
+
+		// Allow the event if the same requestor is connected. Otherwise, filter it out.
+		return m.Requestor != nil && m.Requestor.Username == requestor.CallerUsername() && m.Requestor.Protocol == requestor.CallerProtocol()
```
tomponline, line 232: @markylaing just a nit, not a blocker, but im not a fan of these special case return statements at the end of a function, prefer to have each condition a separate `if` block with a simple default return value at the bottom. This makes it easier to refactor the function in the future in my experience.

markylaing: Hmmm yeah that makes sense. Will change (either later today or in follow up)

tomponline (approved):
LGTM thanks!

tomponline:
@markylaing does this close any issues?

markylaing:
> @markylaing does this close any issues?

I'll double check

PR #5277, opened 2018-11-14
Snapshot scheduling

monstermunchkin:
This resolves #2610

Issue #2610, opened 2016-11-14
Snapshot scheduling


nlienard:
Hello

Is it possible to implement some extra snapshot feature (ex:  TTL on snapshot)  in order to have them automatically created and deleted when TTL expired.

For the moment i have to use "zfSnap" in crontab; a tool which add a TTL to snapshot and automatically delete the snapshot when TTL is expired.

     # Snap LXC
     0 * * * * /infra/tools/lxd_snap.sh > /tmp/cron_lxd_snap.log 2>&1

basically, it runs : 
/usr/sbin/zfSnap -s -z -a 24h -r zdata/lxd/containers

another cron at midnight to delete old snap  : /usr/sbin/zfSnap -d -s -S

Results is to have a snap per hour : 

    root@inxovh-hy002 (node2):~# ls -la /var/lib/lxd/containers/inxovh-adm001.zfs/.zfs/snapshot/
    total 0
    dr-xr-xr-x 2 root root 2 Nov 14 11:11 .
    dr-xr-xr-x 1 root root 0 Nov  7 20:50 ..
    dr-xr-xr-x 1 root root 0 Nov 14 11:57 2016-11-14_10.45.00--24h
    dr-xr-xr-x 1 root root 0 Nov 14 11:57 2016-11-14_10.46.00--1w
    dr-xr-xr-x 1 root root 0 Nov 14 11:57 2016-11-14_11.00.00--24h
    dr-xr-xr-x 1 root root 0 Nov 14 11:57 2016-11-14_11.11.00--24h

So, that means, snapshot with 24h TTL will be deleted tomorrow and the one with "1w" will be deleted in 7 days.

Using snapshot is useful *not to restore an entire container* (it almost never happens) but to restaure specific files for any reason (- bad modification, deleted by error; wrong update after a patch etc).

So, would it be possible to add a snapshot feature which integrates:

- a scheduler to plan automatic snapshot for containers
- a TTL for each snapshot
- a scheduler to plan automatic remove of old snapshot according to its TTL
- a way to restore specific files without having to to to .zfs/snapshots/files  : 

lxc snapshot <container_name> <frequency> <TTL>

ex:

    lxc snapshot inxovh-adm001  "0 * * * *" 24h
    => will add scheduler to do an hourly snap with 24h TTL for container named  inxovh-adm001 

    lxc snapshot inxovh-adm001  "0 23 * * 7" 1w
    => will add scheduler to do a snap every sunday at 23h00 for container named inxovh-adm001

[...]

PR #15998, opened 2025-07-12
Disable template recursion and add unit tests for `snapshots.pattern`

simondeziel:


shared/util.go
```diff
@@ -1349,7 +1349,7 @@ func RenderTemplate(template string, ctx pongo2.Context) (string, error) {
 
 	// Looks like we're nesting templates so run pongo again
 	if strings.Contains(ret, "{{") || strings.Contains(ret, "{%") {
-		return RenderTemplate(ret, ctx)
+		return "", errors.New("Template contains nested template tags, which is not allowed")
```
tomponline: Shall we limit nesting depth instead to avoid breaking changes? Does the original commit that added that support hint as to a use case we can use as an indicator of a realistic nesting depth limit?

simondeziel: This was introduced by #5277 which addressed #2610. I don't see any justifications for recursion but will add it back with a limit of 5 unless you can come up with a better magic number :)

lxd/instance/instance_utils.go
```diff
@@ -997,7 +997,7 @@ func NextSnapshotName(s *state.State, inst Instance, defaultPattern string) (str
 
 	pattern, err = shared.RenderTemplate(pattern, pongo2.Context{
 		"creation_date": time.Now(),
-	})
+	}, 5)
```
tomponline, line 1000: Can we have a constant for this and a comment explaining why?

simondeziel: https://github.com/canonical/lxd/pull/16022

PR #16001, opened 2025-07-14
lxd: Reorganise checks in images GET handler.

markylaing:
Enforce that an untrusted caller cannot list images in non-default projects before making database calls. Additionally, don't check the effective project if the caller is not trusted, since the request must be for the default project, which must have images enabled.


tomponline (approved):
Thanks!

PR #16004, opened 2025-07-14
tests: Test instance file push using strings.Reader

MusicDin:
Test file push using `strings.Reader` which causes the `Content-Length` to be set by `(net/http).NewRequest()`. This test is for https://github.com/canonical/lxd/pull/15974


tomponline:
@MusicDin tests failing

tomponline (approved):
Thanks!

PR #16005, opened 2025-07-14
Correctly set project of private image remote on init/launch/refresh

markylaing:
In #15957 I disallowed creation of image secrets for images that are not in the default project. This turns out to be a mistake because it's possible to download a private image from any project (as exemplified by `lxc image copy`). In any case, the change caused a test failure in `clustering_groups` that took a little while to investigate.

The line that failed is https://github.com/canonical/lxd/blob/7c479927ae89ca218338abc9cdbf527725eeaef5/test/suites/clustering.sh#L4058

The command expects to initialise instance `c1` on remote `cluster` in project `foo`. The `testimage` is an alias of the busybox image that was imported earlier in the suite to the default remote; the image is private.

The only reason that this command currently succeeds is that the busybox image was imported into the `default` project in the `cluster` remote earlier in the suite. Here is a run down of what is happening:
1. Client queries `(local) GET /1.0`.
2. Client queries `(local) GET /1.0/images/aliases/testimage?project=foo` on the local remote. 
    - This is because the `--project` flag is being applied to both the image and instance remotes. 
    - In this context, the `--project` flag should only mean "where to put the instance", and not "where to get the image". 
    - Even though the image was not imported into project `foo`, this works because project `foo` has `features.images=false`.
4. Client queries `(local) GET /1.0/images/{fingerprint}?project=foo` using the fingerprint of the alias.
5. Client queries `(local) POST /1.0/images/{fingerprint}/secret?project=foo` because the image is private.
6. Client creates an `api.InstanceSource` using the `api.Image`, the image secret, and connection details from `GET /1.0`. Then queries `(cluster) POST /1.0/instances?project=foo`. Crucially, the `project` field is omitted from the instance source.
7. The `cluster` remote sees that it already has an image with the same fingerprint and continues with that image.

At step 7, if the `cluster` remote did not already have the image, it would make an untrusted call to `GET /1.0/images/{fingerprint}?secret={secret}`. This call would fail because the image token operation is defined in project `foo` (and not `default`, which is implicit from the lack of a `project` parameter).

When image secrets were disallowed for non-default projects, this failed at step 4.

This PR does three things to fix this:
1. Use the current project for the image remote when performing an `init`, `launch`, or `refresh`. We can later add a flag to override this (such as `--image-project` or `--source-project`) but crucially, the `--project` flag is not used for the image remote.
2. Set the project in the `api.InstanceSource` before performing the `POST /1.0/instances`.
3. Plumb the source project into the `ImageDownloadArgs` when creating or refreshing an instance.

I didn't add any API extensions for this because all the relevant fields were already in place, they just needed to be connected up. 

This is a change in behaviour for `init`/`launch`/`refresh` if the image remote is a private LXD server, but I don't think it's likely that anyone is relying on this behaviour.

### DIFF
--- a/test/suites/remote.sh
+++ b/test/suites/remote.sh
@@ -397,7 +397,38 @@ test_remote_usage() {
   # The alias should be set to `bar`.
   [ "${alias}" = "bar" ]
 
+  # Clean up
+  lxc_remote delete lxd2:c1 -f
+  lxc_remote image delete lxd2:bar
   lxc_remote image alias delete localhost:foo
+  lxc image list -f csv -c f | xargs -I{} lxc image delete {} # Delete all images (there are none in any other projects)
+
+  # Test instance init where image is in non-default project.
+  lxc_remote project create lxd2:foo
+  LXD_DIR="${LXD2_DIR}" deps/import-busybox --project foo --alias foo-img
+  ! lxc init lxd2:foo-img foo-container || false
+  lxc_remote project switch lxd2:foo
+  lxc init lxd2:foo-img foo-container
+
+  # Revert changes
+  lxc delete foo-container
+  lxc image delete "$(lxc image list -f csv -c f)"
+  lxc_remote project switch lxd2:default
+
+  # Test instance init where image is in non-default project and the target project is also non-default
+  lxc project create foo2
+  lxc profile show default | lxc profile edit default --project foo2
+  ! lxc init lxd2:foo-img foo-container --project foo2 || false
+  lxc_remote project switch lxd2:foo
+  lxc init lxd2:foo-img foo-container --project foo2
+  lxc_remote project switch lxd2:default
+
+  # Final clean up
+  lxc_remote image delete lxd2:foo-img --project foo
+  lxc_remote project delete lxd2:foo
+  lxc delete foo-container --project foo2
+  lxc image delete --project foo2 "$(lxc image list -f csv -c f --project foo2)"
+  lxc project delete foo2
 
   lxc_remote remote remove lxd2
   lxc_remote remote remove lxd2-public
--- a/client/lxd.go
+++ b/client/lxd.go
@@ -499,6 +499,12 @@ func (r *ProtocolLXD) getSourceImageConnectionInfo(source ImageServer, image api
 		return nil, nil
 	}
 
+	// If the image source is a private LXD server, set the project of the client as the project for the instance source.
+	instanceServer, ok := source.(*ProtocolLXD)
+	if ok && instanceServer.project != "" {
+		instSrc.Project = instanceServer.project
+	}
+
 	// Minimal source fields for remote image
 	instSrc.Mode = "pull"
 
--- a/lxc/utils.go
+++ b/lxc/utils.go
@@ -393,6 +393,16 @@ func getImgInfo(d lxd.InstanceServer, conf *config.Config, imgRemote string, ins
 		if err != nil {
 			return nil, nil, err
 		}
+
+		// Check if the image server is actually a private LXD server. If it is, tell it to use the project that is
+		// configured for that remote. This is because "GetImageServer" will overwrite the configured project with the
+		// --project flag, but for init, launch, and rebuild we don't want to apply the project override to the image
+		// remote. Contextually, the --project flag in this case is for the private LXD server where the instance is
+		// being created, not the image source.
+		instanceServer, ok := imgRemoteServer.(lxd.InstanceServer)
+		if ok {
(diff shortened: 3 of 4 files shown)


markylaing:
The tests fail before this change and pass afterwards.

tomponline:
> 1. Use the current project for the image remote when performing an `init`, `launch`, or `refresh`. We can later add a flag to override this (such as `--image-project` or `--source-project`) but crucially, the `--project` flag is not used for the image remote.

"Current project" is the temporarily set the same as `--project` if the `--project` is specified right?

test/suites/remote.sh
```diff
@@ -397,7 +397,38 @@ test_remote_usage() {
...
+  lxc_remote project create lxd2:foo
+  LXD_DIR="${LXD2_DIR}" deps/import-busybox --project foo --alias foo-img
+  ! lxc init lxd2:foo-img foo-container || false
+  lxc_remote project switch lxd2:foo
+  lxc init lxd2:foo-img foo-container
+
+  # Revert changes
+  lxc delete foo-container
+  lxc image delete "$(lxc image list -f csv -c f)"
+  lxc_remote project switch lxd2:default
```
roosterfish, line 416: Just cosmetic but do you think it's worth deleting `lxd2:foo` project here too and recreate it for the test below so its easier to follow the test? 

Otherwise at a first glance you might not get why `! lxc init lxd2:foo-img foo-container --project foo2 || false` below is failing.
You need to connect the dots that it also still has the image from the test before.

roosterfish: Or let's just have a comment mentioning this in the test below.

tomponline:
@skozina what are your thoughts on this one?

@markylaing is proposing to use the current project in the `lxc` CLI's config.yaml (the one switched to with with `lxc project switch`) to guide the project used for an image source service, and use the `--project` override flag to only influence the target instance server.

This is a new paradigm as far as we can work out, with the exception of `lxc image copy` which appears to be using `--project` and `--target-project` as the same value and should be fixed.

I'm hesitant to introduce new CLI paradigms to the old CLI, especially if its used in only one place.

It'd be more consistent to fix `lxc image copy` and then introduce a `--target-project` to `lxc [init|launch]` but that would be very disruptive.

@markylaing proposal is less disruptive but  at the cost of doing something novel for this one command; breaking the relationship between `lxc project switch` and the `--project` flag.

test/suites/remote.sh
markylaing: Will update, depending on the outcome of https://github.com/canonical/lxd/pull/16005#issuecomment-3070710134

markylaing:
> This is a new paradigm as far as we can work out, with the exception of `lxc image copy` which appears to be using `--project` and `--target-project` as the same value and should be fixed.

I was concerned that this might break users, so I recommended that we just fix the help text in the issue (see the bottom of the description of #16028). This behaviour has been around since LXD 4.20 apparently.

> @markylaing proposal is less disruptive but at the cost of doing something novel for this one command; breaking the relationship between `lxc project switch` and the `--project` flag.

This PR aligns `lxc init` with `lxc image copy`, where the project used for the image/source remote is defined by the current project (via `lxc project switch`) and is not overridden by the `--project` flag. The `--project` flag only overrides the project for the instance/destination remote.

tomponline:
> This PR aligns `lxc init` with `lxc image copy`, where the project used for the image/source remote is defined by the current project (via `lxc project switch`) and is not overridden by the `--project` flag. The `--project` flag only overrides the project for the instance/destination remote.

Yes, except once that `lxc image copy` bug is fixed to correctly use `--target-project` then it wont align anymore.
Thus this change will remain the exception to all other commands AFAIK.

skozina:
Just adding a note for myself here, we've agreed not to introduce the new paradigm of breaking the `lxc project switch` and `--project` flag. Instead we'd just revert the changes done to `lxc image copy` to fix the non-standard behavior there, and rely on `--target-project` instead.
That, and we should also review our docs and `--help` messages to be consistent with the behavior of `--project` and `--target-project`.

tomponline:
> Just adding a note for myself here, we've agreed not to introduce the new paradigm of breaking the `lxc project switch` and `--project` flag. Instead we'd just revert the changes done to `lxc image copy` to fix the non-standard behavior there, and rely on `--target-project` instead. That, and we should also review our docs and `--help` messages to be consistent with the behavior of `--project` and `--target-project`.

And we are going introduce a `--target-project` flag to `lxc [init|launch|rebuild]` commands.

tomponline:

markylaing:
Yep go for it

PR #16019, opened 2025-07-14
lxd: Handle DevLXD image export directly in DevLXD handler

markylaing:
Auth handling in the image export handler is complicated because:
- Untrusted callers can use the endpoint if the image they are requesting is public.
- Both trusted and untrusted callers may provide a secret, which if valid overrides any other auth decisions.
- The handler is also called by the DevLXD handler directly. This bypasses the AccessHandler and has it's own rules.

This PR extracts the logic to actually return the image export response into its own function, then implements image export in the DevLXD handler directly.

### DIFF
--- a/lxd/devlxd.go
+++ b/lxd/devlxd.go
@@ -16,6 +16,8 @@ import (
 
 	"github.com/canonical/lxd/lxd/auth"
 	"github.com/canonical/lxd/lxd/cloudinit"
+	"github.com/canonical/lxd/lxd/db"
+	"github.com/canonical/lxd/lxd/db/cluster"
 	"github.com/canonical/lxd/lxd/events"
 	"github.com/canonical/lxd/lxd/instance"
 	"github.com/canonical/lxd/lxd/lifecycle"
@@ -265,6 +267,11 @@ var devLXDImageExportEndpoint = devLXDAPIEndpoint{
 	Get:  devLXDAPIEndpointAction{Handler: devLXDImageExportHandler},
 }
 
+// devLXDImageExportHandler returns a file response containing the image files. The requested fingerprint must match
+// exactly, and the project must be "default". Images are only made available over DevLXD if they are public or cached.
+//
+// Note: This endpoint used to call into the image export handler directly, it therefore returns full API responses
+// rather than DevLXDErrorResponses for compatibility.
 func devLXDImageExportHandler(d *Daemon, r *http.Request) response.Response {
 	_, err := getInstanceFromContextAndCheckSecurityFlags(r.Context(), devLXDSecurityKey, devLXDSecurityImagesKey)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


tomponline:
@markylaing tests failing

markylaing:
> @markylaing tests failing

Failed on "Post install python": https://github.com/canonical/lxd/actions/runs/16268050722/job/45928336934?pr=16019#step:54:2

I've just rerun it

lxd/images.go
```diff
@@ -4347,12 +4330,16 @@ func imageExport(d *Daemon, r *http.Request) response.Response {
...
+}
 
-	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+// imageExportFiles returns the [response.FileResponse] for the specified image files, the image can be local or remote.
+func imageExportFiles(ctx context.Context, s *state.State, imgInfo *api.Image, requestProjectName string) response.Response {
+	var address string
+	err := s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
 		// Check if the image is only available on another node.
+		var err error
 		address, err = tx.LocateImage(ctx, imgInfo.Fingerprint)
```
kadinsayani, line 4342:
```suggestion
		address, err := tx.LocateImage(ctx, imgInfo.Fingerprint)
```

lxd/devlxd.go
```diff
@@ -278,7 +285,44 @@ func devLXDImageExportHandler(d *Daemon, r *http.Request) response.Response {
 		return response.Forbidden(err)
 	}
 
-	return imageExport(d, r)
+	fingerprint, err := url.PathUnescape(mux.Vars(r)["fingerprint"])
+	if err != nil {
+		return response.SmartError(err)
+	}
+
```
kadinsayani, line 292: Should we add fingerprint validation here?

markylaing: I can do :) I'm not sure it's necessary though. The endpoint is gated by `security.devlxd.images` and the SQL query performs an exact match.

lxd/images.go
markylaing: I can't make this change as it would shadow the `address` variable that is defined outside of the scope of the function.

What I will do however, is remove the `context.TODO()` on the preceding line and change to the given context.

markylaing: Done.

kadinsayani: Gotcha, I missed spotting the shadowing. Thanks!

lxd/devlxd.go
kadinsayani: Thanks for clarifying. In that case, I don't think its necessary.

kadinsayani (dismissed):

lxd/devlxd.go
```diff
@@ -278,7 +285,44 @@ func devLXDImageExportHandler(d *Daemon, r *http.Request) response.Response {
...
+	if projectName != api.ProjectDefaultName {
+		// Disallow requests made to non-default projects.
+		return response.NotFound(nil)
+	}
+
+	s := d.State()
+
+	var imgInfo *api.Image
+	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+		// Perform exact match on image fingerprint and project.
```
tomponline, line 303: Is this truly exact match?

lxd/devlxd.go
```diff
@@ -278,7 +285,44 @@ func devLXDImageExportHandler(d *Daemon, r *http.Request) response.Response {
...
+		if err != nil {
+			return err
+		}
+
+		imgInfo, err = dbImage.ToAPI(ctx, tx.Tx(), api.ProjectDefaultName)
+		return err
+	})
+	if err != nil && api.StatusErrorCheck(err, http.StatusNotFound) {
+		// Return a generic not found response. If an image is found but it is not public or cached, we'll later return
+		// a generic not found error, so we need to mask this error to avoid information leakage.
```
tomponline: Didn't follow

> we'll later return
		// a generic not found error, so we need to mask this error to avoid information leakage.

We are returning the generic here are we not?

lxd/devlxd.go
markylaing: Yes, the `GetImage` function is generated by lxd-generate and uses `WHERE ( project = ? AND images.fingerprint = ? )` (where `project` is `projects.name` from a join).

lxd/devlxd.go
markylaing: Yes we are. On line 322 a generic not found error is also returned. So this is to make the response consistent in both cases.

tomponline: OK please can you have a go at clarifying that comment, as it seems to indicate they are related somehow.

lxd/devlxd.go
tomponline: cool, thanks.

lxd/devlxd.go
markylaing: Will do :)

markylaing: I realised there's a way to write the handler to make it more efficient and (hopefully) clearer.

tomponline (approved):

Issue #15492, opened 2025-04-30
LXD wont start if has listeners bound to managed bridge addresses when evacuated


tomponline:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

24.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
N/A
```

### Output of "lxc info" or system info if it fails

```shell
N/A
```

### Issue description

The `init()` function has this section before the listeners are started:

```go
	// Setup the networks.
	if !d.db.Cluster.LocalNodeIsEvacuated() {
		logger.Infof("Initializing networks")

		err = networkStartup(d.State)
		if err != nil {
			return err
		}
	}
```

https://github.com/canonical/lxd/blob/f56d1ba3ff6d53ceda06708afb1b7a2be0fecd6d/lxd/daemon.go#L1744-L1752

But if the LXD member was evacuated and then shut down, LXD will then not start up again if the subsequent listeners are configured to listen on one of the managed bridge network addresses.

This bug allows users to inadvertently render their LXD server unstartable.

### Steps to reproduce

lxc network create lxdbr0
lxc config set core.metrics_address=<IP of lxdbr0 bridge>
lxc cluster evacuate
systemctl stop snap.lxd.daemon
systemctl start snap.lxd.daemon # It will fail to start

### Information to attach

- [ ] Any relevant kernel output (`dmesg`)
- [ ] Instance log (`lxc info NAME --show-log`)
- [ ] Instance configuration (`lxc config show NAME --expanded`)
- [ ] Main daemon log (at `/var/log/lxd/lxd.log` or `/var/snap/lxd/common/lxd/logs/lxd.log`)
- [ ] Output of the client with `--debug`
- [ ] Output of the daemon with `--debug` (or use `lxc monitor` while reproducing the issue)

PR #16035, opened 2025-07-15
Waitready enhancements

roosterfish:
Implements the first portion of changes from the [waitready enhancements](https://docs.google.com/document/d/1Rf46WtmkFnZ1wA1h7uoOSuqtwVgrBCxlysCnBOiJkvU/edit?tab=t.0) spec which allow using the `lxd waitready --network` command during cluster evacuation and restore.

Also fixes https://github.com/canonical/lxd/issues/15492 implicitly.

The PR introduces two new functions `Evacuate()` and `Restore()` on the networks interface which are implemented by the `bridge`, `physical` and `ovn` type networks to remote/setup BGP related peers and prefix as well as the active chassis in case of OVN.

By extending the internal `networkStartup` to be able to cope with both network initialization and restore we can safely revert the changes introduced in https://github.com/canonical/lxd/commit/97b597e4d5629e0fdaa6f71e844685e26ccb2085 and always start networks when initializing an evacuated cluster member.

As an effect for the user you can now restart an evacuated cluster member and run `lxd waitready --network --storage` to wait until those resources are ready on this cluster member so you can safely run `lxc cluster restore <name>` without fearing the to be restored instances are consuming any non-ready resources.

This also lays the groundwork for the second step which adds a pre-check to `lxc cluster restore` that allows checking for resource readiness even before initiating the restore operation.


roosterfish:
I have checked the OVN chassis and BGP prefixes are removed accordingly. The chassis doesn't anymore show using `ovn-nbctl find ha_chassis` and in case of the upstream BGP (when connected to all LXD cluster members as peers), the route information changes accordingly and doesn't anymore use the evacuated peer.

~~Bridge and physical network tests pending.~~

~~I'll also add some additional checks to the waitready test suite.~~

roosterfish:
After this one lands into `main`, we should also backport it into `5.21` to make use of it in the MicroCloud pipelines. As we use `5.21/edge` there we don't have to wait for the next release and can proceed.

lxd/networks.go
```diff
@@ -1786,6 +1808,12 @@ func networkStartup(stateFunc func() *state.State, restoreOnly bool) error {
 			for pn := range initNetworks[priority] {
 				err := loadAndInitNetwork(s, pn, priority, true, restoreOnly)
 				if err != nil {
+					// When restoring a network the operation is not allowed to fail.
+					// The network is already started at this stage which might have taken multiple retries.
```
tomponline, line 1812: @roosterfish can it happen that LXD is still trying to start up a network in a loop in a different thread (started during LXD start up), and then `lxc cluster restore` is run before the local network is started, and then this will call `n.Restore()` before the network itself is started?

tomponline: also on the flip side, if a network fails to start up, and enters the start up loop, and then the member is evacuated using `lxc cluster evacuate`, and then the local problem preventing the network starting up is resolved, could the network complete its normal start up without taking into account the member has been evacuated?

tomponline: Maybe we should add a "local member is evacuated" indicator to `State` and then it can be accessed on each internal iteration, via:

```go
s := stateFunc() // Get fresh state in case global config has been updated.
```

tomponline: or if thats going to be too invasive to reason about at this time, then we can at least run `d.db.Cluster.LocalNodeIsEvacuated()` internally for each iteration here?

roosterfish: > can it happen that LXD is still trying to start up a network in a loop in a different thread (started during LXD start up), and then lxc cluster restore is run before the local network is started

For this I have already added a check inside `networkStartup()`:

```go
if n.Status() != api.NetworkStatusCreated {
	return fmt.Errorf("Cannot restore network %q when not in created state", n.Name())
}
```

But I just realized instead of `Status` this should rather be `LocalStatus` as the networks non-local state will be `Created` even if the local one isn't yet started. 

For the evacuation I haven't added a check as `networkStop()` doesn't return errors but only logs if something didn't work. But I agree there should be at least an error in the logs indicating we couldn't evacuate the network.

To fix this properly, why not also adding a pre-check to `lxc cluster evacuate` and block evacuation of cluster members until all networks/storage pools are started? In the follow up PR to this one I'll add such a check for `lxc cluster restore` but we can just perform the same for evacuation too which makes most sense IMHO.

tomponline: Sounds good thanks!

tomponline (approved):
LGTM thanks!

lxd/network/driver_ovn.go
```diff
@@ -3101,14 +3116,10 @@ func (n *ovn) Start() error {
 
 		// Check if we should enable the chassis.
 		chassisEnabled, err = n.chassisEnabled(ctx, tx)
-		if err != nil {
-			return err
-		}
-
-		return nil
+		return err
```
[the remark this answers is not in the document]
roosterfish: Looks like a false positive when not adding additional context and simply returning the error (either `nil` or an actual error) as it was before.

PR #16036, opened 2025-07-15
test: Add tests for public images APIs

markylaing:
These tests represent the current state of the images APIs from the perspective of an untrusted caller, or from a restricted or fine-grained TLS client with no projects/groups.

### DIFF
--- a/test/suites/image.sh
+++ b/test/suites/image.sh
@@ -1,5 +1,6 @@
 test_image_expiry() {
   local LXD2_DIR LXD2_ADDR
+  # shellcheck disable=2153
   LXD2_DIR=$(mktemp -d -p "${TEST_DIR}" XXX)
   spawn_lxd "${LXD2_DIR}" true
   LXD2_ADDR=$(cat "${LXD2_DIR}/lxd.addr")


test/suites/image.sh
```diff
@@ -199,3 +200,184 @@ test_image_refresh() {
...
+  query /1.0/images?project=default | jq -e '(.metadata | length) == 0 and .status_code == 200'
+  query /1.0/images?recursion=1 | jq -e '(.metadata | length) == 0 and .status_code == 200'
+  query /1.0/images?recursion=1\&project=default | jq -e '(.metadata | length) == 0 and .status_code == 200'
+
+  if [ -z "${CERT_NAME:-}" ]; then
+    # Untrusted callers see a generic 404 for project foo, or a 403 if using "all-projects".
+    query /1.0/images?project=foo | jq -e '.error == "Not Found" and .error_code == 404'
+    query /1.0/images?recursion=1\&project=foo | jq -e '.error == "Not Found" and .error_code == 404'
+    query /1.0/images?project=bar | jq -e '.error == "Not Found" and .error_code == 404'
+    query /1.0/images?recursion=1\&project=bar | jq -e '.error == "Not Found" and .error_code == 404'
```
tomponline, line 284: Can we use single quotes to avoid need to escape &?

tomponline (approved):

PR #16037, opened 2025-07-15
Auth: Use correct entity type when enriching storage pool access entitlements

### LINKED ISSUE #16033: Storage pool list ignores entitlements
### Please confirm

### Distribution

Ubuntu

### Distribution version

snap

```shell
latest/edge
```

```shell
-
```

### Issue description

The `GET /1.0/storage-pools?recursion=1&with-access-entitlements=can_edit,can_delete` request does not return any access information on the pools in the response.

When issuing `GET /1.0/storage-pools/lvm1?recursion=1&with-access-entitlements=can_edit,can_delete` on a pool that is also part of the list response, the `access_entitlements` field is filled.

Expectation is for the storage pool list endpoint to also include the `access_entitlements`.

### Steps to reproduce

1. Authenticate with a restricted TLS certificate and have server admin permissions.
2. Create a storage pool.
3. Fetch list of storage pools with entitlements.
4. See the entitlements missing from the api response.

### Information to attach

markylaing:
The `reportEntitlements` function accepts a `map[*api.URL]auth.EntitlementReporter`. 

If this map contains a single entry, then `(auth.Authorizer).CheckPermission` is called for each requested entitlement against the single URL. `CheckPermission` doesn't need an entity type because this can be derived from the given URL.

If the map contains multiple entries, then `(auth.Authorizer).GetPermissionChecker` is called for each requested entitlement, using the given `entity.Type`. This is done for efficiency.

In this case, the wrong entity type was passed into `reportEntitlements` and subsequently `GetPermissionChecker`. When the permission checker was called with storage pool URLs, it wasn't able to match this against any storage volume URLs that it held internally.

In fact, it was logging an error:
```
ERROR  [2025-07-15T12:17:00Z] Unexpected permission checker input URL       actual_entity_type=storage_pool driver=embedded-openfga entitlement=can_edit entity_type=storage_volume expected_entity_type=storage_volume protocol=oidc url=/1.0/storage-pools/foo username=test-user@example.com
```

This PR changes the entity type in the `reportEntitlements` call to `entity.StoragePool` so that entitlements are reported correctly when listing storage pools. Tests have been added to catch this case.

Closes #16033

### DIFF
--- a/lxd/storage_pools.go
+++ b/lxd/storage_pools.go
@@ -236,7 +236,7 @@ func storagePoolsGet(d *Daemon, r *http.Request) response.Response {
 	}
 
 	if len(withEntitlements) > 0 {
-		err = reportEntitlements(r.Context(), s.Authorizer, s.IdentityCache, entity.TypeStorageVolume, withEntitlements, urlToPool)
+		err = reportEntitlements(r.Context(), s.Authorizer, s.IdentityCache, entity.TypeStoragePool, withEntitlements, urlToPool)
 		if err != nil {
 			return response.SmartError(err)
 		}


tomponline:
Should that error scenario be changed from a log to an error response?

markylaing:
> Should that error scenario be changed from a log to an error response?

Potentially. That would require changing the `auth.PermissionChecker` signature. Often we're checking these things in a loop like:
```
for _, inst := range instances {
    if !canView(entity.InstanceURL(inst.Project, inst.Name)) {
        continue
    }
    
    /* ... */
}
```
so this would mean an error check on all of those calls. It's annoying but probably better then a silent failure and would be clearer to debug

markylaing:
> Should that error scenario be changed from a log to an error response?

Should this be a follow up? Or would you like me to address it now?

test/suites/auth.sh
```diff
@@ -1221,13 +1221,28 @@ entities_enrichment_with_entitlements() {
...
+
+  # Test multiple pools
+  lxc storage create foo dir
+  lxc auth group permission add test-group storage_pool foo can_edit
   all_storage_pools=$(lxc_remote query "oidc:/1.0/storage-pools?recursion=1&with-access-entitlements=can_edit,can_delete")
-  if ! jq -e -r --arg pool_name "$pool_name" '.[] | select(.name == $pool_name) | .access_entitlements | sort | @csv == "\"can_delete\",\"can_edit\""' <<< "$all_storage_pools";  then
-    echo "Second check failed (or pool not found)!" >&2
-    false
-  fi
+  jq -e -r --arg pool_name "$pool_name" '
```
tomponline: Please can you use long form flags where they exist, for documentation reasons.

Also can we have a comment explaining this complex command.

tomponline:
> > Should that error scenario be changed from a log to an error response?
> 
> Should this be a follow up? Or would you like me to address it now?

Can we include in this?

test/suites/auth.sh
markylaing: Gabriel added these checks a while ago and this is (almost) a copy/paste. It would be good to do a check of all of them as there was some unnecessary logic I removed.

markylaing:
@tomponline Before I refactor I just want to clarify.

The existing signature for the `PermissionChecker` is:
```go
type PermissionChecker func(entityURL *api.URL) bool
```
Which can only log errors, but has the advantage that it's usage is clean:
```go
if !checkAccess(myurl) {
    continue
}
```

Do you prefer:
```go
type PermissionChecker func(entityURL *api.URL) (bool, error)
```
with usage:
```go
hasAccess, err := checkAccess(myurl)
if err != nil {
} else if !hasAccess {
    continue
}
```

Or 
```go
type PermissionChecker func(entityURL *api.URL) error
```
with usage
```go
err := checkAccess(myurl)
if err != nil {
    if auth.IsDeniedError(err) {
        continue
    }

}
```

I am favouring the second option.

tomponline:
2nd option seems fine to me, both are clear though.

markylaing:
> 2nd option seems fine to me, both are clear though.

I've been working on this for a while now and it doesn't make sense. For these errors to be informative we need to surface the url that we failed to check permission with. We can't do that because we will likely expose URLs of entities that the caller can't view. For now I will leave it as an error log and we can reinvestigate in the future.

tomponline:
> For now I will leave it as an error log and we can reinvestigate in the future.

OK

markylaing:
This PR mainly now consists of improvements to the entitlement reporting tests.

test/suites/auth.sh
```diff
@@ -1199,6 +1199,23 @@ auth_ovn() {
...
+  # 5. Evaluates a conditional (asserting that the access entitlements are correct).
+  #
+  # When performing assertions on an array of responses, the "all" function is used (see https://jqlang.org/manual/#all).
+  # This function evaluates a conditional for all elements of an array, and returns true only if the conditional returns
+  # true for all elements. In all cases, this:
+  # 1. Calls "all".
+  # 2. Performs different assertions (steps 2-5 above) for each element in the array.
+  # 3. Has an "else" condition containing only "false". This asserts that we do not expect any extra array elements.
+  #
+  # Sometimes an --arg is passed into jq. This is to make bash variables referencable within jq.
```
simondeziel: `s/referencable/referenceable/` I think.

markylaing: Fixed.

test/suites/auth.sh
```diff
@@ -1446,12 +1508,10 @@ entities_enrichment_with_entitlements() {
...
   lxc auth group permission add test-group network_zone zone2 can_delete project=default
-  [ "$(lxc_remote query "oidc:/1.0/network-zones/zone1?project=default&with-access-entitlements=can_view,can_edit,can_delete" | jq -r '.access_entitlements | sort | @csv')" = '"can_view"' ]
-  [ "$(lxc_remote query "oidc:/1.0/network-zones/zone2?project=default&with-access-entitlements=can_view,can_edit,can_delete" | jq -r '.access_entitlements | sort | @csv')" = '"can_delete","can_edit","can_view"' ]
-  all_network_zones=$(lxc_remote query "oidc:/1.0/network-zones?project=default&recursion=1&with-access-entitlements=can_view,can_edit,can_delete")
-  if ! jq -e -r '
+  lxc_remote query "oidc:/1.0/network-zones/zone1?project=default&with-access-entitlements=can_view,can_edit,can_delete" | jq -e '.access_entitlements | sort | @csv == "\"can_view\""'
+  lxc_remote query "oidc:/1.0/network-zones/zone2?project=default&with-access-entitlements=can_view,can_edit,can_delete" | jq -e '.access_entitlements | sort | @csv == "can_delete","can_edit","can_view"'
+  lxc_remote query "oidc:/1.0/network-zones?project=default&recursion=1&with-access-entitlements=can_view,can_edit,can_delete" | jq -e -r '
```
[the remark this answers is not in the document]
markylaing: Fixed.

tomponline:
@markylaing ready for review?

markylaing:
Yep good to go :)

test/suites/auth.sh
tomponline, line 1: Great comment thanks!

tomponline (approved):
LGTM!

PR #16038, opened 2025-07-15
lxd/request: Never get existing Info when setting up request context

markylaing:
When setting up the context info, we should never use the existing value verbatim. This is because the UI multiplexes requests over HTTP2, and per https://pkg.go.dev/net/http#Request.Context, this context is cancelled when the clients connection is cancelled. This means that multiple requests are made on the same connection and they can all share a context.

Happily, when we set the Info in the request context, we perform a shallow clone of the Request and pass it down to further handlers. This means that we can set an empty value when the request is received and rely on it being correct when we later retrieve it because it is set higher in the context tree (this behaviour is also relied upon by mux for HTTP vars).

### DIFF
--- a/lxd/daemon.go
+++ b/lxd/daemon.go
@@ -817,8 +817,8 @@ func (d *Daemon) createCmd(restAPI *mux.Router, version string, c APIEndpoint) {
 			return
 		}
 
-		// Ensure request context info is set.
-		reqInfo := request.SetupContextInfo(r)
+		// Initialise the request info.
+		reqInfo := request.InitContextInfo(r)
 
 		// Set the "trusted" value in the request context.
 		reqInfo.Trusted = trusted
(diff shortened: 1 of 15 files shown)


MusicDin:
Nice catch! 🚀 Can we add a test for this? Maybe `test/lxd-client` could be used?

tomponline (dismissed):
Thanks!

tomponline:
Unit tests failing

markylaing:
> Unit tests failing

This was caused by initialising the request info as empty in the devlxd handler, which caused the `net.Conn` to be missing from the info struct.

I've reverted some of @MusicDin's commits so that we again have a separate `CtxConn` context key containing the `net.Conn`. We should keep this separate to keep it self contained and because it is used by the `SaveConnectionInContext` function that we use as the `ConnContext` field when setting up an `http.Server`.

tomponline:
> > Unit tests failing
> 
> This was caused by initialising the request info as empty in the devlxd handler, which caused the `net.Conn` to be missing from the info struct.
> 
> I've reverted some of @MusicDin's commits so that we again have a separate `CtxConn` context key containing the `net.Conn`. We should keep this separate to keep it self contained and because it is used by the `SaveConnectionInContext` function that we use as the `ConnContext` field when setting up an `http.Server`.

Yes sounds good to separate per connection and per request data.

PR #16040, opened 2025-07-16
lxd: Instance device removal using instance patch

MusicDin:
Allow instance device removal using `PATCH /1.0/instances/{name}` by setting device value to `null`.

For example:
```
lxc query -X PATCH /1.0/instances/c1 -d '{"devices": {"disk01": null}}' 
```


lxd/instance_patch.go
```diff
@@ -175,12 +175,21 @@ func instancePatch(d *Daemon, r *http.Request) response.Response {
 	if req.Devices == nil {
 		req.Devices = c.LocalDevices().CloneNative()
 	} else {
+		// Retain devices that are not present in the request.
 		for k, v := range c.LocalDevices() {
 			_, ok := req.Devices[k]
```
tomponline, line 180: Can we do this instead of another loop?

```go
or k, v := range c.LocalDevices() {
			rv, ok := req.Devices[k]
            if !ok {
				req.Devices[k] = v
			} else if ok && rv == nil {
                delete(req.Devices, k)
            }
```

MusicDin: Depends what is the expected behavior when trying to remove a non-existing device?
Currently, if the device does not exist, it is a no-op.

In your case, if the device does not exist it will add the new device and later fail the device validation.

tomponline: Ah I see OK, lets leave as is then.

test/suites/container_devices_disk.sh
```diff
@@ -184,3 +185,28 @@ _container_devices_disk_char() {
   lxc config device remove foo char
   lxc stop foo -f
 }
+
+_container_devices_disk_patch() {
+  lxc launch testimage c1 || true
```
tomponline: Why || true?

tomponline: Why do we need a running instance at all, can we use an empty one?

MusicDin: > Why || true?

No need, left over from my local testing. Fixed.

> Why do we need a running instance at all, can we use an empty one?

Fixed.

test/suites/container_devices_disk.sh
```diff
@@ -184,3 +185,28 @@ _container_devices_disk_char() {
   lxc config device remove foo char
   lxc stop foo -f
 }
+
+_container_devices_disk_patch() {
+  lxc init testimage c1
```
tomponline: --empty no need for an image to be used

MusicDin: Ahh, right. Fixed.

PR #16047, opened 2025-07-17
Container: Check for monitor process candidate's `NSpid` value  (from Incus)

tomponline:
When a process inside a container accesses `/dev/lxd` LXD will now check the `NSpid` status variable of each candidate monitor process to confirm the process has no PIDs in other namespaces. If it does then it can't be the LXC monitor process, even if it has "[lxc monitor]" in its process name.

Inspired by ideas from https://github.com/lxc/incus/pull/2285


lxd/api_devlxd.go
```diff
@@ -253,48 +309,33 @@ func devlxdFindContainerForPID(s *state.State, originPID int32) (instance.Contai
...
-			pid = int32(ppid)
-			break
-		}
+		// Don't return error to avoid leaking details about the process.
+		return nil, errPIDNotInContainer
 	}
 
 	// Fallback to loading all containers and checking their PID namespaces.
+	// This is the approach used when the process isn't actually a descendant of the container's init process,
+	// such as when using `lxc exec`.
```
roosterfish, line 302: How are the following lines related to the container's init process, isn't `instance.LoadNodeAll` just running a SQL query? Probably I cannot follow the comment here.

roosterfish: Ah, should the comment rather be on the loop below?

roosterfish (commented):
LGTM, only a small comment for clarification.

lxd/api_devlxd.go
tomponline: No, the comment is in the right place.

Its say that we use this fallback approach when the origin process of the devlxd request isn't actually a descendant of the container's init process (which is what happens when accessing /dev/lxd from a shell launched with `lxc exec`).

roosterfish: Ah I see it's just a header for the entire block. Thanks.

roosterfish (approved):

PR #16049, opened 2025-07-17
VM: Disk device hotplug improvements and fixes

tomponline:
- Moves waiting for PCI bridge population into the hotplug device add function rather than doing it at VM start time - this maintains the fix that allows adding custom directory disk devices immediately after VM start but should remove the flakiness of the previous implementation.
- Custom disk directory devices now use their own dedicated PCIe slot at boot time - this fixes an issue when removing custom disk devices that were also removing the config drive devices. This way it makes pre-boot slot allocation the same as hotplug slot allocation.
- Fixes panic for `virtio-blk` `io.bus` panic.

### DIFF
--- a/lxd/instance/drivers/driver_qemu.go
+++ b/lxd/instance/drivers/driver_qemu.go
@@ -3949,13 +3949,23 @@ func (d *qemu) addDriveDirConfig(cfg *[]cfgSection, bus *qemuBus, fdFiles *[]*os
 		}
 	}
 
+	// Setup a bus allocator for use with generating QEMU pre-boot config file.
+	busAllocate := func() (busName string, busAddress string, multi bool, err error) {
+		// Use per-LXD device multi-function group so that both virtiofs and 9p devices use same bus slot.
+		busName, busAddr, multi := bus.allocate("lxd_" + driveConf.DevName)
+		return busName, busAddr, multi, nil
+	}
+
 	// If there is a virtiofsd socket path setup the virtio-fs share.
 	if virtiofsdSockPath != "" {
 		if !shared.PathExists(virtiofsdSockPath) {
 			return fmt.Errorf("virtiofsd socket path %q doesn't exist", virtiofsdSockPath)
 		}
 
-		devBus, devAddr, multi := bus.allocate(busFunctionGroup9p)
+		devBus, devAddr, multi, err := busAllocate()
+		if err != nil {
+			return err
+		}
 
 		shortPath, err := d.shortenedFilePath(virtiofsdSockPath, fdFiles)
 		if err != nil {
@@ -3979,7 +3989,10 @@ func (d *qemu) addDriveDirConfig(cfg *[]cfgSection, bus *qemuBus, fdFiles *[]*os
 	}
 
 	// Add 9p share config.
-	devBus, devAddr, multi := bus.allocate(busFunctionGroup9p)
+	devBus, devAddr, multi, err := busAllocate()
+	if err != nil {
+		return err
+	}
 
 	fdSource, ok := driveConf.DevSource.(deviceConfig.DevSourceFD)
 	if !ok {


lxd/instance/drivers/driver_qemu.go
```diff
@@ -4243,20 +4217,17 @@ func (d *qemu) addDriveConfig(qemuDev map[string]any, bootIndexes map[string]int
...
+	qemuDev := make(map[string]any)
 	qemuDev["id"] = qemuDeviceIDPrefix + escapedDeviceName
 	qemuDevDrive, ok := blockDev["node-name"].(string)
 	if !ok {
 		return nil, errors.New("Failed getting block device node-name")
 	}
 
 	qemuDev["drive"] = qemuDevDrive
-
+	qemuDev["driver"] = busName
```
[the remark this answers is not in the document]
tomponline: fixed

simondeziel:
>     * Custom disk directory devices now use their own dedicated PCIe slot at boot time - this fixes an issue when removing custom disk devices that were also removing the config drive devices. This way it makes pre-boot slot allocation the same as hotplug slot allocation.

I'm far from an expert with PCIe root ports, switches/bridges and multifunction devices concepts but I think there is a fairly low limit on dedicated PCIe slots (aka root ports?), right?

Would it make sense to dedicate a PCIe bridge to shared disk directories? Or is the added complexity not justified?

tomponline:
> > ```
> > * Custom disk directory devices now use their own dedicated PCIe slot at boot time - this fixes an issue when removing custom disk devices that were also removing the config drive devices. This way it makes pre-boot slot allocation the same as hotplug slot allocation.
> > ```
> 
> I'm far from an expert with PCIe root ports, switches/bridges and multifunction devices concepts but I think there is a fairly low limit on dedicated PCIe slots (aka root ports?), right?
> 
> Would it make sense to dedicate a PCIe bridge to shared disk directories? Or is the added complexity not justified?

Too big a change without research. This change just makes boot time setup the same as hotplug setup.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -4046,7 +4020,17 @@ func (d *qemu) addDriveDirConfig(cfg *[]cfgSection, bus *qemuBus, fdFiles *[]*os
 }
 
 // addDriveConfig adds the qemu config required for adding a supplementary drive.
-func (d *qemu) addDriveConfig(qemuDev map[string]any, bootIndexes map[string]int, driveConf deviceConfig.MountEntryItem) (monitorHook, error) {
+func (d *qemu) addDriveConfig(busAllocate busAllocator, bootIndexes map[string]int, driveConf deviceConfig.MountEntryItem) (monitorHook, error) {
+	// Check if the user has overridden the bus.
+	busName := "virtio-scsi"
+	for _, opt := range driveConf.Opts {
+		name, found := strings.CutPrefix(opt, "bus=")
```
simondeziel, line 4027: For consistency, the loop below this to handle `cache=` might also be switched to `strings.CutPrefix()`.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -4279,21 +4248,24 @@ func (d *qemu) addDriveConfig(qemuDev map[string]any, bootIndexes map[string]int
...
-		qemuDev["driver"] = bus
+		d.logger.Debug("Using device port to plug drive into", logger.Ctx{"device": driveConf.DevName, "port": devBus})
+
+		// Populate the qemu device with port info.
+		qemuDev["bus"] = devBus
+		qemuDev["addr"] = devAddr
+
+		if multi {
+			qemuDev["multifunction"] = true
+		}
```
simondeziel, line 4268:
```suggestion
		qemuDev["multifunction"] = multi
```

`true` will render to `on` and `false` will simply omit the config line.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -2503,36 +2453,66 @@ func (d *qemu) deviceAttachNIC(deviceName string, netIF []deviceConfig.RunConfig
...
+		if !hasUsedBridges(devices) {
+			// Wait for PCI bridges to be created (with the internal devices LXD adds) to ensure QEMU
+			// has had time to initialise the QMP query-pci output.
+			// Otherwise we'll end up trying to hotplug into a used PCI slot.
+			d.logger.Debug("Waiting for PCI to initialize")
+
+			select {
+			case <-ctx.Done():
+				return "", "", false, fmt.Errorf("Timed out waiting for PCI to initialize: %w", ctx.Err())
+			case <-time.After(500 * time.Millisecond):
```
simondeziel, line 2493: If the first iteration barely misses it, it adds a long delay to the boot. Would it be OK to burn a bit more CPU to more actively poll here and hopefully break out earlier?

lxd/instance/drivers/driver_qemu.go
```diff
@@ -2503,36 +2453,66 @@ func (d *qemu) deviceAttachNIC(deviceName string, netIF []deviceConfig.RunConfig
...
+			}
 
-	return "", errors.New("No available PCI hotplug slots could be found")
+			// Skip used bridges.
+			if len(dev.Bridge.Devices) > 0 {
+				continue
+			}
+
+			// Found an empty slot.
+			return dev.DevID, "00.0", false, nil
```
simondeziel, line 2511: I know this `00.0` was just moved but a brief comment explaining what it refers to would be nice. Does it indicate it uses a dedicated port and not one behind a bridge?

simondeziel (approved):
LGTM

tomponline:
@simondeziel ta, I've got a follow up pr in the works for a different related hotplug issue so will address ur feedback in that.

tomponline:
> > ```
> > * Custom disk directory devices now use their own dedicated PCIe slot at boot time - this fixes an issue when removing custom disk devices that were also removing the config drive devices. This way it makes pre-boot slot allocation the same as hotplug slot allocation.
> > ```
> 
> I'm far from an expert with PCIe root ports, switches/bridges and multifunction devices concepts but I think there is a fairly low limit on dedicated PCIe slots (aka root ports?), right?
> 
> Would it make sense to dedicate a PCIe bridge to shared disk directories? Or is the added complexity not justified?

https://gitlab.com/qemu-project/qemu/-/blob/master/docs/pcie.txt?ref_type=heads is useful on this topic.

But using a single dedicated root port is a non starter as only 8 functions.

So we would also need to figure out how to avoid the issue fixed in this pr (removing a single function not the whole port when hot unplugging) and opportunistically filling partially filled virtiofs/9p in use ports when hot plugging.

I've got an idea around dynamically shrinking the number of empty ports we add at startup to accommodate more devices at start up (sacrificing hotplug capacity).

lxd/instance/drivers/driver_qemu.go
tomponline: https://github.com/canonical/lxd/pull/16067/

lxd/instance/drivers/driver_qemu.go
tomponline: https://github.com/canonical/lxd/pull/16067/

lxd/instance/drivers/driver_qemu.go
tomponline: https://github.com/canonical/lxd/pull/16067/

lxd/instance/drivers/driver_qemu.go
tomponline: https://github.com/canonical/lxd/pull/16049

PR #16050, opened 2025-07-17
github: Increase allowed `lxc` binary size (16MiB->17MiB)

kadinsayani:
This is required for cluster links and replicas. With the addition of the new commands we're exceeding the current limit: https://github.com/canonical/lxd/actions/runs/16353437660/job/46206281586?pr=14884.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -202,7 +202,7 @@ jobs:
           set -eux
 
           # bin/max (sizes are in MiB)
-          SIZES="lxc 16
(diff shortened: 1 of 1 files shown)


simondeziel:
@kadinsayani as you probably figured out already, this change needs to happen in the same PR that will lead to a bigger binary.

kadinsayani:
> @kadinsayani as you probably figured out already, this change needs to happen in the same PR that will lead to a bigger binary.

Thanks, I've included this change with #14884.

Issue #12716, opened 2024-01-09
Removing a `disk` device with an existing target path on a VM leads to unreachable target directory 


gabrielmougard:
# Required information
Distribution: ubuntu
Distribution version: 22.04
The output of "lxc info" or if that fails:
Kernel version:
LXC version:
LXD version: 5.19-8635f82
Storage backend in use: zfs

# Issue description
Similarly to https://github.com/canonical/lxd/issues/12648, but for a VM instance this time, we observe an issue while removing a disk device that was bind mounted over an existing empty directory. This time, the underlying directory doesn't get removed but cannot but interacted with :

```bash
mkdir /tmp/empty-dir
lxc launch ubuntu:jammy v1 --vm
lxc config device add v1 test disk source=/tmp/empty-dir path=/opt
lxc config device remove v1 test
lxc exec v1 -- ls /opt
ls: cannot access '/opt': Transport endpoint is not connected
```

I would expect that, for an already exiting directory, the target directory would be unmounted but still present, whereas a target directory created through LXD during the mount woulld be unmounted and deleted when unmounted. 

We should also check that, within the VM, the target path we're attempting to unmount has no overmounts as calling `umount` on the target directory will first unmount the last mount but not all of them.

gabrielmougard:
@tomponline I opened a new issue as this is a bit different than https://github.com/canonical/lxd/issues/12648

tomponline:
@gabrielmougard thanks, I suspect this is due to:

https://discuss.linuxcontainers.org/t/incus-0-3-has-been-released/18351#hot-plughot-remove-of-paths-in-virtual-machines-5

> Worth noting that the agent will not automatically unmount the filesystem prior to hot-remove. If the filesystem is mounted in the guest, you’ll get an error during hot-remove.

tomponline:
Thanks for opening this @gabrielmougard I suspect we'll need to add support to the lxd-agent to trigger an unmount when the device is being removed.

PR #12700, opened 2024-01-05
Device: Don't remove an existing target directory when unmounting a `disk` device if the original dir hasn't been created by LXD 

gabrielmougard:
closes #12648 
closes #12716
____

# Overall description

When a disk device is removed while relying on a host directory and mapped to a target within a container or a VM,
we detect if the target directory has been created by LXD or not in order to not delete the content of a target
directory during the unmount. Additionaly, with the VM case, we cleanly unmount the target path inside the VM.  

# Container case

Example:

- Let say we mounted a custom host empty directory (`test`) on the existing `/opt` directory of the container,
when unmounted (`lxc device remove ...`), the target `/opt` won't be removed, because we marked it as NOT being created
by LXD at mount time.

- If, however, we created a custom empty target directory at mount time:
`lxc config device add u1 test disk source=/home/user/test path=/new_dir`,
the directory `new_dir` will be created on the target instance and if we decide to unmount `test`,
the target `/new_dir` will be removed because is has been created by LXD

# Particular case for VMs

In addition to that, this fix also cleanly unmount the target directory inside the VM through an new LXD-agent API call. Before that, here is what would have happened:

```bash
mkdir /tmp/empty-dir
lxc launch ubuntu:jammy v1 --vm
lxc config device add v1 empty-dir disk source=/tmp/empty-dir path=/opt
lxc config device remove v1 empty-dir

lxc shell v1 -- stat /opt
stat: cannot statx '/opt': Transport endpoint is not connected
``` 

This happens because the mounted device and its associated char device were removed using QEMU's QMP without unmounting the target.

Now, we inspect for the mounts and the over-mounts if any on the VM, and unmount them in the right order. 

## Benchmark

Each time, there are 10 runs on each benchmark case (10 starts, 10 stops)

`master` branch (CONTAINER):

Startup time with:                   Stop time (`--force`) with:

* 0 disk:    `598.2 ms ±  14.3 ms`   * 0 disk:    `883.5 ms ±  37.1 ms`
* 5 disk:    `595.0 ms ±   8.2 ms`   * 5 disk:    `892.6 ms ±  23.3 ms`
* 10 disks:  `601.0 ms ±   8.5 ms`   * 10 disks:  `900.1 ms ±  22.0 ms`
* 20 disks:  `606.9 ms ±  12.7 ms`   * 20 disks:  `886.8 ms ±  23.4 ms`
* 100 disks: `634.3 ms ±  10.1 ms`   * 100 disks: `934.1 ms ±  16.4 ms`

[...]

PR #16051, opened 2025-07-17
Device: Clean up disk device removal from VM (WIP)

nmezhenskyi:
Fixes https://github.com/canonical/lxd/issues/12716.  

Includes cherry-picks from https://github.com/canonical/lxd/pull/12700.


tomponline:
@nmezhenskyi as discussed, lets see if we can depend on the "remove" event?

lxd-agent/events.go
```diff
@@ -132,63 +136,103 @@ func eventsProcess(event api.Event) {
...
 	}
 
-	// Only care about device additions, we don't try to handle remove.
-	if e.Action != "added" {
-		return
-	}
-
-	// We only handle disk hotplug.
+	// We only handle disk hotplug/removal.
 	if !filters.IsDisk(e.Config) {
```
tomponline, line 145: can we also add an explicit check for the e.Action too please

tomponline: Also would be great if you can define contants for the e.Action values in the lxd-agent package and then use then instead of hardcoded values in LXD where the events are sent.

This way correlating the event sender and receiver is easier.

nmezhenskyi: Yes, great idea, thanks

nmezhenskyi: I've added constants to `lxd-agent/api` package because this is what we import in LXD to interface with the lxd-agent.

lxd-agent/events.go
```diff
@@ -132,63 +137,108 @@ func eventsProcess(event api.Event) {
...
-		time.Sleep(500 * time.Millisecond)
-	}
+		defer mountInfoFile.Close()
+
+		var mountPoints []string
+		scanner := bufio.NewScanner(mountInfoFile)
+
+		for scanner.Scan() {
+			fields := strings.Fields(scanner.Text())
+			if len(fields) >= 10 && strings.HasPrefix(fields[4], strings.TrimSuffix(targetPath, "/")) {
```
simondeziel: Could there be a problem if a user mounts a directory as well as another dir that's a child of the first? In other words, should we do prefix matching or exact matching?

I agree with the `strings.TrimSuffix(targetPath, "/")` part but I think it should be done outside of the loop.

nmezhenskyi: > Could there be a problem if a user mounts a directory as well as another dir that's a child of the first? In other words, should we do prefix matching or exact matching?

Yes, there is actually still a problem with unmounting the parent in this case. I'll fix the condition.

Or are you saying that the user might not expect the child to be unmounted as a result of the parent unmount?

> I agree with the strings.TrimSuffix(targetPath, "/") part but I think it should be done outside of the loop.

Good point, I'll do that.

lxd-agent/events.go
```diff
@@ -132,63 +137,108 @@ func eventsProcess(event api.Event) {
...
 
-	l.Info("Failed to mount hotplug", logger.Ctx{"err": err})
+		err = scanner.Err()
+		if err != nil {
+			l.Error("Error reading /proc/self/mountinfo", logger.Ctx{"err": err})
+			return
+		}
+
+		if len(mountPoints) == 0 {
+			l.Error("No mount points found for " + targetPath)
```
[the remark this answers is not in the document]
nmezhenskyi: Done

lxd-agent/events.go
```diff
@@ -132,63 +137,108 @@ func eventsProcess(event api.Event) {
...
+		}
+
+		// Reverse the slice to unmount in reverse order.
+		// This is needed to unmount potential over-mounts first.
+		slices.Reverse(mountPoints)
+
+		for _, mountPoint := range mountPoints {
+			err = unix.Unmount(mountPoint, unix.MNT_DETACH)
+			if err != nil {
+				l.Error("Error unmounting "+mountPoint, logger.Ctx{"err": err})
```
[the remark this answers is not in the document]
nmezhenskyi: Done

lxd-agent/events.go
nmezhenskyi: I fixed the condition. Now it should match the exact target path and nested mounts, without unintended mounts that have similar path name to the target path.

lxd-agent/events.go
```diff
@@ -132,63 +137,111 @@ func eventsProcess(event api.Event) {
...
+			l.Error("No mount points found")
+			return
+		}
+
+		// Reverse the slice to unmount in reverse order.
+		// This is needed to unmount potential over-mounts first.
+		slices.Reverse(mountPoints)
+
+		for _, mountPoint := range mountPoints {
+			err = unix.Unmount(mountPoint, unix.MNT_DETACH)
```
nmezhenskyi: The call to `unix.Unmount(mountPoint, unix.MNT_DETACH)` fails for all nested mounts, even when using `MNT_FORCE` flag, with the error: `transport endpoint is not connected`.

I think this is because the parent disk device is already removed from the VM at that point.

tomponline: Yes, not much we can do about that - but probably quite a rare scenario.

lxd-agent/events.go
```diff
@@ -132,63 +137,111 @@ func eventsProcess(event api.Event) {
...
 		}
 
-		time.Sleep(500 * time.Millisecond)
-	}
+		defer mountInfoFile.Close()
+
+		var mountPoints []string
+		scanner := bufio.NewScanner(mountInfoFile)
+
+		trimmedPath := strings.TrimSuffix(targetPath, "/")
```
simondeziel, line 212: Assuming I understood what's going on properly, adding a `nestedParentPath := trimmedPath+"/"` here would make the `HasPrefix` bit easier to understand.

lxd-agent/events.go
nmezhenskyi: True, maybe in that case we should skip trying to unmount nested mounts completely? We can try to unmount the target path only and let it fail if there are any nested mounts. What do you think? @tomponline

nmezhenskyi: Alternatively, we could try to do some kind of validation before device removal and notify the user if there are any nested mounts present. This will likely need an additional HTTP endpoint on the `lxd-agent`'s side.

tomponline: > We can try to unmount the target path only and let it fail if there are any nested mounts.

sounds good for now.

lxd-agent/events.go
nmezhenskyi: I removed the part with `HasPrefix` completely. We agreed with Tom to unmount the target path only for now and let it fail if there are any nested mounts.

simondeziel (approved):

tomponline:
@nmezhenskyi when ready please mark as ready to review. Ta

nmezhenskyi:
VM tests from `lxd-ci` are passing locally.

tomponline (approved):
LGTM

PR #16052, opened 2025-07-18
Cache `apparmor_parser --version` info during daemon init

simondeziel:
Also address a couple CodeQL alerts.

### DIFF
--- a/lxd/sys/os.go
+++ b/lxd/sys/os.go
@@ -102,10 +102,13 @@ type OS struct {
 	LXCFeatures map[string]bool
 
 	// OS info
-	ReleaseInfo   map[string]string
-	KernelVersion version.DottedVersion
-	Uname         *shared.Utsname
-	BootTime      time.Time
+	ReleaseInfo map[string]string
+	Uname       *shared.Utsname
+	BootTime    time.Time
+
+	// Version info
+	KernelVersion   version.DottedVersion
+	AppArmorVersion *version.DottedVersion
 }
 
 // DefaultOS returns a fresh uninitialized OS instance with default values.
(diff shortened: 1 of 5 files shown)


lxd/sys/os.go
```diff
@@ -102,10 +102,13 @@ type OS struct {
...
-	KernelVersion version.DottedVersion
-	Uname         *shared.Utsname
-	BootTime      time.Time
+	ReleaseInfo map[string]string
+	Uname       *shared.Utsname
+	BootTime    time.Time
+
+	// Version info
+	KernelVersion   version.DottedVersion
+	AppArmorVersion *version.DottedVersion
```
tomponline, line 113: There's an AppArmor section of this struct this seems to make sense to go into. Also why is this a pointer? Is this so it can be optional? Can we make it 0 if it's not present, there's already an AppArmorAvailable field in this struct too.

lxd/sys/os.go
```diff
@@ -77,6 +77,8 @@ type OS struct {
 	AppArmorStacked   bool
 	AppArmorStacking  bool
 	AppArmorFeatures  AppArmorFeaturesInfo
+	AppArmorCacheLoc  string
```
tomponline, line 80: "Loc"? Can this be "Path"?

tomponline (approved):

lxd/sys/os.go
tomponline: As this is internal struct its fine as is

lxd/sys/os.go
tomponline: As this is internal struct its fine as is

simondeziel: I went with `CacheLoc` to match the `apparmor_parser` argument of `--cache-loc`.

lxd/sys/os.go
simondeziel: I went for a pointer as that's what `version.Parse()` returns and it's used in `appArmorGetVersion()`.

I can change it if you'd like.

PR #16054, opened 2025-07-18
tests: Run Tiobe TICS on self-hosted runners

MusicDin:
Run Tiobe TICS on self-hosted runners.


tomponline:
@MusicDin probably easiest to merge the single line change and then see if it works ok on next schedule run

tomponline (approved):

PR #16057, opened 2025-07-20
Revert "tests: Run Tiobe TICS on self-hosted runners"

MusicDin:
Reverts canonical/lxd#16054

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -414,7 +414,7 @@ jobs:
 
   tics:
     name: Tiobe TICS
-    runs-on: tiobe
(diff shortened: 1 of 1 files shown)


tomponline (approved):

roosterfish:
I was about to make this change for Micro*. What was the reason for it being reverted?

MusicDin:
The self-hosted runner was assigned successfully, but the step failed with no clear output what was wrong. Based on the reported logs it hanged on some operation (or more likely did not report the actual failure). We will retry the switch, it was just not the best timing due to the upcoming release.

Issue #15644, opened 2025-05-23
Cluster restore doesn't wait for networks and storage pools


roosterfish:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

22.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
.
```

### Output of "lxc info" or system info if it fails

```shell
.
```

### Issue description

When restoring a cluster member using `lxc cluster restore`, the daemon won't wait until both the networks and storage pools are ready.

Storage pools are always started when the daemon comes up regardless whether or not the member's state is evacuated.
However networks are started as part of the restore operation. In case the first try to start a network fails, LXD spawns a background routine which retries every 60s. 
But this will cause any attempts to restart instances on those networks/pools to return an error as they aren't ready yet.

Running the member restore operation should wait until both the networks and storage pools are ready to prevent running into errors when restoring instances on this member.

### Proposed solution

Add a new generic `wait`/`waitready` flag to the `lxc cluster restore` command which indicates to wait for both networks and storage pools. A timeout can be provided too. To stay backwards compatible if the flag isn't provided LXD won't wait for the resources to be ready.

### Steps to reproduce

1. Evacuate a member `foo` with instances
2. Break `foo`'s storage pool/network on this evacuated member
3. `lxc cluster restore foo` <- Errors out with something equal to `Error: Failed to wait for migration to finish: Instance move to destination failed on source: Failed migration on source: Error from migration control target: Failed creating instance on target: Storage pool is unavailable on this server`

### Information to attach

- [ ] Any relevant kernel output (`dmesg`)
- [ ] Instance log (`lxc info NAME --show-log`)
- [ ] Instance configuration (`lxc config show NAME --expanded`)
- [ ] Main daemon log (at `/var/log/lxd/lxd.log` or `/var/snap/lxd/common/lxd/logs/lxd.log`)
- [ ] Output of the client with `--debug`
- [ ] Output of the daemon with `--debug` (or use `lxc monitor` while reproducing the issue)

tomponline:
> To stay backwards compatible if the flag isn't provided LXD won't wait for the resources to be ready.

Or we default to a sensible timeout at the CLI level so as not to change the API behaviour but introduce useful defaults for those using the CLI.

tomponline:
[...]

PR #16066, opened 2025-07-21
Cluster: Add pre-checks for evacuation/restore

roosterfish:
Fixes https://github.com/canonical/lxd/issues/15644

The PR leverages the latest enhancements to the local waitready infrastructure allowing to wait until all networks and storage pools are started before performing relevant cluster member operations.

With the changes in this PR a cluster member can only ever be evacuated/restored when all of its networks and storage pools are ready.
This is effectively a safety mechanism so that you don't accidentally restore cluster members which might cause instances to be moved over that depend on non-ready networks and storage pools which will lead to errors.


lxd/api_cluster.go
```diff
@@ -3074,6 +3074,20 @@ func clusterNodeStatePost(d *Daemon, r *http.Request) response.Response {
 		return response.BadRequest(err)
 	}
 
+	// Run some pre-checks before evacuating or restoring the cluster member.
+	// It's important that those checks runs on the to be restored cluster member.
+	select {
+	case <-d.State().NetworkReady.Done():
```
tomponline: If you're not waiting for a context, its clearer in my view to do a non-blocking call, and avoid the select, so:

```go
if d.State().NetworkReady.Err() == nil {
    return response.BadRequest(fmt.Errorf("Cannot %s %q because some networks aren't started yet", req.Action, d.serverName))
}
```

lxd/api_cluster.go
```diff
@@ -3074,6 +3074,20 @@ func clusterNodeStatePost(d *Daemon, r *http.Request) response.Response {
...
+	default:
+		return response.BadRequest(fmt.Errorf("Cannot %s %q because some networks aren't started yet", req.Action, d.serverName))
+	}
+
+	select {
+	case <-d.State().StorageReady.Done():
```
tomponline: Can we avoid multiple calls to State() in a single function please.

tomponline: Its already loaded here https://github.com/canonical/lxd/pull/16066/commits/b06e929f5816e224a27b1cb922081e198e737b23#diff-07fccfa2de6bd472d27c84cb9ca8cd9319e32135864bf114267111df3d3fad2fR3062

lxd/storage.go
```diff
@@ -62,6 +62,11 @@ func storageStartup(s *state.State) error {
 	if err != nil {
 		if response.IsNotFoundError(err) {
 			logger.Debug("No existing storage pools detected")
+
+			// There aren't any storage pools.
+			// Unblock all the waitready callers using the --storage flag.
+			s.StorageReady.Cancel()
```
tomponline, line 68: Do we need the same for networks?

lxd/api_cluster.go
roosterfish: True, also ways less lines. Fixed.

lxd/api_cluster.go
roosterfish: Haven't seen this one, fixed.

lxd/storage.go
roosterfish: The networks are looped a bit different which causes the canceller to be cancelled in case all networks are started or if there aren't any networks. For clarity I have added a comment to the respective line in the preceding commit.

For full test coverage I have now extended the last commit which adds checks to the test suite to check both networks and storage pools when trying cluster member evacuation. Now both evacuate and restore are checked for both networks and storage pools.

tomponline (approved):
LGTM thanks!

Issue #16003, opened 2025-07-14
Make LXD concurrent device add requests proceed sequentially


taurus-forever:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

24.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
> snap list --all lxd core20 core22 core24 snapd
Name    Version         Rev    Tracking       Publisher   Notes
core20  20250526        2599   latest/stable  canonical✓  base
core22  20250612        2045   latest/stable  canonical✓  base
core24  20250526        1006   latest/stable  canonical✓  base
lxd     5.21.3-c5ae129  33110  5.21/stable    canonical✓  -
snapd   2.70            24792  latest/stable  canonical✓  snapd
```

### Output of "lxc info" or system info if it fails

```shell
> lxc info
config:
  core.https_address: '[::]'
api_extensions:
- storage_zfs_remove_snapshots
- container_host_shutdown_timeout
- container_stop_priority
- container_syscall_filtering
- auth_pki
- container_last_used_at
- etag
- patch
- usb_devices
- https_allowed_credentials
- image_compression_algorithm
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
- unix_device_rename
- storage_lvm_use_thinpool
- storage_rsync_bwlimit
- network_vxlan_interface
- storage_btrfs_mount_options
- entity_description
- image_force_refresh
- storage_lvm_lv_resizing
- id_map_base
- file_symlinks
- container_push_target
- network_vlan_physical
- storage_images_delete
- container_edit_metadata
- container_snapshot_stateful_migration
- storage_driver_ceph
- storage_ceph_user_name
- resource_limits
- storage_volatile_initial_source
- storage_ceph_force_osd_reuse
- storage_block_filesystem_btrfs
- resources
- kernel_limits
- storage_api_volume_rename
- network_sriov
[...]

PR #16067, opened 2025-07-21
VM: Re-generate agent-mounts.json file when directory disks are hotplugged

tomponline:
Fixes 2nd issue reported in https://github.com/canonical/lxd/issues/16003 which was that when hotplugging a directory disk immediately after VM start (but before the lxd-agent had loaded) caused the devlxd notification event to be missed by the agent and so the directory disk share was not mounted inside the guest.

E.g.
```
lxc start v1; lxc config device add v1 dir1 disk source=/mnt path=/mnt
```

This PR fixes that by re-writing the agent-mounts.json file in the VM's config drive share when a directory disk is hotplugged, so that its mount config is available to the lxd-agent when it eventually starts.

Also addresses some of @simondeziel comments from https://github.com/canonical/lxd/pull/16049

### DIFF
--- a/lxd/instance/drivers/driver_qemu.go
+++ b/lxd/instance/drivers/driver_qemu.go
@@ -4229,10 +4229,7 @@ func (d *qemu) addDriveConfig(busAllocate busAllocator, bootIndexes map[string]i
 		// Populate the qemu device with port info.
 		qemuDev["bus"] = devBus
 		qemuDev["addr"] = devAddr
-
-		if multi {
-			qemuDev["multifunction"] = true
-		}
+		qemuDev["multifunction"] = multi
 	}
 
 	if bootIndexes != nil {


lxd/instance/drivers/driver_qemu.go
```diff
@@ -5833,6 +5801,15 @@ func (d *qemu) Update(args db.InstanceArgs, userRequested bool) error {
 	cpuLimitWasChanged := false
 
 	if isRunning {
```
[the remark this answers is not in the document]
tomponline: As this mounts config file potentially needs updating when devices are added, removed or re-ordered (to account for correct mount path ordering) it is simpler and safer to always evaluate the resulting device set and then compare if the on-disk json file has changed and in that case only write out the new config if needed.

This reduces write I/O at least.

The file itself is small and so the read I/O is minimal.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -5801,6 +5801,15 @@ func (d *qemu) Update(args db.InstanceArgs, userRequested bool) error {
...
+		// Re-generate the agent mounts file so that it reflects the current devices set.
+		// This way if a directory disk is added immediately after VM start but before the lxd-agent has
+		// started in the guest (such that it misses the devlxd notification event), the agent will still
+		// be able to see the mount config for the new disk when it starts.
+		err = d.generateAgentMountsFile()
```
roosterfish, line 5804: Does the agent load the mounts file after starting its event listener? Or is there still a small window in which you could potentially add a device which isn't observed by the agent?

1. agent starts and loads the mounts file
2. new device gets added
3. agent is about to start the listener so it misses the event and has already loaded the previous version of the mounts file
4. agent is mounting all devices except the one from (2)

tomponline: Yes there is a small window where this can occur as lxd-agent doesn't report to LXD its online and start its HTTP server until after the mount file has been read.

However changing that ordering would require careful evaluation and testing of the impacts of changing when the instance's status (such as process count) is reported during guest OS start up. As many of our tests use a non-zero process count reported by the agent to mean the agent has done its work.

We may be able to move `c.mountHostShares()` after `startHTTPServer()`, but before `c.startStatusNotifier()` in the lxd-agent, but would require more testing and consideration. The `c.startStatusNotifier()` notifies LXD that its online and this triggers LXD to start sending it events. However if we call `c.startStatusNotifier()` before `c.mountHostShares()`  then we've potentially introduced another race whereby a non-zero process count will be reported by `/1.0/instances/<name>/state` (indicating the agent is started) but the start-time disks haven't been mounted yet.
 
This PR fixes the immediate issue with Juju which is that it starts a VM and then immediately adds directory disks to the instance's config.

roosterfish: I see, thanks for providing the additional details. Let's not over complicate the fix then as it already fixes most of the not previously covered window in which new devices could be added to the VM.

PR #16071, opened 2025-07-22
Backport waitready enhancements (stable-5.21)

roosterfish:
Mainly backports #16035 and #16066 but also some other PRs/commits to retain a clean history:

* https://github.com/canonical/lxd/pull/15926
* https://github.com/canonical/lxd/pull/15934
* https://github.com/canonical/lxd/commit/158073c54a73e61e1b48d9dfd04f44cea155f4e9 (#15820)
* https://github.com/canonical/lxd/commit/ebfb168bf63b599b5135d9fa211ad3aa68718921 (#15820)
* https://github.com/canonical/lxd/pull/16035
* https://github.com/canonical/lxd/pull/16066

This allows making use of the new features in the MicroCloud test suite which pulls LXD `5.21/edge`.

### DIFF
--- a/lxd/api_cluster.go
+++ b/lxd/api_cluster.go
@@ -3323,7 +3323,7 @@ func evacuateClusterMember(s *state.State, gateway *cluster.Gateway, r *http.Req
 		}
 
 		// Stop networks after evacuation.
-		networkingStop(s)
+		networkStop(s)
 
 		revert.Success()
 		return nil
--- a/lxd/api_cluster.go
+++ b/lxd/api_cluster.go
@@ -3322,8 +3322,8 @@ func evacuateClusterMember(s *state.State, gateway *cluster.Gateway, r *http.Req
 			return err
 		}
 
-		// Stop networks after evacuation.
-		networkStop(s)
+		// Evacuate networks too.
+		networkStop(s, true)
 
 		revert.Success()
 		return nil
--- a/lxd/api_cluster.go
+++ b/lxd/api_cluster.go
@@ -3068,6 +3068,14 @@ func clusterNodeStatePost(d *Daemon, r *http.Request) response.Response {
 		return response.BadRequest(err)
 	}
 
+	// Run some pre-checks before evacuating or restoring the cluster member.
+	// It's important that those checks runs on the to be restored cluster member.
+	if s.NetworkReady.Err() == nil {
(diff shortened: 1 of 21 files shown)


roosterfish:
~~I was able to backport all commits as is except the ones that added some additional shell variables:~~

~~E.g. this line on `main` (https://github.com/canonical/lxd/blob/main/test/suites/clustering_waitready.sh#L109) had to be changed to only use `DEBUG=""` instead of `CLIENT_DEBUG="" SHELL_TRACING=""`.
That change was also applied for the other lines in this test.~~

simondeziel (dismissed):
LGTM, thanks

tomponline:
> I was able to backport all commits as is except the ones that added some additional shell variables:
> 
> E.g. this line on `main` (https://github.com/canonical/lxd/blob/main/test/suites/clustering_waitready.sh#L109) had to be changed to only use `DEBUG=""` instead of `CLIENT_DEBUG="" SHELL_TRACING=""`. That change was also applied for the other lines in this test.

Shall we drop commits that had to be massaged and then they'll get brought through during normal backport process, otherwise this is going to require manual massaging back the other way again right?

tomponline:
Or can you also include commits that changed those var names as well, or is that too invasive?

roosterfish:
> Or can you also include commits that changed those var names as well, or is that too invasive?

I would prefer backporting those too so it's complete. Let me check.

roosterfish:
I was able to retain a clean history by backporting a little bit more.

There was only one non-direct conflict in https://github.com/canonical/lxd/pull/15934/commits/5981ebf6288ef04b839ac57fd952b5cf4ff53b21 (#15934) regarding some token/trust password/accept certificate changes in a line not changed in this exact commit. 
But as the original changes/PRs are dating back to October last year, I assume applying those changes was skipped for a reason, also because 5.21 still has password support:

* The line on `main` https://github.com/canonical/lxd/blame/main/test/suites/remote.sh#L22
* Commits skipped that cause the conflict:
  * https://github.com/canonical/lxd/commit/891810bb0a3152c6feff64af150bb35ccdb86e45
  * https://github.com/canonical/lxd/commit/55773796b4f12a96e1435dbb57bdf0d6f97f0b2f
  * https://github.com/canonical/lxd/commit/767d0d13a113b2ae33fd73577a8894a97d26c8e4

PR #16075, opened 2025-07-22
Cluster: Validate a cluster group member removal

### LINKED ISSUE #16074: Cluster group management is broken
### Please confirm

### Distribution

Ubuntu

### Distribution version

snap

```shell
core24  20250618     1055   latest/stable  canonical✓  base
lxd     6.4-2336111  34099  6/stable       canonical✓  disabled
lxd     6.4-1c216d4  34285  6/stable       canonical✓  -
snapd   2.68.5       24718  latest/stable  canonical✓  snapd,disabled
snapd   2.70         24792  latest/stable  canonical✓  snapd
```

```shell
-
```

### Issue description

On editing a cluster group. When removing a member, that is not in any other group -- the removal of the member is ignored and a success returned via api.

### Steps to reproduce

1. Have a clustered LXD with at least one member
2. Create a cluster group and put the member in that group
3. Remove the member from the default group. Save. Ok.
4. Remove the member from the newly created group. Save. Member is still in the group

Expected is either the removal of the member, or returning an error, that the member cannot be removed from the group, because it has to be in at least one group.

### Information to attach

edlerd:
# Done
- validate a cluster group edit does not ignore a member removal

fixes #16074

### DIFF
--- a/lxd/api_cluster.go
+++ b/lxd/api_cluster.go
@@ -4133,6 +4133,27 @@ func clusterGroupPut(d *Daemon, r *http.Request) response.Response {
 			return err
 		}
 
+		members, err := tx.GetClusterGroupNodes(ctx, name)
+		if err != nil {
+			return err
+		}
+
+		// Every member must belong to at least one group.
+		for _, oldMember := range members {
+			// On removing member, ensure it is in at least one other group.
+			if !slices.Contains(req.Members, oldMember) {
+				// Get all cluster groups this member belongs to.
+				groups, err := tx.GetClusterGroupsWithNode(ctx, oldMember)
+				if err != nil {
+					return err
+				}
+
+				if len(groups) == 1 {
+					return fmt.Errorf("Cannot remove %q from group as member needs to belong to at least one group", oldMember)
+				}
+			}
+		}
+
 		obj := dbCluster.ClusterGroup{
 			Name:        group.Name,
 			Description: req.Description,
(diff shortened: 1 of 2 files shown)


lxd/api_cluster.go
```diff
@@ -4133,6 +4133,25 @@ func clusterGroupPut(d *Daemon, r *http.Request) response.Response {
...
+		// ensure no member is removed that is the only member of one group
+		for _, oldMember := range members {
+			if !slices.Contains(req.Members, oldMember) {
+				groups, err := tx.GetClusterGroupsWithNode(ctx, oldMember)
+				if err != nil {
+					return err
+				}
+
+				if len(groups) == 1 {
+					return errors.New("Member " + oldMember + " cannot be removed, it must be in at least one group.")
```
tomponline: Please dont add fullstops to error messages.

lxd/api_cluster.go
```diff
@@ -4133,6 +4133,25 @@ func clusterGroupPut(d *Daemon, r *http.Request) response.Response {
...
 
+		members, err := tx.GetClusterGroupNodes(ctx, name)
+		if err != nil {
+			return err
+		}
+
+		// ensure no member is removed that is the only member of one group
+		for _, oldMember := range members {
+			if !slices.Contains(req.Members, oldMember) {
+				groups, err := tx.GetClusterGroupsWithNode(ctx, oldMember)
```
tomponline, line 4146: Looking further down this function we see the same loop again, calling `groups, err := tx.GetClusterGroupsWithNode(ctx, oldMember)` for each member.

Please can we avoid iterating through the members list twice and loading each member's group twice from the DB?

edlerd: Yes, we actually don't need the check below. Adjusting it to do just one call to database per update.

edlerd:
Thanks for the review @tomponline 

Addressed the concerns above and added a test for the new error case.

lxd/api_cluster.go
```diff
@@ -4133,6 +4133,27 @@ func clusterGroupPut(d *Daemon, r *http.Request) response.Response {
...
+			// On removing member, ensure it is in at least one other group.
+			if !slices.Contains(req.Members, oldMember) {
+				// Get all cluster groups this member belongs to.
+				groups, err := tx.GetClusterGroupsWithNode(ctx, oldMember)
+				if err != nil {
+					return err
+				}
+
+				if len(groups) == 1 {
+					return errors.New("Member " + oldMember + " cannot be removed, it must be in at least one group")
```
tomponline: We quote variables in errors, so please can this be:

```go
fmt.Errorf("Member %q cannot be removed, it must be in at least one group", oldMember)
```

tomponline (approved):
Thanks!

lxd/api_cluster.go
```diff
@@ -4133,6 +4133,27 @@ func clusterGroupPut(d *Daemon, r *http.Request) response.Response {
...
+			// On removing member, ensure it is in at least one other group.
+			if !slices.Contains(req.Members, oldMember) {
+				// Get all cluster groups this member belongs to.
+				groups, err := tx.GetClusterGroupsWithNode(ctx, oldMember)
+				if err != nil {
+					return err
+				}
+
+				if len(groups) == 1 {
+					return fmt.Errorf("Member %s cannot be removed, it must be in at least one group", oldMember)
```
tomponline: Oh this needs to be `%q` not `%s`

tomponline: As per https://github.com/canonical/lxd/pull/16075#discussion_r2254204785

edlerd:
Thanks for the review, adjusted the message as there is the same case handled on PATCH below and that messaged seemed a bit better. While at it also changed the member name to be quoted in the PATCH endpoint as well.

tomponline (approved):
LGTM thanks!

PR #16076, opened 2025-07-22
doc: clarify upgrade instructions

minaelee:
This PR:
- Differentiates update vs. upgrade in the Manage the snap guide
- In that guide, moves section about upgrading (changing the snap channel) closer to the section on updates & improves its content
- Adds information about where to find non-snap installation database backup
- Improves the cluster update/upgrade guide (formerly used 'upgrade' to mean both update and upgrade and gave erroneous information such as that upgrades can happen automatically)
- Adds more links between pages about upgrading


doc/howto/snap.md
```diff
@@ -37,6 +37,13 @@ lxd   5.21.3-c5ae129  33110  5.21/stable  canonical✓  -
 
 The first part of the version string corresponds to the LXD release (in this sample, `5.21.3`).
 
+(howto-snap-updates-vs-upgrades)=
+## Updates versus upgrades
+
+Updating the LXD snap means to refresh it to the release most recently published to its tracked channel. These updates are always within the same major version; they can be automatically or manually performed.
```
tomponline: > These updates are always within the same major version;

This isn't true if following `latest/*` channels.

doc/howto/snap.md
```diff
@@ -171,6 +171,21 @@ Example:
 sudo enterprise-store override lxd stable=25846
 ```
 
+(howto-snap-change)=
+## Change the snap channel
+
+While it is possible to change the channel used at installation, proceed with caution.
+
+You can upgrade (move to a newer {ref}`track <ref-snap-tracks>`, such as from {{current_lts_track}} to {{current_feature_track}}), as well as move to different {ref}`risk level <ref-snap-risk>` with the same track. However, downgrading (moving to a channel with an older track, such as from {{current_feature_track}} to {{current_lts_track}}) is neither recommended nor supported, as breaking changes can exist between major versions.
```
tomponline: > as well as move to different {ref}`risk level <ref-snap-risk>` with the same track.

Sometimes this isn't true when switching risk levels in `latest/*` or the current feature track, as new DB schema changes may have landed in `{latest,6}/{edge,candidate}` that are not in `/stable` that would prevent switch back to `/stable`.

doc/howto/snap.md
```diff
@@ -171,6 +171,21 @@ Example:
...
+
+To change the channel, run:
+
+```bash
+sudo snap refresh lxd --channel=<target channel>
```
tomponline: should we mention the `--cohort=+` thing here for cluster members?

doc/howto/snap.md
```diff
@@ -184,7 +184,15 @@ To change the channel, run:
 sudo snap refresh lxd --channel=<target channel>
 ```
 
-This command immediately updates the installed snap version.
+This also updates the installed snap version to the most recent release in the target channel.
+
+For the channel, you must include at least the track, such as {{current_lts_track}}. You can optionally include the risk-level; when omitted, the default is `stable`. For example, this command changes the snap channel to `5.21/stable`:
```
tomponline: Can we use `{{current_lts_track}}` in the example here too rather than hardcoding 5.21?

tomponline:
@minaelee any news on this one?

doc/howto/snap.md
minaelee: Afraid not, or at least there's no straightforward solution for including substitutions inside code blocks. I've seen some roundabout ways to do it and intend to explore it more next cycle.

minaelee:
Closed https://github.com/canonical/lxd/pull/16306 per similar content & updated this PR to address review comments on both this PR and that one, including information regarding what kinds of downgrading/reverting is supported (change from higher to lower revision or risk level _only_ supported on LTS tracks, not on feature or latest tracks due to potential breaking changes even within the same track)

PR #16085, opened 2025-07-24
doc: linkcheck ignore ceph.io

minaelee:
ceph.io reachable by browser but returning 500 Internal Server Error for sphinx linkchecks and curl requests (even curl requests with a browser user-agent set). This PR causes all pages under https://ceph.io to be ignored by the linkcheck to avoid test failures.


tomponline (approved):
Ta

PR #16087, opened 2025-07-25
lxd/sys/apparmor: do not try to get the cache dir if Apparmor is not available

simondeziel:
Fixes commit f4be44fa9214ac9d88a9ba256077a64580f070ca that accidentally introduced a `panic` if Appamor is not available.

Reported-by: Katie May <katie.may@canonical.com>
Co-authored-by: Mark Laing <mark.laing@canonical.com>

### DIFF
--- a/lxd/sys/apparmor.go
+++ b/lxd/sys/apparmor.go
@@ -63,11 +63,16 @@ func (s *OS) initAppArmor() []cluster.Warning {
 
 	// Detect AppArmor cache directories.
 	s.AppArmorCacheLoc = shared.VarPath("security", "apparmor", "cache")
-	s.AppArmorCacheDir, err = appArmorGetCacheDir(s.AppArmorVersion, s.AppArmorCacheLoc)
-	if err != nil {
-		logger.Warn("AppArmor feature cache directory detection failed", logger.Ctx{"err": err})
+	if s.AppArmorAvailable {
+		s.AppArmorCacheDir, err = appArmorGetCacheDir(s.AppArmorVersion, s.AppArmorCacheLoc)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


lxd/sys/apparmor.go
```diff
@@ -63,11 +63,15 @@ func (s *OS) initAppArmor() []cluster.Warning {
...
+	if s.AppArmorAvailable {
+		s.AppArmorCacheDir, err = appArmorGetCacheDir(s.AppArmorVersion, s.AppArmorCacheLoc)
+		if err != nil {
+			logger.Warn("AppArmor feature cache directory detection failed", logger.Ctx{"err": err})
+		} else {
+			logger.Debug("AppArmor feature cache directory detected", logger.Ctx{"cache_dir": s.AppArmorCacheDir})
+		}
 	} else {
-		logger.Debug("AppArmor feature cache directory detected", logger.Ctx{"cache_dir": s.AppArmorCacheDir})
+		s.AppArmorCacheDir = s.AppArmorCacheLoc
```
tomponline, line 75: Could do with a comment here as to why we are setting this when apparmor isnt available?

simondeziel: This is essentially to mimic what was done before. For example, `s.AppArmorStacking` and `s.AppArmorStacked` are set irrespective of `s.AppArmorAvailable`.

Not sure how to clearly convey this in a comment.

Maybe the better way would be to change the other fields to not be set if `s.AppArmorAvailable == false`?

tomponline: I think a comment would suffice for now. It seems unexpected to be setting the CacheDir when apparmor wasnt detected.

tomponline (approved):

PR #16089, opened 2025-07-25
lxd/instance: Allow setting lxc.net config keys through raw.lxc (from Incus)

simondeziel:


tomponline:
@simondeziel whats the reason behind needing this?

@skozina should we be adding more raw.* support for specific container runtimes?

skozina:
Yup, I thought we want to limit the number of low-level tunables in general, to avoid the cases when LXD lags behind the latest changes in the lower layer configs. Curious to hear more about the need here.

simondeziel:
> @simondeziel whats the reason behind needing this?

Saw this pass by and found the (dropped) validation could have been done more efficiently. Since this was only reachable via `raw.lxc`, I felt it was OK to allow for more exotic config which are outside of the supported path anyway.

simondeziel:
> @skozina should we be adding more raw.* support for specific container runtimes?

It's no adding support for more top level `raw.*` key but about relaxing the constraints on what you can do with the existing `raw.lxc` key.

tomponline:
@simondeziel thanks, is this needed for something you're working on?

simondeziel:
> @simondeziel thanks, is this needed for something you're working on?

No. It seems like you have reasons not to want the cherry-pick which is fine with me (I just don't know why but that's OK :)
Should I drop the single cherry-pick from the PR and keep the rest which is mostly tests.

skozina:
Sound good to me. Considering this is an existing unsupported tunnable, there's really no need to have the explicit restrictions for the IP address only.

PR #16094, opened 2025-07-28
doc: combine UI and CLI tutorials into a single improved tutorial

minaelee:
This PR:
- Combines the UI tutorial into the CLI tutorial at `doc/tutorial/first_steps.md`. The existing two tutorials did basically the same things, including installing and initializing LXD (through the CLI in both) then creating and using instances (through the CLI in one and through the UI in the other). This version uses CLI/UI tabs instead..
- Deletes the tutorial index (no longer needed since there are no longer two tutorials).
- Updates toctrees and redirects to direct tutorial links to first_steps.md.
- Uses tabs to provide instructions for both CLI and UI in first_steps.md.
- Updates content for the CLI and UI tutorial steps so that they perform the same operations instead of diverging unnecessarily.
- Removes profiles and storage sections of UI tutorial (they do not exist in CLI tutorial, could be added back later as a more advanced tutorial, but I'd like to keep the "First steps" tutorial at a more reasonable length and scope).
- Reduces the lengthy configuration section by removing the section on configuring the size of a VM's root disk device.
- Reduces creating 4 separate containers via the CLI to 2 containers (same as what the UI does).
- Uses section headings instead of numbered lists. This makes it easier to organize content in tabs and improves the level of detail in the table of contents panel.
- Adds creating a second VM via the CLI from the `images:` remote server to demonstrate configuring an instance at creation and using a non-default image server; the UI instructions already did this, so this also aligns the CLI instructions with the UI ones.
- Uses very simple `hello world` text file to simplify the demo of pushing and pulling files, instead of /etc/hosts and syslog.
- Adds optional sections for deleting all instances at the end and holding updates.
- Adds friendlier wording throughout (less "Do this, do that" and more friendly contextual guidance).

### DIFF
--- a/doc/tutorial/first_steps.md
+++ b/doc/tutorial/first_steps.md
@@ -1,380 +1,776 @@
 (first-steps)=
+(tutorial-first-steps)=
 # First steps with LXD
+
+This tutorial guides you through your first steps with LXD. You'll begin by installing and initializing LXD. Then you'll use its CLI or graphical web UI to work with instances, including both containers and virtual machines. You'll learn how to create and configure instances, create snapshots, and more.
 
 ````{only} integrated
 
 ```{admonition} For MicroCloud users
 :class: note
-The MicroCloud setup process installs and initializes LXD. Thus, you do not need to follow the steps on this page.
+The MicroCloud setup process installs and initializes LXD. Thus, you can skip those sections on this page.
 ```
 
 ````
 
-This tutorial guides you through the first steps with LXD.
-It covers installing and initializing LXD, creating and configuring some instances, interacting with the instances, and creating snapshots.
+(tutorial-requirements)=
+## Requirements
 
-After going through these steps, you will have a general idea of how to use LXD, and you can start exploring more advanced use cases!
+- At least 20 GiB free disk space
+- A Linux distribution installed
 
-```{note}
-Ensure that you have 20 GiB free disk space before starting this tutorial.
+(tutorial-install)=
+## Install LXD using snap
+
+This section of the tutorial assumes that you have the `snap` packaging system available on your system, which is the recommended way to use LXD.
+
+If you use a Linux distribution that does not support `snap`, see {ref}`installing-other` and skip to the next section of this tutorial.
+
+**LXD snap installation requirements**
+
+- The LXD snap must be [available for your Linux distribution](https://snapcraft.io/lxd#distros).
+- The [`snapd` daemon](https://snapcraft.io/docs/installing-snapd) must be installed.
+
+The `snapd` daemon that manages snap packages comes pre-installed on many distributions by default. To confirm whether it is available on your system, run:
+
+```bash
+snap version
 ```
 
-<!-- Include start tutorial installation -->
+If you see an error message indicating that `snap` is not installed, visit the [Snap installation documentation](https://snapcraft.io/docs/installing-snapd) and follow the instructions there to install it.
 
-## Install and initialize LXD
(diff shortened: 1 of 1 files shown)


doc/tutorial/first_steps.md
```diff
@@ -1,380 +1,706 @@
...
+```bash
+sudo snap install lxd
+```
 
-      If you see a table of version numbers, snap is installed and you can continue with the next step of installing LXD.
+If you see an error message that the LXD snap is already installed, run the following command to refresh it and ensure that you are running an up-to-date version:
 
-   1. If the command returns an error, run the following commands to install the latest version of ``snapd`` on Ubuntu:
+```bash
+sudo snap refresh lxd
```
tomponline, line 59: Should we instruct the user to check they have at least LXD 5.21 here so that if they are on an older OS with an older LXD preinstalled we can ensure they can be aware up front that the tutorial isn't going to be 100% accurate for them?

doc/tutorial/first_steps.md
```diff
@@ -1,380 +1,706 @@
...
+```
 
-          sudo apt update
-          sudo apt install snapd
+(tutorial-adduser)=
+## Add the current user to the `lxd` group
```
tomponline, line 65: We should probably mention (or link to the existing statement on this matter) about the ramifications of adding the current user to the `lxd` user, in that it gives them effectively root access.

doc/tutorial/first_steps.md
minaelee: I added instructions for them to check the version, refresh/update if it's the current LTS track or higher, and upgrade if lower. I'd rather not hard code a version number here, but it does mean we need to keep the tutorial updated to work with the current LTS and higher (which should be a given).

doc/tutorial/first_steps.md
minaelee: I added the same security notice as used in the how to install guide here.

doc/tutorial/first_steps.md
```diff
@@ -1,380 +1,706 @@
...
-If you prefer a different installation method, or use a Linux distribution that is not supported by the snap package, see {ref}`installing`.
+- The LXD snap must be [available for your Linux distribution](https://snapcraft.io/lxd#distros).
+- The [`snapd` daemon](https://snapcraft.io/docs/installing-snapd) must be installed.
 
-1. Install `snapd`:
+The `snapd` daemon that manages snap packages comes pre-installed on many distributions by default. To confirm whether it is available on your system, run:
 
-   1. Run `snap version` to find out if snap is installed on your system:
+```bash
+snap version
```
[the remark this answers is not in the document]
minaelee: I don't understand this as it's already marked as a bash code block.

minaelee:
I am conducting UX testing and am converting to draft while I incorporate the feedback.

Main feedback: Make sure that explanations about what commands do are always provided before asking the user to blindly run the commands, then providing the explanation after the commands. Also, `newgrp` did not work for the tester, so include a note about closing and re-opening the terminal if needed.

minaelee:
Additional UX testing session completed, and feedback from the session incorporated, including:
- Supplement use of `{terminal}`block with the actual command above it in simple Markdown code blocks, because there is no `Copy` button in the `{terminal}` block, and the user expects to be able to copy commands.
- Where bash commands to use `lxc shell` are followed by other commands in the same code block, copying and pasting these blocks results in the commands after `lxc shell` to be held in buffer until the shell is exited, then run on the host system. Updated to separate out commands following `lxc shell` into separate code blocks.

PR #16095, opened 2025-07-28
VM: Persist device bus number order to volatile config

tomponline:


tomponline:
Replaced by https://github.com/canonical/lxd/pull/16708

PR #16120, opened 2025-07-30
DB: Add `ScanValue` method and interfaces.

markylaing:
`sql.Scanner` implementations all perform the same logic which calls `driver.TYPE.ConvertValue`, then performs a type assertion.

This PR adds intermediary interfaces that can be implemented and a `ScanValue` method that moves the repeated logic into one place.

This blocks #15832 (which will use `query.ScanValue` to scan the `type` column in the `secrets` table).

### DIFF
--- a/lxd/db/cluster/entities.go
+++ b/lxd/db/cluster/entities.go
@@ -11,6 +11,7 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/canonical/lxd/lxd/db/query"
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/entity"
 )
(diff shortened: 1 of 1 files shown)


lxd/db/cluster/identities.go
```diff
@@ -158,12 +143,22 @@ func (i *IdentityType) Scan(value any) error {
...
+		return fmt.Errorf("Unknown identity type `%d`", identityTypeCode)
 	}
 
 	return nil
 }
 
+// Scan implements sql.Scanner for IdentityType. This converts the integer value back into the correct API constant or
+// returns an error.
+func (i *IdentityType) Scan(value any) error {
+	if value == nil {
```
roosterfish: Just a nit, can those checks for null be moved in the general `ScanValue` func? Maybe `ScanValue` can grow another arg which indicates whether or not the to be scanned value can be null?

roosterfish (dismissed):
Looks good to me, only one small question.

lxd/db/cluster/identities.go
markylaing: Ah yes good idea :)

lxd/db/query/scan.go
```diff
@@ -0,0 +1,58 @@
...
+
+		i, ok := driverVal.(int64)
+		if !ok {
+			return fmt.Errorf("Expected int64, got `%v` (%T)", driverVal, driverVal)
+		}
+
+		return scanner.ScanInteger(i)
+	}
+
+	return fmt.Errorf("Cannot scan %v, target %v does not implement any scanner interfaces", value, target)
```
tomponline: Can we include %T here so we know what type is the problem?

lxd/db/cluster/identities.go
markylaing: Done

lxd/db/query/scan.go
markylaing: Done

PR #16128, opened 2025-07-31
lxd/cluster/config: Remove `ClusterTx` from `Config`

markylaing:
A pointer to `config.Config` is saved to the `Daemon` struct but the transaction it holds is not valid outside of the initial
transaction where it was loaded. This is pretty strange. Instead, let's pass a transaction only into the methods that require it.


skozina (commented):
I do believe the `Config.tx` transaction is basically useless after the initial `daemon.init()` is done, so good to have it removed. Thanks!

PR #16129, opened 2025-07-31
Add `volatile.uuid` to `config` table

markylaing:
This global UUID can serve as a stable identifier for the cluster, and will be used for both OIDC sessions and token based authentication, as it will be present in JWTs as the issuer.

Depends in #16128 so will keep as draft for now.

Also, to get this to work with cluster config was a little weird. This was the simplest way I could get it to work without doing any large scale refactoring of cluster configuration.

### DIFF
--- a/test/suites/serverconfig.sh
+++ b/test/suites/serverconfig.sh
@@ -6,8 +6,28 @@ test_server_config() {
   _server_config_access
   _server_config_storage
   _server_config_auth_secret
+  _server_config_cluster_uuid
 
   kill_lxd "${LXD_SERVERCONFIG_DIR}"
+}
+
+_server_config_cluster_uuid() {
+  # Validate that the cluster UUID cannot be changed
+
+  # PUT
+  ! lxc config unset volatile.uuid || false
+  ! lxc config set volatile.uuid="$(uuidgen)" || false
+  cluster_uuid="$(lxc config get volatile.uuid)"
+  lxc config set volatile.uuid="${cluster_uuid}"
+
+  # PATCH
+  my_curl -X PATCH "https://${LXD_ADDR}/1.0" -d "{\"config\":{\"core.https_address\":\"${LXD_ADDR}\"}}" | jq -e '.status == "Success" and .status_code == 200'
+  my_curl -X PATCH "https://${LXD_ADDR}/1.0" -d "{\"config\":{\"core.https_address\":\"${LXD_ADDR}\",\"volatile.uuid\":\"\"}}" | jq -e '.error == "The cluster UUID cannot be changed" and .error_code == 400'
+  my_curl -X PATCH "https://${LXD_ADDR}/1.0" -d "{\"config\":{\"core.https_address\":\"${LXD_ADDR}\",\"volatile.uuid\":\"$(uuidgen)\"}}" | jq -e '.error == "The cluster UUID cannot be changed" and .error_code == 400'
+  my_curl -X PATCH "https://${LXD_ADDR}/1.0" -d "{\"config\":{\"core.https_address\":\"${LXD_ADDR}\",\"volatile.uuid\":\"${cluster_uuid}\"}}" | jq -e '.status == "Success" and .status_code == 200'
+
+  # Check that the cluster UUID matches the server UUID
+  [ "$(cat "${LXD_DIR}/server.uuid")" = "${cluster_uuid}" ]
 }
 
 _server_config_auth_secret() {


tomponline:
> Also, to get this to work with cluster config was a little weird

@skozina knows all about this, perhaps we should schedule some time to refactor out the "Schema" concept in global config.

tomponline:
> I am a little confused of what validation we do through the `Validator()` function of the config struct vs. what validation we're doing directly in the code from the API handlers

I'm not too keen on extending the Validator with additional functionality (this feels like a responsibility beyond what the Validator should be doing) and would prefer to see a list of unchangeable keys in being checked for change in the handler itself (along with tests to check its taking effect at blocking change).

The Validator can then just be concerned with whether the value is valid or not.

tomponline:
> could we use a v7 uuid so we can extract the deployment date of the cluster (at least the first node) if needed later?

Also whilst we're at it we could change the per-node uuid file we write (cc @roosterfish ) to use v7 uuids too for the same reason.

roosterfish:
> Also whilst we're at it we could change the per-node uuid file we write (cc @roosterfish ) to use v7 uuids too for the same reason.

We are using the `server.uuid` also for NVMe/TCP protocol when crafting the host's NQN and for the host ID. 
In case of the host's NQN the string is crafted using `nqn.2014-08.org.nvmexpress:uuid:<uuidv4 from LXD>` and I don't know if we can just replace this to v7.
Or is there a way to convert from v7 to v4?

tomponline:
It would need to be tested, but they are the same format at the end of the day, just with encoded info in them.

tomponline:
@roosterfish @markylaing i think we should make it so that the initial volatile.uuid generated matches the node uuid generated and stored in the file (i.e generate it and save it to 2 different places, the volatile.uuid config key and on disk). That way it works the same as the cluster certificate being the same as the initial node's server cert (which I think is the case right?).

markylaing:
@tomponline @roosterfish @skozina this is ready for review when you have a moment, thanks.

lxd/daemon.go
```diff
@@ -1158,6 +1158,33 @@ func (d *Daemon) init() error {
 	d.startStopLock.Lock()
 	defer d.startStopLock.Unlock()
 
+	// Setup and load the server's UUID file.
```
tomponline: Not a blocker, but this would have been a great opportunity to reduce cyclomatic complexity by moving this to a separate function on `Daemon` that can be called from here.

lxd/patches.go
```diff
@@ -1568,4 +1569,30 @@ func patchMoveBackupsImagesStorage(name string, d *Daemon) error {
 	return nil
 }
 
+// patchClusterConfigVolatileUUID checks if the clusterUUID is defined and if not, adds the server UUID to config under
+// 'volatile.uuid'. Note that this means for existing deployments (i.e. for whom this patch applies) the cluster UUID is
+// a v4 UUID, whereas new installations will have a v7 UUID.
+func patchClusterConfigVolatileUUID(name string, d *Daemon) error {
```
tomponline, line 1576: @markylaing have you considered that in addition to:

> Note that this means for existing deployments (i.e. for whom this patch applies) the cluster UUID is
// a v4 UUID, whereas new installations will have a v7 UUID.

That the UUID will not necessarily be the Node UUID associated with the cluster certificate.

I wonder if for both of those reasons we just instead generate a new random v7 uuid for existing deployments.

What do you think?

test/suites/serverconfig.sh
```diff
@@ -6,10 +6,27 @@ test_server_config() {
   _server_config_access
   _server_config_storage
   _server_config_auth_secret
+  _server_config_cluster_uuid
 
   kill_lxd "${LXD_SERVERCONFIG_DIR}"
 }
 
+_server_config_cluster_uuid() {
```
tomponline, line 14: worth checking its the same as the member ID too here?

lxd/patches.go
markylaing: I was back and forth between the two and in the end settled on this but I'm open to just generating a new v7 UUID :)

Will make the change

tomponline: At least that way its always a v7

roosterfish: Just stumbled across this. 
Do I understand it right that for existing deployments the `volatile.uuid` is different from the `server.uuid` in the daemon's directory because it's not of type v7 and we (of course) cannot simply replace it?

roosterfish:
I was finally able to test this, seems to be all working with PowerFlex hosts and the v7 UUID.

lxd/patches.go
markylaing: Yep that's correct. This means the cluster UUID is always a v7 UUID. For new deployments, each server UUID will be v7 and the cluster UUID will be the UUID of the bootstrapping cluster member.

PR #16130, opened 2025-07-31
Backports (stable-5.21)

tomponline:


tomponline:
@simondeziel @skozina I also included the previously skipped Content-Type enforcement changes in this PR because we later changed the enforcement checks in `main` from always checking Content-Type (which would have broken older clients) to only enforcing Content-Type for browser clients, which then made this set of changes backportable.

Issue #12648, opened 2023-12-10
Removing an LXD disk device deletes an empty directory it was bind mounted over


dmitry-lyfar:
# Required information

 * Distribution: ubuntu
 * Distribution version: 23.04
 * The output of "lxc info" or if that fails:
   * Kernel version:
   * LXC version:
   * LXD version: 5.19-8635f82
   * Storage backend in use: dir

# Issue description

LXD behaves inconsistently when it deals with bind-mounts on empty instance's directories. If an instance is running and one removes a disk device that was bind mounted over an existing empty directory, the underlying directory gets removed as well. Conversely, when the instance is stopped, removing the disk device would not lead to removing the instance's empty directory.

As per the [source code](https://github.com/canonical/lxd/blob/e90ae16cc7c573d5a6ad37783e888190afda9ffd/lxd/instance/drivers/driver_lxc.go#L1744), that seems to be a designed behaviour as LXD creates the target (path) directories if those did not exist on the host. However, it leads to removing directories that were not created by LXD.

# Steps to reproduce

Note: /opt is an existing and empty directory.

```
lxc launch ubuntu:22.04 craft-0758614d
lxc start craft-0758614d
lxc config device add craft-0758614d test disk source=/home/dmitry/test path=/opt
lxc config device remove craft-0758614d test
lxc exec craft-0758614d -- ls /opt
ls: cannot access '/opt': No such file or directory
```

roosterfish:
Hi @dmitry-lyfar it looks like if you have a file with the same name as `path=XYZ` inside the container, it gets removed too:

```
lxc launch ubuntu:22.04 c1
lxc exec c1 -- touch /tmp/test
lxc config device add c1 test disk source=/tmp path=/tmp/test
lxc config device remove c1 test
lxc exec c1 -- ls /tmp/test
ls: cannot access '/tmp/test': No such file or directory
```

gabrielmougard:
I managed to find a workaround that solves both issues but I don't quite know is this is ideal as it introduces a potential 'marking file' inside the mounted directory in the container:  

```diff
diff --git a/lxd/instance/drivers/driver_lxc.go b/lxd/instance/drivers/driver_lxc.go
index 4e939d0e1..70156e8c1 100644
--- a/lxd/instance/drivers/driver_lxc.go
+++ b/lxd/instance/drivers/driver_lxc.go
[...]

PR #16131, opened 2025-07-31
Add `path.create` configuration option for `disk` devices

kadinsayani:
Closes https://github.com/canonical/lxd/issues/12648.

This pull request introduces a new configuration option, `path.create`, to control whether the mount path inside a container is automatically created for `disk` devices.


test/suites/container_devices_disk.sh
```diff
@@ -183,6 +183,15 @@ _container_devices_disk_char() {
   lxc restart -f foo
   [ "$(lxc exec foo -- stat /root/zero -c '%F')" = "character special file" ]
   lxc config device remove foo char
+  # Can't create mount when "path.create" is false and path does not exist in container.
+  ! lxc config device add foo char disk source=/dev/zero path=/root/test path.create=false || false
+  lxc exec foo -- touch /root/test
```
simondeziel, line 188: I'd like a test where the path exists but is of the wrong type. Like a dir when a file is needed and vice-versa.

kadinsayani:
With my current approach, the new `CreatePath` field adds complexity since it needs to be considered every time we add a mount and not just for `disk` devices. This is why the `container_devices_gpu` tests are failing.

One potential approach is to include device type or the entire device config in `MountEntryItem`, that way we can check the type and only skip path creation for `disk` devices. This could also help us in the future if we ever have additional considerations for mounts based on device type.

test/suites/container_devices_disk.sh
kadinsayani: Currently, we don't have checks for this. If a `/some/path/to/*.txt` file exists, the only requirement is that the path exists, even if we're trying to mount a directory. I'll add checks with this PR and some tests :)

kadinsayani:
It looks like Incus has taken a different approach: https://github.com/lxc/incus/pull/2335/commits/3150202a805efe8b8dfe22edfd57d6e99f855e57.

test/suites/container_devices_disk.sh
kadinsayani: Relevant Incus PR: https://github.com/lxc/incus/pull/2326.

tomponline:
> It looks like Incus has taken a different approach: [lxc/incus@3150202](https://github.com/lxc/incus/commit/3150202a805efe8b8dfe22edfd57d6e99f855e57).

@kadinsayani I'd be happy to go with that approach. 

This effectively changes the behaviour to create mount points (if needed), but doesn't delete them, except if they are under `/dev` right?

Any objections @skozina ?

kadinsayani:
> > It looks like Incus has taken a different approach: [lxc/incus@3150202](https://github.com/lxc/incus/commit/3150202a805efe8b8dfe22edfd57d6e99f855e57).
> 
> @kadinsayani I'd be happy to go with that approach.
> 
> This effectively changes the behaviour to create mount points (if needed), but doesn't delete them, except if they are under `/dev` right?

Correct, I think it makes sense to only remove mount points created in `/dev`.

kadinsayani:
@tomponline shall I also open a separate PR cherry-picking https://github.com/lxc/incus/pull/2326?

tomponline:
> @tomponline shall I also open a separate PR cherry-picking [lxc/incus#2326](https://github.com/lxc/incus/pull/2326)?

yes please

kadinsayani:
Closing in favor of https://github.com/canonical/lxd/pull/16137.

PR #16138, opened 2025-08-01
VM: Prevent file mounts (from Incus)

kadinsayani:

PR #16156, opened 2025-08-05
tests: Add basic disk mount VM checks

kadinsayani:
Follow up to https://github.com/canonical/lxd/pull/16138.

### DIFF
--- a/test/suites/vm.sh
+++ b/test/suites/vm.sh
@@ -51,4 +51,12 @@ test_vm_empty() {
     # Cleanup custom changes from the default profile
     lxc profile unset default migration.stateful
   fi
+
+  # This check must be run after unsetting "migration.stateful" to workaround 9p limitations; shared filesystems are incompatible.
+  echo "==> Disk mounts"
+  lxc init --vm --empty v1 -c limits.memory=128MiB -d "${SMALL_ROOT_DISK}"
+  lxc config device add v1 char disk source=/dev/zero path=/dev/zero
+  # Attempting to mount a single file as a disk device is not supported for VMs; this should fail at start time.
+  ! lxc start v1 || false
+  lxc delete v1
 }


test/suites/vm.sh
```diff
@@ -47,8 +49,18 @@ test_vm_empty() {
...
-  if grep -qxF 'VERSION_ID="22.04"' /etc/os-release; then
+  if [ "${isJammyRunner}" = "0"]; then
     # Cleanup custom changes from the default profile
     lxc profile unset default migration.stateful
   fi
+
+  if [ "${isJammyRunner}" = 1 ]; then
+    # Only check filesystem mounts where virtiofsd is supported/compatible.
+    echo "==> Disk mounts"
+    lxc launch --vm --empty v1 -c limits.memory=128MiB -d "${SMALL_ROOT_DISK}"
```
tomponline: We don't need to start the VM to test this right, its validated at add time?

kadinsayani: Its validated at start time in `(*disk).startVM`.

tomponline: Ah OK then

tomponline (approved):

tomponline:
@kadinsayani github has messes up, can you force push s change to retriggee workflows as I can't manually do it.

test/suites/vm.sh
```diff
@@ -9,7 +9,9 @@ test_vm_empty() {
     return
   fi
 
-  if grep -qxF 'VERSION_ID="22.04"' /etc/os-release; then
+  isJammyRunner=$(grep -qxF 'VERSION_ID="22.04"' /etc/os-release && echo 0 || echo 1)
```
tomponline: This doesn't look right as jammy runners should match 22.04 and don't support virtiofsd

tomponline (changes requested):
Test version checks look wrong

tomponline (changes requested):
I've been thinking about this and we dont need the Jammy (or rather Noble) check for directory disk hotplug support because we should be checking for directory source paths even when using 9p to pass the directory to the VM.

Now 9p doesn't support hotplug, so we just need to add the disk to a stopped VM and then try and start the VM and it shouldn't start if trying to pass a file as a disk source.

test/suites/vm.sh
```diff
@@ -51,4 +51,13 @@ test_vm_empty() {
...
   fi
+
+  echo "==> Disk mounts"
+  lxc init --vm --empty v1 -c limits.memory=128MiB -d "${SMALL_ROOT_DISK}"
+  lxc config device add v1 char disk source=/dev/zero path=/dev/zero
+  # File mounting to filesystem paths is not supported for virtual machines. This is checked at start time.
+  ! lxc start v1 || false
+  lxc config device remove v1 char
+  lxc config device add v1 char disk source=/dev/ path=/dev/
+  lxc delete --force v1
```
tomponline, line 62: you could just have done `lxc delete --force

test/suites/vm.sh
```diff
@@ -51,4 +51,13 @@ test_vm_empty() {
...
     lxc profile unset default migration.stateful
   fi
+
+  echo "==> Disk mounts"
+  lxc init --vm --empty v1 -c limits.memory=128MiB -d "${SMALL_ROOT_DISK}"
+  lxc config device add v1 char disk source=/dev/zero path=/dev/zero
+  # File mounting to filesystem paths is not supported for virtual machines. This is checked at start time.
+  ! lxc start v1 || false
+  lxc config device remove v1 char
+  lxc config device add v1 char disk source=/dev/ path=/dev/
```
tomponline: you didnt try starting the VM after this line so you've not tested that directories can be added.

tomponline (changes requested):
nearly there, but not quite right yet. Thanks

kadinsayani:
> nearly there, but not quite right yet. Thanks

Should be good to go now, thanks :)

simondeziel (approved):
LGTM, thanks!

PR #16158, opened 2025-08-05
Mount point removal follow up

kadinsayani:
Follow up to https://github.com/kadinsayani/lxd/pull/new/tests/mount-point-removal.

This pull request adds tests to ensure removal of `/dev` directories when a disk device is removed and warns rather than returns an error when removal fails.

### DIFF
--- a/lxd/instance/drivers/driver_lxc.go
+++ b/lxd/instance/drivers/driver_lxc.go
@@ -1696,7 +1696,10 @@ func (d *lxc) deviceHandleMounts(mounts []deviceConfig.MountEntryItem) error {
 
 				// Only remove mountpoints created in /dev.
 				if strings.HasPrefix(mount.TargetPath, "dev/") {
-					return files.Remove(relativeTargetPath)
+					err = files.Remove(relativeTargetPath)
+					if err != nil {
+						d.logger.Warn("Could not remove the device path inside container", logger.Ctx{"err": err, "path": relativeTargetPath})
+					}
 				}
 			}
 
--- a/test/suites/container_devices_unix.sh
+++ b/test/suites/container_devices_unix.sh
@@ -117,6 +117,7 @@ _container_devices_unix() {
   rm "${testDev}"
   sleep 1
   ! lxc exec "${ctName}" -- mount | grep -F "/tmp/testdev" || false
+  lxc exec "${ctName}" -- test -e /tmp/testdev
   ! test -e "${LXD_DIR}"/devices/"${ctName}"/unix.test--dev--dynamic.tmp-testdev || false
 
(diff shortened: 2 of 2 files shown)


lxd/instance/drivers/driver_lxc.go
```diff
@@ -1696,7 +1696,10 @@ func (d *lxc) deviceHandleMounts(mounts []deviceConfig.MountEntryItem) error {
 
 				// Only remove mountpoints created in /dev.
 				if strings.HasPrefix(mount.TargetPath, "dev/") {
-					return files.Remove(relativeTargetPath)
+					err = files.Remove(relativeTargetPath)
+					if err != nil {
+						d.logger.Warn("Could not remove the device path inside container", logger.Ctx{"err": err})
```
tomponline: Lets include the path in the error context please

test/suites/container_devices_unix.sh
```diff
@@ -113,10 +113,17 @@ _container_devices_unix() {
   [ "$(lxc exec "${ctName}" -- stat -c '%F %a %t %T' /tmp/testdev)" = "${deviceTypeDesc} 660 0 0" ]
   [ "$(stat -c '%F %a %t %T' "${LXD_DIR}"/devices/"${ctName}"/unix.test--dev--dynamic.tmp-testdev)" = "${deviceTypeDesc} 660 0 0" ]
 
+  # Check removal of mount point devices created in /dev.
+  lxc config device add "${ctName}" test-dev5 disk source=/dev/zero path=/dev/test
+  lxc config device remove "${ctName}" test-dev5
+  ! lxc exec "${ctName}" -- mount | grep -F "/dev/test" || false
```
tomponline, line 119: Suggest using something like this:

https://github.com/canonical/lxd-ci/pull/547/files#diff-3f95fb5cc9095f27cc0eb1d002484ae66c5d2a832218fcdd6853d90931a8d0a1R370

tomponline: @kadinsayani do we need to change the lxd-agent behaviour and tests in lxd-ci for VM behaviour?

kadinsayani: The busybox test image is missing `findmnt`. What's wrong with the current method of checking?

tomponline: ah ok, only that it was using grep rather than exact match

kadinsayani: > @kadinsayani do we need to change the lxd-agent behaviour and tests in lxd-ci for VM behaviour?

I believe we are covered by https://github.com/canonical/lxd/pull/16051/commits/af5e54fb37849af4fa4bc1176bf06a3ce7416c72 for `lxd-agent`.

PR #15990, opened 2025-07-09
Per-project backups and images storage

skozina:

PR #16160, opened 2025-08-05
Followup project storage

skozina:
Follow up on #15990


lxd/config/map.go
```diff
@@ -275,6 +275,18 @@ const (
 	DaemonStorageTypeBackups DaemonStorageType = "backups"
 )
 
+// String returns the string representation of the DaemonStorageType.
+func (t DaemonStorageType) String() string {
```
tomponline: You don't need this as the underlying type value is string. You can just cast it to string if needed.

lxd/node/config.go
```diff
@@ -112,30 +112,30 @@ func (c *Config) StorageBucketsAddress() string {
 // daemonStorageVolume returns the volume configured as images or backups storage for target project.
 // If project is not specified, or if project has no specifid storage volume configured, the daemon
 // storage volume is returned.
-func (c *Config) daemonStorageVolume(projectName string, storageType string) string {
+func (c *Config) daemonStorageVolume(projectName string, storageType config.DaemonStorageType) string {
 	// If project is not specified, return daemon storage.
 	if projectName == "" {
-		return c.m.GetString("storage." + storageType + "_volume")
+		return c.m.GetString("storage." + storageType.String() + "_volume")
```
tomponline: Can just do `string(storageType)`

lxd/config/map.go
skozina: Ack, I've replaced this with simple type conversion.

lxd/node/config.go
skozina: Replaced with `string(storageType)`.

tomponline (approved):
LGTM thanks!

PR #16161, opened 2025-08-05
Refactor identity type handling with type-safe interface

kadinsayani:
This pull request replaces string-based utility functions with a `Type` interface that provides type-safe access to identity type behavior. This change moves away from utility functions like `IsFineGrainedIdentityType` and `IsRestrictedIdentityType` to a centralized, interface-based approach.

Key changes:
- Add `identity.Type` interface with methods for authentication, admin status, caching, and fine-grained permission capabilities
- Implement `New()` function to validate identity type strings and return concrete types that implement the interface
- Create individual identity type structs (`OIDCClient`, `CertificateClient`, etc.) with embedded `typeInfoCommon` for shared default behavior
- Update all callers to use the new interface methods instead of utility functions
- Preserve behavior while improving type safety and code organization

### DIFF
--- a/lxd/identity/cache.go
+++ b/lxd/identity/cache.go
@@ -70,20 +70,23 @@ func (c *Cache) Get(authenticationMethod string, identifier string) (*CacheEntry
 }
 
 // GetByType returns a map of identifier to CacheEntry, where all entries have the given identity type.
-func (c *Cache) GetByType(identityType string) map[string]CacheEntry {
+func (c *Cache) GetByType(identityTypeStr string) map[string]CacheEntry {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
 
-	// Explicitly ignore the error here. It is expected that the caller will use the constants defined in shared/api.
-	authenticationMethod, _ := AuthenticationMethodFromIdentityType(identityType)
-	entriesByAuthMethod, ok := c.entries[authenticationMethod]
+	identityType, err := New(identityTypeStr)
+	if err != nil {
+		return nil
+	}
+
+	entriesByAuthMethod, ok := c.entries[identityType.AuthenticationMethod()]
 	if !ok {
 		return nil
 	}
 
 	entriesOfType := make(map[string]CacheEntry)
 	for _, entry := range entriesByAuthMethod {
-		if entry != nil && entry.IdentityType == identityType {
+		if entry != nil && entry.IdentityType == identityTypeStr {
 			entriesOfType[entry.Identifier] = *entry
 		}
 	}
--- a/lxd/auth/drivers/openfga.go
+++ b/lxd/auth/drivers/openfga.go
@@ -183,8 +183,13 @@ func (e *embeddedOpenFGA) checkPermission(ctx context.Context, entityURL *api.UR
 	logCtx["protocol"] = id.AuthenticationMethod
 	l := e.logger.AddContext(logCtx)
 
+	identityType, err := identity.New(id.IdentityType)
+	if err != nil {
+		return err
+	}
+
 	// If the identity type does not use fine-grained auth use the TLS driver instead.
-	if !identity.IsFineGrainedIdentityType(id.IdentityType) {
+	if !identityType.IsFineGrained() {
 		return e.tlsAuthorizer.CheckPermission(ctx, entityURL, entitlement)
 	}
 
@@ -392,8 +397,13 @@ func (e *embeddedOpenFGA) getPermissionChecker(ctx context.Context, entitlement 
 	logCtx["protocol"] = id.AuthenticationMethod
 	l := e.logger.AddContext(logCtx)
 
+	identityType, err := identity.New(id.IdentityType)
+	if err != nil {
+		return nil, err
+	}
+
 	// If the identity type does not use fine-grained auth, use the TLS driver instead.
-	if !identity.IsFineGrainedIdentityType(id.IdentityType) {
+	if !identityType.IsFineGrained() {
 		return e.tlsAuthorizer.GetPermissionChecker(ctx, entitlement, entityType)
 	}
 
--- a/lxd/auth/util.go
+++ b/lxd/auth/util.go
@@ -41,12 +41,12 @@ func IsServerAdmin(ctx context.Context, identityCache *identity.Cache) (bool, er
 		return false, fmt.Errorf("Failed to get caller identity: %w", err)
 	}
 
-	isRestricted, err := identity.IsRestrictedIdentityType(id.IdentityType)
+	identityType, err := identity.New(id.IdentityType)
 	if err != nil {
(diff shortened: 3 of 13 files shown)


lxd/identity/type.go
```diff
@@ -0,0 +1,247 @@
...
+// isFineGrained returns false by default.
+func (typeInfoCommon) isFineGrained() bool {
+	return false
+}
+
+type identityTypeOIDCClient struct {
+	typeInfoCommon
+}
+
+func (identityTypeOIDCClient) authenticationMethod() string {
```
tomponline: There's no need to prefix these types with "identity" when they are in the "identity" package, it just leads to stammering and unnecessarily long type names.

lxd/identity/type.go
```diff
@@ -0,0 +1,247 @@
...
+	return api.AuthenticationMethodOIDC
+}
+
+func (identityTypeOIDCClient) isCacheable() bool {
+	return true
```
tomponline: where an identity type diverges from the common implementation it would be useful to readers of the code for a comment to added explaining why the difference exists.

lxd/identity/type.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func (identityTypeOIDCClient) isFineGrained() bool {
+	return true
+}
+
+type identityTypeCertificateClient struct {
```
tomponline: Please can you separate each identity type type definition by (at least) a title comment explaining what the identity type is for or perhaps into separate files, as otherwise the reviewer is faced with a long block of type definitions with no clear separation.

lxd/identity/type.go
```diff
@@ -0,0 +1,247 @@
...
+
+// Type represents an identity type in LXD.
+//
+// To create a new identity type:
+// 1. Add a new const in db/cluster/identities.go.
+// 2. Implement db functions in db/cluster/identities.go.
+// 3. Add an API type in shared/api/auth.go.
+// 4. Add a new const [Type].
+// 5. Create a type that implements typeInfo and add it to the [identityTypes] map.
+type Type string
```
tomponline: Because you've not implemented a type loader (like `lxd/storage/drivers/Load()` or `lxd/instance/drivers/load()`) you've been forced to handle the case of an an invalid string being cast to a `Type` in every attached function, even if the is no other reason to return an error (like `IsAdmin`) which then complicates every function.

Suggest instead implementing a `New` function that takes an identity `Type` string so that it can be used to validate the identity type early on and then the returned struct can then have simpler functions on it.

tomponline: Separately I don't really see the need for the outer type `Type` that has the exported functions on them and then internally just passes through the call to the driver's unexported function.

How about we simplify this:

1. Create a `Type` type as an interface that defines the exported (and optionally unexported) functions an identity type needs to implement.
2. Keep the `typeInfoCommon` struct by alter its implementations to comply with the parts of the new `Type` interface that are relevant.
3. Create a per-type struct that implements the full `Type` interface (optionally by embedding `typeInfoCommon` and if needed overriding certain functions if needed).
4. Create a `New(identityType string) (Type, error)` function to validate the `identityType` string and return a pointer to the relevant identity type struct that will be cast to a `Type` interface as its returned.

tomponline (changes requested):
I think the idea is good, but the implementation is a bit overly complicated for my liking.

I think we can simplify this which will make the caller's life easier and be more consistent with how we've done other subsystem drivers.

lxd/identity/type.go
markylaing: I recommended this pattern because it's what we use for entity types but I can see the pain points here. This suggestion sounds good to me :)

tomponline: @markylaing was there a reason the entity types needed to use that more abstracted pattern?

tomponline: The pattern as it is more like the storage subsystem where we have the outer `backendLXD` implementation that implements the `Pool` interface. With hindsight the `Pool` interface isn't really needed, and we could rename `backendLXD` to `Pool` and just use the concrete type everywhere.

However the 2-tier abstraction at the storage subsystem layer is still beneficial in my view as the `backendLXD` implementation deals with commonalities between the storage drivers, as well as DB management, and the storage drivers themselves are implementing a `Driver` interface, which `backendLXD` interacts with.

This separation of concerns works well.

But in this case (and maybe Entity types too) if the outer implementation is just acting as a "dumb" pass through to the per-driver implementation then it doesn't offer any benefits in my view, and is just unnecessary abstraction.

markylaing: > @markylaing was there a reason the entity types needed to use that more abstracted pattern?

I don't recall, but it's possible that it was necessary because entity types need to be split into the public (shared/entity) and private (lxd/db/cluster) implementations

lxd/identity/type.go
kadinsayani: I've added additional comments for each identity type's implementations.

For the default implementations I've gone with `return false` for `IsAdmin()`, `IsFineGrained()`, and `IsCacheable()`. We could reduce LOC by changing this for `IsFineGrained()` and `IsCacheable()`, but I think setting the defaults to `false` forces behavior of future identities to be carefully considered when implementing their methods. 

I've applied similar logic for `AuthenticationMethod()` where the only differing implementation is for OIDC clients. There is no default implementation, so although we have repetition, it forces defining the authentication method when adding a new type. What do you think about this?

lxd/identity/cache.go
```diff
@@ -70,20 +70,20 @@ func (c *Cache) Get(authenticationMethod string, identifier string) (*CacheEntry
...
 // GetByType returns a map of identifier to CacheEntry, where all entries have the given identity type.
-func (c *Cache) GetByType(identityType string) map[string]CacheEntry {
+func (c *Cache) GetByType(identityTypeStr string) map[string]CacheEntry {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
 
 	// Explicitly ignore the error here. It is expected that the caller will use the constants defined in shared/api.
-	authenticationMethod, _ := AuthenticationMethodFromIdentityType(identityType)
-	entriesByAuthMethod, ok := c.entries[authenticationMethod]
+	identityType, _ := New(identityTypeStr)
```
tomponline: You need to test the error here to avoid a panic for invalid types on next line.

tomponline (dismissed):

lxd/identity/cache.go
kadinsayani: I didn't handle the error because that comment states it's expected that the caller always uses the identity type API constants. It's probably better to return `nil` if `GetByType()` receives an invalid string though. Alternatively, we could change the param from `string` to `Type` and call `New()` at the call site, then pass the correct identity type to `GetByType()`.

tomponline: I'm fine either way as long as we close down the panic possibility.

tomponline (approved):
Lgtm

PR #16162, opened 2025-08-06
Network: Internal OVN load balancers and forwards (WIP)

nmezhenskyi:
This PR adds support for internal OVN load balancers and network forwards, enabling the use of internal OVN IPs as a listen address.


lxd/db/network_load_balancers.go
```diff
@@ -276,13 +277,18 @@ func (c *ClusterTx) GetProjectNetworkLoadBalancerListenAddressesByUplink(ctx con
...
 	JOIN projects ON projects.id = networks.project_id
+	JOIN networks_config AS nc_filter on networks.id = nc_filter.network_id
+	LEFT JOIN networks_config AS nc_ipv4 ON networks.id = nc_ipv4.network_id AND nc_ipv4.key = 'ipv4.address'
+	LEFT JOIN networks_config AS nc_ipv6 ON networks.id = nc_ipv6.network_id AND nc_ipv6.key = 'ipv6.address'
 	WHERE (
-		(networks_config.key = "network" AND networks_config.value = ?1)
+		(nc_filter.key = "network" AND nc_filter.value = ?1)
 		OR (projects.name = "default" AND networks.name = ?1)
 	)
 	`)
```
nmezhenskyi, line 294: This query is used in:
 https://github.com/canonical/lxd/blob/7b3a50b2bd82d1e343e0ad2935a1fa35067b7b30/lxd/network/driver_common.go#L214
 
 https://github.com/canonical/lxd/blob/7b3a50b2bd82d1e343e0ad2935a1fa35067b7b30/lxd/network/driver_common.go#L1224
 
 https://github.com/canonical/lxd/blob/7b3a50b2bd82d1e343e0ad2935a1fa35067b7b30/lxd/project/limits/permissions.go#L1770
 
 In all these contexts, it is safe to omit listen addresses that are internal OVN IPs.
 
 But alternatively, I can add an additional `bool` parameter (i.e., `externalOnly`) to the `GetProjectNetworkLoadBalancerListenAddressesByUplink()` function signature to explicitly determine whether internal listen addresses should be filtered out.

tomponline: at least you should update the function's comment to explain it I think.

lxd/network/driver_ovn.go
```diff
@@ -5952,6 +5961,95 @@ func (n *ovn) checkAddressNotInUse(netip *net.IPNet) (bool, error) {
...
+		// Listen address is not an internal OVN IP, therefore, it is not used internally.
+		return false, nil
+	}
+
+	// Check that the listen address is not taken by the OVN gateway.
+	if netIP.Equal(listenAddress) {
+		return true, fmt.Errorf("Listen address %q is already in use by %q of network %q", listenAddress, netIPKey, n.name)
+	}
+
+	var forwards map[int64]string
```
tomponline, line 6000: tip:

```go
var forwards, loadBalancers map[int64]string
```

lxd/db/network_load_balancers.go
nmezhenskyi: Makes sense, I'll update it

PR #16163, opened 2025-08-06
lxd/db/cluster: Add missing id column to GetCoreAuthSecrets

markylaing:
#15832 unfortunately broke the OIDC flow because this column was missing. I'm confused because I did check that OIDC was working after performing the final changes, but perhaps I forgot to rebuild LXD :facepalm:


tomponline:
@markylaing looks like the tests arent covering this then as they passed too. Can we add some to test this at least at a basic level somehow?

markylaing:
> @markylaing looks like the tests arent covering this then as they passed too. Can we add some to test this at least at a basic level somehow?

This would be covered by the UI e2e tests, they just haven't run yet as I caught it early enough. I can add unit tests though?

tomponline:
@markylaing unit tests would be good too thanks

markylaing:
By adding the unit tests I found another issue. When the secrets were rotated, the newly created secret held in-memory did not have the `ID` set, which could have meant that old secrets would not be properly deleted when the secrets rotate again.

PR #16164, opened 2025-08-06
Recovery: Learn about volumes from instance's backup config

roosterfish:
Includes a cherry-picked [commit](https://github.com/lxc/incus/pull/678/commits/350fa762a4e5044a25ff4650fc85f7c9e6763f2b) from Incus that has fixed the issue https://github.com/lxc/incus/issues/646 which also affects `lxd recover`.

This PR essentially makes use of the new and updated backup config stored alongside instances to recover both instances and attached custom volumes.
For custom volumes which aren't attached to any instance we fall back to the old behavior of recovering the volume solely from its name.

The changes in this PR pave the way to implement the `ListVolumes` driver func of PowerFlex, Pure and Alletra to enable recovery also for those storage drivers.

### DIFF
--- a/test/main.sh
+++ b/test/main.sh
@@ -551,6 +551,7 @@ if [ "${1:-"all"}" != "cluster" ]; then
     run_test test_warnings "Warnings"
     run_test test_metrics "Metrics"
     run_test test_storage_volume_recover "Recover storage volumes"
+    run_test test_storage_volume_recover_by_container "Recover storage volumes by container"
     run_test test_syslog_socket "Syslog socket"
     run_test test_lxd_user "lxd user"
     run_test test_waitready "waitready"
--- a/test/suites/backup.sh
+++ b/test/suites/backup.sh
@@ -69,6 +69,95 @@ EOF
   shutdown_lxd "${LXD_IMPORT_DIR}"
 }
 
+test_storage_volume_recover_by_container() {
+  LXD_IMPORT_DIR=$(mktemp -d -p "${TEST_DIR}" XXX)
+  spawn_lxd "${LXD_IMPORT_DIR}" true
+
+  poolName=$(lxc profile device get default root pool)
+  poolDriver=$(lxc storage show "${poolName}" | awk '/^driver:/ {print $2}')
+
+  # Create another storage pool.
+  poolName2="${poolName}-2"
+  lxc storage create "${poolName2}" "${poolDriver}"
+
+  # Create container.
+  ensure_import_testimage
+  lxc init testimage c1 -d "${SMALL_ROOT_DISK}"
+
+  # Create a custom volume and attach to the instance.
+  lxc storage volume create "${poolName}" vol1 size=32MiB
+  lxc storage volume snapshot "${poolName}" vol1
+  lxc storage volume attach "${poolName}" vol1 c1 /mnt
+
+  # Create a custom volume in a different pool and attach to the instance.
+  lxc storage volume create "${poolName2}" vol2 size=32MiB
+  lxc storage volume snapshot "${poolName2}" vol2
+  lxc storage volume attach "${poolName2}" vol2 c1 /mnt2
+
+  # Get the volume's UUIDs before deleting it's database entries.
+  vol1_uuid="$(lxc storage volume get "${poolName}" vol1 volatile.uuid)"
+  vol2_uuid="$(lxc storage volume get "${poolName2}" vol2 volatile.uuid)"
+
+  # Delete database entries of the created custom volumes.
+  lxd sql global "PRAGMA foreign_keys=ON; DELETE FROM storage_volumes WHERE name='vol1'"
+  lxd sql global "PRAGMA foreign_keys=ON; DELETE FROM storage_volumes WHERE name='vol2'"
+
+  # Ensure the custom volumes are no longer listed.
+  ! lxc storage volume show "${poolName}" vol1 || false
+  ! lxc storage volume show "${poolName2}" vol2 || false
+
+  # Recover custom volumes.
+  cat <<EOF | lxd recover
+no
+yes
+yes
+EOF
+
+  # Ensure custom storage volumes have been recovered.
+  lxc storage volume show "${poolName}" vol1 | grep -xF 'content_type: filesystem'
+  lxc storage volume show "${poolName2}" vol2 | grep -xF 'content_type: filesystem'
+
+  # Ensure the custom volumes still have the same UUIDs.
+  # This validates that the custom storage volumes were recovered from the instance's backup config.
+  [ "${vol1_uuid}" = "$(lxc storage volume get "${poolName}" vol1 volatile.uuid)" ]
+  [ "${vol2_uuid}" = "$(lxc storage volume get "${poolName2}" vol2 volatile.uuid)" ]
+
+  # Detach the custom volumes from the instance.
+  lxc storage volume detach "${poolName}" vol1 c1
+  lxc storage volume detach "${poolName2}" vol2 c1
+
+  # Delete database entries of the created custom volumes.
+  lxd sql global "PRAGMA foreign_keys=ON; DELETE FROM storage_volumes WHERE name='vol1'"
+  lxd sql global "PRAGMA foreign_keys=ON; DELETE FROM storage_volumes WHERE name='vol2'"
+
+  # Ensure the custom volumes are no longer listed.
+  ! lxc storage volume show "${poolName}" vol1 || false
+  ! lxc storage volume show "${poolName2}" vol2 || false
+
+  # Recover custom volumes.
+  cat <<EOF | lxd recover
+no
+yes
+yes
+EOF
+
+  # Ensure custom storage volumes have been recovered.
+  lxc storage volume show "${poolName}" vol1 | grep -xF 'content_type: filesystem'
+  lxc storage volume show "${poolName2}" vol2 | grep -xF 'content_type: filesystem'
+
+  # Check the custom volumes got different UUIDs.
+  # This validates that the custom storage volumes were recovered by name which looses all of their configuration.
+  [ "${vol1_uuid}" != "$(lxc storage volume get "${poolName}" vol1 volatile.uuid)" ]
+  [ "${vol2_uuid}" != "$(lxc storage volume get "${poolName2}" vol2 volatile.uuid)" ]
+
+  # Cleanup
+  lxc storage volume delete "${poolName}" vol1
+  lxc storage volume delete "${poolName2}" vol2
+  lxc delete -f c1
+  lxc storage delete "${poolName2}"
+  shutdown_lxd "${LXD_IMPORT_DIR}"
+}
+
 test_container_recover() {
   LXD_IMPORT_DIR=$(mktemp -d -p "${TEST_DIR}" XXX)
   spawn_lxd "${LXD_IMPORT_DIR}" true


lxd/storage/backend_lxd.go
```diff
@@ -7661,11 +7661,11 @@ func (b *lxdBackend) detectUnknownInstanceVolume(vol *drivers.Volume, projectVol
...
 
-	if instID > 0 && volume != nil {
-		return nil // Instance record and storage record already exists in DB, no recovery needed.
-	} else if instID > 0 {
+	// Check if either instance or storage volume already exist in the DB.
+	// If only one of them exists this is considered to be an error.
+	if instID > 0 && volume == nil {
 		return fmt.Errorf("Instance %q in project %q already has instance DB record", instName, projectName)
-	} else if volume != nil {
+	} else if instID < 0 && volume != nil {
```
tomponline: <= 0?

lxd/storage/backend_lxd.go
```diff
@@ -7669,36 +7669,39 @@ func (b *lxdBackend) detectUnknownInstanceVolume(vol *drivers.Volume, projectVol
 		return fmt.Errorf("Instance %q in project %q already has storage DB record", instName, projectName)
 	}
 
-	// Check snapshots are consistent between storage layer and backup config file.
-	_, err = b.CheckInstanceBackupFileSnapshots(backupConf, projectName, nil)
-	if err != nil {
-		return fmt.Errorf("Instance %q in project %q has snapshot inconsistency: %w", instName, projectName, err)
-	}
+	// Instance record and storage record don't exist in DB, continue recovering the instance.
+	if instID < 0 && volume == nil {
```
tomponline: <= 0?

lxd/storage/backend_lxd.go
tomponline, line 1: Why would this be the case?

> In case the instance also needs to be restored we delete all the custom volumes from the list so we don't
accidentally report the custom volumes twice to the caller.

lxd/storage/backend_lxd.go
```diff
@@ -7669,6 +7669,58 @@ func (b *lxdBackend) detectUnknownInstanceVolume(vol *drivers.Volume, projectVol
...
+	// Iterate over the custom volumes attached to the instance.
+	for _, customVol := range backupConf.Volumes {
+		// Skip the instances root volume.
+		if customVol == rootVol {
+			continue
+		}
+
+		// The custom volume might be located on a different pool.
+		// Therefore try to load the right pool.
+		customVolPool, err := backupVolConfCache.getPool(customVol.Pool)
```
tomponline, line 7692: Nice to see this cache concept coming in useful here too.

lxd/storage/backend_lxd.go
```diff
@@ -7669,6 +7669,58 @@ func (b *lxdBackend) detectUnknownInstanceVolume(vol *drivers.Volume, projectVol
...
+		if err != nil {
+			// We don't know the pool which hosts the custom volume. Skip it for now.
+			// At this point we would have to notify the user about the new pool and ask for it to be recovered too.
```
tomponline, line 7695: This comment is odd, it seems to suggest we need to notify the user, but we're ignoring the error.
Can you clarify the comment please.

lxd/storage/backend_lxd.go
```diff
@@ -7669,6 +7669,58 @@ func (b *lxdBackend) detectUnknownInstanceVolume(vol *drivers.Volume, projectVol
...
+			if api.StatusErrorCheck(err, http.StatusNotFound) {
+				continue
+			}
+
+			return fmt.Errorf("Failed to load pool %q for custom volume %q in project %q: %w", customVol.Pool, customVol.Name, customVol.Project, err)
+		}
+
+		// Check if any entry for the custom volume already exists in the DB.
+		// This will return no record for any temporary pool structs being used (as ID is -1).
```
tomponline, line 7704: This comment confused me a bit, can it be clarified why this comment is relevant here?

lxd/storage/backend_lxd.go
```diff
@@ -7669,6 +7669,58 @@ func (b *lxdBackend) detectUnknownInstanceVolume(vol *drivers.Volume, projectVol
...
+		volume, err := VolumeDBGet(customVolPool, customVol.Project, customVol.Name, drivers.VolumeTypeCustom)
+		if err != nil && !response.IsNotFoundError(err) {
+			return err
```
tomponline: Worth wrapping the error here?

lxd/storage/backend_lxd.go
```diff
@@ -7669,6 +7669,58 @@ func (b *lxdBackend) detectUnknownInstanceVolume(vol *drivers.Volume, projectVol
...
+		if projectVols[customVol.Project] == nil {
+			projectVols[customVol.Project] = []*backupConfig.Config{backupConf}
+		} else {
+			projectVols[customVol.Project] = append(projectVols[customVol.Project], backupConf)
+		}
+	}
+
+	// Unset the custom volumes as we don't want them to be returned twice.
+	// Once as a custom volume backup config and another time as part of the instance's backup config.
+	backupConf.Volumes = slices.DeleteFunc(backupConf.Volumes, func(vol *backupConfig.Volume) bool {
```
tomponline, line 7729: The lingering question I have thats not been answered by the comments is *why* did we extract the attached volumes into the `projectVols` map if they were also usable in the instance's backupConfig.Volumes later? Why go to the work if they need to be removed. I think the comments here could be clarified on the why this is needed as it seems counter-intuitive.

lxd/api_internal_recover.go
```diff
@@ -72,6 +72,33 @@ type internalRecoverImportPost struct {
 	Pools []api.StoragePoolsPost `json:"pools" yaml:"pools"`
 }
 
+// volumeConfigIsLatest checks whether the given volumeConfig's volume is newer than the existingVolumeConfig's volume.
+// The volume configs are expected to have a single volume in their list of volumes.
```
tomponline, line 76: Lets enforce this with a check in this function then, as the `for` loop suggests that this might not be the case.

lxd/api_internal_recover.go
```diff
@@ -99,6 +99,72 @@ func volumeConfigIsLatest(volumeConfig *backupConfig.Config, existingVolumeConfi
 	return false, false
 }
 
+// appendUnknownVolumeConfig tries to add the given volume in the global map of discovered volumes.
+// In case it contains the most recent set of information, it will replace an already existing one.
+func appendUnknownVolumeConfig(originalPoolName string, projectName string, volumeConfig *backupConfig.Config, poolsProjectVols map[string]map[string][]*backupConfig.Config) error {
+	// We only ever expect a single volume per volume config.
+	// In case an instance has custom volumes attached, those are extracted an should be presented as their own custom volume config.
```
tomponline: /an/and/

lxd/api_internal_recover.go
```diff
@@ -99,6 +99,72 @@ func volumeConfigIsLatest(volumeConfig *backupConfig.Config, existingVolumeConfi
...
+	// This ensures the same behavior performed by detecUnknownCustomVolume which puts every custom volume discovered by name
+	// into its own backup config struct.
+	if len(volumeConfig.Volumes) != 1 {
+		// The backup config for buckets is special as they don't have a volume listed.
+		// In case it's not a bucket's backup config, consider it to be an error.
+		if volumeConfig.Bucket == nil {
```
tomponline: can be combined with the if above.

lxd/api_internal_recover.go
```diff
@@ -99,6 +99,72 @@ func volumeConfigIsLatest(volumeConfig *backupConfig.Config, existingVolumeConfi
...
+			return errors.New("Backup config of unknown volume contains more or less than one volume")
+		}
+	}
+
+	var unknownVol *backupConfig.Volume
+
+	// Check if we can extract an actual volume from the backup config.
+	// This does not work for a bucket's config as it doesn't track neither pool nor volume.
+	if len(volumeConfig.Volumes) > 0 {
```
tomponline: It feels like this can be combined with the if statement section above, as it feels related but some what duplicated here.

lxd/api_internal_recover.go
```diff
@@ -99,6 +99,72 @@ func volumeConfigIsLatest(volumeConfig *backupConfig.Config, existingVolumeConfi
...
+		unknownVol = volumeConfig.Volumes[0]
+
+		// Change the original pool the volume was discovered on as the given volume config describes a volumes on another pool.
+		// This is imporant when the volume doesn't yet exist so it gets created under the right pool.
+		originalPoolName = unknownVol.Pool
```
tomponline, line 127: This part confused me, can you explain the reason why the `volumeConfig.Volumes[0]` has come to have the wrong pool name in it?

lxd/api_internal_recover.go
```diff
@@ -99,6 +99,72 @@ func volumeConfigIsLatest(volumeConfig *backupConfig.Config, existingVolumeConfi
...
+			if volumeExists {
+				return nil
+			}
+		}
+	}
+
+	// In case the volume couldn't be used to replace an existing one or it doesn't exist at all, append it.
+	// This is always true for buckets.
+	if !volumeExists {
+		if poolsProjectVols[originalPoolName] == nil {
```
tomponline, line 155: why are we switching between using `unknownVol.Pool` above and back to `originalPoolName` here, this section has quite confused me.

lxd/api_internal_recover.go
```diff
@@ -99,6 +99,72 @@ func volumeConfigIsLatest(volumeConfig *backupConfig.Config, existingVolumeConfi
 	return false, false
 }
 
+// appendUnknownVolumeConfig tries to add the given volume in the global map of discovered volumes.
+// In case it contains the most recent set of information, it will replace an already existing one.
+func appendUnknownVolumeConfig(originalPoolName string, projectName string, volumeConfig *backupConfig.Config, poolsProjectVols map[string]map[string][]*backupConfig.Config) error {
```
tomponline, line 104: for clarity should we use a pointer to `poolsProjectVols` to make it clear we're modifying it?

tomponline: does `volumeConfig` need to be a pointer?

lxd/api_internal_recover.go
tomponline: What i'd prefer to see is first a check for if the  volume exists, then check if `poolsProjectVols` contains the relevant pool and project and if not create those entries, and then add the volume to the slice if needed.

lxd/api_internal_recover.go
```diff
@@ -414,6 +414,9 @@ func internalRecoverScan(ctx context.Context, s *state.State, userPools []api.St
 					var displayType, displayName string
 					var displaySnapshotCount int
 
+					// Usually the volume's pool name is the one where it originates from.
```
tomponline: "Usually"? Can we refactor so this is always the case?

lxd/storage/backend_lxd.go
```diff
@@ -6647,7 +6647,15 @@ func (b *lxdBackend) ImportCustomVolume(projectName string, poolVol *backupConfi
...
 
 	// Copy volume config from backup file if present (so VolumeDBCreate can safely modify the copy if needed).
-	vol := b.GetNewVolume(drivers.VolumeTypeCustom, drivers.ContentType(customVol.ContentType), volStorageName, customVol.Config)
+	vol := b.GetVolume(drivers.VolumeTypeCustom, drivers.ContentType(customVol.ContentType), volStorageName, customVol.Config)
+
+	// Only modify the UUID if not already set by the original volume.
+	// A custom volume can either be imported directly through its name (which loses any further information)
+	// or by checking an instances backup config (which lists all of the volumes original configuration).
+	// In the latter case we don't want to override the original UUID to allow a consistent recovery.
+	if vol.Config()["volatile.uuid"] == "" {
```
tomponline, line 6656: Is this a general bug not specific to recovery? Can we add a test for it?

lxd/storage/backend_lxd.go
roosterfish: Yes we can do this to ensure we catch all cases even though I think there won't ever be an instance with id `0` as the autoincrement starts at 1?

lxd/storage/backend_lxd.go
roosterfish: In case `tx.GetInstanceID` returns a "not found", it sets the returned ID to -1. But as before, let's add this in to catch all cases. Fixed.

lxd/storage/backend_lxd.go
roosterfish: I tried to keep this PR as minimal as possible.

Since we now know all of the attached custom volume's pools, at this point we can notify the user about its presence (so that he creates the pool using the discussed `source.reuse` config key).

As of today we don't have this information and don't perform any action on this so I thought I'll just add a comment here mentioning it for the follow up.

lxd/storage/backend_lxd.go
roosterfish: I copied the comment from the other invocations of `VolumeDBGet` on the `detectUnkonwn*` functions. 

My understanding is that if you add another to be recovered pool as part of `lxd recover`, it's only present in memory during the recovery (not in the DB), so if you try to fetch it from the DB (from a temporary pool), it cannot return any record. 
The statement "as ID is -1" should refer to the volume's ID which doesn't exist as there is no DB entry.

lxd/storage/backend_lxd.go
roosterfish: Fixed, also for the other invocation of `VolumeDBGet` in this func.

lxd/storage/backend_lxd.go
roosterfish: Up in the callchain (when evaluating the response from `ListUnknownVolumes` during recovery), we don't anymore want to deal with backup configs that contain multiple volumes. 

Today each of the returned backup configs contain only a single volume. No matter if its the backup config of an instance volume or custom volume.
As we could now have multiple volumes per instance's volume backup config, we already "unpack" them here into their own backup configs (same as they would be returned when running `detecUnknownCustomVolume`). 

Ultimately this makes it easier later on because when dealing with a backup config we can be sure it only every contains a single volume which makes the handling more convenient. Also there is a check in the recovery code that tests for this and errors out if a backup config contains more than one volume.

roosterfish: Happy to update the docstring if you see fit.

lxd/storage/backend_lxd.go
roosterfish: Hope I have answered it here https://github.com/canonical/lxd/pull/16164/commits/ab27c376d4a9077b0d778db84f85e974f66eed50#r2343473395.

lxd/api_internal_recover.go
roosterfish: We already check in `appendUnknownVolumeConfig` which calls this func.

lxd/api_internal_recover.go
roosterfish: > for clarity should we use a pointer to poolsProjectVols to make it clear we're modifying it?

I see your point but isn't it more cumbersome to write something like `(*poolsProjectVols)[unknownVol.Pool]` inside the func just for the sake of having a pointer argument? To make it more clear I have already indicated in the functions docstring that it appends to the global map.

`ListUnknownVolumes` returns pointers to the unknown backup configs. So I continued using the pointer to only pass the reference to the backup config struct. In case an instance has various custom vols attached, the struct might grow to a significant size.

lxd/api_internal_recover.go
roosterfish: Thanks, fixed.

lxd/api_internal_recover.go
roosterfish: Fixed.

lxd/api_internal_recover.go
```diff
@@ -72,6 +72,99 @@ type internalRecoverImportPost struct {
...
+func appendUnknownVolumeConfig(originalPoolName string, projectName string, volumeConfig *backupConfig.Config, poolsProjectVols map[string]map[string][]*backupConfig.Config) error {
+	// We only ever expect a single volume per volume config.
+	// In case an instance has custom volumes attached, those are extracted an should be presented as their own custom volume config.
+	// This ensures the same behavior performed by detecUnknownCustomVolume which puts every custom volume discovered by name
+	// into its own backup config struct.
+	if len(volumeConfig.Volumes) != 1 {
+		// The backup config for buckets is special as they don't have a volume listed.
+		// In case it's not a bucket's backup config, consider it to be an error.
+		if volumeConfig.Bucket == nil {
+			return errors.New("Backup config of unknown volume contains more or less than one volume")
```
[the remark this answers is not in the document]
roosterfish: Fixed

lxd/api_internal_recover.go
roosterfish: The first `if` checks if there is more or less than one volume in case the given backup config is not a bucket's backup config. The length of volumes can either be 0, 1, or x.

The second `if` is only valid in case there is at least one volume (not applicable for buckets as they don't have a volume in the backup config). 

So we could potentially have a nested `if`, the outer checking for `volumeConfig.Bucket == nil` and the inner checking `len(volumeConfig.Volumes) != 1`. But I think it's easier to read this way as we first tackle the potential error and then continue on extracting the vol if it exists.

However I'll change `len(volumeConfig.Volumes) > 0` to `len(volumeConfig.Volumes) == 1` to make it more explicit when this condition is true.

In the other comment you mentioned why we ensure there is always one vol per backup config and why do we need to create a new backup config for each custom volume attached to an instance and then remove all the custom vols from the instance's backup config. That's why :)

roosterfish: Fixed.

lxd/api_internal_recover.go
roosterfish: `originalPoolName` is the pool we have executed the `ListUnknownVolumes` on. It returns both volumes from this pool and maybe from other pools. For the latter we have to align the pool name and pick the actual one which is set in the volume's backup config. 

Updated the comment to make it more clear.

lxd/api_internal_recover.go
```diff
@@ -72,6 +72,99 @@ type internalRecoverImportPost struct {
...
+
+	// If we were able to extract a volume from the config, check if it already exists.
+	// If it exist we might want to use it to replace a volume with less enriched information.
+	if unknownVol != nil && poolsProjectVols[unknownVol.Pool] != nil {
+		for i, existingVolConfig := range poolsProjectVols[unknownVol.Pool][unknownVol.Project] {
+			volumeLatest := false
+
+			volumeExists, volumeLatest = volumeConfigIsLatest(volumeConfig, existingVolConfig)
+			if volumeLatest {
+				poolsProjectVols[unknownVol.Pool][unknownVol.Project][i] = &backupConfig.Config{Volumes: []*backupConfig.Volume{unknownVol}}
```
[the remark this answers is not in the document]
roosterfish: Very interesting suggestion. And it proved to be true. 

Let me unfold the issue. In case the volume is already known and we have one which is newer that can replace it, we do the replacement but miss out the last modified field.
This now means any other backup config describing this volume is either newer or has an identical zero modified time set which might again cause a replacement.

In a real world scenario this would have never happened because we might have discovered a vol just by its name and then by checking another pool find out that we have a newer backup config so we replace the original one. There is no "third" replacement only in case the vol is used by any other instance which happens to have an ever more recent version of the backup config (due to potential failed backup config updates before).

Fixed by picking the actual `volumeConfig` whithout having to craft a new one.

lxd/api_internal_recover.go
roosterfish: We have to deal with all three types: instances, custom vol and bucket backup configs. The bucket one doesn't have anything set so that is the reason for passing `originalPoolName` to the `appendUnknownVolumeConfig` func.

Now it might be that we are dealing with a custom volume from another pool, so we have to at some point update the `originalPoolName` so that we use the correct pool name.

lxd/api_internal_recover.go
roosterfish: Absolutely. Fixed.

lxd/storage/backend_lxd.go
roosterfish: We call `ImportCustomVolume` only during recovery (from `internalRecoverScan`) so it's only recovery related.

I would classify it as a bug yes.

roosterfish: Actually I discovered it whilst writing the test in the next commit which explicitly checks the volumes `volatile.uuid` in `test_storage_volume_recover_by_container`. So we now have a test :)

lxd/storage/backend_lxd.go
```diff
@@ -6647,7 +6647,15 @@ func (b *lxdBackend) ImportCustomVolume(projectName string, poolVol *backupConfi
 	volStorageName := project.StorageVolume(projectName, customVol.Name)
```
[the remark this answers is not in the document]
roosterfish: Can be ignored as I have specifically added a comment describing the issue.

roosterfish:
> I'm wondering if we can refactor so that the pool name in the volume config is always correct, so that it doesn't need to be corrected higher up in the call stack?

The pool name in the volume's backup config is correct (we ensure this in all of the `detectUnknown*` funcs and is what we use as the baseline throughout the process to figure out if the volume is already located on another pool.

Now we iterate through all pools, invoke `ListUnknownVolumes` on them and then iterate over all the returned backup configs, check their volume's actual pool and then move the backup config to the respective pool if we don't see fit.

lxd/storage/backend_lxd.go
tomponline: right got you.

lxd/storage/backend_lxd.go
tomponline: Right, so i think you're saying you want to split the instance's backup config so that it doesn't contain its attached custom volumes, so that they can be returned individually? Right?

lxd/api_internal_recover.go
tomponline: would it make more sense to push the check down into here?

lxd/storage/backend_lxd.go
tomponline: Adding "TODO" somewhere in that sentence would have clarified it didn't exist yet.

lxd/storage/backend_lxd.go
roosterfish: Yes, that is it. I have adapted this "behavior" because I saw that the current custom volume recovery also creates a backup config with a single volume for every recovered custom vol so I thought lets continue using this pattern and adapt the custom volume recovery from an instance's config to it.

lxd/storage/backend_lxd.go
```diff
@@ -7539,10 +7539,11 @@ func (b *lxdBackend) ListUnknownVolumes(op *operations.Operation) (map[string][]
 	return projectVols, nil
 }
 
-// detectUnknownInstanceVolume detects if a volume is unknown and if so attempts to mount the volume and parse the
-// backup stored on it. It then runs a series of consistency checks that compare the contents of the backup file to
+// detectUnknownInstanceAndCustomVolumes parses the given volume's backup config and detects if an instance's volume is unknown.
+// It then runs a series of consistency checks that compare the contents of the backup file to
 // the state of the volume on disk, and if all checks out, it adds the parsed backup file contents to projectVols.
-func (b *lxdBackend) detectUnknownInstanceVolume(vol *drivers.Volume, projectVols map[string][]*backupConfig.Config, op *operations.Operation) error {
+// In any case it also checks whether or not the instance has unknown custom volumes attached and appends them to projectVols too.
```
tomponline, line 7553: this docblock comment could have been expanded to explain that it removes custom volume config from the instance backupConfig entry in projecVols.

tomponline (approved):
LGTM thanks!

As a side note, in the future please consider breaking larger functions into smaller ones rather than adding new stanzas to already complicated functions, as this will help to reduce our cyclomatic complexity score (some of the functions in this area of the code base are very complicated and long).

roosterfish:
> LGTM thanks!
> 
> As a side note, in the future please consider breaking larger functions into smaller ones rather than adding new stanzas to already complicated functions, as this will help to reduce our cyclomatic complexity score (some of the functions in this area of the code base are very complicated and long).

Will shortly send another PR to fix the remaining comments.

lxd/api_internal_recover.go
roosterfish: By checking it in `appendUnknownVolumeConfig`, we only run it once per unknown volume. 

Because `volumeConfigIsLatest` is called for each of the already processed unknown vols, so we would repeat the check for the same unknown volume multiple times as we iterate over all the pool's volumes in the respective project to find the same one.

lxd/storage/backend_lxd.go
roosterfish: Fixed in https://github.com/canonical/lxd/pull/16524.

PR #16166, opened 2025-08-07
init: Ensure the LXD_DIR exists

markylaing:
Ensure this at the very beginning of the Daemon init function

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -435,11 +435,10 @@ jobs:
       - name: Reclaim some memory
         uses: ./.github/actions/reclaim-memory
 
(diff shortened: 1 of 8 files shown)


tomponline:
@markylaing can you also add removal of the lxd-installer package to the github workflows because this manifested itself as in the lxd-ui tests as trying to install the LXD snap by triggering lxd-installer via its socket.

markylaing:
@tomponline @skozina I've removed the `lxd-installer` from any jobs that are running on 24.04 runners (TiCS & UI e2e) as well as moved the `ServerUUID` plus associated logic to `sys.OS`.

PR #16165, opened 2025-08-06
Cluster links identity type

kadinsayani:
Depends on https://github.com/canonical/lxd/pull/16161.

This pull request adds support for a new identity type, `CertificateClusterLink`, and its pending variant. The new identity types are required for https://github.com/canonical/lxd/pull/14884.

PR #16167, opened 2025-08-07
Cluster links identity type follow up

kadinsayani:
Follow up to https://github.com/canonical/lxd/pull/16165.

This pull extends the type interface by adding `CanUpdateCert()` and `IsPending()`, and their implementations. Additionally, query construction is simplified by using a slice to store identity types.


lxd/db/cluster/identities.go
kadinsayani, line 1: I'm not sure about this change. Although it makes it easier to add identity types by leveraging the `Type` interface and `IsPending()`, it moves filtering from the query to Go, making the function less performant. I don't think this change is worth including since we don't add new pending identity types very often.

tomponline: > moves filtering from the query to Go, making the function less performant

No we don't want that indeed

lxd/identities.go
```diff
@@ -1285,8 +1290,8 @@ func updateIdentity(authenticationMethod string) func(d *Daemon, r *http.Request
 			return response.SmartError(err)
 		}
 
-		// Only identities of type api.IdentityTypeCertificateClient may update their own certificate
-		if id.Type != api.IdentityTypeCertificateClient {
+		// Only identities of type api.IdentityTypeCertificateClient and api.IdentityTypeCertificateClusterLink may update their own certificate
```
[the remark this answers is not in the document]
kadinsayani: I'd like to mention the specific identity types for now. If this list grows, we can consider generalizing the comment.

lxd/identities.go
```diff
@@ -1529,7 +1534,7 @@ func patchIdentity(authenticationMethod string) func(d *Daemon, r *http.Request)
 		}
 
 		// Only identities of type api.IdentityTypeCertificateClient and api.IdentityTypeCertificateClusterLink may update their own certificate
```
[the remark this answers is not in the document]
kadinsayani: I'd like to mention the specific identity types for now. If this list grows, we can consider generalizing the comment.

lxd/db/cluster/entity_type_identity.go
```diff
@@ -13,6 +15,39 @@ func (e entityTypeIdentity) code() int64 {
...
+	builder.WriteByte('(')
+	for i, t := range identityTypes {
+		if i > 0 {
+			builder.WriteString(", ")
+		}
+
+		builder.WriteString(strconv.FormatInt(t, 10))
+	}
+
+	builder.WriteByte(')')
```
tomponline: why do you switch to WriteByte here rather than use WriteString?

kadinsayani: I was going for a micro-optimization but after looking closer at their implementations it appears that any performance improvement would be marginal.

lxd/identity/type_interface.go
```diff
@@ -13,13 +13,18 @@ import (
...
+// 5. Add an API type in shared/api/auth.go.
+// 6. Add a new struct that implements this interface.
+// 7. Add a case to [New] for the new identity type.
 type Type interface {
 	// AuthenticationMethod returns the authentication method used by this identity type.
 	AuthenticationMethod() string
 
+	// CanUpdateCert returns true if this identity type supports updating its certificate.
+	CanUpdateCert() bool
+
```
markylaing: This is a bit confusing to me because all non-pending TLS identities can update their certificates, it's just that legacy certificate clients need to use the /1.0/certificates API instead if /1.0/auth/identities/tls.

I think it would be clearer if this function was `LegacyCertificateType() (certificate.Type, error)`

We can then have a blanket check on `PUT/PATCH/DELETE/POST /1.0/auth/identities/tls/{id}` with:
```
idType, err := identity.New(id.Type) 
if err != nil {
}

_, err := idType.LegacyCertificateType()
if err == nil {
    return response.BadRequest(fmt.Errorf("Identities of type %q cannot be modified by this API", id.Type))
}
```

There is also a `toCertificateType` function in `lxd/db/cluster` that it would simplify.

lxd/db/cluster/identities.go
```diff
@@ -444,15 +460,23 @@ func ActivateTLSIdentity(ctx context.Context, tx *sql.Tx, identifier uuid.UUID,
 	return nil
 }
 
-var getPendingTLSIdentityByTokenSecretStmt = fmt.Sprintf(`
+var pendingIdentities = []int64{identityTypeCertificateClientPending, identityTypeCertificateClusterLinkPending}
```
markylaing: Can we instead have a list of all identity types in `identity`, then have the function below iterate over the list to get the identity types that are pending?

Perhaps this one is for another follow up :)

lxd/db/cluster/identities.go
```diff
@@ -151,13 +152,13 @@ func (i *IdentityType) ScanInteger(identityTypeCode int64) error {
...
 // returns an error.
 func (i *IdentityType) Scan(value any) error {
 	return query.ScanValue(value, i, false)
 }
 
-// Value implements driver.Valuer for IdentityType. This converts the API constant into an integer or throws an error.
+// Value implements [driver.Valuer] for [IdentityType]. This converts the API constant into an integer or throws an error.
 func (i IdentityType) Value() (driver.Value, error) {
 	switch i {
 	case api.IdentityTypeCertificateClientRestricted:
```
markylaing: Do you think we could have a `Code` method on the type interface? Then this method would simplify to:
```
idType, err := identity.New(string(i))
if err != nil {
    return nil, err
}

return idType.Code(), nil
}
```

Similarly, the identity package could have a `NewFromCode` method to get an identity type from an int64 code, simplifying the ScanInteger method above

lxd/db/cluster/entity_type_identity.go
```diff
@@ -13,6 +15,39 @@ func (e entityTypeIdentity) code() int64 {
 	return entityTypeCodeIdentity
 }
 
+// Only the following identity types are included here because they are the
+// ones currently supported for use in the identity entity type context.
+// See identity/type_interface.go for details on adding additional identity types.
+var identityTypes = []int64{identityTypeOIDCClient, identityTypeCertificateClient, identityTypeCertificateClientPending, identityTypeCertificateClusterLink, identityTypeCertificateClusterLinkPending}
+
```
markylaing: This is a list of all identity types that are not legacy certificate types. I think this can be simplified by listing all identity types and calling `LegacyCertificateType`, if the error is not nil, include the `Code` in the list.

markylaing (commented):
Making great steps here :rocket:. I think we can go a little further to reduce the number of slices/switch statements that are needed but I'm happy to do this before adding bearer identities if you like :)

lxd/db/cluster/identities.go
tomponline: Lets get it right in this PR and then we can put this to bed.

lxd/identity/type_interface.go
markylaing: Looking at the usage of `CanUpdateCert` I'm actually wrong here. `CanUpdateCert` is basically:
```
// Uses TLS, is fine-grained, and is not pending.
idType.AuthenticationMethod() == api.AuthenticationMethodTLS && idType.IsFineGrained() && !idType.IsPending()
```
So we can alter the call site to do this check instead

I think we should still have `LegacyCertificateType` to simplify the `toCertificateType` method though.

lxd/db/cluster/identities.go
kadinsayani: Great idea, thanks!

lxd/db/cluster/entity_type_identity.go
```diff
@@ -13,6 +16,43 @@ func (e entityTypeIdentity) code() int64 {
...
+		WHEN %d THEN '%s'
+	END`, authMethodTLS, api.AuthenticationMethodTLS, authMethodOIDC, api.AuthenticationMethodOIDC)
+}
+
+// typeInClause returns the SQL IN clause for identity types.
+func (e entityTypeIdentity) typeInClause() string {
+	// Only include identity types that don't have a legacy certificate type.
+	var identityTypes []int64
+	for _, t := range identity.IdentityTypes {
+		_, err := t.LegacyCertificateType()
```
markylaing: I realise this is still draft but I think this should just check for `IsFineGrained`. If it's fine grained it's managed via /auth/identities and is therefore of the `identity` entity type.

lxd/db/cluster/certificates.go
```diff
@@ -113,11 +114,11 @@ SELECT identities.id, identities.auth_method, identities.type, identities.identi
...
-	strconv.Itoa(int(identityTypeCertificateServer)) + `, ` +
-	strconv.Itoa(int(identityTypeCertificateClientUnrestricted)) + `, ` +
-	strconv.Itoa(int(identityTypeCertificateClientRestricted)) + `, ` +
-	strconv.Itoa(int(identityTypeCertificateMetricsUnrestricted)) + `, ` +
-	strconv.Itoa(int(identityTypeCertificateMetricsRestricted)) + `)
+	strconv.Itoa(int(identity.IdentityTypeCertificateServer)) + `, ` +
+	strconv.Itoa(int(identity.IdentityTypeCertificateClientUnrestricted)) + `, ` +
+	strconv.Itoa(int(identity.IdentityTypeCertificateClientRestricted)) + `, ` +
+	strconv.Itoa(int(identity.IdentityTypeCertificateMetricsUnrestricted)) + `, ` +
+	strconv.Itoa(int(identity.IdentityTypeCertificateMetricsRestricted)) + `)
```
markylaing: These are all the identity types that have a valid `LegacyCertificateType`

lxd/db/cluster/entity_type_certificate.go
```diff
@@ -42,11 +44,11 @@ WHERE '' = ?
...
-		identityTypeCertificateClientRestricted,
-		identityTypeCertificateClientUnrestricted,
-		identityTypeCertificateServer,
-		identityTypeCertificateMetricsRestricted,
-		identityTypeCertificateMetricsUnrestricted,
+		identity.IdentityTypeCertificateClientRestricted,
+		identity.IdentityTypeCertificateClientUnrestricted,
+		identity.IdentityTypeCertificateServer,
+		identity.IdentityTypeCertificateMetricsRestricted,
+		identity.IdentityTypeCertificateMetricsUnrestricted,
```
markylaing: And here also, these are all the identities that have a valid `LegacyCertificateType`

lxd/identity/type_interface.go
```diff
@@ -47,6 +47,38 @@ var IdentityTypes = []Type{
...
+	IdentityTypeCertificateClient int64 = 7
+
+	// IdentityTypeCertificateClientPending represents identities for which a token has been issued but who have not yet authenticated with LXD.
+	IdentityTypeCertificateClientPending int64 = 8
+
+	// IdentityTypeCertificateClusterLink represents cluster links that authenticate using TLS and whose permissions are managed via group ownership.
+	IdentityTypeCertificateClusterLink int64 = 9
+
+	// IdentityTypeCertificateClusterLinkPending represents cluster links for which a token has been issued but who have not yet authenticated with a linked LXD cluster.
+	IdentityTypeCertificateClusterLinkPending int64 = 10
```
markylaing: Can these be unexported please

lxd/db/cluster/entity_type_certificate.go
kadinsayani: I'm working on pulling the `typeInClause()` and `authMethodCaseClause()` helper functions up so we can use them here as well.

lxd/identity/type_interface.go
```diff
@@ -33,6 +33,20 @@ type Type interface {
...
+	&CertificateClient{},
+	&CertificateClientPending{},
+	&CertificateClientRestricted{},
+	&CertificateClientUnrestricted{},
+	&CertificateMetricsRestricted{},
+	&CertificateMetricsUnrestricted{},
+	&CertificateServer{},
+	&CertificateClientClusterLink{},
+	&CertificateClientClusterLinkPending{},
+}
```
markylaing, line 102: Why do these need to be references? Also, can we unexport `IdentityTypes` and then have a function `IdentityTypes` that returns a copy of the unexported list? That way we know it can never be modified.

lxd/db/cluster/entity_type_certificate.go
kadinsayani: I'm thinking of having something like this:

```Go
// In entity_type_certificate.go
var identityTypes = func() []int64 {
	var result []int64
	for _, t := range identity.IdentityTypes {
		_, err := t.LegacyCertificateType()
		if err == nil {
			result = append(result, t.Code())
		}
	}

	return result
}

// In stmt.go or a new util.go
// typeInClause returns the SQL IN clause for identity types.
func typeInClause(identityTypes func() []int64) string {
	idTypes := identityTypes()

	if len(idTypes) == 0 {
		return "()"
	}

	var builder strings.Builder
	builder.WriteString("(")
	for i, t := range idTypes {
		if i > 0 {
			builder.WriteString(", ")
		}

		builder.WriteString(strconv.FormatInt(t, 10))
	}

	builder.WriteString(")")
	return builder.String()
}
```

Or, we can have a function in the `Type` interface called `LegacyCertificateTypes()` that returns all identity types with legacy certificates. The same approach can be implemented for fine-grained identity types.

lxd/identity/type_interface.go
```diff
@@ -112,3 +112,34 @@ func New(identityType string) (Type, error) {
...
+		return &CertificateMetricsUnrestricted{}, nil
+	case IdentityTypeCertificateServer:
+		return &CertificateServer{}, nil
+	case IdentityTypeCertificateClusterLink:
+		return &CertificateClientClusterLink{}, nil
+	case IdentityTypeCertificateClusterLinkPending:
+		return &CertificateClientClusterLinkPending{}, nil
+	default:
+		return nil, api.StatusErrorf(http.StatusBadRequest, "Unrecognized identity type code %d", identityTypeCode)
+	}
```
markylaing, line 124: I think even these functions can now be simplified. Let's have something like:
```go
var types = []Type{ /* list of all types */ }
var nameToType = make(map[string]Type, len(types))
var codeToType = make(map[int64]Type, len(types))

func init() {
    for _, t := range types {
        nameToType[t.Name()] = t
        codeToType[t.Code()] = t
    }
}

func New(name string) (Type, error) {
    t, ok := nameToType[name]
    if !ok {
        return nil, api.StatusErrorf(http.StatusBadRequest, "Unrecognized identity type %q", identityType)
    }

    return t, nil
}

func NewFromCode(code int64) (Type, error) {
    t, ok := codeToType[code]
    if !ok {
        // This might need to just be an internal error. This should be used when reading from the database, if we've
        // ever written an invalid type code to the DB then something has gone wrong internally!
        return nil, api.StatusErrorf(http.StatusBadRequest, "Unrecognized identity type code %d", identityTypeCode)
    }

    return t, nil
}
```

lxd/db/cluster/entity_type_certificate.go
markylaing: I think `typeInClause` looks a little complicated. Why does the argument need to be a function? Can it just be the slice? Or a variadic?

Then perhaps create a `[]string` from that via `strconv.FormatInt` and call `strings.Join` on it. This would make it similar to `query.Params`

lxd/identity/type_interface.go
kadinsayani: Great suggestion, thanks :)

lxd/identity/type_interface.go
kadinsayani: They don't need to be references, thanks for pointing that out. Sounds good, I've gone with `Types()` since it will be called as `identity.Types()`.

lxd/db/cluster/entity_type_certificate.go
kadinsayani: I've gone with `...int64`. At the call sites, we have:

```Go
...
var identityTypes = func() (result []int64) {
	for _, t := range identity.Types() {
		if t.IsFineGrained() {
			result = append(result, t.Code())
		}
	}

	return result
}
...
```

Then, `typeInClause()` can be used like this:
```Go
func (e entityTypeIdentity) allURLsQuery() string {
	return fmt.Sprintf(`
SELECT 
	%d, 
	identities.id, 
	'', 
	'', 
	json_array(
		%s,
		identities.identifier
	) 
FROM identities
WHERE type IN %s`,
		e.code(),
		authMethodCaseClause(),
		typeInClause(identityTypes()...),
	)
}
...
```

lxd/identity/type_interface.go
kadinsayani: Thanks for the suggestion. I've added `LegacyCertificateType()` to simplify `toCertificateType()`. I've also updated the checks: https://github.com/canonical/lxd/pull/16167/commits/efb9826956b71adf78f4071503a8925e502c6c9b.

lxd/db/cluster/entity_type_certificate.go
kadinsayani: I've arrived at something simpler that is better aligned with existing conventions throughout the codebase and avoids the addition of `typeInClause()`, please see https://github.com/canonical/lxd/pull/16167/commits/bac503849f33870be50976794ac45e00d4c16e47 :)

lxd/db/cluster/entity_type_identity.go
```diff
@@ -79,3 +82,25 @@ CREATE TRIGGER %s
...
+		return "()"
+	}
+
+	strTypes := make([]string, len(types))
+	for i, t := range types {
+		strTypes[i] = strconv.FormatInt(t, 10)
+	}
+
+	return "(" + strings.Join(strTypes, ", ") + ")"
+}
```
markylaing: This function isn't really about identity types, it returns a string like `(1,2,3)` based on the input. I think it would be worth creating a util for it in the `lxd/db/query` package. It could look like:
```go
func IntParams[T int | int8 | int16 | int32 | int64 | uint | uint8 | uint16 | uint32 | uint64](args ...T) string {
	strs := make([]string, 0, len(args))
	for _, a := range args {
		if a > 0 {
			strs = append(strs, strconv.FormatUint(uint64(a), 10))
			continue
		}

		strs = append(strs, strconv.FormatInt(int64(a), 10))
	}

	return "(" + strings.Join(strs, ", ") + ")"
}
```

lxd/db/cluster/entity_type_identity.go
```diff
@@ -2,13 +2,26 @@ package cluster
...
+	"strings"
 
+	"github.com/canonical/lxd/lxd/identity"
 	"github.com/canonical/lxd/shared/api"
 )
 
 // entityTypeIdentity implements entityTypeDBInfo for an Identity.
 type entityTypeIdentity struct{}
 
+var identityTypes = func() (result []int64) {
```
markylaing: Can this function be defined on the `entityTypeIdentity` struct?

kadinsayani: Sure, I'll define the analogous function in `entity_type_certificate.go` on the `entityTypeCertificate` struct as well.

lxd/db/cluster/entity_type_identity.go
kadinsayani: Good idea, thanks. I've also added unit tests for `Params()` and `IntParams()`.

lxd/db/cluster/entity_type_identity.go
kadinsayani: Actually, I'll keep `certIdentityTypes()` as is to make it easier to use at its other call sites.

lxd/identities.go
```diff
@@ -1268,28 +1273,23 @@ func updateIdentity(authenticationMethod string) func(d *Daemon, r *http.Request
 			return response.NotImplemented(fmt.Errorf("Identities of type %q cannot be modified via this API", id.Type))
 		}
 
+		if identityType.AuthenticationMethod() == api.AuthenticationMethodTLS && identityType.IsPending() {
```
markylaing: Should this check include `&& identityPut.TLSCertificate != ""`?

It looks like this will prevent modification of group membership for pending identities.

lxd/identities.go
```diff
@@ -1506,16 +1506,16 @@ func patchIdentity(authenticationMethod string) func(d *Daemon, r *http.Request)
...
```
markylaing: same here

markylaing: Can you add a small test case for this in the `auth.sh` suite? E.g. 

```
# Pending identities can be added to groups
lxc auth identity create tls/foo
lxc auth identity group add tls/foo admins
lxc auth identity delete tls/foo
```

markylaing (changes requested):
LGTM except for a couple of tiny nits :)

lxd/identities.go
kadinsayani: Nice catch, thanks!

lxd/identities.go
kadinsayani: I've added the test :)

markylaing (dismissed):
I've just rebased my token auth branch onto this and it works an absolute treat! Thank you :partying_face:

kadinsayani:
> I've just rebased my token auth branch onto this and it works an absolute treat! Thank you 🥳

That's great to hear! Glad it's working well. Thanks again for all your feedback and for helping shape the new identity type system :smile:

lxd/db/query/expr.go
```diff
@@ -17,3 +18,19 @@ func Params(n int) string {
 
 	return "(" + strings.Join(tokens, ", ") + ")"
 }
+
+// IntParams returns a parameters expression with the given integer(s).
+// E.g. IntParams(1, 2, 3) -> "(1, 2, 3)". Useful for IN expressions and VALUES expressions.
+func IntParams[T int | int8 | int16 | int32 | int64 | uint | uint8 | uint16 | uint32 | uint64](args ...T) string {
```
tomponline, line 24: cool!

PR #16172, opened 2025-08-07
Revert "github: Remove lxd-installer package."

simondeziel:
This reverts commit d60d56d62749479d836311d24872c6d8b59c5900.

The `system-tests` job still runs on 22.04 (unfortunately) and the UI one should have `lxd-installer` dealt with by the `actions/install-lxd-runtimedeps` since commit e3951e66bc886d4d3fc0da1a72307117dcd77926


tomponline (commented):
The tics step runs on 24.04 though right and the action removing the snap is only on that part?

.github/workflows/tests.yml
```diff
@@ -436,10 +436,11 @@ jobs:
       - name: Reclaim some memory
         uses: ./.github/actions/reclaim-memory
 
-      - name: Remove lxd-installer
+      - name: Purge LXD snap
+        # TODO: drop this when moving away from `ubuntu-22.04` runners to `ubuntu-24.04`
```
tomponline: We don't need to purge the snap as this part runs on 24.04

simondeziel: Right, I fixed the commit message and went a bit further to stop doing this bit of the cleanup because for TiCS, LXD is never executed so the snap or the `lxd-installer` presence shouldn't matter, right?

PR #16176, opened 2025-08-08
Recover: Rearrange `detectUnknownInstanceVolume`

roosterfish:
This prepares the `detectUnknownInstanceVolume` function to be able to cope with volumes using transformed names (PowerFlex & Pure). 
Also it paves the way for further investigation of the instance's backup config file to be able to start scanning attached custom volumes to feed them into the recovery process.

After this one I'll open up another PR which implements the `ListVolumes()` func for both PowerFlex & Pure which will then automatically enable them for recovery.

### DIFF
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -7584,10 +7584,6 @@ func (b *lxdBackend) detectUnknownInstanceVolume(vol *drivers.Volume, projectVol
 		}
 	}
 
-	if instName != backupConf.Instance.Name {
-		return fmt.Errorf("Instance %q in project %q has a different instance name in its backup file (%q)", instName, projectName, backupConf.Instance.Name)
-	}
-
 	apiInstType, err := VolumeTypeToAPIInstanceType(volType)
 	if err != nil {
 		return fmt.Errorf("Failed checking instance type for instance %q in project %q: %w", instName, projectName, err)
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -7651,13 +7651,6 @@ func (b *lxdBackend) detectUnknownInstanceVolume(vol *drivers.Volume, projectVol
 		return fmt.Errorf("Instance %q in project %q already has storage DB record", instName, projectName)
 	}
 
-	// Add to volume to unknown volumes list for the project.
-	if projectVols[projectName] == nil {
-		projectVols[projectName] = []*backupConfig.Config{backupConf}
-	} else {
-		projectVols[projectName] = append(projectVols[projectName], backupConf)
-	}
-
 	// Check snapshots are consistent between storage layer and backup config file.
 	_, err = b.CheckInstanceBackupFileSnapshots(backupConf, projectName, nil)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


roosterfish:
I have the follow up PR to implement PowerFlex's `ListVolume()` ready but its pending proper testing as I still have limited environment access. Therefore this PR is also still in draft.

lxd/storage/backend_lxd.go
```diff
@@ -7531,14 +7531,42 @@ func (b *lxdBackend) ListUnknownVolumes(op *operations.Operation) (map[string][]
 // backup stored on it. It then runs a series of consistency checks that compare the contents of the backup file to
 // the state of the volume on disk, and if all checks out, it adds the parsed backup file contents to projectVols.
 func (b *lxdBackend) detectUnknownInstanceVolume(vol *drivers.Volume, projectVols map[string][]*backupConfig.Config, op *operations.Operation) error {
+	backupYamlPath := filepath.Join(vol.MountPath(), "backup.yaml")
+	var backupConf *backupConfig.Config
+	var err error
+
+	// If the instance is running, it should already be mounted, so check if the backup file
+	// is already accessible, and if so parse it directly, without disturbing the mount count.
+	if shared.PathExists(backupYamlPath) {
```
tomponline: This is what vol.MountTask should do already, so i dont see the need to duplicate the optimisation here?

tomponline: I see this is moved code, so not a blocker, but interesting its here and wonder if it can be removed and the logic simplified.

lxd/storage/backend_lxd.go
tomponline, line 1: I cant see where in this commit the change for:

>  this allows getting the necessary information as their volume names are transformed and don't provide this type of information.

is

tomponline: Oh I see its in the next commit - that prior commit message was misleading in my view.

tomponline (approved):
LGTM thanks!

lxd/storage/backend_lxd.go
roosterfish: I think it is there to differentiate between instances which are already running (we likely assume their `vol.MountTask` was already executed) and instances which are stopped. By checking for the existence of the `backup.yaml` file we can argue that the instance is mounted and must be running.

lxd/storage/backend_lxd.go
roosterfish: I see where this gets confusing. It was more meant as a reason why we are moving this check instead of a description what the actual change is about.

Will keep it in mind, thanks.

lxd/storage/backend_lxd.go
tomponline: Indeed, but vol.MountTask does that for us anyway.

roosterfish: Ah do you mean because the drivers `MountVolumeSnapshot` and `MountVolume` will simply continue on in case the vol is already present? Let me add a new PR for this.

tomponline: yeah, and reference counting would ensure that its not unmounted unless its needed.

One scenario where this might be needed specifically here (and could do with a comment added if true) is the case where the DB was deleted and LXD restarted whilst the instance was still running and thus the mount was kept but the LXD mount counter reference was lost.

This might be why its needed to detect if the instance is already running without needing mount reference counts in LXD.

roosterfish: > One scenario where this might be needed specifically here 

Thanks for the summary, I was just about to comment about this one. In this case the ref counter is decremented to 0 which will cause the volume to be actually unmounted which then crashes the instances. 
I'll add a comment to make this clear.

roosterfish: Fixed in https://github.com/canonical/lxd/pull/16251

PR #16177, opened 2025-08-08
Storage: Ensure Pure volume is unmapped after changing quota

roosterfish:
Compared to PowerFlex, the Pure driver is taking a different approach for unmap after volume size changes. SetVolumeQuota can be invoked in two cases, when the volume is already mapped and when the volume is not mapped. In the latter case the volume should be unmapped again which won't happen if the cleanup hook from getMappedDevPath is only added to the reverter which is cleared right before exiting the function.

### DIFF
--- a/lxd/storage/drivers/driver_pure_volumes.go
+++ b/lxd/storage/drivers/driver_pure_volumes.go
@@ -779,9 +779,6 @@ func (d *pure) GetVolumeUsage(vol Volume) (int64, error) {
 // SetVolumeQuota applies a size limit on volume.
 // Does nothing if supplied with an non-positive size.
 func (d *pure) SetVolumeQuota(vol Volume, size string, allowUnsafeResize bool, op *operations.Operation) error {
-	revert := revert.New()
-	defer revert.Fail()
-
 	// Convert to bytes.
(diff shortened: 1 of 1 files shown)


roosterfish:
@MusicDin is this change viable? I haven't had a chance to test this yet due to lack of access to the env. Please let me know what you think.

MusicDin:
That makes perfect sense. I'll report back once tested

tomponline:
@MusicDin  @roosterfish  any progress here?

MusicDin:
I cannot test it because PureStorage is currently inaccessible. I've escalated the issue, but it has not been resolved yet.

tomponline:
@MusicDin @mihalicyn does this need backporting into stable-5.21 too?

MusicDin:
Yes, it would be very welcome to make sure unmapping is reliable

tomponline:
> Yes, it would be very welcome to make sure unmapping is reliable

Please can you open a PR to cherry-pick this ASAP ready for the LXD 5.21.4 release

tomponline:
@MusicDin ready for cherry pick

tomponline:
@MusicDin please can you cherry-pick this into lxd-pkg-snap latest-candidate

PR #16181, opened 2025-08-11
lxd/idmap/linux: address `golanglint-ci` issues

simondeziel:


lxd/idmap/idmapset_linux.go
```diff
@@ -822,14 +822,8 @@ func getFromProc(fname string) ([][]int64, error) {
 
 	scanner := bufio.NewScanner(f)
 	for scanner.Scan() {
-		// Skip comments
```
tomponline, line 794: Did you check history of this change and why it was added?

tomponline (dismissed):
Thanks!

tomponline:
@simondeziel are you able to reopen this for the cla check?

lxd/idmap/idmapset_linux.go
simondeziel: I had not done it because the files in questions are coming from the `procfs` which contains stuff build dynamically by the kernel. Having a comment show up there would be weird.

That said, I tracked it down to being introduced by commit dc64c84363de76a7a70e957fb2c4be903a275926 in 2017 and appears to have been a copy-n-paste from the `getFromShadow()` function that predates it. The shadow variant deals with files created by humans so comments are possible.

Lastly, I did not include it in the final list of tests but even if comments are somehow present in what the `getFromProc` reads, it has logic to skip bogus lines. I've now reinstate an unlikely test where a comment is present.

tomponline: > and appears to have been a copy-n-paste from the `getFromShadow()` function that predates it.

Ah yeah seems likely indeed cheers

simondeziel:
With that PR, I wanted to squash the 6 integer overflow alerts that CodeQL found in that file. Despite not indicating the PR was achieving this, it seems like it did once merged into `main`: https://github.com/canonical/lxd/security/code-scanning?query=is%3Aclosed+branch%3Amain+path%3Alxd%2Fidmap%2Fidmapset_linux.go+tool%3ACodeQL+rule%3Ago%2Fincorrect-integer-conversion+

We have quite a few other alerts to tend to but slightly less than before :)

PR #16183, opened 2025-08-11
DB: Detect conflicts in generated db code

markylaing:
#14884 proposes to add a conditional unique index to the identities table to ensure that no two cluster links can have the same name. We are considering going a step further and making all TLS identities have a unique name.

The error that is returned by DQLite on a constraint violation is a `driver.Error` whose `Code` is `sqlite3.ErrConstraintUnique`. This is not caught by the current implementation of `SmartError`, and so API responses will typically return status 500 instead of 409.

My initial thoughts were to update `SmartError` to be able to handle constraint violations, however this is not particularly useful because the sqlite error is not very API friendly (e.g. `UNIQUE CONSTRAINT FAILED...`). So instead this PR updates the db generator to check for constraint violations and return an `api.StatusError`. Call sites for create/update/rename functions should use `SmartError` to ensure the correct code is returned.

Additionally, since the database schema should be responsible for defining constraints, I've removed the existence check from all generated `Create` functions. Many API handlers do this anyway. In many cases we're performing the check more than once.

Please note that this morning I've have seen https://github.com/lxc/incus/pull/2365 which does some similar work. This is purely coincidental. I began work on this last week without any knowledge of the PR.

### DIFF
--- a/lxd/db/generate/db/method.go
+++ b/lxd/db/generate/db/method.go
@@ -741,18 +741,8 @@ func (m *Method) create(buf *file.Buffer, replace bool) error {
 		}
 
 		kind := "create"
-		if mapping.Type != AssociationTable {
-			if replace {
-				kind = "create_or_replace"
-			} else {
-				buf.L("// Check if a %s with the same key exists.", m.entity)
-				buf.L("exists, err := %sExists(ctx, tx, %s)", lex.Camel(m.entity), strings.Join(nkParams, ", "))
-				m.ifErrNotNil(buf, true, "-1", "fmt.Errorf(\"Failed to check for duplicates: %w\", err)")
-				buf.L("if exists {")
-				buf.L(`        return -1, api.StatusErrorf(http.StatusConflict, "This \"%s\" entry already exists")`, entityTable(m.entity, m.config["table"]))
-				buf.L("}")
-				buf.N()
-			}
+		if mapping.Type != AssociationTable && replace {
+			kind = "create_or_replace"
 		}
 
 		if mapping.Type == AssociationTable {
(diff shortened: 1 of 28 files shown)


lxd/db/generate/db/method.go
```diff
@@ -792,7 +792,12 @@ func (m *Method) create(buf *file.Buffer, replace bool) error {
 			m.ifErrNotNil(buf, true, "-1", fmt.Sprintf(`fmt.Errorf("Failed to get \"%s\" prepared statement: %%w", err)`, stmtCodeVar(m.entity, kind)))
 			buf.L("// Execute the statement. ")
 			buf.L("result, err := stmt.Exec(args...)")
-			m.ifErrNotNil(buf, true, "-1", fmt.Sprintf(`fmt.Errorf("Failed to create \"%s\" entry: %%w", err)`, entityTable(m.entity, m.config["table"])))
+			buf.L("if err != nil {")
+			buf.L("if query.IsConflictErr(err) {")
+			buf.L(`return -1, api.StatusErrorf(http.StatusConflict, "This \"%s\" entry already exists")`, entityTable(m.entity, m.config["table"]))
+			buf.L("}\n")
+			buf.L(`return -1, fmt.Errorf("Failed to create \"%s\" entry: %%w", err)`, entityTable(m.entity, m.config["table"]))
```
simondeziel, line 789: For consistency, the "Failed to create" should be reordered to align with the rename and update failure messages.

markylaing:
On reflection, this isn't ideal. Returning the `api.StatusError` without wrapping the actual error doesn't allow for deeper inspections further up the call stack. Additionally, any constraints enforced by triggers can have a custom error message that we'd be overwriting here. 

I think the better approach is to go with handling this in SmartError as I mentioned last week.

tomponline:
Why don't you wrap it and return a status error, best of both.

tomponline:
> however this is not particularly useful because the sqlite error is not very API friendly (e.g. `UNIQUE CONSTRAINT FAILED...`).

Agreed. We would prefer entity specific errors returned from api.

markylaing:
> Why don't you wrap it and return a status error, best of both.

Because there is no need. API handlers can call `query.IsConflictError` to check if it's a conflict and return a more API friendly error. Otherwise, `SmartError` will catch it and at least return a 409

markylaing:
Opened again over here https://github.com/canonical/lxd/pull/16198 because github won't let me reopen this one for some reason

Issue #15923, opened 2025-07-02
`powerflex.clone_copy` setting and default value is inconsistent with other drivers' `clone_copy` settings


tomponline:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

ubuntu

### Distribution version

24.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
N/A
```

### Output of "lxc info" or system info if it fails

```shell
N/A
```

### Issue description

The Powerflex driver has a setting called [`powerflex.clone_copy` ](https://documentation.ubuntu.com/lxd/latest/reference/storage_powerflex/#storage-powerflex-pool-conf:powerflex.clone_copy) described as:

> Whether to use non-sparse copies for snapshots
> If this option is set to true, PowerFlex makes a non-sparse copy when creating a snapshot of an instance or custom volume. See [Limitations](https://documentation.ubuntu.com/lxd/latest/reference/storage_powerflex/#storage-powerflex-limitations) for more information.
> Default: true

This is not using a Powerflex "clone" concept (as one does not exist), but rather doing a full raw copy between block volumes using LXD.

In contrast, ZFS and Ceph RBD have their own `clone_copy` settings:

- [ceph.rbd.clone_copy](https://documentation.ubuntu.com/lxd/latest/reference/storage_ceph/#storage-ceph-pool-conf:ceph.rbd.clone_copy) -  "Whether to use RBD lightweight clones: Enable this option to use RBD lightweight clones rather than full dataset copies."
- [zfs.clone_copy ](https://documentation.ubuntu.com/lxd/latest/reference/storage_zfs/#storage-zfs-pool-conf:zfs.clone_copy)- "Whether to use ZFS lightweight clones: Set this option to true or false to enable or disable using ZFS lightweight clones rather than full dataset copies. Set the option to rebase to copy based on the initial image."

Both default to `true` also. 

This makes the Powerflex `clone_copy` setting inconsistent with the other storage drivers because when they are set to `true` it means take a snapshot/lightweight copy whereas with PF it means do a full LXD-side copy, and when the PF setting is set to `false` it means take a snapshot. So the meanings are flipped.

**Proposal:**

Rename Powerflex `clone_copy` to `snapshot_copy` and make its default value `false`, along with a patch to convert any manually specified `clone_copy=false` to `snapshot_copy=true`.

[...]

PR #16186, opened 2025-08-11
Storage: Add new `powerflex.snapshot_copy` replacement config key

roosterfish:
Fixes https://github.com/canonical/lxd/issues/15923.

Adds a new `powerflex.snapshot_copy` pool config key that replaces the `powerflex.clone_copy` key to avoid confusion with the `*.clone_copy` key in the ZFS and Ceph RBD drivers.

Also adds a patch to move all usages of `powerflex.clone_copy` to `powerflex.snapshot_copy`.


tomponline (approved):
Thanks!

PR #16192, opened 2025-08-11
Improve constraint violation detection in `response.SmartError`

markylaing:
This PR improves detection of constraint violations errors so that `SmartError` returns a 409 Conflict.

To do this, the logic used to determine status codes needed to be modified such that `response.Init` can accept functions with signature `func(error) int`. A default function is defined to catch sentinel errors. Additional functions can call `errors.As` to inspect errors more deeply.

### DIFF
--- a/lxd/main.go
+++ b/lxd/main.go
@@ -54,7 +54,7 @@ func (c *cmdGlobal) Run(cmd *cobra.Command, args []string) error {
 	operations.Init(daemon.Debug)
 
 	// Set debug for the response package
-	response.Init(daemon.Debug, db.SmartErrors)
+	response.Init(daemon.Debug, db.SentinelErrors, db.SmartErrFuncs...)
 
 	// Setup logger
 	syslog := ""


lxd/response/smart.go
```diff
@@ -9,51 +9,76 @@ import (
 	"github.com/canonical/lxd/shared/api"
 )
 
+// ErrFunc represents the signature of additional functions that can be passed to Init. This allows deeper error
+// inspections using e.g. [errors.As]. SmartError will evaluate each function in errFuncs, if any return a valid HTTP
+// status code, then that status code will be used (otherwise [http.StatusInternalServerError] is returned).
+type ErrFunc func(error) int
```
roosterfish: Does this `ErrFunc` need to be exported? The callers of `Init` just pass a slice of `func(error) int` so I think you can lowercase this.

lxd/response/response.go
```diff
@@ -26,11 +26,15 @@ import (
...
 var debug bool
 
-// Init sets the debug variable to the provided value and registers any additional smart error mappings.
-func Init(d bool, smartErrors map[int][]error) {
+// Init sets the debug variable to the provided value and registers any additional error mappings. Error mappings are
+// provided as a map of sentinel errors (e.g. those that can be checked with [errors.Is]), where map keys are HTTP
+// status codes, and map values are a slice of sentinel errors that correspond to the status code. Additionally, a list
+// of ErrFunc can be provided. These are functions that return a HTTP code given an error, and are necessary if
+// inspecting error types (e.g. using [errors.As]).
+func Init(d bool, sentinelErrors map[int][]error, additionalErrFuncs ...ErrFunc) {
```
roosterfish: For anybody else wondering about some specifics on `Is` and `As` I  can recommend https://adrianlarion.com/golang-error-handling-demystified-errors-is-errors-as-errors-unwrap-custom-errors-and-more/.

I see there is a need to differentiate between the sentinel errors and custom ones. But should we merge the two variables just to have a more simple interface? 
The `type ErrFunc func(error) int` already returns the status code, so can't we also represent the `sentinelErrors` using this type?

`func Init(d bool, smartErrors ...ErrFunc) {`

E.g.:
```go
IsNotFoundErr := func(err error) int {
    return errors.Is(err, os.ErrNotExist) || errors.Is(err, sql.ErrNoRows)
}
```

I saw in the other commit you have already defined `IsConflictErr`. By doing this you also get another "helper" func that might become handy.

roosterfish (changes requested):
Looks good overall, just a comment on the params passed to `Init` as this was not a clear to me at a first glance.

lxd/response/smart.go
markylaing: Happy to remove altogether and just have a `func(error) int` but this was mainly to have clear documentation about what is expected from the error funcs.

lxd/response/response.go
markylaing: We certainly could do this - this was my first attempt!

However, I realised that the `response` package is used by all the micro* products so I didn't wan't to change the signature of `Init`. With this implementation, all of the given sentinel errors are appended to the `httpErrorCodes` map as before, but the usage of that map is moved into the default ErrFunc.

roosterfish: IIRC its just Microcluster that runs `response.Init` as part of its own `init` func. I would be fine changing this also in Microcluster v3/v2 if we think it's better this way.

lxd/response/smart.go
```diff
@@ -9,51 +9,76 @@ import (
...
-	for httpStatusCode, checkErrs := range httpResponseErrors {
-		for _, checkErr := range checkErrs {
-			if errors.Is(err, checkErr) {
-				if err != checkErr {
-					// If the error has been wrapped return the top-level error message.
-					return &errorResponse{httpStatusCode, err}
-				}
-
-				// If the error hasn't been wrapped, use a generic error.
-				return &errorResponse{httpStatusCode, nil}
```
markylaing, line 38: @roosterfish it turns out we do still need the map response errors. I had missed this logic in my refactor, which returns the generic HTTP status text when a sentinel error is encountered and hasn't been wrapped.

lxd/response/smart.go
markylaing: I've removed the exported `ErrFunc` in favour of just using `func(error) int`

lxd/response/response.go
markylaing: Unfortunately as per https://github.com/canonical/lxd/pull/16192#discussion_r2269870714 I've had to keep this as is. I've also updated the initial commit to keep the logic which tells SmartError to use the generic status text if it encounters a bare (not wrapped) sentinel error.

lxd/response/smart.go
```diff
@@ -9,51 +9,89 @@ import (
...
+// checkSentinelErrors checks all sentinel errors in sentinelErrors and returns the appropriate code if a match is found.
+// Otherwise, it returns zero (which is not a valid HTTP status code, and so indicates that there was no match).
+// Additionally, checkSentinelErrors returns a boolean which indicates if the given error exactly matched the given error
+// (e.g. the error was not wrapped). This is used by SmartError to determine if it should render only the generic HTTP
+// status text.
+func checkSentinelErrors(err error) (code int, generic bool) {
+	for code, errList := range sentinelErrors {
+		for _, sentinelErr := range errList {
+			if errors.Is(err, sentinelErr) {
+				if err == sentinelErr {
```
roosterfish: Not a blocker but you can reduce one level of nesting here by moving the `if err == sentinelErr` check right before checking for `errors.Is`.

lxd/response/smart.go
roosterfish: Mh, this makes me thinking if the list of `func(error) int` could rather be `func(error) (status int, generic bool)`? This way we could still have only this single list of funcs which can be passed to `Init` and its up to the caller to decide to either implement the func using `errors.Is` or `errors.As`. 

Mh but that would also require repeating the `if err == sentinelErr` I believe for every sentinel error.

markylaing: Yeah I'm not sure there is a one-size fits all solution here unfortunately.

roosterfish: Happy to approve it as is. And we don't require any modifications in Microcluster.

roosterfish (dismissed):

lxd/response/smart.go
markylaing: Fixed

lxd/response/smart.go
markylaing: Thanks, I've made one more change to make it more flexible:

The initialisation and handling of sentinel errors is kept as is. Now we'll have a list of functions with `func(err) (code int, msg string)` allowing the registered error matchers to overwrite the error message sent over the API. This could be useful if we figure out a way to automatically map constraint violations to API friendly messages.

nmezhenskyi (approved):
LGTM!

lxd/db/storage_buckets.go
```diff
@@ -300,9 +298,8 @@ func (c *ClusterTx) CreateStoragePoolBucket(ctx context.Context, poolID int64, p
 		VALUES (?, ?, ?, ?, (SELECT id FROM projects WHERE name = ?))
 		`, poolID, nodeID, info.Name, info.Description, projectName)
 	if err != nil {
-		var dqliteErr dqliteDriver.Error
 		// Detect SQLITE_CONSTRAINT_UNIQUE (2067) errors.
-		if errors.As(err, &dqliteErr) && dqliteErr.Code == 2067 {
```
tomponline, line 305: Is sqlite3.ErrConstraintUnique the same as `dqliteErr.Code == 2067`?

lxd/response/smart.go
```diff
@@ -9,51 +9,83 @@ import (
 	"github.com/canonical/lxd/shared/api"
 )
 
-var httpResponseErrors = map[int][]error{
+// sentinelErrors contains some default mappings of HTTP status codes to sentinel errors that are defined in the Go
+// standard library. Additional errors can be added to this map on a call to Init.
+var sentinelErrors = map[int][]error{
```
tomponline, line 14: Rather than having to define sentinel errors separately from `additionalErrFuncs` used by `Init` above it would seem preferable to me to model error to status code mapping problem in a consistent way by making this map something like:

```go
var httpResponseErrors = map[int][]func(error) bool
```

Then for sentinel errors we can define an entry as follows:

```
http.StatusForbidden: {func() {return checkSentinelErrors(os.ErrNotExist, sql.ErrNoRows)}},
```

Then we can have a single implementation that iterates over `httpResponseErrors` calling the function(s) defined for that status code and if any one returns true we return that status code and message from the error if present.

lxd/db/storage_buckets.go
markylaing: Yes. The dqlite.Error codes are the codes from sqlite

lxd/response/smart.go
markylaing: Did you see these threads about this:
- https://github.com/canonical/lxd/pull/16192#discussion_r2269143003
- https://github.com/canonical/lxd/pull/16192#discussion_r2269870714

Happy to make the change, just want to double check you have all the context :)

tomponline: OK think we need a meeting to discuss this as i dont understand why we cant do this because of MicroCloud/cluster. It sounds like MicroCloud/cluster needs to move away from using LXD's response package.

markylaing: Basically to keep the same behaviour regarding sentinel errors, then all functions that are passed into `response.Init` need to check if the given error is an exact match, and if so return the generic http.StatusText.

markylaing: This ends up like:
```go
// lxd/response/smart.go

var httpResponseErrors = map[int][]func(error) (bool, string){
	http.StatusNotFound:  {checkSentinelError(os.ErrNotExist, http.StatusNotFound), checkSentinelError(sql.ErrNoRows, http.StatusNotFound)},
	http.StatusForbidden: {checkSentinelError(os.ErrPermission, http.StatusForbidden)},
}

// checkSentinelErrors checks all sentinel errors in sentinelErrors and returns the appropriate code if a match is found.
// Otherwise, it returns zero (which is not a valid HTTP status code, and so indicates that there was no match).
// Additionally, checkSentinelErrors returns a string message which overwrites the API error message contents if non-empty.
func checkSentinelError(targetErr error, code int) func(checkErr error) (isMatch bool, msg string) {
	return func(checkErr error) (isMatch bool, msg string) {
		if checkErr == targetErr {
			return true, http.StatusText(code)
		}

		return errors.Is(checkErr, targetErr), ""
	}
}

// lxd/db/errors.go

var SmartErrors = map[int][]func(error) (bool, string){
	http.StatusConflict: {
		func(err error) (bool, string) {
			return query.IsConflictErr(err), ""
		},
	},
	http.StatusServiceUnavailable: {
		func(err error) (bool, string) {
			if err == driver.ErrNoAvailableLeader {
				return true, http.StatusText(http.StatusServiceUnavailable)
			}

			return errors.Is(err, driver.ErrNoAvailableLeader), ""
		},
	},
}
```

markylaing: > OK think we need a meeting to discuss this as i dont understand why we cant do this because of MicroCloud/cluster. It sounds like MicroCloud/cluster needs to move away from using LXD's response package.

Microcluster isn't the problem. The problem is the existing behaviour of `SmartError` for sentinel errors, where if the sentinel error is an exact match, then the generic http.StatusText is returned instead of `err.Error()`.

If we're happy to get rid of this, then I'm happy to make the change.

tomponline: lets have a meeting as im lacking too much content i think

markylaing:
Closing for now. We will add conflict detection in the database generator which will return an api.StatusError with a 409.

Calling functions can return wrap the error or return a more appropriate error message by checking with e.g. `api.StatusErrorCheck`.

I'll add a PR to remove the `sqlite.ErrConstraintUnique` from `lxd/db/errors.go`.

PR #16198, opened 2025-08-11
DB: Improvements to generated code

markylaing:
All generated DB methods accept a `context.Context` but not all were using it. Additionally, when #16192 is merged, a 409 will be automatically returned by SmartError if one was encountered.

If an API handler does not check for the existence of an entity before creating a new one, it should check the error with `query.IsConflictErr` (from #16192) and return a more API friendly error message.


markylaing:
Before continuing with this one I'll rebase onto #16192 and audit entity creation to ensure we're catching constraint violations in handlers and returning a sensible error. Otherwise we'll start seeing a lot of `UNIQUE constraint failed: ...` when creating resources.

markylaing:
I have updated this PR to add conflict detection to all handlers that were missing it, and tests for these cases.

This PR now depends on https://github.com/canonical/lxd/pull/16192. So will wait for that to be merged before rebasing and requesting review here.

PR #16205, opened 2025-08-11
Auth: Enforce unique identity names (except for OIDC)

markylaing:
This PR enforces name uniqueness per-authentication method for non-OIDC identities.

There are a number of reasons for this:
1. If two identities have the same name but different certificate fingerprints, it's not possible for an administrator to reason about which one belongs to who (unless they are of different identity types).
2. The `/1.0/auth/identities/{method}/{nameOrID}` route accepts a name only if that name is unique. Otherwise it requires the full fingerprint for TLS identities, or the email for OIDC identities.
3. We're increasingly aware that we need some form of certificate rotation, both within cluster and between clusters (links). For this we may need to have multiple certificates per identity.

The drawback of enforcing name uniqueness for all identities using the same authentication method is that it will not be possible for e.g. a cluster link identity to have the same name as a client certificate. I'm open to just enforcing this per-identity type, but we'll need to be careful to enforce it for both pending and active types.

For the implementation I have chosen to go with triggers on the `identities` table on `INSERT` and `UPDATE`. This is because creating a conditional unique index as part of a schema upgrade is problematic:
1. The index can't be created if any duplicate names already exist.
2. If duplicate names exist, we can't rename them because a) we should avoid modifying existing deployments as much as possible, b) it's not clear what we'd rename them to, and c) even if we did rename them, we still can't create the index until the transaction is committed (requiring two schema upgrades).

Using triggers will disallow creation of new identities with duplicate names, but will not modify any existing ones. Additionally, it has the benefit that a custom error message can be returned.

### DIFF
--- a/lxd/db/cluster/entity_type_identity.go
+++ b/lxd/db/cluster/entity_type_identity.go
@@ -9,7 +9,9 @@ import (
 )
 
 // entityTypeIdentity implements entityTypeDBInfo for an Identity.
-type entityTypeIdentity struct{}
+type entityTypeIdentity struct {
+	entityTypeCommon
+}
 
 // identityTypes returns the list of identity type codes that are considered fine-grained.
 func (e entityTypeIdentity) identityTypes() (types []int64) {
@@ -42,10 +44,6 @@ WHERE type IN %s`,
 		e.code(),
 		authMethodCaseClause(),
 		query.IntParams(e.identityTypes()...))
-}
-
-func (e entityTypeIdentity) urlsByProjectQuery() string {
-	return ""
 }
 
 func (e entityTypeIdentity) urlByIDQuery() string {
--- a/lxd/db/cluster/entity_type_identity.go
+++ b/lxd/db/cluster/entity_type_identity.go
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/canonical/lxd/lxd/db/query"
 	"github.com/canonical/lxd/lxd/identity"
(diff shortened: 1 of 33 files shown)


lxd/db/cluster/stmt.go
```diff
@@ -67,20 +69,41 @@ func StmtString(code int) (string, error) {
...
+
+		return nil
+	}
+
+	for entityType, entityTypeInfo := range entityTypes {
+		name, stmt := entityTypeInfo.onDeleteTriggerSQL()
+		err := applyTrigger(name, stmt, entityType)
+		if err != nil {
+			return err
+		}
```
kadinsayani, line 97: If an entity type returns a trigger where either `name` or `stmt` are empty, `applyTriggers()` will return an error early and subsequent triggers won't be applied. Rather than returning an error, should we collect or log errors and continue iterating over `entityTypes`?

markylaing: I don't think so, `applyTriggers` is called when the database is opened and ideally LXD will fail to start if it can't apply them. Then we're never in danger of passing tests while these errors are lost in all the logs.

kadinsayani: Won't all of the entity types with the common implementations of `onInsertTriggerSQL()` and `onUpdateTriggerSQL()` fail since they return empty strings?

kadinsayani: Ah I just saw the first check in `applyTriggers()` which returns `nil`.

kadinsayani (approved):
LGTM, thanks!

lxd/db/cluster/stmt.go
markylaing: Yeah they need to define both the name and the trigger contents for a trigger to be created. If neither are defined, no trigger is created. I added a comment about this to the `entityTypDBInfo` interface but let me know if I can make it any clearer.

lxd/db/cluster/entity_type_identity.go
```diff
@@ -88,3 +89,33 @@ func authMethodCaseClause() string {
...
+
+// onInsertTriggerSQL enforces that newly created identities whose authentication method is not OIDC have a unique name.
+func (e entityTypeIdentity) onInsertTriggerSQL() (name string, sql string) {
+	name = "on_identity_insert"
+	sql = `CREATE TRIGGER ` + name + `
+	BEFORE INSERT ON identities
+	WHEN NEW.auth_method != ` + strconv.FormatInt(authMethodOIDC, 10) + `
+		AND (SELECT COUNT(*) FROM identities WHERE name = NEW.name AND auth_method = NEW.auth_method) > 0
+	BEGIN
+		SELECT RAISE(ABORT, 'An identity with this name and authentication method already exists');
```
tomponline, line 102: This error "name and authentication method already exists" doesn't exactly align with the statement:

> enforces that newly created identities whose authentication method is not OIDC have a unique name.

As the statement (and the trigger itself) seem to make name unique across all authentication methods (except OIDc).

tomponline (commented):
Just one comment about the triggers.

lxd/db/cluster/entity_type_identity.go
markylaing: `SELECT COUNT(*) FROM identities WHERE name = NEW.name AND auth_method = NEW.auth_method`

returns the number of identities with the same name and authentication method as the identity being created. The trigger runs when this number is greater than zero, so this only makes identity names unique within the authentication method.

tomponline: thanks, the comment then is still misleading:

> enforces that newly created identities whose authentication method is not OIDC have a unique name.

should be:

> enforces that newly created identities have a unique name within the authentication method (where method is not OIDC)

tomponline (changes requested):
Just a nit to clarify the comment please.

lxd/db/cluster/entity_type_identity.go
markylaing: Fixed :)

lxd/db/cluster/entity_type_identity.go
```diff
@@ -90,3 +89,33 @@ func authMethodCaseClause() string {
...
```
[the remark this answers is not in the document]
markylaing: This isn't needed because the error message from the trigger isn't wrapped in generated `CreateIdentity` functions and passed up to the API handlers. This is because general constraint violation errors are not very API friendly (e.g. `UNIQUE constraint failed: identities.identifier, identities.auth_method`), so a generalised error is returned with a 409.

lxd/db/cluster/entity_type_identity.go
```diff
@@ -90,3 +89,33 @@ func authMethodCaseClause() string {
...
+// identities using that authentication method). This trigger only runs if the name of the identity is being changed,
+// this allows pre-existing identities with duplicated names to continue to update normally.
+func (e entityTypeIdentity) onUpdateTriggerSQL() (name string, sql string) {
+	name = "on_identity_update"
+	sql = `CREATE TRIGGER ` + name + `
+	BEFORE UPDATE ON identities
+	WHEN OLD.name != NEW.name AND NEW.auth_method != ` + strconv.FormatInt(authMethodOIDC, 10) + `
+		AND (SELECT COUNT(*) FROM identities WHERE name = NEW.name AND auth_method = NEW.auth_method) > 0
+	BEGIN
+		SELECT RAISE(ABORT, 'An identity with this name and authentication method already exists');
```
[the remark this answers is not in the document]
markylaing: As in the comment above, this error message is never seen by the end user at the moment.

markylaing:
> Just a nit to clarify the comment please.

Fixed :)

tomponline (approved):

PR #16210, opened 2025-08-12
lxd/storage: Improve error message in `CreateCustomVolumeFromBackup()`

skozina:
Add the same checks in `CreateCustomVolumeFromBackup()` like we have in `CreateCustomVolumeFromISO()`.
This changes the error message in case the volume already exists from:
```
# lxc storage volume import default backup2.tar.gz backup
Error: Create custom volume from backup: Error inserting volume "backup" for project "default" in pool "default" of type "custom" into database "UNIQUE constraint failed: index 'storage_volumes_unique_storage_pool_id_node_id_project_id_name_type'"
```
to
```
# lxc storage volume import default backup2.tar.gz backup
Error: Create custom volume from backup: Cannot create volume, already exists on target storage
```


lxd/storage/backend_lxd.go
```diff
@@ -8234,6 +8234,15 @@ func (b *lxdBackend) CreateCustomVolumeFromBackup(srcBackup backup.Info, srcData
 
 	vol := b.GetNewVolume(drivers.VolumeTypeCustom, drivers.ContentType(customVol.ContentType), volStorageName, customVol.Config)
 
+	volExists, err := b.driver.HasVolume(vol)
+	if err != nil {
+		return err
+	}
+
+	if volExists {
+		return errors.New("Cannot create volume, already exists on target storage")
```
roosterfish: Worth adding the name of the volume to the error message? Imagine you export the custom vol, by default the name of the downloaded archive is `backup.tar.gz`. Now when importing using `lxc storage volume import <pool> backup.tar.gz`, you might wonder what is the name which is causing the conflict. 
Because providing the name of the resulting vol is optional. 
By default it takes what is shipped with the backup inside the config file.

roosterfish (changes requested):

lxd/storage/backend_lxd.go
skozina: Good point. I've changed the error message for both the backup import and ISO import (where it's not necessary, as the volume name must be provided in the request) to:
```
root@test:~# lxc storage volume import default backup2.tar.gz backup
Error: Create custom volume from backup: Cannot create volume "backup", volume already exists on storage pool "default"
```

roosterfish (approved):

tomponline:
Thanks!

PR #16211, opened 2025-08-12
Import tarballs as LXD volumes

skozina:


tomponline:
Can be rebased

skozina:
> Can be rebased

Rebased. Thanks for the merges!

client/lxd_storage_volumes.go
```diff
@@ -1065,13 +1065,7 @@ func (r *ProtocolLXD) GetStoragePoolVolumeBackupFile(pool string, volName string
...
 }
 
-// CreateStoragePoolVolumeFromISO creates a custom volume from an ISO file.
-func (r *ProtocolLXD) CreateStoragePoolVolumeFromISO(pool string, args StoragePoolVolumeBackupArgs) (Operation, error) {
-	err := r.CheckExtension("custom_volume_iso")
-	if err != nil {
-		return nil, err
-	}
-
+func (r *ProtocolLXD) createStoragePoolVolumeFromFile(pool string, args StoragePoolVolumeBackupArgs, xLxdType string) (Operation, error) {
```
roosterfish: `volumeType` is already used in the other client funcs to differentiate between LXD's volume type. But should this variable be named something like `contentType` rather than `xLxdType`? For example there is [`ContentTypeISO`](https://github.com/canonical/lxd/blob/main/lxd/storage/drivers/volume.go#L75) const also using this nomenclature.

client/lxd_storage_volumes.go
```diff
@@ -1133,6 +1133,20 @@ func (r *ProtocolLXD) CreateStoragePoolVolumeFromISO(pool string, args StoragePo
 	return r.createStoragePoolVolumeFromFile(pool, args, "iso")
 }
 
+// CreateStoragePoolVolumeFromTar creates a custom volume from an ISO file.
```
roosterfish: Comment still says `ISO`.

lxd/storage_volumes.go
```diff
@@ -1007,11 +1007,14 @@ func storagePoolVolumesPost(d *Daemon, r *http.Request) response.Response {
 
 	// If we're getting binary content, process separately.
 	if r.Header.Get("Content-Type") == "application/octet-stream" {
-		if r.Header.Get("X-LXD-type") == "iso" {
+		switch r.Header.Get("X-LXD-type") {
+		case "iso":
 			return createStoragePoolVolumeFromISO(s, r, requestProjectName, projectName, r.Body, poolName, r.Header.Get("X-LXD-name"))
+		case "tar":
+			return createStoragePoolVolumeFromTarball(s, r, requestProjectName, projectName, r.Body, poolName, r.Header.Get("X-LXD-name"))
```
roosterfish, line 1025: The corresponding client func is using `CreateStoragePoolVolumeFromTar`. Should this be consistent and either say "Tar" or "Tarball"?

lxd/storage_volumes.go
```diff
@@ -2540,6 +2543,67 @@ func createStoragePoolVolumeFromISO(s *state.State, r *http.Request, requestProj
 	return operations.OperationResponse(op)
 }
 
+func createStoragePoolVolumeFromTarball(s *state.State, r *http.Request, requestProjectName string, projectName string, data io.Reader, poolName string, volName string) response.Response {
+	revert := revert.New()
+	defer revert.Fail()
+
+	if poolName == "" {
```
roosterfish: Is this check required for a specific purpose? Just asking because I haven't found it in the other two funcs for backups and ISOs.

lxd/storage/backend_lxd.go
```diff
@@ -8173,6 +8174,154 @@ func (b *lxdBackend) CreateCustomVolumeFromISO(projectName string, volName strin
...
+// CreateCustomVolumeFromTarball creates a custom volume from the given backup info.
+func (b *lxdBackend) CreateCustomVolumeFromTarball(projectName string, volName string, srcData io.ReadSeeker, op *operations.Operation) error {
+	l := b.logger.AddContext(logger.Ctx{"project": projectName, "volume": volName})
+	l.Debug("CreateCustomVolumeFromTarball started")
+	defer l.Debug("CreateCustomVolumeFromTarball finished")
+
+	// Define function to unpack a volume from a backup tarball file.
+	unpackTarball := func(r io.ReadSeeker, tarArgs []string, unpacker []string, mountPath string) error {
+		// Prepare tar arguments.
+		args := append(tarArgs, []string{
```
roosterfish: IIRC the `createStoragePoolVolumeFromBackup` is extracting the archive before calling the driver specific func and instead passes the "extracted" `io.ReadSeeker`. Is there a specific reason for doing this differently here?

roosterfish: The args look similar to the ones in `archive.Unpack`. Can you use this one instead?

roosterfish (changes requested):
Looking good already. Only a few minor comments especially about reuse of the already existing `archive` package.

Also I was wondering about the use case of this. Is it meant to upload and unpack arbitrary tar archives into custom volumes? In your tests you are using an archive containing an ISO image but I was wondering if there is another common pattern where this might be used.
Whilst thinking about it I wondered you could potentially already do this today by creating a custom volume, attaching it to an instance and use `lxc file push` to upload the contents into the respective instance. But of course you would need to do the unpacking first and you have to transfer potentially more data.

client/lxd_storage_volumes.go
skozina: Hm... `volumeType` is the LXD volume type, eg. `container`, `image` or `custom` in this case.
`contentType` might mean either the HTTP Content-Type (so `application/octet-stream` in this case), but here it rather means the Content-type of the LXD storage volume (so `ISO`, or `filesystem` in this case). Already confusing!;-)
The `X-LXD-type` feels different from either of these, it differentiates between the various types of the import files (so `iso` or `tar` in this case).
I've renamed this to `fileType` for now, WDYT?

client/lxd_storage_volumes.go
skozina: Thanks! Fixed.

lxd/storage_volumes.go
skozina: Yes, `...fromTarball()` is better naming. Thanks! Renamed.

lxd/storage_volumes.go
skozina: You are right, the HTTP request parsing wouldn't match the right endpoint if pool name is not provided in the URL (`/1.0/storage-pools//volumes/custom`). So there's no point in checking if the pool name is not empty in the handler itself. Thanks, I've removed this.

lxd/storage/backend_lxd.go
skozina: > IIRC the `createStoragePoolVolumeFromBackup` is extracting the archive before calling the driver specific func and instead passes the "extracted" `io.ReadSeeker`. Is there a specific reason for doing this differently here?

It works a little differently. The `createStoragePoolVolumeFromBackup()` detects the compression algorithm, but it doesn't fully unpack the tar archive. By calling `backup.GetInfo()`->`TarReader()`, it only decompresses the archive in case it's compressed with gzip or similar. The uncompressed content is then piped to tar reader to find the content of the `backup/index.yaml` metadata file, otherwise is discarded. The ultimate extraction of the archive is done by the driver specific func, but all drivers eventually end up calling `genericVFSBackupUnpack()`, as the tar archive is unoptimized source storage. Yes, this means that we end up de-compressing the archive twice, once in the `GetInfo()` by calling gzip/bzip directly, and then again in `genericVFSBackupUnpack()` by calling `tar -z`, `tar -j` or similar.

The code in this PR is indeed inspired by the `genericVFSBackupUnpack()` so it should work quite similar.

skozina: > The args look similar to the ones in `archive.Unpack`. Can you use this one instead?

Yes! Thank you for this tip, this made the code much nicer;-) Done.

skozina:
> Looking good already. Only a few minor comments especially about reuse of the already existing `archive` package.

Thank you for that tip! Yes, using `archive.Unpack()` I think made this much nicer.

Thank you for the review!

client/lxd_storage_volumes.go
roosterfish: Yeah I see the conflict with the `Content-Type` header. `fileType` sounds good!

lxd/storage/drivers/utils.go
```diff
@@ -37,8 +37,8 @@ const MinBlockBoundary = 8192
 // blockBackedAllowedFilesystems allowed filesystems for block volumes.
 var blockBackedAllowedFilesystems = []string{"btrfs", "ext4", "xfs"}
 
-// wipeDirectory empties the contents of a directory, but leaves it in place.
-func wipeDirectory(path string) error {
+// WipeDirectory empties the contents of a directory, but leaves it in place.
+func WipeDirectory(path string) error {
```
roosterfish: Is this commit maybe a leftover? I cannot see where in the other changes `WipeDirectory` is used which would explain exporting it.

client/lxd_storage_volumes.go
skozina: Cool, thanks!

lxd/storage/drivers/utils.go
skozina: Oh yes, it is a leftover. I've dropped this commit, thank you!
I was trying to wipe the mounted directory before we extract the tar content in there to see if the CodeQL security checks would be happier, as that's something we do in the backup extract case (in `genericVFSBackupUnpack()`). The idea to use `archive.Unpack()` changed that code completely, and I forgot to drop the export commit. Thank you!

roosterfish:
Regarding the failing CodeQL check, I see the `lxd/backup` package already has a `ValidateBackupName` which maybe could be made generic so it can also be used here? I think the other CodeQL checks for all the storage drivers are just an outcome of having this new tar func which might call each of the drivers.

skozina:
> Regarding the failing CodeQL check, I see the `lxd/backup` package already has a `ValidateBackupName` which maybe could be made generic so it can also be used here? I think the other CodeQL checks for all the storage drivers are just an outcome of having this new tar func which might call each of the drivers.

The new code uses the `X-LXD-name` header to provide target volume name. We validate it with `drivers.ValidVolumeName()`, pretty much the same as we do for ISO or backup imports, and this validation is stricter than the one done by `ValidateBackupName()`. Here: https://github.com/canonical/lxd/pull/16211/files#diff-029fdd0e1afdacb0e66630c0d7e813772122db9a7d036060902838db315d9c4dR8184

I am not much sure what more we can do, but I'm hesitant to just dismiss these warnings, after all we *do* create a custom volume and mount it based on a external provided input value.

skozina:
I've tried two more options to please the CodeQL X-LXD-name header validation failures:
 - overwrite the volume name local variable with the validated version of the input
 - call the validation function directly in the API handler function

None of this really silenced the warnings;-/ I think we should review the code manually, and dismiss the warnings based on the review. Thanks for all the review feedback!

tomponline:
> ```
> $ lxc storage volume import $POOL file.tar $VOLUME --type=tar
> ```

Does this only support uncompressed tarballs or this that handled?

tomponline:
@skozina apologies if I've already asked this somewhere before, but should we also support raw image files for block type volumes?

lxc/storage_volume.go
```diff
@@ -2843,7 +2843,7 @@ func (c *cmdStorageVolumeImport) command() *cobra.Command {
 		Create a new custom volume using backup0.tar.gz as the source.`))
 	cmd.Flags().StringVar(&c.storage.flagTarget, "target", "", i18n.G("Cluster member name")+"``")
 	cmd.RunE = c.run
-	cmd.Flags().StringVar(&c.flagType, "type", "", i18n.G("Import type, backup or iso (default \"backup\")")+"``")
+	cmd.Flags().StringVar(&c.flagType, "type", "", i18n.G("Import type: backup, iso or tar (default \"backup\")")+"``")
```
tomponline: Can we expand this help text to differentiate a "backup" tarball from a plain tarball?

tomponline:
@skozina please can you rebase, ta

skozina:
> Does this only support uncompressed tarballs or this that handled?

The implementation relies on `shared.DetectCompression()`: https://github.com/canonical/lxd/blob/ed290bd7cf386745df16f4951353c3ff797171cd/shared/archive.go#L11

This means that all currently recognized compression formats should work:
```
# file content.tar.gz 
content.tar.gz: gzip compressed data, from Unix, original size modulo 2^32 10240
# lxc storage volume import default content.tar.gz vol --type=tar
```

skozina:
> @skozina apologies if I've already asked this somewhere before, but should we also support raw image files for block type volumes?

There are two possible extensions to this PR:
 - support raw image files to import as block type volumes
 - support for custom volume export as a tarball (the opposite operation to the one implemented here)

I'd prefer not to implement any of these until a specific use is requested, just to keep the implementation simpler.

tomponline:
> * support raw image files to import as block type volumes

Agree with not implementing now, but would like us to consider the possibility of adding it in the future to the current design so we don't inadvertently block it in a distasteful way. 

What I mean is today `--type=iso` implies the equivalent of `lxc storage volume create --type=block`, but `--type=tar` could mean either `lxc storage volume create  --type=filesystem` or `--type=block`.

So how would we support block import in the future?

Would we perhaps support a `--type=raw` on `lxc storage volume import` that treats the input file as a raw block file (and implies `lxc storage volume create  --type-block`)?

lxc/storage_volume.go
skozina: Sure. I've updated the help message so now it looks like:
```
Flags:
      --target        Cluster member name
      --type=backup   Import file is a custom volume backup (default option)
      --type=iso      Import file is an iso image, will be imported as iso volume
      --type=tar      Import file is a tarball, will be imported as custom volume
```

tomponline: > Import file is a tarball, will be imported as custom volume

Import file is a tarball, will be imported as custom filesystem volume

skozina:
> > * support raw image files to import as block type volumes
> 
> What I mean is today `--type=iso` implies the equivalent of `lxc storage volume create --type=block`, but `--type=tar` could mean either `lxc storage volume create --type=filesystem` or `--type=block`.

Today, `storage volume import --type=tar` creates custom storage volume with content-file=filesystem (so yes, analogous to `storage volume create --type=filesystem`).
 
> So how would we support block import in the future?
> 
> Would we perhaps support a `--type=raw` on `lxc storage volume import` that treats the input file as a raw block file (and implies `lxc storage volume create --type-block`)?

Yes, I was thinking that the `--type=raw` would be a new option, which would create custom storage volume with `content-type=block` (analogous to `storage volume create --type=block`).

I don't think there's much need to put raw images files into tarballs, or even compress those, so the input file type would be directly the raw image (`lxc storage volume import $pool $raw_file_image.bin --type=raw`).

tomponline:
> > > * support raw image files to import as block type volumes
> > 
> > 
> > What I mean is today `--type=iso` implies the equivalent of `lxc storage volume create --type=block`, but `--type=tar` could mean either `lxc storage volume create --type=filesystem` or `--type=block`.
> 
> Today, `storage volume import --type=tar` creates custom storage volume with content-file=filesystem (so yes, analogous to `storage volume create --type=filesystem`).
> 
> > So how would we support block import in the future?
> > Would we perhaps support a `--type=raw` on `lxc storage volume import` that treats the input file as a raw block file (and implies `lxc storage volume create --type-block`)?
> 
> Yes, I was thinking that the `--type=raw` would be a new option, which would create custom storage volume with `content-type=block` (analogous to `storage volume create --type=block`).
> 
> I don't think there's much need to put raw images files into tarballs, or even compress those, so the input file type would be directly the raw image (`lxc storage volume import $pool $raw_file_image.bin --type=raw`).

Cool in that case we have a plan, thanks!

lxc/storage_volume.go
skozina: I've changed the help message again, now it looks like:
```
Flags:
      --target   Cluster member name
      --type     Type of the import file. Valid options are:
                 - backup: custom volume backup (default option)
                 - iso: iso image, will be imported as iso volume
                 - tar: tarball, will be imported as custom filesystem volume
```

client/lxd_storage_volumes.go
```diff
@@ -1123,65 +1119,33 @@ func (r *ProtocolLXD) CreateStoragePoolVolumeFromISO(pool string, args StoragePo
...
-func (r *ProtocolLXD) CreateStoragePoolVolumeFromBackup(pool string, args StoragePoolVolumeBackupArgs) (Operation, error) {
-	err := r.CheckExtension("custom_volume_backup")
+// CreateStoragePoolVolumeFromISO creates a custom volume from an ISO file.
+func (r *ProtocolLXD) CreateStoragePoolVolumeFromISO(pool string, args StoragePoolVolumeBackupArgs) (Operation, error) {
+	err := r.CheckExtension("custom_volume_iso")
 	if err != nil {
 		return nil, err
 	}
 
-	if args.Name != "" {
```
tomponline, line 1133: Is this not relevant also for ISO uploads?

client/lxd_storage_volumes.go
```diff
@@ -1133,6 +1133,20 @@ func (r *ProtocolLXD) CreateStoragePoolVolumeFromISO(pool string, args StoragePo
 	return r.createStoragePoolVolumeFromFile(pool, args, "iso")
 }
 
+// CreateStoragePoolVolumeFromTarball creates a custom volume from a tarball.
```
tomponline: > // CreateStoragePoolVolumeFromTarball creates a custom filesystem volume from a tarball.

lxc/storage_volume.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/spf13/cobra"
 	"go.yaml.in/yaml/v2"
 
-	"github.com/canonical/lxd/client"
+	lxd "github.com/canonical/lxd/client"
```
tomponline: whoops

client/lxd_storage_volumes.go
skozina: I'm not sure I understand your comment.
The volume name argument is mandatory for ISO and Tarball file imports, while it's optional for backup file imports (because the volume name is contained in the backup metadata).

The diff is a bit hard to read due to refactoring. The resulting code contains this mandatory check in `CreateStoragePoolVolumeFromISO()` and `CreateStoragePoolVolumeFromTarball()` functions:
```
	if args.Name == "" {
		return nil, errors.New("Missing volume name")
	}
```

client/lxd_storage_volumes.go
skozina: Thanks, added.

lxc/storage_volume.go
skozina: Ugh. I must have fixed it already, I don't see it in my current local branch. I'll push and ensure this is gone. Thanks for catching this!

skozina: Yup, it's gone.

skozina:
Another change to this PR is that various device nodes (such as files under `dev/*`) are now not filtered out from the tarball file import:
https://github.com/canonical/lxd/pull/16211/files#diff-00ff72430f4f6a2dcf590fdf4192333322a98893eeb23e5445a4bb9d10949ebdR141

tomponline (approved):
Thanks!

tomponline:
@skozina LGTM, please can you rebase

skozina:
Thanks! Rebased.

PR #16223, opened 2025-08-15
Auth: Refactoring in preparation for token based authentication

markylaing:
**Motivation**

Maintaining a consistent view of the calling identity is critical for access control and auditing purposes. This is especially true when requests can be forwarded between cluster members. Our current approach needs improvement for four reasons:
1. At the moment, the `request.Info` struct contains all information about the calling identity, including forwarding information. The `request.CreateRequestor` function and many utility functions in the `auth` package all inspect the `request.Info` to determine the calling identity. Each new utility that is created must be aware of request forwarding, and not use e.g. `request.Info.Username` directly.
2. Upcoming work on [events filtering](https://github.com/canonical/lxd/pull/15991) will add a new `Requestor` field to the `Operation` response. To facilitate this, an internal representation of the requestor should be used in the `operations.Operations` struct that can be converted to an `api.EventLifecycleRequestor` or `api.OperationRequestor` as required.
3. When `request.Info` was added, we used it as a catch-all for anything stored in the request context. This caused an issue for HTTP2 connections because the request context [actually persists across multiple requests](https://github.com/canonical/lxd/pull/16038). Because of this, we reset the `request.Info` on every request. We could improve performance on HTTP2 connections by using the existing `request.Info` (double checking it of course) if we remove the `EffectiveProject` from it.
4. Every time authorization is performed, we are currently:
    i. Getting the `request.Info`.
    ii. Resolving the caller identity (checking if the request was forwarded).
    iii. Getting the identity from the identity cache.
    iv. Performing the access check.
    This could be greatly simplified if the identity cache entry is stored in the request context. We have to get the identity from the cache in order to authenticate them, so there should be no reason not to store the information.

**Implementation**

1. Removes `EffectiveProject` from `request.Info`, and changes `request.Info` to `request.RequestorDetails`. This naming scopes the usage of the type.
2. ~~Adds a `Requestor` interface, which `RequestorDetails` implements.~~ Adds a Requestor type with unexported fields.
4. Adds an `identity` field to `Requestor`, of type `*identity.CacheEntry`. This can be nil. The identity should be set during authentication.

Note: to prevent this PR from going into orbit, there are two commits that are not optimal. I intend to address these in follow up PRs. The commits are:
- ~~*lxd: Update post authentication logic to set the requestor.* With the addition of bearer token authentication, I would like to refactor authentication into three drivers. This will reduce the cyclomatic complexity of both the Daemon.Authenticate and Daemon.createCmd functions. The `tls` and `bearer` drivers would always be set, with the `oidc` driver only set if configured.~~
- *lxd/request: Update CreateRequestor to use the internal Requestor.* As part of ongoing work to ensure all lifecycle events have a requestor, I would like to remove this function entirely so that an empty requestor is never silently ignored.

Depends on #16167

### DIFF
--- a/lxd/request/const.go
+++ b/lxd/request/const.go
@@ -31,18 +31,18 @@ const (
 
 // Headers.
 const (
-	// HeaderForwardedAddress is the forwarded address field in request header.
-	HeaderForwardedAddress = "X-LXD-forwarded-address"
+	// headerForwardedAddress is the forwarded address field in request header.
+	headerForwardedAddress = "X-LXD-forwarded-address"
 
-	// HeaderForwardedUsername is the forwarded username field in request header.
-	HeaderForwardedUsername = "X-LXD-forwarded-username"
+	// headerForwardedUsername is the forwarded username field in request header.
+	headerForwardedUsername = "X-LXD-forwarded-username"
 
-	// HeaderForwardedProtocol is the forwarded protocol field in request header.
-	HeaderForwardedProtocol = "X-LXD-forwarded-protocol"
+	// headerForwardedProtocol is the forwarded protocol field in request header.
+	headerForwardedProtocol = "X-LXD-forwarded-protocol"
 
-	// HeaderForwardedIdentityProviderGroups is the forwarded identity provider groups field in request header.
+	// headerForwardedIdentityProviderGroups is the forwarded identity provider groups field in request header.
 	// This will be a JSON marshalled []string.
-	HeaderForwardedIdentityProviderGroups = "X-LXD-forwarded-identity-provider-groups"
+	headerForwardedIdentityProviderGroups = "X-LXD-forwarded-identity-provider-groups"
 )
 
 const (
--- a/lxd/request/requestor.go
+++ b/lxd/request/requestor.go
@@ -131,23 +131,23 @@ func (r *Requestor) IsForwarded() bool {
 // ForwardProxy returns a proxy function that adds the requestor details as headers to be inspected by the receiving cluster member.
 func (r *Requestor) ForwardProxy() func(req *http.Request) (*url.URL, error) {
 	return func(req *http.Request) (*url.URL, error) {
-		req.Header.Add(HeaderForwardedAddress, r.CallerAddress())
+		req.Header.Add(headerForwardedAddress, r.CallerAddress())
 
 		username := r.CallerUsername()
 		if username != "" {
-			req.Header.Add(HeaderForwardedUsername, username)
+			req.Header.Add(headerForwardedUsername, username)
 		}
 
 		protocol := r.CallerProtocol()
 		if protocol != "" {
-			req.Header.Add(HeaderForwardedProtocol, protocol)
+			req.Header.Add(headerForwardedProtocol, protocol)
 		}
 
 		identityProviderGroups := r.CallerIdentityProviderGroups()
 		if identityProviderGroups != nil {
 			b, err := json.Marshal(identityProviderGroups)
 			if err == nil {
-				req.Header.Add(HeaderForwardedIdentityProviderGroups, string(b))
+				req.Header.Add(headerForwardedIdentityProviderGroups, string(b))
 			}
 		}
 
@@ -167,10 +167,10 @@ func (r *Requestor) ForwardingMemberFingerprint() (string, error) {
 
 // setForwardingDetails validates and sets forwarding details from the request headers.
 func (r *Requestor) setForwardingDetails(req *http.Request) error {
-	forwardedAddress := req.Header.Get(HeaderForwardedAddress)
-	forwardedUsername := req.Header.Get(HeaderForwardedUsername)
-	forwardedProtocol := req.Header.Get(HeaderForwardedProtocol)
-	forwardedIdentityProviderGroupsJSON := req.Header.Get(HeaderForwardedIdentityProviderGroups)
+	forwardedAddress := req.Header.Get(headerForwardedAddress)
+	forwardedUsername := req.Header.Get(headerForwardedUsername)
+	forwardedProtocol := req.Header.Get(headerForwardedProtocol)
+	forwardedIdentityProviderGroupsJSON := req.Header.Get(headerForwardedIdentityProviderGroups)
 
 	// Requests can only be forwarded from other cluster members.
 	if r.protocol != ProtocolCluster {


lxd/request/const.go
```diff
@@ -39,3 +44,26 @@ const (
...
+	// Note: If core.trust_ca_certificates is false, the peer certificate is additionally verified via mTLS and
+	// RequestorDetails.Protocol is set to api.AuthenticationMethodTLS.
+	//
+	// Note: Regardless of whether `core.trust_ca_certificates` is enabled, if an identity corresponding to the clients
+	// peer certificate exists in the [identity.Cache], then protocol should be set to api.AuthenticationMethodTLS and
+	// the identity should be set as the RequestorDetails.Identity.
+	ProtocolPKI string = "pki"
+
+	// ProtocolDevLXD is the authentication method for interacting with the devlxd API.
+	ProtocolDevLXD = "devlxd"
```
kadinsayani:
```suggestion
	// ProtocolCluster is set as the [RequestorDetails.Protocol] when the request is authenticated via mTLS and the peer
	// certificate is present in the trust store as type [certificate.TypeServer].
	ProtocolCluster string = "cluster"

	// ProtocolUnix is set as the [RequestorDetails.Protocol] when the request is made over the unix socket.
	ProtocolUnix string = "unix"

	// ProtocolPKI is set as the [RequestorDetails.Protocol] when a `server.ca` file exists in LXD_DIR, the peer
	// certificate of the request was signed by the CA file, and core.trust_ca_certificates is true.
	//
	// Note: If core.trust_ca_certificates is false, the peer certificate is additionally verified via mTLS and
	// [RequestorDetails.Protocol] is set to [api.AuthenticationMethodTLS].
	//
	// Note: Regardless of whether `core.trust_ca_certificates` is enabled, if an identity corresponding to the clients
	// peer certificate exists in the [identity.Cache], then protocol should be set to [api.AuthenticationMethodTLS] and
	// the identity should be set as the [RequestorDetails.Identity].
	ProtocolPKI string = "pki"

	// ProtocolDevLXD is the authentication method for interacting with the devlxd API.
	ProtocolDevLXD = "devlxd"
```

Can we please include go doc links where possible?

lxd/request/requestor.go
```diff
@@ -0,0 +1,291 @@
...
+	return info, nil
+}
+
+// GetForwardedRequestorDetails gets requestor details from the request headers. It should only be called when the request
+// was sent from another cluster member.
+func GetForwardedRequestorDetails(r *http.Request) (username string, protocol string, address string, identityProviderGroups []string, err error) {
+	address = r.Header.Get(HeaderForwardedAddress)
+	username = r.Header.Get(HeaderForwardedUsername)
+	protocol = r.Header.Get(HeaderForwardedProtocol)
+
```
kadinsayani: Can we verify the request is from another cluster member here? That way we can provide a useful error if a caller tries to use this function in an incorrect context.

```suggestion
// GetForwardedRequestorDetails gets requestor details from the request headers. It returns an error if the request is not sent from another cluster member.
func GetForwardedRequestorDetails(r *http.Request, requestorProtocol string) (username string, protocol string, address string, identityProviderGroups []string, err error) {
    if requestorProtocol != request.ProtocolCluster {
        return "", "", "", nil, fmt.Errorf("Invalid requestor protocol %q, reqeustorProtocol)
    }

	address = r.Header.Get(HeaderForwardedAddress)
	username = r.Header.Get(HeaderForwardedUsername)
	protocol = r.Header.Get(HeaderForwardedProtocol)

```

lxd/request/requestor.go
```diff
@@ -0,0 +1,291 @@
...
+	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/api"
+)
+
+// ErrRequestorNotPresent is a sentinel error used when getting the Requestor from the request context.
+var ErrRequestorNotPresent = errors.New("No requestor was found in the given context")
+
+// Requestor is a view of the original caller. An interface is used here to make the caller think twice about accessing
+// the raw details (which can be done via type assertion if needed) because e.g. RequestorDetails.Username might be the
+// fingerprint of another cluster member, and not the callers actual username.
```
kadinsayani: This comment is a bit confusing to me. If a caller should think twice about accessing raw details, why export them?

lxd/request/requestor.go
```diff
@@ -0,0 +1,291 @@
...
+	"net/url"
+	"slices"
+
+	"github.com/canonical/lxd/lxd/identity"
+	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/api"
+)
+
+// ErrRequestorNotPresent is a sentinel error used when getting the Requestor from the request context.
+var ErrRequestorNotPresent = errors.New("No requestor was found in the given context")
```
kadinsayani: Can this go in a separate `errors.go` file in the `request` package?

kadinsayani:
I believe the tests are failing because `originalProtocol` (`cluster`) is passed to `ValidateAuthenticationMethod()`:

https://github.com/markylaing/lxd/blob/d3082b082c027abe2e1d88398504163932cba2b9/lxd/daemon.go#L898-L908

lxd/request/const.go
markylaing: Have updated.

PS: I don't think we need the square brackets when referencing within a package? I've added brackets for anything outside the package.

lxd/request/requestor.go
markylaing: This is now changed to an internal method on `Requestor` which validates that the headers are not set unless the protocol is "cluster" (internal) and then parses and sets them if present.

lxd/request/requestor.go
markylaing: This is all changed now. We have an exported type `Requestor` with exported methods but no exported fields, and there is no interface.

lxd/request/requestor.go
markylaing: Done thanks :)

markylaing:
> I believe the tests are failing because `originalProtocol` (`cluster`) is passed to `ValidateAuthenticationMethod()`:
> 
> https://github.com/markylaing/lxd/blob/d3082b082c027abe2e1d88398504163932cba2b9/lxd/daemon.go#L898-L908

Yeah there was a whole bunch wrong with my first few attempts. It's all very streamlined now ;)

markylaing:
@tomponline I'm thinking it'll be better to break this up into smaller PRs. Going to mark as draft again and will submit tomorrow morning.

lxd/request/const.go
kadinsayani: Thanks! I think we should include square brackets for any exported identifiers in the current package and when referring to any identifiers from other packages: https://tip.golang.org/doc/comment#doclinks.

kadinsayani: Some IDE's will automatically provide a reference when hovering over a known identifier, but including square brackets in the cases mentioned above makes our Go Doc comments more robust and consistent across all environments.

lxd/request/requestor.go
kadinsayani: Nice!

lxd/request/requestor.go
kadinsayani: Awesome, looks good!

kadinsayani (approved):
LGTM, thanks! The new `Requestor` type and its methods definitely streamline auth :)

markylaing:
@tomponline thanks for merging those smaller PRs. Unfortunately I can't make this one any smaller but it's now ready for review when you're ready, thanks.

lxd/request/requestor.go
```diff
@@ -0,0 +1,323 @@
...
+
+	if r.identityType == nil {
+		return false
+	}
+
+	return r.identityType.IsAdmin()
+}
+
+// CallerAddress returns the original caller address.
+func (r *Requestor) CallerAddress() string {
```
tomponline, line 72: OriginAddress would have been clearer in my view, but not a blocker.

lxd/request/requestor.go
```diff
@@ -0,0 +1,323 @@
...
+			}
+		}
+
+		return shared.ProxyFromEnvironment(req)
+	}
+}
+
+// ForwardingMemberFingerprint returns the fingerprint of the cluster member that forwarded the request. It returns
+// an error if the request was not sent by another cluster member.
+func (r *Requestor) ForwardingMemberFingerprint() (string, error) {
```
tomponline, line 160: Fingerprint of what? Is this a TLS certificate fingerprint? I think we can make the function name or comment clearer here.

lxd/request/requestor.go
```diff
@@ -0,0 +1,323 @@
...
+	callerProtocol := r.CallerProtocol()
+
+	// Handle untrusted case
+	if !r.trusted {
+		// If the caller is not trusted, there should not be a username.
+		if callerUsername != "" {
+			return errors.New("Caller is not trusted but a username was set")
+		}
+
+		// The only allowed protocols for the untrusted case are ProtocolDevLXD, or empty.
```
tomponline, line 280: when is it empty?

lxd/request/requestor.go
```diff
@@ -0,0 +1,323 @@
...
+			return errors.New("Unsupported protocol set for untrusted request")
+		}
+
+		SetContextValue(req, ctxRequestor, r)
+		return nil
+	}
+
+	// Trusted
+
+	// DevLXD requests must always be untrusted.
```
tomponline, line 291: this will change soon right?

lxd/request/requestor.go
```diff
@@ -0,0 +1,323 @@
...
+	if callerProtocol == ProtocolDevLXD {
+		return errors.New("Received trusted request over DevLXD")
```
tomponline, line 293: the word "unexpected" would be good here as the error message as it is right now sounds like a good thing.

lxd/auth/drivers/openfga.go
```diff
@@ -159,24 +159,24 @@ func (e *embeddedOpenFGA) checkPermission(ctx context.Context, entityURL *api.UR
 	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 
-	// Untrusted requests are denied.
-	if !auth.IsTrusted(ctx) {
-		return api.NewGenericStatusError(http.StatusForbidden)
+	requestor, err := request.GetRequestor(ctx)
+	if err != nil {
+		return err
```
tomponline, line 164: Please can you wrap the error here

lxd/auth/drivers/openfga.go
```diff
@@ -373,24 +368,24 @@ func (e *embeddedOpenFGA) getPermissionChecker(ctx context.Context, entitlement
 		return nil, fmt.Errorf("Failed to get a permission checker: %w", err)
 	}
 
-	// Untrusted requests are denied.
-	if !auth.IsTrusted(ctx) {
-		return allowFunc(false), nil
+	requestor, err := request.GetRequestor(ctx)
+	if err != nil {
+		return nil, err
```
tomponline, line 373: Please wrap error

lxd/devlxd.go
```diff
@@ -487,11 +487,12 @@ func registerDevLXDEndpoint(d *Daemon, apiRouter *mux.Router, apiVersion string,
...
 	// Function that handles the request by calling the appropriate handler.
 	handleFunc := func(w http.ResponseWriter, r *http.Request) {
-		// Initialise the request context info.
-		reqInfo := request.InitContextInfo(r)
-
 		// Set devLXD auth method to identify this request as coming from the /dev/lxd socket.
-		reqInfo.Protocol = request.ProtocolDevLXD
+		err := request.SetRequestor(r, d.identityCache, request.RequestorArgs{Protocol: request.ProtocolDevLXD})
+		if err != nil {
+			_ = response.DevLXDErrorResponse(api.StatusErrorf(http.StatusInternalServerError, "%v", err)).Render(w, r)
```
tomponline, line 493: should we be using SmartError here rather than forcing it to InternalServiceError?

tomponline (approved):
LGTM

lxd/devlxd.go
markylaing: AFIAK no - we need to use the devlxd error response at this point. That writes a plaintext error and sets the status code.

markylaing:
> Thanks @markylaing overall I really like this structure, it improves readability a lot.
> 
> I'll merge this, but ive left a few comments/questions, and please can you wrap error returned from request.GetRequestor(ctx) call.

Thanks :) I'll submit a follow up PR to address your comments. In the meantime, I'll open #16225 up for review :)

lxd/daemon.go
```diff
@@ -491,39 +502,51 @@ func (d *Daemon) Authenticate(w http.ResponseWriter, r *http.Request) (trusted b
...
 		if err != nil {
-			return false, "", "", nil, err
+			return nil, err
 		}
 
-		u, err := user.LookupId(strconv.FormatUint(uint64(cred.Uid), 10))
-		if err != nil {
-			return true, fmt.Sprint("uid=", cred.Uid), request.ProtocolUnix, nil, nil
+		uid := strconv.FormatUint(uint64(cred.Uid), 10)
+		username := "uid=" + uid
```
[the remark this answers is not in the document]
markylaing: Not true, `uid` is used in the next statement in `user.LookupId`

lxd/request/requestor.go
markylaing: No, but the comment could be updated in #16225 to `Requests using ProtocolDevLXD must always be untrusted`.

This is because the protocol is not `ProtocolDevLXD` when the identity in the guest authenticates with a bearer token. It is `api.AuthenticationMethodBearer`.

lxd/request/requestor.go
markylaing: It's empty when a main API request is untrusted, e.g. during an untrusted GET /1.0/images. Will update the comment

lxd/request/requestor.go
markylaing: Yep the TLS certificate fingerprint. Will update.

lxd/request/requestor.go
markylaing: I've made it `SourceAddress` to align it with the field name in `Requestor`. If you prefer I can make them both "origin address".

tomponline: I worry that SourceAddress might be confused with the source address of the requesting cluster member when its forwarded.

markylaing: Ok will update both to OriginAddress :+1:

markylaing:
Follow up PR here: https://github.com/canonical/lxd/pull/16265

Issue #15857, opened 2025-06-23
`lxc network forward create --allocate` can allocate the physical network's `ipv4.gateway`


MggMuggins:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

24.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
$ snap list --all lxd core20 core22 core24 snapd
Name    Version         Rev    Tracking        Publisher   Notes
core20  20250429        2582   latest/stable   canonical✓  base,disabled
core20  20250526        2599   latest/stable   canonical✓  base
core22  20250425        1981   latest/stable   canonical✓  base,disabled
core22  20250528        2010   latest/stable   canonical✓  base
core24  20250504        988    latest/stable   canonical✓  base,disabled
core24  20250526        1006   latest/stable   canonical✓  base
lxd     5.21.3-c5ae129  33110  5.21/candidate  canonical✓  disabled,in-cohort
lxd     5.21.3-27d70fa  34068  5.21/candidate  canonical✓  in-cohort
snapd   2.68.4          24505  latest/stable   canonical✓  snapd,disabled
snapd   2.68.5          24718  latest/stable   canonical✓  snapd
```

### Output of "lxc info" or system info if it fails

```shell
$ snap list --all microovn microceph microcloud
Name        Version                 Rev   Tracking      Publisher   Notes
microceph   19.2.0+snapd37ac0ff9d   1344  squid/stable  canonical✓  disabled
microceph   19.2.0+snap3b53da1c21   1355  squid/stable  canonical✓  -
microcloud  2.1.0-3e8b183           1144  2/stable      canonical✓  -
microovn    24.03.2+snapa2c59c105b  667   24.03/stable  canonical✓  -
$ uname -a
Linux houston 6.8.0-60-generic #63-Ubuntu SMP PREEMPT_DYNAMIC Tue Apr 15 19:04:15 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
```

### Issue description

For this UPLINK configuration:
```yaml
name: UPLINK
description: ""
type: physical
managed: true
status: Created
config:
  dns.nameservers: 10.0.0.1,8.8.8.8
  ipv4.gateway: 10.0.0.1/24
  ipv4.ovn.ranges: 10.0.0.20-10.0.0.80
  ipv4.routes: 10.0.0.0/28
  volatile.last_state.created: "false"
```

[...]

PR #16224, opened 2025-08-15
Network: Prevent auto allocating uplink's gateway IP as a listen address for OVN forwards and load balancers

nmezhenskyi:
Fixes https://github.com/canonical/lxd/issues/15857.

## Changes:

- Check that the listen address for allocation is not taken by the uplink's gateway IP.
- Fix the condition in `randomAddressInSubnet()` function that determines whether there are any available addresses left to try to find.

### DIFF
--- a/test/suites/network_ovn.sh
+++ b/test/suites/network_ovn.sh
@@ -140,6 +140,22 @@ test_network_ovn() {
   lxc network unset "${uplink_network}" ipv4.routes
   lxc network unset "${uplink_network}" ipv6.routes
 
+  echo "Set ipv4.gateway and ipv6.gateway for the uplink."
+  lxc network set "${uplink_network}" ipv4.gateway=192.0.2.1/24
+  lxc network set "${uplink_network}" ipv6.gateway=2001:db8:1:2::1/64
+
+  echo "Update the uplink's ipv4.routes and ipv6.routes to include the gateway addresses."
+  lxc network set "${uplink_network}" ipv4.routes=192.0.2.0/29
+  lxc network set "${uplink_network}" ipv6.routes=2001:db8:1:2::/125
+
+  echo "Check that automatic allocation does not allocate uplink's ipv4.gateway and ipv6.gateway for forwards."
+  auto_allocate_forwards_ip4 "192.0.2"
+  auto_allocate_forwards_ip6 "2001:db8:1:2"
+
+  echo "Check that automatic allocation does not allocate uplink's ipv4.gateway and ipv6.gateway for load balancers."
+  auto_allocate_load_balancers_ip4 "192.0.2"
+  auto_allocate_load_balancers_ip6 "2001:db8:1:2"
+
   echo "Clean up created networks."
   lxc network delete "${ovn_network}"
   lxc network delete "${uplink_network}"
@@ -473,8 +489,41 @@ test_network_ovn() {
   echo "Check that instance NIC passthrough with ipv6.routes.external does not allow using volatile.network.ipv6.address."
   ! lxc launch testimage c1 -n "${ovn_network}" -d eth0,ipv6.routes.external="${volatile_ip6}/128" || false
 
+  echo "Test automatic allocation of an allowed external IP addresses for forwards and load balancers."
+
+  echo "Update uplink's routes to include the uplink's IPv4 and IPv6 gateway addresses."
+  lxc network set "${uplink_network}" ipv4.routes=10.10.10.0/29
+  lxc network set "${uplink_network}" ipv6.routes=fd42:4242:4242:1010::/125
+
+  echo "Check that automatic allocation does not allocate uplink's ipv4.address and ipv6.address for forwards."
+  auto_allocate_forwards_ip4 "10.10.10"
+  auto_allocate_forwards_ip6 "fd42:4242:4242:1010"
+
+  echo "Check that automatic allocation does not allocate uplink's ipv4.address and ipv6.address for load balancers."
+  auto_allocate_load_balancers_ip4 "10.10.10"
+  auto_allocate_load_balancers_ip6 "fd42:4242:4242:1010"
+
+  echo "Set ipv4.dhcp.gateway for the uplink."
+  lxc network set "${uplink_network}" ipv4.dhcp.gateway=192.0.2.1
+
+  echo "Update the uplink's ipv4.routes to include the gateway address."
+  lxc network set "${uplink_network}" ipv4.routes=192.0.2.0/29
+
+  echo "Check that automatic allocation does not allocate uplink's ipv4.dhcp.gateway for forwards."
+  auto_allocate_forwards_ip4 "192.0.2"
+
+  echo "Check that automatic allocation does not allocate uplink's ipv4.dhcp.gateway for load balancers."
+  auto_allocate_load_balancers_ip4 "192.0.2"
+
   echo "Delete the OVN network in the default project."
   lxc network delete "${ovn_network}"
+
+  echo "Unset ipv4.dhcp.gateway for the uplink."
+  lxc network unset "${uplink_network}" ipv4.dhcp.gateway
+
+  echo "Reset the uplink's routes."
+  lxc network set "${uplink_network}" ipv4.routes=192.0.2.0/24
+  lxc network set "${uplink_network}" ipv6.routes=2001:db8:1:2::/64
 
   echo "Test ha_chassis removal on shutdown."
   shutdown_lxd "${LXD_DIR}"
@@ -640,4 +689,80 @@ test_network_ovn() {
   assert_row_count
 
   unset_ovn_configuration
+}
+
+auto_allocate_forwards_ip4() {
+  # Network X.X.X.0/29 has 5 usable addresses (.2, .3, .4, .5, .6), excluding the uplink's gateway (.1), broadcast address (.7), and network address (.0).
+  echo "Allocate all available IPv4 addresses for forwards."
+  for _ in $(seq 5); do
+    lxc network forward create "${ovn_network}" --allocate=ipv4
+  done
+
+  echo "Check that there is no forward with uplink's IPv4 gateway."
+  ! lxc network forward show "${ovn_network}" "${1}.1" || false
+
+  echo "Check that there is no more available IPv4 addresses left."
+  ! lxc network forward create "${ovn_network}" --allocate=ipv4 || false
+
+  echo "Clean up forwards."
+  for i in $(seq 2 6); do
+    lxc network forward delete "${ovn_network}" "${1}.${i}"
+  done
+}
+
+auto_allocate_forwards_ip6() {
+  # Network X:X:X:X::/125 has 6 usable addresses (::2, ::3, ::4, ::5, ::6, ::7), excluding the uplink's gateway (::1) and the subnet-router anycast address (::).
+  echo "Allocate all available IPv6 addresses for forwards."
+  for _ in $(seq 6); do
+    lxc network forward create "${ovn_network}" --allocate=ipv6
+  done
+
+  echo "Check that there is no forward with uplink's IPv6 gateway."
+  ! lxc network forward show "${ovn_network}" "${1}::1" || false
+
+  echo "Check that there is no more available IPv6 addresses left."
+  ! lxc network forward create "${ovn_network}" --allocate=ipv6 || false
+
+  echo "Clean up forwards."
+  for i in $(seq 2 7); do
+    lxc network forward delete "${ovn_network}" "${1}::${i}"
+  done
+}
+
+auto_allocate_load_balancers_ip4() {
+  # Network X.X.X.0/29 has 5 usable addresses (.2, .3, .4, .5, .6), excluding the uplink's gateway (.1), broadcast address (.7), and network address (.0).
+  echo "Allocate all available IPv4 addresses for load balancers."
+  for _ in $(seq 5); do
+    lxc network load-balancer create "${ovn_network}" --allocate=ipv4
+  done
+
+  echo "Check that there is no load balancer with uplink's IPv4 gateway."
+  ! lxc network load-balancer show "${ovn_network}" "${1}.1" || false
+
+  echo "Check that there is no more available IPv4 addresses left."
+  ! lxc network load-balancer create "${ovn_network}" --allocate=ipv4 || false
+
+  echo "Clean up load balancers."
+  for i in $(seq 2 6); do
+    lxc network load-balancer delete "${ovn_network}" "${1}.${i}"
+  done
+}
+
+auto_allocate_load_balancers_ip6() {
+  # Network X:X:X:X::/125 has 6 usable addresses (::2, ::3, ::4, ::5, ::6, ::7), excluding the uplink's gateway (::1) and the subnet-router anycast address (::).
+  echo "Allocate all available IPv6 addresses for load balancers."
+  for _ in $(seq 6); do
+    lxc network load-balancer create "${ovn_network}" --allocate=ipv6
+  done
+
+  echo "Check that there is no load balancer with uplink's IPv6 gateway."
+  ! lxc network load-balancer show "${ovn_network}" "${1}::1" || false
+
+  echo "Check that there is no more available IPv6 addresses left."
+  ! lxc network load-balancer create "${ovn_network}" --allocate=ipv6 || false
+
+  echo "Clean up load balancers."
+  for i in $(seq 2 7); do
+    lxc network load-balancer delete "${ovn_network}" "${1}::${i}"
+  done
 }


tomponline:
Thanks for this @nmezhenskyi we should add tests in this repo to check the validation is working though.

nmezhenskyi:
The tests are currently flaky, I'm investigating what's the issue. When creating forwards / load balancers in a loop with auto allocated listen address, the allocation operation sometimes exceeds the context deadline even when there is still an available address. Perhaps there is some lock contention happening with the database calls.

nmezhenskyi:
The issue actually occurs in this function:
https://github.com/canonical/lxd/blob/0d701e470f583fa5525b0997aa678f364b746af0/lxd/network/driver_ovn.go#L235

Sometimes it does not find the available address in time, especially if there is only one available address left.

nmezhenskyi:
As a workaround, I'm trying to increase the context timeout for the `randomExternalAddress()` function and use `sleep 0.5` between the iterations in the for loops in the tests.

tomponline:
> As a workaround, I'm trying to increase the context timeout for the `randomExternalAddress()` function and use `sleep 0.5` between the iterations in the for loops in the tests.

which part of randomExternalAddress is taking the time?

tomponline:
Is it that the random number generator is not arriving at the desired address in time?

nmezhenskyi:
> Is it that the random number generator is not arriving at the desired address in time?

@tomponline I'm still debugging but so far it seems that `randomAddressInSubnet()` function sometimes returns `noAvailableAddressErr` when there is only one available address left. This is inconsistent, because on the next try it often succeeds and finds the last available address correctly. I found that this is reproducible even without the loop.

nmezhenskyi:
@tomponline @markylaing I think I fixed the issue now. It seems the `randomAddressInSubnet()` function was incorrectly tracking the number of IPs to attempt to allocate. 

The loop with random numbers generator continues until all available addresses have been exhausted/visited. It uses the length of the map with visited IPs to determine whether there are any non-visited IPs left. However, it was always expecting the number of possibly available IPs to be 1 less than actually is.

The reason why it sometimes worked is that random number generator sometimes correctly produced the last possible IP before the loop would determine that it visited all the IPs.

tomponline:
> @tomponline @markylaing I think I fixed the issue now. It seems the `randomAddressInSubnet()` function was incorrectly tracking the number of IPs to attempt to allocate.
> 
> The loop with random numbers generator continues until all available addresses have been exhausted/visited. It uses the length of the map with visited IPs to determine whether there are any non-visited IPs left. However, it was always expecting the number of possibly available IPs to be 1 less than actually is.
> 
> The reason why it sometimes worked is that random number generator sometimes correctly produced the last possible IP before the loop would determine that it visited all the IPs.

Nice. One of the 3 hardest problems in computing. ;)

lxd/network/driver_ovn.go
```diff
@@ -5938,6 +5938,10 @@ func (n *ovn) forPeers(f func(targetOVNNet *ovn) error) error {
 // checkAddressNotInUse checks that a given network subnet does not fall within
 // any existing OVN network external subnets on the same uplink.
 func (n *ovn) checkAddressNotInUse(netip *net.IPNet) (bool, error) {
```
tomponline: So a couple of things I'd like to say about this function:

1. This isn't due to your change. But I am not enjoying the fact that `checkAddressNotInUse` both previously and now extended by your change, loads records from the database (fine in and of itself), but it is called many times when auto allocating an IP - which means its loading the same data from the database multiple times. This is bad because its inefficient. Instead I think we should load the necessary "external subnets in use" where checkAddressNotInUse is called and then have it passed to it on each invocation.
2. Further more, I think the newly added code in this function to load the gateway IPs on the uplink should be moved into its own function and then called from where `checkAddressNotInUse` is called, and the gateway IPs added as subnets to the  "external subnets in use" list as /32 entries.

This way we improve the efficiency of this function and simplify it by ensuring it only needs to consider subnets.

nmezhenskyi: Sounds good, I'll work on that

test/suites/network_ovn.sh
```diff
@@ -140,6 +140,86 @@ test_network_ovn() {
...
+
+  echo "Check that there is no more available IPv4 addresses left."
+  ! lxc network forward create "${ovn_network}" --allocate=ipv4 || false
+
+  echo "Clean up forwards."
+  for i in $(seq 2 6); do
+    lxc network forward delete "${ovn_network}" "192.0.2.${i}"
+  done
+
+  # Network 2001:db8:1:2::/125 has 6 available addresses, excluding the uplink's ipv6.gateway and subnet anycast address.
```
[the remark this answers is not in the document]
nmezhenskyi: Updated

test/suites/network_ovn.sh
```diff
@@ -140,6 +140,86 @@ test_network_ovn() {
...
+  lxc network set "${uplink_network}" ipv4.gateway=192.0.2.1/24
+  lxc network set "${uplink_network}" ipv6.gateway=2001:db8:1:2::1/64
+
+  echo "Update the uplink's ipv4.routes and ipv6.routes to include the gateway addresses."
+  lxc network set "${uplink_network}" ipv4.routes=192.0.2.0/29
+  lxc network set "${uplink_network}" ipv6.routes=2001:db8:1:2::/125
+
+  echo "Check that automatic allocation does not allocate uplink's ipv4.gateway and ipv6.gateway for forwards."
+
+  # Network 192.0.2.0/29 has 5 available addresses, excluding the uplink's ipv4.gateway, broadcast address, and network address.
```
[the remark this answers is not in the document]
nmezhenskyi: Updated

lxd/network/driver_ovn.go
```diff
@@ -1166,6 +1178,79 @@ func (n *ovn) getLogicalRouterPeerPortName(peerNetworkID int64) openvswitch.OVNR
...
+	uplinkBridgeIP4AddrKey := "ipv4.address"             // Uses CIDR notation.
+	uplinkBridgeIP6AddrKey := "ipv6.address"             // Uses CIDR notation.
+	uplinkBridgeIP4DHCPGatewayKey := "ipv4.dhcp.gateway" // Uses bare IP.
+
+	subnetsInUse := make([]externalSubnetUsage, 0, 5)
+
+	// Parse IPs in CIDR notation and record usage.
+	for _, key := range []string{uplinkPhysicalIP4GatewayKey, uplinkPhysicalIP6GatewayKey, uplinkBridgeIP4AddrKey, uplinkBridgeIP6AddrKey} {
+		if uplinkNetwork.Config[key] != "" {
+			ipAndPrefix := strings.Split(uplinkNetwork.Config[key], "/")
```
simondeziel: Trimming any CIDR with `strings.Cut()` would avoid dealing with a slice:
```suggestion
			ip, _, _ := strings.Cut(uplinkNetwork.Config[key], "/")
```

nmezhenskyi: Nice, thanks!

markylaing (dismissed):
This looks great :partying_face: 
+ Well commented
+ Clear
+ More efficient
+ Fixes bug!

My only concern is that now we're getting external subnet usage in advance of searching for a free IP. If it takes a while to get a free IP is there a possibility that an IP will be taken while we're searching? 

I think this is an unlikely scenario so not a blocker, but something to consider. Perhaps this is something that requires a lock.

lxd/network/driver_ovn.go
markylaing: `strings.CutPrefix` is also useful here

tomponline:
> My only concern is that now we're getting external subnet usage in advance of searching for a free IP. If it takes a while to get a free IP is there a possibility that an IP will be taken while we're searching?

This was always an issue to be fair @markylaing

tomponline:
> I think this is an unlikely scenario so not a blocker, but something to consider. Perhaps this is something that requires a lock.

A local lock wont be sufficient for this sadly. Its another example of needing cluster wide locking.

tomponline:
Suggest log a separate issue so we remember to address it and proceed with the current fix.

lxd/network/driver_ovn.go
```diff
@@ -1166,6 +1166,78 @@ func (n *ovn) getLogicalRouterPeerPortName(peerNetworkID int64) openvswitch.OVNR
...
+
+	if uplinkNetwork == nil {
+		return nil, errors.New("Failed to load uplink network")
+	}
+
+	uplinkPhysicalIP4GatewayKey := "ipv4.gateway"        // Uses CIDR notation.
+	uplinkPhysicalIP6GatewayKey := "ipv6.gateway"        // Uses CIDR notation.
+	uplinkBridgeIP4AddrKey := "ipv4.address"             // Uses CIDR notation.
+	uplinkBridgeIP6AddrKey := "ipv6.address"             // Uses CIDR notation.
+	uplinkBridgeIP4DHCPGatewayKey := "ipv4.dhcp.gateway" // Uses bare IP.
```
tomponline, line 1195: ovn networks dont have this option https://documentation.ubuntu.com/lxd/latest/reference/network_ovn/#configuration-options

lxd/network/driver_ovn.go
```diff
@@ -1166,6 +1166,78 @@ func (n *ovn) getLogicalRouterPeerPortName(peerNetworkID int64) openvswitch.OVNR
...
+
+	subnetsInUse := make([]externalSubnetUsage, 0, 5)
+
+	// Parse IPs in CIDR notation and record usage.
+	for _, key := range []string{uplinkPhysicalIP4GatewayKey, uplinkPhysicalIP6GatewayKey, uplinkBridgeIP4AddrKey, uplinkBridgeIP6AddrKey} {
+		if uplinkNetwork.Config[key] != "" {
+			bareIP, _, _ := strings.Cut(uplinkNetwork.Config[key], "/")
+
+			ipToParse := bareIP + "/32"
```
tomponline, line 1204: You can use `ParseIPToNet` here instead.

tomponline: Sorry, I meant `ParseIPCIDRToNet`

lxd/network/driver_ovn.go
```diff
@@ -1166,6 +1166,78 @@ func (n *ovn) getLogicalRouterPeerPortName(peerNetworkID int64) openvswitch.OVNR
...
+			}
+
+			_, ipNet, err := net.ParseCIDR(ipToParse)
+			if err != nil {
+				return nil, fmt.Errorf("Failed to parse IP address: %w", err)
+			}
+
+			subnetsInUse = append(subnetsInUse, externalSubnetUsage{
+				subnet:         *ipNet,
+				networkProject: api.ProjectDefaultName,
```
tomponline: suggest using uplinkNetwork.Project here, because you use api.ProjectDefaultName earlier in the filter, and now you can use the result of that query to record the actual entities returned (which ofc should be the same, but this is more correct in my view because its only using the constant in one place).

tomponline: Also in general I would encourage you to use the values from the DB once you've looked up a record, rather than the input values, so here `netowrkName` would be better to use `uplinkNetwork.Name` in my view.

This way it decouples the input from the DB record found.

lxd/network/driver_ovn.go
```diff
@@ -1166,6 +1166,78 @@ func (n *ovn) getLogicalRouterPeerPortName(peerNetworkID int64) openvswitch.OVNR
...
+				networkName:    uplinkName,
+				usageType:      subnetUsageGateway,
+			})
+		}
+	}
+
+	// Parse ipv4.dhcp.gateway as bare IP and record usage.
+	if uplinkNetwork.Config[uplinkBridgeIP4DHCPGatewayKey] != "" {
```
tomponline, line 1224: I dont think we need to keep this bit, but if we do then we can use `ParseIPToNet` here.

lxd/network/driver_ovn.go
```diff
@@ -269,6 +269,18 @@ func (n *ovn) randomExternalAddress(ctx context.Context, ipVersion int, uplinkRo
 		}
 	}
 
+	externalSubnetsInUse, err := n.getExternalSubnetInUse(n.config["network"])
```
tomponline: How about we update `randomExternalAddress` to take a `[]externalSubnetUsage` as well, which it can pass to `validator`, then in `allocateUplinkAddress` we can move the subnet loading up to before `if listenIPAddress.IsUnspecified() {` line and de-dupe this code.

lxd/network/driver_ovn.go
tomponline: Oh I see, this is the uplink keys ofc, so the uplink can be a managed bridge and have this option. All good.

lxd/network/driver_ovn.go
tomponline: We do need this and we should use `ParseIPToNet` here.

test/suites/network_ovn.sh
tomponline, line 1: There appears to be a lot of duplication here in the additions.
Can we de-dupe them by using separate bash functions to run the common setup & checks from multiple places?

tomponline (changes requested):
Thanks, a few minor changes to be made, but overall this is looking great!

lxd/network/driver_ovn.go
nmezhenskyi: Makes sense, will do.

lxd/network/driver_ovn.go
nmezhenskyi: Oh right, I missed this function, thanks

lxd/network/driver_ovn.go
nmezhenskyi: Good idea, will do.

lxd/network/driver_ovn.go
nmezhenskyi: I replaced `net.ParseCIDR()` with `ParseIPCIDRToNet()`, but this bit with `ipToParse := bareIP + "/32"` is still needed because `ParseIPCIDRToNet()` doesn't handle this logic.

lxd/network/driver_ovn.go
nmezhenskyi: Updated

lxd/network/driver_ovn.go
nmezhenskyi: Updated

lxd/network/driver_ovn.go
nmezhenskyi: Updated

test/suites/network_ovn.sh
nmezhenskyi: I've split the loops into separate functions now, that should reduce the duplication and make the test flow more readable.

test/suites/network_ovn.sh
```diff
@@ -641,3 +690,79 @@ test_network_ovn() {
 
   unset_ovn_configuration
 }
+
+auto_allocate_forwards_ip4() {
+  # Network X.X.X.0/29 has 5 usable addresses (.2, .3, .4, .5, .6), excluding the uplink's gateway (.1), broadcast address (.7), and network address (.0).
+  echo "Allocate all available IPv4 addresses for forwards."
+  for _ in $(seq 5); do
+    lxc network forward create "${ovn_network}" --allocate=ipv4
+  done
```
simondeziel, line 699: Thinking out loud here. How about manually allocating one IP in the middle of the range (`.4`) and then auto allocate the remaining 4 IPs? Would that better ensure the random is always able to find the free IPs to pick from?

It might be a pointless variation of what you are doing already as I don't know how auto allocation works when mixed with manual ones.

simondeziel (approved):

lxd/network/driver_ovn.go
```diff
@@ -1166,6 +1166,78 @@ func (n *ovn) getLogicalRouterPeerPortName(peerNetworkID int64) openvswitch.OVNR
...
+
+	subnetsInUse := make([]externalSubnetUsage, 0, 5)
+
+	// Parse IPs in CIDR notation and record usage.
+	for _, key := range []string{uplinkPhysicalIP4GatewayKey, uplinkPhysicalIP6GatewayKey, uplinkBridgeIP4AddrKey, uplinkBridgeIP6AddrKey} {
+		if uplinkNetwork.Config[key] != "" {
+			bareIP, _, _ := strings.Cut(uplinkNetwork.Config[key], "/")
+
+			ipToParse := bareIP + "/32"
+			if strings.Contains(bareIP, ":") {
```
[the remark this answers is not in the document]
simondeziel: Surely, looking for `:` is good enough. We don't care to distinguish for IPv4 embedded in IPv6 as those are not supported to begin with.

test/suites/network_ovn.sh
nmezhenskyi: > How about manually allocating one IP in the middle of the range (.4) and then auto allocate the remaining 4 IPs? Would that better ensure the random is always able to find the free IPs to pick from?

This should not make any difference because auto allocation first randomly generates the proposed IP from the subnet range and then validates if it's actually available before allocating it

simondeziel: All good then, thanks!

PR #16215, opened 2025-08-14
Doc: Rewrite DevLXD documentation

markylaing:
The DevLXD documentation is quite old and is written informally. The API specification is also hand written and hard to navigate.

With upcoming extensions via bearer token authentication and new endpoints for the k8s CSI, I figured now would be a good time to update it.

This PR updates the page and uses swagger to represent the API.

Unfortunately I wasn't able to generate the swagger from the Go files without modifying the main `rest-api.yaml` file, so I had to hand write it. Also, I wasn't sure where to place it in the `doc` folder, it seemed to disappear unless I placed it in `.sphinx/_static`. Please advise where would be best for this :)

PR #16225, opened 2025-08-16
Auth: Token based authentication for DevLXD

markylaing:
Adds token based authentication for requests over DevLXD.

Depends on:
- #16205 (which depends on #16192)
- #16223 (which depends on #16167)
- #16215


tomponline:
@markylaing think this needs a rebase

markylaing:
> @markylaing think this needs a rebase

Ah sorry. Yes. Also I'll submit a PR separately for the short form identity type aliasing in the CLI.

go.sum
```diff
@@ -129,6 +129,8 @@ github.com/goccy/go-json v0.10.5/go.mod h1:oq7eo15ShAhp70Anwd5lgX2pLfOS3QCiwU/PU
 github.com/gofrs/uuid v4.2.0+incompatible/go.mod h1:b2aQJv3Z4Fp6yNu3cdSllBxTCLRxnplIgP/c0N/04lM=
```
markylaing: Need to drop this commit as it is adding the wrong dependency version (which is updated later).

lxd/db/cluster/identities.go
```diff
@@ -63,8 +63,9 @@ import (
 type AuthMethod string
 
 const (
-	authMethodTLS  int64 = 1
-	authMethodOIDC int64 = 2
+	authMethodTLS    int64 = 1
+	authMethodOIDC   int64 = 2
+	authMethodBearer int64 = 3
```
simondeziel, line 68: Should we leave out the type determination to the compiler rather than force it to be `int64`?  Maybe it's dependent of the underlying CPU arch but in the playground, it seems to pick `int` for such cases: https://go.dev/play/p/Uv8tSZ9taJ0

But maybe I'm splitting hairs here :)

tomponline:
@markylaing did you drop the commit to allow copilot to run?

simondeziel (commented):
I've only done a quick/partial first pass. The only thing I picked up was the `int64` const so not much, yet :)

lxd-agent/devlxd.go
```diff
@@ -71,13 +72,19 @@ func devLXDServer(d *Daemon) *http.Server {
...
 		return nil, err
 	}
 
+	args := &lxd.ConnectionArgs{}
+	token, ok := strings.CutPrefix(r.Header.Get("Authorization"), "Bearer ")
+	if ok {
+		args.BearerToken = token
+	}
+
 	server, err := lxd.ConnectDevLXDHTTPWithContext(context.Background(), nil, client)
```
MusicDin: Args need to be passed into connection func:
```go
lxd.ConnectDevLXDHTTPWithContext(context.Background(), args, client)
                                                       ^^^^
```

lxd/db/cluster/identities.go
tomponline: We tend to stick with int64 for DB values as a convention.

tomponline: In most cases int == int64 anyway.

lxd-agent/devlxd.go
simondeziel: Good thing you caught this as that's the kind of thing Copilot is also good at spotting ... when it works ;)

markylaing:
> @markylaing did you drop the commit to allow copilot to run?

Not yet, just been in a meeting with Din about using this for the CSI. Will sort shortly.

lxd-agent/devlxd.go
markylaing: Fixed :+1:

go.sum
markylaing: Fixed.

markylaing:
Have rebased onto https://github.com/canonical/lxd/pull/16261 and fixed the immediate issues

markylaing:
For some reason it won't let me re-request a review from copilot

simondeziel:
> For some reason it won't let me re-request a review from copilot

You need to have it enabled in your account it seems, it just let me do it.

markylaing:
I'm just marking as draft for a minute while I add a couple of tests for token expiration

lxc/auth.go
```diff
@@ -1440,6 +1489,152 @@ func (c *cmdIdentityGroupRemove) run(cmd *cobra.Command, args []string) error {
...
+	cmd.RunE = c.run
+
+	return cmd
+}
+
+func (c *cmdIdentityTokenIssue) run(cmd *cobra.Command, args []string) error {
+	_, err := c.global.CheckArgs(cmd, args, 1, 1)
+	if err != nil {
+		return err
+	}
```
MusicDin: We return an error when `exit` is true. This ensures the `nil` error is returned when no argument is provided, and the help message is displayed without an error.

```go
	exit, err := c.global.CheckArgs(cmd, args, 1, 1)
	if exit {
		return err
	}
```


In the current state, when a `nil` error is returned the code does not exit, resulting in a panic when no argument is provided (in `c.identity.resolveIdentityArg(args[0])`).

Same applies for both `issue` and `revoke` commands:
```
$ lxc auth identity token revoke
Description:
  Revoke the current token for a bearer identity

Usage:
  lxc auth identity token revoke [<remote>:]<authentication_method>/<name_or_identifier> [flags]

Global Flags:
      --debug          Show all debug messages
      --force-local    Force using the local unix socket
  -h, --help           Print help
      --project        Override the source project
  -q, --quiet          Don't show progress information
      --sub-commands   Use with help or --help to view sub-commands
  -v, --verbose        Show all information messages
      --version        Print version number
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
main.(*cmdIdentityTokenRevoke).run(0xc0002f7170, 0xc0001a39f8?, {0x0, 0x0, 0x0?})
        github.com/canonical/lxd/lxc/auth.go:1603 +0x39a
github.com/spf13/cobra.(*Command).execute(0xc000323b08, {0x1421ba0, 0x0, 0x0})
        github.com/spf13/cobra@v1.9.1/command.go:1015 +0xaaa
github.com/spf13/cobra.(*Command).ExecuteC(0xc00020ec08)
        github.com/spf13/cobra@v1.9.1/command.go:1148 +0x46f
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra@v1.9.1/command.go:1071
main.main()
        github.com/canonical/lxd/lxc/main.go:333 +0x1b25
```

lxd/identities.go
```diff
@@ -82,6 +83,21 @@ var oidcIdentitiesCmd = APIEndpoint{
...
 }
 
+var tokenIdentitiesCmd = APIEndpoint{
+	Name:        "identities",
+	Path:        "auth/identities/bearer",
+	MetricsType: entity.TypeIdentity,
+
+	Get: APIEndpointAction{
+		Handler:       getIdentities(api.AuthenticationMethodBearer),
+		AccessHandler: allowAuthenticated,
```
markylaing, line 99: Double check why this is `allowAuthenticated`

lxd/auth/encryption/key.go
```diff
@@ -0,0 +1,52 @@
...
+
+// CookieBlockKey returns a key suitable for cookie encryption with AES-256.
+func CookieBlockKey(secret []byte, salt []byte) ([]byte, error) {
+	return deriveKey(secret, salt, "ENCRYPTION", 32)
+}
+
+// TokenSigningKey returns a key suitable for signing JWTs with HMAC-SHA512.
+func TokenSigningKey(secret []byte, salt []byte) ([]byte, error) {
+	return deriveKey(secret, salt, "SIGNATURE", 64)
+}
```
simondeziel: Those should be easy to add unit tests for, if you feel like it :)

lxd/devlxd.go
```diff
@@ -487,8 +489,24 @@ func registerDevLXDEndpoint(d *Daemon, apiRouter *mux.Router, apiVersion string,
...
+
+		// Check if the caller has a bearer token.
+		isBearerRequest, token, subject := bearer.IsRequest(r, d.globalConfig.ClusterUUID())
+		if isBearerRequest {
+			bearerRequestor, err := bearer.Authenticate(token, subject, d.identityCache, api.IdentityTypeBearerTokenDevLXD)
+			if err == nil {
+				requestor = *bearerRequestor
+			} else if !auth.IsDeniedError(err) {
+				// Log any internal errors
+				logger.Warn("Failed to authenticate DevLXD bearer token", logger.Ctx{"err": err})
```
markylaing: Just return an error. Easy to relax restriction but hard to restrict if a problem

MusicDin: I would just like to confirm this is desired because that would mean that passing an invalid token in authorization header would fail the request, but passing no header at all would succeed?

Shouldn't the access handler fail if the user is required to be authenticated, but is not? Therefore, can we skip both the log and the error, and let access handler handle it?

MusicDin: Ahh ignore me, this is in case something else went wrong. +1 for error

markylaing:
> I'm just marking as draft for a minute while I add a couple of tests for token expiration

Token expiration tests were failing because I added a minute leeway to the jwt parser during authentication. I will update to not allow leeway and always use UTC time. We expect that there is no time skew between cluster members for the auth secret and this further requires no time skew.

lxc/auth.go
markylaing: Fixed, thanks for spotting :)

lxd/devlxd.go
markylaing: @tomponline and I discussed this earlier. We said we'd reject the request if a token is sent that is not valid, regardless or the error. I didn't do this for the reason you have mentioned above

lxd/identities.go
markylaing: This is because identities can always view themselves. When listing identities there is a permission checker which filters out identities the caller cannot see.

markylaing:
This will be ready again once #16268 is merged.

markylaing:
Ah.. forgot to change the tests after immediately rejecting requests with bad tokens. Will fix in the morning

lxd/db/cluster/secrets.go
```diff
@@ -301,3 +301,64 @@ func deleteSecretsByID(ctx context.Context, tx *sql.Tx, ids ...int) error {
 
 	return nil
 }
+
+// GetAllBearerIdentitySigningKeys returns a map of identity ID to token signing keys. It should only be used to refresh
```
tomponline: Just a nit, no need to change, but comments like this look pretty odd to me, why not something like this?

```
// GetAllBearerIdentitySigningKeys returns a map of identity ID to token signing keys. 
// It should only be used to refresh the identity cache.
```

markylaing:
Have now updated the DevLXD test suite to expect failures when sending an invalid token. The tests now perform exact equality on the returned error messages too.

lxd/devlxd.go
markylaing: This is now changed to always reject invalid tokens, regardless of the endpoint they're hitting. We can relax this restriction later if required.

lxd/db/cluster/secrets.go
```diff
@@ -301,3 +301,64 @@ func deleteSecretsByID(ctx context.Context, tx *sql.Tx, ids ...int) error {
...
+	err := query.Scan(ctx, tx, q, scanFunc, EntityType(entity.TypeIdentity), SecretTypeBearerSigningKey)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get auth secrets: %w", err)
+	}
+
+	return identityIDToSigningKey, nil
+}
+
+// DeleteBearerIdentitySigningKey deletes any signing keys for the identity.
+func DeleteBearerIdentitySigningKey(ctx context.Context, tx *sql.Tx, identityID int) (int64, error) {
```
tomponline: I'm not sure about this pattern of returning rowsAffected for a delete function call.
Whats the expected use case here? Allow the caller to be able to detect if the entity existed to be deleted?
If so, why not return an api.StatusErr not found if RowsAffected <= 0, this would be more consistent with the errors we return when getting a row right?

tomponline: Not a blocker right now, but just stood out as a little odd to me.

tomponline: The docblock should at least state what the int64 return value is though as otherwise it requires the caller to inspect the implementation of the function to see what it does. This shouldn't be required.

tomponline:
@markylaing can be rebased

lxd/db/cluster/secrets.go
markylaing: Yeah we don't have a convention for this. I just have a line guide in my IDE for 120 characters and tend to add a newline when I reach it. 

Maybe we can just have a convention like Ruth suggested for docs where we always add a newline per sentence? Then it makes changes clear in diffs

lxd/db/cluster/secrets.go
markylaing: This was to be able to re-use the function in different contexts. For example:
- Token revocation
  - Return 404 if no rows deleted
  - Return 200 if 1 or more rows were deleted
- Token rotation
  - Don't check number of rows deleted.

Now that I think about it. This would be much clearer if the function did that logic itself and returned an appropriate api.StatusError. Will fix.

lxd/db/cluster/secrets.go
tomponline: Yeah I tend to try and break things up by sentence by default, but also not have an unlimited line length to enable effective review in Github. So I would say use your judgement.

tomponline:
@markylaing can be rebased, and all standalone tests failing

lxd/auth/encryption/sign.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/canonical/lxd/shared"
+)
+
+const (
+	// defaultExpiry is used to set an expiry of "never", meaning 10 years.
+	defaultExpiry = "10y"
+)
+
+// GetSignedJWT generates and signs a token. If a UUID salt is provided, a unique signing key is derived from the secret using
```
tomponline: Please can you expand this docblock to explain the different use-cases for the different modes this function provides (salted and unsalted) - perhaps even it should be split into different functions, one for each purpose.

Right now the docblock is a bit abstract and depends on the caller understanding the implications of the different modes, which I feel we could improve here and provide guidance around the use-cases.

lxd/auth/encryption/sign.go
```diff
@@ -0,0 +1,58 @@
...
+// the salt and the UUID is set as the token ID (jti). Otherwise, the token is directly signed by the secret.
+func GetSignedJWT(secret []byte, salt *uuid.UUID, subject string, clusterUUID string, expiry string) (string, error) {
```
tomponline: Can this function take a time.Time for expiry rather than a string, otherwise you're tying this function to the specific configuration language used by shared.GetExpiry. Lets decouple as its not the concern of this function how the user defines the expiry in the DB/API.

lxd/identities.go
```diff
@@ -128,6 +144,43 @@ var oidcIdentityCmd = APIEndpoint{
 	},
 }
 
+var bearerIdentityCmd = APIEndpoint{
+	Name:        "identity",
+	Path:        "auth/identities/bearer/{nameOrIdentifier}",
+	MetricsType: entity.TypeIdentity,
+
+	Get: APIEndpointAction{
+		Handler:       getIdentity,
```
tomponline, line 159: The handler names are not aligned with the convention of `<entity><verb>` of the rest of the access handler names.

tomponline: Now I look closer it seems that the identities endpoints have all gone their own way with the naming convention. We should bring them back inline with the rest of the project.

lxd/identities.go
```diff
@@ -293,6 +346,211 @@ func createIdentityTLS(d *Daemon, r *http.Request) response.Response {
...
+		return nil
+	})
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	networkCert := s.Endpoints.NetworkCert()
+	serverCert := s.ServerCert()
+	notify := newIdentityNotificationFunc(s, r, networkCert, serverCert)
+	lc, err := notify(lifecycle.IdentityCreated, api.AuthenticationMethodBearer, newIdentityID.String(), false)
```
tomponline, line 433: A little comment explaining why updateCache is false would be useful.

lxd/identities.go
```diff
@@ -293,6 +346,211 @@ func createIdentityTLS(d *Daemon, r *http.Request) response.Response {
...
+			return err
+		}
+
+		return nil
+	})
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	token, err := encryption.GetSignedJWT(secret, nil, id.Identifier, s.GlobalConfig.ClusterUUID(), req.Expiry)
```
tomponline: By separating the parsing of req.Expiry from this function, it allows you do validate the incoming request more before rotating the (potentially existing) token.

lxd/identities.go
```diff
@@ -293,6 +346,211 @@ func createIdentityTLS(d *Daemon, r *http.Request) response.Response {
...
+		return response.SmartError(err)
+	}
+
+	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+		rowsAffected, err := dbCluster.DeleteBearerIdentitySigningKey(ctx, tx.Tx(), id.ID)
+		if err != nil {
+			return fmt.Errorf("Failed to revoke token: %w", err)
+		}
+
+		if rowsAffected == 0 {
```
tomponline: Yeah this would be better handled inside DeleteBearerIdentitySigningKey in my view.

lxd/auth/bearer/bearer.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	// Expect the issuer to be "lxd:{cluster_uuid}".
+	expectIssuer := "lxd:" + clusterUUID
+	if issuer != expectIssuer {
+		return false, "", ""
+	}
+
+	// Expect the audience to be "lxd:{cluster_uuid}" (the token is signed symmetrically).
+	gotAudience, err := t.Claims.GetAudience()
```
tomponline: would `audience` be sufficient here?

lxd/auth/bearer/bearer.go
```diff
@@ -0,0 +1,108 @@
...
+// type. It then verifies that the token was signed by the secret associated with that identity, and that the token has
+// not expired.
+func Authenticate(token string, subject string, identityCache *identity.Cache, expectedIdentityType string) (*request.RequestorArgs, error) {
+	// Get the identity from the cache by the subject.
+	entry, err := identityCache.Get(api.AuthenticationMethodBearer, subject)
+	if err != nil {
+		return nil, err
+	}
+
+	// Validate that it is of the expected type. (When/if we have token authentication to the main API, we want to make
```
tomponline: @markylaing wrt to this point and the `"lxd:{cluster_uuid}"` audience string above.

How about we use a different audience for devlxd tokens vs future general API tokens?

We could then validate the tokens are for use with the specific API earlier in the request.

E.g. `"devlxd:{cluster_uuid}"`

lxd/db/cluster/secrets.go
markylaing: I've updated a few things in the database handling.
1. Comments have newlines for new sentences (long sentences are still wrapped at 120 characters).
2. The `DeleteBearerIdentitySigningKey` function returns:
    i. `nil` if one key is deleted.
    ii. `api.StatusError` with 404 if no key is deleted.
    iii. A plain error if more than one key is deleted.
3. `RotateBearerIdentitySigningKey` allows either `nil` or `404` when deleting an existing key. If an identity has more than one key associated with them (should never happen), then the keys cannot be rotated. This would need to be fixed with a patch or something.
4. `GetAllBearerIdentitySigningKeys` handles identities found to have multiple signing keys and removes them from the returned map so they cannot authenticate. A list of offenders is returned to the caller. When updating the cache, if this list has length greater than zero, a warning is logged. This is so that any internal error leading to an identity with multiple keys:
    i. Does not prevent the identity cache from refreshing.
    ii. Prevents offending identities from authenticating.

lxd/auth/bearer/bearer.go
markylaing: That is a very good point and is exactly what the audience is for!! Wonderful. Will update. Thanks.

lxd/devlxd.go
```diff
@@ -487,8 +488,25 @@ func registerDevLXDEndpoint(d *Daemon, apiRouter *mux.Router, apiVersion string,
...
-		err := request.SetRequestor(r, d.identityCache, request.RequestorArgs{Protocol: request.ProtocolDevLXD})
+		var requestor request.RequestorArgs
+
+		// Set [request.ProtocolDevLXD] by default identify this request as coming from the /dev/lxd socket.
+		requestor.Protocol = request.ProtocolDevLXD
+
+		// Check if the caller has a bearer token.
+		isBearerRequest, token, subject := bearer.IsRequest(r, d.globalConfig.ClusterUUID())
+		if isBearerRequest {
+			bearerRequestor, err := bearer.Authenticate(token, subject, d.identityCache, api.IdentityTypeBearerTokenDevLXD)
```
tomponline: After this point the `requestor.Protocol` value with be `api.AuthenticationMethodBearer` right?

Does that mean we've now lost the ability to identify this request as a devlxd request using requestor.Protocol.

I'm thinking this is going to be a problem for logging, API metrics, debugging, and potentially other use cases.

I'm more concerned because `api.AuthenticationMethodBearer` will likely be used for non-devlxd purposes in the future.

How do we identify these requests as devlxd requests after this point?

lxc/auth.go
```diff
@@ -933,6 +940,48 @@ func (c *cmdIdentityCreate) createTLSIdentity(remote string, name string, certFi
 	return nil
 }
 
+func (c *cmdIdentityCreate) createBearerIdentity(remoteName string, identityName string, identityType string) error {
```
tomponline, line 943: missing docblock

lxd-agent/devlxd.go
```diff
@@ -71,14 +72,20 @@ func devLXDServer(d *Daemon) *http.Server {
...
-func getDevLXDVsockClient(d *Daemon) (lxd.DevLXDServer, error) {
+func getDevLXDVsockClient(d *Daemon, r *http.Request) (lxd.DevLXDServer, error) {
 	// Try connecting to LXD server.
 	client, err := getClient(d.serverCID, int(d.serverPort), d.serverCertificate)
 	if err != nil {
 		return nil, err
 	}
 
-	server, err := lxd.ConnectDevLXDHTTPWithContext(context.Background(), nil, client)
+	args := &lxd.ConnectionArgs{}
```
tomponline, line 82: Please can you add a comment here, something like:

> Extract optional bearer token from client and pass it to the LXD server for authentication.

lxd/devlxd.go
```diff
@@ -118,10 +118,21 @@ func devLXDAPIGetHandler(d *Daemon, r *http.Request) response.Response {
 		state = api.Started
 	}
 
+	requestor, err := request.GetRequestor(r.Context())
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	clientAuth := "untrusted"
```
tomponline, line 126: now feels like a good time to introduce constants for these 2 values as they are now being used in multiple places and will be used for comparisons too. Probably need to be in shared/api.

tomponline (changes requested):
Thanks @markylaing this is looking really good.

Mostly my comments are nits/style suggestions.

There were just 3 main issues:

- The lost of protocol devlxd when using a bearer token potentially confusing future external bearer requests with devlxd requests.
- encryption.GetSignedJWT to accept a time.Time for expiry.
- The devlxd jwt audience.

lxd/auth/encryption/sign.go
markylaing: I've simplified this to only issue DevLXD tokens and added a comment about what it creates including how the claims are populated.

lxd/auth/encryption/sign.go
markylaing: Done.

lxd/identities.go
markylaing: Added.

lxd/identities.go
markylaing: Yep done.

lxd/auth/bearer/bearer.go
markylaing: Done.

lxd/devlxd.go
markylaing: The identity is saved in the `request.Requestor`, so we can do e.g.:
```
requestor, err := request.GetRequestor(r.Context())
if err != nil {
    return err
}

identityType := requestor.CallerIdentityType()
if identity == nil {
    return errors.New("No identity type")
}

if identityType.Name() == api.IdentityTypeDevLXDTokenBearer {
   // Authenticated as bearer identity over devlxd
   // Do stuff ...
}
```

markylaing:
Assuming the tests pass I've addressed the majority of comments but unfortunately now I need to go, sorry!

lxd/devlxd.go
tomponline: True. But I am still concerned that we may end up adding logic that checks for `Protocol == devlxd` and forget that sometimes the `Protocol` can be `bearer` for devlxd requests.

For example we are currently sending protocol to loki:

```go
context["requester-protocol"] = lifecycleEvent.Requestor.Protocol
```

So once external bearer concept is added these requests become indistinguishable from authenticated devlxd requests.

Same goes for `EventLifecycleRequestor` and `ToLogging`.

I'll add this point along with the other outstanding issues to a jira card that we can discuss when you get back.

Thanks!

lxd/auth/encryption/sign.go
```diff
@@ -0,0 +1,52 @@
...
+// GetDevLXDBearerToken generates and signs a token for use with the DevLXD API. For claims it has:
+// - Subject (sub): Identity identifier (UUID)
+// - Issuer (iss): "lxd:{cluster_uuid}"
+// - Audience (aud): "devlxd:{cluster_uuid}"
+// - Not before (nbf): time now (UTC)
+// - Issued at (iat): time now (UTC)
+// - Expiry (exp): The given time (UTC).
+func GetDevLXDBearerToken(secret []byte, identityIdentifier string, clusterUUID string, expiresAt time.Time) (string, error) {
+	if expiresAt.Location() != time.UTC {
+		expiresAt = expiresAt.UTC()
```
tomponline, line 34: you could just call this line always - it would be the same result.

lxd/db/cluster/secrets.go
```diff
@@ -301,3 +304,97 @@ func deleteSecretsByID(ctx context.Context, tx *sql.Tx, ids ...int) error {
...
+		// Check the identity ID has not already been determined to have multiple keys.
+		if slices.Contains(identityIDsWithMultipleSigningKeys, identityID) {
+			return nil
+		}
+
+		// Check our running map, if a signing key is already present for the ID, delete the existing one from the map
+		// and append the identity ID to the list of offenders.
+		_, ok := identityIDToSigningKey[identityID]
+		if ok {
+			delete(identityIDToSigningKey, identityID)
```
tomponline, line 335: no need to delete here, because you're replacing it on line 341 anyway

lxd/db/cluster/secrets.go
```diff
@@ -301,3 +304,97 @@ func deleteSecretsByID(ctx context.Context, tx *sql.Tx, ids ...int) error {
...
 	return nil
 }
+
+// GetAllBearerIdentitySigningKeys returns a map of identity ID to token signing keys.
+// It should only be used to refresh the identity cache.
+// Any identities with more than one signing key (this should never happen) are omitted from the returned map and added
+// to the list of identity IDs with multiple signing keys.
+// This so that the identity cache refresh handling can handle this at the call site (e.g. so as not to prevent
+// refreshing the cache for all identities).
+func GetAllBearerIdentitySigningKeys(ctx context.Context, tx *sql.Tx) (identityIDToSigningKey map[int]AuthSecretValue, identityIDsWithMultipleSigningKeys []int, err error) {
```
tomponline, line 314: I don't really like how we have special handling in multiple places (and call sites) for something that should never happen.

Rather than bake into the code detection and handling for a scenario that shouldn't even happen, why do we put a trigger in the DB to prevent it, that way if we ever try to do it accidentally it will fail at source.

lxd/identities.go
```diff
@@ -2149,10 +2150,20 @@ func updateIdentityCache(d *Daemon) {
...
+
+		if len(identityIDsWithMultipleSigningKeys) > 0 {
+			logger.Warn("Found identities with multiple signing keys - these will be omitted from the identity cache", logger.Ctx{"identity_ids": identityIDsWithMultipleSigningKeys})
+		}
+
 		return nil
 	})
 	if err != nil {
-		logger.Warn("Failed reading certificates from global database", logger.Ctx{"err": err})
+		logger.Warn("Failed reading identities from global database", logger.Ctx{"err": err})
```
tomponline, line 2166: ideally would have been a separate commit so we can backport the fix without the bearer stuff

lxd/auth/bearer/bearer.go
```diff
@@ -0,0 +1,104 @@
...
+		return false, "", ""
+	}
+
+	// There must be a subject
+	sub, err := t.Claims.GetSubject()
+	if err != nil {
+		return false, "", ""
+	}
+
+	// Expect the issuer to be "lxd:{cluster_uuid}".
```
tomponline, line 44: Comment out of date

tomponline (approved):
LGTM thanks!

I'll open a jira card for you with the follow up items and discussion points.

lxd/auth/bearer/bearer.go
markylaing: This one is correct. It's the audience that's changed to `devlxd:{cluster_uuid}`.

PR #16226, opened 2025-08-16
Haproxy

simondeziel:


simondeziel:
@minaelee The new reference HAProxy config should replace the old one that was inlined in the docs. Since the new one needs to be generated to be readable/useful, I'd like to discuss with you how we could do that and avoid duplicating it. Thanks!

doc/authentication.md
```diff
@@ -156,84 +156,16 @@ To enable this feature, set the following server configuration:
 - {config:option}`server-acme:acme.agree_tos`: Must be set to `true` to agree to the ACME service's terms of service.
 - {config:option}`server-acme:acme.ca_url`: The directory URL of the ACME service. By default, LXD uses "Let's Encrypt".
 
-For this feature to work, LXD must be reachable from port 80.
-This can be achieved by using a reverse proxy such as [HAProxy](http://www.haproxy.org/).
+LXD currently only supports the [`HTTP-01`](https://letsencrypt.org/docs/challenge-types/#http-01-challenge) challenge and this requires handling incoming HTTP requests on port 80.
```
minaelee:
```suggestion
LXD supports the [`HTTP-01 challenge`](https://letsencrypt.org/docs/challenge-types/#http-01-challenge), which requires handling incoming HTTP requests on port 80.
```

doc/authentication.md
```diff
@@ -156,84 +156,16 @@ To enable this feature, set the following server configuration:
...
+This can be achieved by using a reverse proxy such as [HAProxy](https://www.haproxy.org/).
 
-Here's a minimal HAProxy configuration that uses `lxd.example.net` as the domain.
+Here's a HAProxy configuration that uses `lxd.example.net` as the domain.
```
minaelee:
```suggestion
The HAProxy configuration example below uses `lxd.example.net` as the domain.
```

doc/authentication.md
```diff
@@ -156,84 +156,16 @@ To enable this feature, set the following server configuration:
...
 After the certificate has been issued, LXD will be reachable from `https://lxd.example.net/`.
+It applies filtering to minimize the amount of undesired traffic coming from the Internet reaching the protected LXD cluster.
```
minaelee:
```suggestion
It applies filtering to minimize the amount of undesired traffic coming from the internet reaching the protected LXD cluster.
```

simondeziel:
@minaelee no matter what I tried, generating/rendering the HAProxy config at build time didn't work on RTD despite working locally.

Anyway, I switched the approach to instead put the reference config in the docs' Markdown and simply use `sed` to carve it out for reuse in the tests. It's not pretty but it works and would fail loudly if it breaks due to unexpected changes in the docs.

tomponline (approved):
Excellent, thanks!

PR #16229, opened 2025-08-18
doc: add security hardening how-to guide

minaelee:
This PR:
- Adds a new how-to guide for security hardening
- Adds a few missing reference link targets in pages that the hardening guide needs for linking
- Updates `related_links.yaml`, the how-to index page for navigation, and spelling exceptions in `.custom_wordlist.txt` to accommodate the new guide
- Moves a section about container name leakage from the security explanation guide to the hardening guide

### DIFF
--- a/doc/howto/security_harden.md
+++ b/doc/howto/security_harden.md
@@ -206,6 +206,16 @@ sudo snap set lxd daemon.user.group=lxdusers
 
 You can confine users to specific projects, which can be configured with stricter restrictions to prevent misuse. For details, see: {ref}`projects-confine-users`, {ref}`exp-projects`, and {ref}`restricted-tls-certs`.
 
+(howto-security-harden-name-leakage)=
+### Prevent name leakage
+
+The default server configuration makes it possible to list all cgroups on a system, and by extension, all running containers. Prevent container name leakage by blocking access to `/sys/kernel/slab` and `/proc/sched_debug` before you start any containers. To do so, run:
+
+```bash
+chmod 400 /proc/sched_debug
+chmod 700 /sys/kernel/slab/
+```
+
 (howto-security-harden-host)=
 ## Harden the LXD host OS
 


doc/howto/security_harden.md
```diff
@@ -0,0 +1,251 @@
...
+### Use a supported version
+
+Use only supported LTS releases or the latest feature release of LXD, and ensure that you update it regularly to receive security updates and bugfixes. See: {ref}`ref-releases`.
+
+(howto-security-harden-disable-unused)=
+### Disable unused features
+
+If you are not using certain LXD features, disable them to reduce the attack surface. For example:
+
+- Disable the UI: `lxc config set ui.enable false`
```
simondeziel: ATM, there is not such key. One has to disable it at the the snap level:
```
sudo snap set lxd ui.enable=false
sudo systemctl reload snap.lxd.daemon
```

It's too bad that `lxc config set ui.enable` doesn't exist as it feels much more natural and would have been much easier than fiddling with the snap config. Maybe something to consider @tomponline and @edlerd ?

doc/howto/security_harden.md
```diff
@@ -0,0 +1,251 @@
...
+```bash
+sudo lxc profile create hardened1
+sudo lxc profile set hardened1 limits.cpu=2 limits.memory=4GiB limits.processes=500
+sudo lxc profile add <my-container> hardened1
+```
+
+(howto-security-harden-nesting-disable)=
+### Disable container nesting
+
+The instance configuration option {config:option}`instance-security:security.nesting` enables nested container capability. This increases complexity and can broaden the attack surface. The default for this setting is `false`. Do not set this to `true` unless absolutely needed.
```
simondeziel, line 96: We have considered the idea of possibly making `security.nesting=true` in the past so I'd hope that doesn't equates to weakening the security too much.

I feel like it's a tradeoff because with nesting enabled, even with a single layer of containerization, you are then able to leverage namespace features inside your container to for example, use `systemd` hardening on various units/services running in the said container.

doc/howto/security_harden.md
```diff
@@ -0,0 +1,251 @@
...
+
+Setting this option to `true` is especially dangerous in combination with {config:option}`instance-security:security.privileged` set to `true` because it provides root access to the host.
```
simondeziel, line 98: That's good to stress this one out though as that combo is as bad as it gets AFAIK.

doc/howto/security_harden.md
```diff
@@ -0,0 +1,251 @@
...
+
+(howto-security-harden-isolate)=
+### Isolate containers
+
+If the containers on a bridge network do not need to share data with each other, enable {config:option}`instance-security:security.idmap.isolated` on the bridge NIC. This causes the NIC to use non-overlapping UID/GID maps for each container, preventing potential DoS attacks on other containers.
```
simondeziel: idmap isolation is about giving a dedicated slice of UID/GID to each instances so that they don't each get the same slice.

As such, it is not related with bridges, NIC or networking.

doc/howto/security_harden.md
```diff
@@ -0,0 +1,251 @@
...
+(howto-security-harden-logging-auditd)=
+### Use `auditd` rules
+
+Use `auditd` rules to track LXD command execution and configuration file changes.
+
+Configure the audit daemon to track all commands to the LXD daemon:
+
+```bash
+-a always,exit -F path=/usr/sbin/lxc -p x -k lxd_execution
+-a always,exit -F path=/usr/sbin/lxd -p x -k lxd_execution
```
simondeziel:
```suggestion
-a always,exit -F path=/snap/bin/lxc -p x -k lxd_execution
-a always,exit -F path=/snap/bin/lxd -p x -k lxd_execution
```

doc/howto/security_harden.md
minaelee: Thanks, let me know if you see any issues with the rewrite:

```
If a set of containers do not need to share data with each other, enable the instance option 
{config:option}`instance-security:security.idmap.isolated` on each one. This configures them to use unique 
UID/GID maps, preventing potential {abbr}`DoS (Denial of Service)` attacks from one container to another. 
Only unprivileged containers can use this option.
```

doc/howto/security_harden.md
minaelee: Thanks to @mihalicyn for pointing that one out :)

doc/howto/security_harden.md
simondeziel: Yep, LGTM, thanks!

doc/howto/security_harden.md
```diff
@@ -0,0 +1,256 @@
...
+
+This is especially important for containers and their {config:option}`instance-resource-limits:limits.cpu`, {config:option}`instance-resource-limits:limits.memory`, and {config:option}`instance-resource-limits:limits.processes` options, which by default are set without limits. Review the {ref}`instance-options-limits` reference guide for other options you might want to restrict.
+
+Rather than applying these limits on a per-instance basis, use either {ref}`projects`, {ref}`profiles <images-profiles>`, or both. Example syntax for using a profile:
+
+```bash
+sudo lxc profile create hardened1
+sudo lxc profile set hardened1 limits.cpu=2 limits.memory=4GiB limits.processes=500
+sudo lxc profile add <my-container> hardened1
+```
```
edlerd: Should we give an example of a profile that sets all the things mentioned in the guide?
We can also set `security.nesting` and `security.idmap.isolated` in the hardened profile example.

doc/howto/security_harden.md
```diff
@@ -0,0 +1,256 @@
...
+## Instance security
+
+Along with the recommendations below, review all {ref}`instance security options <instance-options-security>` for further options that might be relevant to your setup.
+
+(howto-security-harden-unprivileged)=
+### Use unprivileged containers
+
+By default, LXD containers are unprivileged. If you need to use privileged containers, make sure to put appropriate security measures in place. For more information, see: {ref}`container-security`.
+
+(howto-security-harden-instance-resource-limits)=
```
edlerd, line 86: There is also the `LXD_UNPRIVILEGED_ONLY` environment variable we might want to mention here.

simondeziel: This env variable is indeed related and not widely known. However, IIRC, Tom didn't really liked it and possibly wanted it to be replaced by something more user friendly like a server config or some such. @tomponline do I remember correctly?

doc/howto/security_harden.md
minaelee: Good call, thanks! I've moved the above to its own section called `Use profiles` and added the `security.nesting` and `security.idmap.isolated` options to the example profile.

doc/howto/security_harden.md
```diff
@@ -0,0 +1,275 @@
...
+### Use a supported version
+
+Use only supported LTS releases or the latest feature release of LXD, and ensure that you update it regularly to receive security updates and bugfixes. See: {ref}`ref-releases`.
+
+(howto-security-harden-disable-unused)=
+### Disable unused features
+
+If you are not using certain LXD features, disable them to reduce the attack surface. For example:
+
+- Disable the UI:
```
simondeziel: I'm not qualified to make a good evaluation but my (probably) naive understanding of how the UI works is the following (please correct me):

* LXD serves a bunch of static files (HTML, JS, CSS, etc) to a browser
* browser authenticates to LXD and then makes API calls

So the browser needs to authenticate and make API calls but that's something all clients need to do.

So what attack surface are we actually closing by disabling the UI? Just the serving the static assets that make up the UI?

doc/howto/security_harden.md
```diff
@@ -0,0 +1,275 @@
...
+Also see: {ref}`howto-security-harden-restricted-group`.
+
+(howto-security-harden-remote)=
+### Harden remote API access
+
+For {ref}`authentication`, LXD can use either {abbr}`TLS (Transport Layer Security)` client certificates or OpenID Connect:
+
+- Client certificates:
+   - Ensure that only clients with certificates issued by your trusted Certificate Authority (CA) can connect. The {config:option}`server-core:core.trust_ca_certificates` option is `false` by default. To prevent auto-trusting of CA-signed certificates, ensure it remains disabled.
+   - Regularly audit and remove unused client certificates from the trust store.
```
simondeziel, line 34: Since we recommend issuing short-lived access tokens for OIDC, we should recommend short-lived certificates for private CA owners too. The appropriate is probably specific to each environment. For a university student, a cert valid for a semester would make sense and for a contractor, a cert valid for the term of the work contract would be adequate.

I would also mention cert revocation and managing the CRL (Certificate Revocation List). I don't know if the CRL is checked when using individual trusted entries along with `core.trust_ca_certificates=false`.

doc/howto/security_harden.md
```diff
@@ -0,0 +1,275 @@
...
+
+- OpenID Connect:
+   - Only set `oidc.client.secret` if required by the identity provider.
+   - Configure your OIDC provider to issue short-lived access tokens.
+   - Require multi-factor authentication (MFA) in your identity provider for accounts that can access LXD.
```
simondeziel: I'm not familiar with how IdP work but I would not limit the recommendation for MFA to only accounts granting access to LXD. IMHO, everything you do with your IdP has the potential to be sensitive.

I'd assume that an IdP admin would be able to tweak which accounts can or cannot use LXD so that admin account needs to be equally protected by MFA.

doc/howto/security_harden.md
```diff
@@ -0,0 +1,275 @@
...
+
+For {ref}`authorization`, use {ref}`restricted-tls-certs` or {ref}`fine-grained-authorization` where relevant to your setup.
+
+Refer to the {ref}`authentication` and {ref}`authorization` pages for details.
+
+(howto-security-harden-auth-expiry)=
+### Decrease token expiry
+
+By default, LXD cluster join tokens are set to expire after 3 hours, and LXD remote authentication tokens (used to access the LXD UI and API remotely) are set to never expire. Decrease the expiry times, such as to 15 minutes:
```
simondeziel: It's good to provide tighter recommendations here but I think we really need to set an upper bound on those LXD remote authentication tokens.

Infinity is just a bad default. Changing that default has not been done yet to not catch people off guard but maybe we could go with something generous like 15d? @tomponline WDYT.

tomponline: sounds reasonable to me

doc/howto/security_harden.md
tomponline: Yes thats right it just disables serving static files, doesn't reduce attach surface really.

doc/howto/security_harden.md
tomponline: These docs are oriented towards the LXD snap, and there's no official way to set LXD_UNPRIVILEGED_ONLY at this time for the snap (no `snap set` equivalent AFAIK) . Maybe this is something we could add before the LXD 6 LTS.

cc @mionaalex @mihalicyn

tomponline: @simondeziel id personally prefer this be a sever level config key rather than an env var/snap control though.

simondeziel: Yes, definitely environment variables and snap configs are just not user friendly.

tomponline:
moving back to draft until comments are addressed

doc/howto/security_harden.md
minaelee: Removed to only recommending deletion of unused storage pools and networks.

doc/howto/security_harden.md
minaelee: Updated to remove reference to the current defaults so that this part does not become outdated when the default for remote authentication tokens are changed.

doc/howto/security_harden.md
minaelee: Added recommendation for short-lived certificates and mention of certificate revocation via the CRL.

doc/howto/security_harden.md
minaelee: Marking as resolved as it does not seem like this discussion affects this doc yet (can update once the server level config key is added).

minaelee:
> moving back to draft until comments are addressed

Thanks, all comments should now be addressed.

tomponline:
doc tests failures

tomponline (approved):
Thanks!

doc/howto/security_harden.md
```diff
@@ -0,0 +1,268 @@
...
+### Use profiles
+
+Instead of applying {ref}`instance-options` on a per-instance basis, use either {ref}`projects`, {ref}`profiles <images-profiles>`, or both. This enables you to use a consistent hardened configuration.
+
+The set of commands to create and use a profile below are provided as an example only, including the instance options explicitly mentioned in this guide. Review all instance options and decide if there are other options you want to set for your hardened profile.
+
+```bash
+sudo lxc profile create hardened1
+sudo lxc profile set hardened1 limits.cpu=2 limits.memory=4GiB limits.processes=500
+sudo lxc profile set hardened1 security.idmap.isolated=true security.nesting=true
```
edlerd, line 115: Should this be `security.nesting=false` as recommended above?

tomponline: Good catch!

minaelee: Yikes, thanks. On it.

minaelee: https://github.com/canonical/lxd/pull/16594

PR #16232, opened 2025-08-18
doc: Fix broken symlink

markylaing:


markylaing:
> Thanks! I was working on the same.
> 
> I also noticed a few trailing spaces in `doc/devlxd-api.yaml` also introduced by the same commit that introduced the symlink.

Okey dokey I'll fix that as well then while I'm here, thanks

simondeziel:
`sed -i 's/ *$//' doc/devlxd-api.yaml` should do :)

PR #16249, opened 2025-08-18
Refresh RTD

simondeziel:


simondeziel:
@minaelee I couldn't find why the git repo was "unshallowed" but it seems to work if we don't do this step. It also saves ~15s.

minaelee:
> @minaelee I couldn't find why the git repo was "unshallowed" but it seems to work if we don't do this step. It also saves ~15s.

It was unshallowed because at the time I added sphinx-sitemap as newly required this cycle, there _was_ a default setting in sphinx-sitemap that used sphinx-last-updated-by-git, which is needed to set the sphinx-sitemap's last updated times for each page in the sitemap (preferred for SEO). This also fixed a long-standing bug where the "Last updated on" note at the bottom of each page in the docs was incorrect (showed the same date for every page). However, since that addition, sphinx-sitemap changed its default not to use sphinx-last-updated-by-git, which had the effect of removing the benefits noted above.

So right now it works with a shallow clone, but I will need to update our implementation of sphinx-sitemap so that it needs an unshallowed clone again. Long story short, please leave it as an unshallowed clone.

Edit: Since it's already merged, I can wait to revert the change when I update (and version lock) sphinx-sitemap!

simondeziel:
> > @minaelee I couldn't find why the git repo was "unshallowed" but it seems to work if we don't do this step. It also saves ~15s.
> 
> It was unshallowed because at the time I added sphinx-sitemap as newly required this cycle, there _was_ a default setting in sphinx-sitemap that used sphinx-last-updated-by-git, which is needed to set the sphinx-sitemap's last updated times for each page in the sitemap (preferred for SEO). This also fixed a long-standing bug where the "Last updated on" note at the bottom of each page in the docs was incorrect (showed the same date for every page). However, since that addition, sphinx-sitemap changed its default not to use sphinx-last-updated-by-git, which had the effect of removing the benefits noted above.

Thanks for the explanation. I'm surprised the "Last updated on" note needs a full clone but I must be missing some context. Anyway, in a shallow clone:

```
$ git clone --depth 1 https://github.com/canonical/lxd && cd lxd
$ git log -1 --format="%aD" doc/authentication.md
Mon, 18 Aug 2025 20:45:36 +0100
```

> Edit: Since it's already merged, I can wait to revert the change when I update (and version lock) sphinx-sitemap!

OK, thanks.

minaelee:
> Thanks for the explanation. I'm surprised the "Last updated on" note needs a full clone but I must be missing some context. 

Per: https://sphinx-sitemap.readthedocs.io/en/latest/advanced-configuration.html#adding-last-modified-timestamps

PR #16251, opened 2025-08-19
lxd/storage/backend: Clarify mount approach in `detectUnknownInstanceVolume`

roosterfish:


lxd/storage/backend_lxd.go
```diff
@@ -7543,6 +7543,12 @@ func (b *lxdBackend) detectUnknownInstanceVolume(vol *drivers.Volume, projectVol
 
 	// If the instance is running, it should already be mounted, so check if the backup file
 	// is already accessible, and if so parse it directly, without disturbing the mount count.
+	// It is important to not always run the volume's MountTask.
+	// In a situation in which the instance is still running, but the LXD daemon was restarted,
```
tomponline: > In a situation in which the instance is still running, but the DB was lost and the LXD daemon was restarted, the mount's ref counter is lost and not restored on start up.

roosterfish:
Tested the behavior we discussed in https://github.com/canonical/lxd/pull/16176#discussion_r2282393923.
When running the recovery after restarting the lost LXD daemon with a fresh DB, we get this error back from the unmount operation which describes the issue with the ref counter:

`Error: Failed validation request: Failed checking volumes on pool "zpool": Failed to unmount "/var/snap/lxd/common/lxd/storage-pools/zpool/virtual-machines/microcloud-test_v1": device or resource busy`

lxd/storage/backend_lxd.go

tomponline (approved):

PR #16257, opened 2025-08-19
test/main: allow calling `_foo_setup` and/or `_foo_teardown` if present

simondeziel:


simondeziel:
@markylaing I'm not sure if I captured your suggestion properly but now is the time to clarify any misunderstanding :)

I tried converting the last test I added (`clustering_ha`) but had to backtrack as the setup and teardown phases need to access some stuff that's available only in the primary test function itself. IIRC you had some use cases in mind though so maybe you can refresh my memory.

test/main.sh
```diff
@@ -132,6 +132,13 @@ cleanup() {
   unset SERVER_DEBUG
   unset SHELL_TRACING
 
+  # Run _foo_teardown function if any.
+  TEARDOWN_FUNC="_${TEST_CURRENT#test_}_teardown"
+  if declare -F "${TEARDOWN_FUNC}" >/dev/null; then
+    echo "==> TEARDOWN: ${TEST_CURRENT_DESCRIPTION}"
+    ${TEARDOWN_FUNC}
+  fi
+
```
markylaing: Ideally the teardown func would run after each test as well as during clean up (like the setup func is running before each test).

This way we can move all of the:
```
# Clean up
... do stuff ...
```
blocks from the bottom of each test into the teardown function

I think also moving the teardown after the inspect shell would be helpful, as then the shell will have the exact state.

Have to go now but this is looking great :) thanks

tomponline:
I'm wondering if we shouldn't be looking at replacing the bash testing "framework" with something more proper that would offer this sort of functionality out of the box.

simondeziel:
> I'm wondering if we shouldn't be looking at replacing the bash testing "framework" with something more proper that would offer this sort of functionality out of the box.

It's still based on `bash` but I think MicroOVN folks use [`bats`](https://github.com/bats-core/bats-core). So not a good replacement but I'm not sure of how we can replace all that glue code :/

There is one upside to having extensive testing in `bash`, it gives a good overview of how to use LXD to achieve things.

tomponline:
Closing for now.

PR #16260, opened 2025-08-20
Doc: Create partity between PowerFlex & Pure name tables

roosterfish:
Ensure that both the PowerFlex and Pure docs use the same examples for the naming of volumes. Additionally add some more examples.


roosterfish:
I'll add the missing `Snapshot` row for PowerFlex as part of https://github.com/canonical/lxd/pull/16253#issuecomment-3204478664 when adding the actual prefix.

tomponline (approved):
ta

PR #16261, opened 2025-08-20
lxc/auth: Use shortform identity types

markylaing:
This PR factors out logic that parses the `[remote]<authentication_method>/<name_or_identifier>` argument and changes `<authentication_method>` to `<type>`, where `<type>` now refers to a short form of an identity type.

This will be used in #16225 for the `devlxd` type.

### DIFF
--- a/po/ar.po
+++ b/po/ar.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: lxd\n"
 "Report-Msgid-Bugs-To: lxd@lists.canonical.com\n"
-"POT-Creation-Date: 2025-07-15 11:36+0100\n"
+"POT-Creation-Date: 2025-08-20 15:53+0100\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Automatically generated\n"
 "Language-Team: none\n"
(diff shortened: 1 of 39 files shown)


lxc/auth.go
```diff
@@ -728,6 +728,29 @@ type cmdIdentity struct {
 	global *cmdGlobal
 }
 
+// resolveIdentityTypeShorthand takes a shorthand and returns an authentication method and identity type or an error.
+// If the shorthand resolves to more than one identity type, it returns an empty string for the identity type.
+func (c *cmdIdentity) resolveIdentityArg(identityArg string) (remote string, method string, identityType string, nameOrID string, err error) {
```
simondeziel, line 733: That's the kind of function for which it's trivial to write unit tests for. That said, maybe the function is so simple that it isn't worth it to have tests?

tomponline:
@simondeziel we seem to be encountering more copilot errors since adding the copilot setup file - is that a coincidence do you think?  Or does it just not like @markylaing PRs?

tomponline (approved):

lxc/auth.go
markylaing: The happy path is covered by the `authorization` system test but we could add some incorrect shortform type names there and cover the edge cases?

Generally in LXD I only write unit tests when I'm doing some complex logic. That's when TTD is most helpful for me. I also often write unit tests for database patches and other critical things. Happy to change our approach though :)

tomponline: > The happy path is covered by the `authorization` system test

I'd like to see the integration tests also cover the unhappy paths too to check validation is occurring properly end to end.

Not opposed to unit tests either.

markylaing: I'll add a couple more tests :)

markylaing: https://github.com/canonical/lxd/pull/16266

Issue #12869, opened 2024-02-12
Import keepalive feature and encrypted client key from Incus


simondeziel:
https://github.com/lxc/incus/pull/295 => keepalive feature: keep connection alive for a certain time avoiding the setup cost
https://github.com/lxc/incus/pull/468 => support SSH encrypted key for authentication: allow decrypting the client.key once when combined with the keepalive feature
https://github.com/lxc/incus/pull/481 => followup on PR 468

PR #16267, opened 2025-08-21
UI command

edlerd:
# Done
- add `lxc ui` command to spawn quick access to the UI proxied through the client and authenticated with a url parameter

From Incus, see also
- https://github.com/lxc/incus/pull/1338
- https://github.com/lxc/incus/pull/295
- https://github.com/canonical/lxd/issues/12869

To test this locally:
1. Build the ui from https://github.com/canonical/lxd-ui/pull/1449 with `yarn build`
2. Set LXD_UI variable to the build output, such as `export LXD_UI=/home/foo/lxd-ui/build/ui/`
3. build and run from this branch
4. run `lxc ui` and see the browser open with the ui authenticated, loaded from localhost.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -208,7 +208,7 @@ jobs:
           set -eux
 
           # bin/max (sizes are in MiB)
-          SIZES="lxc 16
+          SIZES="lxc 17
                  lxd-agent 14"
           MIB="$((1024 * 1024))"
 
--- a/client/lxd_oidc.go
+++ b/client/lxd_oidc.go
@@ -273,7 +273,7 @@ func (o *oidcClient) authenticate(issuer string, clientID string, audience strin
 	fmt.Printf("URL: %s\n", u.String())
 	fmt.Printf("Code: %s\n\n", resp.UserCode)
 
-	_ = openBrowser(u.String())
+	_ = OpenBrowser(u.String())
 
 	token, err := rp.DeviceAccessToken(ctx, resp.DeviceCode, time.Duration(resp.Interval)*time.Second, provider)
 	if err != nil {
--- a/client/util.go
+++ b/client/util.go
@@ -274,7 +274,8 @@ type HTTPTransporter interface {
 	Transport() *http.Transport
 }
 
-func openBrowser(url string) error {
+// OpenBrowser opens the specified URL in the default web browser.
+func OpenBrowser(url string) error {
 	var err error
 
 	browser := os.Getenv("BROWSER")
--- a/client/util_test.go
+++ b/client/util_test.go
@@ -205,9 +205,9 @@ func Test_openBrowser(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Setenv("BROWSER", tt.env)
-			err := openBrowser(tt.url)
+			err := OpenBrowser(tt.url)
 			if err != nil {
-				t.Errorf("openBrowser() unexpected error = %v", err)
+				t.Errorf("OpenBrowser() unexpected error = %v", err)
 			}
 		})
 	}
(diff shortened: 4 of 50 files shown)


markylaing:
@edlerd @tomponline I suspect this can be made a lot simpler if we have token based authentication to the main API.

I'm thinking we can have `lxc ui`:
1. Only talk to the local LXD over unix.
2. Set `core.https_address`
3. Create a bearer identity and add the identity to the `admins` group.
4. Issue a token for the identity (set a session expiry).
5. Open a browser and set an `access_token` query parameter in the URL and have the UI save it and set it as an `Authorization: Bearer {token}` header for subsequent requests.

tomponline:
> * Set `core.https_address`

I dont think that is a good idea. 

Think we need more discussion on the spec.

Also need to firm up with @mionaalex when shes back about her requirements for this.

edlerd:
> @edlerd @tomponline I suspect this can be made a lot simpler if we have token based authentication to the main API.
> 
> I'm thinking we can have `lxc ui`:
> 
> 1. Only talk to the local LXD over unix.
> 2. Set `core.https_address`
> 3. Create a bearer identity and add the identity to the `admins` group.
> 4. Issue a token for the identity (set a session expiry).
> 5. Open a browser and set an `access_token` query parameter in the URL and have the UI save it and set it as an `Authorization: Bearer {token}` header for subsequent requests.

I prefer the current approach for a couple of reasons. The main one being the goal of this PR: Easy onboarding for new users. A new user has to run a single command on any ubuntu machine: `lxc ui` and they have the ui authenticated and can start using lxc without any previous knowledge.

Another reason is this works also with other remotes. So if you are a LXD power user with many remotes, you can quickly launch lxc ui for any of your configured remotes and it will again just work, no setup of tls needed for each of your remotes in the browser.

Lastly, I suspect the bearer token handling in the ui would need extra complexity to support. The current solution needs only minimal changes in the ui, see linked PR https://github.com/canonical/lxd-ui/pull/1449 is already everything needed to make the current approach fly.

tomponline:
> > @edlerd @tomponline I suspect this can be made a lot simpler if we have token based authentication to the main API.
> > I'm thinking we can have `lxc ui`:
> > 
> > 1. Only talk to the local LXD over unix.
> > 2. Set `core.https_address`
> > 3. Create a bearer identity and add the identity to the `admins` group.
> > 4. Issue a token for the identity (set a session expiry).
> > 5. Open a browser and set an `access_token` query parameter in the URL and have the UI save it and set it as an `Authorization: Bearer {token}` header for subsequent requests.
> 
> I prefer the current approach for a couple of reasons. The main one being the goal of this PR: Easy onboarding for new users. A new user has to run a single command on any ubuntu machine: `lxc ui` and they have the ui authenticated and can start using lxc without any previous knowledge.
> 
> Another reason is this works also with other remotes. So if you are a LXD power user with many remotes, you can quickly launch lxc ui for any of your configured remotes and it will again just work, no setup of tls needed for each of your remotes in the browser.
> 
> Lastly, I suspect the bearer token handling in the ui would need extra complexity to support. The current solution needs only minimal changes in the ui, see linked PR [canonical/lxd-ui#1449](https://github.com/canonical/lxd-ui/pull/1449) is already everything needed to make the current approach fly.

Yeah i think this approach is pretty neat. It doesn't change the authentication mechanism, you still can use TLS, OIDC or unix, but you can use join tokens and let the `lxc` CLI handle the certificate generation rather than having to faff installing it into the browser.

It also doesn't modify the LXD server configuration to expose it onto the network and would work for remote servers and clusters.

But I'm not sure this is what @mionaalex had in mind so we should do a PoC with demo and see what the response is.

tomponline:
@edlerd closing for now until we've agreed the approach with @mionaalex and @escabo

PR #16268, opened 2025-08-21
Auth: Add encryption package

markylaing:
These are a couple of commits from #16225 that can be moved into a separate PR

### DIFF
--- a/lxd/auth/encryption/key.go
+++ b/lxd/auth/encryption/key.go
@@ -1,0 +1,63 @@
+package encryption
+
+import (
+	"crypto/hmac"
+	"crypto/sha512"
+	"fmt"
+)
+
+var hashFunc = sha512.New
+
+// CookieHashKey returns a key suitable for cookie integrity checks with HMAC-SHA512.
+func CookieHashKey(secret []byte, salt []byte) ([]byte, error) {
+	return deriveKey(secret, salt, "INTEGRITY", 64)
+}
+
+// CookieBlockKey returns a key suitable for cookie encryption with AES-256.
+func CookieBlockKey(secret []byte, salt []byte) ([]byte, error) {
+	return deriveKey(secret, salt, "ENCRYPTION", 32)
+}
+
+// deriveKey uses HMAC to derive a key from a secret, a salt, and a separator. We can use HMAC directly because our
+// initial key material is uniformly random and of sufficient length.
+func deriveKey(secret []byte, salt []byte, usageSeparator string, length uint) ([]byte, error) {
+	if len(salt) < 16 {
+		return nil, errInsufficientSalt
+	}
+
+	if usageSeparator == "" {
+		return nil, errNoUsage
+	}
+
+	if length < 32 {
+		return nil, errKeyTooShort
+	}
+
+	maxSize := hashFunc().Size()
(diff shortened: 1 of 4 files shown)


markylaing:
@simondeziel added tests for this per your comment here https://github.com/canonical/lxd/pull/16225#discussion_r2290908677

tomponline (commented):
LGTM, just the copilot nit on error value.

lxd/auth/encryption/key.go
```diff
@@ -0,0 +1,63 @@
...
+
+	if usageSeparator == "" {
+		return nil, errNoUsage
+	}
+
+	if length < 32 {
+		return nil, errKeyTooShort
+	}
+
+	maxSize := hashFunc().Size()
```
simondeziel, line 36: Could (/should?) that be a global `const` or `var`?

markylaing:
> LGTM, just the copilot nit on error value.

Fixed

lxd/auth/encryption/key_test.go
```diff
@@ -0,0 +1,121 @@
...
+			},
+			want: []byte{0xc6, 0x8a, 0x4a, 0x97, 0x41, 0x73, 0x8c, 0xe3, 0x34, 0xf9, 0x55, 0xc2, 0x31, 0x7f, 0x97, 0xf9, 0x2e, 0xd9, 0xcc, 0x4a, 0x7c, 0xa1, 0xa9, 0xf2, 0xd7, 0x5, 0xbb, 0xa5, 0x7c, 0x33, 0xb6, 0x1c, 0x1c, 0xdd, 0x77, 0xbf, 0x97, 0xd0, 0xd, 0x24, 0x50, 0x23, 0xb2, 0xe2, 0x4f, 0x7b, 0x5b, 0x14, 0xb2, 0xc4, 0x6, 0x11, 0x39, 0x95, 0x52, 0x59, 0x55, 0xa2, 0x79, 0x2d, 0xcf, 0x79, 0x15, 0x7a},
+		},
+		{
+			name: "valid, length 32",
+			args: args{
+				secret:         slices.Repeat([]byte{'0'}, hashFunc().Size()),
+				salt:           slices.Repeat([]byte{'0'}, 16),
+				usageSeparator: "TEST",
+				length:         uint(32),
```
simondeziel: The `unit()` cast is probably not needed.

simondeziel (approved):

markylaing:
> LGTM, just the copilot nit on error value.

This is fixed. I'll just address @simondeziel's comments and we'll be good to go.

lxd/auth/encryption/key.go
markylaing: I don't think it necessarily needs to be at this point because `deriveKey` is not exported. It's only called by the exported functions above

lxd/auth/encryption/key_test.go
markylaing: Fixed.

markylaing:
> > LGTM, just the copilot nit on error value.
> 
> This is fixed. I'll just address @simondeziel's comments and we'll be good to go.

Done. Ready when you are :)

tomponline (approved):

PR #16270, opened 2025-08-21
Backup: Add metadata to the config struct

roosterfish:
As part of the recovery it's important to understand whether one backup config file is newer than another one. This PR adds unexported metadata to the `Config` struct.

As part of custom volume discovery from an instance's backup config, each of the custom volumes gets packaged into its own backup config. Those backup configs derive their last modification time from their parent instance.
When listing all the unknown volumes attached to instances and the ones discovered just by name, we can then consult this last modification time to figure out whether or not the current backup config should be used instead as it might replace one which is outdated or not enriched enough.

Custom volumes recovered only from name have a backup config with a last modification time of zero. So they will be replaced in all cases by backup configs from custom volumes discovered through instance backup configs.

The follow up PR https://github.com/canonical/lxd/pull/16164 can be rebased once this PR is merged as it contains the commits from this PR and makes use of the new metadata.

### DIFF
--- a/lxd/backup/backup_config_utils.go
+++ b/lxd/backup/backup_config_utils.go
@@ -88,7 +88,7 @@ func ConfigToInstanceDBArgs(state *state.State, c *config.Config, projectName st
 // In case the requested format is already present it's a noop.
 func ConvertFormat(backupConf *config.Config, version uint32) (*config.Config, error) {
 	// Create a copy of the original config.
-	copyBackupConf := &config.Config{}
+	copyBackupConf := config.NewConfig(backupConf.LastModified())
 	err := shared.DeepCopy(backupConf, copyBackupConf)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to deep copy backup config: %w", err)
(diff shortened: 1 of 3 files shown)


roosterfish:
The CodeQL checks seems not right, as we do run this a few lines before:

```go
// Defend against path traversal attacks.
err = instancetype.ValidName(instanceRef, false)
if err != nil {
    return nil, fmt.Errorf("Invalid instance name %q: %w", instanceRef, err)
}
```

roosterfish:
> The CodeQL checks seems not right, as we do run this a few lines before:

I don't see why this would not catch path traversal, inside `ValidName` we run `err := validate.IsHostname(instanceName)` which will fail if the `instanceName` contains a non-valid hostname which would be true in case of an attack.

roosterfish:
@tomponline that is the one I was mentioning earlier in the meeting :) The other recovery related PRs are all in draft for now but this one should come next.

tomponline (approved):
LGTM thanks!

PR #16278, opened 2025-08-22
devlxd: Volume management API

MusicDin:
This PR adds the devLXD endpoints that allow volume management, primarily meant for LXD CSI driver.
This includes volume creation, deletion, and modification, as well as instance device management, which allows attaching and detaching volumes.

### Access

The endpoints are disabled by default and require `security.devlxd.management.volumes` to be enabled on a particular instance. Additionally, endpoints `/1.0/storage-pools/*` and `/1.0/instances/*` require a caller to authenticate using a valid devLXD bearer token. Finally, devLXD allows attaching and detaching devices from other instances within the same project, if the identity has sufficient permissions. 

### Ownership

DevLXD allows only management of owned volumes and instance devices. This means that only volumes created by the devLXD identity can be later managed or removed.

Volume owner is stored in volume configuration `volatile.owner`, while instance device ownership is stored in instance configuration `volatile.<device>.devlxd.owner`. In both cases, the value is set to devLXD identity identifier. 

---

Rebase after:
- [x] https://github.com/canonical/lxd/pull/16329
- [x] https://github.com/canonical/lxd/pull/16450
- [x] https://github.com/canonical/lxd/pull/16451

### DIFF
--- a/doc/.custom_wordlist.txt
+++ b/doc/.custom_wordlist.txt
@@ -54,6 +54,7 @@ CRL
 cron
 cryptographic
 cryptographically
+CSI
 CSM
 CSV
 CUDA


lxd/response/response.go
```diff
@@ -51,12 +52,27 @@ type Response interface {
...
+	etag        string
 	contentType string
 	err         error
 }
 
 // Render renders a response for requests against the /dev/lxd socket.
 func (r *devLXDResponse) Render(w http.ResponseWriter, req *http.Request) (err error) {
+	etag := r.etag
+
+	// Strip quotes from ETag if present, as the Etag may be quoted if received
```
tomponline: @MusicDin why are they quoted from the LXD server?

lxd/response/response.go
```diff
@@ -65,6 +81,10 @@ func (r *devLXDResponse) Render(w http.ResponseWriter, req *http.Request) (err e
 			return SmartError(r.err).Render(w, req)
 		}
 
+		if etag != "" {
+			w.Header().Set("ETag", `"`+etag+`"`)
```
tomponline: Needs a comment.

lxd/response/response.go
```diff
@@ -74,6 +94,10 @@ func (r *devLXDResponse) Render(w http.ResponseWriter, req *http.Request) (err e
 		return nil
 	}
 
+	if etag != "" {
+		w.Header().Set("ETag", `"`+etag+`"`)
```
tomponline: why is this one still being quoted too? It doesn't appear to be happening over the vsock? Perhaps im misreading.

lxd/devlxd.go
```diff
@@ -650,6 +652,26 @@ func allowDevLXDPermission(entityType entity.Type, entitlement auth.Entitlement,
 	}
 }
 
+// getDevLXDIdentity extracts the identity of the caller from the request context.
+// The identity is expected to be present and have a valid identifier. Otherwise, an error is returned.
+func getDevLXDIdentity(ctx context.Context) (*identity.CacheEntry, error) {
```
tomponline: Do we need a devlxd specific one of these, it seems to not have anything devlxd specific in it?

lxd/devlxd.go
```diff
@@ -556,7 +556,19 @@ func registerDevLXDEndpoint(d *Daemon, apiRouter *mux.Router, apiVersion string,
...
 
-			return f(d, r, action.Handler)
+			// All API endpoint acctions should either have an
+			// access handler or allow untrusted requests.
+			if action.AccessHandler == nil && !action.AllowUntrusted {
+				return response.DevLXDErrorResponse(api.StatusErrorf(http.StatusInternalServerError, "Access handler not defined for %s %s", r.Method, r.URL.RequestURI()))
+			}
+
+			// If the request is not trusted, only call the handler
+			// if the action allows it.
```
tomponline: No need to wrap the comment

lxd/devlxd.go
```diff
@@ -556,7 +556,19 @@ func registerDevLXDEndpoint(d *Daemon, apiRouter *mux.Router, apiVersion string,
 				return response.DevLXDErrorResponse(api.NewGenericStatusError(http.StatusNotImplemented))
 			}
 
-			return f(d, r, action.Handler)
+			// All API endpoint acctions should either have an
```
tomponline: No need to wrap the comment.

lxd/devlxd_util.go
tomponline, line 1: I'm not keen on these generic "util" files/packages, they tend to become a dumping ground for an assortment of functions and don't well describe their grouping.

Is there a reason we need these new functions specifically in the devlxd context? They dont appear to contain anything devlxd specific. 

Could they go in the `request` package if needed?

lxd/devlxd_util.go
```diff
@@ -63,3 +68,101 @@ func NewRequestWithContext(ctx context.Context, method string, url string, data
...
+// header, body, and status code for later inspection.
+type responseCapture struct {
+	header     http.Header
+	body       *bytes.Buffer
+	statusCode int
+	url        url.URL
+}
+
+// NewResponseCapture creates a new responseCapture for the given request.
+func NewResponseCapture(req *http.Request) *responseCapture {
```
tomponline: Can these go into the `response` package, they dont appear to be devlxd specific?

lxd/devlxd.go
```diff
@@ -146,11 +146,39 @@ func devLXDAPIGetHandler(d *Daemon, r *http.Request) response.Response {
 		clientAuth = "trusted"
 	}
 
+	var supportedDrivers []api.DevLXDServerStorageDriverInfo
+
+	// Include supported storage drivers if the instance has the devLXD volume
+	// management security flag enabled.
+	if inst.ExpandedConfig()[string(devLXDSecurityMgmtVolumesKey)] == "true" {
```
tomponline: shared.IsTrue()

lxd/devlxd.go
```diff
@@ -146,11 +146,39 @@ func devLXDAPIGetHandler(d *Daemon, r *http.Request) response.Response {
...
+		url := api.NewURL().Path("1.0").URL
+		req, err := NewRequestWithContext(r.Context(), http.MethodGet, url.String(), nil, "")
+		if err != nil {
+			return response.DevLXDErrorResponse(err)
+		}
+
+		info := &api.Server{}
+
+		resp := api10Get(d, req)
```
tomponline: Rather than calling `api10Get` here (which gets a lot of other info unnecessarily here) and then discard all but the storage drivers info, please can you factor out the supported storage driver code from `api10Get` into its own function which you can call from `api10Get` and here.

shared/api/devlxd_storage.go
```diff
@@ -0,0 +1,30 @@
...
+	// Pool status (Pending, Created, Errored or Unknown)
+	// Read only: true
+	// Example: Created
+	//
+	// API extension: clustering
+	Status string `json:"status" yaml:"status"`
+
+	// Storage pool configuration map (refer to doc/storage.md)
+	// Example: {"volume.block.filesystem": "ext4", "volume.size": "50GiB"}
+	Config map[string]string `json:"config" yaml:"config"`
```
tomponline, line 55: @MusicDin @markylaing we need to be very careful here as storage pool config can contain passwords to the external storage appliances. 

IIRC we only grant access to storage pool config when you are an admin, is that right @markylaing ?

@MusicDin whats the reason for needing access to this Config map?

Perhaps we need to move the pool level secret keys into the new `secrets` table and relax access to Pool config first before this change lands.

shared/api/devlxd_storage.go
tomponline, line 1: Some weird rejigging of the new struct added in the previous commit here. Looks like it needs a rebase and tidy up.

shared/api/devlxd_storage.go
```diff
@@ -13,18 +13,73 @@ type DevLXDStoragePool struct {
...
+	// Volume name.
+	// Example: foo
+	Name string `json:"name" yaml:"name"`
+
+	// Volume type.
+	// Example: custom
+	Type string `json:"type" yaml:"type"`
+
+	// Migration source.
+	Source StorageVolumeSource `json:"source" yaml:"source"`
```
tomponline: Why do we need this?

lxd/devlxd_storage.go
```diff
@@ -46,3 +47,83 @@ func devLXDStoragePoolGetHandler(d *Daemon, r *http.Request) response.Response {
...
+
+	resp := storagePoolVolumesGet(d, req)
+	_, err = RenderToStruct(req, resp, &vols)
+	if err != nil {
+		return response.DevLXDErrorResponse(err)
+	}
+
+	respVols := make([]api.DevLXDStorageVolume, 0, len(vols))
+	for _, vol := range vols {
+		if vol.Config["volatile.owner"] != identity.Identifier {
```
tomponline: This is easy to miss, lets break it out into a separate function and then add a comment explaining what is happening here.

lxd/devlxd_storage.go
```diff
@@ -127,3 +130,75 @@ func devLXDStoragePoolVolumesGetHandler(d *Daemon, r *http.Request) response.Res
 
 	return response.DevLXDResponse(http.StatusOK, respVols, "json")
 }
+
+func devLXDStoragePoolVolumesPostHandler(d *Daemon, r *http.Request) response.Response {
```
tomponline, line 168: missing doc block

lxd/devlxd_storage.go
```diff
@@ -46,3 +47,83 @@ func devLXDStoragePoolGetHandler(d *Daemon, r *http.Request) response.Response {
...
+	Path: "storage-pools/{poolName}/volumes",
+	Get:  devLXDAPIEndpointAction{Handler: devLXDStoragePoolVolumesGetHandler, AccessHandler: allowDevLXDAuthenticated},
+}
+
+var devLXDStoragePoolVolumesTypeEndpoint = devLXDAPIEndpoint{
+	Path: "storage-pools/{poolName}/volumes/{type}",
+	Get:  devLXDAPIEndpointAction{Handler: devLXDStoragePoolVolumesGetHandler, AccessHandler: allowDevLXDAuthenticated},
+}
+
+func devLXDStoragePoolVolumesGetHandler(d *Daemon, r *http.Request) response.Response {
```
tomponline, line 86: missing doc block

shared/api/devlxd_storage.go
MusicDin: > @MusicDin @markylaing we need to be very careful here as storage pool config can contain passwords to the external storage appliances.

Sorry, this seems to be incorrectly rebased. In the final version (last commit) the storage pool config is not exposed. Only storage volume config.

lxd/devlxd_storage.go
```diff
@@ -127,3 +130,75 @@ func devLXDStoragePoolVolumesGetHandler(d *Daemon, r *http.Request) response.Res
...
+	if err != nil {
+		return response.DevLXDErrorResponse(api.StatusErrorf(http.StatusInternalServerError, "Failed decoding request body: %w", err))
+	}
+
+	if volType == "" {
+		volType = "custom"
+	}
+
+	// Reject non-custom volume type.
+	if volType != "custom" {
```
tomponline: Can we check against an existing constant here and line 158?

lxd/devlxd_storage.go
```diff
@@ -127,3 +130,75 @@ func devLXDStoragePoolVolumesGetHandler(d *Daemon, r *http.Request) response.Res
...
+		return response.DevLXDErrorResponse(api.NewStatusError(http.StatusBadRequest, "Only custom storage volumes can be created"))
+	}
+
+	if vol.Type != "" && vol.Type != volType {
+		return response.DevLXDErrorResponse(api.NewStatusError(http.StatusBadRequest, "URL volume type does not match the volume type in body"))
+	}
+
+	// Set caller's identity ID as the volume owner.
```
tomponline: Please expand the comment to explain why.

shared/api/devlxd_storage.go
MusicDin: For volume cloning. I can drop this part and add it in a separate PR if you would like me to?

lxd/devlxd_storage.go
```diff
@@ -202,3 +203,80 @@ func devLXDStoragePoolVolumesPostHandler(d *Daemon, r *http.Request) response.Re
...
+		Pool:        vol.Pool,
+		Type:        vol.Type,
+		Config:      vol.Config,
+		Location:    vol.Location,
+	}
+
+	return respVol, etag, nil
+}
+
+func devLXDStoragePoolVolumeGetHandler(d *Daemon, r *http.Request) response.Response {
```
tomponline, line 305: missing doc block

lxd/devlxd_storage.go
```diff
@@ -280,3 +281,46 @@ func devLXDStoragePoolVolumeGetHandler(d *Daemon, r *http.Request) response.Resp
 
 	return response.DevLXDResponseETag(http.StatusOK, vol, "json", etag)
 }
+
+func devLXDStoragePoolVolumeDeleteHandler(d *Daemon, r *http.Request) response.Response {
```
tomponline, line 431: missing doc block

lxd/devlxd_storage.go
```diff
@@ -282,6 +283,83 @@ func devLXDStoragePoolVolumeGetHandler(d *Daemon, r *http.Request) response.Resp
 	return response.DevLXDResponseETag(http.StatusOK, vol, "json", etag)
 }
 
+func devLXDStoragePoolVolumePutHandler(d *Daemon, r *http.Request) response.Response {
```
tomponline, line 339: missing doc block

lxd/storage/utils.go
```diff
@@ -545,6 +545,15 @@ func poolAndVolumeCommonRules(vol *drivers.Volume) map[string]func(string) error
...
 		rules["volatile.uuid"] = validate.Optional(validate.IsUUID)
+
+		// lxdmeta:generate(entities=storage-btrfs,storage-cephfs,storage-ceph,storage-dir,storage-lvm,storage-zfs,storage-powerflex,storage-pure; group=volume-conf; key=volatile.owner)
+		//
+		// ---
+		//  type: string
+		//  defaultdesc: owner's identity ID
+		//  shortdesc: The identity ID of the volume owner
+		//  scope: global
+		rules["volatile.owner"] = validate.Optional(validate.IsUUID)
```
tomponline: looking at the previous commit, for the device level `devlxd.owner` key, should this be also `volatile.devlxd.owner` for consistency?

MusicDin: Yes, I think this makes the key’s scope self-explanatory

lxd/devlxd_instances.go
```diff
@@ -0,0 +1,21 @@
...
+// newDevLXDDeviceAccessValidator returns a device validator function that
+// checks if the given device is accessible by the devLXD.
+//
+// For example, disk device is accessible if the appropriate security flag
+// is enabled on the instance and the device represents a custom volume.
+func newDevLXDDeviceAccessValidator(inst instance.Instance) devLXDDeviceAccessValidator {
+	diskDeviceAllowed := hasInstanceSecurityFeatures(inst, devLXDSecurityMgmtVolumesKey)
+
+	return func(device map[string]string) bool {
+		return filters.IsCustomVolumeDisk(device) && diskDeviceAllowed
```
tomponline: Suggest switching round the conditions here as its cheaper to check a boolean first.

lxd/devlxd_instances.go
```diff
@@ -3,10 +3,29 @@ package main
...
 	"github.com/canonical/lxd/lxd/instance"
+	"github.com/canonical/lxd/shared/api"
 )
 
 type devLXDDeviceAccessValidator func(device map[string]string) bool
 
+// getOwnedDevices extracts instance devices that are owned by the provided identity.
+// Two maps are returned, one containing owned devices, and the other containing the remaining
+// devices.
+func getOwnedDevices(inst api.Instance, identityID string) (ownedDevices map[string]map[string]string, otherDevices map[string]map[string]string) {
```
tomponline: remember you are in the `main` package and so IDEs will suggest functions to use for the whole of LXD.

So make sure that if a function or variable in the main package is devlxd specific (as this one is) that its prefixed with `devLXD`.

lxd/devlxd_instances.go
```diff
@@ -3,10 +3,29 @@ package main
...
 	"github.com/canonical/lxd/lxd/device/filters"
 	"github.com/canonical/lxd/lxd/instance"
+	"github.com/canonical/lxd/shared/api"
 )
 
 type devLXDDeviceAccessValidator func(device map[string]string) bool
 
+// getOwnedDevices extracts instance devices that are owned by the provided identity.
+// Two maps are returned, one containing owned devices, and the other containing the remaining
+// devices.
```
tomponline: no need to wrap here.

lxd/devlxd_instances.go
tomponline, line 1: Seeing the amount of devlxd specific functionality going into the main package I am starting to think its time to split devlxd specific functionality into its own `devlxd` package.

tomponline:
I'm thinking for future readers' benefit it would be helpful to expand the commit message for each of the API endpoints you are adding explaining why they are being added and what purpose they serve for the CSI. Rather than just a one line naming the endpoint as we have now. Thanks

lxd/devlxd_instances.go
```diff
@@ -1,13 +1,95 @@
...
+//   - Action:    Adds new device to the instance and set its owner in the instance config.
+//
+// - Update:
+//   - Condition: Existing device that devLXD can manage and is present in the new devices.
+//   - Action:    Updates existing device in the instance.
+//
+// - Remove:
+//   - Condition: Existing device that devLXD can manage is set to "null" in the request.
+//   - Action:    Removes existing device from the instance and removes its owner from the instance config.
+func patchInstanceDevices(inst *api.Instance, req api.DevLXDInstancePut, identityID string, isDeviceAccessible devLXDDeviceAccessValidator) error {
```
tomponline: This function seems complicated enough that it warrants some unit tests, what do you think?

lxd/devlxd_instances.go
```diff
@@ -4,13 +4,74 @@ import (
...
 )
 
 type devLXDDeviceAccessValidator func(device map[string]string) bool
 
+var devLXDInstanceEndpoint = devLXDAPIEndpoint{
+	Path: "instances/{name}",
+	Get:  devLXDAPIEndpointAction{Handler: devLXDInstanceGetHandler, AccessHandler: allowDevLXDPermission(entity.TypeInstance, auth.EntitlementCanView, "name")},
+}
+
+func devLXDInstanceGetHandler(d *Daemon, r *http.Request) response.Response {
```
tomponline, line 31: missing doc block

lxd/devlxd_instances.go
```diff
@@ -4,13 +4,74 @@ import (
...
+	}
+
+	resp := instanceGet(d, req)
+	etag, err := RenderToStruct(req, resp, &targetInst)
+	if err != nil {
+		return response.DevLXDErrorResponse(err)
+	}
+
+	// Get owned devices.
+	ownedDevices, _ := getOwnedDevices(targetInst, identity.Identifier)
```
tomponline: does getOwnedDevices need to return other devices at all?

lxd/devlxd_instances.go
```diff
@@ -4,13 +4,74 @@ import (
...
+
+	// Filter devices that are not accessible to devLXD.
+	deviceAccessChecker := newDevLXDDeviceAccessValidator(inst)
+	for name, device := range ownedDevices {
+		if !deviceAccessChecker(device) {
+			delete(ownedDevices, name)
+		}
+	}
+
+	// Map to devLXD type.
```
tomponline: Suggest "Convert" to avoid confusion with built-in type `map`

lxd/devlxd_instances.go
```diff
@@ -72,6 +74,79 @@ func devLXDInstanceGetHandler(d *Daemon, r *http.Request) response.Response {
 	return response.DevLXDResponseETag(http.StatusOK, respInst, "json", etag)
 }
 
+func devLXDInstancePatchHandler(d *Daemon, r *http.Request) response.Response {
```
tomponline, line 91: missing doc block

shared/api/devlxd_storage.go
tomponline: Yes if there's additional PRs coming lets do that - they can all some under the single API extension though

client/devlxd_operations.go
```diff
@@ -1,13 +1,89 @@
...
+var _ Operation = &devLXDOperation{}
+
+// The Operation type represents an ongoing LXD operation (asynchronous processing).
+type devLXDOperation struct {
+	api.DevLXDOperation
+
+	r *ProtocolDevLXD
+}
+
+// Cancel is a no-op.
```
tomponline: Why?

lxd/devlxd_storage.go
```diff
@@ -349,7 +349,7 @@ func devLXDStoragePoolVolumePutHandler(d *Daemon, r *http.Request) response.Resp
 
 	err = addStoragePoolVolumeDetailsToRequestContext(d.State(), req)
 	if err != nil {
-		response.DevLXDErrorResponse(err)
+		return response.DevLXDErrorResponse(err)
```
tomponline: this should be added to the original commit that added `devLXDStoragePoolVolumePutHandler`

lxd-agent/response.go
tomponline, line 1: please expand commit message to explain what is being fixed.

lxd-agent/response.go
```diff
@@ -12,19 +13,38 @@ type devLXDResponse struct {
...
+	// Handle error responses.
+	if r.code != http.StatusOK {
 		http.Error(w, fmt.Sprint(r.content), r.code)
-	} else if r.ctype == "json" {
+		return nil
+	}
+
+	// Set ETag if provided.
+	if r.etag != "" {
+		// Remove quotes from ETag if present.
```
tomponline: Still not clear on this quoting business. Seems like we should be breaking this out into separate function(s) to de-dupe this quoting logic as ive seen it in a few places now.

lxd-agent/response.go
```diff
@@ -65,6 +85,15 @@ func okResponse(ct any, ctype string) *devLXDResponse {
 	}
 }
 
+func okResponseETag(ct any, ctype string, etag string) *devLXDResponse {
```
tomponline: This got me thinking, why does the main LXD not need an `okResponseEtag`, but devlxd does?

lxd/instance/drivers/driver_qemu.go
```diff
@@ -5771,6 +5771,7 @@ func (d *qemu) Update(args db.InstanceArgs, userRequested bool) error {
 			"security.csm",
 			"security.devlxd",
 			"security.devlxd.images",
+			"security.devlxd.management.volumes",
```
tomponline: do we need the same thing for containers?

tomponline (changes requested):
Thanks @MusicDin looking good so far.

I've been trying to think of a way we can split this up into smaller PRs as its rather unwieldy as it stands.

How about we break this PR up into the following 2 PRs:

1. LXD side devlxd APIs, devlxd client and associated container tests.
2. lxd-agent side devlxd APIs to add VM support.

They can both be under the same API extension though.

The only big area of concern I saw was the storage pool Config field.
I did not see any mechanism to prevent leaking storage appliance credentials to devlxd users.

lxd/devlxd_instances.go
MusicDin: No. Previously we used PUT and we had to retain non-owned devices in the request. Now that we use patch, this is no longer needed.

lxd/response/response.go
MusicDin: Because LXD quotes ETag in sync response:
https://github.com/canonical/lxd/blob/d0301013b227c5915dbe2330431e0795d4ad7507/lxd/response/response.go#L176-L184

lxd/response/response.go
MusicDin: To make it consistent with LXD itself, the ETag is always quoted. In case of devLXD it ensures it is first unquoted to prevent double quoting (same for devLXD over vsock).

lxd/devlxd.go
MusicDin: We can make it generic (`getIdentity()`?). The main intention of the function is just to avoid duplication.

lxd/devlxd_instances.go
MusicDin: Done.

lxd-agent/response.go
MusicDin: LXD has `SyncResponseEtag()`

MusicDin:
> I've been trying to think of a way we can split this up into smaller PRs as its rather unwieldy as it stands.
> How about we break this PR up into the following 2 PRs:
> - LXD side devlxd APIs, devlxd client and associated container tests.
> - lxd-agent side devlxd APIs to add VM support.

I've separated the PR as per your suggestion. Also, removed everything snapshots and volume cloning related, which was still in the PR (for example: volume `Restore`/`Source` fields) as this will also land in separate PRs.

doc/.wordlist.txt
```diff
@@ -7,6 +7,7 @@ APIs
 balancer
 Charmhub
 CLI
+devLXD
```
markylaing: Shouldn't need this. I think we should be using DevLXD everywhere (or `devlxd` if necessary).

doc/metadata.txt
```diff
@@ -5147,6 +5162,14 @@ This number is then incremented by one for the new name.
 Specify either a cron expression (`<minute> <hour> <dom> <month> <dow>`), a comma-separated list of schedule aliases (`@hourly`, `@daily`, `@midnight`, `@weekly`, `@monthly`, `@annually`, `@yearly`), or leave empty to disable automatic snapshots (the default).
 ```
 
+```{config:option} volatile.devlxd.owner storage-btrfs-volume-conf
+:defaultdesc: "devLXD owner identity ID"
```
markylaing:
```suggestion
:defaultdesc: "DevLXD owner identity ID"
```

lxd/api_vsock.go
```diff
@@ -37,6 +37,14 @@ func hoistReqVM(d *Daemon, r *http.Request, handler devLXDAPIHandlerFunc) respon
...
 
 	request.SetContextValue(r, request.CtxDevLXDInstance, inst)
+
+	if accessHandler != nil {
+		resp := accessHandler(d, r)
+		if resp != response.EmptySyncResponse {
+			return resp
+		}
+	}
+
```
markylaing: I'm a little confused as to why the access handler needs to go in the hoist functions.

The hoist functions were used to determine the `instance.Instance` before passing it into the handlers, which at the time, took an `instance.Instance` as an argument. 

Since we now store the instance in the request context, we no longer need to have the hoist functions work this way. 

I think we should have a separate PR which makes this change and introduces access handlers to the devlxd API.

lxd/devlxd.go
markylaing: `request.GetCallerIdentityFromContext(ctx context.Context)` could be useful. 

In the main API, in most cases, we need to call `GetRequestor` first because an identity is not always present (e.g. for unix requests). That said, I can see why this would be useful if you know an identity must be present in the context due to it being validated in e.g. an access handler

lxd/api_vsock.go
MusicDin: The reason why hoist func needs to be called first, is to retrieve the instance and its project to prevent cross-project actions.

We could move the call to the handler from hoist func. Then this would allow us to first call hoist func to retrieve the instance, and later call the remaining handlers.

markylaing: > We could move the call to the handler from hoist func. Then this would allow us to first call hoist func to retrieve the instance, and later call the remaining handlers.

Yeah that's what I'm suggesting. I think these can now be functions that return the instance or an error, and are called at the beginning of the main handler. The main handler can then set the instance in the context for the access and main handler to use.

MusicDin: Also, I would allow multiple access handlers to be added. This way we don't need wrap them in another function when, for example, a security flag and a token have to be checked.

MusicDin: Discussion moved to https://github.com/canonical/lxd/pull/16329

MusicDin:
Based on @markylaing comments, also fixed:
- Rename constant `devLXDSecurityMgmtVolumesKey` -> `devLXDSecurityManagementVolumesKey`
- Remove unqoting of ETags, they are set in the header as is. The ETags from LXD are always quoted anyway.
- Added several comments.

lxd/response/response.go
MusicDin: Removed unqoting. ETag is not set in the header as provided because LXD ETags are always quoted anyway.

lxd-agent/response.go
MusicDin: This is now removed - ETag is set in the header as provided. LXD ETags are already quoted.

lxd/devlxd_response_render.go
```diff
@@ -0,0 +1,23 @@
+package main
+
+import (
+	"net/http"
+
+	"github.com/canonical/lxd/lxd/response"
+)
+
+// Render renders the response and returns a potential error.
+func Render(req *http.Request, resp response.Response) error {
```
tomponline: This doesnt seem to be devlxd specific, yet its in a devlxd file?

lxd/devlxd_response_render.go
```diff
@@ -21,3 +21,24 @@ func Render(req *http.Request, resp response.Response) error {
 
 	return nil
 }
+
+// RenderToStruct renders the response into a struct and returns the ETag.
+func RenderToStruct(req *http.Request, resp response.Response, target any) (etag string, err error) {
```
tomponline: not devlxd specific, but in devlxd file.

lxd/devlxd_response_render.go
MusicDin: Should I move that to `lxd/response`, where `responseCapture` is located?

Actually, since those functions create new response capture, they can simply be receiver functions of `responseCapture`, which would simplify the invocation and reduce the amount of globally accessible functions in `lxd/response` package

tomponline: Sounds good

lxd/request/request.go
```diff
@@ -1,14 +1,74 @@
...
+	return NewRequestWithContext(context.Background(), method, url, data, ETag)
+}
+
+// NewRequestWithContext creates a new HTTP request with the specified method, URL, data, and ETag.
+func NewRequestWithContext(ctx context.Context, method string, url string, data any, ETag string) (*http.Request, error) {
+	var req *http.Request
+	var err error
+
+	// Create new HTTP request.
+	if data != nil {
```
tomponline: This section looks very similar to that of client.RawQuery.

Can we de-dupe it somehow? Perhaps by moving this to the client package and using it in LXD?

shared/api/devlxd.go
```diff
@@ -41,6 +41,21 @@ type DevLXDGet struct {
 	// Whether the caller is trusted or untrusted
 	// Example: untrusted
 	Auth string `json:"auth" yaml:"auth"`
+
+	// List of supported storage drivers.
+	// Example: [{"name": "zfs", "remote": false}, {"name": "btrfs", "remote": true}]
```
tomponline: Should include the API extension that added this field in the comment

MusicDin: Any naming suggestion? How about `devlxd_volume_management` to cover new API as well?

tomponline: yep sounds good -  did you not add an api extension yet for this large api extension ;)

lxd/devlxd_storage.go
```diff
@@ -296,6 +297,51 @@ func devLXDStoragePoolVolumeGetHandler(d *Daemon, r *http.Request) response.Resp
...
 
+// devLXDStoragePoolVolumeDeleteHandler deletes the specified custom storage volume if it is owned by the caller.
+// If the volume is not found or not owned by the caller, it returns a generic not found error.
+func devLXDStoragePoolVolumeDeleteHandler(d *Daemon, r *http.Request) response.Response {
+	inst, err := getInstanceFromContextAndCheckSecurityFlags(r.Context(), devLXDSecurityKey, devLXDSecurityManagementVolumesKey)
+	if err != nil {
+		return response.DevLXDErrorResponse(err)
+	}
+
+	poolName := mux.Vars(r)["poolName"]
```
tomponline: Small nit, but we should try to avoid calling `mux.Vars()` repeatedly (in all API handlers).

lxd/devlxd_storage.go
```diff
@@ -296,6 +297,51 @@ func devLXDStoragePoolVolumeGetHandler(d *Daemon, r *http.Request) response.Resp
...
+	if target != "" {
+		url = url.WithQuery("target", target)
+	}
+
+	req, err := request.NewRequestWithContext(r.Context(), http.MethodDelete, url.String(), nil, "")
+	if err != nil {
+		return response.DevLXDErrorResponse(err)
+	}
+
+	err = addStoragePoolVolumeDetailsToRequestContext(d.State(), req)
```
tomponline: @markylaing @MusicDin why isn't this being done by an API route level access handler like in `storagePoolVolumeTypeCmd`?

lxd/devlxd.go
```diff
@@ -687,23 +687,32 @@ func allowDevLXDPermission(entityType entity.Type, entitlement auth.Entitlement,
...
+	}
+
+	return inst, nil
+}
+
+// hasInstanceSecurityFeatures checks whether the instance has the provided devLXD security features enabled.
+func hasInstanceSecurityFeatures(inst instance.Instance, keys ...DevLXDSecurityKey) bool {
 	for _, key := range keys {
-		value := config[string(key)]
+		value := inst.ExpandedConfig()[string(key)]
```
tomponline: can this function be passed ExpandedConfig as an argument, we shouldn't be calling ExpandedConfig in a loop like we are here.

lxd/devlxd_instances.go
```diff
@@ -3,13 +3,78 @@ package main
...
+	// Get identity from the request context.
+	identity, err := request.GetCallerIdentityFromContext(r.Context())
+	if identity == nil {
+		return response.DevLXDErrorResponse(err)
+	}
+
+	// Fetch instance.
+	targetInst := api.Instance{}
+
+	url := api.NewURL().Path("1.0", "instances", targetInstName).Project(projectName).WithQuery("recursion", "1").URL
```
tomponline: No need for `.URL` on the end, you can just use `api.NewURL()` returned and call `.String()`  on it when needed.

client/devlxd.go
```diff
@@ -39,6 +39,10 @@ type ProtocolDevLXD struct {
 	isDevLXDOverVsock bool
 
 	bearerToken string
+
+	// clusterTarget is the name of the cluster member
+	// that this client is targeting.
```
tomponline: no need to overflow the line here

lxd/devlxd_storage.go
```diff
@@ -337,7 +337,8 @@ func devLXDStoragePoolVolumePutHandler(d *Daemon, r *http.Request) response.Resp
 	}
 
 	// Ensure the volume owner cannot be changed.
-	if !isDevLXDVolumeOwner(vol.Config, identity.Identifier) {
+	owner, ok := vol.Config["volatile.devlxd.owner"]
```
tomponline: why not using helper function here?

tomponline: this looks like a bad rebase

MusicDin: Yes, it is bad rebase. But the second approach is correct:
```go
owner, ok := vol.Config["volatile.devlxd.owner"]
if ok && owner != identity.Identifier {}
```

Because we want to check the owner only if the owner is set in the request (e.g. when doing get and update). 
If the owner is not set, we simply retain the owner.

We could alternatively do:
```go
_, ok := vol.Config["volatile.devlxd.owner"]
if ok && !isDevLXDVolumeOwner(...) {}
```

lxd/request/request.go
MusicDin: Fixed. Also de-duped from DevLXD and LXD clients.

shared/api/devlxd.go
MusicDin: Done.

client/devlxd.go
MusicDin: Fixed

lxd/devlxd_storage.go
MusicDin: Fixed.

tomponline: I prefer the approach you proposed that uses isDevLXDVolumeOwner() so we dont duplicate logic.

tomponline: And we can see all use-case of checking device ownership.

lxd/devlxd.go
MusicDin: Fixed.

lxd/devlxd_storage.go
MusicDin: Thanks. Fixed, and also ensured values are unescaped.

MusicDin:
The only open question remaining is:

> @markylaing @MusicDin why isn't this being done by an API route level access handler like in storagePoolVolumeTypeCmd?

This is comment on adding volume details to the request in `GET/PUT/DELETE storage-pools/{poolName}/volumes/{type}/{volumeName}` DevLXD handlers:
```go
err = addStoragePoolVolumeDetailsToRequestContext(d.State(), req)
if err != nil {
        return response.DevLXDErrorResponse(err)
}
```

Based on our previous discussion, the storage details are not added during in the main API handlers because the target needs to be resolved beforehand. However, we do not enforce that the target is set in DevLXD either, so this means that target is effectively not resolved prior calling any of those endpoints. 

However, I would expect the (main LXD) handler to error out if multiple storage volumes are found, but the target is not provided. Effectively enforcing target requirement if needed.

I'm open to clarifications and corrections :)

lxd/devlxd_storage.go
tomponline: What i'd like to understand is why cant be like the main API handler:

```go
var storagePoolVolumeTypeCmd = APIEndpoint{
	Path:        "storage-pools/{poolName}/volumes/{type}/{volumeName}",
	MetricsType: entity.TypeStoragePool,

	Delete: APIEndpointAction{Handler: storagePoolVolumeDelete, AccessHandler: storagePoolVolumeTypeAccessHandler(entity.TypeStorageVolume, auth.EntitlementCanDelete)},
	Get:    APIEndpointAction{Handler: storagePoolVolumeGet, AccessHandler: storagePoolVolumeTypeAccessHandler(entity.TypeStorageVolume, auth.EntitlementCanView)},
	Patch:  APIEndpointAction{Handler: storagePoolVolumePatch, AccessHandler: storagePoolVolumeTypeAccessHandler(entity.TypeStorageVolume, auth.EntitlementCanEdit)},
	Post:   APIEndpointAction{Handler: storagePoolVolumePost, AccessHandler: storagePoolVolumeTypeAccessHandler(entity.TypeStorageVolume, auth.EntitlementCanEdit)},
	Put:    APIEndpointAction{Handler: storagePoolVolumePut, AccessHandler: storagePoolVolumeTypeAccessHandler(entity.TypeStorageVolume, auth.EntitlementCanEdit)},
}
```

tomponline:
> The only open question remaining is:
> 
> > @markylaing @MusicDin why isn't this being done by an API route level access handler like in storagePoolVolumeTypeCmd?
> 
> This is comment on adding volume details to the request in `GET/PUT/DELETE storage-pools/{poolName}/volumes/{type}/{volumeName}` DevLXD handlers:
> 
> ```go
> err = addStoragePoolVolumeDetailsToRequestContext(d.State(), req)
> if err != nil {
>         return response.DevLXDErrorResponse(err)
> }
> ```
> 
> Based on our previous discussion, the storage details are not added during in the main API handlers because the target needs to be resolved beforehand. However, we do not enforce that the target is set in DevLXD either, so this means that target is effectively not resolved prior calling any of those endpoints.
> 
> However, I would expect the (main LXD) handler to error out if multiple storage volumes are found, but the target is not provided. Effectively enforcing target requirement if needed.
> 
> I'm open to clarifications and corrections :)

Lets discuss at team meeting today as i still dont see why devlxd needs to be different to main API handler here.

MusicDin:
@tomponline If the above issue is the only one remaining, could we get that merged, and I can address the issue after our discussions?

This would unblock me for the CSI itself, and all subsequent PRs will be much smaller and easier to review.

tomponline:
> @tomponline If the above issue is the only one remaining, could we get that merged, and I can address the issue after our discussions?
> 
> This would unblock me for the CSI itself, and all subsequent PRs will be much smaller and easier to review.

I wont block on this issue, but its pretty important.

Also i didnt manage to make it through the whole PR yet - in the future if we can break up large PRs like this into smaller deliverables it would help to get them in quicker as wont need such a long review each pass.

shared/api/devlxd.go
tomponline: still missing API extension

MusicDin: Ahh, I see. I'll add it rn

lxd/devlxd_storage.go
tomponline: this point as not been addressed yet

shared/api/devlxd.go
MusicDin: Now it is really done :)

doc/api-extensions.md
```diff
@@ -2711,5 +2711,22 @@ See {ref}`DevLXD bearer tokens <devlxd-authentication-bearer>`.
...
+
+New DevLXD endpoints:
+- `GET /1.0/storage-pools/{pool}` — Retrieve a storage pool.
+- `GET /1.0/storage-pools/{pool}/volumes` — List owned custom volumes in a pool.
+- `GET /1.0/storage-pools/{pool}/volumes/{volume}` — Retrieve an owned custom volume.
+- `POST /1.0/storage-pools/{pool}/volumes` — Create a new owned custom volume in a pool.
+- `PUT /1.0/storage-pools/{pool}/volumes/{volume}` — Update an owned custom volume.
+- `DELETE /1.0/storage-pools/{pool}/volumes/{volume}` — Delete an owned custom volume.
+- `GET /1.0/instances/{inst}` — Retrieve an instance and its owned devices.
+- `PATCH /1.0/instances/{inst}` — Add new instance devices or modify owned devices.
```
simondeziel: Could you please run `make doc-lint` as I think those `- ` prefixes seem to upset it.

Maybe using `* ` would fare better?

MusicDin: Thanks. That was the issue + missing line before list elements.

lxd/devlxd.go
```diff
@@ -616,6 +616,32 @@ func registerDevLXDEndpoint(d *Daemon, apiRouter *mux.Router, apiVersion string,
...
+
+	if projectParam == "" {
+		// Ensure the project query parameter is always set.
+		// This is needed by the permission checkers to determine the correct project.
+		q := r.URL.Query()
+		q.Set("project", instProject)
+		r.URL.RawQuery = q.Encode()
+	} else if projectParam != instProject {
+		// Disallow cross-project access.
+		return api.StatusErrorf(http.StatusForbidden, "Cross-project access is not allowed")
```
tomponline: Lets just use a generic error here as "Cross-project" access allows enumeration of projects.

MusicDin: Fixed.

Used `return api.NewGenericStatusError(http.StatusForbidden)` instead.

lxd/devlxd_storage.go
MusicDin: Done.

lxd/devlxd_storage.go
```diff
@@ -353,6 +354,92 @@ func devLXDStoragePoolVolumeGetHandler(d *Daemon, r *http.Request) response.Resp
...
+	inst, err := getInstanceFromContextAndCheckSecurityFlags(r.Context(), devLXDSecurityKey, devLXDSecurityManagementVolumesKey)
+	if err != nil {
+		return response.DevLXDErrorResponse(err)
+	}
+
+	projectName := inst.Project().Name
+	target := r.URL.Query().Get("target")
+	pathVars := mux.Vars(r)
+
+	poolName, err := url.PathUnescape(pathVars["poolName"])
```
tomponline, line 349: The extraction of projectName, target, pathVars, poolName, volName and volType seem to be repeated for GET, PUT and DELETE. Is it worth creating a function that does this and then calling it from these 3 functions?

lxd/devlxd_instances.go
```diff
@@ -0,0 +1,21 @@
...
+// newDevLXDDeviceAccessValidator returns a device validator function that
+// checks if the given device is accessible by the devLXD.
+//
+// For example, disk device is accessible if the appropriate security flag
+// is enabled on the instance and the device represents a custom volume.
+func newDevLXDDeviceAccessValidator(inst instance.Instance) devLXDDeviceAccessValidator {
+	diskDeviceAllowed := hasInstanceSecurityFeatures(inst.ExpandedConfig(), devLXDSecurityManagementVolumesKey)
+
+	return func(device map[string]string) bool {
+		return diskDeviceAllowed && filters.IsCustomVolumeDisk(device)
```
tomponline, line 256: Nice use of `filters` package here.

lxd/devlxd_instances_test.go
tomponline, line 1: Well done for adding tests for that non-trivial function. Thanks!

lxd/devlxd_instances.go
```diff
@@ -2,14 +2,85 @@ package main
...
+// devLXDInstanceGetHandler retrieves instance information for the specified instance name.
+// The returned instance information includes only the devices that are accessible to devLXD
+// and are owned by the devLXD identity making the request.
+func devLXDInstanceGetHandler(d *Daemon, r *http.Request) response.Response {
+	inst, err := getInstanceFromContextAndCheckSecurityFlags(r.Context(), devLXDSecurityKey, devLXDSecurityManagementVolumesKey)
+	if err != nil {
+		return response.DevLXDErrorResponse(err)
+	}
+
+	// Allow access only to the projectName where current instance is running.
```
tomponline: should we reject requests with `project=` in it for a non-matching project to the source instance?

test/suites/devlxd.sh
```diff
@@ -244,8 +246,150 @@ test_devlxd_volume_management() {
...
+    lxc auth identity group add "${authIdentity}" "${authGroup}"
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client instance get "${inst}" | jq -e .name
+
+    # Test devLXD authorization (volume management security flag).
+    # Fail when the security flag is not set.
+    lxc config set "${inst}" --project "${project}" security.devlxd.management.volumes=false
+    [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client instance get "${inst}")" = "Forbidden" ]
+    lxc config set "${inst}" --project "${project}" security.devlxd.management.volumes=true
+
+    # Get storage pool.
```
tomponline, line 274: Do we need a check here for insufficient permission too?

test/suites/devlxd.sh
```diff
@@ -244,8 +246,150 @@ test_devlxd_volume_management() {
...
+
+    # Fail - already exists.
+    [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage create-volume "${pool}" "${vol2}")" = "Volume by that name already exists" ]
+
+    # Verify created storage volumes.
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage volumes "${pool}" | jq 'length == 2'
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage get-volume "${pool}" custom vol-01
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage get-volume "${pool}" custom vol-02
+
+    # Update storage volume.
```
tomponline, line 305: Please can you add tests checking that we cant put, patch, delete non-custom volumes too?

tomponline:
@markylaing please can you let me know if you're happy with the access control measures in this PR. Thanks

lxd/devlxd_instances.go
MusicDin: It is already checked by the access handler

tomponline: ah so the comment is misleading then as its not the one doing the access control.

test/suites/devlxd.sh
MusicDin: Done.

lxd/devlxd_instances.go
tomponline: May change "allow access" to "Force ..."

tomponline: with a comment explaining the access handler checks this, but its just belt-and-braces.

MusicDin: Agree. Have updated the comment

lxd/devlxd_storage.go
```diff
@@ -0,0 +1,513 @@
...
+var devLXDStoragePoolVolumesEndpoint = APIEndpoint{
+	Path: "storage-pools/{poolName}/volumes",
+	Get:  APIEndpointAction{Handler: devLXDStoragePoolVolumesGetHandler, AccessHandler: allowDevLXDAuthenticated},
+	Post: APIEndpointAction{Handler: devLXDStoragePoolVolumesPostHandler, AccessHandler: allowDevLXDAuthenticated},
+}
+
+var devLXDStoragePoolVolumesTypeEndpoint = APIEndpoint{
+	Path: "storage-pools/{poolName}/volumes/{type}",
+	Get:  APIEndpointAction{Handler: devLXDStoragePoolVolumesGetHandler, AccessHandler: allowDevLXDAuthenticated},
+	Post: APIEndpointAction{Handler: devLXDStoragePoolVolumesPostHandler, AccessHandler: allowDevLXDAuthenticated},
```
markylaing: This access handler should be checking if the devlxd identity has permission to create volumes in the project.

lxd/devlxd_storage.go
```diff
@@ -0,0 +1,513 @@
...
+		Status: pool.Status,
+	}
+
+	return response.DevLXDResponseETag(http.StatusOK, respPool, "json", etag)
+}
+
+var devLXDStoragePoolVolumesEndpoint = APIEndpoint{
+	Path: "storage-pools/{poolName}/volumes",
+	Get:  APIEndpointAction{Handler: devLXDStoragePoolVolumesGetHandler, AccessHandler: allowDevLXDAuthenticated},
+	Post: APIEndpointAction{Handler: devLXDStoragePoolVolumesPostHandler, AccessHandler: allowDevLXDAuthenticated},
```
markylaing: This access handler should be checking if the devlxd identity has permission to create volumes in the project.

lxd/devlxd_storage.go
```diff
@@ -0,0 +1,513 @@
...
+	resp := storagePoolVolumesPost(d, req)
+	err = response.NewResponseCapture(req).Render(resp)
+	if err != nil {
+		return response.DevLXDErrorResponse(err)
+	}
+
+	return response.DevLXDResponse(http.StatusOK, "", "raw")
+}
+
+var devLXDStoragePoolVolumeTypeEndpoint = APIEndpoint{
```
markylaing: Can you please move all of these `APIEndpoint` definitions to the top of the file.

markylaing (changes requested):
Most handlers look fine from an auth perspective except for creation of storage volumes.
Even with the storage volume management security key, we should not allow creation of storage volumes unless the identity has `can_create_storage_volumes` on the project.

lxd/devlxd_storage.go
MusicDin: Thanks, will fix

lxd/devlxd_storage.go
```diff
@@ -0,0 +1,513 @@
...
+
+var devLXDStoragePoolEndpoint = APIEndpoint{
+	Path: "storage-pools/{poolName}",
+	Get:  APIEndpointAction{Handler: devLXDStoragePoolGetHandler, AccessHandler: allowDevLXDAuthenticated},
+}
+
+var devLXDStoragePoolVolumesEndpoint = APIEndpoint{
+	Path: "storage-pools/{poolName}/volumes",
+	Get:  APIEndpointAction{Handler: devLXDStoragePoolVolumesGetHandler, AccessHandler: allowDevLXDAuthenticated},
+	Post: APIEndpointAction{Handler: devLXDStoragePoolVolumesPostHandler, AccessHandler: allowDevLXDPermission(entity.TypeProject, auth.EntitlementCanCreateStorageVolumes)},
```
markylaing, line 29: Nice, thanks

PR #16280, opened 2025-08-22
Use lxd-mount when mounting to a target that already exists

kadinsayani:
Closes https://github.com/canonical/lxd/issues/16264.


lxd/instance/drivers/driver_lxc.go
```diff
@@ -7768,7 +7768,21 @@ func (d *lxc) insertMount(source, target, fstype string, flags int, idmapType id
...
+		// replace a placeholder at the mount target as part of its attach logic.
+		// When the target already exists, this can implicitly change or overwrite
+		// the path's mode/uid/gid or other metadata.
+		pid := d.InitPID()
+		if pid > 0 {
+			if !strings.HasPrefix(target, "/") {
+				target = "/" + target
+			}
+
+			if !shared.PathExists("/proc/" + strconv.Itoa(pid) + "/root" + target) {
```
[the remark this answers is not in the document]
simondeziel: I think it's fair to assume we'll only ever be dealing with Unix paths when it comes to the lxc driver reading from `/proc`. `filepath.Join()` is thus not strictly needed IMHO.

lxd/instance/drivers/driver_lxc.go
```diff
@@ -7768,7 +7768,21 @@ func (d *lxc) insertMount(source, target, fstype string, flags int, idmapType id
...
+		// Only use liblxc's mount injection when the target path does not already
+		// exist inside the container. liblxc's mount injection may create or
+		// replace a placeholder at the mount target as part of its attach logic.
+		// When the target already exists, this can implicitly change or overwrite
+		// the path's mode/uid/gid or other metadata.
+		pid := d.InitPID()
+		if pid > 0 {
+			if !strings.HasPrefix(target, "/") {
+				target = "/" + target
+			}
```
simondeziel, line 7780: This bit comes for free if you use `filepath.Join()` I think.

test/suites/container_devices_disk.sh
```diff
@@ -94,7 +95,36 @@ _container_devices_disk_shift() {
...
+  lxc exec foo -- mkdir -p /opt/target
+  lxc exec foo -- chmod 754 /opt/target
+  lxc exec foo -- chown 12345:12345 /opt/target
+  lxc config device add foo bar disk source="$(mktemp -d)" path=/opt/target
+  lxc config device remove foo bar
+  # Check permissions and ownership remain after removal.
+  [ "$(lxc exec foo -- stat -c '%a %u %g' /opt/target)" = "754 12345 12345" ]
+
+  # Add a mount point that points to an existing file in the instance.
+  lxc exec foo -- sh -c 'echo "hello" > /opt/target-file'
```
simondeziel, line 113:
```suggestion
  echo "hello" | lxc exec foo -- tee /opt/target-file
```

Or use `lxc file push -` maybe?

test/suites/container_devices_disk.sh
```diff
@@ -94,7 +95,36 @@ _container_devices_disk_shift() {
...
   lxc storage volume delete "${POOL}" foo-shift
-  lxc stop foo -f
+}
+
+_container_devices_disk_mount() {
+  # Add a mount that points to an existing path in the instance.
+  lxc exec foo -- mkdir -p /opt/target
+  lxc exec foo -- chmod 754 /opt/target
+  lxc exec foo -- chown 12345:12345 /opt/target
+  lxc config device add foo bar disk source="$(mktemp -d)" path=/opt/target
```
simondeziel: This leaves a tmp dir behind. This should put it in the `TEST_DIR` that is cleanup at the end:

```suggestion
  lxc config device add foo bar disk source="$(mktemp -d -p "${TEST_DIR}" XXX)" path=/opt/target
```

test/suites/container_devices_disk.sh
```diff
@@ -94,7 +95,36 @@ _container_devices_disk_shift() {
...
+  lxc config device remove foo bar
+  # Check permissions and ownership remain after removal.
+  [ "$(lxc exec foo -- stat -c '%a %u %g' /opt/target)" = "754 12345 12345" ]
+
+  # Add a mount point that points to an existing file in the instance.
+  lxc exec foo -- sh -c 'echo "hello" > /opt/target-file'
+  lxc exec foo -- chmod 754 /opt/target-file
+  lxc exec foo -- chown 12345:12345 /opt/target-file
+  lxc config device add foo bar disk source="$(mktemp)" path=/opt/target-file
```
simondeziel:
```suggestion
  lxc config device add foo bar disk source="$(mktemp -p "${TEST_DIR}" XXX)" path=/opt/target-file
```

test/suites/container_devices_disk.sh
kadinsayani: `lxc file push` works well, thanks :)

test/suites/container_devices_disk.sh
kadinsayani: Nice catch, thanks.

lxd/instance/drivers/driver_lxc.go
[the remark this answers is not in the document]
kadinsayani: I've gone with `fmt.Sprintf()` for readability and consistency in `driver_lxc.go`.

tomponline:
Would appreciate @mihalicyn thoughts on this one. Thanks

lxd/instance/drivers/driver_lxc.go
```diff
@@ -7768,7 +7768,21 @@ func (d *lxc) insertMount(source, target, fstype string, flags int, idmapType id
...
+
+			if !shared.PathExists(fmt.Sprintf("/proc/%d/root%s", pid, target)) {
```
tomponline, line 7782: @mihalicyn is this safe to do on an untrusted container rootfs?

test/suites/container_devices_disk.sh
```diff
@@ -97,6 +98,32 @@ _container_devices_disk_shift() {
   lxc stop foo -f
 }
 
+_container_devices_disk_mount() {
+  lxc start foo
```
tomponline, line 102: Does this only affecting hotplugging files or should we also be testing for boot-time file passthrough permissions too?

kadinsayani: The issue is only observed when hot-plugging disk devices.

tomponline: Do you know why?

kadinsayani: Here is the relevant code in `lxc`:

https://github.com/lxc/lxc/blob/main/src/lxc/lxccontainer.c#L4910-L5088

After mounting, the underlying target’s original permissions and ownership are not preserved; they are removed and replaced by a fresh placeholder with 0000 perms, which is then covered by the mount.

`create_mount_target()` unconditionally calls `(void)remove(dest)`, deleting the target directory before the new mount target is created.

kadinsayani: On the LXD side, the `idmapType` param determines if `insertMount()` calls `moveMount()`, `insertMountLXC()`, or `insertMountLXD()`. `idmapType` is determined here: 

https://github.com/canonical/lxd/blob/3975f51d0f2aed28d8e1524140b9cb4d74500e3c/lxd/instance/drivers/driver_lxc.go#L1661-L1673

When the disk device config `shift` is set to `false`, we end up calling `insertMountLXC()`, since `idmapType = IdmapStorageNone`.

`insertMountLXC()` calls `forkmount lxc-mount`, which uses the `liblxc` mount api.

Here is the relevant code in lxc:

https://github.com/lxc/lxc/blob/main/src/lxc/lxccontainer.c#L4910-L5088

After mounting, the underlying target’s original permissions and ownership are not preserved; they are removed and replaced by a fresh placeholder with 0000 perms, which is then covered by the mount. This is because `create_mount_target()` unconditionally calls `(void)remove(dest)`, deleting the target directory before the new mount target is created.

When the instance isn't running, rather than calling `insertMount()`, we set `lxc.mount.entry`. I followed the code path for setting boot-time mounts in `liblxc` which led me to this function:

https://github.com/lxc/lxc/blob/95cad77e7d36484e49fa5c1d5a7f3e5fb3d64abc/src/lxc/utils.c#L219-L245

When `create=dir/file` is set, the directory/file won't be created if it already exists, so the original target is not overridden before mounting.

kadinsayani: Perhaps we should be aligning the behavior of  `do_lxcapi_mount()` and `lxc.mount.entry` when `create=dir/file` is passed as an arg, rather than going with the fix in this PR which just skips `insertMountLXC()` and calls `insertMountLXD()`.

cc. @mihalicyn

tomponline: Yes this seems like an inconsistency in liblxc

tomponline: @kadinsayani @mihalicyn please let me know how you get on with deciding the best course of action with this one?

Thanks!

kadinsayani: @tomponline I've shared my findings with @mihalicyn and we've confirmed this is a liblxc issue. I'll submit a fix for the issue and if its accepted we can avoid working around the problem as this pull request aims to do.

kadinsayani: https://github.com/lxc/lxc/pull/4581

tomponline:
Closing for now in preference to https://github.com/lxc/lxc/pull/4581

PR #16283, opened 2025-08-24
Improve TiCS

simondeziel:


client/lxd.go
```diff
@@ -438,8 +449,8 @@ func (r *ProtocolLXD) rawWebsocket(url string) (*websocket.Conn, error) {
 	}
 
 	// Set TCP timeout options.
-	remoteTCP, _ := tcp.ExtractConn(conn.UnderlyingConn())
-	if remoteTCP != nil {
+	remoteTCP, err := tcp.ExtractConn(conn.UnderlyingConn())
+	if err == nil && remoteTCP != nil {
```
kadinsayani:
```suggestion
	// Set TCP timeout options.
	remoteTCP, err := tcp.ExtractConn(conn.NetConn())
	if err != nil {
		logger.Warn("Failed extracting TCP connection", logger.Ctx{"err": err})
	}

	err = tcp.SetTimeouts(remoteTCP, 0)
	if err != nil {
		logger.Warn("Failed setting TCP timeouts on remote connection", logger.Ctx{"err": err})
	}

```

I don't think there is any code path that returns `nil, nil` so we  can just handle any returned error and then use `remoteTCP`.

We can also switch to `conn.NetConn()` since `conn.UnderlyingConn()` is deprecated.

lxd/instance/drivers/driver_lxc.go
```diff
@@ -8507,6 +8507,9 @@ func (d *lxc) getFSStats() (*metrics.MetricSet, error) {
 
 	for scanner.Scan() {
 		fields := strings.Fields(scanner.Text())
+		if len(fields) < 3 {
+			continue
+		}
```
kadinsayani, line 8512: Can we avoid using a magic number here? Or, add a comment explaining why we expect `len(fields)` to be at least 3.

lxd/metrics/metrics.go
```diff
@@ -161,13 +160,8 @@ func MetricSetFromAPI(metrics *Metrics, labels map[string]string) (*MetricSet, e
...
 			labels := map[string]string{"mode": mode}
-			cpu := ""
-
-			if dev != "cpu" {
-				_, _ = fmt.Sscanf(dev, "cpu%s", &cpu)
-			}
-
-			if cpu != "" {
+			cpu, found := strings.CutPrefix(dev, "cpu")
+			if found {
```
kadinsayani, line 164: Nice! Much cleaner.

lxd/response/response.go
```diff
@@ -454,8 +454,8 @@ func (r *fileResponse) Render(w http.ResponseWriter, req *http.Request) error {
 	// For a single file, return it inline
 	if len(r.files) == 1 {
 		remoteConn := ucred.GetConnFromContext(req.Context())
-		remoteTCP, _ := tcp.ExtractConn(remoteConn)
-		if remoteTCP != nil {
+		remoteTCP, err := tcp.ExtractConn(remoteConn)
```
kadinsayani, line 457: Same comment as https://github.com/canonical/lxd/pull/16283/files#r2299000540.

lxd/migration_connection.go
```diff
@@ -98,8 +98,8 @@ func (c *migrationConn) AcceptIncoming(r *http.Request, w http.ResponseWriter) e
 	}
 
 	// Set TCP timeout options.
-	remoteTCP, _ := tcp.ExtractConn(c.conn.UnderlyingConn())
-	if remoteTCP != nil {
+	remoteTCP, err := tcp.ExtractConn(c.conn.UnderlyingConn())
+	if err == nil && remoteTCP != nil {
```
kadinsayani: Same comment as https://github.com/canonical/lxd/pull/16283/files#r2299000540.

shared/netutils/network_linux_cgo.go
```diff
@@ -166,19 +166,9 @@ func NetnsGetifaddrs(initPID int32, hostInterfaces []net.Interface) (map[string]
...
-			if goAddrString == "::1" {
-				scope = "local"
-			}
-
-			if strings.HasPrefix(goAddrString, "169.254") {
-				scope = "link"
-			}
-
-			if strings.HasPrefix(goAddrString, "fe80:") {
+			} else if strings.HasPrefix(goAddrString, "169.254.") || strings.HasPrefix(goAddrString, "fe80:") {
```
kadinsayani: Can we extract a function here? That way we can also use it in `lxd/resources/network.go`. Perhaps something like:

```Go
func Scope(goAddrString string) string {
	if strings.HasPrefix(goAddrString, "127.") || goAddrString == "::1" {
		return "local"
	} else if strings.HasPrefix(goAddrString, "169.254.") || strings.HasPrefix(goAddrString, "fe80:") {
		return "link"
	}

	return "global"
}
```

kadinsayani (changes requested):

lxd/instance/drivers/driver_lxc.go
simondeziel: The line right after accesses `fields[2]` so the surrounding context dispels some of the magic, IMHO :)

Maybe you'd find this more readable?

```
	for scanner.Scan() {
		fields := strings.Fields(scanner.Text())
		if len(fields) >= 2 {
			mountMap[fields[0]] = mountInfo{Mountpoint: fields[1], FSType: fields[2]}
		}
	}
```

I'm ambivalent personally.

shared/netutils/network_linux_cgo.go
simondeziel: Would `shared/network_ip.go` be a good home for this one?

client/lxd.go
simondeziel: The initial code ignoring the returned error was fine AFAICT but TiCS didn't like it which is why I added the `err == nil` bit. 

With your proposed change, on error, it would warn and then try to set a timeout on `remoteTCP == nil`, isn't it?

Re the deprecated `conn.UnderlyingConn()`, thanks I don't know why I didn't notice :)

kadinsayani: Ah yes, we shouldn't set a timeout when `remoteTCP == nil`. I think it's still useful to include a warning when we receive an error from `tcp.ExtractConn`. Something like this should work:

```suggestion
	// Set TCP timeout options.
	remoteTCP, err := tcp.ExtractConn(conn.NetConn())
	if err != nil {
		logger.Warn("Failed extracting TCP connection", logger.Ctx{"err": err})
	} else {
	    err = tcp.SetTimeouts(remoteTCP, 0)
	    if err != nil {
	        logger.Warn("Failed setting TCP timeouts on remote connection", logger.Ctx{"err": err})
	    }
	}
```

simondeziel: @kadinsayani now I know why I didn't pick up the deprecation: https://github.com/gorilla/websocket/pull/998

simondeziel: Re the suggested warning, isn't that going to be noisy with Unix connections?

lxd/instance/drivers/driver_lxc.go
kadinsayani: Although the surrounding context is useful, the magic number can lead to future bugs if there is ever an unlikely scenario where `fields` changes. Something like this would be nice:

```suggestion
	// Fields in /proc/mounts lines are space-separated, e.g.:
	//   source target fstype options dump pass
	const (
		fieldSource    = 0
		fieldTarget    = 1
		fieldFSType    = 2
		minMountFields = 3 // we only need the first three
	)

	for scanner.Scan() {
		fields := strings.Fields(scanner.Text())
		if len(fields) < minMountFields {
			continue
		}

		mountMap[fields[fieldSource]] = mountInfo{
			Mountpoint: fields[fieldTarget],
			FSType:     fields[fieldFSType],
		}
	}
```

This also acts as documentation for us :)

client/lxd.go
kadinsayani: > @kadinsayani now I know why I didn't pick up the deprecation: [gorilla/websocket#998](https://github.com/gorilla/websocket/pull/998)

Nice catch, thanks for putting that PR up!

kadinsayani: > Re the suggested warning, isn't that going to be noisy with Unix connections?

I didn't consider that. Perhaps that's why the error was ignored in the first place.

shared/netutils/network_linux_cgo.go
kadinsayani: Yes, I think that's a good spot.

client/lxd.go
kadinsayani: I think we can go ahead with your original approach.

lxd/instance/drivers/driver_lxc.go
kadinsayani: Just a comment describing the fields in `/proc/mounts` also works.

simondeziel: Hmm, while that's definitely more idiomatic I feel like it goes against the style of the surrounding code and makes it stick out a bit. So unless you feel strongly about it, I'd keep the magic `3`.

kadinsayani (approved):
LGTM, thanks!

client/lxd.go
tomponline: Yes we we should only try and set TCP timeouts on a TCP connection :)

lxd/instance/drivers/driver_lxc.go
tomponline: Yeah for me, 3 isn't a magic number here anyway, its clearly the required length of the slice because we want that field.

A magic number in my view is something that we send/receive into an API that has meaning outside of its numeric value.

tomponline (approved):
thanks!

lxd/instance/drivers/driver_lxc.go
kadinsayani: @tomponline thanks for clarifying :)

PR #16284, opened 2025-08-24
Drop a few unused function parameters

simondeziel:


lxd/api_project.go
```diff
@@ -732,7 +732,7 @@ func isProjectInUse(projectUsedBy []string) bool {
 }
 
 // Common logic between PUT and PATCH.
-func projectChange(ctx context.Context, s *state.State, project *api.Project, req api.ProjectPut) response.Response {
+func projectChange(s *state.State, project *api.Project, req api.ProjectPut) response.Response {
```
tomponline, line 735: Rather than removing this we should be using it on the DB transaction line instead of the shutdown ctx:

```
err = s.DB.Cluster.Transaction(s.ShutdownCtx, func(ctx context.Context, tx *db.ClusterTx) error {
```

lxd/operations.go
```diff
@@ -1174,7 +1174,7 @@ func autoRemoveOrphanedOperationsTask(stateFunc func() *state.State) (task.Func,
 // behind if a cluster member abruptly becomes unreachable. If the affected cluster members comes
 // back online, these operations won't be cleaned up. We therefore need to periodically clean up
 // such operations.
-func autoRemoveOrphanedOperations(ctx context.Context, s *state.State) error {
+func autoRemoveOrphanedOperations(s *state.State) error {
```
tomponline, line 1177: Rather than removing it, it should be passed into the DB transaction.

PR #15448, opened 2025-04-25
api: Improve logging on file push

skozina:
This is a follow up on https://github.com/canonical/lxd/pull/15400 .
With these changes, if UID is adjusted on `file push`, LXD logs an INFO line such as:
```
INFO   [2025-04-25T08:47:57Z] Requested UID 1522341878 not within idmaprange  file=/root/filemanip instance=filemanip project=test
```

PR #16285, opened 2025-08-25
fix bug about file owner/group change not work on vm

heat1024:
Hello,

I found a bug about file push so push commit for fix it.

After https://github.com/canonical/lxd/pull/15448 merged, file's uid/gid change is available in container type instance only, because code got return from applyEffectiveFileOwnership before run file.Chown if not container type.

The problem is here.

https://github.com/canonical/lxd/blob/lxd-6.4/lxd/instance_file.go#L379-L381

And after return that function, there is no chance run chown.

Here is a test.

- in LXD 6.2, UID and GID can set

```
root@testlxd-1:/home/heat1024# lxc file push test vm-1/root/test --uid 65534 --gid 65534

root@testlxd-1:/home/heat1024# lxc exec vm-1 -- bash -c 'ls -al /root/test'
-rw-r--r-- 1 nobody nogroup 5 Aug 25 17:21 /root/test
```

- after LXD 6.4, UID and GID cannot set if instance is VM

```
root@testlxd-2:/home/heat1024# lxc file push test vm-2/root/test --uid 65534 --gid 65534

root@testlxd-2:/home/heat1024# lxc exec vm-2 -- bash -c 'ls -al /root/test'
-rw-r--r-- 1 root root 5 Aug 25 17:22 /root/test
```

So in this PR, I make a branch before enter applyEffectiveFileOwnership function.
if instance type is not container, set file owner just like old time.

It makes only container type instance will check uidmap allow range.

### DIFF
--- a/lxd/instance_file.go
+++ b/lxd/instance_file.go
@@ -376,9 +376,6 @@ func instanceFileHead(inst instance.Instance, path string) response.Response {
 func applyEffectiveFileOwnership(inst instance.Instance, headers *shared.LXDFileHeaders, file *sftp.File) error {
 	uid := headers.UID
 	gid := headers.GID
-	if inst.Type() != instancetype.Container {
-		return nil
-	}
 
 	c, ok := inst.(instance.Container)
 	if !ok {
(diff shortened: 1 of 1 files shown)


lxd/instance_file.go
```diff
@@ -376,9 +376,6 @@ func instanceFileHead(inst instance.Instance, path string) response.Response {
 func applyEffectiveFileOwnership(inst instance.Instance, headers *shared.LXDFileHeaders, file *sftp.File) error {
 	uid := headers.UID
 	gid := headers.GID
-	if inst.Type() != instancetype.Container {
-		return nil
-	}
```
simondeziel, line 381: Since the function is still only called for the container case, why drop this check?

simondeziel (commented):
Thanks for finding and fixing that bug! BTW, I've confirmed the bug and the fix :)

This PR is looking good, I'd just like some clarification on the changes done to `applyEffectiveFileOwnership()`.

lxd/instance_file.go
simondeziel: NVM, I see there is only one caller for `applyEffectiveFileOwnership()` so it makes sense to only check this once.

simondeziel (approved):

skozina:
Thank you @heat1024 for finding and fixing this!
@tomponline yes, I think this is good to go. I'll look at adding the tests. Thanks!

tomponline:
@skozina please can you cherry-pick this commit into lxd-pkg-snap latest-candidate?

PR #16286, opened 2025-08-25
Init: Recovery of existing storage pool (disk) during join

roosterfish:
Addresses scenario (5) (and (10) implicitly) from the recovery enhancements spec in https://docs.google.com/document/d/1yQFZFOM2TK-cEak10tMAVqKKG5Ma0OwcOc0aW9WEuqY/edit?tab=t.0.
 
In case a cluster member crashed (similar to https://discourse.ubuntu.com/t/lxd-host-server-crashed-need-some-advice/66746/2), but the disk(s) used for the storage pool(s) could be recovered, a user might want to move the disks into a new machine and join this machine back into the existing cluster.

However this requires that `lxd init` can cope with new cluster members which contain disks that already have the storage pool and potential instances/custom volumes.
This PR solves this by simply extending the storage pool's creation. First we probe if we can already mount a storage pool based on the provided config. If we can then the pool already exists on the underlying storage and we just need to recreate its existence in the DB.
Recreating the necessary config is easy as the user is getting asked for each member specific config key when adding the new cluster member (and we now also print a default).
Afterwards the user can run `lxd recover` on this member to restore instances and custom volumes which are already located on the recovered disk.

As a side effect this also allows manually recreating a storage pool's DB entry if you have all the necessary configuration at hand.

Furthermore a new cluster test suite is added to check this scenario.

### DIFF
--- a/doc/howto/cluster_form.md
+++ b/doc/howto/cluster_form.md
@@ -117,7 +117,9 @@ Basically, the initialization process consists of the following steps:
 1. Confirm that all local data for the server is lost when joining a cluster.
 1. Configure server-specific settings (see {ref}`clustering-member-config` for more information).
 
-   You can accept the default values or specify custom values for each server.
+   You can specify custom values for each server.
+   In case you are restoring a lost server but you were able to recover the storage pool's disk, you might want to accept the default
+   values which should help telling LXD how to access the existing underlying storage pool.
 
 <details>
 <summary>Expand to see full examples for <code>lxd init</code> on additional servers</summary>
@@ -131,9 +133,9 @@ Are you joining an existing cluster? (yes/no) [default=no]: yes
 Do you have a join token? (yes/no/[token]) [default=no]: yes
 Please provide join token: eyJzZXJ2ZXJfbmFtZSI6InJwaTAxIiwiZmluZ2VycHJpbnQiOiIyNjZjZmExZDk0ZDZiMjk2Nzk0YjU0YzJlYzdjOTMwNDA5ZjIzNjdmNmM1YjRhZWVjOGM0YjAxYTc2NjU0MjgxIiwiYWRkcmVzc2VzIjpbIjE3Mi4xNy4zMC4xODM6ODQ0MyJdLCJzZWNyZXQiOiJmZGI1OTgyNjgxNTQ2ZGQyNGE2ZGE0Mzg5MTUyOGM1ZGUxNWNmYmQ5M2M3OTU3ODNkNGI5OGU4MTQ4MWMzNmUwIn0=
 All existing data is lost when joining a cluster, continue? (yes/no) [default=no] yes
-Choose "size" property for storage pool "local":
-Choose "source" property for storage pool "local":
-Choose "zfs.pool_name" property for storage pool "local":
+Choose "size" property for storage pool "local" [default=9GiB]:
+Choose "source" property for storage pool "local" [default=local]:
+Choose "zfs.pool_name" property for storage pool "local" [default=local]:
 Would you like a YAML "lxd init" preseed to be printed? (yes/no) [default=no]:
 ```
 
@@ -141,6 +143,9 @@ Would you like a YAML "lxd init" preseed to be printed? (yes/no) [default=no]:
 
 After the initialization process finishes, your server is added as a new cluster member.
 You can check this with [`lxc cluster list`](lxc_cluster_list.md).
+
+In case you have restored a cluster member with a disk that was recovered from a previous cluster member, run
+the `lxd recover` command on this cluster member to recover instances and volumes located on the disk's storage pool.
 
 ## Configure the cluster through preseed files
 
--- a/lxd/api_cluster.go
+++ b/lxd/api_cluster.go
@@ -285,7 +285,7 @@ func clusterGetMemberConfig(ctx context.Context, cluster *db.Cluster) ([]api.Clu
 	}
 
 	for pool, config := range pools {
-		for key := range config {
+		for key, value := range config {
 			if strings.HasPrefix(key, instancetype.ConfigVolatilePrefix) {
 				continue
 			}
@@ -294,6 +294,7 @@ func clusterGetMemberConfig(ctx context.Context, cluster *db.Cluster) ([]api.Clu
 				Entity:      "storage-pool",
 				Name:        pool,
 				Key:         key,
+				Value:       value,
 				Description: fmt.Sprintf("\"%s\" property for storage pool \"%s\"", key, pool),
 			}
 
@@ -302,7 +303,7 @@ func clusterGetMemberConfig(ctx context.Context, cluster *db.Cluster) ([]api.Clu
 	}
 
 	for network, config := range networks {
-		for key := range config {
+		for key, value := range config {
 			if strings.HasPrefix(key, instancetype.ConfigVolatilePrefix) {
 				continue
 			}
@@ -311,6 +312,7 @@ func clusterGetMemberConfig(ctx context.Context, cluster *db.Cluster) ([]api.Clu
 				Entity:      "network",
 				Name:        network,
 				Key:         key,
+				Value:       value,
 				Description: fmt.Sprintf("\"%s\" property for network \"%s\"", key, network),
 			}
 
--- a/lxd/main_init_interactive.go
+++ b/lxd/main_init_interactive.go
@@ -269,10 +269,10 @@ func (c *cmdInit) askClustering(config *api.InitPreseed, server *api.Server) err
 			}
 
 			for i, config := range cluster.MemberConfig {
-				question := "Choose " + config.Description + ": "
+				question := fmt.Sprintf("Choose %s [default=%s]: ", config.Description, config.Value)
 
 				// Allow for empty values.
-				configValue, err := c.global.asker.AskString(question, "", validate.Optional())
+				configValue, err := c.global.asker.AskString(question, config.Value, validate.Optional())
 				if err != nil {
 					return err
 				}
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -216,30 +216,41 @@ func (b *lxdBackend) Create(clientType request.ClientType, op *operations.Operat
 		return nil
 	}
 
-	// Create the storage pool on the storage device.
-	err = b.driver.Create()
-	if err != nil {
-		return err
-	}
-
-	revert.Add(func() { _ = b.driver.Delete(op) })
-
-	// Mount the storage pool.
+	// Check if we can already mount the pool before we have created it.
+	// This is an indicator to check whether the pool already exists on storage.
+	// If it doesn't, then try to create it and perform the mount a second time.
 	ourMount, err := b.driver.Mount()
 	if err != nil {
-		return err
+		// Create the storage pool on the storage device.
+		err = b.driver.Create()
+		if err != nil {
+			return err
+		}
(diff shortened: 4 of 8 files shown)


doc/api-extensions.md
```diff
@@ -2875,3 +2875,13 @@ Endpoints converted to asynchronous behavior:
 * `PATCH /storage-pools/{pool}/volumes/{type}/{vol}/snapshots/{snap}` - Patch storage volume snapshot
 * `PUT /1.0/profiles/{name}` - Update profile
 * `PATCH /1.0/profiles/{name}` - Patch profile
+
+## `storage_source_recover`
+
+As part of the recovery process it might be necessary to recover already existing remote storage pools or physical drives when creating
```
tomponline: > already existing remote storage pools or physical drives

This covers all pool types, so we could make this more concise to just:

> As part of the recovery process it might be necessary to recover existing storage pools previously created by LXD.

doc/api-extensions.md
```diff
@@ -2875,3 +2875,13 @@ Endpoints converted to asynchronous behavior:
...
+a new storage pool in LXD.
+
+Previously this was only possible for some of the drivers (e.g. `ceph.osd.force_reuse`).
```
tomponline: I wouldn't mention `ceph.osd.force_reuse` here because we are dropping this config key, and because they key wasn't intended for use with recovery anyway.

doc/api-extensions.md
```diff
@@ -2875,3 +2875,13 @@ Endpoints converted to asynchronous behavior:
...
+The new pool level `source.recover` configuration key can be set per node to allow reuse of an existing `source`.
```
tomponline: We say "member" not "node" in public facing literature.

tomponline: > The new pool `source.recover` configuration key...

tomponline: > of an existing pool `source`

lxd/storage/utils.go
```diff
@@ -612,7 +612,16 @@ func validatePoolCommonRules() map[string]func(string) error {
 		//  defaultdesc: `false`
 		//  shortdesc: Whether to wipe the block device before creating the pool
 		//  scope: local
-		"source.wipe":             validate.Optional(validate.IsBool),
+		"source.wipe": validate.Optional(validate.IsBool),
+		// lxdmeta:generate(entities=storage-dir,storage-lvm,storage-btrfs,storage-zfs,storage-ceph,storage-cephfs; group=pool-conf; key=source.recover)
+		// Set this option to `true` to recover an existing `source`
+		// which might already contain volumes.
```
tomponline: Suggest:

> Set this option to `true` to recover an existing `source` which was previously created by LXD

tomponline: Whether or not it has volumes is irrelevant.

lxd/api_cluster.go
```diff
@@ -299,6 +300,7 @@ func clusterGetMemberConfig(ctx context.Context, cluster *db.Cluster) ([]api.Clu
 				Entity:      "network",
 				Name:        network,
 				Key:         key,
+				Value:       value,
 				Description: fmt.Sprintf("\"%s\" property for network \"%s\"", key, network),
```
tomponline: whilst you're working in this area, can you also fix up these value quotes to use `%q` ta

lxd/main_init_interactive.go
```diff
@@ -269,10 +269,22 @@ func (c *cmdInit) askClustering(config *api.InitPreseed, server *api.Server) err
 			}
 
 			for i, config := range cluster.MemberConfig {
+				var defaultAnswer string
 				question := "Choose " + config.Description + ": "
 
+				// Don't populate the 'source' key as it is likely depending on the used system.
```
tomponline: Please can this comment be clarified. It seems to be contradicting itself.
It starts with 

> Don't populate the 'source' key as it is likely depending on the used system.

Then goes onto argue that in some cases it'd be OK to do this, and then seems to go ahead and populate it anyway on line 282.

lxd/storage/backend_lxd.go
tomponline, line 1: > This is the case when recovering lost cluster members which still have the underyling storage present but lost LXD.

"lost LXD" would be clearer as "lost the LXD database"

tomponline, line 1: From our 1:1 discussions I had gotten the impression you were going to just call `pool.Mount()` when in recovery mode, rather than calling `pool.Create()` and adding recovery support to it, but this doesn't appear to be what is happening here.

Also unexpected is that `pool.Create()` seems to be silently continuing to create the pool if the mount fails during recovery mode. This doesn't seem correct. In what scenario should a full-blown pool creation occur during recovery mode?

lxd/storage/drivers/driver_ceph.go
```diff
@@ -254,24 +254,18 @@ func (d *ceph) Create() error {
...
-
-			d.config["volatile.pool.pristine"] = "false"
-		} else {
-			// Create placeholder storage volume. Other LXD instances will use this to detect whether this osd
-			// pool is already in use by another LXD instance.
-			err := d.rbdCreateVolume(placeholderVol, "0")
-			if err != nil {
-				return err
-			}
+			return fmt.Errorf("Pool '%s' in cluster '%s' seems to be in use by another LXD instance", d.config["ceph.osd.pool_name"], d.config["ceph.cluster_name"])
```
tomponline: Lets use `%q` here rather than `'%s'`

lxd/storage/drivers/driver_ceph.go
tomponline, line 1: lets land this as a separate PR, along with any other commits from this PR that are not directly related to the recovery changes please

lxd/api_cluster.go
roosterfish: Yes, let me add this in another PR.

lxd/storage/drivers/driver_ceph.go
roosterfish: I'll add this to the other PR about quotes.

lxd/storage/drivers/driver_ceph.go
roosterfish: I wanted to prevent a situation in which we drop the key and don't yet have `source.recover` landed. But as it's edge anyway I think that is not an issue. Will move to another PR we can then merge before this one.

lxd/storage/backend_lxd.go
roosterfish: When we are in recovery mode (`source.recover=true`), we only have to create the missing pool's DB entry which is already implemented in `pool.Create`. Therefore I have modified `pool.Create` to first try `driver.Mount` and only `driver.Create` in case (`source.recover=false`). I didn't consider `pool.Mount` as this one is always called to just mount and call `driver.Mount`.

Because now from a backend's perspective you can always run `pool.Create` (which has the recovery logic) and `pool.Mount` to actually do the mount.

But you raised a good point, we should exit out when `source.recover=true` and we were not able to recover and not just continue on to create the pool as a fall back. Will fix now.

tomponline:
@roosterfish please can you rebase

doc/api-extensions.md
roosterfish: Thanks, updated this and the other comments on this text.

doc/api-extensions.md
roosterfish: Fixed, used LVM's `lvm.vg.force_reuse` instead.

doc/api-extensions.md
roosterfish: Fixed.

roosterfish: Fixed.

roosterfish: Fixed.

lxd/storage/utils.go
roosterfish: Fixed.

lxd/api_cluster.go
roosterfish: Moved to https://github.com/canonical/lxd/pull/17214

lxd/main_init_interactive.go
roosterfish: Fixed by clarifying it's about not populating a default for `source`.

lxd/storage/backend_lxd.go
roosterfish: Fixed by updating the commit message.

lxd/storage/drivers/driver_ceph.go
roosterfish: Fixed and moved to https://github.com/canonical/lxd/pull/17214.

lxd/storage/drivers/driver_ceph.go
roosterfish: Moved to https://github.com/canonical/lxd/pull/17212.

test/includes/clustering.sh
```diff
@@ -273,13 +285,10 @@ EOF
...
     value: lxdtest-$(basename "${TEST_DIR}")-${ns}
-  - entity: storage-pool
-    name: data
-    key: size
-    value: 1GiB
 EOF
       fi
-      if [ "${driver}" = "btrfs" ]; then
+      # shellcheck disable=SC2235
+      if ([ "${driver}" = "btrfs" ] && [ "${source}" = "" ]) || ([ "${driver}" = "zfs" ] && [ "${source}" = "" ]) || ([ "${driver}" = "lvm" ] && [ "${source}" = "" ]); then
```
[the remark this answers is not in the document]
roosterfish: Fixed.

roosterfish:
@tomponline ready for another pass. I have added one more check to only set `source.recover` if recovery mode is entered, so we don't always carry an explicit `source.recover=false` on the pool when not doing recovery.

lxd/storage/backend_lxd.go
```diff
@@ -301,30 +305,52 @@ func (b *lxdBackend) Create(clientType request.ClientType, op *operations.Operat
...
+		}
 
-	// Mount the storage pool.
-	ourMount, err := b.driver.Mount()
-	if err != nil {
-		return err
-	}
+		// Unmount the pool if we have mounted it as part of probing for its existence.
+		if ourMount {
+			defer func() { _, _ = b.driver.Unmount() }()
```
tomponline, line 321: we dont need to defer here right?

doc/howto/cluster_form.md
```diff
@@ -142,6 +144,9 @@ Would you like a YAML "lxd init" preseed to be printed? (yes/no) [default=no]:
 After the initialization process finishes, your server is added as a new cluster member.
 You can check this with [`lxc cluster list`](lxc_cluster_list.md).
 
+In case you have restored a cluster member with a disk that was recovered from a previous cluster member, run
+the `lxd recover` command on this cluster member to recover instances and volumes located on the disk's storage pool.
```
tomponline, line 148: @roosterfish shall we add an this suggestion to the output of `lxd init` if `source.recover` was used? To prompt the user to recover instances? Perhaps we could even offer a question as to whether to jump into `lxd recover` at the end of `lxd init`?

lxd/storage/backend_lxd.go
roosterfish: Actually no, let me fix this in a follow up.

doc/howto/cluster_form.md
roosterfish: Yeah good point, let me explore this option.

lxd/storage/backend_lxd.go
roosterfish: https://github.com/canonical/lxd/pull/16286#discussion_r2609897122

PR #16289, opened 2025-08-25
Network: Limit OVN MAC binding (from Incus)

nmezhenskyi:
Includes cherry-picks from https://github.com/lxc/incus/pull/1160.

## Changes:
This configures OVN so it doesn't fill the MAC_Binding table with every single MAC address it sees through ARP. Instead MAC_Binding entries will be generated for the default gateway only, leading to two entries per network.

### DIFF
--- a/lxd/network/driver_ovn.go
+++ b/lxd/network/driver_ovn.go
@@ -3916,7 +3916,7 @@ func (n *ovn) InstanceDevicePortStart(opts *OVNInstanceNICSetupOpts, securityACL
 				break
 			}
 
-			time.Sleep(100 * time.Millisecond)
+			time.Sleep(250 * time.Millisecond)
 		}
 
 		for _, dynamicIP := range dynamicIPs {
--- a/lxd/network/openvswitch/ovn.go
+++ b/lxd/network/openvswitch/ovn.go
@@ -359,7 +359,16 @@ func (o *OVN) LogicalRouterAdd(routerName OVNRouter, mayExist bool) error {
 		args = append(args, "--may-exist")
 	}
 
-	_, err := o.nbctl(append(args, "lr-add", string(routerName))...)
+	// Create a logical router.
+	args = append(args, "lr-add", string(routerName), "--")
+
+	// Set its properties.
+	args = append(args, "set", "logical_router", string(routerName),
+		"options:always_learn_from_arp_request=false",
+		"options:dynamic_neigh_routers=true",
+	)
+
+	_, err := o.nbctl(args...)
 	if err != nil {
 		return err
 	}


lxd/network/openvswitch/ovn.go
```diff
@@ -359,7 +359,16 @@ func (o *OVN) LogicalRouterAdd(routerName OVNRouter, mayExist bool) error {
 		args = append(args, "--may-exist")
 	}
 
-	_, err := o.nbctl(append(args, "lr-add", string(routerName))...)
+	// Create a logical router.
+	args = append(args, "lr-add", string(routerName), "--")
+
+	// Set its properties.
+	args = append(args, "set", "logical_router", string(routerName),
+		"options:always_learn_from_arp_request=false",
```
tomponline, line 367: Wont this break keepalived which @escabo has been testing recently?

What was the prompt to adopt this change?

nmezhenskyi: > What was the prompt to adopt this change?

@tomponline I was going through the import candidates list to see if there are any improvements that could be added together with the `ipv4.dhcp.ranges` functionality. Enabling `always_learn_from_arp_request=false` and `dynamic_neigh_routers=true` reduces unnecessary and stale neighbor entries in the logical router. I thought this could be useful in environments with a lot of instances connected to the network.

> Wont this break keepalived?

This is a good point. This change makes it so that the OVN network will not learn from Keepalived gratuitous ARP requests, only from its gratuitous ARP replies. Keepalived can be configured to use gratuitous ARP replies as broadcasts, but this is not the default functionality.

nmezhenskyi: > Keepalived can be configured to use gratuitous ARP replies as broadcasts, but this is not the default functionality.

@tomponline Should we hold off on this change then and close the PR for now?

tomponline: I'm not clear what this setting does.

Does it affect only learning on the uplink or does it also affect learning on the internal logical switch too?

I think we should bring this up with @fnordahl at the next LXD/OVN catch up to discuss impact and recommendations.

lxd/network/driver_ovn.go
tomponline, line 1: we can land this separately

lxd/network/openvswitch/ovn.go
tomponline: @nmezhenskyi before we change anything we should add an IP failover test using ARP/NDP announcements into lxd-ci to ensure we dont regression behaviour we depend on.

nmezhenskyi: > Does it affect only learning on the uplink or does it also affect learning on the internal logical switch too?

It affects both.

nmezhenskyi: > before we change anything we should add an IP failover test using ARP/NDP announcements into lxd-ci to ensure we dont regression behaviour we depend on.

Makes sense, I'll work on this.

nmezhenskyi:
Closing this for now until the IP failover test using ARP/NDP announcements is added to `lxd-ci`. I'll make a separate PR for this small change https://github.com/canonical/lxd/pull/16289/commits/ecb5e032b2080c2081693afccc5c37967c2bc608.

PR #16304, opened 2025-08-26
Fix remaining `golangci-lint` issues

simondeziel:


simondeziel:
The `golangci-lint` GHA is slower (~1m40s vs 50s) but I'm hoping that's attributable to the cache miss from the `actions/setup-go` as it is supposed to leverage it. Time will tell.

simondeziel:
> The `golangci-lint` GHA is slower (~1m40s vs 50s) but I'm hoping that's attributable to the cache miss from the `actions/setup-go` as it is supposed to leverage it. Time will tell.

It is super fast again now when doing cache hits, it lints the whole deal inside ~12s.

PR #16306, opened 2025-08-26
doc: info about snap upgrades, downgrades, revert

minaelee:
This PR adds a section to releases-snap.md to discuss upgrading/downgrading/reverting snaps, including that:
- downgrades from a newer to an older track is not supported
- reverting from a newer to an older feature release revision is also not supported
- includes cross-linking between this section and the existing "Change the snap channel" how-to guide

### DIFF
--- a/doc/reference/releases-snap.md
+++ b/doc/reference/releases-snap.md
@@ -119,6 +119,15 @@ New LXD releases are [published progressively as snaps](https://documentation.ub
 
 This can cause issues when updating the LXD snap for {ref}`clusters <exp-clusters>`, as cluster members must use the same version of the snap at all times. For a guide on how to avoid this issue using the `--cohort` flag, see {ref}`howto-snap-updates-sync`.
 
+(ref-snap-upgrades-downgrades)=
+### Upgrades, downgrades, and reverting
+
+To upgrade LXD means to move to a newer {ref}`track <ref-snap-tracks>`, such as from {{current_lts_track}} to {{current_feature_track}}. For instructions, see: {ref}`howto-snap-change`.
+
+Downgrading (moving to a channel with an older track, such as from {{current_feature_track}} to {{current_lts_track}}) is neither recommended nor supported, as breaking changes can exist between major versions.
+
+Breaking changes can be introduced between feature release versions as well. Thus, we also do not support reverting to an older feature release revision.
+
 ## Related topics
 
 How-to guides:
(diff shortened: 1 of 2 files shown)


tomponline (commented):
@minaelee @roosterfish showed me we have some reference to this in https://documentation.ubuntu.com/lxd/latest/installing/#upgrade-lxd should we combine the warnings/info so we can reference it from one place?

doc/reference/releases-snap.md
```diff
@@ -119,6 +119,15 @@ New LXD releases are [published progressively as snaps](https://documentation.ub
 
 This can cause issues when updating the LXD snap for {ref}`clusters <exp-clusters>`, as cluster members must use the same version of the snap at all times. For a guide on how to avoid this issue using the `--cohort` flag, see {ref}`howto-snap-updates-sync`.
 
+(ref-snap-upgrades-downgrades)=
+### Upgrades, downgrades, and reverting
+
+To upgrade LXD means to move to a newer {ref}`track <ref-snap-tracks>`, such as from {{current_lts_track}} to {{current_feature_track}}. For instructions, see: {ref}`howto-snap-change`.
```
simondeziel, line 125: Moving from 5.21.3 to 5.21.4 happens in the same `5.21/stable` track but I'd still call it an upgrade because the version moves forward.

doc/reference/releases-snap.md
```diff
@@ -119,6 +119,15 @@ New LXD releases are [published progressively as snaps](https://documentation.ub
...
+
+Downgrading (moving to a channel with an older track, such as from {{current_feature_track}} to {{current_lts_track}}) is neither recommended nor supported, as breaking changes can exist between major versions.
```
simondeziel, line 127: Moving from 5.21.4 to 5.21.3 happens in the same `5.21/stable` track but because the versions moves backward, I'd call this a downgrade. However, it's a supported downgrade path because the track stayed the same.

doc/reference/releases-snap.md
minaelee: How would you differentiate between an `upgrade` and an `update` of the snap?

simondeziel: I think we can agree that an `upgrade` is a bigger change than an `update`. However, I'm not sure if `update` should be reserved to snap revisions or "patch level" jumps (like 5.21.3->5.21.4).

tomponline:
Any update on this @minaelee ?

minaelee:
Closing per overly similar to earlier PR https://github.com/canonical/lxd/pull/16076 — I've merged the changes in this one with that PR, including addressing @tomponline's comment re: moving the info from `installing.md`. @simondeziel I'd appreciate a review over there, though I have left the `upgrade` definition as-is (changing to a higher/newer track only, not updates that occur within the same track) per discussions.

PR #16314, opened 2025-08-26
shared/util: Break slice aliasing in RemoveElementsFromSlice

MusicDin:
Make sure that slice aliasing is broken (slice is cloned) before removing elements using `slices.Delete` in helper function `RemoveElementsFromSlice`. The `slices.Delete` zeros out the slice tail after element removal which affects the original slice.

This was caught during Pure testing.
When VM volume is being refreshed, a slice of snapshot names that need to be refreshed are passed into `refreshVolume` function twice - once for FS and once for block volume. 
Now, when the FS volumes are refreshed we remove snapshot names that were found from the passed slice. When the slice is passed again, it contains empty name.


shared/util.go
```diff
@@ -738,6 +738,7 @@ func StringPrefixInSlice(key string, list []string) bool {
 // RemoveElementsFromSlice returns a slice equivalent to removing the given elements from the given list.
 // Elements not present in the list are ignored.
 func RemoveElementsFromSlice[T comparable](list []T, elements ...T) []T {
+	list = slices.Clone(list)
```
tomponline, line 742: please update description to mention the input is cloned

MusicDin:
> Does this need cherry picking 5.21 LTS too?

Yes please, I'll open PR now

tomponline:
@MusicDin please can you cherry-pick this into lxd-pkg-snap latest-candidate

PR #16320, opened 2025-08-26
Network: Add support for IPv4 DHCP ranges for OVN networks (WIP)

nmezhenskyi:
## Changes:
- Adds `ovn_dhcp_ranges` API extension.
- Adds `ipv4.dhcp.ranges` config option for `ovn` networks that allows specifying a list of IPv4 ranges
reserved for dynamic allocation using DHCP. IPs outside of these ranges will not be used for DHCP.


test/suites/network_ovn.sh
```diff
@@ -473,6 +473,37 @@ test_network_ovn() {
...
+  setup_instance_ip4_interface "c2"
+  setup_instance_ip4_interface "c3"
+
+  echo "Check that the 4th instance creation fails because all reserved dynamic addresses are taken."
+  ! lxc launch testimage c4 --network "${ovn_network}" || false
+
+  echo "Check IPs assigned to instances."
+  [ "$(lxc info c1 | awk '/inet:.*(global)/ {split($2, a, "/"); print a[1]}')" = "10.24.140.10" ]
+  [ "$(lxc info c2 | awk '/inet:.*(global)/ {split($2, a, "/"); print a[1]}')" = "10.24.140.11" ]
+  [ "$(lxc info c3 | awk '/inet:.*(global)/ {split($2, a, "/"); print a[1]}')" = "10.24.140.12" ]
```
simondeziel: That's some fancy `awk` ;)
Would it also work with `[ "$(lxc list -f csv -c 4 c1)" = "10.24.140.10 (eth0)"` ]`?

nmezhenskyi: Yes, seems to work, thanks! This is definitely more readable :)

doc/api-extensions.md
```diff
@@ -2714,3 +2714,8 @@ This change removes the previous limitation on `ovn` networks that load balancer
 
 Adds a new `bearer` authentication method and enables authentication to the DevLXD API.
 See {ref}`DevLXD bearer tokens <devlxd-authentication-bearer>`.
+
+## `ovn_dhcp_ranges`
+
+This introduces support for the {config:option}`network-ovn-network-conf:ipv4.dhcp.ranges` configuration key for `ovn` networks. This key allows specifying a list of IPv4 ranges
```
tomponline, line 2738: what about `ipv6.dhcp.ranges`?

LXD OVN networks support stateful DHCPv6 https://documentation.ubuntu.com/lxd/latest/reference/network_ovn/#network-ovn-netwo

nmezhenskyi: @tomponline I found that only IPv4 IPs can be specified in the OVN's `exclude_ips` field, so wasn't sure how to proceed with `ipv6.dhcp.ranges`.

tomponline: Ah right I see.

tomponline:
@nmezhenskyi needs a rebase please

doc/api-extensions.md
nmezhenskyi: From what I understand, the only way to exclude an IPv6 IP from being allocated by DHCP is to statically assign it to a specific switch port.

tomponline: If its an OVN limitation we cannot support it at this time. cc @fnordahl

nmezhenskyi:
@tomponline rebased

lxd/network/driver_ovn.go
```diff
@@ -2190,15 +2202,71 @@ func (n *ovn) validateUplinkNetwork(ctx context.Context, tx *db.ClusterTx, p *ap
...
+	routerIntPortIPv4, ipv4Net, err := n.parseRouterIntPortIPv4Net()
 	if err != nil {
 		return nil, fmt.Errorf("Failed parsing router's internal port IPv4 Net for DHCP reservation: %w", err)
 	}
 
 	var dhcpReserveIPv4s []shared.IPRange
 
 	if routerIntPortIPv4 != nil {
-		dhcpReserveIPv4s = []shared.IPRange{{Start: routerIntPortIPv4}}
+		if n.config["ipv4.dhcp.ranges"] == "" {
```
tomponline, line 2210: Can we break this new section out into a new function on the `ovn` struct that is called by the `getDHCPv4Reservations` function.

lxd/network/driver_ovn.go
```diff
@@ -2190,15 +2202,71 @@ func (n *ovn) validateUplinkNetwork(ctx context.Context, tx *db.ClusterTx, p *ap
...
+			dhcpReserveIPv4s = []shared.IPRange{{Start: routerIntPortIPv4}}
+		} else {
+			allowedNets := []*net.IPNet{n.DHCPv4Subnet()}
+			dhcpRanges, err := shared.ParseIPRanges(n.config["ipv4.dhcp.ranges"], allowedNets...)
+			if err != nil {
+				return nil, err
+			}
+
+			lastIP := routerIntPortIPv4
```
tomponline: if im reading this right we seem to lose the `routerIntPortIPv4` from the DHCP reservation list because its overwritten below when parsing the dhcp ranges before it is added to the `dhcpReserveIPv4s`.

We want to add `routerIntPortIPv4` to `dhcpReserveIPv4s` in all cases right?

lxd/network/driver_ovn.go
```diff
@@ -2190,15 +2202,71 @@ func (n *ovn) validateUplinkNetwork(ctx context.Context, tx *db.ClusterTx, p *ap
...
+			containsIP := false
+			for _, reservedDhcpRange := range dhcpReserveIPv4s {
+				containsIP = reservedDhcpRange.ContainsIP(ip)
+				if containsIP {
+					break
+				}
+			}
+
+			if !containsIP {
+				dhcpReserveIPv4s = append(dhcpReserveIPv4s, shared.IPRange{Start: ip})
```
tomponline: Why are we including static DHCP assignments in the logical switch's `ExcludeIPv4` list?

These are excluded from DHCP anyway because they will be assigned to a logical switch port.

Am I missing something?

tomponline: I think you might be being lead by an inaccurate comment in the call site:

> // Get a list of all NICs connected to this network that have static DHCP IPv4 reservations.

I don't think this is relevant anymore since we moved to NIC add-time setup of logical switch ports rather than start-time logical switch ports, and this might explain why the Incus cherry-picks are different too.

lxd/network/driver_ovn.go
nmezhenskyi: > We want to add routerIntPortIPv4 to dhcpReserveIPv4s in all cases right?

@tomponline Yes, we do.

This commit in general is faulty, but is a prerequisite for the next one, which improves on it. Since they're both cherry-picked from Incus in a succession I wasn't sure how to organize this better.

Should I try to merge them into a single commit and leave out the incorrect part?

tomponline: > Should I try to merge them into a single commit and leave out the incorrect part?

yes please

lxd/network/driver_ovn.go
nmezhenskyi: I've simplified this section with the use of a separate `complementRangesIP4()` function. It should be more readable now. But I can split it even further if needed.

lxd/network/driver_ovn.go
nmezhenskyi: Updated

lxd/network/driver_ovn.go
nmezhenskyi: > I don't think this is relevant anymore since we moved to NIC add-time setup of logical switch ports rather than start-time logical switch ports, and this might explain why the Incus cherry-picks are different too.

Yes, indeed. Thanks for pointing out. I've removed that part.

tomponline:
@nmezhenskyi please rebase

lxd/network/driver_ovn.go
```diff
@@ -2190,15 +2200,38 @@ func (n *ovn) validateUplinkNetwork(ctx context.Context, tx *db.ClusterTx, p *ap
...
+		if n.config["ipv4.dhcp.ranges"] == "" {
+			dhcpReserveIPv4s = []shared.IPRange{{Start: routerIntPortIPv4}}
+		} else {
+			allowedNets := []*net.IPNet{n.DHCPv4Subnet()}
+			dhcpRanges, err := shared.ParseIPRanges(n.config["ipv4.dhcp.ranges"], allowedNets...)
+			if err != nil {
+				return nil, err
+			}
+
+			sort.Slice(dhcpRanges, func(i, j int) bool {
```
tomponline: is the sorting important for `complementRangesIP4`? If so it should be done inside `complementRangesIP4`.

lxd/network/network_utils.go
```diff
@@ -1504,3 +1505,67 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
 
 	return allowedUplinkNetworkNames, nil
 }
+
+// complementRangesIP4 returns the complement of the provided IPv4 network ranges.
+// It calculates the IPv4 ranges that are *not* covered by the input slice.
+func complementRangesIP4(ranges []*shared.IPRange, netAddr *net.IPNet) ([]shared.IPRange, error) {
```
tomponline, line 1517: what is `netAddr` here? Comment could be improved.

tomponline: Do we need the complexity of this function now that we are not including instance NIC static DHCP reservations in this excluded IPs list?

Can we convert the ranges to a map of unique IPs and return that?

lxd/network/driver_ovn.go
nmezhenskyi: Yes, the sorting is important for the `complementRangesIP4` because the function assumes that each IP range in the loop will start after the previous end.

lxd/network/network_utils.go
nmezhenskyi: > what is `netAddr` here? 

It's the network address on which IP ranges are specified. I.e, an OVN IPv4 network address.

nmezhenskyi: > Can we convert the ranges to a map of unique IPs and return that?

@tomponline I'm not sure I understand this correctly.

`complementRangesIP4` returns a list of IP ranges that are "outside" of a given list of ranges parameter. This is done because we want to write the IPs that are "outside" of the `ipv4.dhcp.ranges` into the `exclude_ips` field on the OVN network.

Do you mean we should return a map with all these "outside" IPs stored individually instead?

nmezhenskyi:
Rebased

lxd/network/network_utils.go
tomponline: Ah I see, right ofcourse, please can you add a comment to the call site of `complementRangesIP4` explaining what it is doing and why as I did not follow when reviewing.

nmezhenskyi: Will do, sorry for that

lxd/network/driver_ovn.go
```diff
@@ -2274,7 +2305,7 @@ func (n *ovn) setup(update bool) error {
 		updatedConfig["bridge.mtu"] = strconv.FormatUint(uint64(bridgeMTU), 10)
 	}
 
-	// Get a list of all NICs connected to this network that have static DHCP IPv4 reservations.
+	// Get a list of all reserved IPv4 addresses that must not be used by the network's DHCP.
```
nmezhenskyi, line 2309: I've also updated the comment here, where `n.getDHCPv4Reservations()` is called.

lxd/network/driver_ovn.go
```diff
@@ -2188,17 +2197,39 @@ func (n *ovn) validateUplinkNetwork(ctx context.Context, tx *db.ClusterTx, p *ap
...
+		} else {
+			allowedNets := []*net.IPNet{n.DHCPv4Subnet()}
+			dhcpRanges, err := shared.ParseIPRanges(n.config["ipv4.dhcp.ranges"], allowedNets...)
+			if err != nil {
+				return nil, err
+			}
+
+			// Get the list of IP ranges that are "outside" of the "ipv4.dhcp.ranges".
+			// This is needed because we want to write these IPs into the "exclude_ips"
+			// field on the OVN network to prevent DHCP from using them.
```
nmezhenskyi, line 2221: Added this comment.

lxd/network/driver_ovn.go
nmezhenskyi: Updated. Now the sorting is done inside the `complementRangesIP4` function

lxd/network/network_utils.go
nmezhenskyi: Added more comments

lxd/network/network_utils.go
```diff
@@ -1504,3 +1507,73 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
 
 	return allowedUplinkNetworkNames, nil
 }
+
+// complementRangesIP4 returns the complement of the provided IPv4 network ranges.
+// Accepts a slice of IPv4 ranges and its network's address as parameters.
+// It calculates the IPv4 ranges that are *not* covered by the input slice and
```
tomponline, line 1513: The comment doesn't yet mention what `netAddr` is used for?

lxd/network/network_utils.go
```diff
@@ -1504,3 +1507,73 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
...
+// returns the result.
+func complementRangesIP4(ranges []*shared.IPRange, netAddr *net.IPNet) ([]shared.IPRange, error) {
+	var complement []shared.IPRange
+
+	sort.Slice(ranges, func(i, j int) bool {
```
tomponline, line 1523: Adding a comment here from your post in https://github.com/canonical/lxd/pull/16320#discussion_r2336721135 would be useful for future readers:

> the sorting is important for the complementRangesIP4 because the function assumes that each IP range in the loop will start after the previous end.

lxd/network/network_utils.go
```diff
@@ -1504,3 +1507,73 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
...
+		if err != nil {
+			return nil, err
+		}
+
+		endAddr, err := netip.ParseAddr(r.End.String())
+		if err != nil {
+			return nil, err
+		}
+
+		if startAddr.Compare(previousEnd.Next()) == 1 {
```
tomponline: Lets not call previousEnd.Next twice here and line below, so suggest storing it in var before passing it to `.Compare()`.

lxd/network/network_utils.go
```diff
@@ -1504,3 +1507,73 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
...
+	if previousEnd.Compare(endAddr) == -1 {
+		complement = append(complement, shared.IPRange{Start: net.ParseIP(previousEnd.Next().String()), End: net.ParseIP(endAddr.String())})
+	}
+
+	return complement, nil
+}
+
+// ipInRanges checks whether the given IP address is contained within any of the
+// provided IP network ranges.
+func ipInRanges(ipAddr net.IP, ipRanges []shared.IPRange) bool {
```
tomponline, line 1594: Can we use `range.Overlaps()` here instead?

lxd/network/driver_ovn.go
```diff
@@ -2188,17 +2197,39 @@ func (n *ovn) validateUplinkNetwork(ctx context.Context, tx *db.ClusterTx, p *ap
...
 	if err != nil {
 		return nil, fmt.Errorf("Failed parsing router's internal port IPv4 Net for DHCP reservation: %w", err)
 	}
 
 	var dhcpReserveIPv4s []shared.IPRange
 
 	if routerIntPortIPv4 != nil {
-		dhcpReserveIPv4s = []shared.IPRange{{Start: routerIntPortIPv4}}
+		if n.config["ipv4.dhcp.ranges"] == "" {
+			dhcpReserveIPv4s = []shared.IPRange{{Start: routerIntPortIPv4}}
```
tomponline, line 2211: Is the `routerIntPortIPv4` address still included in dhcpReserveIPv4s even if `ipv4.dhcp.ranges` is set to a range that includes the `routerIntPortIPv4`?

lxd/network/network_utils.go
```diff
@@ -1504,3 +1507,73 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
...
+		endAddr, err := netip.ParseAddr(r.End.String())
+		if err != nil {
+			return nil, err
+		}
+
+		if startAddr.Compare(previousEnd.Next()) == 1 {
+			newStart := previousEnd.Next()
+			newEnd := startAddr.Prev()
+
+			if newStart.Compare(newEnd) == 0 {
```
tomponline, line 1560: This could do with a comment here because the var naming doesnt make it clear what we are trying to ascertain - which i think is if the previous end of the range is the same as the start of the new range?

tomponline: Or is it checking if the range is a single IP?

lxd/network/network_utils.go
nmezhenskyi: `netAddr` is used to establish the boundaries of the network. That is, to find the first IP address of the network and the last IP (broadcast address). Without it, the function wouldn't know about the full IP space that the ranges belong to.

I'll clarify this with the comment.

lxd/network/driver_ovn.go
nmezhenskyi: > Is the `routerIntPortIPv4` address still included in dhcpReserveIPv4s even if `ipv4.dhcp.ranges` is set to a range that includes the `routerIntPortIPv4`?  

Yes, this case is handled below in the `else` clause in the very end:

 ```Go
 // Add the router IP in the reserved IP list, if it is not there yet.
 if !ipInRanges(routerIntPortIPv4, dhcpReserveIPv4s) {
	dhcpReserveIPv4s = append(dhcpReserveIPv4s, shared.IPRange{Start: routerIntPortIPv4})
}
```

File: lxd/network/network_utils.go (hunk 13 above)
Comment (nmezhenskyi) [reply lxd/network/network_utils.go:1594]: > Can we use range.Overlaps() here instead?  

I don't think so, because `range.Overlaps()` function checks whether two IP ranges have IP addresses in common. Whereas `ipInRanges()` checks whether a given IP address is included within any of the IP ranges in the given list.

`ipInRanges()` is just a convenience function really and is used only in one place in `driver_ovn.go` `getDHCPv4Reservations()`: to check that the `routerIntPortIPv4` is always included in the reserved IPs list. I could get rid of this function altogether and explicitly loop over the reserved IP ranges to see if there's `routerIntPortIPv4`.

What do you think?

File: lxd/network/network_utils.go (hunk 10 above)
Comment (nmezhenskyi) [reply lxd/network/network_utils.go:1513]: Updated the function header comment for `complementRangesIP4()` to mention this.

File: lxd/network/network_utils.go (hunk 11 above)
Comment (nmezhenskyi) [reply lxd/network/network_utils.go:1523]: Good idea, added

File: lxd/network/network_utils.go [hunk 16]
```diff
@@ -1504,3 +1507,73 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
...
+
+		if endAddr.Compare(previousEnd) == 1 {
+			previousEnd = endAddr
+		}
+	}
+
+	endAddr, err := netip.ParseAddr(dhcpalloc.GetIP(netAddr, -1).String()) // Returns broadcast address.
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
nmezhenskyi: > The comment states 'Returns broadcast address' but dhcpalloc.GetIP with -1 typically returns the last usable IP address in the subnet, not the broadcast address. For a /24 network, this would be .254, not .255.

This is incorrect. `dhcpalloc.GetIP(netAddr, -1)` returns the broadcast address, whereas `dhcpalloc.GetIP(netAddr, -2)` would return the IP ending in `.254`.

Here is the `GetIP` implementation:

```Go
// GetIP returns a net.IP representing the IP belonging to the subnet for the host number supplied.
func GetIP(subnet *net.IPNet, host int64) net.IP {
	// Convert IP to a big int.
	bigIP := big.NewInt(0)
	bigIP.SetBytes(subnet.IP.To16())

	// Deal with negative offsets.
	bigHost := big.NewInt(host)
	bigCount := big.NewInt(host)
	if host < 0 {
		mask, size := subnet.Mask.Size()

		bigHosts := big.NewFloat(0)
		bigHosts.SetFloat64((math.Pow(2, float64(size-mask))))
		bigHostsInt, _ := bigHosts.Int(nil)

		bigCount.Set(bigHostsInt)
		bigCount.Add(bigCount, bigHost)
	}

	// Get the new IP int.
	bigIP.Add(bigIP, bigCount)

	// Generate an IPv6.
	if subnet.IP.To4() == nil {
		newIP := bigIP.Bytes()
		return newIP
	}

	// Generate an IPv4.
	newIP := make(net.IP, 4)
	binary.BigEndian.PutUint32(newIP, uint32(bigIP.Int64()))
	return newIP
}
```

lxd/network/network_utils.go
nmezhenskyi: > Or is it checking if the range is a single IP?

Yes, this is checking if the calculated gap is just a single IP.

If it is, we then create an IP range consisting of only one IP. If not, we then create an IP range with both `Start`  and `End` fields set.

lxd/network/network_utils.go
```diff
@@ -1504,3 +1507,73 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
...
+		if err != nil {
+			return nil, err
+		}
+
+		if startAddr.Compare(previousEnd.Next()) == 1 {
+			newStart := previousEnd.Next()
+			newEnd := startAddr.Prev()
+
+			if newStart.Compare(newEnd) == 0 {
+				complement = append(complement, shared.IPRange{Start: net.ParseIP(newStart.String())})
```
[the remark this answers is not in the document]
nmezhenskyi: This is not necessary because `IPRange` allows setting only the `Start` field to indicate that the range consists of just a single IP.

```Go
// IPRange defines a range of IP addresses.
// Optionally just set Start to indicate a single IP.
type IPRange struct {
	Start net.IP
	End   net.IP
}
```

lxd/network/network_utils.go
nmezhenskyi: I've added a comment for this section.

lxd/network/network_utils.go
nmezhenskyi: Updated. I wasn't sure how to properly name a variable with `previousEnd.Next()`, so I just named it `previousEndNext`, but I added an explanation comment.

This var basically stores the IP immediately following the previous end of a range.

lxd/network/network_utils.go
tomponline: much clearer thanks!

lxd/network/network_utils.go
tomponline: thanks!

lxd/network/network_utils.go
```diff
@@ -1504,3 +1507,97 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
...
+	sort.Slice(ranges, func(i, j int) bool {
+		return bytes.Compare(ranges[i].Start, ranges[j].Start) < 0
+	})
+
+	ipv4NetPrefix, err := netip.ParsePrefix(netAddr.String())
+	if err != nil {
+		return nil, err
+	}
+
+	// Initialize a cursor to the start of the network. It tracks
```
tomponline, line 1532: just a nit (no need to change) but the comment line wrapping here is odd, this would have been clearer in my view:

> // Initialize a cursor to the start of the network. 
> // It tracks the end of the last covered IP range.

lxd/network/network_utils.go
```diff
@@ -1504,3 +1507,97 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
...
+		// A gap is present only if the start of this range comes *after* the IP
+		// immediately following the previous end (previousEnd.Next()).
+		// This correctly handles adjacent ranges (e.g., ending in .20, starting at .21) by not
+		// flagging them as a gap.
+		if startAddr.Compare(previousEndNext) == 1 {
+			newStart := previousEndNext
+			newEnd := startAddr.Prev()
+
+			// Check if the calculated gap is just a single IP or a multi-IP range.
+			if newStart.Compare(newEnd) == 0 {
```
tomponline, line 1560: Thanks for the extra comments, this is great now.

For future reference this could have been written slightly differently to reduce duplication e.g.

```go
newStart := previousEndNext
newEnd := startAddr.Prev()
newRange := shared.IPRange{Start: net.ParseIP(newStart.String())}
if if newStart.Compare(newEnd) != 0 { // Gap covers multiple IPs so specify an end IP.
    newRange.End: net.ParseIP(newEnd.String())
}

complement = append(complement, newRange)
```

lxd/network/network_utils.go
```diff
@@ -1504,3 +1507,97 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
...
+		return nil, err
+	}
+
+	// Initialize a cursor to the start of the network. It tracks
+	// the end of the last covered IP range.
+	previousEnd := ipv4NetPrefix.Addr()
+
+	// Iterate over the sorted list of given IP ranges to find the gaps between them.
+	for _, r := range ranges {
+		startAddr, err := netip.ParseAddr(r.Start.String())
```
tomponline, line 1538: I don't think we need to be converting to net.IP to string and then into netip.Addr, but instead we should use:

https://pkg.go.dev/net/netip#AddrFromSlice

lxd/network/network_utils.go
```diff
@@ -1504,3 +1507,97 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
...
+		if err != nil {
+			return nil, err
+		}
+
+		endAddr, err := netip.ParseAddr(r.End.String())
```
tomponline, line 1543: https://pkg.go.dev/net/netip#AddrFromSlice

lxd/network/network_utils.go
```diff
@@ -1504,3 +1507,97 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
...
+		// flagging them as a gap.
+		if startAddr.Compare(previousEndNext) == 1 {
+			newStart := previousEndNext
+			newEnd := startAddr.Prev()
+
+			// Check if the calculated gap is just a single IP or a multi-IP range.
+			if newStart.Compare(newEnd) == 0 {
+				// The gap is a single IP. It is allowed to set only the "Start"
+				// field for single IP ranges.
+				complement = append(complement, shared.IPRange{Start: net.ParseIP(newStart.String())})
```
tomponline, line 1563: https://pkg.go.dev/net/netip#AddrFromSlice

lxd/network/network_utils.go
```diff
@@ -1504,3 +1507,97 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
...
+			} else {
+				// The gap consists of multiple IPs, therefore, append a range with both "Start" and "End" fields set.
+				complement = append(complement, shared.IPRange{Start: net.ParseIP(newStart.String()), End: net.ParseIP(newEnd.String())})
```
tomponline, line 1566: https://pkg.go.dev/net/netip#AddrFromSlice

lxd/network/network_utils.go
```diff
@@ -1504,3 +1507,97 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
...
+
+		// Advance the cursor to the end of the current range if it extends further
+		// than the previous one. This correctly handles overlapping ranges.
+		if endAddr.Compare(previousEnd) == 1 {
+			previousEnd = endAddr
+		}
+	}
+
+	// Set "endAddr" to the end of the network (broadcast address).
+	endAddr, err := netip.ParseAddr(dhcpalloc.GetIP(netAddr, -1).String()) // Returns broadcast address.
```
tomponline, line 1578: https://pkg.go.dev/net/netip#AddrFromSlice

lxd/network/network_utils.go
```diff
@@ -1504,3 +1507,97 @@ func AllowedUplinkNetworks(ctx context.Context, tx *db.ClusterTx, projectConfig
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Check for a final gap between the end of the last processed range
+	// and the end of the network.
+	if previousEnd.Compare(endAddr) == -1 {
+		complement = append(complement, shared.IPRange{Start: net.ParseIP(previousEnd.Next().String()), End: net.ParseIP(endAddr.String())})
```
tomponline, line 1586: https://pkg.go.dev/net/netip#AddrFromSlice

lxd/network/network_utils.go
nmezhenskyi: Updated in https://github.com/canonical/lxd/pull/16525.

lxd/network/network_utils.go
nmezhenskyi: Updated in https://github.com/canonical/lxd/pull/16525.

lxd/network/network_utils.go
nmezhenskyi: Updated in https://github.com/canonical/lxd/pull/16525.

lxd/network/network_utils.go
nmezhenskyi: Updated in https://github.com/canonical/lxd/pull/16525.

lxd/network/network_utils.go
nmezhenskyi: We can't use `AddrFromSlice` here because we're converting from `netip.Addr` to `net.IP`. I couldn't find a way that would avoid parsing a string here.

lxd/network/network_utils.go
nmezhenskyi: Updated

PR #16325, opened 2025-08-26
Minor tweaks

simondeziel:


shared/validate/validate.go
```diff
@@ -482,15 +482,11 @@ func IsNetworkMTU(value string) error {
 
 // IsNetworkPort validates an IP port number >= 0 and <= 65535.
 func IsNetworkPort(value string) error {
-	port, err := strconv.ParseUint(value, 10, 32)
+	_, err := strconv.ParseUint(value, 10, 16)
 	if err != nil {
 		return fmt.Errorf("Invalid port number %q", value)
 	}
 
-	if port > 65535 {
```
tomponline, line 494: the old error was more user friendly in my view than the new error we will get.

simondeziel: Indeed, `Invalid port number "2555555"` is less friendly but we are more consistent with one providing `-1`.

PR #16329, opened 2025-08-27
devlxd: Add API access handlers

MusicDin:
This PR adds support for defining access handlers in preparation for devLXD volume management endpoints.

Access handlers must be run before the endpoint handler, but also after the devLXD instance is retrieved, as devLXD is not allowed to do any cross-project operations. Therefore, the hoist functions are updated to also accept the access handler.

Access handlers follow the same approach as main LXD API - either access handler has to be defined or `AllowUntrusted` has to be enabled.

### DIFF
--- a/lxd/api_vsock.go
+++ b/lxd/api_vsock.go
@@ -9,35 +9,39 @@ import (
 	"github.com/canonical/lxd/lxd/db"
 	"github.com/canonical/lxd/lxd/db/cluster"
 	"github.com/canonical/lxd/lxd/instance"
-	"github.com/canonical/lxd/lxd/request"
-	"github.com/canonical/lxd/lxd/response"
 	"github.com/canonical/lxd/lxd/state"
 	"github.com/canonical/lxd/lxd/util"
 	"github.com/canonical/lxd/shared/api"
 )
 
+// vSockServer creates an http.Server capable of handling /dev/lxd requests over vsock.
 func vSockServer(d *Daemon) *http.Server {
-	isVsock := true
-
 	return &http.Server{
-		Handler: devLXDAPI(d, hoistReqVM, isVsock),
+		Handler: devLXDAPI(d, vSockAuthenticator{}),
 	}
 }
 
-// hoistReqVM authenticates a VM accessing /dev/lxd over vsock using its agent certificate,
-// identifies and retrieves the corresponding instance, and passes it to the handler if trusted.
-func hoistReqVM(d *Daemon, r *http.Request, handler devLXDAPIHandlerFunc) response.Response {
+// vSockAuthenticator implements DevLXDAuthenticator for vsock connections.
+type vSockAuthenticator struct{}
+
+// IsVsock returns true indicating that this authenticator is used for vsock connections.
+func (vSockAuthenticator) IsVsock() bool {
+	return true
+}
+
+// AuthenticateInstance authenticates a VM accessing /dev/lxd over vsock using its agent certificate,
+// and returns the corresponding VM instance.
+func (vSockAuthenticator) AuthenticateInstance(d *Daemon, r *http.Request) (instance.Instance, error) {
 	trusted, inst, err := authenticateAgentCert(d.State(), r)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


MusicDin:
Fixed unnecessary comment wrapping (based on comments from https://github.com/canonical/lxd/pull/16278)

lxd/api_vsock.go
```diff
@@ -26,7 +26,7 @@ func vSockServer(d *Daemon) *http.Server {
 
 // hoistReqVM authenticates a VM accessing /dev/lxd over vsock using its agent certificate,
 // identifies and retrieves the corresponding instance, and passes it to the handler if trusted.
-func hoistReqVM(d *Daemon, r *http.Request, handler devLXDAPIHandlerFunc) response.Response {
+func hoistReqVM(d *Daemon, r *http.Request, handler devLXDAPIHandlerFunc, accessHandler devLXDAPIHandlerFunc) response.Response {
```
markylaing: I think we should do away with passing the handler into the hoist function. The `devLXDAPIHandlerFunc` no longer accepts an instance as an argument and instead expects the instance to be present in the request context. 

Instead, these should be rewritten to something like:
```go
func authenticateVSock(d *Daemon, r *http.Request) (instance.Instance, error) { ... }
```
and
```go
func authenticateContainer(d *Daemon, r *http.Request) (instance.Instance, error) { ... }
```

These can be called in the main devLXD HTTP handle func, with the instance set in the context before setting the handler.

markylaing: Alternatively these functions could be defined on structs to implement some interface, which can be passed in when instantiating the devlxd or vsock servers.

lxd/devlxd.go
```diff
@@ -46,11 +46,20 @@ type devLXDAPIHandlerFunc func(*Daemon, *http.Request) response.Response
...
+	// Handler is the function that handles the request.
 	Handler devLXDAPIHandlerFunc
+
+	// AccessHandler is an optional function that can be used to check access
+	// permissions before calling the main handler.
+	AccessHandler devLXDAPIHandlerFunc
+
+	// Whether this action allows untrusted requests.
+	// This option must be enabled if the access handler is not configured.
+	AllowUntrusted bool
```
markylaing: I don't see a reason to use a separate struct here vs the standard `APIEndpointAction` we have everywhere else.

markylaing (commented):
Looks mostly good but I think this is an opportunity to remove further cruft that separates DevLXD API handlers from handlers in the main API. As we expand the DevLXD API, I think it's important to keep things as consistent as we can.

lxd/devlxd.go
MusicDin: DevLXD expanded to the point to match LXD API endpoints. Will fix 🚀

lxd/devlxd.go
```diff
@@ -485,6 +485,11 @@ func registerDevLXDEndpoint(d *Daemon, apiRouter *mux.Router, apiVersion string,
 	handleFunc := func(w http.ResponseWriter, r *http.Request) {
 		var requestor request.RequestorArgs
 
+		// Indicate whether the devLXD is being accessed over vsock. This allowes the handler
+		// to determine the correct response type. The responses over vsock are always
+		// in api.Response format, while the responses over Unix socket are in devLXDResponse format.
+		request.SetContextValue(r, request.CtxDevLXDOverVsock, authenticator.IsVsock())
```
markylaing, line 491: It looks like this is already done in the vsock autheticate function? Perhaps `IsVsock` isn't required

MusicDin: Thanks, I've removed context setters from authenticator.

I was experimenting with both options. However, I find it easier if authenticator function does not set anything in context. Instead, `IsVsock()` reports whether it is used for vsock or not, and instance is similarly returned from `AuthenticateInstance()`. This way, everything that is set in the request context is visible in one place (`registerDevLXDEndpoint()`). 

Any thoughts?

markylaing: Yeah that makes sense to me, thanks :)

markylaing (approved):
LGTM

tomponline (approved):

PR #7380, opened 2020-05-18
Storage: Support VM image unpack size greater than volume size

tomponline:
This PR allows images larger than the default 10GB block disk size to be used.

Fixes https://github.com/lxc/lxd/issues/7346 

Create a VM image larger than the default 10GB block disk size:
```
lxc init images:ubuntu/focal/cloud v1 --vm
lxc config device override v1 root size=15GB // Enlarge beyond default 10GB block disk size
lxc start v1  // Wait for boot and cloud init to resize rootfs inside guest
lxc exec v1 -- df -h  | grep /dev/sda2
/dev/sda2        14G  796M   13G   6% /
lxc stop v1
lxc publish v1 --alias myvm
lxc delete v1
```

Now check we can create new VM from the image when the storage pool has no `volume.size` set.
```
lxc init myvm v1
sudo zfs get volsize default/images/f037ad0b451277b594d9f5d80f518f77a6f1a7184a2c7d55939946192ee9a896.block
NAME                                                                                   PROPERTY  VALUE    SOURCE
default/images/f037ad0b451277b594d9f5d80f518f77a6f1a7184a2c7d55939946192ee9a896.block  volsize   14.0G    local
sudo zfs get volsize default/virtual-machines/v1.block
NAME                               PROPERTY  VALUE    SOURCE
default/virtual-machines/v1.block  volsize   14.0G    local
```

Now check that this fails if `volume.size` is set on the pool smaller than the image size when the larger optimized snapshot volume already exists.
```
lxc delete v1
lxc storage set default volume.size 9GB
lxc init myvm v1
Creating v1
Error: Create instance from image: Source image size (14999994368) exceeds specified volume size (9000000000)
```

Now check the optimized snapshot volume creation fails for the same reason:

```
lxc storage volume delete default image/f037ad0b451277b594d9f5d80f518f77a6f1a7184a2c7d55939946192ee9a896
lxc init myvm v1
Creating v1
Error: Create instance from image: Source image size (14999994368) exceeds specified volume size (9000000000)
```

Now check we can set pool's `volume.size` larger than image and optimized snapshot and resulting instance volume use the larger size.

```
lxc storage set default volume.size 16GB
lxc init myvm v1
sudo zfs get volsize default/images/f037ad0b451277b594d9f5d80f518f77a6f1a7184a2c7d55939946192ee9a896.block
NAME                                                                                   PROPERTY  VALUE    SOURCE
[...]

PR #16330, opened 2025-08-27
Storage: Change powerflex to ensure the same volume size setting is used as other drivers

roosterfish:
For drivers using optimized image storage picking only the size which is explicitly defined is important. See for reference https://github.com/canonical/lxd/pull/7380.

PowerFlex doesn't use optimized image storage but for consistency reasons we also use the explicitly defined volume size.


roosterfish:
Tests running...

tomponline (approved):

roosterfish:
Tests passing.

tomponline:
No need to back port or cherry pick as its a no op

PR #16331, opened 2025-08-27
api: fix file push bug about owner/group change not work on vm (stable-5.21)

skozina:


tomponline (approved):
Thanks!

tomponline:
@skozina good to be cherry-picked into 5.21-candidate in lxd-pkg-snap now.

PR #15400, opened 2025-04-15
api: On file push to containers, only change uid/gid if it fits within uidmap range

skozina:
Fixes #15368

PR #16335, opened 2025-08-27
Fix `file push` of a directory owned by uid outside of idmap

skozina:
This is analogous to #15400.

If we use `lxc file push -p` into a container with the option to create dirs (`-p`) , lxd will try to change directory ownership to a uid possibly outside of the allowed idmap range, causing an error (which will cancel the operation before the file itself is tranferred):

```
# ls -l file.txt
-rw-r--r-- 1 1234123412 ubuntu 0 Aug 27 09:36 file.txt
# lxc file push -p file.txt c1/root/temp/file.txt
Error: sftp: "chown /root/temp: invalid argument" (SSH_FX_FAILURE)
```


tomponline (approved):
Thanks!

PR #16340, opened 2025-08-28
lxd/storage/drivers/generic_vfs: don't ask to map volume if mounted

mihalicyn:
We shouldn't instruct getDevicePath() callback to map the volume if we see that mount refcounter is not zero as it means that volume is mapped and accessible from the machine.

For Pure/Powerflex this change is just an optimization.

For HPE Alletra driver this fixes a real issue when "nvme connect" fails with a really weird errors like:
```
Failed to run: nvme connect --transport tcp --traddr 192.168.168.2 --nqn nqn.2020-07.com.hpe:7297d0bc-f8fe-4815-ac36-30957526ec6b --hostnqn nqn.2014-08.org.nvmexpress:uuid:0198ccbf-f421-71bd-9bf0-d71e6af8e308 --hostid 0198ccbf-f421-71bd-9bf0-d71e6af8e308: exit status 1 (Failed to write to /dev/nvme-fabrics: Input/output error\ncould not add new controller: failed to write to nvme-fabrics device)
```
after a volume resize.

### DIFF
--- a/lxd/storage/drivers/generic_vfs.go
+++ b/lxd/storage/drivers/generic_vfs.go
@@ -1213,7 +1213,7 @@ func mountVolume(d Driver, vol Volume, getDevicePath getVolumePathFunc, op *oper
 	defer revert.Fail()
 
 	// Activate volume if needed.
-	volDevPath, cleanup, err := getDevicePath(vol, true)
+	volDevPath, cleanup, err := getDevicePath(vol, !vol.MountInUse())
 	if err != nil {
 		return err
 	}


lxd/storage/drivers/generic_vfs.go
```diff
@@ -1213,7 +1213,7 @@ func mountVolume(d Driver, vol Volume, getDevicePath getVolumePathFunc, op *oper
 	defer revert.Fail()
 
 	// Activate volume if needed.
-	volDevPath, cleanup, err := getDevicePath(vol, true)
+	volDevPath, cleanup, err := getDevicePath(vol, !vol.MountInUse())
```
simondeziel, line 1216: I wonder if that could help with the `zfs` migration issue I was having. I'll keep an eye on this one, thanks.

tomponline:
@mihalicyn any idea why this is only needed for hpe? Just don't want to be papering over an issue.

lxd/storage/drivers/generic_vfs.go
mihalicyn: This function is only used for Pure/Powerflex/HPE Alletra for now.

mihalicyn:
> @mihalicyn any idea why this is only needed for hpe? Just don't want to be papering over an issue.

1. As I mentioned in the commit message this change is useful for all these remote storages, otherwise I would just do some extra check in `getDevicePath` callback in HPE Alletra driver instead of changing a generic function.

2. I don't know what is the root cause, as most likely it is something on the array side. We are getting a failure from the kernel while doing `nvme connect ...`. All the command line arguments are fully correct. But I've noticed that our attempt to connect was meaningless, because we were connected to the array at this point. Then I started to trace this back and noticed that we basically calling map volume each time even if it is not needed. It can be that HPE thing is just more picky about this NVMe/TCP sessions than Pure/Powerflex. Let me know if you want me to dive in the kernel to research more.

tomponline:
> I don't know what is the root cause

OK that is what I was a bit concerned by.
I'll merge for now, but my worry here is that it'll come back and bite us elsewhere as we don't understand what is hspoenkng here - might be a restriction we're not aware of.

When we last chatted you were planning to setup a minimal reproducer. Did that work out?

tomponline:
@mihalicyn @roosterfish @MusicDin just to check: 

Will this cause any impact to the mount reference counter itself (I.e cause it to not be incremented?). Thanks

mihalicyn:
>OK that is what I was a bit concerned by.
>I'll merge for now, but my worry here is that it'll come back and bite us elsewhere as we don't understand what is >hspoenkng here - might be a restriction we're not aware of.

Most likely the restriction is to connect to the array if you are already connected.
It is a proprietary software, not so much we can do.

> When we last chatted you were planning to setup a minimal reproducer. Did that work out?

Of course.

```
lxc launch ubuntu:j v1 -s demo-alletra-pool -d root,size=3584MiB --vm
lxc config device set v1 root size=4GiB
lxc stop -f v1
lxc start v1
```

tomponline:
> > OK that is what I was a bit concerned by.
> > I'll merge for now, but my worry here is that it'll come back and bite us elsewhere as we don't understand what is >hspoenkng here - might be a restriction we're not aware of.
> 
> Most likely the restriction is to connect to the array if you are already connected. It is a proprietary software, not so much we can do.
> 
> > When we last chatted you were planning to setup a minimal reproducer. Did that work out?
> 
> Of course.
> 
> ```
> lxc launch ubuntu:j v1 -s demo-alletra-pool -d root,size=3584MiB --vm
> lxc config device set v1 root size=4GiB
> lxc stop -f v1
> lxc start v1
> ```

OK thanks. And I can see the mount ref is increased in this function so not forcing activation will be fine. Thanks

tomponline (approved):

mihalicyn:
> Will this cause any impact to the mount reference counter itself (I.e cause it to not be incremented?).

No, it still will be incremented, as this only prevents an additional and unnecessary attempt to map the volume twice.
Mount refcounter will be increased (in the same function `mountVolume()`.

mihalicyn:
In general, I have a feeling that we need to introduce a refcounter for volume maps with the same semantics as for volume mounts. Obviously, it will be specific only to a remote filesystems like ceph, etc. But it looks like a very useful thing.

tomponline:
> In general, I have a feeling that we need to introduce a refcounter for volume maps with the same semantics as for volume mounts. Obviously, it will be specific only to a remote filesystems like ceph, etc. But it looks like a very useful thing.

Ideally we shouldn't need it as they should 1:1 match to "mount" ref counts. So I would want us to think twice about that and check if there's an alternative (or whether there's s genuine scenario where its legitimate to map without incrementing the "mount counter").

mihalicyn:
> > 
> 
> Ideally we shouldn't need it as they should 1:1 match to "mount" ref counts. So I would want us to think twice about that and check if there's an alternative (or whether there's s genuine scenario where its legitimate to map without incrementing the "mount counter").

In `SetVolumeQuota` we may need only a block device to be present, without mount.

mihalicyn:
Hmm, I see that we have already introduced something similar to what I'm talking about in LVM https://github.com/canonical/lxd/commit/9ff322c248f21b77a33f1866f1bcf0efd367a124

tomponline:
> Hmm, I see that we have already introduced something similar to what I'm talking about in LVM [9ff322c](https://github.com/canonical/lxd/commit/9ff322c248f21b77a33f1866f1bcf0efd367a124)

Yeah this was for some lvm specific behaviour related to snapshot parents iirc. Not sure if relevant to hpe.

tomponline:
> > > 
> > 
> > 
> > Ideally we shouldn't need it as they should 1:1 match to "mount" ref counts. So I would want us to think twice about that and check if there's an alternative (or whether there's s genuine scenario where its legitimate to map without incrementing the "mount counter").
> 
> In `SetVolumeQuota` we may need only a block device to be present, without mount.

Do we even need block device for pf, pure and hpe for resize?

mihalicyn:
> Do we even need block device for pf, pure and hpe for resize?

of course,
https://github.com/canonical/lxd/blob/3ed0d834c245e0b84e8e9ad910a7b20d3b1a45a1/lxd/storage/drivers/driver_pure_volumes.go#L897
https://github.com/canonical/lxd/blob/3ed0d834c245e0b84e8e9ad910a7b20d3b1a45a1/lxd/storage/drivers/driver_pure_volumes.go#L851
https://github.com/canonical/lxd/blob/3ed0d834c245e0b84e8e9ad910a7b20d3b1a45a1/lxd/storage/drivers/driver_pure_volumes.go#L826

PR #16342, opened 2025-08-28
Minor CI tweaks for speed

simondeziel:


simondeziel:
A boring PR (to write and review, sorry) with mostly mechanical changes. It shaves ~20s to the standalone ceph case.

PR #16344, opened 2025-08-29
storage/pure: Ensure ISO image volume size is at least 1MiB

MusicDin:
When creating a new volume from an ISO image, ensure that the volume size is at least minimum allow volume size for PureStorage volume (`1MiB`).


tomponline (commented):
Why is this needed only for isos?

MusicDin:
> Why is this needed only for isos?

That was my question in https://github.com/canonical/lxd/pull/16345 -> should we always round to at least 1 MiB. 

Closing this PR in favor of https://github.com/canonical/lxd/pull/16345

PR #16345, opened 2025-08-29
storage/pure: Round volume size to make it divisible by 512B

MusicDin:
PureStorage driver rejects sizes that are not divisible by `512B`.
For example creating a volume with volume size `10MB` is rejected.

Therefore, we ensure that the size is multiple of 512 by rounding it up if necessary.
Additionally, we ensure that the volume size is `1MiB` (minimum PureStorage volume size) if smaller size is provided.

---

- [x] One question that I have is whether we should also always make sure the size is 1MiB?
For example, if user enters volume size `10KB`, should we round it to `1MiB` (minimum allowed size), or error out as we currently do?

### DIFF
--- a/lxd/storage/drivers/driver_pure.go
+++ b/lxd/storage/drivers/driver_pure.go
@@ -28,6 +28,9 @@ var pureSupportedConnectors = []string{
 	connectors.TypeNVME,
 }
 
+// pureMinVolumeSizeBytes defines the minimum size of a Pure Storage volume, which is 1MiB.
+const pureMinVolumeSizeBytes = 1024 * 1024
+
 type pure struct {
 	common
 
@@ -374,4 +377,15 @@ func (d *pure) MigrationTypes(contentType ContentType, refresh bool, copySnapsho
 			Features: rsyncFeatures,
 		},
 	}
+}
+
+// roundVolumeBlockSizeBytes rounds the given size (in bytes) up to the next
+// multiple of 512 bytes, which is the minimum allocation unit on Pure Storage.
+// It also enforces a minimum volume size of 1 MiB.
+func (d *pure) roundVolumeBlockSizeBytes(_ Volume, sizeBytes int64) int64 {
+	if sizeBytes < pureMinVolumeSizeBytes {
+		return pureMinVolumeSizeBytes
+	}
+
+	return roundAbove(512, sizeBytes)
 }
--- a/lxd/storage/drivers/driver_pure_volumes.go
+++ b/lxd/storage/drivers/driver_pure_volumes.go
@@ -48,7 +48,7 @@ func (d *pure) commonVolumeRules() map[string]func(value string) error {
 		//  type: string
 		//  defaultdesc: same as `volume.size`
 		//  shortdesc: Size/quota of the storage volume
-		"size": validate.Optional(validate.IsMultipleOfUnit("512B")),
+		"size": validate.Optional(validate.IsSize),
 	}
 }
 
--- a/lxd/storage/drivers/driver_pure_volumes.go
+++ b/lxd/storage/drivers/driver_pure_volumes.go
@@ -69,6 +69,8 @@ func (d *pure) CreateVolume(vol Volume, filler *VolumeFiller, op *operations.Ope
 		return err
 	}
 
+	sizeBytes = d.roundVolumeBlockSizeBytes(vol, sizeBytes)
+
 	// Create the volume.
 	err = client.createVolume(vol.pool, volName, sizeBytes)
 	if err != nil {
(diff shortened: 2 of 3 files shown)


simondeziel:
> One question that I have is whether we should also always make sure the size is 1MiB? For example, if user enters volume size `10KB`, should we round it to `1MiB` (minimum allowed size), or error out as we currently do?

Shouldn't this rounding be done by implementing `roundVolumeBlockSizeBytes()` like some other drivers do? As for failing vs silently rounding I'd go with the later like other drivers do.

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -786,6 +784,8 @@ func (d *pure) SetVolumeQuota(vol Volume, size string, allowUnsafeResize bool, o
 		return nil
 	}
 
+	sizeBytes = d.roundVolumeBlockSizeBytes(vol, sizeBytes)
```
roosterfish, line 787: This line made me thinking a little bit. In PowerFlex we don't implicitly round as this has a rather "big" impact due to the required multiples of 8GiB. The rounding is only done when e.g. importing an ISO so we can pick the right vol size.

```bash
root@julian:~# lxc storage volume create pf foo
Storage volume foo created
root@julian:~# lxc storage volume set pf foo size=10GiB
Error: Invalid value for volume "default_foo" option "size": Value not in multiples of 8GiB: 10GiB
root@julian:~# lxc storage volume create pf foo2 size=12GiB
Error: Invalid value for volume "foo2" option "size": Value not in multiples of 8GiB: 12GiB
root@julian:~# lxc storage volume create pf foo2 size=16GiB
Storage volume foo2 created
```
But in this case its multiples of 512B which I think is acceptable. 
However someone might ask why the PowerFlex driver doesn't provide you with a 16 GiB vol if you try to create one using 10 GiB. At this point we could also argue should we return one using 8GiB as this one is nearer to the original value? 
What I am asking myself now is where do we draw the line between 1) rounding is acceptable and done implicitly and 2) rounding is too expensive so we error out.

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -48,7 +48,7 @@ func (d *pure) commonVolumeRules() map[string]func(value string) error {
 		//  type: string
 		//  defaultdesc: same as `volume.size`
 		//  shortdesc: Size/quota of the storage volume
-		"size": validate.Optional(validate.IsMultipleOfUnit("512B")),
+		"size": validate.Optional(validate.IsSize),
```
roosterfish, line 51: Should this be documented in the Pure driver docs? Meaning if you don't provide a multiple of 512B vol size, you are getting a rounded one instead?

lxd/storage/drivers/driver_pure.go
```diff
@@ -375,3 +378,14 @@ func (d *pure) MigrationTypes(contentType ContentType, refresh bool, copySnapsho
 		},
 	}
 }
+
+// roundVolumeBlockSizeBytes rounds the given size (in bytes) up to the next
+// multiple of 512 bytes, which is the minimum allocation unit on Pure Storage.
+// It also enforces a minimum volume size of 1 MiB.
+func (d *pure) roundVolumeBlockSizeBytes(vol Volume, sizeBytes int64) int64 {
```
roosterfish: A minor tweak, you can do `_ Volume`.

roosterfish (changes requested):

lxd/storage/drivers/driver_pure_volumes.go
MusicDin: This was my exact thought.

Rounding to nearest 512B seems acceptable to me because the size is so small that it practically has no impact. However, rounding to the minimum volume size of `1MiB`, while it feels small, I share your opinion that there should be a clear rule/approach.

IMO, we should just keep the approach consistent across all drivers. Either we always error out if the size conditions are not matched or we round up the size, where round up means:
If size < min allowed volume size -> Bump it to min allowed size.
If size % allocation unit != 0 -> round up to next multiple of AU.

roosterfish: > If size < min allowed volume size -> Bump it to min allowed size.

This!

> If size % allocation unit != 0 -> round up to next multiple of AU

I feel this is too much in case of PowerFlex. At the end it would mean you request a 8193MiB vol and get a 16384MiB which is almost double the size.

How likely is it to hit this limitation in Pure? As its multiple of 512B, I would assume you almost never really encounter this as I assume you usually provide rounded vol sizes like 1GiB or 10GiB/100GiB/500GiB which match the `% 512B == 0` check.

MusicDin: > How likely is it to hit this limitation in Pure? As its multiple of 512B, I would assume you almost never really encounter this as I assume you usually provide rounded vol sizes like 1GiB or 10GiB/100GiB/500GiB which match the % 512B == 0 check.

Should be only if you provide the size in KB/MB/GB instead of KiB/MiB/GiB, or directly in bytes.

lxd/storage/drivers/driver_pure_volumes.go
MusicDin: I've added explanation next to the volume size constraints - please check

lxd/storage/drivers/driver_pure.go
MusicDin: Fixed

lxd/storage/drivers/driver_pure_volumes.go
tomponline: Powerflex is a special case in my view as its rounding amount is "too much" (8GB) so I think the pf driver should continue to require multiples of 8GB from user but other drivers can round a few kB.

PR #16352, opened 2025-09-01
lxd/storage/drivers/{ceph,zfs,lvm,generic}: unmap block device for ISO volumes

mihalicyn:
We shouldn't forget to unmap a block device when unmounting ISO volumes.

### DIFF
--- a/lxd/storage/drivers/driver_ceph_volumes.go
+++ b/lxd/storage/drivers/driver_ceph_volumes.go
@@ -1631,7 +1631,7 @@ func (d *ceph) UnmountVolume(vol Volume, keepBlockDev bool, op *operations.Opera
 		}
 
 		ourUnmount = true
-	} else if vol.contentType == ContentTypeBlock {
+	} else if IsContentBlock(vol.contentType) {
 		// For VMs, unmount the filesystem volume.
 		if vol.IsVMBlock() {
 			fsVol := vol.NewVMBlockFilesystemVolume()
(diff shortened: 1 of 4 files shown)


lxd/storage/drivers/driver_ceph_volumes.go
```diff
@@ -1631,7 +1631,7 @@ func (d *ceph) UnmountVolume(vol Volume, keepBlockDev bool, op *operations.Opera
 		}
 
 		ourUnmount = true
-	} else if vol.contentType == ContentTypeBlock {
+	} else if IsContentBlock(vol.contentType) {
```
roosterfish, line 1634: Nice catch!

roosterfish (approved):

lxd/storage/drivers/driver_ceph_volumes.go
roosterfish: > I'm pretty sure that the same can be reproduced on Powerflex/Pure.

PowerFlex is using the generic driver's mount/unmount so its fixed by this PR too.

mihalicyn: > PowerFlex is using the generic driver's mount/unmount so its fixed by this PR too.

sure, I meant that without this fix it should be reproducible :)

tomponline: Good catch !

tomponline:
@mihalicyn @MusicDin please can you back port this fix and cherry pick into latest-candidate and 5.21-candidate

Thanks!

mihalicyn:
> @mihalicyn @MusicDin please can you back port this fix and cherry pick into latest-candidate and 5.21-candidate
> 
> Thanks!

I'm on it.

PR #16355, opened 2025-09-01
Auth: DevLXD token auth follow-up

markylaing:


lxd/request/requestor.go
```diff
@@ -198,7 +198,7 @@ func (r *Requestor) setForwardingDetails(req *http.Request) error {
 }
 
 // setIdentity validates and sets the [identity.CacheEntry] in the Requestor.
-// It must be called after setForwardingDetails.
+// It must only be called when Requestor.trusted is true, and after setForwardingDetails has been called.
```
kadinsayani, line 201: Can we add early checks for this?

```Go
if !r.trusted {
	return errors.New("Cannot set identity for untrusted request")
}

if r.forwardedOriginAddress == "" || r.forwardedUsername == "" || r.forwardedProtocol == "" || r.forwardedIdentityProviderGroups == nil {
    return errors.New("Cannot set identity without forwarded request information")
}
```

markylaing: We could do.. this function is not exported and is only called by `SetRequestor`, the main reason for separating it out was to simplify that function, where those checks are already performed.

Also, the forwarded request information may well be empty, that's not an error. It's just that `setIdentity` should be called after the request headers are examined to ensure the `callerUsername` and `callerProtocol` are correct.

kadinsayani: Gotcha, thanks. The note about `setIdentity()` only being callable under certain conditions feels like a code smell, since the function itself doesn’t enforce those conditions. That said, because it’s unexported and only invoked by `SetRequestor()`, I think it’s fine to avoid duplicating the checks.

PR #16358, opened 2025-09-02
Avoid explicit type declarations for global vars

simondeziel:
Inspired by Mark's PR #16351


lxd/storage/connectors/connector_iscsi.go
```diff
@@ -26,7 +26,7 @@ const (
 	iscsiErrCodeNotFound = 21
 )
 
-var _ Connector = &connectorISCSI{}
+var _ = &connectorISCSI{}
```
markylaing: Beware of this one - it may be important.

I expect that this is an inline assertion that enforces that `&connectorISCSI{}` implements a `Connector` interface, preventing compilation if not. I've had edge cases where this is required in the past, e.g. where it is not being explicitly returned as that interface type. @roosterfish may know more.

lxd/storage/connectors/connector_nvme.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/canonical/lxd/shared/revert"
 )
 
-var _ Connector = &connectorNVMe{}
+var _ = &connectorNVMe{}
```
markylaing: Same here as for the ISCSI one

lxd/storage/connectors/connector_sdc.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/canonical/lxd/shared/revert"
 )
 
-var _ Connector = &connectorSDC{}
+var _ = &connectorSDC{}
```
markylaing: And again here

lxd/storage/connectors/connector_iscsi.go
simondeziel: Alright, that echoes chat Copilot only picked on the `starlak_test.go` file despite me asking for 2 reviews because those you flagged seems utterly similar.

Thanks, I'll keep those `Connector` types in.

PR #16359, opened 2025-09-02
Auth: Fix permission management for entitlements on bearer identities

markylaing:
Earlier I was demonstrating the permissions API for the anbox team. I noticed that running `lxc auth permission list` returned a bogus URL for a `DevLXD token bearer` identity. The URL looked like:
```
/1.0/auth/identities//{uuid}
```
e.g. the authentication method was missing. 

I then attempted to grant a group the `can_edit` entitlement on the identity and found it to be broken, with all combinations of `lxc auth group permission add <group> identity {bearer,devlxd}/{name,UUID} can_edit` failing.

This PR does a few things:
1. Updates `lxd/db/cluster.AuthMethod` to use a map of database code to text when scanning from the db.
2. Uses the above map to generate a case statement when querying for entity URLs via the permissions APIs.
3. Updates `lxc auth group permission add/remove` to parse the shortform identity type when the entity type that the entitlement is being grant against is an identity. This allows e.g. `lxc auth group permission add <group> identity devlxd/{uuid} can_edit`.

### DIFF
--- a/lxc/auth.go
+++ b/lxc/auth.go
@@ -649,7 +649,7 @@ func parsePermissionArgs(args []string) (*api.Permission, error) {
 
 	if entityType == entity.TypeServer {
 		if len(args) != 3 {
-			return nil, errors.New("Expected three arguments: `lxc auth group grant [<remote>:]<group> server <entitlement>`")
+			return nil, errors.New("Expected three arguments: `lxc auth group permission add [<remote>:]<group> server <entitlement>`")
 		}
 
 		return &api.Permission{
@@ -660,7 +660,7 @@ func parsePermissionArgs(args []string) (*api.Permission, error) {
 	}
 
 	if len(args) < 4 {
-		return nil, errors.New("Expected at least four arguments: `lxc auth group grant [<remote>:]<group> <object_type> <object_name> <entitlement> [<key>=<value>...]`")
+		return nil, errors.New("Expected at least four arguments: `lxc auth group permission add [<remote>:]<group> <object_type> <object_name> <entitlement> [<key>=<value>...]`")
 	}
 
 	entityName := args[2]
--- a/lxc/auth.go
+++ b/lxc/auth.go
@@ -680,9 +680,9 @@ func parsePermissionArgs(args []string) (*api.Permission, error) {
 
 	pathArgs := []string{entityName}
 	if entityType == entity.TypeIdentity {
-		authenticationMethod, identifier, ok := strings.Cut(entityName, "/")
-		if !ok {
-			return nil, fmt.Errorf("Malformed identity argument, expected `<type>/<identifier>`, got %q", entityName)
+		authenticationMethod, _, identifier, err := resolveIdentityTypeShorthand(entityName)
+		if err != nil {
+			return nil, err
 		}
 
 		pathArgs = []string{authenticationMethod, identifier}
--- a/lxc/auth.go
+++ b/lxc/auth.go
@@ -728,29 +728,42 @@ type cmdIdentity struct {
 	global *cmdGlobal
 }
 
-// resolveIdentityTypeShorthand takes a shorthand and returns an authentication method and identity type or an error.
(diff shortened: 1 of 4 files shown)


lxc/auth.go
```diff
@@ -728,29 +728,42 @@ type cmdIdentity struct {
...
+		return api.AuthenticationMethodOIDC, api.IdentityTypeOIDCClient, idName, nil
 	case "devlxd":
-		return remoteName, api.AuthenticationMethodBearer, api.IdentityTypeBearerTokenDevLXD, idName, nil
+		return api.AuthenticationMethodBearer, api.IdentityTypeBearerTokenDevLXD, idName, nil
+	}
+
+	return "", "", "", fmt.Errorf("Unrecognized identity type shorthand %q", shorthandType)
+}
+
+// resolveIdentityTypeShorthand takes an identity argument of the form [<remote>:]<type>/<name> and returns the remote
```
kadinsayani, line 752:
```suggestion
// resolveIdentityArg...
```

lxc/auth.go
```diff
@@ -728,29 +728,42 @@ type cmdIdentity struct {
...
-	remoteName, resourceName, err := c.global.conf.ParseRemote(identityArg)
-	if err != nil {
-		return "", "", "", "", err
-	}
-
-	shorthandType, idName, ok := strings.Cut(resourceName, "/")
+// resolveIdentityTypeShorthand takes an identity argument of the form <type>/<name> and returns an authentication
+// method, an identity type, and a name (or an error).
+// If the shorthand <type> resolves to more than one identity type, it returns an empty string for the identity type.
+func resolveIdentityTypeShorthand(identityArg string) (method string, identityType string, nameOrID string, err error) {
```
kadinsayani, line 734: Perhaps this function would be a good candidate for unit testing?

lxd/db/cluster/identities.go
```diff
@@ -68,19 +68,21 @@ const (
...
-	case authMethodTLS:
-		*a = api.AuthenticationMethodTLS
-	case authMethodOIDC:
-		*a = api.AuthenticationMethodOIDC
-	case authMethodBearer:
-		*a = api.AuthenticationMethodBearer
-	default:
+	text, ok := authMethodCodeToText[authMethodCode]
+	if !ok {
 		return fmt.Errorf("Unknown authentication method `%d`", authMethodCode)
```
kadinsayani, line 82:
```suggestion
		return fmt.Errorf("Unknown authentication method code `%d`", authMethodCode)
```

kadinsayani (changes requested):
Just a few nits, thanks!

lxc/auth.go
markylaing: I'm not sure it's really necessary as this will be covered by the main suite but I will if you insist :wink:

lxd/db/cluster/identities.go
markylaing: Fixed

lxc/auth.go
markylaing: Fixed

lxc/auth.go
kadinsayani: It might help TICS. Your call :wink:

markylaing: I thought we'd figured out TICS so that it gets coverage from our main suite? (as well as unit tests)

kadinsayani: Yes, it does. To get full coverage points, we just want to make sure all statements in `resolveIdentityTypeShorthand()` are covered in the main test suite via CLI commands. https://github.com/canonical/lxd/blob/314e78c7ff6e54e82b3156e8e2d5ae42869a8530/test/suites/auth.sh#L198-L199 covers us for bearer identities. Do we also have similar checks for invalid/valid shorthand expressions for TLS and OIDC identities?

markylaing: Yep we also have checks for TLS and OIDC identities :)

kadinsayani (approved):
LGTM, thanks!

tomponline (approved):
Ta

PR #16367, opened 2025-09-03
Reduce dependency on `gorilla/mux` (from Incus)

simondeziel:
The long term plan would be to completely replace `gorilla/mux` and use Go stdlib `net/http` instead.


lxd-agent/operations.go
```diff
@@ -43,9 +41,9 @@ var operationWait = APIEndpoint{
 }
 
 func operationDelete(d *Daemon, r *http.Request) response.Response {
-	id, err := url.PathUnescape(mux.Vars(r)["id"])
-	if err != nil {
-		return response.SmartError(err)
+	id := r.PathValue("id")
```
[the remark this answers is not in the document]
kadinsayani: https://github.com/canonical/lxd/blob/dc94228a8d4fd8ad61561be0ad60a39bef06caf6/lxd-agent/operations.go#L21-L26

lxd-agent/devlxd.go
```diff
@@ -164,8 +162,8 @@ var devLXDConfigKeyEndpoint = devLXDAPIEndpoint{
 }
 
 func devLXDConfigKeyGetHandler(d *Daemon, r *http.Request) *devLXDResponse {
-	key, err := url.PathUnescape(mux.Vars(r)["key"])
-	if err != nil {
+	key := r.PathValue("key")
```
tomponline: @simondeziel does r.PathValue do the unescaping?

lxd-agent/devlxd.go
```diff
@@ -415,13 +412,7 @@ func registerDevLXDEndpoint(d *Daemon, apiRouter *mux.Router, apiVersion string,
 		}
 	}
 
-	route := apiRouter.HandleFunc(uri, handleFunc)
```
MusicDin, line 429: Why is this being dropped?

markylaing (commented):
Overall this is fine as a drop in replacement for `gorilla/mux`. As Tom says, we just need to ensure that `r.PathValue("...")` is performing the path unescaping.

Something to also consider is that the "pattern" argument for the stdlib http handlers accepts a method and a path e.g. `PUT /1.0/instances`. Utilising this might make the `createCmd` functions a bit simpler, but might also cause "method not supported" errors to become not found errors or similar.

tomponline:
Switching back to draft until comments are addressed.

lxd-agent/devlxd.go
MusicDin: DevLXD is not using the named routes, so I am OK with this being dropped.
However, main LXD relies on named routes to differentiate between `/containers` and `/instances` endpoints. To keep that backward compatibility, we could save the route name in the request context, which is what `gorilla/mux` is actually doing behind the scene. When this lands in LXD, I would prefer seeing that in DevLXD as well, just to make sure they are consistent.

@simondeziel If others agree on this approach, I can help you land this in a separate PR.

lxd-agent/devlxd.go
simondeziel: Seems like it does but Incus doesn't do it but it sounds like we better keep it. Let me fix.

lxd-agent/devlxd.go
simondeziel: As discussed in our 1:1, I appreciate the solution you showed and agree we'll likely need it when moving forward with the rest of LXD. That's material for a separated PR IMHO.

simondeziel:
> Something to also consider is that the "pattern" argument for the stdlib http handlers accepts a method and a path e.g. `PUT /1.0/instances`. Utilising this might make the `createCmd` functions a bit simpler, but might also cause "method not supported" errors to become not found errors or similar.

Yeah, I think that's something to explore but probably in a separated PR as this one is mostly to get the cherry-pick.

simondeziel:
> Switching back to draft until comments are addressed.

Done, PTAL.

lxd-agent/devlxd.go
tomponline: How are we recording this so it happens? Is there a GH issue with discussion points for switching to the built in mux?

tomponline:
> Overall this is fine as a drop in replacement for `gorilla/mux`. As Tom says, we just need to ensure that `r.PathValue("...")` is performing the path unescaping.
> 
> Something to also consider is that the "pattern" argument for the stdlib http handlers accepts a method and a path e.g. `PUT /1.0/instances`. Utilising this might make the `createCmd` functions a bit simpler, but might also cause "method not supported" errors to become not found errors or similar.

@simondeziel did you get an answer for these 2? Especially the first one as its important.

simondeziel:
> > Overall this is fine as a drop in replacement for `gorilla/mux`. As Tom says, we just need to ensure that `r.PathValue("...")` is performing the path unescaping.

This was addressed.

> > Something to also consider is that the "pattern" argument for the stdlib http handlers accepts a method and a path e.g. `PUT /1.0/instances`. Utilising this might make the `createCmd` functions a bit simpler, but might also cause "method not supported" errors to become not found errors or similar.

This is left for a later PR.

tomponline:
> This was addressed.
> 
> > >

was it a response or a code change? I couldnt see it.

tomponline:
> This is left for a later PR.

so no API behaviour change now?

simondeziel:
> > > Overall this is fine as a drop in replacement for `gorilla/mux`. As Tom says, we just need to ensure that `r.PathValue("...")` is performing the path unescaping.
> 
> This was addressed.

Sorry, that's not true. Something went wrong with one of my rebase so the changes were dropped. Will fix.

lxd-agent/devlxd.go
tomponline: @simondeziel please can you create a GH issue for the 26.04 milestone to drop gorilla mux and move to the built in one - i dont want us to stay in a half way situation for an entire LTS release.

simondeziel:
@tomponline the `url.PathUnescape()` calls are back in as they are needed.

@MusicDin suggested to not do the checks for emptiness as that might introduce a subtle behaviour change on error and I agree so I dropped this part of the cherry-pick. This means, we stay aligned with the old behaviour so it should be safe.

Mark's suggestion about the HTTP method was not explored in this PR so there is no API behaviour change.

lxd-agent/devlxd.go
simondeziel: https://github.com/canonical/lxd/issues/16587

lxd-agent/devlxd.go
```diff
@@ -265,9 +264,9 @@ var devLXDImageExportEndpoint = devLXDAPIEndpoint{
 
 func devLXDImageExportHandler(d *Daemon, r *http.Request) *devLXDResponse {
 	// Extract the fingerprint.
-	fingerprint, err := url.PathUnescape(mux.Vars(r)["fingerprint"])
+	fingerprint, err := url.PathUnescape(r.PathValue("fingerprint"))
 	if err != nil {
-		return smartResponse(err)
+		return smartResponse(errors.New("Missing image fingerprint"))
```
tomponline: It might not be missing, it might be invalid, also we've lost the actual error now, suggest:

```go
return smartResponse(api.StatusErrorf(http.StatusBadRequest, "Failed parsing image fingerprint: %w", err))
```

Also there's several other places in the code base doing `url.PathUnescape(mux.Vars(r)["fingerprint"])` so suggest you also apply the same change there for consistency.

lxd-agent/operations.go
```diff
@@ -43,9 +40,9 @@ var operationWait = APIEndpoint{
...
```
tomponline: We've lost the `url.PathUnescape` again here.

lxd-agent/devlxd.go
tomponline: @simondeziel id suggest leaving the return error changes out of this PR and then doing a separate pass of these endpoints error messages and codes, because there's places like `return errorResponse(http.StatusBadRequest, "bad request")` above that could also do with some attention.

lxd-agent/operations.go
```diff
@@ -63,9 +60,9 @@ func operationDelete(d *Daemon, r *http.Request) response.Response {
 }
 
 func operationGet(d *Daemon, r *http.Request) response.Response {
-	id, err := url.PathUnescape(mux.Vars(r)["id"])
-	if err != nil {
-		return response.SmartError(err)
+	id := r.PathValue("id")
```
tomponline: lost `url.PathUnescape`

lxd-agent/operations.go
```diff
@@ -152,9 +149,9 @@ func operationsGet(d *Daemon, r *http.Request) response.Response {
 }
 
 func operationWebsocketGet(d *Daemon, r *http.Request) response.Response {
-	id, err := url.PathUnescape(mux.Vars(r)["id"])
-	if err != nil {
-		return response.SmartError(err)
+	id := r.PathValue("id")
```
tomponline: lost `url.PathUnescape`

lxd-agent/operations.go
tomponline, line 1: Can we include these changes in the original commit please.

simondeziel:
> Lets drop the changes to error responses,

Done.

> and include the path escaping in the original commit so as to not change something and then change it back in the next commit.

Ack, I've squashed them together. I didn't do it initially due to wanting as close to original cherry-picks but I'm fine with squashing.

PTAL.

lxd-agent/operations.go
```diff
@@ -43,7 +41,7 @@ var operationWait = APIEndpoint{
 }
 
 func operationDelete(d *Daemon, r *http.Request) response.Response {
-	id, err := url.PathUnescape(mux.Vars(r)["id"])
+	id, err := url.PathUnescape(r.PathValue("id"))
```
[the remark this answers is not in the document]
simondeziel: Sigh, rather than speculating on the Go version, you should take a look at the `go.mod` file.

lxd-agent/server.go
```diff
@@ -101,12 +97,6 @@ func createCmd(restAPI *mux.Router, version string, c APIEndpoint, cert *x509.Ce
 			}
 		}
 	})
-
-	// If the endpoint has a canonical name then record it so it can be used to build URLS
```
tomponline, line 105: Is this not needed anymore?

lxd-agent/operations.go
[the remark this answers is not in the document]
kadinsayani: This would be another good candidate for copilot instructions: https://github.com/canonical/lxd/pull/16596.

lxd-agent/server.go
simondeziel: It was dropped by the cherry-picked commit and I hadn't look into it much. I'll have to defer to those who better understand the code I'm afraid.

tomponline: Lets move this back to draft until we get chance to look at it properly.

MusicDin: @tomponline If you agree with comment https://github.com/canonical/lxd/pull/16367#discussion_r2333933808, this can be merged and we'll prepare a PR for named routes and replace `gorilla/mux` on devLXD side as well.

simondeziel: @tomponline what do you think of Din's suggestion?

tomponline: yep sounds fine to me

simondeziel: @tomponline it's been a while this PR went through CI, would you like me to rebase it?

tomponline: yes please

simondeziel:
@tomponline CI green now.

PR #16369, opened 2025-09-03
github: also run `golangci/golangci-lint-action` on schedule

simondeziel:
This will ensure the `main` and `stable-*` branches have warm caches which should then speed up the check done in PRs.


tomponline:
ready for rebase

PR #16375, opened 2025-09-04
lxd/api_project: Fix daemon storage unmount on project deletion

skozina:
This is serious follow-up bug fix on #15990, and should be backported with #15990.

The issue here is that on project deletion, if there is no project-specific daemon storage set (`storage.project.foo.images_volume` and the like), we unmount the daemon storage itself (normally set with `storage.images_volume`). This makes the `images` directory to disappear (because it was on the volume we just unmounted) causing all following image operations to fail.

### DIFF
--- a/test/suites/serverconfig.sh
+++ b/test/suites/serverconfig.sh
@@ -106,6 +106,11 @@ _server_config_storage() {
   lxc config set storage.backups_volume "${pool}/backups"
   lxc config set storage.images_volume "${pool}/images"
 
+  # Regression test for the case where project deletion unmounted the daemon storage volume.
+  lxc project create foo
+  lxc project delete foo
+  [ -d "${LXD_DIR}/storage-pools/${pool}/custom/default_images/images" ]
+
   # Validate the old location is really gone
   [ ! -e "${LXD_DIR}/backups" ]
   [ ! -e "${LXD_DIR}/images" ]


lxd/api_project.go
```diff
@@ -1032,7 +1032,9 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
 		}
 
 		// Unmount the project-specific storage volumes.
-		for _, projectStorageVolume := range []string{config.StorageImagesVolume(name), config.StorageBackupsVolume(name)} {
+		localConfig := s.LocalConfig.Dump()
```
tomponline, line 1035: @skozina what is the difference between `node.ConfigLoad` above (in the 1config` var) and `s.LocalConfig.Dump()` in the `localConfig` var? They both seem to be containing local node config.

tomponline:
@skozina ready for rebase

lxd/api_project.go
skozina: They do. The problem is that the raw config values are private to the `node` package. The private values can be accessed with the `StorageImagesVolume()` public function, but it tries to be smart -- if the project config is not set, it returns the daemon storage, if configured. Look at `daemonStorageVolume`. And that's the whole cause of the bug here.
So to fix this, we need to real the real raw values, which can only be access in the raw `Dump()`.

tomponline:
Thanks!

simondeziel:
@escabo sounds like this could fix the bug you found with while doing e2e testing on MicroCloud. But maybe you know already and reached out to Stan directly?

tomponline:
> @escabo sounds like this could fix the bug you found with while doing e2e testing on MicroCloud. But maybe you know already and reached out to Stan directly?

https://chat.canonical.com/canonical/pl/u6dum6xqqty98nnnkjagxs5yma

tomponline:
Ready for cherry-pick

PR #16396, opened 2025-09-05
Extract doNetworkDelete() from network DELETE handler

kadinsayani:


lxd/networks.go
```diff
@@ -1140,74 +1140,80 @@ func networkDelete(d *Daemon, r *http.Request) response.Response {
...
-			return response.BadRequest(errors.New("The network is currently in use"))
+			return api.StatusErrorf(http.StatusBadRequest, "The network is currently in use")
 		}
 	}
 
 	if n.LocalStatus() != api.NetworkStatusPending {
 		err = n.Delete(clientType)
 		if err != nil {
-			return response.InternalError(err)
+			return api.StatusErrorf(http.StatusInternalServerError, "Failed to delete network: %w", err)
```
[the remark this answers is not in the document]
kadinsayani: https://github.com/canonical/lxd/blob/d250a89aa9b792ae2b36bc99da100d71c35079da/shared/api/error.go#L30-L36

lxd/networks.go
```diff
@@ -1140,74 +1140,80 @@ func networkDelete(d *Daemon, r *http.Request) response.Response {
...
+
+// doNetworkDelete deletes the named network in the given project.
+// Returns an [api.StatusError] with an HTTP status code on failure.
+func doNetworkDelete(ctx context.Context, s *state.State, name string, effectiveProjectName string, requestProjectConfig map[string]string, clusterNotification bool, clientType clusterRequest.ClientType) error {
 	// Get the existing network.
-	n, err := network.LoadByName(s, effectiveProjectName, details.networkName)
+	n, err := network.LoadByName(s, effectiveProjectName, name)
 	if err != nil {
-		return response.SmartError(fmt.Errorf("Failed loading network: %w", err))
+		return api.StatusErrorf(http.StatusBadRequest, "Failed loading network: %w", err)
```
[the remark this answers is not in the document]
kadinsayani: https://github.com/canonical/lxd/blob/d250a89aa9b792ae2b36bc99da100d71c35079da/shared/api/error.go#L30-L36

lxd/networks.go
```diff
@@ -1140,74 +1140,80 @@ func networkDelete(d *Daemon, r *http.Request) response.Response {
...
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	return response.EmptySyncResponse
+}
+
+// doNetworkDelete deletes the named network in the given project.
+// Returns an [api.StatusError] with an HTTP status code on failure.
+func doNetworkDelete(ctx context.Context, s *state.State, name string, effectiveProjectName string, requestProjectConfig map[string]string, clusterNotification bool, clientType clusterRequest.ClientType) error {
```
tomponline: You don't need to return an api.StatusError unless you're specifically wanting to return an http response code.

Otherwise you should just return the error and then use response.SmartError in the call site so it detects the error type and uses right response code.

lxd/networks.go
tomponline: Don't force bad request here, it might he an internal dB error, you don't know here, just return error like before.

lxd/networks.go
```diff
@@ -1140,74 +1140,80 @@ func networkDelete(d *Daemon, r *http.Request) response.Response {
...
 	if err != nil {
-		return response.SmartError(fmt.Errorf("Failed loading network: %w", err))
+		return api.StatusErrorf(http.StatusBadRequest, "Failed loading network: %w", err)
 	}
 
 	// Check if project allows access to network.
-	if !project.NetworkAllowed(details.requestProject.Config, details.networkName, n.IsManaged()) {
-		return response.SmartError(api.StatusErrorf(http.StatusNotFound, "Network not found"))
+	if !project.NetworkAllowed(requestProjectConfig, name, n.IsManaged()) {
+		return api.StatusErrorf(http.StatusNotFound, "Network not found")
```
tomponline: This change makes sense to return api.StatusError

tomponline: But you can use NewStatusError as no formatting needed

tomponline (changes requested):

lxd/networks.go
```diff
@@ -1140,74 +1140,80 @@ func networkDelete(d *Daemon, r *http.Request) response.Response {
...
 		// Quick checks.
 		inUse, err := n.IsUsed()
 		if err != nil {
-			return response.SmartError(err)
+			return err
 		}
 
 		if inUse {
-			return response.BadRequest(errors.New("The network is currently in use"))
+			return api.StatusErrorf(http.StatusBadRequest, "The network is currently in use")
```
tomponline: NewStatusError

lxd/networks.go
tomponline: Just wrap the error here no need for status error

lxd/networks.go
kadinsayani: Thanks for clarifying how `api.StatusError` should be used. I also looked closer at `shared/api/error.go` to make sure I fully understand the conventions.

lxd/networks.go
kadinsayani: Got it :)

lxd/networks.go
```diff
@@ -1140,74 +1140,79 @@ func networkDelete(d *Daemon, r *http.Request) response.Response {
...
+	err = doNetworkDelete(r.Context(), s, details.networkName, effectiveProjectName, details.requestProject.Config, isClusterNotification(r), clusterRequest.UserAgentClientType(r.Header.Get("User-Agent")))
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	return response.EmptySyncResponse
+}
+
+// doNetworkDelete deletes the named network in the given project.
+func doNetworkDelete(ctx context.Context, s *state.State, name string, projectName string, requestProjectConfig map[string]string, clusterNotification bool, clientType clusterRequest.ClientType) error {
```
tomponline: This function shouldn't need both `clusterNotification bool` and `clientType clusterRequest.ClientType` as the latter covers the former.

markylaing:
> Lets simplify the client type/notification detection to use just `clusterRequest.UserAgentClientType(`

Cluster notification and client type could also be moved to `request.Requestor` to centralise this kind of logic

tomponline:
> > Lets simplify the client type/notification detection to use just `clusterRequest.UserAgentClientType(`
> 
> Cluster notification and client type could also be moved to `request.Requestor` to centralise this kind of logic

Good idea, and we should just switch to client type (one of which is notification).

kadinsayani:
I've added a `clientType` field to the `Requestor` struct, alongside a helper func `GetClientType()` which returns the client type, whether the request is a cluster notification, and any error. If the changes look good, I can open up another PR which makes use of `GetClientType()` elsewhere, and we can deprecate `isClusterNotification()`.

tomponline:
Happy with this @markylaing ?

lxd/request/request.go
```diff
@@ -22,3 +23,15 @@ func CreateRequestor(ctx context.Context) *api.EventLifecycleRequestor {
 func SaveConnectionInContext(ctx context.Context, connection net.Conn) context.Context {
 	return context.WithValue(ctx, CtxConn, connection)
 }
+
+// GetClientType returns the client type, whether the request is a cluster notification, and any error.
+func GetClientType(ctx context.Context) (clientType request.ClientType, clusterNotification bool, err error) {
```
tomponline: I dont think we need to special case the clusterNotification return value, the client type can just be compared to the `request.ClientTypeNotifier` constant.

tomponline: @markylaing do you think we even need this function?

tomponline (commented):
LGTM, we just need to confirm the approach to getting client type out of context.

lxd/request/requestor.go
```diff
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"slices"
 
+	"github.com/canonical/lxd/lxd/cluster/request"
```
markylaing: This package contains only the clienttype.go file with these definitions. Can we just move them here so we only have one `request` package?

tomponline: happy with that if it doesn't enlarge the import size of lxd-agent, but its likely already importing request package anyway

lxd/request/requestor.go
```diff
@@ -260,6 +267,7 @@ func (r *Requestor) setIdentity(cache *identity.Cache) error {
 func SetRequestor(req *http.Request, identityCache *identity.Cache, args RequestorArgs) error {
 	r := &Requestor{
 		trusted:                args.Trusted,
+		clientType:             request.UserAgentClientType(req.Header.Get("User-Agent")),
```
markylaing: IIRC these user agent headers should only be set if receiving a request from another cluster member? Shall we check here if the protocol is "cluster"?

tomponline: yes that makes sense i think and if so its normal?

lxd/request/request.go
markylaing: I think `Requestor` should have an `IsClusterNotification` function, and a `ClientType` function. Then we should evaluate if it's required to have this function.

tomponline: Can we just have ClientType() and then compare to the ClientTypeNotifier constant? Having `IsClusterNotification` seems redundant.

markylaing: It is a bit redundant. I'm thinking this should get rid of the `isClusterNotification` function, so just trying to ease that transition a little

tomponline: We can switch to ClientType() == notifier for same effect

markylaing (changes requested):
Just following up on my earlier comments - for this PR I would like to see:
- The user agent header extracted and validated as part of `request.SetRequestor`. Return an error if the protocol is not "cluster" but an internal user agent client type has been set.
- Remove the `isClusterNotification` function. This function only checks the user agent, but does not check the requestor protocol. Which IMO is a little dangerous as it relies on earlier validation being performed during authentication.
- Add utils to `Requestor` first before considering adding utils that accept a context and call GetRequestor themselves. In most cases, we need more than one thing from the Requestor, so it's often better to get it at the beginning of a handler/function and use its' utils directly.

markylaing:
> Just following up on my earlier comments - for this PR I would like to see:
> 
> * The user agent header extracted and validated as part of `request.SetRequestor`. Return an error if the protocol is not "cluster" but an internal user agent client type has been set.
> * Remove the `isClusterNotification` function. This function only checks the user agent, but does not check the requestor protocol. Which IMO is a little dangerous as it relies on earlier validation being performed during authentication.
> * Add utils to `Requestor` first before considering adding utils that accept a context and call GetRequestor themselves. In most cases, we need more than one thing from the Requestor, so it's often better to get it at the beginning of a handler/function and use its' utils directly.

@tomponline in the spirit of having easy to review PRs - this could actually go in it's own PR. I was considering doing this when I initially added the `request.Requestor`, but was trying not to let the refactoring get out of hand.

tomponline:
> > Just following up on my earlier comments - for this PR I would like to see:
> > 
> > * The user agent header extracted and validated as part of `request.SetRequestor`. Return an error if the protocol is not "cluster" but an internal user agent client type has been set.
> > * Remove the `isClusterNotification` function. This function only checks the user agent, but does not check the requestor protocol. Which IMO is a little dangerous as it relies on earlier validation being performed during authentication.
> > * Add utils to `Requestor` first before considering adding utils that accept a context and call GetRequestor themselves. In most cases, we need more than one thing from the Requestor, so it's often better to get it at the beginning of a handler/function and use its' utils directly.
> 
> @tomponline in the spirit of having easy to review PRs - this could actually go in it's own PR. I was considering doing this when I initially added the `request.Requestor`, but was trying not to let the refactoring get out of hand.

Much agree

kadinsayani:
@tomponline @markylaing https://github.com/canonical/lxd/pull/16446 :)

lxd/request/request.go
kadinsayani: Please see https://github.com/canonical/lxd/pull/16446/commits/ba22b9a60a09387deb5d871b564a28c88da86fe5 and the associated commit description. I'm okay with either approach :)

PR #16399, opened 2025-09-06
Rework CI caching strategy

simondeziel:
This changes how `minio` and `dqlite`/`liblxc` are cached as well as introduce a cache for snap dependencies.

* `minio` is now cached for up to 7 days
* `dqlite`/`liblxc` is still cached only for 1 day
* snap dependencies are cached for only 1 day

A notable change is that `minio` and `dqlite`/`liblxc` are no longer saved in separated caches and duplicated in the `system-test-deps` artifact.

The concept of artifact is kept to group the things that are build during a given CI run and the other cached assets are using the `actions/cache` mechanism alone to be saved and reused independently between the parent and derived PRs.

This change of strategy will reduce the network consumption (a non-goal but still a benefit) but will also reduce the cache space consumption and improve the cache hit slightly which should come with a little CI speedup.

### DIFF
--- a/test/includes/snap.sh
+++ b/test/includes/snap.sh
@@ -1,0 +1,90 @@
+# download_snap: downloads a snap to the cache dir.
+download_snap() {
+    local name="${1}"
+    local channel="${2:-"latest/edge"}"
+    local cache_dir="${SNAP_CACHE_DIR:-${HOME}/snap-cache}"
+    local dir="${cache_dir}/${name}/${channel/\//-}"
+
+    [ -d "${dir}" ] || mkdir -p "${dir}"
+    (
+        set -eux
+        cd "${dir}"
+        snap download "${name}" --channel="${channel}" --cohort="+"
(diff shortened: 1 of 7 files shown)


test/includes/snap.sh
```diff
@@ -0,0 +1,90 @@
...
+                recursive_call=true
+                break
+            fi
+        done
+
+        # If files are missing and we're not in a recursive call
+        if { [ -z "${assert}" ] || ! [ -e "${snap}" ]; } && [ "${recursive_call}" = "false" ]; then
+            echo "Opportunistically downloading ${name} before installation"
+            if download_snap "${name}" "${channel}"; then
+                echo "Download successful, retrying installation"
```
tomponline, line 70: why "retrying installation", it hasn't tried to install yet?

simondeziel: Good catch, will include in a subsequent PR.

Issue #16347, opened 2025-08-29
`lxc info --resources` should learn about `bcache`


simondeziel:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

On a machine with `bcache` configured:

```sh
# lsblk
NAME        MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
loop0         7:0    0  50.8M  1 loop /snap/snapd/25202
loop1         7:1    0  66.8M  1 loop /snap/core24/1055
loop2         7:2    0 112.4M  1 loop /snap/lxd/35505
sda           8:0    0  10.9T  0 disk 
└─bcache0   251:0    0  10.9T  0 disk 
sdb           8:16   0  10.9T  0 disk 
└─bcache1   251:128  0  10.9T  0 disk 
sdc           8:32   0  10.9T  0 disk 
sdd           8:48   0  10.9T  0 disk 
sde           8:64   0  10.9T  0 disk 
sdf           8:80   0  10.9T  0 disk 
nvme2n1     259:0    0 447.1G  0 disk 
nvme0n1     259:1    0 447.1G  0 disk 
nvme4n1     259:2    0 447.1G  0 disk 
├─bcache0   251:0    0  10.9T  0 disk 
└─bcache1   251:128  0  10.9T  0 disk 
nvme5n1     259:3    0 447.1G  0 disk 
nvme1n1     259:4    0   5.8T  0 disk 
├─nvme1n1p1 259:5    0   512M  0 part /boot/efi
└─nvme1n1p2 259:6    0   5.8T  0 part /
nvme3n1     259:7    0   5.8T  0 disk 

# blkid
/dev/sdb: UUID="251de71f-d1e3-490e-870d-11b4e766510a" TYPE="bcache"
/dev/nvme1n1p2: LABEL="root" UUID="4c955080-edc6-4ddd-9e45-cfd8eeb24c9a" BLOCK_SIZE="4096" TYPE="ext4" PARTUUID="50736f4f-2393-4f5f-b154-4dbbeb9c4298"
/dev/nvme1n1p1: LABEL_FATBOOT="efi" LABEL="efi" UUID="B833-48BF" BLOCK_SIZE="512" TYPE="vfat" PARTUUID="009dcac8-736d-4d9b-8236-439d63152ebd"
/dev/sda: UUID="42207410-d17e-4e49-a175-ec8370e038ee" TYPE="bcache"
[...]

PR #16435, opened 2025-09-08
Resources: Handle bcache storage devices

roosterfish:
Fixes https://github.com/canonical/lxd/issues/16347

When querying the `/1.0/resources` endpoint, we should also show `bcache` devices. Furthermore a `bcache` device can (and usually does) have devices configured as cache and backing device. See https://www.kernel.org/doc/html/latest/admin-guide/bcache.html.

Those devices should be filtered out from the regular list in any case as they should only ever be used through the `bcache` device.

### DIFF
--- a/test/main.sh
+++ b/test/main.sh
@@ -528,6 +528,7 @@ if [ "${1:-"all"}" != "cluster" ]; then
     run_test test_storage_volume_import "storage volume import"
     run_test test_storage_volume_initial_config "storage volume initial configuration"
     run_test test_resources "resources"
+    run_test test_resources_bcache "resources bcache"
     run_test test_kernel_limits "kernel limits"
     run_test test_console "console"
     run_test test_query "query"


simondeziel:
In terms of terminology, it seems like at one point, upstream moved to "caching" device and "backing" device(s) (as a replacement for "slave") leaving only the `sysfs` paths using the older terminology. I think we should do the same.

roosterfish:
> In terms of terminology, it seems like at one point, upstream moved to "caching" device and "backing" device(s) (as a replacement for "slave") leaving only the `sysfs` paths using the older terminology. I think we should do the same.

I have removed almost all occurrences but left the variable names tied to the sysfs path for clarity purposes.

test/suites/resources.sh
```diff
@@ -10,3 +10,47 @@ test_resources() {
...
+  # Print for debugging purposes.
+  lxc query /1.0/resources | jq '.storage.disks'
+
+  # Check the bcache device is returned by LXD.
+  [ "$(lxc query /1.0/resources | jq -r '.storage.disks.[] | select(.id == "bcache0")')" != "" ]
+
+  # Get the bcache cache and backing devices.
+  cache_device_base="$(basename "${loop_device_1}")"
+  backing_device_base="$(basename "${loop_device_2}")"
+  cache_device="$(cat "/sys/block/${cache_device_base}/dev")"
```
simondeziel:
```suggestion
  cache_device="$(< "/sys/block/${cache_device_base}/dev")"
```

I've replaced all those instances were a small file is read by `$(< ...)` to avoid forking. It's a tiny optimization but for consistency, would you mind doing it here too, please?

simondeziel (dismissed):

test/suites/resources.sh
roosterfish: Fixed. Also moved the `modprobe` part into the test itself, found it's better suited there, also in case you run the test suite locally and don't have the mod loaded.

simondeziel (approved):
Awesome and the runtime cost barely registers even with `ceph`.

lxd/resources/storage.go
```diff
@@ -158,6 +199,11 @@ func GetStorage() (*api.ResourcesStorage, error) {
 			mountedIDs[fields[2]] = true
 		}
 
+		bcacheDevs, err := bcacheDevices(entries)
```
tomponline: where is this used?

lxd/resources/storage.go
```diff
@@ -246,7 +247,14 @@ func GetStorage() (*api.ResourcesStorage, error) {
 				return nil, fmt.Errorf("Failed to read %q: %w", filepath.Join(entryPath, "dev"), err)
 			}
 
-			disk.Device = strings.TrimSpace(string(diskDev))
+			diskDevStr := strings.TrimSpace(string(diskDev))
+
+			// Skip the device if it is used as either bcache cache or backing device.
+			if slices.Contains(bcacheDevs, diskDevStr) {
```
tomponline: im not sure about this because im not sure what the contract of `GetStorage()` is.

The comment to the function just says:

> // GetStorage returns a filled api.ResourcesStorage struct ready for use by LXD.


but what does that mean?

Does it mean "only storage devices that can be used as backing devices to new storage pools", or does it mean "all storage devices on the system", this line is only valid depending on which definition we choose.

tomponline: api.ResourcesStorage is defined as "represents the local storage", not "ResourcesStorage represents the local storage available for use with a pool".

We need to remember that the MAAS team use the resources package too and understand if/how they use this function.

lxd/resources/storage.go
roosterfish: Ups, that is confusing, moved it to two commits later when it's actually used. Fixed.

lxd/resources/storage.go
roosterfish: I see your concern. 

From what I can see we only use it for the `lxc info --resources` command. I'll check the MAAS usage.

If possible I'd like to narrow down the list by definition to return only "usable" disks. But if we rather want to use the list as general information about disks present in the system, we could also think about adding another field like `bcache_parent` which links to another device. Certainly this change will cause less friction but more work on e.g. MicroCloud to do proper disk filtering.

tomponline: Im fine to change it if MAAS dont mind.

roosterfish: Fixed by adding a `used_by` field which can be populated with the respective virtual device type like `bcache` for any device which has a virtual device parent.

doc/api-extensions.md
```diff
@@ -2714,3 +2714,7 @@ See {ref}`DevLXD bearer tokens <devlxd-authentication-bearer>`.
 Changes to existing endpoints:
 
 * `GET /1.0` — Adds a `supported_storage_drivers` field to the response, which is populated when `security.devlxd.management.volumes` is enabled.
+
+## `resources_disk_used_by`
+
+Adds the field `used_by` to potential storage disk returned by the resources end point to indicate its use by any virtual parent device, e.g. bcache.
```
simondeziel, line 2734: If we ever support mdadm, DRBD and such, will we need new extensions?

roosterfish: The idea would be to just set `used_by` to e.g. `mdadm` instead which we could just slide in at some point without having to add a new extension.

simondeziel: OK, cool, thanks!

roosterfish: An information we don't transport with this field is the actual virtual device which is using this disk. But I wasn't too sure if this concept would fit for all the various disk configurations so I instead settled on just telling if a disk is used by any of those virtual disk systems.
This is fine for the MicroCloud use case because we just need to filter out any disk which has `used_by != ""`.

lxd/resources/storage.go
```diff
@@ -433,6 +439,12 @@ func GetStorage() (*api.ResourcesStorage, error) {
 				return nil, err
 			}
 
+			// Identify if the disk is in use by any bcache device.
+			// The bcache device's own 'bcache' path is a link, not a directory.
+			if sysfsIsDir(filepath.Join(devicePath, "bcache")) {
+				disk.UsedBy = "bcache"
```
tomponline, line 445: would it be more useful to include the sysfs path to the used_by device rather than just the device type?

roosterfish: I was now running this through multiple tests. However there unfortunately is no 1:1 mapping between a virtual bcache dev and an actual cache device. Something like this is possible where you have three bcache devs (each having their own backing dev) but they all use the same cache device `/dev/sde`:

```bash
sdb         8:16   0    10G  0 disk 
└─bcache0 250:0    0    10G  0 disk 
sdc         8:32   0    10G  0 disk 
└─bcache1 250:128  0    10G  0 disk 
sdd         8:48   0    10G  0 disk 
└─bcache2 250:256  0    10G  0 disk 
sde         8:64   0    10G  0 disk 
├─bcache0 250:0    0    10G  0 disk 
├─bcache1 250:128  0    10G  0 disk 
└─bcache2 250:256  0    10G  0 disk 
```

So the `.UsedBy` would need to accommodate a list of devices. For sure `UsedBy` could be a `[]string` but I wonder if this would also fit for the other types like mdadm, multipath etc. 
But for them we could potentially only set one element in `UsedBy`. What do you guys think @tomponline @simondeziel ?

tomponline: OK lets use a string and put what ever identifier we can use for the parent, in this case just "bcache"

roosterfish: Ok great, then the PR is ready for another check as this is the current behavior. I have just rebased to fix the API extension conflict.

lxd/resources/storage.go
```diff
@@ -164,9 +164,15 @@ func GetStorage() (*api.ResourcesStorage, error) {
 			entryPath := filepath.Join(sysClassBlock, entryName)
 			devicePath := filepath.Join(entryPath, "device")
 
-			// Only keep the main entries not partitions
+			// Only keep the main entries not partitions.
+			// Also account for bcache devices.
 			if !sysfsExists(devicePath) {
-				continue
+				if !sysfsExists(filepath.Join(entryPath, "bcache")) {
+					continue
```
tomponline, line 171: isnt this still excluding bcache devices?

lxd/resources/utils.go
```diff
@@ -53,6 +53,15 @@ func sysfsExists(path string) bool {
 	return err == nil
 }
 
+func sysfsIsDir(path string) bool {
```
tomponline, line 56: this function has nothing to do with sysFs, like the `sysfsExists` function above it, seems like a weird name to pick.

lxd/resources/storage.go
roosterfish: The hierarchy is like this:

```
/sys/class/block
\_ /sda
    \_ device (link)
\_ /sdb
    \_ device (link)
\_ /sdc
     \_ device (link)
\_ /bcache0 # no "device" link
```

In case of physical disks, their `/sys/class/block/sd*/` contains a link to the actual device. Whereas the bcache doesn't have such a `device` link as its virtual.
So in this case the `entryPath` needs to have a `bcache` directory.

lxd/resources/utils.go
roosterfish: https://github.com/canonical/lxd/pull/16474

PR #13035, opened 2024-03-04
API: Unembed Put structs from Get structs

MusicDin:
Unembed Put structs from Get structs.

Fixes https://github.com/canonical/lxd/issues/12993

PR #16442, opened 2025-09-09
Public microcloud flag

### LINKED ISSUE #16384: Indicate microcloud membership to logged out users
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

The LXD-UI is changing logo, title and favicon when the LXD backend is running under a microcloud. Currently, the server setting `user.microcloud` is used to initiate the switch. However this setting is not readable to logged out users. See also the blocked issue in lxd-ui: https://github.com/canonical/lxd-ui/issues/1460

### Describe the solution you'd like

I propose one of the two solutions:

1. Make the server setting `user.microcloud` readable by unauthenticated users via `GET /1.0` api endpoint
2. Add a new field to the `GET /1.0` endpoint that signals the microcloud membership to logged in and logged out users.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

edlerd:
# Done
- make the `user.microcloud` flag on the server config publicly visible

Fixes #16384
Unblocks https://github.com/canonical/lxd-ui/issues/1460

### DIFF
--- a/doc/rest-api.yaml
+++ b/doc/rest-api.yaml
@@ -6045,9 +6045,9 @@ definitions:
                 x-go-name: ClientCertificate
             config:
                 additionalProperties: {}
-                description: Server configuration map (refer to doc/server.md)
+                description: Server configuration map (refer to doc/server.md) The available fields for public endpoint (before authentication) are limited.
                 example:
-                    core.https_address: :8443
+                    user.microcloud: "true"
                 type: object
                 x-go-name: Config
             environment:
@@ -6307,6 +6307,13 @@ definitions:
                 readOnly: true
                 type: boolean
                 x-go-name: ClientCertificate
+            config:
+                additionalProperties: {}
+                description: Server configuration map (refer to doc/server.md) The available fields for public endpoint (before authentication) are limited.
+                example:
+                    user.microcloud: "true"
+                type: object
+                x-go-name: Config
             public:
                 description: Whether the server is public-only (only public endpoints are implemented)
                 example: false


shared/api/server.go
```diff
@@ -179,6 +179,8 @@ type ServerPut struct {
 //
 // swagger:model
 type ServerUntrusted struct {
+	ServerPut `yaml:",inline"`
```
tomponline: I'm hesistant to embed ServerPut in case we add something to this later that shouldnt be in the untrusted ServerUntrusted struct type.

Instead I think we should be explicit and add 

```
Config map[string]any `json:"config" yaml:"config"`
```
 to this struct directly.

lxd/api_1.0.go
```diff
@@ -261,6 +261,14 @@ func api10Get(d *Daemon, r *http.Request) response.Response {
 
 	// If not authenticated, return now.
 	if !requestor.IsTrusted() {
+		daemonConfig, _ := daemonConfigRender(s)
+		_, flagExists := daemonConfig["user.microcloud"]
+		if flagExists {
+			// Unprivileged users may see the user.microcloud config key
+			srv.Config = map[string]any{}
+			srv.Config["user.microcloud"] = daemonConfig["user.microcloud"]
```
tomponline: can initialise and set at once using:

```go
srv.Config = map[string]any{
     "user.microcloud": daemonConfig["user.microcloud"]
}
```

doc/rest-api.yaml
```diff
@@ -6307,6 +6307,13 @@ definitions:
                 readOnly: true
                 type: boolean
                 x-go-name: ClientCertificate
+            config:
+                additionalProperties: {}
+                description: Server configuration map (refer to doc/server.md)
+                example:
+                    core.https_address: :8443
```
tomponline: this example is another reason why `config` should be in ServerUntrusted as its own field so we can use the user.microcloud example here

minaelee: Was this added manually? `rest-api.yaml` must be [generated from the code](https://documentation.ubuntu.com/lxd/latest/contributing/#api-updates) via `make update-api` rather than manually edited (will be overwritten next time it's generated from the code).

minaelee (changes requested):

doc/rest-api.yaml
edlerd: Yes, I used `make update-api` to create this.

shared/api/server.go
```diff
@@ -217,15 +217,19 @@ type ServerUntrusted struct {
...
 }
 
 // Server represents a LXD server
 //
 // swagger:model
 type Server struct {
 	WithEntitlements `yaml:",inline"`
-	ServerPut        `yaml:",inline"`
 	ServerUntrusted  `yaml:",inline"`
+	ServerPut        `yaml:",inline"`
```
edlerd: Needed to switch the order here, so the doc for the config field from `ServerUntrusted` gets correctly overridden with the one from `ServerPut`

shared/api/server.go
edlerd: This has the downside, that the `server.Config` references to a `type Server` object are now ambiguous. See also the failing CI. We could change them to be explicit by using `server.ServerPut.Config` everywhere, but not sure if that is a good way. Wdyt?

tomponline: I mean we shouldn't put ServerPut in ServerUntrusted, and instead add `Config` field directly to `ServerUntrusted`.

tomponline: Ah I see what you mean, ServerUntrusted is embedded in Server for unconditionally untrusted fields.
Now we have a partial untrusted field this complicates matters.

Lets re-group and re-think the plan as this isnt going to work.

markylaing:
FYI https://github.com/lxc/incus/pull/1071/commits is how Incus did this

tomponline:
> FYI https://github.com/lxc/incus/pull/1071/commits is how Incus did this

Yeah dislike for same reason that it confuses Put fields with untrsuted Get fields.

Perhaps we should stop embedding ServerPut in Server at all, and just duplicate the Config fields. @MusicDin did something similar a while back for other API types.

shared/api/server.go
edlerd: We can have a dedicated `ServerConfig` type with only the `config` field. It has a generic description that works both logged in and logged out. Include `Serverconfig` in `ServerPut` and `ServerUntrusted`. Then avoid including `ServerPut` in `Server`.

tomponline: I think lets just put the Config field in both rather than having a new struct that contains just one field. Its clear the field is for config after all.

This aligns with the type of untangling that @MusicDin did a while back.

https://github.com/canonical/lxd/pull/13035

And while this is a small GO SDK API break, its not a REST API break and updating affected apps to will continue to work with old and new LXD servers.

edlerd: All right, went with duplicating the config field in `ServerUntrusted` and `Server`.

tomponline:
@edlerd if ServerPut only as Config in it it doesn't need to be embedded in Server anymore.

edlerd:
> @edlerd if ServerPut only as Config in it it doesn't need to be embedded in Server anymore.

There is a problem with removing, as it breaks the writable helper inside of server.go:

```
func (srv *Server) Writable() ServerPut {
	return srv.ServerPut
}
```

shared/api/server.go
```diff
@@ -217,6 +217,10 @@ type ServerUntrusted struct {
 	//
 	// API extension: client_cert_presence
 	ClientCertificate bool `json:"client_certificate" yaml:"client_certificate"`
+
+	// Server configuration map limited to publicly visible keys (refer to doc/server.md)
+	// Example: {"user.microcloud": ":true"}
```
minaelee: Is `":true"` intended to be `"true"` or is that `:` intentional?
Shows up like this in the [preview](https://canonical-ubuntu-documentation-library--16442.com.readthedocs.build/lxd/16442/api/#/server/server_get_untrusted):
image

minaelee (changes requested):

shared/api/server.go
edlerd: Whoops, that is a typo indeed!

edlerd: Thanks for raising this, fixed it just now.

tomponline:
> Server) Writable()

Ah I see, makes sense, ok keep as is then, so Config in all 3 structs.

tomponline:
@edlerd btw do we have any `user.ui.` keys left that are set by the UI?

tomponline:
> > Server) Writable()
> 
> Ah I see, makes sense, ok keep as is then, so Config in all 3 structs.

ah no, we should have `server.Writable()` return a new `ServerPut` populated with its `Config`, and then unembed `ServerPut` from `Server`.

Like the other `Writable()` functions do.

minaelee (approved):
LGTM docs-wise, ty!

edlerd:
@tomponline everything above should be resolved by now, please have another go at this one.

shared/api/server.go
```diff
@@ -217,14 +217,19 @@ type ServerUntrusted struct {
...
+	// Server configuration map limited to publicly visible keys (refer to doc/server.md)
+	// Example: {"user.microcloud": "true"}
+	Config map[string]any `json:"config" yaml:"untrusted_config"` // config field is overridden downstream
 }
 
 // Server represents a LXD server
 //
 // swagger:model
+//
+//nolint:govet // config field is overridden downstream
```
tomponline: Do we need this now?

edlerd: Indeed, it can be removed. Did it just now.

tomponline: Lovey stuff thanks!

shared/api/server.go
```diff
@@ -217,14 +217,17 @@ type ServerUntrusted struct {
 	//
 	// API extension: client_cert_presence
 	ClientCertificate bool `json:"client_certificate" yaml:"client_certificate"`
+
+	// Server configuration map limited to publicly visible keys (refer to doc/server.md)
+	// Example: {"user.microcloud": "true"}
+	Config map[string]any `json:"config" yaml:"untrusted_config"` // config field is overridden downstream
```
tomponline: Do we need "// config field is overridden downstream" any more?

shared/api/server.go
```diff
@@ -241,12 +244,18 @@ type Server struct {
 	// API extension: auth_user
 	AuthUserMethod string `json:"auth_user_method" yaml:"auth_user_method"`
 
+	// Server configuration map (refer to doc/server.md)
+	// Example: {"core.https_address": ":8443"}
+	Config map[string]any `json:"config" yaml:"config"` // override the inherited field
```
tomponline: Do we need "// override the inherited field" anymore?

shared/api/server.go
edlerd: I thought it might be good to mention as the untrusted server object is still included in the server object.

shared/api/server.go
edlerd: This is overriding the field from `ServerUntrusted`, so maybe worth keeping?

shared/api/server.go
tomponline: Ah yes I see. Well now originally I was not happy with having `ServerPut` embedded in `ServerUntrusted`, so we've now untangled that by moving `Config` into `ServerUntrusted` directly and removing `ServerPut` from `Server`.

How about we now remove `Config` from `Server` and rely on the field from `ServerUntrusted`, as its embedded in `Server` already. This will remove any ambiguity.

The only downside is that we'd need to have only a single "example" field in the docs, which I suggest we keep as `user.microcloud`.

This overall feels simpler and safer.

What do you think?

tomponline:
@edlerd would you mind following up with a test that checks A) the user.microcloud key is accessible without auth, and B) that other config keys are not.

edlerd:
> @edlerd would you mind following up with a test that checks A) the user.microcloud key is accessible without auth, and B) that other config keys are not.

Sure, planning it in next pulse with this ticket: https://warthogs.atlassian.net/browse/WD-26747

tomponline:
> > @edlerd would you mind following up with a test that checks A) the user.microcloud key is accessible without auth, and B) that other config keys are not.
> 
> Sure, planning it in next pulse with this ticket: https://warthogs.atlassian.net/browse/WD-26747

Thanks, please include a CLI test in this repo using `lxc config get <key>` so we can catch any regressions early.

PR #16446, opened 2025-09-09
Request package refactorings

kadinsayani:
Refactorings identified during review of https://github.com/canonical/lxd/pull/16396.

### DIFF
--- a/lxd/storage_pools.go
+++ b/lxd/storage_pools.go
@@ -328,7 +328,7 @@ func storagePoolsPost(d *Daemon, r *http.Request) response.Response {
 	lc := lifecycle.StoragePoolCreated.Event(req.Name, requestor.EventLifecycleRequestor(), ctx)
 	resp := response.SyncResponseLocation(true, nil, lc.Source)
 
-	clientType := request.UserAgentClientType(r.Header.Get("User-Agent"))
+	clientType := requestor.ClientType()
 
 	if requestor.IsClusterNotification() {
 		// This is an internal request which triggers the actual
@@ -832,9 +832,7 @@ func storagePoolPut(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
-	clientType := request.UserAgentClientType(r.Header.Get("User-Agent"))
-
-	response := doStoragePoolUpdate(s, pool, req, targetNode, clientType, r.Method, s.ServerClustered)
+	response := doStoragePoolUpdate(s, pool, req, targetNode, requestor.ClientType(), r.Method, s.ServerClustered)
 
 	ctx := logger.Ctx{}
 	if targetNode != "" {
@@ -1000,7 +998,7 @@ func storagePoolDelete(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
-	clientType := request.UserAgentClientType(r.Header.Get("User-Agent"))
+	clientType := requestor.ClientType()
 	clusterNotification := requestor.IsClusterNotification()
 	var notifier cluster.Notifier
 	if !clusterNotification {
--- a/lxd/network_acls.go
+++ b/lxd/network_acls.go
@@ -588,9 +588,12 @@ func networkACLPut(d *Daemon, r *http.Request) response.Response {
 		}
 	}
 
-	clientType := request.UserAgentClientType(r.Header.Get("User-Agent"))
+	requestor, err := request.GetRequestor(r.Context())
+	if err != nil {
+		return response.SmartError(err)
+	}
 
-	err = netACL.Update(&req, clientType)
+	err = netACL.Update(&req, requestor.ClientType())
 	if err != nil {
 		return response.SmartError(err)
 	}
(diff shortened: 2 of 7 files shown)


kadinsayani:
@markylaing we can go over this in our 1:1. Marking as draft in the meantime.

lxd/request/requestor.go
```diff
@@ -283,6 +289,8 @@ func SetRequestor(req *http.Request, identityCache *identity.Cache, args Request
 	callerUsername := r.CallerUsername()
 	callerProtocol := r.CallerProtocol()
 
+	r.clientType = UserAgentClientType(req.Header.Get("User-Agent"))
```
markylaing: Unexport `UserAgentClientType`

lxd/request/requestor.go
```diff
@@ -283,6 +289,8 @@ func SetRequestor(req *http.Request, identityCache *identity.Cache, args Request
...
+
```
markylaing: Move below "Trusted" comment:

```go
clientType := UserAgentClientType(req.Header.Get("User-Agent"))
if args.Protocol == ProtocolCluster {
    r.clientType = clientType
} else {
    if clientType != ClientTypeNormal {
         return api.StatusError(http.StatusForbidden, "oainepfinw") // Same error we return in Daemon.Authenticate
    }
}
```

markylaing: For now let's allow unix socket to have the user agent set as well. Documenting the exception

lxd/storage_pools.go
```diff
@@ -822,18 +822,21 @@ func storagePoolPut(d *Daemon, r *http.Request) response.Response {
...
+	requestor, err := request.GetRequestor(r.Context())
+	if err != nil {
+		return response.SmartError(err)
+	}
 
-	response := doStoragePoolUpdate(s, pool, req, targetNode, clientType, r.Method, s.ServerClustered)
+	response := doStoragePoolUpdate(s, pool, req, targetNode, requestor, r.Method, s.ServerClustered)
 
-	requestor := request.CreateRequestor(r.Context())
+	eventLifecycleRequestor := request.CreateRequestor(r.Context())
```
markylaing: `requestor.EventLifecycleRequestor()`

tomponline:
@kadinsayani ready for rebase

lxd/request/requestor.go
```diff
@@ -312,6 +324,20 @@ func SetRequestor(req *http.Request, identityCache *identity.Cache, args Request
...
+	// If "core.trust_ca_certificates" is false, the peer certificate is additionally verified via mTLS and
+	// RequestorArgs.Protocol is set to [api.AuthenticationMethodTLS].
+	// XXX: We allow ProtocolUnix because initDataNodeApply() in lxd/init.go uses a local client to join a cluster. initDataNodeApply() is used by 'lxd init' and PUT /1.0/cluster.
+	allowedClusterProtocols := []string{ProtocolCluster, ProtocolUnix}
+	_, err = r.ClusterMemberTLSCertificateFingerprint()
+	if err == nil {
+		allowedClusterProtocols = append(allowedClusterProtocols, api.AuthenticationMethodTLS)
+	}
+
+	if r.clientType != ClientTypeNormal && !slices.Contains(allowedClusterProtocols, callerProtocol) {
```
tomponline, line 344: @markylaing @kadinsayani If im reading this correctly we are allowing notification requests from clients that are authenticating using TLS but are not cluster members. Is this right? And if so, why would we allow this?

markylaing: I'm also a little confused about this. Can we just have this check at the very beginning of the function to avoid confusion with other checks? I think it should be:
```go
func SetRequestor(req *http.Request, identityCache *identity.Cache, args RequestorArgs) error {
    clientType := userAgentClientType(req.Header.Get("User-Agent"))
    if clientType != ClientTypeNormal && !slices.Contains([]string{ProtocolCluster, ProtocolUnix}, args.Protocol) {
        // XXX: We allow ProtocolUnix because initDataNodeApply() in lxd/init.go uses a local client to join a cluster. initDataNodeApply() is used by 'lxd init' and PUT /1.0/cluster.
        return errors.New("Cluster notification isn't using trusted server certificate")
    }

    r := &Requestor{
		trusted:                args.Trusted,			trusted:                args.Trusted,
		originAddress:          req.RemoteAddr,			originAddress:          req.RemoteAddr,
		username:               args.Username,			username:               args.Username,
		protocol:               args.Protocol,			protocol:               args.Protocol,
		identityProviderGroups: args.IdentityProviderGroups,			identityProviderGroups: args.IdentityProviderGroups,
		clientType:             clientType,
    }

    /* Remaining function body */
}
```

tomponline: +1 from me to keep this stuff together and do it early.

Not sure about your code block though @markylaing ;)

kadinsayani: Ah, I see where my confusion stemmed from. The `clustering_ha` test was failing at a check using `nc` to verify direct connectivity. `callerProtocol` is set to the original caller protocol (`tls` in the case of the direct connectivity check). I incorrectly included `tls` as an allowed protocol. https://github.com/canonical/lxd/pull/16470 fixes this.

PR #16447, opened 2025-09-09
Minor tweaks to `lxd-agent`

simondeziel:


lxd-agent/devlxd.go
```diff
@@ -427,9 +426,10 @@ func registerDevLXDEndpoint(d *Daemon, apiRouter *mux.Router, apiVersion string,
 
 // Create a new net.Listener bound to the unix socket of the devLXD endpoint.
 func createDevLXDListener(dir string) (net.Listener, error) {
-	path := filepath.Join(dir, "lxd", "sock")
+	parentDir := dir + "/lxd"
+	path := parentDir + "/sock"
```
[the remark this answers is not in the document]
simondeziel: The whole `lxd-agent` assumes a Linux host so joining Windows paths isn't worth considering, at least not now.

lxd-agent/devlxd.go
```diff
@@ -427,9 +426,10 @@ func registerDevLXDEndpoint(d *Daemon, apiRouter *mux.Router, apiVersion string,
...
 
-	err := os.MkdirAll(filepath.Dir(path), 0755)
+	err := os.MkdirAll(parentDir, 0755)
```
MusicDin, line 432: `filepath` is used in Go to handle file paths and will take the OS specific path separator. I would prefer having better file path handling over micro optimization, as this ensures file paths are always correctly handled.

tomponline: yes i tend to agree.

lxd-agent/daemon.go
```diff
@@ -58,25 +57,27 @@ func (d *Daemon) init() error {
 
 	// Check whether we should start the devlxd server in the early setup. This way, /dev/lxd/sock
 	// will be available for any systemd services starting after the lxd-agent.
-	if shared.PathExists("agent.conf") {
-		f, err := os.Open("agent.conf")
-		if err != nil {
-			return err
+	f, err := os.Open("agent.conf")
+	if err != nil {
+		if os.IsNotExist(err) {
```
tomponline: IsNotExist is deprecated see https://pkg.go.dev/os#IsNotExist

lxd-agent/devlxd.go
simondeziel: I've tracked `filepath.Join()` and it goes pretty deep due to calling `filepath.Clean()` among other things. I stopped tracking it 5 levels deep.

I don't feel strongly about it but it felt like an innocuous place to just concatenate strings together for something that is definitely not going to be port to Windows any time soon.

tomponline: It only happens on lxd-agent start up, if it were happening in a fast path that might be more of an argument to micro optimise. In this case as the values are static i dont mind keeping it as it is in the PR.

But in general I'd like us to try and target our optimisations on more beneficial hot-path areas where the improvement can be noticeable.

simondeziel: Agreed, it's just that I was looking at the `lxd-agent` startup in relation with the early boot problem we have with `cloud-init` detecting it's inside a LXD instance.

It was not "an optimization effort" per say, more like a drive-by thing.

tomponline (approved):

PR #16450, opened 2025-09-10
devlxd: Volume managent API preparation

MusicDin:
Helper functions that are subset of commits from DevLXD volume management API (https://github.com/canonical/lxd/pull/16278).
The main intention is to reduce the size of the volume management API PR.

### DIFF
--- a/lxd/response/response_capture.go
+++ b/lxd/response/response_capture.go
@@ -84,3 +84,23 @@ func (rc *responseCapture) Render(resp Response) error {
 
 	return nil
 }
+
+// RenderToStruct renders the response into a struct and returns the ETag.
+func (rc *responseCapture) RenderToStruct(resp Response, target any) (etag string, err error) {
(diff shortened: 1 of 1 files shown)


lxd/request/request.go
```diff
@@ -22,3 +24,23 @@ func CreateRequestor(ctx context.Context) *api.EventLifecycleRequestor {
...
+	if err != nil {
+		return nil, err
+	}
+
+	identity := requestor.CallerIdentity()
+	if identity == nil {
+		return nil, errors.New("Request context identity is missing")
+	}
+
+	if identity.Identifier == "" {
```
tomponline, line 41: @markylaing should this check be added to `requestor.CallerIdentity()` such that it returns nil if `identity.Identifier` is empty?

Do you think we need this `GetCallerIdentityFromContext` at all, and if so, should we be updating other call sites of `requestor.CallerIdentity()` to use it?

PR #16451, opened 2025-09-10
devlxd: Include supported storage drivers in DevLXD state

MusicDin:
This PR adds API extension `devlxd_volume_management` and includes a list of supported storage drivers in `/1.0` endpoint when security flag `devlxd.security.management.volumes` is enabled on the instance.


doc/api-extensions.md
```diff
@@ -2708,3 +2708,8 @@ This change removes the previous limitation on `ovn` networks that load balancer
 
 Adds a new `bearer` authentication method and enables authentication to the DevLXD API.
 See {ref}`DevLXD bearer tokens <devlxd-authentication-bearer>`.
+
+## `devlxd_volume_management`
+
+Changes to existing endpoints:
+- `GET /1.0` — Adds a `supported_storage_drivers` field to the response, which is populated when `security.devlxd.management.volumes` is enabled.
```
tomponline, line 2715: Changes to existing endpoints -> Changes to existing devlxd endpoints

MusicDin: Fixed in https://github.com/canonical/lxd/pull/16278

simondeziel: This `- ` prefix seems to upset the doc linter: https://github.com/canonical/lxd/actions/runs/17614289729/job/50043413615#step:6:23

MusicDin: Missed that. I'll fix that

simondeziel: In your defense, the check only triggers once merged so you didn't actually missed it :)
We've made some trades offs with the doc checks as they are too brittle/prone to fail so they only run on merges.

PR #16453, opened 2025-09-10
Allow viewing operations with same Requestor

markylaing:
`GET /1.0/operations` currently filters out operations where the caller does not have `can_view_operations` on the operations' project.

This PR changes access handling in `GET /1.0/operations` such that the operation will be visible if:
- The requestor of the operations matches the requestor of the caller.
- It has no project, and the caller has `can_edit` on `server`. These are background tasks and are not currently properly handled. See commit messages.
- It has a project, and the caller has `can_view_operations` on the project (as before).

In addition, the `GET /1.0/operations/{id}` and `GET /1.0/operations/{id}/wait` handlers were not performing sufficient access control (see commit messages). This is now added to match what is performed when listing operations.

Depends on #15991

### DIFF
--- a/test/suites/tls_restrictions.sh
+++ b/test/suites/tls_restrictions.sh
@@ -46,6 +46,44 @@ test_tls_restrictions() {
   # Allow access to project blah
   lxc config trust show "${FINGERPRINT}" | sed -e "s/projects: \[\]/projects: ['blah']/" -e "s/restricted: false/restricted: true/" | lxc config trust edit "${FINGERPRINT}"
 
+  # The restricted caller can listen for events on all projects, but the events are filtered to only those in the projects they have access to.
+  monfile_root="${TEST_DIR}/mon-root.jsonl"
+  lxc monitor --all-projects --type lifecycle --format json > "${monfile_root}" &
+  mon_root_pid=$!
+  sleep 0.1
+
+  monfile_restricted="${TEST_DIR}/mon-restricted.jsonl"
+  lxc remote switch localhost
+  lxc monitor --all-projects --format json > "${monfile_restricted}" &
+  mon_restricted_pid=$!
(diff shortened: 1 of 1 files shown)


markylaing:
Marking as draft while #15991 is still in review

markylaing:
Once this is merged the UI should be unblocked

tomponline:
> * It has no project, and the caller has `can_edit` on `server`. These are background tasks and are not currently properly handled. See commit messages.

Shouldnt this be `can_view_operations` on the server?

tomponline:
@markylaing test failures

markylaing:
@tomponline this one is ready when you are. Thanks.

markylaing:
Ok maybe not ready - tests blew up

markylaing:
Tests should pass now

markylaing:
Tests still flaky 🤦‍♂️ will look in the morning

markylaing:
Tests really should be more stable now. Ready for review.

lxd/operations.go
```diff
@@ -984,8 +984,30 @@ func operationWaitGet(d *Daemon, r *http.Request) response.Response {
...
 	if err == nil {
-		if secret != "" && op.Metadata()["secret"] != secret {
-			return response.Forbidden(nil)
+		if secret != "" {
+			// If a secret is provided and it matches the operation, allow access.
+			if op.Metadata()["secret"] != secret {
+				return response.Forbidden(nil)
+			}
+		} else {
+			// Allow view access if the caller is the requestor.
```
tomponline, line 993: this code looks duplicated from the previous commit, can we have a separate function to avoid the duplication and reducing our TICS score?

markylaing: https://github.com/canonical/lxd/pull/16591

PR #16458, opened 2025-09-10
Revert "lxd-benchmark: log failure to close CSV"

simondeziel:
This reverts commit d14605c2e8699b2a92eb570e09693bb70a1665bc.

Having the happy path closing the file and also logging errors in the deferred close leads to noise:

```
[Sep 10 14:55:24.217] Batch processing completed in 0.675s
[Sep 10 14:55:24.217] Written report file /tmp/lxd-test.tmp.OhwL/Ox2
[Sep 10 14:55:24.217] Error closing file /tmp/lxd-test.tmp.OhwL/Ox2: close /tmp/lxd-test.tmp.OhwL/Ox2: file already closed
```

### DIFF
--- a/lxd-benchmark/benchmark/report.go
+++ b/lxd-benchmark/benchmark/report.go
@@ -28,13 +28,18 @@ type CSVReport struct {
 }
 
 // Load reads current content of the filename and loads records.
-func (r *CSVReport) Load() error {
+func (r *CSVReport) Load() (err error) {
 	file, err := os.Open(r.Filename)
 	if err != nil {
 		return err
 	}
 
-	defer func() { _ = file.Close() }()
+	defer func() {
+		cerr := file.Close()
+		if err == nil {
+			err = cerr
+		}
+	}()
 
 	reader := csv.NewReader(file)
 	for line := 1; err != io.EOF; line++ {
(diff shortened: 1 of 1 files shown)


simondeziel:
@tomponline it's quite annoying that `file.Close()` an already closed file leads to an error. I don't see how we could please TiCS by checking errors in deferred closure.

I still think it's worthwhile to fo `file.Close()` on the happy path where applicable to catch IO errors returned by the OS. In the `Load()` function, I left it with a final `return nil` because the file is only opened for reading from it.

simondeziel:
> @tomponline it's quite annoying that `file.Close()` an already closed file leads to an error. I don't see how we could please TiCS by checking errors in deferred closure.
> 
> I still think it's worthwhile to fo `file.Close()` on the happy path where applicable to catch IO errors returned by the OS. In the `Load()` function, I left it with a final `return nil` because the file is only opened for reading from it.

Based on the "Update 2" section of https://www.joeshaw.org/dont-defer-close-on-writable-files/, I think the `Write()` function should `return file.Sync()` and have the deferred closure update the error if `Sync()` was successful.

However, that same section warns about the performance hit incurred by the sync call so I'd reserve this for known valuable data. WDYT?


**Update**: the whole blog post is quite old as it mentions that calling `File.Close()` multiple times is OK due to the file descriptor being set to `-1` once closed the first time. This was true at the time of writing (Go 1.8) but changed in Go 1.9 :/

simondeziel:
The CI run failed during the [concurrent test](https://github.com/canonical/lxd/actions/runs/17622332377/job/50071337312?pr=16458#step:17:1459):

```
Error: Failed deleting instance "concurrent-13" in project "default": Error deleting storage volume: Failed to run: zfs destroy -r lxdtest-x4g/containers/concurrent-13: exit status 1 (cannot open 'lxdtest-x4g/containers/concurrent-13': dataset does not exist)
+ '[' 1 -eq 127 ']'
+ cleanup
```

I'm probably biased ATM but I feel like we should invest a bit into making ZFS more reliable as it's the primary FS advertised for local storage with MicroCloud...

PR #16459, opened 2025-09-10
Network: We don't need an OVN volatile uplink when the address is none

escabo:
Observed this while trying to edit a network definition to change its ipv4 segment. The validation expected an ipv6 volatile entry but ipv6 was actually turned off with none.

Fixes https://github.com/canonical/lxd/issues/16607.

### DIFF
--- a/lxd/network/driver_ovn.go
+++ b/lxd/network/driver_ovn.go
@@ -3375,7 +3375,7 @@ func (n *ovn) Update(newNetwork api.NetworkPut, targetNode string, clientType re
 			uplinkAddrKey = ovnVolatileUplinkIPv6
 		}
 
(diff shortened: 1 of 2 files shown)


tomponline:
Moving back to draft until comments are addressed. Thanks

tomponline:
@escabo are you OK if @nmezhenskyi finishes this one off?

escabo:
> @escabo are you OK if @nmezhenskyi finishes this one off?

Yes. I would have liked to get to it but it's better if we wrap this up quickly. Thanks @nmezhenskyi

tomponline:
@escabo im preparing a LXD interim release so could cherry-pick it into it if its done in time is all.

tomponline (approved):
LGTM thanks both!

PR #16460, opened 2025-09-10
Early redirection to leader

simondeziel:


lxd/acme.go
tomponline, line 1: There's a tradeoff to be made here, do we do simple and cheap validation twice, or do we forward early (which is costly) when the cheaper validation might have avoided the need to in case of invalid requests.

simondeziel: This being an unauthenticated endpoint, an attacker wanting to DoS LXD would provide valid parameters to avoid any early return and maximize the processing needing to be done by LXD.

This is why I opted for making it as cheap as possible on the forwarding member.

tomponline: I see your point.

In general though if we take the cluster as a whole, the member that servers the request from the client can be seen as the "boundary" of the cluster, so in my mind dropping invalid requests early (especially if the check is cheap to do so) would make sense. 

I take your point so this isn't a blocker, but something to keep in mind that forwarded requests have, in my view, effectively "made it" into the cluster.

lxd/api_cluster.go
```diff
@@ -2001,7 +2001,6 @@ func clusterNodeDelete(d *Daemon, r *http.Request) response.Response {
 
 	// Redirect all requests to the leader, which is the one with
```
tomponline, line 2002: this comment looks like it can be dropped/moved, doesnt make much sense here.

lxd/api_cluster.go
tomponline, line 1: We can change the `logger.Debugf("Redirect member accept request to %s", leaderInfo.Address)` to use contextual logging.

tomponline, line 1: We can change the `logger.Debugf("Redirect handover request to %s", leaderInfo.Address)` to use contextual logging.

PR #16462, opened 2025-09-10
Misc ZFS tweaks

simondeziel:


lxd/storage/drivers/driver_zfs_utils.go
```diff
@@ -305,24 +285,22 @@ func (d *zfs) getDatasetProperties(dataset string, keys ...string) (map[string]s
...
 func (d *zfs) version() (string, error) {
 	// Loaded kernel module version
-	if shared.PathExists("/sys/module/zfs/version") {
-		out, err := os.ReadFile("/sys/module/zfs/version")
-		if err == nil {
-			return strings.TrimSpace(string(out)), nil
-		}
+	outBytes, err := os.ReadFile("/sys/module/zfs/version")
+	if err == nil {
+		return strings.TrimSpace(string(outBytes)), nil
```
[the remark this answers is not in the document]
simondeziel: No, because few lines below `out` is used and it has a different type (`string`) so it wouldn't even compile.

lxd/storage/drivers/driver_zfs_volumes.go
```diff
@@ -1923,22 +1922,30 @@
...
 func (d *zfs) getVolumeDiskPathFromDataset(dataset string) (string, error) {
-	zvolUdevLink := filepath.Join("/dev/zvol", dataset)
-
 	// Shortcut for udev.
-	sb, err := os.Stat(zvolUdevLink)
+	// If the udev link exists, resolve it and confirm it's a block device
+	// before returning it. Chasing the symlink to the actual device node
+	// (`/dev/zdX`) ensures consistency with the "slow path" where `zvol_id` is
+	// invoked directly.
+	devPath, err := os.Readlink("/dev/zvol/" + dataset)
```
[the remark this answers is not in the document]
simondeziel: Not a new alert.

simondeziel:
@tomponline I re-ran the ZFS standalone tests multiple times for good measure. It never failed. Only the `btrfs` one failed for unrelated reasons.

simondeziel:
@tomponline the failed CI is due to CodeQL rediscovering an old alert due to code churn.

lxd/storage/drivers/driver_zfs_volumes.go
tomponline, line 1: Please can you add a better commit message rather than just "improve ..." what improved?

simondeziel: I tweaked just the commit message.

lxd/storage/drivers/driver_zfs_utils.go
tomponline, line 1: FWIW im not a big fan of doing this because it makes it easier, in my view, to mess up the error handling when refactoring later in the code because the last code block in a function is "special" from the rest.

lxd/storage/drivers/driver_zfs_volumes.go
tomponline: much better thanks!

lxd/storage/drivers/driver_zfs_utils.go
simondeziel: Fair, should I drop those commits?

tomponline: If you dont feel strongly about it.

simondeziel: I've dropped them in the end. Don't want to make any refactor any more dangerous.

Issue #15833, opened 2025-06-19
`AllowProjectUpdate` has its own (partial) implementation of `checkRestrictionsAndAggregateLimits`


tomponline:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

24.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
N/A
```

### Output of "lxc info" or system info if it fails

```shell
N/A
```

### Issue description

This issue is related to https://github.com/canonical/lxd/issues/15822 but broader in scope.

Put simply, when something is added/changed inside a project its restrictions and limits are checked for violation using a function called `checkRestrictionsAndAggregateLimits()`.

However when a project's own settings are updated, a function called `AllowProjectUpdate()` is called.

Inside this function it checks that restrictions are still valid by calling `checkInstanceRestrictions()`, however that is not entirely correct which is covered by https://github.com/canonical/lxd/issues/15822.

Separately to that `AllowProjectUpdate()` also seems to have its own implementation of `checkRestrictionsAndAggregateLimits()` by individually considering each limit key individually:

https://github.com/canonical/lxd/blob/5290649b7b8f1665413f2a02cf9ec8af961de76e/lxd/project/limits/permissions.go#L1121-L1159

As we as not being updated to handle `limits.networks` (see https://github.com/canonical/lxd/issues/15823) it is also undesirable to have two different implementations for limit checks and we should instead use `checkRestrictionsAndAggregateLimits` when updating a project's settings to ensure the project's entities comply with the project's new restrictions & limits.

### Steps to reproduce

N/A

### Information to attach

- [ ] Any relevant kernel output (`dmesg`)
- [ ] Instance log (`lxc info NAME --show-log`)
- [ ] Instance configuration (`lxc config show NAME --expanded`)
- [ ] Main daemon log (at `/var/log/lxd/lxd.log` or `/var/snap/lxd/common/lxd/logs/lxd.log`)
- [ ] Output of the client with `--debug`
- [ ] Output of the daemon with `--debug` (or use `lxc monitor` while reproducing the issue)

Issue #15823, opened 2025-06-17
Project `limits.networks` is not checked during project update via `checkAggregateLimits`


tomponline:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

24.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
N/A
```

### Output of "lxc info" or system info if it fails

```shell
N/A
```

### Issue description

The `limits.networks` setting doesn't appear to be validated via `checkAggregateLimits` during project update.

### Steps to reproduce

```
lxc project create footest -c features.networks=true -c limits.networks=1
lxc project switch footest
lxc network create foo --type=ovn
lxc network create foo2 --type=ovn # Should not be allowed
lxc project set footest limits.networks=0 # Should not be allowed as there is currently 1 network
```

### Information to attach

- [ ] Any relevant kernel output (`dmesg`)
- [ ] Instance log (`lxc info NAME --show-log`)
- [ ] Instance configuration (`lxc config show NAME --expanded`)
- [ ] Main daemon log (at `/var/log/lxd/lxd.log` or `/var/snap/lxd/common/lxd/logs/lxd.log`)
- [ ] Output of the client with `--debug`
- [ ] Output of the daemon with `--debug` (or use `lxc monitor` while reproducing the issue)

nmezhenskyi:
I noticed [a related issue](https://github.com/canonical/lxd/issues/15833) to this. It looks like it's better to tackle them together. So I've assigned it to myself as well.

nmezhenskyi:
@tomponline Should we close this as fixed by https://github.com/canonical/lxd/pull/16463?

PR #16463, opened 2025-09-10
Project: Improve validation during project update

nmezhenskyi:
Fixes https://github.com/canonical/lxd/issues/15833 and https://github.com/canonical/lxd/issues/15823.

This PR refactors the project restrictions and limits validation logic and reduces code duplication. It also adds validation for `limits.networks` limit which wasn't previously checked during the project update.

### DIFF
--- a/test/suites/network_ovn.sh
+++ b/test/suites/network_ovn.sh
@@ -706,10 +706,32 @@ test_network_ovn() {
   lxc network load-balancer create "${project_ovn_network}" 192.0.2.2
   lxc network load-balancer create "${project_ovn_network}" 2001:db8:1:2::2
 
-  # Clean up
+  echo "==> Delete load balancers and the network."
   lxc network load-balancer delete "${project_ovn_network}" 192.0.2.2
   lxc network load-balancer delete "${project_ovn_network}" 2001:db8:1:2::2
   lxc network delete "${project_ovn_network}"
+
+  echo "==> Test that limits.networks is enforced."
+
+  echo "==> Set limits.networks=1 for the project."
+  lxc project set testovn limits.networks=1
+
+  echo "==> Create project network ovn1."
+  lxc network create ovn1 --type=ovn network="${uplink_network}" ipv4.address=192.0.2.1/24 ipv6.address=2001:db8:1:2::1/64
+
+  echo "==> Check that creating more networks is not allowed because limits.networks=1."
+  ! lxc network create ovn2 --type=ovn network="${uplink_network}" || false
+
+  echo "==> Check that limits.networks is enforced during project update."
+  ! lxc project set testovn limits.networks=0 || false
+
+  echo "==> Delete the network."
+  lxc network delete ovn1
+
+  echo "==> Check that updating limits.networks is allowed now that there are no created networks."
+  lxc project set testovn limits.networks=0
+
+  echo "==> Clean up the project."
   lxc project switch default
   lxc project delete testovn
 


lxd/project/limits/permissions.go
```diff
@@ -343,14 +349,18 @@ func GetImageSpaceBudget(ctx context.Context, globalConfig *clusterConfig.Config
...
+// to the project-wide instance count should be checked separately where needed.
+func checkInstanceRestrictionsAndAggregateLimits(globalConfig *clusterConfig.Config, tx *db.ClusterTx, info *projectInfo) error {
 	// List of config keys for which we need to check aggregate values
 	// across all project instances.
 	aggregateKeys := []string{}
 	isRestricted := false
 
 	for key, value := range info.Project.Config {
-		if slices.Contains(allAggregateLimits, key) || strings.HasPrefix(key, projectLimitDiskPool) {
+		isKnownLimitsKey := slices.Contains(allInstanceAggregateLimits, key) || strings.HasPrefix(key, projectLimitDiskPool)
```
simondeziel: Could you only evaluate this after checking that `value != ""`?

lxd/project/limits/permissions.go
```diff
@@ -25,8 +25,14 @@ import (
 	"github.com/canonical/lxd/shared/validate"
 )
 
-// projectLimitDiskPool is the prefix used for pool-specific disk limits.
-var projectLimitDiskPool = "limits.disk.pool."
+const (
```
simondeziel, line 28: Nice :)

test/suites/network_ovn.sh
```diff
@@ -706,10 +706,32 @@ test_network_ovn() {
...
   lxc network load-balancer delete "${project_ovn_network}" 2001:db8:1:2::2
   lxc network delete "${project_ovn_network}"
+
+  echo "==> Test that limits.networks is enforced."
+
+  echo "==> Set limits.networks=1 for the project."
+  lxc project set testovn limits.networks=1
+
+  echo "==> Create project network ovn1."
+  lxc network create ovn1 --type=ovn network="${uplink_network}"
```
simondeziel: Generating random addresses isn't cheap (see #16521):

```
2025-09-18T03:20:29.8200707Z ++ timeout --foreground 120 /home/runner/go/bin/lxc network create ovn1 --type=ovn network=uplink6164 --verbose
2025-09-18T03:20:33.5038758Z Network ovn1 created
```

That's ~4s. Could you use `ipv4.address=none ipv6.address=none` to avoid most of the overhead, please?

nmezhenskyi: > Could you use ipv4.address=none ipv6.address=none to avoid most of the overhead, please?  

In this case the network creation fails with this error: `Error: No internal IPs defined for network router`

nmezhenskyi: I could try just specifying an available IP

tomponline: you can use the test ip ranges https://en.wikipedia.org/wiki/Reserved_IP_addresses

nmezhenskyi: Updated

lxd/project/limits/permissions.go
```diff
@@ -1188,6 +1194,27 @@ func validateInstanceCountLimit(instances []api.Instance, key, value, project st
...
+
+	limit, err := strconv.Atoi(limitValue)
+	if err != nil {
+		return err
+	}
+
+	count := len(networks)
+
+	if limit < count {
+		return fmt.Errorf("network limit exceeded: there currently are %d networks", count)
```
simondeziel: Should we use a capital letter here?

lxd/project/limits/permissions.go
```diff
@@ -1076,89 +1089,51 @@ func checkUplinkUse(ctx context.Context, tx *db.ClusterTx, projectName string, c
...
 
-		case "limits.processes":
-			fallthrough
-		case "limits.cpu":
-			fallthrough
-		case "limits.memory":
-			fallthrough
-		case "limits.disk":
-			aggregateKeys = append(aggregateKeys, key)
-		}
```
simondeziel, line 1149: I must have missed it but can't find/understand where those are checked now.

lxd/project/limits/permissions.go
```diff
@@ -1076,89 +1089,51 @@ func checkUplinkUse(ctx context.Context, tx *db.ClusterTx, projectName string, c
...
 			err := validateTotalInstanceCountLimit(info.Instances, config[key], projectName)
 			if err != nil {
-				return fmt.Errorf("Can't change limits.instances in project %q: %w", projectName, err)
+				return fmt.Errorf("Can't change %q in project %q: %w", key, projectName, err)
 			}
 
-		case "limits.containers":
-			fallthrough
-		case "limits.virtual-machines":
+		case "limits.containers", "limits.virtual-machines":
```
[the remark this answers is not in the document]
simondeziel: I disagree here. Doing a comparison for 2 items reads better and is also a bit more performant.

lxd/project/limits/permissions.go
```diff
@@ -25,8 +25,14 @@ import (
 	"github.com/canonical/lxd/shared/validate"
 )
 
-// projectLimitDiskPool is the prefix used for pool-specific disk limits.
-var projectLimitDiskPool = "limits.disk.pool."
+const (
+	// projectLimitDiskPool is the prefix used for pool-specific disk limits.
+	projectLimitDiskPool string = "limits.disk.pool."
```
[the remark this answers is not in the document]
simondeziel: Wait, what?

lxd/project/limits/permissions.go
nmezhenskyi: These are now checked by the `checkAggregateLimits()` function, which is called inside of the `checkInstanceRestrictionsAndAggregateLimits()`.

nmezhenskyi: `checkAggregateLimits()` accepts a list of limits keys to check and we pass in there the keys from `allInstanceAggregateLimits` variable, if they have been set by the updated project config.

```Go
var allInstanceAggregateLimits = []string{
	"limits.cpu",
	"limits.disk",
	"limits.memory",
	"limits.processes",
}

func checkAggregateLimits(info *projectInfo, aggregateKeys []string) error {
	if len(aggregateKeys) == 0 {
		return nil
	}

	totals, err := getTotalsAcrossProjectEntities(info, aggregateKeys, false)
	if err != nil {
		return fmt.Errorf("Failed getting usage of project entities: %w", err)
	}

	for _, key := range aggregateKeys {
		keyName := key

		// Handle pool-specific limits.
		if strings.HasPrefix(key, projectLimitDiskPool) {
			keyName = "limits.disk"
		}

		parser := aggregateLimitConfigValueParsers[keyName]
		maxValue, err := parser(info.Project.Config[key])
		if err != nil {
			return err
		}

		if totals[key] > maxValue {
			return fmt.Errorf("Reached maximum aggregate value %q for %q in project %q", info.Project.Config[key], key, info.Project.Name)
		}
	}

	return nil
}
```

nmezhenskyi: This is how we already check these limits in `AllowInstanceCreation()`, `AllowVolumeCreation()`, etc.  
I just made it so that `AllowProjectUpdate` now uses the same approach.

lxd/project/limits/permissions.go
nmezhenskyi: Updated

lxd/project/limits/permissions.go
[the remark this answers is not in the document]
nmezhenskyi: I was looking at it for 2 minutes trying to understand is it me or is it Copilot who got it wrong :)

simondeziel (approved):
LGTM, thanks!

lxd/project/limits/permissions.go
```diff
@@ -358,7 +358,8 @@ func checkInstanceRestrictionsAndAggregateLimits(globalConfig *clusterConfig.Con
 	isRestricted := false
 
 	for key, value := range info.Project.Config {
-		if slices.Contains(allInstanceAggregateLimits, key) || strings.HasPrefix(key, projectLimitDiskPool) {
+		// Check that the key is a known limits key and the value is not empty.
+		if value != "" && (slices.Contains(allInstanceAggregateLimits, key) || strings.HasPrefix(key, projectLimitDiskPool)) {
```
tomponline, line 362: why do we need to do this?

lxd/project/limits/permissions.go
```diff
@@ -1089,90 +1089,46 @@ func checkUplinkUse(ctx context.Context, tx *db.ClusterTx, projectName string, c
...
-	info.Instances, err = expandInstancesConfigAndDevices(globalConfigDump, info.Instances, info.Profiles)
+	// Set the fetched project's config to new config for validation.
+	info.Project.Config = config
+
+	// This checks if restricted uplinks are used within this project and if limits
+	// for uplink IP usage are respected.
+	// This is done separately because this may require getting network info from the database.
+	err = checkUplinkUse(ctx, tx, projectName, info.Project.Config)
 	if err != nil {
 		return err
```
tomponline, line 1105: should we wrap this error for clarity?

lxd/project/limits/permissions.go
```diff
@@ -1089,90 +1089,46 @@ func checkUplinkUse(ctx context.Context, tx *db.ClusterTx, projectName string, c
...
-
-			continue
-		}
-
 		switch key {
 		case "limits.instances":
 			err := validateTotalInstanceCountLimit(info.Instances, config[key], projectName)
 			if err != nil {
-				return fmt.Errorf("Can't change limits.instances in project %q: %w", projectName, err)
+				return fmt.Errorf("Can't change %q in project %q: %w", key, projectName, err)
```
tomponline, line 1121: this should be a separate commit

lxd/project/limits/permissions.go
```diff
@@ -1089,90 +1089,46 @@ func checkUplinkUse(ctx context.Context, tx *db.ClusterTx, projectName string, c
...
 			}
 
-		case "limits.containers":
-			fallthrough
-		case "limits.virtual-machines":
+		case "limits.containers", "limits.virtual-machines":
```
tomponline, line 1124: this should be a separate commit

lxd/project/limits/permissions.go
```diff
@@ -1089,90 +1089,46 @@ func checkUplinkUse(ctx context.Context, tx *db.ClusterTx, projectName string, c
...
 			err := validateInstanceCountLimit(info.Instances, key, config[key], projectName)
 			if err != nil {
 				return fmt.Errorf("Can't change %q in project %q: %w", key, projectName, err)
 			}
 
-		case "limits.processes":
```
tomponline, line 1141: are these covered by the call to `checkInstanceRestrictionsAndAggregateLimits` above?

tomponline:
Thanks!

lxd/project/limits/permissions.go
nmezhenskyi: > why do we need to do this?

To check that the value is non-empty? There was an issue without this check where we would try to parse empty string as a valid limit. This is specific to `AllowProjectUpdate()` because we pass `""` string values in a new config to signify that a project option should be unset.

tomponline: Got you, thanks!

lxd/project/limits/permissions.go
nmezhenskyi: > are these covered by the call to `checkInstanceRestrictionsAndAggregateLimits` above?

Yes, this is covered by the `checkInstanceRestrictionsAndAggregateLimits()`.

lxd/project/limits/permissions.go
nmezhenskyi: > should we wrap this error for clarity?

Yes, I think we should do this, because some of the errors returned by `checkUplinkUse()` are a bit unclear. I'll open a small follow-up PR.

PR #16467, opened 2025-09-12
Container: Ignore clean stop of SFTP server

### LINKED ISSUE #15925: Forcibly stopping an instance should not spam logs about leftover `sftp` server
### Steps to reproduce

Create an instance, pull a file from it and immediately stop it with force:

```
$ lxc init ubuntu-minimal-daily:24.04 c1
$ lxc start c1; lxc file pull c1/etc/hostname -; lxc stop --force c1
c1
```

See the log spam:

```
$ journalctl -n 3 -u snap.lxd.daemon
Jul 02 16:48:33 sdeziel-lemur lxd.daemon[767970]: time="2025-07-02T16:48:33-04:00" level=warning msg="SFTP server stopped with error" err="signal: terminated" instance=c1 instanceType=container project=default stderr=
```

This `sftp` server is kept around for a few seconds to avoid spawning a new one for each request, that's good but it shouldn't matter if it's SIGTERM'ed.

simondeziel:
If a SFTP server was spawned and is still around when a container is stopped, it will be `SIGTERM`'ed to do an orderly shutdown. This is perfectly normal so it shouldn't cause any log message.

Fixes #15925

### DIFF
--- a/lxd/instance/drivers/driver_lxc.go
+++ b/lxd/instance/drivers/driver_lxc.go
@@ -5930,11 +5930,14 @@ func (d *lxc) FileSFTPConn() (net.Conn, error) {
 		close(chReady)
 
 		// Wait for completion.
-		err = forkfile.Wait()
-		if err != nil {
-			d.logger.Warn("SFTP server stopped with error", logger.Ctx{"err": err, "stderr": strings.TrimSpace(stderr.String())})
+		// Do not log anything if the process received a SIGTERM (exit status 128+15=143)
+		// as that likely means the instance is being stopped.
+		exitStatus, err := shared.ExitStatus(forkfile.Wait())
+		if exitStatus == 143 {
 			return
 		}
+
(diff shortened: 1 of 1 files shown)


lxd/instance/drivers/driver_lxc.go
```diff
@@ -5932,6 +5932,18 @@ func (d *lxc) FileSFTPConn() (net.Conn, error) {
 		// Wait for completion.
 		err = forkfile.Wait()
 		if err != nil {
+			// Check if this is an ExitError
+			var exitErr *exec.ExitError
```
tomponline: Can you use the existing function we have called `ExitStatus()` here?

simondeziel: Thanks, I didn't know about this handy helper!

simondeziel: Worked quite well.

tomponline (approved):
Ta

PR #16476, opened 2025-09-12
Auth: OIDC sessions

markylaing:
Implements server side sessions for OIDC authenticated identities. 

When a user authenticates to LXD via an IdP, LXD issues a JWT and stores it in a `session` cookie. The JWT does not contain any user identifiable information. LXD can verify the JWT and inspect the payload to determine the session and identity.

When a session JWT expires but is otherwise valid, LXD will attempt to start a new session by reverifying the identity with the IdP using credentials saved server side. 

For the CLI, no credentials are saved server side. If a token sent by the CLI has expired, LXD will use the Bearer token (which the CLI continues to send).

Note that old CLI versions will not be able to authenticate with a version of LXD with this API extension, as they cannot persistently store cookies. LXD expects the client to express that they are able to store cookies by setting a `cookiejar` feature in the user agent header.

Also note that it is no longer permitted for the IdP subject to change. This is intentionally made static for security reasons.

Fixes https://github.com/canonical/lxd/issues/14174

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -162,7 +162,7 @@ jobs:
           set -eux
 
           # bin/max (sizes are in MiB)
-          SIZES="lxc 16
+          SIZES="lxc 17
                  lxd-agent 14"
           MIB="$((1024 * 1024))"
 


lxd/auth/oidc/oidc.go
```diff
@@ -481,26 +481,36 @@ func (o *Verifier) setRelyingParty(ctx context.Context, host string) error {
...
+	http.SetCookie(w, &http.Cookie{
+		Name:     cookieNameSession,
+		Path:     "/",
+		Secure:   true,
+		HttpOnly: true,
+		SameSite: http.SameSiteStrictMode,
+		Value:    token,
+		// The cookie has to have an expiry for the persistent cookie jar (CLI) to save it to disk.
+		// This sets the cookie to expire one week after the session expires. This allows sessions to be refreshed.
+		Expires: expiry.Add(time.Hour * 24 * 7),
```
markylaing: We need to define this buffer on the expiry somewhere as a const, as it should also be used to clean up tokens from the database.

lxd/auth/oidc/oidc.go
```diff
@@ -514,6 +514,18 @@ func (o *Verifier) startSession(r *http.Request, w http.ResponseWriter, res Auth
 	return nil
 }
 
+// endSession deletes the session cookie. It does not delete the session from the database.
+func (o *Verifier) endSession(w http.ResponseWriter) {
+	http.SetCookie(w, &http.Cookie{
```
markylaing, line 698: We need to also delete the session from the database.

lxd/oidc_sessions.go
```diff
@@ -0,0 +1,313 @@
...
+	Name:        "oidc_session",
+	Path:        "auth/oidc-sessions/{id}",
+	MetricsType: entity.TypeIdentity,
+	Get: APIEndpointAction{
+		Handler:       oidcSessionGet,
+		AccessHandler: oidcSessionAccessHandler(auth.EntitlementCanView),
+	},
+	Delete: APIEndpointAction{
+		Handler:       oidcSessionDelete,
+		AccessHandler: oidcSessionAccessHandler(auth.EntitlementCanDelete),
```
markylaing, line 54: This is `can_delete` so that oidc identities can delete their own sessions without any special logic.

shared/version/useragent.go
```diff
@@ -62,3 +52,65 @@ func UserAgentFeatures(features []string) {
...
+	}
+
+	return &ClientUserAgent{
+		Name:         matches[0][1],
+		Version:      matches[0][2],
+		IsLTS:        isLTS,
+		OS:           strings.Split(matches[0][4], ";"),
+		Capabilities: strings.Split(matches[0][5], ";"),
+	}, nil
+}
```
markylaing: This is just a proposal for what the client user agent could look like. I'm open to suggestions, but was going by the spec here https://www.rfc-editor.org/rfc/rfc9110#name-user-agent

lxc/main.go
```diff
@@ -22,6 +22,18 @@ import (
 	"github.com/canonical/lxd/shared/version"
 )
 
+var capabilities = []string{
+	"cookiejar",
```
markylaing: Probably need to make this a const in the API package because the OIDC verifier will be looking for the same string in client capabilities before trying to set a cookie.

lxd/auth/oidc/oidc.go
```diff
@@ -140,7 +142,16 @@ func (o *Verifier) userInfo(ctx context.Context, token string) (*oidc.UserInfo,
 // authenticateBearerToken calls the /userinfo endpoint with the given token to retrieve information about the user. Then
 // starts a new session.
 func (o *Verifier) authenticateBearerToken(r *http.Request, w http.ResponseWriter, accessToken string) (*AuthenticationResult, error) {
-	err := o.ensureConfig(r.Context(), r.Host)
+	agent, err := version.ParseClientUserAgent(r.UserAgent())
+	if err != nil {
+		return nil, api.StatusErrorf(http.StatusBadRequest, "OIDC authentication requires cookie support. Please update your client")
+	}
+
+	if !slices.Contains(agent.Capabilities, "cookiejar") {
```
markylaing: Need a const in API package for `cookiejar`.

markylaing:
@tomponline Of everything in this PR, the only thing that I think I can extract into it's own PR is the client user agent definition :cry: 

It's ready for an initial overview if you want to get a sense of it. 

@nmezhenskyi I will book a meeting for us to go over the details on wednesday

lxd/auth/oidc/oidc.go
markylaing: Defined as a const in the oidc package, which is imported into the daemon (lxd) package.

lxd/auth/oidc/oidc.go
markylaing: Done.

markylaing:
One last item on my list for this is to see if I can find a workable alternative to https://github.com/juju/persistent-cookiejar, which hasn't been updated in over 8 years.

lxd/auth/bearer/bearer.go
```diff
@@ -22,43 +24,76 @@ func IsDevLXDRequest(r *http.Request, clusterUUID string) (isRequest bool, token
...
 	if err != nil {
-		return false, "", ""
+		return "", nil, errors.New("Token is not a JWT")
 	}
 
 	// There must be an issuer
 	issuer, err := t.Claims.GetIssuer()
 	if err != nil {
-		return false, "", ""
+		return "", nil, errors.New("Token does not have an issuer")
```
kadinsayani: Is the reason for not wrapping errors returned by `GetIssuer()` and `GetSubject()` to avoid leaking information from the token?

lxd/auth/encryption/sign.go
```diff
@@ -22,6 +23,12 @@ func Issuer(clusterUUID string) string {
 	return strings.Join([]string{"lxd", clusterUUID}, ":")
 }
 
+// LXDAudience is the expected audience for the main API tokens (including OIDC session tokens).
+// This is the same as the [Issuer].
+func LXDAudience(clusterUUID string) string {
+	return Issuer(clusterUUID)
```
kadinsayani, line 29: Can we just use Issuer() at the call-site since it's exported?

shared/api/auth.go
```diff
@@ -367,3 +371,38 @@ type PermissionInfo struct {
...
+	// Username is the name of the user that holds the session.
+	// Example: Jane Doe
+	Username string `json:"username" yaml:"username"`
+
+	// IP is the IP address of the user that holds the session.
+	// Example: 10.21.242.46
+	IP string `json:"ip" yaml:"ip"`
+
+	// UserAgent is the UserAgent of the user that holds the session.
+	// Example: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/139.0.0.0 Safari/537.36
```
kadinsayani, line 398: Should this example follow the format of LXD user agent strings?

lxd/db/cluster/oidc_sessions.go
```diff
@@ -0,0 +1,193 @@
...
+	AccessToken  string
+	RefreshToken string
+	IP           string
+	UserAgent    string
+	ExpiryDate   time.Time
+}
+
+// ToAPI returns an [api.OIDCSession] from the [OIDCSession].
+func (s OIDCSession) ToAPI() (*api.OIDCSession, error) {
+	sessionUUID, err := uuid.Parse(s.UUID)
```
kadinsayani: Consider updating and using `IsUUID()` from the `shared/validate` package.

lxd/auth/oidc/oidc.go
```diff
@@ -27,6 +27,14 @@ import (
 	"github.com/canonical/lxd/shared/logger"
 )
 
+// SessionHandler is used where session handling must call the database. Methods should only be called after a client
+// session cookie or bearer token has been verified.
```
kadinsayani: Could calling these methods before verifying a client session cookie or bearer token cause any unintended side effects?

shared/api/client.go
kadinsayani, line 1: Alternatively, this constant could go in `shared/api/auth.go`.

lxd/auth/bearer/bearer.go
markylaing: I'll double check - can't remember why I did this.

lxd/auth/encryption/sign.go
markylaing: We could do. I just added this for clarity at the call site as it might not be obvious that you need to set the audience to be the same as the issuer.

shared/api/auth.go
markylaing: It could be a CLI or it could be a browser. The user agent we use for the CLI is subject to change, and I'd rather not use the LXD one as LXD shouldn't be connecting to other LXD's via OIDC.

lxd/db/cluster/oidc_sessions.go
markylaing: I don't think it even needs to be parsed here. If we're adding validation on UUIDs I would create a type for it an implement `driver.Valuer` & `sql.Scanner` so a bad value is never written to the DB.

lxd/auth/oidc/oidc.go
markylaing: The side effect is that an unauthenticated user would be able to cause LXD to write/read from the database using bogus tokens. I'll add to the comment.

shared/api/client.go
markylaing: It could. I have no objections to moving it there, but also no objections to keeping it where it is :)

What would be the reason for moving it? I just created a new file for this so that if another client feature is added, it's clear where to find it.

tomponline: its not about auth so no reason to place it in shared/api/auth

lxd/db/cluster/identities.go
```diff
@@ -288,17 +288,27 @@ type OIDCMetadata struct {
 
 // Subject returns OIDC subject from the identity metadata. The [AuthMethod] of the [Identity] must be [api.AuthenticationMethodOIDC].
 func (i Identity) Subject() (string, error) {
```
tomponline: should this be renamed to OIDCSubject now we're using an "OIDC" prefix for OIDC specific functions?

tomponline: Or should we lose this function entirely and update callers to use i.OIDCMetadata().Subject?

shared/api/auth.go
tomponline: makes sense thanks

lxd/db/cluster/oidc_sessions.go
tomponline: agreed - this is reading something that is already in DB

lxd/db/cluster/oidc_sessions.go
```diff
@@ -0,0 +1,193 @@
...
+		Email:     s.Email,
+		Username:  s.Username,
+		IP:        s.IP,
+		UserAgent: s.UserAgent,
+		ExpiresAt: s.ExpiryDate,
+		CreatedAt: creationDate,
+	}, nil
+}
+
+const allSessionsQuery = `SELECT
```
tomponline, line 48: any scope to be using `cluster.RegisterStmt(` here for commonly executed queries?

lxd/auth/oidc/oidc.go
```diff
@@ -27,6 +27,14 @@ import (
...
+type SessionHandler interface {
```
tomponline, line 34: Is this interface actually needed? Do we have multiple implementations of it, if not shall we drop it and pass the concrete type around instead?

lxd/db/oidc/session.go
```diff
@@ -0,0 +1,240 @@
...
+	"github.com/canonical/lxd/lxd/lifecycle"
+	"github.com/canonical/lxd/lxd/request"
+	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/logger"
+)
+
+// NewSessionHandler returns a new [oidc.SessionHandler]. The getSessionExpiry function must return the current value
+// of oidc.session.expiry for the server configuration.
+func NewSessionHandler(db *db.Cluster, events *events.Server, getSessionExpiry func() string) oidc.SessionHandler {
```
tomponline, line 28: seems like we can just return `sessionHandler` here and only make the fields/functions exported that are in the interface.

lxd/db/oidc/session.go
```diff
@@ -0,0 +1,240 @@
...
+// a newer cluster secret.
+func (s *sessionHandler) StartSession(r *http.Request, res oidc.AuthenticationResult, tokens *zitadelOIDC.Tokens[*zitadelOIDC.IDTokenClaims], expiryOverride *time.Time) (*uuid.UUID, *time.Time, error) {
+	// Get a new session UUID. This is a v7 UUID from which we can extract the session creation date.
+	sessionID, err := uuid.NewV7()
+	if err != nil {
+		return nil, nil, fmt.Errorf("Failed to create new session UUID: %w", err)
+	}
+
+	// Get the session expiry. Use the override, else get from config.
+	var expiry time.Time
```
tomponline, line 55: why not make the `expiry` var the argument to the function instead of `expiryOverride` and then if its nil then set it to the config one. Why the need to have separate var for expiryOverride?

tomponline: also why the pointer for expiryOverride rather than requiring time.Time and checking for `IsZero()`?

lxd/db/oidc/session.go
```diff
@@ -0,0 +1,240 @@
...
+	} else {
+		expiryConfig := s.expiryFunc()
+		expiry, err = shared.GetExpiry(time.Now().UTC(), expiryConfig)
+		if err != nil {
+			return nil, nil, fmt.Errorf("Failed to get session expiry: %w", err)
+		}
+	}
+
+	// Extract the IP address to store in the session.
+	addrport, err := netip.ParseAddrPort(r.RemoteAddr)
```
tomponline: remoteAddr would be clearer no? "Address" includes port normally in LXD parlance.

lxd/db/oidc/session.go
```diff
@@ -0,0 +1,240 @@
...
+				// as social login) and should still have the same permissions.
+				//
+				// This is dangerous because it means that permissions are tied to an email address and not who the IdP
+				// considers to be a unique user. Consider that in a large organisation, it's possible for an email address
+				// to be re-used (someone joins the org with the same name as someone that has already left the org).
+				//
+				// So we will instead restrict changes to the IdP subject. The consequence is that email address/subject
+				// conflicts will need to be resolved by an administrator. If there does turn out to be a problem with
+				// social login, we may need a mechanism to "join accounts" - but we can approach it from a secure position.
+				logger.Warn("Encountered new OIDC login with unexpected subject", logger.Ctx{"email": res.Email, "subject": res.Subject})
```
tomponline: should we extend this warning with a suggestion on what needs to be resolved?

lxd/db/oidc/session.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	var action lifecycle.IdentityAction
+	err = s.db.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+		// Get the identity from their email address. If none is found, it's a first time login.
+		var firstTimeLogin bool
+		identity, err := cluster.GetIdentity(ctx, tx.Tx(), api.AuthenticationMethodOIDC, res.Email)
+		if err != nil {
+			if !api.StatusErrorCheck(err, http.StatusNotFound) {
+				return err
```
tomponline: should wrap the error here please

lxd/db/oidc/session.go
```diff
@@ -0,0 +1,240 @@
...
+
+			firstTimeLogin = true
+		}
+
+		// Check if we need to update the identity metadata
+		var doUpdateIdentity bool
+		if !firstTimeLogin {
+			existingMetadata, err := identity.OIDCMetadata()
+			if err != nil {
+				return err
```
tomponline: wrap the error here please

lxd/db/oidc/session.go
```diff
@@ -0,0 +1,240 @@
...
+// the [cluster.Identity] that holds the session.
+func (s *sessionHandler) GetIdentityBySessionID(ctx context.Context, sessionID uuid.UUID) (res *oidc.AuthenticationResult, tokens *zitadelOIDC.Tokens[*zitadelOIDC.IDTokenClaims], sessionExpiry *time.Time, err error) {
+	var identity *cluster.Identity
+	var session *cluster.OIDCSession
+	err = s.db.Transaction(ctx, func(ctx context.Context, tx *db.ClusterTx) error {
+		identity, session, err = cluster.GetIdentityAndSessionDetailsFromSessionID(ctx, tx.Tx(), sessionID)
+		return err
+	})
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
```
tomponline: this isn't needed as its already detect and converted in https://github.com/canonical/lxd/pull/16476/commits/8a5272b89487e90237a86b9d42b0209a9780e214#diff-05c8a91814014d829c71f3fed609672a586e19bb7c1ccadf667f80cfa08c4443R142

lxd/auth/encryption/sign.go
tomponline: makes sense thanks @markylaing

lxd/auth/bearer/bearer.go
tomponline: good point @kadinsayani needs a comment if there's a specific reason for this @markylaing

lxd/auth/bearer/bearer.go
```diff
@@ -22,43 +23,57 @@ func IsDevLXDRequest(r *http.Request, clusterUUID string) (isRequest bool, token
...
-		return false, "", ""
+		return "", nil, errors.New("Token issuer does not match")
 	}
 
-	// Expect the audience to be "devlxd:{cluster_uuid}".
-	expectAudience := encryption.DevLXDAudience(clusterUUID)
 	audience, err := t.Claims.GetAudience()
 	if err != nil {
-		return false, "", ""
+		return "", nil, errors.New("Token does not have an audience")
```
tomponline: Same point as @kadinsayani made above, why are we losing the original error here?

lxd/auth/bearer/bearer.go
```diff
@@ -22,43 +23,57 @@ func IsDevLXDRequest(r *http.Request, clusterUUID string) (isRequest bool, token
...
 
-	if len(audience) != 1 || audience[0] != expectAudience {
-		return false, "", ""
+	if len(audience) != 1 || audience[0] != expectedAudience {
+		return "", nil, errors.New("Token does not contain the expected audience")
+	}
+
+	issuedAt, err := t.Claims.GetIssuedAt()
+	if err != nil {
+		return "", nil, errors.New("Token does not contain an issued at field")
```
tomponline: Same point as @kadinsayani made above, why are we losing the original error here?

lxd/auth/bearer/bearer.go
```diff
@@ -86,12 +86,28 @@ func Authenticate(token string, subject string, identityCache *identity.Cache) (
 		return nil, err
 	}
 
+	err = verifyToken(token, func() ([]byte, error) {
+		return entry.Secret, nil
+	})
+	if err != nil {
+		return nil, err
```
tomponline: should we wrap this?

lxd/auth/bearer/bearer.go
```diff
@@ -31,6 +32,25 @@ func IsDevLXDRequest(r *http.Request, clusterUUID string) (isRequest bool, token
 	return true, token, subject
 }
 
+// IsSessionToken returns the session UUID and the issued at claim, or an error if it is not a LXD token.
+// The issued at claim is used to determine the cluster secret used when the signing key was derived.
+// The session ID is used as a salt when deriving the signing key from the cluster secret.
+// An error is returned here instead of a boolean (see IsDevLXDRequest) because the token is contained in a cookie
```
tomponline: @markylaing im afraid I was able to make the leap as to why "because the token is contained in a cookie that we set server side" was a reason for "An error is returned here instead of a boolean". So perhaps this could do with clarifying?

lxd/auth/oidc/oidc.go
```diff
@@ -505,6 +505,41 @@ func (o *Verifier) startSession(ctx context.Context, w http.ResponseWriter, idTo
 	return nil
 }
 
+// getSecretFromUsedAtTime returns the secret that should have been used to encrypt a cookie or sign a token at the
+// given unix time in seconds. A boolean is returned to indicate if the returned secret is not the most recent, this
+// prompts the Verifier to refresh the session token, so that it doesn't become unverifiable.
+func (o *Verifier) getSecretFromUsedAtTime(ctx context.Context, usedAtTimeUnixSeconds int64) (*cluster.AuthSecret, bool, error) {
```
tomponline: what about naming the bool return type "needsRefresh" rather than indicating its most recent, to make it clearer to the caller what they need to do?

lxd/auth/oidc/oidc.go
```diff
@@ -114,6 +114,26 @@ func (o *Verifier) Auth(w http.ResponseWriter, r *http.Request) (*Authentication
...
+	if err != nil {
+		return nil, err
+	}
+
+	// We only expect bearer tokens.
+	req.Header.Set("Authorization", "Bearer "+token)
+	var userinfo oidc.UserInfo
+	err = httphelper.HttpRequest(o.relyingParty.HttpClient(), req, &userinfo)
+	if err != nil {
+		return nil, err
```
tomponline: should we wrap this error?

lxd/auth/oidc/oidc.go
```diff
@@ -414,7 +414,19 @@ func (o *Verifier) Callback(w http.ResponseWriter, r *http.Request) {
 	}
 
 	handler := rp.CodeExchangeHandler(func(w http.ResponseWriter, r *http.Request, tokens *oidc.Tokens[*oidc.IDTokenClaims], state string, rp rp.RelyingParty) {
-		err := o.startSession(r.Context(), w, tokens.IDToken, tokens.RefreshToken)
+		userInfo, err := o.userInfo(r.Context(), tokens.AccessToken)
```
tomponline, line 525: could use an anonymous function here to only have a single error handler, e.g.

```go
err := func() error {
userInfo, err := o.userInfo(r.Context(), tokens.AccessToken)
...
res, err := o.getResultFromClaims(userInfo, userInfo.Claims)
...
}
		if err != nil {
			_ = response.ErrorResponse(http.StatusInternalServerError, fmt.Errorf("Failed to start a new session: %w", err).Error()).Render(w, r)
			return
		}
```

lxd/auth/oidc/oidc.go
```diff
@@ -134,40 +132,30 @@ func (o *Verifier) userInfo(ctx context.Context, token string) (*oidc.UserInfo,
...
-// is returned if no error occurs.
-func (o *Verifier) authenticateAccessToken(ctx context.Context, accessToken string) (*AuthenticationResult, error) {
-	claims, err := op.VerifyAccessToken[*oidc.AccessTokenClaims](ctx, accessToken, o.accessTokenVerifier)
+// authenticateBearerToken calls the /userinfo endpoint with the given token to retrieve information about the user. Then
+// starts a new session.
+func (o *Verifier) authenticateBearerToken(r *http.Request, w http.ResponseWriter, accessToken string) (*AuthenticationResult, error) {
+	err := o.ensureConfig(r.Context(), r.Host)
 	if err != nil {
-		return nil, AuthError{Err: fmt.Errorf("Failed to verify access token: %w", err)}
+		return nil, err
```
tomponline: should we wrap error here?

lxd/auth/oidc/oidc.go
```diff
@@ -134,40 +132,30 @@ func (o *Verifier) userInfo(ctx context.Context, token string) (*oidc.UserInfo,
...
-			Email:                  id.Identifier,
-			Name:                   id.Name,
-			Subject:                claims.Subject,
-			IdentityProviderGroups: o.getGroupsFromClaims(claims.Claims),
-		}, nil
-	} else if !api.StatusErrorCheck(err, http.StatusNotFound) {
-		return nil, fmt.Errorf("Failed to get OIDC identity from identity cache by their subject (%s): %w", claims.Subject, err)
+	res, err := o.getResultFromClaims(userInfo, userInfo.Claims)
+	if err != nil {
+		return nil, err
```
tomponline: should we wrap error here?

lxd/auth/oidc/oidc.go
```diff
@@ -134,40 +132,30 @@ func (o *Verifier) userInfo(ctx context.Context, token string) (*oidc.UserInfo,
...
 	}
 
-	userInfo, err := rp.Userinfo[*oidc.UserInfo](ctx, accessToken, oidc.BearerToken, claims.Subject, o.relyingParty)
+	err = o.startSession(r, w, *res, nil, nil)
 	if err != nil {
-		return nil, AuthError{Err: fmt.Errorf("Failed to call user info endpoint with given access token: %w", err)}
+		return nil, err
```
tomponline: should we wrap error here?

lxd/auth/oidc/oidc.go
```diff
@@ -158,64 +159,143 @@ func (o *Verifier) authenticateBearerToken(r *http.Request, w http.ResponseWrite
...
 	}
 
-	idTokenAny := tokens.Extra("id_token")
-	if idTokenAny == nil {
-		return nil, AuthError{Err: errors.New("ID tokens missing from OIDC refresh response")}
+	// If we signed the session with a key derived from an old cluster secret, start a new session and override the expiry.
+	if startNewSession {
+		err = o.startSession(r, w, *res, tokens, expiry)
+		if err != nil {
+			return nil, err
```
tomponline: wrap error?

lxd/cluster/config/config.go
```diff
@@ -756,6 +763,40 @@ var ConfigSchema = config.Schema{
...
+		exp, err := shared.GetExpiry(now, s)
+		if err != nil {
+			return err
+		}
+
+		if exp.Sub(now) < time.Hour {
+			return errors.New("OIDC session expiry cannot be set to less than one hour")
+		}
+
+		// TODO: This should be less than the auth secret expiry.
```
tomponline: TODO?

lxd/daemon.go
```diff
@@ -641,84 +638,25 @@ func (d *Daemon) Authenticate(w http.ResponseWriter, r *http.Request) (*request.
...
-			})
-		})
-		if err != nil {
-			return fmt.Errorf("Failed to update OIDC identity information: %w", err)
-		}
-
-		action = lifecycle.IdentityUpdated
+// handleOIDCAuthenticationResult checks the identity cache for the OIDC identity by their email address.
+// If no identity is found, the cache is refreshed.
+// This is for new OIDC logins and is currently required for authorization to work because group membership is saved to the cache.
```
tomponline, line 643: and if group membership changes does this trigger a cluster notification to refresh the cache?

lxd/oidc_sessions.go
```diff
@@ -0,0 +1,319 @@
...
+			return response.BadRequest(fmt.Errorf("Bad session ID: %w", err))
+		}
+
+		var session *dbCluster.OIDCSession
+		err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+			session, err = dbCluster.GetOIDCSessionByUUID(ctx, tx.Tx(), sessionID)
+			return err
+		})
+		if err != nil {
+			if api.StatusErrorCheck(err, http.StatusNotFound) {
```
tomponline, line 79: why not response.SmartError here so that the DB level "Session not found" bubbles up rather than generic not found error?

lxc/main.go
```diff
@@ -466,7 +467,8 @@ Or for a virtual machine: lxc launch ubuntu:24.04 --vm`)
 		}
 	}
 
-	// Set the user agent
+	// Set the user agent, indicating that we are able to store cookies.
+	version.UserAgentFeatures([]string{api.ClientFeatureCookieJar})
```
tomponline: @markylaing shall we update this version.UserAgentFeatures function to return an error if the features container spaces?

lxd/auth/oidc/oidc.go
```diff
@@ -146,7 +148,16 @@ func (o *Verifier) userInfo(ctx context.Context, token string) (*oidc.UserInfo,
...
 // starts a new session.
 func (o *Verifier) authenticateBearerToken(r *http.Request, w http.ResponseWriter, accessToken string) (*AuthenticationResult, error) {
-	err := o.ensureConfig(r.Context(), r.Host)
+	agent, err := request.ParseUserAgent(r.UserAgent())
+	if err != nil {
+		return nil, api.StatusErrorf(http.StatusBadRequest, "Failed to parse user agent to determine cookie capability: %w", err)
+	}
+
+	if !slices.Contains(agent.Features, api.ClientFeatureCookieJar) {
+		return nil, api.StatusErrorf(http.StatusBadRequest, "OIDC authentication requires cookie support. Please update your client")
```
tomponline: requires persistent cookie support

lxd/auth/oidc/oidc.go
```diff
@@ -146,7 +148,16 @@ func (o *Verifier) userInfo(ctx context.Context, token string) (*oidc.UserInfo,
 // authenticateBearerToken calls the /userinfo endpoint with the given token to retrieve information about the user. Then
 // starts a new session.
 func (o *Verifier) authenticateBearerToken(r *http.Request, w http.ResponseWriter, accessToken string) (*AuthenticationResult, error) {
-	err := o.ensureConfig(r.Context(), r.Host)
+	agent, err := request.ParseUserAgent(r.UserAgent())
+	if err != nil {
+		return nil, api.StatusErrorf(http.StatusBadRequest, "Failed to parse user agent to determine cookie capability: %w", err)
```
tomponline: persistent cookie feature

lxd/oidc_sessions.go
```diff
@@ -317,3 +324,76 @@ func oidcSessionDelete(d *Daemon, r *http.Request) response.Response {
...
+	f := func(ctx context.Context) {
+		s := stateFunc()
+
+		leaderInfo, err := s.LeaderInfo()
+		if err != nil {
+			logger.Error("Failed to get leader cluster member address", logger.Ctx{"err": err})
+			return
+		}
+
+		if !leaderInfo.Clustered {
```
tomponline: why would we return if not clustered?

shared/api/client.go
kadinsayani: > if another client feature is added, it's clear where to find it.

Makes sense, thanks.

lxd/oidc_sessions.go
markylaing: Whoops, will fix

markylaing: Whoops, will fix

markylaing: Whoops, will fix

lxc/main.go
markylaing: Sounds good, will fix

lxd/cluster/config/config.go
markylaing: Ahhh I forgot about this. We need to validate that the session expiry is not longer than the Auth secret expiry. Else we could be in a situation where the token was signed by a key derived from a secret that is too old.

Will fix

lxd/auth/oidc/oidc.go
markylaing: It's needed so that the database package is not imported by the `auth/oidc` package. IMO this is good for keeping imports simple in case we ever want to use this OIDC logic elsewhere.

lxd/db/oidc/session.go
markylaing: I wanted to make it clear that in general, the session handler is responsible for setting the expiry, but the caller can override this only when necessary.

lxd/auth/bearer/bearer.go
markylaing: These are now wrapped. I couldn't figure out why I didn't wrap in the first place.

lxd/db/cluster/identities.go
markylaing: Have removed the function, it is unused after this PR anyway.

lxd/db/cluster/oidc_sessions.go
markylaing: Have used cluster.RegisterStmt for one query that is used on the hot path for authentication

lxd/db/oidc/session.go
markylaing: Changed to `remoteAddr`.

lxd/db/oidc/session.go
markylaing: done.

lxd/db/oidc/session.go
markylaing: Fixed.

lxd/auth/bearer/bearer.go
markylaing: This is now wrapped

lxd/auth/bearer/bearer.go
markylaing: I've updated the function comment.

lxd/db/cluster/oidc_sessions.go
markylaing: Need to parse the UUID here as the session ID is a v7 UUID. It is used to populated the `created_at` field for the session. I'll wrap the error.

lxd/auth/oidc/oidc.go
markylaing: Done

lxd/auth/oidc/oidc.go
markylaing: Done

lxd/cluster/config/config.go
markylaing: Fixed. Also updated the `serverconfig.sh` suite to test this.

lxd/daemon.go
markylaing: Yes, this is in the `auth_groups` API handlers.

lxc/main.go
markylaing: Fixed in https://github.com/canonical/lxd/pull/16625 and will update this PR when merged.

lxd/db/cluster/oidc_sessions.go
markylaing: It turns out that `uuid.UUID` implements the `sql.Scanner` and `driver.Valuer` interfaces to convert the UUID into a string in the database. This allows us to use it directly with the DB and means we don't need to do the conversion at this point.

lxd/oidc_sessions.go
markylaing: Fixed.

markylaing:
I believe I have addressed everything and/or replied to any comments, so ready for re-review when you are. Thanks

tomponline (dismissed):
Great work thanks!

One question, why do we still add the jujue github.com/juju/go4 dep?

https://github.com/canonical/lxd/pull/16476/commits/a294cfc946a3d5db8fdd10621bcf498443aad4fe#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R110

tomponline:
@markylaing one test failed with:

>  jq -e '."127.0.0.1"."127.0.0.1;/;session" | .Name == "session" and .Path == "/" and .Secure and .HttpOnly and .Persistent and .HostOnly and .Domain == "127.0.0.1" and .SameSite == "SameSite=Strict"' /tmp/lxd-test.tmp.X1m0/ZJy/jars/

flaky?

markylaing:
> Great work thanks!
> 
> One question, why do we still add the jujue github.com/juju/go4 dep?
> 
> [a294cfc#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R110](https://github.com/canonical/lxd/commit/a294cfc946a3d5db8fdd10621bcf498443aad4fe#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R110)

Ahh sorry should be able to drop that now

markylaing:
> @markylaing one test failed with:
> 
> > jq -e '."127.0.0.1"."127.0.0.1;/;session" | .Name == "session" and .Path == "/" and .Secure and .HttpOnly and .Persistent and .HostOnly and .Domain == "127.0.0.1" and .SameSite == "SameSite=Strict"' /tmp/lxd-test.tmp.X1m0/ZJy/jars/
> 
> flaky?

Interesting. Not sure what happened there. Will look into it.

markylaing:
> > @markylaing one test failed with:
> > > jq -e '."127.0.0.1"."127.0.0.1;/;session" | .Name == "session" and .Path == "/" and .Secure and .HttpOnly and .Persistent and .HostOnly and .Domain == "127.0.0.1" and .SameSite == "SameSite=Strict"' /tmp/lxd-test.tmp.X1m0/ZJy/jars/
> > 
> > 
> > flaky?
> 
> Interesting. Not sure what happened there. Will look into it.

I was able to repeat this reliably by spamming calls to an OIDC remote (which will repeatedly read from and write to the cookie jar). 

Instead of opening the existing file in write mode (without append), the solution was to always call `os.Create` and truncate the file. We're always fully overwriting the contents and discarding any old data, so I'm not sure why I made it more complicated than it needed to be.

I've also removed the dependency update which added the go4 library.

markylaing:
Will unmark as draft when I see a clean test run

markylaing:
@tomponline ok now I think it's done :rocket:

tomponline:
@markylaing what uses golang.org/x/net/publicsuffix ?

lxc/cookiejar/jar.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil, fmt.Errorf("Failed to open jar: %w", err)
+	}
+
+	return jar, nil
+}
+
+// newJar creates a new [Jar] but does not attempt to read existing contents of the jar.
+func newJar(filepath string) (*Jar, error) {
+	// Use the default public suffix list.
+	jar, err := cookiejar.New(&cookiejar.Options{PublicSuffixList: publicsuffix.List})
```
tomponline, line 36: @markylaing I see this is what needs the publicsuffix list, but why is it needed?

markylaing: If you look at the [godoc](https://pkg.go.dev/net/http/cookiejar#New) for `net/http/cookiejar`, in the example under `New` it says "All users of cookiejar should import "golang.org/x/net/publicsuffix".

Also see the comment under [`Options`](https://pkg.go.dev/net/http/cookiejar#Options) where it says a `nil` value isn't secure.

tomponline: Yep I saw that, but I was wondering whether we can set this dynamically to the domain of the LXD remote?

markylaing: Oh... that is a good idea! I'll create a follow up issue to investigate what we can do here.

PR #16478, opened 2025-09-12
Add "volatile.cluster.group" config key

kadinsayani:
Includes cherry-picks from https://github.com/lxc/incus/pull/696.

This pull request introduces a new configuration key, `volatile.cluster.group`, which records the target cluster group specified during instance creation or migration and ensures instances are scheduled to the correct cluster group during events such as evacuation.

### DIFF
--- a/lxd/instance_post.go
+++ b/lxd/instance_post.go
@@ -645,10 +645,6 @@ func instancePostClusteringMigrate(ctx context.Context, s *state.State, srcPool 
 		return nil, errors.New("The destination cluster member is evacuated")
 	}
 
-	// Save the original value of the "volatile.apply_template" config key,
-	// since we'll want to preserve it in the copied instance.
-	origVolatileApplyTemplate := srcInst.LocalConfig()["volatile.apply_template"]
-
 	// Check we can convert the instance to the volume types needed.
 	volType, err := storagePools.InstanceTypeToVolumeType(srcInst.Type())
 	if err != nil {
@@ -812,23 +808,6 @@ func instancePostClusteringMigrate(ctx context.Context, s *state.State, srcPool 
 				err = tx.UpdateInstanceConfig(int(id), map[string]string{"volatile.cluster.group": targetGroupName})
 				if err != nil {
 					return fmt.Errorf(`Failed setting "volatile.cluster.group" config key: %w`, err)
-				}
-			}
-
-			// Restore the original value of "volatile.apply_template".
-			err = tx.DeleteInstanceConfigKey(ctx, id, "volatile.apply_template")
-			if err != nil {
-				return fmt.Errorf("Failed to remove volatile.apply_template config key: %w", err)
-			}
-
-			if origVolatileApplyTemplate != "" {
-				config := map[string]string{
-					"volatile.apply_template": origVolatileApplyTemplate,
-				}
-
-				err = tx.CreateInstanceConfig(ctx, int(id), config)
-				if err != nil {
-					return fmt.Errorf("Failed to set volatile.apply_template config key: %w", err)
 				}
 			}
 


lxd/instance/instancetype/instance.go
```diff
@@ -452,6 +452,13 @@ var InstanceConfigKeysAny = map[string]func(value string) error{
 	//  shortdesc: The origin of the evacuated instance
 	"volatile.evacuate.origin": validate.IsAny,
 
+	// lxdmeta:generate(entities=instance; group=volatile; key=volatile.cluster.target)
+	// The target cluster group at instance creation or migration time. This is used during scheduling events such as evacuation to ensure the instance is placed correctly.
+	// ---
+	// type: string
+	// shortdesc: The target cluster group
```
minaelee, line 472: Is there a default to note here if unset, such as the `default` group?

kadinsayani: There is no default value, it is only set when `--target` is used to create or move an instance.

test/suites/clustering.sh
```diff
@@ -3543,74 +3543,97 @@ test_clustering_evacuation() {
...
+  LXD_DIR="${LXD_TWO_DIR}" lxc start c7
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc info c7 | grep -cF "Location: node1")" = 1 ]
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc info c7 | grep -cF "Status: RUNNING")" = 1 ]
+
   # Evacuate first node
   LXD_DIR="${LXD_TWO_DIR}" lxc cluster evacuate node1 --force
 
   # Ensure the node is evacuated
-  LXD_DIR="${LXD_TWO_DIR}" lxc cluster show node1 | grep -F "status: Evacuated"
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc cluster show node1 | grep -cF "status: Evacuated")" = 1 ]
```
simondeziel: Looks more bloated to me, what's the benefit?

test/suites/clustering.sh
```diff
@@ -3543,74 +3543,97 @@ test_clustering_evacuation() {
...
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc info c4 | grep -cF "Status: RUNNING")" = 1 ]
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc info c4 | grep -cF "Location: node1")" = 0 ]
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc info c5 | grep -cF "Status: STOPPED")" = 1 ]
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc info c5 | grep -cF "Location: node1")" = 0 ]
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc info c6 | grep -cF "Status: RUNNING")" = 1 ]
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc info c6 | grep -cF "Location: node2")" = 1 ]
+
+  # Check "volatile.target.cluster" (initial target cluster group) is respected.
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc info c7 | grep -cF "Status: RUNNING")" = 1 ]
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc info c7 | grep -cF "Location: node3")" = 1 ]
```
simondeziel: This was reworked for fewer `lxc` calls in `main` but I am now wondering if we shouldn't go with a single `lxc list -f json > list.json` followed by a bunch of `jq -e 'assertion checks for each instance' < list.json`.

test/suites/clustering.sh
kadinsayani: I thought it was preferred to use shell comparison in general. I'll drop that change.

kadinsayani: Do we just want to avoid `grep -xF`?

simondeziel: It's just the counting occurrences that I find risky. `grep -xF` is something I consider pretty safe so it's the recommended way to look for a specific line **IF** multiple lines are provided.

If a single line is what you get, then shell comparison is the safest bet IMHO.

kadinsayani: Gotcha, thanks for clarifying :)

test/suites/clustering.sh
```diff
@@ -3545,22 +3559,27 @@ test_clustering_evacuation() {
   # For debugging
   LXD_DIR="${LXD_TWO_DIR}" lxc list -c nsL
 
+  # Check "volatile.target.cluster" (initial target cluster group) is respected.
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c sL c7)" = "RUNNING,node3" ]
+
   # Check instance status
   [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c s  c1)" = "RUNNING" ]
-  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -L    c1)" != "node1" ]
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c L  c1)" != "node1" ]
```
simondeziel, line 3577: Ah, you picked that one too! Thanks and sorry for the broken test I added :/

kadinsayani: No worries!

kadinsayani: It's easy to miss those.

test/suites/clustering.sh
kadinsayani: I'll leave that call to you :)

Something to consider is that if we start using that pattern across different test functions it could become a source for error if not cleaned up properly.

test/suites/clustering.sh
simondeziel: Yeah, I haven't yet found a way to make those negative string comparisons safe(r) to use :/

test/suites/clustering.sh
simondeziel: Yep, good point. I guess this could be mitigated by saving the JSON in a `local` variable and then use `jq -e '...' <<< "${JSON}"`.

I was thinking of doing the repeated `jq -e` checks only when replacing many "read" operations that just inspect various things. It would not be widely used.

kadinsayani: I'll leave as is for now but we can consider your suggested approach for a separate PR :)

lxd/instance/instancetype/instance.go
```diff
@@ -452,6 +452,13 @@ var InstanceConfigKeysAny = map[string]func(value string) error{
...
+	"volatile.cluster.target": validate.IsAny,
```
tomponline: should we validate this starts with an `@` seeing as we are only recording the group?

lxd/api_cluster.go
```diff
@@ -4497,6 +4485,15 @@ func evacuateClusterSelectTarget(ctx context.Context, s *state.State, candidateM
 		return nil, err
 	}
 
+	// Find the least loaded cluster member which supports the instance's architecture.
+	err = s.DB.Cluster.Transaction(ctx, func(ctx context.Context, tx *db.ClusterTx) error {
```
tomponline: @kadinsayani why are we starting a new transaction here rather than doing the query for `GetNodeWithLeastInstances` in the transaction just above?

lxd/instance_post.go
```diff
@@ -791,12 +797,25 @@ func instancePostClusteringMigrate(ctx context.Context, s *state.State, srcPool
...
 
-			// Restore the original value of "volatile.apply_template".
 			id, err := dbCluster.GetInstanceID(ctx, tx.Tx(), projectName, newInstName)
 			if err != nil {
 				return fmt.Errorf("Failed to get ID of moved instance: %w", err)
 			}
 
+			// Set the cluster group record if needed.
+			if targetGroupName != "" {
+				err = tx.DeleteInstanceConfigKey(ctx, id, "volatile.cluster.target")
```
tomponline: @kadinsayani would you mind updating the pattern used in this function of `DeleteInstanceConfigKeys` followed by `CreateInstanceConfig`, and replace it with using the `UpdateInstanceConfig` function, as this will use `INSERT OR REPLACE INTO` to avoid 2 queries.

tomponline: Also, is there a reason we cant inject the group name into the config provided in `InstancePut: srcInstInfo.Writable(),` above, as that contains the source instance's config, could we modify `InstancePut.Config` afterwards?

Or are the volatile keys being stripped during migration, and we need to put them back because this is an intra-cluster migration?

Basically I'm looking at this section where we are updating the instance after it has been migrated to put back some volatile config keys and I'm wondering why they are getting stripped out/changed in the first place.

We have the concept of `ClusterMoveSourceName` now plumbed through the migration subsystem (you can see it being set in the call to `newMigrationSource()` in this function and passing in the source instance's name.

This concept is defined in `MigrateArgs` as:

> ClusterMoveSourceName // Will be empty if not a cluster move, othwise indicates the source instance.

Oh there's a typo in the comment in the comment "othwise", please can you fix too :)

If we can update the migration receiver to detect a "cluster move" type migration, then perhaps we can relax the volatile stripping logic on the receiver side and remove this whole section on the sending side (avoiding all the DB queries).

tomponline: Sorry this was a bit rambling as my thoughts were evolving as I typed it, so here's the ideas more concisely:

1. Dont issue Delete then Insert queries, just use a single Replace query.
2. Can we avoid the Replace query by inserting the original value into the migration request config?
3. Probably not, as its likely already there and being stripped on the receiver side (migrations do strip some volatile config because if you're sending an instance to another host in a different cluster not all volatile keys make sense to be kept - `volatile.cluster.target` being one of them).
4. But, as the migration receiver can now differentiate a migration between standalone LXD servers and a "cluster move" type migration, we should be able to avoid doing all of these queries, and instead relax the volatile config key stripping somewhat on the receiver side.

In that way we wont need to undo the stripping as it wont have happened.

tomponline (changes requested):
Thanks. Lets take this opportunity to remove the hacky "restore volatile keys that were stripped after migration" and see if we can update the receiver to avoid stripping them in the first place when doing a cluster move.

lxd/api_cluster.go
kadinsayani: We can perform the query in the earlier transaction.

lxd/instance_post.go
kadinsayani: For `volatile.apply_template`, we restore the value because the template is applied when the instance starts. After applying the template, the key is removed.

For `volatile.cluster.target`, we want to set the value to any new target group (if we have one) during an instance move. That way we aren't stuck with the original target group if a new one is provided.

tomponline: > For volatile.apply_template, we restore the value because the template is applied when the instance starts. After applying the template, the key is removed.

Question is, why do we do that if its already been applied, or has it not, and if not why was it removed?

> For volatile.cluster.target, we want to set the value to any new target group (if we have one) during an instance move. That way we aren't stuck with the original target group if a new one is provided.

Right makes sense. But where is the decision for instance placement taking place? And why isn't the recorded cluster group recorded there?

kadinsayani: > Right makes sense. But where is the decision for instance placement taking place? And why isn't the recorded cluster group recorded there?

`evacuateClusterSelectTarget()` is where we filter candidate members. That's where we want to use `volatile.cluster.target`, so if an instance is created or moved with `--target` specified, we respect that targeted cluster group during evacuation.

kadinsayani: I've added a check to avoid updating `volatile.apply_template` during intra-cluster moves, since the migration receiver preserves the original value:

```Go
			// Update "volatile.apply_template".
			// For intra-cluster moves, the migration receiver preserves the original value,
			// so we only need to restore it for cross-cluster moves.
			config := map[string]string{}
			if srcMigration.clusterMoveSourceName == "" && origVolatileApplyTemplate != "" {
				config["volatile.apply_template"] = origVolatileApplyTemplate

				err = tx.UpdateInstanceConfig(int(id), config)
				if err != nil {
					return fmt.Errorf(`Failed setting "volatile.apply_template" config key: %w`, err)
				}
			}
```

lxd/instance/instancetype/instance.go
```diff
@@ -462,6 +462,20 @@ var InstanceConfigKeysAny = map[string]func(value string) error{
 	//  shortdesc: The origin of the evacuated instance
 	"volatile.evacuate.origin": validate.IsAny,
 
+	// lxdmeta:generate(entities=instance; group=volatile; key=volatile.cluster.target)
+	// The target cluster group at instance creation or migration time. This is used during scheduling events such as evacuation to ensure the instance is placed correctly.
+	// ---
+	// type: string
+	// shortdesc: The target cluster group
+	"volatile.cluster.target": func(value string) error {
+		targetGroupPrefix := "@"
```
tomponline: We shouldnt define `targetGroupPrefix` again, but instead move the one in `main` into this package so it can be used throughout the code base.

lxd/api_cluster.go
```diff
@@ -3404,31 +3404,11 @@ func evacuateInstances(ctx context.Context, opts evacuateOpts) error {
...
-		targetMemberInfo, err := evacuateClusterSelectTarget(ctx, opts.s, candidateMembers)
+		// Find a new location for the instance.
+		targetMemberInfo, err := evacuateClusterSelectTarget(ctx, opts.s, inst)
 		if err != nil {
 			if api.StatusErrorCheck(err, http.StatusNotFound) {
-				// Skip migration if no target is available
+				// Skip migration if no target is available.
 				l.Warn("No migration target available for instance")
 				continue
 			}
```
tomponline, line 3412: what happens if `err != nil` AND it is not an http.StatusNotFound error?

Do we proceed to use a nil `targetMemberInfo` and crash?

lxd/api_cluster.go
```diff
@@ -4479,9 +4479,25 @@ func evacuateClusterSelectTarget(ctx context.Context, s *state.State, inst insta
 			return fmt.Errorf("Failed getting cluster members: %w", err)
 		}
 
+		// Filter candidates by group if needed.
+		_, group := shared.TargetDetect(inst.LocalConfig()["volatile.cluster.target"])
```
tomponline: It looks like `TargetDetect` is only used by LXD, so this would be a good opportunity to move it out of `shared` and update it to use the targetGroupPrefix constant too.

lxd/api_cluster.go
```diff
@@ -3404,31 +3404,11 @@ func evacuateInstances(ctx context.Context, opts evacuateOpts) error {
...
-
-			return nil
-		})
-		if err != nil {
-			return err
-		}
-
-		targetMemberInfo, err := evacuateClusterSelectTarget(ctx, opts.s, candidateMembers)
+		// Find a new location for the instance.
+		targetMemberInfo, err := evacuateClusterSelectTarget(ctx, opts.s, inst)
```
tomponline, line 3406: @kadinsayani iiuc correctly, if an instance has a cluster group specified, and no members within that cluster group can take this instance during evacuation, then the instance is left on the member being evacuated.

This is fine for now.

Looking forward, I would like us to ensure that the evacuation policy is consistent when using either cluster group targeting or placement group targeting, so it would be worth keeping this in mind so that we dont end up having LXD do different things for different modes.

I believe your spec was leaning towards evacuating instances to a members not considered candidates for the desired cluster/placement group in the event that no preferable candidates were available.

We may want to reconsider this approach.

tomponline (commented):
In Incus there is this:

https://github.com/lxc/incus/blob/f9d5bb51246ce05b01b7e3bfc8892ceb7d2df557/cmd/incusd/instance_post.go#L594-L600

Which in LXD is equivalent to `instancePostClusteringMigrateWithRemoteStorage`.

So I think we need to consider whether we need this part too.

lxd/instance_post.go
```diff
@@ -796,25 +802,29 @@ func instancePostClusteringMigrate(ctx context.Context, s *state.State, srcPool
...
 			}
 
-			if origVolatileApplyTemplate != "" {
-				config := map[string]string{
-					"volatile.apply_template": origVolatileApplyTemplate,
-				}
+			// Update "volatile.apply_template".
+			// For intra-cluster moves, the migration receiver preserves the original value,
+			// so we only need to restore it for cross-cluster moves.
+			config := map[string]string{}
```
tomponline: do we need to define this outside of the `if`?

lxd/instance_post.go
```diff
@@ -796,25 +802,29 @@ func instancePostClusteringMigrate(ctx context.Context, s *state.State, srcPool
...
+			if srcMigration.clusterMoveSourceName == "" && origVolatileApplyTemplate != "" {
```
tomponline: In what scenario does this function `instancePostClusteringMigrate()` get called when its not an intra-cluster move? I think this comment could be clarified to explain.

lxd/api_cluster.go
kadinsayani: Yes, that would cause a panic. We should return an `err` in this case to avoid a nil pointer dereference.

lxd/instance_post.go
kadinsayani: Nope, we can move this in.

lxd/api_cluster.go
kadinsayani: > iiuc correctly, if an instance has a cluster group specified, and no members within that cluster group can take this instance during evacuation, then the instance is left on the member being evacuated.

Your assumption is correct, if there are no candidate members, that instance will be skipped (unless we encounter the situation discussed in https://github.com/canonical/lxd/pull/16478#discussion_r2433123848). 

> Looking forward, I would like us to ensure that the evacuation policy is consistent when using either cluster group targeting or placement group targeting, so it would be worth keeping this in mind so that we dont end up having LXD do different things for different modes.

Agreed, evacuation should be consistent with cluster group targeting and placement groups. 

> I believe your spec was leaning towards evacuating instances to a members not considered candidates for the desired cluster/placement group in the event that no preferable candidates were available.

> We may want to reconsider this approach.

For now, I think we can align placement group evacuation behavior with current cluster group targeting behavior, logging a warning and leaving the instance on the member being evacuated.

In the future, rather than take a stance on what to do during evacuation when there are no candidate members (regardless of whether cluster group targeting or placement group policy is used), perhaps we can introduce a new evacuation action. The action can determine whether we skip the instance and leave it on member being evacuated, schedule it wherever we can (member with fewest instances), or fail.

lxd/api_cluster.go
kadinsayani: Thanks for the suggestion!

lxd/api_cluster.go
tomponline: sounds good for now.

lxd/instance_post.go
kadinsayani: I've removed this part entirely. It turns out `(lxc/qemu).MigrateReceive()` already has logic to preserve `volatile.apply_template` during cluster moves:

https://github.com/canonical/lxd/blob/94edcfbf7259419e9e78112726404973999454a9/lxd/instance/drivers/driver_qemu.go#L7657-L7662

https://github.com/canonical/lxd/blob/94edcfbf7259419e9e78112726404973999454a9/lxd/instance/drivers/driver_lxc.go#L5421-L5426

kadinsayani:
> In Incus there is this:
> 
> https://github.com/lxc/incus/blob/f9d5bb51246ce05b01b7e3bfc8892ceb7d2df557/cmd/incusd/instance_post.go#L594-L600
> 
> Which in LXD is equivalent to `instancePostClusteringMigrateWithRemoteStorage`.
> 
> So I think we need to consider whether we need this part too.

Thanks for pointing that out. I've added that part.

lxd/api_cluster.go
kadinsayani: I've added a note to the spec for this :)

lxd/instance_post.go
tomponline: Excellent! Thanks

test/suites/clustering.sh
```diff
@@ -3546,9 +3546,23 @@ test_clustering_evacuation() {
   LXD_DIR="${LXD_ONE_DIR}" lxc launch testimage c6 --target=node2
   LXD_DIR="${LXD_ONE_DIR}" lxc config set c6 boot.host_shutdown_timeout=1
 
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster group create foo
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster group assign node3 default,foo
+
+  LXD_DIR="${LXD_ONE_DIR}" lxc launch testimage c7 --target=@foo
+  LXD_DIR="${LXD_ONE_DIR}" lxc config set c7 boot.host_shutdown_timeout=1
+
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc config get c7 volatile.cluster.group)" = "@foo" ]
```
tomponline: Worth checking this is on node3?

test/suites/clustering.sh
```diff
@@ -3546,9 +3546,23 @@ test_clustering_evacuation() {
...
+
   # For debugging
   LXD_DIR="${LXD_TWO_DIR}" lxc list -c nsL
 
+  # Move c7 to node1 to test "volatile.cluster.group" handling post-evacuation.
+  LXD_DIR="${LXD_TWO_DIR}" lxc stop c7 -f
```
tomponline: Why are you doing stop, move, start rather than just move?

test/suites/clustering.sh
```diff
@@ -3546,9 +3546,23 @@ test_clustering_evacuation() {
...
+
+  LXD_DIR="${LXD_ONE_DIR}" lxc launch testimage c7 --target=@foo
+  LXD_DIR="${LXD_ONE_DIR}" lxc config set c7 boot.host_shutdown_timeout=1
+
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc config get c7 volatile.cluster.group)" = "@foo" ]
+
   # For debugging
   LXD_DIR="${LXD_TWO_DIR}" lxc list -c nsL
 
+  # Move c7 to node1 to test "volatile.cluster.group" handling post-evacuation.
```
tomponline: This could be clarified with not just the *what* but also explaining what it represents (a node outside of the target group).

test/suites/clustering.sh
kadinsayani: Yeah, thanks for suggesting :)

test/suites/clustering.sh
kadinsayani: I've switched to just move. No need to start this instance.

test/suites/clustering.sh
kadinsayani: Ack, added.

kadinsayani:
@tomponline tests passing ✅

test/suites/clustering.sh
```diff
@@ -3546,9 +3546,22 @@ test_clustering_evacuation() {
   LXD_DIR="${LXD_ONE_DIR}" lxc launch testimage c6 --target=node2
   LXD_DIR="${LXD_ONE_DIR}" lxc config set c6 boot.host_shutdown_timeout=1
 
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster group create foo
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster group assign node3 default,foo
+
+  LXD_DIR="${LXD_ONE_DIR}" lxc init testimage c7 --target=@foo
+  LXD_DIR="${LXD_ONE_DIR}" lxc config set c7 boot.host_shutdown_timeout=1
```
simondeziel: Is `c7` ever started? If not, could we drop that `boot.host_shutdown_timeout`?

kadinsayani: It's not started. I dropped that bit, thanks :)

tomponline:
@kadinsayani please can you also look at this https://github.com/lxc/incus/pull/2571 related to this PR. Thanks

tomponline:
@kadinsayani also https://github.com/lxc/incus/pull/2569 seems pertinent whilst you're working on migration refactors (the first commit)?

kadinsayani:
Thanks @tomponline, I'll take a look this week :)

PR #16482, opened 2025-09-15
WIP: test delete-database branch

marco6:
Do not merge.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,7 @@ dqlite:
 
 	cd "$(DQLITE_PATH)" && \
 		autoreconf -i && \
-		./configure --enable-build-raft && \
+		./configure && \
 		make -j
 
 .PHONY: liblxc
--- a/.github/actions/cache-dqlite-liblxc/action.yml
+++ b/.github/actions/cache-dqlite-liblxc/action.yml
@@ -13,26 +13,35 @@ runs:
         ARCH="$(dpkg --print-architecture)"
         DATE="$(date --utc '+%Y%m%d')"
         YESTERDAY="$(date --utc '+%Y%m%d' --date=yesterday)"
+        DQLITE_REV="$(cat /home/runner/work/lxd/lxd-test/vendor/dqlite/.gitref)"
 
         # i.e: deps-ubuntu-24.04-amd64-${DATE}
-        echo "KEY=deps-${ID}-${VERSION_ID}-${ARCH}-${DATE}" >> $GITHUB_OUTPUT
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -157,7 +157,7 @@ dqlite:
 	cd "$(DQLITE_PATH)" && \
 		autoreconf -i && \
 		./configure --enable-build-raft && \
-		make
+		make -j
```
simondeziel, line 160: Sounds like something we should always have. I'll scoop this one in, thanks!

simondeziel:
> Do not merge.

I see not problem in using those test PRs personally but maybe you'd be better served with an easy way to replicate locally in a VM?

If you are interested, we have this documented [here](https://documentation.ubuntu.com/lxd/latest/contributing/#development-environment-setup) but here's the distilled version hoping to speed things up for you.

You first need LXD installed (if not `sudo snap install lxd` then `sudo lxd init --auto`) and a custom profile:

```
# You need to modify this
cd path/to/lxd/git

# this needs to be run from inside the git repostory
GIT_ROOT="$(git rev-parse --show-toplevel)"
# create or edit the profile based on the provided template
lxc profile list | grep -qwF lxd-test || lxc profile create lxd-test
sed "s|@@PATH_TO_LXD_GIT@@|${GIT_ROOT}|" "${GIT_ROOT}/doc/lxd-test.yaml" | lxc profile edit lxd-test
```

Then create a test VM using it:

```
lxc launch ubuntu-minimal-daily:24.04 v1 --vm -p lxd-test
sleep 30
# this may take a while as many packages need to be installed
lxc exec v1 -- cloud-init status --wait --long
```

Then you are ready to build LXD+dqlite in it:

```
# start a root shell in the VM
lxc exec v1 -- bash

# go into the git repo
cd lxd

# build deps and LXD binaries
make deps && make

# get an interactive test shell session with all the needed environment variables to use and test LXD
make test-shell

# run the just the problematic test
LXD_VERBOSE=server LXD_INSPECT=1 ./main.sh clustering_remove_raft_node
```

That's it, if you have questions please let me know.

tomponline:
@marco6 good to close this now?

marco6:
@simondeziel thanks for the guide. The problem is that I couldn't reproduce this on my machine and I really need to be sure about what the workflow logic is doing. Also, those tests take a long time and it is nice for me not to clog my CPU while they run.

Regardless, the new caching mechanism is using the cache as a requirement instead of as a speedup, so I probably can't easily trigger a new build after an update. I think the best I can do for now is to watch out for failures in main and close this.

tomponline:
@marco6 you're welcome to keep this open, I only thought perhaps you didn't' need it because you found the issue https://github.com/canonical/dqlite/pull/846 ? 

:)

PR #16483, opened 2025-09-15
Go: Bump minimum version to 1.24.6 and update dependencies (stable-5.0)

tomponline:
Documentation issues are being worked on separately by myself and @minaelee


lxd/instances_get.go
```diff
@@ -236,7 +236,7 @@ func instancesGet(d *Daemon, r *http.Request) response.Response {
 	}
 
 	logger.Debugf("DBERR: containersGet, db is locked")
-	logger.Debugf(logger.GetStack())
+	logger.Debug(logger.GetStack())
```
simondeziel, line 239: Weird that the previous line didn't need to lose its `f` as well.

PR #16502, opened 2025-09-15
Pure/HPE Alletra: ISCSI/multipath mode volume resize fixes

mihalicyn:
Requires https://github.com/canonical/lxd-pkg-snap/pull/921

- [x] HPE Alletra tested
- [ ] Pure tested

### DIFF
--- a/lxd/storage/drivers/driver_alletra_volumes.go
+++ b/lxd/storage/drivers/driver_alletra_volumes.go
@@ -1356,6 +1356,11 @@ func (d *alletra) SetVolumeQuota(vol Volume, size string, allowUnsafeResize bool
 
 			defer cleanup()
 
+			err = block.RefreshDiskDeviceSize(d.state.ShutdownCtx, devPath)
+			if err != nil {
+				return fmt.Errorf("Failed refreshing volume %q size: %w", vol.name, err)
+			}
+
 			// Always wait for the disk to reflect the new size.
 			// In case SetVolumeQuota is called on an already mapped volume,
 			// it might take some time until the actual size of the device is reflected on the host.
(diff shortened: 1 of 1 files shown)


lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -891,25 +906,30 @@ func (d *pure) SetVolumeQuota(vol Volume, size string, allowUnsafeResize bool, o
...
 		}
 
-		// Move the VM GPT alt header to end of disk if needed (not needed in unsafe resize mode as it is
-		// expected the caller will do all necessary post resize actions themselves).
-		if vol.IsVMBlock() && !allowUnsafeResize {
-			devPath, cleanup, err := d.getMappedDevPath(vol, true)
-			if err != nil {
-				return err
-			}
+		devPath, cleanup, err := d.getMappedDevPath(vol, true)
```
MusicDin, line 909: Should we do disk device refresh only if mode is `iSCSI`? This way we avoid unnecessary volume mapping for NVME/tcp

mihalicyn:
From my side I can confirm that with this patches applied HPE Alletra driver passes all the tests:
```
export LXD_SIDELOAD_PATH=/home/user1/lxd-ci/lxd.debug
export ALLETRA_WSAPI=https://<redacted>
export ALLETRA_WSAPI_VERIFY=false
export ALLETRA_USERNAME=<redacted>
export ALLETRA_PASSWORD=<redacted>
export ALLETRA_CPG=SSD_r6
export ALLETRA_MODE=iscsi

snap remove lxd --purge
./bin/local-run tests/storage-vm latest/edge alletra

snap remove lxd --purge
./bin/local-run tests/storage-volumes-vm latest/edge alletra
```

lxd/storage/drivers/driver_pure_volumes.go
mihalicyn: Yeah, I agree, but we should look closer into this place, because as you can see I have also added a missing call to `block.WaitDiskDeviceResize` which we had in all the other cases. My point is that we should always call `block.WaitDiskDeviceResize` when block device is present, isn't it?

tomponline: > My point is that we should always call block.WaitDiskDeviceResize when block device is present, isn't it?

I see no reason why we shouldnt do this.

PR #16508, opened 2025-09-15
gomod: Update go-dqlite (stable-4.0)

tomponline:


.github/workflows/tests.yml
```diff
@@ -450,7 +445,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.22.x
+          go-version: 1.24.x
```
simondeziel: Go being available on all GH runners, we could also drop this `Install Go` step altogether like with did in `main`'s `.github/actions/lp-snap-build/action.yml`

PR #16517, opened 2025-09-16
lxd/storage/drivers/pure: fix block device unmap logic for iSCSI

mihalicyn:
We found that our previous approach with iSCSI volumes (in multipath mode) unmap doesn't work, because we have a race condition with systemd-udevd which tries to access process a device changes and as a part of it also tries to access a block device /dev/dm-X. But as long as by that time we have removed a vLUN from the array, this access attempt traps the process to get into D-state. A clean solution is to:
1. remove a block device first
2. remove vLUN

Also, for multipath device we can't just remove the device, we have to use a command "multipath -f /dev/dm-X" to do that. Because otherwise we also at risk of stepping on our tail. Because to remove a multipath device we need to remove it slaves (backing block devices), and once all of them are removed then /dev/dm-X device becomes inaccessible too. And again, we can race with udevd who can (and he is!) try to access it and get stuck.

Port of 6b6c112 ("lxd/storage/drivers/alletra: fix block device unmap logic for iSCSI").

Requires https://github.com/canonical/lxd-pkg-snap/pull/921

### DIFF
--- a/lxd/storage/drivers/driver_pure_util.go
+++ b/lxd/storage/drivers/driver_pure_util.go
@@ -1308,61 +1308,69 @@ func (d *pure) unmapVolume(vol Volume) error {
 		return err
 	}
 
+	// Get a path of a block device we want to unmap.
+	volumePath, _, _ := d.getMappedDevPath(vol, false)
+
+	// When iSCSI volume is disconnected from the host, the device will remain on the system.
+	//
+	// To remove the device, we need to either logout from the session or remove the
+	// device manually. Logging out of the session is not desired as it would disconnect
+	// from all connected volumes. Therefore, we need to manually remove the device.
+	//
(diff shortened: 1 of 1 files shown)


MusicDin:
Thanks, I'll report when tests are done

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -1370,6 +1358,26 @@ func (d *pure) unmapVolume(vol Volume) error {
...
 		}
+
+		// Device is not there anymore.
+		volumePath = ""
+	}
+
+	// Disconnect the volume from the host and ignore error if connection does not exist.
+	err = d.client().disconnectHostFromVolume(vol.pool, volName, host.Name)
+	if err != nil && !api.StatusErrorCheck(err, http.StatusNotFound) {
+		return err
```
tomponline, line 1369: please wrap errors as it helps when users report problems

PR #16525, opened 2025-09-16
Network: Follow-up to OVN IPv4 DHCP ranges

nmezhenskyi:
Follow-up to https://github.com/canonical/lxd/pull/16320.

## Changes:

Refactor the `complementRangesIP4` function.

- Avoid unnecessary string parsing of IP addresses.
- Fix the comment for the `previousEnd` var.
- Simplify checking if the calculated gap is a single IP or multi-IP.

### DIFF
--- a/lxd/network/driver_ovn.go
+++ b/lxd/network/driver_ovn.go
@@ -2227,7 +2227,7 @@ func (n *ovn) getDHCPv4Reservations() ([]shared.IPRange, error) {
 			dhcpReserveIPv4s = append(dhcpReserveIPv4s, reservedIPs...)
 
 			// Add the router IP in the reserved IP list, if it is not there yet.
-			if !ipInRanges(routerIntPortIPv4, dhcpReserveIPv4s) {
+			if !ipInRanges(routerIntPortIPv4.To4(), dhcpReserveIPv4s) {
 				dhcpReserveIPv4s = append(dhcpReserveIPv4s, shared.IPRange{Start: routerIntPortIPv4})
 			}
 		}
--- a/lxd/network/network_utils.go
+++ b/lxd/network/network_utils.go
@@ -1529,20 +1529,20 @@ func complementRangesIP4(ranges []*shared.IPRange, netAddr *net.IPNet) ([]shared
 		return nil, err
 	}
 
-	// Initialize a cursor to the start of the network. It tracks
-	// the end of the last covered IP range.
+	// Initialize a cursor to the start of the network.
+	// It tracks the end of the last covered IP range.
 	previousEnd := ipv4NetPrefix.Addr()
 
 	// Iterate over the sorted list of given IP ranges to find the gaps between them.
 	for _, r := range ranges {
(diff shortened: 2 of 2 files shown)


lxd/network/network_utils.go
```diff
@@ -1555,16 +1555,15 @@ func complementRangesIP4(ranges []*shared.IPRange, netAddr *net.IPNet) ([]shared
 		if startAddr.Compare(previousEndNext) == 1 {
 			newStart := previousEndNext
 			newEnd := startAddr.Prev()
+			newRange := shared.IPRange{Start: net.ParseIP(newStart.String())}
```
nmezhenskyi: We can't use `netip.AddrFromSlice` here because we're converting from `netip.Addr` to `net.IP`. I couldn't find a way that would avoid parsing a string here.

tomponline: https://pkg.go.dev/net/netip#Addr.AsSlice maybe?

nmezhenskyi: Nice, this seems to work, thanks!

nmezhenskyi: The interesting part is that `netip.AddrFromSlice` seems to break things, I'm looking into it...

nmezhenskyi: > The interesting part is that `netip.AddrFromSlice` seems to break things, I'm looking into it...

There was a comment in an Incus PR (https://github.com/lxc/incus/pull/2179) mentioning something to that effect:

```Go
// Convert the 4-byte IPv4 address returned by 'dhcpalloc.GetIP' to a 16-byte form
// using net.ParseIP, to ensure compatibility with other IPs stored in 16-byte format.
// This is necessary because direct comparison with 4-byte IPs would fail.
ovnRouter := net.ParseIP(dhcpalloc.GetIP(ipv4Net, -2).String())
```

tomponline: How about https://pkg.go.dev/net/netip#Addr.As4 then?

tomponline: or https://pkg.go.dev/net/netip#AddrFrom4

nmezhenskyi: Ensuring that `net.IP` value is in IPv4 representation using `.To4()` before passing it to the `netip.AddrFromSlice()` function seems to work

lxd/network/driver_ovn.go
```diff
@@ -2227,7 +2227,7 @@ func (n *ovn) getDHCPv4Reservations() ([]shared.IPRange, error) {
 			dhcpReserveIPv4s = append(dhcpReserveIPv4s, reservedIPs...)
 
 			// Add the router IP in the reserved IP list, if it is not there yet.
-			if !ipInRanges(routerIntPortIPv4, dhcpReserveIPv4s) {
+			if !ipInRanges(routerIntPortIPv4.To4(), dhcpReserveIPv4s) {
```
nmezhenskyi, line 2230: This change is important because otherwise the condition always evaluates to `true` and `routerIntPortIPv4` gets appended to the list even if it's already there.

PR #16526, opened 2025-09-16
Storage: Enable PowerFlex recovery

roosterfish:
This PR leverages the new backup config parsing behavior to enable recovery of volumes from PowerFlex backed storage pools.

In addition it adds another field `TransformedVolumeName` to the driver's `Info` struct which allows identifying whether or not a driver is using transformed volume names. This is important during recovery to figure out whether or not custom volumes can be recovered simply by name or not.
In addition this allows running an extra check during instance recovery to retrieve an instance root volume's actual name to allow checking if its already running or not.

Furthermore some errors and `volatile.uuid` generation on instance and custom volume import are improved/fixed.

### DIFF
--- a/test/suites/devlxd.sh
+++ b/test/suites/devlxd.sh
@@ -385,7 +385,7 @@ EOF
     [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage delete-volume "${pool}" "${instType}" "${inst}")" = "Only custom storage volume requests are allowed" ]
 
     # Delete storage volumes.
-    [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage delete-volume "${pool}" custom non-existing-volume)" = "Storage pool volume not found" ]
+    [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage delete-volume "${pool}" custom non-existing-volume)" = "Storage volume not found" ]
     lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage delete-volume "${pool}" custom vol-01
     lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage delete-volume "${pool}" custom vol-02
 


roosterfish:
@tomponline @MusicDin @mihalicyn Today I realized that we require an extra check for the instance's name to figure out whether or not it is already running to not disturb running workload when doing `lxd recover` in parallel. This is because we have to know the right path of the instance's backup config for the check but the path contains the instance's real name which we don't have before checking its backup config (chicken and egg). However we can safely leverage the UUID to resolve the name.

This is now ready for review. 
@MusicDin @mihalicyn for Pure and Alletra it's way less involved as this PR creates the groundwork. All which is required is setting the drivers `TransformedVolumeNames` to `true` and implementing the `ListVolumes` func. I'll now go ahead and continue doing it for Pure.
Maybe we can finish it too this week for Alletra :)

lxd/storage/drivers/driver_types.go
```diff
@@ -20,6 +20,7 @@ type Info struct {
 	IOUring                      bool         // Whether the driver supports io_uring.
 	MountedRoot                  bool         // Whether the pool directory itself is a mount.
 	PopulateParentVolumeUUID     bool         // Whether the volume should have parent UUID populated before any action.
+	TransformedVolumeName        bool         // Whether the volume name is transformed using its UUID.
```
tomponline: @roosterfish I think we are missing the definition/explanation of what "transformed volume name" actually is from here, with the expectation that the reader is already familiar with the powerflex, pure, hpe drivers. 

I think it would be more accessible for new developers if we expanded this comment to explain how and why this is used/needed.

lxd/storage/backend_lxd.go
```diff
@@ -7512,6 +7512,14 @@ func (b *lxdBackend) ListUnknownVolumes(op *operations.Operation) (map[string][]
 	for _, poolVol := range poolVols {
 		volType := poolVol.Type()
 
+		// Skip custom volumes from drivers using transformed volume names.
+		// We cannot recover the volume as we cannot retrieve the original volume's name.
+		// Instead the volume should be attached to any of the instances so it can be
+		// recovered through the respective backup config file.
+		if volType == drivers.VolumeTypeCustom && b.driver.Info().TransformedVolumeName {
```
tomponline: @roosterfish rather than adding the concept of `TransformedVolumeName` to the pool features list, can we just update the drivers that dont support custom volume recovery from returning custom volumes in `b.driver.ListVolumes`?

lxd/storage/backend_lxd.go
```diff
@@ -7560,6 +7560,49 @@ func (b *lxdBackend) ListUnknownVolumes(op *operations.Operation) (map[string][]
 // the state of the volume on disk, and if all checks out, it adds the parsed backup file contents to projectVols.
 // In any case it also checks whether or not the instance has unknown custom volumes attached and appends them to projectVols too.
 func (b *lxdBackend) detectUnknownInstanceAndCustomVolumes(vol *drivers.Volume, projectVols map[string][]*backupConfig.Config, op *operations.Operation) error {
+	volUUID := vol.Config()["volatile.uuid"]
+
+	// In case a driver is using volumes with transformed names, we have to identify which instance in LXD
```
tomponline: This feels like the wrong place to do this task (also this function is already too big so adding more to it is not attractive).


Lets think more about phases in the caller of this function:

1. Ask the storage driver for volumes - and as we know, for powerflex, pure and hpe, the volume names they return will be uuids not the actual volume name we want in the DB.
2. The missing step: We need to transform those volumes (if we can) into volume structs acceptable for the rest of the system (and potentially remove ones that aren't compatible, such as custom volumes, although im not sure if that part isn't better done in the storage driver).
3. Call `detectUnknownInstanceAndCustomVolumes` passing in the transformed volume struct.

tomponline: So it seems to be like we should have a new function that is called from `(b *lxdBackend) ListUnknownVolumes` that takes the output of `b.driver.ListVolumes` and cleans it up before we call `detectUnknownInstanceAndCustomVolumes`.

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -1295,3 +1295,19 @@ func (d *powerflex) getVolumeName(vol Volume) (string, error) {
 
 	return volName, nil
 }
+
+// getUUIDFromVolumeName transforms the volume's name to the respective UUID.
+// It expects the volume name without any prefix/suffix.
+func (d *powerflex) getUUIDFromVolumeName(name string) (string, error) {
```
tomponline: @roosterfish should this be on common driver so it can be re-used?

tomponline: also in the interests of keeping the native format as long as possible, can this return a `*uuid.UUID` rather than a string.

lxd/storage/backend_lxd.go
tomponline: Or we can include it in the volume transform function i suggest below https://github.com/canonical/lxd/pull/16526/files#r2354991298

lxd/storage/drivers/driver_powerflex_utils.go
roosterfish: Yes good point, otherwise we would need to make it common once we land the next `ListVolumes` for the next driver.

lxd/storage/backend_lxd.go
roosterfish: > can we just update the drivers that dont support custom volume recovery from returning custom volumes 

Mhh that is an interesting one. Initially I refrained from doing this because there are two more use cases which call `ListVolumes`. One is to delete image vols (irrelevant in this case as image vol != custom vol) and one in `patchBucketNames` which is also not relevant. 

I was about to say yes let's do this but inside `detectUnknownInstanceAndCustomVolumes` (or somewhere else when I have moved this check out), we have to recognize somehow if transformed vol names are used. 
So I think having the extra field is necessary but I'll make sure its properly documented and will apply the suggestion to move the logic out of the already big func.

tomponline: Yeah I appreciate we still need to recognise uuid volume names for instance volumes.

lxd/storage/drivers/driver_types.go
tomponline: How about we call this `UUIDVolumeNames` to be explicit (that way also there's less confusion about when/who is doing the transforming - otherwise we'll end up with logic that says `if TransformedVolumeName { doTransforming() }` which is odd.

lxd/storage/backend_lxd.go
tomponline: On further consideration, as we have to return UUID volume names for instances, lets just do it for all of them, and let the consumers figure out which they need/want.

lxd/storage/drivers/driver_powerflex_utils.go
roosterfish: Mh actually Pure and Alletra use a slightly different encoding scheme as some of the regular base64 characters are not allowed AFAIK.
So I think it should stay with the PowerFlex driver.

roosterfish: Fixed the return value.

lxd/storage/backend_lxd.go
roosterfish: Fixed, thanks for the suggestion.

lxd/storage/backend_lxd.go
roosterfish: > Or we can include it in the volume transform function i suggest below

Moved and fixed.

lxd/storage/drivers/driver_types.go
roosterfish: Renamed.

lxd/db/storage_volumes.go
```diff
@@ -118,6 +118,53 @@ WHERE storage_volumes.id = ?
...
+JOIN storage_pools ON storage_pools.id = storage_volumes.storage_pool_id
+JOIN projects ON projects.id = storage_volumes.project_id
+LEFT JOIN nodes ON nodes.id = storage_volumes.node_id
+WHERE storage_volumes_config.key = "volatile.uuid" AND storage_volumes_config.value = ?
+`
+
+	err := c.tx.QueryRowContext(ctx, stmt, volumeUUID).Scan(&response.ID, &response.Name, &response.Description, &response.CreationDate, &rawVolumeType, &response.NodeID, &response.PoolName, &response.ProjectName)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return StorageVolumeArgs{}, api.StatusErrorf(http.StatusNotFound, "Storage pool volume not found")
```
tomponline: We seem to mix "Storage pool volume not found" with "Storage volume not found" in various places.

To me, the latter feels more natural, could we standardise on that please?

lxd/storage/backend_lxd.go
```diff
@@ -7495,6 +7495,68 @@ func (b *lxdBackend) CheckInstanceBackupFileSnapshots(backupConf *backupConfig.C
 	return backupConf.Snapshots, nil
 }
 
+// resolveAndSkipUnknownVolumes tries to resolve volumes returned from a driver's ListVolumes.
+// In addition it skips unknown custom volumes which cannot be recovered by name in case its transformed.
+// In case a driver is using volumes with transformed names, we have to identify which instance in LXD
```
tomponline: "transformed names" is not the concept anymore, we're calling it `UUIDVolumeNames` now. We could make it explicit too like we do here:

https://github.com/canonical/lxd/blob/bb83ac977d03e2afd2af0b243a9ef9e5910dbee1/lxd/devlxd.go#L519

lxd/storage/backend_lxd.go
```diff
@@ -7495,6 +7495,68 @@ func (b *lxdBackend) CheckInstanceBackupFileSnapshots(backupConf *backupConfig.C
...
+		if b.driver.Info().UUIDVolumeNames {
+			// Skip custom volumes from drivers using transformed volume names.
+			// We cannot recover the volume as we cannot retrieve the original volume's name.
+			// Instead the volume should be attached to any of the instances so it can be
+			// recovered through the respective backup config file.
+			if vol.Type() == drivers.VolumeTypeCustom {
+				continue
+			}
+
+			err := b.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
```
tomponline: this function should take a context and have it passed here for future proofing.

lxd/storage/backend_lxd.go
```diff
@@ -7495,6 +7495,68 @@ func (b *lxdBackend) CheckInstanceBackupFileSnapshots(backupConf *backupConfig.C
...
+// Therefore try to lookup the instance's root volume using the volume's UUID.
+// An instance's root volume always has the same name as its parent instance.
+func (b *lxdBackend) resolveAndSkipUnknownVolumes(poolVols []drivers.Volume) ([]drivers.Volume, error) {
+	resolvedUnknownVols := make([]drivers.Volume, 0, len(poolVols))
+
+	for _, vol := range poolVols {
+		volUUID := vol.Config()["volatile.uuid"]
+
+		if vol.Name() == "" {
+			return nil, fmt.Errorf("Cannot detect instance with UUID %q on pool %q using an empty volume name", volUUID, b.name)
```
tomponline: this error doesnt correlate with the check, its checking volume name and complaining about UUID, furthermore this check isn't restricted to only pools with `UUIDVolumeNames`, so doesn't make sense in a general context.

lxd/storage/backend_lxd.go
tomponline: can we start the transaction outside of the loop to avoid needing to start a transaction per volume?

lxd/storage/backend_lxd.go
```diff
@@ -7495,6 +7495,68 @@ func (b *lxdBackend) CheckInstanceBackupFileSnapshots(backupConf *backupConfig.C
...
+			// recovered through the respective backup config file.
+			if vol.Type() == drivers.VolumeTypeCustom {
+				continue
+			}
+
+			err := b.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+				var err error
+
+				// Fetch the instance's volume from the DB in case it already exists by using its unique UUID.
+				dbVol, err := tx.GetStoragePoolVolumeWithUUID(ctx, volUUID)
```
tomponline, line 7539: should we verify that the volume type of the record we got matches the volume's type from the storage driver?

tomponline: as an aside, what are we doing around volume snapshot recovery?

lxd/storage/drivers/driver_powerflex_volumes.go
```diff
@@ -657,9 +658,86 @@ func (d *powerflex) GetVolumeDiskPath(vol Volume) (string, error) {
...
+			contentType = ContentTypeISO
+			volName = strings.TrimSuffix(volName, powerFlexISOVolSuffix)
+		} else if volType == VolumeTypeVM || isBlock {
+			contentType = ContentTypeBlock
+			volName = strings.TrimSuffix(volName, powerFlexBlockVolSuffix)
+		}
+
+		volUUID, err := d.getUUIDFromVolumeName(volName)
+		if err != nil {
+			d.logger.Debug("Ignoring malformed volume name", logger.Ctx{"name": vol.Name})
```
tomponline: when does this occur? Should it be a warning?

lxd/storage/drivers/driver_powerflex_volumes.go
```diff
@@ -657,9 +658,86 @@ func (d *powerflex) GetVolumeDiskPath(vol Volume) (string, error) {
...
+
+		// This is important to allow subsequent operations on the volume struct (e.g. driver's HasVolume) to be able to
+		// resolve the volume's name using the its volatile.uuid.
+		volConfig := map[string]string{
+			"volatile.uuid": volUUID.String(),
+		}
+
+		// We cannot determine the volume's name.
+		// Therefore we set it to the UUID.
+		// If we would keep the volume's name on storage, it might contain special characters.
```
tomponline, line 729: nice comment, good reasoning, thanks

lxd/storage/backend_lxd.go
```diff
@@ -8048,7 +8048,14 @@ func (b *lxdBackend) ImportInstance(inst instance.Instance, poolVol *backupConfi
 	volStorageName := project.Instance(inst.Project().Name, inst.Name())
 
 	// Copy the volume's config so VolumeDBCreate can safely modify the copy if needed.
-	vol := b.GetNewVolume(volType, contentType, volStorageName, volumeConfig)
+	// Don't generate a new UUID as we are importing an existing instance's volume.
+	vol := b.GetVolume(volType, contentType, volStorageName, volumeConfig)
+
+	// In case the instance got lost on an old LXD which didn't set a volume's UUID,
+	// ensure we set a new one.
+	if vol.Config()["volatile.uuid"] == "" {
```
tomponline: just a nit, but we could avoid calling vol.Config() twice if its empty by storing it in a temporary variable.

lxd/storage/backend_lxd.go
```diff
@@ -8075,7 +8082,14 @@ func (b *lxdBackend) ImportInstance(inst instance.Instance, poolVol *backupConfi
 
 				// Copy volume config from backup file if present,
 				// so VolumeDBCreate can safely modify the copy if needed.
-				snapVol := b.GetNewVolume(volType, contentType, fullSnapName, poolVolSnap.Config)
+				// Don't generate a new UUID as we are importing an existing instance's snapshot volume.
+				snapVol := b.GetVolume(volType, contentType, fullSnapName, poolVolSnap.Config)
+
+				// See the comment for the instance's root volume above.
+				// The same also applies for its snapshots.
+				if snapVol.Config()["volatile.uuid"] == "" {
```
tomponline: just a nit, but we could avoid calling snapVol.Config() twice if its empty by storing it in a temporary variable.

lxd/storage/backend_lxd.go
```diff
@@ -6671,7 +6671,13 @@ func (b *lxdBackend) ImportCustomVolume(projectName string, poolVol *backupConfi
 
 		// Copy volume config from backup file if present
 		// (so VolumeDBCreate can safely modify the copy if needed).
-		snapVol := b.GetNewVolume(drivers.VolumeTypeCustom, drivers.ContentType(poolVolSnap.ContentType), fullSnapName, poolVolSnap.Config)
+		snapVol := b.GetVolume(drivers.VolumeTypeCustom, drivers.ContentType(poolVolSnap.ContentType), fullSnapName, poolVolSnap.Config)
+
+		// See the comment for the custom volume above.
+		// The same also applies for its snapshots.
+		if snapVol.Config()["volatile.uuid"] == "" {
```
tomponline: just a nit, but we could avoid calling snapVol.Config() twice if its empty by storing it in a temporary variable.

lxd/db/storage_volumes.go
roosterfish: Fixed. Also added a commit to fix the other errors in this file.

lxd/storage/backend_lxd.go
roosterfish: Fixed and removed all traces of "transformed" from code and commit messages.

lxd/storage/backend_lxd.go

lxd/storage/backend_lxd.go
roosterfish: Made it a warning instead with more generic message. 
As the volume's UUID is the only identifier at this stage, I have added a `uuid` log ctx field. Should we rather use `volume_uuid` or `volatile.uuid`? I found just `uuid` to be most consistent with the rest.

lxd/storage/backend_lxd.go

lxd/storage/backend_lxd.go
roosterfish: Do you see a case where a volume of another type uses the same UUID? 

> as an aside, what are we doing around volume snapshot recovery?

This is essentially the early stage of recovery. We got back a list of vols from the driver and can now resolve or skip volumes. Later on inside the actual `detectUnknown*` funcs we then look further ahead to also find the snapshots.

The driver's `ListVolumes` func doesn't return snapshots.

lxd/storage/drivers/driver_powerflex_volumes.go
roosterfish: It might happen if somebody puts another volume in the storage pool (aside from LXD) with a name that we cannot understand. 
Fixed it to be a warning as it shouldn't really matter but we should notify about this.

lxd/storage/backend_lxd.go

lxd/storage/backend_lxd.go

lxd/storage/backend_lxd.go

lxd/storage/backend_lxd.go
tomponline: > Do you see a case where a volume of another type uses the same UUID?
> 
> >

no just defensive coding

lxd/storage/backend_lxd.go
tomponline: uuid is fine by me

lxd/storage/backend_lxd.go
roosterfish: > no just defensive coding

I see, added a safety check.

lxd/storage/backend_lxd.go
```diff
@@ -7495,6 +7495,75 @@ func (b *lxdBackend) CheckInstanceBackupFileSnapshots(backupConf *backupConfig.C
...
+	resolvedUnknownVols := make([]drivers.Volume, 0, len(poolVols))
+
+	err := b.state.DB.Cluster.Transaction(ctx, func(ctx context.Context, tx *db.ClusterTx) error {
+		for _, vol := range poolVols {
+			volUUID := vol.Config()["volatile.uuid"]
+
+			// This is likely an indicator that the storage driver's ListVolumes returned volumes without setting a name.
+			// We have to skip those volumes as the name is mandatory to create the correct volume's mount path.
+			if vol.Name() == "" {
+				logger.Warn("Cannot resolve or skip unknown volume with empty name", logger.Ctx{"pool": b.name, "uuid": volUUID})
```
tomponline: although you are skipping it :)

lxd/storage/backend_lxd.go
```diff
@@ -7495,6 +7495,75 @@ func (b *lxdBackend) CheckInstanceBackupFileSnapshots(backupConf *backupConfig.C
...
+// resolveAndSkipUnknownVolumes tries to resolve volumes returned from a driver's ListVolumes.
+// In addition it skips unknown custom volumes which cannot be recovered by name.
+// This might be the case if the volume name is based on its UUID which cannot be translated to the original name.
+// Instead we have to identify which instance in LXD belongs to the respective volume to resolve its name.
+// This is necessary to not disturb potentially running instances in case we are trying to re-mount an instance's volume
+// which is already mounted to the system by its actual name which we don't yet know because we haven't looked into
+// the instances backup config file.
+// Therefore try to lookup the instance's root volume using the volume's UUID.
+// An instance's root volume always has the same name as its parent instance.
+func (b *lxdBackend) resolveAndSkipUnknownVolumes(ctx context.Context, poolVols []drivers.Volume) ([]drivers.Volume, error) {
```
tomponline: this isn't a blocker, but i think the name here is quite confusing, as it sounds like "resolve and skip unknown volumes" suggesting that unknown volumes are not returned from this function, but ofcourse they are.

perhaps "normalise" would be a better verb, without the "skip" part?

lxd/storage/backend_lxd.go
roosterfish: Let me pick another name for the func as you have suggested below. Then I'll also pick a more descriptive warning message. It's indeed odd. Will fix.

lxd/storage/backend_lxd.go

roosterfish:
Pushed one more commit, see https://github.com/canonical/lxd/compare/61e202f4cb27ab5dc066e4ca742aec2e6431ad4f..93e4e2a4430ff96b982126f27361a61c1fed5f69 or the latest change set.

I figured (whilst doing many recover, delete DB records, recover, delete DB records, ...) that there is a left over directory for the instance's mount path (when mounting using the volume's UUID as path base).
Added another cleanup step to ensure we don't leave unused directories behind.

tomponline:
> Pushed one more commit, see https://github.com/canonical/lxd/compare/61e202f4cb27ab5dc066e4ca742aec2e6431ad4f..93e4e2a4430ff96b982126f27361a61c1fed5f69 or the latest change set.
> 
> I figured (whilst doing many recover, delete DB records, recover, delete DB records, ...) that there is a left over directory for the instance's mount path (when mounting using the volume's UUID as path base). Added another cleanup step to ensure we don't leave unused directories behind.

We have `Volume.mountCustomPath` which is used by `MountPath()`.

Is this something the UUID based drivers could utilise when returning volumes from `ListVolumes()` to avoid (push down) some of this special casing?

roosterfish:
> Is this something the UUID based drivers could utilise when returning volumes from `ListVolumes()` to avoid (push down) some of this special casing?

Would be great, but we cannot set it inside the driver (ListVolumes) as we haven't yet mounted the vol to figure out the actual instance name. We do this afterwards in the backend but there we cannot anymore access the unexported `mountCustomPath`.

roosterfish:
But we could do something like `vol.SetConfigSize` called `.SetCustomMountPath` to override the internal field? Maybe that's the cleanest.

roosterfish:
> But we could do something like `vol.SetConfigSize` called `.SetCustomMountPath` to override the internal field? Maybe that's the cleanest.

Nvm, this does not work due to chicken and egg situation. We cannot set the actual instance's mount path before checking its backup config which we can only do after mounting using its UUID as basename.

roosterfish:
@tomponline pushed the discussed changes.

Regarding the special error message inside `detectUnknownInstanceAndCustomVolumes` I found this approach to be the cleanest: If the driver isn't using UUID based vol names, return the error using the vol's actual name. In all other cases instead of the vols name used the mount path which contains the volume's UUID for later reference. This way we don't need any extra "unknown" string but still print all the information we have about the vol.

lxd/storage/backend_lxd.go
```diff
@@ -7495,6 +7495,78 @@ func (b *lxdBackend) CheckInstanceBackupFileSnapshots(backupConf *backupConfig.C
...
+
+				// In case the new volume's name is empty, set a custom mount path based on the volume's UUID.
+				// This ensures when mounting the volume, we are picking a unique path as the actual volume's name is empty.
+				if newVolName == "" {
+					newVol.SetMountCustomPath(drivers.GetVolumeMountPath(b.name, newVol.Type(), volUUID))
+				}
+
+				normalizedUnknownVols = append(normalizedUnknownVols, newVol)
+			} else {
+				normalizedUnknownVols = append(normalizedUnknownVols, vol)
```
tomponline, line 7573: should we also check for volume name not being empty here too?

lxd/storage/backend_lxd.go
```diff
@@ -7632,6 +7633,38 @@ func (b *lxdBackend) ListUnknownVolumes(op *operations.Operation) (map[string][]
...
+// In case an unknown volume wasn't known in the DB, we couldn't normalize it to find out its actual name.
+// This caused the volume to be mounted under its UUID (last part of the mount path).
+// After recovery the volume gets mounted under its actual name which would leave an empty directory behind.
+func (b *lxdBackend) cleanupUnknownVolumeMountPath(poolVol *drivers.Volume) error {
+	if b.driver.Info().UUIDVolumeNames {
+		mountPath := poolVol.MountPath()
+		mountPathBase := path.Base(mountPath)
+
+		// It looks like the mount path already used the instance's name so skip the cleanup.
+		if mountPathBase != poolVol.Config()["volatile.uuid"] {
```
tomponline, line 7656: should we check for empty volume name here too to make it explicit?

lxd/storage/backend_lxd.go
roosterfish: Good catch, thanks fixed.

lxd/storage/backend_lxd.go
roosterfish: Thanks fixed, added it to the outer if.

lxd/storage/backend_lxd.go
```diff
@@ -7495,6 +7495,78 @@ func (b *lxdBackend) CheckInstanceBackupFileSnapshots(backupConf *backupConfig.C
...
+// An instance's root volume always has the same name as its parent instance.
+func (b *lxdBackend) normalizeUnknownVolumes(ctx context.Context, poolVols []drivers.Volume) ([]drivers.Volume, error) {
+	normalizedUnknownVols := make([]drivers.Volume, 0, len(poolVols))
+
+	err := b.state.DB.Cluster.Transaction(ctx, func(ctx context.Context, tx *db.ClusterTx) error {
+		for _, vol := range poolVols {
+			// This is likely an indicator that the storage driver's ListVolumes returned volumes with a set name.
+			// We have to skip those volumes as they are not following the protocol.
+			if vol.Name() != "" {
+				logger.Warn("Cannot normalize unknown volume with already set name", logger.Ctx{"pool": b.name, "volume": vol.Name()})
```
tomponline: wont this now fail for volumes on pools where UUIDVolumeNames == true?

What I thought we needed was:

- if UUIDVolumeNames then vol.Name() must be empty
- if !UUIDVolumeNames then vol.Name() must be non-empty

roosterfish: My bad, I was doing this too fast. Now I have added the corresponding warning for both cases `UUIDVolumeNames=(true|false)`. Test passing locally, also checked on PowerFlex env.

tomponline (approved):
Thanks!

PR #16531, opened 2025-09-16
Client user agent

markylaing:
This PR adds a different user agent to the client used by `lxc` from the one exported by `shared/version`. The new User-Agent looks like:
```
lxc/6.5 (LTS(false);OS(Linux;x86_64);capabilities(cap1;cap2))
```

I'm very open to suggestions about what the format of the user agent should look like. I was using this [RFC](https://www.rfc-editor.org/rfc/rfc9110#name-user-agent) as a guide. I've added everyone as a reviewer to encourage rigorous debate :wink: 

The only requirement is that the client must be able to advertise it's capabilities. A bit like how LXD exposes it's API extensions so the client knows whether what it is trying to do is supported or not.

This is required for #15030 so that OIDC auth will fail if the client is too old* to store cookies in a local file.

**Or not old enough... since we had cookie support until it was removed when macaroon auth was removed.*

### DIFF
--- a/lxc/main.go
+++ b/lxc/main.go
@@ -22,6 +22,18 @@ import (
 	"github.com/canonical/lxd/shared/version"
 )
 
+var capabilities = []string{
+	"cookiejar",
+}
+
+var userAgent = version.ClientUserAgent{
+	Name:         "lxc",
+	Version:      version.Version,
+	IsLTS:        version.IsLTSVersion,
+	OS:           version.GetOSTokens(),
+	Capabilities: capabilities,
+}
+
 type cmdGlobal struct {
 	asker cli.Asker
 
@@ -467,7 +479,7 @@ Or for a virtual machine: lxc launch ubuntu:24.04 --vm`)
 	}
 
 	// Set the user agent
-	c.conf.UserAgent = version.UserAgent
+	c.conf.UserAgent = userAgent.String()
 
 	// Setup the logger
 	err = logger.InitLogger("", "", c.flagLogVerbose, c.flagLogDebug, nil)
--- a/shared/version/useragent.go
+++ b/shared/version/useragent.go
@@ -1,7 +1,10 @@
 package version
 
 import (
+	"fmt"
+	"regexp"
 	"runtime"
+	"strconv"
 	"strings"
 
 	"golang.org/x/text/cases"
@@ -16,19 +19,6 @@ var userAgentStorageBackends []string
 var userAgentFeatures []string
 
 func getUserAgent() string {
-	archID, err := osarch.ArchitectureId(runtime.GOARCH)
-	if err != nil {
-		panic(err)
-	}
-
-	arch, err := osarch.ArchitectureName(archID)
-	if err != nil {
-		panic(err)
-	}
-
-	osTokens := []string{cases.Title(language.English).String(runtime.GOOS), arch}
-	osTokens = append(osTokens, getPlatformVersionStrings()...)
-
 	// Initial version string
 	agent := "LXD " + Version
 	if IsLTSVersion {
@@ -36,7 +26,7 @@ func getUserAgent() string {
 	}
 
 	// OS information
-	agent = agent + " (" + strings.Join(osTokens, "; ") + ")"
+	agent = agent + " (" + strings.Join(GetOSTokens(), "; ") + ")"
 
 	// Storage information
 	if len(userAgentStorageBackends) > 0 {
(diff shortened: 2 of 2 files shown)


kadinsayani:
Have we considered adopting user agent client hints? With client hints, the significant version can be used to determine if a client includes a specific capability.

tomponline:
> I'm very open to suggestions about what the format of the user agent should look like. I was using this [RFC](https://www.rfc-editor.org/rfc/rfc9110#name-user-agent) as a guide. I've added everyone as a reviewer to encourage rigorous debate 😉

Please could you include in the PR description what the resulting header looks like to save jumping straight into the code to figure it out. Thanks

tomponline:
> Have we considered adopting user agent client hints? With client hints, the significant version can be used to determine if a client includes a specific capability.

mmm interesting, thanks, there's list of client hint headers here:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers#client_hints

But the key bit from @markylaing is:

> The only requirement is that the client must be able to advertise it's capabilities. A bit like how LXD exposes it's API extensions so the client knows whether what it is trying to do is supported or not.

And I don't see a supported way of sending this using client hint headers.

We're trying to avoid version number logic and focus on capabilities, similar to API extensions concept, so that if this functionality is backported it starts working.

So it looks like we'd still be left encoding these into one of those fields.

Also it looks like those headers are not standardised just yet and are liable to change.

tomponline:
@markylaing it looks like we already have the needed functionality, by way of the `UserAgentFeatures()` function which:

> // UserAgentFeatures updates the list of advertised features.

What seems less than ideal is that if `userAgentStorageBackends` is <= 0 then its not included, and we have no way of identifying what each section is.

 So perhaps we have make a small modification (to which @simondeziel will need to be aware of to update his parser), to include the section name inside the brackets, e.g.

Today:
```go
version.UserAgentStorageBackends([]string{"foo 1.0", "bar 2.0"})
version.UserAgentFeatures([]string{"mycap"})
// LXD 6.5 (Linux; x86_64; 6.8.0; Ubuntu; 24.04) (foo 1.0; bar 2.0) (mycap)
```

Propose:

```go
version.UserAgentStorageBackends([]string{"foo 1.0", "bar 2.0"})
version.UserAgentFeatures([]string{"mycap"})
// LXD 6.5 (Linux; x86_64; 6.8.0; Ubuntu; 24.04) (storage:foo 1.0; bar 2.0) (features:mycap)
```

markylaing:
> @markylaing it looks like we already have the needed functionality, by way of the `UserAgentFeatures()` function which:
> 
> > // UserAgentFeatures updates the list of advertised features.
> 
> What seems less than ideal is that if `userAgentStorageBackends` is <= 0 then its not included, and we have no way of identifying what each section is.
> 
> So perhaps we have make a small modification (to which @simondeziel will need to be aware of to update his parser), to include the section name inside the brackets, e.g.
> 
> Today:
> 
> ```go
> version.UserAgentStorageBackends([]string{"foo 1.0", "bar 2.0"})
> version.UserAgentFeatures([]string{"mycap"})
> // LXD 6.5 (Linux; x86_64; 6.8.0; Ubuntu; 24.04) (foo 1.0; bar 2.0) (mycap)
> ```
> 
> Propose:
> 
> ```go
> version.UserAgentStorageBackends([]string{"foo 1.0", "bar 2.0"})
> version.UserAgentFeatures([]string{"mycap"})
> // LXD 6.5 (Linux; x86_64; 6.8.0; Ubuntu; 24.04) (storage:foo 1.0; bar 2.0) (features:mycap)
> ```

I thought of using this initially, however I wasn't sure a) about the use of spaces (the EBNF for the user agent in the [RFC](https://www.rfc-editor.org/rfc/rfc9110#name-user-agent) is quite specific about this, and b) whether changes to the user agent here would affect parsing for metrics.

Also, what I'm aiming for is something that can be used by any client, not just the go client. So perhaps even the LTS field should be removed in favour of having it in the version string.

tomponline:
> EBNF 

?

tomponline:
> Also, what I'm aiming for is something that can be used by any client, not just the go client

What do you mean?

tomponline:
> whether changes to the user agent here would affect parsing for metrics.

The last thing we want to do is have 2 different types of user agent formats in LXD, so both my proposal and your proposal require changes to this anyway (although my proposal is less dramatic).

tomponline:
@markylaing did you consider just json encoding the struct into the header?

markylaing:
> > Also, what I'm aiming for is something that can be used by any client, not just the go client
> 
> What do you mean?

Let's say e.g. I want to create another CLI or desktop application for LXD. I might want to write it in Python and use PyLXD. Ideally we'd have a way of expressing client features that is clear and well structured.

> @markylaing did you consider just json encoding the struct into the header?

I did, and RE my previous point I think this would be the clearest way to do it from a client perspective. It just doesn't match what a user agent is "supposed" to be according to the RFC.

> > EBNF
>
> ?

Sorry they call it "ABNF" in the spec. It's just the syntax that is used to define what the format of the user agent should take.

tomponline:

markylaing:
> @markylaing shall we close this for now?

Yes. Will use existing user agent for now and revisit later.

PR #16533, opened 2025-09-16
Forced project deletion

kadinsayani:


lxd/api_project.go
```diff
@@ -1078,6 +1080,252 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
 
+// doProjectForceDelete handles force deletion of project entities.
+func doProjectForceDelete(ctx context.Context, s *state.State, projectName string) error {
+	// Get the project entities and config.
+	var usedBy []string
+	var projectConfig map[string]string
+	err := s.DB.Cluster.Transaction(ctx, func(ctx context.Context, tx *db.ClusterTx) error {
+		project, err := dbCluster.GetProject(ctx, tx.Tx(), projectName)
+		if err != nil {
+			return fmt.Errorf("Fetch project %q: %w", projectName, err)
```
tomponline: Failed loading project

lxd/api_project.go
```diff
@@ -1078,6 +1080,252 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+	localNodeName := s.ServerName
+
+	for _, u := range usedBy {
+		if u == defaultProfile {
+			continue
+		}
+
+		url, err := url.Parse(u)
+		if err != nil {
+			return err
```
tomponline: wrap error

tomponline: and lets parse all the urls from usedBy first before we start doing any deletion.

validate all, then delete

lxd/api_project.go
```diff
@@ -1078,6 +1080,252 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+		}
+
+		entityType, _, target, pathArgs, err := entity.ParseURL(*url)
+		if err != nil {
+			return err
+		}
+
+		// Skip entities that are not local to this node (only applies to instances and storage volumes).
+		if (entityType == entity.TypeInstance || entityType == entity.TypeStorageVolume) && target != "" && target != localNodeName {
```
tomponline: why not use s.ServerName directly here instead of saving it to a local var?

lxd/api_project.go
```diff
@@ -1078,6 +1080,252 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+		if err != nil {
+			return err
+		}
+
+		entityType, _, target, pathArgs, err := entity.ParseURL(*url)
+		if err != nil {
+			return err
+		}
+
+		// Skip entities that are not local to this node (only applies to instances and storage volumes).
```
tomponline: expand the comment with the *why* please

lxd/api_project.go
```diff
@@ -1078,6 +1080,252 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+		if (entityType == entity.TypeInstance || entityType == entity.TypeStorageVolume) && target != "" && target != localNodeName {
+			continue
+		}
+
+		// Delete the entity based on its type.
+		switch entityType {
+		case entity.TypeInstance:
+			instanceName := pathArgs[0]
+			err = s.Authorizer.CheckPermission(ctx, entity.InstanceURL(projectName, instanceName), auth.EntitlementCanDelete)
```
tomponline: @markylaing if you have can delete on a project, do you implicitly have can delete on all entities inside that project, or do we need to check each entity separately?

lxd/api_project.go
```diff
@@ -1078,6 +1080,252 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+				return err
+			}
+
+			// Load instance.
+			inst, err := instance.LoadByProjectAndName(s, projectName, instanceName)
+			if err != nil {
+				return err
+			}
+
+			if inst.State() != strings.ToUpper(api.Stopped.String()) {
```
tomponline: shouldnt this logic be handled by `doInstanceDelete` perhaps with a force argument?

This should stand out as something that is not the responsibility of this function.

lxd/api_project.go
```diff
@@ -1078,6 +1080,252 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
 	return nil
 }
 
+// doProjectForceDelete handles force deletion of project entities.
+func doProjectForceDelete(ctx context.Context, s *state.State, projectName string) error {
```
tomponline, line 1082: this function has a high cyclomatic complexity. I think we're going to need to break it out more.

lxd/api_project.go
```diff
@@ -1078,6 +1080,252 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+			// Delete profile.
+			err = doProfileDelete(ctx, s, profileName, projectName)
+			if err != nil {
+				return err
+			}
+
+			// Delete default profile if "features.profiles" is enabled.
+			if shared.IsTrue(projectConfig["features.profiles"]) {
+				err = s.DB.Cluster.Transaction(ctx, func(ctx context.Context, tx *db.ClusterTx) error {
+					return dbCluster.DeleteProfile(ctx, tx.Tx(), projectName, "default")
```
tomponline: do we even need this part? profiles are just DB records linked to the project itself, and when the project record is removed the profile records will be too.

lxd/api_project.go
```diff
@@ -1078,6 +1080,252 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+			err = s.DB.Cluster.Transaction(ctx, func(ctx context.Context, tx *db.ClusterTx) error {
+				imageID, _, err = tx.GetImage(ctx, fingerprint, dbCluster.ImageFilter{Project: &projectName})
+				return err
+			})
+			if err != nil {
+				return err
+			}
+
+			// Delete image.
+			op, err := doImageDelete(ctx, s, fingerprint, imageID, projectName)
```
tomponline: was there a reason that these entity type specific delete functions dont take the name/fingerprint of the entity and load the DB record internally? Why are they passed IDs?

tomponline: Is this because in the API handlers themselves this has already been loaded by the API handler middleware?

lxd/api_project.go
```diff
@@ -1104,18 +1357,34 @@ func projectDelete(d *Daemon, r *http.Request) response.Response {
...
 
+	force := shared.IsTrue(r.FormValue("force"))
+	target := request.QueryParam(r, "target")
+
 	// Quick checks.
 	if name == api.ProjectDefaultName {
 		return response.Forbidden(errors.New("The 'default' project cannot be deleted"))
 	}
 
+	// Handle requests targeted to a node.
```
tomponline: projects are not node specific so they cant be targetted at a node.

lxd/api_project.go
kadinsayani: Yes, the request context includes an `imageDetails` struct which has fields for fingerprint, ID, and an `api.Image`.

lxd/api_project.go
kadinsayani: You're right, we don't need this part.

lxd/api_project.go
kadinsayani: Good idea, I'll add a force param to `doInstanceDelete()` and open a separate PR.

kadinsayani: https://github.com/canonical/lxd/pull/16560

kadinsayani:
> When ever you find yourself coding a massive switch statement that covers each entity type, stop, consider the maintenance of this code, and try to come up with a better way of doing it. We already have an interface for entities to implement for entity type specific behaviour, could we utilise this @markylaing ?

I've cleaned the function up a bit and added a few TODO comments for us to discuss during our 1:1 tomorrow. I've been considering different approaches and design patterns for bulk operations in LXD but I don't want us to end up in a scenario where we go with one approach and require something different if/when we are looking at an external scheduler or task operation tool, hence why I went with the giant for loop + switch statement approach.

One idea I had is to add another level of abstraction to each `do*Delete()` function, where `*` is the entity. We could have `delete*()` functions with a generalized set of params (only what is required: `context.Context`, `state.State`, and the entity name). That way we could map each entity type to its delete function.

Another idea I had is to add a new interface called `EntityAction` to a new package. Each entity type can implement a `Delete()` function that gathers the required params for its `do*Delete()` function and then calls it.

tomponline:
> One idea I had is to add another level of abstraction to each `do*Delete()` function, where `*` is the entity. We could have `delete*()` functions with a generalized set of params (only what is required: `context.Context`, `state.State`, and the entity name). That way we could map each entity type to its delete function.
> 
> Another idea I had is to add a new interface called `EntityAction` to a new package. Each entity type can implement a `Delete()` function that gathers the required params for its `do*Delete()` function and then calls it.

This sounds like 2 parts of the same idea: an interface and multiple implementations.

Is it possible to generalise them though?

tomponline:
> bulk operations in LXD

Yeah we don't need to consider that for this pr

lxd/api_project.go
markylaing: You need to check everything separately.

@kadinsayani we'll need to do this first when parsing URLs before actually deleting anything so that the whole action either succeeds or fails

tomponline: > You need to check everything separately.

Why is that? If I van delete a project isn't it implicit I can delete its contents?

tomponline: Strong agree on doing prechecks up front though!

kadinsayani:
I've gone with a more generalized approach following the strategy pattern. 

The general idea is to call different `do*Delete()` functions based on the entity type. To achieve this, I've added a `entity.Reference` struct which is constructed from project name, entity type, location, and path args (which are all extracted from the entity URL). There is a "*deleter" type for each entity type, which implements its own `Delete()` function.

There's still work to be done to further generalize so we can adapt the new interface to more types and follow the same pattern for other entity specific actions. Additionally, the naming for the new files, interfaces, and types is all subject to change pending further discussion.

kadinsayani:
`doProjectForceDelete()` is much cleaner now, however, I'm still working out how to handle the permission checks since we probably also want to include checks in the `(entityDeleter).Delete()` functions. 

I think it may be helpful to add a `can_force_delete` project entitlement which inherits `can_delete` entitlements for all relevant entities. However, we would still end up checking permissions twice.

lxd/entity_action_types.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// Optional fields depending on type.
+	name     string // instance, profile, network, ACL, zone, volume, or bucket name, or image fingerprint
+	poolName string // storage pool for volumes/buckets
+	volType  string // storage volume type name (e.g. custom)
+	target   string // cluster member name (for location-specific entities: instances, storage volumes, and storage buckets)
+}
+
+// entityDeleter defines how to delete a specific entity type.
+type entityDeleter interface {
```
kadinsayani: Instead of `entityDeleter`, we could have `entityActor` which could grow to define additional actions. This would also make organizing the implementations easier since we could have a separate file for each entity type, e.g. entity_instance_action.go which includes `type instanceActor struct{}` and its `Delete()` implementation, and eventually more implementations.

kadinsayani:
I've made an attempt at de-duplicating by pulling up a `Reference` type from `lxd/db/cluster/entities.go` to the `shared/entity` package. Still a WIP but getting closer...

@tomponline I'll further de-dupe and split this PR up following our chat tomorrow :)

tomponline:
@markylaing @kadinsayani and I were chatting about permissions and I wondered if the `project_manager` was named somewhat misleadingly as it suggests it allows managing of a single project, but actually allows management of *all* projects.

Shall we rename this to `all_projects_manager`?

lxd/entity_action_types.go
kadinsayani: This can be considered at a later time.

lxd/api_project.go
```diff
@@ -1077,6 +1078,85 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+
+		// Build an entity reference directly from the URL.
+		ref, err := entity.ReferenceFromURL(*parsedURL)
+		if err != nil {
+			return err
+		}
+
+		// Skip instances and storage volumes not hosted on this node when a target is specified.
+		// These entities will be deleted by the node they are hosted on when a delete request is a cluster notification.
+		if (ref.EntityType == entity.TypeInstance || ref.EntityType == entity.TypeStorageVolume) && ref.Location != "" && ref.Location != s.ServerName {
```
simondeziel, line 1127: Unless `s.ServerName` can be empty, I'd go with the simpler:
```suggestion
		if (ref.EntityType == entity.TypeInstance || ref.EntityType == entity.TypeStorageVolume) && ref.Location != s.ServerName {
```

Or maybe even split that into 2 `if` with the first checking only the `ref.Location` condition.

lxd/api_project.go
```diff
@@ -1077,6 +1078,85 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+			return err
+		}
+
+		toDelete = append(toDelete, ref)
+	}
+
+	// Delete entities.
+	for _, ref := range toDelete {
+		deleter := getEntityDeleter(ref.EntityType)
+		if deleter != nil {
```
simondeziel: Continuing on `deleter == nil` would reduce the indentation:
```suggestion
		if deleter == nil {
		    continue
	    }
```

test/suites/projects.sh
```diff
@@ -1415,3 +1415,85 @@ test_projects_backups_volume() {
...
+  [ "$(lxc storage bucket list s3 --all-projects | grep -cF bucket1)" = 0 ]
+  [ "$(lxc network list --all-projects | grep -cF foonet)" = 0 ]
+  [ "$(lxc network acl list --all-projects | grep -cF acl1)" = 0 ]
+  [ "$(lxc network zone list --all-projects | grep -cF zone1)" = 0 ]
+  [ "$(lxc profile list --all-projects | grep -cF profile1)" = 0 ]
+  [ "$(lxc image list --all-projects | grep -cF testimage)" = 0 ]
+  [ "$(lxc list --all-projects | grep -cF c1)" = 0 ]
+
+  echo "Check project is deleted."
+  ! lxc project show foo || false
```
simondeziel, line 1516: I'd do that earlier, before checking individual entities are deleted.

test/suites/projects.sh
```diff
@@ -1415,3 +1415,85 @@ test_projects_backups_volume() {
...
+
+  echo "Check all entities are deleted."
+  [ "$(lxc storage volume list "${pool}" --all-projects | grep -cF vol1)" = 0 ]
+  [ "$(lxc storage bucket list s3 --all-projects | grep -cF bucket1)" = 0 ]
+  [ "$(lxc network list --all-projects | grep -cF foonet)" = 0 ]
+  [ "$(lxc network acl list --all-projects | grep -cF acl1)" = 0 ]
+  [ "$(lxc network zone list --all-projects | grep -cF zone1)" = 0 ]
+  [ "$(lxc profile list --all-projects | grep -cF profile1)" = 0 ]
+  [ "$(lxc image list --all-projects | grep -cF testimage)" = 0 ]
+  [ "$(lxc list --all-projects | grep -cF c1)" = 0 ]
```
simondeziel: If I tracked things properly, everything should be gone except for the uplink network, if OVN was enabled, right?

I'd move the OVN cleanup bit right after the forced project deletion of `foo` and then I'd check all the resources to all be empty.

```
[ "$(lxc storage volume list "${pool}" --all-projects || echo fail)" = "" ]
[ "$(lxc storage bucket list s3 --all-projects || echo fail)" = "" ]
...
```

lxd/api_project.go
kadinsayani: Thanks, I've added an `error` return value to `getEntityDeleter()` to reduce the indentation.

test/suites/projects.sh
kadinsayani: You got it. I've moved the OVN clean up steps up.

I want to explicitly search for the entity names while using `--all-projects` in the checks, hence why I've piped the output to `grep -cF` instead of using the new `[ "$(cmd || echo fail)" = "" ]` pattern.

test/suites/clustering.sh
```diff
@@ -4557,3 +4557,97 @@ test_clustering_trust_add() {
...
+  LXD_DIR="${LXD_ONE_DIR}" lxc network zone create zone1 --project foo
+
+  echo "Create profile in project."
+  LXD_DIR="${LXD_ONE_DIR}" lxc profile create profile1 --project foo
+
+  echo "Add image to project."
+  LXD_DIR="${LXD_ONE_DIR}" deps/import-busybox --project foo --alias testimage
+
+  echo "Create instance in project on node1."
+  LXD_DIR="${LXD_ONE_DIR}" lxc init testimage c1 --project foo --target node1 -s pool1
```
simondeziel: I'd use `--empty` container here.

test/suites/clustering.sh
```diff
@@ -4557,3 +4557,97 @@ test_clustering_trust_add() {
...
+
+  echo "Create another instance on node2."
+  LXD_DIR="${LXD_ONE_DIR}" lxc init testimage c2 --project foo --target node2 -s pool1
```
simondeziel: Same here.

test/suites/clustering.sh
```diff
@@ -4557,3 +4557,97 @@ test_clustering_trust_add() {
...
+  setup_clustering_netns 1
+  LXD_ONE_DIR=$(mktemp -d -p "${TEST_DIR}" XXX)
+  ns1="${prefix}1"
+  spawn_lxd_and_bootstrap_cluster "${ns1}" "${bridge}" "${LXD_ONE_DIR}"
+
+  # Add a newline at the end of each line. YAML has weird rules.
+  cert=$(sed ':a;N;$!ba;s/\n/\n\n/g' "${LXD_ONE_DIR}/cluster.crt")
+
+  # Spawn a second node
+  setup_clustering_netns 2
```
simondeziel, line 4577: Good call to make a 2 member cluster.

test/suites/clustering.sh
```diff
@@ -4557,3 +4557,97 @@ test_clustering_trust_add() {
...
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc network acl list --all-projects | grep -cF acl1)" = 0 ]
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc network zone list --all-projects | grep -cF zone1)" = 0 ]
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc profile list --all-projects | grep -cF profile1)" = 0 ]
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc image list --all-projects | grep -cF testimage)" = 0 ]
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc list --all-projects | grep -cF c1)" = 0 ]
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc list --all-projects | grep -cF c2)" = 0 ]
+
+  echo "Check project is deleted from both nodes."
+  ! LXD_DIR="${LXD_ONE_DIR}" lxc project show foo || false
+  ! LXD_DIR="${LXD_TWO_DIR}" lxc project show foo || false
```
simondeziel, line 4632: I'd move that right after the force deletion, before checking entities.

test/suites/clustering.sh
```diff
@@ -4557,3 +4557,97 @@ test_clustering_trust_add() {
...
+
+  echo "Check all entities are deleted from both nodes."
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc storage volume list --all-projects | grep -cF vol1)" = 0 ]
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc storage volume list --all-projects | grep -cF vol2)" = 0 ]
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc network acl list --all-projects | grep -cF acl1)" = 0 ]
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc network zone list --all-projects | grep -cF zone1)" = 0 ]
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc profile list --all-projects | grep -cF profile1)" = 0 ]
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc image list --all-projects | grep -cF testimage)" = 0 ]
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc list --all-projects | grep -cF c1)" = 0 ]
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc list --all-projects | grep -cF c2)" = 0 ]
```
simondeziel: How about doing a "before" and "after" comparison instead of counting entities that should be removed.

```
VOLUMES_BEFORE="$(LXD_DIR="${LXD_ONE_DIR}" lxc storage volume list -f csv -c n --all-projects)"
ACLS_BEFORE="..."
...

# project creation
# entities creation
# force removal

VOLUMES_AFTER="$(LXD_DIR="${LXD_ONE_DIR}" lxc storage volume list -f csv -c n --all-projects)"
ACLS_AFTER="..."

echo "Verify entities were all cleaned up"
[ "${VOLUMES_BEFORE}" = "${VOLUMES_AFTER}" ]
[ "${ACLS_BEFORE}" = "${ACLS_AFTER}" ]
...
```

If you don't like the "before" vs "after" way, could you then keep those `grep` but use faster and limited CSV listings, like this:

```
[ "$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c n --all-projects | grep -c "c[12]")" = 0 ]
```

test/suites/projects.sh
simondeziel: Same suggestion about "before" vs "after" comparison and if you don't like it, please go with `-f csv -c n` listings alone. Thanks!

test/suites/clustering.sh
kadinsayani: I've gone with the `list -f csv -c n` approach as it's consistent with other `lxc <entity> list` checks. If you'd prefer the before/after method, happy to go with that too :)

test/suites/projects.sh
```diff
@@ -1415,3 +1415,87 @@ test_projects_backups_volume() {
...
+    lxc network set "${uplink_network}" ipv4.ovn.ranges=192.0.2.100-192.0.2.254
+    lxc network set "${uplink_network}" ipv6.ovn.ranges=2001:db8:1:2::100-2001:db8:1:2::254
+
+    echo "Set IP routes that include OVN ranges."
+    lxc network set "${uplink_network}" ipv4.routes=192.0.2.0/24
+    lxc network set "${uplink_network}" ipv6.routes=2001:db8:1:2::/64
+
+    echo "Create OVN network in project."
+    lxc network create foonet --type ovn --project foo network="${uplink_network}"
+  fi
```
simondeziel, line 1489: Should a plain bridge be created in an `else` here? If you add one, please use `ipv[46].address=none` for speed ;)

kadinsayani: Thanks for catching that, added :)

test/suites/projects.sh
```diff
@@ -1415,3 +1415,90 @@ test_projects_backups_volume() {
...
+  ! lxc project delete foo || false
+
+  echo "Check force delete of non-existent project."
+  ! lxc project delete nonexistent --force --yes || false
+
+  echo "Check force delete of default project fails."
+  ! lxc project delete default --force --yes || false
+
+  echo "Create and start instance."
+  pool="lxdtest-$(basename "${LXD_DIR}")"
```
simondeziel: `pool` is already defined earlier.

test/suites/projects.sh
```diff
@@ -1415,3 +1415,90 @@ test_projects_backups_volume() {
...
+  lxc launch testimage c1 --project foo -s "${pool}"
```
simondeziel: I'm not familiar with projects. Do they come with a profile that has a NIC such that `c1` is hooked to `foonet`?

test/suites/projects.sh
```diff
@@ -1415,3 +1415,90 @@ test_projects_backups_volume() {
...
+  echo "Create project with all features enabled."
+  lxc project create foo -c features.networks=true -c features.networks.zones=true -c features.images=true -c features.profiles=true -c features.storage.volumes=true -c features.storage.buckets=true
+
+  echo "Create storage volume in project."
+  pool="lxdtest-$(basename "${LXD_DIR}")"
+  lxc storage volume create "${pool}" custom/vol1 --project foo
+
+  echo "Create storage bucket in project."
+  create_object_storage_pool s3
+  lxc storage bucket create s3 bucket1 --project foo
```
simondeziel, line 1441: If you could move that along with the `pool=...` bit at the very beginning of the function that'd be cleaner IMHO.

This would bring more symmetry with the cleanup section.

test/suites/projects.sh
```diff
@@ -1415,3 +1415,90 @@ test_projects_backups_volume() {
   lxc project delete foo
   lxc storage volume delete "${pool}" vol
 }
+
+test_projects_force_delete() {
+  echo "Create project with all features enabled."
+  lxc project create foo -c features.networks=true -c features.networks.zones=true -c features.images=true -c features.profiles=true -c features.storage.volumes=true -c features.storage.buckets=true
+
+  echo "Create storage volume in project."
+  pool="lxdtest-$(basename "${LXD_DIR}")"
```
simondeziel: Should be the first line of the function, IMHO.

test/suites/projects.sh
```diff
@@ -1415,3 +1415,90 @@ test_projects_backups_volume() {
...
+  echo "Force delete project."
+  lxc project delete foo --force --yes
+
+  echo "Check project is deleted."
+  ! lxc project show foo || false
+
+  echo "Clean up OVN parent network."
+  if ovn_enabled; then
+    lxc network delete "${uplink_network}"
+    ip link delete dummy0
```
simondeziel, line 1521:
```suggestion
    ip link delete dummy0
    unset_ovn_configuration
```

test/suites/projects.sh
kadinsayani: The default project-scoped profile is empty so a network must be specified when creating a project-scoped instance. I looked into project-scoped networks further and I don't think we can create a project-scoped bridge network, only OVN is supported. I'll remove the `else` case when OVN is not enabled.

simondeziel: I see. How about adding the OVN network to the default profile in `foo`? This way, it'd be used by `c1` for the OVN case.

test/suites/projects.sh
```diff
@@ -1415,3 +1415,88 @@ test_projects_backups_volume() {
...
+
+  echo "Force delete project."
+  lxc project delete foo --force --yes
+
+  echo "Check project is deleted."
+  ! lxc project show foo || false
+
+  echo "Clean up OVN parent network."
+  if ovn_enabled; then
+    lxc network delete "${uplink_network}"
```
simondeziel, line 1520: I presume that this acts as the validation that `foonet` was deleted along with the project. I think that having that network being used by `c1` would also make sure that even networks being used are deleted.

test/suites/projects.sh
kadinsayani: Good idea, thanks. While testing this locally, I noticed that some of the `lxc <entity> list -f csv -c n` commands fail locally, but not on CI. This is because not all `lxc <entity> list` commands support column-shorthand chars. Additionally, I think we need to capture output and use the `cmd || echo fail` pattern.

I'll go ahead with the before/after check you previously suggested :)

simondeziel: Ah, right, indeed not all support the `-c n`.

test/suites/projects.sh
kadinsayani: Thanks for the great suggestions, and for the thorough review :)

This got me thinking about the potential for race conditions due to the order of entity deletion. I'll conduct additional manual testing and assess if adding sorting by entity types and/or additional `force` parameters to `do*Delete()` functions is required.

kadinsayani: I think we should be ok since entities are collected by `projectUsedBy()`, which gathers entity URLs in a deterministic order, but better to catch any potential timing issues now.

kadinsayani: The failing tests confirm my race condition hypothesis. I'll investigate further.

test/suites/projects.sh
```diff
@@ -1416,3 +1416,122 @@ test_projects_backups_volume() {
...
+  ! lxc project delete foo || false
+
+  echo "Check force delete of non-existent project."
+  ! lxc project delete nonexistent --force --yes || false
+
+  echo "Check force delete of default project fails."
+  ! lxc project delete default --force --yes || false
+
+  echo "Create and start instance."
+  lxc launch testimage c1 --project foo -s "${pool}" -p profile1
```
simondeziel, line 1501: I'd like to see a test with an instance with `security.protection.delete=true`.

shared/entity/ref.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/canonical/lxd/shared/api"
+)
+
+// Reference represents a canonical entity reference.
+type Reference struct {
+	EntityType  Type
+	ProjectName string
+	Location    string
+	PathArgs    []string
+	url         *api.URL
```
tomponline, line 17: why a pointer?

shared/entity/ref.go
```diff
@@ -0,0 +1,85 @@
...
+		if len(r.PathArgs) >= 1 {
+			return r.PathArgs[0], true
+		}
+	}
+
+	return "", false
+}
+
+// StorageVolumeParts returns pool name, volume type, and volume name for storage volumes.
+func (r Reference) StorageVolumeParts() (poolName string, volType string, volName string, ok bool) {
```
tomponline: this function and StorageBucketParts stood out to me as unexpected in this file, why only these "*Part()" functions are needed?

lxd/api_project.go
```diff
@@ -1078,6 +1078,104 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+	})
+	if err != nil {
+		return err
+	}
+
+	// Parse entity URLs.
+	defaultProfile := api.NewURL().Path(version.APIVersion, "profiles", api.ProjectDefaultName).Project(projectName).String()
+	toDelete := make([]entity.Reference, 0, len(usedBy))
+	for _, u := range usedBy {
+		if u == defaultProfile {
```
tomponline, line 1110: this should have a comment i think - especially as in theory the user deleting the project might not have `auth.EntitlementCanDelete` on it, and yet is allowed to delete the project and its default profile.

cc @markylaing

lxd/api_project.go
```diff
@@ -1078,6 +1078,104 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+
+		// Build an entity reference directly from the URL.
+		ref, err := entity.ReferenceFromURL(*parsedURL)
+		if err != nil {
+			return err
+		}
+
+		// Skip instances and storage volumes not hosted on this node when a location is specified.
+		// These entities will be deleted by the node they are hosted on when a delete request is a cluster notification.
+		if ref.Location != "" && ref.Location != s.ServerName {
```
tomponline: Can we just do `ref.Location != s.ServerName` as for non-clusters, both should be "None" right?

lxd/api_project.go
```diff
@@ -1078,6 +1078,104 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+			continue
+		}
+
+		projectSpecific, err := ref.EntityType.RequiresProject()
+		if err != nil {
+			return err
```
tomponline: wrap error?

lxd/api_project.go
```diff
@@ -1078,6 +1078,104 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+		}
+
+		// Skip entities that are not project-specific.
+		if !projectSpecific {
+			continue
```
tomponline, line 1128: what entities can reference a project (such that they are in this project's usedBy list) that are not themselves project specific? Can you give me an example?

lxd/api_project.go
```diff
@@ -1078,6 +1078,104 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+		err = s.Authorizer.CheckPermission(ctx, ref.URL(), auth.EntitlementCanDelete)
+		if err != nil {
+			return err
+		}
+
+		toDelete = append(toDelete, ref)
+	}
+
+	// Sort entities by type to ensure proper deletion order (instances first, then profiles).
+	slices.SortFunc(toDelete, func(a, b entity.Reference) int {
```
tomponline, line 1141: Nice!

lxd/api_project.go
```diff
@@ -1078,6 +1078,104 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+		}
+
+		return getPriority(a.EntityType) - getPriority(b.EntityType)
+	})
+
+	// Delete entities.
+	for _, ref := range toDelete {
+		deleter, err := getEntityDeleter(ref.EntityType)
+		if err != nil {
+			return err
```
tomponline: wrap error?

lxd/api_project.go
```diff
@@ -1078,6 +1078,104 @@ func projectNodeConfigDelete(d *Daemon, s *state.State, name string) error {
...
+		}
+
+		err = deleter.Delete(ctx, s, ref)
+		if err != nil {
+			return err
```
tomponline: wrap error?

lxd/api_project.go
```diff
@@ -1124,20 +1236,22 @@ func projectDelete(d *Daemon, r *http.Request) response.Response {
...
 			return fmt.Errorf("Fetch project %q: %w", name, err)
 		}
 
-		empty, err := projectIsEmpty(ctx, project, tx)
-		if err != nil {
-			return err
-		}
+		if !force {
+			// Verify the project is empty.
+			empty, err := projectIsEmpty(ctx, project, tx)
```
tomponline, line 1241: I was just looking at projectIsEmpty and wondering how we can de-dupe this special "isn't the default profile" logic. Maybe we can get `projectIsEmpty` to call `isProjectInUse`?

lxd/api_project.go
```diff
@@ -1239,7 +1239,7 @@ func projectDelete(d *Daemon, r *http.Request) response.Response {
 	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
 		project, err := dbCluster.GetProject(ctx, tx.Tx(), name)
 		if err != nil {
-			return fmt.Errorf("Fetch project %q: %w", name, err)
+			return fmt.Errorf("Failed loading project %q: %w", name, err)
```
tomponline, line 1236: thanks

lxc/project.go
```diff
@@ -199,6 +203,9 @@ func (c *cmdProjectDelete) command() *cobra.Command {
 	cmd.Long = cli.FormatSection(i18n.G("Description"), i18n.G(
 		`Delete projects`))
 
+	cmd.Flags().BoolVarP(&c.flagForce, "force", "f", false, i18n.G("Force delete project and its entities (instances, profiles, images, networks, network ACLs, network zones, storage volumes, and storage buckets)")+"``")
```
tomponline: Lets not list all entities here, this will likely go out of date quickly and is also server version dependent.

lxc/project.go
```diff
@@ -199,6 +203,9 @@ func (c *cmdProjectDelete) command() *cobra.Command {
...
+	cmd.Flags().BoolVar(&c.flagNonInteractive, "yes", false, i18n.G("Don't require user confirmation for using --force"))
```
tomponline: we dont have this for `lxc delete --force`, `lxc rebuild --force` `lxc cluster remove --force`, `lxc network acl rule remove --force`, `lxc network forward port remove --force`.

So this seems unnecessary?

tomponline: Oh wait, we do have it indeed for `lxc cluster remove` i take it back. TIL.

tomponline: I suppose as the damage impact from an accidental forced project removal can be very bad, so makes some sense to require interactive approval.

Although I would argue that needing `--force --interactive no` is still just as easy to mess up when changing a previously executed commands as just needing `--force`.

Doesn't seem to offer much value to me.

shared/entity/ref.go
kadinsayani: For the other entities, we just need the name from the path parts. For storage volumes, we need storage pool name, volume type, and volume name. For storage buckets, we need the storage pool name and bucket name. Instead of adding these 2 functions, I can add a `PathParts(numParts int)` function and include comments at the call-sites indicating which path parts we need for different entities.

lxd/api_project.go
kadinsayani: This check is redundant. There are no entities that are in a project's used by list that are not project specific. I'll remove this bit.

lxd/api_project.go
kadinsayani: Yes, that should work.

lxc/project.go
kadinsayani: Sounds good, ack.

lxc/project.go
kadinsayani: Agreed, I'll drop the confirmation bit.

kadinsayani:
> Thanks @kadinsayani
> 
> Overall looks great, ive added some questions & suggestions.

Thanks for the review :)

shared/entity/ref.go
kadinsayani: For interoperability with existing functions in `lxd/db/cluster/entities.go`.

lxd/api_project.go
kadinsayani: I checked and this doesn't work because the location does not get set to "none"; it's empty.

shared/entity/ref.go
tomponline: Great idea!

lxd/api_project.go
kadinsayani: Yeah that would work well!

kadinsayani: Thanks for the suggestion.

lxd/api_project.go
kadinsayani: There's a couple reasons why the location is empty:

1) 

```Go
func (e entityTypeStorageVolume) allURLsQuery() string {
	return fmt.Sprintf(`
SELECT 
	%d, 
	storage_volumes.id, 
	projects.name, 
	replace(coalesce(nodes.name, ''), 'none', ''), 
...
```

An empty string is returned.

2)

```Go
// Target sets the "target" query parameter in the URL if the clusterMemberName is not empty or "default".
func (u *URL) Target(clusterMemberName string) *URL {
	if clusterMemberName != "" && clusterMemberName != "none" {
		queryArgs := u.Query()
		queryArgs.Add("target", clusterMemberName)
		u.RawQuery = queryArgs.Encode()
	}

	return u
}
```

The location isn't set by `Target()` when the provided arg is "none".

kadinsayani: I think we can stick with the current guard which first checks if the location is not empty, and then compares it to the server name.

shared/entity/ref.go
tomponline: Hrm I dont see the need for a pointer, the Reference struct doesn't need to container a pointer inside it.

tomponline: It can be converted to a pointer where needed if really necessary.

shared/entity/ref.go
```diff
@@ -0,0 +1,65 @@
...
+type Reference struct {
+	EntityType  Type
+	ProjectName string
+	Location    string
+	PathArgs    []string
+	url         *api.URL
+}
+
+// NewReference constructs a [Reference] and validates it by attempting to build the canonical URL.
+func NewReference(projectName string, entityType Type, location string, pathArgs ...string) (Reference, error) {
```
tomponline, line 21: Cant this return a pointer to `Reference` so that errors can return nil.

This is the convention we use (unless returning an interface as they are pointers themselves).

shared/entity/ref.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return Reference{}, err
+	}
+
+	return NewReference(projectName, entityType, location, pathArgs...)
+}
+
+// URL returns an [*api.URL] for this [Reference].
+func (r Reference) URL() *api.URL {
+	return r.url
```
tomponline, line 46: this can be `return &r.url`

lxd/api_project.go
```diff
@@ -1146,19 +1254,31 @@ func projectDelete(d *Daemon, r *http.Request) response.Response {
...
-	// Send notification to other cluster members to update the node schema.
+	// Send notification to other cluster members to update the node schema and handle force deletion (if requested).
 	notifier, err := cluster.NewNotifier(s, s.Endpoints.NetworkCert(), s.ServerCert(), cluster.NotifyAlive)
 	if err != nil {
 		return response.SmartError(err)
 	}
 
 	err = notifier(func(member db.NodeInfo, client lxd.InstanceServer) error {
+		if force {
+			return client.ForceDeleteProject(name)
```
tomponline, line 1279: in the future please order the commits such that the called functions are defined before they are called. Ta

tomponline (approved):
LGTM thanks!

Please open a follow up wrt to the pointer comments I made. Ta

lxd/api_project.go
kadinsayani: Ack :)

client/interfaces.go
```diff
@@ -342,6 +342,7 @@ type InstanceServer interface {
 	UpdateProject(name string, project api.ProjectPut, ETag string) (err error)
 	RenameProject(name string, project api.ProjectPost) (op Operation, err error)
 	DeleteProject(name string) (err error)
+	ForceDeleteProject(name string) (err error)
```
tomponline, line 345: @kadinsayani when i was reviewing your multi-volume PR and I got to the Go SDK changes bit I noticed we had `DeleteClusterMember(name string, force bool) (err error)`.

So I think we should remove `ForceDeleteProject` and update `DeleteProject()` to accept a `force bool` argument.
We can return an error if the server doesn't have the specific API extension.

tomponline: I realise this is a Go SDK API break, but as we are approaching the next LTS soon, now is the time to make these changes.

kadinsayani: Sounds good! I'll open up a follow-up PR addressing this.

kadinsayani: https://github.com/canonical/lxd/pull/16636

PR #16536, opened 2025-09-16
lxd: extract doStoragePoolVolumeDelete()

kadinsayani:


lxd/storage_volumes.go
```diff
@@ -2478,7 +2478,7 @@ func doStoragePoolVolumeDelete(ctx context.Context, s *state.State, name string,
 	case cluster.StoragePoolVolumeTypeImage:
 		err = pool.DeleteImage(name, op)
 	default:
-		return api.StatusErrorf(http.StatusBadRequest, `Storage volumes of type %q cannot be deleted with the storage API`, volType.String())
+		return api.StatusErrorf(http.StatusBadRequest, `Invalid storage volume type %q`, volType.String())
```
tomponline: This new error seems more misleading that the old error - the new error suggests the volume type is invalid (i.e its not defined), but that isn't necessarily true.

How about:

```
return api.StatusErrorf(http.StatusBadRequest, `Storage volumes of type %q cannot be deleted directly`, volType.String())
```

kadinsayani: Sounds good, I've gone with your suggestion. We can also remove that default case since we check the volume type at the beginning of the function and return an error there.

tomponline (approved):
Cheers!

PR #16547, opened 2025-09-18
Auth: Enforce user can view project when listing resources

markylaing:
Also, enforce that a group cannot be granted entitlements on a project, or any entity in that project if members of the group cannot view the project.

Also a bugfix that I'll put in a separate PR tomorrow.

### DIFF
--- a/lxd/db/images_test.go
+++ b/lxd/db/images_test.go
@@ -83,7 +83,7 @@ func TestGetImage(t *testing.T) {
 		id, img, err := tx.GetImage(ctx, "a", cluster.ImageFilter{Project: &project})
 		require.NoError(t, err)
 		assert.True(t, img.Public)
-		assert.NotEqual(t, id, -1)
+		assert.NotEqual(t, -1, id)
 
 		// non-public image with 'default' project
 		err = tx.CreateImage(ctx, project, "abcd2", "x.gz", 16, false, false, "amd64", time.Now(), time.Now(), map[string]string{}, "container", nil)
@@ -101,7 +101,7 @@ func TestGetImage(t *testing.T) {
 		id, img, err = tx.GetImage(ctx, "a", cluster.ImageFilter{Project: &project, Public: &public})
 		require.NoError(t, err)
 		assert.True(t, img.Public)
-		assert.NotEqual(t, id, -1)
+		assert.NotEqual(t, -1, id)
 
 		return nil
 	})


markylaing:
Will need to rebase after https://github.com/canonical/lxd/pull/16551 is merged. Then will look again at test failures and open for review

markylaing:
Ready for review but depends on #16551 so will need rebasing when that is merged.

tomponline:
@markylaing can be rebased

markylaing:
@tomponline Have rebased. Previous test run passed (though ceph was still running when I pushed the rebase).

lxd/auth/types.go
```diff
@@ -48,6 +48,9 @@ type Authorizer interface {
 	//
 	// Warn: You almost never need this function. You should use GetPermissionChecker instead.
 	GetPermissionCheckerWithoutEffectiveProject(ctx context.Context, entitlement Entitlement, entityType entity.Type) (PermissionChecker, error)
+
+	// GetViewableProjects accepts a list of permissions and returns a list of projects that a member of a group with these permissions is able to view.
+	GetViewableProjects(ctx context.Context, permissions []api.Permission) ([]string, error)
```
tomponline, line 53: Fell at the first hurdle :)

> accepts a list of permissions and returns a list of projects that a member of a group with these permissions is able to view.

I found this pretty cryptic and couldn't reason about it after several attempts.

What permissions around I pass into this?

test/suites/devlxd.sh
```diff
@@ -291,7 +292,7 @@ test_devlxd_volume_management() {
     vol1='{\"name\": \"vol-01\", \"type\": \"custom\", \"config\": {\"size\": \"10MiB\"}}'
 
     # Create a custom storage volume (fail - insufficient permissions).
-    [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage create-volume "${pool}" "${vol1}")" = "Not Found" ]
+    [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage create-volume "${pool}" "${vol1}")" = "Forbidden" ]
```
tomponline, line 295: is this correct?

lxd/daemon.go
```diff
@@ -317,70 +317,76 @@ func allowPermission(entityType entity.Type, entitlement auth.Entitlement, muxVa
 
 // allowProjectResourceList should be used instead of allowAuthenticated when listing resources within a project.
 // This prevents a restricted TLS client from listing resources in a project that they do not have access to.
-func allowProjectResourceList(d *Daemon, r *http.Request) response.Response {
-	requestor, err := request.GetRequestor(r.Context())
-	if err != nil {
-		return response.SmartError(err)
-	}
+func allowProjectResourceList(allowAllProjects bool) func(d *Daemon, r *http.Request) response.Response {
```
tomponline, line 321: allowAllProjects param is not documented in the doc block

tomponline: is this basically controlling whether to allow access to endpoint of all-projects=true?

lxd/auth/types.go
markylaing: > What permissions around I pass into this?

I don't follow sorry.

In the API, only groups can be assigned permissions. Identities are granted access only via the groups they are a member of. Given a list of permissions for a group, this function returns a list of project names that a member of only that group would be able to see.

test/suites/devlxd.sh
markylaing: Yes. On line 64 I've granted view access to the project. So rather than seeing a "Not found" here. The caller sees Forbidden (they can view the project but don't have `can_create_storage_volumes` on it).

lxd/daemon.go
markylaing: This controls whether a restricted TLS client is allowed to use `all-projects=true`

lxd/auth/types.go
tomponline: OK right, so this function takes one or more permissions (like server admin or can_view on a particular project) and then returns a list of projects that the permissions give you access to `view`?

tomponline: So ultimately is this just doing the same `lxc project list`?

lxd/daemon.go
markylaing: Added to the doc block

lxd/auth/types.go
tomponline: But iiuc this is not based on a user's permissions, but a group's permissions?

lxd/auth/drivers/tls.go
```diff
@@ -34,6 +34,11 @@ func (t *tls) load(ctx context.Context, identityCache *identity.Cache, opts Opts
 	return nil
 }
 
+// GetViewableProjects is not implemented for the TLS authorizer.
+func (t *tls) GetViewableProjects(ctx context.Context, permissions []api.Permission) ([]string, error) {
+	return nil, api.NewStatusError(http.StatusNotImplemented, "not implemented")
```
tomponline: Capital letter at start of errors please

lxd/auth/drivers/openfga.go
```diff
@@ -113,6 +116,83 @@ func (e *embeddedOpenFGA) load(ctx context.Context, identityCache *identity.Cach
...
+	// Use a different context from the given context.
+	// Why? Because the OpenFGA datastore will attempt to load a cache that is optimised for permission checking an individual request.
+	// In this case we are asking the datastore to enumerate projects based on contextual tuples. The cache is not useful for this, so we don't
+	// want to trigger loading it.
+	var cancel context.CancelFunc
+	deadline, ok := ctx.Deadline()
+	if ok {
+		ctx, cancel = context.WithTimeout(context.Background(), time.Until(deadline))
+	} else {
+		ctx, cancel = context.WithTimeout(context.Background(), 5*time.Second)
```
tomponline: is that the 5s timeout creeping back in again?

lxd/auth/drivers/openfga.go
```diff
@@ -113,6 +116,83 @@ func (e *embeddedOpenFGA) load(ctx context.Context, identityCache *identity.Cach
...
+	}
+
+	defer cancel()
+
+	// Construct a list objects request using a dummy identity as a member of a dummy group.
```
tomponline, line 132: I was just thinking in:

https://github.com/canonical/lxd/pull/16547/commits/91d26d11f5837f56e27aed98e2de2108b2b60265#r2362665840

Is this the same as `lxc project list`, but the only problem is we have an identity there, whereas in the case of group management we don't have an identity yet. Then I thought I wonder if we could split apart `projectsGet()` and feed it a psuedo identity via the request context to save implementing part of `projectsGet` here.

lxd/auth/drivers/openfga.go
```diff
@@ -113,6 +116,83 @@ func (e *embeddedOpenFGA) load(ctx context.Context, identityCache *identity.Cach
...
+					Object:   groupObject,
+				},
+			},
+		},
+	}
+
+	// Add a contextual tuple for each given permission
+	for _, permission := range permissions {
+		req.ContextualTuples.TupleKeys = append(req.ContextualTuples.TupleKeys, &openfgav1.TupleKey{
+			User:     groupObject + "#member", // Members of the dummy group have permission, not the group itself.
```
tomponline: is the "#member" here the same as the "member" string above, can these be a local constant (with "#" prepended) to show the relationship?

lxd/auth/drivers/openfga.go
```diff
@@ -113,6 +116,83 @@ func (e *embeddedOpenFGA) load(ctx context.Context, identityCache *identity.Cach
...
+		if err != nil {
+			return nil, fmt.Errorf("Invalid project url %q returned from list object request", err)
+		}
+
+		_, project, _, _, err := entity.ParseURL(*u)
+		if err != nil {
+			return nil, fmt.Errorf("Invalid project url %q returned from list object request", err)
+		}
+
+		projects = append(projects, project)
```
tomponline, line 188: @markylaing aside from this PR in question, I feel like this technique could be useful in other places.

The fact that you can generate a list of objects of a certain entity type the "user" has access to could be beneficial in the API handlers too, as we could then avoid needing to load all entities and filtering certain ones out, and instead change the API handlers to only load the ones needed:

https://github.com/canonical/lxd/blob/cea4064a353c71c8d6c885071cdd376a9a08bd10/lxd/api_project.go#L167-L174

lxd/auth/types.go
markylaing: > So ultimately is this just doing the same lxc project list?

Yes but it does it while impersonating an identity that is a member of a group with only these permissions.

> But iiuc this is not based on a user's permissions, but a group's permissions?

Yep. This is why we can't use the permission checker, as we don't want the projects that the current identity can view.

lxd/auth/drivers/openfga.go
markylaing: Ah yes will remove. In fact.. I think a better solution here might be to explicitly set the openfga cache to nil

lxd/auth/drivers/openfga.go
markylaing: Unfortunately that wouldn't work. The access checking in getting a permission checker only uses groups that have been defined in the database. Whereas this function doesn't require the group to be defined yet, allowing it to be used for validation.

lxd/auth/drivers/openfga.go
markylaing: This could work well but we would need to be extremely careful about project feature flags. 

Perhaps this is something that can be more easily achieved after rewiring the authorizer to use entity IDs instead

tomponline:
@markylaing linter failures

markylaing:
> @markylaing linter failures

Should be fixed now (one I should have caught, the others didn't show up for me - had to update my local golangci-lint).

lxd/auth/drivers/openfga.go
markylaing: Ah I forgot to do this. I'll fix now and push again. Then we should be ready.

tomponline (approved):
Thanks!

tomponline:
@markylaing with this change would you expect `lxc list --project=a-non-existent-project` to return a 404 "project not found" type error rather than 200 OK with an empty list?

cc @escabo @simondeziel

tomponline:
@markylaing there looks like still an inconsistency between `lxc list`, `lxc image list`, `lxc operation list` and other listing commands that list entities belonging to projects.

e.g.

```
lxc list --project=a-non-existent-project
<no error, blank output>

lxc image list lxc list --project=a-non-existent-project
+-------+-------------+--------+-------------+--------------+------+------+-------------+
| ALIAS | FINGERPRINT | PUBLIC | DESCRIPTION | ARCHITECTURE | TYPE | SIZE | UPLOAD DATE |
+-------+-------------+--------+-------------+--------------+------+------+-------------+
<no error, blank output>

lxc operation list --project=a-non-existent-project
+----+------+-------------+--------+------------+---------+
| ID | TYPE | DESCRIPTION | STATUS | CANCELABLE | CREATED |
+----+------+-------------+--------+------------+---------+
<no error, blank output>
```
_Also there's an unrelated inconsistency in the CLI display output of these commands when there is no results, 2 of them display an empty table, but `lxc list` displays nothing_

Compared to:
```
lxc network list  --project=a-non-existent-project
Error: Failed loading project "a-non-existent-project": Project not found

lxc profile list --project=a-non-existent-project
Error: Failed loading project "a-non-existent-project": Project not found

lxc network acl list --project=a-non-existent-project
Error: Failed loading project "a-non-existent-project": Project not found
```

I understood we are wanting all such listing commands to exhibit the latter behaviour (returning an error) rather than no-error and an empty list (unless called with `--all-projects` flag).

Should we log a new issue for this?

cc @escabo @simondeziel

markylaing:
Will log a new issue and track in JIRA as it sounds like there are some cases I missed here or didn't test well enough.

PR #16548, opened 2025-09-18
shared/filter: dereference pointer and handle nil pointer in ValueOf()

kadinsayani:
Closes https://github.com/canonical/lxd/issues/16546.


tomponline (approved):
ta

PR #16551, opened 2025-09-18
test/suites: When asserting empty output, get output and test on different lines

markylaing:
We set -e in the test suite to ensure an exit on error. Unfortunately, as far as I can tell -e doesn't apply inside a test. The man page for bash (https://man7.org/linux/man-pages/man1/bash.1.html) has:

> The shell does not exit if the
> command that fails is part of the command list
> immediately following a while or until reserved
> word, part of the test following the if or elif
> reserved words, part of any command executed in a
> && or || list except the command following the
> final && or ||, any command in a pipeline but the
> last (subject to the state of the pipefail shell
> option), or if the command's return value is being
> inverted with !.

Which to me indicates that it should be failing, but at least locally for me it is not.

This means that any test that runs `[ "$(cmd)" = "" ]` is not erroring out it `cmd` fails, and the test is asserting that the stdout from `cmd` is empty. This is usually true if the command fails, as it will send output to stderr instead.

As I see it, the only fix for this is to get the output on one line and test it on the next.

### DIFF
--- a/test/lint/empty-test-fail.sh
+++ b/test/lint/empty-test-fail.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+set -eu
+set -o pipefail
+
+# We set -e in the test suite to ensure an exit on error. Unfortunately,
+# `-e` doesn't apply inside a test.
+#
+# This means that any test that runs `[ "$(cmd)" = "" ]` is not erroring
+# out if cmd fails, and the test is asserting that the stdout from cmd
+# is empty. This is usually true if the command fails, as it will send
+# output to stderr instead.
+#
+# This linter enforces that any tests of this kind are of the form `[ "$(cmd)" = "" || echo fail ]`.
+# This means that if the command fails, bash will execute the OR statement
+# and echo "fail", causing the test to fail.
+out="$(git grep --untracked -E '^\s+\[\s+"\$\(.*\)"\s+=\s+["'\'']{2}\s+]' test/ | grep -Fv '|| echo fail)' || true)"
+if [ "${out}" != '' ] ; then
+  # shellcheck disable=SC2016
+  echo 'One or more assertions of type [ "$(cmd)" = '' ] require `|| echo fail`'
+  echo '=========='
+  echo "${out}"
+  exit 1
+fi
+
+exit 0
--- a/test/suites/basic.sh
+++ b/test/suites/basic.sh
@@ -148,16 +148,13 @@ test_basic_usage() {
 
   # Test container rename
   lxc move foo bar
+  [ "$(lxc list -c n -f csv)" = "bar" ]
 
   # Check volatile.apply_template is altered during rename.
   [ "$(lxc config get bar volatile.apply_template)" = "rename" ]
 
-  [ "$(lxc list -c n | grep -F foo)" = "" ]
-  [ "$(lxc list -c n | grep -F bar)" != "" ]
-
   lxc rename bar foo
-  [ "$(lxc list -c n | grep -F bar)" = "" ]
-  [ "$(lxc list -c n | grep -F foo)" != "" ]
+  [ "$(lxc list -c n -f csv)" = "foo" ]
   lxc rename foo bar
 
   # Test container copy
--- a/test/suites/basic.sh
+++ b/test/suites/basic.sh
@@ -765,7 +765,7 @@ EOF
   lxc launch testimage c3 --ephemeral
 
   lxc stop -f c1 c2 c3
-  [ "$(lxc list -f csv -c n)" = "" ]
+  [ "$(lxc list -f csv -c n || echo fail)" = "" ]
 
   # Cleanup
   fingerprint="$(lxc config trust ls --format csv | cut -d, -f4)"
--- a/test/README.md
+++ b/test/README.md
@@ -49,7 +49,7 @@ test_vm() {
   # Ephemeral cleanup
   lxc launch --vm --empty --ephemeral v1 -c limits.memory=128MiB -d "${SMALL_ROOT_DISK}"
   lxc stop -f v1
-  [ "$(lxc list -f csv -c n)" = "" ]
+  [ "$(lxc list -f csv -c n || echo fail)" = "" ]
 }
 ```
 
@@ -64,7 +64,7 @@ Good:
   echo "==> Ephemeral cleanup"
   lxc launch --vm --empty --ephemeral v1 -c limits.memory=128MiB -d "${SMALL_ROOT_DISK}"
   lxc stop -f v1
-  [ "$(lxc list -f csv -c n)" = "" ]
+  [ "$(lxc list -f csv -c n || echo fail)" = "" ]
 ```
 
 This way, debug logs will be easier to read.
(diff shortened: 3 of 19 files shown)


markylaing:
Lo and behold the tests should never have been passing! https://github.com/canonical/lxd/actions/runs/17823329320/job/50671315261#step:17:17732

markylaing:
Tests are passing now.

tomponline:
> Error: Storage pool driver "ceph" does not support buckets

simondeziel:
> This means that any test that runs `[ "$(cmd)" = "" ]` is not erroring out it `cmd` fails, and the test is asserting that the stdout from `cmd` is empty. This is usually true if the command fails, as it will send output to stderr instead.
> 
> As I see it, the only fix for this is to get the output on one line and test it on the next.

How about this:

```
[ "$(lxc storage bucket list oidc: --all-projects --format csv || echo fail)" = "" ]
```

markylaing:
> > This means that any test that runs `[ "$(cmd)" = "" ]` is not erroring out it `cmd` fails, and the test is asserting that the stdout from `cmd` is empty. This is usually true if the command fails, as it will send output to stderr instead.
> > As I see it, the only fix for this is to get the output on one line and test it on the next.
> 
> How about this:
> 
> ```
> [ "$(lxc storage bucket list oidc: --all-projects --format csv || echo fail)" = "" ]
> ```

That seems like a difficult thing to remember to do. I'd be tempted to start using `lxc query` with `jq` instead as it feels more robust

simondeziel:
> > > This means that any test that runs `[ "$(cmd)" = "" ]` is not erroring out it `cmd` fails, and the test is asserting that the stdout from `cmd` is empty. This is usually true if the command fails, as it will send output to stderr instead.
> > > As I see it, the only fix for this is to get the output on one line and test it on the next.
> > 
> > 
> > How about this:
> > ```
> > [ "$(lxc storage bucket list oidc: --all-projects --format csv || echo fail)" = "" ]
> > ```
> 
> That seems like a difficult thing to remember to do. I'd be tempted to start using `lxc query` with `jq` instead as it feels more robust

I agree on the hard to remember part. I'd be happy to add a linter check that flags all the empty stdout comparison ( ` ...)" = "" ]`) that are missing the `|| echo fail` bit. I prefer the `|| echo fail` alternative to splitting on two lines. WDYT?

Re `lxc query` + `jq` sounds like a good move too.

tomponline:
> Re `lxc query` + `jq` sounds like a good move too.

the downside of that is that it wont help with our CLI code coverage metrics.

markylaing:
> > > > This means that any test that runs `[ "$(cmd)" = "" ]` is not erroring out it `cmd` fails, and the test is asserting that the stdout from `cmd` is empty. This is usually true if the command fails, as it will send output to stderr instead.
> > > > As I see it, the only fix for this is to get the output on one line and test it on the next.
> > > 
> > > 
> > > How about this:
> > > ```
> > > [ "$(lxc storage bucket list oidc: --all-projects --format csv || echo fail)" = "" ]
> > > ```
> > 
> > 
> > That seems like a difficult thing to remember to do. I'd be tempted to start using `lxc query` with `jq` instead as it feels more robust
> 
> I agree on the hard to remember part. I'd be happy to add a linter check that flags all the empty stdout comparison ( ` ...)" = "" ]`) that are missing the `|| echo fail` bit. I prefer the `|| echo fail` alternative to splitting on two lines. WDYT?
> 
> Re `lxc query` + `jq` sounds like a good move too.

Ok happy to go with this and will add a linter to catch future issues

test/suites/basic.sh
```diff
@@ -148,16 +148,13 @@ test_basic_usage() {
...
   # Check volatile.apply_template is altered during rename.
   [ "$(lxc config get bar volatile.apply_template)" = "rename" ]
 
-  [ "$(lxc list -c n | grep -F foo)" = "" ]
-  [ "$(lxc list -c n | grep -F bar)" != "" ]
-
   lxc rename bar foo
-  [ "$(lxc list -c n | grep -F bar)" = "" ]
-  [ "$(lxc list -c n | grep -F foo)" != "" ]
+  [ "$(lxc list -c n -f csv)" = "foo" ]
```
simondeziel, line 157: Ah, nice!

test/suites/basic.sh
```diff
@@ -152,11 +152,11 @@ test_basic_usage() {
   # Check volatile.apply_template is altered during rename.
   [ "$(lxc config get bar volatile.apply_template)" = "rename" ]
 
-  [ "$(lxc list -c n | grep -F foo)" = "" ]
+  [ "$(lxc list -c n | grep -F foo || echo fail)" = "" ]
   [ "$(lxc list -c n | grep -F bar)" != "" ]
 
   lxc rename bar foo
-  [ "$(lxc list -c n | grep -F bar)" = "" ]
+  [ "$(lxc list -c n | grep -F bar || echo fail)" = "" ]
```
simondeziel: I preferred your previous fix of just checking for ` = "foo" ]`

test/lint/empty-test-fail.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash -eu
+
+out="$(grep --include \*.sh -rE '^\s+\[\s+"\$\(.*\)"\s+=\s+""\s+]' .)"
```
simondeziel:
```suggestion
out="$(git grep --untracked -E '^\s+\[\s+"\$\(.*\)"\s+=\s+["'\'']{2}\s+]' test/)"
```

This covers both `[ "$(cmd)" = "" ]` and `[ "$(cmd)" = '' ]`.

test/lint/empty-test-fail.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash -eu
```
simondeziel: It's best to put the `set -eu` on a separated line to also apply if someone uses `bash test/lint/empty-test-fail.sh` for example.

I'd also add the `set -o pipefail` while at it.

test/lint/empty-test-fail.sh
markylaing: Updated the linter, it now uses this:
```
out="$(git grep --untracked -E '^\s+\[\s+"\$\(.*\)"\s+=\s+["'\'']{2}\s+]' test/ | grep -Fv '|| echo fail)' || true)"
```

1. Grep for all `[ $(cmd)" = "" ] ` or `[ "$(cmd)" = '' ].
2. Remove occurrences where `|| echo fail` is used.
3. Don't error if no occurrences are found, or if all occurrences have `|| echo fail`.

test/lint/empty-test-fail.sh
markylaing: Done

tomponline (approved):
Thanks!

markylaing:
@tomponline this is ready when you are, thanks.

markylaing:
> @tomponline this is ready when you are, thanks.

Oh I missed that you'd already approved it :laughing:

tomponline:
> > @tomponline this is ready when you are, thanks.
> 
> Oh I missed that you'd already approved it 😆

yeah i wanted to make sure the tests passed ;)

simondeziel:
Thanks @markylaing that's a really nice improvement!

PR #16553, opened 2025-09-18
client/lxd_server: Copy all fields in UseProject and UseTarget

jonathan-conder:
Currently the `httpUnixPath` field isn't copied over. This should hopefully prevent further issues of this type!


jonathan-conder:
Sure, I gave it a go

PR #16565, opened 2025-09-19
Fix OIDC scopes default

edlerd:
* Adding a default string to the docs for the server config `oidc.scopes`, `core.bgp_asn` and `instances.migration.stateful`.


markylaing (approved):
Thanks!

markylaing (dismissed):
Can you please update all the descriptions in separate commits and then update the metadata at the end. Thanks

tomponline (approved):
ta

PR #16567, opened 2025-09-19
Storage: Enable Pure recovery

roosterfish:
Implements Pure's `ListVolumes` function to enable recovery.

PR #16568, opened 2025-09-19
Storage: Recovery followups 

roosterfish:
Addresses the comment here https://github.com/canonical/lxd/pull/16567#discussion_r2362523668.

Also I somehow missed properly testing VM recovery :(. The `container` type was always the default (int 0) so I didn't encounter the issue during volume normalization.

Just a few more lines and missing bits to get VM recovery working too for both PowerFlex & Pure.

I have now cross tested everything, also by having Pure & PowerFlex instances running in parallel, having vols on PowerFlex attached to Pure backed instances and so on.

### DIFF
--- a/lxd/storage/drivers/driver_pure_volumes.go
+++ b/lxd/storage/drivers/driver_pure_volumes.go
@@ -997,10 +997,10 @@ func (d *pure) ListVolumes() ([]Volume, error) {
 		contentType := ContentTypeFS
 		if volType == VolumeTypeCustom && strings.HasSuffix(volName, pureContentTypeSuffixes[ContentTypeISO]) {
 			contentType = ContentTypeISO
-			volName = strings.TrimSuffix(volName, pureContentTypeSuffixes[ContentTypeISO])
+			volName = strings.TrimSuffix(volName, "-"+pureContentTypeSuffixes[ContentTypeISO])
 		} else if volType == VolumeTypeVM || isBlock {
 			contentType = ContentTypeBlock
-			volName = strings.TrimSuffix(volName, pureContentTypeSuffixes[ContentTypeBlock])
+			volName = strings.TrimSuffix(volName, "-"+pureContentTypeSuffixes[ContentTypeBlock])
 		}
 
 		volUUID, err := d.getUUIDFromVolumeName(volName)
--- a/lxd/storage/drivers/driver_pure_volumes.go
+++ b/lxd/storage/drivers/driver_pure_volumes.go
@@ -1010,7 +1010,7 @@ func (d *pure) ListVolumes() ([]Volume, error) {
 		}
 
 		// This is important to allow subsequent operations on the volume struct (e.g. driver's HasVolume) to be able to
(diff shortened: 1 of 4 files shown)


roosterfish (commented):
I have the feeling we should make this consistent across drivers.
Ceph and PowerFlex have the delimiter included in the constant, Pure and Alletra don't. 

Including the delimiter in the constant make it easier to use later on as you don't have to manually add the `-` everytime you use any of the suffixes.

tomponline:
> I have the feeling we should make this consistent across drivers.

make what consistent?

lxd/storage/drivers/driver_pure_volumes.go
```diff
@@ -997,20 +997,20 @@ func (d *pure) ListVolumes() ([]Volume, error) {
 		contentType := ContentTypeFS
 		if volType == VolumeTypeCustom && strings.HasSuffix(volName, pureContentTypeSuffixes[ContentTypeISO]) {
 			contentType = ContentTypeISO
-			volName = strings.TrimSuffix(volName, pureContentTypeSuffixes[ContentTypeISO])
+			volName = strings.TrimSuffix(volName, "-"+pureContentTypeSuffixes[ContentTypeISO])
```
roosterfish, line 1000: I have the feeling we should make this consistent across drivers.
Ceph and PowerFlex have the delimiter included in the constant, Pure and Alletra don't.

Including the delimiter in the constant make it easier to use later on as you don't have to manually add the `-` everytime you use any of the suffixes.

roosterfish: @tomponline that should have been a code comment, removed the original one.

tomponline: We can for HPE, but isn't it too late for Pure, as there could be running VMs on upgrade that prevent patching volume name renames?

roosterfish: It's just how we handle it in the code. The volume name on storage is identical.

Ceph and PowerFlex use constants like `.block` or `.b` but Pure and Alletra just do `b` so you always have to think about adding the delimiter (`.`, `-`) when dealing with the suffix.

tomponline: oh well it should be standardised then

roosterfish: Happy to land this in a follow up PR for Pure & Alletra if this is ok for you. Because we should again run all test suites then just in case we miss something as it touches integral parts of the code.

Whereas this just sets Pure's `ListVolume` to the format used today.

tomponline: Well we cant on HPE

tomponline: Suggest landing it carefully and @mionaalex mentioned HP are going to do some tests themselves later

tomponline (approved):

Issue #16392, opened 2025-09-05
API should not allow creating a certificate add token with an empty client name


markylaing:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

22.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
edge build 8262f9d0987b1616c10bdffb1f694682bf7d0eff
```

### Output of "lxc info" or system info if it fails

```shell
-
```

### Issue description

It is possible via the API to create a certificate add token that does not contain a client name. This should not be allowed, because the `shared.CertificateTokenDecode` function does not consider a token to be valid unless a name is present.

https://github.com/canonical/lxd/blob/8262f9d0987b1616c10bdffb1f694682bf7d0eff/shared/cert.go#L531-L533

### Steps to reproduce

```
$ curl -s -k --cert client.crt --key client.key https://127.0.0.1:8443/1.0/certificates -X POST -d '{"token": true, "type": "client"}' | jq .
{
  "type": "async",
  "status": "Operation created",
  "status_code": 100,
  "operation": "/1.0/operations/86e9aeb1-26db-4004-a02f-62a95ee43542",
  "error_code": 0,
  "error": "",
  "metadata": {
    "id": "86e9aeb1-26db-4004-a02f-62a95ee43542",
    "class": "token",
    "description": "Executing operation",
    "created_at": "2025-09-05T15:59:40.142670986+01:00",
    "updated_at": "2025-09-05T15:59:40.142670986+01:00",
    "status": "Running",
    "status_code": 103,
    "resources": {},
    "metadata": {
      "addresses": [
        "127.0.0.1:8443"
      ],
      "fingerprint": "118a5add840380b57293fff8c054009d0351a0db370edb609383ecfff43336b2",
      "request": {
        "name": "",
        "type": "client",
        "restricted": false,
        "projects": [],
        "certificate": "",
        "password": "",
        "trust_token": "",
        "token": true
      },
      "secret": "ef4f124ee0f340d65aa9a40674890d05138932ab04f46fd1c322cd7e771c12a1"
    },
    "may_cancel": true,
    "err": "",
    "location": "none"
  }
}
```

### Information to attach

- [ ] Any relevant kernel output (`dmesg`)
- [ ] Instance log (`lxc info NAME --show-log`)
- [ ] Instance configuration (`lxc config show NAME --expanded`)
[...]

PR #16572, opened 2025-09-19
lxd: Reject certificate add token with an empty client name

nmezhenskyi:
Fixes https://github.com/canonical/lxd/issues/16392.

## Changes:
Extend client name validation during certificate add token in the `POST /1.0/certificates` request to reject empty values.

### DIFF
--- a/lxd/certificates.go
+++ b/lxd/certificates.go
@@ -490,12 +490,12 @@ func certificatesPost(d *Daemon, r *http.Request) response.Response {
 
 	localHTTPSAddress := s.LocalConfig.HTTPSAddress()
 
-	// Quick check.
-	if req.Token && req.Certificate != "" {
-		return response.BadRequest(errors.New("Can't use certificate if token is requested"))
-	}
-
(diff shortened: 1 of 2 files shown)


lxd/certificates.go
```diff
@@ -500,6 +500,10 @@ func certificatesPost(d *Daemon, r *http.Request) response.Response {
 			return response.BadRequest(errors.New("Tokens can only be issued for client certificates"))
 		}
 
+		if req.Name == "" {
```
simondeziel, line 503: While you are in there, would you mind merging the outer `if req.Token {` block with the one just before?

simondeziel (commented):
I know it's still in draft but just in case you had not plan for it, having a quick `curl` to test this would make a nice addition ;)

lxd/certificates.go
nmezhenskyi: Updated :)

simondeziel (approved):
LGTM, thanks!

tomponline (approved):
Ta!

PR #16586, opened 2025-09-22
devlxd: Report whether server is clustered to authenticated users

MusicDin:
When CSI is deployed on non-clustered LXD, we need to avoid setting `target` to server name in the request, as it would result in `Cluster member not found`. However, we are currently unable to determine that from the information provided by DevLXD.

This PR introduces `server_name` field that would report that, but only if the client is authenticated.

In addition, it introduces `DevLXDGetUntrusted` struct (similar to `ServerUntrusted` for LXD) to avoid returning fields that are not populated for the untrusted client.

### DIFF
--- a/test/suites/devlxd.sh
+++ b/test/suites/devlxd.sh
@@ -250,6 +250,12 @@ test_devlxd_volume_management() {
     # Fail when token is not passed.
     [ "$(lxc exec "${inst}" --project "${project}" -- devlxd-client instance get "${inst}")" = "You must be authenticated" ]
 
+    # Ensure "environment" is not included in the API response for unauthenticated clients.
+    # When using LXD go-client, default values are used for missing fields, so "environment.server_clustered" will be false.
+    [ "$(lxc exec "${inst}" --project "${project}" -- devlxd-client get-state | jq -e '.environment.server_clustered')" = "false" ]
+    # However, "environment" must be missing in the API response.
+    [ "$(lxc exec "${inst}" --project "${project}" -- devlxd-client query GET /1.0 | jq -e '.environment')" = "null" ]
+
     # Fail when a valid identity token is passed, but the identity does not have permissions.
     lxc auth identity create "${authIdentity}"
     token=$(lxc auth identity token issue "${authIdentity}" --quiet)
@@ -265,6 +271,8 @@ test_devlxd_volume_management() {
     lxc auth group permission add "${authGroup}" instance "${inst}" can_view project="${project}"
     lxc auth identity group add "${authIdentity}" "${authGroup}"
     lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client instance get "${inst}" | jq -e .name
+    [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client get-state | jq -e '.environment.server_clustered')" = "false" ]
+    [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client query GET /1.0 | jq -e '.environment.server_clustered')" = "false" ]
 
     # Test devLXD authorization (volume management security flag).
     # Fail when the security flag is not set.


shared/api/devlxd.go
```diff
@@ -59,6 +63,14 @@ type DevLXDServerStorageDriverInfo struct {
...
 }
 
+// DevLXDServerEnvironment represents the read-only environment fields of a DevLXD server.
+type DevLXDServerEnvironment struct {
+	// Whether the underlying LXD server is part of a cluster.
+	// This information is populated only for trusted clients.
+	// Example: false
+	ServerClustered *bool `json:"server_clustered" yaml:"server_clustered"`
+}
+
```
MusicDin, line 78: @tomponline Note that I've set `ServerClustered` to *bool, which means `null` will be returned to unauthenticated users.

I could return `false` (default bool value), but this may be misleading, as some client may not realize it is not authenticated and do subsequent operations based on the reported value that LXD is not clustered

shared/api/devlxd.go
```diff
@@ -46,6 +46,10 @@ type DevLXDGet struct {
 	// Example: [{"name": "zfs", "remote": false}, {"name": "ceph", "remote": true}]
 	// API extension: devlxd_volume_management
 	SupportedStorageDrivers []DevLXDServerStorageDriverInfo `json:"supported_storage_drivers" yaml:"supported_storage_drivers"`
+
+	// Read-only information about DevLXD server environment.
+	// API extension: devlxd_volume_management
```
MusicDin, line 56: I've reported that as part of the `devlxd_volume_management` API extension because it has not been released yet. Please confirm whether this is acceptable. Thanks.

shared/api/devlxd.go
```diff
@@ -46,6 +46,10 @@ type DevLXDGet struct {
...
+	DevLXDServerEnvironment `json:"environment" yaml:"environment"`
```
tomponline: On type Server this field is called `Environment` and is of a type called `ServerEnvironment` rather than embedding, can we do this here too for consistent?

shared/api/devlxd.go
```diff
@@ -59,6 +63,14 @@ type DevLXDServerStorageDriverInfo struct {
 	Remote bool `json:"remote" yaml:"remote"`
 }
 
+// DevLXDServerEnvironment represents the read-only environment fields of a DevLXD server.
+type DevLXDServerEnvironment struct {
+	// Whether the underlying LXD server is part of a cluster.
+	// This information is populated only for trusted clients.
+	// Example: false
+	ServerClustered *bool `json:"server_clustered" yaml:"server_clustered"`
```
tomponline: why is this a pointer to a bool, whereas in ServerEnvironment its just a bool?

shared/api/devlxd.go
tomponline: yes thats fine thanks

shared/api/devlxd.go
tomponline: I think we should look at how its done for the main /1.0 and mirror that rather than changing data types.

shared/api/devlxd.go
MusicDin: Dup of: https://github.com/canonical/lxd/pull/16586#discussion_r2368435667

shared/api/devlxd.go
MusicDin: Hmm. Currently we have a DevLXD client function `GetState() (api.DevLXDGet, error)`. The `api.DevLXDGet` would be populated with default values anyway, which means the unauthenticated client would get:
```json
{
  ...
  "environment": {
    "server_clustered": false
  }
}
```

I just thought this may be misleading

tomponline: In my view better to behave consistently between external and devlxd APIs than to introduce subtle differences between the two.

tomponline: Also this is a Go SDK problem, not a REST API issue right?

tomponline: @MusicDin what does the external /1.0 REST API raw output look like today for untrusted requests?

MusicDin: Correct. This is fixed now. Note that LXD Agent checks `state.Auth` field, and if it indicates the untrusted user, it only returns untrusted part of the response.

shared/api/devlxd.go
MusicDin: Thanks for spotting that. Fixed.

MusicDin:
Hmm, no curl in our test image.
```
+++ timeout --foreground 120 /home/runner/go/bin/lxc exec devlxd-volume-mgmt-container --project devlxd-volume-mgmt --verbose --force-local -- curl --unix-socket /dev/lxd/sock lxd/1.0
Error: Command not found
+ '[' '' = null ']'
```

I've commented the test. However, it works both for VMs and containers. 
```sh
$ lxc exec v1 -- curl --unix-socket /dev/lxd/sock lxd/1.0 | jq
{
  "state": "Started",
  "api_version": "1.0",
  "instance_type": "virtual-machine",
  "location": "recovery",
  "auth": "untrusted",
  "supported_storage_drivers": []
}

$ lxc exec c1 -- curl --unix-socket /dev/lxd/sock lxd/1.0 | jq
{
  "state": "Started",
  "api_version": "1.0",
  "instance_type": "container",
  "location": "recovery",
  "auth": "untrusted",
  "supported_storage_drivers": []
}
```

And if authenticated:
```sh
token=$(lxc auth identity token issue devlxd/csi --quiet)

$ lxc exec v1 -- curl --unix-socket /dev/lxd/sock -H "Authorization: Bearer ${token}" lxd/1.0 | jq
{
  "state": "Started",
  "api_version": "1.0",
  "instance_type": "virtual-machine",
  "location": "recovery",
  "auth": "trusted",
  "supported_storage_drivers": [],
  "environment": {
    "server_clustered": false
  }
}

$ lxc exec c1 -- curl --unix-socket /dev/lxd/sock -H "Authorization: Bearer ${token}" lxd/1.0 | jq
{
  "state": "Started",
  "api_version": "1.0",
  "instance_type": "container",
  "location": "recovery",
  "auth": "trusted",
  "supported_storage_drivers": [],
  "environment": {
    "server_clustered": false
  }
}
```

MusicDin:
@tomponline @markylaing This is ready for review. 

I've commented the test that checks the `environment` is not included in API response when client is not authenticated as test image does not have `curl` (see comment https://github.com/canonical/lxd/pull/16586#issuecomment-3320159815). I will make sure to include this in LXD CI, where VM tests will run.

markylaing:
`devlxd-client` has a `get-state` function which I added to test devlxd tokens. Should be able to use that rather than curling /1.0

MusicDin:
> devlxd-client has a get-state function which I added to test devlxd tokens. Should be able to use that rather than curling /1.0

Yeah, we need plain API response. Go will populated missing fields with default values (field `environment` and sub-field `server_clustered` are not pointers).

MusicDin:
I've added `devlxd-client query` command that allows arbitrary DevLXD API invocations.

Issue #16561, opened 2025-09-18
docs: state variable bullet point in duplicated


ethandcosta:
In the ACL Rules section, information about the state variable being enabled by default is duplicated in the second and last bullet points. Since the last BP is more informative, it would be better to get rid of this second BP. 

---
*Reported from: https://documentation.ubuntu.com/lxd/latest/howto/network_acls/*

CLodBard:
@tomponline Can I get this one ?

tomponline:
Go for it! Thanks :)

PR #16590, opened 2025-09-23
DOC: Changes to ACL Rules bullet point in Network ACLs

CLodBard:
- Addresses issue #16561 (State value being enabled by default is mentioned twice)

- Changed the formatting of the bullet points, as it became a bit unclear on what was the Key and was was the Value in some cases. Some lines don't have any emphasis on  Key or Value, while some had double quotes for both Keys and Values and some had double quotes only for Values.
   -- Changed it to **BOLD** for Keys and "double quotes" for Values, across the list

### DIFF
--- a/doc/howto/network_acls.md
+++ b/doc/howto/network_acls.md
@@ -313,14 +313,13 @@ state: <enabled|disabled|logged>
 
 `````
 
-- The `action` property is required.
-- The `state` property defaults to `"enabled"` if unset.
-- The `source` and `destination` properties can be specified as one or more CIDR blocks, IP ranges, or {ref}`selectors <network-acls-selectors>`. If left empty, they match any source or destination. Comma-separate multiple values.
-- If the `protocol` is unset, it matches any protocol.
-- The `"destination_port"` and `"source_port"` properties and `"icmp_code"` and `"icmp_type"` properties are mutually exclusive sets. Although both sets are shown in the same rule above to demonstrate the syntax, they never appear together in practice.
-   - The `"destination_port"` and `"source_port"` properties are only available when the `"protocol"` for the rule is `"tcp"` or `"udp"`.
-   - The [`"icmp_code"`](https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml#icmp-parameters-codes) and [`"icmp_type"`](https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml#icmp-parameters-types) properties are only available when the `"protocol"` is `"icmp4"` or `"icmp6"`.
-- The `"state"` is `"enabled"` by default. The `"logged"` value is used to {ref}`log traffic <network-acls-log>` to a rule.
+- The **`action`** property is required.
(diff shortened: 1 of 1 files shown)


tomponline:
@CLodBard please can you rebase and drop the non-doc related commits?

doc/howto/network_acls.md
```diff
@@ -313,14 +313,13 @@ state: <enabled|disabled|logged>
...
+- The **`action`** property is required.
+- The **`source`** and **`destination`** properties can be specified as one or more CIDR blocks, IP ranges, or {ref}`selectors <network-acls-selectors>`. If left empty, they match any source or destination. Comma-separate multiple values.
+- If the **`protocol`** is unset, it matches any protocol.
+- The **`destination_port`** and **`source_port`** properties and **`icmp_code`** and **`icmp_type`** properties are mutually exclusive sets. Although both sets are shown in the same rule above to demonstrate the syntax, they never appear together in practice.
+   - The **`destination_port`** and **`source_port`** properties are only available when the **`protocol`** for the rule is `"tcp"` or `"udp"`.
+   - The [**`icmp_code`**](https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml#icmp-parameters-codes) and [**`icmp_type`**](https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml#icmp-parameters-types) properties are only available when the **`protocol`** is `"icmp4"` or `"icmp6"`.
+- The **`state`** is `"enabled"` by default. The `"logged"` value is used to {ref}`log traffic <network-acls-log>` to a rule.
```
minaelee:
```suggestion
- The **`action`** property is required.
- The **`source`** and **`destination`** properties can be specified as one or more CIDR blocks, IP ranges, or {ref}`selectors <network-acls-selectors>`. If left empty, they match any source or destination. Comma-separate multiple values.
- If the **`protocol`** is unset, it matches any protocol.
- The **`destination_port`** and **`source_port`** properties and **`icmp_code`** and **`icmp_type`** properties are mutually exclusive sets. Although both sets are shown in the same rule above to demonstrate the syntax, they never appear together in practice.
   - The **`destination_port`** and **`source_port`** properties are only available when the **`protocol`** for the rule is `tcp` or `udp`.
   - The [**`icmp_code`**](https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml#icmp-parameters-codes) and [**`icmp_type`**](https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml#icmp-parameters-types) properties are only available when the **`protocol`** is `icmp4` or `icmp6`.
- The **`state`** is `enabled` by default. The `logged` value is used to {ref}`log traffic <network-acls-log>` to a rule.
```

minaelee: Thanks for pointing out the inconsistency with quotes. Let's go ahead and format the values _without_ quotes (just in code-format backticks) since the default tab shown is for YAML, and YAML does not require quotes around strings. JSON does, but JSON also requires quotes around the keys.

CLodBard: Thank you for the notes Minae, applied them to the commit and should now be reflecting the changes you mentioned.

tomponline:


@CLodBard ^

tomponline:
Thanks!

PR #16591, opened 2025-09-23
lxd: Extract operation operation access check to function

markylaing:
Follow to #16453 to deduplicate the operation access check.

### DIFF
--- a/lxd/operations.go
+++ b/lxd/operations.go
@@ -158,33 +158,14 @@ func operationGet(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
-	requestor, err := request.GetRequestor(r.Context())
-	if err != nil {
-		return response.SmartError(err)
-	}
-
 	var body *api.Operation
 
 	// First check if the query is for a local operation from this node
 	op, err := operations.OperationGetInternal(id)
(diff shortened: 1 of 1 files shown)


lxd/operations.go
```diff
@@ -1087,6 +1047,42 @@ func operationWaitGet(d *Daemon, r *http.Request) response.Response {
 	return response.ForwardedResponse(client)
 }
 
+func checkOperationViewAccess(ctx context.Context, op *operations.Operation, authorizer auth.Authorizer, secret string) error {
+	// If a secret is provided and it matches the operation, allow access.
+	if secret != "" && op.Metadata()["secret"] == secret {
```
tomponline, line 1052: This seems to go against your commit message here:

> Previously, we relied on the fact that operation IDs are UUIDs and are
therefore not guessable. So if the user cannot list operations via
/1.0/operations, they can't get the operation. This does not account for
a bad actor acquiring an operation ID by other means.

https://github.com/canonical/lxd/pull/16453/commits/eb4bf028b47fb1d0e8999425dac6b8a93bb4ffd2

tomponline: Oh I see the UUIDs are not secret now, but the secret still is ofcourse.

PR #16593, opened 2025-09-23
User agent parser

markylaing:
Adds a user agent parser for LXD user agents (see `shared/version/useragent.go`). This will be used by #15030, which will inspect the list of features to determine if a client that sends an OIDC access token as a bearer token is able to store cookies.

Note the commit adding storage pool driver name constants - I didn't want to use string literals. I can prepare a few follow up PRs replacing other string literals throughout the codebase if desired.

### DIFF
--- a/lxd/request/user_agent.go
+++ b/lxd/request/user_agent.go
@@ -1,0 +1,278 @@
+package request
+
+import (
+	"errors"
+	"fmt"
+	"slices"
+	"strings"
+	"unicode"
+
+	"github.com/canonical/lxd/shared"
+)
+
+// UserAgent represents a LXD user agent.
+type UserAgent struct {
+	Product  UserAgentProduct
+	Host     UserAgentHost
+	Storage  map[string]string
+	Features []string
+}
+
+// UserAgentProduct contains information about the product (first part of user agent).
+type UserAgentProduct struct {
+	Name    string
+	Version string
+	LTS     bool
+}
+
+// UserAgentHost contains host information stored in the user agent.
+type UserAgentHost struct {
+	OS            string
+	Arch          string
+	KernelVersion string
+	Distro        string
+	DistroVersion string
+}
+
+// ParseUserAgent parses a user agent to return a [UserAgent].
+func ParseUserAgent(userAgent string) (*UserAgent, error) {
+	groups, err := splitUserAgent(userAgent)
+	if err != nil {
+		return nil, err
+	}
+
+	product, err := getProductInfo(groups[0])
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get product information: %w", err)
+	}
+
+	host, err := getHostInfo(groups[1])
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get host information: %w", err)
+	}
+
+	ua := &UserAgent{
+		Product: *product,
+		Host:    *host,
+	}
+
+	if len(groups) == 2 {
+		return ua, nil
+	}
+
+	storage, isFeatureGroup, err := tryGetStorageInfo(groups[2])
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get storage information: %w", err)
+	}
+
+	if storage != nil {
+		ua.Storage = storage
+	} else if isFeatureGroup {
+		features, err := getFeatureInfo(groups[2])
+		if err != nil {
+			return nil, fmt.Errorf("Failed to get feature information: %w", err)
+		}
+
+		ua.Features = features
+	}
+
+	if len(groups) == 3 {
+		return ua, nil
+	}
+
+	if storage == nil {
+		return nil, errors.New("Feature group may not precede storage group")
+	}
+
+	features, err := getFeatureInfo(groups[3])
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get feature information: %w", err)
+	}
+
+	ua.Features = features
+	return ua, nil
+}
+
+// getProductInfo returns the product, product_version and product_lts based on the basic User-Agent information.
+func getProductInfo(product string) (*UserAgentProduct, error) {
+	fields := strings.Fields(product)
+	if fields[0] != "LXD" {
+		return nil, errors.New("Only LXD user agents are currently supported")
+	}
+
+	if len(fields) < 2 {
+		return nil, errors.New("Product does not contain a version")
+	}
+
+	name := fields[0]
+	version := fields[1]
+
+	if len(fields) > 3 {
+		return nil, errors.New(`Product must be of the form "LXD <version> [LTS]"`)
+	}
+
+	var lts bool
+	if len(fields) == 3 {
+		if fields[2] != "LTS" {
+			return nil, errors.New(`Malformed product LTS field`)
+		}
+
+		lts = true
+	}
+
+	if !lts {
+		// 4.0 and 5.0 didn't have the LTS flag in the User-Agent string.
+		// So derive the LTS status from the major and minor version numbers and any patch level.
+		lts = strings.HasPrefix(version, "4.0.") || strings.HasPrefix(version, "5.0.")
+	}
+
+	p := UserAgentProduct{
+		Name:    name,
+		Version: version,
+		LTS:     lts,
+	}
+
+	return &p, nil
+}
+
+// getHostInfo returns the host_os, host_arch, host_kernel, host_distro and host_distro_version based on the User-Agent information.
+func getHostInfo(hostGroup string) (h *UserAgentHost, err error) {
+	// The host information contains: host_os, host_arch, host_kernel, host_distro, host_distro_version.
+	// Among those, only the host_os and host_arch are mandatory.
+	parts := shared.SplitNTrimSpace(hostGroup, "; ", 6, true)
+	if len(parts) < 2 {
+		return nil, errors.New("Host group must contain OS and architecture")
+	}
+
+	if len(parts) > 5 {
+		return nil, errors.New("Host group cannot contain more than 5 elements")
+	}
+
+	// If not enough parts were found, fill in the missing parts with empty strings.
+	if len(parts) < 5 {
+		parts = append(parts, make([]string, 5-len(parts))...)
+	}
+
+	h = &UserAgentHost{
+		OS:            parts[0],
+		Arch:          parts[1],
+		KernelVersion: parts[2],
+		Distro:        parts[3],
+		DistroVersion: parts[4],
+	}
+
+	return h, nil
+}
+
+// tryGetStorageInfo inspects the given group to return storage information. If the first element in the group does not
+// contain any whitespace, a boolean is returned to indicate that this not a storage group and should be parsed as a
+// feature group.
+func tryGetStorageInfo(group string) (map[string]string, bool, error) {
+	// The storage information is optional and may not be present in the User-Agent string.
+	if group == "" {
+		return nil, false, nil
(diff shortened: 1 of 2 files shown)


lxd/request/user_agent.go
```diff
@@ -0,0 +1,328 @@
...
+
+// getStorageOrFeatureInfo returns the Alletra, Btrfs, Ceph, CephFS, Dir, LVM, Powerflex, Pure and ZFS information based on the User-Agent information.
+func getStorageOrFeatureInfo(storageGroup string) ([]UserAgentStorage, []string, error) {
+	// The storage information is optional and may not be present in the User-Agent string.
+	if storageGroup == "" {
+		return nil, nil, nil
+	}
+
+	allDriverNames := api.StorageDriverNames()
+	parts := strings.Split(storageGroup, "; ")
```
simondeziel: A good candidate for `shared.SplitNTrimSpace()` maybe?

lxd/request/user_agent.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+
+	return groups, nil
+}
+
+// splitUserAgent splits the User-Agent string into groups.
+func splitUserAgent(ua string) ([]string, error) {
+	// Find where the basic information (`LXD X.Y[ LTS]`) ends and the extended part begins.
+	division := strings.Index(ua, " (")
+	if division == -1 || division < len("LXD X.Y") {
```
simondeziel: Why check for `division == -1`? I think the `division < len("LXD X.Y")` is a superset.

tomponline:
> Adds a user agent parser for LXD user agents (see `shared/version/useragent.go`). This will be used by #15030, which will inspect the list of features to determine if a client that sends an OIDC access token as a bearer token is able to store cookies.
> 
> Note the commit adding storage pool driver name constants - I didn't want to use string literals. I can prepare a few follow up PRs replacing other string literals throughout the codebase if desired.

I don't think we should be adding pool driver names to our api package, they are lxd server version dependent and not part of our api.

They should be (and are) defined server side and exposed through the 1.0 endpoint.

lxd/request/user_agent.go
```diff
@@ -0,0 +1,328 @@
...
+		case ')':
+			depth--
+			if depth < 0 {
+				return nil, errors.New("User agent contains unbalanced parentheses")
+			} else if depth == 0 && start < i {
+				groups = append(groups, ua[start+1:i])
+			}
+		}
+	}
+
```
simondeziel, line 229: Should it error if `len(groups) == 0`?

lxd/request/user_agent.go
```diff
@@ -0,0 +1,328 @@
...
+	// Product info and host info are mandatory.
+	groups = append(groups, basicUA)
+
+	// The host information is mandatory and always the first group.
+	infoGroups, err := extractParenthesesGroups(ua[division+1:])
+	if err != nil {
+		return nil, err
+	}
+
+	if len(infoGroups) > 3 {
```
simondeziel, line 259: Shouldn't we also check for `len(infoGroups) < 1` too?

lxd/request/user_agent.go
```diff
@@ -0,0 +1,328 @@
...
+
+// ParseUserAgent returns the information from the User-Agent string.
+func ParseUserAgent(userAgent string) (*UserAgent, error) {
+	groups, err := splitUserAgent(userAgent)
+	if err != nil {
+		return nil, err
+	}
+
+	product := getProductInfo(groups[0])
+	host, err := getHostInfo(groups[1])
```
simondeziel: I think we should check `len(groups)` before accessing indices > 0.

lxd/request/user_agent_test.go
```diff
@@ -0,0 +1,725 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/version"
+)
+
+func Test_getAllInfo(t *testing.T) {
```
simondeziel, line 11: Unit tests are cheap to run so why not have all the small functions tested individually?

markylaing:
> > Adds a user agent parser for LXD user agents (see `shared/version/useragent.go`). This will be used by #15030, which will inspect the list of features to determine if a client that sends an OIDC access token as a bearer token is able to store cookies.
> > Note the commit adding storage pool driver name constants - I didn't want to use string literals. I can prepare a few follow up PRs replacing other string literals throughout the codebase if desired.
> 
> I don't think we should be adding pool driver names to our api package, they are lxd server version dependent and not part of our api.
> 
> They should be (and are) defined server side and exposed through the 1.0 endpoint.

Where are they defined server side? I can only see string literals in `driver.Info()`. Even if they were defined as constants somewhere internally, I can't import them into this package if it is to be used externally.

lxd/request/user_agent.go
markylaing: It's an internal function and is caught in a later check.

lxd/request/user_agent.go
markylaing: As above - it's an internal function and is caught by the caller.

lxd/request/user_agent.go
markylaing: There are unit tests for this.

lxd/request/user_agent_test.go
markylaing: Unit tests for the only exported function have 100% coverage :)

tomponline:
> > > Adds a user agent parser for LXD user agents (see `shared/version/useragent.go`). This will be used by #15030, which will inspect the list of features to determine if a client that sends an OIDC access token as a bearer token is able to store cookies.
> > > Note the commit adding storage pool driver name constants - I didn't want to use string literals. I can prepare a few follow up PRs replacing other string literals throughout the codebase if desired.
> > 
> > 
> > I don't think we should be adding pool driver names to our api package, they are lxd server version dependent and not part of our api.
> > They should be (and are) defined server side and exposed through the 1.0 endpoint.
> 
> Where are they defined server side? I can only see string literals in `driver.Info()`. Even if they were defined as constants somewhere internally, I can't import them into this package if it is to be used externally.

They may not be defined as constants but we have a driver map that exposes the names via the /1.0 endpoint so they must be "defined" in the server.

They must not be in the `shared/api` package (for the reason I mentioned above).
We must be careful about not putting LXD server version specific code into the `shared/api` package (and in general the `shared` package). In fact we should ultimately be *removing* as much as possible from the `shared/*` packages to limit our external commitments from an API perspective.

> I can't import them into this package if it is to be used externally.

I think we have a different definition of "externally" here. 
For me `shared/api` is a contract we have with consumers outside of the LXD team.
But in this case the "external" consumer of this package is an internal (from an organisation perspective) tool.
Also the UA parser itself is in the `lxd/request` package and not in shared, which conflicts with the idea that tools outside of the LXD repo cannot import packages outside of `shared/*`.

Can you import the `lxd/storage/drivers` package and call `AllDriverNames()`?

tomponline:
> > > > Adds a user agent parser for LXD user agents (see `shared/version/useragent.go`). This will be used by #15030, which will inspect the list of features to determine if a client that sends an OIDC access token as a bearer token is able to store cookies.
> > > > Note the commit adding storage pool driver name constants - I didn't want to use string literals. I can prepare a few follow up PRs replacing other string literals throughout the codebase if desired.
> > > 
> > > 
> > > I don't think we should be adding pool driver names to our api package, they are lxd server version dependent and not part of our api.
> > > They should be (and are) defined server side and exposed through the 1.0 endpoint.
> > 
> > 
> > Where are they defined server side? I can only see string literals in `driver.Info()`. Even if they were defined as constants somewhere internally, I can't import them into this package if it is to be used externally.
> 
> They may not be defined as constants but we have a driver map that exposes the names via the /1.0 endpoint so they must be "defined" in the server.
> 
> They must not be in the `shared/api` package (for the reason I mentioned above). We must be careful about not putting LXD server version specific code into the `shared/api` package (and in general the `shared` package). In fact we should ultimately be _removing_ as much as possible from the `shared/*` packages to limit our external commitments from an API perspective.
> 
> > I can't import them into this package if it is to be used externally.
> 
> I think we have a different definition of "externally" here. For me `shared/api` is a contract we have with consumers outside of the LXD team. But in this case the "external" consumer of this package is an internal (from an organisation perspective) tool. Also the UA parser itself is in the `lxd/request` package and not in shared, which conflicts with the idea that tools outside of the LXD repo cannot import packages outside of `shared/*`.
> 
> Can you import the `lxd/storage/drivers` package and call `AllDriverNames()`?

On a related point, im not entirely clear why the user agent parser needs to know the possible storage driver names anyway? We could, in principle, remove a storage driver in future, but older versions of LXD might still have support for it.

The user agent says what pool drivers are in use, so why does the parser need to know the possible values?

markylaing:
> On a related point, im not entirely clear why the user agent parser needs to know the possible storage driver names anyway? We could, in principle, remove a storage driver in future, but older versions of LXD might still have support for it.

> The user agent says what pool drivers are in use, so why does the parser need to know the possible values?

What we can do instead is just expect that storage drivers contain a space, and enforce that features do not contain a space.

lxd/request/user_agent.go
```diff
@@ -0,0 +1,328 @@
...
+		// Only one product exists at the moment: "LXD".
+		Product:         "LXD",
+		Version:         v,
+		SemanticVersion: getUserAgentSemver(v),
+		LTS:             lts,
+	}
+
+	// Always expect our version to be a basic semantic version. If it isn't it's not LTS.
+	// This makes e.g. 5.0.2-qnap5 not an LTS version.
+	if p.SemanticVersion.String() != v {
```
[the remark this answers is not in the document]
markylaing: Addressed. No longer attempting to parse semantic version.

lxd/request/user_agent.go
markylaing: Done

lxd/request/user_agent.go
markylaing: Fixed, but different fix to suggestion so needs re-review :)

lxd/request/user_agent.go
markylaing: Added a comment explaining why it's not possible for the host group to be missing at this point.

lxd/request/user_agent.go
```diff
@@ -0,0 +1,328 @@
...
+			Name:            driverName,
+			Version:         v,
+			SemanticVersion: getUserAgentSemver(v),
+		}
+
+		driverList = append(driverList, driver)
+		foundDriverNames = append(foundDriverNames, driverName)
+	}
+
+	return driverList, nil, nil
```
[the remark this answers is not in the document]
markylaing: Now renamed function to `tryGetStorageInfo` which returns a boolean indicating if the group should be considered a feature group instead.

markylaing:
This is now updated such that:
- Driver names are not checked. Making sure these are valid is the responsibility of the caller.
- The difference between storage drivers and features is that storage drivers must contain a space to delimit the driver name and version, whereas features must not contain a space.
- Semantic versions are not parsed. We just save the entire contents of the version string (trimmed whitespace).
- Storage drivers are just a map of driver name to version string.

tomponline:
> * The difference between storage drivers and features is that storage drivers must contain a space to delimit the driver name and version, whereas features must not contain a space.

Should we enforce that in `UserAgentFeatures()` by updating it to return an error if any of the features have spaces in them?

PR #16595, opened 2025-09-23
Set a default for `core.remote token expiry`

simondeziel:
Follow-up on review comment: https://github.com/canonical/lxd/pull/16229#discussion_r2349532611

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2012,6 +2012,7 @@ This introduces the storage bucket API. It allows the management of S3 object st
 This updates the storage bucket API to return initial admin credentials at bucket creation time.
 
 ## `metrics_cpu_effective_total`
+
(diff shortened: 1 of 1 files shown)


doc/metadata.txt
```diff
@@ -4721,7 +4721,7 @@ If this option is not specified, LXD falls back to the `NO_PROXY` environment va
 ```
 
 ```{config:option} core.remote_token_expiry server-core
-:defaultdesc: "`15d`"
+:defaultdesc: "no expiry"
```
tomponline: Why changed back?

simondeziel: I likely edited the wrong file and the `make update-metadata` undid my change.

simondeziel: Fixed.

PR #16596, opened 2025-09-23
Add go version to copilot instructions

kadinsayani:


.github/copilot-instructions.md
```diff
@@ -4,6 +4,8 @@
 
 LXD is a modern, secure system container and virtual machine manager written in Go.
 
+LXD uses Go version 1.24.6
```
simondeziel: Thanks for picking that up!

Makefile
```diff
@@ -239,6 +239,12 @@ check-gomin:
 		echo "Please update the Go version in 'doc/requirements.md' to be $(GOMIN) instead of $(DOC_GOMIN)"; \
 		exit 1; \
 	fi
+	@echo "Check copilot-instructions.md mentions the right Go minimum version"
+	$(eval COPILOT_GOMIN := $(shell sed -n 's/^LXD uses Go version \([0-9.]\+\).*/\1/p' .github/copilot-instructions.md))
+	if [ "$(COPILOT_GOMIN)" != "$(GOMIN)" ]; then \
+		echo "Please update the Go version in '.github/copilot-instructions.md' to be $(GOMIN) instead of $(COPILOT_GOMIN)"; \
+		exit 1; \
+	fi
```
simondeziel, line 247: Now that we have 3 files to keep in sync, should we have a `make update-gomin` target that accepts a version to then update all 3 files and suggest to commit them?

kadinsayani: Added :)

Makefile
```diff
@@ -239,6 +239,38 @@ check-gomin:
...
+
+.PHONY: update-gomin
+update-gomin:
+ifndef NEW_GOMIN
+	@echo "Usage: make update-gomin NEW_GOMIN=1.x.y"
+	@echo "Current Go minimum version: $(GOMIN)"
+	exit 1
+endif
```
simondeziel, line 259: If `GOMIN == NEW_GOMIN` it should error out.

Also, if not error'ing out, we should unconditionally try to commit all 3 files.

Makefile
```diff
@@ -239,6 +239,38 @@ check-gomin:
...
+	@echo "Updating Go minimum version from $(GOMIN) to $(NEW_GOMIN)"
+	
+	@# Update GOMIN in Makefile
+	sed -i 's/^GOMIN=.*/GOMIN=$(NEW_GOMIN)/' Makefile
```
simondeziel: Is it `make` that allows the variable to be resolved even inside the single quotes?

Makefile
```diff
@@ -239,6 +239,38 @@ check-gomin:
...
+	
+	@# Update doc/requirements.md
+	sed -i 's/^LXD requires Go [0-9.]* or higher/LXD requires Go $(NEW_GOMIN) or higher/' doc/requirements.md
+	
+	@# Update .github/copilot-instructions.md
+	sed -i 's/^LXD uses Go version [0-9.]*/LXD uses Go version $(NEW_GOMIN)/' .github/copilot-instructions.md
```
simondeziel: I'd go with a safer regex than `*` gives us. Also, could we use the same wording to allow editing both files in one go?

```
sed -i 's/^\(LXD uses Go version\) [0-9.]+ /\1 $(NEW_GOMIN) /' doc/requirements.md .github/copilot-instructions.md
```

I'd also use the same version number matching bit for the `Makefile` one while at it.

Makefile
```diff
@@ -239,6 +239,38 @@ check-gomin:
...
+	
+	@echo "Go minimum version updated to $(NEW_GOMIN)"
+	if [ -t 0 ] && ! git diff --quiet -- ./Makefile ./doc/requirements.md ./.github/copilot-instructions.md; then \
```
simondeziel: As mentioned before, for this target to be successful, at this point, all 3 files need to have been altered so I require for all 3 to have a diff.

Makefile
kadinsayani: `$(NEW_GOMIN)` is preprocessed by `make` before the `shell` command is executed.

tomponline (approved):
LGTM!

Issue #14843, opened 2025-01-23
Containers do not start again when the host is not shut down properly


dbruno74:
LXD: latest/edge
OS: core24

After creating containers with the nvidia gpu device, and the host is not shut down properly [ i.e. it loses power ], the instances won't start again after the host is started, even withj the manual start command.

LXD returns as exception:

```
Failed to start device "nvidia-gpu": Failed to create device "/var/snap/lxd/common/lxd/devices/oel-ogrp623/cdi.unix.nvidia--gpu.dev-nvidia0" for "/dev/nvidia0": file exists
```

For clarity, devices config is:
```
devices:
  nvidia-gpu:
    gputype: physical
    id: [nvidia.com/gpu=all](http://nvidia.com/gpu=all)
    type: gpu
```

Note that checking the lxd devices dir, the following files are still present:
```
-rw-r--r-- 1 root    root           0 Jan 22 17:43 cdi.disk.nvidia--gpu.etc-vulkan-icd.d-nvidia_icd.json
-rw-r--r-- 1 root    root           0 Jan 22 17:43 cdi.disk.nvidia--gpu.etc-vulkan-implicit_layer.d-nvidia_layers.json
-rw-r--r-- 1 root    root           0 Jan 22 17:43 cdi.disk.nvidia--gpu.usr-bin-nvidia--smi
-rw-r--r-- 1 root    root           0 Jan 22 17:43 cdi.disk.nvidia--gpu.usr-lib-x86_64--linux--gnu-libEGL_nvidia.so.550.127.05
-rw-r--r-- 1 root    root           0 Jan 22 17:43 cdi.disk.nvidia--gpu.usr-lib-x86_64--linux--gnu-libGLESv1_CM_nvidia.so.550.127.05
-rw-r--r-- 1 root    root           0 Jan 22 17:43 cdi.disk.nvidia--gpu.usr-lib-x86_64--linux--gnu-libGLESv2_nvidia.so.550.127.05
-rw-r--r-- 1 root    root           0 Jan 22 17:43 cdi.disk.nvidia--gpu.usr-lib-x86_64--linux--gnu-libGLX_nvidia.so.550.127.05
-rw-r--r-- 1 root    root           0 Jan 22 17:43 cdi.disk.nvidia--gpu.usr-lib-x86_64--linux--gnu-libcuda.so.550.127.05
[...]

PR #16600, opened 2025-09-24
lxd/device/gpu_physical: fix stale CDI-related files cleanup logic

mihalicyn:
We need to take our pathname encoding into account when looking for a files to clean up.

Fixes: #14843


mihalicyn:
Tested on Ubuntu Core with CDI mode GPU passthrough.

tomponline (approved):
Thanks!

PR #16336, opened 2025-08-27
[WIP] Multi-volume snapshots

kadinsayani:
Depends on https://github.com/canonical/lxd/pull/16334.
Based on https://github.com/canonical/lxd/pull/15349.

This pull request introduces support for multi-volume snapshots for instances and their attached volumes, ensuring crash consistency across volumes.

I've gone through each commit and optimized code where possible. There are two `TODO` comments for making loops concurrent; I don't think using concurrency for snapshotting and restoring volumes will help much in those cases since an instance usually does not have many attached volumes and snapshot/restore operations are fairly quick for most storage drivers.

Outstanding todos:
- [ ] expand godoc comments
- [ ] docs
- [ ] lxd-ci tests

PR #16603, opened 2025-09-24
Return generated UUID from (*lxdBackend).CreateCustomVolumeSnapshot()

kadinsayani:
Required for #16336.

When creating a snapshot of an attached custom storage volume, we need a way to map attached volumes to their snapshots. `volatile.attached_volumes` will be a JSON-serialized map of attached volume UUIDs to the UUIDs of their corresponding snapshots. We can use this key to identify attached volume snapshots during restore and deletion.

### DIFF
--- a/lxd/storage/pool_interface.go
+++ b/lxd/storage/pool_interface.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/google/uuid"
+
 	"github.com/canonical/lxd/lxd/backup"
 	backupConfig "github.com/canonical/lxd/lxd/backup/config"
 	deviceConfig "github.com/canonical/lxd/lxd/device/config"
@@ -133,7 +135,7 @@ type Pool interface {
 	CreateCustomVolumeFromTarball(projectName string, volName string, srcData *os.File, op *operations.Operation) error
 
 	// Custom volume snapshots.
-	CreateCustomVolumeSnapshot(projectName string, volName string, newSnapshotName string, newDescription string, newExpiryDate *time.Time, op *operations.Operation) error
+	CreateCustomVolumeSnapshot(projectName string, volName string, newSnapshotName string, newDescription string, newExpiryDate *time.Time, op *operations.Operation) (*uuid.UUID, error)
 	RenameCustomVolumeSnapshot(projectName string, volName string, newSnapshotName string, op *operations.Operation) error
 	DeleteCustomVolumeSnapshot(projectName string, volName string, op *operations.Operation) error
 	UpdateCustomVolumeSnapshot(projectName string, volName string, newDesc string, newConfig map[string]string, newExpiryDate time.Time, op *operations.Operation) error
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -6719,17 +6719,19 @@ func (b *lxdBackend) ImportCustomVolume(projectName string, poolVol *backupConfi
 }
 
 // CreateCustomVolumeSnapshot creates a snapshot of a custom volume.
-func (b *lxdBackend) CreateCustomVolumeSnapshot(projectName, volName string, newSnapshotName string, newDescription string, newExpiryDate *time.Time, op *operations.Operation) error {
+// A new UUID is generated for the snapshot and returned upon success.
+// The UUID is used to fill "volatile.attached_volumes" for multi-volume snapshot and restore functionality.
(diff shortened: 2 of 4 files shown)


lxd/storage/pool_interface.go
```diff
@@ -133,7 +133,7 @@ type Pool interface {
 	CreateCustomVolumeFromTarball(projectName string, volName string, srcData *os.File, op *operations.Operation) error
 
 	// Custom volume snapshots.
-	CreateCustomVolumeSnapshot(projectName string, volName string, newSnapshotName string, newDescription string, newExpiryDate *time.Time, op *operations.Operation) error
+	CreateCustomVolumeSnapshot(projectName string, volName string, newSnapshotName string, newDescription string, newExpiryDate *time.Time, op *operations.Operation) (string, error)
```
tomponline: could we return a uuid.UUID type for clarity and convenience?

tomponline: you're already creating one internally anyway

kadinsayani: Yeah sounds good!

kadinsayani:
@tomponline ready to go.

tomponline (commented):
Suggest returning a pointer to uuid so we can return nil on error.

@roosterfish @MusicDin @mihalicyn are there any concern re DR discovery or with remote storage drivers for generating volume snap shot uuids?

Will this interfere with naming snapshots on the appliance that derive their name from the uuid?

@kadinsayani do these volume snapshots have a volatile.uuid of the parent today already? Are we changing that here?

kadinsayani:
> @kadinsayani do these volume snapshots have a volatile.uuid of the parent today already? Are we changing that here?

The `Volume` struct has a field `parentUUID` which is set to the parent volume's `volatile.uuid`. We capture the parent volume's UUID, set it, and then move on to generating a new UUID for the snapshot volume:

```Go
	parentUUID := parentVol.Config["volatile.uuid"]
	if parentUUID == "" {
		return nil, fmt.Errorf(`Volume %q is missing the required "volatile.uuid" setting`, parentVol.Name)
	}

	// Get the volume name on storage.
	volStorageName := project.StorageVolume(projectName, fullSnapshotName)
	vol := b.GetNewVolume(drivers.VolumeTypeCustom, contentType, volStorageName, parentVol.Config)

	// Set the parent volume's UUID.
	vol.SetParentUUID(parentUUID)

	// Generate snapshot volume UUID.
	snapshotUUID := uuid.New()
	vol.Config()["volatile.uuid"] = snapshotUUID.String()
```

tomponline:
> > @kadinsayani do these volume snapshots have a volatile.uuid of the parent today already? Are we changing that here?
> 
> The `Volume` struct has a field `parentUUID` which is set to the parent volume's `volatile.uuid`. We capture the parent volume's UUID, set it, and then move on to generating a new UUID for the snapshot volume:
> 
> ```go
> 	parentUUID := parentVol.Config["volatile.uuid"]
> 	if parentUUID == "" {
> 		return nil, fmt.Errorf(`Volume %q is missing the required "volatile.uuid" setting`, parentVol.Name)
> 	}
> 
> 	// Get the volume name on storage.
> 	volStorageName := project.StorageVolume(projectName, fullSnapshotName)
> 	vol := b.GetNewVolume(drivers.VolumeTypeCustom, contentType, volStorageName, parentVol.Config)
> 
> 	// Set the parent volume's UUID.
> 	vol.SetParentUUID(parentUUID)
> 
> 	// Generate snapshot volume UUID.
> 	snapshotUUID := uuid.New()
> 	vol.Config()["volatile.uuid"] = snapshotUUID.String()
> ```

I meant today do snapshots inherit the parents uuid?

kadinsayani:
It looks like `GetNewVolume()` already generates a new UUID, so we can skip repeating that step in `CreateCustomVolumeSnapshot()` and just parse `volatile.uuid` and return it.

```Go
// GetNewVolume returns a drivers.Volume that doesn't yet exist in the database.
// It contains copies of the supplied volume config, including a new UUID, and the pools config.
// Use the returned drivers.Volume as the base for actions performed on the new volume.
func (b *lxdBackend) GetNewVolume(volType drivers.VolumeType, contentType drivers.ContentType, volName string, volConfig map[string]string) drivers.Volume {
	newVol := b.GetVolume(volType, contentType, volName, volConfig)

	// Set a new UUID.
	newVol.Config()["volatile.uuid"] = uuid.New().String()
	return newVol
}
```

kadinsayani:
> I meant today do snapshots inherit the parents uuid?

No, we just save it before calling `GetNewVolume()` which generates a new UUID.

tomponline (approved):
Thanks!

PR #16605, opened 2025-09-24
Snapshot volatile config validation

kadinsayani:
Required for #16336.


kadinsayani:
> Hrm, this feels like a hack to me.
> 
> Shouldn't we instead change the ValidConfig to take a `isSnapshot` bool so that the config can be validated based on whether its a snapshot or not?

Yes, that would be a better approach. I'll update this.

kadinsayani:
Closing as I don't think this is actually required. We can just delete `volatile.attached_volumes` from the config during multi-volume restores.

PR #16609, opened 2025-09-24
lxd: Prevent instance migration to evacuted member

nmezhenskyi:
Fixes https://github.com/canonical/lxd/issues/16468.

## Changes:
- Extend validation inside of the `instancePostClusteringMigrate()` function to prevent instance migration if the target member is evacuated.

### DIFF
--- a/test/suites/clustering.sh
+++ b/test/suites/clustering.sh
@@ -3562,6 +3562,9 @@ test_clustering_evacuation() {
   c4_location="$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c L c4)"
   c5_location="$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c L c5)"
 
+  echo "==> Check that instance migration to an evacuated node is not allowed."
+  [[ "$(LXD_DIR="${LXD_TWO_DIR}" lxc move c5 --target=node1 2>&1)" == *"Error: The destination cluster member is evacuated"* ]]
+
   # Restore first node with "skip" mode.
   # "skip" mode restores cluster member status without starting instances or migrating back evacuated instances.
   LXD_DIR="${LXD_TWO_DIR}" lxc cluster restore node1 --action=skip --force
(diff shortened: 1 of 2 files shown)


test/suites/clustering.sh
```diff
@@ -3562,6 +3562,9 @@ test_clustering_evacuation() {
   c4_location="$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c L c4)"
   c5_location="$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c L c5)"
 
+  echo "==> Check that instance migration to an evacuated node is not allowed."
+  ! LXD_DIR="${LXD_TWO_DIR}" lxc move c6 --target=node1 || false
```
simondeziel: I'd pick another `cX` because `c6` is running so wouldn't be allowed to move anyway. `c5` seems like a good candidate as it's `STOPPED` and NOT on `node1`.

There's a lot going on with those 6 containers so maybe that's a case where we should also inspect the error we get to be the right one.

simondeziel (changes requested):

test/suites/clustering.sh
simondeziel: FYI, I fished out the previous CI run logs and even with `c6` being running, it error'ed out due to `node1` being evacuated. That said, I still think checking the error is a safe choice.

nmezhenskyi: Updated

test/suites/clustering.sh
```diff
@@ -3562,6 +3562,9 @@ test_clustering_evacuation() {
   c4_location="$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c L c4)"
   c5_location="$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c L c5)"
 
+  echo "==> Check that instance migration to an evacuated node is not allowed."
+  [[ "$(LXD_DIR="${LXD_TWO_DIR}" lxc move c5 --target=node1 2>&1)" == *"Error: The destination cluster member is evacuated"* ]]
```
nmezhenskyi, line 3566: I didn't use exact matching here because the command produced some extra output:
```sh
+ '[' '++ timeout --foreground 120 /home/runner/go/bin/lxc move c5 --target=node1 --verbose
Error: The destination cluster member is evacuated' = 'Error: The destination cluster member is evacuated' ']' 
```

tomponline (approved):
Thanks!

PR #16610, opened 2025-09-24
Test improvements

simondeziel:


test/suites/config.sh
```diff
@@ -277,12 +281,11 @@ test_config_edit() {
     [ "$(lxc config get foo --property description)" = "bar" ]
 
     # Check instance name is included in edit screen.
-    cmd=$(unset -f lxc; command -v lxc)
-    output=$(EDITOR="cat" timeout --foreground 120 "${cmd}" config edit foo)
-    echo "${output}" | grep "name: foo"
+    output=$(EDITOR="cat" ${unbuffer} timeout --foreground 120 "${_LXC}" config edit foo)
```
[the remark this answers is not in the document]
simondeziel: Nope, quoting that variable causes problem if the command is not there (which is OK when not needed) but there is TTY connected. This is the case for local runners (!= GH ones).

test/suites/auth.sh
```diff
@@ -873,7 +873,7 @@ auth_project_features() {
   remote="${1}"
 
   # test-group must have no permissions to start the test.
-  [ "$(lxc query /1.0/auth/groups/test-group | jq '.permissions | length')" -eq 0 ]
+  [ "$(lxc query /1.0/auth/groups/test-group | jq -r '.permissions')" = "[]" ]
```
markylaing: I've been enjoying the use of `-e` with `jq` for assertions:

```suggestion
lxc query /1.0/auth/groups/test-group | jq -e '(.permissions | length) == 0'
```

test/suites/config.sh
```diff
@@ -404,38 +407,39 @@ test_container_metadata() {
...
     # Check instance name is included in edit screen.
-    cmd=$(unset -f lxc; command -v lxc)
-    output=$(EDITOR="cat" timeout --foreground 120 "${cmd}" config edit foo/snap0)
-    echo "${output}" | grep "name: snap0"
+    output=$(EDITOR="cat" ${unbuffer} timeout --foreground 120 "${_LXC}" config edit foo/snap0)
+    echo "${output}" | grep -xF "name: snap0"
 
     # Check expanded config isn't included in edit screen.
-    ! echo "${output}"  | grep "expanded" || false
+    ! echo "${output}" | grep -F "expanded" || false
```
markylaing, line 442: For me it's preferable to never try to grep for something that doesn't exist. Can we use `yq` here to make assertions like we do with `jq`? Not sure if it supports it.

test/suites/devlxd.sh
```diff
@@ -241,7 +240,7 @@ test_devlxd_volume_management() {
     lxc exec --project "${project}" "${inst}" -- devlxd-client
 
     # Ensure supported storage drivers are included in /1.0 only when volume management security flag is enabled.
-    lxc exec "${inst}" --project "${project}" -- devlxd-client get-state | jq -e '.supported_storage_drivers | length == 0'
+    [ "$(lxc exec "${inst}" --project "${project}" -- devlxd-client get-state | jq -e '.supported_storage_drivers')" = "[]" ]
```
markylaing: Why is this better than using `jq` to perform the assertion?

markylaing: Also now I think the `-e` flag for `jq` has no effect, since it's running in a command substitution inside a test

test/suites/devlxd.sh
```diff
@@ -285,7 +284,7 @@ test_devlxd_volume_management() {
     lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage get "${pool}"
 
     # Get storage volumes (ok - custom volumes requested).
-    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage volumes "${pool}" | jq 'length == 0'
+    [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage volumes "${pool}")" = "[]" ]
```
markylaing:
```suggestion
lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage volumes "${pool}" | jq -e 'length == 0'
```

test/suites/devlxd.sh
```diff
@@ -379,7 +378,7 @@ EOF
 
     etag=$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client instance get-etag "${inst}")
     lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client instance update "${inst}" "${detachReq}" "${etag}"
-    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client instance get "${inst}" | jq '.devices' | jq -e 'length == 0'
+    [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client instance get "${inst}" | jq -e '.devices')" = "{}" ]
```
markylaing:
```suggestion
lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client instance get "${inst}" | jq -e '.devices | length == 0'
```

test/suites/devlxd.sh
```diff
@@ -397,7 +396,7 @@ EOF
 
     etag=$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client instance get-etag "${inst2}")
     lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client instance update "${inst2}" "${detachReq}" "${etag}"
-    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client instance get "${inst2}" | jq '.devices' | jq -e 'length == 0'
+    [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client instance get "${inst2}" | jq -e '.devices')" = "{}" ]
```
markylaing:
```suggestion
lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client instance get "${inst2}" | jq -e '.devices | length == 0'
```

test/suites/devlxd.sh
```diff
@@ -416,7 +415,7 @@ EOF
     lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage delete-volume "${pool}" custom vol-02
 
     # Ensure storage volumes are deleted.
-    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage volumes "${pool}" | jq 'length == 0'
+    [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage volumes "${pool}")" = "[]" ]
```
markylaing:
```suggestion
lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage volumes "${pool}" | jq -e 'length == 0'
```

simondeziel:
> I like using `jq` for assertions,

Fair, will apply those good suggestions, thanks!

> though we do need to have the `-e` flag consistently applied

Yep, I plan on adding a linter to enforce it.

> and we should avoid multiple invocations when one will suffice.

True that!

test/suites/config.sh
simondeziel: If you don't mind, I'll leave that one for another PR.

simondeziel:
@markylaing thanks a lot for you review feedback. I took your suggestion about `jq -e` a bit further by replacing some `length` checks with empty type assertions.

test/includes/check.sh
```diff
@@ -190,7 +190,9 @@ check_log_order() {
 }
 
 # runsMinimumKernel: check if the running kernel is at least the minimum version.
-runsMinimumKernel() (
+runsMinimumKernel() {
+    local min_version min_major min_minor
+    local running_version running_major running_minor
```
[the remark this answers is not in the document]
simondeziel: I don't see what you mean.

markylaing (approved):
LGTM thanks :)

test/suites/auth.sh
tomponline, line 1: We should use long form flags for clarity when reading and not typing commands. A note for the future please.

test/suites/devlxd.sh
```diff
@@ -57,7 +57,7 @@ test_devlxd() {
   lxc auth identity delete devlxd/foo
 
   # No secret remains in the database after the identity was deleted
-  [ "$(lxd sql global --format csv 'SELECT COUNT(*) FROM secrets WHERE entity_id = (SELECT id FROM identities WHERE name = "foo")')" = 0 ]
```
tomponline, line 60: I preferred the former, it was clearer than comparing empty string to me.
What was the problem with it?

simondeziel: There was no problem but to me it read as if there was something else in that table needing to be excluded while in fact, at that point, the table is supposed to be empty.

I've been looking into it after noticing this script would work when ran twice in a row. Turns out the issue was not the DB but that's what "brought me here" :)

test/suites/auth.sh
simondeziel: If you are suggesting long flags to `jq`, then I think it will make a future lint script harder cause I'd like to catch all of the use of `jq` that are lacking `-e`. Now if it also needs to consider `--exit-status` this will be quite a bit harder I'm afraid.

test/suites/devlxd.sh
tomponline: ``SELECT COUNT(*) FROM secret` = 0` was clearer to me then

test/suites/auth.sh
tomponline: why? we should always use --exit-status and not -e?

test/suites/devlxd.sh
simondeziel: I know nothing about SQL so maybe it's a common idiom? I dropped the count as I felt it was an extra indirection when checking for emptiness.

I'll put it back.

test/suites/auth.sh
simondeziel: I guess it would work to insist on the long flag so the linter would error out when just using `jq -e`.

A quick count showed that 392 call sites will need to be changed to apply please that linter :/

test/suites/devlxd.sh
tomponline: to me thats clearer that we're saying "no rows", rather than "no rows or a row with an empty value"

simondeziel: Ah, I didn't even noticed that the check for emptiness is what forced the `|| echo fail` bit in. I agree that the count reads better ;)

test/suites/auth.sh
tomponline: yeah that would be better, we've spoken before about avoiding short form flags in tests where possible.

tomponline: doesnt have to be 392 commits ;)

test/suites/devlxd.sh
simondeziel: >  rather than "no rows or a row with an empty value"

Oh, that's a very important point, thanks for bringing it!

tomponline:
@simondeziel tests are failing intermittently now you've reduced the token expiry sleep time:

https://github.com/canonical/lxd/actions/runs/18037519953

```
2025-09-26T12:50:32.9754095Z ++ lxc auth identity token issue devlxd/foo --quiet --expiry 1S
2025-09-26T12:50:32.9757036Z ++ timeout --foreground 120 /home/runner/go/bin/lxc auth identity token issue devlxd/foo --quiet --expiry 1S --verbose
2025-09-26T12:50:32.9915268Z + devlxd_token4=***
2025-09-26T12:50:32.9919804Z + lxc exec --env DEVLXD_BEARER_TOKEN=*** devlxd -- devlxd-client get-state
2025-09-26T12:50:32.9920501Z + jq -e '.auth == "trusted"'
2025-09-26T12:50:32.9929737Z + timeout --foreground 120 /home/runner/go/bin/lxc exec --env DEVLXD_BEARER_TOKEN=*** devlxd --verbose --force-local -- devlxd-client get-state
2025-09-26T12:50:33.0219949Z parse error: Invalid numeric literal at line 1, column 7
```

I fixed this previously here https://github.com/canonical/lxd/pull/16275 but seems to have been undone by your latest PR.

simondeziel:
> @simondeziel tests are failing intermittently now you've reduced the token expiry sleep time:
> 
> https://github.com/canonical/lxd/actions/runs/18037519953
> 
> ```
> 2025-09-26T12:50:32.9754095Z ++ lxc auth identity token issue devlxd/foo --quiet --expiry 1S
> 2025-09-26T12:50:32.9757036Z ++ timeout --foreground 120 /home/runner/go/bin/lxc auth identity token issue devlxd/foo --quiet --expiry 1S --verbose
> 2025-09-26T12:50:32.9915268Z + devlxd_token4=***
> 2025-09-26T12:50:32.9919804Z + lxc exec --env DEVLXD_BEARER_TOKEN=*** devlxd -- devlxd-client get-state
> 2025-09-26T12:50:32.9920501Z + jq -e '.auth == "trusted"'
> 2025-09-26T12:50:32.9929737Z + timeout --foreground 120 /home/runner/go/bin/lxc exec --env DEVLXD_BEARER_TOKEN=*** devlxd --verbose --force-local -- devlxd-client get-state
> 2025-09-26T12:50:33.0219949Z parse error: Invalid numeric literal at line 1, column 7
> ```
> 
> I fixed this previously here #16275 but seems to have been undone by your latest PR.

Ah, thanks I wanted to check at that error also pop for me once. Will reintroduce the fix with a comment.

PR #16612, opened 2025-09-25
client/devlxd: Add UseBearerToken function 

MusicDin:
Add `UseBearerToken` function for devLXD client to allow changing token without creating a new client. This is helpful for CSI - when we detect the new token we can create a copy of the client with new bearer token (while retaining http client, context, etc) and start using it from that moment onward, allowing us to keep the client for the entire lifecycle of the CSI container.

### DIFF
--- a/client/devlxd.go
+++ b/client/devlxd.go
@@ -95,6 +95,13 @@ func (r *ProtocolDevLXD) UseTarget(name string) DevLXDServer {
 	}
 }
 
+// UseBearerToken returns a client that will use the provided bearer token for authentication.
+func (r *ProtocolDevLXD) UseBearerToken(bearerToken string) DevLXDServer {
+	server := *r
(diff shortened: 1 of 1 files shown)


tomponline (commented):
Should we call this UseAuthBearerToken to indicate its for auth?

tomponline (dismissed):
I see the field itself is called `BearerToken` in the client `ConnectionArgs` so lets keep it consistent with that.

MusicDin:
Ops, just changed the name. Will revert

tomponline:
> Ops, just changed the name. Will revert

Sorry

MusicDin:
Fixed

MusicDin:
Done, updated `UseTarget` as well

tomponline (approved):
Oh you did already, GH diff view failed me, thanks!

PR #16622, opened 2025-09-26
Improve test speed and correctness

simondeziel:
Improves `jq` usage correctness by replacing multiple `jq -r` with `jq --exit-status` to reflect the assertion result in the test's outcome.

Speedup `container_devices_nic_bridged` test from ~74s to ~48s.


tomponline (approved):
Thanks!

PR #16623, opened 2025-09-26
Prune cached images during project delete

kadinsayani:
Closes https://github.com/canonical/lxd/issues/13507.


lxd/api_project.go
```diff
@@ -1361,12 +1361,45 @@ func projectStateGet(d *Daemon, r *http.Request) response.Response {
...
+			ref, err := entity.ReferenceFromURL(*parsedURL)
+			if err != nil {
+				return false, err
+			}
+
+			// Check for cached images.
+			if ref.EntityType == entity.TypeImage {
+				fingerprint := ref.Name()
+
+				image, err := dbCluster.GetImage(ctx, tx.Tx(), project.Name, fingerprint)
```
tomponline: this will run a query per-image, can we use `dbCluster.GetImages` to load all the cached imagses in a project in a single query and then use that result to identify urls to remove from this list?

Also, to save loading the cached images again outside of this function, why dont we call `dbCluster.GetImages` in the call site, and then pass in a list of "ignore urls" to this function?

kadinsayani: Sounds good, I've made those changes :)

lxd/api_project.go
```diff
@@ -970,7 +970,7 @@ func projectPost(d *Daemon, r *http.Request) response.Response {
 				return fmt.Errorf("Failed loading project %q: %w", name, err)
 			}
 
-			empty, err := projectIsEmpty(ctx, project, tx)
+			empty, err := projectIsEmpty(ctx, project, tx, nil)
```
tomponline, line 973: not a blocker, but remember there are variadic arguments which means the caller doesnt have to provide the argument at all if they dont have anything to provide.

https://golangdocs.com/variadic-functions-in-golang

lxd/api_project.go
```diff
@@ -1236,9 +1236,21 @@ func projectDelete(d *Daemon, r *http.Request) response.Response {
 			return fmt.Errorf("Failed loading project %q: %w", name, err)
 		}
 
+		cached := true
```
tomponline, line 1239: does this var need this level of scope or can it be moved inside the `if !force{` stanza?

tomponline (approved):
Thanks!

lxd/api_project.go
kadinsayani: Ah right, thanks for the reminder :)

lxd/api_project.go
kadinsayani: We can move it in.

kadinsayani:
> Thanks!

Np! Thanks for the review! 

Quick follow-up PR here: https://github.com/canonical/lxd/pull/16654.

Issue #13374, opened 2024-04-23
Remove cluster healing feature


iatrou:
https://documentation.ubuntu.com/lxd/en/latest/api-extensions/#cluster-healing

Please clarify what the "threshold" unit is. Seconds? Something else?
Furthermore, the default value is 0, disabled.
Please provide a recommended value when it's enabled. 
If there are values that are known to be problematic (e.g. causing false positives) please call it out.

---
Document: api-extensions.md

- [x] Clarify units and add caution (https://github.com/canonical/lxd/pull/16624)
- [ ] Add cluster healing event and update docs

ru-fu:
> https://documentation.ubuntu.com/lxd/en/latest/api-extensions/#cluster-healing
> 
> Please clarify what the "threshold" unit is. Seconds? Something else? 

It's seconds - see the documentation for the configuration option (which is linked):
https://documentation.ubuntu.com/lxd/en/latest/server/#server-cluster:cluster.healing_threshold

> Furthermore, the default value is 0, disabled. Please provide a recommended value when it's enabled. If there are values that are known to be problematic (e.g. causing false positives) please call it out.

Not sure if we can give recommendations, since it depends on the individual setup, size of the cluster, network speed ...
Any ideas @tomponline ? Or pointers to who can give more input?

tomponline:
Thanks @ru-fu 

@iatrou @ru-fu is correct that the `cluster.healing_threshold` is environment & workload specific.

I would not suggest setting it too low as any short live issues in the network, or short lived high load on a single host that may cause it to stop responding to heartbeats in a timely manner could trigger the cluster healing for that mechanism.

So my recommendation is to set it as high as possible that the organisation's availability targets are met whilst avoiding the chances of short lived "events" triggering an unwanted forced evacuation of a cluster member.

I would also like to draw your attention to some known issues with cluster healing and evacuation:

https://github.com/canonical/lxd/issues/13083 - this is also a risk if the cluster healing threshold is set too low, such that any manual maintenance windows that involve rebooting a cluster member cleaning can result in the member being considered "offline" and being evacuated, meaning that its workloads are started up elsewhere and when the member is back up again, it will have to be manually restored to service using `lxc cluster restore` which will then move the workloads back (with disruption to the workloads).

[...]

PR #16624, opened 2025-09-27
doc: partial port of cluster healing section (from Incus)

minaelee:
This ports and adapts information about cluster healing (automatic evacuation of workloads from clusters when a cluster member goes offline) from the Incus documentation, with edits for LXD terminology and context.

Inspired-by: [fc1d7d0191 (\"doc/clustering: Better document healing\")](https://github.com/lxc/incus/pull/1075/files#diff-dd7677da7f945b89e7e144c9848ca0603c332793beac980926c06021037e3620R88)

Details:
Specifically, I adapted two parts of the cluster healing documentation from Incus in the above-referenced commit:
- Incus added a `note` admonition that "This automatic cluster healing only applies to instances on shared storage and which don't use any local devices." I lightly edited this and placed it inline.
- Incus added a `warning` admonition about the risk of automatic evacuation/cluster healing when cluster members are  incorrectly deemed offline. I added a similar `warning`, including advice [based on this comment from @tomponline](https://github.com/canonical/lxd/issues/13374#issuecomment-3339051526) that the cluster healing threshold should be set high enough to avoid cluster members being evacuated during short-lived network issues or high load.
  - Note that I did not include Incus's [advice to use third party tooling](https://github.com/lxc/incus/blob/main/doc/howto/cluster_manage.md?plain=1#L96) to watch for the `cluster-member-healed` event because this [event does not exist yet in LXD](https://github.com/canonical/lxd/blob/main/shared/api/event_lifecycle.go#L8). 

I also added clarification that "automatic evacuation" and "cluster healing" are referring to the same concept in this case, and that the cluster healing threshold value is set in seconds (which was the original request for clarification in issue https://github.com/canonical/lxd/issues/13374). + a couple more relevant links.

### DIFF
--- a/doc/howto/cluster_manage.md
+++ b/doc/howto/cluster_manage.md
@@ -112,12 +112,25 @@ If an instance is not suitable for live migration, it will be shut down cleanly 
 Any instance that you plan to live-migrate must have its {config:option}`instance-migration:migration.stateful` configuration option set to `true`. Be aware that this option can only be set while the instance is stopped. Thus, for any instance to have the ability to be live-migrated in the future, this option must be set to `true` ahead of time.
 ```
 
+(cluster-healing)=
 (cluster-automatic-evacuation)=
-### Automatic evacuation
+## Cluster healing
 
-If you set the {config:option}`server-cluster:cluster.healing_threshold` configuration to a non-zero value, instances are automatically evacuated if a cluster member goes offline.
+To enable cluster healing, set the {config:option}`server-cluster:cluster.healing_threshold` configuration to a non-zero value (in seconds). If a cluster member is offline for longer than this threshold, LXD automatically sets its state to "evacuated" and starts its instances on another member. This behavior only applies to instances that use shared storage storage and have no local devices attached.
 
-When the evacuated server is available again, you must manually restore it.
+Syntax:
+
+```bash
+lxc cluster set cluster.healing_threshold <value in seconds>
+```
+
+When the healed cluster member is available again, you must manually {ref}`restore <cluster-restore>` it to remove its "evacuated" state and return instances to it.
+
+```{warning}
+Enabling the cluster healing threshold carries the risk that LXD might incorrectly judge a cluster member as offline while it is still running workloads. Short-lived network issues or temporary high load might cause a cluster member to briefly stop responding to heartbeat or ICMP packets. If a healing threshold is set, LXD might then start that member's instances on another cluster member even though they're still active on the original. Since cluster members share the same storage, this can lead to data corruption.
+
+To reduce the chance of false healing events, set {config:option}`server-cluster:cluster.healing_threshold` as high as possible within your availability targets.
+```
 
 (cluster-manage-delete-members)=
 ## Delete cluster members


doc/howto/cluster_manage.md
```diff
@@ -115,9 +115,15 @@ Any instance that you plan to live-migrate must have its {config:option}`instanc
 (cluster-automatic-evacuation)=
 ### Automatic evacuation
 
-If you set the {config:option}`server-cluster:cluster.healing_threshold` configuration to a non-zero value, instances are automatically evacuated if a cluster member goes offline.
+If you set the {config:option}`server-cluster:cluster.healing_threshold` configuration to a non-zero value (in seconds), instances are automatically evacuated if a cluster member goes offline. This behavior, also known as cluster healing, only applies to instances that share storage and do not use any local devices.
```
tomponline: @minaelee shall we lose the term "evacuated" here, because that, to me, is a different type of operation.
One `evacuates` a server **before** losing access to it (such as a before maintenance).
It can move instances irrespective of whether they are on shared storage or not.

Whereas cluster healing is only for **after** you lose access to a cluster member unexpectedly and by that point the only instances that can be moved to a different cluster member are those on shared storage.

tomponline: Shall we use the term "moved" instead so as not to conflate meanings?

doc/howto/cluster_manage.md
```diff
@@ -115,9 +115,15 @@ Any instance that you plan to live-migrate must have its {config:option}`instanc
...
 
-When the evacuated server is available again, you must manually restore it.
+When the evacuated cluster member is available again, you must manually {ref}`restore <cluster-restore>` it.
+
+```{warning}
+Enabling automatic evacuation (cluster healing) carries risks if LXD incorrectly deems a cluster member as offline while it is still running workloads. This can happen due to short-lived network issues or high load on a single cluster member, either of which might cause it to briefly stop responding to heartbeat or ICMP packets. This might result in workloads being evacuated to another cluster member while still continuing to run on the original one, which can cause data corruption since they share the same storage.
```
tomponline: shall we say "moved" rather than "evacuated"?

doc/howto/cluster_manage.md
minaelee: We're using "migrated" or "live migrated" everywhere else to describe moving instances between servers (when not using "evacuated"). However, for this operation specifically, it looks like it's tied to evacuation in the code: [func internalClusterHeal](https://github.com/canonical/lxd/blob/main/lxd/api_cluster.go#L3175) returns [evacuateClusterMember](https://github.com/canonical/lxd/blob/main/lxd/api_cluster.go#L3239). Meaning it's set to an evacuated state and can be restored using `lxc cluster restore` after it returns online. This makes the case that it can be considered evacuation rather than migration. Wdyt?

tomponline: Hrm, this is tricky from a terminology basis.

The source member is unreachable, so the instance's can't be live migrated.

What happens is that the instance's location is updated in the DB and then the instance is simply started on the new cluster member. Nothing from the source member is copied (because its unreachable).

So whether that is a "migration" or a "move" I don't know.

I see your point about it setting the member's status to evacuated so that manual action has to be taken before it is considered for future placement when/if it comes back online.

But I'm concerned about conflating the verb "evacuating" with "cluster healing" - they are very different, even if the end result is similar in some sense.

Could we just say:

> Instances that are running on shared storage are started on another cluster member and the cluster member's state is set to evacuated

?

minaelee: That sounds completely reasonable. I will update.

tomponline (approved):
Thanks!

PR #16625, opened 2025-09-29
User agent features validation

markylaing:
The parser added in #16593 relies on the storage drivers having whitespace separating the driver name and version and features not containing whitespace. This PR enforces that any features set in `version.UserAgentFeatures` must contain whitespace. I didn't return an error if this happens and instead logged a warning, since it isn't critical for operation of LXD.


tomponline (approved):
ta

PR #16639, opened 2025-09-29
Auth: pass a single remote address to cookie jar

### LINKED ISSUE #16634: Fix cookiejar public suffix list to only the URL of a given remote
We're using `golang.org/x/net/publicsuffix` for the client cookie jar but we could restrict this to only save cookies from the URL of the remote, since cookies should not be required for other URLs the client calls out to.

See https://github.com/canonical/lxd/pull/16476#discussion_r2387959547_

dhvogel:
This change fixes #16634 by shrinking the PublicSuffixList of a CookieJar from the recommended publicsuffixlist to just the LXD remote that is passed to the client. (Context: https://github.com/canonical/lxd/pull/16476#discussion_r2387886168).

Dan

### DIFF
--- a/lxc/config/remote.go
+++ b/lxc/config/remote.go
@@ -359,7 +359,7 @@ func (c *Config) getConnectionArgs(name string) (*lxd.ConnectionArgs, error) {
 				}
 			}
 
-			jar, err := cookiejar.Open(c.CookiesPath(name))
+			jar, err := cookiejar.Open(c.CookiesPath(name), remote.Addr)
 			if err != nil {
 				return nil, err
 			}
--- a/lxc/cookiejar/jar.go
+++ b/lxc/cookiejar/jar.go
@@ -6,24 +6,23 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/cookiejar"
+	"net/url"
 	"os"
 	"reflect"
 	"unsafe"
-
-	"golang.org/x/net/publicsuffix"
 
 	"github.com/canonical/lxd/shared/api"
 )
 
 // Open opens a new jar at the given location, reading contents from disk if present.
-func Open(filepath string) (*Jar, error) {
-	jar, err := newJar(filepath)
+func Open(filepath string, host string) (*Jar, error) {
+	jar, err := newJar(filepath, host)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to open jar: %w", err)
 	}
 
-	err = jar.openJar()
-	if err != nil {
+	if err := jar.openJar(); 
+	err != nil {
 		return nil, fmt.Errorf("Failed to open jar: %w", err)
 	}
 
@@ -31,14 +30,18 @@ func Open(filepath string) (*Jar, error) {
 }
 
 // newJar creates a new [Jar] but does not attempt to read existing contents of the jar.
-func newJar(filepath string) (*Jar, error) {
-	// Use the default public suffix list.
-	jar, err := cookiejar.New(&cookiejar.Options{PublicSuffixList: publicsuffix.List})
+func newJar(filepath string, remote string) (*Jar, error) {
+	jar, err := cookiejar.New(&cookiejar.Options{})
 	if err != nil {
 		return nil, fmt.Errorf("Failed to instantiate a cookie jar: %w", err)
 	}
 
-	return &Jar{Jar: jar, filepath: filepath}, nil
+	remoteURL, err := url.Parse(remote)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to parse remote URL: %w", err)
+	}
+
+	return &Jar{Jar: jar, filepath: filepath, remote: *remoteURL}, nil
 }
 
 // Jar implements [http.CookieJar] by embedding a [*cookiejar.Jar].
@@ -47,6 +50,7 @@ func newJar(filepath string) (*Jar, error) {
 type Jar struct {
 	filepath string
 	*cookiejar.Jar
+	remote url.URL
 }
 
 // openJar opens the jar file and obtains a read lock on it. Then reads the contents into the jar and unlocks the file.
@@ -175,4 +179,15 @@ func (j *Jar) Save() error {
 	}
 
 	return nil
+}
+
+// SetCookies implements [http.CookieJar.SetCookies]. It only accepts cookies from the remote host.
+func (j *Jar) SetCookies(u *url.URL, cookies []*http.Cookie) {
+	// Only accept cookies from the remote host.
+	host := u.Hostname()
+	allowedHost := j.remote.Hostname()
+
+	if host == allowedHost {
+		j.Jar.SetCookies(u, cookies)
+	}
 }
--- a/lxc/cookiejar/jar_test.go
+++ b/lxc/cookiejar/jar_test.go
@@ -3,6 +3,7 @@
 package cookiejar
 
 import (
+	"fmt"
 	"net/http"
 	"net/url"
 	"os"
@@ -15,7 +16,8 @@ import (
 
 type cookieJarSuite struct {
 	suite.Suite
-	j *Jar
+	host string
+	j    *Jar
 }
 
 func TestCookieJarSuite(t *testing.T) {
@@ -25,9 +27,84 @@ func TestCookieJarSuite(t *testing.T) {
 func (s *cookieJarSuite) SetupTest() {
 	tmpDir := s.T().TempDir()
 	cookieFile := filepath.Join(tmpDir, "cookies.json")
-	j, err := Open(cookieFile)
+	remote := "https://127.0.0.1:8443/"
+	j, err := Open(cookieFile, remote)
 	s.Require().NoError(err)
 	s.j = j
+}
+
+func (s *cookieJarSuite) TestSetCookies() {
+	require := s.Require()
+
+	// --- Domain example ---
+	domainHost := "example.com"
+	uRemote, err := url.Parse("https://" + domainHost)
+	require.NoError(err)
+
+	uSubdomain, err := url.Parse("https://foo." + domainHost)
+	require.NoError(err)
+
+	uOther, err := url.Parse("https://evil.com/")
+	require.NoError(err)
+
+	cookies := []*http.Cookie{
+		{Name: "foo", Value: "bar"},
+	}
+
+	// Should set cookies for remote domain
+	s.j.remote = *uRemote
+	s.j.SetCookies(uRemote, cookies)
+	got := s.j.Cookies(uRemote)
+	require.Len(got, 1)
+	require.Equal("foo", got[0].Name)
+	require.Equal("bar", got[0].Value)
+
+	// Should not set cookies for subdomain
+	s.j.SetCookies(uSubdomain, cookies)
+	gotSub := s.j.Cookies(uSubdomain)
+	require.Empty(gotSub)
+
+	// Should not set cookies for other domain
+	s.j.SetCookies(uOther, cookies)
+	gotOther := s.j.Cookies(uOther)
+	require.Empty(gotOther)
+
+	// --- IP example ---
+	ipHost := "http://127.0.0.1:8443/"
+	uIP, err := url.Parse(ipHost)
+	fmt.Printf("Parsed URL is %s\n", uIP.String())
+	require.NoError(err)
+
+	uIPOther, err := url.Parse("http://192.168.1.1/")
+	require.NoError(err)
+
+	uIPSub, err := url.Parse("http://sub.127.0.0.1:8443/")
+	require.NoError(err)
+
+	ipCookies := []*http.Cookie{
+		{Name: "ipcookie", Value: "123"},
+	}
+
+	// Remote IP URL
+	uIPRemote, err := url.Parse("http://127.0.0.1:8443/")
+	require.NoError(err)
+	s.j.remote = *uIPRemote
+
+	// Exact IP allowed
+	s.j.SetCookies(uIPRemote, ipCookies)
+	gotIP := s.j.Cookies(uIPRemote)
+	require.Len(gotIP, 1)
+	require.Equal("ipcookie", gotIP[0].Name)
+
+	// Other IP should be rejected
+	s.j.SetCookies(uIPOther, ipCookies)
+	gotIPOther := s.j.Cookies(uIPOther)
+	require.Empty(gotIPOther)
+
+	// Subdomain of IP should be rejected
+	s.j.SetCookies(uIPSub, ipCookies)
+	gotIPSub := s.j.Cookies(uIPSub)
+	require.Empty(gotIPSub)
 }
 
 func (s *cookieJarSuite) TestLocking() {
@@ -86,6 +163,16 @@ func (s *cookieJarSuite) TestJar() {
 			HttpOnly: true,
 			SameSite: http.SameSiteStrictMode,
 		},
+		{
+			Name:   "shouldnotbehere",
+			Value:  "evil",
+			Domain: "evil.com",
+		},
+		{
+			Name:   "shouldalsonotbehere",
+			Value:  "evil",
+			Domain: "1.2.3.4:5678",
+		},
 	}
 
 	// Set cookies on the suite jar.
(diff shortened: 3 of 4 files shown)


lxc/config/remote.go
```diff
@@ -14,7 +14,7 @@ import (
 
 	"github.com/zitadel/oidc/v3/pkg/oidc"
 
-	"github.com/canonical/lxd/client"
+	lxd "github.com/canonical/lxd/client"
```
tomponline: please can you revert this to what it was, thanks

dhvogel: yes

tomponline:
Also please can you ensure that each commit in this PR is the final version of single logical change - i.e dont add more commits showing the evolution of the PR. As LXD project doesn't use squash merges your PR commits will land into the main branch as-is. Thanks

dhvogel:
I've squashed the changes. 

Re-reading what I wrote, I think that this change might need some more work. Right now it builds a suffix list that is derived from the lxd server address, but that would enable the cookie jar for all connections with that same suffix. I will take another pass today to see if a restriction can be made for the actual address, not just the suffix, and to see if there is a test I can write for it.

lxc/cookiejar/jar_test.go
```diff
@@ -130,7 +133,7 @@ func (s *cookieJarSuite) TestJar() {
 	require.NoError(err)
 
 	// Should open without error.
-	j2, err := Open(s.j.filepath)
+	j2, err := Open(s.j.filepath, s.host)
```
markylaing, line 220: What we're aiming for in this work is to prevent the cookie jar from saving cookies that are not from the specified remote. So for this test, the cookiejar should only be saving cookies from `fakehost` - as the test currently stands we want it to fail.

dhvogel: I moved the domain check from jar creation to when the cookies are set, which allows for the expected test behavior. Now, the cookie jar is created with the configured remote address, and each SetCookie checks the incoming domain with the domain of the remote, and only sets the cookie when they match. This is in addition to the PublicSuffixList check, which (contrary to my initial understanding) functions as a deny list for any cookies that have a domain maliciously set to ".com", ".org", etc.

I've added a test for the SetCookie logic, in addition to modifying the cookies in the TestJar test, to include two cookies coming from wrong domains that should be ignored. PTAL at that new test setup.

dhvogel:
I've signed off on the commit and added a key signing as well.

One note is that SetCookie logic will still set cookies for addresses that have a subdomain on an IP address, e.g. `http://sub.127.0.0.1:8443/`, which isn't really a possible address, but I'm new to LXD and I don't know how the networking is configured, and it's better than the publicsuffixlist denylist, but just let me know if these cookies should absolutely be rejected instead.

lxc/cookiejar/jar.go
```diff
@@ -16,29 +18,28 @@ import (
...
 	}
-
 	return jar, nil
 }
 
 // newJar creates a new [Jar] but does not attempt to read existing contents of the jar.
-func newJar(filepath string) (*Jar, error) {
-	// Use the default public suffix list.
+func newJar(filepath string, remote string) (*Jar, error) {
 	jar, err := cookiejar.New(&cookiejar.Options{PublicSuffixList: publicsuffix.List})
```
markylaing: I guess since we're rejecting all addresses except for the given remote there is no need to import the public suffix list anymore.

lxc/cookiejar/jar.go
```diff
@@ -16,29 +18,28 @@ import (
...
 	if err != nil {
-		return nil, fmt.Errorf("Failed to instantiate a cookie jar: %w", err)
+		return nil, fmt.Errorf("failed to instantiate a cookie jar: %w", err)
```
markylaing: Please start all errors with capital letters. We realise this is counter to Go conventions but it's a convention in this codebase.

lxc/cookiejar/jar.go
```diff
@@ -16,29 +18,28 @@ import (
...
 	}
-
-	return &Jar{Jar: jar, filepath: filepath}, nil
+	remoteURL, err := url.Parse(remote)
```
markylaing, line 39: Please add a newline after each functional block. E.g.:
```go
/* logic */
if err != nil {
    /* handle err */
}

remoteURL, err := url.Parse(remote)
```

We have a linter for this, as it generally improves readability.

lxc/cookiejar/jar.go
```diff
@@ -176,3 +178,18 @@ func (j *Jar) Save() error {
...
+	// Note: This will still accept cookies from a subdomain on an IP address, e.g. sub.1.2.3.4
+	host := u.Hostname()
+	allowedHost := j.remote.Hostname()
+
+	if host == allowedHost || strings.HasSuffix(host, "."+allowedHost) {
+		// store cookies with the original URL (includes port)
+		j.Jar.SetCookies(u, cookies)
+	} else {
+		fmt.Printf("Rejected cookie from %s\n", host)
+	}
```
markylaing, line 192: Please just keep this to identical host names for now. If subdomains become a problem in the future, we can fix it then.

Also, please just ignore cookies from other hosts and don't print anything. We don't want to print to stdout as this might break shell scripts.

markylaing (changes requested):
Thank you @dhvogel this is looking really good. Overwriting SetCookies seems like the better place to do this.

There are now just a few small things to look at :)

lxc/cookiejar/jar.go
dhvogel: Ok, I've put a newline between code blocks

lxc/cookiejar/jar.go
dhvogel: removed the subdomain logic and print statement

lxc/cookiejar/jar.go
dhvogel: removed the publicsuffixlist denylist

lxc/cookiejar/jar.go
dhvogel: Capitalized error statements

lxc/cookiejar/jar.go
```diff
@@ -6,39 +6,41 @@ import (
...
-	jar, err := newJar(filepath)
+func Open(filepath string, host string) (*Jar, error) {
+	jar, err := newJar(filepath, host)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to open jar: %w", err)
 	}
 
-	err = jar.openJar()
-	if err != nil {
+	if err := jar.openJar(); err != nil {
```
markylaing: Please keep the assignment and error checking on separate lines

lxc/cookiejar/jar.go
```diff
@@ -176,3 +179,14 @@ func (j *Jar) Save() error {
 
 	return nil
 }
+
+// SetCookies implements [http.CookieJar.SetCookies]. It only accepts cookies from the remote host or its subdomains.
```
markylaing: Comment needs updating to remove the subdomain part

lxc/cookiejar/jar.go
```diff
@@ -6,39 +6,41 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/cookiejar"
+	"net/url"
 	"os"
 	"reflect"
 	"unsafe"
 
-	"golang.org/x/net/publicsuffix"
```
markylaing, line 13: Now that this is removed, can you please run `make update-godeps` and commit the result? We keep these lists to prevent accidental addition/removal of dependencies to some of our packages.

markylaing (changes requested):
Hi @dhvogel thanks for the updates.

I've left a few more comments - we're nearly there :)

Once you addressed those, could you please revert this commit: https://github.com/canonical/lxd/commit/6bc210b2ce71d9555e66c6a8c1fcbbd513771f50

This increased a limit that we place on the `lxc` binary and was required because of the public suffix import. Since the import is no longer required, the lxc binary should come back down in size. Thanks

lxc/cookiejar/jar.go
dhvogel: Yes, I've updated the dependency list for lxc

lxc/cookiejar/jar.go
dhvogel: updated the comment to remove mention of subdomains

lxc/cookiejar/jar.go
dhvogel: separated out the error check to another line

dhvogel:
Ok, I've made the fixes, updated the lxc dependencies, and reverted the binary size change in 6bc210b. I did not know that the publicsuffixlist was big enough to make a meaningful impact on the binary size, removing it seems like a nice bonus.

dhvogel:
It looks like the binary size exceeds 16 MiB in a test?

```
FAIL: lxc binary size exceeds 16MiB
+ echo 'FAIL: lxc binary size exceeds 16MiB'
```

```
+ cur=16843008 + min=15 + min_mib=15728640 + max_mib=16777216
```

The publicsuffixlist is definitely removed, because when I look in the binary for a string that the publicsuffixlist should contain, nothing comes back:

```
$ strings lxc | grep '\.co\.uk'
$
```

Beyond that I'm not really sure how to check for what is the source of the extra 64.3 KiB.

markylaing:
@dhvogel I'm out of office now and for the next few weeks but @tomponline will point you in the right direction.

I suspect there is still a left over import pushing the binary size up.

tomponline:
Also please can you ensure your commits are verified https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

Thanks!

tomponline:
Lets remove the CLI size check revert commit.

dhvogel:
Ok, I've removed the reverted change, signed the CLA, and made sure my change is verified.

Thanks for your help @markylaing and hope you have some restful time off.

tomponline:
Linter is failing, looks like code isnt correctly formatted with `go fmt`

dhvogel:
I think I fixed the lint issues, but I'm having trouble reproducing the CI `golangci-lint` results on my local.

I purposefully left in `require.Len(gotSub, 0)` and ran `golangci-lint run ./lxc --config .golangci.yaml` from the root directory, and it did not see the error:

```
~/Workspaces/lxd $ golangci-lint run ./lxc --config .golangci.yaml
WARN [runner/exclusion_rules] Skipped 0 issues by rules: [Text: "var-naming: avoid meaningless package names", Linters: "revive"] 
0 issues.
```

Any tips on how to set up my local linter so it matches what CI is doing?

dhvogel:
Ok, I was able to reproduce the test linting by installing `testifylint` from `https://github.com/Antonboom/testifylint`.

```
$ ~/go/bin/testifylint ./lxc/cookiejar/
... /jar_test.go:65:2: empty: use require.Empty
```

The source issue was found by `go fmt`.

dhvogel:
I am seeing: 

```
ERROR: changed dependencies for lxc/config; please make sure this is what you want:
```

Looking at the last ten or so changes to test/godeps/lxc-config.list, it looks like they were all made by direct commit, as opposed to a PR with this CI check. I'm not sure how to fix this error in a PR.

tomponline:
Hi, yes you need to update one or more of the files in https://github.com/canonical/lxd/tree/main/test/godeps to confirm you expect the change in go imports for this applications.

dhvogel:
This file appears to be changed.

test/godeps/lxc-config.list

It appears that the change to this file is triggering the CI error, or am I reading this wrong?

tomponline:
yeah, looks like also need to revert the extra dependencies that were added previously.

dhvogel:
Could you point me to another PR that successfully changed test/godeps/lxc-config.list? Maybe I could learn from what the author did there. 

The only changes I see to this file are direct commits that are made specifically to update dependencies. 

One idea for a path forward is to follow that existing pattern, to submit a code change, and then follow up with a dependency file commit.

tomponline:
You just need to undo this change https://github.com/markylaing/lxd/commit/d53f1fcc9dbe9458038591fe95716771c1591a0f without also undoing other changes to that file.

dhvogel:
Ok, I've removed the publicsuffix dependency and left the file otherwise untouched. I guess the next step is to wait for CI to run and see if it passes the PR checks.

dhvogel:
Just checking in... it's been about 24 hours since I last updated this PR and CI hasn't run yet. I'm not sure about CI scheduling, do you know if there is a way to trigger it manually?

tomponline:
> Just checking in... it's been about 24 hours since I last updated this PR and CI hasn't run yet. I'm not sure about CI scheduling, do you know if there is a way to trigger it manually?

Ah yeah please ping me here so i can trigger it.

tomponline:
Please use this style:

```go
err := jar.openJar(); 
if err != nil {
```

dhvogel:
I see, I have changed

```
err := jar.openJar(); if err != nil {
```

to

```
err := jar.openJar() 
if err != nil {
```

tomponline:
looks like go fmt wasnt run on it again, linter is complaining

dhvogel:
This appears to be the change that I just fixed (https://github.com/canonical/lxd/pull/16639#issuecomment-3377550045):

```
  Error: lxc/cookiejar/jar.go:24:1: File is not properly formatted (gci)
  	if err := jar.openJar()
  ^
  1 issues:
  * gci: 1
  
  Error: issues found
  Ran golangci-lint in 94993ms
  ```
  
When I run `go fmt`, then it combines this into one line:

```
err := jar.openJar() 
if err != nil {
```

to  

```
err := jar.openJar(); if err != nil {
```

But this is what makes a custom CI check fail later. 

I see that I did not add a semicolon as you suggested here:
https://github.com/canonical/lxd/pull/16639#issuecomment-3377508054

I have added a semi colon at the end of line 39. I hope this works.

**Comment by tomponline** [comment]
> This appears to be the change that I just fixed ([#16639 (comment)](https://github.com/canonical/lxd/pull/16639#issuecomment-3377550045)):
> 
> ```
>   Error: lxc/cookiejar/jar.go:24:1: File is not properly formatted (gci)
>   	if err := jar.openJar()
>   ^
>   1 issues:
>   * gci: 1
>   
>   Error: issues found
>   Ran golangci-lint in 94993ms
> ```
> 
> When I run `go fmt`, then it combines this into one line:
> 
> ```
> err := jar.openJar() 
> if err != nil {
> ```
> 
> to
> 
> ```
> err := jar.openJar(); if err != nil {
> ```
> 
> But this is what makes a custom CI check fail later.
> 
> I see that I did not add a semicolon as you suggested here: [#16639 (comment)](https://github.com/canonical/lxd/pull/16639#issuecomment-3377508054)
> 
> I have added a semi colon at the end of line 39. I hope this works.

Ive not seen go fmt do that, it wont comply with the linter check.

What about using `gofmt` ?

**Comment by dhvogel** [comment]
gofmt -w ./lxc/cookiejar yields:

```
--- a/lxc/cookiejar/jar.go
+++ b/lxc/cookiejar/jar.go
@@ -21,8 +21,7 @@ func Open(filepath string, host string) (*Jar, error) {
                return nil, fmt.Errorf("Failed to open jar: %w", err)
        }
 
-       if err := jar.openJar(); 
-       err != nil {
+       if err := jar.openJar(); err != nil {
                return nil, fmt.Errorf("Failed to open jar: %w", err)
        }
```

This would fail the second CI check

tomponline:
Please take a look at other examples in the code base where we do error handling, its pretty common.

e.g. (picked at random)

https://github.com/canonical/lxd/blob/main/lxd/api.go#L335-L336

dhvogel:
Tom, I appreciate your continued guidance here. 

I have tried a number of ways to get this if-statement to pass CI, semi-colon, no semi-colon, one line, two lines. The problem is, that there are formatting CI tests that I cannot reproduce on my local, that seem to conflict with each other. One formatting test requires that if statements are one-liners, another requires that they are two-liners. I'm not sure what CI looked like when the above change was submitted two years ago, or if it is the same for two-variable assignments as one-variable assignments, but I need to move on to other things.

tomponline:
> One formatting test requires that if statements are one-liners, another requires that they are two-liners.

This is incorrect. They should be 2 liners. I think you're just getting that impression because your `go fmt` is applying that policy. I've not seen that before, have you tried `gofmt` instead?

tomponline:
ill re-run the CI again as the code looks good to me.

tomponline:
ok thanks, ill fix this one up

tomponline:
Opened https://github.com/canonical/lxd/pull/16697

Issue #16557, opened 2025-09-18
Update firewall guidance for modern Docker compatibility


astrojuanlu:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

[The docs on how to run Docker alongside LXD](https://documentation.ubuntu.com/lxd/stable-5.21/howto/network_bridge_firewalld/#prevent-connectivity-issues-with-lxd-and-docker) give a bunch of not-so-great options. However, according to [one of the maintainers of Moby](https://github.com/moby/moby/issues/14041#issuecomment-3113858552), part of that might be out of date. In particular, this warning:

https://github.com/canonical/lxd/blob/b3bed5d27c68d3193ea8d5ab809d10b4737d9b64/doc/howto/network_bridge_firewalld.md?plain=1#L139-L143

Indeed, from https://www.docker.com/blog/docker-engine-28-hardening-container-networking-by-default/,

> In Docker 28.0, we now explicitly drop unsolicited inbound traffic to each container’s internal IP unless that port was explicitly published (-p or --publish). This doesn’t affect local connections from the Docker host itself, but it does block remote LAN connections to unpublished ports.

### Describe the solution you'd like

Rather than recommending users to uninstall Docker altogether (effective, but probably impractical for many users) or to manipulate `iptables` rules themselves, would it be more helpful to remove that warning from the docs (which seems no longer necessary) and highlight the "Enable IPv4 forwarding" option?

### Additional context

I arrived here after [a long rabbit hole](https://matrix.to/#/!NPPCseDHKRvSBMUEXN:ubuntu.com/$uPbuNF1ZjgSDFGm57fqpumjR0M7l0ZIYM6i1WzxHSD0?via=ubuntu.com&via=matrix.org) that started with me not being able to do `rockcraft pack` after installing Docker.

Plainly applying the advice from [the docs](https://documentation.ubuntu.com/lxd/stable-5.21/howto/network_bridge_firewalld/#prevent-connectivity-issues-with-lxd-and-docker) seemed scary at first, but then when I clicked on the linked issue https://github.com/moby/moby/issues/14041, I found out that it has already been addressed.

tomponline:
Sounds reasonable to me!

astrojuanlu:
[...]

PR #16640, opened 2025-09-29
doc: update docker compatibility guidance

minaelee:
In https://documentation.ubuntu.com/lxd/latest/howto/network_bridge_firewalld/#prevent-connectivity-issues-with-lxd-and-docker, we recommend three approaches to resolve connectivity issues when Docker is installed on the same system as LXD. One of those ways is to enable IPv4 forwarding; however, this included a warning about how this can cause Docker container ports to be reachable from any machine on your local network.

That issue has been resolved on Docker's side as of v28 (see this [closed issue](https://github.com/moby/moby/issues/14041)). The warning in our docs is updated in this PR to include this, along with a link to their [blog announcement](https://www.docker.com/blog/docker-engine-28-hardening-container-networking-by-default/) about the fix.

This PR also moves the "Uninstall Docker" approach to the bottom and moves the IPv4 forwarding solution to the top, since it now involves less of a security concern.

Closes https://github.com/canonical/lxd/issues/16557


tomponline:
linter failures

PR #16644, opened 2025-09-30
lxd/auth/drivers: Set list object max results to 10000

### LINKED ISSUE #16614: Listing instances through fine grained TLS auth is not reliable at scale

```shell
$ snap list --all
Name        Version                 Rev    Tracking       Publisher   Notes
core24      20250729                1151   latest/stable  canonical✓  base
lxd         git-3265003             35903  latest/edge    canonical✓  in-cohort,held
microceph   19.2.1+snap30863f37a3   1518   squid/stable   canonical✓  in-cohort,held
microcloud  2.1.1-202e275           1781   2/stable       canonical✓  in-cohort,held
microovn    24.03.2+snapa2c59c105b  667    24.03/stable   canonical✓  in-cohort,held
snapd       2.71                    25202  latest/stable  canonical✓  snapd
```

### Issue description

On a 7-member cluster loaded with 2500 tiny instances (`busybox` containers without NIC), listing the instances using Unix works reliably. Listing them through a remote using fine grained TLS auth is not reliable:

```
root@pc1-rb3-n1:~# time lxc list -c n -f csv | wc -l
2500

real	0m1.028s
user	0m0.573s
sys	0m0.250s
root@pc1-rb3-n1:~# time lxc list -c n -f csv mc-tls: | wc -l
988

real	0m1.062s
user	0m0.389s
sys	0m0.201s
root@pc1-rb3-n1:~# time lxc list -c n -f csv mc-tls: | wc -l
1002

real	0m0.963s
user	0m0.409s
sys	0m0.214s
root@pc1-rb3-n1:~# lxc list -c n -f csv mc-tls: >/dev/null; echo $?
0
```

The `lxc list` command itself does not return any error though.

The TLS fine grained auth was setup like this:

```
lxc auth identity create tls/foo --group admins
lxc remote add mc-tls https://127.0.0.1:8443 --auth-type tls --token <...>
```

and the 2500 instances were created with `lxd-benchmark init testimage --count 2500 --parallel 25`

The default profile (using ZFS and no NIC):

```
$ lxc profile show default | grep -vF -- '- /1.0/instances/benchmark-'
name: default
description: Default LXD profile
config: {}
devices:
  root:
    path: /
    pool: local
    type: disk
used_by:
project: default
```

markylaing:
Previously, calls to `(openfgav1.OpenFGAServiceServer).ListObjects` limited the number of results to the default value of 1000 (see https://github.com/openfga/openfga/blob/2a454e3772389af0e06f5472eb2d898c264cc732/pkg/server/config/config.go#L25).

This commit increases the limit to 10000 so as to always return all data. It assumes that there are never more than 10000 of a LXD resource.

Note that increasing this limit might unnecessarily increase memory usage because OpenFGA creates a channel of this size when executing the query.

Closes #16614

### DIFF
--- a/lxd/auth/drivers/openfga.go
+++ b/lxd/auth/drivers/openfga.go
@@ -89,6 +89,9 @@ func (e *embeddedOpenFGA) load(ctx context.Context, identityCache *identity.Cach
 		// Enable context propagation to the datastore so that the OpenFGA cache
 		// created for each request can be accessed from our OpenFGA datastore implementation.
 		server.WithContextPropagationToDatastore(true),
+		// Do not limit the number of results returned by calls to ListObjects.
+		// This assumes that there are never more than 10000 of any LXD resource.
+		server.WithListObjectsMaxResults(10_000),
 	}
 
 	e.server, err = server.NewServerWithOpts(openfgaServerOptions...)


lxd/auth/drivers/openfga.go
```diff
@@ -89,6 +89,9 @@ func (e *embeddedOpenFGA) load(ctx context.Context, identityCache *identity.Cach
 		// Enable context propagation to the datastore so that the OpenFGA cache
 		// created for each request can be accessed from our OpenFGA datastore implementation.
 		server.WithContextPropagationToDatastore(true),
+		// Do not limit the number of results returned by calls to ListObjects.
+		// This assumes that there are never more than 10000 of any LXD resource.
+		server.WithListObjectsMaxResults(10_000),
```
tomponline, line 94: whats with the underscore?

markylaing: Go allows underscores for clarity when writing large numbers. I can remove it if preferred?

tomponline: I'd not seen that pattern before, fine with keeping it in if you prefer it for clarity.

tomponline:
@markylaing do we also need to set https://pkg.go.dev/github.com/openfga/openfga/pkg/server#WithListObjectsIteratorCacheMaxResults

markylaing:
> @markylaing do we also need to set https://pkg.go.dev/github.com/openfga/openfga/pkg/server#WithListObjectsIteratorCacheMaxResults

No - we're not using OpenFGA's caching as it means it could return incorrect results if they're still within the TTL

simondeziel:
In https://github.com/canonical/lxd/issues/16614, multiple checks didn't always return the same number of instances, 988 and 1002 for 2 repeated calls. That's pretty close to the 1000 limit you found but I'm wondering why it was not returning a constant partial list?

markylaing:
> In #16614, multiple checks didn't always return the same number of instances, 988 and 1002 for 2 repeated calls. That's pretty close to the 1000 limit you found but I'm wondering why it was not returning a constant partial list?

I wondered about this too and I'm not sure. I was getting 1000 results consistently (though never in the same order).

PR #16655, opened 2025-10-02
Instance: Fix snapshot restore clearing description

### LINKED ISSUE #15256: Instance snapshot restore clears description
### Distribution

snap

### Distribution version

6/stable

```shell
lxd                                   6.3-d704dcb                     32918  6/stable         canonical**     -
```

```shell
.
```

### Issue description

When restoring an instance snapshot, the description of the instance is reset to empty string. This seems like a bug. From the research below, I think the description should be not changed on snapshot restore.

For storage volumes, the description is unchanged on restoring an older snapshot. Which seems okay.

On instance snapshots, the profiles applied to the instance are not part of the snapshot and not restored either. They are kept as is after the restore process. This seems okay.

### Steps to reproduce

1. lxc launch ubuntu:24.04 precise-ray
2. lxc config edit precise-ray (set a description)
3. lxc config show precise-ray (show the description)
4. lxc info precise-ray (show snapshots)
5. lxc snapshot precise-ray (create a snapshot)
6. lxc restore precise-ray snap0 (restore snapshot)
7. lxc info precise-ray (description is reset to empty string)

tugbataluy:
Closes #15256

### DIFF
--- a/lxd/instance/drivers/driver_common.go
+++ b/lxd/instance/drivers/driver_common.go
@@ -750,6 +750,7 @@ func (d *common) snapshotCommon(inst instance.Instance, name string, expiry *tim
 		Stateful:     stateful,
 		ExpiryDate:   *expiry,
 		CreationDate: snapshotCreationDate,
+		Description:  inst.Description(),
 	}
 
 	// Create the snapshot.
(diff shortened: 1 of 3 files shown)


tugbataluy:
Sure, I'll add it. Thank you.

tomponline:
@tugbataluy shall we close this one or move it to draft?

tugbataluy:
I think we can close it

PR #16658, opened 2025-10-02
binfmt_misc apparmor policy fix and documentation

kadinsayani:
https://github.com/canonical/lxd/pull/14148 introduced support for `binfmt_misc` namespaces. This pull request updates LXD's apparmor policy and adds documentation on enabling `binfmt_misc` in an unprivileged container.

lxd-ci tests: https://github.com/canonical/lxd-ci/pull/591.

### DIFF
--- a/doc/container-environment.md
+++ b/doc/container-environment.md
@@ -81,6 +81,21 @@ If LXCFS is present on the host, it is automatically set up for the container.
 This normally results in a number of `/proc` files being overridden through bind-mounts.
 On older kernels, a virtual version of `/sys/fs/cgroup` might also be set up by LXCFS.
 
+(container-runtime-environment-binfmt-misc)=
+### `binfmt_misc`
+
+The `binfmt_misc` kernel feature allows the system to run binaries for foreign architectures by registering interpreters for different binary types. See the [kernel documentation](https://docs.kernel.org/admin-guide/binfmt-misc.html) for more details.
+
+On kernels that lack `binfmt_misc` namespace support, LXD automatically bind-mounts `/proc/sys/fs/binfmt_misc` from the host into the container. This allows containers to use the host's registered binary types.
+
+On kernels with `binfmt_misc` namespace support, unprivileged containers can mount their own isolated `binfmt_misc`. To enable `binfmt_misc` in an unprivileged container, run:
+
+```bash
+mount binfmt_misc -t binfmt_misc /proc/sys/fs/binfmt_misc
+```
+
+The container can now register its own binary types without affecting the host or other containers. This is useful for running foreign architecture binaries within the container.
+
 ## PID1
 
 LXD spawns whatever is located at `/sbin/init` as the initial process of the container (PID 1).


lxd/apparmor/instance_lxc.go
```diff
@@ -366,7 +379,7 @@ profile "{{ .name }}" flags=(attach_disconnected,mediate_deleted) {
   deny /proc/sys/net?*{,/**} wklx,
 
   # Block dangerous paths under /sys
-  deny /sys/[^fdck]*{,/**} wklx,
+  deny /sys/[^fdc]*{,/**} wklx,
```
simondeziel: should read as `[^cdf]` I think.

lxd/apparmor/instance_lxc.go
```diff
@@ -327,7 +326,21 @@ profile "{{ .name }}" flags=(attach_disconnected,mediate_deleted) {
...
+  deny /proc/sys/fs/bin[^f]*{,/**} wklx,
+  deny /proc/sys/fs/binf[^m]*{,/**} wklx,
+  deny /proc/sys/fs/binfm[^t]*{,/**} wklx,
+  deny /proc/sys/fs/binfmt[^_]*{,/**} wklx,
+  deny /proc/sys/fs/binfmt_[^m]*{,/**} wklx,
+  deny /proc/sys/fs/binfmt_m[^i]*{,/**} wklx,
+  deny /proc/sys/fs/binfmt_mi[^s]*{,/**} wklx,
+  deny /proc/sys/fs/binfmt_mis[^c]*{,/**} wklx,
+  deny /proc/sys/fs/binfmt_misc?*{,/**} wklx,
+  deny /proc/sys/fs?*{,/**} wklx,
```
simondeziel, line 343: Gah, I really don't like this hack but I don't see a way around it :/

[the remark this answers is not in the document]
kadinsayani: https://github.com/canonical/lxd-ci/pull/591

doc/container-environment.md
```diff
@@ -81,6 +81,20 @@ If LXCFS is present on the host, it is automatically set up for the container.
...
+### `binfmt_misc`
+
+The `binfmt_misc` kernel feature allows the system to run binaries for foreign architectures by registering interpreters for different binary types. See the [kernel documentation](https://docs.kernel.org/admin-guide/binfmt-misc.html) for more details.
+
+On kernels that lack `binfmt_misc` namespace support, LXD automatically bind-mounts `/proc/sys/fs/binfmt_misc` from the host into the container. This allows containers to use the host's registered binary types.
+
+On kernels with `binfmt_misc` namespace support, unprivileged containers can mount their own isolated `binfmt_misc`. To enable `binfmt_misc` in an unprivileged container, run:
+
+```bash
+mount binfmt_misc -t binfmt_misc /proc/sys/fs/binfmt_misc
```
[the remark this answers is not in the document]
kadinsayani: https://docs.kernel.org/admin-guide/binfmt-misc.html

lxd/apparmor/instance_lxc.go
kadinsayani: Likewise :/

doc/container-environment.md
```diff
@@ -81,6 +81,20 @@ If LXCFS is present on the host, it is automatically set up for the container.
 This normally results in a number of `/proc` files being overridden through bind-mounts.
 On older kernels, a virtual version of `/sys/fs/cgroup` might also be set up by LXCFS.
 
+### `binfmt_misc`
```
minaelee, line 85:
```suggestion
(container-runtime-environment-binfmt-misc)=
### `binfmt_misc`
```

minaelee (changes requested):
LGTM, just one request to add a link target for any new documentation headings, making it available for linking from elsewhere. (You could link using the HTML anchor, but that's fragile as the heading text can change, or the file could move to a different directory.)

lxd/apparmor/instance_lxc.go
tomponline: What is the hack here? 

Sounds like it could do with a comment as to what its doing and why there isn't a better way.

tomponline:
LGTM, but will wait for @mihalicyn feedback on this one. Thanks!

lxd/apparmor/instance_lxc.go
simondeziel: The problem is that `deny` rules have precedence over `allow` ones. This means that to allow `/proc/sys/fs/binfmt_misc/**` but not `/proc/sys/fs/binfmt_misd/**` we need to do this gymnastic with denying all possible alternations leading to the desired `allow` rule.

tomponline: Excuse my ignorance, but shouldnt we be denying everything and only allowing the things we want through?

simondeziel: By default, that's how Appamor works. Everything not explicitly allowed is denied. Everything explicitly denied takes precedence over what's explicitly allowed. Weak default `deny` < `allow` < explicit `deny`.

I **think**, the `file` rule at the top of this profile is what provides the explicit global `allow` that we then need surgically restrict for known dangerous bits like stuff under `/proc/sys` that we want to permit but only partly.

tomponline: OK thanks.

Before proceeding I'd like to chat with @mihalicyn about this and understand why we dont have a default block policy and only allow the things we want.

mihalicyn: Yeah, this seems correct to me.

> understand why we dont have a default block policy and only allow the things we want.

unfortunately, if we prohibit something, we can no longer allow it with the rule that follows after.
That's why we have this trickery with having a very precise rules to deny everything except a part of the path, then more and so on.

PR #16660, opened 2025-10-02
tests: add bash completions for test funcs

kadinsayani:


test/main.sh.bash-completion
```diff
@@ -0,0 +1,35 @@
...
+    # Extract all test function names from main.sh
+    # Remove 'test_' prefix and 'run_test test_' wrapper
+    local script_dir
+    script_dir="$(dirname "${BASH_SOURCE[0]}")"
+    if [ ! -f "$script_dir/main.sh" ]; then
+        script_dir="$(pwd)"
+    fi
+
+    local tests
+    tests=$(grep -oP 'run_test test_\K[a-z_]+' "$script_dir/main.sh" 2>/dev/null | sort -u)
```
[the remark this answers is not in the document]
kadinsayani: We can make that change if/when needed.

test/main.sh.bash-completion
```diff
@@ -0,0 +1,35 @@
...
+
+    # Special keywords
+    local keywords="all cluster standalone test-shell"
+
+    # Extract all test function names from main.sh
+    # Remove 'test_' prefix and 'run_test test_' wrapper
+    local script_dir
+    script_dir="$(dirname "${BASH_SOURCE[0]}")"
+    if [ ! -f "$script_dir/main.sh" ]; then
+        script_dir="$(pwd)"
```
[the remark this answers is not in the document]
kadinsayani: I think we're ok here. Setting `script_dir` to `cwd` is already the fallback.

tomponline (approved):
Thanks!

test/main.sh.bash-completion
```diff
@@ -0,0 +1,35 @@
...
+_lxd_test_completion() {
+    local cur="${COMP_WORDS[COMP_CWORD]}"
+
+    # Special keywords
+    local keywords="all cluster standalone test-shell"
+
+    # Extract all test function names from main.sh
+    # Remove 'test_' prefix and 'run_test test_' wrapper
+    local script_dir
+    script_dir="$(dirname "${BASH_SOURCE[0]}")"
```
simondeziel, line 19: We already set `MAIN_DIR` in `test/main.sh`.

test/main.sh.bash-completion
simondeziel, line 25: Wow, nice use of `\K`. That's something that should be reused in `test/lint/test-tests.sh` which does a similar extraction.

simondeziel (commented):
Thanks for putting this up!

test/main.sh.bash-completion
kadinsayani: We need to set it here too since we are using the completions before sourcing `test/main.sh`.

test/main.sh.bash-completion
kadinsayani: Copilot suggested using `\K` in this grep expression :)

test/main.sh.bash-completion
simondeziel: Maybe we use it differently. How else than `make test-shell` do you use those completions?

kadinsayani: I only use the completions in `make test-shell`, but others may run `./main.sh` from the main shell on their host.

tomponline: > but others may run ./main.sh from the main shell on their host

Indeed I do

simondeziel: Thank you both. I've reworked #16664 to remain compatible with direct invocations of `./main.sh`.

Issue #16481, opened 2025-09-12
MicroCloud on Raspberry Pi 5 – Error parsing /proc/cpuinfo


raphael-adeshiyan:
Hi, I’m trying to set up a MicroCloud cluster on two Raspberry Pi 5 boards.

* **piserver1**: Raspberry Pi 5, 16 GB RAM – where I ran `microcloud init`
* **piserver2**: Raspberry Pi 5, 4 GB RAM – where I tried to join the cluster

After entering the secret word on piserver2 and accepting the join on piserver1, I get this error:

```
Error: Failed to get system resources of peer "piserver1": 
Failed to retrieve CPU information: 
Failed to parse /proc/cpuinfo: Unexpected line "Revision  : e04171"
```

I tried a workaround by manually editing `/proc/cpuinfo` to remove the `Revision` line. Then I retried the join, but got a similar error:

```
Error: Failed to get system resources of peer "piserver1": 
Failed to retrieve CPU information: 
Failed to parse /proc/cpuinfo: Unexpected line "Serial    : cd3d211d47ad44c2"
```

So it looks like MicroCloud’s CPU parser doesn’t recognize the extra fields that the Raspberry Pi 5 reports in `/proc/cpuinfo` (`Revision`, `Serial`, etc.).

* How can I fix or work around this so the cluster setup can complete?
* Is there a version of MicroCloud/LXD where ARM `/proc/cpuinfo` parsing has already been fixed?
* Or is there a recommended patch/hack?

Thanks in advance!

simondeziel:
@raphael-adeshiyan it would help if you could attach the content of `/proc/cpuinfo` from both systems. It would be useful to reproduce the issue and also augment our unit tests to make sure it never regresses once fixed. Thanks!

borathexplorer:
Hey folks, @raphael-adeshiyan @simondeziel 

I had the exact same issue.

This is caused by a small detail on LXD cpu information scanner, not an issue on the MicroCloud side. Scanner expects each block of the contents of `/proc/cpuinfo` to start with the word `processor`. However, pi5 seems to have some metadata at the end and therefore causes the scan to fail.

In this case, running `journalctl -u snap.lxd.daemon -n 200 --no-pager`on the unit (`piserver1`) should show the exact same error.

It looks like MicroCloud is just not able to get the resource information from LXD because of this.

[...]

PR #16662, opened 2025-10-03
lxd/resources/cpu: make /proc/cpuinfo parser less strict (stable-5.21)

mihalicyn:
While for many CPU architectures in /proc/cpuinfo output, each CPU info block starts with "processor\s+:" (regex) block, it is not always the case.

Also, some architectures, for example, arm tend to add an additional text block at the beginning of the file or at the end.

Let's make our /proc/cpuinfo parser less strict and just skip the block we can't handle.

Fixes: #16481

(cherry picked from commit a6c9f361eae184625943869d4dcd7ad7bebb4764)


tomponline:
Thanks! Please can you also cherry-pick this into 5.21-candidate?

mihalicyn:
> Please can you also cherry-pick this into 5.21-candidate?

Have done https://github.com/canonical/lxd-pkg-snap/pull/937

PR #16679, opened 2025-10-06
Fix Instance snapshot restore clears description

### LINKED ISSUE #15256: Instance snapshot restore clears description

tugbataluy:
Closes #15256


test/suites/snapshots.sh
```diff
@@ -182,6 +182,14 @@ snap_restore() {
 
   ## prepare snap0
   lxc launch testimage bar -d "${SMALL_ROOT_DISK}"
+
+  # set description via config edit YAML
+  temp_yaml=$(mktemp)
+  lxc config show bar > "${temp_yaml}"
+  echo "description: test_description_snap0" >> "${temp_yaml}"
```
tomponline: can you use `lxc config set --property` to set description?

test/suites/snapshots.sh
```diff
@@ -256,6 +264,13 @@ snap_restore() {
       false
     fi
 
+     # Check description has been restored to snap0 value via YAML
+    if ! lxc config show bar | grep "description: test_description_snap0" >> /dev/null; then
```
tomponline: can you use `lxc config get --property` here?

tugbataluy: of course, thanks :)

test/suites/snapshots.sh
tugbataluy: Sure :)

tomponline:
Please can you rebase and squash your 2 test commits together so they show the final version and not the iteration of the pr.

test/suites/snapshots.sh
```diff
@@ -182,7 +182,11 @@ snap_restore() {
 
   ## prepare snap0
   lxc launch testimage bar -d "${SMALL_ROOT_DISK}"
-  echo snap0 > state
+
+  # set description
+  lxc config set bar --property description="test_description_snap0"
+
+  echo snap0 >state
```
tomponline: @tugbataluy would you mind changing this line back to how it was?

test/suites/snapshots.sh
```diff
@@ -204,7 +208,7 @@ snap_restore() {
 
   ## prepare snap1
   lxc start bar
-  echo snap1 > state
+  echo snap1 >state
```
tomponline: @tugbataluy would you mind changing this line back to how it was, it doesn't seem related to the PR aim?

test/suites/snapshots.sh
```diff
@@ -440,9 +453,9 @@ test_snap_volume_db_recovery() {
   lxc snapshot c1
   lxc start c1
   lxc stop -f c1
-  lxd sql global 'DELETE FROM storage_volumes_snapshots' # Remove volume snapshot DB records.
+  lxd sql global 'DELETE FROM storage_volumes_snapshots'                               # Remove volume snapshot DB records.
```
tomponline: looks like your editor is changing things on save, but these are not related to the test commit:

> test/suites/snapshot: Add test to preserve definition in snap_restore 

So they should either be reverted or added in a separate commit.

test/suites/snapshots.sh
```diff
@@ -301,8 +314,8 @@ snap_restore() {
   # Check config value in snapshot has been restored
   cpus=$(lxc config get bar limits.cpu)
   if [ "${cpus}" != "1" ]; then
-   echo "==> config didn't match expected value after restore (${cpus})"
-   false
+    echo "==> config didn't match expected value after restore (${cpus})"
```
tomponline, line 315: Unrelated to commit.

test/suites/snapshots.sh
```diff
@@ -256,6 +260,15 @@ snap_restore() {
       false
     fi
 
+    # Check container description is restored
+    description=$(lxc config get bar --property description)
+    if [ "${description}" != "test_description_snap0" ]; then
```
tomponline, line 265: This section is quite verbose and doesn't tell us any more than the simpler assertion:

```
[ "${description}" = "test_description_snap0" ]
```

If this evaluates to `false` then the test will fail.

tomponline (changes requested):
Thanks @tugbataluy 

This is nearly there.

Please can you simplify the test by just using an assertion, and ensure that any unrelated whitespace changes are in separate commits. Thanks

tomponline (changes requested):
Hi @tugbataluy the last commit "test/suites/snapshot: Remove unnecessary whitespaces in snaphot.sh" still includes the change that re-adds the whitespace removed from the earlier commit from ">state" to "> state".

That change should be squashed into the earlier commit, and the last commit should just contain the whitespace removal.

tugbataluy:
Okay thanks Thomas I'll be fixing that :smile:

PR #16681, opened 2025-10-06
lxd/storage: source for ceph and cephfs pools is a global config option

edlerd:
# Done
- lxd/storage: source for ceph and cephfs pools is a global config option
- see https://github.com/canonical/lxd-ui/issues/1526


edlerd:
Closing as internally the source field is always handled local, so the current state is correct.

PR #16686, opened 2025-10-06
lxd: Mark images as cached consistently

### LINKED ISSUE #16576: Marking images as cached is inconsistent in projects which share images with the default project
### Please confirm

### Distribution

Ubuntu

### Distribution version

24.04

```shell
Name    Version      Rev    Tracking       Publisher   Notes
core24  20250729     1151   latest/stable  canonical✓  base
lxd     6.5-22da890  35616  6/stable       canonical✓  -
snapd   2.71         25202  latest/stable  canonical✓  snapd
```

```shell
config: {}
api_extensions:
- storage_zfs_remove_snapshots
- container_host_shutdown_timeout
- container_stop_priority
- container_syscall_filtering
- auth_pki
- container_last_used_at
- etag
- patch
- usb_devices
- https_allowed_credentials
- image_compression_algorithm
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
- unix_device_rename
- storage_lvm_use_thinpool
- storage_rsync_bwlimit
- network_vxlan_interface
- storage_btrfs_mount_options
- entity_description
- image_force_refresh
- storage_lvm_lv_resizing
- id_map_base
- file_symlinks
- container_push_target
- network_vlan_physical
- storage_images_delete
- container_edit_metadata
- container_snapshot_stateful_migration
- storage_driver_ceph
- storage_ceph_user_name
- resource_limits
- storage_volatile_initial_source
- storage_ceph_force_osd_reuse
- storage_block_filesystem_btrfs
- resources
- kernel_limits
- storage_api_volume_rename
- network_sriov
- console
- restrict_devlxd
- migration_pre_copy
- infiniband
- maas_network
- devlxd_events
- proxy
- network_dhcp_gateway
- file_get_symlink
- network_leases
- unix_device_hotplug
- storage_api_local_volume_handling
- operation_description
- clustering
- event_lifecycle
- storage_api_remote_volume_handling
- nvidia_runtime
- container_mount_propagation
- container_backup
- devlxd_images
- container_local_cross_pool_handling
- proxy_unix
- proxy_udp
- clustering_join
- proxy_tcp_udp_multi_port_handling
- network_state
- proxy_unix_dac_properties
- container_protection_delete
- unix_priv_drop
- pprof_http
- proxy_haproxy_protocol
- network_hwaddr
- proxy_nat
- network_nat_order
- container_full
- backup_compression
- nvidia_runtime_config
- storage_api_volume_snapshots
- storage_unmapped
- projects
- network_vxlan_ttl
- container_incremental_copy
- usb_optional_vendorid
- snapshot_scheduling
- snapshot_schedule_aliases
- container_copy_project
- clustering_server_address
- clustering_image_replication
- container_protection_shift
- snapshot_expiry
- container_backup_override_pool
- snapshot_expiry_creation
- network_leases_location
- resources_cpu_socket
- resources_gpu
- resources_numa
- kernel_features
- id_map_current
- event_location
- storage_api_remote_volume_snapshots
- network_nat_address
- container_nic_routes
- cluster_internal_copy
- seccomp_notify
- lxc_features
- container_nic_ipvlan
- network_vlan_sriov
- storage_cephfs
- container_nic_ipfilter
- resources_v2
- container_exec_user_group_cwd
- container_syscall_intercept
- container_disk_shift
- storage_shifted
- resources_infiniband
- daemon_storage
- instances
- image_types
- resources_disk_sata
- clustering_roles
- images_expiry
- resources_network_firmware
- backup_compression_algorithm
- ceph_data_pool_name
- container_syscall_intercept_mount
- compression_squashfs
- container_raw_mount
- container_nic_routed
- container_syscall_intercept_mount_fuse
- container_disk_ceph
- virtual-machines
- image_profiles
- clustering_architecture
- resources_disk_id
- storage_lvm_stripes
- vm_boot_priority
- unix_hotplug_devices
- api_filtering
- instance_nic_network
- clustering_sizing
- firewall_driver
- projects_limits
- container_syscall_intercept_hugetlbfs
- limits_hugepages
- container_nic_routed_gateway
- projects_restrictions
- custom_volume_snapshot_expiry
- volume_snapshot_scheduling
- trust_ca_certificates
- snapshot_disk_usage
- clustering_edit_roles
- container_nic_routed_host_address
- container_nic_ipvlan_gateway
- resources_usb_pci
- resources_cpu_threads_numa
- resources_cpu_core_die
- api_os
- container_nic_routed_host_table
- container_nic_ipvlan_host_table
- container_nic_ipvlan_mode
- resources_system
- images_push_relay
- network_dns_search
- container_nic_routed_limits
- instance_nic_bridged_vlan
- network_state_bond_bridge
- usedby_consistency
- custom_block_volumes
- clustering_failure_domains
- resources_gpu_mdev
- console_vga_type
- projects_limits_disk
- network_type_macvlan
- network_type_sriov
- container_syscall_intercept_bpf_devices
- network_type_ovn
- projects_networks
- projects_networks_restricted_uplinks
- custom_volume_backup
- backup_override_name
- storage_rsync_compression
- network_type_physical
- network_ovn_external_subnets
- network_ovn_nat
- network_ovn_external_routes_remove
- tpm_device_type
- storage_zfs_clone_copy_rebase
- gpu_mdev
- resources_pci_iommu
- resources_network_usb
- resources_disk_address
- network_physical_ovn_ingress_mode
- network_ovn_dhcp
- network_physical_routes_anycast
- projects_limits_instances
- network_state_vlan
- instance_nic_bridged_port_isolation
- instance_bulk_state_change
- network_gvrp
- instance_pool_move
- gpu_sriov
- pci_device_type
- storage_volume_state
- network_acl
- migration_stateful
- disk_state_quota
- storage_ceph_features
- projects_compression
- projects_images_remote_cache_expiry
- certificate_project
- network_ovn_acl
- projects_images_auto_update
- projects_restricted_cluster_target
- images_default_architecture
- network_ovn_acl_defaults
- gpu_mig
- project_usage
- network_bridge_acl
- warnings
- projects_restricted_backups_and_snapshots
- clustering_join_token
- clustering_description
- server_trusted_proxy
- clustering_update_cert
- storage_api_project
- server_instance_driver_operational
- server_supported_storage_drivers
- event_lifecycle_requestor_address
- resources_gpu_usb
- clustering_evacuation
- network_ovn_nat_address
- network_bgp
- network_forward
- custom_volume_refresh
- network_counters_errors_dropped
- metrics
- image_source_project
- clustering_config
- network_peer
- linux_sysctl
- network_dns
- ovn_nic_acceleration
- certificate_self_renewal
- instance_project_move
- storage_volume_project_move
- cloud_init
- network_dns_nat
- database_leader
- instance_all_projects
- clustering_groups
- ceph_rbd_du
- instance_get_full
- qemu_metrics
- gpu_mig_uuid
- event_project
- clustering_evacuation_live
- instance_allow_inconsistent_copy
- network_state_ovn
- storage_volume_api_filtering
- image_restrictions
- storage_zfs_export
- network_dns_records
- storage_zfs_reserve_space
- network_acl_log
- storage_zfs_blocksize
- metrics_cpu_seconds
- instance_snapshot_never
- certificate_token
- instance_nic_routed_neighbor_probe
- event_hub
- agent_nic_config
- projects_restricted_intercept
- metrics_authentication
- images_target_project
- cluster_migration_inconsistent_copy
- cluster_ovn_chassis
- container_syscall_intercept_sched_setscheduler
- storage_lvm_thinpool_metadata_size
- storage_volume_state_total
- instance_file_head
- instances_nic_host_name
- image_copy_profile
- container_syscall_intercept_sysinfo
- clustering_evacuation_mode
- resources_pci_vpd
- qemu_raw_conf
- storage_cephfs_fscache
- network_load_balancer
- vsock_api
- instance_ready_state
- network_bgp_holdtime
- storage_volumes_all_projects
- metrics_memory_oom_total
- storage_buckets
- storage_buckets_create_credentials
- metrics_cpu_effective_total
- projects_networks_restricted_access
- storage_buckets_local
- loki
- acme
- internal_metrics
- cluster_join_token_expiry
- remote_token_expiry
- init_preseed
- storage_volumes_created_at
- cpu_hotplug
- projects_networks_zones
- network_txqueuelen
- cluster_member_state
- instances_placement_scriptlet
- storage_pool_source_wipe
- zfs_block_mode
- instance_generation_id
- disk_io_cache
- amd_sev
- storage_pool_loop_resize
- migration_vm_live
- ovn_nic_nesting
- oidc
- network_ovn_l3only
- ovn_nic_acceleration_vdpa
- cluster_healing
- instances_state_total
- auth_user
- security_csm
- instances_rebuild
- numa_cpu_placement
- custom_volume_iso
- network_allocations
- storage_api_remote_volume_snapshot_copy
- zfs_delegate
- operations_get_query_all_projects
- metadata_configuration
- syslog_socket
- event_lifecycle_name_and_project
- instances_nic_limits_priority
- disk_initial_volume_configuration
- operation_wait
- cluster_internal_custom_volume_copy
- disk_io_bus
- storage_cephfs_create_missing
- instance_move_config
- ovn_ssl_config
- init_preseed_storage_volumes
- metrics_instances_count
- server_instance_type_info
- resources_disk_mounted
- server_version_lts
- oidc_groups_claim
- loki_config_instance
- storage_volatile_uuid
- import_instance_devices
- instances_uefi_vars
- instances_migration_stateful
- container_syscall_filtering_allow_deny_syntax
- access_management
- vm_disk_io_limits
- storage_volumes_all
- instances_files_modify_permissions
- image_restriction_nesting
- container_syscall_intercept_finit_module
- device_usb_serial
- network_allocate_external_ips
- explicit_trust_token
- shared_custom_block_volumes
- instance_import_conversion
- instance_create_start
- instance_protection_start
- devlxd_images_vm
- disk_io_bus_virtio_blk
- metrics_api_requests
- projects_limits_disk_pool
- ubuntu_pro_guest_attach
- metadata_configuration_entity_types
- access_management_tls
- network_allocations_ovn_uplink
- network_ovn_uplink_vlan
- state_logical_cpus
- vm_limits_cpu_pin_strategy
- gpu_cdi
- images_all_projects
- metadata_configuration_scope
- unix_device_hotplug_ownership_inherit
- unix_device_hotplug_subsystem_device_option
- storage_ceph_osd_pool_size
- network_get_target
- network_zones_all_projects
- vm_root_volume_attachment
- projects_limits_uplink_ips
- entities_with_entitlements
- profiles_all_projects
- storage_driver_powerflex
- storage_driver_pure
- cloud_init_ssh_keys
- oidc_scopes
- project_default_network_and_storage
- client_cert_presence
- clustering_groups_used_by
- container_bpf_delegation
- override_snapshot_profiles_on_copy
- resources_device_fs_uuid
- backup_metadata_version
- storage_buckets_all_projects
- network_acls_all_projects
- networks_all_projects
- clustering_restore_skip_mode
- disk_io_threads_virtiofsd
- oidc_client_secret
- pci_hotplug
- device_patch_removal
api_status: stable
api_version: "1.0"
auth: trusted
public: false
auth_methods:
- tls
client_certificate: false
auth_user_name: root
auth_user_method: unix
environment:
  addresses: []
  architectures:
  - x86_64
  - i686
  backup_metadata_version_range:
  - 1
  - 2
  certificate: |
    -----BEGIN CERTIFICATE-----
    MIIB1zCCAV2gAwIBAgIQdF3vBNY8i8wJLUIq93V0zTAKBggqhkjOPQQDAzAgMQww
    CgYDVQQKEwNMWEQxEDAOBgNVBAMMB3Jvb3RAdjEwHhcNMjUwOTIxMjMxMzE2WhcN
    MzUwOTE5MjMxMzE2WjAgMQwwCgYDVQQKEwNMWEQxEDAOBgNVBAMMB3Jvb3RAdjEw
    djAQBgcqhkjOPQIBBgUrgQQAIgNiAATeWo/8Apa5iI5o+MUT4yUN3808K578hByl
    l99uSuFPoZYNKNS9Qp4tkKLdr12WfHd0Fa8k/9hGeZMLI6dul5ts2tblxgwlSUoU
    PUV/eGaTEjBKoO7l61v8aEUbANji18CjXDBaMA4GA1UdDwEB/wQEAwIFoDATBgNV
    HSUEDDAKBggrBgEFBQcDATAMBgNVHRMBAf8EAjAAMCUGA1UdEQQeMByCAnYxhwR/
    AAABhxAAAAAAAAAAAAAAAAAAAAABMAoGCCqGSM49BAMDA2gAMGUCMQDGQRwlRtHu
    wVLrBb2pR2w+vO3DSj0GbSl8ngDVXFBWCK7SRx/9MKMPyyug2cDNN40CMHRUgP+/
    Ciocxgn/TVHz/YLLOcsss2MKnGy4Gi63mh6Asby1rsPIhHDG97Xhwy3r1g==
    -----END CERTIFICATE-----
  certificate_fingerprint: 231e98983eb01f27dd155a877aeabb8fa4964a5fd36063c0738f0ecd80d94872
  driver: lxc | qemu
  driver_version: 6.0.4 | 8.2.2
  instance_types:
  - container
  - virtual-machine
  firewall: nftables
  kernel: Linux
  kernel_architecture: x86_64
  kernel_features:
    bpf_token: "false"
    idmapped_mounts: "true"
    netnsid_getifaddrs: "true"
    seccomp_listener: "true"
    seccomp_listener_continue: "true"
    uevent_injection: "true"
    unpriv_binfmt: "true"
    unpriv_fscaps: "true"
  kernel_version: 6.8.0-71-generic
  lxc_features:
    cgroup2: "true"
    core_scheduling: "true"
    devpts_fd: "true"
    idmapped_mounts_v2: "true"
    mount_injection_file: "true"
    network_gateway_device_route: "true"
    network_ipvlan: "true"
    network_l2proxy: "true"
    network_phys_macvlan_mtu: "true"
    network_veth_router: "true"
    pidfd: "true"
    seccomp_allow_deny_syntax: "true"
    seccomp_notify: "true"
    seccomp_proxy_send_notify_fd: "true"
  os_name: Ubuntu
  os_version: "24.04"
  project: p1
  server: lxd
  server_clustered: false
  server_event_mode: full-mesh
  server_name: v1
  server_pid: 2592
  server_version: "6.5"
  server_lts: false
  storage: dir
  storage_version: "1"
  storage_supported_drivers:
  - name: cephobject
    version: 19.2.1
    remote: true
  - name: dir
    version: "1"
    remote: false
  - name: lvm
    version: 2.03.16(2) (2022-05-18) / 1.02.185 (2022-05-18) / 4.48.0
    remote: false
  - name: powerflex
    version: 2.8 (nvme-cli)
    remote: true
  - name: cephfs
    version: 19.2.1
    remote: true
  - name: pure
    version: 2.1.9 (iscsiadm) / 2.8 (nvme-cli)
    remote: true
  - name: zfs
    version: 2.2.2-0ubuntu9.2
    remote: false
  - name: btrfs
    version: 6.6.3
    remote: false
  - name: ceph
    version: 19.2.1
    remote: true
```

### Issue description

When using a project with the setting `features.images=false`, working with images usually works the same as using the default project. An exception is whether LXD sets `cached` or not.

In the default project, and projects with `feature.images=true`:
- Creating an instance will set `cached=true` if the image doesn't exist locally.
- Manually downloading an image will set `cached=false` regardless of whether the image exists locally.

I would expect projects with `features.images=false` to work the same, but:
- Creating an instance will never set `cached=true` on the image.
- Manually downloading an image will only set `cached=false` if the image doesn't exist locally.

### Steps to reproduce

In the following commands I assume the image isn't updated between steps. This usually happens around 1AM UTC. If the image is updated, you can update the `fp` variable manually.

First, check `cached=true` for a new image, downloaded implicitly:
```console
$ lxc init images:alpine/3.22 a1
Creating a1
$ fp="$(lxc config show a1 | sed -En 's/^.*volatile.base_image: (.*)$/\1/p')"
$ lxc delete a1
$ lxc image info "$fp" | grep Cached
Cached: yes
```

Next, check `cached=false` when downloading manually:
```console
$ lxc image copy images:alpine/3.22 local:
Image copied successfully!
$ lxc image info "$fp" | grep Cached
Cached: no
```

The same output should be observed if you create a new project first:
```console
$ lxc image delete "$fp"
$ lxc project create --network lxdbr0 --storage default p1
Project p1 created
$ lxc project switch p1
$ lxc init images:alpine/3.22 a1
Creating a1
$ lxc delete a1
$ lxc image info "$fp" | grep Cached
Cached: yes
$ lxc image copy images:alpine/3.22 local:
Image copied successfully!
$ lxc image info "$fp" | grep Cached
Cached: no
```

However, if we set `features.images=false`, then `cached=false` for implicit downloads:
```console
$ lxc image delete "$fp"
$ lxc project set p1 features.images=false
$ lxc init images:alpine/3.22 a1
Creating a1
$ lxc delete a1
$ lxc image info "$fp" | grep Cached
Cached: no
```

Also, explicit downloads don't set `cached=false`:
```console
$ lxc image delete "$fp"
$ lxc project switch default
$ lxc init images:alpine/3.22 a1
Creating a1
$ lxc delete a1
$ lxc image info "$fp" | grep Cached
Cached: yes
$ lxc project switch p1
$ lxc image copy images:alpine/3.22 local:
Image copied successfully!
$ lxc image info "$fp" | grep Cached
Cached: yes
```

### Information to attach

nmezhenskyi:
This PR makes marking images as cached consistent among projects that have `features.images` disabled.

Fixes https://github.com/canonical/lxd/issues/16576.

## Changes:
- If `features.images` is disabled for the project, associate the image with the `default` project inside of the `ensureDownloadedImageFitWithinBudget()` function. This ensures that `cached=yes` is set for a new image downloaded implicitly using `lxc init <image> <instance_name>`.
-  If `features.images` is disabled for the project, associate the image with the `default` project during `lxc copy`. This ensures that `cached=no` is set for a new image downloaded explicitly using `lxc image copy`.
- Correctly associate profile IDs for images that are shared between the `default` project and custom projects with `features.images=false`.
- Enforce consistent use of `shared.IsFalseOrEmpty()` when checking `features.images` value for a project.

### DIFF
--- a/lxd/db/cluster/profiles.go
+++ b/lxd/db/cluster/profiles.go
@@ -5,6 +5,8 @@ package cluster
 import (
 	"context"
 	"database/sql"
+	"fmt"
+	"strings"
 
 	"github.com/canonical/lxd/shared/api"
 )
@@ -116,4 +118,69 @@ func GetProfilesIfEnabled(ctx context.Context, tx *sql.Tx, projectName string, n
 	}
 
 	return profiles, nil
+}
+
+// GetProfileIDsByProjectAndName finds profile IDs matching specific names within
+// a specific list of projects.
+func GetProfileIDsByProjectAndName(ctx context.Context, tx *sql.Tx, projectNames []string, profileNames []string) ([]int64, error) {
+	if len(projectNames) == 0 || len(profileNames) == 0 {
+		return []int64{}, nil
+	}
+
+	args := make([]any, 0, len(projectNames)+len(profileNames))
+	var stmt strings.Builder
+
+	stmt.WriteString("SELECT profiles.id FROM profiles ")
+	stmt.WriteString("JOIN projects ON profiles.project_id = projects.id ")
+	stmt.WriteString("WHERE projects.name IN (")
+
+	// Add project name placeholders.
+	for i, projectName := range projectNames {
+		if i > 0 {
+			stmt.WriteString(",")
+		}
+
+		stmt.WriteString("?")
+		args = append(args, projectName)
+	}
+
+	stmt.WriteString(") AND profiles.name IN (")
+
+	// Add profile name placeholders.
+	for i, profileName := range profileNames {
+		if i > 0 {
+			stmt.WriteString(",")
+		}
+
+		stmt.WriteString("?")
+		args = append(args, profileName)
+	}
+
+	stmt.WriteString(")")
+
+	rows, err := tx.QueryContext(ctx, stmt.String(), args...)
+	if err != nil {
+		return nil, fmt.Errorf("Failed fetching profile IDs: %w", err)
+	}
+
+	defer func() { _ = rows.Close() }()
+
+	profileIDs := make([]int64, 0)
+	for rows.Next() {
+		var id int64
+
+		err := rows.Scan(&id)
+		if err != nil {
+			return nil, fmt.Errorf("Failed scanning profile ID: %w", err)
+		}
+
+		profileIDs = append(profileIDs, id)
+	}
+
+	err = rows.Err()
+	if err != nil {
+		return nil, err
+	}
+
+	return profileIDs, nil
 }
--- a/lxd/db/images.go
+++ b/lxd/db/images.go
@@ -584,7 +584,9 @@ func (c *ClusterTx) UnsetImageCached(ctx context.Context, projectName string, fi
 }
 
 // UpdateImage updates the image with the given ID.
-func (c *ClusterTx) UpdateImage(ctx context.Context, id int, fname string, sz int64, public bool, autoUpdate bool, architecture string, createdAt time.Time, expiresAt time.Time, properties map[string]string, project string, profileIDs []int64) error {
+// If profileIDs parameter is nil, profile associations will not be updated for the image.
+// If profileIDs parameter is not nil, profile associations will be completely overwritten by the new values.
+func (c *ClusterTx) UpdateImage(ctx context.Context, id int, fname string, sz int64, public bool, autoUpdate bool, architecture string, createdAt time.Time, expiresAt time.Time, properties map[string]string, profileIDs []int64) error {
 	arch, err := osarch.ArchitectureId(architecture)
 	if err != nil {
 		arch = 0
@@ -623,27 +625,15 @@ func (c *ClusterTx) UpdateImage(ctx context.Context, id int, fname string, sz in
 		}
 	}
 
-	if project != "" && profileIDs != nil {
-		enabled, err := cluster.ProjectHasProfiles(ctx, c.tx, project)
+	if profileIDs != nil {
+		// Delete all profile associations from all projects for this image.
+		q := `DELETE FROM images_profiles WHERE image_id = ?`
+		_, err = c.tx.ExecContext(ctx, q, id)
 		if err != nil {
 			return err
 		}
 
-		if !enabled {
-			project = "default"
-		}
-
-		q := `DELETE FROM images_profiles
-				WHERE image_id = ? AND profile_id IN (
-					SELECT profiles.id FROM profiles
-					JOIN projects ON profiles.project_id = projects.id
-					WHERE projects.name = ?
-				)`
-		_, err = c.tx.ExecContext(ctx, q, id, project)
-		if err != nil {
-			return err
-		}
-
+		// Save new profile associations for this image.
 		sql = `INSERT INTO images_profiles (image_id, profile_id) VALUES (?, ?)`
 		for _, profileID := range profileIDs {
 			_, err = c.tx.ExecContext(ctx, sql, id, profileID)
--- a/lxd/images.go
+++ b/lxd/images.go
@@ -365,7 +365,7 @@ func compressFile(compress string, infile io.Reader, outfile io.Writer) error {
  * This function takes a container or snapshot from the local image server and
  * exports it as an image.
  */
-func imgPostInstanceInfo(s *state.State, req api.ImagesPost, op *operations.Operation, projectName string, builddir string, budget int64) (*api.Image, error) {
+func imgPostInstanceInfo(s *state.State, req api.ImagesPost, op *operations.Operation, instanceProject string, imageProject string, builddir string, budget int64) (*api.Image, error) {
 	info := api.Image{}
 	info.Properties = map[string]string{}
 
@@ -398,7 +398,7 @@ func imgPostInstanceInfo(s *state.State, req api.ImagesPost, op *operations.Oper
 		info.Public = false
 	}
 
-	c, err := instance.LoadByProjectAndName(s, projectName, name)
+	c, err := instance.LoadByProjectAndName(s, instanceProject, name)
 	if err != nil {
 		return nil, err
 	}
@@ -460,7 +460,7 @@ func imgPostInstanceInfo(s *state.State, req api.ImagesPost, op *operations.Oper
 	} else {
 		var p *api.Project
 		err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
-			project, err := dbCluster.GetProject(ctx, tx.Tx(), projectName)
+			project, err := dbCluster.GetProject(ctx, tx.Tx(), instanceProject)
 			if err != nil {
 				return err
 			}
@@ -551,7 +551,7 @@ func imgPostInstanceInfo(s *state.State, req api.ImagesPost, op *operations.Oper
 	info.CreatedAt = time.Now().UTC()
 
 	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
-		_, _, err = tx.GetImage(ctx, info.Fingerprint, dbCluster.ImageFilter{Project: &projectName})
+		_, _, err = tx.GetImage(ctx, info.Fingerprint, dbCluster.ImageFilter{Project: &imageProject})
 
 		return err
 	})
@@ -564,7 +564,7 @@ func imgPostInstanceInfo(s *state.State, req api.ImagesPost, op *operations.Oper
 	}
 
 	/* rename the file to the expected name so our caller can use it */
-	finalName := filepath.Join(s.ImagesStoragePath(projectName), info.Fingerprint)
+	finalName := filepath.Join(s.ImagesStoragePath(imageProject), info.Fingerprint)
 	err = shared.FileMove(imageFile.Name(), finalName)
 	if err != nil {
 		return nil, err
@@ -575,7 +575,7 @@ func imgPostInstanceInfo(s *state.State, req api.ImagesPost, op *operations.Oper
 
 	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
 		// Create the database entry
-		return tx.CreateImage(ctx, c.Project().Name, info.Fingerprint, info.Filename, info.Size, info.Public, info.AutoUpdate, info.Architecture, info.CreatedAt, info.ExpiresAt, info.Properties, info.Type, nil)
+		return tx.CreateImage(ctx, imageProject, info.Fingerprint, info.Filename, info.Size, info.Public, info.AutoUpdate, info.Architecture, info.CreatedAt, info.ExpiresAt, info.Properties, info.Type, nil)
 	})
 	if err != nil {
 		return nil, err
@@ -584,7 +584,7 @@ func imgPostInstanceInfo(s *state.State, req api.ImagesPost, op *operations.Oper
 	return &info, nil
 }
 
-func imgPostRemoteInfo(ctx context.Context, s *state.State, req api.ImagesPost, op *operations.Operation, project string, budget int64) (*api.Image, error) {
+func imgPostRemoteInfo(ctx context.Context, s *state.State, req api.ImagesPost, op *operations.Operation, profileProject string, imageProject string, budget int64) (*api.Image, error) {
 	var err error
 	var hash string
 
@@ -605,7 +605,7 @@ func imgPostRemoteInfo(ctx context.Context, s *state.State, req api.ImagesPost, 
 		Type:              req.Source.ImageType,
 		AutoUpdate:        req.AutoUpdate,
 		Public:            req.Public,
-		ProjectName:       project,
+		ProjectName:       imageProject,
 		Budget:            budget,
 		SourceProjectName: req.Source.Project,
 		UserRequested:     true,
@@ -617,7 +617,7 @@ func imgPostRemoteInfo(ctx context.Context, s *state.State, req api.ImagesPost, 
 	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
 		var id int
 
-		id, info, err = tx.GetImage(ctx, info.Fingerprint, dbCluster.ImageFilter{Project: &project})
+		id, info, err = tx.GetImage(ctx, info.Fingerprint, dbCluster.ImageFilter{Project: &imageProject})
 		if err != nil {
 			return err
 		}
@@ -633,7 +633,7 @@ func imgPostRemoteInfo(ctx context.Context, s *state.State, req api.ImagesPost, 
 		profileIDs := make([]int64, len(req.Profiles))
 
 		for i, profile := range req.Profiles {
-			profileID, _, err := tx.GetProfile(ctx, project, profile)
+			profileID, _, err := tx.GetProfile(ctx, profileProject, profile)
 			if response.IsNotFoundError(err) {
 				return fmt.Errorf("Profile %q doesn't exist", profile)
 			} else if err != nil {
@@ -643,9 +643,20 @@ func imgPostRemoteInfo(ctx context.Context, s *state.State, req api.ImagesPost, 
 			profileIDs[i] = profileID
 		}
 
+		// Handle the case when an image is being copied into the default project.
+		if imageProject == api.ProjectDefaultName {
+			// Find IDs of profiles that have matching names in projects with "features.images=false".
+			otherProfileIDs, err := resolveProfileIDs(ctx, tx, profileProject, req.Profiles)
+			if err != nil {
+				return err
+			}
+
+			profileIDs = append(profileIDs, otherProfileIDs...)
+		}
+
 		// Update the DB record if needed
 		if req.Public || req.AutoUpdate || req.Filename != "" || len(req.Properties) > 0 || len(req.Profiles) > 0 {
-			err := tx.UpdateImage(ctx, id, req.Filename, info.Size, req.Public, req.AutoUpdate, info.Architecture, info.CreatedAt, info.ExpiresAt, info.Properties, project, profileIDs)
+			err := tx.UpdateImage(ctx, id, req.Filename, info.Size, req.Public, req.AutoUpdate, info.Architecture, info.CreatedAt, info.ExpiresAt, info.Properties, imageProject, profileIDs)
 			if err != nil {
 				return err
 			}
@@ -660,7 +671,7 @@ func imgPostRemoteInfo(ctx context.Context, s *state.State, req api.ImagesPost, 
 	return info, nil
 }
 
-func imgPostURLInfo(ctx context.Context, s *state.State, req api.ImagesPost, op *operations.Operation, project string, budget int64) (*api.Image, error) {
+func imgPostURLInfo(ctx context.Context, s *state.State, req api.ImagesPost, op *operations.Operation, imageProject string, budget int64) (*api.Image, error) {
 	var err error
 
 	if req.Source.URL == "" {
@@ -714,7 +725,7 @@ func imgPostURLInfo(ctx context.Context, s *state.State, req api.ImagesPost, op 
 		Alias:         hash,
 		AutoUpdate:    req.AutoUpdate,
 		Public:        req.Public,
-		ProjectName:   project,
+		ProjectName:   imageProject,
 		Budget:        budget,
 		UserRequested: true,
 	})
@@ -725,7 +736,7 @@ func imgPostURLInfo(ctx context.Context, s *state.State, req api.ImagesPost, op 
 	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
 		var id int
 
-		id, info, err = tx.GetImage(ctx, info.Fingerprint, dbCluster.ImageFilter{Project: &project})
+		id, info, err = tx.GetImage(ctx, info.Fingerprint, dbCluster.ImageFilter{Project: &imageProject})
 		if err != nil {
 			return err
 		}
@@ -1164,10 +1175,16 @@ func imagesPost(d *Daemon, r *http.Request) response.Response {
 
 	// Load the project entry so we have a valid project name.
 	var dbProject *dbCluster.Project
+	var projectConfig map[string]string
 	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
 		dbProject, err = dbCluster.GetProject(ctx, tx.Tx(), projectName)
 		if err != nil {
 			return fmt.Errorf("Failed loading project %q: %w", projectName, err)
+		}
+
+		projectConfig, err = dbCluster.GetProjectConfig(ctx, tx.Tx(), projectName)
+		if err != nil {
+			return fmt.Errorf("Failed loading config for project %q: %w", projectName, err)
 		}
 
 		return nil
@@ -1325,17 +1342,33 @@ func imagesPost(d *Daemon, r *http.Request) response.Response {
 			/* Processing image upload */
 			info, err = getImgPostInfo(s, r, builddir, dbProject.Name, post, imageMetadata)
 		} else {
+			// Project to associate image with.
+			imageProject := dbProject.Name
+
+			// If "features.images" is disabled for the project, associate the image with the "default" project.
+			if shared.IsFalseOrEmpty(projectConfig["features.images"]) {
+				imageProject = api.ProjectDefaultName
+			}
+
+			// Project to associate profiles with.
+			profileProject := dbProject.Name
+
+			// If "features.profiles" is disabled for the project, associate the profiles with the "default" project.
+			if shared.IsFalseOrEmpty(projectConfig["features.profiles"]) {
+				profileProject = api.ProjectDefaultName
+			}
+
 			switch req.Source.Type {
 			case api.SourceTypeImage:
 				/* Processing image copy from remote */
-				info, err = imgPostRemoteInfo(ctx, s, req, op, dbProject.Name, budget)
+				info, err = imgPostRemoteInfo(ctx, s, req, op, profileProject, imageProject, budget)
 			case "url":
 				/* Processing image copy from URL */
-				info, err = imgPostURLInfo(ctx, s, req, op, dbProject.Name, budget)
+				info, err = imgPostURLInfo(ctx, s, req, op, imageProject, budget)
 			default:
 				/* Processing image creation from container */
 				imagePublishLock.Lock()
-				info, err = imgPostInstanceInfo(s, req, op, dbProject.Name, builddir, budget)
+				info, err = imgPostInstanceInfo(s, req, op, dbProject.Name, imageProject, builddir, budget)
 				imagePublishLock.Unlock()
 			}
 		}
@@ -3494,8 +3527,29 @@ func imagePut(d *Daemon, r *http.Request) response.Response {
 	profileIDs := make([]int64, len(req.Profiles))
 
 	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+		projectConfig, err := dbCluster.GetProjectConfig(ctx, tx.Tx(), projectName)
+		if err != nil {
+			return fmt.Errorf("Failed loading config for project %q: %w", projectName, err)
+		}
+
+		// Project to associate image with.
+		imageProject := projectName
+
+		// If "features.images" is disabled for the project, associate the image with the "default" project.
+		if shared.IsFalseOrEmpty(projectConfig["features.images"]) {
+			imageProject = api.ProjectDefaultName
+		}
+
+		// Project to associate profiles with.
+		profileProject := projectName
+
+		// If "features.profiles" is disabled for the project, associate the profiles with the "default" project.
+		if shared.IsFalseOrEmpty(projectConfig["features.profiles"]) {
+			profileProject = api.ProjectDefaultName
+		}
+
 		for i, profile := range req.Profiles {
-			profileID, _, err := tx.GetProfile(ctx, projectName, profile)
+			profileID, _, err := tx.GetProfile(ctx, profileProject, profile)
 			if response.IsNotFoundError(err) {
 				return fmt.Errorf("Profile %q doesn't exist", profile)
 			} else if err != nil {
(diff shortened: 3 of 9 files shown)


tomponline:
@nmezhenskyi please can you add a test for this, ta

nmezhenskyi:
> please can you add a test for this  

@tomponline Will do! 

There's just one more edge case to handle when it comes to explicitly copying images with `lxc image copy` to a project with `features.images=false`, so I'm working on that.

 I also noticed that we have `features.profiles` option for projects that behaves similarly by letting projects with `features.profiles=false` use profiles from the `default` project, so I'm checking whether this works correctly as well.

tomponline:
> > please can you add a test for this
> 
> @tomponline Will do!
> 
> There's just one more edge case to handle when it comes to explicitly copying images with `lxc image copy` to a project with `features.images=false`, so I'm working on that.
> 
> I also noticed that we have `features.profiles` option for projects that behaves similarly by letting projects with `features.profiles=false` use profiles from the `default` project, so I'm checking whether this works correctly as well.

Nice one!

lxd/db/images.go
```diff
@@ -644,7 +644,7 @@ func (c *ClusterTx) UpdateImage(ctx context.Context, id int, fname string, sz in
 			return err
 		}
 
-		sql = `INSERT INTO images_profiles (image_id, profile_id) VALUES (?, ?)`
+		sql = `INSERT OR IGNORE INTO images_profiles (image_id, profile_id) VALUES (?, ?)`
```
nmezhenskyi: This is needed because the profile could already be associated with the image by previously calling `CreateImage()` function.   

`CreateImage()` itself also handles this query using `INSERT OR IGNORE`:
https://github.com/canonical/lxd/blob/5e69be9583bf52f7803a2cc8f5ba8a26dcd2beda/lxd/db/images.go#L751-L768

lxd/images.go
```diff
@@ -1291,17 +1297,33 @@ func imagesPost(d *Daemon, r *http.Request) response.Response {
...
+				imageProject = api.ProjectDefaultName
+			}
+
+			// Project to associate profiles with.
+			profileProject := dbProject.Name
+
+			// If "features.profiles" is disabled for the project, associate the profiles with the "default" project.
+			if shared.IsFalse(projectConfig["features.profiles"]) {
+				profileProject = api.ProjectDefaultName
+			}
```
nmezhenskyi, line 1359: This part is not strictly necessary because `UpdateImage()` database layer function checks whether the project has `features.profiles` disabled on its own.  

I thought it is still a good idea to add this check here as well for consistency and to make it clear that both images and profiles can independently refer to the `default` project. This also makes it clear why we are using two different project parameters (`profileProject` and `imageProject`) in the `imgPostRemoteInfo()` function.

But I can remove this part if needed.

lxd/db/images.go
tomponline: arent the images_profiles rows deleted just above this?

tomponline: but it seems safe in general to do this.

lxd/images.go
tomponline: yeah i think clarity is a good idea here.

test/suites/image.sh
```diff
@@ -398,3 +398,76 @@ run_images_public() {
   lxc project delete foo
   lxc image delete "${fingerprint}"
 }
+
+test_image_cached() {
+  echo "==> Test that images are being cached consistently between projects."
+
+  echo "==> Create a new instance in the default project with a remote image that is not stored locally."
+  lxc init images:alpine/edge c1
```
tomponline: @simondeziel have we used a separate LXD server in the tests as an image remote source when testing caching behaviour previously, to avoid needing to tie the tests to an external image server?

test/suites/image.sh
```diff
@@ -398,3 +398,76 @@ run_images_public() {
...
+
+  local fingerprint
+  fingerprint="$(lxc config show c1 | sed -En 's/^.*volatile.base_image: (.*)$/\1/p')"
+
+  echo "==> Delete the instance."
+  lxc delete c1
+
+  echo "==> Check that the locally saved image is marked as cached."
+  lxc image info "${fingerprint}" | grep -xF "Cached: yes"
```
tomponline, line 492: Can you use `lxc image get-property` here?

test/suites/image.sh
```diff
@@ -398,3 +398,76 @@ run_images_public() {
...
 }
+
+test_image_cached() {
+  echo "==> Test that images are being cached consistently between projects."
+
+  echo "==> Create a new instance in the default project with a remote image that is not stored locally."
+  lxc init images:alpine/edge c1
+
+  local fingerprint
+  fingerprint="$(lxc config show c1 | sed -En 's/^.*volatile.base_image: (.*)$/\1/p')"
```
tomponline: can you use `lxc config get` here?

lxd/db/images.go
nmezhenskyi: > arent the images_profiles rows deleted just above this?

They are, but only for a single project associated with the image. There is an edge case where this does not work: when a project has `features.images=false`, but `features.profiles=true`. This means that the image is associated to both the `default` project's profiles and to the other custom project's profiles. The `DELETE` query will only delete the association with the `default` project's profiles because the function uses a single parameter both for image and profile project name.

Should I fix this query as well?

test/suites/image.sh
nmezhenskyi: > Can you use lxc image get-property here?

This doesn't seem to work. `Cached` value only shows up in `lxc image info`, and is not shown in `lxc image show` and `lxc image get-property <image> cached`.

test/suites/image.sh
nmezhenskyi: Updated

test/suites/image.sh
tomponline: ah ok, i think image properties are different other properties :(

lxd/db/images.go
tomponline: > Should I fix this query as well?

What do you think is best, leaving it as insert ignore?
should we do that instead of the other deletion?

nmezhenskyi: > What do you think is best, leaving it as insert ignore?

Hm, I'm still trying to understand the correct approach for updating image profiles in `UpdateImage()`.

Using `insert ignore` seems to work in this particular case, but I'm afraid it can introduce a subtle bug, especially noticeable when there are multiple projects with `features.profiles=true` and `features.images=false`. 

Let's say we have three projects: `default`, `project1`, `project2`.

`project1` and `project2` have `features.profiles=true` and `features.images=false`. This means they have their own profiles, but they share images with the `default` project.

All three projects have a single profile named `profile1`. The `default` project has image named `image1` and it is associated with `profile1`.

 This means that the image actually has profile associations like these in the database:

**images_profiles rows**:

| Image | Profile |
| -------- | ------- |
| default_project_image1 | default_project_profile1 |
| default_project_image1 | project1_profile1 |
| default_project_image1 | project2_profile1 |

Now, the problem occurs when we want to remove `profile1` from `image1`, the `delete` query in the `UpdateImage()` will only do it for the `default` project.

nmezhenskyi: Currently I think the best way to handle this in `UpdateImage()` is to delete all profile associations for the image and then insert what is provided in the `profileIDs` parameter. It would be the function caller responsibility to specify all required profile IDs for the image.

tomponline: Right makes sense, and then we'd expect the caller to figure out all the profile IDs across all projects that reference this image.

nmezhenskyi: I've updated the approach as described, but it seems to have broken other tests. I'm looking into this.

nmezhenskyi: Reproduced this locally

nmezhenskyi: Seems to be working now, I think I've fixed it

test/suites/image.sh
simondeziel: You mean a dedicated LXD image server on which we'd host our `testimage` and a copy of the `ubuntu-minimal-daily:24.04`?

Not that I know but it sounds like an interesting idea.

simondeziel (dismissed):
LGTM and even with the repeated image fetching, it still runs inside ~20s on `ceph` so not bad at all.

test/suites/image.sh
tomponline: No I mean using a temporary lxd instance as a remote to use the standard test image as s cached one. I'll find it in the existing tests.

lxd/images.go
```diff
@@ -5086,3 +5086,32 @@ func createTokenResponse(s *state.State, r *http.Request, projectName string, fi
...
+func getProfileIDsFromOtherProjects(ctx context.Context, tx *db.ClusterTx, currentProject string, profileNames []string) ([]int64, error) {
+	projects, err := dbCluster.GetProjectNames(ctx, tx.Tx())
+	if err != nil {
+		return nil, fmt.Errorf("Failed to load project names: %w", err)
+	}
+
+	profileIDs := make([]int64, 0)
+
+	for _, projectName := range projects {
+		if projectName == currentProject {
```
tomponline: out of interest, why do we exclude currentProject here rather than just including it in the returned profile IDs list?

test/suites/image.sh
tomponline: @nmezhenskyi take a look at `test_image_auto_update` and how we used the default lxd server as an image remote to auto update the built-in test image, rather than depending on an external image server.

lxd/images.go
nmezhenskyi: > out of interest, why do we exclude currentProject here rather than just including it in the returned profile IDs list?

Good point, now there's not much of a reason to do this. I originally did it to avoid duplicating current project IDs because I fetched other IDs conditionally (only when image's project was the `default` project) while the current project IDs were always fetched, but this was incorrect. Since now we always fetch all these IDs, there's no need to exclude `currentProject`, so I can probably just combine these two parts.

nmezhenskyi: After looking at it again, I see that there is still a use case for separating fetching profile IDs for the current project and other projects.

When we fetch profile IDs for the current project before calling `getProfileIDsFromOtherProjects()`, we also ensure that all requested profiles actually exist within the current project. Otherwise, we return an error specifying that a profile is missing.

However, this is not a requirement for fetching profile IDs from other projects because other projects are allowed to not have the same profiles.

So the reason we exclude `currentProject` inside of `getProfileIDsFromOtherProjects()` function is to avoid having duplicate IDs in the result list.

test/suites/image.sh
nmezhenskyi: Nice! I've updated the tests and it's working perfectly.

test/suites/image.sh
```diff
@@ -398,3 +398,93 @@ run_images_public() {
...
+  fi
+
+  echo "==> Set up a local LXD server as an image remote."
+  local LXD2_DIR LXD2_ADDR
+  LXD2_DIR=$(mktemp -d -p "${TEST_DIR}" XXX)
+  spawn_lxd "${LXD2_DIR}" true
+  LXD2_ADDR=$(< "${LXD2_DIR}/lxd.addr")
+
+  echo "==> Use busybox image for testing."
+  (LXD_DIR=${LXD2_DIR} deps/import-busybox --alias testimage --public)
```
simondeziel: Please drop the `()` as this shouldn't be needed. This was probably taken from `test_image_auto_update` which I still need to fix ;)

test/suites/image.sh
```diff
@@ -398,3 +398,93 @@ run_images_public() {
...
+  lxc image copy l2:testimage local:
+
+  echo "==> Check that the locally saved image is now marked as non-cached."
+  lxc image info "${fingerprint}" | grep -xF "Cached: no"
+
+  echo "==> Delete the local image."
+  lxc image delete "${fingerprint}"
+
+  echo "==> Create a new project p1. It has features.images=true by default."
+  defaultPoolName="$(lxc storage show "$(lxc profile device get default root pool)" | awk '/^name:/ {print $2}')"
```
simondeziel:
```suggestion
  defaultPoolName="$(lxc profile device get default root pool)"
```

test/suites/image.sh
nmezhenskyi: Updated :)

test/suites/image.sh
nmezhenskyi: Updated

simondeziel (dismissed):

test/suites/image.sh
```diff
@@ -398,3 +398,93 @@ run_images_public() {
...
+
+  echo "==> Delete the local image."
+  lxc image delete "${fingerprint}"
+
+  echo "==> Create a new project p1. It has features.images=true by default."
+  defaultPoolName="$(lxc profile device get default root pool)"
+  lxc project create --storage "${defaultPoolName}" p1
+
+  echo "==> Switch to project p1."
+  lxc project switch p1
```
tomponline: For tests I recommend using `--project` flag rather than persistent switching as its clearer when reviewing the tests which project each action is occurring in rather than having to remember the state.

test/suites/image.sh
tomponline, line 1: Can we test (if not already) whether the same image can be cached in one project but not cached in another project (when features.images=true)?

Also, related to the Cached property is the `expires_at` property, can you expand the tests to check that is correctly being set  too?

Thanks!

test/suites/image.sh
nmezhenskyi: Makes sense, updated

test/suites/image.sh
nmezhenskyi: > Can we test (if not already) whether the same image can be cached in one project but not cached in another project (when features.images=true)?

@tomponline Can you please expand on what is the expected behavior here?

Currently I see that if we do `lxc init images:alpine/edge c1` in the `default` project and then do the same for project `p1` with `features.images=true`, both projects will show that their image is cached (`lxc image info <image>` and `lxc image info <image> --project p1`). Is that wrong?

tomponline: That is correct. Both images were created in the project as a side effect of doing `lxc init`.

My question was for a different scenario:

1. lxc init images:alpine/edge c1 in the default project - expect cached image in default project.
2. lxc image copy images:alpine/edge local: in the p1 project - expect non-cached image in p1 project (and expect that same image in default project is still marked as cached).

nmezhenskyi: > My question was for a different scenario

Now I understand, thanks. I've added a check for this now.

nmezhenskyi: > Also, related to the Cached property is the expires_at property, can you expand the tests to check that is correctly being set too?

@tomponline While working on expanding the tests to check for the `expires_at`, I've noticed that we never set this property for images downloaded from `https://images.lxd.canonical.com` (it's always set to never expire), but we do set it for images downloaded from `https://cloud-images.ubuntu.com/minimal/daily`. Is this a bug?

nmezhenskyi: And similarly in the test suite, when we use a Busybox test image, its `expires_at` field is set to never expire regardless of whether the image is cached or not

tomponline: which part of the code is contributing to the difference?

nmezhenskyi: > which part of the code is contributing to the difference?

@tomponline After today's meeting I've confirmed that `expires_at` field is set by the remote server itself.

Here's the relevant section of code inside of the `ImageDownload` function:
 https://github.com/canonical/lxd/blob/936bd063af43f710f2c21ba8c9d26299a3b0b0ab/lxd/daemon_images.go#L113-L129
 
`info, _, err = remote.GetImage(fp)` returns the image metadata which includes the `expires_at` field. Then, we just use this value for the downloaded image.

nmezhenskyi: Also, regarding the `images.remote_cache_expiry` config option, we already have a test for it `test_image_expiry()`:

https://github.com/canonical/lxd/blob/936bd063af43f710f2c21ba8c9d26299a3b0b0ab/test/suites/image.sh#L71-L85

lxd/images.go
```diff
@@ -612,9 +612,17 @@ func imgPostRemoteInfo(ctx context.Context, s *state.State, req api.ImagesPost,
 			profileIDs[i] = profileID
 		}
 
+		// Find IDs of profiles that have matching names in other projects.
+		otherProfileIDs, err := getProfileIDsFromOtherProjects(ctx, tx, profileProject, req.Profiles)
```
tomponline: In our earlier conversation about moving profile ID list generation to the caller of `tx.UpdateImage()` I said:

> Right makes sense, and then we'd expect the caller to figure out all the profile IDs across all projects that reference this image [that match the requested profile names].

But looking at `getProfileIDsFromOtherProjects` now, it seems to be returning:

> all the profile IDs across all projects that [that match the requested profile names]

This doesn't seem quite right to me. Why would we associate an image with profile IDs belonging to projects that dont have the image?

tomponline: Also please can you check, and add if not, tests for checking this profile association behaviour. Thanks

nmezhenskyi: > This doesn't seem quite right to me. Why would we associate an image with profile IDs belonging to projects that dont have the image?

True, this is wrong now that I'm looking at it again. I got a bit misguided when trying to solve an edge case for `lxc image copy` between projects. I'll change it.

nmezhenskyi: > Right makes sense, and then we'd expect the caller to figure out all the profile IDs across all projects that reference this image [that match the requested profile names].

Initially I approached this by only fetching profiles of projects that have `features.images=false` and only when determined that the `current` project has `features.images=false` or if that's the `default` project. 

However, this approach broke `tls_restrictions` test, where it copies local images between projects. I'm investigating why this is happening.

nmezhenskyi: > However, this approach broke tls_restrictions tests, where it copies local images between projects. I'm investigating why this is happening.

I think I've found the issue. There's an edge case when copying an image from `default` project into another project with `features.images=false`. Logically, this should be a no-op because the image still remains in the `default` project, but current implementation doesn't handle this. As a result, this deletes all profile associations from the `default` project because of the updated `UpdateImage()` function implementation. I'll fix this.

Though, I don't understand why `tls_restrictions` is setting `features.images=false` on a project:
https://github.com/canonical/lxd/blob/59ef526735cda321e42f90cda992a1822a0d6ae3/test/suites/tls_restrictions.sh#L155-L156
and then tries to copy an image into that project from `default` project:
https://github.com/canonical/lxd/blob/59ef526735cda321e42f90cda992a1822a0d6ae3/test/suites/tls_restrictions.sh#L301-L305

But this helped me find the edge case :)

nmezhenskyi: > There's an edge case when copying an image from `default` project into another project with `features.images=false`. Logically, this should be a no-op because the image still remains in the `default` project, but current implementation doesn't handle this.

Or should it be an error? @tomponline

tomponline: > Though, I don't understand why tls_restrictions is setting features.images=false on a project:

@markylaing might know why?

tomponline: > Or should it be an error? @tomponline

What happens within the same project if you copy an image from a remote into a project that already has an existing image of the same fingerprint (Cached or not)?  My thinking is that it should do the same thing if copying into a different project with features.images=false if the default project has the same image existing.

lxd/images.go
```diff
@@ -5195,3 +5195,34 @@ func createTokenResponse(s *state.State, r *http.Request, projectName string, fi
...
+// It searches the "default" project and all projects where "features.images" is false.
+func resolveProfileIDs(ctx context.Context, tx *db.ClusterTx, currentProject string, targetNames []string) ([]int64, error) {
+	// Get the "default" project and all projects with "features.images=false".
+	candidateProjects, err := dbCluster.GetProjectsSharingDefaultImages(ctx, tx.Tx())
+	if err != nil {
+		return nil, fmt.Errorf("Failed to load project names: %w", err)
+	}
+
+	// Filter out the current project to avoid SQL constraint violation when adding duplicate profile IDs.
+	searchableProjects := make([]string, 0, len(candidateProjects))
```
nmezhenskyi, line 5274: I could remove this part completely by changing the `INSERT` query for the `images_profiles` table to use `INSERT OR IGNORE`.

lxd/db/cluster/profiles.go
```diff
@@ -117,3 +119,68 @@ func GetProfilesIfEnabled(ctx context.Context, tx *sql.Tx, projectName string, n
...
+
+		stmt.WriteString("?")
+		args = append(args, profileName)
+	}
+
+	stmt.WriteString(")")
+
+	rows, err := tx.QueryContext(ctx, stmt.String(), args...)
+	if err != nil {
+		return nil, fmt.Errorf("Failed fetching profiles IDs: %w", err)
```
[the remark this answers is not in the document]
nmezhenskyi: Fixed

lxd/images.go
```diff
@@ -3494,6 +3527,27 @@ func imagePut(d *Daemon, r *http.Request) response.Response {
...
+		// Project to associate profiles with.
+		profileProject := projectName
+
+		// If "features.profiles" is disabled for the project, associate the profiles with the "default" project.
+		if shared.IsFalseOrEmpty(projectConfig["features.profiles"]) {
+			profileProject = api.ProjectDefaultName
+		}
+
 		for i, profile := range req.Profiles {
 			profileID, _, err := tx.GetProfile(ctx, projectName, profile)
```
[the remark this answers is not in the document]
nmezhenskyi: Fixed

nmezhenskyi:
@tomponline This should be ready for review now.

lxd/images.go
tomponline: Sounds good to me.

tomponline: But do you think it might have other side effects we dont want?

tomponline (approved):
Lovely stuff thanks!

PR #16692, opened 2025-10-07
Instance: Refactor start operation for LXC driver to wait for stop on failure (WIP)

nmezhenskyi:
Fixes https://github.com/canonical/lxd/issues/16182.

## Changes:
Refactor `Start()` function for LXC driver to wait for stop on failure. Also, if post start hooks are not run, use reverter to clean up instance devices.


nmezhenskyi:
Hm, this change breaks device setup after forced restart for containers. I'm looking into this

nmezhenskyi:
Oh, I see why.

This section of code where we clear the revert function inside the `Start()` does not handle the case where `op.Action() == "restart"`. As a result, we remove all the devices from the container even on successful restart.

```go
if op.Action() == "start" {
		revert.Success() // Instance started successfully.

		d.logger.Info("Started instance", ctxMap)
		d.state.Events.SendLifecycle(d.project.Name, lifecycle.InstanceStarted.Event(d, nil))
}
```

nmezhenskyi:
Seems to be working correctly now.

lxd/instance/drivers/driver_lxc.go
```diff
@@ -2393,6 +2400,11 @@ func (d *lxc) Start(stateful bool) error {
 
 		d.logger.Error("Failed starting instance", ctxMap)
 
+		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+		defer cancel()
+
+		_ = op.Wait(ctx)
```
tomponline: @nmezhenskyi we should log a warning if the error returned here is non-nil.

lxd/instance/drivers/driver_lxc.go
```diff
@@ -2393,6 +2400,11 @@ func (d *lxc) Start(stateful bool) error {
 
 		d.logger.Error("Failed starting instance", ctxMap)
 
+		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
tomponline, line 2401: We should add a comment here explaining what we are waiting for any why it may not happen.

lxd/instance/drivers/driver_lxc.go
```diff
@@ -2410,8 +2422,12 @@ func (d *lxc) Start(stateful bool) error {
 	}
 
 	if op.Action() == "start" {
+		revert.Success() // Instance started successfully.
```
tomponline: can we just move this 2 lines up and avoid the duplication below?

lxd/instance/drivers/driver_lxc.go
nmezhenskyi: Updated

lxd/instance/drivers/driver_lxc.go
nmezhenskyi: Updated

lxd/instance/drivers/driver_lxc.go
nmezhenskyi: Updated

lxd/instance/drivers/driver_lxc.go
```diff
@@ -2393,6 +2400,18 @@ func (d *lxc) Start(stateful bool) error {
 
 		d.logger.Error("Failed starting instance", ctxMap)
 
+		// Use this context timeout to wait for potential stop operation to complete in case instance start fails.
+		// If run, stop operation will inherit the current operation lock.
+		// In case stop operation never runs, op.Wait() will return immediately.
```
tomponline: this line seems to be in conflict with the next line

lxd/instance/drivers/driver_lxc.go
```diff
@@ -2346,13 +2347,19 @@ func (d *lxc) Start(stateful bool) error {
...
 	// Run the shared start code.
-	configPath, postStartHooks, err := d.startCommon()
+	cleanup, configPath, postStartHooks, err := d.startCommon()
 	if err != nil {
 		op.Done(err)
 		return err
 	}
 
+	// Use reverter to clean up instance devices on failure.
+	revert.Add(cleanup)
```
tomponline: if the stop hooks have run they will have cleaned up the devices and unmounted, so dont think we need to do it here too. 

We somehow should differentiate whether the stop hooks have run or not and only run the cleanup if they have not.

lxd/instance/drivers/driver_lxc.go
nmezhenskyi: > this line seems to be in conflict with the next line

Maybe I worded it poorly, but what I meant with the next line (`Stop operation may never run, if the forkstart fails early`) is to provide the reason for why Stop operation might not run.

nmezhenskyi: Essentially, the stop hooks will not run unless `forkstart` calls `go-lxc.Start()` on the container and it fails. Does this make sense? @tomponline

lxd/instance/drivers/driver_lxc.go
nmezhenskyi: > We somehow should differentiate whether the stop hooks have run or not and only run the cleanup if they have not.

True. 

If `stopErr := op.Wait(ctx)` failed with an error, then either stop hooks have not run (and deadline timed out), have failed, or are taking a very long time. We could use this error as a sign to use the reverter and otherwise skip it. 

Or alternatively, we could perhaps communicate whether the stop hooks have run using channels.

What do you think? @tomponline

lxd/instance/drivers/driver_lxc.go
nmezhenskyi: Ohh, I see now. The issue is with this line you've highlighted actually. Yes, it is incorrect.

tomponline: thanks!

lxd/instance/drivers/driver_lxc.go
tomponline: Yeah i think detecting if the `stopErr` is context cancelled and only then trigger the cleanup makes sense.

nmezhenskyi: Updated

lxd/instance/drivers/driver_lxc.go
```diff
@@ -2393,6 +2397,23 @@ func (d *lxc) Start(stateful bool) error {
...
+		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+		defer cancel()
+
+		stopErr := op.Wait(ctx)
+		if stopErr != nil {
+			d.logger.Warn("Failed running instance stop hooks after failed start", ctxMap, logger.Ctx{"err": stopErr})
+
+			if errors.Is(stopErr, context.DeadlineExceeded) {
+				// Use reverter to clean up instance devices because stop hooks likely have not run.
+				revert.Add(cleanup)
```
tomponline: Thanks this is looking great.

One last thing, I don't think we need a reverter at all in this function, because we just need to call `cleanup()` here.

The only other place that the reverter can fun before `revert.Success()` is called below is after an error returned from `d.runHooks(postStartHooks)`, and we can see that in that scenario the container is stopped via `d.Stop()` which will in turn trigger the device cleanup normally.

nmezhenskyi: Oh yes, indeed

nmezhenskyi: Updated

lxd/main_forkstart.go
```diff
@@ -91,5 +92,11 @@ func (c *cmdForkstart) run(cmd *cobra.Command, args []string) error {
 		_ = unix.Dup3(int(logFile.Fd()), 2, 0)
 	}
 
-	return d.Start()
+	err = d.Start()
+	if err != nil {
+		// Wait for container to be stopped if the start failed.
```
tomponline, line 97: @nmezhenskyi one last nit, maybe this comment would benefit from being expanded to explain that the status goes through "stopping" whilst some hooks are being run. Otherwise it seems weird to wait for stopped after start failed.

PR #16704, opened 2025-10-09
Refined caching

simondeziel:
Improve how some assets are cached.
Improve the `install_snap` helper to verify any installed snap is from the right channel.
Use `snapd`'s `latest/candidate` like done in `lxd-ci`. This costs us ~50MiB per cache bundle.


simondeziel:
I was thinking that `core24` might be best pulled from `latest/candidate`, what do you think @tomponline?

PR #16708, opened 2025-10-10
VM: Persistent PCIe bus numbers

tomponline:
Reworks PCIe bus management using a per-device `volatile.<device_name>.bus` key to store the bus order number.

e.g.

```
volatile.eth0.bus=4
```

This allows to persistently record the PCIe bus layout for a VM in order to be able to restore it at reboot.
This was a device hotplugged whilst the VM is running will get the same bus number on reboot.

Moving to DB based bus management also resolves the issues we were having with QEMU's QMP `query-pci` command not showing timely information about which slots were being used. 

Instead we now only use `query-pci` to ascertain the number of root ports available (which doesn't change per boot) and the accounting of slot usage is done in the DB.

Related to https://github.com/canonical/lxd/issues/16003

- [x] Handle upgrades with running VMs (added `volatile.bus.mode=persistent` to detect if a running VM supports dynamic allocation mode).
- [ ] Handle reverting volatile config keys if device attach fails.

### DIFF
--- a/test/suites/vm.sh
+++ b/test/suites/vm.sh
@@ -44,7 +44,53 @@ test_vm_empty() {
   lxc stop -f v1
   lxc delete v1
 
+  echo "==> Percentage memory limits"
   lxc launch --vm --empty v1 -c limits.memory=1% -d "${SMALL_ROOT_DISK}"
+
+  echo "==> Device PCIe bus numbers"
+  # Check initial boot bus allocation.
+  [ "$(lxc config get v1 volatile.eth0.bus)" = "4" ]
+
+  # Hotplug device and check bus allocation.
+  lxc config device add v1 aaa0 nic nictype=p2p
+  [ "$(lxc config get v1 volatile.aaa0.bus)" = "5" ]
+
+  # Check that the NIC devices are maintained when restarted.
+  lxc restart -f v1
+  [ "$(lxc config get v1 volatile.aaa0.bus)" = "5" ]
+  [ "$(lxc config get v1 volatile.eth0.bus)" = "4" ]
+
+  lxc stop -f v1
+
+  # Make the root disk device to consume a PCIe slot and check bus allocation.
+  lxc config device set v1 root io.bus=virtio-blk
+  lxc start v1
+  [ "$(lxc config get v1 volatile.eth0.bus)" = "4" ]
+  [ "$(lxc config get v1 volatile.aaa0.bus)" = "5" ]
+  [ "$(lxc config get v1 volatile.root.bus)" = "6" ]
+
+  # Check hotplugging block volume.
+  poolName=$(lxc profile device get default root pool)
+  poolDriver=$(lxc storage show "$(lxc profile device get default root pool)" | awk '/^driver:/ {print $2}')
+
+  if [ "$(lxc config get --expanded v1 migration.stateful || echo fail)" = "" ] || [ "${poolDriver}" = "ceph" ]; then
+    # Check using PCIe based virtio-blk when using shared storage or migration.stateful disabled.
+    lxc storage volume create "${poolName}" v1block --type=block size=1MiB
+    lxc config device add v1 v1block disk source=v1block pool="${poolName}"
+    [ "$(lxc config get v1 volatile.eth0.bus)" = "4" ]
+    [ "$(lxc config get v1 volatile.aaa0.bus)" = "5" ]
+    [ "$(lxc config get v1 volatile.root.bus)" = "6" ]
+    [ "$(lxc config get v1 volatile.v1block.bus || echo fail)" = "" ] # Uses virtio-scsi by default so no PCIe bus number
+    lxc config device set v1 v1block io.bus=virtio-blk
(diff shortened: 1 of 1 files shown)


lxd/instance/drivers/driver_qemu.go
```diff
@@ -3566,16 +3681,58 @@ func (d *qemu) generateQemuConfigFile(cpuInfo *cpuTopology, mountInfo *storagePo
...
+		} else if a.BusNum > 0 && b.BusNum == 0 {
+			return -1
+		}
+
+		return cmp.Compare(a.BusNum, b.BusNum)
+	})
+
+	// Number of spare hotplug ports to allocate.
+	// Could go negative by the time its used (below) if there are gaps in the bus numbers.
+	spareHotplugPorts := 8
```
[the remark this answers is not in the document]
tomponline: Nonsense. 

> If n <= 0, the loop does not run any iterations. 

https://go.dev/ref/spec

lxd/instance/drivers/driver_qemu.go
```diff
@@ -3566,16 +3681,58 @@ func (d *qemu) generateQemuConfigFile(cpuInfo *cpuTopology, mountInfo *storagePo
 		busName, busAddress, multifunction = bus.allocate(multifunctionGroup)
 		if busName != "" {
 			d.logger.Debug("Using bus to plug device into", logger.Ctx{"device": deviceName, "busType": bus.name, "bus": busName})
+
+			if bus.name == "pcie" {
+				if lastBusName != busName {
+					lastBusName = busName
+					lastBusNum++ // Increment bus number when bus name changes.
```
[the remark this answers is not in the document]
tomponline: This isn't true.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -3557,6 +3662,16 @@ func (d *qemu) generateQemuConfigFile(cpuInfo *cpuTopology, mountInfo *storagePo
 	}
 
 	// Setup a bus allocator for use with generating QEMU pre-boot config file.
+	volatileSet := make(map[string]string)
+	if d.localConfig["volatile.bus.mode"] == "" {
+		volatileSet["volatile.bus.mode"] = "persistent"
+	}
```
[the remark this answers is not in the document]
tomponline: its just before its needed

lxd/instance/drivers/driver_qemu.go
```diff
@@ -2216,8 +2229,100 @@ func (d *qemu) deviceStart(dev device.Device, instanceRunning bool) (*deviceConf
...
+		}
+
+		busNum := strconv.FormatUint(uint64(i), 10)
+		err = d.VolatileSet(map[string]string{deviceVolatileKey: busNum})
+		if err != nil {
+			return "", "", false, fmt.Errorf("Failed setting volatile keys: %w", err)
+		}
+
+		busName = busDevicePortPrefix + busNum
+		d.logger.Debug("Using bus to hotplug device into", logger.Ctx{"device": deviceName, "busType": "pcie", "bus": busName})
```
MusicDin: Small nit, but maybe `Hotplugging device onto bus` or `Found bus to hotplug device into` would be clearer. In case of `Using bus to hotplug device into` I would expect target where device is being plugged.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -2231,10 +2336,10 @@ func (d *qemu) busAllocatePCIeHotplug(deviceName string, _ bool) (busName string
 		pciDevID++
 	}
 
-	busName = busDevicePortPrefix + strconv.Itoa(pciDevID)
+	busName = busDevicePortPrefix + strconv.FormatUint(uint64(pciDevID), 10)
 	busAddress = "00.0" // First function on the bus.
 
-	d.logger.Debug("Using bus to hotplug device into", logger.Ctx{"device": deviceName, "busType": "pcie", "bus": busName})
+	d.logger.Debug("Using bus to hotplug device into (legacy)", logger.Ctx{"device": deviceName, "busType": "pcie", "bus": busName})
```
MusicDin: Same as above

lxd/instance/instancetype/instance.go
```diff
@@ -1194,6 +1194,13 @@ var InstanceConfigKeysVM = map[string]func(value string) error{
 	//  shortdesc: Whether to regenerate VM NVRAM the next time the instance starts
 	"volatile.apply_nvram": validate.Optional(validate.IsBool),
 
+	// lxdmeta:generate(entities=instance; group=volatile; key=volatile.bus.mode)
+	// Set to `persistent` when persistent bus allocation mode is enabled.
```
MusicDin, line 1198: What is the default value? Could we use stricter validation allowing empty or `persistent`? Or are there other allowed values?

tomponline: ill add improved validation.

tomponline:
@MusicDin ready for re-review.

MusicDin (approved):
Thanks 🚀

PR #16713, opened 2025-10-12
Misc test tweaks

simondeziel:
The `test-shell` feature now avoids spawning an extra daemon when invoking `main.sh` and also avoids reporting `bogus` setup error.

It also prepare things for `snap` based tests involving bigger VMs.

### DIFF
--- a/test/suites/basic.sh
+++ b/test/suites/basic.sh
@@ -247,17 +247,11 @@ test_basic_usage() {
   lxc profile delete priv
 
   # Test that containers without metadata.yaml are published successfully.
-  # Note that this quick hack won't work for LVM, since it doesn't always mount
-  # the container's filesystem. That's ok though: the logic we're trying to
-  # test here is independent of storage backend, so running it for just one
-  # backend (or all non-lvm backends) is enough.
-  if [ "$lxd_backend" = "lvm" ]; then
-    lxc init testimage nometadata
-    rm -f "${LXD_DIR}/containers/nometadata/metadata.yaml"
-    lxc publish nometadata --alias=nometadata-image
-    lxc image delete nometadata-image
-    lxc delete nometadata
-  fi
+  lxc init testimage nometadata
+  rm -f "${LXD_DIR}/containers/nometadata/metadata.yaml"
+  lxc publish nometadata --alias=nometadata-image
+  lxc image delete nometadata-image
+  lxc delete nometadata
 
   # Test public images
   lxc publish --public bar --alias=bar-image2


test/suites/vm.sh
```diff
@@ -1,9 +1,4 @@
 test_vm_empty() {
```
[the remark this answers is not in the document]
simondeziel: No, `main.sh` now checks for this.

test/backends/btrfs.sh
tomponline, line 1: why?

simondeziel: This is in preparation for `snap` based testing which involves bigger VMs (per the commit subject).

simondeziel:
> Can we use a constant for the 24MiB and 3GiB settings?

The `volume.size` is a parameter but could definitely be turned into `DEFAULT_VOLUME_SIZE` and used with drivers that support it. Same for the `size` which could be turned into `DEFAULT_POOL_SIZE` and used where applicable.

Is that what you asked?

tomponline:
> > Can we use a constant for the 24MiB and 3GiB settings?
> 
> The `volume.size` is a parameter but could definitely be turned into `DEFAULT_VOLUME_SIZE` and used with drivers that support it. Same for the `size` which could be turned into `DEFAULT_POOL_SIZE` and used where applicable.
> 
> Is that what you asked?

sounds good, i just saw a lot of duplication of those values.

simondeziel:
@tomponline ready for you to have another look, thanks

lxc/profile.go
```diff
@@ -732,6 +732,7 @@ Default column layout is: ndu
 Column shorthand chars:
 n - Profile Name
 d - Description
+e - Project
```
tomponline, line 735: @simondeziel in the future please separate functional non-test changes out of "misc tweaks" type PRs so they more visible. Thanks

test/includes/lxd.sh
```diff
@@ -142,10 +142,13 @@ kill_lxd() {
...
         # Delete all networks
-        echo "==> Deleting all managed networks"
-        for network in $(timeout -k 2 2 lxc network list --force-local --format csv | awk -F, '{if ($3 == "YES") {print $1}}'); do
-            timeout -k 10 10 lxc network delete "${network}" --force-local || true
-        done
+        echo "==> Deleting all managed networks from all projects"
+        while IFS=, read -r project network _ managed _; do
+            # Only delete managed networks.
+            [ "${managed}" != "YES" ] && continue
+            echo "   ⛔ Deleting network ${network} from project ${project}"
```
tomponline, line 149: are these symbols going to render in github test output ok?

lxc/profile.go
simondeziel: Indeed, noted.

test/includes/lxd.sh
simondeziel: UTF-8 seems to be widely supported but we won't seem them on GH because the `cleanup` is skipped on ephemeral runners.

simondeziel:
> Please can I ask you in the future to reduce the size of these type of PRs by splitting them into more focussed "themed" PRs.

I'll try to produce more focused PRs, it surely will help with reviewing.
 
> In general we don't want to see "misc ..." type PRs.

I must admit I often find an urge to fix anything amiss in the vicinity of what I'm looking at/fixing.

Thanks

PR #16735, opened 2025-10-14
github: free up disk space for the `standalone` test group

simondeziel:
This should reduce the likelyhood of running out of space with GH action runners.


simondeziel:
While reclaiming the disk space might have worked well, I'd rather avoid the ~1.5-2 minutes of runtime cost and instead specify a smaller pool for the problematic test(s) (see #16736).

PR #16738, opened 2025-10-14
Add `snap` test group

simondeziel:
All test functions with the name prefix `test_snap_` will have the LXD `latest/edge` snap installed and the `lxd`, `lxc` and `lxd-agent` binaries sideloaded onto it for the execution of the test function.

It is possible to mix regular (`test_*`) and snap-based (`test_snap_*`) functions in a given run, but it's best to group them to minimize the overhead of installing, sideloading, purging the LXD snap.

### DIFF
--- a/.github/actions/download-snaps/action.yml
+++ b/.github/actions/download-snaps/action.yml
@@ -66,5 +66,6 @@ runs:
         # Seed the snap dependencies cache
         download_snap snapd "latest/candidate"
         download_snap core24 "latest/stable"
+        download_snap lxd
         download_snap microceph
         download_snap microovn


test/suites/console.sh
```diff
@@ -51,3 +51,19 @@ EOF
...
   lxc delete cons1
 }
+
+test_snap_console_vm() {
+  lxc launch ubuntu-minimal-daily:24.04 v1 --vm -c limits.memory=384MiB -d "${SMALL_VM_ROOT_DISK}"
+  waitInstanceReady v1
+
+  # The VGA console is available for VMs
+  echo "===> Check VGA console address"
+  OUTPUT="$(timeout --foreground --signal KILL 0.1 lxc console --type vga v1 || true)"
```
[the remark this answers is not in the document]
simondeziel: 0.1 seconds is an eternity for a computer. Let's hope our CI runners are not overloaded to the point where they cannot generate that URL during those 100ms.

test/includes/lxd.sh
```diff
@@ -267,7 +267,7 @@ kill_lxd() {
         sleep 2
 
         # Cleanup devlxd and shmounts (needed due to the forceful kill)
-        find "${LXD_DIR}" \( -name devlxd -o -name shmounts \) -exec "umount" "-q" "-l" "{}" + || true
+        find "${LXD_DIR}" -type d -xdev \( -name devlxd -o -name shmounts \) -exec "umount" "-q" "-l" "{}" + || true
```
kadinsayani, line 272: Do we need to unlink in `kill_lxd_snap`?

kadinsayani: Perhaps a comment would be helpful here.

test/includes/snap.sh
```diff
@@ -106,3 +106,17 @@ install_snap() {
         snap switch "${name}" --channel="${channel}"
     )
 }
+
+# sideload_lxd_snap: installs the lxd snap and sideloads the lxc, lxd and lxd-agent binaries.
```
kadinsayani, line 110: Nice! In the future, maybe we can add a new `make` target, `test-snap`, that spawns a test shell and sideloads the lxc, lxd, and lxd-agent binaries?

test/suites/vm.sh
```diff
@@ -65,3 +65,48 @@ test_vm_empty() {
...
+  lxc stop -f v1
+  lxc delete v1
+
+  lxc launch --vm --empty v1 -c limits.memory=1% -d "${SMALL_ROOT_DISK}"
+  lxc delete --force v1
+
+  echo "==> Ephemeral cleanup"
+  lxc launch --vm --empty --ephemeral v1 -c limits.memory=128MiB -d "${SMALL_ROOT_DISK}"
+  lxc stop -f v1
+  [ "$(lxc list -f csv -c n || echo fail)" = "" ]
```
kadinsayani: Since these checks are similar to what we have in `test_vm_empty()`, perhaps we can extract a `vm_empty_common()` function for use in both `test_snap_vm_empty()` and `test_vm_empty()`?

test/main.sh
```diff
@@ -291,6 +291,16 @@ run_test() {
   fi
 
   if [ "${skip}" = false ]; then
+
```
kadinsayani, line 294: Instead of installing the snap and removing it between interleaved checks, perhaps we can sort arguments and only add the snap overhead once. For example, `./main.sh server_info snap_lxd_user metrics snap_basic clustering_enable` can be parsed so we run the `snap_*` checks first (with one install and cleanup) and the non-snap tests after.

test/includes/snap.sh
simondeziel: That's an idea worth exploring. I'll give it a think once I'm done rewriting the `main.sh` to be more structured. Ultimately, it should be easier to decide which kind of initial LXD daemon you want to interact with but for now, I've settled with the regular (non-snap) that we are used to.

test/main.sh
simondeziel: I also thought of that but then I found myself wanting to test the cleanup between mixed tests. Ultimately, I refrain from implementing it as it only affects folks running tests manually and I didn't want to remove control from the operator.

test/suites/vm.sh
simondeziel: True, I tried with the `console` one but had to walk back because containers and VMs behaves very differently.

For this one, I think I might be able to cut the duplication down, thanks for prodding me.

test/includes/lxd.sh
simondeziel: The snap is different in that it's mounted in the daemon NS so I decided to simply "walk over" those oddities.

test/main.sh
kadinsayani: Ack, we should leave control with the operator.

test/includes/lxd.sh
simondeziel: Comment added.

test/suites/vm.sh
simondeziel: Worked beautifully, thanks again!

kadinsayani (approved):
LGTM, thanks!

test/includes/lxd.sh
```diff
@@ -1,5 +1,71 @@
...
+    unset -f lxc lxc_remote
+
+    # shellcheck disable=SC2153
+    local lxd_backend="${LXD_BACKEND}"
+    if [ "$LXD_BACKEND" = "random" ]; then
+        lxd_backend="$(random_storage_backend)"
+    fi
+
+    if [ "${lxd_backend}" = "ceph" ] && [ -z "${LXD_CEPH_CLUSTER:-}" ]; then
+        echo "A cluster name must be specified when using the CEPH driver." >&2
```
tomponline, line 24: why is this? You can use ceph with a non-clustered lxd?

test/includes/lxd.sh
```diff
@@ -1,5 +1,71 @@
...
+    echo "==> Started lxd snap (PID is ${LXD_PID})"
+
+    echo "==> Binding to network"
+    lxc config set core.https_address "127.0.0.1:8443"
+
+    if [ -n "${SHELL_TRACING:-}" ]; then
+        set -x
+    fi
+
+    echo "==> Setting up networking"
```
tomponline, line 62: can these 2 steps be done in a common place for snap and non-snap?

test/suites/vm.sh
```diff
@@ -65,3 +69,48 @@ test_vm_empty() {
   ! lxc start v1 || false
   lxc delete v1
 }
+
+test_snap_vm_empty() {
```
tomponline: this doesnt need to be a specific snap test does it, it seems to be this would be preferable as a non-snap test?

test/includes/lxd.sh
simondeziel: It is referring to the `ceph` cluster name to create OSDs in. It's also included in the regular non-snap function which is why I included it there as well.

test/suites/vm.sh
simondeziel: Indeed it doesn't need to be snap specific and it isn't either. The bulk of the test is run with regular and with snap except for the gymnastic around 9p/`migration.stateful` limitation.

test/includes/lxd.sh
simondeziel: By 2 steps, you are referring to the storage backend and the network configuration? If so, only the network bit is common as for the storage, since snap tests involve fat Ubuntu VMs, they need to configure a bigger `volume.size`.

tomponline: I see. ta

test/suites/vm.sh
tomponline: I'm still confused why there appears to be validation tests (wrt to instance names, memory size, snapshot names) in `test_snap_vm_empty` which shouldnt be in there if we are just using snap tests for functional/integration tests, rather than testing business logic.

simondeziel: You are right, that's duplicating something already tested for the most part. Shall I drop it?

tomponline: yes please, we should aim to keep snap testing to just functional stuff we cant test in the normal VM tests.

simondeziel:
@tomponline changes made and CI green.

test/suites/basic.sh
```diff
@@ -755,6 +755,64 @@ EOF
...
+  [ "$(lxc list -f csv -c s)" = "FROZEN" ]
+  ! lxc stop v1 || false
+  [ "$(lxc list -f csv -c s)" = "FROZEN" ]
+  lxc start v1
+  [ "$(lxc list -f csv -c s)" = "RUNNING" ]
+  lxc stop -f v1
+  [ "$(lxc list -f csv -c s)" = "STOPPED" ]
+
+  echo "==> VM with snapshots"
+  lxc snapshot v1
```
tomponline: These name validation checks dont need to be in the snap test as not snap specific. Please can they be in the basic non snap VM test?

test/suites/basic.sh
```diff
@@ -755,6 +755,64 @@ EOF
...
+  ! lxc snapshot v1 "\\\\" || false
+  ! lxc snapshot v1 "/" || false
+  [ "$(lxc list -f csv -c S v1)" = "1" ]
+  lxc start v1
+  lxc snapshot v1
+  [ "$(lxc list -f csv -c S v1)" = "2" ]
+  lxc delete --force v1
+
+  # Test randomly named VM creation
+  RDNAME="$(lxc init --vm --empty --quiet -c limits.memory=384MiB -d "${SMALL_VM_ROOT_DISK}" | sed 's/Instance name is: //')"
```
tomponline: This isn't snap specific right?

test/suites/basic.sh
```diff
@@ -755,6 +755,64 @@ EOF
...
+  lxc delete "${RDNAME}"
+
+  echo "==> Create a VM suitable for stateful stop/start"
+  lxc launch ubuntu-minimal-daily:24.04 v1 --vm -c migration.stateful=true -c limits.memory=384MiB -d root,size.state=384MiB -d "${SMALL_VM_ROOT_DISK}"
+  waitInstanceReady v1
+
+  # XXX: for the dir driver, we can also check the existence of the state file
```
tomponline: This comment is defunct right?

test/suites/devlxd.sh
```diff
@@ -431,3 +431,71 @@ EOF
       lxc project delete "${project}"
   fi
 }
+
+test_snap_devlxd_vm() {
```
tomponline, line 435: Can we make this non snap now we have the cached Ubuntu image available to the non snap tests? They don't seem snap specific.

test/suites/basic.sh
simondeziel: Correct, I've stashed to add to `test_vm_empty` in another PR.

test/suites/basic.sh
simondeziel: Indeed, I've picked the few improvements and will fold them in the existing test around that in another PR.

test/suites/devlxd.sh
simondeziel: Unfortunately for non snap tests, we can only use `--empty` VMs with `migration.stateful=true`. The Ubuntu images for non snap will require the compiling `virtiofsd` which is not ready yet.

tomponline: do they not support 9p?

test/suites/basic.sh
```diff
@@ -755,6 +755,47 @@ EOF
...
+  lxc launch ubuntu-minimal-daily:24.04 v1 --vm -c limits.memory=384MiB -d "${SMALL_VM_ROOT_DISK}"
+  waitInstanceReady v1
+  lxc list --fast
+
+  echo "==> Check exec operations"
+  lxc exec v1 -- true
+  ! lxc exec v1 -- false || false
+  [ "$(lxc exec v1 -- hostname)" = "v1" ]
+
+  echo "==> Check VM state transitions"
```
tomponline, line 768: For the follow up PR,these dont look snap specific.

test/suites/devlxd.sh
simondeziel: Oh OK I misunderstood your request. So yeah I am sure they would support it but the bits to cache the `ubuntu-minimal-daily:24.04` image is something I'll implement next week. Until then, would you like me to drop this test?

tomponline: im ok to merge as this allows you to progress with your roadmap item, but id like to revisit in a follow up PR and move any non-snap specific tests into the main test suite run (even if they need to rely on a cached image).

test/suites/lxd_user.sh
```diff
@@ -38,3 +38,34 @@ test_lxd_user() {
 
   LXD_DIR="${bakLxdDir}"
 }
+
+snap_lxc_user() {
+    sudo -Hu testuser LXD_DIR=/var/snap/lxd/common/lxd-user lxc "${@}"
+}
+
+test_snap_lxd_user() {
```
tomponline, line 46: this test is a good example of a snap specific test, as it relies on the interplay between snap's systemd socket listeners and lxd-user inside the snap.

tomponline (approved):
LGTM thanks!

As discussed lets revisit this PR and where possibly move non-snap specific tests out of the `snap` prefixed test functions. If we provide an image for use with the non-snap tests then it opens up tests using `lxc exec` etc.

test/suites/basic.sh
simondeziel: This was a rebase issue in fact. This will land in `vm_empty` in the follow-up PR.

test/suites/devlxd.sh
simondeziel: https://github.com/canonical/lxd/pull/16756

PR #16739, opened 2025-10-15
Storage: Add explicit pool source validation

roosterfish:
This essentially splits the pool's source validation which is often performed during the driver's `Create` into its own driver level `ValidateSource` function.

In addition a driver level `SourceIdentifier` func is added which allows the backend to identify whether or not two individual pools with slightly different configs are actually pointing to the same underlying source. This further hardens the check around not allowing to reuse the same underlying storage for multiple storage pools in LXD.

The following identifiers are used per driver:
* Alletra: WSAPI address + CPG + name of volume set
* Ceph: Cluster name + pool name
* Cephfs: Cluster name + `source`
* PowerFlex: ID of used pool (pool always exists)
* Pure: ID of used pool (or name if not yet exists)
* Cephobject: RadosGW endpoint address
* Dir: `source` config setting
* Btrfs: `source` config setting
* LVM: Volume group name
* ZFS: Pool name

In case of Alletra we might also want to use the ID of the volume set (if exists) but we can only test the retrieval of it once we again have access to an environment. The code currently doesn't have any function for this.

This PR is the first set of changes required to properly introduce a `source.reuse` config key for storage pools.

### DIFF
--- a/lxd/storage/backend_lxd.go
+++ b/lxd/storage/backend_lxd.go
@@ -264,6 +264,23 @@ func (b *lxdBackend) Create(clientType request.ClientType, op *operations.Operat
 
 	revert.Add(func() { _ = os.RemoveAll(path) })
 
+	// Fill in the missing config.
+	// This is required before asking the driver for further source validation.
+	// The storage driver also expects the missing config to be present
+	// before trying to create or mount the actual storage pool.
+	err = b.Driver().FillConfig()
+	if err != nil {
+		return err
+	}
+
+	// Validate source.
+	// Ensure this is executed after creating the pool's storage path.
+	// Some drivers like dir expect the directory to be present for source validation.
+	err = b.validateSource()
+	if err != nil {
+		return err
+	}
+
 	if b.driver.Info().Remote && clientType != request.ClientTypeNormal {
 		if !b.driver.Info().MountedRoot {
 			// Create the directory structure.


lxd/storage/backend_lxd.go
```diff
@@ -116,6 +116,66 @@ func (b *lxdBackend) Validate(config map[string]string) error {
 	return b.Driver().Validate(config)
 }
 
+// validateSource checks whether or not the provided underlying source (based on the config) can be used.
+func (b *lxdBackend) validateSource() error {
+	// First let the source be validated by the driver itself.
+	err := b.Driver().ValidateSource()
+	if err != nil && !errors.Is(err, drivers.ErrNotSupported) {
```
tomponline: Shouldnt all drivers be required to validate source?

lxd/storage/backend_lxd.go
```diff
@@ -116,6 +116,66 @@ func (b *lxdBackend) Validate(config map[string]string) error {
...
+	// Second check if the same source is already used by any storage pool on this node.
+	sourceIdentifier, err := b.Driver().SourceIdentifier()
+	if err != nil {
+		return err
+	}
+
+	var poolNames []string
+
+	// Fetch the node local config of each storage pool.
+	err = b.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
```
tomponline: could use d.state.ShutdownCtx here

lxd/storage/backend_lxd.go
```diff
@@ -116,6 +116,66 @@ func (b *lxdBackend) Validate(config map[string]string) error {
...
+		var err error
+
+		poolNames, err = tx.GetStoragePoolNames(ctx)
```
tomponline: should we only get pools in a created state (in terms of the DB anyway).

lxd/storage/backend_lxd.go
roosterfish: Ups, left over from testing because at this stage not all drivers had their `ValidateSource` implemented.

lxd/storage/backend_lxd.go

lxd/storage/backend_lxd.go
roosterfish: Thanks fixed. Great we already have a func for this.

lxd/storage/drivers/driver_cephfs.go
```diff
@@ -138,6 +144,15 @@ func (d *cephfs) Create() error {
 		return errors.New("cephfs.path must match the source")
 	}
 
+	return nil
+}
+
+// Create is called during pool creation and is effectively using an empty driver struct.
+// WARNING: The Create() function cannot rely on any of the struct attributes being set.
```
tomponline, line 151: @roosterfish is the warning true anymore now you're validating the config up front?

tomponline: this is just a nit though

lxd/storage/drivers/driver_pure.go
```diff
@@ -210,17 +210,25 @@ func (d *pure) Validate(config map[string]string) error {
...
-	err := d.FillConfig()
+// SourceIdentifier returns a combined string consisting of the pool ID.
+// The underlying storage pool might not exist.
+// In this case the name of the pool is returned instead as a temporary identifier.
+func (d *pure) SourceIdentifier() (string, error) {
+	pool, err := d.client().getStoragePool(d.name)
 	if err != nil {
-		return err
+		// Pool does not yet exist, simply return its name instead.
+		if api.StatusErrorCheck(err, http.StatusNotFound) {
```
tomponline, line 220: Maybe the callers of `SourceIdentifer()` should be updated to handle `api.StatusErrorCheck(err, http.StatusNotFound)` instead, it seems a bit odd to be returning different identifiers based on whether the pool exists or not?

lxd/storage/backend_lxd.go
```diff
@@ -116,6 +116,66 @@ func (b *lxdBackend) Validate(config map[string]string) error {
...
+		if poolName == b.name {
+			continue
+		}
+
+		pool, err := LoadByName(b.state, poolName)
+		if err != nil {
+			return fmt.Errorf("Failed loading storage pool %q: %w", poolName, err)
+		}
+
+		// Skip pools with other drivers.
```
tomponline, line 161: this is a *what* but a *why* would be more helpful

tomponline (approved):
LGTM thanks!

There might need to be a couple of tweaks in a follow up PR.

The main thing was the pseudo identifier returned for Pure when no pool exists rather than the ID.

lxd/storage/drivers/driver_powerflex.go
```diff
@@ -154,14 +154,20 @@ func (d *powerflex) FillConfig() error {
...
 
-// Create is called during pool creation and is effectively using an empty driver struct.
-// WARNING: The Create() function cannot rely on any of the struct attributes being set.
-func (d *powerflex) Create() error {
-	err := d.FillConfig()
+// SourceIdentifier returns a combined string consisting of the pool ID.
+// The underlying storage pool always has to exist as the PowerFlex driver does
+// not create a new pool on the storage array.
+func (d *powerflex) SourceIdentifier() (string, error) {
+	pool, err := d.resolvePool()
```
tomponline, line 161: So this will return "not found" if the pool doesn't exist, so why is this different from a `SourceIdentifier()` compared to Pure?

lxd/storage/drivers/driver_alletra.go
```diff
@@ -219,14 +219,28 @@ func (d *alletra) Validate(config map[string]string) error {
...
 
-// Create is called during pool creation and is effectively using an empty driver struct.
-// WARNING: The Create() function cannot rely on any of the struct attributes being set.
-func (d *alletra) Create() error {
-	err := d.FillConfig()
-	if err != nil {
-		return err
+// SourceIdentifier returns a combined string consisting of the WSAPI address, CPG and volume set name.
+// We might want to use a volume sets ID, same as we do with pool ID for PowerFlex and Pure.
+func (d *alletra) SourceIdentifier() (string, error) {
```
tomponline, line 224: @roosterfish this seems different compared to Pure, here the `SourceIdentifier` comes purely from config, whereas the Pure driver is using the config to reach into the appliance and get the pool ID.

Why cant we do something similar here? Because from what I can see the Alletra driver does create something on the appliance (unlike Powerflex) via `CreateVolumeSet()`.

Should we not get a volume set ID (or stamp on in via the Alletra description field) for use as a source identifier?

Thanks

tomponline: or should Pure just use config for `SourceIdentifier` like this driver does?

lxd/storage/drivers/driver_cephfs.go
roosterfish: I agree we can get rid of those. Whilst this sentence is quite generic, I assume it was added because you cannot rely on any of the config being set. Will fix in the follow up PR.

lxd/storage/drivers/driver_alletra.go
roosterfish: > Should we not get a volume set ID

Yes we should, but I was not able to test this as we currently not have an Alletra array, and neither do our current API structs contain any field we could use and know it exists. Therefore I went with the combination of those three infos URL, name of CPG and volume set.

> or should Pure just use config for SourceIdentifier like this driver does

In case of Pure (and PowerFlex) we deal with a pool. PowerFlex requires the pool to be there so we can just rely on its ID.
In Pure we create the pool (which also has an ID) so we can also return this ID.
Ultimately Alletra should also use the ID. But until we can properly test this I have added a comment to the `SourceIdentifier` of Alletra about this.

The reason why the pure driver might either return the ID of the pool or its name is because when creating the Pure pool, the backend runs `SourceIdentifier` on all existing Pure pools (yielding the pool IDs). But it also runs `SourceIdentifier` on the to be create pool (which doesn't yet have an actual pool and pool ID on Pure) so to still return something unique we just return the chosen name of the pool as it is also unique compared to all the actual pool IDs of the existing Pure pools.

If this is too much differen for each of the drivers we could approach it the same way as for Alletra and simply return a combination of URL + (+ x if there is something like Alletra CPG) + pool name.

lxd/storage/drivers/driver_pure.go
roosterfish: Commented here https://github.com/canonical/lxd/pull/16739#discussion_r2481935956

lxd/storage/drivers/driver_alletra.go
tomponline: > But it also runs `SourceIdentifier` on the to be create pool (which doesn't yet have an actual pool and pool ID on Pure) so to still return something unique we just return the chosen name of the pool as it is also unique compared to all the actual pool IDs of the existing Pure pools.

Could we instead return a 404 and then detect that in the caller and skip it (it cant conflict if it doesnt exists right)?

roosterfish: Yes that would work.

Whilst thinking a bit more about it, why not approaching it like this for all pools:

> If this is too much differen for each of the drivers we could approach it the same way as for Alletra and simply return a combination of URL + (+ x if there is something like Alletra CPG) + pool name.

Because this way don't even need to reach out to the array to request any pool ID because we can just craft the identifier based on information we already have in the pool's config?

tomponline: yeah if that works then fine by me, I thought for some reason you wanted to get an appliance side identifier?

roosterfish: I found the ID to be the most unique we have, but let's rather make it consistent. Will add a follow up PR to fix this in PowerFlex, Pure and Alletra.

roosterfish: https://github.com/canonical/lxd/pull/16926

Issue #15995, opened 2025-07-11
nvidia-container dependency broken since v1.17.7


tomponline:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

24.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
latest/edge
```

### Output of "lxc info" or system info if it fails

```shell
N/A
```

### Issue description

We previously bumped our snap package dependencies of https://github.com/NVIDIA/libnvidia-container and https://github.com/NVIDIA/nvidia-container-toolkit to v1.17.6 and found that the GPU container passthrough feature stopped working (the container wouldn't start).

Recently we tried again and bumped it to v1.17.8 and this time the containers will start, but our [tests](https://github.com/canonical/lxd-ci/blob/main/.github/workflows/testflinger/nvidia-cdi-job.yml) are failing with nvidia-smi failing to find any devices.

https://github.com/canonical/lxd-ci/actions/runs/16213696751

```
2025-07-11T07:12:44.4628964Z + _run lspci -k
2025-07-11T07:12:44.4629908Z 0000:00:00.0 PCI bridge: NVIDIA Corporation Device 22b2
2025-07-11T07:12:44.4630562Z 	Kernel driver in use: pcieport
2025-07-11T07:12:44.4630976Z 0002:00:00.0 PCI bridge: NVIDIA Corporation Device 22b2
2025-07-11T07:12:44.4631391Z 	Kernel driver in use: pcieport
2025-07-11T07:12:44.4631767Z 0004:00:00.0 PCI bridge: NVIDIA Corporation Device 22b2
2025-07-11T07:12:44.4632169Z 	Kernel driver in use: pcieport
2025-07-11T07:12:44.4632810Z 0004:01:00.0 Non-Volatile memory controller: Samsung Electronics Co Ltd NVMe SSD Controller PM9A1/PM9A3/980PRO
2025-07-11T07:12:44.4633735Z 	Subsystem: Samsung Electronics Co Ltd NVMe SSD Controller PM9A1/PM9A3/980PRO
2025-07-11T07:12:44.4634431Z 	Kernel driver in use: nvme
2025-07-11T07:12:44.4634740Z 	Kernel modules: nvme
2025-07-11T07:12:44.4635085Z 0005:00:00.0 PCI bridge: NVIDIA Corporation Device 22b8
[...]

PR #16742, opened 2025-10-15
Bump Go minimum version to 1.25.3 and update Go modules

simondeziel:


simondeziel:
#15995 will need to be retested with an updated snap to see if the nvidia-container-toolkit v1.17.9 works better than v1.17.8

Issue #16643, opened 2025-09-30
Refactor cluster healing to not make internal API calls and generate lifecycle events


tomponline:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

24.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
N/A
```

### Output of "lxc info" or system info if it fails

```shell
N/A
```

### Issue description

Currently the auto cluster healing feature is triggered by a task on the leader member which then makes loop back API calls to itself to trigger healing. This is counter to how we prefer to do things these days, similar to how we avoided this type of action when doing forced project deletion.

We should refactor the auto cluster healing similar to https://github.com/lxc/incus/commit/ea8a49596c4a8b210fcd7ecbbfdf01efab911ccf (ensuring correct attribution where things can be cherry-picked).

Additionally we should include lifecycle events from: 

https://github.com/lxc/incus/commit/6326d9be173b858c5f97bb81be23cc6ff5a71e5a

And then update the cluster healing warning in the docs similar to:

https://github.com/lxc/incus/pull/1075/files#diff-dd7677da7f945b89e7e144c9848ca0603c332793beac980926c06021037e3620R88

### Steps to reproduce

N/A

### Information to attach

- [ ] Any relevant kernel output (`dmesg`)
- [ ] Instance log (`lxc info NAME --show-log`)
- [ ] Instance configuration (`lxc config show NAME --expanded`)
- [ ] Main daemon log (at `/var/log/lxd/lxd.log` or `/var/snap/lxd/common/lxd/logs/lxd.log`)
- [ ] Output of the client with `--debug`
- [ ] Output of the daemon with `--debug` (or use `lxc monitor` while reproducing the issue)

aviralgarg05:
I would like to solve the issue for hacktoberfest

tomponline:
Thanks @aviralgarg05 please can you ensure that any cherry-picks from Incus maintain their DSO attribution. Thanks

aviralgarg05:
Yes sure

aviralgarg05:
Can I get the issue assigned, with respect to hacktoberfest now

tomponline:
done

aviralgarg05:
Is it not under hacktoberfest? although I will still work on the issue

PR #16743, opened 2025-10-15
Refactor cluster healing and add lifecycle events

kadinsayani:
Closes https://github.com/canonical/lxd/issues/16643.

Includes cherry-picks from Incus (tracking s/s updated).

This pull request refactors and improves cluster healing. The main changes include removing the internal cluster heal endpoint, refactoring the healing logic to be more robust and event-driven, and adding lifecycle events during instance migration and cluster healing.

* Removed the internal `/cluster/heal/{name}` API endpoint and handler.
* Refactored the automatic cluster healing task to directly call a new `healClusterMember` function for each offline member, which now handles migration and emits lifecycle events.
* Updated the `evacuateClusterMember` function to use context and operation objects.
* Added lifecycle events for cluster member healing and restoration, enabling external systems to monitor and react to these events.
* Updated the cluster healing documentation to recommend monitoring for `cluster-member-healed` events.

### DIFF
--- a/doc/howto/cluster_manage.md
+++ b/doc/howto/cluster_manage.md
@@ -129,6 +129,8 @@ When the healed cluster member is available again, you must manually {ref}`resto
 ```{warning}
 Enabling the cluster healing threshold carries the risk that LXD might incorrectly judge a cluster member as offline while it is still running workloads. Short-lived network issues or temporary high load might cause a cluster member to briefly stop responding to heartbeat or ICMP packets. If a healing threshold is set, LXD might then start that member's instances on another cluster member even though they're still active on the original. Since cluster members share the same storage, this can lead to data corruption.
 
+To avoid this, it's critical to ensure that any server marked as offline is actually offline and not still running instances. You can automate this by monitoring for `cluster-member-healed` events and shutting off the affected server through its remote power interface, such as a Baseboard Management Controller (BMC) or Power Distribution Unit (PDU).
+
 To reduce the chance of false healing events, set {config:option}`server-cluster:cluster.healing_threshold` as high as possible within your availability targets.
 ```
 
--- a/doc/.custom_wordlist.txt
+++ b/doc/.custom_wordlist.txt
@@ -30,6 +30,7 @@ bootable
 bpffs
 BPFFS
 BPF
+BMC
 Btrfs
 bugfix
 bugfixes
@@ -210,6 +211,7 @@ OVS
 Pbit
 PCI
 PCIe
+PDU
 passthrough
 passthroughs
 peerings


doc/howto/cluster_manage.md
```diff
@@ -129,6 +129,8 @@ When the healed cluster member is available again, you must manually {ref}`resto
 ```{warning}
 Enabling the cluster healing threshold carries the risk that LXD might incorrectly judge a cluster member as offline while it is still running workloads. Short-lived network issues or temporary high load might cause a cluster member to briefly stop responding to heartbeat or ICMP packets. If a healing threshold is set, LXD might then start that member's instances on another cluster member even though they're still active on the original. Since cluster members share the same storage, this can lead to data corruption.
 
+It's critical to ensure that a server which is considered offline is in fact offline and isn't still running its instances. To automatically achieve this, monitor for `cluster-member-healed` events and promptly cut the power to the server in question by interacting with its BMC or PDU.
```
minaelee:
```suggestion
To avoid this, it's critical to ensure that any server marked as offline is actually offline and not still running instances. You can automate this by monitoring for `cluster-member-healed` events and shutting off the affected server through its remote power interface, such as a Baseboard Management Controller (BMC) or Power Distribution Unit (PDU).
```

kadinsayani: Thanks for the suggestion!

lxd/api_cluster.go
```diff
@@ -4576,5 +4582,12 @@ func healClusterMember(s *state.State, gateway *cluster.Gateway, op *operations.
 		return nil
 	}
 
-	return evacuateClusterMember(context.Background(), s, gateway, op, name, "migrate", nil, migrateFunc)
+	err := evacuateClusterMember(context.Background(), s, gateway, op, name, "migrate", nil, migrateFunc)
```
tomponline: should "migrate" argument here be "heal", so it aligns with https://github.com/canonical/lxd/pull/16743/commits/aaa4f5b4c0ba595310640c1e6354227fe20cc544#diff-07fccfa2de6bd472d27c84cb9ca8cd9319e32135864bf114267111df3d3fad2fR3277

lxd/api_cluster.go
```diff
@@ -4576,5 +4582,12 @@ func healClusterMember(s *state.State, gateway *cluster.Gateway, op *operations.
...
+	if err == nil {
```
tomponline: the structure here is odd, normally we do happy path, so should be `if err != nil { logger.Error("Failed healing cluster memb...` etc

lxd/api_cluster.go
```diff
@@ -4519,6 +4519,9 @@ func autoHealClusterTask(stateFunc func() *state.State, gateway *cluster.Gateway
 }
 
 func healClusterMember(s *state.State, gateway *cluster.Gateway, op *operations.Operation, name string) error {
+	logger.Info("Starting cluster healing", logger.Ctx{"server": name})
```
tomponline: there seems to be a convention of "member: name" looking in the code base rather than "server".

tomponline:
@kadinsayani as this PR touches instance migration logic, please can you confirm that live migration works both between identical cluster members and standalone installs. Thanks! 

As we're still not there yet with https://github.com/canonical/lxd-ci/pull/256

Thanks!

lxd/api_cluster.go
kadinsayani: Ah yes, thanks.

lxd/api_cluster.go
kadinsayani: Ack, changed.

kadinsayani:
> @kadinsayani as this PR touches instance migration logic, please can you confirm that live migration works both between identical cluster members and standalone installs. Thanks!
> 
> As we're still not there yet with [canonical/lxd-ci#256](https://github.com/canonical/lxd-ci/pull/256)
> 
> Thanks!

Tested and all good :white_check_mark:

tomponline (approved):
Thanks!

PR #16746, opened 2025-10-16
Backend: Lock immutable volume config settings

roosterfish:
Fixes a recent finding which allows unsetting immutable volume config settings.

### DIFF
--- a/lxd-agent/devlxd_storage.go
+++ b/lxd-agent/devlxd_storage.go
@@ -109,7 +109,7 @@ func devLXDStoragePoolVolumesPostHandler(d *Daemon, r *http.Request) *devLXDResp
 var devLXDStoragePoolVolumeTypeEndpoint = devLXDAPIEndpoint{
 	Path:   "storage-pools/{pool}/volumes/{type}/{volume}",
 	Get:    devLXDAPIEndpointAction{Handler: devLXDStoragePoolVolumeGetHandler},
-	Put:    devLXDAPIEndpointAction{Handler: devLXDStoragePoolVolumePutHandler},
+	Patch:  devLXDAPIEndpointAction{Handler: devLXDStoragePoolVolumePatchHandler},
 	Delete: devLXDAPIEndpointAction{Handler: devLXDStoragePoolVolumeDeleteHandler},
 }
 
@@ -145,7 +145,7 @@ func devLXDStoragePoolVolumeGetHandler(d *Daemon, r *http.Request) *devLXDRespon
 	return okResponseETag(vol, "json", etag)
 }
 
-func devLXDStoragePoolVolumePutHandler(d *Daemon, r *http.Request) *devLXDResponse {
+func devLXDStoragePoolVolumePatchHandler(d *Daemon, r *http.Request) *devLXDResponse {
 	poolName, err := url.PathUnescape(r.PathValue("pool"))
 	if err != nil {
 		return errorResponse(http.StatusBadRequest, err.Error())


roosterfish:
Mhh, looks `test_devlxd_volume_management` is depending on the `volatile.uuid` unset behavior.

MusicDin:
> Mhh, looks test_devlxd_volume_management is depending on the volatile.uuid unset behavior.

What do you mean by depending on `volatile.uuid` being unset? It could be that I've done a mistake, as only important volatile keys in this test are `volatile.<device>.devlxd.owner` (for instances) and `volatile.devlxd.owner` (for volumes)

roosterfish:
> What do you mean by depending on `volatile.uuid` being unset? 

It updates the storage vols using the `devlxd-client storage update-volume` and omits the `volatile.uuid` (and `block.filesystem` if set) in the `{...}` payload which causes removal of those config keys. 

Your test seems right, it's just that it depends on the bug which shouldn't allow unsetting those config keys.

roosterfish:
Think I found a good option to keep all of the original config and only modify the description and size.

MusicDin:
Hmm, I think it would be best if we shift to using patch instead of put for storage volumes through DevLXD. Any new field added in the future will be wiped by DevLXD, especially when considering that DevLXD does not return the entire volume state.

tomponline:
> Hmm, I think it would be best if we shift to using patch instead of put for storage volumes through DevLXD. Any new field added in the future will be wiped by DevLXD, especially when considering that DevLXD does not return the entire volume state.

Agree, good idea.

MusicDin:
The patch is already supported, so I suggest just changing `UpdateStoragePoolVolume` in DevLXD client to use patch:

https://github.com/canonical/lxd/blob/c3dafbf5969144cc0f19bbd131c785db54efee03/client/devlxd_storage.go#L66-L76

roosterfish:
> The patch is already supported, so I suggest just changing `UpdateStoragePoolVolume` in DevLXD client to use patch:

That sounds like a great idea. Then I can drop all the changes made to this test from the PR.

MusicDin:
And please, if you could adjust the LXD agent to expose patch instead of put. Otherwise, DevLXD client won't work with LXD agent

https://github.com/canonical/lxd/blob/c3dafbf5969144cc0f19bbd131c785db54efee03/lxd-agent/devlxd_storage.go#L109-L114

lxd/storage/backend_lxd.go
```diff
@@ -6248,11 +6248,10 @@ func (b *lxdBackend) UpdateCustomVolume(projectName string, volName string, newD
...
 		}
 
-		if changedConfig["security.shifted"] != "" {
-			// Confirm that no running instances are using it when changing shifted state.
-			if inst.IsRunning() && changedConfig["security.shifted"] != "" {
-				return errors.New("Cannot modify shifting with running instances using the volume")
-			}
+		securityShifted, ok := changedConfig["security.shifted"]
+		// Confirm that no running instances are using it when changing shifted state.
+		if ok && inst.IsRunning() && (securityShifted != "" || securityShifted == "") {
```
tomponline: why is this needed `(securityShifted != "" || securityShifted == "")`? if "its empty or not empty" seems redundant? 
Just its presence in `changedConfig` should be sufficient no?

lxd/storage/backend_lxd.go
```diff
@@ -6269,12 +6268,14 @@ func (b *lxdBackend) UpdateCustomVolume(projectName string, volName string, newD
 		}
 
 		// Check that the volume's block.filesystem property isn't being changed.
-		if changedConfig["block.filesystem"] != "" {
+		blockFilesystem, ok := changedConfig["block.filesystem"]
+		if ok && (blockFilesystem != "" || blockFilesystem == "") {
```
tomponline: i think we just need to test for `ok == true`

lxd/storage/backend_lxd.go
roosterfish: Mhh valid point. The case we haven't checked before (which is the bug) is in case `ok==false` which would automatically mean the key also is `==""`.

tomponline: ok==false means its not changed.

lxd/storage/backend_lxd.go

tomponline (approved):
Thanks!

PR #16749, opened 2025-10-16
`actions/cache`'s `restore-keys` is a multiline string, not a list

simondeziel:
The bogus `- ` prefix what found when looking at a CI job trying to restore an artifact after a manual purge:
```
Cache not found for input keys: deps-ubuntu-22.04-amd64-20251016, - deps-ubuntu-22.04-amd64-20251015
```

Upstream doc: https://github.com/actions/cache?tab=readme-ov-file#inputs


tomponline (approved):
So the cache wasn't working?

simondeziel:
> So the cache wasn't working?

It was working but never succeeded restoring an older asset, the one from yesterday. Pretty harmless bug but still a bug.

PR #16753, opened 2025-10-17
GitHub: Create dqlite directory only if not exist

roosterfish:
In case the /home/runner/go/bin/dqlite directory was restored from cache, don't try recreating it.

See failure https://github.com/canonical/lxd/actions/runs/18585640979/job/52988847361?pr=16746.

### DIFF
--- a/.github/actions/cache-dqlite-liblxc/action.yml
+++ b/.github/actions/cache-dqlite-liblxc/action.yml
@@ -42,7 +42,7 @@ runs:
         make deps
 
         # Include dqlite libs in dependencies for system tests.
-        mkdir /home/runner/go/bin/dqlite
+        mkdir -p /home/runner/go/bin/dqlite
         mv /home/runner/work/lxd/lxd-test/vendor/dqlite/include /home/runner/go/bin/dqlite/include
         mv /home/runner/work/lxd/lxd-test/vendor/dqlite/.libs /home/runner/go/bin/dqlite/libs
 


tomponline:
@roosterfish any idea why we've not seen that issue before?

Could be related to https://github.com/canonical/lxd/pull/16749 perhaps?

roosterfish:
> Could be related to #16749 perhaps?

Potentially yes. It might be that it never restored the cache so it always succeeded in recreating the directory.

But what I don't understand is why it didn't fail on this PR already.

tomponline (approved):
ta

simondeziel:
It seems that the cache was restored using the **fallback key** (so YESTERDAY's artifact):

```
+ YESTERDAY=20251016
+ echo KEY=deps-ubuntu-22.04-amd64-20251017
+ echo FALLBACK_KEY=deps-ubuntu-22.04-amd64-20251016
Run actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809
Cache hit for restore-key: deps-ubuntu-22.04-amd64-20251016
Received 11819068 of 11819068 (100.0%), 11.5 MBs/sec
Cache Size: ~11 MB (11819068 B)
/usr/bin/tar -xf /home/runner/work/_temp/4ceae9b0-0f28-4f04-aa88-fec632ca2ce2/cache.tzst -P -C /home/runner/work/lxd/lxd --use-compress-program unzstd
Cache restored successfully
Cache restored from key: deps-ubuntu-22.04-amd64-20251016
```

But then it went on and built `dqlite` and `liblxc` despite this supposedly being gated by `if: ${{ steps.cache-deps.outputs.cache-hit != 'true' }}`. I'll dig into this because it's only since https://github.com/canonical/lxd/pull/16749 was merged that we can use the fallback mechanism.

simondeziel:
https://github.com/actions/cache?tab=readme-ov-file#restoring-and-saving-cache-using-a-single-action explains the `cache-hit` is only set to `true` if the primary key is restored, it will remain `false` if a fallback key does a hit... I feel like this violates the principle of least surprise but oh well, at least it's documented!

Should be addressed by https://github.com/canonical/lxd/pull/16755

Issue #16642, opened 2025-09-30
Suspected regression in cluster healing that stops network on member that triggers healing


tomponline:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu 

### Distribution version

24.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
N/A
```

### Output of "lxc info" or system info if it fails

```shell
N/A
```

### Issue description

Since https://github.com/canonical/lxd/commit/90ed1c7ee8511746a544da065de08db1b88bab75 the `evacuateClusterMember` function has stood-down the networks on the local member where it is called.

However that change failed to take into account that this function is also called from `internalClusterHeal()` when run on a different cluster member than the member that down/unavailable.

I suspect therefore that now when auto cluster healing kicks in on the leader node it will stand down all of the networks on that member leading to cluster distruption.

### Steps to reproduce

No confirmed reproducer at this time.

But suspect you need to set https://documentation.ubuntu.com/lxd/latest/server/#server-cluster:cluster.healing_threshold to non-zero and then stop one cluster member and then watch to see if the leader member turns its OVN networks off after that timeout has elapsed.

### Information to attach

- [ ] Any relevant kernel output (`dmesg`)
- [ ] Instance log (`lxc info NAME --show-log`)
- [ ] Instance configuration (`lxc config show NAME --expanded`)
- [ ] Main daemon log (at `/var/log/lxd/lxd.log` or `/var/snap/lxd/common/lxd/logs/lxd.log`)
- [ ] Output of the client with `--debug`
- [ ] Output of the daemon with `--debug` (or use `lxc monitor` while reproducing the issue)

PR #16757, opened 2025-10-17
Fix network evacuation regression during cluster healing

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/16642.

### DIFF
--- a/lxd/instance/drivers/driver_common.go
+++ b/lxd/instance/drivers/driver_common.go
@@ -1677,18 +1677,18 @@ func (d *common) canMigrate(inst instance.Instance) (migrate bool, live bool) {
 	config := d.ExpandedConfig()
 	val, ok := config["cluster.evacuate"]
 	if !ok {
-		val = "auto"
+		val = api.ClusterEvacuateModeAuto
 	}
 
-	if val == "migrate" {
+	if val == api.ClusterEvacuateModeMigrate {
 		return true, false
 	}
 
-	if val == "live-migrate" {
+	if val == api.ClusterEvacuateModeLiveMigrate {
 		return true, true
 	}
 
-	if val == "stop" {
+	if val == api.ClusterEvacuateModeStop {
 		return false, false
 	}
 


test/suites/clustering.sh
```diff
@@ -1337,6 +1337,111 @@ test_clustering_network() {
...
 
+test_clustering_heal_networks_stop() {
+  echo "==> Test: cluster healing does not shut down networks on the leader node when evacuating an offline member"
+  # Regression test for https://github.com/canonical/lxd/issues/16642.
+  # This test uses OVN networks because they have more impactful Evacuate() operations (removing chassis entries).
+  # While bridge networks also call bgpClear() on evacuation, OVN network evacuation is more likely to expose issues.
+
+  if ! ovn_enabled; then
+    echo "==> SKIP: OVN not configured. Skipping cluster healing network test..."
+    return
```
simondeziel: There is a nice mechanism for unmet dependencies. It's unfortunately not widely used (still working on it) but it'd look like this:
```suggestion
    export TEST_UNMET_REQUIREMENT="OVN not configured"
    return
```

This will result in the following:

```
echo "==> SKIP: test_clustering_heal_networks_stop OVN not enabled"

File: test/suites/clustering.sh [hunk 2]
```diff
@@ -1337,6 +1337,111 @@ test_clustering_network() {
...
+  echo "Setup OVN on all nodes"
+  (
+    LXD_DIR="${LXD_ONE_DIR}" setup_ovn
+    LXD_DIR="${LXD_TWO_DIR}" setup_ovn
+    LXD_DIR="${LXD_THREE_DIR}" setup_ovn
+  )
+
+  echo "Create a physical uplink network for OVN"
+  uplink_network="${prefix}uplink"
+  ip link add dummy0 type dummy
```
simondeziel: For those `ip link add`, I tend to add a `ip link delete ... || true` right before to make manual test runs less frustrating in case the test fails mid way ;)

test/suites/clustering.sh
```diff
@@ -1337,6 +1337,111 @@ test_clustering_network() {
...
+  rm -f "${LXD_ONE_DIR}/unix.socket"
+
+  teardown_clustering_netns
+  teardown_clustering_bridge
+
+  kill_lxd "${LXD_ONE_DIR}"
+  kill_lxd "${LXD_THREE_DIR}"
+
+  # Cleanup OVN and physical interface
+  LXD_DIR="${LXD_ONE_DIR}" unset_ovn_configuration || true
```
simondeziel: I think this needs to happen before `LXD_DIR="${LXD_ONE_DIR}" lxd shutdown`

test/suites/clustering.sh
kadinsayani: Nice! I'll use that moving forward :)

test/suites/clustering.sh
```diff
@@ -1337,6 +1337,127 @@ test_clustering_network() {
...
+
+  # Clean up dummy interface if it exists from previous failed run
+  nsenter -n -t "${daemon_pid1}" -- ip link delete dummy0 || true
+  nsenter -n -t "${daemon_pid2}" -- ip link delete dummy0 || true
+  nsenter -n -t "${daemon_pid3}" -- ip link delete dummy0 || true
+
+  # Create dummy interface in each node's namespace
+  nsenter -n -t "${daemon_pid1}" -- ip link add dummy0 type dummy
+  nsenter -n -t "${daemon_pid2}" -- ip link add dummy0 type dummy
+  nsenter -n -t "${daemon_pid3}" -- ip link add dummy0 type dummy  # Create physical network with per-node parent configuration
```
simondeziel: `ip link` seems to support `netns $PID` which might be handy here.

test/suites/clustering.sh
```diff
@@ -1337,6 +1337,127 @@ test_clustering_network() {
...
+test_clustering_heal_networks_stop() {
+  echo "==> Test: cluster healing does not shut down networks on the leader node when evacuating an offline member"
+  # Regression test for https://github.com/canonical/lxd/issues/16642.
+  # This test uses OVN networks because they have more impactful Evacuate() operations (removing chassis entries).
+  # While bridge networks also call bgpClear() on evacuation, OVN network evacuation is more likely to expose issues.
+
+  if ! ovn_enabled; then
+    export TEST_UNMET_REQUIREMENT="OVN not configured"
+    return
+  fi
```
simondeziel: I'd move that right at the beginning, even before the first `echo`.

test/suites/clustering.sh
```diff
@@ -1337,6 +1337,127 @@ test_clustering_network() {
...
+  LXD_DIR="${LXD_THREE_DIR}" lxc network info "${ovn_net}"
+
+  echo "Cleanup"
+  # Cleanup OVN and physical interface
+  LXD_DIR="${LXD_ONE_DIR}" unset_ovn_configuration || true
+
+  # Delete dummy interfaces from each node's namespace
+  nsenter -n -t "${daemon_pid1}" -- ip link delete dummy0 || true
+  nsenter -n -t "${daemon_pid2}" -- ip link delete dummy0 || true
+  nsenter -n -t "${daemon_pid3}" -- ip link delete dummy0 || true  LXD_DIR="${LXD_THREE_DIR}" lxd shutdown
```
simondeziel: Same thing about `ip link netns $PID`

test/suites/clustering.sh
```diff
@@ -1337,6 +1337,127 @@ test_clustering_network() {
...
+  LXD_DIR="${LXD_ONE_DIR}" lxd shutdown
+  sleep 0.5
+  rm -f "${LXD_THREE_DIR}/unix.socket"
+  rm -f "${LXD_ONE_DIR}/unix.socket"
+
+  teardown_clustering_netns
+  teardown_clustering_bridge
+
+  kill_lxd "${LXD_ONE_DIR}"
+  kill_lxd "${LXD_THREE_DIR}"
```
simondeziel, line 1434: I noticed that `LXD_TWO_DIR` is never `kill_lxd`, intended?

kadinsayani: Nice catch, we should still kill the second node during cleanup.

test/suites/clustering.sh
kadinsayani: Nice, that would work well.

However, I don't think my approach for testing this regression will work since we are using network namespaces for each cluster, and we need a shared L2 for OVN.

We're looking to ensure the `Evacuate()` method is not called during automatic cluster healing. For network drivers other than OVN, the `Evacuate()` implementation is trivial and in some cases a no-op.

kadinsayani: The bug fix for this regression is also trivial:

```Go
-       // Evacuate networks too.
-       networkStop(s, true)
+       if mode != "heal" {
+               // Evacuate networks too.
+               networkStop(s, true)
+       }
```

kadinsayani: I can't think of a better way to test at this time :/

kadinsayani: Thanks for your review @simondeziel. I'm going to skip adding tests for this at this time. Perhaps we can revisit this when we add nested VM tests.

roosterfish (approved):

tomponline (changes requested):
Thanks!

@kadinsayani did you test this?

Can we add tests for it using the BGP internal API to check whether a bridge network is stepped down when evacuating by checking there's no exported routes?

Also can we add constants for the modes?

test/suites/clustering.sh
```diff
@@ -1337,6 +1337,105 @@ test_clustering_network() {
...
+  LXD_THREE_DIR=$(mktemp -d -p "${TEST_DIR}" XXX)
+  ns3="${prefix}3"
+  spawn_lxd_and_join_cluster "${ns3}" "${bridge}" "${cert}" 3 1 "${LXD_THREE_DIR}" "${LXD_ONE_DIR}"
+
+  echo "Create bridge network to start BGP listener on"
+  bgpbr="${prefix}bgpbr"
+  LXD_DIR="${LXD_ONE_DIR}" lxc network create "${bgpbr}" --target node1
+  LXD_DIR="${LXD_ONE_DIR}" lxc network create "${bgpbr}" --target node2
+  LXD_DIR="${LXD_ONE_DIR}" lxc network create "${bgpbr}" --target node3
+  LXD_DIR="${LXD_ONE_DIR}" lxc network create "${bgpbr}"
```
simondeziel: Could you hardcode IP ranges to avoid this taking a while?

test/suites/clustering.sh
```diff
@@ -1337,6 +1337,105 @@ test_clustering_network() {
...
+  echo "Create network forward"
+  LXD_DIR="${LXD_ONE_DIR}" lxc network forward create "${net}" 198.51.100.1
+
+  echo "Check forward is exported via BGP prefixes"
+  LXD_DIR="${LXD_ONE_DIR}" lxc query /internal/testing/bgp | grep -F "198.51.100.1/32"
+
+  echo "Enable the BGP listener"
+  LXD_DIR="${LXD_ONE_DIR}" lxc config set core.bgp_address="${bgpIP}:8874"
+  LXD_DIR="${LXD_ONE_DIR}" lxc config set core.bgp_asn=65536
+  LXD_DIR="${LXD_ONE_DIR}" lxc config set core.bgp_routerid="${bgpIP}"
```
simondeziel: Worth doing in one go?

kadinsayani:
> Thanks!

Np!

> @kadinsayani did you test this?

Yeah, I tested locally a few times.
 
> Can we add tests for it using the BGP internal API to check whether a bridge network is stepped down when evacuating by checking there's no exported routes?

Great idea! I've added tests following your suggested approach.

> Also can we add constants for the modes?

Done :white_check_mark:

test/suites/clustering.sh
kadinsayani: Sure, thanks for suggesting!

test/suites/clustering.sh
```diff
@@ -1337,6 +1337,103 @@ test_clustering_network() {
...
+  LXD_DIR="${LXD_TWO_DIR}" lxc config set cluster.healing_threshold 1
+
+  echo "Stop node2 (a non-leader member) to trigger healing"
+  LXD_DIR="${LXD_TWO_DIR}" lxd shutdown
+  sleep 0.5
+  rm -f "${LXD_TWO_DIR}/unix.socket"
+  kill_lxd "${LXD_TWO_DIR}"
+
+  echo "Wait for the member to be detected as offline and healing to kick in"
+  sleep 45
```
tomponline, line 1416: is this long sleep needed?

test/suites/clustering.sh
```diff
@@ -1337,6 +1337,103 @@ test_clustering_network() {
...
+  LXD_DIR="${LXD_ONE_DIR}" lxc query /internal/testing/bgp | grep -F "198.51.100.1/32"
+
+  echo "Set a low offline threshold so the member is marked offline quickly"
+  LXD_DIR="${LXD_TWO_DIR}" lxc config set cluster.offline_threshold 11
+
+  echo "Set the healing threshold to automatically heal offline members"
+  LXD_DIR="${LXD_TWO_DIR}" lxc config set cluster.healing_threshold 1
+
+  echo "Stop node2 (a non-leader member) to trigger healing"
+  LXD_DIR="${LXD_TWO_DIR}" lxd shutdown
```
tomponline, line 1410: we should simulate a crash by just killing lxd and not shutting it down ideally

tomponline (approved):
Thanks!

test/suites/clustering.sh
kadinsayani: Yes, this is as low as I could go and still consistently observe the regression (without the fix).

test/suites/clustering.sh
kadinsayani: Quick follow-up here: https://github.com/canonical/lxd/pull/16791.

test/suites/clustering.sh
tomponline: do you know why?

kadinsayani: No, but I will investigate further.

kadinsayani: `autoHealClusterTask()` runs every minute:

https://github.com/canonical/lxd/blob/c989f9c996029a78092f6c4bb22def44df214cef/lxd/api_cluster.go#L4515

kadinsayani: It looks like `cluster.offline_threshold` is the value actually used for the healing threshold:

https://github.com/canonical/lxd/blob/c989f9c996029a78092f6c4bb22def44df214cef/lxd/cluster/config/config.go#L244-L247

In the test we have `cluster.healing_threshold = cluster.offline_threshold = 11`. Therefore, at minimum we have to wait 22s. Factoring in run time and waiting for `autoHealClusterTask()`, I think 45s is fair.

tomponline: @kadinsayani we have a `/internal/...` endpoint for triggering image refresh iirc (which is called by the tests), so perhaps we could have something similar to trigger healing evaluation?

kadinsayani: Great idea, I'll add that endpoint and update the tests.

kadinsayani: https://github.com/canonical/lxd/pull/16793

PR #16758, opened 2025-10-17
Makefile improvements

simondeziel:
* Simplify how coverage-enabled binaries are produced
* Split up build targets to allow build specific binaries
* Create the `GOCOVERDIR` when needed

The second change should speed up the workflows like this:

1. change code
2. `make lxd`   <== this became faster
3. run a test
4. repeat

`make lxd` is now noticeably faster for both the first build and subsequent rebuilds (no-change here):

|Type|Before|After|
|:---|:---|:---
|first build|30s|14s|
|rebuild|3.7s|1.7s|

### DIFF
--- a/test/main.sh
+++ b/test/main.sh
@@ -384,6 +384,9 @@ export SMALL_ROOT_DISK="${SMALL_ROOT_DISK:-"root,size=32MiB"}"
 export SMALLEST_VM_ROOT_DISK="3584MiB"
 export SMALL_VM_ROOT_DISK="${SMALL_VM_ROOT_DISK:-"root,size=${SMALLEST_VM_ROOT_DISK}"}"
 
+# Create GOCOVERDIR if needed
+[ -n "${GOCOVERDIR:-}" ] && mkdir -p "${GOCOVERDIR}"
+
 # Spawn an interactive test shell when invoked as `./main.sh test-shell`.
 # This is useful for quick interactions with LXD and its test suite.
 if [ "${1:-"all"}" = "test-shell" ]; then


Makefile
```diff
@@ -377,11 +347,8 @@ check: default check-gomin check-unit test-binaries
 
 .PHONY: check-unit
 check-unit:
-ifeq "$(GOCOVERDIR)" ""
-	CGO_LDFLAGS_ALLOW="$(CGO_LDFLAGS_ALLOW)" go test -mod=readonly -v -failfast -tags "$(TAG_SQLITE3)" $(DEBUG) ./...
-else
-	CGO_LDFLAGS_ALLOW="$(CGO_LDFLAGS_ALLOW)" go test -mod=readonly -v -failfast -tags "$(TAG_SQLITE3)" $(DEBUG) ./... -cover -test.gocoverdir="$(GOCOVERDIR)"
-endif
+	$(shell [ -n "$(GOCOVERDIR)" ] && mkdir -p "$(GOCOVERDIR)")
```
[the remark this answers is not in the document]
simondeziel: Did you ran out of space chars?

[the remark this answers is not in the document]
simondeziel: This turned out to be a bogus suggestion, the `$(shell )` wrapper is required otherwise non-coverage builds fail with `[ -n "" ] && mkdir -p ""` error'ing out.

Makefile
```diff
@@ -48,95 +55,48 @@ ifeq "$(TAG_SQLITE3)" ""
...
 client:
-ifeq "$(GOCOVERDIR)" ""
-	go install -v -trimpath $(DEBUG) ./lxc
-else
-	go install -v -trimpath -cover $(DEBUG) ./lxc
-endif
-
-	@echo "LXD client built successfully"
+	go install -v -trimpath $(COVER) $(DEBUG) ./lxc
+	@echo "LXD $@ built successfully"
```
kadinsayani, line 75: Is this echo supposed to include "LXD"?

kadinsayani: Nevermind, I see that the other echo statements don't include "LXD" because their name starts with "lxd".

simondeziel: Yeah, it's probably to avoid the confusion with `liblxc` from before we called it that way.

PR #16760, opened 2025-10-19
`lxc/exec`: Warn about using lxc exec for privileged instances

### LINKED ISSUE #16058: warn when user lxd shells or lxd execs into a privileged container
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

I'm always frustrated when I can't find out why `snap install something` fails with strange error messages because my default profile contains `security.privileged: "true"`.

### Describe the solution you'd like

every time the user shells into a privileged container, log a warning that:
- links to a website that describes the consequence of privileged mode, including that:
  - snaps/systemd won't work on noble
- suggests using LXD VMs if you really need privileged

### Describe alternatives you've considered

n/a

### Additional context

n/a

haanhvu:
fixes #16058

I checked the fix locally and it worked:
```
$ ~/go/bin/lxc config get juju-7a3df1-0 security.privileged
true
$ ~/go/bin/lxc exec juju-7a3df1-0 -- echo "ok"
WARNING[2025-10-19T18:59:57+07:00] juju-7a3df1-0 is a privileged container (security.privileged: true), so snapd/systemd won't work on noble. To fix this, please consider using LXD VMs if you need privileged security. 
ok
```

### DIFF
--- a/lxc/exec.go
+++ b/lxc/exec.go
@@ -11,7 +11,7 @@ import (
 	"github.com/gorilla/websocket"
 	"github.com/spf13/cobra"
 
-	"github.com/canonical/lxd/client"
+	lxd "github.com/canonical/lxd/client"
 	"github.com/canonical/lxd/shared/api"
 	cli "github.com/canonical/lxd/shared/cmd"
 	"github.com/canonical/lxd/shared/i18n"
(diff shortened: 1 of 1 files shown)


tomponline:
Thanks for taking a stab at this.

But for the reasons explained in https://github.com/canonical/lxd/issues/16058#issuecomment-3441815163 I don't think we will want to proceed with this approach as it currently stands.

PR #16761, opened 2025-10-19
github: workaround unreliable partition mapping in `actions/setup-microceph`

simondeziel:


simondeziel:
https://github.com/canonical/lxd/actions/runs/18635296232/job/53128595039?pr=16761#step:14:372 shows that the retry mechanism worked.

tomponline (approved):
Ta

roosterfish:
I think we should have the same in MicroCloud as I remember seeing this error there too.

simondeziel:
> I think we should have the same in MicroCloud as I remember seeing this error there too.

I couldn't find where the ephemeral disk was partitioned and the result `microceph disk add`'ed. If you get the error again, please let me know.

roosterfish:
> I couldn't find where the ephemeral disk was partitioned and the result `microceph disk add`'ed. If you get the error again, please let me know.

In case of MicroCloud the `disk add` part is performed by MicroCloud itself, so we "only" have to wait until the disk is ready before starting to `microcloud init` something. Maybe we can just check until the path appears?

simondeziel:
There is already some wait logic in place as the action waits for the partitions to appear in `/proc/partitions` before using them to back a loop device. Then the loop dev is feed to `microceph disk add`. After that, the error complains the partition's `/dev/disk/by-id` symlink is not there.

It's not entirely clear how or what goes from the loop dev to the symlink but maybe there is something else more reliable we could wait on.

PR #16764, opened 2025-10-20
lxd/instance/instancetype/instance: Allow unsetting volatile.<name>.bus config key

tomponline:


tomponline:
> Can this also be unset when the instance is running?

yes, we allow manual control of this.

PR #16774, opened 2025-10-20
storage: Return StatusLocked error code when custom volume is in use

MusicDin:
Returning `StatusLocked` helps us identify when the volume is in use.
In the LXD CSI driver, we need this to differentiate *volume in use* from other internal errors.

---

Additionally, this PR enables VM tests for DevLXD volume management and ensures the reported error code is 423 (StatusLocked) when updating a volume that is in use.

### DIFF
--- a/test/suites/devlxd.sh
+++ b/test/suites/devlxd.sh
@@ -477,6 +477,42 @@ EOF
     # Ensure storage volumes are deleted.
     lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage volumes "${pool}" | jq --exit-status '. == []'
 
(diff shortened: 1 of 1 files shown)


test/suites/devlxd.sh
tomponline, line 1: Could we have tested devlxd locking status codes with containers or does it have to be VMS due to needing a block volume?

MusicDin: I could try shrinking FS volumes for certain local storage drivers, which would error out with `ErrInUse` even for FS volumes.

However, the benefit of testing VMs is that both DevLXD and LXD Agent are tested. They both need to report correct status for the test to pass. This means that DevLXD itself is indirectly tested.

PR #16775, opened 2025-10-20
Workflow hardening

simondeziel:
Apply some best practices with GH actions to defend against some attacks.

Also, rework how client binaries are built to better align with how they are normally produced by the `Makefile`.


tomponline (dismissed):

tomponline (approved):
thanks!

PR #16779, opened 2025-10-20
`volatile.cluster.group` fixes, add cluster group in use checks and used_by tests (from Incus)

kadinsayani:
Follow-up to https://github.com/canonical/lxd/pull/16478.

### DIFF
--- a/test/suites/clustering.sh
+++ b/test/suites/clustering.sh
@@ -3651,16 +3651,16 @@ test_clustering_evacuation() {
   LXD_DIR="${LXD_ONE_DIR}" lxc cluster group create foo
   LXD_DIR="${LXD_ONE_DIR}" lxc cluster group assign node3 default,foo
 
-  LXD_DIR="${LXD_ONE_DIR}" lxc init testimage c7 --target=@foo
+  echo 'Create c7 on node1 and manually set "volatile.cluster.group" to "foo" to test evacuation respects the group constraint.'
+  LXD_DIR="${LXD_ONE_DIR}" lxc init testimage c7 --target=node1
+  LXD_DIR="${LXD_ONE_DIR}" lxc config set c7 volatile.cluster.group foo
   [ "$(LXD_DIR="${LXD_THREE_DIR}" lxc config get c7 volatile.cluster.group)" = "foo" ]
+  # "volatile.cluster.group" is only checked during scheduling events (creation, migration, evacuation).
+  # Expected: c7 created on node1.
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c sL c7)" = "STOPPED,node1" ]
 
   # For debugging
   LXD_DIR="${LXD_TWO_DIR}" lxc list -c nsL
-
-  echo 'Move c7 to node1 (outside target cluster group "foo") to test "volatile.cluster.group" handling post-evacuation.'
-  # Expected: Instance should be evacuated to node3 (only node in group "foo") during evacuation.
-  LXD_DIR="${LXD_TWO_DIR}" lxc move c7 --target=node1
-  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c sL c7)" = "STOPPED,node1" ]
 
   echo "Evacuate first node"
   LXD_DIR="${LXD_TWO_DIR}" lxc cluster evacuate node1 --force
@@ -3671,7 +3671,7 @@ test_clustering_evacuation() {
   # For debugging
   LXD_DIR="${LXD_TWO_DIR}" lxc list -c nsL
 
-  echo 'Check "volatile.cluster.group" (initial target cluster group) is respected.'
+  echo 'Check c7 respects "volatile.cluster.group" and evacuates to the "foo" group (node3).'
   [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c sL c7)" = "STOPPED,node3" ]
 
   echo "Check status and location of all instances post-evacuation."
@@ -3733,7 +3733,7 @@ test_clustering_evacuation() {
   LXD_DIR="${LXD_TWO_DIR}" lxc cluster restore node1 --force
 
   # For debugging
-  LXD_DIR="${LXD_TWO_DIR}" lxc list
+  LXD_DIR="${LXD_TWO_DIR}" lxc list -c nsL
 
   echo 'Ensure the instances were moved back to the origin'
   [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c sL c1)" = "RUNNING,node1" ]
@@ -3747,6 +3747,13 @@ test_clustering_evacuation() {
   echo 'Move c7 to @default to check "volatile.cluster.group" is updated.'
   LXD_DIR="${LXD_TWO_DIR}" lxc move c7 --target=@default
   [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc config get c7 volatile.cluster.group)" = "default" ]
+
+  echo 'Move c7 to verify "volatile.cluster.group" is cleared when moving to an explicit node outside the group.'
+  LXD_DIR="${LXD_TWO_DIR}" lxc move c7 --target=node1
+
+  echo 'Verify c7 is on the target node and "volatile.cluster.group" is cleared'
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c L c7)" = "node1" ]
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc config get c7 volatile.cluster.group || echo fail)" = "" ]
 
   echo 'Clean up'
   LXD_DIR="${LXD_TWO_DIR}" lxc delete -f c1 c2 c3 c4 c5 c6 c7


lxd/instance_post.go
```diff
@@ -798,14 +798,9 @@ func instancePostClusteringMigrate(ctx context.Context, s *state.State, srcPool
 				return fmt.Errorf("Failed updating cluster member to %q for instance %q: %w", newMember.Name, newInstName, err)
 			}
 
-			id, err := dbCluster.GetInstanceID(ctx, tx.Tx(), projectName, newInstName)
-			if err != nil {
-				return fmt.Errorf("Failed to get ID of moved instance: %w", err)
-			}
-
```
simondeziel: Nice!

test/suites/clustering.sh
```diff
@@ -3550,17 +3550,14 @@ test_clustering_evacuation() {
   LXD_DIR="${LXD_ONE_DIR}" lxc cluster group create foo
   LXD_DIR="${LXD_ONE_DIR}" lxc cluster group assign node3 default,foo
 
-  LXD_DIR="${LXD_ONE_DIR}" lxc init testimage c7 --target=@foo
+  echo 'Create c7 on node1 and manually set "volatile.cluster.group" to @foo to test evacuation respects the group constraint.'
+  LXD_DIR="${LXD_ONE_DIR}" lxc init testimage c7 --target=node1
+  LXD_DIR="${LXD_ONE_DIR}" lxc config set c7 volatile.cluster.group @foo
```
simondeziel: At first, I was wondering if `set`'ing this was immediately relocating the container. I'm assuming it doesn't, based on the previous comment. Worth testing it's still on `node1`?

kadinsayani: `volatile.cluster.group` is only checked during scheduling events, but I think it's worth adding an additional check to prevent any potential regressions. Thanks for suggesting!

kadinsayani: I've added a comment as well :)

lxd/db/cluster/cluster_groups.go
```diff
@@ -95,3 +119,22 @@ WHERE projects_config.key = 'restricted.cluster.groups'`
 
 	return projectURLs, nil
 }
+
+// GetInstancesInClusterGroup returns instances assigned to the specified cluster group.
+// It filters instances by checking the "volatile.cluster.group" configuration key.
+func GetInstancesInClusterGroup(ctx context.Context, tx *sql.Tx, instances []Instance, groupName string) ([]Instance, error) {
+	filteredInstances := []Instance{}
+	for _, instance := range instances {
+		config, err := GetInstanceConfig(ctx, tx, instance.ID)
```
tomponline: This is doing an additional query for each instance in the loop.

Instead of passing around slices of instances retrieved from `GetInstances()` above, I suggest we use the existing `InstanceList()` function which allows a custom function to be passed to it where its called for each instance.

It is more efficient as it performs an initial instance query using `GetInstances` but rather than calling `GetInstanceConfig` for each instance, instead bulk-retrieves all config for those instances in a single query using `InstancesToInstanceArgs`.

In general iterating instances and performing DB queries per instance should be a warning sign.

kadinsayani: Thanks, I've reworked this PR and cleaned up the code to be more aligned with other areas of the codebase. `GetClusterGroupsUsedBy()` is renamed to `GetProjectsUsingRestrictedClusterGroups()`. We now have `clusterGroupUsedBy()` which uses `InstanceList()` and `GetProjectsUsingRestrictedClusterGroups()`. Rather than fill `UsedBy` in `ToAPI()`, we fill it ad-hoc in `GET` request handlers. This is similar to how we set `UsedBy` for projects. I went with this approach to avoid dependency leakage across modules. Alternatively, we could keep `GetClusterGroupUsedBy()` in `lxd/db/cluster` and have one query that handles everything like this: https://github.com/canonical/lxd/blob/396a3391e0b6d62dc5a6ecff25b2eae5c6a64231/lxd/db/cluster/placement_groups.go#L146-L181

I'm fine with either approach.

Additionally, I've renamed `GetClusterGroupMemberInstances()` to `ClusterGroupNodeInUse()` and reworked the query to be more simple and efficient. Rather than getting all instances which is more work than required, we can check the cluster group member for any instance where `volatile.cluster.group` is set to the cluster group name.

kadinsayani:
> Suggest splitting the cluster group in use checks from the migration improvements as separate PRs. ta

https://github.com/canonical/lxd/pull/16787

lxd/api_cluster.go
```diff
@@ -4600,3 +4600,31 @@ func healClusterMember(s *state.State, gateway *cluster.Gateway, op *operations.
...
+	var err error
+
+	usedBy, err = dbCluster.GetProjectsUsingRestrictedClusterGroups(ctx, tx.Tx(), name)
+	if err != nil {
+		return nil, err
+	}
+
+	err = tx.InstanceList(ctx, func(inst db.InstanceArgs, p api.Project) error {
+		// Only consider instances where "volatile.cluster.group" is set to the provided cluster group.
+		val, ok := inst.Config["volatile.cluster.group"]
```
tomponline: whilst correct, this logic seems needlessly verbose, the same could be achieved with:

```go
// Check if instance references cluster group in "volatile.cluster.group" config key.
if inst.Config["volatile.cluster.group"] == instancetype.TargetClusterGroupPrefix+name {
    usedBy = append(usedBy, entity.InstanceURL(inst.Project, inst.Name).String())
}

return nil
```

lxd/api_cluster.go
tomponline, line 1: @kadinsayani whilst reviewing this PR I noticed an issue with the new `volatile.cluster.group` key value.

When we switched from `volatile.cluster.target` to `volatile.cluster.group` I assumed you'd drop the `TargetClusterGroupPrefix` from the value stored in it (because its explicitly a group name now and not a generic target).

But it looks like I missed that in the final review, and I just spotted it now.
Please can you drop storing `TargetClusterGroupPrefix` in the `volatile.cluster.group` key.

Thanks.

lxd/api_cluster.go
```diff
@@ -4186,6 +4198,12 @@ func clusterGroupPatch(d *Daemon, r *http.Request) response.Response {
 			return err
 		}
 
+		usedBy, err := clusterGroupUsedBy(ctx, s, tx, name)
```
tomponline: Do we need to have the overhead of populating `clusterGroup.UsedBy` for PATCH?

lxd/db/node.go
```diff
@@ -684,6 +684,21 @@ func (c *ClusterTx) UpdateNodeClusterGroups(ctx context.Context, id int64, group
...
+			return err
+		}
+
+		inUse, err := c.ClusterGroupNodeInUse(ctx, clusterGroup.Name, nodeInfo.Name)
+		if err != nil {
+			return err
+		}
+
+		if inUse {
+			return errors.New("Cluster group member is currently in use")
```
tomponline: @kadinsayani do we need this check?

Do you see a problem with allowing a member (we don't refer to them as "nodes" anymore)  to be removed from a cluster group if it has an instance on it that was placed in that group.

True, this might cause the instance to be moved during a future "rebalance" operation, but I'm not sure we should prevent the member itself from being removed from a group.

The group still exists, so the instance referencing it is fine.

What do you think?

lxd/api_cluster.go
```diff
@@ -3988,6 +3988,15 @@ func clusterGroupPost(d *Daemon, r *http.Request) response.Response {
 			return fmt.Errorf("Name %q already in use", req.Name)
 		}
 
+		usedBy, err := clusterGroupUsedBy(r.Context(), s, tx, name)
```
tomponline: Some of our "used by" functions take a `firstOnly bool` argument, which would be set true in cases like this, so they can use an optimisation of stopping the search as soon as the first user of the group has been found.

Would be good to use that same technique here in `clusterGroupUsedBy`

lxd/instance_post.go
```diff
@@ -299,6 +300,36 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
 		targetGroupName = target
 	}
 
+	// Unset "volatile.cluster.group" if the instance is manually moved to a node outside of the specified cluster group.
```
tomponline: lets no reference "nodes" in comments please

tomponline: This is interesting.

Whilst I agree we need to have some sort of reset logic for the `volatile.cluster.group` key, I'm somewhat surprised to see this specific logic. Is this what Incus is doing also?

Lets imagine a group called "mygroup" with members: `m1` and `m2`.

If I do:

```
lxc move c1 --target=@mygroup # Sets volatile.cluster.group to "mygroup" and placed on m2.
```

Then do:

```
lxc move c1 --target=m2
```

For the second mode operation, firstly ofc no move should take place, but I am now requesting that c1 should live on `m2`, so despite `m2` being part of `mygroup`, I would expect to see `volatile.cluster.group` removed. Right?

Assuming you agree, then we do need to consider the evacuation scenario, where LXD triggers a move to a different member. But I think we handle that already by checking for if the move request is from a notification right?


The other scenario that is interesting is related to my earlier comment in this PR about whether we should allow removal of a member from a group that still has instances referencing that group on it.

If we do allow such a scenario, then in that case, after removing the instance's cluster member from the group the instance references, running:

```
lxc move c1 --target=@mygroup
```

Should trigger a move to a new member as the instance's current member is no longer in that group.

lxd/api_cluster.go
kadinsayani: Sorry I missed that! I'll fix it with this PR.

lxd/api_cluster.go
kadinsayani: Hmm yeah we can probably skip that.

lxd/api_cluster.go
kadinsayani: Nice, I like that approach.

lxd/db/node.go
kadinsayani: You raise a good point, I think we should actually allow an in use member to be removed from a cluster group. That way, since the cluster group still exists, running `lxc move <instance> --@<group>` will move the instance to a different member in the group. We already have a check ensuring members belong to at least one group, so we'll never end up in a scenario where an instance cannot be scheduled.

lxd/instance_post.go
kadinsayani: > Whilst I agree we need to have some sort of reset logic for the `volatile.cluster.group` key, I'm somewhat surprised to see this specific logic. Is this what Incus is doing also?

Indeed, this is the approach Incus has taken.

> For the second mode operation, firstly ofc no move should take place, but I am now requesting that c1 should live on `m2`, so despite `m2` being part of `mygroup`, I would expect to see `volatile.cluster.group` removed. Right?

Agreed. I'll update this logic to unset `volatile.cluster.group` during any manual targeting event, not just when the member is outside of the referenced cluster group.

> Assuming you agree, then we do need to consider the evacuation scenario, where LXD triggers a move to a different member. But I think we handle that already by checking for if the move request is from a notification right?

Yes, we already handle this:

https://github.com/canonical/lxd/blob/c989f9c996029a78092f6c4bb22def44df214cef/lxd/instances_post.go#L1314-L1317

> The other scenario that is interesting is related to my earlier comment in this PR about whether we should allow removal of a member from a group that still has instances referencing that group on it.
> 
> If we do allow such a scenario, then in that case, after removing the instance's cluster member from the group the instance references, running:
> 
> ```
> lxc move c1 --target=@mygroup
> ```
> 
> Should trigger a move to a new member as the instance's current member is no longer in that group.

Agreed.

lxd/instance/instancetype/instance.go
```diff
@@ -470,13 +470,7 @@ var InstanceConfigKeysAny = map[string]func(value string) error{
...
 	// type: string
 	// shortdesc: The target cluster group
-	"volatile.cluster.group": func(value string) error {
-		if !strings.HasPrefix(value, TargetClusterGroupPrefix) {
-			return fmt.Errorf("Missing %q prefix", TargetClusterGroupPrefix)
-		}
-
-		return nil
-	},
+	"volatile.cluster.group": validate.IsAny,
```
tomponline: we should probably use `validate.Optional()` here, with `clusterGroupValidateName()`.

kadinsayani: I've moved `clusterGroupValidateName()` to `shared/validate` and updated its usage.

lxd/api_cluster.go
```diff
@@ -4559,3 +4559,34 @@ func healClusterMember(s *state.State, gateway *cluster.Gateway, op *operations.
...
+			usedBy = append(usedBy, entity.InstanceURL(inst.Project, inst.Name).String())
+
+			if firstOnly {
+				return db.ErrListStop
+			}
+		}
+
+		return nil
+	})
+	if err != nil {
```
tomponline: you need to check for `err != db.ErrListStop` here like in `VolumeUsedByExclusiveRemoteInstancesWithProfiles` otherwise the `clusterGroupUsedBy` will return an error when `firstOnly == true` and there is >0 entities referencing the cluster group.

lxd/api_cluster.go
```diff
@@ -4559,3 +4559,34 @@ func healClusterMember(s *state.State, gateway *cluster.Gateway, op *operations.
...
 }
+
+// clusterGroupUsedBy returns the list of resource URLs that reference the cluster group.
+// The returned slice contains project URLs (for projects whose "restricted.cluster.groups" configuration includes the group) and instance URLs (for instances whose config contains the group in the "volatile.cluster.group" key).
+// If firstOnly is true then search stops at first result.
+func clusterGroupUsedBy(ctx context.Context, s *state.State, tx *db.ClusterTx, name string, firstOnly bool) ([]string, error) {
+	var usedBy []string
+	var err error
+
+	usedBy, err = dbCluster.GetProjectsUsingRestrictedClusterGroups(ctx, tx.Tx(), name)
```
tomponline, line 4591: if `len(usedBy) >0 && firstOnly` then we should return the first element of `usedBy` here and not proceed to `tx.InstanceList` below.

lxd/instance_post.go
```diff
@@ -300,6 +300,28 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
 		targetGroupName = after
 	}
 
+	// Unset "volatile.cluster.group" if the instance is manually moved to a cluster member.
+	if targetMemberInfo != nil && targetGroupName == "" {
+		err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+			config, err := dbCluster.GetInstanceConfig(ctx, tx.Tx(), inst.ID())
```
tomponline: why do we need to load instance config again here, we already have it via `inst.LocalConfig()` no?

kadinsayani: Yeah, we can use `inst.LocalConfig()`.

lxd/api_cluster.go
kadinsayani: Fixed, thanks.

lxd/api_cluster.go
```diff
@@ -4559,3 +4559,38 @@ func healClusterMember(s *state.State, gateway *cluster.Gateway, op *operations.
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(usedBy) > 0 && firstOnly {
+		return usedBy, nil
```
tomponline, line 4597: > we should return the first element of usedBy 

As it's "first only"

lxd/instance_post.go
```diff
@@ -300,6 +300,23 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
 		targetGroupName = after
 	}
 
+	// Unset "volatile.cluster.group" if the instance is manually moved to a cluster member.
+	if targetMemberInfo != nil && targetGroupName == "" {
+		err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+			if inst.LocalConfig()["volatile.cluster.group"] != "" {
```
tomponline, line 306: We should do this check before starting a transaction as it might not be needed.

tomponline (approved):
Thanks! I will merge this as-is as the behaviour is correct . However there's a couple of minor tweaks for efficiency for a follow up. Cheers

kadinsayani:
> Thanks! I will merge this as-is as the behaviour is correct . However there's a couple of minor tweaks for efficiency for a follow up. Cheers

Thanks for your review! Follow-up here: https://github.com/canonical/lxd/pull/16818. Cheers!

PR #16780, opened 2025-10-20
Ubuntu image caching

simondeziel:
Depends on #16777.


simondeziel:
@tomponline when you get a chance, that one's ready to be reviewed as well now that the dependent PR was merged (thanks!).

PR #16781, opened 2025-10-20
Placement groups

kadinsayani:
This pull request introduces placement group support, including schema updates, a new API, client library methods, and CLI commands. It also includes comprehensive unit and integration tests, new documentation on using placement groups, CLI completions, and forced project deletion integration.

Inspired by https://github.com/canonical/lxd/pull/15408 and includes cherry-picks from https://github.com/markylaing/lxd/tree/placement-groups.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -180,7 +180,7 @@ jobs:
           set -eux
 
           # bin/max (sizes are in MiB)
-          SIZES="lxc 16
+          SIZES="lxc 17
                  lxd-agent 14"
           MIB="$((1024 * 1024))"
 


doc/explanation/clusters.md
```diff
@@ -159,6 +159,15 @@ However, you can control this behavior with the {config:option}`cluster-cluster:
    - The instance is targeted to live on this cluster member.
    - The instance is targeted to live on a member of a cluster group that the cluster member is a part of, and the cluster member has the lowest number of instances compared to the other members of the cluster group.
 
+### Placement groups
```
minaelee, line 163:
```suggestion
(exp-clusters-placement)=
### Placement groups
```

lxd/db/cluster/update.go
```diff
@@ -118,6 +118,30 @@ var updates = map[int]schema.Update{
...
+);
+
+CREATE TABLE placement_groups_config (
+	id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+	placement_group_id INTEGER NOT NULL,
+	key TEXT NOT NULL,
+	value TEXT,
+	UNIQUE (placement_group_id, key),
+	FOREIGN KEY (placement_group_id) REFERENCES placement_groups (id) ON DELETE CASCADE
+);
```
simondeziel, line 142: That's unfortunate there is no way to avoid this duplication :/

minaelee (approved):
Approved for docs part, ty!

tomponline:
@kadinsayani needs a rebase please

lxd/db/cluster/entities_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 func TestEntityStatementValidity(t *testing.T) {
 	schema := Schema()
-	db, err := schema.ExerciseUpdate(71, nil)
+	db, err := schema.ExerciseUpdate(78, nil)
```
tomponline: @kadinsayani is there a way we can have this number be pulled from the patches list?

lxd/db/cluster/placement_groups.go
```diff
@@ -40,3 +53,229 @@ type PlacementGroupFilter struct {
...
 }
+
+// CreatePlacementGroupConfig creates config for a new placement group with the given ID.
+func CreatePlacementGroupConfig(ctx context.Context, tx *sql.Tx, placementGroupID int64, config map[string]string) error {
+	stmt, err := tx.Prepare("INSERT INTO placement_groups_config (placement_group_id, key, value) VALUES(?, ?, ?)")
+	if err != nil {
+		return err
+	}
+
+	defer func() { _ = stmt.Close() }()
```
tomponline: we should be logging warnings in defer/revert functions for errors now.

lxd/db/cluster/placement_groups.go
```diff
@@ -40,3 +53,229 @@ type PlacementGroupFilter struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed finding references to placement group %q: %w", name, err)
+	}
+
+	return urls, nil
+}
+
+// GetAllPlacementGroupUsedByURLs returns a map of project name to map of placement group name to a list of URLs of instances and
+// profiles that reference the placement group in their configuration. If a project is given, used by URLs will only be returned for placement groups in that project.
+func GetAllPlacementGroupUsedByURLs(ctx context.Context, tx *sql.Tx, project *string) (map[string]map[string][]string, error) {
```
tomponline: we should use `projectName string` not `project string` to be explicit and consistent with other functions.

also suggest using a filter struct with pointe fields like some of our other functions to make clear this is a filter, and can easily be extended for other filter fields later.

tomponline: Can we use this function and remove the need for `GetPlacementGroupUsedBy` above, by extending the filtering to support place group name? As there is some duplication by having both of these functions.

lxd/db/cluster/placement_groups.go
```diff
@@ -40,3 +53,229 @@ type PlacementGroupFilter struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed retrieving used by URLs for placement groups: %w", err)
+	}
+
+	return urlMap, nil
+}
+
+// GetPlacementGroupNames returns a map of project name to slice of placement group names. If a project name is provided,
+// only groups in that project are returned. Otherwise, the returned map will contain all projects.
+func GetPlacementGroupNames(ctx context.Context, tx *sql.Tx, project *string) (map[string][]string, error) {
```
tomponline: Lets use a filter struct arg here too please.

lxd/placement_groups.go
```diff
@@ -0,0 +1,783 @@
...
+}
+
+func doPlacementGroupDelete(ctx context.Context, s *state.State, name string, projectName string) error {
+	err := s.DB.Cluster.Transaction(ctx, func(ctx context.Context, tx *db.ClusterTx) error {
+		_, err := cluster.GetPlacementGroupID(ctx, tx.Tx(), name, projectName)
+		if err != nil {
+			return err
+		}
+
+		usedBy, err := cluster.GetPlacementGroupUsedBy(ctx, tx.Tx(), projectName, name)
```
tomponline: GetPlacementGroupUsedBy could do with a `firstOnly bool` argument to allow stopping at first user.

lxd/placement_groups.go
```diff
@@ -0,0 +1,783 @@
...
+
+		err := validator(v)
+		if err != nil {
+			return fmt.Errorf("Invalid placement group key %q value", k)
+		}
+	}
+
+	// Policy is required
+	policy, ok := config["policy"]
+	if !ok || policy == "" {
```
tomponline: The `|| policy == ""` is redundant isn't it, as that would have been caught by the key validator above?
You're only checking for missing keys here right?

lxd/placement_groups.go
```diff
@@ -0,0 +1,783 @@
...
+		return api.StatusErrorf(http.StatusBadRequest, "Missing required config key: %q", "policy")
+	}
+
+	// Rigor is required
+	rigor, ok := config["rigor"]
```
tomponline: To avoid duplication can we do it like this:

```go
for k := range []string{"policy", "rigor"} {
    _, found := config[k]
    if ! found {
       return api.StatusErrorf(http.StatusBadRequest, "Missing required config key: %q", k)
    }
}
```

lxd/placement/filtering.go
```diff
@@ -0,0 +1,243 @@
...
+
+// GetInstancesInPlacementGroup returns a map of instance ID to node ID for all instances that reference the given
+// placement group either directly or indirectly via a profile.
+func GetInstancesInPlacementGroup(ctx context.Context, tx *sql.Tx, project string, placementGroupName string) (map[int]int, error) {
+	// Query notes:
+	// 1. Union of profile config and instance config to perform single query. Since we plan to run this on the leader,
+	//    this may not be necessary.
+	// 2. The apply order must be a selected column, as we need to perform the ORDER BY to expand config correctly, and
+	//    the ORDER BY can only be used after the UNION statement.
+	// 3. The apply_order of 1000000 for instances is used to ensure that instance config is applied last.
```
tomponline: is there "max int" type constant in sqlite (or if not Go) we can use here rather than a special hardcoded value?

lxd/placement/filtering.go
```diff
@@ -0,0 +1,243 @@
...
+	instances_config.value,
+	1000000 AS apply_order
+FROM instances
+	JOIN projects ON instances.project_id = projects.id
+	JOIN instances_config ON instances.id = instances_config.instance_id 
+WHERE projects.name = ? AND instances_config.key = 'placement.group'
+	ORDER BY instances.id, apply_order`
+	args := []any{project, project}
+
+	// Keep a map of pointers so that each value is mutable.
```
tomponline: this comment doesn't say "why", and also the two maps defined after it are not maps of pointers, so doubly confusing.

lxd/placement/filtering.go
```diff
@@ -0,0 +1,243 @@
...
+
+	"github.com/canonical/lxd/lxd/db"
+	"github.com/canonical/lxd/lxd/db/cluster"
+	"github.com/canonical/lxd/lxd/db/query"
+	"github.com/canonical/lxd/shared/api"
+)
+
+// GetInstancesInPlacementGroup returns a map of instance ID to node ID for all instances that reference the given
+// placement group either directly or indirectly via a profile.
+func GetInstancesInPlacementGroup(ctx context.Context, tx *sql.Tx, project string, placementGroupName string) (map[int]int, error) {
```
tomponline: `projectName string` to stick with convention.

lxd/placement/filtering.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	return result, nil
+}
+
+// Filter filters the candidates using the placement group. If the excludeInstanceIDs argument is passed, those instances
+// are ignored (this is used to check if a running instance satisfies its placement group policy, or during evacuation).
+func Filter(ctx context.Context, tx *sql.Tx, candidates []db.NodeInfo, excludeInstanceIDs []int, placementGroup cluster.PlacementGroup) ([]db.NodeInfo, error) {
+	// Get the placement group config.
+	apiPlacementGroup, err := placementGroup.ToAPI(ctx, tx)
```
tomponline: rather than calling `ToAPI` here, I'd prefer to see the `apiPlacementGroup` passed in as an argument, this way it makes clear what the function needs to operate, and the caller is left to convert `placementGroup` to `apiPlacementGroup`, this way it can avoid needing to do this multiple times if calling `Filter` multiples times.

lxd/placement/filtering.go
```diff
@@ -0,0 +1,243 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Query to get node ID from node name.
+	sourceNodeID := tx.GetNodeID()
+
+	// Build a list of instance IDs on the source node to exclude from placement filtering.
+	var excludeInstances []int
+	for instID, nodeID := range instToNode {
```
tomponline: Can we add a filter struct argument to allow excluding instances on a specific member to `GetInstancesInPlacementGroup` and push this filtering down into the database query?

lxd/instances_post.go
```diff
@@ -1299,10 +1300,41 @@ func instancesPost(d *Daemon, r *http.Request) response.Response {
 	}
 
 	if s.ServerClustered && !clusterNotification && targetMemberInfo == nil {
+		var globalConfigDump map[string]string
+		if s.GlobalConfig != nil {
```
tomponline: why would s.GlobalConfig by nil?

lxd/instances_post.go
```diff
@@ -1299,10 +1300,41 @@ func instancesPost(d *Daemon, r *http.Request) response.Response {
...
+
+		expandedConfig := instancetype.ExpandInstanceConfig(globalConfigDump, req.Config, profiles)
+
 		// If no target member was selected yet, pick the member with the least number of instances.
 		if targetMemberInfo == nil {
 			err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
-				targetMemberInfo, err = tx.GetNodeWithLeastInstances(ctx, candidateMembers)
+				// Check if instance is using a placement group.
+				placementGroupName, ok := expandedConfig["placement.group"]
+				if !ok {
```
tomponline: `if placementGroupName == "" {` would be more explicit here in my view (as its a belt+braces approach).

tomponline: @kadinsayani is there anything stopping us moving the contents of this transaction up into the transaction above where `candidateMembers` are loaded?

Also, we should probably be calling `instancetype.ValidName` higher up in that transaction too, before we do all of the expensive placement queries.

lxd/db/cluster/update.go
tomponline: what duplication?

simondeziel: The same is already in the `schema.go` definition.

tomponline: one is the patch to add the table and the other is resulting output for fresh schema (built from the patches when running `make update-schema`).

simondeziel: I know but that's still copied and pasted bits which is unfortunate but probably not avoidable if it's been that way since day one.

tomponline: Its not manually copied/pasted, and its quite useful to be able to see what a fresh schema would be without needing for it to be computed at run time.

simondeziel: Ah thanks, I was assuming this was manually copied. All good then, thanks for clarifying.

lxd/db/cluster/entities_test.go
kadinsayani: We can use this: https://github.com/canonical/lxd/blob/c853f78bdeacec2ba1891abab851c8cbe09dbdcd/lxd/db/cluster/update.go#L40-L41

lxd/placement/filtering.go
kadinsayani: [`math.MaxInt64`](https://cs.opensource.google/go/go/+/refs/tags/go1.25.3:src/math/const.go;drc=68370187fafe9c355e489d6976c35cd908f37a00;l=50) should work well.

lxd/placement_groups.go
kadinsayani: Correct, that extra check is redundant.

lxd/placement/filtering.go
kadinsayani: I'll add a better comment here.

lxd/placement/filtering.go
kadinsayani: Thanks for the suggestion, I like that approach :)

lxd/db/cluster/placement_groups.go
kadinsayani: Yeah, we can just have one function with a filter struct. Thanks for the suggestion!

lxd/db/cluster/placement_groups.go
kadinsayani: We can actually drop this function and just use `GetPlacementGroups()` with `PlacementGroupFilter{Project: projectFilter}`.

lxd/instances_post.go
kadinsayani: I don't think it would ever be nil at this point. @markylaing any particular reason why this check was included in https://github.com/markylaing/lxd/commit/8328351109179e247fa7172a99f5af71220ddf9d?

lxd/placement_groups.go
kadinsayani: Great idea, I've added that.

lxd/placement/filtering.go
kadinsayani: I've moved `GetInstancesInPlacementGroup()` to `lxd/db/cluster/placement_group.go` and utilized the new `PlacementFilter` struct. Additionally, I've pushed the filtering logic down and simplified the query.

lxd/instances_post.go
kadinsayani: > @kadinsayani is there anything stopping us moving the contents of this transaction up into the transaction above where `candidateMembers` are loaded?

No, we can move it up :)

> Also, we should probably be calling `instancetype.ValidName` higher up in that transaction too, before we do all of the expensive placement queries.

Done :white_check_mark:

lxd/instances_post.go
kadinsayani: Resolving for now, feel free to re-open :)

lxd/placement/filtering.go
kadinsayani: I also got rid of `FilterForEvacuation()` and added an `evacuation` bool arg to `Filter()` instead. Additionally, I've extracted a `getCompliantMembers()` helper function to encapsulate the placement policy and rigor logic.

```Go
// Filter filters the provided slice of candidate cluster members using the provided [api.PlacementGroup].
func Filter(ctx context.Context, tx *db.ClusterTx, candidates []db.NodeInfo, apiPlacementGroup api.PlacementGroup, evacuation bool) ([]db.NodeInfo, error)
```

```Go
// getCompliantMembers gets compliant cluster members from the provided candidates based on the given placement policy and rigor.
func getCompliantMembers(policy string, rigor string, candidates []db.NodeInfo, memberToInst map[int][]int) ([]db.NodeInfo, error) {
```

I think this is much cleaner and easier to work with now :)

lxd/db/cluster/placement_groups.go
kadinsayani: I was able to de-dupe quite a bit thanks to your suggestion! The db functions and their associated queries are looking much cleaner.

lxd/db/cluster/placement_groups.go
```diff
@@ -41,3 +55,211 @@ type PlacementGroupFilter struct {
...
+// ToAPI converts the [PlacementGroup] to an [api.PlacementGroup], querying for extra data as necessary.
+func (p *PlacementGroup) ToAPI(ctx context.Context, tx *sql.Tx) (*api.PlacementGroup, error) {
+	// Get config
+	config, err := GetPlacementGroupConfig(ctx, tx, p.ID)
+	if err != nil {
+		return nil, fmt.Errorf("Failed getting placement group config: %w", err)
+	}
+
+	// Get used by
+	usedBy, err := GetPlacementGroupUsedBy(ctx, tx, PlacementGroupFilter{Project: &p.Project, Name: &p.Name}, false)
```
tomponline, line 143: @kadinsayani based on what you did for https://github.com/canonical/lxd/pull/16779 should we be populating `UsedBy` here?

lxd/instances_post.go
tomponline, line 1: There's a fair bit of duplication in these changes, can you look at de-duping this by breaking out into smaller functions please?

lxd/api_cluster.go
```diff
@@ -4405,9 +4396,61 @@ func evacuateClusterSelectTarget(ctx context.Context, s *state.State, inst insta
 			return err
 		}
 
+		if ok {
+			// Clear "volatile.cluster.group".
+			if clusterGroupName != "" {
+				err = tx.DeleteInstanceConfigKey(ctx, int64(inst.ID()), "volatile.cluster.group")
```
tomponline, line 4402: I'm a bit confused why this is happening in `evacuateClusterSelectTarget`?

Shouldn't this be wiped when the instance is assigned to the placement group rather than the next time its evacuated?

lxd/api_cluster.go
```diff
@@ -4405,9 +4396,61 @@ func evacuateClusterSelectTarget(ctx context.Context, s *state.State, inst insta
...
+				if err != nil {
+					return fmt.Errorf(`Failed removing "volatile.cluster.group" config key: %w`, err)
+				}
+			}
+
+			// Filter candidates by placement group.
+			placementGroup, err := dbCluster.GetPlacementGroup(ctx, tx.Tx(), placementGroupName, inst.Project().Name)
```
tomponline, line 4409: as an optimisation, we could look at pre-loading placement group configs rather than loading potentially the same data repeatedly if multiple instances are in the same group on the member being evacuated.

lxd/db/cluster/placement_groups.go
kadinsayani: We can skip populating `UsedBy` here.

lxd/instances_post.go
kadinsayani: Certainly, thanks for pointing that out :)

lxd/api_cluster.go
kadinsayani: Yes, we can move this clearing logic to `(lxc/qemu).Update()`.

lxd/api_cluster.go
kadinsayani: I've added a placement group cache for this, similar to what we have for storage pools in `storage/cache.go`.

kadinsayani:
> Thanks!
> 
> LGTM, there will need to be some follow ups.
> 
> And @markylaing and @minaelee may yet have comments on this.

Thanks @tomponline for your review! Follow-up here: https://github.com/canonical/lxd/pull/16841.

kadinsayani:
> Absolutely smashing work @kadinsayani well done!

Thanks @markylaing! This feature wouldn’t have come together without all the initial work you put in 🎉

PR #16782, opened 2025-10-21
Network: Extend target IP address validation for network forwards and load balancer backends (from Incus)

nmezhenskyi:
This PR extends target IP address validation for network forwards and load balancer backends to check that the target address is not a network address and not a broadcast address.

Inspired by https://github.com/lxc/incus/pull/2555.


simondeziel (approved):
LGTM thanks!

PR #16786, opened 2025-10-21
Docs: Add how-to for recovery using storage replication

roosterfish:
Follow up on the `lxd recover` story to properly document the required steps to recover workload when using storage replication.

This PR also extends the storage driver support matrix with another line for volume recovery which is not yet present for the Alletra driver.

### DIFF
--- a/doc/production-setup.md
+++ b/doc/production-setup.md
@@ -34,6 +34,7 @@ How to back up your server and recover from failure:
 
 :diataxis:Back up a server </backup>
 :diataxis:Recover instances </howto/disaster_recovery>
+:diataxis:Recover with storage replication </howto/disaster_recovery_replication>
 ```
 
 ## Related topics


lxd/storage/drivers/driver_ceph.go
```diff
@@ -332,7 +332,14 @@ func (d *ceph) Validate(config map[string]string) error {
...
-		"ceph.cluster_name":    validate.IsAny,
+		"ceph.cluster_name": validate.IsAny,
+		// lxdmeta:generate(entities=storage-ceph; group=pool-conf; key=ceph.osd.force_reuse)
+		//
+		// ---
+		//  type: bool
+		//  defaultdesc: `false`
+		//  shortdesc: Whether to allow reuse of an existing OSD storage pool already used by LXD
+		//  scope: global
 		"ceph.osd.force_reuse": validate.Optional(validate.IsBool), // Deprecated, should not be used.
```
tomponline, line 343: @roosterfish this key is deprecated, should we really be adding it to the docs?

roosterfish: The idea is to replace it with the general `source.reuse` setting once we have it. Today it's really the only option to allow LXD reusing this pool.

Another option would be to instruct the user to remove the placeholder vol in case it got accidentally replicated but I found this option more robust.

tomponline: ok will include for now then

doc/howto/disaster_recovery_replication.md
```diff
@@ -0,0 +1,147 @@
...
+
+```{note}
+There is still the possibility for collisions in the network in case the primary location comes back up and LXD starts up any instances.
+This is out of scope for storage replication but proper measures have to be put into place to prevent conflicts.
+```
+
+Switching back service to the primary location can be performed in two ways. For both of them the operations on the storage layer are identical
+but choosing the right option depends on what has happened to the instances and custom volumes on the secondary location:
+
+1. Shutdown the resources on the primary location and bring them back up on the primary
```
mionaalex: Should this be "shutdown the resources on the **secondary** location and bring them back up on the primary"?

roosterfish: Good catch thanks, fixed.

tomponline (approved):

PR #16795, opened 2025-10-22
Coverage fixes

simondeziel:


test/includes/lxd.sh
```diff
@@ -518,15 +518,22 @@ coverage_enabled() {
...
   coverage_enabled || return 0
 
   local instance="${1}"
   echo "==> Setting up LXD agent coverage gathering inside the ${instance} VM"
 
-  # Mount the host's GOCOVERDIR into the instance.
-  lxc config device add "${instance}" gocoverdir disk source="${GOCOVERDIR}" path="${GOCOVERDIR}"
+  # Ubuntu 22.04 currently requires running with `migration.stateful=true` to
+  # avoid incompatibilities with `virtiofsd` and force a fallback to `9p`. This
+  # in turn prevents mounting host directories into the instance. As a
```
tomponline, line 529: does the guest kernel not support 9p?

simondeziel: Yes, `9p` is supported by the guest but LXD doesn't seem to make any distinction between `9p` and `virtiofs` shares: https://github.com/canonical/lxd/blob/main/lxd/device/disk.go#L655

tomponline: ack, i see, probably safer, especially as we are likely removing most 9p support soon anyway.

PR #16796, opened 2025-10-22
Run `lxd-ci` tests with support for collecting coverage data

simondeziel:


.github/workflows/tests.yml
```diff
@@ -411,6 +427,279 @@ jobs:
           # Never allow public SSH access
           limit-access-to-actor: true
 
+  snap-tests:
+    if: github.event_name != 'schedule' || github.repository == 'canonical/lxd'
```
tomponline, line 431: if we dont run on schedule how will this help with code coverage collection?

tomponline (approved):

tomponline:
Thanks!

.github/workflows/tests.yml
simondeziel: It's the same condition as the `system-tests` one. It won't run on a schedule **unless** it's the `canonical/lxd` repo.

This is what I could come up with to avoid schedule runs on forks while still keeping them for the canonical repo.

tomponline: Ah yes I must've misread it as "and" 

Thanks

PR #16803, opened 2025-10-23
doc: Fix storage driver ref names in CSI docs

MusicDin:
Adds custom names for references to make them match the text.

Before:

image

After:

image

### DIFF
--- a/doc/reference/storage_drivers.md
+++ b/doc/reference/storage_drivers.md
@@ -11,7 +11,7 @@ Storage drivers are divided into subgroups based on their accessibility.
 See the corresponding pages for driver-specific information and configuration options.
 
 (storage-drivers-local)=
-## Local volumes
+## Local
 
 A local volume resides on the storage pool of a single LXD server and is only accessible to instances running on that server.
 In a cluster, other members cannot access local volumes directly.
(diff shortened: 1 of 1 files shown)


tomponline (dismissed):

tomponline:
@MusicDin why are those driver headings labelled as "volumes" in the first place though?

MusicDin:
The ref name is taken from the referenced title. Please, take a look at titles in https://documentation.ubuntu.com/lxd/latest/reference/storage_drivers/ - they contain `* volumes`.

Should I also change the titles to `* drivers` / `* storage` instead of `* volumes`?

tomponline:
> The ref name is taken from the referenced title. Please, take a look at titles in https://documentation.ubuntu.com/lxd/latest/reference/storage_drivers/ - they contain `* volumes`.
> 
> Should I also change the titles to `* drivers` / `* storage` instead of `* volumes`?

suggest just drop "volumes" from the heading

tomponline (commented):
do we need the override of labels on the csi.md page now?

MusicDin:
> do we need the override of labels on the csi.md page now?

I would prefer them to ensure they are lower cased, but could drop the overrides in tables

tomponline:
> > do we need the override of labels on the csi.md page now?
> 
> I would prefer them to ensure they are lower cased, but could drop the overrides in tables

I suppose if they are not lower cased it shows they are a concept.

PR #16806, opened 2025-10-23
Fix segfault in test_clustering_heal_networks_stop()

kadinsayani:


tomponline:
> Isn't that mostly undoing [a7085d8](https://github.com/canonical/lxd/commit/a7085d8c9c4c37b1f69098d7a0fc24a8c16b1e3a) where LXD was intentionally abruptly killed?

Thats a good point. Do we know why killing lxd is generating a core dump?

kadinsayani:
I took a closer look at `kill_lxd()` and _only_ when the `unix.socket` exists we run:

```
timeout -k 30 30 lxd shutdown || kill -9 "${LXD_PID}" 2>/dev/null || true
```

Additionally, `spawn_lxd()` runs the following:

```
 nsenter -n -m -t "${pid}" lxd --logfile "${LXD_DIR}/lxd.log" "${SERVER_DEBUG-}" "$@" 2>&1 &
```

I believe the segfault occurred because the test deleted `${LXD_TWO_DIR}/unix.socket` too early, preventing `kill_lxd()` from performing a clean shutdown.

Still investigating...

kadinsayani:
`kill_lxd()` isn't killing the LXD daemon process when `${LXD_TWO_DIR}/unix.socket` is removed too early.

kadinsayani:
I've reworked the test to just kill the LXD process and save cleanup for the end of the test. similar to what we do in `test_clustering_rebalance()`.

simondeziel:
> `kill_lxd()` isn't killing the LXD daemon process when `${LXD_TWO_DIR}/unix.socket` is removed too early.

Indeed, if there is no `unix.socket` when calling `kill_lxd()`, `lxd shutdown || kill -9` is not done, presumably because an orderly shutdown of LXD cleans up the socket. Maybe `kill_lxd()` should still check if the found PID exist though and kill it hard.

I'm still not clear as to where the coredumps come from though :/

test/suites/clustering.sh
```diff
@@ -1409,8 +1409,7 @@ test_clustering_heal_networks_stop() {
   LXD_DIR="${LXD_ONE_DIR}" lxc config set cluster.healing_threshold 11
 
   echo "Kill node2 (a non-leader member)"
-  rm -f "${LXD_TWO_DIR}/unix.socket"
-  kill_lxd "${LXD_TWO_DIR}"
+  kill -9 "$(< "${LXD_TWO_DIR}/lxd.pid")"
```
simondeziel, line 1412: Sounds like a clean and easy way to simulate a hard crash.

simondeziel (approved):

kadinsayani:
> Indeed, if there is no `unix.socket` when calling `kill_lxd()`, `lxd shutdown || kill -9` is not done, presumably because an orderly shutdown of LXD cleans up the socket. Maybe `kill_lxd()` should still check if the found PID exist though and kill it hard.

Perhaps, but I'd rather leave that change for another day as we have quite a few `kill_lxd()` calls and red CI is scarier than halloween :laughing: 
 
> I'm still not clear as to where the coredumps come from though :/

Perhaps the background process is still trying to write to the log file?

simondeziel:
> > I'm still not clear as to where the coredumps come from though :/
> 
> Perhaps the background process is still trying to write to the log file?

Writing to a deleted file is will still work for as long as the FD is valid. It will keep the filesystem object alive although mostly "invisible" to most tools.

I've not looked at the dump closely as they are not entirely fun to explore with all the libs that need to be obtained for GDB to be happy. But maybe you were more persistent and got a backtrace?

kadinsayani:
> Writing to a deleted file is will still work for as long as the FD is valid. It will keep the filesystem object alive although mostly "invisible" to most tools.

Ah yes.

> I've not looked at the dump closely as they are not entirely fun to explore with all the libs that need to be obtained for GDB to be happy. But maybe you were more persistent and got a backtrace?

I was able to get a backtrace in `gdb` after compiling dqlite and LXD with debug symbols:

```
(gdb) bt
#0  runtime.raise () at runtime/sys_linux_amd64.s:154
#1  0x0000000000484f0d in runtime.raisebadsignal (sig=11, c=0x7a6bcf7f9550) at runtime/signal_unix.go:1038
#2  0x0000000000485385 in runtime.badsignal (sig=11, c=0x7a6bcf7f9550) at runtime/signal_unix.go:1147
#3  0x0000000000483ef2 in runtime.sigtrampgo (sig=11, info=0x7a6bcf7f96f0, ctx=0x7a6bcf7f95c0) at runtime/signal_unix.go:468
#4  0x00000000004af586 in runtime.sigtramp () at runtime/sys_linux_amd64.s:352
#5  <signal handler called>
#6  0x00007a6c1fcadd55 in __GI___libc_free (mem=0x1) at ./malloc/malloc.c:3375
#7  0x00007a6c1feede54 in convertClearFollower (r=0x7a6bc4003d50) at src/raft/convert.c:49
#8  convertClear (r=0x7a6bc4003d50) at src/raft/convert.c:134
#9  0x00007a6c1feee37f in convertToUnavailable (r=0x7a6bc4003d50) at src/raft/convert.c:273
#10 0x00007a6c1ff045e4 in uvFireRecvCb (s=s@entry=0x7a6ba0029570) at src/raft/uv_recv.c:190
#11 0x00007a6c1ff04800 in uvServerReadCb (stream=<optimized out>, nread=<optimized out>, buf=<optimized out>) at src/raft/uv_recv.c:309
#12 0x00007a6c1fb04cab in ?? () from /lib/x86_64-linux-gnu/libuv.so.1
#13 0x00007a6c1fb05028 in ?? () from /lib/x86_64-linux-gnu/libuv.so.1
#14 0x00007a6c1fb0abdb in ?? () from /lib/x86_64-linux-gnu/libuv.so.1
#15 0x00007a6c1faf1ce8 in uv_run () from /lib/x86_64-linux-gnu/libuv.so.1
--Type <RET> for more, q to quit, c to continue without paging--
#16 0x00007a6c1fee16d8 in taskRun (d=0x7a6bc4003018) at src/server.c:711
#17 taskStart (arg=0x7a6bc4003018) at src/server.c:774
#18 0x00007a6c1fc9caa4 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:447
#19 0x00007a6c1fd29c6c in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
(gdb) frame 7
#7  0x00007a6c1feede54 in convertClearFollower (r=0x7a6bc4003d50) at src/raft/convert.c:49
49                      raft_free(r->follower_state.current_leader.address);
(gdb) info local
__func__ = "convertClearFollower"
(gdb) info locals
__func__ = "convertClearFollower"
(gdb) p r
$1 = (struct raft *) 0x7a6bc4003d50
(gdb) p r->follower_state.current_leader.address
$2 = 0x1 <error: Cannot access memory at address 0x1>
(gdb) p r->follower_state
$3 = {randomized_election_timeout = 3942, current_leader = {id = 0, address = 0x1 <error: Cannot access memory at address 0x1>}, append_in_flight_count = 0, reserved = {0, 0, 0, 0, 0, 0, 0}}
```

I also noticed this appearing in the logs:

```
ERROR  [2025-10-23T20:07:54Z] Error updating raft members                   err="attempt to write a readonly database"
```

Definitely a dqlite issue. It looks like the crash cause is the `free(0x1)` in `convertClearFollower`. I suspect `current_leader.address` is corrupted.

I didn't go any further than this to avoid sinking time.

simondeziel:
> I didn't go any further than this to avoid sinking time.

Indeed, that's already very nice findings. If you could put the coredump and the debug enabled libs/binaries into a tarball attached to a dqlite issue, I think that'd be a very good starting point for them to have a close look. Thanks!

kadinsayani:
> > I didn't go any further than this to avoid sinking time.
> 
> Indeed, that's already very nice findings. If you could put the coredump and the debug enabled libs/binaries into a tarball attached to a dqlite issue, I think that'd be a very good starting point for them to have a close look. Thanks!

Np! TIL GitHub has an upload limit of 25MB for attachments, and they only allow `.gz`, `.tgz`, and `.zip` archives so other compression types aren't an option :(

simondeziel:
> Np! TIL GitHub has an upload limit of 25MB for attachments, and they only allow `.gz`, `.tgz`, and `.zip` archives so other compression types aren't an option :(

Maybe a link to Google Drive?

kadinsayani:
> > Np! TIL GitHub has an upload limit of 25MB for attachments, and they only allow `.gz`, `.tgz`, and `.zip` archives so other compression types aren't an option :(
> 
> Maybe a link to Google Drive?

I'll add a note letting them know I can provide the binaries upon request. Looking at the code, it should be pretty easy to reproduce:

https://github.com/canonical/dqlite/blob/a7dbde52b7f3a8ecf8a68982df7bd5a0f163c3ed/src/raft/convert.c#L43-L52

These are the type of bugs Rust shines at preventing ;)

tomponline (approved):
Thanks!

simondeziel:
Thanks for filling https://github.com/canonical/dqlite/issues/857

PR #16809, opened 2025-10-23
Minimize snap system tests

simondeziel:
Adapt the `console_vm` test to work outside of the `snap` and test it a bit more.

### DIFF
--- a/test/main.sh
+++ b/test/main.sh
@@ -593,6 +593,7 @@ if [ "${1:-"all"}" != "snap" ] && [ "${1:-"all"}" != "cluster" ]; then
     run_test test_resources_bcache "resources bcache"
     run_test test_kernel_limits "kernel limits"
     run_test test_console "console"
+    run_test test_console_vm "console VM"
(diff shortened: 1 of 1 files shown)


simondeziel:
@tomponline I tried doing `snap_storage_volume_attach_vm` while at it but there the limitation with `migration.stateful=true` is really problematic:

```
+ lxc storage volume attach lxdtest-SLq vol1 v1      
+ timeout --foreground 120 /root/go/bin/lxc storage volume attach lxdtest-SLq vol1 v1 --verbose
Error: Failed add validation for device "vol1": Only additional disks coming from a shared storage pool are supported with migration.stateful=true
+ cleanup
```

I think this one is best kept until we deal with the `virtiofsd` issue proper (or move to 24.04 runners).

tomponline:
> I think this one is best kept until we deal with the `virtiofsd` issue proper (or move to 24.04 runners).

Fine by me.

tomponline (approved):
Thanks!

PR #16813, opened 2025-10-24
Check project existence when listing entities

### LINKED ISSUE #16802: Resource list should return 404 if project does not exist
Should also return 404 if project does exist but caller does not have `can_view` for it

@markylaing there looks like still an inconsistency between `lxc list`, `lxc image list`, `lxc operation list` and other listing commands that list entities belonging to projects.

e.g.

```
lxc list --project=a-non-existent-project
<no error, blank output>

lxc image list lxc list --project=a-non-existent-project
+-------+-------------+--------+-------------+--------------+------+------+-------------+
| ALIAS | FINGERPRINT | PUBLIC | DESCRIPTION | ARCHITECTURE | TYPE | SIZE | UPLOAD DATE |
+-------+-------------+--------+-------------+--------------+------+------+-------------+
<no error, blank output>

lxc operation list --project=a-non-existent-project
+----+------+-------------+--------+------------+---------+
| ID | TYPE | DESCRIPTION | STATUS | CANCELABLE | CREATED |
+----+------+-------------+--------+------------+---------+
<no error, blank output>
```
_Also there's an unrelated inconsistency in the CLI display output of these commands when there is no results, 2 of them display an empty table, but `lxc list` displays nothing_

Compared to:
```
lxc network list  --project=a-non-existent-project
Error: Failed loading project "a-non-existent-project": Project not found

lxc profile list --project=a-non-existent-project
Error: Failed loading project "a-non-existent-project": Project not found

lxc network acl list --project=a-non-existent-project
Error: Failed loading project "a-non-existent-project": Project not found
```

I understood we are wanting all such listing commands to exhibit the latter behaviour (returning an error) rather than no-error and an empty list (unless called with `--all-projects` flag).

Should we log a new issue for this?

cc @escabo @simondeziel

_Originally posted by @tomponline in https://github.com/canonical/lxd/issues/16547#issuecomment-3396171025_

markylaing:
#16547 Enforced that fine-grained identities are able to view a project in order to list entities contained within said project. 

However, the following behaviour was noted by @escabo as a local (unix socket) user:
```
$ lxc list --project nonexistent
+------+-------+------+------+------+-----------+
| NAME | STATE | IPV4 | IPV6 | TYPE | SNAPSHOTS |
+------+-------+------+------+------+-----------+
$ lxc operation list --project nonexistent
+----+------+-------------+--------+------------+---------+
| ID | TYPE | DESCRIPTION | STATUS | CANCELABLE | CREATED |
+----+------+-------------+--------+------------+---------+
$ lxc warning list --project nonexistent
+------+------+--------+----------+-------+---------+-----------+
| UUID | TYPE | STATUS | SEVERITY | COUNT | PROJECT | LAST SEEN |
+------+------+--------+----------+-------+---------+-----------+
$ lxc image list --project nonexistent
+-------+-------------+--------+-------------+--------------+------+------+-------------+
| ALIAS | FINGERPRINT | PUBLIC | DESCRIPTION | ARCHITECTURE | TYPE | SIZE | UPLOAD DATE |
+-------+-------------+--------+-------------+--------------+------+------+-------------+
$ lxc image alias list --project nonexistent
+-------+-------------+------+-------------+
| ALIAS | FINGERPRINT | TYPE | DESCRIPTION |
+-------+-------------+------+-------------+
```

These examples fail if authenticated as a restricted TLS client or as a fine-grained identity, because the authorization check for `can_view` on the project returns a 404 if the project doesn't exist. As a unix or unrestricted TLS identity however, the authorizer just allows the request to continue, and it is up to the API handlers themselves to perform appropriate checks for project existence.

This PR updates these handlers to check if the requested project exists. Note that if the caller is a fine-grained identity, then the project ID is effectively being looked up twice. This can be optimised when the openfga authorizer uses IDs instead of URLs. 

Also note that the fix for the image and image alias list handlers was quite subtle. This was caused by a regression introduced in https://github.com/canonical/lxd/pull/15988 whereby the `GetProjectConfig` function did not return a 404 not found error if the project does not exist. This caused all nonexistent projects to be treated as if they were the default project:
```
$ lxc project list
+-------------------+--------+----------+-----------------+-----------------+----------+---------------+---------------------+---------+
|       NAME        | IMAGES | PROFILES | STORAGE VOLUMES | STORAGE BUCKETS | NETWORKS | NETWORK ZONES |     DESCRIPTION     | USED BY |
+-------------------+--------+----------+-----------------+-----------------+----------+---------------+---------------------+---------+
| default (current) | YES    | YES      | YES             | YES             | YES      | YES           | Default LXD project | 3       |
+-------------------+--------+----------+-----------------+-----------------+----------+---------------+---------------------+---------+
$ lxc image list --project default
+-------+--------------+--------+-----------------------------------+--------------+-----------+---------+-------------------------------+
| ALIAS | FINGERPRINT  | PUBLIC |            DESCRIPTION            | ARCHITECTURE |   TYPE    |  SIZE   |          UPLOAD DATE          |
+-------+--------------+--------+-----------------------------------+--------------+-----------+---------+-------------------------------+
|       | 7d4a2be617fb | no     | Alpine edge amd64 (20251024_0050) | x86_64       | CONTAINER | 3.55MiB | Oct 24, 2025 at 11:50am (UTC) |
+-------+--------------+--------+-----------------------------------+--------------+-----------+---------+-------------------------------+
$ lxc image list --project none
+-------+--------------+--------+-----------------------------------+--------------+-----------+---------+-------------------------------+
| ALIAS | FINGERPRINT  | PUBLIC |            DESCRIPTION            | ARCHITECTURE |   TYPE    |  SIZE   |          UPLOAD DATE          |
+-------+--------------+--------+-----------------------------------+--------------+-----------+---------+-------------------------------+
|       | 7d4a2be617fb | no     | Alpine edge amd64 (20251024_0050) | x86_64       | CONTAINER | 3.55MiB | Oct 24, 2025 at 11:50am (UTC) |
+-------+--------------+--------+-----------------------------------+--------------+-----------+---------+-------------------------------+
```

See the commit message for more details.

Closes #16802

### DIFF
--- a/test/suites/projects.sh
+++ b/test/suites/projects.sh
@@ -810,7 +810,8 @@ test_projects_limits() {
   lxc profile device set default root size=100MiB
   lxc config device add c2 root disk path="/" pool="${pool}" size=50MiB
 
-  if [ "${LXD_BACKEND}" = "lvm" ]; then
+  lxd_backend=$(storage_backend "$LXD_DIR")
+  if [ "${lxd_backend}" = "lvm" ]; then
     # Can't set the project's disk limit because not all volumes have
     # the "size" config defined.
     pool1="lxdtest1-$(basename "${LXD_DIR}")"
@@ -860,7 +861,7 @@ test_projects_limits() {
   # Run the following part of the test only against the dir or zfs backend,
   # since it on other backends it requires resize the rootfs to a value which is
   # too small for resize2fs.
-  if [ "${LXD_BACKEND}" = "dir" ] || [ "${LXD_BACKEND}" = "zfs" ]; then
+  if [ "${lxd_backend}" = "dir" ] || [ "${lxd_backend}" = "zfs" ]; then
     # Add a remote LXD to be used as image server.
     local LXD_REMOTE_DIR
     LXD_REMOTE_DIR=$(mktemp -d -p "${TEST_DIR}" XXX)
@@ -926,7 +927,7 @@ test_projects_limits() {
   lxc project switch default
   lxc project delete p1
 
-  if [ "${LXD_BACKEND}" = "dir" ] || [ "${LXD_BACKEND}" = "zfs" ]; then
+  if [ "${lxd_backend}" = "dir" ] || [ "${lxd_backend}" = "zfs" ]; then
     lxc remote remove l2
     kill_lxd "$LXD_REMOTE_DIR"
   fi
@@ -1350,7 +1351,8 @@ test_projects_images_volume() {
 
   lxc project create foo
 
-  if [ "${LXD_BACKEND}" = "ceph" ]; then
+  lxd_backend=$(storage_backend "$LXD_DIR")
+  if [ "${lxd_backend}" = "ceph" ]; then
     # This won't work on ceph because it's not a multi-node storage
     ! lxc config set storage.project.foo.images_volume="${pool}/vol" || false
     # Clean up
@@ -1431,7 +1433,8 @@ test_projects_backups_volume() {
 
   lxc project create foo
 
-  if [ "${LXD_BACKEND}" = "ceph" ]; then
+  lxd_backend=$(storage_backend "$LXD_DIR")
+  if [ "${lxd_backend}" = "ceph" ]; then
     # This won't work on ceph because it's not a multi-node storage
     ! lxc config set storage.project.foo.backups_volume="${pool}/vol" || false
     # Clean up


test/suites/image.sh
```diff
@@ -297,21 +297,32 @@ run_images_public() {
   query /1.0/images?recursion=1\&project=default | jq --exit-status '(.metadata | length) == 0 and .status_code == 200'
```
simondeziel, line 297: Those tests above are making sure that every kind of auth will seen an empty list with a 200 when there are not public images.

Would it be worth adding a check that they can all list images that are indeed public? Maybe that's already done elsewhere but couldn't find it quickly.

markylaing: Yep that's a little below in the same test: https://github.com/canonical/lxd/blob/5bfffbcab530b9399c90bddae90d1659494d9a25/test/suites/image.sh#L367-L371

tomponline:
@markylaing 

tests failing:

> Error: Storage pool driver "ceph" does not support buckets

markylaing:
> @markylaing can we use access handlers to load the requested project into the context to avoid needing the project ID checks?

In most cases there may not be a "requested project", I think all of these API endpoints accept an `all-projects` query parameter. 

More generally, I would like to check project existence much earlier, e.g. in `(Daemon).createCmd` and set the project name and ID in the request context. We can do this now, but it would lead to a lot of duplicated DB calls unless I do a lot of refactoring

markylaing:
@tomponline ready for review when you are. I would like to add a general check for the current project and load it into context at some point, but I think that will coincide some bigger changes to data access patterns in LXD, which will be required if changing access control to use entity IDs rather than URLs.

tomponline:
> @tomponline ready for review when you are. I would like to add a general check for the current project and load it into context at some point, but I think that will coincide some bigger changes to data access patterns in LXD, which will be required if changing access control to use entity IDs rather than URLs.

Thanks! The main thing is that we have behavioural tests added in this PR to catch regressions.

markylaing:
Will rebase when https://github.com/canonical/lxd/pull/16857 is merged

tomponline (approved):
Thanks!

PR #16818, opened 2025-10-24
Cluster group used_by follow-up

kadinsayani:
Follow-up to https://github.com/canonical/lxd/pull/16779.

### DIFF
--- a/lxd/instance_post.go
+++ b/lxd/instance_post.go
@@ -301,13 +301,11 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
 	}
 
 	// Unset "volatile.cluster.group" if the instance is manually moved to a cluster member.
-	if targetMemberInfo != nil && targetGroupName == "" {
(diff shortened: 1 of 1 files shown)


lxd/api_cluster.go
```diff
@@ -4594,7 +4594,7 @@ func clusterGroupUsedBy(ctx context.Context, s *state.State, tx *db.ClusterTx, n
 	}
 
 	if len(usedBy) > 0 && firstOnly {
-		return usedBy, nil
+		return usedBy[:1], nil
```
simondeziel, line 4597: TIL this nice and simple way of getting a slice with just the first element.

simondeziel (approved):

lxd/api_cluster.go
kadinsayani: It's rarely used, but indeed, quite nice :)

tomponline (approved):
ta

PR #16819, opened 2025-10-24
Download and cache `virtiofsd` on 22.04 runners

simondeziel:


simondeziel:
Moving back to draft after discussing with Din with the reverting in `test/suites/devlxd.sh` that should be done more cleanly.

test/suites/devlxd.sh
```diff
@@ -391,7 +391,9 @@ EOF
 
     # Manage device on a different instance.
     inst2="${inst}-2"
-    lxc launch testimage "${inst2}" --project "${project}" --storage "${pool}"
+
+    # shellcheck disable=SC2086 # Variable "opts" must not be quoted, we want word splitting.
+    lxc init --empty "${inst2}" --project "${project}" --storage "${pool}" ${opts}
```
MusicDin: > The original call was missing `${opts}` so it always created a container.

This can be either a container or VM (used container because it is smaller). The idea is just to make sure that volumes can be attached/detached on another instance (in terms of LXD CSI, controller handles volume attachment on all nodes).  

> I'm assuming that removing the volume from a running instance is not supported
so instead of `lxc launch` let's use a `lxc init` and use `--empty` while at
it.

Removing a volume from a running instance must be supported in order for LXD CSI to be usable.

simondeziel: OK then it seems this is not working yet per the error in the commit message.

I'll go back to using a running container, thanks for the clarification.

MusicDin: If you go with container approach, please just wrap the image cleanup to prevent removing container image before VM tests:

Before:
```bash
lxc image delete "$(lxc config get "${inst}" volatile.base_image --project "${project}")" --project "${project}"
```

After:
```bash
if [ "${instType}" = "virtual-machine" ]; then
  lxc image delete "$(lxc config get "${inst}" volatile.base_image --project "${project}")" --project "${project}"
fi
```

simondeziel: Worked well, thanks.

Issue #12637, opened 2023-12-07
Synchronous storage volume API endpoints may result in blocking server operations


mas-who:
# Required information
 * Distribution: Ubuntu snap
 * Distribution version: 5.19

# Issue description

The api endpoints for updating a volume `PUT /1.0/storage-pools/{poolName}/volumes/{type}/{volumeName}` and updating a volume snapshot `PUT /1.0/storage-pools/{poolName}/volumes/{type}/{volumeName}/snapshots/{snapshotName}` are synchronous at the moment. Making a request to either endpoint results in the below successful response:
```
{
  type: "sync",
  status: "Success",
  status_code: 200,
  operation: "",
  error_code: 0,
  error: "",
  metadata: {},
}
```
This seems to be inconsistent with other API endpoints which typically results in asynchronous operations. This is concerning since it may result in long and blocking operations on the server. Suggestion is to make these endpoints asynchronous.

tomponline:
@MusicDin this has been reported from our LXD UI team so would be good to take a look and evaluate.

tomponline:
@MusicDin please can you look at this one as your next bug, as if we need to change the API behaviour to make it consistent then we need to do so before LXD 5.21 LTS is released.

tomponline:
@MusicDin if we made these endpoints async would this affect your devlxd APIs? @mionaalex this might be something we want to do before 6 LTS as its an api break.

MusicDin:
> If we made these endpoints async would this affect your devlxd APIs?

Yes, in the same way the main LXD API is affected - the operation needs to be returned and potentially waited for on the client side. 

Should I (re)open a PR for this?

tomponline:
@MusicDin did you previously envision supporting both sync and async modes? 

Also would the updated SDK just return a nil-operation when talking with an older LXD without the API extension?

tomponline:
@MusicDin presumably it would make sense to decide upon this one now before we land the devlxd routes?

MusicDin:
Hmm. I'm all for updating those endpoints, I'm just a bit concerned that it would even further postpone the CSI. How about I do the first PR addressing the routes that are included in initial devLXD volume management? And afterwards the remaining routes?

MusicDin:
> did you previously envision supporting both sync and async modes?
> Also would the updated SDK just return a nil-operation when talking with an older LXD without the API extension?

I see the situation like this:

|            |  Old client | New Client
| ---------- | ----------- | ---
[...]

PR #16842, opened 2025-10-27
devlxd: Allow volume cloning

MusicDin:
Allow setting volume source when creating a new volume through DevLXD.
This change enables volume cloning in LXD CSI.

The PR includes:
- Changes to DevLXD's `POST /storage-pools/<pool>/volumes/<type>` (volume creation) endpoint to allow providing volume source
- Tests for volume cloning

Requires rebase after:
- https://github.com/canonical/lxd/pull/16882 (+ Remove `CreateStoragePoolFromSource` which will no longer be required once  `POST /storage-pools/<pool>/volumes/<type>` returns an operation).

### DIFF
--- a/test/suites/devlxd.sh
+++ b/test/suites/devlxd.sh
@@ -410,6 +410,44 @@ EOF
 
     lxc delete "${inst2}" --project "${project}" --force
 
+    # Use existing volume owned by the identity as a source for the new volume.
+    vol3=$(cat <<EOF
+{
+    "name": "vol-03",
+    "type": "custom",
+    "pool": "${pool}",
+    "source": {
+        "name": "vol-01",
+        "pool": "${pool}",
+        "type": "copy"
+    }
+}
+EOF
+)
+
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage create-volume "${pool}" "${vol3}"
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage get-volume "${pool}" custom vol-03
+
+    # Ensure creating a volume from a source volume not owned by the identity fails.
+    lxc storage volume create "${pool}" unowned size=8MiB
+
+    volUnownedSource=$(cat <<EOF
+{
+    "name": "vol-unowned",
+    "type": "custom",
+    "pool": "${pool}",
+    "source": {
+        "name": "unowned",
+        "pool": "${pool}",
+        "type": "copy"
+    }
+}
+EOF
+)
+
+    [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage create-volume "${pool}" "${volUnownedSource}")" = "Source volume not found" ]
+    lxc storage volume delete "${pool}" unowned
+
     # Delete storage volume (fail - insufficient permissions).
     [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage delete-volume "${pool}" "${instType}" "${inst}")" = "Forbidden" ]
 
@@ -423,6 +461,7 @@ EOF
     [ "$(lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage delete-volume "${pool}" custom non-existing-volume)" = "Storage volume not found" ]
     lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage delete-volume "${pool}" custom vol-01
     lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage delete-volume "${pool}" custom vol-02
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage delete-volume "${pool}" custom vol-03
 
     # Ensure storage volumes are deleted.
     lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage volumes "${pool}" | jq --exit-status '. == []'


lxd/devlxd_operation.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/canonical/lxd/client"
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/lxd/shared/api"
+)
+
+var devLXDOperationWaitEndpoint = APIEndpoint{
+	Path: "operations/{id}/wait",
+	Get:  APIEndpointAction{Handler: devLXDOperationsWaitHandler, AccessHandler: allowDevLXDAuthenticated},
+}
```
MusicDin, line 17: @markylaing Could you advise whether `AccessHandler: allowDevLXDAuthenticated` is sufficient for operations endpoint, or should I additionally perform some checks to verify DevLXD identity can access the operation?

We only need `GET operations/{id}/wait` to wait for the operation to complete and `DELETE operations/{id}` to potentially cancel the operation.

client/devlxd_storage.go
```diff
@@ -62,6 +67,23 @@ func (r *ProtocolDevLXD) CreateStoragePoolVolume(poolName string, vol api.DevLXD
...
+		return nil, errors.New("Volume source type is not set")
+	}
+
+	op, _, err := r.queryOperation(http.MethodPost, url.String(), vol, "")
+	if err != nil {
+		return nil, err
+	}
+
+	return op, nil
+}
```
MusicDin: Please pay attention to this function during review.

`ProtocolLXD` defines `CopyStoragePoolVolume`, which operates between two LXD servers — source and destination (the source may be the same as the destination).

In DevLXD, I’ve implemented this as `CreateStoragePoolVolumeFromSource`. Unlike LXD’s copy operation, it takes only a single remote because cross-remote and cross-project copies aren’t allowed. Additionally, DevLXD must fetch the source volume to verify ownership by the caller’s identity.

If we agree on this approach, I suggest avoiding the name `CopyStoragePoolVolume` for the DevLXD client to prevent confusion with the original LXD method.

lxd/devlxd_storage.go
```diff
@@ -229,6 +229,74 @@ func devLXDStoragePoolVolumesPostHandler(d *Daemon, r *http.Request) response.Re
...
+
+		// Populate request context with source volume details.
+		err = addStoragePoolVolumeDetailsToRequestContext(d.State(), req)
+		if err != nil {
+			if api.StatusErrorCheck(err, http.StatusNotFound) {
+				return response.DevLXDErrorResponse(api.NewStatusError(http.StatusNotFound, "Source volume not found"))
+			}
+
+			return response.DevLXDErrorResponse(err)
+		}
```
MusicDin, line 272: Since DevLXD needs to fetch the source volume to verify ownership, we need to adjust mux vars because the source volume is not part of the original URL. Maybe there is a better approach?

lxd/devlxd_operation.go
markylaing: To be consistent with the main operations API, the only identities that should be able to view an operation are those that can view operations for the whole project, or the requestor of the operation.

In case, the access handling is being performed within the main API handlers themselves. Since you're calling the main API handlers, this should already be correct.

client/interfaces.go
```diff
@@ -537,6 +537,9 @@ type DevLXDServer interface {
 	UpdateStoragePoolVolume(poolName string, volType string, volName string, vol api.DevLXDStorageVolumePut, ETag string) error
 	DeleteStoragePoolVolume(poolName string, volType string, volName string) error
 
+	// DevLXD operations.
+	GetOperationWait(uuid string, timeout int) (*api.DevLXDOperation, string, error)
```
markylaing, line 548: Not sure what our current conventions are but if passing a UUID as an argument to a client method I'd prefer it to be a `uuid.UUID`. If there are already `uuid string` arguments then disregard this comment :)

client/devlxd_storage.go
markylaing: It looks like the bodies of `CreateStoragePoolVolume` and `CreateStoragePoolVolumeFromSource` are identical. Is there really a need to have two different functions?

markylaing: Nevermind - I see one returns an operation and one doesnt :facepalm:

client/interfaces.go
tomponline: This would force consumers of the SDK to depend on a specific UUID package (as there's not a built in one).
If our SDK package already requires this because of internal usage then perhaps that's be OK, but in general it'd be good to avoid tying consumers to external packages in our SDK API.

lxd/devlxd_storage.go
markylaing: Perhaps you can refactor `addStoragePoolVolumeDetailsToRequestContext` to accept arguments? The access handler can extract these and pass them in.

tomponline:
> * DevLXD operation endpoints (when volume is created from a source volume, an operation is returned in response)

@MusicDin as https://github.com/canonical/lxd/issues/12637 is in for this cycle, and https://github.com/canonical/lxd/issues/12637#issuecomment-3291113001 comment, is this laying the ground work for that item?

It sounds like we should ensure that https://github.com/canonical/lxd/issues/12637 is completed before LXD 6.6 goes out (after the sprint) so that we dont land the new devlxd endpoints and then immediately break them in the next release.

MusicDin:
> It sounds like we should ensure that https://github.com/canonical/lxd/issues/12637 is completed before LXD 6.6 goes out (after the sprint) so that we dont land the new devlxd endpoints and then immediately break them in the next release.

Agree. This PR does not set the ground work for the above issue. I would rather say that it works around it, introducing an additional client function for cases when an operation is returned. 

For example, if `POST /storage-pools/<pool>/volumes/<type>` would always return an operation, volume cloning can be achieved through an existing client function (`CreateStoragePoolVolume`) because only `Source` needs to be provided in the request.

@tomponline Should I process with the endpoint changes and move this PR to draft?

tomponline:
> @tomponline Should I process with the endpoint changes and move this PR to draft?

Yes I think so, it sounds like at the very least we should sync on the correct order of delivery of these items.

I certainly don't like the sound of working around something that is going away soon anyway.

shared/api/devlxd_storage.go
```diff
@@ -75,4 +75,26 @@ type DevLXDStorageVolumesPost struct {
...
+	// Example: foo
+	Name string `json:"name" yaml:"name"`
+
+	// Source type. Currently, only "copy" is supported.
+	// Example: copy
+	Type string `json:"type" yaml:"type"`
+
+	// Source storage pool.
+	// Example: local
+	Pool string `json:"pool" yaml:"pool"`
```
tomponline, line 95: Does CSI allow for cloning to/from different pools? Wondering if we need this?

lxd/devlxd_storage.go
tomponline: @MusicDin did this get addressed? Can it be marked as resolved?

lxd/devlxd_storage.go
tomponline, line 1: Do we check if devlxd identity has access to source volume (what checks are used for the real REST API for that matter)?

cc @markylaing

shared/api/devlxd_storage.go
MusicDin: Yes, it does.

From the docs (https://kubernetes.io/docs/concepts/storage/volume-pvc-datasource/#introduction):
> Cloning is supported with a different Storage Class.
> - Destination volume can be the same or a different storage class as the source.
> - Default storage class can be used and storageClassName omitted in the spec.

Effectively, a user can have two storage classes, each pointing to one storage pool, and clone between them.

lxd/devlxd_storage.go
MusicDin: Yes. I fetch the source volume and ensure it is owned by the identity requesting volume cloning. If the identity does not own the source volume, a not found error is returned.

tomponline: @MusicDin can you point me to that please?

lxd/devlxd_storage.go
```diff
@@ -229,6 +229,70 @@ func devLXDStoragePoolVolumesPostHandler(d *Daemon, r *http.Request) response.Re
...
+		resp := storagePoolVolumeGet(d, req)
+		_, err = response.NewResponseCapture(req).RenderToStruct(resp, &sourceVol)
+		if err != nil {
+			return response.DevLXDErrorResponse(err)
+		}
+
+		// Ensure the source volume is owned by the caller.
+		if !isDevLXDVolumeOwner(sourceVol.Config, identity.Identifier) {
+			return response.DevLXDErrorResponse(api.NewStatusError(http.StatusNotFound, "Source volume not found"))
+		}
```
MusicDin, line 283: ^ @tomponline Just pointing to the check that ensures the volume is owned by the identity requesting volume cloning.

lxd/devlxd_storage.go
MusicDin: See comment https://github.com/canonical/lxd/pull/16842#discussion_r2534325341

lxd/devlxd_storage.go
tomponline: should we be checking fine-grained auth here?

What does the external REST API do here?

MusicDin: In addition:
- the DevLXD security flags are checked
- the access handler verifies the identity is allowed to create storage volumes. 
- project is restricted to the instance project (also verified by the access handler, if `project` query parameter is provided, it must match the project where the instance is running)

tomponline: cool, im asking about the access checks we perform on the source volume specified in the POST payload aligns with what we do for the external REST API - and wanted to check with @markylaing what we should be doing on the external endpoint too.

MusicDin: The DevLXD endpoint uses the the same check as LXD API:
```go
var devLXDStoragePoolVolumesTypeEndpoint = APIEndpoint{
	Path: "storage-pools/{poolName}/volumes/{type}",
	Post: APIEndpointAction{Handler: devLXDStoragePoolVolumesPostHandler, AccessHandler: allowDevLXDPermission(entity.TypeProject, auth.EntitlementCanCreateStorageVolumes)},
	...
}
```

Any internal checks are performed by the LXD API handler anyway, as the handler is called directly.

tomponline: Isnt the URL being checked the entity volume being created rather than the source though?

lxd/devlxd_storage.go
MusicDin: No, I'll try to refactor this in a follow-up PR to make `addStoragePoolVolumeDetailsToRequestContext` accept arguments, so that we do not need to manually adjust mux vars in the handler.

lxd/devlxd_storage.go
MusicDin: The URL does not contain the source volume. The source volume is part of the request payload/content.

MusicDin: I suggest we wait for Mark's input

tomponline: >  The URL does not contain the source volume. The source volume is part of the request payload/content.

Exactly :)

lxd/devlxd_storage.go
```diff
@@ -229,6 +229,70 @@ func devLXDStoragePoolVolumesPostHandler(d *Daemon, r *http.Request) response.Re
...
+
+		if vol.Source.Name == "" {
+			return response.DevLXDErrorResponse(api.StatusErrorf(http.StatusBadRequest, "Source volume name must be provided when source is configured"))
+		}
+
+		// Fetch a source volume.
+		sourceVol := api.StorageVolume{}
+
+		sourceURL := api.NewURL().Path("1.0", "storage-pools", vol.Source.Pool, "volumes", "custom", vol.Source.Name).Project(projectName)
```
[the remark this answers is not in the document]
MusicDin: We can validate, but the `poolName` is set in mux vars and it is validated in `storagePoolVolumesPost` (LXD API handler)

lxd/devlxd_storage.go
markylaing: Following along I gather the contention here is whether handler should check `can_view` on the source volume before copying. 

Din is right that it isn't doing that in the main API handler but that doesn't mean it's correct ;)

I think we should add this check in both places.

@MusicDin could you please add a new access handler for storage volume creation which checks  `can_create_storage_volumes` on the target project and then, if a source is provided, checks `can_view` on the source volume? Then please use the same access handler in both places. Thanks.

markylaing: This should not break anyone. Prior to this change if anyone was including a source volume that they are not permitted to view then we should wonder how they figured out the name of the volume in the first place.

tomponline: > Din is right that it isn't doing that in the main API handler but that doesn't mean it's correct ;)

OK will merge this for now and we can discuss more separately in general.

PR #16846, opened 2025-10-28
docs: CSI volume cloning

MusicDin:
Marks volume cloning as supported and adds an example of how to clone a volume.

Requires:
- https://github.com/canonical/lxd-csi-driver/pull/72

### DIFF
--- a/doc/howto/storage_csi.md
+++ b/doc/howto/storage_csi.md
@@ -284,6 +284,32 @@ Access mode        | Supported drivers                                          
 `ReadOnlyMany`     | {ref}`storage-drivers-shared`                                                                  | Mounted as read-only by many Pods across nodes.
 `ReadWriteMany`    | {ref}`storage-drivers-shared`                                                                  | Mounted as read-write by many Pods across nodes.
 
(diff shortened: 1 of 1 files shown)


doc/howto/storage_csi.md
```diff
@@ -295,6 +295,31 @@ Access mode        | Supported drivers
...
+kind: PersistentVolumeClaim
+metadata:
+  name: app-data
+spec:
+  accessModes:
+    - ReadWriteOnce
+  storageClassName: lxd-csi-sc
+  resources:
+    requests:
+      storage: 10Gi             # Must be equal to or larger than the size of the source volume.
```
simondeziel, line 306:
```suggestion
      storage: 10GiB            # Must be equal to or larger than the size of the source volume.
```

MusicDin: In Kubernetes, bytes (`B`) are not specified: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory

simondeziel: Oh TIL, thanks.

PR #16848, opened 2025-10-28
Apparmor policy fix and tests (from Incus)

kadinsayani:
Fixes https://github.com/canonical/lxd/pull/16732#issue-3513677322 by correctly introducing the apparmor policy changes required for namespaced `binfmt_misc` support in unprivileged containers. 

This pull request also adds apparmor restriction tests to prevent future regressions.

Related lxd-ci tests: https://github.com/canonical/lxd-ci/pull/591.

### DIFF
--- a/test/main.sh
+++ b/test/main.sh
@@ -636,6 +636,7 @@ if [ "${1:-"all"}" != "cluster" ] && [ "${1:-"all"}" != "standalone" ]; then
     run_test test_snap_vm_empty "snap empty VM"
     run_test test_snap_lxd_user "snap lxd-user"
     run_test test_snap_storage_volume_attach_vm "snap attaching storage volumes to VMs"
+    run_test test_snap_apparmor "snap apparmor restrictions"
 fi
 
 # shellcheck disable=SC2034


test/suites/apparmor.sh
```diff
@@ -0,0 +1,115 @@
+test_apparmor() {
+  if [ ! -e /sys/module/apparmor/ ]; then
+    echo "==> SKIP: AppArmor /sys/kernel/* tests (missing kernel support)"
```
simondeziel: This one calls for `TEST_UNMET_REQUIREMENT`.

test/includes/lxd.sh
```diff
@@ -389,6 +389,35 @@ waitInstanceReady() {
...
+  local instProj="${2:-}"
+  local i
+
+  # Wait for the instance to be ready
+  waitInstanceReady "${instName}" "${instProj}"
+
+  # Then wait for the boot sequence to complete.
+  sleep 1
+  rc=0
+  state="$(lxc exec --project "${instProj}" "${instName}" -- timeout "${maxWait}" systemctl is-system-running --wait)" || rc="$?"
```
simondeziel: This being `systemd` specific will not work with busybox containers.

test/suites/apparmor.sh
```diff
@@ -0,0 +1,115 @@
...
+  echo "Verify read access to other /proc/sys/kernel paths allowed (only write denied)"
+  if lxc exec "${ctName}" -- test -f /proc/sys/kernel/kptr_restrict; then
+    lxc exec "${ctName}" -- cat /proc/sys/kernel/kptr_restrict
+  else
+    echo "    kptr_restrict not available (skipping read test)"
+  fi
+
+  echo "Verify write access to denied paths blocked by AppArmor"
+  if lxc exec "${ctName}" -- test -f /proc/sys/kernel/kptr_restrict; then
+    ! echo 0 | lxc file push - "${ctName}"/proc/sys/kernel/kptr_restrict || false
```
simondeziel: I would use `echo 0 | lxc exec ... tee` as file pushing is handled differently.

test/suites/apparmor.sh
```diff
@@ -0,0 +1,115 @@
+test_apparmor() {
+  if [ ! -e /sys/module/apparmor/ ]; then
+    echo "==> SKIP: AppArmor /sys/kernel/* tests (missing kernel support)"
+    return
+  fi
+
+  ensure_import_testimage
+  ensure_has_localhost_remote "${LXD_ADDR}"
```
simondeziel: Is this truly needed?

test/suites/apparmor.sh
```diff
@@ -0,0 +1,115 @@
...
+  if lxc exec "${ctName}" -- test -f /proc/sys/kernel/kptr_restrict; then
+    ! echo 0 | lxc file push - "${ctName}"/proc/sys/kernel/kptr_restrict || false
+  else
+    echo "    kptr_restrict not available (skipping test)"
+  fi
+
+  echo "==> Test /proc read access"
+
+  echo "Verify read access to /proc/stat"
+  lxc exec "${ctName}" -- cat /proc/stat | head -n1 | grep -q "^cpu"
```
simondeziel:
```suggestion
  lxc exec "${ctName}" -- grep -wm1 "^cpu" /proc/stat
```

test/suites/apparmor.sh
```diff
@@ -0,0 +1,115 @@
...
+
+  echo "Verify read access to /proc/meminfo"
+  lxc exec "${ctName}" -- cat /proc/meminfo | head -n1 | grep -q "MemTotal"
```
simondeziel:
```suggestion
  lxc exec "${ctName}" -- grep -wm1 "^MemTotal:" /proc/meminfo
```

test/suites/apparmor.sh
```diff
@@ -0,0 +1,115 @@
...
+  echo "Verify read access to /proc/uptime"
+  lxc exec "${ctName}" -- cat /proc/uptime
+
+  echo "==> Test /proc denied paths"
+
+  echo "Verify access to /proc/kcore denied (rwklx)"
+  ! lxc exec "${ctName}" -- cat /proc/kcore || false
+
+  echo "Verify access to /proc/sysrq-trigger denied (rwklx)"
+  ! lxc exec "${ctName}" -- cat /proc/sysrq-trigger || false
```
simondeziel: Trying to read that file also fails in my host due to it being only writable:
```
$ ll /proc/sysrq-trigger 
--w------- 1 root root 0 Oct 28 12:59 /proc/sysrq-trigger
```

I'd probably try to `echo ... | lxc exec ... tee /proc/sysrq-trigger` into it.

test/suites/apparmor.sh
```diff
@@ -0,0 +1,115 @@
...
+
+  echo "==> Test /proc/sys read access"
+
+  echo "Verify read access to /proc/sys/fs/* allowed (only write denied)"
+  if lxc exec "${ctName}" -- test -f /proc/sys/fs/file-max; then
+    lxc exec "${ctName}" -- cat /proc/sys/fs/file-max
```
simondeziel: Worth making sure writes fail?

test/suites/apparmor.sh
```diff
@@ -0,0 +1,115 @@
...
+  else
+    echo "    /proc/sys/fs/file-max not available (skipping test)"
+  fi
+
+  echo "Verify read access to /proc/sys/net/* allowed (only write denied)"
+  if lxc exec "${ctName}" -- test -f /proc/sys/net/ipv4/ip_forward; then
+    lxc exec "${ctName}" -- cat /proc/sys/net/ipv4/ip_forward
```
simondeziel: I'd try `echo 0 | lxc exec ... tee ...`

test/suites/basic.sh
```diff
@@ -777,6 +777,18 @@ test_snap_basic_usage_vm() {
   lxc delete -f v1
 }
 
+test_snap_basic_usage_container() {
+  ensure_import_testimage
+  ensure_has_localhost_remote "${LXD_ADDR}"
```
simondeziel: Likely unneeded.

test/suites/basic.sh
```diff
@@ -777,6 +777,18 @@ test_snap_basic_usage_vm() {
...
+
+  echo "==> Launch container"
+  lxc launch testimage c1 -c limits.memory=384MiB -d root,size.state=384MiB -d "${SMALL_VM_ROOT_DISK}"
+  waitInstanceBooted c1
```
simondeziel: Won't work because this is a busybox image lacking systemd.

test/suites/basic.sh
```diff
@@ -777,6 +777,18 @@ test_snap_basic_usage_vm() {
   lxc delete -f v1
 }
 
+test_snap_basic_usage_container() {
+  ensure_import_testimage
+  ensure_has_localhost_remote "${LXD_ADDR}"
+
+  echo "==> Launch container"
+  lxc launch testimage c1 -c limits.memory=384MiB -d root,size.state=384MiB -d "${SMALL_VM_ROOT_DISK}"
```
simondeziel: Looks like a copy-n-paste from a VM creation. The `size.state` bit should be dropped and the root disk size should lose the `VM_` bit.

test/main.sh
```diff
@@ -635,6 +636,7 @@ if [ "${1:-"all"}" != "cluster" ] && [ "${1:-"all"}" != "standalone" ]; then
     run_test test_snap_vm_empty "snap empty VM"
     run_test test_snap_lxd_user "snap lxd-user"
     run_test test_snap_storage_volume_attach_vm "snap attaching storage volumes to VMs"
+    run_test test_snap_basic_usage_container "snap basic usage container"
```
simondeziel: Things that are not snap specific should be kept as regular (non-snap) tests. I recently dropped such basic usage snap test because of that.

test/suites/apparmor.sh
kadinsayani: Not needed, copy and paste bit.

test/main.sh
kadinsayani: Sounds good, I'll exclude that additional test.

test/suites/apparmor.sh
kadinsayani: Ah yes, thanks.

test/suites/apparmor.sh
kadinsayani: Thanks for the suggestion, that works well :)

test/suites/apparmor.sh
kadinsayani: Yeah good idea, thanks!

test/suites/basic.sh
kadinsayani: Ack, thanks.

test/suites/apparmor.sh
simondeziel: On second though, this is the kind of thing that is NS aware so it's a `sysctl` that can be tweaked inside containers.

kadinsayani: Yeah this bit was incorrect, both read and write are allowed.

test/suites/apparmor.sh
```diff
@@ -0,0 +1,117 @@
...
+  echo "Verify write access to /sys/kernel/security denied"
+  ! lxc exec c1 -- touch /sys/kernel/security/testfile || false
+
+  echo "Verify write access to /sys/kernel/security/apparmor allowed"
+  lxc exec c1 -- touch /sys/kernel/security/apparmor/.load
+
+  echo "==> Test /sys/firmware/efi/efivars/ access"
+
+  echo "Verify read access to /sys/firmware/efi/efivars/ denied"
+  ! lxc exec c1 -- ls /sys/firmware/efi/efivars/ || false
```
simondeziel: This is the one dir I'm not 100% sure will be present even in the host. I don't know if Azure VMs are CSM or EFI based TBH.

This might need to be gated by the existence of that dir in the host.

test/suites/apparmor.sh
```diff
@@ -0,0 +1,117 @@
...
+  echo "Verify access to /proc/sysrq-trigger denied"
+  ! lxc exec c1 -- cat /proc/sysrq-trigger || false
+
+  echo "Verify access to /proc/acpi/** denied"
+  ! lxc exec c1 -- cat /proc/acpi/wakeup || false
+
+  echo "==> Test /proc/sys access"
+
+  echo "Verify read access to /proc/sys/fs/* allowed and write access denied"
+  if lxc exec c1 -- test -f /proc/sys/fs/file-max; then
```
simondeziel: I think it's fair to assume this file will always be available.

test/suites/apparmor.sh
```diff
@@ -0,0 +1,117 @@
...
+    lxc exec c1 -- cat /proc/sys/fs/file-max
+    ! echo 0 | lxc exec c1 -- tee /proc/sys/fs/file-max || false
+  else
+    echo "    /proc/sys/fs/file-max not available (skipping test)"
+  fi
+
+  echo "==> Test /proc/sys/net/* access"
+
+  echo "Verify read and write access to /proc/sys/net/ipv4/ip_forward is allowed"
+  if lxc exec c1 -- test -f /proc/sys/net/ipv4/ip_forward; then
```
simondeziel: Same deal with ip_forward being always available.

test/suites/apparmor.sh
tomponline: @mihalicyn ?

simondeziel: Locally with my EFI based host, I get a permission denied in the container. On Azure, we get:

```
2025-10-28T19:02:58.7592696Z + lxc exec c1 -- ls /sys/firmware/efi/efivars/
2025-10-28T19:02:58.7593488Z + timeout --foreground 120 /home/runner/go/bin/lxc exec c1 --verbose --force-local -- ls /sys/firmware/efi/efivars/
2025-10-28T19:02:58.7911153Z ls: /sys/firmware/efi/efivars/: No such file or directory
```

test/suites/apparmor.sh
kadinsayani: Sounds good, agreed.

test/suites/apparmor.sh
kadinsayani: I've added a check for existence of the directory.

test/suites/apparmor.sh
```diff
@@ -0,0 +1,113 @@
...
+
+  echo "Verify write access to /sys/kernel/security denied"
+  ! lxc exec c1 -- touch /sys/kernel/security/testfile || false
+
+  echo "Verify write access to /sys/kernel/security/apparmor allowed"
+  lxc exec c1 -- touch /sys/kernel/security/apparmor/.load
+
+  echo "==> Test /sys/firmware/efi/efivars/ access"
+
+  if lxc exec c1 -- test -d /sys/firmware/efi/efivars/; then
```
simondeziel: Aleks would know for sure but I assumed this depended on the `efivars` existing in the host itself.

kadinsayani: That makes sense to me, I've fixed the check.

test/suites/apparmor.sh
```diff
@@ -0,0 +1,113 @@
...
+  echo "Verify read access to /proc/uptime"
+  lxc exec c1 -- cat /proc/uptime
+
+  echo "==> Test /proc denied paths"
+
+  echo "Verify access to /proc/kcore denied"
+  ! lxc exec c1 -- cat /proc/kcore || false
+
+  echo "Verify access to /proc/sysrq-trigger denied"
+  ! lxc exec c1 -- cat /proc/sysrq-trigger || false
```
simondeziel, line 104: This file is write-only:
```
$ ll /proc/sysrq-trigger 
--w------- 1 root root 0 Oct 29 09:53 /proc/sysrq-trigger
```

I'd try to `echo b | ... tee` into it.

test/suites/apparmor.sh
```diff
@@ -0,0 +1,113 @@
...
+  if [ ! -e /sys/module/apparmor/ ]; then
+    export TEST_UNMET_REQUIREMENT="missing AppArmor kernel support"
+    return
+  fi
+
+  ensure_import_testimage
+
+  echo "Create and start a test container"
+  lxc launch testimage c1
+
```
simondeziel, line 12: Here, I'd probably clear the `dmesg` buffer (`dmesg -c`).

test/suites/apparmor.sh
```diff
@@ -0,0 +1,113 @@
...
+  if lxc info | grep -F 'unpriv_binfmt: "true"'; then
+    echo "    binfmt_misc is supported (unpriv_binfmt enabled)"
+
+    echo "Verify binfmt_misc can be mounted and accessed"
+    lxc exec c1 -- mount -t binfmt_misc none /proc/sys/fs/binfmt_misc
+    [ "$(lxc exec c1 -- cat /proc/sys/fs/binfmt_misc/status)" = "enabled" ]
+  else
+    echo "    binfmt_misc not supported (unpriv_binfmt disabled)"
+  fi
+
```
simondeziel, line 141: Here I'd probably look at the Apparmor denial accumulated in the `dmesg` buffer. Showing them would be good and maybe inspecting them to ascertain only the expected ones are present?

`dmesg --notime` might be useful for that but it will require some `sed` trickery. This can probably wait for another PR though.

kadinsayani: Doesn't look like we're getting any useful audit logs from apparmor. I'll investigate further.

test/suites/apparmor.sh
```diff
@@ -0,0 +1,147 @@
+test_apparmor() {
```
simondeziel: Considering that we rely on the `apparmor_parser` to generate the instance profile, I think this one test would make sense to have as a regular one and a `snap` variant. As such, could you extract the testing logic to a helper function like `_instance_apparmor()` that's called by both `test_apparmor()` and `test_snap_apparmor()`.

Sorry for suggesting it this late, hopefully the bulk of the work is reusable.

kadinsayani: Agreed, it would be helpful to include a snap variant for these tests. Thanks for suggesting!

tomponline:
@mihalicyn please can you review

PR #16851, opened 2025-10-29
devlxd: Allow volume snapshot management

MusicDin:
This PR adds DevLXD endpoints for volume snapshot management (list, get, create, delete).
Volume snapshots can be managed only for volumes that are owned by the caller's identity.

Requires rebase after:
- https://github.com/canonical/lxd/pull/16819
- https://github.com/canonical/lxd/pull/16842

### DIFF
--- a/test/suites/devlxd.sh
+++ b/test/suites/devlxd.sh
@@ -519,6 +519,108 @@ EOF
   fi
 }
 
+test_devlxd_volume_management_snapshots() {
+  local testName="devlxd-volume-mgmt"
+
+  local instPrefix="${testName}"
+  local pool="${testName}"
+  local project="${testName}"
+  local authGroup="${testName}-group"
+  local authIdentity="devlxd/${testName}-identity"
+
+  poolDriver="$(storage_backend "$LXD_DIR")"
+  lxc storage create "${pool}" "${poolDriver}"
+
+  if [ "${project}" != "default" ]; then
+    lxc project create "${project}" --config features.images=false
+  fi
+
+  local instTypes="container"
+  ensure_import_testimage
+
+  if [ "${LXD_VM_TESTS:-0}" != "0" ]; then
+    instTypes="${instTypes} virtual-machine"
+    ensure_import_ubuntu_vm_image
+  fi
+
+  for instType in $instTypes; do
+    inst="${instPrefix}-${instType}"
+
+    image="testimage"
+    opts="--storage ${pool} --config security.devlxd.management.volumes=true"
+    if [ "${instType}" = "virtual-machine" ]; then
+        image="ubuntu-vm"
+        opts="$opts --vm --config limits.memory=384MiB --device ${SMALL_VM_ROOT_DISK}"
+    fi
+
+    # shellcheck disable=SC2086 # Variable "opts" must not be quoted, we want word splitting.
+    lxc launch "${image}" "${inst}" $opts --project "${project}"
+    waitInstanceReady "${inst}" "${project}"
+
+    # Install devlxd-client and make sure it works.
+    lxc file push --project "${project}" --quiet "$(command -v devlxd-client)" "${inst}/bin/"
+    lxc exec --project "${project}" "${inst}" -- devlxd-client
+
+    # Create identity with all volume management permissions.
+    lxc auth group create "${authGroup}"
+    lxc auth identity create "${authIdentity}" --group "${authGroup}"
+    token=$(lxc auth identity token issue "${authIdentity}" --quiet)
+
+    # Grant permissions.
+    lxc auth group permission add "${authGroup}" project "${project}" can_view
+    lxc auth group permission add "${authGroup}" project "${project}" storage_volume_manager
+
+    # Create a custom storage volume.
+    vol1='{
+    "name": "vol-01",
+    "type": "custom",
+    "config": {
+        "size": "8MiB"
+    }
+}'
+
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage create-volume "${pool}" "${vol1}"
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage get-volume "${pool}" custom vol-01
+
+    # Create snapshots with custom and auto-generated names.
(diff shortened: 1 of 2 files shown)


test/suites/devlxd.sh
```diff
@@ -527,6 +566,122 @@ EOF
...
+    lxc auth group permission add "${authGroup}" project "${project}" can_view
+    lxc auth group permission add "${authGroup}" project "${project}" storage_volume_manager
+
+    # Create a custom storage volume.
+    vol1=$(cat <<EOF
+{
+    "name": "vol-01",
+    "type": "custom",
+    "config": {
+        "size": "10MiB"
```
simondeziel: Could you go with 8MiB instead of 10MiB, this better aligns with LVM and the PE that forces rounding to multiples of 4MiB.

test/suites/devlxd.sh
```diff
@@ -527,6 +566,122 @@ EOF
...
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage create-volume "${pool}" "${vol1}"
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage get-volume "${pool}" custom vol-01
+
+    # Create snapshots.
+    vol1=$(cat <<EOF
+{
+    "name": "vol-01",
+    "type": "custom",
+    "config": {
+        "size": "10MiB"
```
simondeziel: 8MiB here too, please

test/suites/devlxd.sh
```diff
@@ -460,6 +460,44 @@ EOF
...
+    }
+}
+EOF
+)
+
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage create-volume "${pool}" "${vol3}"
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage get-volume "${pool}" custom vol-03
+
+    # Ensure creating a volume from a source volume not owned by the identity fails.
+    lxc storage volume create "${pool}" unowned size=10MiB
```
simondeziel: 8MiB here too, please

test/suites/devlxd.sh
```diff
@@ -519,6 +519,122 @@ EOF
...
+  local testName="devlxd-volume-mgmt"
+
+  local instPrefix="${testName}"
+  local pool="${testName}"
+  local project="${testName}"
+  local authGroup="${testName}-group"
+  local authIdentity="devlxd/${testName}-identity"
+
+  poolDriver="$(storage_backend "$LXD_DIR")"
+  lxc storage create "${pool}" "${poolDriver}"
```
simondeziel, line 532: Would it work if the existing pool (automatically created) was used?

```
pool=$(lxc profile device get default root pool)
```

I'm asking because creating takes ~5s on Ceph and it's similarly slow to delete.

test/suites/devlxd.sh
```diff
@@ -519,6 +519,122 @@ EOF
...
+{
+    "name": "vol-01",
+    "type": "custom",
+    "config": {
+        "size": "8MiB"
+    }
+}
+EOF
+)
+
```
simondeziel: This looks like an unused redefinition of `vol1`.

test/suites/devlxd.sh
```diff
@@ -519,6 +519,122 @@ EOF
...
+    vol1=$(cat <<EOF
+{
+    "name": "vol-01",
+    "type": "custom",
+    "config": {
+        "size": "8MiB"
+    }
+}
+EOF
+)
```
simondeziel: `cat` and the subshell shouldn't be needed:

```suggestion
    vol1='{
    "name": "vol-01",
    "type": "custom",
    "config": {
        "size": "8MiB"
    }
}'
```

test/suites/devlxd.sh
```diff
@@ -519,6 +519,122 @@ EOF
...
+
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage delete-snapshot "${pool}" custom vol-01 snap0
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage snapshots "${pool}" custom vol-01 | jq --exit-status 'length == 0'
+
+    # Delete storage volume.
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client storage delete-volume "${pool}" custom vol-01
+
+    # Cleanup.
+    lxc image delete "$(lxc config get "${inst}" volatile.base_image --project "${project}")" --project "${project}"
+    lxc delete "${inst}" --project "${project}" --force
```
simondeziel, line 611: I /think/ it's best to delete instances before deleting their `base_image` to avoid creating `zombie` datasets.

test/suites/devlxd.sh
```diff
@@ -519,6 +519,117 @@ EOF
...
+    opts="--storage ${pool} --config security.devlxd.management.volumes=true"
+    if [ "${instType}" = "virtual-machine" ]; then
+        image="ubuntu-vm"
+        opts="$opts --vm --config limits.memory=384MiB --device ${SMALL_VM_ROOT_DISK}"
+
+        orig_volume_size="$(lxc storage get "${pool}" volume.size)"
+        if [ -n "${orig_volume_size:-}" ]; then
+          echo "==> Override the volume.size to accommodate a large VM"
+          lxc storage set "${pool}" volume.size "${SMALLEST_VM_ROOT_DISK}"
+        fi
```
simondeziel: I forgot about needing this for VMs created on the default pool. Thanks!

MusicDin: @simondeziel I've reverted usage of default pool. When I set the `volume.size` and later revert to previous value the LVM standalone test (`test_storage()`) consistently fails with an out of space error. I've confirmed the value is reverted, and no resources are left.

I'll create a separate storage pool for the test and temporary sacrifice the 5s time saving in order to land this in 6.6 and not block any longer. 

I'll open a follow-up PR where we can optimize the test using the default storage pool.

simondeziel: Sure, works for me.

simondeziel (approved):
LGTM, thanks!

tomponline:
Would appreciate a quick check of the security checks in this when you get a chance @markylaing  (they look fine to me).

PR #16853, opened 2025-10-29
Reword cluster error messages for consistency and check operations before cluster member evacuate/restore

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/15877.

### DIFF
--- a/test/main.sh
+++ b/test/main.sh
@@ -453,6 +453,7 @@ if [ "${1:-"all"}" != "snap" ] && [ "${1:-"all"}" != "standalone" ]; then
     run_test test_clustering_failure_domains "clustering failure domains"
     run_test test_clustering_image_refresh "clustering image refresh"
     run_test test_clustering_evacuation "clustering evacuation"
+    run_test test_clustering_evacuation_restore_operations "clustering evacuation/restore operations"
     run_test test_clustering_move "clustering move"
     run_test test_clustering_remove_members "clustering config remove members"
     run_test test_clustering_autotarget "clustering autotarget member"


test/suites/clustering.sh
```diff
@@ -3951,6 +3951,100 @@ test_clustering_evacuation() {
...
+  setup_clustering_netns 2
+  LXD_TWO_DIR=$(mktemp -d -p "${TEST_DIR}" XXX)
+  ns2="${prefix}2"
+  spawn_lxd_and_join_cluster "${ns2}" "${bridge}" "${cert}" 2 1 "${LXD_TWO_DIR}" "${LXD_ONE_DIR}" "${poolDriver}"
+
+  LXD_DIR="${LXD_ONE_DIR}"
+  ensure_import_testimage
+
+  echo "Launch 5 containers on node1"
+  for i in {1..5}; do lxc launch testimage c"${i}" --target node1; done
```
simondeziel: Unless there is a special meaning for 5 instances, I'd go with 2 or 3. Low but still high enough to ensure the normal placement algorithm would like to place at least one instance per member (unless told otherwise).

Also, you can put the `c` prefix in the `{..}` expansion:

```suggestion
  for i in c{1..3}; do lxc launch testimage "${i}" --target node1; done
```

test/suites/clustering.sh
```diff
@@ -3951,6 +3951,100 @@ test_clustering_evacuation() {
...
+
+  echo "Start node1 evacuation in background"
+  lxc cluster evacuate node1 --quiet --force&
```
simondeziel: I find that separating the `&` is easier to read and probably to `grep` for.

test/suites/clustering.sh
```diff
@@ -3951,6 +3951,100 @@ test_clustering_evacuation() {
...
+
+  # Spawn first node
+  setup_clustering_netns 1
+  LXD_ONE_DIR=$(mktemp -d -p "${TEST_DIR}" XXX)
+  ns1="${prefix}1"
+  spawn_lxd_and_bootstrap_cluster "${ns1}" "${bridge}" "${LXD_ONE_DIR}" "${poolDriver}"
+
+  echo "Check the state of the preseeded storage pool shows up as CREATED"
+  LXD_DIR="${LXD_ONE_DIR}" lxc storage list | grep -wF data | grep -wF CREATED
+
```
simondeziel: I saw the same in a similar test but is it needed here as well?

test/suites/clustering.sh
```diff
@@ -3951,6 +3951,100 @@ test_clustering_evacuation() {
...
+      loc=$(lxc list -f csv -c L "${c}")
+      if [ "${loc}" = "node1" ]; then
+        all_evacuated=false
+        break
+      fi
+    done
+    $all_evacuated && break
+    sleep 1
+  done
+  sleep 5 # Wait a bit for the operation to complete
```
simondeziel: Could the `wait_for` helper be used here?

test/suites/clustering.sh
kadinsayani: Oops, copy and paste bit.

test/suites/clustering.sh
kadinsayani: 2 should work fine for this test.

test/suites/clustering.sh
kadinsayani: I think we would have to start the evacuation operation using `wait_for lxc query...`. This wouldn't properly check for the bug and test the associated fix.

test/suites/clustering.sh
kadinsayani: Much cleaner with the `c` prefix before the `{}` expansion, I'll keep that in mind for next time, thanks :)

kadinsayani: Actually, I think 3 would be safer since there is a timing component to this bug. We want to make sure `lxc cluster restore node1` fails if the evacuation operation is still in progress. Given that some storage drivers are faster than others, I think 3 is a safer choice.

simondeziel: Yeah, I also wondered about that 1s sleep.

kadinsayani: > Yeah, I also wondered about that 1s sleep.

We need to wait for the operation to start so it's detected when the next request hits the handler.

kadinsayani: I've reduced the other `sleep`'s to more sensible values.

test/suites/clustering.sh
```diff
@@ -3951,6 +3951,97 @@ test_clustering_evacuation() {
...
+
+  LXD_DIR="${LXD_ONE_DIR}"
+  ensure_import_testimage
+
+  echo "Launch 3 containers on node1"
+  for i in c{1..3}; do lxc launch testimage "${i}" --target node1; done
+
+  echo "Start node1 evacuation in background"
+  lxc cluster evacuate node1 --quiet --force &
+  sleep 1 # Wait a bit for the operation to start
```
simondeziel, line 3991: Would it make sense to save the `lxc cluster evacuate` PID to `EVAC_PID=$!` here and use it further down?

test/suites/clustering.sh
```diff
@@ -3951,6 +3951,97 @@ test_clustering_evacuation() {
...
+
+  echo "Check restore fails while evacuation operation in progress"
+  ! lxc cluster restore node1 --quiet --force || false
```
simondeziel: After that, I'd use `wait "${EVAC_PID}"` to achieve 2 things:

1. Validate that evac was still ongoing while the `lxc cluster restore` was tried and failed.
2. Simplify the pooling for the container evacuation few lines below.

test/suites/clustering.sh
```diff
@@ -3951,6 +3951,97 @@ test_clustering_evacuation() {
...
+  lxc cluster restore node1 --quiet --force &
+  sleep 1 # Wait a bit for the operation to start
+
+  echo "Check evacuation fails while restore operation in progress"
+  ! lxc cluster evacuate node1 --quiet --force || false
+
+  echo "Wait for all containers to be restored to node1"
+  for _ in {1..60}; do
+    all_restored=true
+    for c in c1 c2 c3; do
```
simondeziel: Would it make sense to reuse the `for c in c{1..3}; do` syntax here?

test/suites/clustering.sh
```diff
@@ -3951,6 +3951,97 @@ test_clustering_evacuation() {
...
+  setup_clustering_netns 2
+  LXD_TWO_DIR=$(mktemp -d -p "${TEST_DIR}" XXX)
+  ns2="${prefix}2"
+  spawn_lxd_and_join_cluster "${ns2}" "${bridge}" "${cert}" 2 1 "${LXD_TWO_DIR}" "${LXD_ONE_DIR}" "${poolDriver}"
+
+  LXD_DIR="${LXD_ONE_DIR}"
+  ensure_import_testimage
+
+  echo "Launch 3 containers on node1"
+  for i in c{1..3}; do lxc launch testimage "${i}" --target node1; done
```
simondeziel: I'd use this for consistency with the rest:
```suggestion
  for c in c{1..3}; do lxc launch testimage "${c}" --target node1; done
```

test/suites/clustering.sh
```diff
@@ -3951,6 +3951,97 @@ test_clustering_evacuation() {
...
+      fi
+    done
+    $all_evacuated && break
+    sleep 1
+  done
+  sleep 1 # Wait a bit for the operation to complete
+
+  echo "Verify all containers are no longer on node1"
+  for c in c1 c2 c3; do
+    [ "$(lxc list -f csv -c L "${c}")" != "node1" ]
```
simondeziel: Positive checks are "more precise" so in this case, I think we can safely assume it will always be `node2` due to being a 2-wide cluster.
```suggestion
    [ "$(lxc list -f csv -c L "${c}")" = "node2" ]
```

test/suites/clustering.sh
```diff
@@ -3951,6 +3951,97 @@ test_clustering_evacuation() {
...
+        break
+      fi
+    done
+    $all_restored && break
+    sleep 1
+  done
+  sleep 1 # Wait a bit for the operation to complete
+
+  echo "Verify all containers are back on node1"
+  for c in c1 c2 c3; do
```
simondeziel: Same deal with the `c{1..3}` syntax.

test/suites/clustering.sh
```diff
@@ -3951,6 +3951,97 @@ test_clustering_evacuation() {
...
+    [ "$(lxc list -f csv -c L "${c}")" != "node1" ]
+  done
+
+  echo "Start node1 restore in background"
+  lxc cluster restore node1 --quiet --force &
+  sleep 1 # Wait a bit for the operation to start
+
+  echo "Check evacuation fails while restore operation in progress"
+  ! lxc cluster evacuate node1 --quiet --force || false
```
simondeziel: After looking at the Go code, I think that checking the specific error message is returned would be ideal here. Safer than just expecting a failure.

simondeziel (commented):
I like the new test case even if the many nits might imply otherwise. It's just because you said you didn't mind `review --verbose` ;)

test/suites/clustering.sh
```diff
@@ -3951,6 +3951,97 @@ test_clustering_evacuation() {
...
+  done
+  sleep 1 # Wait a bit for the operation to complete
+
+  echo "Verify all containers are back on node1"
+  for c in c1 c2 c3; do
+    [ "$(lxc list -f csv -c L "${c}")" = "node1" ]
+  done
+
+  echo "Clean up"
+  lxc delete c1 c2 c3 --force
```
simondeziel:
```suggestion
  lxc delete c{1..3} --force
```

kadinsayani:
> I like the new test case even if the many nits might imply otherwise. It's just because you said you didn't mind `review --verbose` ;)

Always appreciate your `review --verbose` output :)

test/suites/clustering.sh
kadinsayani: Great idea, thanks!

test/suites/clustering.sh
kadinsayani: Agreed, thanks for the suggestion.

tomponline:
@kadinsayani please can you rebase

tomponline (approved):
Great work thanks!

tomponline:
Tests are failing though looks like left over bridge interface

test/suites/clustering.sh
```diff
@@ -3952,6 +3952,75 @@ test_clustering_evacuation() {
...
+  echo "Wait for all containers to be restored to node1"
+  wait "${restore_pid}"
+
+  echo "Verify all containers are no longer on node2 and have been restored to node1"
+  for c in c{1..3}; do
+    [ "$(lxc list -f csv -c L "${c}")" = "node1" ]
+  done
+
+  echo "Clean up"
+  lxc delete c{1..3} --force
```
simondeziel, line 4021: As found by Tom, seems to be missing just `lxc network delete "${bridge}"`

kadinsayani: Thanks, fixed.

PR #16857, opened 2025-10-30
lint: Fix errors reported by `perfsprint: concat-loop`

markylaing:
These are such micro-optimisations that I was tempted to just turn the rule off - but it's good to keep it on so that we write more efficient code in future.

The fixes aren't the most efficient that they could possibly be, as there are a lot of unnecessary fmt.Sprintf's remaining.

### DIFF
--- a/shared/filter/clause.go
+++ b/shared/filter/clause.go
@@ -70,12 +70,14 @@ func Parse(s string, op OperatorSet) (*ClauseSet, error) {
 			return nil, errors.New("clause has no value")
 		}
 
-		value := parts[index]
+		var value strings.Builder
+		value.WriteString(parts[index])
 
 		// support strings with spaces that are quoted
 		for _, symbol := range op.Quote {
-			if strings.HasPrefix(value, symbol) {
-				value = value[1:]
+			if strings.HasPrefix(parts[index], symbol) {
+				value.Reset()
+				value.WriteString(parts[index][1:])
 				for {
 					index++
 					if index == len(parts) {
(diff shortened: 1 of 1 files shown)


lxd/apparmor/instance.go
```diff
@@ -146,11 +146,11 @@ func instanceProfileGenerate(sysOS *sys.OS, inst instance) error {
 // instanceProfile generates the AppArmor profile template from the given instance.
 func instanceProfile(sysOS *sys.OS, inst instance) (string, error) {
 	// Prepare raw.apparmor.
-	rawContent := ""
+	var rawContent strings.Builder
 	rawApparmor, ok := inst.ExpandedConfig()["raw.apparmor"]
 	if ok {
 		for line := range strings.SplitSeq(strings.Trim(rawApparmor, "\n"), "\n") {
-			rawContent += "  " + line + "\n"
+			rawContent.WriteString("  " + line + "\n")
```
roosterfish, line 153: Think you need a `rawContent.String()` down below when adding to the `map[string]any`.

roosterfish (dismissed):

shared/filter/clause.go
```diff
@@ -86,7 +86,7 @@ func Parse(s string, op OperatorSet) (*ClauseSet, error) {
 						break
 					}
 
-					value += " " + parts[index]
+					value += " " + parts[index] //nolint:perfsprint
```
tomponline: I think it wants you to use `strings.Builder` for `value`.

tomponline:
```go
var value strings.Builder
value.WriteString(parts[index])

for _, symbol := range op.Quote {
...
    value.WriteString(" " + parts[index])
...
}

clause.Value = value.String()

tomponline:
@markylaing looks like you've lost some implicit type `String()` conversion when switching away from `fmt.Sprintf()`, tests failing.

markylaing:
Wowzer the tests really blew up!

shared/filter/clause.go
markylaing: Couldn't quite work it this way because `value` is overwritten later. I've updated to use strings.Builder now

markylaing:
The CLA check is now failing :facepalm: Other tests should pass now :crossed_fingers:

simondeziel:
> The CLA check is now failing 🤦 Other tests should pass now 🤞

All green now. @markylaing OK to mark as ready for review?

tomponline (approved):
Ta!

Issue #15525, opened 2025-05-06
Instance POST changing target and project/pool cannot be mixed


edlerd:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

ubuntu

### Distribution version

snap

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
lxd         git-54945ae             33432  5.21/edge      canonical✓  in-cohort
```

### Output of "lxc info" or system info if it fails

```shell
-
```

### Issue description

Instance migration can only change target exclusively.

When changing instance target together with pool or project, the target change is ignored. The code path in [instancePost](https://github.com/canonical/lxd/blob/main/lxd/instance_post.go#L345) seems to exit early if project or pool is set in the payload and ignored the target.

This is the client logic in lxd-ui:

![Image](https://github.com/user-attachments/assets/c37869ca-9f67-4f5f-ae72-97a0dd76494c)

If target is set exclusively and pool and targetProject are undefined, it works.

Another observation is that on submitting the migration payload above without target, pool or project, the cluster member goes under heavy load.

### Steps to reproduce

1. Create an instance on a clustered backend
2. Request migration of the instance with migration payload sent to the endpoint mentioned above. Set a target and a project.
3. The instance is migrated to the new project, but the new target is ignored.
4. Expected is both project and target to change.

### Information to attach

- [ ] Any relevant kernel output (`dmesg`)
- [ ] Instance log (`lxc info NAME --show-log`)
- [ ] Instance configuration (`lxc config show NAME --expanded`)
- [ ] Main daemon log (at `/var/log/lxd/lxd.log` or `/var/snap/lxd/common/lxd/logs/lxd.log`)
- [ ] Output of the client with `--debug`
- [ ] Output of the daemon with `--debug` (or use `lxc monitor` while reproducing the issue)

tomponline:
@kadinsayani please can you evaulate this one and see 8f a bug?

kadinsayani:
> @kadinsayani please can you evaulate this one and see 8f a bug?

Sure I'll take a look!

kadinsayani:
Relevant Incus PR which reworks server-side migration: https://github.com/lxc/incus/pull/727. This PR makes it possible to change name, pool, project and/or location in one step.

kadinsayani:
I'll take a look at https://github.com/lxc/incus/pull/567 and https://github.com/lxc/incus/pull/696 as well.

PR #16862, opened 2025-10-30
Fix server-side instance migration (from Incus)

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/15525.

Depends on https://github.com/canonical/lxd/pull/16973.

This is a second attempt at fixing https://github.com/canonical/lxd/issues/15525. Rather than following the same approach as https://github.com/canonical/lxd/pull/15595 which included Incus cherry-picks which significantly reworked server-side migration, I've applied a minimal patch that addresses the issue while keeping the existing `instancePost*` helper functions intact.

I've also included tests that validate the patch.

### DIFF
--- a/test/suites/clustering.sh
+++ b/test/suites/clustering.sh
@@ -3770,7 +3770,7 @@ test_clustering_evacuation() {
   c7_location="$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c L c7)"
 
   echo "Verify that instance migration to an evacuated node is not allowed."
-  [[ "$(LXD_DIR="${LXD_TWO_DIR}" lxc move c5 --target=node1 2>&1)" == *"Error: The destination cluster member is evacuated"* ]]
+  [[ "$(LXD_DIR="${LXD_TWO_DIR}" lxc move c5 --target=node1 2>&1)" == *"Error: Migration operation failure: The destination cluster member is evacuated"* ]]
 
   echo 'Restore first node with "skip" mode.'
   # "skip" mode restores cluster member status without starting instances or migrating back evacuated instances.


tomponline:
@kadinsayani please rebase

lxc/move.go
```diff
@@ -179,77 +178,54 @@ func (c *cmdMove) run(cmd *cobra.Command, args []string) error {
...
-	}
-
-	// Support for server-side move. Currently, such migration can only move an instance to different project
-	// or storage pool. If specific profile, device or config is provided, the instance should be copied (move using copy).
-	if sourceRemote == destRemote && (c.flagStorage != "" || c.flagTargetProject != "") {
+		// Connect to the server.
 		source, err := conf.GetInstanceServer(sourceRemote)
 		if err != nil {
-			return err
+			return false
```
tomponline: @kadinsayani this looks suspect, what happens if there is an error here, we cant answer the question "isServerSide" so we need to return an error, rather than masking it as "is not server side".

lxc/move.go
```diff
@@ -427,11 +307,16 @@ func (c *cmdMove) moveInstance(conf *config.Config, sourceResource string, destR
 		Pool:         c.flagStorage,
 		Project:      c.flagTargetProject,
 		Live:         stateful,
-		Config:       inst.Config,
-		Devices:      inst.Devices,
 	}
 
-	// Overwrite profiles.
+	// Override profiles.
+	var profiles *[]string
```
tomponline: FWIW I'm not a fan of this type of declaration where we define it was a pointer and then have to keep dereferencing it below several times.

In my view it would be clearer to keep it as just `[]profiles` locally and then dereference it where its needed on line 321.

lxc/move.go
```diff
@@ -388,35 +290,13 @@ func (c *cmdMove) moveInstance(conf *config.Config, sourceResource string, destR
...
-	for k, m := range deviceMap {
-		if inst.Devices[k] == nil {
-			inst.Devices[k] = m
-			continue
-		}
-
-		maps.Copy(inst.Devices[k], m)
+	// Set the target if specified.
+	if c.flagTarget != "" {
+		source = source.UseTarget(c.flagTarget)
```
tomponline, line 301: @kadinsayani this caught my eye is `--target` used for both source and destination targetting?

lxd/instance_post.go
tomponline, line 1: > - When a move needs pool/project/config/profile changes, call
instancePostMigration once with targetMemberInfo=nil to copy locally,
then reload and run the cluster hop only if a different member is still
required.

"copy locally", do you mean "move locally"?

lxd/instance_post.go
```diff
@@ -191,10 +191,58 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	body, err := io.ReadAll(r.Body)
```
tomponline, line 194: > - Parse the POST body before cluster placement so the target project
used for limits.CheckTarget is the one in the request.

Can this be a separate commit, perhaps even a separate PR?

tomponline: As it seems like a (welcome) fix for a different bug.

lxc/move.go
kadinsayani: Yeah good point, I'll rework the closure to return `bool, error` and handle any error appropriately.

kadinsayani: This will also allow us to use `CheckExtension()` instead of `HasExtension()`.

lxd/instance_post.go
kadinsayani: Yes, thanks.

lxd/instance_post.go
kadinsayani: https://github.com/canonical/lxd/pull/16973

kadinsayani:
> I ran out of time today, but do I understand correctly that this PR is still doing pool/project/config/profile changes first and then moving cluster member as a 2nd step?
> 
> So if I do:
> 
> ```
> lxc move c1 --storage foo --target=member2
> ```
> 
> It will move c1 to foo storage pool (potentially a local copy) and then move it immediately from there onto member2 (potentially onto a different storage pool)?

Yes, that's correct, we handle local changes first and then cluster moves.

lxc/move.go
kadinsayani: Ack, I'll change the type and simplify the code that follows.

lxc/move.go
kadinsayani: The `--target` flag is only used for destination targeting. https://github.com/canonical/lxd/blob/ad98722480743b5c41b41d8a51784fa75e942711/client/lxd_server.go#L108-L108

tomponline:
> > I ran out of time today, but do I understand correctly that this PR is still doing pool/project/config/profile changes first and then moving cluster member as a 2nd step?
> > So if I do:
> > ```
> > lxc move c1 --storage foo --target=member2
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It will move c1 to foo storage pool (potentially a local copy) and then move it immediately from there onto member2 (potentially onto a different storage pool)?
> 
> Yes, that's correct, we handle local changes first and then cluster moves.

Is there a reason we cant handle it all as part of the migration? It seems inefficient to perform the pool and/or project migration locally and then copy it to another member, as that will likely cause more I/O than doing it as part of the migration.

kadinsayani:
> > > I ran out of time today, but do I understand correctly that this PR is still doing pool/project/config/profile changes first and then moving cluster member as a 2nd step?
> > > So if I do:
> > > ```
> > > lxc move c1 --storage foo --target=member2
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > It will move c1 to foo storage pool (potentially a local copy) and then move it immediately from there onto member2 (potentially onto a different storage pool)?
> > 
> > 
> > Yes, that's correct, we handle local changes first and then cluster moves.
> 
> Is there a reason we cant handle it all as part of the migration? It seems inefficient to perform the pool and/or project migration locally and then copy it to another member, as that will likely cause more I/O than doing it as part of the migration.

I've updated the logic to handle pool/config changes part of the cluster migration.

Project changes require creating a new instance DB record, while cluster moves try to work with an existing record. When attempted together we end up in a scenario where the destination creates a new record, but the source still has the old record. I tried adding a cleanup step that deletes the source record, but discovered that this conflicts with the storage cleanup part of `DeleteInstance()` (the storage cleanup queries the DB for volume records and fails if the instance record is already gone).

Handling project changes first via local migration creates a new instance record and deletes the old one, so the cluster migration can update the single record's node. This separates the create/delete records operation from the update record operation, avoiding DB conflicts.

lxd/instance_post.go
```diff
@@ -320,58 +320,111 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
...
 		// Server-side instance migration.
-		if req.Pool != "" || req.Project != "" {
-			// Check if user has access to target project.
-			if req.Project != "" {
-				err := s.Authorizer.CheckPermission(r.Context(), entity.ProjectURL(req.Project), auth.EntitlementCanCreateInstances)
+		hasConfigOverrides := len(req.Config) > 0 || len(req.Devices) > 0 || req.Profiles != nil || req.OverrideSnapshotProfiles
+		needsLocalCopy := req.Pool != "" || targetProjectName != inst.Project().Name || hasConfigOverrides
+		needsClusterMove := targetMemberInfo != nil
+
+		if needsClusterMove && targetMemberInfo != nil && inst.Location() == targetMemberInfo.Name && targetGroupName == "" && !needsLocalCopy {
```
tomponline: `&& targetMemberInfo != nil` is redundant as you checked this as part of `needsClusterMove := targetMemberInfo != nil`

tomponline: Also please can you add  a comment to this section explaining what is being checked here and the action taken?

lxd/instance_post.go
```diff
@@ -320,58 +320,111 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
 
 	if req.Migration {
 		// Server-side instance migration.
-		if req.Pool != "" || req.Project != "" {
-			// Check if user has access to target project.
-			if req.Project != "" {
-				err := s.Authorizer.CheckPermission(r.Context(), entity.ProjectURL(req.Project), auth.EntitlementCanCreateInstances)
+		hasConfigOverrides := len(req.Config) > 0 || len(req.Devices) > 0 || req.Profiles != nil || req.OverrideSnapshotProfiles
+		needsLocalCopy := req.Pool != "" || targetProjectName != inst.Project().Name || hasConfigOverrides
```
tomponline: please can you add a comment here explaining what `needsLocalCopy` means?

lxd/instance_post.go
```diff
@@ -320,58 +320,111 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
...
 				}
 			}
 
-			// Setup the instance move operation.
-			run := func(op *operations.Operation) error {
-				return instancePostMigration(s, inst, req, op)
+			// Pool/config changes (without project changes) can be handled directly in cluster migration to avoid copying locally then migrating over the network.
+			// Project changes require local copy first since cluster migration protocol assumes same project on source and dest.
+			if targetProjectName == inst.Project().Name {
+				needsLocalCopy = false
```
tomponline: why do we need to disable `needsLocalCopy` here, why was it marked as `true` in the first place if "Pool/config changes (without project changes) can be handled directly in cluster migration" cant we detect that where we first set `needsLocalCopy`?

lxd/instance_post.go
```diff
@@ -320,58 +320,111 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
...
 
-			resources := map[string][]api.URL{}
-			resources["instances"] = []api.URL{*api.NewURL().Path(version.APIVersion, "instances", name)}
-			op, err := operations.OperationCreate(r.Context(), s, projectName, operations.OperationClassTask, operationtype.InstanceMigrate, resources, nil, run, nil, nil)
-			if err != nil {
-				return response.InternalError(err)
+		if needsLocalCopy || needsClusterMove {
+			// Check if user has access to target project.
+			if req.Project != "" {
+				err := s.Authorizer.CheckPermission(r.Context(), entity.ProjectURL(req.Project), auth.EntitlementCanCreateInstances)
```
tomponline: why is this access check gated by `if needsLocalCopy || needsClusterMove {`, shouldnt we always check for access if `req.Project` is non-empty?

lxd/instance_post.go
```diff
@@ -320,58 +320,111 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
...
+					return instancePostMigration(r.Context(), s, currentInst, req, targetMemberInfo, targetGroupName, op)
+				}
 
-			if inst.Type() == instancetype.Container {
-				resources["containers"] = resources["instances"]
+				return nil
 			}
 
+			resources := map[string][]api.URL{}
+			resources["instances"] = []api.URL{*api.NewURL().Path(version.APIVersion, "instances", name)}
```
tomponline, line 438: It looks like you're phasing out `resources["containers"]` as part of this commit, but as that's an API change really it should be done as a separate PR so it can be clearly seen and acknowledge that this is what we want to do.

lxd/instance_post.go
```diff
@@ -320,58 +320,111 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
...
+			// Setup the instance move operation.
 			run := func(op *operations.Operation) error {
-				return migrateInstance(r.Context(), s, inst, targetMemberInfo.Name, targetGroupName, req, op)
-			}
+				currentInst := inst
+
+				// Handle local changes.
+				if needsLocalCopy {
+					// Pass a nil target member so the local copy phase never triggers a cluster move.
+					err := instancePostMigration(r.Context(), s, currentInst, req, nil, "", op)
```
tomponline: we should not ever use the request `r.Context()` inside the operation `fun` function as the original request may have finished by the time the operation calls this part of the `run` function and thus your run functional may fail prematurely.

tomponline: do you need the requester out of the request context for later event logging though?

lxd/api_cluster.go
```diff
@@ -3131,7 +3131,7 @@ func clusterNodeStatePost(d *Daemon, r *http.Request) response.Response {
 				Live: live,
 			}
 
-			err := migrateInstance(r.Context(), s, inst, targetMemberInfo.Name, "", req, op)
+			err := migrateInstance(r.Context(), s, inst, targetMemberInfo.Name, "", req, nil, op)
```
tomponline: It looks like the context used here needs to be reviewed and changed as an earlier PR because passing the `r.Context()` here looks like it could be problematic in the future.

kadinsayani: https://github.com/canonical/lxd/pull/17215

lxd/instance_post.go
kadinsayani: I've added comments and cleaned up the logic.

lxd/instance_post.go
kadinsayani: Done :)

lxd/instance_post.go
kadinsayani: I've reworked and reordered this logic for readability and clarity.

We now have the following steps for determining the required changes:

1. `hasInstanceChanges` is `true` when we have pool, project, or config changes. This variable is new, and is added for readability and simplifying the checks that follow.
2. `needsClusterMove` is `true` when a target member is specified and any of the following conditions are met: 
- The target member is different from the current location.
- A cluster group needs to be set.
- `hasInstanceChanges` is `true`.
3. By default, any instance changes require a local copy (`needsLocalCopy = hasInstanceChanges`). However, when performing a cluster move, only project changes require a local copy.

lxd/instance_post.go
kadinsayani: Yes, we should always check. I've moved this bit up.

lxd/instance_post.go
kadinsayani: I've re-added the `resources["containers"]` handling. We can tackle this separately.

lxd/instance_post.go
kadinsayani: I'll pass `context.TODO()` to `instancePostMigration` as a workaround until we have proper context propagation for operations. Currently, the `op` parameter already includes the request context.

lxd/instance_post.go
```diff
@@ -320,49 +320,116 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
...
-				}
-			}
+		hasConfigOverrides := len(req.Config) > 0 || len(req.Devices) > 0 || req.Profiles != nil || req.OverrideSnapshotProfiles
+		hasInstanceChanges := req.Pool != "" || targetProjectName != inst.Project().Name || hasConfigOverrides
 
-			// Setup the instance move operation.
-			run := func(op *operations.Operation) error {
-				return instancePostMigration(s, inst, req, op)
+		// Check if user has access to target project when changing projects.
+		if req.Project != "" {
```
tomponline: Can this use `targetProjectName != inst.Project().Name` so it benefits from the logic above:

```go
	targetProjectName := req.Project
	if targetProjectName == "" {
		targetProjectName = inst.Project().Name
	}
```

This would also avoid the need for an auth lookup if the project specified in req.Project is the same as instance project.

In general if you've defined something like `targetProjectName` higher up, then we should try and use that all the way down the function to avoid having to duplicate logic.

lxd/instance_post.go
```diff
@@ -320,49 +320,116 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
...
+			(inst.Location() != targetMemberInfo.Name || targetGroupName != "" || hasInstanceChanges)
+
+		// needsLocalCopy determines if we need to copy the instance locally on the current member before migrating.
+		// A local copy creates a new instance on the same member with the desired changes applied.
+		// This is a two-phase operation: first copy locally with changes, then migrate the new instance to the target member.
+		//
+		// By default, any instance changes (pool/project/config) require a local copy.
+		// However, when doing a cluster move, only project changes require a local copy because:
+		//   - The cluster migration protocol assumes the same project on both sides.
+		//   - Pool/config/device/profile changes can be applied directly during the cluster migration itself.
```
tomponline: Not a blocker, but why cant pool/config/device/profile changes not be applied during migration to another LXD cluster/server?

lxd/instance_post.go
```diff
@@ -320,49 +320,116 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
...
 		// Server-side instance migration.
-		if req.Pool != "" || req.Project != "" {
-			// Check if user has access to target project.
-			if req.Project != "" {
-				err := s.Authorizer.CheckPermission(r.Context(), entity.ProjectURL(req.Project), auth.EntitlementCanCreateInstances)
-				if err != nil {
-					return response.SmartError(err)
-				}
-			}
+		hasConfigOverrides := len(req.Config) > 0 || len(req.Devices) > 0 || req.Profiles != nil || req.OverrideSnapshotProfiles
```
tomponline: Is there any difference of meaning between non-nil zero length maps and nil maps for req.Config and req.Devices?

lxd/instance_post.go
```diff
@@ -490,17 +553,18 @@ func instancePostMigration(s *state.State, inst instance.Instance, req api.Insta
 	}
 
 	// Apply previous profiles, if provided profiles are nil.
-	if req.Profiles == nil {
-		req.Profiles = make([]string, 0, len(inst.Profiles()))
+	profileNames := req.Profiles
```
tomponline: why do we use a new var here rather than req.Profiles like previously?

lxd/instance_post.go
```diff
@@ -454,26 +521,22 @@ func instancePost(d *Daemon, r *http.Request) response.Response {
...
-	}
+	sourceName := inst.Name()
+	sourceProject := inst.Project().Name
 
-	statefulStart := false
-	if inst.IsRunning() {
-		if !req.Live {
-			return api.StatusErrorf(http.StatusBadRequest, "Instance must be stopped to move between pools statelessly")
-		}
+	finalProject := req.Project
```
tomponline: why do we use new vars here rather than modifying the copy of req that is passed to this function?

lxd/instance_post.go
```diff
@@ -490,17 +553,18 @@ func instancePostMigration(s *state.State, inst instance.Instance, req api.Insta
...
 		for _, p := range inst.Profiles() {
-			req.Profiles = append(req.Profiles, p.Name)
+			profileNames = append(profileNames, p.Name)
 		}
 	}
 
 	apiProfiles := []api.Profile{}
-	if len(req.Profiles) > 0 {
+	if len(profileNames) > 0 {
 		err := s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
```
tomponline: this can take `ctx` now right?

lxd/instance_post.go
```diff
@@ -631,8 +712,8 @@ func instancePostMigration(s *state.State, inst instance.Instance, req api.Insta
 	}
 
 	// Rename copy from temporary name to original name if needed.
-	if req.Name == inst.Name() && req.Project == inst.Project().Name {
-		err = targetInst.Rename(req.Name, false) // Don't apply templates when moving.
+	if finalName == sourceName && finalProject == sourceProject {
```
tomponline: this logic of `finalName == sourceName && finalProject == sourceProject ` is the same as above line 677, can we combine statements?

tomponline: or at least define the concept in a new var once?

lxd/instance_post.go
kadinsayani: For a slight readability improvement. We can just modify the copy of req though.

lxd/instance_post.go
kadinsayani: For readability. We can just use req.Profiles though. I think the comments suffice for ensuring readability.

lxd/instance_post.go
kadinsayani: They can be applied during a cluster migration, and they will be. Perhaps this comment and the assignment to `needsLocalCopy` that follows is not clear enough. I'll rework this bit.

lxd/instance_post.go
kadinsayani: Yes, we can use that check. Great suggestion, thanks.

lxd/instance_post.go
kadinsayani: Yeah sounds good, we can define the concept once.

lxd/instance_post.go
kadinsayani: There is no semantic difference, but checking for nil maps may improve readability.

tomponline:
@kadinsayani please can you rebase

kadinsayani:
> @kadinsayani please can you rebase

Rebased.

tomponline:
@kadinsayani please rebase

PR #12626, opened 2023-12-06
Fail fast when instance name already exist on create

MusicDin:
If creating an instance from an image, check if instance with the given name already exist before pulling an image from remote.

```sh
$ lxc launch ubuntu:22.04 c1
Creating c1
*Download progress*
Starting c1

$ lxc launch ubuntu:18.04 c1
Creating c1
Error: Failed instance creation: Instance "c1" already exists
```

Fixes #12554

PR #16863, opened 2025-10-30
Check instance name early to fail fast on instance creation and prevent image download

### LINKED ISSUE #12554: `lxc launch` doesn't immediately fail on duplicated instance name if the source image is not cached
If you already have an instance named `foo`, trying to created another one with the same name will not immediately fail if the source is not cached.

Steps to reproduce:

```
$ fingerprint="$(lxc image list -f csv -c lf ubuntu-daily:24.04 type=container | head -n1 | cut -d, -f2)"

$ lxc init "ubuntu-daily:$fingerprint" foo   # create instance and populate local image cache
Creating foo
$ lxc init "ubuntu-daily:$fingerprint" foo   # fail early, good
Creating foo
Error: Failed creating instance record: Add instance info to the database: This "instances" entry already exists

$ lxc image delete "local:$fingerprint"    # remove the locally cached image


$ lxc init "ubuntu-daily:$fingerprint" foo  # fails late after retrieving the image
Creating foo
Retrieving image: metadata: 100% (1.21GB/s)
Retrieving image: rootfs: 1% (2.86MB/s)     
...
Error: Failed creating instance record: Add instance info to the database: This "instances" entry already exists
```

Ideally, the check for duplicated name should be done prior to doing the expensive image fetch.

kadinsayani:
Fixes #12554.

This fix uses uses `GetInstanceID()` so we don't fetch all instances and risk an expensive query on a non-leader node.

### DIFF
--- a/test/suites/basic.sh
+++ b/test/suites/basic.sh
@@ -843,7 +843,7 @@ test_duplicate_detection() {
   [ "$(! "${_LXC}" image alias create testimage "${test_image_fingerprint}" 2>&1 1>/dev/null)" = 'Error: Alias "testimage" already exists' ]
 
   lxc init foo --empty
-  [ "$(! "${_LXC}" init foo --empty 2>&1 1>/dev/null)" = 'Error: Failed creating instance record: Instance "foo" already exists' ]
+  [ "$(! "${_LXC}" init foo --empty 2>&1 1>/dev/null)" = 'Error: Instance "foo" already exists' ]
   lxc init bar --empty
   [ "$(! "${_LXC}" rename bar foo 2>&1 1>/dev/null)" = 'Error: Name "foo" already in use' ]
   lxc delete bar


lxd/instances_post.go
```diff
@@ -91,6 +91,26 @@ func createFromImage(ctx context.Context, s *state.State, p api.Project, profile
 		return response.BadRequest(err)
 	}
 
+	err = s.DB.Cluster.Transaction(ctx, func(ctx context.Context, tx *db.ClusterTx) error {
+		// Check that the name isn't already in use.
+		existingID, err := tx.GetInstanceID(ctx, p.Name, req.Name)
+		if err != nil {
```
tomponline: api.StatusErrorCheck

tomponline (changes requested):
Although cluster move tests failing - because the instance already exists for an intra-cluster move.

simondeziel:
This seems to be lighter than what was initially proposed in #12626 but I'm still concerned by the additional DB query (see https://github.com/canonical/lxd/pull/12626#pullrequestreview-1830131051). Do I understand that currently, LXD only consults the DB once it is ready to expand an image into an instance?

If so, maybe the issue would be better closed as "won't fix, works as expected/designed"?

tomponline:
> This seems to be lighter than what was initially proposed in #12626 but I'm still concerned by the additional DB query (see [#12626 (review)](https://github.com/canonical/lxd/pull/12626#pullrequestreview-1830131051)). Do I understand that currently, LXD only consults the DB once it is ready to expand an image into an instance?
> 
> If so, maybe the issue would be better closed as "won't fix, works as expected/designed"?

This is true, it will incur a cost, but is much reduced from the original proposal as it's reuses an existing transaction and only gets the ID.

So I'm ok with that cost for the benefit of better user experience.

lxd/instances_post.go
```diff
@@ -1247,6 +1247,16 @@ func instancesPost(d *Daemon, r *http.Request) response.Response {
 			return err
 		}
 
+		// Check that the name isn't already in use.
+		if !req.Source.Refresh && !clusterNotification && req.Source.Source == "" {
+			existingID, err := tx.GetInstanceID(ctx, targetProjectName, req.Name)
+			if err != nil && !api.StatusErrorCheck(err, http.StatusNotFound) {
+				return fmt.Errorf("Failed checking for existing instance: %w", err)
+			} else if existingID > 0 {
+				return api.StatusErrorf(http.StatusConflict, "Failed creating instance record: Instance %q already exists", req.Name)
```
tomponline: Can we drop the "Failed creating instance record" prefix?

lxd/instances_post.go
```diff
@@ -1247,6 +1247,16 @@ func instancesPost(d *Daemon, r *http.Request) response.Response {
 			return err
 		}
 
+		// Check that the name isn't already in use.
+		if !req.Source.Refresh && !clusterNotification && req.Source.Source == "" {
```
tomponline, line 1252: can you add a comment here explaining why the logic is here?

tomponline:
DCO checks failing

kadinsayani:
> DCO checks failing

Oops, hopping between dev environments. Should be good now.

tomponline (approved):

PR #16865, opened 2025-10-31
Network: Add per-member `acceleration.parent` setting to `ovn` networks

tomponline:
- Adds `acceleration.parent` to `ovn` networks - this is inherited by `ovn` NICs connected to that network if they have `acceleration` enabled (and they themselves don't override `acceleration.parent` at the NIC device level).
- Makes `acceleration.parent` a per-cluster member setting - so that these interfaces can be different on each member.
- Adds restriction to prevent using NIC device level `acceleration.parent` setting in restricted projects unless low-level settings are allowed.
- Improves validation of `acceleration.parent` settings to be a list of acceptable interface names.

### DIFF
--- a/doc/metadata.txt
+++ b/doc/metadata.txt
@@ -800,11 +800,12 @@ See {ref}`devices-nic-hw-acceleration` for more information.
 ```
 
 ```{config:option} acceleration.parent device-nic-ovn-device-conf
-:managed: "no"
+:managed: "yes"
 :shortdesc: "Physical function interfaces to allocate virtual functions from for hardware acceleration"
 :type: "string"
-Comma separated list of physical function (PF) interfaces to allocate virtual functions (VFs) from for hardware acceleration.
-If this is not specified, and `acceleration` is enabled then all PFs connected to the OVS integration bridge are scanned for a free VF.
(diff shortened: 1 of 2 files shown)


lxd/db/networks.go
```diff
@@ -965,4 +965,5 @@ var NodeSpecificNetworkConfig = []string{
 	"bgp.ipv6.nexthop",
 	"bridge.external_interfaces",
 	"parent",
+	"acceleration.parent",
```
roosterfish, line 968: Before this option was added to the list, was it required that the same interface (or many) are present on all members of the cluster in case `acceleration.parent` was set on the network?

tomponline: Not quite. Before this option (and still applicable if the option isn't specified) if `acceleration={sriov|vdap}` then LXD would search the OVS integration bridge (`br-int`) for physical ports attached to it and then allocate VFs from there.

This was a problem if you wanted to use the physical port for something else (like host connectivity).

roosterfish: I see thanks. From the description I understood you have to list actual interfaces under `acceleration.parent` from which VFs get allocated.

tomponline: you do indeed. Which allows you to still use VFs for OVN NICs, and still use the main PF port for connectivity as it doesn't need to be added to the integration bridge.

PR #16867, opened 2025-10-31
gomod: Update dependencies

simondeziel:


simondeziel:
VS Code suggested running `go mod tidy` to fix some deps being indirect while they should have been direct.

tomponline (approved):

PR #16868, opened 2025-10-31
github: sanitize artifact names in `snap-tests`

simondeziel:


simondeziel:
@tomponline I checked that the action step is ran for the known problematic OVN snap test.

PR #16871, opened 2025-10-31
Fix remote cluster migration

kadinsayani:
https://github.com/canonical/lxd/pull/16478/commits/cafc3e4899d2643bb544f496ef02d56d641fbd29 introduced cluster notifications for remote cluster migrations, conflicting with the following logic: https://github.com/canonical/lxd/blob/781c64e946ef983784ba82c86bbb0685a2fd1b37/lxd/instances_post.go#L245-L253

This pull request fixes the issue by refining the logic for identifying cluster move requests and adds a check to ensure remote cluster migration works.

### DIFF
--- a/lxd/instances_post.go
+++ b/lxd/instances_post.go
@@ -244,11 +244,7 @@ func createFromMigration(ctx context.Context, s *state.State, projectName string
 
 	// Decide if this is an internal cluster move request.
 	var clusterMoveSourceName string
-	if isClusterNotification {
-		if req.Source.Source == "" {
-			return response.BadRequest(errors.New("Source instance name must be provided for cluster member move"))
-		}
-
+	if isClusterNotification && req.Source.Source != "" {
 		clusterMoveSourceName = req.Source.Source
 	}
 


test/suites/clustering_move.sh
```diff
@@ -153,9 +153,13 @@ test_clustering_move() {
   lxc move cluster:c2 --target @foobar1
   lxc move cluster:c3 --target node1
 
+  echo "c4 can be migrated from local cluster to remote cluster"
+  lxc init --empty c4
+  lxc move c4 cluster:c5 --target node1
```
simondeziel, line 158: For simplicity and ease of tracking, can we keep the `c4` name with just `lxc move c4 cluster: --target node1`?

kadinsayani: We need a new name because the remote cluster is the local cluster in `test_clustering_move()`. It's a bit funky but it works to test the expected behavior.

tomponline, line 158: @kadinsayani does targeting node1 ensure an internal forward occurs?

kadinsayani: Yes, confirmed by testing with and without the fix.

PR #16874, opened 2025-10-31
doc: refactor linkcheck exceptions

minaelee:
This PR refactors Sphinx configuration for linkchecks to:
- Ignore all localhost/127.0.0.1 links
- Improve URL regexes to ignore entire domains where the entire domain is the issue
- Improve comments to explain why/where certain sets are ignored
- `web.libera.chat/#lxd` only fails because of a false anchor issue, moved to `custom_linkcheck_anchors_ignore_for_url` list
- Some links only fail when run from GH CI but pass locally, set to only ignore in GH CI so we can still check these locally

### DIFF
--- a/doc/custom_conf.py
+++ b/doc/custom_conf.py
@@ -167,40 +167,38 @@ else:
 ### Link checker
 ############################################################
 
-# Links to ignore when checking links
+# Always ignore these links
 linkcheck_ignore = [
-    'https://127.0.0.1:8443',
-    'https://127.0.0.1:8443/1.0',
-    'https://localhost:8443',
-    'https://web.libera.chat/#lxd',
-    'http://localhost:8000',
-    'http://localhost:8080',
-    'http://localhost:8080/admin',
-    r'/lxd/latest/api/.*',
-    r'/api/.*',
-    # Those links may fail from time to time
-    'https://www.dell.com/',
-    'https://www.dell.com/en-us/shop/powerflex/sf/powerflex',
-    'https://www.gnu.org/licenses/agpl-3.0.en.html',
-    r"https://ceph\.io(/.*)?",
-    # Blocked from GH runners
-    'https://www.schlachter.tech/solutions/pongo2-template-engine/',
-    r"https://.*\.sourceforge\.net/.*",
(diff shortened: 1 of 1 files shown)


doc/custom_conf.py
```diff
@@ -167,40 +167,40 @@
...
-    'https://www.hpe.com/emea_europe/en/hpe-alletra.html',
-    r'https://.*canonical\.com/.*',
-    r'https://snapcraft\.io/.*',
-    r'https://ubuntu\.com/.*',
-    r'https://bugs\.launchpad\.net/.*',
-    r'https://microcloud\.is/.*',
+    r"https?://localhost.*",
+    r"https?://127\.0\.0\.1.*",
+    r"/lxd/latest/api.*",
+    r"/api.*",
```
simondeziel: Do we need the first considering the last one matches more broadly?

doc/custom_conf.py
```diff
@@ -167,40 +167,40 @@
...
+    # These links often/always fail both locally and in GitHub CI
+    r"https://ceph\.io.*",
+    r"https://.*\.sourceforge\.net.*",
+    r"https://www.\.gnu\.org.*",
+    # These links often fail due to infra issues
+    r"https://.*\.canonical\.com.*",
+    r"https://snapcraft\.io.*",
+    r"https://ubuntu\.com.*",
+    r"https://.*\.launchpad\.net.*",
+    r"https://microcloud\.is.*",
```
simondeziel: This shouldn't be used AFAIK as it redirects to https://canonical.com/microcloud

doc/custom_conf.py
minaelee: I'm not sure why it was added originally actually, since that's a RTD-generated URL fragment (`lxd/latest`) and we don't linkcheck in the RTD build. (Also, if it was needed, why wouldn't the other docs versions also be listed?) I think we can live without it.

doc/custom_conf.py
minaelee: Thanks, removed

tomponline (commented):
It would have been better to include the extensive pr description in the commit message too. For future reference.

Thanks!

PR #16882, opened 2025-11-03
devlxd: Add support for operations

MusicDin:
This PR adds support to wait for an operation and cancel (delete) an operation through DevLXD.

It is cherry picked from https://github.com/canonical/lxd/pull/16842 (DevLXD volume cloning) and is needed to seamlessly shift existing DevLXD endpoints to retrieve an operation (when we make LXD endpoints `POST/PUT /1.0/storage-pools/<pool>/volumes/custom{/<vol>}` asynchronous).

### DIFF
--- a/client/devlxd_operations.go
+++ b/client/devlxd_operations.go
@@ -69,7 +69,8 @@ func (r *ProtocolDevLXD) GetOperationWait(uuid string, timeout int) (*api.DevLXD
 
 // DeleteOperation deletes (cancels) a running operation.
 func (r *ProtocolDevLXD) DeleteOperation(uuid string) error {
-	_, _, err := r.query(http.MethodDelete, "/operations/"+url.PathEscape(uuid), nil, "")
+	url := api.NewURL().Path("operations", url.PathEscape(uuid))
+	_, _, err := r.query(http.MethodDelete, url.String(), nil, "")
 	if err != nil {
 		return err
 	}
--- a/lxd-agent/devlxd.go
+++ b/lxd-agent/devlxd.go
@@ -57,6 +57,7 @@ var devLXDEndpoints = []devLXDAPIEndpoint{
 	devLXDDevicesEndpoint,
 	devLXDImageExportEndpoint,
 	devLXDInstanceEndpoint,
+	devLXDOperationEndpoint,
 	devLXDOperationWaitEndpoint,
 	devLXDStoragePoolEndpoint,
 	devLXDStoragePoolVolumeTypeEndpoint,
--- a/lxd-agent/devlxd_operation.go
+++ b/lxd-agent/devlxd_operation.go
@@ -11,6 +11,11 @@ var devLXDOperationWaitEndpoint = devLXDAPIEndpoint{
 	Get:  devLXDAPIEndpointAction{Handler: devLXDOperationWaitHandler},
 }
 
+var devLXDOperationEndpoint = devLXDAPIEndpoint{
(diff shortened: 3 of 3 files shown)


client/devlxd_operations.go
```diff
@@ -0,0 +1,21 @@
...
+	"strconv"
+
+	"github.com/canonical/lxd/shared/api"
+)
+
+// GetOperationWait returns a DevLXDOperation entry for the provided uuid once it's complete or hits the timeout.
+func (r *ProtocolDevLXD) GetOperationWait(uuid string, timeout int) (*api.DevLXDOperation, string, error) {
+	var op api.DevLXDOperation
+
+	etag, err := r.queryStruct(http.MethodGet, "/operations/"+url.PathEscape(uuid)+"/wait?timeout="+strconv.FormatInt(int64(timeout), 10), nil, "", &op)
```
tomponline: Please use ` api.NewURL()` like you did in the earlier commits to build URLs.

client/devlxd_operations.go
```diff
@@ -19,3 +19,13 @@ func (r *ProtocolDevLXD) GetOperationWait(uuid string, timeout int) (*api.DevLXD
 
 	return &op, etag, nil
 }
+
+// DeleteOperation deletes (cancels) a running operation.
+func (r *ProtocolDevLXD) DeleteOperation(uuid string) error {
+	_, _, err := r.query(http.MethodDelete, "/operations/"+url.PathEscape(uuid), nil, "")
```
tomponline: api.NewURL()

client/devlxd_operations.go
```diff
@@ -1,13 +1,89 @@
...
+	return api.Operation{
+		ID:         op.ID,
+		Status:     op.Status,
+		StatusCode: op.StatusCode,
+		Err:        op.Err,
+	}
+}
+
+// Wait lets you wait until the operation reaches a final state.
+func (op *devLXDOperation) Wait() error {
```
tomponline: Do we need to add the non-context legacy function?

MusicDin: We do not necessarily need it, but `Operation` interface exposes it.

```go
type Operation interface {
    ...
    Wait() (err error)
    WaitContext(ctx context.Context) error
}
```

tomponline: ah makes sense to keep it then

MusicDin: I’ve revisited the current implementation and am considering introducing a `DevLXDOperation` interface on the client side:

```go
type DevLXDOperation interface {
	Get() api.DevLXDOperation
	Cancel() error
	WaitContext(context.Context) error
}
```

Compared to the LXD client’s `Operation` interface, this version would rely on `api.DevLXDOperation` instead of `api.Operation`, and allow us to drop currently unnecessary methods (handlers, legacy wait function, etc.).

At the moment, the DevLXD API operation type is `api.DevLXDOperation`, but the DevLXD client uses LXD’s `Operation` interface, which depends on `api.Operation`. This mismatch requires an extra type conversion in the LXD agent.

Alternatively, we could switch DevLXD to use `api.Operation` to match with the main LXD client, but that would reintroduce inconsistency since all other DevLXD responses use `DevLXD*` types.

tomponline: > this version would rely on api.DevLXDOperation instead of api.Operation, and allow us to drop currently unnecessary methods (handlers, legacy wait function, etc.).

Sounds good.

tomponline (approved):

PR #16883, opened 2025-11-03
github: do not reclaim disk space when using `ceph`

simondeziel:


simondeziel:
@tomponline despite some of the problems with QEMU crashing, the changes at hand do work as expected as the `storage-vm ceph` [test](https://github.com/canonical/lxd/actions/runs/19040253440/job/54375996944?pr=16883) no longer tries to reclaim disk space for no reason.

.github/workflows/tests.yml
```diff
@@ -539,7 +539,9 @@ jobs:
         uses: ./.github/actions/reclaim-memory
 
       - name: Reclaim disk space
-        if: ${{ matrix.test == 'conversion' || startsWith(matrix.test, 'storage-') || matrix.test == 'vm-nesting' }}
+        # This can take multiple minutes if the runner is busy/slow so only run it when/where absolutely needed.
```
tomponline, line 542: Can we use the ephemeral disk for all storage drivers by mounting the relevant disks directory onto the ephemeral drive btw?

simondeziel: > Can we use the ephemeral disk for all storage drivers by mounting the relevant disks directory onto the ephemeral drive btw?

I was thinking of always repurposing the ephemeral disk to use it as a block device fed as the `source` for the main pool used by some tests. Is that what you had in mind or were you actually meaning a "directory" inside which we'd host the loop files?

If going with the block device way, the `dir` backend it would require a `mkfs.ext4`.

tomponline: > were you actually meaning a "directory" inside which we'd host the loop files?

Yes this because I understand the tests created multiple pools so would require multiple block devices otherwise.

PR #16884, opened 2025-11-04
Fix tiobe tics for go 1.25

kadinsayani:
This pull request drops usage of `gocov` and `gocov-xml` in favor of `gocover-cobertura`, which appears to be more actively maintained and supports go1.25. We need coverage data in Cobertura XML format for TICS.

Test run: https://github.com/kadinsayani/lxd/actions/runs/19070707294 (this will probably fail due to other failing tests, but I've also tested locally with `GOCOVERDIR=<coverdir> make check-unit`).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -813,15 +813,13 @@ jobs:
 
       - name: Install dependencies
         run: |
-          go install github.com/axw/gocov/gocov@latest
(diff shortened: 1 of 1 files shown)


simondeziel (approved):
Awesome!

tomponline:
Ta thanks for looking into this

PR #16887, opened 2025-11-04
lxd: Use api.NewURL() for path encoding

kadinsayani:
Closes https://github.com/canonical/lxd/issues/16792.

### DIFF
--- a/lxd/instance_backup.go
+++ b/lxd/instance_backup.go
@@ -184,9 +184,7 @@ func instanceBackupsGet(d *Daemon, r *http.Request) response.Response {
 		}
 
 		if !recursion {
-			url := fmt.Sprintf("/%s/instances/%s/backups/%s",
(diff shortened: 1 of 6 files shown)


tomponline (approved):
Cool! Was that all of the remaining ones?

kadinsayani:
> Cool! Was that all of the remaining ones?

This PR covered the remaining ones in our API handlers. There's still `fmt.Sprintf()` usage without path encoding elsewhere, such as in operations metadata.

tomponline:
@kadinsayani I think we need to ensure all URLs are properly path encoded in our API output. 

So the ones you mention likely need fixing too.

kadinsayani:
> @kadinsayani I think we need to ensure all URLs are properly path encoded in our API output. 
> 
> 
> 
> So the ones you mention likely need fixing too. 

Sounds good, I'll double check those today.

PR #16895, opened 2025-11-05
Feature : Alias show - Alias edit

tugbataluy:
What This PR Does

This PR adds two new commands to improve the management of client command aliases:

- lxc alias show: Displays all current aliases in a structured YAML format to stdout.
- lxc alias edit: Allows bulk, simultaneous editing of all aliases. It supports interactive editing via the system editor and piping of configuration via stdin.

 lxc alias edit: Allows bulk, simultaneous editing of all aliases. It supports interactive editing via the system editor and piping of configuration via stdin.

Why This Is Needed

These commands replace the inefficient process of adding or removing aliases one-by-one, enabling quick auditing and configuration deployment through scripting or a single editing session.

Implementation Highlights

-     Consistency: Both commands align with standard LXC configuration commands (like lxc profile edit)
-     Safety: lxc alias edit includes a required validation loop for interactive edits to prevent saving invalid YAML syntax.
-     Compatibility: This implementation fully relies on the existing lxc alias commands and c.global.configPath for persistence in the standard YAML configuration file.

Example Usage
 ```
 lxc alias show
 lxc alias edit
 cat aliases.yaml | lxc alias edit 
```

### DIFF
--- a/po/ar.po
+++ b/po/ar.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: lxd\n"
 "Report-Msgid-Bugs-To: lxd@lists.canonical.com\n"
-"POT-Creation-Date: 2025-12-10 17:19+0000\n"
+"POT-Creation-Date: 2025-12-11 15:59+0300\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Automatically generated\n"
 "Language-Team: none\n"
@@ -93,6 +93,19 @@ msgid ""
 "### Fields \"data\", \"timestamp\", \"digest\" are HEX-encoded.\n"
 "### Field \"attr\" is an unsigned 32-bit integer.\n"
 "###"
+msgstr ""
+
+#: lxc/alias.go:334
+msgid ""
+"### This is a YAML representation of the aliases.\n"
+"### Any line starting with a '#' will be ignored.\n"
+"###\n"
+"### A sample aliases configuration looks like:\n"
+"### list: \"list -c ns46S\"\n"
+"### my-list: \"list -c ns46S\"\n"
+"### start-all: \"start --all\"\n"
+"###\n"
+"### Note that aliases are key-value pairs."
 msgstr ""
 
 #: lxc/config_trust.go:244
@@ -517,15 +530,15 @@ msgstr ""
 msgid "--target cannot be used with instances"
 msgstr ""
 
-#: lxc/alias.go:211
+#: lxc/alias.go:225
 msgid "<alias>"
 msgstr ""
 
-#: lxc/alias.go:58
+#: lxc/alias.go:72
 msgid "<alias> <target>"
 msgstr ""
 
-#: lxc/alias.go:156
+#: lxc/alias.go:170
 msgid "<old alias> <new alias>"
 msgstr ""
 
@@ -558,7 +571,7 @@ msgstr ""
 msgid "ADDRESS"
 msgstr ""
 
-#: lxc/alias.go:139 lxc/image.go:1144 lxc/image_alias.go:267
+#: lxc/alias.go:153 lxc/image.go:1144 lxc/image_alias.go:267
 msgid "ALIAS"
 msgstr ""
 
@@ -653,7 +666,7 @@ msgstr ""
 msgid "Add member to group"
 msgstr ""
 
-#: lxc/alias.go:59 lxc/alias.go:60
+#: lxc/alias.go:73 lxc/alias.go:74
 msgid "Add new aliases"
 msgstr ""
 
@@ -738,18 +751,23 @@ msgstr ""
 msgid "Admin secret key: %s"
 msgstr ""
 
-#: lxc/alias.go:85 lxc/alias.go:190
+#: lxc/alias.go:99 lxc/alias.go:204
 #, c-format
 msgid "Alias %s already exists"
 msgstr ""
 
-#: lxc/alias.go:184 lxc/alias.go:239
+#: lxc/alias.go:198 lxc/alias.go:253
 #, c-format
 msgid "Alias %s doesn't exist"
 msgstr ""
 
 #: lxc/image_alias.go:101 lxc/image_alias.go:156 lxc/image_alias.go:320
 msgid "Alias name missing"
+msgstr ""
+
+#: lxc/alias.go:409
+#, c-format
+msgid "Alias parsing error: %s"
 msgstr ""
 
 #: lxc/publish.go:256
@@ -1728,21 +1746,21 @@ msgid "Deleted OIDC session %q"
 msgstr ""
 
 #: lxc/action.go:33 lxc/action.go:58 lxc/action.go:84 lxc/action.go:111
-#: lxc/alias.go:23 lxc/alias.go:60 lxc/alias.go:110 lxc/alias.go:159
-#: lxc/alias.go:214 lxc/auth.go:37 lxc/auth.go:69 lxc/auth.go:108
-#: lxc/auth.go:162 lxc/auth.go:211 lxc/auth.go:340 lxc/auth.go:400
-#: lxc/auth.go:449 lxc/auth.go:501 lxc/auth.go:524 lxc/auth.go:583
-#: lxc/auth.go:777 lxc/auth.go:820 lxc/auth.go:1020 lxc/auth.go:1088
-#: lxc/auth.go:1148 lxc/auth.go:1210 lxc/auth.go:1337 lxc/auth.go:1394
-#: lxc/auth.go:1418 lxc/auth.go:1470 lxc/auth.go:1524 lxc/auth.go:1550
-#: lxc/auth.go:1612 lxc/auth.go:1670 lxc/auth.go:1692 lxc/auth.go:1874
-#: lxc/auth.go:1912 lxc/auth.go:1964 lxc/auth.go:2013 lxc/auth.go:2132
-#: lxc/auth.go:2192 lxc/auth.go:2241 lxc/auth.go:2292 lxc/auth.go:2315
-#: lxc/auth.go:2368 lxc/auth.go:2431 lxc/auth.go:2457 lxc/auth.go:2510
-#: lxc/auth.go:2582 lxc/cluster.go:33 lxc/cluster.go:126 lxc/cluster.go:218
-#: lxc/cluster.go:275 lxc/cluster.go:334 lxc/cluster.go:407 lxc/cluster.go:489
-#: lxc/cluster.go:533 lxc/cluster.go:591 lxc/cluster.go:682 lxc/cluster.go:776
-#: lxc/cluster.go:899 lxc/cluster.go:983 lxc/cluster.go:1093
+#: lxc/alias.go:29 lxc/alias.go:74 lxc/alias.go:124 lxc/alias.go:173
+#: lxc/alias.go:228 lxc/alias.go:274 lxc/auth.go:37 lxc/auth.go:69
+#: lxc/auth.go:108 lxc/auth.go:162 lxc/auth.go:211 lxc/auth.go:340
+#: lxc/auth.go:400 lxc/auth.go:449 lxc/auth.go:501 lxc/auth.go:524
+#: lxc/auth.go:583 lxc/auth.go:777 lxc/auth.go:820 lxc/auth.go:1020
+#: lxc/auth.go:1088 lxc/auth.go:1148 lxc/auth.go:1210 lxc/auth.go:1337
+#: lxc/auth.go:1394 lxc/auth.go:1418 lxc/auth.go:1470 lxc/auth.go:1524
+#: lxc/auth.go:1550 lxc/auth.go:1612 lxc/auth.go:1670 lxc/auth.go:1692
+#: lxc/auth.go:1874 lxc/auth.go:1912 lxc/auth.go:1964 lxc/auth.go:2013
+#: lxc/auth.go:2132 lxc/auth.go:2192 lxc/auth.go:2241 lxc/auth.go:2292
+#: lxc/auth.go:2315 lxc/auth.go:2368 lxc/auth.go:2431 lxc/auth.go:2457
+#: lxc/auth.go:2510 lxc/auth.go:2582 lxc/cluster.go:33 lxc/cluster.go:126
+#: lxc/cluster.go:218 lxc/cluster.go:275 lxc/cluster.go:334 lxc/cluster.go:407
+#: lxc/cluster.go:489 lxc/cluster.go:533 lxc/cluster.go:591 lxc/cluster.go:682
+#: lxc/cluster.go:776 lxc/cluster.go:899 lxc/cluster.go:983 lxc/cluster.go:1093
 #: lxc/cluster.go:1181 lxc/cluster.go:1305 lxc/cluster.go:1342
 #: lxc/cluster_group.go:33 lxc/cluster_group.go:87 lxc/cluster_group.go:177
 #: lxc/cluster_group.go:263 lxc/cluster_group.go:323 lxc/cluster_group.go:447
@@ -2062,6 +2080,10 @@ msgstr ""
 
 #: lxc/cluster_group.go:322 lxc/cluster_group.go:323
 msgid "Edit a cluster group"
+msgstr ""
+
+#: lxc/alias.go:319
+msgid "Edit aliases"
 msgstr ""
 
 #: lxc/auth.go:1209 lxc/auth.go:1210
@@ -2609,7 +2631,7 @@ msgid ""
 "Are you really sure you want to force removing %s? (yes/no): "
 msgstr ""
 
-#: lxc/alias.go:112 lxc/auth.go:344 lxc/auth.go:1024 lxc/auth.go:2136
+#: lxc/alias.go:126 lxc/auth.go:344 lxc/auth.go:1024 lxc/auth.go:2136
 #: lxc/auth.go:2514 lxc/cluster.go:128 lxc/cluster.go:984
 #: lxc/cluster_group.go:449 lxc/config_template.go:275 lxc/config_trust.go:351
 #: lxc/config_trust.go:433 lxc/image.go:1119 lxc/image_alias.go:182
@@ -3034,6 +3056,11 @@ msgstr ""
 msgid "Import storage volumes"
 msgstr ""
 
+#: lxc/alias.go:386
+#, c-format
+msgid "Imported: %d alias(es)\n"
+msgstr ""
+
 #: lxc/storage_volume.go:2965
 #, c-format
 msgid "Importing custom volume: %s"
@@ -3322,7 +3349,7 @@ msgstr ""
 msgid "List OIDC sessions"
 msgstr ""
 
-#: lxc/alias.go:109 lxc/alias.go:110
+#: lxc/alias.go:123 lxc/alias.go:124
 msgid "List aliases"
 msgstr ""
 
@@ -3774,7 +3801,7 @@ msgstr ""
 msgid "Manage cluster roles"
 msgstr ""
 
-#: lxc/alias.go:22 lxc/alias.go:23
+#: lxc/alias.go:28 lxc/alias.go:29
 msgid "Manage command aliases"
 msgstr ""
 
@@ -4712,11 +4739,11 @@ msgstr ""
 msgid "Press ctrl+c to finish"
 msgstr ""
 
-#: lxc/auth.go:309 lxc/auth.go:1306 lxc/auth.go:2101 lxc/cluster.go:866
-#: lxc/cluster_group.go:405 lxc/config.go:323 lxc/config.go:398
-#: lxc/config.go:1381 lxc/config_metadata.go:157 lxc/config_template.go:239
-#: lxc/config_trust.go:314 lxc/image.go:493 lxc/network.go:773
-#: lxc/network_acl.go:717 lxc/network_forward.go:777
+#: lxc/alias.go:410 lxc/auth.go:309 lxc/auth.go:1306 lxc/auth.go:2101
+#: lxc/cluster.go:866 lxc/cluster_group.go:405 lxc/config.go:323
+#: lxc/config.go:398 lxc/config.go:1381 lxc/config_metadata.go:157
+#: lxc/config_template.go:239 lxc/config_trust.go:314 lxc/image.go:493
+#: lxc/network.go:773 lxc/network_acl.go:717 lxc/network_forward.go:777
 #: lxc/network_load_balancer.go:740 lxc/network_peer.go:699
 #: lxc/network_zone.go:641 lxc/network_zone.go:1334 lxc/placement_group.go:456
 #: lxc/profile.go:628 lxc/project.go:377 lxc/storage.go:361
@@ -5099,7 +5126,7 @@ msgstr ""
 msgid "Remove a network zone record entry"
 msgstr ""
 
-#: lxc/alias.go:213 lxc/alias.go:214
+#: lxc/alias.go:227 lxc/alias.go:228
 msgid "Remove aliases"
 msgstr ""
 
@@ -5175,7 +5202,7 @@ msgstr ""
 msgid "Rename a placement group"
 msgstr ""
 
-#: lxc/alias.go:158 lxc/alias.go:159 lxc/image_alias.go:287
+#: lxc/alias.go:172 lxc/alias.go:173 lxc/image_alias.go:287
 #: lxc/image_alias.go:288
 msgid "Rename aliases"
 msgstr ""
@@ -5723,6 +5750,10 @@ msgstr ""
 msgid "Show OIDC session"
 msgstr ""
 
+#: lxc/alias.go:273 lxc/alias.go:274
+msgid "Show aliases in YAML format"
+msgstr ""
+
 #: lxc/main.go:100
 msgid "Show all debug messages"
 msgstr ""
@@ -6110,7 +6141,7 @@ msgstr ""
 msgid "Symlink target path can only be used for type \"symlink\""
 msgstr ""
 
-#: lxc/alias.go:140
+#: lxc/alias.go:154
 msgid "TARGET"
 msgstr ""
 
@@ -7627,19 +7658,28 @@ msgstr ""
 msgid "info"
 msgstr ""
 
-#: lxc/alias.go:62
+#: lxc/alias.go:76
 msgid ""
 "lxc alias add list \"list -c ns46S\"\n"
 "    Overwrite the \"list\" command to pass -c ns46S."
 msgstr ""
 
-#: lxc/alias.go:216
+#: lxc/alias.go:321
+msgid ""
+"lxc alias edit\n"
+"\tEdit the aliases via interactive terminal.\n"
+"\n"
+"lxc alias edit < aliases.yaml\n"
+"\tEdit the aliases from \"aliases.yaml\"."
+msgstr ""
+
+#: lxc/alias.go:230
 msgid ""
 "lxc alias remove my-list\n"
 "    Remove the \"my-list\" alias."
 msgstr ""
 
-#: lxc/alias.go:161
+#: lxc/alias.go:175
 msgid ""
 "lxc alias rename list my-list\n"
 "    Rename existing alias \"list\" to \"my-list\"."
--- a/po/ber.po
+++ b/po/ber.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: lxd\n"
 "Report-Msgid-Bugs-To: lxd@lists.canonical.com\n"
-"POT-Creation-Date: 2025-12-10 17:19+0000\n"
+"POT-Creation-Date: 2025-12-11 15:59+0300\n"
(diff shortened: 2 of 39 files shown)

### CODE REVIEW

File: lxc/alias.go [hunk 1]
```diff
@@ -245,3 +259,194 @@ func (c *cmdAliasRemove) run(cmd *cobra.Command, args []string) error {
...
+	alias  *cmdAlias
+}
+
+// Command creates a Cobra command to show all aliases in YAML format.
+func (c *cmdAliasShow) command() *cobra.Command {
+	cmd := &cobra.Command{}
+	cmd.Use = usage("show")
+	cmd.Short = i18n.G("Show aliases in YAML format")
+	cmd.Long = cli.FormatSection(i18n.G("Description"), i18n.G(
+		`Show aliases im YAML format`))
```
minaelee:
```suggestion
		`Show aliases in YAML format`))
```

tugbataluy: Thanks Minae, I fixed it :)

minaelee:
I realized there is also a how-to for `lxc alias` that should be updated: https://documentation.ubuntu.com/lxd/latest/howto/lxc_alias/

It's in doc > howto > lxc_alias.md. Let me know if you'd like any guidance. Ty!

tugbataluy:
> I realized there is also a how-to for `lxc alias` that should be updated: https://documentation.ubuntu.com/lxd/latest/howto/lxc_alias/
> 
> It's in doc > howto > lxc_alias.md. Let me know if you'd like any guidance. Ty!

Hi @minaelee,
I added the edit and show commands to the document you mentioned. Thank you.

lxc/alias.go
```diff
@@ -245,3 +258,166 @@ func (c *cmdAliasRemove) run(cmd *cobra.Command, args []string) error {
...
+}
+
+// Command creates a Cobra command to edit aliases either via interactive editor or via pipe to stdin.
+func (c *cmdAliasEdit) command() *cobra.Command {
+	cmd := &cobra.Command{}
+	cmd.Use = usage("edit")
+	cmd.Short = i18n.G("Edit aliases")
+
+	cmd.Example = cli.FormatSection("", i18n.G(`lxc alias edit
+	Update the aliases via interactive terminal.
```
MusicDin: Should we use `Edit` instead of `Update`?

lxc/alias.go
```diff
@@ -245,3 +258,166 @@ func (c *cmdAliasRemove) run(cmd *cobra.Command, args []string) error {
...
+}
+
+// HelpTemplate returns a sample YAML representation of aliases and guidelines for editing.
+func (c *cmdAliasEdit) helpTemplate() string {
+	return i18n.G(
+		`### This is a YAML representation of the aliases.
+### Any line starting with a '#' will be ignored.
+###
+### A sample aliases configuration looks like:
+### list: "list -c ns46S"
```
MusicDin, line 339: Are aliases displayed with `aliases` namespace? If so, maybe we can include the namespace in the example?

```yaml
### aliases:
###   list: "list -c ns46S"
###   ...
```

lxc/alias.go
```diff
@@ -245,3 +258,166 @@ func (c *cmdAliasRemove) run(cmd *cobra.Command, args []string) error {
...
+### my-list: "list -c ns46S"
+### start-all: "start --all"
+###
+### Note that aliases are simple key-value pairs.`)
```
MusicDin: Can we drop `simple`? They are just key-value pairs :)

lxc/alias.go
```diff
@@ -245,3 +258,166 @@ func (c *cmdAliasRemove) run(cmd *cobra.Command, args []string) error {
...
+		}
+
+		err = yaml.Unmarshal(contents, &config)
+		if err != nil {
+			return nil
+		}
+
+		// Update aliases and save config
+		conf.Aliases = newAliases
+		importedCount := len(newAliases)
```
MusicDin: Do we need to verify new aliases? Maybe at least to ensure key/value are not empty?

EDIT:
To clarify, I just want to make sure we do not allow storing something that could cause a parsing error later. Such error would also affect parsing remotes which happens for almost any command.

However, such error should be caught during marshaling.

lxc/alias.go
```diff
@@ -245,3 +258,166 @@ func (c *cmdAliasRemove) run(cmd *cobra.Command, args []string) error {
...
+	}
+
+	for {
+		// Parse the text received from the editor
+		newAliases := make(map[string]string)
+		err = yaml.Unmarshal(content, &newAliases)
+		if err == nil {
+			// Update aliases and save config
+			conf.Aliases = newAliases
+			err = conf.SaveConfig(c.global.confPath)
```
MusicDin: I see you first check if the parsing error is nil. In such case you try to store the config with new aliases. If storing fails, you reopen the terminal.

I'm wondering what is the error if storing a config fails? If we cannot save the config, is there a good reason to offer user to edit configuration again?
Maybe you can just error out in such case:

```go
for {
		newAliases := make(map[string]string)
		err = yaml.Unmarshal(content, &newAliases)
		if err != nil {
		    // Respawn the editor if there was an error
			fmt.Fprintf(os.Stderr, i18n.G("Alias parsing error: %s")+"\n", err)
			if err != nil {
				return err
			}

			if err != nil {
				return err
			}

			continue
		}

	    // Update aliases and save config
	    conf.Aliases = newAliases
	    return conf.SaveConfig(c.global.confPath)
	}
```

lxc/alias.go
tugbataluy: It depends on the command if user uses lxc alias list it only shows key-value pairs. If user uses lxc alias show it looks like:
aliases:
  details: config show
  list: list -c ns4
  list_network: network list
  list_storage: storage list
But  interactive edit basically shows key-value pairs as lxc alias list does.

MusicDin: > But interactive edit basically shows key-value pairs as lxc alias list does.

Perfect, thanks 👍

tomponline:
Moving back to draft after our meeting, ta

lxc/alias.go
tugbataluy: Now both lxc alias list -f yaml's and lxc alias show output's look like:
details: config show
list: list -c ns4
list_network: network list
list_storage: storage list

lxc/alias.go
tugbataluy: When user edits aliases interactively user can see the details of the key-value pairs, but when edits non-interactively I think we should give information to the user (at least that is what I was trying to do)

lxc/alias.go
tomponline: Thanks @tugbataluy !

simondeziel:
> Copilot encountered an error and was unable to review this pull request. You can try again by re-requesting a review.

I wonder if Copilot has issue when there are too many files to review which is probably worsen by updating the translation. Maybe it's worth dropping the translation changes to try if it works better?

Or maybe it's due to GH partial outage that is ongoing.

MusicDin:
> I wonder if Copilot has issue when there are too many files to review which is probably worsen by updating the translation. Maybe it's worth dropping the translation changes to try if it works better?

That's a good point. On my PRs where translations are updated, Copilot also does not work.

lxc/alias.go
```diff
@@ -245,3 +258,154 @@ func (c *cmdAliasRemove) run(cmd *cobra.Command, args []string) error {
...
+		newAliases := make(map[string]string)
+		config := struct {
+			Aliases map[string]string `yaml:"aliases"`
+		}{
+			Aliases: newAliases,
+		}
+
+		err = yaml.Unmarshal(contents, &config)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
tugbataluy: Done

lxc/alias.go
```diff
@@ -245,3 +258,154 @@ func (c *cmdAliasRemove) run(cmd *cobra.Command, args []string) error {
...
+		}
+
```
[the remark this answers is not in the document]
tugbataluy: Done

doc/howto/lxc_alias.md
```diff
@@ -16,4 +16,18 @@ For example, to always ask for confirmation when deleting an instance, create an
     lxc alias add delete "delete -i"
 
 To see all configured aliases, run [`lxc alias list`](lxc_alias_list.md).
+
+To view all aliases in YAML format (useful for exporting or inspection), run [`lxc alias show`](lxc_alias_show.md).
```
minaelee:
````suggestion
To [view all aliases in YAML format](lxc_alias_show.md) (useful for exporting or inspection), run:

```bash
lxc alias show
```
````

File: doc/howto/lxc_alias.md [hunk 10]
```diff
@@ -16,4 +16,18 @@ For example, to always ask for confirmation when deleting an instance, create an
...
+To edit all aliases interactively or via file input, run [`lxc alias edit`](lxc_alias_edit.md). This command opens your system's default text editor and allow you to modify all aliases at once. You can also pipe alias configurations to this command:
+
+    # Export aliases to a file
+    lxc alias show > aliases.yaml
+
+    # Import aliases from a file
+    lxc alias edit < aliases.yaml
+
+    # Import from pipe
+    cat aliases.yaml | lxc alias edit
```
minaelee:
````suggestion
To [edit all aliases](lxc_alias_edit.md) interactively or via file input, run:

```bash
lxc alias edit
```

This command opens your system's default text editor and allows you to modify all aliases at once.

You can also pipe alias configurations to this command. Examples:

```bash
# Export aliases to a file
lxc alias show > aliases.yaml

# Import aliases from a file
lxc alias edit < aliases.yaml

# Import from pipe
cat aliases.yaml | lxc alias edit
```
````

File: doc/howto/lxc_alias.md (hunk 9 above)
Comment (minaelee) [reply doc/howto/lxc_alias.md]: Prefer command syntax in code blocks rather than inline so that the user can see it clearly when scanning. Prefer backticks for code blocks instead of tabs for clarity and the ability to specify a language for syntax highlighting.

File: lxc/alias.go [hunk 11]
```diff
@@ -245,3 +259,170 @@ func (c *cmdAliasRemove) run(cmd *cobra.Command, args []string) error {
...
+		return nil, cobra.ShellCompDirectiveNoFileComp
+	}
+
+	// Convert aliases to YAML and print
+	data, err := yaml.Marshal(&conf.Aliases)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("%s", data)
```
tomponline: We try to avoid using `Printf` et al where possible to avoid reflection.

Here we would use `fmt.Print(string(data))`

tomponline: Should this include a new line too, via `fmt.Println()`?

tomponline (changes requested):
Please can tests be added - suggest in a new test function in the main test suite.

Also please can you rebase. Ta

lxc/alias.go
tugbataluy: OK, thanks.
For fmt.Println() are we talking about the end of the alias list? If that is the case I can use it.

tomponline:
LGTM thanks!

Ive asked @simondeziel to take a look at the tests for consistency. Thanks!

test/suites/alias.sh
```diff
@@ -0,0 +1,158 @@
+test_alias() {
+  ensure_import_testimage
+
+  # Test 1: Add a new alias
+  echo "Test: Add a new alias"
+  lxc alias add my-list "list -c ns46S"
+  lxc alias list --format csv | grep "my-list,list -c ns46S" > /dev/null
```
simondeziel: Since it's a full line match, `-x` is recommended and `-F` is extra precaution to convey that we are searching for a fixed string (!= regex).
```suggestion
  lxc alias list --format csv | grep -xF "my-list,list -c ns46S"
```

Also, we typically do not suppress the output. It does make CI logs more verbose but it also aid navigating them.

Those suggestions apply to many occurrences below that I won't pin point :)

test/suites/alias.sh
```diff
@@ -0,0 +1,158 @@
...
+
+  # Test 2: Try to add duplicate alias (should fail)
+  echo "Test: Prevent adding duplicate alias"
+  ALIAS_ERR="$(! lxc alias add my-list "list -c -s" 2>&1)"
```
simondeziel: Just in case it succeeds unexpectedly. Granted, this one is mostly for consistency as you inspect the output right after.
```suggestion
  ALIAS_ERR="$(! lxc alias add my-list "list -c -s" 2>&1 || echo fail)"
```

test/suites/alias.sh
```diff
@@ -0,0 +1,158 @@
...
+  echo "${ALIAS_ERR}" | grep "Alias my-list already exists" > /dev/null
```
simondeziel: Whenever possible, full matches are even better, especially when done with shell comparison.
```suggestion
  [ "${ALIAS_ERR}" = "Error: Alias my-list already exists" ]
```

test/suites/alias.sh
```diff
@@ -0,0 +1,158 @@
...
+
+  # Test 3: Show alias in YAML format
+  echo "Test: Show alias in YAML format"
+  lxc alias show | grep "my-list: list -c ns46S" > /dev/null
+
+  # Test 4: Rename an alias
+  echo "Test: Rename an alias"
+  lxc alias rename my-list my-list2
+  ! lxc alias list --format csv | grep "my-list," > /dev/null || false
```
simondeziel: It's best to be extra defensive when dealing with "user input". `grep "^my-list,"` would make sure that it only ever matches on the first column and not somewhere random in the alias definition itself.

test/suites/alias.sh
```diff
@@ -0,0 +1,158 @@
+test_alias() {
+  ensure_import_testimage
+
+  # Test 1: Add a new alias
+  echo "Test: Add a new alias"
```
simondeziel: The comment doesn't add much so I'd drop it. It only risks getting out of sync IMHO.

The `echo` however is quite nice to aid navigating CI logs :) Feel free to include the test number in the `echo` if you find it useful.

test/suites/alias.sh
```diff
@@ -0,0 +1,158 @@
...
+  lxc alias show | grep "my-list: list -c ns46S" > /dev/null
+
+  # Test 4: Rename an alias
+  echo "Test: Rename an alias"
+  lxc alias rename my-list my-list2
+  ! lxc alias list --format csv | grep "my-list," > /dev/null || false
+  lxc alias list --format csv | grep "my-list2,list -c ns46S" > /dev/null
+
+  # Test 5: Try to rename to existing alias (should fail)
+  echo "Test: Prevent rename to existing alias"
```
simondeziel: Nice edge case testing :)

test/suites/alias.sh
```diff
@@ -0,0 +1,158 @@
...
+  done
+
+  # Check aliases
+  for alias_name in "${!special_aliases[@]}"; do
+    lxc alias list --format csv | grep "${alias_name},${special_aliases[$alias_name]}" > /dev/null
+  done
+
+  # Test 18: Test bulk operations via edit
+  echo "Test: Bulk update aliases via edit"
+  cat <<EOF | lxc alias edit
```
simondeziel: No need for `cat` for `heredocs`.
```suggestion
  lxc alias edit <<EOF
```

test/suites/alias.sh
```diff
@@ -0,0 +1,158 @@
...
+bulk-alias3: "list --format yaml"
+bulk-alias4: "list --format table"
+EOF
+
+  # Check bulk aliases
+  bulk_formats=("csv" "json" "yaml" "table")
+  for i in {1..4}; do
+    format_index=$((i-1))
+    lxc alias list --format csv | grep "bulk-alias${i},list --format ${bulk_formats[$format_index]}" > /dev/null
+  done
```
simondeziel, line 114: The hardcoded `{1..4}` part would be best avoided. Since you are already using an array, I'd suggest iterating over the array and compute the index. Gemini proposed this (which I only mildly tweaked):

```
bulk_formats=("csv" "json" "yaml" "table")
for format_index in "${!bulk_formats[@]}"; do
  alias_number=$((format_index + 1))
  format_string=${bulk_formats[format_index]}
  lxc alias list --format csv | grep -xF "bulk-alias${alias_number},list --format ${format_string}"
done
```

test/suites/alias.sh
```diff
@@ -0,0 +1,158 @@
...
+  # Verify that Round-trip aliases still exist and being correct
+  for alias_name in "${!roundtrip_aliases[@]}"; do
+    # check with JSON format
+    lxc alias list --format json | jq -e ".[\"${alias_name}\"]" > /dev/null
+
+    # Check with CSV format
+    lxc alias list --format csv | grep "${alias_name},${roundtrip_aliases[$alias_name]}" > /dev/null
+  done
+
+}
```
simondeziel: Your editor (VSCode does that by default IIRC) is not adding the last `\n` that would avoid this warning.

test/suites/alias.sh
```diff
@@ -0,0 +1,158 @@
...
+  # show alias and pipe to edit
+  lxc alias show | lxc alias edit
+
+  # Verify that Round-trip aliases still exist and being correct
+  for alias_name in "${!roundtrip_aliases[@]}"; do
+    # check with JSON format
+    lxc alias list --format json | jq -e ".[\"${alias_name}\"]" > /dev/null
+
+    # Check with CSV format
+    lxc alias list --format csv | grep "${alias_name},${roundtrip_aliases[$alias_name]}" > /dev/null
```
simondeziel: It's a minor thing but when the output of a given command is relatively short (few 10s of lines), I think it's best to save it in a variable outside of the loop so that it is queried only once.

```
ALIAS_LIST_JSON="$(lxc alias list --format json)"
...
for ...
   jq --exit-status '.["my-list"]' <<< "${ALIAS_LIST_JSON}"
...
```

Also, please use long arg names with `jq`, we are in the process of changing all other occurrences in the test suite :)

test/suites/alias.sh
tugbataluy: In appropriate situations, I used the -xF flag; in situations where I did not perform full line control, I used the -F flag.

test/suites/alias.sh
tugbataluy: I added the numbers to the echo lines

test/suites/alias.sh
```diff
@@ -0,0 +1,143 @@
...
+  echo "Test 2: Prevent adding duplicate alias (should fail)"
+  ALIAS_ERR="$(! lxc alias add my-list "list -c -s" 2>&1 || echo fail)"
+  [ "$(echo "${ALIAS_ERR}" | tail -1)" = "Error: Alias my-list already exists" ]
+
+  echo "Test 3: Show alias in YAML format"
+  lxc alias show | grep -xF "my-list: list -c ns46S"
+
+  echo "Test 4: Rename an alias"
+  lxc alias rename my-list my-list2
+  ! lxc alias list --format csv | grep -xF "my-list," || false
```
simondeziel: `-x` should be dropped for this partial match.

test/suites/alias.sh
```diff
@@ -0,0 +1,143 @@
...
+  lxc alias list --format csv | grep -xF "my-list2,list -c ns46S"
+
+  echo "Test 5: Prevent rename to existing alias (should fail)"
+  lxc alias add another-alias "list -c s"
+  ALIAS_ERR="$(! lxc alias rename another-alias my-list2 2>&1 || echo fail)"
+  [ "$(echo "${ALIAS_ERR}" | tail -1)" = "Error: Alias my-list2 already exists" ]
+
+  echo "Test 6: Remove an alias"
+  lxc alias remove my-list2
+  ! lxc alias list --format csv | grep -xF "^my-list2," || false
```
simondeziel: Same re `-x`

test/suites/alias.sh
```diff
@@ -0,0 +1,143 @@
...
+
+  echo "Test 4: Rename an alias"
+  lxc alias rename my-list my-list2
+  ! lxc alias list --format csv | grep -xF "my-list," || false
+  lxc alias list --format csv | grep -xF "my-list2,list -c ns46S"
+
+  echo "Test 5: Prevent rename to existing alias (should fail)"
+  lxc alias add another-alias "list -c s"
+  ALIAS_ERR="$(! lxc alias rename another-alias my-list2 2>&1 || echo fail)"
+  [ "$(echo "${ALIAS_ERR}" | tail -1)" = "Error: Alias my-list2 already exists" ]
```
simondeziel, line 25: Oh I didn't know there was a header/line to drop. LGTM as-is but the `grep` also made sense because of that, sorry.

test/suites/alias.sh
```diff
@@ -0,0 +1,143 @@
...
+  lxc alias list --format csv | grep -xF "complex,list --format csv --all-projects"
+
+  echo "Test 9: Alias list formats"
+  formats=("json" "yaml" "compact" "table")
+  for format in "${formats[@]}"; do
+    lxc alias list --format "${format}" | grep -F "another-alias"
+  done
+
+  echo "Test 10: Edit aliases via stdin (non-interactive)"
+  cat <<EOF | lxc alias edit
```
simondeziel: No need for `cat` here.

test/suites/alias.sh
```diff
@@ -0,0 +1,143 @@
...
+  lxc alias list --format csv | grep -xF "new-alias1,list --format csv"
+  lxc alias list --format csv | grep -xF "new-alias2,list -c ns"
+
+  echo "Test 11: Prevent clearing all aliases with empty input"
+  ALIAS_ERR="$(! echo "" | lxc alias edit 2>&1 || echo fail)"
+  [ "$(echo "${ALIAS_ERR}" | tail -1)" = "Error: no aliases found in input" ]
+
+  echo "Test 12: Handle invalid YAML in edit (non-interactive)"
+  # Store current aliases count
+  alias_count_before=$(lxc alias list --format json | jq length)
```
simondeziel, line 59:
```suggestion
  alias_count_before="$(lxc alias list --format json | jq --exit-status length)"
```

This ensures that `jq` will error our if it receives malformed JSON or cannot compute the length.

New tests should always use `jq --exit-status`. Eventually, there will be a linter script to enforce this.

test/suites/alias.sh
```diff
@@ -0,0 +1,143 @@
...
+  ALIAS_ERR="$(! echo "invalid: yaml: [}" | lxc alias edit 2>&1 || echo fail)"
+  echo "${ALIAS_ERR}" | grep -F "yaml:"
+
+  # Verify aliases were not changed
+  alias_count_after=$(lxc alias list --format json | jq length)
+  [ "${alias_count_before}" = "${alias_count_after}" ]
```
simondeziel:
```suggestion
  lxc alias list --format json | jq --exit-status "length == \"${alias_count_before}\""
```

test/suites/alias.sh
```diff
@@ -0,0 +1,143 @@
...
+
+  echo "Test 13: Verify alias functionality"
+  lxc alias add short-list "list --format csv"
+  lxc init testimage test-alias-instance -d "${SMALL_ROOT_DISK}"
```
simondeziel: When we can get away with an `--empty` instance, it's always better :)
```suggestion
  lxc init --empty test-alias-instance -d "${SMALL_ROOT_DISK}"
```

test/suites/alias.sh
```diff
@@ -0,0 +1,143 @@
+test_alias() {
+  ensure_import_testimage
```
simondeziel: This shouldn't be needed with the `--empty` suggestion.

test/suites/alias.sh
```diff
@@ -0,0 +1,143 @@
...
+  lxc alias add short-list "list --format csv"
+  lxc init testimage test-alias-instance -d "${SMALL_ROOT_DISK}"
+  lxc short-list | grep -F "test-alias-instance"
+
+  echo "Test 14: ls alias for list command"
+  lxc alias ls --format csv | grep -F "new-alias1"
+
+  echo "Test 15: rm alias for remove command"
+  lxc alias rm new-alias1
+  ! lxc alias list --format csv | grep -F "^new-alias1," || false
```
simondeziel: `^` is a regex anchor so it's not compatible with `-F`.

test/suites/alias.sh
```diff
@@ -0,0 +1,143 @@
...
+
+  echo "Test 16: mv alias for rename command"
+  lxc alias mv new-alias2 renamed-alias
+  ! lxc alias list --format csv | grep -F "^new-alias2," || false
```
simondeziel: Same re `^`

test/suites/alias.sh
```diff
@@ -0,0 +1,143 @@
...
+  # Get alias lists once outside the loop
+  ALIAS_LIST_JSON="$(lxc alias list --format json)"
+  ALIAS_LIST_CSV="$(lxc alias list --format csv)"
+
+  # Verify that Round-trip aliases still exist and being correct
+  for alias_name in "${!roundtrip_aliases[@]}"; do
+    # check with JSON format
+    jq --exit-status '.["'"${alias_name}"'"]' <<< "${ALIAS_LIST_JSON}"
+    # Check with CSV format
+    grep -xF "${alias_name},${roundtrip_aliases[$alias_name]}" <<< "${ALIAS_LIST_CSV}"
```
simondeziel, line 142: Since we are using `jq`, might as well also ensure the alias definitions match.
```suggestion
  for alias_name in "${!roundtrip_aliases[@]}"; do
    alias_value="${roundtrip_aliases[$alias_name]}"
    # check with JSON format
    jq --exit-status '.["'"${alias_name}"'"] == "'"${alias_value}"'"' <<< "${ALIAS_LIST_JSON}"
    # Check with CSV format
    grep -xF "${alias_name},${alias_value}" <<< "${ALIAS_LIST_CSV}"
```

test/suites/alias.sh
```diff
@@ -0,0 +1,143 @@
...
+  done
+
```
simondeziel, line 144: I believe that aliases will be "left behind" after running this script. I suggest either cleaning them all or maybe it'd be simpler to backup the original `config.yml` and restore it at the end.

test/suites/alias.sh
simondeziel: In fact, I'd avoid creating an instance altogether. Something like that maybe:

```
lxc alias add running 'list -f csv STATUS=RUNNING'
[ "$(lxc running || echo fail)" = "" ]
```

Instances, however small involve quite a few IOs which are best avoided, especially on slow backends like `ceph`.

The other advantage is that you then don't need to `lxc delete test-alias-instance` ;)

test/main.sh
```diff
@@ -658,6 +658,7 @@ if [ "${1:-"all"}" != "snap" ] && [ "${1:-"all"}" != "cluster" ]; then
     run_test test_waitready "waitready"
     run_test test_vm_empty "Empty VM"
     run_test test_vm_pcie_bus "VM PCIe bus numbers"
+    run_test test_alias
```
simondeziel: Could you please make that one of the first tests? This should help surface any missing cleanup bits.

tomponline:
@tugbataluy sorry please can you rebase again (I merged another PR as I saw the tests on this were still running and now it conflicts with the translations). Thanks!

lxc/alias.go
```diff
@@ -245,3 +259,170 @@ func (c *cmdAliasRemove) run(cmd *cobra.Command, args []string) error {
...
+
+	cmd.ValidArgsFunction = func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		if len(args) == 0 {
+			return nil, cobra.ShellCompDirectiveDefault
+		}
+
+		return nil, cobra.ShellCompDirectiveNoFileComp
+	}
+
+	// if stdin isn't a terminal, read text from it
```
MusicDin: Not sure if we are still enforcing that, but we tend to start the comments uppercase and end them with a dot.

test/suites/alias.sh
```diff
@@ -0,0 +1,185 @@
+test_alias() {
+  # Get list of original aliases before test
+  local original_alias_list
+  original_alias_list=$(lxc alias list --format json | jq --raw-ouput 'keys[]' || echo "")
+
+  echo "Test 1: Add a new alias"
+  lxc alias add my-list "list -c ns46S"
+  lxc alias list --format csv | grep  -xF "my-list,list -c ns46S"
```
MusicDin: There are two spaces between `grep` and `-xF`

test/main.sh
```diff
@@ -497,6 +497,7 @@ fi
 
 if [ "${1:-"all"}" != "snap" ] && [ "${1:-"all"}" != "cluster" ]; then
     #run_test test_concurrent "concurrent startup" # Disabled as flaky.
+    run_test test_alias
```
MusicDin: Should there be the name?
Like `run_test test_alias "import/export aliases"`?

lxc/alias.go
```diff
@@ -245,3 +259,170 @@ func (c *cmdAliasRemove) run(cmd *cobra.Command, args []string) error {
...
+		err = yaml.Unmarshal(contents, &newAliases)
+		if err != nil {
+			return err
+		}
+
+		importedCount := len(newAliases)
+		// Prevent clearing all aliases if input is empty
+		if importedCount == 0 {
+			fmt.Fprintln(os.Stderr, i18n.G("No aliases found in input. Aborting to prevent clearing all aliases."))
+			return errors.New("no aliases found in input")
```
MusicDin: Error capitalized

MusicDin: Also, why is it writing to stderr and returning an error? Will this be shown twice?

PR #16899, opened 2025-11-07
Bump Go minimum to 1.25.4

simondeziel:


simondeziel:
> Looks like openfga introduced a breaking change to the storage interface that we implement: https://github.com/openfga/openfga/blob/main/CHANGELOG.md#changed-3. `opencontainers/runtime-tools` looks like an indirect dependency

I backed down on the `update-gomod` bits for now. I don't want to delay Go 1.25.4 because of that as it comes with several fixes that should reach our users sooner rather than later.

kadinsayani (approved):

Makefile
tomponline, line 1: This should be squashed into earlier commit

simondeziel: Indeed, fixed.

PR #16921, opened 2025-11-10
storage: Async volume creation

MusicDin:
Make `POST /storage-volumes/<pool>/volumes/<type>` asynchronous. The remaining endpoints will land after this one (to keep PRs small*er*). 

This PR sets the scaffolding for operations in DevLXD/LXD Agent. It also introduces the `noopOperation` which is returned when newer LXD Go client interacts with the older LXD server. This allows the caller to call operation functions without the need to check for the appropriate API extension.

Requires rebase after:
- https://github.com/canonical/lxd/pull/16882

### DIFF
--- a/doc/rest-api.yaml
+++ b/doc/rest-api.yaml
@@ -16646,8 +16646,6 @@ paths:
             produces:
                 - application/json
             responses:
-                "200":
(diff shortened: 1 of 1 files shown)


tomponline (approved):
Thanks!

PR #16930, opened 2025-11-11
Cluster: Make state change error message generic

roosterfish:
Found whilst analyzing the log on the following job https://github.com/canonical/microcloud/actions/runs/19261844108/job/55068474262?pr=1080.

The e2e script tries to restore the cluster member but it hasn't yet stabilized itself (status unavailable) after reboot.

### DIFF
--- a/lxc/cluster.go
+++ b/lxc/cluster.go
@@ -1402,7 +1402,7 @@ func (c *cmdClusterEvacuateAction) run(cmd *cobra.Command, args []string) error 
 		var statusErr api.StatusError
 
 		if errors.As(err, &statusErr) && statusErr.Status() == http.StatusServiceUnavailable {
-			return errors.New("Offline cluster members cannot be evacuated")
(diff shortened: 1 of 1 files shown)


roosterfish:
@tomponline I wonder if we want to grow support for something like `lxd waitready --clustering` to be able to wait until the node's status isn't anymore unavailable?

tomponline:
> @tomponline I wonder if we want to grow support for something like `lxd waitready --clustering` to be able to wait until the node's status isn't anymore unavailable?

you mean until the leader has done another heartbeat round?

roosterfish:
> > @tomponline I wonder if we want to grow support for something like `lxd waitready --clustering` to be able to wait until the node's status isn't anymore unavailable?
> 
> you mean until the leader has done another heartbeat round?

Haven't looked closely yet into how to achieve this but yes something like this. My guess would be there are other operations which would also return an error if the node isn't yet ready, or do we just wait until it becomes ready (guarded by a timeout)?

PR #16933, opened 2025-11-11
VM: De-dupe `pcie` device logic

tomponline:
The boot-time and hot-plug code paths for adding `pci` devices to a VM were separate and different, this PR unifies them using a QMP monitor hook function.

Also introduces the ability for a bus allocator to return a revert.Hook cleanup function should the allocation need to be reverted due to a subsequent failure. This lays the ground work for future modification to revert the allocation's volatile config key if needed.

### DIFF
--- a/lxd/instance/drivers/driver_qemu.go
+++ b/lxd/instance/drivers/driver_qemu.go
@@ -2378,6 +2378,13 @@ func (d *qemu) deviceAttachPath(deviceName string) (mountTag string, err error) 
 		return "", fmt.Errorf("Failed to open device socket file %q: %w", virtiofsdSockPath, err)
 	}
 
+	defer func() {
+		err := socketFile.Close()
+		if err != nil {
+			d.logger.Warn("Failed closing device socket file", logger.Ctx{"device": deviceName, "file": virtiofsdSockPath, "err": err})
+		}
+	}()
+
 	shortPath := fmt.Sprintf("/dev/fd/%d", socketFile.Fd())
 
 	addr, err := net.ResolveUnixAddr("unix", shortPath)


lxd/instance/drivers/driver_qemu.go
```diff
@@ -4834,35 +4848,73 @@ func (d *qemu) writeNICDevConfig(mtuStr string, devName string, nicName string,
...
+			pciIOMMUGroup = pciItem.Value
+		default:
+			return nil, errors.New("Unexpected PCI configuration key: " + pciItem.Key)
 		}
 	}
 
-	devBus, devAddr, multi, err := busAllocate(devName, false)
+	if !slices.Contains([]string{"pcie", "pci"}, busName) {
+		return nil, errors.New("Attempting PCI passthrough on a non-PCI system")
+	}
```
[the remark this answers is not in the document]
tomponline: cant add a pci device on a system that doesnt support pci

lxd/instance/drivers/driver_qemu.go
```diff
@@ -4834,35 +4848,73 @@ func (d *qemu) writeNICDevConfig(mtuStr string, devName string, nicName string,
...
-			busName:       busName,
-			devBus:        devBus,
-			devAddr:       devAddr,
-			multifunction: multi,
-		},
-		devName:     devName,
-		pciSlotName: pciSlotName,
+	escapedDeviceName := filesystem.PathNameEncode(devName)
+	qemuDev := map[string]any{
+		"driver":        "vfio-pci",
```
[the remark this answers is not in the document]
tomponline: cant add a pci device on a system that doesnt have pci

lxd/instance/drivers/driver_qemu.go
```diff
@@ -4782,35 +4792,43 @@ func (d *qemu) addNetDevConfig(busName string, busAllocate busAllocator, bootInd
...
-				return nil, errors.New("No PCI IOMMU group supplied")
+			if busCleanup != nil {
+				reverter.Add(busCleanup)
 			}
 
-			vfioGroupFile := "/dev/vfio/" + pciIOMMUGroup
-			err := os.Chown(vfioGroupFile, int(d.state.OS.UnprivUID), -1)
-			if err != nil {
-				return nil, fmt.Errorf("Failed to chown vfio group device %q: %w", vfioGroupFile, err)
+			// Detect physical passthrough device.
```
roosterfish, line 4771: Is there an implication/benefit by moving this code into the `monitorHook` callback returned by `addNetDevConfig` instead of running it directly? Or is it just for consistency compared to the other `else if` block above which also only sets the hook instead of running something directly.

roosterfish: I assume due to consistency as the follow up commit message says so.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -2441,7 +2441,9 @@ func (d *qemu) deviceAttachPath(deviceName string) (mountTag string, err error)
 		return "", err
 	}
 
-	reverter.Add(busCleanup)
+	if busCleanup != nil {
```
roosterfish, line 2444: Worth squashing this with commit https://github.com/canonical/lxd/pull/16933/commits/5395873fe0469fc1fe21974b1673336676a0001b when adding the handling of the `busCleanup`?

roosterfish (changes requested):
Looks good to me, only a nit and question for clarity.

lxd/instance/drivers/driver_qemu.go
tomponline: Yes it was for consistency and also to avoid this function being able to return an error after `busCleanup, devBus, devAddr, multi, err = busAllocate(devName, false)` has run without calling `busCleanup`.

Let me re-look at this in the cold light of day and see if I can do better here.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -2256,19 +2256,22 @@ func (d *qemu) getPCISlotCount() (pciSlots uint8, err error) {
...
 
 // busAllocatePCIeHotplug provides a busAllocator implementation for hotplugging PCIe devices.
-func (d *qemu) busAllocatePCIeHotplug(deviceName string, _ bool) (busName string, busAddress string, multifunction bool, err error) {
+func (d *qemu) busAllocatePCIeHotplug(deviceName string, _ bool) (cleanup revert.Hook, busName string, busAddress string, multifunction bool, err error) {
 	if d.localConfig["volatile.bus.mode"] != qemuBusModePersistent {
 		return d.busAllocatePCIeHotplugLegacy(deviceName, false)
 	}
 
+	reverter := revert.New()
+	defer reverter.Fail()
```
MusicDin, line 2265: The reverter is defined, but no revert functionallity is added to it. Am I missing something?

MusicDin: Is this to avoid returning `nil`?

tomponline: > This lays the ground work for future modification to revert the allocation's volatile config key if needed.

lxd/instance/drivers/driver_qemu.go
tomponline: done

lxd/instance/drivers/driver_qemu.go
tomponline: @roosterfish ill try and refactor this in a follow up PR if thats OK.

For now its consistent and correct, but it could be made somewhat more efficient by doing more of the work earlier and not repeating it for each macvlan device queue.

roosterfish: All right thanks for checking.

roosterfish (approved):

lxd/instance/drivers/driver_qemu.go
tomponline: @roosterfish addressed in https://github.com/canonical/lxd/pull/16941

PR #16944, opened 2025-11-13
lxd/response/capture: Report API response error code instead of response code

MusicDin:
In some cases, response code is not set, so when response is rendered it is set to `200` (Status OK).
When we capture the response in DevLXD and convert it to API response, we find that the response type is an `Error`, and return the API error response. However, the error code was set to http status code instead of API response error code. When the source response had no status code set, the error was returned as success (note that DevLXD has no response "envelope").

### DIFF
--- a/lxd/response/response_capture.go
+++ b/lxd/response/response_capture.go
@@ -64,7 +64,7 @@ func (rc *responseCapture) ToAPIResponse() (*api.Response, string, error) {
 
 	// Handle errors.
 	if response.Type == api.ErrorResponse {
-		return nil, "", api.NewStatusError(rc.statusCode, response.Error)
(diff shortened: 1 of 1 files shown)


tomponline:
@MusicDin sounds like we have gap in tests too

MusicDin:
> @MusicDin sounds like we have gap in tests too

I caught this one in "async volume update" where we check for status locked when block volume is in use. Now, we will have at least 1 test.

This was OK prior operations, but now the operation is successfully returned (200). If it contains error response (operation content), DevLXD has no envelope, so client needs to get response code from the actual API response. Otherwise, the envelope is trimmed, and error returned as success.

tomponline (approved):

PR #16965, opened 2025-11-17
Storage: Ignore if there aren't yet any pools during source validation

roosterfish:
When trying to get all created storage pool names, ignore in case no pool exists.
This currently affects storage pool creation in clusters and returns the error `Error: Failed loading config for storage pools on "<pool>": Storage pool(s) not found`.

The reason this isn't caught in the pipeline is because whenever we create a pool inside a cluster, there already is a pool created. E.g. the `data` pool in the `clustering_evacuation` test suite.

Also the weekly MicroCloud pipeline shows every test run using `latest/edge` failing with the same error https://github.com/canonical/microcloud/actions/runs/19397642748.
We didn't catch this on the regular pipeline as it runs with LXD LTS `5.21/edge`.

### DIFF
--- a/test/suites/clustering.sh
+++ b/test/suites/clustering.sh
@@ -657,6 +657,9 @@ test_clustering_storage() {
   printf 'config: {}\ndevices: {}' | LXD_DIR="${LXD_ONE_DIR}" lxc profile edit default
   LXD_DIR="${LXD_TWO_DIR}" lxc storage delete data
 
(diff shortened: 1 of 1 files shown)


MusicDin (approved):
Thanks!

roosterfish:
> Please can you add a test to catch this regression.

Yep looking into it rn.

PR #16967, opened 2025-11-17
shared: Add unit tests for DetectCompressionFile()

kadinsayani:
Additional unit tests to improve coverage.

### DIFF
--- a/shared/archive_test.go
+++ b/shared/archive_test.go
@@ -1,0 +1,201 @@
+package shared
+
+import (
+	"bytes"
+	"errors"
+	"io"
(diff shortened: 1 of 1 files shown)


kadinsayani:
> The 2 test functions look pretty similar and I don't see what's the additional value in creating real temporary test files. Maybe it helps the coverage?
> 
> I think the test duplication could be reduced if the `file_not_found` test was moved into the first test function.

I've consolidated the test logic into a single function. We can just test `DetectCompressionFile()` and skip creating temp files :)

Issue #16962, opened 2025-11-17
docs: Source installation instructions don't work with `sudo-rs`


astrojuanlu:
https://github.com/canonical/lxd/blob/5aeb56ff68b2c94bbce7a76904adc8a9fe21e09c/doc/installing.md?plain=1#L375-L380

But:

```
$ sudo -E PATH=${PATH} LD_LIBRARY_PATH=${LD_LIBRARY_PATH} $(go env GOPATH)/bin/lxd --group sudo
warning: preserving the entire environment is not supported, `-E` is ignored
```

because: https://github.com/trifectatechfoundation/sudo-rs/issues/1299

And apparently that issue was marked as `wontfix`, so it's not clear to me how to proceed, other than reverting to `sudo-ws`.

---
*Reported from: https://documentation.ubuntu.com/lxd/latest/installing/*

tomponline:
@simondeziel can we update with `--preserve-env=VAR` instead?

PR #16969, opened 2025-11-17
doc/installing: manually run the `lxd` daemon with the `lxd` group

simondeziel:
Reusing the `sudo` group is not standard and might have been convenient but since the `lxd` group is created a few steps earlier, there is no point in abusing the `sudo` group.

Also remove the `-E` argument for `sudo` as needed environment variables are already passed to the `lxd` daemon already and `-E` is not implemented by `sudo-rs`. Fixes https://github.com/canonical/lxd/issues/16962


tomponline (approved):
ta

PR #16972, opened 2025-11-17
Switch to 24.04 runners

simondeziel:


simondeziel:
Aleks' commit went through a lot of repeated runs in his own fork and in mine so we are relatively confident that it is reliable.

The `migration` test takes a performance hit as expected going from ~600s to ~950s.

PR #16974, opened 2025-11-17
github: Output total percent coverage for observability

kadinsayani:
It will be useful to see our total code coverage percent using Go tooling so we have an accurate baseline to compare our TIOBE Tics results with.


kadinsayani:
> Thanks.
> 
> @kadinsayani btw have you had chance to check if the coverage files generated are potentially overwriting each other. I cant recall the method the file names are generated with. Could we be getting collisions?

Followed up on MM.

PR #16977, opened 2025-11-18
Doc: Drop Pure from replication disaster recovery doc

roosterfish:
Regular volume recovery works for Pure using 'lxd recover'. 

Due to the availability of hardware, we haven't yet tested replication between two Pure clusters in order to replay the disaster recovery process. Therefore we drop the link to the docs that might suggest it's supported until we have been able to perform tests.


tomponline (approved):

roosterfish:
@tomponline docs build looks to be ok now after dropping the exepctions for the list style and fixing it.

tomponline (approved):

PR #16979, opened 2025-11-18
lxc/cookiejar: Add freebsd build for filelock

MusicDin:
We are building Terraform provider for freebsd, and since `filelock.go` is built only for linux/darwin, it results in undefined function definitions. 

I used `!windows` to be consistent with `lxc/console_unix.go` / `lxc/exec_unix.go`.
Perhaps, we could just go with `//go:build linux || darwin || freebsd`?

### DIFF
--- a/lxc/cookiejar/filelock_unix.go
+++ b/lxc/cookiejar/filelock_unix.go
@@ -1,4 +1,4 @@
-//go:build linux || darwin
+//go:build !windows
 
 package cookiejar
 


tomponline:
@MusicDin how are you building for freebsd, are you using GH runners or cross compiling? Maybe we should add a compilation target for freebsd like we are doing for windows and macos.

MusicDin:
Just cross compiling. And can confirm that it is working now 👍

PR #16981, opened 2025-11-18
Restore consistent coverage by reverting gocover-cobertura and updating to go1.25 compatible deps

kadinsayani:
This pull request reverts the recent switch from `gocov/gocov-xml` to `gocover-cobertura` for converting coverage to Cobertura XML format. `gocover-cobertura` requires multiple manual workarounds and has ultimately left us in a worse state with inconsistent coverage. Additionally, this pull request updates the dependencies to versions that support go1.25. For `gocov-xml`, we're using a fork that supports go1.25 as a temporary workaround until https://github.com/AlekSi/gocov-xml/pull/20 is merged.

My aim is to get back to consistent code coverage reporting, analyze results while considering the addition of snap tests to see where we're at, and we can go from there.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -825,8 +825,8 @@ jobs:
       - name: Install dependencies
         run: |
           set -eux
-          go install github.com/axw/gocov/gocov@latest
-          go install github.com/AlekSi/gocov-xml@latest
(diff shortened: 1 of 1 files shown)


tomponline:
@kadinsayani will this run with go1.25?

kadinsayani:
> @kadinsayani will this run with go1.25?

Yes.

PR #16985, opened 2025-11-18
tests: Add test for clusterPutDisable() and ReplaceDaemon()

kadinsayani:
This adds a test function covering the following functions: https://github.com/canonical/lxd/blob/7c2b10982d625d580ced4c048ea3c140ab2a94ff/lxd/api_cluster.go#L911 https://github.com/canonical/lxd/blob/7c2b10982d625d580ced4c048ea3c140ab2a94ff/lxd/util/sys.go#L58

### DIFF
--- a/test/main.sh
+++ b/test/main.sh
@@ -462,6 +462,7 @@ if [ "${1:-"all"}" != "snap" ] && [ "${1:-"all"}" != "standalone" ]; then
     run_test test_clustering_ha "clustering high availability"
     run_test test_clustering_handover "clustering handover"
     run_test test_clustering_rebalance "clustering rebalance"
+    run_test test_clustering_rebalance_remove_leader "clustering rebalance remove leader"
     run_test test_clustering_remove_raft_node "clustering remove raft node"
     run_test test_clustering_failure_domains "clustering failure domains"
     run_test test_clustering_image_refresh "clustering image refresh"


test/suites/clustering.sh
```diff
@@ -3133,6 +3133,68 @@ test_clustering_rebalance() {
...
+  # Wait for cluster to stabilize and role rebalancing to complete
+  sleep 5
+
+  echo "Verify we have two database nodes"
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node1 | grep -F "database: true"
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node2 | grep -F "database: true"
+
+  echo "Remove the leader node from the cluster"
+  # When a leader removes itself, clusterPutDisable() is called, which in turn calls ReplaceDaemon().
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster remove node1 --force --yes
```
simondeziel, line 3178: Is it 100% guarantied that `node1` will be the DB leader? If not, maybe loop over the `lxc cluster list -f csv | read -r ...` to nuke the leader?

test/suites/clustering.sh
```diff
@@ -3133,6 +3133,68 @@ test_clustering_rebalance() {
...
+  LXD_TWO_DIR=$(mktemp -d -p "${TEST_DIR}" XXX)
+  ns2="${prefix}2"
+  spawn_lxd_and_join_cluster "${ns2}" "${bridge}" "${cert}" 2 1 "${LXD_TWO_DIR}" "${LXD_ONE_DIR}"
+
+  echo "Verify clustering enabled on both nodes"
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster list | grep -F node1
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster list | grep -F node2
+
+  # Wait for cluster to stabilize and role rebalancing to complete
+  sleep 5
```
simondeziel: Could we loop every `0.5s` waiting for the desired roles to be assigned?

test/suites/clustering.sh
kadinsayani: Since we bootstrap the cluster on `node1`, it should always be the leader. I haven't observed different behavior in repeated runs.

simondeziel: Sounds good, thanks

test/suites/clustering.sh
kadinsayani: Yeah good idea, thanks!

test/suites/clustering.sh
```diff
@@ -3133,6 +3133,75 @@ test_clustering_rebalance() {
...
+  echo "Wait for node1 daemon to restart after removal"
+  # The daemon restarts itself after cluster removal via ReplaceDaemon()
+  LXD_DIR="${LXD_ONE_DIR}" lxd waitready --timeout=30
+
+  echo "Verify node1 is still functional with clustering disabled"
+  LXD_DIR="${LXD_ONE_DIR}" lxc info | grep -F "server_clustered: false"
+
+  echo "Verify node2 is still clustered and is now the only member"
+  LXD_DIR="${LXD_TWO_DIR}" lxc cluster list # For debugging
+  [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc cluster list -f csv | wc -l)" = "1" ]
```
simondeziel, line 3189: This is the only call with `-f csv` so should help with coverage ;)

test/suites/clustering.sh
```diff
@@ -3133,6 +3133,75 @@ test_clustering_rebalance() {
...
+
+  echo "Verify clustering enabled on both nodes"
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster list | grep -F node1
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster list | grep -F node2
+
+  # Wait for cluster to stabilize and role rebalancing to complete
+  echo "Waiting for both nodes to have database role..."
+  for _ in $(seq 10); do
+    if LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node1 | grep -F "database: true" && \
+       LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node2 | grep -F "database: true"; then
```
simondeziel, line 3166: Not sure if that would work? Or maybe it's too cryptic to be worth it?
```
[ "$(LXD_DIR="${LXD_TWO_DIR}" lxc cluster list -f csv | grep -cwE 'database-(leader|standby)')" = "2" ] && break
sleep 0.5
```

kadinsayani: I think that would work but I prefer the current approach for readability :)

PR #16968, opened 2025-11-17
lxd/network/acl: Propagate context to ACL package

markylaing:
First of many pull requests which will allow context propagation throughout the lifetime of a request/operation. This is for #16613 and the wider refactor of our data access patterns.

See https://github.com/canonical/lxd/issues/16613#issuecomment-3547843653

Issue #16613, opened 2025-09-25
Auth: Use entity type and ID for authorization checks


markylaing:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

We initially wrote the OpenFGA datastore to return tuples of the form:
```json
{
   "user": "<entity type>:<entity URL>",
   "relation": "<relation>",
   "object": "<entity type>:<entity URL>"
}
```
e.g.
```json
{
   "user": "group:/1.0/auth/groups/{name}#member",
   "relation": "can_view",
   "object": "instance:/1.0/instances/{name}?project={project_name}"
}
```

This was so that a permission check can be performed against a URL directly. E.g. if checking for access in `/1.0/instances/{name}?project=default` the caller wouldn't need to get the ID of the instance before performing the check.

This made it easier to retrofit fine-grained access control into LXD, but was shortsighted because the OpenFGA datastore spends a lot of time resolving entity types and IDs (from `auth_groups_permissions`) into URLs to return as tuples. This is likely causing fine-grained authorization to be slow.

### Describe the solution you'd like

The OpenFGA datastore should return tuples of the form:
```json
{
   "user": "<entity type>:<entity ID>",
   "relation": "entitlement",
   "object": "<entity type>:<entity ID>"
}
```
e.g. 
```json
{
   "user": "group:1#member",
   "relation": "can_view",
   "object": "instance:10"
}
```

The `auth.Authorizer` interface must be modified to account for the change, and all calls to the authorizer will require getting the entity ID before doing the access check. Note however that this should not slow LXD down, because we always need to load the resource from the database at some point anyway.

markylaing:
Preliminary work for this should change how entity references (used by URLs) are loaded from the database. We need to filter these by the permissions of the caller, so they will need to be of the form `map[entity.Type][]int` or similar.

markylaing:
### `tl;dr`
- No simple solution
- Will add structure/guidelines to our database access patterns in LXD. Then using entity type and ID will follow naturally.

### Investigation

I've investigated a few different approaches and each run into one stumbling block or another. Below I'll try to summarise each of these approaches and why they don't work. This will help to rationalise the much larger refactor I will suggest later.

### Attempt 1: Just start using entity type and ID

This naive approach refactored the Authorizer interface as below with no other changes:
```go
type PermissionChecker func(entityID int) bool

type Authorizer interface {
[...]

PR #16992, opened 2025-11-19
lxd/network/openvswitch: Propagate context

markylaing:
Part of work towards https://github.com/canonical/lxd/issues/16613#issuecomment-3547843653

Depends on #16968


markylaing:
I think because this is not a priority I will close it to avoid conflict with #17089

PR #16997, opened 2025-11-19
cgroup package refactorings

kadinsayani:
This pull request refactors how cgroup resource support is checked and simplifies the representation of cgroup layouts. The main change is to move resource support logic out of a large switch statement and into a generalized function that uses a map of resources to their controller key(s). This reduces cyclomatic complexity and code duplication. Additionally, the cgroup layout type is now a string instead of an integer, improving readability and removing the need for `Mode()` (also reduces cyclomatic complexity).

### DIFF
--- a/lxd/cgroup/init.go
+++ b/lxd/cgroup/init.go
@@ -104,123 +104,14 @@ const (
 	Pids
 )
 
-// SupportsVersion indicates whether or not a given cgroup resource is
-// controllable and in which type of cgroup filesystem.
-func (info *Info) SupportsVersion(resource Resource) (Backend, bool) {
-	switch resource {
-	case Blkio:
-		val, ok := cgControllers["blkio"]
-		if ok {
-			return val, ok
-		}
-
-		val, ok = cgControllers["io"]
-		if ok {
-			return val, ok
-		}
-
-		return Unavailable, false
-	case BlkioWeight:
-		val, ok := cgControllers["blkio.weight"]
-		if ok {
-			return val, ok
-		}
-
-		val, ok = cgControllers["io"]
-		if ok {
-			return val, ok
-		}
-
-		return Unavailable, false
-	case CPU:
-		val, ok := cgControllers["cpu"]
-		return val, ok
-	case CPUAcct:
-		val, ok := cgControllers["cpuacct"]
-		if ok {
-			return val, ok
-		}
-
-		val, ok = cgControllers["cpu"]
-		if ok {
-			return val, ok
-		}
-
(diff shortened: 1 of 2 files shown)


lxd/daemon.go
```diff
@@ -1322,7 +1322,7 @@ func (d *Daemon) init() error {
 
 	dbWarnings = append(dbWarnings, d.os.CGInfo.Warnings()...)
 
-	logger.Infof(" - cgroup layout: %s", d.os.CGInfo.Mode())
+	logger.Infof(" - cgroup layout: %s", d.os.CGInfo.Layout)
```
simondeziel, line 1325: A good candidate for `logger.Info(" - cgroup layout: " + d.os.CGInfo.Layout)`?

simondeziel (commented):
I'm not qualified to review this but it reads nicely :)

lxd/daemon.go
kadinsayani: We'll have to cast `d.os.CGInfo.Layout` to `string`, which probably has similar overhead to just using `fmt.Sprintf()`.

simondeziel: I see, thanks.

lxd/cgroup/init.go
tomponline, line 1: Is it just me or does the new version seem longer and more complex than the current form?

kadinsayani: I've simplified the new version further by adding two separate implementations, one for resources with a single key, and one for resources with a key and fallback key. Now, we have two generalized functions, and a lot less duplication.

kadinsayani: This will improve TICS cyclomatic complexity and code duplication scores for the cgroup package.

lxd/cgroup/resource.go
```diff
@@ -0,0 +1,58 @@
...
+			return val, ok
+		}
+	}
+
+	return Unavailable, false
+}
+
+// resourceMap associates each [Resource] with the [ResourceInterface] implementation that checks its support.
+var resourceMap = map[Resource]ResourceInterface{
+	Blkio:              FallbackResource{Keys: []string{"blkio", "io"}},
```
tomponline: Can we just have FallbackResource (named Resource) that takes a slice of keys?

tomponline: Also do we need the interface at all, can we instead have a map Resource to a slice of keys that the `SupportsVersion` function consumes ?

kadinsayani: Yeah we can do that. Even simpler!

PR #17002, opened 2025-11-20
tests: Use default storage pool in DevLXD tests

MusicDin:
Uses existing storage pool in DevLXD tests to save time creating storage pools.


tomponline:
@MusicDin is this still relevant?

PR #17003, opened 2025-11-20
Test: Fix cleanup in clustering tests

roosterfish:
Currently each clustering test manually kills the cluster members and later also runs `kill_lxd` which should cleanup but cannot anymore as the members are already killed so it skips this logic.
This causes traces left by each test suite (e.g. the `data` pool for clusters) which also causes the underlying storage (e.g. LVM volume groups) to be still around for the next test. 

This PR aims to solely rely on `kill_lxd` to cleanup after every test to benefit from the already existing logic.


roosterfish:
Closing this for now as it seems to be a false alarm.

PR #17007, opened 2025-11-20
Use volume name when getting attached volumes during multi-volume snapshot/restore

kadinsayani:
This pull request addresses two issues:

1. Attaching a volume with a custom device name (affects filesystem and block volumes) and taking a multi-volume instance snapshot; and
2. Attaching a block volume to a VM with a path specified (the path is used as the device name) and taking a multi-volume instance snapshot.

In both scenarios, we want to use the source volume name in the query, not the device name.

lxd-ci tests: https://github.com/canonical/lxd-ci/pull/626.

### DIFF
--- a/test/suites/snapshots.sh
+++ b/test/suites/snapshots.sh
@@ -568,6 +568,14 @@ test_snapshot_multi_volume() {
   lxc delete c1/c1-snap3 --disk-volumes=all-exclusive
   ! lxc storage volume list -f csv | grep -F "Created alongside container c1/c1-snap3 snapshot in project default" || false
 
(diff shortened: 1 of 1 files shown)


kadinsayani:
> How come this only affected block vols?

It affects filesystem volumes too, we just didn't catch it because in the tests the device name matched the volume name. The bug fix addresses two issues:

1) Attaching a volume with a custom device name (affects filesystem and block volumes); and
2) Attaching a block volume to a VM with a path specified (the path is used as the device name).

In both scenarios, we want to use the source volume name in the query, not the device name.

Issue #17005, opened 2025-11-20
Enhance `lxc info` output to show multi-volume snapshot status


mseralessandri:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

Currently, the output of `lxc info <instance>` displays a list of available snapshots for a container or VM, but it does not indicate whether a snapshot is a multi-volume snapshot (i.e., a snapshot that includes storage volumes other than the root disk, see #16336 ).

```
[...]
Snapshots:
+-------+----------------------+------------+----------+
| NAME  |       TAKEN AT       | EXPIRES AT | STATEFUL |
+-------+----------------------+------------+----------+
| snap0 | 2025/11/20 10:46 UTC |            | NO       |
```

The only way to see if a snapshot is multi-volume is to launch `lxc storage volume list <pool>` and filtering for volumes associated with the instance snapshot name (e.g., `snap0`). This table can be huge and requires the user to mentally parse the output to identify volumes tied to a specific instance snapshot, which is not immediately clear or intuitive.

```
+----------------------------+------------------+------------------------------------------------------------------------+--------------+---------+
|            TYPE            |       NAME       |                              DESCRIPTION                               | CONTENT-TYPE | USED BY |
+----------------------------+------------------+------------------------------------------------------------------------+--------------+---------
| custom                     | non-shared       |                                                                        | filesystem   | 1       |
+----------------------------+------------------+------------------------------------------------------------------------+--------------+---------+
| custom (snapshot)          | non-shared/snap0 | Created alongside virtual-machine vm/snap0 snapshot in project default | filesystem   | 0       |
+----------------------------+------------------+------------------------------------------------------------------------+--------------+---------+
| custom                     | shared           |                                                                        | filesystem   | 2       |
+----------------------------+------------------+------------------------------------------------------------------------+--------------+---------+
| custom (snapshot)          | shared/snap0     | Created alongside virtual-machine vm/snap0 snapshot in project default | filesystem   | 0       |
+----------------------------+------------------+------------------------------------------------------------------------+--------------+---------+
| virtual-machine            | vm               |                                                                        | block        | 1       |
+----------------------------+------------------+------------------------------------------------------------------------+--------------+---------+
| virtual-machine (snapshot) | vm/snap0         |                                                                        | block        | 1       |
+----------------------------+------------------+------------------------------------------------------------------------+--------------+---------+
```

### Describe the solution you'd like

[...]

PR #17008, opened 2025-11-20
lxc/info: Add attached volume snapshots to snapshot table

kadinsayani:
Closes https://github.com/canonical/lxd/issues/17005.

Sample output:

```
...
Snapshots:
+----------+----------------------+----------------------+----------+---------------------------+
|   NAME   |       TAKEN AT       |      EXPIRES AT      | STATEFUL | ATTACHED VOLUME SNAPSHOTS |
+----------+----------------------+----------------------+----------+---------------------------+
| c1-snap0 | 2025/11/25 15:55 UTC |                      | NO       |
+----------+----------------------+----------------------+----------+---------------------------+
| c1-snap1 | 2025/11/25 15:55 UTC | 2025/11/25 17:55 UTC | NO       | non-shared/snap0          |
|          |                      |                      |          | shared/snap0              |
+----------+----------------------+----------------------+----------+---------------------------+
| c1-snap2 | 2025/11/25 15:55 UTC | 2025/11/25 17:55 UTC | NO       | shared/snap1              |
|          |                      |                      |          | non-shared/snap1          |
+----------+----------------------+----------------------+----------+---------------------------+
```

### DIFF
--- a/po/ar.po
+++ b/po/ar.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: lxd\n"
 "Report-Msgid-Bugs-To: lxd@lists.canonical.com\n"
-"POT-Creation-Date: 2025-11-17 14:48+0000\n"
+"POT-Creation-Date: 2025-11-25 07:53-0800\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Automatically generated\n"
 "Language-Team: none\n"
@@ -432,7 +432,7 @@ msgid ""
 "### Any line starting with a '# will be ignored."
 msgstr ""
 
-#: lxc/info.go:329
+#: lxc/info.go:330
 #, c-format
 msgid "%d (id: %d, online: %v, NUMA node: %v)"
 msgstr ""
@@ -442,7 +442,7 @@ msgstr ""
 msgid "%s (%d more)"
 msgstr ""
 
-#: lxc/info.go:169
+#: lxc/info.go:170
 #, c-format
 msgid "%s (%s) (%d available)"
 msgstr ""
@@ -461,17 +461,17 @@ msgstr ""
 msgid "(none)"
 msgstr ""
 
-#: lxc/info.go:319
+#: lxc/info.go:320
 #, c-format
 msgid "- Level %d (type: %s): %s"
 msgstr ""
 
-#: lxc/info.go:297
+#: lxc/info.go:298
 #, c-format
 msgid "- Partition %d"
 msgstr ""
 
-#: lxc/info.go:205
+#: lxc/info.go:206
 #, c-format
 msgid "- Port %d (%s)"
 msgstr ""
@@ -513,7 +513,7 @@ msgid "--refresh can only be used with instances"
 msgstr ""
 
 #: lxc/config.go:211 lxc/config.go:482 lxc/config.go:657 lxc/config.go:864
-#: lxc/info.go:461
+#: lxc/info.go:462
 msgid "--target cannot be used with instances"
 msgstr ""
 
@@ -718,7 +718,7 @@ msgstr ""
 msgid "Add rules to an ACL"
 msgstr ""
 
-#: lxc/info.go:209
+#: lxc/info.go:210
 #, c-format
 msgid "Address: %s"
 msgstr ""
@@ -773,12 +773,12 @@ msgstr ""
 msgid "Alternative certificate name"
 msgstr ""
 
-#: lxc/image.go:1017 lxc/info.go:486
+#: lxc/image.go:1017 lxc/info.go:487
 #, c-format
 msgid "Architecture: %s"
 msgstr ""
 
-#: lxc/info.go:135
+#: lxc/info.go:136
 #, c-format
 msgid "Architecture: %v"
 msgstr ""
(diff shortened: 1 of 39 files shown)


tomponline:
@kadinsayani lets discuss approach before merging this, I'm wondering if it wouldn't be more useful to include "Volumes" column and include a list of them associated to snapshot, rather than just the disk volumes mode as a column.

kadinsayani:
> @kadinsayani lets discuss approach before merging this, I'm wondering if it wouldn't be more useful to include "Volumes" column and include a list of them associated to snapshot, rather than just the disk volumes mode as a column.

Perhaps we can keep the "disk volumes mode" column in the snapshots table and add another section for attached volumes. In that section, we can list the instance's attached volumes. For each volume, we can include a snapshots table. What do you think?

tomponline:
> Perhaps we can keep the "disk volumes mode" column in the snapshots table and add another section for attached volumes

Seems unnecessary to have 2 columns, to my mind showing associated volume snapshots of the instance snapshot is more useful as it indicates the "disk volumes mode" was "all-exclusive", but goes further and shows specifically which disk volumes were included at that time.

kadinsayani:
> > Perhaps we can keep the "disk volumes mode" column in the snapshots table and add another section for attached volumes
> 
> Seems unnecessary to have 2 columns, to my mind showing associated volume snapshots of the instance snapshot is more useful as it indicates the "disk volumes mode" was "all-exclusive", but goes further and shows specifically which disk volumes were included at that time.

Sounds good, let's go with your suggested approach.

tomponline:
> | c1-snap1 | 2025/11/21 16:32 UTC | 2025/11/21 18:32 UTC | NO       | shared/snap0, non-shared/snap0 |

Do you know of other columns where we comma-separate a list inside the field rather than separate with newlines (like IP addresses in `lxc list`)?

lxc/info.go
```diff
@@ -642,6 +643,50 @@ func (c *cmdInfo) instanceInfo(d lxd.InstanceServer, name string, showLog bool)
...
 			}
 
+			// Display attached volume snapshots
+			var attachedVolumeSnapshots string
+			if snap.Config["volatile.attached_volumes"] != "" {
+				// Parse the JSON map of volume UUID -> snapshot UUID
+				var attachedVolumeUUIDs map[string]string
+				err := json.Unmarshal([]byte(snap.Config["volatile.attached_volumes"]), &attachedVolumeUUIDs)
+				if err != nil {
+					return err
```
tomponline: I'd encourage the use of error wrapping here so the user cant get vague unmarshal errors.

lxc/info.go
```diff
@@ -642,6 +643,50 @@ func (c *cmdInfo) instanceInfo(d lxd.InstanceServer, name string, showLog bool)
...
+				}
+
+				// Find the volume names and snapshot names from the UUIDs
+				volumeSnapshotNames := make([]string, 0, len(attachedVolumeUUIDs))
+				for volUUID, snapUUID := range attachedVolumeUUIDs {
+					for _, dev := range inst.Devices {
```
tomponline: should this be snap.Devices (the volume snapshot might be of a disk device that is no longer attached to the instance)?

lxc/info.go
```diff
@@ -642,6 +643,50 @@ func (c *cmdInfo) instanceInfo(d lxd.InstanceServer, name string, showLog bool)
...
+						if dev["type"] == "disk" && dev["pool"] != "" && dev["source"] != "" {
+							volume, _, err := d.GetStoragePoolVolume(dev["pool"], "custom", dev["source"])
+							if err != nil {
+								return err
+							}
+
+							if volume.Config["volatile.uuid"] == volUUID {
```
tomponline, line 666: Looking at this relatively expensive process to figure out the disk device associated to a snapshot volume, I wonder if we should not rather store this in the instance snapshot's volatile config in a different way.

Similar to the approach I went with for persistent PCIe bus allocation, by using `volatile.<device>.volume_snapshot` - indicating that this device has an associated volume snapshot, and which volume snapshot uuid it is.

And phase out the "volatile.attached_volumes" concept.
?

kadinsayani:
> > | c1-snap1 | 2025/11/21 16:32 UTC | 2025/11/21 18:32 UTC | NO       | shared/snap0, non-shared/snap0 |
> 
> Do you know of other columns where we comma-separate a list inside the field rather than separate with newlines (like IP addresses in `lxc list`)?

None come to mind, I can rework this to follow the new line approach.

tomponline:
> None come to mind, I can rework this to follow the new line approach.

OK cool, would seem to me to give more space for volume names too and be easier to consume.

lxc/info.go
kadinsayani: While I like the idea and agree that it would improve performance, the change would require many modifications to the multi-volume snapshot/restore associated helper functions.

lxc/info.go
kadinsayani: Ack, added error wrapping.

lxc/info.go
kadinsayani: Yes, fixed.

lxc/info.go
tomponline: Yes and a patch to convert any existing `volatile.attached_volume`.

kadinsayani:
> > None come to mind, I can rework this to follow the new line approach.
> 
> OK cool, would seem to me to give more space for volume names too and be easier to consume.

Agreed, thanks for the suggestion :)

lxc/info.go
kadinsayani: I'll log an issue to revisit when time permits.

kadinsayani: https://github.com/canonical/lxd/issues/17067

kadinsayani:
@tomponline shall we merge this and I can fix the lookup later, or close https://github.com/canonical/lxd/issues/17067 first? I'm prioritizing roadmap items at this time, so not sure when I'll get to https://github.com/canonical/lxd/issues/17067.

tomponline:
> @tomponline shall we merge this and I can fix the lookup later, or close #17067 first? I'm prioritizing roadmap items at this time, so not sure when I'll get to #17067.

Lets add it properly later.

PR #17009, opened 2025-11-20
Remove some unused functions

simondeziel:
```
$ git diff --shortstat main..
 21 files changed, 281 deletions(-)
```


lxd-agent/response.go
```diff
@@ -50,18 +50,6 @@ func (r *devLXDResponse) Render(w http.ResponseWriter, req *http.Request) error
 	return err
 }
 
-func (r *devLXDResponse) String() string {
```
tomponline, line 53: Removing `String()` functions is risky, as we've seen in the past, as it might be being used by `fmt.Sprintf()`.

Have you considered this?

lxd/db/generate/lex/lang.go
```diff
@@ -1,19 +1,11 @@
...
-	"fmt"
-)
-
 // VarDecl holds information about a variable declaration.
 type VarDecl struct {
 	Name string
 	Expr string
 }
 
-func (d VarDecl) String() string {
```
tomponline, line 13: Do we know this isnt used?

lxd/instance/drivers/qmp/commands.go
```diff
@@ -996,22 +996,6 @@ func (m *Monitor) BlockJobComplete(deviceNodeName string) error {
 	return nil
 }
 
-// Eject ejects a removable drive.
-func (m *Monitor) Eject(id string) error {
```
tomponline, line 1000: I can see we may need this so lets keep it.

lxd/ip/link.go
```diff
@@ -130,18 +130,6 @@ func (l *Link) SetAddress(address net.HardwareAddr) error {
 	return nil
 }
 
-// SetAllMulticast when enabled instructs network driver to retrieve all multicast packets from the network to the
-// kernel for further processing.
-func (l *Link) SetAllMulticast(enabled bool) error {
```
tomponline, line 135: Lets keep this as it aligns with the `AllMutlicast` property we do use.

lxd/ip/neigh_proxy.go
```diff
@@ -14,36 +13,6 @@ type NeighProxy struct {
 	Addr    net.IP
 }
 
-// Show list neighbour proxy entries.
-func (n *NeighProxy) Show() ([]NeighProxy, error) {
```
tomponline, line 18: I suggest we keep this its useful functionality on the NeighProxy package.

lxd/ip/vdpa.go
```diff
@@ -90,19 +90,6 @@ type VDPADev struct {
 	VhostVDPA *VhostVDPA
 }
 
-// ParseAttributes parses the attributes of a netlink message for a vDPA management device.
-func (d *MgmtVDPADev) parseAttributes(attrs []syscall.NetlinkRouteAttr) error {
```
tomponline, line 94: this seems to be called from `parseMgmtVDPADevList`

lxd/ip/vdpa.go
```diff
@@ -232,26 +232,6 @@ func newNetlinkAttribute(attrType int, data any) (*nl.RtAttr, error) {
 	}
 }
 
-// parseMgmtVDPADevList parses a list of vDPA management device netlink messages.
-func parseMgmtVDPADevList(msgs [][]byte) ([]*MgmtVDPADev, error) {
```
tomponline, line 249: im not sure if we need this for our vdpa support, so id be inclined to keep it in case we need it in the future, as we've been working on sriov offloading recently.

lxd/lifecycle/network_peer.go
```diff
@@ -14,15 +13,3 @@ const (
 	NetworkPeerDeleted = NetworkForwardAction(api.EventLifecycleNetworkPeerDeleted)
 	NetworkPeerUpdated = NetworkForwardAction(api.EventLifecycleNetworkPeerUpdated)
 )
-
-// Event creates the lifecycle event for an action on a network forward.
-func (a NetworkPeerAction) Event(n network, peerName string, requestor *api.EventLifecycleRequestor, ctx map[string]any) api.EventLifecycle {
```
tomponline, line 19: we will likely need this for our lifecycle logging roadmap item, and the fact its not being called is probably a bug.

lxd/migration/migration_volumes.go
```diff
@@ -269,28 +269,6 @@ func progressWrapperRender(op *operations.Operation, key string, description str
 	}
 }
 
-// ProgressReader reports the read progress.
-func ProgressReader(op *operations.Operation, key string, description string) func(io.ReadCloser) io.ReadCloser {
```
tomponline, line 273: `ProgressWriter` is used, which suggests `ProgressReader` is supposed to be used somewhere. And our progress reporting has known limitations currently, so would tend to leave this in place until we know we've got our progress reporting working correctly.

lxd/seccomp/seccomp.go
```diff
@@ -2664,22 +2664,6 @@ func lxcSupportSeccompNotify(state *state.State) error {
 	return nil
 }
 
-// MountSyscallFilter creates a mount syscall filter from the config.
-func MountSyscallFilter(config map[string]string) []string {
```
tomponline, line 2668: @mihalicyn any idea why this isnt used?

lxd-agent/response.go
simondeziel: No, I had not, thanks for raising this (again).

lxd/seccomp/seccomp.go
simondeziel: This `MountSyscallFilter` function is really similar to `SyscallInterceptMountFilter` that is actually used and also handles `security.syscalls.intercept.mount.fuse`.

PR #17010, opened 2025-11-21
fix: use uint64 for interface statistics counters

troyanov:
The /proc/net/dev as well as /sys/class/net/{iface}/statistics uses struct
rtnl_link_stats64, where all counters are __u64.

Previously, parsing these values into int64 caused errors for large counters:

    strconv.ParseInt: parsing "18446744073709551613": value out of range

This commit ensures that all counters are treated as uint64.

See: https://www.kernel.org/doc/html/latest/networking/statistics.html

### DIFF
--- a/doc/rest-api.yaml
+++ b/doc/rest-api.yaml
@@ -2822,49 +2822,49 @@ definitions:
             bytes_received:
                 description: Number of bytes received
                 example: 192021
-                format: int64
+                format: uint64
                 type: integer
                 x-go-name: BytesReceived
             bytes_sent:
                 description: Number of bytes sent
                 example: 10888579
-                format: int64
+                format: uint64
                 type: integer
                 x-go-name: BytesSent
             errors_received:
                 description: Number of errors received
                 example: 14
-                format: int64
+                format: uint64
                 type: integer
                 x-go-name: ErrorsReceived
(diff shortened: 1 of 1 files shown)


simondeziel:
@troyanov FYI, the CI failures with `Error: Required tool "zpool" is missing. ...` are going to be resolved soon once `latest/edge` updates.

shared/units/units.go
tomponline, line 1: This change seems to be different from the requirement of changing the `InstanceStateNetworkCounters` and `NetworkStateCounters`?

alanbach:
Hello @tomponline Thank you so much for working on this - just a side node we have tested changes proposed by @troyanov on an affected machine and it unblocked the client.

shared/units/units.go
troyanov: `units.GetByteSizeString` is used here to return Network usage, 
https://github.com/canonical/lxd/blob/e56d702bcd18bc7b9ec8f4473cffd6421bd62527/lxc/network.go#L957-L958

troyanov:
> Did you find any usage of these metrics in the LXD code base expecting/setting it to <0?

I did not find anything like this. But TBH I didn't check any sophisticated ways to access the field value through reflection

PR #17016, opened 2025-11-21
Derive a context for use in operations

markylaing:
Operations should never be using the request context inside their run function body. The request context may have been cancelled before the operation is started.

When calling `operations.OperationCreate`, a context is passed in which is used to set the operation `Requestor`. The original context is not persisted beyond this.

This PR changes the `operations.Operation` to remove the `cancel.Canceller` in favour of using a cancellable context. The operation context is just a background context with some key information copied from the input context (e.g. the Requestor). Operations can still monitor the daemon shutdown context via their state field if necessary.

I've then updated all operations in LXD to use `(Operation).Context()` wherever a `context.Context` argument is called for inside an operation.

This is another part of https://github.com/canonical/lxd/issues/16613#issuecomment-3547843653

### DIFF
--- a/lxd/operations/operations.go
+++ b/lxd/operations/operations.go
@@ -196,7 +196,7 @@ func operationCreate(s *state.State, requestor *request.Requestor, args Operatio
 	}
 
 	if op.class == OperationClassToken && op.onRun != nil {
-		return nil, errors.New("Token operations can't have a Run hook")
+		return nil, errors.New("Token operations cannot have a Run hook")
 	}
 
 	operationsLock.Lock()


lxd/operations/operations.go
```diff
@@ -107,21 +109,45 @@ type Operation struct {
...
 
 	state  *state.State
 	events *events.Server
 }
 
+func (op *Operation) setupContext(inCtx context.Context, s *state.State) error {
+	opCtx := context.Background()
+	if s != nil {
+		opCtx = s.ShutdownCtx
+	}
```
markylaing: Thinking a little bit more about this. Should an operation only ever have a `context.Background()` plus any extra contents (such as the requestor)?

If the daemon is shut down, then the shutdown context will cancel, and all operations will cancel. This would mean that LXD wouldn't wait for running exec websockets to close before shutting down.

I'll experiment with this

tomponline: The answer I think is "it depends".

There's value in the operation knowing that the daemon is shutting down, so that when the daemon is waiting for operations to finish they may choose to finish prematurely, but still cleanly.

However at which stage the operation chooses to finish is dependent on the operation.

markylaing: Hmm ok.. Well the operation still has access to the shutdown context via the state, so the particular operation can choose to act on that independently of its own context. I will change this commit so that all operations have a background context plus any extra "contextual" information :)

tomponline: Fine for now to limit the scope of the change.

markylaing: Done. Will unmark as draft if I have a clean test run

lxd/operations/operations.go
```diff
@@ -82,6 +82,9 @@ func OperationGetInternal(id string) (*Operation, error) {
 
 // Operation represents an operation.
 type Operation struct {
+	ctx    context.Context
```
skozina: For durable operations, we'd still persist `Requestor` in the db, and just re-construct the new context with the same requestor entity, is that right?

markylaing: Yes you would need to do that I think.

markylaing:
Ok this one is ready for review now

lxd/request/requestor.go
```diff
@@ -361,3 +361,13 @@ func GetRequestor(ctx context.Context) (*Requestor, error) {
 
 	return r, nil
 }
+
+// CopyRequestor returns a new context derived from the "out" context that contains the Requestor from the "in" context.
+func CopyRequestor(inCtx context.Context, outCtx context.Context) (context.Context, error) {
```
tomponline: `Clone()` is more conventional.

tomponline: Or in this case what its actually doing is "Adding" a requestor to the outCtx.

Also can we rename `inCtx` and `outCtx` to `fromCtx` and `toCtx`?

lxd/operations/operations.go
```diff
@@ -107,21 +109,41 @@ type Operation struct {
...
-	finished cancel.Canceller
-
 	// Locking for concurent access to the Operation
 	lock sync.Mutex
 
 	state  *state.State
 	events *events.Server
 }
 
+func (op *Operation) setupContext(inCtx context.Context, s *state.State) error {
```
tomponline: `ctx` rather than `inCtx` will suffice I think as there's only one context being passed.

tomponline: Reading your comments (thanks btw!), I'm reconsidering whether we should have a context passed in at all via `OperationCreate`.

In general its not good practice to store a context in a struct, and worse still, here we are accepting a context, extracting stuff from it and then discarding it and creating a new context, which is misleading to the caller as it is unconventional. Normally I would expect that if I cancel the passed in context that the downstream is cancelled. Which would not be the case here.

Instead I wonder if we should make it explicit that `OperationCreate` takes a `requestor` argument instead, and totally disassociate it from context.

markylaing: Yep this makes so much more sense!

However, I do want to keep the logic which asserts that the requestor should be present in certain scenarios. Perhaps this can be moved to a util in the request package. I'll have a play with it.

tomponline: > However, I do want to keep the logic which asserts that the requestor should be present in certain scenarios.

Yeah that can be a check done in OperationCreate that ensures the requestor is non-nil

markylaing: Yeah I've refactored slightly to have two functions, one for user requested async operations which must have a requestor, and one for server initiated background operations which do not have a requestor

lxd/request/requestor.go
markylaing: This is now `WithRequestor(ctx context.Context, requestor *Requestor) context.Context`

lxd/operations/operations.go
markylaing: I've implemented this now. Keen to know your thoughts.

markylaing:
Just marking as draft while I wait for tests to pass.

markylaing:
Tests passing now

lxd/operations/operations.go
```diff
@@ -313,8 +328,17 @@ func (op *Operation) Start() error {
 	op.status = api.Running
 
 	if op.onRun != nil {
-		go func(op *Operation) {
-			err := op.onRun(op)
+		// The operation context will be a background context plus the requestor.
```
tomponline: Do we really need to also store it in the operation too then?

markylaing: Yes, else we can't pass it into the context in the operation hooks. Per your last comment I removed the `context.Context` field from `operations.Operation` and reverted to the cancel.Canceler which can't store values

lxd/operations/operations.go
```diff
@@ -102,10 +102,10 @@ type Operation struct {
 	logger      logger.Logger
 
 	// Those functions are called at various points in the Operation lifecycle
-	onRun     func(*Operation) error
-	onCancel  func(*Operation) error
-	onConnect func(*Operation, *http.Request, http.ResponseWriter) error
-	onDone    func(*Operation)
+	onRun     RunFunc
```
skozina: Do we need 4 different types? What about only 3 based on the function type? Or perhaps only two, `OperationHandlerType` for the run/cancel/done, and `WebsocketOperationHandlerType` for the onConnect? I suppose that the onDone handler can be required to always return nil.

lxd/operations/operations.go
```diff
@@ -117,9 +117,34 @@ type Operation struct {
...
+func CreateUserOperation(s *state.State, projectName string, requestor *request.Requestor, opClass OperationClass, opType operationtype.Type, opResources map[string][]api.URL, opMetadata any, onRun RunFunc, onCancel CancelFunc, onConnect ConnectFunc) (*Operation, error) {
+	if requestor == nil || requestor.OriginAddress() == "" {
+		return nil, errors.New("Cannot create user operation, the requestor must be set")
+	}
+
+	return operationCreate(s, projectName, requestor, opClass, opType, opResources, opMetadata, onRun, onCancel, onConnect)
+}
+
+// CreateServerOperation creates a new [Operation] that runs as a server background task.
+func CreateServerOperation(s *state.State, projectName string, opClass OperationClass, opType operationtype.Type, opResources map[string][]api.URL, opMetadata any, onRun RunFunc, onCancel CancelFunc, onConnect ConnectFunc) (*Operation, error) {
```
skozina: FYI: For the durable operations I'll be creating one more wrapper, as the durable operations can't have custom onRun/Cancel/Connect handlers. Initially we'll likely have only server-side durable operations, but I suppose we will also create user-requested durable operations. So we have 2 dimensions to look at, I suppose ending up with 4 wrappers is fine.

lxd/operations/operations.go
skozina: I agree. The operation needs it's own context. The context just needs to be internal to the operation, especially on durable operation restarts we'll need to create a new context for the new operation.

tomponline: I can see why @markylaing has done it this way, as its not recommended to store contexts, and especially cancellable ones have their own go routine.

TBH I need more time to digest this which I didn't have yesterday (bugs), so lets leave it as is for now and ill get back to looking at it soon.

The main thing I want to look at is how the context passed to opRun is canceled.

lxd/operations/operations.go
tomponline: Agreed, seems unnecessary duplication of a callback with the same signature. The argument name to the function can indicate the usage.

lxd/operations/operations.go
tomponline: > durable operations can't have custom onRun/Cancel/Connect handlers

Curious. Can't have `onRun` or `Cancel` handlers?

Lets discuss in our 1:1

lxd/operations/operations.go
```diff
@@ -381,8 +405,14 @@ func (op *Operation) Cancel() (chan error, error) {
...
 	if hasOnCancel {
-		go func(op *Operation, oldStatus api.StatusCode, chanCancel chan error) {
-			err := op.onCancel(op)
+		// As with the context in the run func, we copy the requestor to ensure it is sent if any internal calls are made.
+		cancelCtx := context.Background()
+		if op.requestor != nil {
+			cancelCtx = request.WithRequestor(context.Background(), op.requestor)
+		}
+
+		go func(ctx context.Context, op *Operation, oldStatus api.StatusCode, chanCancel chan error) {
```
tomponline: Why does the `onCancel` callback function need to have a ctx passed to it? 
This indicates the cancellation itself could be cancelled?


Would passing the requester itself be sufficient?

lxd/operations/operations.go
markylaing: I thought to make `RunFunc` and `CancelFunc` the same type, but then I couldn't figure out what to call it.

markylaing: Or I can just leave it as it is. The reason for changing was a) the signature of `OperationCreate` was getting ridiculously long, b) IMO it's nicer to have `var run operations.RunFunc` within a handler.

tomponline: We could switch to having an `operationArgs` struct argument, with the options as fields in that struct.
As well as reducing the function signature size, it would have an additional benefit that we can then use our IDEs to find which call sites use specific options, which is more difficult today with them being function arguments.

skozina: I'm all in favor of having some types for the handlers, as I'll need to store the handlers in some table for the durable workflows. `operationArgs` sound good.

markylaing:
> I'm not sure if we even need cancel.HTTPRequestCanceller anywhere in the code base. Although I notice, with dismay, that its part of the public SDK via BackupFileRequest and ImageFileRequest structs.

Just looked into this. What a nightmare. We can't change this really without introducing a context.Context parameter across the whole client interface.

tomponline:
> > I'm not sure if we even need cancel.HTTPRequestCanceller anywhere in the code base. Although I notice, with dismay, that its part of the public SDK via BackupFileRequest and ImageFileRequest structs.
> 
> Just looked into this. What a nightmare. We can't change this really without introducing a context.Context parameter across the whole client interface.

At the very least we should aim to de-couple it from the `Operation` and allow for `cancel.HTTPRequestCanceller` to be cancelled by a context (from the Operation).

tomponline:
> introducing a context.Context parameter across the whole client interface

This has been on the "TODO" list for quite some time too.

markylaing:
> > introducing a context.Context parameter across the whole client interface
> 
> This has been on the "TODO" list for quite some time too.

Shall we bite the bullet? I'm making the same change everywhere else. Of course this is a big breaking change on the client. We can wait until after the next LTS if preferred.

tomponline:
> > > introducing a context.Context parameter across the whole client interface
> > 
> > 
> > This has been on the "TODO" list for quite some time too.
> 
> Shall we bite the bullet? I'm making the same change everywhere else. Of course this is a big breaking change on the client. We can wait until after the next LTS if preferred.

Lets discuss in the team meeting. 

It shouldn't need to hold this change up though as we should be able to untangle `cancel.HTTPRequestCanceller` and `Operation` at least.

lxd/operations/operations.go
markylaing: This is now changed to OperationArgs. I haven't created types for the operation hook function signatures.

markylaing:
This PR is now updated to:
1. Add a context to an operations' main "run" hook. This context contains the operation requestor and can be cancelled in some situations.
2. Change how operations are instantiated. We now how `CreateUserOperation` and `CreateServerOperation`, with the former requiring a `request.Requestor` to ensure that the owner of the operation is explicitly set for auditing and access control purposes.
3. Remove the channel that is returned from operation.Cancel. As far as I can tell this was not used. Callers should use op.Wait.
4. Cancel image download operations via a HTTPRequestCanceller that automatically cancels requests when a given context is cancelled.

I have not added a context argument to any other hooks other than the "run" hook. It seemed redundant in the "cancel" hook, and the onConnect hook has an `*http.Request` argument which has it's own context.

markylaing:
Test failures fixed. Opened a new issue for the underlying problem: https://github.com/canonical/lxd/issues/17220

lxd/operations/operations.go
```diff
@@ -123,49 +123,75 @@ type Operation struct {
...
+	op.resources = args.Resources
 	op.finished = cancel.New()
 	op.running = cancel.New()
 	op.state = s
+	op.requestor = requestor
 	op.logger = logger.AddContext(logger.Ctx{"operation": op.id, "project": op.projectName, "class": op.class.String(), "description": op.description})
 
 	// An operation is cancellable if the cancel hook is not nil or it is a token operation.
-	op.isCancellable = onCancel != nil || opClass == OperationClassToken
+	op.isCancellable = args.CancelHook != nil || args.Class == OperationClassToken
```
kadinsayani: The comment on line 178 is a bit confusing because it states that an operation is cancellable if it is a token operation, but this check follows:

https://github.com/canonical/lxd/blob/2ccb2e9958e179187cfc3ac27a92aa292eacf058/lxd/operations/operations.go#L174-L176

lxd/operations/operations.go
```diff
@@ -184,17 +210,11 @@ func OperationCreate(ctx context.Context, s *state.State, projectName string, op
 		return nil, errors.New("Token operations can't have a Cancel hook")
```
kadinsayani: Not a blocker, but perhaps we can replace the contractions in these "quick check" error messages (can't -> cannot) as per the new CLI guidelines. I've been incrementally patching error messages in preparation for the new CLI to make our work easier in the future. Another change I've been applying is switching messages that start with `Failed to update/create/delete/*` to `Failed updating/creating/deleting/*`.

lxd/operations/operations.go
```diff
@@ -469,7 +469,7 @@ func (op *Operation) Cancel() (chan error, error) {
 	op.sendEvent(md)
 	op.lock.Unlock()
 
-	return chanCancel, nil
+	return nil
```
kadinsayani, line 431: Since we are no longer returning `chanCancel` and `Wait()` waits for `op.done()`, can we remove its initialization and any references to it in this function?

kadinsayani: It looks like `chanCancel` was capturing an error on line 426 that is now lost. Suggest the following approach:

We keep `chanCancel`, but add `err := <-chanCancel` after the goroutine. We can then gate lines 465-470 with `if err == nil {`, and `return err` from this function.

lxd/operations/operations.go
```diff
@@ -360,22 +395,23 @@ func (op *Operation) Start() error {
...
 		op.lock.Unlock()
-		return nil, errors.New("Only running operations can be cancelled")
+		return errors.New("Only running operations can be cancelled")
 	}
 
-	if !op.mayCancel() {
+	if !op.isCancellable {
 		op.lock.Unlock()
-		return nil, errors.New("This operation can't be cancelled")
+		return errors.New("This operation can't be cancelled")
```
kadinsayani:
```suggestion
		return errors.New("Operation cannot be cancelled")
```

kadinsayani: Are there any details about the operation we can include in this error message?

kadinsayani (dismissed):
Overall, looks really good! Just a few nits and suggestions, and a comment on the changes to `Cancel()`.

lxd/operations/operations.go
kadinsayani: Perhaps the operation type?

lxd/operations/operations.go
tomponline: Agreed, let's not use contractions in error messages please.

lxd/operations/operations.go
```diff
@@ -138,9 +145,12 @@ func OperationCreate(ctx context.Context, s *state.State, projectName string, op
 	op.url = api.NewURL().Path(version.APIVersion, "operations", op.id).String()
 	op.resources = opResources
 	op.finished = cancel.New()
+	op.running = cancel.New()
 	op.state = s
 	op.logger = logger.AddContext(logger.Ctx{"operation": op.id, "project": op.projectName, "class": op.class.String(), "description": op.description})
 
+	// An operation is cancellable if the cancel hook is not nil or it is a token operation.
+	op.isCancellable = onCancel != nil || opClass == OperationClassToken
```
tomponline: Small nit, could we pass a no-op cancel function into the `OperationCreate` function for OperationClassToken operations rather than having this operation-type-specific logic in the generic `OperationCreate` function?

lxd/operations/operations.go
```diff
@@ -118,7 +125,7 @@ type Operation struct {
 
 // OperationCreate creates a new operation and returns it. If it cannot be
 // created, it returns an error.
-func OperationCreate(ctx context.Context, s *state.State, projectName string, opClass OperationClass, opType operationtype.Type, opResources map[string][]api.URL, opMetadata any, onRun func(*Operation) error, onCancel func(*Operation) error, onConnect func(*Operation, *http.Request, http.ResponseWriter) error) (*Operation, error) {
+func OperationCreate(ctx context.Context, s *state.State, projectName string, opClass OperationClass, opType operationtype.Type, opResources map[string][]api.URL, opMetadata any, onRun func(context.Context, *Operation) error, onCancel func(*Operation) error, onConnect func(*Operation, *http.Request, http.ResponseWriter) error) (*Operation, error) {
```
tomponline: Did we previously discuss the idea of getting rid of the `ctx` argument on `OperationCreate` and replacing it with an optional (nillable) `requestor` argument. 

This would make it explicit what the operation uses from the outer context, whilst enforcing the idea that operations cant take a parent context.

tomponline: I see you did this as part of introducing operation args. Thanks!

lxd/operations/operations.go
```diff
@@ -175,8 +176,8 @@ func operationCreate(s *state.State, requestor *request.Requestor, args Operatio
 	op.requestor = requestor
 	op.logger = logger.AddContext(logger.Ctx{"operation": op.id, "project": op.projectName, "class": op.class.String(), "description": op.description})
 
-	// An operation is cancellable if the cancel hook is not nil or it is a token operation.
-	op.isCancellable = args.CancelHook != nil || args.Class == OperationClassToken
+	// An operation is cancellable if a) the caller says so (via args), b) the cancel hook is not nil, c) it is a token operation.
+	op.isCancellable = args.IsCancellable || args.CancelHook != nil || args.Class == OperationClassToken
```
tomponline: @markylaing can we simplify this by getting rid of `IsCancellable` and making it so that if a cancel hook is supplied the operation is cancellable?

lxd/operations/operations.go
tomponline: > Operation of type %s cannot be cancelled

Maybe?

tomponline: Although is that always the case? It might be best to keep it generic as it is now

lxd/operations/operations.go
tomponline: The error that has been removed was the error from `onCancel`, I think we should first figure out if any of the existing `onCancel` hooks can return an error, and then decide if we need this feature.

Because I'm not actually sure we need the `onCancel` hook at all - but that remains to be seen.

Really I'd expect the `op.Cancel()` to just signal to the `onRun` function that it needs to cancel and clean up, and any error would then be returned from `op.Wait()`.

lxd/tokens.go
```diff
@@ -60,7 +60,7 @@ func autoRemoveExpiredTokens(ctx context.Context, s *state.State) {
 
 	opRun := func(ctx context.Context, op *operations.Operation) error {
 		for _, op := range expiredTokenOps {
-			_, err := op.Cancel()
+			err := op.Cancel()
```
tomponline, line 63: should we also `op.Wait()` for each cancelled operation to finish?

lxd/operations.go
```diff
@@ -336,7 +336,7 @@ func operationCancel(ctx context.Context, s *state.State, projectName string, op
 	localOp, _ := operations.OperationGetInternal(op.ID)
 	if localOp != nil {
 		if localOp.Status() == api.Running {
-			_, err := localOp.Cancel()
+			err := localOp.Cancel()
```
tomponline, line 340: these seem to be only used for token cancellation, should we `localOp.Wait` here too?

lxd/daemon.go
```diff
@@ -2103,7 +2103,7 @@ func cancelCancelableOps() error {
 		}
 
 		if opAPI.MayCancel {
-			_, _ = op.Cancel()
+			_ = op.Cancel()
```
tomponline: I think we need to also `op.Wait` here to wait for the cancelled operations to finish before we move on with the shutdown process.

tomponline:
@markylaing this is looking really great, its going to be much easier to use and reason about operations once we get your work merged. Thanks!

lxd/operations/operations.go
markylaing: The caller can already do that. The purpose of this was to make it more explicit and not require e.g. the caller to pass in a stubbed CancelHook to make the operation cancellable.

tomponline: What does a cancellable operation that has no onCancel hook do?

Why not make all such operations cancellable?

markylaing: It just cancels the `running` context. We can for sure make all operations cancellable, I just wasn't sure if that would break anything so was trying to be cautious.

lxd/api_vsock.go
```diff
@@ -17,7 +18,8 @@ import (
 // vSockServer creates an http.Server capable of handling /dev/lxd requests over vsock.
 func vSockServer(d *Daemon) *http.Server {
 	return &http.Server{
-		Handler: devLXDAPI(d, vSockAuthenticator{}),
+		Handler:     devLXDAPI(d, vSockAuthenticator{}),
+		ConnContext: request.SaveConnectionInContext,
```
skozina: Can you advice why we need this change?

lxd/operations/operations.go
tomponline: Happy with cautious, but can we also untangle the `Operation` concept from the type-specific logic by pushing that out to the type specific handlers that create the operations, so they indicate if the operation is cancellable by providing an `onCancel` hook, and if not, then the operation isn't cancellable.

Even if that `onCancel` doesn't do anything.

Because as per comment here:

https://github.com/canonical/lxd/pull/17016#discussion_r2609718304

I'm not sure we even need the `onCancel` hook, but we need to audit the existing `onCancel` hooks to see what they do.

lxd/operations/operations.go
markylaing: Tom and I looked at this yesterday and there are no `onCancel` hooks that actually return an error.

Looking further at this this morning, I think that all existing `onCancel` hooks can be replaced by a simple deferred statement.

Even if I missed something, since adding the run context an equivalent to "onCancel" would be:
```
// inside run hook
defer func() {
    if errors.Is(ctx.Err(), context.Canceled) {
        // whatever onCancel was defined as		
    }
}
```

So I'm going to remove the cancel hook altogether :)

lxd/operations/operations.go
markylaing: See https://github.com/canonical/lxd/pull/17016#discussion_r2613520458

lxd/operations/operations.go
tomponline: Lovely stuff thanks

lxd/api_vsock.go
markylaing: We don't! I will remove this.

The reason the change is present here is convoluted - before we changed operation to use an `OperationArgs` struct, we passed in a `context.Context`.

That context was either derived from a request context, a daemon shutdown context, or a background (or TODO) context. When an operation is instantiated as part of a request, we had no way to enforce that the context contained a requestor. 

The ConnContext field of a http.Server is a function which allows you to add the connection to the request context. We use this to extract the raw connection for a few reasons - I can't remember them off the top of my head, but I think one is the dqlite connection upgrade and another is for a container exec websocket.

So, before adding `OperationArgs`, I was checking if a context contained a `net.Conn` and if so, enforcing that the requestor was present.

This broke tests that Din recently added for devlxd volume management because there was no `net.Conn` in the request context for devlxd requests over vsock.

skozina: Cool;-) Thanks for the thorough explanation!

markylaing:
@tomponline tests finally ran and passed. I've also changed the operation delete endpoint to emit the lifecycle event before waiting on the operation.

Ready for review when you're ready 😌

lxd/operations/operations.go
```diff
@@ -331,8 +324,14 @@ func (op *Operation) Start() error {
 			err := op.onRun(ctx, op)
 			if err != nil {
 				op.lock.Lock()
-				op.status = api.Failure
-				op.running.Cancel()
+				// If the run context was cancelled, the previous state should be "cancelling", and the final state should be "cancelled".
+				if op.running.Err() != nil {
```
tomponline: @markylaing should we also check here that the error returned from `op.onRun` is non-nil and if its a canceled context error, only then set the op.status to Cancelled, and if not take the error from `onRun` as the operations status?

lxd/instance_exec.go
```diff
@@ -272,7 +274,7 @@ func (s *execWs) Do(ctx context.Context, op *operations.Operation) error {
 		}
 
 		// Make VM disconnections (shutdown/reboot) match containers.
-		if cmdErr == drivers.ErrExecDisconnected {
+		if cmdErr == drivers.ErrExecDisconnected || ctx.Err() != nil {
```
tomponline: I'm not sure about this bit, what was the reasoning here?

lxd/instance_exec.go
```diff
@@ -319,6 +321,18 @@ func (s *execWs) Do(ctx context.Context, op *operations.Operation) error {
...
+			select {
+			case <-done:
+			case <-ctx.Done():
+				// If the websocket operation is deleted, close cleanly.
+				closeMsg := websocket.FormatCloseMessage(websocket.CloseNormalClosure, "")
+				_ = conn.WriteMessage(websocket.CloseMessage, closeMsg)
+			}
+		}()
+
 		for {
```
tomponline, line 343: Rather than having a additional go routine here, can we instead add a check for `ctx.Err() != nil` in the for loop on line 336 just below it, and if we detect the context is cancelled then call `conn.Close()` and let the existing logic handle cleaning up the process?

lxd/operations/operations.go
markylaing: Will check `if errors.Is(err, context.Canceled)` and only set it in that case.

lxd/instance_exec.go
markylaing: Ah - I forgot to add a comment, sorry. This is so that if an admin boots a user from an exec session by deleting the operation, the exit code for the user is 129. I figured that would be a good error code to indicate disconnection?

lxd/instance_exec.go
markylaing: I tried that but unfortunately we end up blocking in the for loop waiting for the next message :/

lxd/instance_exec.go
tomponline: OK makes sense thanks

lxd/instance_exec.go
tomponline: Can you try calling `cmdKill` from the for loop when the context is cancelled and then return from the `wgEOF.Go` function. This might avoid the need for the `if cmdErr == drivers.ErrExecDisconnected || ctx.Err() != nil {` bit too and just make it look like the command was killed (which it was).

markylaing: It's not possible to do this in the for loop because it blocks at the beginning of the loop:
```go
for {
	mt, r, err := conn.NextReader() // <--- blocks
	/* do stuff */
}
```
There is no way for NextReader to return early if the context is cancelled.

Happy to call cmdKill from the goroutine that waits on context cancellation, but I don't think the extra routine is avoidable

tomponline: > It's not possible to do this in the for loop because it blocks at the beginning of the loop:
> 
> for {
> 	mt, r, err := conn.NextReader() // <--- blocks
> 	/* do stuff */
> }
> 
> There is no way for NextReader to return early if the context is cancelled.
> 


OK that makes sense.

> Happy to call cmdKill from the goroutine that waits on context cancellation, but I don't think the extra routine is avoidable

Lets do that then, because I don't like the way the context handling is getting involved with websocket responses - that seems to me to be a kind of layering violation - we want to signal that the process should end, and let the existing code deal with websocket handling.

lxd/operations/operations.go
markylaing: Done.

lxd/instance_exec.go
markylaing: I've moved this to a different block with a comment. I've also only set the code to 129 if the existing exit code is 0, to ensure a non-zero exit code.

lxd/instance_exec.go
markylaing: Done.

lxd/instance_exec.go
tomponline: OK thanks, do we need this part anymore, now that we're killing the process on cancel?

lxd/operations/operations.go
```diff
@@ -376,81 +378,43 @@ func (op *Operation) Start() error {
 
 // Cancel cancels a running operation. If the operation cannot be cancelled, it
 // returns an error.
-func (op *Operation) Cancel() (chan error, error) {
+func (op *Operation) Cancel() error {
```
tomponline, line 381: Can we simplify this even more and get rid of the `error` returned from `Cancel()` and have it just detected for non-running operations and return early (it should still cancel the `op.running.Cancel()` canceler though because that can be called multiple times and ensures consistency of behaviour in any derived contexts.

tomponline:
@markylaing btw is `UpdateResources()` unused?

markylaing:
> @markylaing btw is `UpdateResources()` unused?

I'll check

lxd/instance_exec.go
```diff
@@ -277,6 +279,13 @@ func (s *execWs) Do(ctx context.Context, op *operations.Operation) error {
 			cmdErr = nil
 		}
 
+		// If the exec operation was deleted, the exec websocket is closed and the caller is booted from the instance.
+		// In this case, make sure there is a non-zero exit code. The disconnection exit code is used here.
+		if ctx.Err() != nil && cmdResult == 0 {
```
tomponline, line 284: @markylaing i just tested removing this part and I dont see that its needed, please can it be removed.

PR #17017, opened 2025-11-21
doc: replace markdownlint with pymarkdownlnt

minaelee:
This PR replaces the Markdown linter for the docs to the standard one used across Canonical docs.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1194,16 +1194,16 @@ jobs:
           sudo apt-get install aspell aspell-en
           sudo snap install mdl
 
-      - name: Run markdown linter
-        run: |
-          set -eux
-          make doc-lint
-
       - name: Build docs (Sphinx)
         run: |
           set -eux
           make doc
           if [ -s doc/.sphinx/warnings.txt ]; then cat doc/.sphinx/warnings.txt; exit 1; fi
+
+      - name: Run markdown linter
+        run: |
+          set -eux
+          make doc-lint
 
       - name: Run spell checker
         run: |


doc/Makefile
```diff
@@ -44,11 +44,23 @@ objects:
...
+	pymarkdownlnt --config .sphinx/.pymarkdown.json scan \
+		--recurse \
+		--exclude=./.sphinx \
+		--exclude=./_build \
+		--exclude=./reference/manpages/lxc \
+		. \
+	&& echo "Passed" \
+	|| (echo "Failed"; exit 1)
 
-.PHONY: html
```
simondeziel, line 51: Was that dropped by accident?

doc/Makefile
```diff
@@ -44,11 +44,23 @@ objects:
...
 
-.PHONY: lint
-lint:
-	.sphinx/.markdownlint/doc-lint.sh
+.PHONY: pymarkdownlnt-install
+pymarkdownlnt-install:
+	. .sphinx/venv/bin/activate; test -d .sphinx/venv/lib/python*/site-packages/pymarkdown || pip install pymarkdownlnt
+
+.PHONY: lint lint-md
+lint lint-md: pymarkdownlnt-install
```
simondeziel, line 52: I'm not familiar with this multi-target syntax. Is adding the `lint-md` required for compatibility with other docs at Canonical? Not an issue, I'm just curious.

doc/Makefile
minaelee: Yes, thanks!

doc/Makefile
minaelee: Yes - the starter pack uses `lint-md` since it supports both Markdown and reStructuredText; we only use Markdown.

PR #17022, opened 2025-11-24
doc: improve description of feature releases

minaelee:
This PR updates documentation about feature releases. It previously described feature releases as not recommended for production because they are "less tested"; this PR clarifies that the reason is due to the frequency of updates to the API and database in feature releases.


tomponline (approved):
ta

PR #17023, opened 2025-11-24
Backports (stable-5.21)

tomponline:


lxd/identities.go
```diff
@@ -1290,11 +1290,6 @@ func updateIdentity(authenticationMethod string) func(d *Daemon, r *http.Request
 			return response.SmartError(err)
 		}
 
-		// Only identities of type api.IdentityTypeCertificateClient may update their own certificate
-		if id.Type != api.IdentityTypeCertificateClient {
```
tomponline, line 1269: @kadinsayani @markylaing does this backport look right (and the similar check removal below)?

.github/workflows/commits.yml
```diff
@@ -31,6 +31,3 @@ jobs:
 
           echo "Invalid branch target: ${TARGET} != ${TARGET_FROM_PR_TITLE}"
           exit 1
-
-      - name: Check if CLA signed
-        uses: canonical/has-signed-canonical-cla@5d1443b94417bd150ad234a82fe21f7340a25e4d # 2
```
simondeziel, line 36: I see that the Org CLA check was done for this PR so I'm fine with dropping it but just noted that we've reintroduced it in the `main` branch until the official announcement that the Org one is "production ready".

tomponline: yeah ill leave it out of this branch as we should remove it now from main branch too i think

tomponline: ill double check though...

lxd/identities.go
kadinsayani: Followed up on MM. This backport is fine, however, we're missing a check in `main` disallowing updating an OIDC identity with a TLS certificate (only possible as admin). Once that check is added in `main`, we can include its associated commit in a future backport.

.github/workflows/tests.yml
kadinsayani, line 1: @tomponline the original [commit](https://github.com/canonical/lxd/commit/0c7226ae3bbb68604cf6ef641c6509565843479f) was included with placement groups. Was the lxc binary size check failing here?

tomponline: yes

.github/workflows/commits.yml
simondeziel: To drop it from the `main` branch: https://github.com/canonical/lxd/pull/17051

PR #17045, opened 2025-11-24
Durable workflows

skozina:
This depends, or rather contains, also changes included in #17589 and #17462.

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -3033,3 +3033,8 @@ This is primarily useful when combined with the `initial.` `disk` device configu
 Storage pool and network endpoints that were previously synchronous now return background operations. This affects create, update, delete, and rename actions on storage pools, networks, network ACLs, network zones, network zone records, network forwards, network load balancers, network peers, and storage buckets (including bucket keys).
 
 Clients should check for this extension and handle the asynchronous response by waiting on the returned operation. Operation metadata may include additional data, such as storage bucket admin credentials on bucket creation.
+
+(extension-durable-operations)=
+## `durable_operations`
+
+Introduces new operation class for durable operations. Durable operations are guaranteed to run on a different node if the node running the operation previously goes offline.
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -3657,3 +3657,10 @@ Creating a public cluster link is a two-phase process:
 A new `ClusterLinkCertificate` response type is introduced, containing the certificate fingerprint and PEM-encoded certificate returned by the first request.
 
 Public cluster links cannot be used by {ref}`replicators <exp-replicators>` or as a project's `replica.cluster`, since replication requires the remote cluster to authenticate the connection. Both reject a public link at configuration time.
+
+(extension-durable-operations)=
+## `durable_operations`
+
+Introduces new operation class for durable operations.
+Durable operations are restarted on the DQLite raft leader if the member that is running the operation fails to respond to heartbeats.
+If the leader was running the operation and goes offline, the operation is restarted on the newly elected leader.
--- a/lxd/cluster/heartbeat.go
+++ b/lxd/cluster/heartbeat.go
@@ -311,6 +311,42 @@ func (g *Gateway) HeartbeatRestart() bool {
 	return false
 }
 
+func (g *Gateway) cancelDurableOperationsIfHeartbeatNotReceived(ctx context.Context, duration time.Duration) {
+	select {
+	case <-time.After(duration):
+		// Heartbeat not received in time, cancel durable operations.
+		logger.Error("Heartbeat not received in time, cancelling durable operations")
+		operations.CancelLocalDurableOperations()
+	case <-ctx.Done():
+		// Heartbeat received in time, nothing to do.
+	}
+}
+
+// HeartbeatReceived is called when a heartbeat is received on this node.
+// It resets the heartbeat detection timer.
+func (g *Gateway) HeartbeatReceived() {
+	g.heartbeatDetectionCancellerLock.Lock()
+	if g.heartbeatDetectionCanceller != nil {
+		g.heartbeatDetectionCanceller()
+	}
+
+	// Setup the new heartbeat detection context.
+	heartbeatDetectionCtx, cancel := context.WithCancel(g.shutdownCtx)
+	g.heartbeatDetectionCanceller = cancel
+	g.heartbeatDetectionCancellerLock.Unlock()
+
+	// The heartbeats round is send at least each gateway.heartbeatInterval() seconds.
+	// The heartbeats send to each node are evenly distributed in an interval <0, gateway.heartbeatInterval() - 3s) to distribute the load across the cluster.
+	// Each node then has up to 2s to respond to the heartbeat, and another 1s for the HTTP client to get more useful info.
+	// While the heartbeats are sent to the nodes in order of the node IDs, the heartbeat to a specific node can be sent at an arbitrary time in the interval,
+	// in case when other nodes have been removed from the cluster (or new nodes were added with higher IDs) and thus the node distribution has changed.
+	// So we must assume the upper limit of when a heartbeat can be sent to a specific node is the full heartbeat interval.
+	// All together this is 2*gateway.heartbeatInterval(), so that's the duration we wait before considering that we have not received a heartbeat.
+	nextHeartbeatTimeout := 2 * g.heartbeatInterval()
+
+	go g.cancelDurableOperationsIfHeartbeatNotReceived(heartbeatDetectionCtx, nextHeartbeatTimeout)
+}
+
 func (g *Gateway) heartbeat(ctx context.Context, mode HeartbeatMode) {
 	// Avoid concurent heartbeat loops.
 	// This is possible when both the regular task and the out of band heartbeat round from a dqlite
@@ -321,6 +357,19 @@ func (g *Gateway) heartbeat(ctx context.Context, mode HeartbeatMode) {
 	if g.Cluster == nil || g.server == nil || g.memoryDial != nil {
 		// We're not a raft node or we're not clustered
 		return
+	}
+
+	isLeader, err := g.isLeader()
+	if err != nil {
+		logger.Warn("Failed determining if node is leader", logger.Ctx{"err": err})
+	}
+
+	// If we're a leader, we are not going to send heartbeats to ourselves.
+	// So just record like that we have received one.
+	// This is useful if we started as a leader, but later we lost connection to the rest of the cluster, and therefore we lost leadership.
+	// In such case we should have a heartbeat detection enabled and cancel durable operations if heartbeats don't arrive in time.
+	if isLeader {
+		g.HeartbeatReceived()
 	}
 
 	// Acquire the cancellation lock and populate it so that this heartbeat round can be cancelled if a
--- a/lxd/cluster/heartbeat.go
+++ b/lxd/cluster/heartbeat.go
@@ -15,6 +15,7 @@ import (
 	"github.com/canonical/lxd/lxd/db/query"
 	"github.com/canonical/lxd/lxd/db/warningtype"
 	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/lxd/lxd/state"
 	"github.com/canonical/lxd/lxd/task"
 	"github.com/canonical/lxd/lxd/warnings"
 	"github.com/canonical/lxd/shared"
@@ -147,7 +148,7 @@ func (hbState *APIHeartbeat) Update(fullStateList bool, raftNodes []db.RaftNode,
 }
 
 // Send sends heartbeat requests to the nodes supplied and updates heartbeat state.
-func (hbState *APIHeartbeat) Send(ctx context.Context, networkCert *shared.CertInfo, serverCert *shared.CertInfo, localAddress string, nodes []db.NodeInfo, spreadDuration time.Duration) {
+func (hbState *APIHeartbeat) Send(ctx context.Context, s *state.State, networkCert *shared.CertInfo, serverCert *shared.CertInfo, localAddress string, nodes []db.NodeInfo, spreadDuration time.Duration) {
 	heartbeatsWg := sync.WaitGroup{}
 	sendHeartbeat := func(nodeID int64, address string, delay time.Duration, heartbeatData *APIHeartbeat) {
 		defer heartbeatsWg.Done()
@@ -418,14 +419,14 @@ func (g *Gateway) heartbeat(ctx context.Context, mode HeartbeatMode) {
 	// Send stale set to all nodes in database to get a fresh set of active nodes.
 	if mode == HeartbeatInitial {
 		hbState.Update(false, raftNodes, members, g.HeartbeatOfflineThreshold)
-		hbState.Send(ctx, g.networkCert, serverCert, localClusterAddress, members, spreadDuration)
+		hbState.Send(ctx, s, g.networkCert, serverCert, localClusterAddress, members, spreadDuration)
 
 		// We have the latest set of node states now, lets send that state set to all nodes.
 		hbState.FullStateList = true
-		hbState.Send(ctx, g.networkCert, serverCert, localClusterAddress, members, spreadDuration)
+		hbState.Send(ctx, s, g.networkCert, serverCert, localClusterAddress, members, spreadDuration)
 	} else {
 		hbState.Update(true, raftNodes, members, g.HeartbeatOfflineThreshold)
-		hbState.Send(ctx, g.networkCert, serverCert, localClusterAddress, members, spreadDuration)
+		hbState.Send(ctx, s, g.networkCert, serverCert, localClusterAddress, members, spreadDuration)
 	}
 
 	// Check if context has been cancelled.
@@ -468,7 +469,7 @@ func (g *Gateway) heartbeat(ctx context.Context, mode HeartbeatMode) {
 		// If any new nodes found, send heartbeat to just them (with full node state).
 		if len(newMembers) > 0 {
 			hbState.Update(true, raftNodes, members, g.HeartbeatOfflineThreshold)
-			hbState.Send(ctx, g.networkCert, serverCert, localClusterAddress, newMembers, 0)
+			hbState.Send(ctx, s, g.networkCert, serverCert, localClusterAddress, newMembers, 0)
 		}
 	}
 
--- a/lxd/cluster/heartbeat.go
+++ b/lxd/cluster/heartbeat.go
@@ -14,6 +14,7 @@ import (
 	"github.com/canonical/lxd/lxd/db"
 	"github.com/canonical/lxd/lxd/db/query"
 	"github.com/canonical/lxd/lxd/db/warningtype"
+	"github.com/canonical/lxd/lxd/operations"
 	"github.com/canonical/lxd/lxd/response"
 	"github.com/canonical/lxd/lxd/state"
 	"github.com/canonical/lxd/lxd/task"
@@ -193,6 +194,11 @@ func (hbState *APIHeartbeat) Send(ctx context.Context, s *state.State, networkCe
 				})
 				if err != nil {
 					logger.Warn("Failed creating warning", logger.Ctx{"err": err})
+				}
+
+				err = operations.RestartDurableOperationsFromNode(ctx, s, nodeID)
+				if err != nil {
+					logger.Warn("Failed restarting durable operations from offline node", logger.Ctx{"node_id": nodeID, "err": err})
 				}
 			}
 		}
--- a/lxd/cluster/heartbeat.go
+++ b/lxd/cluster/heartbeat.go
@@ -15,6 +15,7 @@ import (
 	"github.com/canonical/lxd/lxd/db/query"
 	"github.com/canonical/lxd/lxd/db/warningtype"
 	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/lxd/lxd/state"
 	"github.com/canonical/lxd/lxd/task"
 	"github.com/canonical/lxd/lxd/warnings"
 	"github.com/canonical/lxd/shared"
@@ -149,7 +150,7 @@ func (hbState *APIHeartbeat) Update(fullStateList bool, raftNodes []db.RaftNode,
 }
 
 // Send sends heartbeat requests to the nodes supplied and updates heartbeat state.
-func (hbState *APIHeartbeat) Send(ctx context.Context, networkCert *shared.CertInfo, serverCert *shared.CertInfo, localAddress string, nodes []db.NodeInfo, spreadDuration time.Duration) {
+func (hbState *APIHeartbeat) Send(ctx context.Context, s *state.State, networkCert *shared.CertInfo, serverCert *shared.CertInfo, localAddress string, nodes []db.NodeInfo, spreadDuration time.Duration, offlineThreshold time.Duration) {
 	heartbeatsWg := sync.WaitGroup{}
 	sendHeartbeat := func(nodeID int64, address string, delay time.Duration, heartbeatData *APIHeartbeat) {
 		defer heartbeatsWg.Done()
@@ -430,16 +431,17 @@ func (g *Gateway) heartbeat(ctx context.Context, mode HeartbeatMode) {
 	// If this leader node hasn't sent a heartbeat recently, then its node state records
 	// are likely out of date, this can happen when a node becomes a leader.
 	// Send stale set to all nodes in database to get a fresh set of active nodes.
+	offlineThreshold := g.offlineThreshold()
 	if mode == HeartbeatInitial {
-		hbState.Update(false, raftNodes, members, g.HeartbeatOfflineThreshold)
-		hbState.Send(ctx, g.networkCert, serverCert, localClusterAddress, members, spreadDuration)
+		hbState.Update(false, raftNodes, members, offlineThreshold)
+		hbState.Send(ctx, s, g.networkCert, serverCert, localClusterAddress, members, spreadDuration, offlineThreshold)
 
 		// We have the latest set of node states now, lets send that state set to all nodes.
 		hbState.FullStateList = true
-		hbState.Send(ctx, g.networkCert, serverCert, localClusterAddress, members, spreadDuration)
+		hbState.Send(ctx, s, g.networkCert, serverCert, localClusterAddress, members, spreadDuration, offlineThreshold)
 	} else {
-		hbState.Update(true, raftNodes, members, g.HeartbeatOfflineThreshold)
-		hbState.Send(ctx, g.networkCert, serverCert, localClusterAddress, members, spreadDuration)
+		hbState.Update(true, raftNodes, members, offlineThreshold)
+		hbState.Send(ctx, s, g.networkCert, serverCert, localClusterAddress, members, spreadDuration, offlineThreshold)
 	}
 
 	// Check if context has been cancelled.
@@ -481,8 +483,8 @@ func (g *Gateway) heartbeat(ctx context.Context, mode HeartbeatMode) {
 
 		// If any new nodes found, send heartbeat to just them (with full node state).
 		if len(newMembers) > 0 {
-			hbState.Update(true, raftNodes, members, g.HeartbeatOfflineThreshold)
-			hbState.Send(ctx, g.networkCert, serverCert, localClusterAddress, newMembers, 0)
+			hbState.Update(true, raftNodes, members, offlineThreshold)
+			hbState.Send(ctx, s, g.networkCert, serverCert, localClusterAddress, newMembers, 0, offlineThreshold)
 		}
 	}
 
--- a/lxd/cluster/heartbeat.go
+++ b/lxd/cluster/heartbeat.go
@@ -14,6 +14,7 @@ import (
 	"github.com/canonical/lxd/lxd/db"
 	"github.com/canonical/lxd/lxd/db/query"
 	"github.com/canonical/lxd/lxd/db/warningtype"
+	"github.com/canonical/lxd/lxd/operations"
 	"github.com/canonical/lxd/lxd/response"
 	"github.com/canonical/lxd/lxd/state"
 	"github.com/canonical/lxd/lxd/task"
@@ -196,6 +197,15 @@ func (hbState *APIHeartbeat) Send(ctx context.Context, s *state.State, networkCe
 				})
 				if err != nil {
 					logger.Warn("Failed creating warning", logger.Ctx{"err": err})
+				}
+
+				offlineTime := time.Now().Add(-offlineThreshold)
+				lastHeartbeat := heartbeatData.Members[nodeID].LastHeartbeat
+				if lastHeartbeat.Before(offlineTime) || lastHeartbeat.Equal(offlineTime) {
+					err = operations.RestartDurableOperationsFromNode(ctx, s, nodeID)
+					if err != nil {
+						logger.Warn("Failed restarting durable operations from offline node", logger.Ctx{"node_id": nodeID, "err": err})
+					}
 				}
 			}
 		}
--- a/lxd/cluster/heartbeat.go
+++ b/lxd/cluster/heartbeat.go
@@ -270,14 +270,19 @@ func HeartbeatTask(gateway *Gateway) (task.Func, task.Schedule) {
 	return heartbeatWrapper, schedule
 }
 
-// heartbeatInterval returns heartbeat interval to use.
-func (g *Gateway) heartbeatInterval() time.Duration {
+// offlineThreshold returns the currently configured offline threshold or the default if not set.
+func (g *Gateway) offlineThreshold() time.Duration {
 	threshold := g.HeartbeatOfflineThreshold
 	if threshold <= 0 {
 		threshold = time.Duration(db.DefaultOfflineThreshold) * time.Second
 	}
 
-	return threshold / 2
+	return threshold
+}
+
+// heartbeatInterval returns heartbeat interval to use.
+func (g *Gateway) heartbeatInterval() time.Duration {
+	return g.offlineThreshold() / 2
 }
 
 // HearbeatCancelFunc returns the function that can be used to cancel an ongoing heartbeat.
--- a/lxd/cluster/heartbeat.go
+++ b/lxd/cluster/heartbeat.go
@@ -323,6 +323,32 @@ func (g *Gateway) HeartbeatRestart() bool {
 	return false
 }
 
+func (g *Gateway) cancelDurableOperationsIfHeartbeatNotReceived(ctx context.Context, duration time.Duration) {
+	timer := time.NewTimer(duration)
+	select {
+	case <-timer.C:
+		// Heartbeat not received in time, cancel durable operations.
+		logger.Error("Heartbeat not received in time, cancelling durable operations")
+		operations.CancelLocalDurableOperations()
+	case <-ctx.Done():
+		// Heartbeat received in time, cancel the timer.
+		timer.Stop()
+	}
+}
+
+// heartbeatReceived is called when a heartbeat is received on this node.
+// It resets the heartbeat detection timer.
+func (g *Gateway) heartbeatReceived() {
+	ctx, newCancel := context.WithCancel(g.shutdownCtx)
+	oldCancel := g.heartbeatDetectionCanceller.Swap(&newCancel)
+	if oldCancel != nil && *oldCancel != nil {
+		(*oldCancel)()
+	}
+
+	// Wait for the next heartbeat and cancel durable operations if not received within the offline threshold.
+	go g.cancelDurableOperationsIfHeartbeatNotReceived(ctx, g.offlineThreshold())
+}
+
 func (g *Gateway) heartbeat(ctx context.Context, mode HeartbeatMode) {
 	// Avoid concurrent heartbeat loops.
 	// This is possible when both the regular task and the out of band heartbeat round from a dqlite
@@ -339,6 +365,19 @@ func (g *Gateway) heartbeat(ctx context.Context, mode HeartbeatMode) {
 	if cluster == nil || server == nil || memoryDial != nil {
 		// We're not a raft node or we're not clustered
 		return
+	}
+
+	isLeader, err := g.isLeader()
+	if err != nil {
+		logger.Warn("Failed determining if node is leader", logger.Ctx{"err": err})
+	}
+
+	// If we're a leader, we are not going to send heartbeats to ourselves.
+	// So just record like that we have received one.
+	// This is useful if we started as a leader, but later we lost connection to the rest of the cluster, and therefore we lost leadership.
+	// In such case we should have a heartbeat detection enabled and cancel durable operations if heartbeats don't arrive in time.
+	if isLeader {
+		g.heartbeatReceived()
 	}
 
 	// Acquire the cancellation lock and populate it so that this heartbeat round can be cancelled if a
--- a/lxd/db/cluster/operations.go
+++ b/lxd/db/cluster/operations.go
@@ -26,6 +26,8 @@ import (
 //
 //go:generate mapper stmt -e operation objects
 //go:generate mapper stmt -e operation objects-by-NodeID
+//go:generate mapper stmt -e operation objects-by-NodeID-and-Class
+//go:generate mapper stmt -e operation objects-by-Class
 //go:generate mapper stmt -e operation objects-by-ID
 //go:generate mapper stmt -e operation objects-by-UUID
 //go:generate mapper stmt -e operation objects-by-Parent
@@ -74,6 +76,7 @@ type OperationFilter struct {
 	NodeID *int64
 	UUID   *string
 	Parent *int64
+	Class  *int64
 }
 
 // RequestorProtocol is the database representation of the Requestor Protocol.
--- a/lxd/db/cluster/operations.go
+++ b/lxd/db/cluster/operations.go
@@ -160,24 +160,21 @@ func (r *RequestorProtocol) Value() (driver.Value, error) {
 // UpdateOperation updates operation status, metadata and error (if set) in the cluster db.
 // This is used to keep DB in sync with the current status of the operation when the operation changes
 // its status, or when calls to commit metadata explicitly.
-func UpdateOperation(ctx context.Context, tx *sql.Tx, opUUID string, updatedAt time.Time, newStatus api.StatusCode, metadata string, opErr string, opErrCode int64) error {
-	stmt := `UPDATE operations SET updated_at = ?, status_code = ?, metadata = ?, error = ?, error_code = ? WHERE uuid = ?`
+// Returns number of rows updated in the DB, or an error.
+func UpdateOperation(ctx context.Context, tx *sql.Tx, opUUID string, nodeID int64, updatedAt time.Time, newStatus api.StatusCode, metadata string, opErr string, opErrCode int64) (int64, error) {
+	stmt := `UPDATE operations SET updated_at = ?, status_code = ?, metadata = ?, error = ?, error_code = ? WHERE uuid = ? AND node_id = ?`
 
-	result, err := tx.ExecContext(ctx, stmt, updatedAt, newStatus, metadata, opErr, opErrCode, opUUID)
+	result, err := tx.ExecContext(ctx, stmt, updatedAt, newStatus, metadata, opErr, opErrCode, opUUID, nodeID)
 	if err != nil {
-		return fmt.Errorf("Failed updating operation status: %w", err)
+		return 0, fmt.Errorf("Failed updating operation status: %w", err)
 	}
 
 	n, err := result.RowsAffected()
 	if err != nil {
-		return fmt.Errorf("Fetch affected rows: %w", err)
+		return 0, fmt.Errorf("Fetch affected rows: %w", err)
 	}
 
-	if n != 1 {
-		return fmt.Errorf("Query updated %d rows instead of 1", n)
-	}
-
-	return nil
+	return n, nil
 }
 
 // GetOperationResources loads operation resources from the cluster db.
--- a/lxd/db/cluster/operations.go
+++ b/lxd/db/cluster/operations.go
@@ -364,3 +364,24 @@ func ClearStaleOperationsFromNodes(ctx context.Context, tx *sql.Tx, nodeIDs ...i
 
 	return nil
 }
+
+// UpdateOperationNodeID updates the node_id field of an existing operation in the cluster db.
+func UpdateOperationNodeID(ctx context.Context, tx *sql.Tx, opUUID string, newNodeID int64, updatedAt time.Time) error {
+	stmt := `UPDATE operations SET node_id = ?, updated_at = ? WHERE uuid = ?`
+
+	result, err := tx.ExecContext(ctx, stmt, newNodeID, updatedAt, opUUID)
+	if err != nil {
+		return fmt.Errorf("Failed updating operation node ID: %w", err)
+	}
+
+	n, err := result.RowsAffected()
+	if err != nil {
+		return fmt.Errorf("Fetch affected rows: %w", err)
+	}
+
+	if n != 1 {
+		return fmt.Errorf("Query updated %d rows instead of 1", n)
+	}
+
+	return nil
+}
--- a/lxd/db/cluster/operations.go
+++ b/lxd/db/cluster/operations.go
@@ -34,13 +34,11 @@ import (
 //go:generate mapper stmt -e operation create
 //go:generate mapper stmt -e operation create-or-replace
 //go:generate mapper stmt -e operation delete-by-UUID
-//go:generate mapper stmt -e operation delete-by-NodeID
 //
 //go:generate mapper method -i -e operation GetMany
 //go:generate mapper method -i -e operation Create
 //go:generate mapper method -i -e operation CreateOrReplace
 //go:generate mapper method -i -e operation DeleteOne-by-UUID
-//go:generate mapper method -i -e operation DeleteMany-by-NodeID
 //go:generate goimports -w operations.mapper.go
 //go:generate goimports -w operations.interface.mapper.go
 
--- a/lxd/db/cluster/operations.go
+++ b/lxd/db/cluster/operations.go
@@ -248,6 +248,12 @@ func GetOperations(ctx context.Context, tx *sql.Tx, includeChildren bool, projec
 	return query.Select[Operation](ctx, tx, clause, args...)
 }
 
+// GetOperationWithChildren gets a single operation and all of its children (if it has any).
+func GetOperationWithChildren(ctx context.Context, tx *sql.Tx, operationUUID string) ([]Operation, error) {
+	clause := "WHERE operations.uuid = ? OR operations.parent = (SELECT operations.id FROM operations WHERE operations.uuid = ?)"
+	return query.Select[Operation](ctx, tx, clause, operationUUID, operationUUID)
+}
+
 // UpdateOperation updates operation status, metadata and error (if set) in the cluster db.
 // This is used to keep DB in sync with the current status of the operation when the operation changes
 // its status, or when calls to commit metadata explicitly. The caller is expected to pass in the current node ID.
--- a/lxd/db/cluster/operations.go
+++ b/lxd/db/cluster/operations.go
@@ -436,3 +436,8 @@ func CountOperationChildren(ctx context.Context, tx *sql.Tx, parentID int64) (in
 
 	return count, nil
 }
+
+// GetOperationsByNodeIDAndClass gets all operations on the given node ID and operation class.
+func GetOperationsByNodeIDAndClass(ctx context.Context, tx *sql.Tx, nodeID int64, class operationtype.Class) ([]Operation, error) {
+	return query.Select[Operation](ctx, tx, "WHERE operations.node_id = ? AND operations.class = ?", nodeID, class)
+}
--- a/lxd/db/cluster/operations.mapper.go
+++ b/lxd/db/cluster/operations.mapper.go
@@ -87,10 +87,6 @@ var operationDeleteByUUID = RegisterStmt(`
 DELETE FROM operations WHERE uuid = ?
 `)
 
-var operationDeleteByNodeID = RegisterStmt(`
-DELETE FROM operations WHERE node_id = ?
-`)
-
 // getOperations can be used to run handwritten sql.Stmts to return a slice of objects.
 func getOperations(ctx context.Context, stmt *sql.Stmt, args ...any) ([]Operation, error) {
 	objects := make([]Operation, 0)
@@ -446,27 +442,6 @@ func DeleteOperation(ctx context.Context, tx *sql.Tx, uuid string) error {
 		return api.StatusErrorf(http.StatusNotFound, "Operation not found")
 	} else if n > 1 {
 		return fmt.Errorf("Query deleted %d Operation rows instead of 1", n)
-	}
-
-	return nil
-}
-
-// DeleteOperations deletes the operation matching the given key parameters.
-// generator: operation DeleteMany-by-NodeID
-func DeleteOperations(ctx context.Context, tx *sql.Tx, nodeID int64) error {
-	stmt, err := Stmt(tx, operationDeleteByNodeID)
-	if err != nil {
-		return fmt.Errorf("Failed getting \"operationDeleteByNodeID\" prepared statement: %w", err)
-	}
-
-	result, err := stmt.ExecContext(ctx, nodeID)
-	if err != nil {
-		return fmt.Errorf("Delete \"operations\": %w", err)
-	}
-
-	_, err = result.RowsAffected()
-	if err != nil {
-		return fmt.Errorf("Fetch affected rows: %w", err)
 	}
 
 	return nil
--- a/lxd/db/cluster/operations.mapper.go
+++ b/lxd/db/cluster/operations.mapper.go
@@ -33,6 +33,22 @@ SELECT operations.id, operations.uuid, nodes.address AS node_address, nodes.name
   ORDER BY operations.id, operations.uuid
 `)
 
+var operationObjectsByNodeIDAndClass = RegisterStmt(`
+SELECT operations.id, operations.uuid, nodes.address AS node_address, nodes.name, operations.project_id, operations.node_id, operations.type, operations.requestor_protocol, operations.requestor_identity_id, operations.entity_id, operations.metadata, operations.class, operations.created_at, operations.updated_at, operations.inputs, operations.status_code, operations.conflict_reference, operations.error, operations.error_code, operations.parent, operations.stage
+  FROM operations
+  JOIN nodes ON operations.node_id = nodes.id
+  WHERE ( operations.node_id = ? AND operations.class = ? )
+  ORDER BY operations.id, operations.uuid
+`)
+
+var operationObjectsByClass = RegisterStmt(`
+SELECT operations.id, operations.uuid, nodes.address AS node_address, nodes.name, operations.project_id, operations.node_id, operations.type, operations.requestor_protocol, operations.requestor_identity_id, operations.entity_id, operations.metadata, operations.class, operations.created_at, operations.updated_at, operations.inputs, operations.status_code, operations.conflict_reference, operations.error, operations.error_code, operations.parent, operations.stage
+  FROM operations
+  JOIN nodes ON operations.node_id = nodes.id
+  WHERE ( operations.class = ? )
+  ORDER BY operations.id, operations.uuid
+`)
+
 var operationObjectsByID = RegisterStmt(`
 SELECT operations.id, operations.uuid, nodes.address AS node_address, nodes.name, operations.project_id, operations.node_id, operations.type, operations.requestor_protocol, operations.requestor_identity_id, operations.entity_id, operations.metadata, operations.class, operations.created_at, operations.updated_at, operations.inputs, operations.status_code, operations.conflict_reference, operations.error, operations.error_code, operations.parent, operations.stage
   FROM operations
@@ -144,7 +160,31 @@ func GetOperations(ctx context.Context, tx *sql.Tx, filters ...OperationFilter) 
 	}
 
 	for i, filter := range filters {
-		if filter.UUID != nil && filter.ID == nil && filter.NodeID == nil && filter.Parent == nil {
+		if filter.NodeID != nil && filter.Class != nil && filter.ID == nil && filter.UUID == nil && filter.Parent == nil {
+			args = append(args, []any{filter.NodeID, filter.Class}...)
+			if len(filters) == 1 {
+				sqlStmt, err = Stmt(tx, operationObjectsByNodeIDAndClass)
+				if err != nil {
+					return nil, fmt.Errorf("Failed getting \"operationObjectsByNodeIDAndClass\" prepared statement: %w", err)
+				}
+
+				break
+			}
+
+			query, err := StmtString(operationObjectsByNodeIDAndClass)
+			if err != nil {
+				return nil, fmt.Errorf("Failed getting \"operationObjects\" prepared statement: %w", err)
+			}
+
+			parts := strings.SplitN(query, "ORDER BY", 2)
+			if i == 0 {
+				copy(queryParts[:], parts)
+				continue
+			}
+
+			_, where, _ := strings.Cut(parts[0], "WHERE")
+			queryParts[0] += "OR" + where
+		} else if filter.UUID != nil && filter.ID == nil && filter.NodeID == nil && filter.Parent == nil && filter.Class == nil {
 			args = append(args, []any{filter.UUID}...)
 			if len(filters) == 1 {
 				sqlStmt, err = Stmt(tx, operationObjectsByUUID)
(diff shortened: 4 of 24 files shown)


lxd/operations.go
```diff
@@ -1299,6 +1427,47 @@ type operationWaitPost struct {
...
+			return fmt.Errorf("Failed to update operation metadata: %q", err)
+		}
+
+		err = op.CommitMetadata()
+		if err != nil {
+			return fmt.Errorf("Failed to commit operation metadata: %q", err)
+		}
+	}
+
+	logger.Warn("Wait handler operation completed")
```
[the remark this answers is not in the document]
skozina: Nope. This handler is only used for testing, the point of Warnf() is to see what's going on.

lxd/operations/linux.go
```diff
@@ -62,3 +143,90 @@ func (op *Operation) sendEvent(eventMessage any) {
...
+		if dbOp.ProjectID != nil {
+			var ok bool
+			projectName, ok = projects[*dbOp.ProjectID]
+			if !ok {
+				logger.Warn("Project ID not found in the map of projects", logger.Ctx{"projectID": *dbOp.ProjectID})
+				continue
+			}
+		}
+
+		// TODO Reconstruct the Requestor of the operation here.
```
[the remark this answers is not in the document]
skozina: Yes, we are tracking this.

lxd/operations.go
```diff
@@ -387,6 +387,110 @@ func operationCancel(ctx context.Context, s *state.State, projectName string, op
...
+			continue
+		}
+
+		// Omit operations that don't have a project if the caller does not have access.
+		if opProjectName == "" && !canViewServerOperations {
+			continue
+		}
+
+		// Omit operations if the caller does not have `can_view_operations` on the operations' project and the caller is not the operation owner.
+		// TODO once we reconstruct requestor on the operation, also return operations owned by the caller.
```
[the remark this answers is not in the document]
skozina: Yes, we are tracking this.

lxd/operations.go
```diff
@@ -1299,6 +1427,47 @@ type operationWaitPost struct {
...
+		logger.Warnf("Wait handler operation running for %d seconds...", elapsed/time.Second)
+		op.Metadata()["elapsed"] = elapsed.String()
+		err = op.UpdateMetadata(op.Metadata())
+		if err != nil {
+			return fmt.Errorf("Failed to update operation metadata: %q", err)
+		}
+
+		err = op.CommitMetadata()
+		if err != nil {
+			return fmt.Errorf("Failed to commit operation metadata: %q", err)
```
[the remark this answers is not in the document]
skozina: We use %q for error formatting.

lxd/operations.go
```diff
@@ -387,6 +387,110 @@ func operationCancel(ctx context.Context, s *state.State, projectName string, op
...
+		if !canViewProjectOperations(entity.ProjectURL(opProjectName)) {
+			continue
+		}
+
+		op := api.Operation{}
+		op.ID = dbOp.UUID
+		op.Class = (operations.OperationClass)(dbOp.Class).String()
+		op.Description = dbOp.Description
+		op.CreatedAt = dbOp.CreatedAt
+		op.Status = api.Frozen.String() // As we are loading the operation from the DB, it's likely not running right now.
```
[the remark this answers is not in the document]
skozina: I don't think so. We know that this operation is not currently running on the node, or at least the node is not responsive. We could update the operation status in the DB first (once the operation is restarted, we'd set it again to Running), but that would mean we'd be writing to the database as part of the `operationsGet()` which would be confusing. I'd leave this as it is.

lxd/operations/linux.go
```diff
@@ -62,3 +143,90 @@ func (op *Operation) sendEvent(eventMessage any) {
...
+			}
+
+			resources[dbOp.ID], err = cluster.GetDurableOperationResources(ctx, tx.Tx(), dbOp.ID)
+			if err != nil {
+				return fmt.Errorf("Failed to load durable operation resources for operation %d: %w", dbOp.ID, err)
+			}
+
+			if dbOp.IdentityID != nil {
+				identityFilter := cluster.IdentityFilter{ID: dbOp.IdentityID}
+				clusterIdentities, err := cluster.GetIdentitys(ctx, tx.Tx(), identityFilter)
```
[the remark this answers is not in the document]
skozina: This is generated function, not worth fixing the generated for proper English naming.

lxd/operations/operations.go
```diff
@@ -310,19 +329,28 @@ func (op *Operation) Start() error {
 		return errors.New("Only pending operations can be started")
 	}
 
-	op.status = api.Running
+	err := op.updateStatus(api.Running)
+	if err != nil {
+		op.lock.Unlock()
+		return fmt.Errorf("failed to update Operation %s status: %w", op.id, err)
+	}
```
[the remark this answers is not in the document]
skozina: This doesn't seem right. The last sentence seems plain wrong:
> the function returns at line 335 without ever releasing the lock acquired at line 326.

As the line before 335 looks like:
```
332		err := op.updateStatus(api.Running)
333		if err != nil {
334			op.lock.Unlock()
335			return fmt.Errorf("failed to update Operation %s status: %w", op.id, err)
336		}
```

doc/api-extensions.md
```diff
@@ -2884,3 +2884,8 @@ Before this was only partially possible for some of the drivers (e.g. by using `
 The new pool `source.recover` configuration key can be set per cluster member to allow reuse of an existing pool `source`.
 What it does not allow is reusing the same source for multiple storage pools.
 The LVM storage driver has the specific `lvm.vg.force_reuse` configuration key for this purpose.
+
+(extension-durable-workflows)=
+## `durable_workflows`
```
kadinsayani: Perhaps we can just title the extension `durable_operations` for consistency?

lxd/db/cluster/operations.go
```diff
@@ -29,17 +38,192 @@ import (
...
+			return nil, err
+		}
+
+		values[key] = value
+	}
+
+	err = rows.Err()
+	if err != nil {
+		return nil, err
+	}
```
kadinsayani: We can use `github.com/canonical/lxd/lxd/db/query.Scan` for this.

lxd/db/cluster/projects.go
```diff
@@ -156,6 +156,39 @@ func GetProjectNames(ctx context.Context, tx *sql.Tx) ([]string, error) {
...
+			return nil, err
+		}
+
+		result[id] = name
+	}
+
+	err = rows.Err()
+	if err != nil {
+		return nil, err
+	}
```
kadinsayani, line 278: `query.Scan()`

lxd/db/operationtype/operation_type.go
```diff
@@ -211,6 +212,8 @@ func (t Type) Description() string {
 		return "Healing cluster"
 	case RemoveExpiredOIDCSessions:
 		return "Remove expired OIDC sessions"
+	case Wait:
+		return "Just chilling"
```
kadinsayani, line 318: Nice 😎

lxd/operations/linux.go
```diff
@@ -33,6 +36,19 @@ func registerDBOperation(op *Operation, opType operationtype.Type) error {
 			opInfo.ProjectID = &projectID
 		}
 
+		if op.requestor != nil {
+			opInfo.RequestorProtocol = op.requestor.CallerProtocol()
+			callerIdentity := op.requestor.CallerIdentity()
+			if callerIdentity != nil {
+				identityID, err := cluster.GetIdentityID(ctx, tx.Tx(), cluster.AuthMethod(callerIdentity.AuthenticationMethod), callerIdentity.Identifier)
+				if err != nil {
+					return fmt.Errorf("Failed to add %q Operation %s to database: %w", opType.Description(), op.id, err)
```
kadinsayani: For non-generated error messages, can we please use gerund style? For example: `Failed adding...` as opposed to `Failed to add...`.

shared/api/operation.go
```diff
@@ -15,6 +15,9 @@ const OperationClassWebsocket = "websocket"
 // OperationClassToken represents the Token OperationClass.
 const OperationClassToken = "token"
 
+// OperationClassDurable represents the Durable OperationClass.
```
kadinsayani: While this godoc comment is aligned with the comments for already existing operation class constants, it just states what we already know from the constant's name and value. Perhaps we could add a bit more detail (even just a single sentence) for each constant to distinguish between the different operation classes?

lxd/operations/operations.go
```diff
@@ -128,7 +137,12 @@ func OperationCreate(ctx context.Context, s *state.State, projectName string, op
 	// Main attributes
 	op := Operation{}
 	op.projectName = projectName
-	op.id = uuid.New().String()
+	if opUUID != "" {
+		op.id = opUUID
+	} else {
+		op.id = uuid.New().String()
+	}
```
kadinsayani:
```suggestion
    op.id = opUUID
    if op.id == "" {
        op.id = uuid.New().String()
    }
```

lxd/operations/operations.go
```diff
@@ -185,7 +199,12 @@ func OperationCreate(ctx context.Context, s *state.State, projectName string, op
 	operations[op.id] = &op
 	operationsLock.Unlock()
 
-	err = registerDBOperation(&op, opType)
+	if opUUID == "" {
+		err = registerDBOperation(&op, opType)
+	} else {
+		err = updateDBOperationNodeID(&op)
+	}
+
```
kadinsayani:
```suggestion
    if opUUID == "" {
        // New operation: register in the DB.
	err = registerDBOperation(&op, opType)
    } else {
        // Existing operation: update the node ID in the DB.
	err = updateDBOperationNodeID(&op)
    }
```

lxd/operations/operations.go
```diff
@@ -310,19 +329,28 @@ func (op *Operation) Start() error {
...
 			err := op.onRun(op)
 			if err != nil {
+				op.logger.Debug("Failure for operation", logger.Ctx{"err": err})
+
 				op.lock.Lock()
-				op.status = api.Failure
 				op.err = err
+				err = op.updateStatus(api.Failure)
+				if err != nil {
+					op.logger.Warn("Failed to update operation status to failure", logger.Ctx{"err": err})
```
kadinsayani: Can we include more info in the context of these log messages?


```suggestion
					op.logger.Warn("Failed to update operation status", logger.Ctx{"err": err, "oldStatus": op.status.String(), "newStatus": api.Failure.String()})
```

lxd/operations/operations.go
```diff
@@ -333,7 +361,11 @@ func (op *Operation) Start() error {
 			}
 
 			op.lock.Lock()
-			op.status = api.Success
+			err = op.updateStatus(api.Success)
+			if err != nil {
+				op.logger.Warn("Failed to update operation status to success", logger.Ctx{"err": err})
```
kadinsayani: Same note on adding more context to these log messages. Since we update the status and construct these warning messages multiple times, perhaps we can extract a helper function:



```Go
func (op *Operation) updateStatusWithWarning(status api.Status) {
	err := op.updateStatus(status)
    if err != nil {
		op.logger.Warn("Failed to update operation status", logger.Ctx{
			"err":    err,
            "oldStatus": op.status.String(),
			"newStatus": status.String(),
		})
	}
}
```

Or, we can expand `updateStatus()` to handle any `error` returned by `updateDBOperationStatus()` and log warnings.

lxd/operations/operations.go
```diff
@@ -719,3 +789,26 @@ func (op *Operation) Class() OperationClass {
...
+var durableOperations map[operationtype.Type]DurableOperationHandlers
+
+// InitDurableOperations initializes the durable operations table.
+// As durable operations can be restarted on other nodes, the durable operation handlers cannot be defined only in the memory of the node.
+// Therefore we maintain a static map of durable operation handlers based on operation type.
+// As this map contains handlers from across many packages, the table itself is defined in the main package.
+// Because this table needs to be accessible from the operations package, we provide this Init function to set it.
+func InitDurableOperations(opTable map[operationtype.Type]DurableOperationHandlers) {
+	durableOperations = opTable
+}
```
kadinsayani:
```suggestion
type DurableOperationTable map[operationtype.Type]DurableOperationHandlers

var (
    durableOperations DurableOperationTable
    durableOperationsOnce sync.Once
)

// InitDurableOperations initializes the durable operations table.
// As durable operations can be restarted on other nodes, the durable operation handlers cannot be defined only in the memory of the node.
// Therefore we maintain a static map of durable operation handlers based on operation type.
// As this map contains handlers from across many packages, the table itself is defined in the main package.
// Because this table needs to be accessible from the operations package, we provide this Init function to set it.
func InitDurableOperations(opTable DurableOperationTable) {
    durableOperationsOnce.Do(func() {
        durableOperations = opTable
    })
}
```

This ensures `InitDurableOperations` can only set the map once, preventing accidental re-initialization and race conditions.

lxd/operations/operations.go
```diff
@@ -719,3 +789,26 @@ func (op *Operation) Class() OperationClass {
...
+
+// CreateDurableOperation creates a new durable operation and returns it.
+// Durable operations need to have handlers defined in a durableOperations map rather than provided via arguments.
+// Arguments to these handlers are then looked up in the metadata.
+func CreateDurableOperation(ctx context.Context, s *state.State, opUUID string, projectName string, opType operationtype.Type, opResources map[string][]api.URL, opMetadata map[string]string) (*Operation, error) {
+	handlers, ok := durableOperations[opType]
```
kadinsayani: Do we need a synchronization mechanism to prevent concurrent access to `durableOperations` here?

lxd/operations.go
```diff
@@ -1353,10 +1520,22 @@ func operationWaitHandler(d *Daemon, r *http.Request) response.Response {
...
+	}
+
 	if err != nil {
 		return response.InternalError(err)
 	}
 
 	return operations.OperationResponse(op)
 }
+
+var DurableOperations = map[operationtype.Type]operations.DurableOperationHandlers{
```
kadinsayani: We can use the proposed `DurableOperationTable` type here too.

kadinsayani (changes requested):
Looking really good so far! I've made an initial pass through and have a few questions, comments and suggestions.

lxd/operations/operations.go
kadinsayani: Ah, I just looked at the call site of `InitDurableOperations()`. Since `d.init()` is thread safe, we're all good! Keeping this unresolved as the type suggestion is still worth considering (`type DurableOperationTable map[operationtype.Type]DurableOperationHandlers`).

lxd/operations/operations.go
kadinsayani: Just a nit for readability :)

lxd/operations/operations.go
kadinsayani: Again, just a nit. The comments improve readability.

doc/api-extensions.md
skozina: I think you're right, we should align on naming. Lets stick with "durable operations" everywhere as that's more native term to LXD. Thanks!

lxd/db/cluster/operations.go
skozina: Thanks! Switched to `query.Scan()`.

lxd/db/cluster/projects.go
skozina: Thanks! Switched to `query.Scan()`.

lxd/operations/linux.go
skozina: Sure! I've updated all error and warning messages to use gerund style. Thanks for the reminder!;-)

shared/api/operation.go
skozina: Good idea! I've added a sentence to each. Appreciate if you could check for correctness;-)

lxd/operations/operations.go
skozina: Good point;-) Thanks, I've done that.

lxd/operations/operations.go
skozina: Thanks! Comments added.

lxd/operations/operations.go
skozina: Thanks! I've done this. I didn't want to directly log the warning from the `updateStatus()`, but adding another wrapper function works great. It also allows to simply replace the setting of the status with this function call without any further changes.

lxd/operations/operations.go
skozina: Thanks for the tip! I've done this change in full. I think including the `sync.Once()` will make it clear that this table is really not supposed to change over the lifetime of the daemon.

lxd/operations/operations.go
skozina: No... this is really just a static table which is set only once in `d.Init()`. Especially with the `sync.Once()` protection I think we don't need any further locking.

lxd/operations.go
skozina: Yup, done (just in a later commit when we're populating the table). Thanks!

lxd/operations/operations.go
kadinsayani: Ack, thanks for clarifying.

shared/api/operation.go
kadinsayani: Awesome thanks, lgtm!

lxd/operations/operations.go
kadinsayani: I'm not sure if the use of `sync.Once()` is strictly required here since `daemon.init()` is already thread safe, however using `sync.Once()` does make the code self-documenting by indicating it should only be created once. Will `InitDurableOperations()` be called anywhere else?

lxd/db/cluster/schema.go
```diff
@@ -442,13 +442,38 @@ CREATE TABLE oidc_sessions (
 CREATE TABLE "operations" (
     id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
     uuid TEXT NOT NULL,
-    node_id TEXT NOT NULL,
+    node_id INTEGER NOT NULL,
     type INTEGER NOT NULL DEFAULT 0,
     project_id INTEGER,
+    description TEXT,
+    requestor_protocol TEXT,
+    identity_id INTEGER,
```
markylaing: I wonder if we should make this `requestor_identity_id` to match the `requestor_protocol` column. Not sure if the additional clarity is required or whether it is too verbose. Up to you :)

lxd/db/cluster/schema.go
```diff
@@ -442,13 +442,38 @@ CREATE TABLE oidc_sessions (
...
+	operation_id INTEGER NOT NULL,
+	key TEXT NOT NULL,
+	value TEXT NOT NULL,
+	FOREIGN KEY (operation_id) REFERENCES operations (id) ON DELETE CASCADE,
+	UNIQUE (operation_id, key, value)
+);
+CREATE TABLE operations_resources (
+	id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+	operation_id INTEGER NOT NULL,
+	resource TEXT NOT NULL,
```
markylaing: What does `resource` represent here? Don't we already know the "resource" from the entity type and ID?

lxd/db/cluster/schema.go
skozina: Yes, `requestor_identity_id` is a better name. I'll leave this open until I get to some more database schema changes. Thanks for the tip!

lxd/db/cluster/schema.go
skozina: The operation resources are defined as `map[string][]api.URL`. The `resource` in the db is the key of the map.
Some operations use multiple different keys, eg.:
https://github.com/canonical/lxd/blob/08d70f8ed6365bfd5fc6cc99d5397f98d375893a/lxd/storage_volumes_snapshot.go#L227-L228

I agree that this might not be necessary, a simple list (rather than a map of lists) of resources might be sufficient. The db schema is designed to be able to capture our current structure of resources.

lxd/db/cluster/operations.go
```diff
@@ -29,12 +31,18 @@ import (
...
 // in the cluster.
 type Operation struct {
-	ID          int64              `db:"primary=yes"`                               // Stable database identifier
-	UUID        string             `db:"primary=yes"`                               // User-visible identifier
-	NodeAddress string             `db:"join=nodes.address&omit=create-or-replace"` // Address of the node the operation is running on
-	ProjectID   *int64             // ID of the project for the operation.
-	NodeID      int64              // ID of the node the operation is running on
-	Type        operationtype.Type // Type of the operation
+	ID                int64              `db:"primary=yes"`                               // Stable database identifier
+	UUID              string             `db:"primary=yes"`                               // User-visible identifier
```
markylaing: Do you think that, if a durable operation can use the UUID column to effectively achieve cluster-wide locking, it would be worth changing the column name?

I'm just thinking aloud. We could have a lock operation for the download of a particular image. In this case the UUID column would be an image fingerprint?

tomponline: Yes, suggest "reference" (i.e not an ID, but a unique reference for the operation).

lxd/operations.go
```diff
@@ -1353,10 +1520,22 @@ func operationWaitHandler(d *Daemon, r *http.Request) response.Response {
...
+	}
+
 	if err != nil {
 		return response.InternalError(err)
 	}
 
 	return operations.OperationResponse(op)
 }
+
+var DurableOperations = operations.DurableOperationTable{
```
markylaing: Can we move this closer to the top of the file? Just a preference but I tend to find declarations easier to find that way

lxd/operations/operations.go
```diff
@@ -719,3 +776,34 @@ func (op *Operation) Class() OperationClass {
...
+// InitDurableOperations initializes the durable operations table.
+// As durable operations can be restarted on other nodes, the durable operation handlers cannot be defined only in the memory of the node.
+// Therefore we maintain a static map of durable operation handlers based on operation type.
+// As this map contains handlers from across many packages, the table itself is defined in the main package.
+// Because this table needs to be accessible from the operations package, we provide this Init function to set it.
+func InitDurableOperations(opTable DurableOperationTable) {
+	durableOperationsOnce.Do(func() {
+		durableOperations = opTable
+	})
+}
```
markylaing: Can this also be moved to the top of the file, or moved to a second file (`durable.go`?)

lxd/operations/linux.go
```diff
@@ -43,6 +79,51 @@ func registerDBOperation(op *Operation, opType operationtype.Type) error {
 	return nil
 }
 
+func updateDBOperationNodeID(op *Operation) error {
```
markylaing: I'm not a huge fan of how we're using build tags to define how operations interact with the database.

I'd like to see something like an interface defined in the operations package. When creating an operation, we can pass an implementation of that interface which does that actual persistence to the DB.

tomponline: there should be no need for build tags to be needed if we can untangle server-side operations from lxd-agent operations.

tomponline: Also we have tended to avoid putting functions that run queries directly into the associated package and prefer to put them into the `db/cluster` package.

lxd/db/cluster/operations.go
skozina: Sounds good! I'll also rename the database column from `uuid` to `reference`.

lxd/db/cluster/schema.go
markylaing: In that case I'm confused. We have:
```
CREATE TABLE operations_resources (
	id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
	operation_id INTEGER NOT NULL,
	resource TEXT NOT NULL,
	entity_type_code INTEGER NOT NULL,
	entity_id INTEGER NOT NULL,
	FOREIGN KEY (operation_id) REFERENCES operations (id) ON DELETE CASCADE,
	UNIQUE (operation_id, resource)
);
```
Is "resource" the key in the map? E.g. `storage_volumes`?

If so, I wonder if we can reduce some redundancy by having a standardised way to pluralise an entity type name.

tomponline: We likely wont have this table in the end result so might be worth holding on further review of this.

lxd/operations.go
skozina: Sure! I'll move this right under the APIEndpoint{} definitions.

lxd/operations/operations.go
skozina: Sure! I've moved this right after the Operation struct definition.

lxd/operations/linux.go
skozina: I am confused, can you help me to understand the comments?
With the current state of arts, both lxd-agent and lxd call `OperationCreate()`, and we need to distinguish the two. That is done with build tags (I don't like it either), and I don't see other way to do it without isolating lxd-agent operations.
These functions are all just wrappers for the build tags, the DB queries are run by `cluster.UpdateOperationNodeID()` function in the `db/cluster` package.

Is there something you'd like me to change?

lxd/db/cluster/schema.go
skozina: > Is "resource" the key in the map? E.g. storage_volumes?

Yes.

lxd/operations/linux.go
tomponline: > Is there something you'd like me to change?

Probably not right now. 

But untangling lxd-agent operations from lxd operations so that the latter can access the DB packages would be nice.

Whether thats done via an injected persistence dependency as @markylaing suggest, or whether we go for a similar approach we've used in the wider LXD of having an interface that defines a server-side Operation and then two implementations of said interface (one for LXD that depends on DB packages, and one for lxd-agent that does not) remains to be seen. Although my preference would be the latter approach as its more in keeping with how its been done in the wider LXD (there could still be a `common` embedded implementation that does not depend on the DB packages but provides the common functionality between both).

skozina: I see, thank you!
For now I've opened #17268 to track this. Thanks!

lxd/db/cluster/schema.go
skozina: I've dropped the `operations_resources` table from the PR. Instead, we only allow up to a single resource per  durable operation, and we ensure this resource, if provided, has the same entity_type as the required permissions defined for the operation type. Then we only store entity_id of this single resource.

lxd/operations/linux.go
```diff
@@ -78,3 +83,43 @@ func (op *Operation) sendEvent(eventMessage any) {
...
+	op.state = s
+	op.requestor = nil
+	op.logger = logger.AddContext(logger.Ctx{"operation": op.id, "project": op.projectName, "class": op.class.String(), "description": op.description})
+
+	if s != nil {
+		op.SetEventServer(s.Events)
+	}
+
+	op.resources = nil
+	op.metadata = nil
```
kadinsayani: This can be cleaned up as follows:

```Go
op := Operation{
    projectName: projectName,
    ...
}

File: lxd/operations/linux.go [hunk 28]
```diff
@@ -55,10 +166,225 @@ func removeDBOperation(op *Operation) error {
...
+		if err != nil {
+			return fmt.Errorf("Failed loading project IDs to names: %w", err)
+		}
+
+		// See if there are any durable operations running on this node which need to be restarted.
+		durableClass := (int64)(OperationClassDurable)
+		filter := cluster.OperationFilter{NodeID: &nodeID, Class: &durableClass}
+		dbOps, err := cluster.GetOperations(ctx, tx.Tx(), filter)
+		if err != nil {
+			return fmt.Errorf("Failed loading durable operations for the node %d: %w", nodeID, err)
```
kadinsayani:
```suggestion
			return fmt.Errorf("Failed loading durable operations for node with ID %q: %w", nodeID, err)
```

lxd/operations/linux.go
```diff
@@ -55,10 +166,225 @@ func removeDBOperation(op *Operation) error {
...
+
+		for _, dbOp := range dbOps {
+			var projectName string
+
+			// Load the project name if provided.
+			if dbOp.ProjectID != nil {
+				var ok bool
+				projectName, ok = projects[*dbOp.ProjectID]
+				if !ok {
+					logger.Warn("Project ID not found in the map of projects", logger.Ctx{"projectID": *dbOp.ProjectID})
```
kadinsayani:
```suggestion
					logger.Warn("Project ID %q not found in the map of projects", logger.Ctx{"projectID": *dbOp.ProjectID})
```

lxd/operations/notlinux.go
```diff
@@ -17,9 +17,41 @@ func registerDBOperation(op *Operation, opType operationtype.Type) error {
...
-		return fmt.Errorf("registerDBOperation not supported on this platform")
+		return fmt.Errorf("removeDBOperation not supported on this platform")
+	}
+
+	return nil
+}
+
+func removeDBOperationMetadata(op *Operation) error {
+	if op.state != nil {
+		return fmt.Errorf("removeOperationMetadata not supported on this platform")
```
kadinsayani:
```suggestion
		return fmt.Errorf("removeDBOperationMetadata not supported on this platform")
```

lxd/operations/operations.go
```diff
@@ -278,6 +342,27 @@ func (op *Operation) done() {
 	op.finished.Cancel()
 	op.lock.Unlock()
 
+	// For durable operations, we just remove it from the internal map and clear metadata from the db.
+	// The database record will be cleared later by a dedicated task.
```
kadinsayani:
```suggestion
	// The database record will be cleared later by pruneExpiredDurableOperationsTask (hourly).
```

lxd/operations/operations.go
```diff
@@ -322,7 +419,11 @@ func (op *Operation) Start() error {
 		return errors.New("Only pending operations can be started")
 	}
 
-	op.status = api.Running
+	err := op.updateStatus(api.Running)
+	if err != nil {
+		op.lock.Unlock()
+		return fmt.Errorf("Failed updating Operation %s status: %w", op.id, err)
```
kadinsayani:
```suggestion
		return fmt.Errorf("Failed updating Operation with ID %q status: %w", op.id, err)
```

lxd/operations.go
```diff
@@ -158,50 +165,97 @@ func operationGet(d *Daemon, r *http.Request) response.Response {
...
+				return err
+			}
 
-		_, body, err = op.Render()
+			if len(ops) < 1 {
+				return api.StatusErrorf(http.StatusNotFound, "Operation not found")
+			}
+
+			if len(ops) > 1 {
+				return errors.New("More than one operation matches")
```
kadinsayani: When is this possible?

lxd/operations.go
```diff
@@ -158,50 +165,97 @@ func operationGet(d *Daemon, r *http.Request) response.Response {
...
-		if len(ops) < 1 {
-			return api.StatusErrorf(http.StatusNotFound, "Operation not found")
+		_, body, err = op.Render()
+		if err != nil {
+			return response.SmartError(err)
 		}
 
-		if len(ops) > 1 {
-			return errors.New("More than one operation matches")
+		// The operations.Operation doesn't contain the node where the operation is running.
```
kadinsayani:
```suggestion
		// The [operations.Operation] doesn't contain the node where the operation is running.
```

godoc links are handy :)

lxd/operations.go
```diff
@@ -314,16 +368,23 @@ func operationDelete(d *Daemon, r *http.Request) response.Response {
...
 	}
 
-	client, err := cluster.Connect(r.Context(), address, s.Endpoints.NetworkCert(), s.ServerCert(), false)
+	// Don't forward to ourselves.
+	if operation.NodeAddress == s.LocalConfig.ClusterAddress() {
+		if operation.Class == int64(operations.OperationClassDurable) && api.StatusCode(operation.Status).IsFinal() {
+			return response.BadRequest(errors.New("Durable operation is already finalized"))
+		}
+
+		return response.SmartError(fmt.Errorf("Operation %q is not running on this node", id))
```
kadinsayani:
```suggestion
		return response.SmartError(fmt.Errorf("Operation with ID%q is not running on this node", id))
```

lxd/operations.go
```diff
@@ -527,6 +599,83 @@ func operationsGet(d *Daemon, r *http.Request) response.Response {
...
+				// Omit operations if the caller does not have `can_view_operations` on the operations' project and the caller is not the operation owner.
+				// TODO once we reconstruct requestor on the operation, also return operations owned by the caller.
+				if !canViewProjectOperations(entity.ProjectURL(projectName)) {
+					continue
+				}
+
+				// Durable operation, convert to API representation now.
+				op, err := operations.NewDurableOperation(ctx, tx.Tx(), s, &dbOp, projectName)
+				if err != nil {
+					return fmt.Errorf("Failed to load durable operation %q: %w", dbOp.Reference, err)
```
kadinsayani:
```suggestion
					return fmt.Errorf("Failed loading durable operation with reference %q: %w", dbOp.Reference, err)
```

lxd/operations.go
```diff
@@ -641,64 +798,56 @@ func operationsGet(d *Daemon, r *http.Request) response.Response {
...
-	}
-
-	// Get all nodes with running operations in this project.
-	var membersWithOps []string
-	var members []db.NodeInfo
-	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
-		var err error
+	// Merge with the durable operations data.
+	for _, o := range durableOps {
+		op := o // Local var for pointer.
```
kadinsayani: Why is this needed? Are we modifying for local use or just accessing fields?

kadinsayani (commented):
Just a few nits, suggestions, and a couple questions :)

This is really coming together!

lxd/operations.go
kadinsayani: Perhaps a comment explaining? :)

lxd/operations/linux.go
skozina: I was little hesitant to use the struct initialization, as we are unable to set all items in the initialization itself (the `operation_type.Permission()` returns multiple values, resources and metadata depends on further function calls). But I've set at least those fields which can be set. Thanks!

lxd/operations/linux.go
skozina: Thanks, I've changed the text, but left `%d` as the formatting character for the `int nodeID`.

lxd/operations/linux.go
skozina: I'm confused -- are you suggesting it would be better to inline the project ID rather than using the logger context?

lxd/operations/notlinux.go
skozina: Fixed, thanks!

lxd/operations/operations.go
skozina: Thanks for the tip! I've put the name of the task function here, but not the `(hourly)`, as if we need to change that later, I'm pretty sure we'd forget to update the comment here.

lxd/operations/operations.go
skozina: Thanks!

lxd/operations.go
skozina: Hopefully never possible;-) 
I've added a comment and turned this into a switch like we do in https://github.com/canonical/lxd/blob/ad5334629f8fe097d9a340afbceb7afacf324654/lxd/db/node.go#L240-L247
Thanks!

lxd/operations.go
skozina: Thanks, added;-)

lxd/operations.go
skozina: Thanks, added;-)

lxd/operations.go
skozina: Thanks, fixed! The `ID` is the API external field which we internally call `Reference` (because ID is the database ID). So in this case I'm going to say `operation ID`, in line with other messages.

lxd/operations.go
skozina: I've copied this from existing code;-) But I think this explains it nicely:
https://stackoverflow.com/questions/38692998/strange-golang-append-behavior-overwriting-values-in-slice
```
The for ... range makes a copy of the elements it loops over, and you append the address of this temporary, loop variable - which is the same in all iterations. 
```

Without the copy of the loop variable, we'd (in the case of `recursion`) return slice of pointers all pointing to the same variable.

lxd/operations/linux.go
kadinsayani: Let's keep the logger context. My apologies, I scanned over that.

lxd/operations/operations.go
kadinsayani: Sounds good, thanks!

lxd/operations.go
kadinsayani: Awesome, thanks!

lxd/operations.go
kadinsayani: I believe this behavior changed in go 1.22: https://go.dev/ref/spec#For_range. `for range` loop vars are now per-iteration, so taking `&o` no longer aliases the same variable each time.

lxd/operations.go
kadinsayani: Sounds good to me, thanks!

lxd/operations.go
tomponline: yeah we dont need this anymore

kadinsayani: https://github.com/canonical/lxd/pull/17376/changes/5a47a513ee5db329a74885bc2feb2be1a61371fe

lxd/operations/linux.go
skozina: Cool, thanks;-)

lxd/operations.go
skozina: Thank you for those references! And thanks also for cleaning up the existing code;-) I've dropped the unnecessary local variable. Thanks!

lxd/db/cluster/operations.go
```diff
@@ -65,9 +66,10 @@ type Operation struct {
 
 // OperationFilter specifies potential query parameter fields.
 type OperationFilter struct {
-	ID     *int64
-	NodeID *int64
-	UUID   *string
+	ID                *int64
+	NodeID            *int64
+	UUID              *string
+	ConflictReference *string
```
markylaing: Why do we need conflict reference logic for the durable workflows?

lxd/operations/linux.go
```diff
@@ -9,15 +9,28 @@ import (
 
 	"github.com/canonical/lxd/lxd/db"
 	"github.com/canonical/lxd/lxd/db/cluster"
+	"github.com/canonical/lxd/lxd/db/operationtype"
 	"github.com/canonical/lxd/shared/api"
 )
 
-func registerDBOperation(op *Operation) error {
+func registerDBOperation(op *Operation, conflictReference string) error {
```
markylaing: Do you need to change the signature here? The conflict reference is a field on Operation isn't it?

lxd/operations/linux.go
```diff
@@ -9,15 +9,28 @@ import (
...
 	if op.state == nil {
 		return nil
 	}
 
 	err := op.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+		// If a conflict reference is provided, check if any conflicting operation is already running before creating the new operation record.
```
markylaing: We have a conditional unique index in the database so we don't need to check first. Just create it and check for a conflict via `query.IsConflictErr`

lxd/db/operationtype/operation_type.go
```diff
@@ -335,3 +335,18 @@ func (t Type) EntityType() entity.Type {
...
+const (
+	// ConflictActionNone means operation has no conflicts, all operations of this type can run concurrently.
+	ConflictActionNone ConflictAction = iota
+	// ConflictActionFail asks to resolve conflicts by failing to create a new operation if a conflicting operation is already running.
+	ConflictActionFail
+)
+
+// ConflictAction returns the action to take if a conflicting operation is already running.
+func (t Type) ConflictAction() ConflictAction {
+	return ConflictActionNone
```
markylaing, line 542: I don't know why we need a `None` action? The action is implicitly `None` if there is no conflict reference?

lxd/operations/operations.go
```diff
@@ -275,11 +280,36 @@ func initOperation(s *state.State, args OperationArgs) (*Operation, error) {
...
 // scheduleOperation schedules a new operation and returns it. If it cannot be created, it returns an error.
 func scheduleOperation(s *state.State, args OperationArgs) (*Operation, error) {
+	// If this is a bulk operation, don't allow more than one level of nesting.
+	if args.Children != nil {
+		for _, child := range args.Children {
+			if child.Children != nil {
+				return nil, errors.New("Bulk operations cannot have nested bulk operations")
+			}
+		}
+	}
```
markylaing, line 328: Seems like this logic should go into `initOperation`

lxd/operations/operations.go
```diff
@@ -425,6 +474,15 @@ func (op *Operation) Start() error {
...
 	}
 
+	// Start child operations
+	for _, childOp := range op.children {
+		err := childOp.Start()
+		if err != nil {
+			op.lock.Unlock()
+			return fmt.Errorf("Failed to start child operation: %w", err)
+		}
+	}
```
markylaing, line 482: Is this where the "stage" logic will live? I don't understand how we can start all of the children at once. Thinking back to the replicator jobs, we can't snapshot an instance and move that snapshot simultaneously?

doc/api-extensions.md
```diff
@@ -2968,3 +2968,8 @@ As with the {ref}`storage and profile operation extension <extension-storage-and
 ## `gpu_cdi_amd`
 
 Adds support for using the Container Device Interface (CDI) specification to configure AMD GPU passthrough in LXD containers. The `id` field of GPU devices now accepts CDI identifiers (for example, `amd.com/gpu=gpu{INDEX}`) for containers, in addition to DRM card IDs.
+
+(extension-durable-operations)=
+## `durable_operations`
+
+Introduces new operation class for durable operations. Durable operations are guaranteed to run on a different node if the node running the operation previously goes offline.
```
markylaing: This needs to be the first commit of the PR please

lxd/operations/operations.go
```diff
@@ -229,6 +229,25 @@ func initOperation(s *state.State, args OperationArgs) (*Operation, error) {
 		return nil, errors.New("LXD is shutting down")
 	}
 
+	// Quick check.
+	if args.Class == OperationClassDurable {
+		if args.RunHook != nil || args.ConnectHook != nil {
+			return nil, errors.New("Durable operations cannot have Run or Connect hooks provided directly")
```
markylaing: Nit but can we make this: `Durable operation run and connect hooks are statically defined`

lxd/db/cluster/operations.mapper.go
```diff
@@ -49,6 +49,14 @@ SELECT operations.id, operations.uuid, nodes.address AS node_address, operations
...
 `)
 
+var operationObjectsByClass = RegisterStmt(`
+SELECT operations.id, operations.uuid, nodes.address AS node_address, operations.project_id, operations.node_id, operations.type, operations.requestor_protocol, operations.requestor_identity_id, operations.entity_id, operations.metadata, operations.class, operations.created_at, operations.updated_at, operations.inputs, operations.status_code, operations.conflict_reference, operations.error, operations.parent, operations.stage
+  FROM operations
+  JOIN nodes ON operations.node_id = nodes.id
+  WHERE ( operations.class = ? )
+  ORDER BY operations.id, operations.uuid
+`)
+
```
markylaing: Can we just have one commit adding all the directives you need?

lxd/operations.go
```diff
@@ -512,6 +523,81 @@ func operationsGet(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(fmt.Errorf("Failed to check caller access to server operations: %w", err))
 	}
 
+	// We'll start by loading all operations from the database and sorting them out to durable operations and other classes.
+	// Durable operations are directly loaded from the database and converted into their API representation.
+	// For all the other operations we'll compile a list of nodes running these and get them directly from those nodes via cluster notifications.
+	var durableOps []api.Operation
+	membersWithOps := make(map[string]struct{})
+	var members []db.NodeInfo
+	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
```
markylaing, line 515: I think we said we would load all operations from the database now that we're storing all the metadata? (except for progress updates)

markylaing: If so, can we perform this work in a separate PR?

lxd/operations.go
```diff
@@ -1488,6 +1488,74 @@ func pruneExpiredDurableOperationsTask(stateFunc func() *state.State) (task.Func
...
+			if op.Class() != operations.OperationClassDurable {
+				continue
+			}
+
+			localOpsMap[op.ID()] = op
+		}
+
+		// Ensure that all durable operations in the database are running locally.
+		for _, op := range dbOps {
+			if op.Status().IsFinal() {
```
markylaing: Can we please always use `op.IsRunning`? This checks the run context directly and cannot race with subsequent state updates.

lxd/operations.go
```diff
@@ -1488,6 +1488,74 @@ func pruneExpiredDurableOperationsTask(stateFunc func() *state.State) (task.Func
 	return f, task.Hourly()
 }
 
+func syncDurableOperationsTask(stateFunc func() *state.State) (task.Func, task.Schedule) {
```
markylaing: Can this function please have a comment that describes the goals of the task - why does it exist? In which situations are local durable operations not already in sync?

lxd/db/cluster/operations.go
skozina: The durable-workflows branch contains all commits in order in which these will land in several independent PRs, at least:
 - schema changes (merged)
 - conflict reference (merged)
 - operation(s)Get returning operations constructed from DB
 - bulk operations
 - durable operations
The ConflictReference is already merged and thus is not in this PR any more.

lxd/operations/linux.go
skozina: Yes, the conflict reference is now a field on Operation, and the signature is not changing here.

lxd/operations/linux.go
skozina: Yup, this is done now:
https://github.com/skozina/lxd/blob/4753c75eb1d68cee92bd578b45a04c486915dd12/lxd/operations/linux.go#L124

lxd/db/operationtype/operation_type.go
skozina: Yes. We discussed this in the conflict reference PR:
https://github.com/canonical/lxd/pull/17721#discussion_r2840227020

lxd/operations/operations.go
skozina: Hm... that means we'll repeat the check for parent and all its child operations. But you're right, it will be good to have all the checks at one place, I've moved this to `initOperation`. Thanks!

lxd/operations/operations.go
skozina: Hm... you're right. Ultimate we'll have few different conditions which we'll need to implement here:
 - stages (don't start the child operation of stages >= 1, these will be started when all ops of previous stage finish)
 - `ConflictActionEnqueue`. Instead of failing operation creation with failing reference, we'll allow it, but we'll only start its go routine when the conflicting op finishes.
 - rate limiting on number of concurrently running operations

You're right that we'll need stages for replicators. I'll look at implementing them.

doc/api-extensions.md
skozina: This will be the first commit of the durable operations PR. All commits before this one will land in separate PRs.
Note that there's another API extension for adding the `Parent` and `Children` fields in the `api.Operation` struct. I've moved that one to also be the first commit of the bulk operations PR.

lxd/operations/operations.go
skozina: Sure, changed per your suggestion. Thanks!

lxd/db/cluster/operations.mapper.go
skozina: I've left one commit (`Add directive to get operations by Parent`) for the bulk operations PR, and merged the other two commits into one for the durable operations PR (`Add directive to get operations by NodeID and Class`).
So there will be only one commit adding directives in each PR.

lxd/operations.go
skozina: Yes, loading all operations from the DB on `GET /1.0/operations` will be a separate PR, roughly the first 10 commits of the durable-workflows branch.
The operations are loaded on line 629 below.

lxd/operations.go
skozina: Hm... Here, the `dbOps` are just reconstructed from the database. So nothing is really running anywhere (yet). I've changed `ConstructOperationFromDB()` to cancel both the `running` and `finished` contexts if operation status is final. This allows us to use the `running` context here for detection.

Ultimately the database is the source of truth of status of durable operations. Fortunately we update the status in the DB *before* we cancel the running context. So we can assume that if the running context is cancelled, the status in the DB is also already final.

I think we *can* use the `op.IsRunning` everywhere, for durable operations it just means we're not looking at the source of truth. I thought I'd double check with you if you think using `op.IsRunning` is safer than the status of the operation for durable operations?

lxd/operations.go
skozina: I've added:
```
// With durable operations, there's a chance that heartbeat replies are lost before reaching the leader node.
// In such case the node will continue running the durable operations (because it's receiving the heartbeats),
// but the leader will also restart those operations (because it's not receiving the heartbeats).
// To avoid this, we have a periodic task on each node checking that the node is doing what the database says.
// In other words, we cancel local operations which are not written in the database, and we start operations which
// are in the database but are not running locally.
```

skozina:
Opened #17783 for getting the operations from DB rather than memory.

lxd/operations.go
skozina: I've changed the PR to use following pattern:
 - code using only `cluster.Operation`, the `operations.Operation` is not even constructed: There's no context, so we are using `status.IsFinal()` to check state
 - code where `cluster.Operation` is converted into `operations.Operation`, but not really used yet, there's no operation running: The conversion code cancels the `running` context if `status.IsFinal()`, so we are using `operation.IsRunning()` to check state, which looks at `running` context
 - code where we are looking at running operations: We are always working with `operations.Operation`, so we are always using `operation.IsRunning()` to check its state.

markylaing:
@tomponline @kadinsayani tests now contain:
- Durable op on non-leader (killed and network partitioned)
- Durable op on leader (killed and network partitioned)

Both the leader and non-leader should be waiting for the offline threshold before cancelling the durable operation locally.

All tests passing. Hopefully we're close to merging now.

tomponline:
Please can you rebase

markylaing:
@tomponline this is ready for you to take a look at when you are able. Thanks

test/suites/clustering.sh
```diff
@@ -7027,4 +7027,236 @@ test_clustering_acme() {
...
+
+  # Respawn the second node.
+  echo "Respawning cluster member 2..."
+  respawn_lxd_cluster_member "${ns2}" "${LXD_TWO_DIR}"
+
+  echo "Started member 2"
+
+  # Wait for node2 to be detected as back online.
+  for _ in $(seq 20); do
+    if [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc query /1.0/cluster/members/node2 | jq --exit-status -r '.status')" = 'Online' ]; then
```
simondeziel, line 7095:
```suggestion
    if LXD_DIR="${LXD_TWO_DIR}" lxc query /1.0/cluster/members/node2 | jq --exit-status '.status == "Online"'; then
```

test/suites/clustering.sh
```diff
@@ -7027,4 +7027,236 @@ test_clustering_acme() {
...
+      break
+    fi
+
+    sleep 1
+  done
+
+  if [ "${succeeded}" = 0 ]; then
+    echo "Durable operation did not move to the leader after network partition"
+    # Clean up partition before failing
+    ip netns exec "${ns1}" iptables -F || true
```
simondeziel, line 7174: Could those iptables rules be left in place for an easier `LXD_INSPECT=1` behavior?

test/suites/clustering.sh
```diff
@@ -7027,4 +7027,236 @@ test_clustering_acme() {
...
+  # Verify the operation was cancelled on the partitioned node by checking it's not in the local operations list
+  # Note: We cannot query the partitioned node's API due to the partition, so we'll verify after removing partition
+
+  # Remove the network partition
+  echo "Removing network partition"
+  ip netns exec "${ns1}" iptables -F
+
+  # Wait for partitioned member to be detected as back online.
+  for _ in $(seq 20); do
+    if [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc query /1.0/cluster/members/node1 | jq --exit-status -r '.status')" = 'Online' ]; then
```
simondeziel, line 7187:
```suggestion
    if LXD_DIR="${LXD_TWO_DIR}" lxc query /1.0/cluster/members/node1 | jq --exit-status '.status == "Online"'; then
```

test/suites/clustering.sh
```diff
@@ -7027,4 +7027,236 @@ test_clustering_acme() {
...
+  # Verify the operation was cancelled on the partitioned node by checking it's not in the local operations list
+  # Note: We cannot query the partitioned node's API due to the partition, so we'll verify after removing partition
+
+  # Remove the network partition
+  echo "Removing network partition"
+  ip netns exec "${leader_ns}" iptables -F
+
+  # Wait for partitioned member to be detected as back online.
+  for _ in $(seq 20); do
+    if [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc query "/1.0/cluster/members/${leader_name}" | jq --exit-status -r '.status')" = 'Online' ]; then
```
simondeziel, line 7238:
```suggestion
    if LXD_DIR="${LXD_ONE_DIR}" lxc query "/1.0/cluster/members/${leader_name}" | jq --exit-status '.status == "Online"'; then
```

test/suites/clustering.sh
```diff
@@ -7027,4 +7027,236 @@ test_clustering_acme() {
...
+      break
+    fi
+
+    sleep 1
+  done
+
+  if [ "${succeeded}" = 0 ]; then
+    echo "Durable operation did not move to the leader after network partition"
+    # Clean up partition before failing
+    ip netns exec "${leader_ns}" iptables -F || true
```
simondeziel, line 7225: Same here, I wouldn't do the cleanup on failure.

test/suites/clustering.sh
```diff
@@ -7027,4 +7027,236 @@ test_clustering_acme() {
...
+      break
+    fi
+
+    sleep 1
+  done
+
+  if [ "${succeeded}" = 0 ]; then
+    echo "Durable operation was not restarted on the newly elected leader"
+    false
+  fi
```
simondeziel, line 7129: Could this be moved to a separated helper function?

shared/api/operation.go
```diff
@@ -6,14 +6,23 @@ import (
...
-const OperationClassWebsocket = "websocket"
+	// OperationClassToken represents the Token OperationClass.
+	// These are used to track the tokens used by new cluster members joining the cluster.
+	OperationClassToken = "token"
 
-// OperationClassToken represents the Token OperationClass.
-const OperationClassToken = "token"
+	// OperationClassDurable represents the Durable OperationClass.
+	// These are like task operations, but are restarted on the leader if the member that was initially running the operation goes offline.
+	OperationClassDurable = "durable"
```
markylaing, line 25: Break into two commits:
 - restructure and and comments to existing
 - Add types for durable class

lxd/operations/db.go
```diff
@@ -267,6 +267,15 @@ func ConstructOperationFromDB(ctx context.Context, tx *sql.Tx, s *state.State, d
 		}
 	}
 
+	if op.class == operationtype.OperationClassDurable {
```
markylaing, line 333: Either:
- Durable parent with children and no run hook
- Durable operation with run hook and no children

lxd/operations/operations.go
```diff
@@ -166,6 +169,25 @@ func ScheduleServerOperation(s *state.State, args OperationArgs) (*Operation, er
 	return scheduleOperation(s, args)
 }
 
+// DurableOperationTable represents the table of durable operation hooks.
+type DurableOperationTable map[operationtype.Type]RunHook
```
markylaing: move this table to the operations package. Add function to add to it.

e.g. 
```go

func init() {
	operations.RegisterDurableOperationRunHook(operationtype.ReplicatorFinalize, replicatorFinalizeDurableHook)
}
```

lxd/operations/operations.go
```diff
@@ -295,6 +301,16 @@ func scheduleOperation(s *state.State, args OperationArgs) (*Operation, error) {
 			return nil, errors.New("Token operations cannot have a Run hook")
 		}
 
+		if op.class == operationtype.OperationClassDurable {
+			// Durable operations must have their hooks defined in the durable operations table.
```
markylaing: Unless they are a parent

lxd/operations/operations.go
```diff
@@ -361,6 +361,35 @@ func scheduleOperation(s *state.State, args OperationArgs) (*Operation, error) {
...
+			reconstructedChildOp, err := loadDurableOperationFromDB(childOp)
+			if err != nil {
+				return nil, fmt.Errorf("Failed reloading durable child operation from database: %w", err)
+			}
+
+			newOp.AddChildren(reconstructedChildOp)
+		}
+
+		op = newOp
+	}
```
markylaing, line 316: Factor this out

lxd/operations/db.go
```diff
@@ -323,3 +323,28 @@ func PruneExpiredOperations(ctx context.Context, s *state.State) error {
 
 	return nil
 }
+
+// loadDurableOperationFromDB reloads a durable operation from the database based on its UUID.
+// This is only used to ease debugging to ensure that the reloaded operation is identical to the one originally created.
+func loadDurableOperationFromDB(op *Operation) (*Operation, error) {
+	var result *Operation
```
markylaing: This is called in a loop :(

Fix

lxd/operations/db.go
```diff
@@ -348,3 +348,115 @@ func loadDurableOperationFromDB(op *Operation) (*Operation, error) {
 
 	return result, nil
 }
+
+// LoadDurableOperationsFromNode returns all durable operations from the db that exist on given node.
+func LoadDurableOperationsFromNode(ctx context.Context, s *state.State, nodeID int64) ([]*Operation, error) {
```
markylaing: Separate into load and update. Then have a "load and ensure"

Set the node ID on all operations in one query.

lxd/operations/db.go
```diff
@@ -348,3 +348,115 @@ func loadDurableOperationFromDB(op *Operation) (*Operation, error) {
...
+
+	// Return values of the operations map as a slice.
+	return result, nil
+}
+
+// RestartDurableOperation creates and starts a new durable operation based on the provided Operation object.
+// This is used to restart operations that were running on a node which failed to respond to heartbeats on other node.
+func RestartDurableOperation(s *state.State, op *Operation) {
+	// Don't restart operations which are already in a final state.
+	if !op.IsRunning() {
```
markylaing, line 504: This should be looking at the operation status rather than the operation context.

lxd/operations/operations.go
```diff
@@ -711,6 +740,50 @@ func (op *Operation) cancelImmediate() {
...
+	op.running.Cancel()
+	for _, child := range op.children {
+		cancelInternal(child)
+	}
+}
+
+// CancelLocalDurableOperations stops all durable operations running on this node.
+// These operations are only removed from the local operations map, the database records are left intact.
+// The cluster leader will later restart these operations.
+func CancelLocalDurableOperations() {
```
tomponline, line 771: should these wait for the cancelled operations to finish?

lxd/operations/db.go
```diff
@@ -701,3 +701,125 @@ func loadAndConstructOperationFromDB(ctx context.Context, s *state.State, opID s
...
+
+	operationsLock.Lock()
+	operations[op.id] = op
+	for _, childOp := range op.children {
+		operations[childOp.id] = childOp
+	}
+
+	operationsLock.Unlock()
+
+	op.logger.Debug("Restarting operation", logger.Ctx{"id": op.id})
```
tomponline, line 904: I think this should be Info level

lxd/cluster/heartbeat.go
```diff
@@ -197,6 +198,15 @@ func (hbState *APIHeartbeat) Send(ctx context.Context, s *state.State, networkCe
 				if err != nil {
 					logger.Warn("Failed creating warning", logger.Ctx{"err": err})
 				}
+
+				offlineTime := time.Now().Add(-offlineThreshold)
```
tomponline, line 202: lets have a comment here saying what its doing.

tomponline: otherwise its just another block of code without a label of what its for.

tomponline: not a blocker - but hb.Send is a weird place to be restarting operations (and clearing warnings for that matter) in my view.

lxd/operations/args.go
```diff
@@ -202,3 +211,19 @@ func (a OperationArgs) validate(isChild bool) error {
...
 }
+
+// SetInputValue sets the given value on the operation inputs. This enforces that the value can be serialized.
+// Values can be retrieved via [GetOperationInputValue].
+func (a *OperationArgs) SetInputValue(key InputKey, value any) error {
+	if a.inputs == nil {
+		a.inputs = map[InputKey]json.RawMessage{}
+	}
+
+	b, err := json.Marshal(value)
```
tomponline, line 222: @markylaing just want to check here, this wont end up storing nested json inside when a.inputs map is encoded to store in the DB will it?

lxd/operations.go
```diff
@@ -1143,6 +1143,33 @@ type operationWaitPost struct {
 	ConflictReference string              `json:"conflict_reference" yaml:"conflict_reference"`
 }
 
+func init() {
+	operations.RegisterDurableOperationRunHook(operationtype.Wait, waitHandlerOperationRunHook)
+}
+
+const operationInputKeyWaitHandlerDuration operations.InputKey = "duration"
+
+func waitHandlerOperationRunHook(ctx context.Context, op *operations.Operation) error {
```
tomponline, line 1152: missing doc block

lxd/operations/args.go
markylaing: I'm not sure what you mean. It's a JSON object where each key is a string but the value can be anything (as if it were a `map[string]any`). The change to this data type is so that we can have a generic function which gets the key and unmarshals it back into the same type that it was set as, rather than getting back an `any` that we need to perform type assertion against.

lxd/operations.go
```diff
@@ -1143,6 +1143,33 @@ type operationWaitPost struct {
...
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-time.Tick(duration):
+	}
+
+	return nil
+}
+
 // operationWaitHandler creates a dummy operation that waits for a specified duration.
 func operationWaitHandler(d *Daemon, r *http.Request) response.Response {
```
tomponline, line 1215: can we add "testing" to this function name so its clear its for testing?

lxd/operations/args.go
markylaing: This change also means that values are unmarshalled lazily, so the values aren't unmarshalled until they are needed.

tomponline: what i mean is that we are marshalling it into `b`, and then that later gets marshalled again as part of `input` map, but i want to check this wont result in a double marshalled json string?

lxd/operations.go
```diff
@@ -1160,13 +1160,54 @@ func waitHandlerOperationRunHook(ctx context.Context, op *operations.Operation)
 		return fmt.Errorf("Invalid duration: %w", err)
 	}
 
-	// Sleep for the duration, or until the run context is cancelled.
-	select {
-	case <-ctx.Done():
-		return ctx.Err()
-	case <-time.Tick(duration):
+	logger.Warnf("Starting wait handler operation for %s", duration.String())
```
tomponline, line 1163: we dont use Warnf any more

tomponline (dismissed):

tomponline:
a few things to confirm or to fix in a follow up PR, thanks!

lxd/operations/args.go
tomponline: but what we want to end up in the DB is the equivalent of if we just marshalled a map[string]any once. Is that is what happening with RawMessage?

lxd/operations.go
simondeziel: Should we have a linter script to enforce this?

tomponline:
@markylaing btrfs cluster check failed:

> Durable operation did not move to the leader after network partition

lxd/operations/args.go
markylaing: Yeah that's exactly what is happening :)

markylaing:
> @markylaing btrfs cluster check failed:
> 
> > Durable operation did not move to the leader after network partition

:facepalm: I have never seen this in CI before and it is not happening locally.

markylaing:
> > @markylaing btrfs cluster check failed:
> > > Durable operation did not move to the leader after network partition
> 
> 🤦 I have never seen this in CI before and it is not happening locally.

It worked on retry. Current theory is a race with operation synchronization logic that has changed since this PR last had test runs. When an operation is internally cancelled, it's status is set to cancelled locally but not in the database. If the operation synchronization task runs, it can update the status to cancelled in the background. Since the operation is then finalized, it is never restarted.

I think I need to leave the status locally as running (because technically it is still running, just being moved to another member)

elijahgreenstein (dismissed):
Docs update LGTM

markylaing:
I think this should be good to go now.

tomponline (commented):
Thanks!

PR #17047, opened 2025-11-24
Cluster: Fix clearing a member when using remote storage

roosterfish:
Whilst working on https://github.com/canonical/lxd/pull/16286 I found that when using remote storage and force removing a cluster member, currently LXD leaves traces of an instance's volume behind in the `storage_volumes` table which causes a subsequent `lxd recover` executed on the pool to error out with:

`Error: Failed validation request: Failed checking volumes on pool "data": Failed to detect unknown instances: Instance "c1" in project "default" already has storage DB record`.

This issue only surfaces in case the force removed member still had instances located on it prior to removal.

### DIFF
--- a/test/main.sh
+++ b/test/main.sh
@@ -481,6 +481,7 @@ if [ "${1:-"all"}" != "snap" ] && [ "${1:-"all"}" != "standalone" ]; then
     run_test test_clustering_waitready "clustering waitready"
     run_test test_clustering_heal_networks_stop "clustering heal networks stop"
     run_test test_clustering_placement_groups "clustering placement groups"
+    run_test test_clustering_force_removal "clustering force removal"
 fi
 
 if [ "${1:-"all"}" != "snap" ] && [ "${1:-"all"}" != "cluster" ]; then


lxd/db/node.go
```diff
@@ -969,20 +970,41 @@ SELECT storage_volumes.name
 }
 
 // ClearNode removes any instance or image associated with this node.
-func (c *ClusterTx) ClearNode(ctx context.Context, id int64) error {
-	_, err := c.tx.Exec("DELETE FROM instances WHERE node_id=?", id)
+func (c *ClusterTx) ClearNode(ctx context.Context, node NodeInfo) error {
+	err := c.InstanceList(ctx, func(inst InstanceArgs, project api.Project) error {
```
tomponline: Can we use a join to delete the volume records based on the instances of a node, and then use the original query to delete the instance records. This will then avoid the need to load the project ID for every instance.

roosterfish: I have tried using `DELETE FROM ... JOIN` but this seems to be not supported by sqlite. 

However I experimented a bit more and the key to success is tuple comparison:

`DELETE FROM storage_volumes WHERE (name, project_id) IN (SELECT name, project_id FROM instances WHERE node_id=?) AND type IN (<vm>, <container>)`.

Just pushed the fixes.

tomponline (approved):
Thanks!

PR #17053, opened 2025-11-24
Add database-client member role (from Incus)

kadinsayani:
Introduces a new `database-client` cluster role to LXD and updates the cluster membership and rebalancing logic to support it. The `database-client` role prevents nodes from being elected as database voters or standbys, ensuring they remain non-voting members in the cluster. Cluster members assigned the `database-client` role are assigned to the Raft spare role.

### DIFF
--- a/doc/reference/dqlite-internals.md
+++ b/doc/reference/dqlite-internals.md
@@ -31,3 +31,4 @@ For more information on the Canonical Dqlite Raft implementation, see [`dqlite/s
 1. `database`: Assigned to cluster members with the `RAFT_VOTER` role.
 1. `database-standby`: Assigned to cluster members with the `RAFT_STANDBY` role.
 1. `database-leader`: Assigned to the current Raft leader.
+1. `database-client`: When assigned to a cluster member, LXD assigns the `RAFT_SPARE` role to that member.


test/suites/clustering.sh
```diff
@@ -3090,42 +3091,83 @@ test_clustering_rebalance() {
...
+  echo "Verify node2 (standby without database-client) was promoted to voter instead."
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node2 | grep -xF "database: true"
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node2 | grep -cF -- "- database-standby")" = "0" ]
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster list # For debugging.
+
+  echo "Clean up"
   LXD_DIR="${LXD_ONE_DIR}" lxd shutdown
   LXD_DIR="${LXD_TWO_DIR}" lxd shutdown
   LXD_DIR="${LXD_THREE_DIR}" lxd shutdown
-  LXD_DIR="${LXD_FOUR_DIR}" lxd shutdown
```
simondeziel, line 3123: I'd keep that line but commented out with an additional note that `LXD_FOUR_DIR` was abruptly killed earlier.

lxd/cluster/membership.go
```diff
@@ -736,7 +736,49 @@ func Rebalance(state *state.State, gateway *Gateway, unavailableMembers []string
 
 	nodes, err := gateway.currentRaftNodes()
 	if err != nil {
-		return "", nil, fmt.Errorf("Get current raft nodes: %w", err)
+		return "", nil, fmt.Errorf("Failed getting current raft nodes: %w", err)
+	}
+
+	var members []db.NodeInfo
+	err = state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
```
roosterfish: Instead of `context.TODO()` can you use `state.ShutdownCtx`?

lxd/api_cluster.go
```diff
@@ -1759,6 +1759,41 @@ func updateClusterNode(s *state.State, gateway *cluster.Gateway, r *http.Request
 		return response.BadRequest(errors.New("Cluster members need to belong to at least one group"))
 	}
 
+	// Prevent assigning "database-client" role to all nodes.
+	clientNodes := 0
+	nodesCount := 0
+	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
+		nodesCount, err = tx.GetNodesCount(ctx)
```
roosterfish: Can you omit doing `tx.GetNodesCount` and instead run `len()` on the response of `tx.GetNodes`?

kadinsayani: Yes, thanks!

lxd/cluster/membership.go
kadinsayani: Done :)

lxd/daemon.go
```diff
@@ -2586,6 +2587,11 @@ func (d *Daemon) nodeRefreshTask(heartbeatData *cluster.APIHeartbeat, isLeader b
 				if node.RaftID == 0 {
 					hasNodesNotPartOfRaft = true
 				}
+
+				// Check if a "database-client" node currently has a raft role other than "spare".
+				if slices.Contains(node.Roles, db.ClusterRoleDatabaseClient) && node.RaftRole != int(db.RaftSpare) {
```
roosterfish, line 2592: Just for my understanding, does this bit trigger demoting the member on the raft layer from role `x` to `spare` in case we set the `database-client` role in between two heartbeats? 

If this is the case, I wonder if we should rather rename `hasDbClientToProcess` to something more descriptive like `is(Demoted|Degraded)ToClient`?

roosterfish (approved):
LGTM, only one remark/question which shouldn't be a blocker.

lxd/daemon.go
kadinsayani: That's correct. I've renamed that variable, and reworded and moved around the comments for clarity.

lxd/cluster/membership.go
```diff
@@ -736,7 +736,49 @@ func Rebalance(state *state.State, gateway *Gateway, unavailableMembers []string
 
 	nodes, err := gateway.currentRaftNodes()
 	if err != nil {
-		return "", nil, fmt.Errorf("Get current raft nodes: %w", err)
+		return "", nil, fmt.Errorf("Failed getting current raft nodes: %w", err)
+	}
+
+	var members []db.NodeInfo
+	err = state.DB.Cluster.Transaction(state.ShutdownCtx, func(ctx context.Context, tx *db.ClusterTx) error {
+		members, err = tx.GetNodes(ctx)
```
kadinsayani, line 746: We should check if Rebalance() call sites already load cluster members.

lxd/cluster/membership.go
```diff
@@ -736,7 +736,49 @@ func Rebalance(state *state.State, gateway *Gateway, unavailableMembers []string
 
 	nodes, err := gateway.currentRaftNodes()
 	if err != nil {
-		return "", nil, fmt.Errorf("Get current raft nodes: %w", err)
+		return "", nil, fmt.Errorf("Failed getting current raft nodes: %w", err)
```
kadinsayani, line 741: Separate commit.

lxd/cluster/membership.go
```diff
@@ -736,7 +736,49 @@ func Rebalance(state *state.State, gateway *Gateway, unavailableMembers []string
...
+		members, err = tx.GetNodes(ctx)
+		if err != nil {
+			return fmt.Errorf("Failed getting cluster members: %w", err)
+		}
+
+		return nil
+	})
+	if err != nil {
+		return "", nil, err
+	}
```
kadinsayani, line 755: We can probably move this to newRoleChanges().

lxd/daemon.go
```diff
@@ -2597,7 +2603,7 @@ func (d *Daemon) nodeRefreshTask(heartbeatData *cluster.APIHeartbeat, isLeader b
 		// If there are offline members that have voter or stand-by database roles, let's see if we can
 		// replace them with spare ones. Also, if we don't have enough voters or standbys, let's see if we
 		// can upgrade some member.
-		if isDegraded || onlineVoters < maxVoters || onlineStandbys < maxStandBy {
+		if isDegraded || onlineVoters < maxVoters || onlineStandbys < maxStandBy || hasNodeToAssignSpareRole {
```
kadinsayani: Check if we can simplify this logic and reduce variables, similar to https://github.com/canonical/lxd/pull/17135/commits/fcec4399850627110b5b026c8be45322cebe7be5.

lxd/daemon.go
```diff
@@ -2603,7 +2603,7 @@ func (d *Daemon) nodeRefreshTask(heartbeatData *cluster.APIHeartbeat, isLeader b
 		// If there are offline members that have voter or stand-by database roles, let's see if we can
 		// replace them with spare ones. Also, if we don't have enough voters or standbys, let's see if we
 		// can upgrade some member.
-		if isDegraded || onlineVoters < maxVoters || onlineStandbys < maxStandBy || hasNodeToAssignSpareRole {
+		if isDegraded || onlineVoters != maxVoters || onlineStandbys != maxStandBy || hasNodeToAssignSpareRole {
```
kadinsayani, line 2605: This can be in a separate PR.

lxd/api_cluster.go
```diff
@@ -1759,6 +1759,38 @@ func updateClusterNode(s *state.State, gateway *cluster.Gateway, r *http.Request
...
+		return nil
+	})
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	if clientNodes+1 >= nodesCount {
+		return response.BadRequest(errors.New(`Assigning the "database-client" role to all nodes is not allowed`))
+	}
+
```
kadinsayani, line 1793: We could consider triggering a rebalance here, moving validation into rebalance function, and trigger reverts if a rebalance is not possible.

test/suites/clustering.sh
```diff
@@ -3119,15 +3119,56 @@ test_clustering_rebalance() {
...
 
+  echo "Add a fifth node to the cluster."
+  # Spawn a fifth node
+  setup_clustering_netns 5
+  LXD_FIVE_DIR=$(mktemp -d -p "${TEST_DIR}" XXX)
+  ns5="${prefix}5"
+  spawn_lxd_and_join_cluster "${ns5}" "${bridge}" "${cert}" 5 1 "${LXD_FIVE_DIR}" "${LXD_ONE_DIR}"
+
+  echo "Remove standby role and add client role to fifth node."
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster role remove node5 database-standby
```
kadinsayani, line 3130: We shouldn't be able to remove automatically assigned roles.

kadinsayani: Test behavior in Incus.

kadinsayani (commented):
Reviewed with @tomponline during 1:1.

I'm going to spend some time considering adding a `database-spare` role which is displayed once a member has actually been assigned to the Raft spare role, and just using the `database-client` role (name subject to change) as the preference. This would also allow us to display a role for offline nodes that are assigned the Raft spare role.

lxd/daemon.go
kadinsayani: https://github.com/canonical/lxd/pull/17171

test/suites/clustering.sh
kadinsayani: It looks like we only have validation for adding/removing the `database` role:

https://github.com/canonical/lxd/blob/2d41139159b30e92ceeefff8a7bd82c966c895a7/lxd/api_cluster.go#L1748-L1755

Incus has the same.

kadinsayani:
Closing for now to reduce PR clutter. I'll reopen this following review and approval of the spec, as there is more thought and discussion required on how we want to introduce this to LXD.

test/suites/clustering.sh
kadinsayani: https://github.com/canonical/lxd/pull/17259 improves cluster role validation.

lxd/daemon.go
kadinsayani: https://github.com/canonical/lxd/pull/17171/changes/ff0f7dcecb0819fd6ab99da4848463ac06b7fccf

PR #17054, opened 2025-11-24
Disallow sending a TLS certificate when updating an OIDC or Bearer identity

markylaing:
Bit of a weird one.

If an identity with `can_edit` on another identity populates the `tls_certificate` field when calling `PUT` or `PATCH` `/1.0/auth/identities/oidc/{email}`, then the identities' identifier becomes the certificate fingerprint and the identity metadata is replaced with the certificate. The certificate that is sent is useless, it can't be used to authenticate with LXD, and there's no way to perform this action in the UI or CLI, so I think this is quite an edge case.

Note that it is not possible for an OIDC identity to populated the tls_certificate field to "update their own certificate" unless they have "can_edit" on their own identity. This is because the access handlers for `PUT`/`PATCH` `/1.0/auth/identities/oidc/{email}` are different from `PUT`/`PATCH` `/1.0/auth/identities/tls/{fingerprint}`.

### DIFF
--- a/test/suites/auth.sh
+++ b/test/suites/auth.sh
@@ -349,6 +349,29 @@ fine_grained: true"
   LXD_CONF="${LXD_CONF6}" CERTNAME=unrestricted my_curl -X GET "https://${LXD_ADDR}/1.0/auth/identities/current" | jq --exit-status '.metadata.fine_grained == false'
   lxc config trust remove "${lxdconf6_fingerprint_short}"
 
+  # Check that it is not possible to send a certificate in the request body to update an OIDC or bearer identity.
+  LXD_CONF7=$(mktemp -d -p "${TEST_DIR}" XXX)
+  LXD_CONF="${LXD_CONF7}" gen_cert_and_key "client"
+  conf7_cert="$(awk '{printf "%s\\n", $0}' "${LXD_CONF7}/client.crt")"
+  lxc auth identity create devlxd/test-bearer
+
+  # As an admin
+  ! lxc query -X PUT /1.0/auth/identities/bearer/test-bearer -d "{\"tls_certificate\":\"${conf7_cert}\",\"name\":\" \",\"identifier\":\"test-user@example.com\"}" || false
+  [ "$("${_LXC}" query -X PUT /1.0/auth/identities/bearer/test-bearer -d "{\"tls_certificate\":\"${conf7_cert}\",\"name\":\" \",\"identifier\":\"test-user@example.com\"}"  2>&1 >/dev/null)" = 'Error: Cannot update certificate for identities of type "DevLXD token bearer"' ]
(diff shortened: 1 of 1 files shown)


test/suites/auth.sh
```diff
@@ -251,30 +251,30 @@ fine_grained: true"
...
-  auth_project_features "oidc"
-  LXD_CONF="${LXD_CONF2}" auth_project_features "tls"
-
-  # Entitlement enrichment
-  entities_enrichment_with_entitlements
-
-  # Access checks with project specific networks.
-  auth_ovn "oidc"
-  LXD_CONF="${LXD_CONF2}" auth_ovn "tls"
+#  LXD_CONF="${LXD_CONF2}" events_filtering
```
tomponline: @markylaing left over commenting out?

markylaing: Whoopsie!

markylaing: Just trying to speed up the tests :wink:

markylaing: Fixed

test/suites/auth.sh
```diff
@@ -349,6 +349,29 @@ fine_grained: true"
   LXD_CONF="${LXD_CONF6}" CERTNAME=unrestricted my_curl -X GET "https://${LXD_ADDR}/1.0/auth/identities/current" | jq --exit-status '.metadata.fine_grained == false'
   lxc config trust remove "${lxdconf6_fingerprint_short}"
 
+  # Check that it is not possible to send a certificate in the request body to update an OIDC or bearer identity.
+  LXD_CONF7=$(mktemp -d -p "${TEST_DIR}" XXX)
+  LXD_CONF="${LXD_CONF7}" gen_cert_and_key "client"
+  conf7_cert="$(awk '{printf "%s\\n", $0}' "${LXD_CONF7}/client.crt")"
```
simondeziel, line 355: It's good to have it extracted into a variable but would you mind doing it for cert4 and cert5 as well for consistency? ATM, those 2 have the `awk` call done "inline" with the `my_curl` calls.

test/suites/auth.sh
```diff
@@ -349,6 +349,29 @@ fine_grained: true"
   LXD_CONF="${LXD_CONF6}" CERTNAME=unrestricted my_curl -X GET "https://${LXD_ADDR}/1.0/auth/identities/current" | jq --exit-status '.metadata.fine_grained == false'
   lxc config trust remove "${lxdconf6_fingerprint_short}"
 
+  # Check that it is not possible to send a certificate in the request body to update an OIDC or bearer identity.
+  LXD_CONF7=$(mktemp -d -p "${TEST_DIR}" XXX)
+  LXD_CONF="${LXD_CONF7}" gen_cert_and_key "client"
```
simondeziel, line 354: Not for now but maybe for clarity and as a simplification we could skip the `LXD_CONFX` part and simply use a different CN for those multiple `clientX` certs?

PR #17070, opened 2025-11-25
Storage volume/instance snapshot edit restriction

### LINKED ISSUE #14423: `lxc storage volume edit` should fail when editing snapshots
I suspect that snapshot volumes cannot have their configuration changed (although I haven't confirmed it). The following command should probably fail and print an error.

## Steps to reproduce
`storage volume edit` opens the editor, and returns 0 regardless of the changes made in the yaml:
```sh 
$ lxc storage volume edit default virtual-machine/vm1/snap0
$ echo $?
0
```

None of the changes made during the edit are persisted (`lxc storage volume show virtual-machine/vm1/snap0`).

tugbataluy:
## Checklist

## What Does This PR Do?
When editing a Storage Volume snapshot, modifications to writable fields are saved successfully. However, changes to read-only fields are simply ignored without notification, which can be confusing for users, as noted in issue #14423.

The same situation applies to instance snapshots.
This PR ensures that if a snapshot is present within the two edit commands related to this situation, input validation is performed, and if a non-editable property is modified, an error is returned to the user.

Fixes #14423

### DIFF
--- a/test/suites/snapshots.sh
+++ b/test/suites/snapshots.sh
@@ -55,6 +55,47 @@ snapshots() {
   # Check if the snapshot's UUID can be modified
   ! lxc storage volume set "${pool}" container/foo/snap0 volatile.uuid "2d94c537-5eff-4751-95b1-6a1b7d11f849" || false
 
+  # Check snapshot configuration editing
+  # Test that only expires_at field can be modified for instance snapshots
+
+  tmp_yaml=$(mktemp)
+  ERROR_MSG="Error: Only \"expires_at\" field(s) can be modified for instance snapshots"
+
+  # Test all non-editable properties
+  for field in name architecture config devices ephemeral profiles stateful created_at last_used_at; do
+    lxc config show foo/snap0 > "$tmp_yaml"
+    case $field in
+      config|devices)
+        sed -i "s/^${field}:.*/${field}: {}/" "$tmp_yaml"
+        ;;
+      profiles)
+        sed -i "s/^${field}:.*/${field}: []/" "$tmp_yaml"
+        ;;
+      ephemeral|stateful)
+        sed -i "s/^${field}:.*/${field}: true/" "$tmp_yaml"
+        ;;
+      created_at|last_used_at)
+        sed -i "s/^${field}:.*/${field}: 2024-01-01T00:00:00Z/" "$tmp_yaml"
+        ;;
+      *)
+        sed -i "s/^${field}:.*/${field}: invalid-${field}/" "$tmp_yaml"
+        ;;
+    esac
+
+    ! lxc config edit foo/snap0 < "$tmp_yaml" 2>&1 | grep -xF "$ERROR_MSG" || false
+  done
+
+  # Test that expires_at can be modified
+  expiry_date=$(date -u -d '+1 day' '+%Y-%m-%dT%H:%M:%SZ')
+  lxc config show foo/snap0 > "$tmp_yaml"
+  sed -i "s/^expires_at:.*/expires_at: ${expiry_date}/" "$tmp_yaml"
+  lxc config edit foo/snap0 <<EOF
+$(cat "$tmp_yaml")
+EOF
+  lxc config show foo/snap0 | grep -F "expires_at: ${expiry_date}"
+
+  rm -f "$tmp_yaml"
+
   lxc snapshot foo
   # FIXME: make this backend agnostic
   if [ "$lxd_backend" = "dir" ]; then


lxc/storage_volume.go
```diff
@@ -1111,13 +1112,37 @@ func (c *cmdStorageVolumeEdit) run(cmd *cobra.Command, args []string) error {
...
+			}
+
+			// Create a copy of the original snapshot with only writable properties changed
+			expectedVol := *snapVol
+			expectedVol.SetWritable(newdata)
+
+			// Validate that only writable properties were modified
+			if !reflect.DeepEqual(check, expectedVol) {
+				return fmt.Errorf("%s", i18n.G("Only 'description' and 'expires_at' fields can be modified for snapshot volumes"))
+			}
```
MusicDin: Why can't we compare `newdata` and `expectedVol`? `newdata` does not seem to be modified, and `check` is parsed only to feed comparison?

lxc/storage_volume.go
```diff
@@ -1184,6 +1209,34 @@ func (c *cmdStorageVolumeEdit) run(cmd *cobra.Command, args []string) error {
...
+				if err != nil {
+					return fmt.Errorf(i18n.G("Failed to parse edited content: %w"), err)
+				}
+
+				// Create a copy of the original snapshot with only writable properties changed
+				expectedVol := *snapVol
+				expectedVol.SetWritable(newdata)
+
+				// Compare if the user tried to modify anything beyond writable fields
+				if !reflect.DeepEqual(check, expectedVol) {
```
MusicDin: Can we avoid using reflection? (also above)

If the comparison is necessary, we could have a function that compares two structs

lxc/config.go
```diff
@@ -416,6 +479,92 @@ func (c *cmdConfigEdit) run(cmd *cobra.Command, args []string) error {
...
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+// validateSnapshotEdit validates that only editable fields are modified.
+func (c *cmdConfigEdit) validateSnapshotEdit(current *api.InstanceSnapshot, edited *api.InstanceSnapshot) error {
```
MusicDin: Huh, there is a lot of reflection used.
Since this is used only for snapshots, I would suggest comparing the snapshot fields directly. While it becomes *less dynamic*, it is more performant (less of a concern in cli), retains compile-time safety, and improves readability. Additionally, you can tailor error messages if needed.

tomponline:
> the operation silently fails without providing any feedback to the user.

It doesn't silently fail, just the read-only fields are not modified, but the writable ones are still persisted.

lxc/storage_volume.go
tomponline: agreed, there seems no need for the duplicate.

tomponline: 1. Get original content into `origdata` (as `api.StorageVolumeSnapshot`).
2. Unmarshal stdin to `newdata` (into `api.StorageVolumeSnapshot`).
3. Get writable fields from `newdata` and store into `origdata` (via `origdata.SetWritable(newdata.Writable))`.
4. Check `newdata` is the same as `origdata` (using `reflect.DeepEqual()`).
5. If not then it means that `newdata` has had some of the non-writable fields change, so error.

lxc/storage_volume.go
tomponline: The structs contain non-scalar fields (such as `map[string]string`) however we could add an `Equal()` function to the api type.

tomponline:
@tugbataluy the tests are failing and this needs a rebase to be mergeable as other conflicting PRs have been merged in the interim.

You can open a draft PR, check the tests pass, before promoting to "reviewable".

lxc/config.go
```diff
@@ -416,6 +478,150 @@ func (c *cmdConfigEdit) run(cmd *cobra.Command, args []string) error {
...
+		return true
+	}
+
+	if a == nil {
+		return len(b) == 0
+	}
+
+	if b == nil {
+		return len(a) == 0
+	}
```
MusicDin: I think you can skip those 3 checks. Iterating over nil map treats it as a zero length map

lxc/config.go
```diff
@@ -416,6 +478,150 @@ func (c *cmdConfigEdit) run(cmd *cobra.Command, args []string) error {
...
```
MusicDin: Same as above

lxc/config.go
```diff
@@ -416,6 +478,150 @@ func (c *cmdConfigEdit) run(cmd *cobra.Command, args []string) error {
...
+
+	return true
+}
+
+func sliceEqual(a, b []string) bool {
+	// Both nil or both empty
+	if len(a) == 0 && len(b) == 0 {
+		return true
+	}
```
MusicDin: Same as above

lxc/config.go
```diff
@@ -416,6 +478,150 @@ func (c *cmdConfigEdit) run(cmd *cobra.Command, args []string) error {
...
+func (c *cmdConfigEdit) validateSnapshotEdit(current *api.InstanceSnapshot, edited *api.InstanceSnapshot) error {
+	// For snapshots, only ExpiresAt is writable. All other fields must remain unchanged.
+
+	// Create expected by copying current and applying writable changes
+	expected := *current
+	expected.SetWritable(edited.Writable())
+
+	// Now compare field by field, being careful with nil maps
+	if expected.Architecture != edited.Architecture {
+		return errors.New("field 'architecture' cannot be modified")
```
MusicDin: Could you please capitalize errors?

tomponline:
Please rebase

lxc/storage_volume.go
```diff
@@ -1111,13 +1112,29 @@ func (c *cmdStorageVolumeEdit) run(cmd *cobra.Command, args []string) error {
...
+				return fmt.Errorf(i18n.G("Failed to parse edited content: %w"), err)
+			}
+
+			// Create a copy of the original snapshot with only writable properties changed
+			expectedVol := *origdata
+			expectedVol.SetWritable(newdata.Writable())
+
+			// Validate that only writable properties were modified
+			if !reflect.DeepEqual(newdata, expectedVol) {
+				return fmt.Errorf("%s", i18n.G("Only 'description' and 'expires_at' fields can be modified for snapshot volumes"))
```
tomponline: Please can you use double quoting rather than single quoting.
Also lets avoid the use of `fmt.Errorf` when not using placeholders.
Instead use `errors.New()`.

tomponline: E.g. ```errors.New( i18n.G(`Only "description" and "expires_at" fields can be modified for snapshot volumes`))```

lxc/storage_volume.go
```diff
@@ -1181,11 +1198,33 @@ func (c *cmdStorageVolumeEdit) run(cmd *cobra.Command, args []string) error {
...
 			err = yaml.Unmarshal(content, &newdata)
 			if err == nil {
+				// Validation: Check if user tried to modify read-only properties
+				// Create a copy of the original snapshot with only writable properties changed
+				expectedVol := *snapVol
+				expectedVol.SetWritable(newdata.Writable())
+
+				// Compare if the user tried to modify anything beyond writable fields
+				if !reflect.DeepEqual(newdata, expectedVol) {
+					fmt.Fprintf(os.Stderr, "%s", i18n.G("Error: Only 'description' and 'expires_at' fields can be modified for snapshot volumes")+"\n")
```
tomponline: Here the editable fields have been duplicated from above.

Can we avoid needing to define them in multiple places by defining them in a slice and then constructing the error message from the slice?

Even better, could we use reflection to figure out the editable yaml field names from the struct type returned from `newdata.Writable()` so that these error messages will automatically update if the struct changes in the future?

lxc/config.go
tomponline: Can we not use `reflect.DeepEqual()` like we have done in the earlier commits? 

I dont personally see a problem with using reflect in these cases because they are not hot-path.

tomponline (changes requested):
Thanks @tugbataluy 

Please can you see if we can simplify the instance snapshot editing checks to be similar to the approach used in the volume snaphot editing checks (as its much simpler there).

If not please don't hesitate to explain why though :)

Also, I think we can do a better job of de-duplicating the definition of which fields can be edited to help with building the error messages (and even better would be to use reflection to extract the editing yaml field names from the writable struct itself).

lxc/storage_volume.go
MusicDin: > Even better, could we use reflection to figure out the editable yaml field names from the struct type returned from newdata.Writable() so that these error messages will automatically update if the struct changes in the future?

I apologize here. She did great job using reflection previously (hope she still has this code laying somewhere around). The change was due to my comments, as I though we want to avoid using it - both for readability and performance reasons. 

But if we go with the reflection, we could derive some more general approach and potentially reuse it for multiple entities to avoid having too much of it across the codebase.

tomponline: I'd like us to use a single approach for both, and in the earlier commit @tugbataluy has used the suggested approach here:

https://github.com/canonical/lxd/pull/17070/changes/074d63054c0e04003f79851ffd5998dcd7426737#r2593200563

Whilst reflection has a performance impact, when editing a single snapshot, as we are doing here, and because its in the CLI (so will also incur the overhead of launching the process) I think the tradeoffs are worth it in this case.

Also, in order to extract the original data from the snapshot to do the comparison there is a network request needed, which will likely be an order of magnitude more than the cost of doing reflection, so likely won't make much difference.

tomponline:
@tugbataluy also needs a rebase now that we've removed translations.

lxc/config.go
MusicDin: I tend to agree with that.

To clarify, my point was just that using reflection to manually derive the diff between the fields for one specific entity seemed overkill (even if used just within a CLI), as this will later serve as an "example" for other comparisons. 

However, if the comparison func (using reflection) would be just renamed and allow accepting a pair of any entity, we would simply solve this issue for all (future) entity comparisons.

lxc/storage_volume.go
MusicDin: Had `validateSnapshotEdit` in mind - and I missed your comment https://github.com/canonical/lxd/pull/17070#discussion_r2613929609.

lxc/config.go
tomponline: Basically like https://github.com/google/go-cmp  (which uses similar techniques to reflect.DeepEqual in some cases)

lxc/storage_volume.go
tugbataluy: I'm looking at it from the positive side, and I've tried both perspectives and solutions. I can go back and implement it that way. I'll update the code taking your reviews into consideration.
But maybe I can use a few helper functions for readability. Because when I implement everything in a single function, it's difficult for me to follow. What do you think?

lxc/config.go
MusicDin: Yep nice example. Our implementation could be simplified and focused on returning the field name of a non-editable field.

lxc/storage_volume.go
```diff
@@ -1255,6 +1303,63 @@ func (c *cmdStorageVolumeEdit) run(cmd *cobra.Command, args []string) error {
...
+
+// formatFieldList formats a slice of field names as a quoted, comma-separated list.
+func (c *cmdStorageVolumeEdit) formatFieldList(fields []string) string {
+	if len(fields) == 0 {
+		return ""
+	}
+
+	quoted := make([]string, len(fields))
+	for i, field := range fields {
+		quoted[i] = fmt.Sprintf(`"%s"`, field)
```
tomponline: %q will suffice here

lxc/storage_volume.go
```diff
@@ -1255,6 +1303,63 @@ func (c *cmdStorageVolumeEdit) run(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+// getEditableYAMLFields uses reflection to extract YAML field names from a struct.
+func (c *cmdStorageVolumeEdit) getEditableYAMLFields(v any) []string {
```
tomponline: this function is not specific to `cmdStorageVolumeEdit` and could be used for other edit commands for similar benefit, so to avoid duplication, suggest you define this `lxc/utils_properties.go`

tugbataluy: Thanks Tom, I wasn't sure where to put that function :D

lxc/storage_volume.go
```diff
@@ -1103,6 +1104,15 @@ func (c *cmdStorageVolumeEdit) run(cmd *cobra.Command, args []string) error {
 		isSnapshot = true
 	}
 
+	// Get editable fields for snapshots
+	var snapshotEditableFields []string
+	var snapshotEditableFieldsMsg string
+	if isSnapshot {
+		// Use a temporary instance to determine editable fields
+		tempSnapshot := api.StorageVolumeSnapshot{}
+		snapshotEditableFields = c.getEditableYAMLFields(tempSnapshot.Writable())
```
tomponline: to simplify tracking where these variables are defined and used, how about creating a function inside this function like so:

```go
snapshotEditableFieldsMsg := func(v api.StorageVolumeSnapshotPut) []string{
     return fmt.Sprintf(i18n.G("Only %s field(s) can be modified for snapshot volumes"), c.formatFieldList(c.getEditableYAMLFields(v)))
}
```

tomponline: Then call it from where you need it, e.g.

```go
    fmt.Fprintf(os.Stderr, "Error: %s\n", snapshotEditableFieldsMsg(newdata.Writable()))
```

tugbataluy: Yes I can do that :)

lxc/utils_properties.go
```diff
@@ -227,3 +227,60 @@ func unpackKVToWritable(writable any, keys map[string]string) error {
...
+		field := typ.Field(i)
+		yamlTag := field.Tag.Get("yaml")
+
+		// Skip fields without yaml tags or with "-"
+		if yamlTag == "" || yamlTag == "-" {
+			continue
+		}
+
+		// Extract the field name from the yaml tag (handle "name,omitempty" format)
+		parts := strings.Split(yamlTag, ",")
```
tomponline: could probably use strings.Cut here

lxc/utils_properties.go
```diff
@@ -227,3 +227,60 @@ func unpackKVToWritable(writable any, keys map[string]string) error {
...
+	return fields
+}
+
+// formatFieldList formats a slice of field names as a quoted, comma-separated list.
+func formatFieldList(fields []string) string {
+	if len(fields) == 0 {
+		return ""
+	}
+
+	quoted := make([]string, len(fields))
```
tomponline: please avoid prefilling the slice with zero entries like this, and instead use `make([]string,0,len(fields))` to pre-create with right capacity but without zero entries.

Then use `append` in the `for` loop.

lxc/utils_properties.go
```diff
@@ -227,3 +227,60 @@ func unpackKVToWritable(writable any, keys map[string]string) error {
...
+	for i, field := range fields {
+		quoted[i] = fmt.Sprintf("%q", field)
+	}
+
+	if len(quoted) == 1 {
```
tomponline: lets use `switch` here

lxc/storage_volume.go
```diff
@@ -17,7 +18,7 @@ import (
 	"github.com/spf13/cobra"
 	"go.yaml.in/yaml/v2"
```
tomponline, line 20: we dont import the client as `lxd` so please update your editor settings to not do this automatically

lxc/storage_volume.go
```diff
@@ -1103,6 +1104,11 @@ func (c *cmdStorageVolumeEdit) run(cmd *cobra.Command, args []string) error {
 		isSnapshot = true
 	}
 
+	// Get editable fields for snapshots.
+	snapshotEditableFieldsMsg := func(v api.StorageVolumeSnapshotPut) string {
+		return fmt.Sprintf(i18n.G("Only %s field(s) can be modified for snapshot volumes"), formatFieldList(getEditableYAMLFields(v)))
```
tomponline: just a nit, but no need for `fmt.Sprintf` here, can just do:

```go
return "Only "+formatFieldList(getEditableYAMLFields(v))+" field(s) can be modified for snapshot volumes"
```

Also no need for `i18n.G` anymore.

lxc/storage_volume.go
```diff
@@ -1181,11 +1203,33 @@ func (c *cmdStorageVolumeEdit) run(cmd *cobra.Command, args []string) error {
 	if isSnapshot {
 		for {
 			// Parse the text received from the editor
-			newdata := api.StorageVolumeSnapshotPut{}
+			newdata := api.StorageVolumeSnapshot{}
 			err = yaml.Unmarshal(content, &newdata)
 			if err == nil {
```
tomponline: We tend to use "happy path" style https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

So rather than handle the no-parse error scenario here, please can you move this "happy path" scenario below the `err != nil` logic below.

Also please can you try and refactor this to avoid duplicating the 

```go
					if err != nil {
						return err
					}

					if err != nil {
						return err
					}
```

piece that is now repeated twice.

lxc/storage_volume.go
```diff
@@ -1111,13 +1117,29 @@ func (c *cmdStorageVolumeEdit) run(cmd *cobra.Command, args []string) error {
...
 			if err != nil {
 				return err
 			}
 
-			op, err := client.UpdateStoragePoolVolumeSnapshot(resource.name, volType, fields[0], fields[1], newdata, "")
+			// Unmarshal stdin to newdata
+			newdata := api.StorageVolumeSnapshot{}
+			err = yaml.Unmarshal(contents, &newdata)
+			if err != nil {
+				return fmt.Errorf(i18n.G("Failed to parse edited content: %w"), err)
```
tomponline: please remove `i18n.G("` calls now.

lxc/config.go
```diff
@@ -221,14 +226,37 @@ func (c *cmdConfigEdit) run(cmd *cobra.Command, args []string) error {
 			var op lxd.Operation
 
 			if isSnapshot {
-				newdata := api.InstanceSnapshotPut{}
+				// For snapshots, we need to parse and validate the input.
+				check := api.InstanceSnapshot{}
+				err = yaml.Unmarshal(contents, &check)
+				if err != nil {
+					return err
```
tomponline: In the earlier `lxc/storage_volume.go` file at the equivalent point you returned an error like:

> Failed to parse edited content: %w

but here there is no wrapping. 

Please can it be updated to return consistent error messages.

lxc/config.go
```diff
@@ -221,14 +226,37 @@ func (c *cmdConfigEdit) run(cmd *cobra.Command, args []string) error {
...
 				}
 
-				op, err = resource.server.UpdateInstanceSnapshot(fields[0], fields[1], newdata, "")
+				// Validate that only writable fields have been modified.
+				// Create expected with original data + writable changes.
+				expected := *snapshotInstance
+				expected.SetWritable(check.Writable())
+
+				// Allow the input if it matches the expected result (original + only writable changes) or if it's partial YAML with only writable fields provided
+				testPartial := api.InstanceSnapshot{}
```
tomponline: why dont we need this concept in the `lxc/storage_volume.go` file? It surprised me that they were not identical login.

lxc/config.go
```diff
@@ -298,11 +324,51 @@ func (c *cmdConfigEdit) run(cmd *cobra.Command, args []string) error {
 				newdata := api.InstanceSnapshotPut{}
 				err = yaml.Unmarshal(content, &newdata)
 				if err == nil {
+					// Parse the full edited snapshot.
+					check := api.InstanceSnapshot{}
+					err = yaml.Unmarshal(content, &check)
+					if err != nil {
+						fmt.Fprintf(os.Stderr, i18n.G("Failed to parse edited content: %s")+"\n", err)
+						fmt.Println(i18n.G("Press enter to open the editor again or ctrl+c to abort change"))
```
tomponline: Please can we de-dupe the 

```go
						if err != nil {
							return err
						}

						if err != nil {
							return err
						}
```

part.

lxc/config.go
tugbataluy: Now, as far as I understand, I'll try to explain. It might be a bit long:
The partial YAML validation for instance snapshot editing addresses a key difference in how these resources are presented versus storage volume snapshots. While storage volume snapshots display a complete, transparent structure that users can edit in full, instance snapshots have several fields that are either hidden or incompletely shown due to Go's omitempty tags. Fields like expanded_config and expanded_devices are automatically omitted when empty in the YAML output, and the devices field often appears as an empty object {} even when it contains actual device data. This means users working with instance snapshots cannot reasonably provide the complete api.InstanceSnapshot structure when using non-interactive editing via stdin. To maintain a practical user experience, we support partial YAML input where users only need to specify the writable fields they want to change(such as just providing expires_at or description) while still validating that only permitted fields are modified. Storage volume snapshots don't require this partial handling because their YAML representation accurately reflects all fields, allowing users to work with the complete structure directly.

I couldn't find any other solution, so I proceeded this way.

I hope this was helpful.

tugbataluy: As we discussed with @tomponline this is not needed, full YAML input will be sufficient for both storage volume snapshots and instance snapshots.

PR #17072, opened 2025-11-25
Reworked system-tests matrix

simondeziel:
## Pull request overview

This PR reworks the system test matrix by introducing test groups and multi-backend execution capabilities. The changes aim to improve test organization, reduce CI durations through selective execution, and enable parallel testing across storage backends.

Key changes:
- Introduces test groups (cluster, standalone, storage, etc.) to categorize tests by functionality
- Adds support for running tests against multiple storage backends in a single execution
- Refactors the test execution flow to support group-based and backend-based filtering
- Updates CI workflow to use the new test group matrix structure

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -13,6 +13,7 @@ To run only the integration tests, run from the test directory:
 Name                           | Default                   | Description
 :--                            | :---                      | :----------
 `LXD_BACKEND`                  | dir                       | What backend to test against (btrfs, ceph, dir, lvm, zfs, or random)
+`LXD_BACKENDS`                 | `LXD_BACKEND`             | Space-delimited list of backends to run test against. Accepts `fast` (`btrfs` or `dir`), `fasts` (`btrfs` and `dir`) and `all` (see the list in `LXD_BACKEND`)
 `LXD_CEPH_CLUSTER`             | ceph                      | The name of the ceph cluster to create osd pools in
 `LXD_CEPH_CEPHFS`              | cephfs                    | Enables the CephFS tests using the specified cephfs filesystem for `cephfs` pools
 `LXD_CEPH_CEPHOBJECT_RADOSGW`  | ""                        | Enables the Ceph Object tests using the specified radosgw HTTP endpoint for `cephobject` pools


test/suites/projects.sh
```diff
@@ -9,7 +9,6 @@ test_projects_default() {
   lxc project show default | grep -xF -- "- /1.0/profiles/default"
   lxc project show default | grep -F -- "- /1.0/images/"
   lxc delete c1
-  lxc image delete testimage
```
kadinsayani: Perhaps a comment here that matches the commit description?

test/includes/test-groups.sh
```diff
@@ -0,0 +1,255 @@
+# Test groups
+test_group_cluster=(
```
kadinsayani: Perhaps we can initialize these variables like this for readability:

`readonly test_group_*=(...`

.github/workflows/tests.yml
```diff
@@ -46,6 +46,7 @@ on:
     - cron: '0 0 * * *'  # Test TICS daily
 
 env:
+  LXD_SKIP_TESTS: "concurrent"
```
kadinsayani: Will we ever consider not skipping concurrent test? Perhaps a `XXX: ` comment here?

test/README.md
```diff
@@ -22,6 +22,7 @@ Name                           | Default                   | Description
 `LXD_LOGS`                     | ""                        | Path to a directory to copy all the LXD logs to
 `LXD_OFFLINE`                  | 0                         | Skip anything that requires network access
 `LXD_REPEAT_TESTS`             | 1                         | Number of times to repeat test(s)
+`LXD_RANDOMIZE_TESTS`          | 0                         | Randomize the order of tests in a group
```
kadinsayani, line 26: Perhaps this should be enabled by default, that way we catch race conditions or missing test cleanup steps early.

kadinsayani: IMHO, it would be nice if the ordering of tests (within a group) did not matter and each test was idempotent, performing all necessary cleanup steps. For tests with high temporal coupling, this may be tricky, but it would help in the long run to catch race conditions.

simondeziel: I aim to do that at one point but we are definitely not there yet... still a long way to go :/

kadinsayani: Glad it's on our radar :)

.github/workflows/tests.yml
simondeziel: I think we should reinstate it whenever we have the available cycles to fix the underlying cause as there is definitely a bug underneath this.

How about `XXX: concurrent is too unreliable/racy` ?

test/suites/projects.sh
simondeziel: The `testimage` is special in the sense that it is rarely cleaned up, same for the more recent `ubuntu` and `ubuntu-vm` one. Because of that, I'm not sure it warrants a comment, no strong opinion though.

.github/workflows/tests.yml
kadinsayani: Sounds good!

.github/workflows/tests.yml
```diff
@@ -323,7 +343,7 @@ jobs:
           chmod uog+x /home/runner/go/bin/*
 
       - name: Setup MicroCeph
-        if: ${{ matrix.backend == 'ceph' }}
+        if: ${{ contains(matrix.backends, 'ceph') || matrix.backends == 'all' }}
```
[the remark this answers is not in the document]
simondeziel: `cephfs` isn't a valid backend name but either way, we'd want `microceph` to be installed. The direct string matching was not used because `backends: btrfs ceph` is a valid one.

test/main.sh
```diff
@@ -248,48 +304,131 @@ cleanup() {
...
+
+# Generate markdown table with test durations across backends
+generate_duration_table() {
+    local output="${GITHUB_STEP_SUMMARY:-"/dev/stdout"}"
+
+    # Collect all unique test names
+    local -a test_names=()
+    for key in "${!durations[@]}"; do
+        local test_name="${key%,*}"
+        if [[ ! " ${test_names[*]} " =~ (^|[[:space:]])${test_name}([[:space:]]|$) ]]; then
```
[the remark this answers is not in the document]
simondeziel: Most of the `generate_duration_table` function is Claude's work. However unreadable that regex is, it works so it will remain ;)

Besides, using `grep -q` would require us to update the `test/lint/test-tests.sh` script to not flag this use of `grep` with `-q`.

test/main.sh
```diff
@@ -384,7 +381,29 @@ if ldd "${_LXC}" | grep -F liblxc; then
...
+    chmod +x "${TEST_DIR}"
+
+    # Verify the dir chain is accessible for other users (other's execute bit has to be `x` or `t` (sticky))
+    # This is to catch if `sudo chmod +x ~` was not run and the TEST_DIR is under `~`
+    INACCESSIBLE_DIRS="$(namei -m "${TEST_DIR}" | awk '/^ d/ {if ($1 !~ "^d.*[xt]$") print $2}')"
+    if [ -n "${INACCESSIBLE_DIRS:-}" ]; then
+        echo "Some directories are not accessible by other users" >&2
+        namei -m "${TEST_DIR}"
+        exit 1
+    fi
```
kadinsayani, line 529: It looks like this is duplicated on lines 268-275, perhaps this can be extracted to a helper function as well?

test/main.sh
```diff
@@ -77,6 +77,44 @@ import_subdir_files() {
     done
 }
 
+dependency_checks() {
```
kadinsayani: Suggest using verb_noun naming style (e.g. `check_dependencies`) for consistency.

test/main.sh
```diff
@@ -185,6 +185,29 @@ export _LXC
...
+  # Ignore qemu core dumps (known crasher, to be fixed later)
+  # Enable extended globbing for the !(pattern) syntax
+  shopt -s extglob
+  if compgen -G "/var/crash/core-!(qemu-system-*).gz" > /dev/null 2>&1; then
+    echo "==> CORE: coredumps found"
+    ls -la /var/crash/
+    shopt -u extglob
+    return 1
+  fi
+  shopt -u extglob
```
kadinsayani: Can this be simplified by moving the `shopt -s extglob` into the `if` block? Then we can remove the `shopt-u extglob` that follows the `if` block.

test/main.sh
```diff
@@ -391,25 +467,15 @@ run_test() {
     fi
   fi
 
-  END_TIME=$(date +%s)
-  DURATION=$((END_TIME-START_TIME))
+  END_TIME=$(date +%s.%N)
+  DURATION=$(awk "BEGIN {printf \"%.2f\", ${END_TIME} - ${START_TIME}}")
```
kadinsayani: Nit: Could we have microsecond precision here (`%.6f`)?

test/main.sh
kadinsayani: Ah I see we already have a `check_dependencies` function.

test/includes/test-groups.sh
```diff
@@ -0,0 +1,250 @@
+# Test groups
```
kadinsayani, line 1: Would it be helpful to add a comment indicating that strings added to `test_group_*` vars should match their corresponding test function names (minus the `test_` prefix)?

test/main.sh
simondeziel: I feel like this becomes a bit too noisy in the table summary. It's harder to read and at that point coughing next to your laptop might cause jitter ;)

test/main.sh
simondeziel: Hmm, maybe I'm not seeing straight but I couldn't find the dup. Maybe you looked at a prior rebase?

test/includes/test-groups.sh
simondeziel: Hmm, not sure I understand what you meant here. There is not relation between `test_` functions and `test_group_` array names. We can see this is in the `group:standalone`:

```
readonly test_group_standalone=(
    "alias"
    "apparmor"
    "authorization"
    "basic_usage"
    ...
)
```

tomponline:
Going to hold of merging this until we have a recent TICs run.

Also needs a rebase please

simondeziel:
> Going to hold of merging this until we have a recent TICs run.

You got it reset right in time for the 5th run to have a chance of working, thanks!

> Also needs a rebase please

Done.

test/main.sh
kadinsayani: I checked out the commit locally and there is no dup. The diff alg just isn't showing the original bit as removed.

simondeziel:
@kadinsayani thanks for your review, I think I've addressed/responded to the points you raised. Let me know if I missed anything, please.

test/includes/test-groups.sh
kadinsayani: Is there a relationship between the array strings and test function names?

kadinsayani: Array string `alias` => `test_alias()`, etc.

simondeziel: Ah I see. Yeah, I could add a note that the string them is prepended by `test_` to execute the said function.

kadinsayani: Thanks! It may be helpful for others that haven't reviewed this PR, and it will also help Copilot understand the structure better.

simondeziel: I added a brief note, let me know if unclear.

test/main.sh
```diff
@@ -77,92 +120,109 @@ import_subdir_files() {
     done
 }
 
+dependency_checks() {
```
kadinsayani:
```suggestion
run_dependency_checks() {
```

test/main.sh
```diff
@@ -77,92 +120,109 @@ import_subdir_files() {
...
+  if ! compgen -G "/var/crash/core-*.gz" > /dev/null; then
+    return 0  # No core dumps at all
+  fi
 
-echo "==> Available storage backends: $(available_storage_backends | sort)"
-if [ "$LXD_BACKEND" != "random" ] && ! storage_backend_available "$LXD_BACKEND"; then
-  if [ "${LXD_BACKEND}" = "ceph" ] && [ -z "${LXD_CEPH_CLUSTER:-}" ]; then
-    echo "Ceph storage backend requires that \"LXD_CEPH_CLUSTER\" be set."
-    exit 1
+  # Ignore qemu core dumps (known crasher, to be fixed later)
```
kadinsayani, line 210: Perhaps we can include a `TODO:` or `XXX:` comment here.

simondeziel: Done, thanks

simondeziel:
@tomponline CI all green now.

test/includes/test-groups.sh
```diff
@@ -0,0 +1,256 @@
...
+# individual test function (without the `test_` prefix that is automatically
+# added by `main.sh` when executing the test).
+#
+readonly test_group_cluster=(
+    "clustering_enable"
+    "clustering_edit_configuration"
+    "clustering_membership"
+    "clustering_containers"
+    "clustering_network"
+    "clustering_publish"
```
tomponline, line 13: @simondeziel I think this should be in `test_group_cluster_storage` because it performs publishing of instances from their storage volumes on a storage pool, so is driver specific.

simondeziel: https://github.com/canonical/lxd/pull/17282

simondeziel:
> LGTM thanks!
> 
> Lets see what the impact on TICs is....

It most probably will as I accidentally forgot to add the `group:image` to the CI workflow :/

https://github.com/canonical/lxd/pull/17282 will address that silly mistake.

PR #17075, opened 2025-11-26
storage: Correctly parse Pure Storage gateway URL

MusicDin:
Correctly parse gateway URL, and prevent usage of `/api` in the provided gateway path.

```sh
$ lxc storage create p pure pure.gateway=https://<ip>/pure/api pure.api.token=<token>
Error: Failed to create storage pool "p": Failed to login: Invalid Pure Storage gateway URL "https://<ip>/pure/api": Pure Storage gateway should not include "/api"

# Pure is not accessible on /pure in our case, but resulting URL is correct.
$ lxc storage create p pure pure.gateway=https://<ip>/pure pure.api.token=<token>
Error: Failed to create storage pool "p": Failed to login: Failed to retrieve supported Pure Storage API versions: Failed to retrieve available API versions from Pure Storage: Failed to read response body from "https://<ip>/pure/api/api_version"

$ lxc storage create p pure pure.gateway=https://<ip> pure.api.token=<token>
Storage pool p created
```

### DIFF
--- a/lxd/storage/drivers/driver_pure_util.go
+++ b/lxd/storage/drivers/driver_pure_util.go
@@ -239,15 +239,21 @@ func (p *pureClient) createBodyReader(contents map[string]any) (io.Reader, error
 
 // request issues a HTTP request against the Pure Storage gateway.
 func (p *pureClient) request(method string, url url.URL, reqBody map[string]any, reqHeaders map[string]string, respBody any, respHeaders map[string]string) error {
-	// Extract scheme and host from the gateway URL.
-	scheme, host, found := strings.Cut(p.driver.config["pure.gateway"], "://")
-	if !found {
-		return fmt.Errorf("Invalid Pure Storage gateway URL: %q", p.driver.config["pure.gateway"])
+	gw := p.driver.config["pure.gateway"]
+	if !strings.Contains(gw, "://") {
+		return fmt.Errorf("Invalid Pure Storage gateway URL %q: Missing protocol", gw)
 	}
 
-	// Set request URL scheme and host.
-	url.Scheme = scheme
-	url.Host = host
+	gwURL, err := url.Parse(gw)
+	if err != nil {
+		return fmt.Errorf("Failed to parse Pure Storage gateway URL %q: %w", gw, err)
+	}
+
+	// Ensure the gateway URL does not include the "/api" path, as LXD will handle
+	// the API versioning.
+	if slices.Contains(strings.Split(gwURL.Path, "/"), "api") {
+		return fmt.Errorf(`Invalid Pure Storage gateway URL %q: Pure Storage gateway should not include "/api"`, gw)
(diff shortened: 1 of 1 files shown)


lxd/storage/drivers/driver_pure_util.go
```diff
@@ -239,15 +239,21 @@ func (p *pureClient) createBodyReader(contents map[string]any) (io.Reader, error
...
-		return fmt.Errorf("Invalid Pure Storage gateway URL: %q", p.driver.config["pure.gateway"])
+	gw := p.driver.config["pure.gateway"]
+	if !strings.Contains(gw, "://") {
+		return fmt.Errorf("Invalid Pure Storage gateway URL %q: Missing protocol", gw)
 	}
 
-	// Set request URL scheme and host.
-	url.Scheme = scheme
-	url.Host = host
+	gwURL, err := url.Parse(gw)
```
tomponline, line 247: Why dont we start with this first, so parse the URL from `p.driver.config["pure.gateway"]`. 
If it fails at this point then stop.

Then if it parses successfully you can check whether there is a `url.Scheme` set and avoid the manual `strings.Cut` step.

lxd/storage/drivers/driver_pure_util.go
```diff
@@ -239,15 +239,21 @@ func (p *pureClient) createBodyReader(contents map[string]any) (io.Reader, error
...
+	if err != nil {
+		return fmt.Errorf("Failed to parse Pure Storage gateway URL %q: %w", gw, err)
+	}
+
+	// Ensure the gateway URL does not include the "/api" path, as LXD will handle
+	// the API versioning.
+	if slices.Contains(strings.Split(gwURL.Path, "/"), "api") {
```
tomponline, line 254: Would https://pkg.go.dev/strings#HasSuffix be sufficient here?

MusicDin: Current approach prevents `/api` anywhere in gw URL.
Suffix is not sufficient, as there may be version included (for example`/api/1.2`).

I can tweak to check only last 2 parts. However, since this was more restrictive, and seems very unlikely someone would expose Pure on `/api` I went with this approach.

tomponline: OK lets leave as is ta

lxd/storage/drivers/driver_pure_util.go
MusicDin: I can tweak. The only thing that I don't like about parsing url directly is the case when port is provided, but scheme is not.

```go
url, err := url.Parse("10.10.10.10:443")
if err != nil {
  return err
}
```

Error:
```
first path segment in URL cannot contain colon
```

By checking for the presence of the `://` ensures this case does not occur (more intuitive error is returned).

tomponline (approved):

PR #17078, opened 2025-11-26
devlxd: Ensure security.devlxd flag is checked /1.0 endpoint

MusicDin:
This PR ensures DevLXD endpoint `/1.0` is not accessible if the `security.devlxd` is disabled.

In case of VMs, if DevLXD was previously disabled, when LXD Agent checks connection, the endpoint will now return `Forbidden`, because the flag is not updated in DB until DevLXD server is started by LXD agent. Therefore, to allow checking for `security.devlxd` flag in DevLXD's endpoint `/1.0`, LXD agent needs to consider error `Forbidden` as *successful connection*. The flag will be switched once LXD agent starts the DevLXD server.

Note: DevLXD cannot be live enabled in case the VM is using old LXD agent (prior this update) and the LXD server contains this change. It will require VM restart. 

Related to PR https://github.com/canonical/lxd/pull/15435

### DIFF
--- a/lxd-agent/api_1.0.go
+++ b/lxd-agent/api_1.0.go
@@ -104,12 +104,28 @@ func api10Put(d *Daemon, r *http.Request) response.Response {
 		return response.ErrorResponse(http.StatusInternalServerError, err.Error())
 	}
 
-	server, err := lxd.ConnectLXDHTTP(nil, client)
+	args := &lxd.ConnectionArgs{
+		SkipGetServer: true,
+	}
+
+	server, err := lxd.ConnectLXDHTTP(args, client)
 	if err != nil {
 		return response.ErrorResponse(http.StatusInternalServerError, err.Error())
 	}
 
 	defer server.Disconnect()
+
+	// Now get the server information to verify the connection.
+	//
+	// If DevLXD was previously disabled, this call will fail as forbidden.
+	// This is expected because DevLXD will allow access only once this call
+	// succeeds and the "security.devlxd" flag is set to true.
+	// Therefore, we ignore forbidden error here, as it is still sufficient to
+	// confirm the DevLXD is actually accessible.
+	_, _, err = server.GetServer()
+	if err != nil && !api.StatusErrorCheck(err, http.StatusForbidden) {
+		return response.SmartError(fmt.Errorf("Failed to verify connection to LXD server: %w", err))
+	}
 
 	// Let LXD know, we were able to connect successfully.
 	d.chConnected <- struct{}{}


MusicDin:
Sorry, there is bad commit message, will fix

tomponline:
> because the flag is not updated in DB until DevLXD server is started by LXD agent

Please could you explain this a bit more? Could we change that order to make it work?

MusicDin:
> > because the flag is not updated in DB until DevLXD server is started by LXD agent

> Please could you explain this a bit more? Could we change that order to make it work?

The instance `Update()` (in `lxd/instance/drivers/driver_qemu.go`) advertises the vsock address to LXD agent by calling its `PUT /1.0` endpoint. If it succeeds, then the instance update proceeds and updates instance config in DB.

Since the `PUT /1.0` is called before instance config is updated in DB, when LXD agent reaches out to LXD, the DevLXD loads the instance with old config, where `security.devlxd` is still false.

---

We could potentially first update the instance, and then call the DevLXD. Or we could even call DevLXD's endpoint `/` (which returns `1.0` and is not protected)?

tomponline:
> Or we could even call DevLXD's endpoint `/` (which returns `1.0` and is not protected)?

You mean do this first, then update the DB, then push the vsock address to the agent?

MusicDin:
> > Or we could even call DevLXD's endpoint / (which returns 1.0 and is not protected)?

> You mean do this first, then update the DB, then push the vsock address to the agent?

No, the `/` endpoint is not protected (does not check the flags). It returns API version.

```go
var apiDevLXD = []APIEndpoint{
	{
		Path: "/",
		Get: APIEndpointAction{
			Handler: func(d *Daemon, r *http.Request) response.Response {
				return response.DevLXDResponse(http.StatusOK, []string{"/1.0"}, "json")
			},
			AllowUntrusted: true,
		},
	},
	...
}
```


So, we can use that approach.
Or we can protect the `/` endpoint as well, and keep the current approach where `Forbidden` is consider successful connection?

Do you have a suggestion to address this cleanly?

tomponline:
> > > Or we could even call DevLXD's endpoint / (which returns 1.0 and is not protected)?
> 
> > You mean do this first, then update the DB, then push the vsock address to the agent?
> 
> No, the `/` endpoint is not protected (does not check the flags). It returns API version.
> 
> ```go
> var apiDevLXD = []APIEndpoint{
> 	{
> 		Path: "/",
> 		Get: APIEndpointAction{
> 			Handler: func(d *Daemon, r *http.Request) response.Response {
> 				return response.DevLXDResponse(http.StatusOK, []string{"/1.0"}, "json")
> 			},
> 			AllowUntrusted: true,
> 		},
> 	},
> 	...
> }
> ```

Right, but im not following you, lets discuss.

MusicDin:
As discussed, will add check for `security.devlxd` on DevLXD's `/` endpoint as well, and consider the connection successful if `Forbidden` is returned.

PR #17080, opened 2025-11-26
Backports (stable-5.21)

tomponline:


tomponline:
@markylaing this auth stuff is ok to backport iirc we previously agreed with @escabo right?

kadinsayani (approved):

PR #17082, opened 2025-11-26
storage: Do not call host updated in Alletra as it is not supported

MusicDin:
UpdateHost always returns an error, and linter complains that `if err != nil` always succeeds.
Therefore, just return an error immediately


tomponline:
@MusicDin how are you testing this?

PR #17089, opened 2025-11-27
Network: Begin transition to libovsdb (from Incus)

nmezhenskyi:
This PR begins a series of changes and cherry-picks to transition LXD to use [libovsdb](https://github.com/ovn-kubernetes/libovsdb), rather than OVN CLI tooling. It continues the work started [here](https://github.com/canonical/lxd/pull/14841).

This PR primarily lays the groundwork needed for migration of our OVN functionality to `libovsdb` by restructuring the `openvswitch` package and adding `libovsdb` clients for Northbound and Southbound databases. The follow-up PR(s) will focus on porting the LXD's OVN network creation functionality to fully rely on `libovsdb` instead of OVN CLI tools.

Includes cherry-picks and inspirations from https://github.com/lxc/incus/pull/319, https://github.com/lxc/incus/pull/325, https://github.com/lxc/incus/pull/688, https://github.com/lxc/incus/pull/1349, https://github.com/canonical/lxd/pull/14841.

**Changes:**
- Split functionality from the `lxd/network/openvswitch` package into two separate packages `lxd/network/ovs` and `lxd/network/ovn`.
- Added generated schemas necessary for `libovsdb` usage.
- Added OVN Southbound database client `type SB struct` (to replace CLI calls to `ovn-sbctl`).
- Added OVN Northbound database client `type NB struct` (to replace CLI calls to `ovn-nbctl`).
- Ported `GetLogicalRouterPortActiveChassisHostname` to use `libovsdb` instead of `ovn-sbctl`.

### DIFF
--- a/lxd/network/driver_ovn.go
+++ b/lxd/network/driver_ovn.go
@@ -2434,19 +2434,13 @@ func (n *ovn) setup(update bool) error {
 		}
 
 		// Create external router port.
-		err = client.LogicalRouterPortAdd(n.getRouterName(), n.getRouterExtPortName(), routerMAC, bridgeMTU, extRouterIPs, update)
+		err = client.LogicalRouterPortAdd(n.getRouterName(), n.getRouterExtPortName(), routerMAC, bridgeMTU, extRouterIPs, n.getChassisGroupName(), update)
 		if err != nil {
 			return fmt.Errorf("Failed adding external router port: %w", err)
 		}
 
 		if !update {
 			revert.Add(func() { _ = client.LogicalRouterPortDelete(n.getRouterExtPortName()) })
-		}
-
-		// Associate external router port to chassis group.
-		err = client.LogicalRouterPortLinkChassisGroup(n.getRouterExtPortName(), n.getChassisGroupName())
-		if err != nil {
-			return fmt.Errorf("Failed linking external router port to chassis group: %w", err)
 		}
 
 		// Create external switch port and link to router port.
@@ -2668,7 +2662,7 @@ func (n *ovn) setup(update bool) error {
 	}
 
 	// Create internal router port.
-	err = client.LogicalRouterPortAdd(n.getRouterName(), n.getRouterIntPortName(), routerMAC, bridgeMTU, intRouterIPs, update)
+	err = client.LogicalRouterPortAdd(n.getRouterName(), n.getRouterIntPortName(), routerMAC, bridgeMTU, intRouterIPs, "", update)
 	if err != nil {
 		return fmt.Errorf("Failed adding internal router port: %w", err)
 	}


lxd/network/ovn/ovn_sb.go
```diff
@@ -0,0 +1,163 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		clientCAPool := x509.NewCertPool()
+		clientCAPool.AddCert(tlsCAcert)
+
+		tlsConfig := &tls.Config{
+			Certificates:       []tls.Certificate{clientCert},
+			InsecureSkipVerify: true,
```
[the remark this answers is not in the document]
nmezhenskyi: I will need to test and reevaluate this. Incus is doing custom verification for OVN TLS certs, but their implementation seems incorrect. I guess the reason they're using custom verification is that `dbAddr` can have multiple endpoints, so each connection attempt would need the certificate to match that specific endpoint. Or perhaps this is done to avoid hostname checking. Either way, I've removed the custom verification for now. I'll test whether it works with the standard TLS verification, and if not, what's the reason.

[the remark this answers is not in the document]
nmezhenskyi: Standard TLS verification does not work indeed

lxd/network/ovn/ovn_nb.go
```diff
@@ -0,0 +1,237 @@
...
+
+		tlsCAcert.IsCA = true
+		tlsCAcert.KeyUsage = x509.KeyUsageCertSign
+
+		clientCAPool := x509.NewCertPool()
+		clientCAPool.AddCert(tlsCAcert)
+
+		tlsConfig := &tls.Config{
+			Certificates:       []tls.Certificate{clientCert},
+			InsecureSkipVerify: true, // Don't use the default TLS verification.
```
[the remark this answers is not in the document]
nmezhenskyi: This should be safe because below we define custom TLS verification function in `VerifyPeerCertificate`.

lxd/network/ovn/ovn_sb.go
```diff
@@ -0,0 +1,225 @@
...
+
+		tlsCAcert.IsCA = true
+		tlsCAcert.KeyUsage = x509.KeyUsageCertSign
+
+		clientCAPool := x509.NewCertPool()
+		clientCAPool.AddCert(tlsCAcert)
+
+		tlsConfig := &tls.Config{
+			Certificates:       []tls.Certificate{clientCert},
+			InsecureSkipVerify: true, // Don't use the default TLS verification.
```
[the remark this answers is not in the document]
nmezhenskyi: This should be safe because below we define custom TLS verification function in `VerifyPeerCertificate`.

nmezhenskyi:
This should now be ready for the initial review.

Some notes and items for discussion:
- Code scanning errors about the TLS verification.
- The addition of `network.ovs.connection` as a global config key instead of making it node-specific.
- The usage of new `network.ovs.connection` setting in the `spawn_lxd()` helper in our test suite and the side-effects of that for some tests (`completions` and `init_dump`).
- Currently each call to OVN/OVS functions requires establishing a new connection to NB/SB/OVS databases. Should we consider using long-living connections for the `libovsdb` clients as part of LXD state perhaps?

Also, there are some cherry-picks from the [original PR](https://github.com/canonical/lxd/pull/14841) I didn't include in here to make this PR more manageable to review. Those are mostly ports of some functions that use `ovn-nbctl` to use `libovsdb` client, so I thought it might be better to add them in a follow-up PR.

Makefile
```diff
@@ -296,7 +296,9 @@ endif
 
 .PHONY: update-ovsdb
 update-ovsdb:
-	go install github.com/ovn-kubernetes/libovsdb/cmd/modelgen@main
+ifeq ($(shell command -v modelgen),)
+	(cd / ; go install github.com/ovn-kubernetes/libovsdb/cmd/modelgen@main)
```
markylaing: Is the `cd /` necessary?

test/suites/network_ovn.sh
```diff
@@ -58,6 +58,10 @@ test_network_ovn() {
 
   setup_ovn
 
+  # This assertions tests that the "BridgeExists" function works correctly and can find the integration bridge.
```
markylaing: `s/assertions/assertion/`

lxd/network/network_utils_bridge.go
```diff
@@ -46,15 +47,15 @@ func IsNativeBridge(bridgeName string) bool {
 }
 
 // AttachInterface attaches an interface to a bridge.
-func AttachInterface(bridgeName string, devName string) error {
+func AttachInterface(s *state.State, bridgeName string, devName string) error {
```
markylaing: I don't think we should pass `state.State` here just to use the OVS connection address.

Can we please just pass the address?

lxd/network/network_utils_bridge.go
```diff
@@ -69,15 +70,15 @@ func AttachInterface(bridgeName string, devName string) error {
 }
 
 // DetachInterface detaches an interface from a bridge.
-func DetachInterface(bridgeName string, devName string) error {
+func DetachInterface(s *state.State, bridgeName string, devName string) error {
```
markylaing: Same here, I'd prefer not to pass state around if at all possible.

test/suites/completions.sh
```diff
@@ -169,10 +169,10 @@ test_completions() {
     [ "$(complete config get localhost:c1 '')" = 'boot.,cloud-init.,cluster.,environment.,limits.,linux.,migration.,nvidia.,placement.,raw.,security.,snapshots.,ubuntu_pro.,user.' ]
     [ "$(complete config get c1 limits.)" = 'limits.cpu,limits.cpu.,limits.disk.,limits.hugepages.,limits.kernel.,limits.memory,limits.memory.,limits.processes' ]
     lxc config set user.foo=bar
-    [ "$(complete config unset '')" = 'c1,c2,core.https_address,localhost:,user.foo' ]
+    [ "$(complete config unset '')" = 'c1,c2,core.https_address,localhost:,network.ovs.connection,user.foo' ]
     [ "$(complete config unset c)" = 'c1,c2,core.https_address' ]
     [ "$(complete config unset l)" = 'localhost:' ]
-    [ "$(complete config unset localhost: '')" = 'core.https_address,user.foo' ]
+    [ "$(complete config unset localhost: '')" = 'core.https_address,network.ovs.connection,user.foo' ]
```
markylaing: Won't this config key only be set if `LXD_OVS_CONNECTION` is set in the environment?

test/suites/init_dump.sh
```diff
@@ -50,6 +51,7 @@ cat <<EOF > expected.yaml
 config:
   core.https_address: 127.0.0.1:9999
   images.auto_update_interval: "15"
+  network.ovs.connection: unix:///var/snap/microovn/common/run/switch/db.sock
```
markylaing: Does this fail when running locally without LXD_OVS_CONNECTION set to the above value?

lxd/network/ovn/ovn_sb.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+
+		clientCAPool := x509.NewCertPool()
+		clientCAPool.AddCert(tlsCAcert)
+
+		tlsConfig := &tls.Config{
+			Certificates:       []tls.Certificate{clientCert},
+			InsecureSkipVerify: true, // Don't use the default TLS verification.
+
+			// We use custom TLS verification here to skip the hostname verification.
```
markylaing, line 95: Perhaps I've missed this but why do we need to skip hostname verification?

lxd/network/ovn/ovn_sb.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	// Create the client struct.
+	client := &SB{
+		client: ovn,
+		cookie: monitorCookie,
+	}
+
+	// Set finalizer to stop the monitor.
+	runtime.SetFinalizer(client, func(o *SB) {
```
markylaing, line 164: Will this finalizer stuff go away eventually?

lxd/network/ovn/ovn_sb_events.go
```diff
@@ -0,0 +1,36 @@
...
+	defer sbEventHandlersMu.Unlock()
+
+	if sbEventHandlers == nil {
+		return nil
+	}
+
+	delete(sbEventHandlers, name)
+
+	return nil
+}
```
markylaing: There are a lot of things in this commit that I'm not a fan of:
1. String literals in event hooks.
2. Handling of event hooks being performed at package level.
3. There seems to be a lot of unnecessary wrapping. The library already has a `cache.EventHandler` interface that could be implemented, but instead it looks like package level hooks are shoehorned into the default `cache.EventHandlerFuncs` implementation.

For now we should probably keep this as it is for the sake of making the cherry-picking easier, but we should revisit this.

markylaing (changes requested):
Overall looking great. I have a few questions and nitpicks to go through. Thanks :)

Makefile
nmezhenskyi: Good catch, it's not needed anymore with recent Go versions. I've removed it.

test/suites/network_ovn.sh
nmezhenskyi: Fixed

lxd/network/network_utils_bridge.go
nmezhenskyi: Good point, I agree

test/suites/completions.sh
nmezhenskyi: > Won't this config key only be set if `LXD_OVS_CONNECTION` is set in the environment?

It would be yes.

I've changed the `spawn_lxd()` helper function to always set `network.ovs.connection` if `LXD_OVS_CONNECTION` is provided. However, I've also updated the GitHub workflow to set this environment variable `LXD_OVS_CONNECTION` to `"unix:///var/snap/microovn/common/run/switch/db.sock"` for the test suite.

So now, each time `spawn_lxd()` is called anywhere in the test suite, it sets the `network.ovs.connection`. The `spawn_lxd()` helper gets called earlier in the `completions` script, and as a side-effect it sets this config option. This is why I've updated the completions test to account for that. Otherwise, it fails.

nmezhenskyi: I've also tried not setting `network.ovs.connection` in `spawn_lxd()`, and instead do that only in the tests that require it. But it turns out we rely on it quite a lot, including in the `completions` test where it queries the list of networks.

tomponline:
> * Added new API extension `network_ovs_connection` which introduces `network.ovs.connection` global configuration key for LXD. This option allows setting a custom connection string to connect to OVS database (i.e., when using MicroOVN).

Why do we need this?

test/suites/init_dump.sh
nmezhenskyi: > Does this fail when running locally without LXD_OVS_CONNECTION set to the above value?

Yes, it does.

I should probably append it conditionally based on `LXD_OVS_CONNECTION` value. What do you think?

nmezhenskyi:
> Added new API extension `network_ovs_connection` ...
> Why do we need this?

@tomponline  This is because Open VSwitch can use different locations for its database socket. For example, if MicroOVN snap is installed, it uses `unix:/var/snap/microovn/common/run/switch/db.sock` instead of the `unix:/run/openvswitch/db.sock`.

Previously, `ovs-vsctl` tool could figure out which socket path to use. But now since we're using `libovsdb`, we need to specify the correct path explicitly.

lxd/network/ovn/ovn_sb.go
nmezhenskyi: > Will this finalizer stuff go away eventually?

@markylaing Can you expand on what do you mean by it? Do you mean we should get rid of this `runtime.SetFinalizer()` function call?

lxd/network/ovn/ovn_sb_events.go
nmezhenskyi: I agree. I'll see what I can do with it for now. I can at least change it to use constants instead of string literals.

tomponline:
> > Added new API extension `network_ovs_connection` ...
> > Why do we need this?
> 
> @tomponline This is because Open VSwitch can use different locations for its database socket. For example, if MicroOVN snap is installed, it uses `unix:/var/snap/microovn/common/run/switch/db.sock` instead of the `unix:/run/openvswitch/db.sock`.
> 
> Previously, `ovs-vsctl` tool could figure out which socket path to use. But now since we're using `libovsdb`, we need to specify the correct path explicitly.

Please can you book a meeting to dicuss the approach. I did not expect an internal implementation change to affect the external API, so something looks off here, but I may have just misunderstood. Thanks

tomponline:
@nmezhenskyi shall we close this for now?

nmezhenskyi:
@tomponline Yes

PR #17091, opened 2025-11-27
lxd: Require all cluster member to process forced project delete notifications

kadinsayani:


tomponline (approved):

markylaing:
This PR has just made me think. With @MusicDin changing the client methods to return an operation, is it worth changing project deletion to be an operation at the same time? Just because force deletion could take a while for a large project.

tomponline:
> at the same time

That time has already passed im afraid.

I am not against it in principle but we must avoid the drip-drip-drip breaking of the API every release (we saw how disruptive it was even before pushing to stable as it required new releases of pylxd and terraform).

tomponline:
I think if we are going to do this we should bite the bullet and make *all* mutating API endpoints use operations in one go.

PR #17092, opened 2025-11-27
Fix inefficiency in instance deleteCommon()

kadinsayani:


kadinsayani:
> IIRC the reason the original parent load was moved up was so we did all DB checks before starting on-disk operations, so suggest we keep it there, but make it conditional.

Ah yes, that's right. Fixed.

lxd/instance/drivers/driver_common.go
```diff
@@ -752,13 +752,6 @@ func (d *common) deleteAttachedVolumeSnapshots(snapInst instance.Instance, diskV
 // - Attached volume snapshot deletion for snapshots (if diskVolumesMode is "all-exclusive").
 // - Parent backup file update for snapshots.
 func (d *common) deleteCommon(inst instance.Instance, force bool, diskVolumesMode string) error {
-	unlock, err := d.updateBackupFileLock(context.Background())
```
tomponline, line 755: Lets keep this here and make it conditional so we dont delete the instance before taking the lock.

tomponline: This seems to have moved down

kadinsayani: Yes but it's before we delete the instance. It's bundled in when we check if the instance is a snapshot before loading the parent.

tomponline: I think I'd be tempted to keep it before the operation lock so its consistent with the ordering of other actions, like `instance.Update()`.

kadinsayani: Sounds good, fixed.

tomponline:
tests sad

tomponline (approved):
LGTM!

Issue #16812, opened 2025-10-24
Fail to configure OIDC authentication if discovery fails


markylaing:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

A user may configure OIDC authentication via:
```
lxc config set oidc.issuer=https://idp.example.com oidc.client.id=12345
```

but LXD doesn't perform discovery until a client attempts to authenticate via OIDC (either UI or CLI). 

This means that LXD may not be able to connect to the issuer when OIDC is initially configured. Connection issues could be:
- Invalid issuer URL
- Invalid issuer certificate
- Issuer certificate not trusted
- Invalid proxy settings

See https://discourse.ubuntu.com/t/add-custom-ca-certs-to-lxd-for-openid-trusts/68749/4 for an example of a user facing these issues.

### Describe the solution you'd like

On a call to `NewVerifier`, the verifier should attempt to perform OIDC discovery at the issuer URL and gracefully fail if connection fails.

### Describe alternatives you've considered

N/A

kwonkwonn:
hey @markylaing 

I would like to work on this issue.

My plan is to update NewVerifier to perform OIDC discovery (likely by initializing the provider) and return a proper error if the issuer is unreachable or invalid. 

Could you please assign this issue to me?

markylaing:
Thanks for picking this up! Yes this should be a case of caller `(*Verifier).ensureConfig` once all of it's fields are set. It might be necessary to wrap the returned error in an `api.StatusError` so that a user friendly error is reported back to the caller.

kwonkwonn:
Hi @markylaing  ,

Thanks for the tip on (*Verifier).ensureConfig. I'll definitely use that.

While looking into where to hook this validation, I noticed a timing issue in `doAPI10Update`. Currently, the triggers run after the in-memory config (`d.globalConfig`) is updated.  (https://github.com/canonical/lxd/blob/a64535d78fbdf86c4a1e88df4d0f42518903760c/lxd/api_1.0.go#L1019-L1030).

I ran a quick test, and if the validation fails inside the trigger, the DB reverts seems valid, but the in-memory config doesn't. This leaves the daemon showing the invalid OIDC config even though the request failed.

To avoid this "split state," I plan to place the validation at the very beginning of doAPI10Update (similar to validateOIDCSessionExpiry). This way, we check the issuer URL before touching any state.

[...]

PR #17098, opened 2025-11-28
lxd: Validate OIDC issuer configuration during update

kwonkwonn:
## Checklist

Fixes: #16812

To maintain the existing code structure, I've used the current `Verifier` and `(*verifier).ensureConfig` methods. While I recognize that using `Verifier` solely for issuer discovery has some drawbacks—such as heavy dependencies and unnecessary overhead from database session handling and cookie management—I've deferred refactoring this approach pending the outcome of discussion in #17087.

I've identified that the `Discover` function from the `zitadel/oidc` package (which we already use) efficiently meets our requirements for issuer validation. If the maintainers prefer, I can modify the `validateOIDC` function to use `Discover` directly instead.

I've performed manual testing for OIDC issuer configuration. Due to the complexity of mocking structures like `*daemon` and `clusterConfig`, I've postponed adding automated tests pending guidance from the maintainers.

Please let me know if you have any concerns or if you'd like me to modify this approach.

Thank you!

### DIFF
--- a/test/suites/oidc.sh
+++ b/test/suites/oidc.sh
@@ -12,7 +12,22 @@ test_oidc() {
 
   # Setup OIDC
   spawn_oidc
-  lxc config set "oidc.issuer=http://127.0.0.1:$(< "${TEST_DIR}/oidc.port")/" "oidc.client.id=device"
+
+  # Ensure a clean state before testing validation & rollback
+  lxc config unset oidc.issuer
+  lxc config unset oidc.client.id
+
+  # Should be failed on wrong issuer.
+  ! lxc config set oidc.issuer="http://127.0.0.1:22/" oidc.client.id="device" || false # Wrong port
+  ! lxc config set "oidc.issuer=http://127.0.0.1:$(< "${TEST_DIR}/oidc.port")/wrong-path" "oidc.client.id=device" || false # Invalid path
+  ! lxc config set "oidc.issuer=https://idp.example.com/" "oidc.client.id=device" || false # Invalid host
+
+
+  # Should remain empty as above tests failed.
+  [ -z "$(lxc config get oidc.client.id || echo fail)" ]
+  [ -z "$(lxc config get oidc.issuer || echo fail)" ]
+
+  lxc config set "oidc.issuer=http://127.0.0.1:$(< "${TEST_DIR}/oidc.port")/" "oidc.client.id=device" # Valid Configuration
 
   # Expect this to fail. No user set.
   ! BROWSER=curl lxc remote add --accept-certificate oidc "${LXD_ADDR}" --auth-type oidc || false


lxd/api_1.0.go
```diff
@@ -780,6 +780,36 @@ func validateOIDCSessionExpiry(config *clusterConfig.Config, requestConfig map[s
...
+			return fmt.Errorf("Failed to generate OIDC verifier: %w", err)
+		}
+		err = verifier.Validate(r.Context(), r.Host)
+		if err != nil {
+			return fmt.Errorf("Failed to validate OIDC configuration: %w", err)
+		}
+	}
+	return nil
+}
+
```
markylaing: Can we please reverse the boolean logic here to be:
```
if not oidc changed {
    return nil
}

err = validate(config)
if err != nil {
    return wrapped err
}

return nil
```

lxd/auth/oidc/oidc.go
```diff
@@ -791,3 +791,7 @@ func NewVerifier(issuer string, clientID string, clientSecret string, scopes []s
 
 	return verifier, nil
 }
+
+func (o *Verifier) Validate(ctx context.Context, host string) error {
```
markylaing: Can this have a Godoc style comment please? Thanks.

lxd/api_1.0.go
```diff
@@ -989,6 +1019,11 @@ func doAPI10Update(d *Daemon, r *http.Request, req api.ServerPut, patch bool) re
 		}
 	})
 
+	err = validateOIDC(d, r, newClusterConfig, clusterChanged)
```
markylaing: Could you please add a comment that says *why* we need to call this validation logic here and not inside the "apply triggers" function? Thanks.

markylaing (changes requested):
Just a few nit picks. Ultimately we'll want to remove this logic and have `oidc.NewVerifier` call `ensureConfig` after initialisation, but as mentioned this currently leads to the situation in #17087

lxd/api_1.0.go
kwonkwonn: i've reversed boolean logic to escape as early as possible when oidc's configuration are not chaged.

kwonkwonn:
I've been thinking about the best approach here. As discussed, the current implementation requires "pre-validation" logic to avoid the memory corruption issue described in #17087. This makes the code unnecessarily complex and disconnected from the actual triggers.

This plan sounds resonable:
1. put this PR on hold for a moment.
2. PR to fix #17087, I plan to add a revert hook immediately after the in-memory replacement in `doAPI10Update`. It would look something like this:
```
revert.Add(func() {
	d.globalConfigMu.Lock()
	d.globalConfig = oldclusterconfig
    d.localConfig= oldnodeconfig
    d.globalConfigMu.Unlock()
}
``` 
This will effectively prevent the split-brain state when triggers fail.

3. Once the fix is merged, I will refactor this PR to perform OIDC validation naturally within the trigger logic.

This sequence will make the OIDC implementation much cleaner and safer, relying on the core rollback mechanism rather than ad-hoc per-field checking.

Does this roadmap look good to you? If so, I'll open the PR for #17087 shortly.

markylaing:
> I've been thinking about the best approach here. As discussed, the current implementation requires "pre-validation" logic to avoid the memory corruption issue described in #17087. This makes the code unnecessarily complex and disconnected from the actual triggers.
> 
> This plan sounds resonable:
> 
> 1. put this PR on hold for a moment.
> 2. PR to fix [In-memory configuration state corruption on trigger failure in doAPI10Update #17087](https://github.com/canonical/lxd/issues/17087), I plan to add a revert hook immediately after the in-memory replacement in `doAPI10Update`. It would look something like this:
> 
> ```
> revert.Add(func() {
> 	d.globalConfigMu.Lock()
> 	d.globalConfig = oldclusterconfig
>     d.localConfig= oldnodeconfig
>     d.globalConfigMu.Unlock()
> }
> ```
> 
> This will effectively prevent the split-brain state when triggers fail.
> 
> 3. Once the fix is merged, I will refactor this PR to perform OIDC validation naturally within the trigger logic.
> 
> This sequence will make the OIDC implementation much cleaner and safer, relying on the core rollback mechanism rather than ad-hoc per-field checking.
> 
> Does this roadmap look good to you? If so, I'll open the PR for #17087 shortly.

Sounds great to me. Much better to fix underlying issues rather than papering over them!

kwonkwonn:
I've opened #17122 to address this as discussed.
Please take a look when you have a moment. Thanks!!

kwonkwonn:
I've drafted the test cases. Unfortunately, I'm facing some environment-specific issues on my local GCP instance, which prevents me from running the full test suite locally.

I'm pushing this to check the results against the official CI. 
Please let me know if you spot any obvious issues in the test logic while the CI runs.

lxd/auth/oidc/oidc.go
```diff
@@ -791,3 +792,20 @@ func NewVerifier(issuer string, clientID string, clientSecret string, scopes []s
...
+	}
+
+	// As this function concentrates on validating the configuration, we do not store the discovered configuration.
+	_, err = client.Discover(ctx, o.issuer, httpClient)
+	if err != nil {
+		return fmt.Errorf("Failed to discover OIDC configuration: %w", err)
+	}
+
+	return nil
+}
```
markylaing, line 799: The zitadel library only needs to perform discovery once. After that, the JWKs are stored in a "remote key set" which will automatically pull new ones from the issuer as needed.

Because of this, I think it's wasteful to perform discovery more than once. I think we can remove this function and simply have `NewVerifier` call `ensureConfig` after the fields are instantiated - returning any error.

test/suites/oidc.sh
```diff
@@ -12,7 +12,21 @@ test_oidc() {
...
+  lxc config unset oidc.client.id
+
+  # Should be failed on wrong issuer.
+  ! lxc config set oidc.issuer="http://127.0.0.1:22/" oidc.client.id="device" || false  # Wrong port 
+  ! lxc config set "oidc.issuer=http://127.0.0.1:$(< "${TEST_DIR}/oidc.port")/wrong-path" "oidc.client.id=device" || false  # Invalid path 
+  ! lxc config set "oidc.issuer=https://idp.example.com/" "oidc.client.id=device" || false  # Invalid host 
+  
+  # Should remain empty as above tests failed. 
+  [ -z "$(lxc config get oidc.client.id || true)" ] 
+  [ -z "$(lxc config get oidc.issuer || true)" ]
```
markylaing: We have a custom linter that is failing on these lines. What you're asserting is that the output of the command is empty, but this doesn't check that the command itself succeeded. We've been bitten by this before!

The fix is to change these lines as follows:
```suggestion
[ -z "$(lxc config get oidc.client.id || echo fail)" ] 
[ -z "$(lxc config get oidc.issuer || echo fail)" ] 
```

This way, if the command fails "fail" will be outputted and the `-z` test will fail. When the command succeeds, `-z` will assert that the output is empty.

This is all because we can't force bash to fail when it encounters a non-zero exit code within a command substitution.

lxd/auth/oidc/oidc.go
```diff
@@ -789,5 +789,11 @@ func NewVerifier(issuer string, clientID string, clientSecret string, scopes []s
 		sessionHandler: sessionHandler,
 	}
 
+	// Ensure configuration is valid with dummy host.
+	err := verifier.ensureConfig(ctx, "localhost")
+	if err != nil {
+		return nil, fmt.Errorf("Failed to ensure new verifier's configuration. %w", err)
+	}
```
markylaing, line 796:
```suggestion
	if err != nil {
		return nil, fmt.Errorf("Failed to ensure new verifier's configuration: %w", err)
	}
```

lxd/auth/oidc/oidc.go
```diff
@@ -789,5 +789,11 @@ func NewVerifier(issuer string, clientID string, clientSecret string, scopes []s
 		sessionHandler: sessionHandler,
 	}
 
+	// Ensure configuration is valid with dummy host.
+	err := verifier.ensureConfig(ctx, "localhost")
```
markylaing: I think we can improve this a little by using the network address of the LXD daemon. Could you please add another new `networkAddress string` argument to `NewVerifier`. The value will be `(*Daemon).endpoints.NetworkAddress()`.

lxd/daemon.go
```diff
@@ -1758,7 +1758,7 @@ func (d *Daemon) init() error {
 		}
 
 		sessionHandler := dbOIDC.NewSessionHandler(d.db.Cluster, d.events, d.globalConfig.OIDCSessionExpiry)
-		d.oidcVerifier, err = oidc.NewVerifier(oidcIssuer, oidcClientID, oidcClientSecret, oidcScopes, oidcAudience, oidcGroupsClaim, d.globalConfig.ClusterUUID(), d.getCoreAuthSecrets, httpClientFunc, sessionHandler)
+		d.oidcVerifier, err = oidc.NewVerifier(d.shutdownCtx, oidcIssuer, oidcClientID, oidcClientSecret, oidcScopes, oidcAudience, oidcGroupsClaim, d.globalConfig.ClusterUUID(), d.getCoreAuthSecrets, httpClientFunc, sessionHandler)
 		if err != nil {
 			return err
```
markylaing: Now that it is possible for `NewVerifier` to return an error. I wonder if we really should be aborting `Daemon.init` if OIDC configuration is incorrect. 

While it is no longer possible to configure OIDC incorrectly, it is possible that the IdP is down at start up time.

I think it is best that we change this for now to log a warning instead.

lxd/auth/oidc/oidc.go
kwonkwonn: Thanks for the catch!
Fixed to follow overall error structure.

kwonkwonn:
> Think we're getting very close now :)
> 
> There are just a few small changes to be made.
> 
> In addition, please could you add one commit per package and per functional change? I suggest:
> 
> * `lxd/auth/oidc: Verify configuration on initialisation.`
> * `lxd: Update calls to NewVerifier`
> * `lxd: Do not abort daemon init on bad OIDC config.`
> * `test/suites: Add tests for OIDC configuration validation.`

Thanks for the detailed review! 

I’ve split the work into functional commits as requested.

One note regarding the commit structure: I combined the first two suggestions (lxd/auth/oidc logic and the NewVerifier updates) into a single atomic commit.
Since the signature change in NewVerifier affects callers across different files, keeping them together ensures the build passes.

If you would still prefer them separated, please let me know and I will adjust accordingly.

lxd/api_1.0.go
```diff
@@ -1280,7 +1280,7 @@ func doAPI10UpdateTriggers(d *Daemon, nodeChanged, clusterChanged map[string]str
 			}
 
 			sessionHandler := dbOIDC.NewSessionHandler(d.db.Cluster, d.events, d.globalConfig.OIDCSessionExpiry)
-			d.oidcVerifier, err = oidc.NewVerifier(oidcIssuer, oidcClientID, oidcClientSecret, oidcScopes, oidcAudience, oidcGroupsClaim, d.globalConfig.ClusterUUID(), s.CoreAuthSecrets, httpClientFunc, sessionHandler)
+			d.oidcVerifier, err = oidc.NewVerifier(s.ShutdownCtx, oidcIssuer, oidcClientID, oidcClientSecret, oidcScopes, oidcAudience, oidcGroupsClaim, d.endpoints.NetworkAddress(), d.globalConfig.ClusterUUID(), s.CoreAuthSecrets, httpClientFunc, sessionHandler)
```
markylaing: The arguments are out of order here, this is causing the tests to fail

markylaing (changes requested):
I think just fix the argument order on the call to `NewVerifier` in `doAPI10UpdateTriggers` and we're good to go. Thanks.

lxd/api_1.0.go
kwonkwonn: My bad, thanks for catching that!
I've corrected the argument order and pushed the fix.

markylaing (approved):
Great work, thank you! :pray:

tomponline (approved):

tomponline:

Please can you sign your commits.

markylaing:
> > Commits must have verified signatures.
> 
> Please can you sign your commits.

Ah I missed that only the first commit was signed

kwonkwonn:
> > > Commits must have verified signatures.
> > 
> > 
> > Please can you sign your commits.
> 
> Ah I missed that only the first commit was signed

Apologies for the oversight.
I've re-signed all commits and force-pushed.

PR #17101, opened 2025-11-28
test/devlxd: Sync FS before creating a snapshot

MusicDin:


MusicDin:
Rejecting in favor of processing LXD agent events synchronously.

PR #17105, opened 2025-11-28
Add dqlite developer documentation and improve dqlite related code comments

kadinsayani:
Adds dqlite developer documentation and comments to Raft and Cluster role constants. I'll continue improving upon the documentation/comments while working on cluster member evacuation improvements.

### DIFF
--- a/doc/reference/dqlite-internals.md
+++ b/doc/reference/dqlite-internals.md
@@ -1,0 +1,33 @@
+---
+relatedlinks: "[Canonical&#32;Dqlite](https://canonical.com/dqlite), [Dqlite&#32;GitHub](https://github.com/canonical/dqlite)"
+---
+
+(dqlite-internals)=
+# Dqlite internals
+
+Dqlite (distributed SQLite) implements a replicated SQLite database by combining the SQLite engine with a Raft-based consensus layer. Each LXD daemon (cluster member) runs a Dqlite node which exposes a SQLite-like API backed by a Raft replicated state machine. A single leader handles writes; followers apply replicated log entries and serve reads depending on configuration.
+
+## Raft
+
+[Raft](https://raft.github.io/) is a consensus algorithm that ensures a cluster of nodes can agree on a sequence of state machine commands even in the presence of failures. Raft handles leader election, log replication, safety, and membership changes.
+
+## Dqlite Raft implementation
+
+Raft nodes in Dqlite move between four runtime states: `RAFT_UNAVAILABLE`, `RAFT_FOLLOWER`, `RAFT_CANDIDATE` and `RAFT_LEADER`. Followers are passive: they accept `AppendEntries` RPCs (remote procedure calls) from an active leader and reset an election timer; when a follower's randomized election timeout elapses without leader contact it becomes a candidate, increments its term and sends `RequestVote` RPCs to gather votes. A candidate becomes leader after receiving votes from a majority of voting servers and then starts replicating log entries to followers using `AppendEntries` (heartbeats are empty `AppendEntries` used to maintain authority).
+
+The election timeout is randomized, but always shorter than the heartbeat interval. Features such as pre-vote and explicit leadership transfer ensure reliable handover even if all servers cannot vote. Leaders also step down if they lose contact with a majority of voters. The {ref}`dqlite-internals-raft-roles` control whether a server participates in quorum and elections.
+
+For more information on the Canonical Dqlite Raft implementation, see [`dqlite/src/raft.h`](https://github.com/canonical/dqlite/blob/main/src/raft.h) and [Dqlite replication](https://canonical.com/dqlite/docs/explanation/replication).
+
+(dqlite-internals-raft-roles)=
+### Dqlite raft roles
+
+1. `RAFT_VOTER`: Replicates the log and participates in quorum/elections.
+1. `RAFT_STANDBY`: Replicates the log but does not participate in quorum/elections.
+1. `RAFT_SPARE`: Does not replicate the log and does not participate in quorum/elections.
+
+### LXD cluster roles
+
+1. `database`: Assigned to cluster members with the `RAFT_VOTER` role.
+1. `database-standby`: Assigned to cluster members with the `RAFT_STANDBY` role.
+1. `database-leader`: Assigned to the current Raft leader.
--- a/doc/.custom_wordlist.txt
+++ b/doc/.custom_wordlist.txt
@@ -305,6 +305,7 @@ Solaris
 SPAs
 SPL
 SquashFS
+SQLite
 SSDs
 SSL
 SSO
--- a/doc/database.md
+++ b/doc/database.md
@@ -19,6 +19,8 @@ Dqlite provides replication, fault-tolerance, and automatic failover without the
 
 When using LXD as a single machine and not as a cluster, the Dqlite database effectively behaves like a regular SQLite database.
 
+For more information, see {ref}`dqlite-internals`.
+
 (database-location)=
 ## File location
 
--- a/doc/internals.md
+++ b/doc/internals.md
@@ -10,6 +10,7 @@
 :diataxis:User namespace setup </userns-idmap>
 :diataxis:OVN implementation </reference/ovn-internals>
 :diataxis:VM live migration implementation </reference/vm_live_migration_internals>
+:diataxis:Dqlite </reference/dqlite-internals>
 ```
 
 ## Related topics


doc/reference/dqlite-internals.md
```diff
@@ -0,0 +1,30 @@
...
+---
+
+(dqlite-internals)=
+# Dqlite internals
+
+Dqlite (distributed SQLite) implements a replicated SQLite database by combining the SQLite engine with a Raft-based consensus layer. Each LXD daemon (cluster member) runs a Dqlite node which exposes a SQLite-like API backed by a Raft replicated state machine. A single leader handles writes; followers apply replicated log entries and serve reads depending on configuration.
+
+## Raft
+
+Raft is a consensus algorithm that ensures a cluster of nodes can agree on a sequence of state machine commands even in the presence of failures. Raft handles leader election, log replication, and safety guarantees. For more information on the Raft algorithm, see the [original Raft paper](https://raft.github.io/raft.pdf).
```
minaelee:
```suggestion
Raft is a consensus algorithm that ensures a cluster of nodes can agree on a sequence of state machine commands even in the presence of failures. Raft handles leader election, log replication, and safety guarantees. For more information on the Raft algorithm, see the [original Raft paper](https://raft.github.io/raft.pdf).
```

Why not link to https://raft.github.io/ instead of the PDF? Since it includes the paper and other useful bits, like a visualization.

doc/reference/dqlite-internals.md
```diff
@@ -0,0 +1,30 @@
...
+
+Raft nodes in Dqlite move between four runtime states: `RAFT_UNAVAILABLE`, `RAFT_FOLLOWER`, `RAFT_CANDIDATE` and `RAFT_LEADER`. Followers are passive: they accept `AppendEntries` RPCs (remote procedure calls) from an active leader and reset an election timer; when a follower's randomized election timeout elapses without leader contact it becomes a candidate, increments its term and sends `RequestVote` RPCs to gather votes. A candidate becomes leader after receiving votes from a majority of voting servers and then starts replicating log entries to followers using `AppendEntries` (heartbeats are empty `AppendEntries` used to maintain authority).
```
minaelee, line 16: Since the paragraphs on line 14 and 16 do not seem to be specific to LXD, is this level of detail about DQlite/Raft necessary in the LXD documentation? We want to avoid documenting upstream products for them in detail, and prefer to summarize / link to their documentation/resources instead as much as possible. How much can we pare this down to what's important for LXD users?

minaelee (changes requested):

doc/reference/dqlite-internals.md
kadinsayani: This is meant to serve as developer documentation, similar to the [OVN implementation](https://documentation.ubuntu.com/lxd/latest/reference/ovn-internals/) and [VM live migration implementation](https://documentation.ubuntu.com/lxd/latest/reference/vm_live_migration_internals/) pages.

kadinsayani: These paragraphs summarize information gathered from the Dqlite docs, source code, and the Raft white paper, in a _concise_ format that isn't directly available upstream, while also providing a mapping of Raft roles to LXD cluster roles. IMHO, this information is valuable to include. 

I'd also like to expand the practice of adding internals/developer documentation to reduce cognitive load when reasoning about different concepts in LXD. Perhaps there is a better location for this type of documentation?

kadinsayani: I'd be fine with including this information as code comments instead of a dedicated doc page. Open to suggestions :)

doc/reference/dqlite-internals.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Raft
+
+[Raft](https://raft.github.io/) is a consensus algorithm that ensures a cluster of nodes can agree on a sequence of state machine commands even in the presence of failures. Raft handles leader election, log replication, safety, and membership changes.
+
+## Dqlite Raft implementation
+
+Raft nodes in Dqlite move between four runtime states: `RAFT_UNAVAILABLE`, `RAFT_FOLLOWER`, `RAFT_CANDIDATE` and `RAFT_LEADER`. Followers are passive: they accept `AppendEntries` RPCs (remote procedure calls) from an active leader and reset an election timer; when a follower's randomized election timeout elapses without leader contact it becomes a candidate, increments its term and sends `RequestVote` RPCs to gather votes. A candidate becomes leader after receiving votes from a majority of voting servers and then starts replicating log entries to followers using `AppendEntries` (heartbeats are empty `AppendEntries` used to maintain authority).
```
minaelee, line 16:
```suggestion
Raft nodes in Dqlite move between four runtime states: `RAFT_UNAVAILABLE`, `RAFT_FOLLOWER`, `RAFT_CANDIDATE` and `RAFT_LEADER`. Leaders coordinate log replication among followers and send out regular heartbeats to maintain authority. Followers listen for heartbeats and participate in electing a new leader when an election timeout elapses before a heartbeat is received. 
```

Do you think this suggested revision provides enough information for an LXD developer, or is the detailed information in the original necessary? 

If you think it is necessary and there is no existing resource that can be linked, then I withdraw my suggestion.

doc/reference/dqlite-internals.md
```diff
@@ -0,0 +1,32 @@
...
+
+Dqlite randomizes the election timeout (default 1000ms, randomized between 1x–2x) and uses a shorter heartbeat timeout (default 100ms) so that leaders periodically prevent elections. The implementation supports pre-vote to reduce unnecessary disruptions caused by isolated or slow nodes, and provides explicit leadership transfer primitives (`TimeoutNow` / `raft_transfer`) to hand over leadership without a full election; leaders also step down if they lose contact with a majority of voters. Roles such as `RAFT_VOTER`, `RAFT_STANDBY` and `RAFT_SPARE` control whether a server participates in quorum and elections, which is used to manage membership changes and promotions.
+
+For more information on the Canonical Dqlite Raft implementation, see [`dqlite/src/raft.h`](https://github.com/canonical/dqlite/blob/main/src/raft.h) and [Dqlite replication](https://canonical.com/dqlite/docs/explanation/replication).
+
+### Dqlite raft roles
```
minaelee, line 23:
```suggestion
(dqlite-internals-raft-roles)=
### Dqlite raft roles
```

doc/reference/dqlite-internals.md
```diff
@@ -0,0 +1,32 @@
...
+
+Dqlite randomizes the election timeout (default 1000ms, randomized between 1x–2x) and uses a shorter heartbeat timeout (default 100ms) so that leaders periodically prevent elections. The implementation supports pre-vote to reduce unnecessary disruptions caused by isolated or slow nodes, and provides explicit leadership transfer primitives (`TimeoutNow` / `raft_transfer`) to hand over leadership without a full election; leaders also step down if they lose contact with a majority of voters. Roles such as `RAFT_VOTER`, `RAFT_STANDBY` and `RAFT_SPARE` control whether a server participates in quorum and elections, which is used to manage membership changes and promotions.
```
minaelee:
```suggestion
The election timeout is randomized, but always shorter than the heartbeat interval. Features such as pre-vote and explicit leadership transfer ensure reliable handover even if all servers cannot vote. Leaders also step down if they lose contact with a majority of voters. The {ref}`dqlite-internals-raft-roles` control whether a server participates in quorum and elections.
```

Same as in the above comment -- please consider if the shorter suggestion is enough, or if the detailed content in the original is necessary for LXD development.

doc/reference/dqlite-internals.md
minaelee: See suggestions below!

doc/reference/dqlite-internals.md
kadinsayani: I prefer the more detailed version. Since this is developer documentation, we don't need to obfuscate details. There is external documentation we can link to, but I want to save other developers (and my future self) from needing to parse information spread across many sources.

doc/reference/dqlite-internals.md
kadinsayani: In this case, I prefer your shorter version. It reads better while still getting my intended points across. Thanks!

kadinsayani: The addition of the `ref` is a great suggestion too :)

kadinsayani:
> Please consider if the suggested condensed changes would be sufficient, or if the detailed original version is preferred. I'll defer if the latter.

Thanks for your review! I've addressed all comments.

PR #17123, opened 2025-12-01
lxd-agent: Make event processing synchronous

MusicDin:
Make LXD agent event processing synchronous. This ensures when the device is hotplugged or removed, it does not respond until the event is processed by the lxd-agent, meaning the guest side mount has been applied/unapplied. 

Fixes the DevLXD volume snapshot issue where the snapshot of the volume was empty. 
The root cause was that when the instance update was finished, the LXD agent may not processed the device attachment event yet, causing the race condition where the files could be written on root volume instead of the attached volume (if they were written immediately after the device was attached).

### DIFF
--- a/lxd-agent/events.go
+++ b/lxd-agent/events.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


lxd-agent/events.go
```diff
@@ -112,15 +113,18 @@ func eventsPost(d *Daemon, r *http.Request) response.Response {
 	}
 
 	// Handle device related actions locally.
-	go eventsProcess(event)
+	err = eventsProcess(event)
+	if err != nil {
+		logger.Error("Failed to process event", logger.Ctx{"err": err, "event_type": event.Type, "project": event.Project})
```
tomponline: I dont see "event_type" used elsewhere in the code base, so can you double check if we've logged event type using a different label previously.

MusicDin: We used just `type` in one place. Will fix

PR #17124, opened 2025-12-01
doc: Add that QEMU is required for VMs

markylaing:
Add `qemu-system` to list of build deps if VM support is needed

## Checklist

### DIFF
--- a/doc/installing.md
+++ b/doc/installing.md
@@ -197,7 +197,9 @@ These builds are for the [`lxc`](lxc.md) client only, not the LXD daemon. For an
 (installing_from_source)=
 ## Install LXD from source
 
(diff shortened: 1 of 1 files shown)


markylaing:
> Can we use `--no-install-recommends` with a specific list of required packages rather than pulling in unneeded packages?

My personal preference for this guide would be to make it as simple as possible and also add a caveat that the guide was written for Ubuntu Noble only.

tomponline:
> > Can we use `--no-install-recommends` with a specific list of required packages rather than pulling in unneeded packages?
> 
> My personal preference for this guide would be to make it as simple as possible and also add a caveat that the guide was written for Ubuntu Noble only.

But that command pulls in a lot of stuff that is not needed.

Suggest these:

- qemu-system-x86
- qemu-block-extra
- qemu-utils
- qemu-system-modules-spice
- virtiofsd
- ovmf
- swtpm

Fine with it only being targeted for Noble though.

Issue #17067, opened 2025-11-25
Optimize tracking attached volume snapshots in volatile config


kadinsayani:
https://github.com/canonical/lxd/blob/91c2ad2e3618effe444b690480e7bb749146e326/lxc/info.go#L646-L688

Looking at this relatively expensive process to figure out the disk device associated to a snapshot volume, I wonder if we should not rather store this in the instance snapshot's volatile config in a different way.

Similar to the approach I went with for persistent PCIe bus allocation, by using `volatile.<device>.volume_snapshot` - indicating that this device has an associated volume snapshot, and which volume snapshot uuid it is.

And phase out the "volatile.attached_volumes" concept.

_Originally posted by @tomponline in https://github.com/canonical/lxd/pull/17008#discussion_r2560414240_

This change would also require modifying multi-volume snapshot/restore associated helper functions, adding a patch to convert any existing use of `volatile.attached_volumes`, a new API extension, and a documentation update.

tomponline:
Another potential option is to change the format of the `volatile.attached_volumes` key to be in the form of `<deviceName>:<volumeUUID>`

Issue #17125, opened 2025-12-01
Volume snapshots can be attached using `source=<vol>/<snap>` rather than requiring use of `source.snapshot` key


tomponline:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

N/A

### Distribution version

N/A

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
N/A
```

### Output of "lxc info" or system info if it fails

```shell
LXD 6.6
```

### Issue description

We should not have 2 ways to attach a volume snapshot as a `disk` device.

There is the documented `source.snapshot` config key, but it also works using `source=<vol>/<snapshot>`.

Only the former should work.

https://documentation.ubuntu.com/lxd/latest/reference/devices_disk/#device-disk-device-conf:source.snapshot

### Steps to reproduce

```
lxc storage volume create default foo
lxc storage volume snapshot default foo
lxc launch ubuntu:24.04 v1 --vm
lxc config device add v1 foo disk source=foo source.snapshot=foo pool=default path=/mnt/foo-snap0
lxc config device add v1 foo disk source=foo/snap0 pool=default path=/mnt/bar # should not be allowed, but works!
```

### Information to attach

- [ ] Any relevant kernel output (`dmesg`)
- [ ] Instance log (`lxc info NAME --show-log`)
- [ ] Instance configuration (`lxc config show NAME --expanded`)
- [ ] Main daemon log (at `/var/log/lxd/lxd.log` or `/var/snap/lxd/common/lxd/logs/lxd.log`)
- [ ] Output of the client with `--debug`
- [ ] Output of the daemon with `--debug` (or use `lxc monitor` while reproducing the issue)

MggMuggins:
@tomponline This is also using the `dir` storage driver as in #17126?

The `source: <vol>/<snap>` syntax was never implemented (or at least, I'm sure I didn't do it intentionally now that I've looked over my PRs in this area). I introduced similar syntax in #14491 but that should have been completely reverted by #14886.

This feels like missing validation in the disk device driver, but it's been a long time since I touched that code.

See also:
- VM root volumes as disk devices (#14532)
- Revert the rest of #14491 (#14921)
- VM snapshots as disk devices (#14930)
- `source-type` -> `source.type` (#14982)

PR #17127, opened 2025-12-01
Optimize volatile.attached_volumes storage and lookups for multi-volume snapshot tracking

kadinsayani:
Closes https://github.com/canonical/lxd/issues/17005 and https://github.com/canonical/lxd/issues/17067.

## New `volatile.attached_volumes` format

Optimizes the `volatile.attached_volumes` format for multi-volume snapshot tracking by changing from volume UUID keys to device name keys. This improves lookup performance and simplifies code for displaying and restoring multi-volume snapshots.

- **Before:** `{"<volume-uuid>": "<snapshot-uuid>", ...}`
- **After:** `{"<device-name>": "<snapshot-uuid>", ...}`

### Motivation

The original volume UUID format required expensive lookups:
- `lxc info`: For each device, we would have to fetch volume info to match UUID, then lookup snapshot UUID
- In snapshot restore: Iterate all volumes to build UUID map, then correlate with devices

The new device name-based format:
- Eliminates volume UUID lookups (device names are already available)
- Enables direct iteration over the map for restore operations
- Simplifies the code and improves readability

## `lxc info`

Adds attached volume snapshots to `lxc info` output:

```
...
Snapshots:
+----------+----------------------+----------------------+----------+---------------------------+
|   NAME   |       TAKEN AT       |      EXPIRES AT      | STATEFUL | ATTACHED VOLUME SNAPSHOTS |
+----------+----------------------+----------------------+----------+---------------------------+
| c1-snap0 | 2025/11/25 15:55 UTC |                      | NO       |
+----------+----------------------+----------------------+----------+---------------------------+
| c1-snap1 | 2025/11/25 15:55 UTC | 2025/11/25 17:55 UTC | NO       | non-shared/snap0          |
|          |                      |                      |          | shared/snap0              |
+----------+----------------------+----------------------+----------+---------------------------+
| c1-snap2 | 2025/11/25 15:55 UTC | 2025/11/25 17:55 UTC | NO       | shared/snap1              |
|          |                      |                      |          | non-shared/snap1          |
+----------+----------------------+----------------------+----------+---------------------------+
```

### DIFF
--- a/po/ar.po
+++ b/po/ar.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: lxd\n"
 "Report-Msgid-Bugs-To: lxd@lists.canonical.com\n"
-"POT-Creation-Date: 2025-11-17 14:48+0000\n"
+"POT-Creation-Date: 2025-12-08 09:07-0800\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Automatically generated\n"
 "Language-Team: none\n"
@@ -432,7 +432,7 @@ msgid ""
 "### Any line starting with a '# will be ignored."
 msgstr ""
 
-#: lxc/info.go:329
+#: lxc/info.go:330
 #, c-format
 msgid "%d (id: %d, online: %v, NUMA node: %v)"
 msgstr ""
@@ -442,7 +442,7 @@ msgstr ""
 msgid "%s (%d more)"
 msgstr ""
 
-#: lxc/info.go:169
+#: lxc/info.go:170
 #, c-format
 msgid "%s (%s) (%d available)"
 msgstr ""
@@ -461,17 +461,17 @@ msgstr ""
 msgid "(none)"
 msgstr ""
 
-#: lxc/info.go:319
+#: lxc/info.go:320
 #, c-format
 msgid "- Level %d (type: %s): %s"
 msgstr ""
 
-#: lxc/info.go:297
+#: lxc/info.go:298
 #, c-format
 msgid "- Partition %d"
 msgstr ""
 
-#: lxc/info.go:205
+#: lxc/info.go:206
 #, c-format
 msgid "- Port %d (%s)"
 msgstr ""
@@ -513,7 +513,7 @@ msgid "--refresh can only be used with instances"
 msgstr ""
 
 #: lxc/config.go:211 lxc/config.go:482 lxc/config.go:657 lxc/config.go:864
-#: lxc/info.go:461
+#: lxc/info.go:462
 msgid "--target cannot be used with instances"
 msgstr ""
 
@@ -718,7 +718,7 @@ msgstr ""
 msgid "Add rules to an ACL"
 msgstr ""
 
-#: lxc/info.go:209
+#: lxc/info.go:210
 #, c-format
 msgid "Address: %s"
 msgstr ""
@@ -773,12 +773,12 @@ msgstr ""
 msgid "Alternative certificate name"
 msgstr ""
 
-#: lxc/image.go:1017 lxc/info.go:486
+#: lxc/image.go:1017 lxc/info.go:487
 #, c-format
 msgid "Architecture: %s"
 msgstr ""
 
-#: lxc/info.go:135
+#: lxc/info.go:136
 #, c-format
 msgid "Architecture: %v"
 msgstr ""
@@ -855,7 +855,7 @@ msgstr ""
 msgid "Authentication type '%s' not supported by server"
 msgstr ""
 
-#: lxc/info.go:228
+#: lxc/info.go:229
 #, c-format
 msgid "Auto negotiation: %v"
 msgstr ""
@@ -895,7 +895,7 @@ msgstr ""
 msgid "Backup exported successfully!"
 msgstr ""
 
-#: lxc/info.go:666 lxc/storage_volume.go:1577
+#: lxc/info.go:722 lxc/storage_volume.go:1577
 msgid "Backups:"
 msgstr ""
 
@@ -935,7 +935,7 @@ msgstr ""
 msgid "Both --all and instance name given"
 msgstr ""
 
-#: lxc/info.go:136
+#: lxc/info.go:137
 #, c-format
 msgid "Brand: %v"
 msgstr ""
@@ -944,11 +944,11 @@ msgstr ""
 msgid "Bridge:"
 msgstr ""
 
-#: lxc/info.go:589 lxc/network.go:957
+#: lxc/info.go:590 lxc/network.go:957
 msgid "Bytes received"
 msgstr ""
 
-#: lxc/info.go:590 lxc/network.go:958
+#: lxc/info.go:591 lxc/network.go:958
 msgid "Bytes sent"
 msgstr ""
 
@@ -968,7 +968,7 @@ msgstr ""
 msgid "COUNT"
 msgstr ""
 
-#: lxc/info.go:364
+#: lxc/info.go:365
 #, c-format
 msgid "CPU (%s):"
 msgstr ""
@@ -977,15 +977,15 @@ msgstr ""
 msgid "CPU USAGE"
 msgstr ""
 
-#: lxc/info.go:539
+#: lxc/info.go:540
 msgid "CPU usage (in seconds)"
 msgstr ""
 
-#: lxc/info.go:543
+#: lxc/info.go:544
 msgid "CPU usage:"
 msgstr ""
 
-#: lxc/info.go:367
+#: lxc/info.go:368
 #, c-format
 msgid "CPUs (%s):"
 msgstr ""
@@ -1002,7 +1002,7 @@ msgstr ""
 msgid "CREATION DATE"
 msgstr ""
 
-#: lxc/info.go:138
+#: lxc/info.go:139
 #, c-format
 msgid "CUDA Version: %v"
 msgstr ""
@@ -1012,7 +1012,7 @@ msgstr ""
 msgid "Cached: %s"
 msgstr ""
 
-#: lxc/info.go:317
+#: lxc/info.go:318
 msgid "Caches:"
 msgstr ""
 
@@ -1084,12 +1084,12 @@ msgstr ""
 msgid "Cannot use metrics type certificate when using a token"
 msgstr ""
 
-#: lxc/info.go:411 lxc/info.go:423
+#: lxc/info.go:412 lxc/info.go:424
 #, c-format
 msgid "Card %d:"
 msgstr ""
 
-#: lxc/info.go:121
+#: lxc/info.go:122
 #, c-format
 msgid "Card: %s (%s)"
 msgstr ""
@@ -1173,7 +1173,7 @@ msgid "Cluster member %s removed from group %s"
 msgstr ""
 
 #: lxc/config.go:147 lxc/config.go:439 lxc/config.go:590 lxc/config.go:797
-#: lxc/config.go:928 lxc/copy.go:63 lxc/info.go:45 lxc/init.go:65
+#: lxc/config.go:928 lxc/copy.go:63 lxc/info.go:46 lxc/init.go:65
 #: lxc/move.go:68 lxc/network.go:338 lxc/network.go:809 lxc/network.go:890
 #: lxc/network.go:1025 lxc/network.go:1291 lxc/network.go:1382
 #: lxc/network.go:1454 lxc/network_forward.go:185 lxc/network_forward.go:267
@@ -1277,7 +1277,7 @@ msgstr ""
 msgid "Content type: %s"
 msgstr ""
 
-#: lxc/info.go:125
+#: lxc/info.go:126
 #, c-format
 msgid "Control: %s (%s)"
 msgstr ""
@@ -1365,12 +1365,12 @@ msgstr ""
 msgid "Copying the storage volume: %s"
 msgstr ""
 
-#: lxc/info.go:325
+#: lxc/info.go:326
 #, c-format
 msgid "Core %d"
 msgstr ""
 
-#: lxc/info.go:323
+#: lxc/info.go:324
 msgid "Cores:"
 msgstr ""
 
@@ -1551,7 +1551,7 @@ msgstr ""
 msgid "Create the instance with no profiles applied"
 msgstr ""
 
-#: lxc/image.go:1024 lxc/info.go:497 lxc/storage_volume.go:1532
+#: lxc/image.go:1024 lxc/info.go:498 lxc/storage_volume.go:1532
 #, c-format
 msgid "Created: %s"
 msgstr ""
@@ -1570,7 +1570,7 @@ msgstr ""
 msgid "Creating the instance"
 msgstr ""
 
-#: lxc/info.go:145 lxc/info.go:254
+#: lxc/info.go:146 lxc/info.go:255
 #, c-format
 msgid "Current number of VFs: %d"
 msgstr ""
@@ -1598,7 +1598,7 @@ msgstr ""
 msgid "DRIVER"
 msgstr ""
 
-#: lxc/info.go:117
+#: lxc/info.go:118
 msgid "DRM:"
 msgstr ""
 
@@ -1769,7 +1769,7 @@ msgstr ""
 #: lxc/image.go:1092 lxc/image.go:1452 lxc/image.go:1547 lxc/image.go:1613
 #: lxc/image.go:1677 lxc/image.go:1740 lxc/image_alias.go:24
 #: lxc/image_alias.go:60 lxc/image_alias.go:124 lxc/image_alias.go:177
-#: lxc/image_alias.go:288 lxc/import.go:29 lxc/info.go:33 lxc/init.go:44
+#: lxc/image_alias.go:288 lxc/import.go:29 lxc/info.go:34 lxc/init.go:44
 #: lxc/launch.go:24 lxc/list.go:50 lxc/main.go:86 lxc/manpage.go:22
 #: lxc/monitor.go:34 lxc/move.go:39 lxc/network.go:36 lxc/network.go:139
 #: lxc/network.go:241 lxc/network.go:331 lxc/network.go:418 lxc/network.go:476
@@ -1910,7 +1910,7 @@ msgstr ""
 msgid "Device from profile(s) cannot be retrieved for individual instance"
 msgstr ""
 
-#: lxc/info.go:274 lxc/info.go:299
+#: lxc/info.go:275 lxc/info.go:300
 #, c-format
 msgid "Device: %s"
 msgstr ""
@@ -1939,16 +1939,16 @@ msgstr ""
 msgid "Disable stdin (reads from /dev/null)"
 msgstr ""
 
-#: lxc/info.go:435
+#: lxc/info.go:436
 #, c-format
 msgid "Disk %d:"
 msgstr ""
 
-#: lxc/info.go:533
+#: lxc/info.go:534
 msgid "Disk total:"
 msgstr ""
 
-#: lxc/info.go:529
+#: lxc/info.go:530
 msgid "Disk usage:"
 msgstr ""
 
@@ -1977,11 +1977,11 @@ msgid ""
 "volumes (non-shared)."
 msgstr ""
 
-#: lxc/info.go:430
+#: lxc/info.go:431
 msgid "Disk:"
 msgstr ""
 
-#: lxc/info.go:433
+#: lxc/info.go:434
 msgid "Disks:"
 msgstr ""
 
@@ -2029,7 +2029,7 @@ msgstr ""
 msgid "Down delay"
 msgstr ""
 
-#: lxc/info.go:113 lxc/info.go:199
+#: lxc/info.go:114 lxc/info.go:200
 #, c-format
 msgid "Driver: %v (%v)"
 msgstr ""
@@ -2300,7 +2300,7 @@ msgid ""
 "AND stdout are terminals (stderr is ignored)."
 msgstr ""
 
-#: lxc/info.go:652 lxc/info.go:703 lxc/storage_volume.go:1564
+#: lxc/info.go:707 lxc/info.go:759 lxc/storage_volume.go:1564
 #: lxc/storage_volume.go:1614
 msgid "Expires at"
 msgstr ""
@@ -2641,17 +2641,17 @@ msgstr ""
 msgid "Found alias %q references an argument outside the given number"
 msgstr ""
 
-#: lxc/info.go:378 lxc/info.go:389 lxc/info.go:394 lxc/info.go:400
+#: lxc/info.go:379 lxc/info.go:390 lxc/info.go:395 lxc/info.go:401
 #, c-format
 msgid "Free: %v"
 msgstr ""
 
-#: lxc/info.go:326 lxc/info.go:337
+#: lxc/info.go:327 lxc/info.go:338
 #, c-format
 msgid "Frequency: %vMhz"
 msgstr ""
 
-#: lxc/info.go:335
+#: lxc/info.go:336
 #, c-format
 msgid "Frequency: %vMhz (min: %vMhz, max: %vMhz)"
 msgstr ""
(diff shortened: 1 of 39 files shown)


kadinsayani:
@tomponline something about supporting two formats seems off. Perhaps we can add a patch to check for the presence of `volatile.attached_volumes` in the old format (`<volatile.uuid>:<snapshot UUID>`) and convert it to the new format (`<device name>:<snapshot UUID>`)? What do you think?

tomponline:
> Perhaps we can add a patch to check for the presence of `volatile.attached_volumes` in the old format (`<volatile.uuid>:<snapshot UUID>`) and convert it to the new format (`<device name>:<snapshot UUID>`)? What do you think?

sounds good

kadinsayani:
The failing system tests for `cluster, btrfs` look unrelated to my changes. However, it's an interesting failure, likely the result of a race condition. Have we seen this one before?

simondeziel:
> The failing system tests for `cluster, btrfs` look unrelated to my changes. However, it's an interesting failure, likely the result of a race condition. Have we seen this one before?

The `sql: transaction has already been committed or rolled back` bit is not uncommon and seems to be more frequent under load/contention.

As for the `clustering_move` test itself, it is one of the "less reliable" ones in the `cluster` group IIRC.

tomponline:
> > The failing system tests for `cluster, btrfs` look unrelated to my changes. However, it's an interesting failure, likely the result of a race condition. Have we seen this one before?
> 
> The `sql: transaction has already been committed or rolled back` bit is not uncommon and seems to be more frequent under load/contention.
> 
> As for the `clustering_move` test itself, it is one of the "less reliable" ones in the `cluster` group IIRC.

https://chat.canonical.com/canonical/pl/nztxeqbgzpf5mmmawx1j8i6dmc

lxd/patches.go
```diff
@@ -1845,4 +1848,164 @@ func patchDefaultStoragePermissions(_ string, d *Daemon) error {
 	return nil
 }
 
+// patchUpdateVolatileAttachedVolumesFormat updates "volatile.attached_volumes" from old format (volume UUID -> snapshot UUID) to new format (device name -> snapshot UUID).
+func patchUpdateVolatileAttachedVolumesFormat(_ string, d *Daemon) error {
```
tomponline, line 1866: Has any consideration been given to what happens when (not if) this patch is run concurrently on multiple cluster members at once during a cluster upgrade? It looks like not, and this will likely be a problem.

Also, there seems to be an unnecessary amount of transactions.
Transactions in dqlite are relatively expensive, especially write ones, so we should make judicious use of them.

It seems to me that this patch can be simplified:

Inside a single transaction with the following steps:

1. Select snapshots with "volatile.attached_volumes" set (you could also filter by volume type).
2. Try unmarshalling that into the `[]string` to see if its in the old format, if that fails then skip it as it must be in the new format (or corrupt and theres nothing we can do about that here).
3. For volumes in the old format, store the unmarshalled slice along with instance ID and snapshot ID into a modified map of `snapshotEntry` keyed on instance ID.
4. Then iterate the update list on the instance ID, meaning we only need to load the instance devices once, resolve the volume uuid, marshal the updated volume json and execute an update query.
5. Commit the transaction.

kadinsayani: Thanks, great suggestions. I'll implement your suggested optimized approach, but I think we may be on a different page for the new format. I was going with a map of device name to snapshot volatile.uuid for the new format (as opposed to the old map of attached volume volatile.uuid to snapshot volatile.uuid). Do you think a string slice containing strings of the format "device name:snapshot volatile.uuid" would be better?

kadinsayani: A `[]string` would make it easier to detect the new format.

tomponline: I'm happy with your proposal for the new format, but my suggestion for `[]string` was for decoding the old format into, if it fails, then its already been converted to the new format, and if it succeeds then you know its in the old format and you have then already unmarshalled it ready for conversion to new format.

kadinsayani: I'm not following your suggestion to unmarshal volatile.attached_volumes into a `[]string` since the old format is `map[string]string`. I've reduced the patch to a single transaction, optimized where possible, and added a check using `selectedPatchClusterMember()` to avoid concurrent updates.

tomponline: What was the original purpose of having vol UUID -> volume snapshot UUID map instead of now device name -> volume snapshot UUID?

lxd/patches.go
```diff
@@ -1845,4 +1848,157 @@ func patchDefaultStoragePermissions(_ string, d *Daemon) error {
...
+		return err
+	}
+
+	if !isSelectedMember {
+		return nil
+	}
+
+	err = s.DB.Cluster.Transaction(s.ShutdownCtx, func(ctx context.Context, tx *db.ClusterTx) error {
+		// Query to get all instance snapshots with "volatile.attached_volumes".
+		rows, err := tx.Tx().QueryContext(ctx, `
```
tomponline: Please use `lxd/db/query/objects.Scan()`

lxd/instance/instancetype/instance.go
```diff
@@ -445,12 +445,12 @@ var InstanceConfigKeysAny = map[string]func(value string) error{
...
 
 	// lxdmeta:generate(entities=instance; group=volatile; key=volatile.attached_volumes)
-	// JSON-serialized map of attached volume UUIDs to the UUIDs of their corresponding
+	// JSON-serialized map of attached volume device names to the UUIDs of their corresponding
 	// snapshots, created as part of a multi-volume snapshot.
 	//
 	// ---
 	//   type: string
-	//   shortdesc: JSON-serialized map of attached volume UUIDs to the UUIDs of their corresponding snapshots.
+	//   shortdesc: JSON-serialized map of attached volume device names to the UUIDs of their corresponding snapshots.
```
tomponline, line 453: isnt this the other way round? UUID -> device names?

lxd/instance/drivers/driver_common.go
```diff
@@ -921,14 +922,15 @@ func (d *common) getAttachedVolumes(inst instance.Instance) (volumes []*db.Stora
 			}
 
 			volumes = append(volumes, vol)
+			deviceNames[vol.Config["volatile.uuid"]] = name
```
tomponline: there seems to be some confusion in the comments in earlier commit about whether this map is a map of volume snapshot uuid -> device name or device name -> volume snapshot uuid, here it is the former, but in the earlier comment it was the latter.

tomponline:
@kadinsayani ive stopped reviewing this for now as im finding it pretty hard going due to lack of clarity around the change. 

Please can you update the PR description with an expanded explanation of the change, showing the old format and the new format examples, along with explanation as to why the volume UUID mapping is no longer required, and only device name is.

lxd/instance/drivers/driver_common.go
kadinsayani: I can see where the confusion is stemming from. This map is returned from `getAttachedVolumes()`. We use this map to lookup the device name by the volume's volatile.uuid in `snapshotCommon()`. Once we have the device name, we create the entry for `volatile.attached_volumes` (device name -> snapshot volatile.uuid).

lxd/instance/instancetype/instance.go
kadinsayani: No, it should be device name -> UUID (snapshot volatile.uuid).

lxd/instance/drivers/driver_common.go
kadinsayani: This could be simplified by just returning a map of device name to volume from `getAttachedVolumes()`.

kadinsayani:
> @kadinsayani ive stopped reviewing this for now as im finding it pretty hard going due to lack of clarity around the change.
> 
> Please can you update the PR description with an expanded explanation of the change, showing the old format and the new format examples, along with explanation as to why the volume UUID mapping is no longer required, and only device name is.

Sure thing, I'll also add more comments for clarity and change a few variable names to be more consistent and aligned with what they represent.

lxd/patches.go
kadinsayani: Ah yes, thanks.

lxd/patches.go
kadinsayani: It looks like the volume volatile.uuid served no purpose other than identifying the volume we want to locate the snapshot for. Using the device name as our map key is much more efficient.

kadinsayani:
@tomponline ready for another review whenever you have a chance. I've updated the PR description, added commit descriptions, optimized code and renamed vars for clarity where possible.

tomponline:
> @tomponline ready for another review whenever you have a chance. I've updated the PR description, added commit descriptions, optimized code and renamed vars for clarity where possible.

Excellent, thank you for the expanded description!

lxd/patches.go
```diff
@@ -1845,4 +1848,152 @@ func patchDefaultStoragePermissions(_ string, d *Daemon) error {
...
+			}
+
+			// Convert from old format (volume UUID -> snapshot UUID) to new format (device name -> snapshot UUID).
+			newVolatileAttachedVolumes := make(map[string]string)
+			for name, dev := range devices {
+				if !filters.IsCustomVolumeDisk(dev.Config) {
+					continue
+				}
+
+				poolID, err := tx.GetStoragePoolID(ctx, dev.Config["pool"])
```
tomponline: Can we push this down into a single query by taking the list of volume uuids keys from `attachedVolumesData` and then running a query that gets the volumes referenced by those UUIDs, along with their pool name and volume name.

Then you can convert that list to a pool name & vol name -> vol uuid map, and use that to find the device(s) that they match to.

BTW I thought of some things to check:

1. If a project doesn't have the `features.storage.volumes` enabled, do the UUIDs point to volumes in the default project, rather than the instance project?
2. Can a single volume be used by multiple instance devices (yes), so this needs to be taken into account when converting the devices (if not already).

lxc/info.go
```diff
@@ -642,6 +643,40 @@ func (c *cmdInfo) instanceInfo(d lxd.InstanceServer, name string, showLog bool)
...
+				err := json.Unmarshal([]byte(snap.Config["volatile.attached_volumes"]), &volatileAttachedVolumes)
+				if err != nil {
+					return fmt.Errorf(`Failed parsing "volatile.attached_volumes" from snapshot %q: %w`, snap.Name, err)
+				}
+
+				attachedVolumeSnapshotNames := make([]string, 0, len(volatileAttachedVolumes))
+				for deviceName, snapshotUUID := range volatileAttachedVolumes {
+					dev, ok := snap.Devices[deviceName]
+					if !ok {
+						continue
```
tomponline, line 666: this handles clients retrieving data from servers using the older format.

lxc/info.go
```diff
@@ -642,6 +643,40 @@ func (c *cmdInfo) instanceInfo(d lxd.InstanceServer, name string, showLog bool)
...
+					}
+
+					volName, _, _ := api.GetParentAndSnapshotName(dev["source"])
```
tomponline: are you expecting `vol/snap` format in `dev[source]`?

tomponline: because its a bug if so https://github.com/canonical/lxd/issues/17125

lxc/info.go
```diff
@@ -642,6 +643,40 @@ func (c *cmdInfo) instanceInfo(d lxd.InstanceServer, name string, showLog bool)
...
+					snapshots, err := d.GetStoragePoolVolumeSnapshots(dev["pool"], "custom", volName)
```
tomponline: This is an expensive call, both from a network and DB perspective, can we cache the result of this in an map outside of the `for _, snap := range inst.Snapshots {` loop so that if another snapshot also references the same poo/vol combo (likely) then we dont have to load it over and over again.

test/suites/snapshots.sh
```diff
@@ -529,6 +529,7 @@ test_snapshot_multi_volume() {
 
   echo "Check multi-volume snapshot."
   lxc snapshot c1 c1-snap2 --disk-volumes=all-exclusive
+  lxc info c1 # For debugging.
```
tomponline: left over debugging output

lxd/patches.go
kadinsayani: > Can we push this down into a single query by taking the list of volume uuids keys from `attachedVolumesData` and then running a query that gets the volumes referenced by those UUIDs, along with their pool name and volume name.
> 
> Then you can convert that list to a pool name & vol name -> vol uuid map, and use that to find the device(s) that they match to.

Yes, I think that approach will work well!

> BTW I thought of some things to check:
> 
> 1. If a project doesn't have the `features.storage.volumes` enabled, do the UUIDs point to volumes in the default project, rather than the instance project?

When `features.storage.volumes` is disabled, we don't allow multi-volume snapshots:

https://github.com/canonical/lxd/blob/6128b86863287664c9ad50e57948e78593e965a9/lxd/instance_snapshot.go#L330-L333

> 2. Can a single volume be used by multiple instance devices (yes), so this needs to be taken into account when converting the devices (if not already).

This will be handled under the new format, in the patch we will:

1. Iterate through each device by name
2. Find the volume UUID it references
3. Map that device name to the corresponding snapshot UUID
4. If multiple devices reference the same volume UUID, they will each get their own entry in the new `volatile.attached_volumes`

tomponline: Great, thanks!

lxc/info.go
kadinsayani: Ah yes, I'll add logic to make sure we handle the old format here.

test/suites/snapshots.sh
kadinsayani: This is also for code coverage, I'll update the comment.

lxc/info.go
kadinsayani: Yeah good idea!

lxc/info.go
kadinsayani: We can just use `dev[source]` and exclude that function call. I'm not sure why I included that, I think it was just out of habit.

lxc/info.go
kadinsayani: I've added caching.

lxc/info.go
kadinsayani: Fixed.

lxd/patches.go
```diff
@@ -1859,4 +1862,165 @@ func patchUnsetCephFSPristineSetting(_ string, d *Daemon) error {
...
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+
+		for _, snap := range snapshots {
+			var volatileAttachedVolumes map[string]string
+			err = json.Unmarshal([]byte(snap.volatileAttachedVolumesValue), &volatileAttachedVolumes)
+			if err != nil || len(volatileAttachedVolumes) == 0 {
```
tomponline: I suggest we lose the `|| len(volatileAttachedVolumes) == 0 ` as if its empty then there's nothing to do.

tomponline: or move it out of the error handling

lxd/patches.go
```diff
@@ -1859,4 +1862,165 @@ func patchUnsetCephFSPristineSetting(_ string, d *Daemon) error {
...
+				if err != nil {
+					logger.Warn(`Failed parsing "volatile.attached_volumes"`, logger.Ctx{"err": err})
```
tomponline: for what instance, snapshot and in what project?

lxd/patches.go
```diff
@@ -1859,4 +1862,165 @@ func patchUnsetCephFSPristineSetting(_ string, d *Daemon) error {
...
+			}
+
+			// Get instance snapshot devices.
+			devices, err := dbCluster.GetInstanceSnapshotDevices(ctx, tx.Tx(), snap.snapshotID)
+			if err != nil {
+				return fmt.Errorf("Failed getting instance snapshot devices: %w", err)
+			}
+
+			// Skip if already in new format (any key matches a device name).
+			isNewFormat := false
```
tomponline: Just an idea, but if you defined this as a function inside this patch function called `isInNewFormat(...) (bool, err)` then you could use early `return` statements and not need the `break` and `if isNewFormat` check later.

lxd/patches.go
```diff
@@ -1859,4 +1862,165 @@ func patchUnsetCephFSPristineSetting(_ string, d *Daemon) error {
...
+			// Get all custom volumes with matching UUIDs.
+			volumes, err := tx.GetStorageVolumes(ctx, true, filter)
+			if err != nil {
+				return fmt.Errorf("Failed getting storage volumes: %w", err)
+			}
+
+			// Create a map of pool/volume -> UUID for matching devices.
+			volumeByPoolAndName := make(map[string]string)
+			for _, vol := range volumes {
+				volumeByPoolAndName[vol.Pool+"/"+vol.Name] = vol.Config["volatile.uuid"]
```
tomponline: You could use a function specific type to avoid needing to encode the pool and vol name as the map key, e.g.

```go
type volKey struct {
	name string
	pool string
}

poolvols := map[volKey]string{
		{
			name: "foo",
			pool: "bar",
		}: "uuid",
}
```

lxd/patches.go
```diff
@@ -1859,4 +1862,165 @@ func patchUnsetCephFSPristineSetting(_ string, d *Daemon) error {
...
+				snapshotUUID, ok := volatileAttachedVolumes[volumeUUID]
+				if !ok {
+					continue
+				}
+
+				newVolatileAttachedVolumes[name] = snapshotUUID
+			}
+
+			// Skip if no volumes were converted.
+			if len(newVolatileAttachedVolumes) == 0 {
```
tomponline, line 2006: is this a safety net, or in what scenario could this happen?

lxd/patches.go
```diff
@@ -1859,4 +1862,165 @@ func patchUnsetCephFSPristineSetting(_ string, d *Daemon) error {
...
+				continue
+			}
+
+			marshalled, err := json.Marshal(newVolatileAttachedVolumes)
+			if err != nil {
+				logger.Warn(`Failed marshalling new "volatile.attached_volumes" format, skipping snapshot`, logger.Ctx{"err": err, "snapshotID": snap.snapshotID, "instanceID": snap.instanceID})
+				continue
```
tomponline: `continuing` here is the wrong thing to do, the patch isn't able to be applied cleanly so it should fail.

lxd/patches.go
```diff
@@ -1859,4 +1862,165 @@ func patchUnsetCephFSPristineSetting(_ string, d *Daemon) error {
...
+`, string(marshalled), snap.snapshotID)
+			if err != nil {
+				return fmt.Errorf(`Failed updating instance snapshot "volatile.attached_volumes": %w`, err)
+			}
+		}
+
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf(`Failed updating "volatile.attached_volumes" to new format: %w`, err)
```
tomponline, line 2028: This error could do with more context about which snapshot is failing.

lxd/patches.go
```diff
@@ -1859,4 +1862,165 @@ func patchUnsetCephFSPristineSetting(_ string, d *Daemon) error {
...
+				continue
+			}
+
+			_, err = tx.Tx().ExecContext(ctx, `
+UPDATE instances_snapshots_config 
+SET value = ? 
+WHERE instance_snapshot_id = ? AND key = "volatile.attached_volumes"
+`, string(marshalled), snap.snapshotID)
+			if err != nil {
+				return fmt.Errorf(`Failed updating instance snapshot "volatile.attached_volumes": %w`, err)
```
tomponline: this error wrapping could do with being shorter as it'll be wrapped with the outer error which is very similar.

lxd/instance/drivers/driver_common.go
```diff
@@ -956,7 +956,7 @@ func (d *common) snapshotCommon(inst instance.Instance, name string, expiry *tim
 	}
 
 	// Load attached volumes for multi-volume snapshot (if requested).
-	var attachedVolumes []*db.StorageVolume
+	var attachedVolumes map[string]*db.StorageVolume
```
tomponline: not a blocker, but do we need to have a map of db.StorageVolume pointers, or can we store the literal value in the map rather than the pointer (as maps are passed as pointers anyway).

lxc/info.go
```diff
@@ -642,6 +644,77 @@ func (c *cmdInfo) instanceInfo(d lxd.InstanceServer, name string, showLog bool)
...
+				err := json.Unmarshal([]byte(snap.Config["volatile.attached_volumes"]), &volatileAttachedVolumes)
+				if err != nil {
+					return fmt.Errorf(`Failed parsing "volatile.attached_volumes" from snapshot %q: %w`, snap.Name, err)
+				}
+
+				attachedVolumeSnapshotNames := make([]string, 0, len(volatileAttachedVolumes))
+				for deviceName, snapshotUUID := range volatileAttachedVolumes {
+					dev, ok := snap.Devices[deviceName]
+					if !ok {
+						// Handle old "volatile.attached_volumes" format (map of volume UUID -> snapshot UUID).
```
tomponline: As this old format was only in 1 feature release, as you had previously gracefully handled the scenario without displaying the attached volumes I felt that was fine, whereas now you've adding the more expensive code we aimed to avoid adding in the first place.

I suggest we leave it as it was.

lxc/info.go
```diff
@@ -642,6 +644,77 @@ func (c *cmdInfo) instanceInfo(d lxd.InstanceServer, name string, showLog bool)
...
+									break
+								}
+							}
+						}
+					}
+
+					// Handle new "volatile.attached_volumes" format (map of device name -> snapshot UUID).
+
+					// Get storage volume snapshots, using cache if possible.
+					storageVolumeSnapshots, ok := storageVolumeSnapshotsCache[dev["pool"]+"/"+dev["source"]]
```
tomponline: just a nit, but I tend to use `foo, found := ...` format as I find it clearer than `ok` (what is "ok" in this context, sometimes it might be "ok" for an element to not be found)?

lxc/info.go
```diff
@@ -642,6 +644,77 @@ func (c *cmdInfo) instanceInfo(d lxd.InstanceServer, name string, showLog bool)
...
+					if !ok {
+						// Storage volume snapshots not in cache yet, fetch them.
+						storageVolumeSnapshots, err = d.GetStoragePoolVolumeSnapshots(dev["pool"], "custom", dev["source"])
+						if err != nil {
+							return err
+						}
+
+						// Cache the storage volume snapshots.
+						storageVolumeSnapshotsCache[dev["pool"]+"/"+dev["source"]] = storageVolumeSnapshots
```
tomponline: could use a local volKey struct type as map key type here too.

lxc/info.go
```diff
@@ -651,6 +724,7 @@ func (c *cmdInfo) instanceInfo(d lxd.InstanceServer, name string, showLog bool)
 			i18n.G("Taken at"),
 			i18n.G("Expires at"),
 			i18n.G("Stateful"),
+			i18n.G("Attached volume snapshots"),
```
tomponline: Shall we shorten this to just "Volume snapshots", their attachment is implicit given they are associated to an instance snapshot.

lxd/patches.go
kadinsayani: This is a safety net and optimization to prevent an unnecessary write of an empty map, which could happen if the snapshot's `volatile.attached_volumes` references deleted volumes.

tomponline: Sounds like a great comment to add there ;)

lxd/instance/drivers/driver_common.go
kadinsayani: We don't need to store pointers.

lxc/info.go
kadinsayani: Nice, I like that approach. I've also updated other bits of this PR to use more descriptive names than `ok`.

lxc/info.go
kadinsayani: Sounds good.

lxd/patches.go
kadinsayani: I haven't shortened it, but I've added the snapshot name and changed the wording, so we'll have: `Failed updating "volatile.attached_volumes" to new format: Failed setting instance snapshot <snapshot name> "volatile.attached_volumes"`. What do you think?

lxd/patches.go
kadinsayani: I've those details to logger context :)

lxd/patches.go
kadinsayani: Great idea, added.

lxd/patches.go
kadinsayani: Ack, we'll return an error here.

kadinsayani:
> LGTM thanks!

Thanks for your review and feedback!

PR #17131, opened 2025-12-02
client: Call mkdir instead of file stat

### LINKED ISSUE #16687: TOCTOU error if images are downloaded concurrently
### Please confirm

### Distribution

Ubuntu

### Distribution version

22.04

```shell
I'll add this if/when I find a reproducer
```

```shell
I'll add this if/when I find a reproducer
```

### Issue description

I noticed this in CI, when launching two LXD VMs [spread](https://github.com/dmitry-lyfar/spread) (note: we use a custom fork of spread which uses VMs, not containers):
```
Launching spread-2-ubuntu-24-04
Error: Failed instance creation: Failed to connect to simple streams server "https://cloud-images.ubuntu.com/releases": mkdir /var/snap/lxd/common/lxd/cache/c75a7cef2922039c9f4206be4b0181b39f899bd6d33524ede301f4356c724cd8: file exists
```

I can't share the link here but feel free to DM me.

### Steps to reproduce

Unfortunately I haven't been able to reproduce it locally yet. But the error message `Failed to connect to simple streams server` is unambiguous. It points to the `ConnectSimpleStreams` function. I think this block is the culprit:
https://github.com/canonical/lxd/blob/8f6b7a7ca7311990c60729d7260e0cde19b9ef5a/client/connection.go#L299-L304

I wrote a standalone version of this to confirm it fails pretty frequently:
```go
package main

import (
	"errors"
	"fmt"
	"os"
	"sync"
)

func main() {
	_ = os.Remove("/tmp/foo")

	var wg sync.WaitGroup
	wg.Add(5)
	for range 5 {
		go func() {
			_, err := os.Lstat("/tmp/foo")
			if errors.Is(err, os.ErrNotExist) {
				err := os.Mkdir("/tmp/foo", 0755)
				if err != nil {
					fmt.Fprintf(os.Stderr, "error: %v\n", err)
				}
			}
			wg.Done()
		}()
	}
	wg.Wait()
}
```

Instead it should be more like this:
```go
package main

import (
	"errors"
	"fmt"
	"os"
	"sync"
)

func main() {
	_ = os.Remove("/tmp/foo")

	var wg sync.WaitGroup
	wg.Add(5)
	for range 5 {
		go func() {
			err := os.Mkdir("/tmp/foo", 0755)
			if err != nil && !errors.Is(err, os.ErrExist) {
				fmt.Fprintf(os.Stderr, "error: %v\n", err)
			}
			wg.Done()
		}()
	}
	wg.Wait()
}
```

### Information to attach

markylaing:
When LXD is concurrently downloading images, it creates a local lock per image. The lock is keyed on image fingerprint.

Unfortunately, we need to instantiate a client before creating said lock because we need a client to resolve an image alias to a fingerprint to key the lock on.

This makes the existing stat+mkdir racey because another routine could create the directory after another routine has checked it's existence.

Closes #16687

## Checklist

### DIFF
--- a/client/connection.go
+++ b/client/connection.go
@@ -3,6 +3,7 @@ package lxd
 import (
 	"context"
 	"crypto/sha256"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -296,11 +297,9 @@ func ConnectSimpleStreams(url string, args *ConnectionArgs) (ImageServer, error)
 			cacheExpiry = time.Hour
 		}
 
-		if !shared.PathExists(cachePath) {
-			err := os.Mkdir(cachePath, 0755)
-			if err != nil {
-				return nil, err
-			}
+		err := os.Mkdir(cachePath, 0755)
+		if err != nil && !errors.Is(err, os.ErrExist) {
+			return nil, err
 		}
 
 		ssClient.SetCache(cachePath, cacheExpiry)


tomponline (approved):
Ta

jonathan-conder:
Looks good, thanks!

PR #17136, opened 2025-12-03
Storage: Adds safety check before ZFS volume restoration

### LINKED ISSUE #16441: `lxc restore` does not handle zfs snapshots properly with `remove_snapshots=true`
### Please confirm

### Distribution

Ubuntu

### Distribution version

24.04

```shell
core20  20250526     2599   latest/stable  canonical✓  base
core22  20250612     2045   latest/stable  canonical✓  base,disabled
core22  20250730     2111   latest/stable  canonical✓  base
core24  20250618     1055   latest/stable  canonical✓  base,disabled
core24  20250729     1151   latest/stable  canonical✓  base
lxd     6.5-72c8a10  35210  6/stable       canonical✓  -
snapd   2.70         24792  latest/stable  canonical✓  snapd,disabled
snapd   2.71         25202  latest/stable  canonical✓  snapd
```

```shell
N/A
```

### Issue description

LXD's zfs driver does not handle removing the snapshots during `restore` when those have copies. As a result, the `restore` command fails but it also removes the snapshots (preceding to the desired) from the instance's record (see `lxc info` on those).

### Steps to reproduce

```
lxc init ubuntu:24.04 w-1 -d root,pool=z1

lxc snapshot w-1 system
lxc snapshot w-1 ros2
lxc snapshot w-1 nav2

lxc copy w-1/nav2 w-2

# fails (expected)
lxc restore w-1 ros2
# fails (LXD renames the nav2 snapshot to deleted-* and cannot find it anymore)
lxc restore w-1 nav2
```

The pool looks like this:

```
z1/containers                                                                         200K  28.1G    24K  legacy
z1/containers/w-1                                                                     160K  28.1G   493M  legacy
z1/containers/w-1@snapshot-system                                                    13.5K      -   493M  -
z1/containers/w-1@snapshot-ros2                                                      13.5K      -   493M  -
z1/containers/w-1@deleted-d0abd73a-5859-4c5e-8a2c-a3e40e10ec86                         14K      -   493M  -
z1/containers/w-2                                                                    15.5K  28.1G   493M  legacy
```

### Information to attach

tugbataluy:
Fixes #16441

This PR adds a safety check before ZFS volume restoration to prevent attempts to delete snapshots that have dependent clones, which would cause the operation to fail. When zfs.remove_snapshots=true is used during restoration, it now validates each snapshot using the existing getClones() method and returns a clear error if any snapshot has clones, providing users with specific information about which snapshot is blocking the operation and how many clones depend on it, preventing failed restoration attempts.

## Checklist

### DIFF
--- a/test/suites/storage_driver_zfs.sh
+++ b/test/suites/storage_driver_zfs.sh
@@ -329,8 +329,53 @@ do_storage_driver_zfs() {
   lxc launch testimage c5
   lxc storage unset lxdtest-"$(basename "${LXD_DIR}")" volume.size
 
+  # Test snapshot restore behavior with dependent clones
+  # Enable remove_snapshots
+  lxc storage set lxdtest-"$(basename "${LXD_DIR}")" volume.zfs.remove_snapshots true
+
+  # Create container with multiple snapshots
+  lxc launch testimage c8
+  lxc snapshot c8 snap0
+  lxc exec c8 -- touch /root/file1
+  lxc snapshot c8 snap1
+  lxc exec c8 -- touch /root/file2
+  lxc snapshot c8 snap2
+
+  # Clone from the middle snapshot (this creates a dependency)
+  lxc copy c8/snap1 c9
+
+  # Store snapshot names before restore attempt
+  snap_list_before=$(lxc info c8 | awk '/^\s+snap/ {print $2}')
+
+  # Try to restore c8 to snap0 (should fail due to dependent clone c9 on snap1)
+  # This tests that snapshots are NOT removed from LXD records on failure
+  ! lxc restore c8 snap0 2>&1 | grep "cannot be restored due to snapshot.*having.*dependent clone" || false
+
+  # Verify that snapshots are still visible in LXD after failed restore
+  snap_list_after=$(lxc info c8 | awk '/^\s+snap/ {print $2}')
+  [ "$snap_list_before" = "$snap_list_after" ] || return 1
+
+  # Also verify that the ZFS snapshots still exist
+  for snap in snap0 snap1 snap2; do
+    zfs list -H -o name "lxdtest-$(basename "${LXD_DIR}")/containers/c8@snapshot-${snap}" 2>&1 || return 1
+  done
+
+  # Delete the dependent clone to allow restoration
+  lxc delete -f c9
+
+  # Now restore should work since dependency is gone
+  lxc restore c8 snap0
+
+  # Verify c8 has no files after restore to snap0
+  ! lxc exec c8 -- test -f /root/file1 || false
+  ! lxc exec c8 -- test -f /root/file2 || false
+
   # Clean up
-  lxc rm -f c1 c3 c11 c21 c4 c5 c6 c7
+  lxc delete -f c8
+  lxc storage unset lxdtest-"$(basename "${LXD_DIR}")" volume.zfs.remove_snapshots
+
+  # Clean up
+  lxc delete -f c1 c3 c11 c21 c4 c5 c6 c7
   lxc storage volume rm lxdtest-"$(basename "${LXD_DIR}")" vol1
   lxc storage volume rm lxdtest-"$(basename "${LXD_DIR}")" vol2
 


lxd/storage/drivers/driver_zfs_volumes.go
```diff
@@ -3398,6 +3398,20 @@ func (d *zfs) restoreVolume(vol Volume, snapVol Volume, migration bool, op *oper
 			return fmt.Errorf("Snapshot %q cannot be restored due to subsequent snapshot(s). Set zfs.remove_snapshots to override", snapshotName)
 		}
 
+		// Check if we can actually delete these snapshots
+		for _, snapName := range snapshots {
+			snapshotPath := fmt.Sprintf("%s@snapshot-%s", d.dataset(vol, false), snapName)
```
simondeziel: We try to avoid `fmt.Sprintf` whenever possible. Since this is handling strings only, please use `+` to concatenate them.

tugbataluy: Ok, but why are we try to avoid from it?

simondeziel: It's to avoid the overhead when not warranted. This is the initial blog that raised our awareness on this: https://www.dolthub.com/blog/2024-11-08-sprintf-vs-concat/

simondeziel: Apparently things have improved with Go 1.23 for `fmt.Sprintf`: https://markaicode.com/go-1-23-fmt-sprintf-performance-boost/

tomponline: It also helps to avoid the implicit use of data type's `String()` functions.

tugbataluy: I did my tiny research and I realized that, I should use + for simple string-only concatenation, and fmt.Sprintf when I need formatting, type conversion, or complex string building with 3+ variables. Thanks Simon :)

lxd/storage/drivers/driver_zfs_volumes.go
```diff
@@ -3398,6 +3398,20 @@ func (d *zfs) restoreVolume(vol Volume, snapVol Volume, migration bool, op *oper
 			return fmt.Errorf("Snapshot %q cannot be restored due to subsequent snapshot(s). Set zfs.remove_snapshots to override", snapshotName)
 		}
 
+		// Check if we can actually delete these snapshots
+		for _, snapName := range snapshots {
+			snapshotPath := d.dataset(vol, false) + "@snapshot-" + snapName
+			clones, err := d.getClones(snapshotPath)
+
```
MusicDin: Looking good. My only comment is to get rid of empty line between error assignment and check.

```go
clones, err := d.getClones(snapshotPath)
if err != nil {
...
```

simondeziel (dismissed):

mseralessandri (approved):

tomponline (changes requested):
Please can you add a test for this check to `test_storage_driver_zfs`

lxd/storage/drivers/driver_zfs_volumes.go
```diff
@@ -3398,6 +3398,19 @@ func (d *zfs) restoreVolume(vol Volume, snapVol Volume, migration bool, op *oper
 			return fmt.Errorf("Snapshot %q cannot be restored due to subsequent snapshot(s). Set zfs.remove_snapshots to override", snapshotName)
 		}
 
+		// Check if we can actually delete these snapshots
+		for _, snapName := range snapshots {
+			snapshotPath := d.dataset(vol, false) + "@snapshot-" + snapName
+			clones, err := d.getClones(snapshotPath)
+			if err != nil {
+				return fmt.Errorf("Failed to check clones for snapshot %q: %w", snapName, err)
```
tomponline: We tend to use this form:

> Failed checking ...

lxd/storage/drivers/driver_zfs_volumes.go
```diff
@@ -3398,6 +3398,19 @@ func (d *zfs) restoreVolume(vol Volume, snapVol Volume, migration bool, op *oper
...
+			}
+
+			if len(clones) > 0 {
+				return fmt.Errorf("Snapshot %q has %d dependent clone(s), cannot delete", snapName, len(clones))
```
tomponline: Lets keep the error here consistent with the earlier error:

> Snapshot %q cannot be restored due to...

Suggest:

> Snapshot %q cannot be restored due to snapshot %q having %d dependent clone(s) which prevents deletion

MusicDin: I assume the suggestion is without `snapshot %q` twice:
```
Snapshot %q cannot be restored due to having %d dependent clone(s) which prevents deletion
```

tomponline: I'm not sure, the caller is asking to restore snapshot A but snapshot B might prevent that as it needs to be deleted and cant be. 

So it would be:

> Snapshot A cannot be restored due to snapshot B having N dependent clone(s) which prevents deletion

MusicDin: Makes sense, thanks for clarification

test/suites/storage_driver_zfs.sh
```diff
@@ -325,12 +237,51 @@ do_storage_driver_zfs() {
...
+
+  # Clone from the middle snapshot
+  lxc copy c8/snap2 c9
+
+  # Verify c9 has file2 but not file3
+  lxc exec c9 -- test -f /root/file2
+  ! lxc exec c9 -- test -f /root/file3 || false
+
+  # Try to restore c8 to snap1 (should fail due to subsequent snapshots)
+  ! lxc restore c8 snap1 2>/dev/null || false
```
simondeziel: We rarely suppress expected error output as those are useful when navigating CI logs.

test/suites/storage_driver_zfs.sh
```diff
@@ -325,12 +237,51 @@ do_storage_driver_zfs() {
...
+  lxc exec c8 -- test -f /root/file3
+  ! lxc exec c8 -- test -f /root/file2_modified || false
+
+  lxc storage set lxdtest-"$(basename "${LXD_DIR}")" volume.size=5GiB
   lxc launch testimage c5
   lxc storage unset lxdtest-"$(basename "${LXD_DIR}")" volume.size
 
   # Clean up
-  lxc rm -f c1 c3 c11 c21 c4 c5 c6 c7
+  lxc rm -f c1 c3 c11 c21 c4 c5 c6 c7 c8 c9
```
simondeziel: While in there, please use `lxc delete`. We try to avoid aliases as much as possible as it makes it harder to track the lifecycle of instances.

test/suites/storage_driver_zfs.sh
```diff
@@ -325,12 +237,51 @@ do_storage_driver_zfs() {
   lxc exec c4 -- test -f /root/foo
   ! lxc exec c4 -- test -f /root/bar || false
 
-  lxc storage set lxdtest-"$(basename "${LXD_DIR}")" volume.size=1GiB
+  # Snapshot restore after cloning
+  # Create container with multiple snapshots
+  lxc launch testimage c8
+  lxc exec c8 -- touch /root/file1
+  lxc snapshot c8 snap1
```
simondeziel, line 340: LXD automatic snapshot would be `snap0` so for clarity, I'd either rely on automatic snapshot names or mimic the numbering scheme.

test/suites/storage_driver_zfs.sh
```diff
@@ -326,11 +326,51 @@ do_storage_driver_zfs() {
   ! lxc exec c4 -- test -f /root/bar || false
 
   lxc storage set lxdtest-"$(basename "${LXD_DIR}")" volume.size=1GiB
+  # Snapshot restore after cloning
+  # Create container with multiple snapshots
+  lxc launch testimage c8
+  lxc exec c8 -- touch /root/file1
```
simondeziel: `file1` is never used so I'd skip creating it.

test/suites/storage_driver_zfs.sh
```diff
@@ -326,11 +326,51 @@ do_storage_driver_zfs() {
...
+  lxc snapshot c8 snap0
+  lxc exec c8 -- touch /root/file2
```
simondeziel: Aligning the `fileX` with `snapX` that contains/includes it would make it more readable.

test/suites/storage_driver_zfs.sh
```diff
@@ -326,11 +326,51 @@ do_storage_driver_zfs() {
...
+  lxc copy c8/snap1 c9
+
+  # Verify c9 has file2 but not file3
+  lxc exec c9 -- test -f /root/file2
+  ! lxc exec c9 -- test -f /root/file3 || false
+
+  # Try to restore c8 to snap0 (should fail due to subsequent snapshots)
+  ! lxc restore c8 snap0 || false
+
+  # Try to restore c8 to snap1 (should work - snap3 is after but we're restoring to snap1)
```
simondeziel: The comment needs to be updated, `snap3` is no more.

test/suites/storage_driver_zfs.sh
```diff
@@ -326,11 +326,51 @@ do_storage_driver_zfs() {
...
   lxc storage set lxdtest-"$(basename "${LXD_DIR}")" volume.size=1GiB
+  # Snapshot restore after cloning
+  # Create container with multiple snapshots
+  lxc launch testimage c8
+  lxc exec c8 -- touch /root/file1
+  lxc snapshot c8 snap0
+  lxc exec c8 -- touch /root/file2
+  lxc snapshot c8 snap1
+  lxc exec c8 -- touch /root/file3
+  lxc snapshot c8 snap2
```
simondeziel: I would touch one less file, the first as it's never used afterward.

I'd also suggest having the convention that `fileX` is included in `snapX`, it's easier to track.

test/suites/storage_driver_zfs.sh
simondeziel: The comment needs an update (`snap3` is no more).

test/suites/storage_driver_zfs.sh
```diff
@@ -326,11 +326,51 @@ do_storage_driver_zfs() {
...
+  # Now try to restore c8 to snap2 (should work - it's the latest)
+  lxc exec c8 -- touch /root/file2_modified  # Modify to differentiate
+  lxc snapshot c8 snap1_modified
+  lxc restore c8 snap2
+
+  # Verify c8 has the original snap2 state (file3, no file2_modified)
+  lxc exec c8 -- test -f /root/file3
+  ! lxc exec c8 -- test -f /root/file2_modified || false
+
+  lxc storage set lxdtest-"$(basename "${LXD_DIR}")" volume.size=5GiB
```
simondeziel: A quick comment as to why that is needed would help.

test/suites/storage_driver_zfs.sh
```diff
@@ -326,11 +326,51 @@ do_storage_driver_zfs() {
   ! lxc exec c4 -- test -f /root/bar || false
 
   lxc storage set lxdtest-"$(basename "${LXD_DIR}")" volume.size=1GiB
```
simondeziel, line 328: I think this was specifically for testing with `c5` only. Now, that 1GiB is used for `c8` and `c9`, is that intentional?

tugbataluy: No it wasn't :)

test/suites/storage_driver_zfs.sh
```diff
@@ -325,12 +325,50 @@ do_storage_driver_zfs() {
   lxc exec c4 -- test -f /root/foo
   ! lxc exec c4 -- test -f /root/bar || false
 
-  lxc storage set lxdtest-"$(basename "${LXD_DIR}")" volume.size=1GiB
```
simondeziel, line 328: I'm not sure this needed to be moved.

tomponline: @tugbataluy what was the outcome of this point?

test/suites/storage_driver_zfs.sh
```diff
@@ -325,12 +325,50 @@ do_storage_driver_zfs() {
...
+
+  # Verify c9 has file1 but not file2
+  lxc exec c9 -- test -f /root/file1
+  ! lxc exec c9 -- test -f /root/file2 || false
+
+  # Try to restore c8 to snap0 (should fail due to subsequent snapshots)
+  ! lxc restore c8 snap0 || false
+
+  # Try to restore c8 to snap1 (should work - snap2 is after but we're restoring to snap1)
+  lxc restore c8 snap1
```
tomponline: @tugbataluy does this implicitly test that snap1 wasnt removed as part of the `lxc restore c8 snap0` command above? Would be good to add a comment as such to this.

tugbataluy: Yes, it tests this situation. I will add a comment about this, thank you.

tomponline:
@tugbataluy thanks!

To aid review, when you have resolved a comment from a reviewer please can you mark the thread as resolved. Ta

test/suites/storage_driver_zfs.sh
tugbataluy: I left these commands in the main branch where they were. I added my own test steps below them.

tomponline: this line is still being moved though

tugbataluy: Sorry, I guess I fixed right now

PR #17141, opened 2025-12-03
Use `microovn` instead of OVS deb

simondeziel:


simondeziel:
@tomponline this fixes the `group:network` that is baking in another PR, yay!

PR #17143, opened 2025-12-03
lxd/instance/drivers: Ignore expected ErrInUse

kadinsayani:


lxd/instance/drivers/driver_common.go
```diff
@@ -1098,7 +1099,8 @@ func (d *common) snapshotCommon(inst instance.Instance, name string, expiry *tim
 
 	defer func() {
 		err := pool.UnmountInstance(inst, d.op)
-		if err != nil {
+		// Unmounting volumes while an instance is running is expected to return [storageDrivers.ErrInUse].
```
tomponline, line 1102: why are we even trying to unmount instance if its running?

kadinsayani: Because earlier in `snapshotCommon()` we call `pool.MountInstance()` (this increments the ref count) to write to backup.yaml, so we need to defer a call to `pool.UnmountInstance()` to decrement the ref count. If the instance is running, the ref count will still be > 0, hence why `ErrInUse` is returned.

kadinsayani: I've added this to the commit description.

tomponline: Very good :)

PR #17156, opened 2025-12-03
Use the snap cache in snap tests

simondeziel:


simondeziel:
This [fixes the problem it needed to](https://github.com/canonical/lxd/actions/runs/19911883015/job/57082750892?pr=17156#step:23:89):
```
+ '[' -e ./lxd_36942.snap ']'
+ echo 'Installing lxd from cache'
+ snap ack ./lxd_36942.assert
Installing lxd from cache
+ snap install ./lxd_36942.snap
lxd git-8d7d5a8 from Canonical** installed
```

While previously all the `Snap` tests were downloading the snaps opportunistically... it showed the fallback of opportunistically downloading worked well :)

PR #17158, opened 2025-12-04
doc: baseurl and sitemap improvements

minaelee:
## PR description

Replaces the statically set value for `html_baseurl` to use RTD environment variable, or `"/"` if not available (for local builds). This ensures that the HTML metadata canonical URL is set correctly, including the version in the URL. This change required also updating the `sitemap_url_scheme` to prevent duplicating the version string, since the sitemap extension uses `html_baseurl` to create the link.

References:
- https://docs.readthedocs.com/platform/stable/canonical-urls.html
- https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_baseurl

Also:
- Adds last modified timestamps to sitemap URLs to improve SEO. This used to be the default for the sitemap but was changed upstream. Requires setting `.readthedocs.yaml` to use a non-shallow clone (see previous discussion in #16249).
- Adds a list of files to exclude from the sitemap.

## Checklist

### DIFF
--- a/doc/.readthedocs.yaml
+++ b/doc/.readthedocs.yaml
@@ -14,6 +14,7 @@ build:
   jobs:
     pre_install:
       - pip install gitpython pyyaml
+      - git fetch --unshallow || true
     pre_build:
       - go build -ldflags "-s -w" -o trimpath -o lxc.bin ./lxc
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -108,15 +108,25 @@ slug = "lxd"
 # Sitemap configuration: https://sphinx-sitemap.readthedocs.io/
 #######################
 
-# Use RTD canonical URL to ensure duplicate pages have a single canonical URL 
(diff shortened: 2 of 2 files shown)


doc/.readthedocs.yaml
```diff
@@ -14,6 +14,7 @@ build:
   jobs:
     pre_install:
       - pip install gitpython pyyaml
+      - git fetch --unshallow || true
```
tomponline, line 17: what is the `|| true` bit for by the way? Seems like we would want to fail if that git command fails?

minaelee: My understanding is that it's a safeguard for if somehow the git clone is already unshallowed, in which case the command will fail. RTD recommends it in its [example](https://docs.readthedocs.com/platform/stable/build-customization.html#unshallow-git-clone) ([commit](https://github.com/readthedocs/readthedocs.org/pull/10485/commits/f3d856abc1064ad7704c85fd1a0e918e1b689c20)).

PR #17161, opened 2025-12-04
Test: Cleanup all daemons in clustering tests

roosterfish:
Doesn't cause any functional issues but it seems `test_clustering_evacuation_restore_operations` is the only suite which doesn't cleanup all of the spawned cluster members.


simondeziel (approved):
Very nice! Thanks

PR #17163, opened 2025-12-04
test/suites/storage_driver_zfs: fix `/dev/zfs` permissions if `udev`  didn't do it

simondeziel:
This is similar to what was needed in `lxd-ci`:
https://github.com/canonical/lxd-ci/pull/563

### DIFF
--- a/test/suites/storage_driver_zfs.sh
+++ b/test/suites/storage_driver_zfs.sh
@@ -21,6 +21,18 @@ do_zfs_delegate() {
     return
   fi
 
+  # XXX: Ensure that `/dev/zfs` has mode 0666 so that any user on the system
+  #      can interact with it. Setting those permissions is udev's job but the
+  #      needed rule ships in the `zfsutils-linux` which might be installed after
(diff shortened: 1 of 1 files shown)


test/suites/storage_driver_zfs.sh
```diff
@@ -21,6 +21,18 @@ do_zfs_delegate() {
     return
   fi
 
+  # XXX: Ensure that `/dev/zfs` has mode 0666 so that any user on the system
```
tomponline, line 24: how come we arent seeing test failures btw?

simondeziel: because we install `zfsutils-linux` before loading `zfs.ko` so that udev is able to be notified and fix it for us.

In local ephemeral VMs, we routinely run into it as it was first noticed in `lxd-ci`.

tomponline: did you notice failures btw?

simondeziel: Yes, when trying to reproduce your other issue this morning.

PR #17168, opened 2025-12-05
Storage: Drop CephFS `volatile.pool.pristine` config key

roosterfish:
The `volatile.pool.pristine` is allowed to be set but not used by the CephFS driver.

### DIFF
--- a/lxd/patches.go
+++ b/lxd/patches.go
@@ -110,6 +110,7 @@ var patches = []patch{
 	{name: "config_remove_instances_placement_scriptlet", stage: patchPreLoadClusterConfig, run: patchRemoveInstancesPlacementScriptlet},
 	{name: "event_entitlement_rename", stage: patchPreLoadClusterConfig, run: patchEventEntitlementNames},
 	{name: "pool_fix_default_permissions", stage: patchPostDaemonStorage, run: patchDefaultStoragePermissions},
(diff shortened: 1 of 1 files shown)


lxd/patches.go
```diff
@@ -1845,4 +1846,44 @@ func patchDefaultStoragePermissions(_ string, d *Daemon) error {
 	return nil
 }
 
+// patchUnsetCephFSPristineSetting unsets the volatile.pool.pristine setting from all CephFS storage pool's configs.
+func patchUnsetCephFSPristineSetting(_ string, d *Daemon) error {
+	err := d.State().DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+		var err error
+
+		// Get all storage pool names.
+		pools, _, err := tx.GetStoragePools(ctx, nil)
```
tomponline: can we do this in a single query using a DELETE with sub-query?

roosterfish: Yeah good point, fixed and much simpler now.

tomponline (dismissed):
Thanks!

The indentation of the query made me a bit sad though.

roosterfish:
> The indentation of the query made me a bit sad though.

No reason to be sad, updated :)

tomponline (approved):
Thanks!

PR #17198, opened 2025-12-08
Backports (stable-5.21)

tomponline:


tomponline:
@minaelee @roosterfish are the DR doc commits in this PR suitable to backport to 5.21?

@minaelee also I've updated some links in the new docs build rules to reference stable-5.21 docs.

tomponline:
> Those three doc commits about recovery make sense. However you cannot recover custom vols from PowerFlex/Pure as 5.21 doesn't store the new metadata format on file to allow recovery of custom vols from an instance's backup config:
> 
>     * [09a57c0](https://github.com/canonical/lxd/commit/09a57c079ab3c273ccc79309bc3ba3d18390dfde)
> 
>     * [94e7ad6](https://github.com/canonical/lxd/commit/94e7ad612a7e7d924b90dc188b726868931f1177)
> 
>     * [9836352](https://github.com/canonical/lxd/commit/98363521fd13dc75b901c02875794310c9d6d004)
> 
> 
> The footnote says this:
> 
> ```
> [6] (1, 2)
> 
> Custom volumes can only be recovered when attached to an instance due to the use of transformed volume names.
> ```
> 
> So we could drop [09a57c0](https://github.com/canonical/lxd/commit/09a57c079ab3c273ccc79309bc3ba3d18390dfde) which would not extend the table but also not introduce a comment about custom volume recovery which is not possible on 5.21 for PowerFlex/Pure.

As discussed, we modified the point 6 footnote to be relevant for 5.21.

Issue #17167, opened 2025-12-04
nvidia support broken for docker inside LXC containers with latest LXD


jocado:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu Core

### Distribution version

24

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
Name    Version      Rev    Tracking          Publisher   Notes
core22  20250822     2133   latest/stable     canonical✓  base,disabled
core22  20250923     2139   latest/stable     canonical✓  base,disabled
core22  20251009     2163   latest/stable     canonical✓  base
core24  20250618     1055   latest/stable     canonical✓  base,disabled
core24  20250829     1196   latest/stable     canonical✓  base,disabled
core24  20251001     1225   latest/stable     canonical✓  base
lxd     6.6-3c9aa6d  36883  latest/candidate  canonical✓  disabled,ignore-validation
lxd     6.6-f7f9c85  36968  latest/candidate  canonical✓  disabled,ignore-validation
lxd     6.6-5b6f7fb  37002  latest/candidate  canonical✓  ignore-validation
snapd   2.67.1       23771  latest/stable     canonical✓  snapd
```

### Output of "lxc info" or system info if it fails

```shell
api_extensions:
- storage_zfs_remove_snapshots
- container_host_shutdown_timeout
- container_stop_priority
- container_syscall_filtering
- auth_pki
- container_last_used_at
- etag
- patch
- usb_devices
- https_allowed_credentials
- image_compression_algorithm
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
- unix_device_rename
- storage_lvm_use_thinpool
- storage_rsync_bwlimit
- network_vxlan_interface
- storage_btrfs_mount_options
- entity_description
- image_force_refresh
- storage_lvm_lv_resizing
- id_map_base
- file_symlinks
- container_push_target
- network_vlan_physical
[...]

PR #17222, opened 2025-12-10
Nvidia CDI: Update symlink creation inside LXD container

nmezhenskyi:
This PR updates symlink creation inside a LXD container when applying CDI hooks. It resolves the link's target path to a path relative to the link's directory.

Fixes https://github.com/canonical/lxd/issues/17167.

### DIFF
--- a/lxd/main_callhook.go
+++ b/lxd/main_callhook.go
@@ -42,25 +42,30 @@ func (c *cmdCallhook) Command() *cobra.Command {
 	return cmd
 }
 
-// resolveTargetRelativeToLink converts a target relative to a link path into an absolute path.
+// resolveTargetRelativeToLink converts a link's target into a path relative to the link's path.
 func resolveTargetRelativeToLink(link string, target string) (string, error) {
 	if !filepath.IsAbs(link) {
 		return "", fmt.Errorf("The link must be an absolute path: %q (target: %q)", link, target)
 	}
 
-	if filepath.IsAbs(target) {
+	// If target is already relative, return as-is.
+	if !filepath.IsAbs(target) {
 		return target, nil
 	}
 
-	linkDir := filepath.Dir(link)
-	absTarget := filepath.Join(linkDir, target)
-	cleanPath := filepath.Clean(absTarget)
-	absPath, err := filepath.Abs(cleanPath)
+	// Clean both paths to normalize them.
+	linkClean := filepath.Clean(link)
+	targetClean := filepath.Clean(target)
+
+	linkDir := filepath.Dir(linkClean)
+
+	// Calculate the relative path from link's directory to the target.
+	relPath, err := filepath.Rel(linkDir, targetClean)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


lxd/main_callhook.go
```diff
@@ -101,25 +101,36 @@ func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) er
...
 
+		linkDir := filepath.Dir(symlink.Link)
+		targetDir := filepath.Dir(target)
+
+		// If link location and target share the same dir, make target path relative.
+		if linkDir == targetDir {
+			relTarget, err := filepath.Rel(linkDir, target)
+			if err == nil {
+				target = relTarget
+			}
```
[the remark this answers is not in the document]
nmezhenskyi: Fixed

lxd/main_callhook.go
```diff
@@ -101,25 +101,38 @@ func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) er
 	// Creating the symlinks
 	for _, symlink := range hooks.Symlinks {
 		// Resolve hook link from target
-		absTarget, err := resolveTargetRelativeToLink(symlink.Link, symlink.Target)
+		target, err := resolveTargetRelativeToLink(symlink.Link, symlink.Target)
```
tomponline, line 109: Hang on.

The `resolveTargetRelativeToLink` function's docblock says:

> // resolveTargetRelativeToLink converts a target relative to a link path into an absolute path.

If the link is already relative, why are we resolving it to an absolute path?

Do we even need this call, and your modification?

nmezhenskyi: > Do we even need this call, and your modification?

@tomponline This is a good point. It doesn't seem like we need the call to `resolveTargetRelativeToLink`. I've tried doing it both ways and it works even with `resolveTargetRelativeToLink` removed or modified to try and return relative link if possible.  

The main reason I went with the current approach is to preserve a default of absolute path for situations where link and target are not in the same directory. However, now that I'm thinking again about this, it seems unnecessary.

> If the link is already relative, why are we resolving it to an absolute path?

I'm not exactly sure because currently for Nvidia CDI we don't seem to rely on any absolute path links. I couldn't find an explanation for this in the PR that added this.

Do you think we should use links provided to the hook as-is or try to resolve them to relative even if they're not?

tomponline: > Do you think we should use links provided to the hook as-is or try to resolve them to relative even if they're not?

It sounds like we should resolve and convert them to relative to support docker snap.

nmezhenskyi:
> Please can you follow up with some unit tests?

@tomponline Will do! Should I add it here or as a follow-up PR?

tomponline:
> > Please can you follow up with some unit tests?
> 
> @tomponline Will do! Should I add it here or as a follow-up PR?

You can open a follow up PR as then we can get this merged and tested with @jocado in latest/edge sooner.

PR #17226, opened 2025-12-10
Dependabot tweaks

simondeziel:


.github/dependabot.yml
tomponline, line 1: the commit message here would have benefitted from having a **why** so in the future we can see why this was added.

PR #17228, opened 2025-12-11
Cluster membership improvements and Assign() refactorings

kadinsayani:
- Improves error messages by changing them to gerund style for consistency.
- Refactors the `Assign()` function for code clarity.
- Adds comments to explain logic for stepping down from voter -> stand-by -> spare.

## Checklist

### DIFF
--- a/lxd/cluster/membership.go
+++ b/lxd/cluster/membership.go
@@ -829,13 +829,13 @@ func Assign(state *state.State, gateway *Gateway, nodes []db.RaftNode) error {
 		defer func() { _ = client.Close() }()
 
 		// Figure out our current role.
-		role := db.RaftRole(-1)
-		cluster, err := client.Cluster(ctx)
+		role := db.RaftRole(-1) // -1 is an invalid role.
+		servers, err := client.Cluster(ctx)
 		if err != nil {
 			return fmt.Errorf("Fetch current cluster configuration: %w", err)
 		}
 
-		for _, server := range cluster {
+		for _, server := range servers {
 			if server.ID == info.ID {
 				role = server.Role
 				break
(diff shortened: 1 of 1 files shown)


lxd/cluster/membership.go
```diff
@@ -829,13 +829,13 @@ func Assign(state *state.State, gateway *Gateway, nodes []db.RaftNode) error {
 		defer func() { _ = client.Close() }()
 
 		// Figure out our current role.
-		role := db.RaftRole(-1)
-		cluster, err := client.Cluster(ctx)
+		role := db.RaftRole(-1) // -1 is an invalid role.
+		servers, err := client.Cluster(ctx)
```
tomponline, line 833: `members` would have been more inline with our existing terminology. But `servers` is certainly still clearer than `cluster`, so happy to keep as-is.

lxd/cluster/membership.go
```diff
@@ -847,8 +847,28 @@ func Assign(state *state.State, gateway *Gateway, nodes []db.RaftNode) error {
 
 		// If we're stepping back from voter to spare, let's first transition
 		// to stand-by first and wait for the configuration change to be
-		// notified to us. This prevent us from thinking we're still voters and
-		// potentially disrupt the cluster.
+		// notified to us. This prevents us from thinking we're still a voter and
+		// potentially disrupting the cluster.
+		//
+		// If we went from voter -> spare directly:
```
tomponline, line 853: Excellent detail, thanks!

tomponline: Although its unfortunate go-dqlite doesn't prevent this sort of thing internally and requires each application to know these fine details.

lxd/cluster/membership.go
kadinsayani: I considered going with `members` or `clusterMembers`, but ended up going with `servers` since that's what `go-dqlite/client.Cluster()` calls the returned variable. While we don't have to follow terminology used by packages, I thought the slight differentiation could be useful in this case.

PR #17241, opened 2025-12-11
Inital pass to drop translations

simondeziel:
The aim of this PR is to reduce the friction introduced by `i18n` related parts. This should reduce the risk of conflicts between multiple in-flight PRs, make reviewers job easier and help Copilot reviews as well due to reducing the number of file changed.

This is just the first step in a longer process of removing all the `i18n.G()` calls.

Related to https://github.com/canonical/lxd/issues/17066

### DIFF
--- a/test/lint/bin-size.sh
+++ b/test/lint/bin-size.sh
@@ -12,7 +12,7 @@ echo "Check binaries size limits"
 
 # bin/max (sizes are in MiB)
 declare -rA sizes=(
-    ["lxc"]=17
+    ["lxc"]=16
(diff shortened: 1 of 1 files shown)


lxc/utils_properties.go
```diff
@@ -219,12 +217,12 @@ func unpackKVToWritable(writable any, keys map[string]string) error {
 		),
 	})
 	if err != nil {
-		return fmt.Errorf(i18n.G("Error creating decoder: %v"), err)
```
tomponline, line 222: should be %w rather than %v really.

PR #17242, opened 2025-12-11
make: Drop tag for golangci-lint

roosterfish:
Fixes https://github.com/canonical/lxd/actions/runs/20138136599/job/57797322714:

```
golangci/golangci-lint info checking GitHub for tag 'latest'
golangci/golangci-lint crit unable to find 'latest' - use 'latest' or see https://github.com/golangci/golangci-lint/releases for details
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -472,7 +472,7 @@ ifeq "$(LXD_OFFLINE)" ""
 	go install fillmore-labs.com/zerolint@latest
 endif
 ifeq ($(shell command -v golangci-lint),)
-	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(GOPATH)/bin latest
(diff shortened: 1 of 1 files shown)


roosterfish:
Looks like this actively changes every now and then. I just had another test run that succeeding over on MicroCloud without this fix. Leaving it as draft for now.

simondeziel:
> Looks like this actively changes every now and then. I just had another test run that succeeding over on MicroCloud without this fix. Leaving it as draft for now.

I just noticed the same. That said, without `latest` it seems to do the right thing of defaulting to `latest`:

```
usage() {
  this=$1
  cat <<EOF
$this: download go binaries for golangci/golangci-lint

Usage: $this [-b <bindir>] [-d] [<tag>]
  -b sets bindir or installation directory, Defaults to ./bin
  -d turns on debug logging
   <tag> is a tag from
   https://github.com/golangci/golangci-lint/releases
   If tag is missing, then the latest will be used.

EOF
  exit 2
}
```

Which downloads from https://github.com/golangci/golangci-lint/releases/latest that is redirecting (ATM at least) to `v2.7.2`

PR #17248, opened 2025-12-12
auth: LXD bearer tokens

MusicDin:
This PR adds support for LXD bearer tokens.

It adds new identity type `bearer`, which similarly to the identity type `devlxd` can be used to issue a bearer token for API authentication. The LXD bearer tokens are trusted only by the main LXD API, while DevLXD bearer tokens are valid only for DevLXD API.

In addition, the LXD bearer token contains server certificate fingerprint which is used to verify received server certificate during initial TLS handshake.

TODO:
- [x] API Extension (`auth_bearer_lxd`)
- [x] Docs

### DIFF
--- a/doc/authentication.md
+++ b/doc/authentication.md
@@ -14,6 +14,7 @@ The following authentication methods are supported:
 
 - {ref}`authentication-tls-certs`
 - {ref}`authentication-openid`
+- {ref}`authentication-bearer`
 
 (authentication-tls-certs)=
 ## TLS client certificates
@@ -283,6 +284,21 @@ backend lxd_cluster_tcp
   server lxd-3 1.2.3.6:8443 check send-proxy
 # EOF
 ```
+
+(authentication-bearer)=
+## Bearer token authentication
+
+LXD supports authenticating to the LXD API using bearer tokens. Bearer tokens provide a secure and temporary way to authenticate API requests without requiring client certificates.
+
+Bearer tokens can be issued for identities of type `bearer`. The permissions associated with a token are derived from the identity it belongs to and are enforced through {ref}`fine-grained-authorization`.
+
+To authenticate an API request using a bearer token, include it in the `Authorization` header
+as `Authorization: Bearer <token>`, where `<token>` represents an actual token value.
+
+By default, bearer tokens expire after 24 hours, unless they are manually revoked.
+The expiration time can be customized when issuing the token.
+
+See {ref}`howto-auth-bearer` to learn how to issue and use bearer token in LXD.
 
 ## Failure scenarios
 
--- a/doc/howto/index.md
+++ b/doc/howto/index.md
@@ -59,5 +59,6 @@ Harden security </howto/security_harden>
 /howto/troubleshoot
 Get support </support>
 Contribute to LXD </contributing>
+/howto/auth_bearer
 /howto/devlxd_authenticate
 ```


client/util.go
```diff
@@ -23,13 +24,47 @@ import (
...
 // It returns the HTTP client with the provided configurations and handles any errors that might occur during the setup process.
-func tlsHTTPClient(client *http.Client, tlsClientCert string, tlsClientKey string, tlsCA string, tlsServerCert string, insecureSkipVerify bool, legacyCurvesOnly bool, proxy func(req *http.Request) (*url.URL, error), transportWrapper func(t *http.Transport) HTTPTransporter) (*http.Client, error) {
+func tlsHTTPClient(client *http.Client, tlsClientCert string, tlsClientKey string, tlsCA string, tlsServerCert string, insecureSkipVerify bool, legacyCurvesOnly bool, proxy func(req *http.Request) (*url.URL, error), transportWrapper func(t *http.Transport) HTTPTransporter, serverCertFingerprint string) (*http.Client, error) {
 	// Get the TLS configuration
 	tlsConfig, err := shared.GetTLSConfigMem(tlsClientCert, tlsClientKey, tlsCA, tlsServerCert, insecureSkipVerify)
 	if err != nil {
 		return nil, err
 	}
 
+	if serverCertFingerprint != "" {
```
MusicDin: And if `tlsServerCert` is NOT provided.

Comment from @markylaing

client/util.go
```diff
@@ -23,13 +24,47 @@ import (
...
+		// If a server fingerprint is provided, we disable default verification and rely on our own
+		// verification logic to confirm the server's identity.
+		// We need to verify that the server's certificate matches the expected fingerprint and that
+		// the certificate is currently valid (not expired).
+		tlsConfig.InsecureSkipVerify = true
+		tlsConfig.VerifyPeerCertificate = func(rawCerts [][]byte, _ [][]*x509.Certificate) error {
```
MusicDin, line 40: Should we verify the server address in the server certificate as well?

client/connection.go
```diff
@@ -446,8 +447,28 @@ func httpsLXD(ctx context.Context, requestURL string, args *ConnectionArgs) (Ins
...
+		// try to extract the server certificate fingerprint from the token.
+		//
+		// TODO: Anonymous struct is used to avoid referencing internal lxd/auth/encryption package.
+		//       Should [encryption.LXDClaims] be moved to lxd/shared instead?
+		var claims struct {
+			jwt.RegisteredClaims
+			ServerFingerprint string `json:"server_cert_fingerprint,omitempty"`
+		}
+
+		_, _, err := jwt.NewParser().ParseUnverified(args.BearerToken, &claims)
```
MusicDin, line 465: ParseUnverified -> What if token is mangled and serverFingerprint replaced?

doc/api-extensions.md
```diff
@@ -2884,3 +2884,9 @@ Before this was only partially possible for some of the drivers (e.g. by using `
 The new pool `source.recover` configuration key can be set per cluster member to allow reuse of an existing pool `source`.
 What it does not allow is reusing the same source for multiple storage pools.
 The LVM storage driver has the specific `lvm.vg.force_reuse` configuration key for this purpose.
+
+(extension-auth-bearer-lxd)=
+## `auth_bearer_lxd`
```
tomponline: I think `auth_bearer` will suffice, the `_lxd` is redundant in my view.

shared/api/auth.go
```diff
@@ -42,6 +42,9 @@ const (
 
 	// IdentityTypeBearerTokenDevLXD represents an identity that bears a LXD token that can be used to interact with the DevLXD API.
 	IdentityTypeBearerTokenDevLXD = "DevLXD token bearer"
+
+	// IdentityTypeBearerTokenLXD represents an identity that bears a LXD token that can be used to interact with the LXD API.
+	IdentityTypeBearerTokenLXD = "LXD token bearer"
```
tomponline: "Client token bearer" to be consistent with the other external API client types?

lxd/identity/lxd_token_bearer.go
```diff
@@ -0,0 +1,32 @@
+package identity
+
+import (
+	"github.com/canonical/lxd/shared/api"
+)
+
+// LXDTokenBearer represents an identity that authenticates using a token issued by LXD
+// and whose permissions are managed via group membership. The token is valid for the LXD API.
+// It supports both caching and fine-grained permissions but is not an admin by default.
+type LXDTokenBearer struct {
```
tomponline: Just "BearerToken" will suffice, the LXD is implicit by way of it being a bearer token for LXD.

tomponline: Or BearerTokenClient to be consistent with CertificateClient

shared/api/auth.go
tomponline: or do you want to align it with the "LXD" bearer audience type?

client/connection.go
```diff
@@ -21,6 +22,16 @@ import (
 	"github.com/canonical/lxd/shared/simplestreams"
 )
 
+// LXDClaims represents the claims of a LXD bearer token.
+// They extend standard [jwt.RegisteredClaims] with custom LXD claims.
+type LXDClaims struct {
```
tomponline: "LXD" here is pretty vague/meaningless. 

Can it be "BearerTokenClaims"?

tomponline: or "ClientBearerTokenClaims"?

tomponline (changes requested):
Looking good to me.

However in general across the board I think we should be avoiding "LXD" prefixes and suffixes for the bearer token types, function names etc.

Instead I think we should go with "BearerToken" or "ClientBearerToken" and the devlxd variant of that can have "DevLXD" suffixed, rather than also having "LXD" on the external API client ones.

shared/api/auth.go
MusicDin: Actually, I followed "DevLXD" approach to be consistent how constants and functions are already named. Will use `Client` instead of LXD.

lxd/identity/type_interface.go
```diff
@@ -84,6 +84,7 @@ var types = []Type{
 	CertificateMetricsUnrestricted{},
 	CertificateServer{},
 	DevLXDTokenBearer{},
+	LXDTokenBearer{},
```
MusicDin: @markylaing @tomponline Mind if I rename both to `TokenBearer{DevLXD,Client}` for consistency?

tomponline: As long as renamimg the DevLXD isnt an SDK affecting change.

shared/api/auth.go
MusicDin: Oh, yeah the `LXD` term came from the audience

tomponline:
@MusicDin this needs a rebase.

Also I'm going to pop this back into draft until its had the security review. Ta

MusicDin:
Hi @canonical/security, friendly reminder about this PR. It would help us move forward once reviewed. 
Thanks in advance!

tomponline:
@MusicDin is this nearly ready for review? Can we land some of it? Thanks

MusicDin:
@tomponline Waiting for ~security~ compliance review. If they approve, it is ready.

0xDSousa:
At @escabo’s request, I’ve arranged for someone from Security Engineering to take a look over the coming week.

MusicDin:
Thank you @0xDSousa! 🚀

tomponline:
@MusicDin this requires a rebase, and then once done, please can you promote to "ready for review", ta

simondeziel:
The `lxd-installer` tests fail when trying to launch a 26.04 VM and then trying to use ZFS. The problem is the Resolute kernel (6.19) just started providing ZFS 2.4 which is not yet supported by `lxd-pkg-snap` (only goes up to ZFS 2.3). I'll fix it in another PR.

tomponline:
> The `lxd-installer` tests fail when trying to launch a 26.04 VM and then trying to use ZFS. The problem is the Resolute kernel (6.19) just started providing ZFS 2.4 which is not yet supported by `lxd-pkg-snap` (only goes up to ZFS 2.3). I'll fix it in another PR.

Thanks!

lxc/config/remote.go
```diff
@@ -392,6 +395,14 @@ func (c *Config) getConnectionArgs(name string) (*lxd.ConnectionArgs, error) {
 		return &args, nil
 	}
 
+	// Check for LXD_AUTH_BEARER_TOKEN environment variable
+	// Stop here if bearer token is used. It takes precedence over the TLS client certificate and key.
+	token, ok := os.LookupEnv("LXD_AUTH_BEARER_TOKEN")
+	if ok {
```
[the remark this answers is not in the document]
tomponline: @MusicDin this seems valid problem

test/suites/auth.sh
```diff
@@ -190,6 +206,27 @@ fine_grained: true"
...
+- test-group
+tls_certificate: ""
+effective_groups:
+- test-group
+effective_permissions: []
+fine_grained: true
+EOF
+)
+
+  # When bearer token is passed using environemnt variable, it will take precedence over the OIDC/TLS
```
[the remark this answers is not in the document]
tomponline: @MusicDin typo

test/suites/auth.sh
```diff
@@ -190,6 +206,27 @@ fine_grained: true"
...
+  # authentication, however, we need HTTPS remote to get the identity info. We simply reuse the remote
+  # used for testing TLS authentication.
+  [ "$(LXD_CONF="${LXD_CONF4}" LXD_AUTH_BEARER_TOKEN=${bearer_identity_token} lxc auth identity info bearer:)" = "${expectedBearerInfo}" ]
```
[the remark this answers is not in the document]
tomponline: @MusicDin this seems valid problem

test/suites/auth.sh
```diff
@@ -734,7 +807,11 @@ user_is_server_admin() {
   certificate_add_token="$(lxc_remote config trust add "${remote}:" --name test --quiet)"
 
   # The token works.
-  LXD_CONF="${TMP_LXD_CONF}" lxc_remote remote add test-remote "${certificate_add_token}"
+  # Make sure to the bearer token does not take precedence remote removal.
```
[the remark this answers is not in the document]
tomponline: @MusicDin agree this can be clarified.

lxd/auth/encryption/sign.go
```diff
@@ -1,12 +1,15 @@
...
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"time"
 
 	"github.com/golang-jwt/jwt/v5"
 	"github.com/google/uuid"
+
+	"github.com/canonical/lxd/client"
```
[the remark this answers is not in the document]
tomponline: this is fine

lxd/identities.go
```diff
@@ -505,7 +505,24 @@ func identityBearerTokenPost(d *Daemon, r *http.Request) response.Response {
...
-	token, err := encryption.GetDevLXDBearerToken(secret, id.Identifier, s.GlobalConfig.ClusterUUID(), expiresAt)
+	var token string
+	switch id.Type {
+	case api.IdentityTypeBearerTokenClient:
+		var serverCertFingerprint string
+
+		// When creating LXD bearer tokens, include the server certificate fingerprint.
+		serverCertFingerprint, err = shared.CertFingerprintStr(string(s.Endpoints.NetworkPublicKey()))
+		if err != nil {
+			return response.SmartError(fmt.Errorf("Failed to parse server certificate fingerprint: %w", err))
```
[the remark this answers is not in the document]
tomponline: this is standard

client/util.go
```diff
@@ -23,13 +24,47 @@
...
 	if err != nil {
 		return nil, err
 	}
 
+	if serverCertFingerprint != "" {
+		// If a server fingerprint is provided, we disable default verification and rely on our own
+		// verification logic to confirm the server's identity.
+		// We need to verify that the server's certificate matches the expected fingerprint and that
+		// the certificate is currently valid (not expired).
+		tlsConfig.InsecureSkipVerify = true
```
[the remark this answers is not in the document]
tomponline: this is needed for custom verification

client/util.go
```diff
@@ -23,13 +24,51 @@ import (
...
+			// Parse the server certificate.
+			serverCert, err := x509.ParseCertificate(rawCerts[0])
+			if err != nil {
+				return fmt.Errorf("Failed to parse server certificate: %w", err)
+			}
+
+			// Verify cert is valid.
+			now := time.Now()
+			if now.Before(serverCert.NotBefore) {
+				return fmt.Errorf("Server certificate is not valid at %s (valid after %s)", now.Format(time.RFC3339), serverCert.NotBefore.Format(time.RFC3339))
```
tomponline: can we use `%q` for strings in errors please

client/util.go
```diff
@@ -23,13 +24,51 @@ import (
...
+			}
+
+			if now.After(serverCert.NotAfter) {
+				return fmt.Errorf("Server certificate is not valid at %s (expired at %s)", now.Format(time.RFC3339), serverCert.NotAfter.Format(time.RFC3339))
```
tomponline: can we use `%q` for strings in errors please

client/util.go
```diff
@@ -23,13 +24,51 @@ import (
...
+			}
+
+			// Verify fingerprint.
+			fingerprint := shared.CertFingerprint(serverCert)
+			if !strings.EqualFold(fingerprint, serverCertFingerprint) {
+				return fmt.Errorf("Server certificate fingerprint mismatch: got %s, want %s", fingerprint, serverCertFingerprint)
```
tomponline: can we use `%q` for strings in errors please

doc/howto/auth_bearer.md
```diff
@@ -0,0 +1,47 @@
...
+    lxc auth identity create bearer/<name> [[--group <group> ]]
+```
+```{group-tab} API
+    lxc query --request POST /1.0/auth/identities/bearer --data '{
+      "name": "<name>",
+      "type": "bearer"
+      "groups": [
+        "<group>"
+      ]
+    }'
```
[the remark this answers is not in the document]
tomponline: @MusicDin this is valid problem

shared/version/api.go
```diff
@@ -472,6 +472,7 @@ var APIExtensions = []string{
 	"storage_source_recover",
 	"instance_force_delete",
 	"operation_metadata_entity_url",
+	"auth_bearer",
```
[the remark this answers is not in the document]
MusicDin: We agreed on "auth_bearer" name for the API extension

test/lint/bin-size.sh
```diff
@@ -12,7 +12,7 @@ echo "Check binaries size limits"
 
 # bin/max (sizes are in MiB)
 declare -rA sizes=(
-    ["lxc"]=16
+    ["lxc"]=17
```
simondeziel, line 15: 😱

test/suites/auth.sh
```diff
@@ -190,6 +206,27 @@ fine_grained: true"
 
   [ "$(LXD_CONF="${LXD_CONF2}" lxc auth identity info tls:)" = "${expectedTLSInfo}" ]
 
+  # Need heredoc to retain double quotes for empty TLS certificate.
+  expectedBearerInfo=$(cat<<EOF
```
simondeziel: No need for `$(cat <<EOF` as `"` can spawn multiple lines.

test/suites/auth.sh
```diff
@@ -734,7 +807,11 @@ user_is_server_admin() {
   certificate_add_token="$(lxc_remote config trust add "${remote}:" --name test --quiet)"
 
   # The token works.
-  LXD_CONF="${TMP_LXD_CONF}" lxc_remote remote add test-remote "${certificate_add_token}"
+  # Make sure the bearer token does not take precedence during remote removal.
+  (
+    unset LXD_AUTH_BEARER_TOKEN
+    LXD_CONF="${TMP_LXD_CONF}" lxc_remote remote add test-remote "${certificate_add_token}"
+  )
```
simondeziel: It's not clear to me why a subshell would be needed here. Why not just `unset` it or using `LXD_AUTH_BEARER_TOKEN="" LXD_CONF="${TMP_LXD_CONF}" lxc_remote remote add test-remote "${certificate_add_token}"` ?

test/suites/devlxd.sh
```diff
@@ -280,8 +280,25 @@ test_devlxd_volume_management() {
...
+
+    # Ensure DevLXD authentication does not accept LXD bearer token.
+    # Set same permissions to the LXD bearer token as to DevLXD bearer token
+    # and ensure it is not allowed to do any operation.
+    lxc auth identity create bearer/test-with-devlxd --group "${authGroup}"
+    lxdToken=$(lxc auth identity token issue bearer/test-with-devlxd --quiet)
+
+    # First, query LXD API directly to ensure the token is trusted by LXD API.
+    lxdAddr="$(lxc config get core.https_address)"
+    curl -k "https://${lxdAddr}/1.0" -H "Authorization: Bearer ${lxdToken}" | jq --exit-status '.metadata.auth == "trusted"'
```
simondeziel: To be consistent with how `curl` is invoked in other tests:
`curl -s -k -H "Authorization: Bearer ${lxdToken}" "https://${LXD_ADDR}/1.0"`

test/suites/devlxd.sh
```diff
@@ -280,8 +280,25 @@ test_devlxd_volume_management() {
...
+    lxc exec "${inst}" --project "${project}" --env DEVLXD_BEARER_TOKEN="${token}" -- devlxd-client query GET /1.0 | jq --exit-status '.auth == "trusted"'
+
+    # Ensure DevLXD authentication does not accept LXD bearer token.
+    # Set same permissions to the LXD bearer token as to DevLXD bearer token
+    # and ensure it is not allowed to do any operation.
+    lxc auth identity create bearer/test-with-devlxd --group "${authGroup}"
+    lxdToken=$(lxc auth identity token issue bearer/test-with-devlxd --quiet)
+
+    # First, query LXD API directly to ensure the token is trusted by LXD API.
+    lxdAddr="$(lxc config get core.https_address)"
```
simondeziel: The test LXD's address should be already in `${LXD_ADDR}`.

client/util.go
```diff
@@ -23,13 +24,51 @@ import (
...
+
+			// Verify cert is valid.
+			now := time.Now()
+			if now.Before(serverCert.NotBefore) {
+				return fmt.Errorf("Server certificate is not valid at %q (valid after %q)", now.Format(time.RFC3339), serverCert.NotBefore.Format(time.RFC3339))
+			}
+
+			if now.After(serverCert.NotAfter) {
+				return fmt.Errorf("Server certificate is not valid at %q (expired at %q)", now.Format(time.RFC3339), serverCert.NotAfter.Format(time.RFC3339))
+			}
```
simondeziel, line 60: This could reuse [`certificateInDate`](https://github.com/canonical/lxd/blob/main/lxd/util/http.go#L160-L171)

tomponline:
@markylaing are you happy with this one?

tomponline:
@MusicDin please can you rebase and address @simondeziel and then if @markylaing is happy we are good to merge. Ta

client/util.go
MusicDin: certificateInDate is in internal package and is private func. We can make it shared, but this seems quite straight forward using std lib to check validity?

test/suites/devlxd.sh
MusicDin: Thanks

test/suites/auth.sh
MusicDin: It was just to unset the bearer token for this specific command, but not globally. Fixed.

test/lint/bin-size.sh
MusicDin: 😢

test/suites/auth.sh
MusicDin: Yeah, the reason why it uses `cat` is because there must be another double quote inside (empty tls_certificate). Single quotes won't work either as I need to populate bearer identity id. 

So those two lines need to resolve as intended:
```
id: ${bearer_identity_id}
tls_certificate: ""
```

MusicDin: Nvm, I could just escape them 🤦

markylaing (approved):
Great work @MusicDin

client/util.go
```diff
@@ -23,13 +24,51 @@ import (
...
+
+			// Verify fingerprint.
+			fingerprint := shared.CertFingerprint(serverCert)
+			if !strings.EqualFold(fingerprint, serverCertFingerprint) {
```
simondeziel, line 64: In [`CheckMutualTLS`](https://github.com/canonical/lxd/blob/main/lxd/util/http.go#L223-L238) we use `subtle.ConstantTimeCompare` to compare the raw cert. Should we do something similar here?

Also, why do the case folding?

simondeziel: @MusicDin sorry for not providing this comment earlier, seems like I forgot to actually save the comment :/

tomponline: > Should we do something similar here?

Yes we should

MusicDin: No problem at all, thanks for bringing that up.

I might be missing something, but we are not comparing certificates here. We are comparing server certificate fingerprint extracted from bearer token against the fingerprint calculated from the certificate reported by the LXD server.

I've used `EqualFold` because server fingerprint is a hex value and case should not matter. I can change the comparison to strict case matching, and just compare strings directly, but I think we should keep using equal fold.

For example, those two are the same fingerprints:
```
7dc4ebe08b0aa19349594fa22bff8677a129a2d352dcf933c94deadc37e7bfbe
7DC4EBE08B0AA19349594FA22BFF8677A129A2D352DCF933C94DEADC37E7BFBE
```

simondeziel: Using constant-time comparison would avoid leaking bits of information to an eventual attacker doing a bruteforce attack. This is explained better than I could do here: https://www.chosenplaintext.ca/articles/beginners-guide-constant-time-cryptography.html

The `tl;dr` version would be that regular string comparison will return earlier if the 2 strings diverge early on and will return later if they share a longer prefix. With made up strings and numbers:

* "foobarbaz" == "abcedfghi": returns after 1ns
* "foobarbaz" == "foodbcdef": returns after 3ns

The attacker guessing the right hand side knows that in the second attempt, he was onto something but not fully there yet.

MusicDin: If we just want the constant time compare I can change that. I see your point, but I still don't see the benefit of doing that on the client side, although it does not hurt to have it.

https://github.com/canonical/lxd/pull/17655

PR #17251, opened 2025-12-12
Improvements to `network_ovn` test

simondeziel:


test/suites/network_ovn.sh
```diff
@@ -45,6 +45,7 @@ test_network_ovn() {
 
   # Function to assert that the associative array matches what is in the northbound database.
   assert_row_count() {
+    local count table_name
```
simondeziel, line 48: This bit avoids leaking `count` to the calling function which is required for another PR I'm working on.

PR #17259, opened 2025-12-15
Rename database role to database-voter and improve cluster role validation

kadinsayani:
## Summary

Preliminary work to the addition of the `control-plane` cluster member role to improve role management and validation. Renames the `database` role to `database-voter` and improves cluster role validation by clearly separating automatic (system-managed) and manual (user-assignable) roles, and introducing additional cluster role validation for PUT/PATCH /1.0/cluster/members/{name}.

## Checklist

### DIFF
--- a/doc/howto/cluster_manage.md
+++ b/doc/howto/cluster_manage.md
@@ -5,7 +5,7 @@ discourse: lxc:[Cluster&#32;member&#32;evacuation](11330)
 (cluster-manage)=
 # How to manage a cluster
 
-After your cluster is formed, use [`lxc cluster list`](lxc_cluster_list.md) to see a list of its members and their status:
+After your cluster is formed, use [`lxc cluster list`](lxc_cluster_list.md) to see a list of its members and their status. Example output:
 
 ```{terminal}
 :input: lxc cluster list


doc/api-extensions.md
```diff
@@ -2884,3 +2884,21 @@ Before this was only partially possible for some of the drivers (e.g. by using `
...
+
+Introduces granular automatic cluster member roles and stricter validation when updating cluster member configuration.
+
+The single `database` role has been replaced with `database-voter`, allowing for three granular roles:
+
+* `database-voter` - Member participates in Raft consensus as a voter
+* `database-standby` - Member is a standby database node ready to be promoted to voter
+* `database-leader` - Member is the current Raft leader (in addition to `database-voter`)
+
+When updating a cluster member (via `PUT /1.0/cluster/{name}` or `lxc cluster edit`), automatic roles (`database-voter`, `database-standby`, `database-leader`) must now be explicitly included in the request if the member currently has them. Previously, omitting these roles would silently preserve them; now it will return an error.
```
tomponline: We didn't discuss this right? So I was surprised to see this included, I don't think we want this behaviour.
The automatic roles are just that, automatic, so the end user should not be forced to specify them, indeed they may change between reading them and setting them.
Please can it be removed.

Also no need for an API change here.

shared/api/cluster.go
```diff
@@ -162,7 +162,7 @@ type ClusterMember struct {
 	// Example: https://10.0.0.1:8443
 	URL string `json:"url" yaml:"url"`
 
-	// Whether the cluster member is a database server
+	// Whether the cluster member is a database server that participates in Raft consensus
```
tomponline: So are database-standby members not included in this?

lxd/db/node.go
```diff
@@ -43,32 +43,50 @@ const ClusterRoleEventHub = ClusterRole("event-hub")
 // ClusterRoleOVNChassis represents a cluster member who operates as an OVN chassis.
 const ClusterRoleOVNChassis = ClusterRole("ovn-chassis")
 
-// ClusterRoles maps role ids into human-readable names.
+// AutomaticClusterRoles contains roles that are automatically assigned by LXD.
+// These roles are managed internally and cannot be manually added or removed by users.
+var AutomaticClusterRoles = map[int]ClusterRole{
```
tomponline: Suggest storing these in a map keyed on type/class/family of role type (manual and automatic) rather than 2 different variables as it allows iterating the map for all role type/class/families rather than having to have duplicate loops when searching for a particular role name.

E.g.

```go
type ClusterRoleClass int
var ClusterRoles map[ClusterRoleClass]map[int]ClusterRole
```

shared/api/cluster.go
kadinsayani: They are, I'll reword this.

doc/api-extensions.md
kadinsayani: Correct, this wasn't discussed. I was honestly a bit unsure of this change as well, so thank you for clarifying your expectations for this behavior. I'll remove this change and the extension.

tomponline: Thanks! If you're not sure of something in the future please raise it via MM or via a comment in the PR itself drawing attention to it, so the emphasis is not on the reviewer to find it. It would be much appreciated by myself :)

kadinsayani: Will do :)

lxd/db/node.go
kadinsayani: Thanks for your suggestion, I like that approach.

kadinsayani:
> Thanks for the extra validation improvements.
> 
> I don't think we should require end users to have to supply matching automatic roles for the current member in their update requests.
> 
> Its not part of the database record they are updating and is added at view time only.

Thanks for your review. I've removed that requirement and addressed all comments.

test/suites/clustering.sh
```diff
@@ -4915,6 +4922,139 @@ test_clustering_events() {
...
+  # Spawn a fifth node.
+  setup_clustering_netns 5
+  LXD_FIVE_DIR=$(mktemp -d -p "${TEST_DIR}" XXX)
+  ns5="${prefix}5"
+  spawn_lxd_and_join_cluster "${ns5}" "${bridge}" "${cert}" 5 1 "${LXD_FIVE_DIR}" "${LXD_ONE_DIR}"
+
+  # Configure cluster with max_voters=3 and max_standby=1
+  LXD_DIR="${LXD_ONE_DIR}" lxc config set cluster.max_voters=3
+  LXD_DIR="${LXD_ONE_DIR}" lxc config set cluster.max_standby=1
+  LXD_DIR="${LXD_ONE_DIR}" lxc config set cluster.offline_threshold 11
```
simondeziel: Could this be a single `set` dealing with all 3 values?

test/suites/clustering.sh
```diff
@@ -4915,6 +4922,139 @@ test_clustering_events() {
...
+
+  sleep 12 # Wait a bit for cluster to stabilize.
+
+  lxc cluster ls
+
+  # Find a member with database-voter role.
+  voter_member=$(LXD_DIR="${LXD_ONE_DIR}" lxc cluster list -f json | jq -r '.[] | select(.roles[] | contains("database-voter")) | .server_name' | head -1)
```
simondeziel: Could you save the `LXD_DIR="${LXD_ONE_DIR}" lxc cluster list -f json` bit in a variable you then check multiple times with `jq --exit-status`?

Is there any way to get rid of the `head -1` bit? It would simplify those changes/checks.

test/suites/clustering.sh
```diff
@@ -4915,6 +4922,139 @@ test_clustering_events() {
...
+  # Find a member without database-leader role (to test adding it).
+  non_leader_member=$(LXD_DIR="${LXD_ONE_DIR}" lxc cluster list -f json | jq -r '.[] | select([.roles[] | contains("database-leader")] | any | not) | .server_name' | head -1)
+  echo "Found non-leader member: ${non_leader_member}"
+
+  # Find a member with database-leader role (to test removing it).
+  leader_member=$(LXD_DIR="${LXD_ONE_DIR}" lxc cluster list -f json | jq -r '.[] | select(.roles[] | contains("database-leader")) | .server_name' | head -1)
+  echo "Found database-leader member: ${leader_member}"
+
+  echo "==> Reject adding automatic role 'database-voter'"
+  [ "$(CLIENT_DEBUG="" SHELL_TRACING="" LXD_DIR="${LXD_ONE_DIR}" lxc cluster role add "${non_voter_member}" database-voter 2>&1)" = "Error: The automatically assigned \"database-voter\" role cannot be added manually" ]
```
simondeziel: That's a case for single quoting the error message that then includes double quotes. Would be more readable IMHO.

test/suites/clustering.sh
```diff
@@ -4915,6 +4922,139 @@ test_clustering_events() {
...
+  rm -f "${LXD_ONE_DIR}/unix.socket"
+
+  teardown_clustering_netns
+  teardown_clustering_bridge
+
+  kill_lxd "${LXD_ONE_DIR}"
+  kill_lxd "${LXD_TWO_DIR}"
+  kill_lxd "${LXD_THREE_DIR}"
+  kill_lxd "${LXD_FOUR_DIR}"
+  kill_lxd "${LXD_FIVE_DIR}"
```
simondeziel, line 5039: Why kill in reverse order when compared to the shutdown?

kadinsayani: Thanks for catching that, I used copilot to scaffold the spawning and cleanup steps.

test/suites/clustering.sh
kadinsayani: Yeah that would be much cleaner, thanks for suggesting!

kadinsayani: The `first` jq filter should work nicely.

test/suites/clustering.sh
simondeziel: The order is unchanged.

test/suites/clustering.sh
```diff
@@ -4915,6 +4922,140 @@ test_clustering_events() {
...
+
+  sleep 12 # Wait a bit for cluster to stabilize.
+
+  lxc cluster ls
+
+  # Get cluster list once and reuse it for all queries.
+  cluster_list=$(LXD_DIR="${LXD_ONE_DIR}" lxc cluster list -f json)
+
+  # Find a member with database-voter role.
+  voter_member=$(echo "${cluster_list}" | jq --exit-status -r '[.[] | select(.roles[] | contains("database-voter")) | .server_name] | first')
```
simondeziel: `s/-r/--raw-output/` please.

simondeziel: Also, with `bash` we can do:

```
voter_member="$(jq --exit-status -r '[.[] | select(.roles[] | contains("database-voter")) | .server_name] | first' <<< "${cluster_list}")"
```

kadinsayani: Cool, thanks!

test/suites/clustering.sh
kadinsayani: Apologies, changed now.

simondeziel (dismissed):
LGTM, let's see if it's green at the end of the pipeline ;)
Thanks!

kadinsayani:
> LGTM, let's see if it's green at the end of the pipeline ;) Thanks!

Thanks for your review :)

simondeziel (dismissed):

lxd/db/node.go
```diff
@@ -144,12 +144,15 @@ func (n NodeInfo) ToAPI(ctx context.Context, tx *ClusterTx, args NodeInfoArgs) (
...
-	if args.LeaderAddress == n.Address {
+	isLeader := args.LeaderAddress == n.Address
+	if isLeader {
 		result.Roles = append(result.Roles, string(ClusterRoleDatabaseLeader))
 		result.Database = true
 	}
 
-	if raftNode != nil && raftNode.Role == RaftVoter {
+	// Add database-voter role only for non-leader voters.
+	// Leaders are implicitly voters, so we don't show database-voter for them.
```
simondeziel, line 153: I like that simplification!

simondeziel (approved):

lxd/db/node.go
kadinsayani: Likewise! Great suggestion by @tomponline during our last 1:1 :)

test/suites/clustering.sh
```diff
@@ -245,7 +245,7 @@ test_clustering_membership() {
   LXD_DIR="${LXD_THREE_DIR}" lxc cluster list
   LXD_DIR="${LXD_TWO_DIR}" lxc cluster show node1 | grep -xF -- '- database-leader'
   [ "$(LXD_DIR="${LXD_THREE_DIR}" lxc cluster list | grep -Fc "database-standby")" = "2" ]
-  [ "$(LXD_DIR="${LXD_FIVE_DIR}" lxc cluster list | grep -Fc "database ")" = "3" ]
+  [ "$(LXD_DIR="${LXD_FIVE_DIR}" lxc cluster list | grep -Fc "database-voter")" = "2" ]
```
simondeziel: I've been thinking that `grep -wFc` would be nice as `-w` makes sure it doesn't match on partial role names. Granted this would have made even more sense before to avoid needing the trailing ` ` in the pattern.

Who knows, maybe we'll get a `database-voter-backup` or `database-voter-denied` role in the future?

The other `grep` patterns use `-x` so should be safe from this.

kadinsayani:
Marking as draft until tests are passing again.

lxd/db/node.go
```diff
@@ -43,32 +54,41 @@ const ClusterRoleEventHub = ClusterRole("event-hub")
...
+// ClusterRoleClasses provides reverse lookup from role name to role class.
+// Populated at init time.
+var ClusterRoleClasses map[ClusterRole]ClusterRoleClass
 
-// ClusterRoleSet provides O(1) membership testing for valid external cluster roles.
-// Automatically populated at init time.
-var ClusterRoleSet map[ClusterRole]bool
+// ManualClusterRoleIDs provides reverse lookup from role name to ID for manual roles.
+// Populated at init time.
+var ManualClusterRoleIDs map[ClusterRole]int
```
tomponline: You shouldn't need this any more, seems like a de-normalisation step too far (especially for an infrequently used API operation like updating cluster roles).
You now have a map of roles by class, so if validating role against all classes you can iterate through ClusterRoles for each class and then each role in a class, or you can jump directly to a particular class if needed.

lxd/db/node.go
```diff
@@ -629,20 +652,18 @@ func (c *ClusterTx) UpdateNodeConfig(ctx context.Context, id int64, config map[s
...
+		if ClusterRoleClasses[role] == ClusterRoleClassAutomatic {
 			continue
 		}
 
-		roleID, valid := ClusterRoleIDs[role]
-		if !valid {
-			return fmt.Errorf("Invalid cluster role %q", role)
-		}
-
+		roleID := ManualClusterRoleIDs[role]
```
tomponline: Can we use the `maps.Values` function to check if the role is present in the `ClusterRoles[ClusterRoleClassManual]` map and avoid the need for having the permanent reverse map in memory?

lxd/db/node.go
```diff
@@ -629,20 +652,18 @@ func (c *ClusterTx) UpdateNodeConfig(ctx context.Context, id int64, config map[s
...
+// Automatic roles are managed by Raft and filtered out.
+// Callers are expected to validate roles before calling this function.
 func (c *ClusterTx) UpdateNodeRoles(id int64, roles []ClusterRole) error {
-	// Translate role names to ids
+	// Translate role names to IDs, filtering out automatic roles.
 	roleIDs := []int{}
 	for _, role := range roles {
-		// Skip internal-only roles.
-		if role == ClusterRoleDatabase || role == ClusterRoleDatabaseStandBy || role == ClusterRoleDatabaseLeader {
+		if ClusterRoleClasses[role] == ClusterRoleClassAutomatic {
```
tomponline: We dont need this check as if the next check fails to find an entry it must be automatic or invalid.

lxd/db/node.go
```diff
@@ -37,38 +48,47 @@ const ClusterRoleDatabaseStandBy = ClusterRole("database-standby")
...
-// configuration which acts as single source of truth for it. This map should
-// only contain LXD-specific cluster roles.
-var ClusterRoles = map[int]ClusterRole{
-	1: ClusterRoleEventHub,
-	2: ClusterRoleOVNChassis,
+// ClusterRoles maps role classes to their respective role definitions.
+// Automatic roles are managed by LXD and cannot be manually modified.
+// Manual roles can be assigned and removed by users via the API.
+var ClusterRoles = map[ClusterRoleClass]map[int]ClusterRole{
+	ClusterRoleClassAutomatic: {
```
tomponline, line 62: Can we add a comment to this section mentioning these indexes are not IDs, unlike the `ClusterRoleClassManual` ones?

test/suites/clustering.sh
kadinsayani: Thanks, I've added the -w flag to those greps.

lxd/db/node.go
kadinsayani: I've removed all reverse lookup maps and updated all references to use `maps.Values` or range based iteration where possible.

lxd/db/node.go
kadinsayani: Done.

lxd/db/node.go
kadinsayani: Thanks for suggesting, much simpler now and more space efficient.

test/suites/clustering.sh
```diff
@@ -3101,7 +3108,7 @@ test_clustering_rebalance() {
   LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node2 | grep -xF "status: Offline"
   LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node2 | grep -xF "database: false"
   LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node4 | grep -xF "status: Online"
-  LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node4 | grep -F -- "- database"
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node4 | grep -F -- "- database-voter"
```
simondeziel: This one is (and was) missing an `-x`.

kadinsayani:
Tests passing :white_check_mark:

simondeziel (dismissed):

test/suites/clustering.sh
kadinsayani: Fixed.

test/suites/clustering.sh
```diff
@@ -4922,6 +4922,128 @@ test_clustering_events() {
...
+  prefix="lxd$$"
+  bridge="${prefix}"
+
+  setup_clustering_netns 1
+  LXD_ONE_DIR=$(mktemp -d -p "${TEST_DIR}" XXX)
+  ns1="${prefix}1"
+  spawn_lxd_and_bootstrap_cluster "${ns1}" "${bridge}" "${LXD_ONE_DIR}"
+
+  # Add a newline at the end of each line. YAML has weird rules.
+  cert=$(sed ':a;N;$!ba;s/\n/\n\n/g' "${LXD_ONE_DIR}/cluster.crt")
```
tomponline, line 4938: For the future this feels like a good example for a function, there are now 43 occurrences of this line in the test suite.

cc @simondeziel

test/suites/clustering.sh
```diff
@@ -4922,6 +4922,128 @@ test_clustering_events() {
...
+  ns4="${prefix}4"
+  spawn_lxd_and_join_cluster "${ns4}" "${bridge}" "${cert}" 4 1 "${LXD_FOUR_DIR}" "${LXD_ONE_DIR}"
+
+  # Spawn a fifth node.
+  setup_clustering_netns 5
+  LXD_FIVE_DIR=$(mktemp -d -p "${TEST_DIR}" XXX)
+  ns5="${prefix}5"
+  spawn_lxd_and_join_cluster "${ns5}" "${bridge}" "${cert}" 5 1 "${LXD_FIVE_DIR}" "${LXD_ONE_DIR}"
+
+  # Configure cluster with max_voters=3 and max_standby=1
```
tomponline, line 4964: The comment describes what we can see is plainly happening in the next line, but it doesn't say why that is significant for this test. Comments are more useful when they explain the **why** rather than the **what** (unless ofcourse if the **what** is non-obvious or complicated, in which case explaining both are beneficial).

tomponline (approved):
Thanks!

test/suites/clustering.sh
kadinsayani: Ack, I'll improve this comment to go deeper into why we want node5 to be a spare in the next PR. It only becomes important in the new tests for the `control-plane` role. I've added checks that verify node5 stays as a spare when it's missing the `control-plane` role, and only becomes eligible for promotion after assigning the `control-plane` role to it.

test/suites/clustering.sh
simondeziel: https://github.com/canonical/lxd/pull/17287 should make that simpler.

kadinsayani: Nice, thanks @simondeziel!

PR #17262, opened 2025-12-16
Fix ZFS origin snapshot not always removed recursively

tugbataluy:
## PR Overview
This pull request fixes a bug in LXD's ZFS storage driver where origin snapshots were not always removed recursively when deleting instances. This issue occurred specifically when the original instance was deleted before its copies, leaving orphaned snapshots in the ZFS pool.

Fixes: https://github.com/canonical/lxd/issues/16448

## Problem Description

As mentioned in [#16448](https://github.com/canonical/lxd/issues/16448), when using ZFS storage, the following scenario would leave orphaned snapshots:

1. Create an instance (`u1`) 
2. Create two copies (`u2`, `u3`) - this creates two `@copy-*` snapshots on the original
3. Delete the original instance (`u1`) - moves to `/deleted` with snapshots intact
4. Delete one copy (`u3`) - should remove its corresponding snapshot from the deleted volume
5. Delete the last copy (`u2`) - should remove the remaining snapshot and clean up

**Expected behavior**: After step 4, only one snapshot should remain.

**Actual behavior**: The deleted volume remained both snapshots instead of cleaning them up recursively. After step 5 there aren't any snapshots remain.

## Root Cause

The issue was in the `deleteDatasetRecursive()` function in `lxd/storage/drivers/driver_zfs_utils.go`. When processing origins that were `@copy-*` or `@deleted-*` snapshots, the logic incorrectly stripped the snapshot portion (`@...`) from the dataset name before attempting deletion, causing the cleanup to target the parent volume instead of the specific snapshot.

## Checklist

### DIFF
--- a/test/suites/storage_driver_zfs.sh
+++ b/test/suites/storage_driver_zfs.sh
@@ -14,6 +14,7 @@ test_storage_driver_zfs() {
   do_zfs_cross_pool_copy
   do_zfs_delegate
   do_zfs_rebase
+  do_recursive_copy_snapshot_cleanup
 }
 
 do_zfs_delegate() {
@@ -388,4 +389,91 @@ do_storage_driver_zfs() {
 
   # shellcheck disable=SC2031
   kill_lxd "${LXD_STORAGE_DIR}"
+}
+
+do_recursive_copy_snapshot_cleanup() {
+  echo "Test recursive copy snapshot cleanup."
+  local storage_pool
+  storage_pool="lxdtest-$(basename "${LXD_DIR}")"
+
+  echo "Create the first container."
+  lxc init testimage t1
+
+  echo "Make two copies."
+  lxc copy t1 t2
+  lxc copy t1 t3
+
+  echo "Verify two copy snapshots exist."
+  [ "$(zfs list -t snapshot -H -o name "${storage_pool}/containers/t1" | grep -cF "@copy-")" -eq 2 ]
+
+  echo "Delete t3, should delete one copy snapshot."
+  lxc delete t3
+
+  echo "Verify one copy snapshot remains."
+  [ "$(zfs list -t snapshot -H -o name "${storage_pool}/containers/t1" | grep -cF "@copy-")" -eq 1 ]
+
+  echo "Delete t2, should delete the remaining copy snapshot."
+  lxc delete t2
+
+  echo "Verify no snapshots remain, should output \"no datasets available\"."
+  [ "$(zfs list -t snapshot "${storage_pool}/containers/t1" 2>&1)" = "no datasets available" ]
+
+  echo "Create two new copies."
+  lxc copy t1 t4
+  lxc copy t1 t5
+
+  echo "Verify two new copy snapshots exist."
+  [ "$(zfs list -t snapshot -H -o name "${storage_pool}/containers/t1" | grep -cF "@copy-")" -eq 2 ]
+
+  echo "Delete the original container t1, should move to deleted pool with snapshots."
+  lxc delete t1
+
+  echo "Verify container moved to deleted pool with both copy snapshots."
+  [ "$(zfs list -rt snapshot "${storage_pool}/deleted/containers" | grep -cF "@copy-")" -eq 2 ]
+
+  echo "Delete t5, should delete its snapshot from deleted container."
+  lxc delete t5
+
+  echo "Verify one snapshot remains in the deleted container."
+  [ "$(zfs list -rt snapshot "${storage_pool}/deleted/containers" | grep -cF "@copy-")" -eq 1 ]
+
+  echo "Delete t4, should delete remaining snapshot, leaving no snapshots."
+  lxc delete t4
+
+  echo "Verify no snapshots remain."
+  [ "$(zfs list -rt snapshot "${storage_pool}/deleted/containers" 2>&1)" = "no datasets available" ]
+
+  echo "Test chain copy snapshot cleanup."
+
+  echo "Create base container."
+  lxc init testimage base
+
+  echo "Create chain of copies."
+  lxc copy base chain1
+  lxc copy chain1 chain2
+  lxc copy chain2 chain3
+
+  echo "Verify base container has one copy snapshot."
+  [ "$(zfs list -t snapshot -H -o name "${storage_pool}/containers/base" | grep -cF "@copy-")" -eq 1 ]
+
+  echo "Verify chain1 has one copy snapshot."
+  [ "$(zfs list -t snapshot -H -o name "${storage_pool}/containers/chain1" | grep -cF "@copy-")" -eq 1 ]
+
+  echo "Verify chain2 has one copy snapshot."
+  [ "$(zfs list -t snapshot -H -o name "${storage_pool}/containers/chain2" | grep -cF "@copy-")" -eq 1 ]
+
+  echo "Verify chain3 has no copy snapshots."
+  [ "$(zfs list -t snapshot -H -o name "${storage_pool}/containers/chain3" | grep -cF "@copy-")" -eq 0 ]
+
+  echo "Delete base, chain1, and chain2 containers."
+  lxc delete base chain1 chain2
+
+  echo "Verify three copy snapshots exist in deleted pool."
+  [ "$(zfs list -rt snapshot "${storage_pool}/deleted/containers" | grep -cF "@copy-")" -eq 3 ]
+
+  echo "Delete the remaining chain3 container."
+  lxc delete chain3
+
+  echo "Verify no snapshots remain in deleted pool."
+  [ "$(zfs list -rt snapshot "${storage_pool}/deleted/containers" 2>&1)" = "no datasets available" ]
 }


test/suites/storage_driver_zfs.sh
```diff
@@ -374,6 +374,70 @@ do_storage_driver_zfs() {
...
+  # Verify one snapshot remains in the deleted container.
+  deleted_container=$(zfs list -rt snapshot test-pool/deleted/containers)
+  copy_snap_count=$(echo "$deleted_container" | grep -c "@copy-")
+  [ "$copy_snap_count" -eq 1 ]
+
+  # Delete t4, should delete remaining snapshot, leaving no snapshots.
+  lxc delete t4
+
+  # Verify no snapshots remain.
+  zfs list -rt snapshot test-pool/deleted/containers 2>&1 | grep -F "no datasets available"
```
simondeziel:
```suggestion
  [ "$(zfs list -rt snapshot default/deleted/containers 2>&1)" = "no datasets available" ]
```

test/suites/storage_driver_zfs.sh
simondeziel: I believe the new test section you added is using `volume.zfs.block_mode=true`, possibly unintentionally. Using `block_mode` is inherently much slower. This might be the reason why those additional tests take ~11s.

test/suites/storage_driver_zfs.sh
```diff
@@ -374,6 +374,70 @@ do_storage_driver_zfs() {
...
+  lxc init testimage t1 --storage test-pool
+
+  # Make two copies.
+  lxc copy t1 t2
+  lxc copy t1 t3
+
+  # Verify two copy snapshots exist.
+  copy_snaps=$(zfs list -t snapshot -H -o name test-pool/containers/t1)
+  copy_snap_count=$(echo "$copy_snaps" | grep -c "@copy-")
+  [ "$copy_snap_count" -eq 2 ]
```
simondeziel: This feels more direct.
```suggestion
  [ "$(zfs list -t snapshot -H -o name test-pool/containers/t1 | grep -cF "@copy-")" -eq 2 ]
```

lxd/storage/drivers/driver_zfs_utils.go
```diff
@@ -123,33 +123,78 @@ func (d *zfs) deleteDatasetRecursive(dataset string) error {
 
 	// Check if the origin can now be deleted.
 	if origin != "" && origin != "-" {
+		var targetToDelete string
+
+		// Check deleted directory first because these volumes need special handling:
+		// They're already marked for deletion.
+		// We want to clean up the parent volume when it becomes empty.
+		// This handles both instances and custom volumes.
 		if strings.HasPrefix(origin, d.config["zfs.pool_name"]+"/deleted") {
```
simondeziel: As an extra precaution, I'd add the trailing `/` after so: `/deleted/`.

This mirrors how in Ceph we look for `zombie_` and not just `zombie`

lxd/storage/drivers/driver_zfs_utils.go
```diff
@@ -123,33 +123,78 @@ func (d *zfs) deleteDatasetRecursive(dataset string) error {
...
 				return err
 			}
 
 			if len(clones) == 0 {
-				// Delete the origin.
-				err = d.deleteDatasetRecursive(dataset)
-				if err != nil {
-					return err
+				// If we're deleting a snapshot from a deleted volume, check if this is the last snapshot. If so, we should also try to delete the parent volume.
+				if strings.Contains(targetToDelete, "@") && strings.HasPrefix(targetToDelete, d.config["zfs.pool_name"]+"/deleted") {
```
simondeziel:
```suggestion
				if strings.Contains(targetToDelete, "@") && strings.HasPrefix(targetToDelete, d.config["zfs.pool_name"]+"/deleted/") {
```

test/suites/storage_driver_zfs.sh
simondeziel: With that simple change of avoiding `block_mode`, the additional tests now take ~8s instead of ~11s.

That's still quite a bit higher than I initially anticipated as only 5 testimage instances are created after all. Looking at it more closely, I noticed that it only takes 2.5s but since the new tests were added in `do_storage_driver_zfs()`, this is repeated 3 times, once for each of `ext4`, `xfs` and `btrfs`.

So I think you'd be better off with creating a new subtest, say `do_zfs_recursive_snap()` or similar and have it called only once from `test_storage_driver_zfs()`.

lxd/storage/drivers/driver_zfs_utils.go
```diff
@@ -123,33 +123,78 @@ func (d *zfs) deleteDatasetRecursive(dataset string) error {
...
+		var targetToDelete string
+
+		// Check deleted directory first because these volumes need special handling:
+		// They're already marked for deletion.
+		// We want to clean up the parent volume when it becomes empty.
+		// This handles both instances and custom volumes.
+		if strings.HasPrefix(origin, d.config["zfs.pool_name"]+"/deleted/") {
+			// Origin is in deleted directory.
+			if strings.Contains(origin, "@deleted-") || strings.Contains(origin, "@copy-") {
+				// This is a deleted or copy snapshot in deleted directory, delete the snapshotitself (don't strip the @part).
```
tomponline: typo `snapshotitself`

tomponline: This comment and the comment below are explaining the **what** of the decision, e.g.

> This is a deleted or copy snapshot in deleted directory, delete the snapshot itself

> This is a regular snapshot in deleted directory, delete the parent volume (strip the @part).

But they don't explain the **why**. This is, in my view, the most important part.

Why is it that we delete the whole parent volume (dataset) when the origin is a regular snapshot in a deleted dataset, and only delete the origin snapshot itself if the snapshot is deleted or copy in a deleted dataset?

Also we should say "dataset" rather than "directory" when referring to the ZFS hierarchy.

lxd/storage/drivers/driver_zfs_utils.go
```diff
@@ -123,33 +123,78 @@ func (d *zfs) deleteDatasetRecursive(dataset string) error {
...
-			dataset, _, _ = strings.Cut(origin, "@")
+		var targetToDelete string
+
+		// Check deleted directory first because these volumes need special handling:
+		// They're already marked for deletion.
+		// We want to clean up the parent volume when it becomes empty.
+		// This handles both instances and custom volumes.
+		if strings.HasPrefix(origin, d.config["zfs.pool_name"]+"/deleted/") {
+			// Origin is in deleted directory.
+			if strings.Contains(origin, "@deleted-") || strings.Contains(origin, "@copy-") {
```
tomponline: This code `if strings.Contains(origin, "@deleted-") || strings.Contains(origin, "@copy-") ` occurs twice in this function now.

Instead of duplicating it, can we define its meaning as a bool variable once above here and then use it multiple places, e.g.

```go
if origin != "" && origin != "-" {
     foo := strings.Contains(origin, "@deleted-") || strings.Contains(origin, "@copy-")
     ..

     if foo {
     ...
     else if foo ...
}
```

lxd/storage/drivers/driver_zfs_utils.go
```diff
@@ -123,33 +123,78 @@ func (d *zfs) deleteDatasetRecursive(dataset string) error {
...
 		} else if strings.Contains(origin, "@deleted-") || strings.Contains(origin, "@copy-") {
-			// Handle deleted snapshots.
-			dataset = origin
+			// Origin is a deleted or copy snapshot but not in deleted directory, delete the snapshot itself (don't strip the @part).
+			targetToDelete = origin
 		} else {
-			// Origin is still active.
-			dataset = ""
+			// Origin is still active (not in deleted, not a special snapshot).
+			targetToDelete = ""
```
tomponline: We dont need this part right? Because `var targetToDelete string` will default to empty.

Then below:

```go
// If targetToDelete is non-empty it means the origin is inactive and should be deleted.
if targetToDelete != "" {
}

File: lxd/storage/drivers/driver_zfs_utils.go [hunk 8]
```diff
@@ -123,33 +123,78 @@ func (d *zfs) deleteDatasetRecursive(dataset string) error {
...
+		if strings.HasPrefix(origin, d.config["zfs.pool_name"]+"/deleted/") {
+			// Origin is in deleted directory.
+			if strings.Contains(origin, "@deleted-") || strings.Contains(origin, "@copy-") {
+				// This is a deleted or copy snapshot in deleted directory, delete the snapshotitself (don't strip the @part).
+				targetToDelete = origin
+			} else {
+				// This is a regular snapshot in deleted directory, delete the parent volume (strip the @part).
+				targetToDelete, _, _ = strings.Cut(origin, "@")
+			}
 		} else if strings.Contains(origin, "@deleted-") || strings.Contains(origin, "@copy-") {
```
tomponline: It looks to me like this if/else block can be simplified:

```go
var originToDelete string // Will be set non-empty if origin (or parent of) should be deleted.

if strings.Contains(origin, "@deleted-") || strings.Contains(origin, "@copy-") {
    originToDelete = origin // Origin is a temporary snapshot for this dataset, delete it.
} else if strings.HasPrefix(origin, d.config["zfs.pool_name"]+"/deleted/") {
    originToDelete, _, _ = strings.Cut(origin, "@") // This is a regular snapshot for a deleted origin dataset, delete the origin parent dataset.
}

if originToDelete == "" {
      return nil // Origin doesn't need to be deleted.
}

// Origin needs to be deleted...
```

lxd/storage/drivers/driver_zfs_utils.go
tomponline: looks like we can simplify the logic to avoid the duplication anyway, see https://github.com/canonical/lxd/pull/17262#discussion_r2635041336 so no need for this.

lxd/storage/drivers/driver_zfs_utils.go
```diff
@@ -123,33 +123,78 @@ func (d *zfs) deleteDatasetRecursive(dataset string) error {
...
+		}
+
+		// Check if the origin (or parent dataset) has any clones.
+		clones, err := d.getClones(originToDelete)
+		if err != nil {
+			return err
+		}
+
+		// Only proceed with deletion if there are no clones.
+		if len(clones) == 0 {
```
tomponline: we can do happy path here to avoid indention, e.g. `if len(clones) >0 { return nil }`

lxd/storage/drivers/driver_zfs_utils.go
```diff
@@ -123,33 +123,78 @@ func (d *zfs) deleteDatasetRecursive(dataset string) error {
...
+		var originToDelete string
+
+		// Determine if the origin is a temporary snapshot created.
+		IsOriginTemporarySnapshot := strings.Contains(origin, "@deleted-") || strings.Contains(origin, "@copy-")
+
+		// Determine deletino strategy based on origin type and location:
+		// - Temporary snapshots (deleted- or copy-) should be deleted themselves.
+		// - Regular snapshots in deleted/ dataset shuould cause parent dataset deletion (when dataset is in deleted/, it's marked for cleanup, so we clean up the parent dataset when its last snapshot is removed).
+
+		if IsOriginTemporarySnapshot {
```
tomponline: we can move the logic for `IsOriginTemporarySnapshot` back here now its not duplicated.

test/suites/storage_driver_zfs.sh
```diff
@@ -389,3 +390,59 @@ do_storage_driver_zfs() {
   # shellcheck disable=SC2031
   kill_lxd "${LXD_STORAGE_DIR}"
 }
+
+do_recursive_copy_snapshot_cleanup() {
+  # Test recursive copy snapshot cleanup.
+  # Create separate storage pool for the test.
+  lxc storage create test-pool zfs
```
simondeziel: Sorry to only pick this up now!

I just ran a test locally and you don't need a new pool for this subtest.

```
local storage_pool
storage_pool="lxdtest-$(basename "${LXD_DIR}")"
```

Should be all you need to interact with the existing pool. This will avoid the need to create and destroy a pool which is rather slow.

test/suites/storage_driver_zfs.sh
```diff
@@ -389,3 +390,56 @@ do_storage_driver_zfs() {
...
   kill_lxd "${LXD_STORAGE_DIR}"
 }
+
+do_recursive_copy_snapshot_cleanup() {
+  # Test recursive copy snapshot cleanup.
+  local storage_pool
+  storage_pool="lxdtest-$(basename "${LXD_DIR}")"
+
+  # Create the first container
+  lxc init testimage t1 --storage "${storage_pool}"
```
simondeziel: That ` --storage "${storage_pool}"` part is implicit.

simondeziel (approved):

test/suites/storage_driver_zfs.sh
```diff
@@ -389,3 +390,56 @@ do_storage_driver_zfs() {
   # shellcheck disable=SC2031
   kill_lxd "${LXD_STORAGE_DIR}"
 }
+
+do_recursive_copy_snapshot_cleanup() {
+  # Test recursive copy snapshot cleanup.
```
kadinsayani: For integration tests, it can be helpful to replace comments with `echo` calls. This makes debugging and interpreting test output much easier. For example:

```
  echo "Test recursive copy snapshot cleanup."
  local storage_pool
  storage_pool="lxdtest-$(basename "${LXD_DIR}")"

  echo "Create the first container"
  lxc init testimage t1 --storage "${storage_pool}"

  echo "Make two copies."
  lxc copy t1 t2
  lxc copy t1 t3

  echo "Verify two copy snapshots exist."
  [ "$(zfs list -t snapshot -H -o name "${storage_pool}/containers/t1" | grep -cF "@copy-")" -eq 2 ]

  echo "Delete t3, should delete one copy snapshot."
  lxc delete t3

...
```

lxd/storage/drivers/driver_zfs_utils.go
```diff
@@ -123,33 +123,74 @@ func (d *zfs) deleteDatasetRecursive(dataset string) error {
...
-		if dataset != "" {
-			// Get all clones.
-			clones, err := d.getClones(dataset)
+		// If there are clones, we can't delete the origin.
+		if len(clones) > 0 {
+			return nil
+		}
+
+		// We need to check if we should also delete the parent dataset.
+		if strings.Contains(originToDelete, "@") && strings.HasPrefix(originToDelete, d.config["zfs.pool_name"]+"/deleted/") {
```
tomponline, line 155: I wonder if this should be a loop that traverses up the tree?

As this code seems to fix the issue with the immediate parent, but is it possible for there to be a parent of a parent, etc?

tugbataluy: Can I create and test a non-immediate parent situation like this? And what should the expected behavior be: (When I repeat the same steps as before the fix, I get the same results).
```
# 1. Create storage pool and base container
lxc storage create z1 zfs
lxc init ubuntu:24.04 base --storage z1

# 2. Create a 3-level chain
lxc copy base chain1      # Creates: base@copy-A → chain1
lxc copy chain1 chain2    # Creates: chain1@copy-B → chain2  
lxc copy chain2 chain3    # Creates: chain2@copy-C → chain3

zfs list -t snapshot z1/containers/base
NAME                                                           USED  AVAIL  REFER  MOUNTPOINT
z1/containers/base@copy-616bdafc-177c-45e4-8222-09ae314b6f1d     0B      -   501M  -

zfs list -t snapshot z1/containers/chain1
NAME                                                             USED  AVAIL  REFER  MOUNTPOINT
z1/containers/chain1@copy-eb7aaefc-57de-439b-a26a-e6916ef1953f     0B      -   501M  -

zfs list -t snapshot z1/containers/chain2
NAME                                                             USED  AVAIL  REFER  MOUNTPOINT
z1/containers/chain2@copy-bb81764e-eca6-4fd1-a5a3-b4ed0cad813d     0B      -   501M  -



# Current zfs structure:
base (active)
└── @copy-A (snapshot)
    └── chain1 (clone, active)
        └── @copy-B (snapshot)
            └── chain2 (clone, active)
                └── @copy-C (snapshot)
                    └── chain3 (clone, active)
                    

# 3. Delete the original base (moves to deleted/ directory)
lxc delete base

zfs list -rt snapshot z1/deleted/containers
# Should show: deleted_base@copy-A exists
NAME                                                                                                   USED  AVAIL  REFER  MOUNTPOINT
z1/deleted/containers/a7a8a69e-e9fe-41e4-94d7-ba3c6d1cac8a@copy-616bdafc-177c-45e4-8222-09ae314b6f1d     0B      -   501M  -


zfs get origin z1/containers/chain1
# Should show: origin = deleted_base@copy-A
NAME                  PROPERTY  VALUE                                                                                                 SOURCE
z1/containers/chain1  origin    z1/deleted/containers/a7a8a69e-e9fe-41e4-94d7-ba3c6d1cac8a@copy-616bdafc-177c-45e4-8222-09ae314b6f1d  -


# 4. Delete the last container in the chain
lxc delete chain3

What should happen:

    chain3 dataset is deleted

    Should its origin snapshot chain2@copy-C be deleted (no more clones) or not? I think it should be deleted.

    Check: Does deleted_base@copy-A get deleted? NO - because chain1 and chain2 still exist

Current output:
zfs list -t snapshot z1/containers/chain2
no datasets available

Check: As far as I understand we should see deleted base. But should we also see the deleted chain3 here I'm not sure.
zfs list -t snapshot -r z1/deleted/containers
NAME                                                                                                   USED  AVAIL  REFER  MOUNTPOINT
z1/deleted/containers/a7a8a69e-e9fe-41e4-94d7-ba3c6d1cac8a@copy-616bdafc-177c-45e4-8222-09ae314b6f1d     0B      -   501M  -

# 5. Delete the middle container
lxc delete chain2

What should happen:

    chain2 dataset is deleted

    Its origin snapshot chain1@copy-B should be deleted (no more clones)

    Check: Does deleted_base@copy-A get deleted?

Current output:
zfs list -t snapshot z1/containers/chain1
no datasets available

Check: As far as I understand we should see deleted base.

zfs list -t snapshot -r z1/deleted/containers
NAME                                                                                                   USED  AVAIL  REFER  MOUNTPOINT
z1/deleted/containers/a7a8a69e-e9fe-41e4-94d7-ba3c6d1cac8a@copy-616bdafc-177c-45e4-8222-09ae314b6f1d     0B      -   501M  -

# 6. Delete the remaining container
lxc delete chain1
What should happen:

    chain1 dataset is deleted

    Its origin snapshot deleted_base@copy-A should now be deleted (no references left)

    The entire deleted base dataset should be removed

Check: No dataset should remain.
zfs list -t snapshot -r z1/deleted/containers
no datasets available

```

tomponline: After doing:

```
# 1. Create storage pool and base container
lxc storage create z1 zfs
lxc init ubuntu:24.04 base --storage z1

# 2. Create a 3-level chain
lxc copy base chain1      # Creates: base@copy-A → chain1
lxc copy chain1 chain2    # Creates: chain1@copy-B → chain2  
lxc copy chain2 chain3    # Creates: chain2@copy-C → chain3
```

What is left if you run `lxc delete base chain1 chain2`?

tomponline: Also, separate question, do we need to check with `getClones()` first for existing clones, or can we do a non-recursive remove that would fail if there were clones, and then we can detect that specific error and ignore it?

tugbataluy: zfs list -rt snapshot z1/deleted/containers
NAME                                                                                                   USED  AVAIL  REFER  MOUNTPOINT
z1/deleted/containers/4a3eb1f0-5585-4a2e-b58d-65a9e08dfca1@copy-a7df0ac8-b4d7-4fdc-bd9c-6f108fbe4a21     0B      -   501M  -
z1/deleted/containers/5a2f3aaa-305c-4b28-b1b0-a6f9e27b8fb4@copy-fe2ce98f-e527-4e11-bff4-275d9a7f495a     0B      -   501M  -
z1/deleted/containers/9c229868-ee92-4a64-bece-31cbef9c5c18@copy-7592a611-e77d-4022-9fab-cd78b1abe915     0B      -   501M  -

tomponline: cool thanks, and does doing `lxc delete chain3` remove all of those?

tugbataluy: Nope :( I need to review my logic

tomponline: ah ok, so that looks like the need for recursing back up the tree I mentioned.

tugbataluy: Yes, I guess so. Can I also add a tiny test case for  chain recursive removal? 
If you hadn't mentioned it, I wouldn't have realized that current behavior was causing the non-immediate parent removal to work incorrectly.

tomponline: > Can I also add a tiny test case for chain recursive removal?

yes please

PR #17263, opened 2025-12-16
Operation context follow up

markylaing:
Follow up to #17016 

## Checklist

### DIFF
--- a/lxd/instance_exec.go
+++ b/lxd/instance_exec.go
@@ -279,13 +279,6 @@ func (s *execWs) Do(ctx context.Context, op *operations.Operation) error {
 			cmdErr = nil
 		}
 
-		// If the exec operation was deleted, the exec websocket is closed and the caller is booted from the instance.
-		// In this case, make sure there is a non-zero exit code. The disconnection exit code is used here.
-		if ctx.Err() != nil && cmdResult == 0 {
-			cmdResult = 129
-			cmdErr = nil
-		}
-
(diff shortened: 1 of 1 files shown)


lxd-agent/operations.go
```diff
@@ -52,11 +52,7 @@ func operationDelete(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
-	err = op.Cancel()
-	if err != nil {
-		return response.BadRequest(err)
-	}
-
+	op.Cancel()
```
tomponline, line 55: We could check the status here and reject with bad request if already cancelled.
This would keep the existing API business logic in the hander.

What do you think?

markylaing: I wonder if we should reintroduce `MayCancel` and have handlers call that.

So `MayCancel` would be false in the API response if the operation is already cancelled or has completed. Then API handlers return a 400 if a request tries to cancel an operation where MayCancel is false

markylaing: ATM MayCancel is hardcoded to true

tomponline: `MayCancel` is looking likely to be due for removal given it should always be true.

Instead for this particular API endpoint we could just check the status and reject if already being cancelled?

tomponline: @markylaing ive merged this PR, but would be good to follow up to restore the API behaviour of preventing repeat cancellation of a cancelled operation.

PR #17266, opened 2025-12-16
Recover: Proactively ask for pool creation instead

roosterfish:
This PR finally makes use of the list of pools introduced with the new metadata format (backup config) stored for instance volumes. 
The pool of each custom volume attached to the instance is recorded there which allows notifying the user of `lxd recover` in case there are further pools which require recovery.
In addition we show the pool's current configuration which makes it straightforward to recover it manually using `lxc storage create`.

This replaces the option to enter additional pools for recovery which was also only working for standalone LXDs and not for clusters. 

**Hint for easier review**: To get an overview over the new flow in `lxd recover` checkout the last commit first which contains updated docs and demonstrates the new capabilities.

### DIFF
--- a/doc/howto/disaster_recovery.md
+++ b/doc/howto/disaster_recovery.md
@@ -10,8 +10,8 @@ discourse: lxc:[New&#32;disaster&#32;recovery&#32;tool](11296)
 
 LXD provides a tool for disaster recovery in case the {ref}`LXD database <database>` is corrupted or otherwise lost.
 
-The tool scans the storage pools for instances and imports the instances that it finds back into the database.
-You need to re-create the required entities that are missing (usually profiles, projects, and networks).
+The tool scans the storage pools for instances and imports the instances, custom volumes and buckets that it finds back into the database.
+You need to re-create the required entities that are missing (usually pools, profiles, projects, and networks).
 
 ```{important}
 This tool should be used for disaster recovery only.
@@ -25,16 +25,14 @@ The tool is available through the `lxd recover` command (note the `lxd` command 
 ## Recovery process
 
 When you run the tool, it scans all storage pools that still exist in the database, looking for missing volumes that can be recovered.
-You can also specify the details of any unknown storage pools (those that exist on disk but do not exist in the database), and the tool attempts to scan those too.
+Any unknown storage pools (those that exist on disk but do not exist in the database) which are discovered whilst scanning existing and unknown volumes
+are printed so they can be created manually using the `lxc storage create` command.
 
 After mounting the specified storage pools (if not already mounted), the tool scans them for unknown volumes that look like they are associated with LXD.
 LXD maintains a `backup.yaml` file in each instance's storage volume, which contains all necessary information to recover a given instance (including instance configuration, attached devices, storage volume, and pool configuration).
-This data can be used to rebuild the instance, storage volume, and storage pool database records.
+This data can be used to rebuild the instance, storage volume, attached custom volumes, and storage pool database records.
 Before recovering an instance, the tool performs some consistency checks to compare what is in the `backup.yaml` file with what is actually on disk (such as matching snapshots).
 If all checks out, the database records are re-created.
-
-If the storage pool database record also needs to be created, the tool uses the information from an instance's `backup.yaml` file as the basis of its configuration, rather than what the user provided during the discovery phase.
-However, if this information is not available, the tool falls back to restoring the pool's database record with what was provided by the user.
 
 The tool asks you to re-create missing entities like networks.
 However, the tool does not know how the instance was configured.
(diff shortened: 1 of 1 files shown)


lxd/storage/backend_lxd.go
```diff
@@ -7951,17 +7951,58 @@ func (b *lxdBackend) detectUnknownInstanceAndCustomVolumes(vol *drivers.Volume,
...
+				Status: api.StoragePoolStatusCreated,
+			}
+
+			customVolPool, err = NewTemporary(b.state, &poolInfo)
+			if err != nil {
+				return fmt.Errorf("Failed to initialise unknown pool %q: %w", customVolConfigPool.Name, err)
+			}
+
+			// Populate configuration with default values.
+			err := customVolPool.Driver().FillConfig()
```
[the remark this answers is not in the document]
roosterfish: We don't care about the outer `err` in this case.

lxd/api_internal_recover.go
```diff
@@ -166,6 +166,54 @@ func appendUnknownVolumeConfig(originalPoolName string, projectName string, volu
...
+
+		// Nothing left to identify, the pool exists.
+		return "", nil
+	}
+
+	// The pool is missing so create a dependency error.
+	configItems := make([]string, 0, len(volPool.Config))
+	for key, value := range volPool.Config {
+		configItems = append(configItems, fmt.Sprintf("%s=%q", key, value))
+	}
```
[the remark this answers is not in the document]
roosterfish: Good point, fixed.

test/suites/backup.sh
```diff
@@ -183,11 +179,94 @@ EOF
...
+
+  in_pipe="${LXD_IMPORT_DIR}/in.pipe"
+  out_pipe="${LXD_IMPORT_DIR}/out.pipe"
+  mkfifo "${in_pipe}" "${out_pipe}"
+  lxd recover < "${in_pipe}" > "${out_pipe}" &
+  lxd_recover_pid="$!"
+
+  # Open both pipes for continuous read/write and keep the fd open.
+  exec 3> "${in_pipe}"
+  exec 4< "${out_pipe}"
```
simondeziel, line 219: Clever!

doc/howto/disaster_recovery.md
```diff
@@ -25,75 +25,114 @@ The tool is available through the `lxd recover` command (note the `lxd` command
...
+ - Pool "default" using driver "zfs"
+Would you like to continue with scanning for lost volumes? (yes/no) [default=yes]:
 Scanning for unknown volumes...
 The following unknown volumes have been found:
- - Container "u1" on pool "default" in project "default" (includes 0 snapshots)
- - Container "u2" on pool "default" in project "default" (includes 0 snapshots)
+ - Virtual-Machine "v1" on pool "default" in project "default" (includes 0 snapshots)
+ - Volume "vol1" on pool "backup" in project "default" (includes 0 snapshots)
+You are currently missing the following:
+ - Pool "backup" using driver "lvm" (lvm.thinpool_name="LXDThinPool" lvm.vg_name="backup" source="backup" volatile.initial_source="/dev/sdc")
```
tomponline, line 64: @roosterfish does the backup.yaml file contain the member specific config for the pool btw?

tomponline: @roosterfish also why did you go for the approach of asking the user to jump into a new terminal and construct a `lxc storage create` command from the setting displayed, rather than offering an option to create the pool needed from within `lxd recover`?

Is this something to do with clustering?

doc/howto/disaster_recovery.md
```diff
@@ -25,75 +25,114 @@ The tool is available through the `lxd recover` command (note the `lxd` command
...
-This is how a recovery process could look:
+This is how a recovery process could look.
+We start by adding the `default` pool we still know about. On this pool we expect an instance `v1` which might use volumes from others unknown pools:
 
 ```{terminal}
+:input: lxc storage create default zfs source=/dev/sdb zfs.pool_name=default source.recover=true
+
+Storage pool default created
+
 :input: lxd recover
```
tomponline: If the user runs `lxd recover` with no storage pools defined, does the tool suggest they go and create them using the `source.recover=true` option?

lxd/api_internal_recover.go
```diff
@@ -166,6 +168,67 @@ func appendUnknownVolumeConfig(originalPoolName string, projectName string, volu
 	return nil
 }
 
+// identifyCustomVolumePool checks if the respective pool of a discovered custom volume is already known or missing.
+// This can be the case if a custom volume was discovered through an instance's backup config on another pool.
+// It may return a potential dependency error as string in case the pool doesn't yet exist.
```
tomponline: this sounds funny, why are we returning special errors as strings rather than `error`s?

lxd/api_internal_recover.go
```diff
@@ -391,13 +395,23 @@ func internalRecoverScan(ctx context.Context, s *state.State, userPools []api.St
 			profileProjectname = project.ProfileProjectFromRecord(projectInfo)
 			networkProjectName = project.NetworkProjectFromRecord(projectInfo)
 
-			for _, poolVol := range poolVols {
-				if poolVol.Instance == nil {
-					continue // Skip dependency checks for non-instance volumes.
+			for _, volConfig := range volConfigs {
+				dependencyErr, err := identifyCustomVolumePool(s, volConfig, pools)
```
tomponline, line 398: can we use a different type of error rather than a string?

doc/howto/disaster_recovery.md
roosterfish: > Is this something to do with clustering?

Yes as `lxd recover` as well as the other `lxd *` commands are scoped for the local LXD and don't have any knowledge about clustering.
Also I think we should not offer two different interfaces to create storage pools and `lxc storage create` already supports both standalone and clustered LXD.

lxd/api_internal_recover.go
```diff
@@ -501,81 +500,6 @@ func internalRecoverScan(ctx context.Context, s *state.State, userPools []api.St
 
 	// If in import mode and no dependency errors, then re-create missing DB records.
 
-	// Create the pools themselves.
-	for _, pool := range pools {
-		// Create missing storage pool DB record if neeed.
-		if pool.ID() == storagePools.PoolIDTemporary {
```
tomponline, line 463: Does this mean we dont need the "temporary pool" and "temporary pool ID concept" anymore too?

doc/howto/disaster_recovery.md
roosterfish: Good idea, let me add this. Right now it will just tell you there is nothing to recover.

tomponline (approved):
Thanks!

tomponline:
@roosterfish please rebase

lxd/api_internal_recover.go
roosterfish: I just wanted to differentiate an `error` occurred in the func from a dependency error which we track separately in a list once returned from the func just to prevent the `func (...) (error, error)` signature which might be not easy to grasp at a first glance.

But I can also name the return vars.

tomponline: yes 2 named return vars would be clearer

lxd/api_internal_recover.go
roosterfish: Fixed.

lxd/api_internal_recover.go
roosterfish: Fixed.

doc/howto/disaster_recovery.md
roosterfish: Fixed.

test/suites/backup.sh
roosterfish: Took me a bit to figure this. I initially just tried to `grep` on the `out_pipe` but `grep` seems to close it after it is done reading which ultimately let me do this approach :)

lxd/api_internal_recover.go
roosterfish: I am still making use of `NewTemporary` when finding custom vols on yet unknown pools. But I have to check whether we can remove `PoolIDTemporary`.

test/suites/backup.sh
simondeziel: Thanks! I'll try to remember when facing similar issue.

lxd/api_internal_recover.go
roosterfish: Ah ok, `PoolIDTemporary` const is assigned inside `storage.NewTemporary` so we have to keep it as it's still used.
And we still have to create the temporary pool so that we can pass it as all of the other pools and can then later create the dependency error about missing pools.

lxd/main_recover.go
```diff
@@ -54,6 +54,10 @@ func (c *cmdRecover) run(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("Failed getting existing storage pools: %w", err)
 		}
 
+		if len(existingPools) == 0 {
+			return errors.New(`This LXD server doesn't have any storage pools. First recover a pool using the "source.recover" config item`)
```
tomponline, line 58: Suggest:

> There are no storage pools defined in the database.
> Please first add the pool(s) to be recovered by running the "lxc storage create" for each pool and specifying its source configuration along with the "source.recover=true" setting to allow adding a pool that exists on the storage device already.

doc/howto/disaster_recovery.md
```diff
@@ -25,80 +25,120 @@ The tool is available through the `lxd recover` command (note the `lxd` command
 ## Recovery process
 
 When you run the tool, it scans all storage pools that still exist in the database, looking for missing volumes that can be recovered.
-You can also specify the details of any unknown storage pools (those that exist on disk but do not exist in the database), and the tool attempts to scan those too.
+Any unknown storage pools (those that exist on disk but do not exist in the database) which are discovered whilst scanning existing and unknown volumes
+are printed so they can be created manually using the `lxc storage create` command.
```
tomponline, line 29: I think its worth mentioning that they need to be added with `source.recover=true` and what that does.

doc/howto/disaster_recovery.md
```diff
@@ -114,11 +154,13 @@ lxc start u1
...
-| NAME |  STATE  |       IPV4        |                    IPV6                     |   TYPE    | SNAPSHOTS |
-+------+---------+-------------------+---------------------------------------------+-----------+-----------+
-| u1   | RUNNING | 192.0.2.49 (eth0) | 2001:db8:8b6:abfe:216:3eff:fe82:918e (eth0) | CONTAINER | 0         |
-+------+---------+-------------------+---------------------------------------------+-----------+-----------+
-| u2   | STOPPED |                   |                                             | CONTAINER | 0         |
-+------+---------+-------------------+---------------------------------------------+-----------+-----------+
++------+---------+----------------------+-----------------------------------------------+-----------------+-----------+
+| NAME |  STATE  |         IPV4         |                     IPV6                      |      TYPE       | SNAPSHOTS |
++------+---------+----------------------+-----------------------------------------------+-----------------+-----------+
+| u1   | RUNNING | 10.178.27.235 (eth0) | fd42:f9e6:53c2:ccc5:216:3eff:fe49:e549 (eth0) | CONTAINER       | 0         |
```
tomponline, line 160: Please can we use the test IP ranges in examples

doc/howto/disaster_recovery.md
roosterfish: Fair point, I'll add it together with the text improvements we talked about yesterday in our 1:1. Thanks!

doc/howto/disaster_recovery.md
roosterfish: Fixed in https://github.com/canonical/lxd/pull/17280.

doc/howto/disaster_recovery.md
roosterfish: Fixed in https://github.com/canonical/lxd/pull/17280. Added a link to the new recovery docs.

lxd/main_recover.go
roosterfish: Added, removed the dot at the end to comply with other error messages and added `command` after "lxc storage create".

Fixed in https://github.com/canonical/lxd/pull/17280.

PR #17269, opened 2025-12-16
Cluster: control-plane cluster role

kadinsayani:
This PR introduces the new `control-plane` cluster member role to LXD. This role provides a mechanism to explicitly designate which nodes are eligible to participate in the dqlite database (Raft) quorum.

By separating the control plane (database members) from the data plane (worker members), LXD clusters can be safely autoscaled. External autoscalers can target non-control-plane members for removal with the guarantee that they are not critical to the cluster's Raft consensus.

Specification: https://docs.google.com/document/d/1mZdZMBvf3wlP0lgxEEv2-YZjnMNguZH7_roPZxQ-FZg/edit?usp=sharing.

### DIFF
--- a/test/suites/clustering.sh
+++ b/test/suites/clustering.sh
@@ -4483,30 +4483,30 @@ test_clustering_roles() {
   [ "$(CLIENT_DEBUG="" SHELL_TRACING="" LXD_DIR="${LXD_ONE_DIR}" lxc cluster role add node1 invalid-role 2>&1)" = 'Error: Invalid cluster role "invalid-role"' ]
 
   # Reject typo in role name
-  [ "$(CLIENT_DEBUG="" SHELL_TRACING="" LXD_DIR="${LXD_ONE_DIR}" lxc cluster role add node1 event-hubb 2>&1)" = 'Error: Invalid cluster role "event-hubb"' ]
+  [ "$(CLIENT_DEBUG="" SHELL_TRACING="" LXD_DIR="${LXD_ONE_DIR}" lxc cluster role add node1 control-planee 2>&1)" = 'Error: Invalid cluster role "control-planee"' ]
 
   # Reject duplicate roles in request
-  [ "$(CLIENT_DEBUG="" SHELL_TRACING="" LXD_DIR="${LXD_ONE_DIR}" lxc cluster role add node1 event-hub,event-hub 2>&1)" = 'Error: Duplicate role "event-hub" in request' ]
+  [ "$(CLIENT_DEBUG="" SHELL_TRACING="" LXD_DIR="${LXD_ONE_DIR}" lxc cluster role add node1 control-plane,control-plane 2>&1)" = 'Error: Duplicate role "control-plane" in request' ]
 
   # Accept valid manual role addition
-  LXD_DIR="${LXD_ONE_DIR}" lxc cluster role add node1 event-hub
-  LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node1 | grep -xF -- "- event-hub"
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster role add node1 control-plane
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node1 | grep -xF -- "- control-plane"
 
   # Accept adding multiple manual roles
   LXD_DIR="${LXD_ONE_DIR}" lxc cluster role add node1 ovn-chassis
-  LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node1 | grep -xF -- "- event-hub"
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node1 | grep -xF -- "- control-plane"
   LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node1 | grep -xF -- "- ovn-chassis"
 
   # Reject adding role member already has
-  [ "$(CLIENT_DEBUG="" SHELL_TRACING="" LXD_DIR="${LXD_ONE_DIR}" lxc cluster role add node1 event-hub 2>&1)" = 'Error: Member "node1" already has role "event-hub"' ]
+  [ "$(CLIENT_DEBUG="" SHELL_TRACING="" LXD_DIR="${LXD_ONE_DIR}" lxc cluster role add node1 control-plane 2>&1)" = 'Error: Member "node1" already has role "control-plane"' ]
 
   # Accept removing manual role
-  LXD_DIR="${LXD_ONE_DIR}" lxc cluster role remove node1 event-hub
-  ! LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node1 | grep -xF -- "- event-hub" || false
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster role remove node1 control-plane
+  ! LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node1 | grep -xF -- "- control-plane" || false
   LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node1 | grep -xF -- "- ovn-chassis"
 
   # Reject removing role member does not have
-  [ "$(CLIENT_DEBUG="" SHELL_TRACING="" LXD_DIR="${LXD_ONE_DIR}" lxc cluster role remove node1 event-hub 2>&1)" = 'Error: Member "node1" does not have role "event-hub"' ]
+  [ "$(CLIENT_DEBUG="" SHELL_TRACING="" LXD_DIR="${LXD_ONE_DIR}" lxc cluster role remove node1 control-plane 2>&1)" = 'Error: Member "node1" does not have role "control-plane"' ]
 
   sub_test "Control plane mode behavior"
   # Test safe incremental assignment: add control-plane to 2 members (below minimum for control plane mode)


test/suites/clustering.sh
```diff
@@ -5013,10 +5014,79 @@ test_clustering_roles() {
...
+
+  echo "==> Test 2-member cluster prevents removing leader when spare lacks control-plane"
+  echo "Reduce max_voters back to 3 before shutting down nodes to maintain quorum"
+  LXD_DIR="${LXD_ONE_DIR}" lxc config set cluster.max_voters=3
+  sleep 12 # Wait for rebalancing
+
+  echo "Shutdown node3, node4, node5"
   LXD_DIR="${LXD_THREE_DIR}" lxd shutdown
+  LXD_DIR="${LXD_FOUR_DIR}" lxd shutdown
+  LXD_DIR="${LXD_FIVE_DIR}" lxd shutdown
```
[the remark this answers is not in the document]
kadinsayani: Actually, the [commit structure table](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#commit-structure) specifies that test changes should use the prefix `tests:`.

doc/api-extensions.md
```diff
@@ -2884,3 +2884,11 @@ Before this was only partially possible for some of the drivers (e.g. by using `
 The new pool `source.recover` configuration key can be set per cluster member to allow reuse of an existing pool `source`.
 What it does not allow is reusing the same source for multiple storage pools.
 The LVM storage driver has the specific `lvm.vg.force_reuse` configuration key for this purpose.
+
+## `clustering_control_plane`
+
+Adds a new `control-plane` cluster member role that can be manually assigned to designate which members participate in Raft consensus. Control-plane mode remains inactive until the number of members with the `control-plane` role reaches `cluster.max_voters`. During this inactive period, all cluster members remain eligible for automatic promotion to database roles. Once control-plane mode activates, only members with the `control-plane` role are eligible to become voters, standbys, or the database leader. Members without the `control-plane` role are automatically assigned the `RAFT_SPARE` role and are excluded from automatic promotion to database roles, enabling safe scaling of cluster members without affecting quorum.
```
tomponline: Thanks for this explanation.

Reading it now I'm not sure about the relationship between `control-plane` members and `cluster.max_voters` defined here though.

The existing `cluster.max_voters` setting takes an odd number >= 3.

Was there a reason you tied the trigger point for control-plane activation to the members role for this setting, rather than specifying that `control-planes` members needed to be an odd number >= 3?

Would it be valid to have cluster members with `control-plane` role be greater than the  `cluster.max_voters` setting?

I'm concerned that strictly tying those two concepts together will make it inflexible when adding/removing cluster members - i.e too easy to accidentally switch the whole cluster back to non-control-plane mode - which in a large cluster would be quite problematic.

kadinsayani: > Thanks for this explanation.
> 
> Reading it now I'm not sure about the relationship between `control-plane` members and `cluster.max_voters` defined here though.
> 
> The existing `cluster.max_voters` setting takes an odd number >= 3.
> 
> Was there a reason you tied the trigger point for control-plane activation to the members role for this setting, rather than specifying that `control-planes` members needed to be an odd number >= 3?

The only reason was semantic alignment with `cluster.max_voters`. Let's say `cluster.max_voters` is set to 5 and `control-plane` is assigned to 3 members. In this case, only 3 members would actually be assigned `database-voter`. Upon further consideration, I think this behavior is fine since it creates a hard boundary of members eligible for promotion tied to the number of members assigned the `control-plane` role.

> Would it be valid to have cluster members with `control-plane` role be greater than the `cluster.max_voters` setting?

Yes. If `cluster.max_voters` is set to 3 but `control-plane` is assigned to 5 members, we have a pool of 5 eligible candidates. 3 of the 5 `control-plane` members will be voters, and the remaining 2 would be standbys (up to `cluster.max_standby`).

Another valid scenario is having an even number of control-plane members. For example, with `cluster.max_voters=3` and `cluster.max_standby=3`, you could assign `control-plane` to 6 members: 3 would become voters (satisfying the odd requirement), and 3 would become standbys. The odd requirement is already enforced by `cluster.max_voters` itself, so there's no need to constrain the number of control-plane members to be odd.

> I'm concerned that strictly tying those two concepts together will make it inflexible when adding/removing cluster members - i.e too easy to accidentally switch the whole cluster back to non-control-plane mode - which in a large cluster would be quite problematic.

You raise a valid concern. We can decouple the `control-plane` role from `cluster.max_voters` and only require that the `control-plane` role is assigned to >= 3 of members for activation. This will create a hard boundary of cluster members eligible for promotion while providing the flexibility to scale independently of the `cluster.max_voters` configuration.

test/suites/clustering.sh
```diff
@@ -4377,6 +4377,71 @@ test_clustering_roles() {
...
+  echo "node5 demoted to spare after losing control-plane"
+
+  echo '==> Test control plane mode becoming inactive when dropping below minimum members'
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster role remove node3 control-plane
+  sleep 12 # Wait for heartbeat to deactivate control plane mode (2 < 3)
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster list
+
+  control_plane_count=$(LXD_DIR="${LXD_ONE_DIR}" lxc cluster list -f json | jq '[.[] | select(any(.roles[]; . == "control-plane"))] | length')
+  echo "Control plane members: ${control_plane_count} (should be 2, mode inactive)"
+  [ "${control_plane_count}" -eq 2 ]
```
simondeziel: This could be done without the shell comparison and with just `jq --exit-status '... | length == 2`, isn't it?

test/suites/clustering.sh
```diff
@@ -4377,6 +4377,71 @@ test_clustering_roles() {
...
+
+  # Control plane mode is now inactive (2 < 3), so all members are eligible again
+  echo "Verify at least one of the former spares (node4 or node5) gets promoted"
+  sleep 12 # Wait for promotion
+  spare_with_db_role=$(LXD_DIR="${LXD_ONE_DIR}" lxc cluster list -f json | jq -r '[.[] | select(.server_name == "node4" or .server_name == "node5") | select(any(.roles[]; test("database-(leader|voter|standby)")))] | .[0].server_name // ""')
+  [ -n "${spare_with_db_role}" ]
+
+  voter_count=$(LXD_DIR="${LXD_ONE_DIR}" lxc cluster list -f json | jq '[.[] | select(any(.roles[]; contains("database-voter") or contains("database-leader")))] | length')
+  [ "${voter_count}" -eq 3 ] # Should have 3 voters (max_voters=3)
```
simondeziel: Same note about a pure `jq` assertion.

test/suites/clustering.sh
```diff
@@ -4377,6 +4377,71 @@ test_clustering_roles() {
...
+  # Control plane mode remains inactive until 3 members have the control-plane role
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster role add node1 control-plane
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster role add node2 control-plane
+  sleep 12 # Wait for heartbeat
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster list
+
+  # Control plane mode is inactive (2 < 3), so all members are still eligible for database roles
+  echo "Verify node4 or node5 (spares without control-plane) can still have database roles"
+  spare_with_db_role=$(LXD_DIR="${LXD_ONE_DIR}" lxc cluster list -f json | jq -r '[.[] | select(.server_name == "node4" or .server_name == "node5") | select(any(.roles[]; test("database-(leader|voter|standby)")))] | .[0].server_name // ""')
+  [ -n "${spare_with_db_role}" ]
```
simondeziel: Could this be done as a pure `jq --exit-status` assertion instead of setting a variable and testing it to be non-null?

kadinsayani: Yes, thanks for pointing these out :)

test/suites/clustering.sh
```diff
@@ -4377,6 +4377,68 @@ test_clustering_roles() {
...
+  ! LXD_DIR="${LXD_ONE_DIR}" lxc cluster show node5 | grep -E -- "- database-(leader|voter|standby)" || false
+  echo "node5 demoted to spare after losing control-plane"
+
+  echo '==> Test control plane mode becoming inactive when dropping below minimum members'
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster role remove node3 control-plane
+  sleep 12 # Wait for heartbeat to deactivate control plane mode (2 < 3)
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster list
+
+  control_plane_count=$(LXD_DIR="${LXD_ONE_DIR}" lxc cluster list -f json | jq --exit-status '[.[] | select(any(.roles[]; . == "control-plane"))] | length == 2')
+  echo "Control plane members: ${control_plane_count} (should be 2, mode inactive)"
```
simondeziel: `control_plane_count` will contain the text `true` as it's the result of the `length == 2` comparison. If the variable and comments are not essential, maybe drop them and only keep the `jq` assertion?

test/suites/clustering.sh
```diff
@@ -4377,6 +4377,68 @@ test_clustering_roles() {
...
+
+  # Control plane mode is now inactive (2 < 3), so all members are eligible again
+  echo "Verify at least one of the former spares (node4 or node5) gets promoted"
+  sleep 12 # Wait for promotion
+  LXD_DIR="${LXD_ONE_DIR}" lxc cluster list -f json | jq --exit-status '[.[] | select(.server_name == "node4" or .server_name == "node5") | select(any(.roles[]; test("database-(leader|voter|standby)")))] | .[0].server_name // empty' > /dev/null
+
+  voter_count=$(LXD_DIR="${LXD_ONE_DIR}" lxc cluster list -f json | jq --exit-status '[.[] | select(any(.roles[]; contains("database-voter") or contains("database-leader")))] | length')
+  [ "${voter_count}" -eq 3 ] # Should have 3 voters (max_voters=3)
```
simondeziel: This could also be a `jq` assertion.

doc/api-extensions.md
kadinsayani: Spec and implementation updated :)

simondeziel (dismissed):

doc/api-extensions.md
```diff
@@ -2884,3 +2884,11 @@ Before this was only partially possible for some of the drivers (e.g. by using `
 The new pool `source.recover` configuration key can be set per cluster member to allow reuse of an existing pool `source`.
 What it does not allow is reusing the same source for multiple storage pools.
 The LVM storage driver has the specific `lvm.vg.force_reuse` configuration key for this purpose.
+
+## `clustering_control_plane`
```
minaelee, line 2985:
```suggestion
(extension-clustering-control-plane)=
## `clustering_control_plane`
```

Please add link targets to new API extension documentation as a matter of course. This will make it easier to link to from, for example, release notes and other docs.

doc/api-extensions.md
```diff
@@ -2884,3 +2884,11 @@ Before this was only partially possible for some of the drivers (e.g. by using `
...
+
+Adds a new `control-plane` cluster member role that can be manually assigned to designate which members participate in Raft consensus. Control plane mode remains inactive until the number of members with the `control-plane` role reaches 3 or more. During this inactive period, all cluster members remain eligible for automatic promotion to database roles. Once control plane mode activates, only members with the `control-plane` role are eligible to become voters, standbys, or the database leader. Members without the `control-plane` role are automatically assigned the `RAFT_SPARE` role and are excluded from automatic promotion to database roles, enabling safe scaling of cluster members without affecting quorum.
+
+If no cluster members have the `control-plane` role assigned (the default), or if fewer than 3 members have the role, all members remain eligible for automatic promotion to database roles, maintaining backwards compatibility with existing cluster behavior.
```
```suggestion
If fewer than 3 members are assigned the `control-plane` role, all members remain eligible for automatic promotion to database roles, maintaining backwards compatibility with existing cluster behavior.
```

doc/api-extensions.md
```diff
@@ -2884,3 +2884,11 @@ Before this was only partially possible for some of the drivers (e.g. by using `
 The new pool `source.recover` configuration key can be set per cluster member to allow reuse of an existing pool `source`.
 What it does not allow is reusing the same source for multiple storage pools.
 The LVM storage driver has the specific `lvm.vg.force_reuse` configuration key for this purpose.
+
+## `clustering_control_plane`
+
+Adds a new `control-plane` cluster member role that can be manually assigned to designate which members participate in Raft consensus. Control plane mode remains inactive until the number of members with the `control-plane` role reaches 3 or more. During this inactive period, all cluster members remain eligible for automatic promotion to database roles. Once control plane mode activates, only members with the `control-plane` role are eligible to become voters, standbys, or the database leader. Members without the `control-plane` role are automatically assigned the `RAFT_SPARE` role and are excluded from automatic promotion to database roles, enabling safe scaling of cluster members without affecting quorum.
```
```suggestion
Adds a new `control-plane` cluster member role that can be manually assigned to designate which members participate in Raft consensus. Control plane mode is inactive by default until the number of members assigned the `control-plane` role reaches 3 or more. During this inactive period, all cluster members are eligible for automatic promotion to database roles. Once control plane mode activates, only members with the `control-plane` role are eligible to become voters, standbys, or the database leader. Members without the `control-plane` role are automatically assigned the `RAFT_SPARE` role and are excluded from automatic promotion to database roles, enabling safe scaling of cluster members without affecting quorum.
```

doc/reference/dqlite-internals.md
```diff
@@ -28,6 +28,10 @@ For more information on the Canonical Dqlite Raft implementation, see [`dqlite/s
 
 ### LXD cluster roles
```
minaelee, line 30:
```suggestion
(dqlite-internals-lxd-cluster-roles)=
### LXD cluster roles
```

doc/explanation/clusters.md
```diff
@@ -55,14 +55,28 @@ Automatic roles are assigned by LXD itself and cannot be modified by the user.
 | `database-voter`      | yes           | Voting member of the distributed database |
 | `database-leader`     | yes           | Current leader of the distributed database |
 | `database-standby`    | yes           | Stand-by (non-voting) member of the distributed database |
+| `control-plane`       | no            | Eligible to participate in Raft as voter, standby, or leader. Members without this role are assigned as spares and excluded from automatic promotion. |
```
minaelee:
```suggestion
| `control-plane`       | no            | Eligible to participate in Raft as voter, standby, or leader. When control plane mode is active, members without this role are assigned as spares and excluded from automatic promotion. |
```

doc/explanation/clusters.md
```diff
@@ -87,7 +101,7 @@ See {ref}`cluster-recover` for more information.
 #### Failure domains
 
 You can use failure domains to indicate which cluster members should be given preference when assigning roles to a cluster member that has gone offline.
-For example, if a cluster member that currently has the database role gets shut down, LXD tries to assign its database role to another cluster member in the same failure domain, if one is available.
+For example, if a cluster member that currently has the `database-voter` role gets shut down and the `control-plane` role is active, LXD tries to promote another `control-plane` cluster member in the same failure domain to voter, if one is available. If no members have the `control-plane` role assigned (the default), any suitable member in the same failure domain can be promoted instead.
```
minaelee:
```suggestion
For example, if a cluster member that currently has the `database-voter` role is shut down and control plane mode is active, LXD tries to promote another `control-plane` cluster member in the same failure domain to voter, if one is available. If no members have the `control-plane` role assigned (the default), any suitable member in the same failure domain can be promoted instead.
```

doc/api-extensions.md
```diff
@@ -2884,3 +2884,11 @@ Before this was only partially possible for some of the drivers (e.g. by using `
...
+
+If no cluster members have the `control-plane` role assigned (the default), or if fewer than 3 members have the role, all members remain eligible for automatic promotion to database roles, maintaining backwards compatibility with existing cluster behavior.
+
+The `control-plane` role is displayed alongside the database role (e.g., `database-leader`, `database-voter`, `database-standby`) for members participating in Raft.
```
```suggestion
If fewer than 3 members are assigned the `control-plane` role, all members remain eligible for automatic promotion to database roles, maintaining backwards compatibility with existing cluster behavior.

The `control-plane` role is displayed alongside the database role (`database-leader`, `database-voter`, `database-standby`) for members participating in Raft.

For more information, see {ref}`dqlite-internals-lxd-cluster-roles` and {ref}`cluster-manage-control-plane`.
```

doc/explanation/clusters.md
```diff
@@ -55,14 +55,28 @@ Automatic roles are assigned by LXD itself and cannot be modified by the user.
...
+Once active, only members with the `control-plane` role can participate in Raft and be assigned as voters, standbys, or the leader.
+Members without the `control-plane` role are automatically assigned the `RAFT_SPARE` role and are excluded from automatic promotion to database roles.
+Spare members can still run instances and act as "worker" members for hosting workloads.
+
+You can assign the `control-plane` role to more members than {config:option}`server-cluster:cluster.max_voters` to create a pool of eligible candidates.
+For example, if you assign control-plane to 5 members when `cluster.max_voters` is 3, all 5 members are eligible for database roles, but only 3 will be promoted to voters based on the configuration.
+
+If no cluster members have the `control-plane` role assigned (the default), or if fewer than 3 members have the role, all members are eligible for automatic promotion to database roles.
+
+See {ref}`cluster-manage` for instructions on using the control-plane role.
```
minaelee:
```suggestion
For example, if you assign `control-plane` to 5 members when `cluster.max_voters` is 3, all 5 members are eligible for database roles, but only 3 will be promoted to voters based on the configuration.

If no cluster members have the `control-plane` role assigned (the default), or if fewer than 3 members have the role, all members are eligible for automatic promotion to database roles.

See {ref}`cluster-manage-control-plane` for instructions on using the `control-plane` role.
```

doc/howto/cluster_manage.md
```diff
@@ -62,12 +62,31 @@ To add or remove a {ref}`member role <clustering-member-roles>` for a cluster me
...
+       lxc cluster role add node1 control-plane
+       lxc cluster role add node2 control-plane
+       lxc cluster role add node3 control-plane
+
+2. Verify activation by running `lxc cluster list` - only members with the `control-plane` role will display database roles.
+
+3. New members joining the cluster will automatically be spares. To make them eligible for database roles, assign the `control-plane` role to them.
+
+You can assign `control-plane` to more members than `cluster.max_voters` to create a pool of eligible candidates. For example, 5 members with control-plane when `cluster.max_voters` is 3 means 3 become voters from the 5 candidates.
+
```
minaelee: Let's move this section out of the note and into its own section following the `Assign member roles` section, starting like this:

```
(cluster-manage-control-plane)=
### Use the control-plane role
```

doc/howto/cluster_manage.md
```diff
@@ -62,12 +62,31 @@ To add or remove a {ref}`member role <clustering-member-roles>` for a cluster me
...
+       lxc cluster role add node1 control-plane
+       lxc cluster role add node2 control-plane
+       lxc cluster role add node3 control-plane
+
+2. Verify activation by running `lxc cluster list` - only members with the `control-plane` role will display database roles.
+
+3. New members joining the cluster will automatically be spares. To make them eligible for database roles, assign the `control-plane` role to them.
+
+You can assign `control-plane` to more members than `cluster.max_voters` to create a pool of eligible candidates. For example, 5 members with control-plane when `cluster.max_voters` is 3 means 3 become voters from the 5 candidates.
```
minaelee:
```suggestion
1. Assign the role to at least 3 members:

       lxc cluster role add <member1> control-plane
       lxc cluster role add <member2> control-plane
       lxc cluster role add <member3> control-plane

2. Verify activation by running `lxc cluster list` — only members with the `control-plane` role will display database roles.

3. New members join the cluster as spares by default. To make them eligible for database roles, assign the `control-plane` role to them.

You can assign `control-plane` to more members than {config:option}`server-cluster:cluster.max_voters` to create a pool of eligible candidates. For example, having 5 `control-plane` members when {config:option}`server-cluster:cluster.max_voters` is 3 means 3 of the 5 candidates become voters. If one of the voters becomes unavailable, one of the remaining two candidates takes its place.

For more information, see: {ref}`clustering-control-plane`.
```

doc/explanation/clusters.md
```diff
@@ -55,14 +55,28 @@ Automatic roles are assigned by LXD itself and cannot be modified by the user.
...
+| `control-plane`       | no            | Eligible to participate in Raft as voter, standby, or leader. Members without this role are assigned as spares and excluded from automatic promotion. |
 | `event-hub`           | no            | Exchange point (hub) for the internal LXD events (requires at least two) |
 | `ovn-chassis`         | no            | Uplink gateway candidate for OVN networks |
 
 The default number of voter members ({config:option}`server-cluster:cluster.max_voters`) is three.
 The default number of stand-by members ({config:option}`server-cluster:cluster.max_standby`) is two.
 With this configuration, your cluster will remain operational as long as you switch off at most one voting member at a time.
 
-See {ref}`cluster-manage` for more information.
+The `control-plane` role is optional and is not assigned by default.
```
minaelee, line 67:
```suggestion

(clustering-control-plane)=
#### The `control-plane` role

The `control-plane` role is optional and is not assigned by default.
```

doc/howto/cluster_recover.md
```diff
@@ -24,6 +24,7 @@ Run `lxd cluster --help` for an overview of all available commands.
 ## Database members
 
 Every LXD cluster has a specific number of members (configured through {config:option}`server-cluster:cluster.max_voters`) that serve as voting members of the distributed database.
+These are cluster members with or without the `control-plane` role that have been assigned the `database-voter` or `database-leader` role.
```
minaelee:
```suggestion
If the cluster uses {ref}`control plane mode <clustering-control-plane>`, only cluster members with the `control-plane` role act as voters.
```

kadinsayani:
> Some structural and clarifying edits requested. Thanks for your work on this!

Comments addressed, thanks for your review!

minaelee (dismissed):
Ty!

lxd/cluster/membership.go
```diff
@@ -311,10 +311,47 @@ func Accept(state *state.State, gateway *Gateway, name, address string, schema,
...
+	err = state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+		// Check if control plane mode is active.
+		controlPlaneActive, err = tx.ControlPlaneActive()
+		if err != nil {
+			return fmt.Errorf("Failed checking %q role: %w", db.ClusterRoleControlPlane, err)
+		}
+
+		if controlPlaneActive {
+			// Check if the new member has the control-plane role.
+			members, err := tx.GetNodes(ctx)
```
tomponline: Can we use something like `GetNodeByAddress` to avoid needing to pull back all node info?

kadinsayani: `GetNodeByAddress` and `GetNodes` both call `ClusterTx.nodes`, so that wouldn't improve performance, but it would read better. Perhaps we can add a new helper, something like `NodeHasRoleByAddress` or `GetNodeRolesByAddresses` with an optimized query.

tomponline: GetNodeByAddress does DB level filtering too. So more efficient if ur looking for a specific member.

kadinsayani: Ah yes, thanks. Updated.

tomponline:
@kadinsayani please rebase

lxd/cluster/membership.go
```diff
@@ -1114,8 +1114,56 @@ func Handover(state *state.State, gateway *Gateway, address string) (string, []d
...
+		}
+
+		// Skip non-control-plane members if control plane mode is active.
+		if controlPlaneActive && !slices.Contains(member.Roles, db.ClusterRoleControlPlane) {
+			continue
+		}
+
+		candidateAddress = candidate.Address
+		break
+	}
```
kadinsayani: New function to return candidates by address.

lxd/daemon.go
```diff
@@ -2595,8 +2604,14 @@ func (d *Daemon) nodeRefreshTask(heartbeatData *cluster.APIHeartbeat, isLeader b
...
 			if needsRebalance {
+				// Build member roles map from heartbeat data.
+				memberRoles := make(map[string][]db.ClusterRole, len(heartbeatData.Members))
+				for _, member := range heartbeatData.Members {
+					memberRoles[member.Address] = member.Roles
+				}
+
 				logger.Debug("Rebalancing member roles in heartbeat", logger.Ctx{"local": localClusterAddress})
-				err := rebalanceMemberRoles(context.Background(), d.State(), d.gateway, unavailableMembers)
+				err := rebalanceMemberRoles(context.Background(), d.State(), d.gateway, unavailableMembers, controlPlaneActive, memberRoles)
```
kadinsayani: rebalanceMemberRoles can accept candidate members.

Function to filter candidates members depending on control plane mode status.

lxd/cluster/membership.go
```diff
@@ -722,6 +722,67 @@ func NotifyHeartbeat(state *state.State, gateway *Gateway) {
...
+
+	return controlPlaneCount >= 3
+}
+
+// filterPromotionCandidates returns the subset of candidates that are eligible for
+// promotion based on control plane mode and member roles. When controlPlaneActive
+// is true, only candidates with the control-plane role are eligible.
+func filterPromotionCandidates(candidates []client.NodeInfo, memberRoles map[string][]db.ClusterRole, controlPlaneActive bool) []client.NodeInfo {
+	if !controlPlaneActive {
```
tomponline: The `controlPlaneActive` argument seems like an odd API choice, as this function is already passed sufficient info to work this out itself by calling `IsControlPlaneActive(memberRoles)` no?

lxd/cluster/membership.go
```diff
@@ -311,10 +311,40 @@ func Accept(state *state.State, gateway *Gateway, name, address string, schema,
...
+		canPromote = slices.Contains(newMember.Roles, db.ClusterRoleControlPlane)
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// If member doesn't have control-plane role, check if control plane mode is active.
+	if !canPromote {
+		memberRoles, err := getMemberRoles(context.TODO(), state)
```
tomponline: Can you tweak `getMemberRoles` to accept a `tx` and then perform this logic and optional retrieval inside the transaction above rather than starting another transaction?

lxd/cluster/membership.go
```diff
@@ -819,7 +819,11 @@ func filterPromotionCandidates(candidates []client.NodeInfo, memberRoles map[str
 //
 // If there's such spare node, return its address as well as the new list of
 // raft nodes.
-func Rebalance(state *state.State, gateway *Gateway, unavailableMembers []string) (string, []db.RaftNode, error) {
+//
+// The optional memberRoles parameter can be provided to avoid redundant database
+// queries when the caller already has this data (e.g., from heartbeats). If
+// memberRoles is nil, member data will be fetched from the database.
+func Rebalance(state *state.State, gateway *Gateway, unavailableMembers []string, memberRoles map[string][]db.ClusterRole) (string, []db.RaftNode, error) {
```
tomponline: This commit should come first before the usage of the updated function.

Also, can we avoid making optional DB retrieval behaviour of a function (as it complicates the function itself) and rather push up the requirement to retrieve the member roles to the caller site, so that the caller always has to provide these to this function.

We did similar re-work for some of @roosterfish recent PRs to avoid optional behaviour.
This might also allow for member roles to be retrieved in an existing transaction from some call sites by the looks of it.

lxd/cluster/membership.go
```diff
@@ -830,6 +834,59 @@ func Rebalance(state *state.State, gateway *Gateway, unavailableMembers []string
...
+
+			// If member doesn't have control-plane role but has a database role, demote it
+			if !slices.Contains(roles, db.ClusterRoleControlPlane) {
+				logger.Info("Found non-control-plane member with database role requiring demotion", logger.Ctx{"address": node.Address, "currentRole": node.Role, "local": localClusterAddress})
+
+				// Set target role to spare. The Assign() function will handle the two-phase
+				// voter -> standby -> spare transition automatically if needed.
+				for i := range nodes {
+					if nodes[i].Address == node.Address {
+						nodes[i].Role = db.RaftSpare
```
tomponline: should we demote down to Standby first so that we dont potentially end up with no database nodes?

For example if you have all database nodes currently outside of the members assigned the `control-plane` role, then if we step all database nodes down to spare, what happens?

lxd/cluster/membership.go
```diff
@@ -842,11 +899,14 @@ func Rebalance(state *state.State, gateway *Gateway, unavailableMembers []string
...
+	// Filter candidates to only those eligible for promotion.
+	candidates = filterPromotionCandidates(candidates, memberRoles, controlPlaneActive)
+	if len(candidates) == 0 {
+		return "", nodes, nil
+	}
 
-	// Check if we have a spare node that we can promote to the missing role.
 	candidateAddress := candidates[0].Address
-	logger.Info("Found cluster member whose role needs to be changed", logger.Ctx{"candidateAddress": candidateAddress, "newRole": role, "local": localClusterAddress})
+	logger.Info("Found cluster member requiring role change", logger.Ctx{"candidateAddress": candidateAddress, "newRole": role, "local": localClusterAddress})
```
tomponline, line 1085: feels like this should be a separate commit

lxd/cluster/membership.go
kadinsayani: I've removed that param, we can use `IsControlPlaneActive`.

lxd/cluster/membership.go
kadinsayani: Great suggestion, thanks.

lxd/cluster/membership.go
kadinsayani: We don't need to demote to standby first in Rebalance() because:

- Assign() already handles the two-phase transition: When it detects voter->spare, it automatically does voter->standby->spare with confirmation polling.
- Control plane mode self-deactivates: If all database nodes lose their control-plane role, you'd have < 3 control-plane members, so IsControlPlaneActive() returns false and this demotion block is skipped ensuring existing voters remain in place.

The scenario can't happen because demotions only occur when >=3 control-plane members exist, and they only target nodes without the control-plane role.

lxd/cluster/membership.go
kadinsayani: Done. I like that approach!

lxd/cluster/membership.go
tomponline: What about this:

9 member cluster (cluster.max_standby = 2, cluster.max_voters = 3), currently members 1,2,3,4 and 5 are automatically assigned DB roles (no control-plane members). Members 6,7,8 and 9 are spare.

Then I assign members 7, 8 and 9 as `control-plane` members.

So now:

> If control plane mode is active, demote non-control-plane members with database roles to spare.

occurs.

The non-control-plane members (1,2,3,4,5) are demoted to spare.

Where is the database at this point?

kadinsayani: Ah, I see the issue now.

kadinsayani: I'll fix this logic to follow the two-phase approach and maintain quorum throughout a transition to control plane mode.

kadinsayani: I've reworked the logic to:

- Promote before demote to maintain quorum;
- Follow a two-phase demotion (voter -> standby -> spare); and
- Avoid demotions if no control-plane candidate is available to prevent leaving the cluster in a degraded state.

lxd/cluster/membership.go
```diff
@@ -822,32 +826,161 @@ func Rebalance(state *state.State, gateway *Gateway, unavailableMembers []string
...
+				}
+
+				if slices.Contains(roles, db.ClusterRoleControlPlane) {
+					hasEligiblePromotion = true
+					break
+				}
+			}
+
+			if !hasEligiblePromotion {
+				logger.Info("Skipping control-plane voter demotion; no eligible control-plane promotion candidates", logger.Ctx{"local": localClusterAddress})
```
tomponline: you mean "non-control-plane voter demotion" right?

lxd/cluster/membership.go
```diff
@@ -805,13 +805,17 @@ func filterPromotionCandidates(candidates []client.NodeInfo, memberRoles map[str
 	return eligible
 }
 
-// Rebalance the raft cluster, trying to see if we have a spare online node
-// that we can promote to voter node if we are below membershipMaxRaftVoters,
-// or to standby if we are below membershipMaxStandBys.
+// Rebalance adjusts raft cluster roles by promoting spares to fill vacancies or
+// demoting members to enforce control-plane policy. Returns the address of the member
```
tomponline: > Returns the address of the member whose role should change and the updated node list

change to what?

lxd/cluster/membership.go
```diff
@@ -805,13 +805,17 @@ func filterPromotionCandidates(candidates []client.NodeInfo, memberRoles map[str
...
+// whose role should change and the updated node list, or empty string if no changes needed.
 //
-// If there's such spare node, return its address as well as the new list of
-// raft nodes.
-func Rebalance(state *state.State, gateway *Gateway, unavailableMembers []string) (string, []db.RaftNode, error) {
+// The memberRoles parameter enforces control-plane restrictions during rebalancing.
+//
+// Called repeatedly by rebalanceMemberRoles, returning one change per call to naturally
```
tomponline: When is it called repeatedly by `rebalanceMemberRoles`?

lxd/cluster/membership.go
```diff
@@ -805,13 +805,17 @@ func filterPromotionCandidates(candidates []client.NodeInfo, memberRoles map[str
 	return eligible
 }
 
-// Rebalance the raft cluster, trying to see if we have a spare online node
-// that we can promote to voter node if we are below membershipMaxRaftVoters,
-// or to standby if we are below membershipMaxStandBys.
+// Rebalance adjusts raft cluster roles by promoting spares to fill vacancies or
```
tomponline: @kadinsayani I've been looking at this in detail this morning and this function has grown from a simple wrapper around the `roles.Adjust` function from go-dqlite, to implementing additional logic ontop of what `roles.Adjust` is suggesting.

My concern with this approach is that there are now 2 functions deciding promotion/demotion logic.

As I dug into the `roles.Adjust` function I came across the `client.NodeMetadata` struct which is used to indicate which failure domain a node should be in (so this is "config" rather than "state").

It got me thinking whether we could extend the `client.NodeMetadata` struct to add a new field to indicate role preference. Or perhaps its not even needed and can we use the `Weight` field that exists already to prefer promoting members in the control plane?

lxd/cluster/membership.go
```diff
@@ -822,32 +826,161 @@ func Rebalance(state *state.State, gateway *Gateway, unavailableMembers []string
...
+			for _, node := range nodes {
+				if node.Role == db.RaftVoter {
+					continue
+				}
+
+				if slices.Contains(unavailableMembers, node.Address) {
+					continue
+				}
+
+				if !HasConnectivity(gateway.networkCert, gateway.state().ServerCert(), node.Address) {
```
tomponline: If we truly need this check then `newRolesChanges` is already doing it it seems, so we'd not want to ping every member multiple times in the same function.

lxd/cluster/membership.go
kadinsayani: Note the `goto again` statements:

https://github.com/canonical/lxd/blob/d70995a9f3008ec795e82b0809ec695b91a9f7dc/lxd/api_cluster.go#L2680-L2715

lxd/cluster/membership.go
kadinsayani: Thanks for raising this. I'll dig further into this and look into the `Weight` field today.

lxd/cluster/membership.go
tomponline: ah indeed, please can you lose that goto as part of this PR (or a separate one) please.

lxd/cluster/membership.go
kadinsayani: We could push more logic into the `RolesChanges` model, but there are trade-offs:

- `RolesChanges.Adjust` is promotion centric and doesn't express the demotion logic required for non-control-plane members during the transition phase. 
- `Weight` can bias candidate selections, but it doesn't forbid promotions or encode the "must demote non-control-plane members" logic. 

We could use `Weight` to deprioritize non‑control‑plane members, but since the promotion/demotion behavior is specific to LXD’s control‑plane feature, it seems clearer to keep it centralized in `Rebalance`.

If we add a new field to indicate role preference, we'll need to maintain our own implementation of `RolesChanges.Adjust`.

Here's what I propose:

- Extract the control plane promotion/demotion logic into a new function;
- Rename `Rebalance` to something more descriptive;
- Replace the `goto` statements in `rebalanceMemberRoles` with a `for` loop that continues until there are no more role changes.

lxd/cluster/membership.go
kadinsayani: Yes.

lxd/cluster/membership.go
tomponline: > If we add a new field to indicate role preference, we'll need to maintain our own implementation of `RolesChanges.Adjust`.

No, I would suggest we propose it go in go-dqlite, not maintain our own copy. That is what I'm trying to avoid.

tomponline: > RolesChanges.Adjust is promotion centric and doesn't express the demotion logic required for non-control-plane members during the transition phase.

Can we expand the `Adjust` function to add said logic change when adding the "preferred role" hint metadata field?

lxd/cluster/membership.go
kadinsayani: https://github.com/canonical/lxd/pull/17377

lxd/cluster/membership.go
kadinsayani: Yes, we can. To keep things moving along, can we keep the control plane logic on the LXD side for now, add a comment documenting future improvements, and re-visit in a later PR? Or would you prefer proposing the go-dqlite changes first?

tomponline: Can you try opening a pr against go-dqlite v3 with what we need, id appreciate getting the logic reviewed by the octo team anyway.

kadinsayani: https://github.com/canonical/go-dqlite/pull/419

kadinsayani:
@tomponline ready for another review when you have a chance. Wondering if I should include the `event-hub` deprecation and transition to `control-plane` here or in a follow-up PR?

tomponline:
> Wondering if I should include the `event-hub` deprecation and transition to `control-plane` here or in a follow-up PR?

yes please

kadinsayani:
> > Wondering if I should include the `event-hub` deprecation and transition to `control-plane` here or in a follow-up PR?
> 
> yes please

Done :)

lxd/cluster/membership.go
```diff
@@ -819,14 +819,245 @@ func filterPromotionCandidates(candidates []client.NodeInfo, memberRoles map[str
 	return eligible
 }
 
+// adjustRoles determines the next role change using the same generic roles
+// algorithm as [app.RolesChanges.Adjust], while applying LXD-specific control-plane restrictions.
+func adjustRoles(roles *app.RolesChanges, leaderID uint64, nodes []db.RaftNode, connectivity map[string]bool, memberRoles map[string][]db.ClusterRole) (client.NodeRole, []client.NodeInfo) {
```
tomponline, line 823: Please can you split this into 2 commits; the first bringing the role adjustment logic as-is from go-dqlite, the second introducing the changes need for control plane.

tomponline:
needs a rebase too please

lxd/cluster/membership.go
kadinsayani: Done.

lxd/cluster/membership.go
```diff
@@ -311,10 +311,41 @@ func Accept(state *state.State, gateway *Gateway, name, address string, schema,
 		return nil, errors.New("Cannot convert maximum standby cluster members to int: Upper bound exceeded")
 	}
 
-	if count > 1 && voters < int(maxVoters) {
-		node.Role = db.RaftVoter
-	} else if standbys < int(maxStandBy) {
-		node.Role = db.RaftStandBy
+	// Check if the new member is eligible for promotion.
+	// Members are always eligible unless control plane mode is active and they lack the control-plane role.
+	var canPromote bool
```
tomponline, line 316: In the interests of reducing cyclomatic complexity I feel this could be split out into a separate function that returns this bool, rather than complicating `Accept()` further?

lxd/api_cluster.go
```diff
@@ -1211,13 +1211,28 @@ func internalClusterPostRebalance(d *Daemon, r *http.Request) response.Response
 
 // Check if there's a dqlite node whose role should be changed, and post a
 // change role request if so.
-func rebalanceMemberRoles(ctx context.Context, s *state.State, gateway *cluster.Gateway, unavailableMembers []string) error {
+func rebalanceMemberRoles(ctx context.Context, s *state.State, gateway *cluster.Gateway, unavailableMembers []string, memberRoles map[string][]db.ClusterRole) error {
 	if s.ShutdownCtx.Err() != nil {
 		return nil
 	}
 
+	if memberRoles == nil {
```
tomponline, line 1219: I'm not a fan of this style of "optional input, and if not supplied, we load it ourselves" logic. It makes reasoning about a function's behaviour unnecessarily complicated. 

I'm fine with adding the new `memberRoles` function argument here, but it should always be provided externally, rather than loaded here if not provided.

We went through similar design choices with @roosterfish storage recover work.

lxd/daemon.go
```diff
@@ -2643,6 +2643,7 @@ func (d *Daemon) nodeRefreshTask(heartbeatData *cluster.APIHeartbeat, isLeader b
 	if isLeader && unavailableMembers != nil && len(heartbeatData.Members) > 1 {
```
tomponline, line 2643: Please can you refactor everything inside this code block that is related to the decision as to whether to call `rebalanceMemberRoles` or `upgradeNodesWithoutRaftRole` into a separate function (including calling the functions mentioned here), and then add this new logic to that function.

This will help reduce the cyclomatic complexity of this `nodeRefreshTask` function, which is getting out of hand.

lxd/cluster/membership.go
```diff
@@ -819,6 +819,125 @@ func filterPromotionCandidates(candidates []client.NodeInfo, memberRoles map[str
 	return eligible
 }
 
+// adjustRoles determines the next role change using the same generic roles algorithm as [app.RolesChanges.Adjust].
```
tomponline, line 822: Does `[app.RolesChanges.Adjust]` link to the go-dqlite package?

lxd/cluster/membership.go
```diff
@@ -819,6 +819,125 @@ func filterPromotionCandidates(candidates []client.NodeInfo, memberRoles map[str
...
+func adjustRoles(roles *app.RolesChanges, leaderID uint64) (client.NodeRole, []client.NodeInfo) {
```
tomponline, line 823: Suggest calling it `rolesAdjust` to keep this aligned with our convention of naming things "<thing><action>" rather than "<action><thing>".

doc/explanation/clusters.md
```diff
@@ -55,14 +55,32 @@ Automatic roles are assigned by LXD itself and cannot be modified by the user.
 | `database-voter`      | yes           | Voting member of the distributed database |
 | `database-leader`     | yes           | Current leader of the distributed database |
 | `database-standby`    | yes           | Stand-by (non-voting) member of the distributed database |
+| `control-plane`       | no            | Eligible to participate in Raft as voter, standby, or leader. When control plane mode is active, members without this role are assigned as spares and excluded from automatic promotion. |
```
tomponline, line 58: The other descriptions in this table don't end with fullstop but this one does.

lxd/cluster/membership.go
kadinsayani: Good suggestion, thanks.

lxd/cluster/membership.go
kadinsayani: Yes it does. I thought it would be helpful to have a quick reference to it.

doc/explanation/clusters.md
kadinsayani: https://github.com/canonical/lxd/pull/17831/changes/7c94281f3de34ed9f8926f700d0f944997a92805

lxd/cluster/membership.go
kadinsayani: https://github.com/canonical/lxd/pull/17831/changes/767b0aa6cc050f29d0932c6f9f2a97175d657822

lxd/daemon.go
kadinsayani: https://github.com/canonical/lxd/pull/17831/changes/7d1179655bd0764b0421613debd24c933d485e83

lxd/api_cluster.go
kadinsayani: https://github.com/canonical/lxd/pull/17831/changes/3caf5b7be753f9c46938b3122586c04b00585a0b

lxd/cluster/membership.go
kadinsayani: https://github.com/canonical/lxd/pull/17831/changes/8d58ce7a6ff991e4c5038aa876f3f4d803e0cfd1

PR #17270, opened 2025-12-17
docs: avoid permanent redirects

simondeziel:


simondeziel:
There is another link check redirection that caught my attention:

```
(howto/oidc_keycloak: line   11) redirect  https://github.com/keycloak/keycloak/releases/download/25.0.4/keycloak-25.0.4.zip - with Found to https://release-assets.githubusercontent.com/github-production-release-asset/11125589/02036f13-b425-468d-af2f-48f71bf417fd?...
```

This 25.0.4 release is not very current, I wonder if we should instead point users to https://github.com/keycloak/keycloak/releases/latest and ask them to download the `.zip` (unless we can get the `latest` tag to lead us to a `.zip`, somehow).

minaelee (approved):
Ty!

minaelee:
> There is another link check redirection that caught my attention:
> 
> ```
> (howto/oidc_keycloak: line   11) redirect  https://github.com/keycloak/keycloak/releases/download/25.0.4/keycloak-25.0.4.zip - with Found to https://release-assets.githubusercontent.com/github-production-release-asset/11125589/02036f13-b425-468d-af2f-48f71bf417fd?...
> ```
> 
> This 25.0.4 release is not very current, I wonder if we should instead point users to https://github.com/keycloak/keycloak/releases/latest and ask them to download the `.zip` (unless we can get the `latest` tag to lead us to a `.zip`, somehow).

I agree. Alternatively, we could point them to https://www.keycloak.org/downloads and ask them to download the Keycloak server zip at the top. It's still not a direct link to the zip but a little easier to find.

The only drawback I can see is that the instructions might fall out of date, but that's also the case with the instructions for other OIDC methods that use web portals, and still better than pointing to an outdated version.

minaelee (approved):

PR #17281, opened 2025-12-19
Dell PowerStore storage driver (part 1)

adanishal:
This PR adds basic support for Dell PowerStore appliances.

## Implemented methods from the `"github.com/canonical/lxd/storage/drivers".Driver` interface

- Internal
  - init
  - load
  - isRemote
  - defaultVMBlockFilesystemSize
  - Info
  - HasVolume
- Export struct details
  - SourceIdentifier
- Pool
  - FillConfig
  - Create
  - Delete
  - Mount
  - Unmount
  - GetResources
  - Validate
  - ValidateSource
  - Update
- Volumes
  - FillVolumeConfig
  - ValidateVolume
  - CreateVolume
  - DeleteVolume
  - RenameVolume
  - UpdateVolume
  - GetVolumeUsage
  - GetVolumeDiskPath
  - ListVolumes
  - MountVolume
  - UnmountVolume
  - SetVolumeQuota

## Methods from the `"github.com/canonical/lxd/storage/drivers".Driver` interface left out
- Volumes
  - CreateVolumeFromCopy
  - RefreshVolume
  - CanDelegateVolume
  - DelegateVolume
- Snapshots
  - MountVolumeSnapshot
  - UnmountVolumeSnapshot
  - CreateVolumeSnapshot
  - DeleteVolumeSnapshot
  - RenameVolumeSnapshot
  - VolumeSnapshots
  - CheckVolumeSnapshots
  - RestoreVolume
- Migration
  - MigrationTypes
  - MigrateVolume
  - CreateVolumeFromMigration
- Backup
  - BackupVolume
  - CreateVolumeFromBackup

## Checklist

### DIFF
--- a/lxd/metadata/configuration.json
+++ b/lxd/metadata/configuration.json
@@ -7064,6 +7064,234 @@
 				]
 			}
 		},
+		"storage-powerstore": {
+			"pool-conf": {
+				"keys": [
+					{
+						"powerstore.discovery": {
+							"defaultdesc": "the list of discovery addresses",
+							"longdesc": "A comma-separated list of NVMe or iSCSI discovery addresses.",
+							"shortdesc": "List of discovery addresses.",
+							"type": "string"
+						}
+					},
+					{
+						"powerstore.gateway": {
+							"longdesc": "",
+							"scope": "global",
+							"shortdesc": "Address of the PowerStore Gateway",
+							"type": "string"
+						}
+					},
+					{
+						"powerstore.gateway.verify": {
+							"defaultdesc": "`true`",
+							"longdesc": "",
+							"scope": "global",
+							"shortdesc": "Whether to verify the PowerStore Gateway's certificate",
+							"type": "bool"
+						}
+					},
+					{
+						"powerstore.mode": {
+							"defaultdesc": "the discovered mode",
+							"longdesc": "The mode gets discovered automatically if the system provides the necessary kernel modules.\nSupported values are `iscsi` and `nvme`.",
+							"scope": "global",
+							"shortdesc": "How volumes are mapped to the local server",
+							"type": "string"
+						}
+					},
+					{
+						"powerstore.target": {
+							"defaultdesc": "the list of target addresses",
+							"longdesc": "A comma-separated list of NVMe or iSCSI target addresses. When empty targets are discovered via discovery endpoints.",
+							"shortdesc": "List of target addresses.",
+							"type": "string"
+						}
+					},
+					{
+						"powerstore.transport": {
+							"defaultdesc": "the discovered transport",
+							"longdesc": "The transport gets discovered automatically if the system provides the necessary kernel modules.\nSupported values are `tcp`.",
+							"scope": "global",
+							"shortdesc": "Transport layer used when transferring volumes data to the local server.",
+							"type": "string"
+						}
+					},
+					{
+						"powerstore.user.name": {
+							"defaultdesc": "`admin`",
+							"longdesc": "",
+							"scope": "global",
+							"shortdesc": "User for PowerStore Gateway authentication",
+							"type": "string"
+						}
+					},
+					{
+						"powerstore.user.password": {
+							"longdesc": "",
+							"scope": "global",
+							"shortdesc": "Password for PowerStore Gateway authentication",
+							"type": "string"
+						}
+					},
+					{
+						"rsync.bwlimit": {
+							"defaultdesc": "`0` (no limit)",
+							"longdesc": "When `rsync` must be used to transfer storage entities, this option specifies the upper limit\nto be placed on the socket I/O.",
+							"scope": "global",
+							"shortdesc": "Upper limit on the socket I/O for `rsync`",
+							"type": "string"
+						}
+					},
+					{
+						"rsync.compression": {
+							"defaultdesc": "`true`",
+							"longdesc": "",
+							"scope": "global",
+							"shortdesc": "Whether to use compression while migrating storage pools",
+							"type": "bool"
+						}
+					},
+					{
+						"volume.size": {
+							"defaultdesc": "`10GiB`",
+							"longdesc": "The size must be in multiples of 1 MiB. The minimum size is 1 MiB and maximum is 256 TiB.",
+							"scope": "global",
+							"shortdesc": "Size/quota of the storage volume",
+							"type": "string"
+						}
+					}
+				]
+			},
+			"volume-conf": {
+				"keys": [
+					{
+						"block.filesystem": {
+							"condition": "block-based volume with content type `filesystem`",
+							"defaultdesc": "same as `volume.block.filesystem`",
+							"longdesc": "Valid options are: `btrfs`, `ext4`, `xfs`\nIf not set, `ext4` is assumed.",
+							"scope": "global",
+							"shortdesc": "File system of the storage volume",
+							"type": "string"
+						}
+					},
+					{
+						"block.mount_options": {
+							"condition": "block-based volume with content type `filesystem`",
+							"defaultdesc": "same as `volume.block.mount_options`",
+							"longdesc": "",
+							"scope": "global",
+							"shortdesc": "Mount options for block-backed file system volumes",
+							"type": "string"
+						}
+					},
+					{
+						"security.shared": {
+							"condition": "virtual-machine or custom block volume",
+							"defaultdesc": "same as `volume.security.shared` or `false`",
+							"longdesc": "Enabling this option allows sharing the volume across multiple instances despite the possibility of data loss.\n",
+							"scope": "global",
+							"shortdesc": "Enable volume sharing",
+							"type": "bool"
+						}
+					},
+					{
+						"security.shifted": {
+							"condition": "custom volume",
+							"defaultdesc": "same as `volume.security.shifted` or `false`",
+							"longdesc": "Enabling this option allows attaching the volume to multiple isolated instances.",
+							"scope": "global",
+							"shortdesc": "Enable ID shifting overlay",
+							"type": "bool"
+						}
+					},
+					{
+						"security.unmapped": {
+							"condition": "custom volume",
+							"defaultdesc": "same as `volume.security.unmappped` or `false`",
+							"longdesc": "",
+							"scope": "global",
+							"shortdesc": "Disable ID mapping for the volume",
+							"type": "bool"
+						}
+					},
+					{
+						"size": {
+							"defaultdesc": "same as `volume.size`",
+							"longdesc": "The size must be in multiples of 1 MiB. The minimum size is 1 MiB and maximum is 256 TiB.",
+							"scope": "global",
+							"shortdesc": "Size/quota of the storage volume",
+							"type": "string"
+						}
+					},
+					{
+						"snapshots.expiry": {
+							"condition": "custom volume",
+							"defaultdesc": "same as `volume.snapshots.expiry`",
+							"longdesc": "Specify an expression like `1M 2H 3d 4w 5m 6y`.",
+							"scope": "global",
+							"shortdesc": "When snapshots are to be deleted",
+							"type": "string"
+						}
+					},
+					{
+						"snapshots.pattern": {
+							"condition": "custom volume",
+							"defaultdesc": "same as `volume.snapshots.pattern` or `snap%d`",
+							"longdesc": "You can specify a naming template that is used for scheduled snapshots and unnamed snapshots.\n\nThe `snapshots.pattern` option takes a Pongo2 template string to format the snapshot name.\n\nTo add a time stamp to the snapshot name, use the Pongo2 context variable `creation_date`.\nMake sure to format the date in your template string to avoid forbidden characters in the snapshot name.\nFor example, set `snapshots.pattern` to `{{ creation_date|date:'2006-01-02_15-04-05' }}` to name the snapshots after their time of creation, down to the precision of a second.\n\nAnother way to avoid name collisions is to use the placeholder `%d` in the pattern.\nFor the first snapshot, the placeholder is replaced with `0`.\nFor subsequent snapshots, the existing snapshot names are taken into account to find the highest number at the placeholder's position.\nThis number is then incremented by one for the new name.",
+							"scope": "global",
+							"shortdesc": "Template for the snapshot name",
+							"type": "string"
+						}
+					},
+					{
+						"snapshots.schedule": {
+							"condition": "custom volume",
+							"defaultdesc": "same as `snapshots.schedule`",
+							"longdesc": "Specify either a cron expression (`\u003cminute\u003e \u003chour\u003e \u003cdom\u003e \u003cmonth\u003e \u003cdow\u003e`), a comma-separated list of schedule aliases (`@hourly`, `@daily`, `@midnight`, `@weekly`, `@monthly`, `@annually`, `@yearly`), or leave empty to disable automatic snapshots (the default).",
+							"scope": "global",
+							"shortdesc": "Schedule for automatic volume snapshots",
+							"type": "string"
+						}
+					},
+					{
+						"volatile.devlxd.owner": {
+							"defaultdesc": "DevLXD owner identity ID",
+							"longdesc": "",
+							"scope": "global",
+							"shortdesc": "The ID of the DevLXD identity which owns the volume",
+							"type": "string"
+						}
+					},
+					{
+						"volatile.idmap.last": {
+							"condition": "filesystem",
+							"longdesc": "",
+							"shortdesc": "JSON-serialized UID/GID map that has been applied to the volume",
+							"type": "string"
+						}
+					},
+					{
+						"volatile.idmap.next": {
+							"condition": "filesystem",
+							"longdesc": "",
+							"shortdesc": "JSON-serialized UID/GID map that has been applied to the volume",
+							"type": "string"
+						}
+					},
+					{
+						"volatile.uuid": {
+							"defaultdesc": "random UUID",
+							"longdesc": "",
+							"scope": "global",
+							"shortdesc": "The volume's UUID",
+							"type": "string"
+						}
+					}
+				]
+			}
+		},
 		"storage-pure": {
 			"pool-conf": {
 				"keys": [
--- a/lxd/storage/connectors/connector_nvme.go
+++ b/lxd/storage/connectors/connector_nvme.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"net"
 	"os"
 	"path/filepath"
 	"slices"
@@ -24,21 +25,98 @@ var _ Connector = &connectorNVMe{}
 // nvmeDiskDevicePrefix is the prefix of the NVMe disk device name in /dev/disk/by-id/.
 const nvmeDiskDevicePrefix = "nvme-eui."
 
-// SubtypeNVMESubsys defines an NVMe subsystem type (from https://github.com/linux-nvme/libnvme/blob/97886cb68d238ccbbed804a275851f63e490b22f/src/nvme/fabrics.c#L99).
-const SubtypeNVMESubsys = "nvme subsystem"
-
 type connectorNVMe struct {
 	common
 }
 
 // NVMeDiscoveryLogRecord represents an NVMe discovery entry.
 type NVMeDiscoveryLogRecord struct {
-	SubType string `json:"subtype"`
-	SubNQN  string `json:"subnqn"`
+	TransportType              string `json:"trtype"`
+	AddressFamily              string `json:"adrfam"`
+	TransportAddress           string `json:"traddr"`
+	TransportServiceIdentifier string `json:"trsvcid"`
+	SubType                    string `json:"subtype"`
+	SubNQN                     string `json:"subnqn"`
 }
+
+const (
+	// NVMeDefaultDiscoveryPort is the default port number for NVMe/TCP discovery
+	// controller.
+	NVMeDefaultDiscoveryPort = "8009"
+
+	// NVMeDefaultTransportPort is the default port number for NVMe/TCP I/O
+	// controller.
+	NVMeDefaultTransportPort = "4420"
+)
+
+// Transport type definitions (from https://github.com/linux-nvme/libnvme/blob/97886cb68d238ccbbed804a275851f63e490b22f/src/nvme/fabrics.c#L73).
+const (
+	nvmeTransportTypeFC  = "fc"
+	nvmeTransportTypeTCP = "tcp"
+)
+
+// Address family definitions (from https://github.com/linux-nvme/libnvme/blob/97886cb68d238ccbbed804a275851f63e490b22f/src/nvme/fabrics.c#L85).
+const (
+	nvmeAddressFamilyIPv4         = "ipv4"
+	nvmeAddressFamilyIPv6         = "ipv6"
+	nvmeAddressFamilyFibreChannel = "fibre-channel"
+)
+
+// SubtypeNVMESubsys defines an NVMe subsystem type (from https://github.com/linux-nvme/libnvme/blob/97886cb68d238ccbbed804a275851f63e490b22f/src/nvme/fabrics.c#L99).
+const SubtypeNVMESubsys = "nvme subsystem"
 
 type nvmeDiscoveryLog struct {
 	Records []NVMeDiscoveryLogRecord `json:"records"`
+}
+
+// nvmeFilterDiscoveryLog filters out entries from the provided discovery log
+// that do not describe proper NVMe over TCP targets.
+func nvmeFilterDiscoveryLog(log *nvmeDiscoveryLog) {
+	if len(log.Records) == 0 {
+		return
+	}
+
+	filteredRecords := make([]NVMeDiscoveryLogRecord, 0, len(log.Records))
+	for _, record := range log.Records {
+		if record.SubType != SubtypeNVMESubsys {
+			continue
+		}
+
+		if record.TransportType != nvmeTransportTypeTCP {
+			continue
+		}
+
+		if record.AddressFamily != nvmeAddressFamilyIPv4 && record.AddressFamily != nvmeAddressFamilyIPv6 {
+			continue
+		}
+
+		filteredRecords = append(filteredRecords, record)
+	}
+
+	log.Records = filteredRecords
+}
+
+// nvmeAddMissingServiceIDToDiscoveryLog ensure all port numbers (transport
+// service identifiers) are set. For non specified ports function uses
+// the discovery controller port if present, otherwise uses the default
+// transport port number.
+func nvmeAddMissingServiceIDToDiscoveryLog(log *nvmeDiscoveryLog, discoveryAddress string) {
+	if len(log.Records) == 0 {
+		return
+	}
+
+	_, transportServiceID, err := net.SplitHostPort(discoveryAddress)
+	if err != nil {
+		transportServiceID = NVMeDefaultTransportPort
+	}
+
+	for i := range log.Records {
+		if log.Records[i].TransportServiceIdentifier != "" {
+			continue
+		}
+
+		log.Records[i].TransportServiceIdentifier = transportServiceID
+	}
 }
 
 // Type returns the type of the connector.
@@ -84,6 +162,7 @@ func (c *connectorNVMe) QualifiedName() (string, error) {
 func (c *connectorNVMe) Connect(ctx context.Context, targetQN string, targetAddresses ...string) (revert.Hook, error) {
 	// Connects to the provided target address, if the connection is not yet established.
 	connectFunc := func(ctx context.Context, session *session, targetAddr string) error {
+		targetAddr = shared.EnsurePort(targetAddr, NVMeDefaultTransportPort)
 		if session != nil && slices.Contains(session.addresses, targetAddr) {
 			// Already connected.
 			return nil
@@ -94,7 +173,12 @@ func (c *connectorNVMe) Connect(ctx context.Context, targetQN string, targetAddr
 			return err
 		}
 
-		_, err = shared.RunCommand(ctx, "nvme", "connect", "--transport", "tcp", "--traddr", targetAddr, "--nqn", targetQN, "--hostnqn", hostNQN, "--hostid", c.serverUUID)
+		transportAddr, transportServiceID, err := net.SplitHostPort(targetAddr)
+		if err != nil {
+			return fmt.Errorf("Bad transport address %q: %w", targetAddr, err)
+		}
+
+		_, err = shared.RunCommand(ctx, "nvme", "connect", "--transport", "tcp", "--traddr", transportAddr, "--trsvcid", transportServiceID, "--nqn", targetQN, "--hostnqn", hostNQN, "--hostid", c.serverUUID)
 		if err != nil {
 			return fmt.Errorf("Failed connecting to target %q on %q via NVMe: %w", targetQN, targetAddr, err)
 		}
@@ -227,14 +311,27 @@ func (c *connectorNVMe) findSession(targetQN string) (*session, error) {
 		// The "address" file contains one line per connection,
 		// each in format "traddr=<ip>,trsvcid=<port>,...".
 		for line := range strings.SplitSeq(string(fileBytes), "\n") {
-			parts := strings.SplitSeq(strings.TrimSpace(line), ",")
-			for part := range parts {
+			parts := strings.Split(strings.TrimSpace(line), ",")
+
+			transportAddr := ""
+			for _, part := range parts {
 				addr, ok := strings.CutPrefix(part, "traddr=")
 				if ok {
-					session.addresses = append(session.addresses, addr)
+					transportAddr = addr
 					break
 				}
 			}
+
+			transportServiceID := NVMeDefaultTransportPort
+			for _, part := range parts {
+				port, ok := strings.CutPrefix(part, "trsvcid=")
+				if ok {
+					transportServiceID = port
+					break
+				}
+			}
+
+			session.addresses = append(session.addresses, net.JoinHostPort(transportAddr, transportServiceID))
 		}
 	}
 
@@ -258,7 +355,13 @@ func (c *connectorNVMe) Discover(ctx context.Context, targetAddresses ...string)
 
 	var discoveryLog nvmeDiscoveryLog
 	for _, targetAddr := range targetAddresses {
-		stdout, err := shared.RunCommand(ctx, "nvme", "discover", "--transport", "tcp", "--traddr", targetAddr, "--hostnqn", hostNQN, "--hostid", c.serverUUID, "--output-format", "json")
+		discoveryAddr, discoveryServiceID, err := net.SplitHostPort(shared.EnsurePort(targetAddr, NVMeDefaultDiscoveryPort))
+		if err != nil {
+			logger.Warn("Bad discovery address", logger.Ctx{"target_address": targetAddr, "err": err})
+			continue
+		}
+
+		stdout, err := shared.RunCommand(ctx, "nvme", "discover", "--transport", "tcp", "--traddr", discoveryAddr, "--trsvcid", discoveryServiceID, "--hostnqn", hostNQN, "--hostid", c.serverUUID, "--output-format", "json")
 		if err != nil {
 			// Exit code 110 is returned if the target address cannot be reached.
 			logger.Warn("Failed connecting to discovery target", logger.Ctx{"target_address": targetAddr, "err": err})
@@ -279,7 +382,10 @@ func (c *connectorNVMe) Discover(ctx context.Context, targetAddresses ...string)
 			return nil, fmt.Errorf("Failed unmarshaling the returned discovery log entries from %q: %w", targetAddr, err)
 		}
 
-		// Unmarshalling the response from the discovery succeeded, break the loop.
+		nvmeAddMissingServiceIDToDiscoveryLog(&discoveryLog, targetAddr)
+		nvmeFilterDiscoveryLog(&discoveryLog)
+
+		// Unmarshaling the response from the discovery succeeded, break the loop.
 		break
 	}
 
--- a/lxd/storage/drivers/driver_powerstore.go
+++ b/lxd/storage/drivers/driver_powerstore.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/canonical/lxd/lxd/operations"
 	"github.com/canonical/lxd/lxd/storage/connectors"
+	"github.com/canonical/lxd/lxd/storage/drivers/powerstoreclient"
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/validate"
 )
@@ -64,6 +65,10 @@ var powerStoreVersion string
 
 type powerstore struct {
 	common
+
+	// Holds the low level client for the PowerStore API.
+	// Use powerstore.client() to retrieve the initialized client struct.
+	apiClient *powerstoreclient.Client
 
 	// Holds the low level connector for the PowerStore driver.
 	// Use powerstore.connector() to retrieve the initialized connector.
--- a/lxd/storage/drivers/driver_powerstore.go
+++ b/lxd/storage/drivers/driver_powerstore.go
@@ -1,0 +1,356 @@
+package drivers
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/canonical/lxd/lxd/operations"
+	"github.com/canonical/lxd/lxd/storage/connectors"
+	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/validate"
+)
+
+// powerStoreDefaultUser represents the default PowerStore user name.
+const powerStoreDefaultUser = "admin"
+
+// powerStoreMinVolumeSizeBytes represents the minimal PowerStore volume size
+// in bytes.
+const powerStoreMinVolumeSizeBytes = 1 * 1024 * 1024 // 1MiB
+
+// powerStoreMinVolumeSizeUnit represents the minimal PowerStore volume size
+// expressed as unit string.
+const powerStoreMinVolumeSizeUnit = "1MiB"
+
+// powerStoreMaxVolumeSizeBytes represents the maximum PowerStore volume size
+// in bytes.
+const powerStoreMaxVolumeSizeBytes = 256 * 1024 * 1024 * 1024 * 1024 // 256TiB
+
+// powerStoreMaxVolumeSizeUnit represents the maximum PowerStore volume size
+// expressed as unit string.
+const powerStoreMaxVolumeSizeUnit = "256TiB"
+
+// powerStoreMinVolumeSizeAlignmentUnit represents the alignment unit for
+// the PowerStore volume size (each volume size needs to be multiplicate of
+// this value).
+const powerStoreMinVolumeSizeAlignmentUnit = "1MiB"
+
+const (
+	powerStoreModeNVME  string = "nvme"
+	powerStoreModeISCSI string = "iscsi"
+)
+
+const (
+	powerStoreTransportTCP string = "tcp"
+)
+
+// powerStoreSupportedConnectorTypes list all connector types supported by
+// the PowerStore driver.
+var powerStoreSupportedModesAndTransports = driverModesAndTransports{
+	{
+		Mode:          powerStoreModeNVME,
+		Transport:     powerStoreTransportTCP,
+		ConnectorType: connectors.TypeNVME,
+	},
+	{
+		Mode:          powerStoreModeISCSI,
+		Transport:     powerStoreTransportTCP,
+		ConnectorType: connectors.TypeISCSI,
+	},
+}
+
+var powerStoreLoaded bool
+var powerStoreVersion string
+
+type powerstore struct {
+	common
+
+	// Holds the low level connector for the PowerStore driver.
+	// Use powerstore.connector() to retrieve the initialized connector.
+	storageConnector connectors.Connector
+}
+
+// load is used to run one-time action per-driver rather than per-pool.
+func (d *powerstore) load() error {
+	// Done if previously loaded.
+	if powerStoreLoaded {
+		return nil
+	}
+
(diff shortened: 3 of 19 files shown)


tomponline (changes requested):
please may you consult https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#commit-structure and ensure commit structure is compliant. Thanks.

lxd/storage/drivers/driver_powerstore.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+	}
+
+	return nil
+}
+
+// Create is called during pool creation and is effectively using an empty driver struct.
+// WARNING: The Create() function cannot rely on any of the struct attributes being set.
+func (d *powerstore) Create() error {
+	return ErrNotSupported
```
roosterfish: Hey @adanishal, since you pinged me to take a look, I have taken a quick look over the code to prepare for the review. 

I was wondering if the code already reached a "ready" state on which base I can do a review. As the `Create` returns an error, IIRC LXD will error out already during pool creation. Please let me know if I am missing something.
Is there a specific section in the code you want to have some feedback on to continue? 

Also I was looking for funcs like the `CreateVolume` which would allow you to create a vol on the array to go through the basic steps/naming functions you have in the code. But the `getVolumeName` seems to be only used by a unit test right now.

roosterfish: Nvm, I was checking an old commit.

roosterfish: Thanks for the ping :)

roosterfish: I see now you have added the features commit by commit.

tomponline:
@adanishal please can you include in the PR summary an explanation as to what is included in this PR and what is still left to do. Thanks

lxd/storage/drivers/driver_powerstore.go
```diff
@@ -0,0 +1,483 @@
...
+
+// powerStoreSupportedTransports list all transports supported by the PowerStore driver.
+var powerStoreSupportedTransports = []string{
+	"tcp",
+}
+
+// powerStoreModeAndTransportToConnectorType maps mode and transport of the PowerStore driver to matching connector type.
+var powerStoreModeAndTransportToConnectorType = map[string]map[string]string{
+	connectors.TypeNVME: {
+		"tcp": connectors.TypeNVME,
```
roosterfish: It feels redundant that you have to indicate the support for `tcp` twice, once in `powerStoreModeAndTransportToConnectorType` and second in `powerStoreSupportedTransports`.

Can you implicitly derive `powerStoreSupportedTransports` from `powerStoreModeAndTransportToConnectorType`?

lxd/storage/drivers/driver_powerstore.go
```diff
@@ -0,0 +1,483 @@
...
+// discoverTargetQualifiedName discovers target QN (qualified name).
+func (d *powerstore) discoverTargetQualifiedName(targetAddress []string) (string, error) {
+	connector, err := d.connector()
+	if err != nil {
+		return "", err
+	}
+
+	discoveryLogRecords, err := connector.Discover(d.state.ShutdownCtx, targetAddress...)
+	if err != nil {
+		return "", fmt.Errorf("discovering targets: %w", err)
```
roosterfish: Errors should start with a capital letter. There are a few more of those at least in this file.

lxd/storage/drivers/driver_powerstore.go
```diff
@@ -0,0 +1,483 @@
...
+
+		case connectors.NVMeDiscoveryLogRecord:
+			if record.SubType != connectors.SubtypeNVMESubsys {
+				continue
+			}
+
+			return record.SubNQN, nil
+
+		default:
+			return "", fmt.Errorf("unsupported discovery log record entry type %T", discoveryLogRecord)
```
roosterfish: Worth aligning this error message with the one used in the other drivers for consistency?

lxd/storage/drivers/driver_powerstore.go
```diff
@@ -0,0 +1,483 @@
...
+}
+
+// Info returns info about the driver and its environment.
+func (d *powerstore) Info() Info {
+	return Info{
+		Name:                         "powerstore",
+		Version:                      powerStoreVersion,
+		DefaultBlockSize:             d.defaultBlockVolumeSize(),
+		DefaultVMBlockFilesystemSize: d.defaultVMBlockFilesystemSize(),
+		OptimizedImages:              false,
```
roosterfish, line 121: Maybe I am not up to date with the latest discussions on this, but have we decided we cannot do optimized images with PowerStore?

lxd/storage/drivers/driver_powerstore.go
```diff
@@ -0,0 +1,483 @@
...
+// FillConfig populates the storage pool's configuration file with the default values.
+func (d *powerstore) FillConfig() error {
+	if d.config["powerstore.user.name"] == "" {
+		d.config["powerstore.user.name"] = powerStoreDefaultUser
+	}
+
+	// Try to discover the PowerStore operation mode and transport.
+	switch {
+	case d.config["powerstore.mode"] == "" || d.config["powerstore.transport"] == "":
+		mode, transport, err := d.discoverModeAndTransport()
```
roosterfish: To me it looks like this `discoverModeAndTransport` can be something defined on `(d *common)` so that also the other drivers can use this func? Same as `defaultBlockVolumeSize`.

lxd/storage/drivers/driver_powerstore.go
```diff
@@ -0,0 +1,483 @@
...
+		transport, err := d.discoverTransport(d.config["powerstore.mode"])
+		if err != nil {
+			return err
+		}
+
+		d.config["powerstore.transport"] = transport
+	}
+
+	// Set default volume size if not provided.
+	if d.config["volume.size"] == "" {
```
roosterfish, line 163: Do we have to configure this here as it's the default 10GiB anyway?

The reason we did it back then for PowerFlex was because the default has to be 8GiB and 10GiB isn't allowed.

lxd/storage/drivers/driver_powerstore.go
```diff
@@ -0,0 +1,483 @@
...
+	}
+
+	return nil
+}
+
+// discoverModeAndTransport attempts to discover operation mode and transport without using the storage pool's configuration.
+func (d *powerstore) discoverModeAndTransport() (mode string, transport string, err error) {
+	for _, transport = range powerStoreSupportedTransports {
+		for _, mode = range powerStoreSupportedModes {
+			connectorType := powerStoreModeAndTransportToConnectorType[mode][transport]
```
roosterfish: Please see my earlier comment on having only `powerStoreModeAndTransportToConnectorType`.

lxd/storage/drivers/driver_powerstore.go
roosterfish: Also there are various funcs for this mode/transport discovery. 
Would it make sense to have a single func that allows passing empty `""` strings for either `mode` or `transport` and then picks a default? This would also move the logic (switch/case) into this func so a storage driver only has to call `discoverModeAndTransport`.

lxd/storage/drivers/driver_powerstore.go
```diff
@@ -0,0 +1,483 @@
...
+		}
+	}
+	return "", errors.New("failed to discover PowerStore transport")
+}
+
+// Validate checks that all provided keys are supported and that no conflicting or missing configuration is present.
+func (d *powerstore) Validate(config map[string]string) error {
+	rules := map[string]func(value string) error{
+		// lxdmeta:generate(entities=storage-powerstore; group=pool-conf; key=powerstore.user.name)
+		// Must have at least SystemAdmin role to give LXD full control over managed storage pools.
```
roosterfish: Just to double check that this isn't a leftover from PowerFlex. Does PowerStore also use `SystemAdmin` role?

lxd/storage/drivers/driver_powerstore.go
```diff
@@ -0,0 +1,483 @@
...
+		//  shortdesc: Whether to verify the PowerStore Gateway's certificate
+		//  scope: global
+		"powerstore.gateway.verify": validate.Optional(validate.IsBool),
+		// lxdmeta:generate(entities=storage-powerstore; group=pool-conf; key=powerstore.target)
+		// A comma-separated list of NVMe or iSCSI target addresses.
+		// ---
+		//  type: string
+		//  defaultdesc: the discovered mode
+		//  shortdesc: List of target addresses.
+		"powerstore.target": validate.Optional(validate.IsListOf(validate.IsNetworkAddress)),
```
roosterfish: In PowerFlex the targets are called SDT. Does PowerStore also use a custom naming or do they use target?

lxd/storage/drivers/driver_powerstore.go
```diff
@@ -0,0 +1,483 @@
...
+}
+
+// Unmount unmounts the storage pool.
+func (d *powerstore) Unmount() (bool, error) {
+	return true, nil
+}
+
+// GetResources returns the pool resource usage information.
+func (d *powerstore) GetResources() (*api.ResourcesStoragePool, error) {
+	metrics, err := d.client().GetApplianceMetrics(context.Background())
```
roosterfish: You can use `d.state.ShutdownCtx`.

lxd/storage/drivers/driver_powerstore.go
```diff
@@ -0,0 +1,483 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	res := &api.ResourcesStoragePool{}
+
+	for _, m := range metrics {
+		res.Space.Total += uint64(m.LastPhysicalTotalSpace)
```
roosterfish: Worth setting the type of `LastPhysicalTotalSpace` to `uint64` directly to omit the conversion?

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,812 @@
...
+	IdleTimeout     time.Duration
+	LastInteraction atomic.Pointer[time.Time]
+	AuthToken       string
+	CSRFToken       string
+}
+
+func newPowerStoreLoginSession(id string, idleTimeout time.Duration, authToken, csrfToken string) *powerStoreLoginSession {
+	ls := &powerStoreLoginSession{
+		ID:          id,
+		IdleTimeout: idleTimeout - 30*time.Second, // subtract to add safety margin for a potential time skew
```
roosterfish: Haven't yet fully checked how the `IdleTimeout` is used here, but what we do in the other drivers (at least Pure and PowerFlex) is that whenever the API returns a 401, we just attempt to login again.
This doesn't require tracking any idle time on our side as this can potentially be also changed on the server side or our access token can just be revoked which would require logging in again.

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,812 @@
...
+func (ls *powerStoreLoginSession) Interacted() {
+	now := time.Now()
+	ls.LastInteraction.Store(&now)
+}
+
+func makePowerStoreLoginSessionKey(url, username, password string) string {
+	url = base64.StdEncoding.EncodeToString([]byte(url))
+	username = base64.StdEncoding.EncodeToString([]byte(username))
+	password = base64.StdEncoding.EncodeToString([]byte(password))
+	return fmt.Sprintf("%s:%s:%s", url, username, password)
```
roosterfish: Do we have to url encode also the user/pass?

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,812 @@
...
+	if err != nil {
+		e.decoderErr = fmt.Errorf("unmarshal HTTP error response body: %w", err)
+	}
+
+	return e
+}
+
+// Error attempts to return all kinds of errors from the PowerStore API in a nicely formatted way.
+func (e *powerStoreError) Error() string {
+	msg := "PowerSore API error"
```
roosterfish: `PowerStore

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,812 @@
...
+type powerStoreErrorInstanceResource struct {
+	Severity    string   `json:"severity"`
+	Code        string   `json:"code"`
+	MessageL10n string   `json:"message_l10n"`
+	Arguments   []string `json:"arguments,omitempty"`
+}
+
+// powerStoreClient holds the PowerStore HTTP API client.
+type powerStoreClient struct {
+	gateway                  string
```
roosterfish: Just saw Alletra is doing it differently than the others. But Pure and PowerFlex just pass the driver `d` itself when deriving a new client. 
This seems to be more straightforward as you require less fields on the `powerStoreClient` struct and can freely access configuration settings from `d.config`.

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,812 @@
...
+	return dst, nil
+}
+
+func (c *powerStoreClient) doHTTPRequest(ctx context.Context, method string, path string, requestData, responseData any, requestEditors ...func(*http.Request) error) (*http.Response, error) {
+	body, err := c.marshalHTTPRequestBody(requestData)
+	if err != nil {
+		return nil, fmt.Errorf("marshall HTTP request body: %s: %w", path, err)
+	}
+
+	url := c.gateway + path
```
roosterfish: Worth using `path.Join` func? Otherwise I suspect `path` always needs to include a `/` prefix`?

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,812 @@
...
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("send request: %w", err)
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode > 299 {
```
roosterfish: Can you please add a comment what is special abour errors > 299? Maybe a link to some docs (if available) would also help for future readers/code modification.

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,812 @@
...
+	if authCookie == nil {
+		return nil, errors.New("starting PowerStore session: missing PowerStore authorization cookie")
+	}
+
+	csrf := resp.Header.Get(powerStoreCSRFHeaderName)
+	if csrf == "" {
+		return nil, errors.New("starting PowerStore session: missing PowerStore CSRF token")
+	}
+
+	return newPowerStoreLoginSession(info[0].ID, time.Duration(info[0].IdleTimeout)*time.Second, authCookie.Value, csrf), nil
```
roosterfish: Just a nit, when you do `sessionInfo := info[0]` you can reuse this var, maybe it looks nicer.

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,812 @@
...
+			return ls, nil // session was already replaced with a new valid session
+		}
+
+		return c.startNewLoginSession(ctx)
+	})
+}
+
+func (c *powerStoreClient) forceLoginSessionRemoval(sessionKey string, sessionToRemove *powerStoreLoginSession) {
+	//nolint:errcheck // Replace returns error only if inner callback returns error.
+	powerStoreTokenCache.Replace(sessionKey, func(ls *powerStoreLoginSession) (*powerStoreLoginSession, error) {
```
roosterfish: If we can safely ignore the error you can do `_ = powerStoreTokenCache.Replace()` to get rid of the `nolint`.

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,812 @@
...
+
+	session, err := c.getOrCreateLoginSession(ctx, sessionKey)
+	if err != nil {
+		return nil, err
+	}
+
+	requestEditors = append([]func(*http.Request) error{c.withLoginSession(session)}, requestEditors...)
+	resp, err := c.doHTTPRequest(ctx, method, path, requestData, responseData, requestEditors...)
+	if resp != nil && resp.StatusCode == http.StatusUnauthorized {
+		// there is something wrong with the session token, remove it
```
roosterfish: To do the same as in Pure/PowerFlex, can we here just reauthenticate if we see 401? This would omit the idle tracking I believe.

roosterfish: Currently what happens when the server deletes our session but the idle timer still thinks the session is alive?

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,812 @@
...
+		c.forceLoginSessionRemoval(sessionKey, session)
+	}
+
+	return resp, err
+}
+
+func (c *powerStoreClient) withBasicAuthorization(username, password string) func(req *http.Request) error {
+	return func(req *http.Request) error {
+		token := base64.StdEncoding.EncodeToString(fmt.Appendf(nil, "%s:%s", username, password))
```
roosterfish: There is another func further up in this file which concatenates the gateway with username and password I believe but it doesn't url encode the user pass. Why is this done here, why is it different?

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,812 @@
...
+	IdleTimeout              int64    `json:"idle_timeout"`
+	IsPasswordChangeRequired bool     `json:"is_password_change_required"`
+	IsBuiltInUser            bool     `json:"is_built_in_user"`
+}
+
+func (c *powerStoreClient) getLoginSessionInfoWithBasicAuthorization(ctx context.Context) (*http.Response, []*powerStoreLoginSessionResource, error) {
+	body := []*powerStoreLoginSessionResource{}
+	resp, err := c.doHTTPRequest(ctx, http.MethodGet, "/api/rest/login_session", nil, &body,
+		c.withBasicAuthorization(c.username, c.password),
+		c.withQueryParams(url.Values{"select": []string{"id,user,role_ids,idle_timeout,is_password_change_required,is_built_in_user"}}),
```
roosterfish: Do I interpret this right, for every request you have to indicate which values you want to be returned by the API? Or are these addition values not included in the regular API response that you still want to be returned by the server?

roosterfish: What is returned if you don't set those?

roosterfish: I wonder if we can get the JSON struct tags from the respective types to generate those strings.

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,812 @@
...
+	ID               string                                 `json:"id,omitempty"`
+	Name             string                                 `json:"name,omitempty"`
+	Description      string                                 `json:"description,omitempty"`
+	Initiators       []*powerStoreHostInitiatorResource     `json:"initiators,omitempty"`
+	OsType           string                                 `json:"os_type,omitempty"`
+	HostConnectivity string                                 `json:"host_connectivity,omitempty"`
+	MappedHosts      []*powerStoreHostVolumeMappingResource `json:"mapped_hosts,omitempty"`
+}
+
+//nolint:revive // For easier readability allow underscores in constants names.
```
roosterfish: I don't think that is something we usually do in the code base.

lxd/storage/drivers/driver_powerstore.go
```diff
@@ -0,0 +1,283 @@
...
+		//  shortdesc: ID of the PowerStore storage pool
+		//  scope: global
+		"powerstore.pool": validate.IsAny,
+		// lxdmeta:generate(entities=storage-powerstore; group=pool-conf; key=powerstore.domain)
+		// This option is required only if {config:option}`storage-powerstore-pool-conf:powerstore.pool` is specified using its name.
+		// ---
+		//  type: string
+		//  shortdesc: Name of the PowerStore protection domain
+		//  scope: global
+		"powerstore.domain": validate.IsAny,
```
MusicDin: Why is domain required when storage pool is specified via its name?
What are the other *options* to specify the storage pool in which case the domain is not required?

lxd/storage/drivers/driver_powerstore_volumes.go
```diff
@@ -0,0 +1,238 @@
...
+		//
+		// ---
+		//  type: string
+		//  condition: block-based volume with content type `filesystem`
+		//  defaultdesc: same as `volume.block.mount_options`
+		//  shortdesc: Mount options for block-backed file system volumes
+		//  scope: global
+		"block.mount_options": validate.IsAny,
+		// lxdmeta:generate(entities=storage-powerstore; group=volume-conf; key=size)
+		// The size must be in multiples of 1 MiB. The minimum size is 1 MiB and maximum is 256 TB.
```
MusicDin: Maximum size is mentioned, but is it verified anywhere?

lxd/storage/drivers/driver_powerstore_volumes.go
```diff
@@ -0,0 +1,238 @@
...
+// getVolumeName returns the fully qualified name derived from the volume.
+func (d *powerstore) getVolumeName(vol Volume) (string, error) {
+	volUUID, err := uuid.Parse(vol.config["volatile.uuid"])
+	if err != nil {
+		return "", fmt.Errorf(`failed parsing "volatile.uuid" from volume %q: %w`, vol.name, err)
+	}
+	volName := base64.StdEncoding.EncodeToString(volUUID[:])
+
+	// Search for the volume type prefix, and if found, prepend it to the volume name.
+	if prefix := powerStoreVolTypePrefixes[vol.volType]; prefix != "" {
```
MusicDin: A small nit, but we tend to avoid oneliners in if statements for readability.

lxd/storage/drivers/driver_powerstore_volumes.go
```diff
@@ -0,0 +1,952 @@
...
+		volName = prefix + powerStoreVolPrefixSep + volName
+	}
+
+	// Search for the content type suffix, and if found, append it to the volume name.
+	suffix := powerStoreVolContentTypeSuffixes[vol.contentType]
+	if suffix != "" {
+		volName = volName + powerStoreVolSuffixSep + suffix
+	}
+
+	poolHash := sha256.Sum256([]byte(vol.Pool()))
```
roosterfish: You can use the `volumeResourceNamePrefix` func from above which already contains those lines.

lxd/storage/drivers/driver_powerstore_volumes.go
```diff
@@ -0,0 +1,238 @@
...
+	if volNameWithoutSuffix, suffix, ok := strings.Cut(volName, powerStoreVolSuffixSep); ok {
+		volName = volNameWithoutSuffix
+		volContentType = powerStoreVolContentTypeSuffixesRev[suffix]
+	}
+
+	binUUID, err := base64.StdEncoding.DecodeString(volName)
+	if err != nil {
+		return poolHash, volType, volUUID, volContentType, fmt.Errorf("failed to decode volume name %q: %w", name, err)
+	}
+	volUUID, err = uuid.FromBytes(binUUID)
```
MusicDin: Again small nit, but we add empty line after logical blocks. Linter will probably catch that.

lxd/storage/drivers/driver_powerstore_volumes.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	if volNameWithoutSuffix, suffix, ok := strings.Cut(volName, powerStoreVolSuffixSep); ok {
+		volName = volNameWithoutSuffix
+		volContentType = powerStoreVolContentTypeSuffixesRev[suffix]
+	}
+
+	binUUID, err := base64.StdEncoding.DecodeString(volName)
+	if err != nil {
+		return poolHash, volType, volUUID, volContentType, fmt.Errorf("failed to decode volume name %q: %w", name, err)
```
MusicDin: Capitalized error

lxd/storage/drivers/driver_powerstore_volumes.go
```diff
@@ -0,0 +1,238 @@
...
+	poolHash := sha256.Sum256([]byte(vol.Pool()))
+	poolName := base64.StdEncoding.EncodeToString(poolHash[:])
+	return poolName + powerStorePoolAndVolSep + volName, nil
+}
+
+// getDataFromVolumeName decodes the volume's name and extracts stored data.
+func (d *powerstore) getDataFromVolumeName(name string) (poolHash string, volType VolumeType, volUUID uuid.UUID, volContentType ContentType, err error) {
+	poolHash, volName, ok := strings.Cut(name, powerStorePoolAndVolSep)
+	if !ok || poolHash == "" || volName == "" {
+		return "", "", uuid.Nil, "", fmt.Errorf("failed to decode volume name %q: invalid name format", name)
```
MusicDin: Capitalize errors

lxd/storage/drivers/driver_powerstore_volumes.go
```diff
@@ -0,0 +1,952 @@
...
+
+// hostResourceName derives the name of a host resource in PowerStore associated with the current node or host and mode. On success, it returns name applicable to use as PowerStore host resource name along of full unencoded name.
+func (d *powerstore) hostResourceName() (resource string, hostname string, err error) {
+	hostname, err = ResolveServerName(d.state.ServerName)
+	if err != nil {
+		return "", "", err
+	}
+
+	resource = fmt.Sprintf("%s%s-%s/%s", powerStoreResourceNamePrefix, hostname, d.config["powerstore.mode"], d.config["powerstore.transport"])
+	if len(resource) > 128 { // PowerStore limits host resource name to 128 characters
```
roosterfish: Can we not always us the same approach (hah) regardless whether or not `resource` is > 128?

lxd/storage/drivers/driver_powerstore_volumes.go
```diff
@@ -0,0 +1,952 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	return volResource != nil, nil
+}
+
+// ListVolumes returns a list of LXD volumes in storage pool.
+// It returns all volumes and sets the volume's volatile.uuid extracted from the name.
+func (d *powerstore) ListVolumes() ([]Volume, error) {
```
roosterfish, line 432: Have you tried running `lxd recover`?

lxd/storage/drivers/driver_powerstore_volumes_test.go
```diff
@@ -0,0 +1,246 @@
...
+	t.Parallel()
+	tests := []struct {
+		name    string
+		given   Volume
+		want    string
+		wantErr bool
+	}{
+		{
+			name: "container-volume",
+			given: NewVolume(
```
MusicDin: `NewVolume` instantiates `Volume`, but it would be easier to read test cases if `Volume{}` struct with named fields was used.

lxd/storage/drivers/driver_powerstore_volumes.go
```diff
@@ -0,0 +1,952 @@
...
+}
+
+// getVolumeResourceByVolume retrieves volume resource associated with the provided volume.
+func (d *powerstore) getVolumeResourceByVolume(vol Volume) (*powerStoreVolumeResource, error) {
+	volResourceName, err := d.volumeResourceName(vol)
+	if err != nil {
+		return nil, err
+	}
+
+	return d.client().GetVolumeByName(context.Background(), volResourceName)
```
roosterfish: You can use `d.state.shutdownCtx` instead. There are a few more in those file (maybe also others).

lxd/storage/drivers/driver_powerstore_volumes.go
```diff
@@ -0,0 +1,952 @@
...
+		typ += ":" + string(vol.volType)
+	}
+
+	if vol.contentType != "" {
+		typ += ":" + string(vol.contentType)
+	}
+
+	volResource := &powerStoreVolumeResource{
+		Name:         volResourceName,
+		Description:  powerStoreSprintfLimit(128, "LXD Name: %s", vol.name), // maximum allowed value length for volume description field is 128
```
roosterfish: The fact that we cut the desc looks odd to me. 

Not sure if it's better to omit this field completely if we cannot always accommodate what we want.

lxd/storage/drivers/token_cache.go
```diff
@@ -0,0 +1,121 @@
...
+import (
+	"sync"
+)
+
+// keyedMutex represents per-key lock (keyed mutex). It provides mutual exclusion locks scoped to arbitrary strings keys.
+//
+// Each distinct key behaves like Mutex from the "sync" package - subsequent calls to Lock with the given key will block until Unlock with the same key is called. Calls with different keys will proceed concurrently.
+//
+// Internally, it automatically allocates and deallocates mutexes as needed, keeping memory usage bounded to the number of concurrently hold locks.
+type keyedMutex struct {
```
roosterfish: Why do we have to perform this effort to lock the token compared to the other drivers?

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -1,13 +1,357 @@
...
+		if ls != session && ls.IsValid() {
+			return ls, nil // session was already replaced with a new valid session
+		}
+		return c.startNewLoginSession(ctx)
+	})
+}
+
+func (c *powerStoreClient) forceLoginSessionRemoval(sessionKey string, sessionToRemove *powerStoreLoginSession) {
+	//nolint:errcheck // Replace returns error only if inner callback returns error.
+	powerStoreTokenCache.Replace(sessionKey, func(ls *powerStoreLoginSession) (*powerStoreLoginSession, error) {
```
MusicDin: `_ = powerStoreTokenCache...` instead of `//nolint:errcheck`

roosterfish (changes requested):
Hi @adanishal, please have a look at my comments.

Quick summary:
1) I would be very interested to know what is already implemented/tested in this PR, meaning aside from creating vols and probably launching instances as I haven't seen any of the other vol funcs for snapshot/restore, migration, copy etc.
2) There seems to be a lot of logic around concurrent token handling so my question would be why is this necessary for PowerStore but not for the other drivers?
3) How the API client deals with expired tokens/session looks different to what we do in the other drivers
4) Have you had a chance to already validate if you can run `lxd recover` to rescue instances/vols? Just to ensure we don't introduce limits in the naming scheme right from the beginning.

Thanks.

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -1,13 +1,357 @@
...
+		ID:          id,
+		IdleTimeout: idleTimeout - 30*time.Second, // subtract to add safety margin for a potential time skew
+		AuthToken:   authToken,
+		CSRFToken:   csrfToken,
+	}
+	ls.Interacted()
+	return ls
+}
+
+// IsValid inform is the token associated with the login session is not expired.
```
MusicDin: Typo `is` -> `if`

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -1,13 +1,357 @@
...
+func (e *powerStoreError) ErrorCode() string {
+	for _, em := range e.details.Messages {
+		if em != nil && em.Code != "" {
+			return em.Code
+		}
+	}
+	return ""
+}
+
+// HTTPStatusCode attempts to extract the HTTP status code value from a PowerStore response.
```
MusicDin: attempts to extract? or simply returns?

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -1,13 +1,357 @@
...
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: c.gatewaySkipTLSVerify,
+			},
+		},
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("send request: %w", err)
```
MusicDin: Errors should be capitalized

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,812 @@
...
+	}
+
+	return body, nil
+}
+
+// GetApplianceMetrics retrieves appliance metrics.
+func (c *powerStoreClient) GetApplianceMetrics(ctx context.Context) ([]*powerStoreApplianceMetricsResource, error) {
+	var metrics []*powerStoreApplianceMetricsResource
+	for page := 0; ; page++ {
+		metricsPage, err := c.getApplianceMetricsByQuery(ctx, nil, powerStorePagination{Page: page})
```
MusicDin: If my interpretation is correct you are fetching all pages from powerstore? Can we fetch everything at once to avoid unnecessary calls?

lxd/storage/drivers/driver_powerstore_volumes.go
```diff
@@ -0,0 +1,952 @@
...
+	}
+
+	revert.Add(func() { _ = d.deleteVolumeResource(volResource) })
+
+	hostResource, err := d.getOrCreateHostWithInitiatorResource()
+	if err != nil {
+		return err
+	}
+
+	revert.Add(func() { _ = d.deleteHostAndInitiatorResource(hostResource) })
```
MusicDin: Could this be moved into utils?

In general we tried to keep `_volume` file for `powerflex/pure/alletra` clean from `connector/driver` specifics. The createVolumeResource should already take care of handling host and the initiator

lxd/storage/drivers/token_cache.go
tomponline: Suggest using `lxd/locking` package for this so we only have 1 per-key locking mutex package.

tomponline:
@adanishal please can you push what you have so far, thanks.

lxd/storage/drivers/driver_powerstore.go
adanishal: Good point. I introduced shared dedicated type to handle all of that `driverModesAndTransports`. With it all info is derived from a single list that serves as a source of truth.

lxd/storage/drivers/driver_powerstore.go
adanishal: Done.

lxd/storage/drivers/driver_powerstore.go
adanishal: PowerStore will support that, but not in phase 1. During phase 2 this flag will be flipped to true.

lxd/storage/drivers/driver_powerstore.go
adanishal: Good idea, but due to the fact that different drivers (in theory) may support differnt combinations of modes and transport, making this code shared is not straight forward. However with introduction of shared type `driverModesAndTransports` and its slice `driverModesAndTransports` I have manged to extract this logic to so drives can share it.

lxd/storage/drivers/driver_powerstore.go
adanishal: Good point. I have removed it.

lxd/storage/drivers/driver_powerstore.go
adanishal: Yea, it is left over. Removed.

lxd/storage/drivers/driver_powerstore.go
adanishal: There is no specific name, but since other drivers use `target` PowerStore driver does it as well.

lxd/storage/drivers/driver_powerstore.go
adanishal: Good point. I will change this (and other places with similar code as well).

lxd/storage/drivers/driver_powerstore.go
adanishal: Yea. Although those types are directly derived from definitions in swagger, so I would prefer to keep them in sync to avoid confusion (ex. why we try to parse response into different types than defined in the spec).

lxd/storage/drivers/driver_powerstore_utils.go
adanishal: In PowerStore we have additional constraint - we want to use sessions and reuse existing session if it is not expired. Idle timeout avoids unnecessary calls to the API.

There is a check if the session was removed on the server (server responds with 401 HTTP response code), that will also invalidate the session on the client side. So this case is handled. However server should not invalidate sessions at all - this would indicate some issue with the API.

Moreover there is another issue. With parallel requests it is possible that we do multiple requests before we realize the session needs refreshing.

lxd/storage/drivers/driver_powerstore_utils.go
adanishal: This is done so regardless what data user enters as URL, username and password we will have a token cache key uniquely identifying the recipient of the session.

lxd/storage/drivers/driver_powerstore_utils.go
adanishal: Yea. Typo... :)

lxd/storage/drivers/driver_powerstore_utils.go
adanishal: Yes, and passing the entire diver have one issue - it creates a circular dependency (`driver -> client -> driver`) that is much harder to reason about. With this design client is completely independent from the driver for the price of a few extra fields.

lxd/storage/drivers/driver_powerstore_utils.go
adanishal: I tried and it handles the prefix with scheme (`http:` or `https:`) poorly.

lxd/storage/drivers/driver_powerstore_utils.go
adanishal: Those are HTTP response codes (not some PowerStore internal error code).

lxd/storage/drivers/driver_powerstore_utils.go
adanishal: Good point. I will add this.

lxd/storage/drivers/driver_powerstore_utils.go
adanishal: Done.

lxd/storage/drivers/driver_powerstore_utils.go
adanishal: With tracking we avoid unnecessary calls to the API. If the server deletes our session there is a check for that that will invalidate session on the client side as well.

lxd/storage/drivers/driver_powerstore_utils.go
adanishal: This function is to derive a token cache key (it uses URL as well as username and password) this code handle basic authentication (done via HTTP header).

lxd/storage/drivers/driver_powerstore_utils.go
adanishal: Yes we could derive them from the struct tags, but it would require reflection and will add a lot complexity. Moreover it would only help for `select` parameter and PowerStore has an entire query language implemented using those parameters - for all other values we would need to use strings anyway.

lxd/storage/drivers/driver_powerstore_utils.go
adanishal: Ok. I will rename those.

lxd/storage/drivers/driver_powerstore_volumes.go
adanishal: Good point.

lxd/storage/drivers/driver_powerstore_volumes.go
adanishal: We could, but hashed name is less readable for admin. Moreover other drivers use straight name as well. The idea is to mimic this behavior (use name the same way as other drivers do) and only switch to different naming scheme if there is a problem (name is too long).

lxd/storage/drivers/driver_powerstore_volumes.go
adanishal: Yes, I thing it has the same results and constraints as with PowerFlex (since we use similar naming scheme).

lxd/storage/drivers/driver_powerstore_volumes.go
adanishal: I will do this for other as well.

lxd/storage/drivers/driver_powerstore_volumes.go
adanishal: This description is not used everywhere. It was added to help admins figure out what is going on since we use have volume name mangling.

lxd/storage/drivers/token_cache.go
adanishal: With PowerStore we wan to use sessions and only request a new one if the previous expired.

I have moved the implementation to use `lxd/locking` package.

lxd/storage/drivers/driver_powerstore.go
adanishal: Opps, that is leftover from other driver code. I will remove it.

lxd/storage/drivers/driver_powerstore_volumes.go
adanishal: I have added verification.

lxd/storage/drivers/driver_powerstore_volumes.go
adanishal: I will refactor it.

lxd/storage/drivers/driver_powerstore_volumes_test.go
adanishal: Good point I will change it.

lxd/storage/drivers/driver_powerstore_utils.go
adanishal: Well depending on the nature of issue HTTP code may be there or not.

lxd/storage/drivers/driver_powerstore_utils.go
adanishal: Sadly PowerSotore have a hard limit on the number of entrees returned in one call. We try to return as many as possible, but if there is more we need to iterate over pages.

However if there is less entrees that the PowerStore limit. The function returns after a single API call.

lxd/storage/drivers/driver_powerstore_volumes.go
adanishal: Good point. I will move it.

tomponline:
@adanishal is the PR description of progress updated to reflect the latest PR state?

tomponline (changes requested):
Please can you add an API extension commit and make it the first one in the PR. Thanks

lxd/storage/connectors/connector_nvme.go
```diff
@@ -279,7 +382,10 @@ func (c *connectorNVMe) Discover(ctx context.Context, targetAddresses ...string)
 			return nil, fmt.Errorf("Failed unmarshaling the returned discovery log entries from %q: %w", targetAddr, err)
 		}
 
-		// Unmarshalling the response from the discovery succeeded, break the loop.
+		nvmeAddMissingServiceIDToDiscoveryLog(&discoveryLog, targetAddr)
```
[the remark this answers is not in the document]
adanishal: This suggestion do not uses the proper logic. Expected behavior is: when discovery port was explicitly set, use it. Otherwise use default transport port.

When discovery is passed without the port we always add default discovery port (note that this port is different from default transport port). So the logic wouldn't work if we follow the above suggestions.

lxd/metadata/configuration.json
```diff
@@ -7064,6 +7064,234 @@
...
+							"longdesc": "Enabling this option allows attaching the volume to multiple isolated instances.",
+							"scope": "global",
+							"shortdesc": "Enable ID shifting overlay",
+							"type": "bool"
+						}
+					},
+					{
+						"security.unmapped": {
+							"condition": "custom volume",
+							"defaultdesc": "same as `volume.security.unmappped` or `false`",
```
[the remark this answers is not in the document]
adanishal: This is a real type but not introduced by this PR.

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,1526 @@
...
+func (ls *powerStoreLoginSession) Interacted() {
+	now := time.Now()
+	ls.LastInteraction.Store(&now)
+}
+
+func makePowerStoreLoginSessionKey(url, username, password string) string {
+	url = base64.StdEncoding.EncodeToString([]byte(url))
+	username = base64.StdEncoding.EncodeToString([]byte(username))
+	password = base64.StdEncoding.EncodeToString([]byte(password))
+	return fmt.Sprintf("%s:%s:%s", url, username, password)
```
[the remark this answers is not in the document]
adanishal: Not strictly nececary, but good to have.

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,1526 @@
...
+	targets, err := d.targets()
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: Inject failure, by forcing use of a single qualified name for all targets.
+	for i := range targets {
+		targets[i].QualifiedName = targets[0].QualifiedName
+	}
+
```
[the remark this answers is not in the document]
adanishal: Leftover from testing. Should be removed.

lxd/storage/drivers/driver_powerstore_utils.go
```diff
@@ -0,0 +1,1532 @@
...
+	url = base64.StdEncoding.EncodeToString([]byte(url))
+
+	// Hash username to enhance security, preventing anyone from retrieving it from
+	// the session key.
+	usernameHash := sha256.Sum256([]byte(username))
+	username = base64.StdEncoding.EncodeToString(usernameHash[:])
+
+	// Hash password to enhance security, preventing anyone from retrieving it from
+	// the session key.
+	passwordHash := sha256.Sum256([]byte(password))
```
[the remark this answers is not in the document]
adanishal: This is not a real issue. This hash is used as session key. The only reason we hash the password is to make sure it will not end up in some error log message or memory dump.

MusicDin:
Could you please land the connector changes as separate PRs because changes in this package affect several storage drivers and need to be tested against them.

Could you also try applying review fixes in the early commits (where the identified issues were discovered)? Otherwise, it is hard to review if subsequent commits keep changing the logic, and the reviews are marked as resolved while the issue is still present at that place.

Thanks

MusicDin:
Addressed in https://github.com/canonical/lxd/pull/18172

PR #17286, opened 2025-12-19
Use `%q` format where appropriate

simondeziel:
Replace occurrences of `'%s'` in strings by `%q`.

Also updated the MAAS controller to report about the right name of the missing subnet.

### DIFF
--- a/lxd/endpoints/local_test.go
+++ b/lxd/endpoints/local_test.go
@@ -70,7 +70,7 @@ func TestEndpoints_LocalUnknownUnixGroup(t *testing.T) {
 	err := endpoints.Up(config)
 
 	assert.EqualError(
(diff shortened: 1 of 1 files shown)


lxd/endpoints/local_test.go
```diff
@@ -70,7 +70,7 @@ func TestEndpoints_LocalUnknownUnixGroup(t *testing.T) {
 	err := endpoints.Up(config)
 
 	assert.EqualError(
-		t, err, "local endpoint: cannot get group ID of 'xquibaz': group: unknown group xquibaz")
+		t, err, "local endpoint: cannot get group ID of \"xquibaz\": group: unknown group xquibaz")
```
tomponline, line 73: For future reference, you can use backticks for the outer string to avoid the need for escaping double quotes.

simondeziel: Indeed, thanks for the tip.

Issue #17265, opened 2025-12-16
Improve error message when `limits.memory` is set far too low


MggMuggins:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

I typoed the value for `limits.memory` this morning (`limits.memory=4` == 4 bytes):
```
$ lxc launch images:ubuntu/questing/desktop --vm q1 -c limits.cpu=8 -c limits.memory=4 --target=sandusky
Launching q1
Error: Failed instance creation: Failed to run: forklimits limit=memlock:unlimited:unlimited fd=3 fd=4 fd=5 -- /snap/lxd/37049/bin/qemu-system-x86_64 -S -name q1 -uuid 720c132e-54bb-4fd1-9367-43247b6c8432 -daemonize -cpu host,hv_passthrough -nographic -serial chardev:console -nodefaults -no-user-config -sandbox on,obsolete=deny,elevateprivileges=allow,spawn=allow,resourcecontrol=deny -readconfig /var/snap/lxd/common/lxd/logs/rhythmbox-00424018_q1/qemu.conf -spice unix=on,disable-ticketing=on,addr=/var/snap/lxd/common/lxd/logs/rhythmbox-00424018_q1/qemu.spice -pidfile /var/snap/lxd/common/lxd/logs/rhythmbox-00424018_q1/qemu.pid -D /var/snap/lxd/common/lxd/logs/rhythmbox-00424018_q1/qemu.log -smbios type=2,manufacturer=Canonical Ltd.,product=LXD -runas lxd: : exit status 1
$ lxc info --show-log q1
Name: q1
Status: STOPPED
Type: virtual-machine
Architecture: x86_64
Location: sandusky
Created: 2025/12/16 08:20 CST

Log:

qemu-system-x86_64: property 'size' of memory-backend-memfd doesn't take value '0'
```

### Describe the solution you'd like

That error message isn't great UX. We should either:
- Set a minimum for `limits.memory` (does QEMU already enforce one?)
- Implement a warning for values of `limits.memory` < `100MiB`, either in `lxc` or in the daemon.

simondeziel:
> Implement a warning for values of `limits.memory` < `100MiB`, either in `lxc` or in the daemon.

To catch missing units, I think we could impose a much smaller limit would still work. I got a `--emtpy` VM with CSM mode boot to PXE with as little 2MiB. EFI mode is way heavier though :/

[...]

PR #17321, opened 2025-12-30
instance/drivers: Improve error message for low limits.memory

copyleftdev:
Prevent users from setting extremely low memory limits (e.g. < 50MiB) which are likely typos (MB vs GB). Returns a helpful error message suggesting the correct unit.

## Checklist

- [ x] I have read the [contributing guidelines](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md) and attest that all commits in this PR are [signed off](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#including-a-signed-off-by-line-in-your-commits), [cryptographically signed](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#commit-signature-verification), and follow this project's [commit structure](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#commit-structure).
- [ x] I have checked and added or updated relevant documentation.

### DIFF
--- a/lxd/instance/instancetype/instance.go
+++ b/lxd/instance/instancetype/instance.go
@@ -301,8 +301,8 @@ var InstanceConfigKeysAny = map[string]func(value string) error{
 			return err
 		}
 
-		if num == 0 {
-			return errors.New("Memory limit can't be 0")
+		if num < 1024*1024 {
+			return errors.New("Memory limit is too low (minimum 1MiB)")
 		}
 
 		return nil


lxd/instance/drivers/util.go
```diff
@@ -30,3 +31,12 @@ func parseMemoryStr(memory string) (valueInt int64, err error) {
 
 	return valueInt, err
 }
+
+// validateMemoryLimit checks if the memory limit is too low (less than 50MiB).
+func validateMemoryLimit(valueInt int64) error {
+	if valueInt < 50*1024*1024 {
+		return fmt.Errorf("memory limit is too low (minimum 50MiB). Did you mean GiB?")
```
tomponline: Errors should start with capital letter please

tomponline: I think we should drop the "Did you mean" bit here too, as we don't have the context of what the user actually asked for here.

lxd/instance/drivers/util.go
```diff
@@ -30,3 +31,12 @@ func parseMemoryStr(memory string) (valueInt int64, err error) {
 
 	return valueInt, err
 }
+
+// validateMemoryLimit checks if the memory limit is too low (less than 50MiB).
+func validateMemoryLimit(valueInt int64) error {
+	if valueInt < 50*1024*1024 {
```
MggMuggins: As discussed in #17265, I think we can probably solve this problem without introducing a practical limit on the minimum memory allocated to a VM: `1` or `2` MiB is probably sufficient. Did you have a different use case in mind than mine from #17265?

lxd/instance/drivers/driver_lxc.go
```diff
@@ -4263,6 +4268,11 @@ func (d *lxc) Update(args db.InstanceArgs, userRequested bool) error {
 					if err != nil {
 						return err
 					}
+
```
MggMuggins: I'm wondering if it makes more sense to include this check in the validation routine for `limits.memory`, since it doesn't require any contextual information about the instance: https://github.com/canonical/lxd/blob/main/lxd/instance/instancetype/instance.go#L281

MggMuggins (changes requested):
Thanks for fixing this! Could you also include `Fixes #17265` in your commit message so that the fix is linked to the commit?

lxd/instance/instancetype/instance.go
```diff
@@ -305,6 +305,10 @@ var InstanceConfigKeysAny = map[string]func(value string) error{
 			return errors.New("Memory limit can't be 0")
 		}
 
+		if num < 1024*1024 {
```
tomponline, line 304: the stanza above can be removed

```go
	if num == 0 {
			return errors.New("Memory limit can't be 0")
		}
```

tomponline:
Does this fix a logged github issue?

PR #17330, opened 2026-01-05
Remove more `i18n`

simondeziel:


lxc/action.go
tomponline, line 1: If adding `` to the flag label is required, can we have a wrapper function to do it, so as to make clear why its needed?

E.g.

```go
cmd.Flags().StringVar(&c.flagConsole, "console", "", flagLabel("Immediately attach to the console"))
```

simondeziel: Since only `StringVar` need it, I went with `FormatStringFlagLabel` as the name of the helper.

Could you let me know what you think about this implementation? If it's OK for you, I'll squash the commits.

tomponline: Cool looks good thanks!

simondeziel: @tomponline done now. The CI problem is unrelated to the changes at hand.

tomponline (approved):
Thanks!

PR #17332, opened 2026-01-05
Fix container live migration check for backwards compatibility

kadinsayani:
In LXD 5.21, when doing a stateless migration/copy of a running container (which is allowed), the source sends `offerHeader.Criu = CRIUType_NONE` to indicate "the container is running, but we're not doing live migration". In LXD 6.6, the receiver was rejecting any migration where `offerHeader.Criu != nil`, even if it's set to NONE (stateless). This broke backward compatibility for `lxc copy --refresh` from 5.21 to 6.6 when the source container is running.

This PR relaxes the live migration check in `MigrateReceive()` to only examine the `args.Live` flag rather than also checking `offerHeader.Criu`.

Reported from: https://discourse.ubuntu.com/t/lxc-copy-refresh-from-5-21-lts-to-latest-stable-stopped-working-since-6-6-a89d075-2025-12-10/74537

### DIFF
--- a/test/suites/migration.sh
+++ b/test/suites/migration.sh
@@ -282,31 +282,22 @@ migration() {
   lxc_remote copy l1:c1 l2:c2
   lxc_remote copy l1:c1 l2:c2 --refresh
 
-  lxc_remote start l1:c1 l2:c2
+  lxc_remote start l1:c1
 
   # Make sure the testfile doesn't exist
   ! lxc file pull l1:c1 -- /root/testfile1 || false
   ! lxc file pull l2:c2 -- /root/testfile1 || false
 
-  #lxc_remote start l1:c1 l2:c2
-
-  # Containers may not be running when refreshing
-  ! lxc_remote copy l1:c1 l2:c2 --refresh || false
-
-  # Create test file in c1
+  # Create test file in c1 (source)
   echo test | lxc_remote file push - l1:c1/root/testfile1
 
-  lxc_remote stop -f l1:c1 l2:c2
-
-  # Refresh the container and validate the contents
+  # Refresh with running source is allowed (refresh is always stateless).
+  # Target must be stopped to receive the refresh.
   lxc_remote copy l1:c1 l2:c2 --refresh
-  lxc_remote start l2:c2
   [ "$(lxc_remote file pull l2:c2/root/testfile1 -)" = "test" ]
-  lxc_remote stop -f l2:c2
 
   # Change the files modification time by adding one nanosecond.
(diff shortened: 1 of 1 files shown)


lxd/instance/drivers/driver_lxc.go
```diff
@@ -5139,7 +5139,7 @@ func (d *lxc) MigrateReceive(args instance.MigrateReceiveArgs) error {
 		return fmt.Errorf("Failed receiving migration offer from source: %w", err)
 	}
 
-	if args.Live || offerHeader.Criu != nil {
```
tomponline, line 5142: should we also reject if offerHeader.Criu isn't type None or nil?

kadinsayani: Yes, fixed.

tomponline:
@kadinsayani the tests seem to be failing because the target instance is running - this isn't allowed for refreshes.

test/suites/migration.sh
```diff
@@ -288,14 +288,16 @@ migration() {
...
-
-  # Containers may not be running when refreshing
-  ! lxc_remote copy l1:c1 l2:c2 --refresh || false
-
   # Create test file in c1
   echo test | lxc_remote file push - l1:c1/root/testfile1
 
+  # Refresh with running source is allowed (refresh is always stateless).
+  # Target must be stopped to receive the refresh.
+  lxc_remote stop -f l2:c2
```
simondeziel: How about not starting `l2:c2` in line 285 then?

kadinsayani: We start the instance to make sure the test file doesn't exist first.

simondeziel: Shouldn't be needed for `lxc file pull` as it also works on stopped containers.

test/suites/migration.sh
```diff
@@ -282,27 +282,22 @@ migration() {
...
-
-  # Refresh the container and validate the contents
+  # Refresh with running source is allowed (refresh is always stateless).
+  # Target must be stopped to receive the refresh.
   lxc_remote copy l1:c1 l2:c2 --refresh
   lxc_remote start l2:c2
   [ "$(lxc_remote file pull l2:c2/root/testfile1 -)" = "test" ]
-  lxc_remote stop -f l2:c2
+
+  lxc_remote stop -f l1:c1 l2:c2
```
simondeziel: Since pulling from a stopped container works, should `l2:c2` even be started?

kadinsayani: We can skip starting `l2:c2`.

tomponline (approved):
ta

Issue #12828, opened 2024-02-07
virtfs-proxy-helper for 9p filesystem shares deprecated in QEMU 8.1


tomponline:
> The 9p proxy filesystem backend driver has been deprecated and will be removed (along with its proxy helper daemon) in a future version of QEMU. Please use -fsdev local or -virtfs local for using the 9p local filesystem backend, or alternatively consider deploying virtiofsd instead.

> The 9p proxy backend was originally developed as an alternative to the 9p local backend. The idea was to enhance security by dispatching actual low level filesystem operations from 9p server (QEMU process) over to a separate process (the virtfs-proxy-helper binary). However this alternative never gained momentum. The proxy backend is much slower than the local backend, hasn’t seen any development in years, and showed to be less secure, especially due to the fact that its helper daemon must be run as root, whereas with the local backend QEMU is typically run as unprivileged user and allows to tighten behaviour by mapping permissions et al by using its ‘mapped’ security model option.

> Nowadays it would make sense to reimplement the proxy backend by using QEMU’s vhost feature, which would eliminate the high latency costs under which the 9p proxy backend currently suffers. However as of to date nobody has indicated plans for such kind of reimplementation unfortunately.

https://www.qemu.org/docs/master/about/deprecated.html#fsdev-proxy-and-virtfs-proxy-since-8-1

LXD uses virtfs-proxy-helper to isolate 9p filesystem shares to a user namespace to allow custom UID/GID mapping.
This is required for the LXD multi-user mode to function safely.

LXD already supports exporting filesystem volumes to VMs using virtiofs, and uses virtiofsd in a user namespace too.
However 9p filesystem shares are the only type of filesystem share that can be live migrated (unlike virtiofs shares).
Additionally some guest OSes do not support virtiofs and only support 9p.

We will need to explore whether we can use the 9p local backend and potentially have to start QEMU itself in a user namespace, although that will almost certainly cause other issues we will need to try and workaround.

Other possibilities include dropping 9p filesystem share support entirely and use only virtiofs, but this would then mean guest OSes without virtiofs support wouldn't be able to access the agent config drive to start the LXD agent - so this would need some alternative solution (and ISO drive export perhaps?), and it would also mean that all filesystem volume shares would be disabled for guests that could be live migratable.

tomponline:
Seeing these warnings in LXD VMs now:

```
[...]

PR #17335, opened 2026-01-06
lxd: Remove support for virtfs-proxy-helper

tomponline:
For our forthcoming switch to QEMU 10 in the 26.04 release we need to drop the requirement for the `virtfs-proxy-helper` program as this is no longer provided by QEMU. 

This program was used to export filesystems to VM guests using the `9p` protocol, whilst allowing custom id mapping rules via `raw.idmap` by way of running the program in a user namespace).

Instead now we will:

1. Continue to export the `config` drive (used by `lxd-agent`) via both `9p` and `virtiofs` protocols. Using QEMU's built in 9p server for the read-only config drive (as we do today). This maintains guest support for starting the `lxd-agent` program inside the guest.
2. Drop support for exporting custom filesystem disks to VMs using the `9p` protocol, and only support using `virtiofs` protocol. This means that guests will require virtiofs drivers in order to access filesystem disks, including cephfs volumes.

Additional cleanup:

1. Switched to `deviceConfig.DevSourcePath` for passing the virtiofsd socket path from the `disk` device to the `qemu` driver now that we don't need to pass both the virtofsd socket path and the virtfs-proxy-helper file descriptor.
2. Updates tests to only check for virtiofs custom filesystem disks whilst maintaining check for readonly config drive via 9p and virtiofs.

Fixes https://github.com/canonical/lxd/issues/12828

### DIFF
--- a/test/suites/vm.sh
+++ b/test/suites/vm.sh
@@ -95,45 +95,42 @@ test_vm_pcie_bus() {
   [ "$(lxc config get v1 volatile.aaa0.bus)" = "5" ]
   [ "$(lxc config get v1 volatile.root.bus)" = "6" ]
 
-  # Check hotplugging block volume.
-  poolDriver=$(lxc storage show "${pool}" | awk '/^driver:/ {print $2}')
+  # Check hotplugging volume using PCIe based virtio-blk.
+  lxc storage volume create "${pool}" v1block --type=block size=1MiB
+  lxc config device add v1 v1block disk source=v1block pool="${pool}"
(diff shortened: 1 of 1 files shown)


tomponline:
I'll update the failing test from lxd-ci after this is merged.

skozina:
This fixes #12828

simondeziel:
> I'll update the failing test from lxd-ci after this is merged.

I took a stab at it: https://github.com/canonical/lxd-ci/pull/648

PR #17336, opened 2026-01-06
Storage: ZFS promotion

tomponline:
Introduces the `zfs.promote` volume setting which is intended to be used with initial `disk` device config during instance copy/refresh, e.g.

`lxc copy c1 c2 -d root,initial.zfs.promote=true`

It allows for specifying that the target instance's ZFS dataset should be promoted using `zfs promote`.

Depends on https://github.com/canonical/lxd/pull/17930

### DIFF
--- a/test/suites/storage_volume_initial_config.sh
+++ b/test/suites/storage_volume_initial_config.sh
@@ -130,6 +130,84 @@ test_storage_volume_initial_config() {
     [ "$(lxc storage volume get "${pool}" container/c zfs.blocksize)" = "512KiB" ]
     [ "$(zfs get volblocksize "${pool}/containers/c" -H -o value)" = "128K" ]
     lxc delete c
+    lxc storage unset "${pool}" volume.zfs.block_mode
+
+    sub_test "Verify initial.zfs.promote functionality."
+    lxc launch "${image}" c --no-profiles --storage "${pool}"
+
+    # > Check that container's origin is the base image snapshot.
+    [ "$(zfs get origin "${pool}/containers/c" -H -o value)" = "${pool}/images/$(lxc config get c volatile.base_image)@readonly" ]
+
+    # > Create a clone of the container as a backup for restore later. Ensure a new volatile.uuid is generated.
+    lxc copy c c-backup --refresh -c "volatile.uuid=$(uuidgen)" -d root,initial.zfs.promote=true
+    [ "$(lxc config get c volatile.uuid)" != "$(lxc config get c-backup volatile.uuid)" ]
+
+    # > Check that the c container has a ZFS origin of c-backup copy snapshot prefix.
+    zfs get origin "${pool}/containers/c" -H -o value | grep "^${pool}/containers/c-backup@copy-"
+
+    # > Create a snapshot of the container to check that after promotion is not allowed with snapshots present.
+    lxc snapshot c snap0
+    # > Check that a container with snapshots cannot be promoted.
+    [ "$("${_LXC}" copy c c-save1 --refresh -c "volatile.uuid=$(uuidgen)" -d root,initial.zfs.promote=true 2>&1 1>/dev/null || false)" = 'Error: Create instance from copy: Cannot promote volume when source volume has snapshots or is a snapshot' ]
+    lxc delete c/snap0
+
+    # > Add some data to the container and create a ZFS promoted saved point with a new volatile.uuid.
+    lxc exec c -- touch /root/testfile1
+    lxc copy c c-save1 --refresh -c "volatile.uuid=$(uuidgen)" -d root,initial.zfs.promote=true
+    [ "$(lxc config get c volatile.uuid)" != "$(lxc config get c-save1 volatile.uuid)" ]
+
+    # > Check that the c container has a ZFS origin of c-save1 copy snapshot prefix.
+    zfs get origin "${pool}/containers/c" -H -o value | grep "^${pool}/containers/c-save1@copy-"
+
+    # > Add some more data to the container and create a 2nd ZFS promoted saved point with a new volatile.uuid.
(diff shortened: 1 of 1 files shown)


lxd/storage/drivers/driver_zfs_volumes.go
```diff
@@ -598,15 +586,38 @@
...
+	}
+
+	if vol.contentType == ContentTypeFS {
+		// Create mountpoint.
+		err := vol.EnsureMountPath()
 		if err != nil {
 			return err
 		}
+
+		revert.Add(func() { _ = os.Remove(vol.MountPath()) })
```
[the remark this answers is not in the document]
tomponline: False positive.

tomponline:
@kadinsayani this isn't quite polished yet (i still need to, at least, add zfs promote handling for migrations), but would appreciate your thoughts on it.

lxd/storage/utils.go
```diff
@@ -256,6 +256,19 @@ func VolumeDBCreate(pool Pool, projectName string, volumeName string, volumeDesc
 		return err
 	}
 
+	// Special zfs.promote handling.
+	// We don't want to store this in the database as it's a one-time operation.
+	// So record its value, remove it from the config, and then restore it after DB insertion.
+	// This avoids future copies of the volume inheriting the zfs.promote setting.
+	volConfig := vol.Config()
+	zfsPromote, zfsPromoteFound := volConfig["zfs.promote"]
```
kadinsayani, line 264:
```suggestion
	zfsPromote, found := volConfig["zfs.promote"]
```

Just a nit, but `found` works fine here since it's declared alongside `zfsPromote` so what is found is implicit, and it's not used anywhere further down.

test/suites/storage_volume_initial_config.sh
```diff
@@ -130,6 +130,83 @@ test_storage_volume_initial_config() {
     [ "$(lxc storage volume get "${pool}" container/c zfs.blocksize)" = "512KiB" ]
     [ "$(zfs get volblocksize "${pool}/containers/c" -H -o value)" = "128K" ]
     lxc delete c
+    lxc storage unset "${pool}" volume.zfs.block_mode
+
+    # > Verify initial.zfs.promote functionality.
```
kadinsayani: Perhaps a good place to use `sub_test`? The rest can remain comments.


```suggestion
    sub_test "Verify initial.zfs.promote functionality."
```

test/suites/storage_volume_initial_config.sh
```diff
@@ -130,6 +130,83 @@ test_storage_volume_initial_config() {
...
+    # > Create a clone of the container for as a backup for restore later. Ensure a new volatile.uuid is generated.
+    lxc copy c c-backup --refresh -c "volatile.uuid=$(uuidgen)" -d root,initial.zfs.promote=true
+    [ "$(lxc config get c volatile.uuid)" != "$(lxc config get c-backup volatile.uuid)" ]
+
+    # > Check that the c container has a ZFS origin of c-backup copy snapshot prefix.
+    zfs get origin "${pool}/containers/c" -H -o value | grep "^${pool}/containers/c-backup@copy-"
+
+    # > Create a snapshot of the container to check that after promotion is not allowed with snapshots present.
+    lxc snapshot c snap0
+    ! lxc copy c c-save1 --refresh -c volatile.uuid= -d root,initial.zfs.promote=true || false
```
kadinsayani: Can we assert the actual error message here?

test/suites/storage_volume_initial_config.sh
```diff
@@ -130,6 +130,83 @@ test_storage_volume_initial_config() {
...
+
+    # > Now restore from backup.
+    lxc copy c-backup c --refresh -c "volatile.uuid=${cUUID}" -d root,initial.zfs.promote=true
+    lxc delete c-backup
+
+    # > Check that container's origin is the base image snapshot.
+    [ "$(zfs get origin "${pool}/containers/c" -H -o value)" = "${pool}/images/$(lxc config get c volatile.base_image)@readonly" ]
+
+    # > Cleanup created containers.
+    lxc delete c c-save1 c-save2 c-save3
```
kadinsayani:
```suggestion
    lxc delete -f c c-save1 c-save2 c-save3
```

Since the containers are started.

lxd/storage/utils.go
tomponline: I've started preferring longer form indication vars, rather than `found` or `ok` as I find when you're reading a lot of code, the various `found` and `ok` lose their meanings.

lxd/instance/drivers/driver_lxc.go
```diff
@@ -4420,6 +4420,15 @@ func (d *lxc) Update(args db.InstanceArgs, userRequested bool) error {
...
 		}
 
+		// Don't store initial. config in database.
+		for devName, devConfig := range devices {
+			for devKey := range devConfig.Config {
+				if strings.HasPrefix(devKey, deviceConfig.ConfigInitialPrefix) {
+					delete(devices[devName].Config, devKey)
+				}
+			}
+		}
```
MusicDin: This is used in several places now, so should we introduce something like `deviceConfig.RemoveInitialOptions(devices)`?

tomponline: good idea

tomponline:
@kadinsayani changes made

lxd/instance/drivers/driver_lxc.go
tomponline: https://github.com/canonical/lxd/pull/17930

lxd/storage/utils.go
kadinsayani: Ack, makes sense.

PR #17341, opened 2026-01-07
test/suites/container_local_cross_pool_handling: simplify test

simondeziel:
Use the default LXD instance and create a single additional pool with the same driver to do the cross-pool copies/moves.

This avoids the creation of multiple unused pools.

Switch to using `--empty` instance while at it.

Before:

```
Test                                |   ceph |    dir |    lvm |    zfs
:---------------------------------- | -----: | -----: | -----: | -----:
container_local_cross_pool_handling | 88.07s | 16.20s | 27.77s | 15.38s
TOTAL                               | 88.07s | 16.20s | 27.77s | 15.38s
```

After:

```
Test                                |   ceph |   dir |    lvm |   zfs
:---------------------------------- | -----: | ----: | -----: | ----:
container_local_cross_pool_handling | 60.06s | 2.91s | 11.08s | 4.25s
TOTAL                               | 60.06s | 2.91s | 11.08s | 4.25s
```

### DIFF
--- a/test/suites/container_local_cross_pool_handling.sh
+++ b/test/suites/container_local_cross_pool_handling.sh
@@ -1,123 +1,78 @@
 test_container_local_cross_pool_handling() {
-  local LXD_STORAGE_DIR lxd_backend
-  lxd_backend=$(storage_backend "$LXD_DIR")
-  LXD_STORAGE_DIR=$(mktemp -d -p "${TEST_DIR}" XXXXXXXXX)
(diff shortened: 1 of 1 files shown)


simondeziel:
> Perhaps also putting a file into the container using `lxc file push` and then testing it exists in the moved container using `lxc file pull` would offer a more thorough test that the actual move has been completed.

Excellent suggestion, thanks! https://github.com/canonical/lxd/pull/17358

PR #17343, opened 2026-01-08
Add  CDI mode support for AMD GPUs

dmitry-lyfar:
## Checklist

### DIFF
--- a/lxd/main_callhook.go
+++ b/lxd/main_callhook.go
@@ -79,7 +79,7 @@ var customCDILinkerConfFile = "00-lxdcdi.conf"
 func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) error {
 	hookFile, err := os.Open(filepath.Join(devicesRootFolder, hooksFilePath))
 	if err != nil {
-		return fmt.Errorf("Failed to open the CDI hooks file at %q: %w", hooksFilePath, err)
+		return fmt.Errorf("Failed opening the CDI hooks file at %q: %w", hooksFilePath, err)
 	}
 
 	defer hookFile.Close()
@@ -87,7 +87,7 @@ func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) er
 	hooks := &cdi.Hooks{}
 	err = json.NewDecoder(hookFile).Decode(hooks)
 	if err != nil {
-		return fmt.Errorf("Failed to decode the CDI hooks file at %q: %w\n", hooksFilePath, err)
+		return fmt.Errorf("Failed decoding the CDI hooks file at %q: %w\n", hooksFilePath, err)
 	}
 
 	fmt.Println("CDI Hooks file loaded:")
@@ -108,20 +108,20 @@ func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) er
 		// Resolve hook link from target
 		target, err := resolveTargetRelativeToLink(symlink.Link, symlink.Target)
 		if err != nil {
-			return fmt.Errorf("Failed to resolve a CDI symlink: %w\n", err)
+			return fmt.Errorf("Failed resolving a CDI symlink: %w\n", err)
 		}
 
 		// Try to create the directory if it doesn't exist
 		err = os.MkdirAll(filepath.Dir(filepath.Join(containerRootFSMount, symlink.Link)), 0755)
 		if err != nil {
-			return fmt.Errorf("Failed to create the directory for the CDI symlink: %w\n", err)
+			return fmt.Errorf("Failed creating the directory for the CDI symlink: %w\n", err)
 		}
 
 		// Create the symlink
 		err = os.Symlink(target, filepath.Join(containerRootFSMount, symlink.Link))
 		if err != nil {
 			if !errors.Is(err, fs.ErrExist) {
-				return fmt.Errorf("Failed to create the CDI symlink: %w\n", err)
+				return fmt.Errorf("Failed creating the CDI symlink: %w\n", err)
 			}
 
 			fmt.Printf("Symlink not created because link %q already exists for target %q\n", symlink.Link, target)
@@ -134,7 +134,7 @@ func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) er
 		ldConfDirPath := filepath.Join(containerRootFSMount, "etc", "ld.so.conf.d")
 		err = os.MkdirAll(ldConfDirPath, 0755)
 		if err != nil {
-			return fmt.Errorf("Failed to create the linker conf directory at %q: %w\n", ldConfDirPath, err)
+			return fmt.Errorf("Failed creating the linker conf directory at %q: %w\n", ldConfDirPath, err)
 		}
 
 		ldConfFilePath := containerRootFSMount + "/etc/ld.so.conf.d/" + customCDILinkerConfFile
@@ -144,7 +144,7 @@ func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) er
 			// and add the ones that are not already there.
 			ldConfFile, err := os.OpenFile(ldConfFilePath, os.O_APPEND|os.O_RDWR, 0644)
 			if err != nil {
-				return fmt.Errorf("Failed to open the ld.so.conf file at %q: %w\n", ldConfFilePath, err)
+				return fmt.Errorf("Failed opening the ld.so.conf file at %q: %w\n", ldConfFilePath, err)
 			}
 
 			existingLinkerEntries := make(map[string]bool)
@@ -160,7 +160,7 @@ func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) er
 					_, err = fmt.Fprintln(ldConfFile, update)
 					if err != nil {
 						ldConfFile.Close()
-						return fmt.Errorf("Failed to write to the linker conf file at %q: %w\n", ldConfFilePath, err)
+						return fmt.Errorf("Failed writing to the linker conf file at %q: %w\n", ldConfFilePath, err)
 					}
 
 					existingLinkerEntries[update] = true
@@ -172,7 +172,7 @@ func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) er
 			// The file does not exist. We simply create it with our entries.
 			ldConfFile, err := os.OpenFile(ldConfFilePath, os.O_CREATE|os.O_WRONLY, 0644)
 			if err != nil {
-				return fmt.Errorf("Failed to create the linker conf file at %q: %w\n", ldConfFilePath, err)
+				return fmt.Errorf("Failed creating the linker conf file at %q: %w\n", ldConfFilePath, err)
 			}
 
 			for _, update := range hooks.LDCacheUpdates {
@@ -180,7 +180,7 @@ func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) er
 				_, err = fmt.Fprintln(ldConfFile, update)
 				if err != nil {
 					ldConfFile.Close()
-					return fmt.Errorf("Failed to write to the linker conf file at %q: %w\n", ldConfFilePath, err)
+					return fmt.Errorf("Failed writing to the linker conf file at %q: %w\n", ldConfFilePath, err)
 				}
 			}
 
@@ -194,7 +194,7 @@ func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) er
 		err = os.Remove(linkerCachePath)
 		if err != nil {
 			if !os.IsNotExist(err) {
-				return fmt.Errorf("Failed to remove the ld.so.cache file: %w\n", err)
+				return fmt.Errorf("Failed removing the ld.so.cache file: %w\n", err)
 			}
 
 			fmt.Printf("Linker cache not found in %q, skipping removal\n", linkerCachePath)
@@ -204,7 +204,7 @@ func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) er
 		ldexec := exec.Command("/sbin/ldconfig", "-r", containerRootFSMount)
 		output, err := ldexec.CombinedOutput()
 		if err != nil {
-			return fmt.Errorf("Failed to run ldconfig in the container rootfs: %w: %s", err, string(output))
+			return fmt.Errorf("Failed running ldconfig in the container rootfs: %w: %s", err, string(output))
 		}
 	}
 


lxd/device/cdi/configure.go
```diff
@@ -238,52 +248,56 @@ func specHookToLXDCDIHook(hook *specs.Hook, hooks *Hooks, l logger.Logger) error
...
+// rather they apply to all devices of a given class/vendor. These edits must be
+// applied if one or multiple devices are getting passed through (e.g. /dev/nvidia-uvm)
+// 2. Process per-device configuration(s): we convert this information into a
+// map of devices (keyed by their path given in the spec, it mapped to a map of
+// device properties). We also collect mounts (but we do not process them yet)
+// and hooks.
+// 3. Process general container edits in the same fashion.
+// 4. Process all the mounts collected from the spec in order to turn
+// them into disk devices. This operations generate a side effect: it generates
+// a list of indirect symlinks (see `specMountToNativeDev`)
```
jonathan-conder:
```suggestion
// a list of indirect symlinks (see `specMountToLXDDev`)
```

lxd/device/cdi/configure.go
```diff
@@ -294,49 +308,31 @@ func GenerateFromCDI(s *state.State, inst instance.Instance, cdiID ID, l logger.
 	configDevices := &ConfigDevices{UnixCharDevs: make([]map[string]string, 0), BindMounts: make([]map[string]string, 0)}
 
 	// 2. Process the specific device configuration
-	lookedUpDevs := make(map[string]struct{})
 	for _, device := range spec.Devices {
-		if cdiID.Name == All {
```
jonathan-conder, line 299: The `All` constant is now unused

lxd/instance/drivers/driver_lxc.go
```diff
@@ -2394,6 +2394,12 @@ func (d *lxc) Start(stateful bool) error {
 			}
 		}
 
+		if lxcLog != "" {
```
tomponline: This looks unrelated to CDI changes and should be at least a separate commit, if not a separate PR. Thanks

lxd/main_callhook.go
```diff
@@ -131,7 +131,13 @@ func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) er
 	// Updating the linker cache
 	l := len(hooks.LDCacheUpdates)
 	if l > 0 {
-		ldConfFilePath := containerRootFSMount + "/etc/ld.so.conf.d/" + customCDILinkerConfFile
+		ldConfDirPath := filepath.Join(containerRootFSMount, "etc", "ld.so.conf.d")
+		err = os.MkdirAll(ldConfDirPath, 0755)
+		if err != nil {
+			return fmt.Errorf("Failed to create the linker conf directory at %q: %w\n", ldConfDirPath, err)
```
tomponline: We tend to use "Failed creating" style errors these days.

lxd/main_callhook.go
```diff
@@ -196,9 +202,15 @@ func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) er
 	}
 
 	// Run `ldconfig` on the HOST (but targeting the container rootFS) to reduce the risk of running untrusted code in the container.
-	err = exec.Command("/sbin/ldconfig", "-r", containerRootFSMount).Run()
-	if err != nil {
-		return fmt.Errorf("Failed to run ldconfig in the container rootfs: %w\n", err)
+	// Allow tests and other environments to skip this step by setting LXD_SKIP_LDCONFIG=1.
+	if os.Getenv("LXD_SKIP_LDCONFIG") != "1" {
+		ldexec := exec.Command("/sbin/ldconfig", "-r", containerRootFSMount)
+		output, err := ldexec.CombinedOutput()
```
tomponline, line 205: Nice catch, this looks like it fixed a potential race where LXD was not waiting for the command to finish.

This should be a separate commit/PR along with an additional commit documenting the `doc/environment.md` please.

lxd/main_callhook.go
```diff
@@ -196,9 +202,15 @@ func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) er
 	}
 
 	// Run `ldconfig` on the HOST (but targeting the container rootFS) to reduce the risk of running untrusted code in the container.
-	err = exec.Command("/sbin/ldconfig", "-r", containerRootFSMount).Run()
-	if err != nil {
-		return fmt.Errorf("Failed to run ldconfig in the container rootfs: %w\n", err)
+	// Allow tests and other environments to skip this step by setting LXD_SKIP_LDCONFIG=1.
+	if os.Getenv("LXD_SKIP_LDCONFIG") != "1" {
```
tomponline: Remind me why this is needed again?

Lets indicate this is specific to CDI, how about `LXD_GPU_CDI_SKIP_LDCONFIG`?

lxd/instance/drivers/driver_lxc.go
dmitry-lyfar: Part of the draft PR, removed now.

lxd/main_callhook.go
dmitry-lyfar: Part of the draft PR, removed now. I only moved the call to the `ldconfig` to be done only if there are new records.

lxd/main_callhook.go
dmitry-lyfar: Part of the draft PR, removed now.

lxd/main_callhook.go
dmitry-lyfar: I can change that but in `main_callhook.go` all other similar log messages use "Failed to", do you want to bring them all to the latest?

lxd/main_callhook.go
dmitry-lyfar: I think there is no race, I changed it to collect full stderr, as otherwise it reports an exit code.

dmitry-lyfar:
> Please can you split up your commits into logical changes, ideally separate ones when crossing package boundaries.

I have logically re-arranged the commits: changes to the CDI implementation, AMD vendor support and docs update. The first is fairly trivial to see now, please let me know if you prefer it in a separate branch.

lxd/main_callhook.go
tomponline: If you have time that would be great thank you.

lxd/device/cdi/configure.go
tomponline, line 1: For the `lxd/device/cdi: add core CDI improvements` commit, is it possible to split that into 3 commits, one for each of the logical changes you mentioned in the commit description, or are they tightly coupled?

The reason this is useful as it helps to see which code changes relate to the specific logical change, whereas when they are bundled together it is harder to reason about.

Thanks

tomponline: For `- uid/gid support for CDI device nodes in configure.go` - please could you explain what that is, does the CDI generator return specific ownership for files inside the guest?

dmitry-lyfar: Yes, the `uid` and `gid` are the optional fields in the CDI spec which are provided by the AMD runtime but aren't by the NVIDIA's as I can tell: https://github.com/cncf-tags/container-device-interface/blob/main/SPEC.md

Without a proper ownership, things like ROCm won't work for a regular user.

However, I've realised that it is going to be troublesome for the groups given differed GIDs for `video` and `render`. It seems that the spec is largely concerned with the privileged containers and does not account for shifting. I don't see LXD handling it either in both cases (CDI or legacy GPU devices), it only checks the format.

I'm guessing it is by design?

dmitry-lyfar: There is a recent addition to the spec: https://github.com/cncf-tags/container-device-interface/commit/eb0daa749172388a852c17c8a1e055823a427a7e which is not implemented by either of the vendors yet. But it will provide a list of GIDs that the container process should have available.

tomponline, line 1: Please can you split this into 3 commits for:

- uid/gid support for CDI device nodes in configure.go
- Fixed major/minor device number handling for devices with minor=0
- Call `ldconfig` only if cache has changed

tomponline: As some may be backportable and others cant, so logical changes are best separated where possible please.

tomponline: > It seems that the spec is largely concerned with the privileged containers and does not account for shifting. I don't see LXD handling it either in both cases (CDI or legacy GPU devices), it only checks the format.

Yes, this was a thought I had too, I will need to dig into this more deeply to understand how we should be doing UID shifting.

tomponline: Is it worth splitting this part out into a separate PR so it does not hold back the AMD support?

dmitry-lyfar: Sorry, I think I need to be a bit more clear here. The shifting itself is not an issue here, the  problem is that the host distro and the container's may have different ids for groups like `render` and `video`. That won't work for a debian running a fedora image, for example, as their `video` groupd IDs are different.

If I run debian on debian it works just fine, LXD handles shifting for the unpriv containers similarly to how it has been done for classic gpu devices.

tomponline: Ah yes ok understood. Yes not much we can do about different guest os group ids.

jonathan-conder: Probably not something to do right now, but I think we should use ACLs instead of group IDs if possible.

There's a udev rule `/usr/lib/udev/rules.d/73-seat-late.rules` which controls this on my 24.04 system. You can actually use it inside a LXD container if you manually launch a `logind` seat. But it feels wrong.

Instead we could create a separate rule like
```
TAG=="uaccess", ENV{MAJOR}!="", RUN+="lxd-agent uaccess"
```
The `lxd-agent uaccess` command (or something similar) would work like https://github.com/systemd/systemd/blob/main/src/udev/udev-builtin-uaccess.c, but query the seat on the host, for the idmapped user, instead of inside the container.

Obviously this would only work with modified container images, but it's at least distro-agnostic as long as they use `systemd`.

For static devices a different approach might be needed, similar to https://github.com/systemd/systemd/blob/v259/src/login/logind-seat.c#L330. But maybe we can get away with ignoring them.

dmitry-lyfar: I've made the split, I suppose it makes sense to have the AMD changes on top as they are stacking on the `generateSpec` signature changes anyway, so the AMD PR would depend on the generic changes.

lxd/device/cdi/configure.go
```diff
@@ -14,13 +14,11 @@ import (
...
 	"github.com/canonical/lxd/lxd/instance"
-	"github.com/canonical/lxd/lxd/state"
 	"github.com/canonical/lxd/shared"
-	"github.com/canonical/lxd/shared/logger"
 )
 
-// specDevToNativeDev builds a list of unix-char devices to be created from a CDI spec.
-func specDevToNativeDev(configDevices *ConfigDevices, d specs.DeviceNode) error {
+// specDevToLXDDev builds a list of unix-char devices to be created from a CDI spec.
+func specDevToLXDDev(configDevices *ConfigDevices, d specs.DeviceNode) error {
```
tomponline: We try nowadays to avoid prefixing everything with "LXD" as that is implicit.
How about `specDevToInstanceDev` to clarify its adding devices to an instance?

tomponline: The comment to this function could be clarified too to explain that it will modify the devices in the supplied `configDevices` argument.

tomponline: I'd also consider swapping the order of the arguments too, because the `specDevToInstanceDev` suggests that the `d` spec argument should come first due to the directional nature of the function name.

lxd/device/cdi/configure.go
```diff
@@ -134,9 +143,8 @@ func specMountToNativeDev(configDevices *ConfigDevices, cdiID ID, mounts []*spec
 
 // specHookToLXDCDIHook will translate a hook from a CDI spec into an entry in a `Hooks`.
 // Some CDI hooks are not relevant for LXD and will be ignored.
-func specHookToLXDCDIHook(hook *specs.Hook, hooks *Hooks, l logger.Logger) error {
+func specHookToLXDCDIHook(hook *specs.Hook, hooks *Hooks) error {
 	if hook == nil {
-		l.Warn("CDI hook is nil")
```
tomponline, line 139: This change is unrelated to commit message, please can it be in a separate commit.

lxd/device/cdi/configure.go
```diff
@@ -41,16 +39,27 @@ func specDevToNativeDev(configDevices *ConfigDevices, d specs.DeviceNode) error
...
-	return nil
-}
+	nativeDev := map[string]string{"type": "unix-char",
+		"source": hostPath,
+		"path":   d.Path,
+		"major":  strconv.FormatInt(d.Major, 10),
+		"minor":  strconv.FormatInt(d.Minor, 10),
+	}
 
-// specMountToNativeDev builds a list of disk mounts to be created from a CDI spec.
```
tomponline, line 48: The removal of the logging change is unrelated to commit message, please can it be in a separate commit.

test/main.sh
tomponline, line 1: This change is unrelated to commit message, please can it be in a separate commit.

lxd/device/cdi/spec.go
```diff
@@ -42,8 +42,8 @@ func defaultNvidiaTegraCSVFiles(rootPath string) []string {
 }
 
 // generateNvidiaSpec generates a CDI spec for an Nvidia vendor.
-func generateNvidiaSpec(s *state.State, cdiID ID, inst instance.Instance) (*specs.Spec, error) {
-	l := logger.AddContext(logger.Ctx{"instanceName": inst.Name(), "projectName": inst.Project().Name, "cdiID": cdiID.String()})
+func generateNvidiaSpec(isCore bool, cdiID ID, inst instance.Instance) (*specs.Spec, error) {
```
tomponline, line 46: This doesn't seem directly related to the uid/gid support, although it is a welcome change as it decouples `generateNvidiaSpec` from `state.State`. However can it be a separate commit so that logical changes (and their justifications) can be clearly recorded?

lxd/device/cdi/spec.go
```diff
@@ -122,6 +123,7 @@ func generateNvidiaSpec(s *state.State, cdiID ID, inst instance.Instance) (*spec
 		nvcdi.WithMode(mode),
 		nvcdi.WithCSVFiles(defaultNvidiaTegraCSVFiles(rootPath)),
 		nvcdi.WithConfigSearchPaths(configSearchPaths),
+		nvcdi.WithMergedDeviceOptions(transform.WithName("all")),
```
tomponline, line 126: What does this do? Is it related to the uid/gid support?

lxd/device/cdi/configure.go
```diff
@@ -346,8 +342,10 @@ func GenerateFromCDI(s *state.State, inst instance.Instance, cdiID ID, l logger.
 	return configDevices, hooks, nil
 }
 
-// ReloadConfigDevicesFromDisk reads the paths to the CDI configuration devices file from the disk.
-// This is useful in order to cache the CDI configuration devices file so that wee don't have to re-generate a CDI spec whhen stopping the container.
+// ReloadConfigDevicesFromDisk reads the paths to the CDI configuration devices
+// file from the disk. This is useful in order to cache the CDI configuration
+// devices file so that wee don't have to re-generate a CDI spec whhen stopping
```
tomponline: typo "wee".

Also, this seems like it should be a separate commit as its not related to uid/gid functionality directly.

lxd/device/cdi/configure.go
```diff
@@ -41,16 +39,27 @@ func specDevToNativeDev(configDevices *ConfigDevices, d specs.DeviceNode) error
...
+		"path":   d.Path,
+		"major":  strconv.FormatInt(d.Major, 10),
+		"minor":  strconv.FormatInt(d.Minor, 10),
+	}
 
-// specMountToNativeDev builds a list of disk mounts to be created from a CDI spec.
-func specMountToNativeDev(configDevices *ConfigDevices, cdiID ID, mounts []*specs.Mount) ([]SymlinkEntry, error) {
-	if len(mounts) == 0 {
-		return nil, errors.New("CDI mounts are empty")
+	if d.UID != nil {
```
tomponline, line 52: Are these d.UID and d.GID lines below the only part of this commit related to UID/GID support?

Are the rest refactors? I'm having trouble ascertaining the reason for each logical change, as it seems like there's been some (welcome) refactoring done as part of introducing the UID/GID support.

However to enable us to more clearly track the reason for each change, and to enable more to be backported, we like to see refactors done as discrete changes before the new functionality is added as a separate commit.

lxd/device/cdi/configure.go
```diff
@@ -28,7 +28,9 @@ func specDevToLXDDev(configDevices *ConfigDevices, d specs.DeviceNode) error {
 		hostPath = d.Path // When the hostPath is empty, the path is the device path in the container.
 	}
 
-	if d.Major == 0 || d.Minor == 0 {
+	// Only fetch major/minor from device if both are unspecified (0).
+	// Some devices legitimately have minor=0, so we can't treat 0 as "unspecified" individually.
```
tomponline, line 32: Great comment, thanks!

lxd/main_callhook.go
```diff
@@ -182,23 +188,24 @@ func applyCDIHooksToContainer(devicesRootFolder string, hooksFilePath string) er
...
+			fmt.Printf("Linker cache not found in %q, skipping removal\n", linkerCachePath)
+		}
 
-	// Run `ldconfig` on the HOST (but targeting the container rootFS) to reduce the risk of running untrusted code in the container.
-	err = exec.Command("/sbin/ldconfig", "-r", containerRootFSMount).Run()
-	if err != nil {
-		return fmt.Errorf("Failed to run ldconfig in the container rootfs: %w\n", err)
+		// Run `ldconfig` on the HOST (but targeting the container rootFS) to reduce the risk of running untrusted code in the container.
+		ldexec := exec.Command("/sbin/ldconfig", "-r", containerRootFSMount)
+		output, err := ldexec.CombinedOutput()
```
tomponline, line 205: This change should be a separate commit as it allows recording why we are switching to recording all output.

lxd/device/cdi/spec.go
```diff
@@ -158,11 +159,67 @@ func generateNvidiaSpec(isCore bool, cdiID ID, inst instance.Instance) (*specs.S
 	return spec, nil
 }
 
+func generateAmdSpec(isCore bool, cdiID ID, inst instance.Instance) (*specs.Spec, error) {
```
tomponline: should this be called "generateAMDSpec"?

lxd/device/cdi/spec.go
```diff
@@ -158,11 +159,67 @@ func generateNvidiaSpec(isCore bool, cdiID ID, inst instance.Instance) (*specs.S
...
+	l := logger.AddContext(logger.Ctx{"project": inst.Project().Name, "instance": inst.Name(), "cdiID": cdiID.String()})
+
+	// Determine which binary to use based on snap mode.
+	amdCTKBinary := "amd-ctk"
+	rootPath := ""
+	if shared.InSnap() {
```
tomponline: This section caught my eye, and then I looked at https://github.com/canonical/lxd-pkg-snap/pull/1021/changes for more context.

From what I can see if LXD is running inside the snap, then it will use the `amd-ctk-wrapper` to `chroot` to outside of the LXD snap's mount namespace (effectively) and then will run `/usr/bin/amd-ctk` command from the host's filesystem (rather than the LXD's snap's bundled version).

https://github.com/canonical/lxd-pkg-snap/pull/1021/changes#diff-81186418dde8aa802befc72f0528cc4d2eb63d7d53d5fbf46620d6e3b6740f6cR41

This doesn't seem desirable (for one thing it won't work on Ubuntu Core) and it also seems to render the bundling of the `amd-ctk` command here https://github.com/canonical/lxd-pkg-snap/pull/1021/changes#diff-e68b9afd67bb77dec8eda77796d3e0b784c7011203114996383eed6533be0cefR212 redundant.

Normally we name wrappers the same as the command they run in a non-snap environment, e.g. https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft/wrappers/nvidia-container-cli so that LXD can just execute the original command and not need to switch commands based on whether it is a snap or not.

For "escaping" the snap's mount namespace we also have https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft/wrappers/run-host which is used in various places https://github.com/canonical/lxd-pkg-snap/blob/6ac455c7ee42e9dba05a9cd1c67bee767984c207/snapcraft/commands/daemon.start#L191-L227 to symlink to in order to allow LXD to run commands using their original name but from the host's filesystem.

For the `generateNvidiaSpec` function, we can see it directly resolves the `nvidia-ctk` command bundled within the snap package, and then passes it to `nvcdi.New()` using `WithNVIDIACDIHookPath` then it passes the host's filesystem path using `WithDriverRoot` and `WithDevRoot`.

I'm not sure if you were trying to "escape" the snap's mount namespace in order to run `amd-ctk` from the host's filesystem, or whether you wanted to run `amd-ctk` bundled inside the snap, but ask it to scan the host's filesystem.

tomponline: Ideally we would use the same approach as the nvidia CDI integration, and run the `amd-ctk` directly from inside the snap, and then pass it an option to scan the host's filesystem if inside the snap?

lxd/device/cdi/spec.go
dmitry-lyfar: This is to include "all" keyword in the output of the spec generator and avoid filtering duplicating devices like [here](https://github.com/canonical/lxd/blob/02ad243673d1abc42257c676a11f9240800719aa/lxd/device/cdi/configure.go#L297). Both generators, AMD's and NVIDIA's, support it as per the standard, so you can get a list of all devices under a single key.

lxd/device/cdi/configure.go
dmitry-lyfar: Removed for now.

lxd/device/cdi/configure.go
dmitry-lyfar: Separated refactors out.

lxd/device/cdi/spec.go
dmitry-lyfar: Separated the refactor out.

test/main.sh
dmitry-lyfar: Removed.

lxd/device/cdi/configure.go
dmitry-lyfar: Separated out into the refactoring commit.

lxd/device/cdi/configure.go
dmitry-lyfar: Separated out in the refactoring commit.

lxd/device/cdi/configure.go
dmitry-lyfar: I only wanted to edit the name as "native" was more confusing, thanks, I renamed it to the instance*. The behavior of the function itself regarding the `configDevices` or signature is not changed. I would consider further change to actually not modify `configDevices` in place at all but return new devices instead. But just to keep the changes to the minimum, I left that one as is.

Separated this one in the refactoring commit.

lxd/device/cdi/spec.go
tomponline: Thanks. So was this not working before? I'm a bit confused what this enables/fixes?

lxd/device/cdi/spec.go
dmitry-lyfar: @tomponline I don't think that it will call `amd-ctk` from the host, the purpose of this wrapper is different:

1. If LXD finds itself in snap, it looks up for the wrapper name `amd-ctk-wrapper` instead of the binary itself. This is because `amd-ctk` has no way of pointing it to the filesystem's root unlike in the `nvidia-ctk` case.
2. The wrapper changes root for the spec generator similarly to how it is done for nvidia but using `chroot`.

BUT. I think there is no need in the wrapper for `amd-ctk` because it is mostly interested in `/dev` and `/sys` to include nodes into the spec. AMD's implementation does not involve bind mounting of some user space libraries OR loading any host-side so libs as in the Nvidia's case. Therefore, Nvidia's spec generator has to see the host's fs (see the branch that handles snap case for the Nvidia's spec) but AMD's is not. I have tested it without the wrapper by side-loading the binary and adding AMD GPU devices.

tomponline: Thanks! My understanding was that the wrapper script itself chrooted to `/var/lib/snapd/hostfs` and then ran `amd-ctk` itself, and when i was testing the `chroot` command behaviour, it ran the specified command after switching to the chroot, so my interpretation was that the amd-ctk binary shipped in the snap was never being called, but rather the one on the host filesystem.

But thats good that it doesn't need to scan the host fs and we can lose the wrapper, ive merged that PR now.

tomponline:
@dmitry-lyfar is this ready for a review yet?

lxd/device/cdi/spec.go
dmitry-lyfar: @tomponline this wasn't not being generated before and therefore the implementation of `GenerateFromCDI` used to go through the list of all cards and de-duplicated it because each card is included at least twice (by index and by UUID). We should rely on the standard keyword `all` for that which provides a full non-duplicated list (also, it makes fetching the specified devices much simpler).

tomponline: Right I see, sounds good. In that case do we need to also remove the old de-dupe logic you mention?

dmitry-lyfar: Yes, I removed it here: https://github.com/canonical/lxd/pull/17343/changes#diff-8ef907052df8c269f99fe050f869f582f2d5bfd730c77a2108def89a0bac46c5L299

We still need a list by index and by UUID as it can be provided as an input if one attaches a single device, so it stayed in the device options.

dmitry-lyfar: I'm having trouble with linking to the exact diff, it's part of the `GenerateFromCDI` changes.

tomponline: Understood, thanks!

lxd/device/cdi/spec.go
dmitry-lyfar, line 144: @tomponline btw, I did not touch this behaviour yet, but this file does not seem to be used elsewhere for anything. It will be kept in the container's log dir and never reloaded (unlike the LXD configuration associated with a given CDI device which does get reloaded, but that's a different file).

lxd/device/cdi/spec.go
tomponline: Have you considered the case where a user adds a LXD GPU device selecting "nvidia/all" for example, and then also adds a LXD GPU device for "nvidia/0". Should we allow this or prevent it?

I'm just thinking about if subsequently the "nvidia/all" device was removed, would it also remove the pass-through files for "nvidia/0".

lxd/device/cdi/spec.go
tomponline: Hrm, good spotting.


Do you mean it wont get removed if the instance or the LXD GPU device is removed?

Or do you mean it wont get re-generated on each LXD GPU device start up?

lxd/device/cdi/spec.go
dmitry-lyfar: Yes, I have tested it, there are a few issues like this. We should probably gradually cover them as we go with integration tests when we get rid of the old configuration options "nvidia.*" (e.g. I reported another issue where you can just add a GPU device with any random index and it won't report an error as it is not validated + some other findings). I changed mostly the parts I needed to add AMD's spec + start some unit testing.

tomponline: Yeah agreed, thanks for being on top of this. It doesn't need to block this PR being merged.

tomponline: > I'm just thinking about if subsequently the "nvidia/all" device was removed, would it also remove the pass-through files for "nvidia/0".

I'm thinking we should just block an "nvidia/all" from being present at the same time as an "nvidia/N" device?

dmitry-lyfar: To be clear `all` was supported in the `lxc config device add... id=all` before, it just did not use the keyword but filtered the spec manually. So, there is not behavioral change.

dmitry-lyfar:
> @dmitry-lyfar is this ready for a review yet?

Yes, I think so, I extracted my refactorings as a separate commit in that PR (the first one). Lmk if that looks good to you.

lxd/device/cdi/spec.go
tomponline: Got you, still, not for this PR, but as a follow up.

I'm focussing specifically on the CDI IDs, so looking at https://documentation.ubuntu.com/lxd/latest/reference/devices_gpu/#gpu-physical

We seem to allow `nvidia.com/igpu=all` and `nvidia.com/gpu=all`.

I wonder if we should continue to allow that, but prevent another LXD device from being added at the same time to an instance that conflicts with it, e.g. `nvidia.com/gpu=0`.

This would simplify the removal logic.

lxd/device/cdi/spec.go
dmitry-lyfar: It won't get removed if the GPU device is removed (but will with the instance removal afaik). It won't be regenerated on the device startup either as the config + devices will be used to have the device attached. It will be regenerated and saved separately if a device with another index/UUID is added.

Keeping this file around can potentially be used for logging/debugging purposes, otherwise, it is never  used.

lxd/device/cdi/spec.go
dmitry-lyfar: Yes, that's another issue, we do allow classes (`igpu` & `gpu`) but I don't think it is possible to filter devices by using nvidia's generator output (possibly, classes need to be requested when generating the spec). At least, the original implementation disregards classes and only accounts for names (which is either UUID or index).

Blocking an individual device from being added if `all` is there appears to be the best way, yes.

dmitry-lyfar: So, there won't be difference between `nvidia.com/igpu=0` and `nvidia.com/gpu=0` now as a result.

tomponline: What was the difference before your change? Shouldn't they be selecting different cards? I'm not familiar with how CDI uses ID and Class separation.

dmitry-lyfar: Sorry, I think the latter is not entirely correct. It hasn't changed after my PR, just observations. CDI has 3 components in the name: vendor, class and ID. The old and the new implementation both use the name to filter out devices that were requested. The spec's output does not include the class of a device in the current output, this is why I said it is not possible to differentiate. 

But our implementation does it differently. If `igpu` is requested, it sets [mode](https://github.com/canonical/lxd/pull/17343/changes#diff-6feacc6dc48a55821d147c7bf2a40193e9f3628d9998dc8e1aedab71c3d332a8L48) to `csv` as opposed to `auto`. Now I have no idea why it's this way and how csv affects the output, but the output of it does not include discrete GPUs on my machine.

dmitry-lyfar: Yes, I think that `nvidia's` generator should use `nvcdi.WithClass()` which it currently does not. Maybe we should move it to issues, I'm not sure that without this option `nvidia-ctk` would respect the class parameter.

tomponline: Right, would you mind opening an issue about it, I agree we should try and limit the scope of this PR to focus on AMD. Thanks!

lxd/device/cdi/configure.go
```diff
@@ -41,16 +39,19 @@ func specDevToNativeDev(configDevices *ConfigDevices, d specs.DeviceNode) error
 		d.Minor = int64(unix.Minor(uint64(stat.Rdev)))
 	}
 
-	configDevices.UnixCharDevs = append(configDevices.UnixCharDevs, map[string]string{"type": "unix-char", "source": hostPath, "path": d.Path, "major": strconv.FormatInt(d.Major, 10), "minor": strconv.FormatInt(d.Minor, 10)})
+	instanceDev := map[string]string{"type": "unix-char",
```
tomponline: just a nit, but id have moved the "type" field to its own line like the rest of the fields.

lxd/device/cdi/spec.go
```diff
@@ -12,10 +12,11 @@ import (
 	"strings"
 
 	"github.com/NVIDIA/nvidia-container-toolkit/pkg/nvcdi"
+	"github.com/NVIDIA/nvidia-container-toolkit/pkg/nvcdi/transform"
+	"github.com/stretchr/testify/assert/yaml"
```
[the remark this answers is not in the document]
tomponline: @dmitry-lyfar the rest of the code base uses `go.yaml.in/yaml/v2`  so Copilot makes a good point here that this looks out of place.

lxd/device/cdi/spec.go
```diff
@@ -157,11 +158,55 @@ func generateNvidiaSpec(isCore bool, cdiID ID, inst instance.Instance) (*specs.S
 	return spec, nil
 }
 
+func generateAMDSpec(isCore bool, cdiID ID, inst instance.Instance) (*specs.Spec, error) {
+	l := logger.AddContext(logger.Ctx{"project": inst.Project().Name, "instance": inst.Name(), "cdiID": cdiID.String()})
+
+	amdCTKBinary := "amd-ctk"
+	amdCTKPath, err := exec.LookPath(amdCTKBinary)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to find the %s binary: %w", amdCTKBinary, err)
```
tomponline: Please can you use `%q` for quoting strings in error messages. Thanks

lxd/device/cdi/spec.go
dmitry-lyfar: Moved here: https://github.com/canonical/lxd/issues/17587

dmitry-lyfar: Also, this: https://github.com/canonical/lxd/issues/17588

PR #17345, opened 2026-01-08
Fix lint errors after golangci-lint update

tugbataluy:
## Checklist

## Summary
This PR addresses lint errors identified by updated golangci-lint rules, specifically focusing on `prealloc` linters.

## Changes
- **Memory optimization**: Preallocated slices with proper capacity to reduce memory allocations (prealloc)

## Files affected
- Core LXD components: database, device drivers, networking, firewall
- Client CLI tools and shared libraries
- Test files

### DIFF
--- a/lxd/storage/drivers/driver_zfs_utils.go
+++ b/lxd/storage/drivers/driver_zfs_utils.go
@@ -68,7 +68,8 @@ func (d *zfs) dataset(vol Volume, deleted bool) string {
 }
 
 func (d *zfs) createDataset(dataset string, options ...string) error {
-	args := []string{"create"}
+	args := make([]string, 0, 1+2*len(options)+1)
+	args = append(args, "create")
 	for _, option := range options {
 		args = append(args, "-o", option)
 	}
@@ -84,7 +85,8 @@ func (d *zfs) createDataset(dataset string, options ...string) error {
 }
 
 func (d *zfs) createVolume(dataset string, size int64, options ...string) error {
-	args := []string{"create", "-s", "-V", strconv.FormatInt(size, 10)}
+	args := make([]string, 0, 4+2*len(options)+1)
+	args = append(args, "create", "-s", "-V", strconv.FormatInt(size, 10))
 	for _, option := range options {
 		args = append(args, "-o", option)
 	}
@@ -268,7 +270,8 @@ func (d *zfs) filterRedundantOptions(dataset string, options ...string) ([]strin
 }
 
 func (d *zfs) setDatasetProperties(dataset string, options ...string) error {
-	args := []string{"set"}
+	args := make([]string, 0, 1+len(options)+1)
+	args = append(args, "set")
 	args = append(args, options...)
 	args = append(args, dataset)
 
@@ -363,7 +366,8 @@ func (d *zfs) version() (string, error) {
 
 // initialDatasets returns the list of all expected datasets.
 func (d *zfs) initialDatasets() []string {
-	entries := []string{"deleted"}
+	entries := make([]string, 0, 1+len(d.Info().VolumeTypes)*2)
+	entries = append(entries, "deleted")
 
 	// Iterate over the listed supported volume types.
 	for _, volType := range d.Info().VolumeTypes {


lxd/device/nic_bridged.go
```diff
@@ -733,16 +734,17 @@ func (d *nicBridged) Update(oldDevices deviceConfig.Devices, isRunning bool) err
 		}
 
 		// Remove old host-side routes from bridge interface.
-
-		oldRoutes := []string{}
+		//nolint:prealloc
+		oldRoutes := make([]string, 0)
```
simondeziel: Why switch to make instead of just the `nolint` directive?

lxd/device/nic_bridged.go
```diff
@@ -552,7 +552,8 @@ func (d *nicBridged) Start() (*deviceConfig.RunConfig, error) {
 	}
 
 	// Apply host-side routes to bridge interface.
-	routes := []string{}
+	routes := make([]string, 0, len(d.config["ipv4.routes"])+len(d.config["ipv6.routes"])+len(d.config["ipv4.routes.external"])+len(d.config["ipv6.routes.external"]))
```
simondeziel: Why provide the capacity here but not in `Update` and `postStop`?

shared/version/useragent.go
```diff
@@ -29,7 +29,9 @@ func getUserAgent() string {
 		panic(err)
 	}
 
-	osTokens := []string{cases.Title(language.English).String(runtime.GOOS), arch}
+	osTokens := make([]string, 2, 2+len(getPlatformVersionStrings()))
```
simondeziel: Should we call `getPlatformVersionStrings()` once just to learn how big of a slice it returns? I don't think we'll recoup the overhead so I'd probably just do:
```
// nolint:prealloc
osTokens := make([]string, 2, 2)
...
```

lxd/device/nic_bridged.go
tugbataluy: Because I forgot to change it back, I tried to allocate some space but I couldn't determine the amount of space.

lxd/device/nic_bridged.go
tugbataluy: My first intention to allocate enough space (not much, not little), but sometimes I couldn't predict the needed amount in that case I just used //nolint:prealloc

lxd/db/storage_volumes.go
```diff
@@ -379,7 +379,12 @@ func (c *ClusterTx) GetLocalStoragePoolVolumeSnapshotsWithType(ctx context.Conte
     AND (storage_volumes.node_id=? OR storage_volumes.node_id IS NULL AND storage_pools.driver IN ` + query.Params(len(remoteDrivers)) + `)
   ORDER BY storage_volumes_snapshots.creation_date, storage_volumes_snapshots.id`
 
-	args := []any{poolID, volumeType, volumeName, projectName, c.nodeID}
+	args := make([]any, 5, 5+len(remoteDrivers))
```
tomponline: How about this form rather than having to specify "5" twice and every field with its only line.

```go
args := make([]any, 0, 5+len(remoteDrivers))
args = append(args, poolID, volumeType, volumeName, projectName, c.nodeID)
```

shared/version/useragent.go
tugbataluy: But the platform version strings' length may vary right?

simondeziel: > But the platform version strings' length may vary right?

Exactly but `append()` will deal with whatever reallocation is needed. Since the function returns a 0-length slice on non-Linux, I think it's best to silence the `prealloc` hint.

lxd/device/nic_bridged.go
tugbataluy: And also I couldn't estimate the needed space for the slices in Update and postStop

tomponline: I would suggest not going too complicated for this - the optimisation here is really not worth it.
If you don't know the size, then pre-allocate the minimal size you know you need or disable the linter for this one.

lxc/config/config.go
```diff
@@ -50,15 +50,17 @@ func (c *Config) GlobalConfigPath(paths ...string) string {
 		configDir = "/etc/lxd"
 	}
 
-	path := []string{configDir}
+	path := make([]string, 1, 1+len(paths))
```
tomponline: we should avoid all similar invocations that pre-allocate and initialise using duplicated values, and instead only pre-allocate and then append. Ta

lxd/db/storage_volumes.go
```diff
@@ -760,7 +761,11 @@ SELECT storage_volumes_snapshots.name FROM storage_volumes_snapshots
    AND storage_pools.name=?
    AND (storage_volumes.node_id=? OR storage_volumes.node_id IS NULL AND storage_pools.driver IN ` + query.Params(len(remoteDrivers)) + `)`
 
-	inargs := []any{typ, name, pool, c.nodeID}
+	inargs := make([]any, 4, 4+len(remoteDrivers))
```
tomponline: whoops, this one didnt get switched to append

lxd/db/schema/schema.go
```diff
@@ -204,7 +204,7 @@ func (s *Schema) Ensure(db *sql.DB) (int, error) {
 // It requires that all patches in this schema have been applied, otherwise an
 // error will be returned.
 func (s *Schema) Dump(db *sql.DB) (string, error) {
-	var statements []string
+	var statements = make([]string, 0, 1)
```
tomponline: this capacity of 1 seems unexpected, id be tempted to leave as it was and disable the linter check for this one.

tomponline: especially as the slice is replaced on line 209 anyway

lxd/device/nic_bridged.go
tomponline: does using `make([]string, 0)` appease the linter?

lxd/ip/vdpa.go
```diff
@@ -394,7 +394,7 @@ func AddVDPADevice(pciDevSlotName string, volatile map[string]string) (*VDPADev,
 
 // DeleteVDPADevice deletes a vDPA management device.
 func DeleteVDPADevice(vDPADevName string) error {
-	header := []*nl.RtAttr{}
+	header := make([]*nl.RtAttr, 0, 1)
```
tomponline: here we can just move the `header` intialisation down below `newNetlinkAttribute` so we can initialise the `header` directly (no need for define followed by `append`).

lxd/network/driver_ovn.go
```diff
@@ -353,7 +353,7 @@ func (n *ovn) validateExternalSubnet(uplinkRoutes []*net.IPNet, projectRestricte
 func (n *ovn) getExternalSubnetInUse(uplinkNetworkName string) ([]externalSubnetUsage, error) {
 	var err error
 	var projectNetworks map[string]map[int64]api.Network
-	var externalSubnets []externalSubnetUsage
+	var externalSubnets = make([]externalSubnetUsage, 0)
```
tomponline: we shouldnt initialise a variable with `make` that we are about to replace with a the result of another function call. So if the linter is complaining about this just disable the check for this point.

lxd/storage/drivers/driver_zfs_utils.go
```diff
@@ -68,7 +68,8 @@ func (d *zfs) dataset(vol Volume, deleted bool) string {
 }
 
 func (d *zfs) createDataset(dataset string, options ...string) error {
-	args := []string{"create"}
+	args := make([]string, 1, 1+2*len(options)+1)
```
tomponline: whoops, not using append here and still initialising instead.

lxd/storage/drivers/driver_zfs_utils.go
```diff
@@ -84,7 +85,11 @@ func (d *zfs) createDataset(dataset string, options ...string) error {
 }
 
 func (d *zfs) createVolume(dataset string, size int64, options ...string) error {
-	args := []string{"create", "-s", "-V", strconv.FormatInt(size, 10)}
+	args := make([]string, 4, 4+2*len(options)+1)
```
tomponline: whoops, not using append here and still initialising instead.

lxd/storage/drivers/driver_zfs_utils.go
```diff
@@ -268,7 +273,8 @@ func (d *zfs) filterRedundantOptions(dataset string, options ...string) ([]strin
 }
 
 func (d *zfs) setDatasetProperties(dataset string, options ...string) error {
-	args := []string{"set"}
+	args := make([]string, 1, 1+len(options)+1)
```
tomponline: whoops, not using append here and still initialising instead.

lxd/subprocess/bgpm_test.go
```diff
@@ -134,7 +134,7 @@ func TestStopRestart(t *testing.T) {
 }
 
 func TestProcessStartWaitExit(t *testing.T) {
-	var a []string
+	var a = make([]string, 0, 1)
```
tomponline: we can just define `a` with the value it needs on line 141 instead of append.

lxd/subprocess/bgpm_test.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 func TestSignalHandling(t *testing.T) {
-	var a []string
+	var a = make([]string, 0, 1)
```
tomponline: we can just initialise with the value need, no need to use append.

lxd/subprocess/bgpm_test.go
```diff
@@ -82,7 +82,7 @@ func TestSignalHandling(t *testing.T) {
 
 // tests newprocess, start, stop, save, import, restart, wait.
 func TestStopRestart(t *testing.T) {
-	var a []string
+	var a = make([]string, 0, 1)
```
tomponline: we can just initialise with the value need, no need to use append.

shared/version/useragent.go
```diff
@@ -29,6 +29,7 @@ func getUserAgent() string {
 		panic(err)
 	}
 
+	//nolint:prealloc
 	osTokens := []string{cases.Title(language.English).String(runtime.GOOS), arch}
```
tomponline:
```go
versionInfo := getPlatformVersionStrings()
osTokens := make([]string,0,len(2+len(versionInfo)))
osTokens = append(osTokens, cases.Title(language.English).String(runtime.GOOS), arch)
osTokens = append(osTokens, versionInfo)
```

would work right?

lxd/firewall/drivers/drivers_xtables.go
```diff
@@ -1236,7 +1236,12 @@ func (d Xtables) iptablesAdd(ipVersion uint, comment string, table string, metho
 		return fmt.Errorf("Asked to setup IPv%d firewalling but %s can't be found", ipVersion, cmd)
 	}
 
-	args := []string{"-w", "-t", table, method, chain}
+	args := make([]string, 5, 5+len(rule)+4)
```
tomponline: whoops, missed this one

shared/version/useragent.go
tugbataluy: Yes, it would.

lxd/storage/drivers/driver_zfs_utils.go
tugbataluy: but initialising gives this warning:
`lxd/storage/drivers/driver_zfs_utils.go:71:2: Consider preallocating args with capacity 1 + 2*len(options) + 1 (prealloc)
        args := []string{"create"}
        ^
lxd/storage/drivers/driver_zfs_utils.go:87:2: Consider preallocating args with capacity 4 + 2*len(options) + 1 (prealloc)
        args := []string{"create", "-s", "-V", strconv.FormatInt(size, 10)}
        ^
lxd/storage/drivers/driver_zfs_utils.go:271:2: Consider preallocating args with capacity 1 + len(options) + 1 (prealloc)
        args := []string{"set"}`

tugbataluy: I'll simply //nolint:prealloc if it suits you.

tomponline: This should work like the others:

```go
args := make([]string, 0, 1+2*len(options)+1)
args = append(args, "create")
...
```

tomponline: basically we dont want to see the 2nd argument to `make` being > 0.

tugbataluy: Ah you were mentioning the second argument sorry :sweat_smile:

lxd/db/schema/schema.go
```diff
@@ -204,7 +204,8 @@ func (s *Schema) Ensure(db *sql.DB) (int, error) {
 // It requires that all patches in this schema have been applied, otherwise an
 // error will be returned.
 func (s *Schema) Dump(db *sql.DB) (string, error) {
-	var statements []string
+	//nolint:prealloc
+	var statements = []string{}
```
tomponline: you dont need to initialise it here, so leave it as:

```
//nolint:prealloc
var statements []string
```

lxd/network/driver_ovn.go
tomponline: this comment hasn't been addressed @tugbataluy

lxd/storage/drivers/driver_zfs_utils.go
tomponline: this comment hasnt been addressed @tugbataluy

tomponline (changes requested):
@tugbataluy there's a few comments from last week that havent been addressed yet.

lxd/storage/drivers/driver_zfs_utils.go
tugbataluy: Currently it's looking like that:
``` 
func (d *zfs) createDataset(dataset string, options ...string) error {
	args := make([]string, 0, 1+2*len(options)+1)
	args = append(args, "create")
	for _, option := range options {
		args = append(args, "-o", option)
}
 ```
	
Could you explain the part that is not addressed?

Comment (tomponline) [reply lxd/storage/drivers/driver_zfs_utils.go]: did you not rebase so that the earlier commit was removed, as when i was going through the commits it still showed the prior version?

Comment (tugbataluy) [reply lxd/storage/drivers/driver_zfs_utils.go]: I believe i rebased that. Currently can you see the corrected version in the changes? I think I can.

Comment (tomponline) [reply lxd/storage/drivers/driver_zfs_utils.go]: this commit is showing the old version https://github.com/canonical/lxd/pull/17345/changes/37297b8d2951c511cc1bbec4635aafa37035fa06

Comment (tugbataluy) [reply lxd/storage/drivers/driver_zfs_utils.go]: Is that still the same?

Comment (tugbataluy) [reply lxd/storage/drivers/driver_zfs_utils.go]: I accidentally commit different changes in different commits with exact same comment. I deleted one of them, I think it's OK right now.

File: lxd/network/driver_ovn.go (hunk 9 above)
Comment (tugbataluy) [reply lxd/network/driver_ovn.go]: I did the same thing as in zfs_utils.go sorry

File: lxd/storage/drivers/driver_zfs_utils.go (hunk 10 above)
Comment (tugbataluy) [reply lxd/storage/drivers/driver_zfs_utils.go]: I guess I cleaned all of the mess :sweat_smile:

File: lxd/network/driver_ovn.go [hunk 19]
```diff
@@ -19,7 +19,7 @@ import (
 
 	"github.com/mdlayher/netx/eui64"
 
-	"github.com/canonical/lxd/client"
+	lxd "github.com/canonical/lxd/client"
```
tomponline: please remove the "lxd" bit.

lxd/network/driver_ovn.go
```diff
@@ -353,8 +353,8 @@ func (n *ovn) validateExternalSubnet(uplinkRoutes []*net.IPNet, projectRestricte
 func (n *ovn) getExternalSubnetInUse(uplinkNetworkName string) ([]externalSubnetUsage, error) {
 	var err error
 	var projectNetworks map[string]map[int64]api.Network
+	//nolint:prealloc
```
tomponline, line 356: do we actually need this when the code isn't appending to it, but simply replacing the variable? Does the linter still file in this case?

lxd/storage/drivers/driver_zfs_utils.go
```diff
@@ -363,6 +366,7 @@ func (d *zfs) version() (string, error) {
 
 // initialDatasets returns the list of all expected datasets.
 func (d *zfs) initialDatasets() []string {
+	//nolint:prealloc
 	entries := []string{"deleted"}
```
tomponline: this can be pre-allocated to `1+len(d.Info().VolumeTypes*2)` right?

lxd/network/driver_ovn.go
tugbataluy: Linter still asks for preallocation, that's why I add that line.

lxd/storage/drivers/driver_zfs_utils.go
tugbataluy: Yes, it can :D

lxd/network/driver_ovn.go
tugbataluy: Removed

simondeziel:
Thanks @tugbataluy!

PR #5706, opened 2019-04-26
Support direct `copy` within a cluster

stgraber:

PR #17347, opened 2026-01-08
Storage: Remove Ceph specific instance copy handling

roosterfish:
Contributes to https://github.com/canonical/lxd/issues/13542. We didn't yet tracked this one, I found it by more carefully searching for `"ceph"` checks, not only in the storage backend code.

This PR removes the check for the specific `ceph` storage driver, and instead uses the driver's `Remote` attribute. 
However @escabo raised a valid point what happens if the instance is still running. As PowerFlex does not support optimized volume copy, that means the volume would be mapped twice on the source machine and the target machine where the copy is made.
Therefore we always have to redirect the request to the source machine.

On top of that there currently is a bug which got introduced with https://github.com/canonical/lxd/pull/5706 for Ceph RBD. If an instance is running and a copy of the instance is made targeting another cluster member, the existing copy functionality cannot be used as its freeze capabilities don't work out across cluster members.

Tests around the copy after start bits are included in https://github.com/canonical/lxd/pull/17466 (should be merged first).

### DIFF
--- a/test/suites/container_copy.sh
+++ b/test/suites/container_copy.sh
@@ -148,6 +148,9 @@ test_container_copy_start() {
   # Spawn a second node.
   spawn_lxd_and_join_cluster "${cert}" 2 1 "${LXD_ONE_DIR}" "${poolDriver}"
 
+  # Spawn a third node.
+  spawn_lxd_and_join_cluster "${cert}" 3 1 "${LXD_ONE_DIR}"
+
   # Set up a TLS identity with admin permissions.
   LXD_DIR="${LXD_ONE_DIR}" lxc auth group create copy
   LXD_DIR="${LXD_ONE_DIR}" lxc auth group permission add copy server admin
@@ -229,6 +232,22 @@ test_container_copy_start() {
     [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc list -f csv -c s c2)" = "RUNNING" ]
 
     LXD_DIR="${LXD_ONE_DIR}" lxc delete -f c2
+
+    # Create an instance on the first cluster member.
+    # Then test if an instance copy triggered on another cluster gets redirected to the source cluster member.
+    # In addition force the placement on a third cluster member different than the source/trigger to also cause a move of the instance.
+    LXD_DIR="${LXD_ONE_DIR}" lxc copy c1 c2 --target node1
+
+    echo "==> Check the instance copy can be triggered from any other member than the source with a final placement on a different member"
+    LXD_DIR="${LXD_TWO_DIR}" lxc copy c2 c3 --target node3 --mode "${mode}" --start
+
+    echo "==> Check the instance was placed on node2"
+    LXD_DIR="${LXD_ONE_DIR}" lxc info c3 | grep -xF "Location: node3"
+
+    echo "==> Check the copied container is running"
+    [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc list -f csv -c s c3)" = "RUNNING" ]
+
+    LXD_DIR="${LXD_ONE_DIR}" lxc delete -f c2 c3
   done
 
   # Cleanup
@@ -241,13 +260,16 @@ test_container_copy_start() {
 
   LXD_DIR="${LXD_ONE_DIR}" lxd shutdown
   LXD_DIR="${LXD_TWO_DIR}" lxd shutdown
+  LXD_DIR="${LXD_THREE_DIR}" lxd shutdown
 
   rm -f "${LXD_ONE_DIR}/unix.socket"
   rm -f "${LXD_TWO_DIR}/unix.socket"
+  rm -f "${LXD_THREE_DIR}/unix.socket"
 
   teardown_clustering_netns
   teardown_clustering_bridge
 
   kill_lxd "${LXD_ONE_DIR}"
   kill_lxd "${LXD_TWO_DIR}"
+  kill_lxd "${LXD_THREE_DIR}"
 }


roosterfish:
Leaving it in draft until I have figured the impact of this change on the other drivers.

lxd/instances_post.go
```diff
@@ -571,19 +571,20 @@ func createFromCopy(ctx context.Context, s *state.State, projectName string, pro
...
-			err = s.DB.Cluster.Transaction(ctx, func(ctx context.Context, tx *db.ClusterTx) error {
-				_, pool, _, err = tx.GetStoragePoolInAnyState(ctx, sourcePoolName)
-
-				return err
-			})
+			pool, err := storagePools.LoadByName(s, sourcePoolName)
 			if err != nil {
-				err = fmt.Errorf("Failed to fetch instance's pool info: %w", err)
-				return response.SmartError(err)
+				return response.SmartError(fmt.Errorf("Failed to load instance's pool %q: %w", sourcePoolName, err))
```
tomponline: "Failed loading..." is the pattern we use for errors

lxd/storage/drivers/driver_types.go
```diff
@@ -35,6 +35,10 @@ type Info struct {
 	// Whether driver generates an optimised backup header file in backup.
 	OptimizedBackupHeader bool
 
+	// Whether driver supports optimized copy of volumes.
```
tomponline: /directly/internally/

lxd/instances_post.go
roosterfish: Fixed.

lxd/storage/drivers/driver_types.go
roosterfish: Thanks, fixed.

roosterfish: Removed this again due to the other bug with Ceph RBD. See the updated PR description. We can instead just check on the source if the instance is running as this is already sufficient.

roosterfish:
Updated the PR description/commits with the up to date problem statement and solution.

lxd/storage/drivers/driver_types.go
roosterfish: Removed from the PR.

lxd/instances_post.go
```diff
@@ -1385,6 +1433,27 @@ func instancesPost(d *Daemon, r *http.Request) response.Response {
...
+		client, err := cluster.Connect(r.Context(), sourceMemberInfo.Address, s.Endpoints.NetworkCert(), s.ServerCert(), false)
+		if err != nil {
+			return response.SmartError(err)
+		}
+
+		// Keep the intended target around for the final move operation.
+		client = client.UseProject(targetProjectName)
+		client = client.UseTarget(targetMemberInfo.Name)
+
+		logger.Debug("Forward instance post copy request", logger.Ctx{"local": s.ServerName, "target": sourceMemberInfo.Name, "targetAddress": sourceMemberInfo.Address})
```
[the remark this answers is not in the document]
roosterfish: Should be fine in this case as we want to log about forwarding the request to the `target` even though in this case it's the `source` of the copy operation.

lxd/instances_post.go
```diff
@@ -676,7 +680,36 @@ func createFromCopy(ctx context.Context, s *state.State, projectName string, pro
 			return err
 		}
 
-		return instanceCreateFinish(s, req, args)
+		var client lxd.InstanceServer
```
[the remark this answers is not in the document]
roosterfish: Fixed.

lxd/instances_post.go
tomponline, line 684: @roosterfish can `run` be defined below the newly added block to move its closer to where its used?

lxd/instances_post.go
```diff
@@ -697,7 +697,36 @@ func createFromCopy(ctx context.Context, s *state.State, projectName string, pro
...
+			}
+
+			// Move to the actual target.
+			targetClient = targetClient.UseTarget(targetMemberInfo.Name)
+
+			logger.Debug("Migrate instance to final target after copy", logger.Ctx{"local": s.ServerName, "target": targetMemberInfo.Name, "targetAddress": targetMemberInfo.Address})
+			op, err := targetClient.MigrateInstance(req.Name, api.InstancePost{
+				// We don't have to handle live migration as the instance is always stopped after copy.
+				// Also we want to always copy the entire instance with all of its snapshots in case they exist.
+				// The previous copy operation would have skipped snapshots when performing an instance only copy.
```
tomponline: what previous copy operation?

lxd/instances_post.go
roosterfish: The new block (moving the instance to the final target) is part of the `run := func() { ... }` so it happens all inside the same operation.

lxd/instances_post.go
roosterfish: A few lines before (in the same operation) we run `instanceCreateAsCopy` which performs the initial copy of the instance (with/without snaps if requested).

lxd/instances_post.go
tomponline: oh this is just GH not showing the diff properly aligned then ta

lxd/instances_post.go
```diff
@@ -1420,6 +1418,27 @@ func instancesPost(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	// Redirect the copy request to the cluster member which currently holds the instance.
```
tomponline, line 1463: Can we do this earlier in the function to avoid duplicated work on original and redirected nodes?

lxd/instances_post.go
```diff
@@ -697,7 +697,36 @@ func createFromCopy(ctx context.Context, s *state.State, projectName string, pro
...
+				return err
+			}
+
+			// Move to the actual target.
+			targetClient = targetClient.UseTarget(targetMemberInfo.Name)
+
+			logger.Debug("Migrate instance to final target after copy", logger.Ctx{"local": s.ServerName, "target": targetMemberInfo.Name, "targetAddress": targetMemberInfo.Address})
+			op, err := targetClient.MigrateInstance(req.Name, api.InstancePost{
+				// We don't have to handle live migration as the instance is always stopped after copy.
+				// Also we want to always copy the entire instance with all of its snapshots in case they exist.
```
tomponline: why is this if the request is for instance only?

roosterfish: I have the feeling mentioning snapshots here is more confusing than helpful.
Overall at this stage we don't have to care about snapshots as we just move the instance which was created/copied earlier (with or without snaps). 

I was going through the code and whilst crafting the `api.InstancePost{}` request body I thought it's helpful for future readers to mention this bit to give some reasoning how the `api.InstancePost` is crafted/populated as I found in other places the body is often created with various fields set without further explanation. 

I can remove/rephrase it.

lxd/instances_post.go
roosterfish: I have already moved it right after the transaction which populated the `sourceMemberInfo`.

Do you prefer splitting the one big transaction into smaller ones instead?

tomponline: This doesn't look like its in a DB transaction to me.

My thinking here is that as this is doing a forward request, that it would make sense to do that forward request early to avoid doing the earlier logic on both the initial node and the forwarded node.

lxd/instances_post.go
tomponline: So is the intention of your comment to explain why you're not setting InstanceOnly true?

tomponline: OK, in that case mention InstanceOnly in the comment please.

lxd/instances_post.go
roosterfish: I wasn't talking about these redirection bits, but the code block before that:

```
transaction {
   load_all_members
   ...
   filter_out_the_source_member (which currently holds the source instance)
   ...
}

redirect to source {

}
```

I wanted to reuse the already existing info we load using `tx.GetNodes(ctx)` inside the transaction as this provides us with the source member info which we require (address) to redirect the request to the source.
So this new redirect block already comes right after the previous transaction returns. 

The only optimization I see is if we split the already existing transaction into two so that we only run the `tx.GetNodes` to get the source member info, then check if we need to redirect and only afterwards run another transaction with the remaining DB queries e.g loading profile.

lxd/instances_post.go
roosterfish: I see thanks, pushed an updated comment.

lxd/instances_post.go
tomponline: why dont we keep the single transaction but break out of it early if we need to redirect?

To avoid doing expensive calls like `limits.AllowInstanceCreation` twice.

roosterfish: Now I see what you mean, yes let me do this.

lxd/instances_post.go
```diff
@@ -693,7 +697,36 @@ func createFromCopy(ctx context.Context, s *state.State, projectName string, pro
...
+		var targetClient lxd.InstanceServer
+
+		// Move the instance in case it's not yet at its requested target.
+		if s.ServerClustered && targetMemberInfo != nil && targetMemberInfo.Name != s.ServerName {
+			targetClient, err = cluster.Connect(ctx, targetMemberInfo.Address, s.Endpoints.NetworkCert(), s.ServerCert(), false)
+			if err != nil {
+				return err
+			}
+
+			// Move to the actual target.
```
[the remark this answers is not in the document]
roosterfish: The instance is already copied in the right project and only moved between members without changing the project.

lxd/instances_post.go
roosterfish: Fixed.

tomponline (approved):
Thanks!

In a follow up PR do you think we can have a test for this?

roosterfish:
> Thanks!
> 
> In a follow up PR do you think we can have a test for this?

The copy/start inside a cluster using remote storage is already executed when running the new `test_container_copy_start`.
Let me add a few more lines to this test which trigger the copy on another member to specifically also test this redirect to source. 
I'll add it to this PR directly if you don't mind.

tomponline:
> > Thanks!
> > In a follow up PR do you think we can have a test for this?
> 
> The copy/start inside a cluster using remote storage is already executed when running the new `test_container_copy_start`. Let me add a few more lines to this test which trigger the copy on another member to specifically also test this redirect to source. I'll add it to this PR directly if you don't mind.

Thanks!

roosterfish:
Test passed, looking good.

test/suites/container_copy.sh
```diff
@@ -229,6 +232,22 @@ test_container_copy_start() {
...
 
     LXD_DIR="${LXD_ONE_DIR}" lxc delete -f c2
+
+    # Create an instance on the first cluster member.
+    # Then test if an instance copy triggered on another cluster gets redirected to the source cluster member.
+    # In addition force the placement on a third cluster member different than the source/trigger to also cause a move of the instance.
+    LXD_DIR="${LXD_ONE_DIR}" lxc copy c1 c2 --target node1
+
+    echo "==> Check the instance copy can be triggered from any other member than the source with a final placement on a different member"
+    LXD_DIR="${LXD_TWO_DIR}" lxc copy c2 c3 --target node3 --mode "${mode}" --start
```
tomponline, line 242: Did this fail before this PR on ceph, or did it work, but just didn't freeze the filesystem?

roosterfish: It worked without freezing as it copied directly on the target member even though the instance was running on the source.

PR #17351, opened 2026-01-08
Config trust add validate projects and restricted flag

tugbataluy:
## Checklist

## Problem Solved

Users could create certificates with --projects but without --restricted, creating confusing certificates that appear restricted but actually have full access as mentioned in #13865.

## Solution

Server-side validation only: Reject certificate creation/updates when Projects is set but Restricted=false.

### CLI-side validation
- Added the same validation in `cmdConfigTrustAdd` to catch user errors immediately with a clear error message
- **Also added a CLI-only warning** when `--restricted` is used without `--projects`, alerting users that the certificate will have no project access

### Why the warning is CLI-only
The warning for `restricted=true` with empty projects is intentionally **not** enforced server-side because:
1. This is a valid state per the API design (users may add projects later)
2. Warnings are user-facing guidance and belong in the CLI layer
3. Server validation should only reject invalid states, not warn about potentially confusing ones

## Key Changes

    lxd/certificates.go: Added validation in certificatesPost() and doCertificateUpdate() to reject Projects without Restricted=true

    Test fixes: Updated existing tests that were creating invalid certificates (had to clear projects before setting restricted=false)

## Impact

    API users get clear error: "Projects can only be specified for restricted certificates"

    Existing invalid certificates remain (grandfathered)

    New certificates must have consistent projects/restricted values

Fixes: #13865 about confusing certificate behavior.

### DIFF
--- a/doc/howto/projects_confine.md
+++ b/doc/howto/projects_confine.md
@@ -38,6 +38,10 @@ To add the client certificate directly:
 
     lxc config trust add <certificate_file> --projects <project_name> --restricted
 
+```{important}
+The `--projects` flag requires `--restricted` to be set. Projects can only be used to restrict certificate access when the certificate is marked as restricted.
+```
+
 The client can then add the server as a remote in the usual way ([`lxc remote add <server_name> <token>`](lxc_remote_add.md) or [`lxc remote add <server_name> <server_address>`](lxc_remote_add.md)) and can only access the project or projects that have been specified.
 ```{note}
 You can specify the `--project` flag when adding a remote.


lxc/config_trust.go
```diff
@@ -137,6 +141,11 @@ func (c *cmdConfigTrustAdd) run(cmd *cobra.Command, args []string) error {
 		return errors.New("The server doesn't implement metrics")
 	}
 
+	// Warn users that a restricted certificate with no projects will have zero access.
+	if c.flagRestricted && c.flagProjects == "" && !c.global.flagQuiet {
+		logger.Warn("Certificate is restricted but no projects specified. This certificate will have no project access.\n")
```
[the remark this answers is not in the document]
tomponline: @tugbataluy logging shouldnt have newline formatting

test/suites/projects.sh
```diff
@@ -1627,3 +1627,93 @@ EOF
...
+  echo "Test 5.1: Create token with projects but without restricted, should fail."
+  ! lxc config trust add --name test-token-fail --projects test-cert-project || false
+
+  echo "Test 5.2: Create token with both projects and restricted, should succeed."
+  lxc config trust add --name test-token-success --projects test-cert-project --restricted
+
+  echo "Test 5.3: Create regular certificate with projects but without restricted, should fail."
+  LXD_CONF_TEST4=$(mktemp -d -p "${TEST_DIR}" XXX)
+  LXD_CONF="${LXD_CONF_TEST4}" gen_cert_and_key "client5"
+  ! lxc config trust add --name test-cert-fail --projects test-cert-project \ "${LXD_CONF_TEST4}/client.crt" || false
```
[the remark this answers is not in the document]
tomponline: @tugbataluy ^

test/suites/projects.sh
```diff
@@ -1627,3 +1627,93 @@ EOF
...
+
+  # Verify restricted is true.
+  echo "${cert_info}" | grep -F "restricted: true"
+
+  # Verify projects list contains test-cert-project.
+  echo "${cert_info}" | grep -E "^- test-cert-project"
+
+  echo "Test 3: Verify that --restricted without --projects succeeds but warns."
+  # This should succeed but the certificate will have no project access.
+  trust_token2="$(lxc config trust add --name test-no-projects --restricted 2>&1)"
```
tomponline: Rather than munging the stdout and stderr together, then having to extract the different parts out afterwards, can you instead record stderr and stdout to different variables for inspection separately?

tomponline:
Have asked @simondeziel to look at the test part of this PR to review the bash code. Thanks!

tomponline:
@simondeziel @kadinsayani looks like the Copilot rules need to be updated so that it doesn't keep flagging to use our deprecated translation wrapper.

test/suites/projects.sh
```diff
@@ -1627,3 +1627,110 @@ EOF
...
+  LXD_CONF_TEST=$(mktemp -d -p "${TEST_DIR}" XXX)
+  LXD_CONF="${LXD_CONF_TEST}" gen_cert_and_key "client"
+  LXD_CONF="${LXD_CONF_TEST}" lxc remote add test-restricted "${trust_token1}"
+
+  # The restricted client should only see the test-cert-project project.
+  fingerprint1="$(cert_fingerprint "${LXD_CONF_TEST}/client.crt")"
+  cert_info="$(lxc config trust show "${fingerprint1}")"
+
+  # Verify restricted is true.
+  echo "${cert_info}" | grep -F "restricted: true"
```
simondeziel: Let's use `-x` to match full lines where possible.

```suggestion
  echo "${cert_info}" | grep -xF "restricted: true"
```

test/suites/projects.sh
```diff
@@ -1627,3 +1627,110 @@ EOF
...
+
+  # Verify projects list contains test-cert-project.
+  echo "${cert_info}" | grep -E "^- test-cert-project"
```
simondeziel:
```suggestion
  echo "${cert_info}" | grep -xF "- test-cert-project"
```

test/suites/projects.sh
```diff
@@ -1627,3 +1627,110 @@ EOF
...
+
+  # Read the outputs.
+  trust_token_stdout=$(cat "$temp_stdout")
+  trust_token_stderr=$(cat "$temp_stderr")
+
+  # Clean up temp files.
+  rm -f "$temp_stdout" "$temp_stderr"
+
+  # Now verify the warning is in stderr.
+  echo "$trust_token_stderr" | grep -i "Certificate is restricted but no projects specified."
```
simondeziel:
```suggestion
  echo "$trust_token_stderr" | grep -F "Certificate is restricted but no projects specified."
```

test/suites/projects.sh
```diff
@@ -1627,3 +1627,110 @@ EOF
...
+  LXD_CONF="${LXD_CONF_TEST2}" gen_cert_and_key "client"
+
+  # Use the extracted token.
+  LXD_CONF="${LXD_CONF_TEST2}" lxc remote add test-restricted-empty "${actual_token}"
+
+  fingerprint2="$(cert_fingerprint "${LXD_CONF_TEST2}/client.crt")"
+  cert_info2="$(lxc config trust show "${fingerprint2}")"
+
+  # Verify restricted is true.
+  echo "${cert_info2}" | grep -F "restricted: true"
```
simondeziel:
```suggestion
  echo "${cert_info2}" | grep -xF "restricted: true"
```

test/suites/projects.sh
```diff
@@ -1627,3 +1627,110 @@ EOF
...
+
+  # Verify projects list is empty.
+  echo "${cert_info2}" | grep -F "projects: []"
```
simondeziel:
```suggestion
  echo "${cert_info2}" | grep -xF "projects: []"
```

test/suites/projects.sh
```diff
@@ -1627,3 +1627,110 @@ EOF
...
+
+  echo "Test 3: Verify that --restricted without --projects succeeds but warns."
+  # This should succeed but the certificate will have no project access.
+
+  # Create temporary files for stdout and stderr.
+  temp_stdout=$(mktemp)
+  temp_stderr=$(mktemp)
+
+  # Run command, redirecting stdout and stderr separately.
+  lxc config trust add --name test-no-projects --restricted >"$temp_stdout" 2>"$temp_stderr"
```
simondeziel, line 1666: Would it work to use `--quiet` and avoid saving the content into a file?
```suggestion
  trust_token2="$(lxc config trust add --quiet --name test-no-projects --restricted 2>"$temp_stderr")"
```

test/suites/projects.sh
```diff
@@ -1627,3 +1627,110 @@ EOF
...
+  echo "Try to update it with projects but without restricted, should fail."
+  ! lxc query -X PATCH -d '{"projects": ["test-cert-project"], "restricted": false}' "/1.0/certificates/${fingerprint3}" || false
+
+  echo "Update it with both projects and restricted, should succeed."
+  lxc query -X PATCH -d '{"projects": ["test-cert-project"], "restricted": true}' "/1.0/certificates/${fingerprint3}"
+
+  # Verify the update
+  cert_info3="$(lxc config trust show "${fingerprint3}")"
+  echo "${cert_info3}" | grep -F "restricted: true"
+  echo "${cert_info3}" | grep -E "^- test-cert-project"
```
simondeziel:
```suggestion
  echo "${cert_info3}" | grep -xF "restricted: true"
  echo "${cert_info3}" | grep -xF "- test-cert-project"
```

test/suites/projects.sh
```diff
@@ -1627,3 +1627,110 @@ EOF
...
+  echo "Test 1: Verify that --projects without --restricted fails."
+  ! lxc config trust add --name test-fail --projects test-cert-project || false
+
+  echo "Test 2: Verify that --projects with --restricted succeeds."
+  trust_token1="$(lxc config trust add --name test-success --projects test-cert-project --restricted --quiet)"
+
+  echo "Verify the certificate was created with the correct settings."
+  LXD_CONF_TEST=$(mktemp -d -p "${TEST_DIR}" XXX)
+  LXD_CONF="${LXD_CONF_TEST}" gen_cert_and_key "client"
+  LXD_CONF="${LXD_CONF_TEST}" lxc remote add test-restricted "${trust_token1}"
```
simondeziel: For ease of tracking, I'd probably use `LXD_CONF_TEST1` and use `client1` as CN.

test/suites/tls_restrictions.sh
```diff
@@ -458,7 +458,13 @@ test_tls_restrictions() {
   # Cleanup
   delete_object_storage_pool s3
   rm "${TEST_DIR}/${test_image_fingerprint}.tar.xz"
+
+  # First clear projects (while still restricted=true, so validation allows it).
+  lxc config trust show "${FINGERPRINT}" | sed -e '/^- blah$/d' -e 's/^projects:$/projects: []/' | lxc config trust edit "${FINGERPRINT}"
+
+  # Then set restricted=false (now projects is already empty, validation passes).
   lxc config trust show "${FINGERPRINT}" | sed -e "s/restricted: true/restricted: false/" | lxc config trust edit "${FINGERPRINT}"
```
simondeziel: I think that repeating that before emptying the project list would be a good idea to ascertain that even an admin cannot save a broken trust (non-empty project list but `restricted=true`).

lxc/config_trust.go
```diff
@@ -83,9 +82,9 @@ type cmdConfigTrustAdd struct {
 
 func (c *cmdConfigTrustAdd) command() *cobra.Command {
 	cmd := &cobra.Command{}
-	cmd.Use = usage("add", i18n.G("[<remote>:] [<cert>]"))
-	cmd.Short = i18n.G("Add new trusted client")
-	cmd.Long = cli.FormatSection(i18n.G("Description"), i18n.G(
+	cmd.Use = usage("add", "[<remote>:] [<cert>]")
+	cmd.Short = "Add new trusted client"
+	cmd.Long = cli.FormatSection("Description",
```
simondeziel: Could you use `cmd.Short` as the prefix for the description to save on repetition:

```
	cmd.Long = cli.FormatSection("Description", cmd.Short+`

The following certificate types are supported:
...

File: lxc/config_trust.go [hunk 14]
```diff
@@ -96,12 +95,15 @@ If the certificate is omitted, a token will be generated and returned. A client
...
-	cmd.Flags().StringVar(&c.flagName, "name", "", i18n.G("Alternative certificate name")+"``")
-	cmd.Flags().StringVar(&c.flagType, "type", "client", i18n.G("Type of certificate")+"``")
+Note: The --projects flag requires --restricted to be set. Projects can only be
+used to restrict certificate access when the certificate is marked as restricted.
+`)
+
+	cmd.Flags().BoolVar(&c.flagRestricted, "restricted", false, "Restrict the certificate to one or more projects")
+	cmd.Flags().StringVar(&c.flagProjects, "projects", "", "List of projects to restrict the certificate to (requires --restricted)")
+	cmd.Flags().StringVar(&c.flagName, "name", "", "Alternative certificate name")
+	cmd.Flags().StringVar(&c.flagType, "type", "client", "Type of certificate")
```
simondeziel: Those with `StringVar` all need the last param to be preprocessed by `cli.FormatStringFlagLabel()` that will be adding the needed `"``"` quirk pattern.

test/suites/tls_restrictions.sh
```diff
@@ -458,7 +458,13 @@ test_tls_restrictions() {
   # Cleanup
   delete_object_storage_pool s3
   rm "${TEST_DIR}/${test_image_fingerprint}.tar.xz"
+
```
simondeziel, line 464: Trying to remove the restriction while the project is non-empty should always fail right? As such, I'd ascertain it indeed fails with:

```suggestion
  ! lxc config trust show "${FINGERPRINT}" | sed -e "s/restricted: true/restricted: false/" | lxc config trust edit "${FINGERPRINT}" || false
```

And then proceed with what you already have.

test/suites/tls_restrictions.sh
```diff
@@ -519,10 +525,12 @@ test_certificate_edit() {
...
 
-  # Cleanup
-  lxc config trust show "${FINGERPRINT}" | sed -e "s/restricted: true/restricted: false/" | lxc config trust edit "${FINGERPRINT}"
+  # Clean up.
+  # First clear projects (while still restricted=true, so validation allows it).
+  lxc config trust show "${FINGERPRINT}" | sed -e '/^- blah$/d' -e 's/^projects:$/projects: []/' | lxc config trust edit "${FINGERPRINT}"
 
-  lxc config trust show "${FINGERPRINT}" | sed -e ':a;N;$!ba;s/projects:\n- blah/projects: \[\]/' | lxc config trust edit "${FINGERPRINT}"
+  # Then set restricted=false (now projects is already empty, validation passes).
+  lxc config trust show "${FINGERPRINT}" | sed -e "s/restricted: true/restricted: false/" | lxc config trust edit "${FINGERPRINT}"
```
simondeziel, line 540: Since we are in the cleanup phase, I'd suggest to simply `lxc config trust delete "${FINGERPRINT}"`.

IMHO, this trust deletion was missing from the test all along.

kadinsayani:
> @simondeziel @kadinsayani looks like the Copilot rules need to be updated so that it doesn't keep flagging to use our deprecated translation wrapper.

https://github.com/canonical/lxd/pull/17368

test/suites/projects.sh
tugbataluy: I'll use `grep -xF -- "- test-cert-project" ` because:

```
grep -xF "- test-cert-project"
# grep sees: -xF "-" <-- thinks this is an option!
# Error: grep: invalid option -- ' '
```

The -- delimiter is a POSIX standard convention used across Unix tools (grep, rm, mv, etc.) to separate options from arguments when the argument might start with a dash.

The -- is needed because our pattern starts with a dash (-). Without it, grep interprets the dash as an option flag.

test/suites/projects.sh
tugbataluy: I don't think so, as far as I understand:
The --quiet flag suppresses both stdout (the token) and stderr (the warning message). Since this test specifically needs to verify that the warning "Certificate is restricted but no projects specified" appears in stderr, using --quiet would suppress the warning and make the test fail.

test/suites/tls_restrictions.sh
tugbataluy: I intentionally avoid deleting the certificate with `lxc config trust remove` because the test suite uses a shared certificate that all tests depend on. If we remove it, subsequent tests like completions fail with "Rejecting request from untrusted client" errors since they can no longer authenticate with LXD (I tested and got the exact error). Instead, I properly clean up by restoring the certificate to its original unrestricted state (clearing the projects array first while still restricted, then setting restricted to false) which leaves the certificate intact for other tests to use while ensuring our test doesn't leave the system in a modified state.

test/suites/projects.sh
simondeziel: Ah yeah, indeed the `--` bit is needed.

test/suites/tls_restrictions.sh
simondeziel: Sounds good.

lxc/config_trust.go
```diff
@@ -96,12 +94,15 @@ If the certificate is omitted, a token will be generated and returned. A client
...
-	cmd.Flags().BoolVar(&c.flagRestricted, "restricted", false, i18n.G("Restrict the certificate to one or more projects"))
-	cmd.Flags().StringVar(&c.flagProjects, "projects", "", i18n.G("List of projects to restrict the certificate to")+"``")
-	cmd.Flags().StringVar(&c.flagName, "name", "", i18n.G("Alternative certificate name")+"``")
-	cmd.Flags().StringVar(&c.flagType, "type", "client", i18n.G("Type of certificate")+"``")
+Note: The --projects flag requires --restricted to be set. Projects can only be
+used to restrict certificate access when the certificate is marked as restricted.
+`)
+
+	cmd.Flags().BoolVar(&c.flagRestricted, "restricted", false, "Restrict the certificate to one or more projects")
+	cmd.Flags().StringVar(&c.flagProjects, "projects", "", "List of projects to restrict the certificate to (requires --restricted)")
```
simondeziel:
```suggestion
	cmd.Flags().StringVar(&c.flagProjects, "projects", "", cli.FormatStringFlagLabel("List of projects to restrict the certificate to (requires --restricted)"))
```

test/suites/projects.sh
```diff
@@ -1627,3 +1627,110 @@ EOF
...
+  trust_token_stderr=$(cat "$temp_stderr")
+
+  # Clean up temp files.
+  rm -f "$temp_stdout" "$temp_stderr"
+
+  # Now verify the warning is in stderr.
+  echo "$trust_token_stderr" | grep -F "Certificate is restricted but no projects specified."
+
+  # Extract just the token from stdout - it's on the last line.
+  actual_token=$(echo "$trust_token_stdout" | tail -n1)
```
simondeziel: I'd probably rename the variable from `actual_token` to `trust_token2` (for consistency) and leverage `tail` and `grep` ability to read from file (for simplicity):

```suggestion
  # Read the outputs.
  # Extract just the token from stdout - it's on the last line.
  trust_token2="$(tail -n1 "${temp_stdout}")"
  # Now verify the warning is in stderr.
  grep -F "Certificate is restricted but no projects specified." "${temp_stderr}"
  
  # Clean up temp files.
  rm -f "$temp_stdout" "$temp_stderr"
```

lxc/config_trust.go
tugbataluy: Oops, I missed that thanks Simon ❇️

test/suites/projects.sh
tugbataluy: Yes it seems more readable and simple, thanks.

PR #17352, opened 2026-01-08
Fix qemu crashes

kadinsayani:


kadinsayani:
@tomponline I'm not sure if this approach is correct, but opening this as draft for observability. Would appreciate your thoughts whenever you can :)

lxd/instance/drivers/qmp/monitor.go
```diff
@@ -84,6 +92,63 @@ func (m *Monitor) start() error {
...
+	}
+
+	// Start ringbuffer monitoring go routine.
+	go func() {
+		var lastRead time.Time
+		ticker := time.NewTicker(ringbufReadInterval)
+		defer ticker.Stop()
+		// Initialize a stopped timer for rate-limited reads.
+		timer := time.NewTimer(0)
+		if !timer.Stop() {
```
kadinsayani: We might not need this.

Issue #16028, opened 2025-07-14
Image: `--project` and `--target-project` have identical usage for `image copy`


markylaing:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

22.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
-
```

### Output of "lxc info" or system info if it fails

```shell
-
```

### Issue description

For image copy, neither `--project` or `--target-project` affect the project for the source server. Both affect the target server.

Note that the help text for `lxc image copy` has: `--project        Override the source project`.

The first issue that mentions this is https://github.com/canonical/lxd/issues/9234. In this issue [it is agreed](https://github.com/canonical/lxd/issues/9234#issuecomment-924443647) that the `--project` flag should affect the project of the destination server. The PR that fixes the issue overwrites the project on the source server (which may have been set by `--project`) with the "current" project for that remote in the config.yml.

Another issue appears later: https://github.com/canonical/lxd/issues/10009. This one incorrectly states that the `--project` flag overrides the source project, when in fact it overrides the target project. The [PR to fix it](https://github.com/canonical/lxd/pull/10068) adds support for a target project when copying instances in push or relay mode, but then also (unnecessarily) adds the `--target-project` flag.

### Steps to reproduce

```bash
$ lxc project list
+-------------------+--------+----------+-----------------+-----------------+----------+---------------+---------------------+---------+
|       NAME        | IMAGES | PROFILES | STORAGE VOLUMES | STORAGE BUCKETS | NETWORKS | NETWORK ZONES |     DESCRIPTION     | USED BY |
+-------------------+--------+----------+-----------------+-----------------+----------+---------------+---------------------+---------+
| default (current) | YES    | YES      | YES             | YES             | YES      | YES           | Default LXD project | 2       |
+-------------------+--------+----------+-----------------+-----------------+----------+---------------+---------------------+---------+
$ lxc image list
+-------+-------------+--------+-------------+--------------+------+------+-------------+
| ALIAS | FINGERPRINT | PUBLIC | DESCRIPTION | ARCHITECTURE | TYPE | SIZE | UPLOAD DATE |
+-------+-------------+--------+-------------+--------------+------+------+-------------+
$ lxc image copy images:alpine/3.20 local:
Image copied successfully!                     
$ lxc image list
+-------+--------------+--------+-----------------------------------+--------------+-----------+---------+------------------------------+
[...]

PR #17356, opened 2026-01-09
lxc/image: Make project flag behavior consistent for image copy command

nmezhenskyi:
This PR fixes the global flag `--project` behavior for `lxc image copy` command. It ensures that `--project` overrides the source project. This behavior is consistent with `lxc copy` and `lxc storage volume copy` commands.

This fix is needed because currently for `lxc image copy`, the `--project` flag dictates the destination (target) project, making it functionally equivalent to `--target-project` flag.

Fixes https://github.com/canonical/lxd/issues/16028.

### DIFF
--- a/test/suites/remote.sh
+++ b/test/suites/remote.sh
@@ -407,8 +407,22 @@ test_remote_usage() {
   [ "${alias}" = "bar" ]
 
   lxc_remote image alias delete localhost:foo
+  lxc_remote image delete lxd2:bar
 
-  lxc_remote image delete localhost:testimage
+  echo "==> Test copying image on the same remote into a different project."
+  lxc_remote project create localhost:p1
+  lxc_remote image copy --quiet localhost:testimage localhost: --project default --target-project p1 --copy-aliases
+  lxc_remote image delete localhost:testimage --project default
+
+  echo "==> Test copying image from one remote's non-default project into another remote's non-default project."
+  lxc_remote project create lxd2:foo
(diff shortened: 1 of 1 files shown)


nmezhenskyi:
@tomponline Based on the context in the issue description, it seemed preferable to change the flag's description in order to avoid introducing a breaking change for some users.

However, I see that in the past [there were conversations](https://github.com/canonical/lxd/pull/16005#issuecomment-3088490138) about reverting the changes done to `lxc image copy` to fix the non-standard behavior of `--project` flag there.

What is the current stance on this?

tomponline:
> @tomponline Based on the context in the issue description, it seemed preferable to change the flag's description in order to avoid introducing a breaking change for some users.
> 
> However, I see that in the past [there were conversations](https://github.com/canonical/lxd/pull/16005#issuecomment-3088490138) about reverting the changes done to `lxc image copy` to fix the non-standard behavior of `--project` flag there.
> 
> What is the current stance on this?

You're referring to @skozina comment here? https://github.com/canonical/lxd/pull/16005#issuecomment-3088490138

@markylaing @skozina there does seem to be some inconsistency here? Any preference on how to proceed?

IIUC the behaviour of the `--project` flag for `lxc image copy` is different than for `lxc copy` and `lxc storage volume copy`, so it seems to me we should fix that issue.

nmezhenskyi:
> You're referring to @skozina comment here? https://github.com/canonical/lxd/pull/16005#issuecomment-3088490138

@tomponline Yes

nmezhenskyi:
@tomponline This should be ready for review now.

tomponline (approved):
ta

PR #17360, opened 2026-01-11
Simplify cluster formation in tests

simondeziel:


simondeziel:
> This isn't a pattern I'd like us to start using as its hard to reason about.

Hmm, good thing we mention you don't like it because I was on my way to also drop another argument (set to `"${LXD_ONE_DIR}`, 99.9% of the time) so changing this

```
  spawn_lxd_and_bootstrap_cluster

  local cert
  cert="$(cert_to_yaml "${LXD_ONE_DIR}/cluster.crt")"

  spawn_lxd_and_join_cluster "${cert}" 2 1 "${LXD_ONE_DIR}"
  spawn_lxd_and_join_cluster "${cert}" 3 1 "${LXD_ONE_DIR}"
```

To:

```
  spawn_lxd_and_bootstrap_cluster

  spawn_lxd_and_join_cluster 2 1
  spawn_lxd_and_join_cluster 3 1
```

> Can we get `spawn_lxd_and_bootstrap_cluster` to return the certificate that can be fed into `spawn_lxd_and_join_cluster` instead to avoid the use of environment variables for passing info between functions.

So you are asking for this instead?

```
  local cert
  cert="$(spawn_lxd_and_bootstrap_cluster)"

  spawn_lxd_and_join_cluster "${cert}" 2 1 "${LXD_ONE_DIR}"
  spawn_lxd_and_join_cluster "${cert}" 3 1 "${LXD_ONE_DIR}"
```

tomponline:
> So you are asking for this instead?
> 
> ```
>   local cert
>   cert="$(spawn_lxd_and_bootstrap_cluster)"
> 
>   spawn_lxd_and_join_cluster "${cert}" 2 1 "${LXD_ONE_DIR}"
>   spawn_lxd_and_join_cluster "${cert}" 3 1 "${LXD_ONE_DIR}"
> ```

Yeah thats it. To make it clear the flow of info, rather than relying on global variable mutation.

simondeziel:
> > So you are asking for this instead?
> > ```
> >   local cert
> >   cert="$(spawn_lxd_and_bootstrap_cluster)"
> > 
> >   spawn_lxd_and_join_cluster "${cert}" 2 1 "${LXD_ONE_DIR}"
> >   spawn_lxd_and_join_cluster "${cert}" 3 1 "${LXD_ONE_DIR}"
> > ```
> 
> Yeah thats it. To make it clear the flow of info, rather than relying on global variable mutation.

Unfortunately, that doesn't work due to `spawn_lxd_and_bootstrap_cluster` creating a process that keeps writing to `stdout` and `stderr` so trying to capture its `stdout` means it never returns until the spawned process returns.. so never:

```
  local cert
  cert="$(spawn_lxd_and_bootstrap_cluster)"  # hangs forever

  spawn_lxd_and_join_cluster "${cert}" 2 1 "${LXD_ONE_DIR}"
  spawn_lxd_and_join_cluster "${cert}" 3 1 "${LXD_ONE_DIR}"
```

So @tomponline should I just close this PR?

tomponline:
> So @tomponline should I just close this PR?

why do you need to capture stdout from it?

simondeziel:
> > So @tomponline should I just close this PR?
> 
> why do you need to capture stdout from it?

Because in shell you can only return a code so if you want textual information, you need to capture the output. So at this point, the PR changes are moot.

tomponline:
> > > So @tomponline should I just close this PR?
> > 
> > 
> > why do you need to capture stdout from it?
> 
> Because in shell you can only return a code so if you want textual information, you need to capture the output. So at this point, the PR changes are moot.

So you cant return a value from a function without going via stdout?

simondeziel:
> > > > So @tomponline should I just close this PR?
> > > 
> > > 
> > > why do you need to capture stdout from it?
> > 
> > 
> > Because in shell you can only return a code so if you want textual information, you need to capture the output. So at this point, the PR changes are moot.
> 
> So you cant return a value from a function without going via stdout?

No. The closest you can achieve (if you don't want to capture `stdout`) is to pass the name of the variable you'd like your function to set, something like this:

```
  local cert
  spawn_lxd_and_bootstrap_cluster cert
  
  spawn_lxd_and_join_cluster "${cert}" 2 1 "${LXD_ONE_DIR}"
  spawn_lxd_and_join_cluster "${cert}" 3 1 "${LXD_ONE_DIR}"
```

The problem with that is `spawn_lxd_and_bootstrap_cluster` would need to grow an argument to know which variable to save the cluster cert into (`cert` here).

tomponline:
OK lets close this then.

Issue #17170, opened 2025-12-05
Evacuating a node does not evacuate instances that were busy setting up at the time of the evacuation


vosdev:
# Description

I just evacuated multiple nodes in sequential order via the web UI.

node1, node2, node3 and node4.

The VMs running on node1 were migrated elsewhere (node2, node3 or node4) but I also evacuated those nodes a few seconds later (And while the VMs/LXCs from node1 were evacuating towards these nodes.)

So now I have node2, node3 and node4 in an evacuated state but there are still VMs and LXCs running on some of them that evacuated from node1 or node2.

This is a mostly UI only issue as the CLI only exits the command when the evacuation process of a node is complete.
 
  # Metadata
    
4 node cluster, LXD 6.6
  UI Version: 0.19

kadinsayani:
@tomponline since this on theme with my current work, I can take a look at it next pulse.

kadinsayani:
@tomponline here are a few potential approaches:

1. Retry loop: keep re‑checking instances for a short window during evacuation; if new ones land, evacuate them too.
2. Two‑phase state: introduce EVACUATING/DRAINING state (unschedulable for new placements). Any in‑flight migration targeting an evacuating member is rejected/rescheduled (or queued for evacuation once a target exists). Mark EVACUATED only once instance count reaches 0.

What do you think?

tomponline:
I interpreted this problem differently to you.

For me the problem is that the members that were receiving instances from another member (or even from outside of the cluster, perhaps from a replicator sync) are allowed to have an evacuation started in the first place.

So, really I think we need some sort of validation on the member being evacuated to check no ongoing instance operations are occurring on that member before marking the member as evacuated and starting the evacuation.

vosdev:
Will that solution still allow us to mark multiple hosts for maintenance mode simultaneously?

For example on a 25 node cluster I want to mark 5 hosts for maintenance and evacuate them. It would be a huge shame if after this bug is fixed we would have to evacuate them 1 at a time and wait until each host is done before we can mark the next host as evacuated.

> For me the problem is that the members that were receiving instances from another member (or even from outside of the cluster, perhaps from a replicator sync) are allowed to have an evacuation started in the first place.

With this logic in mind, if host1 is receiving an instance and I mark host1 for maintenance mode, it should then no longer accept new incoming instances, wait for the current incoming instance to finish migrating and then start evacuating its own instances to other members of the cluster.

tomponline:
[...]

Issue #11636, opened 2023-05-02
Concurent instance creation from same unavailable image cause DB errors


stgraber:
On a reasonably slow system, running those 3 concurrently:
 - lxc launch images:ubuntu/22.04/cloud k8s01 --vm -c limits.cpu=2 -c limits.memory=4GiB --target micro01
 - lxc launch images:ubuntu/22.04/cloud k8s02 --vm -c limits.cpu=2 -c limits.memory=4GiB --target micro02
 - lxc launch images:ubuntu/22.04/cloud k8s03 --vm -c limits.cpu=2 -c limits.memory=4GiB --target micro03

Will cause the first one to download the image, the other two will wait as expected but then after a tiny bit, they both report:
```
Error: Failed instance creation: UNIQUE constraint failed: images.project_id, images.fingerprint
```

This looks like a lock may be released at the time the image is done downloading and starts being unpacked.
It's worth mentioning that this was in a cluster, so the image also would have needed to get internally replicated between machines.

stgraber:
@tomponline mentioned that @ru-fu hit something similar too

tomponline:
Reproduced this issue.

tomponline:
@stgraber I think the problem is that LXD doesn't currently have (or at least not that I'm aware of) any form of distributed locking functionality available. Normally this isn't an issue because most operations occur on a single machine, where we do already have local locking primitives to prevent this sort of thing:

Such as:
https://github.com/search?q=repo%3Alxc%2Flxd%20imageOperationLock&type=code

However in this scenario we are creating instances on each cluster member concurrently and so they all read the DB, notice there is no image record yet, proceed to download the image (whilst locking locally), and then when it comes to creating the DB record in the `images`, `images_profiles` and `images_properties` tables via `s.DB.Cluster.CreateImage` one member will succeed and the others will fail with a duplicate record error as we are seeing.

The way we have overcome cluster wide operations like this in the past, such as `autoSyncImagesTask` task, is to run the task on the DB leader and ship out individual per-member requests as needed.

So I wondered if you have any preferences on:

1. Creating an internal API endpoint for managing locks on the leader that cluster members can request/release (with all the associated complexity surrounding releases on timeout/disconnects). The cluster members can then attempt to create a cluster wide lock before trying to download the image.
2. Or updating the image handling logic so that the actual download and main DB record creation is directed towards the leader to perform (with the existing local locking) and when that is completed, the individual members then proceed to create their entries in the images_nodes table.
[...]

PR #17371, opened 2026-01-12
Prevent concurrent evacuations to prevent races

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/17170. Depends on https://github.com/canonical/lxd/pull/17376.

### DIFF
--- a/doc/rest-api.yaml
+++ b/doc/rest-api.yaml
@@ -4396,7 +4396,7 @@ definitions:
                 type: string
                 x-go-name: ID
             location:
-                description: What cluster member this record was found on
+                description: Which cluster member this record was found on
                 example: lxd01
                 type: string
                 x-go-name: Location


test/suites/clustering.sh
```diff
@@ -3543,14 +3543,23 @@ test_clustering_evacuation_restore_operations() {
...
   lxc cluster evacuate node1 --quiet --force &
   evac_pid=$!
   sleep 1 # Wait a bit for the operation to start
 
+  echo "Check evacuation fails while another evacuation is in progress"
+  # Tests fix for https://github.com/canonical/lxd/issues/17170
+  [ "$(CLIENT_DEBUG="" SHELL_TRACING="" LXD_DIR="${LXD_TWO_DIR}" lxc cluster evacuate node2 --force 2>&1)" = "Error: Failed updating cluster member state: Cannot evacuate \"node2\" while another evacuation is in progress (cluster member \"node1\" is being evacuated)" ]
+
   echo "Check restore fails while evacuation operation in progress"
   [ "$(CLIENT_DEBUG="" SHELL_TRACING="" lxc cluster restore node1 --force 2>&1)" = "Error: Failed updating cluster member state: Cannot restore \"node1\" while an evacuate operation is in progress" ]
```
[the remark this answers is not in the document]
kadinsayani: This is a common pattern across tests. I think readability is fine here.

lxd/api_cluster.go
```diff
@@ -3195,6 +3195,21 @@ func clusterNodeStatePost(d *Daemon, r *http.Request) response.Response {
 			}
 		}
 
+		// Check for ongoing evacuation operations on other cluster members.
+		// Evacuating multiple members simultaneously can cause race conditions where
+		// instances from one evacuation cannot find available targets because potential
+		// targets are also being evacuated.
+		evacuateOps, err := operationsGetByType(r.Context(), s, "", operationtype.ClusterMemberEvacuate)
```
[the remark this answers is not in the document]
kadinsayani: We have comprehensive integration test coverage for evacuation. Adding unit tests is a separate task, not related to this PR.

test/suites/clustering.sh
```diff
@@ -3540,14 +3543,23 @@ test_clustering_evacuation_restore_operations() {
...
 
-  echo "Launch 3 containers on node1"
-  for c in c{1..3}; do lxc launch testimage "${c}" --target node1; done
+  echo "Launch 3 containers on node1 and add data to slow down migration"
+  for c in c{1..3}; do
+    lxc launch testimage "${c}" --target node1
+  done
+
+  echo "Add data to c1 to slow down migration"
+  lxc exec c1 -- dd if=/dev/zero of=/root/data1 bs=1M count=1000 2>/dev/null
```
simondeziel: If a huge file full of zeros slow down the migration maybe the more efficient `fallocate -l 1G` or `truncate -s 1G` would also achieve the same with less disk IOs.

It may not work are those create special files but it's worth a try.

kadinsayani: Thanks, I'll give those a try.

kadinsayani: `fallocate` works well. Thanks :)

test/suites/clustering.sh
```diff
@@ -3540,14 +3543,23 @@ test_clustering_evacuation_restore_operations() {
...
 
-  echo "Launch 3 containers on node1"
-  for c in c{1..3}; do lxc launch testimage "${c}" --target node1; done
+  echo "Launch 3 containers on node1 and add data to slow down migration"
+  for c in c{1..3}; do
+    lxc launch testimage "${c}" --target node1
+  done
+
+  echo "Add data to c1 to slow down migration"
+  lxc exec c1 -- fallocate -l 1G /root/data1
```
simondeziel: Would a `lxc exec c1 -- sleep infinity &` also work to delay the migration due to the clean shutdown trying to wait for the operation to finish?

The it could be `kill`'ed to unblock the migration.

kadinsayani: Great idea, however, we have a timeout of 30s for waiting for a clean shutdown during evacuation, after that the instance is force stopped. I think the data file approach is more reliable.

kadinsayani: The `sleep infinity` would just delay the shutdown phase, but we want to delay the migration phase.

simondeziel: Ack

tomponline: why do we need longer than 30s?

tomponline: and can it be influenced by https://documentation.ubuntu.com/lxd/latest/reference/instance_options/#instance-boot:boot.host_shutdown_timeout ?

test/suites/clustering.sh
```diff
@@ -3543,14 +3543,23 @@ test_clustering_evacuation_restore_operations() {
...
 
   echo "Start node1 evacuation in background"
   lxc cluster evacuate node1 --quiet --force &
   evac_pid=$!
   sleep 1 # Wait a bit for the operation to start
```
tomponline: do we need this sleep before the concurrent evacuation check, because if so, its indication we've got a race where its still possible to launch multiple evacuations at the same time.

This is likely similar to https://github.com/canonical/lxd/issues/11636 where really we need a way to atomically create-if-no-existing-operation-exists rather than iterating through the operations list.

This might be dependent on having more info in the DB, like operation type, and might be dependent on @skozina refactoring of the operations table for the durable operations work.

test/suites/clustering.sh
kadinsayani: Since we're going with the operation approach, we don't need this delay. But figured I'd look into this and clarify:

- We don't need longer than 30s. The data file approach actually ends up taking much less time.
- The shutdown timeout can be influenced by `boot.host_shutdown_timeout`.

tomponline: Can we avoid the data file approach then?

kadinsayani: Yes, we won't need it with the operation based approach.

test/suites/clustering.sh
kadinsayani: I've reduced the `sleep` time to 0.1s, just enough time for the operation to register in the DB.

lxd/operations/linux.go
```diff
@@ -37,6 +57,10 @@ func registerDBOperation(op *Operation, opType operationtype.Type) error {
 		return err
 	})
 	if err != nil {
+		if errors.Is(err, errEvacuationInProgress) {
+			return err
+		}
+
 		return fmt.Errorf("failed to add %q Operation %s to database: %w", opType.Description(), op.id, err)
```
simondeziel:
```suggestion
		return fmt.Errorf("Failed to add %q operation %s to database: %w", opType.Description(), op.id, err)
```

test/suites/clustering.sh
tomponline: but the API shouldnt return to the caller until thats occurred, no sleep should be needed.

kadinsayani: The sleep is only needed because we start the first evacuation in the background and immediately issue the second request. The API does register the operation before returning, but by backgrounding the CLI we’re not waiting for that request to even reach the server. So there’s a short race where node2’s request can run before node1’s request has created the op in the DB.

tomponline: oh right i see. ta

lxd/api_cluster.go
```diff
@@ -3280,6 +3280,7 @@ func clusterNodeStatePost(d *Daemon, r *http.Request) response.Response {
 			Type:        operationtype.ClusterMemberEvacuate,
 			Class:       operations.OperationClassTask,
 			RunHook:     run,
+			Reference:   operationtype.ClusterMemberEvacuate.Description(),
```
tomponline: I'd love it if we could have a comment here explaining **why** this is being done, and even better an explanation in the commit message too. Document the **why** not just the **what** please.

lxd/operations/linux.go
```diff
@@ -17,7 +18,26 @@ func registerDBOperation(op *Operation, opType operationtype.Type) error {
 		return nil
 	}
 
+	var errEvacuationInProgress = errors.New("Another evacuation is already in progress")
+
 	err := op.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+		// Check if an evacuation operation is already in progress on another node.
+		// Evacuating multiple members simultaneously can cause race conditions.
```
tomponline: Rather than tightly coupling the duplicate detection to the operation type description, which is only one form of cluster wide unique operation reference we may use (think image downloads which would use the image fingerprint to derive the reference), instead can you detect the insertion DB conflict error and then convert that to something more generic, like "An operation with the same reference is ongoing" and make it a HTTP conflict response type using `api.StatusErrorf`.

tomponline: Take a look at `CreateStoragePoolBucket`, e.g.

```go
	result, err := c.tx.ExecContext(ctx, `
		INSERT INTO storage_buckets
		(storage_pool_id, node_id, name, description, project_id)
		VALUES (?, ?, ?, ?, (SELECT id FROM projects WHERE name = ?))
		`, poolID, nodeID, info.Name, info.Description, projectName)
	if err != nil {
		if query.IsConflictErr(err) {
			return -1, api.StatusErrorf(http.StatusConflict, "A bucket for that name already exists")
		}

		return -1, err
	}
```

test/suites/clustering.sh
kadinsayani: I've gone with 0.5s instead of 0.1s since the overhead is minimal and it is safer.

lxd/api_cluster.go
kadinsayani: Added. I've added more comments and explanations across the PR too. Working on making this habit :)

lxd/db/cluster/operations.go
```diff
@@ -15,11 +15,13 @@ import (
 //go:generate mapper stmt -e operation objects-by-NodeID
 //go:generate mapper stmt -e operation objects-by-ID
 //go:generate mapper stmt -e operation objects-by-UUID
+//go:generate mapper stmt -e operation create
```
skozina, line 18: I've not really understood why the current `registerDBOperation()` uses `CreateOrReplace` rather than plain `Create`. Even for the durable operations when we need to just update the nodeID or status, I've introduced separate DB functions which only UPDATE whatever column we need to update rather than to use the full CreateOrReplace logic.
Should we just drop the CreateOrReplace and use Create with the logic to detect conflicts everywhere?

lxd/operations/linux.go
```diff
@@ -33,10 +35,30 @@ func registerDBOperation(op *Operation, opType operationtype.Type) error {
 			opInfo.ProjectID = &projectID
 		}
 
-		_, err := cluster.CreateOrReplaceOperation(ctx, tx.Tx(), opInfo)
-		return err
+		if op.fixedRef {
```
skozina: Just one more reason to use the plain INSERT everywhere.

lxd/db/cluster/operations.go
kadinsayani: I did some investigating, and my hypothesis is that `CreateOrReplaceOperation` was used to guard against UUID conflicts. Cluster DB transactions are wrapped in `query.Retry` and can re-run on retriable errors. If a transaction commits but the client sees a retriable error, the retry could insert the same UUID successfully. 

`CreateOrReplaceOperation` also defends against any code path that registers duplicate op ID entries. I don't think we have any today, hence why this was probably just defensive.

kadinsayani: There's no operation that relies on the replace semantics inside `registerDBOperation`. I think we can just use `CreateOperation`. cc. @tomponline

lxd/operations/linux.go
kadinsayani: Great suggestion, I've gone with this approach. Thanks.

lxd/db/cluster/operations.go
kadinsayani: @skozina @tomponline here is the other place where `CreateOrReplaceOperation` is used (in `cluster.Join`):

https://github.com/canonical/lxd/blob/b39fc6efb943d645791ebe9dc51999f82ce76c43/lxd/cluster/membership.go#L594-L606

kadinsayani: I've reworked `registerDBOperation` to always use `CreateOperation` in this PR.

test/suites/clustering.sh
```diff
@@ -1,3 +1,19 @@
+wait_for_evacuation_op() {
```
simondeziel: I'd make that a sub-function in the `test_clustering_evacuation` unless you envision other use in which case I'd put it in `test/includes/operations.sh`.

kadinsayani: I've moved it into `test_clustering_evacuation` and reduced the polling interval, thanks.

test/suites/clustering.sh
```diff
@@ -3165,6 +3165,23 @@ test_clustering_image_refresh() {
 }
 
 test_clustering_evacuation() {
+  wait_for_evacuation_op() {
```
simondeziel, line 3168: Too bad we don't have a `lxc operation wait ${UUID}` ;)

kadinsayani: That would be a handy command! Can you please open a GH issue feature request so we can circle back to it for discussion and potentially implementation?

kadinsayani: Happy to have that assigned to me :)

lxd/operations/linux.go
```diff
@@ -33,10 +35,23 @@ func registerDBOperation(op *Operation, opType operationtype.Type) error {
...
+				return api.StatusErrorf(http.StatusConflict, "Another operation with reference %q already exists", op.id)
+			}
+
+			return err
+		}
+
+		return nil
 	})
 	if err != nil {
+		if api.StatusErrorCheck(err, http.StatusConflict) {
```
tomponline, line 51: This is unnecessary, we dont want 2 places of handling the conflict type.

lxd/operations/linux.go
```diff
@@ -52,7 +52,7 @@ func registerDBOperation(op *Operation, opType operationtype.Type) error {
 			return err
 		}
 
-		return fmt.Errorf("failed to add %q Operation %s to database: %w", opType.Description(), op.id, err)
+		return fmt.Errorf("Failed adding %q Operation %s to database: %w", opType.Description(), op.id, err)
```
tomponline, line 55: do we normally say "creating" or "adding" in DB create errors?

lxd/operations/linux.go
kadinsayani: Resolved in follow-up PR: https://github.com/canonical/lxd/pull/17397.

lxd/operations/linux.go
kadinsayani: Creating. Resolved in follow-up.

lxd/db/cluster/operations.go
skozina: Thank you!
>  my hypothesis is that CreateOrReplaceOperation was used to guard against UUID conflicts

I still don't follow. The `CreateOrReplace` should be *less* defensive than pure `Create`, as it's more likely to just replace the row with the new one (rather than re-try the insert with different UUID), is it?

kadinsayani: It's defensive against callers that may supply the same UUID twice (we don't have any that do this).

kadinsayani: The exact reason why `CreateOrReplace` was used in `registerDBOperation` remains unknown.

PR #17374, opened 2026-01-13
tests: VM live migration

MusicDin:
The PR adds VM live migration using Ceph for instance root volume, as well as an attached custom block volume. The test migrates the VM and checks whether the content on the custom block volume is retained after migration.

Since the test spawns two LXD servers on the same host, an EBUSY error is returned after an instance is migrated when LXD tries to unmount the source volume which is already mounted to the destination instance. 

The instance stop is tweaked in a way to only print a warning when this occurs if the operation action is migration, the error is EBUSY, and the environment variable `LXD_TEST_LIVE_MIGRATION_ON_THE_SAME_HOST` is set to true.

Closes https://github.com/canonical/lxd-ci/pull/256
Closes https://github.com/canonical/lxd-ci/pull/526

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -18,29 +18,30 @@ sudo -E ./main.sh
 
 ## Environment variables
 
-Name                           | Default                   | Description
-:--                            | :---                      | :----------
-`LXD_BACKEND`                  | dir                       | What backend to test against (btrfs, ceph, dir, lvm, zfs, or random)
-`LXD_BACKENDS`                 | `LXD_BACKEND`             | Space-delimited list of backends to run test against. Accepts `fast` (`btrfs` or `dir`), `fasts` (`btrfs` and `dir`) and `all` (see the list in `LXD_BACKEND`)
-`LXD_CEPH_CLUSTER`             | ceph                      | The name of the ceph cluster to create osd pools in
-`LXD_CEPH_CEPHFS`              | cephfs                    | Enables the CephFS tests using the specified cephfs filesystem for `cephfs` pools
-`LXD_CEPH_CEPHOBJECT_RADOSGW`  | ""                        | Enables the Ceph Object tests using the specified radosgw HTTP endpoint for `cephobject` pools
-`LXD_VERBOSE`                  | ""                        | Run lxd, lxc and the shell in verbose mode (used in CI; less verbose than `LXD_DEBUG`)
-`LXD_DEBUG`                    | ""                        | Run lxd, lxc and the shell in debug mode (very verbose)
-`LXD_INSPECT`                  | 0                         | Set to 1 to start an inspection shell in the test environment on failure
-`LXD_LOGS`                     | ""                        | Path to a directory to copy all the LXD logs to
-`LXD_OFFLINE`                  | 0                         | Skip anything that requires network access
-`LXD_REPEAT_TESTS`             | 1                         | Number of times to repeat test(s)
-`LXD_RANDOMIZE_TESTS`          | 0                         | Randomize the order of tests in a group
-`LXD_SKIP_TESTS`               | ""                        | Space-delimited list of test names to skip
-`LXD_TEST_IMAGE`               | "" (busybox test image)   | Path to an image tarball to use instead of the default busybox image
-`LXD_TMPFS`                    | 0                         | Sets up a tmpfs for the whole testsuite to run on (fast but needs memory)
-`LXD_NIC_SRIOV_PARENT`         | ""                        | Enables SR-IOV NIC tests using the specified parent device
-`LXD_IB_PHYSICAL_PARENT`       | ""                        | Enables Infiniband physical tests using the specified parent device
-`LXD_IB_SRIOV_PARENT`          | ""                        | Enables Infiniband SR-IOV tests using the specified parent device
-`LXD_NIC_BRIDGED_DRIVER`       | ""                        | Specifies bridged NIC driver for tests (either native or openvswitch, defaults to native)
-`LXD_REQUIRED_TESTS`           | ""                        | Space-delimited list of test names that must not be skipped if their prerequisites are not met
-`LXD_VM_TESTS`                 | 1                         | Enables tests using VMs and the on-demand installation of the needed tools
+Name                                       | Default                   | Description
+:--                                        | :---                      | :----------
+`LXD_BACKEND`                              | dir                       | What backend to test against (btrfs, ceph, dir, lvm, zfs, or random)
+`LXD_BACKENDS`                             | `LXD_BACKEND`             | Space-delimited list of backends to run test against. Accepts `fast` (`btrfs` or `dir`), `fasts` (`btrfs` and `dir`) and `all` (see the list in `LXD_BACKEND`)
+`LXD_CEPH_CLUSTER`                         | ceph                      | The name of the ceph cluster to create osd pools in
+`LXD_CEPH_CEPHFS`                          | cephfs                    | Enables the CephFS tests using the specified cephfs filesystem for `cephfs` pools
+`LXD_CEPH_CEPHOBJECT_RADOSGW`              | ""                        | Enables the Ceph Object tests using the specified radosgw HTTP endpoint for `cephobject` pools
+`LXD_VERBOSE`                              | ""                        | Run lxd, lxc and the shell in verbose mode (used in CI; less verbose than `LXD_DEBUG`)
+`LXD_DEBUG`                                | ""                        | Run lxd, lxc and the shell in debug mode (very verbose)
+`LXD_INSPECT`                              | 0                         | Set to 1 to start an inspection shell in the test environment on failure
+`LXD_LOGS`                                 | ""                        | Path to a directory to copy all the LXD logs to
+`LXD_OFFLINE`                              | 0                         | Skip anything that requires network access
+`LXD_REPEAT_TESTS`                         | 1                         | Number of times to repeat test(s)
+`LXD_RANDOMIZE_TESTS`                      | 0                         | Randomize the order of tests in a group
+`LXD_SKIP_TESTS`                           | ""                        | Space-delimited list of test names to skip
+`LXD_TEST_IMAGE`                           | "" (busybox test image)   | Path to an image tarball to use instead of the default busybox image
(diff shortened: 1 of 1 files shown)


test/includes/test-groups.sh
```diff
@@ -33,6 +33,7 @@ readonly test_group_cluster=(
     "clustering_evacuation"
     "clustering_evacuation_restore_operations"
     "clustering_move"
+    "clustering_live_migration"
```
MusicDin: Move to `test_group_cluster_storage`.

MusicDin: Done

test/suites/live_migration.sh
```diff
@@ -0,0 +1,88 @@
...
+# and runs live migration. Success is determined by the data being intact after live migration.
+test_clustering_live_migration() {
+  poolDriver=$(lxc storage show "$(lxc profile device get default root pool)" | awk '/^driver:/ {print $2}')
+  if [ "${poolDriver:-}" != "ceph" ]; then
+    export TEST_UNMET_REQUIREMENT="Only supports 'ceph', not ${poolDriver}"
+    return 0
+  fi
+
+  # Spawn the first node and bootstrap the cluster.
+  LXD_LIVE_MIGRATION_TEST_ON_THE_SAME_HOST=true spawn_lxd_and_bootstrap_cluster "${poolDriver}"
```
simondeziel: I'd extend the comment explaining why this extra long variable is set here.

test/suites/live_migration.sh
```diff
@@ -0,0 +1,88 @@
...
+  # After live migration, the volume should be functional and mounted.
+  # Check that the file we created is still there with the same contents.
+  echo "Verifying data integrity after live migration"
+  [ "$(LXD_DIR=${LXD_ONE_DIR} lxc exec vm -- cat /mnt/vol1/bar)" = "vm" ]
+
+  # Cleanup
+  echo "Cleaning up ..."
+  LXD_DIR="${LXD_ONE_DIR}" lxc delete --force vm
+  LXD_DIR="${LXD_ONE_DIR}" lxc storage volume delete "${poolName}" vmdata
+  LXD_DIR="${LXD_ONE_DIR}" lxc storage delete "${poolName}"
```
simondeziel, line 80: This fails in CI due to still having content on the pool. I guess those are the test images that need to be deleted.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -629,9 +629,18 @@ func (d *qemu) onStop(target string) error {
...
+		// The server is running on the same host in which case it is not possible to unmount
+		// the source device because it is already used by the destination (migrated) instance.
+		isLiveMigrationTest := shared.IsTrue(os.Getenv("LXD_LIVE_MIGRATION_TEST_ON_THE_SAME_HOST"))
+		if !isLiveMigrationTest || op.Action() != operationlock.ActionMigrate || !errors.Is(err, storageDrivers.ErrResourceBusy) {
+			err = fmt.Errorf("Failed unmounting instance: %w", err)
+			op.Done(err)
+			return err
+		}
+
+		d.logger.Warn("Failed unmounting source instance during migration", logger.Ctx{"err": err})
```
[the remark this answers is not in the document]
MusicDin: Agree. But then linter is unhappy because if `else` contains return, the condition can be reverted 😕

test/suites/live_migration.sh
```diff
@@ -0,0 +1,88 @@
...
+  fi
+
+  # Spawn the first node and bootstrap the cluster.
+  LXD_LIVE_MIGRATION_TEST_ON_THE_SAME_HOST=true spawn_lxd_and_bootstrap_cluster "${poolDriver}"
+
+  local cert
+  cert="$(cert_to_yaml "${LXD_ONE_DIR}/cluster.crt")"
+
+  # Spawn a second node.
+  LXD_LIVE_MIGRATION_TEST_ON_THE_SAME_HOST=true spawn_lxd_and_join_cluster "${cert}" 2 1 "${LXD_ONE_DIR}" "${poolDriver}"
```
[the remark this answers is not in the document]
MusicDin: The env is correctly set in the test and parsed in the driver. However the comment in the driver indeed contains wrong env var name.

Will change the env name to start with `LXD_TEST` prefix.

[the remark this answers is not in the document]
MusicDin: Same as above.

The env is correctly set in the test and parsed in the driver. However the comment in the driver indeed contains wrong env var name.

Will change the env name to start with LXD_TEST prefix.

lxd/storage/drivers/errors.go
```diff
@@ -17,6 +17,10 @@ var ErrCannotBeShrunk = errors.New("Cannot be shrunk")
 // ErrInUse indicates operation cannot proceed as resource is in use.
 var ErrInUse = errors.New("In use")
 
+// ErrResourceBusy indicates that a device or resource is still busy, while from the perspective of LXD
+// it should not be. For example, when trying to unmap a device that is still in use by another process.
```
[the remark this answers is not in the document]
MusicDin: This is sentinel error and does not wrap anything. It represents EBUSY error that can be asserted outside the package, but I wanted to emphasize that compared to ErrInUse, the LXD *expected* the resource to be free, but it is not.

lxd/instance/drivers/driver_qemu.go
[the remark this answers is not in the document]
MusicDin: Added //nolint to make if statement clearer.

test/suites/live_migration.sh
MusicDin: Fixed

test/suites/live_migration.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
+
+# test_clustering_live_migration spawns a 2-node LXD cluster, creates a virtual machine on top of it,
+# creates and attaches a block volume to the virtual machine, writes some arbitrary data to the volume,
+# and runs live migration. Success is determined by the data being intact after live migration.
+test_clustering_live_migration() {
+  poolDriver=$(lxc storage show "$(lxc profile device get default root pool)" | awk '/^driver:/ {print $2}')
+  if [ "${poolDriver:-}" != "ceph" ]; then
+    export TEST_UNMET_REQUIREMENT="Only supports 'ceph', not ${poolDriver}"
```
tomponline, line 9: why arent we testing live migration for other drivers?

tomponline: this is a good start ofc, but wondered why we didnt go for all drivers?

lxd/storage/drivers/driver_ceph_utils.go
```diff
@@ -238,7 +238,7 @@ again:
 					// EBUSY (currently in use).
 					busyCount++
 					if busyCount == 10 {
-						return err
+						return fmt.Errorf("%v: %w", err, ErrResourceBusy)
```
tomponline: Could we detect the env var here and log a warning and return nil after 10 iterations instead of having the handling of this special case in the qemu driver?

test/suites/live_migration.sh
MusicDin: I've started only with Ceph as this test validate custom volume live migration as well which is only supported for remote storage drivers.

lxd/storage/drivers/driver_ceph_utils.go
MusicDin: Yep, this is how it was done initially, but then we cannot check whether it is part of the instance migration operation. I mean, the environment variable is quite specific so there should be no confusion, but if enabled it would ignore every EBUSY error from Ceph unmap.

tomponline: Understood, in that case can we use ` api.StatusErrorf(http.StatusLocked, "%w", err)` instead of the new error here, as we've used that in some places already when something is in use.

MusicDin: Thanks. Will update

test/suites/live_migration.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
+
+# test_clustering_live_migration spawns a 2-node LXD cluster, creates a virtual machine on top of it,
+# creates and attaches a block volume to the virtual machine, writes some arbitrary data to the volume,
+# and runs live migration. Success is determined by the data being intact after live migration.
+test_clustering_live_migration() {
+  poolDriver=$(lxc storage show "$(lxc profile device get default root pool)" | awk '/^driver:/ {print $2}')
```
simondeziel:
```suggestion
  poolDriver="$(storage_backend "${LXD_INITIAL_DIR}")"
```

test/suites/live_migration.sh
```diff
@@ -0,0 +1,96 @@
...
+  LXD_DIR=${LXD_ONE_DIR} lxc auth group create live-migration
+  LXD_DIR=${LXD_ONE_DIR} lxc auth group permission add live-migration server admin
+
+  token="$(LXD_DIR=${LXD_ONE_DIR} lxc auth identity create tls/live-migration --group=live-migration --quiet)"
+  LXD_DIR="${LXD_ONE_DIR}" lxc remote add cls 100.64.1.101:8443 --token="${token}"
+
+  LXD_DIR="${LXD_ONE_DIR}" ensure_import_ubuntu_vm_image
+
+  # Storage pool created when spawning LXD cluster is "data".
+  poolName=$(LXD_DIR="${LXD_ONE_DIR}" lxc profile device get default root pool)
```
simondeziel: There are many other places in `test/suites/clustering.sh` where `data` is hardcoded so it would be fine by me if we did it here too.

test/suites/live_migration.sh
```diff
@@ -0,0 +1,96 @@
...
+  LXD_DIR="${LXD_ONE_DIR}" lxc storage set "${poolName}" volume.size="${SMALLEST_VM_ROOT_DISK}"
+
+  # Initialize the VM.
+  LXD_DIR="${LXD_ONE_DIR}" lxc init ubuntu-vm vm \
+    --vm \
+    --config limits.cpu=2 \
+    --config limits.memory=768MiB \
+    --config migration.stateful=true \
+    --device root,size="${SMALLEST_VM_ROOT_DISK}" \
+    --config security.devlxd=false \
```
simondeziel: Maybe put a comment as to why `security.devlxd` needs to be disabled, it's not obvious to me at least.

MusicDin: There is no need to disable it. Thanks for spotting this.

simondeziel:
Congrats, it's a nice one to finally have tested regularly!

PR #17375, opened 2026-01-13
doc: simplify first steps tutorial to be more straight forward

edlerd:
## Checklist

- removed the `snap version` command as it is not needed and confusing as local output might differ and it is not needed for the process
- removed the "check if user is already in the group" as that is not needed and confusing the user
- removed the "Confirm virtualization support" as it is not needed to complete the tutorial to get started with lxd.
- made the command for lxd listening on the network compatible with more environments, not assuming a localhost ipv6 address.

### DIFF
--- a/doc/tutorial/first_steps.md
+++ b/doc/tutorial/first_steps.md
@@ -22,27 +22,7 @@ The MicroCloud setup process installs and initializes LXD. Thus, you can skip th
 (tutorial-install)=
 ## Install LXD using snap
 
-This section of the tutorial assumes that you have the `snap` packaging system available on your system, which is the recommended way to install and update LXD. To confirm whether it is available on your system, run:
-
-```bash
-snap version
-```
-
-If `snap` is installed, you will see output similar to this:
-
-```{terminal}
-:user: your-user
-:host: host-system
-
-snap version
-
-snap          2.72
-snapd         2.72
-series        16
-ubuntu        24.04
-kernel        6.14.0-33-generic
-architecture  amd64
-```
+This section of the tutorial assumes that you have the `snap` packaging system available on your system, which is the recommended way to install and update LXD.
 
 To install LXD using `snap`, run:
 
@@ -89,15 +69,7 @@ For more information, see {ref}`security-daemon-access`.
 
 Installing LXD through its snap should automatically create an `lxd` group on your system. The user you are logged in as must be in this group to interact with LXD.
 
-Check to see if the current user is already in that group:
-
-```bash
-getent group lxd | grep "$USER"
-```
-
-If this command returns a result, you're set up correctly and can continue with the next section.
-
-If there is no result, enter the following commands. The first command adds your user to the `lxd` group.
+Enter the following commands. The first command adds your user to the `lxd` group.
 The second command starts a new shell where the group membership takes effect immediately.
 
 ```bash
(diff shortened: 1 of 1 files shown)


doc/tutorial/first_steps.md
```diff
@@ -166,13 +112,7 @@ If you prefer to use the LXD UI, expand and follow the steps below.
...
 
 ```bash
-lxc config set core.https_address [::1]:8443
-```
-
-Confirm that the `core.https_address` is set to `[::1]:8443`:
-
-```bash
-lxc config get core.https_address
+lxc config set core.https_address :8443
```
tomponline: This is exposing LXD to all network interfaces on the host. 

Which is at odds with the statement above:

>  We will use the local network by configuring this to the IPv6 loopback address `[::1]` and port 8443.


I would suggest we stick with `[::1]` (the default port is 8443) and then instruct the user to open a link to `https://[::1]:8443`

tomponline: @edlerd did `https://localhost:8443` not work?

edlerd: The problem is with systems that do not have ipv6 enabled. They run into an error.

tomponline: you dont need ipv6 connectivity to access the ::1 address, and ipv6 has been enabled by default on Ubuntu for a long time.

edlerd: We just ran into an error on this command in a user test. (Sorry for the low resolution. This is a screenshot from the recording).

image

After we switched the command to `lxc config set core.https_address 127.0.0.1:8443` it worked fine.

tomponline: right i see, if we want to support legacy protocols then we have no option to use 127.0.0.1 sadly.

edlerd: Changed it to `127.0.0.1:8443`.

doc/tutorial/first_steps.md
```diff
@@ -22,27 +22,7 @@ The MicroCloud setup process installs and initializes LXD. Thus, you can skip th
...
-snap version
-
-snap          2.72
-snapd         2.72
-series        16
-ubuntu        24.04
-kernel        6.14.0-33-generic
-architecture  amd64
-```
+This section of the tutorial assumes that you have the `snap` packaging system available on your system, which is the recommended way to install and update LXD.
```
minaelee, line 25: I'm not against simplifying this section. However, I suggest leaving in the `snap version` check to see if snap is installed, then in place of the expected output that was deleted, providing a link to https://snapcraft.io/docs/installing-snapd for the user in case snap is not installed.

doc/tutorial/first_steps.md
```diff
@@ -166,13 +112,7 @@ If you prefer to use the LXD UI, expand and follow the steps below.
 By default, LXD is exposed through a Unix socket only and is not accessible over HTTPS. To access and manage LXD through a web browser using HTTPS, we must set the {config:option}`server-core:core.https_address` server configuration option. We will use the local network by configuring this to the IPv6 loopback address `[::1]` and port 8443. Run:
 
 ```bash
-lxc config set core.https_address [::1]:8443
-```
-
-Confirm that the `core.https_address` is set to `[::1]:8443`:
```
minaelee, line 172: The confirmation step was requested by a previous UX tester because otherwise there is no output to confirm whether it was successful. While we don't provide this step for all commands without output, doing it once here is helpful as it demonstrates the use of `lxc config get`.

doc/tutorial/first_steps.md
```diff
@@ -126,34 +98,8 @@ If you do not see any message when running this command, that means it has succe
 The `lxd init` command can be run again later to update the options. Once you have learned more about LXD, you might want to tune the {ref}`initialization options <initialize>` according to your own preferences, or learn how to {ref}`use a preseed file <initialize-preseed>` to initialize LXD from a saved set of options. For now, the minimal configuration is sufficient.
 ```
 
-(tutorial-virtualization)=
```
minaelee, line 129: I agree with removing this from the tutorial, but not removing it from the docs completely. Let's go ahead and remove it in this PR, then in a separate PR after this one, I can re-add it to the how-to guide about creating VMs and link to that from the tutorial section that creates a VM.

doc/tutorial/first_steps.md
```diff
@@ -89,15 +69,7 @@ For more information, see {ref}`security-daemon-access`.
...
-Check to see if the current user is already in that group:
-
-```bash
-getent group lxd | grep "$USER"
-```
-
-If this command returns a result, you're set up correctly and can continue with the next section.
-
-If there is no result, enter the following commands. The first command adds your user to the `lxd` group.
+Enter the following commands. The first command adds your user to the `lxd` group.
```
minaelee, line 72: I agree with simplifying this section. Even if the user is already in the LXD group, it doesn't hurt to add them again, and this provides a single path for users to follow.

doc/tutorial/first_steps.md
edlerd: Isn't it simpler to let the user run the `snap install` directly? Then handle the possible error case in some text below the command. I don't understand why we let them run `snap version` first.

We observed a user who took a long time figuring out if and how to run each command. So it seems a good idea to minimize the commands needed to finish the tutorial.

doc/tutorial/first_steps.md
edlerd: I see, will bring this back.

doc/tutorial/first_steps.md
edlerd: Yes, in our user test, the user ran the "check" command. They interpreted wrongly, that they were already part of the group. This was leading to some pain points later on.

So asking them to run the group add commands directly avoids this risk and friction.

doc/tutorial/first_steps.md
edlerd: I see, we might want to leave a note here linking to the other section then. Do you have a suggestion where to move this section to. Wdyt?

doc/tutorial/first_steps.md
minaelee: Ah, there's already an admonition beneath the `snap install` with the link. Since that is the case, I see no issue with removing this.

doc/tutorial/first_steps.md
minaelee: I can handle adding back this section elsewhere then linking to it from this tutorial after this PR is merged.

minaelee:
Is this draft still in progress?

edlerd:
This is ready to merge from my side.

tomponline (approved):

PR #17377, opened 2026-01-13
Refactor rebalanceMemberRoles and rename Rebalance

kadinsayani:


lxd/api_cluster.go
```diff
@@ -2684,44 +2684,49 @@ func rebalanceMemberRoles(ctx context.Context, s *state.State, gateway *cluster.
...
-	if err != nil {
-		return err
-	}
-
-	if address == "" {
-		// Nothing to do.
-		return nil
-	}
+	for {
+		address, nodes, err := cluster.GetNextRoleChange(s, gateway, unavailableMembers)
```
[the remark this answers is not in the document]
kadinsayani: We have a base case below the call to `GetNextRoleChange`:

```Go
if address == "" {
	// Nothing to do.
	return nil
}
```

lxd/api_cluster.go
```diff
@@ -2684,44 +2684,49 @@ func rebalanceMemberRoles(ctx context.Context, s *state.State, gateway *cluster.
...
+		for _, node := range nodes {
+			if node.Address != address || node.Role != db.RaftSpare {
+				continue
+			}
+
+			if cluster.HasConnectivity(s.Endpoints.NetworkCert(), s.ServerCert(), address) {
+				break
+			}
+
+			logger.Info("Demoting offline member during rebalance", logger.Ctx{"candidateAddress": node.Address})
```
tomponline, line 2709: An interesting thing we should look at in the future is whether the offline concept in `RoleChanges.Adjust` 

```
offlineVoters := c.list(client.Voter, false, nil)
```

Would suffice without needing this additional check.

Also, `newRolesChanges` uses `HasConnectivity` as well, so really we should be using the data gathered from that pass to make decisions here, rather than connecting to them again.

kadinsayani: Follow-up PR here: https://github.com/canonical/lxd/pull/17386.

PR #17379, opened 2026-01-14
Hot-plugging for CDI GPU devices

dmitry-lyfar:
There are two possible options to add a GPU device:

1. On boot. In this case we are required to modify the
linker's configuration at a predictable time. This is needed
to avoid issues with clients who want to run some GPU
workload on start as a systemd service, for example. Currently,
this is done by the lxc.hook.mount hook which
creates configuration (symlinks, linker's paths) and runs
ldconfig to update cache.

2. On a hotplug request. It happens when the instance is already
running. The configuration and cache update would run in
the LXD PostRun hooks, executed after the device's attachment.

Possible scenarios
==================

Not an exhaustive list.

- Container stopped -> GPU added (or updated) -> Container started
- Container stopped -> GPU removed -> Container started
- Container running -> GPU added -> Container restarted
- Container running -> GPU added (id=0) -> GPU removed (id=0) -> GPU added (all) -> Container restarted
- Container running -> GPU added (id=0) -> GPU added (id=1) -> GPU removed (id=0) -> Container restared

Invariants:
- Always correct linker cache
- Always correct bind mounts
- Always correct symlinks
- ldconfig is run from within the container
- No unused json specs are left in the devices directory
- No unused json hooks are left in the devices directory
- No unused cached CDI logs are left in the container's logs directory

Current implementation
======================

The sequence of events for:

Starting a GPU device (boot time):

```
lxc.start -> startCommon -> startDevice (char devices + bind mounts; calls to CDI) -> forkstart -> ... (lxc.hook.mount to setup linker and symlinks) ... -> forkstart -> runHooks
```

Changing the GPU device configuration for an instance. The instance is NOT running:

- Adding a GPU device:

```
inst.Update -> lxc.Update -> d.deviceLoad (validation checks for new or updated devices) -> gpuPhysical.Add (empty)
```

- Updating a GPU device (not supported by GPUs, these are always `remove -> add` if changed):

```
inst.Update -> lxc.Update -> d.deviceLoad (validation checks for new or updated devices) -> gpuPhysical.Remove (empty) -> gpuPhysical.Add (empty)
```

- Removing a GPU device:

```
inst.Update -> lxc.Update -> d.deviceLoad (validation fails are ignored to facilitate removal) -> gpuPhysical.Remove (empty, no postStop hooks cleanup here; it's cleaned on the container's start)
```

Proposed changes
================

Starting a GPU device (boot time):

```
lxc.start -> startCommon -> startDevice (char devices + bind mounts; calls to CDI) -> forkstart -> ... (lxc.hook.mount to setup linker and symlinks; no cache update) ... -> forkstart -> runHooks  (update linker cache via exec)
```

Adding a GPU device. The instance IS running:

```
lxc.Update -> d.devicesUpdate -> d.deviceStart -> startDevice (char devices + bind mounts; calls to CDI) -> gpuPhysical.Start -> runHooks (create symlinks and update linker config, update linker via exec)
```

Issues
======

1. If container is stopped and GPU is removed, what happens to the linker's configuration? symlinks?

A: these are not getting removed, leaving symlinks, configuration, and empty files for the bind-mounts. It may
trigger false positives in tests that add / remove GPU device for a single instance. The issue is reported here: https://github.com/canonical/lxd/issues/17877

2. Current implementation removes ldconfig cache manually. It is unnecessary and, potentially, dangerous as it may corrupt ldconfig cache if called concurrently. I have separated `ldconfig` update to be executed via lxc exec in post hooks and removed the cache removal logic. This call may fail (e.g. no `ldconfig`) with a warning. Ideally, we should not call `ldconfig` manually for the cases when a device was added to a stopped container due to the modern systemd's one-shot service, see `systemctl cat ldconfig.service`. It might be useful for hotplugging, but the linker's cache feels like something that LXD should not concern itself with at the device level.

3. What happens to the bind mounts and (possibly outdated) symlinks of the running containers when the userspace is updated on the host?

A: currently it is not handled. The user would have to restart the container for such update to work in the nvidia's case.

## Checklist

### DIFF
--- a/lxd/device/cdi/hooks.go
+++ b/lxd/device/cdi/hooks.go
@@ -156,6 +156,12 @@ func ApplyHooksToContainer(hooksFilePath string, containerRootFS string) error {
 				existingLinkerEntries[strings.TrimSpace(scanner.Text())] = true
 			}
 
+			err = scanner.Err()
+			if err != nil {
+				ldConfFile.Close()
+				return fmt.Errorf("Failed reading the linker conf file at %q: %w", ldConfFilePath, err)
+			}
+
 			for _, update := range hooks.LDCacheUpdates {
 				if !existingLinkerEntries[update] {
 					_, err = fmt.Fprintln(ldConfFile, update)
--- a/lxd/device/gpu_physical.go
+++ b/lxd/device/gpu_physical.go
@@ -394,6 +394,24 @@ func (d *gpuPhysical) startContainer() (*deviceConfig.RunConfig, error) {
 					{Key: cdi.CDIHookDefinitionKey, Value: filepath.Base(hooksFile)},
 				}...)
 
+			if d.inst.IsRunning() {
+				// If the device is being hotplugged, we need to apply the hooks
+				// immediately as the container is already running and won't have the
+				// chance to run the `lxc.hook.mount` hooks. We also update the ldconfig
+				// cache as there is a chance that some CDI listed libraries are not in
+				// the standard paths.
+				runConf.PostHooks = append(runConf.PostHooks, func() error {
+					err := cdi.ApplyHooksToContainer(hooksFile, d.inst.RootfsPath())
+					if err != nil {
+						return err
+					}
+
+					// For hotplug, ldconfig is called immediately after device attachment.
+					cdi.UpdateLDCache(d.inst)
+					return nil
+				})
+			}
+
 			return &runConf, nil
 		}
 	}
@@ -683,26 +701,9 @@ func (d *gpuPhysical) stopCDIDevices(configDevices cdi.ConfigDevices, runConf *d
 }
 
 // CanHotPlug returns whether the device can be managed whilst the instance is running.
-// CDI GPU are not hotpluggable because the configuration of a CDI GPU requires a LXC hook that
-// is only run at instance start. A classic GPU device can be hotplugged.
+// Both CDI and classic GPU devices can be hotplugged for containers.
 func (d *gpuPhysical) CanHotPlug() bool {
-	if d.inst.Type() != instancetype.Container {
-		return false
-	}
-
-	if d.config["id"] == "" {
-		return true
-	}
-
-	// Check if the id of the device matches a CDI format.
-	cdiID, _ := cdi.ToCDI(d.config["id"])
-	if cdiID != nil {
-		// CDI devices cannot be hot-plugged because they rely on a start hook for setting
-		// up files inside the container.
-		return false
-	}
-
-	return true
+	return d.inst.Type() == instancetype.Container
 }
 
 // Stop is run when the device is removed from the instance.
--- a/test/suites/container_devices_gpu.sh
+++ b/test/suites/container_devices_gpu.sh
@@ -114,6 +114,17 @@ gpu_verify_nvidia_mounts() {
   lxc exec "${ctName}" -- mount | grep -wF /dev/nvidia-uvm
   lxc exec "${ctName}" -- mount | grep -wF /dev/nvidia-uvm-tools
   lxc exec "${ctName}" -- mount | grep -wF /dev/nvidiactl
+
+  # Verify ldconfig configuration exists
+  lxc exec "${ctName}" -- test -f /etc/ld.so.conf.d/00-lxdcdi.conf
+
+  # Verify the CDI library paths are in the config
+  lxc exec "${ctName}" -- grep -q "/usr/lib" /etc/ld.so.conf.d/00-lxdcdi.conf
+
+  # Verify key NVIDIA libraries are accessible (mounted via CDI)
+  lxc exec "${ctName}" -- test -f /usr/lib/x86_64-linux-gnu/libcuda.so.1 || \
+    lxc exec "${ctName}" -- test -f /usr/lib64/libcuda.so.1 || \
+    lxc exec "${ctName}" -- test -f /usr/lib/libcuda.so.1
 }
 
 gpu_run_nvidia_tests() {
@@ -133,26 +144,23 @@ gpu_run_nvidia_tests() {
     return
   fi
 
-  # Check the Nvidia specific devices are mounted correctly.
+  # The instance is stopped.
   lxc stop -f "${ctName}"
+  lxc config device add "${ctName}" gpu-cdi-stopped gpu id=nvidia.com/gpu=0
+  lxc start "${ctName}"
+  gpu_verify_nvidia_mounts "${ctName}"
+  lxc stop -f "${ctName}"
+  lxc config device remove "${ctName}" gpu-cdi-stopped
+  lxc start "${ctName}"
+
+  # The instance is running (hot plugging).
   lxc config device add "${ctName}" gpu-cdi gpu id=nvidia.com/gpu=0
-  lxc start "${ctName}"
-
   gpu_verify_nvidia_mounts "${ctName}"
-
-  lxc stop -f "${ctName}"
   lxc config device remove "${ctName}" gpu-cdi
-  lxc start "${ctName}"
 
-  lxc stop -f "${ctName}"
-  lxc config device add "${ctName}" gpu-cdi gpu id=nvidia.com/gpu=all
-  lxc start "${ctName}"
-
+  lxc config device add "${ctName}" gpu-cdi-all gpu id=nvidia.com/gpu=all
   gpu_verify_nvidia_mounts "${ctName}"
-
-  lxc stop -f "${ctName}"
-  lxc config device remove "${ctName}" gpu-cdi
-  lxc start "${ctName}"
+  lxc config device remove "${ctName}" gpu-cdi-all
 }
 
 gpu_run_amd_tests() {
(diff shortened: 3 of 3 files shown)


lxd/device/cdi/hooks.go
tomponline, line 1: Thanks @dmitry-lyfar would you mind breaking this commit up into several smaller commits, one for each logical change, as it helps to see the changes being made.

I suggest, if possible, to break them down as 1 commit per line in the detailed commit message.

lxd/device/cdi/hooks.go
```diff
@@ -42,3 +57,156 @@ type ConfigDevices struct {
...
+// UpdateLDCache updates the linker cache inside the instance.
+func UpdateLDCache(inst instance.Instance) {
+	l := logger.AddContext(logger.Ctx{"project": inst.Project().Name, "instance": inst.Name()})
+	cmd, err := inst.Exec(api.InstanceExecPost{
+		Command:   []string{"/sbin/ldconfig"},
+		WaitForWS: false,
+	}, nil, nil, nil)
+
+	if err != nil {
+		l.Warn("Failed executing ldconfig in the container", logger.Ctx{"error": err})
```
tomponline: To aid with identifying which step failed, lets log a different message here, how about:

> Failed starting ldconfig in the container

lxd/device/gpu_physical.go
```diff
@@ -398,6 +398,18 @@ func (d *gpuPhysical) startContainer() (*deviceConfig.RunConfig, error) {
...
 				}...)
 
+			if d.inst.IsRunning() {
+				runConf.PostHooks = append(runConf.PostHooks, func() error {
+					// Container is already running (hotplug), apply hooks directly
+					return cdi.ApplyHooksToContainer(hooksFile, d.inst.RootfsPath())
+				})
+			}
+
+			runConf.PostHooks = append(runConf.PostHooks, func() error {
```
tomponline: I'd appreciate a comment here explaining that this is intended for both hotplug and boot time, and explain that at boot time the `ldconfig` will be called during the instance's boot process.

lxd/device/cdi/hooks.go
```diff
@@ -42,3 +57,156 @@ type ConfigDevices struct {
...
+
+const (
+	// customCDILinkerConfFile is the name of the linker conf file we will write to
+	// inside the container. The `00-lxdcdi` prefix is chosen to ensure that these libraries have
+	// a higher precedence than other libraries on the system.
+	customCDILinkerConfFile = "00-lxdcdi.conf"
+)
+
+// resolveTargetRelativeToLink converts a link's target into a path relative to the link's path.
+func resolveTargetRelativeToLink(link string, target string) (string, error) {
```
tomponline, line 69: As this function doesn't resolve the symlink's target specifically, can we rename this to `cleanTargetRelativeToLink`, as "resolve" suggests its resolving the symlink (which when run from inside the `lxc.mount.hook` before `pivot_root` is concerning).

lxd/device/cdi/hooks.go
```diff
@@ -42,3 +57,156 @@ type ConfigDevices struct {
...
+	relPath, err := filepath.Rel(linkDir, targetClean)
+	if err != nil {
+		return "", err
+	}
+
+	return relPath, nil
+}
+
+// ApplyHooksToContainer applies CDI hooks to a container by creating symlinks
+// and updating the linker cache. This function can be called both during
```
tomponline: > updating the linker cache

It doesn't do this anymore right?

lxd/device/cdi/hooks.go
```diff
@@ -42,3 +57,156 @@ type ConfigDevices struct {
...
+		// Create the symlink
+		err = os.Symlink(target, filepath.Join(containerRootFS, symlink.Link))
+		if err != nil {
+			if !errors.Is(err, fs.ErrExist) {
+				return fmt.Errorf("Failed creating the CDI symlink: %w", err)
+			}
+		}
+	}
+
+	// Updating the linker cache
```
tomponline: This comment isn't true anymore.

tomponline:
Please can this be rebased. Ta

lxd/device/cdi/hooks.go
dmitry-lyfar: I can rename it, but this is the original name I left for easier refactoring and, probably, backporting. Lmk.

tomponline:
> @tomponline I've made updates as per the comments except for making the commits more detailed as I think the refactoring commit is a single logical change about `ApplyHooksToContainer`. See the corresponding commit message. Let me know how you prefer it in the codebase. I'd probably consider adding unit tests for the hooks logic as a separate commit as there was none. That would allow us to reason about the application of CDI hook(s) logic correctness and still rely on this refactoring. It is only now tested in the integration tests for CDI and indirectly.
> 
> Also, confirmed that we should rely on the one-shot service from systemd (we should extend our lxd-ci with that set of tests and move some of the e2e tests from to that pipeline, I'll do it separately in that repository).

I think the moving of ApplyHooksToContainer should be its own commit, before the refactoring, that way the move can be backported without any functional changes. This keeps the code more aligned for future fixes.

lxd/device/cdi/hooks.go
```diff
@@ -42,3 +57,156 @@ type ConfigDevices struct {
...
+			ldConfFile.Close()
+		} else {
+			return fmt.Errorf("Could not stat the linker conf file to add CDI linker entries at %q: %w", ldConfFilePath, err)
+		}
+	}
+	return nil
+}
+
+// UpdateLDCache updates the linker cache inside the instance.
+func UpdateLDCache(inst instance.Instance) {
```
tomponline, line 203: I thought we would not update ld cache anymore?

tomponline:
> Also, confirmed that we should rely on the one-shot service from systemd (we should extend our lxd-ci with that set of tests and move some of the e2e tests from to that pipeline, I'll do it separately in that repository).

Excellent thanks!

test/suites/container_devices_gpu.sh
```diff
@@ -133,26 +144,23 @@ gpu_run_nvidia_tests() {
...
   gpu_verify_nvidia_mounts "${ctName}"
-
   lxc stop -f "${ctName}"
-  lxc config device remove "${ctName}" gpu-cdi
-  lxc start "${ctName}"
-
-  lxc stop -f "${ctName}"
-  lxc config device add "${ctName}" gpu-cdi gpu id=nvidia.com/gpu=all
+  lxc config device remove "${ctName}" gpu-cdi-stopped
   lxc start "${ctName}"
```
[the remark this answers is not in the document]
dmitry-lyfar: Yes, check this https://github.com/canonical/lxd/issues/17877

lxd/device/cdi/hooks.go
```diff
@@ -42,3 +57,159 @@ type ConfigDevices struct {
...
+		if err != nil {
+			return fmt.Errorf("Failed creating the directory for the CDI symlink: %w", err)
+		}
+
+		// Create the symlink
+		err = os.Symlink(target, filepath.Join(containerRootFS, symlink.Link))
+		if err != nil {
+			if !errors.Is(err, fs.ErrExist) {
+				return fmt.Errorf("Failed creating the CDI symlink: %w", err)
+			}
```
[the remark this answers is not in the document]
tomponline: @dmitry-lyfar this is something worth considering, but not for this PR.

lxd/device/cdi/hooks.go
```diff
@@ -42,3 +57,159 @@ type ConfigDevices struct {
...
+			if err != nil {
+				return fmt.Errorf("Failed opening the ld.so.conf file at %q: %w", ldConfFilePath, err)
+			}
+
+			existingLinkerEntries := make(map[string]bool)
+			scanner := bufio.NewScanner(ldConfFile)
+			for scanner.Scan() {
+				existingLinkerEntries[strings.TrimSpace(scanner.Text())] = true
+			}
+
```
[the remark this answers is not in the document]
tomponline: @dmitry-lyfar this seems like a good thing to address in a follow up PR?

lxd/main_callhook.go
```diff
@@ -239,14 +66,36 @@ func (c *cmdCallhook) Run(cmd *cobra.Command, args []string) error {
...
 			if err != nil {
 				return err
 			}
 		}
 
+		if len(cdiHooksFiles) > 0 {
+			// NOTE: LXD does need to call ldconfig after applying the hooks here, as
+			// the instance is not running and we don't want to run the host's
+			// ldconfig in the container's mount namespace. systemd (starting from
+			// v215) includes a one-shot service that runs ldconfig at boot at a
```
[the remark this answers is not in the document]
tomponline: @dmitry-lyfar something for the follow up PR please

lxd/device/cdi/hooks.go
tomponline, line 1: There's some changes in the `lxd/device/cdi: refactor: add unit tests to ApplyHooksToContainer` that are not adding tests, and should be in the prior commit please.

lxd/device/gpu_physical.go
```diff
@@ -394,6 +394,23 @@ func (d *gpuPhysical) startContainer() (*deviceConfig.RunConfig, error) {
...
+				// immediately as the container is already running and won't have the
+				// chance to run the `lxc.hook.mount` hooks. We also update the ldconfig
+				// cache as there is a chance that some CDI listed libraries are not in
+				// the standard paths.
+				runConf.PostHooks = append(runConf.PostHooks, func() error {
+					err := cdi.ApplyHooksToContainer(hooksFile, d.inst.RootfsPath())
+					if err != nil {
+						return err
+					}
+					// For hotplug, ldconfig is called immediately after device attachment.
```
tomponline, line 409: new line before comment here please

lxd/device/cdi/hooks.go
```diff
@@ -42,3 +57,159 @@ type ConfigDevices struct {
...
+		return "", err
+	}
+
+	return relPath, nil
+}
+
+// ApplyHooksToContainer applies CDI hooks to a container by creating symlinks
+// and updating the linker configuration. This function can be called both
+// during container start (from LXC hook) and during hotplug.
+func ApplyHooksToContainer(hooksFilePath string, containerRootFS string) error {
```
tomponline, line 97: @dmitry-lyfar I've been thinking more about this function, now that it is being called during hot-plug from the host's mount namespace and operating on the instance's root mount (running as root).

This is unlike previously where this function was being called only from liblxc's `lxc.hook.mount` hook which was run inside the container's namespace (albeit before pivot root so still had access outside of the container's root):

> A hook to be run in the container's namespace after mounting has been done, but before the pivot_root. 

My concern is that every file operation inside this function, like `os.MkdirAll` or `os.Symlink`, could be tricked into operating on the host's filesystem by an existing malicious symlink in the container's root fs, which Go will follow outside of its root path during file operations.

I think we can, and should, do better from a security standpoint (for both hotplug and boot time calls of this function).

Go 1.24 added support for `os.Root` https://go.dev/blog/osroot

> Root provides methods to operate on files within the root. These methods all accept filenames relative to the root, and disallow any operations that would escape from the root either using relative path components ("..") or symlinks.

And we've made use of it already in LXD, with the `recursivePullFile` function.

I think this function should take, rather than `containerRootFS string`, a `containerRoot os.Root` argument.

Then it will be forced to use that `os.Root` when making file changes in the isntance's root fs:

Go 1.25 seems to support everything we need:

https://pkg.go.dev/os#Root

root.MkdirAll
root.Symlink

tomponline: The other thing I've been considering, in addition to using os.Root, is whether, for hot-plugging, we should be using `d.inst.RootfsPath()` (which is the path to the instance's root fs mount) or whether we should be using `/proc/<d.inst.InitPID()>/root` which represents the view of the instance's filesystem from the container's init process (including any subsequent mounts that may have been made inside the container).

On the one hand, using `/proc/<PID>/root` would better reflect the state of the running instance's view of the root filesystem (and so symlinks will be created atop of other mounts if needed), however on the other hand using `d.inst.RootfsPath()`, despite its shortcomings, will represent the same behaviour as the boot-time `lxc.mount.hook` behaviour.

tomponline: OK, thinking even more about this, we also need to think about userns ID mapping, because whereas before when this was running in `lxc.mount.hook` (which was run inside the container's namespaces as I understand it) that means it would have taken on the container's user ID mappings when creating directories and links, now we are running it on the host as part of the post hooks during hotplugging, this ID mapping will not occur.

For containers that are not using idmapped mounts, this might cause issues with UID/GIDs not being correct when viewed inside the container.

So here's what I think we should do:

1. For boot time scenarios `ApplyHooksToContainer` should use `os.Root` when run from the `lxc.mount.hook` - this is a belt and braces security improvement, because even though the hook is run inside the container's mount namespace, it is before the pivot_root, so still (necessarily) has access to the host's filesystem.
2. For hot-plug scenarios `ApplyHooksToContainer` should use `d.inst.FileSFTP()` to get an SFTP client directly into the container (this is the same functionality we use for `lxc file` operations).

Using `inst.FileSFTP()` won't be as performant as directly modifying the files, but it will be the most safe, because under the hood it will use `forkfile` to attach to the running container's namespaces:

https://github.com/canonical/lxd/blob/main/lxd/main_forkfile.go#L104-L124

tomponline: Using `inst.FileSFTP()` is actually kind of interesting, because it works for both running and stopped containers...

Perhaps this means we can use during the device start function, and not even need to use post hooks. That way if the device is being started as part of an instance boot (so the instance isn't running yet), then it would directly modify the container's filesystem (inside the `forkfile`'s chroot that it sets up).

This would handle idmappings and malicious symlinks.

tomponline: This combined with the removal of the boot-time ldconfig requirements would mean we wouldn't need to use the `lxc.mount.hook` at all, and have a common implementation for both boot time and hot-plug starting.

lxd/device/cdi/hooks.go
```diff
@@ -42,3 +57,159 @@ type ConfigDevices struct {
...
+		// Create the symlink
+		err = os.Symlink(target, filepath.Join(containerRootFS, symlink.Link))
+		if err != nil {
+			if !errors.Is(err, fs.ErrExist) {
+				return fmt.Errorf("Failed creating the CDI symlink: %w", err)
+			}
+		}
+	}
+
+	// Updating the linker configuration.
```
simondeziel, line 134: I know this bit of code is mostly just being moved but I see potential for reducing code duplication and improved consistency. Anyway, just a drive-by comment as this likely won't go in this PR.

lxd/device/cdi/hooks.go
dmitry-lyfar: @tomponline it's a good idea to use SFTP, however, not possible for the scenario when `deviceStart` is called during the boot, as SFTP connection function hangs on the operation waiting and deadlocks because of that. Apparently, SFTP connection cannot be opened when the container is starting even if I try it in the post hooks as hooks are also run before the operation is marked as finished.

A possible option for the stopped instance could be applying symlink and configuration changes in `gpuPhysical.Add()`. But there is a caveat, bind mounts for the user space libs do not exist at that point resulting in symlinks being broken (which will be fixed once the container starts). It might be a compromise to explore, but it feels messy.  I changes this PR to demonstrate using sftp.

I'll give it more thinking tomorrow.

dmitry-lyfar: No, actually, moving it to `Add` does not make any difference because sftp connection won't be created as the instance is not running:

```
	if op.Action() != operationlock.ActionUpdate || !d.IsRunning() {
		_ = op.Wait(context.Background())
	}
```

tomponline: We may be able to uncouple the public api which uses locking from the intetnal api that can be used during device start and stop. I suggest i merge this and then take on converting it to either os.Root or sftp.

lxd/device/cdi/hooks.go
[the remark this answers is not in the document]
dmitry-lyfar: @tomponline added here, it looks like an important check.

lxd/device/cdi/hooks.go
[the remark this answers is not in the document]
dmitry-lyfar: @tomponline yes, I think it should be handled together here, added a comment: https://github.com/canonical/lxd/issues/17877

lxd/device/cdi/hooks.go
dmitry-lyfar: @tomponline sounds good, I restored the original version. Using sftp simplifies the code quite a bit if available.

tomponline:
@dmitry-lyfar ive been working on this today

PR #17395, opened 2026-01-16
shared: Add context parameter to RunCommand

markylaing:
- Removes deprecated `shared.RunCommand`.
- Renames `shared.RunCommandContext` -> `shared.RunCommand`.
- Updates all call sites.

## Checklist

### DIFF
--- a/lxd/storage/drivers/driver_alletra_volumes.go
+++ b/lxd/storage/drivers/driver_alletra_volumes.go
@@ -295,7 +295,7 @@ func (d *alletra) unmapVolume(vol Volume) error {
 
 		devName := filepath.Base(volumePath)
 		if strings.HasPrefix(devName, "dm-") {
-			_, err := shared.RunCommand(d.state.ShutdownCtx, "multipath", "-f", volumePath)
+			_, err := shared.RunCommand(context.Background(), "multipath", "-f", volumePath)
 			if err != nil {
 				return fmt.Errorf("Failed to unmap volume %q: Failed to remove multipath device %q: %w", vol.name, devName, err)
 			}
--- a/lxd/storage/drivers/driver_pure_util.go
+++ b/lxd/storage/drivers/driver_pure_util.go
@@ -1366,7 +1366,7 @@ func (d *pure) unmapVolume(vol Volume) error {
 
 		devName := filepath.Base(volumePath)
 		if strings.HasPrefix(devName, "dm-") {
-			_, err := shared.RunCommand(d.state.ShutdownCtx, "multipath", "-f", volumePath)
+			_, err := shared.RunCommand(context.Background(), "multipath", "-f", volumePath)
 			if err != nil {
 				return fmt.Errorf("Failed to unmap volume %q: Failed to remove multipath device %q: %w", vol.name, devName, err)
 			}


shared/util.go
```diff
@@ -1113,22 +1113,70 @@ func RunCommandWithFds(ctx context.Context, stdin io.Reader, stdout io.Writer, n
...
-		if err == nil {
-			break
+		if opts.SleepInterval != 0 {
+			runOpts.SleepInterval = opts.SleepInterval
 		}
+	}
 
-		time.Sleep(500 * time.Millisecond)
+	_, ok := ctx.Deadline()
+	if !ok {
```
roosterfish: Worth mentioning in the functions doc that in case the provided context has a set deadline and `runOpts.Timeout` is set, the context's deadline takes precedence?

shared/util.go
```diff
@@ -1113,22 +1113,70 @@ func RunCommandWithFds(ctx context.Context, stdin io.Reader, stdout io.Writer, n
...
+		defer cancel()
 	}
 
-	return output, err
+	runCtx := ctx
+	if runOpts.NoKill {
+		runCtx = context.Background()
+	}
+
+	out, err := RunCommand(runCtx, cmd, args...)
```
roosterfish: Why not merging this into the `for {}` loop right before the `select` statement? And then once `<-time.Tick` returns you can `continue`.

Logically it's the same but this would prevent defining the `RunCommand` twice.

lxc/image.go
```diff
@@ -718,7 +718,7 @@ func (c *cmdImageImport) packImageDir(path string) (string, error) {
 	defer func() { _ = outFile.Close() }()
 
 	outFileName := outFile.Name()
-	_, err = shared.RunCommandContext(context.TODO(), "tar", "-C", path, "--numeric-owner", "--restrict", "--force-local", "--xattrs", "-cJf", outFileName, "rootfs", "templates", "metadata.yaml")
+	_, err = shared.RunCommand(context.TODO(), "tar", "-C", path, "--numeric-owner", "--restrict", "--force-local", "--xattrs", "-cJf", outFileName, "rootfs", "templates", "metadata.yaml")
```
roosterfish, line 721: Can you also use the `cmd.Context()` here instead of `TODO`?

lxd/storage/drivers/utils.go
```diff
@@ -28,6 +28,10 @@ import (
 	"github.com/canonical/lxd/shared/logger"
 )
 
+var tryCommandOpts = &shared.TryRunCommandOpts{
+	NoKill: true,
```
roosterfish: Mhh, should the field rather be called `Kill` so the default is `false`?

roosterfish: Just thinking we might introduce an issue somewhere when we suddenly start killing commands? 
Here you explicitly disable it for the commands in the next commits, but maybe at some point we could miss that?

roosterfish (changes requested):
Only a smaller remark on the `NoKill` field of the opts.

shared/util.go
markylaing: I thought I did say that. I can make it more obvious.

tomponline: runOpts.Timeout seems redundant to me given it takes a context.

lxc/image.go
markylaing: Ah I'm not sure how that was missed.

lxd/storage/drivers/utils.go
markylaing: My thinking here is that it should always be possible to kill a command, so the default behaviour should be to kill a command if the context is cancelled. However, @tomponline assured me that this has caused issues in the past (specifically for running Ceph commands), so I've added this to maintain behaviour in the storage drivers.

shared/util.go
tomponline: you could return an error if the supplied context doesnt have a deadline to force the caller to specify one, or default to a specific timeout. But having 2 ways to specify the timeout is confusing IMHO.

shared/util.go
```diff
@@ -1113,22 +1113,70 @@ func RunCommandWithFds(ctx context.Context, stdin io.Reader, stdout io.Writer, n
...
+}
+
+// TryRunCommand repeatedly runs a command according to the given options and context.
+// If the input context has a deadline, then commands are run until that deadline is exceeded.
+// If the input context does not have a deadline, the [TryRunCommandOpts.Timeout] is used.
+// If [TryRunCommandOpts] is nil or [TryRunCommandOpts.Timeout] is unset, a default timeout of 10 seconds is applied.
+// [TryRunCommandOpts.SleepInterval] determines how long to sleep between command invocations.
+// [TryRunCommandOpts.NoKill] is used to prevent sending a kill signal to the command when the deadline is exceeded
+// or the context is cancelled. If this is set to true, a background context is used when actually executing the command.
+func TryRunCommand(ctx context.Context, opts *TryRunCommandOpts, cmd string, args ...string) (string, error) {
```
tomponline: Can we call it `RunCommandRetry`?

markylaing: Yeah that's nicer :+1:

shared/util.go
```diff
@@ -1113,22 +1113,70 @@ func RunCommandWithFds(ctx context.Context, stdin io.Reader, stdout io.Writer, n
...
-func TryRunCommand(name string, arg ...string) (string, error) {
-	var err error
-	var output string
+// TryRunCommandOpts contains retry options for running commands.
+type TryRunCommandOpts struct {
+	// Timeout is applied to the input context of [TryRunCommand] if it does not already have a deadline.
+	Timeout time.Duration
+
+	// SleepInterval determines how frequently to retry the command on failure.
+	SleepInterval time.Duration
```
tomponline: Wonder if this could be a slice of strategies instead, see https://pkg.go.dev/github.com/Rican7/retry@v0.3.1/strategy#Strategy

tomponline:
> * Removes deprecated `shared.RunCommand`.
> 
> * Renames `shared.RunCommandContext` -> `shared.RunCommand`.

Could you split these 2 into a separate PR as we can merge that whilst we discuss the retry API.

shared/util.go
markylaing: Will open a new PR for `RunCommandRetry` and we can discuss there.

markylaing:
This PR now only has the changes to `shared.RunCommand`.

shared/util.go
```diff
@@ -168,7 +168,7 @@ func HostPathFollow(path string) string {
 	// Rely on "readlink -m" to do the right thing.
 	path = HostPath(path)
 	for {
-		target, err := RunCommand("readlink", "-m", path)
+		target, err := RunCommand(ctx, "readlink", "-m", path)
```
tomponline: IMHO this isn't a good function (`HostPathFollow()`) to add a context argument to, because cancelling the context wont cause it to return an error (which is clearly identifiable) and instead alters its return value in a way that isn't clear if the value is "true" or whether its due to the context cancellation.

Instead I think it would be better, and more efficient to implement the logic inside this function using Go rather than call out to `readlink` in the first place. Which avoids an external dependency and avoids executing another program.

But for now I'd suggest leaving the signature of this function alone.

markylaing: Done.

lxd/storage/drivers/driver_alletra_volumes.go
```diff
@@ -295,7 +295,7 @@ func (d *alletra) unmapVolume(vol Volume) error {
 
 		devName := filepath.Base(volumePath)
 		if strings.HasPrefix(devName, "dm-") {
-			_, err := shared.RunCommandContext(d.state.ShutdownCtx, "multipath", "-f", volumePath)
+			_, err := shared.RunCommand(d.state.ShutdownCtx, "multipath", "-f", volumePath)
```
tomponline: Lets avoid killing unmap/unmount calls prematurely to avoid leaving the system in an inconsistent state.

markylaing: Ok, I'll do that as a separate commit as it seems a different logical change

markylaing: Done.

PR #17396, opened 2026-01-16
Instance: Copy and refresh de-dupes and cleanups

tomponline:
A precursor to the ZFS promotion PR (https://github.com/canonical/lxd/pull/17336) .

- Fixes an issue that allowed protected config keys to be deleted by sending them as empty string.
- Adds `volatile.uuid` as protected config key.
- De-dupes several functions to do with instance config handling.

### DIFF
--- a/lxd/storage/drivers/driver_zfs_volumes.go
+++ b/lxd/storage/drivers/driver_zfs_volumes.go
@@ -586,17 +586,6 @@ func (d *zfs) CreateVolumeFromCopy(vol VolumeCopy, srcVol VolumeCopy, allowIncon
 		revert.Add(func() { _ = d.DeleteVolume(fsVol.Volume, op) })
 	}
 
-	var err error
-
-	// Retrieve snapshots on the source.
-	snapshots := []string{}
-	if !srcVol.IsSnapshot() && len(vol.Snapshots) > 0 {
-		snapshots, err = d.VolumeSnapshots(srcVol.Volume, op)
-		if err != nil {
-			return err
-		}
-	}
-
 	if vol.contentType == ContentTypeFS {
 		// Create mountpoint.
 		err := vol.EnsureMountPath()
@@ -606,6 +595,8 @@ func (d *zfs) CreateVolumeFromCopy(vol VolumeCopy, srcVol VolumeCopy, allowIncon
 
 		revert.Add(func() { _ = os.Remove(vol.MountPath()) })
 	}
+
+	var err error
 
 	// When not allowing inconsistent copies and the volume has a mounted filesystem, we must ensure it is
 	// consistent by syncing and freezing the filesystem to ensure unwritten pages are flushed and that no
@@ -625,7 +616,7 @@ func (d *zfs) CreateVolumeFromCopy(vol VolumeCopy, srcVol VolumeCopy, allowIncon
 	rebase := d.config["zfs.clone_copy"] == "rebase" && (srcVol.volType == VolumeTypeContainer || srcVol.volType == VolumeTypeVM)
 
 	// Use full copy mode when zfs.clone_copy is false or rebase mode is enabled or source volume has snapshots.
-	fullCopy := shared.IsFalse(d.config["zfs.clone_copy"]) || rebase || len(snapshots) > 0
+	fullCopy := shared.IsFalse(d.config["zfs.clone_copy"]) || rebase || len(vol.Snapshots) > 0
 
 	var srcSnapshot string
 	if srcVol.volType == VolumeTypeImage {
@@ -687,7 +678,7 @@ func (d *zfs) CreateVolumeFromCopy(vol VolumeCopy, srcVol VolumeCopy, allowIncon
 		args := []string{"send"} // Use send mode when doing full copy.
 
 		// Handle transferring snapshots.
-		if len(snapshots) > 0 {
+		if len(vol.Snapshots) > 0 {
 			// The `--replicate` mode will cause the destination to be based on the source's origin.
 			args = append(args, "--replicate")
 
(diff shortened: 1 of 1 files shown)


tomponline:
@roosterfish your thoughts on "[lxd/storage/drivers/driver/zfs/volumes: Base expected snapshot names on snapshot volumes supplied from DB in CreateVolumeFromCopy](https://github.com/canonical/lxd/pull/17396/changes/0e0bff6c6196bc68fc5bedc1b483d2e39fcce842)" specifically would be appreciated.

lxd/storage/drivers/driver_zfs_volumes.go
```diff
@@ -687,7 +678,7 @@ func (d *zfs) CreateVolumeFromCopy(vol VolumeCopy, srcVol VolumeCopy, allowIncon
 		args := []string{"send"} // Use send mode when doing full copy.
 
 		// Handle transferring snapshots.
-		if len(snapshots) > 0 {
+		if len(vol.Snapshots) > 0 {
```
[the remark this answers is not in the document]
tomponline: Wanted to chat with @roosterfish about this.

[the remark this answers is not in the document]
roosterfish: I don't see why both wouldn't be identical. The only difference is that `vol.Snapshots` is populated based on what's in the database and `d.VolumeSnapshots` returns what the storage driver thinks are snapshots of the volume.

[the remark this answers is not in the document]
roosterfish: Looks like a false alarm to me.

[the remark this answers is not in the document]
tomponline: > I don't see why both wouldn't be identical. The only difference is that vol.Snapshots is populated based on what's in the database and d.VolumeSnapshots returns what the storage driver thinks are snapshots of the volume.

Yes thats my thinking, and by using what we have already loaded from the DB avoids pulling from the storage driver.

tomponline:
@roosterfish ready for review.

lxd/instance/drivers/driver_common.go
```diff
@@ -2401,3 +2401,34 @@ func (d *common) removeDiskDevices() error {
...
+	if shared.StringPrefixInSlice(deviceConfig.ConfigInitialPrefix, allUpdatedKeys) {
+		for devName, newDev := range addDevices {
+			for k, newVal := range newDev {
+				if !strings.HasPrefix(k, deviceConfig.ConfigInitialPrefix) {
+					continue
+				}
+
+				oldDev, ok := removeDevices[devName]
+				if !ok {
+					return errors.New("New device with initial configuration cannot be added once the instance is created")
```
roosterfish: Worth adding the device name in the error?

lxd/storage/drivers/driver_zfs_volumes.go
```diff
@@ -607,6 +597,16 @@ func (d *zfs) CreateVolumeFromCopy(vol VolumeCopy, srcVol VolumeCopy, allowIncon
 		}
 	}
 
+	if vol.contentType == ContentTypeFS {
```
roosterfish, line 589: What is the reason for moving this?

roosterfish (changes requested):
Looking good, just one smaller remark and a question on the ZFS driver modification.

lxd/storage/drivers/driver_zfs_volumes.go
tomponline: For me it read more clearly that we do the `vol.IsVMBlock()` recursion into this function sooner rather than later, so that it was clear there was recursion going on from the get go.

Also moving it after the `d.VolumeSnapshots` read was a way to delay mutating the filesystem in case there was an error.

lxd/instance/drivers/driver_common.go
tomponline: fixed

lxd/storage/drivers/driver_zfs_volumes.go
roosterfish: I see thanks! A was thinking a bit on this commit but couldn't figure if this was required by the follow up commit or something else.

roosterfish (approved):

lxd/instance/drivers/driver_common.go
```diff
@@ -2401,3 +2401,100 @@ func (d *common) removeDiskDevices() error {
...
+		// our root disk device is not attached to another instance.
+		if shared.IsTrue(oldExpandedConfig["security.protection.start"]) && shared.IsFalseOrEmpty(d.expandedConfig["security.protection.start"]) {
+			var dbVolType dbCluster.StoragePoolVolumeType
+			switch d.dbType {
+			case instancetype.Container:
+				dbVolType = dbCluster.StoragePoolVolumeTypeContainer
+			case instancetype.VM:
+				dbVolType = dbCluster.StoragePoolVolumeTypeVM
+			default:
+				return fmt.Errorf("Unknown instance type %q for checking security.protection.start removal", d.dbType)
```
kadinsayani, line 2488:
```suggestion
				return fmt.Errorf(`Unknown instance type %q for checking "security.protection.start" removal`, d.dbType)
```

lxd/storage/drivers/driver_zfs_volumes.go
kadinsayani: Perhaps a comment on this?

kadinsayani (commented):
Just a couple nits, thanks!

lxd/instance/drivers/driver_common.go
tomponline: https://github.com/canonical/lxd/pull/17420

tomponline:
https://github.com/canonical/lxd/pull/17420

PR #17401, opened 2026-01-16
lxd: Temporary LXD UI access link

MusicDin:
This PR adds support for a UI login using a bearer token, and extends `lxd init` to generate a temporary LXD UI access link.

The user can generate the temporary UI access link by answering positively to the interactive question. Alternatively, the `--ui-temporary-access-link` flag can be used to directly generate and print the temporary UI access link. The flag also works with --auto, in which case the link is printed at the end.

When access link generation is enabled, the `lxd init` creates an identity named `ui-admin-temporary`, adds it to the group named `admins`, and issues a client bearer token for it. The URL is then constructed as `https://<server_address>?token=<bearer_token>`.

The following example show the new question, which is asked if the server address (`core.https_address`).
```
...
Would you like the LXD server to be available over the network? (yes/no) [default=no]: yes
Address to bind LXD to (not including port) [default=all]: 
Port to bind LXD to [default=8443]: 
Would you like to create a temporary LXD UI  access link? (yes/no) [default=no]: yes
...
UI temporary identity (type: Client token bearer): ui-admin-temporary
UI temporary access link (expires: 2026-01-17 16:36): https://10.2.0.42:8443?token=<bearer_token>
```

When the URL is accessed, LXD reads the bearer token from the query parameter, validates it, and sets as a response cookie if valid. Afterwards, the user is redirected to the `/ui`.

Example response:
```
$ curl -i -k -H "User-Agent: Mozilla" https://10.2.0.42:8443?token=<bearer_token>
HTTP/2 302 
cache-control: no-store
content-type: text/html; charset=utf-8
location: /ui/
set-cookie: token_bearer_session=<bearer_token>; Path=/; Max-Age=86400; HttpOnly; Secure; SameSite=Strict
content-length: 27
date: Fri, 16 Jan 2026 15:54:00 GMT

<a href="/ui/">Found</a>.
```

Using the cookie, the UI can authenticate with the server.

Example request:
```
$ curl -s -k https://10.2.0.42:8443/1.0 -H "User-Agent: Mozilla" -H "Cookie: token_bearer_session=<bearer_token>" 
{
  ...
  "metadata": {
    ...
    "auth": "trusted",
    "auth_user_name": "3fb82c34-19df-4aef-b4f1-f170ce4b66c6",
    "auth_user_method": "bearer",
  }
}
```

Finally, the user can logout using `/bearer/logout` endpoint, which always returns an expired empty cookie (regardless of the client type) and redirects to `/ui/login`. The empty cookie instructs the browser to remove the the cookie if it exists.

Example request:
```
$ curl -k https://10.2.0.42:8443/bearer/logout
HTTP/2 302 
cache-control: no-store
content-type: text/html; charset=utf-8
location: /ui/login
set-cookie: token_bearer_session=; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT; Max-Age=0; HttpOnly; Secure; SameSite=Strict
content-length: 32
date: Fri, 16 Jan 2026 16:26:49 GMT

<a href="/ui/login">Found</a>.
```

---

TODO:
- [x] Rebase after https://github.com/canonical/lxd/pull/17643
- [x] Rebase after https://github.com/canonical/lxd/pull/17248
- [x] Docs

### DIFF
--- a/test/suites/auth.sh
+++ b/test/suites/auth.sh
@@ -1804,3 +1804,116 @@ entities_enrichment_with_entitlements() {
 
   lxc auth group permission remove test-group server admin
 }
+
+test_ui_temporary_access_link() {
+  echo "==> Test temporary UI access link"
+  lxd init --ui-temporary-access-link
+
+  # Regenerate while identity already exists.
+  lxd init --ui-temporary-access-link
+
+  # Ensure error is returned if identity exists but does not have sufficient permissions.
+  lxc auth identity edit bearer/ui-admin-temporary <<< "groups: []"
+  ! lxd init --ui-temporary-access-link || false
+
+  # Regenerate after deleting the identity.
+  lxc auth identity delete bearer/ui-admin-temporary
+  lxd init --ui-temporary-access-link
+
+  # Parse UI access URL and bearer token.
+  output=$(lxd init --ui-temporary-access-link)
+  url="https://${output#*https://}"
+  token="${url#*token=}"
+
+  echo "==> Testing temporary UI link access"
+  loginOutput=$(curl -s -k -i -H "User-Agent: Mozilla" "${url}")
+
+  if ! grep -q "token_bearer_session=" <<< "${loginOutput}"; then
+    echo "Error: Cookie not set when accessing generated temporary UI link"
+    return 1
+  fi
+
+  # Check redirect location
+  if ! grep -i "Location: /ui/" <<< "${loginOutput}"; then
+    echo "Error: Redirect to /ui/ not set when accessing generated temporary UI link"
+    return 1
+  fi
+
+  # Extract the cookie value
+  cookie=$(grep -i "set-cookie: token_bearer_session=" <<< "${loginOutput}" | head -n 1 | sed 's/.*token_bearer_session=\([^;]*\).*/\1/')
+  if [ "${cookie}" != "${token}" ]; then
+    echo "Error: Cookie value does not match token"
+    return 1
+  fi
+
+  echo "==> Testing LXD access with cookie"
+  serverInfoOutput=$(curl -s -k -H "User-Agent: Mozilla" -H "Cookie: token_bearer_session=${cookie}" "https://${LXD_ADDR}/1.0")
+
+  # Ensure we are trusted.
+  if ! echo "${serverInfoOutput}" | jq -e '.metadata.auth == "trusted"'; then
+    echo "Error: Client is not trusted when accessing using token_bearer_session cookie"
+    return 1
+  fi
+
+  # Ensure authentication method is bearer.
+  if ! echo "${serverInfoOutput}" | jq -e '.metadata.auth_user_method == "bearer"'; then
+    echo "Error: Auth method is not bearer when accessing using token_bearer_session cookie"
+    return 1
+  fi
+
+  echo "==> Testing current identity information"
+  currentIdentityOutput=$(curl -s -k -H "User-Agent: Mozilla" -H "Cookie: token_bearer_session=${cookie}" "https://${LXD_ADDR}/1.0/auth/identities/current")
+
+  if ! echo "${currentIdentityOutput}" | jq -e '.metadata.authentication_method == "bearer"'; then
+    echo "Error: Current identity information does not include correct token"
+    return 1
+  fi
+
+  # Ensure token expiry date is set.
+  if ! echo "${currentIdentityOutput}" | jq -e '.metadata.expires_at? != null'; then
+    echo "Error: LXD info should include bearer token expiration date"
+    return 1
+  fi
+
+  echo "==> Testing bearer logout"
+  logoutOutput=$(curl -s -k -i -H "Cookie: token_bearer_session=${cookie}" "https://${LXD_ADDR}/bearer/logout")
+
+  # Ensure bearer logout redirects to the UI login page.
+  if ! grep -i "Location: /ui/login" <<< "${logoutOutput}"; then
+    echo "Error: Redirect to /ui/login not found on logout"
+    return 1
+  fi
+
+  # Ensure an empty cookie is set on logout.
+  if ! grep -i "set-cookie: token_bearer_session=;" <<< "${logoutOutput}"; then
+    echo "Error: Cookie not cleared on logout"
+    return 1
+  fi
+
+  # Ensure cookie max-age is set to 0 on logout.
+  if ! grep -i "Max-Age=0" <<< "${logoutOutput}"; then
+    echo "Error: Cookie Max-Age not set to 0 on logout"
+    return 1
+  fi
+
+  echo "==> Testing revoked token access"
+  lxc auth identity token revoke bearer/ui-admin-temporary
+  loginOutput=$(curl -s -k -i -H "User-Agent: Mozilla" "${url}")
+
+  if grep -q "token_bearer_session=" <<< "${loginOutput}"; then
+    echo "Error: Cookie set when accessing generated temporary UI link with invalid token"
+    return 1
+  fi
+
+  # Despite invalid token, LXD should still redirect to the UI.
+  if ! grep -i "Location: /ui/" <<< "${loginOutput}"; then
+    echo "Error: Redirect to /ui/ not set when accessing generated temporary UI link"
+    return 1
+  fi
+
+  # Ensure access is forbidden when using the revoked token.
+  curl -s -k -H "User-Agent: Mozilla" -H "Cookie: token_bearer_session=${cookie}" "https://${LXD_ADDR}/1.0/auth/identities/current" | jq -e '.error_code == 403'
+
+  # Cleanup.
+  lxc auth identity delete bearer/ui-admin-temporary
+}
--- a/test/includes/test-groups.sh
+++ b/test/includes/test-groups.sh
@@ -122,6 +122,7 @@ readonly test_group_standalone=(
     "alias"
     "apparmor"
     "authorization"
+    "ui_temporary_access_link"
     "basic_usage"
     "basic_version"
     "certificate_edit"


lxd/auth/bearer/bearer.go
```diff
@@ -16,6 +16,9 @@ import (
 	"github.com/canonical/lxd/shared/api"
 )
 
+// CookieNameSession is the name of the cookie used to store session bearer tokens.
+const CookieNameSession = "lxd_session"
```
MusicDin: Rename to `token_bearer_session`

MusicDin: This @tomponline suggestion. @markylaing What do you think? We wanted to avoid `lxd_session` as it might be used in the future for other (non-authentication) session information.

lxd/api.go
```diff
@@ -168,14 +170,47 @@ func restServer(d *Daemon) *http.Server {
 	})
 
 	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Set("Content-Type", "application/json")
-
 		if isBrowserClient(r) {
+			// Check for magic link token.
+			isUIMagicLink, token, subject := bearer.IsQueryRequest(r, d.globalConfig.ClusterUUID())
```
MusicDin: Move to dedicated function to reduce complexity of the func

lxd/api.go
```diff
@@ -199,13 +199,10 @@ func restServer(d *Daemon) *http.Server {
...
 					// Prevent the session cookie from being cached by browsers or proxies.
 					w.Header().Set("Cache-Control", "no-store")
-
-					http.Redirect(w, r, "/ui/", http.StatusFound)
-					return
 				}
 			}
 
-			http.Redirect(w, r, "/ui/", http.StatusMovedPermanently)
+			http.Redirect(w, r, "/ui/", http.StatusFound)
```
MusicDin, line 195: Temporary redirect?

client/util.go
```diff
@@ -23,13 +24,51 @@ import (
...
 	if err != nil {
 		return nil, err
 	}
 
+	if !insecureSkipVerify && tlsServerCert == "" && serverCertFingerprint != "" {
+		// If a server fingerprint is provided, we disable default verification and rely on our own
+		// verification logic to confirm the server's identity.
+		// We need to verify that the server's certificate matches the expected fingerprint and that
+		// the certificate is currently valid (not expired).
+		tlsConfig.InsecureSkipVerify = true
```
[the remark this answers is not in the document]
MusicDin: This is the reason for the comment above.

shared/api/server.go
```diff
@@ -244,6 +248,11 @@ type Server struct {
 	// API extension: auth_user
 	AuthUserMethod string `json:"auth_user_method" yaml:"auth_user_method"`
 
+	// When the bearer token expires. Only included for bearer token authenticated identities.
+	// Read only: true
+	// Example: 2026-12-31T23:59:59Z
+	AuthTokenExpiresAt *time.Time `json:"auth_token_expiry,omitempty" yaml:"auth_token_expires_at,omitempty"`
```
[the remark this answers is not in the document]
MusicDin: Nice 🚀

lxd/api.go
```diff
@@ -167,15 +170,34 @@ func restServer(d *Daemon) *http.Server {
...
+		http.Redirect(w, r, "/ui/login", http.StatusFound)
+	})
+
+	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
 		if isBrowserClient(r) {
-			http.Redirect(w, r, "/ui/", http.StatusMovedPermanently)
+			handleUIMagicLink(w, r, d.globalConfig.ClusterUUID(), d.identityCache)
+
+			http.Redirect(w, r, "/ui/", http.StatusTemporaryRedirect)
 			return
```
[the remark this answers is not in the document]
MusicDin: Agree, nice.

lxd/api.go
```diff
@@ -167,15 +170,34 @@ func restServer(d *Daemon) *http.Server {
...
+		w.Header().Set("Cache-Control", "no-store")
+		http.Redirect(w, r, "/ui/login", http.StatusFound)
+	})
+
+	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
 		if isBrowserClient(r) {
-			http.Redirect(w, r, "/ui/", http.StatusMovedPermanently)
+			handleUIMagicLink(w, r, d.globalConfig.ClusterUUID(), d.identityCache)
+
+			http.Redirect(w, r, "/ui/", http.StatusTemporaryRedirect)
```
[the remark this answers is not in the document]
MusicDin: @kimanhou @edlerd Could you advice which status code is more proper? Should I use `StatusFound` instead of `StatusTemporaryRedirect`?

lxd/api.go
```diff
@@ -446,6 +468,37 @@ func setCORSHeaders(rw http.ResponseWriter, req *http.Request, config *clusterCo
...
+		Path:     "/",
+		Secure:   true, // Only send the cookie over HTTPS.
+		HttpOnly: true, // Do not allow JavaScript to access the cookie.
+		SameSite: http.SameSiteStrictMode,
+		MaxAge:   int(expiresIn.Seconds()),
+	})
+
+	// Prevent caching of the response containing a session cookie.
+	w.Header().Set("Cache-Control", "no-store")
+}
```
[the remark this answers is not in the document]
MusicDin: @kimanhou @edlerd Also would like get your input on this one. Thanks

lxd/api.go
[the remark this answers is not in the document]
edlerd: Yes, 302 status found is much more common for such redirects. I'd also say use it instead of 307.

lxd/api.go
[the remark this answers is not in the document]
edlerd: Yes, setting `Referrer-Policy: no-referrer` seems like a good idea, so the browser doesn't leak referrer information on being redirected.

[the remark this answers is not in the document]
MusicDin: Thank you

lxd/api.go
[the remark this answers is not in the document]
MusicDin: Thank you

lxd/main_init.go
```diff
@@ -258,3 +284,57 @@ func (c *cmdInit) defaultHostname() string {
...
+		return fmt.Errorf("Failed to check for existing temporary UI identity: %w", err)
+	}
+
+	// Create identity if it doesn't exist.
+	if uiAdminIdentity == nil {
+		uiAdminIdentityReq := api.IdentitiesBearerPost{
+			Name:   uiAdminIdentityName,
+			Type:   api.IdentityTypeBearerTokenClient,
+			Groups: []string{"admins"},
+		}
```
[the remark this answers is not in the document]
MusicDin: I've fixed that to always ensure the group admins exists and has exactly 1 permission (server admin). Similarly, it ensures the ui identity is always part only of the group admins. If the auth group admins or identity ui-admin-temporary is ever modified/deleted, the lxd init will restore it back.

lxd/auth/encryption/sign.go
tomponline, line 1: commit message "lxd/auth/encryption: Add server cert fingerprint in LXD bearer token payload" seems wrong.

tomponline (commented):
we can land the token expiry stuff from the beginning of the PR now as a separate PR if you like?

lxd/auth/bearer/bearer.go
```diff
@@ -16,6 +16,9 @@ import (
 	"github.com/canonical/lxd/shared/api"
 )
 
+// CookieNameSession is the name of the cookie used to store session bearer tokens.
+const CookieNameSession = "token_bearer_session"
```
tomponline, line 20: does it need to be exported?

lxd/api.go
```diff
@@ -446,6 +450,41 @@ func setCORSHeaders(rw http.ResponseWriter, req *http.Request, config *clusterCo
 	}
 }
 
+// handleUIMagicLink sets the session cookie if the request represents a magic-link request.
+func handleUIMagicLink(w http.ResponseWriter, r *http.Request, clusterUUID string, identityCache *identity.Cache) {
```
tomponline: Please lets lose the term "magic" throughout this PR please. Lets choose a better more descriptive name for this concept.

tomponline:
Hi @minaelee please can you look at the docs part of this PR, thank you!

lxd/auth/bearer/bearer.go
MusicDin: Yes, it is accessed from the `lxd/api.go`. We keep it in auth/bearer to be consistent with OIDC

lxd/api.go
MusicDin: Agree. I use term "UI access link" everywhere now with "temporary" prefix where needed.

lxd/auth/encryption/sign.go
MusicDin: Fixed.

test/suites/auth.sh
```diff
@@ -1792,3 +1792,82 @@ entities_enrichment_with_entitlements() {
 
   lxc auth group permission remove test-group server admin
 }
+
+test_ui_temporary_access_link() {
```
MusicDin, line 1808: @simondeziel Could you please take a look at tests? Thanks

doc/howto/access_ui.md
```diff
@@ -41,7 +41,42 @@ If you have already set up access to the UI, you will see the {guilabel}`Instanc
...
-Access to the UI requires both a browser certificate and a trust token.
+Access to the UI can be obtained in two ways:
+
+- Temporary access via a UI access link (bearer token, valid for 1 day).
+- Permanent access using a browser certificate and trust token.
+
+(access-ui-setup-temporary-access-link)=
+### UI access using temporary link
+
+The `lxd init` guides you through the LXD initialization process.
```
minaelee:
```suggestion
The `lxd init` command guides you through the LXD initialization process.
```

doc/howto/access_ui.md
```diff
@@ -41,7 +41,42 @@ If you have already set up access to the UI, you will see the {guilabel}`Instanc
...
+When the server address is configured during the initialization process, the `lxd init` offers an option to generate a temporary UI access link. If you agree with that option, a temporary LXD UI access URL that is valid for 24 hours is printed at the end of the process, as shown below.
```
minaelee:
```suggestion
When the server address is configured during the initialization process, LXD offers an option to generate a temporary UI access link. If you agree with that option, a temporary LXD UI access URL that is valid for 24 hours is printed at the end of the process, as shown below.
```

doc/howto/access_ui.md
```diff
@@ -41,7 +41,42 @@ If you have already set up access to the UI, you will see the {guilabel}`Instanc
...
+
+```
```
minaelee:
````suggestion
```bash
````

File: doc/howto/access_ui.md [hunk 16]
```diff
@@ -41,7 +41,42 @@ If you have already set up access to the UI, you will see the {guilabel}`Instanc
...
+UI temporary access link (expires: 2026-01-17 16:36): https://127.0.0.1:8443/?token=<bearer_token>
+```
+
+Open this URL in your browser to immediately access the UI as an admin.
+This method is intended for initial access and setup only. After logging in, configure the permanent authentication (mTLS or OIDC) for continued access.
+
+To obtain a new temporary UI access link, run `lxd init` again.
+For convenience, the `--ui-temporary-access-link` flag can be used to non-interactively generate a new one.
+
+```
```
minaelee:
````suggestion
```bash
````

File: doc/howto/access_ui.md [hunk 17]
```diff
@@ -41,7 +41,42 @@ If you have already set up access to the UI, you will see the {guilabel}`Instanc
 
 <!-- Include start access UI -->
```
minaelee:
```suggestion
```

doc/howto/access_ui.md
```diff
@@ -41,7 +41,42 @@ If you have already set up access to the UI, you will see the {guilabel}`Instanc
...
+To obtain a new temporary UI access link, run `lxd init` again.
+For convenience, the `--ui-temporary-access-link` flag can be used to non-interactively generate a new one.
+
+```
+lxd init --ui-temporary-access-link
+```
+
+(access-ui-setup-certificate)=
+### Permanent UI access using browser certificate
+
```
minaelee, line 76:
```suggestion

<!-- Include start access UI -->

```

doc/howto/access_ui.md
minaelee: Moving this comment line to where I suggest it below will prevent the current RTD build failure, which stems from having reference link targets duplicated in the tutorial via an `{include}` role. 

All the text between this comment and the ending comment on line 102 is included/duplicated in doc/tutorial/first_steps.md (see lines 128-130 there), including the reference link targets, and Sphinx will fail if there are duplicated link targets.

Later, we can look at updating the tutorial to use the temporary link, but since the new option to generate one during initialization defaults to no, and the tutorial uses `lxd init --minimal` instead of stepping through the initialization process, it is likely to need more extensive revision. For now, it's practical to move the comment line for this PR, which keeps the tutorial as-is (not using the temporary UI link) and should successfully build the docs on RTD.

MusicDin: Thank you! 🚀 

> Later, we can look at updating the tutorial to use the temporary link, but since the new option to generate one during initialization defaults to no, and the tutorial uses lxd init --minimal instead of stepping through the initialization process, it is likely to need more extensive revision.

We could keep `--minimal` and show optional steps to configure server address (`lxc config set core.https_address=...`), followed by direct UI access link generation (`lxd init --ui-temporary-access-link`)

MusicDin: Or shift to `--auto`, which can be combined with flag `--ui-temporary-access-link`. Although I would personally prefer keeping minimal with optional steps to be more explicit.

tomponline: @MusicDin @minaelee did this get resolved?

minaelee: Yes, it's resolved!

MusicDin: I've fixed it, but just provided some suggestion how we may include UI access token generation in tutorials. In any case, changes to tutorials should lend separately IMO because UI team is waiting for this PR.

test/suites/auth.sh
```diff
@@ -1804,3 +1804,90 @@ entities_enrichment_with_entitlements() {
...
+
+  # Check redirect location
+  if ! grep -i "Location: /ui/" <<< "${loginOutput}"; then
+        echo "Error: Redirect to /ui/ not set when accessing generated temporary UI link"
+        return 1
+  fi
+
+  # Extract the cookie value
+  cookie=$(grep -i "set-cookie: token_bearer_session=" <<< "${loginOutput}" | head -n 1 | sed 's/.*token_bearer_session=\([^;]*\).*/\1/')
+  [ "${cookie}" != "${token}" ] && echo "Error: Cookie value does not match token" && return 1
```
simondeziel: If you think a specific error should be printed out, please use this form instead:

```suggestion
  if [ "${cookie}" != "${token}" ]; then
    echo "Error: Cookie value does not match token"
    return 1
  fi
```

test/suites/auth.sh
```diff
@@ -1804,3 +1804,90 @@ entities_enrichment_with_entitlements() {
...
+
+  if ! grep -q "token_bearer_session=" <<< "${loginOutput}"; then
+      echo "Error: Cookie not set when accessing generated temporary UI link"
+      return 1
+  fi
+
+  # Check redirect location
+  if ! grep -i "Location: /ui/" <<< "${loginOutput}"; then
+        echo "Error: Redirect to /ui/ not set when accessing generated temporary UI link"
+        return 1
```
simondeziel: That indentation is a inconsistent with the rest

test/suites/auth.sh
```diff
@@ -1804,3 +1804,90 @@ entities_enrichment_with_entitlements() {
...
+  lxd init --ui-temporary-access-link
+  lxc auth group show admins
+
+  # Parse UI access URL and bearer token.
+  output=$(lxd init --ui-temporary-access-link)
+  url="https://${output#*https://}"
+  token="${url#*token=}"
+
+  echo "==> Testing temporary UI link access"
+  loginOutput=$(curl -k -i -H "User-Agent: Mozilla" "${url}")
```
simondeziel:
```suggestion
  loginOutput=$(curl -s -k -i -H "User-Agent: Mozilla" "${url}")
```

This should avoid showing the download progress

test/suites/auth.sh
```diff
@@ -1804,3 +1804,90 @@ entities_enrichment_with_entitlements() {
...
+        echo "Error: Redirect to /ui/ not set when accessing generated temporary UI link"
+        return 1
+  fi
+
+  # Extract the cookie value
+  cookie=$(grep -i "set-cookie: token_bearer_session=" <<< "${loginOutput}" | head -n 1 | sed 's/.*token_bearer_session=\([^;]*\).*/\1/')
+  [ "${cookie}" != "${token}" ] && echo "Error: Cookie value does not match token" && return 1
+
+  echo "==> Testing LXD access with cookie"
+  serverInfoOutput=$(curl -s -k -H "User-Agent: Mozilla" -H "Cookie: token_bearer_session=${cookie}" "https://${LXD_ADDR}/1.0")
```
simondeziel, line 1850: `/1.0` doesn't require authentication so why try to access it with a cookie? Was the intention to test another endpoint?

```
$ curl -s -k -H "User-Agent: Mozilla" -H "Cookie: token_bearer_session=bad" https://c2d:8443/1.0 | jq .status
"Success"
```

test/suites/auth.sh
```diff
@@ -1804,3 +1804,90 @@ entities_enrichment_with_entitlements() {
...
+  fi
+
+  # Ensure token expiry date is set.
+  if ! echo "${currentIdentityOutput}" | jq -e '.metadata.expires_at? != null'; then
+    echo "Error: LXD info should include bearer token expiration date"
+    return 1
+  fi
+
+  echo "==> Testing bearer logout"
+  logoutOutput=$(curl -k -i -H "Cookie: token_bearer_session=${cookie}" "https://${LXD_ADDR}/bearer/logout")
```
simondeziel: Could you confirm the cookie is no longer valid after the logout? The rest of the test makes sure the cookie is invalidated client side but it'd be nice to test server side.

simondeziel:
```suggestion
  logoutOutput=$(curl -s -k -i -H "Cookie: token_bearer_session=${cookie}" "https://${LXD_ADDR}/bearer/logout")
```

Also to avoid download progress.

test/suites/auth.sh
MusicDin: To confirm `/1.0` reports the client is trusted and auth method is `bearer` when using a cookie.

test/suites/auth.sh
MusicDin: Could you elaborate on this one? 

The logout response is always the same, an empty cookie with a non-positive max-age.
I've added one test case below where I revoke the bearer token and try to login again. Did you have that in mind?

test/suites/auth.sh
simondeziel: Oh, indeed, sorry about that!

test/suites/auth.sh
simondeziel: Yep, the additional check post-logout is exactly what I was asking for, thanks!

lxd/api.go
```diff
@@ -446,6 +450,41 @@ func setCORSHeaders(rw http.ResponseWriter, req *http.Request, config *clusterCo
...
 
+// handleUIAccessLink sets the session cookie if the request represents a temporary UI  access link.
+func handleUIAccessLink(w http.ResponseWriter, r *http.Request, clusterUUID string, identityCache *identity.Cache) {
+	isUIAccessLink, token, subject := bearer.IsQueryRequest(r, clusterUUID)
+	if !isUIAccessLink {
+		return
+	}
+
+	requestorArgs, err := bearer.Authenticate(token, subject, identityCache)
+	if err != nil || requestorArgs == nil || requestorArgs.ExpiresAt == nil {
```
tomponline, line 479: @markylaing in your future security events PR, is this the sort of place you'd log an invalid authentication attempt if `bearer.Authenticate` returns non-nil err?

lxd/main_init.go
```diff
@@ -82,6 +93,10 @@ func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
 		return errors.New("Can't use --minimal and --auto together")
 	}
 
+	if c.flagUITemporaryAccessLink && (c.flagPreseed || c.flagDump || c.flagMinimal) {
+		return errors.New("Can't use --ui-temporary-access-link with --preseed, --dump, or --minimal")
```
tomponline: "Cannot" please don't use contractions in docs/error messages.

lxd/main_init.go
```diff
@@ -258,3 +286,113 @@ func (c *cmdInit) defaultHostname() string {
...
+		return fmt.Errorf("Failed to refresh LXD server info: %w", err)
+	}
+
+	var serverAddress string
+	if len(server.Environment.Addresses) > 0 {
+		serverAddress = server.Environment.Addresses[0]
+	}
+
+	if serverAddress == "" {
+		return errors.New("LXD server address is not set, can't create UI temporary access link")
```
tomponline: "cannot" please dont use contractions.

lxd/main_init.go
```diff
@@ -258,3 +286,113 @@ func (c *cmdInit) defaultHostname() string {
...
+	}
+
+	uiAdminIdentityName := "ui-admin-temporary"
+	uiAdminIdentityGroup := "admins"
```
tomponline, line 308: Can we use/define a constant here for the pre-initialised "admins" group so we can easily find where its referenced in the code base?

lxd/main_init.go
```diff
@@ -258,3 +286,113 @@ func (c *cmdInit) defaultHostname() string {
...
+	}
+
+	// Ensure admins group exists and has the expected permissions.
+	adminsGroup, etag, err := d.GetAuthGroup(uiAdminIdentityGroup)
+	if err != nil && !api.StatusErrorCheck(err, http.StatusNotFound) {
+		return fmt.Errorf("Failed to check for existing temporary UI identity: %w", err)
+	}
+
+	if adminsGroup == nil {
+		// Create group if it doesn't exist.
```
tomponline, line 323: I don't think we should have 2 duplicate implementations where the "admins" group is created.

Do we need this bit, should we not expect the admins group to exist?

tomponline: @markylaing what do you think about this?

lxd/main_init.go
```diff
@@ -258,3 +286,113 @@ func (c *cmdInit) defaultHostname() string {
...
+				Description: uiAdminIdentityGroupDesc,
+				Permissions: []api.Permission{uiAdminIdentityGroupPerm},
+			},
+		}
+
+		err := d.CreateAuthGroup(adminsGroupReq)
+		if err != nil {
+			return fmt.Errorf("Failed to create admin auth group: %w", err)
+		}
+	} else if len(adminsGroup.Permissions) != 1 ||
```
tomponline: I can understand why `len(adminsGroup.Permissions) <= 0` would be a trigger to add permissions, but i dont see why `!= 1` would be a trigger, after all the `admins` group could already have multiple permissions, and still have the correct one needed, so this feels brittle to me.

lxd/main_init.go
```diff
@@ -258,3 +286,113 @@ func (c *cmdInit) defaultHostname() string {
...
+			Name:   uiAdminIdentityName,
+			Type:   api.IdentityTypeBearerTokenClient,
+			Groups: []string{uiAdminIdentityGroup},
+		}
+
+		err := d.CreateIdentityBearer(uiAdminIdentityReq)
+		if err != nil {
+			return fmt.Errorf("Failed to create temporary UI identity: %w", err)
+		}
+	} else if len(uiAdminIdentity.Groups) != 1 || uiAdminIdentity.Groups[0] != uiAdminIdentityGroup {
```
tomponline: I'm not sure this is a good idea. If the group identity exists already we shouldn't be adding new groups to it implicitly.

lxd/main_init.go
MusicDin: Hmm, does any other permission have an impact if an identity has server admin role?

My point was, ensure that `admins` have exactly server admin permission and ensure `ui-temporary-access` identity is part of `admins` group.

lxd/main_init.go
MusicDin: We just make sure the identity has exactly one group named `admins`, which gives full access, so everything else *should* be redundant.

I can error out instead if this is desired?

tomponline: We need to be careful that if the `admins` group exists already and has any existing membership then we shouldn't be increasing its permissions.

lxd/main_init.go
tomponline: The key area of concern for me is if the `admins` group already exists and has existing members, we shouldn't be increasing its permission level.

If its empty or doesn't exist we can set it up with increased permissions level, but giving additional permissions to existing users is something we should avoid in my view.

lxd/main_init.go
MusicDin: Yeah, admins group is created as part of the schema initialization. That would mean that if the group is removed or modified later, we can no longer generate the temporary ui access token, unless user manually recreates the group.

Since we allow group admins to be modified or removed, IMO we should at least be able to recreate it if missing.

MusicDin:
@tomponline Now instead of recovering admin group and UI identity, I create them only if missing. If they exist I check if they have expected server admin permission or group admins, respectively. If not, the process errors out.

lxd/main_init.go
tomponline: re-creating if missing is fine

markylaing:
Late to the party reviewing this, apologies. All looking good to me! Looking forward to trying it 🎉

PR #17402, opened 2026-01-16
Fix snapshot config volume edit

tugbataluy:
## Checklist

Fixed a bug in lxc config edit and lxc storage volume edit where specific error messages were not displayed to users during interactive editing sessions. Previously, when API errors occurred or users attempted to modify read-only fields (such as non-editable snapshot properties), the commands would only show a generic "Press enter to open the editor again" prompt without displaying the actual error message, making it difficult for users to understand what went wrong. The fix ensures that all errors (both validation errors and API errors) are properly displayed to stderr before respawning the editor, restoring the expected behavior where users can see exactly what needs to be fixed. Additionally, refactored the error handling code to use a consistent pattern across both commands, improving maintainability and reducing code duplication.

### DIFF
--- a/lxc/storage_volume.go
+++ b/lxc/storage_volume.go
@@ -1184,9 +1184,11 @@ func (c *cmdStorageVolumeEdit) run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if isSnapshot {
-		for {
-			// Parse the text received from the editor.
+	for {
+		var updateErr error
+
+		// Parse the text received from the editor.
+		if isSnapshot {
 			newdata := api.StorageVolumeSnapshot{}
 			err = yaml.Unmarshal(content, &newdata)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


lxc/storage_volume.go
```diff
@@ -1184,68 +1184,54 @@ func (c *cmdStorageVolumeEdit) run(cmd *cobra.Command, args []string) error {
...
-			}
+		// If successful, break out of the loop.
+		if updateErr == nil {
+			break
 		}
 
-		// Respawn the editor for any error condition.
-		fmt.Println(i18n.G("Press enter to open the editor again or ctrl+c to abort change"))
+		// Show error and respawn editor.
+		fmt.Fprintf(os.Stderr, "Config parsing error: %s\n", updateErr)
```
tomponline: @tugbataluy shall we just go with:

```
			fmt.Fprintf(os.Stderr, "Error: %s\n", updateErr)
```

Here, the same as the other commit, as not all errors that can occur are config parsing errors.

tugbataluy: Before performing the edit validation, I looked at the old version in my previous commits. I could send the entire code, but specifically in the interactive edit section, I see that the error is printed as follows:

```
// Spawn the editor
		content, err := shared.TextEditor("", []byte(c.helpTemplate()+"\n\n"+string(data)))
		if err != nil {
			return err
		}

		for {
			if isSnapshot {
				newdata := api.InstanceSnapshotPut{}
				err = yaml.Unmarshal(content, &newdata)
				if err == nil {
					var op lxd.Operation
					op, err = resource.server.UpdateInstanceSnapshot(fields[0], fields[1],
						newdata, etag)
					if err == nil {
						err = op.Wait()
					}
				}
			} else {
				newdata := api.InstancePut{}
				err = yaml.Unmarshal(content, &newdata)
				if err == nil {
					var op lxd.Operation
					op, err = resource.server.UpdateInstance(resource.name, newdata, etag)
					if err == nil {
						err = op.Wait()
					}
				}
			}

			// Respawn the editor
			if err != nil {
				fmt.Fprintf(os.Stderr, i18n.G("Config parsing error: %s")+"\n", err) <-- This line

				if err != nil {
					return err
				}

				if err != nil {
					return err
				}

				continue
			}

			break
		}

```

This tells me that any error is printed with the header “config parsing error.” Am I thinking correctly?

tomponline: yeah lets change that to be the same as the instance edit message.

tugbataluy: So, instead of config parse error, let's just print the error directly. I think this is reasonable, but I wanted to point it out because that's how it was before.
Then
`fmt.Fprintf(os.Stderr, “Error: %s\n”, <errorr>)`
I'm changing it to this as you suggested.

PR #17405, opened 2026-01-16
test/suites: use shorter `sleep`

simondeziel:


simondeziel:
Those shorter sleeps were tested several times on GH and in local runners to not introduce any race.

test/suites/clustering.sh
```diff
@@ -1865,7 +1865,7 @@ test_clustering_shutdown_nodes() {
   wait "${daemon_pid3}"
 
   # Wait for raft election to take place and become aware that quorum has been lost (should take 3-6s).
-  sleep 10
+  sleep 7
```
simondeziel, line 1868: In some other places, I've seen `sleep 5` used when waiting for raft changes, dunno if that would also work reliably here?

kadinsayani: When waiting for raft roles to update, 5s suffices. Here, we're waiting for an election, so perhaps 7s is safer?

kadinsayani (approved):
Thanks for shortening those!

test/suites/clustering.sh
simondeziel: Ack, I'll stick with that 7s as it's been reliable in testing. Thanks

PR #17406, opened 2026-01-17
test/suites/incremental_copy: Tweaks

simondeziel:


test/suites/incremental_copy.sh
```diff
@@ -76,7 +84,7 @@ do_copy() {
   lxc copy c1 c2 --refresh ${targetPoolFlag}
   lxc config show c2/snap0
   lxc config show c2/snap1
-  [ "$(lxc storage volume get "${target_pool}" container/c2 user.foo)" = "main" ]
+  [ "$(lxc storage volume get "${target_pool}" container/c2 user.foo)" = "init" ]
```
simondeziel, line 87: @tomponline this one surprised me TBH. I would have expected the prior `lxc copy c1 c2 --refresh` to align `c2`'s `user.foo` value to that of `c1`'s.

tomponline (approved):

test/suites/incremental_copy.sh
tomponline: @kadinsayani is looking at refresh logic currently, if you have chance would you mind checking if this is correct, and whether we should also be updating the logic to be consistent with client-side refresh of config like we are doing in https://github.com/canonical/lxd/pull/17678 Thanks!

kadinsayani: I'll look into this, thanks for raising :)

kadinsayani: I've addressed this in https://github.com/canonical/lxd/pull/17678.

simondeziel: So it was indeed a bug and you've fixed it, thanks!

PR #17408, opened 2026-01-19
Update `TryRunCommand` to accept a context

markylaing:
- Renames `TryRunCommand` to `RunComandRetry`.
- Adds a context argument.
- Adds an options argument.
- Options contain a retry func, which accepts the number of attempts and the last error. The retry func should perform any necessary sleeps. It defaults to a 500ms sleep if not specified.
- Options also contain a `NoKill` bool, which instructs `RunCommandRetry` to use a background context when running commands so that they are not killed when the input context is done.

Depends on #17395

## Checklist

### DIFF
--- a/lxd/storage/drivers/utils.go
+++ b/lxd/storage/drivers/utils.go
@@ -423,7 +423,7 @@ func makeFSType(path string, fsType string, options *mkfsOptions) (string, error
 	// Always add the path to the device as the last argument for wider compatibility with versions of mkfs.
 	cmd = append(cmd, path)
 
-	msg, err = shared.TryRunCommand(cmd[0], cmd[1:]...)
+	msg, err = shared.RunCommandRetry(context.TODO(), noKillRetryOpts, cmd[0], cmd[1:]...)
 	if err != nil {
 		return msg, err
 	}
@@ -527,11 +527,11 @@ func growFileSystem(fsType string, devPath string, vol Volume) error {
 		var err error
 		switch fsType {
 		case "ext4":
-			_, err = shared.TryRunCommand("resize2fs", devPath)
+			_, err = shared.RunCommandRetry(context.TODO(), noKillRetryOpts, "resize2fs", devPath)
 		case "xfs":
-			_, err = shared.TryRunCommand("xfs_growfs", mountPath)
+			_, err = shared.RunCommandRetry(context.TODO(), noKillRetryOpts, "xfs_growfs", mountPath)
 		case "btrfs":
-			_, err = shared.TryRunCommand("btrfs", "filesystem", "resize", "max", mountPath)
+			_, err = shared.RunCommandRetry(context.TODO(), noKillRetryOpts, "btrfs", "filesystem", "resize", "max", mountPath)
 		default:
 			return fmt.Errorf("Unrecognised filesystem type %q", fsType)
 		}
--- a/lxd/storage/drivers/driver_ceph.go
+++ b/lxd/storage/drivers/driver_ceph.go
@@ -185,7 +185,7 @@ func (d *ceph) Create() error {
 
 	if !poolExists {
 		// Create new osd pool.
-		_, err := shared.TryRunCommand("ceph",
+		_, err := shared.RunCommandRetry(context.TODO(), noKillRetryOpts, "ceph",
 			"--name", "client."+d.config["ceph.user.name"],
 			"--cluster", d.config["ceph.cluster_name"],
 			"osd",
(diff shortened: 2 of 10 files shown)


shared/util.go
```diff
@@ -1103,22 +1103,73 @@ func RunCommandWithFds(ctx context.Context, stdin io.Reader, stdout io.Writer, n
...
+	// RetryFunc instructs RunCommandRetry as to whether it should perform another attempt.
+	// It is called after each command failure until the context is cancelled or times out
+	// and must return true to perform another attempt.
+	// The lastErr argument is a [RunError], which can be used to inspect stdout and stderr if necessary.
+	RetryFunc func(attempt uint, lastErr error) bool
+
+	// UseBackgroundCtx instructs RunCommandRetry to use a background context instead of the
+	// input context when running the command. This prevents sending a kill signal to the
+	// command when the deadline is exceeded or the context is cancelled.
+	UseBackgroundCtx bool
```
tomponline: This field name seems unnecessarily tied to the implementation (the **how** it works) rather than named after the **why** of when you would use it?

How about `NoKill`?

shared/util.go
```diff
@@ -1103,22 +1103,73 @@ func RunCommandWithFds(ctx context.Context, stdin io.Reader, stdout io.Writer, n
...
-func TryRunCommand(name string, arg ...string) (string, error) {
-	var err error
-	var output string
+// RunCommandRetryOpts contains options for running commands.
+type RunCommandRetryOpts struct {
+	// RetryFunc instructs RunCommandRetry as to whether it should perform another attempt.
+	// It is called after each command failure until the context is cancelled or times out
+	// and must return true to perform another attempt.
+	// The lastErr argument is a [RunError], which can be used to inspect stdout and stderr if necessary.
+	RetryFunc func(attempt uint, lastErr error) bool
```
roosterfish, line 1111: I am curious in which situation are we about to use this func? Is it for commands where we want to wait until they return a specific string on stdout/stderr?

shared/util.go
```diff
@@ -1103,22 +1103,73 @@ func RunCommandWithFds(ctx context.Context, stdin io.Reader, stdout io.Writer, n
...
+		defer cancel()
+	}
+
+	runCtx := ctx
+	if runOpts.UseBackgroundCtx {
+		runCtx = context.Background()
 	}
 
-	return output, err
+	var stdout string
```
tomponline: these vars, with the exception of `attempt` can be defined inside the `for` loop to reduce their scope.

tomponline: so `stdout, err := RunCommand(runCtx, cmd, args...)`

shared/util.go
```diff
@@ -1103,22 +1103,73 @@ func RunCommandWithFds(ctx context.Context, stdin io.Reader, stdout io.Writer, n
...
-		output, err = RunCommand(context.TODO(), name, arg...)
-		if err == nil {
-			break
+// RunCommandRetry repeatedly runs a command according to the given options and context.
+// If the input context has a deadline, then commands are run until that deadline is exceeded.
+// If the input context does not have a deadline, a 10 second timeout is applied.
+//
+// [RunCommandRetryOpts.UseBackgroundCtx] instructs RunCommandRetry to use a background context instead of the
+// input context when actually running the command. This prevents sending a kill signal to the command when the
+// deadline is exceeded or the context is cancelled.
```
tomponline: The doc block should explain what the returned  `string` contains.

shared/util.go
markylaing: I'll revert to `NoKill`.

shared/util.go
markylaing: My thinking was that you may only want to retry if the command exits with a certain code or output. Alternatively you may want to stop retries if a certain code or output is returned.

shared/util.go
markylaing: Done

shared/util.go
markylaing: Done

shared/util.go
markylaing: Done

lxd/storage/drivers/utils.go
```diff
@@ -28,6 +28,12 @@ import (
 	"github.com/canonical/lxd/shared/logger"
 )
 
+// commandRetryOpts is used as the default [shared.RunCommandRetryOpts] for storage operations.
+// It instructs [shared.RunCommandRetry] not to kill the command when the context is done.
+var commandRetryOpts = &shared.RunCommandRetryOpts{
```
tomponline: Could we call this `noKillRetryOpts` so its clear when reading it being used what its for?

tomponline (dismissed):
Just one nit.

lxd/storage/drivers/utils.go
markylaing: yeah makes sense

PR #17426, opened 2026-01-19
Start using mermaid diagrams in docs

kadinsayani:
Adds mermaid diagram extension and converts vm live migration diagrams from static images to mermaid diagrams.


tomponline (commented):
@kadinsayani how does this work with offline (static) rendering, and what size impact does it have on the snap (what format do the diagrams get built to)?

kadinsayani:
> @kadinsayani how does this work with offline (static) rendering, and what size impact does it have on the snap (what format do the diagrams get built to)?

Out of the box this is client‑side only: `sphinxcontrib.mermaid` emits Mermaid JS and the diagrams render to inline SVG at page load. There’s no pre‑rendered/static output, so JS‑disabled/offline renderers (PDF, etc.) won’t show them unless we add `mermaid‑cli` and set `mermaid_output_format` to svg/png. The diagrams are not built to files in the current config; they render to SVG in the DOM. This removes ~1.07 MB of SVG assets but HTML output will now include Mermaid JS (roughly 2–3 MB).

We can configure this to output SVG or PNG at build time, but that requires using `npx` to invoke `mermaid-cli`.

tomponline:
Thanks. Lets hold off on this for now until we have clearer understanding of the need and have discussed with @minaelee

kadinsayani:
I'll close this for now. @minaelee feel free to ping me for further discussion or cherry-pick these commits at any time.

minaelee:
> I'll close this for now. @minaelee feel free to ping me for further discussion or cherry-pick these commits at any time.

Thanks for looking into the use of Mermaid @kadinsayani. My initial impression is that the default rendering size of the diagram is not ideal (smaller and less readable, though it does have a zoom option):

Original:
image

Mermaid:
image

To adjust the size, it looks like we'd either need to add custom CSS or invoke mermaid-cli through npx as you mentioned. This seems like adding complexity without a clear current need. I'm good with leaving this closed until there is a need for adding more diagrams that could benefit from this kind of diagrams-as-code approach.

kadinsayani:
> To adjust the size, it looks like we'd either need to add custom CSS or invoke mermaid-cli through npx as you mentioned. This seems like adding complexity without a clear current need. I'm good with leaving this closed until there is a need for adding more diagrams that could benefit from this kind of diagrams-as-code approach.

Thanks for your input, sounds like a good plan!

PR #17427, opened 2026-01-19
Minor test tweaks

simondeziel:
Try to be more efficient during test:

* use `--empty` where possible
* avoid starting if not needed
* avoid unneeded `--force` during deletion

A small but notable addition is that `lxc file delete` of multiple files at once is now tested.

Also, add `uuidgen` helper to avoid depending on `uuid-runtime` package.

### DIFF
--- a/test/suites/security.sh
+++ b/test/suites/security.sh
@@ -138,10 +138,6 @@ test_security_protection() {
   ensure_import_testimage
 
   # Test deletion protecton
-  lxc init --empty c1 -d "${SMALL_ROOT_DISK}"
-  lxc snapshot c1
-  lxc delete c1
-
   lxc profile set default security.protection.delete true
 
   lxc init --empty c1 -d "${SMALL_ROOT_DISK}"


test/suites/basic.sh
```diff
@@ -728,10 +728,9 @@ test_basic_usage() {
   lxc storage delete bla
 
   # Test rebuilding an instance with its original image.
-  lxc init testimage c1
-  lxc start c1
+  lxc launch testimage c1
   lxc exec c1 -- touch /data.txt
-  lxc stop c1
+  lxc stop c1 # Intentionally not using force to ensure `/data.txt` is flushed to disk.
```
tomponline: did you consider using `lxc exec  -- sync`?

simondeziel: Let me time it and see if there's a win.

simondeziel: It's worth it.

Before:

```
Test              |    dir
:---------------- | -----:
basic_usage (1/3) | 53.79s
basic_usage (2/3) | 52.08s
basic_usage (3/3) | 51.94s
basic_usage (avg) | 52.60s
TOTAL (avg)       | 52.60s
```

After:

```
Test              |    dir
:---------------- | -----:
basic_usage (1/3) | 49.82s
basic_usage (2/3) | 50.07s
basic_usage (3/3) | 51.51s
basic_usage (avg) | 50.47s
TOTAL (avg)       | 50.47s
```

Issue #17126, opened 2025-12-01
Volume snapshots `disk` devices are writable


tomponline:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

N/A

### Distribution version

N/A

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
N/A
```

### Output of "lxc info" or system info if it fails

```shell
LXD 6.6
```

### Issue description

When attaching a volume snapshot as a `disk` device the resulting mount in the guest is writable, furthermore the the volume itself is writable (at least when using `dir` pool).

**LXD snapshots should not be writable.**

### Steps to reproduce

```
lxc storage volume create default foo
lxc storage volume snapshot default foo
lxc launch ubuntu:24.04 v1 --vm
lxc config device add v1 foo disk source=foo source.snapshot=foo pool=default path=/mnt/foo-snap0
lxc exec v1 -- findmnt /mnt/foo-snap0
TARGET           SOURCE            FSTYPE   OPTIONS
/mnt/foo-snap0 lxd_foo--snap0 virtiofs rw,relatime # should be mounted read only

# Even if not mounted readonly should still not be writable
lxc exec v1 -- touch /mnt/foo-snap0/hello # should not work!
```

### Information to attach

- [ ] Any relevant kernel output (`dmesg`)
- [ ] Instance log (`lxc info NAME --show-log`)
- [ ] Instance configuration (`lxc config show NAME --expanded`)
- [ ] Main daemon log (at `/var/log/lxd/lxd.log` or `/var/snap/lxd/common/lxd/logs/lxd.log`)
- [ ] Output of the client with `--debug`
- [ ] Output of the daemon with `--debug` (or use `lxc monitor` while reproducing the issue)

MggMuggins:
[Original commit](https://github.com/canonical/lxd/pull/14930/commits/164bce22d9265ed30e4fe8f72731f1fa75078514) that should have implemented this. Looks like there has been refactoring in the disk device driver that dropped that check.

The test added [here](https://github.com/canonical/lxd-ci/pull/413/files#diff-fa08dbef3d980253eea453366d7bdb7b55ec5f7bcec31243014a94cc4bbe035dR277) must not work.

PR #17429, opened 2026-01-19
Make volume snapshot disk devices read-only

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/17126/.

### DIFF
--- a/test/includes/test-groups.sh
+++ b/test/includes/test-groups.sh
@@ -77,6 +77,7 @@ readonly test_group_instance=(
     "exec"
     "exec_exit_code"
     "lxd_benchmark_basic"
+    "vm_storage_volume_attach"
     "vm_empty"
     "vm_pcie_bus"
 )
@@ -176,7 +177,6 @@ readonly test_group_snap=(
     "snap_apparmor"
     "snap_basic_usage_vm"
     "snap_lxd_user"
-    "snap_storage_volume_attach_vm"
     "snap_vm_empty"
 )
 
--- a/test/suites/storage_volume_attach.sh
+++ b/test/suites/storage_volume_attach.sh
@@ -107,7 +107,7 @@ EOF
   lxc storage volume delete "${pool}" testvolume
 }
 
-test_snap_storage_volume_attach_vm() {
+test_vm_storage_volume_attach() {
   local pool
   local orig_volume_size
   pool="lxdtest-$(basename "${LXD_DIR}")"


lxd/device/disk.go
```diff
@@ -1658,28 +1671,24 @@ func (d *disk) mountPoolVolume() (func(), string, *storagePools.MountInfo, error
...
-
-		revert.Add(func() { _, _ = d.pool.UnmountCustomVolume(storageProjectName, volumeName, nil) })
-	}
+		if d.config["source.snapshot"] != "" {
+			// Custom volume snapshots must be mounted read-only to prevent guest writes.
+			snapVol := d.pool.GetVolume(volumeType, storageDrivers.ContentType(dbVolume.ContentType), volStorageName, dbVolume.Config)
+			err = d.pool.Driver().MountVolumeSnapshot(snapVol, nil)
+			if err != nil {
+				return nil, "", nil, fmt.Errorf(`Failed mounting storage volume snapshot "%s/%s" from storage pool %q: %w`, dbVolumeType, volumeName, d.pool.Name(), err)
+			}
```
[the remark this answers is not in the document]
kadinsayani: `StoragePoolVolumeType` has a `String` method.

lxd/device/disk.go
```diff
@@ -1633,6 +1634,20 @@ func (d *disk) mountPoolVolume() (func(), string, *storagePools.MountInfo, error
 
 	instProj := d.inst.Project()
 	storageProjectName := project.StorageVolumeProjectFromRecord(&instProj, dbVolumeType)
+	err = d.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+		dbVolume, err = tx.GetStoragePoolVolume(ctx, d.pool.ID(), storageProjectName, dbVolumeType, volumeName, true)
+		return err
+	})
+	if err != nil {
+		return nil, "", nil, fmt.Errorf("Failed to fetch local storage volume record: %w", err)
```
tomponline: "Failed loading..."?

lxd/device/disk.go
```diff
@@ -1642,11 +1642,9 @@ func (d *disk) mountPoolVolume() (func(), string, *storagePools.MountInfo, error
 		return nil, "", nil, fmt.Errorf("Failed to fetch local storage volume record: %w", err)
 	}
 
-	var volStorageName string
+	volStorageName := project.Instance(storageProjectName, volumeName)
```
tomponline: Using a `switch` here with a the 3 allowed types, and a default scenario that errors would be safer.

lxd/device/disk.go
```diff
@@ -1639,7 +1639,7 @@ func (d *disk) mountPoolVolume() (func(), string, *storagePools.MountInfo, error
 		return err
 	})
 	if err != nil {
-		return nil, "", nil, fmt.Errorf("Failed to fetch local storage volume record: %w", err)
+		return nil, "", nil, fmt.Errorf("Failed fetching local storage volume record: %w", err)
```
tomponline: "Failed loading" would be more in-keeping wouldnt it?

lxd/device/disk.go
```diff
@@ -1671,12 +1671,24 @@ func (d *disk) mountPoolVolume() (func(), string, *storagePools.MountInfo, error
...
-		mountInfo, err = d.pool.MountCustomVolume(storageProjectName, volumeName, nil)
-		if err != nil {
-			return nil, "", nil, fmt.Errorf(`Failed mounting storage volume "%s/%s" from storage pool %q: %w`, dbVolumeType, volumeName, d.pool.Name(), err)
-		}
+		if d.config["source.snapshot"] != "" {
+			// Custom volume snapshots must be mounted read-only to prevent guest writes.
+			snapVol := d.pool.GetVolume(volumeType, storageDrivers.ContentType(dbVolume.ContentType), volStorageName, dbVolume.Config)
+			err = d.pool.Driver().MountVolumeSnapshot(snapVol, nil)
+			if err != nil {
+				return nil, "", nil, fmt.Errorf(`Failed mounting storage volume snapshot "%s/%s" from storage pool %q: %w`, dbVolumeType, volumeName, d.pool.Name(), err)
```
tomponline: should we include `snapVol` here instead of `volumeName` so that snapshot name is included in the error?

lxd/device/disk.go
```diff
@@ -1671,12 +1671,24 @@ func (d *disk) mountPoolVolume() (func(), string, *storagePools.MountInfo, error
 			}
 		})
 	} else {
-		mountInfo, err = d.pool.MountCustomVolume(storageProjectName, volumeName, nil)
```
tomponline, line 1665: Does this mean that the snapshot wasnt being mounted in actuality, and it was the parent volume?

kadinsayani: Before the fix it was mounting the snapshot. `volumeName` includes the snapshot suffix when `source.snapshot` is set, so `MountCustomVolume` resolved the snapshot volume, but it mounted it with `MountVolume` (read‑write), so the snapshot was writable.

kadinsayani:
> Please can you add the relevant tests to the lxd test suite too. Thanks!

Sure thing, I'll add tests to `test_snap_storage_volume_attach_vm()`. @simondeziel do we need to duplicate [those checks](https://github.com/canonical/lxd/pull/17429/changes/38ca7294ae66b656f14bd00a8c7571b63855b0fe) in lxd-ci or can https://github.com/canonical/lxd-ci/pull/664 be closed?

lxd/device/disk.go
kadinsayani: I've also added support for unmounting custom storage volume snapshots; this was missing.

tomponline:
> > Please can you add the relevant tests to the lxd test suite too. Thanks!
> 
> Sure thing, I'll add tests to `test_snap_storage_volume_attach_vm()`. @simondeziel do we need to duplicate [those checks](https://github.com/canonical/lxd/pull/17429/changes/38ca7294ae66b656f14bd00a8c7571b63855b0fe) in lxd-ci or can [canonical/lxd-ci#664](https://github.com/canonical/lxd-ci/pull/664) be closed?

Please dont add snap tests to lxd repo, use the main vm tests.

We dont need to extend the lxd-ci ones in this case.

tomponline:
> test_snap_storage_volume_attach_vm

Is the "snap" there meaning snapshot or snap package?

kadinsayani:
> Is the "snap" there meaning snapshot or snap package?

snap package.

kadinsayani:
> > > Please can you add the relevant tests to the lxd test suite too. Thanks!
> > 
> > 
> > Sure thing, I'll add tests to `test_snap_storage_volume_attach_vm()`. @simondeziel do we need to duplicate [those checks](https://github.com/canonical/lxd/pull/17429/changes/38ca7294ae66b656f14bd00a8c7571b63855b0fe) in lxd-ci or can [canonical/lxd-ci#664](https://github.com/canonical/lxd-ci/pull/664) be closed?
> 
> Please dont add snap tests to lxd repo, use the main vm tests.
> 
> We dont need to extend the lxd-ci ones in this case.

@simondeziel question for you - where would you recommend adding these checks? I'm considering the following options:

1. `test_vm_storage` - new function in `test/suites/vm.sh`,
2. `test_vm_pcie_bus` - we can reuse a running vm here, but mixes concerns.
3. `test_storage_volume_snapshots` - we can gate the checks with `LXD_VM_TESTS=1`, but this increases the run time since we don't have an existing VM.

tomponline:
> > > > Please can you add the relevant tests to the lxd test suite too. Thanks!
> > > 
> > > 
> > > Sure thing, I'll add tests to `test_snap_storage_volume_attach_vm()`. @simondeziel do we need to duplicate [those checks](https://github.com/canonical/lxd/pull/17429/changes/38ca7294ae66b656f14bd00a8c7571b63855b0fe) in lxd-ci or can [canonical/lxd-ci#664](https://github.com/canonical/lxd-ci/pull/664) be closed?
> > 
> > 
> > Please dont add snap tests to lxd repo, use the main vm tests.
> > We dont need to extend the lxd-ci ones in this case.
> 
> @simondeziel question for you - where would you recommend adding these checks? I'm considering the following options:
> 
>     1. `test_vm_storage` - new function in `test/suites/vm.sh`,
> 
>     2. `test_vm_pcie_bus` - we can reuse a running vm here, but mixes concerns.
> 
>     3. `test_storage_volume_snapshots` - we can gate the checks with `LXD_VM_TESTS=1`, but this increases the run time since we don't have an existing VM.

I suggest re-working `test_snap_storage_volume_attach_vm` so its not in the snap group.

test/suites/storage_volume_attach.sh
```diff
@@ -146,6 +155,7 @@ test_snap_storage_volume_attach_vm() {
 
   # Cleanup
   lxc storage volume delete "${pool}" vol1
+  lxc storage volume delete "${pool}" vol2/snap0
```
simondeziel: AFAIK, snapshot do not prevent the deletion of the volume so I'd remove this bit.

test/includes/test-groups.sh
```diff
@@ -77,6 +77,7 @@ readonly test_group_instance=(
     "exec"
     "exec_exit_code"
     "lxd_benchmark_basic"
+    "vm_storage_volume_attach"
```
simondeziel, line 80: The `group:instance` is tested on `fasts` backends only so are we 100% sure nothing would differ on a per-backend basis?

(I hope the answer is not to add it to the already too slow `group:storage_standalone` but maybe that's needed?)

kadinsayani: Nothing appears to be driver specific, although I did discover that `volume.size` needed to be set properly when testing against `lvm`.

simondeziel (approved):

test/includes/test-groups.sh
tomponline: I would recommend doing this test for all storage drivers because it reaches in and depends on driver specific logic (mounting snapshots).

lxd/device/disk.go
tomponline, line 1: > This avoids a second DB lookup later in mountPoolVolume.

I like to see you're thinking about DB query efficiency, thanks!

test/suites/storage_volume_attach.sh
```diff
@@ -133,9 +139,18 @@ test_snap_storage_volume_attach_vm() {
   lxc exec v1 -- mount | grep -wF /mnt
   lxc exec v1 -- stat /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_vol3
 
+  echo "==> Snapshot custom volume and verify read-only attach"
+  lxc exec v1 -- touch /mnt/snap-data
+  lxc storage volume detach "${pool}" vol2 v1
+  lxc storage volume snapshot "${pool}" vol2
+  lxc config device add v1 vol2-snap disk source=vol2 source.snapshot=snap0 pool="${pool}" path=/mnt-snap
+  lxc exec v1 -- mountpoint -q /mnt-snap
```
tomponline, line 147: Can you also expand this test to check the mount point is marked as read only, and then add an additional test that remounts the mount as writable and tests that its still not actually writable.

This ensures both the guest side mount and the host side mount are readonly.

Please can you do similar for block volume snapshot attachment too. 

Ta

lxd/device/disk.go
kadinsayani: Definitely picking up a lot of these best practices from your suggestions! Thanks for the guidance!

test/includes/test-groups.sh
kadinsayani: Sounds good, I'll move the test over to the `storage_standalone` group.

test/suites/storage_volume_attach.sh
kadinsayani: The `virtiofs` guest mount still reports `rw` even when the underlying host mount is read-only. See this comment:

https://github.com/canonical/lxd/blob/830159411c7d63380f8590b2858b4b0b18d56dd1/lxd/device/disk.go#L1247-L1255

Newer `virtiofsd` builds support `--readonly`, which would allow the guest mount to show `ro`.

tomponline: thats fine - im more interested that the actual mount itself (exported via virtiofsd to the guest) is readonly from a usage perspective.

So thats not relevant to virtiofsd, but how LXD is mounting the volume.

test/includes/test-groups.sh
kadinsayani: https://github.com/canonical/lxd/pull/17469/changes/1bbf013fe3e00a0271233a0613b03fd2485c1abe

test/suites/storage_volume_attach.sh
kadinsayani: https://github.com/canonical/lxd/pull/17469/changes/aaf0f5a03bda0db684baf3282ada67cdece40df0

https://github.com/canonical/lxd/pull/17469/changes/10ecbe86d8c905809f6127651fc2353f77f5ec1d

PR #17431, opened 2026-01-19
Small code deduplication

simondeziel:


shared/util.go
```diff
@@ -228,6 +228,41 @@ func HostPath(path string) string {
 	return "/var/lib/snapd/hostfs" + path
 }
 
+func resolveSnapPath(path string) (string, bool) {
```
kadinsayani, line 186: Even though this function isn't exported, can you please add a godoc comment still? It helps first time readers of a function and makes exporting (if we ever need to) easier :)

shared/util.go
```diff
@@ -228,6 +228,41 @@ func HostPath(path string) string {
...
+
+	// Handle relative paths
+	if path[0] != os.PathSeparator {
+		// Use the cwd of the parent as snap-confine alters our own cwd on launch
+		ppid := os.Getppid()
+		if ppid < 1 {
+			return path, false
+		}
+
+		pwd, err := os.Readlink(fmt.Sprintf("/proc/%d/cwd", ppid))
```
kadinsayani: Good spot for string concatenation?

simondeziel: Thanks for the suggestion!

shared/api/operation.go
```diff
@@ -199,3 +199,32 @@ func (op *Operation) ToClusterJoinToken() (*ClusterMemberJoinToken, error) {
 
 	return &joinToken, nil
 }
+
+func (op *Operation) parseCommonTokenFields() (secret string, fingerprint string, addresses []string, err error) {
```
tomponline, line 203: This would be a good candidate for unit tests.

simondeziel: https://github.com/canonical/lxd/pull/17438

PR #17432, opened 2026-01-19
Validate `source.snapshot` used when attaching volume snapshots

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/17125.

### DIFF
--- a/test/suites/storage_snapshots.sh
+++ b/test/suites/storage_snapshots.sh
@@ -138,7 +138,7 @@ EOF
 
   # Reset expires_at property.
   lxc storage volume show "${storage_pool}" "${storage_volume}/snap0" > "$tmp_yaml"
-  sed -i '/^expires_at:/d' "$tmp_yaml"  # expires_at satırını sil
+  sed -i '/^expires_at:/d' "$tmp_yaml"
   lxc storage volume edit "${storage_pool}" "${storage_volume}/snap0" <<EOF
   $(cat "$tmp_yaml")
 EOF


test/suites/storage_snapshots.sh
```diff
@@ -129,27 +136,28 @@ $(cat "$tmp_yaml")
 EOF
   lxc storage volume show "${storage_pool}" "${storage_volume}/snap0" | grep -xF "expires_at: ${expiry_date}"
 
-  # Reset expires_at property.
+  echo "Reset expires_at property"
   lxc storage volume show "${storage_pool}" "${storage_volume}/snap0" > "$tmp_yaml"
-  sed -i '/^expires_at:/d' "$tmp_yaml"  # expires_at satırını sil
```
simondeziel, line 134: Now I'm curious to know what `satırını sil` means, @tugbataluy ?

test/suites/storage_snapshots.sh
tomponline, line 1: Why are we converting all our comments to echos?

Isnt this just going to further excacerbate the slowness viewing of our ci log files?

kadinsayani: Easier to debug. I hadn't considered the slow log loads.

kadinsayani: @simondeziel and I have been converting comments to echos incrementally.

test/suites/storage_snapshots.sh
tugbataluy: It means "delete the line" in Turkish. Sometimes I accidentally keep comments in Turkish when I was typing and forgot to translate them. My bad sorry 😅

test/suites/storage_snapshots.sh
simondeziel: Indeed, I find it useful to track where we are at in case of failure. Switching all comments might be a bit too much though.

I try to use `echo` for logical steps in a given test and leave comments in for small incremental progress.

test/suites/storage_snapshots.sh
```diff
@@ -14,61 +14,68 @@ test_storage_volume_snapshots() {
...
   lxc storage volume show "${storage_pool}" "${storage_volume}/snap0" | grep 'name: snap0'
   lxc storage volume show "${storage_pool}" "${storage_volume}/snap0" | grep 'expires_at: 0001-01-01T00:00:00Z'
 
-  # Create a snapshot with an expiry date using a YAML configuration
+  echo "Check volume snapshot attachment requires source.snapshot (not source=vol/snap)"
+  lxc init testimage c2 -s "${storage_pool}"
+  [ "$(CLIENT_DEBUG="" SHELL_TRACING="" lxc config device add c2 snap-bad disk source="${storage_volume}/snap0" pool="${storage_pool}" path=/mnt/bad 2>&1)" = "Error: Invalid devices: Device validation failed for \"snap-bad\": \"source\" cannot include a snapshot, use \"source.snapshot\" instead" ]
+  lxc config device add c2 snap-good disk source="${storage_volume}" source.snapshot="snap0" pool="${storage_pool}" path=/mnt/snap-good
+  lxc config device remove c2 snap-good
+  lxc delete -f c2
```
simondeziel:
```suggestion
  lxc init --empty c2 -s "${storage_pool}"
  [ "$(CLIENT_DEBUG="" SHELL_TRACING="" lxc config device add c2 snap-bad disk source="${storage_volume}/snap0" pool="${storage_pool}" path=/mnt/bad 2>&1)" = "Error: Invalid devices: Device validation failed for \"snap-bad\": \"source\" cannot include a snapshot, use \"source.snapshot\" instead" ]
  lxc config device add c2 snap-good disk source="${storage_volume}" source.snapshot="snap0" pool="${storage_pool}" path=/mnt/snap-good
  lxc config device remove c2 snap-good
  lxc delete c2
```

test/suites/storage_snapshots.sh
kadinsayani: I'll drop the commit that changes all of them in that test function. Moving forward I'll just add echos for logical steps throughout a test function.

tomponline: Converting all comments to echos will be too noisy. Echoing out logical sections makes some more sense. But we should agree on a convention for that too.

For example, I've seen some form of these comments, but not sure if they are significant or just someone's personal style:

> # > Verify initial.zfs.blocksize configuration.

I.e the ">" character here?

I think should be discussed rather than doing unilaterally otherwise it will be inconsistent as the rest of the team don't know about it.

Please can you raise it as a discussion point and once agreed team wide we can update the testing docs with the recommendations.

Thanks

tomponline: > Easier to debug. I hadn't considered the slow log loads.

Try opening the CI log on any LXD GH run to see what I mean as your browser grinds to a halt.

test/suites/storage_snapshots.sh
```diff
@@ -36,6 +36,13 @@ test_storage_volume_snapshots() {
   lxc storage volume show "${storage_pool}" "${storage_volume}/snap0" | grep 'name: snap0'
   lxc storage volume show "${storage_pool}" "${storage_volume}/snap0" | grep 'expires_at: 0001-01-01T00:00:00Z'
 
+  # Check volume snapshot attachment requires source.snapshot (not source=vol/snap)
+  lxc init --empty c2 -s "${storage_pool}"
+  [ "$(CLIENT_DEBUG="" SHELL_TRACING="" lxc config device add c2 snap-bad disk source="${storage_volume}/snap0" pool="${storage_pool}" path=/mnt/bad 2>&1)" = "Error: Invalid devices: Device validation failed for \"snap-bad\": \"source\" cannot include a snapshot, use \"source.snapshot\" instead" ]
```
tomponline, line 41: Is there a way in bash to define a string without using double quotes so that the double quotes dont have to be escaped, which makes the test hard to read.

simondeziel: Using single quoting the error string should allow embedding the double quotes without needing any escape.

test/suites/storage_snapshots.sh
kadinsayani: I've started a thread on MM. Thanks for pointing this out :)

test/suites/storage_snapshots.sh
kadinsayani: https://github.com/canonical/lxd/pull/17455

PR #17443, opened 2026-01-21
API: Allow starting instance on copy

roosterfish:
To accommodate one of the fixes for https://github.com/canonical/lxd/issues/13542, we add a new `--start` flag to the `lxc copy` command to allow testing this scenario as it's already supported by the `POST /1.0/instances` endpoint.

When copying instances (and snapshots) to new instances and the destination LXD supports the `instance_create_start` extension, we can indicate the intent to start the new instance with the original API request.
In case the destination doesn't support `instance_create_start`, we can fallback to manually starting the new instance.

I will add an additional test to `lxd-ci` where we already do copy between different LXD versions.

### DIFF
--- a/test/suites/basic.sh
+++ b/test/suites/basic.sh
@@ -170,6 +170,12 @@ test_basic_usage() {
   # Check that volatile.last_state.power is cleared even on refresh.
   [ "$(lxc config get bar volatile.last_state.power || echo fail)" = "" ]
 
+  lxc delete foo -f
+
+  # Test starting container after copy.
+  lxc copy bar foo --start
+  [ "$(lxc list -f csv -c s foo)" = "RUNNING" ]
+
   lxc stop foo -f
   lxc delete bar
 


roosterfish:
> Do we need to keep this bit?

I was looking a few times at this section and always came to the conclusion "yes". However it's actually not different than the other cases as it's still using the same API endpoint. I got distracted by the "migration" phrase in the docstring. 

So yes we should change this too and populate the `Start` field or trigger manual start later for older LXDs without the API extension. Will fix.

roosterfish:
@tomponline I have to revert my previous message. 

Actually there is more to it than I thought originally. Your question sounded too convincing. I have added another commit on this section to clarify why we have to keep it as is. 
This client `copyInstance` func is also used for moving instances which will fall back to migration in case there is no server side move possible.
And the migration protocol currently doesn't allow setting the intent to start the instance on the target.
Therefore we need to manually start the instance after such an operation.

test/suites/basic.sh
```diff
@@ -170,6 +170,12 @@ test_basic_usage() {
   # Check that volatile.last_state.power is cleared even on refresh.
   [ "$(lxc config get bar volatile.last_state.power || echo fail)" = "" ]
 
+  lxc delete foo -f
+
+  # Test starting container after copy.
```
tomponline, line 175: We also need a test to check for not being able to use `--start` with `--refresh` please.

tomponline: Do we also need a test to cover `--start` when copying using the migration protocol?

tomponline (approved):
ta

test/suites/basic.sh
roosterfish: I am adding those to `lxd-ci`. Working on the tests rn.

tomponline: Please can we have basic ones of them in the main lxd test suite too as that supports VM and live migration now.

tomponline: We should ensure that business logic and implementation is tested as much as possible in main lxd test suite, and then basic integration testing done for snap tests.

PR #17447, opened 2026-01-21
test: VM live migration using local storage drivers

MusicDin:
This PR enables live migration test on local storage drivers.
However, only for remote storage drivers live migration with an attached custom block volume is tested.

TODO:
- [ ] Investiage: Live migration using LVM errors out with an error
`Error: Failed migration on source: Failed transferring migration storage snapshot: Specified block job not found`

### DIFF
--- a/test/suites/live_migration.sh
+++ b/test/suites/live_migration.sh
@@ -1,27 +1,37 @@
 #!/bin/bash
 
-# test_clustering_live_migration spawns a 2-node LXD cluster, creates a virtual machine on top of it,
-# creates and attaches a block volume to the virtual machine, writes some arbitrary data to the volume,
-# and runs live migration. Success is determined by the data being intact after live migration.
+# test_clustering_live_migration spawns a 2-node LXD cluster, creates a virtual machine on the first node
+# and live migrates it to the second node. If the storage backend is remote, it also creates a custom
+# storage volume which needs to be live migrated as along the virtual machine.
 test_clustering_live_migration() {
   poolDriver="$(storage_backend "${LXD_INITIAL_DIR}")"
-  if [ "${poolDriver:-}" != "ceph" ]; then
-    export TEST_UNMET_REQUIREMENT="Only supports 'ceph', not ${poolDriver}"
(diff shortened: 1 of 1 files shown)


MusicDin:
@simondeziel @tomponline Could you review this PR? If you agree I would like to include this as well to test live migration on local drivers (without additional storage volume attached).

However, it works for all driver except for the LVM, which I need to investigate, but want to focus on cross LXD live migration first.

test/suites/live_migration.sh
```diff
@@ -1,27 +1,43 @@
...
+    # TODO: LVM live migration runs into:
+    # Error: Failed migration on source: Failed transferring migration storage snapshot: Specified block job not found
+    export TEST_UNMET_REQUIREMENT="Storage driver ${poolDriver} is currently unsupported"
     return 0
   fi
 
+  # For remote storage drivers, we perform the live migration with custom storage pool attached as well.
+  isRemoteDriver=false
+  if [ "${poolDriver}" == "ceph" ]; then
+    isRemoteDriver=true
```
simondeziel, line 18:
```suggestion
    isRemoteDriver=true
    export LXD_TEST_LIVE_MIGRATION_ON_THE_SAME_HOST=true
```

test/suites/live_migration.sh
```diff
@@ -67,13 +88,18 @@ test_clustering_live_migration() {
   # After live migration, the volume should be functional and mounted.
   # Check that the file we created is still there with the same contents.
   echo "Verifying data integrity after live migration"
-  [ "$(LXD_DIR=${LXD_ONE_DIR} lxc exec vm -- cat /mnt/vol1/bar)" = "vm" ]
+  if [ "${isRemoteDriver}" = true ]; then
```
simondeziel, line 84: Please move the `echo` into the `if`.

test/suites/live_migration.sh
simondeziel: Then, during cleanup, I would unconditionally `unset LXD_TEST_LIVE_MIGRATION_ON_THE_SAME_HOST`

PR #17455, opened 2026-01-21
Use single quotes for readability in error assertions

kadinsayani:
Follow-up to https://github.com/canonical/lxd/pull/17432.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -140,3 +140,5 @@ set -e
 
 However, note that in the above example, if the `!` is moved outside of the `[]`, it would also warrant a ` || false` fallback.
 ````
+
+For error message assertions, prefer single-quoted strings so error text with `"` does not require escaping and the comparisons stay readable.


kadinsayani:
> Whenever there is a `'` in the string or a variable we need to embed, the readability grain goes down the drain :/
> 
> It's a question of preference/taste but I'd use the single quoting option only when it avoids escaping.

`aren'"'"'t` is certainly not the most readable :laughing: 

As per the new CLI guidelines, I believe we aren't supposed to have contractions in error messages, so those should say "are not". Furthermore, I think "have not" is more grammatically correct: "Cannot restore "node1" because some storage pools have not started yet". 

"Cannot restore "node1" because some storage pools are stopped" might be even better.

kadinsayani:
> > Whenever there is a `'` in the string or a variable we need to embed, the readability grain goes down the drain :/
> > It's a question of preference/taste but I'd use the single quoting option only when it avoids escaping.
> 
> `aren'"'"'t` is certainly not the most readable 😆
> 
> As per the new CLI guidelines, I believe we aren't supposed to have contractions in error messages, so those should say "are not". Furthermore, I think "have not" is more grammatically correct: "Cannot restore "node1" because some storage pools have not started yet".
> 
> "Cannot restore "node1" because some storage pools are stopped" might be even better.

https://github.com/canonical/lxd/pull/17460

PR #17458, opened 2026-01-21
Faster live migration test

simondeziel:


test/suites/live_migration.sh
```diff
@@ -75,12 +75,12 @@ test_clustering_live_migration() {
   # Perform live migration of the VM from node1 to node2.
   echo "Live migrating instance 'vm' ..."
   LXD_DIR="${LXD_ONE_DIR}" lxc move vm --target node2
-  LXD_DIR="${LXD_ONE_DIR}" waitInstanceReady vm
 
   # After live migration, the volume should be functional and mounted.
   # Check that the file we created is still there with the same contents.
   if [ "${isRemoteDriver}" = true ]; then
     echo "Verifying data integrity after live migration"
+    LXD_DIR="${LXD_ONE_DIR}" waitInstanceReady vm
```
MusicDin: This change defeats the purpose of the test for local drivers. How do we now the the instance booted after live migration, if we don't wait for that?

Instead I would rather exec into VM after migration in any case and check some arbitrary data to confirm the instance is for sure accessible after live migration.

simondeziel: That's a good point, I'll put a comment as to why we always wait for the lxd-agent to dial back post migration, thanks!

test/suites/live_migration.sh
```diff
@@ -59,38 +56,49 @@ test_clustering_live_migration() {
...
+
+  # Verify the VM was not rebooted
+  [ "$(LXD_DIR=${LXD_ONE_DIR} lxc exec v1 -- cat /proc/sys/kernel/random/boot_id)" = "${INITIAL_BOOT_ID}" ]
 
   # After live migration, the volume should be functional and mounted.
   # Check that the file we created is still there with the same contents.
   if [ "${isRemoteDriver}" = true ]; then
     echo "Verifying data integrity after live migration"
-    [ "$(LXD_DIR=${LXD_ONE_DIR} lxc exec vm -- cat /mnt/vol1/bar)" = "vm" ]
+    LXD_DIR=${LXD_ONE_DIR} lxc exec v1 -- grep -Fxm1 "foo-$$" /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_vmdata
```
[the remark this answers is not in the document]
simondeziel: `-x` does the right thing here, we've been using this method in `lxd-ci` already.

PR #17459, opened 2026-01-21
test/suites/migration: Speed improvements

simondeziel:


simondeziel:
> With ZFS, the runtime goes from ~560s to ~475s

It seems like 475 was a lucky run, it seems to be more around ~500s now.

test/suites/migration.sh
tomponline, line 1: For migration I think its worth while copying non-empty instances to better reflect the real world.

simondeziel: I tried to use them sparingly but yeah, let's go back to `testimage`.

PR #17460, opened 2026-01-21
Reword error messages

kadinsayani:
Rewords error messages to not use contractions as per the new CLI guidelines.


simondeziel (approved):
Thanks

PR #17466, opened 2026-01-22
Test: Add checks for instance start after copy

roosterfish:
Can only be merged after:

* https://github.com/canonical/lxd/pull/17540

Follow up with additional tests on https://github.com/canonical/lxd/pull/17443.

In addition I figured there is another case where we have to manually start instances. It's basically always when copy falls back to migration. As this is tricky to figure out in case the server itself decides to do migration, we simply trigger the start on the server directly.

### DIFF
--- a/test/includes/test-groups.sh
+++ b/test/includes/test-groups.sh
@@ -193,6 +193,7 @@ readonly test_group_standalone_storage=(
     "backup_export_import_recover"
     "bucket_recover"
     "container_copy_incremental"
+    "container_copy_start"
     "container_devices_disk"
     "container_devices_disk_restricted"
     "container_move"


lxc/copy.go
```diff
@@ -343,11 +343,29 @@ func (c *cmdCopy) copyInstance(conf *config.Config, sourceResource string, destR
...
+	// Even when requesting a copy (no migration) from the client, the server might
+	// still perform a migration instead.
+	// Therefore we query the instance's state after the requested copy and check if
+	// it still requires a start.
+	if c.flagStart {
+		// First probe the API extension and perform the extra state lookup only if necessary.
+		if !dest.HasExtension("instance_create_start") {
+			start = true
+		} else {
+			state, _, err := dest.GetInstanceState(destName)
```
tomponline: Why do we need to query the instance state if we just copied it, it will be stopped right?

roosterfish: It might be that it's already started because the server did a copy and we set `Start: true` in the request body.

But we don't know if the server decided to do a migration instead (even though we called `client.CreateInstance`) which wouldn't set `Start: true` as this information cannot be set during migration. Therefore we have to check if we still need to start the instance.

tomponline: @roosterfish if the server decided to do a migration, doesn't that mean the server can talk to the target, and thus it can decide to start the instance back up if `Start: true` was in the original request?

roosterfish: Yep that would work too. Probably better as we can move logic from the client to the server. Let me change this.

tomponline: > we can move logic from the client to the server

This is the mindset i think we should get into generally.

roosterfish:
Ready now for review after https://github.com/canonical/lxd/pull/17540 got merged. The majority of changed lines is a new test suite for the various scenarios in which `lxc cp --start` can be used.

lxd/instances_post.go
```diff
@@ -378,6 +378,15 @@ func createFromMigration(ctx context.Context, s *state.State, projectName string
 		}
 
 		instOp.Done(nil) // Complete operation that was created earlier, to release lock.
+
+		// Start up the instance if requested by the client.
+		if req != nil && req.Start {
+			err := inst.Start(false)
+			if err != nil {
+				return fmt.Errorf("Failed to start instance %q: %w", inst.Name(), err)
```
tomponline, line 386: "Failed starting..."

tomponline: Just a nit, if the 2 tests pass ill merge this.

roosterfish: Ah I am always falling for this. Maybe a candidate for the Copilot instructions? I'll add a follow up PR.

PR #17468, opened 2026-01-22
Send heartbeats regularly rather than randomly

skozina:
## Checklist

### DIFF
--- a/lxd/cluster/heartbeat.go
+++ b/lxd/cluster/heartbeat.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"math/rand"
 	"net/http"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


lxd/cluster/heartbeat.go
```diff
@@ -152,7 +152,7 @@ func (hbState *APIHeartbeat) Send(ctx context.Context, networkCert *shared.CertI
 
 			if spreadRange > 0 {
 				select {
-				case <-time.After(time.Duration(rand.Intn(spreadRange)) * time.Millisecond):
```
tomponline, line 155: Would you mind adding a comment here explaining what its doing, and importantly, **why** we do it like that. Thanks!

tomponline: Also the code doesn't compile

skozina: Sorry! I've added a comment with the reasoning, and actually tested this on 3 node cluster.

PR #17469, opened 2026-01-22
Test improvements

kadinsayani:
Adds a `sub_test` logging helper for better test organization and readability. Uses `sub_test` in existing volume attach tests and introduces a new sub-test for verifying read-only mounts on snapshot custom block volumes. Moves `vm_storage_volume_attach` to the `standalone_storage` test group. 

Additionally, updates the echo guidance to recommend using `sub_test` for future tests. Lastly, includes a Makefile target (`update-copilot-instructions`) for updating GitHub copilot instructions to include the `Recommendations` section of `test/README.md` and the commit message table from `CONTRIBUTING.md`.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -119,6 +119,94 @@ Additionally, some checks may fail and leave behind temporary files; please clea
 
 ## Development Guidelines
 
+### Test Recommendations
+
+<!-- BEGIN TEST RECOMMENDATIONS -->
+
+### `sub_test` usage
+
+Use `sub_test` to label meaningful phases within a test and make logs easier to scan.
+Prefer a small number of focused sub-tests over excessive nesting.
+Use `sub_test` before a logical group of commands that verifies a specific expected behavior for a bug fix or feature.
+Comments within the sub-test block are appropriate to explain why specific commands are used, any setup or initial configuration, and other intent that isn't obvious from the commands.
+
+Good:
+
+```sh
+sub_test "Verify intended behavior X"
+...
+sub_test "Verify intended behavior Y"
+...
+```
+
+### `echo` context
+
+Prefer `sub_test` labels and concise comments for context instead of adding `echo` statements.
+Use `echo` only when you need to debug flaky behavior.
+
+### Expected failure
+
+If a command is expected to fail, special care needs to be used in testing.
+
+Bad:
+
+```sh
+set -e
+...
+
+! cmd_should_fail
+
+some_other_command
+```
+
+Good:
+
+```sh
+set -e
+
+! cmd_should_fail || false
+
+some_other_command
+```
+
+Best:
+
+```sh
+set -e
+
+if cmd_should_fail; then
+  echo "ERROR: cmd_should_fail unexpectedly succeeded, aborting" >&2
+  exit 1
+fi
+
+some_other_command
+```
+
+In the "bad" example, if the command unexpectedly succeeds, the script won't
+abort because `bash` ignores `set -e` for compounded commands (`!
+cmd_should_fail`).
+
+The "good" example works around the problem of compound commands by falling
+back to executing `false` in case of unexpected success of the command.
+
+The "best" example also works around the problem of compound commands but in a
+very intuitive and readable form, albeit longer.
+
+````{note}
+This odd behavior of `set -e` with compound commands does not apply inside `[]`.
+
+```sh
+set -e
+# Does the right thing of failing if the file unexpectedly exist
+[ ! -e "should/not/exist" ]
+```
+
+However, note that in the above example, if the `!` is moved outside of the `[]`, it would also warrant a ` || false` fallback.
+````
+
+For error message assertions, prefer single-quoted strings so error text with `"` does not require escaping and the comparisons stay readable.
+<!-- END TEST RECOMMENDATIONS -->
+
 ### Commit Requirements
 
 - **All commits MUST be signed:** Use `git commit -s`
@@ -127,6 +215,7 @@ Additionally, some checks may fail and leave behind temporary files; please clea
 
 **Commit message structure:**
 
+<!-- BEGIN COMMIT STRUCTURE -->
 | Type                 | Affects files                                    | Commit message format               |
 |----------------------|--------------------------------------------------|-------------------------------------|
 | **API extensions**   | `doc/api-extensions.md`, `shared/version/api.go` | `api: Add XYZ extension`            |
@@ -135,7 +224,11 @@ Additionally, some checks may fail and leave behind temporary files; please clea
 | **Go client package**| Files in `client/`                               | `client: Add XYZ`                   |
 | **CLI changes**      | Files in `lxc/`                                  | `lxc/<command>: Change XYZ`         |
 | **LXD daemon**       | Files in `lxd/`                                  | `lxd/<package>: Add support for XYZ`|
-| **Tests**            | Files in `tests/`                                | `tests: Add test for XYZ`           |
+| **Tests**            | Files in `test/`                                 | `test/<path>: Add test for XYZ`     |
+| **GitHub**           | Files in `.github/`                              | `github: Update XYZ`                |
+| **Makefile**         | `Makefile`                                       | `Makefile: Update XYZ`              |
+
+<!-- END COMMIT STRUCTURE -->
 
 ### Code Style
 


test/suites/storage_volume_attach.sh
```diff
@@ -129,35 +129,53 @@ test_vm_storage_volume_attach() {
...
   lxc exec v1 -- stat /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_vol3
 
-  echo "==> Snapshot custom volume and verify read-only attach"
+  sub_test "Snapshot custom block volume and verify read-only attach"
+  lxc storage volume detach "${pool}" vol3 v1
+  lxc storage volume snapshot "${pool}" vol3
+  lxc config device add v1 v3rs disk source=vol3 source.snapshot=snap0 pool="${pool}"
+  sleep 3
+  SNAP_PATH=/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_v3rs
+  lxc exec v1 -- test -L "${SNAP_PATH}"
```
simondeziel: It was probably copied from somewhere but I don't think it adds value. The `readlink` + `cat "/sys/block/${SNAP_DEVNAME}/ro"` that follows already confirms this was a symlink pointing to a block device.

test/suites/storage_volume_attach.sh
```diff
@@ -129,35 +129,53 @@ test_vm_storage_volume_attach() {
...
+  [ "$(lxc exec v1 -- cat "/sys/block/${SNAP_DEVNAME}/ro")" = "1" ]
+  lxc config device remove v1 v3rs
+  sleep 3
+  ! lxc exec v1 -- stat "${SNAP_PATH}" || false
+  lxc storage volume attach "${pool}" vol3 v1
+  sleep 2
+
+  # Detaching storage volumes
+  lxc storage volume detach "${pool}" vol1 v1
+  lxc storage volume detach "${pool}" vol3 v1
```
simondeziel: attaching and detaching without checking if the block dev showed up in between risks missing a failed attach.

test/includes/logging.sh
```diff
@@ -0,0 +1,5 @@
+# Logging helper for sub-test output.
+sub_test() {
+  echo "==> SUB_TEST: ${1} (${TEST_CURRENT_DESCRIPTION})"
```
simondeziel: I'd probably leave the `SUB_TEST: ` bit out to be more succinct. It's best to have everyone agree on the format though.

I'd also want to see colors used. Those work both in CI and on local runners and leverages the brain's ability to quickly spot them.

test/includes/test-groups.sh
```diff
@@ -234,6 +233,7 @@ readonly test_group_standalone_storage=(
     "storage_local_volume_handling"
     "storage_profiles"
     "storage_volume_attach"
+    "vm_storage_volume_attach"
```
simondeziel: There is an undocumented expectation [*] that the "VM" marker in the name appears **not** at the beginning. For keeping those 2 alpha sorted and grouped by thematic, how about move it to be a `_vm` suffix?


*: `main.sh` looks for `_vm` somewhere in the test name so yours fly under the radar.

test/suites/storage_volume_attach.sh
```diff
@@ -129,35 +129,53 @@ test_vm_storage_volume_attach() {
...
   lxc exec v1 -- stat /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_vol1
   lxc exec v1 -- mount | grep -wF /mnt
   lxc exec v1 -- stat /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_vol3
 
-  echo "==> Snapshot custom volume and verify read-only attach"
+  sub_test "Snapshot custom block volume and verify read-only attach"
+  lxc storage volume detach "${pool}" vol3 v1
+  lxc storage volume snapshot "${pool}" vol3
+  lxc config device add v1 v3rs disk source=vol3 source.snapshot=snap0 pool="${pool}"
+  sleep 3
```
simondeziel: Those `sleep` interval are sometimes 2s after device addition, why 3s here?

kadinsayani: We can use 2s.

test/suites/storage_volume_attach.sh
kadinsayani: I'll simplify this part, we already test detaching earlier in the new sub-test.

test/includes/test-groups.sh
kadinsayani: Thanks for the info, I'll switch to a `_vm` suffix :)

test/includes/logging.sh
kadinsayani: What would be your color preference here? I was thinking we could go with yellow (for the sub-test string `${1}`) since sub-tests can be considered intermediate results within a test function.

test/suites/storage_volume_attach.sh
kadinsayani: Thanks, you raise a good point, I'll remove that check.

test/includes/logging.sh
kadinsayani: Personally, I like the `==> SUB_TEST:` prefix because tests begin with a line prefixed with `==> TEST_BEGIN:`. It makes it easy to spot the sections within a test, and `grep` for them.

simondeziel: I'm fine with yellow or pink or orange ... I'll know if the pick was good when I see it in my dark terminal ;)
That's going to be a one line change if we pick wrong so the stakes are quite low.

kadinsayani: Let's go with orange, on theme with Canonical and different from log warning text :)

simondeziel: I feel like we are on a project to paint a shed ... and I love it :P

kadinsayani: The little bits of color throughout our tests certainly brighten my day :)

tomponline:
As we primarily run the tests in Github Actions, I took a look at how the output appears in the Github action log.

As we previously discussed, the main view details page of a github action step barely loads because of the verbosity, and actually I could not find how the new logging output appeared in that view because I couldn't get the view pane to actually load the relevant part.

But I also downloaded the raw logs, and here the new output looks like this:

```
2026-01-23T15:53:30.6334379Z + sub_test 'Hot plugging storage volumes'
2026-01-23T15:53:30.6355366Z + echo -e '==> SUB_TEST: \033[38;5;208mHot plugging storage volumes\033[0m (storage_volume_attach_vm on random)'
2026-01-23T15:53:30.6356935Z ==> SUB_TEST: [38;5;208mHot plugging storage volumes[0m (storage_volume_attach_vm on random)
```

One of the aims was to reduce noise in the logs and not to increase verbosity unnecessarily.
But unfortunately it seems we've not quite achieved that aim as we now have a 3x log implication for each sub-step entry.

The colour code output is also adding noise to the logs.

Can we do something to improve this for the primary use case?

simondeziel:
> But unfortunately it seems we've not quite achieved that aim as we now have a 3x log implication for each sub-step entry. 

`{ set +x; } 2>/dev/null` trick as used in `lxc_remote` might allow to reduce this down to a single log.

> The colour code output is also adding noise to the logs.

Indeed, the raw logs don't render output, the builtin web view does though.

kadinsayani:
> > But unfortunately it seems we've not quite achieved that aim as we now have a 3x log implication for each sub-step entry.
> 
> `{ set +x; } 2>/dev/null` trick as used in `lxc_remote` might allow to reduce this down to a single log.
> 
> > The colour code output is also adding noise to the logs.
> 
> Indeed, the raw logs don't render output, the builtin web view does though.

@simondeziel to improve the load time of logs, perhaps we should disable shell tracing and shift focus to using `sub_test` to pinpoint failures.

kadinsayani:
> As we primarily run the tests in Github Actions, I took a look at how the output appears in the Github action log.
> 
> As we previously discussed, the main view details page of a github action step barely loads because of the verbosity, and actually I could not find how the new logging output appeared in that view because I couldn't get the view pane to actually load the relevant part.
> 
> But I also downloaded the raw logs, and here the new output looks like this:
> 
> ```
> 2026-01-23T15:53:30.6334379Z + sub_test 'Hot plugging storage volumes'
> 2026-01-23T15:53:30.6355366Z + echo -e '==> SUB_TEST: \033[38;5;208mHot plugging storage volumes\033[0m (storage_volume_attach_vm on random)'
> 2026-01-23T15:53:30.6356935Z ==> SUB_TEST: �[38;5;208mHot plugging storage volumes�[0m (storage_volume_attach_vm on random)
> ```
> 
> One of the aims was to reduce noise in the logs and not to increase verbosity unnecessarily. But unfortunately it seems we've not quite achieved that aim as we now have a 3x log implication for each sub-step entry.
> 
> The colour code output is also adding noise to the logs.
> 
> Can we do something to improve this for the primary use case?

Yeah, we can get rid of color output. I usually download the logs and `grep` through them too; in that use case color doesn't add much value.

simondeziel:
> > > But unfortunately it seems we've not quite achieved that aim as we now have a 3x log implication for each sub-step entry.
> > 
> > 
> > `{ set +x; } 2>/dev/null` trick as used in `lxc_remote` might allow to reduce this down to a single log.
> > > The colour code output is also adding noise to the logs.
> > 
> > 
> > Indeed, the raw logs don't render output, the builtin web view does though.
> 
> @simondeziel to improve the load time of logs, perhaps we should disable shell tracing and shift focus to using `sub_test` to pinpoint failures.

I've got no problem to go in black and white but the (global) tracing is an absolute necessity, sorry :/

kadinsayani:
> > > > But unfortunately it seems we've not quite achieved that aim as we now have a 3x log implication for each sub-step entry.
> > > 
> > > 
> > > `{ set +x; } 2>/dev/null` trick as used in `lxc_remote` might allow to reduce this down to a single log.
> > > > The colour code output is also adding noise to the logs.
> > > 
> > > 
> > > Indeed, the raw logs don't render output, the builtin web view does though.
> > 
> > 
> > @simondeziel to improve the load time of logs, perhaps we should disable shell tracing and shift focus to using `sub_test` to pinpoint failures.
> 
> I've got no problem to go in black and white but the (global) tracing is an absolute necessity, sorry :/

Can we disable the tracing just within the `sub_test` helper? That way we just add 1 line rather than 3.

simondeziel:
> > I've got no problem to go in black and white but the (global) tracing is an absolute necessity, sorry :/
> 
> Can we disable the tracing just within the `sub_test` helper? That way we just add 1 line rather than 3.

Yep, that was my suggestion. It should use the trick mentioned before and also do a `set -x` at the very end if tracing is enabled.

kadinsayani:
> > > I've got no problem to go in black and white but the (global) tracing is an absolute necessity, sorry :/
> > 
> > 
> > Can we disable the tracing just within the `sub_test` helper? That way we just add 1 line rather than 3.
> 
> Yep, that was my suggestion. It should use the trick mentioned before and also do a `set -x` at the very end if tracing is enabled.

Ah I scrolled past your original comment by accident. Ack, thanks :)

.github/copilot-instructions.md
```diff
@@ -135,7 +138,11 @@ Additionally, some checks may fail and leave behind temporary files; please clea
 | **Go client package**| Files in `client/`                               | `client: Add XYZ`                   |
 | **CLI changes**      | Files in `lxc/`                                  | `lxc/<command>: Change XYZ`         |
 | **LXD daemon**       | Files in `lxd/`                                  | `lxd/<package>: Add support for XYZ`|
-| **Tests**            | Files in `tests/`                                | `tests: Add test for XYZ`           |
+| **Tests**            | Files in `test/`                                 | `test/<path>: Add test for XYZ`       |
```
simondeziel: There is too much space padding in the last column.

CONTRIBUTING.md
```diff
@@ -41,7 +41,7 @@ Use separate commits for different types of changes:
 | **Go client package**| Files in `client/`                               | `client: Add XYZ`                   |
 | **CLI changes**      | Files in `lxc/`                                  | `lxc/<command>: Change XYZ`         |
 | **LXD daemon**       | Files in `lxd/`                                  | `lxd/<package>: Add support for XYZ`|
-| **Tests**            | Files in `tests/`                                | `tests: Add test for XYZ`           |
+| **Tests**            | Files in `test/`                                 | `test/<path>: Add test for XYZ`       |
```
simondeziel: Same deal re padding.

Makefile
```diff
@@ -229,6 +229,86 @@ check-gomin:
 		exit 1; \
 	fi
 
+.PHONY: update-copilot-instructions
+# Update copilot instructions from two sources:
+# - Extract the recommendations section from test/README.md.
+# - Extract the commit message table from CONTRIBUTING.md.
+# The awk blocks below replace the marked sections in .github/copilot-instructions.md.
```
simondeziel: At this point, AI took over our Makefile :/
I already found it quite lengthy (and I'm guilty for a good chunk of it) and I don't really have a better to propose, Markdown does not support including from other files :/

kadinsayani: I can move this into a separate script that we source in `Makefile`.

kadinsayani: Perhaps we add a sub-directory for scripts sourced by `Makefile`? We can start moving bits there.

kadinsayani: I've added `update-copilot-instructions.sh` to `scripts/`.

scripts/update-copilot-instructions.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/sh
```
simondeziel: I'd use `bash` here and the use `local` variables inside functions.

scripts/update-copilot-instructions.sh
```diff
@@ -0,0 +1,76 @@
...
+
+tmp="$(mktemp)"
+awk 'f{print} /^## Recommendations$/{f=1; next} /^## /{if(f){exit}}' test/README.md > "${tmp}"
+if [ ! -s "${tmp}" ]; then
+  echo "Failed to extract recommendations from test/README.md"
+  exit 1
+fi
+check_markers '<!-- BEGIN TEST RECOMMENDATIONS -->' '<!-- END TEST RECOMMENDATIONS -->' .github/copilot-instructions.md
+replace_block '<!-- BEGIN TEST RECOMMENDATIONS -->' '<!-- END TEST RECOMMENDATIONS -->' \
+  "${tmp}" .github/copilot-instructions.md 0
```
simondeziel: Could those helpers be simplified to be called like this:

```suggestion
check_markers "TEST RECOMMENDATIONS" .github/copilot-instructions.md
replace_block "TEST RECOMMENDATIONS" "${tmp}" .github/copilot-instructions.md 0
```

In other words, they'd accept the name of a section and would figure out to add `<!-- BEGIN `, `<!-- END ` and `-->` on their own.

Makefile
```diff
@@ -229,6 +229,14 @@ check-gomin:
 		exit 1; \
 	fi
 
+.PHONY: update-copilot-instructions
+# Update copilot instructions from two sources:
+# - Extract the recommendations section from test/README.md.
+# - Extract the commit message table from CONTRIBUTING.md.
+# The script replaces the marked sections in .github/copilot-instructions.md.
```
simondeziel: I'd drop it and move it into the helper script as builtin doc.

scripts/update-copilot-instructions.sh
kadinsayani: Done :)

scripts/update-copilot-instructions.sh
```diff
@@ -0,0 +1,81 @@
...
+  local file="$2"
+  local begin="<!-- BEGIN ${label} -->"
+  local end="<!-- END ${label} -->"
+  if ! grep -qF "${begin}" "${file}" || ! grep -qF "${end}" "${file}"; then
+    echo "Missing ${begin}/${end} markers in ${file}"
+    exit 1
+  fi
+}
+
+replace_block() {
```
simondeziel, line 30: Wouldn't it be simpler if this func would call the `check_markers`?

scripts/update-copilot-instructions.sh
```diff
@@ -0,0 +1,81 @@
...
+awk 'f{print} /^## Recommendations$/{f=1; next} /^## /{if(f){exit}}' test/README.md > "${tmp}"
+if [ ! -s "${tmp}" ]; then
+  echo "Failed to extract recommendations from test/README.md"
+  exit 1
+fi
+check_markers "TEST RECOMMENDATIONS" .github/copilot-instructions.md
+replace_block "TEST RECOMMENDATIONS" "${tmp}" .github/copilot-instructions.md 0
+
+echo "Updating commit structure table from CONTRIBUTING.md"
+tmp="$(mktemp)"
```
simondeziel: Why not reused the previous `tmp` file? Not reusing means leaking a file if I followed correctly.

kadinsayani: `trap cleanup EXIT` deletes the last `tmp` and `tmp_out` on exit.

kadinsayani: And we just reassign the second `mktemp` to `tmp`.

scripts/update-copilot-instructions.sh
kadinsayani: Yeah we can move that in.

scripts/update-copilot-instructions.sh
kadinsayani: Ah I see, the first `tmp` will leak. I'll adjust this.

scripts/update-copilot-instructions.sh
```diff
@@ -0,0 +1,82 @@
...
+tmp_out="$(mktemp)"
+awk 'f{print} /^## Recommendations$/{f=1; next} /^## /{if(f){exit}}' test/README.md > "${tmp}"
+if [ ! -s "${tmp}" ]; then
+  echo "Failed to extract recommendations from test/README.md"
+  exit 1
+fi
+replace_block "TEST RECOMMENDATIONS" "${tmp}" .github/copilot-instructions.md 0
+
+echo "Updating commit structure table from CONTRIBUTING.md"
+> "${tmp}"
```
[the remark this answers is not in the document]
simondeziel: @kadinsayani `awk` will deal with overwriting it, no need for explicit truncation IMHO.

scripts/update-copilot-instructions.sh
```diff
@@ -0,0 +1,82 @@
...
+tmp_out="$(mktemp)"
+awk 'f{print} /^## Recommendations$/{f=1; next} /^## /{if(f){exit}}' test/README.md > "${tmp}"
+if [ ! -s "${tmp}" ]; then
+  echo "Failed to extract recommendations from test/README.md"
+  exit 1
+fi
+replace_block "TEST RECOMMENDATIONS" "${tmp}" .github/copilot-instructions.md 0
+
+echo "Updating commit structure table from CONTRIBUTING.md"
+true > "${tmp}"
```
simondeziel: I'd drop it, `awk` will overwrite it anyway.

scripts/update-copilot-instructions.sh
```diff
@@ -0,0 +1,81 @@
...
+    echo "Missing ${begin}/${end} markers in ${file}"
+    exit 1
+  fi
+}
+
+replace_block() {
+  local label="$1"
+  local content="$2"
+  local file="$3"
+  local blank_after_begin="$4"
```
simondeziel: It's bike shedding territory but could that 4th argument be dropped and the markers be moved around a little to alleviate the need for this extra complexity?

simondeziel (dismissed):

tomponline (dismissed):
Great work thanks!

tomponline:
@kadinsayani needs a rebase please

kadinsayani:
> @kadinsayani needs a rebase please

Rebased.

tomponline:
@kadinsayani sorry needs a rebase again

kadinsayani:
@tomponline rebased

PR #17470, opened 2026-01-22
Use `lxc_remote exec` to avoid side effect of `--` implying `--force-local`

simondeziel:


test/suites/backup.sh
```diff
@@ -301,9 +309,9 @@ EOF
     lxc storage volume create "${poolName}" vol1_test size=32MiB
     lxc storage volume attach "${poolName}" vol1_test c1 /mnt
     lxc start c1
-    lxc exec c1 --project test -- mount | grep /mnt
-    echo "hello world" | lxc exec c1 --project test -- tee /mnt/test.txt
-    [ "$(lxc exec c1 --project test -- cat /mnt/test.txt)" = "hello world" ]
+    lxc_remote exec c1 -- mountpoint /mnt
```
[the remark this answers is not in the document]
simondeziel: Using `mount | grep /mnt` was already imprecise as it would match for a mount on `/mntdir` so instead of papering over it with `mount | grep -wF /mnt` I think using `mountpoint /mnt` is clearer and simpler.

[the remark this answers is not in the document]
simondeziel: As luck would have it, `busybox` doesn't implement `mountpoint` so I had to use the next best thing.

test/suites/backup.sh
```diff
@@ -287,10 +287,18 @@ test_container_recover() {
...
     lxc project create test -c features.images=false -c features.profiles=true -c features.storage.volumes=true
-    lxc profile device add default root disk path=/ pool="${poolName}" --project test
-    lxc profile device add default eth0 nic nictype=p2p --project test
+
+    # Switching project avoids needing to pass `--project test` to every command
+    # except for `lxc exec ... -- ...` because the `--` causes the `lxc` wrapper
+    # to also inject a `--force-local` argument which causes `lxc` to ignore any
+    # configuration file thus assuming the default project. A workaround is to
+    # use the `lxc_remote` wrapper that does not forcibly inject
+    # `--force-local`.
```
simondeziel, line 296: Maybe it's just me but I was caught by surprise that `--force-local` had the side effect of ignoring any lxc configuration file which is how it forces the use of the local Unix socket.

Maybe the `lxc --help` text should touch on this?
```
$ lxc --help | grep force-local
      --force-local    Force using the local unix socket
```

simondeziel:
@markylaing I vaguely remember you complaining about an obscure issue with `bash`... maybe it was related to this?

test/includes/lxc.sh
```diff
@@ -7,33 +7,31 @@ lxc() {
...
-    done
+    # Inject client debug flag if needed
+    if [ -n "${CLIENT_DEBUG:-}" ]; then
+        cmd_args+=("${CLIENT_DEBUG}")
+    fi
 
-    if [ "${injected}" = "0" ]; then
-        [ -n "${CLIENT_DEBUG:-}" ] && cmd_args+=("${CLIENT_DEBUG}")
+    # Inject --force-local if needed
+    # This is only done when "--" is present to avoid breaking commands that
```
tomponline, line 18: what is we want to do `lxc exec --` against a remote server?

test/suites/backup.sh
tomponline: sounds like a bug to me

simondeziel:
> I dont really understand the premise of this PR

I've spent too much time trying to figure out why only the `lxc exec` in `container_recover` had the `--project test` specified  after the `lxc project switch test` and it would stop working the minute I dropped it.

> why are we forcing local in the first place for the `lxc_remote` function?

That's historical behaviour, not something I introduced, this PR just makes it easier to use by not requiring to prefix it with `LXC_LOCAL=''` when used.

My theory as to why it was introduced was to avoid the noisy "welcome" message `lxc` emits when first creating a config. In the test suite that would pollute the logs quite a bit.

test/includes/lxc.sh
simondeziel: Just did a search and it's never been needed, thankfully because yeah, it's not possible (with or without this PR).

test/suites/backup.sh
simondeziel: I think we use it as a proxy to skip the noisy message when the client config is created. So should I keep the silent behaviour but still look at an existing config or (silently) create one if need be?

tomponline:
> My theory as to why it was introduced was to avoid the noisy "welcome" message `lxc` emits when first creating a config. In the test suite that would pollute the logs quite a bit.

Does this not happen with `--force-local` then?

If thats the only problem, maybe we can get rid of all this special handling and perhaps suppress the welcome message with an env var?

simondeziel:
> > My theory as to why it was introduced was to avoid the noisy "welcome" message `lxc` emits when first creating a config. In the test suite that would pollute the logs quite a bit.
> 
> Does this not happen with `--force-local` then?

Any config is [ignored](https://github.com/canonical/lxd/blob/main/lxc/main.go#L389-L399) and the config creation+"welcome" message is [skipped](https://github.com/canonical/lxd/blob/main/lxc/main.go#L411-L469)

tomponline:
> > > My theory as to why it was introduced was to avoid the noisy "welcome" message `lxc` emits when first creating a config. In the test suite that would pollute the logs quite a bit.
> > 
> > 
> > Does this not happen with `--force-local` then?
> 
> Any config is [ignored](https://github.com/canonical/lxd/blob/main/lxc/main.go#L389-L399) and the config creation+"welcome" message is [skipped](https://github.com/canonical/lxd/blob/main/lxc/main.go#L411-L469)

If thats the only reason we use `--force-local`, could we make the welcome message get suppressed by an env var, and I don't see yet if we depend on ignoring the config.

tomponline:
@simondeziel ready for rebase

simondeziel:
> @simondeziel ready for rebase

Done and green CI

simondeziel:
> > > > My theory as to why it was introduced was to avoid the noisy "welcome" message `lxc` emits when first creating a config. In the test suite that would pollute the logs quite a bit.
> > > 
> > > 
> > > Does this not happen with `--force-local` then?
> > 
> > 
> > Any config is [ignored](https://github.com/canonical/lxd/blob/main/lxc/main.go#L389-L399) and the config creation+"welcome" message is [skipped](https://github.com/canonical/lxd/blob/main/lxc/main.go#L411-L469)
> 
> If thats the only reason we use `--force-local`, could we make the welcome message get suppressed by an env var, and I don't see yet if we depend on ignoring the config.

I think the PR as-is is already an improvement so I'd like to keep that suggestion for a follow-up PR, if that's OK with you.

Issue #17062, opened 2025-11-25
Force delete instance from API


kimanhou:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

Could you please expose force delete of a running instance in the API ?

### Describe the solution you'd like

I would like to be able to send a DELETE request to `/1.0/instances/{instanceName}?project={projectName}&force=1` to delete a running instance.
It currently returns an error "The instance is running" because the force flag is ignored from this endpoint.

### Describe alternatives you've considered

I can stop the instance then delete it. But the force delete would be way simpler.

### Additional context

https://github.com/canonical/lxd/blob/main/lxd/instance_delete.go#L81

This is the error I currently see:
https://github.com/canonical/lxd/blob/main/lxd/instance_delete.go#L98-L100

SaiRajeshRamaraju:
Hi! I’m planning to work on this and will submit a PR shortly. Please let me know if anyone else is already addressing it.

kimanhou:
Hello @SaiRajeshRamaraju,  did you get any luck on this?

SaiRajeshRamaraju:
<img width="1144" height="745" alt="Image" src="https://github.com/user-attachments/assets/426e6a30-37b5-4da1-961f-0d2b18e3d9a3" />

@kimanhou  In the delete end point the force value is hard coded as false , which causing the issue . For now I will try to solve this by adding a url parameter to this [ gonna take some time]  with removing the hardcode value and replace it with the parameter value,and create a new PR to solve this issue !

kadinsayani:
@SaiRajeshRamaraju thank you for expressing interest in resolving this issue! We really appreciate it :)

For this one, I already had a fix ready to go and it involves a Client SDK change discussed internally. I've opened up a PR here: https://github.com/canonical/lxd/pull/17473.

Perhaps you can open a PR for https://github.com/canonical/lxd/issues/17133? Since you previously expressed interested in that issue, I've assigned it to you :)

I see that you've also opened https://github.com/canonical/lxd/pull/17373. Thanks for that! It looks like there's a comment for you to address on that PR, whenever you get a chance.

PR #17473, opened 2026-01-23
Add support for server-side instance force delete

kadinsayani:
Fixes https://github.com/canonical/lxd/issues/17062.

### DIFF
--- a/test/suites/basic.sh
+++ b/test/suites/basic.sh
@@ -793,7 +793,9 @@ EOF
   lxc launch testimage c2 --ephemeral
   lxc launch testimage c3 --ephemeral
 
-  lxc stop -f c1 c2 c3
+  # Check deletion via force delete and force stop code paths
+  lxc delete -f c1
+  lxc stop -f c2 c3
   [ "$(lxc list -f csv -c n || echo fail)" = "" ]
 
   # Cleanup


kadinsayani:
> Could you add a `lxc query` based test for this new style of deletion? I'd like to see it fail if force isn't used and then succeed afterward.

I'll update the CLI to use the flag properly so we'll be covered :)

kadinsayani:
> Please can the CLI be updated to make conditional use of this if the server supports it. Ta

Done. I've gone with an approach that avoids a client sdk change (adding a new client function rather than adding a `force` param to the existing `DeleteInstance`). Let me know if you'd prefer adding a `force` param since we're approaching LTS release and may be ok to make that change at this time.

tomponline:
> > Please can the CLI be updated to make conditional use of this if the server supports it. Ta
> 
> Done. I've gone with an approach that avoids a client sdk change (adding a new client function rather than adding a `force` param to the existing `DeleteInstance`). Let me know if you'd prefer adding a `force` param since we're approaching LTS release and may be ok to make that change at this time.

We should make an SDK break to be consistent with the other functions that have force arguments.

client/lxd_instances.go
```diff
@@ -1283,6 +1281,25 @@ func (r *ProtocolLXD) DeleteInstance(name string) (Operation, error) {
...
 }
 
+// DeleteInstanceForce requests that LXD deletes the instance, stopping it first if running.
+func (r *ProtocolLXD) DeleteInstanceForce(name string) (Operation, error) {
+	path, _, err := r.instanceTypeToPath(api.InstanceTypeAny)
+	if err != nil {
+		return nil, err
+	}
+
+	instanceType := strings.TrimPrefix(path, "/")
```
simondeziel: Is that trimming to support `/1.0/instances/foo/?force=1` as well as `/1.0/instances/foo?force=1` ?

simondeziel: NVM, can't read.

lxc/delete.go
```diff
@@ -128,25 +135,30 @@ func (c *cmdDelete) run(cmd *cobra.Command, args []string) error {
...
-
-			err = op.Wait()
-			if err != nil {
-				return fmt.Errorf("Stopping the instance failed: %s", err)
-			}
-
-			if ct.Ephemeral {
-				continue
+			// Check if server supports force delete API extension
+			if !resource.server.HasExtension("instance_force_delete") {
```
tomponline: Wouldn't this logic be better incorporated into the client's `DeleteInstance` function to avoid the caller needing to figure this out?

kadinsayani: Yeah, great suggestion, thanks! I've moved the logic into `DeleteInstance`.

lxd/instance_delete.go
```diff
@@ -99,23 +104,32 @@ func doInstanceDelete(ctx context.Context, s *state.State, name string, projectN
...
+	// Pre-check for immediate 400; RunHook repeats to guard against state changes before execution.
+	if inst.IsRunning() && !force {
+		return nil, api.NewStatusError(http.StatusBadRequest, "Instance is running")
 	}
 
 	rmct := func(ctx context.Context, op *operations.Operation) error {
+		if inst.IsRunning() {
+			if !force {
+				return api.NewStatusError(http.StatusBadRequest, "Instance is running")
+			}
```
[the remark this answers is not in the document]
kadinsayani: I've included a comment above: `// Pre-check for immediate 400; RunHook repeats to guard against state changes before execution.`

test/suites/basic.sh
```diff
@@ -783,18 +782,19 @@ devices:
...
+  lxc delete -f c1
   lxc profile delete foo
 
   # Multiple ephemeral instances delete
   lxc launch testimage c1 --ephemeral
   lxc launch testimage c2 --ephemeral
   lxc launch testimage c3 --ephemeral
 
-  lxc stop -f c1 c2 c3
+  lxc delete -f c1
```
simondeziel, line 797: It's good to have split the `stop -f` from the `delete -f` here but please put a comment as I'm ~guarantied to undo this by accident as soon as I forget about this ;)

kadinsayani: Added :)

lxd/instance_delete.go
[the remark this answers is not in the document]
tomponline: RunHook?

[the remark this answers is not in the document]
tomponline: I'd prefer us to deal with state changes in the `doInstanceStatePut` command rather than requiring the handler to deal with it.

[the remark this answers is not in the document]
kadinsayani: > RunHook?

The closure passed to the new operation.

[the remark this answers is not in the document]
kadinsayani: I've gone with a more minimal patch to fix ephemeral instance delete to keep this PR from growing. The RunHook no-ops when the instance was already deleted by the forced stop.

lxd/instance_delete.go
```diff
@@ -104,6 +104,7 @@ func doInstanceDelete(ctx context.Context, s *state.State, name string, projectN
 		return nil, err
 	}
 
+	deletedOnStop := false
```
tomponline, line 107: Is there a reason the stop action isn't dont async inside the operation along with the delete action? As stop can take some time, I wonder why we are doing that part synchronously?

tomponline:
Thanks!

lxd/instance_delete.go
kadinsayani: There's no "why" in any comments or the commit history. I don't see any reason why the stop can't be async.

tomponline: Lets do it :)

kadinsayani: https://github.com/canonical/lxd/pull/17544

PR #17474, opened 2026-01-23
Add `pro` and `microcloud` features flags

simondeziel:


shared/version/useragent.go
```diff
@@ -71,7 +71,12 @@ func UserAgentFeatures(features []string) error {
 		return errors.New("User agent features may not contain whitespace")
 	}
 
-	userAgentFeatures = features
+	for _, f := range features {
```
tomponline: Are you expecting `features` to contain duplicates?

lxd/ubuntupro/client.go
```diff
@@ -104,6 +106,30 @@ func (proCLI) getGuestToken(ctx context.Context) (*api.DevLXDUbuntuProGuestToken
...
+// isHostAttached returns true if the host is attached to a pro subscription.
+func (proCLI) isHostAttached(ctx context.Context) (bool, error) {
+	// Run pro status command.
+	response, err := shared.RunCommand(ctx, "pro", "status", "--format", "json")
+	if err != nil {
+		return false, fmt.Errorf("Ubuntu Pro client command unsuccessful: %w", err)
+	}
+
+	// Parse response.
+	var statusResponse map[string]any
```
tomponline: Any particular reason you didn't go for an anonymous struct here?

Like in https://github.com/canonical/lxd/blob/main/lxd/instance/drivers/qmp/commands.go#L866-L872

tomponline: Although the current approach allows you detect a missing field (one could do that with a pointer struct field I suppose too), but this is just a nit question, not a request for change.

lxd/daemon.go
```diff
@@ -2001,12 +2001,12 @@ func (d *Daemon) init() error {
 		d.tasks.Add(autoRemoveExpiredTokensTask(d.State))
 	}
 
-	// Start all background tasks
-	d.tasks.Start(d.shutdownCtx)
-
 	// Load Ubuntu Pro configuration before starting any instances.
 	d.ubuntuPro = ubuntupro.New(d.shutdownCtx, d.os.ReleaseInfo["NAME"])
 
+	// Start all background tasks
```
tomponline: The commit message for this change lacks the reason **why**, but as its implied the re-ordering is important, I think it is worth expanding the comment here explaining why we need to start background tasks after initialising the pro configuration.

lxd/daemon.go
```diff
@@ -1632,6 +1623,28 @@ func (d *Daemon) init() error {
...
+	if d.serverClustered {
+		userAgentFeatures = append(userAgentFeatures, "cluster")
+
+		// A clustered LXD might be part of a MicroCloud.
+		d.globalConfigMu.Lock()
+		userMicrocloud := d.globalConfig.UserMicrocloud()
+		d.globalConfigMu.Unlock()
+
+		if userMicrocloud {
+			userAgentFeatures = append(userAgentFeatures, "microcloud")
```
tomponline, line 1637: Is a single node MicroCloud considered `d.serverClustered == true` from LXD's perspective?

If not we'll lose the `microcloud` feature in this scenario.

simondeziel: Yes, I've specifically checked for that and a single node MicroCloud does turn on LXD clustering.

tomponline: nice, thanks

lxd/ubuntupro/client.go
simondeziel: In fact I like your suggestion of the anonymous and annotated struct, I made the change ;)

lxd/daemon.go
simondeziel: I hope I better explained the rational now, let me know what you think.

shared/version/useragent.go
markylaing: IIRC the difference between features and storage backends is that storage backends are set once from one place (e.g. from `lxd/storage`), whereas features are set from many different packages.

For me it makes sense to make the list append only, and to avoid any duplicates.

shared/version/useragent.go
```diff
@@ -71,11 +71,7 @@ func UserAgentFeatures(features []string) error {
 		return errors.New("User agent features may not contain whitespace")
 	}
 
-	for _, f := range features {
-		if !slices.Contains(userAgentFeatures, f) {
-			userAgentFeatures = append(userAgentFeatures, f)
-		}
-	}
+	userAgentFeatures = append(userAgentFeatures, features...)
```
markylaing, line 74: Happy to skip checking for duplicates as the call sites are very clear, but can this commit be squashed into aaa1599becc4cbe9f2b14cac31bf90a5a0028533 please? Thanks

simondeziel: For sure, forgot to do it, thanks for catching it.

PR #17478, opened 2026-01-26
Container: Prevent shell expansion in liblxc hook arguments

tomponline:
Based on https://github.com/lxc/incus/pull/2827

Related to https://github.com/lxc/incus/security/advisories/GHSA-8h3p-58qv-8p53

From @stgraber :

> The LXC driver generates an LXC config file which includes some LXC hooks.
> Those get run through a shell and may therefore get expanded if special characters are allowed.
> 
> Some care was taken around that by using strconv.Quote, but this only leads to double quoted rather than single quoted strings (equivalent to "%q" in fmt.Sprintf) when for something exposed to a shell, we really want single quoted strings.
> 
> To address potential issues, this branch:
> 
> - Introduces a new util.SingleQuote function (sadly strconv doesn't provide this directly)
> - Makes use of util.SingleQuote where strconv.Quote was previously used
> 
> This issue came out from security research work by @rmcnamara-snyk but as any exploitation of this requires full admin privileges on Incus, it didn't qualify as a security issue.

### DIFF
--- a/lxd/instance/drivers/driver_lxc.go
+++ b/lxd/instance/drivers/driver_lxc.go
@@ -869,8 +869,11 @@ func (d *lxc) initLXC(config bool) (*liblxc.Container, error) {
 		return nil, err
 	}
 
+	projectShellQuoted := shared.ShellQuote(d.Project().Name)
+	instanceShellQuoted := shared.ShellQuote(d.Name())
+
 	// Call the onstart hook on start.
-	err = lxcSetConfigItem(cc, "lxc.hook.pre-start", fmt.Sprintf("/proc/%d/exe callhook %s %s %s start", os.Getpid(), shared.VarPath(""), shared.ShellQuote(d.Project().Name), shared.ShellQuote(d.Name())))
+	err = lxcSetConfigItem(cc, "lxc.hook.pre-start", fmt.Sprintf("/proc/%d/exe callhook %s %s %s start", os.Getpid(), shared.VarPath(""), projectShellQuoted, instanceShellQuoted))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


lxd/instance/drivers/driver_lxc.go
```diff
@@ -1091,7 +1091,7 @@ func (d *lxc) initLXC(config bool) (*liblxc.Container, error) {
 	}
 
 	if shared.IsTrue(d.expandedConfig["security.delegate_bpf"]) {
-		err = lxcSetConfigItem(cc, "lxc.hook.start-host", d.state.OS.ExecPath+" callhook "+shared.VarPath("")+" "+strconv.Quote(d.Project().Name)+" "+strconv.Quote(d.Name())+" starthost")
+		err = lxcSetConfigItem(cc, "lxc.hook.start-host", d.state.OS.ExecPath+" callhook "+shared.VarPath("")+" "+shared.ShellQuote(d.Project().Name)+" "+shared.ShellQuote(d.Name())+" starthost")
```
simondeziel: The `initLXC` function calls those helper multiple times on the same 2 values, would it be worth saving the result in variables?

simondeziel (approved):

lxd/instance/drivers/driver_lxc.go
```diff
@@ -869,8 +869,11 @@ func (d *lxc) initLXC(config bool) (*liblxc.Container, error) {
 		return nil, err
 	}
 
+	projectShellQuoted := shared.ShellQuote(d.Project().Name)
+	instanceShellQuoted := shared.ShellQuote(d.Name())
+
 	// Call the onstart hook on start.
-	err = lxcSetConfigItem(cc, "lxc.hook.pre-start", fmt.Sprintf("/proc/%d/exe callhook %s %s %s start", os.Getpid(), shared.VarPath(""), strconv.Quote(d.Project().Name), strconv.Quote(d.Name())))
+	err = lxcSetConfigItem(cc, "lxc.hook.pre-start", fmt.Sprintf("/proc/%d/exe callhook %s %s %s start", os.Getpid(), shared.VarPath(""), projectShellQuoted, instanceShellQuoted))
```
simondeziel, line 876: Hmm, sometimes `fmt.Sprintf()` is used but sometimes it's just `+` concatenation. For another day though.

tomponline: Yep, ive not got time to do a complete refactor.

PR #17479, opened 2026-01-26
Entity name validation restrictions

tomponline:
Based on https://github.com/lxc/incus/pull/2358 and https://github.com/lxc/incus/pull/2827

### DIFF
--- a/shared/validate/validate.go
+++ b/shared/validate/validate.go
@@ -14,6 +14,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"unicode"
 
 	"github.com/google/uuid"
 	"github.com/kballard/go-shellquote"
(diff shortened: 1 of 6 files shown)


shared/validate/validate.go
```diff
@@ -509,14 +510,42 @@ func IsNetworkPortRange(value string) error {
...
+	if len(value) > 64 {
+		return errors.New("Maximum name length is 64 characters")
+	}
+
+	// Check for unicode characters.
+	for _, r := range value {
+		if unicode.IsSpace(r) {
+			return errors.New("Name cannot contain white space")
+		}
+	}
```
simondeziel, line 525: Should control characters also be blocked?
```
for _, r := range value {
    // IsSpace handles: \t, \n, \v, \f, \r, and ' '
    // IsControl handles: \x00 (Null), \x1b (Escape), etc.
    if unicode.IsSpace(r) || unicode.IsControl(r) {
        return errors.New("Name contains invalid or hidden characters")
    }
}
```

shared/validate/validate.go
```diff
@@ -509,14 +510,42 @@ func IsNetworkPortRange(value string) error {
...
+
+	// Check for special URL characters.
+	reservedChars := []string{"?", "&", "+", "\"", "'", "`", "*"}
```
simondeziel, line 528: I wonder if this shouldn't be extended even more for shell use safety:

```
reservedChars := []string{
    "?", "&", "+", "\"", "'", "`", "*", 
    ";", "|", ">", "<", "$", "(", ")", 
    "{", "}", "[", "]",
}

tomponline:
@simondeziel this is still draft so not ready for review yet.

markylaing:
@tomponline did this break too much stuff? Seems like it would be good to have

tomponline:
@markylaing i felt it would be better for MC 4.0 and not for an upgrade to LXD 6 LTS.

This came as part of a  security fix, but the it masked the actual security fix which I have included separately.

PR #17521, opened 2026-01-29
github: `actions/setup-microceph` fallback to using a loop device if no ephemeral disk is found

simondeziel:
To workaround https://github.com/actions/runner-images/issues/13587

### DIFF
--- a/.github/actions/setup-microceph/action.yml
+++ b/.github/actions/setup-microceph/action.yml
@@ -86,9 +86,11 @@ runs:
       run: |
           set -eux
           sudo "${GITHUB_ACTION_PATH}/setup-microceph.sh" install-ceph-common
+          sudo cat /etc/ceph/ceph.conf || true
 
     - name: Wait for MicroCeph to be ready
       shell: bash
       run: |
           set -eux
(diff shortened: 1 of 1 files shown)


tomponline (approved):
ta

.github/actions/setup-microceph/action.yml
tomponline, line 1: if GH are planning to remove ephemeral drive completely, why dont we always use a loop device?

simondeziel: For the GHA runners side of it, it might come to that at one point, I'm keeping an eye on it and will see how it evolves.

For local runners we want to keep the ephemeral drive support because those runners tend to be used for more than a single job. There setting up microceph from scratch between runs would be annoying.

tomponline: Thanks. It'd be good on the GH side to have it always use the same approach so that its simpler and more predictable, rather than seeing potentially different behaviour depending on the runner. Thanks for keeping an eye on it.

simondeziel: I was not ready to let go of testing with "real disks" but it make sense: https://github.com/canonical/lxd/pull/17545

Issue #16559, opened 2025-09-18
Bump minimum requirements for external dependencies to that of previous Ubuntu LTS


simondeziel:
### Describe the solution you'd like

The next LXD LTS will be released at the same time as Ubuntu 26.04 LTS will be. To limit how much backward compat we support, we should align the minimum requirements to be those of 24.04 LTS, so the prior Ubuntu LTS.

simondeziel:
Re-opening to also bump the minimum requirement on `liblxc`.

PR #17527, opened 2026-01-30
Simplify QEMU version handling

simondeziel:


lxd/instance/drivers/driver_qemu.go
```diff
@@ -9137,10 +9143,54 @@ func (d *qemu) checkFeatures(hostArch int, qemuPath string) (map[string]any, err
...
+func qemuVersionCached(qemuPath string) (string, error) {
+	qemuVersionOnce.Do(func() {
+		stdout, err := shared.RunCommandCLocale(qemuPath, "--version")
+		if err != nil {
+			qemuVersionErr = fmt.Errorf("Failed getting QEMU version: %w", err)
+			return
+		}
+
+		qemuOutput := strings.Fields(stdout)
+		if len(qemuOutput) >= 4 {
```
kadinsayani: Can we please include a comment on the output format here?

simondeziel: Good idea, done.

lxd/instance/drivers/driver_qemu.go
```diff
@@ -8920,20 +8932,14 @@ func (d *qemu) Info() instance.Info {
 		return data
 	}
 
-	stdout, stderr, err := shared.RunCommandSplit(context.TODO(), nil, nil, qemuPath, "--version")
+	qemuVersion, err := qemuVersionCached(qemuPath)
 	if err != nil {
-		logger.Errorf("Failed getting version during QEMU initialization: %v (%s)", err, stderr)
+		logger.Errorf("Failed getting version during QEMU initialization: %v", err)
```
tomponline, line 8925: I think the errors in this function should be warnings as this is called during LXD start up and qemu isn't required for LXD functioning.

lxd/instance/drivers/driver_qemu.go
tomponline, line 1: I'm not sure we need this change (or all of it) as ` DriverStatuses()` is already cached.

lxd/daemon.go
```diff
@@ -1147,6 +1147,9 @@ func (d *Daemon) init() error {
 		return err
 	}
 
+	// Cache QEMU version early during daemon startup.
+	_ = instanceDrivers.InitQemuVersionCache()
```
tomponline: This is already handled by ` DriverStatuses()` as is not needed.

tomponline (changes requested):
The ` DriverStatuses()` function calls each driver's `Info()` function and caches the result so subsequent calls to ` DriverStatuses()` do not call `Info()`.

So I question the need for this PR.

simondeziel:
> The ` DriverStatuses()` function calls each driver's `Info()` function and caches the result so subsequent calls to ` DriverStatuses()` do not call `Info()`.

You are entirely right, thanks for straighten things out.

> So I question the need for this PR.

There is much less substance than I initially anticipated but I've re-spin the PR to avoid the version parsing by having it saved to the `DriverStatus` struct.

It's also using the safer (IMHO) `RunCommandCLocale` helper for the version detection.

simondeziel:
> > So I question the need for this PR.
> 
> There is much less substance than I initially anticipated but I've re-spin the PR to avoid the version parsing by having it saved to the `DriverStatus` struct.

Actually, I think that having the driver version readily consumable will be handing for another PR I'm working where we'll enforce a minimum version of `liblxc` (https://github.com/canonical/lxd/issues/16559#issuecomment-3848558646).

kadinsayani (approved):

tomponline:
> It's also using the safer (IMHO) `RunCommandCLocale` helper for the version detection.

This makes sense indeed.

tomponline (approved):
ta

PR #17528, opened 2026-01-30
Do kernel version lookup during init

simondeziel:


lxd/sys/os.go
```diff
@@ -112,7 +112,7 @@ type OS struct {
 	BootTime    time.Time
 
 	// Version info
-	KernelVersion   version.DottedVersion
```
tomponline, line 115: Please can you revert this change, as the zero value for version.DottedVersion is usable as an indicator that there is no value (its all zeroes) and this avoids altering the data interchange format (the struct) towards the implementation of the `version.Parse` function.

It also avoids the need to check if the value is nil before comparing it.

lxd/firewall/firewall_load.go
```diff
@@ -3,12 +3,13 @@ package firewall
...
 	"github.com/canonical/lxd/shared/logger"
+	"github.com/canonical/lxd/shared/version"
 )
 
-// New returns an appropriate firewall implementation.
+// New returns an appropriate firewall implementation using the provided kernel version.
 // Uses xtables if nftables isn't compatible or isn't in use already, otherwise uses nftables.
-func New() Firewall {
-	nftables := drivers.Nftables{}
+func New(kernelVersion *version.DottedVersion) Firewall {
```
tomponline: Lets pass this is as a value rather than a pointer, so its clear its required and cannot be optional.

tomponline (approved):
Ta

PR #17530, opened 2026-02-02
doc: misc starter pack alignment

minaelee:
This PR adds minor changes to the doc/ folder to align with standard files in the [Canonical Sphinx Starter Pack repository](https://github.com/canonical/sphinx-docs-starter-pack) where possible without breaking LXD customizations and improvements. It also cleans up files and dependency installations that are no longer needed due to recent updates, such as removing installing the `mdl` (`markdownlint`) snap from the `tests.yml` workflow since it has been replaced with `pymarkdownlnt`. 

No docs functionality is altered with these changes. An update script and Makefile target provided by the starter pack are added; these can be useful in comparing changes in the future.

## Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1243,12 +1243,6 @@ jobs:
             doc/conf.py
           python-version: '3.x'  # satisfied by any 3.x version already installed
 
-      - name: Install dependencies
-        run: |
-          set -eux
-          sudo apt-get install aspell aspell-en
-          sudo snap install mdl
-
       - name: Build docs (Sphinx)
         run: |
           set -eux


tomponline:
As there is some python in this PR I'll await @simondeziel and @tugbataluy review. Thanks!

doc/.sphinx/update_sp.py
```diff
@@ -0,0 +1,271 @@
...
+
+# Initial update script for the starter pack.
+#
+# Requires some manual intervention, but makes identifying updates and differences easier.
+#
+# For debugging, please run this script with DEBUGGING=1
+# e.g. user@device:~/git/Canonical/sphinx-docs-starter-pack/docs$ DEBUGGING=1 python .sphinx/update_sp.py
+
+
+import glob
```
tugbataluy: It's a style convention but I think imports should be sorted alphabetically per PEP8, and standard library should be separated from third-party. My suggestion is:
```
import glob
import logging
import os
import re
import subprocess
import sys

import requests
from packaging.version import parse as parse_version
from requests.exceptions import RequestException
```

doc/.sphinx/update_sp.py
```diff
@@ -0,0 +1,271 @@
+#! /usr/bin/env python
```
tugbataluy: shebang expression should be updated as
`#! /usr/bin/env python3 ` 
for python version clarity.

doc/.sphinx/update_sp.py
```diff
@@ -0,0 +1,271 @@
...
+SPHINX_UPDATE_DIR = os.path.join(SPHINX_DIR, "update")
+GITHUB_REPO = "canonical/sphinx-docs-starter-pack"
+GITHUB_API_BASE = f"https://api.github.com/repos/{GITHUB_REPO}"
+GITHUB_API_SPHINX_DIR = f"{GITHUB_API_BASE}/contents/docs/.sphinx"
+GITHUB_RAW_BASE = f"https://raw.githubusercontent.com/{GITHUB_REPO}/main"
+
+TIMEOUT = 10  # seconds
+
+# Check if debugging
+if os.getenv("DEBUGGING"):
```
tugbataluy: This check seems to have no effect since when the log level is not set at all, it will still log debug messages.
For reference:
https://docs.python.org/3/library/logging.html#logging-levels

doc/.sphinx/update_sp.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+def main():
+    # Check local version
+    logging.debug("Checking local version")
+    try:
+        with open(os.path.join(SPHINX_DIR, "version")) as f:
+            local_version = f.read().strip()
+    except FileNotFoundError:
+        print("WARNING\nWARNING\nWARNING")
```
tugbataluy: My comment is about style preference again. I think triple "WARNING" print is a bit odd. I suggest using ```logging.warning()```.

doc/.sphinx/update_sp.py
```diff
@@ -0,0 +1,271 @@
...
+            logging.debug("Updated files found and downloaded")
+            print("Differences have been identified in static files.")
+            print("Updated files have been downloaded to '.sphinx/update'.")
+            print("Validate and move these files into your '.sphinx/' directory.")
+        else:
+            logging.debug("No files found to update")
+        # Provide information on NEW files
+        if new_files:
+            logging.debug("New files found and downloaded")
+            print(
```
tugbataluy: print() with multiple arguments adds spaces between them, resulting in:
```
NOTE: New files have been downloaded
 See 'NEWFILES.txt' for all downloaded files
 Validate and merge these files into your '.sphinx/' directory
```
If this is intentional then we can leave it as it is but if the expected is output like this:
```
NOTE: New files have been downloaded
See 'NEWFILES.txt' for all downloaded files
Validate and merge these files into your '.sphinx/' directory
```

you need to replace those lines with as you did in the other parts of the file:
```
print("NOTE: New files have been downloaded")
print("See 'NEWFILES.txt' for all downloaded files")
print("Validate and merge these files into your '.sphinx/' directory")
```
or you can use triple quotes:
```
print("""NOTE: New files have been downloaded.
See 'NEWFILES.txt' for all downloaded files
Validate and merge these files into your '.sphinx/' directory""")
```

doc/.sphinx/update_sp.py
```diff
@@ -0,0 +1,271 @@
...
+                "Validate and merge these files into your '.sphinx/' directory",
+            )
+        else:
+            logging.debug("No new files found to download")
+    else:
+        logging.debug("Local version and release version are the same")
+        print("This version is up to date.")
+
+    # Check requirements are the same
+    new_requirements = []
```
tugbataluy: I think, this variable is assigned but the initial [] is never used, it gets overwritten by the set operation. Can be removed.

doc/.sphinx/update_sp.py
```diff
@@ -0,0 +1,271 @@
...
+
+            new_requirements = requirements - local_reqs
+
+            for req in new_requirements:
+                logging.debug(f"{req} not found in local requirements.txt")
+
+            for req in requirements & local_reqs:
+                logging.debug(f"{req} already exists in local requirements.txt")
+
+            if new_requirements != set():
```
tugbataluy: Again it's just about style: In Python, an empty set is falsy, so `if new_requirements:` is the standard way to check if a set has elements. This works for sets, lists, dicts, etc.

tugbataluy: It can also applicable for 
`if new_file_list != []` you can use `if new_file_list:`

doc/.sphinx/update_sp.py
```diff
@@ -0,0 +1,271 @@
...
+        else:
+            if item["type"] == "file":
+                logging.debug(f"No local version found of {item['name']}")
+                download_file(
+                    item["download_url"], os.path.join(SPHINX_UPDATE_DIR, item["name"])
+                )
+                if item["name"] != "version":
+                    new_file_list.append(item["name"])
+    # Writes return value for parent function
+    if os.path.exists(os.path.join(SPHINX_UPDATE_DIR)):
```
tugbataluy: os.path.join() with single argument is unnecessary:
you can use:
`if os.path.exists(SPHINX_UPDATE_DIR):'`instead.

doc/.sphinx/update_sp.py
```diff
@@ -0,0 +1,271 @@
...
+
+# Examines local files
+def get_local_files_and_paths():
+    """Identify '.sphinx' local files and paths"""
+    logging.debug("Checking local files and paths")
+    try:
+        files = []
+        paths = []
+        patterns = [".*", "**.*"]
+        files, paths = [], []
```
tugbataluy: This line is duplicated so you can remove it.

doc/.sphinx/update_sp.py
```diff
@@ -0,0 +1,271 @@
...
+
+        for pattern in patterns:
+            for file in glob.iglob(os.path.join(SPHINX_DIR, pattern), recursive=True):
```
tugbataluy: There are some problems here:
- ".*" matches hidden files (starting with dot) like .gitignore, not all files.
- Pattern "**.*" with recursive=True would need **/ prefix to work correctly. As written, it matches files with dots in subdirectories but behavior is inconsistent.

My suggestion:
```
patterns = ["*", "**/*"]
for pattern in patterns:
    for filepath in glob.iglob(os.path.join(SPHINX_DIR, pattern), recursive=True):
        if os.path.isfile(filepath):
            files.append(os.path.basename(filepath))
            paths.append(filepath)
```

doc/.sphinx/update_sp.py
```diff
@@ -0,0 +1,271 @@
...
+                files.append(os.path.basename(file))
+                paths.append(file)
+        return files, paths
+    except Exception as e:
+        logging.debug(e)
+        raise RuntimeError("get_local_files_and_paths()") from e
+
+
+# General API query with timeout and RequestException
+def query_api(url):
```
tugbataluy: I believe this approach is a little bit problematic. Currently function returns response without checking status code. Should at least log non-200 responses:
```
def query_api(url):
    """Query an API with a globally set timeout"""
    logging.debug(f"Querying {url}")
    try:
        r = requests.get(url, timeout=TIMEOUT)
        r.raise_for_status()  # Raises HTTPError for 4xx/5xx
        return r
    except RequestException as e:
        raise RuntimeError(f"Failed query_api(): {url}") from e
```

doc/.sphinx/update_sp.py
```diff
@@ -0,0 +1,271 @@
...
+        print("You may experience issues using this functionality.")
+        logging.debug("No local version found. Setting version to None")
+        local_version = "None"
+    except Exception as e:
+        logging.debug(e)
+        raise Exception("ERROR executing check local version")
+    logging.debug(f"Local version = {local_version}")
+
+    # Check release version
+    latest_release = query_api(GITHUB_API_BASE + "/releases/latest").json()["tag_name"]
```
tugbataluy: This comment can be considered for both API calls  in the file. The current approach means:

- No check for HTTP status code (e.g, 404, 403 rate limit)
- No check if "tag_name" exists in response
- If GitHub API fails, this will crash with an unhelpful error.

My suggestion is:
```
response = query_api(GITHUB_API_BASE + "/releases/latest")
if response.status_code != 200:
    raise RuntimeError(f"Failed to fetch latest release: HTTP {response.status_code}")
    
data = response.json()
if "tag_name" not in data:
    raise RuntimeError(f"Unexpected API response: {data}")
    
latest_release = data["tag_name"]
```

doc/.sphinx/update_sp.py
```diff
@@ -0,0 +1,271 @@
...
+        with open(f"{SPHINX_DIR}/NEWFILES.txt", "w") as f:
+            for entry in new_file_list:
+                f.write(f"{entry}\n")
+        logging.debug("Some downloaded files are new")
+        return files_updated, True
+    return files_updated, False
+
+
+# Checks git hash of a file
+def get_git_revision_hash(file) -> str:
```
tugbataluy: No error handling. If file doesn't exist or git isn't available, this crashes.

My suggestion:
```
def get_git_revision_hash(file: str) -> str:
    """Get SHA of local files."""
    logging.debug(f"Getting hash of {os.path.basename(file)}")
    try:
        result = subprocess.check_output(
            ["git", "hash-object", file],
            stderr=subprocess.STDOUT,
            timeout=10
        ).decode("ascii").strip()
        if not result:
            raise ValueError(f"No hash generated for {file}")
        return result
    except (subprocess.CalledProcessError, FileNotFoundError, subprocess.TimeoutExpired) as e:
        raise RuntimeError(f"Failed to get git hash for {file}") from e
```

doc/.sphinx/update_sp.py
tugbataluy: or:

```
print(
    "NOTE: New files have been downloaded",
    "See 'NEWFILES.txt' for all downloaded files",
    "Validate and merge these files into your '.sphinx/' directory",
    sep="\n"
)
```

minaelee: Hi Tugba, thanks for your review! The update_sp.py script is from the upstream starter pack and only needs to be used locally, occasionally. It's not required at all for the docs build. Thus, I think it might be easiest not to include it at all in the repository and only grab it from upstream when needed.

minaelee: Hi Tugba, thanks for your review! The update_sp.py script is from the upstream starter pack and only needs to be used locally, occasionally. It's not required at all for the docs build. Thus, I think it might be easiest not to include it at all in the repository and only grab it from upstream when needed.

minaelee:
> As there is some python in this PR I'll await @simondeziel and @tugbataluy review. Thanks!

With much thanks to @tugbataluy for her review, I removed the python script from this PR since it is maintained upstream and not necessary for the docs build; I will consider applying her comments upstream. Upon reconsideration, I think it will be simpler to grab the script as needed from upstream during cyclic maintenance/updates than include it in this repository.

PR #17540, opened 2026-02-02
Client: Fix waiting for operations during copy

roosterfish:
### Overview

Whilst working on the tests for https://github.com/canonical/lxd/pull/17466 which will trigger the instance start on the server instead of the client, I found that the client currently doesn't wait for both source and target operations when performing a migration for both instances and custom volumes.

This caused the server side start in https://github.com/canonical/lxd/pull/17466 to be performed after the client has already returned.

In addition I found when copying an instance from a snapshot that results in a migration, the wait logic for the source operation is missing. Same for custom volumes, thanks for the hint @tomponline.

### Fix

As we share the same "wait" logic all across the client code base, I have now consolidated everything together into a `waitForSplitOperation` func which accepts both an already existing source operation and a callback func for the target operation.
This approach allows reducing lots of the complexity (additional channels and go routines) and reuses the same logic for all migration scenarios.

### Potential future improvements

When the target operation fails, the code currently tries to cancel the source operation. This is currently defunct as the `_ = op.Cancel()` operation doesn't return immediately and - since it's running inside a `go` routine - the `lxc` client already exits before this call can complete:

1) Try cancel: https://github.com/canonical/lxd/blob/main/client/lxd_instances.go#L779-L783
2) Premature exit once the `chConnect` chan is closed https://github.com/canonical/lxd/blob/main/client/lxd_instances.go#L798

Therefore I have omitted adding this logic also to `waitForSplitOperation`. The operations will time out anyway on the source (current behavior). We might want to allow immediate cancel in the future if possible.

### DIFF
--- a/client/lxd_storage_volumes.go
+++ b/client/lxd_storage_volumes.go
@@ -457,19 +457,15 @@ func (r *ProtocolLXD) MigrateStoragePoolVolume(pool string, volume api.StorageVo
 	return op, nil
 }
 
-func (r *ProtocolLXD) tryMigrateStoragePoolVolume(source InstanceServer, pool string, req api.StorageVolumePost, urls []string) (RemoteOperation, error) {
+func (r *ProtocolLXD) tryMigrateStoragePoolVolume(source InstanceServer, pool string, req api.StorageVolumePost, urls []string, op Operation) (RemoteOperation, error) {
 	if len(urls) == 0 {
 		return nil, errors.New("The source server isn't listening on the network")
-	}
-
-	rop := remoteOperation{
-		chDone: make(chan bool),
 	}
 
 	operation := req.Target.Operation
 
 	// Forward targetOp to remote op
-	go func() {
+	targetOpFunc := func(rop *remoteOperation) error {
 		success := false
 		var errors []remoteOperationResult
 		for _, serverURL := range urls {
@@ -482,10 +478,7 @@ func (r *ProtocolLXD) tryMigrateStoragePoolVolume(source InstanceServer, pool st
 				continue
 			}
 
-			rop := remoteOperation{
-				targetOp: top,
-				chDone:   make(chan bool),
-			}
+			rop.targetOp = top
 
 			for _, handler := range rop.handlers {
 				_, _ = rop.targetOp.AddHandler(handler)
@@ -507,13 +500,13 @@ func (r *ProtocolLXD) tryMigrateStoragePoolVolume(source InstanceServer, pool st
 		}
 
 		if !success {
-			rop.err = remoteOperationError("Failed storage volume creation", errors)
+			return remoteOperationError("Failed storage volume creation", errors)
 		}
 
-		close(rop.chDone)
-	}()
+		return nil
+	}
 
-	return &rop, nil
+	return r.startSplitRemoteOperation(targetOpFunc, op), nil
 }
 
 // tryCreateStoragePoolVolume attempts to create a storage volume in the specified storage pool.
@@ -703,7 +696,7 @@ func (r *ProtocolLXD) CopyStoragePoolVolume(pool string, source InstanceServer, 
 		target.Certificate = info.Certificate
 		sourceReq.Target = &target
 
-		return r.tryMigrateStoragePoolVolume(source, sourcePool, sourceReq, info.Addresses)
+		return r.tryMigrateStoragePoolVolume(source, sourcePool, sourceReq, info.Addresses, op)
 	}
 
 	// Get source server connection information


tomponline:
@roosterfish may you rebase so we can get a clear test run

roosterfish:
@tomponline looking better now, ready for review.

client/lxd_instances.go
```diff
@@ -794,13 +794,20 @@ func (r *ProtocolLXD) tryCreateInstance(req api.InstancesPost, urls []string, op
...
+		// Finish the remote operation only after both sub-operations are done.
+		defer func() {
+			rop.err = err
+			close(rop.chDone)
+		}()
 
-		rop.err = err
-		close(rop.chDone)
+		// Wait for the actual instance creation on the target.
+		err = <-chConnect
```
tomponline: what happens to the `err` value if its overwritten in the following stanza? 

Should it be recorded? Also does chWait get closed if chConnect returns a non-nil error?

client/lxd_instances.go
```diff
@@ -2169,7 +2169,7 @@ func (r *ProtocolLXD) CopyInstanceSnapshot(source InstanceServer, instanceName s
 		target.Certificate = info.Certificate
 		sourceReq.Target = &target
 
-		return r.tryMigrateInstanceSnapshot(source, cName, sName, sourceReq, info.Addresses)
+		return r.tryMigrateInstanceSnapshot(source, cName, sName, sourceReq, info.Addresses, op)
```
tomponline, line 2095: these changes can be squashed into the following commit as they are related and in the same file.

client/lxd_instances.go
```diff
@@ -2317,11 +2321,42 @@ func (r *ProtocolLXD) tryMigrateInstanceSnapshot(source InstanceServer, instance
 			break
 		}
 
-		if !success {
-			rop.err = remoteOperationError("Failed instance migration", errors)
+		if success {
+			chConnect <- nil
```
tomponline: can you just close chConnect as a read from it will return the default zero value, rather than posting nil to it.

client/lxd_instances.go
roosterfish: Mh we could return a combined error message in case both operations return an error?

client/lxd_instances.go
roosterfish: Yeah good idea, fixed.

client/lxd_instances.go
roosterfish: Fixed.

client/lxd_instances.go
roosterfish: Fixed. Went with a combined error `"%w (Source operation failure: %w)"`. Happy to change the format if you have a better idea. Haven't found anything similar in the code base.

roosterfish:
I went with running manual tests as I haven't found a straightforward approach to trigger migration failures on source/target.

1) Trigger immediate failure on target:
  The source waits for control connection (default timeout)

   ```
   julian@thinkpad:~$ lxc cp c1 mc:c1 
   Error: Failed instance creation: Always fail migration on target (Source operation failure: Failed waiting for migration control connection on source: context deadline exceeded)
   ```

2) Trigger immediate failure on source:

   ```
   julian@thinkpad:~$ lxc cp c1 mc:c1
   Error: Failed instance creation: Error transferring instance data: Failed waiting for migration control connection on target: websocket: bad handshake (Source operation failure: Always fail migration on source)
   ```

3) Trigger immediate failure on both source and target:

   ```
   julian@thinkpad:~$ lxc cp c1 mc:c1
   Error: Failed instance creation: Always fail migration on target (Source operation failure: Always fail migration on source)
    ```

tomponline:
> I went with running manual tests as I haven't found a straightforward approach to trigger migration failures on source/target.
> 
>     1. Trigger immediate failure on target:
>        The source waits for control connection (default timeout)
>        ```
>        julian@thinkpad:~$ lxc cp c1 mc:c1 
>        Error: Failed instance creation: Always fail migration on target (Source operation failure: Failed waiting for migration control connection on source: context deadline exceeded)
>        ```
> 
>     2. Trigger immediate failure on source:
>        ```
>        julian@thinkpad:~$ lxc cp c1 mc:c1
>        Error: Failed instance creation: Error transferring instance data: Failed waiting for migration control connection on target: websocket: bad handshake (Source operation failure: Always fail migration on source)
>        ```
> 
>     3. Trigger immediate failure on both source and target:
>        ```
>        julian@thinkpad:~$ lxc cp c1 mc:c1
>        Error: Failed instance creation: Always fail migration on target (Source operation failure: Always fail migration on source)
>        ```

Excellent thank you!

roosterfish:
Mh, whilst adding the same set of changes to the custom volume migration, I found that when trying to copy the same vol/instance again, we are running into the control connection timeout on the source:

```bash
julian@thinkpad:~$ lxc cp c1 mc:c1
julian@thinkpad:~$ lxc cp c1 mc:c1                     
Error: Failed instance creation: Instance "c1" already exists (Source operation failure: Failed waiting for migration control connection on source: context deadline exceeded)
```

So the outcome of the target operation decides how to deal with the source op. Looking into it.

roosterfish:
Ready for another review. 

I came up with an alternative fix that even reduced the overall lines of code and complexity which was in there before. Please see the updated PR description.

client/lxd.go
```diff
@@ -501,3 +501,35 @@ func (r *ProtocolLXD) getSourceImageConnectionInfo(source ImageServer, image api
...
 	return info, nil
 }
+
+// waitForSplitOperation takes a remoteOperation as input and ensures that both the target operation func and source operation
+// are executed before marking the remoteOperation as done.
+// In case the target operation func is returning an error, this indicates we don't anymore have to wait for the source operation
+// and will cancel it.
+func (r *ProtocolLXD) waitForSplitOperation(rop *remoteOperation, targetOpFunc func() error, sourceOp Operation) RemoteOperation {
+	go func() {
+		var err error
```
tomponline, line 516: err doesnt seem to be used

tomponline: oh wait its used in defer, hard to see.

client/lxd.go
```diff
@@ -501,3 +501,35 @@ func (r *ProtocolLXD) getSourceImageConnectionInfo(source ImageServer, image api
 
 	return info, nil
 }
+
+// waitForSplitOperation takes a remoteOperation as input and ensures that both the target operation func and source operation
+// are executed before marking the remoteOperation as done.
+// In case the target operation func is returning an error, this indicates we don't anymore have to wait for the source operation
+// and will cancel it.
+func (r *ProtocolLXD) waitForSplitOperation(rop *remoteOperation, targetOpFunc func() error, sourceOp Operation) RemoteOperation {
```
tomponline: I would expect a function called "wait...()" to be blocking on something, but this function does not.

tomponline: maybe it shouldnt return the `rop` at all, and instead it can be called before `rop` is returned by the caller, and we could name this `StartSplitOperationCompleter()`?

client/lxd.go
roosterfish: Yeah I was thinking on this a bit but found it more clear to first assign the error from both source and target operation to individual vars first and moving it back to `err` later. 
I'll add a comment for future readers.

client/lxd.go
roosterfish: At the end I went with `startSplitRemoteOperation` which creates and returns the `remoteOperation`.
This also allows removing the `remoteOperation` declaration (with the done chan) from each of the client funcs that use `startSplitRemoteOperation`.

client/lxd.go
```diff
@@ -501,3 +501,41 @@ func (r *ProtocolLXD) getSourceImageConnectionInfo(source ImageServer, image api
 
 	return info, nil
 }
+
+// startSplitRemoteOperation starts a new remote operation and ensures that both the target operation func and source operation
+// are executed before marking the returned remote operation as done.
+// The new remote operation gets passed as an argument to the provided target operation func.
+// In case the target operation func is returning an error, this indicates we don't anymore have to wait for the source operation
+// and will cancel it.
```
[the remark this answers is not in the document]
roosterfish: Ah yes, leftover from before I realized cancelling the source operation is useless, see the PR desc comment.

client/lxd.go
```diff
@@ -501,3 +501,40 @@ func (r *ProtocolLXD) getSourceImageConnectionInfo(source ImageServer, image api
...
+			rop.err = err
+			close(rop.chDone)
+		}()
+
+		targetOpErr := targetOpFunc(rop)
+		if targetOpErr != nil {
+			err = targetOpErr
+			return
+		}
+
```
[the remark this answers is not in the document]
roosterfish: Fixed and removed the extra err var.

client/lxd.go
```diff
@@ -501,3 +501,40 @@ func (r *ProtocolLXD) getSourceImageConnectionInfo(source ImageServer, image api
...
+		// Target operation was successful, now wait and check the source operation if provided.
+		if sourceOp != nil {
+			sourceOpErr := sourceOp.Wait()
+			if sourceOpErr != nil {
+				err = sourceOpErr
```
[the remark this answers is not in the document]
roosterfish: Fixed and removed the extra err var.

PR #17544, opened 2026-02-02
Make instance stop async on DELETE /1.0/instances when force=1

kadinsayani:
Follow-up to https://github.com/canonical/lxd/pull/17473.

### DIFF
--- a/lxd/instance_delete.go
+++ b/lxd/instance_delete.go
@@ -92,7 +92,9 @@ func instanceDelete(d *Daemon, r *http.Request) response.Response {
 }
 
 // doInstanceDelete deletes an instance in the given project.
-// If the instance is running and force is true, the instance is force stopped. Otherwise, an error is returned.
+// If the instance is running and force is true, the instance is force stopped asynchronously
+// as part of the delete operation. If the instance is running and force is false, the request
+// fails before the operation is created.
 func doInstanceDelete(ctx context.Context, s *state.State, name string, projectName string, force bool) (*operations.Operation, error) {
 	requestor, err := request.GetRequestor(ctx)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


test/suites/basic.sh
```diff
@@ -586,7 +586,8 @@ test_basic_usage() {
   lxc delete foo -f
 
   lxc launch testimage deleterunning
-  my_curl -X DELETE "https://${LXD_ADDR}/1.0/containers/deleterunning" | grep "Instance is running"
+  op="$(my_curl -X DELETE "https://${LXD_ADDR}/1.0/containers/deleterunning" | jq --exit-status --raw-output .operation)"
+  my_curl "https://${LXD_ADDR}${op}/wait" | jq --exit-status '.error | contains("Instance is running")'
```
simondeziel: There's the `wait_for` helper that should be handy here.

kadinsayani: Thanks!

doc/api-extensions.md
```diff
@@ -2888,3 +2888,7 @@ The LVM storage driver has the specific `lvm.vg.force_reuse` configuration key f
 ## `instance_force_delete`
 
 This adds support for a `force` query parameter to the `DELETE /1.0/instances/{name}` endpoint. When set, running instances will be forcibly stopped before deletion.
+
+## `instance_delete_async_stop`
+
+The server always performs any required stop inside the delete operation rather than synchronously before the operation is created. As a result, deleting a running instance without `force=1` now returns an asynchronous operation that will fail, instead of a synchronous error.
```
tomponline: Hrm, that sounds like the reason to keep it as-is, sorry.

markylaing: Is it not possible to check the state of the instance and validating the request before returning the operation?

tomponline: Mmm, true we could do that indeed.

kadinsayani: I've added a pre-check. When force=1, the stop is performed with the delete operation asynchronously. Otherwise, a synchronous error is returned.

lxd/instance_delete.go
```diff
@@ -104,31 +106,31 @@ func doInstanceDelete(ctx context.Context, s *state.State, name string, projectN
...
-		}
+	// Pre-check for immediate 400 when force isn't set.
+	if inst.IsRunning() && !force {
+		return nil, api.NewStatusError(http.StatusBadRequest, "Instance is running")
 	}
 
 	rmct := func(ctx context.Context, op *operations.Operation) error {
-		if deletedOnStop {
-			return nil
+		if inst.IsRunning() {
```
tomponline: Lets use the result of the earlier `inst.IsRunning` call, as its a relatively expensive call to make.

doc/api-extensions.md
```diff
@@ -2888,3 +2888,8 @@ The LVM storage driver has the specific `lvm.vg.force_reuse` configuration key f
 ## `instance_force_delete`
 
 This adds support for a `force` query parameter to the `DELETE /1.0/instances/{name}` endpoint. When set, running instances will be forcibly stopped before deletion.
+
+(extension-instance-delete-async-stop)=
+## `instance_delete_async_stop`
```
tomponline: No need for an API change for this now you've added the pre-check.

lxd/instance_delete.go
kadinsayani: Fixed.

doc/api-extensions.md
kadinsayani: Ack, thanks.

lxd/instance_delete.go
```diff
@@ -104,31 +106,32 @@ func doInstanceDelete(ctx context.Context, s *state.State, name string, projectN
...
+	instRunning := inst.IsRunning()
+	if instRunning && !force {
+		return nil, api.NewStatusError(http.StatusBadRequest, "Instance is running")
 	}
 
 	rmct := func(ctx context.Context, op *operations.Operation) error {
-		if deletedOnStop {
-			return nil
+		if instRunning {
+			if !force {
```
tomponline: Is it possible to get into this section, won't it be prevented by the earlier check?

kadinsayani: Yeah we can remove that redundant inner check now. I included it earlier in case there is an instance state change between the first check and the start of the operation, hence why there were 2 separate `inst.IsRunning()` calls.

tomponline (approved):
ta

PR #17545, opened 2026-02-02
github: always use loop devices for `actions/setup-microceph`

simondeziel:


simondeziel:
@tomponline CI didn't run to completion due to known GHA problems.

tomponline:
@simondeziel please can you rebase so we can get a clear run?

Issue #16039, opened 2025-07-15
Image properties incorrectly taken from image in another project


markylaing:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

22.04

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
Local edge build (995a5bedbb)
```

### Output of "lxc info" or system info if it fails

```shell
-
```

### Issue description

`lxc image copy` gets image metadata before downloading the image file. If the fingerprint of the image is available locally, the local image is used instead. However, the image properties are copied from the local image rather than taken from the image metadata at the source.

Also, note that the output of `lxc image list --all-projects` is incorrect in the output below. We should see the same image fingerprint twice in the list. Once for each project containing the image.

cc @tomponline @kadinsayani 

### Steps to reproduce

```bash
$ lxc project list
+-------------------+--------+----------+-----------------+-----------------+----------+---------------+---------------------+---------+
|       NAME        | IMAGES | PROFILES | STORAGE VOLUMES | STORAGE BUCKETS | NETWORKS | NETWORK ZONES |     DESCRIPTION     | USED BY |
+-------------------+--------+----------+-----------------+-----------------+----------+---------------+---------------------+---------+
| default (current) | YES    | YES      | YES             | YES             | YES      | YES           | Default LXD project | 2       |
+-------------------+--------+----------+-----------------+-----------------+----------+---------------+---------------------+---------+
mark@RUBIX:~/projects/canonical/lxd/test$ lxc image list
+-------+-------------+--------+-------------+--------------+------+------+-------------+
| ALIAS | FINGERPRINT | PUBLIC | DESCRIPTION | ARCHITECTURE | TYPE | SIZE | UPLOAD DATE |
+-------+-------------+--------+-------------+--------------+------+------+-------------+
$ lxc image copy images:alpine/3.20 local: --alias alpine
Image copied successfully!                   
$ lxc image show alpine | sed -e "s/description: .*/description: HELLO/" | lxc image edit alpine
$ lxc image show alpine
auto_update: false
properties:
  architecture: amd64
  description: HELLO
  os: Alpine
  release: "3.20"
  requirements.secureboot: "false"
  serial: "20250715_0016"
  type: squashfs
  variant: default
public: false
expires_at: 1970-01-01T00:00:00Z
profiles:
- default
$ lxc project create foo
Project foo created
[...]

PR #17546, opened 2026-02-02
Client: Preserve source image properties and filename for the copied image

nmezhenskyi:
This PR ensures that the `lxc` client always fetches the source image metadata when running `lxc image copy` command. It also ensures that the `ProtocolLXD.CopyImage()` function applies the provided source image filename and properties to the copied image. 

Fixes https://github.com/canonical/lxd/issues/16577, https://github.com/canonical/lxd/issues/16039.

### DIFF
--- a/test/suites/remote.sh
+++ b/test/suites/remote.sh
@@ -406,8 +406,32 @@ test_remote_usage() {
   # The alias should be set to `bar`.
   [ "${alias}" = "bar" ]
 
+  echo "==> Ensure that the copied image properties are set from source image."
+
+  # Check that the copied image has the same filename as the source image.
+  filename=$(lxc_remote query localhost:"/1.0/images/${fingerprint}" | jq --exit-status '.filename')
+  lxc_remote query lxd2:"/1.0/images/${fingerprint}" | jq --exit-status --raw-output ".filename == ${filename}"
+
+  # Now, change the description property for the downloaded image in the default project.
+  lxc_remote image set-property lxd2:bar description "TEST"
+
+  # Create another project to download the image into.
+  lxc_remote project create lxd2:foo
+
+  # Copy the same image from source into newly created project.
+  lxc_remote image copy --quiet localhost:testimage lxd2: --alias bar --target-project foo
+
+  # Check that the downloaded image in the new project has the description property from source.
+  [ "$(lxc_remote image get-property lxd2:bar description --project foo)" = "$(lxc_remote image get-property localhost:testimage description)" ]
+
+  # Check that the downloaded image in the default project still has custom description.
+  [ "$(lxc_remote image get-property lxd2:bar description --project default)" = "TEST" ]
+
+  # Clean up.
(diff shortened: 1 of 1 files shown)


test/suites/remote.sh
```diff
@@ -406,6 +406,10 @@ test_remote_usage() {
   # The alias should be set to `bar`.
   [ "${alias}" = "bar" ]
 
+  # Check that the copied image has the same filename as the source image.
+  filename=$(lxc_remote query localhost:"/1.0/images/${fingerprint}" | jq .filename)
+  [ "${filename}" = "$(lxc_remote query lxd2:"/1.0/images/${fingerprint}" | jq .filename)" ]
```
simondeziel: Please always use `jq --exit-status`... the linter to flag this has yet to be added to `test/lint` but it will

For the comparison, please use a `jq` assertion instead of a shell comparison, something like this:

```
lxc_remote query lxd2:"/1.0/images/${fingerprint}" | jq --exit-status --raw-output ".filename == "${filename}"
```

nmezhenskyi: Updated

test/suites/remote.sh
```diff
@@ -406,8 +406,32 @@ test_remote_usage() {
   # The alias should be set to `bar`.
   [ "${alias}" = "bar" ]
 
+  echo "==> Ensure that the copied image properties are set from source image."
+
+  # Check that the copied image has the same filename as the source image.
+  filename=$(lxc_remote query localhost:"/1.0/images/${fingerprint}" | jq .filename)
```
simondeziel:
```suggestion
  filename=$(lxc_remote query localhost:"/1.0/images/${fingerprint}" | jq --exit-status '.filename')
```

This doesn't do much on its own (well, makes sure it got valid JSON) but it future proves it for when we add that linter enforcing all `jq` invocations have `--exit-status`.

nmezhenskyi: Sorry, missed this one

nmezhenskyi: Updated

simondeziel (approved):
LGTM (for the tests) thanks!

test/suites/remote.sh
```diff
@@ -406,8 +406,32 @@ test_remote_usage() {
   # The alias should be set to `bar`.
   [ "${alias}" = "bar" ]
 
+  echo "==> Ensure that the copied image properties are set from source image."
+
+  # Check that the copied image has the same filename as the source image.
+  filename=$(lxc_remote query localhost:"/1.0/images/${fingerprint}" | jq --exit-status '.filename')
+  lxc_remote query lxd2:"/1.0/images/${fingerprint}" | jq --exit-status --raw-output ".filename == ${filename}"
```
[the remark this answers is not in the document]
nmezhenskyi: `jq` returns `false` is the assertion fails.

tomponline (approved):
Thanks!

PR #17553, opened 2026-02-03
Add `pro` feature flag (stable-5.0)

simondeziel:


lxd/ubuntupro/client_test.go
```diff
@@ -0,0 +1,31 @@
...
+		mockProCLI := proCLIMock{
+			mockAttached: true,
+		}
+
+		s := &Client{}
+		s.init(mockProCLI)
+
+		assert.Contains(t, version.UserAgent, "pro")
+	})
+}
```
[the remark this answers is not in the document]
simondeziel: That's a good suggestion that I've taken also for the `main` branch: https://github.com/canonical/lxd/pull/17555

shared/version/useragent.go
```diff
@@ -59,7 +62,17 @@ func UserAgentStorageBackends(backends []string) {
...
+func UserAgentFeatures(features []string) error {
+	hasWhiteSpace := func(s string) bool {
+		return strings.ContainsFunc(s, unicode.IsSpace)
+	}
+
+	if slices.ContainsFunc(features, hasWhiteSpace) {
+		return errors.New("User agent features may not contain whitespace")
+	}
+
+	userAgentFeatures = append(userAgentFeatures, features...)
```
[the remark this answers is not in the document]
simondeziel: There are only a few call sites and they should not result in duplicated flags. Accumulating flags along the way is desired here so it works as intended.

.github/workflows/tests.yml
```diff
@@ -304,66 +304,9 @@ jobs:
...
-          for _ in $(seq 60); do
-            if sudo microceph.ceph pg stat | grep -wF unknown; then
-              sleep 1
-            else
-              break
-            fi
-          done
-          sudo microceph.ceph status
-          sudo rm -f /snap/bin/rbd
+        uses: canonical/lxd/.github/actions/setup-microceph@main
```
tomponline, line 307: I'm not too keen on intra-branch dependencies, would you mind copying across the current setup-microceph implementation to this branch as well?

simondeziel: I don't particularly mind but that feels like more maintenance than necessary. This specific action is already used by other repos (like `lxd-ci` for example) using the `@main`.

In an ideal world, those actions would have been put in `canonical/lxd-gh-actions` or similar to avoid this busy work.

Anyway, if you firmly prefer `stable-5.0` to not depend on actions from the `main` branch, would it be OK to copy the action files in bulk and not bother with the per-commit backporting?

tomponline: > would it be OK to copy the action files in bulk and not bother with the per-commit backporting?

Yes that is fine.

> lxd-ci

Ultimately I'd like to see that repo go away.

I'm mainly concerned that, like the tests in lxd-ci, if the actions have to cater for many LTS series they will have to get overly complicated, whereas if we can maintain branch specific versions we can keep them doing that they need - and if that is identical then we can just copy it from main into the LTS branches.

simondeziel: Sounds good, done.

tomponline, line 307: we should probably use a microceph LTS edge channel associated with the recommended series for MicroCloud 2, so as not to be testing against latest/edge, but rather the appropriate LTS edge channel - same goes for LXD 5.21 if not already doing so.

tomponline: The same way we pin the equivalent version of liblxc and dqlite.

tomponline (approved):
Thanks!

Do we also need changes to the snap packaging with a pro wrapper?

simondeziel:
> Do we also need changes to the snap packaging with a pro wrapper?

Yep, that's next on the list. Then 4.0

.github/workflows/tests.yml
simondeziel: OK, I'll use the MicroCeph and MicroOVN channels that MicroCloud 2 uses. Will do this for 5.21 and below even if below isn't directly relevant with MicroCloud it's still additional exposure.

tomponline: nice one thanks

simondeziel:
> Do we also need changes to the snap packaging with a pro wrapper?

https://github.com/canonical/lxd-pkg-snap/pull/1038

PR #17556, opened 2026-02-03
scripts/check-and-commit: Fix git commit flags

kadinsayani:
The combined -Ssm is parsed as -S with arg sm, causing -s and -m to be ignored and the commit message to be treated as a pathspec.

### DIFF
--- a/scripts/check-and-commit.sh
+++ b/scripts/check-and-commit.sh
@@ -17,7 +17,7 @@ if [ -t 0 ]; then
   if [ "${answer:-y}" = "y" ] || [ "${answer:-y}" = "Y" ]; then
     # Commit changes and exit
     # shellcheck disable=SC2086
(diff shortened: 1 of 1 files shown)


scripts/check-and-commit.sh
```diff
@@ -17,7 +17,7 @@ if [ -t 0 ]; then
   if [ "${answer:-y}" = "y" ] || [ "${answer:-y}" = "Y" ]; then
     # Commit changes and exit
     # shellcheck disable=SC2086
-    exec git commit -Ssm "${MESSAGE}" -- ${FILES}
+    exec git commit -S -s -m "${MESSAGE}" -- ${FILES}
```
simondeziel, line 20: Good catch, when I moved the whole thing to a separated script, I clustered the `-S` and `-sm` bits... which as you found out fools git into thinking `sm` is a key ID :/

simondeziel (approved):
Thanks!

PR #17561, opened 2026-02-04
lxd: Set placement group name variable on instance creation.

markylaing:
While working on other things my IDE warned that the `placementGroupName == ""` predicate guarding whether volatile.cluster.group key on the instance configuration is always true, because the variable is never assigned a value.

We should double check that this is the behaviour we want. IIRC we separated cluster groups from placement groups so I'm not certain what the relation is here.

## Checklist


kadinsayani:
> We should double check that this is the behaviour we want. IIRC we separated cluster groups from placement groups so I'm not certain what the relation is here.

When a placement group is set, we don't need to set `volatile.cluster.group`.

tomponline (approved):
Ta

@kadinsayani this suggests a gap in our tests, would you mind adding a test to ensure we dont regress this logic? Thanks

kadinsayani:
> Ta
> 
> @kadinsayani this suggests a gap in our tests, would you mind adding a test to ensure we dont regress this logic? Thanks

https://github.com/canonical/lxd/pull/17573

PR #17565, opened 2026-02-04
Require `liblxc` 5.0.0 or newer

simondeziel:
Fixes https://github.com/canonical/lxd/issues/16559

### DIFF
--- a/lxd/sys/os.go
+++ b/lxd/sys/os.go
@@ -89,7 +89,6 @@ type OS struct {
 	// Kernel features
 	BPFToken                bool // BPFToken indicates support for BPF token delegation mechanism.
 	CloseRange              bool // CloseRange indicates support for the close_range syscall.
-	ContainerCoreScheduling bool // ContainerCoreScheduling indicates LXC and kernel support for core scheduling.
 	CoreScheduling          bool // CoreScheduling indicates support for core scheduling syscalls.
 	IdmappedMounts          bool // IdmappedMounts indicates kernel support for VFS idmap.
 	NativeTerminals         bool // NativeTerminals indicates support for TIOGPTPEER ioctl.


lxd/cgroup/file.go
```diff
@@ -58,7 +58,7 @@ func NewFileReadWriter(pid int, unifiedCapable bool) (*CGroup, error) {
 		return nil, err
 	}
 
-	cg.UnifiedCapable = unifiedCapable
+	cg.UnifiedCapable = true // cgroup2: introduced in lxc 4.0.0
```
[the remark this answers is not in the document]
simondeziel: No, it was introduced with lxc 4.0.0 so the comment is valid.

lxd/instance/drivers/driver_lxc.go
```diff
@@ -7204,7 +7163,7 @@ func (d *lxc) cgroup(cc *liblxc.Container, running bool) (*cgroup.CGroup, error)
 		return nil, err
 	}
 
-	cg.UnifiedCapable = liblxc.HasAPIExtension("cgroup2")
+	cg.UnifiedCapable = true // cgroup2: introduced in lxc 4.0.0
```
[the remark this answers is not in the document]
simondeziel: No, it was introduced with lxc 4.0.0 so the comment is valid.

lxd/instance/drivers/driver_lxc.go
```diff
@@ -651,16 +651,11 @@ func (d *lxc) initLXC(config bool) (*liblxc.Container, error) {
 		return nil, err
 	}
 
-	if d.state.OS.ContainerCoreScheduling {
+	if d.state.OS.CoreScheduling {
 		err = lxcSetConfigItem(cc, "lxc.sched.core", "1")
 		if err != nil {
 			return nil, err
```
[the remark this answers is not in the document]
simondeziel: We'll never get there due to the `lxc` driver being unavailable in the unlikely case that someone tried to use LXC < 5.0.0.

lxd/seccomp/seccomp.go
```diff
@@ -2678,10 +2670,6 @@ func lxcSupportSeccompNotify(state *state.State) error {
 		return errors.New("Seccomp notify not supported")
 	}
 
-	if !state.OS.LXCFeatures["seccomp_notify"] {
-		return errors.New("LXC doesn't support seccomp notify")
-	}
-
 	c, err := liblxc.NewContainer("test-seccomp", state.OS.LxcPath)
 	if err != nil {
 		return errors.New("Failed to load seccomp notify test container")
```
kadinsayani:
```suggestion
		return errors.New("Failed loading seccomp notify test container")
```

While you're here ;)

lxd/instance/drivers/driver_lxc.go
```diff
@@ -6149,7 +6149,8 @@ func (d *lxc) Exec(req api.InstanceExecPost, stdin *os.File, stdout *os.File, st
 
 	// Prepare the subcommand
 	cname := project.Instance(d.Project().Name, d.Name())
-	args := []string{
+	args := make([]string, 0, 12+len(req.Command)+len(envSlice))
```
kadinsayani: What does 12 represent here?

simondeziel: There are 8 args added in the append right after, then comes `--`, `env` and `--`, `cmd` and their respective slices.

Did I summed them up wrong?

lxd/seccomp/seccomp.go
simondeziel: Done

lxd/instance/drivers/driver_lxc.go
kadinsayani: Sum is correct, but a comment on what 12 represents may be helpful :)

lxd/seccomp/seccomp.go
kadinsayani: Thanks!

lxd/instance/drivers/driver_lxc.go
simondeziel: I fear this comment will inevitably drift over time so I'd like to avoid it.

Would it be sufficiently self descriptive to avoid the need for a comment?

```go
args := make([]string, 0, 8+2+len(req.Command)+2+len(envSlice))
```

simondeziel: This mimics what's done in https://github.com/canonical/lxd/blob/main/shared/cert.go#L267-L270

kadinsayani: How about:

```go
  baseArgs := []string{
        d.state.OS.ExecPath,
        "forkexec",
        cname,
        d.state.OS.LxcPath,
        configPath,
        req.Cwd,
        strconv.FormatUint(uint64(req.User), 10),
        strconv.FormatUint(uint64(req.Group), 10),
  }

  args := make([]string, 0, len(baseArgs)+2+len(envSlice)+2+len(req.Command))
  args = append(args, baseArgs...)
  ```

Comment (kadinsayani) [reply lxd/instance/drivers/driver_lxc.go]: Self-describing and easy to change :)

Comment (simondeziel) [reply lxd/instance/drivers/driver_lxc.go]: Doesn't that go against the prealloc goal of reducing the amount of allocations/re-allocations? Isn't `baseArgs` pure overhead?

Comment (kadinsayani) [reply lxd/instance/drivers/driver_lxc.go]: Yeah but it's quite minimal. If you want to avoid the extra alloc, this works too:

```go
const fixedArgs = 8 // number of args appended below before the separators
args := make([]string, 0, fixedArgs+2+len(envSlice)+2+len(req.Command))
```

kadinsayani: Happy with your suggested approach too

kadinsayani: Or 
```go
sep = "--"
args := make([]string, 0, 8+len(sep)+len(envSlice)+len(sep)+len(req.Command))
```

Super nit picky but I don't like the `8+2` bit :laughing:

kadinsayani: Feel free to ignore this suggestion if you'd prefer going with the `12`. Since the `12` is just used for the prealloc, even if the args change at some point, at worst, we have an extra alloc and not an actual bug.

simondeziel: I seriously hope the compiler has the smart to do the arithmetic first so the runtime only sees a `12` plus a bunch of string sizes.

kadinsayani: The compiler will constant-fold 8+2+2 to 12 either way. Only the `len(<string>)` calls are evaluated at runtime.

kadinsayani: IMHO `const fixedArgs = 8; args := make([]string, 0, fixedArgs+2+len(envSlice)+2+len(req.Command))` is the best option for readability.

simondeziel: It's a subjective matter I guess. We'll need a tie breaker to settle this one ;)

lxd/instance/drivers/driver_lxc.go
```diff
@@ -656,11 +656,6 @@ func (d *lxc) initLXC(config bool) (*liblxc.Container, error) {
 		if err != nil {
 			return nil, err
 		}
-	} else if d.state.OS.CoreScheduling {
```
tomponline, line 659: isnt this a kernel level feature rather than a liblxc feature?

lxd/main_forkexec.go
tomponline, line 1: can this be squashed into the earlier commit to keep the logic change together

tomponline (commented):
The core scheduling changes need more explanation in order for me to understand and review them. Thanks

lxd/instance/drivers/driver_lxc.go
simondeziel: The reverted commit (7ca2b76e01a04787c2c5f308dea830937957dc20) message helps get more context:
>     lxd: support core scheduling for container even without LXC library support
>     
>     Let the forkcoresched hook be useable from a hook in LXD. This allows us
>     to turn on core scheduling for the main container workload even when the
>     LXC shared library doesn't support core scheduling.

simondeziel:
> The core scheduling changes need more explanation in order for me to understand and review them. Thanks


The code now sets `d.os.CoreScheduling` solely based on `canUseCoreScheduling()`.
`canUseCoreScheduling()` performs a functional test where a process is forked and a core scheduling cookie is tentatively obtained from the kernel. If it works, `core_scheduling_aware = true`.

This means the availability of core scheduling depends solely on that of the underlying kernel as **verified at runtime**.

# Container support

In `driver_lxc.go`, if `d.state.OS.CoreScheduling` is true, it sets `lxc.sched.core = 1`. This is safe because the container wouldn't even start with that config set if the kernel was to not support core scheduling.

As such the fallback logic is dead code that's dropped by this PR.

# VM support

In `driver_qemu.go`, `d.setCoreSched(pids)` is used and this function checks `d.state.OS.CoreScheduling` and calls `forkcoresched` as needed to pin the QEMU threads/processes.

This logic is unchanged and still correctly gated by the kernel check.

# `lxc exec` safety

Previously, if a system had `CoreScheduling` and a recent LXC (`ContainerCoreScheduling = true`), `forkexec` would be passed `0` for the `coresched` param. This meant that core scheduling was delegated to `liblxc`.

Now with this PR, core scheduling is always delegated to `liblxc`'s attach function if the kernel supports it. Since we verified that `liblxc >= 5.0.0` is loaded, `liblxc` will automatically adopt the core scheduling settings from the container's payload processes (or create a new domain correctly) during attach.

By ensuring that `liblxc >= 5.0.0` (which implies `ContainerCoreScheduling` is native), the runtime behavior for `lxc exec` is identical to what it was before.

# Kernel lacking core scheduling support

If the kernel check fails, `d.state.OS.CoreScheduling = false` then `lxc.sched.core` is not set for containers and `setCoreSched` returns immediately for VMs.

This means there is nothing we can do about it, same as before this PR.

PR #17462, opened 2026-01-22
Schema changes for durable operations

skozina:
This PR contains complete DB schema changes needed for durable operations.
It depends on (or rather contains) changes from #17589, and is needed for #17045.

## Checklist

- [ ] I have read the [contributing guidelines](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md) and attest that all commits in this PR are [signed off](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#including-a-signed-off-by-line-in-your-commits), [cryptographically signed](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#commit-signature-verification), and follow this project's [commit structure](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#commit-structure).
- [ ] I have checked and added or updated relevant documentation.

PR #17567, opened 2026-02-05
Operation metadata type

markylaing:
We have `shared.ParseMetadata` which parses an any into a
`map[string]any`. This is not called anywhere except in the
operations package. Checking operation create call sites, only a
`map[string]any` (or `shared.JMap`) is used.

It doesn't make sense to perform this parsing. Let's just enforce
that all operation metadata is a JSON object.

Sorry to @skozina if this creates more rebase headaches.

## Checklist

### DIFF
--- a/lxd-agent/exec.go
+++ b/lxd-agent/exec.go
@@ -171,8 +171,8 @@ type execWs struct {
 }
 
 // Metadata returns the metadata for the operation.
-func (s *execWs) Metadata() any {
-	fds := shared.Jmap{}
+func (s *execWs) Metadata() map[string]any {
+	fds := make(map[string]string, len(s.fds))
 	for fd, secret := range s.fds {
 		if fd == execWSControl {
 			fds[api.SecretNameControl] = secret
@@ -181,7 +181,7 @@ func (s *execWs) Metadata() any {
 		}
 	}
 
-	return shared.Jmap{
+	return map[string]any{
 		"fds":         fds,
 		"command":     s.command,
 		"environment": s.env,


tomponline:
@markylaing do we need `map[string]any` or could it be `map[string]string`?

tomponline:
> @markylaing do we need `map[string]any` or could it be `map[string]string`?

Seeing as @skozina is planning to store metadata in a DB table soon with key/value being defined as TEXT.

markylaing:
> > @markylaing do we need `map[string]any` or could it be `map[string]string`?
> 
> Seeing as @skozina is planning to store metadata in a DB table soon with key/value being defined as TEXT.

I think it still needs to be map[string]any. E.g. The instance exec metadata looks like this:
```json
{
    "fds": {
        "control": "<secret>",
        "1": "<secret>",
        "2": "<secret>"
    },
    "command": ["sh", "-c", "whoami"],
    "interactive": true,
    "environment": {
        "EDITOR": "vim"
    }
}
```

markylaing:
It's also `map[string]any` in the api.Operation

tomponline:
> > > @markylaing do we need `map[string]any` or could it be `map[string]string`?
> > 
> > 
> > Seeing as @skozina is planning to store metadata in a DB table soon with key/value being defined as TEXT.
> 
> I think it still needs to be map[string]any. E.g. The instance exec metadata looks like this:
> 
> ```json
> {
>     "fds": {
>         "control": "<secret>",
>         "1": "<secret>",
>         "2": "<secret>"
>     },
>     "command": ["sh", "-c", "whoami"],
>     "interactive": true,
>     "environment": {
>         "EDITOR": "vim"
>     }
> }
> ```

Thanks. 

So is that another blocker to making all operations DB oriented?

tomponline:
> It's also `map[string]any` in the api.Operation

We could potentially tighten the SDK definition here, so that is not a blocker.

tomponline (approved):
ta

tomponline:
I'd personally like to see Jmap removed from the codebase entirely.

markylaing:
> > > > @markylaing do we need `map[string]any` or could it be `map[string]string`?
> > > 
> > > 
> > > Seeing as @skozina is planning to store metadata in a DB table soon with key/value being defined as TEXT.
> > 
> > 
> > I think it still needs to be map[string]any. E.g. The instance exec metadata looks like this:
> > ```json
> > {
> >     "fds": {
> >         "control": "<secret>",
> >         "1": "<secret>",
> >         "2": "<secret>"
> >     },
> >     "command": ["sh", "-c", "whoami"],
> >     "interactive": true,
> >     "environment": {
> >         "EDITOR": "vim"
> >     }
> > }
> > ```
> 
> Thanks.
> 
> So is that another blocker to making all operations DB oriented?

I don't see it that way - I would JSONify all values in the table

tomponline:
> > > > > @markylaing do we need `map[string]any` or could it be `map[string]string`?
> > > > 
> > > > 
> > > > Seeing as @skozina is planning to store metadata in a DB table soon with key/value being defined as TEXT.
> > > 
> > > 
> > > I think it still needs to be map[string]any. E.g. The instance exec metadata looks like this:
> > > ```json
> > > {
> > >     "fds": {
> > >         "control": "<secret>",
> > >         "1": "<secret>",
> > >         "2": "<secret>"
> > >     },
> > >     "command": ["sh", "-c", "whoami"],
> > >     "interactive": true,
> > >     "environment": {
> > >         "EDITOR": "vim"
> > >     }
> > > }
> > > ```
> > 
> > 
> > Thanks.
> > So is that another blocker to making all operations DB oriented?
> 
> I don't see it that way - I would JSONify all values in the table

Hrm, that'd be quite different to other key/value tables in LXD.

Some do today have json encoded values, but these are then returned by the API as string-literals.

But you're thinking that the `value` field in https://github.com/canonical/lxd/pull/17462/changes#diff-a15876f82899798b44f27be9568d0261586296cb95d2a0de0fdd583e598ea05eR161 should also be json encoded and json decoded back into `map[string]any`? Is @skozina aligned on this too?

markylaing:
> > > > > > @markylaing do we need `map[string]any` or could it be `map[string]string`?
> > > > > 
> > > > > 
> > > > > Seeing as @skozina is planning to store metadata in a DB table soon with key/value being defined as TEXT.
> > > > 
> > > > 
> > > > I think it still needs to be map[string]any. E.g. The instance exec metadata looks like this:
> > > > ```json
> > > > {
> > > >     "fds": {
> > > >         "control": "<secret>",
> > > >         "1": "<secret>",
> > > >         "2": "<secret>"
> > > >     },
> > > >     "command": ["sh", "-c", "whoami"],
> > > >     "interactive": true,
> > > >     "environment": {
> > > >         "EDITOR": "vim"
> > > >     }
> > > > }
> > > > ```
> > > 
> > > 
> > > Thanks.
> > > So is that another blocker to making all operations DB oriented?
> > 
> > 
> > I don't see it that way - I would JSONify all values in the table
> 
> Hrm, that'd be quite different to other key/value tables in LXD.
> 
> Some do today have json encoded values, but these are then returned by the API as string-literals.
> 
> But you're thinking that the `value` field in https://github.com/canonical/lxd/pull/17462/changes#diff-a15876f82899798b44f27be9568d0261586296cb95d2a0de0fdd583e598ea05eR161 should also be json encoded and json decoded back into `map[string]any`? Is @skozina aligned on this too?

Well - I haven't thought about this in depth if I'm honest, I just figured this is the only way to do it if we can't change the type of the API field and we're set on having a separate key/value table for operation metadata.

Personally I don't see a reason for a separate table. I'd have a metadata column directly in the operations table as a JSON blob.

markylaing:
@tomponline let's start this thread on the durable operation schema PR (#17462 )

tomponline:
> Personally I don't see a reason for a separate table. I'd have a metadata column directly in the operations table as a JSON blob.

I concur. I think this is another side effect of @skozina initial approach to try and land durable operations separately from task operations, where we could enforce that durable operations only have map[string]string metadata.

But ultimately we do want to align the operation handling for both as closely as possible, otherwise we'll need to maintain effectively two different implementations.

PR #17603, opened 2026-02-09
Auth: Fix operation cancellation permissions

markylaing:
The `Permission` function defined on `operationtype.Type` defines the permissions required to cancel an operation. Unfortunately, not all cases are covered. The `DELETE /1.0/operations/{uuid}` interprets this as "no permissions required" and lets the caller continue. 

It's currently possible for any authenticated caller to cancel e.g. cluster member evacuation.

Prior to https://github.com/canonical/lxd/pull/17016, only select few operations were actually cancellable. So this hasn't been a problem until now.

This PR makes it so that only the requestor or an administrator with `can_edit` on `server` is able to cancel an operation. It also cleans up the `operationtype` package to enforce consistency.

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md) and attest that all commits in this PR are [signed off](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#including-a-signed-off-by-line-in-your-commits), [cryptographically signed](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#commit-signature-verification), and follow this project's [commit structure](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#commit-structure).
- [x] I have checked and added or updated relevant documentation.

PR #17607, opened 2026-02-09
Audit operation "resources"

markylaing:
- Changes operation resources to use an `entity.Type` key.
- Changes resources to only use valid references (e.g. don't reference entities that don't exist yet).
- Adds an `EntityURL` field to `operations.OperationArgs` and validates that it matches the entity type defined for the operation type.

**Reviewers note:** We should *only* use the effective project for any URLs (either the primary URL or any resources). Otherwise we will not be able to resolve the URL to an ID (because project features make LXD hallucinate). I didn't realise this until halfway through changing the resources so that's why this PR is still in draft.

Depends on #17603

## Checklist

### DIFF
--- a/lxc/export.go
+++ b/lxc/export.go
@@ -142,6 +142,8 @@ func (c *cmdExport) run(cmd *cobra.Command, args []string) error {
 	if d.HasExtension("operation_metadata_entity_url") {
 		backupName, _, err = getEntityFromOperationMetadata(op.Get().Metadata)
 	} else {
+		// Use "backups" here and not "entity.TypeInstanceBackup" because the change to use entity type names happened
+		// after the operation_metadata_entity_url extension.
 		backupName, _, err = getEntityFromOperationResources(op.Get().Resources, "backups")
 	}
 
--- a/lxc/init.go
+++ b/lxc/init.go
@@ -418,6 +418,8 @@ func (c *cmdInit) create(conf *config.Config, args []string, launch bool) (lxd.I
 		if d.HasExtension("operation_metadata_entity_url") {
 			name, _, err = getEntityFromOperationMetadata(opInfo.Metadata)
 		} else {
+			// Use "instances"/"containers" here and not "entity.TypeInstance"/"entity.TypeContainer" because the change
+			// to use entity type names happened after the operation_metadata_entity_url extension.
 			name, _, err = getEntityFromOperationResources(opInfo.Resources, "instances", "containers")
 		}
 
--- a/lxc/storage_volume.go
+++ b/lxc/storage_volume.go
@@ -2765,6 +2765,8 @@ func (c *cmdStorageVolumeExport) run(cmd *cobra.Command, args []string) error {
 	if d.HasExtension("operation_metadata_entity_url") {
 		backupName, _, err = getEntityFromOperationMetadata(op.Get().Metadata)
 	} else {
+		// Use "backups" here and not "entity.TypeStorageVolumeBackup" because the change to use entity type names
+		// happened after the operation_metadata_entity_url extension.
 		backupName, _, err = getEntityFromOperationResources(op.Get().Resources, "backups")
 	}
 


markylaing:
@tomponline @skozina FYI this is taking a while because usage of operation types and resources is pretty inconsistent. 

One example: we are using the same operation type for an image download token (`GET /1.0/images/{fingerprint}/export`) and for untrusted image upload (`POST /1.0/images`). One should use the project as it's primary entity because the image doesnt exist yet, whereas the other should use the image itself. Both need to store the image fingerprint in the operation metadata instead of the resources field as a URL.

Another example: We are using the "Instance{Action}" operation types when bulk updating the state of many instances. This will need to be a parent operation (I've added "InstanceStateUpdateBulk") with sub-operations using the correct operation type.

(and so on...)

markylaing:
This *should* :crossed_fingers: pass all tests now. If not, it's very close and I would appreciate a review. Thanks.

markylaing:
@tomponline as far as I can tell none of the test failures are caused by the changes in this PR.

lxd/operations/operations.go
```diff
@@ -151,6 +151,22 @@ func operationCreate(s *state.State, requestor *request.Requestor, args Operatio
...
+			return nil, fmt.Errorf("Invalid operation entity URL: %w", err)
+		}
+
+		if entityType != operationEntityType {
+			return nil, fmt.Errorf("Entity type for URL %q does not match operation entity type %q", args.EntityURL, operationEntityType)
+		}
+	} else if operationEntityType != entity.TypeServer {
+		return nil, errors.New("Operation entity URL required")
+	} else {
+		args.EntityURL = entity.ServerURL()
```
tomponline, line 177: just a nit, not a blocker, but either this line or the logic in EntityURL():

```
if op.entityURL == nil {
		return entity.ServerURL()
	}
```

seems redundant/duplicated.

lxd/operations/operations.go
```diff
@@ -151,6 +151,22 @@ func operationCreate(s *state.State, requestor *request.Requestor, args Operatio
 		return nil, errors.New("LXD is shutting down")
 	}
 
+	operationEntityType := args.Type.EntityType()
```
tomponline, line 164: just a nit not a blocker, but starting this new section with a comment similar to the commit message would be helpful to future readers and help break up the code into sections.

lxd/operations/operations.go
```diff
@@ -501,7 +501,7 @@ func (op *Operation) Render() (string, *api.Operation, error) {
 		for key, value := range resources {
 			var values = make([]string, 0, len(value))
 			for _, c := range value {
-				values = append(values, c.Project(op.Project()).String())
```
tomponline, line 484: where will the project come from now, is it already inside `c.String()`?

lxd/images.go
```diff
@@ -5239,13 +5239,27 @@ func createTokenResponse(s *state.State, r *http.Request, projectName string, fi
 	maps.Copy(meta, metadata)
 
 	meta["secret"] = secret
-
-	resources := map[string][]api.URL{}
-	resources["images"] = []api.URL{*api.NewURL().Path(version.APIVersion, "images", fingerprint)}
+	meta["fingerprint"] = fingerprint
```
tomponline, line 5235: An idea for future PR, not a blocker, but for common/shared metadata fields (like this one, the secret one above, and the entity name one we introduced recently) it would be good to have shared constants in the `api` package we can refer to on client and server side in order to easily see usage.

On that point, could this use `entity_name` rather than introduce a new field?

lxd/images.go
```diff
@@ -4778,6 +4769,7 @@ func imageExportPost(d *Daemon, r *http.Request) response.Response {
 
 	opArgs := operations.OperationArgs{
 		ProjectName: projectName,
+		EntityURL:   api.NewURL().Path(version.APIVersion, "projects", details.image.Fingerprint),
```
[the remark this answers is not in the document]
tomponline: @markylaing thoughts?

lxd/api_project.go
```diff
@@ -1027,10 +1027,10 @@ func projectPost(d *Daemon, r *http.Request) response.Response {
 	}
 
 	args := operations.OperationArgs{
-		ProjectName: "",
```
tomponline, line 1030: should it not have a projectName too?

lxd/instance_snapshot.go
```diff
@@ -564,19 +561,11 @@ func snapshotPut(s *state.State, r *http.Request, snapInst instance.Instance) re
...
-	resources["instances"] = []api.URL{*api.NewURL().Path(version.APIVersion, "instances", parentName)}
-	resources["instances_snapshots"] = []api.URL{*api.NewURL().Path(version.APIVersion, "instances", parentName, "snapshots", snapName)}
-
-	if snapInst.Type() == instancetype.Container {
-		resources["containers"] = resources["instances"]
-	}
-
 	args := operations.OperationArgs{
 		ProjectName: snapInst.Project().Name,
+		EntityURL:   api.NewURL().Path(version.APIVersion, "instances", parentName, "snapshots", snapName),
```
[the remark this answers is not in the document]
tomponline: @markylaing thoughts?

lxd/images.go
markylaing: Very good point. We can definitely standardise operation metadata further.

In future I think what would be most ideal is to have the API response contain the operation type and make the metadata a `json.RawMessage`. The caller should unmarshal into the concrete type associated with the operation type.

For now, I think our only option is to validate it (like we've recently added for `entity_name`). Maybe we can improve on this to include per operationtype validation.

markylaing: I've created an issue. For now I'l change this PR to use entity_name.

lxd/operations.go
```diff
@@ -60,54 +61,58 @@ var operationWebsocket = APIEndpoint{
...
 
-		// If the current operations has a hold on some resources, we keep track of them in the `resourceMap`.
-		// This is used to mark instances and storage volumes as busy to avoid shutting them down / unmounting them prematurely.
-		// This avoids the situation where a single very long running operation can block the shutdown of unrelated instances and the unmount of unrelated storage volumes.
-		for resourceName, resourceEntries := range opAPI.Resources {
-			if resourceName != "instances" {
-				continue
+		opEqual := opEqualsFunc(op)
+		if op.Type().EntityType() == entity.TypeInstance {
+			u := withProject(op.EntityURL())
```
tomponline: why would the op.EntityURL not have the correct project set and needs a `withProject` fixup?

This is the sort of valuable info that should be in a comment in my view.

lxd/operations.go
```diff
@@ -60,54 +61,58 @@ var operationWebsocket = APIEndpoint{
...
+			if !slices.ContainsFunc(res[u], opEqual) {
```
tomponline: It took me quite a while to figure out what you're trying to do here, because its quite abstracted and non-obvious code. But are you just de-duping the URL entries here, so that if an instance is referenced by multiple operations its not included twice? 

A comment here would have been much appreciated.

Also, for clarity, could we make `res := make(map[string][]*operations.Operation)` be `res := make(map[url.URL][]*operations.Operation)` perhaps?

lxd/operations/operations.go
markylaing: Yeah I think I forgot about one while writing the other

lxd/operations/operations.go
markylaing: Yes. The caller should set the project in the resources or entity URL.

lxd/images.go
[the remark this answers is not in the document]
markylaing: Whoopsie. I need to review where `ImageDownload` is used. I think it's for when LXD downloads an image (and not when a client exports an image). If so then it's correct that it should still be a project URL and I need to revert to using `projectName`

lxd/instance_snapshot.go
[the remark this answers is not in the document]
markylaing: Yep. Copilot really keeping me honest today

lxd/api_project.go
markylaing: I thought so, but that actually led me down a rabbit hole yesterday. If you add a project name, then the project rename works server side but the CLI hangs forever!! I found some weirdness and cruft in the client on the back of this that I'll address in another PR.

On balance though, a project rename doesn't happen *inside* a project. It is a server level operation. So I think it should stay as is.

tomponline: comments in code welcome for future travellers.

lxd/operations.go
markylaing: I'll add comments. 

The caller is iterating over instances (potentially loaded from disk and not from DB) and creating URLs to key into this map. I think this could all be made more clear by having a `map[string]map[string][]*operations.Operation` where the first key is project name and the second is instance name. Then we don't need to parse and pass URLs around.

lxd/operations.go
markylaing: Because in general (except for permissions) we display URLs like `/1.0/instances/{name}` without a query parameter if it exists in the default project. The calling function expects the project parameter to always be there because of the way it generates URLs.

I'll clean this up further with a map[{project}]map[{instance}][]*operations.Operation.

tomponline: > The calling function expects the project parameter to always be there because of the way it generates URLs.

Thanks, if that needs to change to make it clearer that could also be an option.

lxd/operations.go
tomponline: sounds good. thanks

lxd/images.go
markylaing: I've added a constant for `entity_url` (it was `_url` not `_name`) but I haven't used it here.

In writing the comment for `entity_url`, I realised that its' usage here is different. `entity_url` is supposed to be used when creating a resource when the client isn't required to specify a name. Here we are creating an image upload or download token where the caller knows what the entity URL is.

lxd/images.go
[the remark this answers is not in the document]
markylaing: I've added an `ImageUpload` operation type. `ImageDownload` is used when LXD downloads an image from another server. `ImageUpload` is when LXD uploads an image to another server. `ImageUpload` is linked to a specific image. `ImageDownload` is linked to a project (because the image doesn't exist yet).

`ImageDownload` was being used in both cases.

lxd/operations.go
markylaing: Made this change. Hopefully it's all a lot clearer now.

markylaing:
@skozina @tomponline all comments addressed. Please review when you can. Thanks

lxd/images.go
tomponline: Thanks makes sense. Shall we have a constant for fingerprint too so we can see its usage?

lxd/instances.go
```diff
@@ -456,21 +456,30 @@ func instancesOnDisk(s *state.State) ([]instance.Instance, error) {
 }
 
 // isInstanceBusy checks if the instance is currently busy: if it has an associated operation that is in a running state.
-func isInstanceBusy(inst instance.Instance, instancesToOps map[string]*operations.Operation, instancesToOpsMu *sync.Mutex) bool {
-	if instancesToOps == nil {
+func isInstanceBusy(inst instance.Instance, instanceOperations map[string]map[string][]*operations.Operation, instancesToOpsMu *sync.Mutex) bool {
+	if instanceOperations == nil {
```
tomponline: this is redundant isnt it, as `len()` of a nil map is 0, so is covered by the below statement.

tomponline:
@markylaing can you check over copilots comments please. Ta

lxd/storage_volumes_snapshot.go
```diff
@@ -937,11 +946,13 @@ func storagePoolVolumeSnapshotTypePatch(d *Daemon, r *http.Request) response.Res
 		return doStoragePoolVolumeSnapshotUpdate(ctx, s, details.pool, effectiveProjectName, dbVolume.Name, details.volumeType, req, op)
 	}
 
-	resources := map[string][]api.URL{}
-	resources["storage_volume_snapshots"] = []api.URL{*api.NewURL().Path(version.APIVersion, "storage-pools", details.pool.Name(), "volumes", details.volumeTypeName, details.volumeName, "snapshots", snapshotName)}
+	resources := map[entity.Type][]api.URL{
+		entity.TypeStorageVolume: {*api.NewURL().Path(version.APIVersion, "storage-pools", details.pool.Name(), "volumes", details.volumeTypeName, details.volumeName).Project(effectiveProjectName)},
```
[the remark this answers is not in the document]
markylaing: This is wrong, the URL correctly points to a volume.

lxd/storage_volumes_snapshot.go
```diff
@@ -1061,11 +1072,13 @@ func storagePoolVolumeSnapshotTypeDelete(d *Daemon, r *http.Request) response.Re
 		return details.pool.DeleteCustomVolumeSnapshot(effectiveProjectName, fullSnapshotName, op)
 	}
 
-	resources := map[string][]api.URL{}
-	resources["storage_volume_snapshots"] = []api.URL{*api.NewURL().Path(version.APIVersion, "storage-pools", details.pool.Name(), "volumes", details.volumeTypeName, details.volumeName, "snapshots", snapshotName)}
+	resources := map[entity.Type][]api.URL{
+		entity.TypeStorageVolumeSnapshot: {*api.NewURL().Path(version.APIVersion, "storage-pools", details.pool.Name(), "volumes", details.volumeTypeName, details.volumeName).Project(effectiveProjectName)},
```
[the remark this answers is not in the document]
markylaing: The snapshot URL is already the primary entity URL. Fixing by correcting the entity type to `entity.TypeStorageVolume`

lxd/instances.go
markylaing: Fixed.

markylaing:
Should be all fixed now.

PR #17608, opened 2026-02-09
VM: Add `limits.max_bus_ports` config option

nmezhenskyi:
This PR introduces a new configuration key for virtual machines `limits.max_bus_ports`. This option controls the maximum allowed number of user configurable devices requiring a dedicated `PCI/PCIe` port for a virtual machine. The number includes both the devices attached before the instance start and the devices hotplugged at runtime.

The default value for `limits.max_bus_ports` is 8 ports.

The PR also includes a DB patch that counts the `volatile.*.bus` settings and adjusts the `limits.max_bus_ports` value for existing VMs that exceed the default.

### DIFF
--- a/test/suites/vm.sh
+++ b/test/suites/vm.sh
@@ -135,6 +135,37 @@ test_vm_pcie_bus() {
   lxc exec v1 -- mount -t virtiofs config /mnt
   ! lxc exec v1 -- touch /mnt/foo || false
 
+  sub_test "Check that limits.max_bus_ports config option enforces the number of allowed PCIe devices"
+
+  # The default value for "limits.max_bus_ports" is 8 ports.
+  # Fill all the available slots with PCIe devices.
+  for i in $(seq 1 5); do
+    lxc config device add v1 "aaa${i}" nic nictype=p2p
+  done
+
+  # Check that another device cannot be hotplugged because no more available PCIe ports left.
+  ! lxc config device add v1 aaa6 nic nictype=p2p || false
+
+  # Stop the VM and attach an additional device.
+  lxc stop -f v1
+  lxc config device add v1 aaa6 nic nictype=p2p
+
+  # Check that the instance start fails because PCIe devices limit is exceeded.
+  ! lxc start v1 || false
+
+  # Increase the PCIe devices limit.
+  lxc config set v1 limits.max_bus_ports=10
+
+  # Start the instance, this time there should enough PCIe ports.
+  lxc start v1
+  waitInstanceReady v1
+
+  # Hotplug an additional device, there should still be one available PCIe port.
+  lxc config device add v1 aaa7 nic nictype=p2p
+
+  # Check that another device cannot be hotplugged because no more available PCIe ports left.
+  ! lxc config device add v1 aaa8 nic nictype=p2p || false
+
   # Coverage data requires clean lxd-agent stop
   prepare_vm_for_hard_stop v1
   lxc delete --force v1


lxd/instance/drivers/driver_qemu.go
```diff
@@ -3676,6 +3701,10 @@ func (d *qemu) generateQemuConfigFile(cpuInfo *cpuTopology, mountInfo *storagePo
 			return nil, "", "", false, fmt.Errorf("Bus allocation not supported for bus type %q", bus.name)
 		}
 
+		if usedSlots >= int(pciSlotCountMax) {
+			return nil, "", "", false, fmt.Errorf("PCI devices limit reached (%s=%d)", "limits.vm.max_bus_ports", pciSlotCountMax)
```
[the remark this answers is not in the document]
nmezhenskyi: Updated

test/suites/vm.sh
```diff
@@ -136,6 +136,37 @@ test_vm_pcie_bus() {
   lxc exec v1 -- mount -t virtiofs config /mnt
   ! lxc exec v1 -- touch /mnt/foo || false
 
+  echo "==> Check that limits.vm.max_bus_ports config option enforces the number of allowed PCI devices"
```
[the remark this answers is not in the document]
nmezhenskyi: Updated

lxd/instance/drivers/driver_qemu.go
```diff
@@ -2266,6 +2266,24 @@ func (d *qemu) getPCISlotCount() (pciSlots uint8, err error) {
 	return pciSlots, nil
 }
 
+// getMaxPCISlotCount returns the maximum allowed number of PCI slots for the instance.
+func (d *qemu) getMaxPCISlotCount() (pciSlotCountMax uint8, err error) {
+	// The default value for "limits.vm.max_bus_ports" is 8 slots.
+	pciSlotCountMax = uint8(8)
```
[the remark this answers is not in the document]
nmezhenskyi: Fixed

doc/api-extensions.md
```diff
@@ -2906,3 +2906,9 @@ The field is only included when a resource is being created asynchronously (oper
 For synchronous resource creation, clients should inspect the `Location` header.
 
 The `resources` field should no longer be relied upon for this information.
+
+(extension-vm-limits-max-bus-ports)=
+## `vm_limits_max_bus_ports`
+
+This introduces support for the {config:option}`instance-resource-limits:limits.vm.max_bus_ports` configuration key for virtual machines. This option controls the maximum number of user configurable PCI devices allowed for a virtual machine.
```
tomponline: Is it just PCI? It covers PCIe and others too right?

lxd/instance/drivers/driver_qemu.go
```diff
@@ -92,6 +92,9 @@ const QEMUDefaultCPUCores = 1
 // QEMUDefaultMemSize is the default memory size for VMs if no limit specified.
 const QEMUDefaultMemSize = "1GiB"
 
+// QEMUDefaultMaxBusPorts is the default number of PCI ports available for VMs.
+const QEMUDefaultMaxBusPorts uint8 = 8
```
tomponline, line 96: does it need to be exported?

lxd/patches.go
```diff
@@ -2040,4 +2043,79 @@ DELETE FROM storage_pools_config WHERE key = "ceph.osd.force_reuse"
...
+		return pciDevices
+	}
+
+	return s.DB.Cluster.Transaction(s.ShutdownCtx, func(ctx context.Context, tx *db.ClusterTx) error {
+		return tx.InstanceList(ctx, func(inst db.InstanceArgs, project api.Project) error {
+			if inst.Type != instancetype.VM {
+				return nil
+			}
+
+			pciDevices := countPCIDevices(inst.Config)
```
tomponline: is this taking into account expanded devices from profiles - it probably should if not.

lxd/patches.go
```diff
@@ -2040,4 +2043,79 @@ DELETE FROM storage_pools_config WHERE key = "ceph.osd.force_reuse"
...
+			_, hasCustomLimitSet := inst.Config["limits.vm.max_bus_ports"]
```
tomponline: is this taking into account expanded config from profiles - it probably should if not.

lxd/patches.go
```diff
@@ -2040,4 +2043,79 @@ DELETE FROM storage_pools_config WHERE key = "ceph.osd.force_reuse"
...
+
+			// Only update the limit if the instance currently does not have it set
+			// and the number of attached PCI devices is higher than the default value.
+			if !hasCustomLimitSet && pciDevices > int(instanceDrivers.QEMUDefaultMaxBusPorts) {
```
tomponline: ah it does need to be exported

lxd/instance/drivers/driver_qemu.go
nmezhenskyi: Yes, it's also used in the patch

lxd/patches.go
nmezhenskyi: Good point, currently it does not account for expanded devices, I'll update it

doc/api-extensions.md
nmezhenskyi: > Is it just PCI? It covers PCIe and others too right?

@tomponline Yes, it covers devices taking up a whole single PCI or PCIe slot (including NICs, GPUs, Infiniband devices, storage devices using `virtio-blk` or NVMe). I wonder what would the correct description be?

Perhaps:
> "This option controls the maximum allowed number of user configurable devices requiring a dedicated `PCI/PCIe` port for a virtual machine."

What do you think? Am I missing something?

simondeziel: The Q35 machine type comes with only PCIe, no regular/legacy PCI is available unless a bridge is added for those (`pcie-pci-bridge`). At least that's my understanding.

As such, I think we should stick to the PCIe terminology. Granted, QEMU driver names are not helping because they all have the `-pci` suffix (like `virtio-rng-pci`) but AFAIK, this is because of historical reason with the `440fx` machine type that used regular PCI.

lxd/instance/instancetype/instance.go
```diff
@@ -1089,6 +1089,16 @@ var InstanceConfigKeysVM = map[string]func(value string) error{
...
 
+	// lxdmeta:generate(entities=instance; group=resource-limits; key=limits.vm.max_bus_ports)
+	// Total number of user configurable PCI devices that can be attached to the VM.
+	// ---
+	//  type: integer
+	//  defaultdesc: `8`
+	//  liveupdate: no
+	//  condition: virtual machine
+	//  shortdesc: Limit of allowed PCI devices
+	"limits.vm.max_bus_ports": validate.Optional(validate.IsUint8),
```
simondeziel: I find the `vm.` bit a bit redundant here.

test/suites/vm.sh
```diff
@@ -136,6 +136,37 @@ test_vm_pcie_bus() {
   lxc exec v1 -- mount -t virtiofs config /mnt
   ! lxc exec v1 -- touch /mnt/foo || false
 
+  sub_test "Check that limits.vm.max_bus_ports config option enforces the number of allowed PCI devices"
+
+  # The default value for "limits.vm.max_bus_ports" is 8 ports.
+  # Fill all the available slots with PCI devices.
+  for i in $(seq 1 5); do
+    lxc config device add v1 "aaa${i}" nic nictype=p2p
+  done
```
simondeziel, line 144: I'm a bit confused as to which ports we are limiting here.

```
$ lxc launch --empty --vm v2
$ sudo grep -F 'bus =' /var/snap/lxd/common/lxd/logs/v2/qemu.conf | sort | uniq -c
      4 bus = "dev-qemu_serial.0"
     13 bus = "pcie.0"
      7 bus = "qemu_pcie0"
      1 bus = "qemu_pcie1"
      2 bus = "qemu_pcie2"
      1 bus = "qemu_pcie3"
```

Don't we already take 4 slots (`qemu_pcie{0..3}`)?

test/suites/vm.sh
```diff
@@ -136,6 +136,37 @@ test_vm_pcie_bus() {
...
+
+  # Stop the VM and attach an additional device.
+  lxc stop -f v1
+  lxc config device add v1 aaa6 nic nictype=p2p
+
+  # Check that the instance start fails because PCI devices limit is exceeded.
+  ! lxc start v1 || false
+
+  # Increase the PCI devices limit.
+  lxc config set v1 limits.vm.max_bus_ports=10
```
simondeziel: I'm wondering if it would make sense to have a `no-hotplug` type of config where the number of devices attached on boot is the upper limit. This way, only the minimum ports are provisioned and it more are needed, it's just a reboot away.

Dunno if there is real demand for it but I'm assuming those ports are not "free" and some might decide they don't want hot-plugging?

doc/api-extensions.md
tomponline: Pci is used on ppcel64 i understand, though had a hard time testing it.

test/suites/vm.sh
nmezhenskyi: > Don't we already take 4 slots (`qemu_pcie{0..3}`)?

Yes, we do. These are considered "required" devices and cannot be configured by the user. So, for the `limits.vm.max_bus_ports` option we only consider the additional user configurable devices.

test/suites/vm.sh
nmezhenskyi: Hm, we could add it as well. But if the user doesn't want hot-plugging, one could just set the `limits.vm.max_bus_ports` value equal to the number of attached devices. The downside is that to add an additional device the limit needs to be increased manually.

lxd/instance/instancetype/instance.go
tomponline: Its vm specific setting so this makes that clear

test/suites/vm.sh
simondeziel: Oh so does that mean that my suggestion of `no-hotplug` magical value is in fact pointless as one could just set `limits.vm.max_bus_ports=0` and be done with it?

test/suites/vm.sh
simondeziel: Is this config just for the number of hot-pluggable ports? If so, a rename would probably make it clearer.

lxd/instance/instancetype/instance.go
simondeziel: There are many other keys with `condition: virtual machine` that don't have the `vm.` part. Similarly, container specific keys like `limits.hugepages.*` or `limits.processes` don't have any `container.` part.

Not a big deal, just feels inconsistent to have it there.

test/suites/vm.sh
nmezhenskyi: > Is this config just for the number of hot-pluggable ports? If so, a rename would probably make it clearer.

Not quite. The config is for all user added PCIe devices (both the ones added before VM start and the ones hot-plugged at runtime). It's just that if you set the limit to equal the number of devices added before VM start, you will not be able to hot-plug additional devices because there will be no free ports for them.

test/suites/vm.sh
nmezhenskyi: > Oh so does that mean that my suggestion of `no-hotplug` magical value is in fact pointless as one could just set `limits.vm.max_bus_ports=0` and be done with it?

Not exactly. One could set for example `limits.vm.max_bus_ports=3` and attach to the instance 3 PCIe devices before starting the VM. In that case, no more PCIe ports would be allocated during VM startup to be used for hot-plugging because the limit is already reached.

The important detail here is that we don't include the "required" devices mentioned above in the calculations. These will always be added regardless.

So, in this particular example the 3 PCIe devices added by the user could be two NICs and one `virtio-blk` disk device.

lxd/patches.go
nmezhenskyi: Updated

lxd/patches.go
nmezhenskyi: Updated

nmezhenskyi:
Hm, I'm getting lint errors from other packages for some reason now

test/suites/vm.sh
tomponline: We can always add additional values or options later if needed, but lets not over complicate this option for now. It sets an upper limit of pci bus ports that can be consumed by custom devices.

lxd/instance/instancetype/instance.go
tomponline: True. The container ones arent really an argument against it as they existed before vm support. But hugepages is a valid point.

Happy to drop "vm" prefix @nmezhenskyi ?

nmezhenskyi: Yes, I agree, let's drop it for consistency with other options

nmezhenskyi: Updated

doc/api-extensions.md
nmezhenskyi: I've updated the wording to describe the option as the maximum allowed number of PCI/PCIe devices

lxd/patches.go
```diff
@@ -2040,4 +2043,85 @@ DELETE FROM storage_pools_config WHERE key = "ceph.osd.force_reuse"
...
+
+	globalConfig := s.GlobalConfig.Dump()
+
+	return s.DB.Cluster.Transaction(s.ShutdownCtx, func(ctx context.Context, tx *db.ClusterTx) error {
+		return tx.InstanceList(ctx, func(inst db.InstanceArgs, project api.Project) error {
+			if inst.Type != instancetype.VM {
+				return nil
+			}
+
+			expandedConfig := instancetype.ExpandInstanceConfig(globalConfig, inst.Config, inst.Profiles)
```
tomponline: @nmezhenskyi apologies for flip flopping on this, but I realised that because we are only considering volatile bus keys, which cannot come from profiles, that we shouldn't need to expand profiles after all and can just use instance local config.

Do you agree?

nmezhenskyi: > Do you agree?

Yes, I agree, this is unnecessary. I somehow forgot that `volatile` keys are only set on the instance itself.

nmezhenskyi: Updated

PR #17615, opened 2026-02-10
Update docs for vm only live migration support

kadinsayani:


doc/howto/instances_backup.md
tomponline, line 1: > doc: Live migration is only supported for containers

For VMs

kadinsayani: Oops, not sure how I typed that :laughing:

PR #17616, opened 2026-02-10
lxd/storage: Add support for Powerflex 5

escabo:
…olumes (as before) or createThinClone for v5 when creating snapshots.

## Checklist

- [x ] I have read the [contributing guidelines](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md) and attest that all commits in this PR are [signed off](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#including-a-signed-off-by-line-in-your-commits), [cryptographically signed](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#commit-signature-verification), and follow this project's [commit structure](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#commit-structure).
- [ x] I have checked and added or updated relevant documentation.

### DIFF
--- a/lxd/storage/drivers/driver_powerflex.go
+++ b/lxd/storage/drivers/driver_powerflex.go
@@ -93,6 +93,11 @@ func (d *powerflex) isRemote() bool {
 	return true
 }
 
+// hasThinCloneSupport returns true if the PowerFlex system version supports thin clones.
+func (d *powerflex) hasThinCloneSupport() bool {
+	return strings.Contains(d.config["powerflex.version"], "R5")
+}
+
 // Info returns info about the driver and its environment.
 func (d *powerflex) Info() Info {
 	return Info{
@@ -150,6 +155,14 @@ func (d *powerflex) FillConfig() error {
 	if d.config["volume.size"] == "" {
 		d.config["volume.size"] = powerFlexDefaultSize
 	}
+
+	// Retrieve and store the PowerFlex system version.
+	systemInfo, err := d.client().getSystemInfo()
+	if err != nil {
+		return err
+	}
+
+	d.config["powerflex.version"] = systemInfo.SystemVersionName
 
 	return nil
 }
@@ -300,6 +313,13 @@ func (d *powerflex) Validate(config map[string]string) error {
 		//  shortdesc: Size/quota of the storage volume
 		//  scope: global
 		"volume.size": validate.Optional(validate.IsMultipleOfUnit("8GiB")),
+		// lxdmeta:generate(entities=storage-powerflex; group=pool-conf; key=powerflex.version)
+		// This field is automatically populated by querying the PowerFlex system.
+		// ---
+		//  type: string
+		//  shortdesc: Software version of the PowerFlex array.
+		//  scope: global
+		"powerflex.version": validate.IsAny,
 	}
 
 	err := d.validatePool(config, rules, d.commonVolumeRules())
--- a/lxd/storage/drivers/driver_powerflex_utils.go
+++ b/lxd/storage/drivers/driver_powerflex_utils.go
@@ -176,6 +176,13 @@ type powerFlexVolume struct {
 	} `json:"mappedSdcInfo"`
 }
 
+// powerFlexSystemInfo stores system information from PowerFlex.
+type powerFlexSystemInfo struct {
+	SystemVersionName string `json:"systemVersionName"`
+	SWID              string `json:"swid"`
+	DaysInstalled     int    `json:"daysInstalled"`
+}
+
 // powerFlexClient holds the PowerFlex HTTP client and an access token factory.
 type powerFlexClient struct {
 	driver *powerflex
@@ -360,6 +367,17 @@ func (p *powerFlexClient) getStoragePoolVolumes(poolID string) ([]powerFlexVolum
 	return actualResponse, nil
 }
 
+// getSystemInfo returns system information from the PowerFlex system.
+func (p *powerFlexClient) getSystemInfo() (*powerFlexSystemInfo, error) {
+	var actualResponse []powerFlexSystemInfo
+	err := p.requestAuthenticated(http.MethodGet, "/api/types/System/instances", nil, &actualResponse)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get system instances: %w", err)
+	}
+
+	return &actualResponse[0], nil
+}
+
 // getProtectionDomainID returns the ID of the protection domain behind domainName.
 func (p *powerFlexClient) getProtectionDomainID(domainName string) (string, error) {
 	body := map[string]any{
@@ -535,6 +553,8 @@ func (p *powerFlexClient) renameVolume(volumeID string, newName string) error {
 // The accessMode can be either ReadWrite or ReadOnly.
 // The returned string represents the ID of the snapshot.
 func (p *powerFlexClient) createVolumeSnapshot(systemID string, volumeID string, snapshotName string, accessMode powerFlexSnapshotMode) (string, error) {
+	method := "createThinClone" // method for when PowerFlex supports ThinClones (R5 and later)
+
 	body := map[string]any{
 		"snapshotDefs": []map[string]string{
 			{
@@ -542,14 +562,20 @@ func (p *powerFlexClient) createVolumeSnapshot(systemID string, volumeID string,
 				"snapshotName": snapshotName,
 			},
 		},
-		"accessModeLimit": accessMode,
 	}
 
 	var actualResponse struct {
 		VolumeIDs []string `json:"volumeIdList"`
 	}
 
-	err := p.requestAuthenticated(http.MethodPost, "/api/instances/System::"+systemID+"/action/snapshotVolumes", body, &actualResponse)
+	// Need to specify the accessModeLimit setting for snapshot creation when PowerFlex does not
+	// support ThinClones. Also revert to classic snapshot method.
+	if !p.driver.hasThinCloneSupport() {
+		body["accessModeLimit"] = accessMode
+		method = "snapshotVolumes"
+	}
+
+	err := p.requestAuthenticated(http.MethodPost, "/api/instances/System::"+systemID+"/action/"+method, body, &actualResponse)
 	if err != nil {
 		powerFlexError, ok := err.(*powerFlexError)
 		if ok {
@@ -1338,7 +1364,8 @@ func (d *powerflex) getVolumeName(vol Volume) (string, error) {
 	// This allows differentiating between snapshots which actually belong to its parent volume
 	// and snapshots which were being created as part of copying a volume using powerflex.snapshot_copy.
 	// The latter are snapshots of the original volume stand on their own and don't use the snapshot prefix.
-	if vol.IsSnapshot() {
+	// This is required for PowerFlex versions that do not support ThinClones.
+	if !d.hasThinCloneSupport() && vol.IsSnapshot() {
 		volName = powerFlexSnapshotPrefix + volName
 	}
 
--- a/doc/metadata.txt
+++ b/doc/metadata.txt
@@ -6440,6 +6440,13 @@ Must have at least SystemAdmin role to give LXD full control over managed storag
 
 ```
 
+```{config:option} powerflex.version storage-powerflex-pool-conf
+:scope: "global"
+:shortdesc: "Software version of the PowerFlex array."
+:type: "string"
+This field is automatically populated by querying the PowerFlex system.
+```
+
 ```{config:option} rsync.bwlimit storage-powerflex-pool-conf
 :defaultdesc: "`0` (no limit)"
 :scope: "global"
--- a/lxd/metadata/configuration.json
+++ b/lxd/metadata/configuration.json
@@ -7122,6 +7122,14 @@
 						}
 					},
 					{
+						"powerflex.version": {
+							"longdesc": "This field is automatically populated by querying the PowerFlex system.",
+							"scope": "global",
+							"shortdesc": "Software version of the PowerFlex array.",
+							"type": "string"
+						}
+					},
+					{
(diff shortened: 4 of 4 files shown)


lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -535,21 +553,30 @@ func (p *powerFlexClient) renameVolume(volumeID string, newName string) error {
...
 		},
-		"accessModeLimit": accessMode,
 	}
 
 	var actualResponse struct {
 		VolumeIDs []string `json:"volumeIdList"`
 	}
 
-	err := p.requestAuthenticated(http.MethodPost, "/api/instances/System::"+systemID+"/action/snapshotVolumes", body, &actualResponse)
+	// PowerFlex version before R5 nedd setting accessModeLimit for snapshot creation
```
simondeziel: Typo: `nedd`

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -360,6 +367,17 @@ func (p *powerFlexClient) getStoragePoolVolumes(poolID string) ([]powerFlexVolum
 	return actualResponse, nil
 }
 
+// getSystemInfo returns system information from the PowerFlex system.
+func (p *powerFlexClient) getSystemInfo() (*powerFlexSystemInfo, error) {
+	var actualResponse []powerFlexSystemInfo
+	err := p.requestAuthenticated(http.MethodGet, "/api/types/System/instances", nil, &actualResponse)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get system instances: %w", err)
```
roosterfish, line 375: Do you know why is it an array that is returned by the API? Does it return all servers which are part of the cluster? 
Or does the API just use a list but always returns a single item?

Something that just came to my mind is maybe this list grows when adding peer systems (used for replication). In this case maybe we have to check for a specific (our) system?

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -535,21 +553,30 @@ func (p *powerFlexClient) renameVolume(volumeID string, newName string) error {
...
 	var actualResponse struct {
 		VolumeIDs []string `json:"volumeIdList"`
 	}
 
-	err := p.requestAuthenticated(http.MethodPost, "/api/instances/System::"+systemID+"/action/snapshotVolumes", body, &actualResponse)
+	// PowerFlex version before R5 nedd setting accessModeLimit for snapshot creation
+	// also revert to old snapshot method.
+	if !strings.Contains(version, "Version: R5") {
+		body["accessModeLimit"] = accessMode
+		method = "snapshotVolumes"
```
roosterfish, line 575: I see so if it's a PowerFlex 5, instead of snapshots the driver does thin clones when snapshotting a volume.

When we first talked about this, I thought it will be something like Ceph RBD where we continue to do snapshots, but in order to be able to mount a snapshot, we would create a thin clone from it which can then later be mounted.
Has this changed or is snapshot/thin clone the same in PF 5?

Another thing that comes to my mind is recovery. The drivers `ListVolumes` func will iterate over all vols and tries to identify snapshots. 
When creating thin clones, do they still appear as snapshots of the parent volume? You can verify this by running `lxd recover` to restore an instance with one or more snapshots. The recovery will fail if the storage returns a different amount of snapshots (or none) compared to what is in the instance's `backup.yaml` file.

lxd/storage/drivers/driver_powerflex.go
```diff
@@ -300,6 +308,13 @@ func (d *powerflex) Validate(config map[string]string) error {
 		//  shortdesc: Size/quota of the storage volume
 		//  scope: global
 		"volume.size": validate.Optional(validate.IsMultipleOfUnit("8GiB")),
+		// lxdmeta:generate(entities=storage-powerflex; group=pool-conf; key=powerflex.system.version)
+		// This field is automatically populated by querying the PowerFlex system.
+		// ---
+		//  type: string
+		//  shortdesc: Software version of the PowerFlex array.
+		//  scope: global
+		"powerflex.system.version": validate.IsAny,
```
roosterfish: I wonder if we should name it just `powerflex.version` without adding this extra `.system.` namespace.

roosterfish: Also same as `powerflex.mode` I think we should not allow this to be ever changed after creating the pool.
Maybe we can also reject any attempt to set the version during pool creation as we always want to fetch it from the API I suppose.

roosterfish (changes requested):
Hi, thanks for this! Just a few thoughts, mainly around always doing thin clones instead of snaps.

Regarding the `powerflex.snapshot_copy` when using PF 5, I think we should disallow the setting as it wouldn't anymore have any effect if we just do thin clones.

Maybe not for this PR, but things like volume copy could now also be improved to use thin clones on PF directly instead of the slow volume copy performed by LXD. This way we can remove `powerflex.snapshot_copy` entirely for PF 5.

lxd/storage/drivers/driver_powerflex_utils.go
tomponline: > You can verify this by running lxd recover to restore an instance with one or more snapshots. 

Do the tests cover this scenario @roosterfish ?

lxd/storage/drivers/driver_powerflex.go
tomponline: > I wonder if we should name it just powerflex.version without adding this extra .system. namespace.

I agree, unless there are other possible version numbers that are relevant?

lxd/storage/drivers/driver_powerflex.go
```diff
@@ -151,6 +151,14 @@ func (d *powerflex) FillConfig() error {
 		d.config["volume.size"] = powerFlexDefaultSize
 	}
 
+	// Retrieve and store the PowerFlex system version.
+	systemInfo, err := d.client().getSystemInfo()
+	if err != nil {
+		return err
+	}
+
+	d.config["powerflex.system.version"] = systemInfo.SystemVersionName
```
tomponline: as we are only interested in the version, can we parse and extract the version number from `SystemVersionName` here rather than storing the unparsed version name here and having the driver parse it on every use.

Because looking at the version parsing code below, there seems to be some duplication in it:

```
Version: R5
```

Can we just store "R5" in the config?

lxd/storage/drivers/driver_powerflex_utils.go
roosterfish: Regular system tests yes, but there isn't yet an extra test specifically for PowerFlex, Pure, ... regarding recovery in `lxd-ci`.

lxd/storage/drivers/driver_powerflex.go
tomponline: or even just store "5"?

then we could treat it as an integer in the code do `<5` type logic on it.

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -535,21 +553,30 @@ func (p *powerFlexClient) renameVolume(volumeID string, newName string) error {
...
 	}
 
 	var actualResponse struct {
 		VolumeIDs []string `json:"volumeIdList"`
 	}
 
-	err := p.requestAuthenticated(http.MethodPost, "/api/instances/System::"+systemID+"/action/snapshotVolumes", body, &actualResponse)
+	// PowerFlex version before R5 nedd setting accessModeLimit for snapshot creation
+	// also revert to old snapshot method.
+	if !strings.Contains(version, "Version: R5") {
```
tomponline: See https://github.com/canonical/lxd/pull/17616/changes/dcfd9ac8ac1da9b2763e2ad7c5426753d471d6db#r2792036923 if `version` was just "5" we could parse it to an int and then do `<5` logic on it. This would future proof it against newer versions.

lxd/storage/drivers/driver_powerflex_utils.go
roosterfish: I have logged https://github.com/canonical/lxd/issues/17620 and assigned it to me.

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -1338,7 +1366,8 @@ func (d *powerflex) getVolumeName(vol Volume) (string, error) {
 	// This allows differentiating between snapshots which actually belong to its parent volume
 	// and snapshots which were being created as part of copying a volume using powerflex.snapshot_copy.
 	// The latter are snapshots of the original volume stand on their own and don't use the snapshot prefix.
-	if vol.IsSnapshot() {
+	// This is only required for PowerFlex versions before R5.
+	if !strings.Contains(version, "Version: R5") && vol.IsSnapshot() {
```
tomponline: As there are 2 places (for now, but sounds like this could grow in the future) that check the version is 5, I think it would be worth de-duping the logic into a separate function or two.

E.g.

```go
func (d *powerflex) load() error {
   var err error
	d.version, err = version.Parse(d.config["powerflex.system.version"]) // e.g. `5`
	if err != nil {
		return err
	}
}

func (d *powerflex) hasThinSnapshots() bool {
	ver5, err := version.Parse("5")
	if err != nil {
	    // panic
	}

    if d.version.Compare(ver5) >= 0 {
         return true
    }

     return false
}
```

Then we can parse the config key once at pool load time, and then use it for logical feature detection, e.g.

```go
if !d.hasThinSnapshots() && vol.IsSnapshot() {
}
```

tomponline:
Also you need to run `make update-metadata` to update the configuration metadata file and commit it.

lxd/storage/drivers/driver_powerflex_utils.go
escabo: > I see so if it's a PowerFlex 5, instead of snapshots the driver does thin clones when snapshotting a volume.
> 
> When we first talked about this, I thought it will be something like Ceph RBD where we continue to do snapshots, but in order to be able to mount a snapshot, we would create a thin clone from it which can then later be mounted. Has this changed or is snapshot/thin clone the same in PF 5?
> 
> Another thing that comes to my mind is recovery. The drivers `ListVolumes` func will iterate over all vols and tries to identify snapshots. When creating thin clones, do they still appear as snapshots of the parent volume? You can verify this by running `lxd recover` to restore an instance with one or more snapshots. The recovery will fail if the storage returns a different amount of snapshots (or none) compared to what is in the instance's `backup.yaml` file.

This is a "quick fix" to get 5 up and running. PF now has "real" snapshot which are more efficient but would require us to do like Ceph RBD i.e. a more involved change.

lxd/storage/drivers/driver_powerflex_utils.go
escabo: > Do you know why is it an array that is returned by the API? Does it return all servers which are part of the cluster? Or does the API just use a list but always returns a single item?
> 
> Something that just came to my mind is maybe this list grows when adding peer systems (used for replication). In this case maybe we have to check for a specific (our) system?

I don't, this is a good point we could test it.

lxd/storage/drivers/driver_powerflex_utils.go
roosterfish: I see thanks. Then the tricky part is how to progress forward when we land support for the new snapshots in the future. We would probably need a patch that moves the vols from thin clones back to snaps? Not sure if this is doable. 

Seems to be one of those things if we don't add it right from the beginning, we likely will be blocked in the future.
Or the driver has to support both methods at the same time.

lxd/storage/drivers/driver_powerflex.go
```diff
@@ -93,6 +93,15 @@ func (d *powerflex) isRemote() bool {
...
 }
 
+// hasThinCloneSupport returns true if the PowerFlex system version supports thin clones.
+func (d *powerflex) hasThinCloneSupport() bool {
+	version := d.config["powerflex.version"]
+	if strings.Contains(version, "R5") {
+		return true
+	}
+	return false
+}
```
simondeziel, line 99:
```suggestion
func (d *powerflex) hasThinCloneSupport() bool {
	return strings.Contains(d.config["powerflex.version"], "R5")
}
```

lxd/storage/drivers/driver_powerflex_utils.go
```diff
@@ -1338,7 +1364,8 @@ func (d *powerflex) getVolumeName(vol Volume) (string, error) {
 	// This allows differentiating between snapshots which actually belong to its parent volume
 	// and snapshots which were being created as part of copying a volume using powerflex.snapshot_copy.
 	// The latter are snapshots of the original volume stand on their own and don't use the snapshot prefix.
-	if vol.IsSnapshot() {
+	// This is required for PowerFlex versions that do not support ThinClones
```
simondeziel:
```suggestion
	// This is required for PowerFlex versions that do not support ThinClones.
```

simondeziel (commented):
Quick drive-by review.

lxd/storage/drivers/driver_powerflex.go
```diff
@@ -300,6 +313,13 @@ func (d *powerflex) Validate(config map[string]string) error {
 		//  shortdesc: Size/quota of the storage volume
 		//  scope: global
 		"volume.size": validate.Optional(validate.IsMultipleOfUnit("8GiB")),
+		// lxdmeta:generate(entities=storage-powerflex; group=pool-conf; key=powerflex.version)
+		// This field is automatically populated by querying the PowerFlex system.
+		// ---
+		//  type: string
+		//  shortdesc: Software version of the PowerFlex array.
+		//  scope: global
+		"powerflex.version": validate.IsAny,
```
simondeziel, line 322: @escabo I think this change is what requires you to run `make update-metadata` (as hinted by Tom) and commit the result. Calling that target should have it prompt you to commit then resulting changes automatically.

escabo:
sorry, I forgot.

PR #17624, opened 2026-02-11
Simplify database function generation

markylaing:
Since cluster links is coming soon, I wanted to put forward this alternative database generator: `simplegen`. Hopefully starting to use this instead will reduce the total amount of work needed when we make the transition (which is all ultimately needed for #16613).

Reasons for the change:
1. None of the developers who created lxd-generate now work on the project.
2. lxd-generate makes assumptions about the structure of the data following LXD conventions.
3. The fundamental access model for lxd-generate assumes that you want to manipulate database objects by the API name and not the primary key (this is counter to what is needed for #16613).

The above points all create friction when adding a new table. We end up writing a lot of SQL manually.

`simplegen` is a halfway house where shorthand comments can be used to describe the relationship that a struct has with the database. This allows `simplegen` to generate methods for that type that implement interfaces defined in the `query` package. These methods can be implemented manually if, for example, it is useful to query for a type but not create/update/delete it.

## Checklist

### DIFF
--- a/lxd/placement/filtering_test.go
+++ b/lxd/placement/filtering_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/canonical/lxd/lxd/db"
 	"github.com/canonical/lxd/lxd/db/cluster"
+	"github.com/canonical/lxd/lxd/db/query"
 	"github.com/canonical/lxd/lxd/instance/instancetype"
 	"github.com/canonical/lxd/shared/api"
 )
@@ -88,8 +89,8 @@ func (s *filteringSuite) TestFilter() {
 			caseSetup: func() {
 				_ = testCluster.Transaction(context.Background(), func(ctx context.Context, tx *db.ClusterTx) error {
 					// Create placement group with spread/strict policy.
-					pgID, err := cluster.CreatePlacementGroup(ctx, tx.Tx(), cluster.PlacementGroup{
-						Project:     "default",
+					pgID, err := query.Create(ctx, tx.Tx(), cluster.PlacementGroupsRow{
+						ProjectID:   1,
 						Name:        "pg-spread-strict",
 						Description: "Spread strict placement group",
 					})
@@ -108,9 +109,11 @@ func (s *filteringSuite) TestFilter() {
 				candidates: candidates,
 				project:    "default",
 				placementGroup: cluster.PlacementGroup{
-					Project:     "default",
-					Name:        "pg-spread-strict",
-					Description: "Spread strict placement group",
+					ProjectName: "default",
+					Row: cluster.PlacementGroupsRow{
+						Name:        "pg-spread-strict",
+						Description: "Spread strict placement group",
+					},
 				},
 			},
 			want:    candidates, // All candidates available for first instance.
@@ -141,9 +144,11 @@ func (s *filteringSuite) TestFilter() {
 				candidates: candidates,
 				project:    "default",
 				placementGroup: cluster.PlacementGroup{
-					Project:     "default",
-					Name:        "pg-spread-strict",
-					Description: "Spread strict placement group",
+					ProjectName: "default",
+					Row: cluster.PlacementGroupsRow{
+						Name:        "pg-spread-strict",
+						Description: "Spread strict placement group",
+					},
 				},
 			},
 			want:    candidatesWithout("member01"), // member01 has instance, exclude it.
@@ -176,9 +181,11 @@ func (s *filteringSuite) TestFilter() {
 				candidates: candidates,
 				project:    "default",
 				placementGroup: cluster.PlacementGroup{
-					Project:     "default",
-					Name:        "pg-spread-strict",
-					Description: "Spread strict placement group",
+					ProjectName: "default",
+					Row: cluster.PlacementGroupsRow{
+						Name:        "pg-spread-strict",
+						Description: "Spread strict placement group",
+					},
 				},
 			},
 			want:    nil,
@@ -201,8 +208,8 @@ func (s *filteringSuite) TestFilter() {
 			name: "spread/permissive: initial placement",
 			caseSetup: func() {
 				_ = testCluster.Transaction(context.Background(), func(ctx context.Context, tx *db.ClusterTx) error {
-					pgID, err := cluster.CreatePlacementGroup(ctx, tx.Tx(), cluster.PlacementGroup{
-						Project:     "default",
+					pgID, err := query.Create(ctx, tx.Tx(), cluster.PlacementGroupsRow{
+						ProjectID:   1,
 						Name:        "pg-spread-permissive",
 						Description: "Spread permissive placement group",
 					})
@@ -221,9 +228,11 @@ func (s *filteringSuite) TestFilter() {
 				candidates: candidates,
 				project:    "default",
 				placementGroup: cluster.PlacementGroup{
-					Project:     "default",
-					Name:        "pg-spread-permissive",
-					Description: "Spread permissive placement group",
+					ProjectName: "default",
+					Row: cluster.PlacementGroupsRow{
+						Name:        "pg-spread-permissive",
+						Description: "Spread permissive placement group",
+					},
 				},
 			},
 			want:    candidates, // All candidates have 0 instances.
@@ -267,9 +276,11 @@ func (s *filteringSuite) TestFilter() {
 				candidates: candidates,
 				project:    "default",
 				placementGroup: cluster.PlacementGroup{
-					Project:     "default",
-					Name:        "pg-spread-permissive",
-					Description: "Spread permissive placement group",
+					ProjectName: "default",
+					Row: cluster.PlacementGroupsRow{
+						Name:        "pg-spread-permissive",
+						Description: "Spread permissive placement group",
+					},
 				},
 			},
 			want:    candidatesOnly("member04", "member05"), // Only nodes with 0 instances (min).
@@ -291,8 +302,8 @@ func (s *filteringSuite) TestFilter() {
 			name: "compact/strict: initial placement",
 			caseSetup: func() {
 				_ = testCluster.Transaction(context.Background(), func(ctx context.Context, tx *db.ClusterTx) error {
-					pgID, err := cluster.CreatePlacementGroup(ctx, tx.Tx(), cluster.PlacementGroup{
-						Project:     "default",
+					pgID, err := query.Create(ctx, tx.Tx(), cluster.PlacementGroupsRow{
+						ProjectID:   1,
 						Name:        "pg-compact-strict",
 						Description: "Compact strict placement group",
 					})
@@ -311,9 +322,11 @@ func (s *filteringSuite) TestFilter() {
 				candidates: candidates,
 				project:    "default",
 				placementGroup: cluster.PlacementGroup{
-					Project:     "default",
-					Name:        "pg-compact-strict",
-					Description: "Compact strict placement group",
+					ProjectName: "default",
+					Row: cluster.PlacementGroupsRow{
+						Name:        "pg-compact-strict",
+						Description: "Compact strict placement group",
+					},
 				},
 			},
 			want:    candidates, // No instances yet, all candidates valid.
@@ -344,9 +357,11 @@ func (s *filteringSuite) TestFilter() {
 				candidates: candidates,
 				project:    "default",
 				placementGroup: cluster.PlacementGroup{
-					Project:     "default",
-					Name:        "pg-compact-strict",
-					Description: "Compact strict placement group",
+					ProjectName: "default",
+					Row: cluster.PlacementGroupsRow{
+						Name:        "pg-compact-strict",
+						Description: "Compact strict placement group",
+					},
 				},
 			},
 			want:    candidatesOnly("member03"), // Only member03.
@@ -358,9 +373,11 @@ func (s *filteringSuite) TestFilter() {
 				candidates: candidatesWithout("member03"), // member03 not in candidates.
 				project:    "default",
 				placementGroup: cluster.PlacementGroup{
-					Project:     "default",
-					Name:        "pg-compact-strict",
-					Description: "Compact strict placement group",
+					ProjectName: "default",
+					Row: cluster.PlacementGroupsRow{
+						Name:        "pg-compact-strict",
+						Description: "Compact strict placement group",
+					},
 				},
 			},
 			want:    nil,
@@ -413,9 +430,11 @@ func (s *filteringSuite) TestFilter() {
 				candidates: candidates,
 				project:    "default",
 				placementGroup: cluster.PlacementGroup{
-					Project:     "default",
-					Name:        "pg-compact-strict",
-					Description: "Compact strict placement group",
+					ProjectName: "default",
+					Row: cluster.PlacementGroupsRow{
+						Name:        "pg-compact-strict",
+						Description: "Compact strict placement group",
+					},
 				},
 			},
 			want:    candidatesOnly("member01"), // member01 has 3 instances vs member02 with 1.
@@ -437,8 +456,8 @@ func (s *filteringSuite) TestFilter() {
 			name: "compact/permissive: initial placement",
 			caseSetup: func() {
 				_ = testCluster.Transaction(context.Background(), func(ctx context.Context, tx *db.ClusterTx) error {
-					pgID, err := cluster.CreatePlacementGroup(ctx, tx.Tx(), cluster.PlacementGroup{
-						Project:     "default",
+					pgID, err := query.Create(ctx, tx.Tx(), cluster.PlacementGroupsRow{
+						ProjectID:   1,
 						Name:        "pg-compact-permissive",
 						Description: "Compact permissive placement group",
 					})
@@ -457,9 +476,11 @@ func (s *filteringSuite) TestFilter() {
 				candidates: candidates,
 				project:    "default",
 				placementGroup: cluster.PlacementGroup{
-					Project:     "default",
-					Name:        "pg-compact-permissive",
-					Description: "Compact permissive placement group",
+					ProjectName: "default",
+					Row: cluster.PlacementGroupsRow{
+						Name:        "pg-compact-permissive",
+						Description: "Compact permissive placement group",
+					},
 				},
 			},
 			want:    candidates, // No instances yet, all candidates valid.
@@ -490,9 +511,11 @@ func (s *filteringSuite) TestFilter() {
 				candidates: candidates,
 				project:    "default",
 				placementGroup: cluster.PlacementGroup{
-					Project:     "default",
-					Name:        "pg-compact-permissive",
-					Description: "Compact permissive placement group",
+					ProjectName: "default",
+					Row: cluster.PlacementGroupsRow{
+						Name:        "pg-compact-permissive",
+						Description: "Compact permissive placement group",
+					},
 				},
 			},
 			want:    candidatesOnly("member02"), // Prefer member02.
@@ -504,9 +527,11 @@ func (s *filteringSuite) TestFilter() {
 				candidates: candidatesWithout("member02"), // member02 not in candidates.
 				project:    "default",
 				placementGroup: cluster.PlacementGroup{
-					Project:     "default",
-					Name:        "pg-compact-permissive",
-					Description: "Compact permissive placement group",
+					ProjectName: "default",
+					Row: cluster.PlacementGroupsRow{
+						Name:        "pg-compact-permissive",
+						Description: "Compact permissive placement group",
+					},
 				},
 			},
 			want:    candidatesWithout("member02"), // All other candidates.
@@ -553,9 +578,11 @@ func (s *filteringSuite) TestFilter() {
 				candidates: candidates,
 				project:    "default",
 				placementGroup: cluster.PlacementGroup{
-					Project:     "default",
-					Name:        "pg-compact-permissive",
-					Description: "Compact permissive placement group",
+					ProjectName: "default",
+					Row: cluster.PlacementGroupsRow{
+						Name:        "pg-compact-permissive",
+						Description: "Compact permissive placement group",
+					},
 				},
 			},
 			want:    candidatesOnly("member02"), // member02 has 3 instances vs member04 with 1.
@@ -583,7 +610,7 @@ func (s *filteringSuite) TestFilter() {
 
 		_ = testCluster.Transaction(context.Background(), func(ctx context.Context, tx *db.ClusterTx) error {
 			// Fetch the placement group to get the full object with ID.
-			placementGroup, err := pgCache.Get(ctx, tx, tt.args.placementGroup.Name, tt.args.placementGroup.Project)
+			placementGroup, err := pgCache.Get(ctx, tx, tt.args.placementGroup.Row.Name, tt.args.placementGroup.ProjectName)
 			if err != nil {
 				return err
 			}


lxd/auth_groups.go
```diff
@@ -189,7 +190,7 @@ func getAuthGroups(d *Daemon, r *http.Request) response.Response {
 	groupsIdentityProviderGroups := make(map[int64][]dbCluster.IdentityProviderGroup)
 	entityURLs := make(map[entity.Type]map[int]*api.URL)
 	err = d.db.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
-		allGroups, err := dbCluster.GetAuthGroups(ctx, tx.Tx())
```
kadinsayani, line 193: `GetAuthGroups()`'s sql statement currently includes `ORDER_BY auth_groups.name`, should the new generator also include default ordering?

lxd/db/cluster/generated.go
```diff
@@ -0,0 +1,55 @@
...
+package cluster
+
+// Generated by simplegen - DO NOT EDIT
+
+// TableName returns the table name for [AuthGroup] entities.
+func (a AuthGroup) TableName() string {
+	return "auth_groups"
+}
+
+func (a AuthGroup) ScanColumns() []string {
```
kadinsayani: Missing godoc

markylaing:
> As the intention is to partly replace `lxd-generate`, could you also wire it into the `make update-schema` so that it contributes to code coverage.

I'm not sure what you mean. `make update-schema` just calls `go generate ./...`. The directive in `lxd/db/cluster/stmt.go` will build and run the generator. The generated files will be included in coverage because they are used by LXD?

Or do you want the generator code itself to be included in coverage? If so I'm not sure why?

lxd/auth_groups.go
markylaing: The primary goal of this generator is to keep it simple. If ordering is required, then a wrapper function is easy to write:
```
func GetAllAuthGroupsOrderByName(ctx context.Context, tx *sql.Tx) ([]AuthGroup, error) {
    return query.Select[AuthGroup](ctx, tx, "ORDER BY auth_groups.name")
}
```

markylaing:
> I'm excited to use this! Definitely looks much easier to work with. Just wondering, have you thought about extending the new generator to also create associated `_config` tables? Perhaps this can be controlled by a flag:
> 
> ```
> // db:model cluster_links config=true
> ```

As above I think this generator should remain as simple as possible. It's for CRUD operations on a single table. Also since config is so ubiquitous and most tables are effectively identical, I wouldn't want to generate separate types/functions for it and instead would centralise it.

markylaing:
> > I'm excited to use this! Definitely looks much easier to work with. Just wondering, have you thought about extending the new generator to also create associated `_config` tables? Perhaps this can be controlled by a flag:
> > ```
> > // db:model cluster_links config=true
> > ```
> 
> As above I think this generator should remain as simple as possible. It's for CRUD operations on a single table. Also since config is so ubiquitous and most tables are effectively identical, I wouldn't want to generate separate types/functions for it and instead would centralise it.

I'll put up a PR for config and for another type after this to show you what I mean

simondeziel:
> Or do you want the generator code itself to be included in coverage? If so I'm not sure why?

Yep, exactly. As discussed in our 1:1, that's to avoid adding Go code in our repo that would be 0% tested from the TiCS point of view. Silly, I know but I'd rather not lose points for such stupid thing.

lxd/db/cluster/generated.go
markylaing: Fixed

markylaing:
> > Or do you want the generator code itself to be included in coverage? If so I'm not sure why?
> 
> Yep, exactly. As discussed in our 1:1, that's to avoid adding Go code in our repo that would be 0% tested from the TiCS point of view. Silly, I know but I'd rather not lose points for such stupid thing.

Fixed thanks

lxd/db/cluster/auth_groups.go
```diff
@@ -4,49 +4,39 @@ import (
...
+
+// AuthGroupExists returns true if an AuthGroup exists and false otherwise.
+func AuthGroupExists(ctx context.Context, tx *sql.Tx, groupName string) (bool, error) {
+	_, err := GetAuthGroup(ctx, tx, groupName)
+	if err != nil {
+		if api.StatusErrorCheck(err, http.StatusNotFound) {
+			return false, nil
+		}
+
+		return false, fmt.Errorf("Failed to query for auth groups by name: %w", err)
```
kadinsayani:
```suggestion
		return false, fmt.Errorf("Failed executing query for auth groups by name: %w", err)
```

lxd/db/query/generic.go
```diff
@@ -0,0 +1,237 @@
...
+// Delete deletes all rows from the table described by T that match the given clause.
+func Delete[T TableNamer](ctx context.Context, tx *sql.Tx, clause string, args ...any) (int64, error) {
+	// This is probably way OTT but we can't use t = *new(T) is in the above functions because the T might be a pointer.
+	// This line indirects a new TableNamer and calls TableName on the indirection.
+	// Since we know that T implements TableNamer, this is safe to do.
+	tableName := reflect.Indirect(reflect.ValueOf(*new(T))).MethodByName("TableName").Call(nil)[0].String()
+	stmt := "DELETE FROM " + tableName + " " + clause
+	result, err := tx.ExecContext(ctx, stmt, args...)
+	if err != nil {
+		return -1, fmt.Errorf("Failed to delete from table %q: %w", tableName, err)
```
kadinsayani:
```suggestion
		return -1, fmt.Errorf("Failed deleting from table %q: %w", tableName, err)
```

lxd/db/query/generic.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	n, err := result.RowsAffected()
+	if err != nil {
+		return -1, fmt.Errorf("Failed to get number of deleted rows: %w", err)
```
kadinsayani:
```suggestion
		return -1, fmt.Errorf("Failed getting number of deleted rows: %w", err)
```

lxd/db/simplegen/main.go
```diff
@@ -0,0 +1,215 @@
...
+
+	if len(os.Args) > 3 {
+		return errors.New("Unknown additional arguments")
+	}
+
+	// Directory to inspect.
+	packageDirectory := os.Args[1]
+	info, err := os.Stat(packageDirectory)
+	if err != nil {
+		return fmt.Errorf("Could not stat package directory: %w", err)
```
kadinsayani:
```suggestion
		return fmt.Errorf("Cannot stat package directory: %w", err)
```

lxd/db/simplegen/main.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	return nil
+}
+
+// getPackageSpecs parses each .go file in the package and creates a spec for each tagged struct.
+func getPackageSpecs(packageDirectory string) (packageName string, specs []Spec, err error) {
+	paths, err := filepath.Glob(filepath.Join(packageDirectory, "*.go"))
+	if err != nil {
+		return "", nil, fmt.Errorf("Failed to list package files: %w", err)
```
kadinsayani:
```suggestion
		return "", nil, fmt.Errorf("Failed listing package files: %w", err)
```

lxd/db/simplegen/main.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	var allSpecs []Spec
+	var pkgName string
+	for _, path := range paths {
+		// Parse file AST.
+		file, err := parser.ParseFile(token.NewFileSet(), path, nil, parser.ParseComments)
+		if err != nil {
+			return "", nil, fmt.Errorf("Failed to parse Go source file %q: %w", path, err)
```
kadinsayani:
```suggestion
			return "", nil, fmt.Errorf("Failed parsing Go source file %q: %w", path, err)
```

lxd/db/simplegen/main.go
```diff
@@ -0,0 +1,215 @@
...
+			// Hardcode primary key as "id" column, as that is always true in LXD.
+			if newFieldSpec.ColumnName == tableName+".id" {
+				newSpec.PrimaryKey = newFieldSpec
+			}
+
+			newSpec.Fields = append(newSpec.Fields, newFieldSpec)
+		}
+
+		if newSpec.PrimaryKey.FieldName == "" {
+			return nil, fmt.Errorf("Did not find a primary key for struct %q", newSpec.StructName)
```
kadinsayani:
```suggestion
			return nil, fmt.Errorf("Cannot find a primary key for struct %q", newSpec.StructName)
```

lxd/db/cluster/auth_groups.go
markylaing: Fixed.

lxd/db/query/generic.go
markylaing: Fixed.

lxd/db/query/generic.go
markylaing: Fixed.

lxd/db/simplegen/main.go
markylaing: Fixed.

lxd/db/simplegen/main.go
markylaing: Fixed.

lxd/db/simplegen/main.go
markylaing: Fixed.

lxd/db/simplegen/main.go
markylaing: Fixed.

lxd/db/cluster/generated.go
```diff
@@ -0,0 +1,56 @@
+//go:build linux && cgo && !agent
```
roosterfish, line 1: I wonder if the file name should contain the table name? E.g. `auth_groups_generated.go`?

lxd/db/query/generic.go
```diff
@@ -0,0 +1,242 @@
...
+// Updatable defines a type that knows how to update itself.
+type Updatable interface {
+	Referencable
+	Creatable
+}
+
+// BaseQuerier specifies how to query for all the basic data for a given type.
+// The order of the columns should match the order from [ScanArger].
+type BaseQuerier interface {
+	BaseQuery() string
```
roosterfish: When you run `Select` or `SelectFunc` it's always invoking `BaseQuery` which effectively results in a `SELECT * FROM` to get all columns. 

If I just want to do a `SELECT single_colmn FROM`, do I need to create an additional struct implementing `BaseQuerier` which only has the `single_column` field? Let's talk about it later in our meet.

lxd/db/cluster/generated.go
markylaing: As it is right now, all generated code is written to a single file. I personally think this is clearer because we will otherwise have quite a lot of short generated files that litter the directory.

lxd/db/query/generic.go
markylaing: As discussed, I think in general we should always query for all data in a row but if querying a single column is required then you can either write a custom query and func (using `query.Scan`) or you can create a type that implements `BaseQuerier`.

lxd/db/cluster/generated.go
roosterfish: Ah ok, I missed the bit that everything goes into one file.

markylaing:
@tomponline this is ready when you have a minute

lxd/db/query/generic.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// ScanArger specifies how a type should advertise arguments for scanning rows.
+// The order of these arguments should match the order of columns from [BaseQuerier].
+type ScanArger interface {
+	ScanArgs() []any
+}
+
+// Create creates a creatable. All columns are set except for the primary key.
```
tomponline: /creatable/Creatable/

tomponline: > All columns are set except for the primary key.

Why?

lxd/db/query/generic.go
```diff
@@ -0,0 +1,244 @@
...
+func Create(ctx context.Context, tx *sql.Tx, c Creatable) (int64, error) {
+	tableName := c.TableName()
+
+	res, err := tx.ExecContext(ctx, c.CreateStmt(), c.CreateValues()...)
+	if err != nil {
+		if IsConflictErr(err) {
+			return -1, api.StatusErrorf(http.StatusConflict, "This entry in table %q already exists: %w", tableName, err)
```
tomponline: Lets drop the "This", so:

> Entry in table %q already exists

lxd/db/query/generic.go
```diff
@@ -0,0 +1,244 @@
...
+// Update updates an [Updatable] type by its primary key. It sets all fields except for the primary key.
+// The expected usage is to get an [Updatable] entry from the database (e.g. via [SelectOne]), change fields on it
+// directly, and then call Update.
+func Update(ctx context.Context, tx *sql.Tx, u Updatable) error {
+	tableName := u.TableName()
+
+	res, err := tx.ExecContext(ctx, u.UpdateStmt(), u.UpdateValues()...)
+	if err != nil {
+		if IsConflictErr(err) {
+			return api.StatusErrorf(http.StatusConflict, "This entry in table %q already exists: %w", tableName, err)
```
tomponline: Lets drop the "This"

lxd/db/query/generic.go
```diff
@@ -0,0 +1,244 @@
...
+		return fmt.Errorf("Failed updating %q entry: %w", tableName, err)
+	}
+
+	n, err := res.RowsAffected()
+	if err != nil {
+		return fmt.Errorf("Failed verifying update to table %q entry: %w", tableName, err)
+	}
+
+	if n != 1 {
+		return fmt.Errorf("Update by primary key where \"%s.%s = %v\" affected %d rows", tableName, u.PKColumn(), u.PKValue(), n)
```
tomponline: What if no changes genuinely  have taken place?

lxd/db/query/generic.go
```diff
@@ -0,0 +1,244 @@
...
+		return nil, err
+	}
+
+	return ts, nil
+}
+
+// SelectFunc selects data as defined by the base query of the given type.
+// A function is called for every row scanned.
+// Supplementary SQL can be passed in to specify additional joins, clauses, or aggregations.
+// The type constraint specifies that the non-pointer type must implement BaseQuerier, but a pointer to the same type
```
tomponline, line 192: Can you expand this comment to explain why more?

lxd/db/query/generic.go
```diff
@@ -0,0 +1,244 @@
...
+// must implement [ScanArger].
+func SelectFunc[T BaseQuerier, PT interface {
+	ScanArger
+	*T
+}](ctx context.Context, tx *sql.Tx, clause string, rowFunc func(T) error, args ...any) error {
+	rows, err := tx.QueryContext(ctx, (*new(T)).BaseQuery()+clause, args...)
```
tomponline: Does it need a space added between `BaseQuery()+clause`?

lxd/db/query/generic.go
tomponline: Please can you use backticks when putting double quotes into a string to avoid the nasty escaping.

lxd/db/query/generic.go
```diff
@@ -0,0 +1,244 @@
...
+		return fmt.Errorf("Failed deleting from table %q: %w", tableName, err)
+	}
+
+	deleted, err := result.RowsAffected()
+	if err != nil {
+		return fmt.Errorf("Failed getting number of deleted rows: %w", err)
+	}
+
+	if deleted != 1 {
+		return fmt.Errorf("Deletion by primary key where \"%s.%s = %v\" affected %d rows", t.TableName(), t.PKColumn(), t.PKValue(), deleted)
```
tomponline: Please can you use backticks when putting double quotes into a string to avoid the nasty escaping.

lxd/db/query/generic.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	return n, nil
+}
+
+// DeleteOne deletes all rows from the table described by the given [TableNamer] that match the given clause.
+// If the number of affected rows is not one, an error is returned (and the transaction should be rolled back).
+// The type constraint enforces that both the type and a pointer to the type implement the [TableNamer] interface.
+// This is the only way to enforce that (*new(T)) is a non-pointer type, and therefore non-nil.
+func DeleteOne[T TableNamer, PT interface {
```
tomponline: I'm confused why we need a `Delete` that fails if more or less than 1 records are deleted and `DeleteOne` that also fails if more or less than 1 records are deleted, but in a different way. Seems there's some redundancy here. Its certainly a confusing API.

lxd/db/cluster/stmt.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/canonical/lxd/shared/entity"
 )
 
+//go:generate simplegen . generated.go
```
tomponline: can we have this at the top of the file please after `//go:build`?

lxd/auth_groups.go
```diff
@@ -730,7 +726,13 @@ func renameAuthGroup(d *Daemon, r *http.Request) response.Response {
 
 	s := d.State()
 	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
-		return dbCluster.RenameAuthGroup(ctx, tx.Tx(), groupName, groupPost.Name)
+		group, err := dbCluster.GetAuthGroup(ctx, tx.Tx(), groupName)
```
tomponline, line 718: This is inefficient in that we have to load the record to update it and then submit the update back, rather than just updating the specific field in the row.

We'll have to see how often this pattern crops up and if it occurs a lot we may want a specific Rename function on the simplegen?

tomponline: or an UpdateField function?

markylaing:
> Overall this looks very clean, I like it!
> 
> There's a few questions/suggestions, but nothing major.
> 
> The main point of confusion for me was the record deletion, the API seems pretty unnatural to me, so if its needed, its not obvious why it needs to be like that.

Thanks for the feedback :)

I'll try to explain my reasoning on the delete APIs and then we can work that reasoning into the function comments to make it clearer.

lxd/auth_groups.go
markylaing: > This is inefficient in that we have to load the record to update it and then submit the update back, rather than just updating the specific field in the row.
> 
> We'll have to see how often this pattern crops up and if it occurs a lot we may want a specific Rename function on the simplegen?

So in this scenario it is less efficient, but this pattern will eventually be to our benefit. The authorizer has already loaded the auth group in order to perform an access check. The grand plan that we'll be always operating on entities via primary key of the resource that was loaded during that access check.

lxd/db/query/generic.go
markylaing: We have:
- `Delete`. This can only delete a `Referencable` via primary key. The intention behind this is to load a resource during authorization and check `can_delete`, then get it from context in the API handler and call `Delete` with it.
- `DeleteMany`. This can delete multiple entities that match a given clause. This is a convenience.
- `DeleteOne`. Deletes a single entity that matches a given clause. This is also a convenience, as I've seen the pattern dotted all over the code base.

They can for sure be named better! Suggestions welcome :)

lxd/db/query/generic.go
markylaing: We perhaps should. All the generated "base queries" have a newline at the end but if we decide to implement the interface separately it could get confusing.

Will change.

lxd/db/query/generic.go
markylaing: Then it's perhaps a good indicator that we shouldn't have called `Update`? Since it is an unnecessary query.

lxd/db/query/generic.go
markylaing: It is assumed that the primary key is auto generated and the generated create statement contains all other fields.

If we have a primary key that needs to be set manually we can't use the generator. 

If we want to use a v7 UUID as the primary key in future, I think we can make it work via something like:
```go

type UUIDv7Auto struct {
    *uuid.UUID
}

func (u UUIDv7Auto) Value() (driver.Value, error) {
    if u.UUID != nil {
        return u.UUID.String()
    }

    v7uuid, err := uuid.NewV7()
    if err != nil {
        return nil, err
    }

    u.UUID = &v7uuid
    return v7uuid.String(), nil
}

/* implement sql.Scanner */
```

lxd/db/query/generic.go
markylaing: Done.

lxd/db/query/generic.go
markylaing: Done.

lxd/db/query/generic.go
markylaing: Done.

lxd/db/query/generic.go
markylaing: Done.

lxd/db/query/generic.go
markylaing: Done.

lxd/db/query/generic.go
markylaing: Done.

lxd/db/cluster/stmt.go
markylaing: Done.

lxd/db/cluster/generated.go
```diff
@@ -0,0 +1,66 @@
...
+
+// CreateStmt returns a query that creates a [AuthGroup] entity.
+func (a AuthGroup) CreateStmt() string {
+	return "INSERT INTO auth_groups (name, description) VALUES (?, ?)"
+}
+
+// UpdateStmt returns a query that updates a [AuthGroup] by primary key.
+func (a AuthGroup) UpdateStmt() string {
+	return "UPDATE auth_groups SET name = ?, description = ? WHERE auth_groups.id = ?"
+}
```
markylaing, line 55: Add `UpdateOne`. To do this, we need to make this update statement NOT have a WHERE clause. The call to `Update` (changed to `UpdateByPrimaryKey`) will add the clause. `UpdateOne` will accept a clause and expect to only update one row.

lxd/db/query/generic.go
markylaing: Have renamed `Delete` and `Update` to `DeleteByPrimaryKey` and `UpdateByPrimaryKey` respectively.

markylaing: Have also added `UpdateOne`, accepting a clause and args, mirroring the `DeleteOne` function. There will be no `UpdateMany` function.

lxd/db/cluster/generated.go
markylaing: Done.

markylaing:
Have renamed the generator to `dbgen`. Also, the generator will now no longer implement `Updatable`, `Creatable`, or `Referencable` for any table specs that contain a JOIN. These will make these types selectable, but not mutable.

lxd/db/cluster/auth_groups.go
```diff
@@ -4,49 +4,39 @@ import (
...
+	return true, nil
 }
 
-// AuthGroupFilter contains fields upon which an AuthGroup can be filtered.
-type AuthGroupFilter struct {
-	ID   *int
-	Name *string
+// GetAuthGroup gets a single AuthGroup by name.
+func GetAuthGroup(ctx context.Context, tx *sql.Tx, groupName string) (*AuthGroup, error) {
+	return query.SelectOne[AuthGroup](ctx, tx, "WHERE name = ?", groupName)
```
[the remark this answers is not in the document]
markylaing: This is now fixed with standardised error messages for 404 Not Found and 409 Conflict. 

Types can implement `PrettyNamer` to output an API friendly name. If this is not implemented, `reflect.TypeOf(t).Name()` returns the type name without the package name.

lxd/db/query/generic.go
```diff
@@ -0,0 +1,282 @@
...
+
+// Create creates a [Creatable]. All columns are set except for the primary key.
+// This is because it is assumed that the primary key is auto-assigned at the database layer.
+func Create(ctx context.Context, tx *sql.Tx, c Creatable) (int64, error) {
+	tableName := c.TableName()
+
+	res, err := tx.ExecContext(ctx, c.CreateStmt(), c.CreateValues()...)
+	if err != nil {
+		if IsConflictErr(err) {
+			return -1, api.StatusErrorf(http.StatusConflict, "Entry in table %q already exists: %w", tableName, err)
```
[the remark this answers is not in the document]
markylaing: This is fixed.

lxd/db/query/generic.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	err := SelectFunc[T, PT](ctx, tx, clause, f, args...)
+	if err != nil {
+		return nil, err
+	}
+
+	if pt == nil {
+		return nil, api.StatusErrorf(http.StatusNotFound, "%T not found", *(new(T)))
+	}
```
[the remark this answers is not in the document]
markylaing: Fixed.

lxd/db/dbgen/main.go
```diff
@@ -0,0 +1,233 @@
...
+		for _, field := range structTypeSpec.Fields.List {
+			// Fields without a tag are ignored.
+			if field.Tag == nil {
+				continue
+			}
+
+			// Set the field name. This is used to map database columns to struct values.
+			newFieldSpec := FieldSpec{
+				FieldName: field.Names[0].Name,
+			}
```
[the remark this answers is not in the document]
markylaing: Embedded fields are now allowed. Anonymous fields are skipped.

lxd/auth_groups.go
```diff
@@ -186,11 +187,11 @@ func getAuthGroups(d *Daemon, r *http.Request) response.Response {
...
-	groupsIdentities := make(map[int][]dbCluster.Identity)
-	groupsIdentityProviderGroups := make(map[int][]dbCluster.IdentityProviderGroup)
+	groupsIdentities := make(map[int64][]dbCluster.Identity)
+	groupsIdentityProviderGroups := make(map[int64][]dbCluster.IdentityProviderGroup)
 	entityURLs := make(map[entity.Type]map[int]*api.URL)
 	err = d.db.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
-		allGroups, err := dbCluster.GetAuthGroups(ctx, tx.Tx())
+		allGroups, err := query.Select[dbCluster.AuthGroup](ctx, tx.Tx(), "")
 		if err != nil {
 			return err
```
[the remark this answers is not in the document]
markylaing: Fixed.

lxd/db/cluster/identities.go
```diff
@@ -460,31 +460,11 @@ func GetPendingTLSIdentityByTokenSecret(ctx context.Context, tx *sql.Tx, secret
...
-
-	err := query.Scan(ctx, tx, stmt, dest, identityID)
-	if err != nil {
-		return nil, fmt.Errorf("Failed getting groups for identity with ID %d: %w", identityID, err)
-	}
-
-	return result, nil
+WHERE identities_auth_groups.identity_id = ?
+`
+	return query.Select[AuthGroup](ctx, tx, clause, identityID)
```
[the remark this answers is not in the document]
markylaing: This is superfluous.

markylaing:
@tomponline getting joins to work well is important for this to be adopted. I've made it work in this PR and demonstrated with placement groups.

The gist of the expected pattern is:
1. Create type `MyTable` that exactly models a row of the corresponding table.
2. Create type `MyTableFull` that joins other tables, embedding the first type.

Event though `MyTableFull` has joins and no `Update/Delete/Create` related methods were generated for it, it still implements the interfaces by way of the embedded `MyTable`. 

To prevent usage of `MyTableFull` with `Update/Create/Delete`, both `MyTable` and `MyTableFull` implement `IsMutable() bool`, which returns `true` for `MyTable` and `false` for `MyTableFull`. 

This causes `Create/Update/Delete` to fail loudly when called with `MyTableFull`, but to work fine when called with `MyTableFull.MyTable`. We can still select `MyTableFull` so that we can e.g. join projects to get the project name.

It's probably best to review how this works with placement groups to get the gist. cc @kadinsayani @nmezhenskyi

lxd/auth_groups.go
```diff
@@ -777,7 +779,7 @@ func deleteAuthGroup(d *Daemon, r *http.Request) response.Response {
 
 	s := d.State()
 	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
-		return dbCluster.DeleteAuthGroup(ctx, tx.Tx(), groupName)
+		return query.DeleteOne[dbCluster.AuthGroup](ctx, tx.Tx(), "WHERE name = ?", groupName)
 	})
 	if err != nil {
```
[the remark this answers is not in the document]
markylaing: Fixed via `reflect.TypeOf(t).Name` and the `PrettyNamer` interface.

markylaing:
> @tomponline getting joins to work well is important for this to be adopted. I've made it work in this PR and demonstrated with placement groups.
> 
> The gist of the expected pattern is:
> 
> 1. Create type `MyTable` that exactly models a row of the corresponding table.
> 2. Create type `MyTableFull` that joins other tables, embedding the first type.
> 
> Event though `MyTableFull` has joins and no `Update/Delete/Create` related methods were generated for it, it still implements the interfaces by way of the embedded `MyTable`.
> 
> To prevent usage of `MyTableFull` with `Update/Create/Delete`, both `MyTable` and `MyTableFull` implement `IsMutable() bool`, which returns `true` for `MyTable` and `false` for `MyTableFull`.
> 
> This causes `Create/Update/Delete` to fail loudly when called with `MyTableFull`, but to work fine when called with `MyTableFull.MyTable`. We can still select `MyTableFull` so that we can e.g. join projects to get the project name.
> 
> It's probably best to review how this works with placement groups to get the gist. cc @kadinsayani @nmezhenskyi

Disregard the above. After some thinking, I didn't like the `IsMutable` pattern because although invalid calls would return an error, they were still allowed.

Rather than embedding a type to represent the row. I've made it an explicit field, which can by convention be called `Row` (but can be named anything). 

For placement groups, this looks like:
```go
// PlacementGroupsRow represents a single row of the placement_groups table.
// db:model placement_groups
type PlacementGroupsRow struct {
	ID          int64  `db:"id"`
	Name        string `db:"name"`
	Description string `db:"description"`
	ProjectID   int64  `db:"project_id"`
}

// PlacementGroup contains [PlacementGroupsRow] with additional joins.
// db:model placement_groups
type PlacementGroup struct {
	Row PlacementGroupsRow

	// db:join JOIN projects ON placement_groups.project_id = projects.id
	ProjectName string `db:"projects.name"`
}
```

The generator will only implement methods required for database mutations for `PlacementGroupsRow`, but methods for querying will be implemented for both. In practice, this looks like:
```go
pg, err := query.SelectOne[PlacementGroup](ctx, tx, "WHERE placement_groups.name = ? AND projects.name = ?", name, projectName)
if err != nil {
    return err
}

// Rename
 pg.Row.Name = newName
query.UpdateByPrimaryKey(ctx, tx, pg.Row)

// Delete
query.DeleteByPrimaryKey(ctx, tx, pg.Row)
```

This leaves no ambiguity

tomponline:
> The generator will only implement methods required for database mutations for `PlacementGroupsRow`, but methods for querying will be implemented for both

Hrm, is this true?

The `UpdateByPrimaryKey` function is looking for input that implements `Updatable`, and by embedding `PlacementGroupsRow` into `PlacementGroup` it means that `PlacementGroup` now implements `Updatable` (albeit it only the fields that can be updated).

But the original problem still remains, one can update the value of a non-updatable field and then pass that struct into `UpdateByPrimayKey` and it will be silently ignored.

I think it would be safer and clearer to unexport `UpdateByPrimayKey` and then have `dbgen` generate `Update` (and delete) functions for each Row-type, that take the concrete Row type as an argument.

This way if one tries to pass `PlacementGroup` into `UpdatePlacementGroup()` it will fail at compile time.

The `UpdatePlacementGroup()` function can then call the generic `UpdateByPrimaryKey()` function to avoid the duplication that the current dbgenerator does.

markylaing:
> The `UpdateByPrimaryKey` function is looking for input that implements `Updatable`, and by embedding `PlacementGroupsRow` into `PlacementGroup` it means that `PlacementGroup` now implements `Updatable` (albeit it only the fields that can be updated).

`PlacementGroupsRow` is not embedded, it is a field. So `PlacementGroup` doesn't receive the methods via struct embedding.

> This way if one tries to pass PlacementGroup into UpdatePlacementGroup() it will fail at compile time

This happens already with the above implementation:

image

lxd/db/query/generic.go
```diff
@@ -0,0 +1,320 @@
...
+
+	return pt, nil
+}
+
+// DeleteByPrimaryKey deletes a [Referencable] type by primary key.
+// The expected usage is to get a [Referencable] (e.g. via [SelectOne]) and then call Delete on it.
+// Note that this function does not call DeleteOne. This is because the Referencable argument might
+// not satisfy the type constraint.
+func DeleteByPrimaryKey[T Referencable](ctx context.Context, tx *sql.Tx, t T) error {
```
tomponline: Referenceable not Referencable

tomponline: or Referable?

tomponline: or Rowable/Rower?

lxd/db/cluster/stmt.go
```diff
@@ -1,5 +1,7 @@
 //go:build linux && cgo && !agent
```
tomponline, line 2: is empty line needed?

lxd/db/cluster/auth_groups.go
```diff
@@ -11,42 +11,17 @@ import (
...
+	Name        string `db:"name"`
+	Description string `db:"description"`
 }
 
-// AuthGroupFilter contains fields upon which an AuthGroup can be filtered.
-type AuthGroupFilter struct {
-	ID   *int
-	Name *string
+// PrettyName implements [query.PrettyNamer] for API friendly error messages.
+func (AuthGroup) PrettyName() string {
```
tomponline: Can we call this `Description()` or `Label()`? Rather than `PrettyName()`, as the "pretty" isn't too useful from a description point of view.

lxd/auth_groups.go
```diff
@@ -185,50 +186,51 @@ func getAuthGroups(d *Daemon, r *http.Request) response.Response {
...
 
 	var groups []dbCluster.AuthGroup
+	var groupURLs []string
 	var authGroupPermissions []dbCluster.Permission
 	groupsIdentities := make(map[int64][]dbCluster.Identity)
 	groupsIdentityProviderGroups := make(map[int64][]dbCluster.IdentityProviderGroup)
 	entityURLs := make(map[entity.Type]map[int]*api.URL)
 	err = d.db.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
-		allGroups, err := dbCluster.GetAuthGroups(ctx, tx.Tx())
+		err := query.SelectFunc[dbCluster.AuthGroup](ctx, tx.Tx(), "ORDER BY auth_groups.name", func(group dbCluster.AuthGroup) error {
```
tomponline: @markylaing we need to be really careful with this pattern.

Because if `d.db.Cluster.Transaction` is restarted (under the hood) during one of its queries then it will re-append the some of the same content multiple times to these slices.

I think it would be safer and cleaner to encourage specific functions to be written that populate slice(s) and return it, which can then be directly assigned to the outer slice(s) - similar to `GetAllIdentitiesByAuthGroupIDs` and `GetAllIdentityProviderGroupsByGroupIDs` below.

lxd/db/cluster/placement_groups.go
```diff
@@ -129,7 +129,7 @@ func (p *PlacementGroup) ToAPI(ctx context.Context, tx *sql.Tx) (*api.PlacementG
 }
 
 // GetPlacementGroupUsedBy returns a list of URLs of all instances and profiles that reference placement groups matching the provided [PlacementGroupFilter].
-func GetPlacementGroupUsedBy(ctx context.Context, tx *sql.Tx, filter PlacementGroupFilter, firstOnly bool) ([]string, error) {
+func GetPlacementGroupUsedBy(ctx context.Context, tx *sql.Tx, name *string, project *string, firstOnly bool) ([]string, error) {
```
tomponline: projectName?

lxd/db/cluster/placement_groups.go
```diff
@@ -199,12 +199,8 @@ WHERE profiles_config.key = 'placement.group'`)
 
 // GetInstancesInPlacementGroup returns a map of member (node) ID to a slice of instance IDs for all instances that reference the given placement group either directly or indirectly via a profile.
 // The target placement group is specified using a [PlacementGroupFilter] which must contain both [PlacementGroupFilter.Project] and [PlacementGroupFilter.Name].
-func GetInstancesInPlacementGroup(ctx context.Context, tx *sql.Tx, filter PlacementGroupFilter) (map[int][]int, error) {
-	if filter.Project == nil || filter.Name == nil {
-		return nil, errors.New("Project and placement group name must be provided")
-	}
-
-	args := []any{*filter.Project, *filter.Name}
+func GetInstancesInPlacementGroup(ctx context.Context, tx *sql.Tx, name string, project string, nodeID *int64) (map[int64][]int64, error) {
```
tomponline: projectName?

lxd/auth_groups.go
markylaing: Here I was thinking I was being so elegant and clever. Completely forgot about that 🤦

Good catch. Will fix in the morning.

lxd/db/query/generic.go
markylaing: I went with `Referencable` (albeit with a misspelling) to keep it contextually similar to a REFERENCE in the database, which is always via the primary key.

markylaing: Will fix the misspelling of course.

lxd/db/cluster/stmt.go
markylaing: I think this is being done by `gofmt`. Will check though.

lxd/db/cluster/auth_groups.go
markylaing: Can I call it `APIName()`? 

I need to avoid anything that could collide with a field (column) name.

lxd/db/cluster/stmt.go
markylaing: Yeah both `goimports` and `gofmt` place a newline there.

lxd/db/query/generic.go
markylaing: Fixed.

lxd/db/cluster/auth_groups.go
markylaing: Changed to `APIName`

lxd/auth_groups.go
markylaing: Fixed. I've created functions to do this (one for auth groups and one for placement groups). Both are defined in the `lxd/db/cluster` package and return exactly what is needed.

I also noticed that the transaction in `GET /1.0/placement-groups` was appending to a `[]*api.PlacementGroup` that was defined outside the transaction, so I've fixed that as well.

lxd/db/cluster/placement_groups.go
markylaing: Fixed.

lxd/db/cluster/placement_groups.go
markylaing: Fixed.

lxd/db/query/generic.go
```diff
@@ -0,0 +1,320 @@
...
+	BaseQuery() string
+}
+
+// ScanArger specifies how a type should advertise arguments for scanning rows.
+// The order of these arguments should match the order of columns from [BaseQuerier].
+type ScanArger interface {
+	ScanArgs() []any
+}
+
+// getName gets a pretty name for a type for returning 404 Not Found or 409 Conflict errors.
```
markylaing: fix

lxd/db/query/generic.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+	return n, nil
+}
+
+// DeleteOne deletes all rows from the table described by the given [TableNamer] that match the given clause.
+// If the number of affected rows is not one, an error is returned (and the transaction should be rolled back).
+// The type constraint enforces that both the type and a pointer to the type implement the [TableNamer] interface.
+// This is the only way to enforce that (*new(T)) is a non-pointer type, and therefore non-nil.
+func DeleteOne[T TableNamer, PT interface {
```
markylaing: T Referenceable

lxd/db/query/generic.go
```diff
@@ -0,0 +1,320 @@
...
+// Updatable defines a type that knows how to update itself.
+type Updatable interface {
+	Referenceable
+	CreateValuer
+	UpdateStmt() string
+}
+
+// BaseQuerier specifies how to query for all the basic data for a given type.
+// The order of the columns should match the order from [ScanArger].
+type BaseQuerier interface {
```
markylaing: Selectable {
TableName
ScanColumns
}

lxd/db/query/generic.go
```diff
@@ -0,0 +1,320 @@
...
+	"reflect"
+	"strings"
+
+	"github.com/canonical/lxd/shared/api"
+)
+
+// APINamer is used by the generic functions below to return API friendly error messages.
+type APINamer interface {
+	APIName() string
+}
```
markylaing, line 17: Generate this

lxd/db/cluster/placement_groups.go
```diff
@@ -14,45 +14,69 @@ import (
...
 
-// PlacementGroupFilter contains fields that can be used to filter results when getting placement groups.
-type PlacementGroupFilter struct {
-	ID      *int // Used to exclude placement group instances on the source cluster member during evacuation.
-	Project *string
-	Name    *string
+// APIName implements [query.APINamer] for API friendly error messages.
+func (PlacementGroup) APIName() string {
+	return "Placement group"
+}
```
markylaing: Should be generated and call the rows APIName

lxd/db/cluster/placement_groups.go
```diff
@@ -14,45 +14,69 @@ import (
...
+// db:model placement_groups
+type PlacementGroupsRow struct {
+	ID          int64  `db:"id"`
+	Name        string `db:"name"`
+	Description string `db:"description"`
+	ProjectID   int64  `db:"project_id"`
+}
+
+// APIName implements [query.APINamer] for API friendly error messages.
+func (PlacementGroupsRow) APIName() string {
```
markylaing, line 27: Add to all interfaces

lxd/db/cluster/auth_groups.go
```diff
@@ -4,49 +4,67 @@ import (
...
-//go:generate mapper method -i -e auth_group Create
-//go:generate mapper method -i -e auth_group DeleteOne-by-Name
-//go:generate mapper method -i -e auth_group Update
-//go:generate mapper method -i -e auth_group Rename
-//go:generate goimports -w auth_groups.mapper.go
-//go:generate goimports -w auth_groups.interface.mapper.go
-
 // AuthGroup is the database representation of an api.AuthGroup.
+// db:model auth_groups
 type AuthGroup struct {
```
markylaing: `AuthGroupsRow`

markylaing:
All discussed changes made :partying_face:

lxd/db/cluster/placement_groups.go
```diff
@@ -42,6 +42,38 @@ func GetPlacementGroup(ctx context.Context, tx *sql.Tx, name string, projectName
...
 }
 
+// GetPlacementGroupsAndURLs queries for all placement groups and then applies the given filter to the result.
+// This is useful when filtering by groups the caller is able to view.
+// The filter must return true to include an entry, and false to reject an entry.
+// A slice of (filtered) placement group URLs is also returned for convenience.
+// If the project name argument is non-nil, only placement groups in that project are returned.
+// If the project name is nil, placement groups from all projects are returned.
+func GetPlacementGroupsAndURLs(ctx context.Context, tx *sql.Tx, projectName *string, filter func(group PlacementGroup) bool) ([]PlacementGroup, []string, error) {
+	clause := "ORDER BY projects.name, placement_groups.name"
```
tomponline: please can you use a strings.Builder and then add the ORDER by after the optional WHERE so that its not duplicated.

lxd/db/query/generic.go
```diff
@@ -0,0 +1,313 @@
...
+		return fmt.Errorf("Failed updating %q entry: %w", tableName, err)
+	}
+
+	n, err := res.RowsAffected()
+	if err != nil {
+		return fmt.Errorf("Failed verifying update to table %q entry: %w", tableName, err)
+	}
+
+	if n != 1 {
+		return fmt.Errorf("Update by primary key where `%s.%s = %v` affected %d rows", tableName, u.PKColumn(), u.PKValue(), n)
```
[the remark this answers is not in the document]
markylaing: This is intentional. If we call `UpdateByPrimaryKey` then we expect the entity to exist. If it doesn't exist then it is an internal error and not a user error.

lxd/db/query/generic.go
```diff
@@ -0,0 +1,313 @@
...
+	if err != nil {
+		return fmt.Errorf("Failed deleting from table %q: %w", tableName, err)
+	}
+
+	deleted, err := result.RowsAffected()
+	if err != nil {
+		return fmt.Errorf("Failed getting number of deleted rows: %w", err)
+	}
+
+	if deleted != 1 {
```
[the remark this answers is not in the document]
markylaing: This is intentional. If we call `DeleteByPrimaryKey` then we expect the entity to exist. If it doesn't exist then it is an internal error and not a user error.

lxd/placement_groups.go
```diff
@@ -630,24 +619,19 @@ func placementGroupPut(d *Daemon, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
-	dbPlacementGroup := cluster.PlacementGroup{
+	dbPlacementGroup := cluster.PlacementGroupsRow{
+		ID:          existingID,
```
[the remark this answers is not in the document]
tomponline: @markylaing this a good point, seems like we should avoid this pattern of plucking specific fields into vars inside the transaction and instead store the entire `dbGroup` and update the fields we need. Otherwise its asking for mistakes to be made - especially when new fields are added in the future.

lxd/db/query/generic.go
```diff
@@ -0,0 +1,313 @@
...
+	return id, nil
+}
+
+// UpdateByPrimaryKey updates an [Updatable] type by its primary key. It sets all fields except for the primary key.
+// The expected usage is to get an [Updatable] entry from the database (e.g. via [SelectOne]), change fields on it
+// directly, and then call Update. This pattern is encouraged so that we get an entity, perform an authorization check,
+// and then update the values.
+func UpdateByPrimaryKey(ctx context.Context, tx *sql.Tx, u Updatable) error {
+	tableName := u.TableName()
+	res, err := tx.ExecContext(ctx, u.UpdateStmt()+" WHERE "+u.PKColumn()+" = ?", append(u.CreateValues(), u.PKValue())...)
```
tomponline: suggest using strings.Builder here to concatenate the various strings.

lxd/db/query/generic.go
```diff
@@ -0,0 +1,313 @@
...
+
+	return nil
+}
+
+// UpdateOne updates a single [Updatable] that matches the given clause. It sets all fields except for the primary key.
+// The args variadic should contain only the bind arguments for the given clause.
+// Remaining bind arguments are defined by the [Updatable] type.
+func UpdateOne(ctx context.Context, tx *sql.Tx, u Updatable, clause string, args ...any) error {
+	tableName := u.TableName()
+	res, err := tx.ExecContext(ctx, u.UpdateStmt()+" "+clause, append(u.CreateValues(), args...)...)
```
tomponline: strings builder might be good here too

lxd/placement_groups.go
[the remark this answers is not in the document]
markylaing: Yep absolutely, was just thinking the same :+1:

lxd/db/query/generic.go
```diff
@@ -0,0 +1,313 @@
...
+// Supplementary SQL can be passed in to specify additional joins, clauses, or aggregations.
+// The type constraint specifies that the non-pointer type must implement BaseQuerier, but a pointer to the same type
+// must implement [ScanArger]. This is important because it means that new(T), a pointer to T, is non-nil.
+// It also means that we call ScanArgs on the pointer to the type, so we scan a row directly into each fields' memory.
+func SelectFunc[T Selectable, PT interface {
+	ScanArgs() []any
+	*T
+}](ctx context.Context, tx *sql.Tx, clause string, rowFunc func(T) error, args ...any) error {
+	t := *new(T)
+	q := `SELECT ` + strings.Join(t.SelectColumns(), ", ") + ` FROM ` + t.TableName()
```
tomponline: strings.Join uses strings.builder under the hood https://cs.opensource.google/go/go/+/refs/tags/go1.26.1:src/strings/strings.go;l=487

but as you're joining other strings as well, it might be better to just use a single strings.builder for all the concatenations.

lxd/db/cluster/auth_groups.go
```diff
@@ -24,6 +25,25 @@ func (AuthGroupsRow) APIName() string {
...
 
+// AuthGroupExists returns true if an AuthGroup exists and false otherwise.
+func AuthGroupExists(ctx context.Context, tx *sql.Tx, groupName string) (bool, error) {
+	_, err := GetAuthGroup(ctx, tx, groupName)
+	if err != nil {
+		if api.StatusErrorCheck(err, http.StatusNotFound) {
+			return false, nil
+		}
+
+		return false, fmt.Errorf("Failed executing query for auth group by name: %w", err)
```
tomponline: Do we need to wrap this error, given that the err returned from `GetAuthGroup` should already be wrapping with a "Failed loading..." type message?

lxd/db/cluster/placement_groups.go
```diff
@@ -124,7 +124,7 @@ func (p *PlacementGroup) ToAPI(ctx context.Context, tx *sql.Tx) (*api.PlacementG
 }
 
 // GetPlacementGroupUsedBy returns a list of URLs of all instances and profiles that reference placement groups matching the provided [PlacementGroupFilter].
-func GetPlacementGroupUsedBy(ctx context.Context, tx *sql.Tx, filter PlacementGroupFilter, firstOnly bool) ([]string, error) {
+func GetPlacementGroupUsedBy(ctx context.Context, tx *sql.Tx, name *string, projectName *string, firstOnly bool) ([]string, error) {
```
tomponline: what was the reason to get rid of `PlacementGroupFilter` btw?

markylaing: It was only used in two places and I thought the result of removing it was clearer, since in one case all of the optional fields were actually required.

tomponline: Shall we stick with the existing filter convention as we were trending towards more filter args rather than less.

markylaing:
> I also wondered why you got rid of the PlacementGroupFilter concept, as that idea is used fairly often through the code base, instead of many optional arguments. So would like to consider not dropping that unless there is a pressing reason to.

Happy to keep it for the one case where it is useful (placement group used by) but the other function (listing instances using a placement group) was using it in a weird way so I'd like to still get rid of that.

tomponline:
> > I also wondered why you got rid of the PlacementGroupFilter concept, as that idea is used fairly often through the code base, instead of many optional arguments. So would like to consider not dropping that unless there is a pressing reason to.
> 
> Happy to keep it for the one case where it is useful (placement group used by) but the other function (listing instances using a placement group) was using it in a weird way so I'd like to still get rid of that.

OK sounds good.

markylaing:
Should be all resolved.

lxd/db/cluster/auth_groups.go
```diff
@@ -24,6 +25,30 @@ func (AuthGroupsRow) APIName() string {
...
+	}
+
+	return true, nil
+}
+
+// GetAuthGroup gets a single AuthGroup by name.
+func GetAuthGroup(ctx context.Context, tx *sql.Tx, groupName string) (*AuthGroupsRow, error) {
+	group, err := query.SelectOne[AuthGroupsRow](ctx, tx, "WHERE name = ?", groupName)
+	if err != nil {
+		return nil, fmt.Errorf("Failed loading authorization group: %w", err)
```
tomponline, line 46: @markylaing not a blocker for merging, but shouldn't `query.SelectOne` handle generating the "Failed loading <name>" error, to save every caller having to do it?

lxd/db/query/generic.go
```diff
@@ -0,0 +1,358 @@
...
+}](ctx context.Context, tx *sql.Tx, clause string, args ...any) ([]T, error) {
+	var ts []T
+	f := func(t T) error {
+		ts = append(ts, t)
+		return nil
+	}
+
+	err := SelectFunc[T, PT](ctx, tx, clause, f, args...)
+	if err != nil {
+		return nil, err
```
tomponline, line 184: Not a blocker, but to avoid callers having to wrap the error from this in in the entity type context, could this function return a wrapped error like:

> Failed loading (entity type)(s) ...

lxd/db/cluster/placement_groups.go
```diff
@@ -43,6 +43,16 @@ type PlacementGroupFilter struct {
 	Name    *string
 }
 
+// GetPlacementGroup gets a [PlacementGroup] by name and project.
+func GetPlacementGroup(ctx context.Context, tx *sql.Tx, name string, projectName string) (*PlacementGroup, error) {
+	group, err := query.SelectOne[PlacementGroup](ctx, tx, "WHERE placement_groups.name = ? AND projects.name = ?", name, projectName)
+	if err != nil {
+		return nil, fmt.Errorf("Failed loading placement group: %w", err)
```
tomponline, line 50: It'd be great if query.SelectOne (or Queryfunc) handle wrapping these errors so callers dont have to.

lxd/db/cluster/placement_groups.go
```diff
@@ -53,6 +53,43 @@ func GetPlacementGroup(ctx context.Context, tx *sql.Tx, name string, projectName
...
+		if filter != nil && !filter(group) {
+			return nil
+		}
+
+		placementGroups = append(placementGroups, group)
+		placementGroupURLs = append(placementGroupURLs, entity.PlacementGroupURL(group.ProjectName, group.Row.Name).String())
+		return nil
+	}, args...)
+	if err != nil {
+		return nil, nil, fmt.Errorf("Failed listing placement groups: %w", err)
```
tomponline, line 87: if SelectFunc could generate an appropriate "Failed loading" type message we wouldnt need this bespoke error here.

PR #17628, opened 2026-02-12
shared/entity: Helper functions to get named URL path parameters

MusicDin:
This is a proposal to add a helper function which parses the entity URL and returns *named* path parameters.

The reason for adding this is that the Terraform provider can construct the entity URL from the user's input, but we also need to be able to *revert* the URL (retrieved from the LXD) back into the user input (in order to compare desired state against actual state). 

Since the helper function `ParseURL` already parses the path arguments in order, we can just link them to path argument names that need to be provided in the same order as found in path. IMO having separate helper function `ParseURLWithNamedArgs` (or something similar) satisfies the goal, and requires no additional changes across LXD. 

Example:
```
TerraformAuthGroupPermission = {
  "entitlement":    "can_view"
  "entity_type":    "instance"
  "entity_project": "foo"
  "args" / "options": {
    "name": "c1"
  }
}

Transaltes to => LXDAuthGroupPermission = {
  "entitlement": "can_view"
  "entity_type": "instance"
  "entity_ref":  "/1.0/instances/c1?project=foo"
}
```

When we parse the permission back, we need to get arguments and their names back in order to check whether the *TerraformPermission* matches the permission in LXD.
In the above example, we need to convert `/1.0/instances/c1?project=foo` into `name=c1` (entity type, project, location are already returned explicitly).

### DIFF
--- a/shared/entity/url_test.go
+++ b/shared/entity/url_test.go
@@ -1,0 +1,299 @@
+package entity
+
+import (
+	"net/url"
+	"slices"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// This test parses the given URL, and construct a new one from the result, where the final/resulting
+// URL must match the original one.
+//
+// The test will fail if any entity type present in [entityTypes] is not covered by the test cases.
+func TestEntityPermissionURL_RoundTrip(t *testing.T) {
+	tests := []struct {
+		Name     string
+		URL      string
+		WantType Type
+		WantArgs map[string]string
+	}{
+		{
+			Name:     "Auth group",
+			URL:      "/1.0/auth/groups/test",
+			WantType: TypeAuthGroup,
+			WantArgs: map[string]string{
+				"name": "test",
+			},
+		},
+		{
+			Name:     "Certificate",
+			URL:      "/1.0/certificates/e974199c-fcb1-4e0d-9db7-92169e8e3f3d",
+			WantType: TypeCertificate,
+			WantArgs: map[string]string{
+				"fingerprint": "e974199c-fcb1-4e0d-9db7-92169e8e3f3d",
+			},
+		},
+		{
+			Name:     "Cluster group",
+			URL:      "/1.0/cluster/groups/foo",
+			WantType: TypeClusterGroup,
+			WantArgs: map[string]string{
+				"name": "foo",
+			},
+		},
+		{
+			Name:     "Cluster member",
+			URL:      "/1.0/cluster/members/foo",
+			WantType: TypeClusterMember,
+			WantArgs: map[string]string{
+				"name": "foo",
+			},
+		},
+		{
+			Name:     "Container",
+			URL:      "/1.0/containers/foo",
+			WantType: TypeContainer,
+			WantArgs: map[string]string{
+				"name": "foo",
+			},
+		},
+		{
+			Name:     "Identity",
+			URL:      "/1.0/auth/identities/oidc/foo",
+			WantType: TypeIdentity,
+			WantArgs: map[string]string{
+				"method":     "oidc",
+				"identifier": "foo",
+			},
+		},
+		{
+			Name:     "Identity provider group",
+			URL:      "/1.0/auth/identity-provider-groups/test",
+			WantType: TypeIdentityProviderGroup,
+			WantArgs: map[string]string{
+				"name": "test",
+			},
+		},
+		{
+			Name:     "Image",
+			URL:      "/1.0/images/000000000000",
+			WantType: TypeImage,
+			WantArgs: map[string]string{
+				"fingerprint": "000000000000",
+			},
+		},
+		{
+			Name:     "Image alias",
+			URL:      "/1.0/images/aliases/ubuntu",
+			WantType: TypeImageAlias,
+			WantArgs: map[string]string{
+				"name": "ubuntu",
+			},
+		},
+		{
+			Name:     "Instance",
+			URL:      "/1.0/instances/c1",
+			WantType: TypeInstance,
+			WantArgs: map[string]string{
+				"name": "c1",
+			},
+		},
+		{
+			Name:     "Instance with project",
+			URL:      "/1.0/instances/c1?project=foo",
+			WantType: TypeInstance,
+			WantArgs: map[string]string{
+				"name": "c1",
+			},
+		},
+		{
+			Name:     "Instance backup",
+			URL:      "/1.0/instances/myvm/backups/mybackup",
+			WantType: TypeInstanceBackup,
+			WantArgs: map[string]string{
+				"instance": "myvm",
+				"name":     "mybackup",
+			},
+		},
+		{
+			Name:     "Instance snapshot",
+			URL:      "/1.0/instances/my-vm/snapshots/snap-0",
+			WantType: TypeInstanceSnapshot,
+			WantArgs: map[string]string{
+				"instance": "my-vm",
+				"name":     "snap-0",
+			},
+		},
+		{
+			Name:     "Network",
+			URL:      "/1.0/networks/lxdbr0",
+			WantType: TypeNetwork,
+			WantArgs: map[string]string{
+				"name": "lxdbr0",
+			},
+		},
+		{
+			Name:     "Network ACL",
+			URL:      "/1.0/network-acls/1.2.3.4",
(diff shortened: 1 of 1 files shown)


tomponline:
Seems reasonable to me!

MusicDin:
@tomponline Had a discussion with @markylaing and since auth is planned to rely on ids in the future, we might just proceed with this approach. So if I have blessing from both of you I am happy with this PR :)

tomponline:
@markylaing @MusicDin what became of putting it in metadata api?

MusicDin:
So we were discussing that, and it would be possible to retain URL template with named placeholders (for example instead of `{placeholder}` we would use `{name}`). However, having that in metadata means the client needs to fetch a lot of data to resolve entity reference. If client does not persist metadata it results in 1 fetch per request. The main benefit is that any client (not just Go clients) benefit from it.

Having that in shared means everything is resolved locally. The only downside is that if new entity is introduced, client needs to update LXD dependency, which most likely would need to be updated anyway for the new client. The benefit however is that this does not require refactoring and is just an additional helper that also provides path argument names.

shared/entity/type.go
```diff
@@ -222,6 +230,10 @@ func (image) path() []string {
 	return []string{"images", pathPlaceholder}
 }
 
+func (image) pathArgNames() []string {
+	return []string{"name"}
```
markylaing: `fingerprint`?

shared/entity/type.go
```diff
@@ -258,6 +278,10 @@ func (certificate) path() []string {
 	return []string{"certificates", pathPlaceholder}
 }
 
+func (certificate) pathArgNames() []string {
+	return []string{"name"}
```
markylaing: `fingerprint`?

shared/entity/type.go
```diff
@@ -342,6 +390,10 @@ func (operation) path() []string {
 	return []string{"operations", pathPlaceholder}
 }
 
+func (operation) pathArgNames() []string {
+	return []string{"id"}
```
markylaing: I'm wondering if, rather than adding code into shared that people may come to rely on, we should remove the `operation` and `warning` entity types altogether. We'll need to check if they are used anywhere. IIRC I added these entity types to match the existing warning type codes but I don't remember doing a comprehensive review of why certain types exist.

shared/entity/type.go
```diff
@@ -408,6 +476,10 @@ func (warning) path() []string {
 	return []string{"warnings", pathPlaceholder}
 }
 
+func (warning) pathArgNames() []string {
+	return []string{"id"}
```
markylaing: `uuid`? As above with warnings.

shared/entity/type.go
```diff
@@ -480,6 +572,10 @@ func (identity) path() []string {
 	return []string{"auth", "identities", pathPlaceholder, pathPlaceholder}
 }
 
+func (identity) pathArgNames() []string {
+	return []string{"authentication_method", "name"}
```
markylaing: `authentication_method` -> `method` for brevity? Do we think it's clear enough given the identity type is `identity`?

Also, `name` -> `identifier`. Because for OIDC it is the email address. For TLS it's the certificate fingerprint. For bearer identities it's their UUID. (Subject to change for the latter two as we've now made names unique, but we can't change the OIDC one).

MusicDin:
How precise do we want to be with naming?

I am wondering if it would be acceptable to use `identifier` as final path argument everywhere? To me *identifier* identifies the entity, so `name`, `uuid`, `identifier` could all be named `identifier`. In such case the user must know what value identifies the entity.

If we use different final path argument names, it makes the naming *correct*, but the user needs to know what key and value identities the entity.

Maybe this is just a nit, but was thinking from Terraform perspective to have something like that:
```
Permission: {
  Entitlement: ...
  EntityType: ...
  EntityProject: ...
  
  EntityIdentifier: <uuid/name/identifier>

  // Other required identifiers.
  Options: {
    "key": "value",
  }
}
```

shared/entity/type.go
MusicDin: Currently, I see warning and operation type constants used only as a metric type.

markylaing: Maybe we can define those in the metrics package instead then

markylaing:
> How precise do we want to be with naming?
> 
> I am wondering if it would be acceptable to use `identifier` as final path argument everywhere? To me _identifier_ identifies the entity, so `name`, `uuid`, `identifier` could all be named `identifier`. In such case the user must know what value identifies the entity.
> 
> If we use different final path argument names, it makes the naming _correct_, but the user needs to know what key and value identities the entity.
> 
> Maybe this is just a nit, but was thinking from Terraform perspective to have something like that:
> 
> ```
> Permission: {
>   Entitlement: ...
>   EntityType: ...
>   EntityProject: ...
>   
>   EntityIdentifier: <uuid/name/identifier>
> 
>   // Other required identifiers.
>   Options: {
>     "key": "value",
>   }
> }
> ```

So for me this doesn't make a lot of sense:
>   // Other required identifiers.
>   Options: {
>     "key": "value",
>   }

Options implies optional, but they are required?

Why not either a) Just use the URL, or b) put everything into the map?

a)
```
Permission: {
  Entitlement: ...
  EntityType: instance
  EntityUrl: "/1.0/instances/v1?project=foo"
}
```
b)
```
Permission: {
  Entitlement: ...
  EntityType: instance
  EntityDetails: {
    "name": "v1",
    "project": "foo"
}
}
```

MusicDin:
Both A and B are *viable*, but B would be the way to go in terraform just because of how references work there. If you create Terraform resource `lxd_instance` and reference for example name of the instance from another resource, Terraform knows the instance must be handled before the other resource.

So in practice those options would become:
A)
```hcl
entity_url: "/1.0/instances" + ${lxd_instance.instance.name} + "?project=" + ${lxd_instance.instance.project}
``` 

B)
```hcl
entity_details: {
  "name":    lxd_instance.instance.name
  "project": lxd_instance.instance.project
}
```

So B is completely acceptable, and probably as nice as it gets in Terraform.
However, I need to add one more helper function in LXD to be able to construct URL from options/details instead of path arguments, otherwise we will need conditional statements for each entity to be able to pass the slice of correctly ordered path arguments instead of passing just map of values.

lxd/db/cluster/entities.go
```diff
@@ -240,6 +240,7 @@ func GetEntityURL(ctx context.Context, tx *sql.Tx, entityType entity.Type, entit
 	entityRef := &EntityRef{}
 	err := entityRef.scan(row.Scan)
 	if err != nil && !errors.Is(err, sql.ErrNoRows) {
+		logger.Error("scan", logger.Ctx{"stmt": stmt, "err": err})
```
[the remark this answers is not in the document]
MusicDin: Ooops..

shared/entity/type.go
```diff
@@ -388,24 +434,16 @@ func (storageVolumeSnapshot) requiresProject() bool {
...
-	return false
+func (storageVolumeSnapshot) path() []string {
+	return []string{"storage-pools", pathPlaceholder, "volumes", pathPlaceholder, pathPlaceholder, "snapshots", pathPlaceholder}
 }
 
-func (warning) path() []string {
-	return []string{"warnings", pathPlaceholder}
+func (storageVolumeSnapshot) pathArgNames() []string {
+	return []string{"pool", "type", "volume", "name"}
 }
```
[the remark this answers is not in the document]
MusicDin: The order seems fine

shared/entity/type.go
```diff
@@ -158,12 +162,10 @@ var entityTypes = map[Type]typeInfo{
 	TypeNetwork:               network{},
 	TypeNetworkACL:            networkACL{},
 	TypeClusterMember:         clusterMember{},
-	TypeOperation:             operation{},
 	TypeStoragePool:           storagePool{},
 	TypeStorageVolume:         storageVolume{},
 	TypeStorageVolumeBackup:   storageVolumeBackup{},
 	TypeStorageVolumeSnapshot: storageVolumeSnapshot{},
```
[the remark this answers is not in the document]
MusicDin: The constants are referenced throughout the codebase. We left them to have all constants in one place. However, we do not want URL construction/parsing for those entities as they are not used in authorization subsystem.

markylaing:
@MusicDin looks good to me

PR #17632, opened 2026-02-12
lxd/events: Refactor internal listener

skozina:
For cluster locking and wait action of conflicting operations we'd want to listen to operation events. The current internal listener has hard-coded filter for message types and event sources which excludes operation event types and sources from other cluster members. This refactoring allows us to easily create another internal listener with different filters.

## Checklist


skozina:
Lets do this later.
Alternative to this PR would be to move the event filtering from the listener level to the handler level (so each handler could have its own filters).

PR #17633, opened 2026-02-12
Network: Fix `getExternalSubnetInUse` function

nmezhenskyi:
This PR fixes the `getExternalSubnetInUse` function that is used for network validation. 

The PR fixes the underlying database-layer functions `GetProjectNetworkForwardListenAddressesByUplink` and `GetProjectNetworkLoadBalancerListenAddressesByUplink` to correctly handle nullable strings `ipv4_address` and `ipv6_address` when fetching rows from the database.

### DIFF
--- a/test/suites/network_ovn.sh
+++ b/test/suites/network_ovn.sh
@@ -556,6 +556,27 @@ test_network_ovn() {
   # 7. Enable the ipv6.address back.
   lxc network set "${ovn_network}" ipv6.address=fd42:bd85:5f89:5293::1/64 ipv6.nat=true
 
+  sub_test "Check that network forwards and load balancers can be created for IPv6 only network."
+  # 1. Disable the ipv4.address for the network.
+  lxc network set "${ovn_network}" ipv4.address=none ipv4.nat=false
+  # 2. Create one external and one internal network forwards.
+  lxc network forward create "${ovn_network}" 2001:db8:1:2::1
+  lxc network forward create "${ovn_network}" fd42:bd85:5f89:5293::10
+  # 3. Clean up network forwards.
+  lxc network forward delete "${ovn_network}" 2001:db8:1:2::1
+  lxc network forward delete "${ovn_network}" fd42:bd85:5f89:5293::10
+  # 4. Create one external and one internal load balancers.
+  lxc network load-balancer create "${ovn_network}" 2001:db8:1:2::1
+  lxc network load-balancer create "${ovn_network}" fd42:bd85:5f89:5293::10
+  # 5. Clean up load balancers.
+  lxc network load-balancer delete "${ovn_network}" 2001:db8:1:2::1
+  lxc network load-balancer delete "${ovn_network}" fd42:bd85:5f89:5293::10
+  # 6. Check that IPv4 forwards and load balancers cannot be created for IPv6 only network.
+  ! lxc network forward create "${ovn_network}" 192.0.2.1 || false
+  ! lxc network load-balancer create "${ovn_network}" 192.0.2.1 || false
+  # 7. Enable the ipv4.address back.
+  lxc network set "${ovn_network}" ipv4.address=10.24.140.1/24 ipv4.nat=true
+
   echo "Check that instance NIC passthrough with ipv4.routes.external does not allow using volatile.network.ipv4.address."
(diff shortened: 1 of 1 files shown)


test/suites/network_ovn.sh
```diff
@@ -605,6 +605,48 @@ test_network_ovn() {
...
+  lxc network forward delete "${ovn_network}" 192.0.2.1
+  lxc network forward delete "${ovn_network}" 10.24.140.10
+  # 4. Create one external and one internal load balancers.
+  lxc network load-balancer create "${ovn_network}" 192.0.2.1
+  lxc network load-balancer create "${ovn_network}" 10.24.140.10
+  # 5. Clean up load balancers.
+  lxc network load-balancer delete "${ovn_network}" 192.0.2.1
+  lxc network load-balancer delete "${ovn_network}" 10.24.140.10
+  # 6. Check that IPv6 forwards and load balancers cannot be created for IPv4 only network.
+  ! lxc network forward create "${ovn_network}" 2001:db8:1:2::1 || false
```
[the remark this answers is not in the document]
nmezhenskyi: @simondeziel What do you think about the Copilot's suggestion? Should I change how the failure is handled here?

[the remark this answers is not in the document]
simondeziel: @nmezhenskyi it's OK to ignore those as this seems to be a self-inflicted problem. In https://github.com/canonical/lxd/blob/main/.github/copilot-instructions.md#expected-failure we provide a "good" solution (the one you used) and a "best" one which Copilot seems to be aiming for.

The good is good enough and so much lighter to read... probably why we're using it pervasively.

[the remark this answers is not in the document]
nmezhenskyi: @simondeziel I see, thanks!

lxd/db/network_forwards.go
```diff
@@ -309,10 +309,10 @@ func (c *ClusterTx) GetProjectNetworkForwardListenAddressesByUplink(ctx context.
 			var netSubnet *net.IPNet
 			var err error
 
-			if listenAddrIsIP4 && networkIP4Address != "" {
-				_, netSubnet, err = net.ParseCIDR(networkIP4Address)
-			} else if !listenAddrIsIP4 && networkIP6Address != "" {
-				_, netSubnet, err = net.ParseCIDR(networkIP6Address)
+			if listenAddrIsIP4 && networkIP4Address.Valid && networkIP4Address.String != "" {
```
simondeziel: It seems that checking for `.Valid` and `.String != ""` is redundant, isn't it?

simondeziel: In the [Go example](https://pkg.go.dev/database/sql#NullString):
```go
if s.Valid {
   // use s.String
} else {
   // NULL value
}
```

simondeziel: But I think in this case, we should just test for `.String != ""`.

nmezhenskyi: > It seems that checking for `.Valid` and `.String` != "" is redundant, isn't it?

On the one hand, it is, because we don't allow users to set an empty string as an IPv4/IPv6 address.

But I thought it's still good to include this check, because technically the database "could" contain an empty string value and that would be different from `NULL`.

nmezhenskyi: > But I think in this case, we should just test for `.String != ""`.

Yes, we can just do that I think. I'll update the code

nmezhenskyi: Updated

simondeziel (dismissed):

lxd/db/network_forwards.go
tomponline: @nmezhenskyi did you consider solving this in the query itself by using COALLESCE or IFNULL?

nmezhenskyi: > did you consider solving this in the query itself by using COALLESCE or IFNULL?

@tomponline Hm, I could use `COALESCE` and have it return an empty string instead of `NULL` if either `ipv4_address` or `ipv6_address` is not set. Then, there would be no need to use `sql.NullString` types in the Go code.

nmezhenskyi: I've updated the SQL queries to use `COALESCE`

tomponline (approved):
Thanks!

PR #17635, opened 2026-02-12
Fix cluster copy auto-placement when request is forwarded to source

kadinsayani:
Fixes regression from https://github.com/canonical/lxd/commit/7b4ae3b5cf

### DIFF
--- a/lxd/instances_post.go
+++ b/lxd/instances_post.go
@@ -1231,8 +1231,8 @@ func instancesPost(d *Daemon, r *http.Request) response.Response {
 	// Set to true once we find that the request has to be forwarded to the source.
 	redirectToSource := false
 
+	target := request.QueryParam(r, "target")
 	err = s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
-		target := request.QueryParam(r, "target")
 		if !s.ServerClustered && target != "" {
 			return api.StatusErrorf(http.StatusBadRequest, "Target only allowed when clustered")
 		}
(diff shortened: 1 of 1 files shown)


test/suites/clustering.sh
```diff
@@ -368,6 +368,12 @@ test_clustering_containers() {
   [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c L bar)" = "node1" ]
   LXD_DIR="${LXD_THREE_DIR}" lxc delete bar
 
+  echo "Copy the container on node2 without specifying a target, using a client connected to non-source node1."
+  LXD_DIR="${LXD_ONE_DIR}" lxc copy foo auto-copy
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc list -f csv -c n auto-copy)" = "auto-copy" ]
+  [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc list -f csv -c L auto-copy)" != "node2" ]
+  LXD_DIR="${LXD_THREE_DIR}" lxc delete auto-copy
```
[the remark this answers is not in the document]
kadinsayani: We can fix this later.

test/suites/clustering.sh
```diff
@@ -368,6 +368,12 @@ test_clustering_containers() {
   [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c L bar)" = "node1" ]
   LXD_DIR="${LXD_THREE_DIR}" lxc delete bar
 
+  echo "Copy the container on node2 without specifying a target, using a client connected to non-source node1."
+  LXD_DIR="${LXD_ONE_DIR}" lxc copy foo auto-copy
```
roosterfish, line 372: To not silently "break" this test case in the future, should we ensure that `foo` is on `node2`?
Because the test will likely continue to work even if we insert another test somewhere above which moves `foo` to another member.

roosterfish (changes requested):
Thanks for jumping onto this, just one comment on the added test.

test/suites/clustering.sh
roosterfish: I can address this in a follow up PR together with the fix for `--refresh` so we can merge this.

tomponline: yes please thanks @roosterfish

tomponline:
@roosterfish does this fix the refresh connection timeout issue too?

test/suites/clustering.sh
kadinsayani: Thanks @roosterfish!

PR #17638, opened 2026-02-12
Skip broken FAN tests

simondeziel:
GHA runners have switched kernel 6.11 -> 6.14 and it seems that broke the FAN bridges

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -545,7 +545,9 @@ jobs:
         test:
           - cgroup
           - cloud-init
-          - cluster
+          # Known broken on 24.04 with HWE kernel (used by GHA runners starting on 2026-02-12)
+          # (https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2141715)
+          #- cluster
           - container
           - container-copy
           - container-nesting


.github/workflows/tests.yml
```diff
@@ -545,7 +545,9 @@ jobs:
         test:
           - cgroup
           - cloud-init
-          - cluster
+          # Known broken on 24.04 with HWE kernel (used by GHA runners starting on 2026-02-12)
```
tomponline, line 548: Hrm, this seems to broad to be skipping, is the whole test dependent on working fan?

simondeziel: Yes, fortunately in `lxd-ci` the same test keeps running on 22.04, at least.

tomponline: ok, but can you look into getting our tests over to canonical's runners so we can target specific ubuntu os releases with their proper kernels.

simondeziel: Yes, that's long over due by now. We'll likely need to double the testing matrix to cover both the GA and the latest HWE kernels. All feasible.

PR #17643, opened 2026-02-13
identity: Expose bearer token expiration date for current identity

MusicDin:
When client is authenticated using a bearer token, the `/1.0/auth/identities/current` reports token expiration date:

```
lxc query /1.0/auth/identities/current
{
        "authentication_method": "tls",
        "type": "Client certificate (unrestricted)",
        "expires_at": "2035-09-16T13:04:37Z",
        ...
}
```

```
LXD_AUTH_BEARER_TOKEN=<token> lxc query /1.0/auth/identities/current
{
        "authentication_method": "bearer",
        "type": "Client token bearer",
        "expires_at": "2036-02-03T19:42:38Z",
        ...
}
```

### DIFF
--- a/doc/rest-api.yaml
+++ b/doc/rest-api.yaml
@@ -914,6 +914,13 @@ definitions:
                     $ref: '#/definitions/Permission'
                 type: array
                 x-go-name: EffectivePermissions
+            expires_at:
+                description: |-
+                    ExpiresAt is the expiration time of the credential used to authenticate the caller.
+                    It is set only when client is trusted, and authentication method is either bearer or TLS.
+                format: date-time
+                type: string
+                x-go-name: ExpiresAt
             fine_grained:
                 description: |-
                     FineGrained is a boolean indicating whether the identity is fine-grained,


lxd/api_1.0.go
```diff
@@ -438,6 +438,7 @@ func api10Get(d *Daemon, r *http.Request) response.Response {
 	fullSrv.Environment = env
 	fullSrv.AuthUserName = requestor.CallerUsername()
 	fullSrv.AuthUserMethod = requestor.CallerProtocol()
+	fullSrv.AuthTokenExpiresAt = requestor.TokenExpiresAt()
```
markylaing: Does this have to be returned by `GET /1.0`? Could it be returned from `GET /1.0/auth/identities/current`?

tomponline: The same could be said for all of those `requestor` derived fields no?

MusicDin: Fixed.

MusicDin: > The same could be said for all of those requestor derived fields no?

They should all be already included in the current identity response.
I can remove existing ones from `/1.0` but that would break backwards compatibility

tomponline: Why did they get added to /1.0 in the first place, do you know @markylaing was it for the UI needs?

markylaing: I'll have to do a little git archaeology but I think the `trusted` field was added well before the `auth` APIs as that's how the CLI figures out if it added a remote successfully or not. Other fields IIRC we're added so that the UI can show the current user - I think that came before auth APIs as well

markylaing: The UI now uses `GET /1.0/auth/identities/current` to check if the current user is fine-grained so that it can grey out functionality. So putting the token expiration date there won't be an extra call for them.

I haven't reviewed this PR yet but I wonder if this `expiry` could also display certificate expiry for TLS identities?

markylaing: That would allow the UI to display a banner indicating that the client needs to update (as it doesn't have access to the certificate directly like the CLI does).

test/suites/auth.sh
```diff
@@ -206,7 +206,6 @@ fine_grained: true"
 
   [ "$(LXD_CONF="${LXD_CONF2}" lxc auth identity info tls:)" = "${expectedTLSInfo}" ]
 
-  # Need heredoc to retain double quotes for empty TLS certificate.
```
markylaing, line 209: Why remove the comment?

lxd/api_1.0.go
MusicDin: That seems reasonable to me. I suggest renaming the field `token_expires_at` -> `expires_at`. If identity is `bearer` the expiry will indicate token expiry, if identity is TLS, the expiry will indicate certificate expiration date. Does that makes sense?

test/suites/auth.sh
MusicDin: It is old comment. Previously the heredoc was used, but we escaped the quotes within the string instead. The comment should have been removed.

markylaing: Could you please add that as a separate commit then as I was a bit confused :)

lxd/api_1.0.go
markylaing: Sounds good to me :)

To get the certificate expiry, the `SetRequestor` logic can find the certificate in `(*http.Request).TLS.PeerCertificates` that matches the fingerprint in the `RequestorArgs` (there is nothing to stop a client sending multiple certs) and then get the not after.

markylaing: nvm - checking again, on your PR it looks like you can just set the cert expiry directly on the RequestorArgs from the (*Daemon).Authenticate method (much simpler!)

tomponline: > That seems reasonable to me. I suggest renaming the field `token_expires_at` -> `expires_at`. If identity is `bearer` the expiry will indicate token expiry, if identity is TLS, the expiry will indicate certificate expiration date. Does that makes sense?

good idea

test/suites/auth.sh
MusicDin: Fixed

lxd/api_1.0.go
MusicDin: Updated

markylaing:
@MusicDin do we need an API extension so that clients know when they can expect the field to be present?

MusicDin:
Makes sense. How about `identity_credential_expiry`?

tomponline:
> @MusicDin do we need an API extension so that clients know when they can expect the field to be present?

We're going to bundle this in with the bearer token API extension I thought @MusicDin ?

Just because PRs were merged separately, we dont need an API extension for a related feature if its not been released yet.

But you can update the docs for the previous API extension explaining this additional change.

MusicDin:
Perfect, will update

tomponline (approved):
Thanks!

PR #17644, opened 2026-02-13
lxd/api: Include bearer auth method by default

MusicDin:
Include missing auth method in /1.0 endpoint


tomponline:
@MusicDin also rebasing should get you some test fixes.

Issue #17464, opened 2026-01-22
Stateless refresh with running instances is broken


roosterfish:
### Overview

Whilst performing tests around https://github.com/canonical/lxd/pull/17443 (original issue https://github.com/canonical/lxd/issues/13542), I found that running `lxc copy --refresh --stateless` can break the target instance or doesn't even work in the first place. In some cases it cannot be started up anymore afterwards. 
Can we even do a stateful refresh? This seems suspicious to me in the first place. Maybe we just have to block this in the API.

I have made some reproducer clusters just to record the current state.
The same issue can be reproduced in `5.0/stable`, `5.21/stable` and `latest/edge`.
There might be slightly different errors when using VMs instead of containers (as for this reproducer).

Version | Storage | Across members | `--refresh --stateless` works | Breaks target instance
--- | --- | --- | --- | ---
`5.0/stable` | ZFS | yes/no | no | no
`5.0/stable` | Ceph RBD | yes | yes | no
`5.0/stable` | Ceph RBD | no | no | no
`5.21/stable` | ZFS | yes/no | no | no
`5.21/stable` | Ceph RBD | yes | yes | no
`5.21/stable` | Ceph RBD | no | no | no
`6/edge` | ZFS | yes/no | no | no
`6/edge` | Ceph RBD | yes | yes | no
`6/edge` | Ceph RBD | no | yes | yes

See reproducers for all cases below.

### `5.0/stable`

Local storage (ZFS) seems fine, refresh with running instances fails on something else:

```bash
root@micro01:~# lxc ls
+------+---------+---------------------+------+-----------+-----------+----------+
| NAME |  STATE  |        IPV4         | IPV6 |   TYPE    | SNAPSHOTS | LOCATION |
+------+---------+---------------------+------+-----------+-----------+----------+
| c1   | RUNNING | 240.82.0.44 (eth0)  |      | CONTAINER | 0         | micro01  |
+------+---------+---------------------+------+-----------+-----------+----------+
| c2   | RUNNING | 240.203.0.12 (eth0) |      | CONTAINER | 0         | micro02  |
+------+---------+---------------------+------+-----------+-----------+----------+
root@micro01:~# lxc cp c1 c2 --refresh
Error: Error transferring instance data: Failed migration on target: Failed getting instance storage pool name: Instance storage pool not found
root@micro01:~# lxc cp c1 c2 --refresh --stateless
[...]

PR #17645, opened 2026-02-13
Fix: Regressions in instance refresh

roosterfish:
Fixes https://github.com/canonical/lxd/issues/17464 by preventing refresh of running target instances.

Mainly this PR fixes three regressions introduced in https://github.com/canonical/lxd/pull/17347 affecting refresh of instances: 

1) Don't cause instances to be moved to different cluster members after refresh
2) Don't run placement every time we perform a refresh (was ignored/unused before we introduced https://github.com/canonical/lxd/pull/17347 and is now triggering the move after refresh issue).
3) When refreshing internally (remote storage) ensure the target instance is moved back to the source if the source instance is running (same as when doing the initial copy)

By fixing the regressions we also address:
1) Allow refreshing instances on local storage between cluster members without specifying `--target` (fixed by addressing 2))

A matrix test is added to account for all of those scenarios.

### DIFF
--- a/test/suites/clustering.sh
+++ b/test/suites/clustering.sh
@@ -376,6 +376,38 @@ test_clustering_containers() {
   [ "$(LXD_DIR="${LXD_ONE_DIR}" lxc list -f csv -c L auto-copy)" != "node2" ]
   LXD_DIR="${LXD_THREE_DIR}" lxc delete auto-copy
 
+  echo "Refresh a container and check its placement afterwards."
+  # Create stopped base container.
+  LXD_DIR="${LXD_ONE_DIR}" lxc copy foo test-refresh --target node1
+
+  # Perform copy/refresh matrix test from every cluster member to ensure the request forwards work as expected
+  # and test with both started and stopped containers.
+  for state in "" "--start"; do
+    for member in "node1" "node2" "node3"; do
+      echo "Copy base container to target ${member}."
+      # shellcheck disable=SC2248
+      LXD_DIR="${LXD_ONE_DIR}" lxc copy test-refresh test-refresh-target --target "${member}" ${state}
+
+      echo "Check placement is correct."
+      LXD_DIR="${LXD_ONE_DIR}" lxc info test-refresh-target | grep -xF "Location: ${member}"
+
+      echo "Refresh target container."
+      if [ "${state}" = "--start" ]; then
+        echo "Refresh should be blocked if the instance is running."
+        ! LXD_DIR="${LXD_ONE_DIR}" lxc copy test-refresh test-refresh-target --refresh || false
+      else
+        LXD_DIR="${LXD_ONE_DIR}" lxc copy test-refresh test-refresh-target --refresh
+      fi
+
+      echo "Check placement hasn't changed during refresh."
+      LXD_DIR="${LXD_ONE_DIR}" lxc info test-refresh-target | grep -xF "Location: ${member}"
+
+      LXD_DIR="${LXD_ONE_DIR}" lxc delete -f test-refresh-target
+    done
+  done
+
+  LXD_DIR="${LXD_ONE_DIR}" lxc delete -f test-refresh
+
   echo "Copy the container on node2 to node3, using a client connected to node1."
   LXD_DIR="${LXD_ONE_DIR}" lxc copy foo bar --target node3
   [ "$(LXD_DIR="${LXD_TWO_DIR}" lxc list -f csv -c L bar)" = "node3" ]


tomponline:
@roosterfish does this need additional tests to close the gaps?

roosterfish:
> @roosterfish does this need additional tests to close the gaps?

Yes, added a TODO to the desc a few mins ago :)

lxd/instances_post.go
```diff
@@ -685,9 +685,64 @@ func createFromCopy(ctx context.Context, s *state.State, projectName string, pro
...
+			// At this stage we move the entire instance with all of its snapshots.
+			// In case the actual copy operation was requested with InstanceOnly=true, the copied instance doesn't have snapshots.
+			Migration: true,
+		})
+		if err != nil {
+			return err
+		}
+
+		return op.Wait()
+	}
```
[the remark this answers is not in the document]
roosterfish: I don't see where this would panic in it's current form but I have switched to passing the client to make it more obvious.

lxd/instances_post.go
```diff
@@ -1290,6 +1290,28 @@ func instancesPost(d *Daemon, r *http.Request) response.Response {
 			// Identify source member to be able to forward the request to the source in case the request
 			// was made on a different cluster member and the type is copy.
 			if s.ServerClustered && !clusterNotification && req.Source.Type == api.SourceTypeCopy {
+				// When performing a copy with refresh, don't trigger another placement but instead
+				// populate the targetMemberInfo based on the target instance's location.
+				// This ensures the instance isn't moved across the cluster after refresh and
+				// we don't try to identify a different eligible cluster member.
+				if req.Source.Refresh {
+					targetInst, err := instance.LoadInstanceDatabaseObject(ctx, tx, targetProjectName, req.Name)
```
tomponline, line 1354: `instanceCreateAsCopy` also loads the target instance from the DB.

I wonder if we can avoid duplicate DB loading by passing this `targetInst` down from here into that function (via `createFromCopy`) and then using `tx.InstancesToInstanceArgs` rather than `LoadByProjectAndName`?

Something for another PR perhaps.

lxd/instances_post.go
tomponline, line 1: @roosterfish is this a hack? Why does the instance need to be moved back to source (for remote pools) when doing a refresh? How did it end up on the wrong host in the first place?

roosterfish: If the instance is running on the source and we trigger a `lxc cp sourceInst targetInst --refresh`, if we do an internal copy (on remote storage, same pool) the freeze capabilities of the source instance will only take effect when the refresh is running on the member which hosts the source instance. 

Therefore we have to do the same as when doing an actual internal instance copy (on remote storage, same pool) which is doing the copy on the source which allows freezing the FS of the source in case it's running.

roosterfish: Maybe I need to add more explanation to the code.

lxd/instances_post.go
```diff
@@ -1570,8 +1570,29 @@ func instancesPost(d *Daemon, r *http.Request) response.Response {
 		req.Config["volatile.cluster.group"] = targetGroupName
 	}
 
-	// Redirect the request to the target cluster member when not doing an actual copy.
-	if targetMemberInfo != nil && targetMemberInfo.Address != "" && targetMemberInfo.Name != s.ServerName && req.Source.Type != api.SourceTypeCopy {
+	canDoInternalCopy := false
+
+	// When doing a copy check if the request has to be forwarded to the target.
```
tomponline: I'm getting a bit confused by these changes now.

Originally we wanted the copy request to be forwarded to the member hosting the instance so we freeze the source filesystem, but now we are forwarding to the target?

lxd/instances_post.go
tomponline: Right, but the commit message says:

> Move instance back to source when running and on remote storage

and further down the comment says:

> Move the instance to the source member in case of refresh and if the source instance is running.

So we are doing a "copy" but there's multiple references to "moving", and I don't understand why the copied instance is ending up on a member that requires "moving" it back?

lxd/instances_post.go
roosterfish: This is before deciding if we are going the migration or internal copy route.
Let me clarify this with more comments.

lxd/instances_post.go
tomponline: Can we just always have instance copy/refresh calls be redirected to the member hosting the instance?

tomponline:
@roosterfish as discussed lets:

1. Always run internal copy on source irrespective of source run state - to reduce number of variant code paths.
2. Prevent refresh to a running target in all cases.

lxd/instances_post.go
roosterfish: Removed this to avoid confusion.

lxd/instances_post.go
roosterfish: I have now consolidated all cases as much as possible with a clearer decision making an context in comments. Waiting for a successful test run.

roosterfish: Ready for review. Maybe it helps checking the last commit first as it provides the context due to the added comments.

test/suites/clustering.sh
```diff
@@ -369,11 +369,44 @@
...
+  echo "Refresh a container and check it's placement afterwards."
+  # Create stopped base container.
+  LXD_DIR="${LXD_ONE_DIR}" lxc copy foo test-refresh --target node1
+
+  # Perform copy/refresh matrix test from every cluster member to ensure the request forwards work as expected
+  # and test with both started and stopped containers.
+  for state in "" "--start"; do
+    for member in "node1" "node2" "node3"; do
+      echo "Copy base container to target ${member}."
+      LXD_DIR="${LXD_ONE_DIR}" lxc copy test-refresh test-refresh-target --target "${member}" ${state}
```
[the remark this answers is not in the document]
tomponline: @roosterfish did this get addressed?

lxd/instances_post.go
```diff
@@ -685,7 +685,63 @@ func createFromCopy(ctx context.Context, s *state.State, projectName string, pro
 		Stateful:     req.Stateful,
 	}
 
+	moveInstToTarget := func(targetClient lxd.InstanceServer) error {
+		// targetMemberInfo has to be there but this is just a safety check.
+		if targetMemberInfo != nil {
+			targetClient = targetClient.UseTarget(targetMemberInfo.Name)
```
tomponline, line 691: worth creating a new `targetClient` variable rather than modifying the one outside of this function?

Or perhaps if `targetMemberInfo` should be populated and configured by now, instead validate it instead of modifying and return error if not as its just safety check.

test/suites/clustering.sh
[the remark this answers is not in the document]
roosterfish: Jup, I have put the ignore line on top.
We cannot quote the var as `state` can be `""` which would be interpreted as an empty string command line param by LXD if quoted.

lxd/instances_post.go
tomponline, line 582: @roosterfish looks like a "ceph" reference remains in comment here.

roosterfish: Thanks, good catch, I saw this yesterday and missed modifying it.

roosterfish: Fixed in https://github.com/canonical/lxd/pull/17684.

lxd/instances_post.go
roosterfish: Fixed in https://github.com/canonical/lxd/pull/17684. This also helped in reusing the `moveInstToTarget` helper in one more case.

PR #17663, opened 2026-02-16
lxd: Fix various issues with operation entityURLs

skozina:
## Checklist

### DIFF
--- a/test/includes/operations.sh
+++ b/test/includes/operations.sh
@@ -5,6 +5,7 @@ lxd_websocket_operation() {
   local duration="$2"
   local project_name="${3:-default}"
 
+  # op_type 20 is "CommandExec"
   lxc query --wait -X POST -d '{"duration": "'"${duration}"'", "op_class": 2, "op_type": 20, "entity_url": "/1.0/instances/'"${instance_name}"'?project='"${project_name}"'"}' "/internal/testing/operation-wait?project=${project_name}"
 }
 
@@ -16,7 +17,8 @@ lxd_volume_operation() {
   local duration="$3"
   local project_name="${4:-default}"
 
-  lxc query --wait -X POST -d '{"duration": "'"${duration}"'", "op_class": 1, "op_type": 48, "entity_url": "/1.0/storage-pools/'"${pool_name}"'/volumes/custom/'"${volume_name}"'?project='"${project_name}"'"}' "/internal/testing/operation-wait?project=${project_name}"
+  # op_type 32 is "VolumeCopy"
+  lxc query --wait -X POST -d '{"duration": "'"${duration}"'", "op_class": 1, "op_type": 32, "entity_url": "/1.0/storage-pools/'"${pool_name}"'/volumes/custom/'"${volume_name}"'?project='"${project_name}"'"}' "/internal/testing/operation-wait?project=${project_name}"
 }
 
 # check_registered_operations checks for registered operations.


skozina:
Moving to draft, there will be more places where we need to fix effective project of operations entityURL.

lxd/db/cluster/entity_type_storage_volume_backup.go
```diff
@@ -71,7 +71,7 @@ WHERE projects.name = ?
 		WHEN %d THEN '%s' 
 	END = ? 
 	AND storage_volumes.name = ? 
-	AND storage_volumes_backups.name = ?
+	AND storage_volumes_backups.name = concat(storage_volumes.name, '/', ?)
```
markylaing, line 74: This should not be changed unless a storage volume backup name is actually stored in the database as `{parent}/{backup}`

lxd/api_project.go
```diff
@@ -1279,7 +1279,12 @@ func projectDelete(d *Daemon, r *http.Request) response.Response {
 
 		// Prune cached images.
 		for _, image := range cachedImages {
-			op, err := doImageDelete(ctx, s, image.Fingerprint, image.ID, project.Name)
+			effectiveProjectName, err := projecthelpers.ImageProject(ctx, tx.Tx(), project.Name)
```
markylaing: Let's not call this in a loop.

lxd/db/cluster/entity_type_storage_volume_backup.go
skozina: It is stored in the database as `pool/volume`.

skozina: Sorry, I mean `volume/backup`. Eg.: https://github.com/canonical/lxd/blob/63ec667ad6b206be30cd4d10bbd789f7c18a7673/lxd/storage_volumes_backup.go#L417

lxd/instances_post.go
```diff
@@ -756,6 +766,10 @@ func createFromCopy(ctx context.Context, s *state.State, projectName string, pro
 		},
 	}
 
+	if shared.IsSnapshot(req.Source.Source) {
+		opArgs.Type = operationtype.SnapshotCopy
+	}
+
```
markylaing: For me this is very confusing to read. Please can we have:
```go
var entityURL *api.URL
var operationType operationtype.Type
if shared.IsSnapshot(req.Source.Source) {
    cName, sName, _ := api.GetParentAndSnapshotName(req.Source.Source)
    snapshotURL := api.NewURL().Path(version.APIVersion, "instances", cName, "snapshots", sName).Project(req.Source.Project)
    resources[entity.TypeInstanceSnapshot] = []api.URL{*snapshotURL}
    entityURL = snapshotURL
} else {
    instanceURL := api.NewURL().Path(version.APIVersion, "instances", req.Source.Source).Project(req.Source.Project)
    resources[entity.TypeInstance] = []api.URL{*instanceURL}
    entityURL = instanceURL
}
```

lxd/storage_volumes.go
```diff
@@ -1282,7 +1282,7 @@ func doVolumeCreateOrCopy(s *state.State, r *http.Request, requestProjectName st
 		}
 	} else {
 		opType = operationtype.VolumeCopy
-		sourceVolumeURL := entity.StorageVolumeURL(req.Source.Project, req.Source.Location, req.Source.Pool, req.Source.Type, req.Source.Name)
+		sourceVolumeURL := entity.StorageVolumeURL(srcProjectName, req.Source.Location, req.Source.Pool, req.Source.Type, req.Source.Name)
```
markylaing: I'm not sure I follow what the difference is here

lxd/storage_volumes_backup.go
```diff
@@ -660,7 +660,7 @@ func storagePoolVolumeTypeCustomBackupPost(d *Daemon, r *http.Request) response.
 		return nil
 	}
 
-	backupURL := api.NewURL().Path(version.APIVersion, "storage-pools", details.pool.Name(), "volumes", details.volumeTypeName, details.volumeName, "backups", oldName)
+	backupURL := api.NewURL().Path(version.APIVersion, "storage-pools", details.pool.Name(), "volumes", details.volumeTypeName, details.volumeName, "backups", backupName)
```
markylaing, line 655: This seems wrong - it needs to point to the existing backup, which at this point is the one with the old name.

lxd/db/cluster/entity_type_storage_volume_backup.go
markylaing: Do you mean `{volumeName}/{backupName}`?

markylaing: Ah sorry your response didn't load

markylaing: This seems like a bug to me. Why would the backup name encode the parent volume name in the database when it references its parent by foreign key? Very odd. I guess we'll have to work with it for now

skozina: I agree. I suppose we could do a db patch to fix the DB schema and content. I just thought it's now worth it for now.

lxd/storage_volumes.go
skozina: The `srcProjectName` is set on lines 1251 with:
```
	var srcProjectName string
	if req.Source.Project != "" {
		srcProjectName, err = project.StorageVolumeProject(s.DB.Cluster, req.Source.Project, cluster.StoragePoolVolumeTypeCustom)
		if err != nil {
		}
	}
```

The `project.StorageVolumeProject()` call should give us effective project?

lxd/storage_volumes_backup.go
skozina: The `oldName` is set on line 643 with:
```
	oldName := details.volumeName + shared.SnapshotDelimiter + backupName
```

So the `backupName` is still the existing backup, just without the volume prefix. This avoids using entityURL containing things like `volume%2Fbackup`, because the `api.NewURL().Path()` escapes its arguments.

lxd/instances_post.go
skozina: Thanks, I've done this.

lxd/api_project.go
skozina: Pulled out of the loop (the project is the same for all images on project deletion).

lxd/instance_post.go
```diff
@@ -910,9 +907,14 @@ func instancePostClusteringMigrate(s *state.State, srcPool storagePools.Pool, sr
 			return srcMigration.Do(s, op)
 		}
 
+		instanceURL := api.NewURL().Path(version.APIVersion, "instances", srcInstName).Project(srcInst.Project().Name)
+		resources := map[entity.Type][]api.URL{
```
tomponline, line 911: @skozina do we even need to populate resources here now that we have the `EntityURL` for the operation?

tomponline: Something for a future PR though - whether we need to duplicate the EntityURL for an operation into its `Resources` field.

tomponline (approved):
LGTM thanks!

lxd/instance_post.go
skozina: @markylaing might have some details. I think he had to keep some resources at those places where resources were part of external API?

markylaing: Happy to remove. IIRC we said resources would be used as a way of "locking" those resources by way of database triggers. Here this is attempting to say "the primary entity is the source instance and also don't modify it while this operation is running please"

markylaing: Since we don't have that mechanism in place, nor any bugs related to it, we can just omit resources as we please.

tomponline: In that case I suggest we remove all resources being stored for now until such time as we have a reason to use them (i.e for preventing deleting of a related entity), this way we wont need to store a duplicate of the operation's primary entity ID in resources.

PR #17563, opened 2026-02-04
lxd: Add `entity_name` metadata to operations where the name might be generated server side

markylaing:
Will follow with another PR to remove unnecessary URLs from oepration resources.

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md) and attest that all commits in this PR are [signed off](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#including-a-signed-off-by-line-in-your-commits), [cryptographically signed](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#commit-signature-verification), and follow this project's [commit structure](https://github.com/canonical/lxd/blob/main/CONTRIBUTING.md#commit-structure).
- [x] I have checked and added or updated relevant documentation.

PR #17664, opened 2026-02-16
Operations: Missing entity url

markylaing:
#17563 added an API extension to move generated entity URLs out of operation resources and into the operation metadata under an `entity_url` key.

The `entity_url` key is missing on instance creation and also for instance snapshot creation. For instances, this is because the progress writer is calling `UpdateMetadata`. Replacing all the fields. For snapshots, it's because I somehow forgot to add it in the first place :facepalm: 

## Checklist


markylaing:
> @markylaing is this an indication there is a gap in our testing?

It was actually caught by the UI tests and is blocking them. But we should also have tests for it.

tomponline:
> But we should also have tests for it.

I'll merge this PR to unblock @edlerd but can you follow up with a test for this, ta.

Issue #17159, opened 2025-12-04
Remove minio support


tomponline:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

Due to https://github.com/minio/minio/commit/27742d469462e1561c776f88ca7a1f26816d69e2 and https://github.com/minio/minio/commit/7aac2a2c5b7c882e68c1ce017d8256be2feea27f

### Describe the solution you'd like

N/A

Closes https://github.com/canonical/lxd/issues/15026

PR #17673, opened 2026-02-17
Remove `minio` support

simondeziel:
Fixes https://github.com/canonical/lxd/issues/17159

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,6 @@ require (
 	github.com/mdlayher/netx v0.0.0-20230430222610-7e21880baee8
 	github.com/mdlayher/vsock v1.2.1
 	github.com/miekg/dns v1.1.72
-	github.com/minio/minio-go/v7 v7.0.99
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/moby/sys/capability v0.4.0
 	github.com/oklog/ulid/v2 v2.1.1
@@ -80,13 +79,11 @@ require (
 	github.com/cpuguy83/go-md2man/v2 v2.0.7 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/dgryski/go-farm v0.0.0-20240924180020-3414d57e47da // indirect
-	github.com/dustin/go-humanize v1.0.1 // indirect
 	github.com/eapache/channels v1.1.0 // indirect
 	github.com/eapache/queue v1.1.0 // indirect
 	github.com/emirpasic/gods v1.18.1 // indirect
 	github.com/envoyproxy/protoc-gen-validate v1.3.3 // indirect
 	github.com/fsnotify/fsnotify v1.9.0 // indirect
-	github.com/go-ini/ini v1.67.0 // indirect
 	github.com/go-logr/logr v1.4.3 // indirect
 	github.com/go-logr/stdr v1.2.2 // indirect
 	github.com/go-viper/mapstructure/v2 v2.5.0 // indirect
@@ -106,15 +103,11 @@ require (
 	github.com/juju/schema v1.2.0 // indirect
 	github.com/juju/version v0.0.0-20210303051006-2015802527a8 // indirect
 	github.com/k-sone/critbitgo v1.4.0 // indirect
-	github.com/klauspost/compress v1.18.4 // indirect
 	github.com/klauspost/cpuid/v2 v2.3.0 // indirect
-	github.com/klauspost/crc32 v1.3.0 // indirect
 	github.com/kr/fs v0.1.0 // indirect
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mattn/go-runewidth v0.0.21 // indirect
 	github.com/mdlayher/socket v0.5.1 // indirect
-	github.com/minio/crc64nvme v1.1.1 // indirect
-	github.com/minio/md5-simd v1.1.2 // indirect
 	github.com/muhlemmer/gu v0.3.1 // indirect
 	github.com/muhlemmer/httpforwarded v0.1.0 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
@@ -124,7 +117,6 @@ require (
 	github.com/opencontainers/runtime-spec v1.3.0 // indirect
 	github.com/opencontainers/runtime-tools v0.9.1-0.20251114084447-edf4cb3d2116 // indirect
 	github.com/pelletier/go-toml/v2 v2.2.4 // indirect
-	github.com/philhofer/fwd v1.2.0 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
 	github.com/prometheus/client_golang v1.23.2 // indirect
@@ -132,7 +124,6 @@ require (
 	github.com/prometheus/common v0.67.5 // indirect
 	github.com/prometheus/procfs v0.20.1 // indirect
 	github.com/rs/cors v1.11.1 // indirect
-	github.com/rs/xid v1.6.0 // indirect
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
 	github.com/sagikazarmark/locafero v0.12.0 // indirect
 	github.com/sourcegraph/conc v0.3.1-0.20240121214520-5f936abd7ae8 // indirect
@@ -141,7 +132,6 @@ require (
 	github.com/spf13/pflag v1.0.10 // indirect
 	github.com/spf13/viper v1.21.0 // indirect
 	github.com/subosito/gotenv v1.6.0 // indirect
-	github.com/tinylib/msgp v1.6.3 // indirect
 	github.com/vishvananda/netns v0.0.5 // indirect
 	github.com/zeebo/xxh3 v1.1.0 // indirect
 	github.com/zitadel/logging v0.7.0 // indirect
@@ -157,7 +147,7 @@ require (
 	go.uber.org/mock v0.6.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
 	go.yaml.in/yaml/v3 v3.0.4 // indirect
-	golang.org/x/exp v0.0.0-20260218203240-3dfff04db8fa // indirect
+	golang.org/x/exp v0.0.0-20260312153236-7ab1446f8b90 // indirect
 	golang.org/x/mod v0.34.0 // indirect
 	golang.org/x/net v0.52.0 // indirect
 	gonum.org/v1/gonum v0.17.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -104,8 +104,6 @@ github.com/go-acme/lego/v4 v4.32.0 h1:z7Ss7aa1noabhKj+DBzhNCO2SM96xhE3b0ucVW3x8T
 github.com/go-acme/lego/v4 v4.32.0/go.mod h1:lI2fZNdgeM/ymf9xQ9YKbgZm6MeDuf91UrohMQE4DhI=
 github.com/go-chi/chi/v5 v5.2.5 h1:Eg4myHZBjyvJmAFjFvWgrqDTXFyOzjj7YIm3L3mu6Ug=
 github.com/go-chi/chi/v5 v5.2.5/go.mod h1:X7Gx4mteadT3eDOMTsXzmI4/rwUpOwBHLpAfupzFJP0=
-github.com/go-ini/ini v1.67.0 h1:z6ZrTEZqSWOTyH2FlglNbNgARyHG8oLW9gMELqKr06A=
-github.com/go-ini/ini v1.67.0/go.mod h1:ByCAeIL28uOIIG0E3PJtZPDL8WnHpFKFOtgjp+3Ies8=
 github.com/go-jose/go-jose/v4 v4.1.3 h1:CVLmWDhDVRa6Mi/IgCgaopNosCaHz7zrMeF9MlZRkrs=
 github.com/go-jose/go-jose/v4 v4.1.3/go.mod h1:x4oUasVrzR7071A4TnHLGSPpNOm2a21K9Kf04k1rs08=
 github.com/go-kit/log v0.1.0/go.mod h1:zbhenjAZHb184qTLMA9ZjW7ThYL0H2mk7Q6pNt4vbaY=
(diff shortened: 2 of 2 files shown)


doc/howto/storage_buckets.md
```diff
@@ -5,11 +5,10 @@ relatedlinks: "[LXD's&#32;S3&#32;API&#32;-&#32;YouTube](https://youtube.com/watc
 (howto-storage-buckets)=
 # How to manage storage buckets
 
-{ref}`storage-buckets` let you store and manage object-based data using either local or distributed storage.
+{ref}`storage-buckets` let you store and manage object-based data using distributed storage.
```
minaelee:
```suggestion
{ref}`storage-buckets` let you store and manage object-based data using non-local object storage.
```
To align with terms now being used in storage reference documentation.

doc/howto/storage_buckets.md
```diff
@@ -5,11 +5,10 @@ relatedlinks: "[LXD's&#32;S3&#32;API&#32;-&#32;YouTube](https://youtube.com/watc
...
 
 Unlike custom storage volumes, storage buckets cannot be attached to instances. Instead, applications access them directly via a URL using the S3 protocol.
 
-- For local buckets, the LXD server provides the S3-compatible URL via its {ref}`S3 address setting <howto-storage-buckets-create-requirements-local-s3>`.
 - For distributed buckets, a {ref}`Ceph RADOS Gateway endpoint <howto-storage-pools-ceph-requirements-radosgw-endpoint>` provides the S3-compatible URL.
```
minaelee:
```suggestion
- A {ref}`Ceph RADOS Gateway endpoint <howto-storage-pools-ceph-requirements-radosgw-endpoint>` provides the S3-compatible URL.
```

doc/howto/storage_buckets.md
```diff
@@ -46,8 +45,6 @@ Select {guilabel}`Buckets` from the {guilabel}`Storage` section of the main navi
 
 Storage buckets can only be created in storage pools that use a driver capable of **object storage**. View the {ref}`storage-buckets` reference guide's {ref}`storage-drivers-features` table to see which drivers support object storage.
```
minaelee:
```suggestion
```
We can delete this and skip directly to the information in line 51, since there is now only one driver for use with storage buckets.

doc/howto/storage_buckets.md
```diff
@@ -46,8 +45,6 @@ Select {guilabel}`Buckets` from the {guilabel}`Storage` section of the main navi
...
 
-Other requirements must be met before you can create a storage bucket, depending on whether you want to create a distributed or local storage bucket.
-
 (howto-storage-buckets-create-requirements-distributed)=
 #### Distributed storage buckets
```
minaelee:
```suggestion
```

A separate section for distributed storage buckets requirements is no longer needed.

doc/howto/storage_buckets.md
minaelee:
```suggestion
Your LXD server must have access to a {ref}`Ceph Object <storage-cephobject>` storage pool.
```

doc/howto/storage_buckets.md
```diff
@@ -83,63 +80,10 @@ Finally, click {guilabel}`Create bucket`.
...
-
-```bash
-lxc config set core.storage_buckets_address :8555
-```
-
 (howto-storage-buckets-create-single)=
 ### Create a bucket on a single, non-clustered LXD server
 
-To create a local or distributed storage bucket on a non-clustered LXD server, run:
+To create a distributed storage bucket on a non-clustered LXD server, run:
```
minaelee:
```suggestion
To create a storage bucket on a non-clustered LXD server, run:
```

doc/howto/storage_buckets.md
minaelee, line 93:
```suggestion
```

minaelee:
```suggestion
See the {ref}`Ceph Object <storage-cephobject>` documentation for a list of available storage bucket configuration options for the driver.
```

minaelee:
```suggestion
See the {ref}`Ceph Object <storage-cephobject>` documentation for a list of available storage bucket configuration options for the driver.
```

simondeziel:
@minaelee thanks a lot for the early feedback! I think I've incorporated all your suggestions.

lxd/storage/backend_lxd_patches.go
```diff
@@ -10,16 +10,14 @@ import (
...
 	"github.com/canonical/lxd/shared/logger"
 )
 
 var lxdEarlyPatches = map[string]func(b *lxdBackend) error{
-	"storage_missing_snapshot_records":         patchMissingSnapshotRecords,
-	"storage_delete_old_snapshot_records":      patchDeleteOldSnapshotRecords,
-	"storage_prefix_bucket_names_with_project": patchBucketNames,
+	"storage_missing_snapshot_records":    patchMissingSnapshotRecords,
+	"storage_delete_old_snapshot_records": patchDeleteOldSnapshotRecords,
 }
```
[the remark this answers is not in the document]
simondeziel: No because all the drivers still register the `storage_prefix_bucket_names_with_project` with `nil` in the patches maps.

lxd/patches.go
```diff
@@ -2268,3 +2269,37 @@ func patchVMSetMaxBusPorts(_ string, d *Daemon) error {
 }
 
 // Patches end here
+
+// patchStorageRemoveBucketDirectories removes the orphaned "buckets/" directories from local storage pools
```
tomponline: @simondeziel this will delete people's existing data?

lxd/storage/drivers/driver_cephobject_utils.go
```diff
@@ -196,6 +205,118 @@ func (d *cephobject) radosgwadminSubUserDelete(ctx context.Context, user string,
...
+// s3SignRequest signs an HTTP request using AWS Signature V4.
+func s3SignRequest(req *http.Request, creds S3Credentials, payloadHash string) {
+	if payloadHash == "" {
+		payloadHash = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855" // SHA-256 of empty string.
+	}
+
+	now := time.Now().UTC()
+	datestamp := now.Format("20060102")
+	amzDate := now.Format("20060102T150405Z")
+	region := "us-east-1"
```
tomponline, line 281: huh?

doc/api-extensions.md
```diff
@@ -2027,10 +2027,6 @@ This also introduces a change whereby network access is controlled by the projec
 * If `restricted.devices.nic` is set to `allow`, all networks are accessible (dependent on the `restricted.networks.access` setting).
 * If `restricted.devices.nic` is set to `block`, no networks are accessible.
 
-## `storage_buckets_local`
```
tomponline, line 2030: Please can you remove this as part of the removal of the api extension commit earlier, as we add these parts together, so it makes sense to me to remove them together. Ta

doc/explanation/storage.md
```diff
@@ -184,7 +184,7 @@ Instead, applications can access a storage bucket directly using its URL.
 
 Each storage bucket is assigned one or more access keys, which the applications must use to access it.
 
-Storage buckets can be located on local storage (with `dir`, `btrfs`, `lvm` or `zfs` pools) or on distributed storage (with {ref}`Ceph Object <storage-cephobject>` pools). For Ceph Object storage buckets, the {ref}`RADOS Gateway <howto-storage-pools-ceph-requirements-radosgw>` configured on the Ceph cluster acts as an S3 interface.
+Storage buckets must be located on distributed storage (with {ref}`Ceph Object <storage-cephobject>` pools). For Ceph Object storage buckets, the {ref}`RADOS Gateway <howto-storage-pools-ceph-requirements-radosgw>` configured on the Ceph cluster acts as an S3 interface.
```
tomponline: Lets lose the word "distributed" here and just the https://documentation.ubuntu.com/lxd/latest/reference/storage_drivers/#object-storage-backend section, rather than specific drivers.

We can keep the "For Ceph Object storage buckets," section though.

lxd/storage/drivers/driver_cephobject_utils.go
simondeziel: This is apparently the default region for S3-compatible services.Ceph RadosGW doesn't enforce region matching in signatures and only cares about it being consistently throughout the signing process.

MinIO also used it: https://github.com/minio/minio/blob/7aac2a2c5b7c882e68c1ce017d8256be2feea27f/cmd/server-main.go#L1155

lxd/patches.go
simondeziel: The `buckets` directories will be deleted **if empty**, otherwise the patch will fail. I must admit not fully understanding the ramifications of a patch failing at that point. Will LXD refuse to start? Will the user still be able to revert to an older LXD version with local storage bucket support?

I'm open to suggestions on how to improve the story for users relying on this feature we are pulling out.

roosterfish:
> Also @roosterfish please can you advise whether you want some/all of the removed recovery logic to be kept as I'm not sure if you are intending to add ceph object bucket recovery support at some point?

We parked this as a future improvement for the Ceph Object driver. 
Currently `detectUnknownBuckets` relies on a `drivers.Volume` being passed which we create from an unknown vol returned by the driver's `ListVolumes` func. 

I am almost certain that when we land this for Ceph Object, we anyway want to approach it differently and maybe grow a `ListBuckets` func on the driver struct that explicitly returns buckets (if supported by the driver) to not mix RadosGW buckets with vols.
So we would require to pass a different type into `detectUnknownBuckets` at which point we anyway need to rewrite the func. 
The `detectUnknownBuckets` is quite small so I don't think we have to keep it.

tomponline:
@simondeziel as discussed, lets block upgrade if the installation has pre-existing local bucket volume DB records.

simondeziel:
> Some questions:
> 
>     * Was there a patch for removing out-dated config keys from the DB?

Now added, thanks.

>     * what happens to the DB storage volume records and the on-disk volume data when the user updates to this version of LXD?

As suggested, there is now a pre-flight check. Here's what it looks like when trying to start this new LXD (custom debug bin) when there are still some storage buckets using local pools:

```
# lxc ls
Error: LXD unix socket not accessible: Get "http://unix.socket/1.0": EOF

# journalctl -o cat -n 15 -u snap.lxd.daemon
- shared_pidns
- cpuview_daemon
- loadavg_daemon
- pidfds
=> Starting LXD
==> WARNING: Using a custom debug LXD binary!
time="2026-03-10T16:48:08Z" level=error msg="Invalid configuration key: Unknown key" key=core.storage_buckets_address
time="2026-03-10T16:48:08Z" level=error msg="Invalid configuration key: Unknown key" key=core.storage_buckets_address
time="2026-03-10T16:48:08Z" level=error msg="Failed starting daemon" err="Failed to initialize global database: Failed to ensure schema: Failed ensuring schema: This LXD version no longer supports storage buckets backed by local storage drivers. Thi
s server has one or more such buckets and cannot be upgraded. To continue using those buckets, roll back to the previous LXD version. To proceed with the upgrade, back up and delete all locally-backed storage buckets first, then retry."
time="2026-03-10T16:48:08Z" level=warning msg="Dqlite last entry" index=0 term=0
Error: Failed to initialize global database: Failed to ensure schema: Failed ensuring schema: This LXD version no longer supports storage buckets backed by local storage drivers. This server has one or more such buckets and cannot be upgraded. To co
ntinue using those buckets, roll back to the previous LXD version. To proceed with the upgrade, back up and delete all locally-backed storage buckets first, then retry.
Killed
=> LXD failed to start
snap.lxd.daemon.service: Main process exited, code=exited, status=1/FAILURE
snap.lxd.daemon.service: Failed with result 'exit-code'.
```

Removing the storage buckets allows the upgrade to proceed and I confirmed the `core.storage_buckets_address` config to be removed on upgrade.

simondeziel:
> @simondeziel as discussed, lets block upgrade if the installation has pre-existing local bucket volume DB records.

I think I've addressed those so please take another look. The remaining failure in CI are due to `latest/edge` being temporarily broken due to the `liblxc` regression.

lxd/db/cluster/open.go
```diff
@@ -141,6 +141,11 @@ func EnsureSchema(db *sql.DB, address string, dir string, serverUUID string) err
 			return fmt.Errorf("Failed updating cluster member version info for %q: %w", address, err)
 		}
 
+		err = checkNoLocalStorageBuckets(ctx, tx)
+		if err != nil {
+			return err
+		}
+
 		return checkClusterIsUpgradable(ctx, tx, [2]int{len(updates), apiExtensions})
```
tomponline, line 149: Would it make sense to call `checkNoLocalStorageBuckets` from `checkClusterIsUpgradable` seeing as that is the reason we are doing it?

Keeps all the "can we upgrade checks" together?

lxd/db/cluster/update.go
```diff
@@ -123,6 +123,14 @@ var updates = map[int]schema.Update{
 	78: updateFromV77,
 	79: updateFromV78,
 	80: updateFromV79,
+	81: updateFromV80,
+}
+
+func updateFromV80(ctx context.Context, tx *sql.Tx) error {
```
tomponline: As this is not a schema change we should do it using the patch system, e.g. like

```
	{name: "candid_rbac_remove_config_keys", stage: patchPreDaemonStorage, run: patchRemoveCandidRBACConfigKeys},
```

No need to use ` tx.UpdateClusterConfig` though.

lxd/patches.go
tomponline: Resolved by blocking updates.

lxd/db/cluster/update.go
tomponline: You can use `patchStorageRemoveBucketDirectories` and rename to `patchStorageRemoveLocalBuckets` and do both in there.

lxd/storage/backend_lxd_patches.go
```diff
@@ -173,62 +171,3 @@ DELETE FROM storage_volumes WHERE id IN (
 
 	return nil
 }
-
-// patchBucketNames modifies the naming convention of bucket volumes by adding
-// the corresponding project name as a prefix.
-func patchBucketNames(b *lxdBackend) error {
```
tomponline, line 179: Its good we are blocking upgrades if local buckets are defined, otherwise removing this patch would potentially break the patch you are adding.

lxd/db/cluster/update.go
simondeziel: Indeed, cleaner, thanks.

lxd/db/cluster/open.go
simondeziel: `checkNoLocalStorageBuckets` is not only for clusters so I don't think that's a better call site.

tomponline: Ah OK fair enough.

Issue #17641, opened 2026-02-13
Improve operation metadata standardisation


markylaing:
Create constants representing operation metadata fields and enforce per operation type validation.

> An idea for future PR, not a blocker, but for common/shared metadata fields (like this one, the secret one above, and the entity name one we introduced recently) it would be good to have shared constants in the `api` package we can refer to on client and server side in order to easily see usage.
> 
> On that point, could this use `entity_name` rather than introduce a new field? 

 _Originally posted by @tomponline in [#17607](https://github.com/canonical/lxd/pull/17607/changes/e7842f2d7d268a23c3615813d71d0d98838fe75e#r2803239318)_

markylaing:
Tests for this work should include enforcing that the entity URL is present in operation metadata for certain operation types. See https://github.com/canonical/lxd/pull/17664#issuecomment-3909310857

markylaing:
This should also include standardisation of operation progress reporting. There should only ever be one key that ends in `_progress`.

markylaing:
Also standardise event lifecycle metadata to use `entity_url` and not use `old_name` for rename operations.

markylaing:
Agreed to *always* add `entity_url` field to metadata at operation create time when:
- Creating an entity (expected new name).
- Deleting an entity (current name).
- Renaming an entity (expected new name).
- Updating an entity (current name)

For renames we will also add `original_entity_url`, containing the URL of the entity before it was renamed.

Also, if the entity URL is `/1.0`, then lets not add it to metadata as it doesn't give any useful information to the caller.

tomponline:
@markylaing this is done now right?

markylaing:
I'd say so - We've standardised progress reporting and entity URLs. Last thing will be to set the entity URL in metadata to have the same project as the operation

PR #17676, opened 2026-02-17
Operation progress metadata

markylaing:
Ensures that there is only one key ending "_progress" on calls to `(*Operation).UpdateProgress`, which now replaces calls to `shared.UpdateProgressMetadata`. We still need to audit other keys ending `_progress`. I've added this as a prerequisite of #17641. For now, this fixes flickering over various progress keys in the CLI when creating an instance.

## Checklist

### DIFF
--- a/lxd/images.go
+++ b/lxd/images.go
@@ -442,10 +442,7 @@ func imgPostInstanceInfo(s *state.State, req api.ImagesPost, op *operations.Oper
 					processed = value
 				}
 
-				metadata := make(map[string]any)
-				shared.SetProgressMetadata(metadata, "create_image_from_instance_pack", "Image pack", percent, processed, speed)
-				// Replace metadata to show only current progress, avoiding stale keys from previous stages.
-				_ = op.UpdateMetadata(metadata)
+				_ = op.UpdateProgress("create_image_from_instance_pack", "Image pack", percent, processed, speed)
 			},
 			Length: totalSize,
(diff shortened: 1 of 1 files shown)


markylaing:
To avoid conflicts with @skozina I've tried to change as little as possible here. But also to avoid duplication I'm copying the existing metadata, making modifications, and then calling `UpdateMetadata` to actually update the metadata and emit the event.

I've tested that this works and fixes the CLI glitch,.

lxd/operations/operations.go
```diff
@@ -754,3 +757,51 @@ func validateMetadata(metadata map[string]any) (map[string]any, error) {
 
 	return metadata, nil
 }
+
+// UpdateProgress updates the operation metadata with progress information, including
+// the percentage complete, data processed, and speed. It formats and stores these values for
+// both API callers and CLI display purposes.
+func (op *Operation) UpdateProgress(stage string, displayPrefix string, percent int64, processed int64, speed int64) error {
```
tomponline, line 764: Should this function lock the operation so its concurrent update safe?

markylaing: It get's a copy of the current metadata while under lock. Then mutates the copy and calls UpdateMetadata to replace the existing one. So it should be concurrent. We can do better in terms of code duplication here but I wanted to try and avoid more rebase headaches for Stan.

PR #17677, opened 2026-02-17
`golangci-lint` pinning

simondeziel:
* Add `tools/` Go module that depends on `github.com/golangci/golangci-lint/v2` to pin its version.
* Updated the CI workflow to read the pinned `golangci-lint` version and pass it to `golangci-lint-action`.
* Updated `Makefile` and Dependabot config to manage and update the `tools/` module alongside the root module.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -187,11 +187,15 @@ jobs:
           cd /home/runner/work/lxd/lxd-test
           make
 
+      - name: Get pinned golangci-lint version
+        id: golangci-lint-version
+        run: echo "version=$(go list -C tools -m -f '{{.Version}}' github.com/golangci/golangci-lint/v2)" >> "$GITHUB_OUTPUT"
+
       - name: Run golangci-lint
         if: github.event_name == 'pull_request' || github.event_name == 'schedule'
         uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
         with:
-          version: v2.10.1
+          version: ${{ steps.golangci-lint-version.outputs.version }}
 
       - name: Run static analysis
         if: ${{ github.event_name != 'workflow_dispatch' || inputs.static-analysis == 'true' }}


.github/workflows/tests.yml
```diff
@@ -187,11 +187,11 @@ jobs:
...
           make
 
-      # XXX: `make static-analysis` also run golangci-lint but this one provides
-      #      useful feedback in the PR through github-code-scanning bot
       - name: Run golangci-lint
         if: github.event_name == 'pull_request' || github.event_name == 'schedule'
-        uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v8
+        uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
+        with:
+          version: v2.10.1
```
tomponline, line 194: `#v9.2.0` and `version: v2.10.1` is odd?

Makefile
```diff
@@ -276,6 +280,7 @@ endif
 
 	# Enforce minimum go version
 	go mod tidy -go=$(GOMIN)
+	go mod -C tools tidy -go=$(GOMIN)
```
tomponline, line 283: what does "-C" do, it doesn't appear in `go help mod tidy`

.github/workflows/tests.yml
simondeziel: 9.2.0 is the version of the action itself (which Dependabot happily manages) and 2.10.1 is the `golangci-lint` own version which because it's an action input parameter, is normally outside of Dependabot's purview.

tomponline:
@simondeziel would you mind updating the PR description of how you managed to solve this issue of getting dependabot to update golangci-lint using a go module?

Makefile
simondeziel: AFAIK, most (all?) `go` sub-commands support `-C` to change directory, probably inspired by `tar` and company.
`go help build` mentions it at least.

We have a few more users of this facility as it is nicer than having to change dir twice (enter/exit).

tomponline: thanks! I hate short-form flags.

